PR #3789, opened 2018-01-13
Reduce max line length for a diff down to 5k

### LINKED ISSUE #3740: Decrease the max line length limit
As per https://github.com/desktop/desktop/pull/3032#issuecomment-354265785, we can probably cut our max line length limit down quite a bit. It's currently at 500,000. We could try 100,000 or even 50k.

See https://github.com/desktop/desktop/blob/b0d5afb37c36116006bb44855c9e3e81d3f554d7/app/src/lib/git/diff.ts#L41-L45

nerdneha:
Fixes #3740 

Before: 500k (whyy)
After: 5k

### DIFF
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -42,7 +42,7 @@ const MaxReasonableDiffSize = 3000000
  * The longest line length we should try to display. If a diff has a line longer
  * than this, we probably shouldn't attempt it.
  */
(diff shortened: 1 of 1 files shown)


nerdneha:
findings: this works fine for new files but not for existing files. it's a simple change but some thought might go into taking care of existing files and editing the error message. free for someone else to pick up and give another go

PR #3795, opened 2018-01-14
Add a repository settings store

### LINKED ISSUE #934: Introduce a RepositorySettingsStore
See https://github.com/desktop/desktop/pull/860#discussion_r99871828 for a bit more context.

The TL;DR is that right now we're a bit inconsistent in the way we use the dispatcher to read the gitignore, and a better solution would probably be to introduce a `RepositorySettingsStore`.

iAmWillShepherd:
Fixes #934 

- [x] Add store for repo settings
- [x] Move functions that should be in new store from git store
- [x] Fix tests

This PR introduces a repository settings store. 

I moved several functions from `GitStore` were moved into the `RepositorySettingsStore` since they just deal with settings/config for a particular repo.

**Functions moved from `GitStore`**
* `readGitIngore`
* `saveGitIgnore`
* `ignore`
* `formatGitIgnoreContents` _(helper function)_

I also extracted out some common functionality into the abstract class `BaseStore`.
* `emitUpdate`
* `emitError`
* `onDidUpdate`
* `onDidError`

The implications of this changes mean that we no longer need to create an emitter or setup these functions in new stores. We are also able to override this base functionality by merely redefining the function in the implementing class.

I also took the liberty of cleaning up some of the imports I came across, updated some tests to test the things they claimed and added some to exercise the new store.


iAmWillShepherd:
One thing to note is that I'm asserting that a `Repository` is a `GitHubRepository` in the `PullRequestStore` because I believe that is the proper place for this to happen. 

Thoughts from @desktop/core?

app/test/unit/git-store-test.ts
```diff
@@ -23,7 +23,7 @@ import { TipState, IValidBranch } from '../../src/models/tip'
 import { getCommit, getStatus } from '../../src/lib/git'
 
 describe('GitStore', () => {
-  it('can discard changes from a repository', async () => {
+  it.only('can discard changes from a repository', async () => {
```
j-f1: Is `.only` here by accident?

iAmWillShepherd: Nope, forgot to remove it! 🙇

app/src/lib/stores/store.ts
```diff
@@ -0,0 +1,28 @@
...
+  protected emitUpdate(data?: any) {
+    this._emitter.emit('did-update', data || {})
+  }
+
+  protected emitError(error: Error) {
+    this._emitter.emit('did-error', error)
+  }
+
+  /** Register a function to be called when the store updates. */
+  public onDidUpdate(fn: () => void): Disposable {
```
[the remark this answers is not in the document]
shiftkey: Oh, and `AppStore` also has the same sorts of workarounds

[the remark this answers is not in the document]
iAmWillShepherd: Yea, I like the idea of using two base stores, but I'd prefer to name them `BaseStore` and `BaseStore<T>` (Assuming TypeScript allows this, haven't checked yet).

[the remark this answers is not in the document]
iAmWillShepherd: Looks like, it's not allowed. Will use unique names for the base store. 😢

iAmWillShepherd:
🔢

app/src/lib/stores/app-store.ts
```diff
@@ -159,36 +163,33 @@ export class AppStore {
   private currentPullRequestUpdater: PullRequestUpdater | null = null
 
   private repositoryState = new Map<string, IRepositoryState>()
-  private showWelcomeFlow = false
-
+  private _showWelcomeFlow = false
```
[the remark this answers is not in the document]
iAmWillShepherd: In one word, consistency. Not married to it though.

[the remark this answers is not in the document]
shiftkey: Yeah, I see the `_` here as being non-standard so that it stands out and make the reader feel uncomfortable.

Can we undo these changes here to reduce the churn as part of this PR?

iAmWillShepherd:
🔢

app/src/lib/stores/app-store.ts
```diff
@@ -175,15 +175,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
     string,
     RepositorySettingsStore
   >()
-  public readonly _gitHubUserStore: GitHubUserStore
-  private readonly _cloningRepositoriesStore: CloningRepositoriesStore
-  private readonly _emojiStore: EmojiStore
+  public readonly gitHubUserStore: GitHubUserStore
+  private readonly cloningRepositoriesStore: CloningRepositoriesStore
+  private readonly emojiStore: EmojiStore
   private readonly _issuesStore: IssuesStore
```
j-f1: `issuesStore`?

iAmWillShepherd: Nope. it was `_issuesStore` before I started mucking around in this file. When I deleted the underscore,  TS complained about `issuesStore` already being defined. This is likely because the getter is the same name as the field (another reason why I prefer differentiating private members and public properties).

app/src/lib/stores/base-store.ts
```diff
@@ -0,0 +1,55 @@
+import { Emitter, Disposable } from 'event-kit'
+
+export abstract class BaseStore {
+  protected readonly _emitter = new Emitter()
```
shiftkey: Because this has the right access modifier, let's rename this to `emitter` rather than spread the the `_` habit around.

app/src/lib/stores/app-store.ts
```diff
@@ -352,23 +358,10 @@ export class AppStore {
     this.emitQueued = false
     const state = this.getState()
 
-    this.emitter.emit('did-update', state)
+    this._emitter.emit('did-update', state)
```
shiftkey: Minor thing, but we should call `super.emitUpdate(state)` here in case the underlying `eventName` is changed.

iAmWillShepherd:
🔢

shiftkey:
I just pushed 2355e018457e51e534621cc184765a4fd884b828 which tidies up some other conventions I overlooked in the last review, rather than drag this out for another cycle.

PR #3796, opened 2018-01-14
Allows double dashes in branch name

### LINKED ISSUE #3599: Required formatting for branch name not allowed
### Description

The GitHub desktop app currently helps automatically formats the branch name with some variation of kebab-case. However, this is proving a little bit too strict for my company's workflow. We have some automatic actions that run when a name branch is suffixed with `--preview`. The desktop app automatically replaces multiple dashes with a single dash, so something like `branch-name--preview` becomes `branch-name-preview`.

I would very much appreciate it if the app allowed for a little bit more freedom with how users name their branches.

### Version

**GitHub Desktop version:** 10.0.10

**OS version:** MacOS 10.13.1

### Steps to Reproduce

1. Branch -> New Branch
1. Name a branch with--multiple--dashes---
1. The branch name will be reformatted to only allow one dash between words.

**Expected behavior:** More flexible naming conventions are allowed

**Actual behavior:** Strict kebab-case is applied

**Reproduces how often:** Always

JQuinnie:
Fixes #3599 

## Before
test-branch-before

## After
test-branch-after

### DIFF
--- a/app/src/lib/sanitize-branch.ts
+++ b/app/src/lib/sanitize-branch.ts
@@ -8,7 +8,6 @@ const invalidCharacterRegex = /[\x00-\x20\x7F~^:?*\[\\|""<>]|@{|\.\.+|^\.|\.$|\.
 export function sanitizedBranchName(name: string): string {
   return name
     .replace(invalidCharacterRegex, '-')
-    .replace(/--+/g, '-')
     .replace(/^-/g, '')
 }
 
--- a/app/src/lib/sanitize-branch.ts
+++ b/app/src/lib/sanitize-branch.ts
@@ -6,9 +6,7 @@ const invalidCharacterRegex = /[\x00-\x20\x7F~^:?*\[\\|""<>]|@{|\.\.+|^\.|\.$|\.
 
 /** Sanitize a proposed branch name by replacing illegal characters. */
 export function sanitizedBranchName(name: string): string {
-  return name
-    .replace(invalidCharacterRegex, '-')
-    .replace(/^-/g, '')
+  return name.replace(invalidCharacterRegex, '-').replace(/^-/g, '')
 }
 
 /** Validate a branch does not contain any invalid characters */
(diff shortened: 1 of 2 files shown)


JQuinnie:
I am willing to fix [this test](https://github.com/desktop/desktop/issues/3599#issuecomment-357550817) as a part of this PR, if anyone would like.

shiftkey:
@JQuinnie that test is no longer relevant - feel free to remove it

JQuinnie:
remove [this test](https://github.com/desktop/desktop/blob/51c8b182c88189e127e78e84f4fe09f38c3d5e83/app/test/unit/sanitize-test-name-test.ts#L42-L46)  as mentioned [above](https://github.com/desktop/desktop/pull/3796#issuecomment-357720532) to fix #3796

shiftkey:
@JQuinnie there's also a couple of affected tests in [`sanitize-test-name-test.ts`](https://github.com/desktop/desktop/blob/master/app/test/unit/sanitize-test-name-test.ts) that should be updated or removed.

iAmWillShepherd:
What do we do when we encounter consecutive invalid characters? 

At the moment, the regex doesn't seem to be greedy, so we produce _n_ matches for _n_ invalid characters.

https://github.com/JQuinnie/desktop/blob/928d2082e6befa287295e1ed8dddedb78be004a3/app/src/lib/sanitize-branch.ts#L5

Given `bad:?|--branch_name&-`, this regex will produce 4 matches (`:`, `?`, `|`, and `&`). Is this what we want? I'm thinking it would be better to update the regex so that we return one match for a run of invalid characters. So, `bad:?|--branch_name&-` would return 2 matches (':?|' and `&`).

**updated regex:**
`[\x00-\x20\x7F~^:?*\[\\|""<>]+|@{|\.\.+|^\.|\.$|\.lock$|\/$`

shiftkey:
>  I'm thinking it would be better to update the regex so that we return one match for a run of invalid characters. So, `bad:?|--branch_name&-` would return 2 matches (':?|' and `&`).

I like this idea, as it's closer to the original spirit of the regex without intefering with valid dashes.

iAmWillShepherd:
@JQuinnie I updated your body by removing the 2nd fixes since that issue number is the issue number of this PR.

Issue #3834, opened 2018-01-20
Caret always jumps to the end when inserting a completion


CarlRosell:
### Description
When inserting an emoji or mention in the commit message, the caret will always jump to the end of the text.

### Version

**GitHub Desktop version:** 8304e9c271a5e5ab4fda797304cd7bcca7158c87

**OS version:** macOS High Sierra

### Steps to Reproduce

1. Start writing a commit message
2. Go to the beginning of the message
3. Insert an emoji or mention

**Expected behavior:** The caret should stay right where the completion was inserted

**Actual behavior:** The caret jumps to the end of the message

**Reproduces how often:** 100%

shiftkey:
This feels very similar to #3354 which we fixed by being more careful with props and `setState` changes - see https://github.com/desktop/desktop/pull/3411 for the fix.

I'll look at #3835 in the meantime but it'd be nice to see if reusing our fixed `TextBox` component can help to address this problem.

shiftkey:
> I'll look at #3835 in the meantime but it'd be nice to see if reusing our fixed `TextBox` component can help to address this problem.

This requires some refactoring of the autocomplete providers, and probably is worth deferring if #3835 addresses the problem well enough.

PR #3835, opened 2018-01-20
Fix caret position after inserting completion

CarlRosell:
Fixes: #3834.


app/src/ui/autocompletion/autocompleting-text-input.tsx
```diff
@@ -295,17 +286,28 @@ export abstract class AutocompletingTextInput<
     const autoCompleteText = autocompletionState.provider.getCompletionText(
       item
     )
+
+    const textWithAutoCompleteText =
+      originalText.substr(0, range.start - 1) + autoCompleteText + ' '
```
CarlRosell: This seems to be the current behavior on github.com and the old GitHub client. 
Just adding an extra space right after the inserted completion and position the caret there. Some other applications also do it this way. 
Slack seems to do something a bit less naive by actually looking at the next character and only inserting the extra space if the next character isn't one. 
Facebook will not add anything extra at all.

I chose to do it as the other GitHub clients to provide a unified experience.

shiftkey: Yep, makes sense.

app/src/ui/autocompletion/autocompleting-text-input.tsx
```diff
@@ -295,17 +286,28 @@ export abstract class AutocompletingTextInput<
...
+      textWithAutoCompleteText + originalText.substr(range.start + range.length)
+
     element.value = newText
 
     if (this.props.onValueChanged) {
       this.props.onValueChanged(newText)
     }
 
+    // This is pretty gross. Clicking on the list moves focus off the text area.
+    // Immediately moving focus back doesn't work. Gotta wait a runloop I guess?
```
shiftkey: This is already an issue with the current app, and I'm not sure of alternatives to fix this.

@niik any other ideas to handle clicking the autocomplete list item stealing focus from the related text element?

niik: Not of the top of my head, no. But we don't have to solve that in this PR. But it seems like this PR changes to doing the focus hack regardless of whether the autocomplete happened as the result of a keyboard or pointer action. We should _only_ do this onClick.

shiftkey: @niik good shout. I'll see if I can work with @CarlRosell to handle keyboard autocomplete differently to pointer autocompletes :thumbsup:

shiftkey:
@CarlRosell I've added some code to differentiate between keyboard-based autocomplete and mouseclick-based autocomplete, and tidied it up. Feel free to test this and ensure you're happy with the changes, and we can get @niik to sign off on it.

CarlRosell:
Hi @shiftkey and @niik and sorry for my late response. The whole family got the flu and life been chaotic the last two weeks 😅. The changes look good to me and thanks for keeping this alive 👍

PR #2488, opened 2017-08-17
Don't dismiss dialogs when focusing

joshaber:
Fixes #2486

This is a hack. I'm open to better ideas, but I couldn't find a way to associate the click event with the focus event 😞

PR #3843, opened 2018-01-22
Allow window to accept single click on focus

iAmWillShepherd:
This PR addresses feedback from @kevinsawicki: the window required one click to set focus and another to actually perform any useful actions.

- [x] Test on macOS
- [x] Test on Windows
- [x] Make sure dialogs don't respond to focus event
  - [x] macOS
  - [x] Windows

### DIFF
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -52,6 +52,7 @@ export class AppWindow {
         // Enable, among other things, the ResizeObserver
         experimentalFeatures: true,
       },
+      acceptFirstMouse: true,
     }
 
     if (__DARWIN__) {


shiftkey:
Does this make #2488 unnecessary, or is that still required?

iAmWillShepherd:
I don't believe it does. I think that issue was about dialogs responding to any focus event emitted by the renderer process.

nerdneha:
@iAmWillShepherd code looks fine but i'm not getting this to work on my computer

act-more-native

iAmWillShepherd:
I believe this changes requires a rebuild first. @nerdneha Did you rebuild the app before starting?

nerdneha:
it's working

tierninho:
Spotted this: You can right-click into the app, but the focus is still on the Finder menu. 
screen shot 2018-01-31 at 1 53 41 pm

shiftkey:
@tierninho that smells like an Electron issue. If it's reproducible away from Desktop, like using `electron-quick-start`, then we should log it upstream but still take in this change. It might have been fixed in `1.8.x`.

iAmWillShepherd:
@tierninho @shiftkey, turns out this behavior is present in my browser as well (Firefox). Just opened a terminal,  right-clicked in my browser so that a context menu appears, and I typed a little message to myself in my terminal (iTerm2)

shiftkey:
@iAmWillShepherd cool, so it might even be beyond Electron. Anyway, no need to hold this up.

zenmumbler:
A bit late to the party, but this one just shipped in 1.1.0 so I saw this just now.
Clicking once to focus window and then performing another click to act in it IS the native behaviour, at least on macOS. You might want to try it with other apps too. The actual bug / inconsistent behaviour with the GH app is that it does not show that it is inactive by greying out window chrome and interface elements, etc. This behaviour is now even more inconsistent with actual macOS apps than before. This behaviour may be different on Windows, but as far as macOS is concerned, a better name for this branch would be act-even-less-native.

shiftkey:
@zenmumbler 

>  The actual bug / inconsistent behaviour with the GH app is that it does not show that it is inactive by greying out window chrome and interface elements, etc.

This is an interesting bit of feedback. Would you mind opening up a fresh issue about this with an annotated screenshot of the bits that you think need to be changed, so we can track this discussion?

zenmumbler:
Will do, will have time to do this properly later today. Thanks for the response!

iAmWillShepherd:
@zenmumbler , I think I understand the nature of the problem. I used to native macOS apps, TextEdit and Notes. They both respond to a click when out of focus, as does Desktop now, but they also grey out the contents of the window, which Desktop does not do. 



Is my understanding correct and does this gif correctly show the problem?

PR #3844, opened 2018-01-23
render the badge without the commit status if missing

### LINKED ISSUE #3608: Add Pull Request badge number in nav bar if no CI status
If CI is turned on, then any previous Pull Requests without CI status are missing the badge number in the top nav bar when the user selects the branch or pull request.

screen shot 2017-12-14 at 9 24 03 am

I have Travis CI running .... for testing purposes.

1.0.12-beta0, mac

shiftkey:
Fixes #3608

### DIFF
--- a/app/src/ui/branches/pull-request-badge.tsx
+++ b/app/src/ui/branches/pull-request-badge.tsx
@@ -15,21 +15,13 @@ export class PullRequestBadge extends React.Component<
   IPullRequestBadgeProps,
   {}
 > {
-  private renderCIStatus(): JSX.Element | null {
+  public render() {
     const status = this.props.status
 
-    if (!status || status.totalCount === 0) {
-      return null
-    }
-
-    return <CIStatus status={status} />
-  }
-
-  public render() {
     return (
       <div className="pr-badge">
         <span className="number">#{this.props.number}</span>
-        {this.renderCIStatus()}
+        {status && status.totalCount ? <CIStatus status={status} /> : null}
       </div>
     )
   }
--- a/app/src/ui/branches/pull-request-badge.tsx
+++ b/app/src/ui/branches/pull-request-badge.tsx
@@ -15,18 +15,21 @@ export class PullRequestBadge extends React.Component<
   IPullRequestBadgeProps,
   {}
 > {
-  public render() {
+  private renderCIStatus(): JSX.Element | null {
     const status = this.props.status
 
     if (!status || status.totalCount === 0) {
       return null
     }
 
+    return <CIStatus status={status} />
+  }
+
+  public render() {
     return (
       <div className="pr-badge">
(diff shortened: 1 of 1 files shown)


app/src/ui/branches/pull-request-badge.tsx
```diff
@@ -15,18 +15,21 @@ export class PullRequestBadge extends React.Component<
...
+    return <CIStatus status={status} />
+  }
+
+  public render() {
     return (
       <div className="pr-badge">
         <span className="number">#{this.props.number}</span>
-
-        <CIStatus status={status} />
+        {this.renderCIStatus()}
```
j-f1:
```ts
status && status.totalCount ? <CIStatus status={status} /> : null
```
?

j-f1: I would’ve suggested `status && status.totalCount && <CIStatus status={status} />`, but then the `0` count would get rendered as a literal “0” in the app.

app/src/ui/branches/pull-request-badge.tsx
```diff
@@ -18,15 +18,10 @@ export class PullRequestBadge extends React.Component<
...
-    if (!status || status.totalCount === 0) {
-      return null
-    }
-
     return (
       <div className="pr-badge">
         <span className="number">#{this.props.number}</span>
-
-        <CIStatus status={status} />
+        {status && status.totalCount ? <CIStatus status={status} /> : null}
```
iAmWillShepherd: Change `status && status.totalCount` to `status != null && status.totalCount > 0`.

j-f1: 👍, and that’ll let you use `status != null && status.totalCount > 0 && <CIStatus status={status} />` since both `status != null` and `status.totalCount > 0` evaluate to booleans, which don’t get rendered.

shiftkey: @j-f1 I'm just going to leave it as what @iAmWillShepherd said, because I think we should keep the explicit `null` around rather than relying on some obscure React behaviour about rendering booleans.

shiftkey:
:apple:

app/src/ui/branches/pull-request-badge.tsx
j-f1: > some obscure React behaviour about rendering booleans

It’s a [recommended practice](https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored) in their docs.

shiftkey: @j-f1 I'm glad it's in their docs (I honestly haven't tried rendering a boolean yet) but passing `null` is an organic convention we use elsewhere in the codebase, especially with functions that return `JSX.Element | null`.

PR #3846, opened 2018-01-23
bump electron to address security issue

shiftkey:


niik:
I don't understand how this didn't break protocol handlers on Windows. We implemented our own workaround for the RCE attack in https://github.com/desktop/desktop/commit/bb406b63a371215ee5916e4e1bdf687a51bf74fe and I was under the impression that adding the official workaround (which is identical to ours) would break protocol handlers on Windows since we're doing pretty strict checking of the exact format of arguments.

I'm gonna spin up a Windows instance and verify that it still works, it seems odd that we wouldn't have gotten any bug reports for it so I'm assuming it still works somehow.

niik:
Aha! The fix in electron was more clever than I thought and it's backwards compatible with the workaround. Very nice!

PR #3588, opened 2017-12-12
Show tooltip with additional info about the build status

iAmWillShepherd:
Fixes #3134

### After
GitHub Desktop

VS Code

PR #3866, opened 2018-01-26
Database Cleanup

iAmWillShepherd:
Fixes issue ~#3903~, #2954 and #1144.

This PR is intended to make doing database operations easier and less error-prone. 

- [ ] Make sure database usage is consistent across stores
- ~[ ] Make entity field names consistent (I'll be using naming conventions in SQL; ex `createdAt` -> `created_at~`
- [ ] ~Prefix `_` to all entity IDs to make it clearer that it's not to be modified directly; ex `id` -> `_id`~
- [ ] Make distinction between `api` and `entity`results more clear 
- [ ] Add `api -> entity` conversion functions
- [ ] Make sure `api` and `entity` types do not leak from stores
- [ ] Add some tests here and there
- [ ] Remove all dependencies on API from databases; let's keep these concerns separate
- [ ] Ensure we delete records (trying to eliminate orphaned records)

## Current
github desktop

## Proposed [WIP]
github desktop proposed

**Legend**
* `Database` **green**
* `GitHubUserDatabase` **blue**
* `IssuesDatabase` **orange**
* `PullRequestDatabase` **red**
* ~`StatsDatabase`~

_Note that I did not include the `StatsDatabase` in the ER Diagram._

PR #3869, opened 2018-01-27
Work around PR status db object missing statuses property

niik:
As I spun up desktop@master on my Windows machine today I got this error

```
Uncaught (in promise) TypeError: Cannot read property 'map' of undefined
    at PullRequestStore.getPRStatusById (pull-request-store.ts:237)
    at <anonymous>
```

It seems the `statuses` property was added in #3588 and typed as `ReadonlyArray<IAPIRefStatusItem>` but since no corresponding database upgrade function was added to ensure that the property exists on prior database objects we can't rely on it being there.

I could add an upgrade function but seeing as we're pretty close to a release I would rather just mark the property accurately and make sure we don't have any other places which rely on it always being present.

We can revisit and clean this up in a later version.

### DIFF
--- a/app/src/lib/databases/pull-request-database.ts
+++ b/app/src/lib/databases/pull-request-database.ts
@@ -60,8 +60,12 @@ export interface IPullRequestStatus {
   /** The SHA for which this status applies. */
   readonly sha: string
 
-  /** The list of statuses for this specific ref */
-  readonly statuses: ReadonlyArray<IAPIRefStatusItem>
+  /**
+   * The list of statuses for this specific ref or undefined
+   * if the database object was created prior to status support
+   * being added in #3588
+   */
+  readonly statuses: ReadonlyArray<IAPIRefStatusItem> | undefined
 }
 
 export class PullRequestDatabase extends Dexie {
--- a/app/src/lib/stores/pull-request-store.ts
+++ b/app/src/lib/stores/pull-request-store.ts
@@ -234,7 +234,7 @@ export class PullRequestStore {
       return null
     }
 
-    const combinedRefStatuses = result.statuses.map(x => {
+    const combinedRefStatuses = (result.statuses || []).map(x => {
       return {
         id: x.id,
         state: x.state,


shiftkey:
This all seems reasonable. I'll let @iAmWillShepherd give it a look over as he's familiar with these stores and might have additional input.

niik:
With the review from @shiftkey, I'm gonna go ahead and merge this because we need it for the release. We can address any feedback @iAmWillShepherd might have in a subsequent PR

iAmWillShepherd:
> I could add an upgrade function but seeing as we're pretty close to a release I would rather just mark the property accurately and make sure we don't have any other places which rely on it always being present.

@niik turns out there there is no upgrade function needed here since we are not indexing on that property. See Dexie's [docs](http://dexie.org/docs/Tutorial/Design#database-versioning) on database versioning.

What's likely the culprit here is that statuses can be `undefined` when it should probably be required and set to an empty array if a PR does not have any status available. Since not all repos use integration, having a value that represents no statuses is necessary.

https://github.com/desktop/desktop/blob/86edf4c6e1198b19dac474adafe314931097dbdd/app/src/lib/databases/pull-request-database.ts#L68

shiftkey:
@iAmWillShepherd is there a more elegant fix to the problem than what @niik has done? If so, we should get that in and live a happy life™

iAmWillShepherd:
@shiftkey I'm still working that out in #3866. I'm likely going to make sure we always insert something for the status field (something like `[]`) so it can never be `undefined`.

niik:
> @niik turns out there there is no upgrade function needed here since we are not indexing on that property. See Dexie's docs on database versioning.

You're right that we don't need to change the indexes. But the schema has nonetheless changed and now we're left with inconsistent data in the table. When we add a required property to a model that's persisted in indexeddb we need to either explicitly type it as possibly undefined (what this PR did) or we could add a faux db version with an upgrade function which Dexie will take care of calling once where we either clear the table so that fresh, consistent data can come in, or iterate over the records and ensure that we insert an empty array for each property (where that is faster than clearing the table and starting from scratch).

See https://github.com/desktop/desktop/blob/5c1c139641726e282094a9cc8b384099b0893062/app/src/lib/databases/issues-database.ts#L29-L49 for an example of where we cleared the table and also made sure that related data that's not stored in indexeddb was updated.

PR #3879, opened 2018-01-29
Commit together with co-authors

niik:
*This PR has been review internally*

Adds support for quickly adding co-authors to your commit. Great for pairing, excellent for sending a little love to that special someone who helped fix that gnarly bug of yours.

co-authors-demo

Co-authors works for any GitHub or GitHub Enterprise repository and the autocomplete box is populated with the same list of users as you'd see doing an @-mention in an issue. If you need to add someone who hasn't contributed to your repository in any way (issue comments etc) you can still type their GitHub username manually and we'll look them up for you.

What happens behind the scenes when you add a co-author is that Desktop will add a `Co-Authored-By` commit message trailer that looks something like this

```
Co-Authored-By: Foo Bar <foo@bar.com>
```

Or, if the user you're adding has chosen to keep their email private

```
Co-Authored-By: Foo Bar <foo@users.noreply.github.com>
```

This is syntax is what's being recommended by the Git core project and we expect to see more support around co-authors soon.

### DIFF
--- a/app/src/ui/lib/author-input.tsx
+++ b/app/src/ui/lib/author-input.tsx
@@ -628,6 +628,10 @@ export class AuthorInput extends React.Component<IAuthorInputProps, {}> {
       needle
     )
 
+    const exactMatch =
+      hits.length === 1 &&
+      hits[0].username.toLowerCase() === needle.toLowerCase()
+
     const existingUsernames = new Set(this.authors.map(x => x.username))
 
     const list: any[] = hits
@@ -641,7 +645,7 @@ export class AuthorInput extends React.Component<IAuthorInputProps, {}> {
         hint: this.applyCompletion,
       }))
 
-    if (needle.length > 0 && validLoginExpression.test(needle)) {
+    if (!exactMatch && needle.length > 0 && validLoginExpression.test(needle)) {
       list.push({
         text: `@${needle}`,
         username: needle,
--- a/app/src/ui/lib/author-input.tsx
+++ b/app/src/ui/lib/author-input.tsx
@@ -704,7 +704,7 @@ export class AuthorInput extends React.Component<IAuthorInputProps, {}> {
       hintOptions: {
         completeOnSingleClick: true,
         completeSingle: false,
-        closeOnUnfocus: false,
+        closeOnUnfocus: true,
         closeCharacters: /\s/,
         hint: this.onAutocompleteUser,
       },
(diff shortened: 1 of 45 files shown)


j-f1:
> Or, if the user you're adding has chosen to keep their email private
> ```
> Co-Authored-By: Foo Bar <foo@users.noreply.github.com>
> ```

Isn’t the new email format `userid+username@users.noreply.github.com`?

shiftkey:
@j-f1 [it depends](https://help.github.com/articles/about-commit-email-addresses/):

> Note: If you created your GitHub account **after July 18, 2017**, your GitHub-provided no-reply email address is a seven-digit ID number and your username in the form of ID+username@users.noreply.github.com. If you created your GitHub account prior to July 18, 2017, your GitHub-provided no-reply email address is your username in the form of username@users.noreply.github.com.

Not sure if we have enough context to figure this out easily - `created_at` through the `/users/{login}` API would work, but I think it's generating this URL manually currently.

niik:
> Isn’t the new email format userid+username@users.noreply.github.com?

It is, both still work but we intend to switch to the newer format. This feature had an extremely compressed schedule and a few things had to be cut, that was one of them. We currently don't have the user id readily available in the client. There are workarounds but we did't get to them in time. We also didn't have time to get any data on what Enterprise versions support the new format and the percentage of users on those versions.

> Not sure if we have enough context to figure this out easily - created_at through the /users/{login} API would work, but I think it's generating this URL manually currently.

I'm 99% sure it works regardless, it's just about what the default is on dotcom. But regardless, this is the stuff we didn't have time to get to

gamexxx:
@niik  Thank you very much for your work!

Geazy96:
Add design like

app/src/ui/lib/author-input.tsx
```diff
@@ -60,6 +61,7 @@ export class AuthorInput extends React.Component<
   private editor: CodeMirror.Editor | null = null
   private readonly resizeObserver: ResizeObserver
   private resizeDebounceId: number | null = null
+  private hintActive: boolean = false
 
   public constructor(props: IAuthorInputProps) {
```
grantjamie7777-collab: Being rich is all i want and need

grantjamie7777-collab: [9cbb6d23-55f7-4e31-91a5-af9bae2a157c.pdf](https://github.com/user-attachments/files/31231877/9cbb6d23-55f7-4e31-91a5-af9bae2a157c.pdf)
[membershipfee_agreement_v1.0.0 (1).pdf](https://github.com/user-attachments/files/31231878/membershipfee_agreement_v1.0.0.1.pdf)

PR #3884, opened 2018-01-30
Fix expanded avatar stack overflow

niik:
This fixes an issue that was discovered by @tierninho during the internal testing of #3879 where the avatar stack in the commit summary was cut off when you had an insane amount of users.

### Before

lotsa-authors-before

### After

lotsa-authors


tierninho:
@niik 🥇 looks good

nerdneha:
good-to-go

PR #3886, opened 2018-01-30
Edit context menus in text input elements

niik:
While adding a context menu for the co-authors toggle in #3879 I realized that we didn't have any context menus for our input controls. That's not very nice of us. I don't know about macOS but on Windows it's very much expected that there's cut, copy and paste

### Before

screen shot 2018-01-30 at 17 03 36

### After

screen shot 2018-01-30 at 16 34 20

screen shot 2018-01-30 at 16 34 10

This still needs to be tested on Windows. @tierninho I'll try to get to this tonight when I'm home by my PC but if I don't get to it could you please take a look at it for me?


tierninho:
@niik will take a look today

j-f1:
Can you remove the “Paste and match style” item somehow? The text boxes we use aren’t styled, so it doesn’t really make sense.

tierninho:
I'm getting this is OSX:

screen shot 2018-01-30 at 8 55 04 am

agisilaos:
That's super rad! Wanted such thing for such a long time. Thanks for working on it @niik 💯

app/src/ui/lib/author-input.tsx
```diff
@@ -743,18 +745,51 @@ export class AuthorInput extends React.Component<IAuthorInputProps, {}> {
...
+
+    if (__WIN32__) {
+      menu.push({ type: 'separator' })
+    }
+
+    menu.push({
+      label: __DARWIN__ ? 'Select All' : 'Select all',
+      action: () => {
+        // trust me, it's a thing :cry:
+        ;(cm as any).execCommand('selectAll')
```
j-f1: It looks like CodeMirror supports this:

```ts
        execCommand(name: string): void;
```

[source](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a187f6d4476403da23a2a77c3315d3bb573cc1b/types/codemirror/index.d.ts#L344)

niik: Lemme try updating our codemirror types

tierninho:
One small inconsistency found: If you highlight text in the co-auth field, then try to right-click to select any option, the highlighting of the text disappears. This works fine in the Summary and Description fields, but not the co-auth field.

niik:
> One small inconsistency found: If you highlight text in the co-auth field, then try to right-click to select any option, the highlighting of the text disappears. This works fine in the Summary and Description fields, but not the co-auth field.

Yeah, this is because it's not really a text field. I think we can to live with that inconsistency tho

app/src/main-process/main.ts
```diff
@@ -272,19 +273,45 @@ app.on('ready', () => {
...
-        menu.append(item)
+      for (const [ix, item] of items.entries()) {
+        // Special case editMenu in context menus. What we
+        // mean by this is that we want to insert all edit
+        // related menu items into the menu at this spot, we
+        // don't want a sub menu
+        if (item.role === 'editMenu') {
+          const editMenu = Menu.buildFromTemplate([item]).items[0]
+
+          if (editMenu.submenu) {
```
iAmWillShepherd: Let's move the `continue` towards the top to make it super up when to skip.
```ts
...
const editMenu = Menu.buildFromTemplate([item]).items[0]

if (!editMenu.submenu) {
  continue
}
...
```

app/src/main-process/main.ts
```diff
@@ -272,19 +273,45 @@ app.on('ready', () => {
...
+            for (const editItem of editMenu.submenu.items) {
```
iAmWillShepherd: Let's use `filter` here since that seems to be the intent.

```ts
// We don't use styled inputs anywhere at the moment
// so let's filter this for now and when/if we do we
// can make it configurable from the callee
editMenu.submenu.items
  .filter(editItem =>
    editItem.role &&
    editItem.role.toLocaleLowerCase() !== 'pasteandmatchstyle'
  )
  .forEach(editItem => menu.append(editItem))
```

niik:
🐨

niik:
🏷

app/src/lib/menu-item.ts
```diff
@@ -0,0 +1,40 @@
...
+  /** Is the menu item enabled? Defaults to true. */
+  readonly enabled?: boolean
+
+  /**
+   * The predefined behavior of the menu item.
+   *
+   * When specified the click property will be ignored
+   */
+  readonly role?:
+    | 'undo'
```
iAmWillShepherd: Can we extract this out into a separate type?

app/src/main-process/menu/build-context-menu.ts
```diff
@@ -0,0 +1,82 @@
+import { IMenuItem } from '../../lib/menu-item'
+import { Menu, MenuItem } from 'electron'
+
+/**
+ * Gets a value indicating whether or not two roles are considered
+ * equal using a case-insensitive comparison.
+ */
+function roleEquals(x: string | undefined, y: string | undefined) {
+  return (x ? x.toLowerCase() : x) === (y ? y.toLowerCase() : y)
```
iAmWillShepherd: If we use the separate type, we can use it here as well.

app/src/lib/menu-item.ts
niik: I spent some time thinking about this and I decided to just remove the type here. That list of roles wasn't even complete and I think we should wait until Electron adds types for its role property and then reuse that instead of trying to stay in sync with what Electron does

niik:
🍭

package.json
```diff
@@ -107,7 +107,7 @@
     "@types/chai": "^4.0.5",
     "@types/chai-datetime": "^0.0.31",
     "@types/classnames": "^2.2.2",
-    "@types/codemirror": "^0.0.49",
+    "@types/codemirror": "0.0.55",
```
iAmWillShepherd: Just curious if this PR needed anything from the updated types for codemirror.

niik: I couldn't remember but looking at the [commit which introduced that change](https://github.com/desktop/desktop/commit/5bc1c1c0c48ce989849745e06f82a9b2296b9f79) it adds the `execCommand` method, getting rid of an `any` cast.

iAmWillShepherd:
Need's a merge.

niik:
🐨

PR #3348, opened 2017-11-16
Remove the title attribute on the Windows window buttons

j-f1:
See #3338.

PR #3906, opened 2018-02-01
drop unnecessary comments before issue template

shiftkey:
This came up when I was pairing with @nerdneha today.

What's the first thing you see when you create a new issue? The mention of our code of conduct:

That's cool, but there's already a link to it in the header above.

If you remove that, you see our mention of a known issue:

While this is helpful to tell users about, it's very specific and takes up a lot of space.

**I believe both of these things are distracting from the actual purpose of the template - to guide users to fill out the template - and without that instruction they're just filling in the header.**

If we see an uptick in the minimize tooltip after this, I'd recommend including something like this in the header and linking out to a known issues doc to keep this minimized:

```
<!--

Please summarize the issue in the title, and then use the template below to
fill out the details so we can reproduce the issue on our end.

Please also check out our list of known issues that are currently tracked:

https://github.com/desktop/desktop/blob/master/docs/contributing/known_issues.md

-->
```

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,5 @@
 <!--
 
-Have you read GitHub Desktop's Code of Conduct? By filing an Issue, you are
-expected to comply with it, including treating everyone with respect:
-
-https://github.com/desktop/desktop/blob/master/CODE_OF_CONDUCT.md
-
--->
-
-<!--
-
-Are you encountering an issue where the “Minimize” tooltip stays visible
-when you click the minimize button in the window? If so, that is an issue
-with Electron, the framework the app uses. Please subscribe to the issue
-at this link for updates on the issue:
-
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,25 +1,5 @@
...
--->
-
-<!--
-
-Are you encountering an issue where the “Minimize” tooltip stays visible
-when you click the minimize button in the window? If so, that is an issue
-with Electron, the framework the app uses. Please subscribe to the issue
-at this link for updates on the issue:
-
-https://github.com/electron/electron/issues/9943
```
j-f1: I was actually thinking about removing this recently — we removed the tooltip altogether in #3348.

tierninho:
+ 1000000000

shiftkey:
> @shiftkey actually the first thing I noticed was your avatar is upside down...

PR #3907, opened 2018-02-01
Adjust exact match definition

### LINKED ISSUE #3887: Search for user appears in co-author list even though there's an exact mentionable match
@nerdneha spotted this while we were looking at another issue earlier.

screen shot 2018-01-30 at 10 48 26 am

The first two entries are from the mentionables API and the third one is for adding a handle that hasn't interacted with the user before. The last one should not show up when there's an exact match in the list already.

@nerdneha graciously volunteered to fix this 🚀

nerdneha:
Fixes #3887 

Sets exactMatch to true as long as one of the hits exactly matches the needle. 

Before we were setting to true only if there was 1 hit but ran into the counterexample below:

Before:
screen shot 2018-01-30 at 10 48 26 am

After:
screen shot 2018-01-31 at 9 52 38 pm

### DIFF
--- a/app/src/ui/lib/author-input.tsx
+++ b/app/src/ui/lib/author-input.tsx
@@ -629,9 +629,9 @@ export class AuthorInput extends React.Component<IAuthorInputProps, {}> {
       needle
     )
 
-    const exactMatch =
-      hits.length === 1 &&
-      hits[0].username.toLowerCase() === needle.toLowerCase()
+    const exactMatch = hits.some(hit => {
+      return hit.username.toLowerCase() === needle.toLowerCase()
+    })
 
     const existingUsernames = new Set(this.authors.map(x => x.username))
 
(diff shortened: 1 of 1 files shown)


app/src/ui/lib/author-input.tsx
```diff
@@ -629,9 +629,9 @@ export class AuthorInput extends React.Component<IAuthorInputProps, {}> {
       needle
     )
 
-    const exactMatch =
-      hits.length === 1 &&
-      hits[0].username.toLowerCase() === needle.toLowerCase()
+    const exactMatch = hits.some(hit => {
+      return hit.username.toLowerCase() === needle.toLowerCase()
+    })
```
j-f1: `hits.some(hit => hit.username.toLowerCase() === needle.toLowerCase())`?

nerdneha:
🍰

PR #3908, opened 2018-02-01
line endings are hard, lets go shopping

### LINKED ISSUE #3514: "Ignore" context-menu option incorrectly adds carriage return to .gitignore
### Description

When I select a file to ignore from the list of modified files, and then choose "Ignore" from the context menu, the filename is appended to `.gitignore` as desired—but incorrectly suffixed with a `\r` (`^M`).

### Version

**OS version:** Mac OS X El Capitan, 10.11.6 (15G17023)

### Steps to Reproduce

1. Modify a file in the current repository.
1. Select file in "Changes"; choose "Ignore" from context menu.
1. Open .gitignore with text editor.

**Expected behavior:** Filename of ignored file added to .gitignore, suffixed with `\n`.
**Actual behavior:** Filename of ignored file added to .gitignore, suffixed with `\r\n`.
**Reproduces how often:** Always?

### Logs

[2017-12-06.desktop.production.log](https://github.com/desktop/desktop/files/1535093/2017-12-06.desktop.production.log)

#### Additional Information

A suitable setting of `core.autocrlf` will ensure that the carriage return is not actually committed, but it should not have been added in the first place.

shiftkey:
Fixes #3514

### DIFF
--- a/app/src/lib/stores/repository-settings-store.ts
+++ b/app/src/lib/stores/repository-settings-store.ts
@@ -110,6 +110,12 @@ async function formatGitIgnoreContents(
       return
     }
 
+    // fallback to Git default behaviour
+    if (autocrlf == null) {
+      resolve(`${text}\n`)
+      return
+    }
+
     const linesEndInCRLF = autocrlf === 'true'
     if (linesEndInCRLF) {
       resolve(`${text}\n`)
--- a/app/src/lib/stores/repository-settings-store.ts
+++ b/app/src/lib/stores/repository-settings-store.ts
@@ -110,17 +110,16 @@ async function formatGitIgnoreContents(
       return
     }
 
-    // fallback to Git default behaviour
     if (autocrlf == null) {
-      resolve(`${text}\n`)
-      return
-    }
-
-    const linesEndInCRLF = autocrlf === 'true'
-    if (linesEndInCRLF) {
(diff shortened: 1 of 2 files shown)


app/src/lib/stores/repository-settings-store.ts
```diff
@@ -110,6 +110,12 @@ async function formatGitIgnoreContents(
       return
     }
 
+    // fallback to Git default behaviour
```
iAmWillShepherd: I think this should be a part of an else block since it's very much related to the previous condition.

app/test/unit/repository-settings-store-test.ts
```diff
@@ -82,4 +76,34 @@ describe('RepositorySettingsStore', () => {
...
+
+      // do not do any conversion of line endings when committing
+      await GitProcess.exec(
+        ['config', '--local', 'core.autocrlf', 'input'],
+        repo.path
+      )
+
+      const path = repo.path
+
+      await sut.saveGitIgnore('node_modules')
```
iAmWillShepherd: I don't think we need this line.

iAmWillShepherd: Nvm, looks like this is being used to easily add a file to the repo.

shiftkey:
:apple:

PR #3911, opened 2018-02-01
Switch to a saner default gravatar size

niik:
This was supposed to be part of the co-authors PR but this commit got left accidentally on my Windows machine.

The biggest avatar we display in the app right now is 28x28 so 60 should be a good default for retina monitors

### DIFF
--- a/app/src/lib/gravatar.ts
+++ b/app/src/lib/gravatar.ts
@@ -8,7 +8,7 @@ import { getDotComAPIEndpoint } from './api'
  * @param email The email address associated with a user
  * @param size The size (in pixels) of the avatar to render
  */
-export function generateGravatarUrl(email: string, size: number = 200): string {
+export function generateGravatarUrl(email: string, size: number = 60): string {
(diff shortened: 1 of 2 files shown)


app/src/lib/gravatar.ts
```diff
@@ -8,7 +8,7 @@ import { getDotComAPIEndpoint } from './api'
  * @param email The email address associated with a user
  * @param size The size (in pixels) of the avatar to render
  */
-export function generateGravatarUrl(email: string, size: number = 200): string {
+export function generateGravatarUrl(email: string, size: number = 60): string {
```
iAmWillShepherd: I'm a fan of defaults, but I was under the impression that it was determined not use them. I know this was already the case, but just curious if we can start using defaults in a function's parameters?

shiftkey: @iAmWillShepherd I'm happy to review this decision on a case-by-case basis. Feel free to call this out the next time you encounter it on a PR.

PR #3569, opened 2017-12-11
Run Prettier on All The Things

j-f1:
Fixes #2875

- [x] Extract options to `.prettierrc`
- ~❔ Lint the JS/TS in Markdown (`eslint-plugin-markdown`) (` ```ts ` blocks aren’t supported yet) (blocked on https://github.com/eslint/eslint-plugin-markdown/issues/82)~
- [x] Prettier the SCSS
[...]

PR #3915, opened 2018-02-01
Revert "Run Prettier on All The Things"

iAmWillShepherd:
Reverts desktop/desktop#3569


shiftkey:
Yep, I don't mind reviving the original PR to just focus on making the code consistent. But let's not interfere with people's flow.

shiftkey:
I'm going to revisit the original PR and see if we can get most of those changes in.

PR #3916, opened 2018-02-01
update prettier and apply it to SCSS files

shiftkey:
This PR is the core of #3569 but without enforcing conventions on Markdown (as discussed).


script/prettier.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (fileList.length > 0) {
+    console.log('These files are not formatted correctly:\n')
+
+    console.log(result.stdout.toString())
+
+    console.error(
+      chalk`{bold.green → To fix these errors, run {underline yarn lint:prettier --fix}}`
+    )
+  } else {
+    console.log('Something went wrong with invoking prettier:')
```
iAmWillShepherd: Can we use `console.error` for errors if only to make it clearer what our error conditions are?

shiftkey:
:apple:

kokphinchian1682:
Details about it and All First

Owner

PR #3922, opened 2018-02-03
 Add small space between commit summary and description

willnode:
I noticed that the commit list has been redesigned. I think it can be better a bit.

This commit brings down commit description by 3px while keeping commit summary in place.

Preview:

screenshot 28

Close up difference:

screenshot 27

### DIFF
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -5,7 +5,7 @@ import { CommitListItem } from './commit-list-item'
 import { List } from '../lib/list'
 import { IGitHubUser } from '../../lib/databases'
 
-const RowHeight = 48
+const RowHeight = 50
 
 interface ICommitListProps {
   readonly onCommitChanged: (commit: Commit) => void
--- a/app/styles/ui/history/_commit-list.scss
+++ b/app/styles/ui/history/_commit-list.scss
@@ -20,13 +20,14 @@
     .info {
       display: flex;
       flex-direction: column;
-      margin-top: -5px;
+      margin-top: -3px;
       overflow: hidden;
       width: 100%;
 
       .description {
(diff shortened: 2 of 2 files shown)


shiftkey:
@willnode thanks for the contribution, and thanks for the detailed visual comparison.

I'll let @desktop/design chime in about this when they have a moment.

donokuda:
Hi @willnode, thanks for the pull request! Can you explain the reason for bumping down the commit description?

One other thing to note, you'll also need to increase the row height of the list item here (probably by the amount you changed):
https://github.com/desktop/desktop/blob/d2d37b73b2dc5c4fd6096b7393dccacbdb4586ce/app/src/ui/history/commit-list.tsx#L8

This is because the height itself is fixed and the vertical padding becomes uneven as a result (the bottom padding as less space):

willnode:
@donokuda The original design is already looking good. But whenever I see a commit message have difference of only 1 pixel below the profile picture (e.g. emoji or `g` in commit summary), it always come to my mind that I want to bump it down a little bit.

I wasn't knew where to change the row height so thanks for pointing that. I will make it bigger by 2px, but I'm waiting for feedback.

Updated comparison (original, PR at this state, after resized by +2px)

githubdemo

donokuda:
@willnode Thanks for the clarification! 🙇 

I totally agree and this looks like a good improvement. When you push the changes I suggested above, feel free to request my review and I can jump in on it. (Bonus points for attaching a screenshot like how you've done earlier 📸 ✨)

willnode:
@donokuda All set!

And yeah, I done the whole thing with DevTool open. So I just have to edit the stylesheet and not having to clone and rebuild whole thing. The rest is just some magic with `mspaint` and `powerpnt` 🎨.

screenshot 54

PR #3927, opened 2018-02-04
add draft-release command to codify release process

shiftkey:
**[Rendered docs](https://github.com/desktop/desktop/blob/57d9a3e597a01b05385099d94966ba0da5210f5a/docs/process/releasing-updates.md#the-process)**

I was about to cut a new beta release, but wanted to simplify [the current process](https://github.com/desktop/desktop/blob/master/docs/process/releasing-updates.md#the-process) and catch issues that have caught me out before:

 - ensuring `app/package.json` is incremented
 - ensuring there's an entry for this new version

I also wanted to drive this more from the repository state, rather than guessing the process.

So here's what I'm suggesting:

```
$ GITHUB_ACCESS_TOKEN={token} yarn draft-release beta
yarn run v1.3.2
$ ts-node script/draft-release/index.ts beta
Here's what you should do next:

1. Ensure the app/package.json 'version' is set to '1.0.14-beta2'
2. Add this to changelog.json as a starting point:
{
  "1.0.14-beta2": [
    "[???] Add RubyMine support for macOS - #3883. Thanks @gssbzn!",
    "[???] Allow window to accept single click on focus - #3843",
    "[???] Drop unnecessary comments before issue template - #3906",
    "[???] First-class changelog script for generating release notes - #3888",
    "[???] Fix expanded avatar stack overflow - #3884",
    "[???] Revert \"Run Prettier on All The Things\" - #3915",
    "[???] Space - #1. Thanks @shiftkey!",
    "[???] Switch to a saner default gravatar size - #3911",
    "[Fixed] Add a repository settings store - #934",
    "[Fixed] Ensure renames are detected when viewing commit diffs - #3673",
    "[Fixed] Line endings are hard, lets go shopping - #3514",
    "[Fixed] Run Prettier on All The Things - #2875. Thanks @j-f1!"
  ]
}

3. Update the release notes so they make sense and only contain user-facing changes
4. Commit the changes and push them to GitHub
5. Read this to perform the release: https://github.com/desktop/desktop/blob/master/docs/process/releasing-updates.md
✨  Done in 10.36s.
```

Interesting bits:

 - the versioning scheme is defined in `getNextVersionNumber` and I added a bunch of tests in `app/test/unit/script/version-test.ts` to describe how it would flow from beta to stable and stable to beta.
 - I had to extract out `getChangelogEntries` so I could reuse them here. I don't want to touch those for now.

To Do:

 - [x] test drafting a new production release - should pull in all beta release notes (that aren't `-beta1` because we use that to immediately push a new build after `production` is updated)
 - [x] update docs to refer to this New Way™


chavellacinema160:
still confused but im getting it,thank you for being patient with me ,people are counting on me out here

docs/process/releasing-updates.md
```diff
@@ -12,12 +12,77 @@ We have three channels to which we can release: `production`, `beta`, and `test`
...
-1. Bump `version` in [`app/package.json`](../../app/package.json) to `$version`.
-1. Commit & push the changes.
-1. Run `.release! desktop/YOUR_BRANCH to {production|beta|test}`.
-  * We're using `.release` with a bang so that we don't have to wait for any current CI on the branch to finish. This might feel a little wrong, but it's OK since making the release itself will also run CI.
-1. If you're releasing a production update, release a beta update for the next version too, so that beta users are on the latest release.
+From a clean working directory, set the `GITHUB_ACCESS_TOKEN` environment variable to a valid [Personal Access Token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) and run:
+
+```shellsession
+$ yarn draft-release (production|beta)
+```
```
j-f1: For a one-liner, `GITHUB_ACCESS_TOKEN=$(cat) yarn draft-release (production|beta)`, hit enter, paste your token, hit enter, then hit `^D`.

```shellsession
$ FOO=$(cat) bash
foo-bar
^D
> echo $FOO
foo-bar
```

script/draft-release/run.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  throw new Error(`An invalid channel ${arg} has been provided`)
+}
+
+function printInstructions(nextVersion: string, entries: Array<string>) {
+  console.log(
+    `1. Ensure the app/package.json 'version' is set to '${nextVersion}'`
+  )
+  console.log('2. Add this to changelog.json as a starting point:')
```
j-f1: I wonder if it’s possible to do these automatically.

script/draft-release/run.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export async function run(args: ReadonlyArray<string>): Promise<void> {
+  //try {
+  //  await spawn('git', ['diff-index', '--quiet', 'HEAD'])
+  //} catch {
+  //  throw new Error(
+  //    `There are uncommitted changes in the working directory. Aborting...`
+  //  )
+  //}
```
j-f1: 🔥 `//` commented code

script/draft-release/version.ts
```diff
@@ -0,0 +1,55 @@
+import { inc, parse } from 'semver'
+
+export function getNextVersionNumber(
+  version: string,
+  channel: 'production' | 'beta'
```
j-f1: Maybe add a `Channel` typedef to make it easier to add new ones in the future?

script/draft-release/run.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  return latestTag
+}
+
+function parseChannel(arg: string): 'production' | 'beta' {
+  if (arg === 'production' || arg === 'beta') {
+    return arg
+  }
+
+  throw new Error(`An invalid channel ${arg} has been provided`)
```
j-f1: You might be able to `assertNever` here.

docs/process/releasing-updates.md
shiftkey: I left this here because it's totally going to fail in any reasonable shell, but I'm open to a simpler way to demonstrate the options available

script/draft-release/run.ts
shiftkey: I have thoughts on this, but let's agree on the process before we make it more magical

script/draft-release/run.ts
shiftkey: `assertNever` is an application API. I'd rather just have this fail fast than couple this to app code at this point.

docs/process/releasing-updates.md
j-f1:
```shellsession
$ yarn draft-release production
or:
$ yarn draft-release beta
```

shiftkey: Actually, to come back to the earlier point:

> For a one-liner, `GITHUB_ACCESS_TOKEN=$(cat) yarn draft-release (production|beta)`, hit enter, ...

Let's not make this shell-specific. I call out the environment variable in the previous paragraph, and that's applicable to all OSes.

But I'll tweak this sample to show these `production` and `beta` inputs as separate statements...

script/draft-release/version.ts
```diff
@@ -0,0 +1,57 @@
+import { inc, parse } from 'semver'
+
+import { Channel } from './channel'
+
+export function getNextVersionNumber(
+  version: string,
+  channel: Channel
+): string {
+  const parsed = parse(version)
```
iAmWillShepherd: Could we improve naming here? Something like `semanticVersion`?

script/draft-release/version.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (parsed == null) {
+    throw new Error(`Unable to parse input '${version}' into version`)
+  }
+
+  if (channel === 'production') {
```
iAmWillShepherd: Can we be explicit with `channel` and check for all possible values exhaustively. Conditionals work, but I think a switch makes it that much clearer.

```ts
  switch (channel) {
    case 'production': break
    case 'beta': break
    case 'test': break
    case 'development': break
  }
```

script/draft-release/channel.ts
```diff
@@ -0,0 +1 @@
+export type Channel = 'production' | 'beta'
```
iAmWillShepherd: Should we add `test` and `development` as well since `app.tsx` uses the `__RELEASE_CHANNEL__ constant?

https://github.com/desktop/desktop/blob/484856d67a7975aa4d574b7966355a7b873fff92/app/src/lib/globals.d.ts#L27-L31

script/draft-release/version.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (parsed.prerelease.some(p => p.startsWith('beta'))) {
```
iAmWillShepherd: Could we extract the logic from this condition body into a function, `tryIncremenetProductionVersion: (version: SemVer) => string` (or similar name)?

script/draft-release/version.ts
```diff
@@ -0,0 +1,57 @@
...
+    return nextVersion
+  }
+
+  if (parsed.prerelease.some(p => p.startsWith('test'))) {
+    throw new Error(
+      `Unable to resolve production version using test release '${version}'`
+    )
+  }
+
+  const betaTagIndex = version.indexOf('-beta')
```
iAmWillShepherd: This can be extracted into a function, `tryIncrementBetaVersion: (version: string) => string` (or similar name)?

script/draft-release/run.ts
```diff
@@ -0,0 +1,108 @@
...
+  convertToChangelogFormat,
+  getChangelogEntriesSince,
+} from '../changelog/parser'
+
+import { Channel } from './channel'
+import { getNextVersionNumber } from './version'
+
+const jsonStringify: (obj: any) => string = require('json-pretty')
+
+async function getLatestRelease(excludeBetaReleases: boolean): Promise<string> {
```
iAmWillShepherd: Functions that accept a boolean to conditionally do things are sometimes hard to read. Can we extract this into an enum?

```ts
export enum ReleaseOption {
  IncludeBeta,
  ExcludeBeta
}
```

That way, the function call reads better and doesn't require one to look at its definition to figure out what the bool does. ex. `getLatestRelease(ReleaseOption.IncludeBeta)`

script/draft-release/run.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (excludeBetaReleases) {
+    releaseTags = releaseTags.filter(tag => tag.indexOf('-beta') === -1)
+  }
+
+  const releaseVersions = releaseTags.map(tag => tag.substr(8))
+
+  const sortedTags = semverSort(releaseVersions)
+  const latestTag = sortedTags[sortedTags.length - 1]
+
+  if (latestTag instanceof SemVer) {
```
iAmWillShepherd: This can be converted to the following ternary:
```ts
return latestTag instanceof SemVer ? latestTag.raw : latestTag`
```

script/draft-release/run.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  return latestTag
+}
+
+function parseChannel(arg: string): Channel {
+  if (arg === 'production' || arg === 'beta') {
+    return arg
+  }
+
+  throw new Error(`An invalid channel ${arg} has been provided`)
```
iAmWillShepherd: We have 2 other channels defined in the app. Should those go here?

script/draft-release/run.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function printInstructions(nextVersion: string, entries: Array<string>) {
```
iAmWillShepherd: This function can be made a bit clearer:

```ts
function printInstructions(nextVersion: string, entries: Array<string>) {
  const obj: any = {}
  obj[`${nextVersion}`] = entries.sort()

  const steps = [
    `Ensure that the value of the 'version' property in app/package.json is '${nextVersion}'`,
    `Add the following entry to changelog.json as a starting point`,
    `${jsonStringify(obj)}`,
    'Modify the release notes so that they make sense and contain only user-facing changes',
    `Commit the changes and push them to GitHub'`,
    `Read this to perform the release: https://github.com/desktop/desktop/blob/master/docs/process/releasing-updates.md`,
  ]
    .map((value, index) => `${index + 1}. ${value}`)
    .join('\n')

  console.log(steps)
}
```

script/changelog/parser.ts
```diff
@@ -0,0 +1,121 @@
+import * as Path from 'path'
+import * as Fs from 'fs'
+import { gt } from 'semver'
```
iAmWillShepherd: Can we alias this? 
`import { gt as greaterThan } from 'semver'`

script/changelog/parser.ts
```diff
@@ -0,0 +1,121 @@
...
+  return str.charAt(0).toUpperCase() + str.slice(1)
+}
+
+function getChangelogEntry(commit: IParsedCommit, pr: IAPIPR): string {
+  let issueRef = ''
+  let type = PlaceholderChangeType
+  const description = capitalized(pr.title)
+
+  const re = /Fixes #(\d+)/gi
+  let match
```
iAmWillShepherd: Type annotation hear.

script/changelog/parser.ts
```diff
@@ -0,0 +1,121 @@
+import * as Path from 'path'
+import * as Fs from 'fs'
+import { gt } from 'semver'
+
+import { fetchPR, IAPIPR } from './api'
+
+const PlaceholderChangeType = '???'
+const OfficialOwner = 'desktop'
+
+interface IParsedCommit {
```
iAmWillShepherd: Will this need to be updated to handle co-authors

script/draft-release/run.ts
j-f1: or `getLatestRelease({ excludeBetaReleases: false })`

script/changelog/parser.ts
shiftkey: Let's save this discussion for #3949 or later -  this was intended to be extracted without changing the behaviour

script/draft-release/channel.ts
shiftkey: We should only focus on the channels that we actually publish too, so I wouldn't worry about `development` here. I'll add `test` to this list but I won't be implementing it as part of this PR.

script/draft-release/run.ts
shiftkey: While this is a bit terse, it will render `${jsonStringify(obj)}` with it's own number, which we want to avoid.

I'll have a go at tidying this up a bit anyway so it's easier to follow.

shiftkey:
:apple:

script/draft-release/run.ts
shiftkey: 13bd78c5c

iAmWillShepherd: This may seem silly/hacky, but this could work...
`${jsonStringify(obj)` -> `#${jsonStringify(obj)}`
Then replace anything that matches the regex `/\d.\s#/` with the `empty string`

PR #3600, opened 2017-12-13
improved detection of remote ref when deleting branch

JordanMussi:
- Related issue: #2964
- Original PR: #3469

Instead of passing `existOnRemote` through the chain, simply use the `getBranchAheadBehind` function in the dialog.

This is a split from https://github.com/desktop/desktop/pull/3559#discussion_r156792354

PR #3559, opened 2017-12-09
add context menu to branch list

JordanMussi:
Fixes #3202 

In order to do this I improved the fix for #2964 (original fix in #3469).

PR #3469, opened 2017-11-30
Fixes #2964 'Delete Branch' warns that branch already exists on remote after it's deleted

JordanMussi:
Fixes #2964

Issue #3202, opened 2017-10-31
Delete branch without switching to it


adamreisnz:
Pardon me if it's already been asked, but I couldn't locate a similar issue.

One of the things that always bothered me in the old app was that I had to switch to a certain branch before I could delete it. This still appears to be the case in the new app.

Can we get a context menu for right click on a branch in the branches list, and have some options in there, like delete branch?

This way, when I am done with a branch and I have switched back to another and merged it in, I don't need to switch back again, delete the branch, and then switch a third time.

There shouldn't be a need to switch to a particular branch before deleting it as you can also delete another branch via the command line without being on it.

niik:
The old clients used to automatically prune merged local branches (see #750). I'm hoping that we'll be able to get to that at some point which would make deleting local branches manually less of an issue.

That said I don't have anything against also supporting a context menu in the branch list. For now though I'm gonna label this future-proposal and close it out so we can keep our backlog manageable. We'll revisit this in our next planning session.

adamreisnz:
Ok, thanks 👍

adamreisnz:
PS: If you aren't already, may I suggest using [Zenhub](https://www.zenhub.com/) to manage your issues backlog? 

It integrates on top of Github in your browser, and allows you to create as many columns/pipelines as you wish, so you could have one for example "Future proposal" or "Icebox" and put issues that are to be revisited later in there, rather than closing them off. That should still help to keep your actual issues backlog under control.

It's similar to Github's project tab, but a lot better with more functionality.

danqing:
Much needed...

Every time I right click and nothing shows up a kitten dies..

billygriffin:
Hi @adamreisnz, thanks for this. We're working through all the future proposals and it looks like this one should land when https://github.com/desktop/desktop/issues/5803 does, so I'm going to remove the label on this. Just didn't want to you come back to it and wonder what happened.

adamreisnz:
Cheers thanks

PR #3940, opened 2018-02-06
move remote detection into dialog component when deleting branch

shiftkey:
Supersedes #3600 

This takes the original work from @JordanMussi but moves the check closer to the decision time for the user. This avoids the overhead of computing whether the remote exists for every ref, but lets us be more accurate than what we previously had (by querying the remote).

### DIFF
--- a/app/src/ui/delete-branch/delete-branch-dialog.tsx
+++ b/app/src/ui/delete-branch/delete-branch-dialog.tsx
@@ -54,9 +54,6 @@ export class DeleteBranch extends React.Component<
       this.checkingRemote = true
 
       checkBranchExistsOnRemote(this.props.repository, this.props.branch)
-        .then(existsOnRemote => {
-          this.setState({ existsOnRemote }, this.checkCompleted)
-        })
         .catch(err => {
           log.warn(
             `[DeleteBranch] unable to resolve upstream branch: '${
@@ -64,7 +61,11 @@ export class DeleteBranch extends React.Component<
             }'`,
             err
           )
-          this.setState({ existsOnRemote: false }, this.checkCompleted)
+
+          return false
+        })
+        .then(existsOnRemote => {
+          this.setState({ existsOnRemote }, this.checkCompleted)
         })
     }
   }
--- a/app/src/ui/delete-branch/delete-branch-dialog.tsx
+++ b/app/src/ui/delete-branch/delete-branch-dialog.tsx
@@ -37,6 +37,10 @@ export class DeleteBranch extends React.Component<
     }
   }
 
+  private checkCompleted = () => {
+    this.checkingRemote = false
+  }
+
   public componentWillReceiveProps(nextProps: IDeleteBranchProps) {
     if (this.checkingRemote) {
       // a check is being performed, don't start another
@@ -51,9 +55,7 @@ export class DeleteBranch extends React.Component<
 
       checkBranchExistsOnRemote(this.props.repository, this.props.branch)
         .then(existsOnRemote => {
-          this.setState({ existsOnRemote }, () => {
-            this.checkingRemote = false
-          })
+          this.setState({ existsOnRemote }, this.checkCompleted)
         })
         .catch(err => {
           log.warn(
@@ -62,9 +64,7 @@ export class DeleteBranch extends React.Component<
             }'`,
             err
           )
-          this.setState({ existsOnRemote: false }, () => {
-            this.checkingRemote = false
-          })
+          this.setState({ existsOnRemote: false }, this.checkCompleted)
         })
     }
   }
--- a/app/src/ui/delete-branch/delete-branch-dialog.tsx
+++ b/app/src/ui/delete-branch/delete-branch-dialog.tsx
@@ -55,6 +55,9 @@ export class DeleteBranch extends React.Component<
 
       checkBranchExistsOnRemote(this.props.repository, this.props.branch)
         .catch(err => {
+          // if we encounter an unhandled error from Git, log an error
+          // and provide 'false' as the fallback value for the callback
+
           log.error(
             `[DeleteBranch] unable to resolve upstream branch: '${
               this.props.branch.upstream
--- a/app/src/ui/delete-branch/delete-branch-dialog.tsx
+++ b/app/src/ui/delete-branch/delete-branch-dialog.tsx
@@ -97,7 +97,7 @@ export class DeleteBranch extends React.Component<
   }
 
   private renderDeleteOnRemote() {
-    if (this.props.branch.remote && this.state.existsOnRemote == true) {
+    if (this.props.branch.remote && this.state.existsOnRemote === true) {
       return (
         <div>
           <p>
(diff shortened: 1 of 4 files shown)


app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -8,31 +8,68 @@ import { ButtonGroup } from '../lib/button-group'
...
+    }
+
+    const upstreamChanged =
+      this.props.branch.upstream !== nextProps.branch.upstream
+
+    if (this.state.existsOnRemote == null || upstreamChanged) {
+      this.checkingRemote = true
+
+      checkBranchExistsOnRemote(this.props.repository, this.props.branch)
+        .catch(err => {
```
iAmWillShepherd: Can we put the `then` callback before the `catch`?

app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -8,31 +8,68 @@ import { ButtonGroup } from '../lib/button-group'
...
+          log.info(
```
iAmWillShepherd: We should use `log.error` here so errors are filter properly in dev tools.

app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -61,7 +98,7 @@ export class DeleteBranch extends React.Component<
   }
 
   private renderDeleteOnRemote() {
-    if (this.props.branch.remote && this.props.existsOnRemote) {
+    if (this.props.branch.remote && this.state.existsOnRemote === true) {
```
iAmWillShepherd: No need for the `... === true`, `this.state.existsOnRemote` should do just fine.

app/src/ui/delete-branch/delete-branch-dialog.tsx
j-f1: The `return false` line calls the `.then` handler with `false`, which lets the component pretend the branch doesn’t exist on the remote.

shiftkey: @iAmWillShepherd yeah, that's what I had earlier but it resulted in duplicating the `setState` code, like this:

```ts
promise.then(existsOnRemote => {
   this.setState({ existsOnRemote }, this.checkCompleted)
}).catch(error => {
  log.error("oops", error)
  this.setState({ existsOnRemote }, this.checkCompleted)
})
```

By putting the `catch` first we can propagate a default value if the promise errors.

shiftkey:
:apple:

app/src/ui/delete-branch/delete-branch-dialog.tsx
iAmWillShepherd: Could you add that as a comment?

shiftkey: 9708fc8ddc02db4d8794b4fd87542384b05f5ddd

shiftkey:
:apple:

app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -8,31 +8,71 @@ import { ButtonGroup } from '../lib/button-group'
...
       includeRemoteBranch: false,
     }
   }
 
+  private checkCompleted = () => {
+    this.checkingRemote = false
+  }
+
+  public componentWillReceiveProps(nextProps: IDeleteBranchProps) {
+    if (this.checkingRemote) {
```
niik: Triggering the check on `componentWillReceiveProps` means we rely on getting multiple prop updates during the lifetime of the dialog since `componentWillReceiveProps` isn't called for the initial mount. I'm sure we're getting more than one update right now but we really don't need to given that the props we pass in are stable. We should apply the `receiveProps` pattern here instead, see https://github.com/desktop/desktop/blob/458a427b012fe4c6465852905b745fe921221c69/app/src/ui/changes/commit-message.tsx#L120-L122

shiftkey:
> Could you go into a bit more detail about how we are being more accurate?

I thought the original PR was trying to address the period in time where you could fetch, the remote branch was deleted, and then you try and delete it locally. It's a narrow window (a couple of minutes at worst?), but I figured #3600 would be more useful if we moved it closer to the decision time, which leads into...

> We're not querying the remote, are we? We're just checking that the local ref still exists?

Yeah, and I think I glossed over this detail in #3600 because the code lived in a different place and I assumed was expensive (and it was, but for other reasons).

So here's what it takes to actually verify this using `ls-remote`:

```
$ time git ls-remote origin master
458a427b012fe4c6465852905b745fe921221c69	refs/heads/master
git ls-remote origin master  0.04s user 0.01s system 2% cpu 2.229 total
```

4 seconds, whether it exists or not. I'm not a fan.

> I thought the aheadBehind state got updated every time we refresh the repository status which should happen every time you focus the app and after each fetch.

I remember poking around in this a while ago (while touching the dialog at some point) and not trusting what I saw, but maybe this was because I was caught in the window between fetches while it was removed from under me. I'll see if I can break this and report back and make a better case for this change.

cc @JordanMussi for anything else to add, based on your experience with #3600

JordanMussi:
I initially wrote #3469 because the delete branch dialogue didn't check if the remote existed. The dialogue assumed that if the local branch had an upstream, the branch existed on the remote. My change was simply to use the current state (the branch being deleted is currently checked out) to get the `aheadBehind` of the branch (which was of type `IAheadBehind` if exists on remote, else `null`).

This worked for deleting the branch when it is checked out. It became an issue when trying to solve #3202 since we needed to know if the branch exists on the remote without the branch being checked out. My attempt in #3559 used [`getBranchAheadBehind`](https://github.com/desktop/desktop/blob/a29e2d4bdf85d7b13f758b1eb9ed94c6da662010/app/src/lib/git/rev-list.ts#L50-L69). I was then [asked to split the dialogue fix from the addition of the context menu](https://github.com/desktop/desktop/pull/3559#discussion_r156792354) thus #3600 was born.

In #3600 @joshaber [pointed out](https://github.com/desktop/desktop/pull/3600#discussion_r156817680) I wasn't following the architecture for the app by calling `getBranchAheadBehind` in the delete branch dialogue component. So I refactored the whole thing so that the existsOnRemote state was stored in the branch objects (albeit in a poor way). I haven't done any programming in the Flux style (except for the PRs I've made on this repo) so this way seemed much more painful to me because whether the branch exists on the remote didn't feel like a local state that needed to be stored. To me it feels better to determine when launching the dialogue because that is the moment you care if the branch is on the remote (not if it existed on the last fetch).

shiftkey:
I ran through the steps again, trying to break this by merging an PR then getting Desktop to remove the remote branch again:

 - if you haven't fetched, the dialog asks if you want to delete the remote - [this is checked](https://github.com/desktop/desktop/blob/master/app/src/lib/git/branch.ts#L72-L77) and we only try to delete if `ls-remote` finds the branch. That seems fine.

 - if you fetch before deleting the branch, it does lose the tracking information and shows "Publish" rather than "Fetch" - that seems fine, and there's no prompt to remove the remote branch when you delete the branch.

I think that's good enough here, but if someone is able to break this please open a fresh issue with details about the repro steps.

Issue #3263, opened 2017-11-08
The user name or passphrase you entered is not correct.


SilverFoxA:
### Description

Trying to connect to GitHub Desktop on Mac OS High Seria but fails to connect and provides a pop-up screen with the following message "The username or passphrase you entered is not correct".

### Version

**GitHub Desktop version:** 1.0.8

**OS version:** Mac OS 10.13

### Steps to Reproduce

1. Download GitHub desktop from the official website
1. Click on Clone repo or preferences to login to GitHub.com
1. Enter credentials and login

**Expected behavior:** Login to Github.com and fetch the repositories assigned to the user's account

**Actual behavior:** Login process is not complete. Give error "The user name or passphrase you entered is not correct."

**Reproduces how often:** Always (100%)

### Logs

2017-11-08T06:07:19.570Z - info: [ui] [Dispatcher] requesting authenticated user
2017-11-08T06:07:21.592Z - info: [ui] [SignInStore] account resolved
2017-11-08T06:07:21.592Z - info: [ui] [AppStore] adding account SilverFoxA (Abhijit Das) to store
2017-11-08T06:07:21.631Z - error: [ui] Error adding account 'SilverFoxA'
Error: The user name or passphrase you entered is not correct.

shiftkey:
@SilverFoxA a couple of questions to help reproduce the issue:

> Enter credentials and login

Can you share a screenshot of the dialog you used to enter your credentials? I want to confirm it's the GitHub flow and not the generic account flow.

Also, does your account have 2FA enabled?

shiftkey:
@SilverFoxA actually, it looks like this error is really helpful:

```
2017-11-08T06:07:21.631Z - error: [ui] Error adding account 'SilverFoxA'
Error: The user name or passphrase you entered is not correct.
```

The source of this error is this function, which is interoperating with the keychain:

https://github.com/desktop/desktop/blob/7412eda2bb171f1eabd3a25f4c5a959e9592eaea/app/src/lib/stores/accounts-store.ts#L96-L106

Could something in High Sierra be restricting access to the keychain?

w1w1w2w2:
[3331111111111111111111111111.txt](https://github.com/desktop/desktop/files/1460999/3331111111111111111111111111.txt)

yassermog:
I have the same problem

BaldwinLi:
I have same issue after upgraded to macOS Hight Sierra 10.13.1
[...]

Issue #2107, opened 2017-06-24
Window hiding after using a second screen


Raul6469:
### Description

I use a second screen with my laptop, in which I display the GitHub desktop window. The problem is when I disconnect the second screen, the window seems to be again on that second screen, as we can guess with the Windows animations.

Here is the animation : 

![ezgif-1-4bfa617f98](https://user-images.githubusercontent.com/24607388/27507662-df629b12-58d3-11e7-8809-2cf4c57e8601.gif)

### Version

**GitHub Desktop version:** 0.6.2, but the problem was also here in the previous version, 0.6.1

**OS version:** Windows 10.0.15063

### Steps to Reproduce

1. Connect a second screen to your laptop
3. Move GitHub desktop window to that second screen
4. Disconnect the second screen

**Expected behaviour:** The window should be displayed in your laptop screen

**Actual behaviour:** The window seems to be in that inexistant second screen

**Reproduces how often:** Always

Hope this might be useful, this is my first contribution !

Raul6469:
Any thoughts on this ? I'll just add that the issue is independent of screens orientation (not only vertically as I initally wrote).

shiftkey:
@Raul6469 I can't reproduce this at the moment because I'm travelling (and they look at you funny when you try and take an external monitor on a plane), but maybe @iAmWillShepherd is able to reproduce it on his Windows machine.

My gut feeling is that we should be able to get [events](https://github.com/electron/electron/blob/master/docs/api/screen.md) when the screens are added/removed, and maybe we can move the app to a better location.

Raul6469:
Thanks for the answer @shiftkey ! I was just about to close this issue because I found a solution to my problem [here](https://superuser.com/a/53590) with some keyboard tricks. It seems that it can occur for other windows programs as well, so I don't really think that the problem is with desktop...

iAmWillShepherd:
@shiftkey, unfortunately, I can't reproduce because I'm working on my laptop. Haven't actually unpacked everything since moving -_-.

lownignitus:
I just had to deal with this same issue myself. I had GitHub installed on my mass storage HDD. I had to reinstall windows because it was simpler when I got an M.2 drive to use as my main boot drive. Launching from the install save on the HDD brought it up in such a way that it did not display on any screen. I tried the tried and true Shift+Right click to bring up system menu for icon to select move. To my surprise and horror this did bumpkiss... 

[...]

PR #3943, opened 2018-02-06
Add a document with known-issues

### LINKED ISSUE #2972: Folder Redirection Windows
### Description

I'm using a Win7 machine at work and it is configured with folder redirection.

https://technet.microsoft.com/en-us/library/cc732275(v=ws.11).aspx
https://en.wikipedia.org/wiki/Folder_redirection

This configuration forces Github Desktop to install at `\\NETWORK\Redirected\USERNAME\Application Data\GitHub Desktop`. The application opens, but much of the functionality is broken because it has trouble finding and using the data in the folder.

I had the same issue with Atom and I created this issue https://github.com/atom/atom/issues/15663. We were able to workaround by creating an 'electronUserData' folder at `C:\Users\USERNAME\.atom`.

Is it possible to do something similar with Github Desktop to restore functionality?

Thanks for your help.

### Version

**GitHub Desktop version:** [1.0.3]

**OS version:** [Microsoft Windows 6.1.7601]

### Logs
[2017-09-21.desktop.production.log](https://github.com/desktop/desktop/files/1360553/2017-09-21.desktop.production.log)
[2017-09-22.desktop.production.log](https://github.com/desktop/desktop/files/1360554/2017-09-22.desktop.production.log)
[2017-10-05.desktop.production.log](https://github.com/desktop/desktop/files/1360558/2017-10-05.desktop.production.log)

### LINKED ISSUE #4039: Consider using installed Git binaries instead of bundled git/msys-2.0.dll
### Description

Currently GitHub desktop uses bundled git mingw binaries (especially `msys-2.0.dll`). It causes some issues if there is a different version of  `msys-2.0.dll` on the system already installed (likely the case if user has installed Git for Windows).  This causes issues like #3096 . Since multiple versions of the said dll causes  cygheap problems, Avoid using bundled git/mingw when a Git installation is found on the system. 

### Version

**OS version:** Microsoft Windows [Version 10.0.16299.248]

### Steps to Reproduce

1. Install 32 bit Git
2. Install 64 Bit GitHub Desktop app
3. Clone a repository /commit to a repository & you will get cygheap errors.

The issue is not specific to the 32 bit / 64 bit Github desktop setup case, it happens whenever there are multiple versions of msys-2.0.dll.

**Expected behavior:** Use installed git binaries /dlls instead of bundled git whenever possible.

**Actual behavior:** the app currently used bundled binaries

**Reproduces how often:** Always with different version of `msys-2.0.dll`. If all the dlls are of same version and arch, then everything works fine.

nerdneha:
**[Rendered Markdown](https://github.com/desktop/desktop/blob/known-issues-doc/docs/known-issues.md)**

Purpose: a way for us to track "acknowledged" issues without clogging the issue tracker. 
How to use: If you want to close a known issue, add it to this doc with some context

These issues may or may not have workarounds. 

- [x] have README point to it
- [x] add some context (part 2 of doc) around what to do if....

cc @steveward, re: starting to document known issues and having a place to point people to

Closes #2972 
Closes #4039

### DIFF
--- a/docs/known-issues.md
+++ b/docs/known-issues.md
@@ -1,19 +1,65 @@
 # Known Issues
 
-The purpose of this is to document known issues that may have workarounds!
+This document outlines acknowledged issues with GitHub Desktop, including workarounds if known.
 
-## Mac
+## What should I do if...
+
+### I have encountered an issue listed here?
+
+Some known issues have a workaround that users have reported addresses the issue. Please try the workaround for yourself to confirm it addresses the issue.
+
+### I have additional questions about an issue listed here?
+
+Each known issue links off to an existing GitHub issue. If you have additional questions or feedback, please comment on the issue.
+
+### My issue is not listed here? 
+
+Please check the [open](https://github.com/desktop/desktop/labels/bug) and [closed](https://github.com/desktop/desktop/issues?q=is%3Aclosed+label%3Abug) bugs in the issue tracker for the details of your bug. If you can't find it, or if you're not sure, open a [new issue](https://github.com/desktop/desktop/issues/new).
+
+## macOS
+
+### 'The username or passphrase you entered is not correct' error after signing into account - [#3263](https://github.com/desktop/desktop/issues/3263)
+
+This seems to be caused by the Keychain being in an invalid state, affecting applications that try to use the keychain to store or retrieve credentials. Seems to be specific to macOS High Sierra (10.13).
+
+**Workaround:** 
+
+- Open `Keychain Access.app`
+- Right-click on the `login` keychain and try locking it
+- Right-click on the `login` keychain and try unlocking it
+- Sign into your GitHub account again
+
 ## Windows
-[Using GitHub Desktop with a repository that uses Folder Redirection](https://github.com/desktop/desktop/issues/2972)
- * cannot fix because it is unsupported by git
- * no current workaround
+ 
+### Window is hidden after detaching secondary monitor - [#2107](https://github.com/desktop/desktop/issues/2107)
 
-## Linux
+This is related to Desktop tracking the window position between launches, but not changes to your display configuration such as removing the secondary monitor where Desktop was positioned.
 
-# What to do if...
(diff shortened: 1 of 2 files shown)


shiftkey:
@nerdneha just took a shot at structuring the doc and answering the second item on your checklist - feel free to adjust or adapt it to suit

docs/known-issues.md
```diff
@@ -0,0 +1,65 @@
...
+**Workaround:** 
+
+ - Remove `%APPDATA%\GitHub Desktop\window-state.json` 
+ - Restart Desktop
+
+### Using a repository configured with Folder Redirection - [#2972](https://github.com/desktop/desktop/issues/2972)
+
+[Folder Redirection](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc753996(v%3dws.11)) is an feature of Windows for administrators to ensure files and folders are managed on a network server, instea.
+
+**Not supported** as Git is not able to resolve the working directory correctly:
```
nerdneha: not supported initially made me still feel like we might have a workaround. it's clear to me that we don't have one from the desktop side but maybe we can use a different phrase for this to be clearer?

some options
* no workaround
* unable to fix
* fix dependent on _____ (ex: git)

nerdneha:
fyi i added a "close" for the two added issues in this PR description; feel free to edit+take out if we want to keep any of the 3 issues open @shiftkey

shiftkey:
@nerdneha I dropped the #3263 mention because it's already closed, and I think we can do better than the workaround in #2107 and I have access to a spare monitor at the moment so I'm going to see if I can address that separate to this.

shiftkey:
:apple:

shiftkey:
Added a couple of Windows-specific issues, but this can be reviewed and merged at any time as it's not code-related...

iAmWillShepherd (approved):

macotouenaca (commented):
稼ぐ

PR #3949, opened 2018-02-06
drop back to v3 API to retrieve pull request data

shiftkey:
I was meant to be writing docs but instead found myself thinking about how we're listing external contributors in the Desktop release notes.

Turns out we're only looking at the commit message:

https://github.com/desktop/desktop/blob/b6431e6e0a782e1cde78b7fcdb5a56e872596e62/script/changelog/run.ts#L31-L48

This seems suboptimal if we want to start doing more co-authoring, like I did in https://github.com/desktop/desktop/pull/3940/commits/efdbdbc42b7db668375a734cac6f91326e1a16e7 and https://github.com/desktop/desktop/pull/3916/commits/da2b4173a1d00c7aabc2cc9834e764ecb6e3d8b4 to build upon someone else's work.
 
So what have I done here?

 - switch over to use `@octokit/rest.js` as it's battle-tested, has native async/await support and also pagination support (and you can debug it easily by setting `DEBUG=octokit:rest*` as an environment variable)
 - using the API to get all the commit SHAs associated with a PR, as well as the author/committer pairs already visible on a PR. This makes it easier for us to know which commits to scan, without tracing repository history.
 - using the API to get the current core team members (requires `org:read` scope on token) to then exclude when listing external contributors

I don't think we can get the co-author information from the API currently, but a future version of this could use `git interpret-trailers` and the search API to add these users to the existing author/committer pair.


script/changelog/run.ts
```diff
@@ -72,15 +73,19 @@ function getChangelogEntry(commit: IParsedCommit, pr: IAPIPR): string {
   }
 
   let attribution = ''
-  if (commit.owner !== OfficialOwner) {
-    attribution = `. Thanks @${commit.owner}!`
+
+  if (externalContributors.length > 0) {
+    // TODO: format this with an "and" between the last two contributors
```
j-f1: [`listify`](https://www.npmjs.com/package/listify)?

script/changelog/api.ts
```diff
@@ -1,60 +1,128 @@
...
-  readonly data: {
-    readonly repository: {
-      readonly pullRequest: IAPIPR
-    }
+interface IAPIPullRequest {
+  readonly title: string
+  readonly body: string
+  readonly user: {
+    readonly login: string
   }
```
j-f1: `interface IAPIUser { readonly login: string }`?

script/changelog/api.ts
```diff
@@ -1,60 +1,126 @@
...
+interface IAPITeam {
+  readonly name: string
+  readonly id: number
+}
+
+interface IAPITeamMember {
+  readonly login: string
+}
+
+const webflowAccount = ['web-flow']
```
nerdneha: can we put in a comment to describe what 'web-flow' / why it's a member of the core team?

script/changelog/parser.ts
```diff
@@ -75,7 +84,12 @@ export async function convertToChangelogFormat(
         throw new Error(`Unable to get PR from API: ${commit.prID}`)
       }
 
-      const entry = getChangelogEntry(commit, pr)
+      const collaborators = pr.collaborators
+      const externalContributors = collaborators.filter(
+        c => coreMembers.indexOf(c) === -1
```
nerdneha: is:
`!(c in coreMembers)` instead of
`c => coreMembers.indexOf(c) === -1`
clearer?

nerdneha:
@shiftkey can you help me understand why we dropped from graphql to v3? the biggest one i'm seeing is that it has native async/await support but i'm trying to understand the "battle tested" part; was there any additional functionality limitation or risk you saw using it?

script/changelog/api.ts
```diff
@@ -1,60 +1,128 @@
...
+      commitsResponse = await octokit.getNextPage(commitsResponse)
+      data = data.concat(commitsResponse.data)
+    }
+
+    const collaborators: Array<string> = []
+    const commits: Array<string> = []
+
+    for (const commit of data) {
+      commits.push(commit.sha)
+      if (commit.author && collaborators.indexOf(commit.author.login) === -1) {
```
nerdneha: this took some squinting so i tried moving things around so their easier to read; i pulled out consts in the process so if space is a concern or there doesn't feel like there is much to gain, feel free to ignore:
```
    for (const commit of data) {
      const { sha, author, committer } = commit
      commits.push(sha)
      if (author != null && !(author.login in collaborators)) {
        collaborators.push(author.login)
      }
      if (committer != null && !(committer.login in collaborators)) {
        collaborators.push(committer.login)
      }
    }
```

nerdneha: at minimum we can make the first check in `if (commit.author &&` and the one below more explicit, ex: commit.author != null

nerdneha:
🍦

shiftkey:
> can you help me understand why we dropped from graphql to v3? the biggest one i'm seeing is that it has native async/await support but i'm trying to understand the "battle tested" part; was there any additional functionality limitation or risk you saw using it?

@nerdneha our previous use of GraphQL was one artisan query we POSTed over HTTP and knew how to deserialize - not exactly flexible for more scenarios. I looked into using `apollo` and wanting type checking (because TypeScript) and easy pagination (for large PRs) but the dev experience wasn't great. We could come back to this at a later time, but I was grumpy and knew the data I needed so I switched back to the v3 API.

I say "battle tested" here because I've been following the maintainer of the octokit library over the last few months and he's spent a lot of time on it to make experience of using the GitHub API from JS great.

script/changelog/parser.ts
j-f1: The `in` operator checks if the specified **key** is in the object (i.e. `1 in ['a', 'b', 'c']` would return `true` since `['a', 'b', 'c'][1]` exists). There’s `.includes()`, which does the same thing as `.indexOf(...) !== -1`.

Since there’s `includes`, should we open a new issue to discuss removing `indexOf` checks that test against `-1`?

shiftkey:
Moving this into 1.2 as it's not a blocker on 1.1

script/changelog/parser.ts
nerdneha: ah good points @j-f1, maybe we'll punt for a larger discussion in a separate issue

niik:
@shiftkey Could we move the dependencies of this script to its own `package.json`?

shiftkey:
@niik as in introduce a `script/package.json` like we've done with `app/package.json`?

niik:
> @niik as in introduce a script/package.json like we've done with app/package.json?

A `script/changelog/package.json`, but yeah.

shiftkey:
There's a bit more work to do here around organizing the packages, so I'm going to drop the review label here as it's not close to merging.

shiftkey:
Another benefit of making more use of the GitHub API - we can look for labels like `infrastruture` in https://github.com/desktop/desktop/pull/4312 and avoid generating release entries (because they're not user-facing changes to the codebase).

script/changelog/api.ts
shiftkey: I went one better than your snippet and used a `Set<string>` here:

```
if (author != null && !collaborators.has(author.login)) {
  collaborators.add(author.login)
}
```

.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+app/node_modules/
+script/changelog/node_modules
```
j-f1: I’m pretty sure `node_modules` directories are ignored by default.

.prettierignore
```diff
@@ -1,16 +1,19 @@
-out/
+app/node_modules/
```
j-f1: Same here.

.eslintignore
shiftkey: This got caught by

https://github.com/desktop/desktop/blob/d73112a9fc9ca4a953697ae471d3942ae8e07d7b/script/eslint.ts#L17

.prettierignore
shiftkey: 👍 I _think_ I can undo this change given it was related to eslint. But I think the file needed a tidy-up anyway.

shiftkey:
Gonna close this until we have a better motivation for updating the changelog generation process.

PR #3953, opened 2018-02-07
propagate fetch progress when initializing remote for fork

shiftkey:
The first time you open a PR from a fork, Desktop will `git remote add` and then `git fetch` this new remote. This can take a couple of seconds before Desktop will then checkout the PR ref, and we don't display any visual feedback of what's happening.

But now we do!

**Before**

Note the period of no feedback, and then the checkout of the PR branch.

**After**

Note that we now see `Fetching {remote}` (it only takes a couple of seconds to fetch all the refs, so we don't get real progress beyond "Hang on...") and then the checkout occurs.


shiftkey:
Apologies for not crafting the GIFs to pause between loops, but hopefully you get the idea.

Issue #3882, opened 2018-01-30
Opening repository in Git Bash results in catastrophic error


gilmae:
## Description
Unable to open repository in Git Bash from the desktop app. Restarting the app results in the same error every time.

### Version
Version: 1.0.13
OS: Windows 10.0.16299

### Replication Steps
Expected behavior: Opened in Git Bash
Actual behavior: Error on the desktop app
Reproduces how often: 100%

#### Logs
Error: spawn where ENOENT
    at exports._errnoException (util.js:1050:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
    at onErrorNT (internal/child_process.js:367:16)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickCallback (internal/process/next_tick.js:104:9)

Version: 1.0.13
OS: Windows 10.0.16299

Related to #3825

shiftkey:
@gilmae I've not had a chance to look into this due to #3825 being challenging to implement correctly without `START`, but I've published an update to the [beta channel](https://github.com/desktop/desktop#beta-channel) which catches this error and adds some better tracing around launching shells.

If you're able to open this up, trigger the error and look at the logs there should be some more detail about how your setup differs to mine.

If you don't have time to switch, this will land in the stable channel next week.

gilmae:
[2018-02-18.desktop.beta.log](https://github.com/desktop/desktop/files/1734516/2018-02-18.desktop.beta.log)

Log attached from trying to open a repo in Git Bash.

Same results as described in issue description.

shiftkey:
@gilmae thanks for confirming the app now doesn't crash. 

I'm curious why [this message](https://github.com/desktop/desktop/blob/master/app/src/lib/shells/win32.ts#L139) isn't being logged as that would have told me what it was trying to spawn.

We rely on Git for Windows adding a certain registry key that points to it's install location. To see what's set on your machine, you can run this command in PowerShell:

```
PS C:\Users\shiftkey> Get-ItemProperty HKLM:\Software\GitForWindows

CurrentVersion : 2.16.1
InstallPath    : C:\Program Files\Git
LibexecPath    : C:\Program Files\Git\mingw64\libexec\git-core
PSPath         : Microsoft.PowerShell.Core\Registry::HKEY_LOCAL_MACHINE\Software\GitForWindows
PSParentPath   : Microsoft.PowerShell.Core\Registry::HKEY_LOCAL_MACHINE\Software
[...]

Issue #3825, opened 2018-01-19
 Opening repository in command prompt results in catastrophic error


laZaru5ap:
### Description

Unable to open repository in command prompt from the desktop app. Restarting the app results in the same error every time.

### Version
1.0.12

**GitHub Desktop version:** 1.0.12

**OS version:** Windows 6.1.7601 (Windows 7 Ultimate SP1 64-bit)

### Steps to Reproduce

![github error_open in command prompt](https://user-images.githubusercontent.com/28505948/35144763-8ef064b4-fd2b-11e7-9057-3b0990199a7a.png)

**Expected behavior:** Opened in command prompt

**Actual behavior:** Error on the desktop app

**Reproduces how often:** 100%

### Logs
Error: spawn where ENOENT
    at exports._errnoException (util.js:1050:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:193:32)
    at onErrorNT (internal/child_process.js:367:16)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickCallback (internal/process/next_tick.js:104:9)

Version: 1.0.12
OS: Windows 6.1.7601

shiftkey:
@laZaru5ap thanks for the report.

The code to launch `cmd` is rather straightforward:

```
const cp = spawn('START', ['cmd'], { shell: true, cwd: path })
```

My gut feeling is that your `PATH` value isn't configured as expected.

Could you open a command prompt, run these two commands and attach the output here:

```
> echo %PATH%
> where cmd
```

This is what it looks like on one of my developer VMs:

```
Microsoft Windows [Version 10.0.16299.192]
(c) 2017 Microsoft Corporation. All rights reserved.

C:\Users\shiftkey>echo %PATH%
C:\Program Files (x86)\Parallels\Parallels Tools\Applications;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\; ...

C:\Users\shiftkey>where cmd
C:\Windows\System32\cmd.exe
```

gilmae:
I'm seeing the same issue as the OP. Actually, I've seen it no matter what shell I opt for - Git Bash, Powershell, or Command Prompt.

> Microsoft Windows [Version 10.0.16299.192]
> (c) 2017 Microsoft Corporation. All rights reserved.
> 
> C:\Users\username>where cmd
[...]

PR #3954, opened 2018-02-07
better error handling around ENOENT

shiftkey:
This lays the groundwork for resolving #3825 and #3882 by:

 - adding tracing when invoking shells for all OSes, not just Windows
 - handling the `error` event on a spawned child process (if we don't, this crashes the renderer)
 - display the error to the user rather than hiding it away

See for yourself (on Windows):

 - apply this patch to this branch

```diff
diff --git a/app/src/lib/shells/win32.ts b/app/src/lib/shells/win32.ts
index bd814808f..5ab2cb068 100644
--- a/app/src/lib/shells/win32.ts
+++ b/app/src/lib/shells/win32.ts
@@ -141,7 +141,9 @@ export function launch(
       cwd: path,
     })
   } else if (shell === Shell.Cmd) {
-    return spawn('START', ['cmd'], { shell: true, cwd: path })
+    const executable = `"C:\\Windows\\System32\\cmd.exe"`
+    log.info(`launching ${shell} at path: ${executable}`)
+    return spawn(executable, undefined, { shell: true, cwd: path })
   } else {
     return assertNever(shell, `Unknown shell: ${shell}`)
   }
```

 - run the app up
 - use **Repository** | **Open in Command Prompt** to trigger this error

Why this doesn't work as expected is what I'm currently wrestling with, but it'd be nice to get this into the next update so we have better diagnostics.

### DIFF
--- a/app/src/lib/shells/linux.ts
+++ b/app/src/lib/shells/linux.ts
@@ -99,23 +99,22 @@ export async function getAvailableShells(): Promise<
   return shells
 }
 
-export function launch(shell: IFoundShell<Shell>, path: string): ChildProcess {
-  if (shell.shell === Shell.Urxvt) {
-    const commandArgs = ['-cd', path]
-    return spawn(shell.path, commandArgs)
+export function launch(
+  foundShell: IFoundShell<Shell>,
+  path: string
+): ChildProcess {
+  const shell = foundShell.shell
+  switch (shell) {
+    case Shell.Urxvt:
+      return spawn(foundShell.path, ['-cd', path])
+    case Shell.Konsole:
+      return spawn(foundShell.path, ['--workdir', path])
+    case Shell.Xterm:
+      return spawn(foundShell.path, ['-e', '/bin/bash'], { cwd: path })
+    case Shell.Tilix:
+    case Shell.Gnome:
+      return spawn(foundShell.path, ['--working-directory', path])
+    default:
+      return assertNever(shell, `Unknown shell: ${shell}`)
   }
-
-  if (shell.shell === Shell.Konsole) {
-    const commandArgs = ['--workdir', path]
-    return spawn(shell.path, commandArgs)
-  }
-
-  if (shell.shell === Shell.Xterm) {
-    const commandArgs = ['-e', '/bin/bash']
-    const commandOptions = { cwd: path }
-    return spawn(shell.path, commandArgs, commandOptions)
-  }
-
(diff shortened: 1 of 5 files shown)


shiftkey:
Moving this into 1.1 as I think it's a reasonably safe non-functional change.

app/src/lib/shells/linux.ts
```diff
@@ -99,26 +99,23 @@ export async function getAvailableShells(): Promise<
   return shells
 }
 
-export async function launch(
-  shell: IFoundShell<Shell>,
-  path: string
-): Promise<void> {
+export function launch(shell: IFoundShell<Shell>, path: string): ChildProcess {
```
iAmWillShepherd: Seems like this may be a good candidate for a switch statement.

app/src/lib/shells/win32.ts
```diff
@@ -114,52 +114,35 @@ export async function getAvailableShells(): Promise<
...
-
-  cp.on('exit', code => {
-    if (code !== 0) {
-      log.debug(`[${context}] exit code: ${code}`)
-    }
-  })
-}
-
-export async function launch(
+export function launch(
```
iAmWillShepherd: This can also be tidied up with a switch.

app/src/lib/shells/win32.ts
```diff
@@ -114,52 +114,35 @@ export async function getAvailableShells(): Promise<
...
-    const cp = spawn(
-      'START',
-      ['powershell', '-NoExit', '-Command', psCommand],
-      {
-        shell: true,
-        cwd: path,
-      }
-    )
-    addErrorTracing(`PowerShell`, cp)
+    return spawn('START', ['powershell', '-NoExit', '-Command', psCommand], {
```
iAmWillShepherd: I notice you are using `log.info` to give some info about other shells. Any reason not to do it for PowerShell?

shiftkey: @iAmWillShepherd there's no "path" here to log

app/src/lib/shells/linux.ts
shiftkey: 9a7484ea7148345bb2d2d2cadc966778ef2033f8

app/src/lib/shells/win32.ts
shiftkey: fe98607d9660958ea4460ffc844d5f97a40740c4

shiftkey:
:apple:

PR #1430, opened 2017-05-11
update Electron to 1.6.11

shiftkey:
So I was able to workaround much of the incompatibilities between the two type declarations, with some rough edges called out. It'd be nice to ensure the rough edges are sorted before merging, rather than letting it hang around.

@desktop/electron where's the best place to report these issues upstream - an open issue on `electron-typescript-declarations` walking through the samples and cross-referencing the API docs that need to be updated so it all flows through? Or something else?

Fixes #671 

TODO:

 - [x] y u no build
 - [x] verify self-signed certificate flow on macOS
 - [x] verify self-signed certificate flow on Windows

PR #3958, opened 2018-02-07
update Electron to 1.8.3

iAmWillShepherd:
Updates the app to use electron 1.8.2 and updates related dependencies to latest versions.

- [x] Test build
  - [x] macOS
  - [x] Windows
  - [x] Linux


package.json
```diff
@@ -128,10 +128,10 @@
     "@types/uuid": "^3.4.0",
     "@types/winston": "^2.2.0",
     "@types/xml2js": "^0.4.0",
-    "electron": "1.7.11",
-    "electron-builder": "19.48.3",
+    "electron": "1.8.2",
```
shiftkey: You need to bump `app/.npmrc` to this same version to ensure native node modules are built correctly.

package.json
```diff
@@ -128,10 +128,10 @@
...
+    "electron-builder": "19.56.0",
```
shiftkey: Be mindful that we only use this for the Linux installer creation (which isn't done by CI currently).

package.json
iAmWillShepherd: Yes, I spent about 30 mins figuring that out! 👍

package.json
iAmWillShepherd: ok, I will revert this back to the previous version to see if it plays nice with Electron 1.8.

shiftkey:
Don't forget to nuke the `package-lock.json` file :boom::boom::boom:

shiftkey:
I'd also recommend cutting a test release and validating the installer works for these platforms

iAmWillShepherd:
I released a test release in central (everything seems to work 🤞).

iAmWillShepherd:
🔢

niik:
We should discuss whether to hold this until after 1.1. It's a large change and I'd like to have it sit on beta for at least a little while before we ship it.

shiftkey:
@niik yep, that was my plan. I'll move this into 1.2 to make it extra clear.

niik:
Just saw [an issue about crashes on Windows 10 Fall Creator Update](https://github.com/electron/electron/issues/11871) in the electron repo. We should make sure we've tested it out on that particular Windows version if we haven't already.

iAmWillShepherd:
I tested this on the latest version of Windows (full 2 hours of updates and restarts) and it seems to work just fine.

shiftkey:
I've bumped this to match the package upgrades in #4198, and we should do another pass of testing as part of reviewing...

app/package.json
```diff
@@ -58,7 +58,8 @@
   "devDependencies": {
     "devtron": "^1.4.0",
     "electron-debug": "^1.1.0",
-    "electron-devtools-installer": "^2.2.1",
+    "electron-devtools-installer": "^2.2.3",
+    "react-addons-perf": "15.4.2",
```
shiftkey: This dependency needs to be removed :fire:

shiftkey:
> We should make sure we've tested it out on that particular Windows version if we haven't already.

I've tested Electron 1.8.3 against that specific build and it's working fine.

My gut feeling about that upstream issue is that it's related to using Electron within a test runner - we had an issue with `electron-mocha` ages ago https://github.com/desktop/desktop/pull/1430#issuecomment-302373025 which required a fix to the test runner itself https://github.com/jprichardson/electron-mocha/pull/105

niik:
> I've tested Electron 1.8.3 against that specific build and it's working fine.

👍

PR #3962, opened 2018-02-08
tidying up docs

shiftkey:
🌵🌵🌵 Depends on #3975, go review that first 🌵🌵🌵

-  [x] review `editor-integration.md`
-  [x] review `shell-integration.md`
-  [x] review `syntax-highlighting.md`

### DIFF
--- a/docs/technical/editor-integration.md
+++ b/docs/technical/editor-integration.md
@@ -1,4 +1,4 @@
-# "Open in External Editor" integration
+# "Open External Editor" integration
 
 GitHub Desktop supports the user choosing an external program to open their
 local repositories, and this is available from the top-level **Repository** menu 
--- a/docs/technical/editor-integration.md
+++ b/docs/technical/editor-integration.md
@@ -1,8 +1,8 @@
-# Editor Integration
+# "Open in External Editor" integration
 
 GitHub Desktop supports the user choosing an external program to open their
-local repositories, and this is available from the main menu and right-clicking
-on a repository in the sidebar.
+local repositories, and this is available from the top-level **Repository** menu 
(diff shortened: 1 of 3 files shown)


docs/technical/editor-integration.md
```diff
@@ -1,8 +1,8 @@
-# Editor Integration
+# "Open in External Editor" integration
```
nerdneha: maybe 
* "Open External Editor" integration
or 
* "Open in [your external editor]" integration

shiftkey: I kind of want to keep the `in` around because it matches up with the menu item in Desktop, but the latter option feels a bit unclear. I'll drop the `in` and see how it feels...

docs/technical/shell-integration.md
```diff
@@ -46,22 +46,50 @@ use **Git Bash** as a reference for the rest of the process.
...
+This is the example for how we resolve if Git Bash is installed.
 
 ```ts
-const gitBash = await readRegistryKeySafe(
-  'HKEY_LOCAL_MACHINE\\SOFTWARE\\GitForWindows'
-)
-if (gitBash.length > 0) {
-  const installPathEntry = gitBash.find(e => e.name === 'InstallPath')
-  if (installPathEntry) {
+  const gitBashPath = await findGitBash()
```
shiftkey: Just a heads up: these examples are actually based off https://github.com/desktop/desktop/pull/3975 which hasn't been merged yet. These are internals but I found it compelling enough to address now rather than later.

shiftkey:
Dropping this from 1.1 as it's no longer critical to the post I was writing.

Issue #3597, opened 2017-12-13
Cannot use `esc` on an empty Pull Request


tierninho:
If there are no Pull Requests (on a published branch), and you try to close the list with `esc` key, the Pull Request tab is highlighted and the list is not closed as expected.

<img width="348" alt="screen shot 2017-12-13 at 12 08 47 pm" src="https://user-images.githubusercontent.com/14828183/33960236-ef833aa8-dffe-11e7-98c7-81d665fdafaf.png">

1.0.11-beta0, mac

iAmWillShepherd:
It looks like this behavior is consistent with how the branches drop down. 

It looks like we are relying on the filter text input to capture the `esc` key, but since it's possible for the text input to not be in focus or to not even exists, pressing `esc` doesn't always work.

Is this intentional?

cc, @desktop/design

iAmWillShepherd:
Bump for feedback.

donokuda:
My gut says <kbd>esc</kbd> should close the panel.

agisilaos:
Since #3972 is merged we should close this issue right? @iAmWillShepherd

shiftkey:
@agisilaos nice spot!

PR #3972, opened 2018-02-08
Escape behavior should match that of Branch List

iAmWillShepherd:
Fixes issue #3597

## Before
_see original issue for gif_
## After
fix


tierninho:
✔️  looks good

app/src/ui/branches/branches-container.tsx
```diff
@@ -65,17 +67,23 @@ export class BranchesContainer extends React.Component<
     }
   }
 
-  private onFilterKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+  private closeFoldoutOnEsc = (when: () => boolean) => (
```
shiftkey: :lipstick: could this be named to something more descriptive? `shouldCloseFoldout`?

shiftkey: Sorry, I actually meant the `when` parameter here, not the function name 🤦‍♂️

app/src/ui/branches/branches-container.tsx
```diff
@@ -65,17 +67,23 @@ export class BranchesContainer extends React.Component<
     }
   }
 
-  private onFilterKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+  private shouldCloseFoldout = (when: () => boolean) => (
```
shiftkey: Sorry, I actually meant the `when` parameter here, not the function name 🤦‍♂️ 

`onFilterKeyDown` here was fine.

iAmWillShepherd: Where's the Homer Simpson Doh! when you need it?! 😆

iAmWillShepherd:
🔢

app/src/ui/branches/branches-container.tsx
```diff
@@ -134,9 +142,11 @@ export class BranchesContainer extends React.Component<
             allBranches={this.props.allBranches}
             recentBranches={this.props.recentBranches}
             onItemClick={this.onItemClick}
-            filterText={this.state.filterText}
-            onFilterKeyDown={this.onFilterKeyDown}
-            onFilterTextChanged={this.onFilterTextChanged}
+            filterText={this.state.branchFilterText}
+            onFilterKeyDown={this.closeFoldoutOnEsc(
+              () => this.state.branchFilterText.length === 0
```
niik: Having anonymous functions inline like this means that the component will always re-render since every time this function is called a new instance of this function will be instantiated. We have a [tslint rule](https://github.com/desktop/desktop/blob/a991d2a0469120b843e06950add087fd6a2be2d2/tslint-rules/reactNoUnboundDispatcherPropsRule.ts) to prevent this but it only appears to work from within `render()`

This should be `onFilterKeyDown={this.closeFoldoutOnEsc}` and `closeFoldoutOnEsc` should be a bound method (i.e. `=>`) so that it's able to access this.state

app/src/ui/branches/branches-container.tsx
```diff
@@ -181,9 +184,7 @@ export class BranchesContainer extends React.Component<
           pullRequests={pullRequests}
           onSelectionChanged={this.onPullRequestSelectionChanged}
           filterText={this.state.pullRequestFilterText}
-          onFilterKeyDown={this.closeFoldoutOnEsc(
-            () => this.state.pullRequestFilterText.length === 0
-          )}
+          onFilterKeyDown={this.onPullRequestFilterKeyDown()}
```
j-f1: `{this.onPullRequestFilterKeyDown}`?

j-f1: oh, I see how you’re using this. That still ends up creating a new function in every render.

iAmWillShepherd: Suggestions to prevent this? I know I can duplicate the code, but was hoping to avoid that.

j-f1: You could do `onPullRequestFilterKeyDown = this.closeFoldoutOnEsc(...)` in the class body instead of defining it as a method.

iAmWillShepherd: Ahh, so like a field. Seems simple enough 👍

iAmWillShepherd: Linter wouldn't allow me to do that, so I just duplicated code.

iAmWillShepherd:
🔢

iAmWillShepherd:
I merged the branch `show-no-pr` into this PR to avoid confusion on my part. 🙇 @shiftkey for reviewing the PR I merged into this one.

app/src/ui/branches/branches-container.tsx
```diff
@@ -163,39 +181,37 @@ export class BranchesContainer extends React.Component<
...
+        pullRequests={pullRequests}
+        selectedPullRequest={this.state.selectedPullRequest}
+        repositoryName={name}
+        isOnDefaultBranch={!!isOnDefaultBranch}
+        onSelectionChanged={this.onPullRequestSelectionChanged}
+        onCreateBranch={this.onCreateBranch}
+        onCreatePullRequest={this.onCreatePullRequest}
+        filterText={this.state.pullRequestFilterText}
+        onFilterTextChanged={this.onPullRequestFilterTextChanged}
+        onFilterKeyDown={this.closePullRequrstFoldoutOnEsc}
```
shiftkey: Typo: `Requrst`

app/src/ui/branches/branches-container.tsx
```diff
@@ -134,9 +152,9 @@ export class BranchesContainer extends React.Component<
             allBranches={this.props.allBranches}
             recentBranches={this.props.recentBranches}
             onItemClick={this.onItemClick}
-            filterText={this.state.filterText}
-            onFilterKeyDown={this.onFilterKeyDown}
-            onFilterTextChanged={this.onFilterTextChanged}
+            filterText={this.state.branchFilterText}
+            onFilterKeyDown={this.closeBranchFoldoutOnEsc}
```
niik: I think this handler should be attached to the entire container. You should be able to hit escape while navigating the list of PRs as well as in the filter text.

app/src/ui/branches/branches-container.tsx
```diff
@@ -65,17 +66,34 @@ export class BranchesContainer extends React.Component<
     }
   }
 
-  private onFilterKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+  private closePullRequrstFoldoutOnEsc = (
+    event: React.KeyboardEvent<HTMLElement>
+  ) => {
+    if (event.key === 'Escape') {
```
niik: We should check that no handler has taken care of this before us by doing `if (!event.defaultPrevented && e.key ===`

app/src/ui/branches/branches-container.tsx
iAmWillShepherd: Do you mean attaching it to the `TabBar` component? If so, one of the things I struggled with was adding an event handler to plain Dom node and getting it to respond to React's synthetic events.

niik: No I mean we should attach it to the branches-container div. So something like

```tsx
<div className="branches-container" onKeyDown={...}>
```

And the in the handler we'd check whether the event was handled or and if the key was 'escape'. The next change would be to have `TextBox` handle `Escape` when `type=search` and it's not empty. That way we can keep the logic specific to search inputs contained within the text component and everyone else that needs to handle escape can do their thing without having to take search boxes into account. We could take a look at this together next week if you'd like

iAmWillShepherd:
The current fix for this bug introduces a new issue: race conditions. This is due to our reliance on props' values in the filter list. 

@niik suggested we refactor so that we can use the values from the DOM element itself. I will be doing the following in separate PRs:
* Remove `onChange` event in favor of `onValueChanged` in`TextBox` component
* Stop leaking implementation details from `TextBox` by removing 'onInputRef` props and exposing two new public methods, `focus` and `selectAll`

niik:
#4077 and #4080 has landed, let's merge them in here and bring it all together!

j-f1:
Before this gets merged, can you fix the typo in the PR title? (b**a**havior → b**e**havior)

iAmWillShepherd:
🔢

app/src/ui/lib/text-box.tsx
```diff
@@ -177,6 +177,14 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
     )
   }
 
+  private onKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
```
niik: Now that we've moved TextBox to not expose the onChange event directly I think we should move the Escape check in here instead.

```ts
if (event.key === 'Escape' && this.props.type === 'search') {
  if (this.inputElement && this.inputElement.value.length > 0) {
    event.preventDefault()
    this.props.onValueChanged('')
  }
}
```

app/src/ui/lib/text-box.tsx
```diff
@@ -177,6 +177,14 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
...
+    const eventHandler = this.props.onKeyDown
+
+    if (eventHandler != null) {
```
niik: Can we skip the intermediate variable here? Doesn't seem to be giving us much over repeating it in the if-statement and the method call

app/src/ui/lib/text-box.tsx
j-f1: `onEscape`?

iAmWillShepherd: I initially was going to put it in that component, then I thought about the reason why we use `Esc` to clear the contents of the textbox and decided it should be in the filter list. It seems that we only want to `Esc`to clear the contents of in filter-lists and no where else. Is that an incorrect assumption on my part?

niik: @iAmWillShepherd The way I look at it is it's a behavior of a search input field. Like, if we were to use the `TextBox` with `type=search` somewhere outside of a filter list we'd still want to clear its text when you hit escape. It just so happens to be the case that we only use search textboxes inside of filter lists at the moment.

> onEscape

@j-f1 We try to match event handler names to event names so it should be `onKeyDown`

j-f1: @niik I was suggesting having an event that only gets called when the escape key is pressed.

iAmWillShepherd:
🔢

app/src/ui/lib/text-box.tsx
```diff
@@ -178,6 +178,21 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
   }
 
   private onKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (
+      this.state.value !== '' &&
+      this.props.type === 'search' &&
+      event.key === 'Escape'
+    ) {
+      const value: string = ''
```
j-f1: I don’t think you need the `: string` here.

iAmWillShepherd: I didn't either, I put it there to appease VSCode. I'll try restarting my editor to see if it still complains when I remove the type annotation.

iAmWillShepherd:
All good @j-f1, turns out a simple restart was all that I needed.

app/src/ui/lib/filter-list.tsx
```diff
@@ -340,7 +341,14 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
       return
     }
 
-    if (event.key === 'ArrowDown') {
+    const filterText = this.props.filterText
+    if (key === 'Escape' && (filterText && filterText.length !== 0)) {
```
niik: This section would appear to be redundant now since it's handled in the TextBox. Could we remove it please?

app/src/ui/lib/text-box.tsx
```diff
@@ -177,6 +177,27 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
...
+
+      event.preventDefault()
+      this.setState({ value })
+
+      if (this.props.onValueChanged) {
+        this.props.onValueChanged(value)
+      }
+    }
+
+    if (this.props.onKeyDown != null) {
```
niik: Could we make this `(this.props.onKeyDown !== undefined)` since it's typed as undefined or function or just `if(this.props.onKeyDown)`?

app/src/ui/branches/branches-container.tsx
```diff
@@ -34,58 +33,36 @@ interface IBranchesProps {
   readonly isLoadingPullRequests: boolean
 }
 
-interface IBranchesState {
+interface IBranchesContainerState {
   readonly selectedBranch: Branch | null
   readonly selectedPullRequest: PullRequest | null
-  readonly filterText: string
+  readonly branchFilterText: string
+  readonly pullRequestFilterText: string
```
niik: I really like that we're lifting this state into BranchesContainer

iAmWillShepherd:
> Remove the onDismissed prop from clone-repository.tsx

@niik , I've addressed all points except ☝️. We use that `onDismissed` prop to close the dialog. I did remove it from `CloneGithubRepository` thought.

🔢

PR #3975, opened 2018-02-09
add further tracing when spawning shells on Windows

shiftkey:
I was refreshing the docs in #3962 and realised this area of the codebase had gotten rather gnarly, so here's what I want to address:

 - [x] improve readability of the main function by extracting the logic out and making the flow a bit easier to understand
 - [x] use `pathExists` where appropriate as a sanity check (if it doesn't exist, we're gonna have a bad time) or log a debug message if this assumption fails
 - [x] test these shells locally to ensure they work as expected for known good configs
    - [x] cmd
    - [x] PowerShell
    - [x] Hyper
    - [x] Git Bash
 - [x] measure % change in time resolving all shells (worst case, all installed)


app/src/lib/shells/win32.ts
```diff
@@ -43,75 +45,127 @@ export async function getAvailableShells(): Promise<
...
+  if (
+    first.type === RegistryValueType.REG_EXPAND_SZ ||
+    first.type === RegistryValueType.REG_SZ
+  ) {
+    const path = first.data.replace(
+      /^%SystemRoot%/i,
+      process.env.SystemRoot || 'C:\\Windows'
+    )
+
+    if (await pathExists(path)) {
```
shiftkey: The cumulative effect of checking the path exists is a couple of milliseconds (`master` takes `~26ms`, this branch takes `~31ms`) and given this is deferred until you first open the Options/Preferences dialog - and is cached - I'm fine with being more safe here.

maxbed:
09.02.2018. 05.59, "Brendan Forster" <notifications@github.com> је
написао/ла:

shiftkey:
This is less pressing given that #4044 is the root cause of our shells issue, but it's still some nice improvements to have. Moving to 1.2 to review once things have settled down.

PR #3774, opened 2018-01-12
Add Elixir syntax highlighting

joaovitoras:
CodeMirror no longer accepting new modes into the main distribution. I had to add codemirror-mode-elixir package for elixir syntax highlighting
Example files can be found on this PR: https://github.com/desktop/highlighter-tests/pull/16

Example:
Elixir Example

PR #3990, opened 2018-02-11
latest yarn merges two lockfile entries

shiftkey:
If you have yarn `v1.3.2` installed and run `yarn` on master (from the repository root) you'll probably see the lockfile is modified (tested on both macOS and Windows).

I think this crept in as part of #3774, and it seems to be tidying up the lockfile contents and not a behaviour change.


iAmWillShepherd:
Skipping 24-hour hold to get this into `master` due to this issue blocking work.

Issue #1749, opened 2017-05-24
Prevent drag+drop of repos while performing certain actions?


tierninho:
Noticed that you can drag+drop repos in Desktop at pretty much anytime, including when you are confirming a repo deletion, branch renaming, and merge into current branch. Only potential "problem" is that it switches the focus of Desktop onto the new repo, while leaving open the dialogue from the previous repo. (Note the action performed is still performed regardless).

Should we block this from happening?

The dialogue matches the repo:
<img width="1109" alt="screen shot 2017-05-24 at 1 55 25 pm" src="https://cloud.githubusercontent.com/assets/14828183/26418041/cf648f94-4088-11e7-8a94-72a739a81b40.png">

The dialogue does not matches the repo:
<img width="1119" alt="screen shot 2017-05-24 at 1 55 46 pm" src="https://cloud.githubusercontent.com/assets/14828183/26418040/cf46086c-4088-11e7-9456-e4643e23b866.png">

v 0.5.8, OS X10.11.5

PR #3996, opened 2018-02-13
Disable drag-and-drop interaction when a popup is in the foreground

iAmWillShepherd:
Fixes issue #1749

## Before
before

## After
after

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -540,6 +540,9 @@ export class App extends React.Component<IAppProps, IAppState> {
     }
 
     document.body.ondrop = e => {
+      if (this.state.currentPopup != null) {
+        return
+      }
       const files = e.dataTransfer.files
       this.handleDragAndDrop(files)
       e.preventDefault()


j-f1:
Is there a way to make the cursor have a 🚫 on it so the folder doesn’t just get dropped into oblivion?

shiftkey:
@j-f1 maybe. I think you want to listen to the `dragover` event and set `e.dataTransfer.dropEffect = 'none'` to reject it based on [this doc](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect). Not sure about the other options, or how this works on other OSes.

I'll let @iAmWillShepherd chime in too.

j-f1:
Maybe `move` or `copy` for when the drag-and-drop will be accepted?

tierninho:
✅  tested for UI and functionality

shiftkey:
Moving the enhancement discussion out to https://github.com/desktop/desktop/issues/4004 as that can be addressed separate to this fix.

iAmWillShepherd:
@shiftkey you got to it before I could. 

@j-f1, that should be a separate issue since we don't change the cursor on drag-and-drop events when there isn't a popup in the foreground.

PR #4000, opened 2018-02-13
Ignore Visual Studio files

iAmWillShepherd:
Noticed we had some VS files in the repo after I opened the project on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ app/node_modules/
 .DS_Store
 .awcache
 .idea/
+.vs/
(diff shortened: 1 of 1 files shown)


shiftkey:
@iAmWillShepherd interesting - could this be because of VS2017 being used to build the native modules?

iAmWillShepherd:
I believe VS performs some indexing to make code-navigation faster as well as to provide intellisense. All the work it performs is placed in that folder along with, potentially, user settings and such.

PR #4016, opened 2018-02-14
Remove preview features doc and update feature flagging.

iAmWillShepherd:
Reading through the docs, noticed that we don't call out the features under preview. 

@desktop/core feels on screenshots or gifs to illustrate each feature?

### DIFF
--- a/docs/contributing/preview-features.md
+++ b/docs/contributing/preview-features.md
@@ -11,4 +11,6 @@ the Desktop app.
 
 ## Preview Features
 
-Coming Soon™
+* Enhanced branch selector that allows you to view pull requests
+
+* The [status](https://help.github.com/articles/about-statuses/) for each pull request is not surfaced in the UI
--- a/docs/contributing/preview-features.md
+++ b/docs/contributing/preview-features.md
@@ -1,3 +1,0 @@
-# Preview Features
-
-We are hard at work building the latest features, we will have something to show soon...
--- a/docs/contributing/preview-features.md
+++ b/docs/contributing/preview-features.md
@@ -1,16 +1,3 @@
 # Preview Features
 
-If you're running GitHub Desktop you can enable **"preview features"**, which
-we consider something we want to eventually ship to everyone, but needs some
-sort of refinement before we can do that, and in the meantime we would like
-feedback from users on these features.
-
-To opt-in for testing preview features, set the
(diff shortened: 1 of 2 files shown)


shiftkey:
@iAmWillShepherd this document is kind of sad, because we're just now mentioning features that have been on the beta channel for a while.

I think we should move the technical details about `GITHUB_DESKTOP_PREVIEW_FEATURES` into `docs/technical/feature-flagging.md` (it already gets a mention) and figure out a better way to share what stuff is currently on the beta channel that end users will be more likely to notice.

iAmWillShepherd:
Sounds good. And yea, I just noticed we had nothing there, so I figured I'd add it while no one was looking 😆

docs/technical/feature-flagging.md
```diff
@@ -49,6 +49,9 @@ feature once things are stabilized.
...
 
-To opt-in for testing preview features, set the
-`GITHUB_DESKTOP_PREVIEW_FEATURES` environment variable to any value and launch
-the Desktop app.
+Set the `GITHUB_DESKTOP_PREVIEW_FEATURES` environment variable to any value to opt-in for preview featurs.
+
+Don't have that environment variable? 
+No worries, simply create it. (here's a [handy guide](https://www.schrodinger.com/kb/1842) for doing that on most major OSs).
+
+Note, if you would like to opt-out of preview feautures, you will need to remove this variable from your environment.
```
j-f1: Do you have to restart Desktop each time you change the setting?

iAmWillShepherd: Good point. We should make that clear.

iAmWillShepherd:
🔢

docs/contributing/preview-features.md
```diff
@@ -1,14 +1,3 @@
...
-feedback from users on these features.
-
-To opt-in for testing preview features, set the
-`GITHUB_DESKTOP_PREVIEW_FEATURES` environment variable to any value and launch
-the Desktop app.
-
-## Preview Features
-
-Coming Soon™
+We are hard at work building the latest features, we will have something to show soon...
```
shiftkey: Let's just remove this whole file (and open an issue to track this idea properly) :fire::fire::fire:

iAmWillShepherd:
🔢

Issue #3955, opened 2018-02-07
1.2 Launch Checklist - Relationship between branches lite


nerdneha:
Date: ~Mon Apr 23~ ~Mon Apr 30~ ~Mon May 7~ Mon May 21 (if we can QA in time)

### Major Features
* Compare Branches Feature Tracker https://github.com/desktop/desktop/issues/3956
* Multi-file select https://github.com/desktop/desktop/pull/4188

#### OPEN MILESTONE + POPULATE ISSUES ✅ 

- [x] **[HD]** Create help docs issue - https://github.com/github/help-docs/issues/7091
- [x] **[BP]** Create blog post PR - https://github.com/github/blog/pull/279
- [x] **[GHD-MS]** Create marketing site issue - https://github.com/github/desktop.github.com/issues/99
- [x] **[RT]** Create release tracker card (@nerdneha) - https://github.com/github/technology/projects/2#card-9456023

#### EXECUTION PHASE

- [x] Notify help docs to start content strategy (@iAmWillShepherd)
- [x] Close 1.1.x milestone - https://github.com/desktop/desktop/milestone/13
- [x] Close all open issues in 1.2 milestone - https://github.com/desktop/desktop/milestone/12
- [x] **[HD]** help docs PR is created - https://github.com/github/help-docs/pull/7263
- [x] **[GHD-MS]** marketing site PR is created - https://github.com/github/desktop.github.com/pull/101
- [x] **[V]** Create video issue - https://github.com/github/video/issues/466
- [x] **[V]** Notify video to start video-ing - 

#### FINALIZE THE RELEASE DATE

- [x] Release to beta (latest beta on 4/27 can be downloaded [here](https://s3.amazonaws.com/github-desktop/releases/1.1.2-beta7-9e856ecf/GitHubDesktop.zip))
- [x] Flip feature flags for production and move to stable - https://github.com/desktop/desktop/pull/4584
- [x] **[QA]**  QA signoff: https://github.com/desktop/desktop/issues/4734
- [x] **[HD]** Help docs are ready to be merged 
- [x] **[V]** Gifs are created and in the blog post
- [x] **[BP]** Create draft for blog post
- [x] ~**[GHD-MS]**  Put marketing site changes on staging~
- [x] Update the changelog for `1.1.2`

#### IT'S GO TIME: LAUNCH THE THING
- [x] Release to Prod as `1.2.0`
[...]

PR #4018, opened 2018-02-15
release the kraken

shiftkey:
This makes the PR integration available for the next stable update :tada:

I also added a placeholder flag for the 1.2 work so I wouldn't have to eliminate the `enableBetaFeatures` function.

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -26,6 +26,11 @@ function enableBetaFeatures(): boolean {
   return enableDevelopmentFeatures() || __RELEASE_CHANNEL__ === 'beta'
 }
 
+/** Should the new Compare view be enabled? */
+export function enableCompareBranch(): boolean {
+  return enableBetaFeatures()
+}
+
 /** Should PR integration be enabled? */
 export function enablePRIntegration(): boolean {
   return true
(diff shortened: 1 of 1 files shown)


app/src/lib/feature-flag.ts
```diff
@@ -26,9 +26,14 @@ function enableBetaFeatures(): boolean {
   return enableDevelopmentFeatures() || __RELEASE_CHANNEL__ === 'beta'
 }
 
+/** Should the new Compare view be enabled? */
+export function enableCompareBranch(): boolean {
+  return enableBetaFeatures()
+}
+
 /** Should PR integration be enabled? */
 export function enablePRIntegration(): boolean {
```
shiftkey: I decided to leave this function around just in case we see something in the wild and want to back out.

nerdneha: thoughts on adding a task to remove this later on at some point? one idea could be the 1.2 launch checklist? #3955?

nerdneha:
🎏

app/src/lib/feature-flag.ts
shiftkey: Eh, I'd do it before the 1.2 launch - :thumbsup: to `tech-debt` task.

shiftkey: #4025

nerdneha:
🚀

Issue #2041, opened 2017-06-18
Write EOF error when comitting


cloewen8:
### Description

I get the following error when making a commit. The error occurs consistently.

> Error: write EOF
>     at exports._errnoException (util.js:1022:11)
>     at WriteWrap.afterWrite (net.js:804:14)
> 
> Version: 0.6.1
> OS: Windows 10.0.15063

The error may be related to the amount of staged files (3846 files). I am making the commit to remove the unused template. Some files may be missing from the template.

### Version

**OS version:** Microsoft Windows [Version 10.0.15063]

### Steps to Reproduce

The template was setup by OpenShift using the repository at [https://github.com/strongloop/openshift-cartridge-strongloop](https://github.com/strongloop/openshift-cartridge-strongloop).

OpenShift creates a new git repository using the manifest.yml file (located at [https://raw.github.com/strongloop/openshift-cartridge-strongloop/master/metadata/manifest.yml](https://raw.github.com/strongloop/openshift-cartridge-strongloop/master/metadata/manifest.yml).

The same repository structure can be achieved by:

1. Creating a new local repository.
2. Copying the files from [https://github.com/strongloop/openshift-cartridge-strongloop/tree/master/template](https://github.com/strongloop/openshift-cartridge-strongloop/tree/master/template).
3. Staging and committing the added files.

Attempting to commit the removal of all of the files causes the error to occur.

**Expected behaviour:** GitHub for Windows should show a waiting status, then a few seconds later the files should be successfully committed.

**Actual behaviour:** GitHub for Windows closes after waiting a few seconds with the same error message.

**Reproduces how often:** This error occurs consistently.

### Logs

The error logged stays consistent:

> 2017-06-18T04:58:02.738Z - info: [ui] Stats reported.
> 2017-06-18T04:58:47.905Z - error: [main] Error: write EOF
>     at exports._errnoException (util.js:1022:11)
>     at WriteWrap.afterWrite (net.js:804:14)
> 2017-06-18T04:58:47.940Z - info: [main] Error report submitted

#### Additional Information

A package.json and README.md file were added in the same commit (unchecking them yields the same result).

cjcool:
I tried adding the files to a commit manually. Git says, 

> Filename too long

shiftkey:
@cjcool thanks for the detailed report! this seems like enough to reproduce the issue but I'll report back if I can't see the same behaviour.

joshaber:
[...]

PR #4027, opened 2018-02-16
report EPIPE and EOF errors betterer

shiftkey:
There's a couple of common-but-obscure errors sitting in Haystack that I'd like to understand more about. 

This one only occurs on macOS:

```
error: write EPIPE
exports._errnoException (util.js:1050:11)
WriteWrap.afterWrite (net.js:813:14)
```

This one only occurs on Windows:

```
error: write EOF
exports._errnoException (util.js:1050:11)
WriteWrap.afterWrite (net.js:813:14)
```

Digging into more details about these errors, I stumble upon this project: https://github.com/mhart/epipebomb

> By default, node throws `EPIPE` errors if `process.stdout` is being written to and a user runs it through a pipe that gets closed while the process is still outputting (eg, the simple case of piping a node app through `head`).
> 
> This seemed a little overzealous to me, so I wrote this to suppress such errors.

I'm not quite sure how this might apply to Desktop, but the details of this suppression was very interesting: they do this by attaching to the `error` event on `process.stdout`.

This is interesting, for a couple of reasons:

 - #3954 was a situation where I did the exact same thing to listen for `error` events from the spawned process (to stop the renderer from crashing)
 - the stack trace for both errors are the same, which makes me think I can handle both errors by handling the `error` event on `process.stdout`

~~There's also `GitProcess` in `dugite` that might be responsible for these errors, and I'll address that separately to this, but~~ this PR should help us identify the source of these errors better than what's currently delivered to Haystack.

### DIFF
--- a/app/src/lib/git/show.ts
+++ b/app/src/lib/git/show.ts
@@ -2,7 +2,7 @@ import { ChildProcess } from 'child_process'
 
 import { git } from './core'
 import { spawnAndComplete } from './spawn'
-import { protectProcessOutput } from '../process'
+import { reportProcessOutputError } from '../process'
 
 import { Repository } from '../../models/repository'
 
@@ -30,7 +30,7 @@ export async function getBlobContents(
   const successExitCodes = new Set([0, 1])
   const setBinaryEncoding: (process: ChildProcess) => void = cb => {
     cb.stdout.setEncoding('binary')
-    protectProcessOutput(cb, 'getBlobContents')
+    reportProcessOutputError(cb, 'getBlobContents')
   }
 
   const args = ['show', `${commitish}:${path}`]
--- a/app/src/lib/git/show.ts
+++ b/app/src/lib/git/show.ts
@@ -2,6 +2,7 @@ import { ChildProcess } from 'child_process'
 
 import { git } from './core'
 import { spawnAndComplete } from './spawn'
+import { protectProcessOutput } from '../process'
 
 import { Repository } from '../../models/repository'
 
@@ -27,8 +28,10 @@ export async function getBlobContents(
   path: string
 ): Promise<Buffer> {
   const successExitCodes = new Set([0, 1])
-  const setBinaryEncoding: (process: ChildProcess) => void = cb =>
+  const setBinaryEncoding: (process: ChildProcess) => void = cb => {
     cb.stdout.setEncoding('binary')
+    protectProcessOutput(cb, 'getBlobContents')
+  }
 
   const args = ['show', `${commitish}:${path}`]
   const opts = {
--- a/app/src/lib/git/spawn.ts
+++ b/app/src/lib/git/spawn.ts
@@ -1,6 +1,6 @@
 import { GitProcess } from 'dugite'
 import * as GitPerf from '../../ui/lib/git-perf'
-import { protectProcessOutput } from '../process'
+import { reportProcessOutputError } from '../process'
 
 type ProcessOutput = {
   output: Buffer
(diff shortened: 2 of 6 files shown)


shiftkey:
> There's also GitProcess in dugite that might be responsible for these errors, and I'll address that separately to this, 

Actually, I'm going to leave that aside for now as I think the `execFile` callback in `dugite` (which handles ENOENT and should bubble up other error codes when we `await` it) should cover us: https://github.com/desktop/dugite/blob/master/lib/git-process.ts#L183

The fact that the error lacks any Desktop source code in the stack trace makes me think this is related to the spawn-based APIs, and we need a similar fix to what we did in #3954.

shiftkey:
@desktop/core I'd like to get this into 1.1 as it's been around for the last couple of versions, and it might help ameliorate the error reports we're seeing. Let me know if you have any feels on this.

nerdneha:
i'm 👍 for some more info, will let others chime in

niik:
I don't believe that stdout is the problem here and I'd be surprised if the logging in this PR picks up anything useful for us (though I have a strong track record of being wrong). To me this looks like an error that I'd expect if I'm piping data to a child process which closes its *stdin*.

Here's a small repro

```js
require('child_process')
  .spawn("head", ["-c", "1"])
  .stdin.end('x'.repeat(1024 * 1024))
```

output

```
process-input-closed markus$ node index.js
events.js:163
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at exports._errnoException (util.js:1050:11)
    at WriteWrap.afterWrite [as oncomplete] (net.js:813:14)
```

The only case that I see stdout writes being the cause of this error is for data that we ourselves (i.e. desktop) are writing to stdout which should only be log messages (for the main app) and since we haven't made any real changes to that in the release where these errors spiked I think we should consider that to be of fairly low probability.

We did, however, added two new git calls that write to git's standard input in the last release, `mergeTrailers` and `parseTrailers`. I strongly suspect that it's one of those two that's the cause, that for some reason `git interpret-trailers` is terminating abruptly but we never see that actual error because node throws the stdin error first and we terminate the app before we can ever see the actual error from git and the exit code.

As far as I'm concerned a child process closing its stdin should not warrant an error. I think we should suppress it and wait for dugite's error logic to pick up on the exit code and parse (if any) stderr/stdout.

This won't solve the underlying problems obviously but our errors should now be thrown in a way that we can handle them and hopefully catch some stacks that allow us to pinpoint which git call is prematurely closing its input and why.

So to summarize I think we should address this in dugite by suppressing the error both for `execFile` and `spawn`. For `spawn` we could conceivably do this in the app ourselves without involving dugite but for `execFile` we can't since the `processCallback` where we'd be able to attach our error handler runs after we flush the stdin data.

Should consumers of dugite want to be notified when stdin closes they can either subscribe to `ChildProcess.stdin.on('close')` or `ChildProcess.stdin.on('error')` and override our suppression but I think this would make for a better default behavior.

I'll open a PR against dugite and we can continue the discussion there.

shiftkey:
> I'll open a PR against dugite and we can continue the discussion there.

@niik I'm into it

shiftkey:
> I don't believe that stdout is the problem here and I'd be surprised if the logging in this PR picks up anything useful for us (though I have a strong track record of being wrong). 

Thinking about this some more, the `write EOF` reports in Haystack (like #2041) have around for a while and seem to be related to committing a large numbers of files, and one of the commands we depend on - `update-index` does use `stdin` to pass these file paths to Git:

https://github.com/desktop/desktop/blob/bcf0b669e29ec21131ed9cd8a4613caac1a01c06/app/src/lib/git/update-index.ts#L90-L94

I'm still annoyed to have not been able to reproduce this behaviour myself, but I'm keen to see what you dig up. I'm also happy to pick this up and get a fix into `dugite` to improve our diagnostics, if you've got enough on your plate next week.

PR #4034, opened 2018-02-17
Add a custom global whitelist

### LINKED ISSUE #3782: Use a custom whitelist of globals allowed in the source
https://github.com/desktop/desktop/pull/3780 happened because `window.name` is a string property of `window`, which is used for <sub>*mumble mumble*</sub>. However, we probably don’t need to use any of the properties of `window` (or even `window` — `global` should be enough).

We could check for this error with ESLint’s `no-undef` rule, and whitelisting globals as needed (or just accessing them under `global`).

j-f1:
Feedback on the list in `.eslintrc.js` is appreciated.

Fixes #3782.

### DIFF
--- a/app/src/lib/source-map-support.ts
+++ b/app/src/lib/source-map-support.ts
@@ -32,7 +32,7 @@ function retrieveSourceMap(source: string) {
 
   if (__DEV__ && path.startsWith('http://')) {
     try {
-      const xhr = new XMLHttpRequest()
+      const xhr = new window.XMLHttpRequest()
       xhr.open('GET', path, false)
       xhr.send(null)
       if (xhr.readyState === 4 && xhr.status === 200) {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,32 @@
+const appInfo = require('./app/app-info')
+
 module.exports = {
   extends: './eslint-config.yml',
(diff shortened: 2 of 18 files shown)


j-f1:
💻

nerdneha:
@j-f1 looks like we still have some failing tests in circleci; are you able to access?

app/src/lib/source-map-support.ts
```diff
@@ -32,7 +32,7 @@ function retrieveSourceMap(source: string) {
 
   if (__DEV__ && path.startsWith('http://')) {
     try {
-      const xhr = new XMLHttpRequest()
+      const xhr = new window.XMLHttpRequest()
```
shiftkey: I'm not quite sure this is the best approach, as these changes now cause the TS compiler to consider these errors. I'm also not across how best to mash tsc and eslint together at this point, so I'm not sure how to proceed...

shiftkey:
Dropping this from the 1.2 milestone as it's not related to feature work.

shiftkey:
I had another shot at trying to get TS to play nicer with the changes here, and couldn't really figure out a way to mesh this together with the ESLint work here.

@j-f1 do you mind if we close this PR and keep #3782 open to track the desired goal?

j-f1:
Sounds fine

Issue #3971, opened 2018-02-08
Diff too big warning/hint not easily visible


tooomm:
### Description

Warning/hint not easily visible and badly placed.
I assume the intention is to have it displayed centered, not left aligned?

Also, it could be bold for example.

(the picture seems to be left aligned, too!)

### Version

**GitHub Desktop version:** 1.0.13
**OS version:** Microsoft Windows [Version 10.0.16299.125]

### Additional information
![warning](https://user-images.githubusercontent.com/9874850/35986599-c1d48bf2-0cf9-11e8-9ea1-205cab8f9cd0.png)

iAmWillShepherd:
cc @desktop/design for thoughts.

niik:
Welp, that doesn't look great at all. Seems like it looked a little bit better when it was introduced in https://github.com/desktop/desktop/pull/1893 but I don't believe this ever received design input. I don't think we should use that graphic here and I'm not wild about the messaging either.

I think something like this would be an improvement

<img src="https://user-images.githubusercontent.com/634063/36023997-a0628c72-0d8e-11e8-9c96-c72616d276e5.png" width="650" />

What do you think @donokuda?

donokuda:
Looks fine to me. 👍  Thoughts on making sure the warning is centered in that area? I'm also unsure if we should use the blue primary button or a white secondary button instead, but I'm pretty impartial with that call.

niik:
> Thoughts on making sure the warning is centered in that area?

Horizontally? Yes, definitely, with a max width of around 400px.
Vertically? I tried it out and I believe top align with double margin is the way to go.

> I'm also unsure if we should use the blue primary button or a white secondary button instead, but I'm pretty impartial with that call.

<img src="https://user-images.githubusercontent.com/634063/36028414-24d13f20-0d9f-11e8-9a9b-20f02cf7376b.png" width="650" />

I think you're right, let's go with the secondary.

tooomm:
I would center such a warning horizontally for sure!
Vertically not neccessarly. Could also be 1/3 to the top etc.

About the buttons I can't help too much as I don't know about the underlying concept exactly.

perperson1:
![multixnxx-all things gay -7](https://user-images.githubusercontent.com/36304871/36036288-ca6dc858-0db9-11e8-94c7-e31e5f9f69b7.jpg)

PR #4050, opened 2018-02-20
Makes the diff too large message prettier

iAmWillShepherd:
Fixes issue #3971

In addition to fixing up the diff too large message on the History tab, I've updated the layout of the Changes tab to be consistent with the History tab.

* if the buffer is >= 268MB, display image only
* if the buffer >= 4MB or diff  is >= 500k lines, display image and "show diff" button that can be clicked on to show the diff

#### Large Diff
1. Initialize a new repository
1. Create a file that is at least 4MB in the repo
1. Diff should state this diff is large and give you the option to view
1. Commit the file
1. Go to the History tab and try to view the diff
1. Diff should state this diff is large and give you the option to view
#### Not able to render
1. Initialize a new repository
1. Create a file that is at least 270MB in the repo
1. Diff should state this diff cannot be rendered
1. Commit file
1. Go to History tab and try to view the diff
1. Diff should state this diff cannot be rendered

Note that it may take a while for the diff to appear, #4052 addresses that issue.


app/src/ui/diff/index.tsx
```diff
@@ -952,6 +974,65 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+    return (
+      <div className="panel empty large-diff">
+        <img src={NoDiffImage} />
+        <p>The diff is too large to be displayed.</p>
+      </div>
+    )
+  }
+
+  private renderLargeText(diff: ILargeTextDiff) {
+    if (diff.hunks == null || diff.text == null) {
```
nerdneha: we do this check in the switch statement in `renderDiff` right before we call this method. if there's any concern then we can just have all `DiffType.LargeText` call the `this.renderLargeText` method (or something similarly titled) and branch inside.

app/src/ui/diff/index.tsx
```diff
@@ -952,6 +974,65 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+      return fatalError('Cannot render diff withouth hunks and text')
```
nerdneha: i'm not sure this fatalError will be needed if we follow my above suggestion but in case we keep it, typo: s/withouth/without

app/src/lib/git/diff.ts
```diff
@@ -365,6 +362,52 @@ function diffFromRawDiffOutput(output: Buffer): IRawDiff {
   return parser.parse(pieces[pieces.length - 1])
 }
 
+/**
+ * Utility function that takes a buffer and outputs
+ * either a largeTextDiff
```
nerdneha: looks like this comment was accidentally incomplete

app/src/ui/diff/index.tsx
```diff
@@ -1044,52 +1125,31 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
       }
-
-      return this.renderTextDiff(diff)
+      default:
+        return null
     }
+  }
 
-    return null
+  private onShowLargeDiff = () => {
```
nerdneha: i'm not sure i'm seeing much consistency in the codebase, in fairness, but i think if we're creating a private method, the method would ideally be `showLargeDiff` (and the "onShowLargeDiffs" are the methods that are being passed in as props from another component)

nerdneha:
pulling in @niik and @donokuda here as there's a new button to show a diff if we can make it happen (i think). i'm 👍 for it

nerdneha:
🐝

app/src/lib/git/diff.ts
```diff
@@ -110,14 +111,13 @@ export async function getCommitDiff(
     repository.path,
     'getCommitDiff'
   )
-  if (isBufferTooLarge(output)) {
-    return { kind: DiffType.TooLarge, length: output.length }
+
+  const largeDiff = buildLargeTextDiff(output)
+  if (largeDiff != null) {
+    return largeDiff
```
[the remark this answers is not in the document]
iAmWillShepherd: There are 3 conditions that must be met for an `ILargeTextDiff` to be returned from `buildLargeTextDiff`. If none of the following are met, we return `null` because the passed in buffer doesn't meet the conditions to be an `ILargTextDiff` object.
1. `!didReadAllBytes || !isValidBuffer(buffer)`
1. `isBufferTooLarge(buffer)`
1. `isDiffTooLarge(diffText)`

> ...maybe bufferTooLargeCommitDiff, diffTooLargeCommitDiff, invalidBufferCommitDiff since the information we provide for each commit diff is unique to the error case but we don't care what's inside much beyond that.

The reason why the name is not _[noun]TooLarge_ or _invalid[noun]_ is because the words `too` and `invalid` imply, at least to me, that there is no further work we can do with the named object. By omitting any indication that something is too large or invalid allows me to reason more clearly about _how_ this object is to be used: as a way to defer opening a diff that could degrade the UX. 

> I prefer the isBufferTooLarge/etc. boolean checks that were there before...Are we switching to the buildLargeTextDiff method because we want to switch over to doing all checks in every case (buffer invalid, buffer too large, diff too large)?

The reason the function `buildLargeTextDiff` was extracted is because we were duplicating logic in the two functions `getCommitDiff` and `getWorkingDirectoryDiff`. 

> ...ex: maybe bufferTooLargeCommitDiff, diffTooLargeCommitDiff, invalidBufferCommitDiff...

Not sure I'm clear on this. Are you implying that we creating different interfaces for the three different conditions we check for?

If so would something like the following satisfy that requirement?
`type LargeTextDiff = 'bufferTooLarge' | 'diffTooLarge' | ILargeTextDiff`
Doing this would allow us to distinguish between a `ILargeTextDiff` that can be rendered and one that cannot (we can currently do this now, but by inspecting properties).

app/src/ui/diff/index.tsx
iAmWillShepherd: There is no concern, just a design decision. 

Since `hunks` and `text` are optional properties on `ILargeTextDiff` and `ITextDiff` does not allow for optional values, I have to prove to the TSC that those properties are not `null`. The two options I had was to use a conditional statement or to use the `!` operator and a comment e.g. `diff.hunks!  \\guaranteed to be set at this point`. I chose the former because you'd have to look for that condition in another method and I felt it was better off inside the method.

shiftkey:
Needs a merge from `master` to unblock the CircleCI build

iAmWillShepherd:
💯

app/src/lib/git/diff.ts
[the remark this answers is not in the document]
nerdneha: > If none of the following are met, we return null because the passed in buffer doesn't meet the conditions to be an ILargTextDiff object.

Once you read the implementation of `buildLargeTextDiff`, it's clear we are checking against 3 test conditions and maybe returning the text diff object. I don't think returning an `object|null` is as intuitive as exposing one or two boolean checks UNLESS we need to check for all 3 cases in both `getCommitDiff` and `getWorkingDirectoryDiff` functions in the exact same way.

> The reason the function buildLargeTextDiff was extracted is because we were duplicating logic in the two functions getCommitDiff and getWorkingDirectoryDiff.

In this case, I'm not sure it's worth it since we have to do a check against null now in both places instead. The tradeoff is between clarity and duplication.

> Are you implying that we creating different interfaces for the three different conditions we check for?

more like we bring back the boolean checks but have a helper method that defines the output object. ex:
```
if (isDiffTooLarge(diffText)) {	
  return bufferTooLargeCommitDiff()	
}
```
we could even provide better boolean checks (re: your comment of `too` being unclear as a descriptor), ex: s/`isDiffTooLarge`/`diffExceedsMaxLineLength`

app/src/ui/diff/index.tsx
```diff
@@ -1044,52 +1122,31 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
-
-        return <div className="panel empty">No content changes found</div>
+        return this.state.forceShowLargeDiff
+          ? this.renderLargeText(largeTextDiff)
+          : this.renderLargeTextDiff()
       }
-
-      return this.renderTextDiff(diff)
+      default:
+        return null
```
nerdneha: looks like we usually put an `assertNever` here with an error message, see:
https://github.com/desktop/desktop/blob/d5c4b4ad9483670aeec4d0974f4aba1c82e748c6/app/src/ui/branches/ci-status.tsx#L60

app/src/ui/diff/index.tsx
```diff
@@ -952,6 +974,62 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+        <Button onClick={this.showLargeDiff}>Show diff</Button>
+      </div>
+    )
+  }
+
+  private renderDiffTooLarge() {
+    return (
+      <div className="panel empty large-diff">
+        <img src={NoDiffImage} />
+        <p>The diff is too large to be displayed.</p>
```
nerdneha: Could you help me understand why we removed the helper text: `The diff returned by Git is {diffSizeMB}MB ({diff.length} bytes), which is larger than what can be displayed in GitHub Desktop.` I thought it was a nice touch. Maybe even helping people be aware of what the limit is since it's a const.

iAmWillShepherd: I asked @desktop/design for their input and they produced some awesome mocks for me to work from.

app/src/ui/diff/index.tsx
```diff
@@ -1044,52 +1122,31 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+      case DiffType.Text:
+        return this.renderText(diff)
+      case DiffType.Binary:
+        return this.renderBinaryFile()
+      case DiffType.Image:
+        return this.renderImage(diff)
+      case DiffType.LargeText: {
+        const largeTextDiff = diff as ILargeTextDiff
+
+        if (largeTextDiff.hunks == null || largeTextDiff.text == null) {
```
nerdneha: the only type we would end up in this case is if `!isValidBuffer(buffer) === true` right? bufferTooLarge and diffTooLarge both have hunks and text from `diffFromRawDiffOutput(buffer)` of type `IRawDiff` which has hunks and contents. 

If that's true, would it make sense to add a DiffType.InvalidBuffer as a separate case in the switch statement then to result in `this.renderDiffTooLarge()`?

app/src/ui/diff/index.tsx
```diff
@@ -952,6 +974,62 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+    return (
+      <div className="panel empty large-diff">
+        <img src={NoDiffImage} />
+        <p>The diff is too large to be displayed.</p>
+      </div>
+    )
+  }
+
+  private renderLargeText(diff: ILargeTextDiff) {
+    // gauranteed to be set since this function won't be called if text or hunks are null
```
nerdneha: s/gauranteed/guaranteed

app/src/lib/git/diff.ts
```diff
@@ -365,6 +362,52 @@ function diffFromRawDiffOutput(output: Buffer): IRawDiff {
...
+
+  const diffText = diffFromRawDiffOutput(buffer)
+
+  if (isBufferTooLarge(buffer)) {
+    // we don't want to render by default
+    // but we keep it as an option by
+    // passing in text and hunks
+    const largeTextDiff: ILargeTextDiff = {
+      kind: DiffType.LargeText,
+      length: buffer.length,
```
nerdneha: if we remove the length in the helper text in the component, we don't need this any more and we can combine this with the one below, right? am i missing where it's being used?

nerdneha:
Philosophical Debate -- in the name of our objective to `Minimize the scary bits`, do we want to only provide the `Show diff` button to a certain point and after that we hide it because it'll freeze the app? Or is the warning message enough? Thoughts?

app/src/ui/diff/index.tsx
donokuda: I don't find the actual size of the diff helpful other than that it being a very large number. On top of that, it's information that our users don't necessarily have control over (e.g., they can't reduce the size of the diff itself, as far as I know.)

I'm more in favor of the terse "this is the situation right now. You can choose to bypass this, but this effect might happen as a result" type of messaging.

app/src/ui/diff/index.tsx
```diff
@@ -952,6 +974,62 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+    return (
+      <div className="panel empty large-diff">
+        <img src={NoDiffImage} />
+        <p>
+          The diff is too large to be displayed by default.
+          <br />
+          You can try to show it anyways, but performance may be negatively
+          impacted.
+        </p>
+        <Button onClick={this.showLargeDiff}>Show diff</Button>
```
nerdneha: We should probably case this, ex:
`{__DARWIN__ ? 'Untrusted Server' : 'Untrusted server'}`

donokuda:
> Philosophical Debate -- in the name of our objective to Minimize the scary bits, do we want to only provide the Show diff button to a certain point and after that we hide it because it'll freeze the app? Or is the warning message enough? Thoughts?

I think the warning is enough because I'm not exactly sure if/when a large diff will absolutely crash the app.

app/src/ui/diff/index.tsx
iAmWillShepherd: Yea, I like that because it's more explicit.

iAmWillShepherd:
🔢

app/src/models/diff.ts
```diff
@@ -118,18 +120,27 @@ export interface IBinaryDiff {
...
-  /**
-   * The length of the diff output from Git which exceeds the runtime limits:
-   *
-   * 268435441 bytes = 256MB - 15 bytes
-   */
-  readonly length: number
+export interface ILargeTextDiff {
+  readonly kind: DiffType.LargeText
+  /** The unified text diff - including headers and context */
+  readonly text?: string
```
nerdneha: does the `text` value need to be optional? looks like the only time we call it, it is set

app/src/models/diff.ts
```diff
@@ -118,18 +120,27 @@ export interface IBinaryDiff {
...
+  /** The diff contents organized by hunk - how the git CLI outputs to the caller */
+  readonly hunks?: ReadonlyArray<DiffHunk>
```
nerdneha: does the `hunks` value need to be optional? looks like the only time we call it, it is set

app/src/ui/diff/index.tsx
```diff
@@ -1044,52 +1124,29 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+      case DiffType.Unrenderable:
+        return this.renderUnrenderableDiff()
+      default:
+        return assertNever(diff, `Unsupported diff type: ${diff}`)
     }
+  }
 
-    return null
+  private showLargeDiff = () => {
+    this.setState({ forceShowLargeDiff: !this.state.forceShowLargeDiff })
```
nerdneha: if we click on this button, we just show it, right? is there a hide that makes this more of a toggle? should it just be 
```
this.setState({ forceShowLargeDiff: true })
```
? or should we offer to hide it later

app/src/ui/diff/index.tsx
```diff
@@ -1044,52 +1124,29 @@ export class Diff extends React.Component<IDiffProps, {}> {
...
+  private renderDiff(diff: IDiff): JSX.Element | null {
+    switch (diff.kind) {
+      case DiffType.Text:
+        return this.renderText(diff)
+      case DiffType.Binary:
+        return this.renderBinaryFile()
+      case DiffType.Image:
+        return this.renderImage(diff)
+      case DiffType.LargeText: {
+        const largeTextDiff = diff as ILargeTextDiff
```
nerdneha: this is actually an open question:
The above methods that take `diff`, ex: `this.renderText(diff)` don't need to cast diff as `ITextDiff` because the method `renderText` checks that diff is `ITextDiff` in its input. Accordingly, is there a special need to cast diff as `ILargeTextDiff`? it looks like the `renderLargeText` method expects that type as its input anyway.

If we already talked about this, I'm sorry! This looks vaguely familiar.

nerdneha:
🐭

app/src/lib/git/diff.ts
```diff
@@ -83,7 +83,7 @@ const imageFileExtensions = new Set(['.png', '.jpg', '.jpeg', '.gif', '.ico'])
  * Render the difference between a file in the given commit and its parent
  *
  * @param commitish A commit SHA or some other identifier that ultimately dereferences
- *                  to a commit.
+ *                  a commit.
```
j-f1: I think the original is correct. The commitish dereferences **to** something. It does not dereference something.

nerdneha: you're totally right @j-f1 thanks for pointing that out

app/src/models/diff.ts
iAmWillShepherd: Oh you're right! That's a holdover to the old way of doing things .

app/src/ui/diff/index.tsx
iAmWillShepherd: No, we don't have a toggle. Using `true` seems much better.

app/src/ui/diff/index.tsx
iAmWillShepherd: This was a leftover from refactoring `ILargeTextDiff` to not have multiple purposes. Now that we have `IUnrenderableDiff` we don't have to worry about casting any longer.

iAmWillShepherd:
🔢

iAmWillShepherd:
@nerdneha I've added reproduction steps in the PR body.

app/src/lib/git/diff.ts
```diff
@@ -30,13 +32,13 @@ import { DiffParser } from '../diff-parser'
  * This is a hard limit on how big a buffer can be and still be converted into
  * a string.
  */
-const MaxDiffBufferSize = 268435441
+const MaxDiffBufferSize = 268435441 //~268MB
```
nerdneha: This number represents the limit where we no longer show the diff even if the user insists. 

I tried to render a diff for a file that was ~70MB (it showed the initial screen and then clicked on the show diff button) and it cut off after some point (image)
screen shot 2018-03-06 at 12 16 50 am

Maybe we need to bring MaxDiffBufferSize to something <=70MB. It rendered a 62MB file fine.

nerdneha:
everything else works correctly. i tested:
* new file < 4MB -- renders normally
* new file @ 62MB ( > 4MB && <268MB ) -- renders with a warning but ability to `show diff`
* existing file @ 62MB ( > 4MB && <268MB ) -- renders with a warning but ability to `show diff`
* new file > 70MB -- renders with a warning but ability to `show diff` **but** does not render the full diff (see comment above)
* new file with 100k characters -- renders normally
* new file with 690k characters -- renders with a warning but ability to `show diff`

⚙️⚙️⚙️⚙️

iAmWillShepherd:
🔢

app/src/lib/git/diff.ts
```diff
@@ -26,23 +26,23 @@ import { spawnAndComplete } from './spawn'
...
  * Where `MaxDiffBufferSize` is a hard limit, this is a suggested limit. Diffs
  * bigger than this _could_ be displayed but it might cause some slowness.
  */
-const MaxReasonableDiffSize = MaxDiffBufferSize / 64 //~4MB
+const MaxReasonableDiffSize = MaxDiffBufferSize / 16 // ~4.375MB in decimal
 
 /**
  * The longest line length we should try to display. If a diff has a line longer
- * than this, we probably shouldn't attempt it.
+ * tha obably shouldn't attempt it.
```
j-f1: Did you mean to change this?

iAmWillShepherd: No, I did not...

PR #3980, opened 2018-02-09
Make diff too large screen better

iAmWillShepherd:
Fixes issue #3971

PR #4052, opened 2018-02-20
Give up and assume diff is too large

iAmWillShepherd:
*Please make sure this is merged before PR #4050, so that I can fix merge conflicts in that branch*

I noticed this issue while working on PR #3980, which is now closed to make code review easier. Shoutout to @shiftkey for letting me know that `spawn` can be given the maximum number of bytes to read.

Notice how long it takes the UI to catch up and let me know the diff can't be displayed. This changes means that we give up sooner and assume a diff is too large so the UX can be a bit better.

### Before
before

### After
after

### DIFF
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -200,13 +200,14 @@ export async function getWorkingDirectoryDiff(
     ]
   }
 
-  const { output, error } = await spawnAndComplete(
+  const { output, error, didReadAllBytes } = await spawnAndComplete(
     args,
     repository.path,
     'getWorkingDirectoryDiff',
-    successExitCodes
+    successExitCodes,
+    MaxBytesToRead
   )
-  if (isBufferTooLarge(output)) {
+  if (!didReadAllBytes || isBufferTooLarge(output)) {
     // we know we can't transform this process output into a diff, so let's
     // just return a placeholder for now that we can display to the user
     // to say we're at the limits of the runtime
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -113,12 +113,14 @@ export async function getCommitDiff(
     args.push(file.oldPath)
   }
 
-  const { output } = await spawnAndComplete(
+  const { output, didReadAllBytes } = await spawnAndComplete(
     args,
     repository.path,
-    'getCommitDiff'
+    'getCommitDiff',
+    undefined,
+    MaxBytesToRead
   )
-  if (isBufferTooLarge(output)) {
+  if (!didReadAllBytes || isBufferTooLarge(output)) {
     return { kind: DiffType.TooLarge, length: output.length }
   }
 
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -39,6 +39,14 @@ const MaxDiffBufferSize = 268435441
 const MaxReasonableDiffSize = 3000000
 
 /**
+ * Where `MaxReasonableDiffSize` is a soft limit, and `MaxDiffBufferSize`
+ * is an absolute limit, this is the MAX number of bytes to read from the
+ * buffer before _assuming_ the current buffer being read is `MaxDiffBufferSize`.
+ * This is done so that the UI isn't waiting for the entire buffer to be read.
+ */
+const MaxBytesToRead = MaxDiffBufferSize / 8 //~32MB
+
+/**
  * The longest line length we should try to display. If a diff has a line longer
  * than this, we probably shouldn't attempt it.
  */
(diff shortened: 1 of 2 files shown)


tierninho:
✅  tested for UI and functionality

bluebunny4:
hangout?

app/src/lib/git/diff.ts
```diff
@@ -38,6 +38,14 @@ const MaxDiffBufferSize = 268435441
  */
 const MaxReasonableDiffSize = 3000000
 
+/**
+ * Where `MaxReasonableDiffSize` is a soft limit, and `MaxDiffBufferSize`
+ * is an absolute limit, this is the MAX number of bytes to read from the
+ * buffer before _assuming_ the current buffer being read is `MaxDiffBufferSize`.
+ * This is done so that the UI isn't waiting for the entire buffer to be read.
+ */
+const MaxBytesToRead = MaxDiffBufferSize / 8 //~32MB
```
niik: Couldn't we just use MaxReasonableDiffSize instead? Do we care about that small zone in between `MaxBytesToRead` and `MaxReasonableDiffSize`?

app/src/lib/git/diff.ts
```diff
@@ -105,12 +113,14 @@ export async function getCommitDiff(
...
+  const { output, didReadAllBytes } = await spawnAndComplete(
     args,
     repository.path,
-    'getCommitDiff'
+    'getCommitDiff',
+    undefined,
+    MaxBytesToRead
   )
-  if (isBufferTooLarge(output)) {
+  if (!didReadAllBytes || isBufferTooLarge(output)) {
```
niik: Following on from my comment about using `MaxReasonableDiffSize`, could we use `output.length >= MaxReasonableDiffSize` here and avoid introducing a property on the return type here? We could extract that logic into a method similar to `isBufferToLarge` i.e. `isBufferReasonable` and then once we get to the point where we want to be able to tell this method whether to load the diff anyway we can elect to not care about the soft limit.

app/src/lib/git/diff.ts
iAmWillShepherd: If we used `MaxReasonableDiffSize` instead, that would lead to very relatively small diffs being classified as too large (~3MB).  Having a separate value allows us to classify the diff as potentially too large with some wiggle room (up to 32MB) to display it.

app/src/lib/git/diff.ts
iAmWillShepherd: We're doing that already with `isBufferTooLarge`. Getting the length of the buffer will not tell us that `spawnAndComplete` exited before reading the entire buffer, so we need that property. The alternative is to pass a callback to `spawnAndComplete` that get's fired if it doesn't complete reading a buffer.

app/src/lib/git/diff.ts
iAmWillShepherd: I also want to point out that this is being used over in #4050 and I'm using each of these values to determine when to show a diff and when to mark it as not renderable.

app/src/lib/git/diff.ts
niik: > Getting the length of the buffer will not tell us that spawnAndComplete exited before reading the entire buffer

Do we care though? If the length is greater than or equal to the `MaxReasonableDiffSize` we shouldn't be displaying the diff unless the user explicitly asks us to. I might be missing something but it seems to me the only case where we couldn't determine whether `spawnAndComplete` exited before reading all of it is when the output is exactly as large as `MaxReasonableDiffSize` and at that point does it even matter?

app/src/lib/git/diff.ts
niik: > If we used MaxReasonableDiffSize instead, that would lead to very relatively small diffs being classified as too large (~3MB). Having a separate value allows us to classify the diff as potentially too large with some wiggle room (up to 32MB) to display it.

What's that middle ground state giving us? The way I look at it there's three states of a diff from the perspective of the app. There's a diff that we'll always show (<3mb), there's diffs that we won't show by default but the user can explicitly tell us to load (>=3mb) and then there's diffs which we will never be able to represent due to limitations in node (MaxDiffBufferSize).

So there's a soft limit and a hard limit. When we go beyond the soft limit we should exit quickly in this method and show the user that we believe the diff to be unreasonably large. If they choose to try loading it anyway we should call this method again and tell it not to care about the soft limit. What am I missing?

iAmWillShepherd:
So, it looks like we won't need this PR anymore since we limit diff sizes to 70MBs due to larger diffs not rendering completely (see [comment](https://github.com/desktop/desktop/pull/4050#issuecomment-370606540)). Before, the max diff size was around 256MB and waiting up to 10 seconds just for the UI to update and show the diff was too large to render was unacceptable; however, at 70MB, the max time to wait is around 1-2 seconds **max**. Going to close this out since the delay is no longer an issue.

PR #4070, opened 2018-02-22
bump to beta3

shiftkey:
This one gets the fix from #4057 out to beta users.


shiftkey:
Tested the update is applied on macOS and Windows :tada:

PR #3902, opened 2018-01-31
Update to TypeScript 2.7.1

iAmWillShepherd:
[TypeScript 2.7](https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/) is here!!! 

- [x] Bump version
- [x] Ensure everything builds
- [x] Wait for `typescript-eslint-parser` to update with support for this version of TS
- [x] Wait for `prettier` 1.11 to support the new `!` definite assignment assertions suffix

```typescript
let x!: number[];
initialize();
x.push(4);

function initialize() {
    x = [0, 1, 2, 3];
}
```

- [x] remove `strictPropertyInitialization: false` and review breakages for anything that can be initialized correctly
- [ ] Make sure [the new behavior of `instanceof` and `in` don't bite us](https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/#in-operator-narrowing-and-accurate-instanceof)

PR #4077, opened 2018-02-22
Removes ability to get the underlying DOM element from a TextBox component

iAmWillShepherd:
This issue came up while chatting with @niik about PR #3972 (see [comment](https://github.com/desktop/desktop/pull/3972#issuecomment-367764382)).

### DIFF
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -153,7 +153,9 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
   }
 
   public componentDidMount() {
-    this.filterTextBox && this.filterTextBox.selectAll()
+    if (this.filterTextBox != null) {
+      this.filterTextBox.selectAll()
+    }
   }
 
   public render() {
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -159,6 +159,7 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
 
         <Row className="filter-field-row">
           <TextBox
+            ref={this.filterTextBoxRef}
             type="search"
             autoFocus={true}
             placeholder="Filter"
@@ -175,6 +176,10 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
         <div className="filter-list-container">{this.renderContent()}</div>
       </div>
     )
+  }
+
+  private filterTextBoxRef = (component: TextBox) => {
+    this.filterTextBox = component
   }
 
   private renderContent() {
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -144,7 +144,7 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
   IFilterListState<T>
 > {
   private list: List | null = null
-  private filterInput: HTMLInputElement | null = null
+  private filterTextBox: TextBox | null = null
 
   public constructor(props: IFilterListProps<T>) {
     super(props)
@@ -309,7 +309,7 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
       const input = this.filterInput
       if (input) {
         event.preventDefault()
-        input.focus()
+        textBox.focus()
       }
     }
   }
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -165,12 +165,12 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
 
         <Row className="filter-field-row">
           <TextBox
-            ref={this.filterTextBoxRef}
+            ref={this.onTextBoxRef}
             type="search"
             autoFocus={true}
             placeholder="Filter"
             className="filter-list-filter-field"
-            onChange={this.onFilterChanged}
+            onValueChanged={this.onFilterChanged}
             onKeyDown={this.onKeyDown}
             value={this.props.filterText}
             disabled={this.props.disabled}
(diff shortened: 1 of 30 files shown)


app/src/ui/lib/text-box.tsx
```diff
@@ -116,6 +115,24 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
     }
   }
 
+  public selectAll() {
+    if (this.inputElement == null) {
```
niik: `===`?

app/src/ui/lib/text-box.tsx
```diff
@@ -116,6 +115,24 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
...
+      return
+    }
+
+    if (this.state.value && this.state.value.length !== 0) {
+      this.inputElement.select()
+    }
+  }
+
+  public focus() {
+    if (this.inputElement == null) {
```
niik: `===`?

app/src/ui/lib/text-box.tsx
```diff
@@ -132,10 +149,8 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
     })
   }
 
-  private onRef = (instance: HTMLInputElement | null) => {
-    if (this.props.onInputRef) {
-      this.props.onInputRef(instance)
-    }
+  private inputRef = (element: HTMLInputElement) => {
```
niik: I think this should be `element: HTMLInputElement | null` right?

app/src/ui/lib/text-box.tsx
```diff
@@ -116,6 +115,24 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
     }
   }
 
+  public selectAll() {
```
niik: Could we document this?

app/src/ui/lib/text-box.tsx
```diff
@@ -116,6 +115,24 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
...
+    if (this.inputElement == null) {
+      return
+    }
+
+    if (this.state.value && this.state.value.length !== 0) {
+      this.inputElement.select()
+    }
+  }
+
+  public focus() {
```
niik: Could we document this?

app/src/ui/lib/filter-list.tsx
```diff
@@ -178,6 +184,10 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
     )
   }
 
+  private filterTextBoxRef = (component: TextBox) => {
```
niik: `component: TextBox | null`

app/src/ui/lib/text-box.tsx
j-f1: `==` checks `=== null || === undefined`.

iAmWillShepherd: We want to check for `null` and `undefined` here. If we did strict equality here, we could try to call `.select()` on `undefined`, and that can be real bad. Using `==` allows us to check for both without the much longer `x === null || x === undefined`.

app/src/ui/lib/text-box.tsx
iAmWillShepherd: Same explanation as above.

app/src/ui/lib/text-box.tsx
niik: I don't see how `inputElement` could ever be undefined though? It's typed as `HTMLInputElement | null`. In what scenario are we expecting it to be `undefined`? If there's such a scenario we should type it accordingly

app/src/ui/lib/text-box.tsx
iAmWillShepherd: Yes, according to their docs, `null` is passed when the component unmounts. Good catch!

app/src/ui/lib/text-box.tsx
iAmWillShepherd: Actually, it will never be undefined because I initialized it with a value of `null`. So yea, we should use the stronger check here. 

For fun, here's a program where I have told TS that a variable cannot be undefined, yet somehow, the variable is undefined. Maybe it's an opportunity for the TS guys to prevent a variable that can't be `undefined` to be declared without an assignment.

```ts
let cantBeUndefined: number | null; //maybe TS should say no here???

cantBeUndefined //undefined
```

niik: @iAmWillShepherd Defining the variable without assigning it is okay in that scenario but attempting to use it before assigning it should yield an error

screen shot 2018-02-27 at 18 21 13

For the property TypeScript recently added tougher checks for unassigned properties but we'll have to wait on your upgrade PR (#3902) to make use of them.

iAmWillShepherd:
💯

app/src/ui/lib/text-box.tsx
iAmWillShepherd: That's just the thing tho, I didn't produce an error for me.



I'm using 2.6.2 in the REPL I produced this on. Maybe VS Cdoe isn't surfacing the error somehow???

app/src/ui/lib/text-box.tsx
```diff
@@ -116,6 +115,26 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
...
   }
 
+  /** Selects all the text in this component */
+  public selectAll() {
+    if (this.inputElement === null) {
+      return
+    }
+
+    if (this.state.value && this.state.value.length !== 0) {
+      this.inputElement.select()
```
niik: Is there any harm in always calling select? Could we simplify this method into

```ts
public selectAll() {
   if (this.inputElement !== null) {
     this.inputElement.select()
   }
}

iAmWillShepherd:
🔢

PR #4080, opened 2018-02-22
Use values instead of events

iAmWillShepherd:
This PR removes `onChange` from props and updates all usages of that prop to use the `onValueChange` callback instead. 

This was done to limit implementation details from leaking out of the component. See this [comment](https://github.com/desktop/desktop/pull/3972#issuecomment-367764382) for more context.

cc/ @niik

### DIFF
--- a/app/src/ui/lib/configure-git-user.tsx
+++ b/app/src/ui/lib/configure-git-user.tsx
@@ -100,14 +100,14 @@ export class ConfigureGitUser extends React.Component<
             label="Name"
             placeholder="Your Name"
             value={this.state.name}
-            onChange={this.onNameChange}
+            onValueChanged={this.onNameChange}
           />
 
           <TextBox
             label="Email"
             placeholder="your-email@example.com"
             value={this.state.email}
-            onChange={this.onEmailChange}
+            onValueChanged={this.onEmailChange}
           />
 
           <Row>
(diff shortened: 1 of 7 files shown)


app/src/ui/lib/text-box.tsx
```diff
@@ -118,11 +115,6 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
 
   private onChange = (event: React.FormEvent<HTMLInputElement>) => {
     const value = event.currentTarget.value
-
-    if (this.props.onChange) {
-      this.props.onChange(event)
-    }
-
     const defaultPrevented = event.defaultPrevented
```
shiftkey: I think I'm into this change, especially if we can't think of a scenario where we want to block the text changing, but this line here should also be :fire:'d (and maybe some others, I haven't read this whole function closely....

iAmWillShepherd: Oh yes, you're right! ⚡️

app/src/ui/lib/configure-git-user.tsx
```diff
@@ -131,16 +131,15 @@ export class ConfigureGitUser extends React.Component<
     )
   }
 
-  private onNameChange = (event: React.FormEvent<HTMLInputElement>) => {
+  private onNameChange = (name: string) => {
     this.setState({
-      name: event.currentTarget.value,
+      name,
       email: this.state.email,
```
niik: I think we can drop email and avatarUrl here and just update name. I realize that this is ancient code but it's a nice cleanup.

iAmWillShepherd:
🔢

PR #4081, opened 2018-02-22
Don't let changes sidebar get too small

iAmWillShepherd:
Adds a minimum width (up from the default of `150px`) for `RepositoryView`.

### Before
before

### After
after

I'd love to know what @desktop/design thoughts are. I felt the default of `150px` was a bit too cramped. Is setting a min width of `225px` too much?

### DIFF
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -22,6 +22,9 @@ import { Account } from '../models/account'
 /** The widest the sidebar can be with the minimum window size. */
 const MaxSidebarWidth = 495
 
+/** The narrowist the sidebar can be with the minumum window size. */
+const MinSideBarWidth = 250
+
 interface IRepositoryProps {
   readonly repository: Repo
   readonly state: IRepositoryModelState
@@ -149,6 +152,7 @@ export class RepositoryView extends React.Component<IRepositoryProps, {}> {
         onReset={this.handleSidebarWidthReset}
         onResize={this.handleSidebarResize}
         maximumWidth={MaxSidebarWidth}
+        minimumWidth={MinSideBarWidth}
       >
         {this.renderTabs()}
         {this.renderSidebarContents()}
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -23,7 +23,7 @@ import { Account } from '../models/account'
 const MaxSidebarWidth = 495
 
 /** The narrowist the sidebar can be with the minumum window size. */
-const MinSideBarWidth = 250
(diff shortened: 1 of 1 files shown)


agisilaos:
That's reasonable. @iAmWillShepherd I'm thinking though what if the path of the file is bigger than the min width of `225px`. It will look simillar like that (a bit shrinked): 

screen shot 2018-02-23 at 13 30 52

Note: 
Take that screenshot as an example of what it's gonna look like in the `Changes` tab

niik:
> I'd love to know what @desktop/design thoughts are. I felt the default of 150px was a bit too cramped. Is setting a min width of 225px too much?

It is cramped but I don't think we should be restrictive here. We initially had a [minimum width of 200px](https://github.com/desktop/desktop/commit/f2c37777acbec8757dee55e80b022976431c7a75) but I decided to decrease it to 150px.

It's not necessarily about whether the sidebar looks good in 150px but rather about making more room for the diff temporarily. I think the question we should be asking ourselves when we look at resizable widths is the minimum width where all components can still be used. In this case the limiting factor is whether we can fit the 'Changes' and 'History' tabs in the width needed and we can in 150px.

So in summary I don't think we should increase the minimum width until we have components in the sidebar which become unusable at the minimum width.

j-f1:
What about “snapping” it to 225px if you resize it to somewhere between 220px and 230px? This would allow us to suggest a minimum width, while still allowing you to make it narrower if you need to.

niik:
> What about “snapping” it to 225px if you resize it to somewhere between 220px and 230px? This would allow us to suggest a minimum width, while still allowing you to make it narrower if you need to.

You can get back to the default width by double clicking on the resizable. I don't necessarily have anything against snapping as well other than that I don't think there's enough precedence for that kind of interaction. I suspect most people will leave the sidebar at its starting position which is what we've determined to be a good look for the application. Ultimately I'd want to work from a clear use case and example of this being an actual pain point before we look at solutions.

iAmWillShepherd:
> So in summary I don't think we should increase the minimum width until we have components in the sidebar which become unusable at the minimum width.

Fair enough 👍

iAmWillShepherd:
> I'm thinking though what if the path of the file is bigger than the min width of 225px. It will look simillar like that (a bit shrinked):

@agisilaos It would likely still looked cramped, but to a lesser degree which is what I was going for really. one of the things that actually spurred me to make this change is my commit messages typically are more than 3 words, so I figured why not make it wide enough for at least 6 words?! 😆

agisilaos:
@iAmWillShepherd So damn true, same here! 😆

PR #4082, opened 2018-02-22
tidy up some unnecessary braces

garritfra:


shiftkey:
@garritfra we're using Prettier to formatt the code, and I think these changes will clash with that

j-f1:
@shiftkey Some of these changes remove `{}` from arrow functions, which [Prettier explicitly doesn’t modify](https://prettier.io/docs/en/rationale.html#what-prettier-is-not-concerned-about). This PR might be valuable if you prefer to avoid braces on single-line functions.

shiftkey:
@j-f1 fair point - I'll reopen this and see if @garritfra wants to address the CI issues (which are around whitespacing and should be addressable with a `yarn lint:fix`

garritfra:
I'll make sure to do that after work

garritfra:
done, need review👀

PR #4090, opened 2018-02-23
the remote's URL can also be the GitHub htmlURL

mathieudutour:
I have some instance of local repositories that have their remote set to the `htmlURL` of the GitHub repo (eg. without the `.git` at the end). Git and GitHub don't seem to mind so we should probably support it as well


shiftkey:
@mathieudutour @j-f1 thanks for calling this out. While this is the simplest fix, I wanted to capture more context around the change - so I extracted it out and added some tests.

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +94,22 @@ export function matchExistingRepository(
     }) || null
   )
 }
+
+/**
+ * Check a GitHub repository matches a given remote.
```
niik: Either `Check if...` or `Check whether or not a...`

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +94,22 @@ export function matchExistingRepository(
...
+ * @param gitHubRepository the GitHub repository details found from the GitHub API
+ * @param remote the remote details found in the Git repository
+ */
+export function repositoryMatchesRemote(
+  gitHubRepository: GitHubRepository,
+  remote: IRemote
+): boolean {
+  return (
+    gitHubRepository.cloneURL === remote.url ||
+    gitHubRepository.htmlURL === remote.url
```
niik: Should we make these checks case-insensitive given that they're GitHub specific? Also, should we approach this the same way we do in `matchRemoteWithAccount` i.e. not care about the protocol?

shiftkey: :thumbsup: to case-insensitve

> i.e. not care about the protocol?

I think the only place where this will come is with `cloneURL` not matching `remote.url` when they are SSH/HTTPS. I'll give it a shot and see how it feels

shiftkey:
:apple:

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +95,47 @@ export function matchExistingRepository(
...
+ * @param gitHubRepository the GitHub repository details found from the GitHub API
+ * @param remote the remote details found in the Git repository
+ */
+export function repositoryMatchesRemote(
+  gitHubRepository: GitHubRepository,
+  remote: IRemote
+): boolean {
+  const htmlUrl = gitHubRepository.htmlURL || ''
+
+  if (caseInsensitiveCompare(htmlUrl, remote.url) === 0) {
```
niik: While this happens to work here I don't think we should conflate comparison with equality. The comparison function is for sorting. Let's either do a manual lower case comparison here or add a dedicated `caseInsensitiveEquals` method in `equality.ts` and use that.

I'm assuming that we're doing this straight up equality check on htmlUrl here as a happy path? If so, shouldn't we be doing the same thing for cloneURL?

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +95,47 @@ export function matchExistingRepository(
...
+ * .git suffix) or it might match the htmlURL field (without a .git suffix)
+ *
+ * @param gitHubRepository the GitHub repository details found from the GitHub API
+ * @param remote the remote details found in the Git repository
+ */
+export function repositoryMatchesRemote(
+  gitHubRepository: GitHubRepository,
+  remote: IRemote
+): boolean {
+  const htmlUrl = gitHubRepository.htmlURL || ''
```
niik: Granted, this is an academic discussion but Could we opt to not perform htmlURL checks if it's null rather than coalescing to an empty string?

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +95,47 @@ export function matchExistingRepository(
...
+
+  if (caseInsensitiveCompare(htmlUrl, remote.url) === 0) {
+    return true
+  }
+
+  const cloneUrl = parseRemote(gitHubRepository.cloneURL || '')
+  const remoteUrl = parseRemote(remote.url)
+
+  if (remoteUrl == null || cloneUrl == null) {
```
niik: Could we make these triple equals?

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +95,47 @@ export function matchExistingRepository(
...
+  gitHubRepository: GitHubRepository,
+  remote: IRemote
+): boolean {
+  const htmlUrl = gitHubRepository.htmlURL || ''
+
+  if (caseInsensitiveCompare(htmlUrl, remote.url) === 0) {
+    return true
+  }
+
+  const cloneUrl = parseRemote(gitHubRepository.cloneURL || '')
```
niik: Same comment here about opting not to parse when null rather than coalescing to an empty string

shiftkey:
:apple:

app/src/lib/repository-matching.ts
```diff
@@ -92,3 +95,52 @@ export function matchExistingRepository(
     }) || null
   )
 }
+
+export function repositoryMatchesRemote(
```
niik: Could we slap a little bit of documentation on this please?

shiftkey:
:apple:

PR #4117, opened 2018-02-27
handle error event in FSWatcher

shiftkey:
Found this in Haystack today:

```
Error watching file for changes: EPERM
exports._errnoException (util.js:1050:11)
FSEvent.FSWatcher._handle.onchange (fs.js:1386:11)
```

Looking at the source for `onchange`:

```js
  this._handle.onchange = function(status, eventType, filename) {
    if (status < 0) {
      self._handle.close();
      const error = !filename ?
          errnoException(status, 'Error watching file for changes:') :
          errnoException(status,
                         `Error watching file ${filename} for changes:`);
      error.filename = filename;
      self.emit('error', error); // this is the relevant line
    } else {
      self.emit('change', eventType, filename);
    }
  };
```

Like what we did with https://github.com/desktop/dugite/pull/164, not handling the `error` event is meaning this gets thrown anyway, so let's do something about it.

I set the log level to `warn` here because we don't need to use `log.error` to see this in Haystack, as we can't really do much about this within the app.


app/src/lib/tailer.ts
```diff
@@ -43,6 +43,10 @@ export class Tailer {
 
     try {
       const watcher = Fs.watch(this.path, this.onWatchEvent)
+      watcher.on('error', error => {
```
niik: Does an error automatically close the watcher or should we make sure that it's closed?

Also, should we expose this error through some event for consumers? Seems like they'll never know that tailing has failed?

shiftkey: > Does an error automatically close the watcher or should we make sure that it's closed?

I believe so. Manually closing the FSWatcher closes the underlying handle (of type `FSEvent`).

```js
FSWatcher.prototype.close = function() {
  this._handle.close();
};
```

Inside `onchange` it'll do the same:

```js
  this._handle.onchange = function(status, eventType, filename) {
    if (status < 0) {
      self._handle.close();
      ...
      self.emit('error', error);
    } else {
      self.emit('change', eventType, filename);
    }
  };
```

> Also, should we expose this error through some event for consumers? Seems like they'll never know that tailing has failed?

I haven't thought much about what consumers would want to do to handle an `error` event being emitted. This is currently only use for parsing Git LFS progress, and an error (either when initializing the file or when `onchange` occurs) means that we "give up" on reporting progress.

If there's something UI-wise that we should be doing here to report the lack of progress, then I'm :thumsbup: to emitting this. But I wouldn't hold up this fix to the unhandled error unless you think it's important.

shiftkey:
:apple:

app/src/lib/tailer.ts
niik: > I believe so. Manually closing the FSWatcher closes the underlying handle (of type FSEvent).

Excellent!

> This is currently only use for parsing Git LFS progress, and an error (either when initializing the file or when onchange occurs) means that we "give up" on reporting progress and then the tailer is disposed.

> I wouldn't hold up this fix to the unhandled error unless you think it's important.

I concur that it has little practical application in our use case at the moment but I think it makes conceptual sense to have the Tailer class expose errors. The decision about whether or not to care about not being able to tail a file should fall on the consumer, not be hidden away in the implementation. Similarly I think we should capture and propagate the error event from the read stream and let the LFS consumer handle that as well.

Since we're not doing a release this week I think it makes sense to address this while we're in here.

shiftkey:
:apple:

PR #4125, opened 2018-02-27
make the keychain known issue more clear

shiftkey:
This is already captured as a known issue, but I wanted to massage the error into something less vague. I've not been able to reproduce this on High Sierra (so great!) which is why I'm trying to isolate it from the original code path (which would just emit the error, untouched).


app/src/lib/stores/accounts-store.ts
```diff
@@ -83,7 +83,27 @@ export class AccountsStore extends BaseStore {
...
-      this.emitError(e)
+
+      if (__DARWIN__) {
+        const error = e as Error
+
+        if (
+          error.message &&
+          error.message.startsWith(
+            'The user name or passphrase you entered is not correct'
+          )
```
niik: Could we extract this check into a helper function? `IsKeychainError` or something? That way we should be able to clean this method up and get rid of a bunch of these nested statements to something similar to

```ts
catch (e) {
  if (__DARWIN__ && isKeyChainError(e)) {
    this.emitError(new Error(...))
  } else {
     this.emitError(...)
  }
}
```

shiftkey:
:apple:

PR #4139, opened 2018-03-01
ensure compare is comparing the right sort of string

shiftkey:
Found while tidying up #4090 and traced it back to dd63bc0fc5a5f735539da2875bb0077e1024b192 which seems to suggest this was a copy-paste gone wrong, and both should be `toLowerCase()`.

### DIFF
--- a/app/src/lib/compare.ts
+++ b/app/src/lib/compare.ts
@@ -71,7 +71,7 @@ export function compareDescending<T>(x: T, y: T): number {
  * used in a sort operation the comparands will be sorted in ascending order.
  */
 export function caseInsensitiveCompare(x: string, y: string): number {
-  return compare(x.toLowerCase(), y.toLocaleLowerCase())
+  return compare(x.toLowerCase(), y.toLowerCase())
 }
 
 /**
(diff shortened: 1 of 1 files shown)


app/src/lib/compare.ts
```diff
@@ -71,7 +71,7 @@ export function compareDescending<T>(x: T, y: T): number {
  * used in a sort operation the comparands will be sorted in ascending order.
  */
 export function caseInsensitiveCompare(x: string, y: string): number {
-  return compare(x.toLowerCase(), y.toLocaleLowerCase())
+  return compare(x.toLowerCase(), y.toLowerCase())
```
j-f1: Or should these both be `toLocaleLowerCase`?

shiftkey: From [StackOverflow](https://stackoverflow.com/a/34283467/1363815):

> In most cases, this will produce the same result as toLowerCase(), but for some locales, such as Turkish, whose case mappings do not follow the default case mappings in Unicode, there may be a different result.

I don't think we have plans to be doing string comparisons in the app that might trigger the ["Turkisk I" problem](https://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/) so I'm fine with leaving this as-is.

Got any further feels @niik?

niik: > Got any further feels @niik?

I think we should use `toLowerCase` in this particular method. If we end up wanting to do locale-aware sorting in our lists for example we should introduce a new function that specifically does that. When we start to look at that though we'll most likely want to take a broader look at how we want sorting to work (i.e. likely lexicographical)

PR #4145, opened 2018-03-01
Remove curly rule

OiYouYeahYou:
Omitting the rule 'curly' from eslint can mean that the visual complexity and clutter can be reduced for very simple if statements, etc.

Instead of this

```typescript
    if (this.stopped) {
      return
    }

    const shouldFetch = this.shouldPerformFetch(this.repository)
    if (shouldFetch) {
      try {
        await this.fetch(this.repository)
      } catch (e) {
        log.error('Error performing periodic fetch', e)
      }
    }

    if (this.stopped) {
      return
    }

    const interval = await this.getFetchInterval(repository)
    if (this.stopped) {
      return
    }

    this.timeoutHandle = window.setTimeout(
      () => this.performAndScheduleFetch(repository),
      interval
    )
```

Can look something like this:

```typescript
    if (this.stopped) return

    const shouldFetch = this.shouldPerformFetch(this.repository)
    if (shouldFetch) {
      try {
        await this.fetch(this.repository)
      } catch (e) {
        log.error('Error performing periodic fetch', e)
      }
    }

    if (this.stopped) return

    const interval = await this.getFetchInterval(repository)
    if (this.stopped) return

    this.timeoutHandle = window.setTimeout(
      () => this.performAndScheduleFetch(repository),
      interval
    )
```

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -52,7 +52,6 @@ rules:
   ###########
   # BUILTIN #
   ###########
-  curly: error
   no-new-wrappers: error
   no-redeclare:
     - error


shiftkey:
I'm somewhat 👎 to this change, because I worry about early `return`s being hard to follow - despite using prettier everywhere to ensure we're consistent. I'll elaborate more after the :coffee: kicks in.

cc @desktop/core and @desktop/comrades for feels

lee-dohm:
While removing curly braces where they are not required may reduce visual clutter, it has never been shown to increase comprehension and has been shown to marginally increase the number of bugs inadvertently introduced. Some style rules are simply that, style or personal preference. Other style rules, like this one, have a definite impact on things that are important to a successful project.

niik:
Thank you so much for wanting to contribute @OiYouYeahYou. Unfortunately it looks like there's not enough interest in this change for us to bring it in so I'm going to close it out. Feel free to continue the discussion here or in an issue.

PR #4148, opened 2018-03-01
when there are no open pull requests we should still clear the cache

### LINKED ISSUE #4122: Merged pull request still shown in Desktop several hours after merging/closing
### Description

When a pull request is created, it is shown in the 'Pull requests' tab in the 'Current branch' menu. I then navigate to GitHub.com and merge the commit, but the pull request is still shown in GitHub desktop 'Pull requests' tab. 

If I continue to commit to this branch and then try and 'Create a pull request', the menu option has changed to 'Show pull request' and does not allow me to create a pull request (even though the previous pull request has already been merged).

### Version

**GitHub Desktop version:** 1.1.0

**OS version:** Windows 10

### Steps to Reproduce

1. Create branch and perform a commit
1. Create and approve a pull request
1. Perform another commit on the same branch
1. Try and create another pull request

**Expected behavior:** I should be able to create another pull request after the previous one has been merged

**Actual behavior:** IThe menu item has changed, only allowing me to show the previous pull request

**Reproduces how often:** Every time

#### Additional Information

This has only happened since GitHub Desktop v1.1.0

shiftkey:
Fixes #4122 

This issue seems to be related to the scenario where you merge out the last PR in a repository, and the API now returns no open pull requests. In theory, we should be clearing the table in IndexedDB when this occurs, but instead we exit early because we don't have any pull request to add.

Instead, we should always clear the underlying table, and conditionally insert the new pull requests.

### DIFF
--- a/app/src/lib/stores/pull-request-store.ts
+++ b/app/src/lib/stores/pull-request-store.ts
@@ -388,16 +388,7 @@ export class PullRequestStore extends TypedBaseStore<GitHubRepository> {
       })
     }
 
-    if (prsToInsert.length <= 0) {
-      return
-    }
-
     return this.pullRequestDatabase.transaction('rw', table, async () => {
-      // since all PRs come from the same repository
-      // using the base repoId of the fist element
-      // is sufficient here
-      const repoDbId = prsToInsert[0].base.repoId!
-
       // we need to delete the stales PRs from the db
       // so we remove all for a repo to avoid having to
       // do diffing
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/pull-request-store.ts
```diff
@@ -388,16 +388,7 @@ export class PullRequestStore extends TypedBaseStore<GitHubRepository> {
...
 
-    if (prsToInsert.length <= 0) {
-      return
-    }
-
     return this.pullRequestDatabase.transaction('rw', table, async () => {
-      // since all PRs come from the same repository
-      // using the base repoId of the fist element
-      // is sufficient here
-      const repoDbId = prsToInsert[0].base.repoId!
```
shiftkey: We already define `repoDbId` at the start of the function, so this is unnecessary work

iAmWillShepherd: Good catch!

bobsmith947:
Thanks for fixing this. The issue was pretty annoying.

PR #4150, opened 2018-03-02
hide github-desktop-* from branches list

### LINKED ISSUE #4127: branch list containing many forks and refs is hard to navigate
**Background:** I work with a dozen remotes —forks of my project— at any given time, as I need to edit and rebase submitted PRs and push changes back to those remotes. Since "git remote add" brings in all their branches there are >100 branches listed in the Branches dropdown list.

**Request:** I'd like to be able to collapse the remotes in the list to reduce the number of top-level items and reduce visual noise.

shiftkey:
Fixes #4127 to make the branch list easier to browse

Before: branches and branches for days

After: all `github-desktop-*` refs are no longer listed, other remotes still listed

- [x] confirm branch operations like checkout fork PRs aren't affected by this

### DIFF
--- a/app/src/lib/git/for-each-ref.ts
+++ b/app/src/lib/git/for-each-ref.ts
@@ -102,6 +102,14 @@ export async function getBranches(
       continue
     }
 
+    if (ref.startsWith('refs/remotes/github-desktop-')) {
+      // hide refs from our known remotes as these are considered plumbing
+      // and can add noise to everywhere in the user interface where we
+      // display branches as forks will likely contain duplicates of the same
+      // ref names
+      continue
+    }
+
     branches.push(
       new Branch(name, upstream.length > 0 ? upstream : null, tip, type)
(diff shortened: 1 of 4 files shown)


shiftkey:
I've tested this out by checking out fork PRs without issues, but you'll probably also encounter #4149 at the same time.

app/src/lib/git/for-each-ref.ts
```diff
@@ -102,6 +102,14 @@ export async function getBranches(
       continue
     }
 
+    if (ref.startsWith('refs/remotes/github-desktop-')) {
```
iAmWillShepherd: There's a constant we export for `github-desktop`, maybe use that?
https://github.com/desktop/desktop/blob/71fb470e4fcc234a4d471b330704ec71432f44d8/app/src/lib/stores/pull-request-store.ts#L26

PR #4153, opened 2018-03-02
Add message to gitignore dialog when not on the default branch

### LINKED ISSUE #3720: Update `gitignore` dialogue if not on default branch
Add differentiating text to the `gitignore` dialogue if the user is not on the default branch. 

As each branch can have its own `gitignore`, the Repo Settings > Ignored Files text should have a subtle confirmation to the user that they are saving a `gitignore` file only for that branch.

screen shot 2018-01-03 at 10 38 19 am

https://github.com/desktop/desktop/blob/a61a5bdc94ee8237dfff328957cdaee99a9b61e1/app/src/ui/repository-settings/git-ignore.tsx#L16

OSX, 1.0.12-beta0

cc @lee-dohm

lee-dohm:
Fixes #3720

I'm still not super confident on the language in the dialog but this is the best I've come up with so far. Thoughts?

/cc @shiftkey @tierninho

### DIFF
--- a/app/src/ui/repository-settings/git-ignore.tsx
+++ b/app/src/ui/repository-settings/git-ignore.tsx
@@ -16,9 +16,9 @@ export class GitIgnore extends React.Component<IGitIgnoreProps, {}> {
     return (
       <DialogContent>
         <p>
-          Editing <Ref>.gitignore</Ref>. This per-branch file specifies
-          intentionally untracked files that Git should ignore. Files already
-          tracked by Git are not affected.{' '}
+          Editing <Ref>.gitignore</Ref>. This file specifies intentionally
+          untracked files that Git should ignore. Files already tracked by Git
+          are not affected.{' '}
           <LinkButton onClick={this.props.onShowExamples}>
             Learn more
           </LinkButton>
--- a/app/src/ui/repository-settings/git-ignore.tsx
+++ b/app/src/ui/repository-settings/git-ignore.tsx
@@ -19,8 +19,8 @@ export class GitIgnore extends React.Component<IGitIgnoreProps, {}> {
 
     return (
       <p>
-        You are not on the default branch, so changes here may not be applied to
-        the repository when you switch branches.
+        You are not on the default branch, changes here will only be stored
+        locally on this branch.
       </p>
     )
   }
--- a/app/src/ui/repository-settings/git-ignore.tsx
+++ b/app/src/ui/repository-settings/git-ignore.tsx
@@ -18,7 +18,7 @@ export class GitIgnore extends React.Component<IGitIgnoreProps, {}> {
         <p>
           Editing <Ref>.gitignore</Ref>. This per-branch file specifies
           intentionally untracked files that Git should ignore. Files already
-          tracked by Git are not affected.
+          tracked by Git are not affected.{' '}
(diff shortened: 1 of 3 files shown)


app/src/ui/repository-settings/git-ignore.tsx
```diff
@@ -23,6 +37,9 @@ export class GitIgnore extends React.Component<IGitIgnoreProps, {}> {
           for more information about the file format, or simply ignore a file by
           right clicking on it in the uncommitted changes view.
         </p>
+
+        {this.renderBranchWarning()}
```
j-f1: `{this.props.isDefaultBranch && <p>...</p>}`?

lee-dohm: That would return `false` instead of `null` in the negative case. Would that render differently?

j-f1: [Nope](https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored):

> `false`, `null`, `undefined`, and `true` are valid children. They simply don’t render.

lee-dohm: Thanks for that @j-f1. Still, six of one, half dozen of the other. I feel this looks cleaner and the same pattern is used in other UI elements.

tierninho:
@lee-dohm language looks good to me. Slight tweak?

Current:
You are not on the default branch, so changes here may not be applied to the repository when you switch branches.

New?:
As you are not on the default branch, changes here may not be applied to the repository when you switch branches.

lee-dohm:
Thanks @tierninho. Updated :+1:

lee-dohm:
@niik I totally understand your concerns. I'll take a stab at it. I like the wording you include in your screenshots, I'll see if I can expand on it just a little bit.

lee-dohm:
Here's the updated view:

screen shot 2018-03-05 at 9 00 51 am

niik:
@lee-dohm I think that's an improvement but I worry that the 'per-branch' phrasing may be confusing, that it could be interpreted as gitignore files somehow being attached to branches. I don't have a great alternative though 😕

lee-dohm:
I can take out the "per-branch" but leave the improved wording. We still need to decide what to do with #3720 though. Should we just punt on it for now by closing it and mark it for the future? Or leave it open?

shiftkey:
@lee-dohm @niik I'm fine with closing out #3720 if we don't think there's value in talking about "per-branch" - @tierninho any objections given the discussion above?

niik:
> @lee-dohm @niik I'm fine with closing out #3720 if we don't think there's value in talking about "per-branch" here

I'm happy with that. @lee-dohm if you have time please pull that per-branch phrasing and then we'll be ready to land this once @tierninho has weighed in.

lee-dohm:
Done.

niik:
> I'm happy with that. @lee-dohm if you have time please pull that per-branch phrasing and then we'll be ready to land this once @tierninho has weighed in.

I just realized that @tierninho is off today so I'll just bring this in and we can continue to iterate when he gets back.

lee-dohm:
Awesome, thanks @niik :+1:

PR #4178, opened 2018-03-05
Close issues faster

iAmWillShepherd:
Figured this is as good a place as any to ask the question is 14 days too long?

Close out issues labelled `more information needed` faster.


iAmWillShepherd:
Question was answered internally.

Issue #3956, opened 2018-02-07
[Feature Tracker - RBB] Compare Branches Feature tracker


nerdneha:
### Purpose

When working on a local branch, I should be able to get a clear understanding of the commits I've made and how that compares to master or any other target branch. I can understand how I've diverged from a branch and, if I'm behind that branch by any number of commits, I can pull that branch's commits into my branch.
* notifications around diverging from master branch and proactively detecting merge conflicts is in a later release
* continued from initial discussion here: #2639 

### Dependencies:
- [x] Compare to Branch #3967 - compare my work to any branch and see the commits that are ahead of that branch
- [x]  Ahead/Behind Branch Commit Counts #3968 - tags showing ahead/behind for each branch when choosing a branch 
- [x] Ahead/Behind Toggle #3997 - toggle between commits ahead and behind for any branch
- [x] Merge #3998 - merge in any branch to my branch for the "behind" tab
- [x] #4318 - Loading ahead/behind info for each branch is slow 
- [x] Analytics Discussion https://github.com/github/analytics/issues/347
- [x] #4439 - Analytics Implementation

Designs below - https://github.com/desktop/desktop/issues/3956#issuecomment-363998416

iAmWillShepherd:
For some additional context, here's @donokuda's [mockups](https://github.com/desktop/desktop/issues/2639#issuecomment-333621771) from #2639

Design time! 😄 Here are a couple problems we are aiming to solve with comparing between two branches:

-  There currently isn't a way to see how far we've diverged from a branch. As a result, it's easy for our working branch to diverge from a default branch (such as master) over time.
- There's no easy to see what kind of changes are about to be brought in. This also adds to the difficulty of not being able to see if there are any overlapping work.

![image](https://user-images.githubusercontent.com/1174461/30655899-24154946-9de7-11e7-90cd-cc1b69334dd3.png)

I started off with a version where I played around with a graph/timeline/subway map visualization to show which commits existed on which branch. The goal at the time was to be able to visualize the result of a merge (as well as understand which specific commits would be brought in:

<img src="https://user-images.githubusercontent.com/1174461/30396666-3ba54c86-987f-11e7-8f18-a7bb7ea5f1a1.png" width="400" />

After some discussion, we've agreed on some of the limitations of that design:

- The graph doesn't offer any value and (not seen in the crop) doesn't accurately represent events such as merges.
- We're not convinced of the value of showing all of the commits
- We're losing precious horizontal space

[...]

Issue #750, opened 2016-12-14
Merged branches stay around until they're manually deleted


joshaber:
@shiftkey writes

> In previous versions of Desktop, we had this behaviour for pruning:
> 
> - after a sync was completed, this is a chance to find branches that have been merged and removed locally
> - if the current branch is in a detached HEAD or unknown state, skip pruning
> - we look for branches which have been merged into the default branch (not the current branch)
> - on Windows we also had some additional guards to skip reserved branch names, like `gh-pages`
>
> I've created a branch [`prune-merged-braches-demo`](https://github.com/desktop/desktop/compare/prune-merged-braches-demo) that shows how you might add this into our process, but is deliberately incomplete - the `TODO` statements need to be addressed as part of implementing this.
>
> There might be some changes with how we do this, but this is a great chance for someone who cares about this feature to have a go at implementing it!

**Problem statement:**

Branches stick around in Desktop indefinitely until manually deleted. I want the branches that show up in Desktop to be relevant to me and not muddied by stale branches.

**Proposed solution:**

If 
- a branch is deleted on GitHub.com **and** 
- merged locally into the default branch in Desktop **and** 
- has no commits ahead of the default branch

then 
- they should be deleted in Desktop after a certain amount of time (to allow for people who frequently reuse the same branch for their work if that's a common workflow).

**Considerations:**
1. If there's a difference between the commits of the remote on GitHub.com and your local branch, we should not delete the local branch.
2. If GitHub has a concept of reserved branch names as @shiftkey references above (like `gh-pages`), we should skip pruning those branches.
3. If there's work that's **not** merged into default on GitHub.com (for example, a PR that's closed and deleted), we should not delete the local branch because it has work that's ahead of the default branch.

**Questions:**
1. How long before we automatically prune branches? (14 days)
2. How frequently do we check branches for ability to be pruned? (once per day)
3. Do we show anything in the UI to indicate that a branch will be pruned? (We don't think so - we're being intentionally conservative here and we don't think we'll put you in a position to lose any work here)
4. Are there reserved branches here that we should account for? (unknown)
5. Does it make sense to consider squashed and merged branches here (#1771)? (unknown)

alrz:
Copying my comment from https://github.com/desktop/desktop/issues/1128#issuecomment-302165993 at @shiftkey request.

>when I choose to update all remotes it'd be nice if it run in parallel -- `git remote update --prune` does it sequentially.

caiofbpa:
[...]

PR #4182, opened 2018-03-06
Compare to Branch

### LINKED ISSUE #3967: [RBB] Compare to Branch
This is part of #3956 and is for viewing changes as compared to any branch (ex: master).

**Changes:**
- [x] create a feature flag (ideally we don't change anything in the `HistorySideBar` component)
- [x] copy change (history -> compare, changes -> commit [may not be reflected in mocks, talk to @niik])
- [x] `branch selecting input` with new icon in input
- [x] list of branches on `branch selecting input` click (if it's empty) which can be "x"ed out
- [x] clicking on a branch shows commits ahead of the target branch
- [x] filtering

**User Flow to check functionality:**
- click on compare tab and see the commits you made on your branch + all commits from where you branched off of (existing behavior)
- click into the "compare a branch..." input and see a list of branches just like you would for switching branches if you clicked on the "current branch" menu item
- click on "master" and "master" becomes the text in the input and you see all the commits you have ahead of "master"
- press "x" in the branch input and see list of branches again
- select another branch, ex: "new-branch" and "new-branch" becomes the text in the input and you see all the commits you have ahead of "new-branch"
- press "x" in the branch input and see list of branches again
- start typing the name of an existent branch in "compare a branch..." input 
- see all branches filtered with same behavior as we have for the branches container
- select a new target branch, ex: "filtered-branch" and "filtered-branch" becomes the text in the input and you see all the commits you have ahead of "filtered-branch"
- press "x" in the branch input and see list of branches again
- see the compare tab return to default state

**Other pieces**:
- ahead/behind counts for each branch is in #3968 
- toggle for behind/ahead or any behind info in #3997

**Notes:**
@donokuda said this may change soon

**Before:** 
screen shot 2018-02-08 at 2 11 27 am

(branch list in app for reference)
screen shot 2018-02-12 at 11 02 10 pm

**After:**
(default view)
default view

(select branch options)
select branch options

(selecting branch)
selecting branch

(filtering branch)
filtering branch

### LINKED ISSUE #3968: [RBB]  Ahead/Behind Branch Commit Counts
This is part of #3956 and shows how many commits your branch has diverged from other branches

**User Flow to check functionality:**
- click on compare tab
- click into the "compare a branch..." input and see a list of branches
- for each branch, see the # of commits that branch is ahead and behind your current branch
- start typing the name of an existent branch in "compare a branch..." input 
- see the same #s for ahead/behind without recalculating

**Changes:**
- [x] # of commits your branch is `behind \/` that branch
- [x] # of commits your branch is `ahead ^` that branch

**Other pieces**:
- clicking on a branch to compare is in #3967 

**Before:** 
screen shot 2018-02-08 at 2 11 27 am

(branch list in app for reference)
screen shot 2018-02-12 at 11 02 10 pm

**After:**
filtering branch

### LINKED ISSUE #3997: [RBB] Ahead/Behind Toggle
This is part of #3956 and shows how your branch has diverged from your target branch (ex: master). This will help you understand what changes can be brought into your branch (behind) and what commits are unique to your branch (ahead).

**User Flow to check functionality:**
- go to compare view with **no toggle** showing
- click on compare tab, click into the "compare a branch..." input, and click on "master" and see:
  - the **toggle appear** in the compare view
  - `Behind (x)` selected by default where x is the # of commits your branch is behind "master" (old view showed commits ahead of master)
  - the commits that are missing from your branch that exist in the target branch
- click on `Ahead (y)` in the toggle where y is the # of commits your branch is ahead of "master" and see:
  - the commits that exist only in your branch and not the target branch
- press "x" in the branch input, select another target branch, and see:
  - the compare view where you can toggle behind/ahead
  - the respective commits listed

**Changes:**
- [x] create toggle
- Behind **(x)** & Ahead **(y)** counts
- [x] `Behind (x)` is selected by default
- [x] default view shows the commits that your branch is behind (relative to the target branch)
- [x] clicking `Ahead (y)` shows commits 
- [x] clicked `Ahead (y)` shows the commits that your branch is ahead (relative to the target branch)

**Other pieces**:
- merging into my branch is in #3998 

**Before (shows "ahead" by default):**
screen shot 2018-02-08 at 2 11 27 am

**After:**
image

### LINKED ISSUE #3998: [RBB] Merge changes into your branch
This is part of #3956 and provides the option to bring changes from a target branch into your branch using a merge button

**User Flow to check functionality:**
- click on compare tab, click into the "compare a branch..." input, and click on "master" and see:
  - `Behind (x)` in the toggle selected with the commits that are missing from your branch that exist in the target branch
  - a **merge button** saying `Merge into my-branch-name`
  - supporting text `This will merge x commits from (target branch name)`
- click on `Ahead (y)` in the toggle where y is the # of commits your branch is ahead of "master" and see:
  - **no merge button**
- click "x" in the branch input, select another target branch, and see:
  - the option to **merge a different target branch's commits into your branch** in the default behind view only
- click back to `Behind (x)` and click on the **merge button** and see: 
  - commits disappear from Behind area and toggle becomes `Behind (0)`
  - Ahead is not affected
- click "x" in the branch input and focus out of the input and see:
  - the recently merged commits included in the history now

**Changes:**
- [x] merge button with text `Merge into (my-branch-name)`
- [x] supporting text below merge button with a count matching x in the `Behind (x)` toggle
- [x] clicked `Merge into (my-branch-name)` merges those changes into your branch
- [x] disabled merge button when there is nothing to merge (ex: after completing a merge)
- [x] reverting on conflict + error message

**Other pieces**:
- toggling is in #3997
- predicting merge conflicts will happen in a later issue

**Before (shows "ahead" by default):**
screen shot 2018-02-08 at 2 11 27 am

**After:**
image

iAmWillShepherd:
#So, I accidentally built all the things in #3956, my bad. I was confused because I used the image in the first issue to build out my tasks list.

Fixes #3967
Fixes #3968 
Fixes #3997 
Fixes #3998

I also created issues #4317, #4458, and #4318 to track additional work that needs to be done around this feature.

### MVP
- [x] Add feature flag
- [x] Rename _Changes_ tab -> _Commit_
- [x] Rename _History_ tab -> _Compare_
- [x] Add UI to select a branch
- [x] Add `TabBar` component to allow selecting *ahead* or *behind*
- [x] Update commit list based on `CompareType`
- [x] If no branch is selected we should show all commits (basically do what we've always done) and hide the _ahead/behind_ buttons
- [x] Update ahead/behind info when branch is switched
- [x] Add a button to merge commits that are **ahead** of the current branch

### Polish
- [x] Switch from select input to a text input
  - [x] Input should show a filter-list of branches on focus
  - [x] Each branch should show the number of commits ahead and behind the current branch
    - [x] Figure out a way to load ahead/behind for every branch (All branches currently show the same ahead/behind info of the currently selected branch).
  - [x] Selecting a branch should populate the text input with its name
  - [x] Text input should have an icon that, when clicked, clears the text in the input field and takes user back to normal layout (pressing <kbd>escape</kbd> should do the same thing)
- [x] Add a new `CompareBranchListItem` component that has all the logic to load the compare state async
- [x] When typing in non-existent branch names the branch list is eventually empty, which resets the filter, which then displays all branches again
- [x] after selecting a branch to compare, switching Compare -> Commit -> Compare remembers most of the form state except for the filter text - **fixed in #4396**
- [x] Blank-slate image positioning is off (see [comment](https://github.com/desktop/desktop/pull/4390#issuecomment-378569611)) 
- [x] Change empty list message from **No History** -> **No Commits** -  **fixed in #4396**
- [x] Improve styling so that longer branch names don't display weirdly (see [comment](https://github.com/desktop/desktop/pull/4182#issuecomment-378806869))

### QA
- [x] Update top-level menu to reflect new `Compare` and `Commit` tab names
~Disable <kbd>return</kbd> key when no matches exist (this is harder than it seems, may need to track in a new issue)~ (tracking in #4458)
- [x]  Auto-complete it __too__ eager, we want to match character-by-character
- [x] Ahead/Behind tab bar is too sticky, when a branch is switched, default to behind tab
- [x] Any branch with commits beyond roughly 100, do not show when scrolling down
- [x] User cannot "arrow-key-down" from compare-branch filter to the search results. Nor can the user "arrow-key down" from the Behind or Ahead tab to the commits.
- [x] Behind/Ahead placement is not consistent. A/B or B/A?
- [x] when selecting a branch, the user is automatically navigated to the Behind/Ahead tabs, yet the Behind tab is not highlighted blue until the user clicks it
- [x] Odd behavior... select "origin/upgrade-yarn-time" as branch to compare. Behind/Ahead tabs shown with results as expected. Now place cursor back in search bar, and the "origin/upgrade-yarn-time" should show as a search result. Now hit enter, and notice there is nothing there. Behind/Ahead tabs should be shown.

### Incomplete
* #3998: reverting on conflict + error message


iAmWillShepherd:
I took @shiftkey's advice and went back to using a select input because it's much easier to test switching branches. The code to do auto-completion is complete, so switching back shouldn't be too much of an issue.

iAmWillShepherd:
Currently working through issue where `ahead/behind` isn't updated for a selected branch until you click on one of the radio buttons.

iAmWillShepherd:
Right now, since I am relying on the `BranchList` component to show branches, I am not able to render the _ahead/behind_ info. I'm considering modifying the component to accept a `BranchItem` render prop so that I can customize the rendering of the list. I also updated the `FilterList` to accept a `TextBox` component. This is needed because I need the `TextBox` to always be visible, so I cannot rely on the one that's built into the `FilterList` component.

iAmWillShepherd:
Most functionality is here, but this will need some CSS magic to really make it pop.

iAmWillShepherd:
### Branches List
The `TextBox` takes up too much vertical space, pushing the branch list down. Removing the CSS rule `flex: 1;` seems to make the element take up only the space it needs.



### Commit List
The commit list should be visible directly underneath the two radio buttons. @shiftkey mentioned this is because it's a React-Virtualized component and will only render things when it has space to do so. Giving it some `height` in Chrome dev tools works **only** when I set the value to a height in pixels first (I can use `height: 100%` afterwards)

iAmWillShepherd:
@nerdneha check out the description under the _Merging_ header.

app/src/ui/lib/filter-list.tsx
```diff
@@ -118,7 +118,8 @@ interface IFilterListProps<T extends IFilterListItem> {
   /** Called when there are no items to render.  */
   readonly renderNoItems?: () => JSX.Element | null
 
-  readonly filterTextBox?: TextBox
+  /** Used to get a reference to a TextBox that will be used to control this component */
+  readonly filterTextBox?: TextBox //Todo: don't rely on passing textbox to component
```
iAmWillShepherd: @niik and I talked about this and agreed we can approach this better. I created #4317 to track that work.

app/src/ui/lib/filter-list.tsx
```diff
@@ -154,6 +154,12 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
     this.state = createStateUpdate(props)
   }
 
+  public componentWillMount() {
```
iAmWillShepherd: Should this be `componentDidMount` instead?

app/src/ui/branches/branch-list.tsx
```diff
@@ -231,15 +231,11 @@ export class BranchList extends React.Component<
...
-    if (this.props.canCreateNewBranch) {
-      return (
-        <Button className="new-branch-button" onClick={this.onCreateNewBranch}>
-          {__DARWIN__ ? 'New Branch' : 'New branch'}
-        </Button>
-      )
-    } else {
-      return null
-    }
+    return this.props.canCreateNewBranch ? (
```
iAmWillShepherd: Using ternary because of the example in our [engineering values](https://github.com/desktop/desktop/blob/master/docs/contributing/engineering-values.md#immutability-is-good). However, I'm not against switching to normal `if/else` statements for the sake of readability. I do like the succinctness of this though.

app/src/lib/stores/app-store.ts
```diff
@@ -699,26 +699,27 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-        this.updateCompareState(repository, state => ({
-          compareType,
-          branch,
-          commitSHAs: compare.commits.map(commit => commit.sha),
-          ahead: compare.ahead,
-          behind: compare.behind,
-        }))
-      }
+    if (branch == null) {
+      log.debug(`oops, we can't compare if the branch isn't selected`)
```
iAmWillShepherd: All 👂 for a better error message here.

app/src/ui/history/compare.tsx
```diff
@@ -18,6 +18,7 @@ import { FancyTextBox } from '../lib/fancy-text-box'
 import { OcticonSymbol } from '../octicons'
 
 enum SelectedTab {
+  None = -1,
```
iAmWillShepherd: I feel that *None* does a better job of communicating intent than *Default* or defining the type of `selectedTab` as `readonly selectedTab: SelectedTab | null`.

app/src/ui/lib/fancy-text-box.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  /** Callback used to get a reference to internal TextBox */
+  readonly onRef: (textbox: TextBox) => void
+}
+
+interface IFancyTextBoxState {
+  readonly isFocused: boolean
+}
+
+export class FancyTextBox extends React.Component<
```
iAmWillShepherd: We wanted to extend the `TextBox` component so we wouldn't have to add more _configuration props_. I'm all 👂 on a name if the name is a bit too fancy 🥂 .

app/src/ui/branches/branch-list.tsx
```diff
@@ -91,6 +92,13 @@ interface IBranchListProps {
    * to prepopulate the new branch name field.
    */
   readonly onCreateNewBranch?: (name: string) => void
+
+  readonly textbox?: TextBox
+
+  readonly onRenderItem?: (
```
iAmWillShepherd: This may seem weird, but I needed a way to control how an item is rendered and settled on passing a function that the component can delegate to. If this is the wrong approach, I'd love to chat about it.

app/src/ui/history/compare-branch-list-item.tsx
```diff
@@ -0,0 +1,72 @@
...
+  public constructor(props: ICompareBranchListItemProps) {
+    super(props)
+
+    this.state = {
+      compareResult: null,
+    }
+  }
+
+  public async componentWillMount() {
+    const compareResult = await this.props.dispatcher.GetCompareResult(
```
iAmWillShepherd: The performance here is kind of bad. I attempted to use  caching in the `GitStore` but ran into a problem with invalidating entries. If we do use caching, we need to make sure that we update it at the right time to avoid running into issues like never updating the *ahead/behind* info.

app/src/ui/branches/branch-list.tsx
shiftkey: Render props are a fine way craft higher-order components, which sounds like what you're aiming for. This is already something that we need to support with `List` and it's related components, so it's not a foreign pattern.

To take this further, you could make this a required function and extract the logic for the old code path up into `BranchesContainer`.

app/src/ui/branches/branch-list.tsx
```diff
@@ -162,6 +200,15 @@ export class BranchList extends React.Component<
     )
   }
 
+  private renderGroupHeader = (id: string) => {
+    const identifier = id as BranchGroupIdentifier
```
shiftkey: :art: Rather than an implicit cast here, we could do an explicit lookup:

```ts
function parse(label: string): BranchGroupIdentifier | null {
  switch (label) {
    case 'default':
    case 'recent':
    case 'other':
      return label

    default:
      return null
  }
}

...

  private renderGroupHeader = (id: string) => {
    const identifier = parse(id)
    if (identifier == null) {
      return
    }

    return (
      <div className="branches-list-content filter-list-group-header">
        {this.getGroupLabel(identifier)}
      </div>
    )
  }

```

app/src/ui/branches/branch-list.tsx
shiftkey: https://github.com/desktop/desktop/pull/4337

app/src/ui/lib/text-box.tsx
```diff
@@ -228,4 +260,16 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
       </div>
     )
   }
+
+  private onFocus = (event: React.FocusEvent<any>) => {
```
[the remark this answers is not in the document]
iAmWillShepherd: Curious why we use `HTMLInputElement` over `FocusEvent`. Here's what the [docs](https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent) say on it.

app/src/lib/stores/app-store.ts
```diff
@@ -657,6 +678,64 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        commitSHAs: commits,
+        branch,
+        ahead: 0,
+        behind: 0,
+      }))
+      return
+    }
+
+    if (branch == null) {
+      log.debug(`oops, we can't compare if the branch isn't selected`)
```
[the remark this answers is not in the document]
ahmadmiz: Er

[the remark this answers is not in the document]
ahmadmiz: Er

app/src/lib/app-state.ts
```diff
@@ -590,3 +597,25 @@ export interface IChangesState {
...
 }
+
+export interface ICompareState {
+  /**
+   * The ordered local commit SHAs.
+   */
+  readonly commitSHAs: ReadonlyArray<string>
+
+  /**
+   * The branch to compare against the base branch
```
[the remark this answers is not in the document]
iAmWillShepherd: `Branch` should never be `null` here, but I don't have access to a branch when getting the initial state in `AppStore`. Incorporating the changes you suggested (using union types and interfaces, see [40c4db6](https://github.com/desktop/desktop/pull/4182/commits/40c4db624dfff2315ea8f3bd1bc37b64c28acbef)) solved this issue.

app/src/ui/lib/text-box.tsx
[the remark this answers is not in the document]
shiftkey: @iAmWillShepherd not sure I understand. My comment was a suggestion to avoid using `any` unless where we can be more explicit.

iAmWillShepherd:
🔢

app/src/ui/lib/text-box.tsx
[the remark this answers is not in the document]
iAmWillShepherd: oh, gotcha. 👍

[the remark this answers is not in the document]
iAmWillShepherd: I actually talked with niik about my usage of any in the situations a couple of weeks back. I settled on any because we don't really care about the event. Went ahead and updated it tho.

tierninho:
v1.1.1-beta4, OSX 10.13.2

Things I noticed:

- History is still referenced in the menu
screen shot 2018-03-28 at 10 53 06 am

- If you search for a random string, the messaging says "No Branch found", however if you hit `enter` the messaging changes to "No History". Suggest disable `enter` or consistent messaging.

- The autocomplete works, however if you type something similar it still locks you in the that branch. For example, branch `bump-dugite-for-less-EOF` still shows up if I type `bum22222`. Perhaps the user needs to verify that `bum22222` is not a branch?

- If you exit a compare-branch while still on the `Ahead` tab, and search for a new branch to compare, then the default highlighted tab will be the `Ahead` branch. Notice the 5 commits shown is the `Behind` branch, which should be highlighted. 
screen shot 2018-03-28 at 12 44 41 pm

- Any branch with commits beyond roughly 100, do not show when scrolling down

- Got a error that crashed Desktop when switched compare-branch:

```TypeError: Cannot read property 'name' of null
    at CompareSidebar.renderMergeCTAMessage (app/src/ui/history/compare.tsx:206:44)
    at CompareSidebar.renderMergeCTA (app/src/ui/history/compare.tsx:237:3)
    at CompareSidebar.renderActiveTab (app/src/ui/history/compare.tsx:161:5)
    at CompareSidebar.renderTabBar (app/src/ui/history/compare.tsx:273:3)
    at CompareSidebar.renderCommits (app/src/ui/history/compare.tsx:128:3)
    at CompareSidebar.render (app/src/ui/history/compare.tsx:116:3)
    at finishClassComponent (app/node_modules/react-dom/cjs/react-dom.development.js:7873:1)
    at updateClassComponent (app/node_modules/react-dom/cjs/react-dom.development.js:7850:1)
    at beginWork (app/node_modules/react-dom/cjs/react-dom.development.js:8225:1)
    at performUnitOfWork (app/node_modules/react-dom/cjs/react-dom.development.js:10224:1)```

**Comment by j-f1** [comment]
> If you search for a random string, the messaging says "No Branch found", however if you hit enter the messaging changes to "No History". Suggest disable enter or consistent messaging.

How about support for arbitrary refs?

**Comment by iAmWillShepherd** [comment]
> If you search for a random string, the messaging says _No Branch found_, however if you hit enter the messaging changes to "No History". Suggest disable enter or consistent messaging.

@tierninho , this is happening because the input allows you to enter whatever text you like. Suppos we have the following branches in our repo:
* `brownies`
* `marshmallows`
* `cookies`
* `cake`

If I entered `ce` in the input, the list will update to show:
* `cookies`
* `cake`

If I continue typing so that `cent` is now the value of the input:
* `No branch found`

This is correct, because there is no branch which matches `cent`. Now the user can take two actions: 

1. Back out by hitting <kbd>escape</kbd>, deleting the text, or clicking that cancel button that pops up in the input
2. Soldier forward by hitting <kbd>enter</kbd>

It seems that choosing option 2 produces the incorrect behavior because continuing will put you in a state where you are trying to view commits for a branch that doesn't exist. This is the state you are in when you see the message _No history_.

I'm cool with disabling the enter-key when no branches match, but I'd want @desktop/design to weigh in on that one. If everyone agrees on disabling the <kbd>enter</kbd> key when no branches are found, then we can create a separate issue for it.

**Comment by tierninho** [comment]
- User cannot "arrow-key-down" from compare-branch filter to the search results. Nor can the user "arrow-key down"  from the Behind or Ahead tab to the commits.

- Behind/Ahead placement is not consistent. A/B or B/A?
screen shot 2018-03-28 at 2 29 20 pm
![screen shot 2018-03-28 at 2 29 13 pm](https://user-images.githubusercontent.com/14828183/38057519-96ae2602-3294-11e8-9c4b-f0709ff8b357.png

File: app/src/ui/lib/text-box.tsx (hunk 10 above)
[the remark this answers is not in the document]
Comment (shiftkey) [reply app/src/ui/lib/text-box.tsx]: > I settled on any because we don't really care about the event.

If you really don't care about using the event, you could just drop the `event` parameter, e.g:

```
private onFocus = () => {
  if (!this.props.autoFocus && this.props.onFocus !== undefined) {
    this.props.onFocus()
  }
}
```

**Comment by iAmWillShepherd** [comment]
> History is still referenced in the menu
See [comment](https://github.com/desktop/desktop/pull/4182#issuecomment-377014492) for more context.

What should we use in the menu title for this feature? If we follow the pattern, the title wouldn't make much sense IMO (ex. _Show Compare_ _Show Commit). Our previous names were nouns, but out current titles are verbs.

cc @desktop/design

**Comment by donokuda** [comment]
> I'm cool with disabling the enter-key when no branches match, but I'd want @desktop/design to weigh in on that one. If everyone agrees on disabling the enter key when no branches are found, then we can create a separate issue for it.

I think we should disable the enter key if no branches are found.

> What should we use in the menu title for this feature? If we follow the pattern, the title wouldn't make much sense IMO (ex. Show Compare _Show Commit). Our previous names were nouns, but out current titles are verbs.

🤔Off the top of my head, we could change "Show Changes" to "Create Commit" and "Show History" to "Compare Branches." Alternatively, we could just match the titles of the tab and say "Commit" and "Compare" but I'm not sure if that makes as much sense in the "View" category.

This one is a tough one for me so I'm interested in what others think.

File: app/src/ui/history/compare.tsx [hunk 13]
```diff
@@ -0,0 +1,466 @@
...
+        selectedBranch={this.state.selectedBranch}
+        canCreateNewBranch={false}
+        onSelectionChanged={this.onSelectionChanged}
+        onFilterTextChanged={this.onBranchFilterTextChanged}
+        renderBranch={this.renderCompareBranchListItem}
+      />
+    )
+  }
+
+  private renderMergeCTAMessage() {
```
shiftkey: `CTA` isn't an acronym I'm familiar with, and I might be found out to be foolish for asking but I'm gonna do it anyway:

 - what does it stand for?
 - can something more descriptive be used here?

j-f1: Probably call-to-action, but I’m not positive.

iAmWillShepherd: Yep, what @j-f1 said.

shiftkey: We have a few mentions of `CallToAction` already in the codebase, so my gut feeling is that we could simplify this down to `renderMergeSummary` and name the below function `renderMergeCallToAction` rather than abbreviate it especially here. I'll add this into #4390 while I'm cleaning up things.

iAmWillShepherd:
Notice an issue when we have longer branch names.
long-branch-names

shiftkey:
Some notes after walking through the flow a few times:

 - after completing a successful merge into the current branch, it switches to showing the history for your branch. There's no real indicator that it succeeded aside from the ahead/behind indicator changing to indicate you can push new commits. **I wonder if we should show to confirm the clean merge? Or whether we should keep comparing to the current branch to show that your "behind" commits are now empty?** Something to indicate that you were successful here would be :gem:

 - there's some other corner cases where the SHA of the branch tip has changed (after merge, after making a commit) where we don't yet refresh the cached ahead/behind comparisons. I'm looking into these at the moment because I want to find a balance between computing the ahead/behind quickly for the user and not continually doing it because the tip is changing quickly and there's lots of refs to compare.

donokuda:
> after completing a successful merge into the current branch, it switches to showing the history for your branch. There's no real indicator that it succeeded aside from the ahead/behind indicator changing to indicate you can push new commits. I wonder if we should show to confirm the clean merge? Or whether we should keep comparing to the current branch to show that your "behind" commits are now empty? Something to indicate that you were successful here would be 💎

Hmm, I have to play around with this. I agree that if we switch views, we could benefit from something like a success flash messaging.

iAmWillShepherd:
> after completing a successful merge into the current branch, it switches to showing the history for your branch. There's no real indicator that it succeeded aside from the ahead/behind indicator changing to indicate you can push new commits. I wonder if we should show to confirm the clean merge? Or whether we should keep comparing to the current branch to show that your "behind" commits are now empty? Something to indicate that you were successful here would be 

@shiftkey, I actually have a stash where I started working to use the merge branch popup to handle the merge, that should fix the lack of indication that the merge was successful. It should also give us the merge conflict dialog for free.

shiftkey:
There's another workflow here that I wanted to put forward to source ideas for:

 - I create a branch `open-should-specify-full-path` to fix an issue
 - I push the branch to the remote and open PR #4381
 - Some time later, this gets approved and merged into `master`
 - I update `master` locally to get the changes
 - I switch back to `open-should-specify-full-path` and it's not clear to me that this branch has been merged into `master` and can be cleaned up

Here's what I see that is confusing me a bit:

I see the `Publish branch` hint here because the upstream ref has been removed (`git status` also tells me this). And because the PR is no longer open, the branch doesn't show the PR badge any more:



When I compare to `master` again, I see there are a bunch of new commits marked as "behind" - including the merge commit for this PR at the top:



There are also no commits "ahead" for this branch because the work has been merged in:

screen shot 2018-04-06 at 1 54 26 pm

We're already tracking "prune merged branches" as a background task in #750, but I wanted to see if we could use this available space to help the user:

 - explain that this "ahead" panel is empty because the branch you're comparing to (which is also the default branch) now has the commits from this branch. In Git terminology:

```
$ git branch --contains HEAD | cat
  master
* open-should-specify-full-path
```

 - help guide the user to clean up this branch locally as part of ending the flow of "create branch" -> "make commits" -> "create pull request" -> "pull request merged" -> "delete branch"

app/src/ui/lib/filter-list.tsx
```diff
@@ -117,6 +117,9 @@ interface IFilterListProps<T extends IFilterListItem> {
 
   /** Called when there are no items to render.  */
   readonly renderNoItems?: () => JSX.Element | null
+
+  /** Used to get a reference to a TextBox that will be used to control this component */
+  readonly filterTextBox?: TextBox //Todo: don't rely on passing textbox to component
```
nerdneha: Maybe I'm opening a can of worms here but I wonder if we can avoid a lot of these "if" statements and the extra component for `FancyTextBox` by adding an `icon` prop to the `TextBox` that can be null or set to something else (ex: the octicon). That way we don't have to pass in an override or deal with finding a creative way to `selectAll()` on either a `TextBox` or a `FancyTextBox`. 

I have a feeling I'm missing some nuance here, though.

app/src/ui/lib/filter-list.tsx
```diff
@@ -220,6 +234,20 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
     )
   }
 
+  public selectFirstItem(focus: boolean = false) {
+    if (this.list !== null) {
+      const next = this.list.nextSelectableRow('down', -1)
```
nerdneha: i saw that we use `this.list.nextSelectableRow('down', -1)` three times in this file to get the first item. do we want to extract that to something that's easy to recognize? like `firstRow()`

app/src/ui/branches/branches-container.tsx
```diff
@@ -91,6 +97,13 @@ export class BranchesContainer extends React.Component<
     )
   }
 
+  private renderBranch = (
```
nerdneha: Because we're not using any props or local variables for the renderBranch method, we can probably import `renderDefaultBranch` and use it inside the `BranchList` component. Feels like there's nothing unique we're doing at this higher level.

app/src/ui/lib/filter-list.tsx
iAmWillShepherd: That was the original idea, but I felt it was not good to overload the `TextBox` component to do more than it does now. I'm have no strong opinions either way. One thing to keep in mind is that I created #4317 so that this can be refactored.

app/src/ui/history/compare.tsx
```diff
@@ -0,0 +1,477 @@
...
+        }
+
+        if (this.textbox) {
+          this.textbox.blur()
+        }
+      }
+    } else if (key === 'Escape') {
+      this.handleEscape()
+    } else if (key === 'ArrowDown') {
+      console.log('arrow is down')
```
nerdneha: let's take this `console.log` out

app/src/ui/history/compare.tsx
```diff
@@ -0,0 +1,477 @@
...
+  }
+
+  private renderCommitList() {
+    const compareState = this.props.compareState
+    const selectedCommit = this.state.selectedCommit
+    const commitSHAs = compareState.commitSHAs
+
+    const emptyListMessage =
+      compareState.formState.kind === ComparisonView.None
+        ? 'No history'
```
nerdneha: do we want to change the casing based on mac v windows or is that only for buttons and menu items?

app/src/ui/history/compare.tsx
```diff
@@ -0,0 +1,477 @@
...
+    const count = formState.aheadBehind.behind
+
+    if (count > 0) {
+      const pluralized = count === 1 ? 'commit' : 'commits'
+      return (
+        <div className="merge-message">
+          This will merge
+          <strong>{`${count} ${pluralized}`}</strong>
+          {` `}from{` `}
+          <strong>{branch.name}</strong>)
```
nerdneha: I think this is the extraneous `)` that @meaghanlewis found. saw it on my computer, too. let's remove it

app/src/ui/history/compare.tsx
```diff
@@ -0,0 +1,478 @@
...
+      this.textbox.focus()
+    }
+  }
+
+  public render() {
+    const formState = this.props.compareState.formState
+    const placeholderText =
+      formState.kind === ComparisonView.None
+        ? __DARWIN__
+          ? 'Select Branch To Compare...'
```
nerdneha: We're not 💯 consistent but usually small conjunction words like "in" and "to" and "and" are lowercase (cross-checked with the menu bar). If we follow suit, this should be:
```js
? 'Select Branch to Compare...'
```

app/src/ui/lib/filter-list.tsx
nerdneha: 👍

nerdneha:
Sometimes when I switch to a new branch and go to compare mode, the ahead/behinds take a while to load (even though we already calculate them after we click on them). Is this intended behavior? See gif.
ahead-behind-not-update

nerdneha:
Found myself trying to clear the textbox without using my escape button and couldn't figure it out. No matter what I do, the text keeps coming back. Here's a (dramatic) reenactment. Is this intended behavior? cc @donokuda @niik 
remove-text-from-textbox

nerdneha:
🥝

app/src/lib/stores/helpers/ahead-behind-updater.ts
```diff
@@ -0,0 +1,116 @@
...
+  private executeTask = (
+    from: string,
+    to: string,
+    callback: (error: Error | null, result: IAheadBehind | null) => void
+  ) => {
+    if (this.comparisonCache.has(from, to)) {
+      return
+    }
+
+    const range = `${from}...${to}`
```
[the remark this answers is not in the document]
shiftkey: I've opened #4507 to address the `${to}...${from}` syntax.
We can extend this to the `...${from}` syntax later.

app/src/ui/history/compare-branch-list-item.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export class CompareBranchListItem extends React.Component<
+  ICompareBranchListItemProps,
+  {}
+> {
+  public constructor(props: ICompareBranchListItemProps) {
+    super(props)
+
+    this.state = {
```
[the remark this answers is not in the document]
shiftkey: Nice spot, tidied up in 63db804fda7bb84716b0781b08aa9fcbeb3a1ac4

app/src/lib/stores/app-store.ts
```diff
@@ -419,6 +431,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
         currentPullRequest: null,
         isLoadingPullRequests: false,
       },
+      compareState: {
+        formState: { kind: ComparisonView.None },
+        commitSHAs: [],
+        aheadBehindCache: new ComparisonCache(),
+        allBranches: [],
```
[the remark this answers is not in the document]
shiftkey: This is being pulled in from `branchesState` when we initialize the compare:

https://github.com/desktop/desktop/blob/63db804fda7bb84716b0781b08aa9fcbeb3a1ac4/app/src/lib/stores/app-store.ts#L732-L751

It's only really for the branches list inside the compare tab (`_executeCompare` doesn't depend on it).

The benefit of this duplication means that we can pass just `ICompareState` into the `CompareSidebar` component, rather than also passing in `IBranchesState`.

I'd be :thumbsup: to revisiting this after shipping and see if we can simplify this area.

app/src/lib/stores/app-store.ts
```diff
@@ -664,6 +696,138 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+    const compareState = state.compareState
+
+    const cachedState = compareState.formState
+
+    const action = initialAction ? initialAction : getInitialAction(cachedState)
+    this._executeCompare(repository, action)
+
+    if (currentBranch != null && this.currentAheadBehindUpdater != null) {
+      let allOtherBranches = [...recentBranches, ...allBranches]
```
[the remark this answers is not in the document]
shiftkey: Yep, and we don't need to do any of this work because `allBranches` is already doing the  heavy lifting to exclude the current branch:

https://github.com/desktop/desktop/blob/34ece60342d5a0e6e3c7b07a4c1ef5f162fcc391/app/src/lib/stores/app-store.ts#L744-L747

Tidied up in 49debc372

app/src/lib/stores/app-store.ts
```diff
@@ -664,6 +696,138 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      )
+
+      return
+    }
+
+    const updater = new AheadBehindUpdater(repository, aheadBehindCache => {
+      this.updateCompareState(repository, state => ({
+        aheadBehindCache,
+      }))
+      this.emitUpdate()
```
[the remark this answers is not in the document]
shiftkey: I left out the `return` from this example because this function is a callback inside the `AheadBehindUpdater`:

https://github.com/desktop/desktop/blob/71e52ea0f3e48c12d127cd008f11d50a9351c27a/app/src/lib/stores/app-store.ts#L710-L715

The other examples use the latter pattern to mash together the right control flow with a void-returning function in more complex code. 

Happy to revisit this outside the PR and see if we can figure out how to be more consistent here.

app/src/ui/history/compare.tsx
shiftkey: This is what the current History view shows on macOS when you have no commits:

app/src/ui/lib/filter-list.tsx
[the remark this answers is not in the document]
shiftkey: I'll let @iAmWillShepherd chime in on this, but I think this can be removed.

app/src/ui/history/compare.tsx
```diff
@@ -0,0 +1,475 @@
...
+
+    return <div className="merge-message">There are no commits</div>
+  }
+
+  private onTabClicked = (index: number) => {
+    const formState = this.props.compareState.formState
+
+    if (formState.kind === ComparisonView.None) {
+      // the tab control should never be shown in this case
+      // TODO: can we enforce this with TYPES?
```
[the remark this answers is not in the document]
shiftkey: :fire:

nerdneha:
@shiftkey 
> @nerdneha because the ahead/behind stats work is proportional to the number of local and remote branches in a repository, the work to compute branches is queued up serially using requestIdleCallback - see #4399 for more context.

after reading the code, that made sense; in fact that wasn't the part that felt off when i was testing; it was the fact that the component was calculating the `ahead`/`behind` when i clicked on a `compare-branch` and then it felt like the `branch list component` hadn't "remembered". programmatically i get that it's queued and it'll show up soon but still felt weird as a user.

shiftkey:
>  it was the fact that the component was calculating the ahead/behind when i clicked on a compare-branch and then it felt like the branch list component hadn't "remembered". 

Right, now I'm with you. Mind opening a new issue and indicate that selecting a branch (which computes the ahead/behind stats) should also push that into the cache (which should then refresh the list the next time you switch to it)?

nerdneha:
feeling really good about this feature and solid updates based on code review! 🎈 

IMO given that @donokuda opened up a PR for some of the visual pieces and that we have an open PR for updating the way we use `FilterList`, I don't think there's anything that's blocking a merge. (although I'd like to get rid of that TODO if there's nothing that really needs it). 

things i tested:
 * comparing to a branch with 0 behind and X ahead
 * comparing to a branch with X behind and 0 ahead
 * comparing to a branch with X behind and Y ahead
    * merging successfully + undoing
    * merging that resulted in merge conflicts, undoing the commit that caused conflict, discarding changes, and successfully merging
* comparing with many branches back-to-back
* filtering for branches in different ways 
* filtering on something not found to see "no branches" view
* trying to merge when there's nothing to merge
* branching from a branch + then merging back into that branch
* merging between two branches
* merging master into something with >2k commits

it's got a 👍 from me. ccing @niik in case there's anything he wants to comment on

saiKumarGanji:
When will the exe file will be released which will have this feature

shiftkey:
@saiKumarGanji once #4545 is merged and we publish an update to the beta channel it will be available there

jonnyijapan:
This feature shows "only" differing commits, and then the changed files in those specific diffs. It would be helpful if Github Desktop could show the summarized diff of all the files in all of the differing commits. Like the diff when using Github.com in a pull request etc.

shiftkey:
> Sorry for a late "request".

@jonnyijapan if this is something that is still of interest to you, please open a [suggestion](https://github.com/desktop/desktop/issues/new?template=problem-to-raise.md) so we can track it separate to the original work.

Issue #4180, opened 2018-03-05
Discard changes (for a single change) is very slow


kmscode:
## Version

* GitHub Desktop: 1.1.0

* Operating system: Windows 10 Pro 1709

### Expected Behavior
Change should be discarded without delay.

### Actual Behavior
Change is discarded, but it takes about 3 seconds for the popup to go away. This makes the UI feel very sluggish. 

## Additional Information

Not sure if this is just a UI issue where the popup needs to be destroyed, or if this is related to the backend / electron APIs for sending the file to the recycle bin as noted in #2629

Desktop Client system hardware specs:
Dell Precision 7510
[Core i7-6920HQ](https://ark.intel.com/products/88972/Intel-Core-i7-6920HQ-Processor-8M-Cache-up-to-3_80-GHz)
32 Gigs of RAM
512 Gig SSD

Repo in that I've been using most recently is GitHub Enterprise. Server has 2 ms ping times and is 4 hops away on a very fast wired/gigabit corporate network:
```
> ping [REPO HOSTNAME]

Pinging [REPO HOSTNAME] [[REPO IP]] with 32 bytes of data:
Reply from [REPO IP]: bytes=32 time=2ms TTL=61
Reply from [REPO IP]: bytes=32 time=2ms TTL=61
Reply from [REPO IP]: bytes=32 time=2ms TTL=61
Reply from [REPO IP]: bytes=32 time=2ms TTL=61

Ping statistics for [REPO IP]:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 2ms, Maximum = 2ms, Average = 2ms
> tracert [REPO HOSTNAME]

Tracing route to [REPO HOSTNAME] [REPO IP]
over a maximum of 30 hops:

  1    <1 ms    <1 ms    <1 ms  [HOP1] [HOP1 IP]
  2     2 ms     3 ms     1 ms  [HOP2] [HOP2 IP]
  3    <1 ms    <1 ms    <1 ms  [HOP3] [HOP3 IP]
  4     2 ms     2 ms     2 ms  [REPO HOSTNAME] [REPO IP]
```

<sub><sup>Running some tests for some build scripts that will ultimately change some files and check them in. For now as I'm tweaking I need to discard the changed and it's super annoying how slow it is...</sup></sub>

### Logs

[...]

PR #4186, opened 2018-03-06
Check for submodules before getting status

kmscode:
Addresses the slowness as reported in #4180

### DIFF
--- a/app/src/lib/git/submodule.ts
+++ b/app/src/lib/git/submodule.ts
@@ -8,13 +8,13 @@ export async function listSubmodules(
   repository: Repository
 ): Promise<ReadonlyArray<SubmoduleEntry>> {
 
-	let submodulesFile = await pathExists(Path.join(repository.path, '.gitmodules'))
-	let submodulesDir = await pathExists(Path.join(repository.path, '.git', 'modules'))
+  const submodulesFile = await pathExists(Path.join(repository.path, '.gitmodules'))
+  const submodulesDir = await pathExists(Path.join(repository.path, '.git', 'modules'))
 
-	if(!submodulesFile && !submodulesDir) {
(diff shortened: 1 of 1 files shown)


shiftkey:
@kmscode I recommend just running `yarn lint:fix` locally as that should format your code how Prettier prefers.

kmscode:
@shiftkey Thanks!

I found prittier-atom before I saw your comment. I think that should do the trick as well. Will find out in a few minutes.

app/src/lib/git/submodule.ts
```diff
@@ -1,10 +1,24 @@
...
+import { pathExists } from '../file-system'
+import * as Path from 'path'
 
 export async function listSubmodules(
   repository: Repository
 ): Promise<ReadonlyArray<SubmoduleEntry>> {
+  const submodulesFile = await pathExists(
+    Path.join(repository.path, '.gitmodules')
+  )
+  const submodulesDir = await pathExists(
```
niik: Could we run these two checks in parallel rather than sequentially please?

```ts
const [ submodulesFile, subModulesDir ] = await Promise.all([
  pathExists(Path.join(repository.path, 'gitmodules'),
  pathExists(Path.join(repository.path, 'git', 'modules'),
])
```

kmscode: Solid recommendation. Thanks! Done!

niik:
Great work @kmscode! I'm gonna let this sit for a little while just so that everyone on the team in their various time zones has a chance to see it. Since the weekend is coming up it's likely we won't actually merge it until next week.

kmscode:
@niik Thanks for the status update. That sounds fair enough. Glad I could be of some help! 😄

Issue #4189, opened 2018-03-06
Multiple issues related to SSL Interception (environs with proxy, enterprise MITM, etc)


DeepMac:
## Description

GitHub Desktop exhibits issues with authentication and repository cloning when SSL interception is present, such as Enterprise MITM configurations (Proxies and other cybersecurity products).

1) Logging into GitHub account via username and password within the GUI gives an error indicating the username and password is incorrect when they are correct. Logging in via browser still works. But error message is mis-leading and could trigger user interaction loops (reset password, try new password, reset it again, etc).

2) Cloning a repository throws cryptic errors about SChannel. Documented in #3326 I suspect this is at least partially related to interception, as switching to OpenSSL can still throw errors because of expected certificate signatures not matching.

3) Instructions for the workaround here are inconsistent and don't seem to work. Do I use http.CAInfo, http.CAFile or http.CAPath? Do I specify the path in Windows notation (C:\foo\bar) or the Git bash shell form (/c/foo/bar)? Regardless of what I set, I could not get this solution to work. Instead, I had to issue a "git config --global http.sslVerify false" so I could use the openSSL backend and get around the problems with accessing the cert file.

4) Only that didn't actually work. Because now I get "fatal: .git/index: index file open failed: Permission denied". Or hey maybe it did, the files are there. But I just wasted 1.5 hours on this.

## Version

* GitHub Desktop: 1.1.0

* Operating system: Microsoft Windows 10.0.14393

## Steps to Reproduce

Reproduction would require access to a restricted corporate environment.
Recommend setting up a lab and configuring an Enterprise grade security product such as BlueCoat or similar that does MITM, then using internal certificates for the pass-back.

### Expected Behavior

1) Sign-in directly with no error, or get a message indicating issues communicating with auth server.

2) Clone the repo without any hassle.

3) Should never get here but if we do, shouldn't have confusing workarounds that don't work.

### Actual Behavior

Augh. See above. I'm hating this form now.

### Logs

[2018-03-06.desktop.production.log](https://github.com/desktop/desktop/files/1786763/2018-03-06.desktop.production.log)

DeepMac:
I'd like to suggest these and potential other failures related to this sort of environment be addressed during install. Detect the existence of MITM, alert the user. If they verify they are running in a restricted corporate environment, activate a "MITM Mode" where all configurations related to checking cert validity, revocation, etc are disabled. Optionally check for local CA's in the Windows cert storage.

shiftkey:
[...]

PR #4190, opened 2018-03-06
make this workaround clearer

shiftkey:
See discussion here: https://github.com/desktop/desktop/issues/4189#issuecomment-370935990

cc @DeepMac - does this help?


DeepMac:
Yep! Thanks @shiftkey

PR #4192, opened 2018-03-07
Delete gitignore file when no entries provided

### LINKED ISSUE #1896: Deleted Git Ignore files are persistent in Changes
In a new repo, if a user alters a blank git ignore file, but decides to delete those changes before committing the file, then the file should not show in the Changes column as nothing was really changed.

v0.5.9, OSX 10.11.5

Reproduce:
- (Could only reproduce with a fresh File > New Repo, push to dotcom)
- Repository > Repository Settings > Ignored Files
- Enter is some text, Save, git-ignore file appears in Changes
- Go back to Ignored Files and remove the text
- Changes file is still there, with blank line

Expected: git-ignore file should be removed from Changes (For existing repos with git-ignore files, this is working okay)
Actual: git-ignore file remains with a blank line item

screen shot 2017-06-05 at 11 17 34 am

Note the cursor location. It won't save over the first "line" and ultimately allow the placeholder text to appear.
screen shot 2017-06-05 at 11 18 07 am

lee-dohm:
Fixes #1896 

I tested the following manually:

1. Use the Repository Settings dialog to create a `.gitignore` file
1. Use the Repository Settings dialog to delete the `.gitignore` file by deleting the contents
1. Create the file with one line without committing it via the command line, then delete it via the Repository Settings dialog by emptying it

The only scenario that I could find that wasn't covered was to create the `.gitignore` file as empty via `touch .gitignore`, then open the Repository Settings dialog and attempt to delete it. Because the file contents don't change, it doesn't delete the file. Do we want to cover this case?

### DIFF
--- a/app/src/lib/stores/repository-settings-store.ts
+++ b/app/src/lib/stores/repository-settings-store.ts
@@ -49,17 +49,29 @@ export class RepositorySettingsStore extends BaseStore {
   public async saveGitIgnore(text: string): Promise<void> {
     const repository = this._repository
     const ignorePath = Path.join(repository.path, '.gitignore')
-    const fileContents = await formatGitIgnoreContents(text, repository)
 
-    return new Promise<void>((resolve, reject) => {
-      FS.writeFile(ignorePath, fileContents, err => {
-        if (err) {
-          reject(err)
-        } else {
-          resolve()
-        }
+    if (text === '') {
+      return new Promise<void>((resolve, reject) => {
+        FS.unlink(ignorePath, err => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve()
+          }
+        })
       })
-    })
+    } else {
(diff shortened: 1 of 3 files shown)


tierninho:
✅ tested the functionality. I'll defer to @shiftkey for the technical review.

app/src/lib/stores/repository-settings-store.ts
```diff
@@ -49,17 +49,29 @@ export class RepositorySettingsStore extends BaseStore {
...
+        FS.unlink(ignorePath, err => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve()
+          }
+        })
       })
-    })
+    } else {
```
shiftkey: We can drop the `else` here to avoid a bit of unnecessary nesting...

PR #4198, opened 2018-03-07
Add the ability to build arm64 builds

jkleinsc:
This PR adds the ability to build arm64 builds of Desktop.

There are a couple of caveats:
1. It requires Electron 1.8.x because arm64 builds were not available in Electron before 1.8.x
2. The build needs to be run on arm64 hardware or on x86 hardware via cross arch docker containers (eg via [multiarch/debian-debootstrap:arm64-jessie](https://hub.docker.com/r/multiarch/debian-debootstrap/))

To test this PR, I used Electron's arm64 CI here: 
https://arm64-ci.electronjs.org/blue/organizations/jenkins/desktop/detail/desktop/17/pipeline


zeke:
💪6️⃣:four:

shiftkey:
@jkleinsc thanks for the PR! 

> It requires Electron 1.8.x because arm64 builds were not available in Electron before 1.8.x

We've got #3958 underway which we'll get in and test before this - I gather should we be testing `1.8.3` as well, given that just landed earlier this week (ping @iAmWillShepherd).

niik:
Hey @jkleinsc #3958 has been merged so this PR needs a little merge from master.

jkleinsc:
@niik I rebased.  Arm64 CI running here: https://arm64-ci.electronjs.org/blue/organizations/jenkins/desktop/detail/desktop/18/pipeline

iAmWillShepherd:
@zeke , @jkleinsc, I'd like to test on an actual device. Am I correct in assuming I can use a Raspberry Pi to test this?

jkleinsc:
@iAmWillShepherd you would need a Raspberry PI 3 and it needs to be running a 64 bit OS.  Another alternative is a pine64 from https://www.pine64.org/

app/package.json
```diff
@@ -25,7 +25,7 @@
     "codemirror-mode-elixir": "1.1.1",
     "deep-equal": "^1.0.1",
     "dexie": "^2.0.0",
-    "dugite": "1.58.0",
+    "dugite": "1.60.0",
```
shiftkey: This is important because `dugite` now has ARM64 support, correct? If so, we should run though the various Git operations in the app as part of the review.

jkleinsc: @shiftkey yes, I had to upgrade dugite to 1.60.0 for arm64 support.

jkleinsc:
@iAmWillShepherd if you want compiled binaries, I updated the arm64 ci build to archive the dist directory: https://arm64-ci.electronjs.org/blue/organizations/jenkins/desktop/detail/desktop/20/artifacts

shiftkey:
Moving this into 1.2 as it doesn't _seem_ urgent to merge as part of a bugfix release.

nerdneha:
@shiftkey so we're not holding off a merge for this feature, what are your thoughts on us adding a feature flag for this instead? that way we can merge (we can even default it to false for dev if we're worried)

shiftkey:
@nerdneha I don't believe this needs to be feature flagged, I'm just deferring it to after 1.1.x.

Aside from the update to Git (which we should test like usual) the other changes look to be safe to include as-is.

shiftkey:
Dropping myself from the review process because I've contributed to this PR.

Dockerfile.arm64v8
```diff
@@ -0,0 +1,42 @@
+FROM arm64v8/ubuntu:16.04
+
+RUN groupadd --gid 114 jenkins \
+  && useradd --uid 110 --gid jenkins --shell /bin/bash --create-home jenkins
```
niik: Is this typical? Could the user be named `desktop` or something?

jkleinsc:
@niik I have usually seen Dockerfiles at the root of a project, but I'm thinking that the Dockerfile doesn't really belong in this PR.  It is not needed to build for arm64, it is the Dockerfile I used to test on https://arm64-ci.electronjs.org/blue/organizations/jenkins/desktop.  That's why it has a jenkins user in it.  I'll remove that file and I'll add a blurb about how to build on arm64.

jkleinsc:
@niik I removed the file and added the documentation.  Let me know if the doc additions make sense.  Also, I realized in putting together the docs that the option of building the arm64 build through a cross architecture docker image (eg https://hub.docker.com/r/multiarch/debian-debootstrap/) doesn't work.  The build needs arm64 hardware.

niik:
@jkleinsc Lovely, thank you so much for going the extra mile and adding that documentation. I'm gonna let this sit for a little bit so that all TZs have a chance to see it before I merge.

zeke:
🍏

PR #4216, opened 2018-03-10
Trim leading and trailing whitespace

### LINKED ISSUE #4213: Trim whitespace from start and end of URLs pasted into "Clone repository"
## Description

Please trim whitespace from start and end of URLs pasted into "Clone repository" under the "URL" tab. When copying and pasting a URL it is easy to copy extra whitespace by mistake and GitHub desktop gives the following error:

fatal: I don't handle protocol ' http'

## Version

* GitHub Desktop: 1.1.0

* Operating system: Windows 10 / Windows 7

## Steps to Reproduce

1. Copy URL for a git repository with some whitespace before the "http"
2. Paste URL into "Clone repository" / "URL"
3. Click "Clone" button

### Expected Behavior

GitHub desktop should trim whitespace from start and end of URL before attempting to clone repo
### Actual Behavior

Error message is displayed:

fatal: I don't handle protocol ' http'

iAmWillShepherd:
Fixes #4213 

This PR also prevents leading and trailing white-space from being entered.

### Before
before

### After
after


j-f1:
@shiftkey I feel like it should stay in `onChange` — we don’t want people typing spaces in their URLs, either.

niik:
> @iAmWillShepherd while this does fix the issue, I wonder if we could handle the paste explicitly, rather than hijacking onChange.

> @shiftkey I feel like it should stay in onChange — we don’t want people typing spaces in their URLs, either.

I'm gonna jump in here and pile on even more opinions. I don't think this should be a user-facing thing at all. Preventing user input without any indication of why isn't a very common pattern, the closest thing would be a max length input which isn't what's happening here. Also, not binding the actual input value back immediately will be treated by the browser/react as resetting the input value meaning that cursor position is also reset.

I'd prefer we trim once we pass the string off to the actual logic for cloning and that we add a check to not enable the clone button if the input is all whitespace.

nerdneha:
> I'd prefer we trim once we pass the string off to the actual logic for cloning and that we add a check to not enable the clone button if the input is all whitespace.

i'm ➕1️⃣ on @niik's comment. removing myself from being assigned while we decide what to do.

FYI, right now, if i put in a bunch of spaces into the input so that the input is all whitespace, the error message is `The destination already exists` which is weird to me. Removing the whitespaces does not remove the error, the error persists until i type something that could theoretically exist. see gif:
clone-repo-whitespace

nerdneha:
removing the `ready-for-review` flag since we have some discussing to do first

shiftkey:
> I'd prefer we trim once we pass the string off to the actual logic for cloning and that we add a check to not enable the clone button if the input is all whitespace.

This is also fine

iAmWillShepherd:
🔢

iAmWillShepherd:
I'm now trimming `path` and `url` inside of `cloneImpl()`. I've also gone ahead and added a new error conditions for whitespace and invalid URLS (we should consider using [url-regex]() instead of my naive solution of relying on the return type of `parseRepositoryIdentifier()`.

iAmWillShepherd:
🔢

niik:
I appreciate the extra effort here to try and address the weak areas of the clone dialog but I think what would be the best use of our time right now would be to get back to the original intention of this PR, i.e. trimming leading and trailing white space before handing it of to the clone implementation.

As the PR stands right now it classifies valid repository urls as invalid because the parse function is made specifically for GitHub-formatted repository urls and there's some UX peculiarities that would need to be looked at.

Rather than try to address that I think we should steer clear of trying to validate the format of the clone urls all together. Git [supports formats](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a) which aren't technically well-formatted urls (`git@github.com:foo/bar`) so we should just leave that to Git (which is what we're doing today) to deal with and produce an error message if it considers it invalid.

As for the case of only white space in the repo url input I don't think it warrants a dialog error, it should be fairly obvious that the reason you can't hit the clone button is because you haven't filled out the path.

So, can we either get this PR back to its initial intention i.e. only trimming leading and trailing white space and doing so before we pass the url off to the actual logic or put this fix on the back burner (i.e. close it out) while we figure out 1.2 and then get back to it afterwards?

iAmWillShepherd:
🔢

iAmWillShepherd:
@niik no problem at all. It only took a revert! 👍

Issue #4132, opened 2018-02-28
Enable autofocus on 2FA login window


tierninho:
As part of the login process (via Preferences), the auto focus is missing from the 2FA window. 

> I think adding autoFocus={true} to the right textbox should be easy enough to address that.

![36690352-e86b1114-1ae6-11e8-84a7-972559e0a38b](https://user-images.githubusercontent.com/14828183/36805662-da6c5aea-1c72-11e8-8d94-ddfdf4c5c11a.png)

related to: https://github.com/desktop/desktop/issues/4109

shiftkey:
This textbox is fine:

https://github.com/desktop/desktop/blob/aebd9d8759c0b9894ecd6c03fdcdcc4364244b3f/app/src/ui/lib/two-factor-authentication.tsx#L78-L83

This textbox is not:

https://github.com/desktop/desktop/blob/a70d7c5b1321bbc55e1d84e8c6467ae289a5db6d/app/src/ui/sign-in/sign-in.tsx#L236-L242

shiftkey:
@lukePeavey go for it!

iAmWillShepherd:
@lukePeavey , When you open your PR, can you be sure to include this line at the top of your PR body? `Fixes issue #4231`

shiftkey:
@iAmWillShepherd is there a particular reason why you need to add `issue` as well? That seems to interfere with the existing `Fixes #{id}` syntax which auto-closes the issue when a PR is merged.

lukePeavey:
@iAmWillShepherd 
> can you be sure to include this line at the top of your PR body?

Sure, no problem.

iAmWillShepherd:
> @iAmWillShepherd is there a particular reason why you need to add issue as well? That seems to interfere with the existing Fixes #{id} syntax which auto-closes the issue when a PR is merged.

@shiftkey  Nope, updated comment 

@lukePeavey please use `Fixes #4231` at the top of your PR body instead and sorry for the confusion.

lukePeavey:
Ok

lukePeavey:
@shiftkey about the delay on this. I found that auto focus still wasn't working consistently after adding `autoFocus=true` to the `TextBox`. So i was trying to figure out what was causing that before submitting my PR.

shiftkey:
@lukePeavey no problem - we're working on carving out other tasks that contributors can help out with, so keep an eye on the issue tracker for other things that you may be able to help with!

PR #4220, opened 2018-03-11
added autoFocus=true for 2FA textBox

WaleedAshraf:
Fixes https://github.com/desktop/desktop/issues/4132

**Changes:** added autoFocus=true for 2FA textBox


WaleedAshraf:
@shiftkey Sure, np 👍 🎉

PR #4227, opened 2018-03-12
Bump toolbar contrast

### LINKED ISSUE #4219: Improve text contrast and make the UI elements more visible
## Description

The text in the desktop application has a very low contrast. The menu bar is almost invisible. Please see the snapshot below in Additional Information section. Also, see the highlighted part where there is grey text on black background. It makes it really hard to read. Also, we should develop keeping in mind those who don't have perfect vision. There are no themes for high-contrast which makes.

## Version
* GitHub Desktop: 1.1.0
* Operating system: Windows 10, Version 1709.

## Steps to Reproduce
1. Open the desktop app.
2. Try to read any text.
3. Compare it with reading text in a simple text editor
### Expected Behavior
Expectation is to be able to read the text and find menu bar easily.
### Actual Behavior
Difficult to read text and other UI elements like menu bar and close, minimize and maximize buttons.

## Additional Information

image

niik:
Fixes #4219.

Bumps the text secondary color for the toolbar one notch from gray-400 to gray-300. Changes the menu item text from the secondary to the primary color (white) to match the visual weight of the primary toolbar button text.

### Before and after

win-merged

@donokuda Please try this out. Initially I felt like going to 300 would make the description and title elements in the toolbar button too similar but after having run it for a little while I can't say that I do any longer. If you feel like the change is too strong we should mix our own gray-350.

### DIFF
--- a/app/src/ui/index.tsx
+++ b/app/src/ui/index.tsx
@@ -149,14 +149,15 @@ document.body.classList.add(`platform-${process.platform}`)
 dispatcher.setAppFocusState(remote.getCurrentWindow().isFocused())
 
 ipcRenderer.on('focus', () => {
-  dispatcher.setAppFocusState(true)
-
   const state = appStore.getState().selectedState
-  if (!state || state.type !== SelectionType.Repository) {
-    return
+
+  // Refresh the currently selected repository on focus (if
+  // we have a selected repository).
+  if (state && state.type === SelectionType.Repository) {
+    dispatcher.refreshRepository(state.repository)
   }
 
-  dispatcher.refreshRepository(state.repository)
+  dispatcher.setAppFocusState(true)
 })
 
 ipcRenderer.on('blur', () => {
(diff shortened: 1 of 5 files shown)


app/src/ui/index.tsx
```diff
@@ -150,12 +150,14 @@ dispatcher.setAppFocusState(remote.getCurrentWindow().isFocused())
 
 ipcRenderer.on('focus', () => {
   const state = appStore.getState().selectedState
```
nerdneha: while this isn't part of your changed lines, how do you feel about us renaming `state` to `selectedState`, ex:
```
  const { selectedState } = appStore.getState()

  // Refresh the currently selected repository on focus (if
  // we have a selected repository).
  if (selectedState && selectedState.type === SelectionType.Repository) {
    dispatcher.refreshRepository(selectedState.repository)
  }
```
i'd prefer to call it selectedState just so we don't mistake it for having all the state info

niik: I love it

Issue #2121, opened 2017-06-26
Branches button is disabled while on an unborn branch


tierninho:
If you create a second branch, and then remove the initial commit before publishing, you cannot go back to the master branch. The UI of the commit button says `master` while the branch is locked on X. 

v0.6.2, OSX 10.11.5

Reproduce:
- File > New Repo, follow steps but don't publish
- Create new branch, switch to new branch
- Undo initial commit

Expected result: User can switch branches back to `master`, and `commit button` and branch are matching.

<img width="739" alt="screen shot 2017-06-26 at 1 20 21 pm" src="https://user-images.githubusercontent.com/14828183/27551921-5cd9a892-5a73-11e7-8674-80143e022a1f.png">

OnclickSell:
I think it is because the user is not identified by the GitHub client. run these commands inside your current repository if worked:  git config --global user.email "you@example.com" 
git config --global user.name "Username"

shiftkey:
I can reproduce this in 1.1.0, and it's related to the undo commit step at the very end. Here's what Git reports after that:

```
~/Documents/GitHub/doooot/ git status
On branch doot

No commits yet

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.gitattributes
	.gitignore
	LICENSE

nothing added to commit but untracked files present (use "git add" to track)
```

And the branch list contains `master` only:

```
~/Documents/GitHub/doooot/ git branch
  master
```

The reason the UI is ending up in this state is "No commits yet" is what we consider an unborn repository, and `canOpen` is set to false in here:

https://github.com/desktop/desktop/blob/a0172aaa541c0118162987d15dd9199ff1d6e680/app/src/ui/toolbar/branch-dropdown.tsx#L104-L107

In theory, we do a more relaxed check of this:

```diff
diff --git a/app/src/ui/toolbar/branch-dropdown.tsx b/app/src/ui/toolbar/branch-dropdown.tsx
index 5238d4091..3c96169f4 100644
--- a/app/src/ui/toolbar/branch-dropdown.tsx
+++ b/app/src/ui/toolbar/branch-dropdown.tsx
@@ -104,7 +104,7 @@ export class BranchDropdown extends React.Component<IBranchDropdownProps> {
     } else if (tip.kind === TipState.Unborn) {
       title = tip.ref
       tooltip = `Current branch is ${tip.ref}`
-      canOpen = false
+      canOpen = branchesState.allBranches.length > 0
[...]

PR #4236, opened 2018-03-13
Branches button is disabled while on an unborn branch

agisilaos:
Fixes: #2121

As discussed in #2121 this issue isn't "exactly" fixed, but I thought it would be a good idea to kick off this PR since this is a UX issue and I'm so into them. ❤️  

@shiftkey You think we should leave it as it is with the solution you proposed or make any other changes for example in the error message that's appeared after this change etc? Would be more than happy to make any other changes so we're in a really good position to merge this in!


agisilaos:
🐨

agisilaos:
Great, thank you for the detailed comment @niik 👍

tierninho:
@agisilaos Confirming this is now showing the new messaging, as decided.

Issue #4004, opened 2018-02-14
add visual indicator that a folder can be dropped on Desktop


shiftkey:
A discussion in https://github.com/desktop/desktop/pull/3996 indicates we're not exactly providing any visual indication that you can drop a folder on to Desktop (or that there are now times when you can't).

Currently, we have these two event handlers setup in `app.tsx`:

```ts
    document.ondragover = document.ondrop = e => {
      e.preventDefault()
    }

    document.body.ondrop = e => {
      if (this.state.currentPopup != null) {
        return
      }

      const files = e.dataTransfer.files
      this.handleDragAndDrop(files)
      e.preventDefault()
    }
```

Looking at the MDN docs, [this link](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect) suggests that we could do something like this inside the `dragover` event handler: 

```js
e.dataTransfer.dropEffect = 'none';
```

### What To Do

Investigate what happens when split up these event handlers to look like the below snippet. I haven't tested this code, so you might find something unexpected.

```ts
    document.ondragover = e => {
      if (this.state.currentPopup != null) {
        e.dataTransfer.dropEffect = 'none'
      } else {
        e.dataTransfer.dropEffect = 'copy'
      }
 
      // this seems important, we should keep it around
      e.preventDefault()
    }

    document.ondrop = e => {
      e.preventDefault()
    }
```

There's two scenarios to test here with this change:

 - whether the cursor changes when you drag a folder onto Desktop (to indicate it can be added)
 - whether the cursor also changes when a modal is displayed (to indicate that it is being ignored)

There's also probably some cross-platform testing to do here, but we can do that at review time if you don't have access to multiple OSes.

agisilaos:
Currently working on this @shiftkey, opening a PR soon.

agisilaos:
@shiftkey Jus opened a PR #4239

Issue #4068, opened 2018-02-22
optimization: avoid a fetch when checkout out a pull request


shiftkey:
In #4057 @niik mentions an opportunity to avoid a `fetch` if we know the branch exists:

```ts
      // TODO: I think we could skip this fetch if we know that we have the branch locally
      // already. That way we'd match the behavior of checking out a branch.
      if (defaultRemote) {
        await this._fetchRemote(
          repository,
          defaultRemote,
          FetchType.UserInitiatedTask
        )
      }
```

I'll let @niik add any additional thoughts on the best approach here.

agisilaos:
Would be more than happy to pick this up @shiftkey 💯

agisilaos:
@shiftkey Is it possible to report the comment above as spam? because I don't have this option available.
![screen shot 2018-02-24 at 18 41 14](https://user-images.githubusercontent.com/10944108/36632707-5850da78-1992-11e8-94b3-84ceb06fd098.png)

iAmWillShepherd:
@agisilaos. Go for it! I'll add it to our project board for tracking issues our community have claimed.

agisilaos:
@niik Before I start working on this, I would love to know your additional thoughts or any other info that may help me during the process.

iAmWillShepherd:
@agisilaos , when you open your PR, can you be sure to include `Fixes #4068` at the top of the PR body?

agisilaos:
Of course 🌮

niik:
> Before I start working on this, I would love to know your additional thoughts or any other info that may help me during the process.

I believe I was referring to using the same logic that already exists in the forked-case: https://github.com/desktop/desktop/blob/321c3de0f143e29cfaab702dbc6e5766f8d91fda/app/src/lib/stores/app-store.ts#L3323-L3333

So I'd say the work consists of extracting that logic into something reusable and then verifying that my assumption was correct in the first place.

agisilaos:
Just finished working on this one. Will create a PR soon @niik!

agisilaos:
PR #4240 is up! 💯

PR #4239, opened 2018-03-14
Add visual indicator that a folder can be dropped on Desktop

agisilaos:
Fixes: #4004 

Since this is another UX related kind of issue I thought I will chime in and start working on it.

@shiftkey Here's the results of the tests cases that you posted in #4004:

> There's two scenarios to test here with this change:
> whether the cursor changes when you drag a folder onto Desktop (to indicate it can be added)

This seems to be the case. Now when you drag a folder into Github Desktop a plus icon appears near the folder (macOS)

> whether the cursor also changes when a modal is displayed (to indicate that it is being ignored)

The folder remains as it is neither the cursor change it's appearance.

Note: I tested the app only in a macOS enviroment so it would be a good idea to test this with a windows machine as well.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -535,7 +535,19 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   public componentDidMount() {
-    document.ondragover = document.ondrop = e => e.preventDefault()
+    document.ondragover = e => {
+      if (this.state.currentPopup != null) {
+        e.dataTransfer.dropEffect = 'none'
+      } else {
+        e.dataTransfer.dropEffect = 'copy'
+      }
+
+      e.preventDefault()
+    }
+
+    document.ondrop = e => {
+      e.preventDefault()
+    }
 
     document.body.ondrop = e => {
       if (this.state.currentPopup != null) {
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -536,7 +536,7 @@ export class App extends React.Component<IAppProps, IAppState> {
 
   public componentDidMount() {
     document.ondragover = e => {
-      if (this.state.currentPopup != null) {
+      if (this.isShowingModal != null) {
         e.dataTransfer.dropEffect = 'none'
       } else {
         e.dataTransfer.dropEffect = 'copy'
@@ -550,7 +550,7 @@ export class App extends React.Component<IAppProps, IAppState> {
     }
 
     document.body.ondrop = e => {
-      if (this.state.currentPopup != null) {
+      if (this.isShowingModal != null) {
         return
       }
       const files = e.dataTransfer.files
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -122,7 +122,7 @@ export class App extends React.Component<IAppProps, IAppState> {
    * modal dialog such as the preferences, or an error dialog.
    */
   private get isShowingModal() {
-    return this.state.currentPopup || this.state.errors.length
+    return this.state.currentPopup !== null || this.state.errors.length > 0
   }
 
   public constructor(props: IAppProps) {
@@ -536,7 +536,7 @@ export class App extends React.Component<IAppProps, IAppState> {
 
   public componentDidMount() {
     document.ondragover = e => {
-      if (this.isShowingModal != null) {
+      if (this.isShowingModal) {
         e.dataTransfer.dropEffect = 'none'
       } else {
         e.dataTransfer.dropEffect = 'copy'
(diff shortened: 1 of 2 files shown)


agisilaos:
🌮

iAmWillShepherd:
@agisilaos I edited your description to remove an e-mail address.

agisilaos:
I was working on issue #4068 and by accident I commited this change in this branch. I reverted the change.

app/src/ui/app.tsx
```diff
@@ -535,7 +535,19 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   public componentDidMount() {
-    document.ondragover = document.ondrop = e => e.preventDefault()
+    document.ondragover = e => {
+      if (this.state.currentPopup != null) {
```
niik: Could we use `this.isShowingModal` here instead so that we handle the case where an error dialog is showing? And when you make that change I'd really appreciate it if you could clean up our old sins and apply the same fix to the condition inside of the `document.body.ondrop` handler as well.

agisilaos: @niik Some questions so I can clarify some things before I dive into the changes you suggested.

> Could we use this.isShowingModal here instead so that we handle the case where an error dialog is showing?

So you say to replace `this.state.currentPopup` with `this.isShowingModal`? or add it to the if statement. I think adding it would be a better "solution" for the reason that we will do more checks, one for the popup & the other for the error dialog but not sure 🤔 

> And when you make that change I'd really appreciate it if you could clean up our old sins

What you mean by that?

tierninho:
@agisilaos I noticed the new indicator disappears when dragging a folder on Desktop and the following snippet is automatically run in the background.

screen shot 2018-03-14 at 11 13 09 am

Any way to have the indicator persist? 

Reproduce:
- Drag folder on Desktop, but do not drop it
- Drag folder outside of Desktop, but do not drop it
- Drag folder back to Desktop and indicator is missing

```debug: [ui] Executing getStatus: git status --untracked-files=all --branch --porcelain=2 -z
debug: [ui] Executing getBranches: git for-each-ref --format=%(refname)%00%(refname:short)%00%(upstream:short)%00%(objectname)%00%(author)%00%(committer)%00%(parent)%00%(symref)%00%(subject)%00%(body)%00%(trailers:unfold,only)%00%1F refs/heads refs/remotes
debug: [ui] Executing getRecentBranches: git log -g --no-abbrev-commit --pretty=oneline HEAD -n 2500 --
debug: [ui] Executing getConfigValueInPath: git config -z trailer.separators
debug: [ui] Executing getCommits: git log github-desktop-agisilaos/agis/visual-indicator-for-dropping-folder-in-Desktop..pr/4239 --date=raw --max-count=100 --pretty=%H%x1F%s%x1F%b%x1F%an <%ae> %ad%x1F%cn <%ce> %cd%x1F%P%x1F%(trailers:unfold,only) -z --no-color --
debug: [ui] Executing getRemotes: git remote -v
debug: [ui] Executing getAuthorIdentity: git var GIT_AUTHOR_IDENT```

File: app/src/ui/app.tsx (hunk 1 above)
Comment (agisilaos) [reply app/src/ui/app.tsx]: **Update**:  I think I understood what you meant. Just updated the PR!

**Comment by agisilaos** [comment]
Ahh that's a weird "edge" case. Nice catch! @tierninho 

> Any way to have the indicator persist?

Can't think of a way rn. Maybe @shiftkey, @niik can chime in here?

File: app/src/ui/app.tsx [hunk 2]
```diff
@@ -535,10 +535,22 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   public componentDidMount() {
-    document.ondragover = document.ondrop = e => e.preventDefault()
+    document.ondragover = e => {
+      if (this.isShowingModal != null) {
```
niik: Oh! I totally did not realize that isShowingModal wasn't returning a boolean. That's a bug as far as I'm concerned and while I do feel a little bit bad about piling on to your PR I would really appreciate if you could help us fix that. If you change the `isShowingModal` implementation to something like this:

```ts
  private get isShowingModal() {
    return this.state.currentPopup !== null || this.state.errors.length > 0
  }
```

Then we should be able to change this line to be just `if(this.isShowingModal)` which would be much nicer in my opinion.

agisilaos:
🐢

app/src/ui/app.tsx
agisilaos: Indeed that's a much cleaner solution 👍

niik:
> Can't think of a way rn. Maybe @shiftkey, @niik can chime in here?

I'm afraid I don't have much experience with drag/drop in the context of JS/Chromium. There may very well be workaround but I think this PR is a solid improvement on its own so I'm gonna get this merged and we ca see if anyone wants to pick up and try to work around that edge case later.

PR #4240, opened 2018-03-14
avoid a fetch when checkout a pull request

agisilaos:
Fixes: #4068 

My only improvement to this PR would be to make the lines of the variables I added reusable but I'm not sure how I can do this. We can leave it as it is as well. What's your thinking on that?


agisilaos:
🔢

app/src/lib/stores/app-store.ts
```diff
@@ -3279,7 +3279,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
       const defaultRemote = await getDefaultRemote(repository)
       // TODO: I think we could skip this fetch if we know that we have the branch locally
```
shiftkey: I think this comment can be rewritten to drop the `TODO` and describe the intent here

nerdneha:
@agisilaos it looks like we have some repeated logic:
lines `3282`-`3286`
seem to match
lines `3322`-`3327`

specifically these lines:
```js

      const localBranchName = `pr/${pullRequest.number}`
      const doesBranchExist =
        gitStore.allBranches.find(branch => branch.name === localBranchName) !=
        null
```

Maybe we could pull some of that logic up since we have it repeated? It doesn't look like it's extractable into a private method since one of the condition cases uses `localBranchName` so maybe pulling it up is better
(could i get a second pair of 👀 on this?)

app/src/lib/stores/app-store.ts
```diff
@@ -3279,7 +3279,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
       // already. That way we'd match the behavior of checking out a branch.
-      if (defaultRemote) {
+      const gitStore = this.getGitStore(repository)
+      const localBranchName = `pr/${pullRequest.number}`
```
shiftkey: > (could i get a second pair of 👀 on this?)

@nerdneha @agisilaos I don't believe this isn't the right name to lookup for the local branch **when the PR is associated with a branch in the repository, not a fork**.

Let me try and explain my thinking:

 - before we hit this code path, we know the PR comes from the original repository and not a fork
 - after this new code, we continue to git checkout `head.ref`, and not a new `pr/{id}` ref like we would for a fork PR
 - so `doesBranchExists` should be looking for `head.ref` as the local branch

I think that might affect your suggestion to eliminate the duplication, or change it at least.

shiftkey: Actually, there's definitely more nuance to this. Let me poke at some `desktop/desktop` PRs and clarify what checks we should be doing here.

shiftkey: Here's the rough approach that I think we need to do here (avoiding a patch to ensure we're all on the same page about the steps to follow):

```ts
if (isRefInThisRepo) {
  const defaultRemote = await getDefaultRemote(repository)
  // if we don't have a default remote here, it's probably going to just crash and burn
  // on checkout, but that's okay
  if (defaultRemote) {
    // the remote ref will be something like `origin/my-cool-branch` so we should look
    // for that rather than a local branch named `my-cool-branch`
    const remoteRef = `${defaultRemote.name}/${head.ref}`
    // branch.name here is the fully-qualified ref - perhaps that isn't quite clear from the name?
    const remoteRefExists =
      gitStore.allBranches.find(branch => branch.name === remoteRef) != null

    // only try a fetch here if we can't find the ref
    if (!remoteRefExists) {
      await this._fetchRemote(
        repository,
        defaultRemote,
        FetchType.UserInitiatedTask
      )
    }
  }

  // now we should rely on Git's behaviour for finding the remote branch if
  // no local branch matching the name exists, so no change necessary here
  await this._checkoutBranch(repository, head.ref)
} else if (head.gitHubRepository != null) {
```

agisilaos:
Will look into that today and follow up with questions etc, thank you for the detailed comments @shiftkey, @nerdneha! 💯

shiftkey:
Dropping `ready-for-review` because we're still confirming this is the right approach...

shiftkey:
@agisilaos given we've figured out the approach but this has sat for a while without any follow up, could we close this out and update the issue to reference this discussion? You're more than welcome to come back to it when you've got time with a fresh PR, I'm just trying to get on top of our outstanding PRs...

agisilaos:
No need to do that @shiftkey! Will update the PR tonight. Didn't had time before because of exams and stuff

app/src/lib/stores/app-store.ts
```diff
@@ -3283,16 +3283,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-        )
+        // the remote ref will be something like `origin/my-cool-branch`
+        const remoteRef = `${defaultRemote.name}/${head.ref}`
+        const gitStore = this.getGitStore(repository)
+
+        const remoteRefExists =
+          gitStore.allBranches.find(branch => branch.name === remoteRef) != null
+
+        // only try a fetch here if we can't find the ref
+        if (!remoteRefExists) {
```
shiftkey: This should be an explicit comparison to `null`, rather than relying on truthiness.

app/src/lib/stores/app-store.ts
```diff
@@ -3283,16 +3283,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
       head.gitHubRepository.cloneURL === gitHubRepository.cloneURL
 
     if (isRefInThisRepo) {
-      // We need to fetch FIRST because someone may have created a PR since the last fetch
       const defaultRemote = await getDefaultRemote(repository)
-      // TODO: I think we could skip this fetch if we know that we have the branch locally
-      // already. That way we'd match the behavior of checking out a branch.
+      // if we don't have a default remote here, it's probably going
+      // to just crash and burn on checkout, but that's okay
       if (defaultRemote) {
```
shiftkey: This should be an explicit comparison to `null` too.

app/src/lib/stores/app-store.ts
shiftkey: Apologies: I realise now that this is a `boolean`, so there's no need to change it.

agisilaos:
🥖

PR #4243, opened 2018-03-14
simplify constructors across the codebase

shiftkey:
@iAmWillShepherd and I noticed this in a slide deck at TSConf, and I wanted to open this up as a discussion for some `tech-debt` cleanup opportunities.

Basically, TS has a neat feature called "parameter properties" which lets you simplify your constructors by passing in the access modifier:

> Parameter properties are declared by prefixing a constructor parameter with an accessibility modifier or `readonly`, or both. **Using `private` for a parameter property declares and initialises a private member; likewise, the same is done for `public`, `protected`, and `readonly`.**

This eliminates a potential source of errors where you forget to assign a constructor parameter to the corresponding member. `tsc` does identify these and report them as compiler errors already, so the benefits to this are more aesthetic-based:

screen shot 2018-03-15 at 8 33 04 am

I chose the `Account` class to illustrate the aesthetic changes in the attached diff, and that we can move existing JSDoc comments into the constructor to retain Intellisense when browsing the members:

If we're :thumbsup: on this direction I propose we capture a task to tidy up our `app/src/models/` folder and see how that feels.

### DIFF
--- a/app/src/models/account.ts
+++ b/app/src/models/account.ts
@@ -6,43 +6,27 @@ import { getDotComAPIEndpoint, IAPIEmail } from '../lib/api'
  * This contains a token that will be used for operations that require authentication.
  */
 export class Account {
-  /** The access token used to perform operations on behalf of this account */
-  public readonly token: string
-  /** The login name for this account  */
-  public readonly login: string
-  /** The server for this account - GitHub or a GitHub Enterprise instance */
-  public readonly endpoint: string
-  /** The current list of email addresses associated with the account */
-  public readonly emails: ReadonlyArray<IAPIEmail>
-  /** The profile URL to render for this account */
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
I like this a lot!! 🎈  This will remove a lot of ceremony code for sure.

nerdneha:
> right now given how tight our schedule is and where we're at with our roadmap ... Maybe this is something we could aim to do for 1.3.
i agree! seems like an elegant fix once we've gotten some new features for our users

shiftkey:
@nerdneha @niik :thumbsup: i've extracted https://github.com/desktop/desktop/issues/4272 to track this (it's not marked `help-wanted` to avoid someone starting this early)

Issue #3913, opened 2018-02-01
Use common mouse/keyboard combinations for zoom


fkaelberer:
### Description

Please use common keyboard mouse combinations for zooming:
`Ctrl`+`+` for zoom in
`Ctrl`+`-` for zoom out
`Ctrl`+`0` for zoom reset

Issue 1: Shortcut description in the menu is wrong
Issue 2: Numpad `+`/`-` is not supported as in Firefox, Chrome or IE
Issue 3: Mouse scrolling is not supported as in Firefox, Chrome or IE

### Version
**GitHub Desktop version:** 1.0.13
**OS version:** Windows 7

### Steps to Reproduce (Issue 1)
Open the View menu

**Expected behavior:** 
it shows
`Ctrl`+`+` for zoom in

**Actual behavior:**  
Menu screenshot: 
![grafik](https://user-images.githubusercontent.com/4624768/35694499-d8c0f2b0-0781-11e8-939e-3724c71dadba.png)

### Steps to Reproduce (Issue 2)
Press `Ctrl`+`NumPad-+`

**Expected behavior:** it zooms in as in Firefox, Chrome, or IE
**Actual behavior:** nothing happens

### Steps to Reproduce (Issue 3)
Hold `Ctrl` key and scroll with the mouse wheel

**Expected behavior:** it zooms in or out as in Firefox, Chrome, or IE
**Actual behavior:** nothing happens

iAmWillShepherd:
Thanks for taking the time to open an issue! 🙇 

Just to make sure I understand correctly, the problem is that the zoom shortcut is labelled `Ctrl + =` and not `Ctrl +` and GitHub Desktop does not behave like Chrome, Firefox, Safari/IE?

fkaelberer:
> Just to make sure I understand correctly, the problem is that the zoom shortcut is labelled Ctrl + = and  not Ctrl + and GitHub Desktop does not behave like Chrome, Firefox, or Safari/IE/Edge?

That's correct, but I'd rather say: It behaves differently than *any* other zoom-enabled software that I use on my computer, including MS Office, MS Visual Studio [Code], Qt Creator, Thunderbird, Acrobat Reader, Sumatra Pdf, IrfanView, Audacity, even Windows 10's built-in file explorer, in the sense that it does not support `ctrl`+mousewheel-zooming. Among the software titles which supports `ctrl`+`+/-`-zooming, all support also `ctrl`+`numpad+/-`.

[...]

PR #4244, opened 2018-03-14
Fix zoom's order of options and key bindings

agisilaos:
Fixes: #3913 

## Before

screen shot 2018-03-15 at 00 32 49

## After 

screen shot 2018-03-15 at 00 36 32

Here's my question @iAmWillShepherd: How I can change the key binding of the Zoom in from `CMD=` to `CMD+`? I tried to change  the `CmdOrCtrl+=` to `CmdOrCtrl++` and I got as a result:

Any ideas how I can fix that?

screen shot 2018-03-15 at 00 28 15

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -152,11 +152,6 @@ export function buildDefaultMenu(
       },
       separator,
       {
-        label: __DARWIN__ ? 'Reset Zoom' : 'Reset zoom',
-        accelerator: 'CmdOrCtrl+0',
-        click: zoom(ZoomDirection.Reset),
-      },
-      {
         label: __DARWIN__ ? 'Zoom In' : 'Zoom in',
         accelerator: 'CmdOrCtrl+=',
         click: zoom(ZoomDirection.In),
@@ -165,6 +160,11 @@ export function buildDefaultMenu(
         label: __DARWIN__ ? 'Zoom Out' : 'Zoom out',
         accelerator: 'CmdOrCtrl+-',
         click: zoom(ZoomDirection.Out),
+      },
+      {
+        label: __DARWIN__ ? 'Reset Zoom' : 'Reset zoom',
+        accelerator: 'CmdOrCtrl+0',
+        click: zoom(ZoomDirection.Reset),
       },
       separator,
       {


nerdneha:
cc @desktop/design

donokuda:
Hi @agisilaos!

Is there a reason for changing the ordering? I checked the issue you linked to but didn't see the ordering mentioned in there.

agisilaos:
Hm, good question! I was checking other apps such as Bear - Note Editor, Visual Studio Code, Atom among others and checked that's how they order zoom options so I thought it would be a good idea to follow this pattern.

fkaelberer:
Fixes: #3913

Hi, @agisilaos. Which of the three issues in #3913 does this fix?

donokuda:
@agisilaos Makes sense 😄I checked Slack out of curiosity and they order it differently and use a different term for resetting the zoom:



Given that there doesn't seem to be a consensus, and that the menu order isn't a workflow-breaking issue, I think it's okay to not change the ordering in this PR.

agisilaos:
@donokuda
>  Given that there doesn't seem to be a consensus, and that the menu order isn't a workflow-breaking issue, I think it's okay to not change the ordering in this PR.

I think it makes sense 👍 

So my problem persists here. How I can change the `CMD+=` to `CMD++`. Any ideas on that? It's been pretty weird and as I explained above when I try to replace the `=` with `+` the Zoom In Key binding dissapears from the list. 🤔

donokuda:
> So my problem persists here. How I can change the CMD+= to CMD++. Any ideas on that? It's been pretty weird and as I explained above when I try to replace the = with + the Zoom In Key binding dissapears from the list. 🤔

Total shot in the dark, but have you tried something like ` accelerator: 'CmdOrCtrl+Plus',` ?

I think <kbd>+</kbd> is a special keycode where you must to spell out the key. More info in the [accelerator docs](https://electronjs.org/docs/api/accelerator#available-key-codes).

Let me know if that works! 😄 Otherwise, I'm happy to debug on my side.

agisilaos:
@donokuda It worked! Thank you so much I will consider saving this link in my bookmarks for future reference.

This PR is ready for review again. 🦃

agisilaos:
Just an FYI: I updated the title of the PR so it can reflect the updated changes I made to it.

agisilaos:
Suggestion: It would be lovely if someone can test this with a numpad in their keyboard so we can see if we missed anything there.

fkaelberer:
As far as I can tell, your fix repairs the issue that the accelerator hint in the menu is displayed correctly, independent of keyboard layout (before, the label was correct by accident on english keyboards). I don't see that this PR fixes any of the main issues of #3913.

j-f1:
Does Cmd + `=` still work?

agisilaos:
@j-f1 Yes!

agisilaos:
🔢

niik:
Hey @agisilaos, sorry for not getting back to this sooner. I think there's a discrepancy between what you're seeing and what @fkaelberer is experiencing because they are on Windows. I wanted to do some testing of this feature myself on Windows and couldn't get to a machine until late yesterday. The result of that testing is that I'm gonna dismiss my review here so that we can try to figure out what the best approach is.

While this change works wonders on macOS it seems the <kbd>Ctrl</kbd>+<kbd>Plus</kbd> combo no longer works as expected on Windows. This seems to be a known issue in electron (https://github.com/electron/electron/issues/6731). I found one project which used this workaround: https://github.com/zulip/zulip-electron/commit/1948ba2cc3b052faa51e02a66f9ecb0865371133 where they special case for Windows and keep the `Ctrl+=` binding. That obviously doesn't solve the original issue though. Interestingly enough I can't get the `Ctrl+-` binding to work either on Windows which is unrelated to this PR but still a problem.

Additionally we're unfortunately not relying on electron's own logic for expanding the keyboard shortcuts into human readable form but are instead [rolling our own logic](https://github.com/desktop/desktop/blob/525936a9aec13202206887ea0ab3e2124d7211ad/app/src/ui/app-menu/menu-list-item.tsx#L37-L86) and we're not handling the 'Plus' identifier.

zoom-kay

Looking at other apps on Windows (briefly) it seems like most use `Ctrl++`.

Visual Studio code seems to do the Zoom in and Zoom out actions in the same way that we want them to be used but they have their own system for dealing with keyboard shortcuts: https://github.com/Microsoft/vscode/blame/3a70cdfd8f84136e858b3d39e5a709e637fc35e7/src/vs/workbench/electron-browser/main.contribution.ts#L69-L87

I don't have a solution to this but I found that a good way to test whether we're getting the behavior we want is to open dev tools side by side with Desktop and play around with hitting Ctrl+/Ctrl-, it has exactly the behavior we want as far as I can tell, including supporting different keyboard layouts.

I think the path forward here is one of two things. Either we make this change conditional to macOS and consider it a quick win but one that doesn't solve the original issue. Or we dig deeper and try to understand what the actual solution for the Windows problem is.

niik:
I'm gonna drop the ready for review label here until we decide on a course of action

nerdneha:
Hi @agisilaos, it looks like you've gotten a little busy lately with exams and whatnot! Is this still something you're interesting in pursuing? Additionally, did we miss any outstanding or blocking questions?

If not, we thought we might close it just for the time being until you have the bandwidth to open it up again.

PR #2619, opened 2017-09-05
Install LFS filters

joshaber:
Fixes #2227 

1. Run `git lfs install --skip-repo` silently on launch to setup the global filters.
1. Ask the user if they want to run `git lfs install` to setup the hooks when adding a repository which uses LFS.

screen shot 2017-09-05 at 2 54 32 pm

PR #4255, opened 2018-03-15
WIP: [Improvement] additionnal support for submodules

icosamuel:
## Support for submodule
In an effort to better the experience for repositories containing submodules, I've added some options for calling submodule commands such as
```
git submodule init
git submodule update --recursive
git submodule update --recursive --force
```
so that our userbase could make use of the submodules without knowledge of this topic. 

### Progress
At the moment, the options for init and update are accessible from the drop-down menu `Repository`. The Init function is also called by a pop-up that appears when adding an existing repository with submodules.

_Note that this process currently interferes with the `git LFS init` pop-up, because only one can be shown. I need some input on the correct path to take in order to queue them or merge them_

### Design considerations
I would also like to introduce these commands in the `pull` list of commands that follow the use of the `push-pull-button`. If there are any merge conflict when updating, I plan to show a pop-up in order to ask the user if he wants to `--force`the operation and remove any local changes.

This would be an attempt at making the integration of submodules as transparent as possible. Our users that have to work in those subms would include the specific repositories as separate and would chose not to `--force` the subms updates.

I would love any input on these design considerations


shiftkey:
@icosamuel before diving too deep into the code, I want to call out this part:

> In an effort to better the experience for repositories containing submodules
> ... 
> This would be an attempt at making the integration of submodules as transparent as possible.

This could become a significant chunk of work, and it'd be nice to talk about the problems that users face when dealing with submodules, so we can frame this discussion appropriately.

I'm asking this for a couple of reasons:

 - Feedback that is more specific or detailed helps to ensure we're addressing the right thing
 - Feedback in the context of "what is the user trying to achieve?" (i.e. user workflows) helps us to think about how to address the problem within Desktop

Adding new menu items is certainly one way to go about it, but I'd love to figure out ways to address a problem without requiring new menu items or interface components - so the user can keep working without having to learn new steps in their day-to-day work.

Would you mind going into some more detail about our current limitations with submodules, as you see it?

icosamuel:
Our goal in adding more features submodule-wise is to allow users and developers quick access to pulling changes made in the submodules. Ie. If developer A spends most of his time working on part X of a project that uses submodules 1, 2, and 3, we would want him to be able to pull changes from those modules (or all modules) if developer B modified any of those. Dev B would specify which submodule version (commit id) to be pulled by developer A.

At the moment, I noticed that Desktop does not populate the folders where the submodules should be. Althought the `.gitmodules` is present, there is no way of using the code or opening the repo of those submodules. (Or maybe I'm overlooking something obvious) Therefore we have concluded that it would be necessary to further develop the support for submodules. 

In order to reach our goals, there are a multitude of possible solutions, which could be developed on our side only. I created this PR in order to probe for interest in this area and to try to orient ourselves in a direction that will be productive for the whole community.

The current changes in this branch are more of a proof of concept. Again, I realize that the core team want to lead the design of these improvement, and that is why I opened a discussion. But I also consider the possibilities of being in a ratter unique situation and that we could be interfering ratter than helping.

shiftkey:
@icosamuel thanks for the extra details. I'm going to jump around a bit in this reply to hopefully help share my thinking.

> At the moment, I noticed that Desktop does not populate the folders where the submodules should be. 

This is interesting, because if you clone in Desktop it should recursively clone submodules. I gather the repository has come from elsewhere, and that's fine.

This gap reminds me a bit of how we implemented LFS support, and in particular this PR https://github.com/desktop/desktop/pull/2619 where we added support for checking and initialising LFS support once you encounter a repository which needs it.

This is a nice example of the philosophy we want when rolling out new features:

 - prompt the user when the app know something needs to be done
 - guide the user to help them get the repository to a good place
 - once that's done, get out of the way

The fewer things we need to keep in front of the user, the better. Pushing complexity away from the core experience is also ideal here, because submodules aren't something that everyone needs to deal with.

> Althought the `.gitmodules` is present, there is no way of using the code or opening the repo of those submodules. (Or maybe I'm overlooking something obvious) 

One trick (I think I can credit @grokys for this?) is to add the submodule folders to Desktop so that it tracks them, and you can work alongside them in the main repository. But this loses the context of "this repository is a submodule of XYZ" nor do you get any help from the tool, like "these are the changes in submodule 123, want to make a commit to the main repository?" to help with managing submodules.

> If developer A spends most of his time working on part X of a project that uses submodules 1, 2, and 3, we would want him to be able to pull changes from those modules (or all modules) if developer B modified any of those. Dev B would specify which submodule version (commit id) to be pulled by developer A.

I think this is the ideal sort of approach to start from when tackling our submodule support. This gets us away from implementation details immediately and makes us think about what the tool should do to help the user, especially in collaborative scenarios.

At the moment we've got a [number of issues open about submodule support](https://github.com/desktop/desktop/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+submodule) that haven't been prioritized. Some of them touch on things you've raised here, and I think using issues to figure out the way forward is preferred here.

icosamuel:
Thank you @shiftkey for your guidance. We will personalize the workflow for our environment. On every `git pull`, we will verify `git submodules status` and then trigger a `git submodule update --init --recursive` when needed. 

We will try to tackle every edge case as they arise in our workflow. Such edge case includes checking out a branch (or pulling/merging changes) with _less_ submodules than the current state (in which case git does not remove the folders and leaves the user in an unhappy place)

I will follow your recommendation and submit issues when appropriate

PR #4267, opened 2018-03-16
add support for opening repository in PowerShell Core

### LINKED ISSUE #3791: Add support for launching PowerShell Core shell from Desktop
### Description

GitHub Desktop allows me to open Windows PowerShell but not the new [PowerShell Core](https://docs.microsoft.com/en-us/powershell/scripting/whats-new/what-s-new-in-powershell-core-60?view=powershell-5.1).

### Version

**GitHub Desktop version:** 1.0.12

**OS version:** Windows 10

### Steps to Reproduce

1. Options->Advanced->Shell

**Expected behavior:** I should be allowed to open PSC.

**Actual behavior:** No way to open it, just traditional Windows PowerShell.

saschanaz:
Fixes #3791. @shiftkey, would you like to review this? 😆


agisilaos:
Hello @saschanaz thank you for the contribution! I think we have some code formatting errors & that's why the tests fail. Can you please check the logs & fix the errors? Here's the logs from Travis:

screen shot 2018-03-16 at 20 34 57

**Note** Don't forget to merge master in your branch after your fix these errors.

Let me know if there's anything else that you may want to ask! 😃

app/src/lib/shells/win32.ts
```diff
@@ -136,8 +136,8 @@ async function findPowerShellCore(): Promise<string | null> {
 
   const first = powerShellCore[0]
   if (first.type === RegistryValueType.REG_SZ) {
-    if (await pathExists(path)) {
-      return path
+    if (await pathExists(first.data)) {
```
agisilaos: Why you changed this one from `path` to `first.data`, I suspect that this is the reason build fails. @saschanaz

shiftkey:
@saschanaz three initial questions:

 - could you also update [the documentation](https://github.com/desktop/desktop/blob/master/docs/technical/shell-integration.md) to mention PowerShell Core?
 - is it worth adding support for macOS or Linux versions while we're in here, or shall we leave that to someone else to add in later?
 - installing [the latest version](https://github.com/PowerShell/PowerShell/releases/tag/v6.0.2) is all that we need to test this out, right?

shiftkey:
Moving this into 1.2 to ensure we don't rush this out as a 1.1.x hotfix.

app/src/lib/shells/darwin.ts
```diff
@@ -25,6 +26,10 @@ export function parse(label: string): Shell {
     return Shell.iTerm2
   }
 
+  if (label === Shell.PowerShellCore) {
+    return Shell.iTerm2
```
agisilaos: I think this needs to be replaced with `Shell.PowerShellCore`

saschanaz:
>could you also update the documentation to mention PowerShell Core?

Done!

>is it worth adding support for macOS or Linux versions while we're in here, or shall we leave that to someone else to add in later?

I pushed https://github.com/desktop/desktop/pull/4267/commits/410984eae516a1d656b425ae61748275ea1448f5 😃

>installing the latest version is all that we need to test this out, right?

Yup!

app/src/lib/shells/darwin.ts
```diff
@@ -36,6 +41,8 @@ function getBundleID(shell: Shell): string {
       return 'com.googlecode.iterm2'
     case Shell.Hyper:
       return 'co.zeit.hyper'
+    case Shell.PowerShellCore:
+      return 'com.microsoft.powershell'
```
shiftkey: Minor thing: I tested this with the [binary installation](https://github.com/PowerShell/PowerShell/blob/master/docs/installation/macos.md#installation-via-binary-archive) steps and Desktop didn't detect it. `brew cask install powershell` worked fine, and I gather the direct install option also works.

I'm not sure how many people would install it manually _as well as want to use it from Desktop_, but if we have to support that option we might have to fall back to how Linux resolves it's shells. But I won't hold up this PR for that.

app/src/lib/shells/linux.ts
```diff
@@ -114,6 +127,14 @@ export function launch(
     case Shell.Tilix:
     case Shell.Gnome:
       return spawn(foundShell.path, ['--working-directory', path])
+    case Shell.PowerShellCore:
+      const psCoreCommand = `"Set-Location -LiteralPath '${path}'"`
+      return spawn(foundShell.path, [
+        'pwsh',
```
shiftkey: I currently see this error when trying to open a repository in PowerShell Core on Ubuntu 16.04:

> debug: [ui] [PowerShell Core] stderr: 'The argument 'pwsh' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.'
> debug: [ui] [PowerShell Core] exit code: 64

I thought removing the standalone `pwsh` here would be the fix (it's how PowerShell should work on Windows), and the error does go away, but then I don't see a shell being displayed and I don't see any other errors, so I'm not sure what's happened here.

saschanaz: Removing `pwsh` should work. (Sorry for including it!)

Does `pwsh -NoExit -Command "Set-Location -LiteralPath '(pathname)'"` on bash shell work there? It seems to take some time at first time.

daviwil:
The default install path for pwsh on macOS is /usr/local/bin/pwsh, so you'll need an extra case for that. In Linux it is indeed /usr/bin/pwsh.

app/src/lib/shells/linux.ts
shiftkey: > Does `pwsh -NoExit -Command "Set-Location -LiteralPath '(pathname)'"` on bash shell work there? 

Haven't tried it out, but the fact that you're running it within an existing shell means it's not quite the same setup as what Desktop is trying to do.

I think we need to leverage `$TERM` here to run `pwsh` within the context of a terminal emulator to get it appearing (hat tip to @daviwil for the clue):

shiftkey:
@daviwil we actually avoid looking for the paths explicitly on macOS by using the bundle ID to resolve the application path on disk, which I called out here https://github.com/desktop/desktop/pull/4267#discussion_r178492456

app/src/lib/shells/linux.ts
shiftkey: EDIT: actually, `$TERM` isn't set when launching Desktop like you would a regular application (this only really worked because I was doing `yarn start:dev` to start Desktop). Back to the drawing board.

shiftkey: Actually, if `pwsh` on Linux depends on running in an existing terminal, then it doesn't really make sense to have this as a menu option for Linux.

Any objections to just deferring first-class Linux support until someone figures out what that should look like @saschanaz @daviwil?

saschanaz: Turns out that `require("child_process").spawnSync("/usr/bin/pwsh", ['-NoExit'])` just emits stdout to pipe and then exits.

Issue #1476, opened 2017-05-16
Don't show the "Select a repository" button if there are no repositories to select


joshaber:
@niik and I talked about this. We might do more, but at minimum the "Select a repository" button is a lie if there are no repositories to select.

shiftkey:
If someone wants to take a shot at this, I made this patch which changes the button behaviour:

```diff
diff --git a/app/src/ui/app.tsx b/app/src/ui/app.tsx
index 4032be2d2..af34928bb 100644
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1384,9 +1384,12 @@ export class App extends React.Component<IAppProps, IAppState> {
     if (repository) {
       icon = iconForRepository(repository)
       title = repository.name
-    } else {
+    } else if (this.state.repositories.length > 0) {
       icon = OcticonSymbol.repo
       title = __DARWIN__ ? 'Select a Repository' : 'Select a repository'
+    } else {
+      icon = OcticonSymbol.question
+      title = __DARWIN__ ? 'No Repositories' : 'No repositories'
     }
 
     const isOpen =
```

<img width="335" src="https://user-images.githubusercontent.com/359239/36771341-f8488850-1ca3-11e8-95dc-7cf41f05e9a3.png">

Feel free to take it further

agisilaos:
PR's up #4268

PR #4268, opened 2018-03-16
Update "Select a repository" button if there are no repositories to select

agisilaos:
Fixes: #1476

screen shot 2018-03-16 at 16 02 23


agisilaos:
🦃

tierninho:
tested = ✅

app/src/ui/app.tsx
```diff
@@ -1391,9 +1391,12 @@ export class App extends React.Component<IAppProps, IAppState> {
     if (repository) {
       icon = iconForRepository(repository)
       title = repository.name
-    } else {
+    } else if (this.state.repositories.length > 0) {
       icon = OcticonSymbol.repo
       title = __DARWIN__ ? 'Select a Repository' : 'Select a repository'
+    } else {
+      icon = OcticonSymbol.question
```
niik: Could we stick to the `OcticonSymbol.repo` icon here please? That should be the base icon since that's the core responsibility of the button.

agisilaos: I had a feeling that you will want the icon to stay the same but I "followed" what's proposed in the issue. Doing this change right now.

agisilaos:
🐢

PR #4234, opened 2018-03-13
drafting release notes for next beta

shiftkey:
Once we're happy with these, and we've done enough testing on `master` to be confident, we should cut a new beta - there's enough value here to get out to users!

PR #4288, opened 2018-03-19
relax regex rule to catch more merged PRs when generating changelog

shiftkey:
This came up as part of #4234, and it looks like some PRs can link to issues using `Fixes: #id` (with the colon). This is valid syntax for auto-closing issues when the PR is merged:

Our parser doesn't handle this, so it was overlooked in the release notes generator:

```
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

$ GITHUB_ACCESS_TOKEN=[token] yarn draft-release beta | grep \#4004

$ git checkout relax-regex-for-fixed-issue
Switched to branch 'relax-regex-for-fixed-issue'

$ GITHUB_ACCESS_TOKEN=[token] yarn draft-release beta | grep \#4004
    "[Fixed] Add visual indicator that a folder can be dropped on Desktop - #4004. Thanks @agisilaos!",
```

 - [x] extract this into something more testable
 - [x] throw a bunch of the documented keywords at it

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,9 @@
   "scripts": {
     "cli": "ts-node --require ./app/test/globals.ts --require ./app/src/cli/dev-commands-global.js app/src/cli/main.ts",
     "test:integration": "cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe mocha -t 30000 --require ts-node/register app/test/integration/*.ts",
-    "test:unit": "cross-env GIT_AUTHOR_NAME=\"Joe Bloggs\" GIT_AUTHOR_EMAIL=\"joe.bloggs@somewhere.com\" GIT_COMMITTER_NAME=\"Joe Bloggs\" GIT_COMMITTER_EMAIL=\"joe.bloggs@somewhere.com\" TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe electron-mocha -t 10000 --renderer --require ts-node/register --require ./app/test/globals.ts app/test/unit/*.{ts,tsx} app/test/unit/**/*.{ts,tsx}",
-    "test": "yarn test:unit && yarn test:integration",
+    "test:unit": "cross-env GIT_AUTHOR_NAME=\"Joe Bloggs\" GIT_AUTHOR_EMAIL=\"joe.bloggs@somewhere.com\" GIT_COMMITTER_NAME=\"Joe Bloggs\" GIT_COMMITTER_EMAIL=\"joe.bloggs@somewhere.com\" TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe electron-mocha -t 10000 --renderer --require ts-node/register --require ./app/test/globals.ts app/test/unit/*.{ts,tsx} app/test/unit/**/*.{ts,tsx} script/changelog/test/*.ts",
+    "test:script": "mocha -t 10000 --require ts-node/register script/changelog/test/*.ts",
+    "test": "yarn test:unit && yarn test:script && yarn test:integration",
     "test:setup": "ts-node script/test-setup.ts",
     "test:review": "ts-node script/test-review.ts",
     "postinstall": "cd app && yarn install --force && cd .. && git submodule update --recursive --init && yarn compile:tslint",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "cli": "ts-node --require ./app/test/globals.ts --require ./app/src/cli/dev-commands-global.js app/src/cli/main.ts",
     "test:integration": "cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe mocha -t 30000 --require ts-node/register app/test/integration/*.ts",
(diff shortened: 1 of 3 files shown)


script/changelog/parser.ts
```diff
@@ -40,7 +40,7 @@ function getChangelogEntry(commit: IParsedCommit, pr: IAPIPR): string {
   let type = PlaceholderChangeType
   const description = capitalized(pr.title)
 
-  const re = /Fixes #(\d+)/gi
+  const re = /Fixes:? #(\d+)/gi
```
j-f1: Are any other characters allowed here?

agisilaos: @j-f1 I think yes! See: https://help.github.com/articles/closing-issues-using-keywords/

If by characters you mean `:`, `?`, I'm not sure but for words we can definitely use more words than the word `Fixes`. More here https://help.github.com/articles/closing-issues-using-keywords/

shiftkey: I knew this was going to be a can of worms...

staabm: Should it also support more then a single space in between colon and #

shiftkey: @staabm probably, yes

package.json
```diff
@@ -7,8 +7,9 @@
...
     "test:integration": "cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe mocha -t 30000 --require ts-node/register app/test/integration/*.ts",
-    "test:unit": "cross-env GIT_AUTHOR_NAME=\"Joe Bloggs\" GIT_AUTHOR_EMAIL=\"joe.bloggs@somewhere.com\" GIT_COMMITTER_NAME=\"Joe Bloggs\" GIT_COMMITTER_EMAIL=\"joe.bloggs@somewhere.com\" TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe electron-mocha -t 10000 --renderer --require ts-node/register --require ./app/test/globals.ts app/test/unit/*.{ts,tsx} app/test/unit/**/*.{ts,tsx}",
-    "test": "yarn test:unit && yarn test:integration",
+    "test:unit": "cross-env GIT_AUTHOR_NAME=\"Joe Bloggs\" GIT_AUTHOR_EMAIL=\"joe.bloggs@somewhere.com\" GIT_COMMITTER_NAME=\"Joe Bloggs\" GIT_COMMITTER_EMAIL=\"joe.bloggs@somewhere.com\" TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe electron-mocha -t 10000 --renderer --require ts-node/register --require ./app/test/globals.ts app/test/unit/*.{ts,tsx} app/test/unit/**/*.{ts,tsx} script/changelog/test/*.ts",
+    "test:script": "mocha -t 10000 --require ts-node/register script/changelog/test/*.ts",
```
shiftkey: This is the first sort of testing infrastructure we've got for our scripts, and I felt that they belonged under `script` rather than under `app/test`, so a separate command from our existing `test:unit` (because they don't rely on Electron) made sense.

script/changelog/parser.ts
```diff
@@ -35,20 +35,28 @@ function capitalized(str: string): string {
...
 }
 
-function getChangelogEntry(commit: IParsedCommit, pr: IAPIPR): string {
+export function findIssueRef(body: string): string {
   let issueRef = ''
-  let type = PlaceholderChangeType
-  const description = capitalized(pr.title)
 
-  const re = /Fixes #(\d+)/gi
+  const re = /(close[s]?|fix(e[sd])?|resolve[sd]):?\s*#(\d+)/gi
```
shiftkey: The use of `(e[sd])` here means a second group will exist when matching `Fixes: #{id}` and I'm not sure of a better way to do this - but it's easy enough here to poke at "the last group" below so I'm somewhat okay with this.

j-f1: `(?:e[sd])` does the same thing as `(e[sd])`, except it doesn’t create a group number that you can access later.

package.json
iAmWillShepherd: Did you mean to leave `script/changelog/test/*.ts` at the end of line 10?

package.json
```diff
@@ -7,8 +7,9 @@
...
+    "test": "yarn test:unit && yarn test:script && yarn test:integration",
```
iAmWillShepherd: Do the tests for scripts need to be a part of our CI process?

script/changelog/parser.ts
shiftkey: Nice shout. Actually, maybe I want to keep the group number around (so it's always three). Thinking on this some more...

package.json
shiftkey: I did not - Good spot!

package.json
shiftkey: I could go either way on this, but here's my gut feeling on why I put them in here:

 - running these tests as part of CI saves us having to remember to run them manually
 - not catching failures early might affect our ability to release when it comes time to ship, because we'll have to spend time troubleshooting what these scripts are doing - and what they should be doing
 - there may be implicit coupling between application code and script code (I think they should be kept separate for sanity reasons)

shiftkey:
:apple:

shiftkey:
:apple:

Issue #3502, opened 2017-12-04
Get smarter about fetching


iAmWillShepherd:
> Maaaaybe. But then whenever the background fetcher kicks in, we'll also refresh PRs. Which may be ok, but that duplicates work we're already doing in the pull request updater.

So our options seem to be: (1) make fetch smarter so that it only fetches PRs when not called from the background fetcher, or (2) the background fetcher now refreshes PRs as part of its fetch and the pull request updater is only in charge of the post-push logic.

There's probably some other options I'm not thinking of too 😄 I'd lean towards (1) for now for simplicity's sake, but I'm totally open to other thoughts.

@joshaber made the case that we should only refresh PRs in the fetch function when it's not initiated by the backgroundFetcher.

See PR #3448 for additional context.

shiftkey:
:thumbsup: to the first approach

PR #4311, opened 2018-03-23
refresh issues when user manually fetches

### LINKED ISSUE #4076: Allow for issues to be manually refreshed
### Description

When creating issues on GitHub while GitHub Desktop is open, they won't show up when referencing them in commit field (#`issue-number`).

### Version

**OS version:** Microsoft Windows [Version 10.0.16299.248]

### Steps to Reproduce

1. Open GitHub Desktop
1. Create an issue on GitHub
1. GitHub Desktop fails to show it when referencing the issue in commit field.

**Expected behavior:** Being able to select the issue that was recently created.

**Actual behavior:** Fails to show the issue.

#### Additional Information

No additional information 😜

shiftkey:
Fixes #4076

I didn't like how we use `update` and `fetch` in this API, so I took the chance to clean this up so it looked similar to the PR APIs inside `app-store` and other components - these behave similarly in terms of handling the refreshing under the hood, might as well ensure they're named consistently.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -823,7 +823,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const gitHubRepository = repository.gitHubRepository
 
     if (gitHubRepository != null) {
-      this._updateIssues(gitHubRepository)
+      this._refreshIssues(gitHubRepository)
       this.loadPullRequests(repository, async () => {
         const promiseForPRs = this.pullRequestStore.fetchPullRequestsFromCache(
           gitHubRepository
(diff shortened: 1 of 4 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -2439,6 +2439,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
         if (fetchType === FetchType.UserInitiatedTask) {
```
niik: Why are we only doing this for manual fetches? Shouldn't we be refreshing PRs and issues in the background?

shiftkey: We avoid doing this for background fetching because we already have a `PullRequestUpdater` that's working in the background and updating every 10mins - #3502 has some more context for why we went down that path. I think the goal here is to force a refresh of PRs as a part of the user fetching changes from the remote repository.

We don't really have an equivalent for fetching issues in the background currently - `IssuesAutocompletionProvider` manages this inside it's own `ThrottledScheduler` which runs ever minute, but this is only started _after_ an autocompletion request is initiated - and the list won't be refreshed when it does have new results (I think you'd need to change the input to re-render the list).

https://github.com/desktop/desktop/blob/d73112a9fc9ca4a953697ae471d3942ae8e07d7b/app/src/ui/autocompletion/issues-autocompletion-provider.tsx#L51-L59

PR #4319, opened 2018-03-24
add back a dropped dependency

shiftkey:
Looks like one of my merged depedency PRs didn't catch this lockfile update.

You'll notice this if you run `yarn` against `master` such as 58af53bf0ca8fb7d053bb68d6a2bef6413cfe4fd.

It doesn't affect the running app, just a bit of nuisance if you're working against the latest changes.


shiftkey:
#jolo

PR #4330, opened 2018-03-26
tweak pull request status text

### LINKED ISSUE #3521: Should CI status say "Failure" or "failed"?
A bit nit-picky, but I think we should use the word "failed" and not "failure" for the CI status check, as "failed" in surfaced in the current UI on GitHub.com:

screen shot 2017-12-06 at 11 22 39 am

screen shot 2017-12-06 at 11 33 12 am

Also, "failed" is a clearer message that the action is in the past.

shiftkey:
Fixes #3521

When no statuses were found, we'd just UPPERCASE the reported status. That doesn't seem friendly, so I've opened this PR to tweak this output and add some tests for reference.

I liked the feedback in #3521 to map `failure` to `failed`, aside from that the functionality is unchanged.

 - [x] can we do better than the `0/1 checks OK` message when it's a failed status?


app/src/ui/branches/pull-request-status.ts
```diff
@@ -0,0 +1,29 @@
+import { PullRequestStatus } from '../../models/pull-request'
+import { APIRefState } from '../../lib/api'
+import { assertNever } from '../../lib/fatal-error'
+
+export function toFriendlyText(state: APIRefState): string {
```
niik: This doesn't seem like it needs to be exported?

app/src/ui/branches/pull-request-status.ts
```diff
@@ -0,0 +1,29 @@
...
+    case 'pending':
+    case 'success':
+      return state
+
+    default:
+      return assertNever(state, `Unknown APIRefState value: ${state}`)
+  }
+}
+
+export function getSummary(prStatus: PullRequestStatus): string {
```
niik: Could we document this please? Additionally, could we name it `getPRStatusSummary` or something similar?

shiftkey: f49cd887b7a3f356f581be3abfe8e5184a03d260 has the docs

shiftkey:
:apple:

tierninho:
Noticed the tooltips in GH/GH were different, however, the consistency now on Desktop is 👍 

screen shot 2018-03-26 at 1 43 26 pm

shiftkey:
@tierninho that second message with the specific error is something I'd like to see if we can emulate in Desktop - it's better context than what we're currently displaying with the `0/1 checks OK` message.

niik:
Removing the ready-for-review since there seems to be more work to do here

shiftkey:
:apple:

PR #4331, opened 2018-03-26
add support for publishing to a team within an organization

### LINKED ISSUE #826: specify a team (optional?) when publishing repository
[Create Repository](https://developer.github.com/v3/repos/#create) has this optional parameter for the team:

We should make this available in the Publish Repository popover if you've selected an organization.

shiftkey:
Fixes #826 

We haven't seen many requests for this, but it seemed simple enough to add that I took a shot at it now.

- [ ] test `None` behaves as expected when publishing
- [ ] test selecting a team works as expected
- [ ] filter teams that can only publish a repository (`push` or `admin` based on `permission` listed [here](https://developer.github.com/v3/teams/#add-or-update-team-repository))
- [ ] add a tooltip to explain that selecting a team is optional, and this team will own the repository

### DIFF
--- a/app/src/ui/publish-repository/publish-repository.tsx
+++ b/app/src/ui/publish-repository/publish-repository.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { Account } from '../../models/account'
-import { API, IAPIUser } from '../../lib/api'
+import { API, IAPIUser, IAPITeam } from '../../lib/api'
 import { TextBox } from '../lib/text-box'
 import { Select } from '../lib/select'
 import { DialogContent } from '../dialog'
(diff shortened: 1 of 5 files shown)


app/src/ui/publish-repository/publish-repository.tsx
```diff
@@ -138,6 +166,43 @@ export class PublishRepository extends React.Component<
...
 
+  private renderTeams(): JSX.Element | null {
+    if (this.state.teams.length === 0) {
+      return null
+    }
+
+    const options = new Array<JSX.Element>()
+    options.push(
+      <option value={-1} key={-1}>
+        None
```
j-f1: What happens if there’s a team called “None?”

shiftkey: It's a potential source of confusion, but basically:

 - It'll appear below this in the list
 - It'll have it's own distinct `key` derived from it's index, which we should correctly map when the select it

I'm open to suggestions about what to put here for the "i don't want to select a team" thing here, but I've borrowed this form the org selection list.

Somewhat related: GitHub The Website does it a bit differently when creating a new repository, showing your user account as the owner by default when you're creating a new repository. 



Maybe we could make this experience be more consistent with that?

shiftkey:
Given the general lack of demand for this feature, I'm going to close this out and focus on some tougher challenges. I'll leave this branch open for someone else to pick up and follow up on the remaining tasks.

PR #3497, opened 2017-12-04
SVGs now have a visual diffs

mathieudutour:
context: https://github.com/desktop/desktop/pull/2383#issuecomment-328639486 (sorry for the delay)

- Add a new tab to the existing image diffs mode for file that supports it (only svg for now).
- In case it's an added or deleted file, it only shows 2 tabs: "Visual" and "Text"

PR #4348, opened 2018-03-28
reorganize diff models and root component

shiftkey:
The current `app/src/models/diff.ts` weighs in at 571 lines, and `app/src/ui/diff/index.tsx` weighs in at 1157 lines - they've both accrued a bunch of diff-related concerns, and #3497 only adds to it.

This PR is work to organize things better before I pull in the #3497 work:

 - `app/src/models/diff.ts` is now split into four specialized modules and hidden behind an `index.ts`
    - `raw-diff.ts` - this is the Git representation of a diff that Desktop builds upon
    - `diff-line.ts` - the base model for our interactive diffs
    - `diff-selection.ts`- the current state of a selected diff (all/none/partial)
    - `image.ts` - the base model for an image diff
    - `FileSummary` is dead code :fire::fire::fire::fire:

 - lots of APIs don't really need to take an `ITextDiff` when they only need the underlying `ReadonlyArray<DiffHunk>` contents - so refactoring those to pass in the specific `hunks` decouples these functions from interfaces to make future work easier

 - syntax highlighting work in `app/src/ui/diff/index.tsx` now mostly lives in it's own module `app/src/ui/diff/syntax-highlighting/index.ts`

 - extract `TextDiff` component to contain the underlying `CodeMirrorHost` and the necessary interactions for diff selection

TODO:

 - [x] ponder whether these abstractions can be refined at all
 - [x] a bunch more testing to ensure I haven't impacted anything in unexpected ways


app/src/lib/git/diff.ts
```diff
@@ -267,9 +267,9 @@ export async function convertDiff(
   commitish: string,
   lineEndingsChange?: LineEndingsChange
 ): Promise<IDiff> {
-  if (diff.isBinary) {
-    const extension = Path.extname(file.path)
+  const extension = Path.extname(file.path).toLowerCase()
```
shiftkey: This actually catches a bug where `.PNG` is skipped as a file extension, rather than `.png`. We talked about that in #3497 and I wanted to ensure this didn't get missed.

PR #4355, opened 2018-03-29
generate an analysis of the renderer bundle when building for prod

shiftkey:
Currently our renderer bundle is the biggest part of Desktop. After a `yarn build:prod` this is the contrast:

```sh
$ du -sh out/*.js
4.0K	out/ask-pass.js
 32K	out/cli.js
180K	out/crash.js
252K	out/highlighter.js
160K	out/main.js
1.4M	out/renderer.js
```

A couple of times in the past I've spelunked into this to understand more of the situation, and I found myself doing it again yesterday. So why not make this available to everyone?

To make it easier for everyone to explore and analyze things, this PR adds support for generating a breakdown of the emitted renderer bundle contents in the background, using `webpack-bundle-analyzer`.

This is what a report looks like:

It's an interactive diagram, and lets you view based on file size, parsed size or gzipped size (I think we're interested in parsed size as affects launch time). For example, I saw `moment.js` was a significant feature of the bundle, and started digging into how we're using it to see if we could switch to something leaner.

### DIFF
--- a/app/webpack.production.js
+++ b/app/webpack.production.js
@@ -7,6 +7,8 @@ const merge = require('webpack-merge')
 const ExtractTextPlugin = require('extract-text-webpack-plugin')
 const BabelPlugin = require('babel-webpack-plugin')
 
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer')
+
 const config = {
   devtool: 'source-map',
   plugins: [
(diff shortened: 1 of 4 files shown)


j-f1:
It might be interesting to add [`webpack-bundle-analyzer-sunburst`](https://www.npmjs.com/package/webpack-bundle-analyzer-sunburst), which shows the same information in a different way.

shiftkey:
@j-f1 I found the above view good enough for identifying these large dependencies, but we can revisit the end visualization later if we want more detail or a different representation..

PR #4357, opened 2018-03-29
do license parsing at build time to avoid bundling a hefty YAML parser

shiftkey:
Once upon a time, we merged #727 which bundled a collection of licenses in a submodule with the app and parsed them out to display to the user in a list, and things were good.

Fast forward to today, and we're dealing with a really complex `renderer.js` file which weighs in at over 1.4MB. This matters because there might be opportunities for us to defer work that's occurring as part of initializing the app.

One of the places I started looking was the report in #4355 and one thing stood out - esprima.js

It's parsed size is 137,268 bytes, or 9% of the 1,464,048 parsed bytes of the renderer file. That's alot.

And what are we using it for? This is what `front-matter` is using under the hood, and Desktop uses it to create a list of licenses when you create a new repository:

If you look at the code inside `app/src/ui/add-repository/licenses.ts` you'll see that it's:

 - scanning a directory for files
 - parsing each of these files for the YAML front matter
 - building up and caching the list of licenses

Doing the scanning of licenses at runtime feels inefficient, so this PR moves all this computation into the build script. This means:

 - we compute the license information once, and embed a JSON file in the app containing the metadata
 - we no longer need to include `front-matter` as a runtime dependency
 - we no longer need to include the `choosealicense.com` submodule in the app

TODO: 

 - [x] use async method for reading new JSON file to address lint error
 - [x] ensure we are still bundling the license for choosealicense.com that was previously in the submodule

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -30,7 +30,6 @@
     "event-kit": "^2.0.0",
     "file-uri-to-path": "0.0.2",
     "file-url": "^2.0.2",
-    "front-matter": "^2.1.2",
     "fs-extra": "^2.1.2",
     "fuzzaldrin-plus": "^0.6.0",
     "keytar": "^4.0.4",
--- a/app/src/ui/add-repository/licenses.ts
+++ b/app/src/ui/add-repository/licenses.ts
@@ -1,20 +1,11 @@
 import * as Path from 'path'
 import * as Fs from 'fs'
 
-interface IFrontMatterResult<T> {
-  readonly attributes: T
-  readonly body: string
-}
-
-const frontMatter: <T>(
-  path: string
-) => IFrontMatterResult<T> = require('front-matter')
-
-interface IChooseALicense {
-  readonly title: string
-  readonly nickname?: string
+interface ILicenseWithMetadata {
+  readonly name: string
   readonly featured?: boolean
   readonly hidden?: boolean
+  readonly body: string
 }
 
 export interface ILicense {
@@ -39,21 +30,7 @@ interface ILicenseFields {
   readonly year: string
 }
 
-const root = Path.join(__dirname, 'static', 'choosealicense.com', '_licenses')
-
 let cachedLicenses: ReadonlyArray<ILicense> | null = null
-
-function readFileAsync(path: string): Promise<string> {
-  return new Promise((resolve, reject) => {
-    Fs.readFile(path, 'utf8', (err, data) => {
-      if (err) {
-        reject(err)
-      } else {
-        resolve(data)
-      }
-    })
-  })
-}
 
 /** Get the available licenses. */
 export function getLicenses(): Promise<ReadonlyArray<ILicense>> {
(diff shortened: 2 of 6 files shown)


j-f1:
What do you think about using something like [`val-loader`](https://webpack.js.org/loaders/val-loader/) here? It would let you avoid IO at runtime, and it might be useful in other places in the codebase as well.

---

I did some digging into why a YAML parser would include a JS parser, and it seems to be because you can put this in your YAML:

```yaml
foo: !!js/function 'function () { doSomething() }'
```

and you’ll get `{ foo: function() { doSomething() } }` when you parse it.

j-f1:
Maybe something like

```ts
export = () => {
  const chooseALicense = path.join(outRoot, 'static', 'choosealicense.com')
  const licensesDir = path.join(chooseALicense, '_licenses')

  const files = fs.readdirSync(licensesDir)

  const licenses = new Array<ILicense>()
  for (const file of files) {
    const fullPath = path.join(licensesDir, file)
    const contents = fs.readFileSync(fullPath, 'utf8')
    const result = frontMatter<IChooseALicense>(contents)
    const license: ILicense = {
      name: result.attributes.nickname || result.attributes.title,
      featured: result.attributes.featured || false,
      hidden:
        result.attributes.hidden === undefined || result.attributes.hidden,
      body: result.body.trim(),
    }

    if (!license.hidden) {
      licenses.push(license)
    }
  }

  return {
    code: `module.exports = ${JSON.stringify(licenses)}`,
    contextDependencies: [chooseALicense],
    cacheable: true,
  }
}
```

(the function can be an `async () => ...` too)

shiftkey:
@j-f1 

> What do you think about using something like `val-loader` here? It would let you avoid IO at runtime, and it might be useful in other places in the codebase as well.

I'm trying to move things that aren't important to the initial launch out of the renderer script - there's other ideas I have around this, but this was some low-hanging fruit. From what I understand, `val-loader` is about being able to generate source code to embed this data - I want to avoid adding more to the renderer script at this stage.

This IO (that's only actually important when you initialize a repository) was already asynchronously executed, except it was occurring as N file reads (for however many files were in the directory) versus a single JSON file (which is 172 KB because it contains all the license text).

j-f1:
> which is 172 KB because it contains all the license text

Oh yeah, you don’t want to have to wait for that to parse at app startup.

shiftkey:
Some side-by-side stats using `du -h` and the macOS build. The changes become less significant once it's packaged and the installer is generated, but the less JS included in the app is the goal here:

|                                                   | `master` | `we-dont-need-to-ship-a-yaml-parser` |
|---------------------------------------------------|----------|--------------------------------------|
| out/static/*                                      | 2.8 MB   | 1.4 MB                               |
| out/renderer.js                                   | 1.4 MB   | 1.2 MB                               |
| out/renderer.js.map                               | 4.2 MB   | 3.8 MB                               |
| dist/GitHub Desktop-darwin-x64/GitHub Desktop.app | 157 MB   | 155 MB                               |
| dist/GitHub Desktop.zip                           | 64 MB    | 63 MB                                |

PR #4362, opened 2018-03-29
Fix for Cloning gitlab URL chooses GitHub repository with same name

### LINKED ISSUE #4154: Cloning gitlab URL results in GitHub repository being cloned.
### Description

If I try to clone a gitlab.com URL that also exists on github.com, the github.com URL will be cloned instead.

### Version

**GitHub Desktop version:** 1.1.1-beta1
**OS version:** win10

### Steps to Reproduce

1. Go to File -> Clone Repository
2. select the URL tab
3. paste in https://gitlab.com/inkscape/inkscape.git 
4. Click Clone

**Expected behavior:** https://gitlab.com/inkscape/inkscape.git  is cloned

**Actual behavior:** https://github.com/inkscape/inkscape.git is cloned

**Reproduces how often:** Every time afaics

### Logs

```
2018-03-02T08:31:40.671Z - info: [ui] Executing fetch: git -c credential.helper= fetch --progress --prune origin (took 1.099s)
2018-03-02T08:33:18.028Z - info: [ui] [AppStore.getAccountForRemoteURL] account found for remote: https://github.com/inkscape/inkscape.git - grokys (has token)
2018-03-02T08:35:03.033Z - info: [ui] Executing clone: git -c credential.helper= clone --recursive --progress -- https://github.com/inkscape/inkscape.git D:\projects\inkscape 
```

JordanMussi:
Fixes #4154 

I have also added some unit tests for the `findAccountForRemoteURL` method and would like some feedback:
- Should I be using the repository given as an example in #4154 (`https://gitlab.com/inkscape/inkscape.git`/`https://github.com/inkscape/inkscape.git`) or use one provided by the desktop team? https://github.com/desktop/desktop/blob/9931007e2604ebcae96e7d6aa7e026b0adc10761/app/test/unit/find-account-test.ts#L31-L34
- How should I test that the correct GitHub account is selected for a repository?
https://github.com/desktop/desktop/blob/9931007e2604ebcae96e7d6aa7e026b0adc10761/app/test/unit/find-account-test.ts#L52-L57 This is what I want to do but since the credentials are being faked [the call](https://github.com/desktop/desktop/blob/9931007e2604ebcae96e7d6aa7e026b0adc10761/app/src/lib/find-account.ts#L57) to `canAccessRepository` returns false so the account is ignored and the "anonymous" account is returned instead.

<br>

Also there is a discrepancy between how an account is identified for a parsed GitHub URL (`https://github.com/inkscape/inkscape.git`) and an owner/name (`inkscape/inkscape`).
For the former `canAccessRepository` is not called, so the first GitHub account found is returned. https://github.com/desktop/desktop/blob/9931007e2604ebcae96e7d6aa7e026b0adc10761/app/src/lib/find-account.ts#L90-L104
Whereas `canAccessRepository` is called when identifying from owner/name. https://github.com/desktop/desktop/blob/9931007e2604ebcae96e7d6aa7e026b0adc10761/app/src/lib/find-account.ts#L56-L61

Except for those queries this is good to go 🙃


JordanMussi:
Hmm not sure why Travis is failing this test, it works for me and the other to CIs...
https://github.com/desktop/desktop/blob/9931007e2604ebcae96e7d6aa7e026b0adc10761/app/test/unit/find-account-test.ts#L46-L50

```
 1 failing
  1) findAccountForRemoteURL
       finds the anonymous account for public GitHub owner/name repository:
      AssertionError: expected null to not equal null
      + expected - actual
      at it (/home/travis/build/desktop/desktop/app/test/unit/find-account-test.ts:48:28)
      at <anonymous>
```

shiftkey:
@JordanMussi I've added some tracing to those tests - I couldn't trigger the same error on my Ubuntu VM, and I'm not sure what Linux-specific thing you might have tripped with these changes...

JordanMussi:
@shiftkey I'm running the tests locally on Windows. Is there any reason there might be discrepancies caused by the order of the tests in the file? (See the builds of my previous commits to this branch)

Please could you also address some of the points I raised in the first post? 😃

shiftkey:
> Is there any reason there might be discrepancies caused by the order of the tests in the file? (See the builds of my previous commits to this branch)

Not that I'm aware of - Mocha tests should be run in the order they're declared in the file.

> Should I be using the repository given as an example in #4154

This seems fine. 

> How should I test that the correct GitHub account is selected for a repository?

My gut feeling at this point is that `canAccessRepository` is behaviour that we want to control in the test runner, so aren't relying on the real GitHub API for some of these tests. That probably means extracting that function out - if you think that's the right approach, I'm happy to work through figuring out the next steps.

> For the former `canAccessRepository` is not called, so the first GitHub account found is returned.
> ...
> Whereas `canAccessRepository` is called when identifying from owner/name.

If you've given a full URL we'll look for an existing account associated with that host. I think the latter check is there to confirm that `owner/name` does match a real repository on GitHub (or GitHub Enterprise). Unless you've got concerns about that behaviour, I'm happy to leave that as-is.

denitsa-cm:
Any idea when this will be merged? :)

wallaby.js
```diff
@@ -0,0 +1,17 @@
+module.exports = function (wallaby) {
```
iAmWillShepherd: Curious if you intended to commit this file? Wallaby.js is not free software, so we'd prefer not to use it here.

JordanMussi: No, my mistake I didn't mean to commit that. I was testing out wallaby while doing this PR.

shiftkey:
> As long as @shiftkey is fine with the strange problem I was having with the tests which then started passing only after adding console.log statements then it's just about testing the correct account is returned.

I'm not, and I'm assigning myself back to this to dig into it more. The tests are great here, and I might get them reviewed and merged separately to figuring out what we're doing with this fix.

shiftkey:
Alright, I went a slightly different path with #5200 but I think that addresses all the outstanding issues here.

I'm going to close this out in favour of that, but feel free to :eyes: it over and add your feedback there.

Thanks again @JordanMussi for the time spent here!

PR #4365, opened 2018-03-30
Fix lfs log file creation

### LINKED ISSUE #4307: Failure to create the LFS progress tracking file should not abort a clone
#4304 describes a scenario where the clone operation fails because we're unable to create a temporary file for Git LFS to write its progress output to. I think we should be much more lenient and default to logging the error and not tracking LFS progress in the rare case where we're unable to create that file.

The function which creates the file is `createLFSProgressFile`: https://github.com/desktop/desktop/blob/acda0e0b95a0cee29c7de4808e1023c038fb49e9/app/src/lib/progress/lfs.ts#L6-L17

It's used from `executionOptionsWithProgress` https://github.com/desktop/desktop/blob/acda0e0b95a0cee29c7de4808e1023c038fb49e9/app/src/lib/progress/from-process.ts#L18-L38

We should catch any errors from `createLFSProgressFile` inside of `executionOptionsWithProgress` and continue on without LFS progress tracking.

MagicMarvMan:
fixes #4307

### DIFF
--- a/app/src/lib/progress/from-process.ts
+++ b/app/src/lib/progress/from-process.ts
@@ -26,11 +26,10 @@ export async function executionOptionsWithProgress(
     try {
       lfsProgressPath = await createLFSProgressFile()
       env = { GIT_LFS_PROGRESS: lfsProgressPath }
-    } catch(e) {
-      console.log("Error writen LFS log: "+e)
+    } catch (e) {
+      console.log('Error writen LFS log: ' + e)
       env = { GIT_LFS_PROGRESS: null }
     }
-
   }
 
   return merge(options, {
--- a/app/src/lib/progress/from-process.ts
+++ b/app/src/lib/progress/from-process.ts
@@ -23,8 +23,14 @@ export async function executionOptionsWithProgress(
   let lfsProgressPath = null
   let env = {}
   if (options.trackLFSProgress) {
-    lfsProgressPath = await createLFSProgressFile()
-    env = { GIT_LFS_PROGRESS: lfsProgressPath }
+    try {
+      lfsProgressPath = await createLFSProgressFile()
+      env = { GIT_LFS_PROGRESS: lfsProgressPath }
+    } catch(e) {
+      console.log("Error writen LFS log: "+e)
+      env = { GIT_LFS_PROGRESS: null }
+    }
+
   }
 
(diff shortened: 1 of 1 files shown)


app/src/lib/progress/from-process.ts
```diff
@@ -23,8 +23,13 @@ export async function executionOptionsWithProgress(
   let lfsProgressPath = null
   let env = {}
   if (options.trackLFSProgress) {
-    lfsProgressPath = await createLFSProgressFile()
-    env = { GIT_LFS_PROGRESS: lfsProgressPath }
+    try {
+      lfsProgressPath = await createLFSProgressFile()
+      env = { GIT_LFS_PROGRESS: lfsProgressPath }
+    } catch (e) {
+      console.log('Error writen LFS log: ' + e)
```
JordanMussi: Perhaps use the `log` global? `log.error('Error writing LFS progress file:', e)`

JordanMussi: Example from another part of the codebase:
https://github.com/desktop/desktop/blob/614f8489850fd1f409a41490c8a983f8c8160017/app/src/lib/stats/stats-store.ts#L143-L145

MagicMarvMan:
I fixed it, thank you!

app/src/lib/progress/from-process.ts
```diff
@@ -23,8 +23,13 @@ export async function executionOptionsWithProgress(
   let lfsProgressPath = null
   let env = {}
   if (options.trackLFSProgress) {
-    lfsProgressPath = await createLFSProgressFile()
-    env = { GIT_LFS_PROGRESS: lfsProgressPath }
+    try {
+      lfsProgressPath = await createLFSProgressFile()
+      env = { GIT_LFS_PROGRESS: lfsProgressPath }
+    } catch (e) {
+      log.error('Error writing LFS progress file:', e)
```
niik: Could we drop the trailing `:` in the error message please? I know we've been a bit inconsistent here but when you're passing an error the message should stand by itself. The logging framework is responsible for the formatting of the error + message.

niik:
Circle CI is upset about me pushing to this branch for some reason. Not gonna spend time figuring out what's going on there.

PR #4378, opened 2018-04-02
Tweak a handful of styles for comparing branches

donokuda:
- Octicon in text field is dimmed unless field is in focus
- Tweaked some padding around the octicon in the text field
- Updated behind/ahead text styles (a little smaller and a little grayer)
- Bolded number of commits in merge area
- Changed text field type to `search` to get that "clear field ✖️button" for free

image

image

/cc @desktop/design


iAmWillShepherd:
This looks great! 💖

app/styles/ui/history/_history.scss
```diff
@@ -80,16 +81,40 @@
...
 
-  .branches-list-item .description {
-    margin-right: 0;
+  .branches-list-item {
+    .description {
+      margin-right: 0;
+    }
+
+    .name {
+      max-width: 60%;
```
niik: This isn't immediately clear to me. What do we gain by capping the max width as opposed to disabling shrinking of the description? Is this some workaround to get truncation to work in the flexbox?

donokuda: Whoops, this was left over from some stuff I was doing trying to get the commit counters to look good by capping how wide the branch name gets. 😅Turns out it didn't make any real difference in the end. Removed in b496c91

Here's what it looks like without that style:

screen shot 2018-04-03 at 9 29 13 am

PR #4379, opened 2018-04-03
smooth out filtering behaviour

shiftkey:
This addresses an issue I have with the current filter text setup, which can be best illustrated by this GIF that occurs when I first click into it and click `p` to filter on PRs (I have multiple of them open locally):

What seems to be happening:

 - inside `FilterList.createStateUpdate` sees that there's some filter text set, so it "selects" the first branch it has filtered

https://github.com/desktop/desktop/blob/4405c3bb22d33b92ed4682072e7e205420c88ef9/app/src/ui/lib/filter-list.tsx#L419-L423

 - as `selectedItem` is now different, `Compare.onSelectionChanged` is fired with the first branch in the filtered list
 - the `Compare` tab sets the filter text to the first branch name updates it's state
 - `FilterList` filters on this new filter text, it matches only one branch, everything else is hidden

TODO:

This is roughly where I want to take it, feel free to clarify things if I've missed some context:

 - [x] updating filter text should show valid branches
 - mouse click on branch in list should
    - [x] compare that branch to the current one
    - [x] update filter text to full name
 - <kbd>Enter</kbd> when focused in filter text should
    - [x] compare with the focused branch in the list
    - [x] update filter text to full name
 - ~~keyboard navigation and <kbd>Enter</kbd> in list~~ the use of a custom textbox here breaks this, punting to another PR as I need to wrap my head around the internals better

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -413,26 +413,29 @@ export class CompareSidebar extends React.Component<
     this.setState({ filterText: text })
   }
 
+  private clearFilterState = () => {
+    this.setState({
+      selectedBranch: null,
+      filterText: '',
+      compareType: CompareType.None,
+    })
+
+    this.props.dispatcher.loadCompareState(
+      this.props.repository,
+      DisplayHistory
+    )
+  }
+
   private onSelectionChanged = (branch: Branch | null) => {
     if (branch === null) {
-      this.setState({
-        selectedBranch: null,
-        filterText: '',
-        compareType: CompareType.None,
-      })
-
-      this.props.dispatcher.loadCompareState(
-        this.props.repository,
-        DisplayHistory
-      )
-
+      this.clearFilterState()
       return
     } else {
       const { branches } = this.branchState
       const selectedBranch = branches.find(b => b.name === branch.name) || null
 
       if (selectedBranch === null) {
-        this.onSelectionChanged(null)
+        this.clearFilterState()
         return
       }
 
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -436,26 +436,31 @@ export class CompareSidebar extends React.Component<
       return
     }
 
-    const { branches } = this.branchState
-    const selectedBranch = branches.find(b => b.name === branch.name) || null
-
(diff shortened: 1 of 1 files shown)


app/src/ui/history/compare.tsx
```diff
@@ -413,41 +415,51 @@ export class CompareSidebar extends React.Component<
...
+
+    this.props.dispatcher.loadCompareState(
+      this.props.repository,
+      DisplayHistory
+    )
+  }
+
+  private onSelectionChanged = (
+    branch: Branch | null,
+    source: SelectionSource
```
shiftkey: @iAmWillShepherd just a heads up, I think we've gone back and forth about this in the past but I think that the lack of `source` here meant the selection events from `FilterList` were a bit confusing. Hope that helps with consuming the component now as part of the Compare list.

iAmWillShepherd: Oh this is great!!! You actually solved the thing I was working through today. I didn't think about adding a source param to that function. I thought I had did something else wrong and decided to look at how I passed the textbox component to the filter-list component.

shiftkey:
Ready for review 🍎

app/src/ui/history/compare.tsx
```diff
@@ -338,20 +339,25 @@ export class CompareSidebar extends React.Component<
...
-            branch =>
-              branch.name.toLowerCase() === this.state.filterText.toLowerCase()
-          ) || null
-
-        this.props.dispatcher.loadCompareState(
-          this.props.repository,
-          DisplayHistory
-        )
-        this.setState({ selectedBranch: branch })
+        if (this.state.selectedBranch == null) {
```
iAmWillShepherd: Can we call `clearFilterState` here or does the call to `setState` in that method make it not possible?

iAmWillShepherd:
Gonna go ahead and merge since it's going into a PR branch.

app/src/ui/history/compare.tsx
shiftkey: I wasn't sure, so I left it as-is.

iAmWillShepherd: I think it should be fine either way, so I went ahead and merged it.

PR #4030, opened 2018-02-16
Add context menu item to open file in your preferred editor in the changes tab

iAmWillShepherd:
Fixes issue #4028

PR #4381, opened 2018-04-03
"open file in external editor" should provide the full path

shiftkey:
Found this one while trying to address a merge conflict by editing the file and VS Code would continue to show me a file without merge conflicts. Relates to #4030.

**Repro steps:**

 - have multiple versions of a repository on disk (I'm using `desktop` as an example)
 - use these both in VS Code (Atom does the same thing).
 - modify a file so that it appears in the `Changes` list. For this repro, I've deleted all the lines in `tslint.json`.

 - right-click and select `Open in [External Editor]`

**Expected:**

 - editor is launched, file shown matches expected diff from Desktop (no lines)

**Actual:**

 -  editor is launched, but file is different local clone (original content shown)

If you break into in the `launchExternalEditor` function in `app/src/lib/editors/launch.ts` you'll see the relative file path is passed through to the `spawn` API:

It looks like VS Code and Atom are smart enough to find a workspace containing the file, but because it lacks the repository directory it keeps choosing the wrong workspace - leading me to be very confused.

### DIFF
--- a/app/src/lib/dispatcher/dispatcher.ts
+++ b/app/src/lib/dispatcher/dispatcher.ts
@@ -603,11 +603,10 @@ export class Dispatcher {
   }
 
   /**
-   * Opens a path (can either be a folder representing a repository or the
-   * path to a file) in the external editor selected by the user.
+   * Opens a path in the external editor selected by the user.
    */
-  public async openInExternalEditor(path: string): Promise<void> {
-    return this.appStore._openInExternalEditor(path)
+  public async openInExternalEditor(fullPath: string): Promise<void> {
+    return this.appStore._openInExternalEditor(fullPath)
   }
 
   /**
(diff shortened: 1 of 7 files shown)


app/src/lib/dispatcher/dispatcher.ts
```diff
@@ -602,7 +602,10 @@ export class Dispatcher {
     }
   }
 
-  /** Opens a Git repository in the user provided program */
+  /**
+   * Opens a path (can either be a folder representing a repository or the
+   * path to a file) in the external editor selected by the user.
+   */
   public async openInExternalEditor(path: string): Promise<void> {
```
niik: Should the `path` parameter be renamed to `fullPath` here as well since it's been renamed elsewhere?

shiftkey:
:apple:

PR #4390, opened 2018-04-04
improving the performance of the Compare Branches experience

shiftkey:
This is myself and @iAmWillShepherd pairing on the Compare Branches experience to address a few concerns raised:

 - the ahead/behind for each branch is being computed when the component loads, leading to some repainting issues
 - the state and actions of the Compare Tab is unclear
 - components are carrying around a lot of internal state - we should push this up into `AppState`

Goals:

 - [x] refactor the `CompareState` to better align with the visual model of the Compare tab, and introduce `CompareAction` to represent actions that the component sends to update the app state
 - [x] reduce the amount of state in the `Compare` view to simplify the component
 - [x] compute and store the ahead/behind info for all refs against the current branch in a key-value store 
 - [x] emit updates when the ahead/behind on branches are computed, so that the list updates when it's expanded
 - ~~[ ] ensure foreground updates (for a selected branch) can be prioritized over background updates (for branches in the list)~~ 👢 to the next pairing session
 - [x] invalidate the compare cache when switching branches


shiftkey:
Other polish bits for tomorrow:

 - When typing in non-existent branch names the branch list is eventually empty, which resets the filter, which then displays all branches again
   - **expected behaviour:** show "no branches" message, filter text persists despite not matching any branches



 - branch selector in compare tab should have "clear filter" icon like our branch switcher does
  - **expected behaviour:** when filter text shown, <kbd>Esc</kbd> should clear filter text. When no filter text:  <kbd>Esc</kbd> should blur text box.



 - after selecting a branch to compare, switching `Compare` -> `Commit` -> `Compare` remembers most of the form state **except for the filter text**

shiftkey:
Some other little visual things to call out

1. Repro Steps (I think):
 - checkout a branch that's not master
 - compare to `master` 
 - with the compare visible, checkout `master`

You should see the compare end up in this state:



**Expected behaviour:** reset the compare to show the history of the new branch

Also note that the "No branches" graphic doesn't seem to be in the right spot relative to the container

2. The default text when a comparison has no commits is "No history" - I think this should be something like "No commits"

PR #4403, opened 2018-04-06
display the differences file size of images

### LINKED ISSUE #4380: [Feature request] Display the differences file size of images
## Description

In some of case, I have to optimize the assets. The filesize become very important information to me. Will be better if can shows the diff bytes and percents. such as `(-50.6kb 25%)`.

## Version

* GitHub Desktop: 1.1.1
* Operating system: 10.13.3

ggajos:
Fixes #4380

t1
t2


.gitignore
```diff
@@ -9,3 +9,4 @@ app/node_modules/
 .idea/
 .vs/
 .eslintcache
+*.iml
```
iAmWillShepherd: Missing a final new line.

app/src/ui/diff/image-diffs/two-up.tsx
```diff
@@ -30,37 +31,52 @@ export class TwoUp extends React.Component<ITwoUpProps, {}> {
...
+            <div className="image-diff-footer">
+              <span className="strong">W:</span> {currentImageSize.width}px |{' '}
+              <span className="strong">H:</span> {currentImageSize.height}px |{' '}
+              <span className="strong">Size:</span> {currentImageBytes.format()}
+            </div>
           </div>
         </div>
+        <div className="image-diff-summary">
+          Diff:{' '}
+          <span className={diffBytes.bytes > 0 ? 'added' : 'removed'}>
```
iAmWillShepherd: What if there is no size difference?

app/src/ui/lib/bytes.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Manipulate bytes for displaying purposes.
+ */
+export class Bytes {
```
iAmWillShepherd: I don't think we need a class for this. Can we turn these into free-standing functions?

app/src/ui/lib/bytes.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  /**
+   * Display bytes in human readable format like:
+   * - 23gb
+   * - 43bytes
+   * @param {boolean} forceSign Show `+123kb` instead of `123kb`
+   * @return {string} formatted string
+   */
+  public format(forceSign: boolean = false) {
```
iAmWillShepherd: Could we make a type or enum for this?

app/src/ui/lib/bytes.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!Number.isFinite(this.bytes)) {
+      return 'Unknown'
+    }
+    const sizes = ['bytes', 'kb', 'mb', 'gb', 'tb']
```
iAmWillShepherd: The unit for a byte is **B**, so can we update to `['B', 'KB', 'MB', 'GB', 'TB']`?

app/src/ui/lib/bytes.ts
```diff
@@ -0,0 +1,50 @@
...
+    const i = Math.floor(Math.log(Math.abs(this.bytes)) / Math.log(1024))
```
iAmWillShepherd: Could we give this a descriptive name?

app/src/ui/lib/bytes.ts
```diff
@@ -0,0 +1,50 @@
...
+    const sign = forceSign && this.bytes > 0 ? '+' : ''
```
iAmWillShepherd: Can you help me understand why we'd want the `+` symbol but not the `-` symbol?

.gitignore
ggajos: @iAmWillShepherd Thanks, added

app/src/ui/diff/image-diffs/two-up.tsx
ggajos: @iAmWillShepherd Good point, I've added some logic for this

no size diff

app/src/ui/lib/bytes.ts
ggajos: @iAmWillShepherd ok, I've refactored this

app/src/ui/lib/bytes.ts
ggajos: @iAmWillShepherd Sure, added

app/src/ui/lib/bytes.ts
ggajos: @iAmWillShepherd I think I followed the initial design to strictly, updated, added also `PB`.

app/src/ui/lib/bytes.ts
ggajos: @iAmWillShepherd Sure, updated to `sizeIndex`

app/src/ui/lib/bytes.ts
ggajos: @iAmWillShepherd Sure, the `-` symbol is added to all numbers by default. I've added also the comment in the code, hope it's clear:

```
/**
 * Display bytes in human readable format like:
 *    23GB
 *   -43B
 * It's also possible to force sign in order to get the
 * plus sign in case of positive numbers like:
 *   +23GB
 *   -43B
 */
```

ggajos:
@iAmWillShepherd Thanks for the feedback, please have a look at the updated PR

iAmWillShepherd:
@desktop/design , could you do a pass through on the SCSS changes? I've done a cursory review, so really want to make sure those changes are appropriate.

app/styles/ui/_diff.scss
```diff
@@ -286,17 +286,36 @@
...
+  flex-direction: column;
 
-  .image-diff-previous {
-    margin-right: 7.5px;
+  .image-diff-two-up {
+    display: flex;
+    align-items: center;
+
+    .image-diff-previous {
+      margin-right: 7.5px;
```
iAmWillShepherd: @donokuda is this a good place to use `var(--spacing)` since the difference in values is only 2.5px?

donokuda: Thought on replacing this with `margin-right: var(--spacing);` instead? This will create the standard spacing between the image diffs.

app/styles/ui/_diff.scss
```diff
@@ -286,17 +286,36 @@
...
+      margin-bottom: 7.5px;
```
donokuda: Is there a reason why you're using a half pixel? I tend to avoid them since they're specific to high density monitors and haven't found a compelling reason to use them instead of a whole number.

Also, are you trying to apply this margin to create space between the diff size summary below the two images? If so, it might make sense move this margin under `.image-diff-two-up` on Line 296  and remove this line and 306. This should have the same visual effect while adding the benefit of having to change code in one place.

app/styles/ui/_diff.scss
```diff
@@ -286,17 +286,36 @@
...
+    }
+    .image-diff-current {
+      margin-left: 7.5px;
+      margin-bottom: 7.5px;
+    }
```
donokuda: I think lines 304-307 can be removed if you take go forward with my suggestions above.

app/src/ui/diff/image-diffs/two-up.tsx
```diff
@@ -27,40 +29,68 @@ export class TwoUp extends React.Component<ITwoUpProps, {}> {
...
+          Diff:{' '}
+          <span
+            className={classNames({
+              added: diffBytes > 0,
+              removed: diffBytes < 0,
+            })}
+          >
+            {diffBytes !== 0
+              ? `${formatBytes(diffBytes)} | ${diffPercent}`
+              : 'No size difference'}
```
donokuda: Thoughts on adding a plus/minus sign to indicate if there was a positive or negative change in size? (For example `+2KB | +26%` or `-2KB | -26%`) It'll help make the number a little quicker to parse, especially for those with red/green color blindness.

screen shot 2018-04-30 at 10 19 25 am

shiftkey: >  It'll help make the number a little quicker to parse, especially for those with red/green color blindness.

Big 👍 for this point

app/styles/ui/_diff.scss
ggajos: `7.5` pixel was the original code; I just added this fragment to the container (that's why it's modifed). I wanted to position the `Diff: xx%` below properly. The idea behind the margin-right/bottom is to position the images in somehow in the "center" of the views. Please have a look at the image below:

diff

app/styles/ui/_diff.scss
ggajos: Thanks, good idea

ggajos: See https://github.com/desktop/desktop/pull/4403#discussion_r187312119

app/styles/ui/_diff.scss
ggajos: Unfortunately this won't work, see https://github.com/desktop/desktop/pull/4403#discussion_r187312119

app/src/ui/diff/image-diffs/two-up.tsx
ggajos: Thanks! That was my idea as well, just slipped through during some code refresh, should be added properly now.

ggajos:
@j-f1 tx, updated

ggajos:
@donokuda Please have a look at replies

Also, the build failed but without any meaningful reason https://github.visualstudio.com/Desktop/_build?buildId=103&tab=details&_a=summary any idea what might went wrong?

shiftkey:
> Also, the build failed but without any meaningful reason [snip] any idea what might went wrong?

@ggajos that failure is unrelated to the change - I've been evaluating VSTS as a CI platform, and I've restarted the build to see what's up.

ggajos:
@shiftkey I guess VSTS don't like Windows :)

ggajos:
@donokuda can you have a look if all good now?

app/src/ui/diff/image-diffs/two-up.tsx
```diff
@@ -27,40 +29,69 @@ export class TwoUp extends React.Component<ITwoUpProps, {}> {
...
+              <span className="strong">H:</span> {currentImageSize.height}px |{' '}
+              <span className="strong">Size:</span>{' '}
+              {formatBytes(this.props.current.bytes)}
+            </div>
           </div>
         </div>
+        <div className="image-diff-summary">
+          Diff:{' '}
+          <span
+            className={classNames({
```
donokuda: Should this logic be moved outside of `return()`? Only asking since I'm not familiar with this pattern.

j-f1: No, that looks reasonable to me.

nerdneha: looks fine to me 👍

ggajos:
@shiftkey do you think can we merge with failing VSTS build?

shiftkey:
@ggajos yeah, we've just been working to get 1.2.0 out the door over the last couple of days - once that's sorted we've got a few fixes like this that are ready to go out in a minor update.

app/src/ui/diff/image-diffs/two-up.tsx
```diff
@@ -27,40 +29,69 @@ export class TwoUp extends React.Component<ITwoUpProps, {}> {
...
+              added: diffBytes > 0,
+              removed: diffBytes < 0,
+            })}
+          >
+            {diffBytes !== 0
+              ? `${formatBytes(diffBytes, Sign.Forced)} | ${diffPercent}`
```
jessetan: Suggestion:  replace `| ${diffPercent}` with `(${diffPercent})`
The pipe is used earlier in this page to seperate unrelated informational units (e.g. width and height), while  absolute bytes and percentage are different ways to describe the same information.
Also because the pipe is rendered in the same green/red as the numbers next to it, it can be read as a `1` at first glance

ggajos: @jessetan I think it's nice suggestion, thanks, updated.

ggajos:
@shiftkey welcome!

Thanks all for help.

PR #4411, opened 2018-04-09
add escape hatch to disable hardware acceleration when launching 

### LINKED ISSUE #3921: Black screen
### Description

Program screen when black and had to exit out and reopen to see program.

### Version

**OS version:** Windows 10 Pro 1709 build 17083.1000

### Steps to Reproduce

1. Open program
1. Make changes in a repository 
1. Add description and commit
1. Enjoy black screen for the program

**Expected behavior:** Push changes to github

**Actual behavior:** Black screen 

**Reproduces how often:** Have not tried to reproduce

#### Additional Information

Same issues appear in chrome, could be an issue with using hardware acceleration.

shiftkey:
Fixes #3921

- [x] on launch, if environment variable set run the workaround
- [x] document this workaround
- [x] test it on Windows, confirm it emits a message to the log

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -170,6 +170,13 @@ function setAsDefaultProtocolClient(protocol: string) {
   }
 }
 
+if (process.env.GITHUB_DESKTOP_DISABLE_HARDWARE_ACCELERATION) {
+  log.info(
+    `GITHUB_DESKTOP_DISABLE_HARDWARE_ACCELERATION environment variable set, disabling hardware acceleration`
+  )
+  app.disableHardwareAcceleration()
+}
+
 app.on('ready', () => {
   if (isDuplicateInstance || handlingSquirrelEvent) {
     return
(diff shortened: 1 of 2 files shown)


shiftkey:
For reference, this is the "default" GPU settings on one of my Windows VMs:



And with this flag, the app launches and now has these GPU settings:

screen shot 2018-04-09 at 11 42 29 am

This definitely affects the snappiness (that's the non-technical term) of our virtualized lists in the app, but I feel it's a decent workaround to let users who encounter this issue run the app.

j-f1:
> `2d_canvas: "unavailable_software"`

Does that mean we can’t use `<canvas>` without hardware acceleration, or does it mean that the canvas is implemented in software rather than hardware?

shiftkey:
@j-f1 I'm pretty sure this flag means it's implemented in software. From the relevant Chromium [source](https://cs.chromium.org/chromium/src/content/browser/gpu/compositor_util.cc?type=cs&q=unavailable_software&sq=package:chromium&l=305-307):

```C++
    } else if (gpu_feature_data.status == gpu::kGpuFeatureStatusSoftware) {
      status = "unavailable_software";
    } else {
```

iAmWillShepherd:
@shiftkey, I'm 👍 on shipping the GPU deceleration (?) as a part of the next beta. Are there any gotchas we should be concerned about? Does this mean integrated graphics won't take care of the heavy lifting?

shiftkey:
@iAmWillShepherd 

> Are there any gotchas we should be concerned about? 

At the moment it's behind an environment variable, so it's opt-in and low-risk to revert later. The biggest issue I saw was that our virtualized lists feel slower to scroll, but that seems fine to live with.

> Does this mean integrated graphics won't take care of the heavy lifting?

Correct. But for these users the application doesn't work - either at launch or during some usage scenarios, so I'd like to see us have something usable for them at the moment and deal with performance things later.

PR #4417, opened 2018-04-09
draft of release notes for next beta

shiftkey:
We're reviewing the big part of the 1.2 release over in #4182, but it'd be nice to get out another beta with the fixes already on `master` before that lands. ~~I also opened #4411 which is a minor fix for an issue affecting some users trying to open the app on Windows - maybe that can go out?~~

 - [x] merge #4411 and then merge into this branch


shiftkey:
:apple:

PR #4418, opened 2018-04-10
update node @types to v8

shiftkey:
Turns out we're still working against the Node 7 APIs both in the app and our related tooling. Oops.

This upgrade highlighted two areas where our handling of return values needed to be improved:

 - `process.env` now correctly returns `string | undefined` when indexing into it
 - the result from `URL.parse()` now has a different shape for `query`, which is also an index. This now requires handling `string | string[] | undefined` for querystring terms. 

The second bullet point is a bit contentious, but I poked at this API in Chrome Dev Tools and could trigger these three scenarios:

### DIFF
--- a/app/src/lib/parse-app-url.ts
+++ b/app/src/lib/parse-app-url.ts
@@ -40,6 +40,7 @@ export type URLActionType =
   | IOpenRepositoryFromPathAction
   | IUnknownAction
 
+// eslint-disable-next-line typescript/interface-name-prefix
 interface ParsedUrlQueryWithUndefined {
   // `undefined` is added here to ensure we handle the missing querystring key
   // See https://github.com/Microsoft/TypeScript/issues/13778 for discussion about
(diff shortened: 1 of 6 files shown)


app/src/lib/parse-app-url.ts
```diff
@@ -70,28 +107,23 @@ export function parseAppURL(url: string): URLActionType {
...
+    const branch = getQueryStringValue(query, 'branch')
+    const filepath = getQueryStringValue(query, 'filepath')
 
-    const pr = queryString.pr
-    const branch = queryString.branch
-    const filepath = queryString.filepath
-
-    if (pr) {
-      // if anything other than a number is used for the PR value, exit
+    if (pr != null) {
```
iAmWillShepherd: We can use the explicit check for `null` since `getQueryStringValue` returns a value or `null`

shiftkey: There's a few places where we do this beyond this, and I'd rather leave it as `!=` here to ensure it also catches `undefined`, like the previous code did. How strongly do you feel about this?

iAmWillShepherd: Not very.

PR #4424, opened 2018-04-10
Use an actual value to enable preview features

iAmWillShepherd:
This PR just updates the logic around enabling preview features and the associated docs.

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -14,7 +14,10 @@ function enableDevelopmentFeatures(): boolean {
     return true
   }
 
-  if (process.env.GITHUB_DESKTOP_PREVIEW_FEATURES) {
+  if (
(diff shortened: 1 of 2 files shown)


app/src/lib/feature-flag.ts
```diff
@@ -14,7 +14,10 @@ function enableDevelopmentFeatures(): boolean {
     return true
   }
 
-  if (process.env.GITHUB_DESKTOP_PREVIEW_FEATURES) {
+  if (
+    process.env.GITHUB_DESKTOP_PREVIEW_FEATURES !== undefined &&
+    process.env.GITHUB_DESKTOP_PREVIEW_FEATURES === '1'
```
j-f1: This can just be `process.env.GITHUB_DESKTOP_PREVIEW_FEATURES === '1'` — if it’s `'1'`, it’s not `undefined`.

PR #4429, opened 2018-04-10
Add hint in placeholder for required commit message

### LINKED ISSUE #4354: Commit button disabled for unclear reasons
## Description

When using GitHub Desktop, the commit button is disabled until a commit summary is typed in. However, for users who are new to Git & GitHub, its not entirely clear why the button is disabled.

## Version
* GitHub Desktop: 1.1.1
* Operating system: MacOS 10.13.3

## Steps to Reproduce
1. Make some changes to a git repo
2. See that the commit button is disabled and there is no indication of why

### Expected Behavior
The Commit Summary field is highlighted in red or indicated as required in some manner

### Actual Behavior
No indication is given.

iAmWillShepherd:
Workaround for feedback reported in #4354 

**Before**
before

**After**
after

### DIFF
--- a/app/src/ui/autocompletion/autocompleting-text-input.tsx
+++ b/app/src/ui/autocompletion/autocompleting-text-input.tsx
@@ -28,6 +28,9 @@ interface IAutocompletingTextInputProps<ElementType> {
   /** Disabled state for input field. */
   readonly disabled?: boolean
 
+  /** Indicates if input field should be required */
+  readonly isRequired?: boolean
+
   /**
    * Called when the user changes the value in the input field.
    */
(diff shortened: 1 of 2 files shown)


shiftkey:
> Fixes #4354

I'm not sure this is the entire solution to what was discussed #4354 - @donokuda mentioned this work as being a short-term solution, and others in that issue were looking for some more aggressive intervention in the app.

iAmWillShepherd:
I think that something more aggressive can be tackled in a separate PR since we now provide _some_ indication as to why the commit button is disabled. What we have is good enough for now until @desktop/design comes up with something we all agree on.

shiftkey:
Updated the issue description to ensure the original issue isn't accidentally closed.

PR #4431, opened 2018-04-11
upgrade fs-extra by a few major versions

shiftkey:
We're using `fs-extra` in a couple of places for convenience, but as part of the [3.0 release](https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md#300--2017-04-27) they added support for promises throughout the API, so this is a chance for us to move away from 

 - our crafted promises that wrap asynchronous APIs
 - a bunch of our tests using blocking APIs

TODO:

 - [x] review `fs-extra` changelog to identify changes that can be made
 - [x] run through functionality that depends on `fs-extra` and confirm it's not impacted


app/test/helpers/repositories.ts
```diff
@@ -49,10 +51,10 @@ export function setupFixtureRepository(repositoryName: string): string {
     entry => Path.basename(entry.path) === '_git'
   )
 
-  submodules.forEach(entry => {
+  submodules.forEach(async entry => {
```
niik: This change looks like it's moving us from synchronous to asynchronous without awaiting. I don't know if these submodule directories needs to be renamed in sequence or not but I would expect `setupFixtureRepository` to guarantee that all operations are completed when its resultant promise resolves.

Either we make this a regular `for` statement and await each or (if we don't care about the internal order) we map submodules into promises and await all of them using `Promise.all`

shiftkey: @niik good shout - I'll ensure this isn't fire-and-forget

shiftkey:
:apple:

PR #4459, opened 2018-04-17
Don't show option to not show message again when discarding more than one file

iAmWillShepherd:
Addresses a loose-end in #4426.

Before, when discarding all files, we would show the checkbox that disables the confirmation dialog. This doesn't make much sense since that setting is ignored when discarding all files. This PR addresses that by only showing the checkbox when there is exactly one change being discarded.

**Discard**
discard

**Discard All**
discard_all

- [x] Update PopupType to include a bit of context on how it's being used
- [x] Move away from using the number of files to determine if we should show option

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -210,11 +210,6 @@ export enum PopupType {
   DeletePullRequest,
 }
 
-export enum DiscardChangesSource {
-  manualSelection,
-  all,
-}
-
 export type Popup =
   | { type: PopupType.RenameBranch; repository: Repository; branch: Branch }
   | {
@@ -227,7 +222,7 @@ export type Popup =
       type: PopupType.ConfirmDiscardChanges
       repository: Repository
       files: ReadonlyArray<WorkingDirectoryFileChange>
-      source: DiscardChangesSource
+      forceConfirmation: boolean
     }
   | { type: PopupType.Preferences; initialSelectedTab?: PreferencesTab }
   | { type: PopupType.MergeBranch; repository: Repository }
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -2,7 +2,6 @@ import { Account } from '../models/account'
 import { CommitIdentity } from '../models/commit-identity'
 import { IDiff } from '../models/diff'
 import { Repository } from '../models/repository'
-import { IAheadBehind } from './git'
 import { Branch } from '../models/branch'
 import { Tip } from '../models/tip'
 import { Commit } from '../models/commit'
@@ -26,6 +25,7 @@ import { CloneRepositoryTab } from '../models/clone-repository-tab'
 import { BranchesTab } from '../models/branches-tab'
 import { PullRequest } from '../models/pull-request'
 import { IAuthor } from '../models/author'
+import { IAheadBehind } from './git'
 
 export { ICommitMessage }
 export { IAheadBehind }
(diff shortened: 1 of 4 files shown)


say25:
So how does this effect #4188? Are we expecting users to have to click to confirm if multi-selecting some files to discard?

iAmWillShepherd:
@say25 that is correct. Is there a use-case where this breaks down?

say25:
It’s just kinda weird to me that after manually multi selecting I have to hit the confirm to discard. To me only when I do discard all should I have to confirm. But honestly it is a matter of opinion

iAmWillShepherd:
@say25, I can get behind that idea. Will make the appropriate changes. 👍

iAmWillShepherd:
@say25, I've gone ahead and updated so this PR plays nicely with the work you've done. 🥇

app/src/lib/app-state.ts
```diff
@@ -2,7 +2,6 @@ import { Account } from '../models/account'
 import { CommitIdentity } from '../models/commit-identity'
 import { IDiff } from '../models/diff'
 import { Repository } from '../models/repository'
-import { IAheadBehind } from './git'
```
shiftkey: :art: tidy up the diff by moving this line back to here?

app/src/lib/app-state.ts
```diff
@@ -222,6 +222,7 @@ export type Popup =
       type: PopupType.ConfirmDiscardChanges
       repository: Repository
       files: ReadonlyArray<WorkingDirectoryFileChange>
+      forceConfirmation: boolean
```
shiftkey: We've not really been good at writing docs for our popup types, but would you mind adding some docs to this field to help the caller understand what it represents?

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -65,15 +66,17 @@ export class DiscardChanges extends React.Component<
...
-          <Checkbox
-            label="Do not show this message again"
-            value={
-              this.state.confirmDiscardChanges
-                ? CheckboxValue.Off
-                : CheckboxValue.On
-            }
-            onChange={this.onCheckboxChanged}
-          />
+          {!this.props.forceConfirmation ? (
```
shiftkey: The `!` here made me question what the right flow was. I gather we're aiming for:

 - if `forceConfirmation` is true, don't show the checkbox
 - otherwise, show the checkbox with the confirmation

If that's the case, could we make this a bit easier to follow - extract it to a function rather than inlining?

shiftkey:
Looks like some backups for unmerged files have crept into this PR:

shiftkey:
Dropped the `ready-for-review` label until #4510 is merged and this PR can be updated/rebased.

iAmWillShepherd:
💯

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -88,6 +81,27 @@ export class DiscardChanges extends React.Component<
     )
   }
 
+  private renderCheckbox() {
```
niik: Could we rename this to something a little bit less generic? Such as `renderConfirmDiscardChanges`? And similarly rename the `onCheckboxChanged` to `onConfirmDiscardChangesChanged`?

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -15,6 +15,7 @@ interface IDiscardChangesProps {
   readonly dispatcher: Dispatcher
   readonly files: ReadonlyArray<WorkingDirectoryFileChange>
   readonly confirmDiscardChanges: boolean
+  readonly showDiscardChangesSetting: boolean
```
niik: Could we document this please?

iAmWillShepherd:
💯

PR #4477, opened 2018-04-19
extract nextSelectableRow into stateless function

shiftkey:
This makes our core `List` component a bit less complicated by extracting `nextSelectableRow` into a stateless function named `findNextSelectableRow`, for a few reasons:

 - it's only internal state it's checking is a couple of props - these can easily enough be converted to parameters, making it testable
 - I think the name needed to be a bit clearer that it may return `null`
 - having to call into `List` to get this value meant storing a ref to the component - turns out this tidies up `AutocompletingTextInput` a bit because that's no longer necessary

TODO:

 - [x] confirm docs are accurate and detailed enough
 - [x] test interactions in `List`
 - [x] test interactions in `AutocompletingTextInput`


app/src/ui/lib/list/selection.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  const { direction, row } = action
+  const wrap = action.wrap === undefined ? true : action.wrap
+
+  // If we've been given a row that's out of bounds
+  // we'll coerce it to a valid index starting either
+  // at the bottom or the top depending on the direction.
+  //
+  // Given a row that would be below the last item and
```
iAmWillShepherd: I know this comment is old, but it's pretty confusing.

As I understand it, we need to restrict the row to be within a range of 0 to rowCount-1 inclusive.
* Given a row GTE `rowCount` and an upward direction; set to last row
* Given a row GTW `rowCount` and a downward directions; set to first row
* Given a row LT 0 and an upward direction; set to last row
* Given a row LT 0 and a downward direction; set to first row

Could we make the comment say that? I think it's a bit clearer and less wordy

app/src/ui/lib/list/selection.ts
```diff
@@ -0,0 +1,124 @@
...
+      // We've reached the top, wrap around to the bottom
+      // if we're allowed to or give up
+      if (wrap) {
+        currentRow = rowCount - 1
+      } else {
+        break
+      }
+    }
+
+    if (canSelectRow(currentRow) && row !== currentRow) {
```
iAmWillShepherd: We should swap the order of these arguments since it could result in doing less work.

shiftkey:
:apple:

PR #4506, opened 2018-04-23
always fast-forward the default branch

shiftkey:
This is a follow up to #4447 after I got some feedback that the latest beta wasn't fast-forwarding `master`. We were skipping _all_ branches here for performance reasons, but I think we should handle `master` differently because it saves a lot of friction here like:

 - keeping `master` up-to-date means you don't need to switch to `master`, pull `master`, switch back, and then merge.
 - comparing locally means you don't need to consciously differentiate `origin/master` and `master`
 - the GitHub flow is centered on the default branch, creating new branches will mean you're always working from the latest code

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2196,12 +2196,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const currentBranchName =
       tip.kind === TipState.Valid ? tip.branch.name : null
 
-    // A branch is only eligible for being fast forwarded if:
-    //  1. It's local.
-    //  2. It's not the current branch.
-    //  3. It has an upstream.
-    //  4. It's not ahead of its upstream.
-
     let eligibleBranches = branches.filter(b =>
       eligibleForFastForward(b, currentBranchName)
     )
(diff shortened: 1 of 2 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -2211,11 +2211,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
     if (eligibleBranches.length >= FastForwardBranchesThreshold) {
       log.info(
-        `skipping fast-forward work because there are ${
+        `skipping fast-forward for all branches as there are ${
           eligibleBranches.length
         } local branches - this will run again when there are less than ${FastForwardBranchesThreshold} local branches tracking remotes`
       )
-      return
+
+      const defaultBranch = state.branchesState.defaultBranch
```
iAmWillShepherd: This looks like the fifth reason a branch would be eligible for an update. Can we add that the default branch is always eligible to be fast-forwarded as a comment?

shiftkey: :thumbsup: this also reminds me I need to do the same checks in here to ensure I'm not currently on `master`:

https://github.com/desktop/desktop/blob/00669f788655ac4606ee8c4d61e46e3a75f62e1b/app/src/lib/stores/app-store.ts#L2204-L2210

shiftkey:
:apple:

app/src/lib/stores/app-store.ts
vanfumin08: Master

PR #4508, opened 2018-04-23
Add some UI polish for relationship between branches feature

donokuda:
This pull request brings in a few polish items for the [Relationship Between Branches feature](https://github.com/desktop/desktop/issues/3956)

---

### Updated no commit messaging for Behind & Ahead states
_Also removed the "There are no commits" messaging next to the merge button_

I thought the messaging could be improved by explaining why no commits are displayed. Another benefit is a "success" message when someone merges commits from the branch they are comparing against.

📷 **Commits Behind (Before)**

compare-before-empty

📷 **Commits Behind (After)**

compare-master-before-empty

📷 **Commits Ahead (Before)**

compare-ahead-empty

📷 **Commits Ahead (After)**

compare-after-empty

---

### Fix messaging in the merge commit area

1. Removed `)` at the end of the "This will merge..." messaging
2. Remove "There are no commits" message under merge button when current branch is up to date with the compared branch

📷 **Before**

compare-master-merge-message

📷 **After**

compare-master-merge-message

compare-master-before-empty

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -265,7 +265,7 @@ export class CompareSidebar extends React.Component<
       )
     }
 
-    return <div className="merge-message">There are no commits</div>
+    return null
   }
 
   private onTabClicked = (index: number) => {
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -174,21 +174,26 @@ export class CompareSidebar extends React.Component<
 
     let emptyListMessage
     if (compareState.formState.kind === ComparisonView.Ahead) {
-        const currentlyComparedBranchName = compareState.formState.comparisonBranch.name
-        emptyListMessage = (
-          <p>
-            The compared branch (<Ref>${currentlyComparedBranchName}</Ref>)
-            is up to date with your branch
-          </p>
-        )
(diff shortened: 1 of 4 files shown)


app/src/ui/history/compare.tsx
```diff
@@ -171,10 +172,31 @@ export class CompareSidebar extends React.Component<
...
+
+      emptyListMessage = (
+        <p>
+          Your branch is up to date with the compared branch (<Ref>
+            {currentlyComparedBranchName}
+          </Ref>)
+        </p>
+      )
+    } else {
+      emptyListMessage = 'No history'
```
nerdneha: the only kinds of `formState.kind`s are Ahead and Behind so this can get simplified because I don't think it's possible to hit the "else" case.
if my hunch is right, then the `<p>` tags so that can get pulled up to the component that puts in `{emptyListMessage}` and leaves us with a simpler ternary

iAmWillShepherd: There are actually 3 options for `Form.kind`: `none`, `ahead`, `behind`.  You could use a nested ternary instead, but I have opinions about those.

iAmWillShepherd:
@nerdneha , I've removed myself from reviewing since I've contributed to this PR.

PR #4509, opened 2018-04-23
Wire up select all edit menu item to multi-select lists

### LINKED ISSUE #3821: no keyboard shortcut for selecting all entries in changes list
### Version

**GitHub Desktop version:** 1.0.12

**OS Version:** macOS

**Expected behavior:**
- Command-A selects all modified files

**Actual behavior:**
- Command-A selects some text.

niik:
Fixes #3821

This is a continuation of #4188 that wires up the "Select All" Edit menu item such that when keyboard focus resides within a List (which supports range or multi selection) the list deals with the event rather than the default electron action which selects all selectable text in the currently focused window.

We do this by moving the handling of the select-all menu item to the renderer and dispatching a custom DOM event to the element where keyboard focus currently resides and letting it bubble. Components then have the option to intercept that event and cancel the default action.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -307,10 +307,12 @@ export class App extends React.Component<IAppProps, IAppState> {
     })
 
     const target = document.activeElement
+    const webContents = remote.getCurrentWebContents()
 
     if (target.dispatchEvent(event)) {
+      webContents.selectAll()
     } else {
-      remote.getCurrentWebContents().unselect()
+      webContents.unselect()
     }
   }
 
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -298,7 +298,11 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   /**
-   * Handler for the edit menu item 'Select All'.
+   * Handler for the 'select-all' menu event, dispatches
+   * a custom DOM event originating from the element which
+   * currently has keyboard focus. Components have a chance
+   * to intercept this event and implement their own 'select
+   * all' logic.
    */
   private selectAll() {
     const event = new CustomEvent('select-all', {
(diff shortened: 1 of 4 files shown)


shiftkey:
@niik this feels like it will close out #3821, right?

niik:
Yup!

app/src/ui/lib/list/list.tsx
```diff
@@ -323,9 +336,49 @@ export class List extends React.Component<IListProps, IListState> {
...
+  }
+
   private onRef = (element: HTMLDivElement | null) => {
+    if (element === null && this.list !== null) {
+      this.list.removeEventListener('select-all', this.onSelectAll)
+    }
+
     this.list = element
 
+    if (element) {
```
shiftkey: :art: `element != null`

app/src/ui/app.tsx
```diff
@@ -290,11 +290,31 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+   * to intercept this event and implement their own 'select
+   * all' logic.
+   */
+  private selectAll() {
+    const event = new CustomEvent('select-all', {
+      bubbles: true,
+      cancelable: true,
+    })
+
+    if (document.activeElement.dispatchEvent(event)) {
```
iAmWillShepherd: Can we be explicit about what version of false we're checking against.

shiftkey: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) this seems to only be a `boolean`. Not sure what else we can do here...

iAmWillShepherd: @niik, you can ignore this. shiftkey and I discussed earlier.

niik:
🔩

PR #4510, opened 2018-04-23
Refactor discardChanges

iAmWillShepherd:
While working on #4459, I ran into a problem where right-clicking discard changes would not offer the option to never show a confirmation dialog. This is due to the changes introduced over in #4188.

I've updated the function to take a single argument and I’m also checking if there is exactly one file because we want to do what we did before (not fallback to discard all changes).

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -193,27 +193,29 @@ export class ChangesList extends React.Component<
     this.props.onDiscardAllChanges(this.props.workingDirectory.files)
   }
 
-  private onDiscardChanges = (paths: string | string[]) => {
+  private onDiscardChanges = (files: ReadonlyArray<string>) => {
     const workingDirectory = this.props.workingDirectory
 
-    if (paths instanceof Array) {
-      const files: WorkingDirectoryFileChange[] = []
-      paths.forEach(path => {
-        const file = workingDirectory.files.find(f => f.path === path)
-        if (file) {
-          files.push(file)
+    if (files.length === 1) {
+      const modifiedFile = workingDirectory.files.find(f => f.path === files[0])
+
+      return modifiedFile && this.props.onDiscardChanges(modifiedFile)
+    } else {
+      const modifiedFiles: Array<WorkingDirectoryFileChange> = []
+
+      modifiedFiles.forEach(file => {
+        const modifiedFile = workingDirectory.files.find(
+          f => f.path === file.path
+        )
+
+        if (modifiedFile !== undefined) {
+          modifiedFiles.push(file)
         }
       })
-      if (files.length) {
-        this.props.onDiscardAllChanges(files)
-      }
-    } else {
-      const file = workingDirectory.files.find(f => f.path === paths)
-      if (!file) {
-        return
-      }
 
-      this.props.onDiscardChanges(file)
+      if (modifiedFiles.length) {
+        this.props.onDiscardAllChanges(modifiedFiles)
+      }
     }
   }
 
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -208,7 +208,7 @@ export class ChangesList extends React.Component<
       files.forEach(file => {
         const modifiedFile = workingDirectory.files.find(f => f.path === file)
 
-        if (modifiedFile !== undefined) {
+        if (modifiedFile != null) {
           modifiedFiles.push(modifiedFile)
         }
       })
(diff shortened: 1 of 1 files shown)


app/src/ui/changes/changes-list.tsx
```diff
@@ -193,27 +193,29 @@ export class ChangesList extends React.Component<
...
-      paths.forEach(path => {
-        const file = workingDirectory.files.find(f => f.path === path)
-        if (file) {
-          files.push(file)
+    if (files.length === 1) {
+      const modifiedFile = workingDirectory.files.find(f => f.path === files[0])
+
+      return modifiedFile && this.props.onDiscardChanges(modifiedFile)
+    } else {
+      const modifiedFiles: Array<WorkingDirectoryFileChange> = []
```
shiftkey: :art: `const files = new Array<WorkingDirectoryFileChange>()`

app/src/ui/changes/changes-list.tsx
```diff
@@ -193,27 +193,29 @@ export class ChangesList extends React.Component<
...
-    if (paths instanceof Array) {
-      const files: WorkingDirectoryFileChange[] = []
-      paths.forEach(path => {
-        const file = workingDirectory.files.find(f => f.path === path)
-        if (file) {
-          files.push(file)
+    if (files.length === 1) {
+      const modifiedFile = workingDirectory.files.find(f => f.path === files[0])
+
+      return modifiedFile && this.props.onDiscardChanges(modifiedFile)
```
shiftkey: I'm not the greatest fan of the implicit `null` comparison here as well as the `return void`. Could we make this a bit easier to read?

```ts
if (modifiedFile != null) {
  this.props.onDiscardChanges(modifiedFile)
}
```

app/src/ui/changes/changes-list.tsx
```diff
@@ -193,27 +193,29 @@ export class ChangesList extends React.Component<
...
-        this.props.onDiscardAllChanges(files)
-      }
-    } else {
-      const file = workingDirectory.files.find(f => f.path === paths)
-      if (!file) {
-        return
-      }
 
-      this.props.onDiscardChanges(file)
+      if (modifiedFiles.length) {
```
shiftkey: :art: `modifiedFiles.length > 0`

iAmWillShepherd:
💯

app/src/ui/changes/changes-list.tsx
```diff
@@ -193,27 +193,31 @@ export class ChangesList extends React.Component<
...
-      const files: WorkingDirectoryFileChange[] = []
-      paths.forEach(path => {
-        const file = workingDirectory.files.find(f => f.path === path)
-        if (file) {
-          files.push(file)
-        }
-      })
-      if (files.length) {
-        this.props.onDiscardAllChanges(files)
+    if (files.length === 1) {
```
agisilaos: Quick question @iAmWillShepherd: Why you used 3 `===` here?

j-f1: We tend to use `===` in JS because it’s more type-safe.

agisilaos: Thank you @j-f1, makes sense 👍

shiftkey: Yeah, the general rule of thumb we have is loose equality when dealing with the `undefined | null` situation, and strict equality for everything else.

MDN has [a nice writeup](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) of the differences between the two (and some other approaches).

app/src/ui/changes/changes-list.tsx
```diff
@@ -193,27 +193,31 @@ export class ChangesList extends React.Component<
...
+        this.props.onDiscardChanges(modifiedFile)
       }
     } else {
-      const file = workingDirectory.files.find(f => f.path === paths)
-      if (!file) {
-        return
-      }
+      const modifiedFiles = new Array<WorkingDirectoryFileChange>()
+
+      modifiedFiles.forEach(file => {
```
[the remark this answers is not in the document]
iAmWillShepherd: o0o0, that's a big one!

iAmWillShepherd:
💯

app/src/ui/changes/changes-list.tsx
```diff
@@ -193,27 +193,29 @@ export class ChangesList extends React.Component<
...
-      if (!file) {
-        return
-      }
+      const modifiedFiles = new Array<WorkingDirectoryFileChange>()
+
+      files.forEach(file => {
+        const modifiedFile = workingDirectory.files.find(f => f.path === file)
 
-      this.props.onDiscardChanges(file)
+        if (modifiedFile !== undefined) {
```
shiftkey: Alright, I'm gonna be _really_ pedantic here and suggest we do `!= null` here to be consistent with line 202:

```ts
if (modifiedFile != null) {
```

iAmWillShepherd:
💯

PR #4516, opened 2018-04-24
asRange tidy up

shiftkey:
I was talking with @nerdneha last night about [dotted ranges](https://git-scm.com/docs/gitrevisions#_dotted_range_notations) in Git and realised that I should make `asRange` a bit more explicit, given that `..` and `...` have very different meaning in Git.

TL;DR:

 - `revRange` aka `..` is the one that we should be using everywhere in Compare Branches, as this is the "show everything reachable from `to` but exclude things reachable from `from`" case.
 - `revSymmetricDifference` aka `...` is "the set of commits that are reachable from either `from` or `to` but not from both" and we have one usage of that


app/src/lib/git/rev-list.ts
```diff
@@ -4,14 +4,29 @@ import { Repository } from '../../models/repository'
...
+ * `from`.
  *
  * Each parameter can be the commit SHA or a ref name.
  *
  * @param from The start of the range
  * @param to The end of the range
  */
-export function asRange(from: string, to: string) {
+export function asDistinctRange(from: string, to: string) {
+  return `${from}..${to}`
```
shiftkey: You may note here that `asRange` was previously a wrapper for `...` - after reading the docs I think these usages should be `..` because we want the unique commits between `from` and `to`.

I've dropped the `ready-for-review` label here to ponder and test some more to ensure this is the right functionality for the compare tab experience, let me know if you have any other feels.

shiftkey: Nevermind, the existing behaviour was accurate - I'll update this PR to make it clearer.

shiftkey:
Alright, this is ready for review again. I'm totally open to naming these better, but I wanted to stick close to what Git calls these, which is hard!

agisilaos:
We should definitely consider a different name for `asDistinctRange` & `asSymmetricDifferenceRange`, because when I personally read these variables names the `...` & `..` aren't even a thought in my mind to make the "connection".

nerdneha:
The best I could come up with is
`getRangeFrom`
`getRangesBetween`

iAmWillShepherd:
I think `revRange` and `revSymmetricDifferenceRange` could work. They keep the names that are used in Git documentation, which is my preference.

shiftkey:
After being away from it for a few days I settled on `revRange` (shorthand for "revision range") and `revSymmetricDifference` (shorthand for "revision symmetric difference") for the names which is mostly what @iAmWillShepherd proposed but I dropped `Range` from the second example to potentially avoid confusion from interpreting the latter as a range.

shiftkey:
:apple:

PR #4517, opened 2018-04-24
add compare value into cache

shiftkey:
This addresses a UX issue where the cached value when selecting a branch isn't kept around:

From https://github.com/desktop/desktop/pull/4182:

> Sometimes when I switch to a new branch and go to compare mode, the ahead/behinds take a while to load (even though we already calculate them after we click on them). 
>
> 
>
> after reading the code, that made sense; in fact that wasn't the part that felt off when i was testing; it was the fact that the component was calculating the ahead/behind when i clicked on a compare-branch and then it felt like the branch list component hadn't "remembered". programmatically i get that it's queued and it'll show up soon but still felt weird as a user.

 - ~~[ ] investigate corner case when background caching seems to stop once it gets to the currently selected compare branch??!?!?~~ i think this is just an interaction issue, and will eventually come good

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -820,6 +820,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
           commitSHAs: compare.commits.map(commit => commit.sha),
         }))
 
+        const tip = gitStore.tip
+
+        let currentSha: string | null = null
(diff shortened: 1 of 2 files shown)


iAmWillShepherd:
Merging this so we have this fix in the latest beta. #JOLO.

PR #4522, opened 2018-04-25
promote the compare tab to the beta channel

shiftkey:
This is necessary to ensure users on the beta channel get the new Compare tab.


nerdneha:
holding on merging this until Monday when we're more staffed :)

niik:
After discussing this with @nerdneha earlier we decided to go ahead and get ready for a beta release today

Issue #4526, opened 2018-04-25
The menu item Branch -> Create Pull Request is disabled / greyed out


tobireif:
## Description

When I open the GitHup repo in the browser -> "Your recently pushed branches" I can hit "Compare & Pull Request", it works - it shows all my commits I made in that branch.

But in GitHub Desktop the menu item Branch -> Create Pull Request is disabled / greyed out.

## Versions
* GitHub Desktop 1.1.1
* MacOS 10.13.4 (17E202)

agisilaos:
Hello @tobireif, I think that you're searching for the equivalent of `Compare & Pull Request`.
In Github Desktop is the `Compare on Github`. You can find it under the Branch menu

tobireif:
Branch -> `Compare on Github` is greyed out as well 😐

shiftkey:
@tobireif that button state is based off of two fields:

https://github.com/desktop/desktop/blob/d5b02678ed528551467fce823b8f3acefd059e36/app/src/lib/menu-update.ts#L229-L232

Can you share a screenshot of the app at the time - it might help us reason with why the menu item might be disabled?

tobireif:
The project is hosted on GitHub, and I had published the branch (I can view it at github.com). Here's the screenshot:

<img width="1440" alt="screenshot" src="https://user-images.githubusercontent.com/80411/39302418-dce7dca6-4952-11e8-89d8-d2da5b45b4a0.png">

niik:
Hey @tobireif 

![screen shot 2018-04-30 at 17 02 20](https://user-images.githubusercontent.com/634063/39434245-4cdf4fce-4c98-11e8-9cf4-3b2c1b3a56a7.png)

That little computer screen icon from your screenshot tells me that GitHub Desktop seems to think that the repository is a local one. Could you double check to see that you're still signed in by going to the `GitHub Desktop` menu item and selected `Preferences`. If you are signed in to GitHub could you please open the repository settings dialog (available under the repository menu item) and verify that the 'Primary remote repository (origin)' field matches what the repository url on GitHub.com?

tobireif:
I did `GitHub Desktop` -> `Preferences` sign in, now it works - thanks!

Perhaps it could prompt to Sign In in the future.

I thought the Create Pull Request feature would create the PR in GitHub Desktop, but it opens the new-pull-request page in the browser. It's not dramatically different from starting the PR manually in the web UI. But anyways, the feature works now.

Thanks!

[...]

PR #4528, opened 2018-04-25
correct menu id to prevent clash with existing compare-branch id

shiftkey:
Found this while looking at #4526 - 87a19b138f5e617b9b338d9f69b9cba6a738bb34 renamed the menu ID, but we already have a `'compare-branch'` menu ID - so this might cause issues with updating the menu state.

If someone else wants to pick this up, maybe `'compare-branch'`  should also be renamed to indicate it's related to `Compare on GitHub`

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -112,7 +112,7 @@ const allMenuIds: ReadonlyArray<MenuIDs> = [
   'repository',
   'create-branch',
   'create-commit',
-  'compare-branch',
+  'compare-to-branch',
   'show-repository-list',
   'show-branches-list',
   'open-working-directory',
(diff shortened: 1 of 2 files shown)


shiftkey:
To whoever is reviewing this, ensure that they launch the app and test out these menu states:

 - no repository selected - `Compare Branches` and `Compare on GitHub` should be disabled
 - non-GitHub repository - `Compare Branches` enabled, `Compare on GitHub` disabled
 - GitHub repository, unpublished branch - `Compare Branches` enabled, `Compare on GitHub` disabled
 - GitHub repository, published branch - `Compare Branches` and `Compare on GitHub` enabled

shiftkey:
To see this issue in action, `master` should show `Compare Branches` is disabled when viewing a valid repository:



And this menu item is enabled, which should only be set when the branch is tracking a remote branch:

PR #4529, opened 2018-04-25
Fixes the build

iAmWillShepherd:
There was one place that was using the old instance method to get the next-selectable row, this PR makes use of the new function added to replace that method.


shiftkey:
Build seems to be good now. Feel free to merge when it goes green.

PR #4439, opened 2018-04-13
Add tracking for RBB Lite

iAmWillShepherd:
Fixes #4438

**Plumbing**
- [x] Add `branchComparisons` field to `IDailyMeasures` in `stats` database
- [x] Add ~`comparisonsToMaster`~ `defaultBranchComparisons` field to `IDailyMeasures` in `stats` database
- [x] Add `mergesInitiatedFromComparison` field to `IDailyMeasures` in `stats` database
- [x] Add `updateFromDefaultBranchMenuCount` field to `IDailyMeasures` in `stats` database
- [x] Add `mergeIntoCurrentBranchMenuCount` field to `IDailyMeasures` in `stats` database
- [x] Add methods for each new measure to `statsStore`
- [x] Add a new fields for the new measures to `DefaultDailyMeasure` in `statsStore`

**Back-end (Central)**
- [x] Add measure to `desktop_usage_event.rb`
- [x] Update `usage_test.rb`
- [x] Update `fake_desktop_data` in `helper.rb`

**Database**
- [x] Open an issue on _github/analytics_ to get new field pulled into view, update airflow, and update reports

**Other**
- [x] Update the [usage-data page](https://desktop.github.com/usage-data/) on desktop.github.com 

_This depends on PRs in [central](https://github.com/github/central/pull/366), [desktop.github.com](https://github.com/github/desktop.github.com/pull/97), and [analytics](https://github.com/github/analytics/issues/347) being merged._

PR #4531, opened 2018-04-25
Wire up RBB analytics

iAmWillShepherd:
This is a continuation of work done over in #4439. That PR introduced the plumbing for analytics and this PR actually makes use of that prior work. 

This PR also wraps up the analytics piece in #3956

### DIFF
--- a/app/src/lib/dispatcher/dispatcher.ts
+++ b/app/src/lib/dispatcher/dispatcher.ts
@@ -1178,7 +1178,7 @@ export class Dispatcher {
    * Increments the `mergeIntoCurrentBranchMenuCount` metric
    */
   public recordMenuInitatedMerge() {
-    return this.appStore._recordMenuInitatedMerge()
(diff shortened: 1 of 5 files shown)


app/src/lib/dispatcher/dispatcher.ts
```diff
@@ -1173,4 +1173,25 @@ export class Dispatcher {
   public executeCompare(repository: Repository, action: CompareAction) {
     return this.appStore._executeCompare(repository, action)
   }
+
+  /**
+   * Increments the `mergeIntoCurrentBranchMenuCount` metric
+   */
+  public recordMenuInitatedMerge() {
```
nerdneha: minor ⛏ for a few of these, s/`Initated`/`Initiated`
(we should probably do a find & replace because it's in this file a few times)

iAmWillShepherd:
💯

iAmWillShepherd:
💯

PR #4533, opened 2018-04-26
Add Dockerfile Highlights

say25:
Things of Note:
1. This isn't working yet.... even though this would suggest it should work to me. https://github.com/codemirror/CodeMirror/blob/master/mode/meta.js#L39
1. This change is more invasive than previous additions to Highlighter. Wanted to loop you guys in before spending more time on it. Couldn't determine a way around passing more than extension.
1. I still need to add a test file.


niik:
> This change is more invasive than previous additions to Highlighter. Wanted to loop you guys in before spending more time on it. Couldn't determine a way around passing more than extension.

I'm happy with the approach you've taken here. We could discuss whether to call it base name or filename but I'm happy either way.

app/src/lib/highlighter/worker.ts
```diff
@@ -27,6 +27,7 @@ const workerUri = encodePathAsUrl(__dirname, 'highlighter.js')
  */
 export function highlight(
   contents: string,
+  basename: string,
   extension: string,
   tabSize: number,
   lines: Array<number>
```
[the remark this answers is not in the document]
say25: I was going for the least invasive change :/

app/src/highlighter/index.ts
```diff
@@ -152,8 +155,9 @@ function guessMimeType(contents: string) {
 
 function detectMode(request: IHighlightRequest): CodeMirror.Mode<{}> | null {
   const mimeType =
-    extensionMIMEMap.get(request.extension.toLowerCase()) ||
-    guessMimeType(request.contents)
+    extensionMIMEMap.get(
+      request.extension.toLowerCase() || request.basename.toLowerCase()
+    ) || guessMimeType(request.contents)
```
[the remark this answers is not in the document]
say25: I kinda feel like thats wrong as Dockerfile in particular is not supposed to have an extension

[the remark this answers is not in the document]
say25: Think about Dockerfile.java. Is it Java or Docker?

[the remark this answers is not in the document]
niik: I think that we might be overloading the `extensionMIMEMap` here and that's making it tricky to reason about. Since this is introducing a special case for files which have an exact basename match (seeing as a basename includes the extension if one exists) I think we should have a separate map for that. How about we create a `basenameMIMEMap` and check that first before checking the `extensionMIMEMap`?

say25:
@niik noticed PR #4764. I will wait on this PR until that one is merged down

niik:
Hey @say25, #4764 has landed so if you wanna merge master in and make the appropriate changes that'd be great!

say25:
I still have to finish implementing this but here is the corresponding Highlighter Tests PR: https://github.com/desktop/highlighter-tests/pull/27

say25:
@niik Good news; bad news.

***Good News:*** Made some progress on this branch and I figured out why the highlights aren't working. Its because I'm not using the simplemode addon as specified in the docs.

***Bad News:*** Not sure if I can make any more progress on my own, I'm not quite sure how I would start using simple mode to get the highlights working. Maybe you would have more luck/context as you implemented it to begin with?

Docs: https://codemirror.net/mode/dockerfile/

shiftkey:
@say25 @niik I might have gotten this working - here's how it looks in action:





I also opened https://github.com/desktop/highlighter-tests/pull/30 to track the example Dockerfile for regression testing.

say25:
Well that’s a shame that’s all I was missing :(

shiftkey:
@say25 all good, it required just a bit more knowledge about how we're setting up Codemirror within the app

shiftkey:
> @shiftkey any way we can get this reviewed sometime?

I'll check in with the team next week and see if we can move this along - we have a general rule of thumb that if someone has pushed commits to a branch they shouldn't be the primary reviewer, so this isn't something that I should be merging in myself.

app/src/highlighter/index.ts
```diff
@@ -26,85 +26,86 @@ interface IModeDefinition {
...
    */
-  readonly extensions: {
+  readonly mappings: {
     readonly [key: string]: string
   }
 }
 
 /**
- * Array describing all currently supported modes and the file extensions
+ * Array describing all currently supported extentionModes and the file extensions
```
niik:
```suggestion
 * Array describing all currently supported extensionModes and the file extensions
```

app/src/highlighter/index.ts
```diff
@@ -26,85 +26,86 @@ interface IModeDefinition {
...
  * that they cover.
  */
-const modes: ReadonlyArray<IModeDefinition> = [
+const extentionModes: ReadonlyArray<IModeDefinition> = [
```
niik:
```suggestion
const extensionModes: ReadonlyArray<IModeDefinition> = [
```

app/src/highlighter/index.ts
```diff
@@ -204,55 +205,82 @@ const modes: ReadonlyArray<IModeDefinition> = [
...
  * documentation for the IModeDefinition interface
  * for more information
  */
 const mimeModeMap = new Map<string, IModeDefinition>()
 
-for (const mode of modes) {
-  for (const [extension, mimeType] of Object.entries(mode.extensions)) {
-    extensionMIMEMap.set(extension, mimeType)
-    mimeModeMap.set(mimeType, mode)
+for (const extensionMode of extentionModes) {
```
niik:
```suggestion
for (const extensionMode of extensionModes) {
```

app/src/ui/diff/code-mirror-host.tsx
```diff
@@ -7,6 +7,9 @@ import 'codemirror/addon/selection/mark-selection'
 // Autocompletion plugin
 import 'codemirror/addon/hint/show-hint'
 
+// Required for some syntax highlighting support like Docker
+import 'codemirror/addon/mode/simple'
```
niik: @shiftkey I don't see why this would be needed in the renderer. We use a custom-built mode for our syntax highlighting so we really shouldn't need to explicitly import anything. The dockerfile mode automatically imports the simplemode *inside of the highlighter web worker* when necessary, as do many other modes.

Could you provide some more context on this change?

shiftkey: From memory I thought this was related to some styling issues, but let me pull it out and confirm it's not needed.

shiftkey: Yeah, I retested the Dockerfile styles with dark and light theme and couldn't spot anything weird on 8ddf521705672e46367fbbc690d515ebd3119fbf

niik:
Better late than never, thank you so much for this @say25!

Issue #4019, opened 2018-02-15
Double tapping a file should open with the default program


danielgalasko:
### Description

<img width="473" alt="screen shot 2018-02-15 at 11 25 31 am" src="https://user-images.githubusercontent.com/4479146/36249329-f84fd92c-1242-11e8-88a9-57445942a002.png">

When I am viewing files, I have to right click and select "Open with default program" to open a file. It would be awesome if this could be done by double clicking.

### Version

**GitHub Desktop version:** [1.0.13]

**OS version:** [10.12.6]

### Steps to Reproduce

1. Make changes to a source file
2. Observe the file in the Files Changed view
3. Double tap any file in the list

**Expected behavior:** [The file opens in my default program]

**Actual behavior:** [Nothing happens]

**Reproduces how often:** [All the time]

shiftkey:
@danielgalasko thanks for the feedback. We're tracking this in #2620, so please follow along with that.

Issue #3276, opened 2017-11-08
Double clicking file should Open in Default Program


digitalmaster:
### Description

Just figured this is something that others might like to have as well. Right now to open a file I always just right click and select "Open with default program" -- nothing happens when I double click so would be nice to make that the default double click behavior 😬

### Version

**GitHub Desktop version:** 1.0.8

**OS version:** 10.13 (17A405)

### Steps to Reproduce

1. Make a change
1. Open Github Desktop
1. Double Click a changed file

**Expected behavior:** File opens with Default Program.

**Actual behavior:** Nothing happens

**Reproduces how often:** 100%

### Logs

<!--

There may be some relevant information in log files generated by GitHub
Desktop:

 - If you are on macOS, attach the most recent log file from:
   `~/Library/Application Support/GitHub Desktop/logs/*.desktop.production.log`
 - If you are on Windows, attach the most recent log file from:
   `%APPDATA%\GitHub Desktop\logs\*.desktop.production.log`

The log files are organized by date, so see if anything was generated for
today's date.

Issue #1542, opened 2017-05-17
Could not open editor(e.g. Xcode) when double click filename table row in "Changes" pane.


r-plus:
### Description

Could not open editor(e.g. Xcode) when double click filename table row in "Changes" pane.

### Version

**GitHub Desktop version:** [version 0.5.3]
**OS version:** [version 10.12.4]

### Steps to Reproduce

1. Double click filename in "Changes" pane.

**Expected behaviour:**

Open double clicked file in editor that feature is implemented in GitHub Desktop ver.222.

**Actual behaviour:**

No effect.

**Reproduces how often:**

Always.

stevemoser:
I would add to this right click to open the file in an external editor or reveal in Finder, etc.

joshaber:
We made the conscious decision not to implement this, but we'd welcome an issue around @stevemoser's suggestion of right-clicking.

stevemoser:
I created #1566 for the right click functionality.

r-plus:
Okey, thanks all guys.

koenpunt:
> We made the conscious decision not to implement this

@joshaber where was that decision based on? And what about a keyboard shortcut then?

joshaber:
@koenpunt Unfortunately that decision was made in our closed team chat 😞 I think a keyboard shortcut would be reasonable.

koenpunt:
@joshaber I see, but that doesn't give much clarity. Just trying to understand why this feature, that was present in the previous version, has no change of getting in the new version

joshaber:
Yeah, sorry @koenpunt, I should have provided some more context there 😅 

If I recall, and it's been a while, our thinking was mainly that it was a difference in behavior between Mac and Windows which wasn't a high enough priority to special case for just Mac.

That said, if we get a lot of feedback around this we can always revisit it.

PR #4541, opened 2018-04-26
Feature/open files from sidebar

### LINKED ISSUE #4539: Double clicking a file to open in default editor
## Description

The previous version of github desktop allowed for double clicking a file in the sidebar and then it would open in the default editor. The new one lacks this feature. I added this to my fork https://github.com/HeroProtagonist/desktop/tree/feature/open-files-sidebar. I have tested and it works on Mac Sierra and Windows 10. If this feature would be wanted in the project I would be happy to open a pull request. 

## Version

* GitHub Desktop: 1.1.1 (But all since the logo changed)

* Operating system: All

### Expected Behavior

Double click file name, open it in default editor

### Actual Behavior

Nothing

## Additional Information

I did not see this mentioned in other issues. Apologies if it has been addressed elsewhere.

HeroProtagonist:
Resolves #4539

This was my solution to the issue I opened. Let me know if this makes sense. Thanks!

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -368,6 +368,7 @@ export class ChangesList extends React.Component<
           invalidationProps={this.props.workingDirectory}
           onRowClick={this.props.onRowClick}
           onOpenInExternalEditor={this.props.onOpenInExternalEditor}
+          repository={this.props.repository}
         />
 
         <CommitMessage
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -367,6 +367,7 @@ export class ChangesList extends React.Component<
           onSelectionChanged={this.props.onFileSelectionChanged}
           invalidationProps={this.props.workingDirectory}
           onRowClick={this.props.onRowClick}
+          onOpenInExternalEditor={this.props.onOpenInExternalEditor}
         />
 
         <CommitMessage
--- a/app/src/ui/lib/focus-container.tsx
+++ b/app/src/ui/lib/focus-container.tsx
@@ -8,6 +8,13 @@ interface IFocusContainerProps {
 
   /** Callback used when focus is within container */
   readonly onFocusWithinChanged?: (focusWithin: boolean) => void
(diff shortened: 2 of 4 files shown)


niik:
Hi @HeroProtagonist, thank you so much for wanting to contribute to Desktop!

This feature has been suggested in the past https://github.com/desktop/desktop/issues/1542, https://github.com/desktop/desktop/issues/2620, https://github.com/desktop/desktop/issues/3276 https://github.com/desktop/desktop/issues/4019. Not including that feature in the electron version of Desktop was a conscious decision, see https://github.com/desktop/desktop/issues/2620#issuecomment-328665777

> We made the conscious decision not to add it back in the rewrite... If I recall, and it's been a while, our thinking was mainly that it was a difference in behavior between Mac and Windows which wasn't a high enough priority to special case for just Mac.

We opted instead for the context menu route. I'm personally not sure how I feel about supporting double click but I'm definitely open to having a discussion about it. I think that needs to happen before we start looking at implementation though.

That said, I have a few small pieces of feedback for the change proposed here. The `List` component is a generic component for rendering vertical, scrollable, and virtualized lists. We use it to render the list of repositories, changed files, commits, branches, and even the menu items on Windows. With that in mind it should be agnostic with regard to what information it's presenting. Passing an (optional) repository and path prop to the `List` and `ListRow` components isn't compatible with that approach. Instead I would recommend following the approach we've taken for normal clicks on a row (onRowClick). I.e. the List should have an (optional) `onRowDoubleClick` callback prop that's unaware of what type of item has been double clicked. The `ChangesList` should then be the component responsible for passing a handler to the list which is in turn responsible for opening the external editor.

Let me know if that explanation didn't make sense.

I'm going to close this PR now until we have consensus on whether or not to move forward with double clicking opening files in external editors but don't worry, we can easily reopen it again. Let's move the discussion about whether or not to support double click back into the original issue (it seems we've pointed all previous issues about this to https://github.com/desktop/desktop/issues/2620 so I'll close your issue and point it to that one instead. Don't mind the fact that it's closed, that just means we're not currently planning on working on it, that too can be reopened.

Issue #3385, opened 2017-11-20
Beta Release Notes Don't Point to the Right Place


say25:
### Beta Release Notes Don't Point to the Right Place

what's new / release notes dont work with beta versions.

### Version
**GitHub Desktop version:** [1.0.10-beta0]

### Steps to Reproduce

1. Open app.
1. See new update.
1. Click `what's new` link or `release notes` in About screen.

**Expected behavior:** See release notes for beta version.

**Actual behavior:** See release notes for last release version.

**Reproduces how often:** `100%`

#### Additional Information

More than will to help solve this issue. 

I know that #2774 might solve this differently. 

I think https://desktop.github.com/release-notes/beta would be cool too. 

![image](https://user-images.githubusercontent.com/7467062/33046461-a5a4c324-ce16-11e7-8f57-b662c94f1bea.png)

shiftkey:
> I know that #2774 might solve this differently.

Kind of. Off the top of my head the things not quite covered by that issue:

 - we serve up the release notes from an API endpoint - we need to teach that endpoint about channels, and also get it to serve up the right content. That might be covered by #2774 implicitly, but its still work that needs to be done alongside.
 - `desktop.github.com` doesn't have a corresponding page to serve up the beta release notes

msftrncs:
The In-App 'what's new' are nice, but since you only see the most recent beta release, so if you missed one, you have to go looking around in the changelog for them.

PR #4545, opened 2018-04-27
Beta release of RBB lite and multi-select lists

niik:
After discussing with @nerdneha earlier today we decided that shipping RBB lite and multi-select to beta was the easiest way to get mileage on the changes as well as deliver updates to QA.

There was some concerns earlier about potential stability issues but our (my) take on this is that we'll accept some risk on the beta channel, that's what it's there for. Our need for mileage is high enough that it's important for us to get this to the beta channel as soon as possible.

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -1,9 +1,14 @@
 {
   "releases": {
     "1.1.2-beta4": [
-      "[Fixed] Add support for multiple selection in changes list - #1712. Thanks @icosamuel!",
+      "[New] New Compare tab allowing visualization of the relationship between branches",
+      "[New] Support for selecting more than one file in the changes list - #1712. Thanks @icosamuel!",
       "[Fixed] 'Select All' shortcut now works for changes list - #3821",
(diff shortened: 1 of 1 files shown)


nerdneha:
Looks like lines 7 and 8 are the same; I’m not in front of my computer right now but I can push and update if you’re clocked out for the week

niik:
Cheers, that’d be great. I’m not too concerned since I don’t believe
there’s a way to view the beta release notes anyway

shiftkey:
> I’m not too concerned since I don’t believe
there’s a way to view the beta release notes anyway

Yep, this is being tracked in #3385

nerdneha:
I think we didn't merge it just in case we wanted to rollback easily and just release master to beta. Things seems to be good now though, happy to merge

shiftkey:
@nerdneha it's be easy enough to revert #4522 and push a new beta update if we wanted to rollback the RBB changes for whatever reason.

Issue #4548, opened 2018-04-27
Purpose of the debug script and differences with the start script


streeter:
## Description
I was trying to run the local, debug version of GitHub Desktop, and found the `yarn debug` command. At first it doesn't work, and I noticed that it needs a bit of fixing (which has been done in https://github.com/desktop/desktop/pull/3675/files#diff-165034287d03bd2dd5ecb9b849bcad3d). However, I'm not sure what it does. If I want to run the Desktop locally, can I do that with other commands? Can I do `yarn build:dev && yarn start`? What does `yarn debug` do that the other ones don't?

## Version
* GitHub Desktop: `master`
* Operating system: `Mac OS 10.13.4`

## Steps to Reproduce

1. `yarn install`
2. `yarn debug` (which fails)
3. `yarn build:dev`
4. `yarn start`

### Expected Behavior
Not sure. I don't know what `script/debug` does that the `script/start` script does not.

### Actual Behavior
Again, the `yarn debug` issue in the script seems to be have fixed in an upcoming PR. But I'm afraid that using `yarn start` does not give the dev middleware, or the hot middleware.

iAmWillShepherd:
We have a [guide](https://github.com/desktop/desktop/blob/master/docs/contributing/setup.md) on building Desktop. Please check there for details. 

I'm going to close this issue since there doesn't seem to be a problem with the application.

streeter:
Ok, thanks @iAmWillShepherd. I'd seen that doc, but it doesn't mention the `debug` script. So I wasn't sure if that script was even needed, or if it provided additional debug functionality. I noticed an issue on the doc, so I'll make a PR for that.

iAmWillShepherd:
@streeter, we've also updated the docs to remove the `yarn debug` stuff that was giving you problems. Thanks for opening this issue and submitting your PR.

streeter:
Awesome! Thanks!

PR #4562, opened 2018-04-30
Cleanup debug script

shiftkey:
Relates to #4548

I wanted to remember what we were using `script/debug` for, but at first glance it appears the script no longer works:

```
~/src/desktop/ yarn debug
yarn run v1.6.0
$ cross-env NODE_ENV=development node script/debug
/Users/shiftkey/src/desktop/script/debug:14
    publicPath: config.output.publicPath,
                              ^

TypeError: Cannot read property 'publicPath' of undefined
    at Object.<anonymous> (/Users/shiftkey/src/desktop/script/debug:14:31)
    at Module._compile (module.js:662:30)
    at Object.Module._extensions..js (module.js:673:10)
    at Module.load (module.js:575:32)
    at tryModuleLoad (module.js:515:12)
    at Function.Module._load (module.js:507:3)
    at Function.Module.runMain (module.js:703:10)
    at startup (bootstrap_node.js:193:16)
    at bootstrap_node.js:660:3
```

This seems to be the necessary fix to get the script working, based on what `script/start` does:

```diff
diff --git a/script/debug b/script/debug
index ad4855567..4085107af 100644
--- a/script/debug
+++ b/script/debug
@@ -4,14 +4,16 @@
 
 const express = require('express')
 const webpack = require('webpack')
-const config = require('../app/webpack.development')
+const configs = require('../app/webpack.development')
 const server = express()
-const compiler = webpack(config)
+const compiler = webpack(configs)
 const port = process.env.PORT || 3000
 
+const [mainConfig, rendererConfig, askPassConfig] = configs
+
 server.use(
   require('webpack-dev-middleware')(compiler, {
-    publicPath: config.output.publicPath,
+    publicPath: rendererConfig.output.publicPath,
     noInfo: true,
   })
 )
```

This script was added initially so that you could debug the app from VSCode, which we added in #914 but then removed in #2457 because it didn't seem to work right, so I think this should be cleaned up until we can revisit it and get it right.

cc @iAmWillShepherd as you were involved with adding this


iAmWillShepherd:
Going to go ahead with a merge since this isn't user facing and removes what is essentially dead scripts.

PR #4563, opened 2018-04-30
Disable textbox if there are no branches to compare

### LINKED ISSUE #4559: Show message when no branches to compare
## Description
A suggestion for an enhancement. When there are no branches other than master and going to compare and select a branch from the list the pane shows "Sorry, I can't find that branch" when really there are just no other branches to select.

## Version
* GitHub Desktop: 1.1.2-beta4
* Operating system: Mac and Windows

## Steps to Reproduce
1. Use a repository that only has master branch
2. Go to `Compare` tab
3. Click in `Select branch to compare...`

### Expected Behavior
See a message like "There are no branches to compare"

### Actual Behavior
The current message says: "Sorry, I can't find that branch"

no other branches exist

iAmWillShepherd:
Fixes #4559.

Instead of showing a message saying there is nothing to compare, this PR goes one step further by not allowing the textbox to be selected when there are fewer than 2 branches in the repo.

**Before**
See original issue...

**After**

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -117,7 +117,6 @@ export class CompareSidebar extends React.Component<
 
   public render() {
     const formState = this.props.compareState.formState
-    const numBranches = this.props.compareState.allBranches.length
     const placeholderText =
       numBranches <= 1
         ? __DARWIN__
@@ -128,6 +127,7 @@ export class CompareSidebar extends React.Component<
             ? 'Select Branch to Compare...'
             : 'Select branch to compare...'
           : undefined
+    const { allBranches } = this.props.compareState
 
     return (
       <div id="compare-view">
@@ -138,7 +138,7 @@ export class CompareSidebar extends React.Component<
             placeholder={placeholderText}
             onFocus={this.onTextBoxFocused}
             value={this.state.filterText}
-            disabled={numBranches <= 1}
+            disabled={allBranches.length <= 1}
             onRef={this.onTextBoxRef}
             onValueChanged={this.onBranchFilterTextChanged}
             onKeyDown={this.onBranchFilterKeyDown}
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -117,12 +117,17 @@ export class CompareSidebar extends React.Component<
 
   public render() {
     const formState = this.props.compareState.formState
+    const numBranches = this.props.compareState.allBranches.length
     const placeholderText =
-      formState.kind === ComparisonView.None
+      numBranches <= 1
         ? __DARWIN__
-          ? 'Select Branch to Compare...'
-          : 'Select branch to compare...'
-        : undefined
+          ? 'There Are No Branches to Compare'
+          : 'There are no branches to compare'
+        : formState.kind === ComparisonView.None
+          ? __DARWIN__
+            ? 'Select Branch to Compare...'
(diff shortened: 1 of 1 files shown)


app/src/ui/history/compare.tsx
```diff
@@ -117,12 +117,17 @@ export class CompareSidebar extends React.Component<
...
     const formState = this.props.compareState.formState
+    const numBranches = this.props.compareState.allBranches.length
     const placeholderText =
-      formState.kind === ComparisonView.None
+      numBranches <= 1
         ? __DARWIN__
-          ? 'Select Branch to Compare...'
-          : 'Select branch to compare...'
-        : undefined
+          ? 'There Are No Branches to Compare'
```
iAmWillShepherd: Sorry about the nested ternary. If there were a function that could handle casing for me, I could remove the nesting.

iAmWillShepherd:
@meaghanlewis how do you feel about shortening the placeholder message to say _No Branches to Compare_?

meaghanlewis:
This LGTM, thanks @iAmWillShepherd ✨

j-f1:
How about “Nothing to Compare With”? Also, can you make the area behind the Octicon gray instead of white?

donokuda:
> How about “Nothing to Compare With”? Also, can you make the area behind the Octicon gray instead of white?

Good catch 👀 I didn't even notice that in the screenshot!

app/src/ui/history/compare.tsx
```diff
@@ -117,12 +117,17 @@ export class CompareSidebar extends React.Component<
 
   public render() {
     const formState = this.props.compareState.formState
+    const numBranches = this.props.compareState.allBranches.length
```
shiftkey: :art: could we just use `allBranches` here instead of `numBranches`?

app/src/ui/history/compare.tsx
shiftkey: I think `Are` should be lower-cased here. But I'm also :thumbsup: to `No Branches to Compare` here which would make that change unnecessary.

shiftkey: If you're worried about the nesting, why not extract to a function?

```ts
function getPlaceholderText(compareState: ICompareState): string | undefined {
  const formState = compareState.formState
  if (compareState.allBranches.length <= 1) {
    return __DARWIN__
      ? 'There Are No Branches to Compare'
      : 'There are no branches to compare'
  } else if (formState.kind === ComparisonView.None) {
    return __DARWIN__
      ? 'Select Branch to Compare...'
      : 'Select branch to compare...'
  } else {
    return undefined
  }
}
```

japan: Most style guides ask for all verbs to be capitalised in titles and headings. _"Capitalize My Title"_ is a great tool when in doubt - https://capitalizemytitle.com

shiftkey: @japan thanks for the tip - that will definitely help for the future!

iAmWillShepherd:
@j-f1 @donokuda , I've opened a separate issue to track the textbox color.

tierninho:
Found an edge case that makes this reproducible. 

If no branches exist, and user selects the `Compare to Branch` option from the Branch menu, then we still see:

screen shot 2018-09-21 at 9 46 56 am

Perhaps the fix can be folded in to https://github.com/desktop/desktop/issues/5603 ?

1.4.1-beta0, mac

PR #4583, opened 2018-05-02
 Fix background color behind octicon when textfield is disabled

### LINKED ISSUE #4580: Octicon color does not match when textbox is disabled
This issue was introduced as a result of PR #4563 and was pointed out by @j-f1.

> Also, can you make the area behind the Octicon gray instead of white?

Notice that the Octicon background next to the textbox is white, but the background of the textbox is gray.

donokuda:
Fixes #4580 

**Before**

screen shot 2018-05-02 at 3 19 13 pm

**After**

screen shot 2018-05-02 at 3 16 10 pm


shiftkey:
As this is a minor change I'd like to get into the next beta, I'm going to merge this in now rather than wait for the 24hr window.

PR #4586, opened 2018-05-03
actually get the styling fix for the disabled compare branch selector in

shiftkey:
Looks like #4583 was targeting a non-`master` branch, and merging it in  _after_ #4563 means that it hasn't actually made it to `master`.


shiftkey:
#jolo

Issue #3119, opened 2017-10-22
reset test fails when account name contains a #


shiftkey:
Repro:

 - Create a Windows account with a # in the name
 - Clone down the `desktop/desktop` repository
 - `yarn test:unit`

This error seems to be related to Git for Window's experimental support for [`reset --stdin`](https://github.com/desktop/desktop/blob/master/app/src/lib/git/reset.ts#L79-L90) with this specific account - I need to put together a patch for that and get it reviewed upstream.

```
  1) git/reset resetPaths resets discarded staged file:
     GitError: fatal: C:\Users\THEKON~1\AppData\Local\Temp\desktop-git-test-117920-3376-4eobcj.x96dy\README.md: 'C:\Users\THEKON~1\AppData\Local\Temp\desktop-git-test-117920-3376-4eobcj.x96dy\README.md' is outside repository
```

This needs to be fixed in Git for Windows, but for now I'm opening it up here to track it in case other users encounter it.

PR #4599, opened 2018-05-04
skip this test until we can address the underlying Git issue

shiftkey:
We're already tracking this in #3119, and the VSTS build agent is using an account with special characters in it which triggers the issue.


shiftkey:
#jolo

PR #4620, opened 2018-05-07
ensure FocusContainer also has key set

shiftkey:
🌵 **This fix depends on #4618 when running in development mode. Go review that first.** 🌵

This is a warning that React Dev Tools has picked up on Windows when you run up `master`:

screen shot 2018-05-07 at 1 01 29 pm

The `FocusContainer` was added in 2a59ef5a70f1c717a54e4182c0cd7f3b9efd44b1 while we were adding multi-select support in #4188, but there's a difference between macOS and Windows here - we render a "fake" scroll inside the list:

https://github.com/desktop/desktop/blob/617abad4611bc5b3700f2d3bb7d581771b1bc703/app/src/ui/lib/list/list.tsx#L691-L697

Because we can be returning a list of elements from this method, React expects each of them to be named with a distinct `key` attribute so they can be identified. There's more details about this in [the docs](http://fb.me/react-warning-keys).

### DIFF
--- a/app/src/ui/lib/list/list.tsx
+++ b/app/src/ui/lib/list/list.tsx
@@ -728,6 +728,7 @@ export class List extends React.Component<IListProps, IListState> {
       <FocusContainer
         className="list-focus-container"
         onKeyDown={this.onKeyDown}
+        key="focus-container"
       >
         <Grid
           aria-label={''}


j-f1:
You could also use a Fragment here (`<>{this.renderGrid(width, height)}{this.renderFakeScroll(height)}</>`), which avoids the need for a key.

shiftkey:
@j-f1 thanks for the tip! I'll open up a `tech-debt` task to revisit this - it's a critical area, and I want to ensure there's no unexpected side-effects.

PR #4622, opened 2018-05-07
[WIP] improve SSH diagnostics and troubleshooting

shiftkey:
It's Hack Week™ and I'm picking up something that I know that's been suboptimal in Desktop - working with SSH remotes. I started thinking about this a few weeks ago when I was jetlagged and wide awake at 3am, scribbling notes down so I could braindump and get back to sleep.

I'm timeboxing this at a week because we've got the 1.2 ship to get out the door next week, but here are some of my goals for the week:

 - improve troubleshooting SSH issues - guide the user through common issues, provide good diagnostics as a fallback
 - help the user setup their SSH environment
 - stretch goal: get Desktop to launch and manage the SSH environment automagically

### What happens currently?

For reference, this is what you currently see when you encounter an SSH-specific issue:

It's very generic, and there's lots of moving parts with SSH that need to be just right to have it work.

### What's changing?

I wanted to keep the scope narrow, so this workflow starts when you encounter an SSH authentication issue. Rather than the above screenshot, I want to have Desktop support troubleshooting and guide the user to a good place.

Here's a quick demo from earlier today:

These are the corner cases that I'm focusing on:

 - make it easy to get a diagnostics log of the SSH issue
 - handle host verification
 - create a new SSH key
 - upload SSH key to GitHub
 - add to running ssh-agent

There's probably a lot of opportunities for polish here, but for now I'm going to focus on the fundamentals and getting a robust solution together that tackles common pain points.

For the technical details about this, please read the docs here: [`docs/technical/ssh-troubleshooting.md`](https://github.com/desktop/desktop/blob/experimental-ssh-setup/docs/technical/ssh-troubleshooting.md). I'll be keeping that up to date throughout the week so it can be referred to later.

There's also a lot of platform-specific things with SSH that I need to keep in mind as I jump between the three platforms - those should be written up somewhere.

### Blockers

These things need to be resolved before the feature can be merged, in addition to whatever comes up with the review process:

 - the OAuth scopes associated with Desktop do not current support uploading SSH public keys - we need to figure out how to re-auth an existing token to grant these new scopes
 - ???


shiftkey:
Adding in an updated GIF of the flow (the last dialog needs a lot of polish):



I can create the new SSH key, but completing this flow requires uploading the SSH key through the API, which requires figuring out how to upgrade a token as I called out earlier.

shiftkey:
Closing in favour of #4660 and addressing the other blockers.

Issue #3903, opened 2018-01-31
Databases Schema Change


iAmWillShepherd:
I've been putting in work trying to understand the underlying data-model the app uses.

So far, I've produced two diagrams of our application's data model.

**Legend**
* `Database` **green**
* `GitHubUserDatabase` **blue**
* `IssuesDatabase` **orange**
* `PullRequestDatabase` **red**
* ~`StatsDatabase`~

_Note that I did not include the `StatsDatabase` in the ER Diagram._

## Current
![github desktop](https://user-images.githubusercontent.com/1715082/35654545-46581d3c-06b3-11e8-8fb7-56321f43b785.png)

## Proposed [WIP]
![github desktop proposed](https://user-images.githubusercontent.com/1715082/35654553-4ada0a3c-06b3-11e8-96a6-06cb31d99ab5.png)

iAmWillShepherd:
Posting questions I asked in slack here for reference.

> 1. Why are _mentionables_ associated with a `repository` and not a `github_repository`?
> 2. Do we want to maintain referential integrity?
> 3. Should a pull request be associated to a `repository` or `github_repository`?
> 4. Is `github_repository.parent_id` a foreign key? If so, what table does it refer to?

iAmWillShepherd:
Posting response from @shiftkey 

> initial thoughts without diving into the code:
> 1. mentionables are currently only used for autocomplete, and that’s only needed when a repository is local?
> 2. probably, unless we’re fighting the underlying database to achieve it. also, keep in mind that the v3 API now supports returning GraphQL identifiers to assist with migration: https://developer.github.com/changes/2017-12-19-graphql-node-id/
> 3. `githubRepo` makes sense from a data model perspective, especially if we start to do more API-backed features.
> 4. maybe that’s the way to represent a fork from the API?

shiftkey:
Tentatively marking this as `tech-debt` so it doesn't get mixed up with other tasks.

iAmWillShepherd:
I'm actually going to merge this into #2954 since this change is driven by the need to clean up the db.

Issue #2954, opened 2017-10-03
Make sure you remove repos from database


iAmWillShepherd:
While reviewing PR #2918, I noticed that there are records in the `gitHubRepositoryDatabase` that should no longer be there.

iAmWillShepherd:
For more context, I believe the `upgrade` function on the database needs to do this cleanup. Will research indexDB to see if it offers referential integrity.

Issue #1144, opened 2017-04-06
GitHub repository associations are never cleared


joshaber:
Once a `Repository` is associated with a `GitHubRepository`, that association is never cleared. But there are certainly cases where it should be:

1. GitHub repository is deleted.
2. User loses access to GitHub repository.
3. Etc.

PR #4630, opened 2018-05-07
No more dexie

iAmWillShepherd:
This PR spike is somewhat related to PR #3866 that I gave up on due to the difficulty of modifying our _schema_. My goal for this spike is to come up with a way to migrate away from [Dexie.js](http://dexie.org/) and [IndexedDB](https://w3c.github.io/IndexedDB/) to [LokiDb](https://w3c.github.io/IndexedDB/).

xref:  #3903, #2954 and #1144 

**Why the need to switch?**
Dexie.js provides a higher-level API to IndexedDB that removes _some_ the details of IndexedDB from our code. IndexedDB is great for what it does, but working with it can be a pain. 

* Unable to inspect the database outside of Chrome dev tools or without some other specialized tool
* Dexie.js does not allow native JS promises because they are not supported on all browsers
* Upgrade scripts for Dexie.js aren’t as powerful as I expected, e.g. it’s not possible to **easily** rename properties. If your mind works in SQL, then imagine being unable to easily rename a column in a table
* Versioning multiple data stores is hard

**What is our current data model?**
35654545-46581d3c-06b3-11e8-8fb7-56321f43b785

Note that our data model is relational.

**Why does ☝️  matter?**
Since IndexedDB is a noSQL database, it does not have the concept of relationships baked in. This means its up to the developer to maintain those relationships and, more importantly, we have to ensure referential integrity. That is both hard and not an effective use of time.

**Why not switch to a relational database?**
I know of no relational database built in JS that run in node.js. All the relational database I know are meant to be connected to from some client. Those databases do not fit out needs.

**What’s the goal of this PR?**
I do not intend to do a full migration in this PR. As stated earlier, this is more of a spike/POC. In this spike, I will try to accomplish the following:

* Rework our a data model so that it fits the document-based paradigm
* Replace a dexie database in one `Store` with this new implementation
* Design and implement a migration strategy from Dexie to LokiDB
* Perform a database upgrade using LokiDB
* Figure out if renaming a property on a collection is easier than Dexie.js
* Define at least one view

**Updated data model**

I've reworked our data model a bit so that it works in a document-oriented database and since Desktop does not store a ton of data, duplicating data should not be a concern. The root object is the repository. This way, when we switch repos, we only need to load one document that includes all the data we'd need to do any work.

```ts
export interface IRepositoryModel {
  readonly kind: 'repository'  //key
  readonly name: string
  readonly path: string //key
  readonly isMissing: boolean
  readonly ghRepository?: IGHRepository
}

interface IIssueModel {
  readonly number: number  //key
  readonly title: string
  readonly updatedAt: Date
}

interface IUserModel {
  readonly name: string | null
  readonly login: string //key
  readonly email: string | null
  readonly endpoint: string //key
  readonly avatarUrl: string
}

interface IGHRepositoryModel {
  readonly kind: 'gh-repository'
  readonly name: string
  readonly defaultBranch: string
  readonly isPrivate: boolean
  readonly cloneUrl: string  //key
  readonly htmlUrl: string | null
  readonly issues: ReadonlyArray<IIssue>
  readonly owner: IUser
  readonly parent: IGHRepository | null
  readonly mentionables: ReadonlyArray<IUser>
  readonly pullRequests: ReadonlyArray<IPullRequest>
}
```

One benefit to this data model that I had not originally considered is that it maps 1:1 with the data models we use throughout the app. This is because we no longer need to store database IDs for expression relationships between entities.

**The _new_ plan**:
* Create files for each main piece of the data model: `Repository` and `GHRepository`, `User`, `Issue`, `PullRequest`, and `Mentionable`.
* Export two objects, `Command` and `Query`, from each file
* Re-export as named exports from `index` e.g., `PullRequestCommands`
* Use the commands/queries inside of each store (we no longer need to pass a database to the store)
* Build some mapper functions that can convert between the database model boundary and the store model boundary i.e, stores use `APIResult` (my new term for the interfaces that map data from our API to TS objects) while the database uses `Model` (the word _model_ is not part of the type name however). My initial goal was to replace the instance-based models e.g., `Repository`, but the tight coupling makes that difficult at the moment.

### Notes
Desktop does not store a ton of data, so duplicating data should not be a concern. I’m thinking we can represent the entire state of a repo in one document.

While going through this, I've noticed that our stores are tightly coupled to our database and the technology. This slowed me down significantly because I could not simple replace the database with another implementation which forced me to change my approach. Now, I am building the data layer so that it doesn't leak implementation details into each store. I'm accomplishing this by using a hacky variant of CQS.

With the new plan mentioned above, our migration strategy is greatly simplified because we can ship code that uses the commands to recreate the objects in our database file.


j-f1:
You could use SQLite, which is file-based and doesn’t require connecting to a server.

iAmWillShepherd:
Closing now that hack week is over. Will continue working on this in my off time.

PR #4585, opened 2018-05-03
introduce distinct issue templates

shiftkey:
[Improved issue templates](https://blog.github.com/2018-05-02-issue-template-improvements/) was just made available to everyone, and I wanted to take this opportunity to see if we can improve how we get feedback for Desktop by introducing two templates:

 - a bug report template - what we know and have been working with for ages
 - an enhancement template - a simpler, free-form template to gather ideas

I'm going to look around for some examples of good enhancement templates, and I'm open to ideas from others about good ones they've seen elsewhere.

 - [x] introduce a good enhancement template for Desktop
 - [x] remove old `ISSUE_TEMPLATE.md`

I might need to look at a subsequent PR to address the docs usage - it's not clear to me from the GitHub docs how to refer to a specific template in the URL. Let me know if you've spotted this already.

PR #4631, opened 2018-05-08
use more granular links when documenting to open a new issue

shiftkey:
Now that #4585 is merged, there are two important changes to the issue tracker:

 - the **New Issue** button now navigates to `/issues/new/choose` which displays the available templates :tada:
 - Some of our docs point to `/issues/new`, which is now the empty template. Oh well.

This PR updates all the places where we point to `/issues/new` to be more specific. I also tweaked some words while I was in there.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -108,7 +108,7 @@ information you would like to add.
 Enhancement suggestions are tracked as [GitHub issues](https://guides.github.com/features/issues/).
 
 Simply create an issue on the [GitHub Desktop issue tracker](https://github.com/desktop/desktop/issues/new?template=feature_request.md)
-and fill out the template.
+and fill out the provided issue template.
 
 Some additional advice:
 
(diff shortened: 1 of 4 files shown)


JordanMussi:
The in-app link from the menu should be updated as well 🙂
https://github.com/desktop/desktop/blob/b9a4e82b5108f9b8cd4e06935cbc4b33cd4ef21e/app/src/main-process/menu/build-default-menu.ts#L331

shiftkey:
@JordanMussi nice spot - added to my list!

JordanMussi:
As you said in the first post, `/issues/new` is blank with no template. So people currently following the link through from the app won’t be given a template so perhaps your [canned reply](https://github.com/desktop/desktop/issues/4633#issuecomment-387298817) should be more helpful as they may have not even been presented with the template.

j-f1:
Perhaps we should add a default issue template that directs people to the template chooser.

shiftkey:
> Perhaps we should add a default issue template that directs people to the template chooser.

We've already got issues with submitters ignoring the template or removing it, and a process to deal with those. I wouldn't aim to make this too convoluted.

iljasergijenko:
# This is an <h1> tag
## This is an <h2> tag
###### This is an <h6> tag

PR #4640, opened 2018-05-09
keep the branch name readable in the compare tab

shiftkey:
Found this while poking at the compare tab:

By default our `.ref-component` uses `word-break: all` to ensure it doesn't overflow, but breaking in the middle of a word doesn't seem optimal when we have space.

**Before**

**After**


shiftkey:
I'd like to get this in for the 1.2.0 launch - is anyone able to take this over given @niik is on holidays this week?

Issue #4647, opened 2018-05-09
Move unpublished commits to new branch


nerdneha:
## Goal
As a user, before I accidentally made a bunch of commits on the wrong branch (ex: protected or master), I want the app to warn me and guide my commits to a new branch so that I don't have to use command line magic to get them to a new branch or reset my work and risk losing everything.

#### Background
Inspired by previous issue [Move unpublished commits on master to new branch](https://github.com/desktop/desktop/issues/1021) that's about to land in 1.2. 

Colloquially known as **New Branch, Maybe?**

## Solution
[to be filled out by @niik]

#### Alternatives Considered
[to be filled out by @niik]

#### Task List
- [ ] a
- [ ] b
- [ ] c
- [ ] incorporate analytics

#### Mockups
[@niik put mockups here]

#### User Testing Strategy
[@niik/@nerdneha describe when/what/how for user testing]

#### Related
- [ ] inform help-docs on user-facing updates
- [ ] update marketing page screenshots and add feature page
- [ ] update blog post for GitHub blog

billygriffin:
I'm going to close this since we have this work captured in https://github.com/desktop/desktop/issues/1021.

Issue #4646, opened 2018-05-09
[Feature NDDB] - Notification of Diverging From Default Branch


nerdneha:
## Goal 

As a user, when I am working on my own branch, I want to know when a relevant branch (ex: master) has new commits so that I can merge those changes in as soon as possible to avoid conflicts over time and so that I don't have to break my workflow to check if there are updates.

**Background**
Built upon the Compare Branch feature that's about to land in 1.2.

## Solution

This issue tracks all the work needed to be done around this feature. See the task list for a breakdown of what needs to be done.

The following is a high-level overview of the solution. Implementation-specific details will be placed in the appropriate sub-issue and/or pull request.

**History Tab**
* Desktop will try to infer what base branch should be used to compare with the current branch, so the user won't have to perform any setup to use the feature
* If the number of commits behind is different from the last time Desktop checked, a blue notification dot will be placed on the History tab.
* To provide additional context for the notification, a banner will be rendered between the History tab and the compare branch selector; The banner will include the info on the number of commits the current branch is behind and two call-to-action buttons: Merge and Compare
* If the Merge button is clicked, the Merge dialog will appear with the base branch preselected
* If the Compare button is clicked, users will be taken to the compare view
* The notification dot will be cleared if the user clicks on the History tab; the banner will be persisted until dismissed through one of the actions or being dismissed

**Changes Tab**
* ~To avoid confusion, the notification dot will be replaced with a badge that contains the number of changes to be committed~ (pulled from scope, see Related Issues section below)

**Alternatives Considered**

* @donokuda and I spoke about potentially adding a bit of context to the notification dot such as the number of commits the current branch is behind the base branch. We agreed we can iterate to that point if we decide to move in that direction.

**Risks**

| Risk | Mitigation |
|:--|:--|
|  We scrap the current design for the Compare Branch feature | Focus on making implementation UI/UX agnostic |

### Mockups

![image](https://user-images.githubusercontent.com/1174461/40444350-04f5ee64-5e7e-11e8-961c-58d7b7931c99.png)

A breakdown of the UI for some of the newer elements:

![image](https://user-images.githubusercontent.com/1174461/40444589-a233aab8-5e7e-11e8-9c09-258686840a9d.png)

<details>
<summary>Testing</summary>

## User Testing Strategy

We need to actors to test this feature:
* Desktop team member
* Mona (the user)

#### Setup
1. Create repository, `test`, on GitHub
1. Clone `test` using Desktop
[...]

PR #4649, opened 2018-05-09
Update Issue Template

nerdneha:
After using the issue template when describing two new features, I found myself restructuring the headers anyway and trying to create some hierarchy so it's easier to read. I also found myself wanting a tl;dr at the top (aka goal).

Here's what I used:

- [example 1 - Move unpublished commits to new branch](https://github.com/desktop/desktop/issues/4647)
- [example 2 - Notify when there are Compared Branch updates](https://github.com/desktop/desktop/issues/4646)

**Caveat**
It's possible that this should be a different template altogether and only one that we use when we're creating feature-issues for major features being shipped. I'm open to putting this somewhere else and not replace the existing template if it doesn't help with the general requested feature case.


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -3,15 +3,27 @@ name: 🎉 Feature request
...
 
-**Is your feature request related to a problem? Please describe.**
-A clear and concise description of what the problem is. Ex. I have an issue when [...]
+## Goal
+A clear and concise description of what the problem and request are. Ex. I have an issue when [...problem...]. I'd like [...feature...] so that [...reason...]
 
-**Describe the solution you'd like**
-A clear and concise description of what you want to happen. Add any considered drawbacks.
+#### Background
+Any related issues or any preceding features that this would depend on. (linked from desktop repository)
```
shiftkey: A couple of :art: suggestions:

 - `preceding features that this might depend on` to allow for some flexibility of interpretation
 - `Links to other Desktop discussions are appreciated.` to encourage cross-referencing

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -3,15 +3,27 @@ name: 🎉 Feature request
...
 
-**Teachability, Documentation, Adoption, Migration Strategy**
-If you can, explain how users will be able to use this. This will help the core team understand how the feture would work, and assist with future work like documentation.
-Screenshots or mockups here are a great help!
\ No newline at end of file
+## Teachability, Documentation, Adoption, Migration Strategy*
+This will help the core team understand how the feature would work, and assist with future work like documentation.
+Screenshots or mockups here are a great help! Here are some example subheaders:
+
+#### Task List
```
shiftkey: This feels like an implementation-specific thing - should we put this up-front or only add this after we've been able to come to an agreement on things?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -3,15 +3,27 @@ name: 🎉 Feature request
...
+- [ ] a
+- [ ] b
+- [ ] c
+- [ ] incorporate analytics
+
+#### Mockups
```
shiftkey: Personal feels: I'd move this up below the `Solution` section as this feels like they both belong together.

nerdneha: i'm open to considering; the only reason i had them at the end was that i feel like no one reads after the mockups since they can get so long and i wanted the text together. 
do you read much beyond? if so, maybe it's just a worry that isn't grounded in anything real!

.github/ISSUE_TEMPLATE/feature_request.md
nerdneha: 👍 you're right, no point in listing it unless the solution is 👍ed. added `(once solution is agreed upon)` after this line

.github/ISSUE_TEMPLATE/feature_request.md
shiftkey: > i'm open to considering; the only reason i had them at the end was that i feel like no one reads after the mockups since they can get so long and i wanted the text together.

Yeah, I can see the side of that. Alternatively, if mockups become significant they could be done in their own comment after, with details there. I guess I'd ask how much detail we need to  capture in the feature request template at the start? We might have a couple of mockups initially for something - those might help alongside the words. Someone might then have more detailed mocks - I guess those could be added to a follow-up comment with their own words to separate them better.

> do you read much beyond? 

Initially, probably not. But having the words alongside the text might help better for wrapping my head around a feature in terms of the flow - moving mockups away from the words about the feature that might be confusing?

.github/ISSUE_TEMPLATE/feature_request.md
shiftkey: If this template is the same one that everyone fills in, does this add value?

PR #4652, opened 2018-05-10
actually get tslint working in the CI builds

### LINKED ISSUE #2371: the linter has forsaken our event handlers
It'd be nice to see if we can get our linters to catch scenarios like #2368.

I thought this was already in place but maybe it regressed?

shiftkey:
Fixes #2371 once and for all.

Here's the problem in a nutshell:

```shellsession
$ git diff
diff --git a/app/src/ui/cli-installed/cli-installed.tsx b/app/src/ui/cli-installed/cli-installed.tsx
index 3ac74414d..a4fb81537 100644
--- a/app/src/ui/cli-installed/cli-installed.tsx
+++ b/app/src/ui/cli-installed/cli-installed.tsx
@@ -11,6 +11,7 @@ interface ICLIInstalledProps {
 
 /** Tell the user the CLI tool was successfully installed. */
 export class CLIInstalled extends React.Component<ICLIInstalledProps, {}> {
+  private onDismissed() {}
   public render() {
     return (
       <Dialog
@@ -20,7 +21,7 @@ export class CLIInstalled extends React.Component<ICLIInstalledProps, {}> {
             : 'Command line tool installed'
         }
-        onSubmit={this.props.onDismissed}
+        onSubmit={this.onDismissed}
       >
         <DialogContent>
           <div>

$ yarn tslint
yarn run v1.6.0
$ tslint ./{script,tslint-rules}/*.ts ./app/{src,typings,test}/**/*.{ts,tsx}
✨  Done in 1.36s.

$ ./node_modules/.bin/tslint ./{script,tslint-rules}/*.ts ./app/{src,typings,test}/**/*.{ts,tsx}
zsh: no matches found: ./app/typings/**/*.ts

$ ./node_modules/.bin/tslint ./{script,tslint-rules}/*.ts ./app/{src,test}/**/*.{ts,tsx}

ERROR: app/src/ui/cli-installed/cli-installed.tsx[24, 19]: A class method is passed as a JSX attribute without having the 'this' reference bound in the constructor: this.onDismissed
```

This actually is detected in the editor:

- [x] confirm CI doesn't catch this defect
- [x] use proper `tslint` command line interface
- [x] confirm this catches linting errors in `app` and `script`  directories, and doesn't just follow what's defined in `./tsconfig.json` (`tslint-rules` is ignored, but I don't want to dig into that and hold this up further).
- [x] fix introduced defect


shiftkey:
> confirm CI doesn't catch this defect

Turns out that macOS doesn't catch this, but Windows and Linux do. A-ma-zing.

shiftkey:
@j-f1 any clues into why this might be the right fix (or even if it is)? CircleCI seems to only incorrectly pass once, and AppVeyor has a green build in there before reverting the problem commit, so I'm not quite sure what to chalk this up to...

shiftkey:
Found a more reliable solution than moving this out to it's own script :tada:

shiftkey:
Alright, I think this is finally ready for review :apple:

Issue #4229, opened 2018-03-12
Add filtering of Pull Requests by pull request number and author


tierninho:
Currently, the Pull Request tab (Current Branch > Pull Requests) has fuzzy search functionality that allows you to search a string and filters the results by finding matches to the PR titles. This is great if you know the title you are searching for or approximate keywords but does not allow you to search by pull request number or by author handle.
For example, if you wanted to find PR #4227, `Bump toolbar contrast`, you could find it by searching "contrast" or "btcrst", etc. but you can't find it by searching "4227" or "niik"....yet 👇 

✏️ 🗒 **Task**: **Index the pull request numbers and authors and allow filtering of results by the PR number and author.**

Suggested Functionality (let's pretend the user is Paula) :
- **match all number occurrences**: if Paula types `2` in the filter (without `#` sign), they will see all the issues with a `2` in it, like: `#21`, `#2`, `#294`, `#321`, `#52`,  and a PR with the title `when 2 becomes 1` (new)
- **filter by `#` sign**: if Paula types `#23` (with `#` sign), they will see all issues *beginning* with `23`, like `#23` and `#232` but not `#123` (new)
- **match all author occurrences**: if Paula types `n` in the filter, they will see all issues with an `n` in it like a PR with the title `Add context menu to...` and all authors with an `n` in it like `nerdneha` and `niik` (new)
- **highlight matching characters**: while Paula types, text matching the filter will continue to be highlighted (existing functionality)
-  **preserve existing fuzzy searching**: when Paula types a string with mixed characters, they will see results that match existing fuzzy search results, like `3I` will show `Drop back to v3 API to retrieve...` Not sure of the expectation when we type "32" though... (existing functionality)

![screen shot 2018-03-12 at 11 20 23 am](https://user-images.githubusercontent.com/14828183/37302081-c07145aa-25e7-11e8-8e1c-9b1682a37099.png)

tierninho:
cc @desktop/design ⬆️

donokuda:
I'm 👍for this. I ran into a situation today where I wanted to search by the pull request number.

> Current fuzzy searching will search for characters separately, ex: "3I" will show `Drop back to v3 API to retrieve...` Not sure of the expectation when we type "32" though...

I'm fine with punting a decision on consecutive numbers until after a first pass at this and getting a feel of it using real information. Basically, let's see what we get by default and go from there.

nerdneha:
[...]

PR #4653, opened 2018-05-10
Add filtering of Pull Requests by pull request number and author

damaneice:
Add filtering of Pull Requests by pull request number and author

Related issue: #4229


damaneice:
I think the PR is ready for feedback.  @iAmWillShepherd @shiftkey @nerdneha

damaneice:
@iAmWillShepherd @shiftkey @nerdneha I think this PR is ready for feedback.

shiftkey:
@damaneice this is looking really neat! Some screenshots to show it in action:







I was also able to filter on the relative date - not sure if that's intended:

app/src/lib/fuzzy-find.ts
```diff
@@ -16,32 +16,34 @@ export interface IMatch<T> {
...
   // matching `query` against itself is a perfect match.
   const maxScore = score(query, query, 1)
   const result = items
     .map((item): IMatch<T> => {
-      const key: string =
-        typeof getKey === 'function'
-          ? (getKey as KeyFunction<T>)(item)
-          : String(item[getKey])
+      const matches: Array<ReadonlyArray<number>> = []
+      const itemTextArray = (getKey as KeyFunction<T>)(item)
```
shiftkey: :fire: No need to cast here if you've got the signature for `getKey` right

app/src/ui/branches/pull-request-list.tsx
```diff
@@ -206,8 +207,12 @@ export class PullRequestList extends React.Component<
 function createListItems(
   pullRequests: ReadonlyArray<PullRequest>
 ): IFilterListGroup<IPullRequestListItem> {
+  function getSubtitle(pr: PullRequest) {
```
shiftkey: :art: Does this need to be defined inline? I think it can live outside `createListItems`, not sure if I'm missing some context...

app/src/ui/branches/branch-list-item.tsx
```diff
@@ -24,17 +24,20 @@ export class BranchListItem extends React.Component<IBranchListItemProps, {}> {
     const lastCommitDate = this.props.lastCommitDate
     const isCurrentBranch = this.props.isCurrentBranch
     const name = this.props.name
+    const hasMatches = this.props.matches.length > 0
+    const matches = hasMatches ? this.props.matches[0] : []
```
shiftkey: We're doing this in a couple of places - could this be extracted to a function like `getFirstMatchesOrEmpty`?

```ts
function getFirstMatchesOrDefault(matches: ReadonlyArray<ReadonlyArray<number>>) {
  returns matches.length > 0 ? matches[0] : []
}

...

const matches = getFirstMatchesOrDefault(this.props.matches)
```

app/src/ui/history/compare-branch-list-item.tsx
```diff
@@ -41,11 +41,13 @@ export class CompareBranchListItem extends React.Component<
       </div>
     ) : null
 
+    const hasMatches = this.props.matches.length > 0
+    const matches = hasMatches ? this.props.matches[0] : []
```
shiftkey: This could be tidied up with a `getFirstMatchesOrDefault`

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -204,13 +204,14 @@ export class CloneGithubRepository extends React.Component<
 
   private renderItem = (
     item: IClonableRepositoryListItem,
-    matches: ReadonlyArray<number>
+    matches: ReadonlyArray<ReadonlyArray<number>>
   ) => {
+    const textMatches = matches.length > 0 ? matches[0] : []
```
shiftkey: This could be tidied up with a `getFirstMatchesOrDefault`

app/src/ui/branches/pull-request-list-item.tsx
```diff
@@ -51,7 +51,11 @@ export class PullRequestListItem extends React.Component<
   public render() {
     const title = this.props.loading === true ? undefined : this.props.title
     const subtitle = this.getSubtitle()
-
+    const matches = this.props.matches
+    const hasTitleMatches = matches.length > 0
+    const titleMatches = hasTitleMatches ? matches[0] : []
```
shiftkey: I'm fine with leaving this as-is rather than refactoring to use `getFirstMatchesOrDefault` because of the `subtitleMatches` bit below.

app/src/ui/branches/pull-request-list.tsx
damaneice: No we don't need to inline this. I will move it out.

app/src/ui/branches/branch-list-item.tsx
shiftkey: I think it's appropriate to keep this in `lib/fuzzy-find.ts` alongside the other logic for this.

app/src/ui/branches/pull-request-list-item.tsx
```diff
@@ -51,7 +51,11 @@ export class PullRequestListItem extends React.Component<
...
+    const hasSubtitleMatches = matches.length > 1
+    const subtitleMatches = hasSubtitleMatches ? matches[1] : []
```
shiftkey: I think at this point the only concern that I have with this change is that the `matches` result is now an array of arrays, and we need to carry around the implicit context like this.

It'd be nice to move this to a more declarative shape, like this:

```ts
interface IMatches {
  readonly title: ReadonlyArray<number>,
  readonly subtitle: ReadonlyArray<number>
}
```

But I think that means revisiting `IMatch<T>` because that's where we "carry around" these maches:

```ts
export interface IMatch<T> {
  /** `0 <= score <= 1` */
  score: number
  item: T
  matches: ReadonlyArray<ReadonlyArray<number>>
}
```

I won't hold up this PR for that, but if anyone has bright ideas I'd love to hear them!

damaneice: I think I like your suggestion. Let me try it out

app/src/lib/fuzzy-find.ts
```diff
@@ -12,36 +12,49 @@ function score(str: string, query: string, maxScore: number) {
...
 export interface IMatch<T> {
   /** `0 <= score <= 1` */
   score: number
   item: T
-  matches: ReadonlyArray<number>
+  matches: IMatches
 }
 
-export type KeyFunction<T> = (item: T) => string
+export function getFirstMatchesOrDefault(
```
shiftkey: This can be removed now we've changed up the shape of the matches result

app/src/lib/fuzzy-find.ts
```diff
@@ -12,36 +12,49 @@ function score(str: string, query: string, maxScore: number) {
   return fuzzAldrin.score(str, query, undefined, options) / maxScore
 }
 
+export interface IMatches {
+  readonly title: ReadonlyArray<number>,
```
shiftkey: Run a `yarn lint:fix` over this to get things up to date

shiftkey:
Thanks for this @damaneice! Works a treat!

I'm going to let it sit for a day or so and see if anyone from @desktop/core or @desktop/comrades wants to also give it a review.

app/src/lib/fuzzy-find.ts
```diff
@@ -12,36 +12,48 @@ function score(str: string, query: string, maxScore: number) {
   return fuzzAldrin.score(str, query, undefined, options) / maxScore
 }
 
+export interface IMatches {
+  readonly title: ReadonlyArray<number>
```
iAmWillShepherd: I think we should re-work this interface to 

```ts
interface IMatch {
  readonly title: number
  readonly subTitle: string
}
```
That way when we pass matches, we are passing an array of objects that don't require us to use an index to relate to pieces of data together.

For example, If we want to get the title and the subtitle of a match, we could do the following:

```ts
interface IMatch {
  readonly title: number
  readonly subTitle: string
}

function foo(matches: ReadonlyArray<IMatch>) {
  const match = matches[0]

  // match contains all the data we need
  console.log(`title: ${match.title}, subtitle: ${match.subtitle}`)
}
```
As opposed to:
``` ts
interface IMatches {
  readonly title: ReadonlyArray<number>
  readonly subTitle: ReadonlyArray<string>
}

function bar(matches: ReadonlyArray<IMatches>) {
  // we have to index into two separate arrays
  const title = matches.title[0]
  const subtitle = matches.subtitle[0]

  console.log(`title: ${title}, subtitle: ${subtitle}`)
}
```

shiftkey: Not sure I follow the proposed change here, as `fuzzAldrin.match` returns the `ReadonlyArray<number>` which represents the indexes of a string to highlight all the way down in `HighlightText` - are we going to change that representation to something else?

iAmWillShepherd: No we are not. We talked this over a bit earlier and I'm fine with the change.

iAmWillShepherd: For more context, I was under the impression that these two properties needed to be related somehow, but after going deeper into how this is used, I realized that there is no need to relate the two. @damaneice and I both agreed that naming is hard and that's what contributed to my confusion. He cleared it up for me though.

aateeque:
Is there a way to configure this for Bitbucket?

shiftkey:
@aateeque we don't have any sort of Bitbucket-specific integration currently.

aateeque:
@shiftkey are there any plans for other providers?

PR #4657, opened 2018-05-10
Add support for WSL shell on Windows 10

### LINKED ISSUE #4641: Add support for WSL Bash on Windows 10
Windows Subsystem for Linux has available `bash` environment that's available as an optional extra in recent versions of Windows 10.

screen shot 2018-05-09 at 2 08 05 pm

It would be nice to light this up as an option in Desktop and let users choose this shell.

The checks that Desktop need to perform should be added into [`app/src/lib/shells/win32.ts`](https://github.com/desktop/desktop/blob/master/app/src/lib/shells/win32.ts) and we have [documentation](https://github.com/desktop/desktop/blob/master/docs/technical/shell-integration.md) outlining how to add support for additional shells.

The high-level steps for looking this up are basically:

 - are you on Windows 10? If not, return `false`
 - Have you enabled WSL? This seems to be the magic registry key `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Component Based Servicing\Notifications\OptionalFeatures\Microsoft-Windows-Subsystem-Linux` and it contains these values:

If not found inthe registry, return `false`

 - Lastly, does `C:\Windows\System32\bash.exe` exist on disk. That seems to be enough for an entry point but it'd also be nice to be able to detect that you have a Linux distro installed (otherwise you would see a prompt to visit the store and install, which is fine). 

We need to ensure that it launches in the user's repository, and I think we can lean on `cwd` like this:

```
return spawn(pathToBashExe, [ ], {
  shell: true,
  cwd: path,
})
```

**EDIT:** @JordanMussi added some extra context from their setup below - feel free to use that alongside the above rules.

This should all just work, but I'm not sure on the right label for this shell to prevent it from being confused with Git Bash. In the start menu you search on the Linux distro you have, which means support for multiple distros is possible - maybe that's a stretch goal for this?

JordanMussi:
Fixes #4641

Currently I'm relying on `C:\Windows\System32` being on the path (reasonable assumption :wink:) because passing `C:\Windows\System32\wsl.exe` to `spawn` doesn't seem to actually launch anything (same for `bash.exe`).

### DIFF
--- a/app/src/lib/shells/win32.ts
+++ b/app/src/lib/shells/win32.ts
@@ -2,6 +2,7 @@ import { spawn, ChildProcess } from 'child_process'
 import * as Path from 'path'
 import { enumerateValues, HKEY, RegistryValueType } from 'registry-js'
 import { pathExists } from 'fs-extra'
+import * as OS from 'os'
 
 import { assertNever } from '../fatal-error'
 import { IFoundShell } from './found-shell'
@@ -12,6 +13,7 @@ export enum Shell {
   PowerShellCore = 'PowerShell Core',
   Hyper = 'Hyper',
   GitBash = 'Git Bash',
+  Wsl = 'WSL',
 }
 
 export const Default = Shell.Cmd
@@ -35,6 +37,10 @@ export function parse(label: string): Shell {
 
   if (label === Shell.GitBash) {
     return Shell.GitBash
+  }
+
+  if (label === Shell.Wsl) {
+    return Shell.Wsl
   }
 
   return Default
@@ -79,6 +85,14 @@ export async function getAvailableShells(): Promise<
     shells.push({
       shell: Shell.GitBash,
       path: gitBashPath,
+    })
+  }
+
+  const wslPath = await findWsl()
+  if (wslPath != null) {
+    shells.push({
+      shell: Shell.Wsl,
+      path: wslPath,
     })
   }
 
@@ -213,6 +227,42 @@ async function findGitBash(): Promise<string | null> {
   return null
 }
 
+async function findWsl(): Promise<string | null> {
+  // First, determine that we are running Windows 10 or greater
+  const winVersion = OS.release()
+  const versionParts = winVersion.split('.')
+
+  if (versionParts.length == 0 && /^[0-9]+$/.test(versionParts[0])) {
+    return null
+  }
+
+  const majorVersion = Number(versionParts[0])
+  if (majorVersion < 10) {
+    return null
+  }
+
+  // Second, check if WSL enabled?
+  const registryPath = enumerateValues(
+    HKEY.HKEY_LOCAL_MACHINE,
+    'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Component Based Servicing\\Notifications\\OptionalFeatures\\Microsoft-Windows-Subsystem-Linux'
+  )
+
+  // TODO - does this check need to be more specific?
+  if (registryPath.length === 0) {
+    return null
+  }
+
+  // Finally, does wsl.exe exit on the disk?
+  const path = Path.join('C:\\Windows\\System32', 'wsl.exe')
+  if (await pathExists(path)) {
+    return path
+  } else {
+    log.debug(`[WSL] registry entry found but does not exist at '${path}'`)
+  }
+
+  return null
+}
+
 export function launch(
   foundShell: IFoundShell<Shell>,
   path: string
@@ -246,6 +296,8 @@ export function launch(
         shell: true,
         cwd: path,
       })
+    case Shell.Wsl:
+      return spawn('START', ['wsl'], { shell: true, cwd: path })
     case Shell.Cmd:
       return spawn('START', ['cmd'], { shell: true, cwd: path })
     default:
--- a/app/src/lib/shells/win32.ts
+++ b/app/src/lib/shells/win32.ts
@@ -14,6 +14,7 @@ export enum Shell {
   Hyper = 'Hyper',
   GitBash = 'Git Bash',
   Wsl = 'WSL',
+  WslBash = 'WSL Bash',
 }
 
 export const Default = Shell.Cmd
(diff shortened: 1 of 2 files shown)


app/src/lib/shells/win32.ts
```diff
@@ -213,6 +240,64 @@ async function findGitBash(): Promise<string | null> {
...
+    return false
+  }
+
+  // Second, check if WSL enabled?
+  const registryPath = enumerateValues(
+    HKEY.HKEY_LOCAL_MACHINE,
+    'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Component Based Servicing\\Notifications\\OptionalFeatures\\Microsoft-Windows-Subsystem-Linux'
+  )
+
+  // TODO - does this check need to be more specific?
```
JordanMussi: Not sure if just the presence of the registry key is enough to confirm it's been installed?

shiftkey: I think this is fine - when the key doesn't exist, you'll get an empty array.

app/src/lib/shells/win32.ts
```diff
@@ -213,6 +240,64 @@ async function findGitBash(): Promise<string | null> {
...
+  if (registryPath.length === 0) {
+    return false
+  }
+
+  // Installed, but check for executable on disk as well!
+  return true
```
shiftkey: :art: fold this up into the previous check:

```
return registryPath.length > 0
```

app/src/lib/shells/win32.ts
```diff
@@ -213,6 +240,64 @@ async function findGitBash(): Promise<string | null> {
...
+}
+
+async function findWsl(): Promise<string | null> {
+  const isInstalled = await isWslInstalled()
+  if (!isInstalled) {
+    return null
+  }
+
+  const path = Path.join('C:\\Windows\\System32', 'wsl.exe')
```
shiftkey: Like we did in #4085, these should use `SYSTEMROOT` if set:

```ts
const windowsRoot = process.env.SystemRoot || 'C:\\Windows'
const path = Path.join(windowsRoot, 'System32', 'wsl.exe')
```

app/src/lib/shells/win32.ts
```diff
@@ -213,6 +240,64 @@ async function findGitBash(): Promise<string | null> {
...
+  return null
+}
+
+async function findWslBash(): Promise<string | null> {
+  const isInstalled = await isWslInstalled()
+  if (!isInstalled) {
+    return null
+  }
+
+  const path = Path.join('C:\\Windows\\System32', 'bash.exe')
```
shiftkey: Same as above:

```ts
const windowsRoot = process.env.SystemRoot || 'C:\\Windows'
const path = Path.join(windowsRoot, 'System32', 'bash.exe')
```

app/src/lib/shells/win32.ts
```diff
@@ -12,6 +13,8 @@ export enum Shell {
   PowerShellCore = 'PowerShell Core',
   Hyper = 'Hyper',
   GitBash = 'Git Bash',
+  Wsl = 'WSL',
```
shiftkey: At this point I think most of my questions are about what we display to the user:

 - It'd be nice to know which distro or shell you're going to end up in. Is that possible without launching `wsl.exe`?
 - If `wsl.exe` potentially puts you in the Bash shell for whichever distro you have, does that negate the need to also have a `WSL Bash` entry?

shiftkey:
@JordanMussi this has sat for a while with pending feedback. Is this something you're interested in picking up again to try and wrap up?

dlloyd123:
I noticed the PR seems to have gone cold, and wondered if you are interested in another implementation I have of this. I did this a while ago in my spare time but left it as I saw someone had already done it (this PR). Being as this seems to have gone cold, I have a bit of a more complete solution (covering all distro's and calling the command line wslconfig to parse out configured distros). Let me know if you want this? I would create a branch and PR but I think I would need permission(s) to do so.

[win32.zip](https://github.com/desktop/desktop/files/2194958/win32.zip)

JordanMussi:
Hi @shiftkey I’ve been rather busy lately and am currently away until next week. I’d definitely like to get this wrapped up when I’m back.

@dlloyd123 just fork the desktop/desktop repository and push your changes to your fork, would be interesting to see what you’ve done differently!

app/src/lib/shells/win32.ts
JordanMussi: - As I commented on the original issue, `wslconfig /list` gives a list of distributions and the default one is launched when using `wsl.exe`. So that can be parsed to work out which distribution will be launched https://github.com/desktop/desktop/issues/4641#issuecomment-387864749
- Yes I suppose it does. It would also require launching WSL to work out if you have setup a different shell and since it is probably a rare use case might be best to avoid confusion

shiftkey:
@JordanMussi no problem, thanks for the update!

@dlloyd123 feel free to follow the advice from @JordanMussi and publish a branch on your fork - this will help us easily see and compare the changes to what's currently here!

dlloyd123:
no problem will do so in a bit

dlloyd123:
Forked a copy and created branch and a PR into my master for you to look at. Happy to amend anything if you wish to use it, or feel free to lift anything from it for this PR.

@JordanMussi @shiftkey 

https://github.com/dlloyd123/desktop/pull/1

shiftkey:
@JordanMussi I'm going to close this out for now to help keep our outstanding PR queue under control, but feel free to comment here when you're back and have time to wrap this up.

@dlloyd123 if you're keen to see this land, feel free to open your own PR that either builds upon this work or does it a different way.

dlloyd123:
Sure ill raise a PR in the next few days, i will have to re-test locally to make sure any new changes havent affected this, and it all builds/passes tests etc.

LinqLover:
Hi, what is the current state of this issue? Can I use GitHub Desktop to control git installed within wsl?

shiftkey:
@LinqLover no, GitHub Desktop uses a version of Git that's embedded in the app.

We added support for opening a WSL shell from within the app in #7821, which is available on the [beta channel](https://github.com/desktop/desktop/#beta-channel) if you'd like to test it out and provide feedback.

PR #4659, opened 2018-05-11
attach scopes associated with account when refreshing

shiftkey:
This is a bit of groundwork that I uncovered when implementing SSH troubleshooting, and it helps the application to answer this simple question: **what permissions does a token currently have?**

This doesn't cover the actual re-authentication flow, because it will require the user to sign in again and it'd be nice to fit this in around the workflow they're currently performing. If they were using the webflow to sign in, they'd be greeted with a dialog like this: 

I've marked this as `infrastructure` because there's no user-facing changes in here.

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -232,6 +232,11 @@ function toGitHubIsoDateString(date: Date) {
   return date.toISOString().replace(/\.\d{3}Z$/, 'Z')
 }
 
+type UserWithTokenScopes = {
+  user: IAPIUser
+  scopes: ReadonlyArray<string>
+}
+
 /**
  * An object for making authenticated requests to the GitHub API
  */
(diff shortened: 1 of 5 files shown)


app/src/lib/api.ts
```diff
@@ -231,6 +231,17 @@ function toGitHubIsoDateString(date: Date) {
   return date.toISOString().replace(/\.\d{3}Z$/, 'Z')
 }
 
+type UserWithTokenScopes = {
```
iAmWillShepherd: Just curious why you chose a type here instead of an interface?

app/src/lib/api.ts
```diff
@@ -279,12 +290,18 @@ export class API {
...
-      const result = await parsedResponse<IAPIUser>(response)
-      return result
+      const user = await parsedResponse<IAPIUser>(response)
+      const scopeHeader = response.headers.get('X-OAuth-Scopes')
+      const scopes =
+        scopeHeader == null
+          ? []
+          : scopeHeader.split(',').map(scope => scope.trim())
+
+      return { user, scopes }
```
iAmWillShepherd: Is it so you wont have to explicitly annotate the return value?

app/src/lib/stores/accounts-store.ts
```diff
@@ -43,6 +43,7 @@ interface IAccount {
   readonly avatarURL: string
   readonly id: number
   readonly name: string
+  readonly scopes?: ReadonlyArray<string>
```
iAmWillShepherd: Does this have to be optional? I'm thinking we can require this be set and use `[]` if we have no scopes.

app/src/lib/stores/accounts-store.ts
```diff
@@ -182,7 +183,8 @@ export class AccountsStore extends BaseStore {
         account.emails,
         account.avatarURL,
         account.id,
-        account.name
+        account.name,
+        account.scopes || []
```
iAmWillShepherd: Actually you do what I just mentioned ☝️, so maybe we can get rid of the optional bit for `scopes`.

app/src/lib/api.ts
shiftkey: Stylistic reasons mostly, not something I'm hugely attached to.

This is also something internal to Desktop rather than an API response, so I avoided following that pattern here.

app/src/lib/api.ts
shiftkey: Nah, I don't think this behaviour is different between interfaces and plain type signatures

app/src/lib/stores/accounts-store.ts
shiftkey: This bit is important because above here we're parsing raw JSON into `IAcccount[]` - and those might be existing . So for backwards-compatibility I'd like to keep that around.

shiftkey: Actually, I'm going to drop this from review and see if there's a chance to migrate this data properly. I want `IAccount` to be as accurate as possible here, and leaving optional fields around does feel like a code smell.

shiftkey:
Maybe I'll come back to this. Maybe not.

Issue #4369, opened 2018-03-31
Enable React 16.3’s strict mode to catch bad practices


j-f1:
React 16.3 introduces a new [strict mode](https://reactjs.org/docs/strict-mode.html), which is enabled by wrapping all or a portion of your components with the `<React.StrictMode>` component.

According to the docs, it:

- identifies usages of lifecycle methods that will be deprecated in React 17 (`componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate`; [migration docs are here](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html))
- warns about using `ref="string"` (there’s a new `React.createRef()` type now that we could use instead of ref functions)
- and runs `constructor`, `render`, the new `getDerivedStateFromProps`, and the first argument to `setState` (if it’s a function) twice to ensure that they’re idempotent

All of these changes only occur when running in dev mode, so behavior in production shouldn’t change.

What do you think about enabling this @desktop/core? (I can’t ping you since the team is private)

shiftkey:
Revisiting this thread, there's a couple of issues I'd like to address before enabling this for development mode:

 - a lot of components are using these old lifecycle functions, so it'll generate a lot of noise in the console until we migrate them to the new API
 - these components aren't really covered by any tests beyond the integration test which isn't satisfactory to catch subtle changes - it'd be nice to use something lean like [`react-testing-library`](https://github.com/kentcdodds/react-testing-library) to capture the "before" behaviour of each component before we open it up to migrating

It's easy to find all mentions of `componentWillMount`, `componentWillReceiveProps` and `componentWillUpdate` in the codebase, but I'd like to craft up a list of refactoring issues that others can help out with so we can track this migration work - on top of everything else we need to worry about currently.

shiftkey:
I've opened #5092 to get `react-virtualized` up to date, and #5093 to track upgrading `react-transition-group`. Once those are in we can look at components in the codebase and figure out that migration strategy.

say25:
I sorta believe this can be closed as its superseded by https://github.com/desktop/desktop/issues/9875 which is 2 years newer

PR #4676, opened 2018-05-14
bump to React 16.3

shiftkey:
This version of React is an additive release with more [info in the blog](https://reactjs.org/blog/2018/03/29/react-v-16-3.html) but for now we can step up to this release with only a couple of breaking changes called out through the typings...


shiftkey:
This also unblocks #4369 to help investigate and assist with the transition to the next version of React.

nerdneha:
do we see any risks in pushing this into 1.2 last minute? or should we wait for 1.2.1?

PR #4691, opened 2018-05-15
add target branch to merge call-to-action

### LINKED ISSUE #4689: [Compare Branches] add a bit more context to the merge call-to-action
There's a bit of confusion here about the result of this merge:

It's been recommended to mention the target of the merge here too:

shiftkey:
Fixes #4689

Before:

After:

screen shot 2018-05-15 at 12 40 36 pm

This is being tested with a long branch name and with the minimal width, to ensure this is readable.

At full width, this is what the new message looks like:

screen shot 2018-05-15 at 12 42 34 pm

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -258,12 +258,12 @@ export class CompareSidebar extends React.Component<
           Merge into <strong>{this.props.currentBranch.name}</strong>
         </Button>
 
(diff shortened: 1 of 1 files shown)


nerdneha:
this could get a little janky if there is a long "from" and "into" branch but i still think it's worth it!
@donokuda for any thoughts on that real estate

PR #4693, opened 2018-05-15
Rename to `compare to branch` menu option to `show history`

### LINKED ISSUE #4684: [Compare Branches] Rename `compare to branch` menu item
Rename `Compare to Branch` to `Show History` (note: this is reverting back to what's in production)

iAmWillShepherd:
Fixes #4684

screen shot 2018-05-14 at 10 16 59 pm

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -132,7 +132,7 @@ export function buildDefaultMenu(
         click: emit('create-commit'),
       },
       {
-        label: __DARWIN__ ? 'History' : 'Hist&ory',
+        label: __DARWIN__ ? 'Show History' : 'Show &History',
         id: 'show-history',
         accelerator: 'CmdOrCtrl+2',
         click: emit('show-history'),
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -132,7 +132,7 @@ export function buildDefaultMenu(
         click: emit('create-commit'),
       },
       {
-        label: __DARWIN__ ? 'Show History' : 'Show &History',
+        label: __DARWIN__ ? 'Show History' : '&History',
         id: 'show-history',
         accelerator: 'CmdOrCtrl+2',
         click: emit('show-history'),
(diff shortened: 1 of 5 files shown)


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -132,10 +132,10 @@ export function buildDefaultMenu(
         click: emit('create-commit'),
       },
       {
-        label: __DARWIN__ ? 'Compare to Branch' : 'C&ompare',
-        id: 'compare-to-branch',
+        label: __DARWIN__ ? 'History' : 'Hist&ory',
```
shiftkey: Two suggestions:
 - I think we prefix this with a verb, like `View History` or `Show History`, for macOS.
 - For non-macOS we originally had `&H` as the keyboard shortcut here (see 24f20606d700fd216bc3c0666342f0222cf6fab1) so I think we can put that back here without having a conflict.

app/src/ui/app.tsx
```diff
@@ -243,7 +243,7 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.pull()
       case 'create-commit':
         return this.createCommit()
-      case 'compare-to-branch':
+      case 'show-history':
         return this.compareToBranch()
```
nerdneha: should we change this method name also?

app/src/main-process/menu/build-default-menu.ts
shiftkey: > IMO, I feel that this menu item belongs under the Repository menu since we want to show history for the repository.

For the moment let's focus just on the name of the menu item. We've been using the View menu as a gathering point to collect all the options for accessing parts of the main user interface, but we can have a discussion about whether that still makes sense later.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -132,10 +132,10 @@ export function buildDefaultMenu(
         click: emit('create-commit'),
       },
       {
-        label: __DARWIN__ ? 'Compare to Branch' : 'C&ompare',
-        id: 'compare-to-branch',
+        label: __DARWIN__ ? 'Show History' : 'Show &History',
```
shiftkey: Just `&History` here for non-Darwin platforms

iAmWillShepherd:
💯

PR #4694, opened 2018-05-15
redirect to commit tab when merge conflicts found

### LINKED ISSUE #4686: [Compare Branches] When merge conflict error message dismissed, switch to Commit tab
This was a bit of feedback that came up as part of user testing for the new Compare Branch feature:

> Almost everyone, if not everyone, expected to see the diff for the conflicts once they encountered merge conflicts (either in a popup or in the commits tabs)

It should be easy enough to introduce a specialized error handler which, when you dismiss the error message, signals to the app to switch the user to the Commit tab.

shiftkey:
Fixes #4686 

Before:

merge-conflicts-before

After:

merge-conflicts-after

Open to feedback for improvements to the copy and dialog flow - I felt that having an explicit (and default) action to view the conflicts felt better than a single button.

### DIFF
--- a/app/src/ui/index.tsx
+++ b/app/src/ui/index.tsx
@@ -12,6 +12,7 @@ import {
   gitAuthenticationErrorHandler,
   externalEditorErrorHandler,
   openShellErrorHandler,
+  mergeConflictHandler,
   lfsAttributeMismatchHandler,
   defaultErrorHandler,
   missingRepositoryHandler,
@@ -138,6 +139,7 @@ dispatcher.registerErrorHandler(defaultErrorHandler)
 dispatcher.registerErrorHandler(upstreamAlreadyExistsHandler)
 dispatcher.registerErrorHandler(externalEditorErrorHandler)
 dispatcher.registerErrorHandler(openShellErrorHandler)
+dispatcher.registerErrorHandler(mergeConflictHandler)
 dispatcher.registerErrorHandler(lfsAttributeMismatchHandler)
 dispatcher.registerErrorHandler(gitAuthenticationErrorHandler)
 dispatcher.registerErrorHandler(pushNeedsPullHandler)
--- a/app/src/ui/merge-conflicts/merge-conflicts-warning.tsx
+++ b/app/src/ui/merge-conflicts/merge-conflicts-warning.tsx
@@ -1,0 +1,55 @@
+import * as React from 'react'
+import { Button } from '../lib/button'
+import { ButtonGroup } from '../lib/button-group'
+import { Dialog, DialogContent, DialogFooter } from '../dialog'
+import { Dispatcher } from '../../lib/dispatcher'
+import { RepositorySection } from '../../lib/app-state'
+import { Repository } from '../../models/repository'
+
+interface IMergeConflictsWarningProps {
+  readonly dispatcher: Dispatcher
+  readonly repository: Repository
+  readonly onDismissed: () => void
+}
+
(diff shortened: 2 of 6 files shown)


app/src/ui/merge-conflicts/merge-conflicts-warning.tsx
```diff
@@ -0,0 +1,56 @@
...
+          </p>
+        </DialogContent>
+
+        <DialogFooter>
+          <ButtonGroup>
+            <Button type="submit">
+              {__DARWIN__ ? 'View Conflicts' : 'View conflicts'}
+            </Button>
+            <Button onClick={this.props.onDismissed}>
+              {__DARWIN__ ? 'Cancel' : 'Cancel'}
```
nerdneha: do we need this conditional here for windows v macOS since it's one word?

nerdneha: not feeling great about `Cancel` -- when I saw the gif, it gave me hope that you had somehow magically addressed "abort". To me `Close` implies you're just stuck here and `Cancel` means you could get out of it. But we can't 😭

app/src/ui/merge-conflicts/merge-conflicts-warning.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Dialog
+        id="merge-conflicts-warning"
+        title={__DARWIN__ ? 'Merge Conflicts Found' : 'Merge conflicts found'}
+        onDismissed={this.props.onDismissed}
+        onSubmit={this.onSubmit}
+      >
+        <DialogContent>
+          <p>Conflicts were detected as part of the last merge operation.</p>
+          <p>
+            You will need to view and resolve these conflicts before finishing
```
nerdneha: should we specifically say `You will need to view and resolve these conflicts via your IDE or terminal before finishing...`? or should we be vague. I just don't want anyone to get the impression that Desktop can help them at this point

app/src/ui/merge-conflicts/merge-conflicts-warning.tsx
shiftkey: LOL, nah :fire:

shiftkey: This is a good shout - the original button had `Close` so I'm fine to use that here instead of `Cancel`.

app/src/ui/merge-conflicts/merge-conflicts-warning.tsx
shiftkey: I settled on `... via your editor or shell before finishing ...` to stick to concepts we talk about in Desktop.

shiftkey:
🍎

app/src/ui/index.tsx
```diff
@@ -138,6 +139,7 @@ dispatcher.registerErrorHandler(defaultErrorHandler)
 dispatcher.registerErrorHandler(upstreamAlreadyExistsHandler)
 dispatcher.registerErrorHandler(externalEditorErrorHandler)
 dispatcher.registerErrorHandler(openShellErrorHandler)
+dispatcher.registerErrorHandler(mergeConflictHandler)
```
shiftkey: > should we be feature flagging any of this?

We could feature-flag this registration here, but given the new flow is only slightly different to what we had before I'm not hugely worried about it.

nerdneha:
Tested on both the drop down option AND via the merge button in compare. 👍 
 * view commits works for both
 * close works for both

PR #4698, opened 2018-05-15
Remove "Discard All Changes..." from File Context Menu

graemeworthy:
Reasons:

- This makes it very easy to delete all your work by accident.
- The File Context Menu typically performs actions on a specific file, not on all Files.
- There is currently no undo to 'Discard All Changes...'
-  It does not say 'Discard Changes in All Files'

There is a Context Menu on The [{x} Changed Files] header which provides this same functionality, in a less confusing context.

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -272,10 +272,6 @@ export class ChangesList extends React.Component<
               : `Discard ${paths.length} selected changes…`,
         action: () => this.onDiscardChanges(paths),
       },
-      {
-        label: __DARWIN__ ? 'Discard All Changes…' : 'Discard all changes…',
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
Thanks for taking the time to open a PR 🥇 

Unfortunately, this isn't something we will merge in because it is removing functionality that is a part of the applications design. If you are having trouble with deleting your files on accident, you can require that you get a confirmation dialog by going to Preferences -> Advanced and checking the last checkbox labelled _Show confirmation dialog before discarding changes_.

If doing that doesn't work for you, please open an issue instead and we can have that discussion.

j-f1:
In the forthcoming update, we’ve changed this button to always prompt for confirmation before discarding everything.

PR #4702, opened 2018-05-15
Adds compare to branch menu item

### LINKED ISSUE #4683: [Compare Branches] Add compare menu item under Branch menu
We want to support multiple ways of comparing a branch, so let's add a new menu item `Compare to Branch` under the Branch top level menu. 

**Add menu item**

**Focus on branch list**

We also should focus the compare branch list so that we see a list of all branches instead of being taking to a text box and doing nothing.

iAmWillShepherd:
Fixes #4683

**After**

**Repo with only 1 branch**
screen shot 2018-05-17 at 2 24 10 pm

**Repo with 2 or more branches**
screen shot 2018-05-17 at 2 23 52 pm

- [x] Add menu item
- [x] Show history tab
- [x] Focus compare branch selector
- [x] Think of a short-cut: <kbd>Command+Shift+B</kbd>


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -297,11 +297,16 @@ export function buildDefaultMenu(
         click: emit('merge-branch'),
       },
       separator,
+      {
+        label: __DARWIN__ ? 'Compare on Desktop' : '&Compare on Desktop',
```
shiftkey: The screenshot shows that local and remote functions are grouped with the separator above - does it make sense to move this up alongside `Update from Default Branch` and `Merge Branch`?

shiftkey: Also I'm still a huge fan of `Compare to Branch` here to avoid excessive mentions of `Desktop` in the app.

app/src/ui/history/compare.tsx
```diff
@@ -33,6 +33,8 @@ interface ICompareSidebarProps {
   readonly dispatcher: Dispatcher
   readonly currentBranch: Branch | null
   readonly sidebarHasFocusWithin: boolean
+  /**Determines whether or not to focus compare branch selector */
+  readonly focusBranchSelector: boolean
```
shiftkey: Can we use a name here that's prefixed with `initial` to indicate that the value is only used when initializing the component. Something like `initialShowBranchList`?

app/src/ui/history/compare.tsx
```diff
@@ -33,6 +33,8 @@ interface ICompareSidebarProps {
   readonly dispatcher: Dispatcher
   readonly currentBranch: Branch | null
   readonly sidebarHasFocusWithin: boolean
+  /**Determines whether or not to focus compare branch selector */
```
shiftkey: :art: Let this comment breath a bit with some newlines and whitespace, and ensure it mentions that this is about the initial state of the branch selector.

app/src/lib/app-state.ts
```diff
@@ -316,11 +316,15 @@ export type Foldout =
...
 
-export enum RepositorySection {
+export enum RepositorySectionTab {
   Changes,
   History,
 }
 
+export type RepositorySection =
+  | { selectedTab: RepositorySectionTab.Changes }
+  | { selectedTab: RepositorySectionTab.History; focusBranchSelector: boolean }
```
shiftkey: Three things:

:art: can this name here also mention the `initial` prefix?
:art: can `focusBranchSelector` here be optional, and default to `false` when doing the lookup?
:art: can you split this type signature over two lines to avoid the `;`, or is this Prettier making me sad?

app/src/lib/dispatcher/dispatcher.ts
```diff
@@ -11,13 +11,13 @@ import {
...
 import { DiffSelection } from '../../models/diff'
 import {
-  RepositorySection,
   Popup,
   PopupType,
   Foldout,
   FoldoutType,
   ImageDiffType,
   CompareAction,
+  RepositorySection,
```
shiftkey: :art: tidy up this no-op change

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -296,12 +296,18 @@ export function buildDefaultMenu(
         accelerator: 'CmdOrCtrl+Shift+M',
         click: emit('merge-branch'),
       },
+      {
+        label: __DARWIN__ ? 'Compare to Branch...' : '&Compare to branch...',
```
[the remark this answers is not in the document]
iAmWillShepherd: That's good to know. 👍

app/src/ui/history/compare.tsx
iAmWillShepherd: I'll use that until I can think of something better...

app/src/lib/app-state.ts
iAmWillShepherd: I can't assign a default value in a type alias definition. I had to handle the `undefined` case in the component that uses that property.

iAmWillShepherd:
💯

shiftkey:
@nerdneha are you also able to help out with a review here - I was involved with tackling the problem and also merging in master, so I want ensure there are other opinions involved here.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -288,6 +288,12 @@ export function buildDefaultMenu(
         accelerator: 'CmdOrCtrl+Shift+U',
         click: emit('update-branch'),
       },
+      {
+        label: __DARWIN__ ? 'Compare to Branch' : '&Compare to branch',
```
shiftkey: This is the same problem as was uncovered in #4618 - both `&Compare to branch` and `&Compare on GitHub` cannot `&` the same letter. I'd suggest moving the latter to `Compare on &GitHub`

iAmWillShepherd: What is that ampersand represent in this case? And what is the level of uniqueness it must have? I see we use `&G...` in multiple places. Right now, I'm assuming sub-menus have their own scope, but I could not verify using the [docs](https://electronjs.org/docs/api/menu). Also, is there a method for picking which character to put the ampersand next to or is it arbitrary?

nerdneha: @iAmWillShepherd i found this in some documentation
> To create an access key for a menu item, precede any character in the item's text string with an ampersand. For example, the text string "&Move" causes the system to underline the letter "M".

https://msdn.microsoft.com/en-us/library/windows/desktop/ms647553(v=vs.85).aspx

i also found a helpful tip (For windows)
> Make sure all the access keys on a menu are unique by right-clicking your menu and choosing Check Mnemonics from the shortcut menu.

https://msdn.microsoft.com/en-us/library/9sewd7e1.aspx

i found some more best practices here: https://webaim.org/techniques/keyboard/accesskey
that indicates that the access key should be the character you show for the shortcut but nothing else that helped me understand which character

j-f1: @nerdneha It looks like that WebAIM link is for the HTML `accesskey` attribute, which is different from the shortcuts used here.

shiftkey: > What is that ampersand represent in this case? 

In Windows applications, when you press <kbd>ALT</kbd> the application menu becomes focused, with the access keys underlined:



This shows that you can open a given menu by pressing the relevant access key.

Submenus can also have access keys:



> And what is the level of uniqueness it must have?

The uniqueness does not need to be global, just within the given menu or submenu. In fact we have an example here where the top-level `Branch` menu item and the inner `New branch...` menu entries both use `B` (it's not case-sensitive) as a shortcut. This is fine, and if you do <kbd>ALT</kbd> + <kbd>B</kbd> + <kbd>B</kbd> you'll quickly open the new branch dialog.

> Also, is there a method for picking which character to put the ampersand next to or is it arbitrary?

I'm not sure of the rules here beyond my gut instincts, but essentially you want something that stands out - either the first letter of a sentence-cased word, or a scope-unique letter near the start of the menu label.

iAmWillShepherd:
@shiftkey does you [comment](https://github.com/desktop/desktop/issues/4725#issuecomment-390034427) apply here as well? I disabled the _Compare to Branch_ menu item when there are no branches to compare with.

app/src/lib/menu-update.ts
```diff
@@ -199,7 +202,6 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
     'create-commit',
     'show-history',
     'show-branches-list',
-    'open-external-editor',
```
shiftkey: Why have we removed this? This list of menu IDs is the ones to hide when a repository is missing from disk, and currently `Open in External Editor` remains enabled:

app/src/lib/menu-update.ts
```diff
@@ -248,6 +250,7 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
       'create-branch',
       !tipStateIsUnknown && !branchIsUnborn
     )
+    menuStateBuilder.setEnabled('compare-to-branch', hasBranchToCompare)
```
shiftkey: Like I mentioned in #4726, we should let the user switch to this view. We already have a helpful message here when you don't have branches to compare to:


So this is one step better than our other merge dialog. Can we tidy up this check?

shiftkey:
@nerdneha I'm going to merge this in when it's green so that we can put together a test build for further QA, but feel free to add any further thoughts you have about this PR

PR #4703, opened 2018-05-16
add option to move repository to trash when removing from app

### LINKED ISSUE #2108: Additional Check to delete repo from disk while removing repo
### Description

Additional Checkbox can be provided while removing a repo. Asking the user to check/uncheck `Delete permanently from disk`

### Version

**GitHub Desktop version:**  0.6.2

**OS version:**  MacOS Sierra

**Expected behaviour:** 

Provide additional checkbox allowing the user to permanently delete the repo from disk.
This will be unchecked by default.
If the user checks the checkbox before removing the repo can be deleted permanently from the disk.

**Actual behaviour:** 

Removing a repository removes only from GitHub desktop. Each time to delete, the user has to navigate and delete that manually.

say25:
Closes #2108


app/src/ui/remove-repository/confirm-remove-repository.tsx
```diff
@@ -29,6 +42,7 @@ export class ConfirmRemoveRepository extends React.Component<
   }
 
   public render() {
+    const trashName = __DARWIN__ ? 'Trash' : 'Recycle Bin'
```
say25: This concept is duplicated should it be made into a base class?

app/src/ui/app.tsx
```diff
@@ -746,12 +746,12 @@ export class App extends React.Component<IAppProps, IAppState> {
...
       })
     } else {
-      this.props.dispatcher.removeRepositories([repository])
+      this.props.dispatcher.removeRepositories([repository], false)
     }
   }
 
   private onConfirmRepoRemoval = (repository: Repository) => {
-    this.props.dispatcher.removeRepositories([repository])
+    this.props.dispatcher.removeRepositories([repository], false)
```
say25: I would like to use `state.includeMoveToTrash` here but I'm struggling to extract it :/

say25:
@desktop/core I'm looking for feedback/help/ideas at this point.

say25:
@shiftkey I'm looking for feedback/help/ideas at this point.

app/src/ui/remove-repository/confirm-remove-repository.tsx
shiftkey: I'd recommend a function that can be exported and imported, like this:

https://github.com/desktop/desktop/blob/4da77727b8430cd433a8ccab5f312eb5d464b0d3/app/src/ui/lib/context-menu.ts#L6-L8

app/src/ui/app.tsx
shiftkey: Check out how we're doing the Discard Changes dialog.

For the application state:

 - introduce some app state that represents the dialog behaviour

https://github.com/desktop/desktop/blob/88082bc0e75c99af4a3afc3f742b48b2c3612332/app/src/lib/app-state.ts#L161-L162

 - set some default state for this value in the App Store

https://github.com/desktop/desktop/blob/88082bc0e75c99af4a3afc3f742b48b2c3612332/app/src/lib/stores/app-store.ts#L241

https://github.com/desktop/desktop/blob/88082bc0e75c99af4a3afc3f742b48b2c3612332/app/src/lib/stores/app-store.ts#L585

For where you need to handle the action:

 - pass the application state down through props

https://github.com/desktop/desktop/blob/88082bc0e75c99af4a3afc3f742b48b2c3612332/app/src/ui/changes/sidebar.tsx#L49

 - in the event handler, react depending on the current state

https://github.com/desktop/desktop/blob/master/app/src/ui/changes/sidebar.tsx#L158-L168

And for the dialog itself:

 - initial state for the dialog is derived from props (that comes from app state)

https://github.com/desktop/desktop/blob/0ce3de645441bb4a30506636210edbaa6975a970/app/src/ui/discard-changes/discard-changes-dialog.tsx#L52-L55

 - when submitting, the action is performed and the state is updated through the dispatcher

https://github.com/desktop/desktop/blob/88082bc0e75c99af4a3afc3f742b48b2c3612332/app/src/ui/discard-changes/discard-changes-dialog.tsx#L136-L146

say25:
@shiftkey, read issue again and realized that the state idea isn't what is asked for by the original requestor. 

With these changes:
- When removing a repo with the confirm modal, it will give the user the option to the move the repo to the trash as well
- Missing Repos are uneffected

app/src/ui/remove-repository/confirm-remove-repository.tsx
```diff
@@ -45,9 +66,26 @@ export class ConfirmRemoveRepository extends React.Component<
...
           </p>
           <p className="description">
-            The repository will be removed from GitHub Desktop but will remain
-            on disk.
+            The repository will be removed from GitHub Desktop.
           </p>
+
+          <div>
+            <p>
+              <strong>
```
shiftkey: Not sure how I feel about this part. Here's how it looks in the rest of the dialog:

app/src/ui/app.tsx
```diff
@@ -746,12 +746,15 @@ export class App extends React.Component<IAppProps, IAppState> {
         repository,
       })
     } else {
-      this.props.dispatcher.removeRepositories([repository])
+      this.props.dispatcher.removeRepositories([repository], false)
```
shiftkey: I was worried that, for cases where the user has said they don't want to be prompted, the app will use the set value for `deleteRepoFromDisk` and you'd potentially find yourself in a situation where you weren't aware you were moving repositories to the trash. But this seems like a good default if you choose not to be prompted about removing it.

app/src/ui/app.tsx
```diff
@@ -736,7 +736,7 @@ export class App extends React.Component<IAppProps, IAppState> {
     }
 
     if (repository instanceof CloningRepository || repository.missing) {
```
shiftkey: There's a really annoying corner case here that I'll call out because I don't have a good answer right now, but if you're cloning a repository and try to remove it in the application, we don't abort the clone, so you end up in a situation where the clone ends up on disk and not in the app.

I think we need to have the ability to tear down a in-flight clone if the user asks for it, but that's a minor thing in the scope of this PR.

app/src/ui/app.tsx
say25: Yea it kinda sucks to me because I think there might need to be another setting for this. Just cause when you are avoiding prompts you may still want to delete them from disk? just thought it would be better than it was today to allow the option when prompted.

app/src/ui/remove-repository/confirm-remove-repository.tsx
say25: I can have it behave either way the team prefers.

shiftkey: Let's see how having _just the checkbox and message_ feels. Add an updated screenshot to this PR and we'll see how the team feels...

say25: Current: (as of this PR) 
image

Proposed:
image

say25: Maybe another option?
image

donokuda: I'm in favor of this direction: remove

I don't think it's necessary to try to emphasize the option to move given that it's not a destructive action in case the person misses that checkbox.

app/src/ui/app.tsx
shiftkey: I'm inclined to put this aside for now and see if we get any reports on remembering this setting across sessions. I think we will, and I think we can track this setting without need to have a UI for it. But let's not overthink this change just yet.

say25:
@shiftkey Im sure you are busy but any other requested changes?

lee-dohm:
What if the repository is on a drive that doesn't have a "Trash" or "Recycle Bin"? Will it detect that and not show the option? cf https://github.com/atom/tree-view/issues/856

shiftkey:
@lee-dohm are you using the [`shell.moveItemToTrash(fullPath)`](https://electronjs.org/docs/api/shell#shellmoveitemtotrashfullpath) for this? Technically we're already using this as part of [the discard changes flow](https://github.com/desktop/desktop/blob/master/docs/technical/discard-changes.md), so I'm surprised we've not encountered this issue already due to that...

lee-dohm:
I'd have to dig into the code but, to my recollection, yes. I'll take a look later today and report back.

lee-dohm:
@shiftkey Yes, it seems so:

https://github.com/atom/tree-view/blob/208d0bb2fb5ddd072f4039230f6a3887222065fa/lib/tree-view.coffee#L660

shiftkey:
@lee-dohm I've moved the discussion about `shell.moveItemToTrash` on Windows out to #5157.

I think there might be some iterations on this down the track, but for now this is a great first step. Thanks again @say25!

PR #4705, opened 2018-05-16
bring back the hover styles for the app menu

### LINKED ISSUE #4636: App menu foldout list items do not change on focus
## Description

The styling for this was removed by #4188 in https://github.com/desktop/desktop/commit/e0d7d9f4aefead97c3792eff6e1fb14ab75a18fe#diff-cad62e98a2e7c4bc492c0a912c3eb447L87.

## Version

* GitHub Desktop: 1.1.2-beta5

## Steps to Reproduce

1. Hover over any of the main app menu items to open its foldout
2. Move the mouse down to over over one of the list items

### Expected Behavior

The list item should behave as in 1.1.1 where the background is blue with white text on hover.

### Actual Behavior

No change on hover.

shiftkey:
Fixes #4636

This was an unfortunate side-effect of #4188 but hasn't made it to `production` yet, so it doesn't need it's own release entry in the changelog.

Before:

After:

~~I also uncovered a keyboard navigation issue in here while testing - I'll extract another issue with details for reproducing it.~~ EDIT: nevermind, I don't think this was related to the multi-select changes. Keeping an eye on this to see if I can nail down what's wrong with it - basically you can _sometimes_ make the menu disappear/lose focus when you expand the app menu and move left/right between menu items.

### DIFF
--- a/app/styles/ui/window/_app-menu-bar.scss
+++ b/app/styles/ui/window/_app-menu-bar.scss
@@ -42,5 +42,15 @@
       background: var(--background-color);
       pointer-events: all;
     }
+
+    .menu-item:hover {
+      &:not(.disabled) {
+        --text-color: var(--box-selected-active-text-color);
+        --text-secondary-color: var(--box-selected-active-text-color);
+
+        color: var(--text-color);
+        background-color: var(--box-selected-active-background-color);
+      }
+    }
   }
 }


app/styles/ui/window/_app-menu-bar.scss
```diff
@@ -42,5 +42,15 @@
       background: var(--background-color);
       pointer-events: all;
     }
+
+    .menu-item:hover {
+      &:not(.disabled) {
```
j-f1: You should be able to merge these selectors together.

shiftkey: 👍 It took me a fair bit to figure out the right syntax here, so I think leaving it like this for future peeps is a safe bet.

icosamuel:
nice catch

PR #4707, opened 2018-05-16
blur branch selector when search text is cleared

### LINKED ISSUE #4690: [Compare Branches] Clearing branch input takes users to user history
## Goal
As a user, when I'm done comparing my branch to another branch, I want to see the full history so that I can see the status of my branch and anything that was recently merged

#### Background
After user testing, we found that:
1. 1 of our 6 user testers could find the fully history view within 2min
2. People expected that when they cleared the user input field, something would change

We want to make the full history view easier to find! Right now you can get to the full history view clearing the branch input text and THEN removing focus on from the input 

## Solution
Remove the step for removing focus AFTER the text is cleared
 - [ ] when a user clears the branch input field by pressing `x`, the branch input field is cleared AND they are taken to the full history view
 - [ ] when a user clears the user input field by hitting <kbd>esc</kbd>, the branch input field is cleared AND they are taken to the full history view

#### Existing (Before)
full-history-old

#### Suggested Solution (After)
full-history-new

#### Related Tasks
- [ ] inform help-docs on user-facing updates
- [ ] update marketing page screenshots and add feature page
- [ ] update blog post for GitHub blog

shiftkey:
Fixes #4690 

### Clearing filter text before selecting a branch to compare

Before:

After:

### Clearing filter text after selecting a branch to compare

Before:

After:

 - [x] tested clearing with only filter text (no branch selected) on macOS
 - [x] tested clearing after selecting branch on macOS
 - [x] tested clearing with only filter text (no branch selected) on Windows
 - [x] tested clearing after selecting branch on Windows

### DIFF
--- a/app/src/ui/lib/fancy-text-box.tsx
+++ b/app/src/ui/lib/fancy-text-box.tsx
@@ -47,6 +47,7 @@ export class FancyTextBox extends React.Component<
           placeholder={this.props.placeholder}
           onKeyDown={this.props.onKeyDown}
           onValueChanged={this.props.onValueChanged}
+          onSearchCleared={this.props.onSearchCleared}
           tabIndex={this.props.tabIndex}
           ref={this.props.onRef}
         />
--- a/app/src/ui/lib/fancy-text-box.tsx
+++ b/app/src/ui/lib/fancy-text-box.tsx
@@ -38,17 +38,9 @@ export class FancyTextBox extends React.Component<
       <div className={componentCSS}>
         <Octicon className={octiconCSS} symbol={this.props.symbol} />
         <TextBox
-          value={this.props.value}
+          {...this.props}
           onFocus={this.onFocus}
           onBlur={this.onBlur}
-          autoFocus={this.props.autoFocus}
-          disabled={this.props.disabled}
-          type={this.props.type}
-          placeholder={this.props.placeholder}
-          onKeyDown={this.props.onKeyDown}
-          onValueChanged={this.props.onValueChanged}
-          onSearchCleared={this.props.onSearchCleared}
-          tabIndex={this.props.tabIndex}
           ref={this.props.onRef}
         />
       </div>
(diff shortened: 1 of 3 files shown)


app/src/ui/lib/fancy-text-box.tsx
```diff
@@ -38,16 +38,9 @@ export class FancyTextBox extends React.Component<
       <div className={componentCSS}>
         <Octicon className={octiconCSS} symbol={this.props.symbol} />
         <TextBox
-          value={this.props.value}
+          {...this.props}
```
iAmWillShepherd: Is the new pattern we're going to adopt?

app/src/ui/lib/text-box.tsx
```diff
@@ -162,8 +167,33 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
     })
   }
 
+  private onSearchTextCleared = (ev: Event) => {
```
iAmWillShepherd: No need for the `ev` parameter.

iAmWillShepherd: No need for the `ev` parameter.

app/src/ui/lib/fancy-text-box.tsx
shiftkey: I'd say this is an exceptional case. Initially I added a new prop to `TextBox` and because `FancyTextBox`'s props extends the props for `TextBox` I thought it'd just work, but it didn't because we were passing in each prop explicitly and the new one was omitted. 

So after confirming explicitly passing the new prop in I went back and cleaned this up so that it wouldn't happen to the next person touching this code. I think we agreed that `FancyTextBox` isn't a long-term thing, so I don't want to encourage this either unless we find specific cases like this that will prevent mistakes.

Alternatively, I can revert this cleanup code. Let me know if that's a more conservative change in the context of this PR.

shiftkey: Reverted :boom:

shiftkey:
🍎

Issue #4574, opened 2018-05-01
Cannot pull after cloning a truly empty repository


MCF:
## Description

If you clone a truly empty repository (no files, no commits yet) from an SSH source github desktop will not let you pull or fetch.  The normal pull/fetch/push tab shows the message "cannot publish unborn head" and does nothing.  The push and pull menu items in the _Repository_ menu are disabled.

If someone else adds files and commits to the remote repository you cannot subsequently do a pull or fetch to pull them in via github desktop.  Obviously you can drop into command line git to do so but this does not seem like an advanced use-case that would require git to solve.

## Version

* GitHub Desktop: 1.1.1
* Operating system: Windows 7, 64 bit

## Steps to Reproduce
Clone an empty repository (no commits) from an SSH source

### Expected Behavior
You would still be able to pull/fetch from the remote.

### Actual Behavior
Pull/fetch is disabled (apparently) through all controls.

### Logs
Nothing relevant in the logs.

shiftkey:
@MCF thanks for the report. I've been able to reproduce this, and I think it's down to how we handle the unborn repository.

The behaviour for this button is in [`push-pull-button.tsx`](https://github.com/desktop/desktop/blob/master/app/src/ui/toolbar/push-pull-button.tsx), and I'll come back to this later to add more details if someone is interested in contributing a fix.

shiftkey:
A rough patch for this:

```diff
diff --git a/app/src/ui/toolbar/push-pull-button.tsx b/app/src/ui/toolbar/push-pull-button.tsx
index ac499b989..2c6370f1e 100644
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -68,7 +68,8 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     // open the publish dialog - we'll handle publishing the current branch afterwards
     // if it exists
     const validState = this.props.remoteName
-      ? this.props.tipState === TipState.Valid
+      ? this.props.tipState === TipState.Valid ||
+        this.props.tipState === TipState.Unborn
       : true
 
     const disabled = !validState || networkActive
@@ -127,6 +128,13 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     if (!this.props.remoteName) {
       return 'Publish repository'
     }
+
+    // repository has a remote but no branch to manage
+    //  -> can fetch to see if someone else has pushed remote commits
[...]

PR #4710, opened 2018-05-16
Can't pull after cloning an empty repository

agisilaos:
Fixes: #4574

I followed @shiftkey's advice and implemented his rough patch for fixing this issue. One of his comments though was:

> This feels suboptimal because we're repeating the same check in a few places, but that's a side-effect of how the component is currently organized. But that should be a starting point for someone to dig into.

Note: Maybe we can change some things about how this component is organized so we don't repeat the same check as much? I'm open to suggestions.


agisilaos:
This is the fresh new one @shiftkey! it's ready for review. 🍄

agisilaos:
Will look into that again today. Thank you for the heads up @shiftkey!

agisilaos:
@shiftkey Spent an hour or so on this one, but I'm not sure what change we need to do regarding the order of the checks? How do you think the checks should look like and in what order? 🤔

shiftkey:
> Spent an hour or so on this one, but I'm not sure what change we need to do regarding the order of the checks?

What I was referring to here was how we check the `aheadBehind` details (which are only set when we have a valid tip) before we check that the repository is unborn.

Here's a rough patch that I think fixes it:

```diff
diff --git a/app/src/ui/toolbar/push-pull-button.tsx b/app/src/ui/toolbar/push-pull-button.tsx
index 400aa2bf8..ec4ed9fea 100644
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -128,9 +128,6 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     if (!this.props.remoteName) {
       return 'Publish repository'
     }
-    if (!this.props.aheadBehind) {
-      return 'Publish branch'
-    }
 
     // repository has a remote but no branch to manage
     // -> can fetch to see if someone else has puhed remote commits
@@ -138,6 +135,10 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
       return 'Fetch'
     }
 
+    if (!this.props.aheadBehind) {
+      return 'Publish branch'
+    }
+
     const { ahead, behind } = this.props.aheadBehind
     const actionName = (function() {
       if (behind > 0) {
@@ -160,14 +161,15 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     if (!this.props.remoteName) {
       return OcticonSymbol.cloudUpload
     }
-    if (!this.props.aheadBehind) {
-      return OcticonSymbol.cloudUpload
-    }
 
     if (this.props.tipState === TipState.Unborn) {
       return OcticonSymbol.sync
     }
 
+    if (!this.props.aheadBehind) {
+      return OcticonSymbol.cloudUpload
+    }
+
     const { ahead, behind } = this.props.aheadBehind
     if (this.props.networkActionInProgress) {
       return OcticonSymbol.sync
```

agisilaos:
Will test it out. Thanks, @shiftkey!

shiftkey:
Dropping `ready-for-review` to indicate there's still some work to be done. @agisilaos please bump this when it's ready for another go.

billygriffin:
Does this affect the use case where you're adding a new repo locally, and you try to publish it to GitHub, but you get a "Cannot publish unborn HEAD" error? I think it's essentially the opposite of this, so instead of not being able to pull, you aren't able to push, and the error message is not descriptive.

shiftkey:
@agisilaos just dropping in to say



> Does this affect the use case where you're adding a new repo locally, and you try to publish it to GitHub, but you get a "Cannot publish unborn HEAD" error?

@billygriffin it's been a while, so I tested out this workflow again with an unborn repository:

 - current version (1.3.4) - you can do the publish flow, I didn't get this error, but button is disabled (original issue)



 - this PR - the button is now enabled, but clicking doesn't show the publish dialog (no errors either)

agisilaos:
I was thinking about picking it up today. Will implement your new suggestion and I guess we can test it again afterward. 😍

shiftkey:
@agisilaos I'm not sure if you're still planning to come back to this, but do you mind if I close this out and we can reopen a fresh PR when you do?

agisilaos:
Yep let's do that as I have some ideas for this issue. Will open a fresh PR soon. @shiftkey.

PR #4713, opened 2018-05-17
jump to compare mode from the branch selector

### LINKED ISSUE #4709: [Compare Branches] access compare branch via branch dropdown
After user testing, we found that all users expected to be able to compare their branch to another branch from the branch dropdown

Right now the only way to get to the compare branch is through the `History` tab. 

## Solution
When you open the branches dropdown and hover over a branch
 - [x] a kabob icon shows up for all branches except your own / currently checked out branch

When you are hovered over the branch and click on the kabob icon
 - [x] a menu item shows up to "compare to this branch"

When you click on the `compare to this branch` menu item
 - [x] the branch menu item closes
 - [x] UI shows the history tab selected with the branch you clicked on populated in the `compare to branch` input

#### Existing (Before)
branch-list-before

#### Suggested Solution (After)

UI spike branch: https://github.com/desktop/desktop/tree/cant-compare-if-you-cant-find-the-feature

compare-discoverability-idea

#### Other considerations from the spike branch

- [x] Clicking on the kabob menu shouldn't change the branch list item's background to blue (You can see this happening in the suggested solution gif)
- [x] (Design/CSS) - kabob menu rectangle background container shouldn't be a solid color since it stands out against the blue background (You can also see this in the gif below)

shiftkey:
Fixes #4709

Hat tip to @donokuda for the CSS magic! 

TODO:

 - [x] wireup handler to switch to compare view
 - [x] kebab effect needs to be tweaked as merge dialog has same hover effect which still hides the timestamp (no DOM elements rendered)
 - [x] compare view sometimes shows empty branch name on first switch (restored when textbox is focused)
 - [x] a bunch more testing to ensure interactions and transitions are smooth as butter


app/styles/ui/_branches.scss
```diff
@@ -229,6 +264,11 @@
       text-align: right;
       white-space: nowrap;
     }
+
+    .menu {
+      background: red;
```
shiftkey: @donokuda this seems unnecessary, right? :trollface:

shiftkey:
@donokuda one little visual thing I wanted to call out was the pressed state of the kebab button - the icon changes to white dots. Not sure if that's intended, but it feels a bit off:

donokuda:
Pushed a fix addressing the opaque background for the ellipsis button:

compare-branch-ellipsis-button

I also updated the merge summary area so that the "from x to y" part starts on a new line. I'm making an assumption that it'll be more readable:

shiftkey:
Just wanted to ensure I put this behind a feature flag in case we didn't want to ship this out to production ✅

PR #4714, opened 2018-05-17
provide this.canSelectRow here to ensure first valid branch is selected

shiftkey:
This fixes a known issue on `1.1.2-beta6` where you're unable to cursor down from the branch selector like you can do in other places within the app (it gets stuck in the header, which isn't a valid row to select).

Before:

After:

 - [x] can move down from filter textbox to branch list
 - [x] fix focus issue - list should be focused after first move, but textbox holds onto it because `setState` is not a blocking function

### DIFF
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -258,11 +258,11 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
     )
 
     if (next !== null) {
-      this.setState({ selectedRow: next })
-    }
-
-    if (focus) {
-      this.list.focus()
+      this.setState({ selectedRow: next }, () => {
+        if (focus && this.list != null) {
+          this.list.focus()
+        }
+      })
     }
   }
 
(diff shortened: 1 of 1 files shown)


app/src/ui/lib/filter-list.tsx
```diff
@@ -244,19 +244,25 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
   }
 
   public selectFirstItem(focus: boolean = false) {
-    if (this.list !== null) {
-      const next = findNextSelectableRow(this.state.rows.length, {
+    if (this.list == null) {
```
iAmWillShepherd: We can be explicit here since list is of type `List | null`.

app/src/ui/lib/filter-list.tsx
```diff
@@ -244,19 +244,25 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
...
+      },
+      this.canSelectRow
+    )
 
-      if (focus) {
-        this.list.focus()
-      }
+    if (next !== null) {
+      this.setState({ selectedRow: next }, () => {
+        if (focus && this.list != null) {
```
iAmWillShepherd: Same thing here.

shiftkey:
:apple:

Issue #2637, opened 2017-09-07
better UX in Merge Branch dialog when there are no other branches


tierninho:
Should we be displaying the Branch > Merge into Current Branch.. option if `master` is the only branch. 

Suggest we grey it out until minimum of two branches ... thoughts?

0.8.1-beta3, Mac

<img width="862" alt="screen shot 2017-09-07 at 9 26 22 am" src="https://user-images.githubusercontent.com/14828183/30156310-daeecdec-93ae-11e7-8972-bc2eb47267e9.png">

joshaber:
My initial thought is that letting the user open the dialog and see they have no other branches is clearer than disabling the menu item.

@desktop/design what do you think?

niik:
> @desktop/design what do you think?

My gut feeling concurs, we shouldn't disable the menu item. That said the merge branch dialog with no other branches than the current isn't great either. I think we could add some UX around that case to make it :sparkles: but unfortunately that'll have to be fairly low down on our list of priorities right now.

PR #4726, opened 2018-05-17
Disable Merge into current branch menu item when there aren't enough branches

### LINKED ISSUE #4725: Merge into current branch menu item is not disabled at the right times
This menu item should be disabled if there are no branches to merge.

screen shot 2018-05-17 at 3 13 36 pm

iAmWillShepherd:
Fixes #4725 

## Before
**New repo**
screen shot 2018-05-17 at 3 27 00 pm

## After
**New repo**
after-no-branches

**New repo after created new branch**
after-created-new-branch

**Existing repo (that already has branches)**
after-with-branches

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -150,6 +150,7 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
   let networkActionInProgress = false
   let tipStateIsUnknown = false
   let branchIsUnborn = false
+  let hasMultipleBranches = false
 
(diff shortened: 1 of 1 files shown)


shiftkey:
I think this change could be confusing, and the discussion in #2637 covers other ways to improve the UX.

iAmWillShepherd:
I see. I know I may not represent the majority, but I find it more confusing to have to click on the menu item just to find out you can't do anything. I'd prefer to know _before_ I click if I can carry out that action. I will close however since I don't have strong feelings on the issue.

PR #4743, opened 2018-05-21
ensure menu actions put compare tab in correct state

### LINKED ISSUE #4696: Compare branches - Cannot select another branch if branch field is already filled
Occurs in Version 1.1.2-beta5 (Windows 10)
Steps to reproduce:
1) Compare a branch typing in the branch field and select one from the list.
2) Type in the field without pressing (X) to clear the field.
3) Pick a different branch than the previous one from the list.
Result: Selection not applied and comparison stays in the previous branch.

PS: If you clear the field pressing the (X) everything goes right.

### LINKED ISSUE #4733: Can't switch between history + compare mode from menu item in beta
## Description
Found two bugs but they're probably related. Basically if my history tab is open, I can't use the file menu items to switch between compare mode and full history mode.

## Version
* GitHub Desktop: 1.1.2-beta7 (latest `beta`)
* Operating system: macOS High Sierra - 10.13.3

## Steps to Reproduce
*** there's two so make sure to scroll beyond the first gif

If (1) I am on the history tab + viewing the full history and (2) I go to `Branch` > `Compare to Branch`
 * I expect it to: focus the input to `Select branch to compare` and show me the branch options to select
 * but instead it: stays in full history mode (see gif below)

compare-to-branch-switch

If (1) I am on the history tab + in compare mode and (2) I go to `View` > `Show History`
 * I expect it to: show me the full history
 * but instead it: stays in compare mode (see gif below)

show-history-switch

shiftkey:
Fixes #4733 
Fixes #4696 

 - [x] put form into right state when switching to `Show History`
 - [x] put form into right state when switching to `Compare to Branch`
 - [x] only update form state when the compare branch has changed, to prevent reverting to previous state
 - [x] one more fiddly state transition - compare branch set, focus on branch switcher, view history to reset -> close branch switcher too


nerdneha:
looks like it works except for the edge case where the focus is on the input `Select Branch to Compare...` and the branch isn't selected

Conditions:
(1) you focus on the input `Select Branch to Compare...` and do NOT select a branch (which means you can either focus and do nothing, focus and type some letters, or focus and click somewhere else in the tab panel)
and
(2) try to `Show History` (either via menu item or by shortcut`)

I expect it to switch to the history right away, but
instead it stays in compare mode. once you click anywhere else outside of the panel, it'll switch back to full history (actually anytime you focus on the input and do not select a branch and THEN click outside of the pane, it'll switch back to full history)

test-history-compare-switch

shiftkey:
@nerdneha yeah, that one is annoying because we've currently got two ways to track whether to show the branch list - originally it was state inside the compare tab, but we've now needed to specify this via props so that menus can hint what the new state should be - and this is one case where it's not clear how to make the prop change take priority over the current state.

I think we need to lift the state up into app state (and maybe the filter text too?), but I don't want that refactor to break things with the imminent ship. :thumbsup: to extracting a new issue for 1.2.1.

app/src/ui/history/compare.tsx
```diff
@@ -78,32 +78,62 @@ export class CompareSidebar extends React.Component<
...
+      // reset form to it's default state
+      this.setState(
+        {
+          filterText: '',
+          focusedBranch: null,
+          showBranchList: nextProps.shouldShowBranchesList,
+        },
+        () => {
+          // ensure filter text behaviour matches the prop value
+          if (this.textbox) {
```
nerdneha:
```js
if (this.textbox !== null) {
```

app/src/ui/history/compare.tsx
```diff
@@ -78,32 +78,62 @@ export class CompareSidebar extends React.Component<
...
+          focusedBranch: newBranch,
+        })
+      }
+    }
 
-      this.setState({
-        filterText: branch.name,
-        focusedBranch: branch,
-      })
+    if (
```
nerdneha: the only way this happens is if `this.props.shouldShowBranchesList` is `false` and `nextProps.shouldShowBranchesList` is `true` right? maybe this can get folded in with what's on line 139? seems like we're just `setState`-ing to be `nextProps.shouldShowBranchesList` if it's different from `this.props.shouldShowBranchesList`

is there a case I'm missing?

shiftkey: This is kind of related to the issue reported https://github.com/desktop/desktop/pull/4743#issuecomment-390822487, and it's related to managing both prop changes and state. This check is helpful for the `Compare to Branch` menu item, but the equivalent checks for `Show History` (`true` -> `false` isn't there).

> maybe this can get folded in with what's on line 139?

I'd rather lift up this state and then figure out what to do with the `sidebarHasFocusWithin` check as part of that work (I think it can go away too).

PR #4747, opened 2018-05-21
The only constant is change

donokuda:
This pull request changes the tab label back from "Commit" to "Changes" as well as the menu item (which before was "Create Commit").

**Before**

Tab bar

Menu item

**After**

Tab bar:

Menu item:


app/src/ui/app.tsx
```diff
@@ -242,7 +242,7 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.push()
       case 'pull':
         return this.pull()
-      case 'create-commit':
+      case 'show-changes':
         return this.createCommit()
```
nerdneha: @donokuda can we change the createCommit() methods to be called showChanges() also?

donokuda: oh good catch. Done in 12424a8

app/src/ui/changes/commit-message.tsx
```diff
@@ -229,7 +229,7 @@ export class CommitMessage extends React.Component<
   }
 
   private onSubmit = () => {
-    this.createCommit()
+    this.showChanges()
```
nerdneha: i think this one was supposed to stay as createCommit since it's the submit button to create the commit (in fairness it's confusing)

donokuda: 😅 

...

Fixed in f5a10ca 🙇

PR #4748, opened 2018-05-21
1.2.0 changelog

shiftkey:
This is the rollup of the changelog entries from the previous betas - I'm going to give the words a pass and ensure they actually make sense.


shiftkey:
@nerdneha I took that tweaked rule a bit further :metal:

PR #4759, opened 2018-05-22
Bump version for next beta

iAmWillShepherd:
Moves beta to `v.1.2.1-beta1`


app/package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.2.0",
+  "version": "1.2.1-beta1",
```
shiftkey: This should be in sync with whatever is the next version in the changelog

PR #4760, opened 2018-05-22
Document when to use beta0

iAmWillShepherd:
Learned today that we should use `*-beta0` for releases that have no changes. This is only used after releasing a new version of the app.


nerdneha:
woops you might not be ready yet, sorry @iAmWillShepherd! if you push it'll probably dismiss my review

iAmWillShepherd:
No problem, I was just going to come back to make sure I didn't misspell something. I just wanted to get the PR up while it was on my mind.

PR #3101, opened 2017-10-19
Syntax highlighted diffs

niik:
Wouldn't it be nice with some syntax highlighting in your diffs?

A screenshot of GitHub Desktop showing a diff with syntax highlighting

### The Problem

Syntax highlighting is a well-understood problem with tons of options. Atom uses TextMate grammars to do theirs but since we're already using CodeMirror I took a stab at implementing ours using that.

Syntax highlighted diffs have been a much appreciated feature of GitHub.com for a long time now and one that I have missed in GitHub Desktop for a long time. Highlighting in diffs presents some added complexity over that of highlighting in a normal source file though. Pretty much all languages are contextual, in that what happened on some line "higher up" affects what's going on further down. As such you can't just pull out a line from a diff and expect it to be highlighted properly. Here's a good example

A screenshot of GitHub Desktop showing a diff with a multi-line comment which is missing the opening statement

Had we just tried to highlight individual lines here we wouldn't have been able to infer that the first line was part of a multi-line comment.

Instead, we have to take the contents of the file before the change, and the contents of it after and run highlighting on both versions. Once that's done we can stitch these together to form one syntax highlighted diff.

### The Approach

When we are about to perform highlighting on a diff we start out by scanning through the diff to figure out which lines we need from which file. Context lines can be pulled from either version while added/removed lines obviously need to come from a particular version. If we find that a file consists entirely of additions or entirely of deletions we can optimize further by adding a preference for one of the versions and thus getting away with loading just one file.

Once we've got that settled we load the first 256kb from both versions (256kb picked arbitrarily because I figured it should cover the majority of source files while adding a very manageable memory overhead for the feature). We then pass this content, along with which lines we want to get tokens for to one or two web workers which then run the modes. CodeMirror modes are synchronous but running them in a web worker means we can get on with other things while we're tokenizing up to half a megabyte of content in up to two different threads (threads in javascript, what have we come to). It also means that we have a real nice containment of the highlighting process and that we can terminate it should it for some reason end up taking a very long time to complete.

When we get the results from the workers we apply our own custom CodeMirror mode which takes the tokens from the language modes and applies them inside of our diff. That means that there's a small window in between when users see the diff and when highlighting gets applied. In my testing it's barely noticeable and it means we can deliver what really matters (the diff) as quickly as we've done before.

### The Performance.

[...]

PR #4764, opened 2018-05-23
Dynamic importing of CodeMirror modes at runtime

niik:
When we initially shipped [syntax highlighted diffs](https://github.com/desktop/desktop/pull/3101) we supported 18 languages and formats. Since then and thanks to community contributions the list has grown to 36 with more on the way.

As the list of supported languages has grown so has the bundle size of the highlighter module and it sits at about 0.5Mb (unminimized) at the moment. Since that file has to be loaded, parsed, and initialized for each highlighter web worker we spin up (in normal usage 2 for the life time of the application) this creates unnecessary overhead when most users rarely combine more than a few languages in their day to day work.

This PR leverages chunk splitting in webpack 4 to split out each language (or technically each CodeMirror mode) into its own, separate javascript file that then gets loaded, at runtime, when that particular language mode is needed through the `importScript` web worker call.

This ultimately leads to less code being loaded into the runtime and faster initial highlighting. Instead of loading 0.5Mb we're now loading the bare minimum (~16kb) plus whatever mode is required.

### Highlighter bundle before

screen shot 2018-05-23 at 10 54 42

### Highlighter bundle after

screen shot 2018-05-23 at 10 45 44

### Performance

Essentially we sacrifice a little bit (in absolute numbers) of performance whenever a new language is loaded in exchaneg for a big gain in initial load time and zero overhead of adding new language modes in the future. These timings are all taken in dev mode on my early 2015 macbook 12"

| Scenario        | Monolith | Split  | Delta (lower is better) |
|:------------- |-------------:| -----:|-----:|
| First highlighting of language 1     | 261ms | 137ms | 52% |
| First highlighting of language 2    | 7ms      |  23ms | 328% |
| Second highlighting of language 1    | 6ms      |  4ms | No significant change |


app/src/highlighter/index.ts
```diff
@@ -11,115 +11,243 @@ import {
...
+ * the require property).
+ */
+interface IModeDefinition {
+  /**
+   * A function that, when called, will attempt to asynchronously
+   * load the required modules for a particular mode. This function
+   * is idempotent and can be called multiple times with no adverse
+   * effect.
+   */
+  readonly require: () => Promise<void>
```
j-f1: How do you feel about calling this `import` or `install` to encourage people.m not to use `require`?

agisilaos:
@niik As I can understand this will change the way for someone to add support for a new programming language? So the docs need to be updated as well in this PR? 🤔

app/src/highlighter/index.ts
```diff
@@ -162,6 +292,14 @@ function detectMode(request: IHighlightRequest): CodeMirror.Mode<{}> | null {
     return null
   }
 
+  const modeDefinition = mimeModeMap.get(mimeType)
+
+  if (!modeDefinition) {
```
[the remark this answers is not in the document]
j-f1: Looks like it should be `modeDefinition == null`.

app/src/highlighter/index.ts
j-f1: I’m kinda liking `install` since the return value is ignored and the function has a side effect, which is what installation typically means.

app/src/highlighter/index.ts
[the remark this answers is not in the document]
shiftkey: Oops, yes, that. Just less truthiness plz.

say25:
@niik and @shiftkey any status on this. There are a couple extensions I'd like to add without introducing merge conflicts and what not.

shiftkey:
@say25 I probably won't get a chance to start reviewing it until next week

app/src/highlighter/index.ts
niik: I have no feelings whatsoever on what we name this but `import` is a reserved keyword which is why I didn't go with that to begin with. So `install` it is.

niik:
> @niik As I can understand this will change the way for someone to add support for a new programming language? So the docs need to be updated as well in this PR? 🤔

I don't *believe* we have any docs that goes into the specifics of how to add a new language. I assume that contributors who've added a theme have probably just copied and pasted a previous mode that was already in the `highlighter/index.tsx` and tweaked it. If someone would like to contribute more docs around this I'd be 😍 but I'm gonna say that it's out of scope for this particular PR.

niik:
Ready for another look :bow:

saul:
@niik the documentation is here: https://github.com/desktop/desktop/blob/master/docs/technical/syntax-highlighting.md#i-want-to-add-my-favorite-language

Issue #1128, opened 2017-04-04
Option to disable auto-fetching (background fetching) in Desktop


cheshire137:
I would love the option to be able to turn off the app fetching the latest changes for me, or really any automatic `git` command done on my behalf. My main use case for the app is to have a pretty diff of the changes I'm about to commit. I do much of my other work, like resolving conflicts and updating a branch with the latest `master`, via the terminal. I often hit errors when I do a flow like `git pull; git checkout master; git pull; git checkout -; git merge master` that I think are caused by the app trying to run `git` commands at the same time I am:

```
Already up-to-date.
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
error: cannot lock ref 'refs/remotes/origin/acceptance-saml': is at 00617906cd4ddf0ffdf2391f4c77187406a9b38a but expected 181445860d11aaeb13b0a281d42e59610e6499e2
From https://github.com/github/github
 ! 181445860d1..00617906cd4  acceptance-saml    -> origin/acceptance-saml  (unable to update local ref)
error: cannot lock ref 'refs/remotes/origin/lw-manage-listings': is at bde39ce0be0b9b4049a575892d705e440744068e but expected 38318bd501856d9f15119d7f6ef77f9a54427e89
 ! 38318bd5018..bde39ce0be0  lw-manage-listings -> origin/lw-manage-listings  (unable to update local ref)
```

I can get around these errors by rerunning my command, but it's frustrating when they occur because it feels like I'm wrestling with the GitHub app being open. I'd like to be able to leave the app open since I commit frequently and love the diff view, but I'd also like to be in control of when it runs `git` commands.

alrz:
>Allow disabling auto-fetching

Please! This is a show stopper for me, I have a lot of remotes and GD tries to fetch them all at startup which cripples my internet connection. If there is an option, it should also consider auto-fetch only the current branch.

shiftkey:
>  I have a lot of remotes and GD tries to fetch them all at startup which cripples my internet connection.

We'll push out https://github.com/desktop/desktop/pull/1435 soon which will stop this from occurring.

> And when I do choose to update all remotes it'd be nice if it run in parallel -- `git remote update --prune` does it sequentially.

We're tracking #750 to do something similar to this - please comment there about that request.

IdanAdar:
I think an option to separate fetching and pulling would be great actually.

Fetch - show the changes that can be pulled from the branch. A preview of sorts
[...]

PR #4770, opened 2018-05-23
Show more info in repositories list

### LINKED ISSUE #2259: Repository list information
This is a rollup issue for showing more information in the repository list, [currently on the roadmap](https://github.com/desktop/desktop/blob/df3c0a089ecacea86462fb55244a108d0a419a53/docs/process/roadmap.md#13) for 1.3.

vanessayuenn:
kapture 2018-05-23 at 18 33 38

Hi all! This PR adds some indicators on the repositories list view to provide more context & information (whether there are uncommited changes, and/or unsynced upstream/local commits) about the respective repo. This is in response to the issues raised by various users as summarized in the roll-up issue (resolves #2259).

All & any feedback is welcome, but there are a few things I specifically would like to hear your thoughts on:

1. Right now, the repo info gets fetched in the background every 5 minutes, which is entirely arbituary. Any concerns about this frequency?
2. I was told it's on the roadmap that the blue dot indicator at the commit tab will go away, which will then make the blue dot in the repo list non-reflective of the rest of the UI. (cc @donokuda 👀)
3. I like @shiftkey's idea about using tooltip to complement this feature, so I tried it out. But as you can see from the screenshot below, it gets quite crowded and so I'm unsure about it. 

screen shot 2018-05-23 at 18 37 26


app/src/ui/app.tsx
```diff
@@ -147,6 +147,10 @@ export class App extends React.Component<IAppProps, IAppState> {
         },
         { timeout: ReadyDelay }
       )
+      const interval = 1000 * 60 * 5
```
shiftkey: :art: let's extract this up to a const with a more descriptive name.

For example, we have this defined earlier in the file:


```
/** The interval at which we should check for updates. */
const UpdateCheckInterval = 1000 * 60 * 60 * 4
```

donokuda:
🙌:sparkles: Really liking how this came out! 

> I was told it's on the roadmap that the blue dot indicator at the commit tab will go away, which will then make the blue dot in the repo list non-reflective of the rest of the UI. (cc @donokuda 👀)

I'm not too concerned about this. For what it's worth, the blue dot is ambiguous enough that it could mean almost anything; and in fact, I'd love to hear your thoughts on having a tooltip for the blue dot that says something along the lines of "There are uncommitted changes in this repository"

> I like @shiftkey's idea about using tooltip to complement this feature, so I tried it out. But as you can see from the screenshot below, it gets quite crowded and so I'm unsure about it.

🙈Yup, that looks pretty crowded! I wonder we can condense the text down to something like "The currently checked out branch is commits behind [and/or ahead] of its tracked branch"

----

I chatted with @ampinsk and she thought of a couple of ways to improve the notification dot (which I pushed to a branch named [`more-repo-info-ui-tweaks`](https://github.com/desktop/desktop/tree/more-repo-info-ui-tweaks)). I ended up settling on this in case you wanted to bring those changes into your branch. I would love to hear your thoughts on this:

image

Here's what changed:

- **Moved the notification dot to the left of the repo icon / name.** This makes it easier to scan for the blue notification dot (it previously felt like it was getting lost in the repo octicon)
- **Lightened the behind/ahead indicator.** The indicator seemed to be visually heavy that my eyes wanted to gravitate towards the right side of the panel. Making the indicator a little lighter and smaller helped alleviate that.

There's a few styling issues happening (specifically when the [background color of the list item is blue](https://user-images.githubusercontent.com/1174461/40456584-026e63d2-5ea7-11e8-8ab1-ab1ff0cbc510.png)), but I can address those myself and wouldn't consider those blockers to this PR

app/src/lib/stores/app-store.ts
```diff
@@ -1751,6 +1753,35 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    this.repositories = await this.getRefreshedRepositories(this.repositories)
+    this.emitUpdate()
+  }
+
+  private async getRefreshedRepositories(
+    repositories: ReadonlyArray<Repository>,
+    targetRepo?: Repository
+  ): Promise<ReadonlyArray<Repository>> {
+    for (const repo of repositories) {
+      if (!targetRepo || targetRepo.id === repo.id) {
```
shiftkey: 🤔I can see two related use cases here - refreshing the local state of all repositories, or just refreshing one. Rather than doing a check like this, could we break this out into two functions?

```ts
  private async getRefreshedRepositories(
    repositories: ReadonlyArray<Repository>
  ): Promise<ReadonlyArray<Repository>> {
      this.refreshRepositoryStatus(repo)
    }
    return repositories
  }

  private async refreshRepositoryStatus(
    repository: Repository
  ): Promise<Repository> {
    // do the actual refresh work here
  }
```

I think this also requires me to wrap my head around how we're currently managing `this.repositories` because we now have three places where we're updating it.

For existing changes to a `Repository` model we use the `RepositoriesStore` to update the database, like this:

https://github.com/desktop/desktop/blob/12af6e4bee8b398ea957ac7f51fdb07cc234946d/app/src/lib/stores/repositories-store.ts#L170-L173

This function then emits an update and the `AppStore` then re-reads what's in the database by invoking `getAll` this through the callback defined in `this.repositoriesStore.onDidUpdate`. I like this pattern because all changes flow the same way, but it doesn't feel like a natural fit for you because these changes don't need to be stored in the database.

app/src/lib/stores/app-store.ts
```diff
@@ -1751,6 +1753,35 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        await this.withAuthenticatingUser(repo, async (repo, account) => {
+          const gitStore = this.getGitStore(repo)
+          repo.setAheadBehind(gitStore.aheadBehind)
```
shiftkey: :art: the backing fields for `setAheadBehind` and `setChangedFiles` are currently public, so they can be set directly without the need for also having a function defined:

```ts
repo.aheadBehind = gitStore.aheadBehind
```

Alternatively, if you wanted to protect access to these you can use getters and setters. This is a TS example of the syntax:

```ts
class Employee {
    private _fullName: string;

    get fullName(): string {
        return this._fullName;
    }

    set fullName(newName: string) {
        if (passcode && passcode == "secret passcode") {
            this._fullName = newName;
        }
        else {
            console.log("Error: Unauthorized update of employee!");
        }
    }
}
```

app/src/lib/stores/app-store.ts
```diff
@@ -339,8 +339,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.accountsStore.onDidError(error => this.emitError(error))
 
     this.repositoriesStore.onDidUpdate(async () => {
-      const repositories = await this.repositoriesStore.getAll()
-      this.repositories = repositories
+      this.repositories = await this.repositoriesStore
+        .getAll()
+        .then(repos => this.getRefreshedRepositories(repos))
```
iAmWillShepherd: Just curious about the decision to mix the `Promise.then` API with `await`. It should be possible to do the following since calling `await` on a Promise resolves it.

```ts
...
const repositories = await this.repositoriesStore.getAll()
this.getRefreshedRepositories(repositories)
this.repositories = repositories
...
```

app/src/lib/stores/app-store.ts
```diff
@@ -1751,6 +1753,35 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    targetRepo?: Repository
+  ): Promise<ReadonlyArray<Repository>> {
+    for (const repo of repositories) {
+      if (!targetRepo || targetRepo.id === repo.id) {
+        await this.withAuthenticatingUser(repo, async (repo, account) => {
+          const gitStore = this.getGitStore(repo)
+          repo.setAheadBehind(gitStore.aheadBehind)
+          await gitStore.fetch(account, true)
+          const status = await gitStore.loadStatus()
+          if (status) {
```
iAmWillShepherd: We try to be explicit with `null` and `undefined` checks for readability.

`status !== null`

app/src/lib/stores/app-store.ts
vanessayuenn: Hrmm perhaps I am not understanding Typescript's `ReadonlyArray` properly -- but I thought `this.repositories` being `readonly` means we cannot selectively update only one of the repositories, right? We will have to iterate through all of them to find that one and reassign the entire array anyway, hence my decision to add a `targetRepo` for when we only want to refresh one in hope for _some_ performance gain. 

I originally wanted to use `RepositoriesStore` to do all the updating as well - it just looked nicer to me, but yes, I ended up deciding against it for the reasons you cited. :)

app/src/lib/stores/app-store.ts
vanessayuenn: To me, it's more clear this way that we are waiting for the whole chain of actions to complete before assigning the results to `this.repositories`. But I don't feel strongly about this at all, so I am ok with changing this for consistency.

vanessayuenn: fixed!

vanessayuenn:
@donokuda @ampinsk I love the tweaks you made! Pulled them into this working branch already. :) 
>I'd love to hear your thoughts on having a tooltip for the blue dot that says something along the lines of "There are uncommitted changes in this repository"

>I wonder we can condense the text down to something like "The currently checked out branch is commits behind [and/or ahead] of its tracked branch"

Instead of a catch-all tooltip, I experimented with splitting it up into three tooltips: 1 for the blue dot, 1 for repo name (what we currently have), and 1 for the arrows. Hopefully that will be less crowded and make it more clear what each indicator means. What do you think?

screen shot 2018-05-25 at 16 03 49
screen shot 2018-05-25 at 16 04 03
screen shot 2018-05-25 at 16 03 58

app/src/lib/stores/app-store.ts
shiftkey: > Hrmm perhaps I am not understanding Typescript's `ReadonlyArray` properly -- but I thought `this.repositories` being `readonly` means we cannot selectively update only one of the repositories, right?

There's a couple of different uses for `readonly` in here, so I'll mention them here and then try and clarify my thoughts further:

 - `ReadonlyArray` is a way to pass around an array of values that won't be mutated by callers
 - the `readonly` keyword in TypeScript is a way to mark properties in a class which shouldn't be reassigned outside the constructor
 - `this.repositories` isn't defined as `readonly` inside `AppStore` so we can reassign it when the `RepositoriesStore` updates:

```ts
  private repositories: ReadonlyArray<Repository> = new Array<Repository>()
```

So while we're using a `ReadonlyArray<Repository>` in `getRefreshedRepositories` to indicate we're not modifying the array, we are mutating each repository in the array to update it's values and then returning the same array, which is then reassigned to `this.repositories`.

shiftkey:
@vanessayuenn we were chatting about this PR yesterday in Slack but here's the two other things to think about aside from the shape of the app state that we've been wrestling with:

 - feature-flagging - are we able to control how we roll this feature out gradually to users, just in case we encounter some unexpected issues in the wild?
 - performance/load testing - currently we're doing a fetch for each repository in the list every five minutes. We already have a request to disable background fetching in #1128, so I'm curious what the risks are of introducing more background work like this, and how we might manage this for users who are tracking a significant number of repositories that are active (meaning `fetch` is often necessary to keep up).

app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -49,6 +56,33 @@ export class RepositoryListItem extends React.Component<
     const path = repository.path
     const gitHubRepo =
       repository instanceof Repository ? repository.gitHubRepository : null
+    let hasChanges = this.props.changedFilesCount > 0
```
shiftkey: This can now be `const`-d

app/src/lib/feature-flag.ts
```diff
@@ -41,3 +41,8 @@ export function enableCompareSidebar(): boolean {
 export function enableNotificationOfBranchUpdates(): boolean {
   return enableDevelopmentFeatures()
 }
+
+/** Should the repository list display info indicators? */
+export function enableRepoInfoIndicators(): Boolean {
```
shiftkey: :art: `boolean` here to be consistent with our other declarations

app/src/lib/stores/app-store.ts
```diff
@@ -1771,6 +1773,40 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+  public async refreshAllRepositories() {
+    this.repositories = await this.getRefreshedRepositories(this.repositories)
+    this.emitUpdate()
+  }
+
+  private async getRefreshedRepositories(
+    repositories: ReadonlyArray<Repository>,
+    targetRepo?: Repository
+  ): Promise<ReadonlyArray<Repository>> {
```
shiftkey: I think we can make this a `Promise<void>`-returning function and update the name now that we're updating a cache based on the repository ID, which will eliminate a bunch of the `this.repositories` assignment.

shiftkey: Also, for functions that are `Promise<void>`-returning we try and structure them so they manage their own `this.emitUpdate()` responsibility. An example of that is `updateRepositorySelectionAfterRepositoriesChanged()` and the check at the end:

```ts
    if (repositoryChanged) {
      this._selectRepository(newSelectedRepository)
      this.emitUpdate()
    }
```

It'd be nice to have this function behave in the same way, so it only signals when it's cache has been added or updated.

app/src/lib/stores/app-store.ts
```diff
@@ -1771,6 +1773,40 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    if (enableRepoInfoIndicators()) {
+      for (const repo of repositories) {
```
shiftkey: :art: you can use `this.repositories` here and avoid passing in the parameter, now that we're no longer mutating the objects we get from the repositories store.

app/src/lib/stores/app-store.ts
```diff
@@ -1780,6 +1780,40 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this._updateCurrentPullRequest(repository)
     this.updateMenuItemLabels(repository)
     this._initializeCompare(repository)
+    this.refreshRepositoryState([repository])
```
shiftkey: @vanessayuenn I've clarified the signature here to help make it's intent clearer, hopefully this makes our discussions about performance better. I'm still not sold on the `refreshRepositoryState` name as it feels too generic, maybe `refreshRepositoryContext` here but even that doesn't feel clearer...

app/src/lib/stores/app-store.ts
```diff
@@ -1780,6 +1780,40 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    }
+
+    const key = `refreshing ${repositories.length} repositories`
+
+    console.time(key)
+    for (const repo of repositories) {
+      await this.withAuthenticatingUser(repo, async (repo, account) => {
+        const gitStore = this.getGitStore(repo)
+        const lookup = this.localRepositoryStateLookup
+        await gitStore.fetch(account, true)
```
shiftkey: @vanessayuenn and I talked a bunch about this part of the background work earlier - we're trying to balance two schools of thought here:

 - background fetching more frequently means that users can get feedback sooner
 - repositories with significant activity here means fetches will take longer (we have #1128 where users want to disable background fetching already)

I'm going to come back with some data to help quantify where we should end up on this spectrum

shiftkey: So we currently have fetching enabled for every refresh, and using the Desktop repository as a reference this is the overhead:

 - macOS: `refreshing 1 repositories: 463.816162109375ms`
 - Windows: `refreshing 1 repositories: 1064ms`

This occurs whenever you switch back to Desktop from another application, or it can occur when switching repositories.

Every five minutes, the `setInterval` callback fires and all current repositories are refreshed (and fetched):

 - macOS: `refreshing 12 repositories: 5383.87890625ms`
 - Windows: `refreshing 9 repositories: 3438ms`

There are some complex repositories in the macOS list which is why it's now significantly worse, but the constraint still seems to be the `git fetch` operations themselves.

shiftkey: Also note that I haven't factored in my internet connection speed (which is very good by world standards) - slower internet connections will likely take longer when fetching. This is offset by the fact this is being done in the background, but still a thing to keep in mind.

vanessayuenn: >This occurs whenever you switch back to Desktop from another application, or it can occur when switching repositories.

@shiftkey I added a bit that skips `fetch` if the heavier `refreshRepository` operation has just been run less than 2 mins ago, which it should have everytime you switch repo and on application coming to focus.

vanessayuenn:
As far as things stand with this PR, is perf the only showstopper? If so, considering this is currently behind the `PREVIEW_FEATURES` feature flag and won't even make it into beta, I propose addressing performance concerns in another PR/issue. Another reason is that perf optimization can be an endless battle if not scoped out properly, and I think we should be able to quantify the goals (i.e. what numbers would be good enough for beta? and for prod?) before diving into it. Thoughts on this?

shiftkey:
I'm going to drop myself from reviewing because I've helped contribute to this PR - as well as some of the underlying refactoring to support this - but it'd be nice if @niik or @iAmWillShepherd could give some :eyes: over how they feel this feature is and what it would take to land this behind a feature flag so we can follow up with the performance side...

app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -52,7 +59,34 @@ export class RepositoryListItem extends React.Component<
...
+        return null
+      }
+      const { ahead, behind } = this.props.aheadBehind
+      if (ahead === 0 && behind === 0) {
+        return null
+      }
+      const commitGrammar = (commitNum: number) =>
+        `${commitNum} commit${commitNum > 1 ? 's' : ''}` // english is hard
+      const aheadBehindTooltip =
+        'The currently checked out branch is' +
```
iAmWillShepherd: Looks like we have to handle the case that `behind` is 0, as it stands, this would output _The currently checkout branch is its tracked branch._

vanessayuenn: If `ahead` and `behind` are both 0, the `if` block right about this would have handled it and returned already.

app/src/lib/stores/app-store.ts
```diff
@@ -1852,6 +1852,38 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+  private async refreshRepositoryState(
+    repositories: ReadonlyArray<Repository>
+  ): Promise<void> {
+    if (!enableRepoInfoIndicators()) {
+      return
+    }
+
+    for (const repo of repositories) {
+      await this.withAuthenticatingUser(repo, async (repo, account) => {
```
iAmWillShepherd: More of a question than a suggestion. Could we do this in parallel? It looks like nothing is modifying any share state except for L#1878.

vanessayuenn: I am not sure I follow -- could you please elaborate on how you'd do this in parallel?

app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -52,7 +59,34 @@ export class RepositoryListItem extends React.Component<
...
     const gitHubRepo =
       repository instanceof Repository ? repository.gitHubRepository : null
-    const tooltip = gitHubRepo
+    const hasChanges = this.props.changedFilesCount > 0
+    const renderAheadBehindIndicator = () => {
+      if (!(repository instanceof Repository) || !this.props.aheadBehind) {
+        return null
+      }
+      const { ahead, behind } = this.props.aheadBehind
+      if (ahead === 0 && behind === 0) {
```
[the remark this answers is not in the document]
vanessayuenn: This part renders the little arrows indicators. So if the repo is neither ahead nor behind its remote, we just leave it blank.

app/src/lib/stats/stats-store.ts
```diff
@@ -355,6 +357,18 @@ export class StatsStore {
     }))
   }
 
+  public recordRepoClicked(repoHasIndicator: boolean): Promise<void> {
```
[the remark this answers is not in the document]
vanessayuenn: Introduced a variable to encompass that expression as per [your suggestion here](https://github.com/desktop/desktop/pull/4770#discussion_r197866420), so this shouldn't be an issue anymore :)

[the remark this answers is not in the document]
j-f1: Alternatively, `recordRepoClicked({ indicator: true })`

app/src/lib/stores/app-store.ts
vanessayuenn: Ah I now see what you mean - I like that!

shiftkey: @vanessayuenn @iAmWillShepherd I don't have strong feelings on either approach, but I felt serially executing these refresh operations felt more suited to this background work. Just needed to ensure you `await`-ed the `this.withAuthenticatingUser` to not create fire-and-forget promises.

app/src/ui/app.tsx
```diff
@@ -149,6 +151,10 @@ export class App extends React.Component<IAppProps, IAppState> {
         },
         { timeout: ReadyDelay }
       )
+
+      window.setInterval(() => {
```
iAmWillShepherd: Is there any need to clean up this resource or can the GC infer when a timer is no longer needed?

app/src/lib/stats/stats-store.ts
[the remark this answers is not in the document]
iAmWillShepherd: @vanessayuenn not sure I understand. Was this comment meant for another line? This line was about introducing an enum to encode what the argument means.

app/src/ui/app.tsx
shiftkey: The handle is tied to the lifecycle of `App`, so unless we were somehow recreating this interval and needed to `clearInterval` the previous value I wouldn't worry about it.

j-f1: This will keep running forever, and its existence will stop the instance from getting garbage collected. I’d recommend cancelling it when the component unmounts.

app/src/lib/stats/stats-store.ts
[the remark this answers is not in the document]
vanessayuenn: @iAmWillShepherd Using an enum in place of a boolean seems like an overkill to me, and the `hasXYZ` naming convention should make it self explanatory that either true/false is expected as an argument. Sorry I wasn't clear the first time around.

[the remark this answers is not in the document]
iAmWillShepherd: You're right, enum's may be overkill, but I figured it was an easier solution that the alternative which is to have 2 functions. Flag paremeters in functions is actually a code smell (here are some articles on the topic:  [article 1](https://www.martinfowler.com/bliki/FlagArgument.html) [article 2](https://8thlight.com/blog/dariusz-pasciak/2015/05/28/alternatives-to-boolean-parameters.html))

We actually have examples of this all throughout the code base, but now that we will get to focus on paying down tech debt and improving quality, may as well start now 👍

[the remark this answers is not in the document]
shiftkey: Let's all keep in mind that the very next method here does the exact same sin:

```ts
public async setOptOut(optOut: boolean): Promise<void> {
```

I'm fine with leaving this as-is - if we find ourselves looking for more complex analytics then this will need to be evolved anyway.

PR #4771, opened 2018-05-23
Removes unused feature flag

iAmWillShepherd:
Noticed that we had two feature flags doing the same thing. The one that I removed was not being used anywhere in the app. I also had to do some linter trickery to get it to leave me alone about the `enableBetaFeatures` function not being used. 

I'm open to other ways of telling the linter to be quiet...


j-f1:
Does `// @ts-ignore` work?

PR #4783, opened 2018-05-24
Adds logic that determines what branch to use as the compare branch (NDDB)

### LINKED ISSUE #4766: Infer what branch should be used as the compare branch (NDDB)
In order to notify users that their current branch is behind, we need some way to determine which branch to use as the compare branch.

### Criteria to determine compare branch (in priority order)
1. If the current branch has a PR associated with it, use the target branch of the PR _(this can be a branch that exists on either`upstream` or`origin`)_
1. If the repository is a fork, use the default branch on `upstream` _(this is usually `master`)_ 
1. If the repository is hosted on GitHub, use the  default branch on `origin` _(this is usually `master`)_
1. Otherwise fall through to the local `master` branch (conventionally, the default branch is `master`)

**Notes on criteria**
Criterion 1 and 3 are the most important ones to capture. Most times, the base branches of these two will be equivalent, `origin/master`.

Criteria 2 will likely be the most difficult as their are some edge cases we need to worry about. As I understand it, GitHub users create forks for two reasons:
* To have a starting point for their own ideas
* To propose changes to the repository they are forking from

We need to make sure we are careful to support both of these use-cases. We should be able to handle this by:
1. Checking the default branch on the forked repository (likely `origin/master`)
2. If no changes are found after performing step 1, check the default branch of the parent repository (likely `upstream/master`)

iAmWillShepherd:
Fixes #4766 

This PR encapsulates the logic detailed in the issue inside of a function.
> ### Criteria to determine compare branch (in priority order)
> 1. If the current branch has a PR associated with it, use the target branch of the PR _(this can be a branch that exists on either`upstream` or`origin`)_
> 1. If the repository is a fork, use the default branch on `upstream` _(this is usually `master`)_ 
> 1. If the repository is hosted on GitHub, use the  default branch on `origin` _(this is usually `master`)_
> 1. Otherwise fall through to the local `master` branch (conventionally, the default branch is `master`)

~Some notes up front, I'm not sure of the actual place this functions belongs, so I put it next to related functions. Since it's self contained, it shouldn't be a huge deal to move it to a better place.~

- [x] Add function to encapsulate logic
- [x] Add helper function that can choose the right branch on forked repositories
- [x] Write some tests to exercise ☝️ 
- [x] Clean up


app/src/lib/stores/helpers/infer-compare-to-branch.ts
```diff
@@ -0,0 +1,153 @@
...
+
+interface IOfFork {
+  kind: 'fork'
+  repository: Repository
+  gitHubRepository: GitHubRepository
+  currentBranch: Branch
+}
+
+// I am struggling to come up with a decent name for this type
+type NamingIsHard = IOfPullRequest | IOfGithub | IOfFork
```
iAmWillShepherd: I don't have a good names for this type nor the 3 interfaces that compose it.

I'm using this type to make it easier for the caller to make use of the function, `inferComparisonBranch`. Leaving it as a list of arguments would require users of this function to have to concern themselves with the implementation details of this function, something I'm looking to avoid. I thought of this fairly quickly, so I'm sure this could be done better. Open to suggestion 👍 

Thoughts @desktop/core

app/src/lib/stores/helpers/infer-compare-to-branch.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * Do **not** call this method; use `inferComparisonBranch` instead
+ *
+ * Returns `master`, the default branch of a Git repository
+ *
+ * @param branches The list of all branches for the repository
+ */
+export function _getMasterBranch(
```
iAmWillShepherd: These functions are prefixed with an `_` (underscore) because I intend for them to not be used outside of this module; however, I still want the ability to test them. @niik and I spoke on this and he mentioned we could put them in a separate file, but we agreed that we can cross that bridge at a later point in time.

app/src/lib/stores/helpers/infer-compare-to-branch.ts
j-f1: When would you need to call this function without knowing what type of object you’re passing? Would it make sense to just export the three more specific functions and call them individually?

app/src/lib/stores/helpers/infer-compare-to-branch.ts
```diff
@@ -0,0 +1,153 @@
...
+  if (defaultBranch === null) {
+    // Fall back to master
+    return _getMasterBranch(branches)
+  }
+
+  const range = revRange(currentBranch.tip.sha, defaultBranch.tip.sha)
+
+  // use the mock if it's passed - for testing only
+  const aheadBehind =
+    (mockingAheadBehind && mockingAheadBehind(range)) ||
```
iAmWillShepherd: I need to be able to test this function by controlling what this line returns. Since this function is meant to be private, I felt it would be OK to pass an optional argument, `mockingAheadBehind`, to this function and call it in this way. I'm open to other ideas for mocking the behavior of the `getAheadBehind` function.

app/src/lib/stores/helpers/infer-compare-to-branch.ts
iAmWillShepherd: > When would you need to call this function without knowing what type of object you’re passing? 

The only case you wouldn't need to know what type the object is, is if you have a non-published repository. In all other case, you'd need to know. 

> Would it make sense to just export the three more specific functions and call them individually?

It would. In fact, that's how I originally had it, but doing it this way ensures the caller doesn't have to handle all of that logic themselves. The reason why this is important is because these checks are ordinal. It really becomes a question of _what level do I want to perform these checks?_

As it stands, I think doing it here is the best since it ensures we only have to care about one function that can delegate the function calls without needing to add yet another layer of abstraction.

Do you feel like this function is doing a bit too much or are you thinking having the caller construct one of these types is the wrong approach?

iAmWillShepherd:
I was initially planning on adding test for the `inferComparisonBranch` function, but I'm thinking that is unnecessary since I've tested all of the functions that make it up. I also ran into a blocker when trying to test GitHub repository that is also a fork. I don't plan to add an additional parameter that would allow me to mock the behavior of `getAheadBehind`.

I open to feedback from @desktop/eng on this one.

app/src/lib/stores/helpers/infer-compare-to-branch.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Infers which branch to use as the comparison branch
+ *
+ * @param branches The list of all branches for the repository
+ * @param needsAName The object that determines how the branch will be inferred
+ */
+export async function inferComparisonBranch(
+  branches: ReadonlyArray<Branch>,
+  needsAName: NamingIsHard
+): Promise<Branch | null> {
+  switch (needsAName.kind) {
```
iAmWillShepherd: Hopefully the fact that this function is just a switch statement eliminates the need to write tests to exercise it. I have a few stashed away, but they are literally duplicates of the test cases for each of the _private_ functions.

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Do **not** call this method; use `inferComparisonBranch` instead
+ *
+ * Returns the default branch of the fork if it's ahead of `currentBranch`.
+ * Otherwise, the default branch of the parent is returned.
+ *
+ * @param branches The list of all branches for the repository
+ * @param repository The repository the branch belongs to
+ * @param ghRepository
+ * @param currentBranch The branch we want the parent of
+ * @param getAheadBehind Callback function used to compute ahead/behind
```
nerdneha: seems like this is different from the function params

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,129 @@
...
+  const defaultBranch = _getDeafultBranchOfGithubRepo(branches, ghRepository)
+
+  if (defaultBranch === null) {
+    // Fall back to master
+    return _getMasterBranch(branches)
+  }
+
+  const range = revRange(currentBranch.tip.sha, defaultBranch.tip.sha)
+
+  // use the mock if it's passed - for testing only
```
nerdneha: i'm not sure how i feel about passing in a param just to make tests pass. in cases like this, i've seen people do one of the following:
1. use `sinon` to get mocking functionality
1. pass the `getAheadBehind` function into the function directly so that it's only imported to be used in the public method and it's not directly exposed to the "private" method. this also makes it testable because you can mock what it returns in the tests
1. fully build the test object so that you can call `revRange` from it and call `getAheadBehind` on it
1. call getAheadBehind in the overseeing public method and pass in the result into this method

app/src/lib/stores/helpers/infer-comparison-branch.ts
iAmWillShepherd: If we move away from passing in a mock, the last one will no longer be needed.

nerdneha:
I didn't highlight this before but I think it's fantastic that you're testing this, especially because it's tedious to do in person and it means that, if future bugs are noticed, we can understand why. also, anyone else adding code can understand how to test additional cases! loved the constructor methods you used to make a realistic shell of a branch stack, etc. shows real understanding of the inner workings!!

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,129 @@
+import { Branch, IAheadBehind } from '../../../models/branch'
+import { PullRequest } from '../../../models/pull-request'
+import { getAheadBehind, revRange } from '../../git'
+import { Repository } from '../../../models/repository'
+import { GitHubRepository } from '../../../models/github-repository'
+
+/**
+ * Infers which branch to use as the comparison branch
```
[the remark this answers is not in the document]
iAmWillShepherd: That's something I thought about, but decided against since the inner workings of this function should be encapsulated. My thinking is we shouldn't leak implementation details as that would require the caller to understand how the branch is determined. I was deliberate in including the word `infer` to communicate that. 

Now, I'm not against adding how this is calculated to the function's JSdoc, but I would be interested in knowing any scenarios where having knowledge of this process would be beneficial.

iAmWillShepherd:
> What happens if we branch off of branchA to make branchB? Should we add this as a test case? My original assumption would be that this inferComparisonBranch would return branchA but according this code, I think it'd just return master, right? Is that intended behavior? Maybe I am misunderstanding what "default branch" means.

@nerdneha you are right, this function would just return `master` and that is the intended behavior. If we all agree that it should return branch `A` instead, I'm happy to open another issue for that discussion, but I think what we have so far is a good starting off point.

app/src/lib/stores/helpers/infer-comparison-branch.ts
iAmWillShepherd: I did something similar to item 2. I am still passing the `getAheadBehind` to the private function because I wanted to keep that logic as close to the place it's being used as possible. Hopefully it's a decent compromise.

donokuda:
> If we all agree that it should return branch A instead, I'm happy to open another issue for that discussion, but I think what we have so far is a good starting off point.

I'm okay with punting on the `Branch A` -> `Branch B` scenario for now since I was under the impression that there were some technical aspects to be considered on inferring the parent branch correctly. <sup>[[citation needed tho]()]</sup>

One thing to note is that the "default branch" doesn't necessarily mean "master". For example, the default branch could be "development" or "gh-pages" in some cases. If that's the case, then that should be the branch we compare against for the notification.

I wasn't able to tell by reading the code, but are we supporting this scenario?

iAmWillShepherd:
@donokuda, yes, that case is currently being supported. In situations where a default branch has not been set, we default to `master`.

iAmWillShepherd:
🥇

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,143 @@
...
+ * @param repository The repository the branch belongs to
+ * @param currentPullRequest The pull request to use for finding the branch
+ * @param currentBranch The branch we want the parent of
+ * @param getAheadBehind function used to calculate ahead/behind
+ */
+export async function inferComparisonBranch(
+  branches: ReadonlyArray<Branch>,
+  repository: Repository,
+  currentPullRequest: PullRequest | null,
+  currentBranch: Branch | null,
```
nerdneha: This is for my education only -- can you help me understand what scenario we could be in where currentBranch is null but we still want to infer a comparison branch?

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+function _getDefaultBranchOfGithubRepo(
+  branches: ReadonlyArray<Branch>,
+  ghRepository: GitHubRepository
+): Branch | null {
+  return branches.find(b => b.name === ghRepository.defaultBranch) || null
+}
+
+/**
+ * For `inferComparisonBranch` case where inferring  for a pull request
```
nerdneha: double space ⛏  between `inferring` and `for`

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * For `inferComparisonBranch` case where inferring for a forked repository
+ *
+ * Returns the default branch of the fork if it's ahead of `currentBranch`.
+ * Otherwise, the default branch of the parent is returned.
+ *
+ * @param branches The list of all branches for the repository
+ * @param repository The repository the branch belongs to
+ * @param ghRepository The github specific information for the repository
+ * @param currentBranch The branch we want the parent of
```
nerdneha: i think we missed the
```js
@param getAheadBehind
```

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,143 @@
...
+  ghRepository: GitHubRepository,
+  currentBranch: Branch,
+  getAheadBehind: (
+    repository: Repository,
+    range: string
+  ) => Promise<IAheadBehind | null>
+): Promise<Branch | null> {
+  const defaultBranchName = ghRepository.defaultBranch
+
+  if (defaultBranchName === null) {
```
nerdneha: it looks like the method `_getDefaultBranchOfGithubRepo` can handle the case where `ghRepository.defaultBranch` is null so I think we can skip to L121

app/src/lib/stores/helpers/infer-comparison-branch.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * For `inferComparisonBranch` case where inferring for a repository
+ * hosted on GitHub, but with no forks
+ *
+ * Returns the default branch of the GitHub repository
+ *
+ * @param branches The list of all branches for the repository
+ * @param ghRepository The repository the branch belongs to
+ */
+function _getDefaultBranchOfGithubRepo(
```
nerdneha: should we capitalize the H in GitHub?

nerdneha:
@iAmWillShepherd should we also a unit test for the `branchB` returning `master` for now as the default branch so that, if we change functionality, we can change that test to expect `branchA`

app/src/lib/stores/helpers/infer-comparison-branch.ts
iAmWillShepherd: There actually isn't one. This exists only to make using this function easier for the caller.That is the reason why I created a discriminated union a bit earlier in the history of this branch. 

The thing is, depending on what we're working with (PR, Fork, Gh Repo, everything else), we need to do different things. Since each helper function accepts different arguments, this function has to accept the union of all of those arguments.

Because of that, I made the 2 parameters that all functions need, `branches` and `repository`, `non-nullable`. I chose to keep `currentPullRequest` and `currentBranch` `nullable` instead of `optional` because the types as they are defined in `app-state` are `nullable`

Does that make sense? Also, I am open to going back to using types, but then I'd be back in the position of needing to come up with good names and the caller would be responsible for knowing the different types this function expects.

app/src/lib/stores/helpers/infer-comparison-branch.ts
iAmWillShepherd: I don't believe so since GitHub is one word and not two. But, I have no strong feelings either way, so I'm 👍 to change it.

app/src/lib/stores/helpers/infer-comparison-branch.ts
iAmWillShepherd: Good catch ⚡️

iAmWillShepherd:
💯

iAmWillShepherd:
> @iAmWillShepherd should we also a unit test for the branchB returning master for now as the default branch so that, if we change functionality, we can change that test to expect branchA

Not sure I understand this one. 

Is the expectation that we return the _parent branch_ (not sure of the proper Git terminology here)? I'll admit my current understanding of this is flaky, but I was under the impression that you can only check if a commit is a child of the tip of the current branch. Wouldn't this mean we can't reliably tell what the parent branch would be? 

Can anyone in @desktop/eng provide some insight on this?

iAmWillShepherd:
💯

PR #4791, opened 2018-05-26
bump embedded Git to latest version

shiftkey:
This upgrades our version of Git to 2.17.1 and also adds some new error messages (as `dugite` supports some new error codes).


app/src/lib/git/core.ts
```diff
@@ -247,6 +247,12 @@ function getDescriptionForError(error: DugiteError): string {
       return 'The push was rejected by the remote server because a required status check has not been satisfied.'
     case DugiteError.BranchRenameFailed:
       return 'The branch could not be renamed.'
+    case DugiteError.PathDoesNotExist:
+      return 'The path does not exist on disk.'
+    case DugiteError.InvalidObjectName:
+      return 'The object was not found in the Git repository.'
```
j-f1: I feel like this error message wouldn’t help a Git beginner figure out what’s wrong. Is there a way to communicate what a object is in Git terms so new users will understand?

changelog.json
```diff
@@ -1,5 +1,8 @@
 {
   "releases": {
+    "1.2.1-test1": [
```
iAmWillShepherd: Is this supposed to be a test release?

PR #4794, opened 2018-05-27
Adds new component for displaying banner notifications (NDDB)

### LINKED ISSUE #4768: Add a banner component to complement the notification displayed when the base branch moves ahead of the current branch (NDDB)
We want to answer _why_ am I receiving a notification when the base branch moves ahead of the current branch. The banner will include a blurb which explains the state of the current branch in relation to the base branch in addition to two CTA buttons, _Compare_ and  _Merge..._.

**Pieces of the component**
* Blurb: "Your branch is **3 commits** behind `origin/master`"
* ~Compare: primary button~ 👈 is now tracked in #4808 
* ~Merge: button~ 👈  is now tracked in #4809 
* Dismiss: small icon in top-right corner

User feedback around this piece is documented in section 2 of [this comment](https://github.com/desktop/desktop/issues/4646#issuecomment-394733008)

iAmWillShepherd:
Fixes #4768 

~@donokuda and I paired on this while prepping a prototype for user interviews. Note that the CTA buttons will be wired up in separate PRs as that work may require other components to be refactored.~

~At the moment, I'm unsure if we should pass a `ref` or the `remote.name` **and** `branch.name`. Since this component is not actually being used at the moment, I just used `ref` and passed the string `origin/master` to produce the screenshot below.~
☝️ Just passed the branch instead.

~According to the design, the _Compare_ button should be the primary button, but the order of the buttons differ from the design due to how the `ButtonGroup` component works. @desktop/design does that mean the mockup is incorrect?~

This PR now adds a banner notification without the CTA buttons. (see [comment](https://github.com/desktop/desktop/pull/4794#issuecomment-395483337) for more details).

**Before**
before

**After**

Todo:
- [x] Add a blurb of the form "Your branch is `x` commits behind `ref`"
- [x] Add _Compare_ button (primary)
- [x] Add Merge button
- [x] Add an _x_ octicon for dismissing
- [x] Refine styles


iAmWillShepherd:
I removed the _refine styles_ item from the task list because that work can be tackled in a separate PR.

app/src/ui/notification/new-commits-banner.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react'
+import { ButtonGroup } from '../lib/button-group'
+import { Button } from '../lib/button'
+import { Ref } from '../lib/ref'
+import { Octicon, OcticonSymbol } from '../octicons'
+import { Branch } from '../../models/branch'
+
+interface INewCommitsBannerProps {
+  /**
+   * The number of commits behind `branch`
```
iAmWillShepherd: I tried to make the prop name as self-documenting as possible, so this could be removed, but I'll leave that to whoever reviews this.

app/src/ui/notification/new-commits-banner.tsx
```diff
@@ -0,0 +1,54 @@
...
+   */
+  readonly commitsBehind: number
+
+  /**
+   * The base branch that is ahead
+   */
+  readonly branch: Branch
```
iAmWillShepherd: This name is a bit generic, but I'm open to suggestions for a better name

iAmWillShepherd: I'm going to name this `baseBranch` to keep inline with what dotcom calls the branch that will be merging incoming commits.

screen shot 2018-06-05 at 3 12 32 pm

shiftkey:
@iAmWillShepherd 

> According to the design, the Compare button should be the primary button, but the order of the buttons differ from the design due to how the `ButtonGroup` component works.

`ButtonGroup` was designed for use in dialogs, and that order rule is there so that consumers of it don't need to worry about the platform differences between macOS and Windows.  Maybe that's not the right component to use here?

app/src/ui/notification/new-commits-banner.tsx
shiftkey: :fire::fire::fire: if you can make `commitsBehind` a bit more contextual - commits behind _what_?

iAmWillShepherd:
@shiftkey, I've spoken with @desktop/design about removing the buttons from the banner (at least for now) so that we can get this merged into master by next week. I'm going to remove the button group for now, and will add each button one-by-one in separate PRs (#4808 and #4809 are tracking this work). 

We will be coming up with a small copy change and updating how this component looks (see [comment]() for context), so I'm going to remove the `ready-for-review` label until those changes are in.

app/src/ui/notification/new-commits-banner.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * The target branch that will accept commits
+   * from the current branch
+   */
+  readonly baseBranch: Branch
+}
+
+/**
+ * Banner used to notify user that there branch is _commitsBehind_
```
shiftkey: Does `_commitsBehind_` need to be emphasized here? Can we just use human-friendly words here? Something like:

>  Banner used to notify user that there branch is a number of commits behind the base branch

iAmWillShepherd:
💯

PR #4795, opened 2018-05-28
update Prettier to 1.13.0

shiftkey:
This jumps us from 1.11.0 over [1.12.0](https://prettier.io/blog/2018/04/11/1.12.0.html) to [1.13.0](https://prettier.io/blog/2018/05/23/1.13.0.html).

TL;DR:

 - support for some new syntax changes coming in TS 2.9
 - better nested ternary indenting
 - better use of parentheses in math expressions
 - unordered lists in markdown now use `-`, not `*` (not enabled for us yet, might follow up with another PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "node-sass": "^4.7.2",
     "octicons": "^7.0.1",
     "parallel-webpack": "^2.3.0",
-    "prettier": "1.11.1",
+    "prettier": "1.13.0",
     "request": "^2.72.0",
     "rimraf": "^2.5.2",
     "sass-loader": "^7.0.1",
--- a/script/prettier.ts
+++ b/script/prettier.ts
@@ -11,7 +11,12 @@ const root = Path.dirname(__dirname)
 const prettier = process.platform === 'win32' ? 'prettier.cmd' : 'prettier'
 const prettierPath = Path.join(root, 'node_modules', '.bin', prettier)
 
-const args = ['**/*.scss', '--list-different']
+const args = [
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -82,7 +82,7 @@
     "node-sass": "^4.7.2",
     "octicons": "^7.0.1",
     "parallel-webpack": "^2.3.0",
-    "prettier": "1.11.1",
+    "prettier": "^1.13.0",
```
j-f1: I’d recommend either using a specific version or using `~` since formatting can change in a minor release (as seen above).

script/prettier.ts
```diff
@@ -11,7 +11,12 @@ const root = Path.dirname(__dirname)
 const prettier = process.platform === 'win32' ? 'prettier.cmd' : 'prettier'
 const prettierPath = Path.join(root, 'node_modules', '.bin', prettier)
 
-const args = ['**/*.scss', '--list-different']
+const args = [
+  '**/*.scss',
+  'app/**/*.{ts,tsx}',
+  'script/**/*.ts',
+  '--list-different',
+]
```
j-f1: Let’s add `package.json` to this list — Prettier 1.13 now prints it in the same way Yarn and npm do, and it would be nice to keep `package.json` formatted correctly so that someone installing a new dependency doesn’t end up changing the file’s formatting.

shiftkey: I'm gonna leave out any additive changes for follow-up PRs - this was mostly for when we jump up to TS 2.9 and start exploring some of the new syntax options.

PR #4804, opened 2018-05-29
tweaking the flow of data when updating the repository local state

shiftkey:
**This is targeting #4770 but at the moment I'm opening this up for a review and feedback. I also need to do some testing to ensure I haven't adversely impacted things**

@vanessayuenn this does a bit of cleanup to the flow of data when refreshing the sidebar, based on what I was touching on in this comment thread: https://github.com/desktop/desktop/pull/4770#discussion_r190431032.

The cliff notes:

 - `Repository` now uses a boolean `hasChanges` rather than referencing the array of `WorkingDirectoryChanges` - I don't think we need the count of files here?
 - `Repository` now doesn't have public setters to update properties around the local state - instead you call `Repository.withLocalStatus` and get a new instance
 - `RepositoriesStore` now tracks the local state for each repository in a cache and handles enriching the database instance for the application's needs
 - `AppStore.refreshLocalState` replaces `AppStore. getRefreshedRepositories`, and instead of returning an array of repositories it performs it's Git operations and then calls into `RepositoriesStore` to perform the update

The net effect of this is that we can now use `this.repositoriesStore.onDidUpdate` and `getAll()` again to update `this.repositories` in one spot.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1764,6 +1764,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
   }
 
+  /**
+   * Refresh the local state for a number of repositories in the sidebar
+   *
+   * @param repositories A collection of repositories to check and update
+   * @param fetchRepository Whether to trigger a fetch as part of each repository update
+   */
   private async refreshLocalState(
     repositories: ReadonlyArray<Repository>,
     fetchRepository: boolean = false
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1764,24 +1764,34 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
   }
 
-  private async getRefreshedRepositories(
+  private async refreshLocalState(
     repositories: ReadonlyArray<Repository>,
-    targetRepo?: Repository
-  ): Promise<ReadonlyArray<Repository>> {
-    for (const repo of repositories) {
-      if (!targetRepo || targetRepo.id === repo.id) {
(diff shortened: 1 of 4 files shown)


vanessayuenn:
>Repository now uses a boolean hasChanges rather than referencing the array of WorkingDirectoryChanges - I don't think we need the count of files here?

I was told (either by @nerdneha or @donokuda; 🙈I forgot who exactly) that the blue indication on the changed files might be swapped for a numerical indicator. If that's the case, it would be nice to also reflect that consistently in the repository list.

shiftkey:
@vanessayuenn 👍 that's easy enough to change to a `number` rather than `ReadonlyArray<WorkingDirectoryChange>` if that's where we're heading

donokuda:
> I was told (either by @nerdneha or @donokuda; 🙈I forgot who exactly) that the blue indication on the changed files might be swapped for a numerical indicator. If that's the case, it would be nice to also reflect that consistently in the repository list.

Yup! That's the plan.

shiftkey:
I had a [eureka](https://en.wikipedia.org/wiki/Eureka_(word)) moment about this a few days ago but didn't really get to coming back to this PR because Life™ has gotten in the way:

 > The `Repository` model is being used in our data stores (as a domain model) but also being used directly in components (as a view model) and these should be separate concerns.

For reference, we define `IRepositoryListItem` as the shape of the repository in the sidebar:

```ts
export interface IRepositoryListItem extends IFilterListItem {
  readonly text: ReadonlyArray<string>
  readonly id: string
  readonly repository: Repositoryish
  readonly needsDisambiguation: boolean
}
```

I'm still wrapping my head around this (`repositories` is a top-level property of `IAppState`, so I need to review the usages of that and maybe do some refactoring) but once I've settled on a good way out of this I'll open a fresh PR.

PR #4805, opened 2018-05-29
missing chai import

shiftkey:
Found this one while investigating how I broke #4804 - basically we shouldn't be relying on globals like this, and you only notice it when you hit the code path.


shiftkey:
Don't worry about that VSTS failure - I re-ran the build and it passed, but the commit status isn't updated.

PR #4814, opened 2018-05-30
promote to production

shiftkey:
I've tested the beta upgrade on macOS and Windows and poked around at the Git operations - didn't uncover anything. If something does come up we can look into it, but for now I'd like to get this out to users to close out the known vulnerability.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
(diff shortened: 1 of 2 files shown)


shiftkey:
Actually, I'm chatting with @dscho about a potential SChannel regression (affecting Windows users). Gonna investigate that before I push this out to everyone.

shiftkey:
I've not been able to help with reproducing the issue, so I'm happy to get this out and monitor it over the next day or so to see if anyone stumbles into it.

Issue #4817, opened 2018-05-30
After restarting my client following a desktop client update prompt, I can no longer Fetch Origin


michellepurcell:
## Description

![image](https://user-images.githubusercontent.com/19187948/40717496-70f6ccba-6404-11e8-93f0-6268f1d02dbd.png)

## Version

* GitHub Desktop:

![image](https://user-images.githubusercontent.com/19187948/40717524-8f17ef6c-6404-11e8-96c3-72ac08d83884.png)

* Operating system:

Windows 7 professional.

## Steps to Reproduce

1. I restarted the GH desktop client after noticing a prompt that there was a new update.
2. I made a change to content in a repo.
3. Now when I click fetch Origin I get the error above. I have not touched my properties, my password or anything.

### Expected Behavior

Fetch original to work as normal.

### Logs

[2018-05-30.desktop.production.log](https://github.com/desktop/desktop/files/2053675/2018-05-30.desktop.production.log)

shiftkey:
@michellepurcell thanks for the report. We published an update a couple of hours ago to address a Git security issue, and I'd received a report about a potential SChannel issue, but I couldn't reproduce it myself.

Could you do me a huge favour and open a shell and run this command? This will help with troubleshooting and reporting this upstream:

```
$ git config --show-origin -l
```

michellepurcell:
![image](https://user-images.githubusercontent.com/19187948/40718246-45c0d218-6407-11e8-875b-f5528b46d472.png)

shiftkey:
@dscho I'm going to put together a test release now to downgrade to 2.17.0 and confirm this is specific to the latest build - is there any diagnostics or tracing here you'd like to gather to help get to the root cause of this?

fschutt:
Is there any way to downgrade right now? I am getting a similar (but not the same error) when I try to push to origin:

![image](https://user-images.githubusercontent.com/12084016/40719504-193405b4-6414-11e8-8f3b-ffc79a90425e.png)

The file it's complaining about, exists, however:

![image](https://user-images.githubusercontent.com/12084016/40719624-72a64e5e-6414-11e8-95d0-4b160b18ec7d.png)

Maybe I should note that I installed git with the MinGW shell? This never happened before the update, however. Here is my git config:

```
file:"C:\\ProgramData/Git/config"       core.symlinks=false
file:"C:\\ProgramData/Git/config"       core.autocrlf=true
file:"C:\\ProgramData/Git/config"       core.fscache=true
[...]

PR #4819, opened 2018-05-30
downgrade Git to 2.17.0

shiftkey:
This is an interim fix to help confirm there's an outstanding issue with Git for Windows 2.17.1.


niik:
This turned out to not be an acceptable course of action, see #4817.

PR #4879, opened 2018-06-06
introduce new props for repository list items

shiftkey:
This is some groundwork to land #4770:

 - a new `localRepositoryStateLookup` in `AppState` to map repositories (by their numeric id) to the local state, without needing to poke at the `GitStore` inside components
 - the new properties to display in the UI are now found in `IRepositoryListItem` - this is a more appropriate place than the repository model, as this is what we use to represent UI-specific data
 - `groupRepositories` - the function used to build up the listed repositories - this now uses the new lookup to populate these objects as part of rendering

I've avoided pulling in the computation and UI updates from #4770, but you can update this local cache in `AppStore` by using a snippet like this:

```ts
    const lookup = this.localRepositoryStateLookup
    const status = await gitStore.loadStatus()
    if (status != null) {
      const changedFilesCount = status.workingDirectory.files.length
      const aheadBehind = gitStore.aheadBehind
      lookup.set(repository.id, { aheadBehind, changedFilesCount })
    }

    this.emitUpdate()
```

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -75,7 +75,7 @@ export interface IAppState {
   /**
    * A cache of the latest repository state values, keyed by the repository id
    */
-  readonly localRepositoryStateLookup: Map<string, ILocalRepositoryState>
+  readonly localRepositoryStateLookup: Map<number, ILocalRepositoryState>
 
   readonly selectedState: PossibleSelections | null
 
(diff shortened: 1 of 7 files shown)


shiftkey:
@iAmWillShepherd I'd like this to sit a bit to give @vanessayuenn a chance to 👍 it before this gets merged

PR #4884, opened 2018-06-06
Adds a way to view ReactDevTools

iAmWillShepherd:
While working on ..., I needed a way to easily modify state in order to show/hide the notification banner. I did the work to integrate ReactDevTools into the apps instance of Chrome DevTools, but doing this was hard to maintain as it required you to have to install the extension in a non-intuitive way.

This PR attempts to rectify that by using the React DevTools Electron app to connect to Desktop. The benefits of this is that we can easily use the latest version of the tool without needing to do anything manually.

At the moment, this is not very ergonomic since you have to:
1. Install React DevTools App - `npm install -g react-devtools`
1. Start React DevTools - `react-devtools`
1. Start Desktop - `yarn start`

screen shot 2018-06-06 at 4 41 50 pm

**What's Left?**
- [ ] Remove the need for performing step 1
- [ ] Add new command option `yarn start --reactDev` to automagically open React DevTool  app and connect to Desktop (step 2)
- [ ] Update our docs around debugging


app/package.json
```diff
@@ -56,9 +56,8 @@
     "winston-daily-rotate-file": "^1.4.5"
   },
   "devDependencies": {
-    "devtron": "^1.4.0",
```
iAmWillShepherd: As far as I know, none of use (other than me) actually installed this plugin, so no need to keep it around

app/src/main-process/main.ts
```diff
@@ -406,19 +406,6 @@ app.on(
 function createWindow() {
   const window = new AppWindow()
 
-  if (__DEV__) {
```
iAmWillShepherd: This existed only to install the plugins, since they are being removed, this can go away as well.

app/package.json
shiftkey: I'd still keep this around if it's maintained upstream, [there are a bunch of nice benefits](https://electronjs.org/devtron).

shiftkey:
@iAmWillShepherd some random notes because I know this is early days:

> Install React DevTools App - `npm install -g react-devtools`

This could just be installed as a `devDependency` in our root `package.json` if we then have package scripts to manage this: `yarn add --dev react-devtools`

> Add new command option `yarn start --reactDev` to automagically open React DevTool app and connect to Desktop (step 2)

The docs for `react-devtools` suggest that we need to embed a script at a special location inside the `index.html` to get this working:

```
<!doctype html>
<html lang="en">
  <head>
    <script src="http://localhost:8097"></script> 
```

Not sure if you've cracked this one yet, but it might require some webpack-fu to interoperate with `webpack.development.js` so that we can avoid touching `index.html` and keeping it production-ready.

app/src/ui/index.tsx
```diff
@@ -1,3 +1,4 @@
+import 'react-devtools'
```
shiftkey: Can we hide this behind the `__DEV__` global? The plan is for the underlying package to be removed when building for release, right?

iAmWillShepherd:
Closing this out because the plugin that gets installed for React dev tools should be sufficient.

https://github.com/desktop/desktop/blob/6f2898eb8377dec92b5f10fd43ec2fb3d51a2e53/app/src/main-process/main.ts#L409-L420

I will instead focus my efforts on ensuring that we always have the latest version of react dev tools which was the driving force for this PR.

PR #4136, opened 2018-02-28
upgrade all the yarns

shiftkey:
[v1.4](https://github.com/yarnpkg/yarn/releases/tag/v1.4.0) and [v1.5.0](https://github.com/yarnpkg/yarn/releases/tag/v1.5.0) have a whole lot of bugfixes, so let's see if we can use it here:

 - [x] upgrade Travis config
 - [x] figure out why `xvfb-maybe` is now failing on Travis
 - [x] upgrade Circle config
 - [x] AppVeyor is already upgraded, no need to do any of these hacks
 - [x] update docs to point to new version
 - [x] review test output on CI platforms to ensure contributors can see success/failure
 - [x] ensure new test script will fail in a predictable way with a test failure
 - [x] ensure new test script will fail in a predictable way with incorrect args provided in script
 - [x] give @desktop/core and @desktop/comrades 2 weeks to update their environments

PR #4887, opened 2018-06-07
add support for running unit tests with Chrome Dev Tools shown

shiftkey:
Want to poke at your unit tests while authoring them? Here's a quick way to get access to the Chrome Dev Tools when running tests. Simply add a `debugger` statement at the start of the test to get CDT to break there, then you can step through the tests like it was the live app.

I've updated the docs to mention this, and also remove the `--grep` tip that's no longer relevant since we moved to a different way to running these tests in #4136. **Let me know if that's something you still want back, and we can open a tracking issue to investigate.**

### DIFF
--- a/docs/contributing/setup.md
+++ b/docs/contributing/setup.md
@@ -166,18 +166,8 @@ problems.
 ## Running tests
 
 - `yarn test` - Runs all unit and integration tests
-- `yarn test:unit` - Runs all unit tests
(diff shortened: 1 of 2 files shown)


iAmWillShepherd:
Going to go ahead and give this a merge since I've been using this for at least 2 weeks now and it doesn't touch any app code.

PR #4871, opened 2018-06-05
use babel-minify for production builds to avoid incorrect mangling of JS

shiftkey:
Resolving the underlying issue(s) with `uglify-es` that we unfortunately stumbled upon.

 - [x] confirm problem JS from #4848 is not emitted
 - [x] confirm problem JS from #4859 is not emitted
 - [x] install test build and confirm features from #4848 and #4859 (i.e. can select a co-author and can checkout a fork PR) are working

Fixes #4842

Feel free to use [`reproduce-minification-issue`](https://github.com/desktop/desktop/compare/reproduce-minification-issue) branch (this is based off `release-1.2.2`) to experiment with reproducing the isolated version of the problem against `uglify-es` to help with reporting this upstream. Annotated commits are mentioned in https://github.com/desktop/desktop/pull/4859#issuecomment-394564362.

PR #4889, opened 2018-06-07
Hotfix release to address minification issues

niik:
This PR is based off of the latest stable release (1.2.2) and cherry-picks c126c25bf1ed8803fdab56242a77f042c6bc2255 from #4871 which disables the built-in minification of webpack 4 (uglify-es) in favor of babel-minify.

We've chosen to only pull this particular fix in rather than releasing from master in order to minimize risk until we can give what's on master a more thorough run-through.


niik:
1.2.3-beta1 and 1.2.3 has shipped off of this branch.

changelog.json
```diff
@@ -1,5 +1,13 @@
 {
   "releases": {
+    "1.2.3": [
+      "[Fixed] No autocomplete when searching for co-authors - #4847",
+      "[Fixed] Error when checking out a PR from a fork - #4842"
+    ],
+    "1.2.3-beta1": [
```
say25: Maybe I am wrong but isn't this technically beta0 as it is the same as what is `1.2.3`?

shiftkey: @say25 yeah, it's in the [release process](https://github.com/desktop/desktop/blob/master/docs/process/releasing-updates.md#3-releasing) but oh well - version numbers are cheap.

niik:
This has shipped and seems to function as intended, gonna merge it in so we don't forget

PR #4895, opened 2018-06-08
Change "restart now" to "restart GitHub Desktop"

### LINKED ISSUE #4891: Offering to 'restart GitHub now' instead of offering to 'restart now' when updates are available, can be confusing.
Offering to restart now discourages people who misinterpret restarting as restarting their computer to update GitHub (the desktop app) as they are in the middle of doing something. But offering instead to 'restart GitHub now', encourages those (most people) who are not using GiHub but still using their computer to restart GitHub to receive the new update.

KennethSweezy:
Specifies that you will restart GitHub Desktop and not your computer when there is an update available.
Closes #4891

### DIFF
--- a/app/src/ui/updates/update-available.tsx
+++ b/app/src/ui/updates/update-available.tsx
@@ -26,7 +26,9 @@ export class UpdateAvailable extends React.Component<
           installed at the next launch. See{' '}
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
           or{' '}
-          <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
+          <LinkButton onClick={this.updateNow}>
+            restart GitHub Desktop now
+          </LinkButton>.
         </span>
 
         <a className="close" onClick={this.dismiss}>
--- a/app/src/ui/updates/update-available.tsx
+++ b/app/src/ui/updates/update-available.tsx
@@ -25,7 +25,8 @@ export class UpdateAvailable extends React.Component<
           An updated version of GitHub Desktop is available and will be
           installed at the next launch. See{' '}
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
-          or <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
+          or{' '}
+          <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
         </span>
 
         <a className="close" onClick={this.dismiss}>
--- a/app/src/ui/updates/update-available.tsx
+++ b/app/src/ui/updates/update-available.tsx
@@ -25,7 +25,7 @@ export class UpdateAvailable extends React.Component<
           An updated version of GitHub Desktop is available and will be
           installed at the next launch. See{' '}
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
-          or <LinkButton onClick={this.updateNow}>restart now</LinkButton>.
+          or <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
         </span>
 
         <a className="close" onClick={this.dismiss}>
--- a/app/src/ui/updates/update-available.tsx
+++ b/app/src/ui/updates/update-available.tsx
@@ -27,7 +27,7 @@ export class UpdateAvailable extends React.Component<
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
           or{' '}
           <LinkButton onClick={this.updateNow}>
-            restart GitHub Desktop now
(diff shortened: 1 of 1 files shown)


niik:
Hey @KennethSweezy, looks like there's some formatting issues that's preventing CI from running. You can try automatically fixing these by running `yarn lint:fix`.

I think for the language here, if we do decide to change this text we shouldn't say that `GitHub` will be restarted but rather `GitHub Desktop`.

j-f1:
How about “relaunch now”?

AB1908:
Perhaps `Restart GitHub Desktop now` or `relaunch now`, as @j-f1 mentioned, would be more technically correct. While your fix is definitely helpful, people new to GitHub Desktop should know precisely what is intended by the message.

app/src/ui/updates/update-available.tsx
```diff
@@ -25,7 +25,7 @@ export class UpdateAvailable extends React.Component<
           An updated version of GitHub Desktop is available and will be
           installed at the next launch. See{' '}
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
-          or <LinkButton onClick={this.updateNow}>restart now</LinkButton>.
+          or <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
```
shiftkey: If we're going to be more explicit here to avoid confusion, I'd rather this be updated to `GitHub Desktop` to match with the installed program name - GitHub could be confused with the website or the company.

AB1908: That's great, but I have an amateur question. All that was added was a word, so how come all the build checks failed?

shiftkey: @AB1908 we use [Prettier](https://github.com/prettier/prettier) to ensure our code is formatted in a consistent way, and the additional text now means that Prettier wants it on the next line:

```diff
diff --git a/app/src/ui/updates/update-available.tsx b/app/src/ui/updates/update-available.tsx
index dd2c711a9..6f2fd7813 100644
--- a/app/src/ui/updates/update-available.tsx
+++ b/app/src/ui/updates/update-available.tsx
@@ -25,7 +25,8 @@ export class UpdateAvailable extends React.Component<
           An updated version of GitHub Desktop is available and will be
           installed at the next launch. See{' '}
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
-          or <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
+          or{' '}
+          <LinkButton onClick={this.updateNow}>restart GitHub now</LinkButton>.
         </span>
 
         <a className="close" onClick={this.dismiss}>
```

If you run `yarn lint:prettier --fix` that should apply the change to your branch.

AB1908: Ah that makes sense. Thanks for explaining.

AB1908:
I wonder why the build failed this time. I couldn't make much sense out of the logs from VSTS.

KennethSweezy:
From what I can tell, it seems like this is an issue with VSTS.

app/src/ui/updates/update-available.tsx
```diff
@@ -25,7 +25,10 @@ export class UpdateAvailable extends React.Component<
           An updated version of GitHub Desktop is available and will be
           installed at the next launch. See{' '}
           <LinkButton uri={this.props.releaseNotesLink}>what's new</LinkButton>{' '}
-          or <LinkButton onClick={this.updateNow}>restart now</LinkButton>.
+          or{' '}
+          <LinkButton onClick={this.updateNow}>
+            restart GitHub Desktop now
```
shiftkey: Just porting over @donokuda's comment from #4941 

> I don't think the "now" is necessary (although I don't strongly feel either way.)

I'm also 👍 to dropping this, but these are not strong feels

shiftkey:
I tested the UI for this by applying this line to the AppStore:

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index cef55da5f..5c7579432 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1304,6 +1304,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
         ? imageDiffTypeDefault
         : parseInt(imageDiffTypeValue)
 
+    this._setUpdateBannerVisibility(true)
+
     this.emitUpdateNow()
 
     this.accountsStore.refresh()
```

This is how it currently looks in macOS at minimum width:



And on Windows at minimum width:

shiftkey:
@desktop/core @desktop/comrades any strong feels on this change, given the visuals above?

shiftkey:
I've gotten some 👍 on this and it's a minor change, so I'm going to take this in. Thanks @KennethSweezy for the contribution!

PR #4849, opened 2018-06-01
First steps towards a dark theme

niik:
This PR adds the first piece of what will become a dark theme for GitHub Desktop. #4780 and #4810 preceded this PR in order to get us to a point where we could start working on a dark theme. The final colors and contrasts have not been locked in yet, think of this as a preview of roughly what we're thinking.

screen shot 2018-06-01 at 15 30 30
screen shot 2018-06-01 at 15 29 56

### Enabling the dark theme

This PR only adds the style sheet itself and doesn't contain any method to select the dark theme in the UI. For testing purposes the theme can be enabled by setting the `theme` localStorage variable to `dark` **and reloading the app**.

We'll work on a proper UI for selecting a theme as well as being able to switch themes without reloading the app in a follow-up PR.

```js
// To enable the dark theme, open DevTools and execute
localStorage.setItem('theme', 'dark');
```
```js
// To revert to default theme
localStorage.removeItem('theme')
```

### Illustration woes

We've had some excellent help from @OiYouYeahYou (#3135, #3773) and @nickcraver (https://github.com/desktop/desktop/issues/1515#issuecomment-388525591) towards making our illustrations dynamically styleable :heart:

Unfortunately that work hasn't made it into this PR. There are unfortunate edge cases with both approaches that we haven't had time to address. In the monochrome svgs colorized with css filters approach the `feColorMatrix` approach didn't want cooperate and always rendered the colors slightly off, I imagine this is due to use of different color spaces or perhaps simply incompetence on my part but I couldn't afford to spend more time looking into it. The second approach of inlining the SVGs into the markup doesn't suffer from that problem but unfortunately makes it a bit trickier for us to scale the illustrations correctly.

I'm hoping we can get back to investigating these approaches and finding a way to accurately colorize our illustrations but for now I've had to settle for the quick-n-dirty approach of regular CSS filters.

### TODO

These are the general TODOs that I've got on my list. I imagine we'll land this PR without solving most of them and instead address them in subsequent PRs but that's TBD.

- [x] I've yet to test this and adjust for Windows, that'll be top priority.
  - [ ] App menus need a little work
- [x] PR badge looks awful in dark mode when branch dropdown is open
- [ ] We've decided to not apply or support at dark theme in the welcome flow. Users should really only see that flow once at the initial startup before they have the option to select a theme. Not styling the welcome flow saves us a ton of time since it's illustration heavy. This escape hatch needs to be implemented. - #4919 
[...]

PR #4899, opened 2018-06-08
Extract toolbar badge variables and sync ahead/behind with pr badge

niik:
Prior to this we had two different colors for the PR badge background and the ahead/behind counter in the toolbar, gray-700 and gray-500. This standardizes them to both be gray-600.

Additionally this introduces a new variable for the background color of the PR badge when the dropdown is open.

This is extracted from #4849 for easier reviewing


j-f1:
`badge-background-color` perhaps?

PR #4915, opened 2018-06-11
Extract variables for toast notification colors

niik:
This makes the toast notifications themeable.

Extracted from #4849 for easier reviewing.


niik:
It's worth noting here that there's a ton of duplication going on between zoom-info and full-screen-notification but I've intentionally opted not to address that in this PR.

PR #4919, opened 2018-06-11
Select appearance (theme) via the preferences dialog

niik:
~~**Note: This builds upon #4849 so the diff is much more complicated than it'll be once that lands**~~

This PR adds the ability to switch (and persist) the application theme using the preferences dialog.

theme-selector

### DIFF
--- a/app/src/ui/app-theme.tsx
+++ b/app/src/ui/app-theme.tsx
@@ -1,0 +1,59 @@
+import * as React from 'react'
+import { ApplicationTheme, getThemeName } from './lib/application-theme'
+
+interface IAppThemeProps {
+  readonly theme: ApplicationTheme
+}
+
+/**
+ * A pseudo-component responsible for adding the applicable CSS
+ * class names to the body tag in order to apply the currently
+ * selected theme.
+ *
+ * This component is a PureComponent, meaning that it'll only
+ * render when its props changes (shallow comparison).
+ *
+ * This component does not render anything into the DOM, it's
+ * purely (a)busing the component lifecycle to manipulat the
+ * body class list.
+ */
+export class AppTheme extends React.PureComponent<IAppThemeProps> {
+  public componentDidMount() {
+    this.ensureTheme()
+  }
+
+  public componentDidUpdate() {
+    this.ensureTheme()
+  }
+
+  public componentWillUnmount() {
+    this.clearThemes()
+  }
+
+  private ensureTheme() {
+    const newThemClassName = `theme-${getThemeName(this.props.theme)}`
+    const body = document.body
+
+    if (body.classList.contains(newThemClassName)) {
+      return
+    }
+
+    this.clearThemes()
+
+    body.classList.add(newThemClassName)
+  }
+
+  private clearThemes() {
+    const body = document.body
+
+    for (const className of body.classList) {
+      if (className.startsWith('theme-')) {
+        body.classList.remove(className)
+      }
+    }
+  }
+
+  public render() {
+    return null
+  }
+}
--- a/app/src/ui/app-theme.tsx
+++ b/app/src/ui/app-theme.tsx
@@ -14,7 +14,7 @@ interface IAppThemeProps {
  * render when its props changes (shallow comparison).
  *
  * This component does not render anything into the DOM, it's
- * purely (a)busing the component lifecycle to manipulat the
+ * purely (a)busing the component lifecycle to manipulate the
  * body class list.
  */
 export class AppTheme extends React.PureComponent<IAppThemeProps> {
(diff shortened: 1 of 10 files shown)


niik:
This is ready for review :bow:

app/src/ui/app-theme.tsx
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * A pseudo-component responsible for adding the applicable CSS
+ * class names to the body tag in order to apply the currently
+ * selected theme.
+ *
+ * This component is a PureComponent, meaning that it'll only
+ * render when its props changes (shallow comparison).
+ *
+ * This component does not render anything into the DOM, it's
+ * purely (a)busing the component lifecycle to manipulat the
```
nerdneha: typo: s/`manipulat`/`manipulate`

app/src/ui/app-theme.tsx
```diff
@@ -0,0 +1,59 @@
...
+  public componentDidUpdate() {
+    this.ensureTheme()
+  }
+
+  public componentWillUnmount() {
+    this.clearThemes()
+  }
+
+  private ensureTheme() {
+    const newThemClassName = `theme-${getThemeName(this.props.theme)}`
```
nerdneha: s/`newThemClassName`/`new*Theme*ClassName`

app/src/ui/preferences/appearance.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react'
+//import { Row } from '../lib/row'
```
nerdneha: will we be using this in a later commit or is this old cruft?

app/src/ui/preferences/appearance.tsx
```diff
@@ -0,0 +1,50 @@
...
+  readonly selectedTheme: ApplicationTheme
+  readonly onSelectedThemeChanged: (theme: ApplicationTheme) => void
+}
+
+const themes: ReadonlyArray<ISegmentedItem> = [
+  { title: 'Light', description: 'The default theme of GitHub Desktop' },
+  {
+    title: 'Dark (beta)',
+    description:
+      'A beta version of our coming dark theme. Still under development. Please report any issues you may find to our issue tracker.',
```
nerdneha: the phrase `coming dark theme` sounds weird to me. thoughts @donokuda? i could also just see us remove the word `coming`

app/src/ui/preferences/appearance.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  public render() {
+    const selectedIndex =
+      this.props.selectedTheme === ApplicationTheme.Dark ? 1 : 0
+
+    return (
+      <DialogContent>
+        <VerticalSegmentedControl
```
nerdneha: not for this PR but maybe a future one; what do you think about having `VerticalSegmentedControl` be able to just take the `selectedItem` as a prop instead of the `index` since it's getting the `items` prop as an array anyway?

app/src/ui/preferences/appearance.tsx
niik: > i could also just see us remove the word coming

I think this is what we should do. Language fail on my part

app/src/ui/preferences/appearance.tsx
niik: Copy paste mistakes

niik:
> One thing I'm trying to understand is if we could avoid some logic by dropping the theme-light or theme-dark class directly into the App component's className list (L1717 -- https://github.com/desktop/desktop/pull/4919/files#diff-14c781d4474997ba566167e985791f94R1717)
> 
> I feel like there's a good reason why we're not doing that but it'd be great to understand that from my perspective.

It has to do with inheritance of the variables. We need to override them as close to (and ideally at) the root pseudo element as possible. Here's what it looks like if we apply it on the App component instead of body.

screen shot 2018-06-15 at 13 32 47

See that label, since we haven't put an explicit color on that it's inheriting the default color from body. Now, one could argue that we should set an explicit color on it but by applying it to body we're safe-guarding against these sorts of inconsistencies.

In the future I want to have the dark theme live in its own compiled css file that gets conditionally loaded instead of toggling it through a class name. That way we can actually override `:root` instead which will let us have themeable backdrops and other overrides for pseudo-elements which doesn't inherit from `:root`.

app/src/ui/preferences/appearance.tsx
niik: > what do you think about having VerticalSegmentedControl be able to just take the selectedItem as a prop instead of the index since it's getting the items prop as an array anyway?

Yeah, I think that would work. Only upside to keeping it this way that I can think of is having a similar API/approach to that of `List`

niik:
Ready for another look :bow:

PR #4930, opened 2018-06-13
Show the local path of selected repository as a tooltip

### LINKED ISSUE #4922: Multiple checkouts of the same repository are indistinguishable
Multiple checkouts of the same repository are indistinguishable under the current repository drop down

Something simple like showing the local file path out next to the repository would be great.

None

Should be pretty self explanatory to the user what showing the local path next to the repository name represents.

Here is how it looks now

image

yongdamsh:
This change resolves #4922 

If there's selected repository, a tooltip that shows the local path of the repository will be appeared on the toolbar button.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1457,13 +1457,13 @@ export class App extends React.Component<IAppProps, IAppState> {
       title = __DARWIN__ ? 'No Repositories' : 'No repositories'
     }
 
-    const tooltip = repository ? repository.path : undefined
-
     const isOpen =
       this.state.currentFoldout &&
       this.state.currentFoldout.type === FoldoutType.Repository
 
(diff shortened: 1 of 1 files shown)


yongdamsh:
:octocat:

app/src/ui/app.tsx
```diff
@@ -1457,6 +1457,8 @@ export class App extends React.Component<IAppProps, IAppState> {
       title = __DARWIN__ ? 'No Repositories' : 'No repositories'
     }
 
+    const tooltip = repository ? repository.path : undefined
```
shiftkey: I'd like for this to be only shown when the repository list isn't open. We have a check below for `isOpen` that you can use. Thoughts?

yongdamsh: Yes, showing a tooltip when dropdown is open is redundant and is unnecessary. I made the change. Thanks!

PR #4932, opened 2018-06-13
Track theme usage

niik:
~~**Note: This builds upon #4919 so the diff is much more complicated than it'll be once that lands**~~

This adds parameter to our usage payload so that we can get some sense of how many are using the dark theme vs the light (default) theme.


shiftkey:
@niik did you want this in for the 1.2.4 update to track how the dark theme is adopted?

niik:
> @niik did you want this in for the 1.2.4 update to track how the dark theme is adopted?

Yeah that'd be nice. Lemme get this cleaned up for review

PR #4874, opened 2018-06-05
Integrate notification banner component into Compare View (NDDB)

iAmWillShepherd:
🌵 🌵 🌵 This PR depends on PRs ~#4794~ and ~#4783~ being merged first🌵 🌵 🌵 

Fixes #4767

This PR is just a continuation of the work done in ☝️ and integrates it into the _Compare view_ UI.

See the gifs 👇 to see it in action.

<details>
<summary>Testing</summary>
If you notice the banner doesn't popup right away, try clicking on another tab or performing a manual fetch ([more info](https://github.com/desktop/desktop/projects/10#card-10732873))

### Unpublished repository
1. Create a new branch (it does not matter if you branch off `master` or another branch)
2. Make some commits on `master`
3. Checkout the new branch you created
4. Notification dot and banner should be visible
5. Comparison branch should be `master`

### Published repository (non-fork)
1. Create and publish a new branch (it does not matter if you branch off the _default branch_ or another branch)
2. Make some commits on the _default branch_ and push
3. Checkout the new branch you created
4. Notification dot and banner should be visible
5. Comparison branch should be _default branch_ (likely master)

### Published repository (non-fork) with PR
1. Create a new pull request targeting a branch that’s not the _default branch_
2. Make some commits on the _target branch_
3. Checkout the PR branch 
4. Notification dot and banner should be visible
5. Comparison branch should be the _target branch_

### Published repository (fork) - Starting point for new project (origin/default-branch is ahead of current branch)
1. Fork a repo and clone the forked repo
2. Create a new branch
3. Make some commits on the _default branch_
3. Checkout the branch you created
4. Notification dot and banner should be visible
5. Comparison branch should be the _default branch_ of the forked repo

### Published repository (fork) - Contributing to another project (origin/default-branch is not ahead of current branch)
1. Fork a repo and clone the forked repo
2. Create a new branch
3. Make some commits on the parent repo’s _default branch_
3. Perform a manual fetch (or wait, your choice)
4. Notification dot and banner should be visible
5. Comparison branch should be the _default branch_ of the parent repo (likely `upstream/master`)

</details>

PR #4934, opened 2018-06-13
Add transitions for notification banner (NDDB)

donokuda:
This pull request adds a dismiss transition for the notification when you've diverged from the default branch.


shiftkey:
> 🌵🌵🌵**NOTE: This depends on the pull request #4874 to be merged first** 🌵🌵🌵 

@donokuda I'm confused - this is actually targeting #4874,  so it feels like it's something that can be done before we merge #4874. Am I understanding it right?

donokuda:
> @donokuda I'm confused - this is actually targeting #4874, so it feels like it's something that can be done before we merge #4874. Am I understanding it right?

You are right. I have no clue how to write a cactus banner. 😅

donokuda:
> Any pointers on testing the animations?

Yup! Make sure you have a branch that's behind the default branch by any number of commits and switch to the history tab.

donokuda:
> Transitions work great, just gonna call out that the dark theme makes this message stand out - would be nice to get that addressed before shipping NDDB.

"Make it pop" 😄 

Good looking out! I'll take care of that one ASAP.

PR #4935, opened 2018-06-13
Support .Cake Highlights

say25:
The .cake extension is used for C# Make Builds. https://github.com/cake-build/cake

### DIFF
--- a/app/src/highlighter/index.ts
+++ b/app/src/highlighter/index.ts
@@ -121,6 +121,7 @@ const modes: ReadonlyArray<IModeDefinition> = [
       '.scala': 'text/x-scala',
       '.sc': 'text/x-scala',
       '.cs': 'text/x-csharp',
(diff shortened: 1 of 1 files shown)


niik:
Hey @say25, thanks for opening this. Would you mind opening a PR with a sample `.cake` file over at https://github.com/desktop/highlighter-tests? I know that it's technically C# which we already support but it'd be nice if we could verify that it looks as intended. I know nothing about cake unfortunately but it seems like they support some sort of preprocessor directives on lines starting with `#` and I'd like to make sure the stock c# parser doesn't balk at them.

say25:
Done https://github.com/desktop/highlighter-tests/pull/25

shiftkey:
Some example screenshots of this PR in action:

screen shot 2018-06-15 at 2 06 59 pm

screen shot 2018-06-15 at 2 07 09 pm

PR #4943, opened 2018-06-15
bump package to address lint warning

shiftkey:
I thought I'd caught this as part of #4854 but CI builds now report the classic "you're using a newer version of TS" message when linting:

```sh
$ yarn lint
yarn run v1.5.1
$ yarn lint:src && yarn lint:prettier
$ yarn tslint && yarn eslint-check && yarn eslint
$ tslint -p .
$ eslint --print-config .eslintrc.* | eslint-config-prettier-check
No rules that are unnecessary or conflict with Prettier were found.
$ ts-node script/eslint.ts
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-eslint-parser.

SUPPORTED TYPESCRIPT VERSIONS: ~2.8.1

YOUR TYPESCRIPT VERSION: 2.9.1

=============
...
```

I _sometimes_ see this warning when linting locally, but not always. Maybe something is being stored in `.eslintcache`. I'm not hugely worried about that for now.

`typescript-eslint-parser` has shipped [an update](https://github.com/eslint/typescript-eslint-parser/releases/tag/v16.0.0) which supports TS 2.9, so this should address the CI warnings.

 - [x] bump package
 - [x] refresh CI builds so they are all using `8.11.x` 
 - [x] upgrade root `package.json` to indicate that you need a recent version of Node
 - [x] update setup docs to ensure they too refer to a recent version of Node
 - [x] allow 1 week for @desktop/core and @desktop/comrades to upgrade their dev environments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ branches:
 
 language: node_js
 node_js:
-  - "8.9.0"
+  - "8.11.0"
 
 cache:
   yarn: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ platform:
   - x64
 
 environment:
-  nodejs_version: "8.11.3"
+  nodejs_version: "8.11"
 
 cache:
   - .eslintcache
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ platform:
   - x64
 
 environment:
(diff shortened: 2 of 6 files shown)


iAmWillShepherd:
I assigned you @shiftkey since the intent is to merge this next week.

PR #4944, opened 2018-06-15
Themeable colors for application menus on Windows/Linux

niik:
The current setup for the menu bars on Windows/Linux was using the exact same styles as for the toolbar dropdowns which works great in the light theme and not so much in the dark theme.

This PR puts in place variables which by default inherit the toolbar colors and is customizeable in other themes.

### Dark theme (before)
ghd-windows-menus-dark-before

### Dark theme (after)
ghd-windows-menus-dark

### Light theme (no change)
ghd-windows-menus-light

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -2,3 +2,8 @@ singleQuote: true
 trailingComma: es5
 semi: false
 proseWrap: always
+
+overrides:
+- files: "*.scss"
+  options:
+    printWidth: 200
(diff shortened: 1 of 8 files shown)


shiftkey:
I'm not sure why VSTS is only reporting an issue with linting on Windows, but I uncovered this issue around how Prettier (both the version we're using and the latest stable version) munges the multi-line comments https://github.com/prettier/prettier/issues/4716 - that's the issue which VSTS is picking up on CI.

PR #4948, opened 2018-06-15
let the user compare when a second branch exists

### LINKED ISSUE #4730: Confusing error message when selecting Compare to Branch when there are no branches to compare
Clicking **Branch** -> **Compare to Branch** takes you to the compare tab, but if there are no branches, you get a message stating the branch can't be found.

I expected to see the history list if there are no branches to compare.

This is more example why we should show the history list: if you click anywhere, it takes you back to the history list which is confusing.

### LINKED ISSUE #4947: Compare branch selector is disabled when there are 2 branches or fewer in a repo
## Description

While testing #4874 using a test repo, I noticed the compare branch selector was disabled. Given that I had 2 branches, I expected that I'd be able to compare. Turns out we do a bit of filtering here: https://github.com/desktop/desktop/blob/45c7b55575644e26a190b75b36987d7a45b289e0/app/src/lib/stores/app-store.ts#L753-L761

What weren't doing was taking that filtering into account when we enabled or disabled the compare branch selector. Basically we are off by 1. 

## Version

* GitHub Desktop: Dev build
* Operating system: macOS

## Steps to Reproduce
1. Create a new repo on GitHub.com and clone it
2. Create a new branch and publish it
3. Verify you cannot compare a branch

### Expected Behavior
I should be able to select a branch to compare against

### Actual Behavior
Compare branch selector is disabled

### Logs
Logs won't be helpful here, but here is an image of the state we use to enable the selector.
screen shot 2018-06-15 at 10 11 19 am

shiftkey:
Fixes #4947
Fixes #4730

Existing behaviour on a repository with two branches (`master` is published, `things-to-compare` is then created without publishing):

New behaviour introduced by this PR:

You can then select and see the other branch in the list:

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -137,7 +137,7 @@ export class CompareSidebar extends React.Component<
             placeholder={placeholderText}
             onFocus={this.onTextBoxFocused}
             value={filterText}
-            disabled={allBranches.length <= 1}
+            disabled={allBranches.length < 1}
             onRef={this.onTextBoxRef}
             onValueChanged={this.onBranchFilterTextChanged}
             onKeyDown={this.onBranchFilterKeyDown}
@@ -475,7 +475,7 @@ export class CompareSidebar extends React.Component<
 function getPlaceholderText(state: ICompareState) {
   const { allBranches, formState } = state
 
-  if (allBranches.length <= 1) {
+  if (allBranches.length < 1) {
     return __DARWIN__ ? 'No Branches to Compare' : 'No branches to compare'
   } else if (formState.kind === ComparisonView.None) {
     return __DARWIN__


app/src/ui/history/compare.tsx
```diff
@@ -137,7 +137,7 @@ export class CompareSidebar extends React.Component<
             placeholder={placeholderText}
             onFocus={this.onTextBoxFocused}
             value={filterText}
-            disabled={allBranches.length <= 1}
+            disabled={allBranches.length < 1}
```
iAmWillShepherd: We can just check if it's equal to 0 here and 👇

iAmWillShepherd: I realize now that it doesn't show the entire diff, but I was pointed at the change that follows this one.

shiftkey:
:apple:

SilkyMarket:
It's more technical than I have thought. It may take some time to understand. Mujahid Bakht. @MBakht

PR #4954, opened 2018-06-18
tidy up temporary directory after LFS operation is completed

### LINKED ISSUE #4414: LFS progress temp directories are not being cleaned up
I left github desktop open for a couple of days, I did maybe 25 commits, all small files. However it is leaving thousands of temp directories/files in the User/App/Local/Temp area (windows 7)

I'm guessing this is maybe from it's frequent automatic polling to github to check if anything needs to be pushed/pulled etc. which is fine behavior, however it really needs to clean up after itself better as leaving all this behind on disk for something else to clean up is not good.

I have to go into Temp and manually clean those out every few days because the desktop app just leaves them there forever:

image

shiftkey:
Resolves #4414 

 - [x] test on Windows and confirm no LFS directories leftover after closing
 - [x] test on macOS and confirm no LFS directories leftover after closing

### DIFF
--- a/app/src/lib/progress/from-process.ts
+++ b/app/src/lib/progress/from-process.ts
@@ -63,11 +63,17 @@ function createProgressProcessCallback(
 
       process.on('close', () => {
         disposable.dispose()
-        const directory = Path.dirname(lfsProgressPath)
-        // this callbacks are empty because the Node runtime might report
-        // an error with unlinking - this is in the TEMP directory and there's
-        // not much the application can do at this point, so we'll just move on
-        Fs.unlink(directory, () => {})
+        // the order of these callbacks is important because
+        //  - unlink can only be done on files
+        //  - rmdir can only be done when the directory is empty
+        //  - we don't want to surface errors to the user if something goes
+        //    wrong (these files can stay in TEMP and be cleaned up eventually)
+        Fs.unlink(lfsProgressPath, err => {
+          if (err == null) {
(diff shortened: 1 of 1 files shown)


app/src/lib/progress/from-process.ts
```diff
@@ -62,9 +63,17 @@ function createProgressProcessCallback(
...
-        Fs.unlink(lfsProgressPath, () => {})
+        // the order of these callbacks is important because
+        //  - unlink can only be done on files
+        //  - rmdir can only be done when the directory is empty
+        //  - we don't want to surface errors to the user if something goes
+        //    wrong (these files can stay in TEMP and be cleaned up eventually)
+        Fs.unlink(lfsProgressPath, err => {
+          if (err == null) {
+            const directory = Path.dirname(lfsProgressPath)
+            Fs.rmdir(directory, () => {})
```
iAmWillShepherd: Is it worth capturing any potential errors when calling `FS.rmdir`? At the moment, they get tossed out.

shiftkey: I don't think it's worth capturing and logging, for two reasons:

 - there's not really much the user can do at this point - we need to rely on the OS to handle these temporary files if we can't clean them up
 - logging this error out might be a minor help for the user, but as this can potentially occur whenever we spawn Git LFS (so fetch/push/pull) it could lead to a lot of noisy logs

PR #4965, opened 2018-06-19
Enforce named highlighter chunks in prod

niik:
The latest beta revealed a problem with our setup of named chunks for the async/incremental loading of syntax highlighting modes where the chunks would get persisted to disk using the name we gave them but imported using their chunk ids.

### DIFF
--- a/app/webpack.common.js
+++ b/app/webpack.common.js
@@ -153,6 +153,7 @@ const highlighterConfig = merge({}, commonConfig, {
     chunkFilename: 'highlighter/[name].js',
   },
   optimization: {
+    namedChunks: true,
     splitChunks: {
       cacheGroups: {
         modes: {


shiftkey:
@niik what's the best way to test this? `yarn build:prod && yarn start:prod` locally, or something else?

shiftkey:
@niik what's the best way to test this? `yarn build:prod && yarn start:prod` locally, or something else?

shiftkey:
Seems to be working fine here:

niik:
> @niik what's the best way to test this? yarn build:prod && yarn start:prod locally, or something else?

Exactly. And I did that with, and without the fix to prove that it did what it was supposed to do. I also manually checked the compiled `highlighter.js` to make sure it looked alright.

PR #4966, opened 2018-06-19
compare tab should fetch commits when scroll nears end

### LINKED ISSUE #4964: History list shows only ~100 items, expecting to see more
## Description
The history list is only showing the latest 100 items or so, previous versions of the app could scroll back to the initial commit.

## Version
* GitHub Desktop: 1.2.3
* Operating system: macOS 10.12.6 (16G1408)

## Steps to Reproduce
1. Open the History pane.
2. Scroll down.

### Expected Behaviour
I expect earlier commits to show up in the list. I think they were loaded dynamically as scrolling happened before.

### Actual Behavior
No older commits are showing.

## Additional Information
n/a

### Logs
[2018-06-19.desktop.production.log](https://github.com/desktop/desktop/files/2114252/2018-06-19.desktop.production.log)

shiftkey:
Fixes #4964 by addressing something we missed with the compare tab switch:

 - `gitStore.loadNextHistoryBatch()` would append to `gitStore.history`, which isn't used by the new Compare tab. I've added a function to support retrieving commits that can be then used to update the Compare tab's list of commits.
 - `Compare.onScroll` fires whenever the user scrolls, so we need to be cautious when loading new commits to ensure we don't fire multiple requests.

It was possible to do this with the old History tab (at worst it could fetch two batches of commits, so you could go from 100 -> 300, and then 300 -> 500 commits). In the log you'd see two `git log` operations fire in step:

That's what I'm trying to be so defensive about with the changes in `compare.tsx`, and yes it's not pretty but I'm open to other ideas about how to address it without diving into the List component.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -887,9 +887,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
-  public _loadNextHistoryBatch(repository: Repository): Promise<void> {
+  public async _loadNextHistoryBatch(repository: Repository): Promise<void> {
     const gitStore = this.getGitStore(repository)
-    return gitStore.loadNextHistoryBatch()
+
+    if (enableCompareSidebar()) {
+      const state = this.getRepositoryState(repository)
+      const { formState } = state.compareState
+      if (formState.kind === ComparisonView.None) {
+        const commits = state.compareState.commitSHAs
+        const lastCommitSha = commits[commits.length - 1]
+
+        const newCommits = await gitStore.loadCommitBatch(lastCommitSha)
+        if (newCommits == null) {
+          return
+        }
+
+        this.updateCompareState(repository, state => ({
+          commitSHAs: commits.concat(newCommits),
+        }))
+        this.emitUpdate()
+      }
+    } else {
+      return gitStore.loadNextHistoryBatch()
+    }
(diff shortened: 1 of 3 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -889,9 +889,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        const commits = state.compareState.commitSHAs
+        const lastCommitSha = commits[commits.length - 1]
+
+        const newCommits = await gitStore.loadCommitBatch(lastCommitSha)
+        if (newCommits == null) {
+          return
+        }
+
+        this.updateCompareState(repository, state => ({
+          commitSHAs: commits.concat(newCommits),
```
iAmWillShepherd: This looks like a good opportunity to use that fancy spread operator`[...commits, ...newCommits]`.

shiftkey: I had that initially but avoided it here because I felt the spreading and repopulating the array would be inefficient when dealing with 100 new commits and N x 100 existing commits.

A [quick benchmark with 2 arrays of 20 numbers](https://www.measurethat.net/Benchmarks/Show/3306/0/arrayprototypeconcat-vs-spread-operator-with-more-numbe) on my machine running Chrome shows spread is slightly slower (with a bit more variation in the timings):

iAmWillShepherd:
Gonna go ahead with the merge since the only change was around using some ES fanciness.

Issue #4808, opened 2018-05-29
Render the compare view when compare button is pressed (NDDB)


iAmWillShepherd:
## Goal

Allow users to see how their current branch differs from its base branch from the notification banner.

**Background**
This issue is tightly-coupled to #4768.

## Solution

* Desktop will use the inferred compare-to branch as the base branch of the compare view
* Add a _Compare_ **primary** button to the notification banner
* When the _Compare_ button is clicked, show the _Compare View_
* Dismiss the notification banner
* Use the inferred branch's name to set the branch selector's value
* Wire up analytics for `divergingBranchBannerInitiatedCompare` and `divergingBranchBannerInfluencedCompare` (see #4883)

PR #4969, opened 2018-06-19
Update diverging notification banner text (NDDB)

donokuda:
This pull request updates iterates on the text shown for the diverging banner:

**Before:**
screen shot 2018-06-19 at 11 52 16 am

**After:**
screen shot 2018-06-19 at 11 50 47 am

I removed the second line since I felt like it made the dialog take up more space than needed. I didn't think "You can view these commits..." was particularly helpful, and I'd rather leave it open for people to use their existing Desktop workflow of either choosing "Update from default branch..." or "Merge into current branch" from the branch menu instead. Since we're planning on adding the "View commits" and "Merge..." action button later on, I don't expect removing the second sentence to have any long-lasting negative impacts.

If we feel like we should have some sort of action, I've explored using a link action to view the commits (but we should consider making that part of the feature functional):

screen shot 2018-06-19 at 11 57 46 am

### DIFF
--- a/app/src/ui/notification/new-commits-banner.tsx
+++ b/app/src/ui/notification/new-commits-banner.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { Ref } from '../lib/ref'
 import { Octicon, OcticonSymbol } from '../octicons'
 import { Branch } from '../../models/branch'
+import { LinkButton } from '../lib/link-button';
 
 interface INewCommitsBannerProps {
   /**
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
We will have a CTA button for viewing the commits soon. I still think it's worth stating that they can view these commits by comparing to the branch noted in the banner itself.

donokuda:
> We will have a CTA button for viewing the commits soon. I still think it's worth stating that they can view these commits by comparing to the branch noted in the banner itself.

In that case, we could try to match the explanatory text to the field below with something like "Select the branch `upstream/not-a-default-branch` below to view the commits." 



Feels redundant to have the base branch mentioned twice, but it's more clear as to what to do/search for next.

iAmWillShepherd:
> Feels redundant to have the base branch mentioned twice, but it's more clear as to what to do/search for next.

Yea, I think I will prioritize #4808 once I get #4874 merged. That should get us back to a place where we aren't being overly redundant.

donokuda:
> Yea, I think I will prioritize #4808 once I get #4874 merged. That should get us back to a place where we aren't being overly redundant.

👍SGTM 🙇

PR #4978, opened 2018-06-20
Support Dark Theme for notification banner

donokuda:
_Targets the `integrate-banner-component` branch_

Now 90% less-blinding!

**Before**

giphy 6

**After**
screen shot 2018-06-20 at 9 18 11 am
giphy 5

### DIFF
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -279,4 +279,9 @@ body.theme-dark {
   .blankslate-image {
     filter: #{'invert()'} grayscale(1) brightness(8) contrast(0.6);
   }
+
+  /** Diverging notification banner colors */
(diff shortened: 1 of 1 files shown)


j-f1:
What do you think about making the background and border dark blue to make it a little more visible?

iAmWillShepherd:
> What do you think about making the background and border dark blue to make it a little more visible?

Or some Royal Purple 💜

donokuda:
> What do you think about making the background and border dark blue to make it a little more visible?

Sorry I missed this! I gave it a shot:

screen shot 2018-06-20 at 9 50 07 am

It's not too bad! I think we might be able to tweak it along with any other dark theme related follow-ups. 🌃🌠

PR #4840, opened 2018-05-31
Display a badge indicating the number of changed files

yongdamsh:
This change resolves #4772 

### Changes
- 47f3029 Add a new component file
- a629c57 Apply styles to the badge
- ec0e651 Refactoring with `createState` function
- a216ca6 Change `FilesChangedBadge` component to stateless

### Demo
filechangedbadge_demo

PR #4980, opened 2018-06-20
cut a new beta build

shiftkey:
There's been a couple of recently-merged PRs that would be nice to get out for beta testers.

4edab7e is also there to enable the new change count badge from #4840 which was hidden behind a feature flag - I think it's safe to make this available to everyone.

### DIFF
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -18,12 +18,8 @@ import {
 import { Dispatcher } from '../lib/dispatcher'
 import { IssuesStore, GitHubUserStore } from '../lib/stores'
 import { assertNever } from '../lib/fatal-error'
-import { Octicon, OcticonSymbol } from './octicons'
 import { Account } from '../models/account'
-import {
-  enableCompareSidebar,
-  enableNotificationOfBranchUpdates,
-} from '../lib/feature-flag'
+import { enableCompareSidebar } from '../lib/feature-flag'
 import { FocusContainer } from './lib/focus-container'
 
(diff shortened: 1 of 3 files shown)


app/src/ui/repository.tsx
```diff
@@ -84,11 +80,7 @@ export class RepositoryView extends React.Component<
       return null
     }
 
-    return enableNotificationOfBranchUpdates() ? (
-      <FilesChangedBadge filesChangedCount={filesChangedCount} />
-    ) : (
-      <Octicon className="indicator" symbol={OcticonSymbol.primitiveDot} />
-    )
+    return <FilesChangedBadge filesChangedCount={filesChangedCount} />
```
iAmWillShepherd: Should this be behind beta flag or does it not matter since this is only targeting the beta release channel?

shiftkey: I didn't think it was worth introducing a separate beta flag for just this feature, and I couldn't really find much context in the original PRs.

iAmWillShepherd: > I couldn't really find much context in the original PRs about why we coupled it to NDDB.

It was originally part of NDDB because with the addition of the notification dot to indicate divergence, the meaning of the new changes notification dot was obscured. To fix, the dot was redesigned to be a badge that provides more useful context. It was de-scoped from the NDDB so I could focus on the MVP and to allow someone from the community to take it on.

shiftkey: Aha, thanks for the context.

Yeah, I think we can ship this outside of the NDDB work.

Issue #4985, opened 2018-06-21
selected commit not kept in sync with compare tab state


shiftkey:
I think we're close to cleaning up the old `History` component which might uncover the root cause of this, but this is a regression I uncovered while testing #4874 so ignore that part of the screenshots.

Repro steps:

 - under new History tab, select a commit (whether you're comparing or not doesn't matter)
 - switch from History -> Changes -> History

**Expected:**

 - commit in diff panel appears selected in list

**Actual:**
 
 - no selection persisted, despite diff being active

<img width="1327" src="https://user-images.githubusercontent.com/359239/41726785-4a36a84c-7549-11e8-969a-12e2010e56ea.png">

This is even more noticeable when you're in the middle of a compare, because it's ignores the commit I had selected and reverts to the commit I had chosen before:

<img width="1329" src="https://user-images.githubusercontent.com/359239/41726901-7e8fe0b8-7549-11e8-98e0-fb2780b954ae.png">

PR #4987, opened 2018-06-21
disable feature flag and remove old History sidebar

shiftkey:
Opening this up to clean up the old History tab - I'll follow up with other refactoring PRs around this area, as we've kind of mashed "HistoryState" and "CompareState" together, and hopefully resolve issues like #4985.


app/src/lib/stores/app-store.ts
```diff
@@ -692,8 +691,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
   }
 
-  /** This shouldn't be called directly. See `Dispatcher`. */
-  public async _loadHistory(repository: Repository): Promise<void> {
+  /**
+   * TODO:
+   * This is some legacy code that no longer works with the new Compare tab.
+   * Need to investigate porting this to "refresh" the Compare tab state.
+   */
+  private async _loadHistory(repository: Repository): Promise<void> {
```
shiftkey: If we port this behaviour to the Compare tab correctly this should resolve #4985. Gonna leave it around for now even though there's no actual UI related to this computation work.

PR #5003, opened 2018-06-22
ensure native modules are in sync with current version of electron

shiftkey:
In #4695 I neglected to also bump this file, which is used when `npm`/`yarn` build their native modules, to ensure they bind to the right version of V8. We should get this in and published to beta to ensure there's no last-minute issues with things that rely on this:

 - reading credentials (`keytar`) 
 - installing the CLI tools (`runas`)
 - listing the available editors and shells (`registry-js`)

### DIFF
--- a/app/.npmrc
+++ b/app/.npmrc
@@ -1,4 +1,4 @@
 runtime = electron
 disturl = https://atom.io/download/electron
-target = 1.8.3
+target = 1.8.7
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
Gonna go ahead and give this a early merge since it's just a dependency  update.

PR #5005, opened 2018-06-22
compile scripts as part of CI builds to catch regressions

shiftkey:
Turns out there are some lingering Typescript errors lurking in our infrastructure scripts:

```sh
$ tsc -p script
script/build.ts:126:9 - error TS2322: Type '{ name: string; platform: "darwin" | "linux" | "win32"; arch: string; asar: false; out: string; i...' is not assignable to type 'Options & IPackageAdditionalOptions'.
  Type '{ name: string; platform: "darwin" | "linux" | "win32"; arch: string; asar: false; out: string; i...' is not assignable to type 'Options'.
    Types of property 'arch' are incompatible.
      Type 'string' is not assignable to type '"all" | "ia32" | "x64" | "armv7l" | "arm64" | arch[] | undefined'.

126   const options: packager.Options & IPackageAdditionalOptions = {
            ~~~~~~~


script/review-logs.ts:11:22 - error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.

11     return path.join(process.env.APPDATA, getExecutableName())
                        ~~~~~~~~~~~~~~~~~~~


```

This PR ensures we check our scripts as part of CI, so that we can catch regressions earlier.

### DIFF
--- a/script/build.ts
+++ b/script/build.ts
@@ -124,11 +124,17 @@ function packageApp(
   }
 
   const toPackageArch = (targetArch: string | undefined): packager.arch => {
+    if (targetArch === undefined) {
+      return 'x64'
+    }
+
     if (targetArch === 'arm64' || targetArch === 'x64') {
       return targetArch
     }
 
-    return 'x64'
+    throw new Error(
+      `Building Desktop for architecture '${targetArch}'  is not supported`
+    )
   }
 
   const options: packager.Options & IPackageAdditionalOptions = {
--- a/script/build.ts
+++ b/script/build.ts
@@ -123,10 +123,18 @@ function packageApp(
     )
   }
 
+  const toPackageArch = (targetArch: string | undefined): packager.arch => {
+    if (targetArch === 'arm64' || targetArch === 'x64') {
+      return targetArch
+    }
+
+    return 'x64'
+  }
+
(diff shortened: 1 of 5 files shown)


vsts.yml
```diff
@@ -10,7 +10,7 @@ phases:
       versionSpec: "1.5.1"
   - script: |
        yarn install --force
-    name: Dependencies
+    name: Install
```
shiftkey: Just some other tidy up to make these steps consistent

script/build.ts
```diff
@@ -123,10 +123,18 @@ function packageApp(
     )
   }
 
+  const toPackageArch = (targetArch: string | undefined): packager.arch => {
+    if (targetArch === 'arm64' || targetArch === 'x64') {
```
saul: The `|| targetArch === 'x64'` is unnecessary here - you're returning `x64` as the fallback anyway

shiftkey: @saul fair point, but I like being explicit here to indicate what is supported from user input.

@iAmWillShepherd any feels either way on this?

niik: If we're being explicit, shouldn't we throw if we're given an arch that we don't support?

PR #5006, opened 2018-06-22
probably the last beta before 1.2.4 ships

shiftkey:


changelog.json
```diff
@@ -1,5 +1,11 @@
 {
   "releases": {
+    "1.2.4-beta4": [
+      "[Fixed] Compare tab appends older commits when scrolling to bottom of list - #4964",
```
iAmWillShepherd: Should we call out the new notification banner or hold off until next week?

shiftkey: That seems dependent on merging #4993 and shipping another beta. I think it can wait.

Issue #4883, opened 2018-06-06
Add metrics (NDDB)


iAmWillShepherd:
## Goal
Gain insights into the effectiveness of this (NDDB) feature.

xref: https://github.com/github/analytics/issues/520

## Solution
Metrics required:
1. `divergingBranchBannerDismissal`
2. `divergingBranchBannerInitatedMerge`
3. `divergingBranchBannerInitiatedCompare`
4. `divergingBranchBannerInfluencedMerge` ([commit](https://github.com/desktop/desktop/pull/5034/commits/fa0e6d961eaf1d803e119377b114955aea575599))
5. `divergingBranchBannerDisplayed`

Tasks
- [x] Add new fields to database
- [x] Add new metrics to `IDailyMeasures`
- [x] Add functions to record metrics
- [x] Wire-up functions to relevant actions
  - [x] Banner displayed
  - [x] Banner dismissal
  - [x] Compare CTA (`divergingBranchBannerInitiatedCompare` and `divergingBranchBannerInfluencedMerge`)
  - [x] Merge CTA
- ~~[ ] Update _usage-data_ on marketing site - [PR](https://github.com/github/desktop.github.com/pull/105)~~ - moved to release tracker https://github.com/github/releases/issues/191
- [x] Update central with new measures - [PR](https://github.com/github/central/pull/383)
- [x] Verify @desktop/analytics is ready to receive new metrics

cc/ @desktop/analytics

telliott27:
Do we have a counter for how many times the banner is displayed? That would let us calculate percentages of banners that lead to action versus are dismissed.

iAmWillShepherd:
Nope, but I can add it tho! 👍 I'll try and land it in time to have it in the beta.

iAmWillShepherd:
@telliott27 I've added the ability to record the number of times the banner is displayed 👍

telliott27:
Awesome @iAmWillShepherd! Thanks for adding it!

iAmWillShepherd:
@telliott27, are you're good to go on your end to receive these metrics?

telliott27:
Yep! I'm all set.

dinaaoteman:
Looking at what you're talking about has left me, who doesn't speak English, more than once lost in thought

shiftkey:
I'm going to close this out as the last task (site updates) now lives in my release checklist for Monday: https://github.com/github/releases/issues/191#issuecomment-407440469

PR #5008, opened 2018-06-22
Track times notification banner is shown (NDDB)

iAmWillShepherd:
Adds new metric that was suggested by @desktop/analytics 

xref: https://github.com/desktop/desktop/issues/4883#issuecomment-399252873

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -829,6 +829,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
           aheadBehindOfInferredBranch.behind
       ) {
         this._setDivergingBranchBannerVisibility(true)
+        this._recordDivergingBranchBannerDisplayed()
       }
     } else if (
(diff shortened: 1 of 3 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -829,6 +829,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
           aheadBehindOfInferredBranch.behind
       ) {
         this._setDivergingBranchBannerVisibility(true)
+        this._recordDivergingBranchBannerDisplayed()
```
shiftkey: This might erroneously get incremented for `true` -> `true` transitions which we handle inside `_setDivergingBranchBannerVisibility`. Can we move this inside `_setDivergingBranchBannerVisibility`?

iAmWillShepherd: Yea, that seems like a much better place to put it. 👍

iAmWillShepherd:
🥇

Issue #2449, opened 2017-08-11
The output from the command could not fit into the allocated stdout buffer. Set options.maxBuffer to a larger value than 10485760 bytes


mwp1990:
### Description
I used the github desktop to pull the code of linux repository from my github.and then issue it:
The output from the command could not fit into the allocated stdout buffer. Set options.maxBuffer to a larger value than 10485760 bytes

### Version

**GitHub Desktop version:** 0.7.2

**OS version:** Microsoft Windows[version 6.1.7601]

### Steps to Reproduce

1. using the desktop to pull my respository from my github account.
2. my github respository is https://github.com/torvalds/linux
3. 
The output from the command could not fit into the allocated stdout buffer. Set options.maxBuffer to a larger value than 10485760 bytes

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

**Reproduces how often:** [What percentage of the time does it reproduce?]

shiftkey:
I haven't tested this out on Windows 7 but a quick test on Windows 10 also fails, but in a slightly different way which I think is the same underlying issue as what you're seeing.

There are known issues with cloning the linux repository down to a Windows filesystem - opening **View** | **Toggle Developer tools** shows me the end of a large error message:

<img width="1369" src="https://user-images.githubusercontent.com/359239/29201470-d22e669e-7ea1-11e7-8d31-2864d3a88797.png">

The interesting bit is these messages buried in the output:

```
...
error: unable to create file drivers/gpu/drm/nouveau/nvkm/subdev/i2c/aux.c: No such file or directory
error: unable to create file drivers/gpu/drm/nouveau/nvkm/subdev/i2c/aux.h: No such file or directory
... 
error: unable to create file include/soc/arc/aux.h: No such file or directory
```

These paths are problematic and not recommended for Windows (because they clash with device names from Ye Old Days™). From [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx):

> Do not use the following reserved names for the name of a file:
CON, PRN, **AUX**, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9. Also avoid these names followed immediately by an extension; for example, NUL.txt is not recommended. For more information, see Namespaces.

[...]

PR #5011, opened 2018-06-22
bump the limit for `git status` to 20MB of raw output

shiftkey:
Addresses a known bottleneck from Git output as reported in #2449

This bottleneck seems to come up when you have a lot of untracked files (thousands) so it'd be nice to pause here and think about whether there's a better solution to bumping up this constraint.

### DIFF
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -59,7 +59,8 @@ export async function getStatus(
   const result = await git(
     ['status', '--untracked-files=all', '--branch', '--porcelain=2', '-z'],
     repository.path,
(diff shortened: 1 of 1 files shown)


j-f1:
Is there a way to get this output as a stream instead? You could then append each received chunk to an internal buffer and pull file information out of the beginning of that buffer once you have all the info you need for the file.

shiftkey:
@j-f1 I might have a branch lying around of porting this to a streaming API to see if it gives us any streaming gains. Might be worth revisiting that.

shiftkey:
Closing in favour of #5053

PR #5016, opened 2018-06-23
Add variables and change colors on dark theme

aputinov:


shiftkey:
@aputinov thanks for the contribution! It looks like this change clashes with #4986 which addresses the same issue in a slightly different way. Are you able to add a screenshot of how this now looks with the dark theme?

shiftkey:
I'm going to close this out as #4986 has already been merged and shipped, and I'd like a fresh PR with more context to understand the other changes (if they're still relevant here).

PR #5024, opened 2018-06-25
oops, not quite ready for NDDB

shiftkey:
If you've just updated to 1.2.4 you might have seen a message like this in the compare view:

screen shot 2018-06-25 at 12 29 48 pm

This is something that we're _close_ to shipping, called **Notifications of Diverging from Default Branch** and summarized in https://github.com/desktop/desktop/issues/4646, but we want to properly bake it on beta and get feedback before it goes out to the world.

This PR should ensure the new dialog isn't shown:

 - [x] confirm locally that this isn't rendered for `production`

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -172,7 +172,7 @@ export class CompareSidebar extends React.Component<
 
   private renderNotificationBanner() {
     if (!enableNotificationOfBranchUpdates()) {
-      return
+      return null
     }
 
     if (!this.props.isDivergingBranchBannerVisible) {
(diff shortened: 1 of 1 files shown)


app/src/ui/history/compare.tsx
```diff
@@ -171,10 +171,11 @@ export class CompareSidebar extends React.Component<
   }
 
   private renderNotificationBanner() {
-    if (
-      !enableNotificationOfBranchUpdates ||
-      !this.props.isDivergingBranchBannerVisible
-    ) {
+    if (!enableNotificationOfBranchUpdates()) {
+      return
```
iAmWillShepherd: We should return `null` here.

shiftkey:
:apple:

PR #5026, opened 2018-06-25
only show the dot when NDDB is enabled

shiftkey:
Missed this little detail when looking at #5024:

Note the little dot?

That's our hint to switch over to the History tab, but with nothing to see right now it's gonna be annoying until we can enable NDDB.

 - [x] validate fix
 - [x] :thumbsup: for changelog
 - [x] ship to `production`
 - [x] bump for next beta
 - [x] ship to `beta`

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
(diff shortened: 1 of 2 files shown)


iAmWillShepherd:
I just tested a prod build and did not see any notification dots.


_I initiated a compare to show that this branch is behind `master` and no notification dot or banner is shown_

PR #5035, opened 2018-06-26
Add some pluralization to the banner (NDDB)

### LINKED ISSUE #4996: Update notification banner message when 1 commit behind default branch
When the current branch is 1 commit behind the default branch the text in the notification banner should be updated to say:
- Your branch is 1 **commit** behind `master`.
- You can view **this commit** using the form below.

screen shot 2018-06-21 at 6 26 31 pm

iAmWillShepherd:
Fixes #4996

screen shot 2018-06-26 at 12 09 14 am

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -182,7 +182,8 @@ export class CompareSidebar extends React.Component<
     const { inferredComparisonBranch } = this.props.compareState
 
     return inferredComparisonBranch.branch !== null &&
-      inferredComparisonBranch.aheadBehind !== null ? (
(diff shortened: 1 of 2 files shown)


iAmWillShepherd:
@j-f1 that case _should_ never happen. Just updated the calling code to only return this component when the number of commits behind is greater than 0.

shiftkey:
> Just updated the calling code to only return this component when the number of commits behind is greater than 0.

If only T Y P E S could let us indicate this is a positive number 🤔🤔🤔

j-f1:
You could always do some kind of assertion.

PR #5049, opened 2018-06-27
Disables return key when no matches are found while filtering in the branch list (RBB)

### LINKED ISSUE #4458: Disable return key when no matches exist
When selecting a branch to compare there is a possibility of being given a search string that does not result in any matches. 

**No matches**
before-enter

When in this state, the return key should be disabled. Although pressing <kbd>return</kbd> doesn't have any show-stopping effects, it puts the compare tab in a weird state.

**Ahead/Behind tabs missing when in branch compare mode**
after-enter

This is a UI behavior was not included in the main branch for this feature because it was not a part of the original feature.

For additional context, check out #4182.

iAmWillShepherd:
Fixes #4458


shiftkey:
Tentatively requesting a review from @niik or @nerdneha as I paired on this with @iAmWillShepherd and shouldn't review this myself.

shiftkey:
> I think we just need to figure out this VSTS CI break

See #5188

PR #5050, opened 2018-06-27
Adds metrics tracking around pushing to remotes

### LINKED ISSUE #4570: Metrics tracking for pushes
We don't currently have metrics tracking for pushing to remotes within GitHub Desktop. It would be great to add this to our metrics.

iAmWillShepherd:
Fixes #4570 
- [x] Add measures to database
- [x] Implement functions to update measures
- [x] Make functions accessible to components (`Dispatcher` and `AppStore`)
- [x] Wire up the functions to their related event
- [x] Update central - [PR](https://github.com/github/central/pull/388)
- [x] Update marketing site - [PR](https://github.com/github/desktop.github.com/pull/107)
- [x] Verify @desktop/analytics is good to go


iAmWillShepherd:
Closing this out for now until @desktop/core is in agreement on what insights we'd gain from collecting these metrics.

cc/ @desktop/analytics

iAmWillShepherd:
Now that we have consensus (https://github.com/github/desktop/issues/541), I'm opening this back up...

app/src/lib/stats/stats-store.ts
```diff
@@ -427,6 +430,27 @@ export class StatsStore {
     }))
   }
 
+  /** Record that the user pushed to GitHub.com */
+  public async recordPushToGitHub(): Promise<void> {
+    return this.updateDailyMeasures(m => ({
+      dotcomPushCount: m.dotcomPushCount + 1,
```
shiftkey: These counts are only incremented per-push, and not including the number of commits being pushed as part of this operation. This means a user pushing each commit would look very different to a person pushing at the end of the day, even if they push the same number of commits.

We know the number of commits the current branch is ahead before pushing, so do we care about the granularity of these pushes?

cc @telliott27

telliott27: I don't think we need to explicitly collect the number of commits per push. We separately track the number of commits made, so assuming all or most commits get pushed, we can estimate the number of commits per push with `numCommits / numPushes`

shiftkey:
Friendly bump on this - looks like there's some linting issues

PR #5053, opened 2018-06-27
Use a spawn API for getting status details

### LINKED ISSUE #2449: The output from the command could not fit into the allocated stdout buffer. Set options.maxBuffer to a larger value than 10485760 bytes

shiftkey:
Tackling the known bottleneck identified in #2449

Fixes #2449

TODO: 

 - [x] confirm #2449 is resolved
 - [x] if `getStatus` fails, log and ensure no stale state persisted in `GitStore`
 - [x] ensure flows which rely on a valid `getStatus` report an error to the user that it wasn't able to complete


shiftkey:
@iAmWillShepherd digging into this a bit more, I added some localized timings to the actions that make up `getStatus`:

```diff
diff --git a/app/src/lib/git/status.ts b/app/src/lib/git/status.ts
index bc5a94f3c..80b51e972 100644
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -56,12 +56,15 @@ function convertToAppStatus(status: FileEntry): AppFileStatus {
 export async function getStatus(
   repository: Repository
 ): Promise<IStatusResult> {
+  console.time(`status git invocation`)
   const result = await git(
     ['status', '--untracked-files=all', '--branch', '--porcelain=2', '-z'],
     repository.path,
     'getStatus'
   )
+  console.timeEnd(`status git invocation`)
 
+  console.time(`status parsing`)
   // Map of files keyed on their paths.
   // Note, map maintains insertion order
   const files = new Map<string, WorkingDirectoryFileChange>()
@@ -136,6 +139,7 @@ export async function getStatus(
 
   const workingDirectory = WorkingDirectoryStatus.fromFiles([...files.values()])
 
+  console.timeEnd(`status parsing`)
   return {
     currentBranch,
     currentTip,
```

This showed how the overall timings were split up when running this on your 100000 untracked files example:

```
status git invocation: 393.72705078125ms
status parsing: 67778.72485351562ms
```

I think that's the core of the performance problem you're seeing - now that Git itself can send the 10MB+ of output it was previously erroring on, we've uncovered a new bottleneck when parsing this 10MB of output.

shiftkey:
By my count we're spending most of the `getStatus` work inside `parsePorcelainStatus`:

```
status git invocation: 323.88720703125ms
output length: 1288925
parse porcelain: 61767.81005859375ms
status parsing: 62007.541259765625ms
```

At quick glance, `Array.shift` might not be the most optimal approach here, as lots of articles talk about the work to reindex the array - which would add overhead when we're working with 100000 entries or more. I did a quick spike of using `double-ended-queue` and the results were, well, amazing:

```
status git invocation: 139.11474609375ms
output length: 1288925
parse porcelain: 78.654052734375ms
status parsing: 331.564208984375ms
```

I'll put together a PR to target this branch and we can see how that affects things.

j-f1:
Out of curiosity, what would happen if you `push`ed the elements onto the array, then reversed it?

shiftkey:
@j-f1 is this related to #5186 rather than this PR?

app/src/lib/stores/git-store.ts
```diff
@@ -523,7 +523,15 @@ export class GitStore extends BaseStore {
...
     // isn't suitable because we should preserve the other working directory
     // changes.
-    const status = await getStatus(repository)
+
+    const status = await this.performFailableOperation(() =>
+      getStatus(this.repository)
+    )
+
+    if (status == null) {
+      return
```
shiftkey: Should we log or report an error here that `undoFirstCommit` couldn't complete because status failed - unlikely as it might be?

j-f1:
Yes it is! Sorry about that.

app/src/lib/status-parser.ts
```diff
@@ -47,10 +49,12 @@ export function parsePorcelainStatus(
   // containing special characters are not specially formatted; no quoting or
   // backslash-escaping is performed.
 
-  const fields = output.split('\0')
+  const tokens = output.split('\0')
+  const queue = new Deque(tokens)
```
shiftkey: @j-f1 I avoided this sort of approach (no need to push onto a new array first, the `split` gives us an array we own):

```ts
const fields = output.split('\0')
fields.reverse()

...

while ((field = queue.pop())) {
```

for a couple of reasons:

 - the algorithm is intended to work from beginning to end, and inverting these semantics might be confusing for future people reading the source
 - I wasn't sure how poorly `Array.reverse` would perform when dealing with an array of 100000 elements or so - this might perform better than calling `Array.shift` to consume the array, but I suspect we'd be paying the O(N) cost up front for large status outputs

A quick test of this in the Chrome dev tools:

```js
let arr = []

for (i = 0; i < 1000000; i++) {
  arr.push(i.toString())
}

console.time('reverse')

arr.reverse()

console.timeEnd('reverse')
```

Gives me a timing of 20-25ms - much improved, but given the worst case time of `parsePorcelainStatus` was below 80ms (and did a lot more) I suspect this approach still comes out on top. Happy to be proven wrong though.

shiftkey:
Flagging this one as in-scope for `1.3.0` as the status improvements will help out in many parts of the app.

app/src/lib/stores/git-store.ts
iAmWillShepherd: I think we should log it.

shiftkey: I think a `log.warn` here is fine - it'll end up on disk and if we get any new reports of undoing the first commit we'll have the `getStatus` info to work with.

shiftkey:
Yep, we're limited on reviewers and resources right now - and we believe these changes are of benefit to everyone - so we're going to ensure this gets the maximum time on beta before the next release. #jolo

PR #5062, opened 2018-06-28
keep compare tab in sync with selection

### LINKED ISSUE #4985: selected commit not kept in sync with compare tab state

### LINKED ISSUE #5069: Selected commit in Compare branch view gets replaced with latest when window regains focus
## Description

When you select a commit in the Compare to Branch view, it's diff is presented in the right part of window. However when the window gets out of focus and then is focused again, this diff is replaced with latest commit's diff. Previously selected commit still looks selected and clicking on it doesn't refresh the diff. One has to select different commit to change presented diff. The issue is only present when a commit from "Behind" list is selected.

## Version

* GitHub Desktop: 1.2.6

* Operating system: macOS 10.13.4

## Steps to Reproduce

1. Go to "Compare to Branch" view
2. Select some branch
3. Select some commit from "Behind" list
4. Get the window out of focus / open different application
5. Refocus the window

### Expected Behavior

The diff should remain same

### Actual Behavior

The diff from the first commit on the branch is presented

shiftkey:
This PR removes the last vestiges of `IHistoryState` by moving selection into a root `ICommitSelection` state and ensuring the Compare tab uses it when displaying it's list of commits. This helps us close out some selection-related bugs that have been lurking for a while.

Fixes #4985
Fixes #5069

- [x] confirm #4985 is fixed
- [x] confirm #5069 is fixed
- [x] 📝 all the public members (old and new)
- [x] :fire: any last History-related code
- test various scenarios
   - [x] when viewing history - scrolling populates with older commits, history remains preserved
   - [x] when viewing history - scroll, select commit, change tabs, return -> commit remains selected
   - [x] when viewing history - scroll, select commit, switch to other app, return -> commit remains selected
   - [x] when viewing compare - scroll, select commit, change tabs, return -> commit remains selected
   - [x] when viewing history - select commit, then compare branch -> first ahead commit selected
   - [x] when viewing history - select commit, then compare branch which is not ahead -> no commit selected
   - [x] when viewing compare - select commit, then switch back to history -> first commit selected

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -342,7 +342,7 @@ export enum RepositorySectionTab {
 }
 
 export interface IRepositoryState {
-  readonly selection: ICommitSelection
+  readonly commitSelection: ICommitSelection
   readonly changesState: IChangesState
   readonly compareState: ICompareState
   readonly selectedSection: RepositorySectionTab
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -562,21 +562,16 @@ export interface IBranchesState {
 }
 
 export interface ICommitSelection {
-  /**
-   * The commit currently selected in the app
-   */
+  /** The commit currently selected in the app */
   readonly sha: string | null
-  /**
-   * The list of files associated with the current commit
-   */
+
+  /** The list of files associated with the current commit */
   readonly changedFiles: ReadonlyArray<CommittedFileChange>
-  /**
-   * The selected file inside the selected commit
-   */
+
+  /** The selected file inside the selected commit */
   readonly file: CommittedFileChange | null
-  /**
-   * The diff of the currently-sel
-   */
+
+  /** The diff of the currently-selected file */
   readonly diff: IDiff | null
 }
 
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -693,6 +693,12 @@ export enum CompareActionKind {
   Branch = 'Branch',
 }
 
+export interface ICompareToBranch {
+  readonly kind: CompareActionKind.Branch
+  readonly branch: Branch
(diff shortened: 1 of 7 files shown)


shiftkey:
Tentatively pulling this into 1.3.0 but will move to a later version if we run out of time

app/src/lib/app-state.ts
```diff
@@ -342,7 +342,7 @@ export enum RepositorySectionTab {
 }
 
 export interface IRepositoryState {
-  readonly historyState: IHistoryState
+  readonly selection: ICommitSelection
```
iAmWillShepherd: Can this be renamed to `commitSelection`?

app/src/lib/app-state.ts
```diff
@@ -561,19 +561,22 @@ export interface IBranchesState {
...
-  readonly file: CommittedFileChange | null
-}
-
-export interface IHistoryState {
-  readonly selection: IHistorySelection
-
-  /** The ordered SHAs. */
-  readonly history: ReadonlyArray<string>
-
+  /**
```
iAmWillShepherd: Newline before the start of this comment

iAmWillShepherd: There are two more instances of this.

shiftkey:
:apple:

PR #5072, opened 2018-06-29
Add user research label

billygriffin:
Some issues are more usability-related than others and would benefit from user research. This could take the form of interviews, observation, usability testing, or other methods, but I think it'll be helpful to organize them so we know areas of potential focus for UX research. This doesn't mean we'll necessarily incorporate every one into our upcoming user research plans, but rather that they've been identified as reasonable to include and will be available for inclusion in those efforts.


shiftkey:
I think that's enough of a quorum to get this in and let @billygriffin continue scheming

Issue #5070, opened 2018-06-29
Dark theme buttons don't look clickable


jtlowe:
**Is your feature request related to a problem? Please describe.**
Certain buttons for the Dark Theme are unrecognizable as buttons because they have a black background and do not differentiate as buttons to me. This includes the "Commit to..." button, as well as "Save" buttons found in the options dialogue.

**Describe the solution you'd like**
The button background/border should be changed to make it stand out as a button. 

**Describe alternatives you've considered**
These buttons could keep the blue color as found in the light theme, or use a similar or lighter color to the "Cancel" buttons in the Options dialogue.

**Teachability, Documentation, Adoption, Migration Strategy**

PR #5074, opened 2018-06-29
Change primary button back to blue

ampinsk:
This changes the primary button color in dark theme from the black background back to a blue background ⚫️ ➡️ 🔵 

|Before|After|
|---|---|
|screen shot 2018-06-29 at 2 51 22 pm|screen shot 2018-06-29 at 2 52 10 pm|

Closes https://github.com/desktop/desktop/issues/5070

### DIFF
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -16,11 +16,6 @@ body.theme-dark {
   --text-secondary-color: $gray-400;
   --background-color: $gray-900;
 
-  --button-background: black;
-  --button-hover-background: darken($gray-900, 5%);
-  --button-text-color: $white;
-  --button-focus-border-color: $blue-600;
-
   --link-button-color: lighten($blue-400, 3%);
   --link-button-hover-color: $blue-400;
 
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -16,6 +16,11 @@ body.theme-dark {
   --text-secondary-color: $gray-400;
   --background-color: $gray-900;
 
+  --button-background: $blue;
+  --button-hover-background: lighten($blue, 5%);
+  --button-text-color: $white;
+  --button-focus-border-color: $blue-600;
+
   --link-button-color: lighten($blue-400, 3%);
   --link-button-hover-color: $blue-400;
 


donokuda:
@niik I'll let you handle this one in case there was a specific reason for the original dark theme primary button colors.

niik:
> @niik I'll let you handle this one in case there was a specific reason for the original dark theme primary button colors.

There was just a ton of colors I had to pick in a very short time, ergo the beta label for the theme. Fully expected there to be areas we needed to address afterwards.

I'm 👍 on switching to blue but I don't think we should drop the variables and have them inherit, but copy paste them instead. There's some work that needs to be done to split up our variables into which needs to be implemented in a theme and which doesn't but until that I think we should take the stance of duplicating/reimplementing all colors and shadows but omit the layout oriented ones such as spacers.

I'm vary that by not duplicating we'll end up in scenarios where we tweak some colors and only verify that they look good in one theme. With duplication we can at least be sure that we haven't made it any worse in other themes.

donokuda:
> I think we should take the stance of duplicating/reimplementing all colors and shadows but omit the layout oriented ones such as spacers.
> 
> I'm vary that by not duplicating we'll end up in scenarios where we tweak some colors and only verify that they look good in one theme. With duplication we can at least be sure that we haven't made it any worse in other themes.

ah yeah, that makes sense. I'm onboard with duplicating the variables so that it's explicit what colors are going to be used in each theme. 👍

ampinsk:
@niik Makes sense to me! I added the vars back in 👍

app/styles/themes/_dark.scss
```diff
@@ -16,8 +16,8 @@ body.theme-dark {
   --text-secondary-color: $gray-400;
   --background-color: $gray-900;
 
-  --button-background: black;
-  --button-hover-background: darken($gray-900, 5%);
+  --button-background: $blue;
+  --button-hover-background: lighten($blue, 5%);
```
j-f1: How does changing this to `darken` look? I feel like it might look better that way because it would be more symmetrical since the button color is always getting closer to the background color on hover.

ampinsk: I really like your explanation of getting closer to the background color! But, when I tried this out it felt a little off. I think it's because the button is so familiar from the light theme that it's weird to see it with a different behavior.

niik:
Thank you @ampinsk!

to7fh:
2018-07-02 20:23 غرينتش+03:00, Jed Fox <notifications@github.com>:

PR #5078, opened 2018-07-02
Style merge button in banner

ampinsk:
This styles the merge button in the notification banner! 

_Targets the `then-i-wanna-merge` branch_

|Light theme|Dark theme|
|---|---|
|screen shot 2018-07-02 at 10 57 41 am|screen shot 2018-07-02 at 10 57 56 am|

### DIFF
--- a/app/src/ui/notification/new-commits-banner.tsx
+++ b/app/src/ui/notification/new-commits-banner.tsx
@@ -3,6 +3,7 @@ import { Ref } from '../lib/ref'
 import { Octicon, OcticonSymbol } from '../octicons'
 import { Branch } from '../../models/branch'
 import { Button } from '../lib/button'
+import { ButtonGroup } from '../lib/button-group'
 import { Dispatcher } from '../../lib/dispatcher'
 import { Repository } from '../../models/repository'
 import {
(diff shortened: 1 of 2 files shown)


ampinsk:
Hey @niik, @donokuda and weren't sure if `ButtonGroup` was necessary to use here, do you have any thoughts?

donokuda:
@ampinsk Got any thoughts on having the buttons be the same size? I don't have any particularly strong feels either way but thought it looked nice:

ampinsk:
@donokuda I don't feel too strongly either, but my only thought is that it might look more intrusive if the buttons are bigger?

donokuda:
> @donokuda I don't feel too strongly either, but my only thought is that it might look more intrusive if the buttons are bigger?

Yeah, I can see that. I'm fine with leaving it as is. 👍 

I just realized that the font size for the button is bigger than the text in the banner. 😅Can you adding the `.small-button` class to each of the buttons? That should decrease the font-size and keep the buttons compact in the banner.

https://github.com/desktop/desktop/blob/1cd48a0e26fb64fc4c635ffd4f7cbf1478441a2a/app/styles/ui/_button.scss#L81-L85

ampinsk:
@donokuda I agree this looks cleaner ✨ 

screen shot 2018-07-02 at 3 45 33 pm

niik:
> Hey @niik, @donokuda and weren't sure if ButtonGroup was necessary to use here, do you have any thoughts?

I don't think it's necessary. The `ButtonGroup` will switch the order of the actions on Windows and that should be reserved for 'Yes/No'-type actions in dialogs.

Other than that I think this looks great!

PR #5096, opened 2018-07-04
Support legacy macOS right click alternative

### LINKED ISSUE #4942: Ctrl-click is not exactly the same as right-click
## Description
As far as I know (I might be wrong) ctrl-click should result in the same action as a right-click. For example, when selecting a set of files in the 'changed files' list and then ctrl-clicking them then Desktop will be confused and deselect all but the one item that received the event.

When right-clicking, all items will retain their selection status. As expected.

## Version

* GitHub Desktop: 1.2.3
* Operating system: 10.12.6 (16G1408)

## Steps to Reproduce
1. Select multiple items in the source list.
2. Ctrl-click on one of them.

### Expected Behavior
All items should visually retain their selected status.

### Actual Behavior
All selected items are still selected, but visually it looks like only one is currently selected.

niik:
macOS treats Ctrl + Mouse1 as right click for legacy reasons. This updates the multi-selected support in our lists to recognize that and avoid deselection.

Fixes #4942

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -41,7 +41,7 @@
     "react-addons-shallow-compare": "^15.6.2",
     "react-dom": "^16.3.2",
     "react-transition-group": "^1.2.0",
-    "react-virtualized": "^9.18.5",
+    "react-virtualized": "^9.20.0",
     "registry-js": "^1.0.7",
     "runas": "^3.1.1",
     "source-map-support": "^0.4.15",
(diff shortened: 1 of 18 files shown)


app/src/ui/lib/list/list.tsx
```diff
@@ -822,8 +822,14 @@ export class List extends React.Component<IListProps, IListState> {
         this.props.onRowMouseDown(row, event)
       }
 
+      // macOS allow emulating a right click by holding down the ctrl key while
+      // perorming a "normal" click.
```
shiftkey: typo: `performing`

niik:
🏷

PR #3675, opened 2017-12-24
Convert the scripts to TypeScript

j-f1:
Fixes https://github.com/desktop/desktop/issues/48.

TODO:

- [x] `app-info.js`
- [x] `git-info.js`
- [x] `package-info.js`
- [x] `dev-commands-global.js`
- [x] `webpack.*.js`
- ~[`debug`](https://github.com/desktop/desktop/blob/master/script/debug) (imports [`express`](https://www.npmjs.com/package/@types/express), [`webpack`](https://www.npmjs.com/package/@types/webpack), [`webpack-dev-middleware`](https://www.npmjs.com/package/@types/webpack-dev-middleware), and [`webpack-hot-middleware`](https://www.npmjs.com/package/@types/webpack-hot-middleware)) (maybe merge with `start`?)~
- [x] [`dist-info.js`](https://github.com/desktop/desktop/blob/master/script/dist-info.js)
- [x] [`license-overrides.js`](https://github.com/desktop/desktop/blob/master/script/license-overrides.js) (depended upon by `build.ts`) (indirectly depends on `legal-eagle` which ~has no `@types` package defined~ has TYPES at [`@types/legal-eagle`](https://www.npmjs.com/package/@types/legal-eagle))
- [x] [`publish`](https://github.com/desktop/desktop/blob/master/script/publish) (imports [`aws-sdk`](https://www.npmjs.com/package/aws-sdk) and [`request`](https://www.npmjs.com/package/@types/request))
- [x] [`run.js`](https://github.com/desktop/desktop/blob/master/script/run.js) (depended upon by [`start`](https://github.com/desktop/desktop/blob/master/script/start))
- [x] [`start`](https://github.com/desktop/desktop/blob/master/script/start) (imports [`express`](https://www.npmjs.com/package/@types/express), [`webpack`](https://www.npmjs.com/package/@types/webpack), [`webpack-dev-middleware`](https://www.npmjs.com/package/@types/webpack-dev-middleware), and [`webpack-hot-middleware`](https://www.npmjs.com/package/@types/webpack-hot-middleware))

PR #5109, opened 2018-07-06
test deploy with updated scripts

shiftkey:
Now that #3675 has landed, this deployment ensures that we're still able to cut releases from `master`.


shiftkey:
#jolo

PR #5111, opened 2018-07-06
ensure `github` CLI launches under WSL

### LINKED ISSUE #4998: Fail to start from WSL
## Description

In a [WSL(Windows Subsystem for Linux)](https://docs.microsoft.com/en-us/windows/wsl/about) exclusively available on Windows 10, running `github` CLI command fail to start Github Desktop. 

## Version

* GitHub Desktop: 
Version 1.2.3

* Operating system: 
Microsoft Windows [Version 10.0.17134.112]
WSL enabled with Ubuntu 18.04

## Steps to Reproduce

1. Install WSL on a Windows 10 system by following [Microsoft's Guide](https://docs.microsoft.com/en-us/windows/wsl/install-win10), Ubuntu 18.04 is installed for example.
2. Install Github Desktop.
2. Start `bash`, `wsl` from Start Menu, or start from your installed Linux distro app `ubuntu`
3. Run `github` and you can observe the error message

### Expected Behavior

By running `github` in WSL should the application start as expected as it does in cmd, PowerShell and Git bash on Windows 10.

### Actual Behavior

The application won't start and a script error is printed.
```
sh: 0: Can't open /mnt/d/bob/AppData/Local/GitHubDesktop/bin/..\app-1.2.3\resources\app\static\github.sh
```

## Additional Information

As @shiftkey suggested, [this script](https://github.com/desktop/desktop/blob/master/app/static/win32/github.sh) which do start the application properly in Git Bash, cmd and PowerShell on Windows 10, should be modified to cover WSL.

If it is caused by unexpected behavior from WSL, we might need to check [WSL Github](https://github.com/Microsoft/WSL).

Since WSL keeps adding [more Linux distro](https://aka.ms/wslstore), they should be considered as well.

And regarding to #4641, WSL will be added to supported shell options, so this issue should be fixed by then so Github Desktop can start and be started from WSL.

shiftkey:
Fixes #4998 

There's two parts to this PR:

 1.  our shim for launching Desktop from bash-esque environments at `%LOCALAPPDATA%\GitHubDesktop\bin\github` has a mix of forward slashes and backslashes

I think this needs to be normalized to `/` everywhere but I suspect node is going to keep using `\` because it's Windows. 

2. the launcher doesn't have any checks for it's environment, and based on what VSCode is doing in [their launcher script](https://github.com/Microsoft/vscode/blob/master/resources/win32/bin/code.sh) I think we should do the same.

 - [ ] cut a test release, confirm the shim is generated in the correct format
 - [ ] confirm the `github` launcher works in WSL
 - [ ] confirm the `github` launcher works in Git Bash
 - [ ] confirm the `github` launcher works in PowerShell
 - [ ] confirm the `github` launcher works in CMD

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.3.3-beta1",
+  "version": "1.3.3-test10",
   "main": "./main.js",
   "repository": {
     "type": "git",
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.3.3-test6",
+  "version": "1.3.3-test7",
   "main": "./main.js",
   "repository": {
     "type": "git",
--- a/app/src/main-process/squirrel-updater.ts
+++ b/app/src/main-process/squirrel-updater.ts
@@ -66,7 +66,11 @@ function getBinPath(): string {
 
 function resolveVersionedPath(binPath: string, relativePath: string): string {
   const appFolder = Path.resolve(process.execPath, '..')
-  return Path.relative(binPath, Path.join(appFolder, relativePath))
+  const computedPath = Path.relative(
+    binPath,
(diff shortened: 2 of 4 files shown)


shiftkey:
Closing this out for now because I have more urgent things to tackle, but I'd love for someone more WSL-savvy to see if they can help wrap this up.

PR #5134, opened 2018-07-09
Prepping PR #5053 for review

iAmWillShepherd:
* Wraps the call to `getStatus` inside of `performFailableOperation` in `GitStore` to make sure all errors are handled correctly
* No longer re-logs the error condition; pushing the error to the UI instead now (the actual message hasn't been updated)
* Adds a test to make sure 10k files won't cause an error

~- [ ] Figure out a way to test 100k files without the test timing out~

xref: https://github.com/desktop/desktop/pull/5053

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -495,7 +495,11 @@ export class GitStore extends BaseStore {
     // directory for the next stage. Doing doing a `git checkout -- .` here
     // isn't suitable because we should preserve the other working directory
     // changes.
-    const status = await getStatus(repository)
+
+    const status = await this.performFailableOperation(() =>
+      getStatus(this.repository)
+    )
+
     if (status == null) {
       return
     }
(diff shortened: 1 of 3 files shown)


iAmWillShepherd:
It may be worth figuring out a better way to test later.

shiftkey:
> It may be worth figuring out a better way to test later.

We can manually test the original issue in #2449 in the meantime.

shiftkey:
This PR targets a non-default branch which means it isn't being run on some platforms, so I'm gonna merge it in and we can get back to #5053.

iAmWillShepherd:
> This PR targets a non-default branch which means it isn't being run on some platforms...

Thanks for that, I was wondering why I didn't see all test running.

shiftkey:
@iAmWillShepherd are we able to tweak this test failing on VSTS or skip it if it's going to interfere with CI?  If we want to run some "slow" tests like this (not necessarily run on each build) we can talk about it in #5134:

```
 1) git/status
    getStatus
          Handles at least 25k untracked files without failing:
             Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

iAmWillShepherd:
I looked into why it's failing, but I don't think there's anything I can do other than reverting back to creating 10k files. I only increased it to 25k to get that much closer to the 100k files I originally wanted to test with.

PR #5136, opened 2018-07-10
electron@2.0 upgrade

vanessayuenn:
💥 Upgrades to electron@2.0 (and a bunch of other related dependencies) 💥

- [x] Windows build 🆗
- [x] Linux build 🆗 
- [x] macOS build 🆗 
- [x] test release

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "@types/webpack-merge": "^4.1.3",
     "@types/winston": "^2.2.0",
     "@types/xml2js": "^0.4.0",
-    "electron": "2.0.4",
+    "electron": "2.0.5",
     "electron-builder": "20.19.2",
     "electron-mocha": "^6.0.1",
     "electron-packager": "^12.0.0",
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
     "@types/winston": "^2.2.0",
     "@types/xml2js": "^0.4.0",
     "electron": "2.0.4",
-    "electron-builder": "20.2.0",
+    "electron-builder": "20.19.2",
     "electron-mocha": "^6.0.1",
     "electron-packager": "^12.0.0",
     "electron-winstaller": "2.5.2"
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "electron": "2.0.4",
     "electron-builder": "20.2.0",
     "electron-mocha": "^6.0.1",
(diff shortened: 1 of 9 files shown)


shiftkey:
This error from the VSTS agent seems to suggest a missing dependency when running the unit tests:

```
2018-07-10T14:28:36.9957179Z spawing xvfb-maybe with args: --auto-servernum -- /opt/vsts/work/1/s/node_modules/.bin/electron-mocha -t 10000 --renderer --require ts-node/register --require ./app/test/globals.ts app/test/unit/*.{ts,tsx} app/test/unit/**/*.{ts,tsx}
2018-07-10T14:28:37.1580225Z /opt/vsts/work/1/s/node_modules/electron/dist/electron: error while loading shared libraries: libgtk-3.so.0: cannot open shared object file: No such file or directory
```

It's strange that this works fine on Travis, but whatever, [looks like we need to install this package](https://packages.ubuntu.com/search?suite=precise&arch=any&mode=exactfilename&searchon=contents&keywords=libgtk-3.so.0) as part of the VSTS config:

https://github.com/desktop/desktop/blob/2828ee8e2c3f358e7a2a7d99983035bd6da0cd82/vsts.yml#L24-L26

shiftkey:
I've tentatively marked this as in scope for `1.3.0` - @nerdneha @iAmWillShepherd any questions/concerns about getting this in?

package.json
```diff
@@ -149,10 +149,10 @@
     "@types/webpack-merge": "^4.1.3",
     "@types/winston": "^2.2.0",
     "@types/xml2js": "^0.4.0",
-    "electron": "1.8.7",
-    "electron-builder": "20.2.0",
+    "electron": "2.0.4",
+    "electron-builder": "20.19.2",
```
shiftkey: Just tested this upgrade out in a Linux VM (we only use this for generating Linux installers) and I can still build, package and launch Desktop from there :tada:

shiftkey:
I'm not worried about that VSTS build error - [a queued new build passes](https://github.visualstudio.com/Desktop/_build/results?buildId=2403) but doesn't update the commit status.

shiftkey:
I got a :thumbsup: from @iAmWillShepherd and @nerdneha about merging this in for the 1.3.0 release, so I'm gonna cut a new beta so this can get some external feedback.

PR #5138, opened 2018-07-10
API_ENDPOINT is too broad, causes interop issues

### LINKED ISSUE #5137: Unable to authenticate due to existing API_Endpoint environment variable
## Description

GitHub Desktop is unable to authenticate with GitHub.com because it uses an environment variable called `API_Endpoint` to override the API address and _our_ application uses an environment variable called - you guessed it - `API_Endpoint` for its own nefarious ends.
See https://github.com/desktop/desktop/issues/794

I think the lesson for all of us is to prefix our environment variable names to make them more specific!

## Version
* GitHub Desktop: 1.2.6

shiftkey:
Fixes #5137 

I left a comment in the source rather than writing up additional technical content because it seems too niche.

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -824,7 +824,7 @@ export function getEnterpriseAPIURL(endpoint: string): string {
 
 /** Get github.com's API endpoint. */
 export function getDotComAPIEndpoint(): string {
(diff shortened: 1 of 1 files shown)


shiftkey:
#jolo

Issue #5133, opened 2018-07-09
Markdown headers in dark theme have poor contrast


bonimy:
## Issue
I am using the beta dark theme for GitHub desktop (version 1.2.6) and the syntax coloring for headers/titles in markdown files has a poor contrast to the page background.

![screen shot 2018-07-09 at 12 05 30 pm](https://user-images.githubusercontent.com/12055044/42470890-bb2aae4c-8370-11e8-93b3-720760e088ef.png)

Taken on MacOS High Sierra 10.13.5 

## Preferred solution
Increase the brightness of the markdown header color to be easier on the eyes when using dark theme.

## Alternative solutions
* Let users create custom syntax coloring, editable either as a .json file or more ambitiously in the app itself.

shiftkey:
@bonimy thanks for the report. The blue that is coming through is from the default Codemirror CSS, and thankfully we have the ability to style these diffs however we want if we know the right selector to use. I'm going to leave this open as a chance for someone to contribute a fix to it and figure out what colour works here, but here's a patch I used to help as a starting point:

```diff
diff --git a/app/styles/_variables.scss b/app/styles/_variables.scss
index f9e564a68..dab2d056a 100644
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -343,6 +343,7 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
   --syntax-tag-color: #22863a;
   --syntax-attribute-color: #6f42c1;
   --syntax-link-color: #032f62;
+  --syntax-header-color: #032f62;
 
   // Note that this duration *must* match the `UndoCommitAnimationTimeout`
   // specified in `changes/index.tsx`.
diff --git a/app/styles/themes/_dark.scss b/app/styles/themes/_dark.scss
index d95d687e5..589b04878 100644
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -265,6 +265,7 @@ body.theme-dark {
   --syntax-tag-color: $green-400;
   --syntax-attribute-color: $purple-300;
   --syntax-link-color: $blue-300;
+  --syntax-header-color: $red-300;
 
   // Colors for form errors
   --error-color: $red;
diff --git a/app/styles/ui/_diff.scss b/app/styles/ui/_diff.scss
index 8b9383b26..21b46d03b 100644
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
[...]

PR #5145, opened 2018-07-11
Fix contrast of Markdown headers in dark mode

agisilaos:
I kicked off this PR with the changes @shiftkey proposed in #5133. It's definitely an improvement but I would love to know what @niik, @donokuda, @ampinsk think about it. 

### Light Mode
screen shot 2018-07-12 at 02 18 03

### Dark Mode
42470890-bb2aae4c-8370-11e8-93b3-720760e088ef

### Light Mode
screen shot 2018-07-12 at 02 17 34

### Dark Mode
screen shot 2018-07-12 at 02 00 54

Fixes: #5133

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -355,6 +355,7 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
   --syntax-tag-color: #22863a;
   --syntax-attribute-color: #6f42c1;
   --syntax-link-color: #032f62;
+  --syntax-header-color: #032f62;
 
   // Note that this duration *must* match the `UndoCommitAnimationTimeout`
   // specified in `changes/index.tsx`.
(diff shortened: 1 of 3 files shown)


shiftkey:
@agisilaos can you also include the light theme screenshots?

agisilaos:
Done @shiftkey. 👾 In light mode, I prefer the color we already had. 🤔

agisilaos:
🍄

shiftkey:
>  In light mode, I prefer the color we already had. 🤔

Yep, I chose colours in haste that were supposed to stand out, so the next person would be mortified enough to actually figure out what colours fit with each theme.

agisilaos:
Updated light mode: 

screen shot 2018-07-12 at 03 32 25

And with that, this PR is ready for review. 🎇

PR #5149, opened 2018-07-12
add findAccountForRemoteUrl tests

shiftkey:
Extracting these tests from #4362 for review while we dig into the right fix

### DIFF
--- a/app/test/unit/find-account-test.ts
+++ b/app/test/unit/find-account-test.ts
@@ -26,6 +26,7 @@ describe('findAccountForRemoteURL', () => {
     ),
   ]
 
+  // this test currently fails due to the details outlined in https://github.com/desktop/desktop/issues/4154
   it.skip('gives no account for non-GitHub endpoint', async () => {
     const account = await findAccountForRemoteURL(
       'https://gitlab.com/inkscape/inkscape.git',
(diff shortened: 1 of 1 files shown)


app/test/unit/find-account-test.ts
```diff
@@ -0,0 +1,83 @@
...
+      getEnterpriseAPIURL('https://github.mycompany.com'),
+      'deadbeef',
+      [],
+      '',
+      2,
+      'My Company'
+    ),
+  ]
+
+  it.skip('gives no account for non-GitHub endpoint', async () => {
```
iAmWillShepherd: This test shouldn't be skipped right? It looks like the one test that the original issue was about.

iAmWillShepherd: Alternatively a comment on why this test is being skipped will do.

shiftkey: Yeah, I'm skipping the test so that the build isn't broken. I'll add a comment.

shiftkey:
Dropping the `ready-for-review` label to figure out why Travis has two failing tests and the others pass.

shiftkey:
That seems to be working now? Uh, sure

shiftkey:
> Just want to confirm we want to leave the test on line 30 marked as skipped.

@iAmWillShepherd yes, that's being tracked in #4154 and we don't want to break the build just yet.

PR #5150, opened 2018-07-12
Track whether the app is being interacted with

niik:
This adds a new field to our stats package called `active`. It's a simple boolean which indicates whether the app was interacted with by a user during the reporting window (i.e. one day) or if the app has just sat in the background.

There are currently three actions which we interpret as activity.

1. Clicking on a element which we consider to be an interaction target (i.e. buttons and such)
2. Keyboard activity of any kind
3. Clicking on a non-system (i.e. non-role-based) menu item.

The thought of having any kind of code running on every key stroke and mouse click seemed like a bad idea to me and even more so when taking into account the DOM traversal made to determine whether the click was on an interaction target or not, not to mention updating/reading the dailyMeasure table.

For those reason I've introduced a `UiActivityMonitor` class which can be subscribed to in order to get notified of user activity. This class will only listen for events as long as it itself has subscribers. By doing it this way I could pass that down to the StatsStore which is then responsible for unsubscribing whenever it sees user activity for the first time in a reporting window. Once a stats package has been posted successfully it'll resubscribe again to catch the next activity.

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -181,8 +181,13 @@ export class StatsStore {
     await this.db.launches.add(stats)
   }
 
-  /** Clear the stored daily stats. */
-  private async clearDailyStats() {
+  /**
+   * Clear the stored daily stats. Not meant to be called
+   * directly. Marked as public in order to enable testing
+   * of a specific scenario, see stats-store-tests for more
+   * detail.
+   */
+  public async clearDailyStats() {
     await this.db.launches.clear()
     await this.db.dailyMeasures.clear()
 
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -89,6 +89,8 @@ export class StatsStore {
   /** Has the user opted out of stats reporting? */
   private optOut: boolean
 
+  private activityRecorded: boolean = false
+
   public constructor(db: StatsDatabase) {
     this.db = db
 
@@ -175,6 +177,8 @@ export class StatsStore {
   private async clearDailyStats() {
     await this.db.launches.clear()
     await this.db.dailyMeasures.clear()
+
+    this.activityRecorded = false
   }
 
   /** Get the daily stats. */
@@ -429,6 +433,10 @@ export class StatsStore {
   }
 
   public async recordActivity(): Promise<void> {
+    if (this.activityRecorded) {
+      return
+    }
+
     return this.updateDailyMeasures(m => ({
       active: true,
     }))
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -43,6 +43,7 @@ const DefaultDailyMeasures: IDailyMeasures = {
   divergingBranchBannerInitiatedCompare: 0,
   divergingBranchBannerInfluencedMerge: 0,
   divergingBranchBannerDisplayed: 0,
+  active: false,
 }
 
 interface ICalculatedStats {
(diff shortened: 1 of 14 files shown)


app/src/ui/lib/ui-activity-monitor.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    if (!isInteractionTarget(e.target)) {
+      return
+    }
+
+    this.emit('pointer')
+  }
+
+  private onKeyDown = (e: KeyboardEvent) => {
+    this.emit('keyboard')
```
iAmWillShepherd: For my education only:

Does this mean we count any keyboard press as in-app activity? For instance, if I open the app and press <kbd>J</kbd>, will that count as activity in the app?

I see the mouse event is being inspected to ensure it's not just a random click somewhere in the app. I assume it's not possible to do the same thing with a [keyboard event](https://reactjs.org/docs/events.html#keyboard-events) since it doesn't include any properties that would help there.

app/src/ui/lib/ui-activity-monitor.ts
```diff
@@ -0,0 +1,138 @@
...
+   * kind of action detected (mouse/pointer, keyboard etc).
+   *
+   * @return A disposable object which, when disposed will
+   *         terminate the subscription and prevent any
+   *         further calls to the handler.
+   */
+  public onActivity(handler: (kind: UiActivityKind) => void): Disposable {
+    const emitterDisposable = this.emitter.on('activity', handler)
+
+    if (this.subscriberCount++ === 0) {
```
iAmWillShepherd: Can the this be changed to

```ts
if (this.subscriberCount === 0) {
  this.subscriberCount++
  this.startTracking()
}
```

to keep state changing operators separate from Boolean expressions.

app/src/ui/lib/ui-activity-monitor.ts
```diff
@@ -0,0 +1,138 @@
...
+      this.startTracking()
+    }
+
+    return new Disposable(() => {
+      if (--this.subscriberCount === 0) {
```
iAmWillShepherd: Same story here, but before the conditional statement.

app/src/lib/stats/stats-store.ts
```diff
@@ -84,12 +87,18 @@ type DailyStats = ICalculatedStats & ILaunchStats & IDailyMeasures
...
+  private readonly uiActivityMonitor?: IUiActivityMonitor
+  private uiActivityMonitorSubscription: Disposable | null = null
 
   /** Has the user opted out of stats reporting? */
   private optOut: boolean
 
-  public constructor(db: StatsDatabase) {
+  public constructor(
+    db: StatsDatabase,
+    uiActivityMonitor?: IUiActivityMonitor
```
iAmWillShepherd: Is there ever a case where we won't be passing this type to the store?

app/src/ui/lib/ui-activity-monitor.ts
niik: > Does this mean we count any keyboard press as in-app activity? For instance, if I open the app and press J, will that count as activity in the app?

Yes, see the condition in the PR body. Any keypress is treated as an interaction. I'm not too concerned about false-positives here and it's tricky to figure out if a keyboard event did anything since some components will subscribe to keyboard events on random events (such as the List component).

app/src/lib/stats/stats-store.ts
niik: I made it optional to not have to bother with creating an supplying a fake activity monitor in tests such as `app-test`. I don't feel strongly about it

app/src/ui/lib/ui-activity-monitor.ts
niik: It can, and probably should. Pre-increment and post-increment are not the epitome of easily understandable code.

niik:
🐨

niik:
cc @billygriffin @telliott27

billygriffin:
This seems like exactly what we want to capture if someone is using the app or not, and agreed with your tradeoff decision.

PR #5158, opened 2018-07-13
For all intents and purposes, that's a button

niik:
This was something I found as part of #5150, our clickable diff gutters weren't marked as being buttons so the activity tracking wouldn't react to clicking on them.

This PR fixes that and also adds some CSS to ensure that the cursor isn't a text-selection cursor when hovering over the diff gutter.

### DIFF
--- a/app/src/ui/diff/diff-line-gutter.tsx
+++ b/app/src/ui/diff/diff-line-gutter.tsx
@@ -279,7 +279,7 @@ export class DiffLineGutter extends React.Component<
   }
 
   public render() {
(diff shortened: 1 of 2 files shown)


app/src/ui/diff/diff-line-gutter.tsx
```diff
@@ -279,8 +279,17 @@ export class DiffLineGutter extends React.Component<
   }
 
   public render() {
+    let role =
```
shiftkey: :art: `const`

Issue #1533, opened 2017-05-17
Allow repo list to be grouped by organization or provide option to show full repo name


dvnkshl:
For users who are part of multiple github projects/organizations it would be nice to either be able to group the repo list by organization or have the option to show the full repo name on the list view.

e.g.
<img width="736" alt="screen shot 2017-05-16 at 11 20 21 pm" src="https://cloud.githubusercontent.com/assets/3625028/26141046/c1ba2e62-3a8e-11e7-94db-dca756e0bb88.png">

xt0rted:
This was a feature request I had with the old native version. I'd like to see a way to group the repos so I'd have my own personal repos, forks of other projects, and repos for work.

andyfleming:
Huge +1 here.

Folders to create groups would be ideal. A number of solutions would work.

The specific use cases that hit me with this lacking feature-area are:

* needing a way to group by folder and/or by organization to more easily browse among 50 repos in my sidebar
  * The downside or organizing by org is that forks might be listed in an unintuitive way. Also I might want to organize projects within and org
* 2 clones of a repository in separate locations (and needing to be able to differentiate between the two somehow)

adamreisnz:
Perma +1 from me. I've been advocating for a feature like this since the early days of the Github for Mac beta.

sbonami:
I'd love to filter repo list by keyword/tag as well. Just looking at sdks vs apps or react vs redux, etc. would be great by utilizing Github's repo tags.

shiftkey:
@sbonami are you referring to [Topics](https://blog.github.com/2017-01-31-introducing-topics/)? 

These are currently available as a [preview API](https://developer.github.com/v3/repos/) but this would be an interesting idea to move out it's own issue if you'd like to expand on how this would work when filtering.

sbonami:
@shiftkey Topics, yeah 😅 I'll create a new issue, thanks!

raeesbhatti:
I opened https://github.com/desktop/desktop/pull/5160, that makes small changes to show owner login and repo name in current repository and repositories list.
<img width="1072" alt="screen shot 2018-07-14 at 6 34 59 am" src="https://user-images.githubusercontent.com/10067728/42719503-1053f23a-8730-11e8-85f7-ab40a35c1b97.png">

shiftkey:
Tentatively labelling this as `user-research` as I think we need to think about the original problem a bit. 

@desktop/design would love some :eyes: and ❓s about this when you have bandwidth.

[...]

Issue #1929, opened 2017-06-07
Unable to change Label of Local Repo


mrbryo:
Please summarize the issue in the title, and then use the template below to
fill out the details so we can reproduce the issue on our end.

### Description

When I clone a repo multiple times it has the same name within the application even though each entry has different local folders. I should be able to set a name so I can easily distinguish between each duplicate. You ask why am I duplicating them? Because I use a different tool to upload changes to the server since I don't get direct server access to use the git command line tool to just update the repo on the server.

### Version

What version of GitHub Desktop are you running? This is displayed under the
`About GitHub Desktop` menu item. If you are running from source, include
the commit by running `git rev-parse HEAD` from your local repository.

**GitHub Desktop version:** 0.5.9

The operating system you are running on may also help with reproducing the
issue:

 - If you are on macOS, launch `About This Mac` and write down the OS version
   listed.
 - If you are on Windows, open `Command Prompt` and attach the output of this
   command*: `cmd /c ver`

**OS version:** Mac OS Sierra 10.12.5

### Steps to Reproduce

1. Clone a repo.
2. Clone it again into a new folder.

If the issue involves a specific public repository, including the information
about that repository will make it is easier to recreate the issue.

If you think screenshots or a GIF recording will help demonstrate the issue
better, feel free to add them here.

**Expected behavior:** Change the label that shows in the application for the repo.

**Actual behavior:** No way to change the repo label.

**Reproduces how often:** Every time.

![2017-06-07_11-35-10](https://user-images.githubusercontent.com/1635119/26895133-693544be-4b75-11e7-9a62-ea629b1d0bc4.png)

### Logs

There may be some relevant information in log files generated by GitHub
Desktop:

 - If you are on macOS, attach the most recent log file from:
   `~/Library/Application Support/GitHub Desktop/logs/*.desktop.production.log`
 - If you are on Windows, attach the most recent log file from:
   `%APPDATA%\\GitHub Desktop\\logs\\*.desktop.production.log`

The log files are organized by date, so see if anything was generated for
today's date.

Full Log Pasted

> 2017-05-22T18:02:36.418Z - info: Opt in reported.
> 2017-05-23T16:02:20.192Z - info: Stats reported.
> 2017-06-01T21:43:34.128Z - info: Stats reported.
> 2017-06-07T18:26:51.969Z - info: Stats reported.

#### Additional Information

Any additional information, configuration or data that might be necessary to
reproduce the issue.

[...]

PR #5160, opened 2018-07-14
Show both owner login and repo name in list

raeesbhatti:
Related issue: https://github.com/desktop/desktop/issues/1533


app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -94,7 +91,7 @@ export class RepositoryListItem extends React.Component<
       : path
 
     let prefix: string | null = null
-    if (this.props.needsDisambiguation && gitHubRepo) {
+    if (gitHubRepo) {
       prefix = `${gitHubRepo.owner.login}/`
```
shiftkey: This prefix is used to show the owner in a slightly lighter colour. I'm not sure how the "never disambiguate" change here should impact this:

shiftkey:
@raeesbhatti thanks for opening up this PR. This is something I was thinking about recently as a way to add more context, but I'm going to let this sit for a bit and gather feedback.

Some additional context/history:

 - the goal for using disambiguation here was that we believed we should only show as much as necessary, to avoid cluttering the repository list. For users who mostly work in the same organization, we felt that always showing that organization name felt like redundant information.
 - this disambiguation doesn't help when you have multiple copies of the same GitHub repository and want to differentiate between then. This change to always show the owner doesn't really address that either, and that's okay. The best issue tracking that is probably #1929.
 - we might have to suck it up and investigate more complex groupings (remote -> organization) and how filtering affects this.

raeesbhatti:
@billygriffin My apologies for a late reply. The main reason for implementing this change was to "disambiguate" ownership of repos named like "examples", "sdk", "backend", "site". My work (like many others) spans across repos from multiple orgs, and different workstations have different locally cloned repos; having a repo named "backend" is not very intuitive atm. The existing logic for disambiguation requires two repos with same name to trigger which means that previously mentioned repo names would still not be clear.
The change does not affect any styling or positioning.

shiftkey:
@raeesbhatti thanks for the feedback. For the moment I'm going to close this out as we've not really dug into the underlying issue, and I'd love to explore the problem space better. So please follow along with #1533 in the meantime.

Issue #5152, opened 2018-07-12
`fuzzaldrin-plus` options not being set when calling `.score` & `.match` function in `desktop/app/src/lib/fuzzy-find.ts`


reyronald:
Recently I had to build a similar fuzzy filtering feature like the one that works in many places of Github Desktop. Since this project is open source I decided to take a peak to see how you guys implemented it and what packages you used.

There I noticed that you are passing the `options` to the `fuzzaldrin-plus` functions as a fourth parameter:

https://github.com/desktop/desktop/blob/2da984bbff805a1647c467453b766a8c58939aa9/app/src/lib/fuzzy-find.ts#L11-L13

https://github.com/desktop/desktop/blob/2da984bbff805a1647c467453b766a8c58939aa9/app/src/lib/fuzzy-find.ts#L40-L42

The documentation of that package is kind-of lacking to say the least, so if you check source code for version 0.6.0 you will notice that the options should be the third parameter instead, so right now the `options` object in Desktop's implementation is being completely ignored and using the defaults! :

https://github.com/jeancroy/fuzz-aldrin-plus/blob/84eac1d73bacbbd11978e6960f4aa89f8396c540/src/fuzzaldrin.coffee#L17-L20
```coffee
  score: (string, query, options = {}) ->
    return 0 unless string?.length and query?.length
    options = parseOptions(options, query)
    if options.usePathScoring
```

Also, the option `isPath: boolean` doesn't exist in that version, so I assumed you meant `usePathScoring`. These are the default options the package has...

https://github.com/jeancroy/fuzz-aldrin-plus/blob/84eac1d73bacbbd11978e6960f4aa89f8396c540/src/fuzzaldrin.coffee#L41-L49
```coffee
#Setup default values
parseOptions = (options, query) ->

  options.allowErrors ?= false
  options.usePathScoring ?= true
  options.useExtensionBonus ?= false
  options.pathSeparator ?= defaultPathSeparator
  options.optCharRegEx ?= null
  options.wrap ?= null
```

I debugged Desktop locally just to make sure, and indeed this is the case:

![image](https://user-images.githubusercontent.com/7514993/42643514-4fb309d0-85c7-11e8-88f9-bbb4c5e3d6a9.png)

[...]

PR #5170, opened 2018-07-16
drop usage of fuzzy-find options (these were not applied)

shiftkey:
This relates to #5152 as we discovered these options weren't being applied when invoking `fuzzaldrin-plus`. I tried porting them to the correct options API but this changed the behaviour in strange ways, so until we complete that upgrade and figure it out I'm going to tidy these up to reflect the current reality.

### DIFF
--- a/app/src/lib/fuzzy-find.ts
+++ b/app/src/lib/fuzzy-find.ts
@@ -2,14 +2,8 @@ import * as fuzzAldrin from 'fuzzaldrin-plus'
 
 import { compareDescending } from './compare'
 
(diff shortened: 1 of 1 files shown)


shiftkey:
[We have tests!](https://github.com/desktop/desktop/blob/4f65ea12dee73790edbdef32dc478515227361b6/app/test/unit/fuzzy-find-test.ts)

bond22882:
@shiftkey https//github.com/settings/connections/applications/lv1.00d533fffc873a2

PR #5176, opened 2018-07-17
add the rules for the lock extension

shiftkey:
cc @j-f1 

I moved it down to 60 days for no particular reason aside from "sooner is better".


shiftkey:
Gonna merge this in so the integration can do it's thing.

j-f1:
Just realized you might want to set `exemptLabels` to `['future-proposal']` so we don’t end up locking those.

shiftkey:
@j-f1 done: 56b1fe71f59d3f59202f5949ebc48af606e5f89c

PR #5190, opened 2018-07-18
only perform ahead/behind comparisons when the branch selector is open

### LINKED ISSUE #5142: defer ahead/behind computations until user interacts with the branch selector

With the arrival of #4770 alongside the current compare experience we should take a chance to review and identify background work we can defer in the app. One of these things is the ahead/behind commit counts when seeing which branch you want to compare:

screen shot 2018-07-11 at 2 01 57 pm

While it's very cool to see the ahead/behinds immediately when you open the list, this work is queued and executed using `requestIdleCallback`, which only cares about when Desktop is idle. On underpowered hardware, we've noticed this can consume CPU time when you're working in another app - not something we want to do.

I discussed this with @niik in Slack and I think we need to identify when the user has interacted with the History tab (or even the branch selector) to indicate that the ahead/behind stats are needed. The existing logic for this already ensures work isn't blocking animations, but there might be work in there to balance "don't block animations" and "show the information asap".

An alternative might be to perform proper throttling on the background work, using timers or whatnot. But this doesn't really address the "gotta see them asap" scenario, so I think the previous approach is preferred.

I think the only UX issue here is how to handle asynchronously loading in these items, but I think we can revisit this after underlying engineering work is done to defer the computation.

shiftkey:
This fixes #5142 by using the compare form state to determine when to crunch these stats.

That's the core of this chunk of code in `_updateCompareForm`:

```ts
    if (compareState.showBranchList) {
      this.currentAheadBehindUpdater.schedule(
        currentBranch,
        compareState.defaultBranch,
        compareState.recentBranches,
        compareState.allBranches
      )
    } else {
      this.currentAheadBehindUpdater.clear()
    }
```

 - when the branch list is shown, queue up work to do based on `requestIdleCallback` rules
 - when the branch list is hidden, stop doing work

I also ensured we prioritized the default branch and recent branches before the rest, so you see items filled in from the top.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -964,6 +964,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     if (compareState.showBranchList) {
       this.currentAheadBehindUpdater.schedule(
         currentBranch,
+        compareState.defaultBranch,
         compareState.recentBranches,
         compareState.allBranches
       )
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -952,16 +952,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
-    const currentBranch = branchesState.tip.branch
-    if (currentBranch == null) {
-      return
-    }
-
-    if (this.currentAheadBehindUpdater == null) {
+    if (this.currentAheadBehindUpdater ==== null) {
       return
     }
 
     if (compareState.showBranchList) {
+      const currentBranch = branchesState.tip.branch
+
       this.currentAheadBehindUpdater.schedule(
         currentBranch,
         compareState.defaultBranch,
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -851,10 +851,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const action =
       initialAction != null ? initialAction : getInitialAction(cachedState)
     this._executeCompare(repository, action)
-
-    if (currentBranch != null && this.currentAheadBehindUpdater != null) {
-      this.currentAheadBehindUpdater.schedule(currentBranch, allBranches)
-    }
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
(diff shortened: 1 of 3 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -949,6 +945,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
 
     this.emitUpdate()
+
+    const { branchesState, compareState } = this.getRepositoryState(repository)
+
+    if (branchesState.tip.kind !== TipState.Valid) {
+      return
+    }
+
+    const currentBranch = branchesState.tip.branch
```
iAmWillShepherd: I believe this cannot be null.

https://github.com/desktop/desktop/blob/595b4e7315dc35fd3d8062d0011ae74e6d41a958/app/src/models/tip.ts#L34-L40

app/src/lib/stores/app-store.ts
```diff
@@ -949,6 +945,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    if (currentBranch == null) {
+      return
+    }
+
+    if (this.currentAheadBehindUpdater == null) {
```
iAmWillShepherd: Use `===` here since this type is ` AheadBehindUpdater | null`

app/src/lib/stores/helpers/ahead-behind-updater.ts
```diff
@@ -96,17 +96,45 @@ export class AheadBehindUpdater {
...
-      .filter(to => !this.comparisonCache.has(from, to))
+    const filterBranchesNotInCache = (branches: ReadonlyArray<Branch>) => {
+      return branches
+        .map(b => b.tip.sha)
+        .filter(to => !this.comparisonCache.has(from, to))
+    }
+
+    const otherBranches = [...recentBranches, ...allBranches]
+
+    const branches = defaultBranch
```
iAmWillShepherd: Explicit check here

shiftkey:
:apple:

app/src/lib/stores/app-store.ts
```diff
@@ -952,16 +952,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
-    const currentBranch = branchesState.tip.branch
-    if (currentBranch == null) {
-      return
-    }
-
-    if (this.currentAheadBehindUpdater == null) {
+    if (this.currentAheadBehindUpdater ==== null) {
```
iAmWillShepherd: One too many `=` symbols

PR #5192, opened 2018-07-18
Adds a feature flag for in-app-release notes

iAmWillShepherd:
Put release notes behind a feature flag so we can iterate without affecting our users.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1635,11 +1635,13 @@ export class App extends React.Component<IAppProps, IAppState> {
     if (!this.state.isUpdateAvailableBannerVisible) {
       return null
     }
+    const releaseNotesUri = 'https://desktop.github.com/release-notes/'
 
     return (
       <UpdateAvailable
         dispatcher={this.props.dispatcher}
         newRelease={updateStore.state.newRelease}
+        releaseNotesLink={releaseNotesUri}
         onDismissed={this.onUpdateAvailableDismissed}
       />
     )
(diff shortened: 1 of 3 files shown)


shiftkey:
@iAmWillShepherd looks like you need #5188 on this PR too

app/src/ui/app.tsx
```diff
@@ -1635,11 +1635,13 @@ export class App extends React.Component<IAppProps, IAppState> {
     if (!this.state.isUpdateAvailableBannerVisible) {
       return null
     }
+    const releaseNotesUri = 'https://desktop.github.com/release-notes/'
 
     return (
       <UpdateAvailable
         dispatcher={this.props.dispatcher}
         newRelease={updateStore.state.newRelease}
+        releaseNotesLink={releaseNotesUri}
```
shiftkey: Does this need to be a prop? Seems like this is never going to change, and can just be a local `const` inside the component...

say25: I would think it’s a const but maybe there would be a URL that would support beta release notes

shiftkey: I think we can set that at build time using `__RELEASE_CHANNEL__`, but I need to get back and finish up #3385 so the website can handle this.

iAmWillShepherd: So this change actually reverts some work that was done earlier in this PR. I needed to do this so we can merge this PR and have the app behave as it normally does.

https://github.com/desktop/desktop/blob/0ec5b64ed393efbe43fb58abef39b43d83a9e12d/app/src/ui/app.tsx#L1628-L1635

But to answer your question. No, it does not need to be a prop. I'd rather make that change in another PR since everything here was about getting the feature flag in and getting the component to respond to it.

PR #5200, opened 2018-07-19
validate hostname when cloning a URL

### LINKED ISSUE #4154: Cloning gitlab URL results in GitHub repository being cloned.

shiftkey:
Supersedes #4362

Fixes #4154

This PR addresses two related issues:

 - what's a sane way to test `findAccountForRemoteURL` even though it needs to access the API?
 - can we fix the original bug that was the inspiration for digging into this whole area?

For the first part, I changed the signature to `findAccountForRemoteURL` so that you can pass it a different implementation, to test whether a repository is accessible for that account.

This meant I could stub out the tests to fail unless certain criteria was met:

```ts
const mockCanAccessRepository = (
  account: Account,
  owner: string,
  name: string
) => {
  // private repository, only this person can access it
  if (
    account.endpoint === getDotComAPIEndpoint() &&
    account.login === 'joan' &&
    owner === 'desktop' &&
    name === 'repo-fixture'
  ) {
    return Promise.resolve(true)
  }

  // public repository is accessible to everyone
  if (
    account.endpoint === getDotComAPIEndpoint() &&
    owner === 'inkscape' &&
    name === 'inkscape'
  ) {
    return Promise.resolve(true)
  }

  return Promise.resolve(false)
}
```

I also took some time in there to simplify the internals of `findAccountForRemoteURL`.

For the second part of the issue, I took a different approach to what @JordanMussi did in #4154:

 - return `hostname` from `parseRepositoryIdentifier` if it was found, or `null` otherwise
 - if `hostname` is not set, stick to the current behaviour and probe
 - if `hostname` is set, ensure that the hostname associated with the account matches the `hostname`, otherwise this is a domain mismatch

TODO:

 - [x] test with GitLab repository (full URL) resolves correctly
 - [x] test with GitHub account to ensure valid public repositories (both full URL or `owner/name` alias) resolve correctly
 - [x] test with GitHub account to ensure valid private repositories (both full URL or `owner/name` alias) resolve correctly
 - [x] test with GHE account to ensure valid repositories (both full URL or `owner/name` alias) resolve correctly


app/src/lib/find-account.ts
```diff
@@ -94,10 +70,33 @@ export async function findAccountForRemoteURL(
...
+    const { owner, name, hostname } = repositoryIdentifier
+
+    // Prefer an authenticated dot com account, then Enterprise accounts, and
+    // finally the unauthenticated dot com account.
+    const sortedAccounts = Array.from(allAccounts).sort((a1, a2) => {
+      if (a1.endpoint === getDotComAPIEndpoint()) {
+        return a1.token.length ? -1 : 1
+      } else if (a2.endpoint === getDotComAPIEndpoint()) {
+        return a2.token.length ? 1 : -1
+      } else {
```
iAmWillShepherd: Just curious how we determine an enterprise account here? I'd expect to see a comparison made against `getEnterpriseAPIURL`.

iAmWillShepherd: > While working with things in-memory like this we just have to assume that an account that doesn't match getDotComAPIEndpoint() and has a token is a GHE account.

That's good to know. Could you add that as a comment?

shiftkey: 8fba6c32b

PR #5206, opened 2018-07-20
add metric for detecting when user encounters a merge conflict

### LINKED ISSUE #5201: Attempted merges resulting in merge conflict [metrics]
We'd like to understand how many attempted merges result in a conflict. Compare branch and NDDB features are our first big ones that potentially impact merge conflicts, and that's a problem area we're exploring more generally, so we want to introduce metrics that will help us determine if we're making life easier (or more painful) for our users. This is one of them.

I *think* the instrumentation we need here (if not already present) is:

- Event when someone attempts a merge by any means (NDDB, compare, menu bar)
- Event when someone runs into a merge conflict while attempting a merge

@nerdneha @telliott27 please correct me if I got anything wrong or missed any context that would be helpful for this.

shiftkey:
Closes #5201 

 - [x] get 👍 for new `mergeConflictFromPullCount` and `mergeConflictFromExplicitMergeCount` metric name (cc @telliott27 @billygriffin for :eyes:)
 - [x] get PR for required backend changes up, reviewed and merged - https://github.com/github/central/pull/390
 - [ ] get PR for required website changes up, reviewed and merged - https://github.com/github/desktop.github.com/pull/109

### DIFF
--- a/app/src/lib/dispatcher/error-handlers.ts
+++ b/app/src/lib/dispatcher/error-handlers.ts
@@ -281,11 +281,16 @@ export async function mergeConflictHandler(
     return error
   }
 
-  if (e.metadata.command) {
-    if (e.metadata.command === 'pull') {
-      dispatcher.recordMergeConflictFromPull()
-    } else if (e.metadata.command === 'merge') {
-      dispatcher.recordMergeConflictFromExplicitMerge()
+  const command = e.metadata.command
+
+  if (command != null) {
+    switch (command) {
+      case 'pull':
+        dispatcher.recordMergeConflictFromPull()
+        break
+      case 'merge':
+        dispatcher.recordMergeConflictFromExplicitMerge()
+        break
     }
   }
 
--- a/app/src/lib/dispatcher/error-handlers.ts
+++ b/app/src/lib/dispatcher/error-handlers.ts
@@ -281,7 +281,13 @@ export async function mergeConflictHandler(
     return error
   }
 
-  dispatcher.recordMergeConflictDetected()
+  if (e.metadata.command) {
+    if (e.metadata.command === 'pull') {
+      dispatcher.recordMergeConflictFromPull()
+    } else if (e.metadata.command === 'merge') {
+      dispatcher.recordMergeConflictFromExplicitMerge()
+    }
+  }
 
   dispatcher.showPopup({
     type: PopupType.MergeConflicts,
--- a/app/src/lib/dispatcher/error-handlers.ts
+++ b/app/src/lib/dispatcher/error-handlers.ts
@@ -281,6 +281,8 @@ export async function mergeConflictHandler(
     return error
   }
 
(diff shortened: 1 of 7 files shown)


app/src/lib/stats/stats-database.ts
```diff
@@ -93,6 +93,9 @@ export interface IDailyMeasures {
 
   /** Whether or not the app has been interacted with during the current reporting window */
   readonly active: boolean
+
+  /** The number of times a merge in Desktop has resulted in conflicts that the user needs to resolve */
+  readonly mergeConflictDetected: number
```
iAmWillShepherd: spelling: mergeConflict**s**Detected

shiftkey: I actually prefer singular "merge conflict" label here to focus this metric on the event rather than changing this to "merge conflicts" which might be interpreted as the number of files conflicted or the number of conflicts in a merge. Thoughts?

shiftkey: Or maybe this name can be more explicit: `mergeFailedWithConflictsCount` ?

billygriffin: I like the explicit nature of the last one, but what about just `mergeConflictCount`?

shiftkey: My gut feeling is still that "conflict" on it's own is open to (mis-)intepretation, and more context here is required to avoid that. 

With a bit more thought I probably would have opened this PR with `mergeConflictDetectedCount` here because we use `Count` for lots of these other metrics, but the bike shed is green, not red :trollface:

j-f1: `countOfMergesWithConflicts`?

j-f1: Or `conflictedMergeCount`?

shiftkey:
:apple:

app/src/lib/dispatcher/error-handlers.ts
```diff
@@ -281,6 +281,14 @@ export async function mergeConflictHandler(
     return error
   }
 
+  if (e.metadata.command) {
```
iAmWillShepherd: Explicit check for `e.metadata.command`. Additionally, we could `return` to avoid nesting `if` statements.

app/src/lib/dispatcher/error-handlers.ts
```diff
@@ -281,6 +281,14 @@ export async function mergeConflictHandler(
...
+    if (e.metadata.command === 'pull') {
```
iAmWillShepherd: Maybe we can pull out `e.metadata.command` into its own variable?

app/src/lib/dispatcher/error-handlers.ts
shiftkey: Try out f1ba9308e and see how that feels

shiftkey:
🍒

PR #5216, opened 2018-07-22
updated setup.md

xeon-zolt:
added gnupg which is required by nodejs script on ubuntu 16.04

```
+ curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
(23) Failed writing body
Error executing command, exiting

```
screenshot_20180722_215057

### DIFF
--- a/docs/contributing/setup.md
+++ b/docs/contributing/setup.md
@@ -64,7 +64,7 @@ $ sudo ln -s `find /usr/lib64/ -type f -name "libreadline.so.7.0"` /usr/lib64/li
 First, install curl:
 
 ```shellsession
-$ sudo apt install curl
+$ sudo apt install curl gnupg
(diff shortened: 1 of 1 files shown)


docs/contributing/setup.md
```diff
@@ -64,7 +64,7 @@ $ sudo ln -s `find /usr/lib64/ -type f -name "libreadline.so.7.0"` /usr/lib64/li
 First, install curl:
```
shiftkey: 📝 can we also mention here about the need for a GPG agent:

> First, install curl and a GPG program:

shiftkey:
@xeon-zolt thanks!

PR #5238, opened 2018-07-24
support passing values to "update[thing]State" functions inside AppStore

shiftkey:
The motivation for this came from merging `master` into #5062, running it up locally and breaking the Compare when I tried to compare a branch:

"But Brendan, isn't TypeScript supposed to save us from these errors!"

Yes, but even TS has it's limits about checking code, and this is one of those situations.

Let me explain, using the snippet of problem code that triggers the above issue:

```ts
    this.updateCompareState(repository, s => ({
      formState: {
        comparisonBranch,
        kind: action.mode,
        aheadBehind,
      },
      filterText: comparisonBranch.name,
      commitSHAs,
    }))

    const { ahead, behind } = compare
    const aheadBehind = { ahead, behind }

    const commitSHAs = compare.commits.map(commit => commit.sha)
```

This is valid TS code, and the compiler does infer things correctly:

But TS can't actually verify that these local variables will be initialized in the correct order, because we're passing a callback into `updateCompareState` - turns out this is executed immediately, resulting in the bug.

This PR is intended to support two ways of updating state:

 - pass in a new `Pick<T, K>` with the values that need to be updated (no need to look at the current state)
 - pass in a callback `(state: T) => Pick<T, K>` to look at the existing state and determine what to update

This helps us to tidy up lots of places where we don't care about the current state, improving readability of this massive file, and also helps us catch errors like this earlier:

Moving the uses requiring state to `update[Thing]StateFunc` helps us to then revisit these usages and see if we can pass in values instead.


shiftkey:
Closing this out in favour of some more radical ideas. Stay tuned 😈

PR #5239, opened 2018-07-24
enable NDDB and repository info indicators for everyone

shiftkey:


iAmWillShepherd:
Don't we need to hold off on merging this until next Monday?

shiftkey:
@iAmWillShepherd unless we're planning to ship a hotfix release between now and then, Monday will be the next release. I'm just queuing up all the necessary tasks now.

shiftkey:
> Don't we need to hold off on merging this until next Monday?

Oh, another footnote to this - if we do need to cut an emergency hotfix release for whatever reason we'd branch from the `release-1.2.6` tag rather than what's on `master` anyway. So I'm still keen to get `master` as close to what's being shipped sooner rather than later.

PR #5244, opened 2018-07-25
loading in batch of commit should use correct range syntax

### LINKED ISSUE #5243: History missing latest commit
## Description
As of the latest beta, the "History" tab doesn't show the most recent commit and is always one commit behind. This affects all repos.

## Version
* GitHub Desktop: 1.3.0-beta4
* Operating system: Microsoft Windows [Version 10.0.17134.165]

## Actual

Let me know if you need any further info. I believe the steps to replicate should be pretty self-explanatory.

shiftkey:
Fixes #5243 

Suffixing `^` to a commit ID tells Git to refer to the first parent of that commit, so the work in #4966 introduced a subtle bug where the `HEAD` commit wasn't included when doing `git log` to display the start of the commit history (in `loadCommitBatch`).

This means `_loadNextCommitBatch` also needs to be updated, because it actually needs a `^` to ensure it doesn't load in duplicate commits.

 - [x] add regression test for `loadCommitBatch`
 - [x] fix the bug


shiftkey:
Going to add a regression test or two to also validate the fix, rather than rely on end-user testing.

PR #5246, opened 2018-07-25
swap out fatalError for logged warning when OAuth callback reused

### LINKED ISSUE #1073: Reloading the OAuth URL logs an error
From https://github.com/desktop/desktop/issues/1002:

>In addition, if a user is already signed into the Firefox browser, and clicks the Sign in using your browser linkin the preferences module, Firefox will open a tab with a GitHub Desktop popup seeking permission. I noticed that tab in FF stays open even if you successfully signed into Desktop. (In other browsers it closes or is handled different way.) Clicking the link again in FF again, results in console error below because the user is already signed in.

shiftkey:
Fixes #1073 by moving the message to a logged warning, and returning `null`.

If the browser loads the callback URL again after the user has completed the OAuth flow, the app won't display any errors (because `SignInStore` is no longer waiting for the callback) but the logged message will help us if anyone ends up reaching out to support about it


app/src/lib/oauth.ts
```diff
@@ -40,11 +40,12 @@ export function askUserToOAuth(endpoint: string) {
  */
 export async function requestAuthenticatedUser(
   code: string
-): Promise<Account | null> {
+): Promise<Account | null | undefined> {
```
shiftkey: This was necessary to prevent the original code path from considering "this callback is useless" the same as "i have a token but it doesn't work"  and displaying this error in the dev tools:

iAmWillShepherd: Would you mind updating the doc above with what the difference between `null` and `undefined` is?

PR #5208, opened 2018-07-20
ensure commit message is cleared after successful commit

shiftkey:
Fixes #4046 by moving the responsibility to cleanup higher in the component tree, and clearing the `AppStore` state (which flows back down to the commit form).

For those reviewing, I recommend using a very large repository after #5053 is merged because the problem is clear when it takes a long time for the app to make the commit, giving you more time to switch to the History tab and show it doesn't tidy up as expected.

PR #5260, opened 2018-07-26
fix regression with clearing commit message

shiftkey:
@iAmWillShepherd identified that the latest 1.3.0 beta doesn't clear the commit in the normal scenario because of #5208.

I dug into this and uncovered:

 - my assumption that `GitStore` was the source of truth for this commit message isn't correct
 - `GitStore` is only updated when the component is unmounted (due to switching tabs or switching repositories)
 - because `GitStore` keeps sending `null` message to the component in the normal scenario, the component ignores this in favour of it's local state

So we need to do both "clean up local state" and "clean up `GitStore` state" to cover both scenarios. Sigh.

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -722,7 +722,7 @@ export class GitStore extends BaseStore {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


joshaber:
giphy-43

PR #5263, opened 2018-07-26
update changelog generation for production release

shiftkey:
This PR adds some tweaks that I've made to the release notes generation for production releases.

It assumes:

 - all PRs on `master` were recently published to beta
   - this means we can skip scanning the changelog again to check for new entries
   - this was problematic because we'd edit items and remove entries, and this list would bring them back again
 - `-beta0` is the special beta upgrade to ensure the latest changes are in the beta channel, and can be skipped from future releases
   - not sure why we excluded `-beta1` instead 🤷‍♂️ 
 - only include changelog entries that match the expected format e.g. `[New]`
   - this ensures we exclude test releases, which are more relaxed and don't adhere to this format

### DIFF
--- a/script/draft-release/run.ts
+++ b/script/draft-release/run.ts
@@ -95,7 +95,6 @@ export async function run(args: ReadonlyArray<string>): Promise<void> {
     if (channel === 'production') {
       const existingChangelog = getChangelogEntriesSince(previousVersion)
       const entries = new Array<string>(
-        ...changelogEntries,
         ...existingChangelog
       )
       printInstructions(nextVersion, entries)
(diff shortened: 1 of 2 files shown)


domDavid555:
what now?

script/draft-release/run.ts
```diff
@@ -94,10 +94,7 @@ export async function run(args: ReadonlyArray<string>): Promise<void> {
 
     if (channel === 'production') {
       const existingChangelog = getChangelogEntriesSince(previousVersion)
-      const entries = new Array<string>(
-        ...changelogEntries,
-        ...existingChangelog
-      )
+      const entries = new Array<string>(...existingChangelog)
```
iAmWillShepherd: This can be condensed into `const entries = [...existingChangelog]`

shiftkey:
:apple:

PR #5271, opened 2018-07-27
add .tool-versions file

outofambit:
I like to use [asdf](https://github.com/asdf-vm/asdf) for version management. This file lets you

```shellsession
$ asdf install
```

to get the specififed version of node and python.

(is there anything else we should add to this file?)


shiftkey:
> (is there anything else we should add to this file?)

Not that I can think of :shipit:

PR #5272, opened 2018-07-27
Migrate to CircleCI 2.0

### LINKED ISSUE #4135: CircleCI v1 is being deprecated August 31, 2018
[Just announced today](https://circleci.com/blog/sunsetting-1-0/), our CircleCI builds will stop working  at the end of August, and they are encouraging everyone to migrate to the new [v2](https://circleci.com/docs/2.0/) definitions.

A good starting point for a reference config would probably be https://github.com/CircleCI-Public/circleci-demo-react-native/blob/master/.circleci/config.yml and this section: https://github.com/CircleCI-Public/circleci-demo-react-native/blob/master/.circleci/config.yml#L32-L56

vanessayuenn:
This PR migrates our CircleCI build from 1.0 to 2.0. The main change worth calling out is that since xcode 7.3 is no longer supported, we've moved to [another macOS container](https://circle-macos-docs.s3.amazonaws.com/image-manifest/build-405/index.html) that comes with xcode 9.3.0, node 9.10.0 (previously 8.11), and yarn 1.5.1 (no change).

Resolves #4135.


.circleci/config.yml
```diff
@@ -0,0 +1,56 @@
...
+    - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
+    - restore_cache:
+        keys:
+        # when lock file changes, use increasingly general patterns to restore cache
+        - yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
+        - yarn-packages-v1-{{ .Branch }}-
+        - yarn-packages-v1-
+    - run:
+        name: Install dependencies
+        command: yarn install --force
```
vanessayuenn: I kept the `--force` flag from the 1.0 config, but this essentially renders the caching using `yarn.lock` useless. @shiftkey do we want to keep this?

shiftkey: I'm fine with keeping this around. Reading the [docs](https://yarnpkg.com/en/docs/cli/install) again, I read the use of `--force` as being related to the downloading of packages, ignoring the cache, rather than ignoring the lockfile (`--no-lockfile`):




Let me know if you disagree.

circle.yml
```diff
@@ -1,48 +0,0 @@
...
-test:
-  override:
-    - yarn test:setup
-    - yarn test
-
-deployment:
-  production:
-    branch: /^__release-.*/
-    commands:
-      - yarn run publish
```
shiftkey: This step here hasn't been ported over to the new config format - any context on why that's the case?

script/setup-macos-keychain
```diff
@@ -11,3 +11,6 @@ security unlock-keychain -p travis $KEY_CHAIN
 security set-keychain-settings -t 3600 -u $KEY_CHAIN
 
 security default-keychain -s $KEY_CHAIN
+
+# set key partition list to avoid UI permission popup that causes hanging at CI
+security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k travis $KEY_CHAIN
```
shiftkey: :art: newline at end of file

.circleci/config.yml
vanessayuenn: I was under the impression that we're caching `node_modules` (line 32 just below) but `--force` would reinstall all deps even if they're already in `node_modules`.

shiftkey: You are correct, and I think we brought this in flag because we had problems building when the Electron version would change underneath (caching `node_modules` meant the build would break with an obscure `EEXISTS` error). I'll dig into this on Monday and report back.

circle.yml
vanessayuenn: The context is that I didn't drink enough coffee and have missed this step 😅

.circleci/config.yml
vanessayuenn: Okay! If we end up deciding `--force` flag is necessary to keep, I'd skip caching for `node_modules` since it's an expensive operation and won't be doing anything.

shiftkey: :thumbsup: caching this folder requires rehydrating the folder for each build, which is wasted cycles if we ignore the cache anyway

shiftkey: For the moment I want to keep this old behaviour around, just in case it's helpful for us down the track. I had a chat with @shana a few days ago about [offline mirroring into the repository](https://yarnpkg.com/blog/2016/11/24/offline-mirror/) which commits the packages (not the `node_modules` contents) and would be a great speedup for CI scenarios like this.

I'll open an issue to track this work once I've done some initial scouting.

shiftkey:
One last thing for me to test before I merge this - ensuring I can a test release from the new CircleCI config.

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+      CIRCLE_TEST_REPORTS: /tmp/circleci-test-results
+    macos:
+      xcode: '9.3.0'
+    steps:
+    - checkout
+    - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
+    - restore_cache:
+        keys:
+        # when lock file changes, use increasingly general patterns to restore cache
+        - yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
streeter: Chiming in here. Would you want to also add `-{{ checksum "app/yarn.lock" }}` and then also add `app/node_modules` to the cache?

.circleci/config.yml
```diff
@@ -0,0 +1,79 @@
...
+        command: yarn install --force
+    - save_cache:
+        key: yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}
+        paths:
+        - ~/.cache/yarn
+        - vendor/bundle
+        - .eslintcache
+        - ~/.electron
+        - ~/Library/Caches/Yarn/v1
+        - ./node_modules
```
streeter: Going with [my comment above](https://github.com/desktop/desktop/pull/5272/files#r207064438), I think adding `./app/node_modules` would help by caching that folder as well.

.circleci/config.yml
shiftkey: Thanks for the suggestion. The `postinstall` step will update the `app` with a `yarn install --force`, so caching here is not useful. I've opened #5318 to look into what using an offline cache might look like, and whether we can get it working faster than our current setup.

shiftkey:
Looking good:



Given this bumps us to a newer version of XCode, I just want to confirm the test installer works as expected before we close this out.

PR #5297, opened 2018-07-31
fix for missing .git directory from a tracked repository

### LINKED ISSUE #5291: Getting error (Git returned an unexpected exit code '128') after updating to v1.3.1
## Description
Few minutes ago I've updated to v1.3.1 and after opening GitHub Desktop (everytime) I get this error pops up.

## Version
* GitHub Desktop: 1.3.1
* Operating system: Windows 10 x64

## Steps to Reproduce
1. Update to v1.3.1
2. Open Desktop. You get error.
3. Close and re-open Desktop.
4. You will get this error again.

You will also get this error if you keep GitHub Desktop open. Probably a background task.

### Expected Behavior
No error should raised.

### Actual Behavior
`Git returned an unexpected exit code '128' which should be handled by the caller.'`

## Additional Information
No additional info

### Logs
I could attach log here but I investigated it and no error or similar messages but infos available. I avoid uploading all logs here due to some details about my private repositories in it.

**Update:** This evening I keep GitHub Desktop open in the background. After 5-6 hours, when I maximize it, I see error message already displayed. I just clicked `Close` however, another one pops up. I tried nearly hundred times and it seems GitHub Desktop does this in a loop in the background, so I got probably a thousand of errors. I had to close manually from task manager. Anyway, I just watch my GitHub Desktop today and it's indeed this error pops up every few minutes.

shiftkey:
Fixes #5291 by:

 - reordering the checks for updating the repository indicators because of this call stack:

This code path shows that `withAuthenticatingUser` can indirectly call `getRemotes` (which requires a valid Git repository) before we do a `getStatus`, which causes the error initially reported.

 - teach `getStatus` about exit code 128

By reordering the checks so that `git status` fires before `git fetch` we now also need to check for exit code 128 there. This required returning the exit code from the spawned process response, and handling that correctly. I've added some debug logging to this code path so it doesn't pollute the file logs excessively.

### DIFF
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -75,7 +75,7 @@ export async function getStatus(
     log.debug(
       `'git status' returned 128 for '${
         repository.path
-      }' and is likely missing it's .git directory`
+      }' and is likely missing its .git directory`
     )
     return null
   }
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -67,7 +67,8 @@ export async function getStatus(
   const result = await spawnAndComplete(
     ['status', '--untracked-files=all', '--branch', '--porcelain=2', '-z'],
     repository.path,
-    'getStatus'
+    'getStatus',
+    new Set<number>([0, 128])
   )
 
   if (result.output.length > MaxStatusBufferSize) {
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -71,6 +71,15 @@ export async function getStatus(
     new Set<number>([0, 128])
   )
 
+  if (result.exitCode === 128) {
+    log.debug(
+      `'git status' returned 128 for '${
+        repository.path
+      }' and is likely missing it's .git directory`
+    )
+    return null
+  }
+
   if (result.output.length > MaxStatusBufferSize) {
     log.error(
       `'git status' emitted ${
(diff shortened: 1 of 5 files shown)


app/src/lib/git/status.ts
```diff
@@ -67,9 +67,19 @@ export async function getStatus(
   const result = await spawnAndComplete(
     ['status', '--untracked-files=all', '--branch', '--porcelain=2', '-z'],
     repository.path,
-    'getStatus'
+    'getStatus',
+    new Set<number>([0, 128])
```
j-f1: Is TS smart enough to infer that this is `<number>` without specifying it directly?

shiftkey: Actually, yes. Just a habit of mine. Will tidy this up :art:

app/test/unit/git/status-test.ts
```diff
@@ -103,5 +105,13 @@ describe('git/status', () => {
       const files = status.workingDirectory.files
       expect(files.length).to.equal(numFiles)
     })
+
+    it('returns null for directory without a .git directory', async () => {
+      const emptyDir = mkdirSync('no-repo-here')
```
outofambit: same fixture feedback as #5298 :)

app/src/lib/stores/app-store.ts
```diff
@@ -1918,37 +1918,45 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+    const exists = await pathExists(repository.path)
+    if (!exists) {
+      lookup.delete(repository.id)
+      return
+    }
+
+    const gitStore = this.getGitStore(repository)
+
+    const status = await gitStore.loadStatus()
```
outofambit: 💯

app/test/unit/git/status-test.ts
shiftkey: Done in 940cc9341

PR #5298, opened 2018-07-31
teach getRemotes about handling this error code instead of failing fast

shiftkey:
This is related to exercising #5297 and the steps to reproduce this are roughly:

 - add a repository to Desktop
 - switch to a different repository in the list
 - rename the `.git` directory to something else
 - open Desktop again
 - change to the problem repository
 - :boom:

Inside [`AppStore._refreshRepository`](https://github.com/desktop/desktop/blob/7fc541877d02c0245ebcf4c791bbfce32a4c6fad/app/src/lib/stores/app-store.ts#L1864) there are a bunch of Git operations that assume the `missing` flag is in the right state when it runs.

My testing shows these two operations did not handle this "folder exists but it's not a Git repository" scenario:

 - [`GitStore.loadBranches()`](https://github.com/desktop/desktop/blob/7fc541877d02c0245ebcf4c791bbfce32a4c6fad/app/src/lib/stores/app-store.ts#L1875)
 - [`GitStore.loadRemotes()`](https://github.com/desktop/desktop/blob/7fc541877d02c0245ebcf4c791bbfce32a4c6fad/app/src/lib/stores/app-store.ts#L1892)

I'm going to address the first one in a separate up PR because the app is currently in a loop and doesn't set the repository to `missing=true` , but this fix here was easier to address because:

 - 128 is the generic error code for "no Git repository here, yo"
 - if this error code is encountered, we can just return an empty array of remotes

### DIFF
--- a/app/src/lib/git/remote.ts
+++ b/app/src/lib/git/remote.ts
@@ -7,6 +7,9 @@ import { findDefaultRemote } from '../stores/helpers/find-default-remote'
 export async function getRemotes(
   repository: Repository
 ): Promise<ReadonlyArray<IRemote>> {
+  // TODO: use expectedErrors here to handle a specific error
+  // see https://github.com/desktop/desktop/pull/5299#discussion_r206603442 for
+  // discussion about what needs to change
   const result = await git(['remote', '-v'], repository.path, 'getRemotes', {
     successExitCodes: new Set([0, 128]),
   })
--- a/app/src/lib/git/remote.ts
+++ b/app/src/lib/git/remote.ts
@@ -8,7 +8,7 @@ export async function getRemotes(
   repository: Repository
 ): Promise<ReadonlyArray<IRemote>> {
   const result = await git(['remote', '-v'], repository.path, 'getRemotes', {
-    successExitCodes: new Set<number>([0, 128]),
+    successExitCodes: new Set([0, 128]),
   })
 
   if (result.exitCode === 128) {
--- a/app/src/lib/git/remote.ts
+++ b/app/src/lib/git/remote.ts
@@ -7,7 +7,14 @@ import { findDefaultRemote } from '../stores/helpers/find-default-remote'
 export async function getRemotes(
   repository: Repository
 ): Promise<ReadonlyArray<IRemote>> {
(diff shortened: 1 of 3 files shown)


app/src/lib/git/remote.ts
```diff
@@ -7,7 +7,14 @@ import { findDefaultRemote } from '../stores/helpers/find-default-remote'
 export async function getRemotes(
   repository: Repository
 ): Promise<ReadonlyArray<IRemote>> {
-  const result = await git(['remote', '-v'], repository.path, 'getRemotes')
+  const result = await git(['remote', '-v'], repository.path, 'getRemotes', {
+    successExitCodes: new Set<number>([0, 128]),
```
billygriffin: Is this `<number>` also unnecessary?

app/src/lib/git/remote.ts
```diff
@@ -7,7 +7,14 @@ import { findDefaultRemote } from '../stores/helpers/find-default-remote'
 export async function getRemotes(
   repository: Repository
 ): Promise<ReadonlyArray<IRemote>> {
-  const result = await git(['remote', '-v'], repository.path, 'getRemotes')
+  const result = await git(['remote', '-v'], repository.path, 'getRemotes', {
+    successExitCodes: new Set([0, 128]),
```
outofambit: same feedback as in #5299 regarding `expectedErrors`

app/test/unit/git/remote-test.ts
```diff
@@ -32,6 +33,14 @@ describe('git/remote', () => {
       expect(result[1].name).to.equal('origin')
       expect(result[1].url.endsWith(nwo)).to.equal(true)
     })
+
+    it('returns empty array for directory without a .git directory', async () => {
+      const emptyDir = mkdirSync('no-repo-here')
+      const repository = new Repository(emptyDir, -1, null, false)
```
outofambit: should we make a fixture for this in the vein of `setupEmptyRepository`?
also same thoughts from #5299 about `before` apply

shiftkey: :thumbsup: to helper function for creating the fixture

shiftkey:
:apple:

PR #5299, opened 2018-07-31
teach getBranches to exit early if folder isn't a Git repository

shiftkey:
In the similar vein as #5298, `getBranches` is happily run against a folder that doesn't contain a `.git` directory. For safety's sake, we can just exit early and return an empty array of branches.

### DIFF
--- a/app/src/lib/git/for-each-ref.ts
+++ b/app/src/lib/git/for-each-ref.ts
@@ -41,8 +41,14 @@ export async function getBranches(
   const result = await git(
     ['for-each-ref', `--format=${format}`, ...prefixes],
     repository.path,
-    'getBranches'
+    'getBranches',
+    { successExitCodes: new Set([0, 128]) }
   )
+
+  if (result.exitCode === 128) {
+    return []
+  }
+
   const names = result.stdout
   const lines = names.split(delimiterString)
 
--- a/app/src/lib/git/for-each-ref.ts
+++ b/app/src/lib/git/for-each-ref.ts
@@ -38,6 +38,9 @@ export async function getBranches(
     prefixes = ['refs/heads', 'refs/remotes']
   }
 
+  // TODO: use expectedErrors here to handle a specific error
+  // see https://github.com/desktop/desktop/pull/5299#discussion_r206603442 for
+  // discussion about what needs to change
   const result = await git(
     ['for-each-ref', `--format=${format}`, ...prefixes],
     repository.path,
--- a/app/test/unit/git/for-each-ref-test.ts
+++ b/app/test/unit/git/for-each-ref-test.ts
@@ -3,6 +3,7 @@ import { Repository } from '../../../src/models/repository'
 import {
   setupFixtureRepository,
   setupEmptyRepository,
+  mkdirSync,
 } from '../../helpers/repositories'
 import { getBranches } from '../../../src/lib/git/for-each-ref'
 import { BranchType } from '../../../src/models/branch'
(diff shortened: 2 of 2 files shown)


app/test/unit/git/for-each-ref-test.ts
```diff
@@ -58,5 +59,13 @@ describe('git/for-each-ref', () => {
       const branches = await getBranches(repo)
       expect(branches.length).to.equal(0)
     })
+
+    it('should return empty list for directory without a .git directory', async () => {
+      const emptyDir = mkdirSync('no-repo-here')
```
outofambit: I would prefer this setup be done in a `before` block instead of in the `it` test case

app/src/lib/git/for-each-ref.ts
```diff
@@ -41,8 +41,14 @@ export async function getBranches(
   const result = await git(
     ['for-each-ref', `--format=${format}`, ...prefixes],
     repository.path,
-    'getBranches'
+    'getBranches',
+    { successExitCodes: new Set([0, 128]) }
```
outofambit: what if we use the `expectedErrors` option here to indicate more clearly that this is not a success case but an error we want to handle specially?

shiftkey: I wanted to use `expectedErrors` here, like this:

```ts
    ...
    'getBranches',
    expectedErrors: new Set([GitError.NotAGitRepository]),
  })
```

But upstream Git fixed this error message in https://github.com/git-for-windows/git/commit/fc045fe7d4fa220f19274715c936636fe0516ea3 which is in `2.17.1`:

```
fatal: Not a git repository (or any of the parent directories): .git
```

to this:

```
fatal: not a git repository (or any of the parent directories): .git
```

So I need to publish a new version of `dugite` to address this - https://github.com/desktop/dugite/pull/199 is the upstream issue.

I'll leave a note here to indicate that this is some future cleanup work, but I'm holding off on upgrading Git to 2.18.0 until I can do some more thorough testing.

app/test/unit/git/for-each-ref-test.ts
shiftkey: This is likely a side effect of how our tests have built up over time, but I've favoured `before(Each)` and `after(Each)` setup and teardown when it's applied to a group of tests to eliminate the repetition. 

For a single test like this, it feels like a lot of ceremony - particularly with how mocha structures things (and TS deals with things defined outside scope). I think I'd have to have something like this:

```ts
describe('directory without a .git directory', function() {
  // because I need to set and read this from two different functions
  // this signature is required and makes me feel bad
  let repository: Repository | null = null

  before(function() {
    const emptyDir = mkDirSync('no-repo-here')
    repository = new Repository(emptyDir, -1, null, false)
  });

  it('should return empty list', async () => {
      // gotta ! the value here to tell tsc that it's fiiiiine
      const status = await getRemotes(repository!)
      expect(status).eql([])
  })
});
```

For a single test like this, I might split it up into [AAA](http://wiki.c2.com/?ArrangeActAssert) groupings if there's a lot of setup required, but even that is a flag for a potential code smell.

shiftkey:
:apple:

app/test/unit/git/for-each-ref-test.ts
outofambit: that's fair! (i tend to prefer the `before` even for single test cases because it more clearly delineates setup from test but no big deal in this case)

shiftkey: Yeah, if you feel like grabbing some existing tests and giving them some love we can keep this discussion moving.

PR #5318, opened 2018-08-01
move our yarn usage to offline mode 

shiftkey:
This is an experimental PR started by @shana getting in my ear a few days ago about enabling yarn to cache packages locally.

This has a few benefits:

 - CI builds will have the packages available for each build, which should help speed things up
 - CI will only use packages available in it's cache, which makes things easier to audit
 - development will only use packages available locally (✈️ coding yay)
 - changes to packages needed in the app become more visible in the review process
 
I'm also upgrading yarn during this process as there are a few fixes in the queue around offline packages.

 - [x] figure out what the CI errors are all about
 - [x] update `docs/contributing/working-with-packages.md` to detail workflow changes
 - [ ] compare performance changes by switching to this model (no-cache versus caching in config versus in-repository cache)
 - [ ] update `script/install-arm64-deps.sh` as that uses `--force`


shiftkey:
Apparently this whole process requires regenerating the lockfile(s), which is probably why there's that webpack errror introduced to the process. I wonder if I can avoid that process while still caching here 🤔🤔🤔

app/.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+yarn-offline-mirror "./vendor/cache"
```
j-f1: Should this be changed to `../vendor/cache` so any packages with the same version aren’t duplicated?

shiftkey: Sadly no. I initially had this as just a root `.yarnrc` and what would happen was roughly this:

 - `yarn install` runs against `package.json`
 - when complete, unneeded packages _for `package.json`_ would be removed
 - `cd app && yarn install` runs against `app/package.json`
 - when complete, unneeded packages _for `app/package.json`_ would be removed

So you'd end up in a state where whoever ran last would have their packages in the cache, and the other will be missing some packages it needs (and `yarn install --offline` would error). Just seems to be a design limitation that we can't workaround currently.

shiftkey:
Another reason to do this is that the registry sometimes has Problems™ which introduce false CI failures:

```
2018-08-07T17:56:37.8469498Z Script contents:
2018-08-07T17:56:37.8477687Z yarn install --force
2018-08-07T17:56:37.9360898Z ##[command]"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\65570b08-bed1-4272-a9c3-0563840f512f.cmd""
2018-08-07T17:56:38.5038764Z yarn install v1.5.1
2018-08-07T17:56:38.7453459Z [1/5] Validating package.json...
2018-08-07T17:56:38.7478311Z [2/5] Resolving packages...
2018-08-07T17:56:39.6026596Z [3/5] Fetching packages...
2018-08-07T17:56:58.7030708Z error An unexpected error occurred: "https://registry.yarnpkg.com/eslint-scope/-/eslint-scope-4.0.0.tgz: Request failed \"500 Internal Server Error\"".
2018-08-07T17:56:58.7032483Z info If you think this is a bug, please open a bug report with the information provided in "D:\\a\\1\\s\\yarn-error.log".
2018-08-07T17:56:58.7033087Z info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
2018-08-07T17:57:50.3473578Z ##[error]Cmd.exe exited with code '1'.
2018-08-07T17:57:50.4236432Z ##[section]Finishing: Install
```

shiftkey:
Closing this out for now to cleanup low-priority PRs, but I plan to revisit this later...

Issue #1252, opened 2017-04-20
Investigate whether we can short-circuit fetch


niik:
This is an idea that stems from Desktop Classic that we discussed but never investigated.

From https://github.com/github/availability/pull/357#issuecomment-130076518

>> Did you think about some cheaper way to check if there has been any traffic on a repo? For example, repo reference updates could be assigned some kind of unique ID, and your inquiry could be "get the latest push ID".
>
> We do store a pushed_at field on every repository that is updated with each push. [...]
>
> We expose the pushed_at timestamp in the API with repository get.
> https://developer.github.com/v3/repos/#get

If this is reliable and works well we could use that to avoid doing a fetch at all for background updates unless something has changed on the remote. This could spare us a great deal of network traffic in some scenarios as git fetch always retrieves a list of _all_ references (including our "hidden" pull refs) and _then_ matches those against the local refspec (run `git ls-remote` for an idea of what that mean).

For a large repository with a massive amount of pull requests this can actually add up. A "noop" atom/atom fetch takes a little over 2 seconds on my machine and pulls down 341kb of refs. Getting the pushed_at using the API on the other hand takes 0.6 seconds and pulls down 2k of data.

This would apply to GitHub repositories only.

niik:
> If this is reliable and works well we could use that to avoid doing a fetch at all for background updates unless something has changed on the remote.

I took a quick look at this and it does seem to work reliably for git pushes and web flow commits. It doesn't, however, update when a new PR is created from a fork, targeting the repository in question. For our use case I don't think that's necessary since we don't fetch those refs anyway but it's worth making a note of.

shiftkey:
> Getting the pushed_at using the API on the other hand takes 0.6 seconds and pulls down 2k of data.

It's about the same time with GraphQL but can be trimmed down to ~60 bytes of output:

<img width="933" src="https://user-images.githubusercontent.com/359239/35774584-f79d9c7c-09c7-11e8-960e-5e9fa16a7008.png">

j-f1:
With aliases, you can decrease it even more:

```gql
query {
  _: repository(owner: "desktop", name: "desktop") {
    _: updatedAt
  }
}
```

PR #5320, opened 2018-08-02
Check API for updates before fetching 

### LINKED ISSUE #5312: reduce overhead of performing repository indicator updates
#5310 highlighted that the work to compute the new repository indicators doesn't scale to 100+ repositories. I'd like to review this feature and see if we can address these two goals:

 - avoid running the work concurrently (this background work shouldn't steal CPU time from the app or other tools)
 - run this less frequently (defer the first run until after the app has settled, and run it less frequently in the background)

iAmWillShepherd:
Fixes #5312 

Related discussion: #1252


app/src/lib/stores/app-store.ts
```diff
@@ -1968,7 +1971,42 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
+    const account = getAccountForRepository(this.accounts, repository)
```
shiftkey: :art: Something to keep in mind as you're working through this - I'd love for this logic to be reusable so we can eventually use this in `BackgroundFetcher`

shiftkey (commented):
I like where this is heading, but before I go and try and break this code here's some initial feedback

parvaizmuhd:
Bad performamce

iAmWillShepherd:
💯 

Tested with my local repos. 

Note, I had to modify the following locally to test:

https://github.com/desktop/desktop/blob/7e17011122e5085a4c94e2bba94f9fad3ac6ade8/app/src/lib/stores/app-store.ts#L1938-L1945

Updated that number to 25 (I have 21 local repos)

https://github.com/desktop/desktop/blob/7e17011122e5085a4c94e2bba94f9fad3ac6ade8/app/src/ui/app.tsx#L98

Updated to refresh every minute.

app/src/lib/stores/app-store.ts
```diff
@@ -1961,15 +1980,19 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
     }
 
     this.withAuthenticatingUser(repository, async (repo, account) => {
-      if (this.shouldBackgroundFetch(repo)) {
+      const lastPush = await this.inferLastPushForRepository(
+        this.accounts,
+        gitStore,
+        repository
+      )
+      if (this.shouldBackgroundFetch(repo, lastPush)) {
```
shiftkey: There's some extra work that we need to do in here for these dormant repositories:

 - `gitStore.updateLastFetched()` - this pokes at the `.git/FETCH_HEAD` file on disk
 - `gitStore.loadRemotes()` - this hydrates the remotes information from the repository

If we skip the first item, the `lastFetched` value is always empty when we need to compare it inside `this.shouldBackgroundFetch`. If we skip the second, the inner `gitStore.fetch` is no-op because it doesn't have any remotes to work with.

I think we should move `this.shouldBackgroundFetch` out of `this.withAuthenticatingUser` - remember the large call stack from https://github.com/desktop/desktop/pull/5297#issue-205091781 - because we only need to do this work when it's going to fetch. Here's a diff of how I think this would look:

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index a739b0e48..4ca552eb2 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1986,18 +1986,23 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
-    this.withAuthenticatingUser(repository, async (repo, account) => {
-      const lastPush = await this.inferLastPushForRepository(
-        this.accounts,
-        gitStore,
-        repository
-      )
-      if (this.shouldBackgroundFetch(repo, lastPush)) {
-        await gitStore.performFailableOperation(() => {
+    const lastPush = await this.inferLastPushForRepository(
+      this.accounts,
+      gitStore,
+      repository
+    )
+
+    await gitStore.updateLastFetched()
+
+    if (this.shouldBackgroundFetch(repository, lastPush)) {
+      await gitStore.loadRemotes()
+
+      await this.withAuthenticatingUser(repository, async (repo, account) => {
+        return gitStore.performFailableOperation(() => {
           return gitStore.fetch(account, true)
         })
-      }
-    })
+      })
+    }
 
     lookup.set(repository.id, {
       aheadBehind: gitStore.aheadBehind,
```

Also ensure you `await this.withAuthenticatingUser` because that ensures we keep this running sequentially.

shiftkey (changes requested):

app/src/lib/stores/app-store.ts
```diff
@@ -3992,6 +4005,58 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        const repo = await api.fetchRepository(owner, name)
+
+        if (repo !== null) {
+          return new Date(repo.pushed_at)
+        }
+      }
+    }
+
+    if (repository.gitHubRepository !== null) {
+      const api = API.fromAccount(account)
```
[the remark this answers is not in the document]
iAmWillShepherd: Surprised TSC didn't catch that one. I guess a new block scope let's you do that... ¯\_(ツ)_/¯

app/src/lib/stores/app-store.ts
```diff
@@ -1321,7 +1324,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return false
     }
 
-    return true
+    if (lastPush === null) {
+      return true
+    }
+
+    // we should fetch if the last push happened after the last fetch
+    const shouldFetch = lastFetched < lastPush
```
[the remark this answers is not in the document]
iAmWillShepherd: Actually kept it that way for an easier time debugging, but I'm not married to keeping it that way.

app/src/lib/stores/app-store.ts
```diff
@@ -1350,7 +1372,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       repository,
       account,
       r => this.performFetch(r, account, FetchType.BackgroundTask),
-      r => this.shouldBackgroundFetch(r)
+      r => this.shouldBackgroundFetch(r, null)
```
shiftkey: Can we add a `TODO` in here to follow up on this? I think we should make the background fetcher follow the same rules as refreshing the indicators.

shiftkey (changes requested):

app/src/lib/stores/app-store.ts
iAmWillShepherd: Are you referring to the limits we put in place? or checking the API before fetching?

app/src/lib/stores/app-store.ts
```diff
@@ -1301,27 +1301,46 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
 
-    if (!lastFetched) {
+    if (lastFetched === null) {
       return true
     }
 
     const now = new Date()
     const timeSinceFetch = now.getTime() - lastFetched.getTime()
-
+    const repoName =
```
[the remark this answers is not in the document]
shiftkey: #5327 has been merged

app/src/lib/stores/app-store.ts
shiftkey: Checking the API before fetching

iAmWillShepherd:
💯

shiftkey (approved):

shiftkey:
Going to merge this in early because I want to build upon these changes 🏎

Shohag1985 (commented):
app/src/lib/stores/app-store.ts

PR #5326, opened 2018-08-02
cache new tip for fresh repositories

shiftkey:
If you have a lot of tracked repositories and disable #5311, as part of each background refresh you might see a lot of activity like this:

If you can't spot the pattern, here it is:

After each `git status` there's 2 Git operations to load the tip commit into the app, but we didn't keep it around, so each time it would rehydrate this from the Git repository. We didn't notice this for the active repository because we'd load the commits for the repository when it becomes active - these dormant repositories haven't had a chance to do that yet.

Commits can be keyed by their ID, so this adds the new tip commit to the cache if it's found - otherwise we get the same old error flow. The changes are a bit janky because we need to handle `undefined` and `null`, but TS happily shouted at me until I got it right.


app/src/lib/stores/git-store.ts
```diff
@@ -891,15 +891,18 @@ export class GitStore extends BaseStore {
...
+            (await this.performFailableOperation(() =>
+              getCommit(this.repository, currentTip)
+            )) || null
+
+          if (branchTipCommit != null) {
+            this.commitLookup.set(currentTip, branchTipCommit)
+          } else {
+            throw new Error(`Could not load commit ${currentTip}`)
+          }
         }
```
j-f1: Could this section become just:

```ts
let branchTipCommit =
  this.commitLookup.get(currentTip) ||
  (await this.performFailableOperation(() =>
    getCommit(this.repository, currentTip)
  )) ||
  null;

if (branchTipCommit == null) {
  throw new Error(`Could not load commit ${currentTip}`);
} else {
  this.commitLookup.set(currentTip, branchTipCommit);
}
```

shiftkey: Maybe? I'm just over here quietly crying about having to deal with `null` and `undefined` here but I might have to suck it up and always call `this.commitLookup.set` (should be safe).

app/src/lib/stores/git-store.ts
```diff
@@ -923,6 +914,30 @@ export class GitStore extends BaseStore {
...
+  /**
+   * Find a commit in the local cache, or load in the commit from the underlying
+   * repository.
+   *
+   * This will error if the commit ID cannot be resolved.
+   */
+  private async lookupCommit(sha: string): Promise<Commit> {
+    const cachedCommit = this.commitLookup.get(sha)
+    if (cachedCommit != null) {
+      return Promise.resolve(cachedCommit)
```
shiftkey: I'm doing this here because I remember something about returning before the `await` being problematic, but that might have been before Chrome supported native promises, so this might be unnecessary. Gonna leave this here for safety.

j-f1: I’m pretty sure this will be fine.

PR #5335, opened 2018-08-03
upgrade CI providers to local yarn usage

shiftkey:
A significant source of CI failure is the pre-requisite steps to install Yarn:

```yaml
before_install:
  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.5.1
  - export PATH=$HOME/.yarn/bin:$PATH
```

And with all the CI providers having different versions of Yarn installed by default, it's problematic ensuring we're using the same version across all environments.

This PR works through the CI platforms to upgrade us to Yarn 1.9.x and use the trick from [this blog post](https://yarnpkg.com/blog/2016/11/24/offline-mirror/) so we don't need to install or change CI configs:

Unlike the blog post, I'm going to move it into `vendor` to indicate that it shouldn't be touched as part of normal use. This won't be embedded into our `package.json` usages, but I'll upgrade the `engines` array in `package.json` to ensure we're keeping up locally.

 - [x] add `yarn` source package
 - [x] use `.yarnrc` to point to source `yarn`
 - [x] workaround issue with `yarn` inside NPM scripts
 - [x] confirm all builds are using latest version
 - [x] can cleanup manual installs of yarn
 - [x] upgrade `engines` entry in `package.json` to ensure people are using a recent version


.travis.yml
```diff
@@ -49,15 +46,22 @@ cache:
...
+  - npm_config_clang=1
 
 install:
-  - yarn install --force
+  - node ./vendor/yarn-$YARN_VERSION.js install --force
 
 script:
-  - yarn lint && yarn validate-changelog && yarn build:prod && yarn test:setup && yarn test
+  - |
+    node ./vendor/yarn-$YARN_VERSION.js lint &&
```
j-f1: Does this version of Yarn work properly when subcommands (like this one) run `yarn`?

shiftkey: This is what I'm digging into now. [`.yarnrc`](https://yarnpkg.com/lang/en/docs/yarnrc/) supports `local-path` if you want to override the global yarn with a local version, and Travis seems to be working fine.



I want to verify it across all the platforms, hence doing it one at a time.

shiftkey: Actually, it doesn't seem to be currently. Sigh.

script/post-install.ts
```diff
@@ -0,0 +1,63 @@
...
+  glob('vendor/yarn-*.js', (error, files) => {
+    if (error != null) {
+      throw error
+    }
+
+    // this ensures the paths returned by glob are sorted alphabetically
+    files.sort()
+
+    // use the latest version here if multiple are found
+    const latestVersion = files[files.length - 1]
```
iAmWillShepherd: If `files` is being sorted lexicographically, there is a chance `latestVersion` won't actually be the latest version.

shiftkey: I'm not quite sure how these are sorted, just that `glob`  mentions that they are sorted by default.

Given it's a manual process to change the vendored version of yarn, I'm not sure how defensive we need to be in here to handle multiple yarn packages being present on disk. We can guard against this by using the `engines` property in `package.json`, to ensure a minimum version is used.

What's your gut feeling on this?

iAmWillShepherd: I say we leave it since we have to update yarn manually.

PR #5338, opened 2018-08-03
Update the node version noted in setup.md

erin-lee:
Just updating version to match that of `.node-version`


iAmWillShepherd:
Thank you for your contribution! 🙏

Issue #3814, opened 2018-01-18
lock the file/line selection controls while a commit is running


callumlocke:
### Description

I'm working on a large project that has a slow pre-commit hook (> 10 seconds). Several times, I've used the file checkboxes and individual line selection to do a partial commit, and then without realising the commit is still running, I've started selecting more lines for a second commit. Then I realise the first commit is still running, and it's hard to tell from the UI what is going to get committed.

### Version

**GitHub Desktop version:** 1.0.11

**OS version:** macOS Sierra (10.12.6)

### Steps to Reproduce

1. In a project with a slow pre-commit hook, do a partial commit (just select certain files and/or lines, and then click Commit)
1. While that partial commit is running, try to change the selection of lines/files.

**Expected behavior:**

You shouldn't be able to change the file/line selection while the commit is running.

**Actual behavior:**

You can mess with the file/line selection before the commit is complete, so the UI is now in a state that does not reflect what is really going to get committed.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%.

### Additional information

I'm guessing that the selection is fixed from the moment the commit button is clicked, but if this is the case it would be good if the UI could reflect this. To avoid confusion, I suggest disabling all the checkboxes and line selection controls as soon as the commit button is clicked, and re-enabling them once the commit has succeeded or failed.

bangly631:
![snapshot_20180117](https://user-images.githubusercontent.com/33788088/35100654-d35f30d8-fcc9-11e7-9eef-a547cc417661.JPG)

bangly631:
![snapshot_20180117](https://user-images.githubusercontent.com/33788088/35100680-f46c03b4-fcc9-11e7-8adb-2316ac8ce952.JPG)
![snapshot_20180117](https://user-images.githubusercontent.com/33788088/35100675-edf72a4a-fcc9-11e7-958b-aa802044699e.JPG)

shiftkey:
@callumlocke this is an interesting idea, and something that probably needs some design input to ensure the lock/unlock transition isn't overwhelming.

For reference, we're already tracking the `isCommitting` state at the repository level (so we can disable the commit form), and flowing that down into the other components so they can enable/disable themselves is probably worth exploring:

https://github.com/desktop/desktop/blob/6df8a6906c18b70af916039f2e1f38ce0fe98ad1/app/src/ui/repository.tsx#L96

iAmWillShepherd:
What would happen if the commit fails? Should the file remain locked?

[...]

PR #5341, opened 2018-08-03
prevent user from toggling files when commit is in progress

masungwon:
(@erin-lee and I worked on this PR together!)

Prevents changing the file/line selection while a commit is in progress.

Related to https://github.com/desktop/desktop/issues/3814

### DIFF
--- a/app/src/ui/changes/changed-file.tsx
+++ b/app/src/ui/changes/changed-file.tsx
@@ -12,6 +12,7 @@ interface IChangedFileProps {
   readonly oldPath?: string
   readonly include: boolean | null
   readonly availableWidth: number
+  readonly disabled: boolean
   readonly onIncludeChanged: (path: string, include: boolean) => void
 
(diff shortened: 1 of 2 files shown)


app/src/ui/changes/changed-file.tsx
```diff
@@ -12,6 +12,7 @@ interface IChangedFileProps {
   readonly oldPath?: string
   readonly include: boolean | null
   readonly availableWidth: number
+  readonly disabled: boolean
```
iAmWillShepherd: This can be a bit more descriptive, maybe something along the lines of `disableSelection`

shiftkey:
@masungwon @erin-lee this is a great fix for the checkboxes, and I won't hold this up for the other part of the issue - the partial selections when viewing the diff on a file. I'll add some details about this to the original issue, in case you have time to dig into that.

masungwon:
@shiftkey @iAmWillShepherd Thanks for your feedback!

Issue #4721, opened 2018-05-17
Menu item label is cut off on Windows


iAmWillShepherd:
I noticed this while testing a PR on Windows.

<img width="272" alt="screen shot 2018-05-17 at 11 04 27 am" src="https://user-images.githubusercontent.com/1715082/40189559-5e31bb68-59c2-11e8-9e86-b2e29df03a49.png">

Notice that _Clone Repository..._ has the letter _y_ missing. This is likely due to the keyboard shortcut taking up a ton of space. I'm not sure what the path forward is, but we could start using _repo_ instead of _repository_.

shiftkey:
> This is likely due to the keyboard shortcut taking up a ton of space.

Also note that the `...` is required for this menu shortcut, because the menu item here is supposed to open a new dialog.

> I'm not sure what the path forward is, but we could start using repo instead of repository.

I'm 👎 on this for historical reasons (I think we've made this same mistake before) and being consistent with the rest of the GitHub ecosystem.

iAmWillShepherd:
> Also note that the ... is required for this menu shortcut, because the menu item here is supposed to open a new dialog.

I should have been more clear. I was not advocating we remove the ellipsis. But I still feel like we should show the entire word.

Citing consistency, there menu item above has the entire word, so maybe this one should too?

shiftkey:
Moving this `enhancement` as something for someone with a Windows setup to dig into.

j-f1:
@shiftkey it should also be possible for someone on another platform to lie to Desktop — I did it before with my PR for the tooltip on the minimize button and it worked fairly well.

shiftkey:
@j-f1 can you share how you hacked around it? It'd be neat to figure out how easy that would be to switch between OSes, but I worry about not quite testing the right setup using shortcuts like this (like you can't run the macOS menu on Windows, and might actually :boom: the app I suspect)

iAmWillShepherd:
Fixed in #5348

PR #5348, opened 2018-08-07
Fix for file menu pane cut off on Windows

Daniel-McCarthy:
This change is in response to the problem mentioned in Issue #4721. In the comments for that issue it  is suggested that the text should remain the same and that the menu pane should be made wider to fit each option.

To fix this I have increased the width of the file menu pane by 5 pixels. In my tests this is the absolute minimum amount for "Clone Repository..." to show in full and not be cut off on Windows 10.
Currently I am not able to test this in any other Windows OS.

In case this is not the ideal solution to this problem I would be happy to make changes.

desktop3

### DIFF
--- a/app/styles/ui/_app-menu.scss
+++ b/app/styles/ui/_app-menu.scss
@@ -26,7 +26,7 @@
   }
 
   &.menu-pane-file {
-    width: 200px;
+    width: 205px;
   }
 
   // Open panes (except the first one) should have a border on their


shiftkey:
I don't have strong feelings on this so I'm going to ask for some feedback from others:
 
 - @iAmWillShepherd for the proposed change addressing the problem
 - @donokuda for any feels about the fix from a design perspective

donokuda:
Unrelated to this pull request, it looks like we could tighten up some of the vertical spacing for each menu item. Otherwise, it looks good to me.

Would love to hear @ampinsk's thoughts before approving in case I've missed anything 👀

iAmWillShepherd:
@Daniel-McCarthy this is great! And it looks good on my machine. I'll let this sit to let others from @desktop/design chime in.

PR #5350, opened 2018-08-07
renamed the file to drop the preceding dashes

shiftkey:
Not sure if there was a reason for the file to be named in that way.

@billygriffin let me know if I've missed something.


shiftkey:
> I certainly didn't intentionally name it like that 😄

I think it just sanitized the issue template name and replaced the emoji and whitespace with dashes.

PR #5364, opened 2018-08-09
bump yarn again

shiftkey:
Since I last opened #5335 there was a `1.9.3` update which mentions fixing a perf regression (I think it's https://github.com/yarnpkg/yarn/pull/6204).

I figure this might be good to have handy (1.9.4 is a no-op update). I also added docs on the process to upgrade, in case someone else is keeping an eye on upstream `yarn`.


docs/contributing/working-with-packages.md
```diff
@@ -54,3 +54,33 @@ To remove any packages that are no longer needed:
 ```sh
 > yarn remove [package-name]
 ```
+
+### Upgrading Yarn
+
+Desktop uses a local version of `yarn` to ensure one version is used on all
+platforms that we develop on. To upgrade Desktop to the latest version of `yarn`:
```
outofambit: 💯

Issue #4938, opened 2018-06-14
switching branch triggers an error


leonsegal:
Every time I switch to another branch using 'cmd + b' within a local repo, I get the following error modal:

```
A lock file already exists in the repository, which blocks this operation from completing.
```

I am using a mbp with high sierra 10.13.5 and github desktop 1.2.3

The error can be dismissed without consequence, I just thought you might want to know about it.

The error does not appear when changing branches with point and click in the menus.

I have looked this error up and various sources suggest creating a lock file, then deleting it, but this did not fix the error.

Leon

steveward:
Thanks for opening this @leonsegal. Could you upload a log file so we can take a closer look at what the issue might be?

macOS logs location: `~/Library/Application Support/GitHub Desktop/logs/*.desktop.production.log`

You can also access the log files by going to the file menu in GitHub Desktop and selecting `Help` > `Show Logs in Finder`. Please upload a log file from a day where you experienced this error.

leonsegal:
Log:
```
2018-06-14T15:51:00.018Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: brand-iq - leonsegal (has token)
2018-06-14T15:51:00.019Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: brand-iq - leonsegal (has token)
2018-06-14T15:51:00.048Z - error: [ui] `git -c credential.helper= checkout --progress master --` exited with an unexpected code: 128.
fatal: Unable to create '/Users/leon/Sites/brand-iq/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.

(The error was parsed as 34: A lock file already exists in the repository, which blocks this operation from completing.)
```

leonsegal:
No, this is the weird thing (and why I posted it).

I do use phpstorm, which has git integration, but this a new bug since a recent update.

I will investigate if it is phpstorm causing the error.

shiftkey:
> Every time I switch to another branch using 'cmd + b' within a local repo, I get the following error modal:

This seems to be something fairly easy to reproduce, particularly if it's happening without WebStorm running.

leonsegal:
I am new to this - did you want me to reproduce it? 

Should I take a screen shot?

paulrehkugler:
[...]

Issue #5344, opened 2018-08-05
keyboard navigation and enter will execute menu item twice


UgurAldanmaz:
## Description
When I use the keyboard to open menu and select 'Open in Git Bash', two instances of Git Bash is opened.

## Version

* GitHub Desktop:
1.3.2

* Operating system:
Microsoft Windows [Version 10.0.17134.167]

## Steps to Reproduce
1. Press `Alt`
2. Press `R`
2.1. Opens the Repository menu
3. Use arrow down to select 'Open in Git Bash'
4. Press `Enter`

### Expected Behavior
One Git Bash instance should be opened.

### Actual Behavior
Two instances of Git Bash is opened.

## Additional Information

I've added notes to the following screenshot gif files.

Here is the reproduce gif.

![spawn-issue](https://user-images.githubusercontent.com/5686267/43690174-3c00e7a4-990e-11e8-8ca8-5afc6f5b783c.gif)

Here is the expected behavior when I use the mouse.

![spawn-issue-mouse](https://user-images.githubusercontent.com/5686267/43690175-4b2ba688-990e-11e8-97e7-82a674ec138c.gif)

steveward:
Thanks for the report @UgurAldanmaz. I was able to reproduce this issue on my machine, so I'm going to label this as a bug for the team to look into.

UgurAldanmaz:
Thanks for the answer @steveward. I took a look at the source code and I think that I can fix this bug.

If there is no hurry, I would like to work on it.

steveward:
@UgurAldanmaz that would be ✨!

shiftkey:
@steveward @UgurAldanmaz what version of Git for Windows do you have installed?

steveward:
@shiftkey I'm on v2.18.

UgurAldanmaz:
@shiftkey 2.18.0.windows.1

shiftkey:
Weird - I haven't been able to reproduce this myself on a Windows `10.0.17134.191` VM with the same version of Git and `1.3.2`.

Anyway, let me know if I'm missing something with the repro steps or if I can help with the shell internals in Desktop.

shiftkey:
> I have to use the keyboard shortcuts to get it to happen, using a mouse will not trigger the bug.

Oh, this is a very interesting clue - thanks @dlloyd123!  I stumbled upon something similar over here where the behaviour differed between keyboard and mouse: #5368

dlloyd123:
[...]

PR #5368, opened 2018-08-10
fix keyboard navigation triggering dual checkouts

shiftkey:
A fix for one of the known scenarios reported in #4938 

As `FilterList` is used in a bunch of places, we need to ensure they are not affected by this change:

 - [x] test branches list works as expected when using keyboard navigation and <kbd>Enter</kbd> to checkout a branch
 - [x] test PR list works as expected when using keyboard navigation and <kbd>Enter</kbd> to checkout a PR
 - [x] test compare tab works as expected when using keyboard navigation and <kbd>Enter</kbd> to choose a branch to compare
 - [x] test repository list works as expected when using keyboard navigation and <kbd>Enter</kbd> to switch repositories

### DIFF
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -360,6 +360,14 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
       return
     }
 
+    if (event.key === 'Enter' && this.canSelectRow(row)) {
+      // this event is handled in List.onKeyDown (which fires List.toggleSelection)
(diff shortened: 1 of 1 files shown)


shiftkey:
Dropping this from 1.3.3 while we dig into #5344 as both of these issues seem related: https://github.com/desktop/desktop/issues/5344#issuecomment-412365757

shiftkey:
Closing this out to keep things under control while we investigate the underlying issue

Issue #5331, opened 2018-08-03
When deleting a branch, you can click multiple times on 'Delete', causing error popups


FreHu:
The button 'Delete' should be disabled after first click, while deletion is in progress. Clicking n times will then show n-1 popups saying the branch was not found.

FreHu:
The button in question:
![githubdesktop_2018-08-03_14-10-38](https://user-images.githubusercontent.com/5097067/43642860-7047b1be-9729-11e8-8d88-5fc39dfc10dc.png)

steveward:
Thanks for the report @FreHu! I was able to reproduce this on my Windows 10 machine. The most I was able to get was three error dialogs with some fast clicks.

Here's the error dialog that is thrown for reference:

![screen shot 2018-03-8 at 11-33-02](https://user-images.githubusercontent.com/721500/43651699-174e4450-9711-11e8-8424-c12b15fb9111.png)

adamreisnz:
Having the same issue, I think it's because clicking the first time on the delete button doesn't do anything. Then, automatically, you think you maybe didn't click properly or the app lagged, so you click again which causes the error popup because the branch was in fact already deleted.

shiftkey:
@adamreisnz thanks for the update - #5371 should fix the blocking problem but I'm not sure why it took so long. Are you able to see any slow Git operations in the log around that time that might explain why it took so long?

adamreisnz:
No, the thing is I don't think any operations are still in progress.
The modal just doesn't go away, and remains visible, until I close it manually.

The branch does get deleted, both locally and on the remote.

shiftkey:
> The branch does get deleted, both locally and on the remote.

Maybe this is the clue - we don't surface any Git error if we're unable to delete the remote branch successfully.

I'll add that as a test case to #5371.

PR #5371, opened 2018-08-10
Disable delete button while deleting a branch

iAmWillShepherd:
Fixes: #5331

 - [x] test deleting a local-only branch
 - [x] test deleting a remote branch (happy path)
 - [x] test deleting a remote branch (unhappy path)

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2337,10 +2337,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
     includeRemote: boolean
   ): Promise<void> {
     return this.withAuthenticatingUser(repository, async (r, account) => {
-      const defaultBranch = this.getRepositoryState(r).branchesState
-        .defaultBranch
-      if (!defaultBranch) {
-        throw new Error(`No default branch!`)
+      const { defaultBranch } = this.getRepositoryState(r).branchesState
+
+      if (defaultBranch == null) {
+        throw new Error(
+          `A default branch cannot be found for this repository, so the app is unable to identify which branch to switch to before removing the current branch.`
+        )
       }
 
       const gitStore = this.getGitStore(r)
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2336,23 +2336,23 @@ export class AppStore extends TypedBaseStore<IAppState> {
     branch: Branch,
     includeRemote: boolean
   ): Promise<void> {
-    return this.withAuthenticatingUser(repository, async (repo, account) => {
-      const defaultBranch = this.getRepositoryState(repository).branchesState
+    return this.withAuthenticatingUser(repository, async (r, account) => {
+      const defaultBranch = this.getRepositoryState(r).branchesState
         .defaultBranch
       if (!defaultBranch) {
         throw new Error(`No default branch!`)
       }
 
-      const gitStore = this.getGitStore(repository)
+      const gitStore = this.getGitStore(r)
 
       await gitStore.performFailableOperation(() =>
-        checkoutBranch(repository, account, defaultBranch)
+        checkoutBranch(r, account, defaultBranch)
       )
       await gitStore.performFailableOperation(() =>
-        deleteBranch(repository, branch, account, includeRemote)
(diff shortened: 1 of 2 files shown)


app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -96,6 +103,8 @@ export class DeleteBranch extends React.Component<
 
   private deleteBranch = async () => {
     const { dispatcher, repository, branch, onDeleted } = this.props
+
+    this.setState({ isDeleting: true })
```
iAmWillShepherd: There is no need to reset the state back to false since the component exits after deleting

app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -53,7 +55,12 @@ export class DeleteBranch extends React.Component<
         <DialogFooter>
           <ButtonGroup destructive={true}>
             <Button type="submit">Cancel</Button>
-            <Button onClick={this.deleteBranch}>Delete</Button>
+            <Button
+              onClick={this.deleteBranch}
+              disabled={this.state.isDeleting}
```
shiftkey: :art: could we do something similar to how we show the commit is taking time by also displaying ` Loading` component?

https://github.com/desktop/desktop/blob/0678a8bd012723ed86f2cad0dcc01044985babb8/app/src/ui/changes/commit-message.tsx#L529-L540

iAmWillShepherd: Added in #5372.

niik:
Similar to 
https://github.com/desktop/desktop/pull/5372#issuecomment-412449983, I'd rather see that we used the `disabled` and `loading` properties on the dialog itself. That will also take care of disabling the buttons.

shiftkey:
Here's an up-close GIF of the new interaction. Note the spinner in the dialog header and that both buttons are disabled while the delete is happening:

shiftkey:
@outofambit I helped get those recent changes in, so I shouldn't be involved in approving this too.

app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -95,14 +99,17 @@ export class DeleteBranch extends React.Component<
...
+
+    this.setState({ isDeleting: true })
+
     await dispatcher.deleteBranch(
       repository,
       branch,
       this.state.includeRemoteBranch
     )
-    onDeleted(repository)
+    this.props.onDeleted(repository)
```
shiftkey: This is going against the tidying-up goals of object destructuring earlier but I wanted to make this clear that we were calling a prop rather than _some other function_ (instance or otherwise).

app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -95,14 +99,17 @@ export class DeleteBranch extends React.Component<
...
 
-    return dispatcher.closePopup()
+    await dispatcher.closePopup()
```
shiftkey: The `Button.onClick` callback is a void-returning function, so I updated this to respect that signature rather than returning a `Promise<void>` that wasn't being `await`-ed.

shiftkey:
:apple:

app/src/lib/stores/app-store.ts
```diff
@@ -2336,23 +2336,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
     branch: Branch,
     includeRemote: boolean
   ): Promise<void> {
-    return this.withAuthenticatingUser(repository, async (repo, account) => {
-      const defaultBranch = this.getRepositoryState(repository).branchesState
-        .defaultBranch
-      if (!defaultBranch) {
-        throw new Error(`No default branch!`)
+    return this.withAuthenticatingUser(repository, async (r, account) => {
```
outofambit: just to make sure i understand this right—`r` here is going to to be `repository`, correct?

shiftkey: `withAuthenticatingUser` will try and update the repository to the latest details before running the callback, so I think unless we have a compelling reason to _not_ do that, we should be using `r`. I just updated this to make the scope distinction more clear.

PR #5372, opened 2018-08-10
Adds loading indicator to delete branch dialog

iAmWillShepherd:
**Before**

**After**


niik:
The `Dialog` component has two properties for indicating when there's blocking progress, the [disabled and loading](https://github.com/desktop/desktop/blob/0678a8bd012723ed86f2cad0dcc01044985babb8/app/src/ui/dialog/dialog.tsx#L85-L105) properties. Could we set these instead of manually adding a spinner to the button in order to enforce a consistent behavior across all dialogs?

Setting the disabled/loading indicator also has the added benefit of ensuring that the delete button can't be pressed more than once.

shiftkey:
I'm going to fix this in #5371 given there's existing APIs to use on `<Dialog/>` (so no need to add the inline a `<Loading/>`.

PR #5378, opened 2018-08-11
Add accelerator to Exit menu on non-macOS

### LINKED ISSUE #6507: Add a File menu keyboard shortcut for program Exit
I'm a keyboard-shortcuts kind-of-guy so a **File Menu** shortcut to **Exit** would be lovely (-:

At the moment, for me, it's an `<Alt>-F ; <Up-Arrow> ; <Enter>` affair.

Thanks
[GitHub Desktop 1.5.1 on Windows 10 (Intel x86_64)]

paulcbetts:
Fixes #6507 

The "quit" role in Electron's MenuItem on Windows is a bit lacking, it doesn't correctly set the accelerator to 'x' (the underline under the menu item that tells you the key to press), and it doesn't list "Alt-F4" as the shortcut. Since we don't really need Quit as a Special Menu Item (unlike "services" in macOS for example), we can just implement this ourselves and give it the right decoration.

## To Test:

1. Open Desktop
1. Hit Alt-F to open the File Menu
1. Hit 'x' to exit

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -324,6 +324,9 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.openCurrentRepositoryInExternalEditor()
       case 'select-all':
         return this.selectAll()
+      case 'quit':
+        remote.app.quit()
+        return
     }
 
     return assertNever(name, `Unknown menu event name: ${name}`)
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -99,7 +99,12 @@ export function buildDefaultMenu(
         click: emit('show-preferences'),
       },
       separator,
-      { role: 'quit' }
+      {
+        label: 'E&xit',
+        id: 'exit',
+        accelerator: 'Alt+F4',
+        click: emit('quit'),
+      }
     )
   }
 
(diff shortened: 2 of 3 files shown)


app/src/ui/app.tsx
```diff
@@ -324,6 +324,9 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.openCurrentRepositoryInExternalEditor()
       case 'select-all':
         return this.selectAll()
+      case 'quit':
+        remote.app.quit()
+        return
     }
 
     return assertNever(name, `Unknown menu event name: ${name}`)
```
paulcbetts: Compiler error when you don't implement all of the cases 😍

paulcbetts:
> What happens if the renderer isn't responding to the emitted event? Will the system-default handler for ALT+F4 take over and kill the app regardless or have we stolen that handler now?

It should, but let me test by just nerfing the handler

> If we're going to show the accelerator (and I don't have any strong feelings as to whether we should or not), could we make the click handler instantly terminate the app rather than emitting an event to the renderer which then tells the main app to quit?

I can do that too

shiftkey:
Dropping `ready-for-review` while there are some pending questions about the fix.

niik:
I spoke to @paulcbetts about this and he doesn't have the bandwidth to complete this at the moment so I'm going to close this out.

Issue #3764, opened 2018-01-11
Update "not found" repo error for missing orgs


tierninho:
Currently we display "Not found" as the error message when a user attempts to publish a repo to an GitHub or Enterprise org that was recently deleted or renamed. 

The error message should be more clear, perhaps "The selected organization was not found. Please try again." 

This error message should also be removed if the user navigates to another tab to publish the repo.

<img width="1115" alt="screen shot 2018-01-11 at 8 54 37 am" src="https://user-images.githubusercontent.com/14828183/34836732-ae29aec8-f6ad-11e7-8540-158cf1266ba3.png">

osx,1.0.12-beta1

JQuinnie:
I would like to work on this issue.

shiftkey:
@JQuinnie the area of the codebase you probably need to modify to fix this is [here](https://github.com/desktop/desktop/blob/master/app/src/lib/api.ts#L348-L373). In here we should check for the `404` and instead throw a more friendly error message.

iAmWillShepherd:
Turns out we are not keeping track of what orgs a user is a part of. We need to add this in before this issue is attempted.

shiftkey:
@iAmWillShepherd do we need to track that? We already render the orgs list when it's publish time, and that should be relatively up to date enough to not bother with extra work.

> Currently we display "Not found" as the error message when a user attempts to publish a repo to an GitHub or Enterprise org that was recently deleted or renamed.

@tierninho to be clear here, are these the repro steps (or similar) you performed?

 1. Create new repository
 2. Open publish repository dialog, which loads list of orgs
 3. Select organization X
 4. Rename or delete organization X so it's no longer valid
 5. Publish the repository
 6. `Not Found` error displayed

iAmWillShepherd:
Ignore that, I did this on a release version of desktop from a *very* long time ago.

Daniel-McCarthy:
I am interested in giving this Issue a try (as long as JQuinnie is no longer interested).

However, I got a bit stuck reproducing it. While I was able to reproduce the exact steps listed above in my stable 1.3.2 installation of GitHub Desktop, any builds I made from a clone of the current Desktop repository were unable to detect any organizations in the Publish Repository window.

I have tried this on a few machines and tried two different accounts I had created an organization on. Currently I'm not sure if there has been a change to how this system works, if it could be a bug, user error, or something on my end.

shiftkey:
[...]

PR #5380, opened 2018-08-12
Added error for publishing repo w/ non-existent org

Daniel-McCarthy:
As the docs say, open pull requests early. I figured would open this and see what others think.

This pull request is in response to Issue #3764. Although it does not cover all the suggestions in it, it is meant to solve the issue of the vague "Not Found" error when attempting to publish a repository with an organization that has been renamed or deleted. This change intends to make a very clear error that specifically tells the user that it is the organization that was not able to be found and shows the name of the organization to review.

I am certainly open to feedback and would be happy to make changes as needed. I am new both to this project and the conventions of this language, so I am sure there are better ways of doing things.

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -374,11 +374,8 @@ export class API {
       return await parsedResponse<IAPIRepository>(response)
     } catch (e) {
       if (e instanceof APIError) {
-        const orgsList: ReadonlyArray<IAPIUser> | null = await this.fetchOrgs()
-        if (orgsList && org) {
-          if (orgsList.indexOf(org) === -1) {
-            throw new Error(`Organization "${org.login}" Not Found`)
-          }
+        if (org !== null) {
+          throw new Error(`Organization "${org.login}" Not Found`)
         }
         throw e
       }
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -374,6 +374,12 @@ export class API {
       return await parsedResponse<IAPIRepository>(response)
     } catch (e) {
       if (e instanceof APIError) {
+        const orgsList: ReadonlyArray<IAPIUser> | null = await this.fetchOrgs()
+        if (orgsList && org) {
+          if (orgsList.indexOf(org) === -1) {
+            throw new Error(`Organization "${org.login}" Not Found`)
+          }
+        }
         throw e
       }
 
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -375,7 +375,11 @@ export class API {
     } catch (e) {
       if (e instanceof APIError) {
         if (org !== null) {
-          throw new Error(`Organization "${org.login}" Not Found`)
+          throw new Error(
+            `Unable to create repository for organization '${
+              org.login
+            }'. Verify it exists and that you have permission to create a repository there.`
+          )
         }
         throw e
       }


app/src/lib/api.ts
```diff
@@ -374,6 +374,12 @@ export class API {
       return await parsedResponse<IAPIRepository>(response)
     } catch (e) {
       if (e instanceof APIError) {
+        const orgsList: ReadonlyArray<IAPIUser> | null = await this.fetchOrgs()
```
shiftkey: Do we need to re-fetch the organizations here? I think if we tried to `POST` to `orgs/${org.login}/repos` and got a 404 here it's reasonable to think this is related to the organization.

app/src/lib/api.ts
```diff
@@ -374,6 +374,12 @@ export class API {
...
+        if (orgsList && org) {
+          if (orgsList.indexOf(org) === -1) {
+            throw new Error(`Organization "${org.login}" Not Found`)
```
shiftkey: We can iterate on this, but here's a suggestion for the error message itself:

> `Unable to create repository for organization '${org.login}' - check it exists and that you have permissions to create a repository there`

app/src/lib/api.ts
Daniel-McCarthy: Wonderful, I was hoping there was a way to avoid fetching again. To clarify: if the org used is `orgs/${org.login}/repos` and received an APIError from this POST request then we have enough information to throw this error, and therefore no additional checks are needed?

shiftkey: @Daniel-McCarthy yeah, I wouldn't worry about being cleverer here than that

app/src/lib/api.ts
```diff
@@ -374,6 +374,13 @@ export class API {
       return await parsedResponse<IAPIRepository>(response)
     } catch (e) {
       if (e instanceof APIError) {
+        if (org) {
+          throw new Error(
+            `Unable to create repository for organization '${
+              org.login
+            }' - Verify it exists and that you have permission to create a repository there.`
```
shiftkey: :art: drop the `V` here from `Verify` or use a `.` instead of a `-` to indicate it's a new sentence.

app/src/lib/api.ts
```diff
@@ -374,6 +374,13 @@ export class API {
       return await parsedResponse<IAPIRepository>(response)
     } catch (e) {
       if (e instanceof APIError) {
+        if (org) {
```
shiftkey: :art: we're not fans of truthiness here - can you make this a bit more explicit by having a `if (org !== null) {` on this line?

Daniel-McCarthy:
Okay, I have updated again with the current set of suggestions. I made the null check more explicit to clarify the intent of it, and I have added a period to clarify the end of the first sentence. (I have kept the period outside of the quotations so that it can not be assumed to be part of the org name.)

Hopefully this fits better stylistically.

app/src/lib/api.ts
Daniel-McCarthy: Just to clarify, do you mean to change the V to not be capitalized? Would this be preferable to the period I have added in the most recent commit?

Daniel-McCarthy:
@shiftkey To clarify, this pull request is ready for review.

Should I clarify this for my other pull requests? I simply hope to get on the same page and do not want to come across as attempting to rush anything.

shiftkey:
One sec, confirming how the dialog looks when you trip this error (the new message has more words than the old message)...

shiftkey:
Flawless victory:

billygriffin:
Thanks @Daniel-McCarthy, this is awesome! Excited to see it go live.

PR #5399, opened 2018-08-14
avoid using getDefaultRemote where possible

shiftkey:
Desktop will refresh the default remote inside `AppStore.refreshRepository`, and I think we can avoid making this call repeatedly when the result is already cached and not likely to change quickly for users.

`AppStore.matchGitHubRepository` does have a number of callers and will fire more regularly, and `AppStore._checkoutPullRequest` is only called by the `Dispatcher`.


app/src/lib/stores/app-store.ts
```diff
@@ -2208,7 +2207,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
-  public async _repositoryWithRefreshedGitHubRepository(
+  private async _repositoryWithRefreshedGitHubRepository(
```
iAmWillShepherd: I think the `_` prefix is only used for public functions that shouldn't be called. Now that this is private, shouldn't the prefix be removed?

app/src/lib/stores/app-store.ts
```diff
@@ -2294,7 +2293,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
   private async matchGitHubRepository(
     repository: Repository
   ): Promise<IMatchedGitHubRepository | null> {
-    const remote = await getDefaultRemote(repository)
+    const gitStore = this.getGitStore(repository)
+    const remote = gitStore.defaultRemote
     return remote ? matchGitHubRepository(this.accounts, remote.url) : null
```
iAmWillShepherd: `return remote !== null ? matchGitHubRepository(this.accounts, remote.url) : null`
I know you didn't touch this, but it's so close to your changes that I think this would be a great cleanup.

app/src/lib/stores/app-store.ts
```diff
@@ -3819,13 +3819,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
       head.gitHubRepository.cloneURL === gitHubRepository.cloneURL
 
     if (isRefInThisRepo) {
-      const defaultRemote = await getDefaultRemote(repository)
+      const gitStore = this.getGitStore(repository)
+      const defaultRemote = gitStore.defaultRemote
```
iAmWillShepherd: This is repeated above, may be worth breaking this out into a function...

shiftkey: I'm going to push back on this change because creating more private functions inside `AppStore` that depend on `this.` is going to add more noise, particularly when it's only replacing two lines of code used in two places.

shiftkey:
:apple:

PR #5401, opened 2018-08-14
Delete 1.2 Section as it has been Released

say25:


shiftkey:
@say25 thanks for noticing this. I'll see if I can give that page a refresh as we get close to figuring out what 1.4 and beyond will look like.

Issue #4753, opened 2018-05-22
Github Desktop and Default text editors + EmEditor


killerwife:
Currently github desktop only supports selective text editors on windows. It does not respect two following key settings:
1. using default text editor as setup in windows for given file type
2. using default text editor set in git config

I have both of these set to EmEditor, the old Github desktop client is capable of using them properly, but the new one does not even give the option to change this. Using notepad is insanity when doing cross platform development, and installing an additional editor just because the client does not support setting it in some sane way is no good either.

Reason for using EmEditor: we use extremely huge files and as a result, most of not all software is incapable of handling it well. (talking about 1GB + files here)

I have not seen an EmEditor thread around here yet.

killerwife:
Absolutely. When I associate a given editor with .cpp or .h or .py, I would expect the Github Desktop app to use that associated program. Just to clarify on windows 10 its the menu called "Choose default apps by file type".

steveward:
> When I associate a given editor with .cpp or .h or .py, I would expect the Github Desktop app to use that associated program

GitHub Desktop should respect that default program when you right-click on a file in the `Changes` view and select `Open with default program`. Let me know if I'm misunderstanding what you are wanting to do here.

killerwife:
Yes. When I used https://imgur.com/a/W2n3mAB "Open with default program", notepad was used. When I used "show in explorer" and doubleclicked the file, EmEditor was used.

shiftkey:
There's some extra context here about the differences between the classic Windows app and the new Electron app, and it requires some understanding of how Windows handles opening files.

Windows app all end up funnelling through [`ShellExecute`](https://msdn.microsoft.com/en-us/library/windows/desktop/bb776886(v=vs.85).aspx) to ask the OS how to handle a given file. Electron too has support for opening a file in the user's preferred program - [`shell.openItem(fullPath)`](https://electronjs.org/docs/api/shell#shellopenitemfullpath) and [`openExternal(url)`](https://electronjs.org/docs/api/shell#shellopenexternalurl-options-callback).

How [`openExternal` works on Windows](https://github.com/electron/electron/blob/eb63bea87c299b3c440d6a36ccced3cc9653e1c4/atom/common/platform_util_win.cc#L302-L317) is important:

```c
  if (reinterpret_cast<ULONG_PTR>(ShellExecuteW(NULL, L"open",
                                                escaped_url.c_str(), NULL, NULL,
                                                SW_SHOWNORMAL)) <= 32) {
[...]

PR #5406, opened 2018-08-15
Add support for EmEditor external editor

killerwife:
Solving issue https://github.com/desktop/desktop/issues/4753
Done through instructions on https://github.com/desktop/desktop/blob/master/docs/technical/editor-integration.md
Missing part: Registry for 32-bit emeditor since I do not know it.
Should there be need for assistance with merging, happy to do anything so that this is merged asap.

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -18,6 +18,7 @@ export enum ExternalEditor {
   VisualStudioCodeInsiders = 'Visual Studio Code (Insiders)',
   SublimeText = 'Sublime Text',
   CFBuilder = 'ColdFusion Builder',
+  EmEditor = 'EmEditor'
 }
 
 export function parse(label: string): ExternalEditor | null {
@@ -35,6 +36,9 @@ export function parse(label: string): ExternalEditor | null {
   }
   if (label === ExternalEditor.CFBuilder) {
     return ExternalEditor.CFBuilder
+  }
+  if (label === ExternalEditor.EmEditor) {
+    return ExternalEditor.EmEditor
   }
 
   return null
@@ -113,7 +117,21 @@ function getRegistryKeys(
             'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Adobe ColdFusion Builder 2016',
         },
       ]
-
+    case ExternalEditor.EmEditor:
+      return [
+        // 64-bit version of EmEditor
+        {
+          key: HKEY.HKEY_LOCAL_MACHINE,
+          subKey:
+            'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{D03753CD-7F12-4281-9709-DDE30EFF8D6D}',
+        },
+        // 32-bit version of EmEditor - TODO: Find registry
+        {
+          key: HKEY.HKEY_LOCAL_MACHINE,
+          subKey:
+            'SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{FFA9716C-7FBB-4B4F-8130-58658D3D15FE}',
+        },
+      ]
     default:
       return assertNever(editor, `Unknown external editor: ${editor}`)
   }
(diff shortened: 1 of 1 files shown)


shiftkey:
@killerwife there's a limitation with EmEditor that means it's not suitable for use in GitHub Desktop currently:

 - the editors we use in the "Open in External Editor" list support opening the directory containing the repository
 - this means we can launch the editor as a simple way for users to work with the whole repository
 - not all editors have this experience - for example, #4120 wanted to add support Notepad++ but we closed that out because it didn't have the right command line flags (and instead would open _all the files_)
 - the only relevant one I could find for EmEditor was `/cd` which "set the current directory as the default folder in the Open dialog box." [[docs](http://www.emeditor.org/en/howto_file_file_commandline.html)]
 - this would open to the directory, but then ask me which files I wanted to open



Unless I've missed something here, I don't think I can take in this PR because of the mismatch between our "Open in External Editor" and what EmEditor currently supports.

killerwife:
@shiftkey how about this: /rr *.*

killerwife:
Also I would like to express the frustration for the propably most used editor not being supported (Notepad++) due to that limitation. All of this editor stuff considerably hampers usability. Can't there be a switch that open in external editor would only show up for select files, instead of whole repository? I personally do not see the usecase, since I use development tools for that. Maybe I am missing the big picture though.

shiftkey:
> how about this: "/rr *.*"

This will open _all the files_, which was the Notepad++ behaviour we realised wasn't desirable.

>  Can't there be a switch that open in external editor would only show up for select files, instead of whole repository?

Historically we've been focused on two flows:
 - open a file in an external program
 - opening a folder in an external program

For the "open a file in an external program" flow we defer to your OS choice here because you may configure different editors for different file extensions, and it didn't seem worth reimplementing that in Desktop (and not covering all potential scenarios).

For the "open a folder in an external program" flow we felt this was better to address within Desktop because there isn't equivalent functionality in your OS already setup.

> Maybe I am missing the big picture though.

Remember [my comment](https://github.com/desktop/desktop/issues/4753#issuecomment-392394769) from a while ago?

> So while `EmEditor` does declare itself as an _editor_ for text files, Electron is only leveraging programs that support _opening_ a given extension.

I think this is the core of your issue with how "Open in external program" is currently working in Desktop. I went through these steps and it seems to work for me:

1. For some file extension you're interested in, choose **Open with** | **Choose another app**



2. Ensure EmEditor is set always set for opening this extension



3. Then try and open a file of that same extension in Desktop:



If you're satisfied with that behaviour and able to see the same thing on your end, we can revisit #4753 and figure out what's left to do.

shiftkey:
@killerwife 

>  I have almost all text based files set to emeditor, and it doesnt use it, and instead opens it in notepad. Any way I can troubleshoot that?

Thanks for the extra context. I poked at `.sql` to set it to EmEditor:



And when I right-click on a `.sql` file I get these two context menu entries:



Note that **Edit** is emphasised as the default behaviour. This means when I double-click or choose **Edit** it'll choose EmEditor, but when I click **Open** it will launch Notepad. That's the core of the issue here, and Desktop is continuing to try and **Open** the file, which I don't seem to be able to workaround completely.

killerwife:
Okay and how is "Open" changable in windows 10. I can't seem to be able to find a solution for that.

shiftkey:
> Okay and how is "Open" changable in windows 10. I can't seem to be able to find a solution for that.

I had to do this in the registry, and I was fine with ruining my machine, so some warnings:

 - **The registry is annoyingly fiddly, and I have broken myself before in here**
 - **This only worked for `.sql` extensions.**
 - **This works on my machine**

1. I found the entry for `.sql` at `Computer\HKEY_CURRENT_USER\Software\Classes\.sql`. It points to an entry named `sql_auto_file`

screen shot 2018-08-15 at 11 58 23 am

2. I went to the entry for `sql_auto_file` at `Computer\HKEY_CURRENT_USER\Software\Classes\sql_auto_file`. It had the `shell\edit\command` registry key which pointed to `EmEditor`

3. I created the `shell\open\command` registry key alongside, and set `"C:\Users\shiftkey\AppData\Local\Programs\EmEditor\EMEDITOR.EXE" "%1"` to the default value.

screen shot 2018-08-15 at 11 55 59 am

4. Back on my `.sql` file, the context menu now defaults to **Open**



5. Now Desktop launches EmEditor when using **Open in External Program**

shiftkey:
I'm going to close this out because we can't merge in EmEditor external editor support in it's current form, and we have #4753 which is tracking the related issue about **Open in External Program** and `core.editor`.

killerwife:
Thx for the thorough explanations, sadly the regedit did not work for me. It seems to have smth to do with mysql though.

shiftkey:
> It seems to have smth to do with mysql though.

Yep, and at this point I'm going to have to stop trying to debug an unknown configuration and registry setup for fear of breaking things.

PR #5420, opened 2018-08-16
Support VSCode user scope installation

### LINKED ISSUE #5281: add support for detecting new User Install of VS Code Insiders
File -> Options -> Advanced -> External editor
image

I believe VS Code Insiders is supported, however Microsoft has recently released a new version of VS Code that installs to the user directory instead of Program Files, to prevent the requirement of running as administrator in order to get automatic updates.
I recently changed to this version of VS Code, and GitHub Desktop is no longer acknowledging my editors installation.

saschanaz:
Fixes #5281

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -64,13 +64,13 @@ function getRegistryKeys(
       return [
         // 64-bit version of VSCode (user) - provided by default in 64-bit Windows
         {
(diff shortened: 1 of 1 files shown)


shiftkey:
@CharlesStover @Elderry @Cologler are you able to :eyes: this change and ensure these GUIDs look correct (for Stable and Insiders)?

PR #5421, opened 2018-08-16
this link should point to Desktop, not Central

shiftkey:


docs/process/releasing-updates.md
```diff
@@ -156,11 +156,11 @@ Testing that an update is detected, downloaded, and applied correctly is very im
 
 If you don't have the app for `beta`, for example, you can always download the previous version on Central to see it update 
 
-** Make sure you move your application out of the Downloads folder and into the Applications folder for macOS or it won't auto-update.
+_Make sure you move your application out of the Downloads folder and into the Applications folder for macOS or it won't auto-update_.
 
 ### 6. Check Error Reporting
 
-If an error occurs during the release process, a needle will be reported to Central's [Haystack](https://haystack.githubapp.com/central).
+If an error occurs during the release process, a needle will be reported to Central's [Haystack](https://haystack.githubapp.com/desktop).
```
niik: I think this was actually correct, right? An error during _deployment_ will actually be reported to central's haystack, not desktop's?

shiftkey: I think you're right, but at the time I was walking @outofambit through the steps and thought this was supposed to be the link to Desktop's haystack (which would be beneficial to have). 

I'll follow up to revert this and see if I can word this section better.

Issue #3089, opened 2017-10-18
Clearing Publish Repository error messages when switching tabs


tierninho:
If a user gets a `Repo creation failed` error when attempting to publish a repo, that error is specific to either GitHub or Enterprise (unless they fail to publish at both). The error message is currently carried though across both tabs. It should be restricted to only the tab that is relevant. 

For instance, the error is showing on the GitHub and Enterprise tab, despite my user only being logged into GitHub.

Repro by trying to publish a repo with a duplicate name.

<img width="447" alt="screen shot 2017-10-18 at 1 53 48 pm" src="https://user-images.githubusercontent.com/14828183/31719763-6c79160a-b40c-11e7-90a2-ed0d1b7e29c1.png">
<img width="459" alt="screen shot 2017-10-18 at 1 53 56 pm" src="https://user-images.githubusercontent.com/14828183/31719764-6c9e1d88-b40c-11e7-9411-e775c825e5bb.png">

v1.0.5beta0, mac

Update: this applies to all errors found in this part of the app.

Daniel-McCarthy:
Hey @tierninho , I see this issue was recently marked `help wanted` and `priority-3`.
Was there still something that needed to be addressed with this issue?

My pull request to show errors on the tabs they were generated on was merged. Let me know if there's any of improvement or other change required for this issue.

tierninho:
@Daniel-McCarthy Ah, good catch! Was just doing some routine cleanup. 

Removing those labels now... 👍

PR #3787, opened 2018-01-12
Clear repository filter text when switching tabs

reyronald:
... in the Clone Repository modal.

Addressing issue #3739

----

The issue was labeled with `help-wanted` so I decided to give it a go, hope you guys don't mind me interfering. 

This is a first approach using @tierninho 's suggestion in https://github.com/desktop/desktop/issues/3739#issuecomment-356107533, so it doesn't take into account my own inquiry in https://github.com/desktop/desktop/issues/3739#issuecomment-357322240.
Would have no problem in changing the implementation if a different behavior is desired by the team.

Let me know what you guys think. 

**Current behavior**:

clear-filter-before

**Pull request behavior**:

clear-filter

PR #5422, opened 2018-08-16
Clear Displayed Publish Repository Error on change of tab

Daniel-McCarthy:
This pull request is in response to Issue #3764 and #3089. Among the suggestions was clearing errors when switching tabs. However, I figured this would be a good opportunity to discuss such a change.

The intent of this change is to stop showing an error in the Publish Repository window when the user has changed tabs and the error from their attempt to publish is no longer related to the current context.

While it makes sense to clear an error such as an issue publishing with the selected organization, it may not make as much sense with errors that may persist the immediate context (such as warning that internet issues may have prevented the publishing).

Therefore, I figured it would be great to discuss :

- If this would be a welcomed change in the first place.
- If there is a better way to implement such a change.
- If a different approach entirely would be preferable (such as an individual solution to individual errors)

(Additionally, in regard to consistency, this is not something that seems to occur in the Clone Repository window. For example, if you type a single character for the URL in Clone a Repository, you will get a "The destination already exists" error that does not clear on deleting the url or switching tabs. However, I feel like it would make sense to do so in that case as well.)

### DIFF
--- a/app/src/ui/publish-repository/publish.tsx
+++ b/app/src/ui/publish-repository/publish.tsx
@@ -247,13 +247,12 @@ export class Publish extends React.Component<IPublishProps, IPublishState> {
   }
 
   private onTabClicked = (index: PublishTab) => {
-    // Clear the selected org since dot com and Enterprise will have a different
-    // set of orgs.
-    const settings = { ...this.state.publishSettings, org: null }
-    this.setState({ currentTab: index, publishSettings: settings })
-
     const isTabChanging = index !== this.state.currentTab
     if (isTabChanging) {
+      // Clear the selected org since dot com and Enterprise will have a different
+      // set of orgs.
+      const settings = { ...this.state.publishSettings, org: null }
+      this.setState({ currentTab: index, publishSettings: settings })
       // Swap the current stored error from the active tab with the error from
       // the inactive tab. So that each tab saves and displays their own error.
       const temporaryError: Error | null = this.state.error
--- a/app/src/ui/publish-repository/publish.tsx
+++ b/app/src/ui/publish-repository/publish.tsx
@@ -47,6 +47,13 @@ interface IPublishState {
    */
   readonly error: Error | null
 
+  /**
+   * The error pertaining to the tab the user is not
+   * actively using. It is swapped when the user changes
+   * the current tab.
+   */
+  readonly inactiveTabError: Error | null
+
   /** Is the repository currently being published? */
   readonly publishing: boolean
 }
@@ -76,6 +83,7 @@ export class Publish extends React.Component<IPublishProps, IPublishState> {
       currentTab: startingTab,
       publishSettings,
       error: null,
+      inactiveTabError: null,
       publishing: false,
     }
   }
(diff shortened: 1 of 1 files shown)


outofambit:
@Daniel-McCarthy thanks for opening this up for discussion!

My two cents is that the error banner can be cleared from the UI upon leaving that tab, but that it should be reinstated when the user returns to the original tab. (I think its ok to clear the banner, even if it is an error that would apply to the other, since the user can still _attempt_ to publish a repository and then get the same error.)

Would love to hear from @desktop/design on this too. :)

Daniel-McCarthy:
Hi @outofambit,

I think that's a great idea and quite a bit better than how I have it currently implemented. I can certainly look into this as well and see how that would be done.

I look forward to any additional feedback.
Thank you for your suggestion.

shiftkey:
@Daniel-McCarthy can you add in a screenshot of the current behaviour for reference?

Daniel-McCarthy:
@shiftkey Of course.

peek 2018-08-20 08-02

- The current behavior can only show a single error at a time and it shows on both tabs.

- The behavior I have implemented in this pull request deletes the error when switching tabs. Errors will not appear on the wrong tab, but will be gone when returning to the tab the error originated on.

- The behavior @outofambit has suggested (and I have implemented in another branch) allows for each tab to have their own error and only display on their respective tab.

shiftkey:
@Daniel-McCarthy thanks for the extra context.

> The current behavior can only show a single error at a time and it shows on both tabs.

This reminds me of an earlier issue where the filter text in the Clone Repository flow would persist when you switch tabs: https://github.com/desktop/desktop/pull/3787

We ended up implementing a fix like this:

https://github.com/desktop/desktop/blob/0e04ed151e4cab557740432112e25cdcdbd9767b/app/src/ui/clone-repository/clone-repository.tsx#L89-L93

I think we could do something similar here (keep the error scoped to the component, but clear it when the repository tab changes). Thoughts @outofambit?

Daniel-McCarthy:
> I think we could do something similar here (keep the error scoped to the component, but clear it when the repository tab changes).

This is a good point. The implementation I have attached to this pull request clears the error onTabClicked. Therefore even if you click the tab you are on it will clear the error, which is less than ideal.

The implementation I did [in another branch](https://github.com/Daniel-McCarthy/desktop/commit/89401b33ec75ac0d7aec083f97b8c9daba0bc7f6) that allows for two errors accounts for this. I made it so that if onTabClicked activates and the tabIndex is changing, then it will swap the error from the current tab Error with the a second Error variable (for the currently inactive tab.)

 A potential issue is that the implementation does not have awareness of which tab is which, however I have not found an edge case where the errors appear on the wrong tab. It seems to work for the following:

- Clicking the tab one is already on. (It stays the same).
- Clicking the opposite tab. (It swaps the errors and shows that tab's error if any exists.)
- Clicking the tab one is already on, but then use the arrow keys to select the opposite tab. (Works as above.)

I would be happy to update this pull request with the changes from that branch as long as we are decided on if we should keep the errors from both tabs and display them on their respective tab.

Alternatively I can update the commits attached to this pull request to only clear the error when the tab is actually changing, like my other branch does.

outofambit:
hey @Daniel-McCarthy and @shiftkey, i just chatted with @ampinsk (of @desktop/design) and we definitely want the error to be preserved for its tab. in other words: 

> The behavior @outofambit has suggested (and I have implemented in another branch) allows for each tab to have their own error and only display on their respective tab.

So @Daniel-McCarthy, this:

> I would be happy to update this pull request with the changes from that branch as long as we are decided on if we should keep the errors from both tabs and display them on their respective tab.

would be awesome. 😀 Thank you so much!

Daniel-McCarthy:
@outofambit Of course, gladly.

I have updated the pull request with the change now.  Great to hear that it's working as intended.

shiftkey:
Moving this to the 1.4.0 milestone as I'd like to get the existing changes published today/tomorrow.

PR #5424, opened 2018-08-16
[WIP] writing a roadmap for our flux transition

shiftkey:
Opening this up now to share and get feedback, especially from @outofambit who has also been pondering this process.

Rather than throwing PRs into review, I thought I'd do a brain dump first to outline how I think we can get there gradually and ensure we're all on board with it. 

[Rendered Markdown](https://github.com/desktop/desktop/blob/docs-for-migrating-from-flux/docs/technical/the-great-stores-migration.md)

Feel free to ask questions/suggest edits/push commits to this branch.

### DIFF
--- a/docs/technical/the-great-stores-migration.md
+++ b/docs/technical/the-great-stores-migration.md
@@ -14,10 +14,10 @@ This document is focused on:
  - `IssuesStore` - backing store for refreshing and caching open GitHub issues for each repository
  - `PullRequestsStore` - backing store for refreshing and caching open GitHub pull requests for each repository
  - `CloningRepositoriesStore` - component for tracking repository clones in the app
- - `EmojiStore` - contains emoji lookup for use in UI components
+ - ~~`EmojiStore` - contains emoji lookup for use in UI components~~ deprecated
  - `GitHubUserStore` - backing store for refreshing and caching GitHub collaborators for each repository
  - `SignInStore` - component for tracking user sign-in state in the UI
- - `RepositorySettingsStore` - backing store for repository-specific settings
+ - ~~`RepositorySettingsStore` - backing store for repository-specific settings~~ deprecated
  - `TokenStore` - used by `AccountStore` to tidy up credentials, but mostly used
    for authenticating Git operations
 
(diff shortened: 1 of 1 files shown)


docs/technical/the-great-stores-migration.md
```diff
@@ -0,0 +1,160 @@
...
+### Stage 4 - Introduce `redux` to handle some dispatcher work
+
+This is where things get interesting. So rather than trying to rewrite the
+`AppStore` and all the functionality it touches, what if we can setup this flow
+within the app:
+
+ - create a `redux` store alongside `AppStore` that has the same `IAppState`
+   shape
+ - identify a part of `IAppState` that we can port to the Redux pattern
+ - implement actions and reducers to replace this functionality
```
j-f1: You might find [`immer`](https://github.com/mweststrate/immer) useful here.

docs/technical/the-great-stores-migration.md
```diff
@@ -0,0 +1,160 @@
+# Organizing Desktop's Store Interactions For Great Good
+
+I've been pondering on what to do with our `AppStore` and other stores given
```
outofambit: are these concerns and examples meant to be generalized to stores like `git-store`, `issue-store`, and on? there are a lot of `*-store.ts` files, so i want to make sure i understand the scope of this document as it currently stands.

shiftkey: 👍 I've mostly been focused on `AppStore` initially but I can expand this to some gut feels about those and how to organize them.

docs/technical/the-great-stores-migration.md
shiftkey: Interesting footnote - `immer` has some confusion about `ReadonlyArray<T>` because the TS declarations assigns every array-like object it finds to `DraftArray<T>`.



This hurts on whenever you want to update an object that has `ReadonlyArray<T>` because it requires manually managing stuff. I'm going to try and upstream a fix for making this handle both `Array<T>` and `ReadonlyArray<T>`.

j-f1: Yeah, that’s unfortunate. Hopefully the issue will be fixable!

mirai-zoom: エラーコードの修正は大変苦労しました（ｐｈｐで）

shiftkey: I'm figuring this out here if y'all want to follow along: https://github.com/mweststrate/immer/pull/176

outofambit:
just want to make a note here that React.js recently added "Contexts", which may be worth considering (over Redux) for our use case https://reactjs.org/docs/context.html

shiftkey:
I plan come back to this later, but for now I'm going to close this out so we can keep our open PR list under control.

Issue #4293, opened 2018-03-20
application should block adding a bare repository


ysabri:
## Description
So when I try and create a new repository in a directory that includes a bare repository, the app recognizes that the directory includes a git repository and asks me to add it instead. So far so good until I add the repository and the app tries to perform a git operation (I don't know which one exactly) and fails with "fatal: This operation must be run in a work tree". 

## Version

* GitHub Desktop: 1.1.0

* Operating system: Windows 10

## Steps to Reproduce

1. Add a bare repository.  
### Expected Behavior

The app rejects adding it since it does not have a working tree or offers to do a local clone.
### Actual Behavior

The app adds it and fails to do a git operation repeatedly till the repo is removed. 

## Additional Information

So I looked at how the app decides if a path is a git directory. I understand how rev-parse --show-cdup is useful but it does not care if the repo is bare or not. I also tried to put a path into the url field of clone and the app complained that the destination already exists. I feel local clones is an enhancement that should be added as well.

tierninho:
@ysabri thanks for filing. Possible to include the logs for this before one of our core members addresses your issue? See the Help menu to find the logs in the Explorer.

Have you tried running one of the following?

Single commands:
- `git --work-tree=/path/to/work/tree checkout master`

Environment variable
- `GIT_WORK_TREE=/path/to/work/tree git status`

ysabri:
Here is the file: [2018-03-20.desktop.production.log](https://github.com/desktop/desktop/files/1831153/2018-03-20.desktop.production.log)
 I also I am not sure what path to feed --work-tree, I do not know what in the bare repository to point the command. I tried refs/heads and it did not work.

shiftkey:
> So when I try and create a new repository in a directory that includes a bare repository, the app recognizes that the directory includes a git repository and asks me to add it instead.

To confirm I understand the issue, are you using `git init --bare` to create a Git repository and then adding that folder to the app?

```
$ cd test
$ mkdir repo
$ cd repo
$ git init --bare
Initialized empty Git repository in C:/Users/shiftkey/src/test/repo/
$ ls
config  description  HEAD  hooks/  info/  objects/  refs/
```

I suspect there are some assumptions in the app around the presence of a `.git` directory _within_ the folder that you provide. @ysabri can you tell me more about your use case here, and why you're avoiding initialising the repository in the typical way?

ysabri:
[...]

PR #5434, opened 2018-08-18
Start blocking the ability to add local bare repositories

Daniel-McCarthy:
This pull request is in response to Issue #4293. The intention is to stop allowing bare repositories from being added via GitHub Desktop as they are not currently supported. To do so, a function that checks if the repository is bare has been added. Additionally, the Add-Existing-Repository dialogue now uses the added function to check for a bare repository and displays a warning and disables the add button.

I am still new to this project and a number of the technologies involved in it so there is likely plenty room for improvement. I am more than happy to make changes to my implementation. I suspect it can be simplified, especially where it deals with `boolean | null` types.

### DIFF
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -55,26 +55,19 @@ export async function getTopLevelWorkingDirectory(
 export async function checkIfRepositoryIsBare(
   path: string
 ): Promise<boolean | null> {
-  let result
-
   try {
-    result = await git(
+    const result = await git(
       ['rev-parse', '--is-bare-repository'],
       path,
       'checkIfRepositoryIsBare'
     )
+    return result.stdout.trim() === 'true'
   } catch (e) {
     if (e.message.includes('not a git repository')) {
       return null
     }
 
     throw e
-  }
-
-  if (result === null) {
-    return null
-  } else {
-    return result.stdout === 'true\n'
   }
 }
 
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -52,6 +52,32 @@ export async function getTopLevelWorkingDirectory(
   return Path.resolve(path, relativePath)
 }
 
+export async function checkIfRepositoryIsBare(
+  path: string
+): Promise<boolean | null> {
+  let result
+
+  try {
+    result = await git(
+      ['rev-parse', '--is-bare-repository'],
+      path,
+      'checkIfRepositoryIsBare'
+    )
+  } catch (e) {
+    if (e.message.includes('not a git repository')) {
+      return null
+    }
+
+    throw e
+  }
+
+  if (result === null) {
+    return null
+  } else {
+    return result.stdout === 'true\n'
+  }
+}
+
 /** Is the path a git repository? */
 export async function isGitRepository(path: string): Promise<boolean> {
   return (await getTopLevelWorkingDirectory(path)) !== null
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -52,6 +52,13 @@ export async function getTopLevelWorkingDirectory(
   return Path.resolve(path, relativePath)
 }
 
+/**
+ * Checks if the repository at a path is bare.
+ *
+ * @param path The path to the Git repository to check.
+ *
+ * @returns null if the path provided does not contain a Git repository.
+ */
 export async function isBareRepository(path: string): Promise<boolean | null> {
   try {
     const result = await git(
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -59,7 +59,7 @@ export async function getTopLevelWorkingDirectory(
  *
  * @returns null if the path provided does not contain a Git repository.
  */
-export async function isBareRepository(path: string): Promise<boolean | null> {
+export async function isBareRepository(path: string): Promise<boolean> {
   try {
     const result = await git(
       ['rev-parse', '--is-bare-repository'],
@@ -69,7 +69,7 @@ export async function isBareRepository(path: string): Promise<boolean | null> {
     return result.stdout.trim() === 'true'
   } catch (e) {
     if (e.message.includes('not a git repository')) {
-      return null
+      return false
     }
 
     throw e
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -52,14 +52,12 @@ export async function getTopLevelWorkingDirectory(
   return Path.resolve(path, relativePath)
 }
 
-export async function checkIfRepositoryIsBare(
-  path: string
-): Promise<boolean | null> {
+export async function isBareRepository(path: string): Promise<boolean | null> {
   try {
     const result = await git(
       ['rev-parse', '--is-bare-repository'],
       path,
-      'checkIfRepositoryIsBare'
+      'isBareRepository'
     )
     return result.stdout.trim() === 'true'
   } catch (e) {
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -55,9 +55,9 @@ export async function getTopLevelWorkingDirectory(
 /**
  * Checks if the repository at a path is bare.
  *
- * @param path The path to the Git repository to check.
+ * @param path The path to the repository to check. An error will be thrown if the path does not exist on disk.
  *
- * @returns null if the path provided does not contain a Git repository.
+ * @returns true if the path contains a bare Git repository. Returns false if it is not bare or is not a Git repository.
  */
 export async function isBareRepository(path: string): Promise<boolean> {
   try {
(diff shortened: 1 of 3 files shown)


app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,32 @@ export async function getTopLevelWorkingDirectory(
   return Path.resolve(path, relativePath)
 }
 
+export async function checkIfRepositoryIsBare(
```
shiftkey: :art: can we rename this function to `isBareRepository` - this feels more On Brand™ alongside `isGitRepository` which is also defined in this file.

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,32 @@ export async function getTopLevelWorkingDirectory(
...
+  path: string
+): Promise<boolean | null> {
+  let result
+
+  try {
+    result = await git(
+      ['rev-parse', '--is-bare-repository'],
+      path,
+      'checkIfRepositoryIsBare'
```
shiftkey: :art: this should match the function being called

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,32 @@ export async function getTopLevelWorkingDirectory(
...
+    )
+  } catch (e) {
+    if (e.message.includes('not a git repository')) {
+      return null
```
shiftkey: Are you able to add some documentation to the header of this function, so we can understand when we should expect `null` versus a boolean versus an error?

app/src/ui/add-repository/add-existing-repository.tsx
```diff
@@ -158,11 +183,18 @@ export class AddExistingRepository extends React.Component<
 
     const path = directory[0]
     const isRepository = await isGitRepository(path)
+    const isBareRepositoryResult = await checkIfRepositoryIsBare(path)
```
shiftkey: :art: I think we can distill this down into a one-liner, based on what I think `null` represents:

```ts
const isBareRepository = (await checkIfRepositoryIsBare(path)) || false
```

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,32 @@ export async function getTopLevelWorkingDirectory(
...
+    }
+
+    throw e
+  }
+
+  if (result === null) {
+    return null
+  } else {
+    return result.stdout === 'true\n'
```
shiftkey: :art: use `.trim()` to avoid the need to check for `\n` here

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,32 @@ export async function getTopLevelWorkingDirectory(
...
+    )
+  } catch (e) {
+    if (e.message.includes('not a git repository')) {
+      return null
+    }
+
+    throw e
+  }
+
+  if (result === null) {
```
shiftkey: Rather than trying to make the try-catch as narrow as possible, let's just lift this up into the try-catch to simplify things:

```
  try {
    const result = await git(
      ['rev-parse', '--is-bare-repository'],
      path,
      'checkIfRepositoryIsBare'
    )
    return result.stdout.trim() === 'true'
  } catch (e) {
    if (e.message.includes('not a git repository')) {
      return null
    }

    throw e
  }
```

app/src/lib/git/rev-parse.ts
shiftkey: This `'checkIfRepositoryIsBare'` string is used for tracing Git operations (in case they run unexpectedly long or if they error, details end up in the log). The convention is that this name should be the same as the function itself.

Daniel-McCarthy: The source of my confusion is that the function I have added is named checkIfRepositoryIsBare and I am passing a string 'checkIfRepositoryIsBare'.

Similar to how getTopLevelWorkingDirectory passes 'getTopLevelWorkingDirectory' as an argument.

Did I make a mistake with the string? One thing I noticed that seems relevant is that there is 'isGitRepostory' and 'getTopLevelWorkingDirectory', with the prior calling the latter. I did not copy this pattern and only have checkIfRepostoryIsBare. Perhaps that is part of my problem.

Sorry to take up your time with this.

shiftkey: We can't follow this pattern for `isGitRepository` because it defers to `getTopLevelWorkingDirectory` for it's Git operation.

The pattern we're trying to follow here is **if a function performs a Git operation, ensure it's named in a way that we can trace back to later**.

What I was looking to ensure here is that the function name itself is in sync with the string we pass into `git()`, and I think we're on the same page. I just wanted to ensure the function name change mentioned above https://github.com/desktop/desktop/pull/5434#discussion_r213014336 is also applied here.

Daniel-McCarthy:
Okay I have updated this pull requests changes in response to the review. Thanks again @shiftkey for the feedback and assistance!

I believe I have addressed each of the requested changes:
-The ```checkIfRepositoryIsBare``` function in the /lib/git/rev-parse.ts is now named ```isBareRepository```.
-The string being passed in ```isBareRepository``` has been updated to match the name change.
-I added documentation above the ```isBareRepository``` function addressing the null return.
-I made use of the trim() function instead of including ```\n``` in the string.
-And I have made the simplification adding the ```|| false```.

Additionally I updated the unit tests to make use of the renamed function's new name.

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,30 @@ export async function getTopLevelWorkingDirectory(
...
+  try {
+    const result = await git(
+      ['rev-parse', '--is-bare-repository'],
+      path,
+      'isBareRepository'
+    )
+    return result.stdout.trim() === 'true'
+  } catch (e) {
+    if (e.message.includes('not a git repository')) {
+      return null
```
iAmWillShepherd: Curious why not just return false here? Given the function name, I'd expect a yes/no answer and not really care too much about the reason why it's not a bare repo.

app/src/ui/add-repository/add-existing-repository.tsx
```diff
@@ -82,14 +85,33 @@ export class AddExistingRepository extends React.Component<
       return
     }
 
+    const isBare = await isBareRepository(this.state.path)
+    if (isBare !== null && isBare === true) {
```
iAmWillShepherd: Returning just a `bool` cleans this up as `null` doesn't seem to serve any purpose here.

app/src/lib/git/rev-parse.ts
Daniel-McCarthy: That's a great point. Thanks for the feedback!

I would be happy to update the pull request again. To clarify, I should make this function return a ```Promise<boolean>``` instead of a ```Promise<boolean | null>``` and remove the "@returns null ..." section I added to the documentation, and cleanup the section mentioned in the add-existing-repository.tsx?

I just want to be on the same page regarding the full extent of the suggested changes.

Daniel-McCarthy: @shiftkey 
I am looking at implementing @iAmWillShepherd 's recommendations. However, since the method would no longer return null it would also cause one of the unit tests you have added to fail.

Should I go ahead with this and remove that particular unit test? If so then I shall do so, double check my changes, and then update the pull request.

shiftkey: If the test no longer catches a case we're interested in, removing it is fine

Daniel-McCarthy: I see. I think it might make more sense to change the unit test to check if it returns false since it is still a potential result from the function. I can update the test with that change and verify the tests still function as expected.

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,30 @@ export async function getTopLevelWorkingDirectory(
   return Path.resolve(path, relativePath)
 }
 
+/**
+ * Checks if the repository at a path is bare.
+ *
+ * @param path The path to the Git repository to check.
+ *
+ * @returns null if the path provided does not contain a Git repository.
```
shiftkey: This comment needs to be updated to indicate what `true` and `false` represent

app/src/lib/git/rev-parse.ts
```diff
@@ -52,6 +52,30 @@ export async function getTopLevelWorkingDirectory(
   return Path.resolve(path, relativePath)
 }
 
+/**
+ * Checks if the repository at a path is bare.
```
shiftkey: It looks like this code can handle a non-Git repository being at the path, but it errors if the path doesn't exist on disk. Can we capture that here and in the `@param path` value below?

app/src/lib/git/rev-parse.ts
Daniel-McCarthy: Ah, of course. I had forgotten to make that change before I made the commit. I'll fix that now.

app/src/lib/git/rev-parse.ts
Daniel-McCarthy: @shiftkey 
In regards to these two changes requested, would the following be sufficient wording? If so I will submit it in another commit.

```
/**
 * Checks if the repository at a path is bare.
 *
 * @param path The path to the repository to check. An error will be thrown if the path does not exist on disk.
 *
 * @returns true if the path contains a bare Git repository. Returns false if it is not bare or is not a Git repository.
 */
```

shiftkey: @Daniel-McCarthy works for me

searchgold:
5207

PR #5388, opened 2018-08-13
bump dugite to 2.18.0 and related packaging improvements

shiftkey:
This is a rollup of a number of changes to our Git library in the last few weeks:

 - Git 2.18.0
 - Git LFS 2.5.0
 - improved error parsing for "not a repository" code path - #5302
 - embedded Git on Linux now built against OpenSSL - https://github.com/shiftkey/desktop/issues/21
 - support for installing and using on 32-bit Windows #1214

Fixes: #5302

TODO:

 - [x] investigate failing test about Git no longer reporting files as `copied`

PR #5478, opened 2018-08-26
add explicit dependencies to libcurl and libsecret when packaging for deb and RPM

shiftkey:
~~🌵 🌵 🌵 🌵 **depends on #5388, go review that first** 🌵 🌵 🌵 🌵~~ 

As part of #5388 we've switched over the embedded Git to target `curl`, and the [list of shared libraries](https://github.com/desktop/dugite-native/pull/102#issuecomment-407065593) that `dugite-native` needs to function looks like this:

```
 0x0000000000000001 (NEEDED)             Shared library: [libcurl.so.4]
 0x0000000000000001 (NEEDED)             Shared library: [libz.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
```

This PR makes the `libcurl.so.4` shared library an explicit dependency, which is available in in different upstream packages dependent on the source. The other packages in the list are the defaults that `electron-builder` generates, which we still need.

 - [x] identify packages containing `libsecret-1.so.0` and add them to the list

cc @immackay who has been looking after the Arch Linux packaging

### DIFF
--- a/script/electron-builder-linux.yml
+++ b/script/electron-builder-linux.yml
@@ -20,6 +20,8 @@ deb:
     - libnss3
     # dugite-native dependencies
     - libcurl3
+    # keytar dependencies
+    - libsecret-1-0
 rpm:
   depends:
     # default Electron dependencies
@@ -28,3 +30,5 @@ rpm:
     - libnotify
     # dugite-native dependencies
     - libcurl
+    # keytar dependencies
+    - libsecret
--- a/script/electron-builder-linux.yml
+++ b/script/electron-builder-linux.yml
@@ -9,3 +9,22 @@ linux:
     - rpm
     - AppImage
   maintainer: 'GitHub, Inc <opensource+desktop@github.com>'
+deb:
+  depends:
+    # default Electron dependencies
+    - gconf2
+    - gconf-service
+    - libnotify4
+    - libappindicator1
+    - libxtst6
+    - libnss3
+    # dugite-native dependencies
+    - libcurl3
+rpm:
+  depends:
+    # default Electron dependencies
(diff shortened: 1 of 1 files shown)


immackay:
Oh my thank you so much. This has been causing a lot of dependency issues with people that have their own version of libcurl installed.

shiftkey:
Dropping from `1.4.0` as #5388 is not in that milestone.

shiftkey:
@outofambit now that #5388 has been merged into `master` this will be handy to also have in for the next update. Let me know how you feel about this - it's only specific to our Linux packaging work, but it's one less thing I need to manage over on my fork.

shiftkey:
> for my own context, can you point me to where this was being handled downstream?

Not sure I understand what "this" is - can you elaborate?

outofambit:
@shiftkey ah sorry i was referring to how you were handling libCurl in the past but i just reread the PR and see that this wasn't an issue until #5388, correct?

shiftkey:
@outofambit the saga of _upstream_ changes is:

 - linking error reported for some Linux users: https://github.com/shiftkey/desktop/issues/21
 - recompile Git to use plain `libcurl`: https://github.com/desktop/dugite-native/pull/102
 - upgrade `dugite` and flag this change: https://github.com/desktop/dugite/releases/tag/v1.70.0
 - upgrade Desktop itself: #5388

This PR is just the last piece of the puzzle, to ensure the generated packages define the right dependencies at install time (users had to do this via trial and error before).

Issue #4588, opened 2018-05-03
Detect Merge Conflicts


shiftkey:
**Goal**: As a user and consumer, when I am considering merging another branch into my own and _before merging_, I want to:
 * know whether I'll get into a conflict
 * be warned on how bad the conflict will be 
so that I can choose the best time to merge and be mentally prepared for it.

### Background
Built upon the [Compare Branch feature](https://github.com/desktop/desktop/issues/3956) that's about to land in 1.2. This is demonstrated in the initial mockups as a message below the merge button, like this:

<img width="236" alt="conflict-mockup" src="https://user-images.githubusercontent.com/359239/39558810-1b09433a-4ed5-11e8-87ba-19d7c57d90e0.png">

Implementation-wise, this is how we will accomplish the goal:
 - **know whether I'll get into a conflict** = test merging another branch without actually performing the merge 
 - **be warned on how bad the conflict will be** = display a number of files that will be conflicted if the user decides to proceed

### Interactive Example

I'm going to use https://github.com/desktop/desktop/pull/3675 as an example here, because this currently has a merge conflict.

#### Web Editor Flow

Here's what you see when you view the pull request:

<img width="726" src="https://user-images.githubusercontent.com/359239/39559043-a60edf5c-4ed6-11e8-8dad-93412db9caf3.png">

If we try and resolve this conflict in the editor, we see that it's just one file and one conflicted change:

<img width="1200" src="https://user-images.githubusercontent.com/359239/39559094-0ee2ca3e-4ed7-11e8-8c94-d25521080ab2.png">

#### Current Compare Branch Flow

There's no indicator when you compare to `master` that you're gonna have a bad time:

<img width="409" src="https://user-images.githubusercontent.com/359239/39559263-2fa7be2c-4ed8-11e8-811e-230defa9d45f.png">

When you merge `master` into the PR branch, you'll see this error:

<img width="693" src="https://user-images.githubusercontent.com/359239/39559268-3f344996-4ed8-11e8-82ac-052f7314be2a.png">

And when you switch to the Commit tab you'll see there's our conflicted file to resolve:

<img width="1211" src="https://user-images.githubusercontent.com/359239/39559279-506d1404-4ed8-11e8-9132-6cd4edaa3c46.png">

### Command Line Example

If you use `git merge --no-commit [other-branch]` to test this, it will still apply the result to the index - requiring a `git merge --abort` to undo the test merge. So how can we do this in Git _without_ actually performing the merge?

[...]

PR #5493, opened 2018-08-28
add parseMergeResult functionality and tests to detect merge conflicts 

shiftkey:
This PR is the underlying Git work to support #4588. I generated a bunch of test data by scanning the `desktop/desktop`, `electron/electron` and `microsoft/vscode` repositories for examples of merge conflicts - these are now baked as tests into the project, and caught some corner cases I hadn't uncovered with the original code.

For the moment `parseMergeResult`  just scans the Git output to build up an array of entries, to then determine what files have conflict. But there's already details in the diff output about _how_ files conflict that we can use down the track.

For reference, this is how I generated the merge conflict files - if there's interest in me adding this as something more reusable to regenerate tests cases when we upgrade `dugite` let me know and we can figure out how to do that:

```ts
  const repo = new Repository('/Users/shiftkey/src/vscode', -1, null, false)
  const allBranches = await getBranches(repo)
  const start =
    allBranches.find(b => b.nameWithoutRemote === 'master') || null

  if (start == null) {
    throw new Error('Unable to find current branch!')
  }

  await checkoutBranch(repo, null, start)

  const status = await getStatusOrThrow(repo)

  if (status.currentBranch == null) {
    throw new Error('Unable to find current branch!')
  }

  const ours =
    allBranches.find(b => b.nameWithoutRemote === status.currentBranch) ||
    null

  if (ours == null) {
    throw new Error('Unable to find ours!')
  }

  const otherBranchesToCompare = allBranches.filter(
    b => b.tip.sha !== ours.tip.sha
  )

  for (const theirs of otherBranchesToCompare) {
    const fileName = `merge-${theirs.nameWithoutRemote}-into-${
      ours.name
    }.txt`.replace('/', '-')

    const mergeBase = await getMergeBase(repo, ours.tip.sha, theirs.tip.sha)

    if (mergeBase === ours.tip.sha) {
      continue
    }

    if (mergeBase === theirs.tip.sha) {
      continue
    }

      ['merge-tree', mergeBase, ours.tip.sha, theirs.tip.sha],
      repo.path,
      { maxBuffer: 100 * 1024 * 1024 }
    )
    const stdout = result.stdout

    if (stdout.indexOf('<<<<<<') === -1) {
      continue
    }

    const relativePath = Path.join(
      __dirname,
      `../../fixtures/merge-parser/vscode/${fileName}`
    )

    await FSE.writeFile(relativePath, stdout)
  }
})
```


app/src/lib/git/merge.ts
```diff
@@ -21,3 +25,42 @@ export async function getMergeBase(
...
+ * @param repository The repository containing the branches to merge
+ * @param ours The current branch
+ * @param theirs Another branch to merge into the current branch
+ */
+export async function mergeTree(
+  repository: Repository,
+  ours: Branch,
+  theirs: Branch
+): Promise<MergeResult | null> {
+  console.time('getMergeBase')
```
shiftkey: I'm going to clean these up after doing the UI work, but for now these are nice to help identify chronic cases with any of the operations underneath.

app/src/lib/git/merge.ts
```diff
@@ -9,15 +13,73 @@ export async function merge(
   await git(['merge', branch], repository.path, 'merge')
 }
 
+/**
+ * Find the base commit between two refs
+ *
+ * @returns the commit id of the merge base, or null if the two refs do not
```
iAmWillShepherd: We should document what type of string we're looking for for the _*Ref_ parameters (like is it expecting a name of a branch or tag, a commit, or some other value).

app/src/lib/git/merge.ts
```diff
@@ -9,15 +13,73 @@ export async function merge(
...
   secondRef: string
-): Promise<string> {
+): Promise<string | null> {
   const process = await git(
     ['merge-base', firstRef, secondRef],
     repository.path,
-    'merge-base'
+    'merge-base',
+    {
+      successExitCodes: new Set([0, 1]),
```
iAmWillShepherd: What do these exit codes mean? Is it possible to pull these codes into constant so we don't have to throw around magic numbers?

app/src/lib/git/merge.ts
```diff
@@ -9,15 +13,73 @@ export async function merge(
...
+  ours: Branch,
+  theirs: Branch
+): Promise<MergeResult | null> {
+  console.time('getMergeBase')
+  const mergeBase = await getMergeBase(repository, ours.tip.sha, theirs.tip.sha)
+  console.timeEnd('getMergeBase')
+
+  if (mergeBase === null) {
+    // TODO: return a better result to indicate something is wrong
+    return null
```
iAmWillShepherd: Could we throw an error here or will you change this as a part of some other work you have planned?

app/src/lib/git/merge.ts
```diff
@@ -9,15 +13,73 @@ export async function merge(
...
+  }
+
+  if (mergeBase === ours.tip.sha) {
+    return { kind: MergeResultKind.Success, entries: [] }
```
iAmWillShepherd: This can be pulled out into a variable since it's being used below.

app/src/lib/git/merge.ts
shiftkey: The docs for `merge-base` say `<commit>` but my testing indicates this can be any commit/ref/tag/`HEAD`. I'll make it clearer in the title (documenting each `@param` felt noisy at the time):

> Find the base commit between two refs

app/src/lib/merge-tree-parser.ts
```diff
@@ -0,0 +1,194 @@
+import { IMergeEntry, MergeResult, MergeResultKind } from '../models/merge'
+
+function updateCurrentMergeEntry(
+  entry: IMergeEntry | undefined,
+  context: string,
+  blobSource: {
```
iAmWillShepherd: This should be pulled out into it's own type.

app/src/lib/git/merge.ts
shiftkey: This is handled in #5495 because of a corner case I wrote more about here: https://github.com/desktop/desktop/issues/4588#issuecomment-417382948

```
  if (mergeBase === null) {
    return { kind: MergeResultKind.Invalid }
  }
```

app/src/lib/git/merge.ts
shiftkey: The docs aren't really clear on what exit codes are expected here but from my testing showed `0` means a common ancestor was found. `1` means one could not be found (no `stdout` or `stderr` occurred).

Given the rest of the function is rather boring does introducing constants here help readability, or is there something else we can do here?

iAmWillShepherd: I think it's a readability improvement. If we can pass the exit codes we care about to the `git` function, I think we should always use a name instead of numbers. I've seen this in a few places now, so I figured passing these codes will become more common in the future.

app/src/lib/git/merge.ts
iAmWillShepherd: Yea, changing the name of the param to `committish` may also work. I agree that having the param docs is a bit noisey.a

app/src/lib/git/merge.ts
shiftkey: >  I've seen this in a few places now, so I figured passing these codes will become more common in the future.

Are we talking about `expectedErrors` (where an error message matches a known regex, and we map this to an enum in code) rather than `successExitCodes` (the raw numbers)?

I'm definitely in favour of preferring `expectedErrors` where it makes sense, but we don't get any `stdout` or `stderr` in this case to indicate anything went wrong - just the `1` exit code. [Here's the relevant Git source code, for reference](https://github.com/git/git/blob/2f743933341f276111103550fbf383a34dfcfd38/builtin/merge-base.c#L11-L28).

app/src/lib/merge-tree-parser.ts
```diff
@@ -0,0 +1,196 @@
...
+const blobEntryRe = /^\s{2}(result|our|their|base)\s+(\d{6})\s([0-9a-f]{40})\s(.+)$/
+
+/**
+ * Parse the Git output of a merge-tree command to identify whether it
+ * has detected any conflicts between the branches to be merged
+ *
+ * @param text the stdout from a `git merge-tree` command
+ *
+ */
+export function parseMergeResult(text: string): MergeResult {
```
iAmWillShepherd: This seems like a good thing to add some tests for.

shiftkey: Check out `app/test/unit/git/parse-merge-result-test.ts`

iAmWillShepherd: ahh, it's at the bottom and the diff is collapsed.

PR #5494, opened 2018-08-28
Fixes regex issue that caused only one issue number to be matched

iAmWillShepherd:
Context is in this comment https://github.com/desktop/desktop/pull/5143#issuecomment-406033398

**After**
screen shot 2018-08-28 at 3 26 32 pm


shiftkey:
@iAmWillShepherd can you add a screenshot here to show the before/after of this change?

iAmWillShepherd:
@shiftkey I still have not figured out how to render this with spaces in-between, but hope to figure it out after lunch.

screen shot 2018-08-28 at 12 49 19 pm

app/src/ui/lib/join.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Insert a separator element between each entry of an array of elements
+ *
+ * @param elements
```
shiftkey: Yo dawg, I meant to write some docs but totally forgot

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -22,8 +23,8 @@ const ReleaseNoteHeaderRightUri = encodePathAsUrl(
   'static/release-note-header-right.svg'
 )
 
-const externalContributionRe = /^(.*)(#\d+)(.*)(@[a-zA-Z0-9\-]+)!(.*)$/
-const otherContributionRe = /^(.*)(#\d+)(.*)$/
+const externalContributionRe = /^(.*?)((\s?\#\d+\s?)+)(.*)(@[a-zA-Z0-9\-]+)!(.*)$/
```
shiftkey: This regex is getting pretty gnarly - are we able to write some tests against this work to sanity check us against future regressions?

iAmWillShepherd: Sure thing

app/src/ui/lib/join.ts
iAmWillShepherd: I'm not even sure we need the docs for the params, I added some anyways, but I'm fine with removing later if y'all are in agreement.

app/src/ui/lib/join.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param separator
+ */
+export function join(
```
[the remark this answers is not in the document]
shiftkey: @outofambit I'm curious what pulling in a dependency like that would do to our overall bundle size (most of the unused code should get stripped out, but 🤷‍♂️ ) and also what other benefits we'd get out of it. Happy to move that into a tech debt task for someone to explore and propose more formally.

Also, it looks like there are TS declarations available for ramda already: https://www.npmjs.com/package/@types/ramda

app/src/ui/lib/join.ts
```diff
@@ -0,0 +1,29 @@
...
+  // build up the array by using a reducer
+  const output = elements.reduce(
+    (prev: Array<JSX.Element | string>, current: JSX.Element) => {
+      return [...prev, current, separator]
+    },
+    []
+  )
+
+  // remove the last element from the generated list
+  return output.slice(0, output.length - 1)
```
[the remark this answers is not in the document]
shiftkey: Yep. Suggestions for the comment on the previous line area welcome.

app/src/ui/lib/join.ts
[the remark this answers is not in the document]
outofambit: @shiftkey i think that's a great idea. i'd be happy to put together some more formal thoughts.

(according to https://bundlephobia.com/result?p=ramda@0.25.0 ramda is `77kb` minified in its entirety)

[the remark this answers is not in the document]
outofambit: see #5499 for future discussion

app/src/ui/lib/join.ts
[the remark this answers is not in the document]
outofambit: i would suggest "remove the trailing delimiter from the generated list"

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -33,38 +36,58 @@ function accountUrl(name: string): string {
...
+  const trimmed = issueNumberLine.trim()
+  const issueNumbers = trimmed.split(' ')
+
+  if (issueNumbers.length === 0) {
+    return null
+  }
+
+  const linkifiedIssueNumbers = issueNumbers.map((issueNumber, index) => {
+    return (
+      <LinkButton key={index} uri={desktopIssueUrl(issueNumber)}>
```
niik: Could we do something similar to

```ts
return (
  <>
    {index === 0 ? null : ' '}
    <LinkButton key={index} uri={desktopIssueUrl(issueNumber)}>{issueNumber}</LinkButton>
  </>
)
```

And thus avoid the `join` method altogether?

app/src/ui/lib/join.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Insert a separator element between each entry of an array of elements
+ *
+ * @param elements - A list of components
+ * @param separator - A string or component that will be placed between every element
+ */
+export function join(
```
niik: I believe we could simplify this method and make it generic at the same time by leveraging the index parameter in `reduce`

```ts
export function join<T, T1>(arr: T[], separator: T1): ReadonlyArray<T | T1> {
  return arr.reduce(
    (prev: Array<T | T1>, current: T, ix: number) =>
      ix === 0 ? [current] : [...prev, separator, current],
    []
  )
}
```

I feel the need to point out that we should only use this method for small arrays as the cost will increase quite dramatically for larger arrays.

I worry though that using something like this for JSX will mean that we'll have duplicated separators which is fine for strings but not so much for JSX elements.

app/src/models/release-notes.ts
```diff
@@ -18,3 +18,6 @@ export type ReleaseSummary = {
   readonly bugfixes: ReadonlyArray<ReleaseNote>
   readonly other: ReadonlyArray<ReleaseNote>
 }
+
+export const externalContributionRe = /^(.*?)((\s?\#\d+\s?)+)(.*)(@[a-zA-Z0-9\-]+)!(.*)$/
+export const otherContributionRe = /^(.*?)((\s?\#\d+\s?)*)(.*)$/
```
niik: Could we please document these? I don't understand the distinction between external and other contributions.

app/test/unit/in-app-releast-notes-test.ts
```diff
@@ -0,0 +1,50 @@
+import { expect } from 'chai'
+import {
+  externalContributionRe,
+  otherContributionRe,
+} from '../../src/models/release-notes'
+
+//todo: test everything
+
+describe('externalContributionRe', () => {
+  it.only('Matches messages that addresses one issue', () => {
```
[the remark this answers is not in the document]
iAmWillShepherd: Will do.
 
I wasn't actually done adding all tests. Going to go ahead and remove the ready for review label while I address all feedback.

shiftkey:
> I think we should repurpose our `RichText` component which already does all the heavy lifting required to linkify issues, and mentions rather than reinvent the wheel here.

There's some tweaks to the parsing that I need to add but I already have a decent test suite for this, so I'm 👍 on this approach here as we can throw away this custom code.

shiftkey:
I've opened https://github.com/desktop/desktop/pull/5505 which should tackle most of the feedback raised in this PR

shiftkey:
#5505 was merged, making this PR obsolete

Issue #788, opened 2017-01-11
warn user when trying to commit with detached HEAD


shiftkey:
In case the user is in the detached HEAD scenario, we should warn the user in case they try to commit, and help them get back to a safe state.

xref https://github.com/desktop/desktop/pull/786#issuecomment-271798234

shiftkey:
Bumping this because the current UI still shows `master` as the branch you're going to commit to

<img width="864"  src="https://user-images.githubusercontent.com/359239/37745207-ad03fd5e-2dc7-11e8-87da-4e3ffd9d9011.png">

PR #5500, opened 2018-08-29
Disable affected menu items when on detached HEAD

say25:
Noticed this when looking into https://github.com/desktop/desktop/issues/788. This does **NOT** address the core problem mentioned in that issue but similarly addresses concerns of confusion that can arise when you are on a detached head. There are menu items under the branch drop down that currently do nothing when in this state.


outofambit:
for reference, this is what the menu looks like when you'r in a detached head state.

screen shot 2018-08-29 at 10 03 40 am

outofambit:
thank you @say25!

shiftkey:
Moving this to the 1.4.0 milestone as I'd like to get some fixes out today/tomorrow.

PR #5505, opened 2018-08-29
use the RichText component when rendering release notes

shiftkey:
This PR supersedes #5494 and instead uses our current `RichText` component. I caught a couple of corner cases with our release notes that I added tests for while I was in here.

One caveat: `RichText` relies on the `repository` prop being set to render links (because it was rendering things inside the repository view and needs to know the GitHub API details), which might have been why I avoided it initially. Not sure how I feel about stubbing a `Repository` like I have in here, but It Work™ and I think I can follow up by isolating the fields that the component _actually_ needs here.

### DIFF
--- a/app/src/ui/release-notes/release-notes-dialog.tsx
+++ b/app/src/ui/release-notes/release-notes-dialog.tsx
@@ -14,6 +14,7 @@ import { DialogHeader } from '../dialog/header'
 
 import { RichText } from '../lib/rich-text'
 import { Repository } from '../../models/repository'
+import { getDotComAPIEndpoint } from '../../lib/api'
 
 const repository = new Repository(
   '',
@@ -24,7 +25,7 @@ const repository = new Repository(
     owner: {
       id: null,
       login: 'desktop',
-      endpoint: 'https://api.github.com/',
+      endpoint: getDotComAPIEndpoint(),
       hash: '',
     },
     private: false,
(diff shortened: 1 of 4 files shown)


app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -8,11 +8,38 @@ import { ReleaseNote, ReleaseSummary } from '../../models/release-notes'
...
+      login: 'desktop',
+      endpoint: 'https://api.github.com/',
+      hash: '',
+    },
+    private: false,
+    parent: null,
+    htmlURL: 'https://github.com/desktop/desktop',
+    defaultBranch: 'master',
+    cloneURL: 'https://github.com/desktop/desktop',
+    endpoint: 'https://api.github.com/',
```
iAmWillShepherd: `endpoint` should be set to `https://api.github.com` since the `Tokenizer` is looking for that to return the right URL.

https://github.com/desktop/desktop/blob/50f7b88ff834119622018e953cd2f1fa444902c3/app/src/lib/text-token-parser.ts#L192

https://github.com/desktop/desktop/blob/50f7b88ff834119622018e953cd2f1fa444902c3/app/src/lib/api.ts#L808-L810

https://github.com/desktop/desktop/blob/50f7b88ff834119622018e953cd2f1fa444902c3/app/src/lib/api.ts#L827-L839

At the moment, clicking on an @-mention navigates you to api.github.com

shiftkey:
:apple:

PR #5513, opened 2018-08-30
Show border when commit description is expanded.

### LINKED ISSUE #5506: show border when commit description is expanded
On current `master` we merged #5471 which expands a long commit message to only the space that's needed (rather than filling the whole space). 

On macOS (I'm sure this is the same on other OSes) when the commit is collapsed, there's a 1px grey border separating the commit description from the file list:

When this is expanded, the border is no longer displayed:

This wasn't an issue before because the expanded commit description would take up the whole panel.

Repro steps:

 - create a new repository
 - add some changes so you can commit
 - make a long commit description so that you trigger the expander
 - switch to the history list, select the new commit
 - expand the commit description

aryyya:
Fixes #5506.

Previously the border between the description and the file list was not displayed when a long description was expanded:

previous behavior 01

Now the border is shown after expansion:

new behaviour 01

A super long description that takes up the whole page will have it's border hidden (the original behavior before #5471):

new behaviour 02

Please disregard the solid black line across the bottom of the images, they are artifacts from Windows Snipping Tool (I need to find a window screenshot tool for Windows that is as elegant as the one built in to macOS, that one even adds a fancy drop shadow).

Let me know if there are any problems, or if you've got any suggestions for this fix, thanks!

### DIFF
--- a/app/src/ui/history/commit-summary.tsx
+++ b/app/src/ui/history/commit-summary.tsx
@@ -120,7 +120,7 @@ function messageEquals(x: Commit, y: Commit) {
 export class CommitSummary extends React.Component<
   ICommitSummaryProps,
   ICommitSummaryState
-  > {
+> {
   private descriptionScrollViewRef: HTMLDivElement | null = null
   private readonly resizeObserver: ResizeObserver | null = null
   private updateOverflowTimeoutId: number | null = null
@@ -154,7 +154,8 @@ export class CommitSummary extends React.Component<
 
   private onResized = () => {
     if (this.descriptionRef) {
-      const descriptionBottom = this.descriptionRef.getBoundingClientRect().bottom
+      const descriptionBottom = this.descriptionRef.getBoundingClientRect()
+        .bottom
       this.props.onDescriptionBottomChanged(descriptionBottom)
     }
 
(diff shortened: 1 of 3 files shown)


aryyya:
Yes, I did explore a CSS option but couldn't get it to work. Thanks for the idea, I will explore that further now.

outofambit:
@aryyya thank you! 😄 I'm not totally sure that the approach I described will work, but I do think it (or something similar) is worth a try.

aryyya:
I can't figure out a simpler way to do this via CSS and a spacer div as the border.

Using a spacer div and  giving `commit-summary-description-scroll-view` more flex weight than the added `commit-spacer` sibling results in weird behavior. At different zoom levels for example the size of the border will noticeably fluctuate depending on how long the description is, and at the higher zoom levels it won't disappear at all.

If only there was a way to hide a flex item once there was no space available for it instead of changing it's size, or wrapping it.

Do you have any other ideas?

shiftkey:
Dropping `ready-for-review` while we work out the best solution here.

cc @desktop/design for thoughts

outofambit:
@aryyya i tried another idea locally and got nowhere 😓 i think i'm good with this as-is. thanks so much for your work on this!

PR #5495, opened 2018-08-28
display merge conflict hint before doing the merge

shiftkey:
~~🌵 🌵 🌵 **Builds upon #5493, go review that first** 🌵 🌵 🌵~~

This PR adds the UX for #4588 in two places, the Compare tab and the Merge Dialog. Both of these areas are decision points when the user decides to merge in some other branch into their current branch, and we want to help guide them.

I'm working from these mocks in this comment https://github.com/desktop/desktop/issues/4588#issuecomment-416663744 but these are likely to change:

screen shot 2018-08-28 at 3 26 28 pm

 - [x] rebase PR on `master` once #5493 is merged
 - [x] wire up the compare tab to show this new information
 - [x] wire up the merge dialog to show this new information
 - [x] in Merge dialog, indicator is shown when selected branch is same as current (should be hidden) (#5514) 
 - [x] polish the styles and layout once we're happy with the mocks (cc @desktop/design) (#5514)
 - [x] investigate double `mergeTree` work in `AppStore`

PR #5514, opened 2018-08-30
polish merge status

donokuda:
(🎯Targets the `display-merge-status-in-compare-tab` branch.)

This pull request starts some work on polishing the merge status.

I'm running into an issue where the UI is pretty jarring when the status gets updated from "Pending" to the " ✅/ ⚠️" status 😞:

| ✅:cry: | ⚠️:sob: |
|---|---|
|merge-pending-green |merge-pending-x |

I tried doing some quick CSS tricks like animating the max-height, but I think a better approach is to prototype out a full transition instead. The transition could include fading in the content as well as doing a transitions with the octicons themselves.

### DIFF
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -1,0 +1,38 @@
+/**
+ * Wrap a promise in a minimum timeout, so that it only returns after both the
+ * timeout and promise have completed.
+ *
+ * This is ideal for scenarios where a promises may complete quickly, but the
+ * caller wants to introduce a minimum latency so that any dependent UI is
+ *
+ *
+ * @param action the promise work to track
+ * @param timeout the minimum time to wait before resolving the promise (milliseconds)
+ */
+export function promiseWithMinimumTimeout<T>(
+  action: () => Promise<T>,
+  timeout: number
+): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    let timeoutExpired = false
+    let result: T | null = null
+
+    const resolveIfTimeout = () => {
+      if (result != null && timeoutExpired) {
+        resolve(result)
+      }
+    }
+
+    window.setTimeout(() => {
+      timeoutExpired = true
+      resolveIfTimeout()
+    }, timeout)
+
+    action()
+      .then(r => {
+        result = r
+        resolveIfTimeout()
+      })
+      .catch(reject)
+  })
+}
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -17,21 +17,22 @@ export function promiseWithMinimumTimeout<T>(
     let timeoutExpired = false
     let result: T | null = null
 
-    const resolveIfTimeout = () => {
+    const resolveIfBothDone = () => {
       if (result != null && timeoutExpired) {
         resolve(result)
+        result = null
       }
     }
 
     window.setTimeout(() => {
       timeoutExpired = true
-      resolveIfTimeout()
+      resolveIfBothDone()
     }, timeout)
 
     action()
       .then(r => {
         result = r
-        resolveIfTimeout()
+        resolveIfBothDone()
       })
       .catch(reject)
   })
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -7,7 +7,7 @@
  *
  *
  * @param action the promise work to track
- * @param timeout the minimum time to wait before resolving the promise (milliseconds)
+ * @param timeout the minimum time to wait before resolving the promise (in milliseconds)
  */
 export function promiseWithMinimumTimeout<T>(
   action: () => Promise<T>,
--- a/app/test/unit/promise-with-timeout-test.ts
+++ b/app/test/unit/promise-with-timeout-test.ts
@@ -6,20 +6,17 @@ jest.useFakeTimers()
 
 describe('promiseWithMinimumTimeout', () => {
   it('handles promise finishing before timeout', async () => {
-    let promiseCallbackFired = false
+    const resolveMock = jest.fn().mockImplementation(resolve => resolve(42))
 
     const fastPromise = new Promise<number>(resolve => {
-      window.setTimeout(() => {
-        resolve(42)
-        promiseCallbackFired = true
-      }, 100)
+      window.setTimeout(() => resolveMock(resolve), 100)
     })
 
     const promise = promiseWithMinimumTimeout(() => fastPromise, 500)
 
     // promise completes
     jest.advanceTimersByTime(250)
-    expect(promiseCallbackFired).is.true
+    expect(resolveMock.mock.calls.length).equals(1)
 
     // timeout completes
     jest.advanceTimersByTime(250)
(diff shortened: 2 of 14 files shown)


shiftkey:
@donokuda I'm looking into some corner cases about the flow itself after testing this more, but I think the "Merge into Current branch" dialog didn't get the same love with the indicator:

shiftkey:
@donokuda I'm juggling keeping downstream PRs #5493 and #5495 focused on their specific changes, so don't worry about those conflicts you're seeing - once you're happy with the UX I'll rebase it so that it's easy to review just what's changed here.

shiftkey:
@desktop/design I think this is pretty close - I did some tidy up of the Merge dialog and it's now looking like the Compare tab:

screen shot 2018-08-31 at 11 38 48 am

The SCSS can probably be better organized - maybe even unified - but I'll leave that up to y'all to figure out what's important here.

donokuda:
😍There's a few spacing inconsistencies that I can jump in and clean up. Overall, thx @shiftkey! 

> The SCSS can probably be better organized - maybe even unified - but I'll leave that up to y'all to 
figure out what's important here.

👍This makes sense especially if the markup in both places are similar. I'll also take a quick look at that.

shiftkey:
Dropping in some fresh screenshots now that the commits are reorganized:




@desktop/design if you've got a local branch and want to keep working on this you'll need to reset it to the remote:

```
$ git fetch --all
$ git checkout polish-merge-status
$ git reset origin/polish-merge-status --hard
```

app/src/lib/promise.ts
```diff
@@ -0,0 +1,39 @@
...
+ * timeout and promise have completed.
+ *
+ * This is ideal for scenarios where a promises may complete quickly, but the
+ * caller wants to introduce a minimum latency so that any dependent UI is
+ *
+ *
+ * @param action the promise work to track
+ * @param timeout the minimum time to wait before resolving the promise (in milliseconds)
+ */
+export function promiseWithMinimumTimeout<T>(
```
shiftkey: @desktop/core @desktop/comrades this is probably the only non-UI change that needs a review as part of this PR. I'd love some input on it.

say25: I guess my main feedback would be add some unit tests? If this is something we intend on using going forward, it would be nice to know its been tested.

pennyluis:
@say25 commented on this pull request

app/src/lib/promise.ts
shiftkey: @say25 good shout. I'll see what I can do to control `window.setTimeout` now that we're on Jest.

outofambit: @shiftkey https://jestjs.io/docs/en/timer-mocks.html 😃

shiftkey: I've pushed 174cbb292 which covers the interesting scenarios to test for

shiftkey:
:apple:

donokuda:
I got one more small detail to push, and then I'll comment with some sort of emoji to indicate that my part is also ready to review.

donokuda:
Was taking a look at this feature in **xXdark modeXx** and saw that the text treatment between the merge dialog and the compare panel could be more consistent.

Specifically, the emphasized words are white while the rest of the sentence are a little gray so that the more important information stands out.

**Before** (👀at the panel to the left)
screen shot 2018-09-04 at 8 37 50 am

**After**
screen shot 2018-09-04 at 8 35 20 am

🥝My part is ready for review now and open to any feedback, especially around CSS stuff like https://github.com/desktop/desktop/blob/25a008527228234eef7743820f3af7c2c124b790/app/styles/ui/_merge-status.scss#L9-L16

app/test/unit/promise-with-timeout-test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const fastPromise = new Promise<number>((resolve, reject) => {
+      window.setTimeout(() => {
+        resolve(42)
+      }, 100)
+    })
+
+    const promise = promiseWithMinimumTimeout(() => fastPromise, 500)
+
+    // promise completes
+    jest.advanceTimersByTime(250)
```
outofambit: should we check that `promise` isn't resolved right after this?

app/test/unit/promise-with-timeout-test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const slowPromise = new Promise<number>((resolve, reject) => {
+      window.setTimeout(() => {
+        resolve(42)
+      }, 1000)
+    })
+
+    const promise = promiseWithMinimumTimeout(() => slowPromise, 500)
+
+    // timeout completes
+    jest.advanceTimersByTime(500)
```
outofambit: should we check that `promise` isn't resolved right after this?

shiftkey: I've pushed 6a2b963d7 which let's us check the state of the promise we control

app/test/unit/promise-with-timeout-test.ts
outofambit: hmmm fair point. i think we could get what we want with a jest function mock

```js
describe('promiseWithMinimumTimeout', () => {
  it('handles promise finishing before timeout', async () => {
    const resolveWrapper = jest.fn().mockImplementation((resolve) => resolve(42))
    const fastPromise = new Promise<number>((resolve, reject) => {
      window.setTimeout(resolveWrapper(resolve), 100)
    })
     const promise = promiseWithMinimumTimeout(() => fastPromise, 500)
     // promise completes
    jest.advanceTimersByTime(250)
    // this would be a little cleaner with jest expect instead of chai expect
    expect(resolveWrapper.calls.length).equals(0)
...
```

this would be a little cleaner with jest instead of chai expect, but yeah.

shiftkey: @outofambit maybe? let me try out how it feels

shiftkey:
@donokuda I added the UI work for "unrelated histories" in  https://github.com/desktop/desktop/pull/5495/commits/264e99032bbdfd83472fac9cc5fcca27843440f9 and I had to merge that in here to address a conflict, but I'll leave it up to you whether you want to restyle these or punt to a follow-up PR:

screen shot 2018-09-04 at 2 50 19 pm

screen shot 2018-09-04 at 2 50 14 pm

To test, use a repository that contains a `gh-pages` branch (like `https://github.com/prettier/prettier`) and choose that as the branch to merge into your current branch.

app/test/unit/promise-with-timeout-test.ts
shiftkey: Try out 16d996dd5 and see how it feels

shiftkey:
:apple:

app/test/unit/promise-with-timeout-test.ts
outofambit: 🎸rockin'!

shiftkey:
@donokuda thanks for the update. I might even be able to get to it as part of #5495 tomorrow if you don't have a chance today.

donokuda:
🥗 

**xXDark ModeXx**

| Merge dialog  | Compare tab   |
|---|---|
|  screen shot 2018-09-04 at 11 41 51 am |   screen shot 2018-09-04 at 11 41 43 am |

**Light mode**

| Merge dialog   | Compare tab  | 
|---|---|
|screen shot 2018-09-04 at 11 42 08 am |screen shot 2018-09-04 at 11 42 02 am |

nerdneha:
woo! this is looking great! can we get a gif of the less jarring transition between `pending` and `end result` now? :D (will also be helpful for QA and maybe help docs)

j-f1:
Would it be better flipped vertically (i.e. the message, then the icon, then the button? That way, the button’s position wouldn’t change so the user would be able to click it even if the status-getting finishes while they’re hovering.

shiftkey:
We've got two approvals here and I've got a couple of tasks left in #5495 to tackle, so I'm going to merge this in so I can stop juggling multiple PRs.

@j-f1 I'll let @desktop/design weigh in on that suggestion as this PR is about polishing the current layout - I have no strong feels either way.

Issue #4678, opened 2018-05-14
This destination already exists error displayed even after removing it from the file system


sverraest:
## Description

When trying to clone a repository to an existing folder on the filesystem the error "this destination already exists" is displayed. After removing it from the filesystem the error is still visible. The outcome should be that it is now possible to clone to that destination

## Version

* GitHub Desktop: 1.1.1

* Operating system: macOS Sierra 10.12.6

## Steps to Reproduce

1. Clone a repository to a folder which already has this repository
2. Observe error
3. Remove folder from filesystem

### Expected Behavior

You should be able to clone the repository now

### Actual Behavior

The error "this destination already exists" is still visible, no option to clone.

shiftkey:
I think I've stumbled on this before, and it'd be nice to hook into the `focus` event for the app and give the dialog a chance to refresh itself:

https://github.com/desktop/desktop/blob/7f02d93f4624aee8a9e46286e5a6525ea341b6bb/app/src/ui/index.tsx#L153-L163

This is how we update the details about the selected repository, if you were making changes in another tool and switched back to Desktop.

Daniel-McCarthy:
I am interested in taking this on. I have looked into this a bit already and have the basics of it implemented. However, I hoped to ask about it's intended behavior.

There are two immediate scenarios where a focus event could assist this:
- An error has appeared due to the folder already existing. However, should go away if the folder now no longer exists and the user returns to the window.
- No error is shown. However, the folder at the path now exists and should show an error.

Verifying the path on focus can handle both of these. However, it would show the error even for the default path that is given when the Clone Repository window is first opened next time ```onFocus``` is called. This might be somewhat unpleasant behavior.

Do we only want to verify the path when this error is showing? (If so, then it will not generate an error if the folder now exists, but they can find that out when clicking the 'Clone' button again.)

shiftkey:
I believe this has been fixed by #5518 but I'll test out our next beta and confirm that's the case.

PR #5518, opened 2018-08-30
Verify Clone-Repository path on focus

Daniel-McCarthy:
This pull request is in response to Issue #4678. It aims to prevent displaying the DestinationAlreadyExists error when the path has been deleted since selecting it. Additionally, it will display the error if the path has been created after it was selected.

However, there is a current upside and drawback to the current implementation.
- If a repository to clone is selected, the path will be checked on refocus to determine if it exists and will clear or throw the destination exists error.
    
    1. This prevents continuing to display the destination exists error when they have since deleted the folder.

    2. This prevents cloning into a folder that now exists but did not when the path was selected.

- It will not verify when a repository has not been selected.

    1. This stops the error from being thrown on the initial screen. Since the default path exists, if the user opens the dialog, switches to another window, and switches back, the DestinationAlreadyExists error would throw before they have done anything.

    2. However, this means that it will not automatically clear the error when a repository has not been selected yet.

Therefore, I would love suggestions on how to improve this. I would love to have it be able to clear the error when a repository is not selected while continuing to not throw the error on the default screen before a user has changed anything.

### DIFF
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -53,7 +53,7 @@ interface ICloneRepositoryState {
   /** The local path to clone to. */
   readonly path: string
 
-  /** A copy of the initial local path to clone to. */
+  /** The a copy of the default path value to compare if the current path is a different directory. */
   readonly initialPath: string
 
   /** Are we currently trying to load the entered repository? */
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -53,6 +53,9 @@ interface ICloneRepositoryState {
   /** The local path to clone to. */
   readonly path: string
 
+  /** A copy of the initial local path to clone to. */
+  readonly initialPath: string
+
   /** Are we currently trying to load the entered repository? */
   readonly loading: boolean
 
@@ -76,9 +79,11 @@ export class CloneRepository extends React.Component<
   public constructor(props: ICloneRepositoryProps) {
     super(props)
 
+    const defaultDirectory = getDefaultDir()
     this.state = {
       url: this.props.initialURL || '',
-      path: getDefaultDir(),
+      path: defaultDirectory,
+      initialPath: defaultDirectory,
       loading: false,
       error: null,
       lastParsedIdentifier: null,
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -387,7 +387,7 @@ export class CloneRepository extends React.Component<
   private onWindowFocus = () => {
     // Verify the path after focus has been regained in case changes have been made.
     const isDefaultPath =
-      this.state.initialPath === this.state.path.replace(/\/$/g, '')
+      this.state.initialPath === this.state.path
     const isURLNotEntered = this.state.url === ''
 
     if (isDefaultPath && isURLNotEntered) {
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -97,6 +97,8 @@ export class CloneRepository extends React.Component<
     if (initialURL) {
       this.updateUrl(initialURL)
     }
+
+    window.addEventListener('focus', this.onFocus)
   }
 
   public render() {
@@ -371,5 +373,12 @@ export class CloneRepository extends React.Component<
     this.props.onDismissed()
 
     setDefaultDir(Path.resolve(path, '..'))
+  }
+
+  private onFocus = () => {
+    // Verify the path after focus has been regained in case changes have been made.
+    if (this.state.url !== ``) {
+      this.updateAndValidatePath(this.state.path)
+    }
   }
 }
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -382,7 +382,18 @@ export class CloneRepository extends React.Component<
 
   private onFocus = () => {
     // Verify the path after focus has been regained in case changes have been made.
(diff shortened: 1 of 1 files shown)


Daniel-McCarthy:
After looking into this again, I have found a solution that I believe is fitting.

Previously I had set it to not validate the path if the `url` was empty, and a repository to clone had not been selected yet. This prevented the `DestinationExists` error from being generated before a repository was selected, but meant that the error would not go away if it had already been generated and the user then cleared the `url`.

Instead, I have made the initial path get stored in the `state` so that can be compared with. Now, if the `path` has not been changed and the `url` is empty/blank, then any `DestinationExists` errors will be cleared `onfocus`. Otherwise, it will validate the path like normal once the user has selected a repository or modified the path.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -97,6 +102,8 @@ export class CloneRepository extends React.Component<
     if (initialURL) {
       this.updateUrl(initialURL)
     }
+
+    window.addEventListener('focus', this.onFocus)
```
niik: Could we move this to `componentDidMount` and also add remove the listener in `componentWillUnmount` to avoid a dangling subscription and subsequent memory leak? See `dialog.tsx` for an example of how we've done this previously.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -372,4 +379,22 @@ export class CloneRepository extends React.Component<
 
     setDefaultDir(Path.resolve(path, '..'))
   }
+
+  private onFocus = () => {
+    // Verify the path after focus has been regained in case changes have been made.
+    const isDefaultPath =
+      this.state.initialPath === this.state.path.replace(/\/$/g, '')
```
niik: Could you elaborate on the regex here and why it's necessary? Keep in mind that forward slashes are only path separator characters on POSIX systems, Windows uses `\`.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -372,4 +379,22 @@ export class CloneRepository extends React.Component<
 
     setDefaultDir(Path.resolve(path, '..'))
   }
+
+  private onFocus = () => {
```
niik: Could we rename this to be more clear about the fact that this method is called when the window receives focus, not the component itself. In `dialog.tsx` we called it `onWindowFocus` which I think is a good name.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -53,6 +53,9 @@ interface ICloneRepositoryState {
   /** The local path to clone to. */
   readonly path: string
 
+  /** A copy of the initial local path to clone to. */
```
niik: Could we elaborate in this documentation and explain why we're keeping a copy of the path? To a new contributor or someone who isn't familiar with the problem this PR is trying to solve the current documentation is only providing information on what it is, not why it's there or what it's used for.

app/src/ui/clone-repository/clone-repository.tsx
Daniel-McCarthy: The reason for the regex was so that the equality will not fail when only a `/` has been added since it is still otherwise the same initial path.

For example, if I have the initialPath `.../GitHub` and want to clone desktop to `.../GitHub/desktop` I can quickly type in `/desktop` and see that it is already taken. So then if I backspace and remove the desktop and am left with `.../GitHub/` and the string equality would fail.

The regex was meant to avoid generating an error in this situation, but is by no means necessary. If you think I should remove it I can definitely do so. Otherwise I can extent it to also consider `\`.

niik: Thanks for the extra info! I think for now let's go without it and if we feel that it's not sufficient we should use something more robust and platform-independent to compare equality between paths such as `path.resolve`. Your call though. If you do decide to extend the check I'd appreciate a comment explaining what it's doing and why.

Daniel-McCarthy: No problem. I will remove the check in that case, I would prefer it to be easier to understand.
I will test these changes and update the pull request soon. Thanks for the feedback!

Daniel-McCarthy:
@niik Thanks again for the feedback. If there are any more suggestions I'd be happy to see what I can do.

I believe I have made the changes as requested and should be ready for review. Thanks!

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -97,6 +102,12 @@ export class CloneRepository extends React.Component<
     if (initialURL) {
       this.updateUrl(initialURL)
     }
+
+    window.addEventListener('focus', this.onWindowFocus)
```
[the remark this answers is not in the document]
Daniel-McCarthy: I am not quite sure I understand. It [looks to already be inside componentDidMount()](https://github.com/desktop/desktop/blob/e65343ca766eae01df3c12485f98a01f5c2f5625/app/src/ui/clone-repository/clone-repository.tsx#L100-L107). I have looked over it a few times and it does not appear to be in the constructor.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -53,6 +53,9 @@ interface ICloneRepositoryState {
   /** The local path to clone to. */
   readonly path: string
 
+  /** The a copy of the default path value to compare if the current path is a different directory. */
```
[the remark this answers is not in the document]
Daniel-McCarthy: Of course, I will make that change. That should cover it quite well.

app/src/ui/clone-repository/clone-repository.tsx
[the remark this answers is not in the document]
Daniel-McCarthy: No worries! I'll push the documentation commit now.

niik:
Thanks @Daniel-McCarthy!

PR #5521, opened 2018-08-30
Fixing status badge markdown so link takes you to the latest build 

chrispat:
The current link takes you to the overall list of builds rather than the latest build for the repo.


shiftkey:
@chrisrpatterson thanks for the improvement!

PR #5524, opened 2018-08-31
add writeup for sslCAInfo bundle issue

shiftkey:
Should have done this ages ago, but we still seem to be getting reports of this issue so here's my summary of it to help others recover.


docs/known-issues.md
```diff
@@ -123,4 +123,40 @@ Enabling Mandatory ASLR affects the MSYS2 core library, which is relied upon by
...
+> git config -l --show-origin
+```
+
+You should have an entry that looks like this:
+
+```
+file:"C:\ProgramData/Git/config" http.sslcainfo=[some value here]
+```
+
+2. Open `C:\ProgramData\Git\config` and remove the corresponding lines that look like this:
```
steveward: Should we add a line in this section about needing admin privileges to edit this file?

shiftkey: Probably, yes

shiftkey:
:apple:

PR #5528, opened 2018-08-31
fix 10k file test on appveyor

### LINKED ISSUE #5527: Appveyor builds are failing silently
I noticed that Appveyor builds seem like they are broken, even though they are reported as passing. Check out the build for a recent master commit (9c96a9f0cac29f1276d8f1174dd78c318b96201a): https://ci.appveyor.com/project/github-windows/desktop/build/10374#L797

But the status is reported as passing.

I'm not sure if the problem was related to moving to jest (see https://github.com/desktop/desktop/pull/5427)

outofambit:
fixes #5527 

The solution for this ended up being [`--detectOpenHandles`](https://jestjs.io/docs/en/cli.html#detectopenhandles), which gave me reliable failures for the 10k untracked file test. This test was only failing because it timed out, so I increased the time limit for that one test.

thanks @shiftkey and @streeter!

### DIFF
--- a/app/test/unit/git/status-test.ts
+++ b/app/test/unit/git/status-test.ts
@@ -86,25 +86,30 @@ describe('git/status', () => {
       expect(files[1].path).to.equal('docs/OVERVIEW.md')
     })
 
-    it('Handles at least 10k untracked files without failing', async () => {
-      const numFiles = 10000
-      const basePath = repository!.path
+    it(
+      'Handles at least 10k untracked files without failing',
+      async () => {
+        const numFiles = 10000
+        const basePath = repository!.path
 
-      await mkdir(basePath)
+        await mkdir(basePath)
 
-      // create a lot of files
-      const promises = []
-      for (let i = 0; i < numFiles; i++) {
-        promises.push(
-          FSE.writeFile(path.join(basePath, `test-file-${i}`), 'Hey there\n')
-        )
-      }
-      await Promise.all(promises)
+        // create a lot of files
+        const promises = []
+        for (let i = 0; i < numFiles; i++) {
+          promises.push(
+            FSE.writeFile(path.join(basePath, `test-file-${i}`), 'Hey there\n')
+          )
+        }
+        await Promise.all(promises)
 
-      const status = await getStatusOrThrow(repository!)
-      const files = status.workingDirectory.files
-      expect(files.length).to.equal(numFiles)
-    })
+        const status = await getStatusOrThrow(repository!)
+        const files = status.workingDirectory.files
+        expect(files.length).to.equal(numFiles)
+      },
+      // needs a little extra time on CI
+      25000
+    )
 
     it('returns null for directory without a .git directory', async () => {
       repository = setupEmptyDirectory()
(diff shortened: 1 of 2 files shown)


RiXXX91:
H

shiftkey:
This seems to be the expected behaviour from that last build:

 - `jest` test suite reports an error, returns exit code 1 (because one failure)
 - `yarn` detects this non-zero exit code and returns that
 - Appveyor build runner detects `yarn` non-zero exit code and 

screen shot 2018-09-04 at 10 00 58 am

The [build](https://ci.appveyor.com/project/github-windows/desktop/build/10374) linked to in #5527 shows a different sort of error - timeout/`EMFILE` errors - and we don't see jest return in the same way - it just continues on to the next test suite:



Perhaps we need to set [this flag](https://jestjs.io/docs/en/cli.html#forceexit)?

screen shot 2018-09-04 at 10 04 35 am

app/test/unit/git/status-test.ts
```diff
@@ -86,25 +86,30 @@ describe('git/status', () => {
...
+          )
+        }
+        await Promise.all(promises)
+
+        const status = await getStatusOrThrow(repository!)
+        const files = status.workingDirectory.files
+        expect(files.length).to.equal(numFiles)
+      },
+      // needs a little extra time on CI
+      25000
```
outofambit: the only non-formatting change in this file

j-f1: https://github.com/prettier/prettier/issues/5061

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "cli": "ts-node --require ./app/test/globals.ts --require ./app/src/cli/dev-commands-global.ts app/src/cli/main.ts",
     "test:integration": "cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe --auto-servernum -- mocha -t 30000 --require ts-node/register app/test/integration/*.ts",
-    "test:unit": "cross-env ELECTRON_RUN_AS_NODE=1 node_modules/.bin/electron ./node_modules/jest/bin/jest --silent --config ./jest.config.js",
+    "test:unit": "cross-env ELECTRON_RUN_AS_NODE=1 node_modules/.bin/electron ./node_modules/jest/bin/jest --detectOpenHandles --silent --config ./jest.config.js",
```
SimenB: you probably don't want this flag on at all times - the overhead of collecting everything async is pretty atrocious (30%+ overhead on node 8)

outofambit: thanks @SimenB! that's good context to have.

it seems to me like the tradeoff is speed vs possible false positives and i'm not sure that speed is worth the risk of missing failing tests. but perhaps there's another way?

SimenB: If you can put together a test case showing that jest exits with `0` when a test times out, we can fix it 🙂 That's a bug, you shouldn't get false positives

outofambit: 💖

SimenB: We do have an open PR for when a test synchronously takes a long time without timing out (which we currently only detect if it happens async), might be it? https://github.com/facebook/jest/pull/6950

PR #5538, opened 2018-09-04
nvm and pyenv and docs

### LINKED ISSUE #5340: have an official/supported node version manager
Choosing a node version manager introduces additional overhead to getting started with github desktop. It would be nice to have documentation (and the corresponding config files) to get started with a particular node version manager thing on desktop.

Probably some docs and a config file. `nvm`? `asdf-nodejs`?

outofambit:
resolves #5340 

* adds config files for `nvm` and `pyenv`
* adds instructions for setting both of those up, and how to install xcode build tools via command line

**Note:** I think I'd like to soon split out the different installation instructions for node, python, etc into per OS docs for the first step since its so different for each platform.


docs/contributing/mac-deps-setup.md
```diff
@@ -0,0 +1,40 @@
...
+- Install pyenv according to https://github.com/pyenv/pyenv-installer#github-way-recommended
+- Install the version of python for GitHub Desktop with `pyenv install`
+- You're good to go!
+
+### Manual
+
+- Install manually from https://www.python.org/downloads/
+- You're good to go!
+
+## XCode Command Line Tools
```
j-f1: s/XCode/Xcode/?

docs/contributing/mac-deps-setup.md
```diff
@@ -0,0 +1,40 @@
...
+## Node
+
+There are two approaches to installing Node, **simple** and **version managed**. We recommend the version managed route, but the simple method will work fine if you don't have node installed on your machine and just want to get running!
+
+### Version Managed
+
+We recommend using [nvm](https://github.com/creationix/nvm) to manage your node version. This makes it easier to know you are running the correct version of node for this project without effecting setup for other projects on your machine.
+
+- Install nvm from https://github.com/creationix/nvm#install-script
+- Install the version of node for GitHub Desktop with `nvm install`
```
shiftkey: Do we need to indicate that these commands should be run within the clone of the repository?

iAmWillShepherd:
Just popping in to drop some slack chats (or a summary 🤔 )

I've had some issues using NVM where it constantly resets itself to the newest version of node. It's very annoying because I'd kick off a build, only to see it fail. I haven't discovered a fix (I haven't been searching honestly), but I have adjusted my flow so that I always set the version of node I want to use to build.

Just for some context, I only use NVM because I needed to use my laptop for teaching class in addition to building Desktop. If we do recommend NVM, I'd prefer it to be a weak suggestion.

iAmWillShepherd:
> do you have a node version manager you prefer to nvm?

@outofambit, no, I do not. I only used nvm before because I use my personal computer to work on Desktop and to teach. Since I no longer teach, I don't have any feels (weak or strong) for what folks use to manage their versions. Honestly, I'd prefer we not be prescriptive for managing node versions.

.nvmrc
```diff
@@ -0,0 +1 @@
+v8.11.4
```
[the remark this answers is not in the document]
iAmWillShepherd: This can be set to version `v8` and nvm will just use the latest point release automagically..

outofambit:
⛷

docs/contributing/setup.md
```diff
@@ -6,13 +6,15 @@ You will need to install these tools on your machine:
 
 ### macOS
 
- - [Node.js v8.11.2](https://nodejs.org/dist/v8.11.2/)
+ - [Node.js v8.11.4](https://nodejs.org/dist/v8.11.4/)
  - [Python 2.7](https://www.python.org/downloads/mac-osx/)
  - Xcode and Xcode Command Line Tools (Xcode -> Preferences -> Downloads)
 
+See [mac-deps-setup.md](./mac-deps-setup.md) for additional help with this step.
```
shiftkey: I'm now trying to reconcile how we've duplicated the list of dependencies in two places now - is it worth just rewriting this new document as the macOS setup guide, with a summary at the top of what's needed, and then more details to walk people through the setup, outlining options?

The Windows section below is also a good candidate for it's own document rather than the complex bullet list, but that's out of scope for this PR.

shiftkey: The benefit of this approach is that we can be more detailed with the setup guide, rather than trying to cram it in, and this document becomes just a list of the platforms for developing on, linking off to the specialized docs.

outofambit: yeah that's the direction i want to head. was thinking i would separate out later more cleanly, but there's no reason not to do a little clean up now

outofambit: 0001615

docs/contributing/mac-deps-setup.md
```diff
@@ -0,0 +1,41 @@
...
+
+You will need to install these tools on your machine:
+
+ - Node.js v8.11.4
+ - Python 2.7
+ - Xcode and Xcode Command Line Tools (Xcode -> Preferences -> Downloads)
+
+## Node.js
+
+There are two approaches to installing Node.js, **manual** and **managed**. We recommend the managed route, but the simple method will work fine if you don't have node installed on your machine and just want to get running!
```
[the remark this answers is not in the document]
shiftkey: > because people at WSC code day encountered the most problems with setup when they already had nodejs and/or python installed on their machine

Can you elaborate on what these problems looked like, for context? It's been a while since I've stumbled upon a problem like this with random Node codebases.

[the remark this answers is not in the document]
outofambit: mostly it was that they had a newer version of nodejs installed on their machine (and didn't want to uninstall it). the brew instructions may have helped some of them, but many wanted to keep their current version of node installed and work on our project too. i think part of it too was that `yarn install` failed for newer versions of node, but i can't remember for sure.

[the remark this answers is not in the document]
shiftkey: > mostly it was that they had a newer version of nodejs installed on their machine (and didn't want to uninstall it).

I've had my eye on getting us up to Node 10 locally which will hopefully make this a non-issue. Last I checked it was just `node-sass` that started doing weird things when we upgraded it. If we can land that and no longer rely on Node 8 exclusively, would that change the sort of docs we need to write here?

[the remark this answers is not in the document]
shiftkey: > If we can land that and no longer rely on Node 8 exclusively, would that change the sort of docs we need to write here?

https://github.com/desktop/desktop/pull/5644

shiftkey:
@outofambit @iAmWillShepherd can we hold off on reviewing this until we figure out #5644? I think if we can land that the goals for these docs become much different...

outofambit:
@shiftkey @iAmWillShepherd i've been thinking about our discussion here and while #5644 would help, i think its a workaround for the core issue here: open source contributors often work on an array of projects and need to be able to manage an array of environments. i think this must be addressed regardless of #5644.

newer open source contributors also have dev environment management challenges, but are less likely to know how to address them. in the interest of lowering the barrier to entry, these docs should address the needs of beginners/newcomers first. more experienced contributors (or rather contributors with stronger opinions about their dev environment) can (and probably will) ignore these instructions. and we can explicitly say that they are only here for guidance should you need it. 

to be extra clear here, i don't use `nvm` and i probably won't on this project, but i do think its the easiest, most approachable way to manage node versions currently. and while you can find docs on it, knowing the phrase "version management" as a thing is a barrier.

i'm going to restore the more detailed instructions in this PR and clarify that they are completely optional. @shiftkey  @iamwillshepherd can we reevaluate this PR with beginner/newcomer contributors as the intended audience?

docs/contributing/mac-deps-setup.md
```diff
@@ -0,0 +1,50 @@
...
+
+You will need to install these tools on your machine:
+
+ - Node.js v8.11.4
+ - Python 2.7
+ - Xcode and Xcode Command Line Tools (Xcode -> Preferences -> Downloads)
+
+## Node.js
+
+There are two approaches to installing Node.js, **manual** and **managed**. If you're not sure which route to go, we recommend the managed route, but do whatever makes the most sense for your development process!
```
shiftkey: >  If you're not sure which route to go, we recommend the managed route, but do whatever makes the most sense for your development process!

This sentence still doesn't feel great to me, and I'll try to elaborate on how it relates to these docs.

I'd like these guides (when we give our Windows and Linux setups the same treatment) to be more of a walkthrough than a bullet list of options, to ensure we can support as many users coming to the project as possible. While I'm still :thumbsup: on ensuring we support Node version managers, I'd prefer that we focus these guides on getting someone setup to work on Desktop quickly and clearly.

I originally had this set of instructions at the end of the setup, as a way to sanity check what you had installed.

```shellsession
$ node -v
v8.11.4

$ yarn -v
1.9.4
 
$ python --version
Python 2.7.13
```

What if we made this page more of walkthrough to address these two goals:

 - verifying you have the right tools 
 - if you don't, explain where to find what you need

This doc does go a lot of the way to achieving that, but I think we can take it further and support "users without Node" as well as "users with an existing Node install" and "users who prefer node version managers" while making it clear exactly what they need.

Here's an example of how I'd structure the Node section to flow in this way. Let me know your thoughts.

-----

## Node.js

Let's see if you have the right version of `node` installed. Open a terminal and run this command inside the Desktop source directory:

```
$ node -v
```

If you see an error about being unable to find `node`, that probably means you don't have any Node tools installed. You can install Node LTS (the version we need) from the [Node.js website](https://nodejs.org/en/download/) and restart your shell.

If you see the output `v8.11.x` (where `x` is any number), you're good to go.

If you see the output `v10.x.y` you're ahead of what we currently support. We have an outstanding issue building GitHub Desktop with Node 10, and hopefully can resolve this soon. If you don't care about the version you are running, you can install the version from the [Node.js website](https://nodejs.org/en/download/) over the top of your current install.

### I need to use different versions of Node.js in different projects!

We currently support these Node version managers: `nvm`,  `n`, `asdf-nodejs`. 

#### Configuring `nvm`

1. Install `nvm` using the instructions [here](https://github.com/creationix/nvm#install-script).

2. Within the Desktop source directory, install version of Node.js it requires:

```shellsession
$ nvm install
```

3. Ensure you are running the right version

```
$ nvm use
```

4. Verify you have the right version by running `node -v` again

```
$ node -v
```

If you see `v8.11.4`, you're good to go.

#### Configuring `n`

...

#### Configuring `asdf-nodejs`

...

outofambit:
🍕@shiftkey @iAmWillShepherd

.nvmrc
streeter: I know this PR was closed, but looks like [`.node-version`](https://github.com/desktop/desktop/blob/b7895b606c25b29cb735e3affce2ffd38a4870bd/.node-version) was missed in the update. Though, I'm not sure what the difference between that file, `.nvmrc`, and `.tool-versions` is for...

outofambit: thanks for the catch @streeter! i'll get that in another pr in just a minute

outofambit: https://github.com/desktop/desktop/pull/5705

PR #5550, opened 2018-09-05
add analytics for how user leverages merge hint to merge

### LINKED ISSUE #5394: [Metrics] Merge conflict resolution
With the general goal of reducing friction for people collaborating on software, we want to instrument the app to understand the following:

- [ ] Given that an attempted merge results in a merge conflict, what percentage of those eventually result in a successful merge?
- [ ] Of all attempted merges, what percentage result in a successful merge?

I think we have attempted merges that result in conflicts, so the missing piece here is to be able to understand whether those conflicts eventually get resolved successfully or whether they're just abandoned. A thing we've heard in user research is people saying some version of, "and then I just had to start over because I couldn't resolve the conflict." We obviously want to avoid this, so understanding how common it is when people do get into conflicts will be helpful in us figuring out whether we're making progress here.

The ideal scenario is that anytime a person attempts to merge code from one branch to another, that they're able to do so successfully.

Tasks
- [x] Create issue in analytics repo: https://github.com/github/analytics/issues/712
- [x] Add metrics to app: #5810 
- [x] Update metrics documentation #5879 
- [x] Update marketing site - https://github.com/github/desktop.github.com/pull/119
- [x] Update central - https://github.com/github/central/pull/411
- [ ] 🚢 

cc: @telliott27

shiftkey:
I'm not sure whether this entirely resolves (it doesn't completely) #5394 but I think this is two key metrics which will help us understand the impact of the new merge hint.

This PR checks **at merge time** what hint was shown to the user:

 - `proceededAfterNoConflictsHintCount` - the number of times a user merged with the "no conflicts" hint shown
 - `proceededAfterConflictsWarningCount` - the number of times a user merged with the "you have X conflicted files" metric

We already have some insight into how the user is initiating the merge flow (we aggregate these up into "total merges" which isn't quite accurate) but these two metrics show how many actually complete the flow and perform the merge to update their repository.

We don't report on the number of conflicted files shown by the dialog because we don't have a good story about sending those sorts of metrics, but we have that data available on `mergeStatus`.

This also requires changes to Central and `desktop.github.com`, but I'm opening this up for discussion to see whether it's on the right track.

- [x] Update Central: https://github.com/github/central/pull/397
- [x] Update marketing site: https://github.com/github/desktop.github.com/pull/115
- [x] Update metrics doc: https://github.com/desktop/desktop/pull/5556

cc @billygriffin @telliott27

### DIFF
--- a/app/src/lib/dispatcher/dispatcher.ts
+++ b/app/src/lib/dispatcher/dispatcher.ts
@@ -19,6 +19,7 @@ import {
   ImageDiffType,
   CompareAction,
   ICompareFormUpdate,
+  MergeResultStatus,
 } from '../app-state'
 import { AppStore } from '../stores/app-store'
 import { CloningRepository } from '../../models/cloning-repository'
@@ -569,8 +570,12 @@ export class Dispatcher {
   }
 
   /** Merge the named branch into the current branch. */
-  public mergeBranch(repository: Repository, branch: string): Promise<void> {
-    return this.appStore._mergeBranch(repository, branch)
+  public mergeBranch(
+    repository: Repository,
+    branch: string,
+    mergeStatus: MergeResultStatus | null
+  ): Promise<void> {
+    return this.appStore._mergeBranch(repository, branch, mergeStatus)
   }
 
   /** Record the given launch stats. */
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -91,6 +91,12 @@ export interface IDailyMeasures {
   /** The number of times the users pushes to a generic remote */
   readonly externalPushCount: number
 
+  /** The number of times the user has merged after seeing the conflict warning */
+  readonly userProceededAfterConflictsWarning: number
+
+  /** The number of times the user has merged after seeing the 'no conflicts' warning */
+  readonly userProceededAfterNoConflictsHint: number
+
   /** Whether or not the app has been interacted with during the current reporting window */
   readonly active: boolean
 
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -91,11 +91,14 @@ export interface IDailyMeasures {
   /** The number of times the users pushes to a generic remote */
   readonly externalPushCount: number
 
-  /** The number of times the user has merged after seeing the conflict warning */
-  readonly proceededAfterConflictsWarningCount: number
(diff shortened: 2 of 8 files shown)


billygriffin:
@shiftkey This is a great start! A couple of thoughts/questions:

1. What happens if a user clicks "merge" while the check for conflicts is pending? It seems like that wouldn't be captured by either of these metrics.
2. @telliott27 and I talked about this, and we think the only thing missing that would resolve #5394 is the eventual disposition of the merges that result in a conflict. In other words, how many of them are resolved "successfully" vs. abandoned? We were thinking a way to determine "successfully" might be if there was a commit in Desktop on that branch after a conflict was hit, but not exactly sure how to instrument that.

shiftkey:
> What happens if a user clicks "merge" while the check for conflicts is pending? It seems like that wouldn't be captured by either of these metrics.

Correct. I think those flows fall into one of these two buckets:

 - user went to merge immediately, without caring about the hint
 - user waited for the merge hint and got impatient waiting

> we think the only thing missing that would resolve #5394 is the eventual disposition of the merges that result in a conflict. In other words, how many of them are resolved "successfully" vs. abandoned? We were thinking a way to determine "successfully" might be if there was a commit in Desktop on that branch after a conflict was hit, but not exactly sure how to instrument that.

Correct - there might be ways to identify these but I believe it's independent of this work (both the merge hint and this PR). We'd need to inspect the `.git` directory before committing to look for the presence of `.git/MERGE_HEAD` and  `.git/MERGE_MSG` to indicate the current state of the repository.

shiftkey:
@billygriffin 👍  I'll add it in

shiftkey:
I've settled on these three names:

 - `mergedWithLoadingHintCount`
 - `mergedWithCleanMergeHintCount`
 - `mergedWithConflictWarningHintCount`

If we're good to go with those, I'll open up tasks to track the other infrastructure changes we need.

billygriffin:
All of those work for me @shiftkey! Thanks!

iAmWillShepherd (approved):

iAmWillShepherd:
@billygriffin are you able to provide any justification for these new metrics over on https://github.com/desktop/desktop/pull/5556/ ?

cc @desktop/analytics

billygriffin:
@iAmWillShepherd Done, thanks!

iAmWillShepherd (approved):

outofambit (approved):
this all sounds and looks good to me!

PR #5552, opened 2018-09-05
release notes dialog spacing tweaks

outofambit:
did a little design tweaking based on talking with @donokuda. mostly just paddings/margins.

two columns

one column

### DIFF
--- a/app/src/ui/release-notes/release-notes-dialog.tsx
+++ b/app/src/ui/release-notes/release-notes-dialog.tsx
@@ -120,9 +120,11 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
   private drawSingleColumnLayout(release: ReleaseSummary): JSX.Element {
     return (
       <div className="container">
-        {this.renderList(release.bugfixes, 'Bugfixes')}
-        {this.renderList(release.enhancements, 'Enhancements')}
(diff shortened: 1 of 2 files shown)


donokuda:
Looking at the second screenshot, I feel like there's a lot of horizontal spacing around the column.

I wonder if a single column should have a larger width, or we could consider reducing the width of the dialog itself.

outofambit:
screen shot 2018-09-06 at 11 00 30 am

iAmWillShepherd:
I'm in agreement with @donokuda on this one. I think we should reduce the width of the dialog or reduce the padding around the content a bit.

outofambit:
@iAmWillShepherd thanks! are you referring to the initial screenshots? or this one: https://github.com/desktop/desktop/pull/5552#issuecomment-419188991 ?

iAmWillShepherd:
Oops, yea, I was referring to the latest one, I still think we can either use more of that white-space, or remove it. But, since we will be iterating on the design of in-app release notes, it may not be worth the time to keep tweaking what you have.

I'm going to go ahead and merge this 👍

outofambit:
@donokuda 🍕

outofambit:
as of 096de75:
screen shot 2018-09-06 at 4 17 04 pm

PR #5554, opened 2018-09-06
MergeCallToActionWithConflicts display logic refactor

outofambit:
addresses my review comments on #5495:
* refactors display logic in `MergeCallToActionWithConflicts` to be more compact
* a little string formatting

### DIFF
--- a/app/src/ui/history/merge-call-to-action-with-conflicts.tsx
+++ b/app/src/ui/history/merge-call-to-action-with-conflicts.tsx
@@ -79,13 +79,9 @@ export class MergeCallToActionWithConflicts extends React.Component<
           <div className="merge-message">
             This will merge
             <strong>{` ${count} ${pluralized}`}</strong>
-            {` `}
-            from
-            {` `}
+            {` from `}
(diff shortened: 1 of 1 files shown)


app/src/ui/history/merge-call-to-action-with-conflicts.tsx
```diff
@@ -62,59 +62,77 @@ export class MergeCallToActionWithConflicts extends React.Component<
     behindCount: number
   ) {
     if (mergeStatus === null || mergeStatus.kind === MergeResultKind.Loading) {
-      return (
-        <div className="merge-message merge-message-loading">
-          Checking for ability to merge automatically...
-        </div>
+      return this.renderLoadingMergeMessage()
+    } else if (mergeStatus.kind === MergeResultKind.Clean) {
```
iAmWillShepherd: No need for `else` statement here since you return in each block. Normal `if` statement is fine, or use a `switch` statement.

outofambit: ah good point :)

PR #5563, opened 2018-09-07
ensure the old layout is fine for users until they see the new merge hint

shiftkey:
This PR ensures the existing message (which will be default until we enable the feature flag) still looks fine.

To test this out, checkout `display-merge-status-in-compare-tab` and disable the feature flag:

```diff
diff --git a/app/src/lib/feature-flag.ts b/app/src/lib/feature-flag.ts
index 1572bf5e0..7b46277b5 100644
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -44,7 +44,7 @@ export function enableRepoInfoIndicators(): boolean {
 
 /** Should the app try and detect conflicts before the user stumbles into them? */
 export function enableMergeConflictDetection(): boolean {
-  return enableDevelopmentFeatures()
+  return false
 }
 
 /** Should `git status` use --no-optional-locks to assist with concurrent usage */
```

Run it up locally and you'll see it looks more squashed compared to what we've currently shipped:

screen shot 2018-09-07 at 10 00 45 am

This PR adds back that top margin, and indicates it can be cleaned up when we're happy with this feature:

screen shot 2018-09-07 at 10 04 16 am

cc @desktop/design for :eyes: and ideas on how to manage styles that are tied to feature flags like this.

### DIFF
--- a/app/src/ui/history/merge-call-to-action.tsx
+++ b/app/src/ui/history/merge-call-to-action.tsx
@@ -50,7 +50,7 @@ export class MergeCallToAction extends React.Component<
     if (count > 0) {
       const pluralized = count === 1 ? 'commit' : 'commits'
       return (
-        <div className="merge-message">
+        <div className="merge-message merge-message-legacy">
           This will merge
           <strong>{` ${count} ${pluralized}`}</strong>
           {` `}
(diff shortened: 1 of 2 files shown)


outofambit:
i would love to move some (or more) of our styles to use https://www.styled-components.com/ at some point as that would make conditional styling much simpler

PR #5564, opened 2018-09-07
extract some low-hanging responsibilities from AppStore

shiftkey:
This PR attacks some technical debt inside `AppStore` by moving out one of it's responsibilities in a way that's easier to test. It's only ~240 lines net loss, but I'll summarize what I've done here:

 - `AppStore` manages an internal `Map<string, IRepositoryState>` lookup so when you're switching between tracked repositories the app doesn't need to immediately read from disk. This functionality now lives in `RepositoryStateCache`, which makes it easier to test in isolation. 
 - all the `constructor` parameters of `AppStore` are now `private readonly`, and components which were accessing these parameters directly (yay coupling!) now have their own constructor dependencies
 - `Dispatcher` methods for updating stats which were just invoking the same `AppStore` methods and producing side-effects without update app state have been removed in favour of the dispatcher just invoking `StatsStore` directly.
 - a couple of enum types defined in `AppStore` have been moved under `app/src/models/`


app/src/lib/stores/app-store.ts
```diff
@@ -1514,15 +1369,11 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return defaultPRLabel
     }
 
-    const repositoryState = this.repositoryState.get(repository.hash)
-
-    if (!repositoryState) {
```
shiftkey: Extracting out the repository state cache showed this was an inconsistent way of poking at the repository state cache, and it would bypass the behaviour in `this.getRepositoryState()` that creates the initial state if it didn't already exist.

shiftkey:
And to ensure our code coverage is going in the right direction, I added tests for this new component now that it lives outside the `AppStore`:

app/src/lib/stores/repository-state-cache.ts
```diff
@@ -0,0 +1,142 @@
...
+} from '../app-state'
+import { ComparisonCache } from '../comparison-cache'
+import { IGitHubUser } from '../databases'
+import { merge } from '../merge'
+import { GitHubUserStore } from './github-user-store'
+
+export class RepositoryStateCache {
+  private repositoryState = new Map<string, IRepositoryState>()
+
+  public constructor(private gitHubUserStore: GitHubUserStore) {}
```
iAmWillShepherd: Could we pass in a function that returns a `GitHubUserStore` instead of keeping this as a dependency?

app/test/unit/app-store-test.ts
```diff
@@ -65,16 +66,21 @@ describe('AppStore', () => {
       repositoriesStore
     )
 
+    const githubUserStore = new GitHubUserStore(db)
+
+    const repositoryStateManager = new RepositoryStateCache(githubUserStore)
+
     return new AppStore(
-      new GitHubUserStore(db),
+      githubUserStore,
```
iAmWillShepherd: Could we do this inline `new GitHubUserStore(db)`

app/src/lib/stores/repository-state-cache.ts
iAmWillShepherd: Also, I only spotted one usage of this store on line 34, maybe we can just pass a callback that make the call to `getUsersForRepository`, that should make testing a bit easier.

shiftkey: Maybe? I hadn't really thought about making that dependency more abstract. I'll have a play around and see how it feels.

shiftkey: Try out https://github.com/desktop/desktop/pull/5564/commits/22c2428c7393e50dc9a4dd491abca14f5b26168b and see how it feels

shiftkey:
Gonna move this into 1.4.1 unless there are significant reservations about it

PR #5576, opened 2018-09-09
Never suffix `.git` to URLs cloned via an opened URL

### LINKED ISSUE #5529: "github clone url" when URL ends in .git adds another .git and causes auth error
## Description
When trying to clone a repository thru using the command `github clone...`, the command opens the Desktop Application, which prompts for your GitHub information. When trying to login with the correct information, the application fails. It's also worth noting, in my case, I was already authenticated in the Application...and was able to clone the repository without any issue using the application's interface.

## Version

* GitHub Desktop: 1.3.6
* Operating system: 10.13.6

## Steps to Reproduce
Installed thru `brew cask install github`
Open Application and sign into Github
Run "Install Command Line Tool..."
Try to clone any repository

### Expected Behavior
For the repository to be cloned, or for the login to authenticate properly.

### Actual Behavior
Repository fails to clone because login fails to authenticate.

### Logs
[2018-09-01.desktop.production.log](https://pastebin.com/QM3kJykT)

j-f1:
If it’s already present, this breaks the URL. Repositories can still be cloned from GitHub without the `.git` at the end, and other hosts might not allow it.

Fixes #5529.


outofambit:
thanks @j-f1! it looks like there are still linting errors right now. but anyways:

wouldn't a more conservative solution be to add `.git` when the URL doesn't already end in it? or does that not work in some cases?

j-f1:
I don’t see why we’d need that. I don’t know how every Git server works, but I wouldn’t be surprised if there was one that didn’t allow the `.git` ending. If a server requires it, people will most likely either manually type it in by hand or copy-paste the URL from a web UI.

outofambit:
> I don’t know how every Git server works, but I wouldn’t be surprised if there was one that didn’t allow the .git ending.

i totally agree but status quo is those don't work at all for github desktop as we always append the `.git` i'd prefer addressing that as a separate issue when we come across it, though i'm curious on what others in @desktop/maintainers think?

i'm also being cautious because of https://github.com/desktop/desktop/issues/5529#issuecomment-418064693

app/src/lib/parse-app-url.ts
```diff
@@ -102,10 +102,7 @@ export function parseAppURL(url: string): URLActionType {
   const parsedPath = pathName.substr(1)
 
   if (actionName === 'openrepo') {
-    const probablyAURL = parsedPath
-
-    // suffix the remote URL with `.git`, for backwards compatibility
```
iAmWillShepherd: I'd prefer not to remove this line given the comment states this is for backwards compatibility.

iAmWillShepherd (changes requested):

iAmWillShepherd:
> Repositories can still be cloned from GitHub without the .git at the end, and other hosts might not allow it.

I'd prefer if we only append `.git` at the end of a url if it's not already there. Although GitHub dotcom may not require that repos end with `.git` our GHE users' instances may require it. 

> f a server requires it, people will most likely either manually type it in by hand or copy-paste the URL from a web UI.

Let's not make assumptions here, if we can remove friction for our users here, why not add this extension automatically, like we do already. I'm in agreement with using @outofambit's [recommendation](https://github.com/desktop/desktop/pull/5576#issuecomment-420007776) here.

say25:
Maybe this is a better "compromise", could we:
1. Append `.git` if its not already there.
2. If the cloning fails with `.git` appended fails retry with out `.git`

niik (dismissed):
> i'm also being cautious because of #5529 (comment)
https://github.com/desktop/desktop/issues/5529#issuecomment-418064693

I should clarify that I meant cautious in terms of the app itself. My concerns were not related to the server side here. GitHub.com, and GitHub Enterprise support cloning with, and without the .git suffix for http remotes and I agree with @j-f1's solution to always accept what we get from users. There's not, to my knowledge, any universally accepted standard for Git URLs prescribing that they end with `.git`.

I was concerned that we'd have some ancient regular expression lying around that relied on the .git extension being there is all and I didn't have time to dig into it any further right then and there.

> i totally agree but status quo is those don't work at all for github desktop as we always append the .git 

We only append the suffix for URLs opened through the 'openrepo' action, i.e. links such as this

screen shot 2018-09-11 at 07 18 02

The `github` CLI command piggybacks on the same protocol to pass the given command line argument url to Desktop.

When users manually type a URL into the clone dialog we don't unconditionally add a `.git` suffix but rather attempt to [look up the canonical clone URL](https://github.com/desktop/desktop/blob/cafa5347583e0d1a56999a166fca27bb745f0494/app/src/ui/clone-repository/clone-repository.tsx#L319-L345) using the API (if it's a GitHub repository and the user is signed in).

> i'd prefer addressing that as a separate issue when we come across it, though i'm curious on what others in @desktop/maintainers think?

I think @j-f1's take on this is correct, the current behaviour of us manipulating all clone URLs is unintuitive. While the clone in desktop links are supposed to be GitHub specific the fact that the CLI piggybacks on that protocol means that we should take what we get at face value (and the clone URLs in the Clone In Desktop links on GitHub.com are always canonical anyway).

> Although GitHub dotcom may not require that repos end with .git our GHE users' instances may require it.

GHE works the exact same way as GitHub.com in this regard and my previous paragraph about resolving canonical clone URLs through the API also applies here.

app/src/lib/parse-app-url.ts
niik: That line was added by @shiftkey during a refactor and the backwards compatibility is as far as I can tell referring to Desktop itself and not some ancient GHE and or GitHub.com version. I think @shiftkey added it more as a way of saying "I don't know if this is really necessary but it's been like this since forever so I'm not gonna deal with it in this refactor". Please correct me if I'm wrong @shiftkey.

In order words I think the comment itself is confusing as it could be interpreted as backwards compatibility with GHE/GitHub.com.

outofambit:
ok! i'm on board with the solution proposed here by @j-f1 :) thanks @niik

app/src/lib/parse-app-url.ts
shiftkey: I'm really hazy on the details but this probably came up because the URLs that were generated on GitHub looked like this:

```
x-github-client://openRepo/https://github.com/desktop/desktop
```

While we can get away with omitting these for GitHub, perhaps I was thinking about non-GitHub remotes that might also support these protocol handlers, but require a `.git` suffix?

Also note that this work pre-dates implementing the CLI that leverages this too, so I wasn't thinking about users being able to specify URLs with or without `.git`.

shiftkey:
>  Although GitHub dotcom may not require that repos end with `.git` our GHE users' instances may require it.

@iAmWillShepherd I believe this is safe change for GHE installations as this behaviour (being able to clone with or without the `.git` suffix) has been baked for enough updates that it should cover the versions that Desktop supports.

shiftkey:
Just pushed the additional tests from #5597 into this PR to ensure this PR doesn't regress SSH URLS

j-f1:
Do the `git@github.com` URLs need the `.git` suffix?

shiftkey:
@j-f1 no

```
$ git clone git@github.com:desktop/dugite
Cloning into 'dugite'...
Warning: Permanently added the RSA host key for IP address '192.30.253.112' to the list of known hosts.
remote: Counting objects: 2879, done.
remote: Compressing objects: 100% (44/44), done.
remote: Total 2879 (delta 24), reused 23 (delta 10), pack-reused 2825
Receiving objects: 100% (2879/2879), 469.83 KiB | 1.43 MiB/s, done.
Resolving deltas: 100% (1690/1690), done.
```

j-f1:
:gem:

app/src/lib/parse-app-url.ts
```diff
@@ -104,10 +104,7 @@ export function parseAppURL(url: string): URLActionType {
...
-    const url = `${probablyAURL}.git`
+    const url = parsedPath
```
shiftkey: :art: can we move this down to line 128:

```ts
     return {
       name: 'open-repository-from-url',
       url: parsedPath,
       branch,
       pr,
       filepath,
     }
```

shiftkey (changes requested):
Just one bit of cleanup, everything else is 🎉

shiftkey (approved):

iAmWillShepherd (approved):

Banner-Keith:
> I don’t see why we’d need that. I don’t know how every Git server works, but I wouldn’t be surprised if there was one that didn’t allow the `.git` ending. If a server requires it, people will most likely either manually type it in by hand or copy-paste the URL from a web UI.

I know this has already been merged and taken care of but I wanted to note that the TFS/VSTS git implementation did not accept `.git` on the end of the url. Our team uses github desktop a lot for our TFS git repositories and we greatly appreciate this fix. Please don't ever go back to adding `.git` if it is not on the url already. :) Just thought I'd add this note as you were correct in suspecting that some servers don't allow `.git`.

Thank you all for this awesome fix.

billygriffin:
That's great to hear @Banner-Keith and helpful for us to better understand the landscape.

PR #5577, opened 2018-09-09
Remove 1.3 from Roadmap

say25:
To be done when 1.4 is ready.


shiftkey:
@say25 thanks for keeping this up to date. Hopefully we'll have some more to share about 1.5 and other upcoming work in the coming weeks!

say25:
@shiftkey should we kill the 1.4 notes or wait until 1.5 ships?

PR #5581, opened 2018-09-10
don't show merge hint when branch can be fast-forwarded

shiftkey:
This fixes a corner case where you are comparing to a branch that can be fast-forwarded (so no commits behind), and there's no point displaying any new UI (because we disable the merge button).

Current behaviour (1.3.5):

Current behaviour (1.4.0-beta2):

New behaviour:


tierninho:
Curious if this merge hint should also be removed for the same reasons, or are we keeping it around for affirmation? Seems like it should be removed or the language changed...

screen shot 2018-09-21 at 2 13 24 pm

shiftkey:
@tierninho we should be hiding the prompt when there are no commits to merge - I think I might have overlooked the merge dialog when verifying this PR.

Feel free to open a new issue with steps if you can reproduce this on the latest build and I can confirm that's the case.

Issue #5582, opened 2018-09-10
Invalid state prevents merge messaging from loading


tierninho:
Merge Conflict testing surfaced the following error:

`Uncaught (in promise) Error: invalid state - trying to append the diff to a merge entry that isn't defined. line: ''
    at Object.t.parseMergeResult (merge-tree-parser.ts:161)
    at Object.t.mergeTree (merge.ts:79)
    at <anonymous>`

Steps to reproduce... a bit murky as it was hard to reproduce:
- set up and publish new branch
- get the NDDB prompt to surface by comparing with default branch
- somehow i managed to get the same `gitignore` file to surface in both the Behind and Ahead tabs, likely by making a change to it
- check the merge-messaging and it should be stuck in "checking" state

<img width="1162" alt="screen shot 2018-09-10 at 8 22 12 am" src="https://user-images.githubusercontent.com/14828183/45307937-2a914f80-b4d5-11e8-8dd4-f7b76046e6a5.png">

shiftkey:
@tierninho are you able to grant me access to that repository?

shiftkey:
@tierninho I don't seem to be able to access the `hello1` branch from your screenshot. It seems to have a new commit that's key to triggering that issue.

tierninho:
@shiftkey just pushed the commit up, let me know if that works

shiftkey:
@tierninho thanks, I can see the same error now!

shiftkey:
This will be fixed by 940c9e1709e9669dfa03a3e4aa992f29cf30153a - thanks for the help with reproducing it!

PR #5583, opened 2018-09-10
avoid the merge-tree operation work in some cases

shiftkey:
This relates to #5581 and investigating #5582 but tweaks some of the merge hint flow:

 - we skip performing the `mergeTree` action if the merge operation is a fast-forward only. In this case we disable the merge button anyway, so there's no point crunching any further data
 - we handle any unexpected errors from `mergeTree` and clear the `mergeStatus` state
 - when `mergeStatus` is null, we don't show any UI - this should help with some of the confusion we've heard from early beta testers

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -987,7 +987,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         return this.currentMergeTreePromise
       }
 
-      if (tip.kind === TipState.Valid) {
+      if (tip.kind === TipState.Valid && aheadBehind.behind > 0) {
         const mergeTreePromise = promiseWithMinimumTimeout(
           () => mergeTree(repository, tip.branch, action.branch),
           500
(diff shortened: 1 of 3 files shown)


shiftkey:
This is ready for review :tada:

shiftkey:
@outofambit two cases come to mind:

 - the initial state _before_ the user has chosen a branch to compare
 - some unknown state _after_ tried to merge, like #5582 - we shouldn't show any UI if we're not sure what the result is, and we'll log out some details so we can help users who notice this to dig into and improve things

app/src/lib/stores/app-store.ts
```diff
@@ -987,17 +987,27 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          .catch(err => {
+            log.warn(
+              `Error occurred while trying to merge ${tip.branch.name} (${
+                tip.branch.tip.sha
+              }) and ${action.branch.name} (${action.branch.tip.sha})`,
+              err
+            )
+            return null
+          })
+          .then(mergeStatus => {
```
iAmWillShepherd: Curious why this appears after the catch? Not that it makes much difference, but the typical ordering of these method calls is `then`, `catch`, then `finally`.

shiftkey: The return value from `catch` is also a Promise, so this order of chaining means:

 - first, we run the async code that might error
 - then, if it errors, the `catch` callback allows us to handle the error and return something
 - finally, the `then` is a single callback to update the UI for both success and error cases

Flipping it around means that we might have to have UI code duplicated in `then` and `catch`

PR #3601, opened 2017-12-13
in-app release notes

shiftkey:
Fixes #2774 

### Data
- [x] compute and group release notes from API
- [x] only compute for the latest release, ignore in-between releases
- [x] tests

### Component
- [x] layout and basic interactions
- [x] single layout mode when we have simple summary
- [x] two-section mode when we have complex summary
- [x] SVG background for header
- [x] render hyperlinks in each row
- [x] render mention in each row
- [x] render pretext (if found)
- [x] ermagherd all the CSS

### Cleanup
- [x] compute release notes when an update is available
- [x] flow release notes from update store into component, make component stateless
- [x] cleanup loading jank - not needed
- [x] can we extract `DialogHeader` and reuse that instead of duplicating work here?
- [x] refresh wording for prompt and dialog buttons
- [x] open task to add support rendering "Pretext" header #5510
- [x] cleanup mentions of `HACK` which are purely for testing/iterating on the UI

PR #5592, opened 2018-09-10
1.4.0-beta3 release notes draft

outofambit:
- [x] finalize release notes
- [x] release to beta channel

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.4.0-beta2",
+  "version": "1.4.0-beta3",
   "main": "./main.js",
   "repository": {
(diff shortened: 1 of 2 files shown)


outofambit:
@shiftkey good to know!

outofambit:
i think i'm gonna replace these with the same release notes as the prior betas for clarity

say25:
Not sure if this is a bug or not, but in all the 1.4.0 betas I havent been seeing release notes after clicking either of the `Quit and Restart` buttons

shiftkey:
@say25 we've only changed the behaviour of this link - previously it just took you to the release notes URL on `desktop.github.com`, but now it will show the dialog in #3601:



I know there are other ideas floating around (for example, I hope @nerdneha opens an issue to track [this comment](https://github.com/desktop/desktop/pull/3601#issuecomment-418841199), so if you've got suggestions about how we can refine this then let us know!

PR #5595, opened 2018-09-11
Warn user if branch exists on remote when creating branch

### LINKED ISSUE #5141: Local branch name conflicts with remote branch name
Ok so i was trying to push files into my repo

and i came across this thing

[Error Click Here](https://cdn.discordapp.com/attachments/425750108223569982/466542651596865556/Untitled.png)

So i tried to pull, but when i tried that

[I saw this](https://cdn.discordapp.com/attachments/425750108223569982/466542648790876161/Untitled2.png) 

> **Can anyone help me? Please?**

> Can't do pull request either, same error for some reason...

**Thanks niik, i made new branch `oof` and made it default, then deleted master branch and now it works.**

Daniel-McCarthy:
Fixes #5141 

It was suggested to warn a user when attempting to create a branch with a name that already exists on remote.

Previously it already could notify and block a user from creating a branch when it existed locally. Now it can also give a to inform the user that the branch already exists on remote while allowing them to still create their local branch.
___
One matter I felt should be discussed is the design:

- The `create-branch-dialog` already allows a warning to be displayed, however it tells things like when an entered name will be sanitized and changed. This should still display and not be overwritten to display a branch exists warning.
- With this pull request, 0-2 warnings can be displayed at a time. The branch exists warning will not effect the above mentioned warnings and will be displayed additionally.

With that said, it may look _odd_ having two separate warnings and symbols. Currently this is how it behaves. Though I can change it to display both warnings, but only one symbol if both warnings are showing at once. But I would see what others think.
___
Here is what it looks like currently with 2 warnings displayed:
image

### DIFF
--- a/app/src/ui/lib/branch-name-warnings.tsx
+++ b/app/src/ui/lib/branch-name-warnings.tsx
@@ -56,16 +56,16 @@ export function renderBranchNameExistsOnRemoteWarning(
       b => b.nameWithoutRemote === sanitizedName && b.type === BranchType.Remote
     ) > -1
 
-  if (alreadyExistsOnRemote) {
-    return (
-      <Row className="warning-helper-text">
-        <Octicon symbol={OcticonSymbol.alert} />
-        <p>
-          A branch named <Ref>{sanitizedName}</Ref> already exists on remote.
-        </p>
-      </Row>
-    )
-  } else {
+  if (alreadyExistsOnRemote === false) {
     return null
   }
+
+  return (
+    <Row className="warning-helper-text">
+      <Octicon symbol={OcticonSymbol.alert} />
+      <p>
+        A branch named <Ref>{sanitizedName}</Ref> already exists on remote.
+      </p>
+    </Row>
+  )
 }
--- a/app/src/ui/lib/branch-name-warnings.tsx
+++ b/app/src/ui/lib/branch-name-warnings.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Branch } from '../../models/branch'
+import { Branch, BranchType } from '../../models/branch'
 
 import { Row } from './row'
 import { Octicon, OcticonSymbol } from '../octicons'
@@ -39,6 +39,29 @@ export function renderBranchHasRemoteWarning(branch: Branch) {
         <p>
           This branch is tracking <Ref>{branch.upstream}</Ref> and renaming this
           branch will not change the branch name on the remote.
+        </p>
+      </Row>
+    )
+  } else {
+    return null
+  }
+}
+
+export function renderBranchNameExistsOnRemoteWarning(
+  sanitizedName: string,
+  branches: ReadonlyArray<Branch>
+) {
+  const alreadyExistsOnRemote =
+    branches.findIndex(
+      b => b.nameWithoutRemote === sanitizedName && b.type === BranchType.Remote
+    ) > -1
+
+  if (alreadyExistsOnRemote) {
+    return (
+      <Row className="warning-helper-text">
+        <Octicon symbol={OcticonSymbol.alert} />
+        <p>
+          A branch named <Ref>{sanitizedName}</Ref> already exists on remote.
         </p>
       </Row>
     )
--- a/app/src/ui/lib/branch-name-warnings.tsx
+++ b/app/src/ui/lib/branch-name-warnings.tsx
@@ -64,7 +64,7 @@ export function renderBranchNameExistsOnRemoteWarning(
     <Row className="warning-helper-text">
       <Octicon symbol={OcticonSymbol.alert} />
       <p>
-        A branch named <Ref>{sanitizedName}</Ref> already exists on remote.
+        A branch named <Ref>{sanitizedName}</Ref> already exists on the remote.
       </p>
     </Row>
   )
--- a/app/src/ui/create-branch/create-branch-dialog.tsx
+++ b/app/src/ui/create-branch/create-branch-dialog.tsx
@@ -40,7 +40,7 @@ enum StartPoint {
 
 interface ICreateBranchState {
   readonly currentError: Error | null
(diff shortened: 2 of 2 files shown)


outofambit:
@desktop/design what do you think about the multiple error display above?

I could imagine that if both errors are present, we show a single error with text like:
"⚠️ Would be created as `test-1` but a branch with that name already exists on the remote."

Daniel-McCarthy:
@outofambit I believe this should be do-able. Implementing it that way should actually remove some redundancy in my changes despite a bit of an added layer of complexity. 

Thanks for the suggestion! I will look into this.

Daniel-McCarthy:
To give an update on the option of combining the warnings:
___
Here are the parts of the code base that are responsible for the current behavior.
- The `Create-Branch-Dialog` [renders it's warnings here.](https://github.com/desktop/desktop/blob/027dfb08b4b3ba4bb27c5d3c8890ef9813eb5851/app/src/ui/create-branch/create-branch-dialog.tsx#L260-L263)
- The `Rename-Branch-Dialog` [renders it's warnings here.](https://github.com/desktop/desktop/blob/1de8d45fc6dc3da6b7099c68f0d4a97c311bf972/app/src/ui/rename-branch/rename-branch-dialog.tsx#L56-L61)
- Both dialogs depend on the [this code to display general name warnings.](https://github.com/desktop/desktop/blob/1de8d45fc6dc3da6b7099c68f0d4a97c311bf972/app/src/ui/lib/branch-name-warnings.tsx#L8-L33)
- The `Rename-Branch-Dialog` also depends[ on this to display it's additional warning](https://github.com/desktop/desktop/blob/1de8d45fc6dc3da6b7099c68f0d4a97c311bf972/app/src/ui/lib/branch-name-warnings.tsx#L34-L47) that is separate from the name warnings it can also display.

So if we were to merge this behavior so that the warnings can be combined, there are two warnings that both dialogs can display, and both dialogs each have an extra warning specific only to that dialog.
___
Currently I don't have a solution for combining these warnings that would be maintainable. 

An ideal situation would be where I can define each unique warning once, then if there is more than one to display, remove the period at the end of one and replace it with " and " and add in the second warning.

- If these warnings were strings, regex would be useful in combining the warnings or displaying an individual one if only one is present. However, as strings they would not be able to display the `<Ref></Ref>` tags they would have as `JSX.Element` objects.
- When working with `JSX.Element` objects instead of strings, the issue is that I do not know of a way where I can create a `JSX.Element` for each warning then modify the contents to combine the text of the two.
- If the only option is to decide the warning on creation of the `JSX.Element` (and I can't use a string and modify it, and I can't edit the HTML content afterwards), I have to have the warning prepared before hand. This would mean I would have to have an object for each of the warnings and each of their combinations.
- If we had to have all of the combinations prepared ahead of time, another complication is that the functionality for both Dialogs are combined. `Create-Branch-Dialog` and `Rename-Branch-Dialog` need to pass different data. One of the warnings require having an existing `Branch` to get the upstream name The other requires a ReadonlyArray<Branch>. If we have to duplicate warnings for separate functions then the maintainability of even small edits like changing the warning text becomes a bit more difficult.
___
Perhaps others with more knowledge of JSX will know a solution for this problem. I feel there is likely a solution for this that I am missing.

j-f1:
How about stringing the messages together, i.e. “Will be created as `test-1`. A branch named `test-1` already exists on the remote.”?

app/src/ui/lib/branch-name-warnings.tsx
```diff
@@ -46,3 +46,26 @@ export function renderBranchHasRemoteWarning(branch: Branch) {
...
+export function renderBranchNameExistsOnRemoteWarning(
+  sanitizedName: string,
+  branches: ReadonlyArray<Branch>
+) {
+  const alreadyExistsOnRemote =
+    branches.findIndex(
+      b => b.nameWithoutRemote === sanitizedName && b.type === BranchType.Remote
+    ) > -1
+
+  if (alreadyExistsOnRemote) {
```
shiftkey: :art: invert this check so you can return early

```tsx
if (!alreadyExistsOnRemote) {
  return null
}

return (
  ...
)
```

Daniel-McCarthy: Sounds good, I will make this change now. 

There is a second function in this file, `renderBranchHasRemoteWarning`, that also follows the if(...) {return warning} else { return null } structure. Would you like me to modify that to conform with this change as well?

billygriffin:
My view is that because the two warnings are unrelated, combining them is not a good idea. You should theoretically be able to remove one of the warnings and the other still be present, and it be obvious that they're separate. I don't think it's super elegant to have them displayed separately, but it is pretty clear which I think is the priority. And for the majority of the time, I'm assuming people will only see one or the other, so it's somewhat of an edge case we're designing for here.

billygriffin:
@Daniel-McCarthy I conferred with @outofambit and @desktop/design. We're good with the two separated warnings here. Thanks for the great question and explanation.

Daniel-McCarthy:
@billygriffin Wonderful. Thanks for your time, and thanks to the rest of the design team for their thoughts. I am happy to hear we have a conclusion on this matter.

app/src/ui/lib/branch-name-warnings.tsx
```diff
@@ -46,3 +46,26 @@ export function renderBranchHasRemoteWarning(branch: Branch) {
...
+
+  if (alreadyExistsOnRemote === false) {
+    return null
+  }
+
+  return (
+    <Row className="warning-helper-text">
+      <Octicon symbol={OcticonSymbol.alert} />
+      <p>
+        A branch named <Ref>{sanitizedName}</Ref> already exists on remote.
```
shiftkey: :art: on _the_ remote

outofambit:
@shiftkey are we good to merge this?

PR #5597, opened 2018-09-11
Adding SSH Tests

say25:
Saw https://github.com/desktop/desktop/pull/5576 and remembered how I typically clone via URL with SSH. 

I had realized there were no tests so I thought I would add them. Tests are the same as the set I renamed to `openRepoViaHttps` with different URL strings.

### DIFF
--- a/app/test/unit/parse-app-url-test.ts
+++ b/app/test/unit/parse-app-url-test.ts
@@ -24,7 +24,7 @@ describe('parseAppURL', () => {
     })
   })
 
-  describe('openRepo', () => {
+  describe('openRepoViaHttps', () => {
     it('returns right name', () => {
       const result = parseAppURL(
         'github-mac://openRepo/https://github.com/desktop/desktop'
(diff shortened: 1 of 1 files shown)


app/test/unit/parse-app-url-test.ts
```diff
@@ -24,7 +24,7 @@ describe('parseAppURL', () => {
     })
   })
 
-  describe('openRepo', () => {
+  describe('openRepoViaHttps', () => {
```
shiftkey: :art: these don't need to be squished together like that (it was just that way because `openRepo` is the command. 

What about if we do `openRepo via HTTPS` and `openRepo via SSH below`?

say25: as comments or seperate `describe`s

shiftkey: Just use spaces in the first parameter string so when they displayed in the test runner they're human-friendly text.

PR #5182, opened 2018-07-18
Upgrade Prettier to 1.14 (WIP)

j-f1:
I’ll comment when Prettier 1.14 is officially released and this is ready for review.

PR #5602, opened 2018-09-11
use a dedicated phase for linting the source on Azure Pipelines

### LINKED ISSUE #2717: Run the linters on Travis on Linux
Currently, the project is tested on macOS using CircleCI and on Windows with AppVeyor. Since Travis has quick-loading container-based builds, we should use those to run our linters, and just run the tests on macOS and Windows.

shiftkey:
Resolves #2717

This is a variation on #2717, and I chose Pipelines because it's only a config change to add a new agent (I'm gonna leave the other providers as-is).

While I'm still skeptical that running it on one platform will cover everything (see https://github.com/desktop/desktop/pull/5182#issuecomment-408877361 for an example) I'd love to see just how much this speeds things up overall.

Yes, there's probably more optimizations we can do here to reduce CI time, but baby steps...

### DIFF
--- a/vsts.yml
+++ b/vsts.yml
@@ -12,9 +12,6 @@ phases:
           yarn install --force
         name: Install
       - script: |
-          yarn lint
-        name: Lint
-      - script: |
           yarn build:prod
         name: Build
       - script: |
@@ -37,9 +34,6 @@ phases:
           yarn install --force
         name: Install
       - script: |
-          yarn lint
-        name: Lint
-      - script: |
           yarn build:prod
         name: Build
       - script: |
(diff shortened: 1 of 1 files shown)


shiftkey:
The time saved from each Pipelines build agent (compared to [latest `master`](https://github.visualstudio.com/Desktop/_build/results?buildId=6670&view=logs))

 - Windows: 13:12 -> 10:51 (Appveyor: total build time 12:09)
 - Linux: 9:26 -> 8:22 (Travis: total build time 13:13)
 - macOS: 8:00 -> 6:31 (CircleCI: total build time 11:43)

vsts.yml
```diff
@@ -57,12 +51,28 @@ phases:
...
+  - phase: Lint
+    queue: Hosted Linux Preview
+    steps:
+      - script: |
+          apt-get update
+          apt-get install -y --no-install-recommends libsecret-1-dev
+      - task: NodeTool@0
+        inputs:
+          versionSpec: '8.11.1'
+      - task: geeklearningio.gl-vsts-tasks-yarn.yarn-installer-task.YarnInstaller@2
```
iAmWillShepherd: Is this how Yarn is installed? If so, should that version be `1.9.4`? that's the version we use locally.

shiftkey: We don't need to keep updating this because #5335 means it'll use whatever is local, but we need _a_ version of `yarn` installed, so I'm just leaving it as what the other phases are set to (macOS already has it installed).

shiftkey:
As this is an infrastructure change and doesn't contain any user-facing changes, this can be merged after the 1.4.0 release today.

PR #5608, opened 2018-09-11
Branch names no longer allow starting with a plus

### LINKED ISSUE #5594: Block invalid refspec in branch names
We allow Desktop to name branches as `+`, which will trigger a `git -c credential.helper= push origin + --set-upstream --progress exited with an unexpected code: 128. fatal: Invalid refspec '+'` error when attempting to publish the branch. 

We should disable the ability for users to name their branches `+` and/or clarify in the messaging why we do not allow such branches to be published.

Note, you cannot name a repo `+`, as is converted into `-`. 

screen shot 2018-09-10 at 3 44 22 pm

Daniel-McCarthy:
Fixes #5594. This is in response to an issue mentioning getting fatal "invalid refspec" errors when creating a branch that begins with +. (For example `+` or `+A`)

Branch names like `A+B` are allowed and do not result in this error. Therefore I have updated the branch name sanitization regular expression to replace '+' if it is the beginning character (and not replace it if it is anywhere after the first character).

This is the behavior with these changes:
peek 2018-09-11 14-41

___
On the original Issue there is discussion about potentially adding to or updating warnings regarding invalid branch names. I have added more details about the current behavior. If there's any changes needed regarding that matter I would be happy to add them in this or a follow up pull request.
___
I have additionally added two unit tests. Branch names already were not allowed to start with `-`. However, there was no unit test verifying this functionality. I have added one, and one that verifies the same for the `+` at the start of a name.

I found two more things worth mentioning related to these unit tests:
- The name of the unit test file is `sanitize-test-name-test`. However, this is somewhat confusing. Perhaps it should be named `sanitize-branch-name-test`?
- `sanitize-test-name-test` only tests branch names. There does not appear to be any tests for sanitizing repository names. I would be happy to add some as a follow up to this pull request.

### DIFF
--- a/app/src/lib/sanitize-branch.ts
+++ b/app/src/lib/sanitize-branch.ts
@@ -6,7 +6,7 @@ const invalidCharacterRegex = /[\x00-\x20\x7F~^:?*\[\\|""<>]+|@{|\.\.+|^\.|\.$|\
 
 /** Sanitize a proposed branch name by replacing illegal characters. */
 export function sanitizedBranchName(name: string): string {
-  return name.replace(invalidCharacterRegex, '-').replace(/^[-\+]/g, '')
+  return name.replace(invalidCharacterRegex, '-').replace(/^[-\+]*/g, '')
 }
 
 /** Validate a branch does not contain any invalid characters */
--- a/app/src/lib/sanitize-branch.ts
+++ b/app/src/lib/sanitize-branch.ts
@@ -6,7 +6,7 @@ const invalidCharacterRegex = /[\x00-\x20\x7F~^:?*\[\\|""<>]+|@{|\.\.+|^\.|\.$|\
 
 /** Sanitize a proposed branch name by replacing illegal characters. */
 export function sanitizedBranchName(name: string): string {
-  return name.replace(invalidCharacterRegex, '-').replace(/^-/g, '')
+  return name.replace(invalidCharacterRegex, '-').replace(/^[-\+]/g, '')
 }
 
 /** Validate a branch does not contain any invalid characters */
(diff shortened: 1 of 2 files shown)


tierninho:
thanks for the PR, @Daniel-McCarthy 

> If there's any changes needed regarding that matter I would be happy to add them in this or a follow up pull request.

I think we should also prevent the same characters in `Branch > Rename...` modal

app/src/lib/sanitize-branch.ts
```diff
@@ -6,7 +6,7 @@ const invalidCharacterRegex = /[\x00-\x20\x7F~^:?*\[\\|""<>]+|@{|\.\.+|^\.|\.$|\
 
 /** Sanitize a proposed branch name by replacing illegal characters. */
 export function sanitizedBranchName(name: string): string {
-  return name.replace(invalidCharacterRegex, '-').replace(/^-/g, '')
+  return name.replace(invalidCharacterRegex, '-').replace(/^[-\+]/g, '')
```
iAmWillShepherd: Should we also at the 0..N modifier `*`? Otherwise, `++but-can-still-keep-the-rest` would be a valid branch name.

Daniel-McCarthy:
@tierninho No problem!
>I think we should also prevent the same characters in Branch > Rename... modal

Yep! The the Create-Branch-Dialog and the Rename-Branch-Dialog rely on this sanitization function and regex. So they both use this updated behavior.

peek 2018-09-11 15-08

app/src/lib/sanitize-branch.ts
Daniel-McCarthy: Great catch, I will update this now! Thank you for the feedback.

Daniel-McCarthy: Well, with the `*` it should be able to detect several + and - symbols at the start. Should I combine the + and - tests into a single test?

PR #5616, opened 2018-09-12
Onboarding metrics

### LINKED ISSUE #5549: [Metrics] Onboarding
One area we intend to explore over the next few months is how easy it is for people to get up to speed with GitHub Desktop after first downloading it. To that end, we'd love to have several metrics in place as a baseline prior to developing solutions. This issue is intended to capture what we want to know, not to specify the exact implementation.

Where action is each of: [adding repo, cloning repo, creating repo, committing, being on a non-default branch, pushing to GitHub]
1. Time from first launch of GitHub Desktop to [action]

This is intended to help us understand how quickly people are getting to the core functions of the app. We want the time between launch and action for each of these to go down, providing the number in 2 below either stays the same or rises.

2. Percent of installs that reach each action specified above.

This is intended to help us understand what percentage of people are actually getting to each action so we know whether they're actually deriving those pieces of value or not. We want these numbers generally to go up.

3. How did they sign in during welcome flow? Did they type username/password or use the web flow?

Knowing this will help us determine which people are using and which we nudge people toward going forward, and how. @niik and I are thinking that we prefer the web flow to make sure we have everything we need so things don't break for them, but that the experience might not be that great, and some users might just be used to a username/password type flow.

4. How long are people spending inside of the wizard from first launch + which part did they make it to before abandoning or finishing?

This is intended to evaluate the effectiveness of the current wizard. It's informational in nature as we decide how to proceed, but isn't something we're trying to move the needle on.

cc: @desktop/analytics @niik

niik:
First steps towards implementing the onboarding metrics outlined in #5549.

I've skipped the last wizard step metric because the wizard is mandatory so there shouldn't ever be a case where the completion step isn't the opt-in step.

### TODO

- [x] Welcome Wizard sign in type
- [x] Manual testing

Fixes #5549


app/src/lib/stats/stats-store.ts
```diff
@@ -24,6 +26,17 @@ const StatsOptOutKey = 'stats-opt-out'
 /** Have we successfully sent the stats opt-in? */
 const HasSentOptInPingKey = 'has-sent-stats-opt-in-ping'
 
+const WelcomeWizardInitiatedAtKey = 'welcome-wizard-initiated-at'
```
iAmWillShepherd: I feel like these keys can be defined as an enum, which has a benefit of giving of addition type safety in the `timeTo` and `getLocalStorageTimestamp` functions.

If you built this to be as generic as possible because we will need to keep track of stepped timing info, then feel free to resolve as what's here is fine for that scenario.

niik: Given that we want to get a release out with this sooner rather than later let's punt on this discussion for now and land this PR without such a change.

PR #5618, opened 2018-09-12
You can now add a co-author to a repo that's added after onboarding experience

### LINKED ISSUE #5490: Co-authors field doesn't appear after clicking 'add' button
In version 1.3.4 in macOS, clicking the "add co-authors" button just toggled the icon from blue to gray, and made the text say "Add co-authors" and "Remove co-authors", but did not make the field to type the co-authors name appear.

image 25

This was the first time running the app, right after signing into GitHub and adding a local repository. Upon exiting and reopening the app, the "add co-authors" button worked as expected.

iAmWillShepherd:
Fixes #5490

**New behavior:**
co-auth

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -160,7 +160,6 @@ import { findAccountForRemoteURL } from '../find-account'
 import { inferLastPushForRepository } from '../infer-last-push-for-repository'
 import { MergeResultKind } from '../../models/merge'
 import { promiseWithMinimumTimeout } from '../promise'
-import { getDefaultRemote } from '../../lib/git/remote'
 
 /**
  * Enum used by fetch to determine if
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3439,7 +3439,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         const addedRepo = await this.repositoriesStore.addRepository(
           validatedPath
         )
-        
+
         // initialize the remotes for this new repository to ensure it can fetch
         // it's GitHub-related details using the GitHub API (if applicable)
         const gitStore = this.getGitStore(addedRepo)
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2358,7 +2358,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
   private async matchGitHubRepository(
     repository: Repository
   ): Promise<IMatchedGitHubRepository | null> {
-    const remote = await getDefaultRemote(repository)
+    const gitStore = this.getGitStore(repository)
+    const remote = gitStore.defaultRemote
     return remote !== null
       ? matchGitHubRepository(this.accounts, remote.url)
       : null
@@ -3438,6 +3439,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
         const addedRepo = await this.repositoriesStore.addRepository(
           validatedPath
         )
+        
+        // initialize the remotes for this new repository to ensure it can fetch
+        // it's GitHub-related details using the GitHub API (if applicable)
+        const gitStore = this.getGitStore(addedRepo)
+        await gitStore.loadRemotes()
+
         const [refreshedRepo, usingLFS] = await Promise.all([
           this.repositoryWithRefreshedGitHubRepository(addedRepo),
           this.isUsingLFS(addedRepo),
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2358,6 +2358,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     repository: Repository
   ): Promise<IMatchedGitHubRepository | null> {
     const gitStore = this.getGitStore(repository)
+    await gitStore.loadRemotes()
     const remote = gitStore.defaultRemote
     return remote !== null
       ? matchGitHubRepository(this.accounts, remote.url)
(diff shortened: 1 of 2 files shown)


app/src/ui/changes/sidebar.tsx
```diff
@@ -76,8 +76,8 @@ export class ChangesSidebar extends React.Component<IChangesSidebarProps, {}> {
 
   private receiveProps(props: IChangesSidebarProps) {
     if (
-      props.repository.id !== this.props.repository.id ||
-      !this.autocompletionProviders ||
+      this.autocompletionProviders === null ||
+      props.repository !== this.props.repository ||
```
shiftkey: I'd love to understand what the actual fix here is, as we're no longer checking `props.repository.id` here which might be a different equality check here.

outofambit: so the problem was that the `id` check doesn't capture situations where the repository `id` hasn't changed, but its properties have (like `Repository.GitHubRepository` going from `null` to something).

so, once a repository with a null `GitHubRepository` has been shown in the sidebar, you would have to trigger one of these conditions to get coauthoring to work in the UI.

shiftkey: Cool, that's good to know.

What if we compared `.hash` instead, which also touches `.gitHubRepository`?

https://github.com/desktop/desktop/blob/9c9c44fb9128edce2821e01310b9ad06a635d52a/app/src/models/repository.ts#L47-L54

app/src/lib/stores/app-store.ts
```diff
@@ -2357,8 +2358,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   private async matchGitHubRepository(
     repository: Repository
   ): Promise<IMatchedGitHubRepository | null> {
-    const gitStore = this.getGitStore(repository)
-    const remote = gitStore.defaultRemote
+    const remote = await getDefaultRemote(repository)
```
shiftkey: I wanted to get away from sprinkling `getDefaultRemote` (which requires spawning Git to enumerate all remotes) around the place as part of #5399 because this value is never cached, and the other code paths around refreshing a repository already do a `gitStore.loadRemotes()` to load and cache the default remote:

 - `AppStore._refreshRepository()`
 - `AppStore._publishRepository()`

Are we able to ensure we're in a good state before doing this first refresh (like the other code paths) by loading the remotes? This patch seems to work fine for me:

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index a8d513107..fe9ac69e0 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -160,7 +160,6 @@ import { findAccountForRemoteURL } from '../find-account'
 import { inferLastPushForRepository } from '../infer-last-push-for-repository'
 import { MergeResultKind } from '../../models/merge'
 import { promiseWithMinimumTimeout } from '../promise'
-import { getDefaultRemote } from '../../lib/git/remote'
 
 /**
  * Enum used by fetch to determine if
@@ -2358,7 +2357,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
   private async matchGitHubRepository(
     repository: Repository
   ): Promise<IMatchedGitHubRepository | null> {
-    const remote = await getDefaultRemote(repository)
+    const gitStore = this.getGitStore(repository)
+    const remote = gitStore.defaultRemote
     return remote !== null
       ? matchGitHubRepository(this.accounts, remote.url)
       : null
@@ -3438,6 +3438,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
         const addedRepo = await this.repositoriesStore.addRepository(
           validatedPath
         )
+
+        // initialize the remotes for this new repository to ensure it can fetch
+        // it's GitHub-related details using the GitHub API (if applicable)
+        const gitStore = this.getGitStore(addedRepo)
+        await gitStore.loadRemotes()
+
         const [refreshedRepo, usingLFS] = await Promise.all([
           this.repositoryWithRefreshedGitHubRepository(addedRepo),
           this.isUsingLFS(addedRepo),
```

app/src/ui/changes/sidebar.tsx
outofambit: i love that idea

outofambit: f2823c7

iAmWillShepherd:
🔢

app/src/lib/stores/app-store.ts
shiftkey: We could exclude it from [`app/src/lib/git/index.ts`](https://github.com/desktop/desktop/blob/master/app/src/lib/git/index.ts) so that we can write tests against it, but callers in the app are not using it.

https://github.com/desktop/desktop/blob/741a5534bcaed63a5b6c5637feb59bd64bdc27c3/app/src/lib/git/index.ts#L19

shiftkey: Alternatively, if those tests are not valuable and there's no public callers we should just :fire: the code and ensure we have good test coverage elsewhere

outofambit: lets continue discussion at #5626

app/src/lib/stores/app-store.ts
```diff
@@ -160,6 +160,7 @@ import { findAccountForRemoteURL } from '../find-account'
 import { inferLastPushForRepository } from '../infer-last-push-for-repository'
 import { MergeResultKind } from '../../models/merge'
 import { promiseWithMinimumTimeout } from '../promise'
+import { getDefaultRemote } from '../../lib/git/remote'
```
shiftkey: This seems pretty close to done - just an unnecessary `import` blocking the compile step

outofambit:
💖

iAmWillShepherd:
@outofambit thanks for that fix 👍I forgot to push that commit!

PR #5623, opened 2018-09-12
ignore envrc

outofambit:
`.envrc` is the default env file path for [direnv](https://direnv.net/)


.gitignore
```diff
@@ -11,3 +11,4 @@ app/node_modules/
 .vs/
 .eslintcache
 *.iml
+.envrc
```
Akshat-Jain: Please add a newline here. Thanks :-)

PR #5624, opened 2018-09-12
Update yarn version in setup.md

joaomlneto:
Update `yarn` version to match #5335

I'm having some trouble running the linux fork and spent quite some time trying to downgrade yarn assuming that was the issue… :-P


outofambit:
also I'm sorry this cost you some time 😞

PR #5626, opened 2018-09-12
tslint deprecate warning

outofambit:
* adds a warning to the output of `yarn lint` when a method that has been [marked as deprecated](http://usejsdoc.org/tags-deprecated.html) is used
* uses `getDefaultRemote` as an example of deprecation

I'm also open to just deleting `getDefaultRemote` and its unit tests but I figured this deprecation mechanism might be useful for other situations in the future.

We can also make this an `error` in linting, if people would be open to that

**To Do:**
- [x] fix new linting errors


outofambit:
@niik thanks for the feedback! i totally agree.

> Is there a reason why we're creating a new tslint.json here instead of reusing the one at the root of the repository?

this is so tslint only checks for deprecation errors in `app/src/`, but i think that putting it at the root makes more sense if we agree to not use deprecate rules in our own code

@shiftkey i'll take a look at this patch, thank you! i might put that in a separate PR, and let this one just deal with adding the deprecate rule to tslint

outofambit:
see #5634

outofambit:
thanks @shiftkey!

shiftkey:
@outofambit is `help wanted` supposed to be on here?

outofambit:
@shiftkey it was, but i went ahead and finished this :)

fixing the deprecation was easy since the error message clearly laid out how to fix it

outofambit:
🍕

app/src/ui/resizable/resizable.tsx
```diff
@@ -113,7 +113,7 @@ export class Resizable extends React.Component<IResizableProps, {}> {
   }
 }
 
-export interface IResizableProps extends React.Props<Resizable> {
+export interface IResizableProps extends React.ClassAttributes<Resizable> {
```
niik: I think we should just drop the extends part here. This was from a time when we needed to get the `children` prop but these days that's automatically included by a type union.

outofambit:
thanks @niik! 🍜

app/src/ui/resizable/resizable.tsx
```diff
@@ -113,7 +113,7 @@ export class Resizable extends React.Component<IResizableProps, {}> {
   }
 }
 
-export interface IResizableProps extends React.Props<Resizable> {
+export interface IResizableProps {
```
iAmWillShepherd: Is this change part of the work related to the deprecation warnings?

outofambit:
@niik are we good to merge this?

Issue #5511, opened 2018-08-29
Code Coverage Metrics


outofambit:
Hi, I would like to explore implementing test coverage metrics in our codebase. Ideally, we would have something that runs as part of our CI process that provides an overall test coverage metrics as well as the ability to drill down and see the contributing factors to those high level metrics. We can also have that information surfaced in some form in open PRs.

Here are what I see as the goals/benefits of that. Do these seem reasonable? Anything I’m missing that we’d like to call out?

**Goals:**
* Provide a baseline understanding of how thoroughly different parts of our codebase are tested.
* Help identify areas we’d like to test more thoroughly and if there are any areas that we feel are overly tested. This could result in `tech-debt` tasks to add coverage on certain areas, or allow us to mark tests we’d remove when they require upkeep.
* Ensure we are intentional about increasing or decreasing code coverage over time through the PR process.

att. @desktop/maintainers

shiftkey:
👍 to this work.

Some team-based ideas off the top of my head:

 - when updating a component, review whether it can be simplified by extracting functionality that makes it easier to test
 - reviewers encouraging PRs for fixes to also include regression tests, where appropriate (e.g. #5434)

Some specific "nice to have" integration ideas:

 - PR integration to indicate how an incoming PR impacts the code coverage stats
 - tracking trends over time about how we're keeping up with coverage goals

shiftkey:
#5810 raised a commit status error because the PR only touches `.md` files - https://codecov.io/gh/desktop/desktop/commit/72e597785defdd9c211d0083fa724b595e2c0464 - not sure how to get it to ignore that situation, logging it here to ensure I don't forget about.

shiftkey:
Just noticed that Appveyor doesn't have `gcov` installed - which is part of GCC - and it quietly errors (but still seems to succeed uploading code coverage):

```
yarn run v1.10.1
$ codecov -f coverage/*.json
  _____          _  
 / ____|        | |  
| |     ___   __| | ___  ___ _____   __  
| |    / _ \ / _` |/ _ \/ __/ _ \ \ / /  
| |___| (_) | (_| |  __/ (_| (_) \ V /  
 \_____\___/ \__,_|\___|\___\___/ \_/  
                                v3.1.0
==> Detecting CI Provider
    AppVeyor CI Detected
==> Configuration: 
    Endpoint: https://codecov.io
{ commit: 'bd3938780b3977a824c0f17727fd49d5038aa6ac',
[...]

PR #5636, opened 2018-09-13
codecov reporting

outofambit:
This is a minimal first step at tracking code coverage (#5511). This PR sets up code coverage metrics to be tracked in [codecov](https://codecov.io/gh/desktop/desktop), but doesn't integrate it into our flow (for the vast majority of cases). Codecov will send a status on PRs, but it should always be green unless the coverage drops dramatically. You can click "details" to see the report.

I'd like to let this sit for a bit while we/I think about what ways we'd like to use this tool going forward.

To Do:

setup up codecov reporting on:
- [x] circleci
- [x] appveyor
- [x] travis

finally:
- [x] make sure reports are merged properly


.codecov.yml
```diff
@@ -0,0 +1,19 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 1
+  round: nearest
+  range: '10...100'
+
+  status:
```
outofambit: configuration for statuses from codecov:
- only on prs
- only fails if coverage drops by 25%

.codecov.yml
```diff
@@ -0,0 +1,19 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 1
+  round: nearest
+  range: '10...100'
```
outofambit: this just configures the color threshholds

.codecov.yml
```diff
@@ -0,0 +1,19 @@
...
+  status:
+    project:
+      default:
+        target: auto
+        threshhold: 25
+        base: pr
+    patch: no
+    changes: no
+
+comment: off
```
outofambit: *no* commenting by codecov with coverage data on PRs

shiftkey: Do you think this is something we'll want to turn on down the track? What does the commenting look like?

.codecov.yml
shiftkey: Maybe throw a YAML comment in there if there's details that will help the next time we're in here making changes?

.codecov.yml
outofambit: perhaps, once we tune our thresholds and whatnot! it looks like this: https://docs.codecov.io/docs/pull-request-comments

.codecov.yml
outofambit: f238bfd

shiftkey:
> I'd like to let this sit for a bit while we/I think about what ways we'd like to use this tool going forward.

I have two general questions about this PR:

 - what happens after we approve and merge this PR?
 - what outstanding questions around code coverage you have in mind that you'd like to figure out beyond this PR (or get feedback on from others)?

outofambit:
@shiftkey 
> what happens after we approve and merge this PR?

we start to get metrics for code coverage for new branches off of master. we'll get statuses from codecov too on new prs.

> what outstanding questions around code coverage you have in mind that you'd like to figure out beyond this PR (or get feedback on from others)?

- what level of integration do we want with code coverage in our prs?
- it looks like there's a way to focus feedback and track specific parts of the codebase in codecov. is that something that would be more useful than a single number?
- what is a good "rule of thumb" number for enough coverage?

shiftkey:
@outofambit I'm 👍 on both of these things, and suggest using a `meta` issue so we have a place to discuss this further.

outofambit:
@shiftkey 👍 does #5511 seem ok for that?

shiftkey:
@outofambit yep, that works for me!

PR #5570, opened 2018-09-07
Add support for Typora editor in macOS, Windows

joaomlneto:
Adds support for the [Typora](https://typora.io) external editor

Small-ish quirk: focus doesn't jump to the editor when opened.

Fixes #5558

PR #5643, opened 2018-09-14
Add support for Typora editor

### LINKED ISSUE #5558: add support for Typora
There are currently two choices for an external editor in a local repo on my Mac OS machine: `Sublime Text` and `BBEdit`. These are likely popular choices, but they're certainly not the only good choices. 

I think the best solution long-term is to allow users to add any app they wish. If that's impractical, I'd like to see at least the following two apps added to Mac OS: 
1. Typora 
2. Text Edit

joaomlneto:
This is PR #5570 minus the git history disaster.

Adds support for the [Typora](https://typora.io) external editor

- macOS: works fine; focus doesn't jump to the editor when opened (https://github.com/typora/typora-issues/issues/1792)
- Windows: works fine
- Linux: seems to work fine

Closes #5558


app/src/lib/editors/win32.ts
```diff
@@ -137,6 +141,14 @@ function getRegistryKeys(
             'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Adobe ColdFusion Builder 2016',
         },
       ]
+    case ExternalEditor.Typora:
+      return [
+        {
+          key: HKEY.HKEY_LOCAL_MACHINE,
+          subKey:
+            'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\{37771A20-7167-44C0-B322-FD3E54C56156}_is1',
```
shiftkey: There's also a 32-bit Windows version of Typora that installs to `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\{37771A20-7167-44C0-B322-FD3E54C56156}_is1`, so I'm not sure if anyone is using it. Happy to leave that out for now, we don't have plans to support 32-bit Windows in the near-term anyway.

I also noticed both Windows installers are not signed, and reported this upstream: https://github.com/typora/typora-issues/issues/1801

joaomlneto: > There's also a 32-bit Windows version of Typora

I missed that! I'll add it for the sake of completion.

Tested and works on my Windows 10.

joaomlneto: (I have no idea how to undismiss this!)

joaomlneto:
Was playing with the `ImgBot` and blindly pushed his changes to the `master` branch in my `desktop` fork. Due to sad standards by me, they were included here (as I'm pushing directly from master) >_> sorry!

... If I worked at Airbus, it would rain planes from the sky. I'll try to be more careful; I'm used to repositories with small number of contributors!

PR #5205, opened 2018-07-20
bump node-sass to version which supports Node 10

shiftkey:
See this upstream issue for all the boring details: https://github.com/sass/node-sass/issues/2345

- [ ] verify **yarn build:dev** && **yarn start** and hot reload on Windows
- [ ] verify **yarn build:prod** && **yarn start:prod** on Windows
- [ ] verify **yarn build:dev** && **yarn start** and hot reload on macOS
- [ ] verify **yarn build:prod** && **yarn start:prod** on macOS
- [ ] verify **yarn build:dev** && **yarn start** and hot reload on Linux
- [ ] verify **yarn build:prod** && **yarn start:prod** on Linux

PR #5644, opened 2018-09-14
upgrade node-sass to a version that supports Node 10

shiftkey:
> "This will be easy," I think to myself as I `yarn upgrade --latest node-sass`. "There's a version out that supports Node 10, and life will be much simpler!"

This started a couple of months ago, when I'd heard @nerdneha having troubles with building Desktop on Node 10. I found #5204 which fixed one issue, but my earlier attempt to fix this in #5205 lead to something shocking - an app devoid of colour.

What did I break? What happened to the beautiful app? This was supposed to be a minor version change! I put this aside, questioning everything and contemplating joining a monastery again (probably the Trappist ones) and went about my life.

The Node 10 issue came up again earlier in the week, and I had another look at this with fresh eyes and lots of Google tabs. The actual problem was that I wasn't seeing the values from `primer-support` inserted into our CSS:

Same thing with the crash view:

After much digging I find that `node-sass` [`v4.8.0`](https://github.com/sass/node-sass/releases/tag/v4.8.0) mentions upgrading to [Futura](https://github.com/sass/libsass/releases/tag/3.5.0) - the release name for `libsass` 3.5 which corresponds to SASS 3.5.

I switch over to the SASS [release notes](http://sass-lang.com/documentation/file.SASS_CHANGELOG.html) and find this gem:

So yes, a cascading series of "minor" updates broke previously-working code with no warnings about the change in behaviour! 

I've updated all the affected rules to use the recommended `#{}` token to ensure our generated stylesheet matches what we were previously generating, but if someone wants to figure out The Correct Way™ to consume `primer-support` styles without these kludges I'm happy to jam on them with it.

But for the moment, with this PR **I'm now able to build Desktop on Node 10 on macOS** (I'll confirm the others as I go through testing out this PR).

 - [x] test `start` and `start:prod` on macOS
 - [x] test `start` and `start:prod` on Windows 
 - [x] test `start` and `start:prod` on Ubuntu 18.04

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -12,32 +12,32 @@
 $overlay-background-color: rgba(0, 0, 0, 0.4);
 
 :root {
-  --color-new: $green;
-  --color-deleted: $red;
-  --color-modified: $yellow-700;
-  --color-renamed: $blue;
-  --color-conflicted: $orange;
+  --color-new: #{$green};
+  --color-deleted: #{$red};
+  --color-modified: #{$yellow-700};
+  --color-renamed: #{$blue};
+  --color-conflicted: #{$orange};
 
-  --text-color: $gray-900;
-  --text-secondary-color: $gray-500;
-  --background-color: $white;
+  --text-color: #{$gray-900};
+  --text-secondary-color: #{$gray-500};
+  --background-color: #{$white};
 
   --button-height: 25px;
 
-  --button-background: $blue;
+  --button-background: #{$blue};
   --button-border-radius: 2px;
-  --button-hover-background: lighten($blue, 5%);
-  --button-text-color: $white;
-  --button-focus-border-color: $blue-600;
+  --button-hover-background: #{lighten($blue, 5%)};
+  --button-text-color: #{$white};
+  --button-focus-border-color: #{$blue-600};
 
-  --link-button-color: $blue;
-  --link-button-hover-color: $blue-600;
+  --link-button-color: #{$blue};
+  --link-button-hover-color: #{$blue-600};
 
-  --secondary-button-background: $gray-000;
-  --secondary-button-hover-background: $white;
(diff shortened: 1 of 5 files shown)


niik:
Great investigation @shiftkey. A bit of a bummer that this is something we need to spend our time on right now. Is getting to node 10 a priority due to some external dependency requiring it? What version of node does Electron 3 use?

I'm concerned about how this affects the compilation of primer itself. If I'm remember things correctly we still consume the raw primer scss and compile it. I think it'd be super helpful if @donokuda or @ampinsk could talk to design-systems and get their thoughts on this before we move forward.

shiftkey:
> Is getting to node 10 a priority due to some external dependency requiring it? 

This was [called out](https://github.com/desktop/desktop/pull/5538#discussion_r217456237) by @outofambit as a thing that new contributors encountered at WSC, and she was working on docs to try and workaround this:

> mostly it was that they had a newer version of nodejs installed on their machine (and didn't want to uninstall it). the brew instructions may have helped some of them, but many wanted to keep their current version of node installed and work on our project too. i think part of it too was that yarn install failed for newer versions of node, but i can't remember for sure.

> What version of node does Electron 3 use? 

Electron's internal Node goes up to 10.2.0 (it was `10.2.0` in Electron `3.0.0-beta1` but they're now at `-beta10` and I've not been following the latest betas), so there might be other native module dependencies that need updating as part of that work (`keytar` I think now works on with Node 10, but there might be others).

> I'm concerned about how this affects the compilation of primer itself.

:thumbsup: to looping in `design-systems` on this, to ensure we're doing things properly with this change.

outofambit:
closing this until we can prioritize it

shiftkey:
I've opened https://github.com/desktop/desktop/issues/5876 to ensure this doesn't get forgotten about.

PR #5645, opened 2018-09-14
documenting an aspirational structure of the app directory

shiftkey:
I've had #5424 fermenting in the background but I wanted to reorganize the application source in a way that makes more sense to gain other benefits.

One example is to avoid import chains that look like this:

`app/src/ui/index.tsx` -> `app/src/lib/stores/app-store.ts` -> `app/src/ui/main-process-proxy.ts` -> `app/src/main-process/build-default-menu.ts`

That's jumping from `ui` to `lib` back to `ui` and then to `main-process`, and suggests a couple of changes need to be made:

 - Should `AppStore` move down into `ui`?
 - Is there code in `build-default-menu.ts` that should be properly shared?

For the moment I wanted to use this document as a starting point to gather consensus and figure out how to proceed. Let me know your thoughts @desktop/core @desktop/comrades.

[**Rendered Markdown**](https://github.com/desktop/desktop/blob/docs-on-structure/docs/technical/application-structure.md)

If we're happy with this direction and can get the doc to a point where it's good to merge, I can sketch out the work involved as the next steps.

PS: I've been noodling on using [`dependency-cruiser`](https://www.npmjs.com/package/dependency-cruiser) as a way to analyze these kinds of coupling (it was easy to identify circular references between modules) and it looks possible to automate detecting scenarios like this once we get to a happier place.


docs/technical/application-structure.md
```diff
@@ -0,0 +1,131 @@
...
+ - better organization of modules required by the renderer process
+ - clarification of modules that can be shared between parts of the app, and
+   things that make sense to keep specific to one part of the app
+ - reflect our current usage patterns
+    - Git operations performed in the renderer
+    - stores created and managed in the renderer
+
+What lives in each of these folders:
+
+ - `components` - contains the React components used in our application. I don't
```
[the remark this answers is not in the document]
shiftkey: I've avoided going too deep on this because I didn't want to introduce too much churn in the codebase. Once we've gotten things more organized broadly, we can deal with this specific area.

There's lots of [examples out there](https://github.com/markerikson/react-redux-links/blob/master/project-structure.md) - including [my favourite](http://react-file-structure.surge.sh/), but I wanted to find one that felt natural for our use case, which I think we're still figuring out, so I'm happy to punt on that for now.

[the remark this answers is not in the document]
outofambit: sounds great!

shiftkey:
Friendly bump on this @desktop/maintainers @desktop/comrades before I merge it in sometime next week.

iAmWillShepherd:
I think we should hold off on merging this into desktop/desktop for now (until our app structure is reflective of what's in this doc). We could throw it in our comrades repo for now. But, I'm all for this.

shiftkey:
@iAmWillShepherd I'd rather land sooner rather than later for a couple of reasons:

 - the document already indicates that it's aspirational and not reflecting reality
 - others can come in later and suggest changes to the document (not just those who can push to this branch)
 - we can update the document as we work through making the actual changes, rather than maintaining this as a long-running branch

docs/technical/application-structure.md
```diff
@@ -0,0 +1,131 @@
...
+
+I'm going to walk through these in the order in what might look like a strange
+order, but hopefully it makes sense by the end.
+
+## Shared Modules
+
+These folders should contain modules that can be shared across any of the
+Webpack bundles that Desktop generates.
+
+ - `app/src/models` - contains the shapes used in the codebase to represent
```
iAmWillShepherd: We have 3 types of models in the app:
* API
* Database
* App (the ones we use in our components and the app-store

Would be nice to have those broken out to better signal their intent.

docs/technical/application-structure.md
```diff
@@ -0,0 +1,131 @@
+# Application Structure
```
outofambit: **blocking**: i'm worried that people will miss that this isn't current, but future. Could we make this more obvious in the title? Something like "Aspirational Application Structure" or "Future Application Structure".

And now that I'm thinking about it, this isn't application structure so much as directory structure, so perhaps "Future Directory Structure" or "Future Repo Structure" makes more sense.

I'd like to see this title reflected in the filename too, for extra clarity.

docs/technical/application-structure.md
shiftkey: I don't want to get too detailed in this document about layout (especially in the `renderer` directory), but here's some thoughts:

`app/src/lib/api.ts` defines all of the API models used. I haven't explored this deeply recently but we have this sort of coupling currently to deal with:

   - `app/src/lib` -> `app/src/model` (I want this to go away)
   - `app/src/lib` -> `app/src/ui` (this is fine)

I suspect our API usage is being done in the renderer, so maybe this would live in `app/src/renderer` eventually, and then we can figure out those details. The `api.ts` module rocks in at almost 900 lines, so this is likely a great candidate to be reorganized once it finds it's home.

`app/src/lib/databases` defines these models defined inline in their respective modules. Same question about whether these should eventually live under `app/src/renderer` and can be decomposed later.

As for application models themselves, whether they live in `app/src/models` or `app/src/renderer/models` is something that can be figured out based on usage.

How much detail should we go into right now about this? We should be evolving the document as we update the layout of things anyway.

docs/technical/application-structure.md
shiftkey: Updated in d22a096

docs/technical/application-structure.md
shiftkey: @iAmWillShepherd bump, this looks it got swept up with me renaming the file

iAmWillShepherd: I think the current level of detail is fine and I'm in favor of updating this as we figure things out. 👍

PR #5646, opened 2018-09-14
deprecate RepositorySettingStore

shiftkey:
We never took advantage of using this store to manage state, so this PR removes the store in favour of calling plain functions. 

Other important bits:

 - I updated the function names to better reflect what they actually do
 - existing tests have been ported over and tidied up
 - callers that were using the dispatcher but not updating state have been removed in favour of directly calling the function 

TODO:

 - [x] fix config for Windows-only test failures


iAmWillShepherd:
clean

iAmWillShepherd:
clean

PR #5662, opened 2018-09-16
fix overflow CSS for dialog boxes with dark theme on Windows

### LINKED ISSUE #5629: White box in the bottom right corner of "Terms and Conditions" and "License and Open Source Notices" dialogs
## Description
In #5598 a bug was fixed that caused a white box to appear in the bottom right of the release notes. This bug is also present in the "Terms and Conditions" and "License and Open Source Notices" dialogs.

## Version
* GitHub Desktop: 1.4.0
* Operating system: Microsoft Windows [Version 10.0.17134.228]

## Steps to Reproduce
1. Use windows
2. View "Terms and Conditions" or "License and Open Source Notices"

### Expected Behavior
No white box in the corner.

### Actual Behavior
There's a white box in the corner.

## Additional Information

image

PS: How do I open the release notes in the app? Is that something that you can only see when you have a pending update? I had assumed the "release notes" link in "About GitHub Desktop" would have been changed to show that.

cwongmath:
Fixes #5629 

Changes overflow: scroll to overflow: auto in _acknowledgements.scss and _terms-and-conditions.scss
This commit aims to fix issue #5629 by removing the horizontal scrollbar.

### DIFF
--- a/app/styles/ui/_acknowledgements.scss
+++ b/app/styles/ui/_acknowledgements.scss
@@ -3,7 +3,7 @@
 
   .dialog-content {
     height: 300px;
-    overflow: scroll;
+    overflow: auto;
 
     .license-text {
       font-family: var(--font-family-monospace);
--- a/app/styles/ui/_acknowledgements.scss
+++ b/app/styles/ui/_acknowledgements.scss
@@ -3,7 +3,7 @@
 
   .dialog-content {
     height: 300px;
-    overflow: auto;
+    overflow: scroll hidden;
 
     .license-text {
       font-family: var(--font-family-monospace);
--- a/app/styles/ui/dialogs/_release-notes.scss
+++ b/app/styles/ui/dialogs/_release-notes.scss
@@ -57,7 +57,7 @@
     flex-direction: row;
     padding-left: var(--spacing-triple);
     padding-right: var(--spacing-triple);
-    overflow: scroll hidden;
+    overflow: auto hidden;
     justify-content: space-around;
 
     .column {
--- a/app/styles/ui/dialogs/_release-notes.scss
+++ b/app/styles/ui/dialogs/_release-notes.scss
@@ -57,7 +57,7 @@
     flex-direction: row;
     padding-left: var(--spacing-triple);
     padding-right: var(--spacing-triple);
-    overflow: auto;
+    overflow: scroll hidden;
     justify-content: space-around;
 
     .column {
--- a/app/styles/ui/_terms-and-conditions.scss
+++ b/app/styles/ui/_terms-and-conditions.scss
@@ -3,7 +3,7 @@
 
   .dialog-content {
     height: 250px;
-    overflow: scroll;
+    overflow: auto;
 
     p,
     ol,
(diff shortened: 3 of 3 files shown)


shiftkey:
@desktop/design we talked about how this change now means there's no margin between the content of the dialog and the footer over here: https://github.com/desktop/desktop/issues/5629#issuecomment-421651791 

Before this change:



After:



This seems to align it better with macOS (here's the light theme equivalent):

screen shot 2018-09-16 at 5 59 37 pm

And I think flowing the content to the edge here is fine. Let me know if you have anything else to add about this PR.

outofambit:
> And I think flowing the content to the edge here is fine

agreed! @donokuda and i talked about this specifically while working on tweaks for this and we came to the same conclusion.

app/styles/ui/_acknowledgements.scss
```diff
@@ -3,7 +3,7 @@
 
   .dialog-content {
     height: 300px;
-    overflow: scroll;
+    overflow: auto;
```
j-f1: How about setting `overflow-x: hidden` and `overflow-y: scroll`? That’ll remove the horizontal scrollbar.

billygriffin:
Just wanted to make sure you saw this @cwongmath in case your notifications are as noisy as mine. 😄

app/styles/ui/_acknowledgements.scss
shiftkey: @cwongmath we can set `overflow` to two values, as per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow):

> The `overflow` property is specified as one or two keywords chosen from the list of values below. If two keywords are specified, the first applies to `overflow-y` and the second to `overflow-x`. Otherwise, both `overflow-x` and `overflow-y` are set to the same value.

So it could be:

```css
overflow: scroll hidden;
```

or this:

```css
overflow-y: scroll
overflow-x: hidden
```

I'm slightly leaning towards the first option over the second, but the order of `-x` and `-y` might be confusing at a glance. Also, I'm not sure how we settled on `scroll` over `auto` here, but I'll defer to @j-f1 and @outofambit as they've been following this PR more closely than I have.

Also, can we ensure we do the same fix in `app/styles/ui/dialogs/_release-notes.scss` which was the original example of the issue?

outofambit: good shout @shiftkey. we should use `auto` over `scroll`. i'm also fine with a single line `overflow`. @cwongmath how does that sound?

cwongmath: :ok_hand:

app/styles/ui/dialogs/_release-notes.scss
```diff
@@ -57,7 +57,7 @@
     flex-direction: row;
     padding-left: var(--spacing-triple);
     padding-right: var(--spacing-triple);
-    overflow: auto;
+    overflow: scroll hidden;
```
outofambit: can we make this one `auto hidden`? that means it will only show scroll bars on the release notes when they are needed. thanks @cwongmath!

cwongmath:
idk how the Travis CI build failed...

billygriffin:
🚀 Thanks again @cwongmath!

PR #5663, opened 2018-09-16
[WIP] extract CI logic to make scripts more friendly

### LINKED ISSUE #2696: Extract CI logic into some shared abstraction
We currently have some hard-coded env vars detecting various CI states. They're all hardcoded for our current providers. See https://github.com/desktop/desktop/pull/2694.

It'd be nice to abstract this into something more general so that we can call a function to get the current SHA without worrying where it's being built.

shiftkey:
Fixes #2696

 - [ ] update `getSha` to support Azure Pipelines build agents
 - [ ] ???


shiftkey:
This is actually something I'd like to pull in for the beta tomorrow so I can test publishing in advance of `1.5.0` - any objections @outofambit @iAmWillShepherd?

outofambit:
@shiftkey seems fine to me! @iAmWillShepherd will you do the honors?

PR #5666, opened 2018-09-17
Added support for Cygwin

### LINKED ISSUE #5654: add support for launching Cygwin from "Open in Shell"
**Provide a step-by-step description of the suggested enhancement**
When going to options there is a predefined list of options for shell to use. As you can see I have 3 options, command prompt, power shell and git bash. My enhancement suggestion is to allow manual selection of a shell. For example, I would like to use my cygwin version to run as a shell by clicking "Repository->Open in Cygwin" so I need to be able to select it in the options.
image

**Explain why this enhancement would be useful**
It would be much more comfortable using a shell you are already used to working on instead of the predefined ones. Also there are issues in other shells for me which I was able to fix on cygwin and instead of wasting time on configuring other shells I would just like to easily use my own by click option inside desktop github (right now I just open it outside of desktop github and navigate to the sandbox location manually).
image

LordOfTheThunder:
Added Cygwin support for Windows in "Open in shell"

Fixes #5654

### DIFF
--- a/app/src/lib/shells/win32.ts
+++ b/app/src/lib/shells/win32.ts
@@ -229,13 +229,27 @@ async function findGitBash(): Promise<string | null> {
 }
 
 async function findCygwin(): Promise<string | null> {
-  const registryPath = enumerateValues(
+  const registryPath64 = enumerateValues(
     HKEY.HKEY_LOCAL_MACHINE,
     'SOFTWARE\\Cygwin\\setup'
   )
+  const registryPath32 = enumerateValues(
+    HKEY.HKEY_LOCAL_MACHINE,
+    'SOFTWARE\\WOW6432Node\\Cygwin\\setup'
+  )
 
-  if (registryPath.length === 0) {
+  if (registryPath64.length === 0 && registryPath32.length === 0) {
     return null
+  }
+
+  let registryPath
+  if (registryPath64.length === 0) {
+    registryPath = registryPath32
+  } else if (registryPath32.length === 0) {
+    registryPath = registryPath32
+  } else {
+    // If picking between 64bit and 32bit, 64bit wins.
+    registryPath = registryPath64
   }
 
   const installPathEntry = registryPath.find(e => e.name === 'rootdir')
--- a/app/src/lib/shells/win32.ts
+++ b/app/src/lib/shells/win32.ts
@@ -12,6 +12,7 @@ export enum Shell {
   PowerShellCore = 'PowerShell Core',
   Hyper = 'Hyper',
   GitBash = 'Git Bash',
+  Cygwin = 'Cygwin',
 }
 
 export const Default = Shell.Cmd
@@ -35,6 +36,10 @@ export function parse(label: string): Shell {
 
   if (label === Shell.GitBash) {
     return Shell.GitBash
+  }
+
+  if (label === Shell.Cygwin) {
+    return Shell.Cygwin
   }
 
   return Default
@@ -79,6 +84,14 @@ export async function getAvailableShells(): Promise<
     shells.push({
       shell: Shell.GitBash,
       path: gitBashPath,
+    })
+  }
+
+  const cygwinPath = await findCygwin()
+  if (cygwinPath != null) {
+    shells.push({
+      shell: Shell.Cygwin,
+      path: cygwinPath,
     })
   }
 
@@ -215,6 +228,30 @@ async function findGitBash(): Promise<string | null> {
   return null
 }
 
+async function findCygwin(): Promise<string | null> {
+  const registryPath = enumerateValues(
+    HKEY.HKEY_LOCAL_MACHINE,
+    'SOFTWARE\\Cygwin\\setup'
+  )
+
+  if (registryPath.length === 0) {
+    return null
+  }
+
+  const installPathEntry = registryPath.find(e => e.name === 'rootdir')
+  if (installPathEntry && installPathEntry.type === RegistryValueType.REG_SZ) {
+    const path = Path.join(installPathEntry.data, 'bin\\mintty.exe')
+
+    if (await pathExists(path)) {
+      return path
+    } else {
+      log.debug(`[Cygwin] registry entry found but does not exist at '${path}'`)
+    }
+  }
+
+  return null
+}
+
 export function launch(
   foundShell: IFoundShell<Shell>,
   path: string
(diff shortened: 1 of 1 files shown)


app/src/lib/shells/win32.ts
```diff
@@ -215,6 +228,30 @@ async function findGitBash(): Promise<string | null> {
   return null
 }
 
+async function findCygwin(): Promise<string | null> {
+  const registryPath = enumerateValues(
+    HKEY.HKEY_LOCAL_MACHINE,
+    'SOFTWARE\\Cygwin\\setup'
```
shiftkey: Do you think we should check for 32-bit Cygwin installation as well?

app/src/lib/shells/win32.ts
```diff
@@ -248,6 +285,17 @@ export function launch(
         shell: true,
         cwd: path,
       })
+    case Shell.Cygwin:
+      const cygwinPath = `"${foundShell.path}"`
+      log.info(`launching ${shell} at path: ${cygwinPath}`)
+      return spawn(
+        cygwinPath,
+        [`/bin/sh -lc 'cd "$(cygpath "${path}")"; exec bash`],
```
shiftkey: Do we need to pass in all these arguments? I thought emulating the desktop shortcut would be enough here...

LordOfTheThunder: Unfortunately I didn't find a better way to open Cygwin from a certain path and this is the best solution google search gave me. The --cd flag that works for Git Bash doesn't work for Cygwin and there is just no flag in general in Cygwin for that function.

app/src/lib/shells/win32.ts
LordOfTheThunder: I forgot about that. I will check for 32-bit Cygwin later and post an update.

app/src/lib/shells/win32.ts
shiftkey: I found there's an unofficially-documented `--dir` flag here: https://github.com/mintty/mintty/wiki/Tips#starting-in-a-particular-directory

This launched me to the repository, mapped under `/cygdrive/`



```diff
diff --git a/app/src/lib/shells/win32.ts b/app/src/lib/shells/win32.ts
index fe48bb699..df000116c 100644
--- a/app/src/lib/shells/win32.ts
+++ b/app/src/lib/shells/win32.ts
@@ -290,7 +290,7 @@ export function launch(
       log.info(`launching ${shell} at path: ${cygwinPath}`)
       return spawn(
         cygwinPath,
-        [`/bin/sh -lc 'cd "$(cygpath "${path}")"; exec bash`],
+        ['-i', '/Cygwin-Terminal.ico', '--dir', path, '-'],
         {
           shell: true,
           cwd: path,
```

LordOfTheThunder: Tried changing this to your solution and getting a problem:
image
Also I wasn't able to find any information about --dir flag.

shiftkey: Try wrapping `path` in quotes like ``"${path}"`` and see if that changes the behaviour (I was slack and didn't have unsafe characters in my account name).

app/src/lib/shells/win32.ts
shiftkey: Were you looking in the right location, i.e. `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Cygwin\setup`? That's where I found it after a test install yesterday.

I'd rather not hard-code paths for installers where the user can change the location easily, because the registry is the expected way for us to identify where a program is located on disk.

app/src/lib/shells/win32.ts
LordOfTheThunder: I spent a long time figuring out why your method didn't work for folders with spaces (your last comment didn't fix it unfortunately) and I think I get it now.
basically what --dir does from my understanding is takes the path you give, runs the command "cygpath" on it and then runs cd on the result.

Now if we look at the example of my directory, I did some string manipulation to get the path in correct format for cygpath (before the spawn command):
      var newpath = path.replace(/\\/g, '\\\\')
      newpath = newpath.replace(/\s/, '\\ ')

So now the path I get is:
C:\\Users\\Max\ Kolchinsky\\Documents\\max_sbox\\desktop
when I run cygpath on it inside cygwin I get:

$ cygpath C:\\Users\\Max\ Kolchinsky\\Documents\\max_sbox\\desktop
/cygdrive/c/Users/Max Kolchinsky/Documents/max_sbox/desktop

But the desired result for "cd" command to work is
/cygdrive/c/Users/Max\ Kolchinsky/Documents/max_sbox/desktop
(basically we need one more backslash to escape the space for the cd command to work.

So since the cd command fails when I open the shell it doesn't redirect me to the desired folder.

By the way could you please tell me how you paste diff changes inside comments here?
I tried finding a way in the "Mastering Markdown" page but didn't find anything useful.

shiftkey: > By the way could you please tell me how you paste diff changes inside comments here?

Check out the "Syntax Highlighting" section here: https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown

shiftkey: > In my honest opinion, I think that my previous solution is better because it takes care of this issue regardless of folders that contain spaces.

I'm fine with that, especially if it handles the corner cases

LordOfTheThunder: Thank you. In that case I propose we keep my change. I also added a commit where I fixed the cygwin 32bit issue you pointed out yestarday.

app/src/lib/shells/win32.ts
```diff
@@ -215,6 +228,44 @@ async function findGitBash(): Promise<string | null> {
...
+  const registryPath32 = enumerateValues(
+    HKEY.HKEY_LOCAL_MACHINE,
+    'SOFTWARE\\WOW6432Node\\Cygwin\\setup'
+  )
+
+  if (registryPath64.length === 0 && registryPath32.length === 0) {
+    return null
+  }
+
+  let registryPath
```
shiftkey: :art: this could be rewritten as a oneliner:

```ts
  // always favour 64-bit registry path, if found
  const registryPath =
    registryPath64.length > 0 ? registryPath64 : registryPath32
```

app/src/lib/shells/win32.ts
```diff
@@ -215,6 +228,44 @@ async function findGitBash(): Promise<string | null> {
...
+    registryPath = registryPath32
+  } else if (registryPath32.length === 0) {
+    registryPath = registryPath32
+  } else {
+    // If picking between 64bit and 32bit, 64bit wins.
+    registryPath = registryPath64
+  }
+
+  const installPathEntry = registryPath.find(e => e.name === 'rootdir')
+  if (installPathEntry && installPathEntry.type === RegistryValueType.REG_SZ) {
```
shiftkey: :art: rather than relying on JS truthiness, can we make the first part of this more explicit?

```ts
if (installPathEntry != null && ...
```

LordOfTheThunder: Thank you! I did the correction and also found another corner case which I fixed (in case both 32 and 64 bit are present but the 64 bit path is broken we want the 32bit path to load).

PR #5676, opened 2018-09-18
fix codedov threshold

outofambit:
Turns out i misspelled a key in the codecov yaml and their validator didn't catch it.

~threshhold~ **threshold**

also twiddled with a few numbers and added in the parsers section with the default values since i no longer trust their yaml parsing to handle that for me


shiftkey:
Gonna merge this in because it was previously marking builds as failing that weren't touching application code, e.g. #5318

Issue #1485, opened 2017-05-16
validate commit summary to handle whitespace being entered


shiftkey:
### Description

You can try and create a commit with an empty summary field. It fails.

### Version

What version of GitHub Desktop are you running? This is displayed under the
`About GitHub Desktop` menu item. If you are running from source, include
the commit by running `git rev-parse HEAD` from your local repository.

**GitHub Desktop version:** 0.5.1

**OS version:** Windows 10

### Steps to Reproduce

1. Create a new local repository, pre-populated with some files.
1. Launch `Repository` -> `Open in Explorer` to get to the filesystem
1. Modify a file so some changes appear in the app
1. Enter a space for the commit summary

**Expected behaviour:** A warning that the commit summary is not valid

**Actual behaviour:** No error

**Reproduces how often:** 100%

#### Additional Information

Currently this fails with a somewhat meaningful error, so this feels more like a polish task

<img width="599" src="https://cloud.githubusercontent.com/assets/359239/26087369/b84aac68-3a33-11e7-9102-c31872b0baf6.png">

There's also some design work here that we'd need to consider given how the commit form is limited.

shiftkey:
Moving this to `help wanted` as this is something good for an external contributor, but much like #5240 there will probably be some discussion about what makes sense here.

Daniel-McCarthy:
I believe that this issue can be closed now that #5677 has been merged.
I likely should have used "Fixes #1485" for this case.

outofambit:
@Daniel-McCarthy good catch! thanks!

PR #5677, opened 2018-09-19
Disable commit button when commit summary is only spaces

Daniel-McCarthy:
This pull request is in response to Issue #1485. Currently when a commit summary is only a space, or several spaces, an error is generated and the commit is not created. To avoid this, I have made it disable the commit button when the summary is only spaces.

peek 2018-09-18 18-48
___
Above shows how the commit button behaves with the changes I have implemented. Here is an explanation of what occurs:
- If the commit summary is empty, only a space, or multiple spaces, then the commit button will be disabled and not allow the user to create the commit, which would have resulted in an error.
- If the commit summary starts with a non-space character then the commit button will be enabled and allow the user to create the commit.
- If the user enters one or more spaces followed by a non-space character ('    Test'), then the commit button will enable and the commit summary will be created without the preceding spaces ('Test').
___
In regards to the user experience with this change, I do not foresee it to be too confusing since entering any non-space character enables the commit button. 

However, this is definitely a place to consider how the design for this should work. If there are suggestions for alternative or additional behavior I would be happy to explore them.

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -470,7 +470,10 @@ export class CommitMessage extends React.Component<
 
   public render() {
     const branchName = this.props.branch ? this.props.branch : 'master'
-    const buttonEnabled = this.canCommit() && !this.props.isCommitting
+
+    const isSummaryWhiteSpace = this.state.summary.match(/^\s+$/g)
+    const buttonEnabled =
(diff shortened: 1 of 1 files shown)


outofambit:
we can disregard the codecov status here. this is an issue with our configuration that will be fixed with #5676 

thanks for another PR @Daniel-McCarthy!!

shiftkey:
I've merged #5676 so if you merge `master` into this PR the `codecov` issues should be sorted

Issue #4272, opened 2018-03-16
simplify constructors in models folder by using parameter properties


shiftkey:
See https://github.com/desktop/desktop/pull/4243 for initial discussion.

[TL;DR](https://en.wikipedia.org/wiki/Wikipedia:Too_long;_didn%27t_read):

 - we're mostly :thumbsup: on this idea, with some concerns about readability
 - this is tentatively in-scope for 1.3 (we want to get some of the features in 1.2 out first)
 - this is probably something that an external contributor can help with, but we've got plenty of contributions to work through currently

We'll update this issue when this is ready to be started.

shiftkey:
I'm going to drop this from a given milestone, but mark it open for people who want to dive into a bit of TS refactoring work.

I suggest porting one file at a time, at least initially, as this will help with us reviewing and catching potential errors.

Daniel-McCarthy:
I have continued looking into the other `models` and found one that seems to generate a bit of an issue (small albeit).

With `models/pull-request.ts` is a property called `number`. In the constructor the argument that is used to initialize `number` is called `number_`. When this constructor is switched to use parameter properties, however, a linting issue appears to result due to the name being the same as the type.

This does not seem to be a major issue, but one worth noting for these changes as it may occur elsewhere. In this situation, it would likely be sufficient to specify the name as `requestNumber`.

shiftkey:
I'm not a fan of the `number_` pattern to get around this constraint that [tslint is enforcing](https://palantir.github.io/tslint/rules/variable-name/), but it's also pretty pragmatic. We likely stumbled into that because the Pull Request API returns a field named `number` and we want to avoid confusion.

What if we used `pullRequestNumber` instead, which was used previously in that file?

Daniel-McCarthy:
Sounds great. I will switch it to `pullRequestNumber` and get it ready to make a pull request for that file.

Daniel-McCarthy:
As an update to this, I am getting close to done with this.
- Except for one file that additionally requires a rename, I have all of the files in the `models/` folder complete and ready for a pull request. (I have not done so for all of them yet, but I was planning on doing them in chunks to keep it manageable).
- I am currently working on the `models/diff` folder.

[...]

PR #5682, opened 2018-09-19
Simplify Branch constructor with parameter properties

Daniel-McCarthy:
Issue #4272 suggests to clean-up and refactor the `app/src/models/` files by making use of Typescript parameter properties. I have made changes to the `models/branch` file in order to reduce some redundancy via this feature.

The Issue also mentions recommending taking things one file at a time in regards to the `/models/` folder. If these changes are fitting then I would be happy to continue to another of the files after this pull request.

### DIFF
--- a/app/src/models/branch.ts
+++ b/app/src/models/branch.ts
@@ -41,15 +41,16 @@ export function eligibleForFastForward(
   )
 }
 
-/**
- * A branch as loaded from Git.
- *
- * @param name The short name of the branch. E.g., `master`.
- * @param upstream The remote-prefixed upstream name. E.g., `origin/master`.
- * @param tip The commit associated with this branch
- * @param type The type of branch, e.g., local or remote.
- */
+/** A branch as loaded from Git. */
 export class Branch {
+  /**
+   * A branch as loaded from Git.
+   *
+   * @param name The short name of the branch. E.g., `master`.
+   * @param upstream The remote-prefixed upstream name. E.g., `origin/master`.
+   * @param tip The commit associated with this branch
+   * @param type The type of branch, e.g., local or remote.
+   */
   public constructor(
     public readonly name: string,
     public readonly upstream: string | null,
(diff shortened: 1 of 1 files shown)


shiftkey:
> If these changes are fitting then I would be happy to continue to another of the files after this pull request.

Works for me :tada:

Daniel-McCarthy:
>Works for me

Wonderful. Would it be preferable to chain them in one pull request or split each into different pull requests?

Some are quite quick and simple, so I worry about adding too many pull requests to go through.

iAmWillShepherd:
@Daniel-McCarthy doing them in separate PRs is fine, just be sure to state if one PR depends on another and you'll be good to go.

shiftkey:
@Daniel-McCarthy I opened #5703 which updated the original refactoring example to use proper jsdoc following the discussion in https://github.com/desktop/desktop/pull/5697#issuecomment-423339885 and that seems to work fine. Are you able to update this PR to do the same?

Daniel-McCarthy:
@shiftkey Certainly.
I am not sure if I have done it perfectly. Last night I set up VSCode to test this. When I used the JSDoc format for the `models/status` file it _seemed_ to be working as intended.

However, I noticed that the documentation showing up for the properties were showing the documentation from the super class, and that sub class did not seem to show any. This Branch file with the new changes seems to be behaving differently with documentation than the Status file was.

It could be behaving as expected, as I am not used to this environment.

app/src/models/branch.ts
```diff
@@ -41,31 +41,21 @@ export function eligibleForFastForward(
   )
 }
 
-/** A branch as loaded from Git. */
+/**
+ * A branch as loaded from Git.
+ *
+ * @param name The short name of the branch. E.g., `master`.
```
shiftkey: I _think_ this should be documenting the `constructor` element below, not the other `class` itself.

shiftkey: @Daniel-McCarthy friendly bump on this change - it'd be nice to close this out before we open more PRs like this

Daniel-McCarthy: Ah, of course. Nice catch! Thanks for the feedback and reminding me of this.

PR #5683, opened 2018-09-19
show resolved conflicts as resolved

### LINKED ISSUE #5609: Reflect resolved conflicts in commit view
When I hit a merge conflict, GitHub Desktop shows what files are conflicted in the commit view, but doesn't reflect that they've been resolved after I've resolved them in my text editor.

Here's the current flow: 

screen shot 2018-09-11 at 2 57 22 pm
screen shot 2018-09-11 at 2 57 40 pm
screen shot 2018-09-11 at 2 59 04 pm

I'll follow up back here with some ideas for solutions! 

cc @outofambit

outofambit:
**Goal:**
provide visual feedback on which files have had their merge conflicts resolved when mid-merge. resolves #5609 

**Implementation:**
- use `git diff --check` on files that are have a git status of `U` (unmerged) to see if any conflict markers remain
- `git diff --check` returns whether a file has any conflict markers in it
- use this along with the git status of a file to determine its AppStatus
- add some tests for detecting conflicted and resolved files
- a little new scss

**Notes:**

_i recommend using the non-whitespace only changes option on reviewing this pr, since most of the test files diff is just indenting_

**In Action:**
screen shot 2018-09-20 at 3 57 27 pm


app/src/lib/git/status.ts
```diff
@@ -35,7 +35,33 @@ export interface IStatusResult {
...
+    // we don't care about the output
+    0
+  )
+  // 2 means conflict markers were found
+  return exitCode === 2
+}
+
+function convertToAppStatus(
+  status: FileEntry,
+  hasConflictMarkers: boolean | null
```
outofambit: does it make more sense to use a single type that bundles a `FileEntry` and a `boolean` here? it seemed like overkill to me but i'm not totally sure

iAmWillShepherd: When using a `FileEntry` do you usually pass it with a `Bool`? If so, then it should probably be a type. If you only are using this `FileEntry` in fewer than 3 places, it's probably overkill.

app/src/lib/git/status.ts
```diff
@@ -35,7 +35,33 @@ export interface IStatusResult {
...
 
-function convertToAppStatus(status: FileEntry): AppFileStatus {
+async function entryHasConflictMarkers(
+  repositoryPath: string,
+  entry: IStatusEntry,
+  status: FileEntry
+): Promise<boolean | null> {
+  // only conflicted files can have conflict markers
+  if (status.kind !== 'conflicted') {
+    return null
```
iAmWillShepherd: Any reason why we can't return `false` here?

app/src/lib/git/status.ts
```diff
@@ -116,6 +144,11 @@ export async function getStatus(
   for (const entry of parsePorcelainStatus(stdout)) {
     if (entry.kind === 'entry') {
       const status = mapStatus(entry.statusCode)
+      const hasConflictMarkers = await entryHasConflictMarkers(
```
iAmWillShepherd: Awaiting in a loop like this means we can't take advantage of concurrency here. Is that the goal in this case? I know we've had issues in the past with Git processes being spawned all over the place, but is there another strategy to make sure this function doesn't take a long time to complete (imagine having 10k files that need to be processed)?

app/test/unit/git/status-test.ts
```diff
@@ -19,110 +20,144 @@ const _temp = temp.track()
...
-
-      await FSE.writeFile(path.join(repo.path, 'foo'), 'foo\n')
-
-      await GitProcess.exec(['add', 'foo'], repo.path)
-      await GitProcess.exec(['commit', '-m', 'Initial commit'], repo.path)
-      await GitProcess.exec(['mv', 'foo', 'bar'], repo.path)
+      it('parses conflicted files', async () => {
+        const status = await getStatusOrThrow(repository!)
+        const files = status.workingDirectory.files
+        expect(files.length).to.equal(1)
```
iAmWillShepherd: L37 is not needed since you have the assertion on L40.

app/test/unit/git/status-test.ts
```diff
@@ -19,110 +20,144 @@ const _temp = temp.track()
...
+          const numFiles = 10000
+          const basePath = repository!.path
+
+          await mkdir(basePath)
+
+          // create a lot of files
+          const promises = []
+          for (let i = 0; i < numFiles; i++) {
+            promises.push(
+              FSE.writeFile(
```
iAmWillShepherd: This can be simplified by using FSE.writeFileSync. Given that this is a test, it's fine if it's not async.

app/src/lib/git/status.ts
outofambit: mmm good call! i'll fix that 🔧

app/src/lib/git/status.ts
```diff
@@ -35,7 +35,33 @@ export interface IStatusResult {
...
+    return null
+  }
+
+  const args = ['diff', '--check', entry.path]
+  const { exitCode } = await spawnAndComplete(
+    args,
+    repositoryPath,
+    'diffCheck',
+    new Set([0, 2]),
+    // we don't care about the output
```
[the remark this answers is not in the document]
shiftkey: Also, I think we've talked about this before but maybe if we can run `git diff --check` (without the path argument) and parse that into a list of unique file names (`Set<string>` will help here) we could pass that into the status parser and avoid needing to execute this per-file.

app/test/unit/git/status-test.ts
outofambit: I didn't write this, I only indented it. I think its worth seeing if this runs faster on CI if done synchronously, but I think I'm leaning towards that being a separate PR/issue. @shiftkey do you have an opinion on this?

app/src/lib/git/status.ts
outofambit: i think this is good then :) thanks!

app/src/lib/git/status.ts
iAmWillShepherd: We may want to keep it this way. I only brought it up to make sure. It's never good to have 10K git processes spawned, which doing this concurrently could cause, so if you were to remove await, we'd want to limit the amount of Git processes that are spawned by employing some sort of rate limiting strategy.

cc @desktop/eng for feels.

app/src/lib/git/status.ts
outofambit: i was thinking that detecting the presence or lack of conflict markers doesn't mean anything if the file isn't `'conflicted'`, so `null` could represent that case. but i'm not sure that that information is actually useful so i'm good with `false` here. 064a919

app/src/lib/git/status.ts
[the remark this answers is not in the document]
outofambit: i'm gonna look into a single `git diff --check` version now

app/test/unit/git/status-test.ts
shiftkey: Personally I'd just do a dumb `for` loop and `await` on the inside rather than create all these promises. Or use the `Sync` API.

app/src/lib/git/diff-check.ts
```diff
@@ -0,0 +1,37 @@
...
+    args,
+    repositoryPath,
+    'getFilesWithConflictMarkers',
+    new Set([0, 2])
+  )
+
+  // result parsing
+  const outputStr = output.toString('utf8')
+  const fileNameCaptureRe = /(.+):\d+: leftover conflict marker/gi
+  return getMatches(outputStr, fileNameCaptureRe, new Set<string>())
```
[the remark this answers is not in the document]
outofambit: yeah, i couldn't find one in the codebase, but maybe i missed it? i'd be happy to move this to a helper if the pattern if there isn't one somewhere already.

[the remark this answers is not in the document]
shiftkey: I couldn't dig one up, but I tried this approach without recursion and it seems to pass the current tests:

```
function getMatches(text: string): Set<string> {
  const re = /(.+):\d+: leftover conflict marker/gi
  let match

  const matches = new Set<string>()

  do {
    match = re.exec(text)
    if (match != null && match.length === 2) {
      matches.add(match[1])
    }
  } while (match != null)

  return matches
}
```

Let me know if you prefer either approach

shiftkey:
@outofambit I'm not sure where we ended up on the UX for this feature but it was _so cool_ to see it switch from "conflicted" to "resolved" when I was playing with this earlier!

app/test/unit/git/status-test.ts
outofambit: I tried this out and it slowed the test execution by ~30% locally (~+.4 seconds) and on CircleCI (~+8 seconds). Let's not remove the promises approach.

app/src/lib/git/status.ts
```diff
@@ -187,3 +165,96 @@ export async function getStatus(
...
+  entry: IStatusEntry,
+  filesWithConflictMarkers: Set<string>
+): Map<string, WorkingDirectoryFileChange> {
+  const status = mapStatus(entry.statusCode)
+
+  if (status.kind === 'ordinary') {
+    // when a file is added in the index but then removed in the working
+    // directory, the file won't be part of the commit, so we can skip
+    // displaying this entry in the changes list
+    if (
```
iAmWillShepherd: This can be combined with the `if` statement above to reduce the nesting.

outofambit: mind if I make a followup issue to refactor getStatus more? the above line was only a whitespace change, but i'd love to dig in some more :)

outofambit: #5796

outofambit:
@iAmWillShepherd thanks for the review! can you take another look? I addressed your comments 🍕

app/src/lib/helpers/regex.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * get all regex catpures within a body of text
```
Daniel-McCarthy: This is extremely minor, but since this pull request is still open I figured it was worth mentioning this typo. 'catpures'=>'captures'

Added as a comment.

iAmWillShepherd: 🦅 👀

app/src/lib/helpers/regex.ts
```diff
@@ -0,0 +1,41 @@
...
+  return captures
+}
+
+/**
+ * get all regex matches within a body of text
+ * @param text string to search
+ * @param re regex to search with. must have global option
+ * @returns set of strings captured by supplied regex
+ */
+export async function getMatches(
```
outofambit: @iAmWillShepherd does this look better? using promises should keep us from overflowing the stack and i've moved the return value from the function arguments.

i'm trying to avoid a `while` loop since that would be blocking

app/src/lib/git/status.ts
```diff
@@ -35,7 +43,18 @@ export interface IStatusResult {
   readonly workingDirectory: WorkingDirectoryStatus
 }
 
-function convertToAppStatus(status: FileEntry): AppFileStatus {
+type StatusHeadersData = {
```
iAmWillShepherd: Let's use an interface here.

app/src/lib/git/status.ts
```diff
@@ -35,7 +43,18 @@ export interface IStatusResult {
...
+  currentBranch: string | undefined
```
iAmWillShepherd: All of these that have the type signature `x | undefined` can be made optional `x?: string`

app/src/lib/git/status.ts
```diff
@@ -103,78 +124,35 @@ export async function getStatus(
...
+    (files, entry) => buildStatusMap(files, entry, filesWithConflictMarkers),
+    new Map<string, WorkingDirectoryFileChange>()
+  )
+
+  const {
+    currentBranch,
+    currentUpstreamBranch,
+    currentTip,
+    branchAheadBehind,
+  }: StatusHeadersData = headers.reduce(parseStatusHeader, {
```
iAmWillShepherd: There's no need to add type annotation here since the compiler can infer this type.

app/src/lib/helpers/regex.ts
```diff
@@ -0,0 +1,41 @@
...
+  text: string,
+  re: RegExp
+): Promise<Array<RegExpExecArray>> {
+  const matches = new Array<RegExpExecArray>()
+  const getNextMatch = () =>
+    new Promise(resolve => {
+      const match = re.exec(text)
+      if (match !== null) {
+        matches.push(match)
+        resolve(getNextMatch())
```
iAmWillShepherd: It's unclear to me when this recurrence would end. When would we satisfy the base condition: `match === null`? `const match = re.exec(text)` doesn't seem to ever change since the text being matched never changes.

Could you move the base condition to the top?

```ts
 const getNextMatch = () =>
    new Promise(resolve => {
      const match = re.exec(text)

      if (match === null) {
        resolve(matches)
        return
      }

      matches.push(match)
      resolve(getNextMatch())
  })
```

app/src/lib/status-parser.ts
```diff
@@ -23,6 +23,18 @@ export interface IStatusEntry {
   readonly oldPath?: string
 }
 
+export function isStatusHeader(
```
iAmWillShepherd: No need for this function, we can do this inline to reduce the amount of ceremony code.

app/src/lib/status-parser.ts
```diff
@@ -23,6 +23,18 @@ export interface IStatusEntry {
...
+  statusItem: StatusItem
+): statusItem is IStatusHeader {
+  return statusItem.kind === 'header'
+}
+
+export function isStatusEntry(
```
iAmWillShepherd: No need for this function, we can do this inline to reduce the amount of ceremony code.

app/src/lib/git/status.ts
```diff
@@ -103,78 +124,35 @@ export async function getStatus(
   }
 
   const stdout = result.output.toString('utf8')
+  const parsed = parsePorcelainStatus(stdout)
+  const headers: ReadonlyArray<IStatusHeader> = parsed.filter(isStatusHeader)
```
iAmWillShepherd: This can be updated to `parsed.filter(x => x.kind === 'header')` since the `isStatusHeader` function is only checking for this condition.

app/src/lib/git/status.ts
```diff
@@ -103,78 +124,35 @@ export async function getStatus(
...
+  const entries: ReadonlyArray<IStatusEntry> = parsed.filter(isStatusEntry)
```
iAmWillShepherd: Same thing here: `parsed.filter(x => x.kind === 'entry')`

app/test/helpers/repositories.ts
```diff
@@ -116,3 +116,50 @@ export async function setupConflictedRepo(): Promise<Repository> {
...
+
+/**
+ * Setup a repository and create a merge conflict with multiple files
+ *
+ * @returns the new local repository
+ *
+ * The current branch will be 'other-branch' and the merged branch will be
+ * 'master' in your test harness.
+ *
+ * The conflicted file will be 'foo', 'bar', and 'baz'.
```
iAmWillShepherd: s/file/files

app/test/helpers/repositories.ts
```diff
@@ -116,3 +116,50 @@ export async function setupConflictedRepo(): Promise<Repository> {
...
+  Repository
+> {
+  const repo = await setupEmptyRepository()
+  const filePaths = [
+    Path.join(repo.path, 'foo'),
+    Path.join(repo.path, 'bar'),
+    Path.join(repo.path, 'baz'),
+  ]
+
+  await FSE.writeFile(filePaths[0], '')
```
iAmWillShepherd: If we use FSE.writeFileSycn, we can do away with the `async` keyword. Since this is a test, I think we are fine with doing things synchronously. That's actually what's happening anyways. I'm don't have a strong opinion either way, so I'll leave that up to you.

app/src/lib/helpers/regex.ts
iAmWillShepherd: Not really a fan of using recursion in general when iteration is typically much faster. If JS has proper tail recursion, this would be a non-issue for me. But at the moment, it seems like we trade the possibility of blowing the stack for having thousands of promises queued up.

app/test/helpers/repositories.ts
outofambit: > I tried this out and it slowed the test execution by ~30% locally (+.4 seconds) and on CircleCI (+8 seconds). Let's not remove the promises approach.

app/src/lib/status-parser.ts
outofambit: this is a [type guard](https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html#user-defined-type-guards) which i'm pretty sure can't be an inline function. i put it in this file because it felt right to keep that defined near the type it applied to. let me know if you think it should live somewhere else, but we can't remove it.

this kind of typecasting function is [necessary when `Array.filter`ing a list by type](https://github.com/Microsoft/TypeScript/pull/10027)

app/src/lib/status-parser.ts
outofambit: see https://github.com/desktop/desktop/pull/5683#discussion_r222136803

app/src/lib/git/status.ts
outofambit: see https://github.com/desktop/desktop/pull/5683#discussion_r222136803

app/src/lib/git/status.ts
outofambit: see https://github.com/desktop/desktop/pull/5683#discussion_r222136803

app/src/lib/helpers/regex.ts
outofambit: in my testing of this function, the promise chain is faster (and doesn't block the process) in comparison to a while loop. on my machine its about 20% faster

this won't create thousands of promises at once because they are created recursively in a chain. aka each promise is only created at the end of the preceding promise.

outofambit:
I've addressed all feedback. since the remaining items are really polish (and this PR has been i review for 11 days) I will be merging this tomorrow.

PR #5697, opened 2018-09-20
Refactored Status constructors with parameter properties

Daniel-McCarthy:
This is continuing off Issue #4272. It has updated the `models/Status` file with parameter properties.

This one is slightly more involved than the previous one. There may be some :art: cosmetic changes that would be preferred that I can follow for future `model` files. There are some standard constructor parameters mixed in with parameter properties. Since the parameter properties include their documentation, perhaps it might be preferable to put them in an order so that they are more uniform.

### DIFF
--- a/app/src/models/status.ts
+++ b/app/src/models/status.ts
@@ -138,12 +138,14 @@ export class FileChange {
   /** An ID for the file change. */
   public readonly id: string
 
+  /**
+   * @param path The relative path to the file in the repository.
+   * @param status The status of the change to the file.
+   * @param oldPath The original path in the case of a renamed file.
+   */
   public constructor(
-    /** the relative path to the file in the repository */
     public readonly path: string,
-    /** the status of the change to the file */
     public readonly status: AppFileStatus,
-    /** The original path in the case of a renamed file */
     public readonly oldPath?: string
   ) {
     this.id = `${this.status}+${this.path}`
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
Thanks for your contributions! They all have been stellar🥇 

I wanted to see if you were willing to try out something on our behalf? We were having a discussion about the location of comments when using parameter properties ([issue](https://github.com/desktop/desktop/pull/4243) where we decided to try them out). Basically we want to see if it's possible to come up with a solution that allows our comments to be placed above the constructor as a JSDoc, while still giving us intellisense support for when we want to access those params as properties of the class and for when we are creating a new instance of a class.

Just so you know, if you decide to tackle this and you come up with a good solution, we'd likely refactor other areas of the codebase to use the pattern you come up with. Also I'm willing to work with you on this if you want or wait to see what you come up with.

iAmWillShepherd:
Also, if it turns out JSDoc can't fulfill this need, it may be worth taking a look at [TSDoc](https://github.com/Microsoft/tsdoc).

Daniel-McCarthy:
>Thanks for your contributions! They all have been stellar1st_place_medal

That's great to hear! I definitely appreciate that and have been hoping it would be helpful.

>I wanted to see if you were willing to try out something on our behalf?

Certainly! I am definitely interesting in exploring this and seeing what can be done with JSDoc and TSDoc. I will read into this issue and see if there's something I can do to help with this matter.

iAmWillShepherd:
@Daniel-McCarthy that's great to hear ❤️ 

Feel free to ping me at any point if you get stuck or just want someone to bounce ideas off.

iAmWillShepherd:
@Daniel-McCarthy, just a heads up. @shiftkey took a stab at doing this and the results turned out OK. Take a look at https://github.com/desktop/desktop/pull/5703 and tell us what you think there (it will remain open and unmerged for 24 hours to allow the team and contributors such as yourself to comment on it).

PR #5698, opened 2018-09-20
Refactored App-Menu constructors with parameter properties

Daniel-McCarthy:
This is continuing off Issue #4272. It has updated the models/app-menu file with parameter properties.


shiftkey:
Dropping `ready-for-review` based on the feedback here: https://github.com/desktop/desktop/pull/5697#issuecomment-423339885

Once that's sorted, we can update this PR to use the same trick.

PR #5702, opened 2018-09-21
add docs about how Desktop is packaged

shiftkey:
I'd written this up a while ago and shared it with @outofambit and then completely forgot I'd done it until someone on Twitter asked me yesterday if I had these details anywhere. 

So here it is!


rahulpnath:
And that Twitter someone is me. Thanks, @shiftkey for getting it in this quick!

billygriffin:
Thanks @rahulpnath for your contribution as well in making it more discoverable!

PR #5703, opened 2018-09-21
switch from inline comments to normal jsdoc

shiftkey:
Related to #5697 and in particular this comment https://github.com/desktop/desktop/pull/5697#issuecomment-423339885.

I went back to the file used in the original PR #4243 and tested how this is inferred using VSCode (like before), and it works in the constructor:

And also works when inspecting the field outside the constructor, but still within the class:

And somewhere else in the codebase:

cc @Daniel-McCarthy

### DIFF
--- a/app/src/models/account.ts
+++ b/app/src/models/account.ts
@@ -11,20 +11,24 @@ export class Account {
     return new Account('', getDotComAPIEndpoint(), '', [], '', -1, '')
   }
 
+  /**
+   * Create an instance of an account
+   *
+   * @param login The login name for this account
+   * @param endpoint The server for this account - GitHub or a GitHub Enterprise instance
+   * @param token The access token used to perform operations on behalf of this account
+   * @param emails The current list of email addresses associated with the account
+   * @param avatarURL The profile URL to render for this account
+   * @param id The database id for this account
+   * @param name The friendly name associated with this account
+   */
   public constructor(
-    /** The login name for this account  */
     public readonly login: string,
-    /** The server for this account - GitHub or a GitHub Enterprise instance */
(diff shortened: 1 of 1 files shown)


Daniel-McCarthy:
Wonderful! So assuming there are no looming issues with this, should I the rest of my pull requests with this format? It looks great to me. It definitely resolves the :art: awkwardness of having the comments mixed in the constructor.

shiftkey:
@Daniel-McCarthy yep, I think we're good to proceed with this new approach - right @iAmWillShepherd?

Daniel-McCarthy:
Excellent. I'm taking a look at the open pull requests I have currently regarding these parameter properties. As for updating them to use the JSDoc format, what should be done with documentation that fits multiple lines?

- Let it stay on one line
```typescript
/**
 * Used by static constructors and transformers.
 *
 * @param menu The menu that this instance operates on, taken from an electron Menu instance and converted into an IMenu model by menuFromElectronMenu.
 */
```
- Let it fall into the next line with no additional spacing
```typescript
/**
 * Used by static constructors and transformers.
 *
 * @param menu The menu that this instance operates on, taken from 
 * an electron Menu instance and converted into an IMenu model by 
 * menuFromElectronMenu.
 */
```
- Space the following lines to match the start of the first
```typescript
/**
 * Used by static constructors and transformers.
 *
 * @param menu The menu that this instance operates on, taken from an
 *             electron Menu instance and converted into an IMenu model
 *             by menuFromElectronMenu.
 */
```

Hope not to be a bother, I was just hoping to get an idea ahead of time.

shiftkey:
@Daniel-McCarthy I personally like the third option

PR #5706, opened 2018-09-21
[wip] tsx unit tests

### LINKED ISSUE #5704: app-test.tsx is not currently run as part of unit test suite
We migrated over to Jest in #5427 but I noticed this line here in our current config yesterday:

https://github.com/desktop/desktop/blob/e4754a27dc6f0a3a3a1e19ba8329f14784434dc4/jest.config.js#L6

This means our `app/test/unit/app-test.tsx` is not being detected and run in jest currently.

I tried to enable it like this:

```
testMatch: ['**/unit/**/*-test.ts', '**/unit/**/*-test.tsx'], 
```

But then I noticed two runtime issues:

 - `Headers is not defined` here
https://github.com/desktop/desktop/blob/e4754a27dc6f0a3a3a1e19ba8329f14784434dc4/app/src/ui/lib/avatar-in-memory-cache.ts#L6

 - `window.isFullScreen()` was still being executed, despite [our corresponding mock](https://github.com/desktop/desktop/blob/master/app/src/lib/__mocks__/window-state.ts).

https://github.com/desktop/desktop/blob/e4754a27dc6f0a3a3a1e19ba8329f14784434dc4/app/src/lib/window-state.ts#L11-L13

cc @outofambit

outofambit:
resolves #5704

### DIFF
--- a/app/test/globals.ts
+++ b/app/test/globals.ts
@@ -5,6 +5,9 @@ const Dexie = require('dexie')
 Dexie.dependencies.indexedDB = require('fake-indexeddb')
 Dexie.dependencies.IDBKeyRange = require('fake-indexeddb/lib/FDBKeyRange')
 
+import { fetch as fetchPolyfill } from 'whatwg-fetch'
+global.fetch = fetchPolyfill
+
 // These constants are defined by Webpack at build time, but since tests aren't
 // built with Webpack we need to make sure these exist at runtime.
 const g: any = global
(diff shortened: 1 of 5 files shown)


app/test/globals.ts
```diff
@@ -5,6 +5,8 @@ const Dexie = require('dexie')
 Dexie.dependencies.indexedDB = require('fake-indexeddb')
 Dexie.dependencies.IDBKeyRange = require('fake-indexeddb/lib/FDBKeyRange')
 
+import 'airbnb-browser-shims/browser-only'
```
shiftkey: :art: maybe a comment here to indicate what how this is useful, in case someone comes in later and wants to clean it up?

outofambit: a2a1be2

PR #5707, opened 2018-09-21
extract some models from inside app-state.ts

shiftkey:
This is some refactoring to extract interfaces used in multiple distinct parts of the app - app state, Git operations, React components, etc - into separate modules under the `app/src/models` folder.

The new modules:

 - `app/src/models/clone-options.ts`
 - `app/src/models/commit-message.ts`
 - `app/src/models/diff/image-diff.ts`
 - `app/src/models/git-account.ts`
 - `app/src/models/popup.ts`
 - `app/src/models/progress.ts`
 - `app/src/models/retry-actions.ts` (previously lived under `app/src/lib` but contains no logic)

These stood out at me as things which were convenient to keep under `app/src/lib`, but having unexpected components reference back into `app/src/lib/` to import these shapes felt like they could stand on their own under `app/src/models/`.


rushjoe18:
Did not receive any call from your side

shiftkey:
:apple:

PR #5711, opened 2018-09-21
Add to max height of .commit-summary-description-scroll-view to accom…

### LINKED ISSUE #5700: Increase height of the Commit Summary area to avoid the use of the Expand button
So the height of the `commit-summary-description-container` is not optimal. Although we created the shadow to let the user know there is more info below (see here 👇) 

https://github.com/desktop/desktop/blob/3756c147716a2f26729dd790a9cba6a4033fe232/app/styles/ui/history/_commit-summary.scss#L47

I think we ought to increase the height of this area by a few pixels to avoid the following scenario.)

Before user clicks expand:
screen shot 2018-09-20 at 3 42 20 pm

After user clicks expand:
screen shot 2018-09-20 at 3 42 33 pm

Literally, I clicked the `Expand` button to remove the shadow from the sentence. The latter pic seems like a reasonable minimal height... and if there is any text below then we should show the shadow. 

cc @desktop/design 

1.4.1-beta0, mac

aryyya:
…modate 3 lines when the description is not expanded.

This fixes #5700.

When a description has 3 lines, expansion is no longer necessary:

3 line commit

If it has 4+ lines, expansion is optional and the white shadow makes it obvious:

4 line commit

Much nicer.

### DIFF
--- a/app/styles/ui/history/_commit-summary.scss
+++ b/app/styles/ui/history/_commit-summary.scss
@@ -90,7 +90,7 @@
   &-description-scroll-view {
     overflow: hidden;
     // The extra pixel makes the space align up with the commit list.
-    max-height: 61px;
+    max-height: 69px;
     flex: 1;
   }
 


outofambit:
holy wow @aryyya that was fast! thank you so much!

ampinsk:
This looks great, thanks for picking this up so quickly!

PR #5715, opened 2018-09-23
Git diff check sparingly

outofambit:
Currently, #5683 runs a `git diff --check` every time we call `getStatus()`, even when there are no conflicted files. This PR refactors `getStatus()` so we can determine if we need to `git diff --check` earlier on, and only do so when needed.

### DIFF
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -117,14 +117,6 @@ export async function getStatus(
 
   const stdout = result.output.toString('utf8')
 
-  const filesWithConflictMarkers = await getFilesWithConflictMarkers(
-    repository.path
-  )
-
-  // Map of files keyed on their paths.
-  // Note, map maintains insertion order
-  const files = new Map<string, WorkingDirectoryFileChange>()
-
   let currentBranch: string | undefined = undefined
   let currentUpstreamBranch: string | undefined = undefined
   let currentTip: string | undefined = undefined
@@ -134,44 +126,18 @@ export async function getStatus(
   const headers: ReadonlyArray<IStatusHeader> = parsed.filter(isIStatusHeader)
   const entries: ReadonlyArray<IStatusEntry> = parsed.filter(isIStatusEntry)
 
-  for (const entry of entries) {
-    const status = mapStatus(entry.statusCode)
-    const hasConflictMarkers = filesWithConflictMarkers.has(entry.path)
+  // run git diff check if anything is conflicted
+  const filesWithConflictMarkers = entries.some(
+    es => mapStatus(es.statusCode).kind === 'conflicted'
+  )
+    ? await getFilesWithConflictMarkers(repository.path)
+    : new Set<string>()
 
-    if (status.kind === 'ordinary') {
-      // when a file is added in the index but then removed in the working
-      // directory, the file won't be part of the commit, so we can skip
-      // displaying this entry in the changes list
-      if (
-        status.index === GitStatusEntry.Added &&
-        status.workingTree === GitStatusEntry.Deleted
-      ) {
-        continue
-      }
-    }
-
-    if (status.kind === 'untracked') {
-      // when a delete has been staged, but an untracked file exists with the
-      // same path, we should ensure that we only draw one entry in the
-      // changes list - see if an entry already exists for this path and
-      // remove it if found
-      files.delete(entry.path)
-    }
-
-    // for now we just poke at the existing summary
-    const summary = convertToAppStatus(status, hasConflictMarkers)
-    const selection = DiffSelection.fromInitialSelection(DiffSelectionType.All)
-
-    files.set(
-      entry.path,
-      new WorkingDirectoryFileChange(
-        entry.path,
-        summary,
-        selection,
-        entry.oldPath
-      )
-    )
-  }
+  // Map of files keyed on their paths.
+  // Note, map maintains insertion order
+  const files = entries.reduce((files, entry) => {
+    return addEntryToFiles(files, entry, filesWithConflictMarkers)
+  }, new Map<string, WorkingDirectoryFileChange>())
 
   for (const header of headers) {
     let m: RegExpMatchArray | null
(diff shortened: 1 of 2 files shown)


app/src/lib/git/status.ts
```diff
@@ -109,83 +116,41 @@ export async function getStatus(
...
+    return addEntryToFiles(files, entry, filesWithConflictMarkers)
+  }, new Map<string, WorkingDirectoryFileChange>())
+
+  const {
+    currentBranch,
+    currentUpstreamBranch,
+    currentTip,
+    branchAheadBehind,
+  }: {
+    currentBranch: string | undefined
```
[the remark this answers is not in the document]
shiftkey: You can get a fair bit of the way by just extracting out the initial value for the reducer:

```ts
  const initialState = {
    currentBranch: undefined,
    currentUpstreamBranch: undefined,
    currentTip: undefined,
    branchAheadBehind: undefined,
    m: null,
  }

  const {
    currentBranch,
    currentUpstreamBranch,
    currentTip,
    branchAheadBehind,
  } = headers.reduce(handleHeader, initialState)
```

app/src/lib/git/status.ts
```diff
@@ -109,83 +116,41 @@ export async function getStatus(
...
-        const behind = parseInt(m[2], 10)
-
-        if (!isNaN(ahead) && !isNaN(behind)) {
-          branchAheadBehind = { ahead, behind }
-        }
-      }
-    }
-  }
+  const files = entries.reduce((files, entry) => {
+    return addEntryToFiles(files, entry, filesWithConflictMarkers)
```
shiftkey: :art: some other tidy up in here 

```ts
  // Map of files keyed on their paths.
  // Note, map maintains insertion order
  const initialList = new Map<string, WorkingDirectoryFileChange>()
  const files = entries.reduce(
    (files, entry) => addEntryToFiles(files, entry, filesWithConflictMarkers),
    initialList
  )
```

app/src/lib/git/status.ts
```diff
@@ -198,3 +163,91 @@ export async function getStatus(
     workingDirectory,
   }
 }
+
+function addEntryToFiles(
```
shiftkey: :art: documentation for what this method is doing? 📝

app/src/lib/git/status.ts
```diff
@@ -198,3 +163,91 @@ export async function getStatus(
...
+      entry.path,
+      summary,
+      selection,
+      entry.oldPath
+    )
+  )
+  return files
+}
+
+function handleHeader(
```
shiftkey: :art: similarly, documentation for this function? :memo:

app/src/lib/status-parser.ts
```diff
@@ -23,6 +23,18 @@ export interface IStatusEntry {
   readonly oldPath?: string
 }
 
+export function isIStatusHeader(
```
shiftkey: :art: I'd just name this `isStatusHeader` as the `I` on the interface is just a marker

app/src/lib/status-parser.ts
```diff
@@ -23,6 +23,18 @@ export interface IStatusEntry {
...
+  statusItem: StatusItem
+): statusItem is IStatusHeader {
+  return statusItem.kind === 'header'
+}
+
+export function isIStatusEntry(
```
shiftkey: :art: same thing as above, but `isStatusEntry`

app/src/lib/git/status.ts
outofambit: 18251e8

app/src/lib/git/status.ts
outofambit: 18251e8

outofambit:
thanks @shiftkey! 🍕

app/src/lib/git/status.ts
```diff
@@ -198,3 +158,97 @@ export async function getStatus(
     workingDirectory,
   }
 }
+
+/** reducer(ish) to create the map of file change statuses
```
shiftkey: I think we can be a bit more expressive here, as `addEntryToFiles` doesn't reflect the fact that the original function did skip over some entries. Here's a quick something I put together to illustrate what I was hoping for:

```ts
/**
 *
 * Build collection of working directory changes using list of known
 * conflicted files.
 *
 * Map is used here to maintain insertion order.
 */
function parseStatusEntry(
  files: Map<string, WorkingDirectoryFileChange>,
  entry: IStatusEntry,
  filesWithConflictMarkers: Set<string>
) {
...
```

This moves the comment on line 130 closer to where it's being used.

app/src/lib/git/status.ts
```diff
@@ -198,3 +158,97 @@ export async function getStatus(
...
+ * from a list of entries
+ */
+function addEntryToFiles(
+  files: Map<string, WorkingDirectoryFileChange>,
+  entry: IStatusEntry,
+  filesWithConflictMarkers: Set<string>
+) {
+  const status = mapStatus(entry.statusCode)
+  const hasConflictMarkers = filesWithConflictMarkers.has(entry.path)
```
shiftkey: :art: Move this down closer where it's needed

app/src/lib/git/status.ts
```diff
@@ -109,83 +116,36 @@ export async function getStatus(
...
-        const behind = parseInt(m[2], 10)
-
-        if (!isNaN(ahead) && !isNaN(behind)) {
-          branchAheadBehind = { ahead, behind }
-        }
-      }
-    }
-  }
+  const files = entries.reduce(
+    (files, entry) => addEntryToFiles(files, entry, filesWithConflictMarkers),
```
shiftkey: :art: defining `files` as variables in a couple of places is probably fine from a compiler side (it usually gets noisy about shadowed variables when it hits them), but from a readability side could we change the inner `files` to something else to indicate the scope here is different?

app/src/lib/git/status.ts
```diff
@@ -198,3 +158,97 @@ export async function getStatus(
...
+      entry.path,
+      summary,
+      selection,
+      entry.oldPath
+    )
+  )
+  return files
+}
+
+/** reducer to calculate the ahead / behind and branch names
```
shiftkey: Like above, we should want the docs to describe what the function does, rather than the broader environment it's being used in.

A suggested alternative:


```ts
/**
 * Update status header based on the current header entry
 */
function parseStatusHeader(
  results: StatusBuilder,
  header: IStatusHeader
) {
```

I also extracted out `StatusBuilder` (feel free to name it something else if inspiration strikes, it's not quite the Builder pattern here) as it feels important for internal consistency:

```ts
type StatusBuilder = {
  currentBranch: string | undefined
  currentUpstreamBranch: string | undefined
  currentTip: string | undefined
  branchAheadBehind: IAheadBehind | undefined
  m: RegExpMatchArray | null
}
```

outofambit:
thanks for the feedback @shiftkey. could you please indicate which of your requested changes you consider blocking to merging this? thanks!

shiftkey:
@outofambit the ones I've marked with :art: are what I consider aesthetic (and not blockers)

PR #5301, opened 2018-07-31
Fix NDDB consistency

iAmWillShepherd:
Fixes #5293

PR #5720, opened 2018-09-24
push isDivergingBranchBannerVisible down into repository state

shiftkey:
As part of reviewing the current application state I noticed that we define `isDivergingBranchBannerVisible` as global state, but then plumb it down into the Repository and Compare components. This was different to how we track the other related state in `ICompareState`, which is tracked per-repository. This might be confusing to people unfamiliar with the code, and potentially a source of confusing behaviour that we're unpacking in #5301.

The first commit is the refactor, and the second commit is my attempt to clarify the flow while I was touching callers to this code. 

Why do that? In current `master` we had this check:

```ts
    } else if (
      inferComparisonBranch !== null ||
      aheadBehindOfInferredBranch === null
    ) {
```

`inferComparisonBranch` is actually the name of the function we're currently in, rather than some sort of variable assigned inside the function.

That won't ever be `null`, and I wasn't sure what it was supposed to be, so I tried to update the flow to reflect what I think should be happening in here.

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -190,8 +190,6 @@ export interface IAppState {
   /** The currently selected tab for the Branches foldout. */
   readonly selectedBranchesTab: BranchesTab
 
-  /** Show the diverging notification banner */
-  readonly isDivergingBranchBannerVisible: boolean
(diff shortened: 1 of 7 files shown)


iAmWillShepherd:
awesome

Issue #4385, opened 2018-04-03
upgrade emoji support to align with GitHub The Website


CarlRosell:
## Description
Some newer emojis are missing. Found this when I was trying to use the `:clown_face:`-🤡 emoji. It is available here on Github.com. Both seem to rely on https://github.com/github/gemoji and the clown_face is available in `master` but not in the commit of the submodule here.

## Version
* GitHub Desktop: 1.1.1
* Operating system: Mac, 10.13.3

## Steps to Reproduce
1. Typing `:clown_face:` in the commit message.

### Expected Behavior
Display the emoji in the auto-complete and in the commit history

### Actual Behavior
Displays `:clown_face:` as a string

shiftkey:
The problem with upgrading the submodule is that this commit no longer bundles the images that Desktop was using: https://github.com/github/gemoji/commit/16c7e21e5e2ff049c1e5affa956b120eb41fc412

The work to translate a character to the on-disk URL is being done here: 

https://github.com/desktop/desktop/blob/68e4dfbe90df4951ba2aca9a1e18b2d39e3c7f86/app/src/lib/stores/emoji-store.ts#L55-L80

We need to figure out some way to workaround this if we're going to catch up with whatever GitHub The Website is doing.

shiftkey:
For reference, I think we need to switch over to native emoji (printing the actual unicode characters instead of an image) and fallback to https://github.com/googlei18n/noto-emoji/ for platforms that don't support native emoji (Win 7, Win 8, Win 8.1 and Linux).

PR #5723, opened 2018-09-24
retire the EmojiStore component in favour of plain function

shiftkey:
Another abstraction that we don't really need, this PR swaps out creating the `EmojiStore` class for a plain `readEmoji` function, and simplifies the work to initialize the emoji in app state.

I also noted that `readEmoji` does throw exceptions on a couple of code paths, but we don't `catch` this inside `AppStore`:

  - error trying to read the `emoji.json` file from the install directory
  - error while building up the array of emoji lookup values

 This tidies that up to ensure we catch and log a message if this fails, and the app now falls back to an empty cache


app/src/lib/read-emoji.ts
```diff
@@ -0,0 +1,130 @@
...
+    const path = Path.join(rootDir, 'emoji.json')
+    Fs.readFile(path, 'utf8', (err, data) => {
+      if (err) {
+        reject(err)
+        return
+      }
+
+      const tmp = new Map<string, string>()
+
+      try {
```
outofambit: i think we can simplify this if we promisify readfile and put this logic in a `.then()`. any errors thrown in there will automatically reject the promise.

perhaps something like this?

promisified readfile:
```javascript
function promisifiedReadFile(path: string, encoding: string): Promise<string> {
  return new Promise((resolve, reject) => {
    Fs.readFile(path, 'utf8', (err, data) => {
      if (err) {
        reject(err)
      }
      resolve(data)
    })
  })
}
```

modified readEmoji:
```javascript
export function readEmoji(rootDir: string): Promise<Map<string, string>> {
  return new Promise<Map<string, string>>(async resolve => {
    const path = Path.join(rootDir, 'emoji.json')
    return resolve(
      await promisifiedReadFile(path, 'utf8').then(data => {
        const tmp = new Map<string, string>()
        const db: IGemojiDb = JSON.parse(data)
        db.forEach(emoji => {
          // Custom emoji don't have a unicode string and are instead stored
          // on disk as their first alias.
          const url = emoji.emoji
            ? getUrlFromUnicodeEmoji(emoji.emoji)
            : getEmojiImageUrlFromRelativePath(`${emoji.aliases[0]}.png`)

          if (!url) {
            log.error(`Could not calculate location of emoji: ${emoji}`)
            return
          }

          emoji.aliases.forEach(alias => {
            tmp.set(`:${alias}:`, url)
          })
        })
        const emoji = new Map<string, string>()

        // Sort and insert into actual map
        const keys = Array.from(tmp.keys()).sort()
        keys.forEach(k => {
          const value = tmp.get(k)
          if (value) {
            emoji.set(k, value)
          }
        })
        return emoji
      })
    )
  })
}
```

shiftkey: I hadn't planned to change the control flow in here, and I'd prefer to use `fs-extra` if we want to do more promisification of things, as it returns promises OOTB. But I keep thinking about #4385 and how we need to revisit our emoji usage in the app, and I'd prefer to use this PR to simplify the internals than dive any deeper currently.

PR #5728, opened 2018-09-24
Format numbers by their locale

### LINKED ISSUE #1245: Format the ahead/behind count in a locale-aware manner
screen shot 2017-04-20 at 3 03 30 pm

That should be `61,083` in 🇺🇸 and probably something else in other places.

iAmWillShepherd:
Fixes #1245

Related issue: https://github.com/electron/electron/issues/9247

I took a shot in the dark on this, but I understand that using the `toLocaleString` function may not be sufficient, so it would be great if we can discuss the reasons why it's not sufficient or better approaches to formatting large numbers.

### Before

screen shot 2018-09-24 at 9 09 34 pm

### After

screen shot 2018-09-24 at 3 08 41 pm

cc @desktop/engineering

### DIFF
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -105,7 +105,7 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     if (ahead > 0) {
       content.push(
         <span key="ahead">
-          {ahead}
+          {ahead.toLocaleString()}
           <Octicon symbol={OcticonSymbol.arrowSmallUp} />
         </span>
       )
@@ -114,7 +114,7 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     if (behind > 0) {
       content.push(
         <span key="behind">
-          {behind}
+          {behind.toLocaleString()}
           <Octicon symbol={OcticonSymbol.arrowSmallDown} />
         </span>
       )


shiftkey:
For reference, we talked about this last in #1257 and I think there were two themes about why we left it as-is:

 - I can't recall how macOS does this, but on Windows you can choose a different locale for formatting numbers/dates/currency, and Desktop doesn't reflect that when using locale-based formatting in Electron
 - this is the only place where we show a number that might need formatting, so it wasn't a thing that most users noticed

The work required to understand the domain well enough, to then write a native module to read the locale info from the OS-specific APIs, and finally get enough test coverage of what users might do wasn't quite worth it then. 

Maybe if we get more interest in proper date formatting (I'm not sure what Momentjs does here, but I suspect it's limited by browser support too) this could become a more practical investment?

outofambit (dismissed):
I don't see the harm in doing this. It may not work for all configurations, but its at worst status quo.

niik:
> I don't see the harm in doing this. It may not work for all configurations, but its at worst status quo.

As @shiftkey said we decided against this approach in https://github.com/desktop/desktop/pull/1257 and I'm not sure what's changed since then that would make us reconsider?

If we can get proper number formatting that takes into account the number formatting locale I'd be super happy to merge but unless something has changed that I'm missing here I don't think we have that information at our disposal at the moment.

> but its at worst status quo.

Could you elaborate on this? For me it would render the example in the PR body as `6,000` where ',' is the thousand separator used in the US whereas for Swedish it should be a space. Comma is the decimal separator in Swedish.

My locale settings

screen shot 2018-10-02 at 10 25 53

screen shot 2018-10-02 at 10 26 09

What this PR proposes

screen shot 2018-10-02 at 10 26 21

What it should be

screen shot 2018-10-02 at 10 27 35

ghost (approved):

outofambit:
thanks @niik that helps a lot to clarify the problem. @iAmWillShepherd do you have thoughts?

outofambit:
for what its worth, the following also works

```
(1234.56).toLocaleString("en-SE")
"1 234,56"
```

iAmWillShepherd:
@outofambit this hasn't been address in Electron or Chromium(?), so there's not much we can do at the moment. I'm going to close this out until a fix is implemented in our dependencies.

mafalamanero (commented):
Recovery codex

PR #5729, opened 2018-09-24
add merge aborting functionality

outofambit:
adds the ability to abort an in-progress, conflicted merge, but doesn't hook this up to any UI

- [x] write happy path merge function
- [x] land https://github.com/desktop/dugite/pull/218
- [x] land #5757
- [x] handle `NoMergeToAbort` error

_Would like to merge this Wed, Oct 10._


outofambit:
(Azure pipelines failed on an unrelated test timeout)

app/src/lib/git/merge.ts
```diff
@@ -80,3 +80,10 @@ export async function mergeTree(
 
   return parseMergeResult(output)
 }
+
+/** Abort a mid-flight (conflicted) merge
+ *  @param repository where to abort the merge
+ */
+export async function abortMerge(repository: Repository): Promise<IGitResult> {
```
iAmWillShepherd: Could you explain the decision to return an `IGitResult` over a `MergeResult`? Most of the git function I've come across get an `IGitResult` back from Dugite and are then parsed to return a type we use in the UI or store.

outofambit: mmmm, that's a good question. I'm honestly not sure what makes the most sense for this function to return. I don't think `IGitResult` will never have any useful information since we're not catching...perhaps it should just return `Promise<null>`.

does that make sense? the only other thing I think it could return is `true` but it would never return anything else.

outofambit: oh sorry i forgot to say that i didn't use `MergeResult` because this aborts a merge and none of the `MergeResult` types make sense for that case. I guess another option is to create a `MergeAbortResult` type and have it be either `Successful` or `Errored`.

I'm thinking `abortMerge` will be used rarely and only fail when the user has done something outside of github desktop so I was thinking it should let errors propagate up.

iAmWillShepherd: `MergeResult` is a union type, so you would be able to extend it with a `MergeAborted` type.

outofambit:
thanks @iAmWillShepherd !

app/test/unit/git/merge-test.ts
```diff
@@ -82,7 +81,29 @@ describe('git/merge', () => {
...
+          /There is no merge in progress, so there is nothing to abort/
+        )
+      })
+    })
+    describe('in the middle of resolving conflicts merge', () => {
+      beforeEach(async () => {
+        repository = await setupConflictedRepo()
+      })
+      it('aborts the merge', async () => {
+        await expect(subject()).resolves.not.toThrow()
```
shiftkey: A suggestion for the next time you're in here - what about also verifying that the Git repository is no longer in a conflicted state?

saidur48:
i have a nvadea gpu .i want to mining zcoin or bitcoin.are you help me.how will i start?

PR #5742, opened 2018-09-25
Adds stats for when a merge is completed/aborted after a merge conflict

### LINKED ISSUE #5394: [Metrics] Merge conflict resolution

iAmWillShepherd:
Fixes #5394

At the moment, all the functions are in place to record the updates, but @outofambit and I were having some trouble getting to the right state. We'd like to be able to inspect the previous state so we can compare against the new state, so we did some programming by wishful thinking and just made up some variables that stand in for the real deal.


app/src/lib/stores/app-store.ts
```diff
@@ -1470,6 +1471,31 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return false
     }
 
+    // were any of these conflicted
+    const inAConflictedMerge: boolean = [...status.workingDirectory.files].some(
+      file => file.status === AppFileStatus.Resolved || file.status === AppFileStatus.Conflicted
+    )
+    const prevouslyInAConflictedMerge = false
```
iAmWillShepherd: The start of our wishful thinking. We've encoded what we'd like to inspect in the comments

shiftkey:
I've merged #5715 into the base branch for this PR, and it looks like some commits here overlap with that - are we able to tidy up this PR to focus on just the new changes?

iAmWillShepherd:
Closed in favor of #5810

Issue #5690, opened 2018-09-20
Confusion between "Discard all changes" and "Discard changes" leading to unintentionally discarding all


bill1234567890abc:
Just now I wanted to discard one local file change. I opened the right mouse click and chose discard changes. Unfortunately I chose "discard all" because "discard one" is close to "discard all". I didn't have a careful look at confirmation dialog and confirmed the action. Then all work I have done today was discarded.

Thanks God. I restored them from recycle bin on windows. I guess I'm not the first that discard changes wrongly. 

My complain is that the two discard changes are too close and easily choose by mistake.

My idea is to separate them within an apparent distance.

Thanks,
Bill

steveward:
Thanks for the feedback about this @bill1234567890abc. When you select `Discard all changes` a confirmation dialog appears that explicitly states the number of files that will have discarded changes, which should help reduce cases where all changes are accidentally discarded. However I can understand how this could be missed if you are assuming you clicked the correct option.

Would changing the title of the `Discard Changes` confirmation button to `Discard All Changes` be helpful in this scenario?

Daniel-McCarthy:
I can certainly see the potential confusion here. I definitely see where changing the dialog title and/or button text can be useful.

Here are two possible iterations.

- Change dialog title & confirm button text.
![image](https://user-images.githubusercontent.com/4957200/46044338-1c9c1b00-c0e0-11e8-841b-1b1e321ff4a4.png)

- Change dialog title solely.
![image](https://user-images.githubusercontent.com/4957200/46044408-4b19f600-c0e0-11e8-81b1-75dcd9668cf9.png)

I think that at a bare minimum the dialog title should be changed as it is the largest and most apparent change. However, the button gives the user a final chance to notice.

graemeworthy:
This has happened a few times, and the dialog doesn't seem to stop it from happening.

It seems like the context dropdown is the wrong place for this item.

Right clicking a file is for actions related to that file
Lets look at the rest of the items in that list 

|Action| Context|
|---|----|
|Discard Changes...| This file... destructive|
|Discard All Changes...| All Files... destructive|
|Ignore File| This file|
|Ignore All .js files | Project wide, using this file as an example|
|Reveal in finder | This file|
|Open in Atom| This file|
|Open with Default Program| This File|

steveward:
[...]

PR #5744, opened 2018-09-26
Update Discard-Changes-Dialog text when discarding all changes

Daniel-McCarthy:
This is in response to Issue #5690. It recommends updating the `Discard-Changes-Dialog` title and confirmation button when discarding all changes. This is to notify the user more clearly for situations where they have accidentally selected "Discard All Changes..." instead of "Discard Change..." or "Discard # Selected Changes..."
___
This is what the dialog would look like when clicking "Discard All Changes...":
image
___
In regards to the implementation of this change, I suspect there may be a simpler way to inform the dialog when the user is attempting to discard all changes. To do this, a boolean was passed to the PopupType so that it could be sent to the dialog via props.

- As an extra note, selecting "Discard Change" or selecting multiple modified files (but not all of them) to discard, the dialog will display as it normally does. However, if one selects all of the modified files and uses "Discard # Selected Files" it will display the Discard All Changes text.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1018,6 +1018,10 @@ export class App extends React.Component<IAppProps, IAppState> {
           popup.showDiscardChangesSetting === undefined
             ? true
             : popup.showDiscardChangesSetting
+        const discardingAllChanges =
+          popup.discardingAllChanges === undefined
+            ? false
+            : popup.discardingAllChanges
 
         return (
           <DiscardChanges
@@ -1029,6 +1033,7 @@ export class App extends React.Component<IAppProps, IAppState> {
               this.state.askForConfirmationOnDiscardChanges
             }
             showDiscardChangesSetting={showSetting}
+            discardingAllChanges={discardingAllChanges}
             onDismissed={this.onPopupDismissed}
             onConfirmDiscardChangesChanged={this.onConfirmDiscardChangesChanged}
           />
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -50,7 +50,7 @@ interface IChangesListProps {
   readonly askForConfirmationOnDiscardChanges: boolean
   readonly onDiscardAllChanges: (
     files: ReadonlyArray<WorkingDirectoryFileChange>,
-    dontDisplayAsAll?: boolean
+    isDiscardingAllChanges?: boolean
   ) => void
 
   /**
@@ -218,15 +218,12 @@ export class ChangesList extends React.Component<
       })
 
       if (modifiedFiles.length > 0) {
-        if (modifiedFiles.length === workingDirectory.files.length) {
-          // Display Discard Changes popup as Display All Changes since all
-          // modified files are selected.
-          this.props.onDiscardAllChanges(modifiedFiles)
-        } else {
-          // Display Discard Changes popup without displaying showing as
-          // Discard All Changes.
-          this.props.onDiscardAllChanges(modifiedFiles, true)
-        }
+        // DiscardAllChanges can also be used for discarding several selected changes.
+        // Therefore, we update the pop up to reflect whether or not it is "all" changes.
+        const discardingAllChanges =
+          modifiedFiles.length === workingDirectory.files.length
+
+        this.props.onDiscardAllChanges(modifiedFiles, discardingAllChanges)
       }
     }
   }
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -49,7 +49,8 @@ interface IChangesListProps {
   readonly onDiscardChanges: (file: WorkingDirectoryFileChange) => void
   readonly askForConfirmationOnDiscardChanges: boolean
   readonly onDiscardAllChanges: (
-    files: ReadonlyArray<WorkingDirectoryFileChange>
+    files: ReadonlyArray<WorkingDirectoryFileChange>,
+    dontDisplayAsAll?: boolean
   ) => void
 
   /**
@@ -217,7 +218,15 @@ export class ChangesList extends React.Component<
       })
 
       if (modifiedFiles.length > 0) {
-        this.props.onDiscardAllChanges(modifiedFiles)
+        if (modifiedFiles.length === workingDirectory.files.length) {
+          // Display Discard Changes popup as Display All Changes since all
+          // modified files are selected.
+          this.props.onDiscardAllChanges(modifiedFiles)
+        } else {
+          // Display Discard Changes popup without displaying showing as
+          // Discard All Changes.
+          this.props.onDiscardAllChanges(modifiedFiles, true)
+        }
       }
     }
   }
--- a/app/src/ui/changes/sidebar.tsx
+++ b/app/src/ui/changes/sidebar.tsx
@@ -168,12 +168,16 @@ export class ChangesSidebar extends React.Component<IChangesSidebarProps, {}> {
   }
 
   private onDiscardAllChanges = (
-    files: ReadonlyArray<WorkingDirectoryFileChange>
+    files: ReadonlyArray<WorkingDirectoryFileChange>,
+    dontDisplayAsAll?: boolean
   ) => {
+    const discardingAllChanges =
+      dontDisplayAsAll === undefined ? true : dontDisplayAsAll === false
     this.props.dispatcher.showPopup({
       type: PopupType.ConfirmDiscardChanges,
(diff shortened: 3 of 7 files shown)


app/src/ui/changes/sidebar.tsx
```diff
@@ -168,12 +168,16 @@ export class ChangesSidebar extends React.Component<IChangesSidebarProps, {}> {
   }
 
   private onDiscardAllChanges = (
-    files: ReadonlyArray<WorkingDirectoryFileChange>
+    files: ReadonlyArray<WorkingDirectoryFileChange>,
+    dontDisplayAsAll?: boolean
```
outofambit: what do you think about using a default value here instead of an optional? that way we can avoid checking for `undefined` in the ternary

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -57,11 +58,19 @@ export class DiscardChanges extends React.Component<
   }
 
   public render() {
+    const discardingAllChanges = this.props.discardingAllChanges
+
     return (
       <Dialog
         id="discard-changes"
         title={
-          __DARWIN__ ? 'Confirm Discard Changes' : 'Confirm discard changes'
```
outofambit: i looked and the components in this codebase are littered with this logic. would you be up for creating a helper function that takes a string and returns the platform appropriate casing of it? and using it here? (no need to replace all instances of that pattern in the codebase—just having the helper function for future use would be amazing!)

app/src/ui/changes/changes-list.tsx
```diff
@@ -49,7 +49,8 @@ interface IChangesListProps {
   readonly onDiscardChanges: (file: WorkingDirectoryFileChange) => void
   readonly askForConfirmationOnDiscardChanges: boolean
   readonly onDiscardAllChanges: (
-    files: ReadonlyArray<WorkingDirectoryFileChange>
+    files: ReadonlyArray<WorkingDirectoryFileChange>,
+    dontDisplayAsAll?: boolean
```
outofambit: blocking: i'm having a hard time reasoning about this variable as its named. could we rename this to be a positively phrased name? like `displayAsAll` or `isDiscardAll` or something like that?

app/src/ui/app.tsx
```diff
@@ -1018,6 +1018,10 @@ export class App extends React.Component<IAppProps, IAppState> {
           popup.showDiscardChangesSetting === undefined
             ? true
             : popup.showDiscardChangesSetting
+        const discardingAllChanges =
```
outofambit: perhaps this should have a default value too?

app/src/ui/changes/changes-list.tsx
```diff
@@ -217,7 +218,15 @@ export class ChangesList extends React.Component<
       })
 
       if (modifiedFiles.length > 0) {
-        this.props.onDiscardAllChanges(modifiedFiles)
+        if (modifiedFiles.length === workingDirectory.files.length) {
```
outofambit: 💯

app/src/ui/changes/sidebar.tsx
Daniel-McCarthy: Sounds great! I definitely like that suggestion better.

app/src/ui/changes/changes-list.tsx
Daniel-McCarthy: Great point! How about `isDiscardingAllChanges`?
It's a little bulky, but I think the clarity would be worth it.

app/src/ui/discard-changes/discard-changes-dialog.tsx
outofambit: @Daniel-McCarthy exactly! so something like function that does: 
`string` -> `string in the current platform's casing`

app/src/ui/app.tsx
Daniel-McCarthy: Unfortunately, after looking into it, this one may not be possible. The variable this one depends on is defined in the `app-state.ts` and when I switch it to the `discardingAllChanges: boolean = true` style it warns me that an initializer like this is only possible in a constructor or function implementation.

I will implement the other changes for now as it does not seem I can progress on this one.

outofambit: oops must have missed some context on this one!

app/src/ui/discard-changes/discard-changes-dialog.tsx
Daniel-McCarthy: Okay, I got a chance to look into this. I have a solution that can emulate what I believe we're going for.
```typescript
  /** Convert input string to current platform's text case preference. */
  private toPlatformCase(inputText: string) {
    inputText = inputText.toLowerCase()

    if (__DARWIN__) {
      // Capitalize the first letter of every word.
      inputText = inputText.replace(/\b[a-z]/gi, $1 => $1.toUpperCase())
    } else {
      // Capitalize the first letter of the first word.
      inputText = inputText.replace(/\b[a-z]/i, $1 => $1.toUpperCase())
    }

    return inputText
  }
```

I haven't made this into a commit yet. Additionally, I was not sure where to place it. I initially placed it in `discard-changes-dialog.tsx` near where it's needed for testing purposes. However, if we want this as a global helper function, I was not exactly confident where it should go.

A somewhat similar bit of code I have bumped into before was the [`src/lib/sanitize-branch.ts`](https://github.com/desktop/desktop/blob/master/app/src/lib/sanitize-branch.ts) which gets exported out. If we were to place it in `/src/lib/` I could additionally add a test for this function.

But that's definitely jumping ahead a bit as there may be changes needed, and decide if we'll continue with this function.

outofambit: this solution looks great to me! and i agree /src/lib/ is a good place for it, probably in its own file for now. and a test would be amazing! full steam ahead :)

outofambit: Thanks for these excellent notes!

> A test for this comes with some additional difficulties

I forgot about that. :) What i've done in the past for is something like this is:

```typescript
    it('does something good', () => {
      if (__WIN32__) {
        // Sentence case expect
      }
      if (__DARWIN__ || __LINUX__) {
        // Title case expect
      }
    })
```

Not the prettiest solution, so if you have another approach I'm all ears!

> It appears that @j-f1 has worked on a much better solution for this problem

Thanks for finding this! I didn't know it existed. There's a lot of good points about limitations and while I think that the solution there is an excellent comprehensive solution, I would say this is still a good incremental step to cover the majority of cases. I'd rather get this in in the meantime.

Thank you again for all your work on this @Daniel-McCarthy! I just wrote a component today that I'm planning to use this new helper in (twice!).

Daniel-McCarthy: Excellent, I have implemented the test as suggested by combining the platform test into one. 

>I'd rather get this in in the meantime.

Sounds great, I will have this ready very soon and update it if there any problems with building.

>Thank you again for all your work on this

No problem, I'm happy to help! I'm glad it will be useful.

iAmWillShepherd:
@outofambit @Daniel-McCarthy 

Sorry to jump into this so late, I just realized this change was made. The function `toPlatformCase` will not work because of reasons I outlined in my review of #5841. Because of that, we need to revert the commit that added the function and go back to using string literals. While that approach feels like doing extra work, the benefits of being able to easily read what will appear in the application beats the decreased work of ensuring the string literal is correct and the brittleness of covering all cases.

Daniel-McCarthy:
@iAmWillShepherd Perfectly understandable. Should I start a pull request to revert that particular commit, or is that already being covered?

iAmWillShepherd:
@outofambit 

>  I'm not convinced that this isn't a worthwhile helper, even if it doesn't cover all cases.

The function currently doesn't do what it states: convert a string into title case. Check [this](https://github.com/desktop/desktop/pull/5841#pullrequestreview-162478757) out for further discussion. If we can't use this function everywhere, what do we gain? I'd prefer we do things consistently versus use a function that doesn't limit the amount of actual work we have to do. We still have to pass the string to the function, so we don't gain anything really, but we do lose the ability to have string literals that match what's seen in the UI. We also lose the ability to know what a string will be displayed as without having to actually run the application and verify manually. 

If I've missed any benefits this function gives us let me know, I admit I only looked at this briefly.

outofambit:
@iAmWillShepherd can we please continue discussion in #5847? I think this is a good discussion, I'd just like to keep it in one place going forward!

iAmWillShepherd:
Yea, my bad, didn't realize you created another issue for it! 👍

Issue #5743, opened 2018-09-26
Lack of import from old app remains barrier to entry


bdw429s:
I've used Github for Desktop for years now and installed the new version back when it first came out but was promptly disappointed when I realised I hadn't upgraded the app but merely installed a second software which contained none of my repos.  I set it aside and figured that if I waited a while, surely a proper import would be implemented.  I have hundreds of repos spread around my computer and have no interest in setting all that up again from scratch.

I finally got a chance to check back into this and was again disappointed to see that not only does a proper import from the old version of Desktop not exist, it has been asked for several times and either met with the instructions to simply re-create the setup from scratch, or the tickets left until a bot came along and closed them!  

Here are all the existing tickets I found.  They are all closed with no resolution, thus this new ticket.

https://github.com/desktop/desktop/issues/4600
https://github.com/desktop/desktop/issues/186
https://github.com/desktop/desktop/issues/1672
https://github.com/desktop/desktop/issues/1752
https://github.com/desktop/desktop/issues/2220
https://github.com/desktop/desktop/issues/4869

I really appreciate all the use I've gotten out of this (free) app over the years but find it surprising that a proper migration was never considered seriously, and even actively rejected by some members.  I appreciate that underlying technologies have changed from your perspective, but from my perspective as a user, I really don't care :)  This is just another upgrade like any other.  Like going from Chrome 68 to Chrome 69.  I don't want the old version around and I expect all my data to be in the new version.  If you changed underlying technologies I'm more than happy for you, but don't push the baggage onto me please.

I read through the other threads including one that had researching into what it would take to perform a seamless migration.  Surely this new version didn't sneak up on everyone, though.  I assume development was ongoing for months and at some point in time, the later releases of the old architecture could have easily exposed their data in a way readable by the upgrade.  How is it this caught everyone off guard?

If an automatic-migration is not going to be considered, please be upfront and state it here, don't just keep referring to other closed tickets like what happened on all the other threads.  I'm not a huge fan of your auto-close bot, who assumes any ticket not replied to must be resolved so I aim to keep this ticket alive until an answer can be reached one way or another lest others stumble across it, auto-closed in a few months, with no discernable resolution. 

Thank you for the considerations and I'm looking forward to finally having a seamless upgrade path from the latest version of this great tool.

j-f1:
[...]

PR #5747, opened 2018-09-26
disable Lock integration

shiftkey:
This was mentioned out in #5743 and indicates that the [Lock](https://github.com/apps/lock) GitHub App isn't doing what it's supposed to be:

>  I'm not a huge fan of your auto-close bot, who assumes any ticket not replied to must be resolved so I aim to keep this ticket alive until an answer can be reached one way or another lest others stumble across it, auto-closed in a few months, with no discernable resolution.

This isn't the first time I've heard this feedback, and I'm in agreement that this can lead to a bad first impression for someone new to the project.

I'm going to put together a script to unlock and unlabel issues/PRs that were caught up in this, but I'm also keen to hear the feedback from others about this bit of automation. I've already uninstalled the GitHub App to prevent any further issues being marked like this.

TODO:

 - [x] unlock issues  marked with `locked-due-to-inactivity`
 - [x] unlabel issues marked with `locked-due-to-inactivity`
 - [x] remove `locked-due-to-inactivity` label

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,7 +1,0 @@
-# Configuration for lock-threads - https://github.com/dessant/lock-threads
-daysUntilLock: 60
-lockLabel: locked-due-to-inactivity
-lockComment: false
-setLockReason: true
-# Proposals need to be handled separately to normal issues
-exemptLabels: ['future-proposal']


shiftkey:
Cleanup work done, gonna #jolo this in to cleanup the old config.

PR #5757, opened 2018-09-26
update embedded Git to latest version

shiftkey:
This upgrades us to Git 2.19.0 and Git LFS 2.5.2, and also adds the new error code from https://github.com/desktop/dugite/pull/218

I'm open to better suggestions for the new user-friendly error message, but I mostly wanted to open this early to catch any unsuspecting Git changes.

### DIFF
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -255,6 +255,8 @@ function getDescriptionForError(error: DugiteError): string {
       return 'This path is not a valid path inside the repository.'
     case DugiteError.LockFileAlreadyExists:
       return 'A lock file already exists in the repository, which blocks this operation from completing.'
+    case DugiteError.NoMergeToAbort:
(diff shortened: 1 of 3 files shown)


app/src/lib/git/core.ts
```diff
@@ -255,6 +255,8 @@ function getDescriptionForError(error: DugiteError): string {
       return 'This path is not a valid path inside the repository.'
     case DugiteError.LockFileAlreadyExists:
       return 'A lock file already exists in the repository, which blocks this operation from completing.'
+    case DugiteError.NoMergeToAbort:
+      return 'There is no merge operation to abort.'
```
outofambit: maybe something like "There's no merge in-progress, so we there's nothing to cancel."

What do people think about the use of **cancel** merge over **abort** merge?

outofambit: cc @desktop/design

PR #5769, opened 2018-09-27
upgrade to typescript 3.1

outofambit:
- [x] bump dependencies
- [x] fix linter parsing
- [x] get tests passing
- [x] fix build errors


outofambit:
@j-f1 do you know of a good way to check for that pattern?

outofambit:
Some output from `yarn lint`:
```shellsession
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

SUPPORTED TYPESCRIPT VERSIONS: ~3.0.1

YOUR TYPESCRIPT VERSION: 3.1.1

=============

```

https://github.com/JamesHenry/typescript-estree doesn't yet support TypeScript 3.1, which is used by https://github.com/eslint/typescript-eslint-parser.

I'm afraid this is blocked until `typescript-estree` supports TS 3.1. Opened this PR to kickstart the process: https://github.com/JamesHenry/typescript-estree/pull/14

j-f1:
> do you know of a good way to check for that pattern?

Unfortunately, I don’t know of a way.

outofambit:
now waiting on https://github.com/eslint/typescript-eslint-parser/issues/524

app/src/ui/app.tsx
```diff
@@ -611,11 +614,13 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   public componentDidMount() {
-    document.ondragover = e => {
-      if (this.isShowingModal) {
-        e.dataTransfer.dropEffect = 'none'
-      } else {
-        e.dataTransfer.dropEffect = 'copy'
+    document!.ondragover = e => {
+      if (e.dataTransfer !== null) {
```
j-f1: `!= null`

app/src/ui/app.tsx
```diff
@@ -629,8 +634,10 @@ export class App extends React.Component<IAppProps, IAppState> {
       if (this.isShowingModal) {
         return
       }
-      const files = e.dataTransfer.files
-      this.handleDragAndDrop(files)
+      if (e.dataTransfer !== null) {
```
outofambit: @j-f1 should this be `!=` too?

j-f1: Yes.

app/src/ui/app.tsx
```diff
@@ -347,7 +347,10 @@ export class App extends React.Component<IAppProps, IAppState> {
       cancelable: true,
     })
 
-    if (document.activeElement.dispatchEvent(event)) {
+    if (
+      document!.activeElement &&
```
shiftkey: We only need to `!` the elements when they might be `null` and we know they are non-null. I think `document` is fine here, so this can become:

```ts
if (
  document.activeElement &&
  document.activeElement.dispatchEvent(event)
) {
  remote.getCurrentWebContents().selectAll()
}
```

app/src/ui/app.tsx
```diff
@@ -611,11 +614,13 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   public componentDidMount() {
-    document.ondragover = e => {
-      if (this.isShowingModal) {
-        e.dataTransfer.dropEffect = 'none'
-      } else {
-        e.dataTransfer.dropEffect = 'copy'
+    document!.ondragover = e => {
```
shiftkey: Same thing as before, no need to `!` here for `document`.

app/src/ui/autocompletion/autocompleting-text-input.tsx
```diff
@@ -155,7 +155,7 @@ export abstract class AutocompletingTextInput<
       : -1
     const rect = element.getBoundingClientRect()
     const popupAbsoluteTop = rect.top + coordinates.top
-    const windowHeight = element.ownerDocument.defaultView.innerHeight
+    const windowHeight = element.ownerDocument!.defaultView!.innerHeight
```
shiftkey: :art: How do you feel here about being defensive with these checks, rather than `!` to ignore the warnings?

```ts
if (element.ownerDocument == null) {
  return null
}

if (element.ownerDocument.defaultView == null) {
  return null
}

const windowHeight = element.ownerDocument.defaultView.innerHeight
```

I suggest this because in this method we can return `null` if we're not in a known good state, rather than continuing on and raising an error because we try to read something that isn't defined.

app/src/ui/app.tsx
outofambit: ah yes, we can. i mistakenly thought `document` needed to be `!`, but i checked again and that is not the case. sorry about that!

app/src/ui/autocompletion/autocompleting-text-input.tsx
outofambit: ah ok. yeah that's a good point. how about we go a step further and do something like b9b04d1?

shiftkey: Works for me 👍

PR #5783, opened 2018-09-29
Clicking History tab should default to commit history list

### LINKED ISSUE #5736: Clicking History tab should default to commit history list
If you click into the filter under the History tab but do not enter any text, you appear to be stuck looking at compare branches unless you know that the `esc` key takes you back to the commit history.

If the user enters any text, the `x` icon is shown as expected, thus allowing the user to escape back to the commit history. However if one was to click into the filter, enter nothing, and then click on the History tab, nothing happens. A bit of an edge case, but an annoyance as it is not clear how to "reset" the History tab, and it takes a slight workaround to get back to the commit history.

Repro:
- click into filter on History tab, do not enter any text
- click History tab, (notice the cursor leaves the filter area)

Expected result: commit history is shown again, not compare branch list
Actual result: Nothing

screen shot 2018-09-25 at 8 46 01 am

Should this enhancement seem reasonable, please assign the `help-wanted` label.

cc @desktop/design 

1.4.1-beta0, mac

JKirkYuan:
Sets showBranchList to false when clicking on a new tab. This should resolve #5736

### DIFF
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -329,5 +329,9 @@ export class RepositoryView extends React.Component<
       this.props.repository,
       section
     )
+
+    this.props.dispatcher.updateCompareForm(this.props.repository, {
+      showBranchList: false,
+    })
   }
 }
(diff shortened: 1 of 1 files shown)


JKirkYuan:
My pleasure @outofambit 😄. Looking forward to helping out more in the future!

PR #5797, opened 2018-10-02
consolidated assets

prophit987:
confirmation notifications_true


say25:
Hey @prophit987 I think this work might have been done on the wrong repo?

PR #5814, opened 2018-10-04
Refactored models/diff/raw-diff constructors with parameter properties

Daniel-McCarthy:
This is continuing off Issue #4272. It has updated the `models/diff/raw-diff` file with parameter properties.

One thing I noticed with the second constructor in the `raw-diff` file is that it contains an optional argument (`sectionHeading?: string | null`) that does not appear to be used. I'm not sure if this was accidentally left in or if it will be used at a later point. I have not made any changes to that, but thought it was mentioning since we're taking a look at this file.


outofambit:
@Daniel-McCarthy, i'm in favor of removing the unused `null` typing. just tried it out locally and everything type-checked fine. @desktop/engineering @desktop/comrades any thoughts?

Daniel-McCarthy:
@outofambit 
> i'm in favor of removing the unused `null` typing

Just to be clear, when you say "the unused `null` typing" are you suggesting switching this optional `sectionHeading` argument to `string` from `string | null`?
I was initially suggesting removing the argument as a whole since it does not seem to be used in the constructor at all.

I just want to make sure we're on the same page.

outofambit:
@Daniel-McCarthy oh! i totally misunderstood. let me look at that :)

outofambit:
> I was initially suggesting removing the argument as a whole since it does not seem to be used in the constructor at all.

@Daniel-McCarthy ok, yes, I am in favor of removing `sectionHeading` entirely :) thanks for double checking!!

Daniel-McCarthy:
Great, I will make that change now and update this very soon.

>thanks for double checking!!

No problem, glad we're on the same page!

outofambit:
looking into why appveyor failed... 🤔

PR #5815, opened 2018-10-04
Refactored models/diff/diff-selection constructors with parameter properties

Daniel-McCarthy:
This is continuing off Issue #4272. It has updated the `models/diff/diff-line` file with parameter properties.

Along with the `models/diff/raw-diff` pull requests this should be the last one.

### DIFF
--- a/app/src/models/diff/diff-selection.ts
+++ b/app/src/models/diff/diff-selection.ts
@@ -51,15 +51,11 @@ function typeMatchesSelection(
  * whose selection state has diverged from the default selection state.
  */
 export class DiffSelection {
-  private readonly defaultSelectionType:
-    | DiffSelectionType.All
-    | DiffSelectionType.None
-
   /* Any line numbers where the selection differs from the default state. */
-  private readonly divergingLines: Set<number> | null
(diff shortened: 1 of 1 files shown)


app/src/models/diff/diff-selection.ts
```diff
@@ -51,15 +51,11 @@ function typeMatchesSelection(
...
-    | DiffSelectionType.All
-    | DiffSelectionType.None
-
   /* Any line numbers where the selection differs from the default state. */
-  private readonly divergingLines: Set<number> | null
+  //private readonly divergingLines: Set<number> | null
 
   /* Optional set of line numbers which can be selected. */
-  private readonly selectableLines: Set<number> | null
+  //private readonly selectableLines: Set<number> | null
```
j-f1: How come these are commented out instead of removed?

Daniel-McCarthy: Oh, thanks! Great catch!

It seems that I had forgotten to remove these and move those comments down after testing that the constructor itself was working. I have updated to fix this!

outofambit: thank you both!

PR #5816, opened 2018-10-04
Fixed auth error prompt when cloning repo from nonexistent URL

### LINKED ISSUE #5661: Auth error prompt when cloning repo from nonexistent URL
## Description

On trying to clone a repo using an incorrect URL, Desktop raises an authentication error instead of indicating that the repo wasn't found in the user's profile. 

## Version

* GitHub Desktop: 1.3.5

* Operating system: macOS High Sierra 10.13.6

## Steps to Reproduce

1. Click on Clone repository. 
2. Navigate to the URL section. 
3. Enter an invalid URL. 

### Expected Behavior

Desktop raises an error prompt stating that the provided repo URL is invalid. 

### Actual Behavior

Desktop goes into a loop of error prompts related to authentication. If correct username and password is provided, then the same prompt is raised again.

## Additional Information

screen shot 2018-09-16 at 5 07 09 pm
screen shot 2018-09-16 at 5 07 27 pm

kanishk98:
Fixes #5661

Fixes issue which caused Desktop to go into a loop of authentication error prompts when providing it with the wrong URL for a repo. https://github.com/desktop/desktop/issues/5661#issue-360637309

@outofambit whenever you're free, do you mind having a look at this and giving me some feedback?


kanishk98:
Seems like this PR is causing issues with darwin on circleci (`yarn install` failing with `node-gyp rebuild` command errors). 
I didn't expect this (other CI systems pass the PR); will borrow a Mac and figure out the problem. 

In the meantime, I'd love it if someone could help me understand why this is happening.

image

outofambit:
@kanishk98 the circleci was just a intermittent failure. i re-ran the workflow and its all good!

tierninho:
Based on the scenarios outlined [here](https://github.com/desktop/desktop/issues/5661#issuecomment-422798314) these are the results:

1. Full GitHub URL pointing to nonexistent repo (i.e. https://github.com/steveward/norepohere)
Result: new messaging is shown ✅ 

2. Shortened user/repo option that points to a nonexistent repo.
Result: Correctly throws the `fatal: repository does not exist error`.

3. GitHub SSH URL pointing to a nonexistent repo (i.e. git@github.com:steveward/norepohere).
Result: Hangs forever, with only a console error showing: `net::ERR_FILE_NOT_FOUND, Uncaught (in promise) TypeError: Failed to fetch.` We should be showing something to the user. ❌ 

4. Non-GitHub URL pointing to a repository that does not exist (i.e. any web site).
Result: Correctly throws the `fatal: repository does not exist error`.

Your messaging seems to work for all these scenarios. @billygriffin Any thoughts on using 👇  for all these scenarios? 
 
screen shot 2018-11-29 at 11 34 17 am

Notes:
- Your error message disappears if you navigate away from Desktop and return. It should be sticky like "The destination already exists" error
- The messaging carries through to the other tabs Github and Enterprise. We should probably fix this.

billygriffin:
@niik I know you're working on the cloning flow in #6278 and wanted to be sure you had 👀 on this as well to make sure it doesn't conflict.

kanishk98:
@tierninho thanks for the feedback. I'll try implementing your suggestions and get back here. :)

tierninho:
@kanishk98 

> I believe this shows up even when the SSH URL points to an existing repo. Am I correct?

I get the following: "Authentication failed. You may not have permission to access the repository or the repository may have been archived. Open preferences and verify that you're signed in with an account that has permission to access this repository.)"

> Edit: The problem's arising because I'm using fetch() to check the validity of the repo URL. I don't think that allows for SSH connections. Has a similar problem been solved by the Desktop team before? I'd prefer to use the same approach if that is the case.

Not sure. cc @outofambit

shiftkey:
Dropping `ready-for-review` and moving to the blocked pile because I don't believe this is the right solution for the original issue.

kanishk98:
@shiftkey thanks for the explanation. I agree that this isn't the right solution - it's an old PR and I don't think I understood the issue properly then. 
I should be able to redo my work here this weekend.

shiftkey:
@kanishk98 thanks for following up.

Feel free to open a new PR if you prefer - a lot has changed since then, and starting fresh might help here. I'll leave this open in the meantime to track our discussion about the underlying issue.

shiftkey:
@kanishk98 thanks for opening #7634 - are we good to close this out now and move our discussion over there?

kanishk98:
> are we good to close this out now and move our discussion over there?

Yeah, no need for this to stay open.

Issue #5555, opened 2018-09-06
Improve merge flow


billygriffin:
Over the past several releases, we've focused on making it easier for people to collaborate using GitHub Desktop. There are several features we've released to that end, including the ability to compare one branch to another and subsequently merge (https://github.com/desktop/desktop/issues/3956) and notifying users when master has commits that their branch does not (https://blog.github.com/2018-07-30-github-desktop-1.3/).

We think it would be helpful to take a step back and evaluate what things are intuitive and working well and where there might be opportunities for improvement. Since we're still exploring our handling of merge conflicts in Desktop, the scope of this evaluation is intended to be from the moment a user decides they'd like to get work from one branch to another (intent, regardless of where they are in the app) up to the point where they actually initiate the merge.

The goals here are:

1. Take a look at the metrics relating to the current merge flow to understand how it's being used to date.
2. Identify the areas of confusion in the current merge flow and where things are not discoverable, and evaluate whether that aligns with what metrics show. This part should involve at least some interaction/usability testing with users outside of GitHub.
3. Propose potential solutions for making that flow more intuitive and discoverable to help reduce friction for individuals and teams that use Desktop as part of their collaboration.

cc: @desktop/design @desktop/analytics 

### Task Issues:
- [x] Make it obvious which branch is being merged into which branch - #5930
- [x] Place merge button in branch list dropdown - #5929

billygriffin:
We discussed this today and I wanted to share a more clarified set of observations and problem statements:

### Observations

We've not seen considerable evidence that people find value in comparing the commits of one branch to another when trying to merge. Additionally, in usability testing, when attempting to merge, most users went first to the "Branch" dropdown, not to the "History" tab where the compare flow currently lives. Therefore, the most discoverable way to merge one branch to another is through the menu bar. If this is working for users, we have no problem if people choose to continue that flow, but we think beginners would benefit from a more intuitive merge flow.

### Problem statements

1. It's not obvious how to merge one branch into another in Desktop.
2. In the branch dropdown, the branch filter input is confusing in its current location next to the "New branch" button. It should be obvious to people what action is for what purpose.
3. The directional merge flow is not as obvious as it could be. It should be perfectly clear which branch you're merging into which branch.
4. There's no way to see a diff of one branch to another to fully understand what's being merged in and what it will look like after the merge is completed.

For the purposes of this initial effort, we're viewing problems 1-3 as in scope, and 4 as an iteration that we may tackle later.

### How will we measure success?

[...]

PR #5817, opened 2018-10-04
Add context menu to branch list menu

### LINKED ISSUE #5803: Add context menu to branch list
Tagging #5555 since this issue is related to point 2 in that issue.

Starting a branch comparison is currently a bit confusing. Desktop requires that the user click on the _History_ tab then a text box to show a list of branches to compare against.

This is problematic because the _History_ tab is about showing commits. While it's true that selecting a comparison branch will indeed produce a list of commits that are ahead/behind the current branch, the location of the action to produce that result feels out of place.

**Proposal**
As a start, we can add a context-menu to the branch list that will allow comparisons to be initiated directly from the branch list. For now, I think the we should bring over all of the items in the _Branch_ menu except for the _New Branch..._ item.

screen shot 2018-10-03 at 9 45 27 am

I'm already aware that context-menus present a discoverability problem; however, this is intended to be a step taken to overhaul the relationship between branches experience. We also make use of these menus in other areas of the app, so there is precedent here. 

cc @desktop/design

iAmWillShepherd:
Fixes #5803

### DIFF
--- a/app/src/ui/branches/branch-list-item.tsx
+++ b/app/src/ui/branches/branch-list-item.tsx
@@ -5,6 +5,8 @@ import { IMatches } from '../../lib/fuzzy-find'
 
 import { Octicon, OcticonSymbol } from '../octicons'
 import { HighlightText } from '../lib/highlight-text'
+import { showContextualMenu } from '../main-process-proxy'
+import { IMenuItem } from '../../lib/menu-item'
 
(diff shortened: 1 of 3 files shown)


iAmWillShepherd:
Closing this out now that we have an agreed upon solution: https://github.com/desktop/desktop/issues/5555#issuecomment-429963614

Issue #3030, opened 2017-10-12
Remove `git directory` warning if no path


tierninho:
If there is no path in the New Repo modal, then we should not show the `git directory` warning message as it's inaccurate.

<img width="389" alt="screen shot 2017-10-12 at 9 56 44 am" src="https://user-images.githubusercontent.com/14828183/31487971-2ac89d7a-af34-11e7-89b7-43d44e4e0e9d.png">

tested: 1.0.4-beta1 (Mac OS 10.11.6)

joshaber:
This seems like a recent regression, so I'm gonna throw it in 1.0.x.

joshaber:
Nevermind, doesn't look recent after all.

shiftkey:
This was fixed in #5828 and will go out in `1.5.0`

PR #5828, opened 2018-10-05
fixes issue #3030 by checking for an empty field

JakeHL:
This PR fixes issue #3030 by checking for an empty directory when conditionally rendering the warning.

### DIFF
--- a/app/src/ui/add-repository/create-repository.tsx
+++ b/app/src/ui/add-repository/create-repository.tsx
@@ -416,7 +416,7 @@ export class CreateRepository extends React.Component<
   private renderGitRepositoryWarning() {
     const isRepo = this.state.isRepository
 
-    if (this.state.path.length && !isRepo) {
+    if ((this.state.path.length && !isRepo) || this.state.path.length === 0) {
       return null
     }
 


app/src/ui/add-repository/create-repository.tsx
```diff
@@ -416,7 +416,7 @@ export class CreateRepository extends React.Component<
   private renderGitRepositoryWarning() {
     const isRepo = this.state.isRepository
 
-    if (this.state.path.length && !isRepo) {
+    if ((this.state.path.length && !isRepo) || this.state.path.length === 0) {
```
Daniel-McCarthy: Thanks for the contribution. This seems to complicate the if condition somewhat. Why not something this below?
```typescript
if (this.state.path.length === 0 || !isRepo) {
```

JakeHL: I completely didn't catch the string type not being nullable. I'll update the PR 👍

JakeHL: I've updated the PR. Good catch on the non-nullable string. It basically removes any possible edge cases.

a0930119952:
從 Windows 10 的郵件傳送

寄件者: Jake Langford
傳送時間: 2018年10月6日 上午 08:51
主旨: Re: [desktop/desktop] fixes issue #3030 by checking for an emptyfield (#5828)

@JakeHL commented on this pull request.

In app/src/ui/add-repository/create-repository.tsx:
> @@ -416,7 +416,7 @@ export class CreateRepository extends React.Component<
   private renderGitRepositoryWarning() {
     const isRepo = this.state.isRepository
 
-    if (this.state.path.length && !isRepo) {
+    if ((this.state.path.length && !isRepo) || this.state.path.length === 0) {
I've updated the PR. Good catch on the non-nullable string. It basically removes any possible edge cases.

tierninho:
UI looks good. 

Simple tests passing:
- [x] Remove the path and no warning is present
- [x] Warning still present if a Git repo
- [x] link in warning is still active

app/src/ui/add-repository/create-repository.tsx
iAmWillShepherd: > It basically removes any possible edge cases.

Typescript will still let you assign `undefined` to a string and will accept `undefined` being passed as an argument to a function.

```ts
function f(a: string) {
  return a.length
}

f(undefined) //Cannot read property 'length' of undefined​​
```

☝️ TSC is happy to **not** call this an error and let it slide.

JakeHL: Ah! okay - I'll implement a check for that case this evening 👍

JakeHL:
Hi All - I've updated the condition once more. Hopefully this covers all bases. 👍

shiftkey:
I'm not sure if this _needs_ to go out with 1.4.3 - had a chance to look over the latest changes @iAmWillShepherd?

iAmWillShepherd:
I'll merge this after the next release, so it can get some beta time first.

Issue #5018, opened 2018-06-25
Keyboard Shortcut for Delete Branch


1pete:
**Is your feature request related to a problem? Please describe.**
Currently the instructions to delete branch are to open menu in menu bar, which require mouse movement and a few clicks. In large monitor menu bar is also quite far from application window 😛 

**Describe the solution you'd like**
Add keyboard shortcut e.g. `cmd+shift+D`, after pressing them, the delete branch popup will appear, just like normal flow when you select menu item from menu bar.

Thanks

tongovan:
@398¢$DDDJJ

PR #5830, opened 2018-10-06
#5018 Added shortcut key to delete branches

JakeHL:
Fixes: #5018

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -278,6 +278,7 @@ export function buildDefaultMenu(
         label: __DARWIN__ ? 'Delete…' : '&Delete…',
         id: 'delete-branch',
         click: emit('delete-branch'),
+        accelerator: 'CmdOrCtrl+Shift+D',
       },
       separator,
       {


agisilaos:
Hello @JakeHL! Thank you so much for this PR. Can you please add `Fixes:` before #5018 so it can be closed automatically after we merge this PR?

JakeHL:
All done 👍

say25:
It’s kinda a shame that CmdOrCtrl+Shift+R is for the repo menu as rename is the only similar without a shortcut now :/

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -278,6 +278,7 @@ export function buildDefaultMenu(
         label: __DARWIN__ ? 'Delete…' : '&Delete…',
         id: 'delete-branch',
         click: emit('delete-branch'),
+        accelerator: 'CmdOrCtrl+Shift+D',
```
say25: It would appear that everywhere else we specify `accelerator` before `click` might make sense to do the same for consistency?

say25:
@iAmWillShepherd I think something like `CmdOrCtrl+Delete` might make more sense as an accelerator for deleting the repository

iAmWillShepherd:
@say25 

> I think something like CmdOrCtrl+Delete might make more sense as an accelerator for deleting the repository

My question was about <kbd>Delete</kbd> not ⌘+Delete.

say25:
@iAmWillShepherd how would that work if you were to click `Delete` inside the Summary Box for instance?

iAmWillShepherd:
> @iAmWillShepherd how would that work if you were to click Delete inside the Summary Box for instance?

It would delete the text as expected, but I think I understand where our understandings diverge. I was under the impression this would make deleting arbitrary branches easier, not deleting the current branch the user is on.

JakeHL:
You fellas decide what you're after and I'll alter my PR ;-)

iAmWillShepherd:
I think @say25 and I are on the same page. I think the command as you have it is ✨

shiftkey:
@say25 no, the order doesn't matter here - it's just an object that gets used to generate the menu shortcuts.

Issue #5827, opened 2018-10-05
`Install and Restart` button should be default button in Release Notes modal


tierninho:
When clicking "what's new" in the `update-available` banner, the release notes has the buttons backward.

Reproduce:
- download latest update
- navigate to update banner

### Expected result:
`Close` has no-color, left button
`Install and Restart` is blue, default button, right button

![oct-05-2018 15-30-31](https://user-images.githubusercontent.com/14828183/46563151-c46dd180-c8b4-11e8-8205-064d3911f22a.gif)

tested in 1.4.2-beta0, mac

JKirkYuan:
I'll take a look at this one. Been playing around with how buttons are working all week! 👍

iAmWillShepherd:
I think that is intended? In other areas of the app, destructive actions are never the default value.

Here's deleting a branch to illustrate:
![screen shot 2018-10-08 at 1 48 17 pm](https://user-images.githubusercontent.com/1715082/46627689-dcd22c00-cb00-11e8-88e7-6efa7a81bb60.png)

Note that the Cancel button is the main button

shiftkey:
> I think that is intended? In other areas of the app, destructive actions are never the default value.

This is still my "2 cents" on the issue, which I touched on with the original PR about in-app release notes: https://github.com/desktop/desktop/pull/3601#issuecomment-418850431

Also note that our About dialog, which is the other place we let the user restart the app, will also default to closing the app: https://github.com/desktop/desktop/pull/3601#issuecomment-419073963

billygriffin:
Thanks for all the feedback here folks! I had a chance to catch up with @tierninho and given https://github.com/desktop/desktop/pull/3601#issuecomment-419073963 and the desire to treat this as a destructive action, the existing implementation seems reasonable. I'm going to go ahead and close this and the associated PR. Thanks so much @JKirkYuan for being willing to dive in. ❤️  This, among other things, has been an impetus for me to move higher on my priority list writing up documentation about our dialogs and how we approach them in line with Apple and Microsoft's guidance for macOS and Windows, respectively.

PR #5831, opened 2018-10-06
`Install and Restart` button should be default button in Release Notes modal

JKirkYuan:
This was a quick fix to #5827 ! To highlight the button `Install and Restart` I added `type="submit"` as an attribute

This is a destructive button so it will default to close when the `Enter` key is pressed.

### DIFF
--- a/app/src/ui/release-notes/release-notes-dialog.tsx
+++ b/app/src/ui/release-notes/release-notes-dialog.tsx
@@ -174,10 +174,10 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
         <DialogContent>{contents}</DialogContent>
         <DialogFooter>
           <ButtonGroup destructive={true}>
-            <Button type="submit">Close</Button>
-            <Button onClick={this.updateNow}>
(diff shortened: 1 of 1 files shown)


billygriffin:
Thanks @JKirkYuan! Would you mind adding screenshots to allow any reviewers to see the before/after? I'm noodling on the definition of a "destructive action" and whether this fits, but I think the screenshots would be a really helpful step in visualizing that if you wouldn't mind.

billygriffin:
@JKirkYuan Friendly ping on the above ask 😄

JKirkYuan:
@billygriffin Hey! Sorry I didn't respond, I've been super busy. I'll do this tonight when I get a chance. There is also a discussion going on #5827 to see if we actually need to highlight the `Install and Restart` button.

billygriffin:
Thanks @JKirkYuan for making this PR, and sorry for the back and forth. I really hope the fact that we're closing this doesn't dissuade you from contributing on other issues, as it was so cool to see your willingness to just dive in and fix something. ❤️ Thanks again!

PR #5841, opened 2018-10-08
Improving toPlatformCase

say25:
I saw that the helpful `.toPlatformCase` was added as part of https://github.com/desktop/desktop/pull/5744/. Shout out to @Daniel-McCarthy. 

This Pull Request:
1. Adds MOAR Unit Tests
2. Add the ability to ignore "default" platform case rules to handle cases like those in [Context-Menu.ts](https://github.com/desktop/desktop/blob/master/app/src/ui/lib/context-menu.ts)

### DIFF
--- a/app/test/unit/platform-case-test.ts
+++ b/app/test/unit/platform-case-test.ts
@@ -38,4 +38,30 @@ describe('string to platform case', () => {
       expect(result).to.equal(' This should be sentence case.')
     }
   })
+  it('handles special case words to detected platform text case', () => {
+    if (__DARWIN__) {
+      const result = toPlatformCase(
+        'Open _in_ External Editor. Show _in your File Manager_'
+      )
+      expect(result).to.equal(
+        'Open in External Editor. Show in your File Manager'
+      )
+    } else {
+      const result = toPlatformCase(
+        'Open _in_ External Editor. Show _in your File Manager_'
+      )
+      expect(result).to.equal(
+        'Open in external editor. Show in your File Manager'
+      )
+    }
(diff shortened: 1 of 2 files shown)


app/test/unit/platform-case-test.ts
```diff
@@ -2,7 +2,7 @@ import { expect } from 'chai'
 import { toPlatformCase } from '../../src/lib/platform-case'
 
 describe('string to platform case', () => {
-  it('convert string to detected platform text case', () => {
+  it('converts lower string to detected platform text case', () => {
```
iAmWillShepherd: s/lower/lowercase

say25:
@iAmWillShepherd makes sense to me. I was trying to improve the test coverage an noticed the same thing you did which is the whole situation with words like `in` and `and`. Given the request for future discussion we should put this on hold it would seem.

say25:
Depending on the outcome of the discussion, we can backout the extra commits.

iAmWillShepherd:
> @iAmWillShepherd makes sense to me. I was trying to improve the test coverage an noticed the same thing you did which is the whole situation with words like in and and. 

Yea I figured and I initially took it as that until I really honed in on the function you're testing. I think you were on the right track with considering test coverage, but I think we may be removing the function you exercised because as you just said, it can get complicated in a hurry! 👍

shiftkey:
> Given the request for future discussion we should put this on hold it would seem.

I'm going to close this out while we discuss this further in #5841 - thanks for digging into this in the meantime @say25!

PR #5850, opened 2018-10-08
Returns the ability to use Emojis when the app starts

iAmWillShepherd:
We received some reports emojis weren't working consistently, this PR addresses that issue.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1665,9 +1665,9 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   private renderUpdateBanner() {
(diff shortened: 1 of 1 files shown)


niik:
I'm uncertain whether this is resolved by #5859 or if it's a separate issue? I'm hesitant to remove the idle callback entirely as I think that there's still value in deferring potentially time consuming but not critical tasks until the app is fully loaded.

If this is not resolved by #5859 I think we should investigate alternative approaches such as only eagerly loading emojis, setting a deadline for the idle callback, or loading emojis on demand.

iAmWillShepherd:
Oops, looks like I forgot to close this PR. Once I realized this fix wasn't consistent, I went back to the drawing board. PR #5859 is the true MVP!

Issue #5566, opened 2018-09-07
show something when repository filter returns no matches


tierninho:
I noticed we surface actions to engage the user to either 'create a new branch' or 'create a new PR' when a filter shows no results. However, the same cannot be said for the results of a Repo, when they are null. 

We just offer up a bunch of white space. We should take advantage of that space to engage the user with buttons for New Repository, Add Local Repo, Clone Repo options (similar to how our landing page looks).

Repo space:
<img width="245" alt="screen shot 2018-09-07 at 10 52 37 am" src="https://user-images.githubusercontent.com/14828183/45235153-d562fd00-b28c-11e8-9c8a-fb82ed4ce87d.png">

Branch/PR suggestions:
<img width="305" alt="screen shot 2018-09-07 at 10 52 52 am" src="https://user-images.githubusercontent.com/14828183/45235161-dbf17480-b28c-11e8-8e94-1c4f60f8432f.png">
<img width="298" alt="screen shot 2018-09-07 at 10 52 59 am" src="https://user-images.githubusercontent.com/14828183/45235162-dbf17480-b28c-11e8-92ba-4af1b4d5d162.png">

cc @desktop/design

ampinsk:
Thanks for pointing this out @tierninho! I think it's definitely space that we should be at least showing an empty state and a good opportunity to point users to a next step. I think my question is more should we: 

1. Surface all three options (New Repository, Add Local Repo, Clone Repo)?
2. Choose one based on some criteria (most frequently used, what users are most likely looking for in this context)? 

I'd lean towards choosing one action based on what's most frequently used of the three. 👍 

@desktop/analytics do we have metrics on these actions?

This is also a side point, but I just realized we speak from the first person here, do we do that anywhere else? I'd expect us to speak more from the third person. 🤔

PR #5878, opened 2018-10-11
Add new repo button in repository dropdown

niik:
We've heard in research and feedback that users often look in the repository dropdown for a way to add repositories. This adds an "add" dropdown button to the repository sidebar similar to what we have in the branches dropdown to hopefully alleviate that problem.

screen shot 2018-10-11 at 16 59 58

This addresses one part of #5566 and I plan to tackle the blank slate view when filtering in a follow-up PR.

### DIFF
--- a/app/src/ui/repositories-list/repositories-list.tsx
+++ b/app/src/ui/repositories-list/repositories-list.tsx
@@ -15,6 +15,8 @@ import { enableRepoInfoIndicators } from '../../lib/feature-flag'
 import { Dispatcher } from '../../lib/dispatcher'
 import { Button } from '../lib/button'
 import { Octicon, OcticonSymbol } from '../octicons'
+import { showContextualMenu } from '../main-process-proxy'
+import { IMenuItem } from '../../lib/menu-item'
 
 interface IRepositoriesListProps {
   readonly selectedRepository: Repositoryish | null
@@ -169,12 +171,42 @@ export class RepositoriesList extends React.Component<
 
   private renderPostFilter = () => {
     return (
-      <Button className="new-repository-button">
+      <Button
+        className="new-repository-button"
+        onClick={this.onNewRepositoryButtonClick}
+      >
         {__DARWIN__ ? 'Add' : 'Add'}
         <Octicon symbol={OcticonSymbol.triangleDown} />
       </Button>
     )
   }
+
+  private onNewRepositoryButtonClick = () => {
+    const items: IMenuItem[] = [
+      {
+        label: __DARWIN__ ? 'Clone Repository…' : 'Clone repository…',
+        action: this.onCloneRepository,
+      },
+      {
+        label: __DARWIN__
+          ? 'Add Existing Repository…'
+          : 'Add existing repository…',
+        action: this.onAddExistingRepository,
+      },
+      {
+        label: __DARWIN__ ? 'Create New Repository…' : 'Create new repository…',
+        action: this.onCreateNewRepository,
+      },
+    ]
+
+    showContextualMenu(items)
+  }
+
+  private onCloneRepository = () => {}
+
+  private onAddExistingRepository = () => {}
+
+  private onCreateNewRepository = () => {}
 
   private noRepositories() {
     return (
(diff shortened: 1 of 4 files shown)


billygriffin:
@niik I love that we're doing this! ❤️ 

In looking at the branch dropdown where "New" is next to the input for filtering, we called out that it could be confusing because the two can be conflated. I wonder if similarly here there would be value in separating the filter input from the "Add" button, which might also give a bit more real estate to allow the button to say something more explicit (like "Add repository" instead of "Add"). Thoughts?

tierninho:

Two notes:
- Should we arrange the list to mimic the order in the File menu?
- If the user enters text in the filter, should the text be carried through to the repo name (as adding a new branch does this).

Build/OS: 1.4.3-beta0, mac

billygriffin:
> If the user enters text in the filter, should the text be carried through to the repo name (as adding a new branch does this).

My feedback was primarily on this. That seems like it might theoretically be nice, but I'm concerned about the conflation of creating a new repo with filter among existing repos. I think they're separate concerns and while it's a slick implementation to type in a new one, I think separating different actions would help prevent confusion for beginners.

app/src/ui/repositories-list/repositories-list.tsx
```diff
@@ -164,6 +170,54 @@ export class RepositoriesList extends React.Component<
     )
   }
 
+  private renderPostFilter = () => {
+    return (
+      <Button
+        className="new-repository-button"
+        onClick={this.onNewRepositoryButtonClick}
+      >
+        {__DARWIN__ ? 'Add' : 'Add'}
```
shiftkey: :art: no need for this ternary, right?

shiftkey:
> If the user enters text in the filter, should the text be carried through to the repo name (as adding a new branch does this).

I'd rather keep this change focused on the button and flow, and if we notice a theme of feedback in this vein we can iterate on the UX later.

ampinsk:
> In looking at the branch dropdown where "New" is next to the input for filtering, @ampinsk called out that it could be confusing because the two can be conflated. 

I think in this instance, with the added dropdown, I don't think this is as easily conflated. 

> I wonder if similarly here there would be value in separating the filter input from the "Add" button, which might also give a bit more real estate to allow the button to say something more explicit (like "Add repository" instead of "Add"). Thoughts?

@niik and I chatted through this in Slack, and we had some concerns about exposing all of the actions since it really crowded the menu in a way that made everything harder to parse, and might be confused for a tab menu. These menus weren't really designed to accommodate exposing multiple actions like this so we're in a little bit of a bind until we can maybe redesign these areas more holistically

> That seems like it might theoretically be nice, but I'm concerned about the conflation of creating a new repo with filter among existing repos. I think they're separate concerns and while it's a slick implementation to type in a new one, I think separating different actions would help prevent confusion for beginners.

I don't feel too strongly either way here. I don't think it's necessarily _confusing_, but if we feel strongly that it's conflating concerns I would be fine not having that behavior. Have we heard of anyone being blocked or confused by this behavior?

billygriffin:
> Have we heard of anyone being blocked or confused by this behavior?

It's definitely an assumption I hold because I've always been confused by it. Traditionally an input field next to a button means that the button is a submit for that input field, so that's the source of my confusion. But if y'all disagree with that assumption, I'm fine to continue as is now and test it later.

> @niik and I chatted through this in Slack, and we had some concerns about exposing all of the actions since it really crowded the menu in a way that made everything harder to parse, and might be confused for a tab menu. These menus weren't really designed to accommodate exposing multiple actions like this so we're in a little bit of a bind until we can maybe redesign these areas more holistically

I also don't think we should expose all the options individually, but rather I was suggesting that the verbiage on the dropdown button could be more explicit potentially. So instead of "Add" we might be able to use "Add repository" with the same three options available in the dropdown. Totally fine if we don't think that's better, but just wanted to clarify. 😄

shiftkey:
> So instead of "Add" we might be able to use "Add repository" with the same three options available in the dropdown. Totally fine if we don't think that's better, but just wanted to clarify. 😄

@billygriffin I think we're constrained by the available horizontal space here, so I'm fine with just using "Add" for now until we can think about this area more broadly, like @ampinsk suggested.

shiftkey:
Nothing further raised here, and it's Thursday, so I'm gonna take this in.

101010431:
/app/styles/ui/_repository-list.scss

PR #5880, opened 2018-10-11
extract GitStoreCache from AppStore to manage state

shiftkey:
Following on from #5564 and `RepositoryStoreCache`, `AppStore` also manages the lifetimes of the `GitStore` instances we create. We can extract these out to be managed by another component.

 - [ ] add tests

### DIFF
--- a/app/src/lib/stores/git-store-cache.ts
+++ b/app/src/lib/stores/git-store-cache.ts
@@ -1,0 +1,44 @@
+import { GitStore } from './git-store'
+import { Repository } from '../../models/repository'
+import { IAppShell } from '../app-shell'
+import { Commit } from '../../models/commit'
+
+export class GitStoreCache {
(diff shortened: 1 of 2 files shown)


app/src/lib/stores/git-store-cache.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  public remove(repository: Repository) {
+    if (this.gitStores.has(repository.hash)) {
+      this.gitStores.delete(repository.hash)
+    }
+  }
+
+  public get(repository: Repository): GitStore {
+    let gitStore = this.gitStores.get(repository.hash)
+    if (!gitStore) {
```
iAmWillShepherd: Let's make this check more explicit: `if (gitStore === undefined)`

shiftkey:
:apple:

PR #5882, opened 2018-10-11
simplify Compare tab state

shiftkey:
This PR changes this set of enums:

```ts
export enum ComparisonView {
  None = 'none',
  Ahead = 'ahead',
  Behind = 'behind',
}

export enum CompareActionKind {
  History = 'History',
  Branch = 'Branch',
}
```

Into this:

```ts
export enum HistoryTabMode {
  History = 'History',
  Compare = 'Compare',
}

export enum ComparisonMode {
  Ahead = 'Ahead',
  Behind = 'Behind',
}
```

Why?

 - We should make the History tab mode a first class field in the application state
 - In hindsight, `ComparisonView` feels vague in the context of the tab, especially when paired with `ComparisonView.None` which actually represents "We're Viewing History Now"
 - `CompareActionKind` was aspirational to represent actions being dispatched to update state, but we can just use `HistoryTabMode` which feels more meaningful
 - the fact we've mixed up `kind` and `mode` in some places makes me think these should be two distinct things

```ts
this.repositoryStateCache.updateCompareState(repository, s => ({
  formState: {
    comparisonBranch,
    kind: action.mode,
    aheadBehind,
  },
  ...
}))
```


app/src/lib/app-state.ts
```diff
@@ -631,27 +631,43 @@ export interface IChangesState {
...
 /**
  * When the user has chosen another branch to compare, using their current
  * branch as the base branch.
  */
 export interface ICompareBranch {
+  readonly kind: HistoryTabMode.Compare
+
   /** The chosen comparison mode determines which commits to show */
-  readonly kind: ComparisonView.Ahead | ComparisonView.Behind
+  readonly mode: ComparisonMode.Ahead | ComparisonMode.Behind
```
iAmWillShepherd: Perhaps we could name use a more descriptive name. Something like `comparisonMode` or `compareMode`?

shiftkey:
:apple:

PR #5895, opened 2018-10-13
Allow cloning repositories to empty folders

### LINKED ISSUE #5857: Clone Repository not working
I have tried to use the File>Clone repository option on my Desktop. The option does not seem to work 

Environment:
GitHub Desktop version: 1.4.2
Git-GUI for windows 2.18
Windows 7 Pro (x64)

source url: "https://github.com/firebase/quickstart-android.git"
destination: D:\WorkDesk\Repos\Android\android

message: "Destination already exists" where the destination is an empty folder created for cloning.

if the destination is a non-existant folder (with immediate parent is available, eg:"D:\WorkDesk\Repos\Android\quickstart-android"), then the message changed to "Sorry, I can't find that repository".

"Open in Desktop" from chrome does not seem to identify GitHub Desktop installation either.

Daniel-McCarthy:
## Overview
Closes #5857. 

## Description
To give some information on the issue:
- The `Clone Repository dialog` validates the user's entered destination path and checks to see if a folder is at the path to alert the user and stop them from cloning to it.
- This validation also occurs if the user had clicked the destination path input or modified it. However, if one copied a folder to where the destination is and switched back to Desktop to clone, no re-validation would occur. GitHub Desktop would then allow them to clone instead of displaying the error, which would result in a pop-up Git error if the folder contained files.
- Since then, the `Clone Repository dialog` has been updated to validate on window focus to catch situations like this. However, this prevents the user from cloning in situations where they have created a fresh empty folder to clone into, rather than having GitHub Desktop create the folder for them.
___
The aim of this pull request is to allow this user workflow by allowing the path validation function to detect if a folder is empty or not and use that information to allow cloning if it is empty or generate the `Destination Exists` error if it contains files.
___

Additionally, I would like to point out two things that I think would be great potential places for feedback:

- With this change, the added `isPathEmptyFolder` function currently has the ability to detect both a situation where a folder is empty and a situation where the folder does not exist. This means that in theory, calling _both_ `pathExists` and `readdirSync` is potentially extraneous work that can be simplified. However, I have not taken pathExists out of use because I feel that `isPathEmptyFolder` is quite subject for change (if it will even be used at all).

- Perhaps there is a way we can detect that files exist in a folder that stops after finding the first file rather than retrieving the names/paths of all files in the folder (since there may be a very large number of files in the folder, and may occur repeatedly).
___
I would love to hear feedback on this and see how we can best improve this.

## Release notes
Notes: Added ability to clone to existing folders if they are empty.


Daniel-McCarthy:
@shiftkey I have updated this and I believe it should address both concerns. The function now uses `fse.readdir` rather than `fse.readdirSync`, additionally it now returns `Promise<boolean>` rather than `any`. It is also closer to how it is handled elsewhere in the code base.

Daniel-McCarthy:
@outofambit I can surely update this to use only one await, that would be no issue at all. If you'd like I can take care of that in this pull request while it's still open.

It certainly works either way.

outofambit:
> I can surely update this to use only one await, that would be no issue at all. If you'd like I can take care of that in this pull request while it's still open.

@shiftkey thoughts?

shiftkey:
@outofambit @Daniel-McCarthy let's address it in this PR

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -336,6 +339,25 @@ export class CloneRepository extends React.Component<
...
+        return
+      }
+
+      return directoryFiles.length === 0
+    } catch (error) {
+      if (error.code === -2) {
+        // Directory does not exist error
+        return false
+      }
+      return error
```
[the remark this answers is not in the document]
Daniel-McCarthy: @shiftkey Since it has been decided to not have both `isPathEmptyFolder` and `doesPathExist` and instead cover both with one function, the goal of this function may need to shift a bit.

Since this function now needs to also cover situations where the folder does not exist as well, we have a few options. If there if the path doesn't exist `readdir` will return an error. We can potentially return `true` in this case, indicating that the path is an empty folder, which is not _exactly_ correct but does not mess with the logic flow.

Otherwise we can change the question we're trying to answer a bit.

```typescript
  private async doesPathContainFiles(path: string): Promise<boolean | Error> {
    try {
      const directoryFiles = await readdir(path)
      return directoryFiles.length !== 0
    } catch (error) {
      if (error.code === 'ENOENT') {
        return false
      }

      return error
    }
  }
```
> I'd also like to avoid returning an error here - we should return something typed or rethrow the error (preferably an Error type) if it should be handled by the caller.

I'm not sure what the full variety of errors that can be thrown by `readdir` are. I would love if I could cover the `ENOENT` path does not exist error and return a boolean in that case and not have to handle an error from there, but I have done `boolean | Error` for now.

I have not committed these changes yet, as I figured it was best to discuss the changes first and see which direction it would be best to go in.

outofambit:
@Daniel-McCarthy here's what i'm thinking: if we do a single `readdir` and catch the errors for nonexistent folders (but not other errors of course), then we know everything we need to, right?Basically if the folder isn't there, or is empty, then clone into that folder. otherwise (if the folder exists and has stuff in it) we throw an error. i think the only other concern then is to make sure we preserve this logic: 

https://github.com/desktop/desktop/blob/13ee781f396d2f84e0e66e0d4a3a427100726530/app/src/ui/clone-repository/clone-repository.tsx#L332-L334

what do you think about that approach?

Daniel-McCarthy:
@outofambit Excellent, I will modify what I currently have and will update it with these suggestions. It sounds as though we don't need an extra function at all which solves my concern.

Sounds good!

Daniel-McCarthy:
@outofambit After looking at the code and figuring out how to best incorporate these changes, I have a set of changes and would love to know what you think about them, and potentially iterate from there.
___
```typescript
  private async doesPathContainFiles(path: string): Promise<boolean | Error> {
    try {
      const directoryFiles = await readdir(path)
      return directoryFiles.length !== 0
    } catch (error) {
      if (error.code === 'ENOENT') {
        return false
      }

      return error
    }
  }
```
Currently this `doesPathContainFiles` function I have tried uses a single `readdir` call and attempts to answer the most appropriate question to handle both of what the current functions do.

- If the folder exists, it will return true or false regarding whether or not it contains files. 
- If the folder does not exist, it will return false.
- If there is an error in a situation other than the folder not existing it will return the error.

Returning the error is something I want to iterate on.  I am not sure what other possible errors can be generated from this `readdir` function. I figured we can return the error and potentially handle it in the future. Or instead, we can assume that if there is an error it is due to the folder not existing, allowing us to not hold onto the error at all and instead return a `Promise<boolean>` instead of `Promise<boolean | Error>` which would be a preferable change.
___
```typescript
  private updateAndValidatePath = async (path: string) => {
    this.setState({ path })

    const doesPathContainFiles = await this.doesPathContainFiles(path)
    const pathHasChanged = this.state.path !== path

    if (doesPathContainFiles === true || pathHasChanged === true) {
      const error: Error = new Error('The destination already exists.')
      error.name = DestinationExistsErrorName

      this.setState({ error })
    } else {
      this.setState({ error: null })
    }
  }
```
If the path the user has selected is a folder that contains files then the destination already exists error will be displayed. Additionally, to preserve the original functionality of throwing the error when the path has changed, it will also do this.

(Another place to iterate on is to move the `pathHasChanged` part into the function in order to discard the results if the path has changed. The reason this has not been done already is that the function currently returns `Promise<boolean|Error>` and it felt as though the possible results of the function would be even murkier until that part is handled)
___
I wanted to share what I am looking at for this and make sure we're all on the same page on this, then I can make those additional changes as well based on what we're thinking on this.

So my questions are:
- Does this shifting from `doesPathExist` and `isPathEmptyFolder` to `doesPathContainFiles` seem to be a good change for this?
- Should we always return false (indicating the folder does not contain files) when an error is thrown? This way we can stop catching other errors, but I wanted to make sure we were on the same page that this is the functionality to go for.
- Should `doesPathContainFiles` simply return when the path has changed, therefore not indicating a positive or negative result, or simply handle that situation outside of the function? (This part might be a bit tied to the previous question)

Daniel-McCarthy:
@shiftkey 
Returning `Promise<Error | null> does simplify matters in regards of checking the results and I do like this approach, as well as having a simple `validateX` function name. Those handle the issue quite sufficiently.
___
My remaining question then is regarding using an `Error` type in place of the JS model.
- If the folder has no files we can just return null.
- If the filesystem returns an error suggesting there is no folder, we can simply return null or throw an error explaining such, allowing the caller to make use of that.

My question is regarding if an error is thrown that I am not specifically expecting (and therefore don't have a prepared message to set it to in the new `Error`. In that case, do I take the message from the JS error and use it in the new `Error` type so that it can be re-thrown by the receiver?
___
Sorry to make this complicated. I understand we want to use a different type, I just wasn't sure if we wanted to re-use the existing error message as a fallback plan when we don't know what type of error it is.
Thanks for the help with this.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -269,9 +269,9 @@ export class CloneRepository extends React.Component<
   private updateAndValidatePath = async (path: string) => {
     this.setState({ path })
 
-    const doesDirectoryExist = await this.doesPathExist(path)
+    const pathValidation = await this.validateEmptyFolder(path)
 
-    if (doesDirectoryExist) {
+    if (pathValidation != null) {
```
shiftkey: Should we use `pathValidation` on the next line?

```ts
if (pathValidation != null) {
  this.setState({ error: pathValidation })
} else {
  this.setState({ error: null })
}
```

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -326,14 +326,28 @@ export class CloneRepository extends React.Component<
...
-    if (this.state.path !== path) {
-      return
-    }
+  private async validateEmptyFolder(path: string): Promise<null | Error> {
+    try {
+      const directoryFiles = await readdir(path)
+
+      // If the path has changed we no longer want the result
+      if (path !== this.state.path) {
+        return new Error('The path has been changed during check.')
```
shiftkey: For other cases like this, where the path has changed under us, I think we just return `null` to indicate no action, rather than returning an error. Does that make sense?

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -326,14 +326,28 @@ export class CloneRepository extends React.Component<
...
+        return new Error('The path folder contains files.')
+      }
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        // Folder does not exist
+        return null
+      }
 
-    return exists
+      return new Error(error.message)
```
shiftkey: Rather than returning the raw error here (which might be terrifying for the user), could we log the error (to help with reporting an issue) and return a friendly message? Something like `"Unable to read location on disk. Please check the path and try again"`

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -326,14 +326,28 @@ export class CloneRepository extends React.Component<
...
+
+      // If the path has changed we no longer want the result
+      if (path !== this.state.path) {
+        return new Error('The path has been changed during check.')
+      }
+
+      if (directoryFiles.length === 0) {
+        return null
+      } else {
+        return new Error('The path folder contains files.')
```
shiftkey: :art: this message should indicate that Git cannot clone to directory with files, if we're going to display it to the user

app/src/ui/clone-repository/clone-repository.tsx
Daniel-McCarthy: I am currently thinking through the results of this change.

After making this change, in the situation of the path having changed while `validateEmptyFolder` runs, then:
- `validateEmptyFolder` would return null
- The caller of `validateEmptyFolder` would then set the UI `Error` state null

If I understand correctly, then the user would not see any error and not be blocked from attempting to cloning to a path, even though the path has not yet been validated.
___
I am not sure if this is the expected behavior.
I would like to make sure we are on the same page on this change before making a commit for this.

However, I may be thinking through this incorrectly.

app/src/ui/clone-repository/clone-repository.tsx
Daniel-McCarthy: I had considered this as it seemed like a pretty natural change. My decision against doing so came down the other possible errors not seeming fit to display to a user.

However, this change would make even more sense since we will no longer be returning the raw errors via the validation function.
___
Even a step further, if we're willing to set the state to `pathValidation` directly, we could even skip the `if/else` entirely since `pathValidation` will always either be an `Error` or `null`.

If we don't need any additional validation after the other changes then this would be a quite nice clean up.

shiftkey: > Even a step further, if we're willing to set the state to `pathValidation` directly, we could even skip the `if/else` entirely since `pathValidation` will always either be an `Error` or null.

I like this idea

app/src/ui/clone-repository/clone-repository.tsx
shiftkey: I think we should check this before updating the state, but that clashes with what we were talking about on the previous line

Daniel-McCarthy: I think the best scenario we have would be to display an error when the path has changed and block cloning, perhaps telling them to try again.

If we don't display an error of some sort when the path has not yet been validated then they will be able to clone (potentially failing because the path can be anything). If the `Error` state does not get updated in this situation, it might potentially remain as no error and allow cloning.

Previous to my changes if a path has been changed during the check it would simply display the generic "The destination already exists." Perhaps "Failed to validate the entered path. Please try again."

shiftkey: > I think the best scenario we have would be to display an error when the path has changed and block cloning, perhaps telling them to try again.

I still don't think this is the right option - as a user it doesn't matter to me if the path has changed while doing this check, I'd just like the app to check the path I chose.

> If we don't display an error of some sort when the path has not yet been validated then they will be able to clone (potentially failing because the path can be anything).

Ideally there would be an updated check coming through where `path === this.state.path` - if this isn't the case then maybe there's a race condition in here?

Daniel-McCarthy:
> I'm going to have to drop myself from reviewing this because I'm on holidays from next week

Thanks for the feedback you've given up to now and I hope you enjoy your well deserved vacation!
___
> I still don't think this is the right option - as a user it doesn't matter to me if the path has changed while doing this check, I'd just like the app to check the path I chose.

I would certainly be inclined to agree. Before these changes, in this scenario it would just display the original "Destination already exists" error. I have tested what would happen if the path has changed to an invalid one and no UI Error has been thrown to block cloning.

The Clone button will become available allowing an attempt to clone. 
Either the path is valid and it initiates a clone successfully, or depending on the invalid path it may throw a git error such as this:
image

One potential change would be to attempt to validate before cloning when the `Clone Repository` button is clicked. That way we have a final opportunity to check and display a UI error to the user instead of trying to clone and potentially throwing git errors.
___
I figured I would lay out these thoughts to have available for the next reviewer.

app/src/ui/clone-repository/clone-repository.tsx
outofambit: ok so here's my take: the problem is that we use `this.state.path` to control what is displayed in the text field the user is typing in. therefore, i think we should never override that. the error state is separate from that and i think its ok if its not guaranteed to be immediately correct (it should only be delayed a little bit).

i think this check (`path !== this.state.path`) is unnecessary (and confusing) for this method. we should simply perform the validation here and let the state concerns be handled by callers. (i'm pretty sure i've indicated something contrary to this in the past so sorry for missing some of this nuance upfront!)

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -326,14 +326,28 @@ export class CloneRepository extends React.Component<
...
+      }
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        // Folder does not exist
+        return null
```
outofambit: shouldn't this return an error so we don't consider it a success case in the caller?

Daniel-McCarthy: I have opted to return `null` here indicating a success, because if the folder does not exist then we should be able to clone to it. (Therefore, `ENOENT` indicates the only scenario in which we would be able to clone previous to these changes)

If you mean we should return an error and catch it in the caller, while still allow cloning when we catch this specific error, we could certainly do that. However, we would no longer be able to do the cleanup @shiftkey had suggested where we set the `error` state directly with the result of the validation function, though. That would be perfectly fine if that's what's preferable.

___

In regards to your input on the matter of when the path no longer matches the state, that's wonderful! I'm very glad to have a solution here so that we can press forward on this!
Thanks for the help on this. 

I will happily make the necessary changes once we are on the same page with your other comment.

outofambit: > I have opted to return null here indicating a success, because if the folder does not exist then we should be able to clone to it. (Therefore, ENOENT indicates the only scenario in which we would be able to clone previous to these changes)

Oh, you're totally right! Disregard my comment on that line. :)

outofambit:
did some local testing, and everything works great 🎉

PR #5898, opened 2018-10-14
don't render the merge status when there are no commits to merge

### LINKED ISSUE #5890: Do not show merge hint in merge dialogue when count is 0
The merge hint should be hidden in the merge dialogue when count is 0, aka there are no merges to commit.

screen shot 2018-10-12 at 11 41 31 am

Repro:
- Select Branch > Merge into current branch
- Find two branches where it shows "This will merge 0 commits from A-branch into B-branch"

Related to: https://github.com/desktop/desktop/pull/5581

tested in 1.4.3-beta0, mac

shiftkey:
Fixes #5890

Current `master` build when you create a new branch and try to merge it into the default branch (note that merge is disabled):

With this change:

### DIFF
--- a/app/src/ui/merge-branch/merge.tsx
+++ b/app/src/ui/merge-branch/merge.tsx
@@ -119,7 +119,8 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
       currentBranch == null ||
       selectedBranch == null ||
       currentBranch.name === selectedBranch.name ||
-      commitCount == null
+      commitCount == null ||
+      commitCount === 0
     ) {
       return null
     }


tierninho:
Tested the change, looking good 💰

niik:
The code change looks good but I'm gonna request a review from @ampinsk on this. While this is a step in the right direction I’m hesitant to just disable the button, especially given how the disabled button state is so close to the enabled state. I think we should add some messaging here to explain why we’re not letting the user merge.

billygriffin:
Just saw that @outofambit pushed this https://github.com/desktop/desktop/pull/5926 which might be useful to address @niik's concerns (also might be too subtle). cc: @ampinsk

ampinsk:
@niik and I chatted about this in Slack, I think a better solution here would be to update the merge status message to say something like "This branch is up to date with [branch name]", so that we give the user a reason why the button is disabled

shiftkey:
Here's an update to the flow to show a different message:

shiftkey:
:apple:

PR #5909, opened 2018-10-16
Removed invalid -P argument from a “mocha” call

### LINKED ISSUE #5908: Reference to “tsonfig.json” in package.json
In `package.json`, line 12…

https://github.com/desktop/desktop/blob/4272bda3720a4e00b379c83643229be9565503cc/package.json#L12

…there’s a reference to `tsonfig.json`. There’s no such file, so it’s probably a typo. It was introduced in ff3c5a8c1e44c0b478dce3730bde15e6df7090e2.

gscscnd:
Closes #5908


shiftkey:
I'm going to take this in because I got a second approval and this doesn't touch app code.

say25:
#jolo

j-f1:
Does this affect the behavior of the `ts-node` module?

shiftkey:
@j-f1 the recommendation from `ts-node` is to use environment variables if you want to control behaviour and don't have the ability to pass in command line arguments, and that's what the default config looks for if `-P` or `--project` is unset.

I don't believe this was being used, and the differences in behaviour between `tsconfig.json` and `script/tsconfig.json` for compiler options is so minor that I'm not worried about it.

PR #5809, opened 2018-10-03
Conflicted merge resolution flow

outofambit:
🎉 resolves #5400

There are a couple outstanding issues (see Known issues section below) but I'd like to get code review kickstarted now.

_Would love to merge this Monday, Oct 29 (or earlier)._

### Merge flow

merge conflict resolution flow

### Abort flow

merge conflict abort flow

### To be fixed in this PR

- [x] merge conflicts popup loads twice sometimes
- [x] handle outstanding TODO in `app.tsx`

### To be fixed in future PRs

I'll convert these into their own issues to be addressed separately

- background updates while popup is displayed
- popups not always centered vertically
- show flow whenever repository is in a conflicted state (ie cover conflicted merges started outside of desktop)

PR #5910, opened 2018-10-16
add createMergeCommit and refactor out commit error handling logic

outofambit:
to support #5809 

I need a way to commit without supplying a string as one is automatically filled in by git, as it does when committing a resolved merge.

Bonuses:
* added some docs
* refactored out specialized error handling from `createCommit` into its own function for use in both commit methods
* converted `git/commit` tests to use `jest`'s `expect` (I needed jest's error throwing matchers for my test)

_I would like to merge this Thursday, October, 18th_

### DIFF
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -48,11 +48,11 @@ export async function createMergeCommit(
   // do the same thing.
   await unstageAll(repository)
   await stageFiles(repository, files)
-  await git(
-    ['commit', '--no-edit'],
-    repository.path,
-    'createMergeCommit'
-  ).catch(logCommitError)
+  try {
+    await git(['commit', '--no-edit'], repository.path, 'createMergeCommit')
+  } catch (e) {
+    logCommitError(e)
+  }
 }
 
 /**
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -64,11 +64,8 @@ function logCommitError(e: Error): void {
   if (e instanceof GitError) {
     const output = e.result.stderr.trim()
 
-    let standardError = ''
-    if (output.length > 0) {
-      standardError = `, with output: '${output}'`
-    }
-    const exitCode = e.result.exitCode
+    const standardError = output.length > 0 ? `, with output: '${output}'` : ''
+    const { exitCode } = e.result
     const error = new Error(
       `Commit failed - exit code ${exitCode} received${standardError}`
     )
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -57,7 +57,8 @@ export async function createMergeCommit(
 
 /**
  * Commit failures could come from a pre-commit hook rejection.
- * So display a bit more context than we otherwise would.
+ * So display a bit more context than we otherwise would,
+ * then re-raise the error.
  */
 function logCommitError(e: Error): void {
   if (e instanceof GitError) {
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -44,21 +44,15 @@ export async function createCommit(
 export async function createMergeCommit(
   repository: Repository,
   files: ReadonlyArray<WorkingDirectoryFileChange>
-): Promise<boolean> {
+): Promise<void> {
   // Clear the staging area, our diffs reflect the difference between the
   // working directory and the last commit (if any) so our commits should
   // do the same thing.
   await unstageAll(repository)
-
   await stageFiles(repository, files)
-
-  try {
-    await git(['commit'], repository.path, 'createMergeCommit')
-    return true
-  } catch (e) {
-    handleCommitError(e)
-    return false
-  }
+  await git(['commit'], repository.path, 'createMergeCommit').catch(
+    handleCommitError
+  )
 }
 
 /**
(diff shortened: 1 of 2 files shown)


app/src/lib/git/commit.ts
```diff
@@ -22,23 +29,57 @@ export async function createCommit(
...
+}
+
+/**
+ * Creates a commit to finish an in-progress merge
+ * assumes that all conflicts have already been resolved
+ *
+ * @param repository repository to execute merge in
+ * @param files files to commit
+ */
+export async function createMergeCommit(
```
shiftkey: It'd be great to get some tests in for this new function while you're sketching out the API (`createCommit` already has some decent coverage).

Ideas that come to mind:
 - happy path testing - repository in conflict, commit files with conflicts, does it succeed? does a merge commit get created?
 - unresolved conflicts - I suspect unstaging all files as part of the operation prevents from occurring, but maybe there's a way for Git to interrupt you?
 - creating a merge commit when not in conflicted state - return false, or throw error?

For higher-level concerns (like ensuring conflict markers are removed) I wouldn't worry about them here, and just let the user proceed.

outofambit: yes, agreed! :) i'm already in the process of adding some tests. will update when its ready.

app/src/lib/git/commit.ts
```diff
@@ -22,23 +29,53 @@ export async function createCommit(
...
```
outofambit: i'm open to better names for this if you have one

app/src/lib/git/commit.ts
```diff
@@ -22,23 +29,53 @@ export async function createCommit(
...
+  repository: Repository,
+  files: ReadonlyArray<WorkingDirectoryFileChange>
+): Promise<void> {
```
outofambit: similar to pattern in #5729

app/src/lib/git/commit.ts
```diff
@@ -4,6 +4,13 @@ import { Repository } from '../../models/repository'
 import { WorkingDirectoryFileChange } from '../../models/status'
 import { unstageAll } from './reset'
 
+/**
+ * Create a commit
```
iAmWillShepherd: How do you feel about not having a comment for well-named functions and just keeping the parameters that need to be explained further?

app/src/lib/git/commit.ts
```diff
@@ -4,6 +4,13 @@ import { Repository } from '../../models/repository'
...
+ *
+ * @param repository repository to execute merge in
+ * @param message commit message
+ * @param files files to commit
+ */
```
iAmWillShepherd: Maybe this could return a string of the commit SHA instead of bool values.

app/src/lib/git/commit.ts
iAmWillShepherd: This could also return a string.

app/src/lib/git/commit.ts
```diff
@@ -22,23 +29,53 @@ export async function createCommit(
...
+export async function createMergeCommit(
+  repository: Repository,
+  files: ReadonlyArray<WorkingDirectoryFileChange>
+): Promise<void> {
+  // Clear the staging area, our diffs reflect the difference between the
+  // working directory and the last commit (if any) so our commits should
+  // do the same thing.
+  await unstageAll(repository)
```
iAmWillShepherd: We could use try/catch here to check all promises instead of the single catch

app/src/lib/git/commit.ts
```diff
@@ -22,23 +29,53 @@ export async function createCommit(
...
-      const error = new Error(
-        `Commit failed - exit code ${exitCode} received${standardError}`
-      )
-      error.name = 'commit-failed'
-      throw error
-    } else {
-      throw e
+    let standardError = ''
+    if (output.length > 0) {
+      standardError = `, with output: '${output}'`
```
iAmWillShepherd: Perhaps we can use a const here.

```ts
const standardError = output.length > 0?  `, with output: '${output}'` : ''
```

app/src/lib/git/commit.ts
outofambit: I really like this idea! I looked into doing this and it requires a bit of extra logic to parse the git output and get the SHA, so I'm going to make this into its own issue for the future. (https://github.com/desktop/desktop/issues/5923)

outofambit:
🍏

app/src/lib/git/commit.ts
shiftkey: Nah, this works for me 👍

app/test/unit/git/commit-test.ts
```diff
@@ -1,15 +1,14 @@
 import * as path from 'path'
 import * as FSE from 'fs-extra'
 
-import { expect } from 'chai'
```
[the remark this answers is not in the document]
outofambit: yes i would love to do that! and then we can eventually 🔥chai 😈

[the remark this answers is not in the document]
shiftkey: How about some ground rules then?

 - if you are writing new tests, use the inbuilt `expect` from `jest`
 - if you are adding tests to an existing module, in a separate commit update the existing usages in tests from `chai` to the inbuilt `expect`

Any other suggestions @desktop/maintainers @desktop/comrades?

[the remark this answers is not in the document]
outofambit: i’m :100: on that. i’ll make an issue once we’ve come to an agreement

PR #5917, opened 2018-10-16
Show actual branch name in "Update From Default Branch" menu item

### LINKED ISSUE #3018: "Update From Default Branch" should use the default branch's name
_From https://github.com/desktop/desktop/issues/3016#issuecomment-335913972_

This is something I've run into too. For me, my eyes look for `master` and don't find it.

a-golovanov:
Fixes #3018

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1458,7 +1458,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
   private getDefaultBranchName(repository?: Repository) {
     const branchesState = this.getBranchesState(repository)
-    if (!branchesState) {
+    if (branchesState == null) {
       return undefined
     }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1424,36 +1424,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
       ? `Open in ${this.selectedExternalEditor}`
       : undefined
 
-    const prLabel = repository
-      ? this.getPullRequestLabel(repository)
-      : undefined
-
     updatePreferredAppMenuItemLabels({
-      editor: editorLabel,
-      pullRequestLabel: prLabel,
-      shell: `Open in ${this.selectedShell}`,
+      editorLabel: editorLabel,
+      pullRequestLabel: this.getPullRequestLabel(repository),
+      shellLabel: `Open in ${this.selectedShell}`,
+      defaultBranchName: this.getDefaultBranchName(
+        repository
+      ),
     })
   }
 
-  private getPullRequestLabel(repository: Repository) {
-    const githubRepository = repository.gitHubRepository
-    const defaultPRLabel = __DARWIN__
-      ? 'Create Pull Request'
-      : 'Create &pull request'
-
-    if (!githubRepository) {
-      return defaultPRLabel
+  private getBranchesState(repository?: Repository) {
+    if (!repository || !repository.gitHubRepository) {
+      return undefined
     }
 
     const state = this.repositoryStateCache.get(repository)
 
     const { branchesState } = state
+    return branchesState
+  }
+
+  private getPullRequestLabel(repository?: Repository) {
+    const branchesState = this.getBranchesState(repository)
+    if (!branchesState) {
+      return undefined
+    }
 
     if (branchesState.currentPullRequest === null) {
-      return defaultPRLabel
+      return undefined
     }
 
     return __DARWIN__ ? 'Show Pull Request' : 'Show &pull request'
+  }
+
+  private getDefaultBranchName(repository?: Repository) {
+    const branchesState = this.getBranchesState(repository)
+    if (!branchesState) {
+      return undefined
+    }
+
+    const defaultBranch = branchesState.defaultBranch
+    return (defaultBranch && defaultBranch.upstreamWithoutRemote) || undefined
   }
 
   private updateRepositorySelectionAfterRepositoriesChanged() {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1462,7 +1462,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return undefined
     }
 
-    const defaultBranch = branchesState.defaultBranch
+    const { defaultBranch } = branchesState
     return (defaultBranch && defaultBranch.upstreamWithoutRemote) || undefined
   }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1445,7 +1445,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
   private getPullRequestLabel(repository?: Repository) {
     const branchesState = this.getBranchesState(repository)
-    if (!branchesState) {
+    if (branchesState == null) {
       return undefined
     }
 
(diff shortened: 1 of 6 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -1424,38 +1424,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-      return defaultPRLabel
+  private getBranchesState(repository?: Repository) {
+    if (!repository || !repository.gitHubRepository) {
+      return undefined
     }
 
     const state = this.repositoryStateCache.get(repository)
 
     const { branchesState } = state
+    return branchesState
```
shiftkey: :art: we can merge this into the previous line to create a one-liner

```ts
return state.branchesState
```

app/src/lib/stores/app-store.ts
```diff
@@ -1424,38 +1424,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  }
+
+  private getPullRequestLabel(repository?: Repository) {
+    const branchesState = this.getBranchesState(repository)
+    if (!branchesState) {
```
shiftkey: We like to be explicit with these sorts of checks, rather than lean on JS truthiness too much

```suggestion
    if (branchesState == null) {
```

app/src/lib/stores/app-store.ts
```diff
@@ -1424,38 +1424,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
     return __DARWIN__ ? 'Show Pull Request' : 'Show &pull request'
   }
 
+  private getDefaultBranchName(repository?: Repository) {
+    const branchesState = this.getBranchesState(repository)
+    if (!branchesState) {
+      return undefined
+    }
+
+    const defaultBranch = branchesState.defaultBranch
```
shiftkey: This one we can use object destructuring to simplify:

```suggestion
    const { defaultBranch } = branchesState
```

app/src/lib/stores/app-store.ts
```diff
@@ -1424,38 +1424,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-      return defaultPRLabel
+      return undefined
     }
 
     return __DARWIN__ ? 'Show Pull Request' : 'Show &pull request'
   }
 
+  private getDefaultBranchName(repository?: Repository) {
+    const branchesState = this.getBranchesState(repository)
+    if (!branchesState) {
```
```suggestion
    if (branchesState == null) {
```

app/src/lib/stores/app-store.ts
```diff
@@ -1424,38 +1424,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      return undefined
+    }
+
+    const defaultBranch = branchesState.defaultBranch
+    return (defaultBranch && defaultBranch.upstreamWithoutRemote) || undefined
```
shiftkey: Similarly, I'd like this to be a bit easier to read:

```ts
if (
  defaultBranch === null ||
  defaultBranch.upstreamWithoutRemote === null
) {
  return undefined
}

return defaultBranch.upstreamWithoutRemote
```

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -16,12 +16,21 @@ const defaultShellLabel = __DARWIN__
...
+  pullRequestLabel?: string
+  defaultBranchName?: string
+}
+
+export function buildDefaultMenu({
+  editorLabel = defaultEditorLabel,
+  shellLabel = defaultShellLabel,
+  pullRequestLabel = defaultPullRequestLabel,
+  defaultBranchName = defaultBranchNameValue,
+}: MenuLabels = {}): Electron.Menu {
```
shiftkey: I think we can do away with the `= {}` usage in here and ensure `main.ts` line 203 passes in an empty object:

```ts
  let menu = buildDefaultMenu({})
```

app/src/lib/stores/app-store.ts
a-golovanov: Is there any policy/recomendation regarding == and === operators?

a-golovanov: Interesting, I would love to know the reasons besides these rules. In any case, I will try to follow them.

app/src/main-process/menu/build-default-menu.ts
a-golovanov: I there any reasoning behind this recommendation besides improved readability? Readability is a subjective thing and sort of hard to follow.

app/src/lib/stores/app-store.ts
a-golovanov: Sorry, I messed up commit description, I'm new to all of this :)

tierninho:

Noticed there is not limit to the truncation in the menu item. This should probably be capped at 25 characters (followed by `...`) to match the truncation we use in the branch list.

screen shot 2018-10-19 at 11 51 26 am

- [x] actual name of default branch is shown in menu item
- [x] changing default name updates the menu item
- [x] user cannot rename default branch
- [x] default branch remains greyed out if user is already on default branch
- [x] keyboard cmd still intact

Build/OS: 1.4.4-beta0, mac, dev 6d847e853143ed40a0bcef7e9b617a0a1c04288e

shiftkey:
@tierninho that's a great find! @a-golovanov are you able to incorporate that?

shiftkey:
> 6d847e853143ed40a0bcef7e9b617a0a1c04288e

@tierninho Just a heads up, this commit ID isn't being linked to in your comment because it doesn't exist on GitHub (so I can't get to it from my machine and "see what you're seeing". For example, I can checkout ddc8145db655dec67b02f55349bf4adfcd6d4296 which is the current tip of the PR and by mentioning it here I see it light up.

app/src/main-process/menu/build-default-menu.ts
shiftkey: We generally try to avoid optional parameters with defaults in TS, because it can be used to obscure what the API is capable of. There's already a fair bit of complexity here around destructuring this parameter and providing default values for it, so I settled on requiring the caller to provide an empty object.

If there were lots of callers that needed to do this I'd likely spend more time figuring out the right API to support them, but as it's one caller (and called exactly once during startup) I was fine with dropping the optional part.

a-golovanov:
> @tierninho that's a great find! @a-golovanov are you able to incorporate that?

I can implement truncation to the fixed number of characters, but I cannot implement it the same way it's done in the branch list. The branch list truncates names with CSS, so the number of visible characters depends on characters widths and width of the list itself.
Should I proceed with truncation to the fixed number of characters even if it will look different comparing to the branch list?

shiftkey:
> Should I proceed with truncation to the fixed number of characters even if it will look different comparing to the branch list?

@a-golovanov this is fine. The branch name should not expand the menu too much, and _how_ the branch list is currently trimmed is a good example. We obviously can't use the same implementation here, but we should try and be consistent.

tierninho:
Truncation UI change = ✅

app/src/lib/truncate-with-ellipsis.ts
```diff
@@ -0,0 +1,32 @@
+/** Truncate a single line unicode string by a given maxLength and add ellipsis if necessary */
+export function truncateWithEllipsis(str: string, maxLength: number) {
```
shiftkey: This seems a great chance to write some tests for this, especially now that we have to be aware of Unicode characters in the string. Are you able to add some tests to your PR using the guide I started in https://github.com/desktop/desktop/pull/5983? I'd love some feedback on if that helps, or what else we can add.

app/src/main-process/menu/build-default-menu.ts
shiftkey: @a-golovanov also, here's some extra context about why we like to avoid these implicit parameters: https://github.com/desktop/desktop/pull/5979#discussion_r227826267

app/src/lib/truncate-with-ellipsis.ts
a-golovanov: I wrote tests, see `app/test/unit/truncate-with-ellipsis-test.ts`. I used `chai` though following other tests as example. I'll migrate my tests to `jest` today.

PR #5924, opened 2018-10-17
Release 1.4.3

outofambit:
- [x] double check release notes
- [x] release

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.4.3-beta1",
+  "version": "1.4.3-beta2",
   "main": "./main.js",
   "repository": {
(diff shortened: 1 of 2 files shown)


billygriffin:
@shiftkey For context, during the release when you were out we talked about standardizing around verbs at the beginning of these just for consistency and shared expectations. I agree that "Fix" after "[Fixed]" is redundant and not the most elegant, and also I think it'd be nice to have a consistent way of presenting what we and our community have done in any given release. We could just standardize around a verb that is not the same as the thing before it (Fix, Improve, Add, etc.). Thoughts?

cc: @nerdneha

shiftkey:
@billygriffin 

> We could just standardize around a verb that is not the same as the thing before it (Fix, Improve, Add, etc.).

This kinda feel like it's duplicating the work that the group headers are there for, and I do worry about being too proscriptive with release notes to the point they end up being dry and boring, but I'm 👍 to figuring out some guidance for these.

PR #5926, opened 2018-10-17
allow Buttons to take a tooltip

outofambit:
_in support of #5809_

just adding a prop and using it for the button `title` :)


app/src/ui/lib/button.tsx
```diff
@@ -16,6 +16,9 @@ export interface IButtonProps {
    */
   readonly onMouseEnter?: (event: React.MouseEvent<HTMLButtonElement>) => void
 
+  /** tooltip text */
```
shiftkey:
```suggestion
  /** An optional tooltip to render when hovering over the button */
```

PR #5931, opened 2018-10-18
 Keyboard selection doesn't work properly in Clone a Repository dialog

### LINKED ISSUE #3596: Arrow keys behaviour unexpected in Clone a Repository dialog
### Description

When cloning a repository, after clicking a repository and using the up/down arrow keys to navigate, the "Local path" text box does not update and continues to show the `C:\....\path` for the originally clicked on repo. Pressing **Clone**, clones the originally clicked repo to the expected path.

### Version

**OS version:** Microsoft Windows [Version 10.0.16299.64]

### Steps to Reproduce

1. File --> Clone repository
1. Click on a repository
1. <kbd>up</kbd>, <kbd>down</kbd>, etc.

githubdesktop-clone-issue

**Expected behavior:** _Local Path_ text box should update to display the correct name of the repository, as it does when clicking the name of the repository

**Actual behavior:** _Local Path_ text box displays the name of the last clicked repository

**Reproduces how often:** 100 %

### Logs

Not considered relevant, can provide if required.

#### Additional Information

None relevant

a-golovanov:
Fixes #3596


tierninho:

However I noticed that a repo will be deselected if you navigate a way from the GitHub.com tab to another tab, and back. The path of the last repo selected remains intact, but the repo is no longer selected. The existing functionality keeps the repo selected. Not sure if this is worth bringing back.

- [x] clicked into repo list and local path updated
- [x] key up/down updates local path
- [x] public/private repos show up in local path field
- [x] destination already exist warning still works as intended
- [x] local path field updates if user clicks search result
- [ ] Enterprise repos

Build/OS: 1.4.3-beta0, mac, dev 41e2b2a7ba78a1982bb9c889d5509ba18fd614cd

a-golovanov:
I looked at the latest release (v1.4.3) and it behaves exactly the same as a local build with my changes when switching between tabs. Namely, if a repository is selected on the GutHub.com tab, switching to any other tab and back removes the selection from the list, but preserves local path and allows to clone the previously selected repository. In my opinion, it is a bug. Either the selection in the list should be restored or local path should be reset and clone button disabled.

Screenshot from v1.4.3:
screen shot 2018-10-19 at 7 18 50 pm

I will look what I can do about that.

a-golovanov:
Another possible solution for selection reset on tab switching is rendering all tabs al the time, as opposit to current implementation that renders only the selected tab.

PR #5940, opened 2018-10-18
codified section about where to start a release from

shiftkey:
@outofambit and I had a chat earlier about where we land on the "release from master" or "release from a branch" debate.

I think our workflow from the last few updates should be codified to indicate that the release owner should cut a branch from `master` which is then released to users, rather than having options.

This has a few benefits:

 - other maintainers can continue to review and merge PRs once that release branch is cut, knowing it won't be shipped to users by accident
 - if a hotfix needs to be expedited after the release branch is cut, a maintainer can target the release branch and work with the release owner to get it included (if the release owner is on board)
 - release branch is only opened to generate the release notes - so it's not a long-lived branch and should not be involved with typical workflows

### DIFF
--- a/docs/process/releasing-updates.md
+++ b/docs/process/releasing-updates.md
@@ -118,20 +118,19 @@ Here's an example of the previous changelog draft after it has been edited:
 }
 ```
 
-Once you've pushed up the version update and the changelog changes, you're ready to release! Get the others on the team to :thumbsup: in a PR if you're not sure. Note that any version change that does not have an associated changelog entry will not successfully release.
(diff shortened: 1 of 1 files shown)


docs/process/releasing-updates.md
```diff
@@ -118,20 +118,19 @@ Here's an example of the previous changelog draft after it has been edited:
 }
 ```
 
-Once you've pushed up the version update and the changelog changes, you're ready to release! Get the others on the team to :thumbsup: in a PR if you're not sure. Note that any version change that does not have an associated changelog entry will not successfully release.
+Create a new branch off `master` to represent the work that will be released to users. Add your new changelog entries and version update, and push this branch to GitHub. This becomes the release branch, and lets other maintainers continue to merge into `master` without affecting your release.
```
outofambit: should we also add here that for `production` releases, it should be branched off of the last `beta` release branch?

PR #5948, opened 2018-10-19
use remote displayed in the push/pull button when performing the push

### LINKED ISSUE #5936: can't push to repository in app (publish repository modal displayed)
## Description
I updated my system from High Sierra to Mojave. Upon update, I lost the ability to publish.

When I'm looking at the Desktop app, I have the "Push" command for my branch in my repository.
screen shot 2018-10-18 at 1 44 24 pm

However, when I click this: 
screen shot 2018-10-18 at 1 49 56 pm

What I have been doing to complete my work is commit here, then use the command line to 
`git push origin marketplace_product_sort`

It seems to be all branches on this repository - at least this is the second branch I've come across.

I have been going through a few issues here but they were all relegated to Windows 10 issues, and they didn't seem to quite match. If I am wrong I'll happily be pointed to a good issue to follow.

## Version

* GitHub Desktop: 1.4.2

* Operating system:
Mac OSX Mojave 10.14

## Steps to Reproduce
1. Click `Push Origin`
2. Get `Publish?` box
3. Close box without publishing, go to CLI
4. `git push origin my_branch`
5. Branch pushes correctly without errors

### Expected Behavior
Github desktop pushes normally like it used to

### Actual Behavior
Refuses to push

### Logs

[2018-10-18.desktop.production.log](https://github.com/desktop/desktop/files/2493044/2018-10-18.desktop.production.log)

shiftkey:
Closes #5936

I haven't been able to reproduce the bug exactly but the reporter gave me enough details to isolate what I believe is the root cause - a mixup with how we track remotes. 

The push/pull button uses `IRepositoryState.remote` field when displaying it's `Push origin` or `Pull origin` message, but inside `_push` we look at `GitStore.remote`, and the user is being asked to publish the repository because it's set to `null`.

This PR now uses the same state as the push/pull button when performing the push, so it should eliminate the confusing behaviour the user is experiencing. I also tidied up some of the repetition in there around checking `tip`.


shiftkey:
I added some comments in https://github.com/desktop/desktop/issues/5936#issuecomment-431377914 about the underlying behaviour, if you want more context.

shiftkey:
@say25 thanks for the :eyes!

> Unit Tests

For the moment I'm not going to wrestle with getting `AppStore` under test for this scenario, but we're tracking it's subpar test coverage so once we get to a better spot we can look at improving our test coverage there.

> Is it worth covering the unlikely condition `tip.kind === TipState.Unknown`?

I think `TipState.Unknown` is only set when initializing a `GitStore` or when `git status` doesn't give us the expected output, so I'm fine with continuing to do nothing here because, particularly if `git status` fails, other parts of the app will have greater problems to handle.

say25:
@shiftkey makes perfect sense to me. Just things to consider not necessarily do. You are clearly making the code better. Just wanted to make sure someone with more expertise thought about it.

PR #5950, opened 2018-10-19
remove duplicate implementation comparing GitHubRepository and remote

shiftkey:
`findUpstreamRemote` has it's own rules for comparing a `GitHubRepository` to a `IRemote`, which looks very similar to something we have in `repository-matching.ts`:

https://github.com/desktop/desktop/blob/0fc2294c119d0186e2d5db75ce74174233d57924/app/src/lib/repository-matching.ts#L99-L113

And `urlMatchesRemote` is just below in the same module:

https://github.com/desktop/desktop/blob/0fc2294c119d0186e2d5db75ce74174233d57924/app/src/lib/repository-matching.ts#L122-L149

This other implementation has some improvements over `findUpstreamRemote`:

 - skip throwing an error on `forceUnwrap` to workaround type checking
 - use the `htmlUrl` value as an fallback check
 - use `caseInsensitiveCompare` to avoid sprinkling `.toLowerCase()` around

We use this to find the pull request associated with the current branch, so I feel pretty confident that it's already working as intended (for PRs from forks and from local branches).


shiftkey:
🍎

Issue #5438, opened 2018-08-19
error: cannot lock ref 'refs/remotes/origin/


natros:
It only happens a few times

```
error: cannot lock ref 'refs/remotes/origin/jetty-9.4.x': is at d8e0f8df1206536612c938b63ae677d2352de14b but expected ddd80deeb96eed136029b944258b36b285af79c6
From github.com:eclipse/jetty.project
 ! ddd80deeb9..d8e0f8df12  jetty-9.4.x -> origin/jetty-9.4.x  (unable to update local ref)
 * [new branch]            jetty-9.4.x-2828-http2_connection_listener_contention -> origin/jetty-9.4.x-2828-http2_connection_listener_contention
```

![image](https://user-images.githubusercontent.com/4996289/44312340-66f9f000-a3ee-11e8-9dc0-7b4f54921409.png)

Version 1.3.3
$ git --version
git version 2.18.0
macOS 10.13.6

zhzoo:
On windows, happens to me nearly every time.
Same exact error as you:
error: cannot lock 'ref/remotes/origin/blahblahblah': is at something but expected somethingelse

Hacky workaround I've been doing is a combination of:
running: git gc --prune=now
and deleteing .git/refs/remote/origin

iAmWillShepherd:
Thanks for taking the time to open an issue! 🙇 

This error is often a symptom of more than one process trying to act on the same repository at the same time.

@natros could you verify if you are using other tools to manage your repository while Desktop is open (some IDEs and text editors have git plugins that may access your repository)? 

cc @desktop/support

natros:
hi,

i'm not using other tools, just using this to keep up to date with last changes from a few repositories. the one that most often is causing me problems is this repo:

```
(jetty-9.4.x) git remote -v
origin	git@github.com:eclipse/jetty.project.git (fetch)
origin	git@github.com:eclipse/jetty.project.git (push)
```

and this happens right now
```
error: cannot lock ref 'refs/remotes/origin/jetty-9.4.x': is at 5b5f2fcf5f8241dd9edb998b97ad5e6cd45130a8 but expected 33a1367325530ec38254a279abb30946505af485
From github.com:eclipse/jetty.project
 ! 33a1367325..5b5f2fcf5f  jetty-9.4.x -> origin/jetty-9.4.x  (unable to update local ref)
[...]

PR #5952, opened 2018-10-19
run status without optional locks for all users

shiftkey:
We rolled this out in `release-1.4.0-beta1` and I've not heard any problems with it from our beta users, so I'd like to propose rolling this out more broadly in a future update.

I see outstanding issues like #5438 and #4938 which are `priority-2` and I believe will be aided by this change (I often hit this when trying to checkout a branch or a PR, and I haven't seen this on `beta` in a while).

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -54,5 +54,5 @@ export function enableInAppReleaseNotes(): boolean {
 
 /** Should `git status` use --no-optional-locks to assist with concurrent usage */
 export function enableStatusWithoutOptionalLocks(): boolean {
-  return enableBetaFeatures()
+  return true
 }


shiftkey:
Some extra context: I tried to open a PR branch inside the app using `1.4.3` but repeatedly hit this error:

```
2018-10-22T18:55:45.371Z - error: [ui] `git -c credential.helper= checkout --progress pr/5822 --` exited with an unexpected code: 128.
fatal: Unable to create '/Users/shiftkey/src/desktop/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.

(The error was parsed as 34: A lock file already exists in the repository, which blocks this operation from completing.)
```

After the sixth attempt I downloaded `1.4.4-beta0` and was able to switch the first time.

PR #5973, opened 2018-10-23
Repository name is now sanitized on publish

### LINKED ISSUE #3090: Block emoji from being entered as a repo name
Emoji shouldn't be allowed in place of a repo name as they are converted anyway to "normal" characters by GitHub.com and Desktop once published.

If you already have a repo with the name`github.com/tierninho/-1` and you decide to create a new repo with an emoji 🐴  as the repo name, the name becomes converted to  "normal" characters as expected. However, if you click "open in GitHub" or "create PR", the existing repo is opened despite the real url being `github.com/tierninho/-`

screen shot 2017-10-18 at 2 34 22 pm
screen shot 2017-10-18 at 2 35 27 pm

Just remove the Emoji option from the Edit menu?

tested 1.0.5beta0, mac

Daniel-McCarthy:
Fixes #3090. This is in response to an issue reporting the ability to enter characters such as emojis as the repository name in the `Publish Repository` dialog. Although the `Create Repository` dialog sanitizes the user entered name to filter out illegal characters, the `Publish Repository` dialog gives an opportunity to modify the name of to be published without sanitization, giving an opportunity to enter illegal characters.

To solve this, I have included the pre-existing repository sanitization functionality in the `Publish Repository` dialog. This is both in the form of a warning that displays what the new name will be, and the dialog using the new sanitized name instead.
___
The new functionality can be seen here:
peek 2018-10-04 15-07
___
Additionally, while testing these changes I noticed that the sanitization function will replace an emoji such as 🐓 with two dashes instead of one. I am not quite sure why this is the case currently. It is somewhat unrelated, but I figured I would mention it.

### DIFF
--- a/app/src/ui/publish-repository/publish-repository.tsx
+++ b/app/src/ui/publish-repository/publish-repository.tsx
@@ -7,6 +7,8 @@ import { DialogContent } from '../dialog'
 import { Row } from '../lib/row'
 import { merge } from '../../lib/merge'
 import { caseInsensitiveCompare } from '../../lib/compare'
+import { sanitizedRepositoryName } from '../add-repository/sanitized-repository-name'
+import { Octicon, OcticonSymbol } from '../octicons'
 
 interface IPublishRepositoryProps {
   /** The user to use for publishing. */
@@ -150,6 +152,8 @@ export class PublishRepository extends React.Component<
           />
         </Row>
 
+        {this.renderSanitizedName()}
+
         <Row>
           <TextBox
             label="Description"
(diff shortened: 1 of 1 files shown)


j-f1:
> Additionally, while testing these changes I noticed that the sanitization function will replace an emoji such as 🐓 with two dashes instead of one. I am not quite sure why this is the case currently. It is somewhat unrelated, but I figured I would mention it.

That’s probably due to the fact that JS uses UTF-16, which represents emoji and certain other characters as two characters. Try `'🐓 '.length` and:

```js
const s = '🐓'
for (let i = 0; i < s.length; i++) {
  console.log(s.charCodeAt(i))
}
```

Daniel-McCarthy:
@j-f1 Ah, fantastic point!
When I first started looking into this issue I had played around with removing utf16 emoji characters specifically until I realized that the current sanitization function already removes such characters. But perhaps if we wanted to make sure an emoji was replaced with one `-` like on GitHub.com, we could have the regex specifically detect UTF 16 codes at whatever range is necessary so that it matches the whole UTF16 character. Though the area that would be allowed would be very small compared to the rest. There's likely a simpler option.

outofambit:
we can file a separate issue for fixing the `--` issue, since that's pre-existing

shiftkey:
@outofambit I'm fine with pulling this into `1.4.4`

tierninho:
testing done for this, looks good ⚡️

PR #5977, opened 2018-10-23
Add a merge button to the branch selector

### LINKED ISSUE #5929: Add a button to the branch selector
It's not obvious how to merge one branch into another in Desktop.

:warning: See #5555 for additional context :warning: 

This issue is marked as help wanted to gives others the chance to get involved in feature iteration. Please leave a comment below if you'd like to pick this up 😄.

### Goal

We want to improve the discoverability of merging branches. At the moment, we have 2 ways of doing this. 

1. The top-level menu: **Branch -> Merge Into Current Branch**
screen shot 2018-10-17 at 4 05 38 pm

1. Initiating a branch comparison

### Proposed Solution

Adding a button to the *Branch Selector* should bring the merge action closer to where most would expect it to be. This button will behave just like the menu button and bring up the _Merge Branch_ dialog where a user can select the branch they'd like to merge.

**Mockup**
46970036-c120d580-d06c-11e8-9fd9-31a54974977c

bruncun:
Fixes #5929

Morning! This is an early PR for #5929 as requested by @iAmWillShepherd. So far it visually mocks out the new merge button - the following still needs to be done:

- [x] Interpolate current branch name into button label
- [x] Render button only in Branches list (currently also renders in Repositories list and merge dialog)
- [x] Wire button by duplicating 'Merge into Current Branch' Menu Item logic
- [x] Identify and prevent edge cases where button ought to be disabled or hidden (e.g. when selected branch matches current branch)
- [x] Tests?

My plan is to complete the above in order and just get it working before reviewing for potential refactors. Please let me know if I missed anything or whether I should go about this feature differently somehow.

### DIFF
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -115,6 +115,7 @@ export class BranchesContainer extends React.Component<
             defaultBranch={this.props.defaultBranch}
             currentBranch={this.props.currentBranch}
             allBranches={this.props.allBranches}
+            isBranchListDropdown={true}
             recentBranches={this.props.recentBranches}
             onItemClick={this.onBranchItemClick}
             filterText={this.state.branchFilterText}
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -76,7 +76,7 @@ export class BranchesContainer extends React.Component<
         {this.renderTabBar()}
         {this.renderSelectedTab()}
         <Row className="merge-button-row">
-          <Button className="merge-button">
+          <Button className="merge-button" onClick={this.onMergeClick}>
             <Octicon className="icon" symbol={OcticonSymbol.gitMerge} />
             <span title={`Commit to ${branchName}`}>
               Choose a branch to merge into <strong>{branchName}</strong>
@@ -197,6 +197,13 @@ export class BranchesContainer extends React.Component<
 
   private onDismiss = () => {
     this.props.dispatcher.closeFoldout(FoldoutType.Branch)
+  }
+
+  private onMergeClick = () => {
+    this.props.dispatcher.showPopup({
+      type: PopupType.MergeBranch,
+      repository: this.props.repository,
+    })
   }
 
   private onBranchItemClick = (branch: Branch) => {
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -12,6 +12,10 @@ import { assertNever } from '../../lib/fatal-error'
 
 import { TabBar } from '../tab-bar'
 
+import { Row } from '../lib/row'
+import { Octicon, OcticonSymbol } from '../octicons'
+import { Button } from '../lib/button'
+
 import { BranchList } from './branch-list'
 import { PullRequestList } from './pull-request-list'
 import { PullRequestsLoading } from './pull-requests-loading'
@@ -63,10 +67,22 @@ export class BranchesContainer extends React.Component<
   }
 
   public render() {
+    const branchName = this.props.currentBranch
+      ? this.props.currentBranch.name
+      : 'master'
+
     return (
       <div className="branches-container">
         {this.renderTabBar()}
         {this.renderSelectedTab()}
+        <Row className="merge-button-row">
+          <Button className="merge-button">
+            <Octicon className="icon" symbol={OcticonSymbol.gitMerge} />
+            <span title={`Commit to ${branchName}`}>
+              Choose a branch to merge into <strong>{branchName}</strong>
+            </span>
+          </Button>
+        </Row>
       </div>
     )
   }
@@ -115,7 +131,6 @@ export class BranchesContainer extends React.Component<
             defaultBranch={this.props.defaultBranch}
             currentBranch={this.props.currentBranch}
             allBranches={this.props.allBranches}
-            isBranchListDropdown={true}
             recentBranches={this.props.recentBranches}
             onItemClick={this.onBranchItemClick}
             filterText={this.state.branchFilterText}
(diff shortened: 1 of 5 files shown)


iAmWillShepherd:
@bruncun, the build is failing due to linter errors. Run `yarn lint:fix` on your machine and commit the changes it makes to get the build to pass.

bruncun:
@iAmWillShepherd Oops, sorry. 😅 Done.

app/src/ui/branches/branches-container.tsx
```diff
@@ -115,6 +115,7 @@ export class BranchesContainer extends React.Component<
             defaultBranch={this.props.defaultBranch}
             currentBranch={this.props.currentBranch}
             allBranches={this.props.allBranches}
+            isBranchListDropdown={true}
```
iAmWillShepherd: You can actually move the button write after this component. Doing that will eliminate the need to pass the `isBranchListDropDown` prop to the `BranchList` component. 

https://github.com/desktop/desktop/blob/9aaf2a1d0ae39ea0500dfa4b5922db2d22a67b7e/app/src/ui/branches/branches-container.tsx#L112-L128

app/src/ui/lib/filter-list.tsx
```diff
@@ -1,11 +1,15 @@
 import * as React from 'react'
 import * as classnames from 'classnames'
 
+import { Branch } from '../../models/branch'
```
iAmWillShepherd: Let's keep this component generic. This component allows us to filter any list of items, not just branches. All of the state you need access to should be available from within the `BranchesContainer` component.

tierninho:
Two suggestions after taking an early look at this branch:

- Remove the button if there is only one branch or just grey it out
- The amount of white space looks odd... not sure if floating it up will look any better
screen shot 2018-10-24 at 9 09 43 am

billygriffin:
@desktop/design Would you mind taking a quick look at @tierninho's thoughts/questions above?

My thoughts: 

> Remove the button if there is only one branch or just grey it out

I definitely don't think we should remove it because I think it's important to convey an expectation that this is where they can merge from (like if they use multiple repos). It's not greyed out in the menu bar at present, so whichever direction we go, I think we should do the same thing in both places. I think greying it out in both places seems reasonable, as there's really nothing you can do in the modal at that point.

> The amount of white space looks odd... not sure if floating it up will look any better

I agree that it looks a bit odd in the case when there are no branches or very few, but again I think the consistency of location might be a good idea to maintain, and I also don't think it floating would look very good unless we decided to change the visual representation of the entire dropdown to only drop vertically by the amount of space needed (I don't think we should consider that here).

ampinsk:
> I definitely don't think we should remove it because I think it's important to convey an expectation that this is where they can merge from (like if they use multiple repos). It's not greyed out in the menu bar at present, so whichever direction we go, I think we should do the same thing in both places. I think greying it out in both places seems reasonable, as there's really nothing you can do in the modal at that point.

👍 Agree with this. Maybe we leave it in case people are curious and want to get a sense of what clicking on it would do? Disabling it might be frustrating especially if it's not clear why it's disabled. I don't feel strongly though, would also be cool with disabling it

> I agree that it looks a bit odd in the case when there are no branches or very few, but again I think the consistency of location might be a good idea to maintain, and I also don't think it floating would look very good unless we decided to change the visual representation of the entire dropdown to only drop vertically by the amount of space needed (I don't think we should consider that here).

👍 Also agree here.

bruncun:
@iAmWillShepherd I refactored the button into BranchesContainer as requested and also wired it to show the merge dialog. I believe this now works for the happy path.

billygriffin:
Based on @ampinsk's feedback above, let's go ahead and leave those things as is from @tierninho's questions.

iAmWillShepherd:
> Based on @ampinsk's feedback above, let's go ahead and leave those things as is from @tierninho's questions.

@billygriffin does that mean the PR is fine as is? Is it worth create issues to capture y'alls thoughts so we may evaluate them in the future? I'm happy to tackle those changes in a separate PR.

@bruncun it looks to me like this may be ready for review. Am I correct?

tierninho:
UI tested ✅, and no issues found.

billygriffin:
> @billygriffin does that mean the PR is fine as is? Is it worth create issues to capture y'alls thoughts so we may evaluate them in the future? I'm happy to tackle those changes in a separate PR.

@iAmWillShepherd It means the things that @tierninho brought up we're going to leave as is for the moment, and if they present problems in the wild we'll re-evaluate. No issues needed at the moment, but thank you! I haven't looked at the rest of the PR, as I assume the review will ensure that it's in line with the designs from the issue and verify that the flow works as expected.

bruncun:
@iAmWillShepherd I believe that’s correct. 🎉

app/src/ui/branches/branches-container.tsx
```diff
@@ -63,10 +67,22 @@ export class BranchesContainer extends React.Component<
   }
 
   public render() {
+    const branchName = this.props.currentBranch
+      ? this.props.currentBranch.name
+      : 'master'
```
iAmWillShepherd:
```suggestion
      : this.props.defaultBranch || 'master'
```

app/src/ui/merge-branch/merge.tsx
```diff
@@ -33,6 +33,9 @@ interface IMergeProps {
    */
   readonly currentBranch: Branch | null
 
+  /**  */
+  readonly isBranchListDropdown?: boolean
```
iAmWillShepherd: This can be deleted.

iAmWillShepherd: Same with the newline that follows.

bruncun:
@iAmWillShepherd Updated. 👍

billygriffin:
Hi @bruncun, thanks again for this great work and being so open to feedback and collaborative! For our 1.5 release one of the things we want to highlight is the contributions from our community and how grateful we are for them. 

Would you mind if we use your GitHub handle and reference your participation in helping make this release possible in our blog post announcing the release?

bruncun:
@billygriffin The pleasure was mine. I don't mind at all being mentioned. 👍

PR #5984, opened 2018-10-24
sanitized-repository-name: escape emojis

### LINKED ISSUE #5980: sanitizedRepositoryName replaces emoji with two dashes
## Description
Discovered in #5973 

`sanitizedRepositoryName` replaces emoji with `--` instead of `-`

## Steps to Reproduce

reproduction steps

### Expected Behavior
emoji is replaced with `-`

### Actual Behavior
emoji is replaced with `--`

## Additional Information

https://github.com/desktop/desktop/pull/5973#issuecomment-432193334
https://github.com/desktop/desktop/pull/5973#issuecomment-432238271

WaleedAshraf:
Closes #5980

### DIFF
--- a/app/src/ui/add-repository/sanitized-repository-name.ts
+++ b/app/src/ui/add-repository/sanitized-repository-name.ts
@@ -3,8 +3,10 @@ export function sanitizedRepositoryName(name: string): string {
   const ranges = [
     '\ud83c[\udf00-\udfff]', // U+1F300 to U+1F3FF
     '\ud83d[\udc00-\ude4f]', // U+1F400 to U+1F64F
-    '\ud83d[\ude80-\udeff]'  // U+1F680 to U+1F6FF
+    '\ud83d[\ude80-\udeff]', // U+1F680 to U+1F6FF
   ]
 
-  return name.replace(new RegExp(ranges.join('|'), 'g'), '-').replace(/[^\w.-]/g, '-')
+  return name
+    .replace(new RegExp(ranges.join('|'), 'g'), '-')
+    .replace(/[^\w.-]/g, '-')
 }
--- a/app/src/ui/add-repository/sanitized-repository-name.ts
+++ b/app/src/ui/add-repository/sanitized-repository-name.ts
@@ -1,4 +1,10 @@
 /** Sanitize a proposed repository name by replacing illegal characters. */
 export function sanitizedRepositoryName(name: string): string {
-  return name.replace(/[^\w.-]/g, '-')
+  const ranges = [
+    '\ud83c[\udf00-\udfff]', // U+1F300 to U+1F3FF
+    '\ud83d[\udc00-\ude4f]', // U+1F400 to U+1F64F
+    '\ud83d[\ude80-\udeff]'  // U+1F680 to U+1F6FF
+  ]
+
+  return name.replace(new RegExp(ranges.join('|'), 'g'), '-').replace(/[^\w.-]/g, '-')
 }
--- a/app/src/ui/add-repository/sanitized-repository-name.ts
+++ b/app/src/ui/add-repository/sanitized-repository-name.ts
@@ -1,12 +1,11 @@
+const ranges = [
+  '\ud83c[\udf00-\udfff]', // U+1F300 to U+1F3FF
+  '\ud83d[\udc00-\ude4f]', // U+1F400 to U+1F64F
+  '\ud83d[\ude80-\udeff]', // U+1F680 to U+1F6FF
+]
+const emojiRegExp = new RegExp(ranges.join('|'), 'g')
+
 /** Sanitize a proposed repository name by replacing illegal characters. */
 export function sanitizedRepositoryName(name: string): string {
-  const ranges = [
-    '\ud83c[\udf00-\udfff]', // U+1F300 to U+1F3FF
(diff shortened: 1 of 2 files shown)


app/src/ui/add-repository/sanitized-repository-name.ts
```diff
@@ -1,4 +1,12 @@
...
 export function sanitizedRepositoryName(name: string): string {
-  return name.replace(/[^\w.-]/g, '-')
+  const ranges = [
+    '\ud83c[\udf00-\udfff]', // U+1F300 to U+1F3FF
+    '\ud83d[\udc00-\ude4f]', // U+1F400 to U+1F64F
+    '\ud83d[\ude80-\udeff]', // U+1F680 to U+1F6FF
+  ]
+
+  return name
+    .replace(new RegExp(ranges.join('|'), 'g'), '-')
```
iAmWillShepherd: How do you feel about moving `ranges` and the creation of the `RegExp` object to outside of this function?

```ts
const ranges = []
const emojiRE = new RegExp(ranges.join('|'), 'g')

export function sanitizedRepositoryName(name: string): string {
}
```

Doing it this way means we won't create the `RegExp` object every time we run this function.

WaleedAshraf: Done

iAmWillShepherd:
Thanks for your contribution @WaleedAshraf, this should get merged tomorrow ❤️

tierninho:

- [x] each emoji is sanitized with a single dash
- [x] emoji can still be entered in name field (initially)
- [x] sanitized repo pushed to GitHub.com; shows a single dash

Build/OS: 1.4.4-beta0, mac, dev 1c048184dd07456788ca998feea18e0fffea4a18

iAmWillShepherd:
@WaleedAshraf would you like to add tests to get this over the line or would you prefer if I or someone else on @desktop/engineering add some tests to your branch? I personally don't mind either way 👍

WaleedAshraf:
Sorry, I got busy.
If this can wait 2,3 days more, I'll test cases.

Otherwise, I'm also okay if someone else wants to contribute.

WaleedAshraf:
@shiftkey thanks for feedback. I'll look into test docs. 👍

WaleedAshraf:
I need a few more days. Will be done by this weekend.

shiftkey:
@WaleedAshraf no problem!

WaleedAshraf:
@shiftkey @iAmWillShepherd 
There is something weird about one test scenario. I'll share input/output of actual function I wrote vs when using through Desktop App.

**Function:**
```
sanitizedRepositoryName('Hello\r\nWorld')
// "Hello--World"


sanitizedRepositoryName('Hello\\r\\nWorld')
// "Hello-r-nWorld"
```

**Github Desktop (testing on my branch - local)**

```
Name: "Hello\r\nWorld"
// Will be created as Hello-r-nWorld


Name: "Hello\\r\\nWord"
// Will be created as Hello--r--nWorld
```


To my understanding there is difference when input is given from **code/testcase** VS **Desktop App**. Special characteres are being treated differently.

Because of this different, test input values are not reflecting actual input. (kind of enviroment difference) between App and tests.

j-f1:
That’s because the text is escaped in the app. Typing “Hello\r\nWorld” will result in `'Hello\\r\\nWorld'` in the code.

shiftkey:
@WaleedAshraf @j-f1 not sure I 100% follow but I'll try and explain things from my side.

With JS strings you can embed newlines using the `\r` and/or `\n` character. In this situation the backtick `\` is escaping the `r` and `n` characters, because that's easier than passing in the `0D0A` bytes to represent `CRLF` in ASCII.

How the string is created is really important:

 - if you create the string directly in the console (i.e. through the Chrome Dev Tools) you can do whatever you want - including passing in raw characters



 - If you try and create this in an `input` element, it will be handled differently as the `\` and `n` are considered distinct characters. The JS string will be encoded like this `hello\\r\\n` and the first `\` is used to escape the second `\`. That's where the duplication is coming in.



 - If you try and press <kbd>Enter</kbd> in the `input` you'll likely trigger a form submit, or if you can't submit the form I suspect the character is ignored.

For reference, we had a PR merged a while ago https://github.com/desktop/desktop/pull/3796 that improved branch names with invalid characters - previously it would generate a dash for each invalid character, but with the change it would fold them up into one to make the sanitized branch name a bit more readable. Maybe that's something we need to do here too?

WaleedAshraf:
@shiftkey I get your point.
Yes, I followed `sanitizedBranchName` tests as a base for these. And in those tests, new lines are being treated differently. That's because we are using a regex in `sanitizedBranchName` function.

I'm not sure if we need it here? To me, this is okay as it is right now. We can add more checks in regex, later if needed.

Your Call.

WaleedAshraf:
I'm not a big fan of regex. Can make things super complex with minor mistakes. So just trying to avoid.

shiftkey:
@WaleedAshraf I'm fine with leaving it as-is - we have a test that verifies the original issue is resolved, and we have additional tests that demonstrate the behaviour, so we can refine what this function does with actual examples to discuss.

shiftkey:
I'll see if @iAmWillShepherd can give it a 👍 but this seems ideal to land for 1.5.0. Thanks for the contribution @WaleedAshraf!

PR #5991, opened 2018-10-25
Fix/focus counter clockwise cycling

### LINKED ISSUE #3596: Arrow keys behaviour unexpected in Clone a Repository dialog

a-golovanov:
Deactivate codemirror’s invisible focus, fixes #3596

Note: it’s no longer possible to scroll diff text with keyboard, but that feature was dubious since user didn’t see the cursor and the scrolling would start only after the invisible cursor reached the edge of the view.


shiftkey:
@a-golovanov thanks for opening this up! For reference, this is part of the Codemirror [manual](https://codemirror.net/doc/manual.html):



> Note: it’s no longer possible to scroll diff text with keyboard, but that feature was dubious since user didn’t see the cursor and the scrolling would start only after the invisible cursor reached the edge of the view.

I'm hesitant to take this in as-is without investigating alternative approaches, because this feels like something that will hurt the accessibility of the app. If we could at least show that the Codemirror editor has focus when you tab in counter-clockwise direction (maybe using `FocusContainer`?) to ensure tab order is visible, maybe then we can then dig into how we can restore controlling the diff with the keyboard.

Assigning myself to dig into this further.

niik:
> I'm hesitant to take this in as-is without investigating alternative approaches, because this feels like something that will hurt the accessibility of the app

I agree. I think this is just a matter of telling code mirror not to deal with the 'Shift+Tab' combo just [as we've done with the 'Tab' key](https://github.com/desktop/desktop/blob/6cfa767416cd5191523c7abb826621d5b1e0d273/app/src/ui/diff/text-diff.tsx#L728). An addition of `'Shift-Tab': false` to the `extraKeys` property should do the trick I believe.

PR #5994, opened 2018-10-25
help improve our release notes by writing them earlier

shiftkey:
## Overview

A suggestion to improve the workflow for new contributions

## Description

This adds a new section to the pull request template (inspired by the Electron repository) to help curate better release notes earlier in the process.

Why?

 - our release notes generator currently just dumps a list of every PR that was merged - I think we can distinguish better between "this PR should be in the release notes" and "this change doesn't affect end users"
 - writing the release notes later means important context may be forgotten, and makes writing the release notes a more time-consuming task than it should be
 - our release notes generator looks in the PR body for linked issues, so it's easy for us to add an additional lookup for this `notes: [blah]` line to decide whether or not to omit it

For the moment I'm going to keep the `infrastructure` around to help us track reviews unrelated to user-facing work, but I'm pondering on different ways to prioritize reviews (maybe the granularity should be on what the PR is affecting - `docs`/`infrastructure`/`tech-debt`/???)

## Release notes

Notes: no-notes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,3 +11,13 @@ If an issue doesn't exist for this pull request (PR) to address, please open one
 ## Description
 
 -
+
+## Release notes
+
+<!--
+If this is related to a user-facing change and should be added to the release notes, please add a summary of the change.
+
+For changes related to documentation or techical debt, just add `no-notes` to indicate this can be skipped.
+-->
+
+Notes:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,7 +15,7 @@ If an issue doesn't exist for this pull request (PR) to address, please open one
 ## Release notes
 
 <!--
-If this is related to a feature, bugfix or improvement, please add a summary of the change so it can be featured in the releaase notes.
+If this is related to a feature, bugfix or improvement, please add a summary of the change so it can be featured in the release notes.
 
 If you don't believe this change needs to be mentioned in the release notes, write "no-notes" to indicate this can be skipped.
 -->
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -11,3 +11,13 @@ If an issue doesn't exist for this pull request (PR) to address, please open one
 ## Description
 
 -
+
+## Release notes
+
+<!--
+If this is related to a feature, bugfix or improvement, please add a summary of the change so it can be featured in the releaase notes.
```
billygriffin:
```suggestion
If this is related to a feature, bugfix or improvement, please add a summary of the change so it can be featured in the release notes.
```

shiftkey:
> Is the expectation here that external contributors will write what they'd like to see included in release notes alongside their feature?

I'd like them to take a shot at it. Things like PR titles or even issue titles (which we currently rely on generating the release notes entry) aren't the best source for these either.

> Should we maybe include an example?

Maybe? I'll try it out and see how it feels.

shiftkey:
:apple:

j-f1:
How should the changelog entries be formatted grammatically? Like a command (“Fix bug X”) or a description (“Fixes bug X”)?

shiftkey:
> How should the changelog entries be formatted grammatically? Like a command (“Fix bug X”) or a description (“Fixes bug X”)?

Excellent question, and something we've wrestled with for a while. Using the examples I borrowed and adapted from existing changelog entries I ended up writing them in present tense, and as a description, because it felt natural and a concise way to describe the change.

The only downside to this was I found myself leaning on the grouping labels (Adds/Fixes/Improves/etc) to start the entry, as it's something I think we should avoid duplicating in the release entry text that ends up in the changelog.

But I'm open to a different approach here if there's other ideas.

iAmWillShepherd:
Merging since there is consensus.

PR #5995, opened 2018-10-25
cleanup metadata support to help with providing support for merge conflicts

shiftkey:
This PR moves the error handling for merge conflicts back into `dispatcher.ts`, and makes some changes to the internals to help support the scenario needed in #5809 about being able to pass additional metadata to the dialog we need to show.

I also refreshed the relevant docs section to indicate This Is How We Should Pass Information To Error Handlers™ In The Future™:

 - [x] update the `pull` error handler to also provide this required metadata

### DIFF
--- a/app/src/lib/dispatcher/error-handlers.ts
+++ b/app/src/lib/dispatcher/error-handlers.ts
@@ -295,21 +295,17 @@ export async function mergeConflictHandler(
       break
   }
 
-  if (gitContext.kind === 'merge') {
-    const { tip, branch } = gitContext
-    if (tip == null || tip.kind !== TipState.Valid) {
-      return error
-    }
-
-    dispatcher.showPopup({
-      type: PopupType.MergeConflicts,
-      repository,
-      currentBranch: tip.branch.name,
-      comparisonBranch: branch,
-    })
-  } else {
-    // TODO: what should we display if this is a pull?
+  const { tip, branch } = gitContext
+  if (tip == null || tip.kind !== TipState.Valid) {
+    return error
   }
+
+  dispatcher.showPopup({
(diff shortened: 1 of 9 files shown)


app/src/lib/dispatcher/error-handlers.ts
```diff
@@ -245,6 +246,74 @@ export async function pushNeedsPullHandler(
...
+    }
+
+    dispatcher.showPopup({
+      type: PopupType.MergeConflicts,
+      repository,
+      currentBranch: tip.branch.name,
+      comparisonBranch: branch,
+    })
+  } else {
+    // TODO: what should we display if this is a pull?
```
shiftkey: @outofambit previously this error handler would also fire when you triggered a conflict via `pull` (which is merging the remote tracking branch with your local branch). It looks like this dialog now expects this to be a "comparison branch" - will providing a remote branch name confuse your dialog? Could `comparisonBranch` be renamed to something more generic in Git (like `theirs`)?

outofambit: oh i see what you're saying. yeah that totally makes sense. perhaps `theirBranch`?

outofambit: (also a remote branch name won't confuse the dialog, that should be totally fine)

shiftkey:
> do you mind if i push up a commit on this branch to make that change?

Working on that now, gimme a couple of mins to wrap it up.

app/src/lib/dispatcher/error-handlers.ts
shiftkey: Cool, let me push a commit to rename this and see how it feels.

shiftkey:
:apple:

PR #6001, opened 2018-10-25
tweak Azure Pipelines config to mirror other build platforms

shiftkey:
## Overview

Currently we've setup Azure Pipelines without any triggers, which means it's running lots of builds by default. I've spotted it running two builds whenever a PR from a local branch is updated (one for the branch being updated, the second to validate the pull request, but they're both the same commit ID so it's rather wasteful).

## Description

[This documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=vsts&tabs=yaml) shows the syntax for defining the triggers, and in particular:

> If you set `batch` to `true`, when a build is running, the system waits until the build is completed, then queues another build of all changes that have not yet been built.

So here's the constraints I've added:

 - batching builds, to avoid wasted builds when a branch is churning due to lots of pushes
 - only building specific branches (matches how we have Appveyor/Circle/Travis configured)

## Release notes

Notes: no-notes

### DIFF
--- a/vsts.yml
+++ b/vsts.yml
@@ -4,8 +4,8 @@ trigger:
     include:
       - master
       - __release-*
-      
-# initialize submodules before running build jobs      
+
+# initialize submodules before running build jobs
 resources:
   repositories:
     - repository: self
--- a/vsts.yml
+++ b/vsts.yml
@@ -4,6 +4,13 @@ trigger:
     include:
       - master
       - __release-*
+      
+# initialize submodules before running build jobs      
+resources:
+  repositories:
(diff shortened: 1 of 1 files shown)


shiftkey:
I don't think Azure Pipelines is our bottleneck at the moment from all the builds, but I do think this will speed up the feedback that contributors will see (given Pipelines runs the dedicated lint step).

shiftkey:
Something isn't right about this based on my testing over on `shiftkey/desktop`. Gonna close this out until I can figure out what's been missed.

PR #6004, opened 2018-10-25
don't call getSha in local build

outofambit:
closes #6000

PR #6003, opened 2018-10-25
allow for local build in getSha

outofambit:
## Overview

**Closes #6000**

## Description

- accounts for local environment by checking if `NODE_ENV` is `development`
- should it return something besides an empty string?

PR #6005, opened 2018-10-26
revert this tracing as it affects local production builds

shiftkey:
## Overview

#6004 helped unblock `yarn build:dev` locally but `yarn build:prod` locally (which sets `NODE_ENV=production`) still falls through to call `getSha`, as mentioned in https://github.com/desktop/desktop/pull/6003#issuecomment-433289489.

We weren't doing that tracing before #5663, so let's just put current `master` back to that state.

## Release notes

Notes: no-notes

### DIFF
--- a/script/build.ts
+++ b/script/build.ts
@@ -43,19 +43,14 @@ import {
 } from '../app/package-info'
 
 import { getReleaseChannel, getDistRoot, getExecutableName } from './dist-info'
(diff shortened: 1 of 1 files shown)


shiftkey:
#jolo to unblock local development

j-f1:
Why not change `getSHA` to run `git rev-parse HEAD` everywhere?

PR #6013, opened 2018-10-26
Add check for attempts to commit >100 MB files without Git LFS

### LINKED ISSUE #997: Warn me before I commit large files without using Git LFS
Dropped a 1G file in my local repo and committed it. No error validation prevented me from doing so. 

Do we have a limit as I tried dropping same file into dotcom UI and was limited to 100MB? Isn't LFS supposed to do this?

screen shot 2017-03-01 at 4 08 57 pm

Daniel-McCarthy:
## Overview
Closes #997

## Description
- What:

This pull request aims to display a pop-up to notify the user in situations where they are committing files larger than 100 MB and do not have Git LFS set up.
___
- How:

The implementation of this extends the standard commit button behavior. When the commit button is pressed it will now check if Git LFS is set up. If so, it will commit normally. If not, it will proceed to check if any of the selected files to commit are over 100 MB in size.

If none are then it will commit normally. If not, it will display a new Popup Dialog called the `OversizedFilesWarning`which informs the user of the situation and offer two options: (1. Cancel the commit. (2. Commit anyway.
___
- Why: 

Currently if a user commits a large file over 100 MB and does not have Git LFS set up, they will get an error when attempting to push to their GitHub remote repository. Often times the user may not have been aware of the size of the file and now in themselves not able to push anymore.

This solution aims to warn users before they get stuck not being able to push and warn them early.
___
- Additional information:

Currently this implementation supports creating a commit for the user from within the dialog via the `Commit Anyway` button. However, normally when the commit button is pressed the commit message and commit summary will be cleared after the commit. This functionality does not currently occur when committing from the dialog due to complications [mentioned in original issue](https://github.com/desktop/desktop/issues/997#issuecomment-433072506).

@outofambit has [opted to address this](https://github.com/desktop/desktop/issues/997#issuecomment-433564825) in another pull request. Thanks again!

___

## Release notes

Notes: Added dialog to check for oversized file issues that would prevent pushing to the remote repository.


app/src/ui/changes/changes-list.tsx
```diff
@@ -377,6 +379,30 @@ export class ChangesList extends React.Component<
...
+
+    for (const file of workingDirectoryFiles) {
+      const fileIsSelected =
+        file.selection.getSelectionType() !== DiffSelectionType.None
+
+      if (fileIsSelected) {
+        try {
+          const filePath = path.join(this.props.repository.path, file.path)
+          // eslint-disable-next-line no-sync
+          const fileStatus = FSE.statSync(filePath)
```
shiftkey: `fs-extra` has a promise-based API, so this can be rewritten to avoid making `eslint` annoyed:

```ts
const fileStatus = await FSE.stat(filePath)
```

app/src/ui/changes/changes-list.tsx
```diff
@@ -377,6 +379,30 @@ export class ChangesList extends React.Component<
...
+          const fileSizeMegabytes = fileStatus.size / 1000000
+          if (fileSizeMegabytes > 100) {
+            fileNames.push(file.path)
+          }
+        } catch (error) {}
```
shiftkey: Do we want to get any sort of tracing into the logs around these errors.

At the least I'd recommend a `log.debug` so we can identify things in the dev tools:

```ts
} catch (error) {
  log.debug(`Unable to get file size for file ${filePath}`, error)
}
```

Daniel-McCarthy: Fantastic, I will make these changes!
Thanks again for the great feedback!

app/src/ui/changes/commit-message.tsx
```diff
@@ -246,19 +249,39 @@ export class CommitMessage extends React.Component<
...
-      // allow single file commit without summary
+    const { summary, description } = this.state
+    // allow single file commit without summary
+    const commitSummary =
       this.props.singleFileCommit && !this.state.summary
         ? this.props.placeholder
-        : summary,
+        : summary
+
+    const lfsSupported = await this.checkForLFS()
```
shiftkey: As part of rolling out new features like this, we want to enable the ability to feature-flag them, so we can control when we enable them for new users. I'm going to push a commit to this branch in a bit to let us conditionally run this check, but let me know if you have any other questions about this area:

```diff
diff --git a/app/src/lib/feature-flag.ts b/app/src/lib/feature-flag.ts
index 442d75d8d..b5ca4ff03 100644
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -61,3 +61,8 @@ export function enableStatusWithoutOptionalLocks(): boolean {
 export function enableRecurseSubmodulesFlag(): boolean {
   return enableBetaFeatures()
 }
+
+/** Should the app check and warn the user about committing large files? */
+export function enableFileSizeWarningCheck(): boolean {
+  return enableDevelopmentFeatures()
+}
diff --git a/app/src/ui/changes/commit-message.tsx b/app/src/ui/changes/commit-message.tsx
index cbb71c3f0..2e1ab2b75 100644
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -25,6 +25,7 @@ import { ITrailer } from '../../lib/git/interpret-trailers'
 import { IAuthor } from '../../models/author'
 import { IMenuItem } from '../../lib/menu-item'
 import { isUsingLFS } from '../../lib/git/lfs'
+import { enableFileSizeWarningCheck } from '../../lib/feature-flag'
 
 const addAuthorIcon = new OcticonSymbol(
   12,
@@ -261,22 +262,24 @@ export class CommitMessage extends React.Component<
         ? this.props.placeholder
         : summary
 
-    const lfsSupported = await this.checkForLFS()
-    if (lfsSupported === false) {
-      const overSizedFiles = await this.checkForLargeFiles()
-
-      if (overSizedFiles.length !== 0) {
-        this.props.dispatcher.showPopup({
-          type: PopupType.OversizedFiles,
-          dispatcher: this.props.dispatcher,
-          fileList: overSizedFiles,
-          commitSummary: commitSummary,
-          commitDescription: description,
-          repository: this.props.repository,
-          trailers: trailers,
-        })
-
-        return
+    if (enableFileSizeWarningCheck()) {
+      const lfsSupported = await this.checkForLFS()
+      if (lfsSupported === false) {
+        const overSizedFiles = await this.checkForLargeFiles()
+
+        if (overSizedFiles.length !== 0) {
+          this.props.dispatcher.showPopup({
+            type: PopupType.OversizedFiles,
+            dispatcher: this.props.dispatcher,
+            fileList: overSizedFiles,
+            commitSummary: commitSummary,
+            commitDescription: description,
+            repository: this.props.repository,
+            trailers: trailers,
+          })
+
+          return
+        }
       }
     }
```

shiftkey:
Tentatively dropping this from the `1.5.0` milestone as we've only just started the review and we might be pushing ourselves to get this baked on beta sufficiently before the 7th (when we're aiming to ship this) along with the other items in our milestone.

billygriffin:
@Daniel-McCarthy As @shiftkey mentioned above, we talked this morning about this one and we're really excited about it, but we don't want to rush it into next week's release. It feels like it'll be important to have it out on beta for a decent period of time to make sure it's working as expected. So it'll likely go into 1.5.x and be available for all users in 1.6.

Daniel-McCarthy:
@billygriffin Sounds great! That's certainly no problem. I see no reason to rush it and expected it might need a decent bit of review and test time since it is so closely tied with core functionality of committing.

app/src/models/popup.ts
```diff
@@ -121,3 +124,12 @@ export type Popup =
       pullRequest: PullRequest
     }
   | { type: PopupType.MergeConflicts; repository: Repository }
+  | {
+      type: PopupType.OversizedFiles
+      dispatcher: Dispatcher
```
shiftkey: We have the `dispatcher` already available in `app.tsx` so we don't need to provide this explicitly

app/src/models/popup.ts
```diff
@@ -121,3 +124,12 @@ export type Popup =
...
+      fileList: string[]
```
shiftkey: Could this be a more descriptive field than `fileList`? Maybe `largeFiles` or `filesExceedingLimit`?

app/src/models/popup.ts
```diff
@@ -121,3 +124,12 @@ export type Popup =
...
+      commitSummary: string
+      commitDescription: string | null
+      repository: Repository
+      trailers?: ReadonlyArray<ITrailer>
```
shiftkey: Seeing these three fields `commitSummary`, `commitDescription` and `trailers` repeated here makes me think this is a great candidate for turning them into a shape like `CommitDetails` - for now I won't hold up this PR but I'll follow up with some tidy-up to introduce this shape:

```ts
export type CommitDetails = {
  summary: string
  description: string | null
}
```

app/src/ui/app.tsx
```diff
@@ -1323,6 +1324,18 @@ export class App extends React.Component<IAppProps, IAppState> {
             onDismissed={this.onPopupDismissed}
           />
         )
+      case PopupType.OversizedFiles:
+        return (
+          <OversizedFiles
+            fileNames={popup.fileList}
+            onDismissed={this.onPopupDismissed}
+            dispatcher={popup.dispatcher}
```
shiftkey: This will become `dispatcher={this.props.dispatcher}` once you've removed `popup.dispatcher`.

app/src/ui/changes/changes-list.tsx
```diff
@@ -377,6 +379,31 @@ export class ChangesList extends React.Component<
     showContextualMenu(items)
   }
 
+  private getNamesOfSelectedOversizedFiles = async () => {
+    const fileNames: string[] = []
```
shiftkey: :art: no need to be explicit about the signature: `const fileNames = new Array<string>()`

app/src/ui/changes/changes-list.tsx
```diff
@@ -377,6 +379,31 @@ export class ChangesList extends React.Component<
...
+    const workingDirectoryFiles = this.props.workingDirectory.files
+
+    for (const file of workingDirectoryFiles) {
+      const fileIsSelected =
```
shiftkey: :art: rather than doing this check in the for loop, you could filter the array up front:

```ts
const workingDirectoryFiles = this.props.workingDirectory.files

const selectedFiles = workingDirectoryFiles.filter(
  f => f.selection.getSelectionType() !== DiffSelectionType.None
)

for (const file of selectedFiles) {
  // ...
}
```

shiftkey (dismissed):
Some initial feedback - I've prefixed the aesthetic changes with :art: to indicate those suggestions are more about the readability of the code. Let me know if anything is unclear.

tierninho:
@Daniel-McCarthy and @shiftkey a few questions regarding testing:

- Do you know of a way to toggle off/on LFS or at least trick Desktop? I've been using Dev Tools > Application > Local Storage to change values but not sure if it will be recorded here.
- For the `commit anyway` button, does this mean GitHub will now allow 100+ MB files even if the user does not have LFS ... however there is no turning back? I imagine github/github would have to make allowance changes for this?
- Do limits apply for Enterprise accounts as well?

billygriffin:
> For the commit anyway button, does this mean GitHub will now allow 100+ MB files even if the user does not have LFS ... however there is no turning back? I imagine github/github would have to make allowance changes for this?

@tierninho It means they won’t be able to push to GitHub, effectively leaving them in the same place as they are today. We don’t want to block their progress if there’s some reason they want to commit these large files anyway, but we do want to make it clear that it will subsequently prevent them from pushing to GitHub. If you think the verbiage in the warning doesn’t adequately convey that, we can discuss altering it a bit.

Daniel-McCarthy:
@tierninho I see that @billygriffin has covered the question about the `Commit Anyway` button. I am not sure about the Enterprise question and will leave that for someone who knows for sure.

> Do you know of a way to toggle off/on LFS or at least trick Desktop?
https://github.com/desktop/desktop/blob/883d42de873dc6f9ddb3a339ca09ac5f38bc81bd/app/src/lib/git/lfs.ts#L27-L36

This is the function that retrieves whether Git LFS is being used or not. I'm not sure if there is something useful there for your situation.

I do know of one way, but it's not quite as convenient and is a little bit more brute-force in way.
(Images not embedded due to their large size).
[Image 1](https://user-images.githubusercontent.com/4957200/47684856-5d38f980-dba2-11e8-9807-f0d25cf48005.png)
- Go to `Dev Tools > Sources`, then under Sources go to webpack:// > . > app > src > ui > changes > commit-message.tsx

[Image 2](https://user-images.githubusercontent.com/4957200/47684705-ca985a80-dba1-11e8-9a05-58742a5f2150.png)
- Place a breakpoint on line 267 of Commit-Message.tsx

[Image 3](https://user-images.githubusercontent.com/4957200/47684757-fadff900-dba1-11e8-9ae7-8aea2fde76f9.png)
- Go down to the scope menu and double click the value for `lfsSupported`. You can manipulate whether it is true or false. But you would need to do this each time.

tierninho:
@billygriffin Ah yes, we are good as-is. Commit ≠ push.

@Daniel-McCarthy Thanks, that should hopefully do it!

shiftkey:
@tierninho @Daniel-McCarthy to enable Git LFS on your repository you can run this command (if you have Git LFS installed on the command line):

```
git lfs track "*.png"
```

If you don't have the Git LFS command line installed you can manually enable Git LFS using the `.gitattributes` file, adding a line for each entry you want to track in Git LFS:

```
*.png filter=lfs diff=lfs merge=lfs -text
```

This raises an interesting point - we're checking for Git LFS _and_ we're testing for large files, but we're not currently checking for specifics about how Git LFS is configured for the user (which files are included):

Here's some scenarios I need to think about further while we're testing this area (your input is welcome here too):

 - can we easily get Git LFS to indicate whether or not it will handle a large file we've detected? 
 - if we have Git LFS enabled but a file will not be handled (because it doesn't match any found expressions), what should Desktop do?
 - if Git LFS is not enabled but we are committing large files to a GitHub repository, should we recommend the user add rules for these files before creating the commit?

Daniel-McCarthy:
@shiftkey If they have Git LFS set up but not configured how it needs to be in order to commit the included files, perhaps we can then either display a different configuration of the dialog or a different dialog entirely.

Perhaps we can display the large files that are of types that have not been configured for and display which types need to be updated in the configuration in order to commit them.

Would it also be practical to potentially offer to update their configuration via a button?

billygriffin:
I'm not sure how the configuration of Git LFS works or how simple it is, but at the very least it seems reasonable to let the user know that the large files they're trying to commit are not configured for Git LFS, and that should likely update the configuration before committing.

So would this be two scenarios that would trigger a warning and one happy path?

Scenario 1:
1. The user does not have LFS and is trying to commit large files
2. We warn them as we've already discussed here.

Scenario 2:
1. The user has Git LFS but the files they're trying to commit are not configured.
2. We warn them that they need to configure Git LFS for those files if they want to be able to push the repo to GitHub (or potentially do the config for them, but only if it's super simple to do so which I'm skeptical of) - I think this results in the same selection options in the dialog, but different verbiage?

Scenario 3:
1. The user has Git LFS and the files are configured for it.
2. They commit without anything warning or blocking them.

Daniel-McCarthy:
@billygriffin I think that layout sounds great.

It seems to me that what we would want to do is in the case of Scenario 2&3 is get a list of each of the file extensions the user is trying to commit. Then use `git lfs track` to get a list of what's being tracked and see if that list includes each extension the user is attempting to commit.

If not, inform the user that they are not tracking that file type in git lfs and allow them to cancel, commit anyway, and potentially an option that adds the extensions for them.

Once we have a list of the file extensions that the user wants to commit but are not being tracked in git lfs, then I believe we could do `git lfs track '.extensionName'` if the user wants to start tracking those extensions.
___
Though perhaps this leaves open a question for files with no extension.

shiftkey:
> It seems to me that what we would want to do is in the case of Scenario 2&3 is **get a list of each of the file extensions the user is trying to commit**. Then use git lfs track to get a list of what's being tracked and see if that list includes each extension the user is attempting to commit.

@Daniel-McCarthy this is something that I'm working on at the moment, so don't worry too much about it.

I'm hoping to open a PR today to explain it further, but this is the idea I'm running with:

 - you've already done the work to identify which files will break when committing
 - I'm working on some plumbing named `isTrackedByLFS` that takes a relative path and works out if it's covered by Git LFS
 - in theory, you'll then be able to filter your list of problem files to skip files already tracked in Git LFS, if existing rules are found in the repository

Daniel-McCarthy:
@shiftkey Wonderful!
That definitely takes care of the main obstacle currently. I think it would be great if we could potentially take care of adding the file to their tracking for them, but it's definitely not an emergency.

With this change and a dialog added to support checking for untracked files we should be able to catch this scenario to about the same capacity as the original scenario. Catch the situation, throw a dialog to inform the user, give them information on how to solve their issue, and allow to `cancel` or `commit anyway`.

I'll continue to think on what that dialog should look like. And once these changes are merged (unless we can just cherry pick them in?) I can try to get this functionality in.

Thanks again for the great work adding support for checking if a file is tracked!

shiftkey:
>  I think it would be great if we could potentially take care of adding the file to their tracking for them, but it's definitely not an emergency.

I've opened #6055 to start that discussion independent of this

shiftkey:
@billygriffin apologies for the late reply. The "Scenario 2" here is not quite what I had planned for this PR:

> 1. The user has Git LFS but the files they're trying to commit are not configured.

Git LFS is only enabled for certain file extensions on an opt-in basis, so we needed more insight than the original `isUsingLFS` function (which checks for _any_ files being tracked by Git LFS). That's why I opened #6053, so we can check whether Git LFS is tracking a specific file in Desktop (and thus the file won't actually be committed).

> 2. We warn them that they need to configure Git LFS for those files if they want to be able to push the repo to GitHub (or potentially do the config for them, but only if it's super simple to do so which I'm skeptical of) - I think this results in the same selection options in the dialog, but different verbiage?

#6053 adds support for callers to identify which files in a repository will not be tracked by Git LFS. This is important because of how Git LFS inserts itself in the commit process (it swaps out the file contents for a pointer file containing metadata, which is committed to the repository). If we know that's going to happen, we shouldn't be warning the user and blocking them from committing.

I've punted on the config changes to #6055 so we can talk about the flow itself and what we want to achieve in more detail.

Daniel-McCarthy:
I see #6053 has been merged so I should be able to continue on this now. I have merged in the latest commits from master and resolved the merge conflicts.

I plan to take a look at [@shiftkey 's comment](https://github.com/desktop/desktop/pull/6053#issuecomment-434697631) and implement those changes. I will get started on the basics of new dialog that's needed to get the functionality in over the weekend. And then I can do a design pass afterwards once that's in place.

Daniel-McCarthy:
I have made some more changes as suggested by @shiftkey. At this point it is ready to accommodate a dialog for handling situations where the user is attempting to commit files that are not tracked by git LFS.
___
It seems that the dialog we have currently is nearly identical to what this new dialog may potentially need. Both would be passed the same arguments, with the primary one being a list of files that will not be able to be pushed if a commit is created. They both inform the user of their reasonably similar situations.

The dialog that has already been designed and implemented is meant for a user that does not have Git LFS installed or set up at all. We are now trying to address a scenario where the user has Git LFS set up, but do not have Git LFS set up to track a type of file they are trying to commit.

In both scenarios it would seem we would be linking them to [this Git LFS tutorial](https://github.com/git-lfs/git-lfs/wiki/Tutorial). So, it may not be ideal to split this into two dialogs after all. Instead, it may be best to update the verbiage of the current dialog when the user already has Git LFS and give them more detailed information specific to their situation. Perhaps we can show the extensions of the files they would need to track in order to commit.

billygriffin:
> In both scenarios it would seem we would be linking them to this Git LFS tutorial. So, it may not be ideal to split this into two dialogs after all. Instead, it may be best to update the verbiage of the current dialog when the user already has Git LFS and give them more detailed information specific to their situation. Perhaps we can show the extensions of the files they would need to track in order to commit.

I admit to being a bit confused now about the specifics due to the implementation discussions, but in general this sounds exactly like what I'd expect as a user.

billygriffin:
Great description @Daniel-McCarthy, thanks for summing it up. I agree with everything you said. Hopefully we'll be able to look at iterating in #6055 as well, which should help ease the pain of the middle ground.

shiftkey (commented):
@Daniel-McCarthy this is pretty close to landing, and I think we're mostly down to small things like figuring out the dialog messaging. I've also opened #6190 to address a case I found while testing this out, but I'm excited to see this make it out to users!

app/src/ui/changes/oversized-files-warning.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <p>
+            The following files are over 100MB.{' '}
+            <strong>
+              If you commit these files, you will no longer be able to push this
+              repository to GitHub.com.
+            </strong>
+          </p>
+          {this.renderFileList()}
+          <p className="recommendation">
+            We recommend you remove these files before committing or use{' '}
```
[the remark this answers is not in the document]
Daniel-McCarthy: > I think this message could be a bit friendlier.

I agree and think this recommendation would work great. I have gone ahead and made these changes and am happy to update this further if there's anything else that looks like it can be improved.

shiftkey:
As I've done a first round of review and helped a lot with the Git LFS specifics, I'd love someone else from @desktop/engineering to give this a :eyes: and ensure they're happy with the feature. Unassigning myself in the meantime.

app/src/ui/changes/oversized-files-warning.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <p>
+            The following files are over 100MB.{' '}
+            <strong>
+              If you commit these files, you will no longer be able to push this
+              repository to GitHub.com.
+            </strong>
+          </p>
+          {this.renderFileList()}
+          <p className="recommendation">
+            We recommend you avoid committing these files before committing or
```
billygriffin:
```suggestion
            We recommend you avoid committing these files or
```

Daniel-McCarthy: Ah, good catch @billygriffin!
I'll update this now.

tierninho:
Note: Initial focus should be on the `Cancel` button as current functionality opens the `Git LFS` link if you hit the `enter` button as soon as the modal opens.

tierninho:
The UI was tested for this change, and overall it functions well based on initial tests. 

- [x] added 100mb+ file, and warning shown
- [x] added 100mb+ file and normal file(s), and warning shown
- [x] added multiple 100mb+ file, and warning shown with all files
- [x] warning buttons functional, user can escape or X out of modal
- [x] warning shows for both GitHub and Enterprise repos
- [x] undoing a large commit is intact, can recommit and get warning
- [x] push blocked after committing large file

tests based on https://github.com/desktop/desktop/pull/6013#issuecomment-434336789 to follow

Build/OS: 1.5.1-beta0, mac

Daniel-McCarthy:
Currently the build is failing due to `ITrailer` being imported but is not being used. I will take care of that and let it rebuild later today as I've run out of time for now. Just wanted to update.

tierninho:
Completed testing of this feature based on https://github.com/desktop/desktop/pull/6013#issuecomment-438812048 however I noticed a few things.

1) The Initialize Git LFS dialogue always appears for my repo, and clicking `Initialize Git LFS` produces the following log entry`Executing installLFSHooks: git lfs install --force` which is already a step as part of the installation outlined here: https://git-lfs.github.com/ Are we being redundant? and if you were to click `Not Now` or dismiss the dialogue, it produces no log entry and seems to have no bearing on the outcome of the the tests (as LFS was already installed). 

2) There were no warnings in terminal when committing large files that weren't previously "approved" in the `gitattribute` file. Is this intended behavior? The commit was committed successfully in Desktop for this test case.

screen shot 2018-11-19 at 10 17 39 am

Happy to provide additional tests around this as needed. Otherwise, signing off ✅

shiftkey:
@tierninho 

> Are we being redundant?

There might be some redundancy here if the user is adding a new local repository that's already initialized correctly, but this step also ensures newly cloned repositories have the `.git/hooks` that Git LFS needs to ensure it will commit and push correctly. 

I'd be 👍 to opening a tracking issue about this, to see if we can improve this setup and skip a correctly initialized repository.

> and if you were to click Not Now or dismiss the dialogue, it produces no log entry and seems to have no bearing on the outcome of the the tests (as LFS was already installed).

That's fine.

> There were no warnings in terminal when committing large files that weren't previously "approved" in the gitattribute file. Is this intended behavior?

Not sure I understand this one. Can you elaborate?

tierninho:
@shiftkey 

> I'd be 👍 to opening a tracking issue about this

Will do. 

> Not sure I understand this one. Can you elaborate?

So if the `gitattribute` file is:
```
*.zip filter=lfs diff=lfs merge=lfs -text
#*.cpgz filter=lfs diff=lfs merge=lfs -text
```
and I attempt to commit a 100+ zip file on Desktop, it works as expected, but a cpgz file surfaces the LFS warning as expected. When attempting the same commits via terminal, there is no warning for the cpgz file and Desktop follows suit in the background.

shiftkey:
> When attempting the same commits via terminal, there is no warning for the cpgz file and Desktop follows suit in the background.

Now I'm with you. Yep, that's by design because there is no support for this in Git core - it's all on us to detect and warn the user before they create the commit.

outofambit:
@Daniel-McCarthy this is looking good to me! i think the only thing left to do is address this outstanding todo: 

> However, normally when the commit button is pressed the commit message and commit summary will be cleared after the commit. This functionality does not currently occur when committing from the dialog due to complications mentioned in original issue.

see #6049 and #6097 specifically as a guide and let me know if you have any questions! ⛷ 

I'd love to get this in soon-ish (if possible) since i'm starting to write something similar for #6213 (checking for conflicts when hitting the commit button) 😄

outofambit:
@Daniel-McCarthy i've reproduced this locally and am working on a fix since its not caused by anything in your pr. will update (soon) when i have more info.

app/src/ui/changes/oversized-files-warning.tsx
```diff
@@ -0,0 +1,104 @@
...
+            </Monospaced>
+          </ul>
+        ))}
+      </div>
+    )
+  }
+
+  private commitAnyway = async () => {
+    const context = this.props.context
+    this.props.dispatcher.commitIncludedChanges(this.props.repository, context)
```
outofambit:
```suggestion
    await this.props.dispatcher.commitIncludedChanges(this.props.repository, context)
```

outofambit: i _think_ we don't want to clear the message until the commit is finished. is that right?

Daniel-McCarthy: @outofambit 
I can certainly make this change. The reason I chose not to await `commitIncludedChanges` is because creating the commit may take a while, especially with a lot of files. It seemed preferable to get the popup out of the way since it was no longer necessary and allow them to see the commit creation now they normally would.

Perhaps we should show a spinning wheel on the Commit Anyway button to indicate that it's working on it? Otherwise I would be fine with just adding the `await`, I'd just like to mention this to see what would be the ideal way for the user.

Daniel-McCarthy: To add to this:
> i think we don't want to clear the message until the commit is finished. is that right?

That is correct, if we clear the commit message while it is committing, the commit summary will display the default commit message for the selected files.

However, we can change that so that it does not get cleared until after the commit. We can solve that regardless of whether or not we want to close the pop-up while committing. If we want to keep the current functionality of closing the popup during the commit, we can just move the `closePopup()` to happen before the await `commitIncludedChanges`.

outofambit: > we can just move the closePopup() to happen before the await commitIncludedChanges.

that's a great suggestion. let's do that! @Daniel-McCarthy :D

outofambit (approved):
✨

billygriffin:
Hey @Daniel-McCarthy! We'd love to call out your contribution on this in the release blog post for 1.6. Would it be ok if we mention you using your handle?

Daniel-McCarthy:
@billygriffin Of course, that's perfectly acceptable.
Thanks for the consideration! I'm quite happy that these changes have landed and hope they will help users in the future!

rcyost:
Is there a way to automatically ignore files over the limit size? I don't see the logic behind trying to push and then causing a nasty error thats hard to reset. Why can't I just write a flag in my gitignore by file size like file type? Right now I can tell LFS to LFS files by type, but thats very manual. I am looking for functionality that lets me simply flip a flag and never have to deal with file size issues again.

PR #6018, opened 2018-10-28
Improve single file commit summary to match GitHub.com

### LINKED ISSUE #6017: improvements to default message when single file chosen to commit

The changes in #5389 around showing a default message don't seem to match up with the behaviour on GitHub, and I think it's a relatively-straightforward change to address.

Two specific things I've noticed:

 - first, on the website the commit summary is by default "Update [filename]" rather than "Update [path-to-file]" that Desktop displays. That's this line:

https://github.com/desktop/desktop/blob/c82e0707ce4072ec4e768f8f143137eee3c0db0f/app/src/ui/changes/changes-list.tsx#L416

For reference, heres's what I see when updating a file named `Global/SlickEdit.gitignore`

 - second, we should differentiate between creating a new file ("Create [file]"), updating an existing file ("Update [file]") or removing a file ("Removed [file]").

An example of creating a new file:

An example of removing a file:

Daniel-McCarthy:
## Overview

**Closes #6017**

## Description

- Updates the default single file commit summary to show the file name and not the rest of the path.
- Updates the default single file commit summary to show "Create" and "Delete" in addition to "Update" depending on the type of changes being committed.

Preview of the changes being made:
image
## Release notes

Notes: Updated the default commit summary to match GitHub.com more closely.

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -31,6 +31,7 @@ import { IAutocompletionProvider } from '../autocompletion'
 import { showContextualMenu } from '../main-process-proxy'
 import { arrayEquals } from '../../lib/equality'
 import { clipboard } from 'electron'
+import * as path from 'path'
 
 const RowHeight = 29
 
@@ -428,7 +429,7 @@ export class ChangesList extends React.Component<
           coAuthors={this.props.coAuthors}
           placeholder={
             singleFileCommit
-              ? `Update ${filesSelected[0].path}`
+              ? `Update ${path.parse(filesSelected[0].path).base}`
               : 'Summary (required)'
           }
           singleFileCommit={singleFileCommit}
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -31,7 +31,7 @@ import { IAutocompletionProvider } from '../autocompletion'
 import { showContextualMenu } from '../main-process-proxy'
 import { arrayEquals } from '../../lib/equality'
 import { clipboard } from 'electron'
-import * as path from 'path'
+import { basename } from 'path'
 
 const RowHeight = 29
 
@@ -430,10 +430,10 @@ export class ChangesList extends React.Component<
           placeholder={
             singleFileCommit
               ? filesSelected[0].status === AppFileStatus.New
-                ? `Create ${path.parse(filesSelected[0].path).base}`
+                ? `Create ${basename(filesSelected[0].path)}`
                 : filesSelected[0].status === AppFileStatus.Deleted
-                  ? `Delete ${path.parse(filesSelected[0].path).base}`
-                  : `Update ${path.parse(filesSelected[0].path).base}`
+                  ? `Delete ${basename(filesSelected[0].path)}`
+                  : `Update ${basename(filesSelected[0].path)}`
               : 'Summary (required)'
           }
           singleFileCommit={singleFileCommit}
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -429,7 +429,11 @@ export class ChangesList extends React.Component<
           coAuthors={this.props.coAuthors}
           placeholder={
             singleFileCommit
-              ? `Update ${path.parse(filesSelected[0].path).base}`
+              ? filesSelected[0].status === AppFileStatus.New
+                ? `Create ${path.parse(filesSelected[0].path).base}`
+                : filesSelected[0].status === AppFileStatus.Deleted
+                  ? `Delete ${path.parse(filesSelected[0].path).base}`
+                  : `Update ${path.parse(filesSelected[0].path).base}`
               : 'Summary (required)'
           }
           singleFileCommit={singleFileCommit}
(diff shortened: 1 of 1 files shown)


app/src/ui/changes/changes-list.tsx
```diff
@@ -31,6 +31,7 @@ import { IAutocompletionProvider } from '../autocompletion'
 import { showContextualMenu } from '../main-process-proxy'
 import { arrayEquals } from '../../lib/equality'
 import { clipboard } from 'electron'
+import * as path from 'path'
```
shiftkey: :art: instead of importing everything, why not just `import { parse } from 'path'`?

Our production minification steps should tidy this up anyway, but it'll simplify the change below and align better with how we import specific functions from core modules.

app/src/ui/changes/changes-list.tsx
```diff
@@ -428,7 +429,11 @@ export class ChangesList extends React.Component<
           coAuthors={this.props.coAuthors}
           placeholder={
             singleFileCommit
-              ? `Update ${filesSelected[0].path}`
+              ? filesSelected[0].status === AppFileStatus.New
```
shiftkey: This feels like it could be lifted out to it's own function. I noodled on this version to illustrate how it might look:

```diff
diff --git a/app/src/ui/changes/changes-list.tsx b/app/src/ui/changes/changes-list.tsx
index 33ed077bd..4747f7828 100644
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -31,7 +31,7 @@ import { IAutocompletionProvider } from '../autocompletion'
 import { showContextualMenu } from '../main-process-proxy'
 import { arrayEquals } from '../../lib/equality'
 import { clipboard } from 'electron'
-import * as path from 'path'
+import { parse } from 'path'
 
 const RowHeight = 29
 
@@ -378,6 +378,32 @@ export class ChangesList extends React.Component<
     showContextualMenu(items)
   }
 
+  private getPlaceholderMessage(
+    files: ReadonlyArray<WorkingDirectoryFileChange>,
+    singleFileCommit: boolean
+  ) {
+    if (!singleFileCommit) {
+      return 'Summary (required)'
+    }
+
+    const firstFile = files[0]
+    // TODO: can this error? what should happen in that case?
+    const fileName = parse(firstFile.path).base
+
+    switch (firstFile.status) {
+      case AppFileStatus.New:
+        return `Create ${fileName}`
+      case AppFileStatus.Deleted:
+        return `Delete ${fileName}`
+      default:
+        // TODO:
+        // this doesn't feel like a great message for AppFileState.Copied or
+        // AppFileStatus.Renamed but without more insight (and whether this
+        // affects other parts of the flow)
+        return `Update ${fileName}`
+    }
+  }
+
   public render() {
     const fileList = this.props.workingDirectory.files
     const fileCount = fileList.length
@@ -427,15 +453,10 @@ export class ChangesList extends React.Component<
           isCommitting={this.props.isCommitting}
           showCoAuthoredBy={this.props.showCoAuthoredBy}
           coAuthors={this.props.coAuthors}
-          placeholder={
+          placeholder={this.getPlaceholderMessage(
+            filesSelected,
             singleFileCommit
-              ? filesSelected[0].status === AppFileStatus.New
-                ? `Create ${path.parse(filesSelected[0].path).base}`
-                : filesSelected[0].status === AppFileStatus.Deleted
-                  ? `Delete ${path.parse(filesSelected[0].path).base}`
-                  : `Update ${path.parse(filesSelected[0].path).base}`
-              : 'Summary (required)'
-          }
+          )}
           singleFileCommit={singleFileCommit}
         />
       </div>
```

I also had a couple of questions about the new implementation:

 - Will `path.parse` throw an error for paths it cannot handle correctly? Is this something we should be paranoid about? 
 - `AppFileStatus` has other states that could be better described than `Update [file]` - are we likely to stumble upon them? Do we care?

Daniel-McCarthy: Excellent, I had hoped to implement it more cleanly than I had but I didn't think to split it out into it's own function at the time. This should give a bit more wiggle room.

Since we're now using a function, the terseness of the one-liner `path.parse(string).base` may not be necessary. It may be simpler to find the last index of `\` or `/` and substring to everything after it. That way we can always know what to expect from it, and does not require anyone to know what either `.parse()` does or `.base` does.

Additionally, with the second question, both `.Copied` and `.Renamed` feel like good candidates for their own case out of the rest of them.

tierninho:

- [x] update single/multiple file(s)
- [x] delete single/multiple file(s)
- [x] add single/multiple file(s)
- [x] combo update/delete/add multiple files
- [x] add new file, commit only partial lines
- [x] undo a commit, test summary default msg
- [x] default msg can be overwritten
- [x] uncheck all files reverts to no dynamic messaging

I did notice that if a file name is too long, then it runs on past the width of the summary field. Not a big deal, just pointing it out.

screen shot 2018-10-29 at 9 48 00 am

Build/OS: 1.5.0-beta1, mac, 3989dc37418e73492e4c7ed1a543a270a4ce50ae

shiftkey:
> I did notice that if a file name is too long, then it runs on past the width of the summary field. Not a big deal, just pointing it out.

Yeah, unless we're somehow chopping off text (we shouldn't be) I think we'll have to live with it until we get time to dig into ideas around the commit authoring experience...

tierninho:
@Daniel-McCarthy Spotted this randomly... shouldn't a new file be labeled as "create"? I am seeing "update" instead.

screen shot 2018-10-30 at 9 06 03 am

tierninho:
I cannot repro, so we are good to go ⚡️

PR #6024, opened 2018-10-29
Improves License and GitIgnore Support

### LINKED ISSUE #6023: Include ISC License terms
ISC is the default license for npm packages (in other words, `npm init` sets ISC by default), but currently GitHub Desktop does not include it by default so I have to manually copy-paste from https://choosealicense.com/licenses/isc/.

Can we include ISC so that users can select it from "Create a new repository" window?

say25:
## Overview
Updates ChooseALicense.com and GitIgnore submodules to latest

## Description
- Decided to also update GitIgnore to latest to pull in latest changes from there

## Release notes

Notes: Updated license and Git ignore templates for initialising a new repository


say25:
This does NOT actually fix the issue. I was too optimistic. As it turns out, we only show the licenses that are not [hidden](https://github.com/desktop/desktop/blob/master/script/build.ts#L420). I guess this seems to be the first request to show more options? Maybe we show the "hidden" ones after the existing ones?

shiftkey:
@say25 I left some more context in https://github.com/desktop/desktop/issues/6023#issuecomment-433872925 about how to change things, but this PR is still nice from a cleanup perspective :metal:

shiftkey:
I've tested the license and gitignore template pickers to confirm they're not affected by any potential layout changes in the repository, so I'm fine to merge this and get it out as part of the next release.

PR #6025, opened 2018-10-29
rename VSTS config to Azure Pipelines

shiftkey:
## Overview

VSTS was rebranded a while ago to Azure DevOps and so this PR cleans up the last mention of it in our codebase. cc @chrisrpatterson 

## Description

After this PR is merged, I will:

 - in the admin portal, switch the CI build config over to point to this path
 - wait a couple of days to confirm new builds work and active PRs are NOT affected (in which case I'll point it back to `vsts.yml`)
 - open a new PR to remove `vsts.yml` from version control

## Release notes

Notes: no-notes

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,82 @@
+jobs:
+  - job: Windows
+    pool:
+      vmImage: vs2017-win2016
+    steps:
+      - task: NodeTool@0
+        inputs:
+          versionSpec: '8.11.1'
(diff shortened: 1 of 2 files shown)


shiftkey:
As we don't rely on Azure Pipelines for publishing releases I'd like to get this in for verification asap.

outofambit:
@shiftkey want to do the honors so you can do the admin portal thing right away?

shiftkey:
@outofambit can do (once it goes green)

shiftkey:
The admin portal now uses the new config, and I'll monitor this until after our next beta to confirm we're good to deprecate the old config.

PR #6027, opened 2018-10-29
Update title of merge dialog to match label in top-level menu

iAmWillShepherd:
Related to: https://github.com/desktop/desktop/issues/5555

@ampinsk and I discusses the state of the merge dialog title and the proposed changes from https://github.com/desktop/desktop/issues/5555#issuecomment-429963614. I felt we were being overly redundant since the same information was being presented in 2 other places on the dialog and that we were deviating from how all out other dialogs are titled (essentially what we expect you to do). We both agreed that we could approach this by updating the title of the dialog to match that in the menu or only call out the current branch (Merge into `master`). This PR is for the first option because we can always iterate to the one that includes the branch name.

**Before**
screen shot 2018-10-29 at 2 13 45 pm

**After**
screen shot 2018-10-29 at 4 28 24 pm

_Long branch name_
screen shot 2018-10-29 at 4 30 11 pm

### DIFF
--- a/app/src/ui/merge-branch/merge.tsx
+++ b/app/src/ui/merge-branch/merge.tsx
@@ -19,6 +19,8 @@ import { MergeResultKind } from '../../models/merge'
 import { MergeStatusHeader } from '../history/merge-status-header'
 import { promiseWithMinimumTimeout } from '../../lib/promise'
 import { truncateWithEllipsis } from '../../lib/truncate-with-ellipsis'
+import { DialogHeader } from '../dialog/header'
+import { Ref } from '../lib/ref'
 
 interface IMergeProps {
   readonly dispatcher: Dispatcher
@@ -292,14 +294,16 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
     return (
       <Dialog
         id="merge"
-        title={
-          __DARWIN__
-            ? `Merge Into ${currentBranchName}`
-            : `Merge into ${currentBranchName}`
-        }
         onDismissed={this.props.onDismissed}
         onSubmit={this.merge}
       >
+        <DialogHeader title={` `} dismissable={true}>
+          <div className={''}>
+            <p>
+              Merge into <Ref>{currentBranchName}</Ref>
+            </p>
+          </div>
+        </DialogHeader>
         <DialogContent>
           <BranchList
             allBranches={this.props.allBranches}
--- a/app/src/ui/merge-branch/merge.tsx
+++ b/app/src/ui/merge-branch/merge.tsx
@@ -282,14 +282,14 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
       this.state.mergeStatus.kind === MergeResultKind.Invalid
 
     const disabled = invalidBranchState || cannotMergeBranch
-
+    const currentBranchName = this.props.currentBranch!.name
     return (
       <Dialog
         id="merge"
         title={
           __DARWIN__
-            ? `Merge Into ${this.props.currentBranch}`
-            : `Merge into ${this.props.currentBranch}`
+            ? `Merge Into ${currentBranchName}`
+            : `Merge into ${currentBranchName}`
         }
         onDismissed={this.props.onDismissed}
         onSubmit={this.merge}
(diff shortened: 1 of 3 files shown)


iAmWillShepherd:
> non-blocking but should into be uppercase in the title case version?

I'm not sure. Apple states the following:
> Prepositions of four or fewer letters, except when the preposition is part of a verb phrase, as in “Starting Up the Computer”

I think this is a verb phrase, but I really don't know.

iAmWillShepherd:
🔢

app/src/ui/merge-branch/merge.tsx
```diff
@@ -283,13 +285,26 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
...
         onSubmit={this.merge}
       >
+        <DialogHeader
+          title={
+            <div className="merge-dialog-header">
+              Merge into <b>{currentBranchName}</b>
+            </div>
+          }
+          dismissable={true}
+          onDismissed={this.props.onDismissed}
```
shiftkey: I'm not a fan of how we've duplicated `onDismissed` here, and was hoping #5955 would help avoid this. I had a go at adding the support for `Dialog` to accept a `JSX.Element` but there's a bit that I'm not sure of (the `???` part).

```
diff --git a/app/src/ui/dialog/dialog.tsx b/app/src/ui/dialog/dialog.tsx
index 63162f312..9ff9cdee7 100644
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -31,7 +31,7 @@ interface IDialogProps {
    * By omitting this consumers may use their own custom DialogHeader
    * for when the default component doesn't cut it.
    */
-  readonly title?: string
+  readonly title?: string | JSX.Element
 
   /**
    * Whether or not the dialog should be dismissable. A dismissable dialog
@@ -178,9 +178,11 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
       this.setState({ titleId: undefined })
     }
 
-    if (this.props.title) {
+    if (this.props.title != null) {
+      // TODO: how can we get a string-like representation of a JSX element?
+      const id = typeof this.props.title === 'string' ? this.props.title : '???'
       this.setState({
-        titleId: createUniqueId(`Dialog_${this.props.id}_${this.props.title}`),
+        titleId: createUniqueId(`Dialog_${this.props.id}_${id}`),
       })
     }
   }
diff --git a/app/src/ui/merge-branch/merge.tsx b/app/src/ui/merge-branch/merge.tsx
index bd4239bc7..0d0ed27a3 100644
--- a/app/src/ui/merge-branch/merge.tsx
+++ b/app/src/ui/merge-branch/merge.tsx
@@ -19,7 +19,6 @@ import { MergeResultKind } from '../../models/merge'
 import { MergeStatusHeader } from '../history/merge-status-header'
 import { promiseWithMinimumTimeout } from '../../lib/promise'
 import { truncateWithEllipsis } from '../../lib/truncate-with-ellipsis'
-import { DialogHeader } from '../dialog/header'
 
 interface IMergeProps {
@@ -293,18 +292,15 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
     return (
       <Dialog
         id="merge"
+        title={
+          <div className="merge-dialog-header">
+            Merge into <b>{currentBranchName}</b>
+          </div>
+        }
+        dismissable={true}
         onSubmit={this.merge}
       >
-        <DialogHeader
-          title={
-            <div className={'merge-dialog-header'}>
-              Merge into <b>{currentBranchName}</b>
-            </div>
-          }
-          dismissable={true}
-          onDismissed={this.props.onDismissed}
-        />
         <DialogContent>
           <BranchList
```

iAmWillShepherd: Could we use the `key` property JSX elements are supposed to have?

outofambit: hey @shiftkey i think something like that makes sense and i'm happy to open that refactor in another PR to help with cases like these. i think we should merge this separately though.

shiftkey: Moved to https://github.com/desktop/desktop/issues/6032

PR #6033, opened 2018-10-30
Updates Dialog to accept JSX and Merge dialog to use it

iAmWillShepherd:
Refactor of #6027. Adds the ability to pass in JSX into a dialog.


iAmWillShepherd:
Using the `key` property on the JSX element would not work since it could be a number or null.

PR #6034, opened 2018-10-30
add support for debugging jest tests in vscode

shiftkey:
## Overview

This PR re-adds support for being able to debug Desktop's unit tests inside VSCode. 

There's two parts to supporting this in VSCode:

 - the node runtime to attach to - in this case we're using our local version of Electron
 - the program to run - we'll use `jest` with our existing unit test config

When you launch the tests, VSCode's terminal will appear and the rest of the experience looks familiar if you've run the tests from the command line:

But to see the real magic, set a breakpoint inside a test (the gutter to the left of the numbers) like this:

At some point after that, the tests should break on the breakpoint you just defined:

You then have access to the variables and stack trace in the left sidebar:

## Description

In a nutshell, you now have two options:

 - `Jest All` to run all unit tests (the equivalent of `yarn test:unit` from the command line)
 - `Jest Current` to run all unit tests in the current file

I'd recommend `Jest Current` when working on tests, as that avoids running the whole test suite.

If you try and `Jest Current` from a file that doesn't contain any tests, you'll see an error like this:

You can set breakpoints anywhere in the code, but be careful where you `Jest Current` from.

I haven't had a chance to confirm this works the same on Windows, but I think it'll be fine...

## Release notes

Notes: no-notes

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,46 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Jest All",
+      "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron",
+      "program": "${workspaceFolder}/node_modules/jest/bin/jest",
+      "args": [
+        "--silent",
+        "--config",
+        "${workspaceFolder}/jest.config.js"
+      ],
+      "console": "integratedTerminal",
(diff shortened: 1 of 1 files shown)


outofambit:
this works really nicely with https://marketplace.visualstudio.com/items?itemName=Orta.vscode-jest, too!

shiftkey:
Gonna #jolo this in now so that I can use it for writing more tests :tada:

PR #6037, opened 2018-10-30
refactor commit message state handling

outofambit:
## Overview

_supports #6049 and #6013_

This PR refactors our handling of commit message state so that it can be successfully cleared from the `dispatcher`. It also removes the `CommitMessage` components reliance on deprecated lifecycle methods.

**This PR will allow us to reset the commit message in the UI via the existing dispatcher method 'setCommitMessage' (which currently doesn't work in all cases).**

## Description

- removes deprecated [`componentWillMount`](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) from `CommitMessage` component
- removes deprecated [`componentWillReceiveProps`](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) from `CommitMessage` component
- (mostly) centralizes commitMessage state in `gitStore`
- uses a `key` to manage un-mounting and mounting a new `CommitMessage` 

## User Flows tested

- undoing commits with and without coauthors
- switching repositories and not losing commit message drafts
- committing

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -18,7 +18,6 @@ import { ComparisonMode } from '../app-state'
 
 import { IAppShell } from '../app-shell'
 import { ErrorWithMetadata, IErrorMetadata } from '../error-with-metadata'
-import { structuralEquals } from '../../lib/equality'
 import { compare } from '../../lib/compare'
 import { queueWorkHigh } from '../../lib/queue-work'
 
@@ -104,8 +103,6 @@ export class GitStore extends BaseStore {
   private _localCommitSHAs: ReadonlyArray<string> = []
 
   private _commitMessage: ICommitMessage | null = null
-
-  private _contextualCommitMessage: ICommitMessage | null = null
 
   private _showCoAuthoredBy: boolean = false
 
@@ -718,14 +715,6 @@ export class GitStore extends BaseStore {
   }
 
   /**
-   * The commit message to use based on the context of the repository, e.g., the
-   * message from a recently undone commit.
-   */
-  public get contextualCommitMessage(): ICommitMessage | null {
-    return this._contextualCommitMessage
-  }
-
-  /**
    * Gets a value indicating whether the user has chosen to
    * hide or show the co-authors field in the commit message
    * component
@@ -1214,25 +1203,6 @@ export class GitStore extends BaseStore {
     })
   }
 
-  /** Load the contextual commit message if there is one. */
-  public async loadContextualCommitMessage(): Promise<void> {
-    const message = await this.getMergeMessage()
-    const existingMessage = this._contextualCommitMessage
-    // In the case where we're in the middle of a merge, we're gonna keep
-    // finding the same merge message over and over. We don't need to keep
-    // telling the world.
-    if (
-      existingMessage &&
-      message &&
-      structuralEquals(existingMessage, message)
-    ) {
-      return
-    }
-
-    this._contextualCommitMessage = message
-    this.emitUpdate()
-  }
-
   /** Reverts the commit with the given SHA */
   public async revertCommit(
     repository: Repository,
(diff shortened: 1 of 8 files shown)


app/src/ui/changes/changes-list.tsx
```diff
@@ -432,6 +432,7 @@ export class ChangesList extends React.Component<
               : 'Summary (required)'
           }
           singleFileCommit={singleFileCommit}
+          key={this.props.repository.id}
```
shiftkey: What's the context for adding this here? Did you uncover something that we weren't doing well before?

app/src/lib/stores/git-store.ts
```diff
@@ -544,7 +544,7 @@ export class GitStore extends BaseStore {
       }
     }
 
-    this._contextualCommitMessage = {
+    this._commitMessage = {
```
shiftkey: With these changes do we have any actual need for `_contextualCommitMessage`?  The only context I can see being done in `GitStore` is loading the merge message, which was part of our old merge conflicts flow. If there's no need for this now that you're favouring `commitMessage` then I think we can clean this up from the stores and components that need it

shiftkey:
> undoing commits with and without coauthors

There's a behaviour we had in the current app around undoing multiple commits and I'm not sure when that's regressed (`master` is also showing this behaviour, so I'm digging into this now to try and figure out when this changed).

What I was hoping to see: when undoing a commit when you have a commit message already populated, the details are merged together (I think the new commit details are supposed to be appended to the existing details)

What I currently see: the commit summary and description are overwritten with the details of the previous commit.

This becomes acutely painful when you undo several commits, and rather than preserving the details you had in the commits you're left with just the details of the most recent commit you undid.

app/src/lib/stores/git-store.ts
outofambit: ooo good catch! i will 🔥it!

app/src/ui/changes/changes-list.tsx
outofambit: > uses a key to manage un-mounting and mounting a new CommitMessage

the key here forces the component to be remounted when the key changes, and only updates when the key is the same. this simplifies the components internal logic around unmounting and saving the draft commit message to the stores

shiftkey:
@outofambit I've traced this same behaviour back to 3307817aad49e6342e4acbeb760d1a43cb48bccf when it was originally implemented, so I must be thinking of the classic apps. I'll open a fresh issue to expand on that.

app/src/ui/changes/commit-message.tsx
```diff
@@ -109,112 +104,46 @@ export class CommitMessage extends React.Component<
...
-    } else if (
-      initializing ||
-      this.props.repository.id !== nextProps.repository.id
-    ) {
-      // We're either initializing (ie being mounted) or someone has switched
-      // repositories. If we receive a message we'll take it
-      if (nextProps.commitMessage) {
-        // Don't update dispatcher here, we're receiving it, could cause never-
-        // ending loop.
+    if (prevProps.commitMessage !== this.props.commitMessage) {
```
shiftkey: Should this comparison look at the contents of `commitMessage`, rather than using `===`, to avoid unnecessarily invoking `setState` below?



Something like `shallowEquals` should do the trick, based on a quick test on my end:

```ts
if (shallowEquals(prevProps.commitMessage, this.props.commitMessage)) {
  return
}
```

outofambit:
@shiftkey 🍏 😄

app/src/ui/changes/commit-message.tsx
outofambit: love it!

outofambit: done in 6b9ca6c

outofambit:
@shiftkey 🍇

shiftkey:
I'm going to leave this for the full 24-hour review period post-approval to ensure I give others a chance to review the changes, and also because I believe the remaining work to close #6049 after this is merged is minor - @outofambit can clarify if that's not the case.

PR #6051, opened 2018-10-31
Add automatic switching between Dark Mode and Light 

### LINKED ISSUE #5037: Add automatic switching between Dark Mode and Light Mode or ability to change it programmatically.
**Request**
I love the new the dark mode and I would love it if it automatically switched with the change between Light and Dark mode on macOS. 

We could add an observer to check which theme the the OS is in, maybe add an option on the prefs to match system mode? You can check the current theme from defaults, here's an example in [swift](https://gist.github.com/nehayward/d5e022ced5487944b7850d622350c1c3).

Alternatively, we could add an option to change this from the CLI or adding a preference file and let the user be able to switch between modes. 

@niik Thanks!

say25:
## Overview
Add automatic switching between Dark Mode and Light Mode or ability to change it programmatically.

**Closes #5037**

## Description

- Add automatic switching between Dark Mode and Light Mode (macOS Mojave Dark Mode).

ezgif-3-dabd0ed50fb3

## Release notes

Notes: Add automatic switching between Dark Mode and Light Mode (macOS Mojave Dark Mode).


say25:
@desktop/engineering not sure where the best place to put the currently commented code would be. We would basically need to subscribe to the electron event `systemPreferences.subscribeNotification('AppleInterfaceThemeChangedNotification', updateThemeBasedOnSystem);` on app start up (maybe DARWIN conditional). Opening an early PR for early feedback. I also think the `electron.d.ts` could be slightly broken :/

app/src/ui/preferences/appearance.tsx
```diff
@@ -32,11 +52,54 @@ export class Appearance extends React.Component<IAppearanceProps, {}> {
     }
   }
 
+  private onAutomaticallySwitchThemeChanged = (
+    event: React.FormEvent<HTMLInputElement>
+  ) => {
+    const value = event.currentTarget.checked
+
+    // const updateThemeBasedOnSystem = () => {
```
shiftkey: I'm not sure where you plan to encapsulate this logic, but I'd check out [`update-store.ts`](https://github.com/desktop/desktop/blob/master/app/src/ui/lib/update-store.ts) as a reference:

 - exports a singleton (single instance) of a class that wraps the relevant Electron APIs
 - it can subscribe to electron lifecycle events to cleanup after itself
 - other code can `import` this module if they want - they get the same instance

app/src/ui/lib/application-theme.ts
```diff
@@ -54,3 +54,26 @@ export function getPersistedThemeName(): string {
...
+// in localStorage.
+const automaticallySwitchApplicationThemeKey = 'autoSwitchTheme'
+
+/**
+ * Load the whether or not the user wishes to automatically switch the selected theme from the persistent
+ * store (localStorage). If no theme is selected the default
+ * theme will be returned.
+ */
+export function getAutoSwitchPersistedTheme(): boolean {
+  return localStorage.getItem(automaticallySwitchApplicationThemeKey) === 'true'
```
shiftkey: We have `getBoolean` and `setBoolean` now that should make some of this plumbing unneccesary - see #5979 for more info

shiftkey:
@say25 thanks for opening this up!

> not sure where the best place to put the currently commented code would be. 

As this is a UI-specific feature, what about a module under `app/src/ui/lib`? `systemPreferences` is a main-process specific API, but you could use `remote` to access it - that's how we've done it for the title bar interaction:

https://github.com/desktop/desktop/blob/412959e45b517f8d5a19e72950f36fd651bb3f59/app/src/ui/window/title-bar.tsx#L55-L58

And like `UpdateStore` you could define an event emitter interface to bubble up the event to the listeners.

Alternatively, you could have this live in `app/src/main-process` and it could just emit IPC calls into the renderer.

Inside `AppWindow` you can see the example of sending messages:

```ts
this.window.on('blur', () => this.window.webContents.send('blur'))
```

And `app/src/ui/index.tsx` has the example of the subscription:

```ts
ipcRenderer.on('blur', () => {
  dispatcher.setAccessKeyHighlightState(false)
  dispatcher.setAppFocusState(false)
})
```

> on app start up (maybe DARWIN conditional).

The module could be conditionalized using `__DARWIN__` to subscribe or no-op (for non-macOS platforms). You might even be able to read the OS build for macOS and not bother with the subscription for builds earlier than Mojave.

app/src/ui/lib/os-dark-mode.ts
```diff
@@ -0,0 +1,34 @@
...
+  return updateThemeBasedOnSystem()
+}
+
+function updateThemeBasedOnSystem() {
+  const useDarkMode = remote.systemPreferences.isDarkMode()
+  const automaticallySwitchTheme = getAutoSwitchPersistedTheme()
+
+  if (automaticallySwitchTheme) {
+    // const theme = useDarkMode ? ApplicationTheme.Dark : ApplicationTheme.Light
+    // Fire theme to dispatcher somehow :/
```
say25: @shiftkey I got stuck here.... I'm not quiet sure how to fire this back to the dispatcher

shiftkey: I ended up going with a different approach to what you had in mind, and I'll walk through it to contrast - hopefully it makes sense!

First, I defined these two helpers around dark mode detection in `app/src/ui/lib/dark-theme.ts` - there were a few checks for this around, so centralizing it helped with my sanity:

```ts
import { remote } from 'electron'
import { isMojaveOrLater } from '../../lib/get-os'

export function supportsDarkMode() {
  if (!__DARWIN__) {
    return false
  }

  if (!isMojaveOrLater()) {
    return false
  }

  return true
}

export function isDarkModeEnabled() {
  if (!supportsDarkMode()) {
    return false
  }

  return remote.systemPreferences.isDarkMode()
}
```

Next, I created the `ThemeChangeMonitor` in `app/src/ui/lib/theme-change-monitor.ts ` whose sole job it is to report when the theme has changed. I think I ported all the logic over, but use this as a starting point for what you're looking to achieve:

```ts
import { remote } from 'electron'
import { ApplicationTheme } from './application-theme'
import { IDisposable, Disposable, Emitter } from 'event-kit'
import { supportsDarkMode, isDarkModeEnabled } from './dark-theme'

class ThemeChangeMonitor implements IDisposable {
  private readonly emitter = new Emitter()

  private subscriptionID: number | null = null

  constructor() {
    this.subscribe()
  }

  dispose() {
    this.unsubscribe()
  }

  private subscribe = () => {
    if (!supportsDarkMode()) {
      return
    }

    this.subscriptionID = remote.systemPreferences.subscribeNotification(
      'AppleInterfaceThemeChangedNotification',
      this.onThemeNotificationFromOS
    )
  }

  private onThemeNotificationFromOS = (event: string, userInfo: any) => {
    // if there is details in the payload from the OS about the theme change we should
    // use that rather than query again for the dark theme state
    const darkModeEnabled = isDarkModeEnabled()

    const theme = darkModeEnabled
      ? ApplicationTheme.Dark
      : ApplicationTheme.Light
    this.emitThemeChanged(theme)

    // do we need to re-subscribe to the event after it was raised? if not, these
    // lines are unnecessary and can be removed
    this.unsubscribe()
    this.subscribe()
  }

  private unsubscribe = () => {
    if (this.subscriptionID !== null) {
      remote.systemPreferences.unsubscribeNotification(this.subscriptionID)
    }
  }

  // subscribers call this function and pass in the callback they want to run
  public onThemeChanged(fn: (theme: ApplicationTheme) => void): Disposable {
    return this.emitter.on('theme-changed', fn)
  }

  private emitThemeChanged(theme: ApplicationTheme) {
    this.emitter.emit('theme-changed', theme)
  }
}

// this becomes our singleton that we can subscribe to from anywhere
export const themeChangeMonitor = new ThemeChangeMonitor()

// this ensures we cleanup any existing subscription on exit
remote.app.on('will-quit', () => {
  themeChangeMonitor.dispose()
})
```

Last, I have the ability to subscribe to this from anywhere in the app - I chose to do this in the `AppStore` constructor (because that's got easy access to the state) but you might have a better idea:

```ts
import { themeChangeMonitor } from '../../ui/lib/theme-change-monitor'

// ...

    themeChangeMonitor.onThemeChanged(theme => {
      if (this.automaticallySwitchTheme) {
        this.selectedTheme = theme
        this.emitUpdate()
      }
    })
```

say25: Makes perfect sense. Thanks for taking the time to explain the thought process. I was also thinking we would have to do something with an `Emitter` and a singleton. Just wanted to make sure I wasn't crazy/over thinking it. Ill work on the implementation soon.

say25:
Merge conflict resolved. This should be `Ready for Review`

say25:
Merge conflict from #6109 resolved. @desktop/engineering this should be `Ready for Review`.

app/src/ui/lib/dark-theme.ts
```diff
@@ -0,0 +1,24 @@
+import { remote } from 'electron'
+import { isMojaveOrLater } from '../../lib/get-os'
+
+export function supportsDarkMode() {
+  if (!__DARWIN__) {
+    return false
+  }
+
+  if (!isMojaveOrLater()) {
```
shiftkey: :art: the rest of this function can be flattened to `return isMojaveOrLater()`

app/src/ui/preferences/appearance.tsx
```diff
@@ -30,13 +50,51 @@ export class Appearance extends React.Component<IAppearanceProps, {}> {
...
+    this.props.onAutomaticallySwitchThemeChanged(value)
   }
 
   public render() {
     const selectedIndex =
       this.props.selectedTheme === ApplicationTheme.Dark ? 1 : 0
+    const doesSupportDarkMode = supportsDarkMode()
 
-    return (
+    return doesSupportDarkMode ? (
```
shiftkey: Rather than repeating the theme selection, can we trying decomposing the main `render` into smaller 
 functions which may return `null` to signal to React that nothing needs to be rendered:

```tsx
public render() {
  return (
    <DialogContent>
      {this.renderThemeOptions()}
      {this.renderAutoSwitcherOption()}
    </DialogContent>
  )
}
```

Where each of these functions is focused on a part of the overall JSX:

```jsx
private renderThemeOptions(): JSX.Element {
  // 
}

private renderAutoSwitcherOption(): JSX.Element | null {
  const isDarkModeSupported = supportsDarkMode()
  if (!isDarkModeSupported) {
    return null
  }

  //
}

File: app/src/ui/preferences/appearance.tsx [hunk 6]
```diff
@@ -21,7 +30,18 @@ const themes: ReadonlyArray<ISegmentedItem> = [
...
-export class Appearance extends React.Component<IAppearanceProps, {}> {
+export class Appearance extends React.Component<
+  IAppearanceProps,
+  IAppearanceState
+> {
+  public constructor(props: IAppearanceProps) {
+    super(props)
+
+    this.state = {
+      automaticallySwitchTheme: this.props.automaticallySwitchTheme,
```
shiftkey: 💭 I'm curious if we actually need to track state here, given we don't for theme changes. I need to have a read of how stateful we're making other Preferences components, because it's not clear to me which stuff should be saved versus immediately updated...

shiftkey:
@say25 do you have any pointers for reviewers to help with testing this out on Mojave? It'd be nice to be able to quickly verify that the theme changes in macOS are propagated to Desktop, and that the preferences setting is respected...

say25:
@shiftkey, feedback addressed.

Test Notes on Mojave:
1. Open Desktop, check Auto Switch theme, see that theme is consistent with OS theme.
1. Close Desktop change OS theme, reopen, see that theme is consistent with OS theme.
1. Change Desktop theme to not OS theme, see that Auto Switch is unchecked and new Desktop theme applied.
1. With Auto Switch still off, change OS theme see that Desktop theme remains unchanged.

Test Notes on Not Mojave:
1. No changes to Desktop

app/src/lib/get-os.ts
```diff
@@ -16,3 +16,16 @@ export function getOS() {
...
   }
 }
+
+/** See the OS we're currently running on is at least Mojave. */
+export function isMojaveOrLater() {
+  if (__DARWIN__) {
+    const parser = new UAParser()
+    const os = parser.getOS()
+    // Check that it is Mojave or later
+    if (os.version && os.version.split('.')[1] > '13') {
```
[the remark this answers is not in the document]
say25: I was thinking that too but how should it behave in the case above? I thought it would return false anyway as it is unable to verify Mojave :/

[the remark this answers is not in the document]
shiftkey: >  I thought it would return false anyway as it is unable to verify Mojave :/

Agreed, just thinking out loud here how much we should rely on JS truthiness here.

If we wanted to be explicit here and not trust `os.version` so much, we could do something like this:

```ts
if (os.version == null) { 
  return false
}

const parts = os.version.split('.')
if (parts.length !== 2) {
  // unknown version format, giving up
  return false
}

return parts[1] >  '13'
```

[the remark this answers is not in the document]
say25: Sounds good to me. Shall make the changes shortly

app/src/ui/preferences/appearance.tsx
say25: @shiftkey, from what I can tell State is needed as this.props is readonly :/

shiftkey: I believe if you're updating `IAppState.automaticallySwitchTheme` through the dispatcher, and then emitting an update, that will trigger a re-render of the `Appearance` component with the updated prop value. That's why we mark the fields in `props` as `readonly` (because components shouldn't be updating them).

say25:
Given that #6069 just got approved (an hopefully shall merge soon) I'll remove the electron 2.0.9 pollyfill at that time

app/src/ui/preferences/appearance.tsx
shiftkey: @say25 all good, feel free to bump this PR when you get around to it

say25:
@shiftkey installed macOS Mojave on her machine, can confirm this works without state. Changed the check you sugguested to `parts.length < 2` as macOS is generally 10.X.Y format though we only care about the middle digit I relaxed the constraint to < 2

say25:
With #6069 merged down to `master`, I have merged `master` into this branch and removed the pollyfill shim.

tierninho:

- [x] toggling Auto-theme checkbox is functional
- [x] switching OS theme is reflected in Desktop checkbox/theme
- [x] Switch theme in Desktop toggles off auto checkbox if it doesn't match OS
- [x] theme key:value is updated in local storage

Notes: cannot test if the setting in Desktop is saved unless this feature hits beta.

Build/OS: 1.5.1-beta0, mac

say25:
@shiftkey welcome back :D, noticed a merge conflict on this a couple days ago, resolved so it isn't currently blocked by me.

app/src/lib/get-os.ts
```diff
@@ -16,3 +16,24 @@ export function getOS() {
...
   }
 }
+
+/** See the OS we're currently running on is at least Mojave. */
+export function isMojaveOrLater() {
+  if (__DARWIN__) {
+    const parser = new UAParser()
+    const os = parser.getOS()
+
+    if (os.version == null) {
```
iAmWillShepherd: `os.version == null` should be `os.version === undefined` since
`os.version` is typed as `string | undefined` and we tend to favor explicit checks.

app/src/lib/get-os.ts
```diff
@@ -16,3 +16,24 @@ export function getOS() {
...
+      return false
+    }
+
+    const parts = os.version.split('.')
+    if (parts.length < 2) {
```
iAmWillShepherd: I don't think `parts.length` should ever be anything other than `3` so we can change this to `parts !== 3`

app/src/lib/get-os.ts
```diff
@@ -16,3 +16,24 @@ export function getOS() {
...
+export function isMojaveOrLater() {
+  if (__DARWIN__) {
+    const parser = new UAParser()
+    const os = parser.getOS()
+
+    if (os.version == null) {
+      return false
+    }
+
+    const parts = os.version.split('.')
```
iAmWillShepherd: I think we can be more descriptive here by pulling these out into variables. `const [major, minor] = os.version.split('.')` Doing it this way makes it clear what each part of the version means.

app/src/lib/get-os.ts
```diff
@@ -16,3 +16,24 @@ export function getOS() {
...
+    if (parts.length < 2) {
+      // unknown version format, giving up
+      return false
+    }
+
+    return parts[1] > '13'
```
iAmWillShepherd: We should also check to make sure `parts[0] === 10` (or `major === 10` if you go with the change suggested above.)

app/src/ui/lib/theme-change-monitor.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const theme = darkModeEnabled
+      ? ApplicationTheme.Dark
+      : ApplicationTheme.Light
+    this.emitThemeChanged(theme)
+  }
+
+  private unsubscribe = () => {
+    if (this.subscriptionID !== null) {
+      remote.systemPreferences.unsubscribeNotification(this.subscriptionID)
```
iAmWillShepherd: `this.subscriptionID` should probably be set to `null` after you unsubscribe from the notification.

app/src/ui/lib/theme-change-monitor.ts
```diff
@@ -0,0 +1,60 @@
...
+    return this.emitter.on('theme-changed', fn)
+  }
+
+  private emitThemeChanged(theme: ApplicationTheme) {
+    this.emitter.emit('theme-changed', theme)
+  }
+}
+
+// this becomes our singleton that we can subscribe to from anywhere
+export const themeChangeMonitor = new ThemeChangeMonitor()
```
iAmWillShepherd: In how many places do we have to subscribe to this object? If it's only in the `AppStore`, I think we can axe this and setup our single instance there.

app/src/ui/lib/theme-change-monitor.ts
```diff
@@ -0,0 +1,60 @@
...
+
+// this ensures we cleanup any existing subscription on exit
+remote.app.on('will-quit', () => {
```
iAmWillShepherd: Not sure if we need to do this...
Can you provide any insight into this, @niik?

app/src/ui/preferences/appearance.tsx
```diff
@@ -30,20 +35,65 @@ export class Appearance extends React.Component<IAppearanceProps, {}> {
...
+    this.props.onAutomaticallySwitchThemeChanged(false)
+  }
+
+  private onAutomaticallySwitchThemeChanged = (
+    event: React.FormEvent<HTMLInputElement>
+  ) => {
+    const value = event.currentTarget.checked
+
+    if (value) {
+      const usingDarkMode = isDarkModeEnabled()
```
iAmWillShepherd: Since we're only using this variable in one place, we can just call `isDarkModeEnabled()` inline below.

app/src/ui/preferences/appearance.tsx
```diff
@@ -30,20 +35,65 @@ export class Appearance extends React.Component<IAppearanceProps, {}> {
...
           selectedIndex={selectedIndex}
           onSelectionChanged={this.onSelectedThemeChanged}
         />
-      </DialogContent>
+      </Row>
+    )
+  }
+
+  public renderAutoSwitcherOption() {
+    const doesSupportDarkMode = supportsDarkMode()
```
iAmWillShepherd: We can axe this line and call the function inline below.

app/src/ui/lib/theme-change-monitor.ts
say25: It is only currently used in the `AppStore`. Were you thinking we do something like `const themeChangeMonitor = new ThemeChangeMonitor()` in the constructor of `AppStore`? 

I can't currently think of a reason why we would want to watch monitor theme changes elsewhere but we at least get the ability to do so this way consistently to my knowledge 🤔

say25:
🎉 
light-gif

billygriffin:
🎉 This is super cool, thanks @say25! May I suggest we hold off on merging this in until https://github.com/desktop/desktop/pull/6660 gets into prod in 1.6.1 next week. I'd like to let this have at least a week in beta prior to shipping to prod in case there's any unexpected goofiness. Does that seem reasonable?

say25:
Seems legit to me.

shiftkey:
@billygriffin @say25 because we've already [merged in a bit of new work](https://github.com/desktop/desktop/compare/release-1.6.0...development) since the release I suspect we'll just cherry-pick the specific changes we want to go out and have a short beta cycle for `1.6.1`, and shuffle this and the other work into `1.6.2`.

PR #6053, opened 2018-10-31
add some new Git APIs to improve our Git LFS support

shiftkey:
## Overview

As part of #6013 we uncovered that we lack some context about how a repository is configured for Git LFS. In particular we didn't know whether a file would be tracked by Git LFS, so this PR adds support for doing this in two ways:

 - directly querying a repository and passing a relative path
 - a convenience function that takes in a list of relative paths and will return only the files not covered by Git LFS

Shout out to @Daniel-McCarthy for digging into this and helping uncover the gaps with how we detect our Git LFS support, and shout out to @ttaylorr for pointing me to `git check-attr` to help with the underlying querying of the Git repository.

## Description

The internals for this are split into two parts:

 - `isTrackedByLFS` - checking a file path is covered by Git LFS
 - `filesNotTrackedByLFS` - pass it a list of files, get back only those which aren't covered by Git LFS

For the workflows in #6013 I think callers should be fine with what `filesNotTrackedByLFS` provides. @Daniel-McCarthy let me know if this makes sense for what you're building out.

There's a few tests covering this currently to help guide the API, but I didn't want to spend too much time on it in case it's not what we need for #6013. As we get to the bottom of #6013 I imagine we'll bake in more tests.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/lfs.ts
+++ b/app/src/lib/git/lfs.ts
@@ -50,10 +50,24 @@ export async function isTrackedByLFS(
   path: string
 ): Promise<boolean> {
   const { stdout } = await git(
-    ['check-attr', '-a', path],
+    ['check-attr', 'filter', path],
     repository.path,
     'checkAttrForLFS'
   )
+
+  // "git check-attr -a" will output every filter it can find in .gitattributes
+  // and it looks like this:
+  //
+  // README.md: diff: lfs
+  // README.md: merge: lfs
+  // README.md: text: unset
+  // README.md: filter: lfs
+  //
+  // To verify git-lfs this test will just focus on that last row, "filter",
+  // and the value associated with it. If nothing is found in .gitattributes
+  // the output will look like this
+  //
+  // README.md: filter: unspecified
 
   const lfsFilterRegex = /(.*): filter: lfs/
 
--- a/app/src/lib/git/lfs.ts
+++ b/app/src/lib/git/lfs.ts
@@ -1,5 +1,6 @@
 import { git } from './core'
 import { Repository } from '../../models/repository'
+import { caseInsensitiveCompare } from '../compare'
 
 /** Install the global LFS filters. */
 export async function installGlobalLFSFilters(force: boolean): Promise<void> {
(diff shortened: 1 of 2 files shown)


Daniel-McCarthy:
This looks fantastic! This should be perfect for what we immediately need.
Thank you for taking the time to implement this, I greatly appreciate that!

shiftkey:
@Daniel-McCarthy after doing a bit of noodling on bringing these changes into #6013 I ended up with something like this for the flow to identify these problem files:

```diff
diff --git a/app/src/ui/changes/sidebar.tsx b/app/src/ui/changes/sidebar.tsx
index 7baa71f4b..e0eba50ab 100644
--- a/app/src/ui/changes/sidebar.tsx
+++ b/app/src/ui/changes/sidebar.tsx
@@ -24,6 +24,9 @@ import { openFile } from '../../lib/open-file'
 import { ITrailer } from '../../lib/git/interpret-trailers'
 import { Account } from '../../models/account'
 import { PopupType } from '../../models/popup'
+import { enableFileSizeWarningCheck } from '../../lib/feature-flag'
+import { filesNotTrackedByLFS } from '../../lib/git/lfs'
+import { getLargeFilePaths } from '../../lib/large-files'
 
 /**
  * The timeout for the animation of the enter/leave animation for Undo.
@@ -114,11 +117,37 @@ export class ChangesSidebar extends React.Component<IChangesSidebarProps, {}> {
     }
   }
 
-  private onCreateCommit = (
+  private onCreateCommit = async (
     summary: string,
     description: string | null,
     trailers?: ReadonlyArray<ITrailer>
   ): Promise<boolean> => {
+    if (enableFileSizeWarningCheck()) {
+      const overSizedFiles = await getLargeFilePaths(
+        this.props.repository,
+        this.props.changes.workingDirectory,
+        100
+      )
+
+      const filesIgnoredByLFS = await filesNotTrackedByLFS(
+        this.props.repository,
+        overSizedFiles
+      )
+
+      if (filesIgnoredByLFS.length !== 0) {
+        this.props.dispatcher.showPopup({
+          type: PopupType.OversizedFiles,
+          oversizedFiles: filesIgnoredByLFS,
+          commitSummary: summary,
+          commitDescription: description,
+          repository: this.props.repository,
+          trailers: trailers,
+        })
+
+        return false
+      }
+    }
+
     return this.props.dispatcher.commitIncludedChanges(
       this.props.repository,
       summary,
```

Note that:

 - we've pulled this logic up into `sidebar.tsx` to avoid having to plumb functions down the list using props
 - `getLargeFilePaths` is now a standalone function that can be used outside of components (makes it easier to write tests against)

```ts
export async function getLargeFilePaths(
  repository: Repository,
  workingDirectory: WorkingDirectoryStatus,
  maximumSizeMB: number
): Promise<ReadonlyArray<string>> {
 ...
}
```

I'll put this code aside for now until we get this one reviewed and merged - I just wanted to share where I think #6013 should head after that lands...

app/src/lib/git/lfs.ts
```diff
@@ -34,3 +35,76 @@ export async function isUsingLFS(repository: Repository): Promise<boolean> {
...
+  //
+  // To verify git-lfs this test will just focus on that last row, "filter",
+  // and the value associated with it. If nothing is found in .gitattributes
+  // the output will look like this
+  //
+  // README.md: filter: unspecified
+
+  const lfsFilterRegex = /(.*): filter: lfs/
+
+  const match = lfsFilterRegex.exec(stdout)
```
ttaylorr: It looks like you're calling `git check-attr filter` instead of `git check-attr -a`, so (if my reading of the surrounding code is correct) I think that Git will only report about the `filter` attribute. Thus, I think this regex match can removed in favor of simply comparing to and dropping the `: filter: lfs` suffix.

shiftkey: @ttaylorr I'm using the capture as well as a sanity check that it's the relative file path I passed in as an argument, but if I decide that I don't need to do that I can simplify the regex 👍

PR #6067, opened 2018-11-01
figure out when outside merge occurred

### LINKED ISSUE #6045: show conflict resolution flow when conflicted merge is started elsewhere
Right now (on `master`, after #5809), we only show the conflict resolution flow when the conflicted merge is started from within Desktop. This should be flow should be expanded to be triggered when a user starts a conflicted merge somewhere else (like a terminal) and then opens desktop in that repository

outofambit:
## Overview

**Closes #6045 **

## Description

- [x] detect when conflicted merge is in progress
- [ ] get branch names for dialog

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -173,6 +173,7 @@ import { readEmoji } from '../read-emoji'
 import { GitStoreCache } from './git-store-cache'
 import { MergeConflictsErrorContext } from '../git-error-context'
 import { setNumber, setBoolean, getBoolean, getNumber } from '../local-storage'
+import { MergeConflictsDialog } from '../../ui/merge-conflicts'
 
 /**
  * As fast-forwarding local branches is proportional to the number of local
@@ -1607,6 +1608,31 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
       return { workingDirectory, selectedFileIDs, diff }
     })
+
+    const inConflictedMerge = status.workingDirectory.files.some(f => {
+      return (
+        f.status === AppFileStatus.Conflicted ||
+        f.status === AppFileStatus.Resolved
+      )
+    })
+
+    // if any files are conflicted or resolved,
+    // we trigger the conflict resolution flow
+    // (if we're not already in it)
+    if (
+      inConflictedMerge &&
+      (this.currentPopup === null ||
+        (this.currentPopup !== null &&
+          this.currentPopup.type !== PopupType.MergeConflicts &&
+          this.currentPopup.type !== PopupType.AbortMerge))
+    ) {
+      this._showPopup({
+        type: PopupType.MergeConflicts,
+        repository: repository,
+        currentBranch: 'currentBranch',
+        theirBranch: 'theirBranch',
+      })
+    }
     this.emitUpdate()
 
     this.updateChangesDiffForCurrentSelection(repository)


outofambit:
`git branch --points-at=MERGE_HEAD` works but only if one branch has the same tip, otherwise we get multiple results

app/src/lib/git/branch.ts
```diff
@@ -117,3 +118,21 @@ async function checkIfBranchExistsOnRemote(
...
+  commitish: string
+): Promise<Array<string>> {
+  const args = [
+    'branch',
+    `--points-at=${commitish}`,
+    '--format=%(refname:short)',
+  ]
+  const { stdout } = await git(args, repository.path, 'branchPointedAt')
+  // split along newlines, which should just be branch names
+  const captures = getCaptures(stdout, /\s*(.+)\n/g)
```
shiftkey: If we use `--format` to control the output, we can avoid the overhead of regexes here. An advanced example of this is how `for-each-ref.ts` enumerates the list of branches - this feels similar to what you're looking for here: https://github.com/desktop/desktop/blob/f2c460236c9bf2cbfeff3c0df64273316dedc1ce/app/src/lib/git/for-each-ref.ts#L17

app/src/lib/stores/app-store.ts
```diff
@@ -1607,6 +1608,41 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      )
+    })
+
+    // if any files are conflicted or resolved,
+    // we trigger the conflict resolution flow
+    // (if we're not already in it)
+    if (
+      status.currentBranch !== undefined &&
+      inConflictedMerge &&
+      (this.currentPopup === null ||
```
shiftkey: Are we able to extract this part of the check out to something more readable, and indicate what we're looking for here?

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -24,8 +24,8 @@ interface IMergeConflictsDialogProps {
   readonly openFileInExternalEditor: (path: string) => void
   readonly externalEditorName?: string
   readonly openRepositoryInShell: (repository: Repository) => void
-  readonly currentBranch: string
-  readonly theirBranch: string
+  readonly ourBranch: string
+  readonly theirBranch?: string
```
shiftkey: 📝 when might this be undefined?

app/src/lib/stores/app-store.ts
```diff
@@ -1607,6 +1608,41 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        (this.currentPopup !== null &&
+          this.currentPopup.type !== PopupType.MergeConflicts &&
+          this.currentPopup.type !== PopupType.AbortMerge))
+    ) {
+      const possibleTheirsBranches = await getBranchesPointedAt(
+        repository,
+        'MERGE_HEAD'
```
shiftkey: Is there a chance that `MERGE_HEAD` might not exist, and we'd need to handle this exit code reported from Git (this is on a repository not addressing merge conflicts):

```
$ git branch --points-at=MERGE_HEAD
error: malformed object name 'MERGE_HEAD'
$ echo $?
129
```

outofambit: it would be hard, to get in that state, but if we do its now handled (7a8d96a)

app/src/lib/git/branch.ts
outofambit: good call! b2fdd57

outofambit:
@shiftkey 🍏

app/src/lib/stores/app-store.ts
```diff
@@ -1607,13 +1608,62 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          : undefined
+      const ourBranch = status.currentBranch
+      this._showPopup({
+        type: PopupType.MergeConflicts,
+        repository,
+        ourBranch,
+        theirBranch,
+      })
+    } catch (e) {
+      log.info(e)
```
shiftkey: Can you also provide a summary message here about the context alongside logging the error?

Or if we don't need it and can push the error handling into `getBranchesPointedAt` that will simplify things.

app/src/lib/stores/app-store.ts
shiftkey: 🤔 I was thinking of a different approach here,  unless you care about the error and want to log it...

 - handle the exit code inside  `getBranchesPointedAt ` for a malformed object name provided by `commitish` , using `successExitCodes`
 - return an empty array of branches, rather than letting the error bubble up to the caller

The benefits of guarding in the Git function itself is that other callers who may provide an invalid `committish` won't have to handle the error.

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -24,8 +24,9 @@ interface IMergeConflictsDialogProps {
   readonly openFileInExternalEditor: (path: string) => void
   readonly externalEditorName?: string
   readonly openRepositoryInShell: (repository: Repository) => void
-  readonly currentBranch: string
-  readonly theirBranch: string
+  readonly ourBranch: string
+  /* `undefined` when we didn't know the branch name upstream */
```
shiftkey: Is `upstream` the right phrase here (relating to remotes and tracking branches) or is this related to the context of the merge (being able to convert the `MERGE_HEAD` commit ID to something meaningful to the end user)?

app/src/lib/stores/app-store.ts
```diff
@@ -1607,13 +1608,62 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    if (!inConflictedMerge) {
+      return
+    }
+    if (status.currentBranch === undefined) {
+      return
+    }
+    // if we're already in the conflict merge resolution flow, bail
+    if (
+      this.currentPopup !== null &&
+      this.currentPopup.type !== PopupType.MergeConflicts &&
```
shiftkey: This check seems to be for if we're not seeing the `MergeConflicts` popup and also not seeing in the `AbortMerge` popup - which cannot both be true - so the code below runs whenever `AppStore._loadStatus` fires.

I gather that's what the comment above is for, but it'd be great if we were expressive enough that we didn't need comments, like this:

```ts
if (this.currentPopup === null) {
  return
}

const mergeConflictFlowActive =
  this.currentPopup.type === PopupType.MergeConflicts ||
  this.currentPopup.type === PopupType.AbortMerge

if (mergeConflictFlowActive) {
  return
}
```

app/src/lib/stores/app-store.ts
outofambit: @shiftkey would prefer that `getBranchesPointedAt` returned `null` when it has received a malformed `committish`? as the caller, i would like to know whether there are no branches or if the ref is bad. (i think if somehow `MERGE_HEAD` doesn't exist, we don't want to show the popup)

shiftkey: You could return `null` in that case

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
outofambit: ah yeah, poor word choice there

outofambit:
@shiftkey 🍏🍏

shiftkey:
@outofambit are you also able to merge `master` into here so that you're using working with the latest changes? I was testing this flow and thought that something in here had regressed #6059 - turns out the fix wasn't in this branch...

outofambit:
@shiftkey updated!

app/src/lib/stores/app-store.ts
```diff
@@ -1606,13 +1607,63 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
     this.emitUpdate()
 
     this.updateChangesDiffForCurrentSelection(repository)
 
     return true
   }
 
+  /** starts the conflict resolution flow, if appropriate */
+  private async _triggerMergeConflictsFlow(
```
shiftkey: One thing I've forgotten to ask about - should this flow be feature-flagged behind `enableMergeConflictsDialog()`? The old version of the app didn't show any UI in this situation, and I think we should do the same here.

outofambit: that's a good question. this will run but the popup won't show because the feature flag is checked when the app is rendering, so i think we're good (https://github.com/desktop/desktop/blob/4c0aaa7e4b36e1ca34c5a17da0df3c8c71a177e7/app/src/ui/app.tsx#L1334)

shiftkey: But a dialog is still shown on the `else` path:

https://github.com/desktop/desktop/blob/15e4f99570eb90a20c3689f4c532841e3a6b2d7e/app/src/ui/app.tsx#L1367-L1375

shiftkey:
@outofambit it's not clear to me how these two PRs are related - can you elaborate?

outofambit:
@shiftkey this PR refactors the `Popup` type for the merge conflict and abort merge dialogs, which #6096 needs to account for since it relies on those types. (Specifically, there is work to do to handle the possibly undefined `theirBranch` and how that affects the merge success banner for this case.)

app/src/lib/stores/app-store.ts
outofambit: 59e4773

PR #6072, opened 2018-11-01
add support for reporting a binary file is in a conflicted state

shiftkey:
## Overview

While I dig into #6063 about the Git internals, this PR helps to adapt the shape of the data returned by `getStatus` - using a number to represent the conflicted files isn't helpful for binary files, so I've created a new shape:

```ts
export type ConflictStatus =
  | {
      readonly kind: 'text'
      readonly conflictMarkerCount: number
    }
  | {
      readonly kind: 'binary'
    }
```

## Description

- This PR just adapts the existing code to the new shape of data - we're not surfacing binary conflicted files currently, and they'll continue to be marked as "resolved" in the app
- I've not done any of the UI work around how this affects the conflict resolution flow, but there's a placeholder list entry now to support when we are able to bubble these up.
- I'm not a fan of the `throw new Error` I've added in here to ensure an unrepresentable state is not encountered, but it'd be cool to maybe figure that out once I've gotten #6063 ready for review
 - 38c3a2ce7 converts `AppFileStatus` to string enums for readability, but it also results in a bit less JS being emitted because it no longer uses numbers for the backing store

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
+++ b/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
@@ -11,6 +11,7 @@ import {
   WorkingDirectoryStatus,
   WorkingDirectoryFileChange,
   AppFileStatus,
+  ConflictStatus,
 } from '../../models/status'
 import { Octicon, OcticonSymbol } from '../octicons'
 import { PathText } from '../lib/path-text'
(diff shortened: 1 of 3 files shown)


app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -133,27 +134,33 @@ export class MergeConflictsDialog extends React.Component<
...
-          <PathText path={path} availableWidth={200} />
-          <div className="file-conflicts-status">{message}</div>
-        </div>
-        <Button onClick={onOpenEditorClick}>
-          {this.editorButtonString(editorName)}
-        </Button>
-      </li>
-    )
+    if (conflictStatus.kind === 'binary') {
+      return <div>TODO</div>
```
outofambit: can we remove this? i don't see a reason to surface this in the UI (except for dev debugging).

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -165,9 +172,15 @@ export class MergeConflictsDialog extends React.Component<
       case AppFileStatus.Resolved:
         return this.renderResolvedFile(file.path)
       case AppFileStatus.Conflicted:
+        if (file.conflictStatus === null) {
+          throw new Error(
```
outofambit: yeah i agree this isn't ideal, but I think to remove it we'd need to restructure status parsing/construction more. i think the solution is to move `conflictStatus` out of `WorkingDirectoryFileChange` and into `AppFileStatus`, but then `AppFileStatus` would have to be more than just an enum. anyways, that's a long way of saying i think this is fine for now.

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
shiftkey: Want it to return `null` or render a successful file?

shiftkey: I made this return `null` for a binary diff in 651781d. LMK if you have any further input.

shiftkey:
:apple:

Issue #5603, opened 2018-09-11
Disable Compare to Branch menu item if modal is active


tierninho:
If the user is engaged with another modal, especially one that requires an action or acknowledgement, we are currently disabling most of our menu options. It seems one slipped by these rules. 

Branch > Compare to Branch should be disabled along with the rest. 

<img width="528" alt="screen shot 2018-09-11 at 9 13 51 am" src="https://user-images.githubusercontent.com/14828183/45373305-f20c7700-b5a3-11e8-9c63-0fc026579fbe.png">

tested in 1.4.0-beta3, mac

kanishk98:
@tierninho Hi. I'm just beginning to explore Desktop and this seems like a good place to start. Since this issue doesn't seem to be assigned to anyone yet, I'd like to take it up.

tierninho:
@kanishk98 thanks for taking this on and please reach out to the team if you have any questions

kanishk98:
So I haven't run any tests yet, but I suppose adding `compare-to-branch` in the `menu-update.ts` file did the job. 

![screen shot 2018-09-15 at 4 25 42 am](https://user-images.githubusercontent.com/24617297/45578506-ac9bc280-b89f-11e8-86f4-65879b8a8c6d.png)

@tierninho As I'm sure you'll notice, the dialog is transparent. I got the following warning in my terminal while running the dev build:
`GitHub Desktop-dev[9528:93939] *** WARNING: Textured window <AtomNSWindow: 0x7feabd55fb00> is getting an implicitly transparent titlebar. This will break when linking against newer SDKs. Use NSWindow's -titlebarAppearsTransparent=YES instead.`

Did a little bit of digging but couldn't find anything relevant. May I know what's happening here?

shiftkey:
@kanishk98 don't worry about that message, it's something that will be fixed as part of the upcoming Electron 3.0 release: https://github.com/electron/electron/issues/12814#issuecomment-421208525

kanishk98:
@shiftkey Sure, I'll ignore the message, no problem, but I'm troubled that the dialog is actually transparent. 
According to https://github.com/electron/electron/issues/11150#issue-274913410 this problem occurs when `titleBarStyle = hidden`, which does seem to be the case on Desktop. 
I think I'm missing something here.

shiftkey:
@kanishk98 that's not a new warning, and Desktop has been working fine despite it over many different builds of Electron which display that warning 🤷‍♂️

[...]

PR #6078, opened 2018-11-02
Disable Compare on GitHub menu item when no repository is selected

niik:
Seems like we've missed this menu item. It's currently enabled even when no repository is selected (i.e. when there are no repositories added to the app).


tierninho:
Adding the related issue so we can close it out: https://github.com/desktop/desktop/issues/5603

PR #6089, opened 2018-11-05
return commit id from commit and merge operations

### LINKED ISSUE #5923: lib/git/commit functions could return SHA on success
(From https://github.com/desktop/desktop/pull/5910#discussion_r225712202)

Currently, `createCommit` and `createMergeCommit` (both in https://github.com/desktop/desktop/blob/master/app/src/lib/git/commit.ts)  return a boolean or null on success and throw an error on failures. It would be nice to have them return the SHA of the commit created on success.

This will require parsing the git output from commit to get the SHA and adding some test coverage.

say25:
## Overview
Have `createCommit` and `createMergeCommit` return SHAs instead of boolean or null.

**Closes #5923**

## Description
- Have `createCommit` and `createMergeCommit` return SHAs instead of boolean or null.
- Add Unit Tests

## Release notes

Notes: No notes

### DIFF
--- a/app/test/unit/git/commit-test.ts
+++ b/app/test/unit/git/commit-test.ts
@@ -68,6 +68,7 @@ describe('git/commit', () => {
       const commits = await getCommits(repository!, 'HEAD', 100)
       expect(commits.length).toEqual(6)
       expect(commits[0].summary).toEqual('Special commit')
+      expect(commits[0].sha.substring(0, 7)).toEqual(sha)
     })
 
     it('commit does not strip commentary by default', async () => {
@@ -91,6 +92,7 @@ describe('git/commit', () => {
       expect(commit).not.toBeNull()
       expect(commit!.summary).toEqual('Special commit')
       expect(commit!.body).toEqual('# this is a comment\n')
+      expect(commit!.sha.substring(0, 7)).toEqual(sha)
     })
 
     it('can commit for empty repository', async () => {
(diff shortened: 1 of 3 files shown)


shiftkey:
> Do you think its worth doing a `git rev-parse HEAD` or something to verify that the new `HEAD` of the current branch is the same as the sha returned by the merge function? 

I'd prefer to avoid this because `git rev-parse HEAD` requires a second Git operation, and I'd prefer the app avoid any non-essential Git operations. If we think about this operation as a "you performed a merge and the merge commit is XYZ" we can avoid making this function more stateful and expensive to run.

👍 to baking a unit test to verify the HEAD after a merge is the one we expect though (or updating our existing tests)

app/test/unit/git/commit-test.ts
```diff
@@ -138,9 +141,10 @@ describe('git/commit', () => {
 
       expect(files.length).toEqual(1)
 
-      await createCommit(repo, 'renamed a file', [
+      const sha = await createCommit(repo, 'renamed a file', [
         files[0].withIncludeAll(true),
       ])
+      expect(sha).toHaveLength(7)
```
shiftkey: Just a warning on this magic number `7`: in a repository where there are lots of commits, the shortened commit ID here may be than 7 characters. For example, on the current `HEAD` of `desktop/desktop` I get a shortened SHA of 8 characters:

```shellsession
$ git rev-parse HEAD
0ce1ba23eff6da91ceedd2e01069c188faf85995
$ git rev-parse --short HEAD
0ce1ba23e
```

For newer repositories it's still likely to be 7 characters, and it's fine for tests to check for this length, but if we go and take this assumption into the application code we need to be aware of this feature of Git.

say25: 👍

say25:
@outofambit, I would agree with @shiftkey. Fresh commit coming very soon to add some tests.

outofambit:
> 👍 to baking a unit test to verify the HEAD after a merge is the one we expect though (or updating our existing tests)

that’s all i was suggesting: a unit test, not to add that verification to the method itself! :)

say25:
@outofambit I think the tests cover that case now

PR #6092, opened 2018-11-05
Upgrade Prettier to v1.15

j-f1:
## Overview

This upgrades Prettier to v1.15. The new version hasn’t been released yet, so I’ll update the PR when that happens.

## Description

N/A

## Release notes

Notes: no-notes


app/src/crash/crash-app.tsx
```diff
@@ -137,7 +137,7 @@ export class CrashApp extends React.Component<ICrashAppProps, ICrashAppState> {
         <p>
           GitHub Desktop encountered a catastrophic error that prevents it from
           launching. This has been reported to the team, but if you encounter
-          this repeatedly please report this issue to the GitHub Desktop{' '}
+          this repeatedly please report this issue to the GitHub Desktop{" "}
```
j-f1: prettier/prettier#5349

j-f1:
CI 💥 is from prettier/prettier#5336.

app/src/lib/shells/win32.ts
```diff
@@ -185,8 +185,8 @@ async function findHyper(): Promise<string | null> {
     const path = commandPieces
       ? commandPieces[2]
       : localAppData != null
-        ? localAppData.concat('\\hyper\\Hyper.exe')
-        : null // fall back to the launcher in install root
+      ? localAppData.concat('\\hyper\\Hyper.exe')
```
shiftkey: I think this is the only real formatting change that bothers me, and maybe this is a chance to avoid the crutch of nested ternary statements.

j-f1: This change was originally made to encourage the use of nested ternaries since they would previously indent for each level of nesting.

shiftkey: And I think I had kind of gotten used to that as a code smell, but I'm happy to defer to others how they feel about a) this change, and b) the use of nested ternaries in general.

j-f1:
A trick the new version supports: you can collapse an expanded object by just deleting the newline after the `{`.

j-f1:
Also, just realized I need to add the HTML files to Prettier.

app/static/error.html
```diff
@@ -21,16 +24,20 @@ <h1>GitHub Desktop failed to launch &#x1F4A5;</h1>
...
+        Would you mind opening an issue on the GitHub Desktop
+        <a
+          onclick="return onClick(event);"
+          href="https://github.com/desktop/desktop/issues"
+          >issue tracker</a
+        >? The details you need are included below:
       </p>
 
-      <!--{{content}}-->
+      <!-- {{content}} -->
```
shiftkey: I need to remember where/how we're setting this placeholder and confirm the introduced whitespace isn't going to break it

j-f1: Is this file even used anymore? It looks like the `crash` webpack config generates its own HTML.

j-f1: I also don’t recall seeing this wording when Desktop crashed in the past.

shiftkey: There's one way to find out 🤔

shiftkey: Aha, I knew this was something I was involved with: https://github.com/desktop/desktop/pull/1198#issuecomment-295160162

shiftkey: The component that rendered this was removed in c76110b34dec2adcd9d375114cfa802f1d312c01 so I'm going to remove it from this PR to avoid any further confusion.

app/src/lib/shells/win32.ts
shiftkey: @desktop/comrades @desktop/engineering I think this is the only bit of feedback I had about the PR, but I'm fine with merging as long as the code is still working (if a bit harder to follow)

j-f1:
🎉🎉🎉

I just have to upgrade to 1.15.2.

shiftkey:
@j-f1 is that something we should take in as part of this PR? We're still on `1.15.1` and it's not clear from [the changelog](https://github.com/prettier/prettier/blob/master/CHANGELOG.md#1152) whether it's worth also doing that here...

app/src/lib/shells/win32.ts
say25: Yea the more nesting kinda encourages me to make sub-functions personally. I think its also easier to read with nesting. Maybe we can turn that off/back?

shiftkey: I don't think we can, but @j-f1 can confirm. A suggestion if we still don't like this change - address these areas where we use nested ternaries in a `tech-debt` PR which we can then land and merge into this PR?

j-f1: > I don't think we can

Correct. There isn’t an option for this.

say25: Maybe we upstream an option? and pull down that version (maybe thats beyond the scope of this though)?

j-f1:
🍯

app/src/lib/shells/win32.ts
j-f1: @say25 Prettier has [a policy](https://prettier.io/docs/en/option-philosophy.html) of avoiding options.

PR #6095, opened 2018-11-05
first pass at a doc about upgrading dependencies

shiftkey:
## Overview

[**Rendered Markdown**](https://github.com/desktop/desktop/blob/upgrade-dependencies-doc/docs/contributing/upgrading-dependencies.md)

This document came out of a discussion I had with @say25 a couple of weeks ago about how the project manages dependencies.

## Description

For the moment I'm just focused on outlining how we upgrade dependencies. We did talk about automation services of these like [Dependabot](https://dependabot.com/) or [Greenkeeper](https://greenkeeper.io/) but I think the process should be clarified first.

## Release notes

Notes: no-notes

### DIFF
--- a/docs/contributing/upgrading-dependencies.md
+++ b/docs/contributing/upgrading-dependencies.md
@@ -1,0 +1,120 @@
+# Upgrading Dependencies
+
+This document outlines how we manage the dependencies that GitHub Desktop needs,
+as some need to be handled differently to others
+
+## The Impact Of A Dependency
+
+We group the dependencies in `package.json` and `app/package.json` into three
+buckets, based on their value to the project and what happens when a bad upgrade
+occurs.
+
+Because of the nature of the Node ecosystem, changes upgrading between versions
+of a package can introduce subtle changes that need to be reviewed and vetted.
+We also have finite engineering resources on the project, so we're a bit choosy
+with how and when we choose to upgrade dependencies, and some dependencies might
+be very stale at a time.
+
+The measured impact of a dependency can be identified by answering these three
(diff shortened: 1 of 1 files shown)


docs/contributing/upgrading-dependencies.md
```diff
@@ -0,0 +1,121 @@
...
+contribution is opened by the community.
+
+### Application dependencies
+
+This group of dependencies represent the rest of the listed dependencies in
+`app/pacakge.json` - there are lots of small dependencies which are approachable
+for external contributions.
+
+If you are interested in upgrading one of these, take a look at where it is
+used in the application and confirm the affected functionality works as
```
say25:
```suggestion
used in the application and confirm that the affected functionality works as
```

docs/contributing/upgrading-dependencies.md
```diff
@@ -0,0 +1,121 @@
...
+expected before opening a pull request.
```
say25: Is it work mentioning about the "test" releases here i.e. `1.4.1-test1`?

docs/contributing/upgrading-dependencies.md
say25: minor thing not really a big deal. I think it reads a little better ¯\_(ツ)_/¯

docs/contributing/upgrading-dependencies.md
shiftkey: I don't think so. I consider those test releases special snowflake builds that we might use to verify a specific fix (it's often easier to link an installer to someone than to get them to work with the source) but beyond that I consider them unofficial.

Issue #6085, opened 2018-11-03
Creating branch from detached HEAD branches from master


jeffdgr8:
## Description
While in a detached HEAD state, creating a new branch results in that branch being branched from master instead of the currently checked out commit.

## Version
* GitHub Desktop:
1.4.3
* Operating system:
Windows

## Steps to Reproduce
1. From terminal: `git checkout <some_commit_hash>`
2. In GitHub Desktop: Branch -> New branch...
3. Type a name.
4. Create branch.

### Expected Behavior
The newly created branch should point to `<some_commit_hash>`, as the dialog states `You do not currently have any branch checked out (your HEAD reference is detached). As such your new branch will be based on your currently checked out commit (<some_commit_hash>).`

### Actual Behavior
The newly created branch points to the HEAD of master.

## Additional Information
This happens on 1.4.2-linux1 from @shiftkey's [fork](https://github.com/shiftkey/desktop) as well.

### Logs
Not applicable

shiftkey:
@jeffdgr8 thanks for the report! I've been able to reproduce this on `master` and I think it comes down to two things:

 - `getStartPoint` seems to default to using the default branch (ignoring the fact we're in a `detached HEAD` state)

```ts
function getStartPoint(
  props: ICreateBranchProps,
  preferred: StartPoint
): StartPoint {
  // 
}
```

 - `createBranch` uses this `startPoint` state to revert to using the default branch name

```ts
let startPoint = undefined

if (this.state.startPoint === StartPoint.DefaultBranch) {
  // This really shouldn't happen, we take all kinds of precautions
  // to make sure the startPoint state is valid given the current props.
  if (!this.props.defaultBranch) {
    this.setState({
      currentError: new Error('Could not determine the default branch'),
    })
    return
  }

  startPoint = this.props.defaultBranch.name
}
```

There's no tests around `getStartPoint` so I want to wrap this current behaviour in tests before I fix the underlying issue.

mm-gmbd:
I was just about to post this exact same bug report but did one last search before submitting and found it already reported. It looks like it's already been reproduced, but I'll go ahead and include the details I had filled out anyway (as the system and GH Desktop versions are different):



## Description

Creating a new branch from a commit that is detached from HEAD works incorrectly. The dialog box says that it will create the branch from the current commit, but when I publish the branch to GitHub, it is clear that the branch is actually created from the default branch by looking at the commit history.


## Version

* GitHub Desktop: Version 1.5.0 (MacOS)

[...]

PR #6100, opened 2018-11-05
extract logic from inside Commit class

shiftkey:
## Overview

While investigating #6085  I needed to create a tree of objects as part of a test `Tip -> Branch -> Commit`, and I had to stub a function named `isWebFlowCommitter` to match the instance function already defined on `Commit`.

This felt wrong for a few reasons:

 - `isWebFlowCommiter` doesn't update any state on `Commit`
 - `isWebFlowCommitter` only needs to read public fields on `Commit`
 - I found a previous test that encountered this exact issue, and stubbed it with a `() => false` which just adds noise

This feels to me like a great situation where it can just be a function!

## Description

This PR moves `isWebFlowCommitter` out to `app/src/lib/commit-identity.ts` and updates the usage to call that function.

 - [x] rename file as commit identity is a different concept

## Release notes

Notes: no-notes


shiftkey:
Marking this as `time-sensitive` because the bugfix in #6104 depends on it currently

shiftkey:
> Could we find another place for the function to live? 

@niik 👍 I'll rename the file to avoid clashing with commit identity

shiftkey:
:apple:

PR #6082, opened 2018-11-02
acknowledge and handle binary files with conflicts differently in the UI

shiftkey:
## Overview 

**Closes #6066**

This PR adds a simple UI to acknowledge binary files with conflicts, rather than considering them resolved.

## Description

 - [x] test that binary files are marked as conflicted
 - [x] retest with #6067 merged so we can confirm the dialog isn't closed after resolving on command line
 - [x] merge conflict detection needs to look for `.git/MERGE_HEAD` rather than working directory state
 - [x] investigate how changes have introduced the "list of files loaded after showing popup, leading to the popup being drawn off center" issue again
 - [ ] ensure the UI does show a "all conflicts resolved" if there is nothing left to do
 - [ ] investigate why counted file does not show an entry in list

## More information
<details><summary>Collapsed away to focus on the feature itself, expand me</summary>

There's some imperfections with the flow currently that I'm thinking about, and might be resolved with other work that's in-progress.

I have two options here based on the command line - "ours" or "theirs":

```
On branch make-a-change
You have unmerged paths.
  (fix conflicts and run "git commit")
  (use "git merge --abort" to abort the merge)

Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)

	deleted by us:   my-cool-image.png
```

1. If I run `git add my-cool-image.png` in the shell to add the file ("theirs"), this is what Git reports.

```
On branch make-a-change
All conflicts fixed but you are still merging.
  (use "git commit" to conclude merge)

Changes to be committed:

	new file:   my-cool-image.png

```

When I switch back to the app, the conflict dialog disappears and it looks like I'm about to make a regular commit.

This should be fixed when #6067 is resolved, and I imagine the app would show an empty list of conflicts and let the user commit the merge.

2. If I run `git rm my-cool-image.png` in the shell to remove the file ("ours"), this is what Git reports:

```
On branch make-a-change
All conflicts fixed but you are still merging.
  (use "git commit" to conclude merge)
```

And when I switch to the app, I don't have the ability to commit to complete the merge (because no changes):

Again, when we resolve #6067 the app should put them back in the conflicts UI and get them to  complete the merge, but we don't necessarily have any conflicts to display as resolved.

@ampinsk did any of your mocks around the conflicts dialog have a "no outstanding conflicts" view? I think users who resolve conflicts externally and stage the changes from the command line will lose context like this, and it'd be nice to just have a "you're good to merge" state for the dialog.

</details>

## Release notes

Notes: no-notes

PR #6116, opened 2018-11-07
use getStatus to detect conflicted repository state

shiftkey:
## Overview

It's late, I can't sleep, and I'm restless from watching the news so I figure now is a great time to open a PR that changes how we detect merge conflicts in the app :tada:

This was started in #6082 but I wanted to move it out from there for more explicit discussion as it touches the core parts of conflict resolution flow.

## Description

Currently we look at the output of `git status` to see which files are conflicted, but if a user decides to stage files outside the app (whether they're resolved or not) the conflict status markers are forgotten by `git status`, so it's possible for Desktop to "forget" it's in the middle of a conflict.

This affects #6082 but also other scenarios that weren't the happy path flow, like the user addressing conflicts in the app but not completing the process by committing before switching back to Desktop.

Instead of relying on `git status` output, this PR switches us over to look for `.git/MERGE_HEAD` in the repository to identify when conflicts need to be resolved.

The flow of this information is then like this:

 - `getStatus` has a new field `mergeHeadFound` that indicates whether it found `MERGE_HEAD`
 - within `AppStore._loadStatus` we compute the new conflicted state for the repository (`getConflictStatus()`)
 - we update the repository state with the conflict information, as well as related stats ~~(still needs some cleanup)~~
 - inside `AppStore._triggerMergeConflictsFlow` we then check the current state of conflicts
 - we use this detail to display the merge conflicts dialog (if required)

This also removes some `_mergeConflictDetected()` which would set the conflicted state, but didn't use `git status` - this centralizes the responsibility to `AppStore._loadStatus()`

TODO:

 - [x] refactor the stats flow to reflect the desired behaviour
 - [x] a bunch more testing

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1589,7 +1589,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return { conflictState: newConflictStatus }
     })
 
-    this._triggerMergeConflictsFlow(repository, status)
+    this._triggerMergeConflictsFlow(repository)
 
     this.emitUpdate()
 
@@ -1599,19 +1599,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /** starts the conflict resolution flow, if appropriate */
-  private async _triggerMergeConflictsFlow(
-    repository: Repository,
-    status: IStatusResult
-  ) {
+  private async _triggerMergeConflictsFlow(repository: Repository) {
     if (!enableMergeConflictsDialog()) {
-      return
-    }
-
-    if (!status.mergeHeadFound) {
-      return
-    }
-
-    if (status.currentBranch === undefined) {
       return
     }
 
@@ -1620,6 +1609,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
       (this.currentPopup.type === PopupType.MergeConflicts ||
         this.currentPopup.type === PopupType.AbortMerge)
     if (alreadyInFlow) {
+      return
+    }
+
+    const repoState = this.repositoryStateCache.get(repository)
+    const { conflictState } = repoState.changesState
+    if (conflictState === null) {
       return
     }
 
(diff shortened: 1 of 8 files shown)


billygriffin:
Cool! Couple questions:

1. Just clarifying: I assume based on the description that this is just relating to detecting when a conflict has already occurred, and therefore doesn’t help solve the mismatch between git merge and git merge-tree we’ve seen pop up occasionally in predicting conflicts.

2. Will this impact our metrics on conflicts resolved successfully vs. abandoned at all?

shiftkey:
@billygriffin 

> 1. Just clarifying: I assume based on the description that this is just relating to detecting when a conflict has already occurred, and therefore doesn’t help solve the mismatch between git merge and git merge-tree we’ve seen pop up occasionally in predicting conflicts.

Yep, this is separate to the "merge hint" feature.

> 2. Will this impact our metrics on conflicts resolved successfully vs. abandoned at all?

I intend to keep that support around - based on the rules previously defined - and the code path gets hit whenever we update the status of the repository (like before) so there shouldn't be any change to that area.

shiftkey:
I'd like to get reviewed in time for the next beta, to give it the maximum time to bake.

The PR that started this off #6082 won't be resolved without these changes, and I plan to open a fresh PR after this lands targeting the 1.5.0 release branch to ensure it gets in.

app/src/lib/stores/app-store.ts
```diff
@@ -1609,7 +1546,51 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return { workingDirectory, selectedFileIDs, diff }
     })
 
-    this._triggerMergeConflictsFlow(repository, status)
+    this.repositoryStateCache.updateChangesState(repository, state => {
```
outofambit: can we pull this into its own method?

shiftkey: Mind if I do this in a follow-up PR? There 's an existing [function above](https://github.com/desktop/desktop/blob/6d0206acc9f4ea543266acf948cf5064260f96e2/app/src/lib/stores/app-store.ts#L1488-L1546) that is also rather lengthy, and I'd love to tackle both of those at the same time.

outofambit:
going to merge this in ~1hr so we can unblock #6122

outofambit:
@shiftkey it looks like we have some conflicts now. would you mind resolving those? i'll merge this after that.

shiftkey:
@outofambit done 🍎

PR #6119, opened 2018-11-07
Return the SHA in merge

### LINKED ISSUE #6110: git merge API should return commit id
following on from #6089 and #5923, i think we should have our git merge api also return the resulting commit id, especially since `createMergeCommit` does.

https://github.com/desktop/desktop/blob/5ff3f0b318b99633e649574a646cd53daf9ca792/app/src/lib/git/merge.ts#L9-L12

should return `Promise<string>` instead, which will require some changes to `app-store` and/or `git-store` to handle properly.

you can (probably) use https://github.com/desktop/desktop/blob/19ced1ae19c16858b4303c2a6f9a03a077995f6c/app/src/lib/git/commit.ts#L69 to do the parsing.

say25:
## Overview
**Closes #6110**

## Description

- Return the SHA in Merge command

## Release notes

Notes: No Notes

### DIFF
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -1,4 +1,4 @@
-import { git, GitError, IGitResult } from './core'
+import { git, GitError, parseCommitSHA } from './core'
 import { stageFiles } from './update-index'
 import { Repository } from '../../models/repository'
 import { WorkingDirectoryFileChange } from '../../models/status'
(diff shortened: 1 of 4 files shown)


shiftkey:
@say25 oops, looks like there's a merge conflict

say25:
@shiftkey merged

say25:
Not sure if the creation of the release branch means this can be merged or not 🤑

outofambit:
> Not sure if the creation of the release branch means this can be merged or not 🤑

it can!! and it will be (soon)!

shiftkey:
@say25 yep, and to add to what @outofambit said - this wasn't a user-facing change so we decided to pull it in after we'd gotten the release for next week ready (we lose Monday due to holidays, so we want to ensure we're stable and relaxed now).

PR #6133, opened 2018-11-08
Detect and prompt users to use command line for text file merge conflicts without markers

### LINKED ISSUE #6123: text files which differ by added/remove are not rendered correctly
In the vein of #6062, there are other scenarios where you can have conflicted files without text markers.

For this repro, I:

 - created a blank repository
 - committed a file named `README.md` which had some content
 - created a branch named `removed-readme` 
 - switched  to `master` and committed some changes to `README.md`

```diff
commit 62cf8e12edf1a1d3bf03cd826a7194ae485db4f9 (HEAD -> master)
Author: Brendan Forster <github@brendanforster.com>
Date:   Wed Nov 7 10:59:00 2018 -0400

    adding more words

diff --git a/README.md b/README.md
index c166706..ea2225d 100644
--- a/README.md
+++ b/README.md
@@ -1 +1,5 @@
 # writing some words
+
+## writing some more words
+
+### oh my yes
```

 - switched to `removed-readme` and committed a change to remove `README.md`

```diff
commit 211f948a0758b41b4efbde630f18645884112d52 (remove-readme)
Author: Brendan Forster <github@brendanforster.com>
Date:   Wed Nov 7 10:59:17 2018 -0400

    removed the README

diff --git a/README.md b/README.md
deleted file mode 100644
index c166706..0000000
--- a/README.md
+++ /dev/null
@@ -1 +0,0 @@
-# writing some words
```

 - switched to Desktop and merged `master` into `remove-readme`

I encountered the conflict view as expected, but the details don't match the options on the command line:

The porcelain details indicate the `DU` conflict marker (deleted in ours, updated in theirs) needs a different resolution to viewing conflicts than viewing in the editor:

```
$ git status --porcelain=2 --branch
# branch.oid 211f948a0758b41b4efbde630f18645884112d52
# branch.head remove-readme
u DU N... 100644 000000 100644 100644 c166706f733ce814b8e54198c2c088740f6c21e4 0000000000000000000000000000000000000000 ea2225de6d102de272bd276a7b364aea07070629 README.md
```

### LINKED ISSUE #6132: text files which differ by added/remove should be resolved via command line
similar to #6066, i think we can do a first iteration on #6123, where we display a similar UI that says "Use command line to resolve this file"

ui reference

outofambit:
## Overview

Closes #6123 

## Description

screen shot 2018-11-08 at 10 50 56 am

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -258,13 +258,13 @@ function getConflictStatus(
   }
 
   if (status.kind === 'conflicted') {
-    const uncoveredConflict =
+    const conflictWithoutMarkers =
       status.them !== GitStatusEntry.UpdatedButUnmerged &&
       status.us !== GitStatusEntry.UpdatedButUnmerged &&
       status.them !== GitStatusEntry.Modified &&
       status.us !== GitStatusEntry.Modified
 
-    if (uncoveredConflict) {
+    if (conflictWithoutMarkers) {
       const { us, them } = status
       return { kind: 'text', conflictMarkerCount: null, us, them }
     }
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -258,14 +258,14 @@ function getConflictStatus(
   }
 
   if (status.kind === 'conflicted') {
+    const { us, them } = status
     const conflictWithoutMarkers =
-      status.them !== GitStatusEntry.UpdatedButUnmerged &&
-      status.us !== GitStatusEntry.UpdatedButUnmerged &&
-      status.them !== GitStatusEntry.Modified &&
-      status.us !== GitStatusEntry.Modified
+      them !== GitStatusEntry.UpdatedButUnmerged &&
+      us !== GitStatusEntry.UpdatedButUnmerged &&
+      them !== GitStatusEntry.Modified &&
+      us !== GitStatusEntry.Modified
 
     if (conflictWithoutMarkers) {
-      const { us, them } = status
       return { kind: 'text', conflictMarkerCount: null, us, them }
     }
   }
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -236,6 +236,7 @@ export async function getStatus(
 
 function getConflictStatus(
   path: string,
+  status: FileEntry,
   conflictState: ConflictState
 ): ConflictFileStatus | null {
   const { filesWithConflictMarkers, binaryFilePathsInConflicts } = conflictState
@@ -254,6 +255,15 @@ function getConflictStatus(
 
   if (conflictMarkerCount != null) {
     return { kind: 'text', conflictMarkerCount }
+  }
+
+  const uncoveredConflict =
+    status.kind === 'conflicted' &&
+    status.them !== GitStatusEntry.UpdatedButUnmerged &&
+    status.us !== GitStatusEntry.UpdatedButUnmerged
+
+  if (uncoveredConflict) {
+    return { kind: 'text', conflictMarkerCount: 0 }
   }
 
   return null
@@ -293,7 +303,7 @@ function buildStatusMap(
     files.delete(entry.path)
   }
 
-  const conflictStatus = getConflictStatus(entry.path, conflictState)
+  const conflictStatus = getConflictStatus(entry.path, status, conflictState)
 
   // for now we just poke at the existing summary
   const summary = convertToAppStatus(status, conflictStatus !== null)
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -257,13 +257,17 @@ function getConflictStatus(
     return { kind: 'text', conflictMarkerCount }
   }
 
-  const uncoveredConflict =
-    status.kind === 'conflicted' &&
-    status.them !== GitStatusEntry.UpdatedButUnmerged &&
-    status.us !== GitStatusEntry.UpdatedButUnmerged
+  if (status.kind === 'conflicted') {
+    const uncoveredConflict =
+      status.them !== GitStatusEntry.UpdatedButUnmerged &&
+      status.us !== GitStatusEntry.UpdatedButUnmerged &&
+      status.them !== GitStatusEntry.Modified &&
+      status.us !== GitStatusEntry.Modified
 
-  if (uncoveredConflict) {
-    return { kind: 'text', conflictMarkerCount: null }
+    if (uncoveredConflict) {
+      const { us, them } = status
+      return { kind: 'text', conflictMarkerCount: null, us, them }
+    }
   }
 
   return null
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -263,7 +263,7 @@ function getConflictStatus(
     status.us !== GitStatusEntry.UpdatedButUnmerged
 
   if (uncoveredConflict) {
-    return { kind: 'text', conflictMarkerCount: 0 }
+    return { kind: 'text', conflictMarkerCount: null }
   }
 
   return null
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -259,11 +259,11 @@ function getConflictStatus(
 
   if (status.kind === 'conflicted') {
     const { us, them } = status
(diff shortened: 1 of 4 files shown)


app/src/lib/git/status.ts
```diff
@@ -256,6 +257,15 @@ function getConflictStatus(
     return { kind: 'text', conflictMarkerCount }
   }
 
+  const uncoveredConflict =
```
shiftkey: :art: I'm trying to think of a more descriptive name for this parameter that fits with how it's used downstream but nothing really jumps out - `conflictRequiresManualResolution` feels too wordy, `conflictWithoutMarkers` maybe? How attached are you to `uncovered` here?

app/src/models/conflicts.ts
```diff
@@ -10,8 +10,10 @@ import { GitStatusEntry, UnmergedEntry } from './status'
 export type ConflictFileStatus =
   | {
       readonly kind: 'text'
-      /** This number should be greater than zero */
-      readonly conflictMarkerCount: number
+      /** This number should be greater than zero, or null
```
shiftkey: :art: We usually start the text of a multi-line comment after the first line to give it some space to breathe around the code

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -180,6 +180,22 @@ export class MergeConflictsDialog extends React.Component<
   ): JSX.Element | null {
     switch (conflictStatus.kind) {
       case 'text':
+        if (conflictStatus.conflictMarkerCount === null) {
+          return (
+            <li className="unmerged-file-status-conflicts">
```
shiftkey: :art: we've now repeated the same `<li>` and `<Octicon>` elements three times in here - any interest in lifting this up into `renderUnmergedFile` to avoid accidentally forgetting to change a `className` in here? e.g.

```
const content = this.renderConflictedFile(
  file.path,
  file.conflictStatus,
  () =>
    this.props.openFileInExternalEditor(
      join(this.props.repository.path, file.path)
    )
)

return (
  <li className="unmerged-file-status-conflicts">
    <Octicon symbol={OcticonSymbol.fileCode} className="file-octicon" />
    {content}
  </li>
)
```

app/src/lib/git/status.ts
outofambit: not attached at all! i've been trying to think of a better name, too. i like `conflictsWithoutMarkers`.

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
outofambit: totally into this, was thinking the same thing :)

outofambit:
@ampinsk, is this ok as an minimum minimum minimum first step to addressing this case?

ampinsk:
Yep! Makes sense to me as a start 👍

app/src/models/status.ts
```diff
@@ -8,27 +8,27 @@ import { ConflictFileStatus } from './conflicts'
...
+  Copied = 'C',
   // .
-  Unchanged,
+  Unchanged = '.',
   // ?
-  Untracked,
+  Untracked = '?',
   // !
-  Ignored,
+  Ignored = '!',
```
shiftkey: :art: now that these are string enums the code comments are unnecessary - feel free to tidy them up while we're in here

app/src/lib/git/status.ts
```diff
@@ -256,6 +257,19 @@ function getConflictStatus(
     return { kind: 'text', conflictMarkerCount }
   }
 
+  if (status.kind === 'conflicted') {
+    const { us, them } = status
+    const conflictWithoutMarkers =
+      them !== GitStatusEntry.UpdatedButUnmerged &&
+      us !== GitStatusEntry.UpdatedButUnmerged &&
+      them !== GitStatusEntry.Modified &&
```
shiftkey: I don't think this check is quite accurate, because it'll also catch situations where one branch has modified the file, but the other has not. 

My original repro passes fine if we make the check something like this:

```diff
diff --git a/app/src/lib/git/status.ts b/app/src/lib/git/status.ts
index 43615b893..2f5d842d5 100644
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -259,11 +259,14 @@ function getConflictStatus(
 
   if (status.kind === 'conflicted') {
     const { us, them } = status
+
+    const bothFilesModified =
+      them === GitStatusEntry.Modified && us === GitStatusEntry.Modified
+
     const conflictWithoutMarkers =
       them !== GitStatusEntry.UpdatedButUnmerged &&
       us !== GitStatusEntry.UpdatedButUnmerged &&
-      them !== GitStatusEntry.Modified &&
-      us !== GitStatusEntry.Modified
+      !bothFilesModified
 
     if (conflictWithoutMarkers) {
       return { kind: 'text', conflictMarkerCount: null, us, them }
```

And with this check I see text files that we previously considered resolved show have conflicts that they shouldn't be handling in an editor:

screen shot 2018-11-09 at 10 29 56 am

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -178,8 +177,21 @@ export class MergeConflictsDialog extends React.Component<
...
     onOpenEditorClick: () => void
   ): JSX.Element | null {
-    switch (conflictStatus.kind) {
-      case 'text':
+    let content = null
+    if (conflictStatus.kind === 'text') {
+      if (conflictStatus.conflictMarkerCount === null) {
+        content = (
+          <li className="unmerged-file-status-conflicts">
+            <Octicon symbol={OcticonSymbol.fileCode} className="file-octicon" />
```
shiftkey: With the patch to change how this path is detected, I see that we didn't cleanup the duplicated DOM here and we see two icons:

screen shot 2018-11-09 at 10 24 11 am

app/src/lib/git/status.ts
shiftkey: Actually, I like this change to the logic of how we detect them to also call out "both added" which also looks to put the content on-disk with merge conflicts (if any):

```
diff --git a/app/src/lib/git/status.ts b/app/src/lib/git/status.ts
index 43615b893..5222dd49f 100644
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -259,11 +259,18 @@ function getConflictStatus(
 
   if (status.kind === 'conflicted') {
     const { us, them } = status
+
+    const bothFilesModified =
+      them === GitStatusEntry.Modified && us === GitStatusEntry.Modified
+
+    const bothFilesAdded =
+      them === GitStatusEntry.Added && us === GitStatusEntry.Added
+
     const conflictWithoutMarkers =
       them !== GitStatusEntry.UpdatedButUnmerged &&
       us !== GitStatusEntry.UpdatedButUnmerged &&
-      them !== GitStatusEntry.Modified &&
-      us !== GitStatusEntry.Modified
+      !bothFilesModified &&
+      !bothFilesAdded
 
     if (conflictWithoutMarkers) {
       return { kind: 'text', conflictMarkerCount: null, us, them }
```

tbuteler:
Could this behaviour be achieved without a dialog? I feel this solution is incredibly intrusive. There is no option to dismiss the dialog without either committing or aborting, but what if I just want to review the changes? It makes the entire app unusable while there's a conflicted merge pending in one of your repositories. Surely we can't assume all users want this, right?

Also, there is no option to customise the commit message.

tierninho:
@tbuteler Please follow along in #6213 for the next iteration on how we're handling merge conflicts. 

> Also, there is no option to customise the commit message.

cc @billygriffin ☝️

shiftkey:
> Also, there is no option to customise the commit message.

@tbuteler can you please move this out to a [new issue](https://github.com/desktop/desktop/issues/new?template=problem-to-raise.md) so we can discuss this further - I'd like to understand more about this use case

tbuteler:
@shiftkey This has been addressed already, so no need to open a new issue. Just being able to close the conflict modal and using the "old" workflow means I can commit with whatever message I choose. Thanks everyone for iterating so quickly on this, latest version is great!

PR #6141, opened 2018-11-09
prevent concurrent fetches between user and status indicator checks

### LINKED ISSUE #5328: `unable to update local ref` error on first Pull for every repo
## Description

When I open GitHub Desktop I start by pulling down several repos (all hosted on GitHub). Since update 1.3.2 I now consistently get an error window pop-up with the following content (obviously the repo and delta count varies each time, but the end result is always the same)

```
remote: Counting objects: 117, done.        
remote: Compressing objects:   3% (1/26)           
[snipped for brevity]
remote: Compressing objects: 100% (26/26), done.        
Receiving objects:   0% (1/117)   
Receiving objects:   1% (2/117)   
Receiving objects:  45% (53/117)   
remote: Total 117 (delta 98), reused 110 (delta 91), pack-reused 0        
Receiving objects:  46% (54/117)   
[snipped for brevity]
Receiving objects: 100% (117/117), 22.94 KiB | 7.65 MiB/s, done.
Resolving deltas:   0% (0/98)   
[snipped for brevity]
Resolving deltas:  94% (93/98)   
Resolving deltas: 100% (98/98), completed with 30 local objects.
error: cannot lock ref 'refs/remotes/origin/master': is at fc0dc1393039da99c24b165a868444429a591cf6 but expected 0967bfcab4d5f8671783696c67065431fb5d35ba
From https://github.com/photonstorm/phaser
 ! 0967bfcab..fc0dc1393  master     -> origin/master  (unable to update local ref)
```

This only happens once, the first time I try to pull. If I dismiss the error, then click Pull again, it works.

## Version

* GitHub Desktop: 1.3.2

## Steps to Reproduce

Launch GitHub Desktop. Click 'Pull' on any repo.

### Expected Behavior

For it to pull down without an error window.

### Actual Behavior

See above.

## Additional Information

It's not too much of a problem because I can just click Pull a second time and it'll work fine. It happens whenever I launch GitHub Desktop. So if I close it down, make a commit from the command line, open Desktop and try to pull, the first one will error.

It has only started happening with 1.3.2 or the previous version, but definitely not before.

### LINKED ISSUE #5438: error: cannot lock ref 'refs/remotes/origin/

### LINKED ISSUE #6121: Error: Cannot lock ref 'refs/remotes/origin/master'
I have some problem with my GitHub. I read similar questions and answers to them in many forums, but unfortunately nothing helped.

**error: cannot lock ref 'refs/remotes/origin/master': unable to resolve reference
'refs/remotes/origin/master': reference broken
From https://github.com/arbuzzer/arbuzzer.github.io
! [new branch] master -> origin/master (unable to update local ref)**

shiftkey:
## Overview

This PR does some tidy up around how our background fetching from #4770 integrates with other operations and hopefully resolves these reported issues: 

Fixes #6121
Fixes #5438
Fixes #5328

## Description

When we initially rolled out #4770 the goal was to be able to fetch in the background without the user knowing (or needing to care about errors). That's why we originally went through `GitStore` to perform the fetch - but this avoided `withPushPull` which checks and prevents concurrent push/pull/fetch operations.

This seemed fine for the most part, but I believe this has left us open to a race condition with concurrent fetches, where the user starts a fetch at the same time as the repository indicator updater is fetching the same repository, or vice versa.

I stumbled across https://github.com/Microsoft/vscode/issues/47141 earlier today where VSCode basically outlined this behaviour between itself and an extension, and this comment comes to the same conclusion about what I believe is happening to these users:

> This issue occurs when git-radar is doing a fetch in background and user runs git fetch/pull. Git has index.lock to lock index during a fetch, so you cannot corrupt the index by doing concurrent fetches. Your git command fails because of this lock prevents git modifying index.
>
> https://github.com/michaeldfallen/git-radar/issues/81#issuecomment-345191136

It's hard to see from the reported issues which Git operation was the one that failed, but I believe it can be tripped with just a fetch (`pull` also performs a `fetch` to update it's refs, I'm not sure about `push` from memory).

What can we do here? I think the simplest answer is to push these status indicator updates through our existing infrastructure - these use repository state to control whether or not an operation should be performed.

The worst case scenarios I can think of about this change:

 - user tries to fetch/push/pull while the status indicator update is in progress - the button will now show details about the status indicator update (it did not before)
 - status indicator update tries to run while user is fetch/push/pull - status indicator update is skipped

## Release notes

Notes: Prevent concurrent fetches corrupting Git index by adding checking to repository indicator updates


app/src/lib/stores/app-store.ts
```diff
@@ -1937,15 +1937,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this._updateCurrentPullRequest(repository)
     this.updateMenuItemLabels(repository)
     this._initializeCompare(repository)
-    this.refreshIndicatorsForRepositories([repository])
+    this.refreshIndicatorsForRepositories([repository], false)
```
shiftkey: Without doing this check we can end up in recursion hell due to this loop when updating the repository indicators: `_fetch -> performFetch -> _refreshRepository -> refreshIndicatorsForRepositories -> refreshIndicatorForRepository -> _fetch -> 🔁`

billygriffin:
This seems like a step in the right direction here.

> status indicator update tries to run while user is fetch/push/pull - status indicator update is skipped

This seems like expected behavior as a user. My hierarchy of things I care about is definitely whatever is happening in Push/Pull first, much more so than the background ops of repo indicators - it's fine for them to update during the next background operation.

user tries to fetch/push/pull while the status indicator update is in progress - the button will now show details about the status indicator update (it did not before)

I'm not positive I understand what you mean by this. What will the button show differently? Does it interfere with the fetch/push/pull op?

More generally, I think you're aligned with this, but since this has been an issue since 1.3, I think it's ok to release in 1.6 or 1.5.1 from a prioritization perspective.

shiftkey:
@billygriffin yeah, this isn't something I planned to pull in for 1.5.0 - I just wanted to open this PR early so I can work through the impact (race conditions are _great_)!

shiftkey:
> I'm not positive I understand what you mean by this. What will the button show differently? Does it interfere with the fetch/push/pull op?

@billygriffin our normal fetch/push/pull flow will take the progress reported by Git and display it in the toolbar button. We do this for two reasons:

 - it's nice to show users that progress, especially when it takes a while to complete
 - if a fetch/push/pull is in progress, we disable the button so the user doesn't start a second operation

It's that second point that I believe is relevant here, because of how the app can arbitrarily run a fetch in the background (with no feedback to the user). If we don't check for an in-flight fetch/push/pull operation in the background, and the fetch can take a while due to slow network or complex repositories, the likelihood of concurrent fetches increases significant.

shiftkey:
Tentatively scheduling this for `1.6.1` because I'd like it to sit on `beta` for some time to verify the underlying issues are resolved.

shiftkey:
Marking this one as `time-sensitive` as I'd like to get a first round of review and see how we feel about getting this into `1.6.1` because of the `priority-2` issues associated with it.

outofambit:
anyone else from @desktop/engineering want to weigh in?

shiftkey:
@tierninho it's kind of tricky because the background fetch doesn't really have any UI, and you'll see this message _after_ it's completed:

https://github.com/desktop/desktop/blob/ba999a3b28a531ad128c55583c68736beb3d0d02/app/src/lib/stores/app-store.ts#L1969-L1973

This also forces the background fetcher through `AppStore.shouldBackgroundFetch`, which means it's more likely to hit the 30min minimum interval for background fetching.

https://github.com/desktop/desktop/blob/ba999a3b28a531ad128c55583c68736beb3d0d02/app/src/lib/stores/app-store.ts#L1286-L1293

Maybe just trying to fetch regularly in the app to see if you can force the background fetcher to trip at the same time as your manual fetch?

shiftkey:
Merging this in so that it's available as part of the first beta

sanjaygupta1011:
You need to update the reference using following Git command on Git bash:
$ git update-ref -d refs/remotes/origin/[locked branch name]
then pull using $git pull

[locked branch name] is the name of the branch that the error is happening because of mismatch of commit Ids.

PR #6142, opened 2018-11-09
add experimental support for v2 of the Git wire protocol

shiftkey:
## Overview

GitHub just enabled support for v2 of the wire protocol on it's repositories, which helps clients and servers save on network traffic: https://blog.github.com/changelog/2018-11-08-git-protocol-v2-support/

## Description

This PR enables Desktop to opt-in for the new behaviour and start testing it out ourselves (currently development mode only).

This checks it is working with a GitHub repository, and for now won't consider GitHub Enterprise repositories (because they need to receive this with a future update, and we don't have a good way to detect this support) and non-GitHub repositories.

There is a way to enable this via config yourself if you're on Git 2.18 or later:

```shellsession
$ git config --global protocol.version 2
```

Alternatively you can use `--local` instead of `--global` to enable it for a specific repository.

More details about wire protocol changes itself can be found here: https://opensource.googleblog.com/2018/05/introducing-git-protocol-version-2.html

TODO:

 - [x] add config check to respect if `protocol.version` is set to `1` to indicate they have set themselves to the old version.

## Release notes

Notes: Opt-in for v2 of the Git Wire protocol for GitHub repositories

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -51,3 +51,8 @@ export function enableRecurseSubmodulesFlag(): boolean {
 export function enableMergeConflictsDialog(): boolean {
   return true
 }
+
+/** Should the app set protocol.version=2 for any fetch/push/pull/clone operation? */
+export function enableGitProtocolVersionTwo(): boolean {
+  return enableDevelopmentFeatures()
+}
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -78,8 +78,10 @@ export async function deleteBranch(
   )
 
   if (branchExistsOnRemote) {
+    const networkArguments = gitNetworkArguments(account)
+
     const args = [
-      ...gitNetworkArguments,
+      ...networkArguments,
       'push',
       remote,
       `:${branch.nameWithoutRemote}`,
@@ -101,8 +103,10 @@ async function checkIfBranchExistsOnRemote(
   account: IGitAccount | null,
   remote: string
 ): Promise<boolean> {
+  const networkArguments = gitNetworkArguments(account)
+
   const args = [
-    ...gitNetworkArguments,
+    ...networkArguments,
     'ls-remote',
     '--heads',
     remote,
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -78,7 +78,7 @@ export async function deleteBranch(
   )
 
   if (branchExistsOnRemote) {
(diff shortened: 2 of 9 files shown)


shiftkey:
@clarkbw we talked a bit about this yesterday, and I skimmed [the old post](https://opensource.googleblog.com/2018/05/introducing-git-protocol-version-2.html) about Wire Protocol v2 that touches on backwards-compatibility - will remotes which don't support v2 ignore this and work fine, like was mentioned in that post?

> Now, in every case, a client can issue a request to use v2, using a transport-specific side channel, and v2 servers can respond using the new protocol while older servers will ignore the side channel and just respond with a ref advertisement. 

I can differentiate between remotes easily enough in the app, but if it's safe to enable for everyone knowing that servers that don't know about v2 will ignore it (and not error) that will simplify this PR further.

cc @peff who knows a thing or two about Git 😀

IntelOSt:
Its tekhnologi my hp

IntelOSt:
This my divice hp

peff:
@shiftkey Yes, it _should_ just auto-negotiate with older implementations, and the trickery to send the initial "hey, I speak v2. Do you?" message has been tested to be ignored with older versions of Git. It's possible that a custom server-side Git implementation would fail, but we (upstream) checked against the major ones (GitHub, JGit, Atlassian) at the time and didn't find any problems.

So it should be pretty safe to turn on. Git older than v2.18 will barf on that config, but since you ship your own internal copy of Git, that should be fine. The biggest risk is that there may be bugs in the client side of v2 that are still shaking out (e.g., my recent git/git@631f0f8c4b3cf86df13534be40b8df56c2fc6dda which is on master but not yet released). I don't know offhand of any that are more than minor annoyances, though.

shiftkey:
@peff thanks for the extra info!

> Git older than v2.18 will barf on that config, but since you ship your own internal copy of Git, that should be fine.

We're going to enable that per-operation and feature flag it to gradually roll it out, rather than mess with user's config.

> The biggest risk is that there may be bugs in the client side of v2 that are still shaking out (e.g., my recent git/git@631f0f8 which is on master but not yet released).

Yep, I'm happy for this to make it in a beta release of Desktop as a first step to ensuring we don't encounter anything unexpected in the wild.

shiftkey:
Marking this as `ready-for-review`.

Someone else is welcome to address any feedback if they want to see this merged before I'm back from holidays.

iAmWillShepherd:
Went ahead and merged since this is behind a feature flag.

Issue #5876, opened 2018-10-11
Desktop toolchain does not support Node 10 currently


shiftkey:
We've had some reports of contributors being unable to build Desktop in Node 10, which has mostly been down to native node modules which weren't compatible with Node 10.

Some recent reports of users who encountered this: #5603, #5654, #5672

Note that this is a separate issue to how we build the Electron itself - we use [`app/.npmrc`](https://github.com/desktop/desktop/blob/master/app/.npmrc) to control how the native modules the app needs are built.

I believe https://github.com/desktop/desktop/pull/5644 is the last piece of the puzzle, but it's blocked on requiring clarification from another team about whether it's the right approach.

shiftkey:
Friendly bump on this because Node 10 was just transitioned to LTS, meaning Node 8 support is going to be phased out - https://nodejs.org/en/blog/release/v10.13.0/

outofambit:
for reference https://github.com/desktop/desktop/issues/5933#issuecomment-438130606

shiftkey:
Tentatively pulling this into 1.6.1 because I have a working branch, but it needs some tooling to catch regressions. I plan to open a few PRs in the coming days in this order:

 - first, extract `legal-eagle` from the main `build.ts` into it's own module and wrap it in a `Promise`, so it can be chained better (top-level `await` isn't currently possible, as far as I'm aware)
 - next, introduce a regression test for any emitted CSS in the app bundle - if it contains a `$` this represents a SassScript value that wasn't encoded properly, so the app will run but _something_ won't look right
 - with all that in place, upgrade node-sass, catch all the new regressions, ensure it passes CI and test the app out
 - test out enabling CI to use Node 10 rather than Node 8. if that works, refresh the docs to simplify things

shiftkey:
With #6661 just approved I'm going to move this to `1.7.0` to address the final stages of this PR:

 - review setup guides to drop mention of this constraint on Node 8
 - upgrade CI to Node 11 
 - PR in a change to documentation and build configs
 - confirm all builds work

shiftkey:
I'm going to close this out as the underlying issue should be resolved for anyone using Node 10/11. 

Based on the feedback when reviewing #6661 I've captured the remaining work to explore Node 10/11 support in the project in these tasks: #6964 #6965

PR #6152, opened 2018-11-09
Bump for last node 8

shiftkey:
## Overview

Appveyor has the latest version of Node 8 installed on its agents, so we currently waste time on CI builds uninstalling that to choose the previous version specified in config, just for running scripts:

```
Install-Product node $env:nodejs_version $env:platform
Uninstalling node 8.12.0 (x86)...
Installing node 8.11.4 (x64)...
```

This is one of our bottlenecks for getting feedback on builds, so hopefully this saves us some time.

## Description

This touches all the places where we mention `8.11.x`, so it's a good reference for what we need to update when we want to move to Node 10 in #5876 

TODO:

 - [x] Azure Pipelines Windows is reverting to Python 3 (Python2 is available, just not on PATH)
 - [x] fix comment in docs

## Release notes

Notes: no-notes


outofambit:
oops sorry, i got excited!

docs/contributing/setup-macos.md
```diff
@@ -16,7 +16,7 @@ $ node -v
 
 If you see an error about being unable to find `node`, that probably means you don't have any Node tools installed. You can install Node LTS (the version we need) from the [Node.js website](https://nodejs.org/en/download/) and restart your shell.
 
-If you see the output `v8.11.x` (where `x` is any number), you're good to go.
+If you see the output `v8.12.0` (where `x` is any number), you're good to go.
```
j-f1: Should “(where `x` is any number)” be removed?

outofambit:
exciting!! i think we also need to change https://github.com/desktop/desktop/blob/952a9b15f0af7f331998a7e415f165a39ca77c0b/package.json#L52

shiftkey:
@outofambit I've got mixed feelings on that, but mostly err on the side of "do we need to force everyone to update?" This original issue came from trying to improve CI scenarios, and I'm fine with not forcing contributors to upgrade too (this is just for tooling).

outofambit:
@shiftkey 👍 i don't feel strongly about this. in the future though i think we should leave off the y version in `package.json` if we don't want to update it along with everything else

shiftkey:
> i think we should leave off the y version in package.json if we don't want to update it along with everything else

@outofambit yep, I'll keep this in mind whenever we get around to #5876, but I had to add this in because of a specific issue with a package: https://github.com/desktop/desktop/pull/4943#discussion_r195795606

Issue #5933, opened 2018-10-18
open in external editor does not handle account names with whitespace


edgale:
## Description

I basically have the same problem as #4246. I switched users and reinstalled github desktop. After that it worked. I noticed that in my old account the user had a non ascii character in the name (ö). So maybe this has something to do with non ascii character paths.

## Version

* GitHub Desktop:Version 1.4.2

* Operating system: Microsoft Windows [Version 10.0.17134.345]

shiftkey:
@edgale thanks for the report! Last time I checked this I thought we were fine with account names with non-ASCII characters, but maybe something has changed recently?

Could you try running through the steps, open `View` | `Toggle Developer Tools`, and enable Verbose logging? We have some extra diagnostics around this that might highlight the error:

<img width="946" src="https://user-images.githubusercontent.com/359239/47159257-d5124480-d2c3-11e8-806e-bc95308a7817.png">

edgale:
Hi @shiftkey, thanks for the reply.
I deleted the user but could recreate the issue with creating a user with the same username. I also forgot to mention that the username has a space in it.
Here are the logs:
```
[AppStore] loading 1 repositories from store
install.ts:31 [AppStore] found account: edgale (Maximilian Röhrle)
2install.ts:35 Executing getStatus: git status --untracked-files=all --branch --porcelain=2 -z
app.tsx:215 App ready time: 29.096399009227753ms
app.tsx:216 Load time: 700.2170879989862ms
app.tsx:217 Renderer ready time: 183.385ms
2install.ts:35 Executing getCommits: git log 182c6e9fbca67805c5d59a8aa9f756d31a9ec27e --date=raw --max-count=1 --pretty=%H%x1F%s%x1F%b%x1F%an <%ae> %ad%x1F%cn <%ce> %cd%x1F%P%x1F%(trailers:unfold,only) -z --no-show-signature --no-color --
install.ts:35 Executing installGlobalLFSFilter: git lfs install --skip-repo
install.ts:31 launching: 1.4.3 (Windows 10.0.17134)
install.ts:31 execPath: 'C:\Users\Maximilian Röhrle\AppData\Local\GitHubDesktop\app-1.4.3\GitHubDesktop.exe'
2install.ts:35 Executing getConfigValueInPath: git config -z trailer.separators
[...]

Issue #5655, opened 2018-09-15
unable to open editors when user account has space in it


LordOfTheThunder:
**details about your environment - which build, which operating system**
Git Desktop Version 1.4.0
Operating System - Windows 10 Education
Visual Studio Code Version 1.27.2

**details about reproducing the issue - what steps to take, what happens, how often it happens**
I installed Visual Studio Code but the "Open in Visual Studio Code" option doesn't react and nothing happens. When I open the folder manually through Visual Studio Code everything works well.

shen5d:
哦？搜噶~

shiftkey:
@LordOfTheThunder thanks for the report!

Could you attach the full log files? The **Help** | **Show Logs in Explorer** menu items should open the logs folder, and inside those are the log files generated by day. This will provide some more information to help troubleshoot.

LordOfTheThunder:
@shiftkey I've made a little investigation and found out that the reason for this problem is, again, because the path of both my sandbox and my visual studio code are has a folder that contains a space.
The issue happens inside launch.ts inside the proc launchExternalEditor
This hardcoded fix makes it run:

```typescript
export async function launchExternalEditor(
  fullPath: string,
  editor: FoundEditor
): Promise<void> {
  var editorPath = editor.path

  const exists = await pathExists(editorPath)
  if (!exists) {
    const label = __DARWIN__ ? 'Preferences' : 'Options'
    throw new ExternalEditorError(
      `Could not find executable for '${editor.editor}' at path '${
        editor.path
      }'.  Please open ${label} and select an available editor.`,
      { openPreferences: true }
    )
  }
  //spawn(editorPath, [fullPath])
  spawn('C:\\Users\\Max Kolchinsky\\AppData\\Local\\atom\\atom.exe', [
    'C:\\Users\\Max Kolchinsky\\Documents\\max_sbox\\desktop',
  ])
}
```

Unfortunately the editorPath contains atom.cmd which is inside bin dir instaed of atom.exe which I used here. I tried multiple options to manipulate both editorPath and fullPath but everything failed.

shiftkey:
@LordOfTheThunder I think we need to encode this argument, like this:

```diff
diff --git a/app/src/lib/editors/launch.ts b/app/src/lib/editors/launch.ts
index 14fcc4b0e..f77e81a9c 100644
--- a/app/src/lib/editors/launch.ts
+++ b/app/src/lib/editors/launch.ts
[...]

Issue #4277, opened 2018-03-18
Open in Visual Studio Code never works for me.


deep5050:
Neither the CTRL+SHIFT+A nor clicking "open in visual studio code" work for me.
![vs](https://user-images.githubusercontent.com/27947066/37567864-04d574c2-2af3-11e8-92ae-deffcd22a9e8.gif)

shiftkey:
What version of VSCode do you have installed? And which build of GitHub Desktop and Windows?

I'm also looking at #4246 which might be related to this issue.

shiftkey:
Actually, I'm going to close this out as a duplicate of whatever is happening in #4246, please follow along with that issue.

Issue #4246, opened 2018-03-15
Launching the external editor is not working on windows 10.


Yash-Pal:
## Description

**I just installed the github desktop on my windows 10 PC. I have installed both editor "Atom" and "Microsoft Visual Studio Code" and selected one is VS Code. Now when I tried to open repository with the editor with shortcut Ctrl+Shift+A it doesn't do anything. I have changed my default editor to "Atom" but still not working.**

## Version

* GitHub Desktop: 1.1.0

* Operating system: Windows 10

### Expected Behavior
Selected Editor should be open when I press **Ctrl+Shift+A** or by opening it from **Repository -> Open in <Selected Editor>**

### Actual Behavior
Nothing happening.

shiftkey:
@Yash-Pal thanks for the report!

Could you attach the full log files? The **Help** | **Show Logs in Explorer** menu items should open the logs folder, and inside those are the log files generated by day. This will provide some more information to help troubleshoot.

shiftkey:
> Now when I tried to open repository with the editor with shortcut Ctrl+Shift+A it doesn't do anything.

Does manually launching the external editor with the menu item work?

Yash-Pal:
Manually launching the external editor with the menu item is not working. And where do I find the log file?

Thanks!

shiftkey:
@Yash-Pal apologies, I had a look over [the implementation](https://github.com/desktop/desktop/blob/release-1.1.0/app/src/lib/editors/win32.ts) of this and it's logging things at the `debug` level, which doesn't end up in these logs.

Could you open the **View** | **Toggle Developer Tools** menu item, run through the steps to trigger the issue, and check the **Console** for any interesting messages that might explain what's failing?

Yash-Pal:
I toggled the Developer tool and tried to open editor with available methods manually and with shortcut too. But console doesn't show any errors warnings etc. And one more thing I want to tell you that I am using windows 10 64 bit operating system.
Device Configuration Info:
![device specification](https://user-images.githubusercontent.com/19922707/37500174-fea8eeb4-28ed-11e8-9e7f-d181b2983d74.png)


Developer Tools screenshot: 
![console](https://user-images.githubusercontent.com/19922707/37500109-a521a0ac-28ed-11e8-8b3e-1939e6820a6a.png)

Thanks!

shiftkey:
@Yash-Pal oops, could you toggle the verbosity from "Info" to "Verbose"? (yay defaults)

[...]

PR #6181, opened 2018-11-13
Win32 LaunchEditor use Shell for .CMD files

msftrncs:
## Overview

Addresses issues #5933, #5655, #4246, #4277, all regarding the commands for opening the repository, or a file, in an external editor, and that editor choice requiring use of a shell script to be spawned and there being spaces in the path to the editor, often in the user account in the case of Atom.

## Description

The listed issues stem from issues in Node JS `spawn` function which doesn't always handle execution of shell scripts correctly.  If both the file path to the shell script file, and an argument being passed to the script include spaces, the spawn does not work, but without any error.

I have added the `usesShell` property to `IFoundEditor` to indicate when the editor uses a shell script and needs to be spawned accordingly.  `getAvailableEditors` arbitrarily adds `usesShell: true` for the editors known to use a shell script when spawned.   `launchExternalEditor` is then altered to detect the `usesShell` property and use a different set of arguments for the spawn, to include the spawn option `shell: true`, along with quotation of the arguments.

## Release notes

 - fixes 'Open In External Editor' for Atom/VS Code on Windows when paths contain spaces.

Notes:

FYI, this is as much work as I have ever done in TypeScript or even JavaScript for that matter.  Please excuse me if I have failed to follow any normal conventions.  It could be possible that additional commenting could be utilized to help insure future additions to the editors list will result in the added editors working correctly.

### DIFF
--- a/app/src/lib/editors/found-editor.ts
+++ b/app/src/lib/editors/found-editor.ts
@@ -1,4 +1,5 @@
 export interface IFoundEditor<T> {
   readonly editor: T
   readonly path: string
+  readonly usesShell?: boolean
 }
--- a/app/src/lib/editors/launch.ts
+++ b/app/src/lib/editors/launch.ts
@@ -18,11 +18,14 @@ export async function launchExternalEditor(
     const label = __DARWIN__ ? 'Preferences' : 'Options'
     throw new ExternalEditorError(
       `Could not find executable for '${editor.editor}' at path '${
-        editor.path
+      editor.path
       }'.  Please open ${label} and select an available editor.`,
       { openPreferences: true }
     )
   }
-
-  spawn(editorPath, [fullPath])
+  if (editor.usesShell) {
+    spawn(`"${editorPath}"`, [`"${fullPath}"`], { shell: true })
+  } else {
+    spawn(editorPath, [fullPath])
+  }
 }
--- a/app/src/lib/editors/launch.ts
+++ b/app/src/lib/editors/launch.ts
@@ -18,7 +18,7 @@ export async function launchExternalEditor(
     const label = __DARWIN__ ? 'Preferences' : 'Options'
     throw new ExternalEditorError(
       `Could not find executable for '${editor.editor}' at path '${
-      editor.path
+        editor.path
       }'.  Please open ${label} and select an available editor.`,
       { openPreferences: true }
     )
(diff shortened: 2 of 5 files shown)


billygriffin:
Thanks @msftrncs for the PR! We're working on getting 1.5 ready to go this week, so the GitHub folks likely won't be able to review this for at least a few days, but I wanted to let you know we really appreciate you digging into this! ❤️

outofambit:
@msftrncs thanks so much for this work! 💟 what are you thoughts about using https://www.npmjs.com/package/cross-spawn? it seems like that might be more resilient since we don't have to explicitly pass in the `shell` option ourselves, but perhaps i'm missing other considerations. let me know what you think!

msftrncs:
@outofambit, I've spent some time to review the code at https://github.com/moxystudio/node-cross-spawn, but have not been able to test it, probably due to my lack of experience with Node/TypeScript/JavaScript (etc...).  I am unable to see how cross-spawn will improve the situation, and instead I am afraid it would make matters worse.  There is a fair bit of support code happening to provide escapes that I don't see a need for.  The only thing that I find cross-spawn doing that spawn might not (I cannot prove it does or doesn't), is automatic detection of using 'shell: true', but it does not resolve the issue of spaces in the file path of the command, or in the passed argument, as it does not wrap both in double quotes as it builds the shell command argument (it wraps the entire command in double quotes, just like spawn does when using 'Shell: true').

This PR provides a relatively simple and minimal solution to the problem (and this problem only).  It does however require the acknowledgement of the issue on 1 platform under particular circumstances, such that if further work is done in the future (adding additional Windows editors that utilize batchscript shims) that care must be taken to note that circumstance.

Please be aware, there is an issue in VS Code that still causes problems if the path to the repository or file GitHub Desktop passes contains a special character, such as '%' or '^'. Atom doesn't seem to exhibit this issue.  See https://github.com/Microsoft/vscode/issues/63348.

outofambit:
@msftrncs thanks for your thoughts! that all makes sense

outofambit:
@tierninho could you verify this fix works for you? let me know if you have any questions.

tierninho:
Tested this fix based issues found in #5933, #5655, #4246, #4277:

- [x] Opened multiple repos with multiple editors, using command, menu and right click
- [x] Opened multiple files (uncommitted, uncommitted) with multiple editors, via right click
- [x] Launched editor(s) in directory that contained spaces
- [x] Launched editor(s) with user's name that contained spaces
- [x] Switching editors is reflected in the UI
- [x] Adding new editor mid-stream is recognized 

tested in v1.5.1-test0, Windows 10 v17763.1, 51ba5315e06ce68d47fcdae250d271f13973d40a

PR #6187, opened 2018-11-13
indicate that we expect this error to occur to avoid it polluting the dev tools

shiftkey:
## Overview

While investigating a potential bug we were getting feedback about this error in the dev tools:

This PR tidies up that warning, because it's a known case and wasn't related to the root cause of the issue.

## Description

We have an `expectedErrors` argument that can be provided when invoking Git to indicate what known errors we expect. This was initially put in place to identify the known ways an operation might fail, and make clear cases that Git encounters which we should handle.

This error is expected and pushed into the error handling pipeline, so we can't set it as expected to kill off the noise in the dev tools.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/merge.ts
+++ b/app/src/lib/git/merge.ts
@@ -2,6 +2,7 @@ import * as FSE from 'fs-extra'
 import * as Path from 'path'
 
 import { git } from './core'
+import { GitError } from 'dugite'
 import { Repository } from '../../models/repository'
 import { Branch } from '../../models/branch'
 import { MergeResult, MergeResultKind } from '../../models/merge'
@@ -13,7 +14,9 @@ export async function merge(
   repository: Repository,
   branch: string
 ): Promise<true> {
-  await git(['merge', branch], repository.path, 'merge')
+  await git(['merge', branch], repository.path, 'merge', {
+    expectedErrors: new Set([GitError.MergeConflicts]),
+  })
   return true
 }
 
--- a/app/src/lib/git/merge.ts
+++ b/app/src/lib/git/merge.ts
@@ -1,7 +1,7 @@
 import * as FSE from 'fs-extra'
 import * as Path from 'path'
 
-import { git, parseCommitSHA } from './core'
+import { git } from './core'
 import { GitError } from 'dugite'
 import { Repository } from '../../models/repository'
 import { Branch } from '../../models/branch'
@@ -13,11 +13,11 @@ import { spawnAndComplete } from './spawn'
 export async function merge(
   repository: Repository,
   branch: string
-): Promise<string> {
+): Promise<boolean> {
   const result = await git(['merge', branch], repository.path, 'merge', {
     expectedErrors: new Set([GitError.MergeConflicts]),
   })
-  return parseCommitSHA(result)
+  return result.exitCode === 0
 }
 
 /**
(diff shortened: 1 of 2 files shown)


outofambit:
all for this cleanup but this PR currently causes a regression in the merge banner behavior:

unwanted merge banner

this probably happens at https://github.com/desktop/desktop/blob/c7dae96cc9143d9df6e505ed122fc9c1d3b0280a/app/src/lib/stores/app-store.ts#L3127

shiftkey:
@outofambit nice spot! I'll see what needs to change downstream of the call...

shiftkey:
@outofambit the issue you uncovered here is related to #6119 -  we shouldn't be trying to `parseCommitSHA` the merge result here because `git merge` doesn't emit any details about the newly-created commit.

Here's an example from just now of the output of a successful merge:

```shellsession
$ git merge master
Merge made by the 'recursive' strategy.
 .node-version                                      |   2 +-
 .nvmrc                                             |   2 +-
...
 docs/contributing/setup-macos.md                   |  10 +-
 docs/contributing/setup.md                         |   4 +-
 docs/technical/oauth.md                            |   2 +-
 docs/technical/syntax-highlighting.md              |   2 +-
 35 files changed, 278 insertions(+), 196 deletions(-)
 create mode 100644 app/src/lib/web-flow-committer.ts
```

And here's a merge that fails due to conflicts:

```
Auto-merging vsts.yml
CONFLICT (content): Merge conflict in vsts.yml
CONFLICT (modify/delete): vendor/node deleted in HEAD and modified in 3-0-x. Version 3-0-x of vendor/node left in tree.
CONFLICT (modify/delete): vendor/libchromiumcontent deleted in HEAD and modified in 3-0-x. Version 3-0-x of vendor/libchromiumcontent left in tree.
Auto-merging tools/run-if-exists.js
CONFLICT (add/add): Merge conflict in tools/run-if-exists.js
Auto-merging spec/webview-spec.js
CONFLICT (content): Merge conflict in spec/webview-spec.js
Auto-merging spec/static/main.js
...
Automatic merge failed; fix conflicts and then commit the result.
```

After I merged #6119 into this branch I poked at the flow again and noticed we were getting a random string when the merge resulted in a conflict (which would be considered truthy by the check afterwards):



I'm going to update `merge` to return `Promise<boolean>` but I think I now need to look at the exit code to determine whether to show `true` or `false` (something we wanted to do in here earlier anyway!)

shiftkey:
:apple:

outofambit:
wow, excellent find @shiftkey. sorry about that. i'll throw some tests on that in another PR soon.

shiftkey:
@outofambit no problem, and the impact of this didn't make it into a release because of how we cut the `1.5.0` release branch :tada:

outofambit:
@shiftkey yeah, that's a really good point. 💜 without a release branch this would have hit production. release branches ftw!

shiftkey:
Also, for whoever ends up writing tests for this I have some suggestions based on this investigation:

 - I'd change `parseCommitSHA` to take a `stdout` string directly, rather than `IGitResult` (makes it easier to test)
 - `parseCommitSHA` should also return `null` if it can't find a valid commit SHA (in case it somehow gets given some output it can't handle)
 - beef up the logic with a regex to look for that `[{branch} {sha}]` header

shiftkey:
Going to merge this in because @outofambit is enjoying a well-deserved day off

PR #6189, opened 2018-11-13
add a unified commit context shape to combine related data

shiftkey:
## Overview

This came up in a review of #6013 as it replicated the pattern elsewhere in the codebase of defining the same three parameters to represent a commit message: https://github.com/desktop/desktop/pull/6013#discussion_r228990995

## Description

To generate a commit message, we need to flow these fields around to eventually reach `formatCommitMessage`:

```ts
const message = await formatCommitMessage(
  repository,
  summary,
  description,
  trailers
)
```

Instead, this PR combines these three fields into a single `ICommitContext` to clean up this replication:

```ts
const message = await formatCommitMessage(repository, context)
```

Another benefit of this PR is that it tidies up places where we had to import `ITrailer` because we'd have to annotate the parameters to each function:

```ts
import { ITrailer } from '../git/interpret-trailers'
```

And we'd have to ensure these parameters were all consistent:

```ts
  public async commitIncludedChanges(
    repository: Repository,
    summary: string,
    description: string | null,
  ): Promise<boolean> {
```

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -22,7 +22,7 @@ import {
 import { BranchesTab } from '../../models/branches-tab'
 import { CloneRepositoryTab } from '../../models/clone-repository-tab'
 import { CloningRepository } from '../../models/cloning-repository'
-import { Commit } from '../../models/commit'
+import { Commit, ICommitContext } from '../../models/commit'
 import {
   DiffSelection,
   DiffSelectionType,
@@ -122,7 +122,6 @@ import {
   getWorkingDirectoryDiff,
   isCoAuthoredByTrailer,
   mergeTree,
-  ITrailer,
   pull as pullRepo,
   push as pushRepo,
   renameBranch,
@@ -1775,9 +1774,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   /** This shouldn't be called directly. See `Dispatcher`. */
   public async _commitIncludedChanges(
     repository: Repository,
-    summary: string,
-    description: string | null,
-    trailers?: ReadonlyArray<ITrailer>
+    context: ICommitContext
   ): Promise<boolean> {
     const state = this.repositoryStateCache.get(repository)
     const files = state.changesState.workingDirectory.files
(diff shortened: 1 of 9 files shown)


shiftkey:
@Daniel-McCarthy I'm happy to land #6013 before this and deal with merge conflicts on my end

app/src/models/commit.ts
```diff
@@ -4,6 +4,22 @@ import { GitAuthor } from './git-author'
...
+  /**
+   * The summary of the commit message (required)
+   */
+  readonly summary: string
+  /**
+   * Additional details for the commit message (optional)
+   */
+  readonly description: string | null
+  /**
+   * Trailers representing co-author information (optional)
```
niik: Technically speaking these trailers could be anything that conforms to the trailer format, not just co-author trailers. Could we say something like

```suggestion
   * An optional array of commit trailers (for example Co-Authored-By trailers) which will be appended to the commit message in accordance with the Git trailer configuration.
```

app/src/lib/stores/app-store.ts
```diff
@@ -1809,6 +1801,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         this.statsStore.recordPartialCommit()
       }
 
+      const { trailers } = context
       if (trailers != null && trailers.some(isCoAuthoredByTrailer)) {
```
niik: I know this isn't part of your change (in fact I think I wrote this) but can we make this `!== undefined` to be clear about the fact that trailers can't be `null`, only `undefined` or an array?
```suggestion
      if (trailers !== undefined && trailers.some(isCoAuthoredByTrailer)) {
```

Daniel-McCarthy:
> I'm happy to land #6013 before this and deal with merge conflicts on my end

@shiftkey That would certainly work. I had expected that this would likely land first and I could pull in these changes and handle merging it into #6013 as I had already done for some previous changes.

Either order is perfectly fine by me. Thanks for that offer as well; I had not considered that.

niik:
@shiftkey Damn, I just realized that this didn't have the ready-for-review label on it. Sorry for the overeager review. Feel free to ignore my comments until it's ready for review

shiftkey:
@niik no problem, and thanks for the feedback - it was mostly done anyway

Issue #6093, opened 2018-11-05
consolidate conflict information in ConflictFileStatus model


outofambit:
from https://github.com/desktop/desktop/pull/6063#pullrequestreview-171370147

> I wonder if we can collapse the `ConflictFileStatus`, `AppFileStatus.Conflicted` and `AppFileStatus.Resolved` into a single field. Right now a `ConflictFileStatus` of `null` means no conflicts, but we don't know if there _were_ ever conflicts without cross-referencing `AppFileStatus`. Perhaps we could remove `AppFileStatus.Resolved` and have a third (and maybe fourth) type for `ConflictFileStatus` that effectively means "resolved", so that a `ConflictState` of `null` would coincide with unconflicted files and there would be a single field to check for conflict information.

shiftkey:
Tentatively assigning myself to this based on #6205 but there might be further work around this after that is reviewed...

Issue #6062, opened 2018-10-31
display manual conflict resolution options in app


ampinsk:
Right now, we're treating binary files as simply resolved:

<img width="531" alt="screen shot 2018-10-31 at 10 44 31 am" src="https://user-images.githubusercontent.com/10404068/47807837-473e4c80-dcfa-11e8-96c3-1f07257ada70.png">

It would probably be more clear to treat them differently to let the user know that they may need to resolve these outside of Desktop, or even better, we could resolve these using an "ours/theirs" sort of approach in the dialog itself 🤔 

I think we can probably address this after 1.5.0 but wanted to track this as a case 👍

shiftkey:
> let the user know that they may need to resolve these outside of Desktop,

Yep, looks like we might need to tweak our detection of conflicts to handle these files.

> we could resolve these using an "ours/theirs" sort of approach in the dialog itself 🤔

I'm into this idea for the sake of simplicity, and here's a quick writeup on how to resolve this on the command line (which we can steal for any eventual solution): https://stackoverflow.com/a/2163926

```
git checkout --theirs -- path/to/conflicted-file.txt
# or
git checkout --ours -- path/to/conflicted-file.txt
```

shiftkey:
Marking this as `priority-2` because we incorrectly consider them as resolved, and I'm pretty sure the merge won't actually complete successfully if the user tries it.

ampinsk:
I just tried it, and it looks like we do let you successfully complete a merge in this case, at least in the UI

shiftkey:
@ampinsk I guess we just stage whatever is currently in the working directory (which could be it's own can of worms). Gonna try and put together a repro of this to test on my end and confirm. 🤷‍♂️

outofambit:
#6066 is the first iteration of this, while this issue will contain further refinement.

shiftkey:
I started noodling on the UI changes for this building on top of #6066 and wanted to throw it open for feedback. With binary files the only typical options you have when merging conflicts is to choose your version of the file, or the one coming from the other branch.

I started on a dumb UI to expose these options (no messing with styles just yet):

<img width="590" src="https://user-images.githubusercontent.com/359239/47872772-3838c680-ddee-11e8-8351-b459eeb186cd.png">

Unlike merging conflicts in text files, we shouldn't immediately update the Git index with the choice you've made - because the file may disappear when you choose "ours". Instead, we should track the option in-memory and apply it before performing the merge.

[...]

PR #6205, opened 2018-11-14
convert AppFileStatus into a first-class shape

shiftkey:
## Overview

In #6093 we called out that there's some potential confusion with how we carry around `ConflictFileStatus`.  As part of tackling #6062 I need to track some more details, so I spent a bit of time trying to improve the situation before I started adding more details .

This PR transforms `AppFileStatus` into a proper object (with `AppFileStatusKind` as a string enum) and adds the ability to attach additional state to the object, based on the status found.

## Description

In a nutshell, this is now what we have for `AppFileStatus`:

```ts
export type PlainFileStatus = {
  kind:
    | AppFileStatusKind.New
    | AppFileStatusKind.Modified
    | AppFileStatusKind.Deleted
    // TODO: where should this live?
    | AppFileStatusKind.Resolved
}

 export type CopiedOrRenamedFileStatus = {
  kind: AppFileStatusKind.Copied | AppFileStatusKind.Renamed
  oldPath: string
}

 export type ConflictedFileStatus = {
  kind: AppFileStatusKind.Conflicted
  conflictStatus: ConflictFileStatus
}

 export type AppFileStatus =
  | PlainFileStatus
  | CopiedOrRenamedFileStatus
  | ConflictedFileStatus
```

## TODO

 - [x] refactor `AppFileStatus` and update usage/tests
 - [x] move conflict state into `ConflictedFileStatus`
 - [x] migrate impacted tests to jest assertions
 - [x] can we deprecate `FileChange.oldPath` and `WorkingDirectoryFileChange.oldPath` in favour of the one in `AppFileStatus`?

## Release notes

Notes: no-notes


app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -238,17 +238,12 @@ export class MergeConflictsDialog extends React.Component<
...
   ): JSX.Element | null {
-    switch (file.status) {
-      case AppFileStatus.Resolved:
+    const { status } = file
+    switch (status.kind) {
+      case AppFileStatusKind.Resolved:
         return this.renderResolvedFile(file.path)
-      case AppFileStatus.Conflicted:
-        if (file.conflictStatus === null) {
-          throw new Error(
```
shiftkey: This was the main benefit of the PR - being able to verify at compile time that we're in a known good state

shiftkey:
Moving this to `time-sensitive` because I have more internals refactorings that I'd like to open to address #6093 and #6062 which depend on these changes.

app/src/models/status.ts
```diff
@@ -33,6 +33,31 @@ export enum AppFileStatus {
...
 }
 
+/** The state of a GitHub Desktop-specific change containing additional metadata */
+export type PlainFileStatus = {
+  kind:
+    | AppFileStatusKind.New
+    | AppFileStatusKind.Modified
+    | AppFileStatusKind.Deleted
+    // TODO: where should this live?
+    | AppFileStatusKind.Resolved
```
outofambit: what are you thoughts here, @shiftkey 
?

i have mixed feelings about putting this under `PlainFileStatus`. these are all files that can exist in a conflicted merge but require no action. but, they all can appear in a regular status, except for resolved.

what if we have two overlapping type unions? one for regular statuses and one for files in conflict that require no action?

```ts
| AppFileStatusKind.New
| AppFileStatusKind.Modified
| AppFileStatusKind.Deleted
```

and 

```ts
| AppFileStatusKind.New
| AppFileStatusKind.Modified
| AppFileStatusKind.Deleted
| AppFileStatusKind.Resolved
```

(or something to that effect)

app/src/models/status.ts
```diff
@@ -138,14 +163,19 @@ export class FileChange {
...
-    public readonly status: AppFileStatus,
-    public readonly oldPath?: string
+    public readonly status: AppFileStatus
   ) {
-    this.id = `${this.status}+${this.path}`
+    if (
+      status.kind === AppFileStatusKind.Renamed ||
+      status.kind === AppFileStatusKind.Copied
+    ) {
+      this.id = `${this.status.kind}+${this.path}+${status.oldPath}`
```
outofambit: what's the significance of this change? is this a bugfix?

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
outofambit: 💯

app/src/models/status.ts
shiftkey: I left it there as a comment to unblock this PR, because I knew I'd need to think about this later.

Rather than go too deep here I'd suggest we figure it out in #6214 what feels more natural.

app/src/models/status.ts
shiftkey: This was another case that was improved by `oldPath` being moved into `AppFileStatus` - we can generate a better id for these files.

app/src/models/status.ts
outofambit: if this is going into `master` and its got a `todo` comment on it and we're not sure where it should live, i'd prefer to have it be its own type in the interim. non-blocking.

shiftkey: @outofambit let me see what I can do here without introducing too much unnecessary churn

shiftkey: bedd01ace

shiftkey:
:apple:

shiftkey:
Taking this in so I can get the PRs it depends on ready for review

PR #6214, opened 2018-11-16
unify conflicted state into a more meaningful model

### LINKED ISSUE #6093: consolidate conflict information in ConflictFileStatus model

shiftkey:
## Overview

This PR clarifies our models in the app to better support these three scenarios:

 - when a file should be resolved by the user, open their preferred editor
 - when a file has no more conflict markers, mark the file as resolved in the app
 - when a file cannot be resolved in an editor, let the user choose the option

For 1.5.0 we settled on deferring to the command line, but with #6062 in scope for 1.5.1 I went back to the drawing board about how we model these two states

**Closes #6093**

## Description

The inspiration from this model came from the Git SCM docs for `git status`, which list this set of values for the conflicted state: https://git-scm.com/docs/git-status#_short_format

```
-------------------------------------------------
D           D    unmerged, both deleted
A           U    unmerged, added by us
U           D    unmerged, deleted by them
U           A    unmerged, added by them
D           U    unmerged, deleted by us
A           A    unmerged, both added
U           U    unmerged, both modified
-------------------------------------------------
```

I believe the last two entries are the only ones where we should be looking for conflict markers, and this lead to updating the type system to flow these details up to the app, avoid duplicate work, and simplify the work needed in the UI.

### `app/src/lib/status-parser.ts`

As the parser of raw `git status`, I needed to update this to return more details as part of the  `UnmergedEntry` type. For example, this represents the two scenarios where we should look for conflict markers:

```ts
type TextConflictEntry = {
  readonly kind: 'conflicted'
} & TextConflictDetails

type TextConflictDetails =
  | {
      readonly action: UnmergedEntrySummary.BothAdded
      readonly us: GitStatusEntry.Added
      readonly them: GitStatusEntry.Added
    }
  | {
      readonly action: UnmergedEntrySummary.BothModified
      readonly us: GitStatusEntry.UpdatedButUnmerged
      readonly them: GitStatusEntry.UpdatedButUnmerged
    }
```

`ManualConflictDetails` is the type representing the other entries sourced from `git status` docs, and the union of these types becomes our new `UnmergedEntry` type:

```ts
type ManualConflictEntry = {
  readonly kind: 'conflicted'
} & ManualConflictDetails

export type UnmergedEntry = TextConflictEntry | ManualConflictEntry
```

I also ensured we were returning `U` (updated but unmerged) instead of `M`(modified) for these files, which was a workaround we did from back before we had a story about conflict resolution in the app.

### `app/src/models/status.ts`

Because have some richer details in `UnmergedEntry`, we can now make our `AppFileStatus` type richer with the details about conflicts.

Previously we had this type:

```ts
export type ConflictedFileStatus = {
  kind: AppFileStatusKind.Conflicted
  conflictStatus: ConflictFileStatus
}
```

`ConflictFileStatus` had some limitations:

 - it replicated some of the underlying Git details, but also had to have some optional fields
 - it differentiated between text diffs and binary diffs - I don't think we need to do this any more, and I'll explain more about why later on

Now that we have a richer `UnmergedEntry`, we can define a richer status for the application models using union types again:

```ts
export type ConflictedFileStatus = ConflictsWithMarkers | ManualConflict

type ConflictsWithMarkers = {
  kind: AppFileStatusKind.Conflicted
  entry: TextConflictEntry
  lookForConflictMarkers: true
  conflictMarkerCount: number
}

type ManualConflict = {
  kind: AppFileStatusKind.Conflicted
  entry: ManualConflictEntry
  lookForConflictMarkers: false
}
```

`lookForConflictMarkers` is a convenience field for the UI when looking for `conflictMarkerCount`, without having to revert to looking for undefined fields. I might rename this to something more descriptive - suggestions welcome.

As we can rely more on the index state to track details about conflicts, I believe we can eliminate the `binary-file.ts` check that we were doing inline when the repository is in a conflicted state. I'm going to do a bit more digging into how our `git diff --check` work interops with the index to confirm.

### UI Components

I think with these types we can cover the three important scenarios, without needing a `AppFileStatusKind.Resolved` status:

 - conflicted files that can be resolved in an editor look like this

```ts
f =>
  isConflictedFile(f.status) &&
  f.status.lookForConflictMarkers &&
  f.status.conflictMarkerCount > 0
```

 - conflicted files with nothing more to do look like this:

```ts
f =>
  isConflictedFile(f.status) &&
  f.status.lookForConflictMarkers &&
  f.status.conflictMarkerCount === 0
```

 - conflicted files that need to be handled manually look like this:

```ts
f =>
  isConflictedFile(f.status) &&
  !f.status.lookForConflictMarkers
```

Where `isConflictedFile` is a helper function

```ts
export function isConflictedFile(
  file: AppFileStatus
}
```

Check out the rest of the diff for how these changes affected the Merge Conflicts dialog. I have another PR to start off which builds upon this too, because I need the `UnmergedEntrySummary` details when displaying messages in the app. That will help illustrate the benefits of these changes too.

### TODO

 - [x] thinking about the scenario where someone stages an unmerged file that still contains conflict markers - I think this might be affected by how we're building up these changes, but I'll test the current app correctly asks the user to open an editor
 - [x] more documentation about new and existing types
 - [x] rebase and retarget this PR on top of `master` once #6205  is resolved
 - [x] more testing of current scenarios in the app

## Release notes

Notes: no-notes


shiftkey:
> thinking about the scenario where someone stages an unmerged file that still contains conflict markers - I think this might be affected by how we're building up these changes, but I'll test the current app correctly asks the user to open an editor

In answer to this question, I created this test which uses the same checks as Desktop does when detecting merge conflicts:

```shellsession
$ git diff --check | grep  patches/common/skia/.patches
patches/common/skia/.patches:1: leftover conflict marker
patches/common/skia/.patches:2: leftover conflict marker
patches/common/skia/.patches:4: leftover conflict marker

$ cat patches/common/skia/.patches
<<<<<<< HEAD
=======
dcheck.patch
>>>>>>> origin/4-0-x
fix_font_thickness.patch

$ git add patches/common/skia/.patches

$ git diff --check | grep  patches/common/skia/.patches

$ 
```

On 1.5.0 and also this PR, as soon as the file is staged in the index (with the conflict markers still in source) it disappears from the output of `git diff --check` that we use to detect conflict markers. I think that's fine for now, because in our typical usage we don't stage changes until we create the merge commit after the files are resolved.

shiftkey:
Marking this as `ready-for-review` while I focus on doing some more testing of the current flow.

app/src/models/status.ts
```diff
@@ -301,3 +379,10 @@ function getIncludeAllState(
 
   return includeAll
 }
+
+/** Typechecker helper to identify conflicted files */
+export function isConflictedFile(
```
shiftkey: This would live in `app/src/lib/status.ts` but I opened #6242 which also creates this file so I plan to rebase this on top of that merged PR and then move it in there (yay merge conflicts!).

shiftkey:
@outofambit I'd love your thoughts on this too because of your experience with building out the conflict resolution UI and logic

shiftkey:
:apple:

shiftkey:
Moving this to `time-sensitive` because I'd like to get it wrapped up before I'm out next week, and leave the remaining feature work in #6218 to someone else to build upon

shiftkey:
I suspect because of Turkey Day and people taking time off this is not going to be properly reviewed until next week, which is then when I'm off some holidays of my own. If someone else on @desktop/engineering wants to pick this up and help land it I'd greatly appreciate it.

PR #6215, opened 2018-11-16
add previous value to ping payload to differentiate between scenarios

shiftkey:
## Overview

I was discussing with @billygriffin the onboarding experience and the reasons for sending this ping event when the user opts in or opts out of sending usage stats, but we spotted a gap in the design of the payload.

## Description

We realised we couldn't differentiate between these two scenarios a user in Desktop might do:

 - user installs the app, on first launch is asked whether they consent to sending usage stats, checks box to agree
 - user installs the app, did not consent, then some time later opens Options/Preferences, agrees to opt-in to sending usage stats, and presses Save to update their stats

Currently both of these actions send the same ping event with `{ optIn: true }` in the payload, but we'd like to be able to distinguish between them. The first scenario is an important step as part of onboarding new users, and the second is one an existing GitHub Desktop user would perform. We'd like to identify and understand any underlying trends around this area.

To distinguish between these two, I've added a `previousValue` field to the payload:

 - for the new user choosing whether or not to submit usage stats, `previousValue` will be `null` to indicate no previous choice had been set
 - for the existing user choosing to change their preference, `previousValue` will be the value found in local storage

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -248,7 +248,7 @@ export class StatsStore {
     // If the user has set an opt out value but we haven't sent the ping yet,
     // give it a shot now.
     if (!getBoolean(HasSentOptInPingKey, false)) {
-      this.sendOptInStatusPing(!this.optOut, storedValue || null)
+      this.sendOptInStatusPing(!this.optOut, storedValue)
     }
 
     this.enableUiActivityMonitoring()
@@ -601,7 +601,7 @@ export class StatsStore {
 
     this.optOut = optOut
 
-    const previousValue = getBoolean(StatsOptOutKey) || null
+    const previousValue = getBoolean(StatsOptOutKey)
 
     setBoolean(StatsOptOutKey, optOut)
 
@@ -772,14 +772,16 @@ export class StatsStore {
    */
   private async sendOptInStatusPing(
     optIn: boolean,
-    previousValue: boolean | null
+    previousValue?: boolean
   ): Promise<void> {
     const direction = optIn ? 'in' : 'out'
+    const previousValueOrNull =
+      previousValue === undefined ? null : previousValue
     try {
       const response = await this.post({
         eventType: 'ping',
         optIn,
-        previousValue,
+        previousValue: previousValueOrNull,
       })
       if (!response.ok) {
         throw new Error(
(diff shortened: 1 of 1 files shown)


shiftkey:
@telliott27 is `previousValue` fine for you to work with, or is there a different name we should use here?

telliott27:
@shiftkey that works perfect for me 😄

app/src/lib/stats/stats-store.ts
```diff
@@ -579,10 +581,12 @@ export class StatsStore {
 
     this.optOut = optOut
 
+    const previousValue = getBoolean(StatsOptOutKey) || null
```
niik: This will cause `previousValue` to be null both when there's no stored value and when the stored value is 'false' which seems odd to me. Should this be
```suggestion
    const storedValue = getBoolean(StatsOptOutKey)
    const previousValue = storedValue === undefined ? null : storedValue
```

app/src/lib/stats/stats-store.ts
```diff
@@ -227,12 +227,14 @@ export class StatsStore {
...
-    this.optOut = getBoolean(StatsOptOutKey, false)
+    const storedValue = getBoolean(StatsOptOutKey)
+
+    this.optOut = storedValue || false
 
     // If the user has set an opt out value but we haven't sent the ping yet,
     // give it a shot now.
     if (!getBoolean(HasSentOptInPingKey, false)) {
-      this.sendOptInStatusPing(!this.optOut)
+      this.sendOptInStatusPing(!this.optOut, storedValue || null)
```
niik: Similar question as below, shouldn't this be
```suggestion
      this.sendOptInStatusPing(!this.optOut, storedValue === undefined ? null : storedValue)
```

Or else we'll end up treating `false` as the same thing as the value not being set at all, right?

app/src/lib/stats/stats-store.ts
shiftkey: 👍 I'll figure out a way to push the type coercion closer to the payload

shiftkey: Should be addressed in 63679a5d82ab97260288c1d22fe385e59a0ff177

app/src/lib/stats/stats-store.ts
shiftkey: 63679a5d82ab97260288c1d22fe385e59a0ff177

shiftkey:
:apple:

PR #6122, opened 2018-11-07
UI changes to handle binary files when conflicted

shiftkey:
## Overview

**Closes #6066**

This improves how we display binary files - which do not contain conflict markers and have a different flow to resolve them - in the new merge conflict dialog.

## Description

This PR closes out the last parts of #6066:

 - conflicted binary diff files are now listed as conflicted and do not encourage you to open them in an editor

 - because binary files need to be staged to be resolved - removing the conflicted markers and potentially removing them from the working directory completely - the dialog needs to show something if it cannot find any conflicted files

TODO: 

 - [x] rebase on top of `master` once #6116 has been merged
 - [x] retarget PR to 1.5.0 release branch 

## Release notes

Notes: no-notes

PR #6218, opened 2018-11-16
[WIP] add support for choosing which version of a file to merge

### LINKED ISSUE #6062: display manual conflict resolution options in app

shiftkey:
## Overview

🌵🌵🌵**builds upon #6214, please review those first**  🌵🌵🌵

This obsoletes our "use command line to resolve binary files" message we added in #6122 so we could land 1.5.0, and replaces it with actual options that hopefully make sense.

The options should hopefully make questions like "do I want to take so-and-so version of this file?" clearer without needing to drop into the command line.

Here's a demo:

resolve-manual-conflicts

**Resolves #6062**

## Description

🌵🌵🌵**This PR does not target `master` to make the diff more readable. I eventually intend to land this on `master`**  🌵🌵🌵

There's a bunch of work still remaining to really polish this, but I wanted to get something up and playable for peeps early because I'm unavailable for three weeks from the 26th of November, so someone might have to wrap this up in my absence:

## TODO

 - [ ] review `createMergeCommit` code as that was thrown together to just get the demo working
 - [ ] review introduced `resolutions` abstraction to see if there's something better here
 - [ ] extract PR for extracting `resolved-file-item` and `conflicted-file-item` from `MergeConflictDialog` as the latter is becoming more advanced as part of this PR (it can show one of three different states)
 - [ ] ???
 - [ ] rebase and retarget this PR on top of `master` once #6205 and #6214 and ??? are resolved

## Release notes

Notes: Add ability to resolve unmerged changes without switching to the command line


shiftkey:
Closing this out because I'm going to be on holidays from next week, so someone else will need to pick this up to help get it reviewed. Once #6214 has been merged I'll rebase this branch and add more context to the original issue.

Issue #5734, opened 2018-09-25
loss of service to a local repository upon computer sleep


msftrncs:
## Description
Services to a local repository that is stored on a network share is lost if the computer goes to sleep or hibernates, or possibly with other network access issues.

## Version
* GitHub Desktop: 1.4.0
* Operating system: Microsoft Windows [Version 10.0.17763.1]

## Steps to Reproduce
Leave GitHub Desktop open on a repository that is stored on a network share (Windows Server 2012 R2) and then walk away and allow the computer to sleep and eventually hibernate.
Return to the computer and wake it up.

### Expected Behavior
I expect the state of GitHub desktop should return to normal (as was previous before the sleep), if not immediately, I expect it to try again (indefinitely until the user says otherwise) to locate the repository at the path it was last seen.

### Actual Behavior
You will find GitHub Desktop reporting 'Can't Find "repository"' along with a path to where it was last seen, and some buttons, 'locate...', 'clone' and 'remove'.

I restarted GitHub desktop, but it just came back up in the same state.  I tried pressing the 'locate...' button, but it took me to some other path way too far away (local system drive) to be worth the effort to navigate to.  I know the path exists because I still had Explorer windows open to the folder, and they are working fine, immediately after the awakening.  All other repositories are fine, even the ones on the same server share, so only the one that was open at the time the machine went to sleep is a problem.

### Logs
[2018-09-25.desktop.production.log](https://github.com/desktop/desktop/files/2413481/2018-09-25.desktop.production.log)

msftrncs:
Apparently the repository became corrupt during the sleep event, locating it is doing no good.  The issue probably happens when the computer sleeps while GitHub has some uncommitted writes to the database.  Fortunately the repository was synced.

msftrncs:
Additional, the database and the older seem to be intact, as GIT and VS Code operated repository just fine.

shiftkey:
@msftrncs thanks for the report.

>  I know the path exists because I still had Explorer windows open to the folder, and they are working fine, immediately after the awakening. 

Once Desktop is unable to find the repository on disk, it'll stay in that "missing" state until being resolved. Anecdotally, we mostly see this "missing" flow occur when someone renames or deletes a folder on disk without updating Desktop. We've not talked about checking the location again to see if the folder is accessible again, which might help for scenarios like repositories on network shares, but I'm not sure how many of our users are working with this sort of setup.

[...]

Issue #4264, opened 2018-03-16
Repository is not rechecked on client restart and there is no option to simply refresh od auto-locate


tobiasschuerg:
## Description
I've some repros on removable storage. Sometimes I start the client before the storage is available and the client will say "can't fin xyz":

![image](https://user-images.githubusercontent.com/1324555/37510795-a15d8012-28fc-11e8-93eb-50ff190cc9bf.png)

There is the "locate" option which works but is quite unhandy if the path didn't change.
Even after a program restart, the path is not checked again and the message still appears.

Therefore I'm missing a "try again", "refresh option" or "auto-locate"-option which just makes the client re-check the already known location if the git repo is available again.

## Version

* GitHub Desktop:
![image](https://user-images.githubusercontent.com/1324555/37510887-fd54c074-28fc-11e8-9267-28d7b8d99c3c.png)

* Operating system:
Windows 10 pro

## Steps to Reproduce
1. creata a virtual drive
2. mount it
3. create a git repro 
4. unmount
5. -> "can't find"
6. mount
7. **"can't find"**
8. restart client
9. still **"can't find"**

### Expected Behavior
That the old repository location is automatically checked again and it is found (bold points above).

6. mount
7. restart client
8. repo is foudn again

### Actual Behavior
Client says "can't find" until the repo is manually relocated.

j-f1:
Thanks! It looks like this issue is with the Classic apps, but this repository is for issues with the redesigned GitHub Desktop (see https://desktop.github.com).

tobiasschuerg:
Oh, right.

tobiasschuerg:
However, Github Desktop has the same issue:

![image](https://user-images.githubusercontent.com/1324555/38076222-46d88996-3335-11e8-9ff1-91fbe8d926fe.png)

Here I would like to have an option such as: *Refresh* or *Try Again* or even better it would refresh automatically (on repo change).

iAmWillShepherd:
@tobiasschuerg, would you mind updating your original issue description with what you just posted?

tobiasschuerg:
done

shiftkey:
I'm going to tag this as `help wanted` as I think this is reasonably isolated from the rest of the application. I'm not sure about the right UX for this, but we can try and implement something and get a feel for it.

[...]

PR #6228, opened 2018-11-18
auto recover missing repository

msftrncs:
## Overview

Addresses issues #4264, #5734.  Attempts to recover the repository where it was last known to be, automatically. 

## Description

Updates to 'missing-repository' to trigger a rechecking of the repository state after a delay, and repetitively with a delay while the repository remains in the missing state.  Technically every action the user performs while the missing repository view is up, triggers a rechecking of the repository, after a delay.

## Release notes
 - no notes

Notes:
On a network environment, some operations may become hesitant depending on the order of events or something.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -131,6 +131,7 @@ import {
   IStatusResult,
   createMergeCommit,
   getBranchesPointedAt,
+  isGitRepository,
 } from '../git'
 import {
   installGlobalLFSFilters,
@@ -1878,6 +1879,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
 
     return Promise.resolve()
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _attemptRecoverMissingRepository(
+    repository: Repository
+  ): Promise<void> {
+    /* 
+        if the repository is marked missing, check to see if the file path exists, 
+        and if so then see if git recognizes the path as a valid repository,
+        and if so, reset the missing status and then attempt to refresh
+      */
+    if (repository.missing) {
+      const updatedRepository = await this._updateRepositoryMissing(
+        repository,
+        (await pathExists(repository.path))
+          ? !(await isGitRepository(repository.path))
+          : true
+      )
+      if (!updatedRepository.missing) {
+        // repository has returned, attempt to refresh it now.
+        return await this._refreshRepository(updatedRepository)
+      }
+    }
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1079,7 +1079,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     setNumber(LastSelectedRepositoryIDKey, repository.id)
 
-    if (repository.missing) {
+    // if repository might be marked missing, try checking if it has been restored
+    const refreshedRepository = await this.recoverMissingRepository(repository)
+    if (refreshedRepository.missing) {
       // as the repository is no longer found on disk, cleaning this up
       // ensures we don't accidentally run any Git operations against the
       // wrong location if the user then relocates the `.git` folder elsewhere
@@ -1087,6 +1089,16 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return Promise.resolve(null)
     }
 
+    return this._selectRepository_refreshTasks(
+      refreshedRepository,
+      previouslySelectedRepository
+    )
+  }
+
+  private async _selectRepository_refreshTasks(
+    repository: Repository,
+    previouslySelectedRepository: Repository | CloningRepository | null
+  ): Promise<Repository | null> {
     this._refreshRepository(repository)
 
     const gitHubRepository = repository.gitHubRepository
@@ -1885,22 +1897,34 @@ export class AppStore extends TypedBaseStore<IAppState> {
   public async _attemptRecoverMissingRepository(
     repository: Repository
   ): Promise<void> {
+    // if repository is missing, try checking if it has been restored
+    if (repository.missing) {
+      const updatedRepository = await this.recoverMissingRepository(repository)
+      if (!updatedRepository.missing) {
+        // repository has been restored, attempt to refresh it now.
+        return this._refreshRepository(updatedRepository)
+      }
+    }
+  }
+
+  private async recoverMissingRepository(
+    repository: Repository
+  ): Promise<Repository> {
     /* 
         if the repository is marked missing, check to see if the file path exists, 
         and if so then see if git recognizes the path as a valid repository,
-        and if so, reset the missing status and then attempt to refresh
+        and if so, reset the missing status as its been restored
       */
-    if (repository.missing) {
-      const updatedRepository = await this._updateRepositoryMissing(
-        repository,
-        (await pathExists(repository.path))
-          ? !(await isGitRepository(repository.path))
-          : true
-      )
-      if (!updatedRepository.missing) {
-        // repository has returned, attempt to refresh it now.
-        return await this._refreshRepository(updatedRepository)
-      }
+    if (
+      repository.missing
+        ? (await pathExists(repository.path))
+          ? await isGitRepository(repository.path)
+          : false
+        : false
+    ) {
+      return this._updateRepositoryMissing(repository, false)
+    } else {
+      return repository
     }
   }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1895,7 +1895,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
-  public async _attemptRecoverMissingRepository(
+  public async _refreshOrRecoverRepository(
     repository: Repository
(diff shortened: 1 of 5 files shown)


msftrncs:
@Daniel-McCarthy, Thanks for the feedback.  I got to noticing that the 'MissingRepository' happens after a window focus event.  Also, it was seeming like the auto recovery was kicking in on a window focus event as well, but in reality it was on any user activity, such as mousing between menus, and I thought that level of triggering the GIT process to detect the return of the repository shouldn't happen, and so changed it to only perform the check after the time delay, instead of before it.

I've followed through the code and determined how the window focus event triggers the detection of the missing repository, and the detection is done first most with `pathExists` in `_refreshRepository` in 'app-store.ts'.  I did not utilize `pathExists` as a precheck before checking with `isGitRepository`, and I should as it will make less activity.   Also by checking the path first, the logic could be changed so the repeat rate could be different when its a git error rather than a root folder existence issue.

Now, more on the window focus event.  This means there is already a window focus event in 'index.tsx'.  It currently calls to refresh the repository if the selected repository is not already in the missing state. I could see this whole process being changed so that `_refreshRepository` is responsible for recovering a repository that has returned, since it was responsible for marking it missing in the first place.

On a further note regarding window focus.  It seems common place today for users to consider all events in the system to be interconnected.  Thus, if I delete or rename a folder in Explorer, I tend to expect (only though examples all around) that GitHub Desktop might see that immediately.  Programmers may be aware of the actual complexity that this interconnected appearance requires (file watchers, etc...) but common users just expect it to be.

msftrncs:
I missed that selecting a missing repository should probably also attempt to recover the repository, and have included a commit to cover that as well.  The approach in 'app-store.ts' (AppStore) `_selectRepository` probably deserves some discussion.  Maybe it (`_selectRepository_refreshTasks`) should be an arrow function, and maybe that concept could be used in `_refreshRepository` too, but I am not sure how the async/await works through there.

app/src/lib/stores/app-store.ts
```diff
@@ -1078,14 +1079,26 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
       return Promise.resolve(null)
     }
 
+    return this._selectRepository_refreshTasks(
+      refreshedRepository,
+      previouslySelectedRepository
+    )
+  }
+
+  private async _selectRepository_refreshTasks(
```
outofambit: we tend to use camelCase for class methods so could `_selectRepository_refreshTasks` be `_selectRepositoryRefreshTasks`? I like that we are encapsulating this logic into its own method

outofambit:
@desktop/design could you take a look at this mockup and see if the button order and text looks good to you? This PR adds the `Refresh` button.

screen shot 2018-11-28 at 8 32 36 am

ampinsk:
Since this is covering an edge case, I don't think we need to emphasize the action with a button. How about moving it as a text link to the end of the message, and rewording to `Check again`?

frame

Daniel-McCarthy:
@msftrncs I have also had to do this recently and hit the same road block. I found that a place in the code base that does this is the About dialog.

To accomplish it, they stored the `LinkButton` in a variable.
https://github.com/desktop/desktop/blob/073b4561245ba2f0b9c4fc7b4914c2e25d3d2148/app/src/ui/about/about.tsx#L248-L250

Then, they embedded the variable with `{' '}` preceding.
https://github.com/desktop/desktop/blob/073b4561245ba2f0b9c4fc7b4914c2e25d3d2148/app/src/ui/about/about.tsx#L271-L272

This is how I ended up doing it as well as I was not able to find another way to get the space to show up as expected.

msftrncs:
https://github.com/desktop/desktop/blob/06d096c77b8baea16a56c328ee7deeca61fe6b8d/app/src/lib/stores/app-store.ts#L1898-L1909

https://github.com/desktop/desktop/blob/06d096c77b8baea16a56c328ee7deeca61fe6b8d/app/src/lib/dispatcher/dispatcher.ts#L248-L252

Would there be any value in changing (App-Store) `_attemptRecoverMissingRepository` to something like `_refreshOrRecoverRepository` and eliminating the ternary in (Dispatcher) `refreshRepository`? (an additional `else` would be needed in the current `_attemptRecoverMissingRepository` to be sure to complete the refresh operation.)  There could be some confusion however from App-Store indicating 'see `Dispatcher`' regarding `_refreshRepository`.

msftrncs:
I just want to mention a thought I have for UI reaction to the 'Check again' concept.  Without any reaction, the user would think nothing is happening.  My thought would be to change the text, from 'Check again' to 'Checking...' for a small period of time, the maximum of which would be some constant period of time (0.5 to 1 second maybe) via a timer, or the length of time it takes to complete, unsuccessfully, the repository recovery tests.

billygriffin:
Seems reasonable to me @msftrncs, always have ❤️ for feedback to users on what's taking place. 😄

ampinsk:
This looks great @msftrncs, thanks so much for making that change! Would it be possible to center align the text as well?

msftrncs:
@ampinsk,  I too noticed the uncentered appearance, but I think centering the text 'It was last seen at "{path}". Check again.' might result in a small side affect.  When the text is wrapped, both the first line and each of the continuing wrapped line(s) will be separately centered.   I'll check this out and see what it looks like.

outofambit:
@msftrncs 

> Would there be any value in changing (App-Store) _attemptRecoverMissingRepository to something like _refreshOrRecoverRepository and eliminating the ternary in (Dispatcher) refreshRepository?

i like this idea a lot! good call on moving logic out of dispatcher into app-store. 👍

ampinsk:
@msftrncs that looks great, thank you!

ampinsk:
@msftrncs Ooh sorry, a couple things I didn't notice before:

- Should the `Check again` button be `Clone again`?
- The `Locate` button should be on the end on the left

cc @outofambit

msftrncs:
@ampinsk, @outofambit, sorry, I had left the button Check Again along with the textlink 'Check again.' for evaluation.  This particular repository that is shown in my image isn't a GitHub repo, so the 'Clone Again' button is not shown.  Also, the centering of the text was not included in my most recent commit, nor was any attempt to provide a UI reaction to clicking the textlink.  I can get the first time items commited later this evening, and I can work on the reaction part in a later PR if that it desired.

tierninho:

- [x] check again, etc is surfaced when repo is moved
- [x] check again finds repo if path restored
- [x] check again ignores non-git repo of same name
- [x] repo is auto-added back if placed back in prior path
- [x] repo is added back if new path located manually

Notes:
- two log identical entries appear each time I click Check Again: `debug: [ui] Executing getTopLevelWorkingDirectory: git rev-parse --show-cdup`
- We should surface the "This is not a git repository." warning msg if I click Locate and attempt to clone a non-git repo. If you cmd-O, and do the same, the warning is shown as so:
screen shot 2018-12-05 at 11 22 42 am
- In the Repo list, the "up arrow" is still intact if the repo is missing. Should we remove this until the repo is located again?
screen shot 2018-12-05 at 11 17 13 am
- A spinning icon next to "Check Again" would be a nice confirmation that the button "worked" or this: https://github.com/desktop/desktop/pull/6228#issuecomment-443555770

Build/OS: 1.5.1-beta1, mac, 549e6ce2db6abd03cd9d819914d5bbc8e2ced8e8

tierninho:
> I have confirmed that the process of selecting a repository has no effect on the 'up' or 'down' arrows shown in the repository list on missing repositories.

@msftrncs Thanks for confirming above. The icon _does_ go away after moving the repo to another location, clicking the Clone Again button, dismissing the error, and then reopening the repo list. Not a good use case, but might offer a clue to fixing it. 💭

outofambit:
this looks good to me from a technical perspective @msftrncs! @ampinsk are we good on the design side? Here's a screenshot of the current state:

screen shot 2019-01-07 at 9 36 27 am

shiftkey:
Assigning myself to also review this because this all happened while I was out on leave last month and I just now noticed it's planned for `1.6.0`. `selectRepository`/`refreshRepository` is a important hot path for the app I'd like to test this out to ensure the default happy path isn't affected by this change.

app/src/lib/stores/app-store.ts
```diff
@@ -1880,6 +1894,43 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    repository: Repository
+  ): Promise<Repository> {
+    /* 
+        if the repository is marked missing, check to see if the file path exists, 
+        and if so then see if git recognizes the path as a valid repository,
+        and if so, reset the missing status as its been restored
+      */
+    if (
+      repository.missing
+        ? (await pathExists(repository.path))
```
shiftkey: :art: I'm not a huge fan of the readability of this nested ternary with multiple `await`-s inside an `if` statement. I had a go at getting these same rules down in a way that's hopefully easier to understand and makes the large code comment less important. Happy to get this into a follow-up PR:

```ts
  private async recoverMissingRepository(
    repository: Repository
  ): Promise<Repository> {
    if (!repository.missing) {
      return repository
    }

    // if the repository is marked missing, check to see if the file path exists,
    // and if so then see if git recognizes the path as a valid repository,
    // and if so, reset the missing status as its been restored
    const exists = await pathExists(repository.path)
    if (exists) {
      const isGitRepo = await isGitRepository(repository.path)
      if (isGitRepo) {
        return this._updateRepositoryMissing(repository, false)
      }
    }

    // otherwise just return the original repository
    return repository
  }
```

outofambit:
Awesome! 🎉 Thanks for your work on this @msftrncs!!

Issue #2480, opened 2017-08-17
large images are not scaled to stay inside diff panel


shiftkey:
### Description

With the new image diff improvements in #2383 there's a scenario where a large image is drawn outside the diff panel. We shouldn't do this.

### Version

**GitHub Desktop version:** current `master` e.g. 82b1d2377b98208d74aff0337c8f482cef2c1e35
**OS version:** Windows 10

### Steps to Reproduce

1. Clone https://github.com/shiftkey/image-examples 
1. Switch to History tab
1. Select this commit https://github.com/shiftkey/image-examples/commit/10c1cf4d08b884f83332d8ae24e14336c5ad73df
1. Ensure you have `2-up` selected (but the others show this problem too)

**Expected behavior:** 
  - Image does not display outside diff view.
  - Can toggle between different modes.

**Actual behavior:** 
  - Image drawn over top menu.
  - Unable to change diff mode.

<img width="1440"  src="https://user-images.githubusercontent.com/359239/29398473-498201a4-8368-11e7-8ad7-7887caafb759.png">

**Reproduces how often:** 100%

joshaber:
😱 

I think we should fix this for 1.0.

joshaber:
Huh, interestingly I can't reproduce this on Mac 🤔

joshaber:
Hrm, I can't reproduce this in my Windows 8 VM either. Gonna try getting a Windows 10 VM running.

joshaber:
I'm having VMWare troubles with Windows 10, so I'm gonna unassign myself until I'm able to reproduce it.

iAmWillShepherd:
I'm unable to reproduce this issue. Are you still having it?

tierninho:
Was just looking at this too, can't reproduce either.

joshaber:
Gonna 👢 this until we can reproduce it.

shiftkey:
Still reproducible on `0.8.1-beta2` on a Windows 10 VM:

<img width="992" src="https://user-images.githubusercontent.com/359239/29948489-771ee34e-8ef2-11e7-97d2-e72426099600.png">

:thumbsup: to defer until someone else can reproduce this.

PragmaticEd:
Why is this closed, if it still happens?
I uploaded git project [here](https://github.com/PragmaticEd/github-large-image-preview-scale-bug), that reproduces the problem. @tierninho  please open the issue again, till fixed.

PragmaticEd:
[...]

PR #6231, opened 2018-11-18
Add workaround to allow toggling between diff modes

06b:
## Overview

**Workaround to allow toggling between different modes in issue #2480**

## Description

When a large image gets drawn out of the diff panel (due to DPI scaling being larger than 100%) it prevents switching the diff mode because the image renders above tab bar. While this doesn't fix the DPI image scaling issue, it does allow someone to be able to toggle the diff modes.

Before:
githubdiffbefore

After:
githubdiffafter

Tested via the developer tools before making the edit.

## Release notes

Notes: Fixes problem where large image prevent toggling of diff modes in Windows when DPI scaling settings are larger than 100%

### DIFF
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
@@ -283,7 +283,7 @@
   .tab-bar {
     width: 350px;
     margin: 10px auto 0;
-    // Workaround to allow toggling between different modes until 
+    // Workaround to allow toggling between different modes until
     // root issue with DPI scaling is addressed, see https://github.com/desktop/desktop/issues/2480
     z-index: 0;
   }
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
@@ -283,6 +283,9 @@
   .tab-bar {
     width: 350px;
     margin: 10px auto 0;
+    // Workaround to allow toggling between different modes until 
+    // root issue with DPI scaling is addressed, see https://github.com/desktop/desktop/issues/2480
+    z-index: 0;
   }
 
   .image-diff-previous {


shiftkey:
Gonna let this sit in case others with CSS skills have a better way to achieve this

06b:
The following gif shows the tab bar will still appearing both the scenario which triggers the bug (in the gif 125% is used for an example) and in the normal scenario of the scaling at 100%
githubdesktop-6231

@shiftkey While testing the different scenarios I think I found a solution which would also address the image being drawn over the top menu for this issue (along with the sliders appearing behind the image which I didn't notice previously). 

Let me know if you would like for me to make those changes to this pull request. The only downside I'm seeing is that the tab bar may shift (Note: They don't appear to shift in the normal scenario [Not shown in gif, but if you want to see it a 100% scaled gif let me know and I'll attach it] ). See gif below:
githubdesktop-6231-part2

shiftkey:
> While testing the different scenarios I think I found a solution which would also address the image being drawn over the top menu for this issue (along with the sliders appearing behind the image which I didn't notice previously).

@06b I'm happy to leave this for a follow-up PR - this current experience is already improved over what the app previously did :tada:

WorkShop-Office:
ok



Sent from my Redmi 4A
On Brendan Forster <notifications@github.com>, Nov 21, 2018 10:48 PM wrote:

Merged #6231<https://github.com/desktop/desktop/pull/6231> into master.

hdsantos89:
Is this supposed to be working already? I don't get the ability to toggle between different views

shiftkey:
@hdsantos89 not sure I have the full context here - could you share more about what you're trying to diff and how the app is currently behaving? Screenshots would be ideal here too if you can provide them.

PR #6238, opened 2018-11-19
Updating usage metrics to connect to GitHub usage

billygriffin:
This PR affects how we generate and submit usage metrics, and we want to be transparent about what it means technically and for the product. Here are some details around why we're making this change and why we believe it will lead to a better product for our users.

## Context

If you've been following the GitHub Desktop repository for a while, you probably noticed recent pull requests to add analytics for new features in GitHub Desktop, such as [the dark theme](https://github.com/desktop/desktop/pull/4932), the [Notifications of Diverging from Default Branch](https://github.com/desktop/desktop/issues/4883) prompts, and the new [merge conflicts flow](https://github.com/desktop/desktop/issues/5394). We’ve been rolling these out alongside features to help us understand how they’re being used in the wild and if they're solving the problems that we're attempting to solve for users.

In tandem with user research, metrics like these help us answer questions like:

- **What OS are our users running GitHub Desktop on?** This helps us be aware of whether users are on recent OS versions, or whether we have a significant user base on older versions.
- **How are they using GitHub Desktop?** Users might be happy to just use GitHub Desktop for viewing their repository alongside the command line, or they might be making lots of commits and pushing to GitHub without needing to use the command line.
- **Is GitHub Desktop fast to launch?** GitHub Desktop is intended to enhance the productivity of its users, and we believe performance is an important piece of delivering the best experience possible.
- **How are new features in GitHub Desktop being used?** High or growing usage of features suggests we’re building the right things for GitHub Desktop’s target audience, but declining or low usage of features might suggest we need to investigate what didn’t work.
- **Is GitHub Desktop improving people's ability to collaborate?** There are various metrics here, but essentially we want to measure things that help us understand whether GitHub Desktop is reducing friction to users' ability to collaborate effectively.

## A Gap in our Understanding

Historically we anonymized the usage statistics submitted. This meant we had no insight into our users' developer journeys, workflows, or preferences, aside from whether they had signed into GitHub Desktop with a GitHub or GitHub Enterprise account.

This has prevented the team and other groups inside GitHub from being able to answer questions like:

- Does GitHub Desktop help users gain new skills and experience as developers?
- Is GitHub Desktop helping people contribute to projects on GitHub?
- How can we better support people's end to end workflows from GitHub Desktop to GitHub.com?

While we do support connecting to other Git remotes, GitHub Desktop is primarily a GitHub product, and we’d love to be able to answer these questions so we can make GitHub and GitHub Desktop better for both our current and future users. We're also conscious of potential concerns around collecting additional metrics, and therefore we only look at aggregate data and trends as opposed to data from any specific individual.

For reference, you can see examples of the usage statistics we currently submit on the GitHub Desktop website: https://desktop.github.com/usage-data/. This is currently being refreshed in co-ordination with this PR.

## Changes to how we report usage statistics

An upcoming update will show a dialog to users explaining these changes and what it means for you:

#### If you opt out from sending usage stats in GitHub Desktop:

This change does not affect you. We’d love for you to opt-in and participate so we can incorporate your workflows into our decisions, but we will continue to respect this setting in GitHub Desktop.

#### If you opt in to sending usage stats in GitHub Desktop, and...

- **You are signed into a GitHub account:** GitHub Desktop will add a `user_id` field containing your GitHub account id with each usage report. On our usage reporting backend, we can use this to identify the user associated with it, and pass that on to our analytics backend.
- **You are only signed into a GitHub Enterprise account:** this change does not affect you. We will continue to use the anonymized identifier `guid` when submitting usage reports.
- **You are not signed into either a GitHub or GitHub Enterprise account:** this change does not affect you. We will continue to use the anonymized identifier `guid` when submitting usage reports.

### DIFF
--- a/app/src/lib/dispatcher/dispatcher.ts
+++ b/app/src/lib/dispatcher/dispatcher.ts
@@ -682,8 +682,11 @@ export class Dispatcher {
   }
 
   /** Set whether the user has opted out of stats reporting. */
-  public setStatsOptOut(optOut: boolean): Promise<void> {
-    return this.appStore.setStatsOptOut(optOut)
+  public setStatsOptOut(
+    optOut: boolean,
+    userViewedPrompt: boolean
+  ): Promise<void> {
+    return this.appStore.setStatsOptOut(optOut, userViewedPrompt)
   }
 
   public markUsageStatsNoteSeen() {
--- a/app/src/lib/dispatcher/dispatcher.ts
+++ b/app/src/lib/dispatcher/dispatcher.ts
@@ -686,6 +686,10 @@ export class Dispatcher {
     return this.appStore.setStatsOptOut(optOut)
   }
 
+  public markUsageStatsNoteSeen() {
+    this.appStore.markUsageStatsNoteSeen()
+  }
+
   /**
    * Clear any in-flight sign in state and return to the
    * initial (no sign-in) state.
--- a/app/src/lib/stats/index.ts
+++ b/app/src/lib/stats/index.ts
@@ -1,3 +1,7 @@
 export { StatsDatabase, ILaunchStats } from './stats-database'
 export { StatsStore, SamplesURL } from './stats-store'
 export { getGUID } from './get-guid'
+export {
+  hasSeenUsageStatsNote,
+  markUsageStatsNoteSeen,
+} from './usage-stats-change'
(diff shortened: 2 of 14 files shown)


iAmWillShepherd (approved):

WorkShop-Office:
yes paa



Sent from my Redmi 4A
On William Shepherd <notifications@github.com>, Nov 21, 2018 10:51 PM wrote:

Merged #6238<https://github.com/desktop/desktop/pull/6238> into master.

iAmWillShepherd:
@telliott27 noticed some odd values come back to us from our beta users which prompted us to double-check if everything was being done properly. Turns out, they weren't. Here's a table of what we expect versus what's being reported.

PR #6242, opened 2018-11-19
extract functions living under the models folder

shiftkey:
## Overview

A bit of cleanup done in isolation of #6214 - these two helper functions should live somewhere else.

## Description

- `mapStatus` - for converting an `AppFileStatusKind` into a string - is now under `app/src/lib/status.ts`

- `iconForStatus ` - for converting an `AppFileStatusKind` into an `OcticonSymbol` - is now under `app/src/ui/octicons/status.ts`

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/models/status.ts
+++ b/app/src/models/status.ts
@@ -1,6 +1,4 @@
 import { DiffSelection, DiffSelectionType } from './diff'
-import { OcticonSymbol } from '../ui/octicons'
-import { assertNever } from '../lib/fatal-error'
 import { ConflictFileStatus } from './conflicts'
 
 /**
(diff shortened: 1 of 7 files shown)


app/src/models/status.ts
```diff
@@ -1,6 +1,4 @@
 import { DiffSelection, DiffSelectionType } from './diff'
-import { OcticonSymbol } from '../ui/octicons'
```
shiftkey: The red flag for me about these functions was that we had models referencing types defined in the UI folder. With this change we've moved `iconForStatus` closer to the things it needed to import.

PR #6244, opened 2018-11-20
Update roadmap post-release

billygriffin:
This is not time sensitive. I did a few things here:

1. Added the things that were previously on the roadmap and shipped in 1.5 to a past category. I'm not wedded to it, but it might be a helpful way to have a place to come back to to revisit the features that shipped in the past few releases at any given time and evaluate success (especially if we have associated metrics).

2. Added upcoming areas of exploration to the roadmap to make it a bit more robust and communicative. Because a few of the things are earlier in the process there's not a ton of amplifying info (like Accessibility), but I think that may be ok. Open to other opinions.

3. Altered formatting somewhat to make it more readable to me (if others don't feel like it is, please say so).

### DIFF
--- a/docs/process/roadmap.md
+++ b/docs/process/roadmap.md
@@ -2,22 +2,43 @@
 
 The following are the larger areas of upcoming work the GitHub Desktop team intends to explore. This is not inclusive of everything we're working on, and it's not written in stone. We'll continue to update it as our priorities evolve.
 
-### Onboarding
+#### Onboarding
   
-- Evaluate and improve onboarding for new users: [#5686](https://github.com/desktop/desktop/issues/5686)
-- Measuring success: [#5549](https://github.com/desktop/desktop/issues/5549)
-  
-### Merge conflicts handling
+- Improve onboarding for new users: [#5686](https://github.com/desktop/desktop/issues/5686)
+- Measuring success: [#5549](https://github.com/desktop/desktop/issues/5549) + usability testing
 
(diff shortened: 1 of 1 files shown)


docs/process/roadmap.md
```diff
@@ -2,22 +2,43 @@
...
-
-### Support rebase workflows
+#### Support rebase workflows
 
 - Improve rebase workflows: [#5953](https://github.com/desktop/desktop/issues/5953)
 - Measuring success: TBD alongside the work + usability testing
+
+#### Accessibility improvements
+
+- To be further defined, but we want to prioritize scoping a chunk of work on improving accessibility for our users
```
say25: Are there something that are clearly in or out of scope? Internationalization, colors used or font used? 

I would imagine internationalization in particular might be highly requested but maybe too large to be part of this effort

shiftkey: > Are there something that are clearly in or out of scope? Internationalization, colors used or font used?

The primary focus here is that Desktop is well-supported with screenreaders like NVDA and VoiceOver - that users with accessibility needs are able to navigate and work with Desktop like everyone else.

say25: Sounds good. Just thought maybe we could be more specific to help those sorts of users vs color blind users

billygriffin: @say25 This one in particular will get more tightly scoped over the next 3 weeks.

Issue #6184, opened 2018-11-13
remove old feature flags


outofambit:
there are a number of feature flags in https://github.com/desktop/desktop/blob/development/app/src/lib/feature-flag.ts that just return `true`. let's remove these functions and all references to them in the codebase. their respective features should be always be on.

- [x] `enableNotificationOfBranchUpdates()` - #6248
- [x] `enableRepoInfoIndicators()` - #6251
- [x] `enableMergeConflictDetection()` - #6250
- [x] `enableInAppReleaseNotes()` - #6243 
- [x] `enableMergeConflictsDialog()` - #6249

shiftkey:
To be extra clear here, I'd recommend doing **one PR per feature flag**, as some changes will be more complex than others and it'll make reviews easier.

Also, `enableStatusWithoutOptionalLocks()` has only been enabled for everyone on `1.5.0` which is going out tomorrow, so if we can leave that to last just in case that'd be :gem:

say25:
Good to close?

PR #6249, opened 2018-11-20
remove merge conflict dialog feature flag

outofambit:
## Overview

#6184 (for merge conflict resolution)

## Description

-

## Release notes

(not user facing)


shiftkey:
Dropping `ready-for-review` because there are outstanding conflicts

iAmWillShepherd:
I was initially going to fix this, but I am no longer sure how we track a files conflict state. We used to use the `AppFileStatusKind` type to delineate between files that had conflicts and used it return an `AbortMergeWarning`. Now that we have removed the `AppFileStatusKind.Resolved` enum value, I'm not sure which value to use instead.

billygriffin:
@iAmWillShepherd let's go ahead and leave this one until @outofambit is back from OOO then if that seems ok.

outofambit:
ready! @iAmWillShepherd 🍏

PR #6258, opened 2018-11-21
Gets rid of some console logging in one of our components

iAmWillShepherd:
Noticed we were using the `console.log` function to log stuff in the `Acknowledgements` component.


app/src/ui/acknowledgements/acknowledgements.tsx
```diff
@@ -48,14 +48,13 @@ export class Acknowledgements extends React.Component<
...
-        console.error('Error loading licenses')
-        console.error(err)
+        log.error('Error loading licenses', err)
         return
       }
 
       const parsed = JSON.parse(data)
       if (!parsed) {
-        console.error(`Couldn't parse licenses!`)
+        log.warn(`Couldn't parse licenses!`)
```
iAmWillShepherd: I'm using warn here since there is no error we capture to log.

PR #6260, opened 2018-11-21
spaces in filenames are problematic for profile scripts

shiftkey:
## Overview

This is the upstreamed version of https://github.com/shiftkey/desktop/pull/87 which addresses a Linux-specific issue with the Debian installer

## Description

- I updated the product name of the package in the Linux fork to `GitHub Desktop` (it was previously `GitHubDesktop` without a space
 - This uncovered that the generated profile script had a filename with a space in it
 - And it also uncovered that the generated script (which adds the install directory to `PATH`) had a space that wasn't handled properly either

## Release notes

Notes: no-notes

### DIFF
--- a/script/linux-after-install.sh
+++ b/script/linux-after-install.sh
@@ -9,8 +9,8 @@ export PATH=$INSTALL_DIR:\$PATH"
 
 case "$1" in
     configure)
-      echo "$SCRIPT" > ${PROFILE_D_FILE};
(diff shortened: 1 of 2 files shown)


shiftkey:
#jolo because this only affects Linux installers

shiftkey:
Actually, we can't #jolo any more without an approval.

@iAmWillShepherd can you give this a 👍 when you have a spare moment?

PR #6265, opened 2018-11-22
Image preview fix for dpi scaling issue

### LINKED ISSUE #2480: large images are not scaled to stay inside diff panel

06b:
## Overview

**Closes #2480**

## Description

When a large image gets drawn out of the diff panel (due to DPI scaling being larger than 100%) and overlaps the top menu. This fixes the issue where it breaks out of the container and also allows for the slider to render above the image.

48798075-d45d3b80-ecd1-11e8-8795-a1a4af686581

## Release notes

Notes: 
 - Fixes problem where large images render above the top bar in Windows when DPI scaling settings are larger than 100%
 - Fixes problem where sliders for swipe & onion skin modes are below image in Windows when DPI scaling settings are larger than 100%
 - Improves preview of newly added large images by allowing scrolling to the preview

### DIFF
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
@@ -386,6 +386,7 @@
 
   .slider {
     align-self: center;
+    z-index: 0;
   }
 
   .sizing-container {
@@ -429,6 +430,7 @@
 
   .slider {
     align-self: center;
+    z-index: 0;
   }
 
   .sizing-container {
--- a/app/styles/ui/changes/_changes-view.scss
+++ b/app/styles/ui/changes/_changes-view.scss
@@ -71,4 +71,5 @@
   display: flex;
   flex-grow: 1;
   min-width: 0;
+  overflow: overlay;
 }


niik:
Hey @06b, thanks so much for opening this. I'd rather not we opened up this container for scrolling as a way to work around this problem as it's not a very discoverable pattern (especially not on macOS where the scrollbars aren't visible until you actually scroll).

Is there any chance we could instead investigate why the sizing logic we have in place isn't working correctly when using a DPI value other than 100%?

Issue #5686, opened 2018-09-19
Improve initial onboarding experience


billygriffin:
Based on feedback from new users and user interviews, we think there's an opportunity to improve the onboarding experience in GitHub Desktop and reduce the time between when people download and when they can be more productive using the app. 

We're working on getting baseline metrics in place in #5549, and this issue is intended to capture the work that follows from that.

Specifically, we'd like to reduce the amount of time to each action captured in #5549 and increase the percentage of new users who get to those actions.

**Actions are:**

- Adding repo, cloning repo, creating repo
- Committing
- Being on a non-default branch
- Pushing to GitHub

**The things we want to move the needle on:**

1. Time from first launch of GitHub Desktop to [action] - we want these to go down
2. Percent of installs that reach each action - we want these % to go up

**Problem statements to focus on for our initial iteration:**

1. As a new user to GitHub Desktop, after I've gone through the initial sign up flow and added a repository, it's not clear to me what value GitHub Desktop offers me or what I should do next.

2. As a new user to GitHub Desktop, after I've gone through the initial setup, I'm landed on a page that gives me three buttons to add a repository, but doesn't help me through that process - which should I choose, what do they mean, is there one that I should likely be using?

Steps:

- [x] Initial design exploration
- [x] Implementation
- [x] Usability testing with small group of beginner users
- [x] Iterate
- [x] Release

billygriffin:
This comment (https://github.com/desktop/desktop/issues/1688#issuecomment-303386625) stemming from this issue #5738 is likely worth examination. Thanks @cheshire137! 😄

tierninho:
@niik Possible to post the mocks for this when ready? Thanks!

niik:
> @niik Possible to post the mocks for this when ready? Thanks!

Sorry @tierninho, I didn't see this until now 🤦‍♂️. There are some screenshots over in #6365 and, you know, the real thing.

shiftkey:
@billygriffin I've updated the checklist of issues now that #6445 has been merged, and I plan to put together a beta release with the last of our changes in there. Let me know if there's anything else that needs to be done as part of 1.6 for this, otherwise I'd like to close it out before the 🚢.

shiftkey:
I'll defer to @niik on when it's good to close this out, given the last part of this feature just landed on `beta` today.

billygriffin:
Since this is all merged into the release branch and ready to 🚢 tomorrow, gonna close it out!

PR #6278, opened 2018-11-23
Add a store for cloneable repositories

### LINKED ISSUE #5937: Not visually apparent which repo is selected to clone
This issue is borne out of some bad experiences with the Clone a Repository dialog. Since the issues all seem to stem from a root design choice, I'm going to group them all here under one issue, with the generic title.

- The three tabs (GitHub.com, Enterprise, URL) have shared state with regards to the repository URL and the local path. This creates weird UX:
  - Changing the local path under any tab affects the others.
  - The 'Clone' button might do something entirely different than you expect for the tab you're on.
    - Use the GitHub.com account shortcut (say, `desktop/desktop`) under the URL tab. Go to GitHub.com tab and change the local path. Ask someone else to look at the screen and predict what will happen if you hit clone.
- The GitHub.com tab's behavior is confusing.
  - There's no visual indication regarding which repository will be cloned. If you've selected something in the list that *might* be what will be cloned, unless a bug like #3596 exists, or you switch tabs and then go back to the list which will result in nothing being visually selected. There's no definitive visual indication about what will be cloned when you click the clone button.
  - Filtering, selecting a repository, then clearing the filter also has issues. The selected repository is now somewhere in your long list of repositories, and there's no way to see which one was selected without scrolling down the list until you see which one has the visual indication.

Personally I think this dialog could use a UI refresh.

- Get rid of shared state across the tabs, make them each independent like they logically seem.
- The main GitHub.com tab should have a 'Selected Repository' UI element that you can clear and which is separate from the selection list, which would be just for selecting, and would be disabled if there is already a selected repository.
- Drop the magic GitHub username and repository text from the URL tab, leave that tab for the simple task of providing a URL to clone into a local path. By making it magic text under that tab you make rich functionality too cumbersome to add, such as checking if it's a valid GitHub.com repo before trying to clone.
  - Instead put this functionality under the GitHub.com tab. Turn the 'Filter' box into a more generic 'Find', and include GitHub.com results under a separate header which will be below 'Your Repositories' (with de-duplication as necessary).

Here are some (awful) mockups of how the GitHub.com tab might look after those suggested changes.

image

image

niik:
## Overview

As part of the onboarding work I'm doing in #5686 I'm going to need access to the list of "cloneable" repositories in the blank slate. Currently the logic for fetching repositories available to the user from the API is contained within the clone dialog itself. This PR extracts that logic into a separate store responsible for loading and refreshing the list of repositories.

The improvements in this PR made the disadvantages of our current shared state model all the more noticeable (you couldn't switch between tabs and maintain selection) so I took the opportunity to remove the shared state n favor of per-tab as users would expect (see #5937). Each tab now maintains its own selection, filter text, path, etc.

Closes #5937 by maintaining separate state for each of the tabs in the clone dialog.

## Description

Additional benefits of this change

 - Since the repositories available for a user is now persisted we no longer have to load the repository list on each tab switch inside the Clone dialog.
 - Users will no longer have to wait for the repository list to be loaded every time they open the dialog and can instead choose to refresh manually with a new button

image

## Memoization

This PR introduces a memoization pattern that we haven't used before in the app which I think could use some further explanation. See [the react blog](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization) for a good write-up but basically I've made the `CloneGithubRepository` component pure and stateless and used memoization to ensure that we only recompute the filter list groups when necessary.

## Release notes

[Improved] Clone dialog remembers selections, filter text, and path in each tab
[Improved] The list of repositories in the clone dialog are no longer loaded from the API every time the dialog is opened or when the selected tab changes.

## TODO

 - [x] More documentation
 - [X] ~~Automatically refresh the repository list when some amount of time has elapsed since last time?~~
 - [x] ~~See if we can't get rid of the null state for repositories in the `CloneGitHubRepository` component~~
   - Punting on these two for now, the PR is big enough as it is.


niik:
@iAmWillShepherd Sorry for the post-review update but I found a race condition that I had missed when the accounts store was updating at the same time as the apiRepositoriesStore was so I decided it was better to fix it now before this PR lands.

app/src/lib/stores/api-repositories-store.ts
```diff
@@ -4,6 +4,41 @@ import { IAPIRepository, API } from '../api'
...
+function resolveAccount(
+  account: Account,
+  accountState: ReadonlyMap<Account, IAccountRepositories>
+) {
+  if (accountState.has(account)) {
+    return account
+  }
+
+  for (const existingAccount of accountState.keys()) {
+    if (accountEquals(existingAccount, account)) {
```
iAmWillShepherd: A comment here stating why we go from reference equality to value equality would help clarify the need to iterate through accounts after checking the set for it. ✨

niik: Pushed an expanded documentation, let me know if that feels better.

niik:
Ready for one last lookiloo!

billygriffin:
Shifting this to 1.5.1 per our team sync today.

PR #6281, opened 2018-11-24
Fixed avatar duplicating when user is co-author and committer

### LINKED ISSUE #6135: If coauthor and committer are identical, their avatar appears twice
Say we have the following scenario:
- Person A authored a commit.
- Person B committed it.
- Person B is also listed as a co-author in the `Co-authored-by` field .

Then the icons for contributors for that commit will show Person B twice.
Example:
image

The co-author and committer should be deduplicated.

bblarney:
## Overview

**Closes #6135**

## Description

If the committer is also a coauthor, their avatar would appear twice. There is currently a check to see if the committer is the author, but there was no check for coauthor, so I added one following the same style. If the committer is also a coauthor, their avatar will only be displayed once.

OLD:

unfixedgh

NEW:

fixedgh

(the number '2' is hiding behind the avatars, indicating both users are listed as author + coauthor)

I also made a small change in a test file where a commit object was created because of the new prop needing to be initialized.

## Release notes

Notes: Fixed user avatars appearing twice when listed as both co-author and committer


Daniel-McCarthy:
Hi @bblarney! Thank you for this contribution.
I see that the automated CI were giving you some trouble with the linting. I figured it was worth mentioning that we now have a [document available](https://github.com/desktop/desktop/blob/master/docs/contributing/linting.md) on linting which includes a command that can attempt to fix current linting issues for you.

You may already be familiar, but just in case it's useful, you can use `yarn lint` to let it run a test and show any linting errors from command line so you can check for them ahead of time.

Additionally there's `yarn lint:fix` which can attempt to fix the linting issues for you so you don't have to do it manually.

___
That said, I see you have closed this pull request. I hope you have not been discouraged as it's always great seeing new contributors! Keep up the great work!

bblarney:
Thanks for the help @Daniel-McCarthy, I had run the tests successfully locally but after I fixed the linting issues that one failed, looks like I didn't have it locally so I synced my repo and hopefully it works now

I was thinking maybe changing the commit.ts may cause more problems than I had thought and maybe I should just do the check in avatar.ts so only that file would need to change, so I closed the PR, but I am reopening with the change to the checkout-test and hopefully it works out.

bblarney:
Hi @niik thanks for the help

I've made the changes you suggested, putting the check in avatar.ts, this does look like a much simpler option.

Let me know if there's anything else I should change

app/src/models/avatar.ts
```diff
@@ -84,10 +85,18 @@ export function getAvatarUsersForCommit(
     )
   )
 
+  coAuthoredByCommitter = commit.coAuthors.some(
```
niik: Seeing how this variable is only computed once and then never assigned to again I think we should make it a constant and drop the `let` above.

```suggestion
  const coAuthoredByCommitter = commit.coAuthors.some(
```

Making it a `const` signals to people (and computers 😛) reading the code that it's immutable and therefore not going to change at any point after it's been assigned to vs `let` where they might have to read the entire method to see if it's being changed anywhere else.

bblarney: Thanks again, I've updated

niik:
Thank you so much for this contribution @bblarney!

billygriffin:
Great stuff @bblarney, thank you! ❤️

PR #6307, opened 2018-11-28
Update yarn.lock

outofambit:
## Overview

looks like our `package.json` and `yarn.lock` got out of sync somehow. this rectifies that :)


niik:
Gonna forego the 24hr merge rule on this

Issue #5030, opened 2018-06-25
visual indicators when hovering would help emphasize interactive parts of the app


shiftkey:
Extracted from #4989 

@jsunsawyer called out some feedback about our current visual theme: 

> There are several instances where a visual indicator (background change or darkened text) on hover would help users be sure that they're in the right spot before they click. Right now, you just have to click and hope that your mouse was in the right spot. It just slows you down a bit.
> 
> Places where hover states would help:
> 
> - Switching repositories
> - Switching tabs (Changes vs History and Branches vs Pull Requests)
> - Selecting a file to view in the Changes or History views
> - Selecting a branch to compare
> - Selecting a previous commit in the History view
> - Selecting a different branch or pull request
>
> At the very least, the cursor should change from an arrow to a pointer when you hover over an interactive element.

xt0rted:
Another spot that lacks a hover indicator is the files list for a commit

j-f1:
I’m wondering if this is a good idea. While many web apps do this sort of thing, I feel like most native apps don’t, and that’s one of the factors that makes them feel native.

shiftkey:
I'm going to drop `themes` from this while we experiment with figuring out how to address this.

billygriffin:
Added the new `user-research` label to convey that this would benefit from validation with external users.

Generally, however, I agree with the idea that giving users an indication that something is clickable is a good practice and that visual feedback can be very helpful. We do this in a few other instances in Desktop, just not everywhere. For example, the tabs "Current Repository", "Current Branch," and Push/Pull/Fetch all have a lighter hover state. 

I'd love thoughts from @donokuda and/or @ampinsk too.

With a bit more scoping and validation, could this also be something appropriate for a contributor from the community?

PR #6310, opened 2018-11-29
Hover states on list items and tabs

ampinsk:
## Overview

Fixes https://github.com/desktop/desktop/issues/5030

This adds a hover state to all list items in both light and dark mode:
- Uncommitted changes
- Commit history
- File list
- Repo list
- Branches list

And to tabs in:
- Changes panel
- Branches panel

Light mode:

hover-state-list

Dark mode:

hover-state-dark

The only problem I've run into and can't figure out how to address is the hover style is also applied to these header items:

screen shot 2018-11-28 at 4 09 19 pm

### DIFF
--- a/app/src/ui/lib/list/list-row.tsx
+++ b/app/src/ui/lib/list/list-row.tsx
@@ -37,6 +37,13 @@ interface IListRowProps {
 
   /** callback to fire when the row receives a keyboard event */
   readonly onRowKeyDown: (index: number, e: React.KeyboardEvent<any>) => void
+
+  /**
+   * Whether or not this list row is going to be selectable either through
+   * keyboard navigation, pointer clicks, or both. This is used to determine
+   * whether or not to present a hover state for the list row.
+   */
+  readonly selectable: boolean
 }
 
 export class ListRow extends React.Component<IListRowProps, {}> {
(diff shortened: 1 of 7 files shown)


ampinsk:
@billygriffin Using the hand pointer felt a little too webby to me, but I don't feel super strongly about it. 😄

billygriffin:
@ampinsk That seems fine. I have a personal preference for the hand pointer, but I just tested 7 different apps on macOS and 4 of 7 used the hand pointer and 3 of 7 used the regular cursor. I think I just like it because it makes it obvious/explicit when you're over something clickable, but I also don't feel strongly so I will defer to your judgment. 😄

j-f1:
For the headers, you could set `pointer-events: none` on them to prevent any interaction.

ampinsk:
@niik and I paired on fixing this! Should be ready to go now 👍

tierninho:
Changes look good. Overkill to add hover here?

screen shot 2018-11-30 at 10 21 11 am

ampinsk:
@tierninho hmm the hover state should be already working there!

tierninho:
@ampinsk Yep, you are right! Not sure why it didn't show for me, but now it works fine.

outofambit:
merging this early so we can have it in for the beta

hdsantos89:
Thanks! So happy to see it on the changes readme!

Issue #6213, opened 2018-11-16
Cannot see all conflicts


decabeza:
**Please describe the problem you think should be solved**
When you merge a branch with conflicts in previous version you could check all files with conflicts. Now appears a no closable popup with the list of all conflicted files and you have to open one by one on text editor instead of see them on Github Desktop. 

Example:

![conflicts](https://user-images.githubusercontent.com/631897/48616935-5299a500-e995-11e8-96f7-aa5315be5ade.png)

Also when you resolve all the conflicts you can't review all changed files before merge 😅 

<img width="539" alt="merge" src="https://user-images.githubusercontent.com/631897/48623080-fe98bb80-e9a8-11e8-901d-f53ef2878de7.png">

<img width="534" alt="merge" src="https://user-images.githubusercontent.com/631897/48624105-dfe7f400-e9ab-11e8-88e5-cb6b9b77b85b.png">

**Do you have any potential solutions in mind?**
I think this popup should have a close button to hide it an can check all files (as before) without have to open one by one.

j-f1:
The dialog could be dismissed into a banner saying “X conflicts remaining. **Resolve →**” that could be clicked to reopen the dialog.

billygriffin:
@decabeza We're discussing your issue internally and I wanted to clarify one additional thing. Are the first and second screenshots in the same flow? In other words, after you resolved the conflicts from the first screenshot, did you then see the second screenshot with only one file displayed?

webmayak:
Subcribed. New merge flow if not comfortable

manaskulkarni786:
New merge flow is annoying. I dont have any of the supported merge editors. So I do it through TortoiseMerge using git mergetool from command line. This adds a .ORIG file and I am forced to commit it since I cant review the changes

chrispday:
Agree that new merge flow has issues.  If anything goes wrong with the merge and can't commit there doesn't seem to be an option to close and manually fix yourself.   Only option is to abort which rolls back all merges done so far. Agree with @j-f1 that there should at least be an option to just close the dialog and leave things as is.

decabeza:
Hi @billygriffin,

I understand the benefits to see all conflicted files. The main problem I see is for example if you merge two branches there is two types of files: the conflicted ones and other files, I would like can review all files to check the merge is correct.

[...]

PR #6321, opened 2018-11-30
Block committing conflicted files in sidebar

outofambit:
## Overview

Part of #6213 

## Description

- add a helper to find conflicted files in a WorkingDirectoryStatus
- use it in CommitMessage component

## Release notes

Notes: Prevents users from accidentally committing unresolved merge conflicts

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -23,7 +23,6 @@ import { IAuthor } from '../../models/author'
 import { IMenuItem } from '../../lib/menu-item'
 import { shallowEquals } from '../../lib/equality'
 import { ICommitContext } from '../../models/commit'
-import { hasConflictedFiles } from '../../lib/status'
 
 const addAuthorIcon = new OcticonSymbol(
   12,


app/src/lib/status.ts
```diff
@@ -38,3 +39,9 @@ export function isConflictedFile(
 ): file is ConflictedFileStatus {
   return file.kind === AppFileStatusKind.Conflicted
 }
+
+export function anyConflictedFiles(
```
j-f1: `hasConflictedFiles`?

app/src/ui/changes/commit-message.tsx
```diff
@@ -489,6 +498,7 @@ export class CommitMessage extends React.Component<
           className="commit-button"
           onClick={this.onSubmit}
           disabled={!buttonEnabled}
+          tooltip={this.commitButtonTooltipText()}
```
j-f1: Why not inline the ternary?

outofambit: i prefer to avoid ternaries inline in react component props for readability even though it adds another method

app/src/ui/changes/commit-message.tsx
```diff
@@ -212,11 +214,18 @@ export class CommitMessage extends React.Component<
...
-      (this.props.anyFilesSelected && this.state.summary.length > 0) ||
-      this.props.singleFileCommit
+      !this.props.anyConflictedFiles &&
+      ((this.props.anyFilesSelected && this.state.summary.length > 0) ||
+        this.props.singleFileCommit)
     )
   }
 
+  private commitButtonTooltipText() {
+    return hasConflictedFiles
```
j-f1: Since this is a function, wouldn’t this always be truthy?

outofambit: good catch!!

PR #6336, opened 2018-12-03
helper method for checking for conflicted files in a working directory status

outofambit:
the part of #6321 we can use with the updated designs for #6213

### DIFF
--- a/app/src/lib/status.ts
+++ b/app/src/lib/status.ts
@@ -40,6 +40,11 @@ export function isConflictedFile(
   return file.kind === AppFileStatusKind.Conflicted
 }
 
+/**
+ * Returns a value indicating whether any of the files in the
+ * working directory is in a conflicted state. See `isConflictedFile`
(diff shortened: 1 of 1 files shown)


app/src/lib/status.ts
```diff
@@ -38,3 +39,9 @@ export function isConflictedFile(
 ): file is ConflictedFileStatus {
   return file.kind === AppFileStatusKind.Conflicted
 }
+
+export function hasConflictedFiles(
```
niik: Sorry for being difficult but seeing as this is an exported utility function could we please add a little bit of documentation here? It can be something simple like

```suggestion
/**
 * Returns a value indicating whether any of the files in the
 * working directory is in a conflicted state. See `isConflictedFile`
 * for the definition of a conflicted file.
 */
export function hasConflictedFiles(
```

outofambit: totally reasonable. thank you!

Issue #6294, opened 2018-11-27
Check for existing README prior to creation


joshhopkins:
When adding an existing local project, choosing to create a new repository, the UI displays an option to `Initialise this repository with a README`.

This option was selected by default and in a tired state I selected 'create'. This completely wiped the README.MD file I had spent an hour writing and there doesn't seem to be a way of recovering this.

**Potential solution**

Check for existing README.MD file and prompt user if one exists.

msftrncs:
@iAmWillShepherd, I know it is common for me to create a folder, and start creating files, before I know I am going to turn it into a GIT repository, so when I do decide to use GIT, I need to create the repository from the files I already have, sometimes including a README.MD.  Maybe its a bad practice, but its the work flow I use over and over again.  I have definitely made the same mistake as @joshhopkins , but fortunately for me, I still had the file open in VS Code in an unsaved state, which then caused the message 'file on disk has changed' in VS Code.

joshhopkins:
@iAmWillShepherd I should clarify, in this case no repo previously existed, as @msftrncs pointed out. It may be an edge case, but there are times where creating a repo before writing the first line of code seems overkill. Often the README.MD file starts off as a scratch pad for me, and a repo is created at the point where some value is achieved. As a user, I would expect a good Git client to throw a warning before overwriting pre-existing files.

billygriffin:
Thanks @msftrncs and @joshhopkins for describing the workflow here. I'm wondering if it's something similar to the warning on this image, but instead it's a warning below the README checkbox and says something like: "It looks like you already have a README.md file in this folder. Checking this box will overwrite your existing README.md file with an empty one."

<img width="415" alt="screen shot 2018-11-28 at 8 24 50 am" src="https://user-images.githubusercontent.com/5091167/49161935-19432c80-f2e7-11e8-95f6-58b93b631860.png">

I don't know what detection of that file looks like from a technical perspective, but if it's fairly lightweight, it seems reasonable that we could support a warning like that. Thoughts? Either of you have interest in looking into how the app might detect whether someone has a pre-existing README?

Daniel-McCarthy:
@billygriffin It should be possible to use `fse` to call a `readDir` on the user's given local path, which will return a list of the file names in the folder (if any). From there, it would be possible to check for one named "ReadMe.md". There's also `fse.pathExists` as well for a specific file name.

[...]

PR #6338, opened 2018-12-04
Display warning if existing readme file will be overwritten

Daniel-McCarthy:
## Overview

**Addresses #6294**
(Does not close the issue since this is the initial change where more may follow.)

## Description

- Adds a warning to the `Create-Repository` dialog when a `ReadMe.md` file exists in the new repository directory that will be overwritten. (When Initialize repository with a ReadMe checkbox is set)

If a user already had a `ReadMe.md` file in the directory they are creating a repository in and had set it to initialize with a readme, there was no notice that the existing readme file will be overwritten, potentially causing a loss of work.

image

Edit: Ready for review.

## Release notes

Notes: A warning is now displayed if a readme file exists and will be overwritten during repository creation.

### DIFF
--- a/app/src/ui/add-repository/create-repository.tsx
+++ b/app/src/ui/add-repository/create-repository.tsx
@@ -133,6 +133,12 @@ export class CreateRepository extends React.Component<
       this.state.name
     )
     this.setState({ readMeExists })
+
+    window.addEventListener('focus', this.onWindowFocus)
+  }
+
+  public componentWillUnmount() {
+    window.removeEventListener('focus', this.onWindowFocus)
   }
 
   private onPathChanged = async (path: string) => {
@@ -582,5 +588,15 @@ export class CreateRepository extends React.Component<
         </DialogFooter>
       </Dialog>
     )
+  }
+
+  private onWindowFocus = async () => {
+    // Verify whether or not a ReadMe.md file exists at the chosen directory
+    // in case one has been added or removed and the warning can be displayed.
+    const readMeExists = await this.doesReadMeExist(
+      this.state.path,
+      this.state.name
+    )
+    this.setState({ readMeExists })
   }
 }
--- a/app/src/ui/add-repository/create-repository.tsx
+++ b/app/src/ui/add-repository/create-repository.tsx
@@ -28,6 +28,7 @@ import { Dialog, DialogContent, DialogFooter, DialogError } from '../dialog'
 import { Octicon, OcticonSymbol } from '../octicons'
 import { LinkButton } from '../lib/link-button'
 import { PopupType } from '../../models/popup'
+import { Ref } from '../lib/ref'
 
 /** The sentinel value used to indicate no gitignore should be used. */
 const NoGitIgnoreValue = 'None'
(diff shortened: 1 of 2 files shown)


app/src/ui/add-repository/create-repository.tsx
```diff
@@ -434,6 +463,25 @@ export class CreateRepository extends React.Component<
...
+      this.state.readMeExists === false
+    ) {
+      return null
+    }
+
+    return (
+      <Row className="warning-helper-text">
+        <Octicon symbol={OcticonSymbol.alert} />
+        <p>
+          This directory contains a <Ref>ReadMe.md</Ref> file already. Checking
```
niik: Seeing as we're using `README` fully capitalized in the label could we make this be `README.md` as well?

```suggestion
          This directory contains a <Ref>README.md</Ref> file already. Checking
```

That would also be in line with what we're using on GitHub.com

image

app/src/ui/add-repository/create-repository.tsx
```diff
@@ -76,6 +77,13 @@ interface ICreateRepositoryState {
 
   /** The license to include in the repository. */
   readonly license: string
+
+  /**
+   * Whether or not a ReadMe.md file already exists in the
+   * directory that may be overwritten by initializing with
+   * a new ReadMe.md
```
niik: See comment below
```suggestion
   * a new README.md
```

app/src/ui/add-repository/create-repository.tsx
```diff
@@ -149,6 +173,11 @@ export class CreateRepository extends React.Component<
     this.setState({ isRepository, path })
   }
 
+  private async doesReadMeExist(path: string, name: string) {
```
niik: Seeing how this method doesn't use any of the local state in the class I think we should either move it outside of the class and turn it into a stateless function or (and I think this would be more in line with what we've done elsewhere) rename it to `updateReadmeExists` and have the method itself call `setState`. Multiple calls to `setState` won't trigger individual updates so there shouldn't be a performance difference.

Daniel-McCarthy: Sounds great, @niik! I will make both changes, and for this specific one, I'll give the `updateReadmeExists` suggestion a try. Thanks for the feedback!

tierninho:

- [x] warning shown if ReadMe already exists and box is checked
- [x] removing the ReadMe file midstream updates the UI
- [x] Existing ReadMe is overwritten with fresh ReadMe if checked
- [x] Existing ReadMe is not overwritten if unchecked
- [x] checkbox is disabled by default
- [x] checkbox toggles the warning message
- [x] only affects the creation of new repos
- [x] only affects ReadMe.md files, not other file types
- [x] publishing the new repo does not affect the prior ReadMe decision (GitHub tested, not Enterprise)

Build/OS: 1.5.1-beta1, mac, 433e896de87f694b948573cc14868a5cdc1d913e

outofambit:
hey @Daniel-McCarthy just a heads up—i'm going to add a feature flag to this momentarily and then merge so we can have it in the next 1.5.1 beta. thanks again for all your work on this one!

PR #6340, opened 2018-12-04
fixed branch icon flashes after the branch being renamed

### LINKED ISSUE #6320: Branch icon flashes after branch is renamed
## Description
Spotted this visual issue while reviewing #6305 

Notice the icon next to the renamed branch changes from the _selected_ icon to the _branch_ icon and finally back to the _selected_ icon.
issue

## Steps to Reproduce
1. Select a branch to rename
1. Rename it

### Expected Behavior
I expected the branch to be renamed without the state of the icon changing

### Actual Behavior
What happens is the icons flashes states

gnehcc:
## Overview

**Closes #6320**

## Description

- fixed branch icon flashes after the branch been renamed.


gnehcc:
Behaviors after this fix:
branch renaming

the pr is ready for review from my side, not sure if this solution is too over though...

app/src/lib/stores/git-store.ts
```diff
@@ -910,8 +910,6 @@ export class GitStore extends BaseStore {
       this._tip = { kind: TipState.Unknown }
     }
 
-    this.emitUpdate()
```
niik: This change makes me nervous. I don't think we can clearly foresee the consequences of not emitting here. This method is from several places in the code base and I'd rather not alter its behaviour.

app/styles/ui/_dialog.scss
```diff
@@ -156,6 +157,7 @@ dialog {
         outline: 0;
       }
     }
+
```
niik: Can we exclude the white space changes from these scss files or is this some new prettier rule that's come into effect?

app/src/ui/rename-branch/rename-branch-dialog.tsx
```diff
@@ -31,10 +33,13 @@ export class RenameBranch extends React.Component<
   public constructor(props: IRenameBranchProps) {
     super(props)
 
-    this.state = { newName: props.branch.name }
+    this.state = { newName: props.branch.name, renaming: false }
   }
 
   public render() {
+    const icon = this.state.renaming ? (
```
niik: Rather than including an icon in the button here we should use the new state that you've added and set the `loading` prop of the `<Dialog>` component instead. See https://github.com/desktop/desktop/blob/master/app/src/ui/add-repository/create-repository.tsx#L459 for an example.

Setting the loading prop of the Dialog will automatically disable all buttons in the dialog and show a spinner in the dialog header. By having a uniform way of signalling that a dialog is busy we can more easily change the behavior in the future should we wish to do so.

app/src/ui/rename-branch/rename-branch-dialog.tsx
```diff
@@ -22,6 +23,7 @@ interface IRenameBranchProps {
 
 interface IRenameBranchState {
   readonly newName: string
+  readonly renaming: boolean
```
niik: Could we add a little piece of documentation to this state property? Perhaps something like

```suggestion
  /**
   * Whether or not the component is currently processing the rename.
   * When this is set the dialog will be considered busy and the dialog
   * buttons will be disabled.
   */
  readonly renaming: boolean
```

app/src/lib/stores/git-store.ts
gnehcc: I had a hesitation as well while deleting this line of code, I will check if there is another way...

app/styles/ui/_dialog.scss
gnehcc: Not sure, the same result after reformating in VS Code, kind of thinking these new added line breaks are correct, cause they are separate different selectors?
I can manually add them back if you think the old format is ok.

gnehcc:
Latest behavior:

gnehcc:
Hi @iAmWillShepherd, I made another approach by adding an extra new isRenaming state to avoid unnecessary rendering while renaming a branch, not sure this is a proper solution...

billygriffin:
Thanks so much @gnehcwu for all the work you've done here. After taking another look at it and given the amount of logic required to address a very limited and non-blocking update, I don't think the benefit outweighs the potential cost. I'm inclined to suggest that we close this PR and the corresponding issue. Anyone have any objection to that?

Thanks again for all the work you've put in here, and sorry it dragged on for so long.

gnehcwu:
Hey @billygriffin, no problem, and thanks for commenting under the thread, I'm totally fine with closing this pr.

PR #6343, opened 2018-12-04
consolidate file counting logic in merge conflicts dialog

### LINKED ISSUE #6342: All conflicts resolved but still cannot commit
From @msftrncs in #6213 (edited to clarify what the bug is):

## Description
I've resolved all conflicts but it still says "2 conflicted files" and won't let me commit. (NOTE: this is only on beta version. The current production version 1.5.0 does not exhibit this discrepancy.)

## Version
* GitHub Desktop: 1.5.1-beta0 (special note, Node 10)
* Operating system: Win 10 1809

## Steps to Reproduce
1. merge which produces conflicts
1. use editor to handle the conflicts, saving file when complete
1. repeat editing until all conflicted files are 'No conflicts remaining'.

### Expected Behavior
'Commit merge' button would become available.

### Actual Behavior
'Commit merge' button remains unavailable.

## Additional Information
My editor is VS Code, so in VS Code I can just 'stage changes' on the conflicted file, and then I can merge, but that workflow is not clear to me from the dialog, nor would it be available if I was using an editor without GIT integration.

This build is built on Node 10.

image

So the core issue would seem to be a lack of communication on the next step, or providing the tools for the next step, which is to stage the files so they are no longer marked conflicted so we can just press the 'Commit merge' button and move on.

However, Kudos to Desktop, VS Code would not let me commit with conflicted files, even though I did not want to merge the conflicted files. Desktop handled it fine.

outofambit:
## Overview

fixes #6342

## Description

- introduces a couple helper functions to prevent repeating logic (and getting it wrong in some places)

### DIFF
--- a/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
+++ b/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
@@ -47,6 +47,27 @@ function getUnmergedFiles(status: WorkingDirectoryStatus) {
   return status.files.filter(f => isConflictedFile(f.status))
 }
 
+/** Filter working directory changes for resolved files  */
+function getResolvedFiles(status: WorkingDirectoryStatus) {
+  return status.files.filter(
+    f =>
+      isConflictedFile(f.status) &&
+      f.status.lookForConflictMarkers &&
+      f.status.conflictMarkerCount === 0
+  )
+}
+
+/** Filter working directory changes for conflicted files  */
+function getConflictedFiles(status: WorkingDirectoryStatus) {
+  return status.files.filter(
+    f =>
+      (isConflictedFile(f.status) && !f.status.lookForConflictMarkers) ||
+      (isConflictedFile(f.status) &&
+        f.status.lookForConflictMarkers &&
+        f.status.conflictMarkerCount > 0)
+  )
+}
+
 function editorButtonString(editorName: string | null): string {
   const defaultEditorString = 'editor'
   return `Open in ${editorName || defaultEditorString}`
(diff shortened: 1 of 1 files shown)


outofambit:
@tierninho thank you! the merge conflict warning is a separate (and pre-existing) problem. i think the console error is also pre-existing but i'll double check.

tierninho:
@outofambit Ok, then I successfully merged conflicts from multiple repos/branches with both an editor and the cmd line. LGTM!

msftrncs:
I was able to build this branch and test and verify that it fixed the issue I was having with builds of master of 1.5.1-beta0 and later.  I repeated the same merge conflict scenario on each of 1.5.0, 1.5.1-beta0, and this branch, showing that this branch returned the functionality of the merge conflict model back to the level of 1.5.0 (able to complete merge).

app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -103,12 +124,8 @@ export class MergeConflictsDialog extends React.Component<
...
    */
   private onCancel = async () => {
-    const anyResolvedFiles = getUnmergedFiles(this.props.workingDirectory).some(
-      f =>
-        isConflictedFile(f.status) &&
-        f.status.lookForConflictMarkers &&
-        f.status.conflictMarkerCount === 0
-    )
+    const anyResolvedFiles =
+      getResolvedFiles(this.props.workingDirectory).length > 0
```
niik: Tiny nitpick, seeing as we're only using this to check if any files are resolved we can rename the `getResolvedFiles` method to `anyResolvedFiles` and change the `.filter` to a `.some` rather than iterating through all of them. That said this is tremendously unimportant for this PR but if we're iterating here in the future I think that would be a good change.

PR #6357, opened 2018-12-06
Fix for "directory not empty" error when no repository selected

niik:
## Overview

When filtering until there are no more results in the clone dialog you'll get an error stating that the directory isn't empty even though no repository is selected and the path points to your default clone directory.

image

This bug has existed for a while but was only made visible more recently when #5931 was merged due to us now receiving updates whenever the selected item changes, not just when a list item is clicked.

## Description

The `updateUrl` function is called by the tab components whenever the selected item changes. It parses the url, figures out what the clone path should be, and then calls `updateAndValidatePath` to persist that path and run some validation on it. The problem here is that `updateAndValidatePath` requires access to the `url` state property to function, the same property that has just been updated through `setState` by the `updateUrl` function. `setState` is not synchronous but rather buffers and coalesces changes until the next appropriate render cycle so when the `updateAndValidatePath` tried to access the `url` state variable it was seeing the value as it was prior to the `setState` call in `updateUrl`.

This PR fixes the problem by having all callsites which update `url` also update the path variable and then use the callback functionality of `setState` to invoke the new (renamed) `validatePath` function _once the state change has been finalised_.

## Release notes

no-notes (the bug only exists on beta, never hit production).

### DIFF
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -24,9 +24,6 @@ import { CloneGithubRepository } from './clone-github-repository'
 import { assertNever } from '../../lib/fatal-error'
 import { CallToAction } from '../lib/call-to-action'
 
-/** The name for the error when the destination already exists. */
-const DestinationExistsErrorName = 'DestinationExistsError'
-
 interface ICloneRepositoryProps {
   readonly dispatcher: Dispatcher
   readonly onDismissed: () => void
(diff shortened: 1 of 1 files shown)


say25:
@Daniel-McCarthy

say25:
Thanks for reviewing Daniel, I knew you were recently in this area.

OunmanNgampeerapong:
Thankyou

PR #6365, opened 2018-12-07
Onboarding improvements: The Blank Slate

niik:
## Overview

This PR marks the first user-facing changes to the on-boarding project (see https://github.com/desktop/desktop/issues/5686) and focuses on the "blank slate", i.e. the view users are presented with when they've completed the welcome wizard (or when they don't have any repositories in the app).

### Before

image

### After

image

This view builds on the user research and metrics collected via (#5616) which showed us clearly that the most common action users take after completing the welcome wizard is cloning a repository. As such, this flow has been heavily optimized for that flow while getting rid of distractions.

Building on #6278 the repository list starts loading the moment the user signs in through the welcome wizard so for the majority of our users the repository list should be all loaded and ready when they get to the blankslate.

It's technically possible (although inconvenient and tricky) for a user to sign in to both a GitHub.com and a GitHub Enterprise account while in the welcome wizard so the dialog supports that as well, showing a tab bar to switch between the repositories when using two accounts

image

The most likely scenario where you'd end up with that however is if you've been using the app for a bit, signed in to two accounts, and then remove all the repositories from the app.

## Testing

Easiest way to test this out is to sign out of all accounts, remove all repositories in the app, and then force the welcome flow by deleting the `has-shown-welcome-flow` key in localStorage (through dev tools or by running `localStorage.removeItem('has-shown-welcome-flow')` in the devtools console.

image

## Release notes

Improved: A new, streamlined, flow for quickly adding your first repository to the application.

## TODO

- [ ] Verify design on Windows
- [ ] Documentation, and cleanup

cc @billygriffin @tierninho


tierninho:

- [x] New onboarding page not visible if repos exist
- [x] New onboarding page not visible if repos exist and just signing back into Desktop
- [x] Repo list auto-loads upon landing on new page if signed in; can see indicator + messaging
- [x] Filter and refresh button work; esc button; null results
- [x] Repo list matches cmd-T list; hovering; can be scrolled; up/down arrows
- [x] Selecting Repo in list actives clone modal
- [ ] ❌Internet error shown if connection lost (see https://github.com/desktop/desktop/pull/6297)
- [ ] GitHub and Enterprise tabs shown if user logged into both (cannot test right now)
- [x] Clone/Create/Add Existing buttons map to right actions
- [x] Drag n drop works
- [x] Deleting your last Repo shows onboarding page
- [x] Zoom does not distort
- [x] Other main menu items not working; greyed-out
- [x] Sign-in/out midstream via Prefs is recognized
- [x] Dark mode

**Notes/Issues:**
- What is the state if user does not initially sign-in? Currently the repo list is removed as expected. Not sure if a filler image/text should be here.
- The order of the large repo buttons is Clone, Create, Existing, yet in the Repo menu it is different, and in the `Add v` button it is also different. 
- If user is not signed in, and the clicks the Clone option, we end up showing “fatal: repository ‘username/repo’ does not exist” if they select the URL tab and enter a repo. Perhaps we need to indicate that they are still not signed in before clicking the Clone option? A new Desktop user may get confused...

Build/OS: 1.5.1-beta1, mac, 8c23520ad1408cbb92825d98dd66bf24e9fdaab2

niik:
This is ready for review :bow:

tierninho:
Tested again against the latest commits, and LGTM overall. 

One observance which I tribute to the lack of authentication: 

When I logout of .com, I am automatically logged out of Enterprise too. Since I copied the user key over from prod to the dev environment, I appear to be logged in but still get `warn: [ui] fetchRepositories: Error: Must authenticate to access this API` in the logs as expected.

As soon as I logged out, both tabs disappeared. I assume this is correct because of above.
screen shot 2018-12-19 at 10 31 13 am

1.5.1-beta6, mac, 1c49d27c2fce2a660aa08c56bf00b4645f4d7909

shiftkey:
I was able to trigger some stale state on 1c49d27c2fce2a660aa08c56bf00b4645f4d7909 by running through these steps (apologies for there being _so many steps_ to trip it):

 - nuke all state, see welcome wizard
 - skip account signin, complete rest of wizard, view blank slate
 - click "Clone a Repository from the Internet"
 - no accounts found, dialog asks you to sign in
 - sign in to account, return to blank slate, loading repositories occurs
 - after loading repositories done, choose a repository, **but don't start the clone**
 - click "File | Preferences" menu and sign out of account -> blank slate now hides repositories list
 - click "Clone a Repository from the Internet" again
 - no accounts found,  dialog asks you to sign in
 - sign in to account, return to blank slate
 - repositories loading starts again **but my selection from earlier is there and I can clone it?**



Ideally I think the selection should be cleared when the blank slate reverts, but I haven't had a chance to spot where in the code that should occur.

app/styles/ui/_blank-slate.scss
```diff
@@ -1,105 +1,171 @@
...
+    right: var(--spacing);
+    bottom: var(--spacing);
 
-    display: flex;
-    flex-direction: column;
-    flex-shrink: 0;
+    width: auto;
+    max-width: 40%;
+    height: 40%;
+    // I hate this but we'll have to live with it for beta.
```
shiftkey: What's the TL;DR on this hack? Is it worth capturing as an issue?

niik: This is duplicated from the welcome styles. The hack is that we're using z-index which I don't like but I don't think it's worth worrying about. I can remove the comment from here

app/src/ui/blank-slate/blank-slate.tsx
```diff
@@ -1,81 +1,355 @@
...
+
+  public componentDidMount() {
+    this.ensureRepositoriesForAccount(this.getSelectedAccount())
+  }
+
+  public componentDidUpdate(prevProps: IBlankSlateProps) {
+    this.ensureRepositoriesForAccount(this.getSelectedAccount())
+  }
+
+  private ensureRepositoriesForAccount(account: Account | null) {
```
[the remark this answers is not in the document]
niik: Given that this is a private method I'm not too fuzzed about this one.

niik:
> I was able to trigger some stale state on [1c49d27](https://github.com/desktop/desktop/commit/1c49d27c2fce2a660aa08c56bf00b4645f4d7909) by running through these steps (apologies for there being _so many steps_ to trip it):

I think this is a limitation in the FilterList where it doesn't signal that the selection isn't valid. Given that it's not trivial to get into this state I think we should land this PR and open an issue to look at that separately.

shiftkey:
I'll merge this in to not hold up this PR, but I wanted to point out a couple of dark theme things to follow up:

1. the icon is hard to contrast against the light blue background of the speech bubble
2. the text and stars inside this box have similar contrast - can we make the background less transparent?



I'll move it out to an issue afterwards if it's worth tracking

Issue #6418, opened 2018-12-14
Provide dev-only helper functionality to toggle the release notes popup


koralcem:
**Please describe the problem you think should be solved**

I wanted to prepare a change to address https://github.com/desktop/desktop/issues/5605. I realized very quickly that to be able to see the effects of my changes, I somehow need to show the release notes popup. There doesn't seem to be an easy way to do so. Suggestions in that issue were to use a unique build with an update pending, or temporarily hardcode some values. 

While I imagine those methods would trigger the popup just fine, they would have to be given to every other person working on that popup in the future.

**Do you have any potential solutions in mind?**

I noticed the "Help" menu already houses some dev-only functionality to aid in verification ("Crash main process...", "Crash renderer process..."). I suggest adding an entry there to trigger popups. This specific issue is written from the perspective of the release notes popup, but it could be expanded in the future to be able trigger any popup.

tierninho:
@koralcem Many thanks for adding the associated [PR](https://github.com/desktop/desktop/pull/6419) and congrats on contributing your first PR too 💛  

As our team focused on getting next week’s release out and with the holidays approaching, we may not be able to get to a review until after the break. 

We will circle back soon, so hold tight. ⚡️

koralcem:
That's all good; no rush. I appreciate the explanation rather than weeks of silence.

Issue #5605, opened 2018-09-11
Link desktop marketing site back to in-app release notes


nerdneha:
Ref: [Comment made on #3601](https://github.com/desktop/desktop/pull/3601#issuecomment-418841199)

With the new [in-app release notes](https://github.com/desktop/desktop/pull/3601), we've replaced the `what's new` link on the banner a link to the [marketing page release notes](https://desktop.github.com/release-notes/) to now show the release notes summary pop-up. 
<img width="1213" src="https://user-images.githubusercontent.com/359239/45357862-0a799300-b59e-11e8-807a-a407023982ae.png">

### Please describe the problem you think should be solved
As @tierninho mentioned in [his comment](https://github.com/desktop/desktop/pull/3601#issuecomment-418815465), if a user is upgrading from v1.2 to v1.4, they would only see the release notes for v1.4 and miss v1.3's release notes. That used to exist when the `what's new` copy linked to the [release notes for the marketing page](https://desktop.github.com/release-notes/) but that is no longer being pointed to. 

### Do you have any potential solutions in mind?
[We agreed](https://github.com/desktop/desktop/pull/3601#issuecomment-418819082) that it would be too overwhelming to have multiple releases' notes in the in-app popup.

I could see us linking the marketing page's release notes in the in-app release notes. FWIW I don't have strong feelings about copy or location but here is what I've brainstormed:
For copy: maybe `(all release notes)` or `View past release notes`? 
For location: It doesn't seem important enough to be a button but maybe we can have it under the version or date?

<img width="400" alt="44813675-c3100180-abb0-11e8-9218-016d8d0e66d7" src="https://user-images.githubusercontent.com/1834387/45382021-df9e3780-b5bb-11e8-8ba7-083fe3adcdf9.png">

For reference, this is what we do in our `About GitHub Desktop` dialogue:
<img width="472" alt="screen shot 2018-09-11 at 12 13 02 pm" src="https://user-images.githubusercontent.com/1834387/45382103-212ee280-b5bc-11e8-9c20-77524f8c3cae.png">

billygriffin:
Thanks @nerdneha! Let's go with "View all release notes" as a link below the date. I added `help wanted` and `good first issue` as I think this is sufficiently scoped for a first-time contributor.

billygriffin:
Good question @koralcem! @desktop/quality-assurance Do you happen to know the answer to that?

tierninho:
[...]

Issue #5510, opened 2018-08-29
display a "pretext" header inside the in-app release notes


shiftkey:
As part of #2774 we proposed support for adding some "pretext" for the latest release. The mockup looked like this:

![](https://user-images.githubusercontent.com/634063/30918386-b01bc34a-a39e-11e7-86d5-2281af62076b.png)

To surface this, we talked about this workflow:

 - in the release notes, add a `[pretext]` entry that contains some text
 - if defined for a release, the GitHub Desktop website should display this entry separate to the other release notes (cc @desktop/design for thoughts on how this might look)
 - if defined for a release, the app should display this entry separate to the other release notes

koralcem:
Has anyone thought any further about how the dialog should look when the release notes contain `pretext`? Would we switch to a single column layout and add the pretext at the beginning, like the mock given above?

PR #6419, opened 2018-12-14
Add a dev-only menu item to show the release notes popup

koralcem:
## Overview

**Closes https://github.com/desktop/desktop/issues/6418**

## Description

The addition is meant to facilitate working on the release notes popup easier. When modifying the popup, one can see the effect of their changes on a popup for a fake release, rather than trying to conjure up an actual pending update via getting a special build, or temporarily hardcoding a one-off new release in their own environment and remembering to leave that out of their commits.

With this addition, the menu looks like this:
menuadd

And clicking it gets you this:
fakeupdate

## Release notes

Notes: Adds a dev-only menu item to show the release notes popup

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -415,6 +415,15 @@ export function buildDefaultMenu({
       {
         label: 'Crash renderer process…',
         click: emit('boomtown'),
+      },
+      {
+        label: 'Show popup',
+        submenu: [
+          {
+            label: 'Release notes',
+            click: emit('show-release-notes-popup'),
+          },
+        ],
       }
     )
   }
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -348,9 +348,66 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.openCurrentRepositoryInExternalEditor()
       case 'select-all':
         return this.selectAll()
+      case 'show-release-notes-popup':
+        return this.showFakeReleaseNotesPopup()
     }
 
     return assertNever(name, `Unknown menu event name: ${name}`)
+  }
+
+  /**
+   * Show a release notes popup for a fake release, intended only to
+   * make it easier to verify changes to the popup. Has no meaning
+   * about a new release being available.
+   */
+  private showFakeReleaseNotesPopup() {
+    if (__DEV__) {
+      this.props.dispatcher.showPopup({
+        type: PopupType.ReleaseNotes,
+        newRelease: {
+          latestVersion: '42.7.99',
+          datePublished: 'Awesomeber 71, 2025',
+          pretext:
+            'There is something so different here that we wanted to include some pretext for it',
+          enhancements: [
+            {
+              kind: 'new',
+              message: 'An awesome new feature!',
+            },
+            {
+              kind: 'improved',
+              message: 'This is so much better',
+            },
+            {
+              kind: 'improved',
+              message:
+                'Testing links to profile pages by a mention to @shiftkey',
+            },
+          ],
+          bugfixes: [
+            {
+              kind: 'fixed',
+              message: 'Fixed this one thing',
+            },
+            {
+              kind: 'fixed',
+              message: 'Fixed this thing over here too',
+            },
+            {
+              kind: 'fixed',
+              message:
+                'Testing links to issues by calling out #42. Assuming it is fixed by now.',
+            },
+          ],
+          other: [
+            {
+              kind: 'other',
+              message: 'In other news...',
+            },
+          ],
+        },
+      })
+    }
   }
 
   /**
--- a/app/src/main-process/menu/menu-event.ts
+++ b/app/src/main-process/menu/menu-event.ts
@@ -28,3 +28,4 @@ export type MenuEvent =
   | 'install-cli'
   | 'open-external-editor'
   | 'select-all'
+  | 'show-release-notes-popup'


koralcem:
My first PR, so please point it out if I've missed anything about the process/guidelines.

Wanted something like this to make working on https://github.com/desktop/desktop/issues/5605 easier, but didn't find it. The PR guidelines suggested having a specific issue for each PR, so I opened https://github.com/desktop/desktop/issues/6418 to explain it. Let me know if I've gone overboard there.

I also have a few questions about the code itself, which I'll put among the diff.

app/src/ui/app.tsx
```diff
@@ -348,6 +348,41 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.openCurrentRepositoryInExternalEditor()
       case 'select-all':
         return this.selectAll()
+      case 'show-release-notes-popup':
+        return this.props.dispatcher.showPopup({
```
koralcem: Is it elegant to have the notes for a fake release sitting directly in the `switch` statement here? Shall I move them out to another spot? If so, where is a good place?

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -415,6 +415,15 @@ export function buildDefaultMenu({
       {
         label: 'Crash renderer process…',
         click: emit('boomtown'),
+      },
+      {
+        label: 'Show popup',
+        submenu: [
+          {
+            label: 'Release notes',
+            click: emit('show-release-notes-popup'),
```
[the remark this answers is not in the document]
shiftkey: > Is expanding by adding a new MenuEvent each time appropriate?

I think this is fine for now. If a separate use-case to "supporting development mode testing" in the default menu comes up which might be helped by refactoring `emit` we can revisit this then.

app/src/ui/app.tsx
shiftkey: I'm not quite sure where the best place to put this is, but I'd love for this to be stripped out of the build when we generate the production bits. We use the `__DEV__` global for this so that webpack can then eliminate dead/unreachable code, but I'm not sure how best to do that within a `switch` statement.

koralcem: I can have the `case` block call a helper function (`showFakeReleaseNotesPopup`?), move all this into the implementation of said helper, and wrap the whole body of the helper with `if (__DEV__)`. Like so:
fake

`showFakeReleaseNotesPopup` becomes an empty/no-op function in `prod` builds. Its declaration is still in the package, and it still doesn't convey to a human that it's dev-only functionality until you actually look at its implementation, which isn't ideal. But if the concern is around having the fake release notes in the build, this does accomplish leaving them out.

shiftkey: Yeah, let's go with this approach. For production builds this should be a no-op, as the menu shouldn't be accessible to users, but for development mode this can work as currently designed.

app/src/ui/app.tsx
```diff
@@ -348,6 +348,41 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+          type: PopupType.ReleaseNotes,
+          newRelease: {
+            latestVersion: '42.7.99',
+            datePublished: 'Awesomeber 71, 2025',
+            pretext: 'kas',
+            enhancements: [
+              {
+                kind: 'new',
+                message: 'An awesome new feature!',
```
shiftkey: Can you define some `#id` and `@author-name` mentions in here so that we can test that linking works for issues and authors? The list on https://desktop.github.com/release-notes/ is a good reference if you want some existing data to work with.

koralcem: Good idea. Sprinkled in one of each. Let me know if you'd like to see more, or other variations of anything.

app/src/ui/app.tsx
```diff
@@ -348,11 +348,64 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+   * make it easier to verify changes to the popup. Has no meaning
+   * about a new release being available.
+   */
+  private showFakeReleaseNotesPopup() {
+    this.props.dispatcher.showPopup({
+      type: PopupType.ReleaseNotes,
+      newRelease: {
+        latestVersion: '42.7.99',
+        datePublished: 'Awesomeber 71, 2025',
+        pretext: 'kas',
```
shiftkey: Not going to hold up this PR, but I'll cross-reference #5510 here which has the context about what this field represents.

app/src/ui/app.tsx
```diff
@@ -348,11 +348,64 @@ export class App extends React.Component<IAppProps, IAppState> {
...
     return assertNever(name, `Unknown menu event name: ${name}`)
   }
 
+  /**
+   * Show a release notes popup for a fake release, intended only to
+   * make it easier to verify changes to the popup. Has no meaning
+   * about a new release being available.
+   */
+  private showFakeReleaseNotesPopup() {
+    this.props.dispatcher.showPopup({
```
shiftkey: @koralcem actually, we talked about wrapping this in a `if(__DEV__)` so it would get excluded from the production build. Would you mind adding this in before I merge it?

koralcem: Wow... Good catch @shiftkey, and I'm so sorry I forgot about that. As you said, we just talked about it too 🤦‍♂️ 

Added.

app/src/ui/app.tsx
koralcem: Thanks for the reference; I realize now what the `pretext` field represents.

But can you clarify what you meant by "not going to hold up this PR, but ..."? Was there something about #5510 that you'd ideally like to see addressed here? Even if I tackled that too, it seems to me like they are two distinct pieces of work that should have their own commits/PRs; this one being about showing the dialog, and that one being about the implementation of the dialog.

shiftkey: No, I just wanted to add a link to this PR into #5510 so that someone can easily find this dev-mode option.

PR #6438, opened 2018-12-17
[HOLD] Upgrade to latest Electron v3 release

### LINKED ISSUE #6391: Upgrade from Electron 2.x to Electron 3
Currently, the latest electron version is 3.0.10, but the version desktop used is still 2.0.13. Why don't we use the latest version?

iAmWillShepherd:
Fixes #6391

Puts us on Electron 3.0.12 ([release notes](https://electronjs.org/blog/electron-3-0) for Electron 3) and updates related dependencies. ~~Note that you will have to upgrade your version of node to ` >= v10.14.0`.~~


shiftkey:
Tweaking the title to ensure this isn't confused for Electron v4, which will be available imminently

iAmWillShepherd:
Currently experience an issue where it builds, but ultimately fails to launch. Will need to do a bit of investigating to figure out the issue since something seems to be failing silently.

Once I get this working on my machine, I'll update `electron-winstaller` to `v2.7.0` and test on Windows.

package.json
```diff
@@ -93,19 +93,19 @@
     "mocha": "^5.2.0",
     "mocha-junit-reporter": "^1.18.0",
     "mocha-multi-reporters": "^1.1.7",
-    "node-sass": "^4.7.2",
+    "node-sass": "^4.11.0",
```
shiftkey: We need to be careful about the version of Node that ships in Electron (which will be Node 10.x in Electron 3) versus the Node version used as part of the toolchain (still Node 8 until we resolve #5876). This is reflected in how we split up the `package.json` dependencies:

 - `package.json` - toolchain for building Desktop
 - `app/package.json` - libraries used inside Desktop

My working assumption here is that anything related to Electron 3 should only require changing versions in `app/package.json` (aside from `electron-*` dependencies). Can we confirm we're not trying to upgrade _everything_ to Node 10 as part of this PR?

package.json
```diff
@@ -49,7 +49,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">= 8.11",
+    "node": ">= 10.0",
```
shiftkey: Put this back to `8.11` if you want to test this as part of CI - this is related to the Node version you have installed locally, which is independent of the Node version embedded in Electron.

say25:
Just something to think about, with v3 can we kill some of our polyfills? https://github.com/desktop/desktop/blob/master/app/src/lib/globals.d.ts#L190-L218

shiftkey:
@say25 not really - `MenuItem` and `RequestOptions` still fill gaps in the original type declarations that we use even after upgrading to 3.x, and `SystemPreferences.getUserDefault` always returns `any` - we can narrow this type for our known usage the API. Maybe we can get them upstreamed after we're on 3.x (or check in with 4.x to see whether it's easier to upstream then).

`ResizeObserver` is a browser API not available everywhere, so I don't expect that to be available in the standard type declarations that ship with TypeScript.

shiftkey:
@iAmWillShepherd I've opened #6440 and #6442 as these dependencies can be updated independent of the Electron migration, and will simplify the number of moving parts we need to touch as part of this PR.

shiftkey:
After chatting today about the Electron and how native modules all interop, I think a better path would be to land #6440 and #6441 (which will work with Electron 2 and 3) before focusing on the upgrade to Electron. Let me know what your thoughts are on this.

shiftkey:
@iAmWillShepherd once #6442 has been merged did you want to pair up on revisiting this PR?

shiftkey:
Closing this out in favour of a fresh branch and PR. I think the changes in [`app/src/main-process/main.ts`](https://github.com/desktop/desktop/pull/6438/files#diff-3aec7eb337ae15a9f0428c486641cebf) are the only code changes we need to make now that the dependencies are merged into `development`.

PR #6439, opened 2018-12-17
upgrade to Electron 2.0.16 release

iAmWillShepherd:
Update to upcoming version of electron. I bumped to latest release to make sure no errors decided to visit.

 - [v2.0.14 release notes](https://github.com/electron/electron/releases/tag/v2.0.14)
 - [v2.0.15 release notes](https://github.com/electron/electron/releases/tag/v2.0.15)
 - [v2.0.16 release notes](https://github.com/electron/electron/releases/tag/v2.0.16)

TODO: 

 - [x] test on macOS
 - [x] test on Windows
 - [x] test on Linux


shiftkey:
We were initially keen to get this into the next update but after our own investigation and [the announcement post](https://electronjs.org/blog/magellan-fix) we're going to hold off on rushing this out with 1.5.1 because we don't deal with WebSQL or remote content that might use WebSQL. Instead we'll pull this in with 1.6.0 (which is planned for mid-January).

shiftkey:
After discussing this with the team we've decided to hold off on this PR and instead focus on the Electron 3 upgrade process.

PR #6440, opened 2018-12-17
replace runas with fs-admin to handle installing installing CLI

### LINKED ISSUE #5669: runas is deprecated, fs-admin is the future for doing elevated filesystem things
Looking into our dependencies that are affected by the Electron 3.0 upgrade, `runas` (which was maintained by the Atom team) doesn't compile, and has been obsoleted in favour of `fs-admin`. They've chosen a more specific abstraction for the operations than "just spawn whatever": https://github.com/atom/fs-admin/blob/master/index.js

We're using this in [`install-cli.ts`](https://github.com/desktop/desktop/blob/master/app/src/ui/lib/install-cli.ts) to create a symlink, which should be supported by `fs-admin`:

https://github.com/desktop/desktop/blob/99916aead756eb9db7332da3dac08883b537f88b/app/src/ui/lib/install-cli.ts#L43-L66 

There are also no type declarations for `fs-admin`, and maybe we can use this as chance to help define some.

shiftkey:
## Overview

Fixes #5669 

This PR moves us to use `fs-admin` for doing elevated filesystem operations in the app. `runas` was maintained by the Atom team, but they're no longer using this so I suggest we follow their lead.

Currently we only use this to install the CLI tool for macOS users (opt-in), so it's a narrow surface to test.

## Description

- [x] update `install-cli.ts` to use new module
- [x] test production build adds `.cmd` file to bundle (only required for Windows usage) using webpack loader
- ~~[ ] see if https://github.com/atom/fs-admin/pull/2 will be merged to add type declarations~~
- ~~[ ] upgrade to version with typings built-in and cleanup unnecessary types~~
- [x] testing on macOS
- [x] testing on Windows (does not expose this menu item)
- [x] testing on Linux (does not expose this menu item)

## Release notes

Notes: no-notes


app/webpack.common.ts
```diff
@@ -88,6 +88,10 @@ export const renderer = merge({}, commonConfig, {
         test: /\.(jpe?g|png|gif|ico)$/,
         use: ['file?name=[path][name].[ext]'],
       },
+      {
+        test: /\.cmd$/,
+        loader: 'file-loader',
```
shiftkey: @niik if you've got any better ideas to satisfy webpack for this corner case instead introducing another loader I'd love to hear it, but basically this is where I ended up:

1. `fs-admin` has a Windows-specific `.cmd` script that it uses to `robocopy` directories
2. this is referenced in source like this: `require.resolve('./src/copy-folder.cmd')`
3. webpack complains when it encounters this code

```
[WEBPACK] Errors building renderer.js
Module parse failed: Unexpected token (1:3)
You may need an appropriate loader to handle this file type.
| if exist %2 rmdir %2 /s /q
| (robocopy %1 %2 /e) ^& IF %ERRORLEVEL% LEQ 1 exit 0
|
```

4. I thought I'd use the step above (with the other static files) but that didn't seem to make the webpack error go away
5. [`file-loader`](https://www.npmjs.com/package/file-loader) seemed to do the trick

shiftkey:
I've omitted the typings tasks because they're not a blocker on getting this reviewed and merged

app/src/ui/lib/install-cli.ts
```diff
@@ -1,11 +1,7 @@
-import * as Fs from 'fs'
+import * as FSE from 'fs-extra'
 import * as Path from 'path'
 
-const runas: (
-  command: string,
-  args: ReadonlyArray<string>,
-  options?: { admin: boolean }
-) => number = require('runas')
+import fsAdmin = require('fs-admin')
```
iAmWillShepherd: Are we unable to do TS style import: `import * as fsAdmin from 'fs-admin`?

shiftkey: I think this was just an oversight because the type declarations weren't setup correctly. Fixed in 4fc3a1717401f46971b5967f6b6b807b7f062149

iAmWillShepherd:
Merged w/o letting sit for a day since the last commit only changed the style of import we use.

PR #6443, opened 2018-12-17
Add link to all release notes in the release notes dialog

koralcem:
## Overview

**Closes https://github.com/desktop/desktop/issues/5605**

## Description

Adds a link to all release notes in to the release notes dialog, as discussed in https://github.com/desktop/desktop/issues/5605

## Release notes

Notes:

Adds a link to all release notes in to the release notes dialog

### DIFF
--- a/app/src/ui/release-notes/release-notes-dialog.tsx
+++ b/app/src/ui/release-notes/release-notes-dialog.tsx
@@ -17,6 +17,7 @@ import { RichText } from '../lib/rich-text'
 import { Repository } from '../../models/repository'
 import { getDotComAPIEndpoint } from '../../lib/api'
 import { shell } from '../../lib/app-shell'
+import { ReleaseNotesUri } from '../lib/releases'
 
 // HACK: This is needed because the `Rich`Text` component
 // needs to know what repo to link issues against.
@@ -194,7 +195,6 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
   }
 
   private showAllReleaseNotes = () => {
-    const releaseNotesUri = 'https://desktop.github.com/release-notes/'
-    shell.openExternal(releaseNotesUri)
+    shell.openExternal(ReleaseNotesUri)
   }
 }
--- a/app/src/ui/release-notes/release-notes-dialog.tsx
+++ b/app/src/ui/release-notes/release-notes-dialog.tsx
@@ -8,6 +8,7 @@ import { ReleaseNote, ReleaseSummary } from '../../models/release-notes'
 import { updateStore } from '../lib/update-store'
 import { ButtonGroup } from '../lib/button-group'
 import { Button } from '../lib/button'
+import { LinkButton } from '../lib/link-button'
 
 import { Dialog, DialogContent, DialogFooter } from '../dialog'
 import { DialogHeader } from '../dialog/header'
@@ -15,6 +16,7 @@ import { DialogHeader } from '../dialog/header'
 import { RichText } from '../lib/rich-text'
 import { Repository } from '../../models/repository'
 import { getDotComAPIEndpoint } from '../../lib/api'
+import { shell } from '../../lib/app-shell'
 
 // HACK: This is needed because the `Rich`Text` component
 // needs to know what repo to link issues against.
@@ -173,6 +175,9 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
 
         <DialogContent>{contents}</DialogContent>
         <DialogFooter>
+          <LinkButton onClick={this.showAllReleaseNotes}>
+            View all release notes
+          </LinkButton>
           <ButtonGroup destructive={true}>
             <Button type="submit">Close</Button>
             <Button onClick={this.updateNow}>
@@ -186,5 +191,10 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
 
   private updateNow = () => {
     updateStore.quitAndInstallUpdate()
+  }
+
+  private showAllReleaseNotes = () => {
+    const releaseNotesUri = 'https://desktop.github.com/release-notes/'
+    shell.openExternal(releaseNotesUri)
   }
 }
--- a/app/styles/ui/dialogs/_release-notes.scss
+++ b/app/styles/ui/dialogs/_release-notes.scss
@@ -41,6 +41,11 @@
     }
   }
 
+  .dialog-footer {
+    flex-direction: row;
+    justify-content: space-between;
+  }
+
   ul {
     list-style: none;
     padding-left: 0;
--- a/app/src/ui/about/about.tsx
+++ b/app/src/ui/about/about.tsx
@@ -12,6 +12,7 @@ import { Disposable } from 'event-kit'
 import { Loading } from '../lib/loading'
 import { RelativeTime } from '../relative-time'
 import { assertNever } from '../../lib/fatal-error'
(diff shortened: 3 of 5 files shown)


app/styles/ui/dialogs/_release-notes.scss
```diff
@@ -9,7 +9,7 @@
   }
 
   .dialog-header {
-    height: 100px;
+    height: 110px;
```
koralcem: Without the bump to the `height`, the upper half of that dialog looks a bit "compressed" (for lack of a better term). The margins of the close button don't look right, and one can argue that the link is too close to the divider.
x

Let me know if there's a better way to handle that instead of just bumping the `height`.

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -187,4 +192,9 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
   private updateNow = () => {
     updateStore.quitAndInstallUpdate()
   }
+
+  private showAllReleaseNotes = () => {
+    const releaseNotesUri = 'https://desktop.github.com/release-notes/'
```
koralcem: This constant is already used [here](https://github.com/desktop/desktop/blob/master/app/src/ui/about/about.tsx#L46) and [here](https://github.com/desktop/desktop/blob/master/app/src/ui/banners/update-available.tsx#L51), and I'm adding another use now. Is there a mechanism through which they can be centralized so we don't repeat it?

niik:
Hey @koralcem, thanks so much for opening this, and thank you for including screenshots, that makes things so much easier!

I'm not sold on the idea of the link being in the dialog header. The illustration in that header is tailored for the release version and date to sit in the middle of the illustration and adding a third line messes with the balance in my opinion (in fact, thanks to your screenshot I realized there's too much spacing between the two as it is, the version and date were supposed to be much closer together than that and likely that got missed in the initial implementation).

I think I'd prefer if we put it in the footer where all the other actions are located. I agree with the comment in the originating issue about it not deserving a button but I think we could stick the link in the footer off to the left. Something like this

image

(Please ignore the fonts and inconsistencies, I had to open an old sketch file to find the original mocks and apparently it couldn't resolve the font any more).

I'm gonna ping the rest of @desktop/design to see if they have any clever ideas. We're coming up on the holidays though and people will be out so it's possible this will have to sit for a while until we're back up to full strength again.

koralcem:
@niik Thanks for chiming in. I'll admit I have no preference myself; I simply assumed from [this comment](https://github.com/desktop/desktop/issues/5605#issuecomment-440370111) by @billygriffin (and the lack of objection for 20 days afterward 😄) that under the release date was the spot we were going with.

I realize the holidays slow things down. Happy for this to hang around until afterward and until further discussion.

billygriffin:
Thanks @koralcem and @niik. I'm really sorry @koralcem for the runaround - I should've made sure we were all aligned internally before I gave the 👍 in the issue. I'll try to do better at that going forward.

@niik's proposed design seems great to me, and if you decide to implement that prior to others weighing in, we won't reverse course on you again. 😄 This is a small enough change and an improvement over current state. Thanks for understanding and hope my lack of organization on this one doesn't dissuade you from continuing to contribute further. ❤️

koralcem:
No need to apologize @billygriffin; I realize designs change.

And I realize the holidays make it slower for anyone who might otherwise chime in. As long as you folks don't mind, I'll let this PR simmer until the first week of January. I'll go with whatever the agreed upon experience is at that point, defaulting to @niik's latest suggestion if nothing else comes in.

billygriffin:
That sounds great from my perspective @koralcem, thank you!

koralcem:
Had a spare moment and attempted niik's design. Looks like this:

screencast 2018-12-26 17-29-15

Mainly wanted to see what it would take to do it. I'm totally fine if this PR still waits, and if what I have here doesn't make the final cut. Please don't _not_ iterate because this is already done.

koralcem:
Is the Travis CI build failure due to the change given in the PR? If so, I can't tell how/why.

I force-pushed a commit that implemented the new approach to this branch. Was that somehow a no-no?

Daniel-McCarthy:
@koralcem It appears that the Azure CI and CircleCI both passed while Travis-CI failed at the installation stage. It seems to be more of an intermittent failure that sometimes occurs and not due to your changes.

The maintainers should be able to re-run the tests.

app/src/ui/release-notes/release-notes-dialog.tsx
shiftkey: For things that are shared in multiple places, I'd create a module under `app/src/ui/lib/` to centralize it. If it were shared between the main and renderer processes I'd move it under `app/src/models` but for now this is only specific to the UI itself.

ere's a rough patch to illustrate the changes:

```diff
diff --git a/app/src/ui/about/about.tsx b/app/src/ui/about/about.tsx
index c690c4282c..dd94db4d3c 100644
--- a/app/src/ui/about/about.tsx
+++ b/app/src/ui/about/about.tsx
@@ -12,6 +12,7 @@ import { Disposable } from 'event-kit'
 import { Loading } from '../lib/loading'
 import { RelativeTime } from '../relative-time'
 import { assertNever } from '../../lib/fatal-error'
+import { ReleaseNotesUri } from '../lib/releases'
 
 interface IAboutProps {
   /**
@@ -42,9 +43,6 @@ interface IAboutProps {
 interface IAboutState {
   readonly updateState: IUpdateState
 }
-
-const releaseNotesUri = 'https://desktop.github.com/release-notes/'
-
 /**
  * A dialog that presents information about the
  * running application such as name and version.
@@ -246,7 +244,7 @@ export class About extends React.Component<IAboutProps, IAboutState> {
     const name = this.props.applicationName
     const version = this.props.applicationVersion
     const releaseNotesLink = (
-      <LinkButton uri={releaseNotesUri}>release notes</LinkButton>
+      <LinkButton uri={ReleaseNotesUri}>release notes</LinkButton>
     )
 
     return (
diff --git a/app/src/ui/banners/update-available.tsx b/app/src/ui/banners/update-available.tsx
index 2d734a799b..4001e3717d 100644
--- a/app/src/ui/banners/update-available.tsx
+++ b/app/src/ui/banners/update-available.tsx
@@ -8,6 +8,7 @@ import { shell } from '../../lib/app-shell'
 
 import { ReleaseSummary } from '../../models/release-notes'
 import { Banner } from './banner'
+import { ReleaseNotesUri } from '../lib/releases'
 
 interface IUpdateAvailableProps {
   readonly dispatcher: Dispatcher
@@ -48,8 +49,7 @@ export class UpdateAvailable extends React.Component<
     if (this.props.newRelease == null) {
       // if, for some reason we're not able to render the release notes we
       // should redirect the user to the website so we do _something_
-      const releaseNotesUri = 'https://desktop.github.com/release-notes/'
-      shell.openExternal(releaseNotesUri)
+      shell.openExternal(ReleaseNotesUri)
     } else {
       this.props.dispatcher.showPopup({
         type: PopupType.ReleaseNotes,
diff --git a/app/src/ui/lib/releases.ts b/app/src/ui/lib/releases.ts
new file mode 100644
index 0000000000..c9eb3a5b3b
--- /dev/null
+++ b/app/src/ui/lib/releases.ts
@@ -0,0 +1 @@
+export const ReleaseNotesUri = 'https://desktop.github.com/release-notes/'
```

koralcem: Thanks for the pointer. Added the constant.

Issue #6349, opened 2018-12-05
Working directory changes committed unintentionally during merge conflict resolution


billygriffin:
**Please describe the problem you think should be solved**

If I have changed files in my working directory prior to merging a different branch into mine (for example, updating my branch by merging in the latest from master) and I encounter merge conflicts, resolve the conflicts, and commit the merge, the files that were in my working directory will also be committed as part of the merge commit even though I likely did not intend for them to be.

This is likely confusing ("where did the stuff I was working on go?") and could be a disincentive to frequently staying up to date with the default branch, which we think is a good practice.

**Do you have any potential solutions in mind?**

My first inclination is that Desktop could ignore the working directory files during the merge and put them back after it's completed. I'm not sure what potential pitfalls exist with this.

Note: This was originally brought to my attention by @niik.

cc: @outofambit, @ampinsk, @nerdneha

ampinsk:
I think that makes sense. Are there any workflows that would expect the working directory files to _not_ be ignored?

billygriffin:
Moving this to 1.6.1 - #6464 should solve #6411 but doesn't get us all the way on this one.

PR #6464, opened 2018-12-20
don't include untracked files in merge commit

### LINKED ISSUE #6411: Merge Conflicts cannot proceed when untracked files exist
### Reproduce:
1. Start with a branch that has 1 or more untracked files.
2. Merge another branch, like master, into branch, such that conflicts will arise (should NOT be changes that track above file(s)).
3. Merge Conflict dialog pops up.
4. Resolve merge conflicts in editor (do not remove untracked files).
5. Press "Commit Merge" button. 
6. See error: "Could not generate a patch for file foo/bar, patch empty".
7. Now you are stuck.

### Work arounds:
1. Move (temporarily) the untracked file(s). 
2. Run `git add foo/bar` outside of the app. 

Neither are particularly pleasing options. 

### Preferred Solutions:
1. Quietly skip untracked files for the merge, unless, of course, the merge adds them. Committing anything that didn't come from the merge makes no sense from the merge's perspective. (Preferred)
2. Do (1) and offer options to add or skip untracked. (Ok, but messier merge) 
3. Allow me to exit the dialog and commit when I'm ready, which is sort of how it was before the Merge Conflict dialog was added. (Ok, less messy than (2), but not great)
4. Warn me that the merge cannot be started with untracked files present, which is hardly better than doing nothing. (Meh, I'll just end up do work around 1 anyway)

outofambit:
~~🌵 🌵 **waiting on #6468** 🌵 🌵~~

## Overview

closes #6411

## Description

- refactors component's click handler and `AppStore`/`Dispatcher`'s `createMergeCommit` into more semantic `finishConflictedMerge`
- adds a little logic to prevent committing untracked files during merge conflict resolution
- adds untracked file to conflicted repo fixture (and fix broken tests as a result)
- adds regression test 🎉 

## Release notes

Notes: Leave untracked files uncommitted during merge conflict resolution

### DIFF
--- a/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
+++ b/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
@@ -104,9 +104,14 @@ export class MergeConflictsDialog extends React.Component<
    *  commits the merge displays the repository changes tab and dismisses the modal
    */
   private onSubmit = async () => {
+    // filter out untracked files so we don't commit them
+    const trackedFiles = this.props.workingDirectory.files.filter(f => {
+      return f.status.kind !== AppFileStatusKind.Untracked
+    })
+
     await this.props.dispatcher.createMergeCommit(
       this.props.repository,
-      this.props.workingDirectory.files,
+      trackedFiles,
       {
         ourBranch: this.props.ourBranch,
         theirBranch: this.props.theirBranch,
--- a/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
+++ b/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
@@ -104,14 +104,9 @@ export class MergeConflictsDialog extends React.Component<
    *  commits the merge displays the repository changes tab and dismisses the modal
    */
   private onSubmit = async () => {
-    // filter out untracked files so we don't commit them
-    const trackedFiles = this.props.workingDirectory.files.filter(f => {
-      return f.status.kind !== AppFileStatusKind.Untracked
-    })
-
-    await this.props.dispatcher.createMergeCommit(
+    await this.props.dispatcher.finishConflictedMerge(
       this.props.repository,
-      trackedFiles,
+      this.props.workingDirectory,
       {
         ourBranch: this.props.ourBranch,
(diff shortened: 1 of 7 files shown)


app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
```diff
@@ -104,9 +104,14 @@ export class MergeConflictsDialog extends React.Component<
    *  commits the merge displays the repository changes tab and dismisses the modal
    */
   private onSubmit = async () => {
+    // filter out untracked files so we don't commit them
+    const trackedFiles = this.props.workingDirectory.files.filter(f => {
+      return f.status.kind !== AppFileStatusKind.New
```
shiftkey: Is there a way we can capture this as a regression test? I'm worried that this also catches `'added'` entries from here (which might be valid as part of the merge commit):

https://github.com/desktop/desktop/blob/b556ff8f9746447b7457fdbe78796b835e1db322/app/src/lib/git/status.ts#L104-L113

outofambit: ah good catch! i will adjust to cover that case and add some tests too

billygriffin:
Excited to see this! Does this also address #6349?

outofambit:
> Does this also address #6349?

@billygriffin no 😢 , but its a step on the way to fixing that one!

shiftkey:
@outofambit I'd still like to see a regression test added that confirms the behaviour and that this fixes the issue

outofambit:
@shiftkey had a run at introducing a more testable version (as well as a regression test). let me know your thoughts!

app/test/unit/app-store-test.ts
```diff
@@ -89,7 +94,7 @@ describe('AppStore', () => {
   it('can select a repository', async () => {
     const appStore = await createAppStore()
 
-    const repo = await setupEmptyRepository()
+    const repo = await setupConflictedRepoWithMultipleFiles()
```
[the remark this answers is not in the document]
outofambit: whoops!

outofambit:
@shiftkey 🍕

PR #6489, opened 2018-12-24
Repository Remove menu label now indicates if it will show a dialog

### LINKED ISSUE #4976: "Remove" default menu item needs to be aware of user preference
Following on from #4975, the top-level **Repository** | **Remove** menu item needs to be aware of whether the user has this set:

screen shot 2018-06-20 at 11 19 07 am

I wasn't able to fix this in #4975 because we don't have an easy way to get access to this setting (which is managed by the renderer) when generating the menu entries (which is managed by the main process).

[`build-default-menu.ts`](https://github.com/desktop/desktop/blob/061fc8e21caf5640363ae8cdb4f3f9c5c09f6b5e/app/src/main-process/menu/build-default-menu.ts#L21-L25) is the place to make this change, as well as the [related IPC call](https://github.com/desktop/desktop/blob/061fc8e21caf5640363ae8cdb4f3f9c5c09f6b5e/app/src/lib/stores/app-store.ts#L1350-L1364) from the renderer, but this is a potential opportunity to think about how to rethink this area rather than adding another parameter.

Daniel-McCarthy:
## Overview

**Closes #4976**

## Description

- The "Remove" menu label under the Repository category will now display as "Remove..." if it will result in a dialog opening. The dialog can be enabled and disabled in the options.

Before changes (With dialog enabled):
image

After changes (With dialog enabled):
image

## Release notes

Notes: The Repository Remove menu label now indicates if it will open a dialog.


app/src/lib/stores/app-store.ts
```diff
@@ -1456,6 +1457,19 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return __DARWIN__ ? 'Show Pull Request' : 'Show &pull request'
   }
 
+  private getRemoveRepoLabel() {
+    const showRemoveRepoDialog = localStorage.getItem(confirmRepoRemovalKey)
```
shiftkey: We have a helper function from #5979 that should simplify this work greatly:

```ts
  const removeRepoLabel = getBoolean(confirmRepoRemovalKey, false)
    ? __DARWIN__
      ? 'Remove...'
      : '&Remove...'
    : __DARWIN__
    ? 'Remove'
    : '&Remove'
```

shiftkey:
Also, this failing test is interesting - not sure what we want to do here:

```
Summary of all failing tests
 FAIL  app/test/unit/git/remote-test.ts
  ● git/remote › getRemotes › should return both remotes
    TypeError: Cannot read property 'origin' of undefined
      at Window.get localStorage [as localStorage] (node_modules/jsdom/lib/jsdom/browser/Window.js:256:26)
      at AppStore.getRemoveRepoLabel (app/src/lib/stores/app-store.ts:36100:90)
      at AppStore.updateMenuItemLabels (app/src/lib/stores/app-store.ts:36053:35)
      at AppStore._refreshRepository (app/src/lib/stores/app-store.ts:36729:14)
```

Daniel-McCarthy:
Ah, sorry. I had not realized that the tests had failed until today.
Thank you for the feedback!

This test is additionally strange to me as the test was passing locally for me after making the suggested changes, however is not when the test runs on the remote. I will have to keep looking into this.

shiftkey:
Alright, I've narrowed down the issue but don't really have a solution just yet.

Here's what's happening:

 - `jest` runs the `setup-test-framework.ts` which stubs the local storage mock (test environments are sandboxed)

https://github.com/desktop/desktop/blob/76ea9ca50bf7dd1f0e926a5ae6528b5c7ad30b75/app/test/setup-test-framework.ts#L1-L4

 - this `AppStore` test is the one affected by this PR change

https://github.com/desktop/desktop/blob/76ea9ca50bf7dd1f0e926a5ae6528b5c7ad30b75/app/test/unit/app-store-test.ts#L91-L101

 - Inside `AppStore._selectRepository()` is an `AppStore._refreshRepository()` **which is not `await`-ed**. I believe we did this because we didn't want to block and wait for the function to complete before continuing, but we can dig into that later

https://github.com/desktop/desktop/blob/76ea9ca50bf7dd1f0e926a5ae6528b5c7ad30b75/app/src/lib/stores/app-store.ts#L1120

 - `AppStore._selectRepository()` completes and the test passes

 - `jest` tears down the local storage mock because it believes the test is completed

 - **some time later**, `AppStore._refreshRepository()` reaches the `AppStore.updateMenuItemLabels` function which depends on local storage

https://github.com/desktop/desktop/blob/76ea9ca50bf7dd1f0e926a5ae6528b5c7ad30b75/app/src/lib/stores/app-store.ts#L1905

 - this fails because the test completed earlier and there's no global environment for the local storage

shiftkey:
Dropping myself from reviewing this issue because I haven't had a chance to come back to this, but I think `await`-ing on this action is the only way to ensure the correct behaviour of the problem test:

https://github.com/desktop/desktop/blob/9e6b012c775e56e238e80442274f2e65c38e211f/app/src/lib/stores/app-store.ts#L1142

There's potentially some side effects (`_refreshRepository` queries the underlying Git repository and updates a lot of local state) that we need to be mindful of as part of this change.

In the meantime I've merged `development` and the changes from #6228 to confirm this still needs work.

billygriffin:
Just following up here. It's a bit hard for me to follow the discussion, but it feels like maybe the complexity potentially introduced here might not be worth the payoff and maybe we should close the PR and issue? If we left it as is, we wouldn't be communicating as much information up front, but removing a repo shouldn't be a super common occurrence and my understanding is that the current behavior will still respect the user's preferences.

Thoughts @shiftkey?

shiftkey:
I'm tentatively 👎 to closing the issue because we've made our context menus follow this pattern when they know a dialog will appear on click:



At the moment this PR is blocked on being able to apply this change, because of how it impacts unit tests. Any objections to me updating #4976 to refer to this PR, summarizing the issue and closing out this PR? cc @Daniel-McCarthy

Daniel-McCarthy:
@shiftkey @billygriffin I am definitely fine with closing out this pull request as it is currently blocked, and the scope involved with fixing the outlying issues would not likely be something that is quite in my domain as of yet. 

I would tend to agree with @shiftkey that adding the relevant information to the issue and keeping it open would be a good idea, and closing this pull request as it is not likely to be moving forward quite soon.

billygriffin:
Works for me, thanks y'all.

PR #6500, opened 2018-12-27
upgrade to Typescript 3.2

shiftkey:
## Overview

VSCode is now using TS 3.2 by default, so you might encounter errors if you haven't forced VSCode to use the workspace version. I looked into what had changed between TS 3.1 and 3.2 and this was the minimum work to get things compiling again - looks like some changes to how type inference works with `Pick`.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -762,11 +762,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
         return
       }
 
+      const newState: IDisplayHistory = {
+        kind: HistoryTabMode.History,
+      }
+
       this.repositoryStateCache.updateCompareState(repository, () => ({
         tip: currentSha,
-        formState: {
-          kind: HistoryTabMode.History,
-        },
+        formState: newState,
         commitSHAs: commits,
         filterText: '',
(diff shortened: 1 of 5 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -762,11 +762,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      const newState: IDisplayHistory = {
+        kind: HistoryTabMode.History,
+      }
+
       this.repositoryStateCache.updateCompareState(repository, () => ({
         tip: currentSha,
-        formState: {
-          kind: HistoryTabMode.History,
-        },
+        formState: newState,
```
j-f1: Why these changes? What kind of errors were thrown?

shiftkey: This change is because the type inference couldn't verify this satisfies the criteria of `formState`:

> Type '{ kind: HistoryTabMode; }' is not assignable to type 'IDisplayHistory | ICompareBranch'.

If I break it out it becomes clearer where the problem lies:



The next time we're in making changes to the compare tab we can revisit these abstractions, but for now I just want to appease the Type System Gods™

Issue #5185, opened 2018-07-18
investigate using @github/octicons-react instead of our hand-rolled solution


shiftkey:
`v7.4.0` of the Octicons release added support for React components, and `v8.0.0` introduced a breaking change to use the `@github/octicons-react` package name instead.

This is of interest to me because the generated `octicons.generated.ts` contains the entire set of icons:

https://github.com/desktop/desktop/blob/56b1fe71f59d3f59202f5949ebc48af606e5f89c/app/src/ui/octicons/octicons.generated.ts#L6-L25

And it makes up a significant part of the generated renderer bundle:

<img width="1242" src="https://user-images.githubusercontent.com/359239/42886049-fe13ab46-8a78-11e8-8590-d5d5c447ce4f.png">

My gut feeling is that webpack isn't stripping out the unused source in the app (because of how the `OcticonSymbol` class is structured) and if we move to using the `@github/octicons-react` package we'd get the benefits of tree-shaking (because it's able to analyze which symbols are actually used in the app) as well as eliminating some boilerplate code for generating the `OcticonSymbol` class.

I'm not going to mark this as `help wanted` just yet, because it's used rather heavily across the codebase, and there are no typings available for it yet.

j-f1:
`@github/octicons-react` [`v8.1.0`](https://github.com/primer/octicons/releases/tag/v8.1.0) supports TypeScript!

outofambit:
i spiked on this a while ago and found it to be more effort than its worth, as much as i would like to see this happen.

PR #6516, opened 2018-12-31
update octicons package to v8.x

shiftkey:
## Overview

This is just some tidy up of the latest release of the `octicons` package. `v8` looks like it has some new icons, and I clicked through the various screens of the app and couldn't spot anything unexpected.

## Release notes

Notes: no-notes


script/generate-octicons.ts
```diff
@@ -46,7 +46,7 @@ async function generateIconData(): Promise<ReadonlyArray<IOcticonData>> {
 
   const results = new Array<IOcticonData>()
 
-  for (const name of Object.keys(octicons)) {
+  for (const name of Object.keys(octicons).sort()) {
```
shiftkey: I added this step to ensure the generated file is sorted alphabetically - the latest package has a different order after doing `const octicons = require('octicons')`

outofambit:
@shiftkey does this help us with https://github.com/desktop/desktop/issues/5185 ? 🤔 What are the benefits of doing this upgrade?

shiftkey:
@outofambit no, this is just a visual refresh of the current icons. I had a shot at migrating over to the new package some time ago and I think it didn't make a noticeable change in the generated bundle size, so I deferred it until I had some time to dig into how our bundles are generated.

shiftkey:
Gonna merge this in while @outofambit looks into the work required to undertake #5185

PR #6517, opened 2018-12-31
port script tests from mocha to jest

shiftkey:
## Overview

I'd like us to move away from `mocha` and `chai` usage in favour of plain `jest` and jest plugins now that we have the unit test suite working nicely. This PR is another step on that journey.

## Description

To avoid confusion I moved the root `jest.config.js` under `app` to move it closer to code it's associated with testing. I then added a `script/jest.config.js` that is associated with testing the scripting functionality. Like before, I also replaced the `chai` assertions with the jest builtins.

 - [x] `jest-junit` now needs to emit distinct test runs to upload to Azure Pipelines (by default it'll overwrite `junit.xml`)

## Release notes

Notes: no-notes

PR #6518, opened 2018-12-31
port integration tests to jest

shiftkey:
## Overview

~~🌵🌵🌵**builds upon #6517, go review that first** 🌵🌵🌵~~

The (almost) last part of moving away from `mocha` to `jest` is this PR.

## Description

Moving away from mocha requiring rewriting the launch test because we couldn't leverage `chai-as-promised` any more (which essentially promisified lots of `chai`'s API without having to think about it). Instead this test uses `spectron` directly to drive the app. I upgraded to the latest `spectron` because we've already been talking about Electron v3 support, but this also gave me a chance to use the typings that came in the box with the `spectron` package.

The resulting code is more procedural but once I get some fixes upstream for the type declarations (everything is a promise-returning API it seems, despite what the typings might say) I feel much more confident in us writing more tests with it (because the magic is no longer hiding in libraries and BDD-esque syntax).

- [x] rebase once #6521 has been merged, confirm all good
- [x] rebase once #6517 has been merged, confirm all good

## Release notes

Notes: no-notes


shiftkey:
I extracted the last part of the Jest migration to https://github.com/desktop/desktop/issues/6645 because I wanted to keep this PR focused on porting the tests over.

app/test/integration/launch-test.ts
```diff
@@ -28,36 +21,38 @@ describe('App', function(this: any) {
...
+    return Promise.resolve()
   })
 
-  it('opens a window on launch', function() {
-    return app.client
-      .waitUntil(() => app.browserWindow.isVisible(), 5000)
-      .getWindowCount()
-      .should.eventually.equal(1)
-      .browserWindow.isMinimized()
-      .should.eventually.be.false.browserWindow.isDevToolsOpened()
```
shiftkey: `isDevToolsOpened()` isn't exposed in the `spectron` typings because they just wrap `electron` - more context here: https://github.com/electron/spectron/issues/343

package.json
```diff
@@ -161,6 +161,7 @@
     "@types/ua-parser-js": "^0.7.30",
     "@types/untildify": "^3.0.0",
     "@types/uuid": "^3.4.0",
+    "@types/webdriverio": "^4.13.0",
```
iAmWillShepherd: What are these types supporting? I don't see an associated `"webdriverio": "<some version>"` above.

shiftkey: This is because we've enabled typings for spectron as part of this PR.

Inside `spectron.d.ts` you'll see [this line](https://github.com/electron/spectron/blob/master/lib/spectron.d.ts#L9):

```ts
    import * as WebdriverIO from "webdriverio";
```

Without adding this package, we're not able to leverage spectron's typings.

shiftkey: This was specifically added in ccd9a10ef1d8bc4f6a37ec23b5d90e3b37fc7f78 with some context, but if there's another place you'd like to add a comment to emphasise this (can't do comments inside JSON) let me know.

iAmWillShepherd: My bad, didn't realize there was some context in the description of that commit.

jtarver2018:
Unwind my ohone

PR #6519, opened 2018-12-31
Give arrow shown in the path for renamed/copied files some margins

### LINKED ISSUE #5114: UI: Arrow is too close to text.
## Description
Let's say that we change the path of a file inside a repo. When such thing happens, the arrow that indicates that we changed the path is super close to the labels.

## Version
* GitHub Desktop: version 1.2.7-beta0
* Operating system: macOS, High Sierra - version 10.13.5

## Steps to Reproduce
- Change the path of a file inside a repo
- Pay attention to this part of the app:

screen shot 2018-07-04 at 15 20 51

### Expected Behavior
- The arrow shouldn't be that close to both of the labels. One indicating the previous path and one indicating the new path.

### Actual Behavior
- The arrow is too close to both labels.

koralcem:
## Overview

**Closes https://github.com/desktop/desktop/issues/5114**

## Description

In the discussions for https://github.com/desktop/desktop/issues/5114 the last comment was to give the rename/copy arrow a fixed margin in the spots that it's used. The proposed change does that, and things look like this with it in place (note both the renamed files in the sidebar, and the file details header):
after

## Release notes

Notes:
- Gives arrow shown in the path for renamed/copied files some margins

### DIFF
--- a/app/styles/ui/_path-label.scss
+++ b/app/styles/ui/_path-label.scss
@@ -1,0 +1,12 @@
+.path-label-component {
+  display: flex;
+  flex-direction: row;
+  flex-grow: 1;
+  align-self: center;
+  min-width: 0;
+  margin-right: var(--spacing-half);
+
+  .rename-arrow {
+    margin: 0 var(--spacing-half);
+  }
+}
--- a/app/styles/ui/_path-text.scss
+++ b/app/styles/ui/_path-text.scss
@@ -7,7 +7,3 @@
     color: var(--text-secondary-color);
   }
 }
-
-.rename-arrow {
-  margin: 0 var(--spacing-half);
-}
--- a/app/styles/ui/_path-text.scss
+++ b/app/styles/ui/_path-text.scss
@@ -7,3 +7,7 @@
     color: var(--text-secondary-color);
   }
 }
+
+.rename-arrow {
+  margin: 0 10px;
+}
--- a/app/styles/ui/_path-text.scss
+++ b/app/styles/ui/_path-text.scss
@@ -9,5 +9,5 @@
 }
 
 .rename-arrow {
-  margin: 0 10px;
+  margin: 0 var(--spacing-half);
 }
--- a/app/src/ui/lib/path-label.tsx
+++ b/app/src/ui/lib/path-label.tsx
@@ -26,7 +26,7 @@ const ResizeArrowPadding = 10
 export class PathLabel extends React.Component<IPathLabelProps, {}> {
   public render() {
     const props: React.HTMLProps<HTMLLabelElement> = {
-      className: 'path',
+      className: 'path-label-component',
     }
 
     const { status } = this.props
--- a/app/styles/_ui.scss
+++ b/app/styles/_ui.scss
@@ -31,6 +31,7 @@
 @import 'ui/foldout';
 @import 'ui/preferences';
 @import 'ui/path-text';
+@import 'ui/path-label';
 @import 'ui/configure-git-user';
 @import 'ui/form';
 @import 'ui/text-box';
--- a/app/styles/ui/_file-list.scss
+++ b/app/styles/ui/_file-list.scss
@@ -48,11 +48,6 @@
       margin-right: var(--spacing-half);
     }
 
-    .rename-arrow {
-      flex: 1;
-      flex-shrink: 0;
-    }
-
     .octicon {
       vertical-align: text-bottom;
     }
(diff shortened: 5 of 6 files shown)


shiftkey:
@koralcem this seems like a good improvement, but I'd like a quick :eyes: from someone on @desktop/design about the spacing and any other ideas before I merge.

niik (changes requested):
Hey @koralcem, thanks so much for opening this!

With the exception of my comment below about decreasing the margin just slightly this is a great improvement and I'd be happy to pull this in as-is but I wanted to see if you'd be interesting in doing a follow-up improvement as well. It's entirely up to you and we can either look at it in a separate PR or in this one.

As you've seen we've tried to keep styles organized on a per-component level, which is why styles for the `PathText` component lives in `path-text-component.scss`. The rename arrow, however, isn't strictly related to the `PathText` component but rather to the `PathLabel` component (confusing, I know).

The styles for the `PathLabel` component are mostly duplicated between the `_file_list.scss` and `_changes_view.scss` files. It would be wonderful if we could share these and create a dedicated style file for the `PathLabel` component. It currently has a rather nondescript `path` class name which I think we should change to `path-label-component` and extract all the shared styles from the files mentioned above, keeping only those who differ.

The `.rename-arrow` selector and its associated styles could then be moved into that newly created file and live underneath the `.path-label-component` so as to not interfere with the global css namespace.

Let me know how you feel about making those changes!

app/styles/ui/_path-text.scss
```diff
@@ -7,3 +7,7 @@
     color: var(--text-secondary-color);
   }
 }
+
+.rename-arrow {
+  margin: 0 10px;
```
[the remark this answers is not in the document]
koralcem: I didn't realize we had sizing variables. That makes this a no-brainer. Committed.

koralcem:
@niik That sounds like a great idea. I was looking for such a central place when I first made the change, didn't have the insight you did, and hence put it at the "top level" in `_path-text.scss` (not in the `.path-text-component` scope). It's cool to see that such refactoring is encouraged even outside of the core contributors.

I attempted what you mentioned. How does it look now? I'm a little weary since I'm changing the name of the class used in a component, so please let me know if I'm missing something. There's still one more use of `.path` [here](https://github.com/desktop/desktop/blob/development/app/styles/ui/_missing-repository-view.scss#L53) but that seems to correspond to [this element](https://github.com/desktop/desktop/blob/development/app/src/ui/missing-repository.tsx#L48), and not actually a use of the `PathLabel` component.

app/styles/ui/_path-label.scss
```diff
@@ -0,0 +1,12 @@
+.path-label-component {
+  display: flex;
+  flex-direction: row;
+  flex-grow: 1;
+  align-self: center;
+  min-width: 0;
+  margin-right: var(--spacing-half);
```
koralcem: Note that this is more of a "merge" of the two CSS rules getting removed rather than a "lowest common denominator"; `file-list` didn't have `align-self: center` (above) and `changes-view` didn't have `margin-right: var(--spacing-half)` (below). I figured it was worth giving one side a center alignment and the other a right margin in exchange for having a clean central spot for `PathLabel` styles. Let me know if I was mistaken.

niik: >  Let me know if I was mistaken.

Quite the opposite. This is exactly what I hoped would happen by centralizing, I.e. that we'd find inconsistencies and eliminate them. A+

niik (approved):
Hey @koralcem, sorry for the delay in reviewing this, I had to prioritize some other work last week.

> I attempted what you mentioned. How does it look now? I'm a little weary since I'm changing the name of the class used in a component, so please let me know if I'm missing something.

This looks great to me ✨ Thanks so much for going the extra mile on this.

> There's still one more use of .path here but that seems to correspond to this element, and not actually a use of the PathLabel component.

Yeah, that's why I'm so happy we're getting this change in, a generic class name like `path` is bound to get duplicated. With the change you've made we're in a much better place.

billygriffin:
Thanks so much for the PR @koralcem! Because we're planning to release Thursday and we're cutting a beta with several things today or early tomorrow, we're going to merge this in a few days so it goes in with the next release after 1.6 (likely 1.6.1). Just wanted to give you a heads up as to the reason for the delay. We've seen smallish changes incorporate regressions and just want to make sure it has at least a few days on beta prior to pulling it into production. Thanks again!

koralcem:
Cool; no rush here. Thanks for the explanation though. Godspeed with 1.6!

niik:
Thank you so much for this @koralcem!

PR #6520, opened 2019-01-01
Fix image overlapping menu

06b:
## Overview

**Workaround to not have image overlap the menu until the root issue is fixed with the DPI scaling issue in #2480**

## Description

- When a large image gets drawn out of the diff panel (due to DPI scaling being larger than 100%) and overlaps the top menu. This prevents the image from overlapping the menu until the root issue with the DPI scaling is fixed. Unlike #6265 this will not add a scroll bar to the container.

githubmenuoverlap

## Release notes

Notes: Fixes problem where large image overlaps menu in Windows when DPI scaling settings are larger than 100%

### DIFF
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
@@ -279,6 +279,9 @@
   display: flex;
   flex: 1;
   padding-top: 10px;
+  // Workaround so images stay within diff view until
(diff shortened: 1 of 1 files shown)


tierninho:
Tested this and LGTM. Pics scale properly and do not overlap the menu.

PR #6531, opened 2019-01-03
A dialog should not close when any text in text area.

### LINKED ISSUE #6369: Double ESC should be enabled for the every filter field
In the merge dialogue modal and clone modal dialogue, if the cursor is in the filter field **and** text is present, the `esc` should clear filter not close the modal. Hitting `esc` again should close the modal.

screen shot 2018-12-07 at 3 53 25 pm

Repro: 
- select Branch > Merge into current branch
- File > Clone Repo

Should be applied to Github, Enterprise tabs, but unsure of URL tab.
screen shot 2018-12-10 at 11 28 12 am

related to: https://github.com/desktop/desktop/pull/6236

cc @altaf933 in case you want to give it a go. 

1.5.1-beta2, mac

altaf933:
dialog
## Overview

**Closes #6369**

## Description
This should clear input text filled area instead close modal.
-

## Release notes

Notes:


app/src/ui/merge-branch/merge.tsx
```diff
@@ -230,6 +231,12 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
     return renderDefaultBranch(item, matches, this.props.currentBranch)
   }
 
+  private onDismissed = () => {
+    this.state.filterText.length > 0
+      ? this.setState({ filterText: '' })
```
shiftkey: I think eslint here is not a fan of this coding style. Switching it back to a plain `if` should make it happy:

```ts
if (this.state.filterText.length > 0) {
  this.setState({ filterText: '' })
} else {
  this.props.onDismissed()
}
```

altaf933:
Wow, complete overview of pros and cons, even solution. I will take look thanks.

billygriffin:
Hi @altaf933, just wanted to follow up here and find out if you were intending to implement the solution recommended above. We'd love it if you're willing but also totally understand if not. Thanks!

altaf933:
@niik You asked for `conventional React subscription`, would you please guid what exactly a way should be ?

niik:
> @niik You asked for `conventional React subscription`, would you please guid what exactly a way should be ?

Certainly! When I said

> When we initially implemented it there was little to no support in React for it and as such we couldn't subscribe to the keydown event as we normally do in other components by adding a `onKeyDown={this.onKeyDown}` attribute and opted instead to manually subscribe to the event through [`addEventListener`](https://github.com/desktop/desktop/blob/f4efc0af4ef610f9abcc91403aa8f5547282fd0f/app/src/ui/dialog/dialog.tsx#L307)

I was talking about this: https://github.com/desktop/desktop/blob/b2249f4772e1b7d58212f1532d382e591e2468c1/app/src/ui/dialog/dialog.tsx#L315

and this

https://github.com/desktop/desktop/blob/b2249f4772e1b7d58212f1532d382e591e2468c1/app/src/ui/dialog/dialog.tsx#L311

We should be able to replace that with a simple attribute on the `<dialog>` element such that instead of manually subscribing and unsubscribing we'd do something like this

```diff
 <dialog
   ref={this.onDialogRef}
    id={this.props.id}
    onClick={this.onDialogClick}
    className={className}
    aria-labelledby={this.state.titleId}
+   onKeyDown={this.onKeyDown}
 >
```

Unfortunately, since the `cancel` dialog event still isn't supported we'll have to leave that as-is which means we're going to have to update the private `onDialogCancel` event handler to accept either a React synthetic event or a regular DOM event.

```diff
-  private onDialogCancel = (e: Event) => {
+  private onDialogCancel = (e: Event | React.SyntheticEvent<HTMLElement>) => {
```

I hope that helps clarify things but please let me know if there's more detail I can provide.

altaf933:
thanks for concrete solution,
` private onDialogCancel = (e: Event | React.SyntheticEvent<HTMLElement>) => {`

Avoid `preventDefault` by put check `defaultPrevented` when `wKeycode` fire but it closes app over onDismiss call instead just dialog.

niik:
> I avoid `preventDefault` by put check at `defaultPrevented` set and when `wKeycode` fire it closes app over onDismiss method call instead just dialog. @niik

Hey @altaf933, sorry for being so late to respond. I'm afraid I don't quite understand what problem you're running into. Perhaps if you could push what you've got I could reproduce it on my machine and then we could take a look and see what's going on.

altaf933:
I am closing this PR , so anyone can work on issue cause i see issues opened since couple of month.

thinklinux:
@niik is there a way for me to continue from this PR or I should start a new one? I want to see if I can fix it. Thanks!

Rexogamer:
@thinklinux you should open a new PR

PR #6544, opened 2019-01-07
Adds feature to auto-prune local branches

### LINKED ISSUE #750: Merged branches stay around until they're manually deleted

iAmWillShepherd:
Fixes #750 

This PR adds a feature that will auto-prune branches based on these conditions:
* The branch is deleted on GitHub.com **and**
* merged locally into the default branch in Desktop **and**
* has no commits ahead of the default branch

It will run once a repo is opened and every 24 hours for that repo, but only when that repo is opened. This means if a repo is not opened in 3 days, the next time it will be pruned is when that repo is opened. If a repo was just pruned, the next check will be 24 hours later *if* that repo is opened. To accomplish this, I'll likely use IndexedDB to store last time a repo was pruned.

Note that this PR does not include logic to only delete branches that haven't been checked out within 14 days. That logic is captured in #6664 and will be integrated after it is merged.

Todo:
- [x] Get list of merged branches
- [x] Filter out currently checked-out branch
- [x] Filter out local branches that still exist on remote
-  ~Filter out local branches that are ahead of default branch~ (Running `git branch --merged` takes care of this)
- [x] Run prune function on an interval that starts when repo is opened
- [x] Update model schema to include last prune date in the repository table
- [x] Remove test code used for dev purposes
- [x] Test - make sure reserved branches don't get deleted
-  ~Add logic to only add branch to prune list if last checkout was more than 14 days ago~ - #6664


billygriffin:
@iAmWillShepherd I'm excited to see this! Prior to adding `ready-to-review` here, would you mind clarifying scope and where you netted out from the conversation in #750? Just want to make sure we're all aligned on where we're starting with this. Thanks!

app/src/ui/repository.tsx
```diff
@@ -290,10 +290,16 @@ export class RepositoryView extends React.Component<
       <UiView id="repository" onKeyDown={this.onKeyDown}>
         {this.renderSidebar()}
         {this.renderContent()}
+        <button onClick={this.onClick}>Merged Branches</button>
```
iAmWillShepherd: This will be removed. I'm only using it for dev purposes.

app/src/lib/stores/app-store.ts
```diff
@@ -1166,18 +1169,55 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+      return
+    }
+
+    const pruner = new BranchPruner(
+      repository,
+      this.gitStoreCache,
+      this.repositoriesStore,
+      this.repositoryStateCache,
+      (repo: Repository) => {
```
shiftkey: :art: we can distill this whole callback down to `repo => this._refreshRepository(repo)` because of TS inference

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    // Get the last time this repo was pruned
+    const lastPruneDate = await this.repositoriesStore.getLastPruneDate(
+      this.repository
+    )
+
+    // Only prune if it's been at least 24 hours since the last time
+    if (
+      !__DEV__ &&
```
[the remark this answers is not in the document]
shiftkey: I don't think this is entirely covered by the feature flag check, as given how aggressive this might be I think you want to have situations where this doesn't run even after it's been stabilised is considered ready for production use.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,143 @@
...
+      lastPruneDate !== null &&
+      lastPruneDate < Date.now() + BackgroundPruneMinimumInterval * 6
+    ) {
+      log.info(`Last prune took place ${new Date(lastPruneDate)} - skipping`)
```
shiftkey: :art: maybe this could be a relative time? Or # of seconds?

app/src/lib/stores/app-store.ts
```diff
@@ -1166,18 +1169,55 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.stopBackgroundFetching()
     this.stopPullRequestUpdater()
     this.stopAheadBehindUpdate()
+    this.stopBackgrounPruner()
```
billygriffin:
```suggestion
    this.stopBackgroundPruner()
```

app/src/lib/stores/app-store.ts
```diff
@@ -1166,18 +1169,55 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
     this.startAheadBehindUpdater(repository)
     this.refreshMentionables(repository)
+    this.startBackgroundPruner(repository)
 
     this.addUpstreamRemoteIfNeeded(repository)
 
     return this.repositoryWithRefreshedGitHubRepository(repository)
   }
 
+  private stopBackgrounPruner() {
```
billygriffin:
```suggestion
  private stopBackgroundPruner() {
```

app/src/lib/stores/helpers/branch-pruner.ts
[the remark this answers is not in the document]
iAmWillShepherd: The `__Dev__` check is for me to easily test, it will be removed before merging.

app/src/lib/git/branch.ts
```diff
@@ -156,3 +156,18 @@ export async function getBranchesPointedAt(
...
   return stdout.split('\n').slice(0, -1)
 }
+
+export async function getMergedBranches(
+  repository: Repository,
+  branchName: string
+): Promise<ReadonlyArray<string>> {
+  const args = ['branch', '--merged', branchName]
+
+  const { stdout } = await git(args, repository.path, 'mergedBranches')
```
niik: I think we should use a similar approach here as we do in `getBranches`, i.e. `git branch --format="%(refname:short)"`. That way we're certain that we're getting the same values from `getBranches` and `getMergedBranches` and it also saves us from having to do string trimming.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,142 @@
...
+import { GitStoreCache } from '../git-store-cache'
+import { getMergedBranches } from '../../git'
+import { fatalError } from '../../fatal-error'
+import { enableBranchPruning } from '../../feature-flag'
+
+/** Check if a repo needs to be pruned at least every 4 hours */
+const BackgroundPruneMinimumInterval = 1000 * 60 * 60 * 4
+
+export class BranchPruner {
+  private timer: NodeJS.Timer | null = null
```
niik: This should be `number | null` and we should use `window.setTimeout`. That's what's actually gonna get stored into that variable regardless of what type information we set on it but there's a conflict of types.

app/src/lib/stores/repositories-store.ts
```diff
@@ -334,4 +335,62 @@ export class RepositoriesStore extends BaseStore {
...
+    if (!repoID) {
+      return fatalError(
+        '`updateLastPruneDate` can only update the last prune date for a repository which has been added to the database.'
+      )
+    }
+
+    const githubRepo = repository.gitHubRepository
+    if (githubRepo === null) {
+      return fatalError(
+        `'updateLastPruneDate' can only update GitHub repositories`
```
niik: I might be reading this wrong but this looks like it's saying that we'll only prune branches for GitHub repositories. Why is that?

app/src/lib/stores/repositories-store.ts
```diff
@@ -334,4 +335,62 @@ export class RepositoriesStore extends BaseStore {
...
+    const gitHubRepositoryID = githubRepo.dbID
+    if (gitHubRepositoryID === null) {
+      return fatalError(
+        `'getLastPruneDate' can only retrieve the last prune date for GitHub repositories that have been stored in the database.`
+      )
+    }
+
+    const record = await this.db.gitHubRepositories.get(gitHubRepositoryID)
+
+    return record!.lastPruneDate
```
niik: I really don't think we should use `!` except for in tests. If we're certain that this can't ever be `undefined` (which I don't think we can be) we should check and throw if it is.

app/src/lib/stores/app-store.ts
```diff
@@ -1181,18 +1185,55 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.stopBackgroundFetching()
     this.stopPullRequestUpdater()
     this.stopAheadBehindUpdate()
+    if (enableBranchPruning()) {
```
shiftkey: This check shouldn't be necessary as the teardown code only runs when `this.currentBranchPruner` is assigned, and that is wrapped in a `enableBranchPruning()` check in `startBackgroundPruner()`

app/src/lib/stores/repository-state-cache.ts
```diff
@@ -99,7 +99,7 @@ export class RepositoryStateCache {
   }
 }
 
-function getInitialRepositoryState(): IRepositoryState {
+export function getInitialRepositoryState(): IRepositoryState {
```
shiftkey: :art: I don't think `export` this is needed any more

app/src/lib/stores/repositories-store.ts
iAmWillShepherd: We initially decided to support only GH repos because we want to be sure we aren't deleting branches we shouldn't be deleting. Since we rely on the default branch to determine what's been merged, we have to ensure we have the correct default branch.

iAmWillShepherd:
> I might be reading this wrong but this looks like it's saying that we'll only prune branches for GitHub repositories. Why is that?

I decided it would be best to continue to only allow GH hosted repos to have their branches pruned. We can continue to iterate in the future if we would like to support 3rd-party hosted repos. 

I wanted to do a quick test to see if we could support external repos, so I decided to test out with BitBucket. 

First thing I did was setup a repo and set `this-is-the-main-one` to be default _development_ branch
screen shot 2019-01-17 at 10 11 33 am

I then cloned the repo and opened the _Create Branch dialog_ to see what we (Desktop) think the default branch is. As you can see in the screenshot, we think the default is `master` which is incorrect
screen shot 2019-01-17 at 10 11 22 am

@shiftkey suggested I check what `refs/remotes/origin/HEAD` pointed to and got the following:
```sh
~/Desktop/test(master) » git symbolic-ref refs/remotes/origin/HEAD
refs/remotes/origin/master
```

Because I was unable to get the result I expected from one of the top git hosting companies, I decided to hold off on supporting all repos in this PR.

tierninho:
Tests for this feature:

- [ ] (((WIP))): Prune occurs if: The branch is deleted on GitHub.com; merged locally into the default branch in Desktop; and has no commits ahead of the default branch
- [x] IndexedDB > Database > GHrepos > name > lastPruneDate records prune date

Had trouble getting the prune to run when the repo was opened. Also, gonna avoid testing the auto-run intervals as it is more important that the prune happens correctly. Lastly, non-GH repos were already tested as you noted above.

I am getting this error when navigating back to a repo with a deleted branch: error: Cannot delete branch 'prune-me' checked out at '/Users/me/github/tests/prunes'. Is this intended?

shiftkey:
> I am getting this error when navigating back to a repo with a deleted branch: `error: Cannot delete branch 'prune-me' checked out at '/Users/me/github/tests/prunes'`. Is this intended?

@tierninho this feels like the branch pruner is trying to delete the branch you're currently on. This might be a case to investigate @iAmWillShepherd

iAmWillShepherd:
Marking this read-for-review since this is behind a feature flag.

shiftkey:
@iAmWillShepherd does #6664 need to be reviewed and merged before this?

iAmWillShepherd:
> @iAmWillShepherd does #6664 need to be reviewed and merged before this?

Naw, I'll work on integrating the two in #6664.

shiftkey:
To simplify the PR changes, the `app/test/fixtures/branch-prune-tests/_git/hooks/*.sample` can be removed as they are just sample files and are not run as part of any Git operation.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,155 @@
...
+  private timer: number | null = null
+
+  public constructor(
+    private readonly repository: Repository,
+    private readonly gitStoreCache: GitStoreCache,
+    private readonly repositoriesStore: RepositoriesStore,
+    private readonly repositoriesStateCache: RepositoryStateCache,
+    private readonly onPruneCompleted: (
+      repository: Repository,
+      prunedBranches: ReadonlyArray<Branch>
```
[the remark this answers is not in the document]
iAmWillShepherd: Nope
🔥🔥🔥

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,153 @@
...
+    const branchesReadyForPruning = new Array<Branch>()
+    for (const branch of mergedBranches) {
+      const localBranch = localBranches.find(
+        localBranch =>
+          localBranch.remote !== null && localBranch.name === branch
+      )
+
+      if (
+        localBranch !== undefined &&
+        !ReservedBranches.includes(branch.toLocaleLowerCase())
```
shiftkey: :art: we could filter on this earlier and save the lookup for some branches

```ts
const candidateBranches = mergedBranches.filter(b => !ReservedBranches.includes(b))

for (const branch of candidateBranches) {
  // ...
  if (localBranch) {
    branchesReadyForPruning.push(localBranch)
  }
}
```

or

```ts
for (const branch of mergedBranches) {
  if (ReservedBranches.includes(branch.toLocaleLowerCase()) {
    continue;
  }

  // ...

  if (localBranch) {
    branchesReadyForPruning.push(localBranch)
  }
}

File: app/test/unit/branch-pruner-test.ts [hunk 15]
```diff
@@ -0,0 +1,261 @@
...
+    repositoriesStore = new RepositoriesStore(repositoriesDb)
+    repositoriesStateCache = new RepositoryStateCache(
+      () => new Map<string, IGitHubUser>()
+    )
+    onPruneCompleted = jest.fn(() => (_: Repository) => {
+      return Promise.resolve()
+    })
+  })
+
+  it('Does nothing on non GitHub repositories', async () => {
```
shiftkey: :art: we usually avoid sentence case for these test names. These were the only exceptions I could find:

screen shot 2019-02-06 at 10 26 21 am

If you run `yarn test:unit --verbose` you'll see what this looks like:

```
...

 PASS  app/test/unit/list-selection-test.ts
  list-selection
    findNextSelectableRow
      ✓ returns first row when selecting down outside list (filter text) (2ms)
      ✓ returns first selectable row when header is first (1ms)
      ✓ returns first row when selecting down from last row (1ms)
      ✓ returns last row when selecting up from top row

 PASS  app/test/unit/cloning-repository-test.ts
  CloningRepository
    name
      ✓ provides the name of the repository being cloned (1ms)
      ✓ extracts the repo name from the url not the path
      ✓ extracts the repo name without git suffix (1ms)

 PASS  app/test/unit/fatal-error-test.ts
  forceUnwrap
    ✓ fails for null (1ms)
    ✓ fails for undefined (1ms)
    ✓ passes for false
    ✓ passes for a hash (1ms)
```

```suggestion
  it('does nothing on non GitHub repositories', async () => {
```

app/test/unit/branch-pruner-test.ts
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+    const branchesBeforePruning = await getBranchesFromGit(repo)
+    await branchPruner.start()
+    const branchesAfterPruning = await getBranchesFromGit(repo)
+
+    expect(branchesBeforePruning).toEqual(branchesAfterPruning)
+  })
+
+  it('Prunes for GitHub repository', async () => {
```
shiftkey:
```suggestion
  it('prunes for GitHub repository', async () => {
```

app/test/unit/branch-pruner-test.ts
```diff
@@ -0,0 +1,261 @@
...
+      'not-deleted-branch-1',
+      'deleted-branch-1',
+    ]
+
+    for (const branch of expectedBranchesAfterPruning) {
+      expect(branchesAfterPruning).not.toContain(branch)
+    }
+  })
+
+  it('Does not prune if the last prune date is less than 24 hours ago', async () => {
```
shiftkey:
```suggestion
  it('does not prune if the last prune date is less than 24 hours ago', async () => {
```

app/test/unit/branch-pruner-test.ts
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+    const branchesBeforePruning = await getBranchesFromGit(repo)
+    await branchPruner.start()
+    const branchesAfterPruning = await getBranchesFromGit(repo)
+
+    expect(branchesBeforePruning).toEqual(branchesAfterPruning)
+  })
+
+  it('Does not prune if there is no default branch', async () => {
```
shiftkey:
```suggestion
  it('does not prune if there is no default branch', async () => {
```

app/test/unit/branch-pruner-test.ts
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+    const branchesBeforePruning = await getBranchesFromGit(repo)
+    await branchPruner.start()
+    const branchesAfterPruning = await getBranchesFromGit(repo)
+
+    expect(branchesBeforePruning).toEqual(branchesAfterPruning)
+  })
+
+  it('Does not prune reserved branches', async () => {
```
shiftkey:
```suggestion
  it('does not prune reserved branches', async () => {
```

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,153 @@
...
+    // Get the last time this repo was pruned
+    const lastPruneDate = await this.repositoriesStore.getLastPruneDate(
+      this.repository
+    )
+
+    // Only prune if it's been at least 24 hours since the last time
+    const dateNow = moment()
+    const threshold = dateNow.subtract(24, 'hours')
+
+    if (lastPruneDate !== null && threshold.isBefore(lastPruneDate)) {
```
shiftkey: I was able to trip the wrong path because I got a `lastPruneDate` of `undefined` in here the first time I switched to a repository when running this up - could the new field have been missing from Dexie, despite `Date | null` being the expected type?

screen shot 2019-02-06 at 10 43 43 am

shiftkey: This eventually came good for me, which I think is down to the `undefined`/`null` distinction:

screen shot 2019-02-06 at 10 49 19 am

The confusing part of this is the log messages:

```
info: [ui] Last prune took place in a day - skipping
```

Not sure whether you want to relax this check to `== null` here for those repositories which haven't been updated to see the new `null`.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,153 @@
...
+        !ReservedBranches.includes(branch.toLocaleLowerCase())
+      ) {
+        branchesReadyForPruning.push(localBranch)
+      }
+    }
+
+    log.info(
+      `Pruning ${branchesReadyForPruning.length} refs from '${
+        this.repository.name
+      } using '${defaultBranch.name} (${defaultBranch.tip.sha})' as base branch`
```
shiftkey: > `info: [ui] Pruning 0 refs from 'desktop using 'development (6400fd767654d62a23c79f01bbdb67214c16c0bc)' as base branch`

adding the loneliest quote mark:

```suggestion
      }' using '${defaultBranch.name} (${defaultBranch.tip.sha})' as base branch`
```

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -0,0 +1,153 @@
...
+    )
+
+    const gitStore = this.gitStoreCache.get(this.repository)
+    await gitStore.performFailableOperation(async () => {
```
shiftkey: Oh, I see you've got the outer `performFailableOperation` here as well as the one I suggested.

I don't think we need that duplication - if the inner wrapper encounters an error it'll proceed and try to delete the next branch in the list (the outer one would give up on the first failure and you'd have to wait 24 hours for it to go again).

app/src/lib/stores/helpers/branch-pruner.ts
iAmWillShepherd: Good call out, I went with the first option since that reduces the runtime complexity by quite a bit!

iAmWillShepherd:
💯

shiftkey:
This has gone through a long review process and is feature-flagged for development currently, so to prevent further delays I plan to merge this in around 1pm SF time today so we can unblock #6664.

PR #6567, opened 2019-01-09
introduce emitter to abstract away ahead/behind interactions

shiftkey:
## Overview

I was pairing with @iAmWillShepherd on #6544 yesterday which introduces more background work into the `AppStore` for the selected repository. That's fine for the task at hand, but it reminded me about how much `AppStore` does and motivated me to revisit how we can extract patterns of responsibility from it that may live elsewhere.

The pattern I'm noticing is this:

 - when a different repository is selected, we tear down and setup components that care about this event
 - sometimes these components need to run on a specific timer that they manage (background fetcher, pull request updater, pruning local branches)
 - other times we want to manually invoke them to do stuff (ahead/behind updater, also pull request updater)
 - a task that we currently perform but haven't adapted to this pattern is `refreshMentionables` (we check it when the repository is selected, but don't do any background work afterwards)

TL;DR: inside `AppStore` we're repeating this pattern: 
  - stop the current component, if alive, and dispose it
  - create the new component, providing it the newly-selected repository
  - assign it to some state inside `AppStore`
  - start the new component

I have a prototype of extracting these lying around from ages ago when I was tinkering with Redux months ago, but I didn't really generalize the pattern because I was caught up in figuring out what was feasible (spoiler alert: [middleware is awesome](https://redux.js.org/advanced/middleware)). But yesterday's pairing made me realise I can do this without Redux, so now that I'm staring at more usage of this pattern I'll start thinking about how we can support this in the current application architecture.

The imminent problem I need to tackle is demonstrated by this PR - how to prevent updaters from being used directly in `AppStore`. If I can isolate these components in a sane manner, they can potentially be managed outside of `AppStore`, meaning less moving parts inside `AppStore` itself. That's the direction I want to head with this.

## Description

This PR abstracts away the interactions between `AppStore` and `AheadBehindUpdater` aside from the `start`/`stop` functions. It does this by introducing an `AheadBehindEmitter` object that lets components subscribe to and raise events related to ahead/behind actions. This lets us make all functionality aside from `start()` and `stop()` inside `AheadBehindUpdater` private to make decoupling easier in the future.

The other important change about this PR is removing `ComparisonCache` - this was only a wrapper over the top of the `Map<string, IAheadBehind>` store, and I don't believe it was adding value by using it directly in application state.

While I was in here testing changes I also noted that `scheduleComparisons` would do some unnecessary work around ahead/behind computation when you click to close the branches list in the compare tab. Basically you'd see it clear the pending comparisons, compute the missing comparisons, and clear them again:

With this change you'll now see only a single `[AheadBehindUpdater] - abandoning [number] pending comparisons` debug message to indicate that it didn't try to re-schedule comparison work.

## Release notes

Notes: no-notes


shiftkey:
Parking this one for now, will come back to it with fresh eyes at some stage.

Issue #6586, opened 2019-01-10
lib/git return type audit


outofambit:
Many of our functions in `lib/git` have the return type `Promise<void>`. This makes it difficult for callers to know if the function succeeded or failed (unless it throws an error).

This is particularly important for cases like https://github.com/desktop/desktop/blob/aa79543d8d373763dccc7a51a26e277db1d5960c/app/src/lib/stores/app-store.ts#L3134 where the merge may fail (for expected reasons) be we need to have that information as the caller.

## Goal

Determine the appropriate return type for each of these functions.

Akin to https://github.com/desktop/desktop/issues/5923, it may be appropriate for some functions to return a sha or some other information on success, but at many of these functions should at least return `true` or a generic "success" type to indicate to the caller that nothing went wrong. (we should have at the very least exit codes from `git()` for all of these functions to judge that by.)

## Functions to Audit

(these are all functions with the return type `Promise<void>` in `lib/git`)

- [ ] `createMergeCommit`
- [ ] `setRemoteURL`
- [ ] `applyPatchToIndex`
- [ ] `renameBranch`
- [ ] `checkoutBranch`
- [ ] `checkoutPaths`
- [ ] `clone`
- [ ] `setGlobalConfigValue`
- [ ] `writeGitDescription`
- [ ] `fetchRefspec`
- [ ] `fetch`
- [ ] `saveGitIgnore`
- [ ] `initGitRepository`
- [ ] `installGlobalLFSFilters`
- [ ] `abortMerge`
- [ ] `pull`
- [ ] `push`
- [ ] `revertCommit`
- [ ] `unstageAllFiles`
- [ ] `resetSubmodulePaths`
- [ ] `stageFiles`
- [ ] `updateRef`

## Notes

Any PRs that change these return types should also include tests that verify those types are returned in the right situations so we don't inadvertently cause regressions.

cc @shiftkey, as well as @say25 for additional thoughts since they've done [prior work](https://github.com/desktop/desktop/pull/6089) like this

say25:
Seems reasonable to me. I wouldn't mind taking this on myself. There should be no downside to exposing more info as we wont be losing any context by no longer returning `void`.

outofambit:
@say25 awesome :) yeah if you want to take this on that sounds great. does one PR per file seem reasonable? or whatever chunking you think makes sense so long as it's not all one giant PR. 🏂

shiftkey:
[...]

PR #6606, opened 2019-01-13
manualConflictResolution and stageManualConflictResolution

outofambit:
## Overview

supports #6062 

## Description

- new `ManualConflictResolution` model
- allow `createMergeCommit` to take these as an argument and apply them in the merge process
- tests
- much stolen/borrowed from @shiftkey's spike on this

## Release notes

Notes:

### DIFF
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -70,6 +70,10 @@ export async function createMergeCommit(
       const file = files.find(f => f.path === path)
       if (file !== undefined) {
         await stageManualConflictResolution(repository, file, resolution)
+      } else {
+        log.error(
+          `couldn't find file ${path} even though there's a manual resolution for it`
+        )
       }
     }
 
@@ -122,11 +126,11 @@ async function stageManualConflictResolution(
   const { status } = file
   // if somehow the file isn't in a conflicted state
   if (!isConflictedFileStatus(status)) {
-    console.error(`tried to manually resolve unconflicted file (${file.path})`)
+    log.error(`tried to manually resolve unconflicted file (${file.path})`)
     return false
   }
   if (!isManualConflict(status)) {
-    console.error(
+    log.error(
       `tried to manually resolve conflicted file with markers (${file.path})`
     )
     return false
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -120,6 +120,14 @@ export async function createMergeCommit(
   }
 }
 
+/**
+ * Stages a file with the given manual resolution method. Useful for resolving binary conflicts at commit-time.
+ *
+ * @param repository
+ * @param file conflicted file to stage
+ * @param manualResolution method to resolve the conflict of file
+ * @returns true if successful, false if something went wrong
+ */
 async function stageManualConflictResolution(
   repository: Repository,
   file: WorkingDirectoryFileChange,
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -82,8 +82,10 @@ export async function createMergeCommit(
     await stageFiles(repository, otherFiles)
     const result = await git(
       [
-        'commit', // no-edit here ensures the app does not accidentally invoke the user's editor
-        '--no-edit', // By default Git merge commits do not contain any commentary (which
+        'commit',
+        // no-edit here ensures the app does not accidentally invoke the user's editor
+        '--no-edit',
+        // By default Git merge commits do not contain any commentary (which
         // are lines prefixed with `#`). This works because the Git CLI will
         // prompt the user to edit the file in `.git/COMMIT_MSG` before
         // committing, and then it will run `--cleanup=strip`.
--- a/app/src/lib/git/commit.ts
+++ b/app/src/lib/git/commit.ts
@@ -1,8 +1,18 @@
 import { git, GitError, parseCommitSHA } from './core'
 import { stageFiles } from './update-index'
 import { Repository } from '../../models/repository'
-import { WorkingDirectoryFileChange } from '../../models/status'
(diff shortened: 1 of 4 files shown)


app/src/lib/git/commit.ts
```diff
@@ -46,20 +56,30 @@ export async function createCommit(
...
+    const otherFiles = files.filter(f => !manualResolutions.has(f.path))
+
+    await stageFiles(repository, otherFiles)
     const result = await git(
       [
-        'commit',
-        // no-edit here ensures the app does not accidentally invoke the user's editor
-        '--no-edit',
-        // By default Git merge commits do not contain any commentary (which
+        'commit', // no-edit here ensures the app does not accidentally invoke the user's editor
```
shiftkey: :art: minor thing, but it looks like the comments got pulled up rather than being kept nearer to their relevant comments

app/src/lib/git/commit.ts
```diff
@@ -94,6 +114,55 @@ export async function createMergeCommit(
...
 
+async function stageManualConflictResolution(
+  repository: Repository,
+  file: WorkingDirectoryFileChange,
+  manualResolution: ManualConflictResolution
+): Promise<boolean> {
+  const { status } = file
+  // if somehow the file isn't in a conflicted state
+  if (!isConflictedFileStatus(status)) {
+    console.error(`tried to manually resolve unconflicted file (${file.path})`)
```
shiftkey: `log.error` to ensure these cases are flushed to our logging infrastructure if they happen in the wild

app/src/lib/git/commit.ts
```diff
@@ -94,6 +114,55 @@ export async function createMergeCommit(
...
+    return false
+  }
+  if (!isManualConflict(status)) {
+    console.error(
```
shiftkey: `log.error` to ensure these cases are flushed to our logging infrastructure if they happen in the wild

app/test/unit/git/commit-test.ts
```diff
@@ -514,24 +516,75 @@ describe('git/commit', () => {
     })
   })
 
-  describe('createMergeCommit with a merge conflict', () => {
-    let repository: Repository
-    describe('with a merge conflict', () => {
+  describe('createMergeCommit', () => {
+    describe('with a simple merge conflict', () => {
+      let repository: Repository
+      describe('with a merge conflict', () => {
```
shiftkey: :art: does this seem necessary given another `describe` has been inserted two lines above?

app/src/lib/git/commit.ts
```diff
@@ -46,20 +56,30 @@ export async function createCommit(
...
   // working directory and the last commit (if any) so our commits should
   // do the same thing.
   try {
     await unstageAll(repository)
-    await stageFiles(repository, files)
+
+    // apply manual conflict resolutions
+    for (const [path, resolution] of manualResolutions) {
+      const file = files.find(f => f.path === path)
+      if (file !== undefined) {
```
shiftkey: Do you think we should add some debug tracing in here to catch if a `manualResolutions` entry doesn't exist in the `files` list?

app/test/unit/git/commit-test.ts
```diff
@@ -514,24 +516,75 @@ describe('git/commit', () => {
...
+        )
+        const manualResolutions = new Map([
+          ['bar', ManualConflictResolutionKind.ours],
+        ])
+        const sha = await createMergeCommit(
+          repository,
+          trackedFiles,
+          manualResolutions
+        )
+        expect(await FSE.pathExists(path.join(repository.path, 'bar'))).toBe(
```
shiftkey: :art: is it worth adding a test in for the "theirs" case here to see that the resolution is applied correctly and the file is removed from the working directory? It's not clear to me what the options were based on the setup in `setupConflictedRepoWithMultipleFiles`...

app/test/unit/git/commit-test.ts
shiftkey: I ask because it looks like the outer case seems to be enough to describe the test case:

```
 createMergeCommit
  - with a simple merge conflict
     - creates a merge commit
```

with this additional `describe`:

```
 createMergeCommit
  - with a simple merge conflict
     - with a merge conflict
        - creates a merge commit
```

app/src/lib/git/commit.ts
```diff
@@ -94,6 +120,55 @@ export async function createMergeCommit(
   }
 }
 
+async function stageManualConflictResolution(
+  repository: Repository,
+  file: WorkingDirectoryFileChange,
+  manualResolution: ManualConflictResolution
+): Promise<boolean> {
```
iAmWillShepherd: What does the return type imply here? It's not obvious what `true` or `false` means. Looking above at https://github.com/desktop/desktop/pull/6606/files#diff-7f18b49a6d37b89590132862b18c161aR72 also provides no hints since we don't capture the result of this function call. I'm thinking we can return an empty promise.

outofambit: `true` is success, while `false` means something went wrong. its true we don't use that information anywhere right now, but i still think it's generally useful (see #6586 for more thoughts on this pattern). i'm happy to add some docs to this function to clarify, do you think that is sufficient @iAmWillShepherd?

outofambit:
@iAmWillShepherd 🍕

app/src/lib/git/commit.ts
iAmWillShepherd: That makes sense, but could we create a type for that so we get back `Success<T>` or `Failure` where `Failure` is an object that has some other info we may need to handle the failure?

As for this PR, adding docs is sufficient.

outofambit: absolutely, i would love that! i'm going to add a link to this comment in that issue

Issue #6593, opened 2019-01-11
Connection to repository broken after installing Git


Migdalin:
## Description

Push to repository from GitHub Desktop failed to arrive at the repository, though no errors were generated, and the UI claims that the change was pushed without issue.

## Version

* GitHub Desktop:  1.5.1

* Operating system:  Windows 10

## Steps to Reproduce

1.  Install GitHub Desktop.
2. Create a repository.  
3.  Merge and push files into the repository.
4.  Pull the repository onto a second machine.
5.  On the first machine, install Git.  (You will be prompted to do this, "if you want to merge a file manually.")
6.  Now, merge changes and push to the repository from machine #1.
7.  Return to machine #2.  Fetch.  
8.  See that the changes weren't fetched.  Look at the repository online.  See that the changes don't appear.

### Expected Behavior

1.  Require that users have Git installed locally before installing GitHub Desktop.  Or, at least forcefully warn them that operations like manual merging will be unavailable.  I don't know for sure that this is what broke the connection, but requiring that Git be installed still makes sense.  Merging changes is a standard operation.  Before the desktop can claim to be installed correctly, the desktop should support merging.
2.  During installation, warn users that, while Git will let them select various editors, GitHub Desktop only supports Atom.  Same rationale as above.
3.  If the installer doesn't require that Git be installed first, handle situations where it is installed afterwards.  For sure, don't pretend that a push to the repository worked when it didn't.

### Actual Behavior

A push to the repository appeared to succeed, but the files never made it there.

## Additional Information

![github desktop screenshot](https://user-images.githubusercontent.com/6393616/51047428-71cc4180-1596-11e9-9509-2b2b1bbb1a76.png)
![github online screenshot](https://user-images.githubusercontent.com/6393616/51047429-71cc4180-1596-11e9-9d26-0edf9dae7b80.png)

### Logs

[2019-01-11.desktop.production.log](https://github.com/desktop/desktop/files/2750464/2019-01-11.desktop.production.log)

Migdalin:
If multiple editors are supported, then improved messaging sounds like a really good idea.  I tried following the link to the full list, but it gives me "page not found."  Anyway, I installed Notepad++, because the Git installer listed it as a supported editor, but Desktop still appears to insist on Atom or nothing.

![install atom screenshot](https://user-images.githubusercontent.com/6393616/51074485-c8488700-164d-11e9-88b8-07a146c85d1f.png)

shiftkey:
[...]

Issue #1011, opened 2017-03-07
Don't disable push button when doing background fetching


cheshire137:
A user experience problem I brought up [in Slack](https://github.slack.com/archives/desktop/p1488925322009929) is that the 'push' button disables with a spinner right when I want to click it. Like I'll make a commit in the app and as I go to click the 'push' button, it will gray out. @shiftkey speculated it might be due to background fetching. I'm able to get around this by going to a term for `git push`, which works. It would be nice to stay in the app for this operation, though.

![github_desktop](https://cloud.githubusercontent.com/assets/82317/23681010/ddc8a264-0352-11e7-9629-f56cbbf46c3e.png)

I'm in 0.0.24.

PR #6610, opened 2019-01-14
a better "push needs pull" dialog to avoid race condition

### LINKED ISSUE #2738: The repository has been updated since you last pulled. Try pulling before pushing.
If this error shows up, would it be possible to change the button "Push origin" to "Pull origin"?

image

Now I'm left with only the push option, which doesn't work, so I have to hunt in the menu to find a pull command or use command line.

Actually, now after about 40 seconds, it changed to "Pull origin" by itself. Guess there might be something wrong with the update of it.

### LINKED ISSUE #5451: Pull option does not appear when "Pull first before pushing" error is thrown
## Description

On MacOS I create a commit, also on Windows machine for the same repository from another account was pushed another commit before, when I press Push on MacOS machine is gave me the error about "Pull first before pushing...blabla" but no Pull is displayed...so i am stuck.

## Version

* GitHub Desktop:1.3.3

* Operating system:
macOS High Sierra 10.13.6

## Steps to Reproduce

In our case it was reproduced like this:
1. From windows machine push to master a commit
2. From macOS machine commit to master
3. From macOS machine try to Push (Github Desktop is showing 1 Push)
4. Appears an error something like - "You have to pull before pushing"
5. Github Desktop continues to show that there is no Pull...
### Expected Behavior
After the step 4 usually GitHub Desktop loads the commit and updates UI -> so we see that there is a Pull, and after pulling we can Push.

### Actual Behavior

In the UI I always see that I have 1 Push and 0 Pulls and whatever I tried(restart Github Desktop, try Pushing again, adding one more commit to push), it doesn't help, It simply don't load the commit that had to be pulled before pushing.

shiftkey:
## Overview

Fixes #2738
Fixes #5451

While troubleshooting #6593 and doing some groundwork testing for the upcoming rebase flows, I noticed a bug with the classic "push needs pull" error handling that meant you could get stuck in a loop and never be able to see what's changed.

## Description

The easiest way to reproduce this issue:

 - create a new repository with a default branch and some commits, push it to GitHub
 - create a new branch, and publish it to GitHub **without making any commits**
 - switch over on GitHub, add a commit to the branch 
 - back in the app, create a different commit
 - try to push the commit

The generic Git error handler is displayed [with a message](https://github.com/desktop/desktop/blob/64367a525e09bb4497b495efedb225162b63f01d/app/src/lib/git/core.ts#L198), and `Dispatcher.fetch` fires in the background:

https://github.com/desktop/desktop/blob/64367a525e09bb4497b495efedb225162b63f01d/app/src/lib/dispatcher/error-handlers.ts#L243-L246

It's at this point that I'd expect to see the push/pull button update to show there's both a commit ahead and behind, but it never updates:

What's happening underneath is the concurrent network operation check (which was set to `true` when the `push` was started) has not been cleared (the error handing is being performed, but this flag is only cleared after the error handling completes), so the `fetch` is considered re-entrant and it's not perforrmed. 

https://github.com/desktop/desktop/blob/64367a525e09bb4497b495efedb225162b63f01d/app/src/lib/stores/app-store.ts#L2579-L2582

I suspect this only way this eventually "comes good" is when the background fetch eventually kicks in, and nothing changes while the impatient user (i.e me) repeatedly presses "Push" and wonders what happened.

This PR replaces that "fire and forget" fetch with a proper dialog to better explain the situation and run the fetch outside the error handler. I haven't finalized the flow yet, but here's the current message you will see:

I'd love your ideas about what we want to explain here @desktop/product @desktop/design 

## TODO

 - [x] confirm bug is resolved
 - [x] polish dialog message
 - [x] cleanup tracing work

## Release notes

Notes: fix "push needs pull" flow to explain context and actions better


outofambit:
> and run the fetch outside the error handler.

@shiftkey what do you mean by this? Are you planning to modify the underlying handling of fetching while another network operation is in process or is this PR only dealing with the dialog displayed?

shiftkey:
@outofambit I don't plan to change the concurrent network check in this PR. #6141 leverages this behaviour to address some long-running issues, and the only motivator for changing this behaviour that I can currently think of is #1011.

The new behaviour is to perform the `fetch` when the user presses `Fetch` in the new dialog.

app/src/ui/push-needs-pull/push-needs-pull-warning.tsx
```diff
@@ -0,0 +1,52 @@
...
+        }
+        onDismissed={this.props.onDismissed}
+        onSubmit={this.onFetch}
+        type="warning"
+      >
+        <DialogContent>
+          <p>
+            Desktop is unable to push commits to this branch because there are
+            new commits on the remote. It needs to fetch these new commits to
+            figure out how to reconcile these changes with your local commits.
```
billygriffin: I wonder if it's possible to make this a bit more beginner-friendly with a small change.

"Desktop is unable to push commits to this branch because there are commits on the remote that are not present on your local branch. It needs to fetch these new commits before pushing in order to reconcile them with your local commits."

I don't love it so feel free to suggest alternatives (or disagree).

allthesignals:
I am also seeing this - I'm not sure if I have my remote tracking set up incorrectly or not, but there is no option anywhere to sync with origin or pull latest commits from the branch.

I know I have commits on the remote but I can't pull them down through this interface. Fine, I can just do it through command line, but then this is pretty disruptive of a common git workflow.

shiftkey:
Dropping in some updated screenshots of the dialog (my screen recording setup is busted right now, apologies for the lo-fi info):



I also added in a Loading spinner to the fetch button because the user might not see the fetch progress because of how the dialog steals the attention (screenshot is a bit darker because I literally had to `debugger` to take a decent shot):



I went down this path because it felt like a little "what is happening" moment and showing some progress before closing the dialog seemed like a good way to help transition the user back to try and pull again.

shiftkey:
I've assigned this to `1.7.0` because it's was something I encountered easily while testing rebase flows, but this could be pulled in sooner if we want to address the original bug #2738 before then.

app/src/ui/push-needs-pull/push-needs-pull-warning.tsx
```diff
@@ -0,0 +1,71 @@
...
+        dismissable={!this.state.isLoading}
+        onDismissed={this.props.onDismissed}
+        onSubmit={this.onFetch}
+        type="warning"
+      >
+        <DialogContent>
+          <p>
+            Desktop is unable to push commits to this branch because there are
+            commits on the remote that are not present on your local branch. It
+            needs to fetch these new commits before pushing in order to
```
say25: Maybe drop the `It needs to `
```suggestion
            Fetch these new commits before pushing in order to
```

shiftkey: There's a preceding `It` on the last line that means I had to push e38b7f9 rather than take this in.

shiftkey:
Just merged `development` into this branch because of some conflicts with #6755

PR #6611, opened 2019-01-14
Change merge hint and merge button order

### LINKED ISSUE #6467: Order of Merge hint and merge button are inconsistent
The order of the merge hint and merge button in the Compare Branch feature and the Merge into Current Branch modal are different. 

It is either:
`Merge button, merge status icon, merge hint`
or
`Merge status icon, merge hint, merge button`

screen shot 2018-12-20 at 11 53 42 am

For consistency, we should make them the same. cc @desktop/design 

v1.5.1, mac

Raul6469:
## Overview

Fixes #6467 

## Description
This changes the order of merge elements in order to be in sync with what is in the merge modal.

image

However, as you can see, I just changed the order by now, but it looks like there is some styling needed, but I'm unsure about what is expected...

### DIFF
--- a/app/styles/ui/_merge-status.scss
+++ b/app/styles/ui/_merge-status.scss
@@ -54,7 +54,7 @@
   }
 
   .merge-message {
-    padding-bottom: 10px;
+    padding-bottom: var(--spacing);
   }
 }
 
--- a/app/styles/ui/_merge-status.scss
+++ b/app/styles/ui/_merge-status.scss
@@ -17,7 +17,6 @@
 
     &:after {
       display: block;
-      border-bottom: var(--base-border);
       content: '';
       position: absolute;
       z-index: 0;
@@ -52,6 +51,10 @@
     strong {
       color: var(--text-color);
     }
+  }
+
+  .merge-message {
+    padding-bottom: 10px;
   }
 }
 
--- a/app/styles/ui/_merge-status.scss
+++ b/app/styles/ui/_merge-status.scss
@@ -1,6 +1,4 @@
 .merge-status-component {
-  margin-top: var(--spacing-third);
-
   .merge-status-icon-container {
     height: 20px;
     position: relative;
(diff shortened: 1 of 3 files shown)


shiftkey:
cc @desktop/design for thoughts about that subtle line wrapping the  ✅ or ❌

ampinsk:
@shiftkey I think we need to get rid of the rule around the octicon and add a little padding between the message and the button. I can push a commit fixing if that works for you?

Raul6469:
Thanks @ampinsk for your help! I did the changes you asked, and here is what it looks like now:
image

shiftkey (approved):
Changes seem fine and address the order issue, but I'd love a quick :eyes: and ✅ from @desktop/design about the visuals with the border now gone. The merge status does seem a bit isolated (if that's the right word?) when surrounded by all this whitespace in the merge dialog footer:

app/styles/ui/_merge-status.scss
```diff
@@ -53,6 +52,10 @@
       color: var(--text-color);
     }
   }
+
+  .merge-message {
+    padding-bottom: 10px;
```
ampinsk: We can use a spacing variable here:
```suggestion
    padding-bottom: var(--spacing);
```

app/styles/ui/_merge-status.scss
```diff
@@ -17,7 +17,6 @@
 
     &:after {
```
ampinsk: I don't think we need this whole `:after` element now that we aren't using the rule anywhere

ampinsk (commented):
This is looking great! Just a few tweaks 😄

ampinsk:
We can also get rid of this margin here: 

https://github.com/desktop/desktop/blob/0a80e0a8c370593b4c0520b317f8ebcff45660c4/app/styles/ui/_merge-status.scss#L1-L2

Thanks again @Raul6469 👏

shiftkey (changes requested):
On hold pending styling tweaks

Raul6469:
Many thanks @ampinsk @shiftkey for the review! I made changes according to your feedback, let me know if I forgot something 👍

shiftkey (approved):
LGTM. Updated screenshots:




And dark theme:

ampinsk (approved):
LGTM! Thanks so much @Raul6469!

Issue #6595, opened 2019-01-11
improve error handling wrapping for Git operations


shiftkey:
Tracking issue two things that overlap with #6586:

 - extracting error handling from inside GitStore to something reusable - https://github.com/desktop/desktop/issues/6586#issuecomment-453522281
 - use a better return type than `undefined` to represent the error raised by a component: https://github.com/desktop/desktop/issues/6586#issuecomment-453562921

PR #6613, opened 2019-01-14
extract and isolate error handling from GitStore internals

shiftkey:
## Overview

**First part of #6595**

Currently the default error handling has been owned by `GitStore`, but as we've moved towards running Git operations outside of a `GitStore` (because we don't need to cache the result), there has been some confusion about the right way to leverage the inbuilt error handling so that unhandled errors do not bring down the app.

## Description

This PR extracts the logic for `GitStore.createFailableOperationHandler` into it's own module, and provides a more familiar `withErrorHandling` function in the places where this is currently used.

## Release notes

Notes: no-notes


shiftkey:
Closing this out for now to ponder on alternative ways to present this

Issue #2768, opened 2017-09-18
Github Desktop not obeying disabling of UI grow animations.


azav:
Github Desktop not obeying defaults setting disabling of UI grow animations. 

### Description
GitHub Desktop is not obeying the defaults setting to disable alerts expanding open in your face.

Alerts and windows "growing" open as the open are useless and distracting animations that slow down the developer as he waits for it to finish growing, then has to focus on the display.  This is counterproductive and unpleasant.  There is an defaults setting that disables the grow animation for alerts and windows.  It is this:

defaults write NSGlobalDomain NSAutomaticWindowAnimationsEnabled -bool NO

Github Desktop ignores this and insists on animating every window open and close.  This is a TERRIBLE user experience.  It is jarring.  It is unsettling. It makes the user wait for this unwanted action to complete.  

Please either remove it or at least obey the defaults setting above.  
### Version
**GitHub Desktop version:** 0.9.1

OS version: 10.12.6

-->

defaults write NSGlobalDomain NSAutomaticWindowAnimationsEnabled -bool NO
### Steps to Reproduce

1. In the terminal, enter this:
defaults write NSGlobalDomain NSAutomaticWindowAnimationsEnabled -bool NO
2. Launch an app and create a new window to verify that windows to not grow open
3. Launch Github desktop and open a new window.
4. Close that window

**Expected behavior:** Defaults setting should be obeyed and no grow animation should be displayed on open or close

**Actual behavior:** Window grows on open and shrinks on close, ignoring the defaults setting.

**Reproduces how often:**100%

#### Additional Information

This is really really annoying and wastes my time.  Don't make the user wait for the UI.  That is antithetical to what a good UI should do.  If they want to turn off the animations, then let them, don't ignore the system settings.

Thank you.

j-f1:
I’m not sure if the app is able to read from `NSUserDefaults` to check for this (undocumented) setting. However, it may be possible to add a separate preference in the app.

shiftkey:
You can explore system preferences from the main process. This returns `got 'false'` on my machine:

```
const { systemPreferences} = require('electron')
const value = preferences.getUserDefault('NSAutomaticWindowAnimationsEnabled', 'boolean')
console.log(`got '${value}'`)
```

[...]

PR #6620, opened 2019-01-14
Blankslate action transitions

niik:
## Overview

This adds enter and exit transitions for the remote action part of the no-changes blank slate (see #6445).

no-local-changes-transitions-hq

## Description

Prior to this PR all actions appearing or exiting did so immediately making it hard to spot when the result of one action (publish for example) was complete and replaced with another (i.e. create PR).

Whenever a remote action (push, pull, publish, create PR) appears the container hosting it will animate its height to fit the new contents after which the action will slide in from the right with a slight bounce effect. When that action is replaced by another the old action will animate with a slight bounce and slide off to the right.

## Release notes

no-notes

cc @desktop/design just FYI as we don't have a whole lot of animations in the app right now but we've been talking forever about using them more.

### DIFF
--- a/app/src/ui/changes/no-changes.tsx
+++ b/app/src/ui/changes/no-changes.tsx
@@ -93,6 +93,10 @@ interface IMenuItemInfo {
   readonly enabled: boolean
 }
 
+interface INoChangesState {
+  readonly enableTransitions: boolean
+}
+
 function getItemAcceleratorKeys(item: MenuItem) {
   if (item.type === 'separator' || item.type === 'submenuItem') {
     return []
@@ -136,12 +140,24 @@ function buildMenuItemInfoMap(
 }
 
 /** The component to display when there are no local changes. */
-export class NoChanges extends React.Component<INoChangesProps, {}> {
+export class NoChanges extends React.Component<
+  INoChangesProps,
+  INoChangesState
+> {
   private getMenuInfoMap = memoizeOne((menu: IMenu | undefined) =>
     menu === undefined
       ? new Map<string, IMenuItemInfo>()
       : buildMenuItemInfoMap(menu)
   )
+
+  private transitionTimer: number | null = null
+
+  public constructor(props: INoChangesProps) {
+    super(props)
+    this.state = {
+      enableTransitions: false,
+    }
+  }
 
   private getMenuItemInfo(menuItemId: MenuIDs): IMenuItemInfo | undefined {
     return this.getMenuInfoMap(this.props.appMenu).get(menuItemId)
@@ -555,6 +571,8 @@ export class NoChanges extends React.Component<INoChangesProps, {}> {
       <>
         <ReactCSSTransitionReplace
           transitionAppear={false}
+          transitionEnter={this.state.enableTransitions}
+          transitionLeave={this.state.enableTransitions}
           overflowHidden={false}
           transitionName="action"
           component="div"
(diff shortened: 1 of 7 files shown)


app/src/ui/changes/no-changes.tsx
```diff
@@ -577,6 +618,19 @@ export class NoChanges extends React.Component<INoChangesProps, {}> {
...
+  public componentDidMount() {
+    this.transitionTimer = window.setTimeout(() => {
+      this.setState({ enableTransitions: true })
+      this.transitionTimer = null
+    }, 500)
+  }
+
+  public componentWillUnmount() {
+    if (this.transitionTimer !== null) {
+      clearTimeout(this.transitionTimer)
```
iAmWillShepherd: Should `this.transitionTimer` also be set to `null` after the call to `clearTimeout`?

j-f1:
Unfortunately [this feature](https://crbug.com/722548) hasn’t made it to Chrome yet, but it’d be perfect for @iAmWillShepherd’s use case.

app/src/ui/changes/no-changes.tsx
niik: It could but there's no reason to since the component is being unmounted. An unmounted component will not be reused so that value won't either.

niik:
> Curious how this animation affects those on low power devices.

Idk. I can't say I see any difference between my macbook and this 2013 Thinkpad PC that I had lying around. Computationally speaking this transition isn't all that more complex than the spinning progress icons, or sliding notification bars we have, just slightly more pixels being shifted around and as long as there's a GPU available it'll be real snappy because the whole action will be lifted into a separate GPU layer.

> Is it possible to conditionally animate based off what we can figure out about the user's machine?

Chromium already does adapts the transition to match the frame rate that the computer can achieve. The complexities involved in boiling down a graphics rendering performance score is significant.

> Unfortunately this feature hasn’t made it to Chrome yet, but it’d be perfect for @iAmWillShepherd’s use case.

As far as I understand it, even that media query will be hinging on some user agent preference. We could certainly add a setting of our own to disable transitions but I want to emphasize that these transitions aren't simply eye candy. We've seen feedback about immediate transitions in other parts of the app causing confusion due to lack of user feedback (i.e. "wait, did that thing do something now?"). We will be exploring and adding more transitions and delays to address this. The exact transitions are certainly up for discussion and ripe to be polished and improved upon though.

Ultimately I think we should wait and see what feedback we get as we slowly make improvements to user feedback and adjust accordingly.

shiftkey:
xref an earlier discussion about animations #2768

shiftkey:
> Is it possible to conditionally animate based off what we can figure out about the user's machine?

Based on my quick testing, if you disable hardware acceleration using the `GITHUB_DESKTOP_DISABLE_HARDWARE_ACCELERATION=1` environment variable, this turns off these new animations (the elements immediately appear in-place).

niik:
> Based on my quick testing, if you disable hardware acceleration using the `GITHUB_DESKTOP_DISABLE_HARDWARE_ACCELERATION=1` environment variable, this turns off these new animations (the elements immediately appear in-place).

If that's true that's quite surprising to me. That env var should only disable hardware acceleration but it should fall back to good old fashioned CPU composition. Are you sure you're actually triggering the animations? Note that they don't run when the view is initially mounted, only when transitioning from one action to another (or from/to no action).

shiftkey:
> Are you sure you're actually triggering the animations? Note that they don't run when the view is initially mounted, only when transitioning from one action to another (or from/to no action).

Aha, it was related to the repositories setup. Disregard the previous comment then. I've been running with hardware acceleration disabled for a while (on macOS, not sure if that's important), and I was puzzled why I just wasn't able to trigger the animation. The environment variable was a [furphy](https://en.wikipedia.org/wiki/Furphy) and my repositories weren't in the ideal state to see it.

katesegrin:
@niik is this sized for social (1200x630px)? If not, could we get a version that is? 🙏

niik:
@katesegrin Here's a new one that's 1200x630px (1.8Mb)

action-transitions-hq-opt

PR #6621, opened 2019-01-14
A repository doesn't need to have a remote to be pushed

### LINKED ISSUE #6598: Publish button in new onboarding section does not activate
The Publish button in the new onboarding section is disabled, despite the top nav Publish button being active. I expect the button to be active.

Repro:
- Add new repo
- Try to click the blue Publish button (class="blankslate-action primary")

image

1.6.0-beta2, WIndows 10

niik:
Fixes #6598.

The 'push' menu item is used for pushing, publishing a branch, and also publishing a repository. Definitely confusing and something we should look into improving but for now it's blocking the release so I've changed the conditions under which the push menu item is disabled to not care about whether the repository has a remote or not.

The slight side effect of this is that the menu item label will still be 'push' and not 'publish repository' or something similar but I think that's a small price to pay for unblocking the release.

Clicking on the 'Push' menu item on a repository with no remote will trigger the publish flow.

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -154,8 +154,6 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
   let tipStateIsUnknown = false
   let branchIsUnborn = false
 
-  let hasRemote = false
-
   if (selectedState && selectedState.type === SelectionType.Repository) {
     repositorySelected = true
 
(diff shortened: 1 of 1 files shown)


shiftkey:
> The slight side effect of this is that the menu item label will still be 'push' and not 'publish repository' or something similar but I think that's a small price to pay for unblocking the release.

I think this bit is worth capturing, though I'm not sure how many people are using the menu to push versus the UI or the keyboard shortcut, so I suspect the impact is negligible right now...

PR #6622, opened 2019-01-14
[WIP] Custom tooltip experiment

ampinsk:
## Overview

We've been discussing the possibility of adding our own custom tooltips instead of using native tooltips lately. In some downtime, I tried getting it started, based mostly on [Primer tooltips](https://github.com/primer/primer/tree/master/modules/primer-tooltips). I ran into some trouble on disabled buttons (since disabled buttons use opacity, it was also dimming the tooltips) so this also restyles disabled buttons. This is definitely buggy, and there are probably a ton of ways to implement this, but this is just to get us started. If you want to try them, they work on the disabled merge button in the merge conflicts modal, and on the repo list.

**On repo list:**

screen shot 2019-01-14 at 2 25 29 pm

**On merge conflicts modal disabled button:**

screen shot 2019-01-14 at 2 25 08 pm

### DIFF
--- a/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
+++ b/app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx
@@ -371,6 +371,7 @@ export class MergeConflictsDialog extends React.Component<
               type="submit"
               disabled={conflictedFilesCount > 0}
               tooltip={tooltipString}
+              className="tooltip tooltip-s"
             >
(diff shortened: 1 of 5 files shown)


billygriffin:
Thanks @ampinsk! After discussing this with @outofambit, I'm going to close this PR as it's definitely something we're interested in but represents too much work that's not currently prioritized to get it to where it's broadly usable.

PR #6640, opened 2019-01-16
extract updateLicenseDump and promisify it

shiftkey:
## Overview

As part of #5876 I need to add a step to the `build.ts` process, which is already rather complex. Before doing that, this PR extracts the `updateLicenseDump` logic out to it's own module.

## Description

- `updateLicenseDump` was written in the old callback style, because the underlying `legal-eagle` module doesn't support promises
- this code was already isolated in it's own function, so we can easily move it out to it's own module
- we can now move `script/license-overrides.ts` into this folder to keep related things together
- this is one of the few async steps we have (we use synchronous `fs` APIs for other tasks), and we need to add another to support #5876 so making these properly chainable is the goal

## Release notes

Notes: no-notes


script/licenses/update-license-dump.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export function updateLicenseDump(
+  projectRoot: string,
+  outRoot: string
+): Promise<void> {
+  const appRoot = path.join(projectRoot, 'app')
+  const outPath = path.join(outRoot, 'static', 'licenses.json')
+
+  return new Promise((resolve, reject) => {
+    legalEagle(
```
j-f1:
```ts
import { promisify } from 'util'

const summary = await promisify(legalEagle)({
  path: appRoot,
  overrides: licenseOverrides,
  omitPermissive: true,
})
```

script/licenses/update-license-dump.ts
```diff
@@ -0,0 +1,60 @@
...
+import { getVersion } from '../../app/package-info'
+
+export async function updateLicenseDump(
+  projectRoot: string,
+  outRoot: string
+): Promise<void> {
+  const appRoot = path.join(projectRoot, 'app')
+  const outPath = path.join(outRoot, 'static', 'licenses.json')
+
+  let summary = await promisify(legalEagle)({
```
j-f1: Since you use this multiple times, it might be clearer to do this:

```ts
import * as legalEagle_ from 'legal-eagle'
const legalEagle = promisify(legalEagle)
```

(I would’ve mentioned this at first but I didn’t see the second usage 📱)

PR #2407, opened 2017-08-06
Open in External Editor, But Simpler™

shiftkey:
This is my proposed implementation of #2009, and should be simpler than #2039.

#### First Pass: Make It Work

 - [x] lookup to see if the user has each program installed
   - [x] Atom
   - [x] VSCode
   - [x] Sublime Text
 - [x] Preferences dialog listing available editors and 
 - [x] Persist this setting somewhere when it changes (localStorage?)
 - [x] Top-level menu item for "Open in External Editor"
   - [x] ensure this actually launches the selected editor

This should be enough to confirm we're on the right track.

#### Second Pass - polishing things

 - [x] top-level menu should show the editor name here as well
 - [x] Repository menu item for "Open in External Editor"
 - [x] performance - this lookup could be expensive - cache it for the session
 - [x] `macOS` should invoke shims instead of launching app
 - [x] docs about how this integration works
 - [x] on first launch, set a default editor - `await`ing to lookup of all editors adds +1s on Windows to [`AppStore.loadInitialState`](https://github.com/desktop/desktop/blob/db92f65aaac467e28ef76d5bdb4fb6b852e3a803/app/src/lib/dispatcher/app-store.ts#L803) when it fails - maybe i shouldn't do this
 - [x] when the editor doesn't exist, you need to go in to Preferences and manually switch to a different editor (the first one _looks_ like it's selected but lol no) for it to actually update. that's bananas.
 - [x] merge in #2413 and get "Open Preferences/Options" action to focus on Advanced tab
 - [x] find a faster registry implementation for Windows

### Screens and Visuals

With editors found:

- [x] Preferences view

With no editors found: 

 - [x] defaults to `Open in Atom`  - **should this be 'Open in External Editor' instead?**
 - [x] Preferences view should show "Go install Atom" link instead of `select`

 - [x] error dialog when you don't have any editors suggests you install Atom

- [x] error dialog when the editor can't be found, and you have to go to preferences and try again

Assorted other notes:

 - `app-path` [[repo]](https://github.com/sindresorhus/app-path) uses a precompiled binary. I'd rather that use `node-gyp` from a purist perspective, but we can PR that later unless we have strong feels.

PR #6643, opened 2019-01-16
Menu state race condition

niik:
## Overview

I spotted this race condition when I was developing the blank slate actions (#6445). Before https://github.com/desktop/desktop/pull/2407 we only ever created the application menu once; i.e at start up. We then sent IPC events from the renderer to the main process to signal which items should be enabled and which shouldn't.

In #2407, however, we started rebuilding the entire menu and replacing it rather than fiddling with updating the labels of individual menu items. This, unfortunately, led to a tiny race condition that hasn't been noticeable until now where whenever the renderer tells the main process to update labels of menu items the enabled/disabled state of menu items are cleared and reset to their default values until the renderer gets around to telling the main process that it wants to change the enabled/disabled state of some menu item.

The blank slate actions depends on the enabled/disabled state of menu items to determine whether actions (which are backed by menu items) should be enabled and the way I noticed the race condition was that I had another app in focus with the desktop app off to the side and saw that an action was disabled. When I then switched back to the desktop app there was a flicker of the button toggling from disabled to enabled and back again. This made it possible to click on a button that was supposed to be disabled by clicking directly on it when the app was blurred.

## Description

This PR solves the above-mentioned race condition by merging the updated labels with the existing enabled/disabled state of individual menu items before replacing the application menu.

Additionally this cleans up a helper method that we used to rely on which has now been implemented directly in electron itself.

## Release notes

Notes: no-notes.


iAmWillShepherd:
Went ahead and merged since the commit that dismissed this review was a spelling fix in a comment.

PR #6647, opened 2019-01-16
add regression test for node-sass migration

shiftkey:
## Overview

As a precursor to upgrading `node-sass` and resolving #5876, I want to ensure that the generated app bundle correctly transforms the SassScript variables we use in stylesheets.

## Description

This PR adds a step inside the build to look for any existing CSS files in the output. A successful build will look like this:

```
Checking all SASS variables have been encoded correctly…
  Checking stylesheet: out/crash.css…
  Checking stylesheet: out/renderer.css…
```

A failing test (for example, if you upgrade `node-sass` and forget to encode any SassScript variables) looks like this:

```
Checking all SASS variables have been encoded correctly…
  Checking stylesheet: out/crash.css…
  Checking stylesheet: out/renderer.css…
SASS variables were found in the generated stylesheets. This means some styles will not render as expected at runtime.
In file: /Users/shiftkey/src/desktop/out/renderer.css
 - Line 683: '  --color-new: $green;'
 - Line 684: '  --color-deleted: $red;'
 - Line 685: '  --color-modified: $yellow-700;'
 - Line 686: '  --color-renamed: $blue;'
 - Line 687: '  --color-conflicted: $orange;'
 - Line 688: '  --text-color: $gray-900;'
 ...
Look for the source of these styles under app/styles/ and ensure they are wrapped in a '#{}'.
Error verifying the Sass variables in the rendered app. This is fatal for a published build.
```

This new function scans the CSS for anything containing a `$` - I might tune this down the track to be a proper regex but for now this seems to catch what I need.

This logic is disabled for development mode because no CSS bundles are generated.

## Release notes

Notes: no-notes


shiftkey:
I have no idea what is happening with that `travis-ci/push` commit status, and why it is failing. Everything else is :gem:.

shiftkey:
I'd love some feedback from @desktop/engineering or @desktop/comrades on how I'm catching SASS errors here, or whether you have any bright ideas about alternative approaches...

shiftkey:
Merging this in so I can move #6661 into a state where it's ready for review

PR #6656, opened 2019-01-17
embed the SHA in the About view in development mode

shiftkey:
## Overview

This one a small change mostly for @tierninho's benefit when he's testing PRs locally.

Currently, if you run the app from source (`yarn build:dev && yarn start`) the About view shows the version (which likely was from the last release and not very accurate):

With this change, the About view now shows `Build` rather than `Version`, alongside the abbreviated commit SHA:

It still has the same "click to copy the version to clipboard" behaviour as before, so that you can quickly paste the build commit ID into an issue or PR when leaving feedback.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1284,12 +1284,14 @@ export class App extends React.Component<IAppProps, IAppState> {
           />
         )
       case PopupType.About:
+        const version = __DEV__ ? __SHA__.substr(0, 10) : getVersion()
+
         return (
           <About
             key="about"
             onDismissed={this.onPopupDismissed}
             applicationName={getName()}
-            applicationVersion={getVersion()}
+            applicationVersion={version}
             onCheckForUpdates={this.onCheckForUpdates}
             onShowAcknowledgements={this.showAcknowledgements}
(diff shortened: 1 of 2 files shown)


app/src/ui/app.tsx
```diff
@@ -1284,12 +1284,14 @@ export class App extends React.Component<IAppProps, IAppState> {
           />
         )
       case PopupType.About:
+        const version = __DEV__ ? __SHA__.substr(0, 10) : getVersion()
```
shiftkey: The `.substr` here is a rough approximation of `git rev-parse --short HEAD` (currently returning 9 characters):

```
$ git rev-parse --short HEAD
7159d344a
```

I _almost_ went with wiring up a Git operation to do this for maximum correctness, but it turns out the way we get `__SHA__` doesn't invoke Git at all:

https://github.com/desktop/desktop/blob/7159d344a026aae6aa8f85c89b130edf6acebbbe/app/git-info.ts#L49-L77

tierninho:
Nice touch, thanks! 💯

tierninho:
Tested and LGTM ✅

PR #6660, opened 2019-01-18
Only render NoChanges when there are.. no changes

### LINKED ISSUE #6658: "No local changes" window flashes every time when switching between changed files
## Version

* GitHub Desktop: 1.6.0

* Operating system: osx 10.14

## Steps to Reproduce

1. click on changed file
2. click on another changed file

### Expected Behavior

smooth change of displayed file in patch subwindow

### Actual Behavior

"No local changes" window opens and disappear before every file change.

## Additional Information

demo

niik:
## Overview

Apparently we've been using the `NoChanges` component as a fallback component while we're waiting for the Diff to load. It's been this way since forever but the last release with the redesigned no changes blank slate made it more noticeable, especially on slower environments.

This PR ensures that we're only rendering the `NoChanges` component when there are, in fact, no changes.

**Closes #6658**

## Release notes

Notes: `[Fix] Don't show the no local changes view when switching between changed files`

### DIFF
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -265,13 +265,13 @@ export class RepositoryView extends React.Component<
           />
         )
       } else {
-        if (selectedFileIDs.length === 0) {
+        if (selectedFileIDs.length === 0 || diff === null) {
           return null
         }
 
         const selectedFile = workingDirectory.findFileWithID(selectedFileIDs[0])
 
-        if (selectedFile === null || diff === null) {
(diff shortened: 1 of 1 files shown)


outofambit:
tried out locally and verified it fixes the issue ✅

tierninho:
Tested and LGTM. 

I did spot a similar "No File Selected" placeholder when scrolling History commits (both as a whole and individual files). Not as much as an eyesore, but still somewhat distracting. Shall I open a new issue?

image

PR #6661, opened 2019-01-18
upgrade node-sass, address SASS breaking changes

shiftkey:
## Overview

I believe this is the last native module that blocks upgrading us the Node used to build and test Desktop to Node 10 aka #5876. If you want more context the original PR #5644 is worth a read (lolcry).

## Description

Armed with the regression test in #6647, this PR is split up into two parts:

 - bump the module
 - find all the places where SassScript values bleed through into the bundled scripts

The thing I haven't really got a solution for here is the development time tooling to catch issues while working on styles.

Currently it's this sort of flow:

 - `yarn build:dev && yarn start`
 - "Oh, something doesn't look right"
 - Inspect Element in the Chrome Dev Tools
 - "Where's the colour gone?"
 - "Oh, right. \*That\* thing!"
 - Make the fix

cc @desktop/design for any bright ideas

 - ~~[ ] ponder improvements to dev-time tooling to catch things~~ punting this for now, it's an annoying problem and I don't have good hooks short of writing my own linting tools 👢 
 - [x] manually diff the CSS bundles before/after to sanity check
 - [x] test update styles on Windows (light theme)
 - [x] test update styles on Windows (dark theme)
 - [x] test update styles on macOS (light theme)
 - [x] test update styles on macOS (dark theme)
 - [x] test update styles on Linux (light theme)
 - [x] test update styles on Linux (dark theme)

## Release notes

Notes: no-notes

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -12,32 +12,32 @@
 $overlay-background-color: rgba(0, 0, 0, 0.4);
 
 :root {
-  --color-new: $green;
-  --color-deleted: $red;
-  --color-modified: $yellow-700;
-  --color-renamed: $blue;
-  --color-conflicted: $orange;
+  --color-new: #{$green};
+  --color-deleted: #{$red};
+  --color-modified: #{$yellow-700};
+  --color-renamed: #{$blue};
+  --color-conflicted: #{$orange};
 
-  --text-color: $gray-900;
-  --text-secondary-color: $gray-500;
-  --background-color: $white;
+  --text-color: #{$gray-900};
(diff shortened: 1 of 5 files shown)


shiftkey:
A quick test on macOS and Windows locally shows with this PR I can now build Desktop using Node 10.15.0 _or_ Node 11.7.0, so hopefully we can simplify our toolchain needs without much more work (both for contributors and for infrastructure) and get any performance-related improvements that have made it into Node 10+.

And as a reminder: this Node version is separate from the version of Node used in Electron, which has it's own upgrade process. You can follow along with #6391 for more details about that.

PR #6663, opened 2019-01-18
Initiate cloning by hitting enter on a clonable repository list selection

### LINKED ISSUE #6570: Clone repo dialog does not clone selected repo when Enter key is pressed
## Description
Pressing enter key on a selection in the Clone Repository dialog does not start cloning.

Related issues:
* #5234

## Version
* GitHub Desktop: 1.5.2-beta0
* Operating system: macOS 10.14.2

## Steps to Reproduce

1. Open repo dropdown list
1. Add -> Clone Repository...
1. Filter repo list (I used my username to return only my personal repos)
1. Use arrow keys to select a repository
1. Press <kbd>Enter</kbd>

### Expected Behavior
I expected the repository to be cloned

### Actual Behavior
Crickets...

## Additional Information

_It's not apparent here, but I am pressing <kbd>Enter</kbd> multiple times_

Daniel-McCarthy:
## Overview

**Closes #6570**

## Description

- Cloning can now be initiated from hitting enter on a repository list selection.

Previously if one selected a repository from the list and hit enter there would be no action. However, a user may expect or wish that they could simply hit enter upon selecting a repository to start cloning.

## Release notes

Notes: In the Clone Repository dialog it is now possible to activate cloning by hitting enter on a repository list selection.


billygriffin:
@Daniel-McCarthy Thanks! Should I label this `ready-for-review`?

Daniel-McCarthy:
@billygriffin That would be great! Thanks Billy!

tierninho:
Tested and over all works well, however I think we should prevent the `enter` button from cloning with there is an inline error message like so:

image

All in all, it should mirror the logic already attributed to `Clone` button, so if the clone button is active, enter should work.

Note: did not test Enterprise do to dev env restrictions.

Daniel-McCarthy:
@tierninho Ah, of course. I will address this soon. Thanks for the feedback!

Daniel-McCarthy:
I have now updated this to only attempt to initiate cloning if the same check that is used to enable the Clone button passes. However, this check is performed in the render function, and the result is not stored.

When this function is called, the check is done separately, and there is a duplication of code occurring. It may be a bit cleaner and easier to maintain if we branch this into it's own function that either returns a boolean directly or updates the state with the result.

However, this is certainly not strictly necessary.

tierninho:
@Daniel-McCarthy thanks for the change. I will let one of the core engineers chime in about above.

I do think we should add the same `enter` button functionality on the repo list load from the new onboarding page. (To get there, just remove all your repos.) Anyone have contradicting feels?

49655531-296fb000-fa3b-11e8-9076-c83955204246

billygriffin:
@tierninho Since @niik designed that page, I'd like to get his 👍 on the direction there.

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -62,6 +62,9 @@ interface ICloneGithubRepositoryProps {
    * available for cloning.
    */
   readonly onRefreshRepositories: (account: Account) => void
+
+  /** Initiate cloning if a repository and cloning path are selected and no current errors. */
+  readonly cloneIfCloningEnabled?: () => void
```
iAmWillShepherd:
```suggestion
  readonly cloneIfCloningEnabled: () => void
```

This can be made non-optional since this component is only used in 1 place.

Daniel-McCarthy: Excellent, thanks for the feedback! Since this no longer needs to be optional I went ahead and cleaned this up a bit to match the other prop function calls.

Daniel-McCarthy:
@niik Excellent, thanks for the feedback. These set of changes definitely make sense.

In regards to:

> I also think that we should avoid the duplication of the logic determining whether the dialog is able to initiate a clone or not as that's likely to diverge in the future causing us issues, we should have a single place (which is currently in the `renderFooter` method) for that.

For this, I could extract the logic from `renderFooter` into it's own function. We could also have a state variable that gets updated by this function, or return the result directly. Is this what you had in mind?

Thanks for your time, much appreciated!

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -62,6 +63,9 @@ interface ICloneGithubRepositoryProps {
    * available for cloning.
    */
   readonly onRefreshRepositories: (account: Account) => void
+
+  /** Initiate cloning if a repository and cloning path are selected and no current errors. */
+  readonly cloneIfCloningEnabled: () => void
```
iAmWillShepherd: **Non-blocking**
I think a name like `onRowKeyDown` does a better job describing what the purpose of the prop is. We're also using that name in other components.

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -95,4 +100,10 @@ export class CloneGithubRepository extends React.PureComponent<
       </DialogContent>
     )
   }
+
+  private onItemClicked = (repository: IAPIRepository, source: ClickSource) => {
+    if (source.kind === 'keyboard' && source.event.key === 'Enter') {
```
iAmWillShepherd: We could pull this logic out into `cloneIfCloningEnabled` in the parent component. That way, the parent is in control over what key triggers the event.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -616,6 +622,14 @@ export class CloneRepository extends React.Component<
     return url
   }
 
+  private cloneIfCloningEnabled = () => {
```
iAmWillShepherd: If we update the type signature here to accept a `ClickSource` we can move the check, `if (source.kind === 'keyboard' && source.event.key === 'Enter')` from the child component.

iAmWillShepherd: We could then rename this to function `onRowKeyDown`.

Daniel-McCarthy: Both changes sound great! Thanks again for the feedback.
I did not like the name `cloneIfCloningEnabled` much at all, so I'm more than happy to make that change ASAP. :+1:

tierninho:
@Daniel-McCarthy hey, was the `enter` key implemented when a repo is highlighted in the list? I do not see it working as of cb89cdc11230786da9f6351fca696eb91b309ed7

Daniel-McCarthy:
@tierninho Is that in reference to the Blank Slate? Could you by chance show me the area you are testing?

tierninho:
@Daniel-McCarthy If a user were to highlight one these repos and then hit the 'enter' button, the repo would be cloned.

image

Daniel-McCarthy:
@tierninho Oh okay, thanks for catching this! I will take a swing at implementing this.

Daniel-McCarthy:
@tierninho Just to clarify, when one clicks the repository, a clone button appears. If one then clicks the clone button in this view a window appears that shows where it will be cloned to and allows you to then confirm/cancel.

Is the expected behavior on a row item enter press to show the window that pops up as it would on button click? If so, I have this ready. I just wanted to make sure we are on the same page here.

tierninho:
@Daniel-McCarthy 

> Is the expected behavior on a row item enter press to show the window that pops up as it would on button click? If so, I have this ready.

Yes, that is correct. Is it in the current build? I didn't not see it implemented.

Daniel-McCarthy:
@tierninho Okay, great!

No, it's not in the current build (It will be very soon). I was on the same page for this change but at some point I had confused this clone flow with the `Clone a repository from the Internet...` button flow.

I have made the required changes for the flow you were testing and will push them now.

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -62,6 +63,12 @@ interface ICloneGithubRepositoryProps {
    * available for cloning.
    */
   readonly onRefreshRepositories: (account: Account) => void
+
+  /**
+   * Called when a key event occurs on a clone-able repository
+   * list item.
+   */
+  readonly onRowKeyDown: (source: ClickSource) => void
```
niik:
```suggestion
  readonly onItemClicked: (repository: IAPIRepository, source: ClickSource) => void
```

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -80,6 +87,7 @@ export class CloneGithubRepository extends React.PureComponent<
             filterText={this.props.filterText}
             onFilterTextChanged={this.props.onFilterTextChanged}
             onRefreshRepositories={this.props.onRefreshRepositories}
+            onItemClicked={this.onItemClicked}
```
niik:
```suggestion
            onItemClicked={this.props.onItemClicked}
```

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -95,4 +103,8 @@ export class CloneGithubRepository extends React.PureComponent<
       </DialogContent>
     )
   }
+
+  private onItemClicked = (repository: IAPIRepository, source: ClickSource) => {
```
niik: We should be able to remove this given that with the change in signature of the onItemClicked prop we can pass that straight to the underlying component.

app/src/ui/clone-repository/clone-github-repository.tsx
```diff
@@ -62,6 +63,12 @@ interface ICloneGithubRepositoryProps {
    * available for cloning.
    */
   readonly onRefreshRepositories: (account: Account) => void
+
+  /**
+   * Called when a key event occurs on a clone-able repository
+   * list item.
```
niik: I think we can just copy and paste the documentation from the `FilterList` component's `onItemClick` prop here.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -309,6 +315,7 @@ export class CloneRepository extends React.Component<
               onRefreshRepositories={this.props.onRefreshRepositories}
               filterText={tabState.filterText}
               onFilterTextChanged={this.onFilterTextChanged}
+              onRowKeyDown={this.onRowKeyDown}
```
niik:
```suggestion
              onItemClicked={this.onItemClicked}
```

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -616,6 +623,14 @@ export class CloneRepository extends React.Component<
     return url
   }
 
+  private onRowKeyDown = (source: ClickSource) => {
```
niik:
```suggestion
  private onItemClicked = (repository: IAPIRepository, source: ClickSource) => {
```

Daniel-McCarthy:
@niik That is absolutely not a problem at all. I'm more than happy to change it back. I will push the changes now. Thank you for the suggested changes as it certainly simplified figuring out what all needed to changed.

niik:
> That is absolutely not a problem at all. I'm more than happy to change it back. I will push the changes now. Thank you for the suggested changes as it certainly simplified figuring out what all needed to changed.

Thank you so much! And I beg your forgiveness for this having sat for so long, I think it's an excellent improvement and I'm happy to finally bring this in.

PR #6687, opened 2019-01-22
disable the app menu bar when there are no tracked repositories

### LINKED ISSUE #6669: App Menu not hidden/disabled when viewing "Let's get started" options
## Description

Installed the new Github desktop app and the visual bug appears if you hover the top part of the select repository window

## Version

* GitHub Desktop:

* Operating system:
Windows 10 Pro 64x latest version

### Expected Behavior

See blank white space
### Actual Behavior

I see a menu that does nothing

shiftkey:
## Overview

Fixes #6669 

## Description

On current `development` on Windows you can hover your cursor along the top and the app menu appears.

With this change, we use the same check for rendering the `BlankSlate` component to skip rendering the `AppMenuBar` (you'll have to imagine me hovering the cursor).

With the dark theme enabled, it no longer renders the app bar menu, which I spotted while investigating the issue:

## Release notes

Notes: ensure menu is not visible when in "Let's get started" view on Windows

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1001,6 +1001,11 @@ export class App extends React.Component<IAppProps, IAppState> {
       return null
     }
 
+    // Don't render the menu bar when the blank slate is shown
+    if (this.state.repositories.length < 1) {
+      return null
+    }
+
     const currentFoldout = this.state.currentFoldout
 
     // AppMenuBar requires us to pass a strongly typed AppMenuFoldout state or


shiftkey:
Tentatively flagging for the next hotfix release, but I could be persuaded to wait if the `priority-2` seems excessive

shiftkey:
@billygriffin what's extra confusing about the current behaviour is that the menu appears on hover, but you can't interact with it. So 👍 to ensuring the attention is on the "Let's get started" view.

jsejcksn:
v1.6.1: Is this the same bug or should I open a new issue?

GitHub Desktop: Menu bar items not visible

shiftkey:
@jsejcksn this will be available in 1.6.2 next week (currently on the beta channel) - no need to open another issue.

jsejcksn:
This was a terrible move for accessibility. Now I can't do anything without first cloning or creating a repo (unless there just happens to be a shortcut that I already know).

**At least show the `File` and `Help` menus.**

@shiftkey

billygriffin:
@jsejcksn Thanks for your thoughts! I'm curious to understand more about how/why this is bad for accessibility? Until you have a repository in Desktop, it's a useless app that you can't do anything with. We wanted to place this front and center (without being distracted by things that aren't yet useful for you) so it became super obvious that this is how you get going.

jsejcksn:
@billygriffin Accessibility is "the ability to access". If a menu cannot be accessed, then it reduces accessibility. Persistent menus expose functionality for both discovery and access. In the case of GitHub Desktop, the menu bar allows _access_ to the functions contained within, and _exposes_ the keyboard shortcuts which allow access to those functions within the app. (Keyboard access is yet another level of accessibility.)

I understand that there are some menu items that relate directly to interacting with repositories in the app, but that's certainly not the case for all of the menu items. **None** of the items in the `File` or `Help` menus interact with an existing repository in Desktop, so those items should still be accessible whether or not a repository is present in the app.

GitHub Desktop: File Menu

GitHub Desktop: Help Menu

billygriffin:
Hi @jsejcksn, thanks for sharing your thoughts on this. I'm of the opinion (at least at present) that in the welcome flow of the app, adding things outside of the flow of getting to the point where the app is meaningfully useful to people just introduces distraction. The menus have been disabled during the welcome flow since the app was launched and we haven't seen that as harmful. Though it's possible we're wrong, for now we're going to continue with the existing pattern, which puts the action that you need to take to get to using the app front and center. Thanks again.

jsejcksn:
@billygriffin It is meaningfully useful to be able to log out of my GitHub account or disable telemetry before further interacting with the app.

veeratcisco:
I downloaded the latest on Windows 10 and menu items are disabled. The issue is fixed?

shiftkey:
@veeratcisco the menu items are not shown when you are in this "Let's get started!" state. Once you add a repository you'll switch to the repository view (and see the app menu again).

PR #6695, opened 2019-01-23
move Dispatcher module into UI folder

shiftkey:
## Overview

As part of implementing the rebase flows I need to change the error handlers to trigger a new dialog when we detect the repository is in the middle of a rebase, so I was poking around in the `Dispatcher` again and realised this was a chance to address something that has been hanging around that influenced my earlier thinking in #5645.

## Description

As part of #5645 I set a goal for having `app/src/lib` to "contain(s) functions that do not depend on executing in a specific environment".

Why I think `Dispatcher` does not belong in this location comes down to these main reasons:

 -  it lives in `app/src/lib` but is importing functionality from modules in the `app/src/ui/` folder

```ts
import { executeMenuItem } from '../../ui/main-process-proxy'
import { installCLI } from '../../ui/lib/install-cli'
import { ApplicationTheme } from '../../ui/lib/application-theme'
```

 - the thing instantiating it is `app/src/ui/index.tsx`, which is bootstrapping the dispatcher, wiring up the error handling, and making it available to our UI.

https://github.com/desktop/desktop/blob/ba999a3b28a531ad128c55583c68736beb3d0d02/app/src/ui/index.tsx#L147-L158

 - it references stores like `StatsStore` and others indirectly though `AppStore`, which use `dexie` and require `IndexedDB`. This to me says "okay, all of these things are UI-specific" and the `Dispatcher` is a good first step because of how well-defined it's interface is.

Aside from the various React components importing it, the only other module that imports it is `app/src/lib/open-file.ts`, which is a wrapper around Electron's `shell.openExternal` to display an error if it fails to open a file. This also feels UI-specific, so I've moved that into `app/src/ui/lib/open-file.ts`.

The other important thing I did in here was to organize the imports using VSCode's inbuilt refactoring tools. This helped to make the dependencies in each module easier to navigate, especially for those complex components like `Dispatcher`.

## Release notes

Notes: no-notes


outofambit:
@desktop/engineering anyone else want to weigh in before i merge this tomorrow morning PT?

shiftkey:
Just rebased this on top of `development` to address some conflicts 🍎

PR #6696, opened 2019-01-23
Desktop 1.6.2-beta0

iAmWillShepherd:
Release branch for Desktop 1.6.2-beta0


shiftkey:
Once this goes green, is approved and merged, we can reload the changelog so the marketing site and app get the right information 🤘

Issue #6693, opened 2019-01-23
binary file merge conflicts detection broken


outofambit:
version: Desktop `1.6.0`, `1.5.1`

## Steps to reproduce:

1. create a `bothModified` conflict for an image file
(create image in base branch, then edit in a feature branch and base branch, merge)
2. take no action to resolve conflicts

## Current Result:

<img width="537" alt="screen shot 2019-01-23 at 7 07 46 am" src="https://user-images.githubusercontent.com/964912/51615872-c00a0a80-1edd-11e9-8e2f-c7ade182ee86.png">

## Desired Result:

Detect that this is a binary conflict and offer command line resolutions (or #6062)

## Notes

blocks #6062

hfsam:
对不起，我不明白!

PR #6697, opened 2019-01-23
add get binary files git function

outofambit:
## Overview

supports #6693 

## Description

- adds a utility function to get the currently modified binary files in a given repository
- the thinking here is that we'll be able to cross reference this with the output of status to help determine which files can have conflict markers and which ones can't

### DIFF
--- a/app/test/unit/git/diff-test.ts
+++ b/app/test/unit/git/diff-test.ts
@@ -22,6 +22,7 @@ import {
   getWorkingDirectoryDiff,
   getWorkingDirectoryImage,
   getBlobImage,
+  getBinaryPaths,
 } from '../../../src/lib/git'
 import { getStatusOrThrow } from '../../helpers/status'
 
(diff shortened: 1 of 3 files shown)


app/test/unit/git/diff-test.ts
```diff
@@ -384,4 +385,41 @@ describe('git/diff', () => {
...
   })
+
+  describe('getBinaryPaths', () => {
+    describe('in empty repo', () => {
+      let repo: Repository
+      beforeEach(async () => {
+        repo = await setupEmptyRepository()
+      })
+      it('returns an empty array', async () => {
+        expect(await getBinaryPaths(repo)).toHaveLength(0)
```
shiftkey: :art: a suggestion to avoid the `async`/`await` ceremony for these short tests:

```ts
expect(getBinaryFiles(repo)).resolves.someAssertion()
```

shiftkey:
@outofambit let me know if you'd like a quicker merge to unblock #6699

outofambit:
> let me know if you'd like a quicker merge to unblock #6699 

@shiftkey that would be great! thanks!

PR #6702, opened 2019-01-24
Fixed some LGTM alerts

imsolost:
Hello there. I was reviewing your project and I found some [alerts on LGTM.com](https://lgtm.com/projects/g/desktop/desktop/alerts/?mode=tree). I decided to fix a few of them, primarily relating to potentially inconsistent state updates when using this.state or this.props inside of setState.

I know this may not be the type of contribution you normally receive, but I figured I could help clean up the codebase.

(Full disclosure: I'm part of the team behind LGTM.com)

Let me know if you have any questions.


app/src/ui/changes/commit-message.tsx
```diff
@@ -139,10 +139,10 @@ export class CommitMessage extends React.Component<
 
     if (!shallowEquals(prevProps.commitMessage, this.props.commitMessage)) {
       if (this.props.commitMessage) {
-        this.setState({
-          summary: this.props.commitMessage.summary,
-          description: this.props.commitMessage.description,
-        })
+        this.setState((_state, props) => ({
+          summary: props.commitMessage.summary,
```
shiftkey: This one is blowing up because we previously checked `this.props.commitMessage` was not null.

I think a change like this is what's needed (which bypasses needing parameters inside `setState`):

```diff
@@ -138,10 +138,11 @@ export class CommitMessage extends React.Component<
     }

     if (!shallowEquals(prevProps.commitMessage, this.props.commitMessage)) {
-      if (this.props.commitMessage) {
+      const { commitMessage } = this.props
+      if (commitMessage) {
         this.setState({
-          summary: this.props.commitMessage.summary,
-          description: this.props.commitMessage.description,
+          summary: commitMessage.summary,
+          description: commitMessage.description,
         })
       } else {
         this.setState({ summary: '', description: null })
```

app/src/ui/preferences/preferences.tsx
```diff
@@ -110,16 +110,16 @@ export class Preferences extends React.Component<
     const availableEditors = editors.map(e => e.editor)
     const availableShells = shells.map(e => e.shell)
 
-    this.setState({
+    this.setState((_state, props) => ({
       committerName,
```
shiftkey: This is also erroring because we moved `committerName` and `committerEmail` into a different scope, so TSC doesn't trust that it's non-null. Here's a fix that works for me:

```diff
@@ -98,9 +98,6 @@ export class Preferences extends React.Component<
       }
     }
 
-    committerName = committerName || ''
-    committerEmail = committerEmail || ''
-
     const [editors, shells, mergeTool] = await Promise.all([
       getAvailableEditors(),
       getAvailableShells(),
@@ -110,16 +107,16 @@ export class Preferences extends React.Component<
     const availableEditors = editors.map(e => e.editor)
     const availableShells = shells.map(e => e.shell)
 
-    this.setState({
-      committerName,
-      committerEmail,
-      optOutOfUsageTracking: this.props.optOutOfUsageTracking,
-      confirmRepositoryRemoval: this.props.confirmRepositoryRemoval,
-      confirmDiscardChanges: this.props.confirmDiscardChanges,
+    this.setState((_prevState, props) => ({
+      committerName: committerName || '',
+      committerEmail: committerEmail || '',
+      optOutOfUsageTracking: props.optOutOfUsageTracking,
+      confirmRepositoryRemoval: props.confirmRepositoryRemoval,
+      confirmDiscardChanges: props.confirmDiscardChanges,
       availableShells,
       availableEditors,
       mergeTool,
-    })
+    }))
   }
 
   public render() {
```

app/src/ui/preferences/preferences.tsx
```diff
@@ -110,16 +107,16 @@ export class Preferences extends React.Component<
...
-    this.setState({
-      committerName,
-      committerEmail,
-      optOutOfUsageTracking: this.props.optOutOfUsageTracking,
-      confirmRepositoryRemoval: this.props.confirmRepositoryRemoval,
-      confirmDiscardChanges: this.props.confirmDiscardChanges,
+    this.setState((_state, props) => ({
+      committerName: committerName || '',
+      committerEmail: committerEmail || '',
+      optOutOfUsageTracking: props.optOutOfUsageTracking,
```
shiftkey: Just a heads up, these three props are now failing the linting because `tslint-microsoft-contrib` is not correctly detecting that `props.[x]` is a valid use of the prop. I tested this against the latest package and it's still happening, so that's either going to need some muting (short-term, unblock PR) or getting this fixed upstream (harder, but better for addressing more of these situations).

imsolost: Yep, it looks like it failing because the rule doesn't understand this usage of setState. I have opened an issue: https://github.com/Microsoft/tslint-microsoft-contrib/issues/796

Thanks for taking a look at this!

shiftkey: I was able to work around this in a74208d72 by using the props to set the initial state - which is correctly detected by our linting tools. We already did this with some props, so it was easy to justify doing it for these problem props. 

Let's see if that improves things at all.

imsolost: Great to hear. I'll check things on my end to see if that clears up the alerts.

app/src/ui/lib/path-text.tsx
```diff
@@ -374,13 +373,13 @@ export class PathText extends React.PureComponent<
     // by not measuring the actual width.
     if (availableWidth === 0) {
       if (this.state.length !== 0) {
-        this.setState({
-          ...this.state,
-          ...createPathDisplayState(this.state.normalizedPath, 0),
+        this.setState(prevState => ({
+          ...prevState,
```
shiftkey: Do we need to pass in `prevState` here as well? I thought `setState` supported passing in just the values that need to be updated...

shiftkey:
As this touches a number of components I'd like to queue this up for some testing of the affected components after `1.6.3` has been released, to ensure this functionality isn't regressed.

app/src/ui/lib/path-text.tsx
imsolost: You are correct, I have removed line 377 here.

shiftkey:
@imsolost thanks for opening this up, and apologies for taking so long to come back to this.

I've opened #7674 to use as a discussion for  how we want to incorporate LGTM and keeping on top of quality issues, so for the moment I plan to close out this PR for a couple of reasons:

 - I'd like us to co-ordinate better how to address this feedback alongside other things currently in flight
 -  I'd like a chance to discuss whether we want to properly integrate LGTM into the project, so we can better identify and handle things that it reports

If you're fine with that, please follow along with #7674...

PR #6707, opened 2019-01-24
Text diff rendering improvements

### LINKED ISSUE #658: Cursor fails to scroll view when performing partial commit selection
When selecting lines for a partial commit, you can only select the visible range.   I expected dragging with the cursor at the edge of the screen to scroll more lines into view.

To reproduce:
* Make a lot of changes to a file
* Make your window small enough where some of the lines you want to commit are off-screen.
* Hold down the mouse button and drag to begin selecting a range of lines near the edge of the view.  Continue dragging off the app and note that the view doesn't scroll. 

cc @Caged

### LINKED ISSUE #1551: Incorrect text selection box on long wrapped line
### Description

Text selection box appears across multiple lines when the selected line is long and wrapped

image

### Version

**GitHub Desktop version:** 0.5.3

**OS version:** 10.0.15063

### Steps to Reproduce

1. Select a long wrapped line
1. Check the selection box

**Expected behaviour:** The box shouldn't appear across the lines

**Actual behaviour:** It appears across the lines

### LINKED ISSUE #2677: Selecting a block of text in a diff shifts upon moving window pane
If you select a specific sentence in a diff, then expand or collapse the `file list` pane, the selected text will shift to another random snippet of text within that diff.

Ideally, the text selected initially should hold. (Note: moving any pane slider can reproduce this bug.)

0.8.1-beta4, mac

screen shot 2017-09-11 at 9 55 18 am
screen shot 2017-09-11 at 9 55 29 am

niik:
## Overview

This PR addresses a source of pain located in one of the older parts of the app, namely text diff rendering.

This is a huge PR and as such I'd be more than happy to pair with whoever ends up reviewing this and in the process spread some context around how text diff rendering works in desktop. I'll also try to give a thorough review of the work done in the description section below.

Fixes #1551
Fixes #658
Fixes #2677

## Description

I'll begin with a description of the current state of affairs before moving on to what this PR actually does. If you're not interested in the back story skip this next section.

### Story time

GitHub Desktop uses CodeMirror internally to render text diffs. Even though we don't use any of its editor capabilities (we always render in read-only mode) it's a really good fit for what we want given that it supports rendering large files with virtualization (or cell reuse if you're so inclined) in combination with handling wrapping lines (a much, much, harder problem than one would think). It also provides us with most of the infrastructure that we need in order to do [syntax highlighted diffs](https://github.com/desktop/desktop/pull/3101).

We've also had several problems with using CodeMirror but to be fair most of those have been because we haven't quite understood how to use it and not problems within CodeMirror itself. While there are some issues that are CodeMirror limitations, accessibility being a huge one, this PR addresses one particular mis-use on our part, namely gutters.

artboard copy

Our gutters differ from those of a normal editor in that the line number doesn't correspond to the line number in the editor, and we need to render two numbers, one for the line number in the previous version of the file and one for the line number in the current version (for deleted/added lines).

Our gutters are also highly interactive as opposed to most editor gutters out there as we use them to support partial committing.

When we initially created these gutters we were under the impression that it wasn't possible to leverage CodeMirror's built-in gutter system without creating all of the gutter elements up front so we opted instead to leverage CodeMirror's `renderLine` event and manually inserted a custom gutter into the line itself which we then styled to appear as if it was pinned to the left of the document. This, it turns out, wasn't our best move ever. Messing with the line in that way caused all sorts of problems, most notably that multi line selection was broken (see #1551).

We also tried shoehorning react into CodeMirror, with each line having its own `DiffLineGutter` component mounted into it. This led to us having to manually keep track of, and garbage collect, gutter instances which we never really got working fully leading to consoles sometimes looking like this

image

### The current state

I set out to try using the built-in gutter support that already exists in CodeMirror and switched us from using the `renderLine` event to `viewportChanged` which lets us insert gutter elements as lines are scrolled into view. I've also moved us away from intermingling react with CodeMirror, opting instead to simply create the DOM elements manually. They are simple enough and even if they weren't we gain so much by not having to keep track of what's mounted and not that we'd probably do it anyway.

I'm very happy with the end result. Not only is it performing really well but the impact on our code base is encouraging with much less code to contend with.

image

I was also able to move _most_ of the previously complicated hover state work out of typescript and into CSS, the exception being hunk hover states which still needs triggering from TS.

### The future

All this work was initiated by me investigating whether we could speed up the time it takes to switch between diffs in desktop. I started with a prototype which showed that we have an awesome opportunity to do so but that our less-than-ideal mixing of react and CodeMirror was throwing a wrench in the works for us so I put that work on pause (I'll open something up soon) and started to take a look at what it would take to address that. So once this PR lands we should be able to reap the benefits of that previous work and enable really fast diff switching.

## Testing

@tierninho What we're looking for here are bugs when selecting or deselecting lines in partial commits through one of the three methods of doing so (clicking on a single line, clicking on the wide gutter border to toggle an entire hunk, or clicking and dragging in the gutter). The way I've been doing it is by having the production app open side by side with the development app and making sure the behavior is the same between the two.

Some things won't be the same though and that's by design. Most apparent is that dragging beyond the last line of a scrolling diff will now scroll the diff. Another is that pressing down with the mouse on a line and then moving the cursor away from the gutter entirely and releasing no longer selects that line but reverts it to its previous state.

Another difference is that the area which triggers hunk selection now extends both right and left of the gutter border as opposed to only on the left side in the previous version. This new behavior is what was intended all along but it was missed when implementing it.

image

## Release notes

Notes:
```
[Fixed] Text selection in wrapped diff lines now allows selection of individual lines
[Fixed] Resizing the diff area preserves text selection range
[Improved] Moving the mouse pointer outside of the visible diff while selecting a range of lines in a partial commit will now automatically scroll the diff.
```


j-f1:
> Another difference is that the area which triggers hunk selection now extends both right and left of the gutter border as opposed to only on the left side in the previous version.

What do you think about making the entire `+`/`-` clickable for this?

tierninho:
Areas to be tested. Please let me know if I missed any. 

- [ ] Select/deselect lines in partial commit while clicking on single line
- [ ] Select/deselect lines in partial commit while clicking on the wide gutter border to toggle an entire chunk
- [ ] Select/deselect lines in partial commit while clicking and dragging in the gutter
- [ ] Resizing the diff area preserves text selection range
- [ ] User can select individual lines from wrapped diff lines 
- [ ] User can scroll the diff by moving the mouse pointer outside of the visible diff while selecting a range of lines in a partial commit
- [ ] test rendering of large diff

v1.6.x, mac

billygriffin:
Added this to 1.6.3 at the end of the month. I'd love to ship it to beta immediately after the release next week so we have it on beta for a full two weeks (or as long as possible) before shipping to prod.

niik:
> Areas to be tested. Please let me know if I missed any.

I think those are great areas @tierninho :heart: We'll wait to merge this until you've given a 👍

tierninho:
Completed my first round of testing. However, selecting part of a wrapped line and continuing to drag the cursor off the screen to the right causes the selection to flicker between two states as seen below. I think app cannot decide whether to show the highlighted area as-is or shift the selection to the next line.

Flickering state:
screen shot 2019-02-12 at 10 14 41 am

Actual state:
screen shot 2019-02-12 at 10 13 21 am

All is all, it seems to function well. I am glad this will be out in the (beta) wild for a bit as there are certain variations and subtleties that need to be tested individually by multiple users/OS.

niik:
> Completed my first round of testing. However, selecting part of a wrapped line and continuing to drag the cursor off the screen to the right causes the selection to flicker between two states as seen below. I think app cannot decide whether to show the highlighted area as-is or shift the selection to the next line.

Thanks @tierninho. I'm pretty sure this behaviour exists in the previous version as well, CodeMirror seems to have trouble when the cursor is moved out of the window while selecting. Since it seems to be a pre-existing condition I don't think we should hold up this PR because of it.

iAmWillShepherd:
Going to go ahead and merge since the only change since my last review is merging `development`.

PR #6713, opened 2019-01-25
Upgrade to Electron 3

### LINKED ISSUE #6391: Upgrade from Electron 2.x to Electron 3

iAmWillShepherd:
Fixes #6391

screen shot 2019-01-25 at 11 28 08 am

Electron 3 [release notes](https://electronjs.org/blog/electron-3-0)

TODO:

 - [x] rebase on top of `development` once #6799 has been merged
 - [x] upgrade to `spectron` `5.0.0` again to keep that in sync

Test plan:

 - [x] test app launches on macOS
 - [x] test API requests work on macOS
 - [x] test app launches on Windows
 - [x] test API requests work on Windows - @iAmWillShepherd tested on Feb 07, 2019 (Tested using clone repo, fetch, and pull requests; all seems in working order)
 - [x] test app launches on Ubuntu
 - [x] test API requests work on Ubuntu

### DIFF
--- a/app/.npmrc
+++ b/app/.npmrc
@@ -1,4 +1,4 @@
 runtime = electron
 disturl = https://atom.io/download/electron
-target = 2.0.13
+target = 3.1.3
 arch = x64
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -96,7 +96,10 @@ let isDuplicateInstance = false
 // We want to let the updated instance launch and do its work. It will then quit
 // once it's done.
 if (!handlingSquirrelEvent) {
-  isDuplicateInstance = app.makeSingleInstance((args, workingDirectory) => {
+  const gotSingleInstanceLock = app.requestSingleInstanceLock()
+  isDuplicateInstance = !gotSingleInstanceLock
+
+  app.on('second-instance', (event, args, workingDirectory) => {
     // Someone tried to run a second instance, we should focus our window.
     if (mainWindow) {
       if (mainWindow.isMinimized()) {
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
     "@types/webpack-merge": "^4.1.3",
     "@types/winston": "^2.2.0",
     "@types/xml2js": "^0.4.0",
-    "electron": "2.0.13",
+    "electron": "3.1.3",
     "electron-builder": "20.28.4",
     "electron-packager": "^12.0.0",
     "electron-winstaller": "2.5.2"
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "rimraf": "^2.5.2",
     "sass-loader": "^7.0.1",
     "semver": "^5.5.0",
-    "spectron": "4.0.0",
+    "spectron": "5.0.0",
     "stop-build": "^1.1.0",
     "style-loader": "^0.21.0",
     "to-camel-case": "^1.0.0",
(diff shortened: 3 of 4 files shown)


shiftkey:
For extra context, the big difference between 3.1.x and 3.0.x is outlined in [the first 3.1.0 beta](https://github.com/electron/electron/releases/tag/v3.1.0-beta.1):

> It's worth highlighting why this release is 3.1.0-beta.1 instead of 3.0.x despite the only changes being bugfixes. **Fixing `#15813` involved changing Electron to use Chromium's Mojo proxy resolver, which was a large enough change to warrant a new beta cycle instead of releasing straight to stable.** However this is the only large change relative to 3.0.x, so it's likely that the 3.1.0 beta cycle will be brief.

The original issue https://github.com/electron/electron/issues/13829 is related to network requests when a proxy involved, so we might get some benefits from this, but it's also important to test that our API calls are not affected, as well as ensuring the UI parts are not regressed by the Chromium upgrade.

This also upgrades the V8 included from 6.1 to 6.6, and I can see some notable improvements from the release notes around [background compilation](https://v8.dev/blog/v8-release-66#background-compilation), [asynchronous performance](https://v8.dev/blog/v8-release-66#asynchronous-performance-improvements) and some [language features](https://v8.dev/blog/v8-release-63#ecmascript-language-features) we might be able to use in the future, as well as performance and memory improvements

app/src/main-process/main.ts
```diff
@@ -96,7 +96,10 @@ let isDuplicateInstance = false
 // We want to let the updated instance launch and do its work. It will then quit
 // once it's done.
 if (!handlingSquirrelEvent) {
-  isDuplicateInstance = app.makeSingleInstance((args, workingDirectory) => {
+  const gotSingleInstanceLock = app.requestSingleInstanceLock()
+  isDuplicateInstance = !gotSingleInstanceLock
+
+  app.on('second-instance', (event, argv, workingDirectory) => {
```
shiftkey: 🎨 you could rename `argv` back to `args` and make the diff even smaller

app/.npmrc
```diff
@@ -1,4 +1,4 @@
 runtime = electron
 disturl = https://atom.io/download/electron
-target = 2.0.13
+target = 3.1.2
```
say25: 3.1.3?

shiftkey: Yep, good catch!

shiftkey:
One interesting thing I've noticed on macOS Mojave  with this build is that the **View** tab has these two new items (seemingly always greyed out):



If you're not sure what the Tab Bar is, here's Finder with "Show Tab Bar" enabled:



Desktop doesn't support this API, so I wonder if we're able to avoid rendering those menu items to make this clearer. I've opened #6842 to track this as it's not a blocker for this upgrade.

shiftkey:
Marked as `time-sensitive` because we're aiming to get an early beta out on Friday that upgrades us to Electron 3, so we can have a long QA period with it.

say25:
So when do we upgrade to Electron 4? :trollface:

shiftkey:
@say25 



But more seriously, there's been upstream things that we need to get in place before we can get there. 

Off the top of my head:

 - Node ABI mismatches between Electron 3 and 4 affecting native modules (this affected `keytar` but is now working with the latest version on NPM) https://github.com/lgeiger/node-abi/issues/54 has the context
 - Spectron support for Electron 4 is unclear: https://github.com/electron/spectron/issues/346

PR #6698, opened 2019-01-23
[rebase] detect and show a dialog containing what needs to be done

shiftkey:
## Overview

Fixes #6548

This PR has evolved to cover detecting a rebase conflict and letting the user continue or abort. It's feature-flagged for development only so the dialog won't be shown until we're ready to ship it.

## Description

The two flows I want to flesh out here are:

 1. detect a rebase with conflicts done externally to Desktop
  - app should show dialog
  - dialog should block while there are conflicts
  - user can abort rebase while underway
  - if no conflicts, let user continue rebase

2. detect when `git pull --rebase` raises an error (see #3422 for tracking issue)
  - error handling is invoked
  - dialog is shown
  - then user is in the flow from scenario 1

I'll follow up on some questions in the PR about the symmetry between Rebase conflicts and Merge conflicts as I'm thinking about them.

TODO:

- [x] when an error occurs that throws a `RebaseConflicts` error from dugite, show the dialog
- [x] when `git status` detects that a rebase is in progress, show a dialog with what needs to be done
- [x] show the list of files and their state
- [x]  the abort button so the user can bail out
- [x]  continue button is enabled when no more conflicts
- [x] need to stage conflicted files before continuing the rebase
- [x] disable `pull --rebase` blocker behind feature flag to enable further testing
- [x] add some tests
- [x] test flows and identify more interesting corner cases
 - [x] add banners for warning and success
- [x] show warning banner when dismissing dialog, warning banner lets you return to dialog
- [x] show success banner after successful rebase
- [x] rebase conflicts dialog shows base branch (if known) and target branch
- [x] extract follow-up tasks as part of these changes
- [x]  rebase on top of #6884 once that is merged

### Duplicated components

 - ~~`app/src/ui/rebase/conflict-files-list.tsx` has been extracted from `app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx` and could be turned into a reusable component - I don't think I've tweaked it too much, gonna wait and see what recent changes @outofambit has been making~~ moved to #6904
 - ~~`app/src/ui/rebase/rebase-conflicts-dialog.tsx` has been heavily influenced by `app/src/ui/merge-conflicts/merge-conflicts-dialog.tsx` but is likely to need a different flow (in particular to show progress while a rebase is occurring in the background) so I'm fine with keeping them separate for now rather than trying to unify them and be too clever~~ this is fine

## Release notes

Notes: detect and guide user through a rebase when encountered in the app

PR #6721, opened 2019-01-25
add Git tests for our current pull behaviour

shiftkey:
## Overview

This PR adds some scaffolding and tests for the `pull` Git operation, to provide some coverage before I dive in with rebasing (which will need similar scaffolding).

[**Crafting Git Tests docs**](https://github.com/desktop/desktop/blob/add-some-tests-for-pull/docs/technical/crafting-git-tests.md)

## Description

This PR came from the need to be able to programatically build up repositories to help with testing, instead of unpacking from an archive.

Why go down this path?

 - Archives of repositories are great for capturing regressions, but are essentially a black box of Git state
 - Updating archives over time is a pain (and also really hard to diff).
 - Building up the repository programatically in a test means we can adapt tests as our requirements and understanding of flows change.

An example of this process is the `app/test/helpers/repository-builder-pull-test.ts` function, which is responsible for setting up a repository that we use in the `pull` tests.

I've kept the scaffolding simple for now, but some examples of functions:

 - `makeCommit` - pass in a collection of files which are written to disk, staged and committed
 - `switchTo` - a quick way to ensure you're on the right branch (creates the branch if not found)

The rest of the changes were about testing how our `pull` implementation responds in these scenarios (with different configuration set):

 - ahead only (local commits not on remote)
 - behind only (remote commits not on current branch)
 - ahead and behind (commits both on remote and local that need to be reconciled)

TODO: 

 - [x] review merge conflict tests and see if they are a good candidate for the current API
 - [x] write some docs for scaffolding Git repository tests
 - [x] refine the API so it's good enough for a v1 ship

## Release notes

Notes: no-notes

### DIFF
--- a/app/test/unit/git/pull/ahead-and-behind-test.ts
+++ b/app/test/unit/git/pull/ahead-and-behind-test.ts
@@ -33,7 +33,7 @@ describe('git/pull', () => {
         entries: [
           {
             path: 'README.md',
-            value: Buffer.from('# HELLO WORLD! \n WORDS GO HERE! \nLOL'),
+            contents: '# HELLO WORLD! \n WORDS GO HERE! \nLOL',
           },
         ],
       }
@@ -45,7 +45,7 @@ describe('git/pull', () => {
         entries: [
           {
             path: 'CONTRIBUTING.md',
-            value: Buffer.from('# HELLO WORLD! \nTHINGS GO HERE\nYES, THINGS'),
+            contents: '# HELLO WORLD! \nTHINGS GO HERE\nYES, THINGS',
           },
         ],
       }
--- a/app/test/unit/git/pull/ahead-and-behind-test.ts
+++ b/app/test/unit/git/pull/ahead-and-behind-test.ts
@@ -1,0 +1,144 @@
+import {
+  fetch,
+  pull,
+  getAheadBehind,
+  revSymmetricDifference,
+} from '../../../../src/lib/git'
+import { Commit } from '../../../../src/models/commit'
+import { Repository } from '../../../../src/models/repository'
+import { createRepository } from '../../../helpers/repository-builder-pull-test'
+import {
+  cloneRepository,
+  makeCommit,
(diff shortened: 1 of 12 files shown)


shiftkey:
I've marked this as `ready-for-review` despite one open task because I'd love some input on [the docs](https://github.com/desktop/desktop/blob/add-some-tests-for-pull/docs/technical/crafting-git-tests.md) about crafting Git tests (which includes the new and old approaches) as well as feedback on the current scaffolding APIs - I've mostly focused on what I think was needed for the merge conflicts and upcoming rebase flows, so I've kept it narrow for now.

If there are additional patterns you'd like to propose, suggest them in this PR and we can extract tasks and figure out when to address them in the future.

app/test/unit/git/pull/ahead-and-behind-test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    describe('with pull.ff=only set in config', () => {
+      beforeEach(async () => {
+        await GitProcess.exec(
+          ['config', '--local', 'pull.ff', 'only'],
+          repository.path
+        )
+      })
+
+      it(`throws an error as the user blocks merge commits on pull`, async () => {
```
outofambit: unused async
```suggestion
      it(`throws an error as the user blocks merge commits on pull`, () => {
```

app/test/unit/git/pull/only-ahead-test.ts
```diff
@@ -0,0 +1,152 @@
...
+        await pull(repository, null, origin)
+
+        newTip = await getTipOrError(repository)
+      })
+
+      it('does not create new commit', async () => {
+        expect(newTip.sha).toBe(previousTip.sha)
+      })
+
+      it('is different to tracking branch', async () => {
```
outofambit:
```suggestion
      it('is different from tracking branch', async () => {
```

shiftkey:
Merging this in to unblock #6698

PR #6684, opened 2019-01-21
port linting to use renamed TS parser

shiftkey:
## Overview

This is the follow up to #6506 because it turns out that the TypeScript + ESLint tooling has undergone a transition to a new organization and consolidating work https://eslint.org/blog/2019/01/future-typescript-eslint

## Description

There's two new packages which are relevant to Desktop:

 - [`@typescript-eslint/parser`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser) - this enables ESLint to parse TS files and apply it's rules
 - [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin) - this is a collection of TS-specific rules that can be enabled for projects

I wanted this PR to upgrade both (replacing `typescript-eslint-parser` and `eslint-plugin-typescript`) but the latter package introduces a lot more churn (some of it new, others seem incorrect or unclear) so I'll extract those details to another issue. The second package also has a lot of rules that have come from `tslint`, so perhaps this is a chance for us to try and unify things under `eslint` (one of the areas [where the TypeScript team wants to invest in 2019](https://github.com/Microsoft/TypeScript/issues/29288)).

So this is just the minimum to get us over to the new parser and address the error you might be seeing when working locally:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: ~3.1.1

YOUR TYPESCRIPT VERSION: 3.2.2

Please only submit bug reports when using the officially supported version.

=============
```

I'll annotate the PR with more specific notes about the other changes.

## Release notes

Notes: no-notes

PR #6742, opened 2019-01-29
fail the CI build if modified files exist after installing packages

shiftkey:
## Overview

After merging #6684 and pulling into my local branch I noticed that `yarn` would leave the lockfile modified. It seems like I'd made a change to the `package.json` file without ensuring I'd committed the lockfile change too.

This PR adds a check to enforce that all CI builds that touch a `package.json` do not leave modifications in the working directory. If it does, the build will fail with a helpful message:

```
yarn run v1.13.0
$ stop-build
⚠️ there are 1 changed file
M yarn.lock
error Command failed with exit code 255.
```

TODO:

 - [x] see all CI builds fail with a helpful message
 - [x] fix the lock file

## Release notes

Notes: no-notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
             - ./node_modules
       - run:
           name: Lint and validate changelog
-          command: yarn lint && yarn validate-changelog
+          command: yarn lint && yarn validate-changelog && yarn check-modified
       - run:
           name: Build
           command: yarn build:prod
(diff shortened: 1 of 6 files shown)


shiftkey:
Marking as `time-sensitive` because this will probably affect anyone working off the `development` branch

shiftkey:
Friendly bump on this - I'd love to keep chipping away at ESLint/TSLint improvements when I need a break from rebasing, and this is currently a blocker (and also a good sanity check to catch whenever we upgrade packages).

Issue #1325, opened 2017-04-28
a better error message when pull occurs on non-existent upstream branch


tierninho:
I created a repo on dotcom with multiple branches, then cloned it on Desktop. I eventually deleted one of the branches on dotcom, and did a fetch on Desktop. 

As expected, Desktop asked me to publish this branch to dotcom. However, I did a pull on the branch while still in the same status, and I got this error:

`Your configuration specifies to merge with the ref 'refs/heads/tierninho-patch-1'
from the remote, but no such ref was fetched.`

Perhaps this can be rewritten in layman terms for the average end user to comprehend?

<img width="915" alt="screen shot 2017-04-28 at 3 13 22 pm" src="https://cloud.githubusercontent.com/assets/14828183/25549188/533a34c2-2c25-11e7-940c-9b6d3cf4a9e4.png">

v0.0.36, OS X 10.11.5

agisilaos:
👋Would love to start working on it. What's the best approach to start thinking about this issue?
I remember that many error messages aren't defined in the `desktop repo`. 🤔

shiftkey:
> I remember that many error messages aren't defined in the desktop repo. 🤔

@agisilaos they're defined over here - https://github.com/desktop/dugite/blob/master/lib/errors.ts

Adding a new `GitError` enum value, and a `GitErrorRegexes` entry to map the error message to the new `GitError` enum is all that needs to be done there (along with a regression test to validate the change). https://github.com/desktop/dugite/pull/175 is a good example to follow along with.

shiftkey:
Dropping `help wanted` from this issue because I want to call out a workflow that ended up being attempted in #6930 that we didn't expect:

**Along with showing a better error message, the dialog lets the user push the branch again to create it on the remote.**

This felt like a different context to how the user would trigger this problem - when trying to pull in changes from the upstream branch, which now doesn't exist - and a potentially different solution:

**Along with showing a better error message, the dialog lets the user stop tracking the upstream branch.**

Why this approach?

 - Branches can be cleaned up on the GitHub remote as part of normal usage - e.g. after a pull request is merged
 - Branches can also be manually removed through the UI
 - This feels like an opportunity to clean up locally, rather than restore the branch on the remote

In #6930 I suggested that the option here could be to `--unset-upstream` on the current branch, which will put the branch back to a `Publish Branch` state. This feels more explicit and less magical to the end user, if we can get the message right to explain why the remote branch no longer exists.

[...]

PR #6755, opened 2019-01-31
show an actionable dialog when files would be overwritten

mathieudutour:
## Description

Shows an actionable dialog when files would be overwritten during a checkout. The action of the dialog is "Discard Changes and Checkout". I was also thinking about having a "Stash" action but there is no notion of stash in Desktop yet.

## Release notes

Notes:
- Show an actionable dialog when files would be overwritten during a checkout

### DIFF
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -264,6 +264,10 @@ function getDescriptionForError(error: DugiteError): string {
       return 'A lock file already exists in the repository, which blocks this operation from completing.'
     case DugiteError.NoMergeToAbort:
       return 'There is no merge in progress, so there is nothing to abort.'
+    case DugiteError.NoExistingRemoteBranch:
+      return 'The remote branch does not exist.'
+    case DugiteError.LocalChangesOverwritten:
+      return 'Some of your changes would be overwritten.'
     default:
       return assertNever(error, `Unknown error: ${error}`)
   }
--- a/app/src/ui/dispatcher/error-handlers.ts
+++ b/app/src/ui/dispatcher/error-handlers.ts
@@ -359,3 +359,67 @@ export async function upstreamAlreadyExistsHandler(
 
   return null
 }
+
+/**
+ * Handler for when we attempt to checkout a branch and there are some files that would
+ * be overwritten.
+ */
+export async function localChangesOverwrittenHandler(
+  error: Error,
+  dispatcher: Dispatcher
+): Promise<Error | null> {
+  const e = asErrorWithMetadata(error)
+  if (!e) {
+    return error
+  }
+
+  const gitError = asGitError(e.underlyingError)
+  if (!gitError) {
+    return error
+  }
+
+  const dugiteError = gitError.result.gitError
+  if (!dugiteError) {
+    return error
+  }
+
+  if (dugiteError !== DugiteError.LocalChangesOverwritten) {
+    return error
+  }
+
+  const { repository, retryAction } = e.metadata
+  if (repository == null) {
+    return error
+  }
+
+  if (!(repository instanceof Repository)) {
+    return error
+  }
+
+  if (!retryAction) {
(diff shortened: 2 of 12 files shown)


app/src/lib/git/core.ts
```diff
@@ -264,6 +264,10 @@ function getDescriptionForError(error: DugiteError): string {
       return 'A lock file already exists in the repository, which blocks this operation from completing.'
     case DugiteError.NoMergeToAbort:
       return 'There is no merge in progress, so there is nothing to abort.'
+    case DugiteError.NoExistingRemoteBranch:
```
say25: Pretty sure I added this to Dugite, I havent gotten around to testing that it works but this PR might also close out #1325

shiftkey: @say25 I'm happy to leave that open because I think it requires more context and/or a more friendly message, like what I did with #6610

shiftkey:
@mathieudutour I think this needs a merge from `development` to get the fixes from #6756 and get CI working again

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -359,3 +359,67 @@ export async function upstreamAlreadyExistsHandler(
...
+
+  // find the overwritten files from the error
+  const overwrittenFiles = []
+
+  const pathRegex = /^\t(.*)/gm
+  const { stderr } = gitError.result
+
+  let match = pathRegex.exec(stderr)
+
+  while (match) {
```
iAmWillShepherd: I'm not sure how this condition is ever **not** met. From my understanding, match is of type `Array` and the `while` loop exits when some condition is `false`, I don't think an empty array will get coalesced into `false`, so it would always evaluate to true. 

Additionally, match is being set to itself every-time on L414 (identical to L410).

mathieudutour: It returns `null` if it doesn't match https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec, not an empty array

iAmWillShepherd: > And a regex is stateful so match is not set to itself but to the next match in the stderr

That's the part I was missing!

iAmWillShepherd:
```suggestion
  while (match !== null) {
```

iAmWillShepherd:
I am not getting the dialog with the action you noted

**Reproduction steps**
1. Add new file to `master`
1. Create new branch, `b1`, based on `master`
1. Checkout `master` and make commit 
1. Make a commit (moves `master` 1 commit ahead of `b1`)
1. Checkout `b1`
1. Add a new file 
1. Checkout `master` 

I expected the new dialog to give me the option to discard but got the old error dialog. Here's a nice gif of it in action.

mathieudutour:
Ah that's because of https://github.com/desktop/dugite/pull/281. The proper dialog should show up when the files are already tracked and will be overwritten.

Updating the dugite dependency when it's published will take care of showing the dialog when files are untracked and will be overwritten

shiftkey:
@mathieudutour @iAmWillShepherd I have just published a new `dugite` release which contains that PR - https://github.com/desktop/dugite/releases/tag/v1.83.0

shiftkey:
@mathieudutour are you able to include a screenshot of the dialog in action, for reference?

iAmWillShepherd:
> @mathieudutour are you able to include a screenshot of the dialog in action, for reference?

iAmWillShepherd:
> @mathieudutour are you able to include a screenshot of the dialog in action, for reference?

Updated PR body with gif.

mathieudutour:
ah forgot about that, thanks for updating the PR!

shiftkey:
@iAmWillShepherd anything in particular holding up this PR from being merged?

shiftkey:
I think this has been approved for long enough. Thanks for the contribution @mathieudutour!

PR #6756, opened 2019-01-31
Fix the Azure build

iAmWillShepherd:
The build config suddenly broken, so here's an attempt to fix it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,7 +25,7 @@ jobs:
         displayName: 'Publish Test Results'
         inputs:
           testResultsFiles: '**junit*.xml'
-          testRunTitle: TestRun ${{ variables['Agent.OS'] }}
+          testRunTitle: $(Agent.OS) Test Results
       - task: PublishCodeCoverageResults@1
         displayName: 'Publish code coverage results'
         inputs:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         displayName: 'Publish Test Results'
         inputs:
           testResultsFiles: '**junit*.xml'
-          testRunTitle: TestRun ${{ parameters.name }} $(node_version)
+          testRunTitle: TestRun ${{ variables['Agent.OS'] }}
```
shiftkey: I think what's happening is that `${{ variables['Agent.OS'] }}` is resolving to an empty string (which is fine) but then the test results are appending a number to each `TestRun` file so that they are unique:



I don't believe `variables` isn't the right value to use here because these seem to represent values defined in the current configuration. I found [this example](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azdevops#using-other-repositories) to illustrate this:

```yaml
# Repo: Contoso/BuildTemplates
# File: common.yml
parameters:
  vmImage: 'ubuntu 16.04'

jobs:
- job: Build
  pool:
    vmImage: ${{ parameters.vmImage }}
  steps:
  - script: npm install
  - script: npm test
```

I think this needs to be written a bit differently to get the right value, based on [this example](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azdevops&tabs=yaml%2Cbatch#parameters-to-yaml-statements):

```yaml
testRunTitle: $(agent.os)
```

shiftkey:
There might be a chance to refine this but for now I'm gonna merge this to unblock other builds affected by the change:

PR #6758, opened 2019-02-01
Use standard link styling on conflicts dialog

ampinsk:
This removes some (seemingly?) unnecessary styling on the link in the conflict resolution dialog that made the link style not pass color contrast in dark mode.

|Before|After|
|---|---|
|screen shot 2019-01-30 at 10 50 23 am|screen shot 2019-01-31 at 2 59 22 pm|


agisilaos:
Had the same idea @outofambit! It's a good call.

outofambit:
done!

PR #6769, opened 2019-02-04
address tslint deprecated warnings because they never go away in a new VSCode install

shiftkey:
## Overview

Getting my new VSCode installation setup, and after installing the recommended extensions I start to see `tslint` warnings. Cool and helpful. The downside to this is that they don't go away, so I constantly need to workaround this list whenever I'm looking at warnings/errors:

I'm not sure why CI doesn't catch these, but the changes were specific and unimportant enough that I just went in and addressed them. I'll revisit this area when I'm next looking at how we can simplify/unify the `eslint`/`tslint` tooling we have.

## Release notes

Notes: no-notes

### DIFF
--- a/script/build.ts
+++ b/script/build.ts
@@ -120,6 +120,7 @@ function packageApp() {
     )
   }
 
+  /* tslint:disable:deprecation */
   const toPackageArch = (targetArch: string | undefined): packager.arch => {
     if (targetArch === undefined) {
       return 'x64'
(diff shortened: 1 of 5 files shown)


j-f1:
You could use `@typescript-eslint/eslint-plugin-tslint` to run everything at once.

shiftkey:
> You could use `@typescript-eslint/eslint-plugin-tslint` to run everything at once.

This is something that interests me, but for the moment I just want to unblock the problem I'm facing.

Once I've gotten the rest of the `@typescript-eslint` transition done, I'll evaluate whether we need a standalone `tslint` to exist.

PR #6771, opened 2019-02-04
finish migration to using @typescript-eslint packages

shiftkey:
## Overview

This is the second part of #6684 and completes the transition to using the new `@typescript-eslint` packages. 

 - [x] investigate linting errors
 - [x] rewrite history to make it easier to follow

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/highlighter/globals.d.ts
+++ b/app/src/highlighter/globals.d.ts
@@ -1,4 +1,4 @@
-/* eslint-disable typescript/interface-name-prefix */
+/* eslint-disable @typescript-eslint/interface-name-prefix */
 
 declare namespace CodeMirror {
   interface EditorConfiguration {
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -33,8 +33,8 @@ rules:
     - functions: false
       variables: false
       typedefs: false
-  ## blocked by https://github.com/nzakas/eslint-plugin-typescript/pull/23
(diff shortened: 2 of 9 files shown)


shiftkey:
@j-f1 I haven't had a chance to dig into this but `383 problems` after upgrading to `1.2.0` of the three `@typescript-eslint` packages has me nervous. Any clues about whether an earlier version might not have this issue?

app/src/highlighter/globals.d.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable typescript/interface-name-prefix */
+/* eslint-disable @typescript-eslint/interface-name-prefix */
```
j-f1: What do you think about getting rid of this rule? Seems like the community has moved away from the `I` prefix.

shiftkey: I've grumbled about this before, but I'd prefer to discuss this outside the scope of this PR and review our use of interfaces overall. I'll open up an issue to braindump my thoughts about this later today.

shiftkey: #6773

shiftkey:
Requesting a review from @desktop/engineering as this now a blocker on some community PRs like #6912

outofambit (approved):
tried this out locally, and seems good!

PR #6791, opened 2019-02-07
Pull request metrics

### LINKED ISSUE #6772: [Metrics] Create pull request from Desktop
We would like to understand how frequently people are creating pull requests today using GitHub Desktop in order to establish a baseline prior to incorporating "Create PR" as one of the top-level suggested next steps in the "no local changes" view. Note: This is a subset of #6714 

I'd like this to be on production for at least two weeks prior to the "create PR" suggested next step shipping, so ideally this would go into next week's release (1.6.2).

Question - Will we need to instrument it separately with the menu item, keyboard shortcut, and suggested next step? Or do we care the pathway folks take and just want to add one overall metric?

Basically, I'd love for us to be able to see whether incorporating it as a prominent suggestion materially increases the % of people using the PR workflow via Desktop.

cc: @niik and anyone else who wants to weigh in

iAmWillShepherd:
Fixes #6772

Adds new metric to capture the number of times a user is taken to the create PR page on GitHub.com from Desktop.

We initially attempted to differentiate between the different pathways to get to this action, but because @niik's implementation of the suggested next step "Create PR" action looks to be to just hijack the menu item action, the easiest path forward here was just to have one that encompasses the three different ways to trigger it (suggested next step, keyboard shortcut, branch menu item).

---
- [x] Open issue in analytics
- [x] Update central
- [x] Update marketing site
- [x] Update `metrics.md` doc - https://github.com/desktop/desktop/pull/6791/commits/a94f799d626a86e0b53ef937d855bee3e43a9344

### DIFF
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -1422,7 +1422,7 @@ export class Dispatcher {
   }
 
   /**
-   * Increments the `CreatePullRequestCount` metric
+   * Increments the `createPullRequestCount` metric
    */
   public recordCreatePullRequest() {
     return this.statsStore.recordCreatePullRequest()
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -1425,14 +1425,7 @@ export class Dispatcher {
    * Increments the `CreatePullRequestFromMenuCount` metric
    */
   public recordCreatePullRequestFromMenu() {
-    return this.statsStore.recordCreatePullRequestFromMenuCount()
-  }
-
-  /**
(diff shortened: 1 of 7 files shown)


billygriffin:
Added `ready-for-review` here because the checklist is complete.

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -1421,6 +1421,13 @@ export class Dispatcher {
     return this.statsStore.recordDivergingBranchBannerInitatedMerge()
   }
 
+  /**
+   * Increments the `CreatePullRequestCount` metric
```
shiftkey:
```suggestion
   * Increments the `createPullRequestCount` metric
```

shiftkey:
Waving this PR through because I know it's unrelated to whatever #6793 is doing

Issue #6793, opened 2019-02-07
investigate uptick in failing integration tests on Travis/Appveyor


shiftkey:
Opening this up to brain dump so it's not just in my head

**Travis CI**:

 - First known occurrence: https://travis-ci.org/desktop/desktop/builds/489742692 ?
 - Pull Requests associated with change: #6776 and #6791
   - https://travis-ci.org/desktop/desktop/builds/490232532
 
**AppVeyor**:

  - First known occurence: https://ci.appveyor.com/project/github-windows/desktop/builds/22206985
  - PRs associated with change: #6776 and #6777 and #6664
     - https://ci.appveyor.com/project/github-windows/desktop/builds/22211836
     - https://ci.appveyor.com/project/github-windows/desktop/builds/22207813

Other infrastructure: unclear

**Problem behaviour**:

 - `jest` runs to execute launch test
 - `spectron` initializes using build application bundle
 - test fails because of a timeout to launch the app

```
$ cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe --auto-servernum -- jest --config ./app/jest.integration.config.js
FAIL app/test/integration/launch-test.ts (12.399s)
  App
    ✕ opens a window on launch (10008ms)
  ● App › opens a window on launch
    Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.
      at mapper (../node_modules/jest-jasmine2/build/queue_runner.js:41:52)
  ● App › opens a window on launch
    TypeError: Cannot read property 'isVisible' of undefined
      40 | 
      41 |   it('opens a window on launch', async () => {
    > 42 |     await app.client.waitUntil(() => app.browserWindow.isVisible(), 5000)
         |                                                        ^
      43 | 
      44 |     const count = await app.client.getWindowCount()
      45 |     expect(count).toBe(1)
      at Object.app.client.waitUntil (test/integration/launch-test.ts:42:56)
      at Timer.tick (../node_modules/webdriverio/build/lib/utils/Timer.js:111:31)
      at Timer.start (../node_modules/webdriverio/build/lib/utils/Timer.js:75:22)
      at new Timer (../node_modules/webdriverio/build/lib/utils/Timer.js:62:14)
      at Object.<anonymous>.exports.default (../node_modules/webdriverio/build/lib/commands/waitUntil.js:25:17)
[...]

PR #6799, opened 2019-02-08
downgrade spectron to known good version for Electron 2.x

shiftkey:
## Overview

This is related to #6793 but I don't believe is the full resolution to why some builds are quietly failing their integration tests.

## Description

`spectron` needs to be updated at the same time Electron is updated, and the [README](https://github.com/electron/spectron#version-map) shows how they relate:

screen shot 2019-02-08 at 3 12 02 pm

I jumped ahead to `5.0.0` in #6518 because I assumed it'd be backwards compatibile with Electron 2 and I didn't see any issues at the time, but I suspect that's not accurate and is at least partly to blame for the issues we're seeing with the integration tests.

## Release notes

Notes: no-notes

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ environment:
 
 cache:
   - .eslintcache
-  - node_modules
+  - node_modules -> package.json
   - '%USERPROFILE%\.electron'
   - '%LOCALAPPDATA%\Yarn'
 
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -8,7 +8,7 @@ environment:
 
 cache:
   - .eslintcache
-  - node_modules
+  - node_modules -> package.json
```
shiftkey: This is the syntax Appveyor supports for saying "cache `node_modules` based on the contents of `package.json`" and this should help prevent cases like #6793 which I believe are down to cache invalidation not working as expected.

shiftkey:
@outofambit I believe this will help with #6800 - feel free to cherry-pick the changes onto that branch

PR #6802, opened 2019-02-08
Stops node_modules from being cached during CI builds

iAmWillShepherd:
Addressing the recent build issues we've been having due to cache invalidation.

**Previous build times (range)**
* Azure: 12-14 mins
* Appveyor: 13-16 mins
* Travis: 10-12 mins
* Cicle: 12-14 mins

☝️ is given as ranges because I couldn't figure out how to export build history in order to do give better descriptive stats. Also all these times are from the past week.

**After this change**
* Azure: 11m44s
* Appveyor: 17m30s
* Travis: 11m41s
* Cicle: 12m34s

Since these times are within the ranges I saw in our build history for successful builds, removing the cache doesn't seem to have made a significant impact.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
             - vendor/bundle
             - .eslintcache
             - ~/.electron
-            - ~/Library/Caches/Yarn/v1
+            - ~/Library/Caches/Yarn/v4
       - run:
           name: Linting
           command: yarn lint && yarn validate-changelog && yarn check-modified
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,6 @@ jobs:
             - .eslintcache
             - ~/.electron
             - ~/Library/Caches/Yarn/v1
-            - ./node_modules
       - run:
           name: Linting
           command: yarn lint && yarn validate-changelog && yarn check-modified
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,6 @@ environment:
 
 cache:
   - .eslintcache
-  - node_modules
   - '%USERPROFILE%\.electron'
   - '%LOCALAPPDATA%\Yarn'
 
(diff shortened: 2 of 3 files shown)


shiftkey:
@iAmWillShepherd are you able to add into this PR the change in time between this build and a recent successful build on `development`? It'd be nice to know roughly how much this slows this down.

.circleci/config.yml
```diff
@@ -32,7 +32,6 @@ jobs:
             - .eslintcache
             - ~/.electron
             - ~/Library/Caches/Yarn/v1
```
shiftkey: I just had a this path on my new machine and I only have a `v4` key. Maybe this needs to be updated?

```
$ ls ~/Library/Caches/Yarn/
v4
```

iAmWillShepherd: yea, I think you're right since I got the same result.

shiftkey:
> * Appveyor: 13-16 mins
> ...
> * Appveyor: 17m30s

This for me is the one I worry about - it's already the slowest build, and this PR makes it even slower compared to the others.

What if we left the change in #6799 in place for Appveyor only?

appveyor.yml
```diff
@@ -8,7 +8,6 @@ environment:
 
 cache:
   - .eslintcache
-  - node_modules
   - '%USERPROFILE%\.electron'
   - '%LOCALAPPDATA%\Yarn'
```
shiftkey: Related to the previous ppoint about where `yarn` uses for a local cache:

```suggestion
   - '%LOCALAPPDATA%\Yarn\Cache\v4'
```

shiftkey: On my Win10 machine I see the `v2` folder has ~800MB of packages, so I think making this explicit will help speed up our Windows builds

shiftkey:
@iAmWillShepherd I'm 👍 to getting this in and monitoring the build times over the `1.6.3` release process. Let me know if there's anything else you plan to do in this PR.

iAmWillShepherd:
Nope, this should be good. 

I went ahead and remove the work-around I was using to get CI to pass in a971a98c94c6027107fef6917249f4e4056d33ad

shiftkey:
Merging this in early to unblock some other CI work

PR #6803, opened 2019-02-08
improvements to integration test execution

shiftkey:
## Overview

As part of #6793 I've found a bunch of ways to improve the integration tests run. It doesn't quite solve the issues we've seen (#6802 is related) but this should help with identifying any future issues.

## Description

- [x] change launch test to use packaged app
- [x] improve `beforeEach` and `afterEach` so `jest` will exit cleanly when the `spectron` test fails, meaning tests are no longer held up until some sort of timeout occurs
- [x] remove usage of `.resolves` inside test because we can't seem to rely on these working
- [x] restore `yarn test:review` for CircleCI to see more information about failing integration tests internals
- [x] add `yarn test:review` to Pipelines config to see more information about failing integration tests internals
- [x] rebase on top of #6802 
- [ ] use the current set of tools to understand why CircleCI is failing

## Release notes

Notes: no-notes

PR #6815, opened 2019-02-10
publish test/coverage results even if the tests report errors

shiftkey:
## Overview

This PR ensures that our CI platforms always do a `yarn test:report` to publish any test results to CodeCov. Azure Pipelines has it's own way to publish test results, but the principle is the same.

For example, [this CircleCI build](https://circleci.com/gh/desktop/desktop/7806) will "Report to CodeCov" when the tests pass.

But [this CircleCI build](https://circleci.com/gh/desktop/desktop/7817) from #6803 returns exit code `1` inside `yarn test` and it will instead skip "Report to CodeCov" and instead check for details about the test failure (added as part of that PR) before cleaning up:

The goal here is to ensure we upload test and code coverage results whether or not the build succeeds. If we're only uploading test results when they pass, we'll only ever be at 100% tests passing and lack insight into what tests are failing over time.

## Release notes

Notes: no-notes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,11 @@ install:
 
 script:
   - yarn lint && yarn validate-changelog && yarn check-modified && yarn
-    build:prod && yarn test:setup && yarn test && yarn test:report
+    build:prod && yarn test:setup && yarn test
+
+after_success:
+  - yarn test:report
 
 after_failure:
+  - yarn test:report
   - yarn test:review
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,7 +40,6 @@ test_script:
   - ./script/test-appveyor.bat
 
 on_success:
-  - yarn test:report
   - yarn run publish
 
 on_finish:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,10 +38,11 @@ build_script:
 test_script:
   - yarn test:setup
   - ./script/test-appveyor.bat
-  - yarn test:report
 
 on_success:
+  - yarn test:report
   - yarn run publish
 
 on_finish:
   - yarn test:review
+  - yarn test:report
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -45,7 +45,11 @@ install:
 
 script:
   - yarn lint && yarn validate-changelog && yarn check-modified && yarn
-    build:prod && yarn test:setup && yarn test && yarn test:report
+    build:prod && yarn test:setup && yarn test
```
shiftkey: The problem here is that we're using `&&` here to only continue when the previous command exited cleanly. This is great to ensure we fail fast, but we should run `test:report` whether or not the tests passed.

appveyor.yml
```diff
@@ -38,10 +38,11 @@ build_script:
 test_script:
   - yarn test:setup
   - ./script/test-appveyor.bat
```
shiftkey: This wrapper script sets the `APPVEYOR_TEST_RESULT` environment variable (for reasons I can't recall) so that Appveyor will behave like the other platforms rather than continue on to the next step of the script in the event of a non-zero exit code.

appveyor.yml
```diff
@@ -38,10 +38,11 @@ build_script:
...
-  - yarn test:report
 
 on_success:
+  - yarn test:report
```
outofambit: i don't think we need this line, since `on_finish` is [always called](https://www.appveyor.com/docs/build-configuration/#build-pipeline) (on success and failure)

outofambit (changes requested):
one small change requested, but this is great!

appveyor.yml
shiftkey: Good spot! Fixed in efdfb2891.

outofambit (approved):

Tunminkhaing:
Tunminkhaing

PR #6818, opened 2019-02-11
Hide White Space in Diff

### LINKED ISSUE #1940: Provide "ignore whitespace" option when viewing diff
### Description

There should be an option to ignore whitespace changes when comparing file versions in the diff view.

### Version

**GitHub Desktop version:** Chocolate-Covered Yaks (3.3.4.0) 50415df

**OS version:** Microsoft Windows [Version 10.0.14393]

### Steps to Reproduce

1. Look at a diff
1. Notice you didn't actually change much in that file, just added/removed some linebreaks etc.
1. Feel the need to ignore those uninteresting changes
1. Note that there's no option for this very basic diffing task

**Expected behaviour:** Have checkbox "ignore whitespace" that, when checked, causes the diff view to show "real" changes only

**Actual behaviour:** No such checkbox

#### Additional Information

Despite intensive googling I couldn't find any info whether such an option exists, so I assume it doesn't. It would be a VERY handy feature.

say25:
## ~WIP Notes~

~Needs to be rebased on https://github.com/desktop/desktop/pull/8045 or after its merged.~

## Overview

**Closes #1940**

## Description

- Adds a "Hide Whitespace" Checkbox to both the `Changes` and `History` tabs

### Changes Tab

changes

### History Tab

- Checkbox has moved to the right since filming this `.gif`

historytab

## Notes to Reviewer

1. Removed feature flag usage for HideWhitespaceInDiff in existing locations
1. Put new functionality behind the same `HideWhitespaceInDiff` feature flag.
1. I am currently using one user setting for HideWhitespaceInDiff in both changes and history.
1. Fixed numerous typos along the way.

## Release notes

Notes: Adds a "Hide Whitespace" Checkbox to both the `Changes` and `History` tabs


shiftkey:
Tagging this one as `needs-design-input` to grab some :eyes: and 💭 from @desktop/design about where/how we should display this checkbox.:

screen shot 2019-02-12 at 6 11 48 pm
screen shot 2019-02-12 at 6 12 05 pm

I'm not sure we have a good example of this in the app currently, so this is a nice time to ponder on it a bit.

app/src/lib/git/diff.ts
```diff
@@ -120,8 +124,10 @@ export async function getCommitDiff(
     args.push(file.status.oldPath)
   }
 
+  const argsToUse = args.filter(arg => arg !== '')
```
shiftkey: :art: rather than needing to parse like this, you could just conditionally insert the `-w` at the right point:

```diff
diff --git a/app/src/lib/git/diff.ts b/app/src/lib/git/diff.ts
index d33a9620f..ca5d09cc9 100644
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -101,22 +101,23 @@ export async function getCommitDiff(
   commitish: string,
   hideWhitespaceInDiff: boolean = false
 ): Promise<IDiff> {
-  const whitespaceArg = hideWhitespaceInDiff ? '-w' : ''
-
   const args = [
     'log',
     commitish,
-    whitespaceArg,
     '-m',
     '-1',
     '--first-parent',
     '--patch-with-raw',
     '-z',
     '--no-color',
-    '--',
-    file.path,
   ]
 
+  if (hideWhitespaceInDiff) {
+    args.push('-w')
+  }
+
+  args.push('--', file.path)
+
   if (
     file.status.kind === AppFileStatusKind.Renamed ||
     file.status.kind === AppFileStatusKind.Copied
@@ -124,10 +125,8 @@ export async function getCommitDiff(
     args.push(file.status.oldPath)
   }
 
-  const argsToUse = args.filter(arg => arg !== '')
-
   const { output } = await spawnAndComplete(
-    argsToUse,
+    args,
     repository.path,
     'getCommitDiff'
   )
```

tierninho:
💯on design input.

I did spot a potential conflict. If the feature is enabled and the user has a file with two empty lines (add one, deleted one), the file "disappears" like so:

before
image

after
image

say25:
Totally agree on needing design input. I asked for input earlier on the issue but it must have gotten lost in the shuffle at some point. Hopefully we can flush out a solution that can elegantly solve the issue for Desktop users in this PR 😃 .

shiftkey:
@tierninho raises an excellent point about how this potentially affects partial commits.

I haven't tested this out for myself, but by enabling `-w` for the Changes tab I suspect you're not going to be able to make partial commits because the diff it shows is different to what Git is expecting when you apply a patch. I also worry about hiding the whitespace there makes the diff not reflect the reality of the working directory state, but we can dig into that later.

shiftkey:
Just poking at this branch to get it back to a state where it's passing CI, as that appears to be one of the reasons it's blocked currently...

app/src/lib/stores/app-store.ts
```diff
@@ -4167,6 +4180,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  public _setHideWhitespaceInDiff(
+    hideWhitespaceInDiff: boolean,
+    repository: Repository,
+    file: CommittedFileChange | null
+  ): Promise<void> {
+    setBoolean(hideWhitespaceInDiffKey, hideWhitespaceInDiff)
+    this.hideWhitespaceInDiff = hideWhitespaceInDiff
+
+    if (file === null) {
+      return this.updateChangesDiffForCurrentSelection(repository)
```
shiftkey: Looks like `AppStore.updateChangesDiffForCurrentSelection` was renamed to `updateChangesWorkingDirectoryDiff` in #7313, but that might not be the entire work required to update here.

say25:
@shiftkey, I'll get this guy building tonight. Hoping that post `1.7.0` we can get this prioritized

ampinsk:
@outofambit and I chatted about this! We agreed that we actually think the current checkbox is working. The only change I'd suggest is moving the checkbox on the commit view over to the right:

checkbox

We thought about different toggle styles, different placement, and putting it in a dropdown, but thought this was the clearest and simplest way to go about it 👍

say25:
Sounds great @ampinsk I’ll run with that for now. I think one of the gotchas called out about is what should happen for files that only have white space changes?

We could just display text explaining it like we do for binary files

ampinsk:
> We could just display text explaining it like we do for binary files

I think that makes sense!

say25:
This should be reviewable from a functionality standpoint.

shiftkey:
@say25 I was having a play around with how this change affects partial commits, and while I was able to create a partial commit with this enabled I noticed that the checkbox to toggle showing/hiding whitespace stopped working afterwards.

Here's a GIF of it in action:

partial-commit-with-ignore-whitespace

The steps I used to trigger this:

 - make some changes to a file that will be affected by "Hide whitespace"
 - put the diff in "Hide whitespace" mode
 - select a line and commit it
 - try and disable "Hide whitespace" to see the full diff

say25:
@shiftkey correct me if I am wrong but after your commit “add one line” you seem to only have changes that do not include white space remaining?

shiftkey:
@say25 I went to try and get a more detailed repro, but couldn't trigger it again. I guess that's on me.

However, I was able to trigger this error which confirmed we're gonna have some pain if we let users try and partially commit with this enabled:



I was working from f24790c2f678b150f8ca498287a40152da0459c2 but any recent commit should work.

Here's the changes I had in the working directory at the time:

```diff
diff --git a/.prettierignore b/.prettierignore
index 5d0aa7e50..a992e7450 100644
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,16 +1,17 @@
-out/
+      out/ 
 dist/
 vendor/
+*.json
+*.md
+
 tslint-rules/*.js
 npm-debug.log
-yarn-error.log
+    yarn-error.log
 .DS_Store
 .awcache
 .idea/
-.eslintcache
+    .eslintcache
 app/coverage
 app/static/common
 app/test/fixtures
 gemoji
-*.json
-*.md
```

Before switching to "Hide whitespace" here's the diff:



Then I "Hide whitespace":

Screen Shot 2019-05-03 at 4 19 25 PM

And I choose the first line to trigger this error:



If you choose the second line it'll create the patch successfully, so I hope it's not too much work to support "Hide whitespace" and partial commits.

say25:
@shiftkey I started taking a look at this. Got a key question, would the expected behavior be to apply all whitespace changes in addition to selected partial diffs when in this state?

shiftkey:
> Got a key question, would the expected behavior be to apply all whitespace changes in addition to selected partial diffs when in this state?

Not sure I understand the question, but I'll try and explain a bit more about the partial commits side of things as we've implemented it and maybe this will help clarify things:

We use the `formatPatch` function inside [`app/src/lib/patch-formatter.ts`](https://github.com/desktop/desktop/blob/development/app/src/lib/patch-formatter.ts) to generate a patch that's compatible with Git from the current diff. It works out which changes have been selected by the user, and combines them with the diff context to generate a patch file.

This patch file is whitespace-sensitive, and I suspect by using a diff which ignores whitespace changes we have the potential to generate a patch that Git isn't able to apply correctly in some situations to the working directory (because the working directory still has those whitespace changes).

I suspect we'll need to add some tests to [`app/test/unit/patch-formatter-test.ts`](https://github.com/desktop/desktop/blob/74b67591694638c7904c5764103470d00bcb27e2/app/test/unit/patch-formatter-test.ts) now that `getWorkingDirectoryDiff` supports disabling whitespace to catch and verify that we're doing things correctly.

Let me know if that helps clarify things at all.

say25:
1. If I have both whitespace and non-whitespace changes in a file and I commit the entire file, should my whitespace changes also be committed?

    Does having the hide whitespace change the answer to the above?

1. If I have both whitespace and non-whitespace changes in a file and I commit only part of the file, should my whitespace changes also be committed?

    Does having the hide whitespace change the answer to the above?

shiftkey:
1.  My gut feeling is that, if you've specifically set "Hide whitespace in diff" and then choose to commit the file, it'd be unexpected to also see whitespace changes in the commit. You probably have "Hide whitespace in diff" enabled for viewing the commit locally, but what about when you push that commit to GitHub? 

2. Like the previous point, you're looking at a specific diff and choosing specific lines - everything else should remain in your working directory. Other lines in this diff are either not visible (because they were all whitespace) or omitted (because you chose to exclude them) so it doesn't feel very polished for the app to present one thing but then commit additional changes.

jkuss:
My opinion, whitespace changes should always be committed. The checkbox is a view state only. 

If you have whitespace only changes in a file and you click to view that file the, file pane should be empty with a message saying "Whitespace only changes in this file *Show whitespace changes*" and the link unchecks the whitespace box.

lourdes1504:
Excelente trabajo,  en conjunto podemos lograr mejores resultados.

billygriffin:
@say25 We had a chance to catch up about this one. And while we're not positive we'll end up shipping this, if you're interested in continuing on it we're certainly open to it depending on how it feels in a more finished state. 

We're trying to figure out a way to make it explicit what will be committed while still allowing people to view the changes without being bogged down by whitespace changes. How would you feel about elegantly disabling partial committing while whitespace is hidden, or potentially disabling committing altogether? 

@ampinsk If we were to disable partial committing while whitespace is hidden, do you have thoughts on how we should let users know why they're unable to select or deselect parts of files?

say25:
> How would you feel about elegantly disabling partial committing while whitespace is hidden, or potentially disabling committing altogether?

That seems like it would meet my needs and address my workflow. I would prefer to disable partial committing but we could do that as an iteration?

Assuming you and @ampinsk are good with it I can try to finish it out. Been a touch busier at work than normal and haven't been able to address @tierninho's concerns on another PR quite yet.

ampinsk:
I'm team disable partial committing while hiding whitespace changes! I think we can just indicate that with a native tooltip `Partial committing is not available while hiding whitespace changes` 👍

outofambit:
@say25 will you please @ me or request a review from me when you're ready? thanks!

say25:
@outofambit. I think this is ready. Didn’t want to bother your weekend

app/src/ui/diff/text-diff.tsx
```diff
@@ -207,11 +208,11 @@ const defaultEditorOptions: IEditorConfigurationExtra = {
...
     // custom search UI.
-    [__DARWIN__ ? 'Cmd-F' : 'Ctrl-F']: showSearch,
+    'CmdOrCtrl+F': showSearch,
     // Disable all other search-related shortcuts so that they
-    // don't interfer with global app shortcuts.
-    [__DARWIN__ ? 'Cmd-G' : 'Ctrl-G']: false, // findNext
-    [__DARWIN__ ? 'Shift-Cmd-G' : 'Shift-Ctrl-G']: false, // findPrev
+    // don't interfere with global app shortcuts.
+    'CmdOrCtrl+G': false, // findNext
+    'Shift+CmdOrCtrl+G': false, // findPrev
```
outofambit: nice little cleanup :)

app/src/ui/history/commit-summary.tsx
```diff
@@ -360,22 +357,20 @@ export class CommitSummary extends React.Component<
             </li>
             {this.renderTags()}
 
-            {enableHideWhitespaceInDiffOption() && (
```
outofambit: thanks for cleaning this up, too!

outofambit (commented):
overall this looks good @say25! thanks again for sticking with us through this.

i'm wondering about the feature flag changes in this PR. is the intent to repurpose the `enableHideWhitespaceInDiffOption()` for this new checkbox location (the non-history one?). it looks like this PR feature flags the code that recomputes the changes without whitespaces, but not the checkbox in the UI on the diff selection page. (i tested this locally by setting `enableHideWhitespaceInDiffOption()` to return `false`)

say25:
@outofambit 
> Is the intent to repurpose the `enableHideWhitespaceInDiffOption()` for this new checkbox location?

The short answer is yes I'm not sure if you want it behind a toggle or not. In reality the existing flag should be named `enableHideWhitespaceInHistoryOption()` but obviously when I originally had these as one PR well we didnt rename it.

jess3691 (dismissed):

niik:
Hey @say25, seems like there's a fair amount of changes (mostly spelling fixes) unrelated to the core feature here. Would you be able to pull those out of this PR and into a separate PR so that we can more easily reason about the changes here?

The unfortunately typo in `divergingBranchBannerInitatedMerge` is one we'll have to live with or else we'll have to update our metrics backend and queries and that doesn't feel worth it IMO. If you want you can rename all instances save for the one in `IDailyMeasures`, that way the methods calls won't have the typo at least and then we can add a comment to the property in `IDailyMeasures` about why it's the way it is.

say25:
@niik Will do :D

alchaki2009 (dismissed):

say25:
@niik Rebased this following the merge of #10617

6068 (approved):

app/src/lib/stores/app-store.ts
```diff
@@ -2196,7 +2199,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const diff = await getWorkingDirectoryDiff(
       repository,
-      selectedFileBeforeLoad
+      selectedFileBeforeLoad,
+      this.hideWhitespaceInDiff
```
sergiou87: I think this should also be gated behind the feature flag, right?
```suggestion
      enableHideWhitespaceInDiffOption() && this.hideWhitespaceInDiff
```

app/src/ui/diff/index.tsx
```diff
@@ -248,6 +249,7 @@ export class Diff extends React.Component<IDiffProps, IDiffState> {
   private renderTextDiff(diff: ITextDiff) {
     if (
       enableExperimentalDiffViewer() ||
+      enableHideWhitespaceInDiffOption() ||
```
sergiou87: What's the reason for this check here? 🤔

app/src/ui/diff/diff-options.tsx
```diff
@@ -71,12 +71,23 @@ export class DiffOptions extends React.Component<
...
+  }
+
+  private onDocumentMouseDown = (event: MouseEvent) => {
+    const { current: ref } = this.diffOptionsRef
+    const { target } = event
+
+    if (ref !== null && target instanceof Node && !ref.contains(target)) {
+      this.closePopover()
+    }
+  }
```
sergiou87: Is this a leftover after rebasing? In theory I got rid of this in a [previous refactor](https://github.com/desktop/desktop/pull/11571/files#diff-2c0fe2274b8c57c0fbfa6c5763627f504e43bf4e59526f131582b456d517197aL84) 😳

app/src/ui/diff/diff-options.tsx
```diff
@@ -139,9 +150,6 @@ export class DiffOptions extends React.Component<
   }
 
   private renderHideWhitespaceChanges() {
-    if (this.props.hideWhitespaceChanges === undefined) {
-      return null
-    }
```
sergiou87: I think we still need some kind of check, using also the feature flag, to prevent this checkbox to be shown in the Changes tab when the feature flag is not enabled. Is that right? 🤔

sergiou87 (changes requested):
Looks great overall, thank you!! 😍 

I just found a couple of issues that need to be addressed, but I can't wait to get this merged!!!

app/src/ui/history/commit-summary.tsx
```diff
@@ -374,24 +375,23 @@ export class CommitSummary extends React.Component<
             )}
 
             {enableSideBySideDiffs() && (
-              <>
-                <li
-                  className="commit-summary-meta-item without-truncation"
-                  title="Split View"
```
say25, line 380: This title doesn't make sense, it controls both Split View and Whitespace

Screen Shot 2021-03-25 at 12 31 11 AM

app/src/ui/history/commit-summary.tsx
```diff
@@ -374,24 +375,23 @@ export class CommitSummary extends React.Component<
             )}
 
             {enableSideBySideDiffs() && (
-              <>
```
say25, line 377: removed bogus Fragment wrapper

app/src/ui/history/commit-summary.tsx
sergiou87: Nice catch!

app/src/ui/changes/changed-file-details.tsx
```diff
@@ -44,6 +45,7 @@ export class ChangedFileDetails extends React.Component<
 
         {enableSideBySideDiffs() && (
           <DiffOptions
+            sourceTab={RepositorySectionTab.Changes}
```
sergiou87, line 48: I think just passing a `boolean` would've been enough, and we avoid importing (coupling) stuff unnecessarily. But since this is only used for a temporary feature flag check and we're only importing an enum, I'm not particularly worried about it 😄

sergiou87 (commented):
LGTM, thanks for the latest changes! 😄 

I tested it and it works fine, but I have some UX concerns:
- Changing the `Hide whitespace changes` setting in one tab (`Changes` or `History`) affects the other one too. I think I could be ok with having only one shared setting for both tabs, except…
- …this PR turns the `Hide whitespace changes` setting into a **destructive action**: it will wipe your partial changes selection if it exists.
I think that requires us to add a prompt if the user has partial status, so that they confirm they want to go ahead enabling this setting and clearing the partial selection.
And once we add that prompt, I think it'd be super weird to see it when you enable this setting from the `History` tab, don't you think?

I think this PR is good enough for beta, but requires to address these rough edges before shipping it to production.

sergiou87 (approved):
It's happening! 🚀

sergiou87:
Thank you for your hard work and patience @say25 !!! We really appreciate it ❤️

billygriffin:
+1 Thanks @say25! ❤️ 

To recap our discussion in slack for posterity, we want to make two small changes to the way this works before it ships to production.

1. Because of @sergiou87's observation above that changing this selection can be destructive when there are parts of files selected, we should disable this checkbox when subsets of files are selected, and inform people via a tooltip in the same way we do in the opposite direction (disabling partial committing when hiding whitespace).
2. Decouple the changes setting for hiding whitespace from the history setting. Even though there's precedent for these to be coupled (split diffs), we think they're different enough domains that they probably shouldn't be coupled together. For example, in (1) above, we're disabling the setting if there are partial sets of files selected, but it shouldn't follow that you also shouldn't be able to hide whitespace when looking at a diff in history under this set of circumstances. Dotcom also treats hiding whitespace as a completely ephemeral state, which begs the question of how long the setting should persist. Since we don't know what people's assumptions will be about this, we're going to ship it to beta with these changes and then see how it feels and what feedback we get.

For reference, here's how dotcom handles it:

Screen Shot 2021-03-25 at 2 48 34 PM

PR #6837, opened 2019-02-13
[rebase] extract functions so they can be shared

shiftkey:
## Overview

A small refactoring to extract these conflict-related functions into a module they can be used in the rebase flow.

These functions are helpful to identify what conflicts work needs to be done, and we have a module for grouping these functions together. I believe I'll need the same rules for the rebase flow, based on my initial prototype and testing, and this felt like a good bit of tidy-up while I was in there.

## Release notes

Notes: no-notes


shiftkey:
Marking as `time-sensitive` because this is a blocker on the rebase conflicts detection in #6698

PR #6830, opened 2019-02-12
[rebase] add Git internals to support rebase flow

shiftkey:
## Overview

Extracting the Git changes from #6698 so they can be reviewed independent of the UI changes.

## Description

 - [x] new feature flag for rebase flows
 - [x] omit `pull.rebase` when working in development mode
 - [x] tests for `pull.rebase` config flag
 - [x] Git operations for starting and continuing the rebase
 - [x] tests for detecting conflicts in a rebase
 - [x] review and unify `stageManualConflictResolution` in it's own module

## Release notes

Notes: no-notes

Issue #6550, opened 2019-01-08
metrics for rebase feature


shiftkey:
This is a placeholder issue for discussing what metrics we want to roll out as part of this feature. There's probably some prior art from the Merge Conflicts feature that we can mirror here.

cc @desktop/product @desktop/analytics

shiftkey:
@desktop/product @desktop/analytics friendly nudge on what metrics should be in for 1.6.3, versus what can be deferred to 1.7.0

scottypate:
Deferring that decision to @iAmWillShepherd. I would lean on the side of making sure what we have is accurate prior to focusing on new metric additions.

billygriffin:
@scottypate I view that project as more of a thing that's happening in parallel but shouldn't stop us from continuing to measure the things we're shipping.

@shiftkey For 1.6.3, I think we ought to measure:

1. Number of times a user pulls with rebase (potentially as a % of regular pulls if that's possible?)
2. Number of times a user encounters a conflict while pull rebasing.
2a. Number of times a pull rebase is completed successfully after a conflict.
2b. Number of times a pull rebase is aborted after a conflict.
3. Number of times a user force pushes (potentially as a % of regular pushes if that's possible?)

The big things we're trying to learn here are whether this is something people are enabling without surfacing anything in the UI, how frequently they're hitting conflicts, and whether they're subsequently force pushing the resolution.

Does that sound reasonable/appropriate?

shiftkey:
> Does that sound reasonable/appropriate?

This all sounds reasonable

scottypate:
`Does that sound reasonable/appropriate?`

I second that this is all good stuff. 

A follow-up measurement that I need to look at is the effect that rebasing in desktop has on other collaborators on the repo that are not using desktop. I have a hypothesis that easier rebasing in Desktop could increase conflicts in other areas of the product. We don't need to collect any data for that here other than what @billygriffin described above.

shiftkey:
> 3\. Number of times a user force pushes (potentially as a % of regular pushes if that's possible?)

Force push is not in scope for `1.6.3` so we don't have anything to support this metric. Moving this out to #6935 to track there.

shiftkey:
An update on what is coming in `1.6.3` for rebase metrics:

 -  Number of times a user pulls with rebase (potentially as a % of regular pulls if that's possible?)

See #6893 for those metrics

[...]

PR #6839, opened 2019-02-13
[rebase] add state changes to support detecting rebase conflicts

shiftkey:
## Overview

Introduce changes to the application internals to support detecting a rebase is underway, as well as what conflicts exists, so that these can be surfaced to the user.

## Description

With the overlap between merge conflicts and rebase conflicts, I eventually settled on this model for the conflicts state that the app maintains:

```ts
export type MergeConflictState = {
  readonly kind: 'merge'
  readonly currentBranch: string
  readonly currentTip: string
  readonly manualResolutions: Map<string, ManualConflictResolution>
}

export type RebaseConflictState = {
  readonly kind: 'rebase'
  readonly currentTip: string
  readonly manualResolutions: Map<string, ManualConflictResolution>
  // TODO: what other state do we need to capture here?
}

export type ConflictState = MergeConflictState | RebaseConflictState
```

We were previously using an interface for `IConflictState` but as we don't implement the interface I took the chance to drop the `I` prefix and come up with some better names.

My thinking on this is that you can be in one of these states with your Git repository, but never multiple states:

 - no conflicts
 - conflicts introduced by a `git merge`
 - conflicts introduced by a `git rebase`
 - (possibly) conflicts introduced by a `git stash pop`
 - (possibly) conflicts introduced by a `git cherry-pick`

The flow-on effects from this:

 - `getStatus` (which transforms our raw `git status` into an app-specific model) now looks for rebase conflict details when the feature flag is enabled
 -  `AppStore._loadStatus` (which determines whether it needs to show the merge conflicts dialog) is now aware of rebase conflicts (there's no dialog to show currently, but I've added the relevant plumbing)
 - `getConflictState` inside `changes-state.ts` (which handles state changes and performs side-effects like updating the `AppStore`)  is now aware of rebase conflicts (but doesn't perform and side-effects, see #6550 for those details)

TODO:

 - [x] write some tests for rebase conflict detection inside `getStatus`
 - [x] write some tests for rebase state transitions inside `getConflictsState`, using an example event from #6550 for reference
 - [x] figure out where the type guards should live (`app-state.ts` should be plain objects wherever possible)
 - [x] rebase and retarget PR once #6830 has been merged
 - [x] test existing merge conflicts behaviour is unaffected
 - [x] review impacted code and en sure important features are documented

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -190,7 +190,7 @@ import {
   ManualConflictResolutionKind,
 } from '../../models/manual-conflict-resolution'
 import { BranchPruner } from './helpers/branch-pruner'
-import { enableBranchPruning } from '../feature-flag'
+import { enableBranchPruning, enableNewRebaseFlow } from '../feature-flag'
 
 /**
  * As fast-forwarding local branches is proportional to the number of local
@@ -1614,13 +1614,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
       conflictState: updateConflictState(state, status, this.statsStore),
     }))
 
-    this._triggerMergeConflictsFlow(repository)
+    this._triggerConflictsFlow(repository)
 
     this.emitUpdate()
 
     this.updateChangesDiffForCurrentSelection(repository)
 
     return true
+  }
+
+  private async _triggerConflictsFlow(repository: Repository) {
+    if (enableNewRebaseFlow()) {
+      this._triggerRebaseFlow(repository)
+      this._triggerMergeConflictsFlow(repository)
+    } else {
+      this._triggerMergeConflictsFlow(repository)
+    }
+  }
+
+  /** display the rebase flow, if necessary */
+  private async _triggerRebaseFlow(repository: Repository) {
+    // TODO: check the current popup is a rebase conflicts dialog
+    // TODO: if already in flow, exit
+    // TODO: otherwise show the popup
   }
 
   /** starts the conflict resolution flow, if appropriate */
@@ -1640,7 +1656,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const repoState = this.repositoryStateCache.get(repository)
     const { conflictState } = repoState.changesState
-    if (conflictState === null) {
+    if (conflictState === null || conflictState.kind !== 'merge') {
       return
     }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -94,6 +94,7 @@ import {
   SuccessfulMergeBannerState,
   MergeConflictsBannerState,
   MergeConflictState,
+  isMergeConflictState,
 } from '../app-state'
 import { IGitHubUser } from '../databases/github-user-database'
 import {
@@ -1709,7 +1710,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const repoState = this.repositoryStateCache.get(repository)
     const { conflictState } = repoState.changesState
-    if (conflictState === null || conflictState.kind !== 'merge') {
+    if (conflictState === null || !isMergeConflictState(conflictState)) {
       return
     }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -134,6 +134,8 @@ import {
   createMergeCommit,
   getBranchesPointedAt,
   isGitRepository,
+  abortRebase,
+  continueRebase,
 } from '../git'
 import {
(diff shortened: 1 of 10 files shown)


shiftkey:
> figure out where the type guards should live (`app-state.ts` should be plain objects wherever possible)

After thinking about this a bunch I'm going to settle on this convention:

> type guard functions belong in the same module as the types they are associated with

We have some existing examples of this around, e.g. `status-parser.ts`:

https://github.com/desktop/desktop/blob/a0b9b182ea4595330e5630e1812a2591e48d484b/app/src/lib/status-parser.ts#L30-L40

In an ideal world the models in this file would live under a `models` directory rather than `lib` (which to me should be for functionality), but that can be a future refactoring task.

This will help us avoid writing duplicate functions that do the same thing (one example I found is `isConflictedFile` and `isConflictedFileStatus` both exist in code base)

shiftkey:
Marking this one as `time-sensitive` because it's a blocker on #6698, and I'd love some input on whether the changes here make sense to others. I've added a 📝 task to document things as well while I'm in here making changes.

app/src/lib/stores/updates/changes-state.ts
```diff
@@ -162,6 +175,76 @@ export function updateConflictState(
...
+}
+
+export function updateConflictState(
+  state: IChangesState,
+  status: IStatusResult,
+  statsStore: IStatsStore
+): ConflictState | null {
+  const prevConflictState = state.conflictState
+
+  const manualResolutions = prevConflictState
```
iAmWillShepherd:
```suggestion
  const manualResolutions = prevConflictState !== null
```

app/src/lib/stores/updates/changes-state.ts
```diff
@@ -96,40 +103,46 @@ export function updateChangedFiles(
...
+      currentTip,
+      manualResolutions,
+    }
   }
 
-  const { currentBranch, currentTip } = status
-  if (currentBranch == null || currentTip == null) {
-    return null
-  }
+  if (status.rebaseContext !== null) {
```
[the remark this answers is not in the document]
shiftkey: Not really, just that something needs to be checked first.

app/src/lib/stores/app-store.ts
```diff
@@ -1621,6 +1625,74 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    const possibleTheirsBranches = await getBranchesPointedAt(
+      repository,
+      'MERGE_HEAD'
+    )
+    // null means we encountered an error
+    if (possibleTheirsBranches === null) {
+      return
+    }
+    const theirBranch =
+      possibleTheirsBranches.length === 1
```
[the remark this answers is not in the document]
shiftkey: This code is duplicating the existing `AppStore._triggerMergeConflictsFlow` function...

https://github.com/desktop/desktop/blob/a23bdc8f6c573b2bffd0a8bbffb53c12cf911edd/app/src/lib/stores/app-store.ts#L1649-L1657

My only change to the flow is to pass in the `conflictState` which is known in the caller function, but because of the feature flag I wanted to preserve the existing flow.

I traced this change to f543e8fba126d89c427e453724856b492ee17021 and #6067, which mentions:

> `git branch --points-at=MERGE_HEAD` works but only if one branch has the same tip, otherwise we get multiple results

So in the case of having zero or multiple branches, we don't show a branch name here to avoid confusing users ("I was on branch A but the app now says branch B, whyyyyy?")

app/src/lib/stores/updates/changes-state.ts
```diff
@@ -96,40 +103,46 @@ export function updateChangedFiles(
...
+      originalBranchTip,
+    }
   }
+
+  return null
 }
 
-export function updateConflictState(
-  state: IChangesState,
+function performEffectsForMergeStateChange(
```
outofambit: i like this putting all the side effects into a single little function

app/src/models/tip.ts
```diff
@@ -1,10 +1,10 @@
 import { Branch } from './branch'
 
 export enum TipState {
-  Unknown,
-  Unborn,
-  Detached,
-  Valid,
+  Unknown = 'Unknown',
```
outofambit: :100:

app/src/lib/stores/updates/changes-state.ts
shiftkey: We've not been great with making side effects stand out from the rest of the code in the past, so hopefully being more explicit like this when we're in making changes will help keep the intent clear.

shiftkey:
@outofambit @iAmWillShepherd any objections to me merging this in later today so I can get #6698 ready to review?

PR #6848, opened 2019-02-14
Allow octicon to take and render alt text

### LINKED ISSUE #6495: Allow Octicon to take and render alt text
Our Octicon component doesn't allow setting an alt text (for screen readers, and other accessibility technologies) via its props. Tackling some accessibility bugs requires that we implement this.

According to this somewhat outdated [article](https://css-tricks.com/accessible-svgs/) you can do this by adding a title tag at the beginning of the svg block, which I think will work nicely with our current implementation.

The Octicon component prop should be called `ariaLabel`, as [this is used by the octicons-react package's components](https://github.com/primer/octicons/tree/v8.1.0/lib/octicons_react#arialabel). This will ease our transition whenever we tackle #5185.

The Octicon component's render method lives here:

https://github.com/desktop/desktop/blob/249c06d51332a7254de848e1c77a60e1ae52d587/app/src/ui/octicons/octicon.tsx#L65-L79

lisavogtsf:
## Overview

**Closes #6495**

## Description

This PR allows the Octicon to take in a `title` prop and incorporate it into the SVG that is rendered in an accessible way.

The `title` prop already exists and is incorporated in the SVG's `path` for the purpose of tool-tips. We may be able to remove that implementation. Currently a value for title is never supplied to Octicons in the github desktop codebase. With these changes--making `<title>` the first child element of `<svg>`--the title will function like an image `alt` attribute. In case assistive technologies do not pick up on that there is an `aria-labelledby` attribute replacing `aria-hidden`, as we now want screenreaders to see the element.

In addition, as values for title are not currently being supplied, this PR also displays the supplied `className` as the SVG `<desc>` tag. The hope was to take advantage of some of the information currently being supplied to Octicons (`dropdownArrow`, `no-newline`), even though not many get descriptive classNames currently.  

## Release notes

Notes:

### DIFF
--- a/app/src/ui/octicons/octicon.tsx
+++ b/app/src/ui/octicons/octicon.tsx
@@ -20,6 +20,11 @@ interface IOcticonProps {
    * An optional string to use as a tooltip for the icon
    */
   readonly title?: string
+
+  /**
+   * An optional string to provide accessible descriptive text
+   */
+  readonly description?: string
 }
 
 /**
@@ -68,6 +73,7 @@ export class Octicon extends React.Component<IOcticonProps, {}> {
     const shortClassName = this.props.className
     const className = classNames('octicon', shortClassName)
     const title = this.props.title
+    const description = this.props.description
 
     return (
       <svg
@@ -77,7 +83,7 @@ export class Octicon extends React.Component<IOcticonProps, {}> {
         viewBox={viewBox}
       >
         <title id="octiconTitle">{title}</title>
-        <desc id="octiconDescription">{shortClassName}</desc>
+        <desc id="octiconDescription">{description}</desc>
         <path d={symbol.d}>{this.renderTitle()}</path>
       </svg>
     )
--- a/app/src/ui/octicons/octicon.tsx
+++ b/app/src/ui/octicons/octicon.tsx
@@ -66,6 +66,7 @@ export class Octicon extends React.Component<IOcticonProps, {}> {
     const symbol = this.props.symbol
     const viewBox = `0 0 ${symbol.w} ${symbol.h}`
     const className = classNames('octicon', this.props.className)
+    const title = this.props.title
 
     return (
       <svg
@@ -74,6 +75,7 @@ export class Octicon extends React.Component<IOcticonProps, {}> {
         version="1.1"
         viewBox={viewBox}
       >
+        <title>{title}</title>
         <path d={symbol.d}>{this.renderTitle()}</path>
       </svg>
     )
--- a/app/src/ui/octicons/octicon.tsx
+++ b/app/src/ui/octicons/octicon.tsx
@@ -72,12 +72,13 @@ export class Octicon extends React.Component<IOcticonProps, {}> {
     return (
       <svg
         aria-hidden="true"
+        aria-labelledby="octiconTitle octiconDescription"
         className={className}
         version="1.1"
         viewBox={viewBox}
       >
-        <title>{title}</title>
-        <desc>{shortClassName}</desc>
+        <title id="octiconTitle">{title}</title>
+        <desc id="octiconDescription">{shortClassName}</desc>
         <path d={symbol.d}>{this.renderTitle()}</path>
       </svg>
     )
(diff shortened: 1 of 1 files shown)


outofambit:
thanks @lisavogtsf! i need to dig into this a little more, but at first glance this looks like a really smart approach. ✨

lisavogtsf:
@outofambit Sounds good. I was only grabbing the Classname as a way to have some kind of content available as alt text, since the title prop is never given a value. But it's true the names are mostly unhelpful. I will make those changes.

lisavogtsf:
@outofambit I made the requested changes

shiftkey:
> (It would be nice if we could substitute the name of the `OcticonSymbol` somehow, but I don't see a way to do that programmatically right now.)

The set of `OcticonSymbol` are generated in `app/src/ui/octicons/octicons.generated.ts` which is made from `script/generate-octicons.ts` - this could easily be updated to add a new field if you wanted to store a string that represents the name of each icon.

A quick patch to test this out:

```diff
diff --git a/script/generate-octicons.ts b/script/generate-octicons.ts
index a5c7d637d..c0e54e4e7 100644
--- a/script/generate-octicons.ts
+++ b/script/generate-octicons.ts
@@ -22,6 +22,7 @@ interface IXML2JSNode {
 
 interface IOcticonData {
   readonly jsFriendlyName: string
+  readonly symbol: string
   readonly pathData: string
   readonly width: string
   readonly height: string
@@ -62,16 +63,17 @@ async function generateIconData(): Promise<ReadonlyArray<IOcticonData>> {
 
     const result = await readXml(octicon.path)
     const pathData = result.path.$.d
+    const symbol = octicon.symbol
     const jsFriendlyName = toCamelCase(octicon.symbol)
 
-    results.push({ jsFriendlyName, width, height, pathData })
+    results.push({ jsFriendlyName, width, height, pathData, symbol })
   }
 
   return results
 }
 
-generateIconData().then(result => {
-  console.log(`Writing ${result.length} octicons...`)
+generateIconData().then(data => {
+  console.log(`Writing ${data.length} octicons...`)
 
   const out = fs.createWriteStream(
     Path.resolve(__dirname, '../app/src/ui/octicons/octicons.generated.ts'),
@@ -90,13 +92,13 @@ generateIconData().then(result => {
   out.write('export class OcticonSymbol {\n')
 
   out.write(
-    '\n  public constructor(public w: number, public h: number, public d: string) { }\n\n'
+    '\n  public constructor(public w: number, public h: number, public d: string, public symbol: string) { }\n\n'
   )
 
-  result.forEach(function(symbol) {
-    const { jsFriendlyName, pathData, width, height } = symbol
+  data.forEach(function(d) {
+    const { jsFriendlyName, pathData, width, height, symbol } = d
     out.write(
-      `  public static get ${jsFriendlyName}() { return new OcticonSymbol(${width}, ${height}, '${pathData}') }\n`
+      `  public static get ${jsFriendlyName}() { return new OcticonSymbol(${width}, ${height}, '${pathData}', '${symbol}') }\n`
     )
   })
 
```

And the resulting code generated using `yarn generate-octicons`:

```ts
/*
 * This file is automatically generated by the generate-octicons tool.
 * Manually changing this file will only lead to sadness.
 */

export class OcticonSymbol {
  public constructor(
    public w: number,
    public h: number,
    public d: string,
    public symbol: string
  ) {}

  public static get alert() {
    return new OcticonSymbol(
      16,
      16,
      'M8.893 1.5c-.183-.31-.52-.5-.887-.5s-.703.19-.886.5L.138 13.499a.98.98 0 0 0 0 1.001c.193.31.53.501.886.501h13.964c.367 0 .704-.19.877-.5a1.03 1.03 0 0 0 .01-1.002L8.893 1.5zm.133 11.497H6.987v-2.003h2.039v2.003zm0-3.004H6.987V5.987h2.039v4.006z',
      'alert'
    )
  }
  public static get archive() {
    return new OcticonSymbol(
      14,
      16,
      'M13 2H1v2h12V2zM0 4a1 1 0 0 0 1 1v9a1 1 0 0 0 1 1h10a1 1 0 0 0 1-1V5a1 1 0 0 0 1-1V2a1 1 0 0 0-1-1H1a1 1 0 0 0-1 1v2zm2 1h10v9H2V5zm2 3h6V7H4v1z',
      'archive'
    )
  }

 ...
```

This means you'll be able to access this inside the `Octicon` component using `this.props.symbol.symbol`.

Feel free to extract this to a fresh issue as follow-up work.

PR #6878, opened 2019-02-19
remove old folder path from codecov config

outofambit:
since #6695, our codecov config has been pointing to a nonexistent directory and causing codecov statuses to error out.

this removes that path from the config. (those files are now included in the ui-components glob for codecov status, so we're still tracking their coverage!)

codecov PR statuses should return after this is merged

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -17,7 +17,6 @@ coverage:
         base: pr
         paths:
           - 'app/src/lib/stores'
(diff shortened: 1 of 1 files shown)


shiftkey:
@outofambit nice spot!

shiftkey:
> codecov PR statuses should return after this is merged

Expediting this PR so we can get these back!

PR #6884, opened 2019-02-19
unify banner state in application to pay down some tech debt

shiftkey:
## Overview

This PR demonstrates how a unified banner API might work in the application. I need to implement some banners as part of the rebase flow in #6698, so I wanted to open this cleanup PR early and get feedback.

Shout out to @outofambit for brain dumping her thoughts earlier today to help me get started.

## Description

This work came from our learnings about managing popups, with a slight improvement that I don't think we've done elsewhere.

First, we defined a unified `Banner` type to represent all the possible states of a banner:

https://github.com/desktop/desktop/blob/0eea8253dc44554cd6b649332e08b97192665d02/app/src/models/banner.ts#L1-L22

Then, we defined a single `currentBanner` state inside `app-state`, replacing `successfulMergeBannerState` and `mergeConflictsBannerState`.

Lastly, `AppStore` and `Dispatcher` now define a single `setBanner`/`clearBanner` functions to manage this state, rather than two pairs of functions to update each state.

This PR differs from our current `Popup` infrastructure by moving the render function out to it's own module, rather than inlining it inside `App`:

https://github.com/desktop/desktop/blob/0eea8253dc44554cd6b649332e08b97192665d02/app/src/ui/app.tsx#L1888-L1896

This means `App` doesn't need to be aware of the components it can render as banners (fewer imports, better separation of concerns), and we can use `assertNever` to ensure `renderBanner` handles all of the states it should:

https://github.com/desktop/desktop/blob/0eea8253dc44554cd6b649332e08b97192665d02/app/src/ui/banners/render-banner.tsx#L12-L41

You might note that the `UpdateAvailable` banner is technically a banner, but for the sake of this PR I'm not touching that logic or it's representation. I think it could live in there at some point, but I haven't thought about it.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -34,6 +34,7 @@ import { ComparisonCache } from './comparison-cache'
 import { ApplicationTheme } from '../ui/lib/application-theme'
 import { IAccountRepositories } from './stores/api-repositories-store'
 import { ManualConflictResolution } from '../models/manual-conflict-resolution'
+import { Banner } from '../models/banner'
 
 export enum SelectionType {
   Repository,
@@ -99,6 +100,7 @@ export interface IAppState {
   readonly focusCommitMessage: boolean
   readonly currentPopup: Popup | null
   readonly currentFoldout: Foldout | null
+  readonly currentBanner: Banner | null
 
   /**
    * A list of currently open menus with their selected items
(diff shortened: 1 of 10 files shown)


shiftkey:
I've tentatively rebased #6698 on top of this branch to help me test the API. It's not a hard dependency on shipping, but given how much this better organizes things I'd like to gather feedback on this while I'm testing #6698

outofambit:
Glad to see the addition of a `render-banner` to this pattern! Thats a worthwhile improvement. 🎉

shiftkey:
FYI @desktop/engineering I plan to merge this in tomorrow morning my time so #6698 will be ready to review, so let me know if there's any other feedback about this change.

PR #6893, opened 2019-02-20
display "pull with rebase" if user has set this in config

### LINKED ISSUE #3422: Option to rebase instead of automatically merge when pulling from upstream
### Description

Github Desktop automatically merges upstream commits. This is fine for some projects but quickly dirties the git history with unnecessary merges. This has bitten me and colleagues multiple times in this version and the previous version of Github Desktop with the magic "Sync" button.

### Proposal

Add an option in the menu to pull rebase instead of merge, or change button text to "Merge" so that I know it's time to go to the terminal so I can pull rebase.

### Version

**GitHub Desktop version:** 1.0.3

**OS version:** OSX 10.12.6

*Note:* I did look for similar issues on this topic but couldn't find them.

### LINKED ISSUE #6553: "Pull remote" button should hint that rebase will occur for branch
The design adds a "with rebase" message to the "Pull origin" action:

This check can be done in one of two ways:

  - for the current branch - [`git config branch.[name].rebase true`](https://git-scm.com/docs/git-config#git-config-branchltnamegtrebase)
  - for the repository - [`git config pull.rebase true`](https://git-scm.com/docs/git-config#git-config-pullrebase)

There are some additional options here rather than just `true` which we need to be aware of:

> When `true`, rebase branches on top of the fetched branch, instead of merging the default branch from the default remote when "git pull" is run. See "branch.<name>.rebase" for setting this on a per-branch basis.
>
> When `merges`, pass the `--rebase-merges` option to git rebase so that the local merge commits are included in the rebase (see [`git-rebase[1]`](https://git-scm.com/docs/git-rebase) for details).
>
> When `preserve`, also pass `--preserve-merges` along to git rebase so that locally committed merge commits will not be flattened by running git pull.
>
> When the value is `interactive`, the rebase is run in interactive mode.
> 
> NOTE: this is a possibly dangerous operation; do not use it unless you understand the implications (see [`git-rebase[1]`](https://git-scm.com/docs/git-rebase) for details).

The first two options seem reasonable to consider, but the third one is not going to be able to be supported right now. I don't see a way to be able to disable that specific case in `git-pull` so that's going to be an interesting one to figure out 🤔

shiftkey:
## Overview

**Closes #6553**
**Closes #3422**

Also adds `pullWithDefaultSettingCount` and `pullWithRebaseCount` metrics from #6550 

## Release notes

Notes: Pull operation in UI indicates whether user has pull.rebase configured

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -427,6 +427,16 @@ export interface IBranchesState {
 
   /** The pull request associated with the current branch. */
   readonly currentPullRequest: PullRequest | null
+
+  /**
+   * Is the current branch configured to rebase on pull?
+   *
+   * This can be configured in one of two ways by the user:
+   *
+   *  - `git config branch.[name].rebase true` (local)
+   *  - `git config pull.rebase true` (local or global)
+   */
+  readonly pullWithRebase?: boolean
 }
 
 export interface ICommitSelection {
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -431,9 +431,8 @@ export interface IBranchesState {
   /**
    * Is the current branch configured to rebase on pull?
    *
-   * This can be configured in one of two ways by the user:
+   * This is equivalent to this config value in Git:
    *
-   *  - `git config branch.[name].rebase true` (local)
    *  - `git config pull.rebase true` (local or global)
    */
   readonly pullWithRebase?: boolean
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -431,9 +431,10 @@ export interface IBranchesState {
   /**
    * Is the current branch configured to rebase on pull?
    *
-   * This is equivalent to this config value in Git:
+   * This is the value returned from git config (local or global) for `git config pull.rebase`
    *
-   *  - `git config pull.rebase true` (local or global)
+   * If this value is not found in config, this will be `undefined` to indicate
+   * that the default Git behaviour will occur.
    */
   readonly pullWithRebase?: boolean
 }
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -62,6 +62,7 @@ import {
   revRange,
   revSymmetricDifference,
   getSymbolicRef,
+  getGlobalConfigValue,
 } from '../git'
 import { RetryAction, RetryActionType } from '../../models/retry-actions'
 import { UpstreamAlreadyExistsError } from './upstream-already-exists-error'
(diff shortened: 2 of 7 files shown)


app/src/lib/stores/git-store.ts
```diff
@@ -320,6 +324,22 @@ export class GitStore extends BaseStore {
     return allBranchesWithUpstream
   }
 
+  private async checkPullWithRebase() {
+    const result = await getGlobalConfigValue('pull.rebase')
```
shiftkey: I worry about how often this will fire to keep it up to date, and whether there's a better way to get the global config and cache in app state. We don't use this value when it's time to perform the pull, but it's needs to be checked more frequently than our other uses in Preferences.

app/src/lib/app-state.ts
```diff
@@ -411,6 +411,15 @@ export interface IBranchesState {
...
   readonly currentPullRequest: PullRequest | null
+
+  /**
+   * Is the current branch configured to rebase on pull?
+   *
+   * This is equivalent to this config value in Git:
+   *
+   *  - `git config pull.rebase true` (local or global)
+   */
+  readonly pullWithRebase?: boolean
```
iAmWillShepherd: It's not clear to me what this property being `undefined` means. Could you add a bit more to the docs to explain?

app/src/lib/stores/git-store.ts
```diff
@@ -320,6 +324,20 @@ export class GitStore extends BaseStore {
...
+    const result = await getConfigValue(this.repository, 'pull.rebase')
+
+    if (result === null || result === '') {
+      this.pullWithRebase = undefined
+    } else if (result === 'true') {
+      this.pullWithRebase = true
+    } else if (result === 'false') {
+      this.pullWithRebase = false
+    } else {
+      log.warn(`Unexpected value found for pull.rebase in config: '${result}'`)
```
iAmWillShepherd: If we get to the `else` clause, should we be explicitly setting `this.pullWithRebase` to `undefined`, clearing what ever the previous value is?

app/src/lib/app-state.ts
iAmWillShepherd: I was able to figure out what it means from reviewing the code below, but I still think a comment of what it means for Desktop would be helpful.

iAmWillShepherd (changes requested):

app/src/lib/app-state.ts
shiftkey: See ff462628c for more context

app/src/lib/stores/git-store.ts
shiftkey: Good shout, fixed in  ff462628c

iAmWillShepherd (dismissed):

iAmWillShepherd:
@shiftkey this is good to merge after resolving conflicts.

shiftkey:
@iAmWillShepherd 👍 rebased on top of `development`

iAmWillShepherd (approved):

karlp:
We're not seeing this with github desk 2.6.3 on windows?  using "open git command prompt" and then git config -l shows pull.rebase=true, but we still always see the plain "pull" button, and get merge commits?

Issue #6460, opened 2018-12-20
Repositories are difficult to find, navigate, and differentiate


billygriffin:
**Please describe the problem you think should be solved**

We've seen many issues suggesting to change the way we either visualize or group repositories in GitHub Desktop. Right now we just organize repos alphabetically and it's tricky to differentiate many of them visually because it's just in the format of [repo name], which doesn't account for a variety of different scenarios. We have a tooltip that displays on hover, but it's relatively undiscoverable, slow to appear, and requires you to look at each repo individually to differentiate. We also group repos that are on GitHub.com separate from those that are purely local or on another remote. We do have filtering repos via a text input in place today, which should be helpful, but obviously not sufficient based on the feedback. 

There have many different implementations suggested, so I wanted to try organize most of the issues I could find in one place to attempt to get our arms around the problem areas as we try to determine a path forward. Ideally, whatever we implement would be something we could work forward from and we could therefore close and/or remove `future proposal` from all the previous issues.

**Relevant issues:**

Group repos by organization and show full repo name: https://github.com/desktop/desktop/issues/1533
Ability to "pin" repos to top of list: https://github.com/desktop/desktop/issues/2038
Allow repo list to be always expanded: #1593 
Group repos by organization (and make orgs collapsable): https://github.com/desktop/desktop/issues/2037
Add categories for repos to be grouped together: https://github.com/desktop/desktop/issues/4945
Add folders for repos to be grouped together: https://github.com/desktop/desktop/issues/1598
Group or filter repos by [Topics](https://blog.github.com/2017-01-31-introducing-topics/): https://github.com/desktop/desktop/issues/4958
Drag and drop and/or add labels to repos: https://github.com/desktop/desktop/issues/2324
Add visual differentiation for repos: https://github.com/desktop/desktop/issues/1929
Add a keyword to repos in Desktop: https://github.com/desktop/desktop/issues/4940
Add placeholder name for repos in Desktop: https://github.com/desktop/desktop/issues/3197
Support multiple windows: https://github.com/desktop/desktop/issues/3606
Don't persist repo filter after selecting one: https://github.com/desktop/desktop/issues/4273
Filter across multiple repos with OR operator: https://github.com/desktop/desktop/issues/5280
Repo order is not maintained alphabetically after filtering: https://github.com/desktop/desktop/issues/4184

Slight variation
Some way to visually distinguish worktrees: https://github.com/desktop/desktop/issues/5764

Additional contextual things:
[...]

PR #6896, opened 2019-02-20
use fullname of github repos for filtering in repo list

outofambit:
## Overview

small step in addressing #6460 

filtering repository list by repository owner

## Description

- adds fullname (`${ownerName}/${repoName}`) of github repos to be filtered on in the repository sidebar

## Release notes

Notes: [Improved] Repository list filtering now searches GitHub owner name

### DIFF
--- a/app/src/ui/repositories-list/group-repositories.ts
+++ b/app/src/ui/repositories-list/group-repositories.ts
@@ -1,4 +1,8 @@
-import { Repository, ILocalRepositoryState } from '../../models/repository'
+import {
+  Repository,
+  ILocalRepositoryState,
+  nameOf,
+} from '../../models/repository'
 import { CloningRepository } from '../../models/cloning-repository'
 import { getDotComAPIEndpoint } from '../../lib/api'
 import { caseInsensitiveCompare } from '../../lib/compare'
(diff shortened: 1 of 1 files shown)


shiftkey:
General question for the crowd: is the ability to filter and get results that don't contain the expected text a potential source of confusion?

This highlighted example is interesting that the first result appears to be `outofambit/jest` but only `jest` is visible in the label:



One possible way to mitigate this could be to show the `owner/name` result when a result matches, but I haven't thought deeply about other approaches.

I also stumbled upon #6898 while pondering this, but that's not something to resolve here.

outofambit:
> General question for the crowd: is the ability to filter and get results that don't contain the expected text a potential source of confusion?

@shiftkey that's a good question! we could also only include the `owner/name` string as searchable for a repo if there are other with the same name that need to be disambiguated with `owner`. 🤔

shiftkey:
> we could also only include the `owner/name` string as searchable for a repo if there are other with the same name that need to be disambiguated with `owner`.

When more than one repository have the same `name` we try and append the `owner`, like in this example:



We look at this flag to determine what to render:

https://github.com/desktop/desktop/blob/3223b072978890b0be9fc043a9101b0ae1b63713/app/src/ui/repositories-list/group-repositories.ts#L77

But maybe search isn't working in the same way and could be improved?

say25:
I will add that the current way it works, if the `nameCount is > 1` but the org names are all the same, we still display the org names

outofambit:
i'm going to close this out as i'm iterating on this locally now. but it'll be back soon! thanks all!

lindsay lohan saying i'll be right back and walking away

PR #6907, opened 2019-02-21
show a "continue rebase" button when conflicts are outstanding

### LINKED ISSUE #6890: commit form should provide better context when dealing with rebase conflicts
As part of #6698 the user can dismiss the dialog and end up in this state:

screen shot 2019-02-20 at 11 43 44 am

If you "Commit to master", you'll trigger the warning about committing files with conflicts - this is fine:

screen shot 2019-02-20 at 11 45 52 am

And then the rebase trigger kicks in and says "we're good to continue" - this is also fine:

screen shot 2019-02-20 at 11 46 35 am

But I'd like to change the commit form to do two things:

 - disable the summary and description inputs - being able to specify these details in the commit will clash with the previous history when the rebase was started, which could lead to more confusion
 - change the commit button text to "Continue rebase" to indicate this is not the normal commit flow

I'm also open to other ideas here about this flow, if you've got them.

shiftkey:
## Overview

**Closes #6890**

## Description

This swaps out the default commit message for a simple "Continue rebase" button to prevent the user from creating commits as usual outside of the "Rebase conflicts" dialog.

The current view in #6698 when you dismiss the "Rebase conflicts" dialog:

While there are outstanding conflicts to resolve:

When those are resolved:

This also hides the "Undo Commit" dialog (which will appear when any rebased commit hasn't been published). 

TODO:

 - [x] review changes to sidebar code to see if it can be better organized
 - [x] confirm `Continue rebase` [does exactly what it says on the tin](https://en.wikipedia.org/wiki/Does_exactly_what_it_says_on_the_tin)

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/changes/continue-rebase.tsx
+++ b/app/src/ui/changes/continue-rebase.tsx
@@ -5,13 +5,13 @@ import { RebaseConflictState } from '../../lib/app-state'
 import { Dispatcher } from '../dispatcher'
 import { Repository } from '../../models/repository'
 import { WorkingDirectoryStatus } from '../../models/status'
+import { getConflictedFiles } from '../../lib/status'
 
 interface IContinueRebaseProps {
   readonly dispatcher: Dispatcher
   readonly repository: Repository
   readonly workingDirectory: WorkingDirectoryStatus
   readonly rebaseConflictState: RebaseConflictState
-  readonly canCommit: boolean
   readonly isCommitting: boolean
 }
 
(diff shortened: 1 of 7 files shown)


billygriffin:
@shiftkey Looks great! Sorry if this is a dumb question, but is the button that replaced the commit area in the bottom screenshot supposed to say "Continue rebase" and not "Rebase {branch_name}"? Since you're already in it and not starting anew, I'd imagine the "continue" verbiage would be important.

shiftkey:
> Since you're already in it and not starting anew, I'd imagine the "continue" verbiage would be important.

@billygriffin this is a good shout, and what I intended to do originally. I'll update the PR screenshots once I've corrected this.

billygriffin:
Thanks @shiftkey! Screenshots of the behavior now align with my expectations.

app/src/ui/changes/continue-rebase.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public render() {
+    const { manualResolutions } = this.props.rebaseConflictState
+
+    let canCommit = true
+
+    let tooltip = 'Continue rebase'
+
+    if (this.props.rebaseConflictState) {
```
iAmWillShepherd: **Non-blocking question**
What are we checking for here or is this check due to a limitation in the TS compiler? I'm asking because the type signature for this property is `RebaseConflictState`.

app/src/ui/changes/continue-rebase.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <div id="continue-rebase" role="group">
+        <Button
+          type="submit"
+          className="commit-button"
+          onClick={this.onSubmit}
+          disabled={!buttonEnabled}
+          tooltip={tooltip}
+        >
+          {loading}
+          <span>{loading ? 'Rebasing' : 'Continue rebase'}</span>
```
iAmWillShepherd:
```suggestion
          <span>{loading !== undefined ? 'Rebasing' : 'Continue rebase'}</span>
```

app/src/ui/changes/continue-rebase.tsx
shiftkey: This is an unnecessary check - dropped in d3a55eccb

PR #6914, opened 2019-02-21
propagate information to rebase conflicts dialog when pull fails

### LINKED ISSUE #6892: add context about pull to rebase error handler
This code from #6698 is not complete to support detecting and propagating when a `git pull --rebase` fails and the user needs to handle conflicts:

https://github.com/desktop/desktop/blob/765521faa946567d828c8bb2272200e870bf88d4/app/src/ui/dispatcher/error-handlers.ts#L403-L408

I need to wrap my head around:

 - what the current behaviour is when testing to close out #6549
 - what context should I set on a `pull` that might trigger rebase conflicts
 - how that meshes/overlaps with what we do inside `mergeConflictHandler`

https://github.com/desktop/desktop/blob/765521faa946567d828c8bb2272200e870bf88d4/app/src/ui/dispatcher/error-handlers.ts#L287-L303

Any events we need to track as part of rebasing are already covered by #6550, so I'm ignoring those as part of this task

shiftkey:
## Overview

**Closes #6892**

## Description

This PR reviews the data that we provide to the error handlers when a pull fails, so we can either show the merge conflicts or rebase conflicts flow. The biggest change here is that we were passing in the `tip` (which might not be on a branch). I've replaced this with more explicit properties:

 - `currentBranch` is the one that the user is performing the `git pull` or `git merge` on
 - `theirBranch` is the other branch involved with the operation

For a merge, it should be the "other" branch. For a rebase, it should be the tracking branch.

I can't come up with a case where the user should be able to `git pull` or `git merge` from a detached HEAD or unknown tip, so I think we should just report an error if we're in an unexpected state, rather than attempt the operation.

 - [x] testing with `git pull --rebase` to trigger conflicts

## Release notes

Notes:

### DIFF
--- a/app/src/lib/error-with-metadata.ts
+++ b/app/src/lib/error-with-metadata.ts
@@ -1,7 +1,7 @@
 import { Repository } from '../models/repository'
 import { CloningRepository } from '../models/cloning-repository'
 import { RetryAction } from '../models/retry-actions'
-import { IGitErrorContext } from './git-error-context'
+import { GitErrorContext } from './git-error-context'
 
 export interface IErrorMetadata {
   /** Was the action which caused this error part of a background task? */
@@ -14,7 +14,7 @@ export interface IErrorMetadata {
   readonly retryAction?: RetryAction
 
   /** Additional context that specific actions can provide fields for */
-  readonly gitContext?: IGitErrorContext
+  readonly gitContext?: GitErrorContext
 }
 
 /** An error which contains additional metadata. */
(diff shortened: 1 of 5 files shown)


outofambit:
i agree that throwing an error is a reasonable approach to invalid tips in this situation. i _think_ it should be impossible to get in that situation with the current UI too, no?

shiftkey:
> i _think_ it should be impossible to get in that situation with the current UI too, no?

I agree, and if there is a case I want to hear about it, because we should only be doing merges and rebases against valid branches.

PR #6922, opened 2019-02-22
chore: Upgrade moment to 2.24.0

### LINKED ISSUE #6921: upgrade moment dependency
I just got a notification about [`CVE-2017-18214`](https://nvd.nist.gov/vuln/detail/CVE-2017-18214) which was raised against the `moment` project:  https://github.com/moment/moment/issues/4163

We use `moment` in a few places, and the [ReDos](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS) vulnerability is an interesting read, but until I hear more I'm going to leave this open for someone to have a go at:

 - upgrading the dependency in `app/package.json` to `2.19.3` or later

https://github.com/desktop/desktop/blob/6eb4064f9941d256289c4c7816e553e40a6cc1ad/app/package.json#L39

 - run `yarn` from the root to update the lock file

 - commit both of these changes

 - test out the app to ensure it still renders relative dates correctly

Places to check as part of verifying the upgrade:

 - the history view - each commit shows a relative time in the description up until some limit, then it switches to the absolute time

 - the branches list shows the time of the last commit to each branch

 - the fetch button shows the relative time

HashimotoYT:
## Overview

**Closes #6921**

## Description

- Upgrades moment from 2.17.1 to 2.24.0 for fix vulnerability.

## Verification

* the history view - each commit shows a relative time in the description up until some limit, then it switches to the absolute time

* the fetch button shows the relative time

* the branches list shows the time of the last commit to each branch

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -36,7 +36,7 @@
     "fuzzaldrin-plus": "^0.6.0",
     "keytar": "^4.3.0",
     "memoize-one": "^4.0.3",
-    "moment": "^2.17.1",
+    "moment": "^2.24.0",
     "mri": "^1.1.0",
     "primer-support": "^4.0.0",
     "queue": "^4.4.2",
--- a/app/yarn.lock
+++ b/app/yarn.lock
@@ -1031,10 +1031,10 @@ mocha@^3.5.0:
     mkdirp "0.5.1"
     supports-color "3.1.2"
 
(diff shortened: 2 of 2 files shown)


iAmWillShepherd:
This doesn't really need to sit any longer.

PR #6923, opened 2019-02-22
group and filter repositories by owner

outofambit:
## Overview

partly addresses #6460
reboot of #6896

filter by owner with grouping by owner

## Description

- group and display github.com repositories by organization
- filter text field searches the owner field
- feature flagged for beta currently

## Release notes

Notes: [Improved] Group and filter GitHub repositories by owner name

### DIFF
--- a/app/src/ui/repositories-list/group-repositories.ts
+++ b/app/src/ui/repositories-list/group-repositories.ts
@@ -10,12 +10,19 @@ import { IFilterListGroup, IFilterListItem } from '../lib/filter-list'
 import { IAheadBehind } from '../../models/branch'
 import { enableGroupRepositoriesByOwner } from '../../lib/feature-flag'
 
+/**
+ * Special, reserved repository group names
+ *
+ * (GitHub.com user and org names cannot contain `_`,
+ * so these are safe to union with all possible
+ * GitHub repo owner names)
+ */
 export enum KnownRepositoryGroup {
-  enterprise = 'enterprise',
-  other = 'other',
+  Enterprise = '_Enterprise_',
+  NonGitHub = '_Non-GitHub_',
 }
 
-export type RepositoryGroupIdentifier = string | KnownRepositoryGroup
+export type RepositoryGroupIdentifier = KnownRepositoryGroup | string
 
 export type Repositoryish = Repository | CloningRepository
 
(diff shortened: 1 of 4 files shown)


outofambit:
marking this time sensitive since i'd like this to ship to beta this week

outofambit:
> marking this time sensitive since i'd like this to ship to beta this week

i would still _like_ to ship this to beta this week, but we don't _need_ to and have other pending work to support currently :)

app/src/ui/repositories-list/group-repositories.ts
```diff
@@ -28,18 +45,24 @@ export function groupRepositories(
   localRepositoryStateLookup: ReadonlyMap<number, ILocalRepositoryState>
 ): ReadonlyArray<IFilterListGroup<IRepositoryListItem>> {
   const grouped = new Map<RepositoryGroupIdentifier, Repositoryish[]>()
+  const gitHubOwners: Array<string> = []
```
shiftkey: Could this be a `Set<string>`? We'd get "avoid duplicate entries" for free, which is helpful later on.

app/src/ui/repositories-list/group-repositories.ts
```diff
@@ -84,9 +120,21 @@ export function groupRepositories(
   }
 
   // NB: This ordering reflects the order in the repositories sidebar.
-  addGroup('github')
-  addGroup('enterprise')
-  addGroup('other')
+  if (enableGroupRepositoriesByOwner()) {
+    const owners = gitHubOwners.reduce((acc, val) => {
```
shiftkey: If we use a `Set<string>` above we don't need to do this reducing to eliminate duplicate keys:

```
const owners = [...gitHubOwners.keys()]
```

outofambit: good call!

tierninho:
Tested the functionality and LGTM. Could not test Enterprise repos as it does not work in dev mode.

app/src/ui/repositories-list/group-repositories.ts
outofambit: done in 0850f18

say25:
Looked this over from a UI perspective, not sure if it is worth re-adding `GitHub.com` and `Enterprise` groupings with Owner grouping below that?

shiftkey:
> not sure if it is worth re-adding `GitHub.com` and `Enterprise` groupings with Owner grouping below that?

`Enterprise` grouping should not be affected by this change. I'm indifferent to whether there's value in displaying `GitHub.com` before the org groups (I don't think we can nest three levels deep in here).

PR #6930, opened 2019-02-25
:heavy_plus_sign: Add NoExistingRemoteBranch Popup

### LINKED ISSUE #1325: a better error message when pull occurs on non-existent upstream branch

say25:
## Overview

🌵 🌵 🌵 Blocked by https://github.com/desktop/dugite/pull/301 🌵 🌵 🌵

**Closes #1325**

## Description

- Adds a better error message when pull occurs on non-existent upstream branch

image

## Release notes

Notes: Added a better error message when pull occurs on non-existent upstream branch

### DIFF
--- a/app/src/ui/dispatcher/error-handlers.ts
+++ b/app/src/ui/dispatcher/error-handlers.ts
@@ -421,3 +421,56 @@ export async function localChangesOverwrittenHandler(
 
   return null
 }
+
+/**
+ * Handler for when a branch not longer exists on remote.
+ */
+export async function noExistingRemoteBranchHandler(
+  error: Error,
+  dispatcher: Dispatcher
+): Promise<Error | null> {
+  const e = asErrorWithMetadata(error)
+  if (!e) {
+    return error
+  }
+
+  const gitError = asGitError(e.underlyingError)
+  if (!gitError) {
+    return error
+  }
+
+  const dugiteError = gitError.result.gitError
+  if (!dugiteError) {
+    return error
+  }
+
+  if (dugiteError !== DugiteError.NoExistingRemoteBranch) {
+    return error
+  }
+
+  const { repository, gitContext } = e.metadata
+  if (repository == null) {
+    return error
+  }
+
+  if (!(repository instanceof Repository)) {
+    return error
+  }
+
+  if (gitContext == null) {
+    return error
+  }
+
+  const { tip } = gitContext
+  if (tip == null || tip.kind !== TipState.Valid) {
+    return error
+  }
+
+  dispatcher.showPopup({
+    type: PopupType.NoExistingRemoteBranch,
+    repository,
+    branchName: tip.branch.name,
+  })
+
+  return null
+}
--- a/app/src/ui/no-existing-remote-branch/no-existing-remote-branch-warning.tsx
+++ b/app/src/ui/no-existing-remote-branch/no-existing-remote-branch-warning.tsx
@@ -1,0 +1,73 @@
+import * as React from 'react'
+import { Button } from '../lib/button'
+import { ButtonGroup } from '../lib/button-group'
+import { Dialog, DialogContent, DialogFooter } from '../dialog'
+import { Dispatcher } from '../../ui/dispatcher'
+import { Repository } from '../../models/repository'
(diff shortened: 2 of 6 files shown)


app/src/ui/no-existing-remote-branch/no-existing-remote-branch-warning.tsx
```diff
@@ -0,0 +1,72 @@
...
+        id="no-existing-remote-branch"
+        title={
+          __DARWIN__ ? 'No Existing Remote Branch' : 'No existing remote branch'
+        }
+        onDismissed={this.props.onDismissed}
+        type="warning"
+      >
+        <DialogContent>
+          <p>
+            The branch <Ref>{this.props.branchName}</Ref> doesn't not exist on
```
shiftkey: Grammar suggestion:

```suggestion
            The branch <Ref>{this.props.branchName}</Ref> does not exist on
```

shiftkey:
[`dugite@1.84.0`](https://github.com/desktop/dugite/releases/tag/v1.84.0) has been published to NPM, which should unblock this PR.

tierninho:
Tested and LGTM 👍

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -418,6 +418,59 @@ export async function localChangesOverwrittenHandler(
...
+
+  if (!(repository instanceof Repository)) {
+    return error
+  }
+
+  if (gitContext == null) {
+    return error
+  }
+
+  const { tip } = gitContext
```
shiftkey: Have a read of the changes in https://github.com/desktop/desktop/pull/6914 - you should be able to use `currentBranch` here instead.

say25: Done. Thanks for the insight!

app/src/ui/no-existing-remote-branch/no-existing-remote-branch-warning.tsx
```diff
@@ -0,0 +1,72 @@
...
+            </Button>
+          </ButtonGroup>
+        </DialogFooter>
+      </Dialog>
+    )
+  }
+
+  private publishBranch = async () => {
+    this.props.dispatcher.closePopup()
+    await this.props.dispatcher.push(this.props.repository)
```
shiftkey: @say25 There's some extra detail in here that I don't think we've discussed, and I wanted to write it up and think about it some more before we merge this change.

From the original issue:

> However, I did a pull on the branch while still in the same status, and I got this error:

The root cause of this issue is that the upstream branch no longer exists on the remote, and while `push`-ing to restore the branch on the remote is one option, **what if the upstream branch was removed for a reason?**

Keep in mind that the user is trying to get remote changes into their branch, and if the remote branch doesn't exist any more maybe this is a hint that this workflow should be questioned.

There's another option here, which might not be known to most Git users: removing the tracking information from a branch without removing the local branch from the repository. In Git terminology this is `git branch --unset-upstream`.

If you unset the upstream branch the push/pull button will then render `Publish branch` to the user, which is a potentially more explicit step.

I'm going to think on this some more and report back, I just wanted to register my concern about only having the option to push the branch to the remote.

shiftkey:
@say25 for the moment I feel this PR has gone beyond what was initially discussed, and is touching on a potential workflow that we haven't really discussed. I'd like to discuss this back in #1325 and clarify how we should address things, because I think this is a chance to improve this workflow. 

Any objections to me closing out this PR for now? If we reach a consensus and this turns out to be the right approach anyway, we've got this code to use as a baseline.

say25:
Works for me

Issue #6390, opened 2018-12-11
Clicking "Undo" button doesn't populate summary field


elias6:
## Description
I have been using GitHub Desktop for months, and overall I've liked it, but something has been acting funny for the last week or two.

Before, whenever I clicked the "Undo" button to undo a commit, the summary and description fields get populated with the description of that commit.

But now, they don't always get populated. It seems to work for a while, and then stop working, seemingly at random. If I restart GitHub Desktop, it starts working again for a while, then randomly stops again.

Before, I could easily make small changes to my latest commit, like running `git commit --amend` from the command line. But now, I have to re-type the summary and description, which can be especially annoying if those things are long, or if I simply don't remember the description I had.

I'd like to see the old behavior, where the inputs are populated every time the "Undo" button is clicked.

## Version
* GitHub Desktop: 1.5.0
* Operating system: macOS 10.13.6

## Steps to Reproduce
Make a commit, then click the "Undo" button. You may need to try this several times, possibly after GitHub Desktop has been running for several hours.

### Expected Behavior
The commit message does not always show in the summary field.

### Actual Behavior
The commit message should always show in the summary field.

### Logs
I am afraid to attach my log file because it may contain sensitive company information. Sorry. I may try to reproduce this on my home laptop, but I can't promise that I will have time any time soon, or be able to reproduce it at all.

billygriffin:
@elias6 Thanks so much for letting us know, and sorry that's happening! That's definitely not intended and I don't remember seeing other issues for it. It sounds like it may be related to the state of the application given that it works sometimes but not others.

To confirm, it just started happening in the past couple weeks, but previously worked totally fine? Are there any characteristics of the repo(s) you've been working in that are different from before the problem started happening? @desktop/support I'm curious if you've seen this elsewhere.

elias6:
I have to say that I don't remember for sure how long it has been
happening. Maybe it was longer than 2 weeks, but it has only been happening
for a small part of the time that I've been using (the redesigned) GitHub
Desktop. I think the behavior could have been caused by a recent update,
but I don't really have any evidence to back that up.

I am not aware of anything unusual we did with our repos. We started using
Prettier with commit hooks. Other than that, I can't think of any other
major changes to the repos, besides the usual kind of commits, pushes, and
merges. And I think this started happening well after we started using
Prettier but I don't remember for sure.

elias6:
[...]

PR #6932, opened 2019-02-25
ensure unit test files match the expected pattern

shiftkey:
## Overview

While reviewing the latest from #6390 I noticed we had a unit test suite that wasn't being run, and a helper file in the wrong place. This PR addresses those changes, so that all our unit tests are named consistently.

## Description

Before this PR, we had some files that didn't match the pattern in our unit test directory:

```
$ find app/test/unit \! -name "*-test.ts" \! -name "*-test.tsx" -type f                                                                                                     
app/test/unit/stores/updates/changes-state-helper.ts
app/test/unit/format-commit-message.ts
```

With this PR branch, we only have files matching the right patterns that Jest uses:

```
$ find app/test/unit -name "*-test.ts" -type f | wc -l    
      85
$ find app/test/unit -name "*-test.tsx" -type f | wc -l             
      1                                                                                                 
```       
 
And no stray files:

```
$ find app/test/unit \! -name "*-test.ts" \! -name "*-test.tsx" -type f | wc -l
      0        
```

## Release notes

Notes: no-notes


shiftkey:
Merging this in because @outofambit is on leave and it's a small infrastructure fix

PR #6933, opened 2019-02-25
[WIP] Event metrics

iAmWillShepherd:
A P.O.C. for a new way of recording metrics. 

Event-based metrics

This PR is a P.O.C. implementation of an event-based telemetry system that enables us to aggregate our data during analysis versus the current approach of sending aggregate for analysis. 

**Current Approach**
```
{
  merges: 34,
  merges_aborted: 5,
  merges_resulting_in_conflict: 3
}
```

**New Approach**
```
{
  events: [
    {
      type: merge_initiated,
      timestamp: 12345,
      source: top_menu
    },
    {
      type: merge_completed,
      timestamp: 123456,
      source: top_menu
    },
    ...
  ]
}
```

The new approach differs from the old way by recording much more granular data which will allow @desktop/analytics to perform more exploratory data analysis without being constrained by the meaning, as defined by the Desktop team, of the metrics.

I've defined the discriminated union, `InstrumentedEvent`, discriminated by `type` to represent an event that we'd like to measure in the app. At a minimum, any event that we'd like to measure should include a `timestamp` that represents the number of milliseconds since the Unix epoch.
	
```
InstrumentedEvent {
  type: string,
  timestamp: number,
  ...
}
```

We are able to push new events into our data store by making use of the new `TelemetryDoodad` class (the name can be changed to something more useful once we settle on a design). This class is responsible for storing all the events we care about by making use of the following function:
`push(event: InsturmentedEvent): Promise<void>`

I purposefully kept the interface to this class small to avoid all the boilerplate we currently do in the `StatsStore` (old way).
https://github.com/desktop/desktop/blob/c00061fb2c3613ce9e085b62a11e89f37dfcaf89/app/src/lib/stats/stats-store.ts#L578-L604
Note how all of these functions are fundamentally doing the same thing, mutating a value in the data store.

Here is a screenshot of the new data store that contains these event-based metrics.
//todo: insert screenshot

More info to come...


iAmWillShepherd:
Closing until after the next release.

PR #6937, opened 2019-02-25
teach branch switcher about rebase conflicts

### LINKED ISSUE #6891: branch switcher does not indicate rebase context
When you are in the middle of rebase conflicts in #6698,  you can dismiss the dialog and see this context in the header:

We don't have this problem with merge conflicts because you merge another branch into your current branch (so Git doesn't end up with a detached HEAD), and I'd like to change the text here to something more friendly ("rebasing", with the target branch name and some sort of progress summary to see how far through?).

There's no Octicon available to represent `rebase` but maybe something visual here might help to differentiate it from detached HEAD.

shiftkey:
## Overview

**Closes #6891**

## Description

I went with the simpler change rather than incorporating "rebase progress" which we don't currently surface in application state.

## Release notes

Notes: branch switcher now indicates when a rebase is in progress

### DIFF
--- a/app/src/ui/toolbar/branch-dropdown.tsx
+++ b/app/src/ui/toolbar/branch-dropdown.tsx
@@ -4,7 +4,7 @@ import { OcticonSymbol } from '../octicons'
 import { Repository } from '../../models/repository'
 import { TipState } from '../../models/tip'
 import { ToolbarDropdown, DropdownState } from './dropdown'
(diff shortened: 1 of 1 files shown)


shiftkey:
I added e5b1df648 because we never actually disable the branch switcher button (you can hover your cursor over it and see it light up). I kept it separate to the existing `canOpen` variable because that's also used to determine whether to show the down arrow for being able to see the branches list.

Issue #6313, opened 2018-11-29
Grayed out menu items, "Cannot read property 'toLowerCase' of undefined"


cheshire137:
## Description

I opened the dev tools and saw a JavaScript error repeated several times:

![image](https://user-images.githubusercontent.com/82317/49247291-897faa00-f3dc-11e8-9473-65fd75e2c530.png)

## Version

* GitHub Desktop:

1.5.1-beta0

<img width="274" alt="screen shot 2018-11-29 at 1 42 29 pm" src="https://user-images.githubusercontent.com/82317/49247369-bcc23900-f3dc-11e8-9033-7aa5f7b9aa7e.png">
<img width="247" alt="screen shot 2018-11-29 at 1 43 27 pm" src="https://user-images.githubusercontent.com/82317/49247379-c481dd80-f3dc-11e8-8276-fa8e5d9d6600.png">
<img width="287" alt="screen shot 2018-11-29 at 1 43 37 pm" src="https://user-images.githubusercontent.com/82317/49247394-cb105500-f3dc-11e8-9372-04c83684f902.png">
<img width="256" alt="screen shot 2018-11-29 at 1 43 50 pm" src="https://user-images.githubusercontent.com/82317/49247410-d3689000-f3dc-11e8-9a10-97db153ebb64.png">

* Operating system:

macOS High Sierra 10.13.6

## Steps to Reproduce

I'm not sure. I have been using the app all day with no troubles. I pushed up a branch and hit Cmd-R in the app to open the 'new pull request' page for the branch in my browser, and noticed no such page opened. I tried it a few more times, no luck. That's when I went to the menu and saw the option was grayed out, along with many other options.

### Expected Behavior

Menu options would work and not be grayed out.

### Actual Behavior

Many menu options are grayed out, including "About GitHub Desktop" which seems like it should always work.

iAmWillShepherd:
What happens if you restart the app? Is that state persisted?

cheshire137:
Nope, exiting and reopening the app made it work normally again!

shiftkey:
Looking at the stack trace, it seems like this is the problem code:

```ts
function userWithLowerCase(user: IGitHubUser): IGitHubUser {
  return {
    ...user,
    email: user.email.toLowerCase(),
    login: user.login.toLowerCase(),
  }
}
```

The stack trace itself refers to line 518, which is the end of the function - but I think that's it's own confused way of saying the second case is undefined (based on my attempts to breakpoint here):

[...]

PR #6940, opened 2019-02-25
log and drop found user when in an invalid state for caching

shiftkey:
## Overview

Adding some resiliency and tracing to `GitHubUserStore._loadAndCacheUser` because I don't entirely trust what it's trying to cache.

**Related to #6313 but I don't quite follow how it would result in a disabled app menu**

## Description

Keep in mind the original issue that the app is calling this function and one of these fields is `undefined`:

https://github.com/desktop/desktop/blob/c00061fb2c3613ce9e085b62a11e89f37dfcaf89/app/src/lib/stores/github-user-store.ts#L512-L518

The stack trace from #6313 mentions these three functions (albeit minified to strip the function names): `GitHubUserStore._loadAndCacheUser` -> `GitHubUserStore.cacheUser` -> `userWithLowerCase`.

And `cacheUser` is invoked only when it finds a valid object for `gitUser`:

https://github.com/desktop/desktop/blob/c00061fb2c3613ce9e085b62a11e89f37dfcaf89/app/src/lib/stores/github-user-store.ts#L233-L238

So what is `_loadAndCacheUser` doing that might be risky?

 - retrieve a cached user based on the received email address
 - if nothing found, use the GitHub API to find the author - using the given commit and email address

My working theory here is that one of these operations is returning a user with an empty `login`, which is triggering the action. The app can work fine without finding an user here, so this PR changes the behaviour of this area to:

  - emit a clear error message when it encounters a `gitUser` that's an invalid state, with details to help us isolate the issue if it gets reported
 - when the app encounters this invalid `gitUser`, it clears the variable so the app doesn't try and cache it

Armed with more details about this issue from someone who has encountered #6313, we can make this area more reliable once we know what piece of data we're missing...

## Release notes

Notes:

### DIFF
--- a/app/src/lib/stores/github-user-store.ts
+++ b/app/src/lib/stores/github-user-store.ts
@@ -241,30 +241,29 @@ export class GitHubUserStore extends BaseStore {
   private async findUserWithAPI(
     account: Account,
     repository: GitHubRepository,
-    sha: string | null,
+    sha: string,
     email: string
   ): Promise<IGitHubUser | null> {
     const api = API.fromAccount(account)
-    if (sha) {
-      const apiCommit = await api.fetchCommit(
-        repository.owner.login,
-        repository.name,
-        sha
-      )
-      if (apiCommit && apiCommit.author) {
-        const avatarURL = getAvatarWithEnterpriseFallback(
-          apiCommit.author.avatar_url,
-          email,
-          account.endpoint
-        )
 
-        return {
-          email,
-          avatarURL,
-          login: apiCommit.author.login,
-          endpoint: account.endpoint,
-          name: apiCommit.author.name || apiCommit.author.login,
-        }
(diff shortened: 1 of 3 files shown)


app/src/lib/stores/github-user-store.ts
```diff
@@ -177,7 +177,7 @@ export class GitHubUserStore extends BaseStore {
   public async _loadAndCacheUser(
     accounts: ReadonlyArray<Account>,
     repository: Repository,
-    sha: string | null,
```
shiftkey: We can drop the `| null` here because there's no longer any caller that is passing in `null`, so we can simplify the code paths in `findUserWithAPI`.

shiftkey:
Closing this out in favour of a more targeted fix now I have uncovered the root cause: https://github.com/desktop/desktop/issues/6313#issuecomment-467208202

PR #6945, opened 2019-02-26
make loadAndCacheUser more resilient to bad input

shiftkey:
## Overview

**Related to #6313**

## Description

This PR addresses a few things I uncovered during the investigation for #6313. I couldn't reproduce the "menus disabled" behaviour but I could trigger everything else.

Here's what I found:

 1. the GitHub API would return an empty object for the `author` for some commits, instead of `null`

This was fixed in 2484ff6 to add a runtime check that the object found is the expected shape and contents.

 2. our use of `IAPIUser` to represent a unified account does not reflect what was returned by the API. I uncovered three different shapes of data,

`IAPIUser` has been deprecated in favour of these three shapes in 8847ef6:

   - `IAPIFullIdentity` - this is the same shape as `IAPIUser` but is only used in two places - `/user` ( the details about the current user) and `/users/{user}` to find details about a GitHub account
   - `IAPIIdentity` - this is the subset of account information that is used in most places of the API - commits, pull requests, etc. No `name` or `email` is provided in this payload, irrespective of the privacy settings defined by the user. I didn't notice any strange issues downstream as a result of this change, but this is something to be mindful of while I'm doing more testing.
   - `IAPIOrganization` - the array of objects returned by `/user/orgs` is similar to `IAPIIdentity`, but for some reason omitted the `type` value (because only organizations are returned by this endpoint?). 

 3. this problem commit had it's email address set to `""` in Git

We have a few places where we look for a `null` email and bail out, so I also added checks for `""` in 9d0a484  to save API calls when searching for a user.

## Release notes

Notes: `[Fixed] caching GitHub users does not handle edge case where email is empty string`

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -362,6 +362,10 @@ export class API {
   public async searchForUserWithEmail(
     email: string
   ): Promise<IAPIIdentity | null> {
+    if (email.length === 0) {
+      return null
+    }
+
     try {
       const params = { q: `${email} in:email type:user` }
       const url = urlWithQueryString('search/users', params)
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -46,7 +46,7 @@ export interface IAPIRepository {
   readonly ssh_url: string
   readonly html_url: string
   readonly name: string
-  readonly owner: IAPIUser
+  readonly owner: IAPIIdentity
   readonly private: boolean
   readonly fork: boolean
   readonly default_branch: string
@@ -192,7 +192,7 @@ export interface IAPIPullRequest {
   readonly number: number
   readonly title: string
   readonly created_at: string
-  readonly user: IAPIUser
+  readonly user: IAPIIdentity
   readonly head: IAPIPullRequestRef
   readonly base: IAPIPullRequestRef
 }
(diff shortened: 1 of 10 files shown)


app/src/lib/api.ts
```diff
@@ -57,13 +59,42 @@ export interface IAPIRepository {
...
+  readonly url: string
+  readonly login: string
+  readonly avatar_url: string
+}
+
+/**
+ * Minimum subset of an identity returned by the GitHub API
+ */
+export interface IAPIIdentity {
+  readonly id: number
```
iAmWillShepherd: **Non-blocking**

Is it possible to use a sum/intersection type here to take advantage of the fact that these same fields are defined in the type above?

shiftkey: I didn't really see the value in coupling these together to save on typings because it felt like a different version of our mistake of assuming `IAPIUser` everywhere. Also, using `IAPIOrganization` (a special case returned by a specific API) to derive `IAPIIdentity` (a more general case used broadly in the API) doesn't make sense to me from a composition side.

ahxlaaaaiy:
‏‫من الـ iPhone الخاص بي‬

‫في 09‏/03‏/2019 الساعة 12:36 ص، كتب/كتبت ‏Brendan Forster ‏<notifications@github.com<mailto:notifications@github.com>>:‬


@shiftkey commented on this pull request.

In app/src/lib/api.ts<https://github.com/desktop/desktop/pull/6945#discussion_r263938484>:

> + *
+ * Because this is specific to one endpoint it omits the `type` member from
+ * `IAPIIdentity` that callers might expect.
+ */
+export interface IAPIOrganization {
+  readonly id: number
+  readonly url: string
+  readonly login: string
+  readonly avatar_url: string
+}
+
+/**
+ * Minimum subset of an identity returned by the GitHub API
+ */
+export interface IAPIIdentity {
+  readonly id: number


I didn't really see the value in coupling these together to save on typings because it felt like a different version of our mistake of assuming IAPIUser everywhere. Also, using IAPIOrganization (a special case returned by a specific API) to derive IAPIIdentity (a more general case used broadly in the API) doesn't make sense to me from a composition side.

PR #6946, opened 2019-02-26
update labels used in project to provide better context

shiftkey:
## Overview

This is a little idea that was inspired by exploring https://github.com/foosel/OctoPrint (cc @foosel for 📣✨) as it had an interesting way of organizing labels that I wanted to bring to the project.

For the moment I'm just using the documentation to sketch out these proposed changes, and once we're on board with the changes and the docs are good we can figure out the transition process around our other work.

## Description

I've focused on pain points that we're aware of from issue triage and general usage, and what any suggested:

 - the status of an issue is sometimes unclear
    -> labels should be more descriptive to guide maintainers and users
 - the meaning of a label is unclear
    -> we should better organize the labels we want to use in the project
 - it's not clear when a label should be used
    -> can we better clarify how labels are used within the project, based on our organic usages?

### A proposed solution

The big thing that I took away the issue tracker for `foosel/OctoPrint` was how labels were grouped with categories. I think this would be a great alternative to our current "bucket of labels" approach, as we could organize the labels to better represent how and when they should be used.

I'd recommend jumping ahead to the new document to get a feel for how things could be organized: [**Rendered Markdown**](https://github.com/desktop/desktop/blob/shiftkey-patch-1/docs/process/labels.md)

I'd like to focus on how the labels are organized initially, and whether the new categories make sense. I'll call out in this PR where these label changes impact our existing processes, and what this might mean, and we can talk about those in more detail once we're happy with the current set of labels.

## Release notes

Notes: no-notes

### DIFF
--- a/docs/process/labels.md
+++ b/docs/process/labels.md
@@ -3,40 +3,47 @@
 This section outlines the labels currently used in the project, and adds context
 about what each label represents.
 
-## Common labels
+## Areas of Work
 
 These labels are used for both issues and pull requests, and are assigned to
-help understand the type of work involved.
+help identify what area a task is associated with. This helps organize upcoming
+work and enables contributors to focus on a specific area of the project that
+they are passionate about.
 
-|                               | Label name       | Description |
-| ----------------------------- | -----------------| ----------- |
-| [:mag_right:][docs]           | `docs`           | Issues and pull requests related to documentation work on the project |
-| [:mag_right:][infrastructure] | `infrastructure` | Issues and pull requests related to scripts and tooling for GitHub Desktop |
-| [:mag_right:][tech-debt]      | `tech-debt`      | Issues and pull requests related to addressing technical debt or improving the codebase |
+|                                       | Label name               | Description |
+| ------------------------------------- | -------------------------| ----------- |
+| [:mag_right:][area:accessibility]     | `area:accessibility`     | Accessibility issues or improvements in the product |
+| [:mag_right:][area:dependencies]      | `area:dependencies`      | Upgrading dependencies that the project uses |
+| [:mag_right:][area:diffs]             | `area:diffs`             | Diff rendering and our use of [CodeMirror](https://codemirror.net/) |
+| [:mag_right:][area:docs]              | `area:docs`              | Documentation for the project |
+| [:mag_right:][area:electron]          | `area:electron`          | How we use [Electron](https://electronjs.org) in the product |
+| [:mag_right:][area:enhancements]      | `area:enhancements`      | Improvements to the product that address a specific problem for users |
+| [:mag_right:][area:external-websites] | `area:external-websites` | External websites that the GitHub maintainers are responsible for |
+| [:mag_right:][area:integrations]      | `area:integrations`      | Editor and shell integrations that ship in the product |
+| [:mag_right:][area:metrics]           | `area:metrics`           | Required to support analytics and metrics in the product |
+| [:mag_right:][area:performance]       | `area:performance`       | Improvements to the product to ensure it is responsive for users|
+| [:mag_right:][area:tech-debt]         | `area:tech-debt`         | Keep the application code maintainable so contributors can continue building features |
+| [:mag_right:][area:themes]            | `area:themes`            | The light or dark themes that ship in the product  |
+| [:mag_right:][area:tooling]           | `area:tooling`           | Scripts and tooling for the project |
 
-## Issue-specific labels
-
-This section is organized into sub-groups, as groups of labels relate to
-specific work being done in the issue tracker.
-
-### Issue triage
+## Issue triage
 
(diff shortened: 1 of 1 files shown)


docs/process/labels.md
```diff
@@ -46,80 +53,75 @@ get involved with.
...
-### Planning
+## Collaboration
 
-We use these labels to track tasks outside the usual flow of addressing bugs or
-implementing features:
+We use these labels to track tasks that represent specific interactions as part of a productive cross-functional team:
 
-|                                   | Label name           |  Description |
-| --------------------------------- | -------------------- |  ----------- |
-| [:mag_right:][meta]               | `meta`               | Issues used to co-ordinate tasks or discuss a feature before the required work is captured |
```
shiftkey: Looking at what is currently assigned to `meta`, I believe we can replace all of these with a specific `area` label that better suits what the work is. And if there's not an `area` that suits it, we can add one easily enough.

docs/process/labels.md
```diff
@@ -3,40 +3,47 @@
 This section outlines the labels currently used in the project, and adds context
 about what each label represents.
 
-## Common labels
+## Areas of Work
```
shiftkey: Areas is the biggest change here, and it feels like the best way to group together issues related to a specific category, while giving us the flexibility to create new labels or deprecate them down the track if they are unused.

Are there any other areas you can think of that we might list here?

docs/process/labels.md
```diff
@@ -3,40 +3,47 @@
...
 
 The triage process is how the maintainers process incoming issues and prioritize
 the work required to address the feedback raised. These labels help us to track
 the flow of an issue through this process
 
-|                                        | Label name                  | Description |
-| -------------------------------------- | ----------------------------| ----------- |
-| [:mag_right:][bug]                     | `bug`                     | Confirmed bugs or reports that are very likely to be bugs |
-| [:mag_right:][enhancement]             | `enhancement`             | Issues that propose to improve the app and solve a problem for users |
-| [:mag_right:][investigation-needed]    | `investigation-needed`    | Likely bugs, but haven't been reliably reproduced by a reviewer |
```
shiftkey: I've removed `investigation-needed` in favour of these labels that I think better reflect the transition from issue triage to product work:

  - `triage:potential-bug` - the step in between confirming a bug or requiring more information
  - `triage:confirmed-bug` - the bug has been confirmed by a reviewer using the information provided
  - `triage:not-a-bug` - the triage process is done and the issue isn't related to a bug, and it's up to someone else to figure out the next steps (e.g. ask clarifying questions about the use case, understand the value, and either move it to `area:enhancement` if it could be done in Desktop, or close the issue to indicate that it's not in scope)

docs/process/labels.md
```diff
@@ -3,40 +3,47 @@
...
+they are passionate about.
+
+|                                       | Label name               | Description |
+| ------------------------------------- | -------------------------| ----------- |
+| [:mag_right:][area:accessibility]     | `area:accessibility`     | Accessibility issues or improvements in the product |
+| [:mag_right:][area:dependencies]      | `area:dependencies`      | Upgrading dependencies that the project uses |
+| [:mag_right:][area:diffs]             | `area:diffs`             | Diff rendering and our use of [CodeMirror](https://codemirror.net/) |
+| [:mag_right:][area:docs]              | `area:docs`              | Documentation for the project |
+| [:mag_right:][area:electron]          | `area:electron`          | How we use [Electron](https://electronjs.org) in the product |
+| [:mag_right:][area:enhancements]      | `area:enhancements`      | Improvements to the product that address a specific problem for users |
```
shiftkey: This might be the most radical change to how we track enhancements currently, as I worry we just slap this label on everything that's "not a bug" without being more thoughtful and explicit about proposed improvements to the product.

@billygriffin I'd specifically love your input about this change and whether this makes your life easier or harder.

tierninho:
@shiftkey I like this, as it allows for better grouping and enhances the search functionality. I agree that the current labeling could use some work, and this a great step forward. Nice ✨

docs/process/labels.md
billygriffin: I think my biggest concern about this relates to prioritization. Currently, the labels associated with prioritization (priority 1 through 3) relate to bugs, not enhancements (although I've seen that applied inconsistently in a few cases). We don't (yet) have a concept of prioritization for enhancements, and that's definitely something I'd love to have at some point. My dream would be that people could search for the combo of `help wanted` and some sense of `high priority` and have a list of ~10 or so things they could pick from - especially @desktop/comrades.

shiftkey: Let's try and sync up next week about this to see if we can figure out other ideas. I'm not strongly attached to this change.

shiftkey:
Closing this out in favour of more thought and more targeted PRs to tackle specific parts of the project logistics

PR #6947, opened 2019-02-26
update TextDiff update logic

### LINKED ISSUE #6938: Line numbers disappear
I'm in version 1.6.3-beta2 and line numbers disappear from the diff sometimes. I viewed a changed file in Desktop, then switched to my editor and modified one of the changed lines, and when I switched back to Desktop, the diff updated to show my latest change but at the same time the line numbers went away.

image

outofambit:
## Overview

Closes #6938 

The primary change in this PR is to call `updateViewport()` at the precisely right time in the codemirror update cycle, and then only when we are in the specific case of a diff that has changed, but hasn't changed in number of lines.

## Description

- added a hook for after a `swapDoc` event (basically whenever we change the text in the diff viewer)
- utilize hook to check for the case mentioned above and update the viewport so we don't lose our gutter

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -705,14 +705,13 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
     if (canSelect(this.props.file)) {
       if (
         !canSelect(prevProps.file) ||
-        this.props.file.selection !== prevProps.file.selection
+        this.props.file.selection !== prevProps.file.selection ||
+        this.props.text === prevProps.text
       ) {
         // If the text has changed the gutters will be recreated
         // regardless but if it hasn't then we'll need to update
         // the viewport.
-        if (this.props.text === prevProps.text) {
-          this.updateViewport()
-        }
+        this.updateViewport()
       }
     }
 
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -702,12 +702,14 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
       return
     }
 
-    if (canSelect(this.props.file)) {
-      if (
-        !canSelect(prevProps.file) ||
-        this.props.file.selection !== prevProps.file.selection ||
-        this.props.text === prevProps.text
-      ) {
+    const { file } = this.props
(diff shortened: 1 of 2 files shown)


tierninho:
Tested, and the disappearing text issue is now fixed 🎉

app/src/ui/diff/text-diff.tsx
```diff
@@ -705,14 +705,13 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
     if (canSelect(this.props.file)) {
       if (
         !canSelect(prevProps.file) ||
-        this.props.file.selection !== prevProps.file.selection
+        this.props.file.selection !== prevProps.file.selection ||
+        this.props.text === prevProps.text
       ) {
         // If the text has changed the gutters will be recreated
```
billygriffin: It seems like this comment is potentially no longer correct?

outofambit: it's still correct! we do need to call `updateViewport()` when the text hasn't changed. The primary change in this PR is to call `updateViewport()` when the selection hasn't changed OR the text is the same. Previously it was AND, which means we missed some needed updates for the underlying code mirror instance.

outofambit: I'm going to make this logic a bit more readable in my next commit!

app/src/ui/diff/text-diff.tsx
```diff
@@ -702,17 +702,18 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
-        this.props.file.selection !== prevProps.file.selection
-      ) {
+    const { file } = this.props
+    const { file: prevFile } = prevProps
+    if (canSelect(file)) {
+      if (!canSelect(prevFile)) {
+        this.updateViewport()
+      } else if (file.selection !== prevFile.selection) {
+        this.updateViewport()
+      } else if (this.props.text === prevProps.text) {
```
iAmWillShepherd: Minor 🎨
```ts
const { file, text } = this.props
const { file: prevFile, text: prevText } = prevProps
```

iAmWillShepherd:
Merging early so we this can get into beta for 1.6.3.

maxbed:
сре, 27. феб 2019. 18.32 William Shepherd <notifications@github.com> је
написао/ла:

PR #6948, opened 2019-02-26
Add stash entry type

iAmWillShepherd:
Adds a type to represent a stash object.

See tracking issue: #6987

### DIFF
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -1,4 +1,4 @@
-export interface StashEntry {
+export interface IStashEntry {
   /** The name of the branch at the time the entry was created. */
   readonly branchName: string
 
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -1,0 +1,7 @@
+export interface StashEntry {
+  /** The name of the branch at the time the entry was created. */
+  readonly branchName: string
+
+  /** The tip commit SHA at the time the entry was created. */
+  readonly sha: string
+}
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -3,5 +3,5 @@ export interface IStashEntry {
   readonly branchName: string
 
   /** The SHA of the commit object created as a result of stashing. */
-  readonly sha: string
+  readonly stashSha: string
 }
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -2,6 +2,6 @@ export interface IStashEntry {
   /** The name of the branch at the time the entry was created. */
   readonly branchName: string
 
-  /** The tip commit SHA at the time the entry was created. */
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
Note to whoever reviews this PR:
I'm looking for feedback on whether this is the right place for this type. I think a case can be made to put this in the `models` directory, but I threw it in this new file as a start.

shiftkey:
> I think a case can be made to put this in the `models` directory, but I threw it in this new file as a start.

@iAmWillShepherd I'd love to understand how broadly you imagine this type being used, because depending on the answer I can see a couple of options:

 - if you can see this being returned by Git operations, stored in application state, and used in components, then you could make a case for this living under `app/src/models`
 - if you can see thing being returned by the Git operation, but being mapped to something else that is then used for application state and components, I'd keep this under `app/src/lib/git` for the moment

Note that this might change over time as you get the functionality in place, but for now I'd keep things as simple and localized as possible.

iAmWillShepherd:
@outofambit, @shiftkey at the moment, this type will be used to represent a stash entry from git, in the future, I may use it in app-state, but it's a bit early for me to know for sure. I'm 👍 for keeping it in `lib/git` until it's needed in more than one place.

shiftkey:
@iAmWillShepherd any other notes about what you'd like reviewed?

iAmWillShepherd:
> @iAmWillShepherd any other notes about what you'd like reviewed?

Actually yes, I'm wondering if the `sha` is necessary.

My original plan for including the SHA was to handle scenarios where a branch is renamed or if we have an ambiguous branch name. My thinking was that we could use the SHA to verify if we are using the right branch.

shiftkey:
> My thinking was that we could use the SHA to verify if we are using the right branch.

What does "right branch" mean in this context?

iAmWillShepherd:
> > My thinking was that we could use the SHA to verify if we are using the right branch.
> 
> What does "right branch" mean in this context?

Right branch in this context means I have the correct stash entry for the current branch. After thinking on this some more, I think I'm going to use the SHA to refer to the stash entry instead of using it to store the tip of the branch.

app/src/lib/git/stash.ts
```diff
@@ -0,0 +1,7 @@
+export interface IStashEntry {
+  /** The name of the branch at the time the entry was created. */
+  readonly branchName: string
+
+  /** The SHA of the commit object created as a result of stashing. */
+  readonly sha: string
```
shiftkey: 💭 could this be something more descriptive than `sha`?

iAmWillShepherd: Changed - https://github.com/desktop/desktop/pull/6948/commits/01b9731821e6da77b4e2bd3d3c873cd3c32fe819

iAmWillShepherd:
🍎

shiftkey:
@iAmWillShepherd feel free to merge whenever you're ready to move on

PR #6951, opened 2019-02-26
the last set of rebase metrics for this release

shiftkey:
## Overview

**Related to #6550**

## Description

This PR implements the remaining metrics for `1.6.3`:

> 2. Number of times a user encounters a conflict while pull rebasing.
>  2a. Number of times a pull rebase is completed successfully after a conflict.
>   2b. Number of times a pull rebase is aborted after a conflict.

The other metrics are in #6893.

I'm going to follow up with PRs for the documentation once both of these PRs are merged, because there's also backend changes to land to consume these metrics.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -83,6 +83,8 @@ const DefaultDailyMeasures: IDailyMeasures = {
   createPullRequestCount: 0,
   rebaseConflictsDialogDismissalCount: 0,
   rebaseConflictsDialogReopenedCount: 0,
+  rebaseAbortedAfterConflictsCount: 0,
+  rebaseSuccessAfterConflictsCount: 0,
 }
 
 interface IOnboardingStats {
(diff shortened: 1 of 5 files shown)


app/src/lib/stats/stats-store.ts
```diff
@@ -257,6 +259,8 @@ type DailyStats = ICalculatedStats &
 export interface IStatsStore {
   recordMergeAbortedAfterConflicts: () => void
   recordMergeSuccessAfterConflicts: () => void
+  recordRebaseAbortedAfterConflicts: () => void
```
iAmWillShepherd: **Non-blocking question**
Is this intended to make testing easier?

shiftkey: This is because we use `IStatsStore` rather than `StatsStore` directly in `changes-state.ts`.

The benefit of this interface is that we can write tests to verify these side effects are invoked in the right scenarios.

Issue #6960, opened 2019-02-27
Warn if there will be conflicts prior to starting a rebase


shiftkey:
Extracted from #6551

Similar to what we show in the Merge Branch dialog:

<img width="307" src="https://user-images.githubusercontent.com/359239/53506317-1db8e600-3a8c-11e9-880d-48758bc99a0c.png">

outofambit:
@billygriffin and I discussed this last week and we've agreed that this is not a high impact enough feature to invest more time and energy on. The signal to noise ratio of this is lower than the merge conflict warning UI, and is more likely only to be seen by user's who don't need to know if there will be conflicts or not before they start a rebase. Closing.

Issue #6959, opened 2019-02-27
[rebase] when branch selected, count number of commits that will be rebased


shiftkey:
Extracted from #6551

Similar to what we show in the Merge Branch dialog:

<img width="307" src="https://user-images.githubusercontent.com/359239/53506317-1db8e600-3a8c-11e9-880d-48758bc99a0c.png">

Issue #6032, opened 2018-10-30
add support for setting JSX for title on Dialog props


shiftkey:
We want to avoid contributors using `Dialog` and nesting `DialogHeader` within if they only want the default behaviour. See https://github.com/desktop/desktop/pull/6027#discussion_r229337387 for more discussion.

shiftkey:
I was going to propose using `react-to-string` here which seems to fit the bill, based on the example:

```js
import React, { Component } from 'react';
import reactToString from 'react-to-string';
 
class App extends Component {
  render() {
    return <div>Hello world!</div>
  }
}
 
reactToString(<App />);
// Returns "Hello world!"
 
reactToString(<p>Are you interested in this library? <a href="signup">Click here!</a></p>);
// Returns "Are you interested in this library? Click here!"
```

But the library does not have a license, so I've opened https://github.com/kevinzwhuang/react-to-string/issues/3 to ask them to define one before I look at pulling this into our codebase.

shiftkey:
> It looks like that doesn’t actually _render_ the element so the code might not work as described.

To be clear,  I'm only looking for a string-like representation of the element for situations where we need something distinct to work with (the original issue was using the title to generate a unique id). I'm not looking to actually render the elements.

shiftkey:
This was the snippet I used in #6027 as a starting point:

```diff
diff --git a/app/src/ui/dialog/dialog.tsx b/app/src/ui/dialog/dialog.tsx
index 63162f312..9ff9cdee7 100644
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -31,7 +31,7 @@ interface IDialogProps {
    * By omitting this consumers may use their own custom DialogHeader
    * for when the default component doesn't cut it.
    */
-  readonly title?: string
+  readonly title?: string | JSX.Element
 
   /**
    * Whether or not the dialog should be dismissable. A dismissable dialog
@@ -178,9 +178,11 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
       this.setState({ titleId: undefined })
     }
 
-    if (this.props.title) {
+    if (this.props.title != null) {
+      // TODO: how can we get a string-like representation of a JSX element?
+      const id = typeof this.props.title === 'string' ? this.props.title : '???'
       this.setState({
-        titleId: createUniqueId(`Dialog_${this.props.id}_${this.props.title}`),
[...]

PR #6961, opened 2019-02-27
[rebase] menu item and dialog to start rebase operation

### LINKED ISSUE #6551: "Rebase current branch" dialog and flow
This is the flow that a user can start from the **Branch** | **Rebase current branch...** menu item which lets the user:

 - select the base branch to merge the current branch onto
 - see at a glance how many commits will be rebased
 - view the progress of the rebase
 - be interrupted by the rebase if it encounters conflicts (#6548 is related to this step)
 - see some sort of success indicator when it completes

The mocks for this can be seen in this comment: https://github.com/desktop/desktop/issues/5953#issuecomment-437527414

TODO:

 - [x] flesh out the checklist of tasks

### LINKED ISSUE #7012: Update Rebasing terms and UI
Rebasing is currently using two prepositions, `on` and `onto` in the UI. I suggest we use `onto` as it is more descriptive in how a rebase looks visually. (We also use `into` when doing merge conflicts).

Three places were on/onto is used... 
- a mention is in the success message banner
- and remaining two are here:
screen shot 2019-03-04 at 3 54 50 pm

Lastly, we should fix the spacing of the spinning icon and the text in the button:
screen shot 2019-03-04 at 3 56 29 pm

76d48b925f, mac

shiftkey:
## Overview

**Closes #6551**
**Closes #7012**
**Closes #6961**

## Description

This is the first part of enabling the user to rebase their current branch from within the app in a guided fashion, and I've omitted much of the background computation from #6959 and #6960 so that I can have something baked to help with testing.

screen shot 2019-02-27 at 12 33 30 pm

screen shot 2019-02-27 at 12 33 37 pm

screen shot 2019-02-27 at 12 33 48 pm

The menu item is enabled only for development currently.

Testing: 

 - [x] can start a rebase, and dialog closes because no conflicts with success banner shown
 - [x] can start a rebase, and dialog switches to show conflicts if encountered
 - [x] can start a rebase, can address conflicts, and continue rebase until success banner shown
 - [x] address `TODO` comments

## Release notes

Notes:

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -114,10 +114,26 @@ export async function abortRebase(repository: Repository) {
   await git(['rebase', '--abort'], repository.path, 'abortRebase')
 }
 
+/** The app-specific results from attempting to rebase a repository */
 export enum RebaseResult {
+  /**
+   * Git completed the rebase without reporting any errors, and the caller can
+   * signal success to the user.
+   */
   CompletedWithoutError = 'CompletedWithoutError',
+  /**
+   * The rebase encountered conflicts while attempting to rebase, and these
+   * need to be resolved by the user before the rebase can continue.
+   */
   ConflictsEncountered = 'ConflictsEncountered',
+  /**
+   * The rebase was not able to continue as tracked files were not staged in
+   * the index.
+   */
   OutstandingFilesNotStaged = 'OutstandingFilesNotStaged',
+  /**
+   * The rebase was aborted and the caller should do any additional cleanup.
+   */
   Aborted = 'Aborted',
 }
 
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -132,7 +132,9 @@ export enum RebaseResult {
    */
   OutstandingFilesNotStaged = 'OutstandingFilesNotStaged',
   /**
-   * The rebase was aborted and the caller should do any additional cleanup.
+   * The rebase was not attempted because it could not check the status of the
+   * repository. The caller needs to confirm the repository is in a usable
+   * state.
    */
   Aborted = 'Aborted',
 }
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -181,10 +181,10 @@ export async function continueRebase(
 
   if (trackedFiles.length === 0) {
     const rebaseHead = Path.join(repository.path, '.git', 'REBASE_HEAD')
-    const rebaseCurrentCommit = await FSE.readFile(rebaseHead)
+    const rebaseCurrentCommit = await FSE.readFile(rebaseHead, 'utf8')
 
     log.warn(
-      `[rebase] no tracked changes to commit for ${rebaseCurrentCommit}, continuing rebase but skipping this commit`
+      `[rebase] no tracked changes to commit for ${rebaseCurrentCommit.trim()}, continuing rebase but skipping this commit`
     )
 
     const result = await git(
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -97,7 +97,7 @@ export async function rebase(
   repository: Repository,
   baseBranch: string,
   targetBranch: string
-): Promise<ContinueRebaseResult> {
+): Promise<RebaseResult> {
   const result = await git(
     ['rebase', baseBranch, targetBranch],
     repository.path,
@@ -115,7 +115,7 @@ export async function abortRebase(repository: Repository) {
   await git(['rebase', '--abort'], repository.path, 'abortRebase')
 }
 
-export enum ContinueRebaseResult {
+export enum RebaseResult {
   CompletedWithoutError = 'CompletedWithoutError',
   ConflictsEncountered = 'ConflictsEncountered',
   OutstandingFilesNotStaged = 'OutstandingFilesNotStaged',
@@ -126,17 +126,17 @@ const rebaseEncounteredConflictsRe = /Resolve all conflicts manually, mark them 
 
 const filesNotMergedRe = /You must edit all merge conflicts and then\nmark them as resolved/
 
-function parseRebaseResult(result: IGitResult): ContinueRebaseResult {
+function parseRebaseResult(result: IGitResult): RebaseResult {
   if (result.exitCode === 0) {
-    return ContinueRebaseResult.CompletedWithoutError
+    return RebaseResult.CompletedWithoutError
   }
 
   if (rebaseEncounteredConflictsRe.test(result.stdout)) {
-    return ContinueRebaseResult.ConflictsEncountered
+    return RebaseResult.ConflictsEncountered
   }
 
   if (filesNotMergedRe.test(result.stdout)) {
-    return ContinueRebaseResult.OutstandingFilesNotStaged
+    return RebaseResult.OutstandingFilesNotStaged
   }
 
   throw new Error(`Unhandled result found: '${JSON.stringify(result)}'`)
@@ -154,7 +154,7 @@ export async function continueRebase(
   repository: Repository,
   files: ReadonlyArray<WorkingDirectoryFileChange>,
   manualResolutions: ReadonlyMap<string, ManualConflictResolution> = new Map()
-): Promise<ContinueRebaseResult> {
+): Promise<RebaseResult> {
   // apply conflict resolutions
   for (const [path, resolution] of manualResolutions) {
     const file = files.find(f => f.path === path)
@@ -177,7 +177,7 @@ export async function continueRebase(
     log.warn(
       `[rebase] unable to get status after staging changes, skipping any other steps`
     )
-    return ContinueRebaseResult.Aborted
+    return RebaseResult.Aborted
   }
 
   const trackedFiles = status.workingDirectory.files.filter(
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -1,8 +1,9 @@
 import * as Path from 'path'
 import * as FSE from 'fs-extra'
+import { GitError } from 'dugite'
 
 import { Repository } from '../../models/repository'
-import { git } from './core'
+import { git, IGitResult } from './core'
 import {
   WorkingDirectoryFileChange,
   AppFileStatusKind,
@@ -10,7 +11,7 @@ import {
 import { ManualConflictResolution } from '../../models/manual-conflict-resolution'
 import { stageManualConflictResolution } from './stage'
 import { stageFiles } from './update-index'
-import { GitError, IGitResult } from 'dugite'
+
 import { getStatus } from './status'
 import { RebaseContext } from '../../models/rebase'
 
@@ -102,8 +103,6 @@ export async function rebase(
     ['rebase', baseBranch, targetBranch],
     repository.path,
     'rebase',
-    // TODO: what about using successExitCodes here?
-    // successExitCodes: new Set([0, 1, 128]),
     { expectedErrors: new Set([GitError.RebaseConflicts]) }
   )
 
@@ -122,20 +121,16 @@ export enum RebaseResult {
   Aborted = 'Aborted',
 }
 
-const rebaseEncounteredConflictsRe = /Resolve all conflicts manually, mark them as resolved/
-
-const filesNotMergedRe = /You must edit all merge conflicts and then\nmark them as resolved/
-
 function parseRebaseResult(result: IGitResult): RebaseResult {
   if (result.exitCode === 0) {
     return RebaseResult.CompletedWithoutError
   }
 
-  if (rebaseEncounteredConflictsRe.test(result.stdout)) {
+  if (result.gitError === GitError.RebaseConflicts) {
     return RebaseResult.ConflictsEncountered
   }
 
-  if (filesNotMergedRe.test(result.stdout)) {
+  if (result.gitError === GitError.UnresolvedConflicts) {
     return RebaseResult.OutstandingFilesNotStaged
   }
 
@@ -197,24 +192,25 @@ export async function continueRebase(
       repository.path,
       'continueRebaseSkipCurrentCommit',
       {
-        successExitCodes: new Set([0, 1, 128]),
+        expectedErrors: new Set([
+          GitError.RebaseConflicts,
+          GitError.UnresolvedConflicts,
+        ]),
       }
     )
 
     return parseRebaseResult(result)
   }
 
-  // TODO: there are some cases we need to handle and surface here:
-  //  - rebase continued and completed without error
-  //  - rebase continued but encountered a different set of conflicts
-  //  - rebase could not continue as there are outstanding conflicts
-
   const result = await git(
     ['rebase', '--continue'],
     repository.path,
     'continueRebase',
     {
-      successExitCodes: new Set([0, 1, 128]),
+      expectedErrors: new Set([
+        GitError.RebaseConflicts,
+        GitError.UnresolvedConflicts,
+      ]),
     }
   )
 
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -91,19 +91,23 @@ export async function getRebaseContext(
  * and it will probably have a different commit history.
  *
  * @param baseBranch the ref to start the rebase from
- * @param featureBranch the ref to rebase onto `baseBranch`
+ * @param targetBranch the ref to rebase onto `baseBranch`
  */
 export async function rebase(
   repository: Repository,
   baseBranch: string,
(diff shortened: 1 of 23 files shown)


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -334,6 +335,15 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+Shift+M',
         click: emit('merge-branch'),
       },
+      {
+        label: __DARWIN__
+          ? 'Rebase Current Branch…'
+          : 'R&ebase current branch…',
+        id: 'rebase-branch',
+        accelerator: 'CmdOrCtrl+Shift+E',
```
shiftkey: I settled on `E` here because `R` or `B` are already assigned to other shortcuts

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -623,6 +623,52 @@ export class Dispatcher {
...
+    baseBranch: string,
+    targetBranch: string
+  ) {
+    const stateBefore = this.repositoryStateManager.get(repository)
+
+    const beforeSha = getTipSha(stateBefore.branchesState.tip)
+
+    log.info(`[rebase] starting rebase for ${beforeSha}`)
+
+    // TODO: this can happen very quickly for a trivial rebase or an OS with
```
shiftkey: Going to keep this around because I have #6962 to tackle 🔜

app/src/lib/git/rebase.ts
```diff
@@ -91,48 +92,46 @@ export async function getRebaseContext(
...
-const filesNotMergedRe = /You must edit all merge conflicts and then\nmark them as resolved/
-
-function parseRebaseResult(result: IGitResult): ContinueRebaseResult {
+function parseRebaseResult(result: IGitResult): RebaseResult {
   if (result.exitCode === 0) {
-    return ContinueRebaseResult.CompletedWithoutError
+    return RebaseResult.CompletedWithoutError
   }
 
-  if (rebaseEncounteredConflictsRe.test(result.stdout)) {
```
shiftkey: I bumped `dugite` to the latest version to replace this regex inspection with proper error codes for the known rebase problem scenarios: https://github.com/desktop/dugite/pull/306

tierninho:
Tested the new menu option/shortcut as of 76d48b925f969aad3c192929f4db3574e5b9a46b and found no issues. 

- Curious if it makes sense to allow the user to rebase the same two branches twice in a row?

shiftkey:
> Curious if it makes sense to allow the user to rebase the same two branches twice in a row?

I don't think it's worth actively trying to block this case unless there's downsides to this. The second rebase would be a clean rebase (because the user resolved any conflicts in the first rebase).

tierninho:
If a merge conflict is in progress and is currently in manual resolution mode (banner present), then the `Rebase current branch` menu option should not be enabled as it results in an error message:

```
install.ts:23 `git rebase sdfsdfsdfsdfsdfsdfsd test-pr-creation` exited with an unexpected code: 1.
.gitignore: needs merge
UnitTestProject3/UnitTestProject3/UnitTest2.cs: needs merge

error: cannot rebase: You have unstaged changes.
error: additionally, your index contains uncommitted changes.
error: Please commit or stash them.
```

image

(⚠️ Note this is part of a wider problem that might want to be addressed elsewhere as multiple menu options are selectable in this state, and doing so can leads to other errors).

shiftkey:
> (⚠️ Note this is part of a wider problem that might want to be addressed elsewhere as multiple menu options are selectable in this state, and doing so can leads to other errors).

👍 you can currently try merging a branch while also dealing with merge conflicts. I'll spin this out to an issue to track tomorrow.

shiftkey:
I've opened #7017 to audit the menu items and figure out which should be enabled when dealing with conflicts.

shiftkey:
Marking this as `time-sensitive` because it's key to all rebase functionality in `1.7.0`

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+    const disabled = selectedBranchIsNotCurrentBranch
+
+    // the amount of characters to allow before we truncate was chosen arbitrarily
+    const currentBranchName = truncateWithEllipsis(
+      this.props.currentBranch.name,
+      40
+    )
+
+    const loading = this.state.isRebasing ? <Loading /> : null
```
niik: Can we use the `<Dialog loading=` attribute here instead of rendering a custom loading component please?

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+            onFilterTextChanged={this.onFilterTextChanged}
+            selectedBranch={selectedBranch}
+            onSelectionChanged={this.onSelectionChanged}
+            canCreateNewBranch={false}
+            renderBranch={this.renderBranch}
+          />
+        </DialogContent>
+        <DialogFooter>
+          <ButtonGroup>
+            <Button type="submit" disabled={disabled}>
```
niik: Can we please use the `<Dialog disabled=` attribute here instead of explicitly disabling the button?

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+              {loading}
+              Rebase <strong>
+                {currentBranch ? currentBranch.name : ''}
```
niik: I haven't checked this branch out so forgive me if I'm reading this wrong but it doesn't look like the `currentBranch` prop from which this is derived can be null or undefined. Could we omit the check here?

app/styles/ui/dialogs/_rebase-conflicts.scss
```diff
@@ -0,0 +1,124 @@
...
+      .green-circle:last-child {
+        margin-left: auto;
+        margin-top: var(--spacing);
+        flex-shrink: 0;
+        flex: 0 1 auto;
+      }
+    }
+
+    .unmerged-file-status-resolved .file-conflicts-status {
+      color: $green;
```
niik: Are the use of color constants being tracked in some other issue? We should never be using primer SCSS variables outside of `_variables.scss`. I suggest we change these right now to use our variables such that it's themeable.

```suggestion
      color: var(--color-new);
```

Side-note, I thought we had to do `#{$green}` these days?

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+    const loading = this.state.isRebasing ? <Loading /> : null
+
+    return (
+      <Dialog
+        id="rebase"
+        onDismissed={this.props.onDismissed}
+        onSubmit={this.startRebase}
+      >
+        <DialogHeader
```
niik: Manually rendering a header like this will break the accessibility support built into the dialog probably causing screen readers to no longer announce the title of the dialog accurately. The `<dialog>` element is marked with an `aria-labelledby` attribute which points to the dialog header id (which is creating on the fly by the dialog component).

The way I'd prefer this to be implemented is by changing the signature of the `Dialog` component `title` prop to allow `string | JSX.Element` and passing the custom markup through that.

```tsx
<Dialog
  id="rebase"
  onSubmit={this.startRebase}
  title={
    <>
      Rebase <strong>{currentBranchName}</strong> onto…
    </>
  }
>
```

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+    const selectedBranchIsNotCurrentBranch =
+      selectedBranch === null ||
+      currentBranch === null ||
+      currentBranch.name === selectedBranch.name
+
+    const disabled = selectedBranchIsNotCurrentBranch
+
+    // the amount of characters to allow before we truncate was chosen arbitrarily
+    const currentBranchName = truncateWithEllipsis(
```
niik: I'm not gonna hold up this review because of this but I'd much prefer it if we'd use CSS to dynamically truncate this string rather than choosing an arbitrary number of characters since we use a variable width font.

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * Returns the branch to use as the selected branch
+   *
+   * The initial branch is used if passed
+   * otherwise, the default branch will be used iff it's
+   * not the currently checked out branch
+   */
+  private resolveSelectedBranch() {
```
niik: Could we extract this out into a regular function that takes these three parameters rather than accessing the instance variable props? Not a huge fan of invoking instance methods from the constructor.

app/styles/ui/dialogs/_rebase-conflicts.scss
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    .unmerged-file-status-conflicts {
+      .file-conflicts-status {
+        color: $orange;
```
niik:
```suggestion
        color: var(--color-conflicted);
```

app/styles/ui/dialogs/_rebase-conflicts.scss
```diff
@@ -0,0 +1,124 @@
...
+      }
+
+      .command-line-hint {
+        color: $gray;
```
niik:
```suggestion
        color: var(--text-secondary-color);
```

app/styles/ui/dialogs/_rebase-conflicts.scss
```diff
@@ -0,0 +1,124 @@
...
+    flex-flow: row nowrap;
+    padding: var(--spacing) 0 var(--spacing-double);
+
+    .message {
+      padding-left: var(--spacing);
+      padding-top: var(--spacing-third);
+    }
+  }
+
+  .cli-link {
```
niik: I don't see this class being used anywhere?

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+  public render() {
+    const selectedBranch = this.state.selectedBranch
+    const currentBranch = this.props.currentBranch
+
+    const selectedBranchIsNotCurrentBranch =
+      selectedBranch === null ||
+      currentBranch === null ||
+      currentBranch.name === selectedBranch.name
+
+    const disabled = selectedBranchIsNotCurrentBranch
```
niik: This should probably be moved below the `loading` constant and be conditional on that, right?

```suggestion
    const disabled = selectedBranchIsNotCurrentBranch || loading
```

Otherwise the button will still be clickable during a rebase

app/styles/ui/dialogs/_rebase-conflicts.scss
```diff
@@ -0,0 +1,124 @@
...
+
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+  }
+
+  .arrow-menu {
+    .octicon {
+      width: 10px;
```
niik: If we're ever setting an explicit width on an octicon we also need to set the height or else vertical alignment will be off. Normally we should never set the width but rather use the height and have the width be automatically calculated using the correct aspect ratio. This icon, unfortunately, does not scale cleanly to 10px width though so we should really add a custom vector here but for now let's just fix the vertical alignment by adding a height.
```suggestion
      width: 10px;
      height: 14px;
```

app/styles/ui/dialogs/_rebase-conflicts.scss
shiftkey: > We should never be using primer SCSS variables outside of `_variables.scss` 

Agreed. I crimped these styles from `app/styles/ui/dialogs/_merge-conflicts.scss` and these styles will go away in #7030 after merging them together, but I'll clean this up and check for any other usages that we should tidy up outside of this PR.

> Side-note, I thought we had to do #{$green} these days?

Me too, and I'm surprised this didn't get raise any issues. As a sanity check I looked at the CSS from `1.6.3` and these are transformed fine:

app/styles/ui/dialogs/_rebase-conflicts.scss
shiftkey: I'll also check whether `app/styles/ui/dialogs/_merge-conflicts.scss` needs this change too

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: This feels like an earlier discussion around this area #6032 - it's been a while since I looked at this prop but I'll give it another shot and see what comes out of it.

app/src/lib/git/rebase.ts
```diff
@@ -91,48 +92,46 @@ export async function getRebaseContext(
...
+  return parseRebaseResult(result)
 }
 
 /** Abandon the current rebase operation */
 export async function abortRebase(repository: Repository) {
   await git(['rebase', '--abort'], repository.path, 'abortRebase')
 }
 
-export enum ContinueRebaseResult {
+export enum RebaseResult {
```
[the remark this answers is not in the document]
shiftkey: Added in 7e6c7851a6bde77d48e20e51506b5e6f9b71acfb

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: Addressed in 9809329c7d4cef3c5c7ec76dea26592283db2223

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: Addressed in 9809329c7d4cef3c5c7ec76dea26592283db2223

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: Addressed in 9809329c7d4cef3c5c7ec76dea26592283db2223

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: This should be a bit easier to follow in de796e24f8285f8135578b822415584312cd9b8d

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: cc @outofambit for :eyes: and 💭

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: Done in d3f2dbdd7a50e57dbcf9fe0aeb2c4af55dcc61ba

app/styles/ui/dialogs/_rebase-conflicts.scss
shiftkey: Cleaned up in ac9ddb4b4500c9b28f7e582497ba7d76a6029ecc

app/src/ui/rebase/rebase-branch-dialog.tsx
shiftkey: Updated the API in 6cf4a2aab1a4a648c7becdd1c4b6777bd1534bb2 and migrated over to  7f4cedd5cd7d71b8cef35ae06a7c956a9715ded9

shiftkey:
FYI: I've extracted the feedback from this PR that can also be applied to the Merge Conflicts dialog into #7043 - that can be picked up once this and #7030 are merged.

app/src/ui/dialog/dialog.tsx
```diff
@@ -187,8 +187,12 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
     }
 
     if (this.props.title) {
+      // createUniqueId handles static strings fine, so in the case of receiving
+      // a JSX element for the title we can just pass in a fixed value rather
+      // than trying to generate a string from an arbitrary element
+      const id = typeof this.props.title === 'string' ? this.props.title : '???'
```
niik: Not gonna hold up this from landing because of this but `???` stands out to me here. I would have preferred if we either left this code as it was, or that we made this constant something like `JSX` rather than triple question marks. It's entirely stylistic though so I'm gonna go ahead and approve and land this.

app/src/ui/rebase/rebase-branch-dialog.tsx
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    return (
+      <Dialog
+        id="rebase"
+        onDismissed={this.props.onDismissed}
+        onSubmit={this.startRebase}
+        loading={loading}
+        disabled={disabled}
+        dismissable={true}
```
niik: Again, stylistic, but dismissible is the default value so we could just omit this.

PR #6969, opened 2019-02-27
Copy commit message props to state using componentWillReceiveProps

### LINKED ISSUE #6390: Clicking "Undo" button doesn't populate summary field

humphd:
## Overview

This fixes the issue where commit message summary and description UI is lost when committing/undoing commits.  The change alters how the summary and description are mirrored in the component's internal `state` from `props` using  `componentWillReceiveProps()`.

Closes #6390

## Description

I've written about this at length in a [walkthrough write-up](https://github.com/humphd/desktop/tree/good-first-experience-issue-6390#walkthrough-fixing-a-bug-in-github-desktop) if more detail is required.

The bug is caused by sync issues between `state` and `props` that were exacerbated by a switch to using `!shallowEquals()` instead of `!==`, which eliminated calls to `setState()` in cases where the `props` objects were different in reference only: after a commit is made, the internal state gets wiped, and when the repo's state changes, the values from `props` need to get copied back again.

I don't work with React very much, so it's possible there is a better way to accomplish what I've done here.  However, my reading pointed to this approach, so I've opted to open a pull request in the hope of moving this bug forward.

## Release notes

Fixes problem with commit message summary and description being lost after undoing a commit.

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -21,7 +21,6 @@ import { showContextualMenu } from '../main-process-proxy'
 import { Octicon, OcticonSymbol } from '../octicons'
 import { IAuthor } from '../../models/author'
 import { IMenuItem } from '../../lib/menu-item'
-import { shallowEquals } from '../../lib/equality'
 import { ICommitContext } from '../../models/commit'
 
 const addAuthorIcon = new OcticonSymbol(
@@ -122,6 +121,20 @@ export class CommitMessage extends React.Component<
     this.props.dispatcher.setCommitMessage(this.props.repository, this.state)
   }
 
+  public componentWillReceiveProps(nextProps: ICommitMessageProps) {
+    const { commitMessage } = nextProps
+    if (!commitMessage) {
+      return
+    }
+
+    if (commitMessage !== this.props.commitMessage) {
+      this.setState({
+        summary: commitMessage.summary,
+        description: commitMessage.description,
+      })
+    }
+  }
+
   public componentDidUpdate(prevProps: ICommitMessageProps) {
     if (
       this.props.autocompletionProviders !== prevProps.autocompletionProviders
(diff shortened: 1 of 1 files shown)


billygriffin:
Thanks @humphd and we're excited to see your first pull request in Desktop! ❤️ Just to set expectations, it's possible we may get to this in the few days, but we're planning a fairly large release for next week that may consume a good chunk of our time until then, so apologies if we aren't able to get to a thorough review until after that ships. Thank you again! 🎉

humphd:
@outofambit thanks for the review.  Based on your feedback, I've made the two corrections you requested.

One question I had while doing this change was whether or not you want to separate the case of the user changing the `summary` vs. `description` when evaluating if the component commit message is "empty."  That is, I could change the `description` but not the `summary`, or the other way around, or both.

I've opted for the simplest thing, namely, treating the empty `summary` and `description` as a single unit in my check before copying over the incoming `props` values.  Let me know if you think this is wrong.

NOTE: I've also dealt with the startup case, where `description` is set to `''` (empty string),  vs the cleared commit case, where it is `null`.  I'm not sure why the code allows for both, but it's clearly intentional, so my check for an "empty" `description` is more broad than is the case for `summary`.

billygriffin:
Thanks for making those changes and your thoughtful response @humphd! We're releasing 1.6.3 tomorrow and @outofambit is out until mid next week, so you should hear from us then.

humphd:
@billygriffin OK, thanks for the info.  I'm not in a rush, so whenever this happens is fine with me.

PR #6984, opened 2019-03-01
provide a better description when rebase in progress

### LINKED ISSUE #6980: Detached HEAD state midstream during Rebase
As part of testing rebasing, specifically [test case](https://github.com/shiftkey/rebase-test-cases/tree/2-simple-rebase-conflict-base-branch/2-simple-rebase-conflict), I noticed a detached HEAD state during the merge conflict process. 

Although this is a common side-effect of the rebasing process, I am wondering if we should consider any other language or UI treatment to convey that this is not an error?

As rebasing is for the more advanced user, this may not be needed. 🤷‍♂️Thoughts? or feel free to close.

image

shiftkey:
## Overview

**Closes #6980**

## Description

This tweaks the push/pull button to avoid mentioning `detached HEAD` when a rebase is in progress:

screen shot 2019-03-01 at 7 44 12 am

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1838,6 +1838,11 @@ export class App extends React.Component<IAppProps, IAppState> {
     const remoteName = state.remote ? state.remote.name : null
     const progress = state.pushPullFetchProgress
 
+    const { conflictState } = state.changesState
+
+    const rebaseInProgress =
+      conflictState !== null && conflictState.kind === 'rebase'
+
     const tipState = state.branchesState.tip.kind
     const { pullWithRebase } = state.branchesState
 
@@ -1852,6 +1857,7 @@ export class App extends React.Component<IAppProps, IAppState> {
         progress={progress}
         tipState={tipState}
         pullWithRebase={pullWithRebase}
+        rebaseInProgress={rebaseInProgress}
       />
     )
   }
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -44,6 +44,9 @@ interface IPushPullButtonProps {
 
   /** Has the user configured pull.rebase to anything? */
   readonly pullWithRebase?: boolean
+
+  /** Is the detached HEAD state related to a rebase or not? */
+  readonly rebaseInProgress: boolean
 }
 
 function getActionLabel(
@@ -189,7 +192,9 @@ export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
     }
 
     if (tipState === TipState.Detached) {
-      return 'Cannot publish detached HEAD'
+      return this.props.rebaseInProgress
+        ? 'Rebase in progress'
+        : 'Cannot publish detached HEAD'
     }
 
     if (tipState === TipState.Unborn) {


tierninho:
LGTM, however I spotted two issues:

- the `Push` menu item is still active in the Repo menu, which triggers a console error:
`Uncaught (in promise) Error: The current repository is in a detached HEAD state`
image

- Aside from above and while in this state, if you navigate to the Changes tab and uncheck the changes, the `Continue rebase` button is still active. I expect to be be disabled until the user checks the changed file(s).

shiftkey:
> the `Push` menu item is still active in the Repo menu, which triggers a console error:
  `Uncaught (in promise) Error: The current repository is in a detached HEAD state`

This is a good find. I assumed that the menu items were aware of a detached HEAD but I'll see if I can lock that down for `1.6.3`.

> * Aside from above and while in this state, if you navigate to the Changes tab and uncheck the changes, the `Continue rebase` button is still active. I expect to be be disabled until the user checks the changed file(s).

This is an interesting point, and I've got some ideas to tackle this:

 - make the `Continue Rebase` button aware of the included files count
 - make the files which are part of the rebase checked and disable the checkbox, so the user cannot uncheck them while in the middle of a rebase

I favour the latter approach because the app should require the user to bring along all the files when continuing the rebase. @billygriffin any objections to figuring out how to achieve that within the Changes tab?

tierninho:
> Make the files which are part of the rebase checked and disable the checkbox, so the user cannot uncheck them while in the middle of a rebase

I also favor the latter. Note, in a traditional merge conflict the `Commit` button is only active if the conflicted file is checked. If I were to introduce and another file and only check that file, the `Commit` button is deactivated. There is no such logic in place for this in rebasing... yet.

billygriffin:
> make the files which are part of the rebase checked and disable the checkbox, so the user cannot uncheck them while in the middle of a rebase

Makes sense to me, agreed! One question, I remember there was some goofiness in the merge conflicts flow around working directory changes being inadvertently committed as part of the conflict resolution process. Does that same concern exist in this flow or no?

shiftkey:
@billygriffin 

> One question, I remember there was some goofiness in the merge conflicts flow around working directory changes being inadvertently committed as part of the conflict resolution process. Does that same concern exist in this flow or no?

Are you referring to _untracked files_ and this issue https://github.com/desktop/desktop/issues/6411? If so, I have the same behaviour in place when continuing a rebase:

https://github.com/desktop/desktop/blob/9f182b9d2309ba3a34e808191f410cc5a150aeab/app/src/lib/stores/app-store.ts#L3363-L3371

For tracked changes, there's two different flows to consider:

 - starting a rebase - the current behaviour is to either block you with a message, or enable `rebase.autostash` as I talked about in https://github.com/desktop/desktop/issues/3422#issuecomment-468699803



 - continuing a rebase - if you make a change to another file as part of resolving conflicts, this will be included in the commit when continuing the rebase. This might be a bit confusing at first listen, but it's the same behaviour as when you rebase via the Git command line (it won't let you continue without staging all changes).

Happy to chat more about this if there's further questions. I've also opened #7006 to track what we should do with the checkboxes for changed files while in a rebase.

outofambit:
@shiftkey I believe @billygriffin is referring to https://github.com/desktop/desktop/issues/6349 (but correct me if i'm wrong)

billygriffin:
Since #7005 and #7006 are now a thing, I think I'm comfortable with this.

PR #6981, opened 2019-02-28
[rebase] enable `--force-with-lease` after rebasing

shiftkey:
## Overview

**Closes #6913**

## Description

This PR touches a few areas to enable `--force-with-lease` after a rebased is completed in the app:

 - when a rebase is complete, add the updated branch to a list of valid branches that can be force pushed later
 - updated the push/pull button to show when the current branch can be force pushed
 - a new dialog to confirm the force push, controlled by a new preference
 - `push` now has the ability to set `--force-with-lease`
 - do the thing (right up until branch protections block me from doing it)

TODO: 

 - [x] no-op rebase should not enable force push (check that tip has changed from original commit ID)
 - [x] icon for "force push" is a plain Octicon - swap out for custom icon
 - [x] a bunch more testing
 - [x] rebase once #6961 is reviewed and merged
 - [x] cleanup commit history

## Release notes

Notes: `[Improved] users are able to force push branches that were rebased within Desktop`

PR #6986, opened 2019-03-01
ensure push goes to the right remote

shiftkey:
## Overview

This is an annoying bug I uncovered while testing #6981, and I think it was significant enough to extract and review before landing more rebase work.

## The Bug

As part of testing #6981 I would sometimes see the force push complete, and sometimes I'd see it remain in the ahead/behind state despite no errors being reported by the `push`.

I have these two remotes configured, and the branch I am working on is tracking `upstream`:

```
$ git remote -v
origin	https://github.com/shiftkey/desktop (fetch)
origin	https://github.com/shiftkey/desktop (push)
upstream	https://github.com/desktop/desktop.git (fetch)
upstream	https://github.com/desktop/desktop.git (push)
```

Puzzled by it sometimes working, I eventually added this code inside `performPush` as a sanity check:

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index 82afe85b7..4de1895af 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2619,6 +2619,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
       if (tip.kind === TipState.Valid) {
         const { branch } = tip
 
+        if (branch.remote !== null && branch.remote !== remote.name) {
+          debugger
+        }
+
         const pushTitle = `Pushing to ${remote.name}`
 
         // Emit an initial progress even before our push begins
```

And after setting up to force push (note the `upstream` mention here):

I was eventually able to trip this check:

screen shot 2019-03-01 at 1 41 54 pm

What is the button showing at this point?

screen shot 2019-03-01 at 1 42 00 pm

Oh, oh no. I peeked at the state of the local variables and I seem to have stumbled upon a state where the "current" remote is `origin` rather than `upstream`:

screen shot 2019-03-01 at 1 42 47 pm

What does this mean? Well, it means `origin` is passed into as an argument to `git push`

screen shot 2019-03-01 at 1 43 11 pm

And the branch is pushed to the wrong remote 😱 

## The Fix

I've broken this PR down into the three places that need to be fixed here to address this race condition:

 - [x] 026122b  - show the tracking branch remote, if set, before the default remote
 - [x] ba3ee4b - use the tracking branch remote on push, if set
 - ~~[ ] `????????` - nail down race condition that causes `loadRemotes` to not have the right remote~~

https://github.com/desktop/desktop/blob/ba3ee4bba9ca9cbd89d1a899c3f6c38658e76a53/app/src/lib/stores/git-store.ts#L977-L980

## Release notes

Notes: `[Fixed] race condition causes branch to be published to wrong remote`

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2619,7 +2619,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
       if (tip.kind === TipState.Valid) {
         const { branch } = tip
 
-        const pushTitle = `Pushing to ${remote.name}`
+        const remoteName = branch.remote || remote.name
+
+        const pushTitle = `Pushing to ${remoteName}`
 
         // Emit an initial progress even before our push begins
         // since we're doing some work to get remotes up front.
@@ -2627,7 +2629,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
           kind: 'push',
           title: pushTitle,
           value: 0,
-          remote: remote.name,
+          remote: remoteName,
           branch: branch.name,
         })
 
@@ -2656,7 +2658,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
             await pushRepo(
               repository,
               account,
-              remote.name,
+              remoteName,
               branch.name,
               branch.upstreamWithoutRemote,
               progress => {
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1835,7 +1835,7 @@ export class App extends React.Component<IAppProps, IAppState> {
       return <RevertProgress progress={revertProgress} />
     }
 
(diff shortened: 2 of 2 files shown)


shiftkey:
>  - [ ] ???????? - nail down race condition between `loadStatus` and `loadRemotes`

I wasn't able to reproduce this with any sort of consistency in isolation. The commits on this branch should be enough to emulate the desired behaviour (show the remote of the tracking branch if set, otherwise show the default remote, and use this value when pushing) to fix the bug.

shiftkey:
Dropping `ready-for-review` because I want to nail down the race condition that's tripping this and get some closure...

shiftkey:
Haven't been able to trip it at all today - going to put this investigating aside for more rebase tasks and see if I trip it later.

PR #6994, opened 2019-03-02
disable new branch for detached head

tierninho:
if rebase in progress, then `New Branch` menu option should be disabled as well to avoid confusion and error: 

`git -c credential.helper= -c protocol.version=2 checkout --progress cccc --recurse-submodules --` exited with an unexpected code: 1. 2-simple-rebase-conflict/LICENSE.md: needs merge, error: you need to resolve your current index first`

image

related to https://github.com/desktop/desktop/pull/6992 and https://github.com/desktop/desktop/pull/6984

Feel free to close or edit if needed.

PR #7005, opened 2019-03-04
disable create branch menu item when dealing with rebase conflicts

shiftkey:
## Overview

**A more targeted fix for #6994**

## Description

When you are in the middle of a rebase, Desktop should not let the user create a new branch as this is a potential source of confusion.

This wasn't blocked previously because the "detached HEAD" scenario is a way for users to create a new branch based on an arbitrary commit in the repository history.

## Release notes

Notes: no-notes


tierninho:
Tested and LGTM

outofambit:
pulling this into 1.6.3 since its ready and low risk!

Issue #6107, opened 2018-11-06
Improve workflows with uncommitted changes


billygriffin:
We've heard a lot that users want to be able to stash changes in the working directory and that they have to leave GitHub Desktop and do this on the command line. We wanted to take the opportunity to clarify the problem statements as we explore potential paths forward in solving the problems.

More about the history of these problems and how people are experiencing them can be found in #1633 and #2304.

1. I realized after I started doing work that I actually wanted to be doing that work on a different branch. I want to be able to switch branches and retain the uncommitted changes I've made without having to leave GitHub Desktop.

2. I'm doing work on a branch, and something comes up where I need to temporarily go to a different branch to do something entirely different. I'm forced to either WIP commit my work and undo it when I come back or leave Desktop to stash it temporarily. I want to just be able to switch branches, do whatever I needed to do, and come back with my changes right where I left them.

3. When I pull from the remote and there are uncommitted changes, if the uncommitted changes conflict with the commits I'm attempting to pull, I'm not able to pull unless I either commit or stash my changes and deal with the conflicts later. This is true for both regular `git pull` and `git pull --rebase`.  **We've decided that this is not in scope for our initial work.**

Of note, we had a feature on an earlier version of GitHub for Mac that attempted to solve problem 2 with "magic stashing," which auto-stashed your uncommitted changes on a current branch when you switched branches, and auto-popped them when you returned to the original branch.

@shiftkey noted some pitfalls with this approach (capturing here so we consider them in any proposed solutions):

- "Magic stashing didn't restore uncommitted changes after syncing"
- "Magic stashing can fail to restore changes with no indication that they still exist"
- "Magic stashing may delete ignored files"
- "Magic stashing fails if untracked files conflict"
- "Magic stashing information will be lost if a repository moves on disk"

The first two problems are almost certainly related and should not be considered in isolation.

agisilaos:
@billygriffin One of the observations I made using `Magic stashing`, is that it doesn't work all the times. Sometimes it may work, other times not at all. When it doesn't work can't understand why and that's kind of a bummer. Not sure if unreliable is the correct word here (kind of harsh) to describe the situation but it's definitely not a pleasant one. 

Using the terminal for stashing isn't that bad but it adds one more step to my workflow and I kinda "hate" that. Because of situations like the one above I now remember every shortcut in Github Desktop without looking at them so that's cool. 😛

Janpot:
95% of the time I need to shash because I did

1. make a PR1
[...]

PR #7016, opened 2019-03-06
put "local changes overwritten" dialog behind feature flag

shiftkey:
## Overview

This PR puts the work contributed in #6755 behind a feature flag so that it won't go out as part of the 1.6.3 release today.

## Description

Why are we unshipping #6755? 

 - #6755 was opened without a corresponding issue. This might seem minor, but we require an issue to give us an opportunity to discuss problems with workflows in sufficient detail, to ensure we land on the right solution. We should have identified this earlier, and that's on the maintainers to better communicate this requirement and ensure the right discussions happen.
 - The changes in #6755 overlap with our potential plans to support stashing changes within the app. You can see the latest designs for this workflow in https://github.com/desktop/desktop/issues/6107#issuecomment-467197085.

It feels inappropriate for us to ship this workflow change in Desktop if we're going to be changing this workflow significantly in the near term, so this PR will feature-flag the error handler (that shows the new dialog) to avoid it being displayed for `beta` or `production` users while we evaluate the designs for stashing support.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -75,3 +75,12 @@ export function enablePullWithRebase(): boolean {
 export function enableGroupRepositoriesByOwner(): boolean {
   return enableBetaFeatures()
 }
+
+/**
+ * As there is some lack of clarity about this feature with how it overlaps
+ * with planned support for stashing changes, this warning dialog is behind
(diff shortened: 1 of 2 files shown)


shiftkey:
cc @mathieudutour

PR #7018, opened 2019-03-06
only trigger conflicts flow for the selected repository

shiftkey:
## Overview

**Related to #6313**

## Description

See reproduction steps in https://github.com/desktop/desktop/issues/6313#issuecomment-470162214

I focused on this area because @cheshire137 mentioned that it seemed to occur when dealing with merge conflicts, which was how I stumbled upon the fix below.

I thought this might have been affected by #6141 (first released in `1.6.2-beta1`) which introduces the code path: 

> - `AppStore.withPushPullFetch` runs
> - ... which calls `AppStore._refreshRepository`
> - ... which calls `AppStore.loadStatus`
> - ... which calls `AppStore._triggerConflictsFlow`

But the original issue was reported against `1.5.1-beta0` so I'm not sure this is completely resolved.

I also pushed c2e7276a4 as some cleanup because I couldn't find any callers for that API.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1616,7 +1616,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
       conflictState: updateConflictState(state, status, this.statsStore),
     }))
 
-    this._triggerConflictsFlow(repository)
+    if (this.selectedRepository === repository) {
+      this._triggerConflictsFlow(repository)
+    }
 
     this.emitUpdate()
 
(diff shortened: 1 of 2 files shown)


iAmWillShepherd:
I'm unable to reproduce the behavior this PR fixes, but I'm inclined to improve it anyways because of the cleanup and the added check to make sure we're in the right repo before kicking off background tasks.

shiftkey:
> I'm unable to reproduce the behavior this PR fixes, 

I'd like someone to sanity check me before this gets merged.

shiftkey:
For extra context, `AppStore.shouldBackgroundFetch` is rather aggressive in avoiding fetches (it looks at the `last_push` field for the repository on the GitHub API, so an inactive repository is likely to be skipped) and it's `AppStore.performFetch()` that then performs `AppStore._refreshRepository()` which triggers the bug.

PR #7020, opened 2019-03-06
improve error handling inside fetchAll

### LINKED ISSUE #7019: Clone repo list blocked by api 500
When loading my cloned repo list, it sometimes reverts to the "cannot find" state due to a 502 in the api GET. 

Cannot reproduce at will, but it happens frequently enough. Not a blocker as it can be resolved by `cmd-R` in the dev console. Refreshing the list does not resolve the issue.

screen shot 2019-03-06 at 7 37 08 am

 `http.ts:135 GET https://api.github.com/user/repos?per_page=100&page=19 502 (Bad Gateway)`
`fetchRepositories: Error: Server Error`

v1.6.3, mac

shiftkey:
## Overview

**Closes #7019**

## Description

This changes how `fetchAll` works in three ways:

 - makes it aware of `5xx` HTTP error codes, logs a helpful message and returns
 - handles all `4xx` HTTP error codes, not just `404`
 - if an error is encountered, return the current pages of deserialized results (not an empty array)

The original error was occurring when fetching page `19` out of an unknown amount, and for the sake of showing _something_ I think this API should return whatever it has currently retrieved before the error occurred.

## Release notes

Notes: `[Improved]

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -484,6 +484,15 @@ export class API {
         return []
       }
 
+      if (response.status >= 500 && response.status < 599) {
+        log.warn(
+          `fetchAll: '${path}' returned a server error code '${
+            response.status
+          }'`
+        )
+        return []
+      }
+
       const items = await parsedResponse<ReadonlyArray<T>>(response)
       if (items) {
         buf.push(...items)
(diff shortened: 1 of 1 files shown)


shiftkey:
Closing this out in favour of #7259

Issue #6982, opened 2019-02-28
Desktop does not accurately reflect the latest pull requests from GitHub


billygriffin:
**Please describe the problem you think should be solved**

Problem 1: The list of pull requests in GitHub Desktop does not quickly reflect the latest information about what pull requests exist from github.com. In broader terms, someone has proposed changes to the code base and I don't have any knowledge of it because GitHub Desktop doesn't reflect it until later or until I manually click fetch with the pull requests tab open.

Problem 2: When it does fetch the latest pull requests, the list is incredibly slow to load and therefore its utility is limited. This was brought up in #5448. In broader terms, I want to be able to see what work has been proposed to merge into the code base, and the list of things takes forever to load.

**Example of problem 1:**

Before fetch:

<img width="524" alt="screen shot 2019-02-28 at 4 11 30 pm" src="https://user-images.githubusercontent.com/5091167/53605081-a52d5480-3b73-11e9-856d-49f04117b776.png">

After fetch:

<img width="520" alt="screen shot 2019-02-28 at 4 12 10 pm" src="https://user-images.githubusercontent.com/5091167/53605086-ab233580-3b73-11e9-9836-ad2d0cb24085.png">

As you can see, I wouldn't have known the newest pull request (#6981) existed until I explicitly clicked "Fetch" with the pull requests tab open and then waited what felt like a long time for the list to load in (it was around 10 seconds total to load 41 PRs).

Example of problem 2:

<img width="524" alt="screen shot 2019-02-28 at 4 19 57 pm" src="https://user-images.githubusercontent.com/5091167/53605474-d22e3700-3b74-11e9-84d6-8df69b4a44a7.png">

cc: @niik

billygriffin:
@niik An interesting point was brought up by @tierninho in https://github.com/desktop/desktop/issues/3628, where he pointed out that if you rename a branch that has a PR in Desktop, you can get into some interesting situations. I'm not sure that's a thing that happens enough to make it worth solving, but wanted to at least make that decision consciously as part of this effort.

PR #7027, opened 2019-03-07
Explicit refresh of PR list and avoid skeleton loading screen when we have PRs to show

niik:
## Overview

References: #6982 

## Description

This is the first in a series of PRs to address how the app retrieves, stores, and presents pull requests. This change adds an explicit refresh button for the PR list in a similar manner to what we have for repositories in the clone dialog. Prior to this change the only way to explicitly refresh the list of pull requests was by clicking on the fetch button in the toolbar which was undiscoverable and to make matters worse the fetch button is unavailable when on an unpublished branch making it impossible to refresh the list on-demand.

This PR also alters the behavior of the PR list such that we no longer replace the list with a skeleton loading view whenever the PR list is loading. The skeleton loading view made it impossible to check out a PR while we were refreshing. Instead we now only show that view when we're doing the initial load (or when the previous load didn't return any PRs) or in other words we only show it when we don't already have any PRs to show in the list.

Here's a before and after comparison.

### Before

pre-loading

### After

post-loading

The keen observer will note that it takes a very long time to refresh the list of PRs. This is due to the fact that we're fetching the latest PR status (i.e. CI status) for each PR before we return the new list. This is something I'm planning on decoupling in another PR very soon. 

## Release notes

Notes:

### DIFF
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -176,7 +176,12 @@ export class BranchesContainer extends React.Component<
 
   private renderPullRequests() {
     if (this.props.isLoadingPullRequests) {
-      return <PullRequestsLoading key="prs-loading" />
+      return (
+        <PullRequestsLoading
+          key="prs-loading"
+          renderPostFilter={this.renderPullRequestPostFilter}
+        />
+      )
     }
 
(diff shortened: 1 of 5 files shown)


tierninho:
Tested and LGTM.

PR #7029, opened 2019-03-07
feat: Add visual feedback on Remove and Discard changes dialogs

### LINKED ISSUE #7015: confirmation dialog shows no visual feedback when performing an action

When discarding all changes for many files, clicking on discarding all changes will begin performing the action, but there is no visual indication it's actually doing anything until the operation is complete. The button simply doesn't appear to do anything, so it's likely the user will press cancel. In the background though, the operation is still working away...

image

Not specific to dark mode:
image

This is also the case when removing a large repository and moving it to the recycle bin.

image

HashimotoYT:
## Overview

**Closes #7015**

## Description

- Added spinners and disables button when "Remove Repository" and "Discard changes" dialogs have a action.

Discarding: 
discarding

Removing:
removing repo

## Release notes

Notes: `[Improved?] "Remove" and "Discard changes" dialogs show visual effects on action`

### DIFF
--- a/app/src/ui/discard-changes/discard-changes-dialog.tsx
+++ b/app/src/ui/discard-changes/discard-changes-dialog.tsx
@@ -84,7 +84,9 @@ export class DiscardChanges extends React.Component<
 
         <DialogFooter>
           <ButtonGroup destructive={true}>
-            <Button type="submit">Cancel</Button>
+            <Button disabled={this.state.isDiscardingChanges} type="submit">
+              Cancel
+            </Button>
             <Button
               onClick={this.discard}
               disabled={this.state.isDiscardingChanges}
--- a/app/src/ui/discard-changes/discard-changes-dialog.tsx
+++ b/app/src/ui/discard-changes/discard-changes-dialog.tsx
@@ -60,6 +60,7 @@ export class DiscardChanges extends React.Component<
 
   public render() {
     const discardingAllChanges = this.props.discardingAllChanges
+    const isDiscardingChanges = this.state.isDiscardingChanges
 
     return (
       <Dialog
@@ -70,7 +71,8 @@ export class DiscardChanges extends React.Component<
             : toPlatformCase('Confirm Discard Changes')
         }
         onDismissed={this.props.onDismissed}
-        loading={this.state.isDiscardingChanges}
+        dismissable={isDiscardingChanges ? false : true}
+        loading={isDiscardingChanges}
         type="warning"
       >
         <DialogContent>
@@ -84,13 +86,10 @@ export class DiscardChanges extends React.Component<
 
         <DialogFooter>
           <ButtonGroup destructive={true}>
-            <Button disabled={this.state.isDiscardingChanges} type="submit">
+            <Button disabled={isDiscardingChanges} type="submit">
               Cancel
             </Button>
-            <Button
-              onClick={this.discard}
-              disabled={this.state.isDiscardingChanges}
-            >
+            <Button onClick={this.discard} disabled={isDiscardingChanges}>
               {discardingAllChanges
                 ? toPlatformCase('Discard All Changes')
                 : toPlatformCase('Discard Changes')}
--- a/app/src/ui/discard-changes/discard-changes-dialog.tsx
+++ b/app/src/ui/discard-changes/discard-changes-dialog.tsx
@@ -70,6 +70,7 @@ export class DiscardChanges extends React.Component<
             : toPlatformCase('Confirm Discard Changes')
         }
         onDismissed={this.props.onDismissed}
+        loading={this.state.isDiscardingChanges}
         type="warning"
       >
         <DialogContent>
@@ -84,7 +85,10 @@ export class DiscardChanges extends React.Component<
         <DialogFooter>
           <ButtonGroup destructive={true}>
             <Button type="submit">Cancel</Button>
-            <Button onClick={this.discard}>
+            <Button
+              onClick={this.discard}
+              disabled={this.state.isDiscardingChanges}
+            >
               {discardingAllChanges
                 ? toPlatformCase('Discard All Changes')
                 : toPlatformCase('Discard Changes')}
--- a/app/src/ui/remove-repository/confirm-remove-repository.tsx
+++ b/app/src/ui/remove-repository/confirm-remove-repository.tsx
@@ -55,13 +55,16 @@ export class ConfirmRemoveRepository extends React.Component<
   }
 
   public render() {
+    const isRemovingRepository = this.state.isRemovingRepository
+
     return (
       <Dialog
         id="confirm-remove-repository"
         key="remove-repository-confirmation"
         type="warning"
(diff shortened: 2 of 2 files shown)


billygriffin:
@HashimotoYT Thanks for the PR! Is this ready for review?

billygriffin:
@HashimotoYT Also, quick question on the discard changes: does this also trigger the spinner when you're just discarding a selection of changes. In other words, if I have 10 changes and I've selected 5 of them to discard, does the spinner show up?

HashimotoYT:
@billygriffin Spinner shows anyways displayed, discarding selected files shows same Dialog component, and yes this is ready for review 🍿.

shiftkey:
@HashimotoYT another example to test out as part of this would be discarding thousands of files (renaming `node_modules` to something that isn't covered by `.gitignore` would be a good way to get that). How does discarding that many files look within the app?

tierninho:
Tested and LGTM 👍

HashimotoYT:
It works as planned (tested on `e/electronjs.org` node_modules, 53k), spinner spinning, discard button doesn't work, Cancel works as well. But on canceling, app starts hanging. GIF:

tierninho:
@HashimotoYT Thanks for surfacing this. The `Cancel` button should be disabled if the user clicks the `Discard` button.

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -84,7 +85,10 @@ export class DiscardChanges extends React.Component<
         <DialogFooter>
           <ButtonGroup destructive={true}>
             <Button type="submit">Cancel</Button>
```
iAmWillShepherd: I think we need to add the following attribute to the _Cancel_ button: `disabled={this.state.isDiscardingChanges}`.

app/src/ui/remove-repository/confirm-remove-repository.tsx
```diff
@@ -87,7 +92,12 @@ export class ConfirmRemoveRepository extends React.Component<
         <DialogFooter>
           <ButtonGroup destructive={true}>
             <Button type="submit">Cancel</Button>
```
iAmWillShepherd: Same thing for this _Cancel_ button.

HashimotoYT:
> The `Cancel` button should be disabled if the user clicks the `Discard` button.

@tierninho  Hm... 🤔 Disabling `Cancel` button does not make dialog impossible to exit, the user can use `x` mark or click somewhere on free space behind dialog for close. Can please give some cause why dialog should be not dismissable manually?

njbmartin:
So this is where it gets a bit complicated. Cancelling the dialog doesn't actually cancel the action (yet?), but as a user I would expect it to.

HashimotoYT:
@iAmWillShepherd changes made

iAmWillShepherd:
You are right, cancelling the dialog won't cancel the action.  Cancelling an in-progress discard is a non-trivial thing to do with how the discard logic works in the application right now and we’d prefer to focus on getting this change in which will at least provide some feedback while the discard is running and  since we’re not going to be able to cancel the operation, we feel like it would be confusing if the cancel button remains enabled after discarding has begun. That's why our opinion is that we should disable the _Cancel_ button to ensure it's clear what's happening. 

If we did cancel an operation while it's running, doing so would put your repository in an unclear state depending on what percentage of the operation was completed. In this case, especially since you're already in a confirm dialog, our assumption is that once a user has clicked confirm to begin the operation, that's it.

The _Cancel_ button here is really about whether you want to start the operation in the first place. Note that the _Cancel_ button is enabled and displayed before confirming, but sticks around after confirmation is made. I can see how that button remaining visible after confirming changes the semantics behind it.

The dialog component actually has a prop to make it not-dismissible:
https://github.com/desktop/desktop/blob/253a27b20116494b0a8dd3e7d11a59c5b8095972/app/src/ui/dialog/dialog.tsx#L49

And a prop to disable all buttons within it:
https://github.com/desktop/desktop/blob/253a27b20116494b0a8dd3e7d11a59c5b8095972/app/src/ui/dialog/dialog.tsx#L103

Using ☝️ means we don't have to modify that ourselves, instead relying on the component to do it.

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -84,7 +84,9 @@ export class DiscardChanges extends React.Component<
 
         <DialogFooter>
           <ButtonGroup destructive={true}>
-            <Button type="submit">Cancel</Button>
+            <Button disabled={this.state.isDiscardingChanges} type="submit">
```
iAmWillShepherd: It looks like you made this change before I got around to explaining how we can do this using the built-in functionality of the `Dialog` component. See [comment].(https://github.com/desktop/desktop/pull/7029#issuecomment-470996180)

I can made the change for you if you like, just let me know 👍

HashimotoYT: Looks this should be how this? When the process started it's making dialog not dismissable. When process ended dialog closes automatically.

```diff
diff --git a/app/src/ui/discard-changes/discard-changes-dialog.tsx b/app/src/ui/discard-changes/discard-changes-dialog.tsx
index 32a901f8e..0a8fba43a 100644
--- a/app/src/ui/discard-changes/discard-changes-dialog.tsx
+++ b/app/src/ui/discard-changes/discard-changes-dialog.tsx
@@ -60,6 +60,7 @@ export class DiscardChanges extends React.Component<
 
   public render() {
     const discardingAllChanges = this.props.discardingAllChanges
+    const isDiscardingChanges = this.state.isDiscardingChanges
 
     return (
       <Dialog
@@ -70,7 +71,8 @@ export class DiscardChanges extends React.Component<
             : toPlatformCase('Confirm Discard Changes')
         }
         onDismissed={this.props.onDismissed}
-        loading={this.state.isDiscardingChanges}
+        dismissable={isDiscardingChanges ? false : true}
+        loading={isDiscardingChanges}
         type="warning"
       >
         <DialogContent>
@@ -84,13 +86,10 @@ export class DiscardChanges extends React.Component<
 
         <DialogFooter>
           <ButtonGroup destructive={true}>
-            <Button disabled={this.state.isDiscardingChanges} type="submit">
+            <Button disabled={isDiscardingChanges} type="submit">
               Cancel
             </Button>
-            <Button
-              onClick={this.discard}
-              disabled={this.state.isDiscardingChanges}
-            >
+            <Button onClick={this.discard} disabled={isDiscardingChanges}>
               {discardingAllChanges
                 ? toPlatformCase('Discard All Changes')
                 : toPlatformCase('Discard Changes')}
```

iAmWillShepherd: @billygriffin how do you feel about the interaction explained above? Basically, once you start discarding, you can't back out of it (dialog is not dismissable and all buttons are disabled).

iAmWillShepherd: @HashimotoYT, yes, that looks good!

PR #7030, opened 2019-03-07
use shared components for merge and rebase conflicts

### LINKED ISSUE #6904: investigate unifying rebase conflicts file rendering with merge conflicts file rendering
Rebase conflicts currently uses a `app/src/ui/rebase/conflict-files-list.tsx` component to render it's items. This should be combined with the work done around resolving manual conflicts in #6777 and #6821 which is now contained in `app/src/ui/merge-conflicts/unmerged-file.tsx`.

There are some differences that need to be thought about:

 - `unmerged-file.tsx` takes in `ourBranch` and `theirBranch` as props - we need to think about what makes sense to show for these options when handling conflicts
 - probably some other things

shiftkey:
## Overview

**Closes #6904**

## Description

This PR addresses some cleanup post `1.6.3` that brings together the merge conflicts and rebase conflicts components, given how much overlap they have:

- introduced `app/src/ui/lib/conflicts/` module to contain shared functionality
- moved small render functions as well as `renderUnmergedFile` into this directory
- gutted most of `RebaseConflictsDialog` in favour of using shared components
- tidied up some usages inside `MergeConflictsDialog` to use the render functions

Testing:

 - [x] trigger merge conflicts, confirm visuals and behaviour is :gem:
 - [x] trigger manual merge conflicts, confirm visuals and behaviour is :gem:
 - [x] trigger rebase conflicts, confirm visuals and behaviour is :gem:
 - [x] trigger manual rebase conflicts, confirm visuals and behaviour is :gem:

## Release notes

Notes:


shiftkey:
I added ec327ff3184947994f871398cc5e7ae2397b6c35 to this branch because rebase terminology uses the opposite names for "ours" and "theirs" while doing a rebase. To illustrate before the change:



With that last commit, the change is correctly associated:



You'll also note that sometimes we lack the context for listing both branches when dealing with rebase conflicts - once Git kicks off the rebase it uses the base commit ID rather than ref, so in this case (with an in-progress rebase) we are unable to resolve the base branch to show in this context menu. Hopefully once we can get the full rebase flow baked we'll be able to track the branch name when the rebase started and flow that through to the other dialogs.

app/src/ui/lib/conflicts/render-functions.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <div className="message">All conflicts resolved</div>
+    </div>
+  )
+}
+
+export function renderShellLink(
+  openThisRepositoryInShell: () => void
+): JSX.Element {
+  return (
+    <div>
```
iAmWillShepherd: Can we return a simple fragment (`<>all the stuff</>`) here since this `div` node doesn't have an Id or class.

app/src/ui/lib/conflicts/render-functions.tsx
```diff
@@ -0,0 +1,36 @@
...
+export function renderUnmergedFilesSummary(conflictedFilesCount: number) {
+  // localization, it burns :vampire:
+  const message =
+    conflictedFilesCount === 1
+      ? `1 conflicted file`
+      : `${conflictedFilesCount} conflicted files`
+  return <h3 className="summary">{message}</h3>
+}
+
+export function renderAllResolved() {
```
iAmWillShepherd: **Non-blocking**
I noticed the block below has `JSX.Element` in its signature, while the two above don't.

app/src/ui/lib/conflicts/unmerged-file.tsx
```diff
@@ -47,7 +47,7 @@ export const renderUnmergedFile: React.SFC<{
    */
   readonly manualResolution?: ManualConflictResolution
   /** branch we were on before starting merge */
-  readonly ourBranch: string
+  readonly ourBranch?: string
```
iAmWillShepherd: 📝 A bit more context on this prop would be helpful. What does passing `undefined`  mean (i.e., why would we want to pass `undefined` to this component instead of the branch name)?

app/src/ui/lib/conflicts/render-functions.tsx
shiftkey: I want to leave this as-is because this is how it was written for use in Merge Conflicts, and cc @outofambit for feels when she returns next week.

app/src/ui/lib/conflicts/render-functions.tsx
shiftkey: Removed in f6c3c9705 as the type inference is Good Enough™

app/src/ui/lib/conflicts/unmerged-file.tsx
shiftkey: Added in caa88e62e8351702e520bd22c71c88b6c30fd6c7 - let me know if there's anything else I can clarify while I'm in here

tierninho:
>  trigger merge conflicts, confirm visuals and behaviour is 💎
  trigger manual merge conflicts, confirm visuals and behaviour is 💎
  trigger rebase conflicts, confirm visuals and behaviour is 💎
  trigger manual rebase conflicts, confirm visuals and behaviour is 💎

Tested the above scenarios and results were 💎as well.

iAmWillShepherd:
Merge early since the last changes were update to type signature and improved comments.

Issue #6987, opened 2019-03-01
Stashing [Tracking Issue]


iAmWillShepherd:
Original issue: #6107

<details><summary><strong>Design notes</strong></summary>

### git-stash list

```sh
> git stash list
stash@{0}: WIP on stashing: 41f7b73 some random commit message
stash@{1}: WIP on stashing: 41f7b73 some random commit message
stash@{2}: On stashing: a message
stash@{3}: On stashing: this is a message
stash@{4}: WIP on stashing: 531bd83 1
stash@{5}: WIP on stashing: 531bd83 1
```

I'll be using `git log -g refs/stash` to emulate the behaviour of `git stash list` since it uses `git log` under the hood ([docs](https://git-scm.com/docs/git-stash#git-stash-listltoptionsgt)). Doing it this way will help us avoid doing extra work which is super important on Windows as process creation is very expensive.

[Git implementation](https://github.com/git/git/blob/ecbdaf0899161c067986e9d9d564586d4b045d62/git-stash.sh#L402-L405)

---

### git-stash push

```sh
> git stash push -m “a message”
Saved working directory and index state On stashing: a message
```

I’ll be using a specially formatted message to differentiate between stash objects created by the user and Desktop on the user’s behalf.
This is not a bullet-proof solution because
* a user can use our message format when they  stash something from the CLI and
* they can edit the message stored in a stash object by modifying _git/logs/refs/stash_

However, we take a similar approach for adding remotes and checking out pull requests, so I think we can safely ignore the risks above.

**Message Format**

`github-desktop: fully qualified branch name`

| Component      | Reasoning                                                                                                                                                                             |
| -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| github-desktop | It’s unlikely anyone would use that in their own stash message and it makes is super clear which entries are created by Desktop.                                                     |
| tip            | Having the tip of the branch at the time the stash object is created should prevent issues related to ambiguous branch names. An added benefit is we could deal with branch renaming. |

My thinking is that I can use the tip to make extra sure the stash entry is based on a specific branch. I am making an assumption that if the commit SHA is a branch’s history, then it likely is that branch. I think this means I can get by without using fully-qualified branch names.

### git-stash apply

**Happy path**

```sh
> git stash apply stash{1}
On branch stashing
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

[...]

PR #7036, opened 2019-03-08
Add function to apply an arbitrary stash entry

iAmWillShepherd:
See tracking issue - #6987

Desktop has to handle the stash entries list being changed externally, so we can't depend on 
* `git stash pop` - it's not guaranteed our entry will be at the top of the stack
*  the entry's name (e.g. `sha@{0}`) - the stash entry they point to can change
Using the the commit hash of the stash entry appears to be the best way to **uniquely id** an entry.

This PR adds a function that can apply an arbitrary stash entry using its commit hash.

### DIFF
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -1,3 +1,6 @@
+import { git } from './core'
+import { Repository } from '../../models/repository'
+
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
Closing in favor of #7182.

PR #7042, opened 2019-03-08
test and refactor current push/pull button behaviour

### LINKED ISSUE #6985: investigate refactoring push/pull button internals
I've had to touch this a few times recently to add new inputs and render different details as part of working on rebase workflows:

  - #6984
  - #6893
  - #6981

I'd like to look at reorganzing the internals so that the logic is a bit easier to follow for future changes.

My gut feeling is that we're duplicating the same checks around the place, and future changes have the potential to put the button into an unclear state.

For example, we check whether `this.props.remoteName` is undefined in three places:

https://github.com/desktop/desktop/blob/40a7abaf9094bdb88c435662ea6a9b8705e4c1d4/app/src/ui/toolbar/push-pull-button.tsx#L146-L149

https://github.com/desktop/desktop/blob/40a7abaf9094bdb88c435662ea6a9b8705e4c1d4/app/src/ui/toolbar/push-pull-button.tsx#L161-L164

https://github.com/desktop/desktop/blob/40a7abaf9094bdb88c435662ea6a9b8705e4c1d4/app/src/ui/toolbar/push-pull-button.tsx#L186-L189

My preferred approach at the moment would be to introduce render functions for each expected state that the button can be in:

```tsx
export class PushPullButton extends React.Component<IPushPullButtonProps, {}> {
  public render() {
    const { progress, remoteName } = this.props
   
    if (progress) {
       return renderProgressButton(progress)
    }

   if (!remoteName) {
      return renderPublishRepositoryButton()
   }

   // other states
  }
}
```

I've also been looking for an opportunity to evaluate automated component testing in anger, and ensuring I render the same elements for the same props feels like a compelling reason to wrap this component in some tests before making changes to `render()`.

shiftkey:
## Overview

**Closes #6985**

## Description

- [x] add component testing to `yarn test` step
- [x] setup test suite for `PushPullButton` to cover majority of output cases
- [x] rewrite `PushPullButton.render` to be more maintainable
- [x] rebase changes on top of #6981 once merged
- [x] add more tests to cover force push scenario
- [x] extract component tests out of this branch
- [x] write up some notes about the experience of doing this component testing

## Notes

Diff details aside, the size of the component in the production bundle increased 18% (from 3048 characters to 3615 characters, so still less than 4KB) for the component, which is mostly because we no longer have a central `React.createElement` inside `render()`.

You can see the human-readable before-vs-after (with whitespace and indenting) output in this gist: https://gist.github.com/shiftkey/a3b4ddb7c7e3537e28127279cd5969c7

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -1,14 +1,18 @@
 import * as React from 'react'
-import { ToolbarButton, ToolbarButtonStyle } from './button'
+
 import { Progress } from '../../models/progress'
-import { Dispatcher } from '../dispatcher'
-import { Octicon, OcticonSymbol } from '../octicons'
 import { Repository } from '../../models/repository'
 import { IAheadBehind } from '../../models/branch'
 import { TipState } from '../../models/tip'
-import { RelativeTime } from '../relative-time'
 import { FetchType } from '../../models/fetch'
+
 import { enablePullWithRebase } from '../../lib/feature-flag'
+
+import { Dispatcher } from '../dispatcher'
+import { Octicon, OcticonSymbol } from '../octicons'
+import { RelativeTime } from '../relative-time'
+
+import { ToolbarButton, ToolbarButtonStyle } from './button'
 
 interface IPushPullButtonProps {
   /**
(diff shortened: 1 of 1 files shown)


app/src/ui/lib/props-type.ts
```diff
@@ -0,0 +1,6 @@
+// based on https://codewithstyle.info/TypeScript-conditional-types-real-life-example/
+export type PropsType<C> = C extends React.Component<infer P>
```
shiftkey: This little helper function means we can avoid `export`-ing interfaces on props.

In the test below I was able to type-check the props I was setting up using this:

```tsx
let defaultProps: PropsType<PushPullButton>
```

shiftkey:
I've archived the tests I used in here over in [this branch](https://github.com/desktop/desktop/compare/shiftkey/refactor/component-tests-demo) as a reference to the notes below about component testing.

### What did I do?

This PR was a chance to explore testing React components while I refactor a component, because this scenario felt like a good fit for it, rather than other scenarios that I've been contemplating.

### Why this area?

This component felt like a good fit for testing for a few reasons:

 - it's a component that has been around basically forever, accruing little bits of technical debt over time
 - it's purpose have changed over time, in particular to support new states and actions
 - while it currently has a lot of props to show, the rendered DOM is relatively stable

### What worked?

It was helpful to use these tests as a regression suite while I was refactoring. What I did was build up a number of cases that I knew the component needed to support, and run these tests to store a snapshot of the rendered DOM:

https://github.com/desktop/desktop/blob/5b5fe30fd21d727a971e00bed18026c1228091ec/app/test/component/push-pull-button-test.tsx#L139-L149

This was a bit exploratory, and required looking over the internals to understand what it was looking for, because of the large number of props this component supports. I think unifying this logic outside of the function, and returning a union type of the possible states, would be the next iteration on the API here.

Once I had this set of tests, I could contemplate changes to the code. When I was originally refactoring this component I went through this sort of repetitive process:

 1. identify a scenario inside `render()` that could be extracted out to it's own function
 2. make a change
 3. run the tests to verify the tests aren't broken
 4. commit the change
 5. go back to step 1

Chipping away at the code eventually got me to what is currently in this PR - having one place where the conditions are checked and each scenario becomes a standalone function.

### What didn't work?

I wouldn't recommend this approach for other situations:

 - tests while writing new components - as things change, even during development, I can see a lot of time wasted on updating snapshots in tests
 - tests wrapping large/complex components - because of how Jest snapshots represent the underlying DOM, I can see situations where changing a core component like `Button` would cause cascading test failures in areas that contributors might not understand the impact (or the resolution)
  - tests with interactions to see how the component re-render - I've been able to do this in other situations, but the value of these tests isn't clear to me

I also had a shot at enabling code coverage for these "component tests" but it clashed with the unit test coverage and made CodeCov submit disappointed comments into this draft PR about my coverage failing. If we want to go down this path, we'd need to better organize the unit tests to appease Codecov.

### Would I do it again?

Yes. The other case I've been exploring is how we can verify React components that currently use APIs which will be removed in the next major version of React, to ensure we can effectively migrate to the new APIs without breaking functionality.

Issue #6638, opened 2019-01-15
Small tweaks to onboarding language


tierninho:
For the new onboarding, I am suggesting a few small edits to some the existing language, as such:

- Swap "Configure which editor you wish to use in preferences" ... to ... "Select your editor in Preferences" as there is really nothing to configure and Preferences should be capitalized
- "View the files in your repository in Finder"... perhaps to "View the files of your repository in Finder" or "View your repository's files in Finder"? 🤔 
👇 
<img width="684" alt="screen shot 2019-01-15 at 2 58 35 pm" src="https://user-images.githubusercontent.com/14828183/51215622-59ca2a00-18d6-11e9-92e0-8791e7e4b2a4.png">

- Do not capitalize "your" in the two boxes.
👇 
<img width="676" alt="screen shot 2019-01-15 at 2 58 45 pm" src="https://user-images.githubusercontent.com/14828183/51215623-59ca2a00-18d6-11e9-9b97-d65aa88e77f9.png">

Feel free to disagree or make edits.

shiftkey:
I'm into this 👍 

One thing to add:

> "View the files in your repository in Finder"... perhaps to "View the files of your repository in Finder" or "View your repository's files in Finder"? 

I prefer "View the files of your repository in Finder"  over the other one

shiftkey:
Moving this to `help-wanted` as I think this is something ideal for an external contributor to help out with. The relevant areas to update:

https://github.com/desktop/desktop/blob/96eb4a8894f048cb997d86ae413f3d0bfc33be38/app/src/ui/changes/no-changes.tsx#L281-L287

https://github.com/desktop/desktop/blob/96eb4a8894f048cb997d86ae413f3d0bfc33be38/app/src/ui/changes/no-changes.tsx#L236-L239

msztech:
I can work on this @shiftkey 😃

shiftkey:
Moving this back to "Available" because we haven't heard anything from the interested contributor.

jamesgeorge007:
@shiftkey I shall take this up :clap:

jamesgeorge007:
@shiftkey Kindly have a look at #7152

shiftkey:
@jamesgeorge007 thanks for the quick turnaround. I'm actually going to ask someone else to review it, because I'm feeling very spread thin right now across a bunch of areas...

PR #7072, opened 2019-03-12
Add newline towards end of License file

### LINKED ISSUE #6999: The LICENSE file is created witout newline at end of file when creating a new repository
## Description
When creating a repository with a license the LICENSE file is created without a newline a the end of file. This is indicated by the client when the LICENSE file is opened from the initial commit. It show a red sign with a tool tip containing the text "No newline at end of file".

The way in which the file is created differs from when creating the same repository from the GitHub website as it adds a newline at the end of the LICENSE file.

The creation of new repository, including the LICENSE file, should be processed identically by the GitHub website and GitHub Desktop client. As the GitHub Desktop client indicates with a warning symbol that the newline misses the correct way of processing is implemented within the GitHub website and therefore the GitHub Desktop client should adopt to that.

## Version
* GitHub Desktop:
1.6.2
* Operating system:
Microsoft Windows [Version 10.0.17134.590]

## Steps to Reproduce
1. Select "File" > "New repository" from the top menu
2. Set a value in the text field "Name"
3. Select any value from the drop down "License"
4. Click on the button "Create repository"
5. Click on the tab panel "History" of the new created repository
6. Select the file "LICENSE"

### Expected Behavior
The LICENSE file contains a newline at the end of the file.

### Actual Behavior
The LICENSE file does not contain a newline at the end of the file.

## Additional Information

> Screenshot of GitHub Desktop, after creating the repository including a LICENSE file, through the GitHub Desktop client. 

github-dekstop-history-license

> Screenshot of GitHub Desktop, after creating the repository including a LICENSE file, through the GitHub website. 

github-dekstop-history-license-2

jamesgeorge007:
Closes #6999 
Adds newline towards the end of license file being created.

### DIFF
--- a/app/src/ui/add-repository/licenses.ts
+++ b/app/src/ui/add-repository/licenses.ts
@@ -88,6 +88,7 @@ export async function writeLicense(
     'project',
     'description',
     'year',
+    '\n'
   ]
 
   const body = replaceTokens(license.body, tokens, fields)
--- a/app/src/ui/add-repository/licenses.ts
+++ b/app/src/ui/add-repository/licenses.ts
@@ -75,11 +75,7 @@ function replaceTokens(
 }
 
 function ensureLineEndingSet(body: string) {
-  if (body.endsWith('\n')) {
-    return body
-  } else {
-    return `${body}\n`
-  }
+  return body.endsWith('\n') ? body : `${body}\n`
 }
 
 /** Write the license to the the repository at the given path. */
@@ -95,7 +91,7 @@ export async function writeLicense(
     'email',
     'project',
     'description',
-    'year'
+    'year',
   ]
 
   const body = replaceTokens(license.body, tokens, fields)
--- a/app/src/ui/add-repository/licenses.ts
+++ b/app/src/ui/add-repository/licenses.ts
@@ -74,6 +74,14 @@ function replaceTokens(
   return newBody
 }
 
+function ensureLineEndingSet(body: string) {
+  if (body.endsWith('\n')) {
+    return body
+  } else {
+    return `${body}\n`
+  }
+}
+
 /** Write the license to the the repository at the given path. */
 export async function writeLicense(
   repositoryPath: string,
@@ -87,10 +95,9 @@ export async function writeLicense(
     'email',
     'project',
     'description',
-    'year',
-    '\n'
+    'year'
   ]
 
   const body = replaceTokens(license.body, tokens, fields)
-  await writeFile(fullPath, body)
+  await writeFile(fullPath, ensureLineEndingSet(body))
 }


jamesgeorge007:
@shiftkey @iAmWillShepherd What do you think?

app/src/ui/add-repository/licenses.ts
```diff
@@ -88,6 +88,7 @@ export async function writeLicense(
     'project',
     'description',
     'year',
+    '\n'
```
shiftkey: I'm not entirely sure this is the right fix, given how this array of values is used in `replaceTokens`:

 - these license templates expect a `[token]` that they replace with a value from `ILicenseFields`
 - these tokens are constrained to the expected fields from `ILicenseFields`, and the compiler should catch this:



 - I think what we need to do is verify the generated license text is ending in `\n`, and if not append a character. This could just be a simple function that checks the string and either returns it or modifies it:

```ts
export function ensureLineEndingSet(body: string) {
  if (body.endsWith('\n')) {
    return body
  } else {
    return `${body}\n`
  }
}
```

 - this could be then used when writing the updated license text which is stored in the `body` variable:

```ts
await writeFile(fullPath, ensureLineEndingSet(body))
```

This also reminded me of how we touch `.gitignore` files, that might end in `\r\n` or `\n` depending on your setup. We have some code lying around to do this correctly, and likely some tests we can use, but for now let's just focus on getting `\n` appended to the generated license.

jamesgeorge007: @shiftkey I was quite unaware of the context here :confused:  Thank you for throwing light into the present scenario :+1:

jamesgeorge007:
@shiftkey Made the changes as you suggested :+1 :

jamesgeorge007:
@shiftkey Made the changes as you suggested :+1:

app/src/ui/add-repository/licenses.ts
```diff
@@ -87,9 +95,9 @@ export async function writeLicense(
     'email',
     'project',
     'description',
-    'year',
+    'year'
```
shiftkey: I think our prettier config expects this trailing `,` and will error otherwise

jamesgeorge007: :+1:

jamesgeorge007:
@shiftkey Hope it's fine now :+1:

app/src/ui/add-repository/licenses.ts
```diff
@@ -74,6 +74,14 @@ function replaceTokens(
   return newBody
 }
 
+function ensureLineEndingSet(body: string) {
+  if (body.endsWith('\n')) {
```
jamesgeorge007: Using conditional `?` operator instead of this if conditional check makes it appear more clean. Just thoughts :thinking:

jamesgeorge007: Using conditional `?` operator instead of this if conditional check makes it appear clean. Just thoughts :thinking:

jamesgeorge007: Using conditional `?` operator instead of this if conditional check makes it clean. Just thoughts :thinking:

jamesgeorge007: Using ternary `?` operator instead of the if conditional check makes it clean. Just thoughts :thinking:

jamesgeorge007: :+1:

jamesgeorge007:
@shiftkey Hope it's fine now :+1:

jamesgeorge007:
@niik What do you think?

shiftkey:
@jamesgeorge007 apologies for the delay in reviewing this.

I was about to merge this but then I realised there's a better solution that doesn't require checking in the app every time a new repository is created. It unfortunately requires some understanding of how we're sourcing these licenses, which I didn't think to explain here.

Please have a read of #7149, test out the changes I proposed, and see if that makes sense.

shiftkey:
@jamesgeorge007 if you'd like something fresh to pick up, I'd recommend https://github.com/desktop/desktop/issues/6638

jamesgeorge007:
@shiftkey Sure, Thanks :+1:

shiftkey:
Closing this out in favour of #7149

PR #6776, opened 2019-02-05
Split repository publication settings into two types

iAmWillShepherd:
Fixes #6546
Related to #6636.

This PR includes a refactor to better support #6636 with types. I've split the publication settings into two distinct types, `DotcomPublicationSettings` and `GhePublicationSettings`. This was done because we were previously sharing state between the two tabs which meant changes made in one tab would not persist once another tab was selected. 

---
**From original PR**

cc @brendonbarreto 

## Release notes

Notes: [Fixed] Publish settings remembered when switching between publish targets

PR #7073, opened 2019-03-12
Enterprise users should be able to publish to an organization

### LINKED ISSUE #7052: GITHUB Enterprise organization selection cant be switched
**Description**
Hi,

When I’m trying to publish my project (Enterprise) and want to link it with specific organization it can be clicked:
https://enterprise.githubsupport.com/attachments/token/Cz7b6MX7AZniCjKcMNGmRwJvw/?name=image001.png

After I click the ‘none’ option it will open my all organization by when I click on one of them it won’t do the selection,

And because of it , it will upload to my users organization.

 

Will be great if you can check it.

Thx 

**Version**
1.6.3

**Operating system**
windows 7 SP1

**Expected Behavior**
organization selection should be switched and chosen according to users selection.

**Actual Behavior**
selection inst changing and only 'NONE' selection is available

shiftkey:
## Overview

**Closes #7052**

## Description

Publish repository shows the available organization for both GitHub and GitHub Enterprise accounts, but there were some subtle changes in #6776 that were also introduced:

 - `renderOrgs` did not render the selected organization if you were on a GitHub Enterprise account - 6eb0a58799d14f0e1d025c67fbccf979c72f1100
 - `onOrgChange` was skipped for an Enterprise account e0d85ddf90d557447514f64d50da06d32ccd95c3

I'm not sure where this requirement came from, but it means that Enterprise users aren't able to publish to an organization of their choosing from in the app. There's a workaround in https://github.com/desktop/desktop/issues/7052#issuecomment-472059849 but if we feel this is urgent enough we should co-ordinate a hotfix update later in the week.

If we agree with that approach, the first two commits should be cherry-picked on top of the release branch - 4eb56e4eb is a cleanup because of #6945.

## Release notes

Notes: `[Fixed] Publish Repository does not let you choose an organization on your Enterprise account`

### DIFF
--- a/app/src/models/publish-settings.ts
+++ b/app/src/models/publish-settings.ts
@@ -20,6 +20,12 @@ export interface IEnterprisePublicationSettings {
 
   /** Should the repository be private? */
   readonly private: boolean
+
+  /**
+   * The org to which this repository belongs. If null, the repository should be
+   * published as a personal repository.
+   */
+  readonly org: IAPIUser | null
 }
 
 export interface IDotcomPublicationSettings {
--- a/app/src/models/publish-settings.ts
+++ b/app/src/models/publish-settings.ts
@@ -25,7 +25,7 @@ export interface IEnterprisePublicationSettings {
    * The org to which this repository belongs. If null, the repository should be
    * published as a personal repository.
    */
-  readonly org: IAPIUser | null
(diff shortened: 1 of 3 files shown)


outofambit:
also, @iAmWillShepherd what do you think?

PR #7063, opened 2019-03-11
1.6.4 hotfix release

shiftkey:
## Overview

This PR is intended to close out some outstanding issues that arose from the `1.6.3` release last week:

 - [Electron Security Fix](https://electronjs.org/blog/filereader-fix) - not directly affecting Desktop, but good to pull in as soon as possible
 - #7028 - we've not been able to reproduce this issue on our end, but given Git for Windows was updated as part of this release we're fairly sure it's a regression that has been fixed in a later release. As we're not tied to any features as part of this release, we can revert back to the version of Git which we were using in `1.6.2`
 - #7038 - we feel this was also an artifact of the Git upgrade, but the experience can be improved on top of reverting to the previous version of Git - see #7062 for more details

There are other related issues like #7040 and #7028 but we believe they're relating back to the same root cause.

## Description

TODO:

 - [x] bump version for beta
 - [x] draft release notes for beta
 - [x] get approval on changes and release notes
 - [x] ship to beta
 - [x] get feedback that #7028 is resolved 
 - [x] get feedback that #7038 is resolved
 - [x] bump version and update release notes for production
 - [x] ship to production
 - [x] merge in `development` so this PR is mergeable

## Release notes

Notes: no-notes

PR #7074, opened 2019-03-12
back out of dugite upgrade

shiftkey:
## Overview

We rolled back to an old version of `dugite` in #7063 but `development` currently has a later version which supports new error codes. While we figure out how to balance upgrading Git with supporting building features for Desktop (more details on this soon) I wanted to ensure we're on a known good version of `dugite` rather than testing with a known problem version of Git.

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -26,7 +26,7 @@
     "deep-equal": "^1.0.1",
     "dexie": "^2.0.0",
     "double-ended-queue": "^2.1.0-0",
-    "dugite": "1.85.0",
+    "dugite": "1.80.0",
     "electron-window-state": "^4.0.2",
     "event-kit": "^2.0.0",
     "file-uri-to-path": "0.0.2",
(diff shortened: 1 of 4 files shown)


shiftkey:
@desktop/engineering there's an interesting note to this problem which I want to call out as impacting the planned rebase support in Desktop, via the [Git documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-rebaseuseBuiltin):

> **`rebase.useBuiltin`**
Set to `false` to use the legacy shellscript implementation of git-rebase[1]. Is `true` by default, which means use the built-in rewrite of it in C.
>
> **The C rewrite is first included with Git version 2.20.** This option serves an an escape hatch to re-enable the legacy version in case any bugs are found in the rewrite. This option and the shellscript version of git-rebase[1] will be removed in some future release.
>
> If you find some reason to set this option to false other than one-off testing you should report the behavior difference as a bug in git.

So we're kind of stuck at the moment:

 - The version of Git in this PR (`2.19.2`) does not contain the built-in rebase command, and will be significantly slower on plaforms like Windows because of how I/O and spawning processes works there
 - We can't use the version currently on `development` (Git `2.20.0`) because of the issues identified with networked drives and long paths
 - We don't have a package yet supporting Git `2.21.1` to support upgrading the app to use the builtin - details here https://github.com/desktop/dugite-native/issues/193
 - Git 2.22.0 is scheduled for release at [the end of May](https://calendar.google.com/calendar/embed?src=jfgbl2mrlipp4pb6ieih0qr3so@group.calendar.google.com&ctz=America/Los_Angeles).

shiftkey:
Superseded by #7146

Issue #7057, opened 2019-03-11
Bump Dugite to version 1.86.0


Daniel-McCarthy:
This issue suggests to bump Dugite to version 1.86.0. This is in order to get the benefits of the changes in [Dugite #308](https://github.com/desktop/dugite/pull/308) that allow the clone connection error to be correctly detected as `DugiteError.HostDown` and display the correct error description. This would be useful for Issue #926 for testing the retry dialog.

However, in order to bump Dugite to version 1.86.0, a change is required. A description needs to be added for `DugiteError.UnresolvedConflicts` in the `getDescriptionForError` function in `src/lib/git/core.ts`.

Without this change, after bumping Dugite and building, and error will occur. `ERROR in [at-loader] ./app/src/lib/git/core.ts:272:26 TS2345: Argument of type 'GitError.UnresolvedConflicts' is not assignable to parameter of type 'never'.`

shiftkey:
@Daniel-McCarthy in the meantime feel free to upgrade the package in your branch to ensure your feature work isn't blocked. If we haven't addressed this by the time it's ready to review, feel free to link to this issue in your PR as a blocker.

PR #7080, opened 2019-03-13
Add option to retry cloning upon thrown clone errors

### LINKED ISSUE #926: Provide a better error message and give users the option to retry a failed clone
As of https://github.com/desktop/desktop/pull/924, we tell the user when a clone has failed, but the user now has to close out of the error message, go back into the clone dialog, select the same repo again, and clone. A better solution would be to provide users the option to immediately retry cloning the same repo if it fails.

Daniel-McCarthy:
## Overview

**Closes #926**

## Description

- Handles thrown clone errors and displays popup to show error and offer to retry or dismiss.

image
image

~~Note: Although the functionality of this pull request does not depend on it, testing the `HostDown` cloning error will not be possible until Issue #7057 is resolved and merged into this branch.~~
**Update:** Dugite has been updated and the development branch has been merged into this PR. This should now be able to be tested appropriately with the `HostDown` error.

## Release notes

Notes: Adds ability to retry cloning when error is shown.

### DIFF
--- a/app/src/ui/clone-repository/retry-clone-dialog.tsx
+++ b/app/src/ui/clone-repository/retry-clone-dialog.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react'
 import { RetryAction } from '../../models/retry-actions'
 import { Button } from '../lib/button'
-import { ButtonGroup } from '../lib/button-group'
 import { Dialog, DialogContent, DialogFooter } from '../dialog'
+import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
 import { Ref } from '../lib/ref'
 import { Dispatcher } from '../dispatcher'
 import { Repository } from '../../models/repository'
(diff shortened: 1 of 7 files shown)


tierninho:
Tested this and the fix looks nice. Only tested a misspelled URL though, with `hostdown` pending the dugite update. 

Not sure if I prefer the "flashing lights" animation between the foreground and background ... but not hung up on it. A wider modal would cover up the background animation.
Mar-14-2019 14-53-02

Daniel-McCarthy:
@tierninho That is definitely a good point. I can certainly widen the popup dialog.
Part of this is due to the fact that I have it set up to close the dialog then initiate attempting to retry cloning (Instead of closing it after cloning is complete).

The reason I chose to do this is because it did not seem preferable to have this dialog up during the entire cloning process.

app/src/ui/clone-repository/retry-clone-dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <em>{this.props.repository.name}</em> or dismiss this warning.
+        </p>
+      )
+    }
+
+    return (
+      <div>
+        {this.props.errorMessage}
+        <p>
+          Would you like to retry cloning <em>{this.props.repository.name}</em>?
```
niik: Just a drive by comment to say that we have a component which should be doing what you're attempting to accomplish with the `<em>` here, namely the `<Ref>` component. It's a bit funnily named but it's only a styling component and we use it frequently for exactly the purpose you're looking for.

Daniel-McCarthy: Great catch! Thanks for letting me know, I will use the `Ref` component instead.

Daniel-McCarthy:
That's great news! No worries at all, thanks for taking on the PR for review!

PR #7032, opened 2019-03-07
Add function that emulates `git stash list`

iAmWillShepherd:
Adds a function that returns a list of stash entries.

See tracking issue: #6987

Issue #6429, opened 2018-12-15
merge conflicts state allows to keep making more merges?


msftrncs:
## Description
Closing the guided conflicts modal reveals that additional merges (or even the same merge again) can be attempted, but results in GIT errors.  I suspect that undo or revert could be bad ideas at this time too.  Error messages are given (hinting conflicts need resolved first), and nothing bad seems to happen.

UNDO could be worse, as it just automatically merges the undo in to your working tree, with the conflict banner remaining on screen, but no conflicts present (via icons) though the original conflicted file remains as if it were conflicted.  (see below)

## Version
* GitHub Desktop: 1.5.1-beta5 (but self-built using Node 10)
* Operating system: Win 10 18298 preview

## Steps to Reproduce
1. close the merge conflict modal
2. attempt another merge because all the merge buttons are still enabled.

### Expected Behavior
I might expect only finishing or aborting the current merge would be possible, or that doing some things should bring up the 'abort conflicted merge' flow (confirmation that is really want you want to do, then unwind the conflicted merge) before continuing.

### Actual Behavior
The merge buttons attempt to merge more changes, or even the same changes again, but results in GIT errors.

## Additional Information
Keep in mind that my understanding of conflict resolution is limited as I have not dealt a lot with it.  I was testing the latest beta since it added the 'close guided conflict modal' feature.

merge button enabled for the exact same merge I just attempted that I am still in conflict resolution for.
![image](https://user-images.githubusercontent.com/26179051/50039210-af6b9800-fff3-11e8-905f-bab2a6425aa2.png)

Merge conflict markers not treated special, is it not supposed to be?
![image](https://user-images.githubusercontent.com/26179051/50039143-9a423980-fff2-11e8-8bc6-3724938caeb1.png)

After pressing 'undo'...   yikes, how am I going to fix this?  (good its just a sandbox!)
The conflict banner went away when I closed Desktop and reopened it, and the guided conflict modal did not reopen either.
![image](https://user-images.githubusercontent.com/26179051/50039320-33268400-fff6-11e8-9fd7-488983303571.png)

shiftkey:
@outofambit @iAmWillShepherd I see you were both involved with #6382 - want to have a read of this and see if it's something you're aware of?

outofambit:
#6428 should help with this, but i’m not sure that it’s a complete solution

tierninho:
[...]

PR #7085, opened 2019-03-14
Adds function to create stash entry

iAmWillShepherd:
Adds function that creates a stash entry.

See tracking issue: #6987


app/test/unit/git/stash-test.ts
```diff
@@ -46,11 +47,39 @@ describe('git/stash', () => {
...
+
+    beforeEach(async () => {
+      repository = await setupEmptyRepository()
+      readme = path.join(repository.path, 'README.md')
+      await FSE.writeFile(readme, '')
+      await GitProcess.exec(['add', 'README.md'], repository.path)
+      await GitProcess.exec(['commit', '-m', 'initial commit'], repository.path)
+    })
+
+    it('creates a stash entry', async () => {
```
iAmWillShepherd: I can add additional tests to cover scenarios like being in the middle of a rebase or merge, but I'm not sure of the value since we do a pretty good job of not allowing actions to take place on invalid repos.

shiftkey:
As #7032 has been merged I changed the base branch back to `development` to indicate this is unblocked.

iAmWillShepherd:
🍎

app/test/unit/git/stash-test.ts
```diff
@@ -46,13 +47,46 @@ describe('git/stash', () => {
...
+      await await FSE.appendFile(readme, 'just testing stuff')
+      const tipSha = await getTipSha(repository)
+
+      await createDesktopStashEntry(repository, 'master', tipSha)
+
+      const result = await GitProcess.exec(['stash', 'list'], repository.path)
+      const entries = result.stdout.trim().split('\n')
+      expect(entries).toHaveLength(1)
```
shiftkey: Is there any other assertions you can do here about the results? `exitCode`? Contents of this line?

app/test/unit/git/stash-test.ts
```diff
@@ -46,13 +47,46 @@ describe('git/stash', () => {
...
+      repository = await setupEmptyRepository()
+      readme = path.join(repository.path, 'README.md')
+      await FSE.writeFile(readme, '')
+      await GitProcess.exec(['add', 'README.md'], repository.path)
+      await GitProcess.exec(['commit', '-m', 'initial commit'], repository.path)
+    })
+
+    it('creates a stash entry', async () => {
+      await await FSE.appendFile(readme, 'just testing stuff')
+      const tipSha = await getTipSha(repository)
```
shiftkey: I'd recommend using the `getTipOrError` helper function here

app/test/unit/git/stash-test.ts
```diff
@@ -46,13 +47,46 @@ describe('git/stash', () => {
...
+      expect(entries).toHaveLength(1)
+    })
+  })
 })
 
-async function stash(repository: Repository, message?: string) {
+async function getTipSha(repository: Repository) {
   const result = await GitProcess.exec(['rev-parse', 'HEAD'], repository.path)
-  const tipSha = result.stdout.trim()
+  return result.stdout.trim()
```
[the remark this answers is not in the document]
shiftkey: And I see now this was partially addressed in https://github.com/desktop/desktop/pull/7089/commits/a61292d4cf1373162805ec37836c6616e804178b but the rest of the points are still relevant

app/test/unit/git/stash-test.ts
iAmWillShepherd: New assertion added - https://github.com/desktop/desktop/pull/7085/commits/df10c173c8f2e000e539dc41fb0b0a52970c26be

app/test/unit/git/stash-test.ts
shiftkey: I'd try it out purely from a "what happens" scenario given the app might not always prevent the user from doing things - for example #6429 lets users try and merge when already in a merge, and it fails with some helpful context - and it'd be good to ensure your code behaves in a known & predictable way.

Try it out using a repository in the middle of a merge conflict (we should have scaffolding for this) - you might find yourself saying "uh-oh, we shouldn't allow any stashing at this time because XYZ" which would influence what your functions do as pre-conditions to doing stash-related work.

shiftkey:
Looks like this test needs to be reviewed:

```
FAIL app/test/unit/git/stash-test.ts
  ● git/stash › getDesktopStashEntries › handles unborn repo by returning empty list

    Unable to find commit for HEAD - check that you are not working with an unborn repository

       7 | 
       8 |   if (commit === null) {
    >  9 |     throw new Error(
         |           ^
      10 |       'Unable to find commit for HEAD - check that you are not working with an unborn repository'
      11 |     )
      12 |   }

      at Object.getTipOrError (test/helpers/tip.ts:9:11)
```

iAmWillShepherd:
Removed `Ready for review` label while I investigate why my tests only fail on Windows.

iAmWillShepherd:
It seems my tests are failing due to me inspecting the `GitError` thrown as a result of calling `createDesktopStashEntry`

https://github.com/desktop/desktop/blob/6035715d99464c45c37b844720801cb6ae0c1047/app/test/unit/git/stash-test.ts#L67-L85

☝️Fails on Windows because I was inspecting `stderr` for the string "fatal: Needed a single revision!". That is not returned on Windows.

**Windows**
Windows

**macOS**
Screen Shot 2019-03-28 at 11 11 55 AM

https://github.com/desktop/desktop/blob/6035715d99464c45c37b844720801cb6ae0c1047/app/test/unit/git/stash-test.ts#L87-L107

☝️Fails on Windows for the same reason as above.

**Windows**
Screen Shot 2019-03-28 at 11 27 04 AM

**macOS**
Screen Shot 2019-03-28 at 11 26 50 AM


The underlying problem is I'm trying to figure out what type of error I have to make sure it's failing the right way. I think I will remove this error inspection so I can move forward.

iAmWillShepherd:
Closing in favor of #7182.

Issue #6962, opened 2019-02-27
[rebase] show progress of an in-flight rebase


shiftkey:
Extracted from #6551

#6961 just shows a loading spinner, but the designs show a dedicated dialog with progress bar:

<img width="359"  src="https://user-images.githubusercontent.com/359239/53507029-8785bf80-3a8d-11e9-91ee-7211c5fcbab5.png">

Issue #3538, opened 2017-12-07
Move away from relying on JS truthiness


iAmWillShepherd:
See PR #3535 for more context. 

I think we should be explicit in our conditional statements instead of relying on truthiness. This should limit gotchas such as the one referenced above.

```ts
let a: ReadonlyArray<number> | null = null
let s: string | undefined = undefined

if (a) { }
if (a != null) { } //preferred

a = [0, 1, 2, 3]

if (a.length) { }
if (a.length > 0) { } //preferred
```

j-f1:
I’ll take a look at this.

1️⃣ ❓:

Should this throw?

```js
let x: boolean = myFunction()

if (x) { }
```

joshaber:
👍 

```js
if (a !== null) { } //preferred
if (s !== undefined) { } //preferred
```

I'd lump these two together and suggest that we prefer:

```js
if (s != null) { } //preferred
```

It's rare that we care about the distinction between `null` and `undefined`, and I'd go further and say that caring is usually an anti-pattern.

iAmWillShepherd:
The only reason I included `undefined` is that we are using optional types. 

Consider this:
```ts
function foo(obj?: SomeObject) {
  let a = obj.someProperty   //obj can't be null here, but it can be undefined, which means this fails
}
```

In this case, we'd have to check for `undefined`. We've been getting around this issue by relying on the truthiness of `undefined`.

iAmWillShepherd:
Should this throw?

```js
let x: boolean = myFunction()

if (x) { }
```

@j-f1 no it should not. 

The intent behind this issue is that we should be explicit instead of relying on truthiness. I think `booleans` are exempt from this since they are the primitives all other conditional expressions evaluate.

joshaber:
> The only reason I included `undefined` is that we are using optional types. 
> 
> Consider this:
> ```ts
> function foo(obj?: SomeObject) {
>   let a = obj.someProperty   //obj can't be null here, but it can be undefined, which means this fails
> }
> ```
> 
> In this case, we'd have to check for `undefined`. We've been getting around this issue by relying on the truthiness of `undefined`.

Hmm, I'm not sure I understand?

`!= null` coerces `undefined` so that both `null != null` and `null != undefined` are false. Or in other words, it generalizes across both `null` and `undefined`.

iAmWillShepherd:
But `null !== undefined` return `true`. 

[...]

PR #7087, opened 2019-03-14
propagate progress when performing a rebase

shiftkey:
## Overview

**Plumbing and internals related to #6962**

## Description

This PR adds the relevant Git plumbing to support propagating progress. This is living inside `rebase.ts` rather than `app/src/lib/progress` for a couple of reasons:

 - rebase progress is emitted by `stdout`, whereas typical Git progress is emitted by `stderr` (that's why i have a standalone `createStdoutProgressProcessCallback`)
 - rebase progress can start from 0% (when starting the rebase) or from some percent less than 100% (when continuing the rebase) - this is what I need to figure out next, but this PR adds support for the progress parser to know where it started from

**For the moment there's no callers that use this, but I have an upcoming PR next that builds upon this to implement #6962.**

There were also some related changes that I introduced that overlapped with these changes:

 - define a `IGitProgressParser` interface - my rebase parser doesn't need `steps`, `stepIndex` and `lastPercent`  that were part of `GitProgressParser`, and I believe those should be internal to the parser anyway 60fe84686737d3782d9bd69de1ee0ae0e755f1da
 - I updated some callers to `continueRebase` to ensure they were passing through the correct `manualResolutions` map 46a5ce85b769110d0db938625166ce7a3eb2314c

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -15,8 +15,7 @@ import { stageManualConflictResolution } from './stage'
 import { stageFiles } from './update-index'
 
 import { getStatus } from './status'
-import { RebaseContext } from '../../models/rebase'
-import { IRebaseProgress } from '../../models/progress'
+import { RebaseContext, RebaseProgressOptions } from '../../models/rebase'
 import { IGitProgress, IGitOutput, IGitProgressParser } from '../progress'
 import { merge } from '../merge'
 
@@ -137,18 +136,6 @@ function createStdoutProgressProcessCallback(
       progressCallback(parser.parse(line))
     })
   }
-}
-
-/**
- * Options to pass in to rebase progress reporting
- */
-export type RebaseProgressOptions = {
-  /** The number of commits already rebased as part of the operation */
-  start: number
-  /** The number of commits to be rebased as part of the operation */
-  total: number
-  /** The callback to fire when rebase progress is reported */
-  progressCallback: (progress: IRebaseProgress) => void
 }
 
 /**
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -16,7 +16,7 @@ import { stageFiles } from './update-index'
 
 import { getStatus } from './status'
 import { RebaseContext, RebaseProgressOptions } from '../../models/rebase'
-import { IGitProgress, IGitOutput, IGitProgressParser } from '../progress'
+import { IGitProgress, IGitOutput } from '../progress'
 import { merge } from '../merge'
 
 /**
@@ -90,7 +90,7 @@ export async function getRebaseContext(
 /**
  * A parser to read and emit rebase progress from Git `stdout`
  */
-class GitRebaseParser implements IGitProgressParser {
+class GitRebaseParser {
   private currentCommitCount = 0
 
   public constructor(
@@ -115,29 +115,6 @@ class GitRebaseParser implements IGitProgressParser {
   }
 }
 
-/**
- * Reads `stdout` and uses the provided parser to emit progress to the caller.
- *
- * Our default progress parsing infrastructure reads `stderr` as this is how Git
- * typically emits progres, but `git rebase` is a special case. Our default
- * progress parsing also supports more general use cases, so this felt simpler
- * to implement as a first iteration than baking more complexity into the
- * defaults currently.
- *
- * @param parser the provided parser to process `stdout`
- * @param progressCallback the callback to invoke with received progress data
- */
-function createStdoutProgressProcessCallback(
-  parser: IGitProgressParser,
-  progressCallback: (progress: IGitProgress | IGitOutput) => void
-): (process: ChildProcess) => void {
-  return process => {
-    byline(process.stdout).on('data', (line: string) => {
-      progressCallback(parser.parse(line))
-    })
-  }
-}
-
 function configureOptionsForRebase(
   options: IGitExecutionOptions,
   progress?: RebaseProgressOptions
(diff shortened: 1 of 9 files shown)


app/src/lib/progress/git.ts
```diff
@@ -185,13 +199,6 @@ export class GitProgressParser {
     }))
   }
 
-  /**
-   * Parse the given line of output from Git, returns either an IGitProgress
-   * instance if the line could successfully be parsed as a Git progress
-   * event whose title was registered with this parser or an IGitOutput
-   * instance if the line couldn't be parsed or if the title wasn't
-   * registered with the parser.
-   */
```
niik: I think we should keep this documentation here and modify the documentation for the interface version of this. A GitProgressParser was specifically intended to parse the well-defined progress format of Git commands which supports `--progress` whereas the rebase progress parser parses a custom format. As such the `steps` concepts in this concrete class doesn't apply to the rebase parser and consequently shouldn't be described in the interface either. So I suggest reinstating this and I'll add a suggested change for the interface documentation.

app/src/lib/progress/git.ts
```diff
@@ -137,6 +137,20 @@ export interface IGitProgressInfo {
...
 }
 
+/**
+ * Base interface for parsing progess reported from Git
+ */
+export interface IGitProgressParser {
+  /**
+   * Parse the given line of output from Git, returns either an IGitProgress
+   * instance if the line could successfully be parsed as a Git progress
+   * event whose title was registered with this parser or an IGitOutput
```
niik:
```suggestion
   * event or a IGitOutput instance if the line couldn't be parsed.
```

app/src/lib/progress/git.ts
```diff
@@ -137,6 +137,20 @@ export interface IGitProgressInfo {
...
+   * instance if the line couldn't be parsed or if the title wasn't
```
niik:
```suggestion
```

app/src/lib/progress/git.ts
```diff
@@ -137,6 +137,20 @@ export interface IGitProgressInfo {
...
+   * registered with the parser.
```
niik:
```suggestion
```

app/src/models/progress.ts
```diff
@@ -99,6 +99,12 @@ export interface IRevertProgress extends IProgress {
   kind: 'revert'
 }
 
+/** An object describing the progress of a rebase operation */
+export interface IRebaseProgress {
```
niik: I think this should follow the convention of the other progress shapes we've got here. I.e have it extend the `IProgress` interface, state a `kind` and use the fractional value rather than a 0-100 integer value. The `title` should be something like `Rebasing x onto y` or something to that effect.

We should also include it in the Progress union type below.

shiftkey: > I.e have it extend the `IProgress` interface, state a `kind` and use the fractional value rather than a 0-100 integer value. The `title` should be something like `Rebasing x onto y` or something to that effect.

I haven't thought about the latter but I think I've already changed `progress` this as part of implementing a working progress bar. I'll pull those changes into this PR.

app/src/lib/progress/git.ts
shiftkey: This has been addressed in bfe7846e7

app/src/models/progress.ts
shiftkey: Addressed in 0ec1ffcfe008ad8656337545f030cdaa32821eba

shiftkey:
:apple:

app/src/lib/git/rebase.ts
```diff
@@ -82,6 +86,73 @@ export async function getRebaseContext(
...
+  public constructor(
+    startCount: number,
+    private readonly totalCommitCount: number
+  ) {
+    this.currentCommitCount = startCount
+  }
+
+  public parse(line: string): IRebaseProgress | null {
+    const match = rebaseApplyingRe.exec(line)
+    if (match == null || match.length !== 2) {
```
niik: Can we make the null check a triple equals please?

```suggestion
    if (match === null || match.length !== 2) {
```

app/src/lib/git/rebase.ts
```diff
@@ -82,6 +86,73 @@ export async function getRebaseContext(
...
+      commitSummary,
+    }
+  }
+}
+
+function configureOptionsForRebase(
+  options: IGitExecutionOptions,
+  progress?: RebaseProgressOptions
+) {
+  if (!progress) {
```
niik: Just to be clear, I think the pattern you're using here is nice and unambiguous but it seems like we've converged on being explicit (https://github.com/desktop/desktop/issues/3538) so for consistency could we make this into a triple equals check please?

```suggestion
  if (progress === undefined) {
```

app/src/lib/git/rebase.ts
```diff
@@ -82,6 +86,73 @@ export async function getRebaseContext(
...
+    }
+
+    const commitSummary = match[1]
+    this.currentCommitCount++
+
+    const value = this.currentCommitCount / this.totalCommitCount
+
+    return {
+      kind: 'rebase',
+      title: `Rebasing ${this.currentCommitCount} of ${
```
niik: Assuming this is something that'll be presented to the user could we add a word about what the unit of work is here?

```suggestion
      title: `Rebasing commit ${this.currentCommitCount} of ${
```

shiftkey: 👍 - current designs can be found here https://github.com/desktop/desktop/issues/6962

PR #7089, opened 2019-03-14
Adds function to return last stash entry created for a branch

iAmWillShepherd:
⚠️ This builds on top of work done in #7085, so review that first if it's not merged ⚠️ 

Adds a function that returns the last entry created for a given branch.

See tracking issue: #6987


app/src/lib/git/stash.ts
```diff
@@ -69,17 +69,41 @@ export async function getDesktopStashEntries(
   return stashEntries
 }
 
+export async function getLastStashEntry(
+  repository: Repository,
+  branchName: string
+) {
+  const entries = await getDesktopStashEntries(repository)
+
+  return entries.find(stash => stash.branchName === branchName) || null
```
iAmWillShepherd: This function relies on the sort order of `git stash list`. Since that ordering is reverse-chronological order, `find` should be sufficient to return the **first** match stash entry.

app/test/unit/git/stash-test.ts
```diff
@@ -25,42 +27,119 @@ describe('git/stash', () => {
...
+  return result.stdout.trim()
+}
+
+async function stash(
+  repository: Repository,
+  branchName: string,
+  message: string | null
+): Promise<string> {
+  const tipSha = await getTipSha(repository)
+  const result = await GitProcess.exec(['stash', 'create'], repository.path)
```
iAmWillShepherd: I need the object Id of the stash entry because I'm asserting based on that in test for `getLastStashEntry`

app/test/unit/git/stash-test.ts
```diff
@@ -25,42 +27,119 @@ describe('git/stash', () => {
...
+
+      const entry = await getLastStashEntry(repository, 'master')
+
+      expect(entry).toBeNull()
+    })
+
+    it('returns last entry made for branch', async () => {
+      const branchName = 'master'
+      await generateTestStashEntry(repository, branchName, true)
+      const lastEntry = await generateTestStashEntry(
```
iAmWillShepherd: I'm using the object Id to assert that I have the last entry created.

app/test/unit/git/stash-test.ts
```diff
@@ -25,42 +27,119 @@ describe('git/stash', () => {
...
+    beforeEach(async () => {
+      repository = await setupEmptyRepository()
+      readme = path.join(repository.path, 'README.md')
+      await FSE.writeFile(readme, '')
+      await GitProcess.exec(['add', 'README.md'], repository.path)
+      await GitProcess.exec(['commit', '-m', 'initial commit'], repository.path)
+    })
+
+    it('creates a stash entry', async () => {
+      await await FSE.appendFile(readme, 'just testing stuff')
```
shiftkey:
```suggestion
     await FSE.appendFile(readme, 'just testing stuff')
```

app/test/unit/git/stash-test.ts
```diff
@@ -25,42 +27,119 @@ describe('git/stash', () => {
...
+
+      const actual = await getLastStashEntry(repository, branchName)
+
+      expect(actual!.stashSha).toBe(lastEntry)
+    })
+  })
 })
 
-async function stash(repository: Repository, message?: string) {
+async function getTipSha(repository: Repository) {
```
shiftkey: 💭 this helper function might be more useful here rather than rolling your own:

https://github.com/desktop/desktop/blob/385be4ddee002f6f04e776f86a2d73df0f00b19d/app/test/helpers/tip.ts#L5-L15

app/test/unit/git/stash-test.ts
shiftkey: 💭 if you find yourself writing comments to explain the intent of a function, maybe these could live as comments in the source (either as API documentation or inside the function, whichever help to explain the intent better)

app/src/lib/git/stash.ts
shiftkey: 📝 this also seems worth a code comment so it's not easily forgotten once this PR is merged

iAmWillShepherd: Added to the docs for `getDesktopStashEntries`.

shiftkey:
As #7032 has been merged I changed the base branch back to `development` to indicate this is unblocked.

app/test/unit/git/stash-test.ts
```diff
@@ -61,47 +62,89 @@ describe('git/stash', () => {
...
+      repository = await setupEmptyRepository()
+      readme = path.join(repository.path, 'README.md')
+      await FSE.writeFile(readme, '')
+      await GitProcess.exec(['add', 'README.md'], repository.path)
+      await GitProcess.exec(['commit', '-m', 'initial commit'], repository.path)
+    })
 
-async function stash(repository: Repository, message?: string) {
-  const tipSha = await getTipSha(repository)
+    it('returns null when no stash entries exist for branch', async () => {
```
shiftkey: What about a test when no stashes are created in the repository?

app/test/unit/git/stash-test.ts
```diff
@@ -61,47 +62,89 @@ describe('git/stash', () => {
...
+  const message = createdByDesktop ? null : 'Should get filtered'
   const readme = path.join(repository.path, 'README.md')
-
-  // simulate stashing from CLI
   await FSE.appendFile(readme, '1')
-  await stash(repository, 'should get filtered')
-
-  await FSE.appendFile(readme, '2')
-  await stash(repository, 'should also get filtered')
+  const objectId = await stash(repository, branchName, message)
```
shiftkey: :art: could just `return await` here instead of using the a local variable

app/test/unit/git/stash-test.ts
iAmWillShepherd: The function operates at the branch level, so I don't think testing for stashes in a repo would give us much. I've updated the name of the function to make its purpose more clear.

shiftkey: I wasn't that worried about the branch name, given it's parameterized accordingly - I just didn't quite understand this test summary versus the implementation.

What about this test description for the current test:

>  `returns null when no stash entries match branch name`

This might then motivate creating an additional test to cover a related scenario

> `returns null when no stash entries created`

shiftkey: > The function operates at the branch level, so I don't think testing for stashes in a repo would give us much. 

This might be correct, but we also don't have a test for `getDesktopStashEntries` to ensure it handles when no stashes exist. We only cover "unborn repository" and "returns created stashes". That's the gap I'd like to cover here.

shiftkey:
I'm going to drop `ready-for-review` here because it's dependent on #7085 which currently has some testing issues.

iAmWillShepherd:
Closing in favor of #7182.

PR #7090, opened 2019-03-14
FRR and issue triage updates

### LINKED ISSUE #5486: Clarify and disambiguate roles of first responder and support
@steveward and I met this morning to chat about how we're handling [first responder rotation](https://github.com/desktop/desktop/blob/master/docs/process/first-responder.md) and his role as the support team member. The goal is to clearly disambiguate the two roles and draw clear lines of responsibility. To help get us there, a question and a few suggestions below. I'd like to update the first responder role doc to be clearer on expectations.

Questions:

1. Does first responder rotation still include reviewing PRs? If so, how do we handle when a non-dev is the first responder?

Suggestions:

- [x] Add a team for desktop/support so we can ping Steve in issues that require more support function (diagnosis of an individual's problem with the app or their setup).
- [ ] First responder responds to clarify the underlying problem, asks clarifying questions, and in that clarification, ping desktop/support if it's not obvious. If it's obvious, just give the answer.
- [ ] Goal is to respond within 24 hours on a weekday to initial issues and follow up replies. If first responder and Steve cannot meet this, they'll let us know in the channel to ask for help on specific portion(s) of the work.

cc: @desktop/core

billygriffin:
Closes #5486

This PR will update the first responder and issue triage docs to be slightly more reflective of current state, if not perfectly aspirational just yet. 

1. Removes reviewing PRs from explicit expectation of the first responder. It's certainly something they can do, but not all first responders do PR review ordinarily. 
2. Properly document the role of @desktop/support. 
3. Remove `future proposal` from issue triage labels because we're not actively using it anymore and trying to remove the existing ones.
4. Clarify that first responders should not be responsible for following up on every previous issue, but rather that they should highlight things that have gone without a response.

Note: There's probably more that could be done to improve the first responder doc in particular, as I'm not sure it's actually reflective of what we're doing. But I wanted to take baby steps.

### DIFF
--- a/docs/process/first-responder.md
+++ b/docs/process/first-responder.md
@@ -11,6 +11,7 @@ Each rotation is a week long. While first responder your primary duties are:
 
 1. Triage issues.
     * Troubleshoot or follow up on troubleshooting started by the previous first responder. The current first responder should always bear responsibility for pushing troubleshooting forward, unless another team member has explicitly taken ownership.
+    * At mention @desktop/support and add `support` label if the issue feels applicable to only the user reporting it, and isn't something more broadly relevant.
     * Ensure issues are labeled accurately.
     * Review issues labeled [`reviewer-needs-to-reproduce`](https://github.com/desktop/desktop/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+sort%3Aupdated-asc+label%3Areviewer-needs-to-reproduce) and close any that have gone 2 weeks with no new activity after the last question by a reviewer.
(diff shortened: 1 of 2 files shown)


nerdneha:
> One thing that doesn’t seem to be working well today is that our FRR doc says that the current first responder is the one who is supposed to follow up on things that are still open from the previous weeks. This hasn’t been practical because 1. they have less context than the original responder at the time, 2. they aren’t notified when the person responds because they were never part of the issue. I’d propose that the person who initial responds is responsible for closing the loop, but that may be out of scope for this PR.

I think the responsibility should not be on the FRR to follow up but it could be their responsibility to highlight to the team + comrades which issues have been left unanswered on our side for longer than X business days to make them more visible (and potential point out which are the highest priority).

maybe X = 5 days?

billygriffin:
Thanks @nerdneha, updates forthcoming.

PR #7096, opened 2019-03-15
Don't crash when failing to create log file

niik:
## Overview

I encountered this issue when testing some PR related functionality inside of a very conservatively sized virtual machine. Our `FileLogger` which is responsible for writing our log messages to disk (as well as rotating the files and cleaning up old log files) normally handles errors when trying to write to a log file when the disk is full but it emits (effectively throws) an unhandled exception when it fails to _create_ the log file in the first place. The reason for this could be that the user doesn't have sufficient permissions to write to the log file directory (unlikely but theoretically possible) or, more likely, due to the disk being completely full and unable to accommodate even an empty file.

In this scenario the application will crash on startup and immediately show the crash reporter dialog to the user.

## Description

Failing to create the log file (for whatever reason) is not a critical condition in the app and as such we shouldn't crash the application because of it. This PR changes the behavior to ignore any errors thrown by the FileLogger.

Now, while this ought to fix one of our biggest sources of crashes (wild, right?) in the app that obviously doesn't mean that the app will work flawlessly in an environment that doesn't have enough space to create a single file. It's even possible we'll start seeing crashes from elsewhere in the app as a result of removing this particular source of crashes but even if that's the case, the statement I made above still holds, we should never crash simply because we're unable to write to the log file.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/log.ts
+++ b/app/src/main-process/log.ts
@@ -49,7 +49,7 @@ function initializeWinston(path: string): winston.LogMethod {
   // disk being full). If logging fails that's not a big deal
   // so we'll just suppress any error, besides, the console
   // logger will likely still work.
-  fileLogger.on('error', _ => {})
+  fileLogger.on('error', () => {})
 
   const consoleLogger = new winston.transports.Console({
     level: process.env.NODE_ENV === 'development' ? 'debug' : 'error',
--- a/app/src/main-process/log.ts
+++ b/app/src/main-process/log.ts
@@ -43,6 +43,14 @@ function initializeWinston(path: string): winston.LogMethod {
     maxFiles: MaxLogFiles,
   })
 
+  // The file logger handles errors when it can't write to an
+  // existing file but emits an error when attempting to create
+  // a file and failing (for example due to permissions or the
+  // disk being full). If logging fails that's not a big deal
+  // so we'll just suppress any error, besides, the console
+  // logger will likely still work.
+  fileLogger.on('error', _ => {})
+
   const consoleLogger = new winston.transports.Console({
     level: process.env.NODE_ENV === 'development' ? 'debug' : 'error',
   })


app/src/main-process/log.ts
```diff
@@ -43,6 +43,14 @@ function initializeWinston(path: string): winston.LogMethod {
     maxFiles: MaxLogFiles,
   })
 
+  // The file logger handles errors when it can't write to an
+  // existing file but emits an error when attempting to create
+  // a file and failing (for example due to permissions or the
+  // disk being full). If logging fails that's not a big deal
+  // so we'll just suppress any error, besides, the console
+  // logger will likely still work.
+  fileLogger.on('error', _ => {})
```
shiftkey: Coding style suggestion:

  - when no arguments are declared on a callback function, use `() => ` 
  - when parameters need to be declared but are unused (e.g. callback only needs the second parameter), use `_` as the argument name (or `_arg` if there are multiple args) as `tsc` can infer whether these are used 







```suggestion
  fileLogger.on('error', () => {})
```

mido1966:
Thanks es ok Dankeschön alles ok

‏‫من الـ iPhone الخاص بي‬

‫في 15‏/03‏/2019 الساعة 4:07 م، كتب/كتبت ‏Markus Olsson ‏<notifications@github.com<mailto:notifications@github.com>>:‬

Overview

I encountered this issue when testing some PR related functionality inside of a very conservatively sized virtual machine. Our FileLogger which is responsible for writing our log messages to disk (as well as rotating the files and cleaning up old log files) normally handles errors when trying to write to a log file when the disk is full but it emits (effectively throws) an unhandled exception when it fails to create the log file in the first place. The reason for this could be that the user doesn't have sufficient permissions to write to the log file directory (unlikely but theoretically possible) or, more likely, due to the disk being completely full and unable to accommodate even an empty file.

In this scenario the application will crash on startup and immediately show the crash reporter dialog to the user.

Description

Failing to create the log file (for whatever reason) is not a critical condition in the app and as such we shouldn't crash the application because of it. This PR changes the behavior to ignore any errors thrown by the FileLogger.

Now, while this ought to fix one of our biggest sources of crashes (wild, right?) in the app that obviously doesn't mean that the app will work flawlessly in an environment that doesn't have enough space to create a single file. It's even possible we'll start seeing crashes from elsewhere in the app as a result of removing this particular source of crashes but even if that's the case, the statement I made above still holds, we should never crash simply because we're unable to write to the log file.

Release notes

Notes: no-notes

You can view, comment on, or merge this pull request online at:

  https://github.com/desktop/desktop/pull/7096

Commit Summary

  *   Don't crash when failing to write to create log file

File Changes

  *   M app/src/main-process/log.ts<https://github.com/desktop/desktop/pull/7096/files#diff-0> (8)

Patch Links:

  *   https://github.com/desktop/desktop/pull/7096.patch
  *   https://github.com/desktop/desktop/pull/7096.diff

PR #7117, opened 2019-03-18
move rebase flow to a multi-dialog flow

shiftkey:
## Overview

This PR moves the rebase flow to a high-level component to support this multi-dialog flow. In particular it now enables these features:

 - #6962 - show progress for an in-flight rebase
 - #6989 - confirm abort rebase if user has experience

Closes #6962 
Closes #6989
Closes #7104  
Closes #7140 

It's currently working but I want to polish this a bit more to focus on testing these flows to catch anything unexpected.

## Description

 - [x] working demo of the new flow
 - [x] wire up error handler for 'pull with rebase' to trigger new flow
 - [x] flesh out the progress dialog to show the extra context of the rebase
 - [x] smooth out the switching between "conflicts" and "progress" so that things aren't appearing and disappearing to be confusing
 - [x] track resolved conflicts that occur as part of rebase conflicts (needs to persist across commits)
 - [x] use this value to then indicate whether the app should confirm the abort rebase (skip if user hasn't resolved anything)
 - [x] ensure `Abort` button state is disabled while the `git rebase --abort` is in progress - #7104
 - [x] extract changes independent of this PR into a pre-tech-debt PR
 - [x] confirm aborting a rebase outside the app cleans up and re-enables menu again - #7140
 - [x] confirm completing a rebase outside the app cleans up and re-enables menu again - #7140
 - [x] rebase once #7129 is merged
 - [x] investigate missing `Dispatcher.loadStatus` related to #7165

## GIFs

Happy path:

Less-than-happy path (dealing with conflicts):

Don't worry too much about the specifics in the progress dialog - #7167 has improved this significantly now that it knows what commits will be rebased.

## Release notes

Notes: no-notes

PR #7129, opened 2019-03-20
some cleanup extracted from the rebase flow changes

shiftkey:
## Overview

This is a pre-requisite for the work in #7117, which can land in advance of that review.

## Description

I'm going to annotate the PR with context about these changes, but the commits are atomic enough that hopefully the rest makes sense.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -87,6 +87,24 @@ export async function getRebaseContext(
   return null
 }
 
+/**
+ * Attempt to read the `.git/REBASE_HEAD` file inside a repository to confirm
+ * the rebase is still active.
+ */
+async function readRebaseHead(repository: Repository): Promise<string | null> {
+  try {
+    const rebaseHead = Path.join(repository.path, '.git', 'REBASE_HEAD')
+    const rebaseCurrentCommitOutput = await FSE.readFile(rebaseHead, 'utf8')
+    return rebaseCurrentCommitOutput.trim()
+  } catch (err) {
+    log.warn(
+      '[rebase] a problem was encountered reading .git/REBASE_HEAD, so it is unsafe to continue rebasing',
+      err
+    )
+    return null
+  }
+}
+
 /** Regex for identifying when rebase applied each commit onto the base branch */
 const rebaseApplyingRe = /^Applying: (.*)/
 
@@ -277,6 +295,11 @@ export async function continueRebase(
     return RebaseResult.Aborted
   }
 
+  const rebaseCurrentCommit = await readRebaseHead(repository)
+  if (rebaseCurrentCommit === null) {
+    return RebaseResult.Aborted
+  }
+
   const trackedFilesAfter = status.workingDirectory.files.filter(
     f => f.status.kind !== AppFileStatusKind.Untracked
   )
@@ -292,11 +315,8 @@ export async function continueRebase(
   )
 
   if (trackedFilesAfter.length === 0) {
-    const rebaseHead = Path.join(repository.path, '.git', 'REBASE_HEAD')
-    const rebaseCurrentCommit = await FSE.readFile(rebaseHead, 'utf8')
-
     log.warn(
-      `[rebase] no tracked changes to commit for ${rebaseCurrentCommit.trim()}, continuing rebase but skipping this commit`
+      `[rebase] no tracked changes to commit for ${rebaseCurrentCommit}, continuing rebase but skipping this commit`
     )
 
     const result = await git(
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -234,6 +234,12 @@ export enum RebaseResult {
    * state.
    */
   Aborted = 'Aborted',
+  /**
+   * An unexpected error as part of the rebase flow was caught and handled.
+   *
+   * Check the logs to find the relevant Git details.
+   */
+  Error = 'Error',
 }
 
 function parseRebaseResult(result: IGitResult): RebaseResult {
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -37,3 +37,18 @@ export function promiseWithMinimumTimeout<T>(
       .catch(reject)
   })
 }
+
+/**
+ * `async`/`await`-friendly wrapper around `window.setTimeout` for places where
+ * callers want to defer async work and avoid the ceremony of this setup and
+ * using callbacks
+ *
+ * @param timeout the time to wait before resolving the promise (in milliseconds)
+ */
+export async function timeout(timeout: number): Promise<void> {
+  return new Promise((resolve, reject) => {
+    window.setTimeout(() => {
+      resolve()
+    }, timeout)
+  })
+}
--- a/app/src/models/banner.ts
+++ b/app/src/models/banner.ts
@@ -33,6 +33,6 @@ export type Banner =
       readonly type: BannerType.RebaseConflictsFound
       /** name of the branch that was used to rebase */
       readonly targetBranch: string
-      /** popup to be shown when opening the banner */
-      readonly popup: Popup
+      /** callback to run when user clicks on link in banner text */
+      readonly onOpenDialog: () => void
     }
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -981,7 +981,7 @@ export class App extends React.Component<IAppProps, IAppState> {
     return repository.gitHubRepository.htmlURL
   }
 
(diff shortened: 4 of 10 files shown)


app/src/lib/git/rebase.ts
```diff
@@ -87,6 +87,24 @@ export async function getRebaseContext(
...
+ * the rebase is still active.
+ */
+async function readRebaseHead(repository: Repository): Promise<string | null> {
+  try {
+    const rebaseHead = Path.join(repository.path, '.git', 'REBASE_HEAD')
+    const rebaseCurrentCommitOutput = await FSE.readFile(rebaseHead, 'utf8')
+    return rebaseCurrentCommitOutput.trim()
+  } catch (err) {
+    log.warn(
+      '[rebase] a problem was encountered reading .git/REBASE_HEAD, so it is unsafe to continue rebasing',
```
shiftkey: This adds better error handling to a scenario that @tierninho identified during QA - if we're trying to continue when there no active rebase, we shouldn't try to proceed.

app/src/lib/promise.ts
```diff
@@ -37,3 +37,18 @@ export function promiseWithMinimumTimeout<T>(
...
 }
+
+/**
+ * `async`/`await`-friendly wrapper around `window.setTimeout` for places where
+ * callers want to defer async work and avoid the ceremony of this setup and
+ * using callbacks
+ *
+ * @param timeout the time to wait before resolving the promise (in milliseconds)
+ */
+export async function timeout(timeout: number): Promise<void> {
```
shiftkey: This helper function was a neater way to support deferring asynchronous work. I have some upcoming changes when I switch between dialogs that made this sort of code a bit more readable:

```ts
  public async componentDidMount() {
    if (this.props.onDidMount) {
      await timeout(500)
      await this.props.onDidMount()
    }
  }
```

app/src/models/banner.ts
```diff
@@ -33,6 +33,6 @@ export type Banner =
       readonly type: BannerType.RebaseConflictsFound
       /** name of the branch that was used to rebase */
       readonly targetBranch: string
-      /** popup to be shown when opening the banner */
-      readonly popup: Popup
+      /** callback to run when user clicks on link in banner text */
+      readonly onOpenDialog: () => void
```
shiftkey: This change is required because in #7117 I no longer have a `popup` for the rebase conflicts - this is moved inside a "flow" that maintains it's own state. For the moment, this callback just wraps the current behaviour.

app/src/ui/banners/rebase-conflicts-banner.tsx
```diff
@@ -20,16 +20,22 @@ export class RebaseConflictsBanner extends React.Component<
...
   private openDialog = () => {
     this.props.onDismissed()
-    this.props.dispatcher.showPopup(this.props.popup)
+    this.props.onOpenDialog()
     this.props.dispatcher.recordRebaseConflictsDialogReopened()
   }
 
+  private onDismissed = () => {
+    log.warn(
+      `[RebaseConflictsBanner] this is not dismissable by default unless the user clicks on the link`
```
shiftkey: I'm not a huge fan of this callback because when I've set `dismissable` to `false` (expecting that this component will handle it's dismissal) it then feels unnecessary to require a `onDismissed` prop, but maybe I'm using this wrong. Thoughts?

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -720,6 +720,8 @@ export class Dispatcher {
         baseBranch: baseBranch,
       })
     }
+
+    return result
```
shiftkey: These two `return results` from the Dispatcher will help with handling state transitions inside the rebase flow.

app/src/ui/app.tsx
```diff
@@ -981,7 +981,7 @@ export class App extends React.Component<IAppProps, IAppState> {
     return repository.gitHubRepository.htmlURL
   }
 
-  private openCurrentRepositoryInShell() {
+  private openCurrentRepositoryInShell = () => {
```
shiftkey: We have a few places inside `App` where we've not used this syntax to ensure it gets bound, but without this change I can't use this in #7117 because tslint warns me:

```
ERROR: /Users/shiftkey/src/desktop/app/src/ui/app.tsx[1598, 36]: A class method is passed as a JSX attribute without having the 'this' reference bound in the constructor: this.openCurrentRepositoryInShell
```

app/src/lib/promise.ts
outofambit: This is neat, I'll just note I would rather see us rely less on `setTimeout` (except for the cases where css animation is involved since i don't know of a a better solution). This isn't blocking though, since I know we have a number of these in the codebase already.

Out of curiousity, @shiftkey is the above example a place you're planning to use this?

app/src/ui/app.tsx
outofambit: i'm all for this :)

app/src/ui/banners/rebase-conflicts-banner.tsx
outofambit: hmmm yeah that's interesting. i think my thought around requiring the `onDismissed` prop was that the banner would always dismiss itself, but i think the is a good counter example to that. i'd be happy with making the `onDismissed` prop in the banner component optional so you don't have to pass this in. how does that sit with you?

app/src/ui/dispatcher/dispatcher.ts
outofambit: can we explicitly add the type for `results` to the function type definition? i'm fine with this pattern (even though we don't generally do it), I just want to make sure it's as clear as possible.

app/src/lib/git/rebase.ts
```diff
@@ -87,6 +87,24 @@ export async function getRebaseContext(
...
 }
 
+/**
+ * Attempt to read the `.git/REBASE_HEAD` file inside a repository to confirm
+ * the rebase is still active.
+ */
+async function readRebaseHead(repository: Repository): Promise<string | null> {
+  try {
+    const rebaseHead = Path.join(repository.path, '.git', 'REBASE_HEAD')
+    const rebaseCurrentCommitOutput = await FSE.readFile(rebaseHead, 'utf8')
```
outofambit: do we care about the contents of `REBASE_HEAD`? i see you're returning it but the docs say we just want to see if its there. this could just be [`FSE.pathExists`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/pathExists.md) but maybe attempting to read the file gives us more confidence that a valid rebase is in progress?

app/src/lib/promise.ts
shiftkey: Yes. I'm open to other alternatives to achieve this, but ultimately I need some way to perform asynchronous work _after_ the component is mounted and displayed for the first time.

app/src/lib/git/rebase.ts
shiftkey: I use it in `continueRebase` to track when the in-flight rebase skips a commit (because no tracked changes):

```
    log.warn(
      `[rebase] no tracked changes to commit for ${rebaseCurrentCommit.trim()}, continuing rebase but skipping this commit`
      `[rebase] no tracked changes to commit for ${rebaseCurrentCommit}, continuing rebase but skipping this commit`
    )
```

app/src/ui/banners/rebase-conflicts-banner.tsx
shiftkey: I think this needs a bigger discussion than this change because `Dialog` is setup the same way, and making `onDismissed` optional would just move the problem elsewhere:

 - a `Banner` could somehow be set to `dismissable={true}` and `onDismissed={undefined}` due to a coding error
 - user click `Close` to dismiss the banner
 - what should happen here when `onDismissed` is undefined - a runtime error? a `development` runtime error but is silently ignored in `production`? something else?

https://github.com/desktop/desktop/blob/2d30990632997cbc192e7cc03ddae40aa819a4f4/app/src/ui/banners/banner.tsx#L39-L43

Screen Shot 2019-03-25 at 2 07 30 PM

 I'm happy to leave this warning in here for the moment rathe than change the accepted props values, but I'd like to think more about alternative ways to ensure the component doesn't get put in an invalid state.

outofambit: okay! that makes sense!

app/src/ui/dispatcher/dispatcher.ts
shiftkey: Addressed in 238cf4fef

PR #7131, opened 2019-03-21
Cancel active selection before swapping document

niik:
## Overview

This fixes a crash when selecting text in the diff while the contents of said diff is changed.

Upstream issue https://github.com/codemirror/CodeMirror/issues/5821

Rollups: 
 - https://haystack.githubapp.com/desktop/rollups/1fcdc5bd8456c02ea223ae71a77720d2
 - https://haystack.githubapp.com/desktop/rollups/59cd7a422fb4df8fbc4fb53fdaaedfae
 - https://haystack.githubapp.com/desktop/rollups/622e8dc1282275b1402d657b3bcb0054

### Steps to reproduce

- Create a text file in any repository and make it contain enough lines that scrolling is necessary in CodeMirror (scrolling isn't strictly necessary but helps to reproduce)
- View the changed file in desktop and scroll all the way down to the last line in the file
- Now switch to whatever editor you prefer and delete all lines in the file
- Switch back to desktop *by clicking, and holding down*, the left mouse button anywhere inside the diff

The diff should reload and you should encounter the crash if you move your cursor just a little.

This happens because when you clicked on the diff CodeMirror started a selection originating from the position you originally clicked. Desktop then (through the window focus event) reloads the diff and swaps out the document. When you then move your cursor again CodeMirror attempts to extend the selection from a position that no longer exists in the current document.

```
Error: There is no line 143 in the document.
    at getLine (app/node_modules/codemirror/lib/codemirror.js:401:34)
    at bidiSimplify (app/node_modules/codemirror/lib/codemirror.js:7543:1)
    at extendTo (app/node_modules/codemirror/lib/codemirror.js:7486:1)
    at extend (app/node_modules/codemirror/lib/codemirror.js:7504:1)
    at CodeMirror.<anonymous> (app/node_modules/codemirror/lib/codemirror.js:7530:13)
    at HTMLDocument.<anonymous> (app/node_modules/codemirror/lib/codemirror.js:3905:12)
```

Looking at the stack above it's clear that this is fully contained within CodeMirror and there isn't anything we can do to catch this exception, in other words the problem is with CodeMirror but we unearthed the behavior with the recent diff rendering improvements I made in #6707 when we started using `swapDoc` instead of `setValue`. See the upstream issue for the nitty gritty.

As a workaround until this is resolved upstream we're going to attempt to cancel any selection prior to swapping the document. Unfortunately for us there's no documented API for doing so which means we're going to have to dig into some undocumented state. I believe I've taken the necessary precautions to ensure that this won't cause us any problems even after this is fixed upstream and we are (believe it or) seeing a number of crashes because of this which makes me confident that this is a reasonable approach vs waiting for an upstream release.

## Release notes

Notes: [Fixed] Desktop could crash when focused by clicking inside of the diff of a file which had been changed since the last time Desktop was focused.

### DIFF
--- a/app/src/ui/diff/code-mirror-host.tsx
+++ b/app/src/ui/diff/code-mirror-host.tsx
@@ -65,6 +65,28 @@ interface ICodeMirrorHostProps {
 }
 
 /**
+ * Attempts to cancel an active mouse selection in the
+ * given editor by accessing undocumented APIs. This is likely
+ * to break in the future.
+ */
+function cancelActiveSelection(cm: CodeMirror.Editor) {
+  if (cm.state && cm.state.selectingText instanceof Function) {
+    try {
+      // Simulate a mouseup event which will cause CodeMirror
+      // to abort its currently active selection. If no selection
+      // is active the selectingText property will not be a function
+      // so we won't end up here.
+      cm.state.selectingText(new CustomEvent('fake-event'))
+    } catch (err) {
+      // If we end up here it's likely because CodeMirror has changed
+      // its internal API.
+      // See https://github.com/codemirror/CodeMirror/issues/5821
+      log.info('Unable to cancel CodeMirror selection', err)
+    }
+  }
+}
+
+/**
  * A component hosting a CodeMirror instance
  */
 export class CodeMirrorHost extends React.Component<ICodeMirrorHostProps, {}> {
@@ -84,6 +106,7 @@ export class CodeMirrorHost extends React.Component<ICodeMirrorHostProps, {}> {
     if (typeof value === 'string') {
       cm.setValue(value)
     } else {
+      cancelActiveSelection(cm)
       cm.swapDoc(value)
     }
   }


tierninho:
@niik Some clarifying questions for testing this:

> Now switch to whatever editor you prefer and delete all lines in the file

Do I have to save the changes in my editor?

> Switch back to desktop *by clicking, and holding down, the left mouse button anywhere inside the diff`

Any idea how to left-click on a mac trackpad?

niik:
> Do I have to save the changes in my editor?

Yup, you need to change the file on disk.

> Any idea how to left-click on a mac trackpad?

Just click like normal with one finger, left click is the default while right clicking is done by clicking with two fingers on the trackpad.

Here's a video that should help explain the scenario

(null) 2019-03-22 10_17_47

tierninho:
Thanks for the update. 

Couldn't test this as saving the file in the editor immediately autoupdates the diff in Desktop. 🤔

PR #7163, opened 2019-03-26
Detach commit statuses from pull requests

niik:
## Overview

**References #6982**

## Description

This is the second step in a series of PRs to address how the app retrieves, stores, and presents pull requests (see [previous work](https://github.com/desktop/desktop/pull/7027)).

This PR removes the direct relationship between pull requests and commit statuses. Prior to this change all pull requests in the app had a commit status attached to them and when we loaded the list of PRs we used the SHA provided in the PR details from the API to look up the commit status. This meant, among other things, that the commit status would not refresh until we refreshed the list of PRs (so that we got the latest SHA). With this decoupling we are now asking the API for the latest commit status by querying the named ref (i.e `refs/pulls/123/head`) instead of providing an explicit SHA such that we're always getting the commit status for the most recent commit of any particular PR.

This decoupling also means that we no longer have to wait for the commit statuses to load in order to show the list of open PRs, this significantly cuts down on the time it takes to display the PR list from a cold cache.

The approach I've taken here is not one that we've used a whole lot in the past but I think the advantages it provides are more than sufficient to merit it. I've introduced a new store; the `CommitStatusStore`. Through this store consumers are able to subscribe to a particular ref (can be a named ref or a SHA) and receive updates for that ref until they choose to unsubscribe. This means that the `CIStatus` component is able to subscribe directly to the store when it's mounted (and unsubscribe when it's unmounted) meaning that we'll now load statuses on-demand as the component appears on screen (with overrun/underrun provided by our list component meaning that it'll load a few statuses before they're scrolled to. This works really well as we're not relying on commit statuses for any application logic, only presentation. Should that change the store can be expanded to provide an awaitable resource to get the latest status for any given ref.

The store will run up to 5 concurrent fetches of status from the API (compared to the previous implementation which did one status at a time serially) which greatly improves performance and minimizes effects of any single status fetch taking a long time to get. The concurrency management is handled by an external library, [p-limit](https://www.npmjs.com/package/p-limit) but in addition to that the store will re-use handle multiple subscriptions to the same ref and ensure that only one fetch is run for each. Additionally the store handles unsubscriptions from a ref during refresh such that switching a repository will stop refreshing refs that are no longer subscribed to. A ref will be fetched _at most_ once a minute but there's a three-minute interval timer that runs while the app is focused and the store will also refresh when the app regains focus.

Statuses are cached in memory (I've removed the database table that was used when statuses were tied to PRs) in a least recently used cache ([quick-lru](https://www.npmjs.com/package/quick-lru)) to prevent unbounded memory growth.

## PR loading screen

Prior to this PR we had a facade "skeleton" loading screen for PRs. That pattern only existed in this particular view and rather than investing time making that compatible with the new store I've switched over to a loading view more in line with what all our other loading lists do (such as the cloneable repository list).

## Notes for testing / backwards incompatibility

This PR deletes a table from the pullRequestsDatabase making it incompatible with prior versions of GitHub Desktop. As such, after running this branch you need to manually delete the PR database to have it recreated again before you can use PRs in development or other branches. Do this by opening DevTools, clicking on the Applications tab, locating the `PullRequestsDatabase` under the `IndexedDB` group in the sidebar and click `Delete database`

image

## Release notes

Notes: [Improved] Pull requests load faster and PR build status updates automatically.


tierninho:
Tested and overall functions well. Super fast and responsive.

- spotted this, which seemed to happen organically
image
- Is this part of this PR? I was wondering if any testing needs to be done around this (if it's new)? `skipping fast-forward for all branches as there are 21 local branches - this will run again when there are less than 20 local branches tracking remotes`

niik:
> Tested and overall functions well. Super fast and responsive.

💖

> * spotted this, which seemed to happen organically

While not related to this PR that's an interesting error and I've opened https://github.com/desktop/desktop/pull/7169 with more context about that.

That error is actually thrown by `fetchCommit` which is used to retrieve user information for a particular commit in order to display the correct name and avatar.

> * Is this part of this PR? I was wondering if any testing needs to be done around this (if it's new)?

It's not new nor part of this PR. It was introduced in https://github.com/desktop/desktop/pull/4447.

shiftkey:
I wanted to write up this interesting case I stumbled upon - just wanted to flag this in case it's intentional.

Here's what I did:

 - walked through the steps to create screenshots for #7170 
 - closed laptop to suspend machine without exiting Desktop
 - relocated to different location (no internet)
 - unlocked machine and switched back to Desktop

I was hoping that the cached values would be shown here, but it looks like - because the network requests failed - the previous values are cleared from the list:



There's mention in the PR about an LRU cache in-memory rather than using IndexedDB, which is fine, but I'll dig into this as part of the PR to see if I can find the specific change related to this behaviour...

app/src/lib/stores/commit-status-store.ts
```diff
@@ -0,0 +1,313 @@
...
+    const { endpoint, owner, name, ref } = subscription
+    const account = this.accounts.find(a => a.endpoint === endpoint)
+
+    if (account === undefined) {
+      return
+    }
+
+    const status = await API.fromAccount(account)
+      .fetchCombinedRefStatus(owner, name, ref)
+      .catch(err => null)
```
shiftkey: I think this is the case I was referring to in this comment https://github.com/desktop/desktop/pull/7163#issuecomment-477201342 - if this errors we don't look for an existing value in the cache - we just update it to whatever we get here. Is it worth keeping the old value around for situations like the network being intermittent/unavailable?

app/src/lib/stores/commit-status-store.ts
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/**
+ * Creates a cache key for a particular ref in a specific repository.
+ *
+ * Remarks: The cache key is currently the same as the canonical API status
+ *          URI but that has no bearing on the functionality, it does, however
+ *          help with debugging.
+ *
+ * @param ep    The repository endpoint (for example https://api.github.com for
```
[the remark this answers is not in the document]
niik: Yeah, this was me trying to shave a few characters to work around prettier's line wrapping. I'll revert it 😢

app/src/lib/stores/commit-status-store.ts
```diff
@@ -0,0 +1,313 @@
...
+  private readonly queue = new Set<string>()
+
+  /**
+   * A concurrency limiter which ensures that we only run `MaxConcurrentFetches`
+   * API requests simultaneously.
+   */
+  private readonly limit = pLimit(MaxConcurrentFetches)
+
+  public constructor(accountsStore: AccountsStore) {
+    accountsStore.getAll().then(accounts => (this.accounts = accounts))
```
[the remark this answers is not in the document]
niik: We can. Again; this is another case where prettier just goes nuts with unnecessary (IMO) wrapping if we use `{}`. I don't feel very strongly about this TBH, we'll always create a continuation promise regardless and whether that returns void or accounts seems inconsequential but I'll extract the update function into a field and reuse it to minimize the noise and we'll have some .

app/src/lib/stores/commit-status-store.ts
niik: > Is it worth keeping the old value around for situations like the network being intermittent/unavailable?

Yes, I think this is a great shout, we definitely shouldn't be throwing the old value away (at least not initially).

app/src/lib/stores/commit-status-store.ts
```diff
@@ -0,0 +1,336 @@
...
+    } catch (err) {
+      // Okay, so we failed retrieving the status for one reason or another.
+      // That's a bummer, but we still need to put something in the cache
+      // or else we'll consider this subscription eligible for refresh
+      // from here on until we succeed in fetching. By putting a blank
+      // cache entry (or potentially reusing the last entry) in and not
+      // notifying subscribers we ensure they keep their current status
+      // if they have one and that we attempt to fetch it again on the same
+      // schedule as the others.
+      log.debug(`Failed retrieving commit status for ref ${ref}`, err)
```
shiftkey: 🤔 because we removed the error handling inside `fetchCombinedRefStatus` so it will bubble up we also lost the context of which `owner/repo` associated with the failing request. Is that worth tracking to make supporting these changes better? This is also only visible in the Dev Tools console, and not logged to disk, which I think is fine.

niik: It's only visible in devtools [_while you're running in dev mode_](https://github.com/desktop/desktop/blob/d5a938749512b5086c853a9c5640c0ef7992efb8/app/src/main-process/log.ts#L54-L56) if I'm understanding things correctly and in that case I'm assuming that the ref is enough context to understand what's going on but I guess there's no harm in including owner & name either.

PR #7188, opened 2019-03-31
Refactor DialogHeader in Dialog component - Release notes

### LINKED ISSUE #7093: cleanup custom DialogHeader inside Release Notes
This is some cleanup to avoid the workarounds that we've used in some dialogs to support complex headers. #6032 talked about being able to provide support for passing in a JSX element rather than a plain string, which means we can clean up this custom element:

https://github.com/desktop/desktop/blob/b5529325ae629a52b02cb3e07699f1ccbcfb2cfa/app/src/ui/release-notes/release-notes-dialog.tsx#L158-L175

**Version:** development aka b5529325ae629a52b02cb3e07699f1ccbcfb2cfa

**Testing:**

To see this dialog for yourself, build the app in development mode and run it up:

```
$ yarn
$ yarn build:dev
$ yarn start
```

With it running, find the `Help` -> `Show Popup` -> `Release notes` menu item:

Screen Shot 2019-03-15 at 10 59 15 AM

This will then show the dialog - make a note of how things look to ensure that the cleanup in here doesn't affect the styles.

Screen Shot 2019-03-15 at 10 59 27 AM

With that change in place, check out the contents of `render()` inside `app/src/ui/release-notes/release-notes-dialog.tsx`. We should be able to remove the `<DialogHeader>` element and only use props on `<Dialog>` now to achieve the functionality:

```ts
    return (
      <Dialog
        id="release-notes"
        title={header} // TODO
      >
        <DialogContent>{contents}</DialogContent>
```

One thing to check while you're in here is that the close icon is positioned in the top-right (like in the original screenshot) - it looks like having a tall header now causes the default dialog styles to change:

Screen Shot 2019-03-15 at 11 08 43 AM

These are the header-specific styles for this dialog that you might need to tweak:

https://github.com/desktop/desktop/blob/b5529325ae629a52b02cb3e07699f1ccbcfb2cfa/app/styles/ui/dialogs/_release-notes.scss#L11-L42

But I think you'll need to look at the base styles for the dialog to get this working correctly again.

alejandronanez:
## Overview

**Closes #7093**

## Description

Add a new validation in `renderHeader`, check if `this.props.title` is a string.

If it is a string, it should render the default `DialogHeader` component.

If it is not a string (assume it is a React Element), render such element.

## Release notes

Notes:


alejandronanez:
Hey @niik!
First of all, thanks for the thoughtful review, I **really** appreciate it!
I'll be more than happy to keep working on this during this week, your comments make sense but I'll probably have a couple of questions once I get to implement them!

**Thanks** again!

alejandronanez:
Heads up, I'm back from vacations and will be able to work on this in the next couple of days!

Thanks for your patience!

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -154,26 +133,29 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
...
-            />
-            <div className="title">
-              <p className="version">Version {release.latestVersion}</p>
-              <p className="date">{release.datePublished}</p>
-            </div>
-            <img
-              className="release-note-graphic-right"
-              src={ReleaseNoteHeaderRightUri}
-            />
-            {this.renderCloseButton()}
```
alejandronanez: This is gone for good now!

app/styles/ui/dialogs/_release-notes.scss
```diff
@@ -55,6 +68,11 @@
     }
   }
 
+  a.close {
+    align-self: flex-start;
+    z-index: 1;
```
alejandronanez: This was needed b/c the close button was not clickable.

alejandronanez:
Hey @niik, this is ready for another review!

Thanks for your patience with this issue 🙏

billygriffin:
@alejandronanez Congrats on getting your first PR merged into GitHub Desktop! It should go out in our next beta release, and then hopefully to production in a few weeks. Thanks so much! ❤️ 🎉

PR #7197, opened 2019-04-01
enable rule to warn about Buffer constructor usage

shiftkey:
## Overview

As part of testing the rebase flow I kept noticing this warning from within the Dev Tools:

>(node:63731) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

Screen Shot 2019-04-01 at 2 26 55 PM

This was a warning from our upgrade to Electron 3 (that now uses Node 10) about some API changes in Node core.

## Description

Some of the [`Buffer` constructor APIs](https://nodejs.org/docs/latest-v10.x/api/buffer.html) have been deprecated since Node 10.0 and will likely be removed at a later date:

The reasons why have been outlined in https://github.com/nodejs/node/issues/4660 and a more in-depth writeup in https://github.com/ChALkeR/notes/blob/master/Lets-fix-Buffer-API.md.

It's not something that is keeping me awake at night currently, but seeing it on every launch while I test locally in dev tools has annoyed me enough to address it.

I've enabled the [`no-buffer-constructor`](https://eslint.org/docs/rules/no-buffer-constructor) eslint to find anywhere we use this API in the codebase, and I plan to clean these up the violations by swapping them to the recommend APIs.

## Release notes

Notes: no-notes

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -70,6 +70,7 @@ rules:
   strict:
     - error
     - global
+  no-buffer-constructor: error
 
   ###########
   # SPECIAL #
--- a/app/src/ui/diff/syntax-highlighting/index.ts
+++ b/app/src/ui/diff/syntax-highlighting/index.ts
@@ -45,7 +45,7 @@ async function getOldFileContent(
     file.status.kind === AppFileStatusKind.New ||
     file.status.kind === AppFileStatusKind.Untracked
   ) {
-    return new Buffer(0)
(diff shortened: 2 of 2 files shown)


shiftkey:
I plan to merge this today once the build goes green, to ensure this goes out with the next beta this week.

PR #7198, opened 2019-04-01
add general versioning guidelines for dependencies

outofambit:
This PR lays out some guidelines for being more conservative around updating critical dependencies of desktop.

## Questions

* Should anything be added or removed from this list?
* Do the version lags make sense for each dependency?


docs/contributing/upgrading-dependencies.md
```diff
@@ -3,6 +3,22 @@
...
 as some need to be handled differently to others
 
+## General Guidelines
+
+In the interest of stability and caution we tend to stay a version (or more) behind on our most critical dependencies. We vary whether this is a major, minor, or patch version depending on the development practices of the dependency's project.
+
+| Dependency  | Versions Behind Latest |
+| --- | --- |
+| electron | >= 1 major |
+| electron-builder | >= 1 major |
```
shiftkey: This project hasn't seen a major update in a while

- we're on `20.28.4` which was published `Sep 5 2018`
- latest is `20.39.0` which was published `Feb 26 2019`

```suggestion
| electron-builder | >= 1 minor |
```

docs/contributing/upgrading-dependencies.md
```diff
@@ -3,6 +3,22 @@
...
+| electron-packager | >= 1 major |
+| electron-winstaller | >= 1 major |
```
shiftkey: I'm also not sure this one bumps it's major version often enough to constrain it like this:

 - we're on `2.5.2` which was tagged `Jan 26, 2017`
 - latest is `3.0.4` which was published `March 2 2019`

```suggestion
| electron-winstaller | >= 1 minor |
```

docs/contributing/upgrading-dependencies.md
```diff
@@ -3,6 +3,22 @@
...
 
+## General Guidelines
+
+In the interest of stability and caution we tend to stay a version (or more) behind on our most critical dependencies. We vary whether this is a major, minor, or patch version depending on the development practices of the dependency's project.
+
+| Dependency  | Versions Behind Latest |
+| --- | --- |
+| electron | >= 1 major |
+| electron-builder | >= 1 major |
+| electron-packager | >= 1 major |
```
shiftkey: 👍 this one revs enough the major version is incremented to indicate breaking changes

shiftkey:
Dropping `ready-for-review` pending further discussion and iterating on the docs

outofambit:
@shiftkey are there parallel conversations about docs i’m missing? i think this PR should still be up for comment/review by others in @desktop/engineering

shiftkey:
@outofambit I put it aside because I hadn't heard anything about the PR (either from others or follow up to my earlier comments) so it wasn't clear what the current state of this was

outofambit:
@shiftkey ah got it. my understanding is that this is an ongoing conversation that wasn't high priority during the last release cycle, so perhaps we can bring more focus to it soon. going to mark ready for review again since there's nothing to do on the author side and its really just a matter of time and priority right now.

shiftkey:
> since there's nothing to do on the author side

I left some suggested edits because, while all projects use semver, not all of them are bumping the major semver enough for us to be a major version behind. Thoughts on those?

PR #7200, opened 2019-04-01
shorten recent repos list to three

outofambit:
3️⃣ 

## Timing

I'd like to have this go out in the next beta.


tierninho:
Looks good but managed to reduce the list down to 2️⃣as I was up to my old tricks 🐰 🥕 

Screen Shot 2019-04-01 at 2 51 45 PM

outofambit:
thanks @tierninho ! can you file another issue for that? i think that bug was present before this PR.

PR #7167, opened 2019-03-26
preview the number of commits that will be rebased

shiftkey:
## Overview

**Closes #6959**

## Description

Similar to the Merge Branch flow, the Rebase Flow now shows the number of commits that will be moved as part of the operation:

#6960 is tracking detecting conflicts at this stage, so for now this PR is focused on just the happy path and it's testing.

 - [x] use the component in the "Choose Branch" stage to show the status of the rebase
 - [x] pass information from APIs added in #7175 into the rebase flow state, so that commit messages can be shown without relying on the underling Git operation
 - [x] tidy up CSS to reduce excessive whitespace
 - [x] tidy up history to make changes easier to follow
 - [x] rebase on top of #7175 once merged, and tidy up history
 - [x] show "up to date" message when rebase does not contain any commits to rebase

Some additional polish related to this area while I'm in here testing:

 - [x] extract `MergeStatusHeader` into a generalized UI component `ActionStatusIcon` (lots of internals are hard-coded to `merge-status`
 - [x] render Emoji in the progress view if the commit message contains any `:emoji:`

## Release notes

Notes: no-notes

PR #7203, opened 2019-04-02
lift state out of rebase flow components to live in repository state

### LINKED ISSUE #7139: filter text disabled when current branch selected in rebase dialog
If you select the current branch in the rebase model, the modal filter will lock up temporarily until you switch branches. 

For consistency with the merge conflict modal, we should allow the branch to be selected, but keep the `rebase` button disabled. 

The filter should always remain intact and functional.

Repro:
- Select `rebase from current branch` option in the menu
- Modal opens, now select the current branch in the branch list, or type it in the filter.

image

24c697c144, mac

shiftkey:
## Overview

This PR helps to clean up a bunch of component state and uses `Dispatcher` and `AppStore` to drive the rebase flow. This helps to tighten up some of the transitions between states in the rebase flow, and does away with most of the `shouldComponentUpdate` work that was handling the transitions while I was getting the flow working.

I've also included the fix for #7139 because it got sufficiently annoying while testing to have the filter box disabled while trying to find a branch (the filter would end up choosing the current branch, which would disable the form, and I'd have to manually find and choose the branch)

I also noticed that the second paragraph in the Abort Rebase dialog could be better.

Previously, you'd see this:

With this PR, it's now this:

Fixes #7139 

## Description

These things are moved into `IRepositoryState.rebaseState`, which required moving more functionality into `Dispatcher` and `AppStore`

  - [x]  `step: RebaseFlowState` - the state of the flow to show to the user
  - [x] `progress: RebaseProgressSummary` - the state of the Git repository which shows through as progress
  - [x] `rebasePreview: RebasePreview | null`  - the result of the user choosing the rebase
  - [x] `userHasResolvedConflicts: boolean` - whether the rebase

Other tidy up related to these changes:

 - [x] cleanup tracing log messages
 - [x] review names of things and see if things can be clarified
 - [x] docs on all the things
 - [x] review state changes and see if there's anything that could be better organized
 - [x] review `RebaseFlow` and see if there's any other cleanup possible
 - [x] ensure teardown of rebase flow clear relevant app state 
 - [x] ensure transitions in and out of progress are smooth (not suddenly disappearing)
 - ~~[ ] find double-counting rebase progress case (e.g. `9 out of 8 commits rebased`)~~ punting this for now, counting is hard

## Release notes

Notes:

### DIFF
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -239,7 +239,7 @@ export async function getStatus(
     branchAheadBehind,
     exists: true,
     mergeHeadFound,
-    rebaseInternalState: rebaseInternalState,
+    rebaseInternalState,
     workingDirectory,
   }
 }
@@ -373,7 +373,9 @@ async function getConflictDetails(
   try {
     if (mergeHeadFound) {
       return await getMergeConflictDetails(repository)
-    } else if (rebaseInternalState !== null) {
+    }
+    
+    if (rebaseInternalState !== null) {
       return await getRebaseConflictDetails(repository)
     }
   } catch (error) {
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -25,9 +25,9 @@ import { IAheadBehind } from '../../models/branch'
 import { fatalError } from '../../lib/fatal-error'
 import { isMergeHeadSet } from './merge'
 import { getBinaryPaths } from './diff'
-import { getRebaseContext } from './rebase'
+import { getRebaseInternalState } from './rebase'
 import { enablePullWithRebase } from '../feature-flag'
-import { RebaseContext } from '../../models/rebase'
+import { RebaseInternalState } from '../../models/rebase'
 
 /**
  * V8 has a limit on the size of string it can create (~256MB), and unless we want to
@@ -60,7 +60,7 @@ export interface IStatusResult {
   readonly mergeHeadFound: boolean
 
   /** details about the rebase operation, if found */
-  readonly rebaseContext: RebaseContext | null
+  readonly rebaseContext: RebaseInternalState | null
 
   /** the absolute path to the repository's working directory */
   readonly workingDirectory: WorkingDirectoryStatus
@@ -199,7 +199,7 @@ export async function getStatus(
   let conflictDetails: ConflictFilesDetails
 
   const mergeHeadFound = await isMergeHeadSet(repository)
-  const rebaseContext = await getRebaseContext(repository)
+  const rebaseContext = await getRebaseInternalState(repository)
 
   if (enablePullWithRebase()) {
     conflictDetails = await getConflictDetails(
@@ -368,7 +368,7 @@ async function getRebaseConflictDetails(repository: Repository) {
 async function getConflictDetails(
   repository: Repository,
   mergeHeadFound: boolean,
-  rebaseContext: RebaseContext | null
+  rebaseContext: RebaseInternalState | null
 ): Promise<ConflictFilesDetails> {
   try {
     if (mergeHeadFound) {
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -374,7 +374,7 @@ async function getConflictDetails(
     if (mergeHeadFound) {
       return await getMergeConflictDetails(repository)
     }
-    
+
     if (rebaseInternalState !== null) {
       return await getRebaseConflictDetails(repository)
     }
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -60,7 +60,7 @@ export interface IStatusResult {
   readonly mergeHeadFound: boolean
 
   /** details about the rebase operation, if found */
-  readonly rebaseContext: RebaseInternalState | null
+  readonly rebaseInternalState: RebaseInternalState | null
 
   /** the absolute path to the repository's working directory */
   readonly workingDirectory: WorkingDirectoryStatus
@@ -199,13 +199,13 @@ export async function getStatus(
   let conflictDetails: ConflictFilesDetails
 
   const mergeHeadFound = await isMergeHeadSet(repository)
-  const rebaseContext = await getRebaseInternalState(repository)
+  const rebaseInternalState = await getRebaseInternalState(repository)
 
   if (enablePullWithRebase()) {
     conflictDetails = await getConflictDetails(
       repository,
       mergeHeadFound,
-      rebaseContext
+      rebaseInternalState
     )
   } else {
     conflictDetails = await getConflictDetails(repository, mergeHeadFound, null)
@@ -239,7 +239,7 @@ export async function getStatus(
     branchAheadBehind,
     exists: true,
     mergeHeadFound,
-    rebaseContext,
+    rebaseInternalState: rebaseInternalState,
     workingDirectory,
   }
 }
@@ -363,17 +363,17 @@ async function getRebaseConflictDetails(repository: Repository) {
  *
  * @param repository to get details from
  * @param mergeHeadFound whether a merge conflict has been detected
- * @param rebaseContext details about the current rebase operation (if found)
+ * @param rebaseInternalState details about the current rebase operation (if found)
  */
 async function getConflictDetails(
   repository: Repository,
   mergeHeadFound: boolean,
-  rebaseContext: RebaseInternalState | null
+  rebaseInternalState: RebaseInternalState | null
 ): Promise<ConflictFilesDetails> {
   try {
     if (mergeHeadFound) {
       return await getMergeConflictDetails(repository)
-    } else if (rebaseContext !== null) {
+    } else if (rebaseInternalState !== null) {
       return await getRebaseConflictDetails(repository)
     }
   } catch (error) {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -207,12 +207,7 @@ import {
 import { Banner, BannerType } from '../../models/banner'
 import { isDarkModeEnabled } from '../../ui/lib/dark-theme'
 import { ComputedAction } from '../../models/computed-action'
-import {
-  RebaseFlowStep,
-  RebaseStep,
-  ShowConflictsStep,
-  ConfirmAbortStep,
-} from '../../models/rebase-flow-step'
+import { RebaseFlowStep, RebaseStep } from '../../models/rebase-flow-step'
 import { RebasePreview } from '../../models/rebase'
 
 /**
@@ -1740,22 +1735,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
 
     if (step.kind === RebaseStep.ShowConflicts) {
-      const updatedStep: ShowConflictsStep = {
-        kind: RebaseStep.ShowConflicts,
-        conflictState,
-      }
-
       this.repositoryStateCache.updateRebaseState(repository, () => ({
-        step: updatedStep,
+        step: { ...step, conflictState },
       }))
     } else if (step.kind === RebaseStep.ConfirmAbort) {
-      const updatedStep: ConfirmAbortStep = {
-        kind: RebaseStep.ConfirmAbort,
-        conflictState,
-      }
-
       this.repositoryStateCache.updateRebaseState(repository, () => ({
-        step: updatedStep,
+        step: { ...step, conflictState },
       }))
     }
   }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -178,7 +178,7 @@ import { getWindowState, WindowState } from '../window-state'
 import { TypedBaseStore } from './base-store'
 import { AheadBehindUpdater } from './helpers/ahead-behind-updater'
 import { MergeResult } from '../../models/merge'
-import { promiseWithMinimumTimeout } from '../promise'
+import { promiseWithMinimumTimeout, timeout } from '../promise'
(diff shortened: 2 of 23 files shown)


shiftkey:
I've marked this one as `ready-for-review` despite #7167 still being under review because this became the chance to avoid stateful components in the rebase flow and the undesirable side effects, as well as cleanup and clarify a bunch of APIs from previous PRs.

I'd love for this to make it into the next beta, even though I worry about exerting pressure on reviewers to hit this deadline. Let me know your thoughts @desktop/engineering.

app/src/lib/git/status.ts
```diff
@@ -239,7 +239,7 @@ export async function getStatus(
     branchAheadBehind,
     exists: true,
     mergeHeadFound,
-    rebaseContext,
+    rebaseInternalState: rebaseInternalState,
```
iAmWillShepherd:
```suggestion
    rebaseInternalState,
```

app/src/lib/git/status.ts
```diff
@@ -363,17 +363,17 @@ async function getRebaseConflictDetails(repository: Repository) {
...
   repository: Repository,
   mergeHeadFound: boolean,
-  rebaseContext: RebaseContext | null
+  rebaseInternalState: RebaseInternalState | null
 ): Promise<ConflictFilesDetails> {
   try {
     if (mergeHeadFound) {
       return await getMergeConflictDetails(repository)
-    } else if (rebaseContext !== null) {
+    } else if (rebaseInternalState !== null) {
```
iAmWillShepherd:
```suggestion
    }
    
    if (rebaseInternalState !== null) {
```

iAmWillShepherd: No need for the `else` since we return above.

app/src/lib/stores/app-store.ts
```diff
@@ -3417,19 +3426,150 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return this._refreshRepository(repository)
   }
 
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public _setRebaseProgressFromState = async (repository: Repository) => {
```
iAmWillShepherd: **Non-blocking**
Are there any benefits to defining the function in this way in this class? Only asking because it's the only function in this class that is defined as a lambda expression.

app/src/lib/stores/app-store.ts
```diff
@@ -3417,19 +3426,150 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        commits,
+      }
+    })
+
+    this.emitUpdate()
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public _setConflictsResolved(repository: Repository) {
+    // an update is not emitted here because there is no need
```
iAmWillShepherd: 👍

app/src/lib/stores/app-store.ts
```diff
@@ -3417,19 +3426,150 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    step: RebaseFlowStep
+  ): Promise<void> {
+    this.repositoryStateCache.updateRebaseState(repository, () => ({
+      step,
+    }))
+
+    this.emitUpdate()
+
+    if (step.kind === RebaseStep.ShowProgress && step.rebaseAction !== null) {
+      await timeout(500)
```
iAmWillShepherd: Why the timeout of 500ms? If this isn't an arbitrary value, it may be worthing throwing it in a constant.

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -712,13 +724,39 @@ export class Dispatcher {
     }))
   }
 
+  public setConflictsResolved(repository: Repository) {
+    return this.appStore._setConflictsResolved(repository)
+  }
+
+  public initializeRebaseProgress(
```
iAmWillShepherd: I think these functions can use some docs since they form the API of rebase flow.

app/src/ui/rebase/progress-dialog.tsx
```diff
@@ -30,21 +22,17 @@ export class RebaseProgressDialog extends React.Component<
...
     const {
       rebasedCommitCount,
-      commits,
+      totalCommitCount,
       value,
       currentCommitSummary,
     } = this.props.progress
+
+    // ensure progress always starts from 0
+    const count = rebasedCommitCount === 0 ? 1 : rebasedCommitCount
```
iAmWillShepherd: It's not clear how this variable or the logic relates to the comment above. It seems that it's possible to set this to some non-zero value.

app/src/ui/rebase/rebase-flow.tsx
```diff
@@ -323,55 +112,33 @@ export class RebaseFlow extends React.Component<
     targetBranch: string,
     commits: ReadonlyArray<CommitOneLine>
   ) => {
-    if (this.state.step.kind !== RebaseStep.ChooseBranch) {
-      throw new Error(`Invalid step to start rebase: ${this.state.step.kind}`)
+    if (this.props.step.kind !== RebaseStep.ChooseBranch) {
+      throw new Error(`Invalid step to start rebase: ${this.props.step.kind}`)
```
iAmWillShepherd: Is this an error that would be displayed in the UI? If so, how would the user get out of an invalid step? Also, I'm assuming this is necessary because the user may initialize a rebase from the CLI.

app/src/ui/rebase/rebase-flow.tsx
```diff
@@ -323,55 +112,33 @@ export class RebaseFlow extends React.Component<
...
+      rebaseAction: startRebaseAction,
     })
   }
 
   private onContinueRebase = async () => {
-    if (this.state.step.kind !== RebaseStep.ShowConflicts) {
+    if (this.props.step.kind !== RebaseStep.ShowConflicts) {
       throw new Error(
-        `Invalid step to continue rebase rebase: ${this.state.step.kind}`
+        `Invalid step to continue rebase rebase: ${this.props.step.kind}`
```
iAmWillShepherd:
```suggestion
        `Invalid step to continue rebase: ${this.props.step.kind}`
```

app/src/ui/rebase/rebase-flow.tsx
```diff
@@ -323,55 +112,33 @@ export class RebaseFlow extends React.Component<
...
-      }
+    this.props.dispatcher.setRebaseFlowStep(this.props.repository, {
+      kind: RebaseStep.ShowProgress,
+      rebaseAction: startRebaseAction,
     })
   }
 
   private onContinueRebase = async () => {
-    if (this.state.step.kind !== RebaseStep.ShowConflicts) {
+    if (this.props.step.kind !== RebaseStep.ShowConflicts) {
```
iAmWillShepherd: Does this line imply that we can only continue a rebase if we need to show conflicts? If so, the doc for this function doesn't make that clear.
> Continue with the rebase after the user has resovled all conflicts with
   tracked files in the working directory.

tierninho:
Tooltip over button ✅ 
Fix for filter ✅ 
Dialogue update ✅ 

Noticed that there are two success messages, one when rebasing with conflicts, one when rebasing without conflicts. Is this intended?

Screen Shot 2019-04-03 at 2 56 16 PM
Screen Shot 2019-04-03 at 2 56 01 PM

I vote for the former as its more descriptive. 

771c708360, mac

app/src/lib/stores/app-store.ts
shiftkey: Arrow functions inside classes (at least as part of ES2017 support we're targeting in Desktop) are used to ensure that `this` is correctly bound when the function is invoked. In the most part this is unnecessary for classes, but if you find yourself passing a class function for use elsewhere (for example, into a React component) you should ensure it is bound correctly.

We already have a decent set of lint rules to catch these usages, e.g. https://github.com/desktop/desktop/pull/7129#discussion_r267523660 so I'm fine to clean this up because it's not important.

For more details, here's an example from [this blog post](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) that talks about the tradeoffs:

```ts
class A {
  static color = "red";
  counter = 0;
  
  handleClick = () => {
    this.counter++;
  }
  
  handleLongClick() {
    this.counter++;
  }
}
```

 As part of the transpilation to JS it moves this inside inside the constructor to confirm it's an instance function:

```js
class A {
  constructor() {
    this.counter = 0;

    this.handleClick = () => {
      this.counter++;
    };
  }
  
  handleLongClick() {
    this.counter++;
  }
}
A.color = "red";
```

shiftkey: Tidied up in b422d8d22f91867382b810ad2ab51a713e2f1721

app/src/lib/stores/app-store.ts
shiftkey: This was an unscientific number chosen to allow the UI time to display itself before the next action was performed (which could be very fast). Digging into some more data on this:

> The average reaction time for humans is 0.25 seconds to a visual stimulus, 0.17 for an audio stimulus, and 0.15 seconds for a touch stimulus.
>
> [[source](https://backyardbrains.com/experiments/reactiontime)]

So it's a generous delay that gives users a better chance to see the rebase progress before it switches back to the conflicts UI (if it encounters an issue).

I'm not sure how much this value might change, or get reused in other places, but I'll add a comment in here to explain this area a bit better.

app/src/ui/dispatcher/dispatcher.ts
shiftkey: Oops, I made a start on this last night but missed these. Addressed in 25e82b0ea26ad8f999fa89eb30b05300d06a816a

shiftkey:
> Noticed that there are two success messages, one when rebasing with conflicts, one when rebasing without conflicts. Is this intended?

@tierninho this message should only differ based on how the rebase was started:

 - If you started a rebase outside Desktop before switching to Desktop, we lack the knowledge of the original branch that was chosen (Git converts that into a commit ID when it performs the rebase). In this situation, we show the short message with only one branch after completing the rebase.
 - If you started the rebase in app, the base branch should be stored in application state, which enables us to show the longer message with both branches when the rebase completes.

Can you confirm that is how the cases differ, rather than whether or not conflicts were resolved? If that's not the case I can dig into whether we're not correctly remembering the base branch in some situation...

app/src/ui/rebase/progress-dialog.tsx
shiftkey: Improved in 912623da75bd6dff5e81e4cf672ef035b3e787d7

app/src/ui/rebase/rebase-flow.tsx
shiftkey: These guard checks were present to help ensure the callers were enforcing the right transitions, and were useful while I was building out the flow, but are less valuable now that things are stable.

I starting down the path in this PR to define a pattern for these callbacks, where the first parameter received is the step that is initiating the transition:

```ts
private showRebaseConflictsBanner = (step: ShowConflictsStep) => {
 ...
}
```

This helps to make clear:

 - what state is initiating the callback
 - that data the callback needs is available in the step
 - that invalid states are no longer representable at compile time, avoiding the need for checks like this at runtime

This makes the props inside a component a bit more complex than plain functions:

```tsx
  readonly onContinueRebase: (step: ShowConflictsStep) => void
  readonly onAbortRebase: (step: ShowConflictsStep) => void
  readonly showRebaseConflictsBanner: (step: ShowConflictsStep) => void
```

I've pushed some more commits to this branch to wrap up this work - let me know how this feels.

app/src/ui/rebase/rebase-flow.tsx
shiftkey: To save duplicate comments, I'll point to the other discussion that overlaps with this https://github.com/desktop/desktop/pull/7203#discussion_r272150557

app/src/lib/stores/app-store.ts
shiftkey: I've added more context behind this to e8aace4b90f54ca640af4806811ab40cbfc955af

app/src/ui/rebase/rebase-flow.tsx
iAmWillShepherd: Overall, this feels nice 👍

tierninho:
> If you started the rebase in app, the base branch should be stored in application state, which enables us to show the longer message with both branches when the rebase completes.

Yes, this works as intended, but only if there are no conflicts. I am **unsuccessful** in seeing this message when there are conflicts that need to be resolved. If I resolve a conflict in Atom and return, does this count as "outside Desktop"?

> If you started a rebase outside Desktop before switching to Desktop, we lack the knowledge of the original branch that was chosen (Git converts that into a commit ID when it performs the rebase). In this situation, we show the short message with only one branch after completing the rebase.

I see this message when initiating and resolving a rebase in the cmd line. If rebase initiated in Desktop, and conflicts resolved in editor, then this message shows.

Recap:
- Desktop initiated rebase, no conflicts = detailed message
- Desktop initiated rebase, with conflicts = short message
- Cmd line initiated rebase = short message

shiftkey:
@tierninho thanks for the extra information!

> Yes, this works as intended, but only if there are no conflicts. I am unsuccessful in seeing this message when there are conflicts that need to be resolved.

Interesting, maybe there is something in this flow where the context gets lost? I suspect this existed before this PR, and its plausible we're clobbering the context incorrectly to "forget" the base branch.

> If I resolve a conflict in Atom and return, does this count as "outside Desktop"?

It should not. Only where the rebase is started matters for this flow, as that's the only chance where we know the base branch. I've opened #7220 to track this so it doesn't get lost amid these PRs.

PR #7204, opened 2019-04-02
azure pipelines builds all PRs and only development branch

outofambit:
This explicitly configures our azure pipelines CI to build
* all PRs
* all pushes to `development`.

Extras:
* previous in-progress builds for a PR are cancelled when new commits are pushed
* branch builds for `development` are batched (which I think means if you push multiple commits at once, it only builds the _latest_ commit)

This _should_ eliminate duplicate builds for PR branches (and i've verified that as much as i can without merging this).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,7 +2,10 @@ variables:
   PYTHON: 'python2.7'
 
 trigger:
-  - development
+  batch: true
+  branches:
+    include:
+      - development
 
 pr:
   autoCancel: true
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,11 +2,10 @@ variables:
   PYTHON: 'python2.7'
 
 trigger:
-  pr:
-    autoCancel: false
-  branches:
-    include:
-      - development
+  - development
+
+pr:
+  autoCancel: false
 
 jobs:
   - job: Windows
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@ trigger:
   - development
 
 pr:
-  autoCancel: false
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,6 +1,12 @@
 variables:
   PYTHON: 'python2.7'
 
+trigger:
```
outofambit: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#branches

azure-pipelines.yml
```diff
@@ -1,6 +1,12 @@
...
+  - development
+
+pr:
```
outofambit: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#pull-request-validation

azure-pipelines.yml
```diff
@@ -1,6 +1,15 @@
 variables:
   PYTHON: 'python2.7'
 
+trigger:
+  batch: true
```
outofambit: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#batching-ci-builds

azure-pipelines.yml
```diff
@@ -1,6 +1,15 @@
...
+  branches:
+    include:
+      - development
```
shiftkey: Given we don't use Azure Pipelines for generating builds, I think this list is fine. But something to keep an eye on in the future if we decide to simplify our CI infrastructure.

outofambit: yep, agreed!

PR #7209, opened 2019-04-03
ensure conflicts popup and banner are cleaned up if conflicts no longer exist

### LINKED ISSUE #7046: Auto update banner if rebase aborted via cmd line
If the rebase banner is present in Desktop and the user runs `git rebase --abort` in the cmd line, then we should auto clear the banner upon returning to Desktop. 

Currently, if you click `View Conflicts` after running the cmd, the banner will only then disappear. I noticed we updated the Fetch/Pull/Push button, but not the banner.

Before manual abort:
Screen Shot 2019-03-08 at 2 27 32 PM

After manual abort:
Screen Shot 2019-03-08 at 2 27 55 PM

Have done this, the user is also left in a state where half of the menu options remain greyed out, like so:
image
image

This seems to only happens in this instance.

v1.6.3, mac

**Update**: tested similar action in Merge Conflict, and same thing happens. cc @outofambit

shiftkey:
## Overview

**Closes #7046**

## Description

This PR adds a `AppStore.clearConflictsFlowVisuals()` function to ensure any UI related to conflicts (both merge conflicts and rebase conflicts) goes away if the rebase/merge is aborted outside the app.

It wasn't clear to me from the original report how deep to go with cleaning up state, but we can iterate on this based on further testing and feedback.

## Release notes

Notes: no-notes


tierninho:
LGTM. Ran both `git merge --abort` and `git rebase --abort` in cmd line and the UI reverted back to pre-conflicted stage. ⚡️

app/src/lib/stores/app-store.ts
```diff
@@ -1728,6 +1729,52 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
   }
 
+  /**
+   * Cleanup any related UI related to conflicts if still in use.
+   */
+  private clearConflictsFlowVisuals() {
```
niik: I feel like we could make this whole method a lot cleaner if we added an optional `popupType` and `bannerType` arguments to the `closePopup` and `clearBanner` methods (similar to what we do in `closeFoldout` except not prescriptive). So something like

```ts
public async _closePopup(popupType?: PopupType): Promise<void> {
  const currentPopup = this.currentPopup

  if (currentPopup == null) {
    return
  }

  if (popupType !== undefined && currentPopup.type !== popupType) {
    return
  }
```

That way we should just be able to list all the popupTypes we want to close here and if any are open they'll get closed

```
this._closePopup(PopupType.MergeConflicts)
this._closePopup(PopupType.AbortMerge)
this._closePopup(PopupType.RebaseFlow)

this._clearBanner(PopupType.MergeConflictsFound)
this._clearBanner(BannerType.RebaseConflictsFound)
```

shiftkey: 👍 I'm into it. Thoughts @outofambit?

outofambit: yeah, i'm into it!

outofambit:
@niik no, i love it more!

PR #7210, opened 2019-04-03
[Stashing] Adds ability to stash changes

iAmWillShepherd:
Related issues: #6107 and #6987

This PR addresses the following user story:

> I'm doing work on a branch, and something comes up where I need to temporarily go to a different branch to do something entirely different. I'm forced to either WIP commit my work and undo it when I come back or leave Desktop to stash it temporarily. I want to just be able to switch branches, do whatever I needed to do, and come back with my changes right where I left them.

**Important bits**
* Created a feature flag that enables this in the dev.
* New dialog that gives user an opportunity to stash changes upon checking out a new branch. Note that this dialog will only appear if there are changes in your working directory.
* Adds styling to _Stashed Changes_ item so it looks like the mockup; interaction is wired up to show the `IStashEntry` associated with the branch
* Stash entries are deleted to enforce the one stash per branch constraint. We decided that we'd instead like to modify the stash entry's message, but that is not straight forward since there is no stash function that we can rely on.
* Stash dialog notifies user that they will _overwrite_ their stash entry when there is already an associated stash for the branch.
* Stash diff view, thanks @outofambit for taking care of this piece.


app/src/ui/lib/list/list-row.tsx
```diff
@@ -27,48 +27,59 @@ interface IListRowProps {
   readonly onRef?: (element: HTMLDivElement | null) => void
 
   /** callback to fire when the row receives a mouseover event */
-  readonly onRowMouseOver: (index: number, e: React.MouseEvent<any>) => void
+  readonly onRowMouseOver?: (index: number, e: React.MouseEvent<any>) => void
```
iAmWillShepherd: These fields were made optional to simply creating a `ListRow` outside of the `List` component.

app/src/ui/repository.tsx
```diff
@@ -241,12 +243,51 @@ export class RepositoryView extends React.Component<
...
       const { changesState } = this.props.state
-      const { workingDirectory, selectedFileIDs, diff } = changesState
+      const {
+        workingDirectory,
+        selectedFileIDs,
+        diff,
+        isShowingStashEntry,
+      } = changesState
+
+      if (isShowingStashEntry && selectedFileIDs.length === 0) {
```
iAmWillShepherd: There is no need to place this block behind the feature flag because `currentStashForBranch` uses  the feature flag in its implementation.

app/src/ui/repository.tsx
```diff
@@ -241,12 +244,55 @@ export class RepositoryView extends React.Component<
...
-      const { workingDirectory, selectedFileIDs, diff } = changesState
+      const {
+        workingDirectory,
+        selectedFileIDs,
+        diff,
+        shouldShowStashedChanges,
+      } = changesState
+
+      if (shouldShowStashedChanges && selectedFileIDs.length === 0) {
+        const stashEntry = this.currentStashForBranch()
```
iAmWillShepherd: There is no need to place this block behind the feature flag because `currentStashForBranch` uses  the feature flag in its implementation.

tierninho:
Some things I noticed while testing this:
- If a changed file is present and you tab over to the Stashed Changes tab, clicking `enter` does not work. I expected doing so would allow me to view diffs.
- You cannot tab or key over to Stashed changes consistently, making it not accessible.
- This popped up a lot in the Console:

```
install.ts:23 `git log -g --pretty=%H@%gs refs/stash` exited with an unexpected code: 128.
fatal: ambiguous argument 'refs/stash': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

- > Stash dialog does not notify user that they will overwrite their stash entry when there is already an associated stash for the branch.

When testing this, is this a valid use case?
Stash changes on Branch-A
Switch to Branch-B and add some changes
Switch back to Branch-A and stash changes on A when prompted

**Result**: User has two stashes stacked on top of each other. I can `git stash pop` them both back into Changes list.

- If we are going to add a UI element to `git stash pop`, are we taking care of this scenario where the user tries to pop a stash but the file is in Changes already?

```
➜  abcde git:(master) ✗ git stash pop
error: Your local changes to the following files would be overwritten by merge:
	.gitignore
Please commit your changes or stash them before you merge.
Aborting
```

6d241b0bd7, mac

app/src/ui/lib/list/list-row.tsx
shiftkey: > to simply creating a `ListRow` outside of the `List`

I'd love some extra context on why we're using `ListRow` outside of `List` here and what we gain, as hiding these callbacks stands out as something worth unpacking.

I'd also love some @desktop/design input on whether there's a better way we can do this - could we borrow the stylings from and create a standalone component that just wraps a `div`?

app/src/lib/app-state.ts
```diff
@@ -35,6 +35,7 @@ import { ApplicationTheme } from '../ui/lib/application-theme'
 import { IAccountRepositories } from './stores/api-repositories-store'
 import { ManualConflictResolution } from '../models/manual-conflict-resolution'
 import { Banner } from '../models/banner'
+import { IStashEntry } from './git/stash'
```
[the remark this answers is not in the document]
iAmWillShepherd: Yea, I'm using the same type in UI. I've moved it to models.

app/src/ui/lib/list/list-row.tsx
iAmWillShepherd: The ListRow component gives me a few things for free:
* Styling - the _Stashed Changes_ item in the design looks like a list item
* Interaction - the _Stashed Changes_  item is selectable and comes with hooks to respond to keyboard and mouse events

I'm fine with coming up with a new component in an iteration however. 👍

iAmWillShepherd:
🍎

shiftkey:
I had a go at testing the latest changes (4e1ad40f7ed935dfa56c1f8208b8054e61bd380f) but wasn't able to see the expected UI.

Here's my repro steps:

 - make some changes to the `README` that was already committed in a repository



- change branch, see the new warning



 - Choose `Yes, stash my changes`, and the checkout completes, but there's no indicator about the stash UI:



 - The Dev Tools show an error that looks like the same error @tierninho found earlier, which is probably relevant to the underlying issue



 - the stash seems to have been stored in the repository at least

iAmWillShepherd:
> Some things I noticed while testing this:
> 
> * If a changed file is present and you tab over to the Stashed Changes tab, clicking `enter` does not work. I expected doing so would allow me to view diffs.
> * You cannot tab or key over to Stashed changes consistently, making it not accessible.
> * This popped up a lot in the Console:

I'm aware of this issue. This element does not respond to keyboard input at the time.

> ```
> install.ts:23 `git log -g --pretty=%H@%gs refs/stash` exited with an unexpected code: 128.
> fatal: ambiguous argument 'refs/stash': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> ```

This is expected and doesn't present itself in the UI, so it's not really an issue. 

> * > Stash dialog does not notify user that they will overwrite their stash entry when there is already an associated stash for the branch.
> 
> When testing this, is this a valid use case?
> Stash changes on Branch-A
> Switch to Branch-B and add some changes
> Switch back to Branch-A and stash changes on A when prompted
> 
> **Result**: User has two stashes stacked on top of each other. I can `git stash pop` them both back into Changes list.
> 

See https://github.com/desktop/desktop/pull/7230

> * If we are going to add a UI element to `git stash pop`, are we taking care of this scenario where the user tries to pop a stash but the file is in Changes already?
> 

No, this scenario would result in the no-op if the stash entry being popped is not found in the list of Desktop created stash entries.

https://github.com/desktop/desktop/blob/fc4ac3660c0feedbd822ddc48226e37d819cd5b5/app/src/lib/git/stash.ts#L156-L189

☝️ Relevant lines are 164-173

> ```
> ➜  abcde git:(master) ✗ git stash pop
> error: Your local changes to the following files would be overwritten by merge:
> 	.gitignore
> Please commit your changes or stash them before you merge.
> Aborting
> ```
> [6d241b0](https://github.com/desktop/desktop/commit/6d241b0bd72ff815299e57ebc1133341756c49f9), mac

Fixed in https://github.com/desktop/desktop/pull/7219/commits/ac1a72fc6523399e634627bd2352c01996d8d75a

iAmWillShepherd:
https://github.com/desktop/desktop/issues/7254 created to track errors logged to console when `refs/stash` file is missing.

app/src/ui/lib/list/list-row.tsx
shiftkey: > Styling - the Stashed Changes item in the design looks like a list item

I see you're already adding some custom styles to `_changes-list` as part of this PR - what if we tried using a mixin here that allowed for reuse of the `list-item` styles in different areas? Would that get you most of the way to something usable here?

https://github.com/desktop/desktop/blob/4e1ad40f7ed935dfa56c1f8208b8054e61bd380f/app/styles/ui/_list.scss#L76-L100

> Interaction - the Stashed Changes item is selectable and comes with hooks to respond to keyboard and mouse events

This is fair, but I think you could crimp the selection logic and click events from inside `ListRow` without needing to change the behaviour of `ListRow`. Looking at the setup, there's a lot of hard-coded values because of what `ListRow` is expecting:

```jsx
<ListRow
          rowCount={1}                             // necessary?
          rowIndex={0}                              // necessary?
          selectable={true}                       // necessary?
          selected={this.props.isShowingStashEntry} 
          onRowClick={this.onStashEntryClicked}
          tabIndex={0}                               // necessary?
          style={StashListRowStyle}        // pass in rowHeight directly?
          className="stash-entry-row" // could live inside new component
        >
```

I'd love to see an exploration of alternatives here because I think you could get away with something simpler than `ListRow` for your use case.

app/src/lib/stores/app-store.ts
```diff
@@ -4671,7 +4724,66 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     this.emitUpdate()
   }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _createStash(repository: Repository, branchName: string) {
+    const { branchesState } = this.repositoryStateCache.get(repository)
+
+    if (branchesState === undefined) {
```
shiftkey: :fire: `branchesState` is never undefined

app/src/lib/stores/git-store.ts
```diff
@@ -967,6 +971,32 @@ export class GitStore extends BaseStore {
...
+  public async loadStashEntries(): Promise<void> {
+    if (!enableStashing()) {
+      return
+    }
+
+    const map = new Map<string, IStashEntry>()
+    const entries = await getDesktopStashEntries(this.repository)
+
+    for (const entry of entries) {
+      if (!map.has(entry.branchName)) {
```
shiftkey: I'm curious about this check - can you add a comment here about why we skip over subsequent entries for a given branch after the first one?

iAmWillShepherd: I think this is skipping over all entries that aren't already a part of the map. I think @outofambit can give a better answer than I though.

outofambit: hmm it looks like this skips subsequent entries. i _think_ this is because we only want the first matching one we find in the stash list. i can add a comment for this.

outofambit: added in 273dd01a995c5a1c92c2f5ace9711944829b1901

app/src/lib/git/stash.ts
```diff
@@ -155,21 +147,45 @@ export async function dropDesktopStashEntry(
...
+      ['stash', 'pop', `${stashToPop.name}`],
+      repository.path,
+      'popStashEntry',
+      {
+        expectedErrors,
+      }
+    )
+  } catch (err) {
+    if (err instanceof GitError) {
+      log.error(err.message)
```
shiftkey: :art: no need to look for `GitError` here - I'd just log out whatever you find:

```ts
} catch (err) { 
  log.error('[popStashEntry] unable to pop a stash successfully', err)
}
```

app/src/models/uncommitted-changes-strategy.ts
```diff
@@ -0,0 +1,5 @@
+export enum UncommittedChangesStrategy {
+  askForConfirmation,
```
shiftkey: :art: not worth holding this up, but it'd be great to make these keys `SentenceCase` and also string enums to be consistent with other parts of the app.

app/src/lib/stores/app-store.ts
```diff
@@ -2523,6 +2581,19 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    if (shouldPopStash) {
+      const stash = await getLastDesktopStashEntryForBranch(
+        repository,
+        foundBranch.name
+      )
+
+      if (stash !== null) {
+        await this._popStashEntry(repository, stash)
+      } else {
+        log.warn('no stash found that matches')
```
shiftkey: :art:  is this important enough to make the user aware of it? Or is a better message here enough for the developers to look over when troubleshooting:

```suggestion
        log.warn(`[AppStore._checkoutBranch] no stash found that matches ${foundBranch.name}`)
```

app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -0,0 +1,125 @@
+/* tslint:disable:button-group-order */
```
shiftkey: :art: can we add some context here about why we've needed to do this?

app/src/ui/lib/list/list-row.tsx
shiftkey: Resolving this because #7263 will address this (and is currently under review)

shiftkey:
Repro steps:

 - make change to `README.md` in repository
 - switch branches, `Yes, stash changes...`
 - return to branch, make a different change to `README.md`
 - choose stash and try to restore change

Expected behaviour:
 - feedback that I can't restore this stash currently

Actual behaviour:
 - no feedback

Log:

```
error: [ui] `git stash pop stash@{0}` exited with an unexpected code: 128.
error: Your local changes to the following files would be overwritten by merge:
    README.md
Please commit your changes or stash them before you merge.
Aborting

error: [ui] error: Your local changes to the following files would be overwritten by merge:
    README.md
Please commit your changes or stash them before you merge.
Aborting
```

shiftkey:
Repro steps:

 - make change to README.md in repository
 - switch branches, `Yes, stash changes...`
 - return to branch, make changes to any file
 - select stash to see contents
 - switch away from Desktop to another app
 - focus on Desktop again

Expected behaviour:
 - stash has focus, no changed files selected

Actual behaviour:
 - stash has focus, but first changed file is selected and diff is shown

Before switching away:



After returning to Desktop:

iAmWillShepherd:
Diff shows old state. Seems to be the previous stash entry.

**Repro steps**
1. Stash changes on branch
1. Switch back to starting branch
1. Click _Stashed Changes_


☝️I reloaded the app to force a state update, after doing so, the contents of the diff is empty

iAmWillShepherd:
> Expected behaviour:
> 
> * stash has focus, no changed files selected

I'd expect the first changed file in the stash diff to be selected. Thoughts?

cc/ @desktop/product @shiftkey

shiftkey:
> I'd expect the first changed file in the stash diff to be selected. Thoughts?

Agree. To be clear, my "no changed files selected" was referring to the Changed files list above the stash, not the files in the stash (words, ermagherd).

tierninho:
First round of testing. Wanted to get it out there as it contains bugs.

- UI elements are functional ✅ 
- Tested `git stash`, `git stash pop`, `git stash drop` cmds ✅ 
- Selecting a PR updates branch, and stash dialogue still pops up ✅ 

Issues:
- ❌ If a stashed file is empty there if no diff visible. We need a placeholder.
Screen Shot 2019-04-11 at 12 04 02 PM

- ❌ Open a file in Atom. Now stash all changes and return to that branch. Go back to Atom and update the file. Restore stash results in below, and no error message shown in UI.
```
sdfsdf.md already exists, no checkout
sdfsdfddd.md already exists, no checkout
Could not restore untracked files from stash entry
```

Suggestions/Questions: 
- `Clear` button should be dynamic, as is `clear` for one file and `clear all` for multiple files, or add checkboxes if we can support popping single files. Might want an "are you sure?" dialogue or an undo?
- The Stashed Changes dialogue should show file count. I know all files get stashed, but I had a few of them unchecked and I could easily see a novice user wondering what happened. Show file count in the toggle box under Changes too.
- Which is it: Stash Branch or Stash Changes?
Screen Shot 2019-04-11 at 12 22 24 PM
Screen Shot 2019-04-11 at 12 23 41 PM

686fbbc04f, mac

billygriffin:
Thanks for the quick feedback @tierninho! Answers to your questions below:

> Clear button should be dynamic, as is clear for one file and clear all for multiple files, or add checkboxes if we can support popping single files. Might want an "are you sure?" dialogue or an undo?

Initially I think "Clear" is intended to be to clear the entire stash. I'm a bit skeptical of allowing people to mess around with things while inside the stashed state, but definitely open to thinking about it more if we get feedback on that front.

> The Stashed Changes dialogue should show file count. I know all files get stashed, but I had a few of them unchecked and I could easily see a novice user wondering what happened. Show file count in the toggle box under Changes too.

In the initial option to stash or bring changes with you? That could be a nice follow up, but definitely doesn't need to be in for beta imo.

> Which is it: Stash Branch or Stash Changes?

It's "Switch branch" vs. "Stash changes". Is that language the confusion?

tierninho:
Round 2 of testing:

- 🐛Pane slider does work here, even though mouse indicator arrow changes
Screen Shot 2019-04-11 at 2 21 08 PM

- 🐛Binary files do not render in diff. Try stashing a pic.
- 🐛Stashed changes disappear when branch renamed. (You can still use `git stash pop` to bring back or revert the name change.)
- Getting instances of a cached file but it is hard to repro. If I reload the application, it goes away and correct stash remains. I think it may have to do with using the UI and also the cmd line... eventually Desktop get confused.
Screen Shot 2019-04-11 at 2 21 08 PM
- Did a restore of a stash and caught this. Not sure if they are associated.
Screen Shot 2019-04-11 at 2 51 39 PM

Questions:
- is `git stash push` relevant in this feature right now? Or any other stash cmds for the matter?

app/src/ui/repository.tsx
```diff
@@ -241,12 +247,71 @@ export class RepositoryView extends React.Component<
...
+              imageDiffType={this.props.imageDiffType}
+              fileListWidth={this.props.stashedFilesWidth}
+              externalEditorLabel={this.props.externalEditorLabel}
+              onOpenInExternalEditor={this.props.onOpenInExternalEditor}
+              repository={this.props.repository}
+              dispatcher={this.props.dispatcher}
+            />
+          )
+        } else if (this.props.state.branchesState.tip.kind === TipState.Valid) {
+          this.props.dispatcher.loadStashedFiles(
```
niik: This is a dangerous pattern, we should not be calling the dispatcher from render methods. The Repository component could get re-rendered any number of times before the `loadStashedFiles` method is is able to complete and emit an update, causing us to shell out to Git for each render.

I think we can find a much better pattern for this by letting the appStore take care of loading stashed files when the right conditions exists for doing so but in the meantime I'm going to open a PR to ensure we're never loading stashed files more than once for any stash entry.

app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+    return (
+      <Dialog
+        id="stash-changes"
+        title={__DARWIN__ ? 'Switch Branch' : 'Switch branch'}
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+        loading={isStashingChanges}
+        disabled={isStashingChanges}
+      >
+        <DialogContent className="dialog-content">
```
niik: See https://github.com/desktop/desktop/pull/7230#discussion_r274384004 and https://github.com/desktop/desktop/pull/7230#discussion_r274475468. 

```suggestion
        <DialogContent>
```

app/src/lib/git/stash.ts
```diff
@@ -39,17 +30,16 @@ export async function getDesktopStashEntries(
...
       'getStashEntries'
     )
   } catch (err) {
     if (err instanceof GitError) {
-      if (
-        !expectedErrorMessages.some(
-          message => err.message.indexOf(message) !== -1
-        )
-      ) {
+      const shouldThrow = !expectedErrorMessages.some(
```
niik: As I said in https://github.com/desktop/desktop/issues/7254#issuecomment-482063241 I believe this code could be removed entirely if we were just to add `128` to the list of expected exit codes. I will open a PR to illustrate what I mean.

app/src/ui/repository.tsx
niik: I've opened https://github.com/desktop/desktop/pull/7267 as a stop-gap measure for this but I think we should address this more holistically before we ship to production.

app/src/lib/git/stash.ts
niik: Pr opened: https://github.com/desktop/desktop/pull/7268

iAmWillShepherd: Merged

app/src/ui/repository.tsx
iAmWillShepherd: Merged.

iAmWillShepherd: https://github.com/desktop/desktop/issues/7270 created to track this.

app/src/lib/stores/git-store.ts
```diff
@@ -967,6 +972,50 @@ export class GitStore extends BaseStore {
...
+  /** A map key on the canonical ref name of GitHub Desktop created stash entries for the repository */
+  public get stashEntries() {
+    return this._stashEntries
+  }
+
+  /**
+   * Updates the latest stash entry with a list of files that it changes
+   */
+  public async loadStashedFiles(stashEntry: IStashEntry) {
+    const files = await getChangedFiles(this.repository, stashEntry.stashSha)
```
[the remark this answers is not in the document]
iAmWillShepherd: Tracking this in https://github.com/desktop/desktop/issues/7272.

app/src/ui/stashing/stash-diff-viewer.tsx
iAmWillShepherd: Tracked in #7273.

app/src/lib/git/stash.ts
```diff
@@ -106,24 +77,27 @@ export async function getLastDesktopStashEntryForBranch(
...
-export function createDesktopStashMessage(branchName: string, tipSha: string) {
-  return `${DesktopStashEntryMarker}<${branchName}@${tipSha}>`
+export function createDesktopStashMessage(branchName: string) {
+  return `${DesktopStashEntryMarker}<${branchName}>`
 }
 
 /**
  * Stash the working directory changes for the current branch
+ *
+ * @param tipSha is only used to format the message.
```
iAmWillShepherd: We no longer use this in the function signature.

app/src/lib/stores/app-store.ts
```diff
@@ -2512,6 +2530,46 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        branchToCheckout: foundBranch,
+        repository,
+      })
+
+      return repository
+    }
+
+    let shouldPopStash = false
+
+    if (currentBranch !== null) {
```
iAmWillShepherd: Put behind feature flag

app/src/lib/stores/app-store.ts
```diff
@@ -2529,6 +2587,26 @@ export class AppStore extends TypedBaseStore<IAppState> {
       )
     )
 
+    if (shouldPopStash) {
```
iAmWillShepherd: Place comment here saying this cannot be true

app/src/lib/stores/app-store.ts
```diff
@@ -4677,7 +4753,145 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      : null
+
+    await createDesktopStashEntry(repository, branchName)
+
+    if (previousStash === null) {
+      return
+    }
+
+    await dropDesktopStashEntry(repository, previousStash.stashSha)
+    log.warn(
```
iAmWillShepherd: This could be `log.info`

app/src/lib/stores/app-store.ts
```diff
@@ -4677,7 +4753,145 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        stashEntry.stashSha
+      }`
+    )
+
+    await this._refreshRepository(repository)
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public _showStashEntry(repository: Repository) {
+    this.repositoryStateCache.updateChangesState(repository, state => {
```
iAmWillShepherd: Put behind feature flag.

app/src/lib/stores/app-store.ts
```diff
@@ -4677,7 +4753,145 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        selectedFileIDs: [],
+      }
+    })
+
+    this.emitUpdate()
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public _hideStashEntry(repository: Repository) {
+    this.repositoryStateCache.updateChangesState(repository, state => {
```
iAmWillShepherd: Put behind feature flag.

app/src/lib/stores/app-store.ts
```diff
@@ -4677,7 +4753,145 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+    this.emitUpdate()
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _loadStashedFiles(
+    repository: Repository,
+    stashEntry: IStashEntry
+  ) {
+    const gitStore = this.gitStoreCache.get(repository)
```
iAmWillShepherd: Feature flag

app/src/lib/stores/app-store.ts
```diff
@@ -4677,7 +4753,145 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    await gitStore.loadStashedFiles(stashEntry)
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _changeStashedFileSelection(
+    repository: Repository,
+    file: CommittedFileChange
+  ): Promise<void> {
+    this.repositoryStateCache.updateChangesState(repository, () => ({
```
iAmWillShepherd: Feature flag

app/src/lib/stores/git-store.ts
```diff
@@ -967,6 +972,83 @@ export class GitStore extends BaseStore {
...
+
+  /** A map key on the canonical ref name of GitHub Desktop created stash entries for the repository */
+  public get stashEntries() {
+    return this._stashEntries
+  }
+
+  /**
+   * Updates the latest stash entry with a list of files that it changes
+   */
+  public async loadStashedFiles(stashEntry: IStashEntry) {
```
iAmWillShepherd: Feature flag

app/src/models/stash-entry.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/** Whether file changes for a stash entry are loaded or not */
+export enum StashedChangesLoadStates {
+  NotLoaded = 'NotLoaded',
+  Loading = 'Loading',
+  Loaded = 'Loaded',
+}
+
+export type StashedFileChanges =
```
iAmWillShepherd: 📝

app/src/models/uncommitted-changes-strategy.ts
iAmWillShepherd: Tracked in #7279.

app/src/ui/changes/changes-list.tsx
```diff
@@ -463,6 +483,41 @@ export class ChangesList extends React.Component<
...
+  private onStashEntryClicked = () => {
+    if (this.props.isShowingStashEntry) {
+      this.props.dispatcher.hideStashEntry(this.props.repository)
+    } else {
+      this.props.dispatcher.showStashEntry(this.props.repository)
+    }
+  }
+
+  private renderStashedChanges() {
+    if (this.props.stashEntry === null) {
```
iAmWillShepherd: Feature flag

app/src/ui/stash-changes/overwrite-stashed-changes-dialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+  readonly repository: Repository
+  readonly branchToCheckout: Branch
+  readonly onDismissed: () => void
+}
+
+interface IOverwriteStashState {
+  readonly isCheckingOutBranch: boolean
+}
+
+export class OverwriteStash extends React.Component<
```
iAmWillShepherd: 📝

app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+  readonly hasAssociatedStash: boolean
+  readonly onDismissed: () => void
+}
+
+interface ISwitchBranchState {
+  readonly isStashingChanges: boolean
+  readonly selectedStashAction: StashAction
+}
+
+export class StashAndSwitchBranch extends React.Component<
```
iAmWillShepherd: 📝

app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { join } from 'path'
+import { Diff } from '../diff'
+import { IDiff, ImageDiffType } from '../../models/diff'
+import { Resizable } from '../resizable'
+import { Button } from '../lib/button'
+import { ButtonGroup } from '../lib/button-group'
+
+interface IStashDiffViewerProps {
+  /** The stash in question. */
+  readonly stashEntry: IStashEntry
```
iAmWillShepherd: Breathing room between fields with comments and others.

app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * _(Like viewing a selected commit in history but for a stash)_
+ */
+export class StashDiffViewer extends React.PureComponent<
+  IStashDiffViewerProps
+> {
+  private onSelectedFileChanged = (file: FileChange) =>
+    this.props.dispatcher.changeStashedFileSelection(
+      this.props.repository,
+      file as CommittedFileChange
```
iAmWillShepherd: Do type check to ensure `file` is of type `CommittedFileChange`

app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -0,0 +1,126 @@
...
+  private onReset = () => this.props.dispatcher.resetStashedFilesWidth()
+
+  public render() {
+    const files =
+      this.props.stashEntry.files.kind === StashedChangesLoadStates.Loaded
+        ? this.props.stashEntry.files.files
+        : new Array<FileChange>()
+
+    const diffComponent =
+      this.props.selectedStashedFile && this.props.stashedFileDiff ? (
```
iAmWillShepherd: Explicit check.

app/src/ui/changes/changes-list.tsx
niik: Seems to have been resolved on line 495.

app/src/lib/stores/git-store.ts
niik: Resolved on line 1016

app/src/lib/stores/app-store.ts
niik: Resolved on line 4880

app/src/lib/stores/app-store.ts
niik: Resolved on line 4868

app/src/lib/stores/app-store.ts
niik: Resolved on line 4846

app/src/lib/stores/app-store.ts
niik: Resolved on line 4832

app/src/lib/stores/app-store.ts
niik: Why was this marked as resolved? Tentatively marking as unresolved.

app/src/lib/stores/app-store.ts
niik: Resolved in e8aad48504410a5954c524f2b443a77a707f1eda

app/src/ui/repository.tsx
niik: No action needed due to explanation above. Resolving.

app/src/ui/stashing/stash-diff-viewer.tsx
niik: Tracked in https://github.com/desktop/desktop/issues/7296

niik:
Approved given that all new functionality in this PR is behind a feature flag. I've done a quick audit of all comments made to ensure that they're tracked separately in issues.

app/src/lib/stores/app-store.ts
outofambit: fixed in https://github.com/desktop/desktop/pull/7210/commits/0ef307e095ea0d1633dc9af2da6003bb46f2e202

niik: Resolved in https://github.com/desktop/desktop/pull/7210/commits/0ef307e095ea0d1633dc9af2da6003bb46f2e202

PR #7219, opened 2019-04-04
[Stashing] Adds workflow that allows users to take changes with them to new branch

### LINKED ISSUE #7248: [Stashing] Merge conflicts resulting from applying a stash entry does not update state properly
The UI isn't updating correctly when taking changes to the new branch causes a merge conflict and results in:
* the stash dialog not being dismissed even thought the branch was successfully checked out
* files in the _Changes_ view aren't updated (switching to _History_ and back to _Changes_ forces the UI to update itself using the new state and we see that applying worked just fine).
* The branch checkout progress indicator never goes away.

**Repro steps**
1. Branch from `dev
1. Commit change on `feature` 
1. Commit change on `dev` 
1. Make a change on `dev`
1. Switch to `master`
1. Select option to take changes to `master`
1. Click _Switch Branch_

bring_changes_to_new_branch_fails_when_conflicts_are_encountered

iAmWillShepherd:
⚠️ Review #7210 first ⚠️ 

Fixes #7248 - ac1a72f and 7fbce46
Related issues: #6107, #6987 

This PR addresses the following user story:

> I realized after I started doing work that I actually wanted to be doing that work on a different branch. I want to be able to switch branches and retain the uncommitted changes I've made without having to leave GitHub Desktop.

**Important bits**

* Changed `git stash apply <stash>` to `git stash pop <stash>`  after learning that I can specify a stash to pop. `git stash drop <stash>` still exists because we still have to support dropping an arbitrary stash.
* Added a new enum to control what stash related tasks occur before/after a checkout.
* Moved logic for each workflow from the component to `AppStore` to reduce simplify the component and make the relationship between stashing and checking out a branch explicit.

 
**What's not here**

* More robust error-handling. I am letting errors bubble up to the UI during the development so any issues are immediately visible.
* UI for viewing diff of stash


app/src/lib/stores/app-store.ts
```diff
@@ -2514,6 +2525,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return repository
     }
 
+    let shouldPopStash = false
+
+    if (
+      currentBranch !== null &&
```
outofambit: this is a (non-blocking) nitpick, but could we pull this conditional into an enclosing `if` statement so its easier to follow that either `uncommittedChangesStrategy` will be acted on if `currentBranch !== null`

app/src/lib/stores/app-store.ts
```diff
@@ -4688,6 +4717,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      branch.name
+    )
+
+    if (stash === null) {
+      log.warn('no stash found that matches')
+      return
+    }
+
+    await popStashEntry(repository, stash.stashSha)
+    log.warn(`Popped stash with commit id ${stash.stashSha}`)
```
outofambit: why is this `warn` and not `info` or `debug`? doesn't seem like its something we should be even mildly worried about.

app/src/ui/stash-and-switch-branch-dialog.tsx
```diff
@@ -88,24 +89,28 @@ export class StashAndSwitchBranch extends React.Component<
...
   }
 
   private onSubmit = async () => {
-    const {
-      repository,
-      currentBranch,
-      branchToCheckout,
-      dispathcer,
-    } = this.props
+    const { repository, branchToCheckout, dispatcher } = this.props
```
outofambit: i like this syntax :)

app/src/models/uncommitted-changes-strategy.ts
```diff
@@ -0,0 +1,5 @@
+export enum UncommittedChangesStrategy {
+  askForConfirmation,
```
outofambit: is this the state we use when the popup is first opened and we don't have an answer from the user yet?

app/src/lib/stores/app-store.ts
iAmWillShepherd: Yea, I'll change to `info`. My main concern is that what ever is logged is actually persisted to disk.

app/src/models/uncommitted-changes-strategy.ts
iAmWillShepherd: This enum controls what to do about uncommitted changes:
* `askForConfirmation` - informs what to do about changes
* `stashOnCurrentBranch`/`moveToNewBranch` - informs what action to take with changes

iAmWillShepherd:
🍎

app/src/lib/stores/app-store.ts
```diff
@@ -450,7 +455,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         this.emitUpdate()
       })
       .catch(err => {
-        log.warn(`Unexpected issue when trying to read emoji into memory`, err)
+        log.info(`Unexpected issue when trying to read emoji into memory`, err)
```
outofambit: i don't think this is the `warn` we wanted to change

outofambit:
```suggestion
        log.warn(`Unexpected issue when trying to read emoji into memory`, err)
```

app/src/lib/stores/app-store.ts
```diff
@@ -4711,6 +4741,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      branch.name
+    )
+
+    if (stash === null) {
+      log.warn('no stash found that matches')
+      return
+    }
+
+    await popStashEntry(repository, stash.stashSha)
+    log.warn(`Popped stash with commit id ${stash.stashSha}`)
```
outofambit: this is still `warn`?
```suggestion
    log.info(`Popped stash with commit id ${stash.stashSha}`)
```

PR #7230, opened 2019-04-05
[Stashing] Notify user that an existing stash will be overwritten

### LINKED ISSUE #7250: [Stashing] When taking changes to new branch, existing stash entry is deleted
Taking changes to a branch with an existing stash results in the existing stash being deleted.

**Repro steps**
1. Create stash on `dev`
1. Switch to `feature` and make a change
1. Switch back to `dev`
1. Select the option to take changes to new branch and click _Switch Branch_

**See (and hear) it in action**
[Dropping all the stashes](https://youtu.be/5j2lGOxkfqA)

iAmWillShepherd:
⚠️ Please review #7219 first ⚠️ 

Fixes #7250 - e655e8a
Related issues: #6107 and #6987

This PR updates the stash changes workflow to ask for confirmation when creating a stash would result in the previous stash being overwritten.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -100,7 +100,8 @@ import {
   initializeRebaseFlowForConflictedRepository,
 } from '../lib/rebase'
 import { BannerType } from '../models/banner'
-import { StashAndSwitchBranch } from './stash-and-switch-branch-dialog'
+import { StashAndSwitchBranch } from './stash-changes/stash-and-switch-branch-dialog'
+import { OverwriteStashedChanges } from './stash-changes/overwrite-stashed-changes-dialog'
 
 const MinuteInMilliseconds = 1000 * 60
 const HourInMilliseconds = MinuteInMilliseconds * 60
@@ -1655,6 +1656,17 @@ export class App extends React.Component<IAppProps, IAppState> {
             currentBranch={currentBranch}
             branchToCheckout={branchToCheckout}
             hasAssociatedStash={hasAssociatedStash}
+            onDismissed={this.onPopupDismissed}
+          />
+        )
+      }
+      case PopupType.ConfirmOverwriteStash: {
+        const { repository, branchToCheckout: branchToCheckout } = popup
+        return (
+          <OverwriteStashedChanges
+            dispatcher={this.props.dispatcher}
+            repository={repository}
+            branchToCheckout={branchToCheckout}
             onDismissed={this.onPopupDismissed}
           />
         )
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -101,7 +101,7 @@ import {
 } from '../lib/rebase'
 import { BannerType } from '../models/banner'
 import { StashAndSwitchBranch } from './stash-changes/stash-and-switch-branch-dialog'
-import { OverwriteStashedChanges } from './stash-changes/overwrite-stashed-changes-dialog'
+import { OverwriteStash } from './stash-changes/overwrite-stashed-changes-dialog'
 
 const MinuteInMilliseconds = 1000 * 60
 const HourInMilliseconds = MinuteInMilliseconds * 60
@@ -1665,7 +1665,7 @@ export class App extends React.Component<IAppProps, IAppState> {
       case PopupType.ConfirmOverwriteStash: {
         const { repository, branchToCheckout: branchToCheckout } = popup
         return (
-          <OverwriteStashedChanges
+          <OverwriteStash
             dispatcher={this.props.dispatcher}
             repository={repository}
             branchToCheckout={branchToCheckout}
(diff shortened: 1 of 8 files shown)


iAmWillShepherd:
I updated the warning to only show when stashing on the current branch since that's the only time we should overwrite an existing stash entry.

app/src/ui/stash-changes/overwrite-stashed-changes-dialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export class OverwriteStashedChanges extends React.Component<
+  IOverwriteStashedChanges,
+  {}
+> {
+  public render() {
+    const title = __DARWIN__
+      ? 'Are You Sure You Want To Overwrite Your Current Stash?'
+      : 'Are you sure you want to overwrite your current stash?'
```
niik: This is a little meaty for a title. Could we incorporate this message into the dialog contents and make the title be something like `Overwrite Stash?`

app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
```diff
@@ -50,7 +57,10 @@ export class StashAndSwitchBranch extends React.Component<
         loading={isStashingChanges}
         disabled={isStashingChanges}
       >
-        <DialogContent>{this.renderStashActions()}</DialogContent>
+        <DialogContent className="content-wrapper">
```
niik: The `DialogContent` component already adds has a class, `dialog-content`. Could we drop this and use the standard `.dialog-content` selector in `_stash-changes.scss` instead?

app/src/ui/app.tsx
```diff
@@ -1645,12 +1646,29 @@ export class App extends React.Component<IAppProps, IAppState> {
         }
 
         const currentBranch = tip.branch
+        const hasAssociatedStash =
```
niik: I think we could simplify this to 

```ts
const hasAssociatedStash = state.stashEntries.has(currentBranch.name)
```

app/src/ui/stash-changes/overwrite-stashed-changes-dialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+          </ButtonGroup>
+        </DialogFooter>
+      </Dialog>
+    )
+  }
+
+  private onClick = async () => {
+    const { dispatcher, repository, branchToCheckout, onDismissed } = this.props
+
+    await dispatcher.checkoutBranch(
```
niik: Can we set an `isCheckingOutBranch` state field before and after this and then use that to set the `loading` and `disabled` flags on `<dialog>` like we do in `stash-and-switch-branch-dialog`. Otherwise users will be able to click the `Continue` button more than once if checkout takes a while. Either that or we immediately close the dialog without awaiting the checkout.

app/src/ui/stash-changes/overwrite-stashed-changes-dialog.tsx
iAmWillShepherd: I decided to get rid of this content entirely since putting it in the dialog content area would be redundant.

iAmWillShepherd:
🍎

app/src/ui/stash-changes/overwrite-stashed-changes-dialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <Dialog
+        id="overwrite-stash"
+        type="warning"
+        title={title}
+        loading={this.state.isCheckingOutBranch}
+        disabled={this.state.isCheckingOutBranch}
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+      >
+        <DialogContent className="dialog-content">
```
niik: Sorry, I probably wasn't clear here. What I meant was that this class is already added by the component itself. Simply removing the className attribute should do what you want.

```suggestion
        <DialogContent>
```

iAmWillShepherd:
🍎

app/styles/ui/dialogs/_stash-changes.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../mixins';
+
+dialog#stash-changes {
+  width: 450px;
```
niik: Don't you still want to color the oction in the dialog?

```suggestion
  width: 450px;
  .dialog-content {
    .octicon {
      color: var(--dialog-warning-color);
      margin-right: var(--spacing);
    }
  }
```

iAmWillShepherd: Hmm, I thought that was also baked in.

niik:
@desktop/engineering I have to head out but given the time sensitive nature, the fact that I've reviewed this and everyone else will be up and around for almost a full day before I get back I think it's fine if this is merged by someone else before I'm back tomorrow.

iAmWillShepherd:
Merging now since this is going into a feature branch and having these changes in one branch makes @desktop/quality-assurance lives easier.

PR #7236, opened 2019-04-07
Add Resx Highlighter Support

### LINKED ISSUE #7235: Add Highlighter Support for .resx files
Resx is a .Net file extension type. It is essentially XML.
https://docs.microsoft.com/en-us/dotnet/framework/resources/creating-resource-files-for-desktop-apps#resources-in-resx-files

say25:
## Overview

**Closes #7235**

## Description

- Add Resx Highlighter Support

## Release notes

Notes: Add Resx Highlighter Support

### DIFF
--- a/app/src/highlighter/index.ts
+++ b/app/src/highlighter/index.ts
@@ -113,6 +113,7 @@ const extensionModes: ReadonlyArray<IModeDefinition> = [
       '.vcxproj': 'text/xml',
       '.vbproj': 'text/xml',
       '.svg': 'text/xml',
+      '.resx': 'text/xml',
     },
   },
   {


shiftkey:
@say25 are you able to also submit a PR to https://github.com/desktop/highlighter-tests with this file format?

say25:
@shiftkey I can. But it seemed like we were lacking examples for svg, .csproj and other xml like things. I can add those as well if desired? Would you prefer us to make a folder called xml with all the extensions we support?

shiftkey:
@say25 I'm mostly worried about capturing new formats as we add them in, to ensure they aren't regressed later by things in Codemirror or Desktop. XML-ish formats might not be worth doing because they're basically one file, but I'd prefer to be consistent here than start to make exceptions.

@desktop/engineering any thoughts?

PR #7242, opened 2019-04-08
warn before rebase if operation requires a force push

### LINKED ISSUE #6963: [rebase] warn the user when the rebase will result in a force push
Extracted from #6551

This occurs after the user chooses to start the rebase, but before the actual rebase is executed:

shiftkey:
## Overview

**Closes #6963**

## Description

- [x] wireup UI to show when rebase
- [x] plug in user preference to warn about force push
- [x] implement logic to if remote branch has commits that will be rebased
- [x] use `Branch` shape rather than `string` as this contains the relevant tracking branch
- [x] update tests to use `Branch` rather than `string`

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/rebase/warn-force-push-dialog.tsx
+++ b/app/src/ui/rebase/warn-force-push-dialog.tsx
@@ -70,7 +70,7 @@ export class WarnForcePushDialog extends React.Component<
         </DialogContent>
         <DialogFooter>
           <ButtonGroup>
-            <Button type="submit">Continue rebase</Button>
+            <Button type="submit">Begin rebase</Button>
             <Button onClick={this.props.onDismissed}>Cancel</Button>
           </ButtonGroup>
         </DialogFooter>
--- a/app/src/ui/rebase/warn-force-push-dialog.tsx
+++ b/app/src/ui/rebase/warn-force-push-dialog.tsx
@@ -51,8 +51,10 @@ export class WarnForcePushDialog extends React.Component<
             onto <strong>{baseBranch.name}</strong>?
           </p>
           <p>
-            This rebase will rewrite the history on the remote branch and
-            require a force push to complete the rebase flow.
+            In order to push this branch to your remote branch, you will need to
+            force push the branch after the rebase is complete. That will alter
+            the history on the remote and potentially cause problems for others
+            collaborating on this branch.
           </p>
           <div>
             <Checkbox
--- a/app/src/ui/rebase/warn-force-push-dialog.tsx
+++ b/app/src/ui/rebase/warn-force-push-dialog.tsx
@@ -41,7 +41,7 @@ export class WarnForcePushDialog extends React.Component<
       <Dialog
         title="This rebase will require a force push"
         onDismissed={this.props.onDismissed}
-        onSubmit={this.onContinueRebase}
+        onSubmit={this.onBeginRebase}
         disableClickDismissalAlways={true}
         type="warning"
       >
@@ -86,7 +86,7 @@ export class WarnForcePushDialog extends React.Component<
     this.setState({ askForConfirmationOnForcePush: value })
   }
 
-  private onContinueRebase = async () => {
+  private onBeginRebase = async () => {
     this.props.dispatcher.setConfirmForcePushSetting(
       this.state.askForConfirmationOnForcePush
     )
--- a/app/src/ui/rebase/warn-force-push-dialog.tsx
+++ b/app/src/ui/rebase/warn-force-push-dialog.tsx
@@ -9,6 +9,7 @@ import { Checkbox, CheckboxValue } from '../lib/checkbox'
 
 import { Dispatcher } from '../dispatcher'
 import { DialogFooter, DialogContent, Dialog } from '../dialog'
+import { Ref } from '../lib/ref'
 
 interface IWarnForcePushProps {
   readonly dispatcher: Dispatcher
@@ -37,9 +38,13 @@ export class WarnForcePushDialog extends React.Component<
   public render() {
     const { baseBranch, targetBranch } = this.props.step
 
+    const title = __DARWIN__
+      ? 'Rebase Will Require Force Push'
+      : 'Rebase will require force push'
+
     return (
       <Dialog
-        title="This rebase will require a force push"
+        title={title}
         onDismissed={this.props.onDismissed}
         onSubmit={this.onBeginRebase}
         disableClickDismissalAlways={true}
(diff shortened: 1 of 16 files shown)


billygriffin:
I know we've been iterating on the verbiage here throughout the design process, but wanted to offer alternate verbiage since it's been awhile since I've looked at it.

Since this is a pre-emptive warning and everything currently exists locally, I think maybe the force pushing and the rewriting history on the remote should be reversed in the second sentence. A possible alternative:

> Are you sure you want to rebase [branch] onto [branch]?

> In order to push this branch to your remote branch, you will need to force push the branch after the rebase is complete. That will alter the history on the remote and potentially cause problems for others collaborating on this branch.

I don't love my proposed language. But overall, my goals here are to 1. make it more human-readable, and 2. help folks understand why it might be a problem and why we're warning them.

shiftkey:
@billygriffin yep, I'm totally 👍 to iterating on the messaging here. I'll pull in your suggestion and see what we can come up with while we're working on this.

shiftkey:
@billygriffin I've updated the PR description with the latest copy - let me know if you have any additional ideas or suggestions.

tierninho:
Two things I noticed while testing this: 

- The "dont show me this" toggle does not appear to be working in sync with the Pref option. 
- When you navigate away from Desktop and return to the app, the dialogue will disappear. (When I returned I clicked this area just in case)

image

Otherwise, all looks good.

shiftkey:
> When you navigate away from Desktop and return to the app, the dialogue will disappear. (When I returned I clicked this area just in case)

Fixed in 6b851fba11749b16cf8d8cb3eec0a1a9ddad5d40 - turns out we had an eager refactor in #7209 that meant we were always trying to cleanup the rebase flow, even when it was in a state that no conflicts were occurring.

shiftkey:
> The "dont show me this" toggle does not appear to be working in sync with the Pref option.

Fixed in 78bc057e9e46d73f1c44712aaab1dd309444d9ec - not sure what I meant to put in there but clearly it wasn't working.

tierninho:
Thanks. Both issues fixed with the latest commits 👍 

One clarifying question though: from a UX perspective if I check the "dont show me again" option but decide to hit the `Cancel` button, should my preference be saved? My expectation would be yes, but we are treating the `Continue rebase` button as a save button.

billygriffin:
Re: @tierninho's comment above

> One clarifying question though: from a UX perspective if I check the "dont show me again" option but decide to hit the Cancel button, should my preference be saved? My expectation would be yes, but we are treating the Continue rebase button as a save button.

I think my expectation is that if you click "Cancel" that the checkbox state for the warning would *not* be saved because you haven't really affirmatively committed to anything yet.

Separately, this dialog appears *before* you begin a rebase, right? Should we consider having that button say "Begin rebase" instead of "Continue rebase"?

tierninho:
>I think my expectation is that if you click "Cancel" that the checkbox state for the warning would not be saved because you haven't really affirmatively committed to anything yet.

My thought was: "I don't want to see this message and I understand the risks" is a separate from "I don't want to proceed force pushing this one instance". No strong feelings on this as it could go either way. 🤷‍♂️ 

> Separately, this dialog appears before you begin a rebase, right? Should we consider having that button say "Begin rebase" instead of "Continue rebase"?

Hard to say, as the user has already selected the menu option "Rebase current branch...", chosen a branch, and clicked "Rebase X onto X" button. I have the same stance as above. We can use "Beginuing" button 😆

shiftkey:
@billygriffin @tierninho 

> I think my expectation is that if you click "Cancel" that the checkbox state for the warning would not be saved because you haven't really affirmatively committed to anything yet.

Just to reiterate - this is only updated when you "continue" - cancelling will not trigger updating this:

https://github.com/desktop/desktop/blob/14160c205e3610e9625925db7ad356fa2feb79a4/app/src/ui/rebase/warn-force-push-dialog.tsx#L89-L92

> Should we consider having that button say "Begin rebase" instead of "Continue rebase"?

We don't say "Begin" or "Start" when choosing the branch:



I still prefer changing this to "Begin rebase" because the corresponding action is not "Abort rebase" (which suggests the rebase is underway) but "Cancel" which suggests it's easier to bail out now.

app/src/ui/rebase/warn-force-push-dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+                  ? CheckboxValue.Off
+                  : CheckboxValue.On
+              }
+              onChange={this.onAskForConfirmationOnForcePushChanged}
+            />
+          </div>
+        </DialogContent>
+        <DialogFooter>
+          <ButtonGroup>
+            <Button type="submit">Continue rebase</Button>
```
billygriffin: Suggestion per our discussion in the thread.
```suggestion
            <Button type="submit">Begin rebase</Button>
```

app/src/ui/rebase/warn-force-push-dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  private onAskForConfirmationOnForcePushChanged = (
+    event: React.FormEvent<HTMLInputElement>
+  ) => {
+    const value = !event.currentTarget.checked
+
+    this.setState({ askForConfirmationOnForcePush: value })
+  }
+
+  private onContinueRebase = async () => {
```
billygriffin: I'm not sure if we change "Continue rebase" to "Begin rebase" if this function also needs to change, or if it's ok as is. I'll defer to you @shiftkey

billygriffin:
Thanks @shiftkey! I think we're aligned on both of those things then. I made the suggestion to change "Continue" to "Begin," but wasn't sure if any of the function names should change for clarity.

app/src/ui/rebase/warn-force-push-dialog.tsx
shiftkey: Renamed in 74cb03cc0abdc75ef8fba8f79943a99b783ebf0c

shiftkey:
Added `time-sensitive` on this because I'd like to get this out with the next beta

app/src/ui/rebase/warn-force-push-dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+      askForConfirmationOnForcePush: props.askForConfirmationOnForcePush,
+    }
+  }
+
+  public render() {
+    const { baseBranch, targetBranch } = this.props.step
+
+    return (
+      <Dialog
+        title="This rebase will require a force push"
```
niik: Could we keep the title as short as possible and also deal with platform casing? Perhaps something like

```ts
const title = __DARWIN__
  ? 'Rebase Will Require Force Push'
  : 'Rebase will require force push'
```

shiftkey:
Refreshed the screenshot in the PR description with the latest changes

PR #7243, opened 2019-04-08
show "force push" label after rebase which requires it

### LINKED ISSUE #4976: "Remove" default menu item needs to be aware of user preference

### LINKED ISSUE #7138: "Push" menu item should say "Force Push" when you can only force push
When a `force push origin` is inflight, the pull menu option is available to be clicked. Not sure what the correct resolution is, perhaps disable the menu option or do we allow the user to stash? I defer to the dev team for this as the error messaging is conflicting. 

Otherwise the end result is:
image

Edit (5/1): We're going to keep the menu items enabled for now, but we have #7017 tracking those more specifically as we iterate on this. For this issue to be closed specifically, we need to update the "Push" menu item to "Force Push" when the user is in that state after rebase.

24c697c144, mac

### LINKED ISSUE #7231: Remove `AppStore.getBranchesState`
I discovered this after noticing that `AppStore._createStash` function did not behave as expected. After doing a bit of digging, it turns out that my test repo not being a `GitHubRepository` was the cause. I confirmed with @desktop/engineering that getting `BranchesState` does not depend on `GitHubRepository`. 

**Problem function**
https://github.com/desktop/desktop/blob/92faa65e495a38c2be373fac16f8b4ea74a7bcac/app/src/lib/stores/app-store.ts#L1621-L1628

This function should be removed because of the following reasons:
* `repository` is optional
* `undefined` is returned if the `repository.githubRepository` is null
* All functions that use this function are circumventing explicit state access

Basically this function does not follow the patterns we've defined elsewhere.

shiftkey:
## Overview

This also pulls in the work from #6489 because I believe some changes around `AppStore._selectRepository` now mean we no longer have a `AppStore.refreshRepository` that isn't `await`-ed.

I also want to add the `...` to the `Force Push` label if the user has configured their preferences to warn them before they force push.

**Closes #4976**
**Closes #7138**
**Closes #7231**

Current `development` branch:

This PR:

And the `...` characters are omitted if you have disable the "Confirm..." settings in your preferences:

## Description

- [x] restore changes from #6489 into PR 
- [x] see if we can get this change working with the test suite
- [x] add `pushLabel` that reflects whether the next action is a push or a force push
- [x] tidy up code

## Release notes

Notes: no-notes


app/src/lib/stores/app-store.ts
```diff
@@ -1606,9 +1606,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
       pullRequestLabel: this.getPullRequestLabel(repository),
       shellLabel: `Open in ${this.selectedShell}`,
       defaultBranchName: this.getDefaultBranchName(repository),
+      removeRepoLabel: this.getRemoveRepoLabel(),
     })
   }
 
+  private getRemoveRepoLabel() {
+    const showRemoveRepoDialog = this.confirmRepoRemoval
```
shiftkey: @Daniel-McCarthy I think this was the missing part from the original PR that avoided those test failures - we didn't _need_ to read local storage because the current value should be store in app state. I'm going to do the same for the `Force Push...` label which is also backed by some app state.

Daniel-McCarthy: Ah, that is a great catch! I'm glad to hear you were able to find a solution to get this to work without breaking the tests!

app/src/lib/rebase.ts
```diff
@@ -66,3 +70,30 @@ export function initializeRebaseFlowForConflictedRepository(
 export function formatRebaseValue(value: number) {
   return Math.round(clamp(value, 0, 1) * 100) / 100
 }
+
+/**
+ * Check application state to see whether the action applied to the current
+ * branch should be a force push
+ */
+export function isCurrentBranchForcePush(
```
outofambit: glad to see this extracted into a helper function!

could we also use this helper function here? https://github.com/desktop/desktop/blob/639afe70d2dc271b0d9dd9183e507adae6346f3a/app/src/ui/changes/no-changes.tsx#L315-L318

shiftkey: Sure can!

shiftkey: Addressed in 12ea76d48b4473e06ce7a63d216c9fcf955f8f91

tierninho:
Tested the changes and here are my notes:

- Why is dialogue different between clicking `Force push origin` button and selecting the `Force Push...` menu option?  When selecting `Force push...` menu option I noticed there is no `cancel` button in the dialogue, which could lead to accidental pushing. 

- When selecting the `Force push...` menu option, I spotted this in the logs:
```
install.ts:23 `git -c credential.helper= -c protocol.version=2 push origin as:as --progress` exited with an unexpected code: 1.
To https://github.com/tierninho/PRlist.git
 ! [rejected]        as -> as (non-fast-forward)
error: failed to push some refs to 'https://github.com/tierninho/PRlist.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

(The error was parsed as 10: The repository has been updated since you last pulled. Try pulling before pushing.)
```

- I changed the preference setting to "do not show me force push" option, and the menu item changed to just `Push`. Couldn't get to change to `Force Push` nor back to `Force Push...`. 

- I turned off the force push preference setting, selected the Push menu option and still got the warning dialogue. I hit `X` and then clicked the `Force push origin` button and it push the code.

shiftkey:
> * When selecting `Force push...` menu option I noticed there is no `cancel` button in the dialogue, which could lead to accidental pushing.

What dialog do you see here? I think this is related to what happens when the menu button fires - I haven't touched the underlying implementation, which just emits an event from the main process:

https://github.com/desktop/desktop/blob/development/app/src/main-process/menu/build-default-menu.ts#L241

Which is associated with this function in the renderer:

https://github.com/desktop/desktop/blob/6d241b0bd72ff815299e57ebc1133341756c49f9/app/src/ui/app.tsx#L293-L294

That explains the second bit of feedback:

> * When selecting the `Force push...` menu option, I spotted this in the logs:

This error occurs because we don't pass `--force-with-lease` by default, which is Good and Safe™. But we need to signal that this is a force push and pass that through to the client.

> * I changed the preference setting to "do not show me force push" option, and the menu item changed to just `Push`. Couldn't get to change to `Force Push` nor back to `Force Push...`.

Good find! I'll see if I can reproduce this on my end.

> * I turned off the force push preference setting, selected the Push menu option and still got the warning dialogue. I hit `X` and then clicked the `Force push origin` button and it push the code.

I think you're seeing the "Push failed" error message - let's revisit this after I've tweaked the main process to send a `force-push` event to the renderer.

shiftkey:
I've pushed two commits to this branch to address @tierninho's feedback:

 - bd690820efc6a7b2bc222c4bf9cec8e73badfaf8 - the first commit which passes through the original context, which we need to know what menu to show and what action to fire. I prefer this approach rather than maintaining two sets of string checks, and I'll come in later to follow up with other cleanup.
 - 740b8fd4da4db62aedbd00caaa6a0b0d9ddc54c2 - the menu labels should be updated immediately when the preferences change, but we weren't doing this. I wasn't happy with the internals of `AppStore.updateMenuItemLabels` so I took the chance here to clean this up and make it more explicit and better documented - this will close out #7231, apologies @iAmWillShepherd

tierninho:
All issues I found have now been resolved and verified. Nice work! 💰 

The "other" dialogue is gone and now matches this, no matter which method you use: menu or button.

image

Did another round of testing, and all LGTM.

tierninho:
Found an instance when the branch is protected and the menu option and the main push button are not in sync.

Try to force push to a protected branch. After dismissing the warning, you should end up with this:
Screen Shot 2019-04-10 at 1 53 11 PM

If you select Force Push from the menu, you get this dialogue (this is the one I saw earlier before your fixes came in).
Screen Shot 2019-04-10 at 1 55 17 PM

Should we have any UI treatment if the branch is protected to prevent this?

shiftkey:
> Should we have any UI treatment if the branch is protected to prevent this?

The problem here is that we don't refresh the menu items after a Git operation - this isn't a scenario we've really had to think about before. If we keep those in sync, it should revert to `Push` and retain the classic behaviour.

There's a bit here that's unclear to me:

> If you select Force Push from the menu, you get this dialogue (this is the one I saw earlier before your fixes came in).
>

The menu label here seems correct if we're not updating the menu labels, but the action itself seems to be doing a push without the `--force-with-lease`, which is the expected behaviour.

Going to see if I can understand this mismatch this morning and wrap this up - thanks for being so methodical @tierninho!

shiftkey:
@tierninho I haven't been able to trip the issue you noticed here:

> If you select Force Push from the menu, you get this dialogue (this is the one I saw earlier before your fixes came in).
> Screen Shot 2019-04-10 at 1 55 17 PM

After force pushing to a protected branch (and seeing it fail) the menu item should be back in sync with the push/pull button state.

tierninho:
@shiftkey So I tested this again and here are the steps:

- Make branch A a protected branch, and rebase branch A onto any other branch.
- When you get to Force Push, click and wait for protected branch warning.
- Dismiss warning and go to Repo menu and it should now be `Push`, select it
- This warning pops up with a console error:

Screen Shot 2019-04-11 at 10 06 13 AM

Is this a normal use case, as I should be pulling instead of pushing?

shiftkey:
> Is this a normal use case, as I should be pulling instead of pushing?

No, and yes. 

This is not the typical use case because we suspect most users will click the push/pull button - which will "pull with merge" or "pull with rebase" depending on your setting. 

The dialog is from #6610 which might not seem as intuitive as it could be, but having improved Protected Branches support here would make this less of an issue.

app/src/lib/stores/app-store.ts
```diff
@@ -1646,50 +1648,85 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-  private getDefaultBranchName(repository?: Repository) {
-    const branchesState = this.getBranchesState(repository)
-    if (branchesState == null) {
+  private isCurrentBranchForcePush(state: IRepositoryState) {
+    const { branchesState, aheadBehind } = state
+    return isCurrentBranchForcePush(branchesState, aheadBehind)
+  }
+
+  private getRemoveRepoLabel() {
+    return this.confirmRepoRemoval
```
outofambit: this is a bit nitpicky, but could we avoid the nested ternary here? (this is partly because it looks like `prettier` doesn't know how to format it properly, which makes it harder to read)

shiftkey: Can we punt on this to a future PR? I want to stop passing strings from `AppStore` to the main process, and instead pass the information needed to decide what labels to show:

> * [bd69082](https://github.com/desktop/desktop/commit/bd690820efc6a7b2bc222c4bf9cec8e73badfaf8) - the first commit which passes through the original context, which we need to know what menu to show and what action to fire. I prefer this approach rather than maintaining two sets of string checks, and I'll come in later to follow up with other cleanup

outofambit: ah ok yes! missed that context.

PR #7257, opened 2019-04-11
[Stashing] Metrics

### LINKED ISSUE #7246: [Metrics] Working with uncommitted changes
In #6107, we're working to improve things for people with uncommitted changes when they're switching branches, and allowing them to stash their changes on their current branch if they want to just leave them there and come back to them. We want to measure several things to learn about the usage of the feature and whether it's successful.

#### 1. What percentage of the time do people bring their changes across to a different branch vs. stashing them in their current branch?

This helps us understand people's usage patterns and preferences around bringing work with them vs. leaving it on a branch to come back to. Our hypothesis is that it's very common for people want to stash work on their current branch when they're switching branches and that's what makes this feature valuable for them, but we want to validate that.

#### 2. What percentage of people always choose the same action (stash or bring work with them) vs. vary their choice when switching branches?

If an individual person almost always chooses the same option, potentially allowing them to default to that action or something might make sense.

#### 3. What percentage of time are people restoring the stash vs. deleting the stash vs. neither of those things? The total of these should equal the number of times users return to a branch that contains a stash.

This helps us understand whether the work users left behind on a branch is something they actually want to come back to, or whether they're just trying to get rid of it.

#### 4. Of the times people return to a branch that contains a stash, what percentage of the time is the stash viewed? 

This will help us learn whether the feature is discoverable and whether people actually care to see the changes they previously stashed on a branch.

#### 5. What percent of changes viewed are via clicking on the "Stashed changes" row directly vs. clicking on the prompt in the suggested next step? 

This is related to (4) above, and basically allows us to understand whether the stash icon itself is sufficient or whether a prompt to view the stash is helpful in increasing discoverability.

#### 6. How many times do users currently stash outside of Desktop, and can we demonstrate that supporting this shows a decline in that overall per-user number? 

This is intended to understand whether we are improving context-switching and allowing people to do more of their overall workflow in Desktop.

**I think if it's possible to get any of these metrics in before the beta, my priority order would be (6) and (4) because those are the ones where having baseline data would likely be most helpful.**

cc: @iAmWillShepherd @nerdneha @scottypate

iAmWillShepherd:
Fixes #7246

**Adds telemetry to answer the questions:**
1. What percentage of the time do people bring their changes across to a different branch vs. stashing them in their current branch?
    * `stashCreatedOnCurrentBranchCount`
    * `changesTakenToNewBranchCount`
2. What percentage of time are people restoring the stash vs. deleting the stash vs. neither of those things? The total of these should equal the number of times users return to a branch that contains a stash.
    * `stashRestoreCount`
    * `stashDiscardCount`
    * `noActionTakenOnStashCount`
3. Of the times people return to a branch that contains a stash, what percentage of the time is the stash viewed?
    * `stashViewedAfterCheckoutCount`
    * `stashNotViewedAfterCheckoutCount`
4. What percent of changes viewed are via clicking on the "Stashed changes" row directly vs. clicking on the prompt in the suggested next step? 
    * `suggestedStepViewStash`
    * `stashViewCount`
5.  How many times do users currently stash outside of Desktop, and can we demonstrate that supporting this shows a decline in that overall per-user number? 
    * `stashEntriesCreatedOutsideDesktop` - Implemented in #7470

## Testing
To see metrics, go to Application -> IndexedDB -> StatsDatabase -> dailyMeasures in chrome devtools.

**`stashCreatedOnCurrentBranchCount`**
1. Note the value of `stashCreatedOnCurrentBranchCount`
1. Make changes
1. Switch branches
1. Select option to "stash on current branch"
1. Verify  `stashCreatedOnCurrentBranchCount` has been incremented by 1

**`changesTakenToNewBranchCount`**
1. Note the value of `changesTakenToNewBranchCount`
1. Make changes
1. Switch branches
1. Select option to "bring changes to new branch"
1. Verify  `changesTakenToNewBranchCount` has been incremented by 1

**`stashRestoreCount`** (performed on branch with existing stash)

1. Note the value of `stashRestoreCount`
1. Click _Stashed Changes_ button to show stash diff
1. Click _Restore_
1. Verify  `stashRestoreCount` has been incremented by 1

_note, metric will increment even if restore fails since it's measuring user intent, not restore success_

**`stashDiscardCount`** (performed on branch with existing stash)
1. Note the value of `stashDiscardCount` 
1. Click _Stashed Changes_ button to show stash diff
1. Click _Discard_
1. Verify  `stashDiscardCount` has been incremented by 1

_note, metric will increment even if restore fails since it's measuring user intent, not discard success_

**`noActionTakenOnStashCount`** (performed on branch with existing stash)
1. Note the value of `noActionTakenOnStashCount` 
1. Click _Stashed Changes_ button to show stash diff
1. Click _Stashed Changes_ button again to hide the stash diff
1. Verify  `noActionTakenOnStashCount` has been incremented by 1

**`stashViewedAfterCheckoutCount`**
1. Note the value of `noActionTakenOnStashCount` 
1. Checkout branch with existing stash
1. Click _Stashed Changes_ button to show stash diff
1. Verify  `stashViewedAfterCheckoutCount` has been incremented by 1
1. Click _Stashed Changes_ button to show stash diff again
1. Verify  `stashViewedAfterCheckoutCount` has **not** changed

**`stashNotViewedAfterCheckoutCount`**
1. Note the value of `stashNotViewedAfterCheckoutCount` 
1. Checkout branch with existing stash
1. Checkout another branch (doesn't matter if it has stash or not)
1. Verify  `stashNotViewedAfterCheckoutCount` has been incremented by 1

**`suggestedStepViewStash`** (performed on branch with existing stash)
1. Note the value of `suggestedStepViewStash` 
1. Click the suggested next step to view stash
1. Verify  `suggestedStepViewStash` has been incremented by 1

**`stashViewCount`** (performed on branch with existing stash)
1. Note the value of `stashViewCount` 
1. Click _Stashed Changes_ button to **show** stash diff
1. Verify  `stashViewCount` has been incremented by 1
1. Click _Stashed Changes_ button to **hide** stash diff
1. Verify  `stashViewCount` has remained unchanged
1. Click _Stashed Changes_ button to **show** stash diff
1. Verify  `stashViewCount` has been incremented by 1

**`stashEntriesCreatedOutsideDesktop`** (start in a repo other than the testing repo)
1. Note the value of `stashEntriesCreatedOutsideDesktop` 
1. Create 1 stash outside of desktop in the testing repo
1. Open the testing repo
1. Verify `stashEntriesCreatedOutsideDesktop` has been incremented by 1

_note, since this metric is incremented on a schedule, once verifying the metric, a new repo will have to be created to get around the 24 hour check._


app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -1916,4 +1922,32 @@ export class Dispatcher {
...
+
+  /**
+   * Hide the UI for stashed changes
+   */
+  public hideStashEntry(repository: Repository) {
+    return this.appStore._hideStashEntry(repository)
+  }
+
+  /** Record when the user takes no action on the stash entry */
+  public recordNoActionTakenOnStash(): Promise<void> {
```
iAmWillShepherd: This function is for you @outofambit. I was thinking you could call it when you click the dismiss button in the diff UI.

iAmWillShepherd: Or me, depending on merge order...

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -2043,4 +2099,9 @@ export class Dispatcher {
       branchToCheckout
     )
   }
+
+  /** Record when the user takes no action on the stash entry */
+  public recordNoActionTakenOnStash(): Promise<void> {
```
iAmWillShepherd: Since we didn't end up implementing the `x` button to close the diff view, I am using the toggle button in the `ChangesList` component as an alternative.

app/src/lib/stats/stats-store.ts
```diff
@@ -989,6 +998,62 @@ export class StatsStore implements IStatsStore {
     }))
   }
 
+  /** Record when the user views a stash entry after checking out a branch */
+  public async recordStashViewedAfterCheckout(): Promise<void> {
```
shiftkey: :art: as none of these new functions use `await`, the `async` attribute can be dropped:

```suggestion
  public recordStashViewedAfterCheckout(): Promise<void> {
```

app/src/lib/stores/app-store.ts
```diff
@@ -373,6 +373,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
   private selectedTheme = ApplicationTheme.Light
   private automaticallySwitchTheme = false
 
+  private hasUserViewedStash: boolean = false
```
shiftkey: :art:

```suggestion
  private hasUserViewedStash = false
```

iAmWillShepherd:
I will add docs for metrics in a separate PR so this won't be held up.

app/src/lib/stats/stats-store.ts
shiftkey: @iAmWillShepherd just an FYI - there's another seven function in here that can use the same treatment - I just didn't explicitly go through them all

iAmWillShepherd: I see there are actually more than 7 that this can be done to. I'll open up a follow-up PR where I fix them all.

iAmWillShepherd: Feedback addressed in https://github.com/desktop/desktop/pull/7640.

tierninho:
Tested the following and results are:
- `stashCreatedOnCurrentBranchCount` ✅ 
- `changesTakenToNewBranchCount` ❌ never incremented for me
- `stashRestoreCount` ✅ 
- `stashDiscardCount` ✅ 
- `noActionTakenOnStashCount` ✅ 
- `stashViewedAfterCheckoutCount` ✅ 
- `stashNotViewedAfterCheckoutCount` ✅ 
- `suggestedStepViewStash` ✅ 
- `stashViewCount` ✅ 
- `stashEntriesCreatedOutsideDesktop` ❌ never incremented for me

Happy to revisit or 🍐 on the ones that didn't work if needed.

app/src/ui/changes/changes-list.tsx
```diff
@@ -594,10 +594,16 @@ export class ChangesList extends React.Component<
   }
 
   private onStashEntryClicked = () => {
-    if (this.props.isShowingStashEntry) {
-      this.props.dispatcher.selectWorkingDirectoryFiles(this.props.repository)
+    const { isShowingStashEntry, dispatcher, repository } = this.props
+
+    if (isShowingStashEntry) {
+      dispatcher.selectWorkingDirectoryFiles(repository)
```
outofambit: these sequential, un-`await`-ed calls to dispatcher gave me pause here (in both the `if` and the `else` statements here) due to possible race conditions but i think its ok since it seems the work done by each of these methods is independent of one another.

it would be nice to consolidate these into a single `dispatcher`/`appStore` method for each case so the component can still "fire and forget" calls to dispatcher here, but i don't think its blocking

outofambit:
> changesTakenToNewBranchCount ❌ never incremented for me

@tierninho @iAmWillShepherd this is something we need to investigate/address. i think its because `this.statsStore.recordChangesTakenToNewBranch()` is on the wrong line in `_checkoutBranch`. (will call out inline in the PR too.)

> stashEntriesCreatedOutsideDesktop ❌ never incremented for me

@tierninho that is ok and expected for this PR. that metric is implemented in #7438, not here.

app/src/lib/stores/app-store.ts
```diff
@@ -2913,6 +2924,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
               repository,
               foundBranch.name
             )
+            this.statsStore.recordChangesTakenToNewBranch()
```
outofambit: i think this metric recording should be moved down to the `if` statement at L2962 (https://github.com/desktop/desktop/pull/7257/files#diff-d6b76b602a5c48740d6d10deda6f71ebR2962) since `stashToPop` can be set in places besides this and is actually popped later.

iAmWillShepherd: I think you're right.  At the moment, it's possible to increment this metric even if checkout fails. I think this will get fixed once we refactor this flow to separate checking a branch out from stashing changes.

outofambit: f317219 and 4743c6a

app/src/lib/stats/stats-store.ts
```diff
@@ -706,22 +715,22 @@ export class StatsStore implements IStatsStore {
   }
 
   /** Record that user dismissed diverging branch notification */
-  public async recordDivergingBranchBannerDismissal(): Promise<void> {
+  public recordDivergingBranchBannerDismissal(): Promise<void> {
```
outofambit: i'd prefer we didn't make all of these signature changes in this PR, but i'm not going to consider it blocking

tierninho: Looks like `changesTakenToNewBranchCount` is now failing. Was working as of 481ab726d6c2c1f50daf0051cba5d707beb3d4c3

outofambit: ok reverted 4743c6a with b9e42bc

app/src/lib/stats/stats-store.ts
shiftkey: #7640 is fine - I just wanted to avoid adding more code that includes the unnecessary keyword

app/src/lib/stats/stats-store.ts
iAmWillShepherd: > i'd prefer we didn't make all of these signature changes in this PR, but i'm not going to consider it blocking

Yea, that was my thinking in https://github.com/desktop/desktop/pull/7257#discussion_r288309274.

shiftkey: > Looks like `changesTakenToNewBranchCount` is now failing. Was working as of [481ab72](https://github.com/desktop/desktop/commit/481ab726d6c2c1f50daf0051cba5d707beb3d4c3)

I'd like to confirm that b9e42bc05b3fa681c6d49b9a1078a312a0f9f76a is the right fix here, and is separate to the discussion about the `async` modifier here. @tierninho are you good with your QA work in here?

app/src/ui/changes/changes-list.tsx
shiftkey: :thumbsup: let's follow up after the beta ship today and see if we can settle on something we're happier with

app/src/ui/changes/no-changes.tsx
```diff
@@ -416,12 +416,14 @@ export class NoChanges extends React.Component<
         buttonText="View stash"
         type="primary"
         disabled={menuItem !== null && !menuItem.enabled}
+        onClick={this.onViewStashClicked}
       />
     )
-
-    return null
```
shiftkey: :art: I opened #7643 to improve our handling of unreachable code like this

app/src/lib/stats/stats-database.ts
```diff
@@ -227,6 +253,11 @@ export interface IDailyMeasures {
    * a keyboard shortcut only `createPullRequestCount` will increment.
    */
   readonly suggestedStepCreatePullRequest: number
+
+  /** The number of times the user has used the view stash action from
```
shiftkey: :art: The first line of a multi-line JSDoc comment should be empty - nudge this down to it's own line

app/src/lib/stores/app-store.ts
```diff
@@ -2977,6 +2988,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
         UncommittedChangesStrategyKind.moveToNewBranch &&
       checkoutSucceeded
     ) {
+      // We increment the metric after checkout succeeds to gaurd
```
```suggestion
      // We increment the metric after checkout succeeds to guard
```

app/src/lib/stats/stats-store.ts
tierninho: Sorry, missed this comment. Yes, all is good on my end.

app/src/lib/stats/stats-store.ts
lilgurl: Jkk

PR #7258, opened 2019-04-11
Only render pr bubble if there are PRs

niik:
## Overview

This is part of the larger PR work that I'm doing and is extracted here in an attempt to reduce the size of what's going to be a fairly large PR for the bigger story.

## Description

This removes the count bubble in the PR tab inside of the branches foldout when there are no PRs

image

This is now consistent with the behavior of the count bubble in the changes tab which only shows when there are changes.

image

image

## Release notes

Notes: no-notes


konifar:
.

PR #7259, opened 2019-04-11
Optional page size and continue predicate for API fetchAll method

niik:
## Overview

This is part of the larger PR work that I'm doing and is extracted here in an attempt to reduce the size of what's going to be a fairly large PR for the bigger story.

## Description

This adds a (currently unused) optional parameter to the `fetchAll` method which lets us pass in the requested page size and an optional predicate that can be used to interrupt the `fetchAll` when some condition has occurred (such as a maximum number of results).

The reason for implementing this is that we're going to load only pull requests that have changed from the last time we fetched as opposed to always loading all open PRs like we do now. Unfortunately the pull request list API doesn't support passing a `since` parameter like the Issues API does so we're going to have to do the next best thing which is to ask the API to sort PRs on last updated descending and only grab as many pages as we need.

## Release notes

Notes: no-notes


app/src/lib/api.ts
```diff
@@ -500,31 +525,27 @@ export class API {
...
-
     do {
       const response = await this.request('GET', nextPath)
-      if (response.status === HttpStatusCode.NotFound) {
-        log.warn(`fetchAll: '${path}' returned a 404`)
-        return []
-      }
-      if (response.status === HttpStatusCode.NotModified) {
-        log.warn(`fetchAll: '${path}' returned a 304`)
+      if (response.status === NotFound || response.status === NotModified) {
```
shiftkey: While you're in here I wanted to bring to your attention #7019 which was a case where we had a `HTTP 502` that wasn't handled correctly and affected cloning. Any interested in generalizing this to look for ranges of error codes (`3xx`, `4xx`, `5xx`) rather than just a few known error codes?

niik: Are you saying that you want to ignore all non-200 status codes?

shiftkey: "Ignore" isn't how I'd exactly put it. I had a play around with this in #7020 and had these things in mind at the time:

 - look for known ranges of problem error codes, rather than explicit values
 - log them out as we currently do, so we can identify them later if users spot issues
 - as soon as we encounter an error code, return the accumulated pages of items in `buf` - the app currently returns an empty array, which impacted #7019 because it would occur intermittently in the middle of a bunch of network requests

niik: Alright, take a look now and see if that's what you're looking for. Just want to make sure I understand.

app/src/lib/api.ts
```diff
@@ -500,31 +525,26 @@ export class API {
...
       }
 
       const items = await parsedResponse<ReadonlyArray<T>>(response)
       if (items) {
         buf.push(...items)
       }
+
       nextPath = getNextPagePath(response)
-    } while (nextPath)
+    } while (nextPath && opts.continue && !opts.continue(buf))
```
shiftkey: Is this check right? I feel if `opts.continue` is undefined this will evaluate to falsy rather than continue on (which was the previous behaviour):

niik: It certainly isn't. My bad, I converted this from being an inline if statement with a break and didn't invert the logic. Great catch!

niik:
🍏

shiftkey:
I'm going to merge this in after CI passes to help unblock other PR-related changes in the works

PR #7265, opened 2019-04-12
Handle text diff component unmount in the middle of syntax highlighting

### LINKED ISSUE #6970: Uncaught (in promise) TypeError: Cannot read property 'getOption' of null
## Description

When fastly switching between files in History tab, Uncaught TypeError shows in the console.

## Version

* GitHub Desktop: 0ff5f83831

* Operating system: Microsoft Windows [Version 10.0.17763.316], macOS Mojave 10.14

## Steps to Reproduce
1. Open History tab.
2. Open any commit with 2 or more files.
3. Fastly switch between files.

### Expected Behavior

Nothing errors in console.

### Actual Behavior

Uncaught TypeError which does not affect anything

## Additional Information

Error Stack:

```
Uncaught (in promise) TypeError: Cannot read property 'getOption' of null
    at TextDiff.initDiffSyntaxMode (text-diff.tsx:232)
    at process._tickCallback (internal/process/next_tick.js:68)
```

niik:
## Overview

@tierninho [spotted an error](https://github.com/desktop/desktop/pull/7210#issuecomment-482394496) while testing stashing

image

Closes #6970

## Description

This error was not related to stashing at all but deserves to be addressed in this release. The syntax highlighting process consists of two big blocks of asynchronous work, first the loading of the full contents of the file in question and second the syntax highlighting web worker doing the actual tokenization.

At any time while these asynchronous actions runs it's possible for the text diff component to become unmounted which would cause the code mirror instance to be reset to null. Unfortunately TypeScript doesn't help us here and the type narrowing on `this.codemirror` done earlier in the function gives us a false sense of security that it'll remain non-null (see https://github.com/Microsoft/TypeScript/issues/16228). I for one would love to see even stricter checks for this but it is what it is.

This PR shuffles things arounds and adds additional type guards to ensure we don't try to use the codemirror instance after we've been unmounted.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -223,14 +223,14 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
     const propsSnapshot = this.props
 
     const lineFilters = getLineFilters(hunks)
+    const tsOpt = this.codeMirror.getOption('tabSize')
+    const tabSize = typeof tsOpt === 'number' ? tsOpt : 4
+
     const contents = await getFileContents(repository, file, lineFilters)
 
     if (!highlightParametersEqual(this.props, propsSnapshot)) {
       return
     }
-
-    const tsOpt = this.codeMirror.getOption('tabSize')
-    const tabSize = typeof tsOpt === 'number' ? tsOpt : 4
 
     const tokens = await highlightContents(contents, tabSize, lineFilters)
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@niik does this also close #6970?

niik:
> @niik does this also close #6970?

It should, yes. I'll update the body to that effect.

PR #7267, opened 2019-04-12
[Stashing] Only load stash files when necessary

niik:
## Overview

This addresses [one piece of feedback](https://github.com/desktop/desktop/pull/7210#discussion_r274882369) that I had in #7210

> This is a dangerous pattern, we should not be calling the dispatcher from render methods. The Repository component could get re-rendered any number of times before the `loadStashedFiles` method is is able to complete and emit an update, causing us to shell out to Git for each render.
>
> I think we can find a much better pattern for this by letting the appStore take care of loading stashed files when the right conditions exists for doing so but in the meantime I'm going to open a PR to ensure we're never loading stashed files more than once for any stash entry.

## Description

This PR does a few things. It removes flickering in the stashed changes view by persisting loaded files for a given stash entry. Prior to this the list of stashed files would get cleared on each refresh (i.e. window focus) but due to the nature of stashes we can be certain that as long as the commit SHA is still the same the list of files will be as well.

Second, this PR now set the `Loading` state when commencing loading files for a stash. This lets us ensure that only one git process is launched per stash entry.

Finally, this takes into account the asynchronous nature of `loadStashedFiles` and accounts for the fact that the stash entry passed to the method may be stale. It's possible that we've reloaded the list of stashes but not yet emitted an update.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -992,11 +992,7 @@ export class GitStore extends BaseStore {
         // If we've already loaded the files for this stash there's
         // no point in us doing it again. We know the contents haven't
         // changed since the SHA is the same.
-        if (
-          existing !== undefined &&
-          existing.stashSha === entry.stashSha &&
-          existing.files.kind === StashedChangesLoadStates.Loaded
-        ) {
+        if (existing !== undefined && existing.stashSha === entry.stashSha) {
           map.set(entry.branchName, {
             ...entry,
             files: existing.files,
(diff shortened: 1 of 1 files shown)


iAmWillShepherd:
Merging into main branch to help make it easier to keep things in sync.

outofambit:
thank you @niik!! this is great.

b0810:
A quoi ça sert votre truc ? vous êtes nombreux à venir sur mon pc pourquoi parce que je ne vous vois pas m'aider?

PR #7268, opened 2019-04-12
[Stashing] Git stash method, delimiter parsing, and don't swallow errors

niik:
## Overview

This addresses [some feedback I gave in](https://github.com/desktop/desktop/pull/7210#discussion_r274887038) #7210. 

## Description

Instead of parsing `stderr` looking for a string we're now simply counting 128 among the successful exit codes. This simplifies the logic and lets all other errors flow freely past this method. Furthermore this PR implements the same parsing logic that we use elsewhere for log-related output. As a side-effect of this I have now included the stash reference name in the output from git log. Conceptually we want to rely on Git behaviors wherever possible and in doing so we're also able to clean up our logic for inferring these stash names.

Lastly I've introduced a `getStashEntryMatchinSha` method which lets us remove some duplicated logic.

## Release notes

Notes:

### DIFF
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -91,11 +91,8 @@ export async function createDesktopStashEntry(
   branchName: string
 ) {
   const message = createDesktopStashMessage(branchName)
-  await git(
-    ['stash', 'push', '--include-untracked', '-m', message],
-    repository.path,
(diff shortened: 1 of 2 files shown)


iAmWillShepherd:
Merging into main line brach to make it easier to keep things in sync.

PR #7301, opened 2019-04-15
[Stashing] Address button order of stash diff viewer for Windows

### LINKED ISSUE #7273: [Stashing] macOS button order is incorrect
This issue was brought up in review of https://github.com/desktop/desktop/pull/7210#discussion_r274839393. 

https://github.com/desktop/desktop/blob/686fbbc04f9b2c229ad98799dd0c4d63c8df5d01/app/src/ui/stashing/stash-diff-viewer.tsx#L1

> 🎨 can we add some context here about why we've needed to do this?

> I concur, especially since I believe that through subverting this rule we've created the wrong order on Windows. I'll comment about that below.

Let's do a bit of validation here to see if this gives the wrong order on Windows. If so, we need to yank this rule and follow our current conventions.

outofambit:
Fixes #7273

**macOS**
macOS

**Windows (before)**
Windows-Before

**Windows (after)**
Windows-After

☝️ is the same as before...

### DIFF
--- a/app/src/ui/stashing/stash-diff-viewer.tsx
+++ b/app/src/ui/stashing/stash-diff-viewer.tsx
@@ -1,5 +1,3 @@
-/* tslint:disable:button-group-order */
-
 import * as React from 'react'
 import { IStashEntry, StashedChangesLoadStates } from '../../models/stash-entry'
 import { FileList } from '../history/file-list'
@@ -119,10 +117,10 @@ const Header: React.SFC<{
     <div className="header">
       <h3>Stashed changes</h3>
       <ButtonGroup>
-        <Button onClick={onClearClick}>Clear</Button>
         <Button onClick={onSubmitClick} type="submit">
           Restore
         </Button>
+        <Button onClick={onClearClick}>Clear</Button>
       </ButtonGroup>
     </div>
   )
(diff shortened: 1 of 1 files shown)


outofambit:
hey @niik i took a quick stab at this, is this what you had in mind?

iAmWillShepherd:
@niik there is no difference in the UI made after the reorder. Is this because one of the buttons is a submit button within the `ButtonGroup` component? I think the `ButtonGroup` is responsible for ordering the buttons, so I'm not sure if this PR is needed other than to follow our conventions.

outofambit:
[I fought the law and the law won!](https://tvtropes.org/pmwiki/pmwiki.php/Main/IFoughtTheLawAndTheLawWon)

shiftkey:
> Is this because one of the buttons is a submit button within the `ButtonGroup` component? I think the `ButtonGroup` is responsible for ordering the buttons, so I'm not sure if this PR is needed other than to follow our conventions.

@iAmWillShepherd @outofambit the missing piece to this change is the `destructive` prop on `ButtonGroup`. From the docs:

https://github.com/desktop/desktop/blob/5b774b11af24dbf5b1ed1dafeaf43e5b73750b6b/app/src/ui/lib/button-group.tsx#L4-L11

Issue #7218, opened 2019-04-04
Rebase suggestions and questions for post release 


tierninho:
Had some questions / suggestions for rebasing. Nothing here is urgent and can be address post 1.7.0 release.

- The `resolve` carrot is warped. Shouldn't it look like the Open in Editor one?
<img width="106" alt="Screen Shot 2019-04-04 at 7 55 42 AM" src="https://user-images.githubusercontent.com/14828183/55580088-78f67d80-56b5-11e9-901d-346743ba359c.png">
<img width="114" alt="Screen Shot 2019-04-04 at 8 39 53 AM" src="https://user-images.githubusercontent.com/14828183/55580091-798f1400-56b5-11e9-840f-bf7cc953a17c.png">

- When using `git rebase --skip` in the cmd line, I noticed there was nothing in the Desktop UI that popped up. Does anything from this need to be surfaced in the Desktop UI?
<img width="564" alt="Screen Shot 2019-04-04 at 8 01 47 AM" src="https://user-images.githubusercontent.com/14828183/55580273-eacec700-56b5-11e9-9a93-93d575396de1.png">

- Should we use an alternate word instead of "Continue rebase" as I noticed that when the button is active, it does not entirely mean the same thing as `git rebase --continue` cmd. This may be confusing to newer users, but I have no strong feelings on this one.
<img width="913" alt="Screen Shot 2019-04-04 at 8 12 38 AM" src="https://user-images.githubusercontent.com/14828183/55580468-5153e500-56b6-11e9-9ad4-b48b04bdfd05.png">

shiftkey:
@tierninho thanks for this input!

> The `resolve` carrot is warped. Shouldn't it look like the Open in Editor one?

Probably - these styles were borrowed from Merge Conflicts so I'll see what's available there.

> * When using `git rebase --skip` in the cmd line, I noticed there was nothing in the Desktop UI that popped up. Does anything from this need to be surfaced in the Desktop UI?

I'm not sure I understand this bit - what `git rebase --skip` does is ignore the current commit in the rebase (because it doesn't contain any tracked changes to use) and continues rebasing commits until it either completes successfully or encounters conflicts.

For the example screenshot above, it looks like you've got conflicts again to resolve, and switching back to Desktop should show those. Do you recall what Desktop did at the time? There might be some context in the log about whether it was able to detect the conflicted state again...

[...]

PR #7302, opened 2019-04-15
conflicts - add space between button label and octicon

shiftkey:
## Overview

Closes out this bit of feedback from #7218:

> * The `resolve` carrot is warped. Shouldn't it look like the Open in Editor one?
> 
> Screen Shot 2019-04-04 at 7 55 42 AM
> 
> Screen Shot 2019-04-04 at 8 39 53 AM

## Description

I ended up just focusing on the spacing between the two elements because the "Open in Editor" flow has two buttons (the left button launching the editor, the second button displaying a context menu) and that didn't feel consistent here (because I only have a context menu to show).

Current `development` branch - resolving merge and rebase conflict where manual options are required:

After this PR:

## Release notes

Notes: `[Improved] button style for manual conflict resolution options`

### DIFF
--- a/app/styles/ui/dialogs/_merge-conflicts.scss
+++ b/app/styles/ui/dialogs/_merge-conflicts.scss
@@ -75,6 +75,8 @@ dialog#rebase-conflicts-list {
         .resolve-arrow-menu {
           .octicon {
             margin-left: var(--spacing-half);
+            width: 10px;
+            height: 14px;
           }
         }
       }
--- a/app/styles/ui/dialogs/_merge-conflicts.scss
+++ b/app/styles/ui/dialogs/_merge-conflicts.scss
@@ -71,6 +71,12 @@ dialog#rebase-conflicts-list {
         display: flex;
         align-items: center;
         justify-content: center;
+
+        .resolve-arrow-menu {
+          .octicon {
+            margin-left: var(--spacing-half);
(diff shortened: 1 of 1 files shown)


tierninho:
LGTM 🎉

app/styles/ui/dialogs/_merge-conflicts.scss
```diff
@@ -71,6 +71,12 @@ dialog#rebase-conflicts-list {
         display: flex;
         align-items: center;
         justify-content: center;
+
+        .resolve-arrow-menu {
+          .octicon {
+            margin-left: var(--spacing-half);
```
niik: While you're in here could I possibly convince you to apply the same size constraints for the arrow that we have for the other button?

Either by copying and pasting:
```suggestion
            margin-left: var(--spacing-half);
            width: 10px;
            height: 14px;
```

Or by changing the class name from `resolve-arrow-menu` to just `arrow-menu` and letting the other styles apply here as well.

shiftkey: I ended up taking in this suggested change because I wasn't sure about the best way to keep the `margin-left` style around and leverage `arrow-menu` (need to differentiate between a manual conflict and a file with conflict markers, but there are no suitable selectors currently).

PR #7306, opened 2019-04-15
[Stashing] Update stash state after dropping entry

### LINKED ISSUE #7280: [Stashing] Overwriting stashes show old diff state
After creating a stash of a file, the diff of stashed file changes upon returning to the branch with an associated stash does not reflect what was just stashed. The diff seems to show old state.

**Repro steps**
1. Update the contents of an existing file
1. Stash changes on the current branch
1. Go back to branch stash was created on
1. View diff of stash (note its content)
1. Update the contents of the same file from step 1
1. Stash changes on the current branch
1. Go back to branch stash was created on
1. View diff of stash (note the content doesn't reflect what was just stashed

☝️I reloaded the app to force a state update because the previous stash entry's files were being show. After reload, the contents of the diff is correct.

iAmWillShepherd:
Fixes #7280

The issue turned out to be two fold:
1. We weren't updating our cache of Desktop stash entries, `GitStore.stashEntries`
1. We weren't updating `ChangesState` when an entry was dropped

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4787,6 +4787,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
       }`
     )
 
+    this._changeStashedFileSelection(repository, null)
+
     await this._refreshRepository(repository)
   }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4786,6 +4786,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
         previousStash.branchName
       }`
     )
+
+    await this._refreshRepository(repository)
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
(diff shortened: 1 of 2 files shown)


tierninho:
I could not verify this change. I am still seeing the same stashed diff when the two versions of the same file are stashed in two different branches. 

**Steps:**
- Select an existing file, edit it and stash it on Branch A
- Switch to branch B, edit the same file (in a different way) and stash on Branch B

**Result:**
- Stashed file: the diff each branch will match (they should not) 
- Popped file: the diffs will be different (as expected)

I also stashed a completely different file on Branch B and the stashed version was the same file from Branch A.

niik:
Given that @tierninho (and myself) are still able to reproduce this I'm going to remove the ready-for-review label. I'm also going to to open a proposed alternate solution to this. I believe that while we might be able to iron out these specific problems in time for beta we will always be fighting our state until we take a step back and figure out an approach that will help us manage state.

niik:
> I'm also going to to open a proposed alternate solution to this.

PR: #7313

outofambit:
closing in favor of #7313

Issue #7290, opened 2019-04-13
[stashing] display right-click menu in stash diff


tierninho:
The right-click menu is also nice when clicking on an individual file, however its use in the stash diff page may lead to two conflicting files.

Yes, the user can still get in this position otherways, but let's not make it too easy. I suggest we remove some or all of the items in the right-click menu.

<img width="840" alt="Screen Shot 2019-04-12 at 3 14 17 PM" src="https://user-images.githubusercontent.com/14828183/56072773-fb2a1600-5d35-11e9-9e9f-6d9e68fb91cd.png">

5977a42c21, mac

billygriffin:
What's interesting about this is that "Open in..." anything might be misleading because these are stashed changes and therefore won't be reflected in the current state of the file until they're restored. So I think I agree with disabling most of the functionality of the context menu (maybe with the exception of "Copy File Path"?), though I think it's a nice to have and definitely doesn't need to get in prior to beta.

shiftkey:
I'm going to look into what's required to be able to disable the context menu. Currently this list is hard-coded inside `file-list.tsx` and it always behaves the same:

https://github.com/desktop/desktop/blob/f3898a8f58aa1d8a3f74e3353b4cff4bc43006f9/app/src/ui/history/file-list.tsx#L107-L159

shiftkey:
For the moment the stash diff does not show a context menu, as #7307 was merged.

I've tweaked the title to indicate this issue can now focus on what to show in the context menu, but feel free to open a fresh issue here if you prefer.

billygriffin:
I'm inclined to just leave it here for now unless we get feedback on it. The stash doesn't need to be an exit point imo unless we hear from users that there's a reason why they're using it for some purpose we haven't thought about. I'm going to close this issue but happy to reopen if y'all disagree.

PR #7307, opened 2019-04-16
disable context menu when viewing files in diff

shiftkey:
## Overview

**Related to #7290**

## Description

This PR refactors the `onContextMenu` callback inside `FileList` in favour of a callback, so that the components can control the behaviour:

 - `FileList` now omits several unnecessary props because `onContextMenu` no longer lives in there
 - `SelectedCommit` now contains the existing behaviour
 - `StashDiffViewer ` cleans up the props which are no longer needed

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/history/file-list.tsx
+++ b/app/src/ui/history/file-list.tsx
@@ -13,7 +13,7 @@ interface IFileListProps {
   readonly selectedFile: CommittedFileChange | null
   readonly onSelectedFileChanged: (file: CommittedFileChange) => void
   readonly availableWidth: number
-  readonly onContextMenu?: () => void
+  readonly onContextMenu?: (event: React.MouseEvent<HTMLDivElement>) => void
 }
 
 /**
@@ -40,7 +40,7 @@ export class FileList extends React.Component<IFileListProps> {
       statusWidth
 
     return (
-      <div className="file" onContextMenu={this.onContextMenu}>
+      <div className="file" onContextMenu={this.props.onContextMenu}>
         <PathLabel
           path={file.path}
           status={file.status}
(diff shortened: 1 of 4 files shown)


shiftkey:
@niik to confirm I understand correctly the suggested changes:

1. Move the `onContextMenu` callback from being applied to each file `div`:

https://github.com/desktop/desktop/blob/1c9f8370bcec18ec2f491d2659d040af629d58ea/app/src/ui/history/file-list.tsx#L40

Up to the parent `div` for the list: 

https://github.com/desktop/desktop/blob/1c9f8370bcec18ec2f491d2659d040af629d58ea/app/src/ui/history/file-list.tsx#L62

2. Add the `React.MouseEvent<HTMLDivElement>` event to the `onContextMenu` prop, so that the caller is now responsible for calling `event.preventDefault()`

3. Clean up the internal `onContextMenu` as it shouldn't be required (just pass in the prop `onContextMenu`):

https://github.com/desktop/desktop/blob/1c9f8370bcec18ec2f491d2659d040af629d58ea/app/src/ui/history/file-list.tsx#L74-L80

niik:
> @niik to confirm I understand correctly the suggested changes:
> 
> 1. Move the `onContextMenu` callback from being applied to each file `div`:

No, I still think it should be on each div, I'm merely suggesting that we'd pass the prop event handler directly, so something like this

```diff
- <div className="file" onContextMenu={this.onContextMenu}> 
+ <div className="file" onContextMenu={this.props.onContextMenu}> 
```

> 2. Add the React.MouseEvent<HTMLDivElement> event to the onContextMenu prop, so that the caller is now responsible for calling event.preventDefault()

Yup!

> 3. Clean up the internal onContextMenu as it shouldn't be required (just pass in the prop onContextMenu):

Yup!

shiftkey:
@niik thanks for clarifying - 6864b313a2e6cbe9449c391357c8a9674efe2afc has those changes, and this has been rebased on top of #7305 which introduced conflicts with this PR.

Issue #7264, opened 2019-04-11
[Stashing] Drop associated stash when branch is deleted


iAmWillShepherd:
Stash entries are orphaned when the branch is deleted. While we could leave them around, we could run into issues if another branch is created with the same name as the one the stash was associated with. In order to avoid this, we should remove the associated stash along with the branch. 

Maybe it's worth alerting the user this will happen as well? 

cc/ @desktop/quality-assurance  @desktop/design

tierninho:
> Maybe it's worth alerting the user this will happen as well?

Alerting the user would allow them to move the stash elsewhere, so I am in favor of it. 

We should also consider what happens when a branch is restored and an associated stash was present before the deletion.

iAmWillShepherd:
@billygriffin and I have decided that it's safer to not delete stash entries. If the user does create a new branch with the same name, they can easily discard it.

PR #7313, opened 2019-04-16
[Stashing] A select selection

### LINKED ISSUE #7270: [Stashing] Move the responsibility of loading stash files from the component to the AppStore
This issue was brought up in review of #7210 and a work around was put into place in #7267, but we should address this fully when we have the opportunity. 

https://github.com/desktop/desktop/blob/75fd70d8a2ce6ba846adc00f45158eb0998254c5/app/src/ui/repository.tsx#L289-L314

☝️ The actual lines that matter are L308-311; the surrounding code is for better context.

> This is a dangerous pattern, we should not be calling the dispatcher from render methods. The Repository component could get re-rendered any number of times before the loadStashedFiles method is is able to complete and emit an update, causing us to shell out to Git for each render.
>
>  I think we can find a much better pattern for this by letting the appStore take care of loading stashed files when the right conditions exists for doing so but in the meantime I'm going to open a PR to ensure we're never loading stashed files more than once for any stash entry.

### LINKED ISSUE #7280: [Stashing] Overwriting stashes show old diff state

### LINKED ISSUE #7281: [Stashing] Focus gets reset from stashed file change list to changes list when app regains focus
Repro steps:

 - make change to README.md in repository
 - switch branches, `Yes, stash changes...`
 - return to branch, make changes to any file
 - select stash to see contents
 - switch away from Desktop to another app
 - focus on Desktop again

Expected behaviour:
 - stash has focus, no changed files selected

Actual behaviour:
 - stash has focus, but first changed file is selected and diff is shown

Before switching away:

After returning to Desktop:

_Originally posted by @shiftkey in https://github.com/desktop/desktop/pull/7210#issuecomment-482336189_

### LINKED ISSUE #7312: [Stashing] Selection is not reset when stash entry disappears
## Description

After selecting a stash entry in the app we will not do any checks to make sure that the selection is valid until the user explicitly chooses to select a file (or unselect the stash entry).

## Steps to Reproduce

1. Go to a branch with a stash
2. Select said stash such that you're viewing the contents of the stash entry
3. Use another tool or the terminal to check out another branch, one that does not have an associated stash entry.
4. Go back to the app

### Expected Behavior

The stash view is no longer visible

### Actual Behavior

The stash button in the sidebar is gone but the right hand side view is blank.

In this scenario we should explicitly move the selection back to the working directory regardless of whether there are changed files there or not.

niik:
## Overview

This PR updates our model to match the constraints of the user interface. In particular this makes sure that since it should only be possible to select either a changed file in the working directory or a stash entry.

**Closes #7281** - Focus gets reset from stashed file change list to changes list when app regains focus
**Closes #7280** - Overwriting stashes show old diff state
**Closes #7270** - Move the responsibility of loading stash files from the component to the AppStore
**Closes #7312** - Selection is not reset when stash entry disappears

## Description

After reading #7264 I came to the conclusion that the way we deal with state surrounding stashes is working against us. I had a similar comment in the initial review of the stashing PR though that was focused more on the performance implications.

> This is a dangerous pattern, we should not be calling the dispatcher from render methods. The Repository component could get re-rendered any number of times before the loadStashedFiles method is is able to complete and emit an update, causing us to shell out to Git for each render.
>
>  I think we can find a much better pattern for this by letting the appStore take care of loading stashed files when the right conditions exists for doing so but in the meantime I'm going to open a PR to ensure we're never loading stashed files more than once for any stash entry.

Prior to this PR we differentiated between whether the user was looking at a stash or not by one flag in changes state called `shouldShowStashedChanges`. This flag was toggled when the user selected or deselected the stash entry in the sidebar. Next to this flag we kept the state of the currently selected stashed file (if any) alongside the currently selected file (or files) in the working directory.

In doing so we've created a scenario where it's perfectly legal from the perspective of the model to have selected a file in the working directory and a file in a stash entry at the same time and we therefore put the responsibility of ensuring that never happened on the UI. Such an approach is prone to errors and confusing to deal with as a developer.

With this PR we're restructuring our models to more closely reflect the reality of the UI, i.e. one may select a file in the working directory, or a file in a stash, but never both.

Additionally this moves us away from having all stash entries for all branches available in the UI layer and instead let the stores take care of figuring out whether the _current_ branch has an associated stash entry or not.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -522,17 +522,29 @@ export interface ICommitSelection {
   /** The diff of the currently-selected file */
   readonly diff: IDiff | null
 }
-
-export interface IChangesState {
-  readonly workingDirectory: WorkingDirectoryStatus
+export type ChangesWorkingDirectorySelection = {
+  readonly kind: 'workingDirectory'
 
   /**
    * The ID of the selected files. The files themselves can be looked up in
    * `workingDirectory`.
    */
   readonly selectedFileIDs: string[]
-
   readonly diff: IDiff | null
+}
+
+export type ChangesStashSelection = {
+  readonly kind: 'stash'
+
+  /** Currently selected file in the stash diff viewer UI (aka the file we want to show the diff for) */
+  readonly selectedStashedFile: CommittedFileChange | null
+
+  /** Currently selected file's diff */
+  readonly selectedStashedFileDiff: IDiff | null
+}
+
+export interface IChangesState {
+  readonly workingDirectory: WorkingDirectoryStatus
 
   /** The commit message for a work-in-progress commit in the changes view. */
   readonly commitMessage: ICommitMessage
@@ -559,14 +571,10 @@ export interface IChangesState {
    */
   readonly conflictState: ConflictState | null
 
-  /** Whether or not to show the UI for a stash entry. */
-  readonly shouldShowStashedChanges: boolean
-
-  /** Currently selected file in the stash diff viewer UI (aka the file we want to show the diff for) */
-  readonly selectedStashedFile: CommittedFileChange | null
-
-  /** Currently selected file's diff */
-  readonly selectedStashedFileDiff: IDiff | null
+  readonly selection:
+    | ChangesWorkingDirectorySelection
+    | ChangesStashSelection
+    | null
 }
 
 /**
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -584,6 +584,12 @@ export interface IChangesState {
    */
   readonly stashEntry: IStashEntry | null
 
+  /**
+   * The current selection state in the Changes view. Can be either
+   * working directory or a stash. In the case of a working directory
+   * selection multiple files may be selected. See `ChangesSelection`
+   * for more information about the differences between the two.
+   */
   readonly selection: ChangesSelection
 }
 
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -522,6 +522,12 @@ export interface ICommitSelection {
   /** The diff of the currently-selected file */
   readonly diff: IDiff | null
 }
+
+export enum ChangesSelectionKind {
+  WorkingDirectory,
+  Stash,
+}
+
 export type ChangesWorkingDirectorySelection = {
   readonly kind: 'workingDirectory'
 
@@ -534,7 +540,7 @@ export type ChangesWorkingDirectorySelection = {
 }
 
 export type ChangesStashSelection = {
-  readonly kind: 'stash'
+  readonly kind: ChangesSelectionKind.Stash
 
   /** Currently selected file in the stash diff viewer UI (aka the file we want to show the diff for) */
   readonly selectedStashedFile: CommittedFileChange | null
@@ -581,7 +587,7 @@ export interface IChangesState {
    */
   readonly stashEntry: IStashEntry | null
 
-  readonly selection: ChangesSelection | null
+  readonly selection: ChangesSelection
 }
 
 /**
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -575,6 +575,12 @@ export interface IChangesState {
    */
   readonly conflictState: ConflictState | null
 
+  /**
+   * The latest GitHub Desktop stash entry for the current branch, or null
+   * if no stash exists for the current branch.
+   */
+  readonly stashEntry: IStashEntry | null
+
   readonly selection: ChangesSelection | null
 }
 
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -530,7 +530,7 @@ export type ChangesWorkingDirectorySelection = {
 
   /**
    * The ID of the selected files. The files themselves can be looked up in
-   * `workingDirectory`.
+   * the `workingDirectory` property in IChangesState.
    */
   readonly selectedFileIDs: string[]
   readonly diff: IDiff | null
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -529,7 +529,7 @@ export enum ChangesSelectionKind {
 }
 
 export type ChangesWorkingDirectorySelection = {
-  readonly kind: 'workingDirectory'
+  readonly kind: ChangesSelectionKind.WorkingDirectory
 
   /**
    * The ID of the selected files. The files themselves can be looked up in
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -371,9 +371,6 @@ export interface IRepositoryState {
   /** The state of the current branch in relation to its upstream. */
   readonly aheadBehind: IAheadBehind | null
 
-  /** A map keyed on the canonical ref name of stash entries created by Desktop. */
-  readonly stashEntries: ReadonlyMap<string, IStashEntry>
-
   /** Is a push/pull/fetch in progress? */
   readonly isPushPullFetchInProgress: boolean
 
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -543,6 +543,10 @@ export type ChangesStashSelection = {
   readonly selectedStashedFileDiff: IDiff | null
 }
 
+export type ChangesSelection =
+  | ChangesWorkingDirectorySelection
+  | ChangesStashSelection
+
 export interface IChangesState {
   readonly workingDirectory: WorkingDirectoryStatus
 
@@ -571,10 +575,7 @@ export interface IChangesState {
    */
   readonly conflictState: ConflictState | null
 
-  readonly selection:
-    | ChangesWorkingDirectorySelection
-    | ChangesStashSelection
-    | null
+  readonly selection: ChangesSelection | null
 }
 
 /**
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1959,7 +1959,16 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
   }
 
-  /** This shouldn't be called directly. See `Dispatcher`. */
+  /**
+   * Changes the selection in the changes view to the working directory and
+   * optionally selects a particular file from the working directory.
+   *
+   *  @param files An array of files to select when showing the working directory.
+   *               If undefined this method will preserve the previously selected
+   *               files or pick the first changed file if no selection exists.
+   *
+   * Note: This shouldn't be called directly. See `Dispatcher`.
+   */
   public async _selectWorkingDirectoryFiles(
     repository: Repository,
     files?: ReadonlyArray<WorkingDirectoryFileChange>
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -601,12 +601,26 @@ export class AppStore extends TypedBaseStore<IAppState> {
       pullWithRebase: gitStore.pullWithRebase,
     }))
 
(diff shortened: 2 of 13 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -2045,12 +2077,126 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          // than references.
+          selectedStashedFile = files.find(x => x.id === file.id) || null
+        } else {
+          selectedStashedFile = null
+        }
+      }
+
+      return {
+        selection: {
+          kind: ChangesSelectionKind.Stash as ChangesSelectionKind.Stash,
```
niik: This, and the other similar casts in here is a concession I had to make to get around a TypeScript compiler error that I haven't had the time to get to the bottom of.

```Type 'ChangesSelectionKind' is not assignable to type 'ChangesSelectionKind.Stash'```

I don't get this error in vs code fwiw.

Comment (niik) [reply app/src/lib/stores/app-store.ts]: > I don't get this error in vs code fwiw.

Oh, I should have realized that's because I had vs code configured to run the typescript version that ships with vs code and not the one configured for our project. So this will be solved the next time we bump typescript most likely.

File: app/src/lib/stores/app-store.ts [hunk 2]
```diff
@@ -2045,12 +2077,126 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      },
       workingDirectory,
     }))
     this.emitUpdate()
   }
 
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _selectStashedFile(
+    repository: Repository,
+    file?: CommittedFileChange | null
```
shiftkey: 💭 are we able to just use `CommittedFileChange | null` here and omit the `?`? Or if the `undefined` case is unavoidable can we document why it's important here and how it differs from `null`?

app/test/unit/stores/updates/update-changed-files-test.ts
```diff
@@ -31,7 +33,7 @@ const files = [
 ]
 
 describe('updateChangedFiles', () => {
-  describe('workingDirectory', () => {
+  describe(ChangesSelectionKind.WorkingDirectory, () => {
```
shiftkey: 💭 any interest in adding a corresponding stash test in the scope of this PR? Feel free to punt to another issue if you're feeling under pressure..

app/src/lib/stores/app-store.ts
niik: Right, I made the effort to document this for the working directory case, I just forgot about it here. Let me fix that.

app/test/unit/stores/updates/update-changed-files-test.ts
niik: lol, I think this is a search and replace gone awry. Let me just reset that. Stashing doesn't have anything to do with `updateChangedFiles`

app/src/lib/stores/app-store.ts
niik: Added some more documentation to address this and explain the difference between undefined and null

niik:
Cleaned up some logic and added more documentation :apple:

app/src/lib/stores/app-store.ts
```diff
@@ -2045,12 +2086,141 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const workingDirectory = WorkingDirectoryStatus.fromFiles(updatedFiles)
 
     this.repositoryStateCache.updateChangesState(repository, () => ({
-      diff,
+      selection: {
+        ...changesState.selection,
```
shiftkey: There's a subtle problem here I wanted to call out and see how we feel.

Because we're inside a callback, `changesState.selection` reverts to the union type `ChangesSelection` rather than the known `ChangesWorkingDirectorySelection` that we were looking at above.

You can see this by extracting out the `selection` to a local variable and stating that it should satisfy `ChangesWorkingDirectorySelection`:

```ts
    this.repositoryStateCache.updateChangesState(repository, () => {
      const selection: ChangesWorkingDirectorySelection = {
        ...changesState.selection,
        diff,
      }

      return {
        selection,
        workingDirectory,
      }
    })
```

Which produces the compiler error because the previous checks for `changesState.selection` are no longer valid:

```
Type '{ diff: IDiff; kind: ChangesSelectionKind.WorkingDirectory; selectedFileIDs: string[]; } | { diff: IDiff; kind: ChangesSelectionKind.Stash; selectedStashedFile: CommittedFileChange | null; selectedStashedFileDiff: ITextDiff | ... 4 more ... | null; }' is not assignable to type 'ChangesWorkingDirectorySelection'.
  Property 'selectedFileIDs' is missing in type '{ diff: IDiff; kind: ChangesSelectionKind.Stash; selectedStashedFile: CommittedFileChange | null; selectedStashedFileDiff: ITextDiff | IImageDiff | IBinaryDiff | ILargeTextDiff | IUnrenderableDiff | null; }' but required in type 'ChangesWorkingDirectorySelection'.ts(2322)
app-state.ts(535, 12): 'selectedFileIDs' is declared here.
```

The way to appease the compiler is to create the `selection` before invoking `updateChangesState` when we can assert it's a valid state, which is what I'd like to see here:

```ts
    const selection: ChangesWorkingDirectorySelection = {
      ...changesState.selection,
      diff,
    }

    this.repositoryStateCache.updateChangesState(repository, () => {
      return {
        selection,
        workingDirectory,
      }
    })
```

Thoughts?

app/test/unit/stores/updates/update-changed-files-test.ts
```diff
@@ -118,39 +120,62 @@ describe('updateChangedFiles', () => {
...
-      const { selectedFileIDs } = updateChangedFiles(prevState, status, false)
+      const { selection } = updateChangedFiles(prevState, status, false)
 
-      expect(selectedFileIDs).toHaveLength(1)
-      // NOTE: `updateChangedFiles` sorts the paths and `app/package.json` will
-      // appear in list before `README.md`
-      expect(selectedFileIDs[0]).toBe(files[1].id)
+      expect(selection.kind).toBe(ChangesSelectionKind.WorkingDirectory)
+
+      if (selection.kind === ChangesSelectionKind.WorkingDirectory) {
```
shiftkey: :art: because of how union type checks don't flow through Jest assertions, we've previously gone through this sort of pattern to ensure assertions are always run:

```ts
expect(selection.kind).toBe(ChangesSelectionKind.WorkingDirectory)
const workingDirectorySelection = selection as ChangesWorkingDirectorySelection
const { selectedFileIDs } = workingDirectorySelection
expect(selectedFileIDs).toHaveLength(1)
expect(selectedFileIDs[0]).toBe(firstFile)
```

I'll put together a cleanup PR for this so it doesn't hold up the PR, just wanted to call this out in advance

app/src/lib/app-state.ts
```diff
@@ -523,16 +520,38 @@ export interface ICommitSelection {
   readonly diff: IDiff | null
 }
 
-export interface IChangesState {
-  readonly workingDirectory: WorkingDirectoryStatus
+export enum ChangesSelectionKind {
+  WorkingDirectory,
```
shiftkey: :art: string enums for better debuggability

```suggestion
  WorkingDirectory = 'WorkingDirectory',
```

app/src/lib/app-state.ts
```diff
@@ -523,16 +520,38 @@ export interface ICommitSelection {
...
+  Stash,
```
shiftkey: :art:

```suggestion
  Stash = 'Stash',
```

tierninho:
My notes from testing this PR:

- Closes #7281 - ✅
- Closes #7280 - ✅overall, but I noticed the diff is missing when you restore a stash. Tab over to History and back and it reappears:
Screen Shot 2019-04-16 at 8 13 52 AM

- Closes #7270 - I will defer to shiftkey's review as mentioned in https://github.com/desktop/desktop/pull/7313#pullrequestreview-227180550
- Closes #7312 - ✅

8488b66e0a, mac

app/src/lib/stores/app-store.ts
niik: That is some weak typechecking, I am disappoint. Given that `changesState` is a `const` and `selection` is a readonly property containing only readonly members. I suspect this is just an type checker performance issue given that extracting `changesState.selection` itself into a const will not yield an error.

> There's a subtle problem here

Just to be super clear, there's not really a problem right? Only a weak checker? I'll apply the suggestion you made regardless, just want to be sure that you didn't actually find a case where this could cause an issue.

niik:
> * overall, but I noticed the diff is missing when you restore a stash. Tab over to History and back and it reappears:

@tierninho I'm unable to reproduce this with the latest changes. Could I ask you kindly to give this another shot or if you've got any tips on what steps I need to take to reproduce.

app/src/lib/stores/app-store.ts
shiftkey: @niik yep, it's an limitation with the typechecker because it's lacks the context we built up earlier in the function

niik:
> @tierninho I'm unable to reproduce this with the latest changes. Could I ask you kindly to give this another shot or if you've got any tips on what steps I need to take to reproduce.

Sorry, I've been able to reproduce it now and I know the cause, gimme a moment and I'll have that fixed.

tierninho:
> Ensure we load the diff after switching to workingDirectory 

Nice, looks to be working now ✨

shiftkey:
> Ensure we load the diff after switching to workingDirectory

I've been able to reproduce something in this vein using 589856cc01a31daaf8f68ee94207d32d0d1def53 when the stash contents change underneath the app (#7280).

Repro steps:

 - create a stash that modifies a tracked `README.md` (my example is removing a paragraph) on a branch (I'm using `linux` for this demo)
 - switch branches, store the stash, then switch back to the branch
 - select the `README.md` file to view the diff



 - switch to the command line and pop the stash to make changes with it

```
$ git stash pop
```

 - make another change to the `README.md` (in my example I added the missing paragraph to the end of the page)
 - stash the changes again using thee right pattern and the branch name

```
$ git stash push --include-untracked -m "\!\!GitHub_Desktop<linux>"
```

 - switch back to the app, and the `README.md` should remain selected (but have the old diff)



Workarounds:

 - :x: switching between `History` and `Changes` tab doesn't seem to refresh the diff
 - :x: I don't have any working directory changes to switch to to potentially refresh the diff.
 - ✅ choosing a different file in the stash and switching back refreshes the diff

niik:
> I've been able to reproduce something in this vein using [589856c](https://github.com/desktop/desktop/commit/589856cc01a31daaf8f68ee94207d32d0d1def53) when the stash contents change underneath the app (#7280).

I've updated the `onGitStoreUpdated` method in b2e46ad to react to when the stash entry sha changes and update the selection when that happens. Would you mind having another go at your repro case?

iAmWillShepherd:
Merged early since https://github.com/desktop/desktop/pull/7313#issuecomment-483820799 is a scenario a user has to try to get into and I don't think that should be a blocker for beta. If https://github.com/desktop/desktop/pull/7313/commits/b2e46ad1bc6c7b75137fc6ce80f9bdb6fa716fd2 does not address the problem, we can fix it in another PR.

PR #7316, opened 2019-04-16
show untracked files in stash

### LINKED ISSUE #7287: [Stashing] Untracked files are not shown in the diff
Untracked files are not shown in the stash diff, but when restored, they are added to the working directory.

**Repro steps**
1. Create a new file with some contents
1. Create stash
1. View diff of stash

**Expected** 
The untracked file is present in the stash diff.

**Actual**
The untracked file is not there. In situations for the untracked file was the only stashed file, the diff view will appear to be empty.

outofambit:
## Overview

closes #7287 

Let's please review this ASAP so we can ship it to beta. 👩‍🎤 

gif

## Description

We now run `git diff` twice to get a list of files for a stash:
- once for tracked files (which we were doing before)
- again for untracked files (which is new)

This is because git stashes store untracked files separately from tracked files (read: in two different commits).

## Discussion

There are some minor opportunities to refactor this function, but I'm not sure they are that worthwhile, and may even make the code harder to read. There's potentially a more core git function we could derive but that would take more reflection from me to isolate.

I also want to add test coverage for this function, and will file a follow up issue for that.

### DIFF
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -175,7 +175,7 @@ export async function getStashedFiles(
 
   const trackedFiles = parseChangedFiles(trackedResult.stdout, committish)
 
-  const untrackedArgs = [...baseArgs, NullTreeSHA, `${committish}^3`, '--']
+  const untrackedArgs = [...baseArgs, `${NullTreeSHA}..${committish}^3`, '--']
   const untrackedResult = await git(
     untrackedArgs,
     repository.path,
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -160,22 +160,22 @@ const NullTreeSHA = '4b825dc642cb6eb9a060e54bf8d69288fbee4904'
  */
 export async function getStashedFiles(
   repository: Repository,
-  committish: string
+  stashSha: string
 ): Promise<ReadonlyArray<CommittedFileChange>> {
   // opt-in for rename detection (-M) and copies detection (-C)
   // this is equivalent to the user configuring 'diff.renames' to 'copies'
   // NOTE: order here matters - doing -M before -C means copies aren't detected
   const baseArgs = ['diff', '-C', '-M', '--name-status', '-z']
-  const trackedArgs = [...baseArgs, `${committish}^..${committish}`, '--']
+  const trackedArgs = [...baseArgs, `${stashSha}^..${stashSha}`, '--']
   const trackedResult = await git(
     trackedArgs,
     repository.path,
     'getStashedFiles (tracked)'
   )
 
-  const trackedFiles = parseChangedFiles(trackedResult.stdout, committish)
+  const trackedFiles = parseChangedFiles(trackedResult.stdout, stashSha)
 
-  const untrackedArgs = [...baseArgs, `${NullTreeSHA}..${committish}^3`, '--']
+  const untrackedArgs = [...baseArgs, `${NullTreeSHA}..${stashSha}^3`, '--']
   const untrackedResult = await git(
     untrackedArgs,
     repository.path,
@@ -193,7 +193,7 @@ export async function getStashedFiles(
   if (untrackedResult.exitCode === 0 && untrackedResult.stdout.length > 0) {
     const untrackedFiles = parseChangedFiles(
       untrackedResult.stdout,
-      `${committish}^3`
+      `${stashSha}^3`
     )
 
     // we want untracked changes to override potential
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -148,8 +148,6 @@ function extractBranchFromMessage(message: string): string | null {
 /**
  * Get the files that were changed in the given stash commit.
  *
- * Runs `git diff` twice, once for tracked changes in the stash
- * and once for untracked changes in the stash.
  * This is different than `getChangedFiles` because stashes
  * have _3 parents(!!!)_
  */
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -8,6 +8,7 @@ import {
 } from '../../models/stash-entry'
 import { CommittedFileChange } from '../../models/status'
 import { parseChangedFiles } from './log'
+import { compare } from '../compare'
 
 export const DesktopStashEntryMarker = '!!GitHub_Desktop'
 
@@ -145,12 +146,6 @@ function extractBranchFromMessage(message: string): string | null {
 }
 
 /**
- * The SHA for the [empty tree](https://stackoverflow.com/questions/9765453/is-gits-semi-secret-empty-tree-object-reliable-and-why-is-there-not-a-symbolic).
- *
- */
-const NullTreeSHA = '4b825dc642cb6eb9a060e54bf8d69288fbee4904'
-
-/**
  * Get the files that were changed in the given stash commit.
  *
  * Runs `git diff` twice, once for tracked changes in the stash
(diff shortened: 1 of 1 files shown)


app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,53 @@ function extractBranchFromMessage(message: string): string | null {
...
+      // because there weren't any untracked files,
+      // and that's okay!
+      successExitCodes: new Set([0, 128]),
+    }
+  )
 
-  // git diff -C -M --name-status -z [STASH_SHA] [STASH_SHA]^
-  const args = ['diff', '-C', '-M', '--name-status', '-z', sha, `${sha}^`, '--']
-  const result = await git(args, repository.path, 'getStashedFiles')
+  const untrackedFiles = parseChangedFiles(
```
shiftkey: I think we should only run this if `untrackedResult.exitCode === 0`, given we handle/suppress when `128` is found.

I also spotted that `untrackedResult.stdout` is an empty string here when there are no uncommitted files in the stash, so that could be another case for skipping this work...

outofambit: ooohh that's a great call. will update!

app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,53 @@ function extractBranchFromMessage(message: string): string | null {
   return match === null || match[1].length === 0 ? null : match[1]
 }
 
-/** Get the files that were changed in the given stash commit.
- *  This is different than `getChangedFiles` because stashes
- *  have _3 parents(!!!)_
+/** The SHA for the null tree. */
```
shiftkey: :memo: pedantic feedback

```suggestion
/** The SHA for the empty tree. */
```

I'd link off to an external discussion about what this value represents. This is my favourite currently: https://stackoverflow.com/questions/9765453/is-gits-semi-secret-empty-tree-object-reliable-and-why-is-there-not-a-symbolic

app/src/lib/git/stash.ts
outofambit: done in b870f5582e4387feb4b4cf7282bd1d354f330545

app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,62 @@ function extractBranchFromMessage(message: string): string | null {
...
 }
 
-/** Get the files that were changed in the given stash commit.
- *  This is different than `getChangedFiles` because stashes
- *  have _3 parents(!!!)_
+/**
+ * The SHA for the [empty tree](https://stackoverflow.com/questions/9765453/is-gits-semi-secret-empty-tree-object-reliable-and-why-is-there-not-a-symbolic).
+ *
+ */
+const NullTreeSHA = '4b825dc642cb6eb9a060e54bf8d69288fbee4904'
```
say25: Maybe it could be worth putting the “Magic” string into Dugite?

app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,62 @@ function extractBranchFromMessage(message: string): string | null {
...
+  // and merge it with the other output
+  if (untrackedResult.exitCode === 0 && untrackedResult.stdout.length > 0) {
+    const untrackedFiles = parseChangedFiles(
+      untrackedResult.stdout,
+      `${committish}^3`
+    )
+
+    // order is important here, since we want untracked changes to
+    // override potential (and uncommon) collisions with tracked changes
+    return [...trackedFiles, ...untrackedFiles]
```
shiftkey: Is this actually going to address the collisions? I worry that doing an object spread of two arrays is equivalent to concatenating them, which means the potential for duplicate values?

I put together this test to validate my assumption, and it currently fails:

```diff
diff --git a/app/test/unit/git/stash-test.ts b/app/test/unit/git/stash-test.ts
index b42581923..4b7f6a9b0 100644
--- a/app/test/unit/git/stash-test.ts
+++ b/app/test/unit/git/stash-test.ts
@@ -10,6 +10,7 @@ import {
   getLastDesktopStashEntryForBranch,
   dropDesktopStashEntry,
   popStashEntry,
+  getStashedFiles,
 } from '../../../src/lib/git/stash'
 import { getStatusOrThrow } from '../../helpers/status'
 import { AppFileStatusKind } from '../../../src/models/status'
@@ -249,6 +250,43 @@ describe('git/stash', () => {
       expect(files).toHaveLength(1)
     })
   })
+
+  describe('getStashedFiles', () => {
+    let repository: Repository
+    let readme: string
+
+    beforeEach(async () => {
+      repository = await setupEmptyRepository()
+      readme = path.join(repository.path, 'README.md')
+      await FSE.writeFile(readme, 'text before removing from index')
+      await GitProcess.exec(['add', 'README.md'], repository.path)
+      await GitProcess.exec(['commit', '-m', 'initial commit'], repository.path)
+    })
+
+    it('displays uncommitted change rather than index change', async () => {
+      // stage the delete in the index
+      await GitProcess.exec(['rm', 'README.md'], repository.path)
+
+      // add a new file to the working directory (should be untracked)
+      await FSE.writeFile(readme, 'new untracked files contents')
+
+      // stash both changes
+      await stash(repository, 'master', null)
+
+      const entries = await getDesktopStashEntries(repository)
+      expect(entries.length).toBe(1)
+
+      const { stashSha } = entries[0]
+
+      const files = await getStashedFiles(repository, stashSha)
+
+      console.log('found files', files)
+
+      expect(files.length).toBe(1)
+
+      expect(files[0].status.kind).toBe(AppFileStatusKind.Untracked)
+    })
+  })
 })
 
 /**
@@ -264,7 +302,13 @@ async function stash(
   message: string | null
 ): Promise<void> {
   const result = await GitProcess.exec(
-    ['stash', 'push', '-m', message || createDesktopStashMessage(branchName)],
+    [
+      'stash',
+      'push',
+      '--include-untracked',
+      '-m',
+      message || createDesktopStashMessage(branchName),
+    ],
     repository.path
   )
```

But its not failing quite in the way I was expecting:

```

  ● git/stash › getStashedFiles › displays uncommitted change rather than index change

    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 2

      283 |       console.log('found files', files)
      284 |
    > 285 |       expect(files.length).toBe(1)
          |                            ^
      286 |
      287 |       expect(files[0].status.kind).toBe(AppFileStatusKind.Untracked)
      288 |     })

      at Object.it (test/unit/git/stash-test.ts:285:28)

  console.log app/test/unit/git/stash-test.ts:283
    found files [ CommittedFileChange {
        path: 'README.md',
        status: { kind: 'Modified' },
        id: 'Modified+README.md',
        commitish: '8cebe5d30b27e96f7d3941579fd0d0dc78af7c83' },
      CommittedFileChange {
        path: 'README.md',
        status: { kind: 'New' },
        id: 'New+README.md',
        commitish: '8cebe5d30b27e96f7d3941579fd0d0dc78af7c83^3' } ]
```

And what's interesting is that, when we try and pop this stash, we encounter an error because `README.md` already exists:

```
$ git stash pop
README.md already exists, no checkout
Could not restore untracked files from stash entry
```

So I'm not too worried about this change given how tricky it is to setup, but I still think implementing way to filter out duplicate paths like how we do in `buildStatusMap` here would be helpful:

https://github.com/desktop/desktop/blob/4d6037998083199ae29a1a633afb286abc1b6ec7/app/src/lib/git/status.ts#L254-258

outofambit: ack, you are totally right, i spaced on that. fixing now!

outofambit: @shiftkey how does https://github.com/desktop/desktop/pull/7316/commits/6371695b2a6854a06367a1b8dfd9032b7b6ee332 feel? we're not using maps here so i went with something more array-ish

tierninho:
Looking good except one minor bug:

A deleted file in Changes is marked as a new file in the Stash diff:

Changes:
Screen Shot 2019-04-16 at 10 08 09 AM

Stash:
Screen Shot 2019-04-16 at 10 08 03 AM

6371695b2a, mac

app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,68 @@ function extractBranchFromMessage(message: string): string | null {
...
 export async function getStashedFiles(
   repository: Repository,
-  sha: string
+  committish: string
 ): Promise<ReadonlyArray<CommittedFileChange>> {
   // opt-in for rename detection (-M) and copies detection (-C)
   // this is equivalent to the user configuring 'diff.renames' to 'copies'
   // NOTE: order here matters - doing -M before -C means copies aren't detected
+  const baseArgs = ['diff', '-C', '-M', '--name-status', '-z']
+  const trackedArgs = [...baseArgs, committish, `${committish}^`, '--']
```
shiftkey: I've got a repro here that feels confusing, and might be related to this area:

 - remove a tracked file from the repository, see the diff in the working directory



 - switch branches, stash in previous branch, then return to previous branch

 - view current stash, see deleted file has correct diff but is considered "added"



I had a look at the dev tools and confirmed this was coming from this Git operation:



I think the order of these should be switched around:

```
$ git diff -C -M --name-status stash@{0} stash@{0}^
A       tslint.json
git diff -C -M --name-status stash@{0}^ stash@{0}
D       tslint.json
```

outofambit: fixed in 84bbde0

app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,68 @@ function extractBranchFromMessage(message: string): string | null {
...
- * TODO: support showing untracked files in the diff
+ * Runs `git diff` twice, once for tracked changes in the stash
+ * and once for untracked changes in the stash.
+ * This is different than `getChangedFiles` because stashes
+ * have _3 parents(!!!)_
  */
 export async function getStashedFiles(
   repository: Repository,
-  sha: string
+  committish: string
```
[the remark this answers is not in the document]
outofambit: done in b98538f

app/src/lib/git/stash.ts
```diff
@@ -144,23 +144,68 @@ function extractBranchFromMessage(message: string): string | null {
...
+    const untrackedFiles = parseChangedFiles(
+      untrackedResult.stdout,
+      `${committish}^3`
+    )
+
+    // we want untracked changes to override potential
+    // collisions with tracked changes
+    const allFiles = Array.from(untrackedFiles)
+    for (const trackedFile of trackedFiles) {
+      if (!untrackedFiles.some(f => f.path === trackedFile.path)) {
```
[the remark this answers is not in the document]
outofambit: yeah, i know this is pretty ugly in its current form. i went ahead and did this in 
0efbcd4 (with a local version of `getChangedFiles`, I didn't touch the one in `log.ts`) and opened an issue to revisit in #7320 for further refinement

app/src/lib/git/stash.ts
```diff
@@ -144,23 +145,58 @@ function extractBranchFromMessage(message: string): string | null {
...
+    getChangedFilesWithinStash(repository, `${stashSha}^3`),
+  ])
+
+  const files = new Map<string, CommittedFileChange>()
+  trackedFiles.forEach(x => files.set(x.path, x))
+  untrackedFiles.forEach(x => files.set(x.path, x))
+  return [...files.values()].sort((x, y) => compare(x.path, y.path))
+}
+
+/**Same thing as `getChangedFiles` but with extra handling for 128 exit code
```
iAmWillShepherd:
```suggestion
/**
* Same thing as `getChangedFiles` but with extra handling for 128 exit code
```

app/src/lib/git/stash.ts
outofambit: yeah i think that's a good idea @say25!

tierninho:
LGTM ⚡️

Issue #7341, opened 2019-04-18
[rebasing] Premature force push warning?


tierninho:
During a rebase we show a force push warning for the user in anticipation of the Force Push button appearing. Under the hood are we still force pushing if the rebase was successful? 

I still see this warning even though the success messaging below is shown. The Force Push button is never shown in this case, so I am not sure if this is intended or not.

<img width="611" alt="Screen Shot 2019-04-17 at 3 37 52 PM" src="https://user-images.githubusercontent.com/14828183/56331236-3017dd80-6127-11e9-98ef-0af386a397da.png">
<img width="817" alt="Screen Shot 2019-04-17 at 3 37 59 PM" src="https://user-images.githubusercontent.com/14828183/56331235-3017dd80-6127-11e9-8ee1-9b6a0dfe343a.png">

Secondly, when the user actually does click the Force Push button (if pref enabled), we show another warning like so. Curious, do we really need both warnings? 

<img width="645" alt="Screen Shot 2019-04-17 at 3 49 23 PM" src="https://user-images.githubusercontent.com/14828183/56331569-918c7c00-6128-11e9-8790-22a37a377950.png">

1.7.0-beta4, mac

billygriffin:
@tierninho I'm not sure how to answer the first question (cc: @shiftkey), but for the second one, that's intentional for now because they're different kinds of problems (one is getting into a potentially tricky spot for yourself locally, and the other is potentially messing up the bigger repo for everyone on the team). It's a matter of who it impacts.

shiftkey:
I think you're interested in this TODO comment, which I deferred from the initial beta:

https://github.com/desktop/desktop/blob/ab7b0e9bfedd27834fa2fd3b8f49f11c0f8dcafc/app/src/lib/stores/app-store.ts#L3922-L3924

Consider this scenario:

 - you rebase a branch cleanly onto the base branch
 - you try to do the exact same thing again

There's no work to do here for a rebase, but we don't perform this check. I think implementing that check and blocking the user from proceeding is the answer here.

Can you confirm that was the steps you were testing?

shiftkey:
> Curious, do we really need both warnings?

I'm inclined to keep them both around for the moment, because they feel like distinct cases:

 - an up-front warning before you dive into the rebase of the downstream impact to others
 - a warning before pushing to give you a chance to back out

tierninho:
> Consider this scenario:
you rebase a branch cleanly onto the base branch
you try to do the exact same thing again

[...]

PR #7343, opened 2019-04-18
check if the target branch has already been rebased 

shiftkey:
## Overview

**Related to #7341**

## Description

This PR addresses a scenario I had in mind when previewing the rebase: "what if there's no work to do?"

To test this out, I check for the merge base between the target branch and the base branch. If the merge base is the same as the base branch, this means the target branch is directly accessible from the base branch - so the rebase is a no-op.

The latest beta shows this message in the footer:

With this PR, it now shows this (and prevents the rebase from proceeding):

I'm not 💯% sold on the words yet, but I mostly wanted to open this up to confirm it's the same scenario from #7341 that would avoid the confusion about an unnecessary rebase...

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3917,20 +3917,31 @@ export class AppStore extends TypedBaseStore<IAppState> {
       targetBranch.tip.sha
     )
 
-    // TODO: in what situations might this not be possible to compute
+    // TODO: in what situations might this not be possible to compute?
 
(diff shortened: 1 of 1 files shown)


tierninho:
> To test this out, I check for the merge base between the target branch and the base branch. If the merge base is the same as the base branch, this means the target branch is directly accessible from the base branch - so the rebase is a no-op.

Tested this scenario, and worked as expected. 👍

PR #7362, opened 2019-04-19
[Stashing] Confirm before discarding stash

### LINKED ISSUE #7348: [Stashing] Add confirm dialog when user clicks “Discard” to discard a stash
First draft of verbage:
```
Are you sure you want to discard these stashed changes?
They cannot be recovered once discarded.

                                     [Discard] [Cancel]
```

_note: we should use ButtonGroup to make sure these buttons are ordered correctly on each OS._

iAmWillShepherd:
Fixes #7348 

Adds a new popup type and dialog to confirm dropping a stash entry.

as of https://github.com/desktop/desktop/pull/7362/commits/9fd383e28334744325e29c1986f0816cabb64713 the popup is now (on Mac):

Screen Shot 2019-04-24 at 2 28 48 PM

I decided not to include the line, "They cannot be recovered once discarded.", since that is not entirely accurate.

cc/ @outofambit and @billygriffin for thoughts.


tierninho:
Should we make this a preference as we do so with discarding changes?

tierninho:
Tested and LGTM ⚡️

app/src/ui/stashing/confirm-discard-stash.tsx
```diff
@@ -0,0 +1,78 @@
...
+        disabled={this.state.isDiscarding}
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+      >
+        <DialogContent>
+          <Row>Are you sure you want to discard these stashed changes?</Row>
+        </DialogContent>
+        <DialogFooter>
+          <ButtonGroup>
+            <Button type="submit">Continue</Button>
```
outofambit: I'd still like the buttons to say "Discard" and "Cancel". Does that seem okay to you?

```suggestion
            <Button type="submit">Discard</Button>
```

app/src/ui/stashing/confirm-discard-stash.tsx
```diff
@@ -0,0 +1,78 @@
+import React = require('react')
```
shiftkey: :art: my favourite suggestion

```suggestion
import * as React from 'react'
```

app/src/ui/stashing/confirm-discard-stash.tsx
```diff
@@ -0,0 +1,78 @@
...
+  readonly onDismissed: () => void
+}
+
+interface IDiscardStashState {
+  readonly isDiscarding: boolean
+}
+/**
+ * Dialog to confirm dropping a stash
+ */
+export class DiscardStash extends React.Component<
```
shiftkey: :art: we use the `Confirm` prefix in other places to make the intent clearer. Also, maybe make clear this is rendering a dialog?

```suggestion
export class ConfirmDiscardStashDialog extends React.Component<
```

PS This will break a bunch of code, only take it in if you really want to do that 😁

app/src/ui/stashing/confirm-discard-stash.tsx
outofambit: oh yeah we should do that

app/src/ui/stashing/confirm-discard-stash.tsx
```diff
@@ -0,0 +1,78 @@
...
+        disabled={this.state.isDiscarding}
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+      >
+        <DialogContent>
+          <Row>Are you sure you want to discard these stashed changes?</Row>
+        </DialogContent>
+        <DialogFooter>
+          <ButtonGroup>
+            <Button type="submit">Discard</Button>
```
billygriffin: I think the type="submit" should be on the `Cancel` button to make that the blue default button on macOS, shouldn't it?
Screen Shot 2019-04-23 at 8 15 04 AM

outofambit: 9fd383e

app/src/ui/stashing/confirm-discard-stash.tsx
outofambit: 0bb3b1f

outofambit: also f324e60

PR #7363, opened 2019-04-19
[Stashing] Adds a bit of explanatory text for restore button

### LINKED ISSUE #7303: [Stashing] Add info verbiage next to restore button in stash diff UI
Add a bit of explanatory text next to the _Restore_ button to inform the user of what it means.

See https://github.com/desktop/desktop/issues/6987#issuecomment-472942310 for context.

There are 2 scenarios that need to be handled:
1. No uncommitted changes
    * _Discard_ and _Restore_ buttons are both enabled
    * _Restore_ button will have the following text next to it: "This will pop this stash as changed files."
2. Uncommitted changes exist
    * _Discard_ is enabled,  _Restore_ is **disabled**
    * _Restore_ button has an :exclamation: icon with the text: "Unable to restore stash when changes are present on your branch."

Some things worth considering:
* For scenario 1, depending on the button order issue (https://github.com/desktop/desktop/pull/7301), we may need to rethink the helper text because if the buttons are reversed the text is no longer next to the "restore" button and it's confusing.

UI

☝️ This image uses the verb _apply_, but we will instead be going with _pop_; the info text in the image will not actually be used, it only serves as a visual aide.

iAmWillShepherd:
Fixes #7303 

~This PR is effected by the outcome of https://github.com/desktop/desktop/issues/7273.~

**No uncommitted changes**

Screen Shot 2019-04-24 at 12 51 46 PM

**Some uncommitted changes**

Screen Shot 2019-04-24 at 12 51 03 PM

cc/ @billygriffin, @ampinsk, @niik for design thoughts.


iAmWillShepherd:
We could also ditch the explanatory text in the header and have it appear on hover as a tool tip. It's less discoverable, but we wouldn't have to worry about how it's placed in the UI. I think it's a fine tradeoff considering the messaging is intended to clarify what the _Restore_ button does (or cannot do in the case of uncommitted changes being present).

billygriffin:
I would defer to @ampinsk's original design unless it's significantly more difficult to implement. I don't think we should just compromise for ease of implementation unless the hurdle is really high or presents other problems.

tierninho:
This is awkward sounding ""This will pop this stash as changed files." Also, not sure what button is being referenced as it is just floating there and rather vague.

Suggest: "Restore to move your stashed files to the Changes list."

app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -93,22 +107,43 @@ const Header: React.SFC<{
...
+    dispatcher.dropStash(repository, stashEntry)
   }
+
   const onSubmitClick = () => {
-    props.dispatcher.popStash(props.repository, props.stashEntry)
+    dispatcher.popStash(repository, stashEntry)
   }
+
+  const restoreMessage = isWorkingTreeClean ? (
+    'This will pop this stash as changed files.'
```
billygriffin:
```suggestion
    'Restore will move your stashed files to the Changes list.'
```

billygriffin: This is based on @tierninho's suggestion after discussing it in our weekly meeting.

outofambit:
cross-posting this from the original issue for visibility :)
> @ampinsk and I would like to keep the layout of the original design (regardless of button order) as long as we use the text "Restore will move your stashed files to the Changes list."

tierninho:
Tested this and ready to go 🚢

app/styles/ui/_stash-diff-viewer.scss
```diff
@@ -6,23 +6,50 @@
...
+        flex-grow: 1;
+      }
+
+      .octicon {
+        fill: var(--text-secondary-color);
+        margin: 0 var(--spacing-half);
+      }
+
+      .octicon {
+        flex-shrink: 0;
```
niik: Could we move this up into the previous `.octicon` selector?

outofambit: lol yeah that's a very reasonable request! thanks for catching that. 73c86d6

PR #7364, opened 2019-04-19
Update confirm dialog when overwriting a previous stash

### LINKED ISSUE #7361: [stashing] Update stash warning language
Based on user feedback here: https://github.com/desktop/desktop/issues/6107#issuecomment-484919607 we should consider updating the language used in the buttons and text.

56428862-e3e7ad00-628e-11e9-9463-e214c64cbfa4

At minimum, change "clear" to "discard" to match the new button name.

Feedback:

> The wording on that might be a little confusing. When I skim it, I see "Clear or restore...", along with a continue button, which can look like it's guiding me somewhere to fix things, rather than deleting the old stash. It makes sense as a whole, but the main things I looked at weren't clearly destructive.

and

> But if it stays as is, the buttons should be something like Cancel and Overwrite.

1.7.0beta4, mac

billygriffin:
**Closes #7361**

This PR updates the confirm dialog to use clearer verbiage, but isn't complete and I'd love some help getting it all the way there.

I wasn't able to make the button coloring work correctly on macOS, and I'd love a gutcheck on how this works on Windows as well.

The "Cancel" button is the default button (if you click `Enter` or `Spacebar` it cancels, which is desired), but it's not blue as the other dialogs are. I tried to pattern match other similar dialogs by switching the `type="submit"` to the Cancel button, but that made the default action Overwrite, so I'm not sure how to fix that.

I also added the destructive boolean to the button group as it seems like that should've been present for overwriting a stash.

Screen Shot 2019-04-19 at 4 03 30 PM

## Release notes

No notes


billygriffin:
This failed lint because of the ordering of the button groups (the primary button is supposed to be first), but as I mentioned above I couldn't get Cancel to be the primary button and also be the button invoked upon clicking `Enter`, so I'd love some help on that - absolutely no rush.

tierninho:
Tested and looks good.

PR #7377, opened 2019-04-22
lock file state when in rebase conflicts

### LINKED ISSUE #7006: restrict checkbox state when dealing with rebase conflicts
This was raised by @tierninho in https://github.com/desktop/desktop/pull/6984#issuecomment-468805303:

> if you navigate to the Changes tab and uncheck the changes, the Continue rebase button is still active. I expect to be be disabled until the user checks the changed file(s).

While the "Continue rebase" state is valid feedback here, the underlying issue is that users can uncheck files that need to be handled as part of resolving conflicts. This doesn't actually affect the rebase (or merge) action, which is applied to all tracked files and ignores the checkbox status.

What I'd like to do to tighten up this behaviour is to lock down the checkbox behaviour:

 - for each tracked file, mark the file as **checked**, disable the checkbox and add a helpful tooltip
 - for each untracked file, mark the file as **unchecked** by default, leave the checkbox enabled (so the user can add the file to the rebase if they want?) and add a helpful tooltip

shiftkey:
## Overview

**Closes #7006**

## Description

This PR updates the `ChangesList` component to lock down the checkboxes when rebase conflicts are detected.

Currently you can change all the checkboxes in the list, but these choices are ignored because the rebase flow needs to stage all tracked files (untracked files are ignored by `git` when rebasing) to continue correctly:

With this PR, we now:

 - disable all checkboxes while you are dealing with rebase conflicts
 - lock the checkbox state to include tracked files, and exclude untracked files

This is how the underlying `git rebase` works by default, and Desktop needs to follow this convention when staging files before continuing the rebase.

TODO: 
 - [x] lock down checkbox state
 - [x] warn when untracked files exist
 - [x] show a subset of the context menu options when in the middle of a rebase

## Release notes

Notes: `[Improved] Changes list state better reflects rebase flow when handling conflicts`

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -321,6 +321,36 @@ export class ChangesList extends React.Component<
     }
   }
 
+  private getRevealInFileManagerMenuItem = (
+    file: WorkingDirectoryFileChange
+  ): IMenuItem => {
+    return {
+      label: RevealInFileManagerLabel,
+      action: () => revealInFileManager(this.props.repository, file.path),
+      enabled: file.status.kind !== AppFileStatusKind.Deleted,
+    }
+  }
+
+  private getOpenInExternalEditorMenuItem = (
+    file: WorkingDirectoryFileChange,
+    enabled: boolean
+  ): IMenuItem => {
+    const { externalEditorLabel, repository } = this.props
+
+    const openInExternalEditor = externalEditorLabel
+      ? `Open in ${externalEditorLabel}`
+      : DefaultEditorLabel
+
+    return {
+      label: openInExternalEditor,
+      action: () => {
+        const fullPath = Path.join(repository.path, file.path)
+        this.props.onOpenInExternalEditor(fullPath)
+      },
+      enabled,
+    }
+  }
+
   private getDefaultContextMenu(
     file: WorkingDirectoryFileChange
   ): ReadonlyArray<IMenuItem> {
@@ -329,16 +359,7 @@ export class ChangesList extends React.Component<
     const extension = Path.extname(path)
     const isSafeExtension = isSafeFileExtension(extension)
 
-    const {
-      workingDirectory,
-      externalEditorLabel,
-      selectedFileIDs,
-      repository,
-    } = this.props
-
-    const openInExternalEditor = externalEditorLabel
-      ? `Open in ${externalEditorLabel}`
-      : DefaultEditorLabel
+    const { workingDirectory, selectedFileIDs } = this.props
 
     const selectedFiles = new Array<WorkingDirectoryFileChange>()
     const paths = new Array<string>()
@@ -420,19 +441,8 @@ export class ChangesList extends React.Component<
     items.push(
       { type: 'separator' },
       this.getCopyPathMenuItem(file),
-      {
-        label: RevealInFileManagerLabel,
-        action: () => revealInFileManager(repository, path),
-        enabled: status.kind !== AppFileStatusKind.Deleted,
-      },
-      {
-        label: openInExternalEditor,
-        action: () => {
-          const fullPath = Path.join(repository.path, path)
-          this.props.onOpenInExternalEditor(fullPath)
-        },
-        enabled,
-      },
+      this.getRevealInFileManagerMenuItem(file),
+      this.getOpenInExternalEditorMenuItem(file, enabled),
       {
         label: OpenWithDefaultProgramLabel,
         action: () => this.props.onOpenItem(path),
@@ -451,12 +461,6 @@ export class ChangesList extends React.Component<
     const extension = Path.extname(path)
     const isSafeExtension = isSafeFileExtension(extension)
 
-    const { externalEditorLabel, repository } = this.props
-
-    const openInExternalEditor = externalEditorLabel
-      ? `Open in ${externalEditorLabel}`
-      : DefaultEditorLabel
-
     const items = new Array<IMenuItem>()
 
     if (file.status.kind === AppFileStatusKind.Untracked) {
@@ -474,19 +478,8 @@ export class ChangesList extends React.Component<
 
     items.push(
       this.getCopyPathMenuItem(file),
-      {
-        label: RevealInFileManagerLabel,
-        action: () => revealInFileManager(repository, path),
-        enabled: status.kind !== AppFileStatusKind.Deleted,
-      },
-      {
-        label: openInExternalEditor,
-        action: () => {
-          const fullPath = Path.join(repository.path, path)
-          this.props.onOpenInExternalEditor(fullPath)
-        },
-        enabled,
-      },
+      this.getRevealInFileManagerMenuItem(file),
+      this.getOpenInExternalEditorMenuItem(file, enabled),
       {
         label: OpenWithDefaultProgramLabel,
         action: () => this.props.onOpenItem(path),
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -186,7 +186,15 @@ export class ChangesList extends React.Component<
   }
 
   private renderRow = (row: number): JSX.Element => {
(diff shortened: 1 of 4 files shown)


billygriffin:
This is great! The only thing I'm curious about is whether people might be confused by the unchecked portion and not being able to do anything about it.

shiftkey:
> The only thing I'm curious about is whether people might be confused by the unchecked portion and not being able to do anything about it.

We could use the commit form to provide more context:



Maybe an ⚠️ icon above the button alongside a message like `Untracked files excluded from rebase` when they're in this situation? I can noodle on how this would look if this feels on the right track.

tierninho:
Tested the functionality and LGTM.

shiftkey:
Dropping `ready-to-review` while I implement this suggested change https://github.com/desktop/desktop/pull/7377#issuecomment-485450532

donokuda:
> Maybe an ⚠️ icon above the button alongside a message like Untracked files excluded from rebase when they're in this situation? I can noodle on how this would look if this feels on the right track.

The ⚠️icon might feel like there's an issue that a user needs to fix in order to move on (like a merge conflict.) Maybe using an [ℹ️icon](https://octicons.github.com/icon/info/) or something to visually associate it more closer to the file it's referencing (for example, like the [diff-added octicon](https://octicons.github.com/icon/diff-added/)):



Another consideration is to hide the checkboxes for all files since it seems like we wouldn't want users to interact with the list at all.

shiftkey:
@donokuda that's a good shout - I'll try it out!

> Another consideration is to hide the checkboxes for all files since it seems like we wouldn't want users to interact with the list at all.

I worry about removing the checkboxes because their state is still helpful to indicate what will be rebased and what will be skipped, and for any unexpected side-effects in the DOM (we've not tried this before).

donokuda:
Thoughts on going even _more_ minimal and starting off with no octicon and putting the text below the button? After looking at my original proposal again, I'm not really sold on how it visually looks with the added icon.



If we find out that it's confusing to a lot of people, we can go back to the drawing board.

> I worry about removing the checkboxes because their state is still helpful to indicate what will be rebased and what will be skipped, and for any unexpected side-effects in the DOM (we've not tried this before).

Makes sense, I'd rather keep this as simple as possible than adding more scope without realizing the consequences of it.

billygriffin:
I'm a big fan of minimal, especially since my concern might be completely unfounded and people will just be like, "yeah, of course it works that way."

shiftkey:
@donokuda I dig the mininalism. Let me push some changes to this branch and get a working demo together...

shiftkey:
Latest screenshots - warning when there are untracked files:



After removing the untracked change, the warning goes away:

billygriffin:
@shiftkey Looks good! Is it "Uncommitted" or "Untracked"? I assume you've got it right - just making sure.

shiftkey:
The last item on my list is what to do with the context menu that is shown:



My current thinking on this is that, when in the middle of a rebase conflict, these options should only be enabled:



We could also show "Discard Changes" if the file is untracked, as a convenient way to address the warning.

I need to do some internals refactoring to be able to change this behaviour, but let me know if you have suggestions about the list to show in this state.

shiftkey:
> Is it "Uncommitted" or "Untracked"? I assume you've got it right - just making sure.

"Uncommitted" is too vague for my liking in this situation. Git has two terms which are more explicit:

 - "Unstaged file" - changes to a tracked that have not been added to the index
 - "Untracked file" - a new files in the repository which is not yet tracked by Git

That's why I've settled on "Untracked" here...

shiftkey:
Here's the context menu entries for a tracked file in the middle of a rebase:



And here's what you see for an untracked file:

app/src/ui/changes/continue-rebase.tsx
```diff
@@ -46,8 +47,16 @@ export class ContinueRebase extends React.Component<IContinueRebaseProps, {}> {
 
     const loading = this.props.isCommitting ? <Loading /> : undefined
 
+    const warnAboutUncommittedFiles = this.props.hasUntrackedChanges ? (
```
billygriffin:
```suggestion
    const warnAboutUntrackedFiles = this.props.hasUntrackedChanges ? (
```

app/src/ui/changes/continue-rebase.tsx
```diff
@@ -46,8 +47,16 @@ export class ContinueRebase extends React.Component<IContinueRebaseProps, {}> {
...
+      <div className="warning-uncommitted-files">
```
billygriffin:
```suggestion
      <div className="warning-untracked-files">
```

app/src/ui/changes/continue-rebase.tsx
```diff
@@ -58,6 +67,8 @@ export class ContinueRebase extends React.Component<IContinueRebaseProps, {}> {
           {loading}
           <span>{loading !== undefined ? 'Rebasing' : 'Continue rebase'}</span>
         </Button>
+
+        {warnAboutUncommittedFiles}
```
billygriffin:
```suggestion
        {warnAboutUntrackedFiles}
```

app/styles/ui/changes/_continue-rebase.scss
```diff
@@ -10,6 +10,11 @@
   background-color: var(--box-alt-background-color);
   padding: var(--spacing);
 
+  .warning-uncommitted-files {
```
billygriffin:
```suggestion
  .warning-untracked-files {
```

tierninho:
Tested and functionally working 👍  

> We could also show "Discard Changes" if the file is untracked, as a convenient way to address the warning.

Did catch one thing:
The right-click on the `n files changed` still shows the `Discard All Changes` option.

Screen Shot 2019-04-23 at 11 07 18 AM

3e3c288508, mac

iAmWillShepherd:
Holding off on final review until we decide if https://github.com/desktop/desktop/pulls?q=is%3Aopen+is%3Apr+no%3Aassignee+label%3Aready-for-review#issuecomment-485977147 should be addressed in this PR.

shiftkey:
> The right-click on the `n files changed` still shows the `Discard All Changes` option.

Nice find @tierninho. I feel bad for hiding it away, given how not discoverable it is. There's some extra context here that I'll write up into a fresh issue in the morning before I do disable it.

shiftkey:
> There's some extra context here that I'll write up into a fresh issue in the morning before I do disable it.

https://github.com/desktop/desktop/issues/7394

app/src/ui/changes/changes-list.tsx
```diff
@@ -57,6 +56,30 @@ const StashIcon = new OcticonSymbol(
 
 const GitIgnoreFileName = '.gitignore'
 
+/** Compute the 'Include All' checkbox value from the repository state */
+function getIncludeAllValue(
+  workingDirectory: WorkingDirectoryStatus,
+  rebaseConflictState: RebaseConflictState | null
+) {
+  if (rebaseConflictState !== null) {
+    // there should be changes to rebase if conflicts are detected
```
iAmWillShepherd: It's unclear to me how this comment relates to what's below. When I read this, I expected to see us look for files with a conflict marker, not files that are untracked.

shiftkey: Let me take a stab at a better comment here. The code in here only really cares about these two cases:

 - all files are tracked by Git -> all files are needed to rebase -> "Include All" is true
 - some untracked files found -> some files are needed for rebase -> "Include All" is mixed state

There's actually a third state here, now I think about it:

 - no changed files in working directory -> rebase will skip the current commit -> "Include All" is false

This is relatively minor because the checkbox is disabled, but I'd like to ensure this reflects reality while we're in here making changes...

shiftkey: Check out 0e2e5406b and see if that makes more sense

iAmWillShepherd:
Merged without updating since doing so would have been the 4th time.

PR #7378, opened 2019-04-22
experiment with using labels to track epics

shiftkey:
## Overview

This is a suggestion to use labels to better track and organize issues arising from our development of workflows in the app. 

## Description

With our recent work on rebase and stashing workflows, we've fallen into a habit of prefixing the issue title with the relevant workflow:

While this feels nice to help glance at the list of open issues, I think there's a couple of issues with it:

 - search - because search is text based, issues not related to the development of a feature but containing the same words can get mixed in with these tasks
 - scope - it's not easy to distinguish which tasks are in scope for the current feature, versus work that might be done in the future
 - triage - it's not clear whether each task has been appropriate triaged and verified that it should be "in scope"
 - consistency - we also seem to have the exact same pattern for pull requests related to a workflow

I've updated the docs to suggest a solution to this problem: **workflow labels**.

These labels should be created when development of a workflow is being scoped that could be considered complex (new or improvements to an existing one), so that tasks can be associated with it when they are created. When development is underway, bugs and feedback should also be assigned to this label to track additional work.

What would this give us that prefixing issue titles doesn't currently?

 - clarity about progress - we use milestones for near-term work, but tasks beyond that (particularly as features can span multiple releases) are harder to follow
 - scope - by using the label to indicate what is "in scope" we can identify work that could be done later (and dropped from the label), or work that isn't worth tracking currently (and could be closed)
 - granularity  - we currently use `enhancement` for "things that aren't bugs" and I think this would be a step in the direction of eventually deprecating the `enhancement` label in favour of something more useful

Thoughts?

TODO:
 - [x] gather feedback and iterate on this idea
 - [x] create labels for active workflows
 - [x] add links to new labels


outofambit:
YES. I've been thinking the similar things lately. Thanks for writing this up!!

shiftkey:
After some lively discussion with the team this morning (thanks all!) we've decided to generalize these labels so they can be used more broadly:

 - we settled on using `epic` here for the label group name, as it felt like the most fitting label to describe the work
 - we also identified that labels aren't the whole solution to tracking progress and priority for work, so we're going to experiment with using a project board for each epic 

I've updated the docs here to make this clearer, but any additional feedback you might have here would be greatly appreciated!

docs/process/labels.md
```diff
@@ -57,6 +57,23 @@ implementing features:
...
+project board should be used to organize and prioritize the work within an epic
+based on the upcoming release schedule.
+
+Each label should be removed from the repository when the project board is closed, so
+these labels do not become overused and add noise to the issue tracker.
+
+|                              | Label name     |  Description |
+| ---------------------------- | -------------- |  ----------- |
+| [:mag_right:][epic:rebase]   | `epic:rebase`  | Supporting rebase flows in the app |
+| [:mag_right:][epic:stashing] | `epic:stashing`| Supporting stashing uncommitted changes in the app |
```
outofambit: i'd like to try out sprinkling some emojis in our labels. how about something like this for stashing? i think it would really help with visual scanning.

```suggestion
| [:mag_right:][epic:stashing] | `📄 epic:stashing`| Supporting stashing uncommitted changes in the app |
```

shiftkey: Personally, I prefer colour to visually differentiate these from other labels. I can't imagine us having more than 2-3 of these on the go, but I'm open to other suggestions.

I'm curious about how this affects the URLs generated (they contain the label name) - how unfriendly does this change make them?

outofambit: makes sense, we could do both color and emoji. i might try out the url thing, but regardless, let's not consider this suggestion blocking in any sense. :)

outofambit (dismissed):

shiftkey:
I've updated this branch to add the two new labels for epics, and I've updated the rebase-related tasks to use them. @iAmWillShepherd @outofambit feel free to experiment with the `epic:stashing` label to help organize your outstanding issues.

iAmWillShepherd (dismissed):
💯

iAmWillShepherd:
Even though this is a non-functional change, I'm going to let this sit for a day or so so others have a chance to chime in.

outofambit:
@billygriffin @nerdneha are you 👍 on this?

docs/process/labels.md
```diff
@@ -57,6 +57,23 @@ implementing features:
 | [:mag_right:][user-research]      | `user-research`      | Issues that may benefit from user interviews, validations, and/or usability testing |
 | [:mag_right:][needs-design-input] | `needs-design-input` | Issues that require design input from the core team before the work can be started |
 
+### Epics
+
+As we work on new parts of the application, or refining existing workflows, we use
+these labels to track issues that arise as part of the development process.
```
nerdneha:
```suggestion
epic labels, ex: `epic:rebase` to track issues and PRs that are part of a common feature focus area, or epic.
```

nerdneha (changes requested):
I'm 👍 on the idea and absolutely love it! As a ⛏, I'd like for us to define epic within this area so people who aren't familiar with agile terminology can get the gist. I attempted to define it here but I'm not married to the definition or where I put it.

nerdneha (dismissed):

outofambit (dismissed):

docs/process/labels.md
```diff
@@ -57,6 +57,23 @@ implementing features:
 | [:mag_right:][user-research]      | `user-research`      | Issues that may benefit from user interviews, validations, and/or usability testing |
 | [:mag_right:][needs-design-input] | `needs-design-input` | Issues that require design input from the core team before the work can be started |
 
+### Epics
+
+As we work on new parts of the application, or refining existing workflows, we use
```
billygriffin:
```suggestion
As we work on new parts of the application or refine existing workflows, we use
```

billygriffin (dismissed):

billygriffin (approved):

outofambit (approved):

PR #7380, opened 2019-04-22
refactor status to detect conflicts from stash

### LINKED ISSUE #7350: [Stashing] Let Users View Stash Conflicts in Changes Tab
Currently popping a stash results in conflicts in your working directory that are improperly represented in the changes tab.

Screen Shot 2019-04-22 at 11 10 12 AM

(A file with conflict markers is marked as resolved and should instead be marked as conflicted.)

We need to fix the status parsing to properly detect and represent these conflicts.

outofambit:
## Overview

Closes #7350 

## Description

There doesn't seem to be a way to tell if you are in the middle of a conflict from popping a stash in git, except to inspect the current statuses of files in the working directory and look for `unmerged` files.

So, this PR adds logic to look for files in conflict, even when we're not in the middle of a rebase or a merge.

**Note: this doesn't address #7383**


app/src/lib/git/status.ts
```diff
@@ -196,7 +196,10 @@ export async function getStatus(
   const headers = parsed.filter(isStatusHeader)
   const entries = parsed.filter(isStatusEntry)
 
-  let conflictDetails: ConflictFilesDetails
+  let conflictDetails: ConflictFilesDetails = {
```
outofambit: a default value

app/src/lib/git/status.ts
```diff
@@ -207,7 +210,7 @@ export async function getStatus(
       mergeHeadFound,
       rebaseInternalState
     )
-  } else {
+  } else if (entries.length > 0) {
```
outofambit: added this little optimization so we don't run these new git commands as often

shiftkey: Can this comment be included in the source? It feels important enough to keep in mind the next time we're in here making changes...

outofambit: done in b055353 :)

shiftkey:
I plan to merge this in when CI passes to unblock other work

PR #7381, opened 2019-04-22
update node and python versions in versioning config

outofambit:
update version manager files to match changes in #7345

- node `10.15.3`
- python `2.7.16`


shiftkey:
@outofambit can we used relaxed ranges rather than pinning to specific (updated) versions? I'd love for these to be configured to "just download the latest Node 10" and "just download the latest Python 2" rather than just moving them to another fixed version...

outofambit:
> can we used relaxed ranges rather than pinning to specific (updated) versions?

@shiftkey i agree that would be ideal, but unfortunately [asdf doesn't support even slightly ambiguous versions](https://asdf-vm.com/#/core-configuration?id=tool-versions). however, nvm does so i've updated that to use the latest v10.

i'm also happy to remove `.tool-versions` from the repo since i'm pretty sure i'm the only one using it and can configure my environment with untracked config files. what do you think?

shiftkey:
> i'm also happy to remove `.tool-versions` from the repo since i'm pretty sure i'm the only one using it and can configure my environment with untracked config files. what do you think?

@outofambit I'm fine with leaving it in for the moment - the docs talk about how to work with multiple versions of Python, so I still think there's value in keeping this around. It just feels tedious that these files and the docs related to them need to be kept in sync.

PR #7382, opened 2019-04-22
Add "View stash entry" action to suggested next steps

### LINKED ISSUE #7353: [Stashing] Suggested next step when stash exists on your branch

When you come back to a branch that you've stashed changes on, the stashed changes icon is unobtrusive and not super discoverable. We'd like a way that is non-blocking to increase the discoverability of the stash so people remember it's there.

The suggested next step seems like a good first step, though we may iterate to move to something more organic. The flow is described by @donokuda [here](https://github.com/desktop/desktop/issues/6987#issuecomment-472658704), and the verbiage is as follows:

- You have [number of files] changes in progress that you have not yet committed
- Second lighter line: When a stash exists, access it at the bottom of the `Changes` tab to the left
- Button text: "View stash"

The stash suggested next step should take prominence over any conflicting suggested next steps. For example, if there are commits to pull, but there's also a stash present, the blue suggestion should be the stash suggestion until there is no longer a stash, at which point the other suggestion would surface as the primary one.

iAmWillShepherd:
Fixes #7353

This PR adds
* suggested next step to view stash entry that takes precedence over all other actions and
* menu item to toggle the visibility of the stash diff UI 
  * <kbd>Alt+1</kbd> was chosen as the shortcut because _View -> Show Changes_ is <kbd>Cmd+1</kbd> and that could potentially conflict with the shortcut on Windows (I am open to better alternatives here).

☝️ was made before I changed the shortcut from <kbd>Ctrl+1</kbd> to <kbd>Alt+1</kbd>

Todo:
- [x] Add "View stashed changes" action to next steps
- [x] Add "View stashed changes" menu item to top-level menu
- [x] Update store to change view to stashed changes diff UI
- [x] Update menu item to "Hide stashed changes" when they are visible


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -165,6 +165,12 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+2',
         click: emit('show-history'),
       },
+      {
+        label: __DARWIN__ ? 'Show Stashed Changes' : 'Sho&w stashed changes',
+        id: 'show-stashed-changes',
+        accelerator: 'Ctrl+1',
```
iAmWillShepherd: I'm open to suggestions around this keyboard shortcut.

outofambit: could we do `Ctrl+Shift+T`? I want to keep our number shortcuts open for being linked to recent branches or repositories.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -165,6 +167,14 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+2',
         click: emit('show-history'),
       },
+      {
+        label: getStashedChangesLabel(isStashedChangesVisible),
+        id: 'toggle-stashed-changes',
+        accelerator: 'Alt+1',
```
shiftkey: The more I think about this menu item, the more it feels like "Go to Summary" which is in the next section. 

"Go to Summary" has an accelerator `CTRL+G`, which means "Show stashed changes"  could have something like `CTRL+H` (I'd have to scan for any clashes with this, but you get the idea).

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -512,6 +522,14 @@ function getPushLabel(
   return __DARWIN__ ? 'Force Push' : 'Force P&ush'
 }
 
+function getStashedChangesLabel(isStashedChangesVisible: boolean): string {
+  if (isStashedChangesVisible) {
+    return __DARWIN__ ? 'Hide Stashed Changes' : 'H&ide stashed'
```
shiftkey: Is there a reason we've dropped this last word?

```suggestion
    return __DARWIN__ ? 'Hide Stashed Changes' : 'H&ide stashed changes'
```

tierninho:
Whatever shortcut we use, I suggest updating the discoverability line from: "When a stash exists, access it at the bottom of the Changes tab to the left." to "Access your stashed changes at the bottom of the Changes tab or [insert shortcut]".

tierninho:
Tested and looks good. 

Curious why we are using "Hide" now as we used Show to toggle the two views. Do we need to update some of the other menu items now?

5069415ed1, mac

shiftkey:
> Do we need to update some of the other menu items now?

Potentially. The current menu items `Show Branches` and `Show Repository List` aren't aware of their visibility and the need to "hide" but maybe they should?

app/src/main-process/menu/build-default-menu.ts
iAmWillShepherd: I would say fat-fingered, but my fingers are built more for poking.

app/src/main-process/menu/build-default-menu.ts
iAmWillShepherd: Changed to <kbd>Ctrl+H</kbd>

app/src/ui/changes/no-changes.tsx
```diff
@@ -338,6 +339,59 @@ export class NoChanges extends React.Component<
     return null
   }
 
+  private renderStashAction() {
```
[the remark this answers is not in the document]
iAmWillShepherd: Also feature flagged top-level menu item.

app/src/lib/stores/app-store.ts
```diff
@@ -2011,6 +2013,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }))
     this.emitUpdate()
 
+    this._hideStashedChanges(repository)
```
shiftkey: Can this be moved inside the  `if (selectedSection === RepositorySectionTab.Changes) {`? I don't think it needs to run every time this is called...

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -165,6 +167,15 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+2',
         click: emit('show-history'),
       },
+      {
+        label: getStashedChangesLabel(isStashedChangesVisible),
```
shiftkey: Do you think this could be moved down below the `Go to Summary` command in the list? 

My thinking here is that this group of menu items feels like it's accessed fairly regularly, while stashed changes would be used less frequently, so it doesn't make sense to have it in the middle of this group.

app/src/ui/changes/no-changes.tsx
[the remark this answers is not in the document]
shiftkey: I don't think this was feature-flagged correctly, and instead a part inside of `app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx` (which is probably already covered by some other code path) was updated. Does that sound right?

app/src/lib/stores/app-store.ts
iAmWillShepherd: We actually want to do this when the user chooses to view changes **and** view history, so I think we do need to hide it every time..

iAmWillShepherd:
@shiftkey 

> I don't think this was feature-flagged correctly, and instead a part inside of app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx (which is probably already covered by some other code path) was updated. Does that sound right?

I don't quite understand what you mean. Are you saying we don't need a feature flag in this spot because it _should be_ impossible to have a stash associated with a branch if `StashAndSwitchBranch` is behind a feature flag?

app/src/main-process/menu/build-default-menu.ts
iAmWillShepherd: Ooops, I saw this after I changed it to <kbd>Ctrl+H</kbd>

https://github.com/desktop/desktop/pull/7382#discussion_r277876252

If you prefer <kbd>Ctrl+Shift+T</kbd>, leave a new comment in the main PR to make it more discoverable and easier to discuss.

shiftkey:
@iAmWillShepherd I was referring to the change inside `app/src/ui/changes/no-changes.tsx` where we added `renderStashAction()` to be rendered before `renderRemoteAction()`

https://github.com/desktop/desktop/blob/82f6b7984fd67dcafb155e61230304f3cd090456/app/src/ui/changes/no-changes.tsx#L624

If we were to disable the feature flag but still have a stash in the repository, I was worried that this prompt might still be rendered but not be supported by the app. I tested this theory but it looks like the stash is not stored in app state, so it exits early and returns `null`:

https://github.com/desktop/desktop/blob/48d4cf592a78e05c20fbef04c67f65fe5c129096/app/src/ui/changes/no-changes.tsx#L350-L353

I don't think there's a need to add another flag in here - just me being a bit paranoid about unexpected UI changes bleeding through before we're ready.

iAmWillShepherd:
> I was referring to the change inside app/src/ui/changes/no-changes.tsx where we added renderStashAction() to be rendered before renderRemoteAction()

I see. I can see how that's non-obvious since I was relying on truthiness to conditionally render the result of the RHS operand.

> I don't think there's a need to add another flag in here - just me being a bit paranoid about unexpected UI changes bleeding through before we're ready.

Should I remove the feature flag I added inside of `NoChanges.renderStashAction`?

shiftkey:
> Should I remove the feature flag I added inside of `NoChanges.renderStashAction`?

That's the thing I'm confused about - I only see this feature flag being imported:

https://github.com/desktop/desktop/blob/48d4cf592a78e05c20fbef04c67f65fe5c129096/app/src/ui/changes/no-changes.tsx#L7

iAmWillShepherd:
> That's the thing I'm confused about - I only see this feature flag being imported:

Ah! I somehow forgot to commit or I discarded it while I was cleaning up. 

Added in https://github.com/desktop/desktop/pull/7382/commits/8015f2dc412a075428ccce5b412dd80ed172d50d

tierninho:
Found an inconsistency where the Show/Hide is one step behind the user.

Reproduce:
- Create a stash and make sure the suggested next step is in view
- Click the Stashed Changes "button"
- `Ctrl-H` to go back to suggested next step

Expected result: User goes back to suggested next step

Actual result: Screen flashes as user is really just going to stashed changes diffs

You can even reserve this and `crtl-H` does "nothing". Just start step 1 on the stashed diff page. 

8015f2dc41, mac

app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
```diff
@@ -94,6 +95,10 @@ export class StashAndSwitchBranch extends React.Component<
   }
 
   private renderStashActions() {
+    if (!enableStashing()) {
```
shiftkey: :art: I don't think this is required, because the two places where we launch this popup are already guarded by the feature flag:

https://github.com/desktop/desktop/blob/8015f2dc412a075428ccce5b412dd80ed172d50d/app/src/lib/stores/app-store.ts#L2781-L2795

https://github.com/desktop/desktop/blob/8015f2dc412a075428ccce5b412dd80ed172d50d/app/src/lib/stores/app-store.ts#L2849-L2863

Feel free to tidy this up to keep the diff readable

iAmWillShepherd:
> I haven't been able to reproduce the issue that @tierninho reported, but otherwise this is feeling close.
> 
> One suggestion to tidy up.

I was able to reproduce. The issue turned out to be related to the "Button" using a different function to show the UI and the menu item. This was fixed by getting rid of the duplicate function. I had to keep `AppStore._hideStashedChanges` around because it should be hidden when the user comes back to the _Changes_ tab from the _History_ tab.

iAmWillShepherd:
@tierninho can you give this another pass to verify it's no longer an issue. I was unable to reproduce this after 1be3984.

app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
iAmWillShepherd: Removed.

shiftkey:
Waiting on @tierninho to confirm he's happy with the recent changes, then I'll check in with @iAmWillShepherd and see when we should merge this to help with the beta tomorrow.

tierninho:
Nice, seems fixed now. ⚡️

PR #7391, opened 2019-04-24
rebase flow - update copy when choosing branch

shiftkey:
## Overview

This PR is an attempt to simplify the words shown when choosing the base branch to use in the rebase. The main thread of feedback we got at the time was a nervousness about the base branch (the one you are choosing) being modified or updated, and I wanted to eliminate this confusion.

## Description

This is the current view you'll see after choosing a branch:

After plenty of thought, I settled on trying to strip back the words to reduce the confusion:

The key goals I settled on:

 - be less wordy - focus on what will happen
 - remove the suggestions that the base branch is affected by this - "onto" -> "on top of", or omitted
 - reduce the mentions of the base branch where possible - emphasize the current branch

I'd love your thoughts on how this changes the information about starting the rebase.

cc @feministy @zeke as you raised feedback in this vein - I'd love to hear if this feels better.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/rebase/choose-branch.tsx
+++ b/app/src/ui/rebase/choose-branch.tsx
@@ -158,7 +158,7 @@ export class ChooseBranchDialog extends React.Component<
         dismissable={true}
         title={
           <>
-            Rebase <strong>{truncatedCurrentBranchName}</strong> onto…
+            Rebase <strong>{truncatedCurrentBranchName}</strong>…
           </>
         }
       >
@@ -180,8 +180,7 @@ export class ChooseBranchDialog extends React.Component<
           {this.renderRebaseStatus()}
           <ButtonGroup>
             <Button type="submit" disabled={disabled} tooltip={tooltip}>
-              Rebase <strong>{currentBranchName}</strong> onto{' '}
-              <strong>{selectedBranch ? selectedBranch.name : ''}</strong>
+              Rebase <strong>{currentBranchName}</strong>
             </Button>
           </ButtonGroup>
         </DialogFooter>
@@ -264,11 +263,9 @@ export class ChooseBranchDialog extends React.Component<
     const pluralized = commitsToRebase === 1 ? 'commit' : 'commits'
     return (
       <>
-        This will rebase
+        This will apply
         <strong>{` ${commitsToRebase} ${pluralized}`}</strong>
-        {` from `}
-        <strong>{currentBranch.name}</strong>
-        {` onto `}
+        {` on top of `}
         <strong>{baseBranch.name}</strong>
       </>
     )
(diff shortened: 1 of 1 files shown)


nerdneha:
> When I click the "Rebase linux" button, it's the same as if I'm on the development branch in the command line and type git rebase linux, right?

Well, the helper text says it’s going on top of `development`. So it’s probably equivalent to `git rebase development` but that’s why it’s confusing. I’m definitely seeing the information being conflicting. Especially if anyone has operated on the command line before.

billygriffin:
@shiftkey Got it, ugh. I do think we need to avoid doing the opposite of what the user would expect based on the git command in the CLI.

What about what @nerdneha suggested for the helper text:

> This will update `linux` by rebasing `27 commits` from `linux` onto `development`

Along with the button that just says "Start rebase" and the title text of the modal is "Choose a branch to rebase"

I think this introduces less confusion and makes clear which branch is being updated, and it aligns with your current branch, which is what I think most people expect (and aligns with the feedback we've heard from @zeke and @feministy)

shiftkey:
@billygriffin 

> What about what @nerdneha suggested for the helper text:

>> This will update linux by rebasing 27 commits from linux onto development

This is close, but I disliked "onto" here because there was some confusion about whether `development` would receive these commits. That's why I switched to `on top of` to try and better emphasize the separation.

I also wanted to use `apply` rather than `rebasing` because that felt closer to how Git itself [describes the command in it's docs](https://git-scm.com/docs/git-rebase):



What about this?

> This will update **linux** by applying **27 commits** from linux on top of **development**

billygriffin:
@shiftkey I like it - your rationale on those two things makes sense to me and I don't think it adds confusion.

shiftkey:
I've updated the PR description with the latest screenshot, but here it also is for reference:

nerdneha:
I feel like we're already short on space right now but I'd love to add "its" in front of 27 commits, ex:
```markdown
This will update **linux** by applying its **27 commits** on top of **development**
```
I got confused for a second and couldn't understand where the 27 commits were coming from and then I remembered it must be from `linux`. I agree that using "from linux" can be redundant but maybe we can at least indicate ownership with the word "its"?

shiftkey:
> I agree that using "from linux" can be redundant but maybe we can at least indicate ownership with the word "its"?

Updated in 7151858

feministy:
I think this is definitely better than what was there before, but I do still find it confusing 🙁 - this might be because I don't rebase that much anymore though, so take that with a grain of salt!

outofambit:
i'm going to merge this in preparation for beta tomorrow!

zeke:
Late to the party, but I approve of the new wording. 👍 

> This will update `linux` by applying its 32 commits on top of `development`

PR #7392, opened 2019-04-24
fix stash diff viewer button order (for mac)

### LINKED ISSUE #7273: [Stashing] macOS button order is incorrect

outofambit:
fixes #7273

Windows (i spoofed windows in code locally):
Screen Shot 2019-04-24 at 2 10 34 PM

Mac:
Screen Shot 2019-04-24 at 2 09 28 PM

### DIFF
--- a/app/src/ui/stashing/stash-diff-viewer.tsx
+++ b/app/src/ui/stashing/stash-diff-viewer.tsx
@@ -139,14 +139,14 @@ const Header: React.SFC<{
   return (
     <div className="header">
       <h3>Stashed changes</h3>
-      <ButtonGroup destructive={false}>
       <div className="row">
-        <Button onClick={onRestoreClick} type="submit">
-          Restore
-        </Button>
-        <Button onClick={onDiscardClick}>Discard</Button>
-      </ButtonGroup>
-      <div className="explanatory-text">{restoreMessage}</div>
+        <ButtonGroup destructive={false}>
+          <Button onClick={onRestoreClick} type="submit">
+            Restore
+          </Button>
+          <Button onClick={onDiscardClick}>Discard</Button>
+        </ButtonGroup>
+        <div className="explanatory-text">{restoreMessage}</div>
       </div>
     </div>
   )
--- a/app/src/ui/stashing/stash-diff-viewer.tsx
+++ b/app/src/ui/stashing/stash-diff-viewer.tsx
@@ -140,12 +140,14 @@ const Header: React.SFC<{
     <div className="header">
       <h3>Stashed changes</h3>
       <ButtonGroup destructive={false}>
+      <div className="row">
         <Button onClick={onRestoreClick} type="submit">
           Restore
         </Button>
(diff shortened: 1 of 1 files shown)


app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -94,20 +94,26 @@ const Header: React.SFC<{
...
       <h3>Stashed changes</h3>
-      <ButtonGroup destructive={true}>
-        <Button onClick={onSubmitClick} type="submit">
+      {/*
+      this is false because ButtonGroup assumes the submit button
+      performs the destructive action. In this case the destructive
+      action is performed by the non-submit button so we lie to the
+      props to get the correct button ordering
+       */}
+      <ButtonGroup destructive={false}>
```
billygriffin: I think this may work just fine without the destructive false, unless I'm missing something?

outofambit: it would! i wanted to pass it explicitly (and with the comment above it) so the intent was super clear 😄

iAmWillShepherd: Moving this comment outside of the return statement will help make the code a bit more readable.

app/src/ui/stashing/stash-diff-viewer.tsx
```diff
@@ -109,14 +109,11 @@ const Header: React.SFC<{
   dispatcher: Dispatcher
   isWorkingTreeClean: boolean
 }> = props => {
-  const { dispatcher, repository, stashEntry, isWorkingTreeClean } = props
-
-  const onClearClick = () => {
-    dispatcher.dropStash(repository, stashEntry)
+  const onDiscardClick = () => {
+    props.dispatcher.dropStash(props.repository, props.stashEntry)
```
iAmWillShepherd: `dispatcher`, `repository`, and `stashEntry` can be destructured to reduce the amount of `props.<field>` expressions.

iAmWillShepherd: I addressed this in c8956f0 since I broke the build with my previous merge.

app/src/ui/stashing/stash-diff-viewer.tsx
outofambit: done in https://github.com/desktop/desktop/pull/7392/commits/bb022a458aac3d3a75188306c4ec12fc574b0cbc

PR #7393, opened 2019-04-24
Update verbiage for leaving changes on current branch vs bringing them to new branch

billygriffin:
## Overview

Based on some user feedback and discussion in Slack, proposing a few changes to the dialog that pops up when switching branches that allows you to either stash your changes or bring them with you to the new branch.

The goals:

1. Remove explicit "Would you like to stash?" and "Yes/No" because that can convey different things to different people depending on their workflow.
2. Use similar language pattern for explanatory text in each box to help with consistency. 
3. Add a bit of separation between the question of what you'd like to do with the changes and your options. Note: This bleeds through to other similar dialogs (like creating a new branch), but I tested them and they all actually benefit from the small additional vertical spacing I think.

Before:

Screen Shot 2019-04-24 at 12 07 07 PM

After:

Screen Shot 2019-04-24 at 2 08 20 PM

## Release notes

No notes

### DIFF
--- a/app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
+++ b/app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
@@ -97,20 +97,20 @@ export class StashAndSwitchBranch extends React.Component<
     const { branchToCheckout } = this.props
     const items = [
       {
-        title: `Yes, stash my changes on ${this.props.currentBranch.name}`,
-        description: 'Stash your in-progress work and return to it later',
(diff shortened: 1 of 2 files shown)


tierninho:
Overall an improvement. Possible to add the `ref-component` UI treatment on the branches like so?

image

billygriffin:
@tierninho Thanks! I like that idea but given that each of them can be highlighted, I think I'd prefer to look at that separately. Also the structure of this is a bit different than where that ref component is used elsewhere, so it's not super straightforward.

PR #7406, opened 2019-04-25
Fix: Cannot use `esc` key to close Repository list and Branch list

### LINKED ISSUE #7177: Cannot use `esc` key to close Branch list
Using the `esc` key to close the `branches-container` is still possible, however I found an edge case where if you click into the white space in the Pull Request list, then using `esc` won't close the list.

Screen Shot 2019-03-28 at 3 03 08 PM

v1.7.0-beta1, mac

roottool:
## Overview

**Closes #7177**

## Description

This PR fixes to close Repository list and Branch list using Escape key.
The reason of this bug is that `onKeyDown` in `<div>` on React is not working.
We can focus to this `<div>` by `tabIndex={0}`.
I referred to [this answer of stackoverflow](https://stackoverflow.com/a/44434971).

### Currently

BeforeFixToCloseDropdown

### After fixing

FixToCloseDropdown

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/toolbar/dropdown.tsx
+++ b/app/src/ui/toolbar/dropdown.tsx
@@ -137,7 +137,7 @@ interface IToolbarDropdownState {
 export class ToolbarDropdown extends React.Component<
   IToolbarDropdownProps,
   IToolbarDropdownState
-  > {
+> {
   private innerButton: ToolbarButton | null = null
 
   public constructor(props: IToolbarDropdownProps) {
--- a/app/src/ui/toolbar/dropdown.tsx
+++ b/app/src/ui/toolbar/dropdown.tsx
@@ -279,7 +279,7 @@ export class ToolbarDropdown extends React.Component<
         <div
           className="foldout"
           style={this.getFoldoutStyle()}
-          tabIndex={0}
+          tabIndex={-1}
           onKeyDown={this.onFoldoutKeyDown}
         >
           {this.props.dropdownContentRenderer()}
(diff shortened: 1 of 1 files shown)


billygriffin:
Thanks @roottool, great to see you back! I'm marking this as `ready-for-review` as it looks like you've got all the elements in the PR body that we're looking for to be able to do a thorough review. As I think you know, we're working toward a big release, so I can't promise you'll get a review quickly since this isn't a time-sensitive fix, but we'll try to get to it as soon as we can. Thanks again!

tierninho:
@desktop/core Anyone else getting build errors when running this on dev? 

Ran into `[WEBPACK] Build failed after 21.243 seconds` etc..

outofambit:
> Anyone else getting build errors when running this on dev?
> 
> Ran into `[WEBPACK] Build failed after 21.243 seconds` etc..

@tierninho i was able to build this without errors locally. happy to help troubleshoot your build offline.

roottool:
@outofambit I noticed that I was making a mistake.
It is possible to focus on this div tag with the `Tab` key.
I think that it is best not to be able to focus this div tag with the `Tab` key.

> A negative value (usually tabindex="-1") means that the element should be focusable, but should not be reachable via sequential keyboard navigation. It's mostly useful to create accessible widgets with JavaScript.
[tabindex - MDN - Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

According to the reference, I will change the value from `tabIndex={0}` to `tabIndex={-1}`,

May I fix that?

## Screenshot

Mistake

outofambit:
@roottool great catch! i totally missed that. yes please, make that change!

PR #7410, opened 2019-04-25
Prevent re-render while in the middle of stashing and checking out

### LINKED ISSUE #7402: checkout branch with stash warning changes branch name while "loading"
## Description

I noticed when switching branches on a test VM that the operation would take a while but the branch name would change in the dialog.

Here's a GIF to try and illustrate this:

## Version

* GitHub Desktop: `1.7.0-beta5`

## Steps to Reproduce

1. Have some local changes that you wish to bring with you
2. Choose a different branch from the branch switcher
3. Choose "Bring my changes along..." (not sure if this is important)
4. Click "Switch branches" and wait

### Expected Behavior

Loading icon appears, buttons are disabled while action happens, then dialog disappears

### Actual Behavior

Midway through the operation occurring the branch updates to the new branch, so in the list you see both branches are the "destination" branch. This suggests some app state update is updating the UI post-checkout, but pre-dismissing the dialog.

iAmWillShepherd:
Fixes #7402 

**Before**

**After**

### DIFF
--- a/app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
+++ b/app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx
@@ -50,6 +50,19 @@ export class StashAndSwitchBranch extends React.Component<
     }
   }
 
+  public shouldComponentUpdate(
+    nextProps: ISwitchBranchProps,
+    nextState: ISwitchBranchState
+  ) {
+    // If we are in the middle of checking out a branch, prevent the component from rendering to avoid
+    // the button text that represents `StashAction.StashOnCurrentBranch` from updating before the component
+    // is dismissed. (see https://github.com/desktop/desktop/issues/7402)
+    return !(
+      nextState.selectedStashAction === StashAction.MoveToNewBranch &&
+      nextState.isStashingChanges
+    )
+  }
+
   public render() {
     const { isStashingChanges } = this.state
     return (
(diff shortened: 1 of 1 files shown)


tierninho:
I am still seeing the branch switch occur. In your "after" gif, try selecting the first option instead and it should appear. When you select the second option, the issue goes away.

shiftkey:
I also wanted to mention that running this up in development mode displays a `Warning: Can't call setState (or forceUpdate) on an unmounted component.` error:



It's not directly related to the change introduced by this PR, but looks to be related to the second `setState` in here:

https://github.com/desktop/desktop/blob/1cad7c13313e2dc368dd8d1c1dbc1349e131b56a/app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx#L146-L152

shiftkey:
Because `setState` doesn't immediately trigger a re-render, I think what you're looking for to address the devtools warning is to use the second parameter to `setState`, which is a callback that fires after the state is actually updated:

```ts
this.setState({ isStashingChanges: false }, () => {
  this.props.onDismissed()
})
```

This ensures you dismiss the dialog after the state update has been applied to the component.

iAmWillShepherd:
> I am still seeing the branch switch occur. In your "after" gif, try selecting the first option instead and it should appear. When you select the second option, the issue goes away.

Could you send a gif of that behavior? I'm unable to reproduce and want to make sure I'm not missing something.

Issue #5342, opened 2018-08-04
UX around Discard Changes is unclear


5acres:
It is possible in 1.3.2 to Discard a single change (right-clicking on it), or "Discard All Changes...", but there is no way to discard only the changes that are selected (checkboxes).

The old version had this.  It is almost a necessary feature.  If you want to discard "all but one" from a long list, for example, you have to discard them one by one.

![screen shot 2018-08-04 at 12 06 50 pm](https://user-images.githubusercontent.com/40286055/43679699-f0b958ce-97de-11e8-9b1c-a64e6b08f568.png)

MichaelHillcox:
Hi @5acres I'd say this is more of a usability issue if you want to call it an issue but what you've described is more than possible but you have to do it slighly differently.

Instead of using the checkboxes use ctr / command ( mac ) and click on the item's you'd like to discard. Once you have selected the ones you no longer want you can right click on any selected item and discard the changes. This will affect all highlighed items.

Hopefully it should look like this 
![](https://i.imgur.com/BSzGFok.png)

5acres:
Interesting, so there are two competing selection models.  How would one know about the Cmd-click selection model, and why not just use the check boxes, which are already a form of selection?  That's what the previous Mac desktop client did.

It's worth noting that all of these are hidden under right-click and command key modifiers, so they are not "discoverable" in the general sense of the word.

But ... thanks for the workaround!

MichaelHillcox:
So, I understand your confusion but I'd also have to argue that Cmd+click is pretty native when being given a list of items like you are presented with within the client. 

The confusing part is when you don't know that the checkboxes are for staging and unstanging changes and nothing to do with the actual act of selecting an item to perform an action.

5acres:
It's bad design.  Conflicting selection paradigms.

You say the checkboxes are "for staging and upstaging changes" and "nothing to do with selecting an item to perform an action..." and yet they select items to perform the Commit action.

Obviously the checkboxes are selection.  Just like Gmail, as an example.  That you decided you needed a second, different selection paradigm for a different set of operations makes no sense.

I've been building software for 30 years and I can't think of another example of two orthogonal selection paradigms for different sets of operations, with the possible exception of the marquee selection vs. masks in Photoshop, but that's pretty esoteric.  Can you think of another example where that's been done, or is actually necessary?

Anyway, thanks for responding quickly.

MichaelHillcox:
[...]

PR #7415, opened 2019-04-26
Improve "Discard all changes" UX

### LINKED ISSUE #7394: improve "discard all changes" UX

For the longest time we've had the ability to discard all changes by right-clicking the "Changes" header:

This was available on at least the classic Windows app, and it came over in #1647 because we needed _somewhere_ to keep this functionality. Context menus for non-interactive elements like headers aren't very intuitive, but we marched on.

Then we added it to the context menu for a changed file in #4188 (requested in #4314):

This has lead to some confusion in #5342 because it is right next to the "Discard Change" menu item, and if a user doesn't have the "Confirm Discard Changes..."  setting enabled they could accidentally move all the files to the trash.

This combination of "header context menu" and "item context menu" makes me think that neither place is the right place for this functionality.

in https://github.com/desktop/desktop/issues/5342#issuecomment-457417470 we identified a list of changes that will help us to improve this area: 

> I think we can solve this through implementing these four things, I believe in this order:
> 
> 1. Remove `Discard All Changes` from file context menu
> 2. Add `Discard All Changes` as a top menu item
> 3. Remove `Discard All Changes` from the header context menu
> 4. Add a UI menu to the changes top bar (containing `Discard All Changes` and potential future actions)
> 
> For anyone who picks this up:
> 
> For 2, I think `Discard All Changes` should live in `Branch` > `Discard All Changes` in a new section between the first and the second section.
> 
> For 4, I think we can wait and see how 1-3 feels first before implementing.

ahuth:
## Overview

**Closes #7394**

This is my first contribution to this project, and it seems like reasonable place to start.

## Description

Implements 3 of the 4 steps outlined in #7394 
- [x] Remove "Discard all changes" from the changes list context menu - 5bcc5e0b2423b42ba34bbeac8c9073cd2061ded3
- [x] Remove "Discard all changes" from the file context menu - 8b0aa3124d40d4b59de1e58661a8b32811a0c955
- [x] Add "Discard all changes" to as a top-level menu item under Branch - ff136f200867e829646e5a7e496d888d7dd175d1

Here's a screenshot of the top-level menu item:

discard all

## Release notes

Notes:
- Update the "Discard all changes" menu items to make it hard to inadvertently activate


app/src/ui/changes/changes-list.tsx
```diff
@@ -239,8 +239,8 @@ export class ChangesList extends React.Component<
       selection === DiffSelectionType.All
         ? true
         : selection === DiffSelectionType.None
-        ? false
-        : null
+          ? false
+          : null
```
ahuth: Prettier changes

app/src/ui/changes/changes-list.tsx
```diff
@@ -304,31 +300,12 @@ export class ChangesList extends React.Component<
           ? `Discard Changes`
           : `Discard changes`
         : __DARWIN__
-        ? `Discard ${files.length} Selected Changes`
-        : `Discard ${files.length} selected changes`
+          ? `Discard ${files.length} Selected Changes`
+          : `Discard ${files.length} selected changes`
```
ahuth: Prettier changes

app/src/ui/changes/changes-list.tsx
```diff
@@ -107,7 +107,7 @@ interface IChangesListProps {
   readonly onDiscardChanges: (file: WorkingDirectoryFileChange) => void
   readonly askForConfirmationOnDiscardChanges: boolean
   readonly focusCommitMessage: boolean
-  readonly onDiscardAllChanges: (
+  readonly onDiscardChangesFromFiles: (
```
ahuth: Renamed the method to more accurately capture its intent after these changes.

ahuth:
This is ready for review

tierninho:
@ahuth Many thanks for the contribution, much appreciated and good work!

> Add "Discard all changes" to as a top-level menu item under Branch

This menu item should be disabled if there are no files in the Changes list. 

image

@shiftkey Are we always going to show a "are you sure?" dialog for discarding **all** files? The discard file option in Preferences is only applicable for single files.

d564b9775e, mac

app/src/lib/menu-update.ts
```diff
@@ -260,6 +263,11 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
       !tipStateIsUnknown && !branchIsUnborn && !rebaseInProgress
     )
 
+    menuStateBuilder.setEnabled(
+      'discard-all-changes',
+      repositoryActive && hasChangedFiles && !rebaseInProgress
```
ahuth: Any other conditions we need to take into account? The previous logic (for the change list context menu) had:

```ts
// need to preserve the working directory state while dealing with conflicts
if (this.props.rebaseConflictState !== null) {
  return
}
```

ahuth:
Good call! The menu item is now disabled when:

- There are no files with changes
- It's in a rebase conflict state

shiftkey:
> @shiftkey Are we always going to show a "are you sure?" dialog for discarding **all** files? The discard file option in Preferences is only applicable for single files.

@tierninho @ahuth I interpret the `Show confirmation dialog before discarding changes` setting as being applicable to any sort of discard operation. So 👍 to showing it when all files are discarded.

shiftkey:
> I interpret the Show confirmation dialog before discarding changes setting as being applicable to any sort of discard operation. So 👍 to showing it when all files are discarded.

Actually, after reviewing the changes and digging into the history of this area it looks like we made a conscious decision to always show the confirmation dialog and always used `Discard All Changes...` for the menu label, so I'm 👍 to preserving the current behaviour (always showing the dialog) rather than changing it to be in sync with the preference as part of this PR.

app/src/lib/menu-update.ts
shiftkey: I think this is fine

shiftkey:
@ahuth one little bit of cleanup we could do while we're in here is to make `isDiscardingAllChanges` a required parameter:

```diff
diff --git a/app/src/ui/changes/changes-list.tsx b/app/src/ui/changes/changes-list.tsx
index d34a44faa..ba5ebcf2f 100644
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -109,7 +109,7 @@ interface IChangesListProps {
   readonly focusCommitMessage: boolean
   readonly onDiscardChangesFromFiles: (
     files: ReadonlyArray<WorkingDirectoryFileChange>,
-    isDiscardingAllChanges?: boolean
+    isDiscardingAllChanges: boolean
   ) => void
 
   /** Callback that fires on page scroll to pass the new scrollTop location */
diff --git a/app/src/ui/changes/sidebar.tsx b/app/src/ui/changes/sidebar.tsx
index 64f734e5a..2d0a4c3c6 100644
--- a/app/src/ui/changes/sidebar.tsx
+++ b/app/src/ui/changes/sidebar.tsx
@@ -234,7 +234,7 @@ export class ChangesSidebar extends React.Component<IChangesSidebarProps, {}> {
 
   private onDiscardChangesFromFiles = (
     files: ReadonlyArray<WorkingDirectoryFileChange>,
-    isDiscardingAllChanges: boolean = true
+    isDiscardingAllChanges: boolean
   ) => {
     this.props.dispatcher.showPopup({
       type: PopupType.ConfirmDiscardChanges,
```

tierninho:
@ahuth Thank you for the updated logic and patience when making the latest changes. 

All looks good now ✨🚢

ahuth:
Thanks for the review/thoughts! I've made the `isDiscardingAllChanges` param cleanup.

ahuth:
Hmm, not sure what's up with the failing tests.

Locally I get `yarn test:integration` failing because a timeout.

On CI, however, I think the build logs show it failed to install both PhantomJS and Electron. Weird.

shiftkey:
@ahuth I believe that's an intermittent network issue - I've re-run the build to see if that addresses it

ahuth:
Thanks!

billygriffin:
@ahuth Thanks so much for the contribution - this is great and I know you'll make a lot of people happy who have inadvertently clicked "Discard all changes." This should get merged in the next couple days, and it'll likely ship to beta later this week and to production in our next large release (probably the one after next so it has a bit of time on beta to make sure it's working correctly). Congrats on your first PR! ❤️

nycdotnet:
Gosh I miss this on the right-click menu on the top.  While I agree this is more safe, this change feels to me like a real regression in UX, and I thought it was a bug.  Perhaps I'm just an old-timer with this app.  :-(

Regardless, thanks for the contribution @ahuth !

billygriffin:
Thanks for the feedback @nycdotnet! If we see quite a bit of evidence from people similar to you who are missing that functionality, I think we have ideas for a reasonable solution to help improve things. So it's really helpful to hear how you're missing it, thanks again.

shiftkey:
> While I agree this is more safe, this change feels to me like a real regression in UX, and I thought it was a bug. Perhaps I'm just an old-timer with this app. :-(

@nycdotnet I think a decent first step to iterate on this would be to give it a keyboard shortcut for easier access. Please bump #7588 if you agree.

fatcerberus:
I don't even know how I first discovered this option - I think I just subconsciously knew it would be there, which doesn't exactly say "bad UX" to me.

I'm sad this command is now gone and moved to the menu bar which isn't anywhere near as ergonomic - this was a big part of my workflow (do some experimental change -> experiment fails -> discard-all and try again).  I didn't find the command particularly unsafe as there's a huge "Are you really sure?" warning that pops up which AFAICT can't be disabled, and right-clicking "X changed files" to discard everything is a natural extension of right-clicking individual files to discard them.

A keyboard shortcut would help, but personally I'm kind of mouse-bound (which is why I want to use a GUI client in the first place); right-clicking things to discover new stuff comes more naturally than hunting through menus, so I tend to miss features if that's the only way to discover them.

Just as I had tried switching to the new client again too... oh well, back to the ancient Classic Desktop again.

shiftkey:
@fatcerberus does the keyboard shortcut proposed in #7650 make it more accessible to you?

fatcerberus:
I’m thinking no, as it’s a two-handed command (or a very gymnastic one-hander :smile:) so I would have to take my hand off the mouse; but my point above was that I don’t find keyboard shortcuts particularly discoverable *in general* compared to just poking around with the right mouse button.

shiftkey:
@fatcerberus thanks for the extra context!

It's challenging to reconcile your scenario with the initial feedback about the context menu being confusing in https://github.com/desktop/desktop/issues/5342, but we'll keep this in mind.

fatcerberus:
I never found it particularly confusing - it just seemed natural that if I can right-click individual files to discard changes in them, I would be able to do the same to the “summary” above the list to discard everything.

The confusion in that issue from my quick skim looks like it’s more about the function of the checkboxes than the discard feature itself (though maybe I didn’t read closely enough); since selecting multiple files didn’t affect the diff view, I figured out pretty quickly that they were only there to determine what gets committed and had nothing to do with anything else going on in the UI.

nycdotnet:
> I never found it particularly confusing - it just seemed natural that if I can right-click individual files to discard changes in them, I would be able to do the same to the “summary” above the list to discard everything.

Exactly this.  The old UX felt very tactile/interactive.

NickCraver:
Can we please bring back the header right click menu?

I used this many times a day. When right clicking on an item like a file, I expect to have options for that thing. When I see a count up top, I naturally right clicked thinking I'd have options for `n` thing (and I did!). I was pretty surprised this was removed. Does *also* having the right click menu there cause some issue? Selecting all the files to do an action or hunting in another place is *not* intuitive to me. It also puts a lot of distance between a global work-loss action and the count of things I'm about to execute it on. That context should be readily apparent. While it did say "Discard all changes" before, it's far away from context now, and I have to go mentally check if I want to really execute that dangerous action. When it was immediately beside the things it affects? Yeah sure, I know where I'm at and what's about to happen. I know that's minor, but I wanted to convey why that feels like a regression in and of itself.

shiftkey:
For anyone else stumbling into this issue, please add your feedback to https://github.com/desktop/desktop/issues/7696

PR #7420, opened 2019-04-28
tidy up tests to avoid needing to exclaim ourselves

shiftkey:
## Overview

I can't recall the PR where this was first discussed, but on Friday I was looking over some Git tests thinking about some optimizations we could make and realized there was a chance to cleanup some of our test setup.

## Description

A while ago we were talking about how our tests are sometimes fighting with the typechecker because of how we setup the objects used in the test:

```
let repo: Repository | null = null
...
repo = await setupEmptyRepository()
...
const after = await getStatusOrThrow(repo!)
```

We don't need to explicitly assign `null` when initializing these variables (we would if this was inside a class because of the `--strictPropertyInitialization` compiler flag that landed in [TypeScript 2.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html)), and if we omit the `| null = null` part here we then avoid the need to tell the typechecker that we expect a value to exist here (using `!`).

I feel this is a safe change because the use of `!` is a compile-time check, and running the tests would verify they are correct no matter which approach we use here - so why not keep things simple?

I also tidied up some places where we'd use `repo` or `r` as an alias and then `const repository = repo!` to workaround this issue.

## Release notes

Notes: no-notes

### DIFF
--- a/app/test/unit/accounts-store-test.ts
+++ b/app/test/unit/accounts-store-test.ts
@@ -3,7 +3,8 @@ import { AccountsStore } from '../../src/lib/stores'
 import { InMemoryStore, AsyncInMemoryStore } from '../helpers/stores'
 
 describe('AccountsStore', () => {
-  let accountsStore: AccountsStore | null = null
+  let accountsStore: AccountsStore
+
   beforeEach(() => {
     accountsStore = new AccountsStore(
       new InMemoryStore(),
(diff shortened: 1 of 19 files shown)


app/test/helpers/random-data.ts
```diff
@@ -0,0 +1,12 @@
+import { randomBytes } from 'crypto'
+
+/**
+ * Generate a pseudo-random sequence of hexadecimal values for use in testing.
+ *
+ * Source: https://blog.abelotech.com/posts/generate-random-values-nodejs-javascript/
+ */
+export function generateString(length: number = 32) {
```
outofambit: does this overlap with #7499? it looks like b19af067816ce760bf643f8ccf8bba1d217c8215 is in both PRs

shiftkey: Yep, and not needed here. Let me update the PR to drop this work.

Issue #3664, opened 2017-12-22
Summary and description lost when closing application


howprice:
### Description

The contents of the Summary and Description fields are lost without warning when closing the application.

### Version

**GitHub Desktop version:** 1.0.11

**OS version:** Microsoft Windows [Version 10.0.15063]

### Steps to Reproduce

1. Make some changes to files
2. Write a Summary and (lengthy, detailed) Description, but _do not Commit_
3. Close the GitHub Desktop application
4. Re-launch GitHub Desktop
5. The Summary and Description fields are now empty. Nooooooo!

**Expected behavior:** Either (a) The application warned on shutdown that the contents of the Summary/Description fields will be lost, if not empty, or (b) The application saved the contents of these two fields to local OS user storage and restored on re-launch

**Actual behavior:** On closing the Github Desktop application, any text entered in Summary and Description is irretrievably lost without warning

**Reproduces how often:** 100%

#### Additional Information

None

iAmWillShepherd:
I'm not sure that we will save state between sessions, but definitely something we can discuss. I think warning people that they have a _commit in progress_ is fair.

seadoggie01:
💔 This is my biggest sticking point with using GitHub desktop. I can't start writing commit messages while still editing my code. I end up saving my commit messages to a sticky note so they don't die, which is just ridiculous! This is 2019, I shouldn't need to supplement an app with MS's Stickies D:

Can we please get some love on this 1.5 year old issue?

Trainax:
Hello. Are there any updates on this issue?

Other than commit summary and description I would also like for selected changed files to be saved and restored on re-launch if possible. I sometimes need to commit only a part of the files I have modified and it would be nice for the file selection to be kept between app launches until the commit is done.

Thank you

cooljeanius:
This applies on Mac across reboots, too; on Mac there is a "Reopen windows when logging back in" API that could be leveraged here; check in `~/Library/Saved Application State/` for examples

steveward:
From #14382:

> ### The feature request
> 
> Currently, the commit message title, summary, co-contributors, and selected files are lost when GitHub Desktop updates, the system restarts, or GitHub Desktop is restarted. This can be especially annoying when composing a long description for a commit that is not completely finalized.
> 
> #### Workarounds
> - Commit with the message and description and then amend the last commit when more changes are made
> - Save the commit message and description to a text document
> - Type them out again
> 
[...]

PR #7427, opened 2019-04-29
get last stored value for stash pane width at launch

### LINKED ISSUE #7416: Stash view state is not remembered after restart
As we remember other panes width in Desktop, we should include the stash pane. 

Not sure if the temporary state of this feature will affect this request.

Screen Shot 2019-04-26 at 3 36 56 PM

Repro:
- adjust width
- restart Desktop

Note: we also do not remember that Stash page was last shown. After restarting, the next steps is shown instead of the stash page

1.7.0beta5, mac

outofambit:
## Overview

closes #7416 

## Release notes

Notes: [Fixed] Stash files list retains width between app restarts

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1572,6 +1572,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
       commitSummaryWidthConfigKey,
       defaultCommitSummaryWidth
     )
+    this.stashedFilesWidth = getNumber(
+      stashedFilesWidthConfigKey,
+      defaultStashedFilesWidth
+    )
 
     this.confirmRepoRemoval = getBoolean(
       confirmRepoRemovalKey,
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -327,9 +327,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
    */
   private appIsFocused: boolean = false
 
-  private sidebarWidth: number = defaultSidebarWidth
-  private commitSummaryWidth: number = defaultCommitSummaryWidth
-  private stashedFilesWidth: number = defaultStashedFilesWidth
+  private sidebarWidth: number = getNumber(
+    sidebarWidthConfigKey,
+    defaultSidebarWidth
+  )
+  private commitSummaryWidth: number = getNumber(
+    commitSummaryWidthConfigKey,
+    defaultCommitSummaryWidth
+  )
+  private stashedFilesWidth: number = getNumber(
+    stashedFilesWidthConfigKey,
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -327,9 +327,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
    */
   private appIsFocused: boolean = false
 
-  private sidebarWidth: number = defaultSidebarWidth
-  private commitSummaryWidth: number = defaultCommitSummaryWidth
-  private stashedFilesWidth: number = defaultStashedFilesWidth
+  private sidebarWidth: number = getNumber(
```
shiftkey: We already update these values in `loadInitialState()`, and I'm curious whether that isn't necessary any more (tradeoffs aside):

https://github.com/desktop/desktop/blob/be399eb44f3a03ed67fc9ae8b52affdaa4c82f7a/app/src/lib/stores/app-store.ts#L1570-L1574

Alternatively, could we just update `stashedFileChanges` there?

outofambit: yup, just saw that. totally my mistake. updating now

outofambit: 🤦‍♀ f5afa9d

billygriffin:
Thanks @outofambit! Just checking, does this also retain knowledge that you were viewing a stash when restarting? That was the other thing @tierninho brought up in the issue.

outofambit:
> does this also retain knowledge that you were viewing a stash when restarting?

@billygriffin no, it does not. i can file a separate issue if we want to discuss/fix that

shiftkey:
> i can file a separate issue if we want to discuss/fix that

👍 we don't remember the tab you were on when you close and restart Desktop, so this is a bit more involved than what's involved with these stashing changes...

billygriffin:
Sounds good to me - I don't have much heartburn about remembering that you were viewing the stash, as it's readily available again for you when you start back up. And it being consistent with other behavior is 👍.

tierninho:
Tested the fix and looks good. 

> Note: we also do not remember that Stash page was last shown. After restarting, the next steps is shown instead of the stash page

We do remember the last repo and last branch when restarting. I thought we got more granular, but no big deal as that is not the case.

shiftkey:
I plan to merge this in later today because the fix has been verified and the change is minor

shiftkey:
> We do remember the last repo and last branch when restarting. 

The last repository is backed by `last-selected-repository-id` in local storage, and the branch is whatever `HEAD` on Git currently points to (so we get that for free).

> I thought we got more granular, but no big deal as that is not the case.

I've not heard many requests for "restore XYZ after closing Desktop" aside from #3664 but happy to discuss options.

PR #7428, opened 2019-04-29
actually disable restore stash button when changes present

### LINKED ISSUE #7409: Block restoring a stash if files in Changes list present
The app is still allowing stashed files to be restored if files are present in the Changes list. Seems like a regression.

`Restore` button is enabled and working:
image

1.7.0beta5, mac

outofambit:
## Overview

closes #7409


tierninho:
Verified the fix 👍

shiftkey:
I plan to merge this in later today because the fix has been verified and the change is minor

PR #7429, opened 2019-04-29
unified and simplified API for updating menu labels based on app state

shiftkey:
## Overview

**Refactoring related to https://github.com/desktop/desktop/pull/7243#discussion_r276333030**

## Description

This PR explores cleaning up some parts of `updatePreferredAppMenuItemLabels`:

 - strings are managed by both the renderer and main process, meaning changes to some menu labels need to be kept in sync manually
 - this event payload is pretty flexible - accepting `undefined` as a fallback value which is mapped to some sort of default in the `buildDefaultMenu` function
 - lack of documentation on the fields used
 - inconsistent naming of things

Changes implemented in this PR:

 - implement two new `models` modules - `menu-labels` and `menu-ids` - that replace things which previously lived in `main-process` but were linked to in `lib` and `ui` (this felt like the wrong direction for importing)
 - a new `MenuLabelsEvent` that replaces `MenuLabels` and only provides the data necessary for the menu changes, not the labels themselves
 - `buildDefaultMenu` is now the source of truth for the labels shown

TODO:

 - [x] more testing

## Release notes

Notes: no-notes


shiftkey:
After reviewing #7415 I decided that there weren't any overlapping changes with this PR (we always show the `...` on `Discard All Changes...`) so these can be landed independent of eachother.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -8,50 +8,58 @@ import { ensureDir } from 'fs-extra'
...
+  hasCurrentPullRequest = false,
+  defaultBranchName = defaultBranchNameValue,
   isForcePushForCurrentRepository = false,
-  askForConfirmationOnForcePush = false,
   isStashedChangesVisible = false,
-}: MenuLabels): Electron.Menu {
+}: MenuLabelsEvent): Electron.Menu {
   defaultBranchName = truncateWithEllipsis(defaultBranchName, 25)
 
+  const removeRepoLabel = askForConfirmationOnRepositoryRemoval
```
iAmWillShepherd: Nice change!

PR #7430, opened 2019-04-29
Use workspace file when opening repository in VSCode

### LINKED ISSUE #7007: Use workspace file when opening repository in VS Code
**Problem**
When I open a repo in Github Desktop the app suggests to open the repo in Visual Studio Code. That's great, but instead of opening the root folder I would like it to open .code (workspace) file when it is available in the root. Either per default or via a setting.

Thanks for a great product!

roottool:
## Overview

**Closes #7007**

## Description

This PR adds a feature to open workspace file (`.code-workspace`) when the repository is opened in Visual Studio Code.
We can turn on/off this feature at `Options -> Advanced` when we choose Visual Studio Code in the external editor.
AddNewOption

This feature detail is the below.

- The repository root folder is opened when we don't use this feature or `.code-workspace` file doesn't exist in the repository root folder.
- When only the one `.code-workspace` file exists, the `.code-workspace` file is opened.
- When the `.code-workspace` file exists more than one, the app pop up a file dialog.
- We can choose an one `.code-workspace` file by a file dialog.
- When the cancel button in a file dialog is clicked, the repository root folder is opened.

## Screenshot

ChooseWorkspaceFile

## Release notes

Notes: Adds a feature to open workspace file when the repository is opened in Visual Studio Code.


shiftkey:
@roottool thanks for opening this up. I haven't had a chance to look over the changes, but wanted to ask more about this item:

> We can turn on/off this feature at `Options -> Advanced` when we choose Visual Studio Code in the external editor.

Are there cases you can think of where you'd want to not open a workspace file? Having "just another checkbox" in here makes me wonder if we can avoid this and have something "just work" without needing users to configure their environment beyond just choosing their editor and have this enabled by default...

roottool:
> Are there cases you can think of where you'd want to not open a workspace file?

I think the other way is that the app pops up [dialog of Electorn](https://electronjs.org/docs/api/dialog) every time if workspace file exists in repository root folder before the app opens the repository or workspace file.
The checkbox will be unnecessary if we are checked by dialog.

I feel troublesome that we checks to open or not workspace file every time by dialog.
Therefore, I implement the checkbox.

app/src/lib/editors/launch.ts
```diff
@@ -23,9 +24,102 @@ export async function launchExternalEditor(
...
+  repoRootFolderPath: string
+) {
+  const usesShell = editor.usesShell ? editor.usesShell : false
+  const useWorkspace = editor.useWorkspace ? editor.useWorkspace : false
+
+  if (!useWorkspace) {
+    spwanExternalEditor(editor.path, repoRootFolderPath, usesShell)
+  } else {
+    const workspacePattern = join(repoRootFolderPath, '*.code-workspace')
+    const glob = require('glob')
```
shiftkey: Please lift this up to the start of the file and use the relevant `import` syntax:

```ts
import * as glob from 'glob'
```

This will allow you to simplify the signature for `glob` usage:

```ts
glob(workspacePattern, (error, files) => {
```

app/src/lib/editors/found-editor.ts
```diff
@@ -2,4 +2,5 @@ export interface IFoundEditor<T> {
   readonly editor: T
   readonly path: string
   readonly usesShell?: boolean
+  useWorkspace?: boolean
```
shiftkey: Unless this needs to be mutable, we prefer to make things `readonly` by default

```suggestion
  readonly useWorkspace?: boolean
```

shiftkey (commented):
> I think the other way is that the app pops up dialog of Electorn every time if workspace file exists in repository root folder before the app opens the repository or workspace file.

As part of this first iteration, I'd like to exclude any "preferences" tracking or work required to remember choices, as I think we should get feedback by enabling this by default when users have already chosen VSCode (we can leverage feature-flags and keep this work on our beta channel for the moment).

> I feel troublesome that we checks to open or not workspace file every time by dialog.

Thinking about this some more, I think there's some interesting possibilities to explore around this area:

 - the first time a user opens a repository in VSCode with a workspace file, we could prompt them (using our custom dialogs) and ask whether they would like to open the default workspace or open the root directory

 - if there are multiple workspace files available, we could show them this as a list of options and ask the user which one they would like to open (or to use the repository root)

This feels more friendly than using the `dialog` API, and gives us a chance to provide the user with more context about what the app will do (and let the user take charge). If that interests you, I'd love to put some time aside later this week to help sketch out what these dialogs might look like and help wire them up with the work you have currently.

Let me know how you feel about these proposed changes.

roottool:
I agree your opinion. It is possible for users to reduce troublesome.

Currently, I have an one question.
Where should users change the location opened in VSCode if your opinion applies?
(e.g. Adds a new tab in repository setting.)
I think a pattern that users want to change from workspace file (or root directory) to root directory (or workspace file).

shiftkey:
> Where should users change the location opened in VSCode if your opinion applies?

Let's not worry about that for now. For the moment let's prompt on every launch, at least while we're developing the flow.

roottool:
I changed from file dialog to your custom dialog.
I noticed that the impact range of  the code in this app is bigger than my assumed if the selected location to open the repository is memorized in the app.
e.g.

- Creates new repository
- Selects an other repository

Currently, when the checkbox of using workspace file is checked and we select `Repository -> Open in Visual Studio Code`, the dialog always pops up.
The dialog doesn't pop up when we select `Context menu -> Open in Visual Studio Code`.

## Screenshot

ChooseWorkspaceFile

app/src/lib/stores/app-store.ts
```diff
@@ -4125,7 +4135,24 @@ export class AppStore extends TypedBaseStore<IAppState> {
         return
       }
 
-      await launchExternalEditor(fullPath, match)
+      const isVisualStudioCode =
+        match.editor === 'Visual Studio Code' ||
+        match.editor === 'Visual Studio Code (Insiders)'
+      const useWorkspaceFileInVSCode = this.getState().useWorkspaceFileInVSCode
```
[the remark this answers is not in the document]
roottool: I added a logic into `onSelectedEditorChanged()` in  `src/ui/preferences/advanced.tsx`.
https://github.com/desktop/desktop/blob/a00a6e9838c6c4a7e707528f519caad9e5f44ca2/app/src/ui/preferences/advanced.tsx#L145-L150
This allows the app to check whether the setting value is the other editor than VSCode or unchecked.

app/src/lib/stores/app-store.ts
```diff
@@ -1,5 +1,6 @@
 import { ipcRenderer, remote } from 'electron'
 import { pathExists } from 'fs-extra'
+import { lstatSync } from 'fs'
```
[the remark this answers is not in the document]
roottool: I fixed it.
Error handling is done by the existing `catch`.
https://github.com/desktop/desktop/blob/a00a6e9838c6c4a7e707528f519caad9e5f44ca2/app/src/lib/stores/app-store.ts#L4150-L4152
Should I add new `try-catch` to catch the `lstat` error?

app/src/ui/open-repository-in-vscode/open-repository-in-vscode.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  private getLocationList = () => {
+    const workspacePattern = join(this.props.repositoryPath, '*.code-workspace')
+    const locationList: ILocationList[] = [
+      {
+        locationName: __DARWIN__
+          ? 'The Repository Root Directory'
+          : 'The repository root directry',
```
Rexogamer: typo: “directry” should be “directory”

roottool: Thank you for the review! I fixed it.

niik:
Hi @roottool, first let me say thank you so much for working on this. I'm sorry that this has languished for so long. I'm afraid this got caught in between some staffing changes and didn't get picked up as it should have. We do really appreciate the work you've done here.

That said this isn't something we've heard a lot of feedback about and we don't have the resources available at the moment to review and help get this across the finish line.

I'm terribly sorry but I figured it'd be better to set expectations clearly here. If our priorities or our resources change in the future we can always reopen and revisit this PR.

4Ferdy:
T

MintThitiratChamnan:
@4Ferdy

Mina012015 (commented):
`app/src/lib/editors/launch.ts `

PR #7432, opened 2019-04-29
Manually drop successful stash pop with conflicts

### LINKED ISSUE #7383: [stashing] successfully popping a stash with conflicts doesn't drop the stash
**Repro steps**
1. Make a stash that touches a file
1. Update that in working directory and commit
1. Pop stash

Note that there will be conflicts and the stash content will be placed in the working directory, but Desktop will continue to show that there is a stash entry. _This is because of underlying git behavior aka desktop shows the stash because `pop` didn't actually remove the stash._

Screen Shot 2019-04-22 at 11 10 12 AM

You'll notice in this screenshot that while a stash has been popped (and resulted in conflicts) it still exists in the UI. We should ensure the stash is dropped as long as the pop is successful. (This doesn't include the case where [popping the stash fails due to conflicts](https://github.com/desktop/desktop/issues/7278).)

outofambit:
## Overview

Closes #7383 

## Description

- adds `--quiet` option to `stash pop`
- adds `1` as a success exit code for `stash pop` and manually inspects for output
- manually drops stash when its popped but not dropped by git
- updates and adds more test coverage for `stash pop`

### DIFF
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -172,8 +172,8 @@ export async function popStashEntry(
     // report an exit code of `1` and are not dropped after being applied.
     // so, we check for this case and drop them manually
     if (result.exitCode === 1) {
-      // because we use --quiet flag, we don't get output successful
-      // stash pops that have conflicts
+      // because we use --quiet flag, we don't get output for
+      // successful stash pops that have conflicts
       if (result.stderr.length > 0 || result.stdout.length > 0) {
         throw new GitError(result, args)
       }
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -172,9 +172,8 @@ export async function popStashEntry(
     // report an exit code of `1` and are not dropped after being applied.
     // so, we check for this case and drop them manually
     if (result.exitCode === 1) {
-      // because we use --quiet flag, we don't get output for
-      // successful stash pops that have conflicts
-      if (result.stderr.length > 0 || result.stdout.length > 0) {
+      if (result.stderr.length > 0) {
+        // rethrow, because anything in stderr should prevent the stash from being popped
         throw new GitError(result, args)
       }
 
--- a/app/src/lib/git/stash.ts
+++ b/app/src/lib/git/stash.ts
@@ -158,11 +158,34 @@ export async function popStashEntry(
   // ignoring these git errors for now, this will change when we start
   // implementing the stash conflict flow
   const expectedErrors = new Set<DugiteError>([DugiteError.MergeConflicts])
+  const successExitCodes = new Set<number>([0, 1])
   const stashToPop = await getStashEntryMatchingSha(repository, stashSha)
 
   if (stashToPop !== null) {
-    const args = ['stash', 'pop', `${stashToPop.name}`]
-    await git(args, repository.path, 'popStashEntry', { expectedErrors })
+    const args = ['stash', 'pop', '--quiet', `${stashToPop.name}`]
+    const result = await git(args, repository.path, 'popStashEntry', {
+      expectedErrors,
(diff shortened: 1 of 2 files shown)


app/src/lib/git/stash.ts
```diff
@@ -158,11 +158,34 @@ export async function popStashEntry(
...
+      successExitCodes,
+    })
+
+    // popping a stashes that create conflicts in the working directory
+    // report an exit code of `1` and are not dropped after being applied.
+    // so, we check for this case and drop them manually
+    if (result.exitCode === 1) {
+      // because we use --quiet flag, we don't get output successful
+      // stash pops that have conflicts
+      if (result.stderr.length > 0 || result.stdout.length > 0) {
```
outofambit: this might be too big of an assumption to make... i suppose we could inspect all lines for `CONFLICT`

```
CONFLICT (content): Merge conflict in README.md
```

🤔

tierninho:
For testing this, just want to check that this fix only pertains to 1) when there are conflicts in the Changes list, and 2) a stash is popped successfully. Therefore we should drop the stash as expected and the files will remain in the Changes list with the conflicted files? 

> `pop [--index] [-q|--quiet] [<stash>]` Remove a single stashed state from the stash list and apply it on top of the current working tree state.

Any changes to the stash list I should look out for and/or is it relevant for testing? Thanks!

outofambit:
>For testing this, just want to check that this fix only pertains to 1) when there are conflicts in the Changes list, and 2) a stash is popped successfully. Therefore we should drop the stash as expected and the files will remain in the Changes list with the conflicted files?

@tierninho this does not apply to situations where there are changes in the working directory. this is only for when you have made commits after making a stash. If those commits touch the same files/lines of any files/lines in the stash, then this comes into play. The expected result is to have the stash popped _and_ dropped.

app/test/unit/git/stash-test.ts
```diff
@@ -231,22 +231,67 @@ describe('git/stash', () => {
...
+        const entries = await getDesktopStashEntries(repository)
+        expect(entries.length).toBe(1)
+
+        const readme = path.join(repository.path, 'README.md')
+        await FSE.writeFile(readme, Math.random()) // eslint-disable-line insecure-random
+
+        const entryToApply = entries[0]
+        expect(
+          popStashEntry(repository, entryToApply.stashSha)
+        ).rejects.toThrowError()
```
shiftkey: 🤔 Didn't we have issues with using `.rejects` like this in the past? 

I pointed @iAmWillShepherd to this pattern in an earlier stashing PR which would be more reliable, despite me feeling sad about its verbosity...

```ts
      try {
        await dropDesktopStashEntry(repository, doesNotExist.stashSha)
      } catch {
        didFail = true
      }

      expect(didFail).toBe(false)
```

outofambit: ugh i totally forgot about that. looks like jest is still trying to figure this out (https://github.com/facebook/jest/issues/8140). 😞 i tried the pattern above and a couple things i found in the jest docs. it looks like the behavior in https://github.com/desktop/desktop/pull/7432/commits/e9abdc153608df3fee28c6fe9ef1eed6ff6661c6 is roughly equivalent if that works for you?

shiftkey: > it looks like the behavior in [e9abdc1](https://github.com/desktop/desktop/commit/e9abdc153608df3fee28c6fe9ef1eed6ff6661c6) is roughly equivalent if that works for you?

If we're fine with that, I'm happy to keep it around. Would also be good to track cleaning up any more of these usages that might be lurking around, but that's not a blocker on this PR.

outofambit: > Would also be good to track cleaning up

i can make an issue for that later today :)

PR #7441, opened 2019-05-01
"view stash" button a11y and styling

### LINKED ISSUE #7274: [Stashing] Stashed changes view is not accessible
This issue was brought up in review of #7210. 

> If a changed file is present and you tab over to the Stashed Changes tab, clicking enter does not work. I expected doing so would allow me to view diffs.
> You cannot tab or key over to Stashed changes consistently, making it not accessible.
-- @tierninho 

☝️ I'm using <kbd>Tab</kbd> to cycle through these elements.

outofambit:
## Overview

Closes #7274

Tabbed onto:
Screen Shot 2019-05-06 at 9 38 50 AM

## Description

- used `button` instead of a `div` for better semantics and tab ordering
- pulled out FocusContainer since it wasn't being used for anything unique to itself
- made the stash toggle button its own style since neither existing list item or button styles were quite right for it (and using them required overriding certain aspects)


outofambit:
needs a couple design tweaks:
1: the icon on selected should be white, and 2: the arrow should be right aligned

PR #7445, opened 2019-05-01
detect conflicts in background when previewing rebase

### LINKED ISSUE #6960: Warn if there will be conflicts prior to starting a rebase

shiftkey:
## Overview

**Closes #6960**

## Description

This PR adds a feature-flag to test whether the commits to be reapplied as part of the rebase will apply cleanly, and report if that isn't the case - this is an indication there will be conflicts to resolve. There's a bunch of work hardening, polishing and testing still to go but I thought I'd open this up early to share how this solution works.

First, what problems were there to address:

 - we need a reliable way to test the commits without performing the whole rebase itself
 - we want to get feedback reasonably quickly if there will problems
 - we want to avoid making changes to the working directory, the index or the current branch

I've settled on a workflow here that uses Git worktrees, which we don't currently support in the app. That's fine, because all of this is internal currently.

I settled on this because it addresses the second two items on that previous list:

 - `git worktree add` shares the same `.git` directory from the main worktree, so we don't need to reclone the repository, 
 - in this new worktree we can change the `HEAD` commit to another branch (or use as a detached HEAD) and the index/working directory state is kept separate to the main worktree

With a worktree for us to maniplate and test independent of the main repository, the rest of the flow becomes pretty straightforward:

 - we checkout the base branch in the new worktree 
 - we run `git format-patch` to generate the patch series that the rebase will perform and emit this to `stdout`
 - we then can use `git apply` with the patch series against the base branch to see if it can be applied cleanly

## Review Plan

As this PR touches a few areas, I'm going to break them up into a series PRs to make reviews more focused. The tentative plan:

 - [ ] introduce abstraction `OnDiskRepository` that is implemented by `Repository` and `WorkTree` - #7819 
 - [ ] introduce worktrees Git APIs and tests
 - [ ] introduce patch format and apply Git and tests
 - [ ] ???

## TODO

 - [ ] more testing
 - [ ] caching to prevent user re-running same rebase preview
 - [ ] ???
 - [ ] docs

## Release notes

Notes: rebase flow will now warn the user if conflicts are expected


shiftkey:

PR #7460, opened 2019-05-03
 cleanup remotes sequentially to avoid lockfile issue

shiftkey:
## Overview

Found this bug while investigating some performance issues, and wanted to review/land this independent of that work as it's a really specific fix

## Description

The repro to this step is to add a lot of remotes from forks that are no longer used, using the `github-desktop-` prefix. After running `fetchAndCachePullRequests` (every 10mins or manually), the app also runs `pruneForkedRemotes` to find any remotes that are no longer associated with forks.

With the current implementation, it creates a bunch of promises (spawning Git processes to work concurrently) and then waits for them to complete.

The problem here is that Git appears to also need to update `.git/packed-refs`, and concurrent Git operations to update this file can trigger an error (like what happens with the index). [More details about packed refs can be found here](https://git-scm.com/docs/git-pack-refs).

This is obviously an extreme case, but we can be better here and run this work sequentially cleaning them up.

Here's a bunch of logs showing the problem:

## Release notes

Notes:

### DIFF
--- a/app/src/lib/stores/pull-request-store.ts
+++ b/app/src/lib/stores/pull-request-store.ts
@@ -187,10 +187,9 @@ export class PullRequestStore extends TypedBaseStore<GitHubRepository> {
     repository: Repository,
     remotes: ReadonlyArray<IRemote>
   ) {
-    const promises: Array<Promise<void>> = []
-
-    remotes.forEach(r => promises.push(removeRemote(repository, r.name)))
-    await Promise.all(promises)
+    for (const remote of remotes) {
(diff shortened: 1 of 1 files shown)


niik:
Hey @shiftkey, turns out I inadvertently solved the same problem in #7501 wherein I also moved the pruning responsibility off of the pull request store. Given that that PR is going to conflict with this would you mind if we close this in favor of #7501?

shiftkey:
@niik I'll have a look at what's happening in #7501 in the morning and confirm this PR is unnecessary

PR #7438, opened 2019-04-30
Baseline stashing metrics

iAmWillShepherd:
partly addresses #7246 

This pr has been updated to adjust the basline stashing metrics implemented in #7470 by merging `getDesktopStashEntries` and `getStash`.

PR #7470, opened 2019-05-06
Reimplement baseline stashing metrics on top of last prod release

iAmWillShepherd:
Get's metrics in place so we have a baseline for future metrics that will be added to address https://github.com/desktop/desktop/issues/7246.

This PR adds two new metrics:
* `stashEntriesCreatedOutsideDesktop` - the number of stash entries we know about for the day. 
* `errorWhenSwitchingBranchesWithUncommmittedChanges` - incremented when the user experiences an error trying to check out a new branch with uncommitted changes

The `stashEntriesCreatedOutsideDesktop` metric is a bit lazy because its value depends on which repositories a user opens during the day. So if you have 100 stashed in your-favorite-repo, but you never open that repo, those stash entries are not counted towards this metric. I chose this approach to limit the amount of background work we do. I'm open to changing the behaviour of this function if needed.

**Review checklist**
- [x] Review https://github.com/github/central/pull/432
- [x] Review https://github.com/github/desktop.github.com/pull/131

**Testing** - https://github.com/desktop/desktop/pull/7438#issuecomment-488133619.

Related PR:  #7438

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.6.5-beta1",
+  "version": "1.6.5-beta2",
   "main": "./main.js",
   "repository": {
     "type": "git",
--- a/app/package.json
+++ b/app/package.json
@@ -26,7 +26,7 @@
     "deep-equal": "^1.0.1",
     "dexie": "^2.0.0",
     "double-ended-queue": "^2.1.0-0",
-    "dugite": "1.82.0",
+    "dugite": "1.87.0",
     "electron-window-state": "^4.0.2",
     "event-kit": "^2.0.0",
     "file-uri-to-path": "0.0.2",
--- a/app/package.json
+++ b/app/package.json
@@ -26,7 +26,7 @@
     "deep-equal": "^1.0.1",
     "dexie": "^2.0.0",
     "double-ended-queue": "^2.1.0-0",
-    "dugite": "1.80.0",
+    "dugite": "1.82.0",
     "electron-window-state": "^4.0.2",
     "event-kit": "^2.0.0",
     "file-uri-to-path": "0.0.2",
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.6.4",
+  "version": "1.6.5-beta1",
   "main": "./main.js",
   "repository": {
     "type": "git",
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.6.5-beta2",
+  "version": "1.6.5",
   "main": "./main.js",
   "repository": {
     "type": "git",
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -267,7 +267,7 @@ function getDescriptionForError(error: DugiteError): string {
     case DugiteError.NoExistingRemoteBranch:
       return 'The remote branch does not exist.'
     case DugiteError.LocalChangesOverwritten:
-      return 'Unable to switch branches as there are working directory changes that would be overwritten. Please commit or stash your changes.'
+      return 'Unable to switch branches as there are working directory changes which would be overwritten. Please commit or stash your changes.'
     case DugiteError.UnresolvedConflicts:
       return 'There are unresolved conflicts in the working directory.'
     default:
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -267,7 +267,7 @@ function getDescriptionForError(error: DugiteError): string {
     case DugiteError.NoExistingRemoteBranch:
       return 'The remote branch does not exist.'
     case DugiteError.LocalChangesOverwritten:
-      return 'Some of your changes would be overwritten.'
+      return 'Unable to switch branches as there are working directory changes that would be overwritten. Please commit or stash your changes.'
     case DugiteError.UnresolvedConflicts:
       return 'There are unresolved conflicts in the working directory.'
     default:
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -268,6 +268,8 @@ function getDescriptionForError(error: DugiteError): string {
       return 'The remote branch does not exist.'
     case DugiteError.LocalChangesOverwritten:
       return 'Some of your changes would be overwritten.'
+    case DugiteError.UnresolvedConflicts:
+      return 'There are unresolved conflicts in the working directory.'
     default:
       return assertNever(error, `Unknown error: ${error}`)
   }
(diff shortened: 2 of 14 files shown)


app/package.json
```diff
@@ -26,7 +26,7 @@
     "deep-equal": "^1.0.1",
     "dexie": "^2.0.0",
     "double-ended-queue": "^2.1.0-0",
-    "dugite": "1.80.0",
+    "dugite": "1.82.0",
```
outofambit: we should skip ahead to https://github.com/desktop/dugite/releases/tag/v1.87.0 since that has all the error handling without the bad version of git for windows that required a hotfix last time.
```suggestion
    "dugite": "1.87.0",
```

iAmWillShepherd:
Removed `ready-for-review` label to figure out why tests are now broken.

app/src/lib/stores/app-store.ts
```diff
@@ -1139,6 +1141,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return Promise.resolve(null)
     }
 
+    this.updateStashEntryCountMetric(repository)
```
shiftkey: The problem you have currently is that the test repositories aren't being inserted into the `RepositoriesStore` as part of the setup, and this code path is being tripped when `AppStore._selectRepository` fires in many cases - sometimes in the test, sometimes in `AppStore.updateRepositorySelectionAfterRepositoriesChanged()` which fires after it detects the list of repositories has updated.

I'm going to spend a bit today trying to get these tests under control, but I'm curious if there's a a better place for us to do this check than "when the repository is selected".

shiftkey: #7484

app/src/lib/stores/app-store.ts
```diff
@@ -3576,6 +3579,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  private async updateStashEntryCountMetric(repository: Repository) {
+    const lastStashEntryCheck = await this.repositoriesStore.getLastStashCheckDate(
+      repository
+    )
+    const dateNow = moment()
+    const threshold = dateNow.subtract(24, 'hours')
+    if (lastStashEntryCheck == null || threshold.isAfter(lastStashEntryCheck)) {
+      // `lastStashEntryCheck` being equal to null means we've never checked for
+      // the given repo
+      const stashSize = await getStashSize(repository)
```
shiftkey: We don't have any error handling in here for unexpected exit codes from the Git operation, and given this is run when the user switches repositories it could be easy to trigger? How defensive do we want to be in here?

app/src/lib/stores/repositories-store.ts
```diff
@@ -178,14 +179,16 @@ export class RepositoriesStore extends BaseStore {
       )
     }
 
-    const gitHubRepositoryID = repository.gitHubRepository
-      ? repository.gitHubRepository.dbID
-      : null
+    const oldRecord = await this.db.repositories.get(repoID)
```
shiftkey: Can we tidy up these checks in existing code if you're satisfied with #7484?

app/src/lib/stores/app-store.ts
```diff
@@ -193,6 +193,8 @@ import {
 import { BranchPruner } from './helpers/branch-pruner'
 import { enableBranchPruning, enablePullWithRebase } from '../feature-flag'
 import { Banner, BannerType } from '../../models/banner'
+import moment = require('moment')
```
```suggestion
import * as moment from 'moment'
```

app/src/lib/git/core.ts
```diff
@@ -264,6 +264,12 @@ function getDescriptionForError(error: DugiteError): string {
       return 'A lock file already exists in the repository, which blocks this operation from completing.'
     case DugiteError.NoMergeToAbort:
       return 'There is no merge in progress, so there is nothing to abort.'
+    case DugiteError.NoExistingRemoteBranch:
+      return 'The remote branch does not exist.'
+    case DugiteError.LocalChangesOverwritten:
+      return 'Some of your changes would be overwritten.'
```
shiftkey: Screen Shot 2019-05-09 at 9 23 26 AM

Can we make this message a bit more descriptive? Something like:

> Unable to switch branches as there are working directory changes that would be overwritten. Please commit or stash your changes.

shiftkey:
I've tested both interactions related to the metric:

 - selecting a repository to check the last stash update (and maybe update a metric)
 - triggering a checkout which would overwrite working directory changes (and fails as expected)  increments a metric

As I've been involved closely with this PR, I'd love another pair of eyes on this review.

app/src/lib/git/core.ts
```diff
@@ -264,6 +264,12 @@ function getDescriptionForError(error: DugiteError): string {
       return 'A lock file already exists in the repository, which blocks this operation from completing.'
     case DugiteError.NoMergeToAbort:
       return 'There is no merge in progress, so there is nothing to abort.'
+    case DugiteError.NoExistingRemoteBranch:
+      return 'The remote branch does not exist.'
+    case DugiteError.LocalChangesOverwritten:
+      return 'Unable to switch branches as there are working directory changes that would be overwritten. Please commit or stash your changes.'
```
shiftkey: Minor wording tweak:

```suggestion
      return 'Unable to switch branches as there are working directory changes which would be overwritten. Please commit or stash your changes.'
```

cc @billygriffin for any other wordsmithery feedback

app/src/lib/stores/app-store.ts
niik: Hey @iAmWillShepherd, me and @billygriffin just looked through this today in our sync-up and noticed that there's a key difference between this and what's being proposed for 1.7 (https://github.com/desktop/desktop/pull/7438) in that the 1.7 release will trigger `updateStashEntryCountMetric` from `_refreshRepository` whereas this is called from `_selectRepository`.

The potential problem that we see with this is for the subset of our users that have a workflow similar to my own, i.e. they work primarily in one repository (rarely switch) and they keep their desktop instances running for a long time (weeks in my case) this approach will lead to us checking and incrementing the count pretty infrequently whereas the proposed solution for 1.7 we will check once every 24 hours regardless of whether they switch or not.

The end result here could be that if the segment we're talking about isn't trivial in size (I would personally guess that it's a decent chunk of our user base) we'd see a spike in stashes created outside of desktop once we ship 1.7 which would be the opposite of what we'd want.

I would propose we amend the 1.6.6 branch with a change that moves this check to the same place as it will be in 1.7 in an effort to avoid collection inconsistencies between versions.

iAmWillShepherd: > I would propose we amend the 1.6.6 branch with a change that moves this check to the same place as it will be in 1.7 in an effort to avoid collection inconsistencies between versions.

I'd rather move the check in the 1.7 PR to `selecectRepository`. For context, the PR from 1.6.6 initially had the check in `refreshRepository`, but I moved it because that code path get's ran a lot and doing DB queries isn't all that fast.

I'm going to link this comment in the PR for 1.7 to make sure I make this change.

billygriffin: @iAmWillShepherd It doesn't seem like you addressed @niik's feedback here.
> The potential problem that we see with this is for the subset of our users that have a workflow similar to my own, i.e. they work primarily in one repository (rarely switch) and they keep their desktop instances running for a long time (weeks in my case) this approach will lead to us checking and incrementing the count pretty infrequently whereas the proposed solution for 1.7 we will check once every 24 hours regardless of whether they switch or not.

> The end result here could be that if the segment we're talking about isn't trivial in size (I would personally guess that it's a decent chunk of our user base) we'd see a spike in stashes created outside of desktop once we ship 1.7 which would be the opposite of what we'd want.

Issue #7464, opened 2019-05-03
performance wins to investigate for the next release


shiftkey:
**Please describe the problem you think should be solved**

We got some feedback from some of our beta users that the app is not as performant as expected after the latest update. I'd like to propose investigating and seeing what we can address before the next update.

The feedback was focused on two areas:

  - macOS, repository with ~8000 refs, application is very sluggish
  - Windows, new stash workflows are not as performant as expected

My personal goals as part of this investigation:

 - skip unnecessary work where possible
 - tune the work we have to do so it runs better under extreme cases
 - better guard against concurrent refreshing

**[Optional] Do you have any potential solutions in mind?**

I paired on this today with @iAmWillShepherd and @outofambit to talk through what we've seen and what we could focus on in the short term, and we came up with some areas to further explore and optimize.

These items are grouped based on when we believe they were introduced into the codebase, with regressions in the beta releases prioritized before problems that exist in the current production release:

### Overlapping `getStatus` calls introduce a bottleneck - #7471 - 🚢 

We identified these chain of calls into `getStatus` that can overlap:

- `checkoutBranch` -> `AppStore.refreshRepository` -> `AppStore._loadStatus` -> `gitStore.getStatus`
- `popStashEntry` -> `AppStore.refreshRepository` -> `AppStore.refreshIndicatorsForRepositories` -> `AppStore.refreshIndicatorForRepository` -> `AppStore._loadStatus` -> `gitStore.getStatus`

`getStatus` itself performs a lot of additional work to `git status`:

 - checking for merge/rebase conflict state
 - checking for conflicted files in working directory
 - checking for binary files in working directory

This feels like a situations where a "refresh" doesn't need to refresh everything, and callers might want to opt-out of updating indicators (or pass in the known state to do the updating).

### Reuse `getStatus` result when refreshing to update indicators - #7475 

Building on the above problem, calling `AppStore._refreshRepository`  will result in two `AppStore._loadStatus` calls:

The first is in `refreshRepository` itself:

https://github.com/desktop/desktop/blob/189d17f44a4744db26e1ce69b658b9cc98f3aba1/app/src/lib/stores/app-store.ts#L2486

The second is in `AppStore.refreshIndicatorsForRepositories` (the original intent here was to have the same behaviour whether this was an active repository or not):

https://github.com/desktop/desktop/blob/189d17f44a4744db26e1ce69b658b9cc98f3aba1/app/src/lib/stores/app-store.ts#L2522

[...]

PR #7474, opened 2019-05-07
Refactors the stashing flow a bit to reduce unnecessary work 

iAmWillShepherd:
Address some concerns brought up in  #7464  under the heading "Only stash when there are problems with checkout" by attempting to implement the flow describe in https://github.com/desktop/desktop/issues/7464#issuecomment-489787219. 

The following are the different flows this PR implements:

**New branch is not ahead of the current branch (ideal)**
1. Perform checkout and rely on Git's default behavior

**New branch is ahead of the current branch; create stash (worst case)**
1. Perform the checkout (this will result in an error)
1. Inspect the error to ensure it's `DugiteError.LocalChangesOverwritten`
1. Create temporary stash
1. Perform checkout (should work since working directory is clean at this point)
1. Pop the entry if the branch checkout succeeds

--- | Create stash on every checkout/Deleted files exist in the index | New branch is not ahead of the current branch| New branch is ahead of the current branch |
|:--:|:--:|:--:|:--:|
| Checkout | - | 180ms | 150ms (error thrown) |
| Stash | 150ms | - | 150ms |
| Checkout | 180ms | - | 180ms |
| Pop | 155ms | - | 155ms |
| Total | 485ms | 180ms (**-403%**) | 635ms (**31%**) |

_Note, all values have been rounded_

This PR differs from the flow referenced above in one key way: we resort to creating a temporary stash when there are deleted files in the working directory. What follows is the flow when deleted files exist in the working directory.

**Deleted files exist**
1. Create a stash entry
2. Checkout new branch
3. Immediately apply the stash entry

_Same performance as the original _Create stash on every checkout_ flow._

Since we are introducing a new flow that relies on creating a temporary stash, `CheckoutBranch` has been modified to return a value so that we can use it in `AppStore._checkoutBranch` to determine if checkout succeeded so that we can pop the temporary stash entry if necessary. 

cc @shiftkey, @outofambit , @billygriffin


app/src/lib/stores/app-store.ts
```diff
@@ -5020,7 +5027,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
   public async _createStash(
     repository: Repository,
     branchName: string,
-    removePreviousStash: boolean = true
+    preAction:
```
iAmWillShepherd: Open to suggestions on better names for these two callbacks. My intent is to give the the calling code more control over what happens before and after a stash entry has been created.

app/src/lib/stores/app-store.ts
```diff
@@ -5031,23 +5041,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
 
-    await dropDesktopStashEntry(repository, previousStash.stashSha)
-    log.info(
-      `Dropped stash '${previousStash.stashSha}' associated with ${
-        previousStash.branchName
-      }`
-    )
+      // if we can't find the stash we just created, something went wrong
+      if (previousStashEntry === null) {
+        throw Error('Something bad happened')
```
iAmWillShepherd: Any suggestions on useful info to put in this message would be helpful. This is a truly exceptional case and I don't know what getting `null` would mean.

app/src/models/uncommitted-changes-strategy.ts
```diff
@@ -1,5 +1,15 @@
-export enum UncommittedChangesStrategy {
+import { IStashEntry } from './stash-entry'
+
+export enum UncommittedChangesStrategyKind {
   askForConfirmation,
   stashOnCurrentBranch,
   moveToNewBranch,
 }
+
+export type UncommittedChangesStrategy =
```
iAmWillShepherd: These type names are descriptive, but super long. I'm open to suggestions.

app/src/lib/stores/app-store.ts
```diff
@@ -2729,7 +2732,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
     repository: Repository,
     name: string,
     startPoint: string | null,
-    uncommittedChangesStrategy: UncommittedChangesStrategy = UncommittedChangesStrategy.askForConfirmation
+    uncommittedChangesStrategy: UncommittedChangesStrategy = {
```
iAmWillShepherd: There are a few other places where we use an `UncommitedChangesStrategy` object literal, is it worth pulling this off into a constant to clean up this function signature?

app/src/lib/stores/app-store.ts
shiftkey: Rather than using `pre` and `post` here, I think these could be more explicit:

 - `onPreviousStashFound`
 - `onNewStashCreated`

This means they both can use the same signature, and it becomes a bit more intuitive than "before" and "after" in terms of what these are intended for.

app/src/lib/stores/app-store.ts
shiftkey:
```suggestion
        throw new Error('The stash was not created. Check the logs for an underlying Git error because this shouldn't happen')
```

app/src/models/uncommitted-changes-strategy.ts
```diff
@@ -1,5 +1,15 @@
-export enum UncommittedChangesStrategy {
+import { IStashEntry } from './stash-entry'
+
+export enum UncommittedChangesStrategyKind {
   askForConfirmation,
```
shiftkey: Finally time to update these to be SentenceCase?

app/src/models/uncommitted-changes-strategy.ts
shiftkey: I'm not too worried about this, but will keep this in mind when doing a proper review

app/src/lib/stores/app-store.ts
iAmWillShepherd: `onPreviousStashFound`'s signature actually can't be different since `getLastDesktopStashEntryForBranch` returns `null` if there is no previous stash. 

https://github.com/desktop/desktop/blob/139dad99ac256d6a9d74ef31614e95fa149be96d/app/src/lib/stores/app-store.ts#L5044-L5050

`onNewStashCreated`'s signature only gets a stash entry for sure because I verify it's there because it really should be there after just being created.

https://github.com/desktop/desktop/blob/139dad99ac256d6a9d74ef31614e95fa149be96d/app/src/lib/stores/app-store.ts#L5054-L5066

outofambit:
While this approach should make switching branches without errors _faster_ by avoiding two git stash operations, i want to note it will most likely make switching branches with errors _slower_.

Here's the before and after for that case.

**Before:**
1. create stash
2. checkout
3. pop stash

**After:**
1. attempt checkout — error encountered
2. create stash
3. checkout
4. pop stash

This is _probably_ the right approach, making the (probably) majority case faster and the (probably) minority case slower. (I'm saying probably here, cause I haven't spent the time to have solid analysis or data to back up this hunch.)

I think its also worth thinking about how we might communicate this to the user so they understand why bringing their changes with them might take a bit longer than the usual switch branches.

iAmWillShepherd:
> I think its also worth thinking about how we might communicate this to the user so they understand why bringing their changes with them might take a bit longer than the usual switch branches.

I definitely agree with that.

outofambit:
> While this approach should make switching branches without errors faster by avoiding two git stash operations, i want to note it will most likely make switching branches with errors slower.

So, I talked through this a bit more with @billygriffin and I think its fine to ignore this concern for this PR. The cases in which an error will be encountered during checkout is definitely a subset of all "bringing changes with you to a branch" workflows. That is to say, it will be slower in those cases, but those cases are most likely quite uncommon.

app/src/lib/stores/app-store.ts
shiftkey: > `onPreviousStashFound`'s signature actually can't be different since `getLastDesktopStashEntryForBranch` returns `null` if there is no previous stash.

Sure, but the only usage I found in the codebase of this parameter was performing some work _only if the previous stash was found_. That felt like an implicit part of the contract of the API: "register a callback to do some work only when a previous stash was found". 

What if we extracted that "function returning promise to its own type alias"?

```ts
export type StashCallback = (stashEntry: IStashEntry) => Promise<void>
```

We could make `AppStore._createStash` only run the callback if both the callback was set and it found a valid previous stash (hence the name `onPreviousStashEntryFound`)

```ts
    if (onPreviousStashEntryFound != null) {
      const previousStashEntry = await getLastDesktopStashEntryForBranch(
        repository,
        branchName
      )

      if (previousStashEntry !== null) {
        await onPreviousStashEntryFound(previousStashEntry)
      }
    }
```

And we could tidy up the signatures to be a bit more clearer:

```ts
  public async _createStash(
    repository: Repository,
    branchName: string,
    onPreviousStashEntryFound: StashCallback | null,
    onNewStashCreated: StashCallback | null
  ) {
```

app/src/models/uncommitted-changes-strategy.ts
shiftkey: Not sure why this was resolved, but it looks like the enum was renamed to be `lowerCamelCase` which is different to our other enums 🤔

app/src/lib/stores/app-store.ts
iAmWillShepherd: Addressed in 71059d3.

shiftkey:
@iAmWillShepherd looks like we've got some conflicts now that #7471 has been merged

app/src/lib/error-with-metadata.ts
```diff
@@ -15,6 +16,9 @@ export interface IErrorMetadata {
 
   /** Additional context that specific actions can provide fields for */
   readonly gitContext?: GitErrorContext
+
+  /** The branch intended to be checked out */
+  readonly checkoutBranch?: Branch
```
outofambit: I think this should be in a new type of `GitErrorContext` instead of added directly in `IErrorMetadata` since this field is very specific to bringing changes during switching branches.

app/src/lib/stores/app-store.ts
```diff
@@ -5018,29 +5027,44 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
   }
 
-  /**
-   *  Note: **drops the existing desktop stash** on that branch (if there is one)
-   */
-  private async createStash(repository: Repository, branchName: string) {
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _createStash(
```
outofambit: I don't think the callback pattern is serving us here. `_createStash` is only called in two places, and in each place one of the callbacks is omitted. This means we're only sharing this line between the two codepaths:

```js
await createDesktopStashEntry(repository, branchName)
```

I think splitting this into two methods would be much less complex and would let us avoid inlining callback methods in `error-handlers.ts` and `_checkoutBranch`.

iAmWillShepherd: Addressed in f5edd65.

app/src/lib/error-with-metadata.ts
iAmWillShepherd: Addressed in 7717a2b.

iAmWillShepherd:
I'm also running into issues that prevent the OverwriteStash dialog from popping up. I've been investigating the root cause, but have yet to figure it out.

shiftkey:
> I couldn't spot anything obvious from the logs, but I'm going to look into this more tomorrow because it'd be great to wrap this flow in some automated tests and give ourselves confidence in this refactoring.

To add some extra context to this issue, this seems to be a limitation of `git checkout` and how it handles deleted tracked files. I've been able to reproduce this on the command line when a deleted file has changed in the other branch:

```shellsession
$ rm .travis.yml
$ git status
On branch development
Your branch is up to date with 'upstream/development'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	deleted:    .travis.yml

no changes added to commit (use "git add" and/or "git commit -a")

$ git checkout linux
Switched to branch 'linux'
Your branch is up to date with 'origin/linux'.

$ git status
On branch linux
Your branch is up to date with 'origin/linux'.

```

It seems to be that we need to stage these deletes before performing the `git checkout`  to correctly . detect the conflicts.

```shellsession
$ rm .travis.yml
$ git add .travis.yml
$ git status
On branch development
Your branch is up to date with 'upstream/development'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	deleted:    .travis.yml

$ git checkout linux
error: Your local changes to the following files would be overwritten by checkout:
	.travis.yml
Please commit your changes or stash them before you switch branches.
Aborting
$ git status
On branch development
Your branch is up to date with 'upstream/development'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	deleted:    .travis.yml

```

Modifying a tracked file doesn't need to be staged for it to correctly warn:

```shellsession
$ echo "add something" >> .travis.yml
$ git status
On branch development
Your branch is up to date with 'upstream/development'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   .travis.yml

no changes added to commit (use "git add" and/or "git commit -a")
$ git checkout linux
error: Your local changes to the following files would be overwritten by checkout:
	.travis.yml
Please commit your changes or stash them before you switch branches.
Aborting

```

app/src/lib/stores/app-store.ts
```diff
@@ -2837,40 +2841,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-    ) {
-      this._showPopup({
-        type: PopupType.StashAndSwitchBranch,
-        branchToCheckout: foundBranch,
-        repository,
-      })
-
-      return repository
-    }
+    let transientStashCreated = false
```
iAmWillShepherd: I'm not a fan of this function, `_checkoutBranch` in its current form because it's doing too much: checking out a branch, asking to stash, stashing changes on current branch, popping transient stash entry. I think it's worthing thinking about how to split this logic up so that it's clearer, but I don't think this PR is the place for that. So if you're reviewing and think this logic is complex, know that I likely agree with you, but I think it's mostly fine for now since it's all grouped together behind a check for `enableStashing`

app/src/lib/stores/app-store.ts
```diff
@@ -2887,26 +2899,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
     )
 
-    if (shouldPopStash) {
-      const stash = await getLastDesktopStashEntryForBranch(
-        repository,
-        foundBranch.name
-      )
-
-      if (stash !== null) {
+    if (
```
shiftkey: Are we concerned here about proceeding with this flow if `checkoutBranch` fails? This is handled in our Git error handling, but we don't poke at the result returned from `performFailableOperation`?

iAmWillShepherd: I'm going to put a fix in place for now with the intent to address it next week when we pair on simplifying this function a bit.

app/src/lib/stores/app-store.ts
```diff
@@ -2843,76 +2847,88 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
       ) {
-        await createDesktopStashEntry(repository, foundBranch.name)
-        shouldPopStash = true
+        const hasDeletedFiles = changesState.workingDirectory.files.some(
+          file => file.status.kind === AppFileStatusKind.Deleted
+        )
+        if (hasDeletedFiles) {
+          await gitStore.performFailableOperation(() => {
+            return createDesktopStashEntry(repository, foundBranch.name)
+          })
```
niik: So it seems to me like we're gonna go ahead and load the last desktop stash entry regardless of whether or not the `createDesktopStashEntry` succeeded or not here given that we're wrapping the `createDesktopStashEntry` call in `performFailableOperation` but not verifying whether it succeeded or not which seems like it could lead to us popping the wrong stash entry.

mido1966:
Pls No Email mor pls Tank you



Von meinem Samsung Galaxy Smartphone gesendet.


-------- Ursprüngliche Nachricht --------
Von: Markus Olsson <notifications@github.com>
Datum: 22.05.19 14:15 (GMT+01:00)
An: desktop/desktop <desktop@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [desktop/desktop] Refactors the stashing flow a bit to reduce unnecessary work  (#7474)


@niik requested changes on this pull request.

Did an initial pass over this and left some questions and suggestions.

In app/src/ui/dispatcher/error-handlers.ts<https://github.com/desktop/desktop/pull/7474#discussion_r286419007>:

> @@ -437,7 +446,21 @@ export async function localChangesOverwrittenHandler(

     return error

   }



-  dispatcher.recordErrorWhenSwitchingBranchesWithUncommmittedChanges()

+  if (!enableStashing()) {

+    dispatcher.recordErrorWhenSwitchingBranchesWithUncommmittedChanges()



This confuses me a bit but I'm probably just lacking context here. In #7246 (comment)<https://github.com/desktop/desktop/issues/7246#issuecomment-487410668> the success criteria for this metric is stated as

This should be > 50% lower. We expect that the way in which we implemented the ability to bring changes with you should provide a nicer experience and fewer error messages.

With this change it seems we're setting ourselves up for an artificial success here since the moment we turn on the stashing feature flag we'll stop recording this metric. Looking further down in this function it would appear as if there are at least two cases where we return the error for the user to see.

In app/src/ui/dispatcher/error-handlers.ts<https://github.com/desktop/desktop/pull/7474#discussion_r286422953>:

>

-  return error

+  if (gitContext == null) {



It took me some time to understand how it could be safe to always invoke dispatcher.moveChangesToBranchAndCheckout when we see this particular error so I think it makes sense to document the context here and at the same time wrap up the check below into this if-statement.

⬇️ Suggested change

-  if (gitContext == null) {

+  // This indicates to us whether the action which triggered the

+  // LocalChangesOverwritten was the AppStore _checkoutBranch method.

+  // Other actions that might trigger this error such as deleting

+  // a branch will not provide this specific gitContext and that's

+  // how we know we can safely move the changes to the destination

+  // branch.

+  if (gitContext === undefined || gitContext.kind !== 'checkout') {



In app/src/ui/dispatcher/error-handlers.ts<https://github.com/desktop/desktop/pull/7474#discussion_r286425444>:

>

-  return error

+  if (gitContext == null) {

+    return error

+  }

+

+  if (gitContext.kind !== 'checkout') {



If we apply my suggested change above we can remove this if-statement.

In app/src/lib/stores/app-store.ts<https://github.com/desktop/desktop/pull/7474#discussion_r286446897>:

>

-      return repository

-    }

+    let createdStashEntry = null



This is a bit confusingly named since we might create a stash entry but not assign anything to this variable (in case of UncommittedChangesStrategyKind.stashOnCurrentBranch. Further down we assign this value or the value of the transientStashEntry to a variable called stashToPop which I find more fitting. Could we perhaps rename this to stashToPop and initially assign to it the value of uncommittedChangesStrategy.transientStashEntry?

In app/src/lib/stores/app-store.ts<https://github.com/desktop/desktop/pull/7474#discussion_r286457416>:

>        ) {

-        await createDesktopStashEntry(repository, foundBranch.name)

-        shouldPopStash = true

+        const hasDeletedFiles = changesState.workingDirectory.files.some(

+          file => file.status.kind === AppFileStatusKind.Deleted

+        )

+        if (hasDeletedFiles) {

+          await gitStore.performFailableOperation(() => {

+            return createDesktopStashEntry(repository, foundBranch.name)

+          })



So it seems to me like we're gonna go ahead and load the last desktop stash entry regardless of whether or not the createDesktopStashEntry succeeded or not here given that we're wrapping the createDesktopStashEntry call in performFailableOperation but not verifying whether it succeeded or not which seems like it could lead to us popping the wrong stash entry.

mido1966:
Pls No Email mor pls Tank you



Von meinem Samsung Galaxy Smartphone gesendet.


-------- Ursprüngliche Nachricht --------
Von: evelyn masso <notifications@github.com>
Datum: 22.05.19 00:09 (GMT+01:00)
An: desktop/desktop <desktop@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [desktop/desktop] Refactors the stashing flow a bit to reduce unnecessary work  (#7474)


@outofambit commented on this pull request.

i think this is good to go. this achieves the stated performance optimizations and in my testing doesn't introduce any regressions.

there are some small refactors we could do to simplify (and shorten) AppStore._checkoutBranch, but i don't think that should be a blocker to getting this out. (and @iAmWillShepherd<https://github.com/iAmWillShepherd> is going to file a followup issue to capture that work).

app/src/lib/stores/app-store.ts
```diff
@@ -2843,76 +2847,88 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-    ) {
-      this._showPopup({
-        type: PopupType.StashAndSwitchBranch,
-        branchToCheckout: foundBranch,
-        repository,
-      })
 
-      return repository
-    }
+    let createdStashEntry = null
```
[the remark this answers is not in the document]
iAmWillShepherd: I like that idea. Unfortunately, we cannot initialize it with the value of `uncommittedChangesStrategy.transientStashEntry` because the `transientStashEntry` property is when `uncommittedChangesStrategy.kind` is `UncommittedChangesStrategyKind.moveToNewBranch`.

I'm 👍 on renaming this variable though. If I do so, L2926 will look like the following code snippet:

```ts
stashToPop = stashToPop || uncommittedChangesStrategy.transientStashEntry
      if (stashToPop !== null) {
        await gitStore.performFailableOperation(() => {
          // we have to use the `!` operator since TS thinks this object can be null...
          return popStashEntry(repository, stashToPop!.stashSha)
        })
      }
```

If you're 👍 on that change, I can go ahead and do it, otherwise, maybe I can choose another name that won't require us to use the non-null assertion.

app/src/lib/stores/app-store.ts
shiftkey: Yeah, I also think it'd be great here to have the result of `createDesktopStashEntry` set to `createdStashEntry`, rather than having this as two separate steps.

iAmWillShepherd: b1cbf11 adds a check to ensure this was successful. 

> Yeah, I also think it'd be great here to have the result of createDesktopStashEntry set to createdStashEntry, rather than having this as two separate steps.

I like this idea. Gonna open an issue to track this so it doesn't get lost after we release.

app/src/lib/stores/app-store.ts
```diff
@@ -2923,11 +2926,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
         UncommittedChangesStrategyKind.moveToNewBranch &&
       checkoutSucceeded
     ) {
-      const stashToPop =
-        createdStashEntry || uncommittedChangesStrategy.transientStashEntry
+      stashToPop = stashToPop || uncommittedChangesStrategy.transientStashEntry
       if (stashToPop !== null) {
         await gitStore.performFailableOperation(() => {
-          return popStashEntry(repository, stashToPop.stashSha)
+          return popStashEntry(repository, stashToPop!.stashSha)
```
iAmWillShepherd: This non-null assertion is required to appease the TS compiler.

shiftkey: There's a slightly better way to appease the compiler without using the `!` here:

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index dd9c7460c..2127741c8 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2928,8 +2928,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
     ) {
       stashToPop = stashToPop || uncommittedChangesStrategy.transientStashEntry
       if (stashToPop !== null) {
+        const stashSha = stashToPop.stashSha
         await gitStore.performFailableOperation(() => {
-          return popStashEntry(repository, stashToPop!.stashSha)
+          return popStashEntry(repository, stashSha)
         })
       }
     }

```

iAmWillShepherd:
🍎

app/src/lib/stores/app-store.ts
outofambit: to add some more context here, `createDesktopStashEntry` on `development` returns `void`, not information about the stash created (or if it was successful). @iAmWillShepherd has done a minimal fix in b1cbf11 (by returning `true` on success) that should address the case @niik called out and prevent us from accidentally popping the wrong stash (or trying to pop nothing).

i agree with @iAmWillShepherd that changing `createDesktopStashEntry` to return a `sha` or `stashEntry` on success is a good idea that we can address in a subsequent PR, since that requires some result parsing and validation.

app/src/lib/stores/app-store.ts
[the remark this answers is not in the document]
outofambit: I think the ideal refactor here would be to encapsulate the following code into its own private function on AppStore that returns `Promise<StashEntry | null>` so we can assign that result directly to `stashToPop`.

https://github.com/desktop/desktop/blob/49d5d54c9096b7a8b9d9b0cc93664033f11c5a04/app/src/lib/stores/app-store.ts#L2885-L2916

Will file an issue to track this refactor.

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -437,7 +446,21 @@ export async function localChangesOverwrittenHandler(
     return error
   }
 
-  dispatcher.recordErrorWhenSwitchingBranchesWithUncommmittedChanges()
+  if (!enableStashing()) {
+    dispatcher.recordErrorWhenSwitchingBranchesWithUncommmittedChanges()
```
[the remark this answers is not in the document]
outofambit: it looks like this was addressed in 7746f58. this seems like a reasonable solution here.

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -437,7 +445,20 @@ export async function localChangesOverwrittenHandler(
...
+  // branch.
+  if (gitContext === undefined || gitContext.kind !== 'checkout') {
+    dispatcher.recordErrorWhenSwitchingBranchesWithUncommmittedChanges()
+    return error
+  }
 
-  return error
+  const { branchToCheckout } = gitContext
+
+  await dispatcher.moveChangesToBranchAndCheckout(repository, branchToCheckout)
```
outofambit: i just noticed this `await` in another read through the code. i think we should not use `await` here so that we don't have a second `_checkoutBranch` executed before the first `_checkoutBranch` can finish. however this requires us to exit from the initial `_checkoutBranch` early if `checkout` is not successful. that feels like a more risky change to the behavior of `checkout`, which prior to now has never checked wether `checkout` was actually successful before making some state updates/refreshes.

since it doesn't seem like this is causing any obvious bugs i'm going to say that this is fine for now, but is something we should keep an eye on. 👀

app/src/lib/stores/app-store.ts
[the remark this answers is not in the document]
outofambit: captured in https://github.com/desktop/desktop/issues/7609

app/src/lib/stores/app-store.ts
outofambit: captured in #7607

tierninho:
Not sure if this error is associated with this PR, but I spotted it a few times. Happy to add repro steps if needed.

Screen Shot 2019-05-22 at 1 47 10 PM

49d5d54c90, mac

outofambit:
> Not sure if this error is associated with this PR, but I spotted it a few times. Happy to add repro steps if needed.

@tierninho thanks for sharing this! Could you file a new issue for that? with repro steps? if you're not seeing any weird behavior along with the error, then i think i'd rather look into that after merging this.

app/src/ui/dispatcher/error-handlers.ts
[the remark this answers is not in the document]
niik: I agree. I think it's possible that the metric itself might not have been fully thought through here.

app/src/lib/stores/app-store.ts
```diff
@@ -2860,76 +2864,91 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-        UncommittedChangesStrategy.moveToNewBranch
+        uncommittedChangesStrategy.kind ===
+        UncommittedChangesStrategyKind.moveToNewBranch
       ) {
-        await createDesktopStashEntry(repository, foundBranch.name)
-        shouldPopStash = true
+        const hasDeletedFiles = changesState.workingDirectory.files.some(
+          file => file.status.kind === AppFileStatusKind.Deleted
+        )
+        if (hasDeletedFiles) {
```
niik: Should this be

```suggestion
        if (
          hasDeletedFiles &&
          uncommittedChangesStrategy.transientStashEntry === null
        ) {
```

or is there a scenario where we want to go ahead with creating a new stash entry even if we've been passed one already?

app/src/lib/stores/app-store.ts
```diff
@@ -5058,30 +5077,52 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _moveChangesToBranchAndCheckout(
+    repository: Repository,
+    branchToCheckout: string
+  ) {
+    if (!enableStashing()) {
+      return
+    }
+
+    await createDesktopStashEntry(repository, branchToCheckout)
```
niik: Could we wrap this in a `performFailableOperation` and only continue with the checkout if the operation succeeds? If we don't I'm assuming the app will crash rather than show an error message if creating the stash fails.

app/src/lib/stores/app-store.ts
iAmWillShepherd: We shouldn't need this check because the working directory would be clean if there were a temporary stash entry at this point since the error handler calls `Dispatcher.moveChangesToBranchAndCheckout` which, after the wrapping the call on L5115 (called out in your comment below) would prevent the checkout from occurring. All that said, it doesn't hurt to have this additional safety check, so I'm going with it. 👍

iAmWillShepherd: Addressed in 89b7437

PR #7475, opened 2019-05-07
avoid invoking status when updating indicator for current repository

shiftkey:
## Overview

Related to #7464 

## Description

Whenever `AppStore._refreshRepository` runs (updating the current repository), it also runs `AppStore.refreshIndicatorsForRepositories` at the end.

Internally this does two expensive things:

 - run `AppStore._loadStatus()` to see whether there are working directory changes
 - run `AppStore.fetch()` to background fetch and detect the ahead/behind of the current branch

This feels rather wasteful to do for the active repository, and we already have a `AppStore._loadStatus()` result from earlier in `AppStore._refreshRepository` that has sufficient information for this UI (are there working changes?).

Given this function opts out of fetching while refreshing, and we have a relatively recent `getStatus` result, this PR introduces a more explicit `AppStore.updateSidebarIndicator` that is now used in `refreshRepository`, and leaves the existing behaviour in `AppStore.refreshAllSidebarIndicators` to be run when the background refresh happens.

This PR eliminates three Git operations that run on every refresh (a `git status` and two `git diff`).

On a `desktop/desktop` repository with a clean working directory:

**Before**

**After**

On a `nodejs/node` repository with 33k files removed in the working directory:

**Before**

**After**

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2487,7 +2487,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const status = await this._loadStatus(repository)
     if (!status) {
       await this._updateRepositoryMissing(repository, true)
-      this.updateRepositoryIndicator(repository, null)
+      this.updateSidebarIndicator(repository, null)
       return
     }
 
@@ -2521,13 +2521,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.updateMenuItemLabels(latestState)
 
     this._initializeCompare(repository)
-    this.updateRepositoryIndicator(repository, status)
+    this.updateSidebarIndicator(repository, status)
   }
 
   /**
    * Update the repository sidebar indicator for the repository
    */
-  private async updateRepositoryIndicator(
+  private async updateSidebarIndicator(
     repository: Repository,
     status: IStatusResult | null
   ): Promise<void> {
@@ -2552,7 +2552,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   /**
    * Refresh sidebar indicators for the set of repositories tracked in the app.
    */
-  public async refreshAllIndicators() {
+  public async refreshAllSidebarIndicators() {
     const startTime = performance && performance.now ? performance.now() : null
 
     // keep a reference to the current set of repositories to avoid the array
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2549,20 +2549,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
     })
   }
 
-  public refreshAllIndicators() {
-    return this.refreshIndicatorsForRepositories(this.repositories)
-  }
-
   /**
-   * Refresh in-memory indicators for a set of repositories
-   *
-   * @param repositories the set of repositories to update
-   * @param tryBackgroundFetch whether the action should also try and fetch new changes from the remote
+   * Refresh sidebar indicators for the set of repositories tracked in the app.
    */
-  private async refreshIndicatorsForRepositories(
-    repositories: ReadonlyArray<Repository>
-  ): Promise<void> {
+  public async refreshAllIndicators() {
     const startTime = performance && performance.now ? performance.now() : null
+
+    // keep a reference to the current set of repositories to avoid the array
+    // changing while this is running
+    const repositories = new Array<Repository>(...this.repositories)
 
     for (const repo of repositories) {
       await this.refreshIndicatorForRepository(repo)
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -144,6 +144,7 @@ import {
   PushOptions,
   RebaseResult,
   getRebaseSnapshot,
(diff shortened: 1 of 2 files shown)


iAmWillShepherd (dismissed):
My comment doesn't need to hold up this PR since what's here works just fine.

app/src/lib/stores/app-store.ts
```diff
@@ -1764,12 +1765,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
   public async _loadStatus(
     repository: Repository,
     clearPartialState: boolean = false
-  ): Promise<boolean> {
+  ): Promise<IStatusResult | null> {
```
outofambit: i'm 👍 with this change, but we should also update this logic to do avoid implicit object boolean-ness: https://github.com/desktop/desktop/blob/e63ab5b33d6012d51a3e98b63563acdbcc59e3cb/app/src/lib/stores/app-store.ts#L2459

app/src/lib/stores/app-store.ts
```diff
@@ -2486,6 +2487,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const status = await this._loadStatus(repository)
     if (!status) {
```
outofambit: can we make this an explicit boolean comparison too?

app/src/lib/stores/app-store.ts
```diff
@@ -2486,6 +2487,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
       await this._updateRepositoryMissing(repository, true)
+      this.updateSidebarIndicator(repository, null)
```
outofambit: i think we can call this right after `_loadStatus` with its result regardless of what it returns and then we can remove the `this.updateSidebarIndicator(repository, status)` at the end of this function entirely.

since `updateSidebarIndicator` is async and fairly straightforward i don't think there's risk there.

app/src/lib/stores/app-store.ts
```diff
@@ -2519,27 +2521,46 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  public async refreshAllSidebarIndicators() {
     const startTime = performance && performance.now ? performance.now() : null
 
+    // keep a reference to the current set of repositories to avoid the array
+    // changing while this is running
+    const repositories = new Array<Repository>(...this.repositories)
+
     for (const repo of repositories) {
-      await this.refreshIndicatorForRepository(repo, tryBackgroundFetch)
+      await this.refreshIndicatorForRepository(repo)
```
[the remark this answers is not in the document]
outofambit: that _might_ be risky if there are a large number of repositories (since we'd flood the process with promises), so i think i have a slight preference for the original implementation.

app/src/lib/stores/app-store.ts
```diff
@@ -2556,15 +2577,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-   *
-   * @param repository the repository to check and update
-   * @param tryBackgroundFetch whether the action should also try and fetch new changes from the remote
+   * Refresh indicator in repository list for a specific repository
    */
-  private async refreshIndicatorForRepository(
-    repository: Repository,
-    tryBackgroundFetch: boolean
-  ) {
+  private async refreshIndicatorForRepository(repository: Repository) {
```
outofambit: i like that this allows us to remove that extra boolean 💯

app/src/lib/stores/app-store.ts
```diff
@@ -2585,16 +2600,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    )
 
-      if (this.shouldBackgroundFetch(repository, lastPush)) {
-        await this._fetch(repository, FetchType.BackgroundTask)
-      }
+    if (this.shouldBackgroundFetch(repository, lastPush)) {
+      await this._fetch(repository, FetchType.BackgroundTask)
     }
 
     lookup.set(repository.id, {
```
outofambit: i wish there was a clean way to use `updateSidebarIndicator` here without doing extra work since the logic is so similar but i can't quite figure out a way to do that right now

outofambit (changes requested):
a couple tweaks, please, before we merge this. overall this is a nice refactor!

app/src/lib/stores/app-store.ts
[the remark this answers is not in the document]
shiftkey: We talked about this in #5312 as there were some performance issue when we initially shipped the repository indicators - we'd create a number promises in the background to do work, and for users with dozens of repositories it would significantly affect things with no indicator of why.

A `for-loop` is fine here.

app/src/lib/stores/app-store.ts
shiftkey: Yeah, I stared at this a bunch while I was in here and couldn't really tease out something I liked either. 

Gonna keep thinking about it, but I won't hold up this PR and the benefits it already gives.

outofambit: > Gonna keep thinking about it, but I won't hold up this PR and the benefits it already gives.

yep, totally agree

outofambit (approved):

PR #7476, opened 2019-05-07
use the index state to detect whether there are conflicted files present

shiftkey:
## Overview

Related to #7464 

## Description

This PR tunes a code path inside `getStatus` that was introduced as part of the new stash flow in #7380. The original implementation was a naive "do we have changes?" check and it turns out that `git diff` is rather expensive when dealing with a large number of working directory changes.

Here's a profile of a refresh in the `nodejs/node` repository when all of the files are marked for deletion (I've highlighted the overall timing):

You'll note that `getFilesWithConflictMarkers` and `getBinaryPaths` take a combined 4.7 seconds, or about 85% of the overall 5.5 seconds it takes to refresh. 

You can also see these fire again outside the scope of the refresh, but I've opened #7475 because I think that work is unnecessary:

The fix I settled on was to look at the index for unmerged files _if we cannot detect a merge or rebase in progress_. If we do not detect any unmerged files in the index, I believe we can skip doing any `git diff` invocations and save time.

## Release notes

Notes: `[Improved] checks for stashing conflicts only when the right state found in index`

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -27,6 +27,7 @@ import { stageFiles } from './update-index'
 import { getStatus } from './status'
 import { getCommitsInRange } from './rev-list'
 import { Branch } from '../../models/branch'
+import { enablePullWithRebase } from '../feature-flag'
 
 /**
  * Check the `.git/REBASE_HEAD` file exists in a repository to confirm
@@ -48,6 +49,10 @@ function isRebaseHeadSet(repository: Repository) {
 export async function getRebaseInternalState(
   repository: Repository
 ): Promise<RebaseInternalState | null> {
+  if (!enablePullWithRebase()) {
+    return null
+  }
+
   const isRebase = await isRebaseHeadSet(repository)
 
   if (!isRebase) {
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -202,18 +202,25 @@ export async function getStatus(
   }
 
   const mergeHeadFound = await isMergeHeadSet(repository)
+  const conflictedFilesInIndex = entries.some(e => e.statusCode === 'UU')
   const rebaseInternalState = await getRebaseInternalState(repository)
 
   if (enablePullWithRebase()) {
     conflictDetails = await getConflictDetails(
       repository,
       mergeHeadFound,
+      conflictedFilesInIndex,
       rebaseInternalState
     )
     // a little optimization so we don't run `getWorkingDirectoryConflictDetails`
     // when there are no files to even detect conflicts in
   } else if (entries.length > 0) {
-    conflictDetails = await getConflictDetails(repository, mergeHeadFound, null)
+    conflictDetails = await getConflictDetails(
+      repository,
+      mergeHeadFound,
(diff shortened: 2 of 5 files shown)


app/src/lib/git/status.ts
```diff
@@ -202,18 +202,25 @@ export async function getStatus(
   }
 
   const mergeHeadFound = await isMergeHeadSet(repository)
+  const conflictedFilesInIndex = entries.some(e => e.statusCode === 'UU')
```
outofambit: should we also look for the other conflicted codes listed in [status-parser](https://github.com/desktop/desktop/blob/189d17f44a4744db26e1ce69b658b9cc98f3aba1/app/src/lib/status-parser.ts#L301-L362)? i'm not totally sure if they can all be generated by a stash pop but i'd rather be safe.

* `DD`
* `AU`
* `UD`
* `UA`
* `DU`
* `AA`
* `UU`

app/src/lib/git/status.ts
```diff
@@ -202,18 +202,25 @@ export async function getStatus(
...
   if (enablePullWithRebase()) {
     conflictDetails = await getConflictDetails(
       repository,
       mergeHeadFound,
+      conflictedFilesInIndex,
       rebaseInternalState
     )
     // a little optimization so we don't run `getWorkingDirectoryConflictDetails`
     // when there are no files to even detect conflicts in
   } else if (entries.length > 0) {
```
outofambit: i think we can remove this tiny optimization now since looking through the entries should cover that case

app/test/unit/git/stash-test.ts
```diff
@@ -263,20 +264,33 @@ describe('git/stash', () => {
...
+      it('restores changes and drops stash', async () => {
+        const status = await getStatusOrThrow(repository)
+        const files = status.workingDirectory.files
         expect(files).toHaveLength(1)
 
         const entriesAfter = await getDesktopStashEntries(repository)
         expect(entriesAfter).not.toContain(entryToApply)
       })
+
+      it('detects conflicted file', async () => {
```
outofambit: i think this test belongs in `status-test.ts`, but i don't feel strongly

app/src/lib/git/status.ts
shiftkey: Good shout! I'll update this tomorrow to try and figure out the most optimal way to do a broad check like this.

app/test/unit/git/stash-test.ts
shiftkey: 👍 let me sleep on this and see how I feel

app/src/lib/git/rebase.ts
```diff
@@ -48,6 +49,10 @@ function isRebaseHeadSet(repository: Repository) {
 export async function getRebaseInternalState(
   repository: Repository
 ): Promise<RebaseInternalState | null> {
+  if (!enablePullWithRebase()) {
```
outofambit: nice

shiftkey:
@outofambit I'm going to revisit the file status checking after the collection of outstanding PRs are reviewed/merged/available on beta, because at this point I want to avoid looking at issues that are resolved elsewhere.

PR #7477, opened 2019-05-07
tweak the output when in development mode for helpful performance context

shiftkey:
## Overview

Some cleanup I would like to land as a result of investigating #7464 

## Description

Two little changes:

 - 655c69cdec5bedb8ed59e32a73ddd7b724c070e9 always shows the Git operations when `__DEV__` is enabled, in production we only log slow Git operations as usual
 - ef487442b9d31ece3beb3f3cc3e686e4ab286c62 this debug message is just noise when `__DEV__` is enabled

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/git-perf.ts
+++ b/app/src/ui/lib/git-perf.ts
@@ -18,7 +18,6 @@ export async function measure<T>(
 ): Promise<T> {
   const id = ++markID
 
-  log.debug(`Executing ${cmd}`)
   const startTime = performance && performance.now ? performance.now() : null
 
   markBegin(id, cmd)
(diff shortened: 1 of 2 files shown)


shiftkey:
I'd love some feedback on whether this PR is helpful when working in development mode

app/src/ui/lib/git-perf.ts
```diff
@@ -18,7 +18,6 @@ export async function measure<T>(
 ): Promise<T> {
   const id = ++markID
 
-  log.debug(`Executing ${cmd}`)
```
outofambit: does this mean this will no longer be in production logs? i've found knowing about all git operations helpful for debugging production stuff in some cases. what's the aim of removing this?

shiftkey: > does this mean this will no longer be in production logs?

`log.debug` messages would only appear in the Dev Tools console for production builds. I wasn't sure if anyone was getting any value out of that there, and it duplicated information as a result of the "always log timings when in development mode" change so I took it out.

Happy to add this back, but rather than adding this back immediately I'll ask: "what information about Git operations would you like to know when using a production build of the app?"

outofambit: ah ok! i think i misremembered where i was looking at logs so i think i'm 👍 with this change. thanks for the extra info.

PR #7488, opened 2019-05-08
Adds some cache to improve baseline metrics performance

iAmWillShepherd:
@shiftkey shared some performance numbers he collected on his machine for IndexedDB queries

> lastStashEntryCheck takes ~150ms (with some outliers in the 400ms+ range) on my Windows test machine. not sure how that makes me feel.

It didn't make me feel great, so we discussed possible solutions which resulted in this PR.

Look at them gains!
Screen Shot 2019-05-08 at 3 50 11 PM

cc @shiftkey  @billygriffin @outofambit

### DIFF
--- a/app/src/lib/stores/repositories-store.ts
+++ b/app/src/lib/stores/repositories-store.ts
@@ -14,6 +14,9 @@ import { BaseStore } from './base-store'
 export class RepositoriesStore extends BaseStore {
   private db: RepositoriesDatabase
 
+  // Key-repo ID, Value-date
+  private lastStashCheckCache = new Map<number, number>()
+
   public constructor(db: RepositoriesDatabase) {
     super()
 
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/repositories-store.ts
```diff
@@ -289,7 +299,12 @@ export class RepositoriesStore extends BaseStore {
       )
     }
 
-    return record!.lastStashCheckDate
+    lastCheckDate = record!.lastStashCheckDate
```
outofambit: feel free to punt on this to the PR for `baseline-stashing-metrics2`, but why do we need to `!` here?

shiftkey: It's the TS way of saying "I know this value might be null/undefined and I don't care, let's use it anyway".

We try to avoid these unless absolutely necessary, and I think we should be able to drop this because of the previous `if (record === undefined)`

iAmWillShepherd: Nope!

shiftkey:
@iAmWillShepherd merge whenever you're ready

PR #7504, opened 2019-05-10
add timing API to understand how long important actions take

shiftkey:
## Overview

Related to #7464 and #7477, this PR adds some infrastructure to help us understand how long things take for the user, to give us an idea of where to spend our time with performance improvements.

## Description

Here's an actual piece of feedback from one of our beta testers:

> *2. create a branch from your current branch and bring changes with you*
> 
>  - prod: 3 seconds
>  - beta: 8 seconds

I want to avoid making users get this stuff manually, for a few reasons:

 - making them re-run scenarios _after_ noticing a problem isn't fun
 - manually gather timings is a complicated task with a production app
 - unless the user remembers to provide more information about their setup (which OS, and even which repository can make a huge difference), more back and forth here with the maintainers is needed to get a bigger picture

We have a place for things like this - the logs. So why not put them there, and the user can just upload the log and we gain these extra insights for free?

After my learnings from #7464 I settled on a simple API and some recommendations about _when_ we should use it.

```ts
const timer = startTimer('stash and checkout', repository)
await this.stashAndCheckout()
timer.done()
```

The inputs to `startTimer` are:

 - a human-friendly name representing the action being performed - this appears in the logs later
 - the repository currently in use - because of how repositories can vary in shape and size, I think this is important to know up front to give a sense of the type of performance issue to investigate

So, when and where do I recommend using this?

 - Apply to operations that are performed by the user _and then block, waiting on them to complete_. These areas are functions where users will notice regressions, and it'd be great to know the impact when things have changed enough to receive feedback. I included some examples in this PR:
   - `checkout branch from list`
   - `stash and checkout`
   - `create branch`
   - `create commit`
 - Apply this to important operations and workflows - we don't want to overload the logs with noise, but we want to know early if a beta release significant affects workflows existing workflows (and by how much)
  - Measure as close to the user interaction as possible - ideally the UI components should invoke an action and wait on the result. If you're measuring something deep inside a data store, I recommend a different approach.
  - Measure actions that are stable - we want to track performance over time, so choose actions where the underlying implementation is unlikely to change radically between updates.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -21,6 +21,7 @@ import { PullRequestList } from './pull-request-list'
 import { IBranchListItem } from './group-branches'
 import { renderDefaultBranch } from './branch-renderer'
 import { IMatches } from '../../lib/fuzzy-find'
+import { startTimer } from '../lib/timing'
 
 interface IBranchesContainerProps {
   readonly dispatcher: Dispatcher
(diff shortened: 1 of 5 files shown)


alaoy24:
H

shiftkey:
Moving this into the review queue to see if there's any interest in pulling this in for a future beta

Issue #6965, opened 2019-02-27
test CI providers on Node 10


shiftkey:
**Please describe the problem you think should be solved**

Our CI providers spent time installing a specific version of Node 8 because that's the version we settled on ages ago. I think we can leverage a later version of Node, and potentially see some improvements in the time to build and test the app.

**[Optional] Do you have any potential solutions in mind?**

These are the relevant build configs that need updating:

 - [`.circleci/config.yml`](https://github.com/desktop/desktop/blob/development/.circleci/config.yml)
 - [`.travis.yml`](https://github.com/desktop/desktop/blob/development/.travis.yml)
 - [`appveyor.yml`](https://github.com/desktop/desktop/blob/development/appveyor.yml)
 - [`azure-pipelines.yml`](https://github.com/desktop/desktop/blob/development/azure-pipelines.yml)

shiftkey:
Dropping `help wanted` here because I don't think PRs will be able to change build configs (as a security measure) so this isn't ideal for external contributors to pick up.

shiftkey:
As part of reviewing #7381 I noticed that Travis is inheriting it's configured version instead of what's in `.nvmrc`. Whoever gets to this work should confirm we can remove this config value so we have one fewer config value to update:

https://github.com/desktop/desktop/blob/4484f85735afd4afa2a396548cafa522663dc308/.travis.yml#L29-L30

shiftkey:
Tweaking the issue title and body to focus on Node 10 (which is still LTS and will be the [active LTS until April 2020](https://nodejs.org/en/about/releases/))

Rexogamer:
The CircleCI config doesn’t seem to mention Node, and all the others have been updated , so should this be closed?

shiftkey:
@Rexogamer we addressed the CircleCI issue in #7836 as well as #7507 and #7900 so I think this is good to close out.

Issue #6888, opened 2019-02-20
upgrade app to use Electron 5


shiftkey:
This should be a smoother upgrade than the Electron 3 process because they use the same version of Node, but #6883 is a known blocker that needs to happen too.

The other issue I'm tracking is https://github.com/electron/spectron/issues/346 which is about using Spectron with Electron 4 - there's a workaround in the issue to get it working with pre-release versions, which is unsatisfying. I'm trying to chase what the status is of this issue.

Given how subtle #6883 was, I'd recommend whoever looks at this issue should try upgrading locally to see if there's any other breakage.

shiftkey:
I think `electron-packager` needs to be updated to address this issue when I try and `yarn build:dev` with Electron 4.0.5:

```
Packaging app for platform darwin x64 using electron v4.0.5
{ Error: ENOENT: no such file or directory, open '/Users/shiftkey/src/desktop/dist/GitHub Desktop-dev-darwin-x64/Electron.app/Contents/Frameworks/Electron Helper EH.app/Contents/Info.plist'
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)
From previous event:
    at packagerPromise (/Users/shiftkey/src/desktop/node_modules/electron-packager/index.js:163:6)
    at packager (/Users/shiftkey/src/desktop/node_modules/electron-packager/index.js:184:18)
    at packageApp (/Users/shiftkey/src/desktop/script/build.ts:198:10)
    at validate_all_1.verifyInjectedSassVariables.catch.then.then (/Users/shiftkey/src/desktop/script/build.ts:100:12)
    at <anonymous>
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/shiftkey/src/desktop/dist/GitHub Desktop-dev-darwin-x64/Electron.app/Contents/Frameworks/Electron Helper EH.app/Contents/Info.plist' }
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'd recommend doing this in advance, because it seems there's been some recent improvements to `electron-packager` for Mojave support, based on the [latest releases](https://github.com/electron-userland/electron-packager/releases).

shiftkey:
[...]

PR #7507, opened 2019-05-10
upgrade infrastructure to node 10

shiftkey:
## Overview

This PR moves us up to a new baseline of Node 10, which is the same major version as in Electron 3.x. There's some things related to #6965 that I'm investigating about this, in no particular order.

## Description

Assorted notes below:

 - I've upgraded the `@types/node` in the root package.json to a `10.x` version. Previously I was worried about how this would blend with Electron 3.x, but given we have Electron 5 on our radar I'm going to keep it simple for now.

 - Updating the CI providers was mostly painless, but I couldn't find a better solution for updating CircleCI than installing and switching over to use `nvm` with that script.

There's probably more to come, but once we're happy with these changes and landing them I'll allow a couple of weeks for contributors to update their machines.

## Release notes

Notes: upgrade infrastructure to use Node 10 as a minimum


.circleci/config.yml
```diff
@@ -5,6 +5,19 @@ defaults: &defaults
   macos:
     xcode: '9.3.1'
```
outofambit: could we upgrade this to 10.1 or higher to get the right version of node and avoid having to use nvm? https://discuss.circleci.com/t/xcode-10-1-image-released/26350

shiftkey: I've been hesitant to upgrade the version of XCode on these images without first investigating whether things will be fine for native modules (Electron 3.x runs on [macOS 10.9 or later](https://github.com/electron/electron/blob/024cb4afd9cebe9a1bf1976ba83c3cdce1a5afd7/docs/tutorial/support.md#macos), so that's our compatibility baseline we need to be mindful of).

Also, based on that link it looks like this takes us up to Node 11 (which is weird, given Node 12 is latest and Node 10 is latest LTS).

The recommendations in other support threads about "how do I enable a specific version of Node?" when I was looking into this all seem to coalesce around using NVM, so I'm not sure how unavoidable it is (short-term or longer-term) if we want to be able to control the version of Node used here without also upgrading XCode, like we can for the other providers.

Thoughts?

package.json
```diff
@@ -141,7 +141,7 @@
     "@types/memoize-one": "^3.1.1",
     "@types/mini-css-extract-plugin": "^0.2.0",
     "@types/mri": "^1.1.0",
-    "@types/node": "^8.10.4",
+    "@types/node": "10.1.4",
```
shiftkey: For the moment I'm going to leave this on a `10.1.x` version, despite there being a Node `10.14.9` available as the latest 10.x version. When upgrading to that version I see a number of errors building the app:

```
[WEBPACK] Errors building ask-pass.js
[at-loader] ./node_modules/@types/node/ts3.2/util.d.ts:6:15
    TS2451: Cannot redeclare block-scoped variable 'custom'.
[at-loader] ./node_modules/@types/node/ts3.2/util.d.ts:9:15
    TS2451: Cannot redeclare block-scoped variable 'custom'.
[at-loader] ./app/src/lib/shell.ts:65:64
    TS2345: Argument of type '{ detached: boolean; stdio: (string | WriteStream)[]; }' is not assignable to parameter of type 'SpawnOptions'.
  Types of property 'stdio' are incompatible.
    Type '(string | WriteStream)[]' is not assignable to type '"pipe" | "ignore" | "inherit" | (number | "pipe" | "ipc" | "ignore" | "inherit" | Stream | null | undefined)[] | undefined'.
      Type '(string | WriteStream)[]' is not assignable to type '(number | "pipe" | "ipc" | "ignore" | "inherit" | Stream | null | undefined)[]'.
        Type 'string | WriteStream' is not assignable to type 'number | "pipe" | "ipc" | "ignore" | "inherit" | Stream | null | undefined'.
          Type 'string' is not assignable to type 'number | "pipe" | "ipc" | "ignore" | "inherit" | Stream | null | undefined'.
[at-loader] ./app/src/lib/shell.ts:69:5
    TS2531: Object is possibly 'null'.
[at-loader] ./app/src/lib/shell.ts:74:5
    TS2531: Object is possibly 'null'.
[at-loader] ./app/src/lib/shell.ts:78:5
    TS2531: Object is possibly 'null'.
[at-loader] ./app/src/ui/lib/list/list.tsx:291:30
    TS2345: Argument of type 'number' is not assignable to parameter of type 'Immediate'.
[at-loader] ./app/src/ui/lib/list/list.tsx:294:13
    TS2322: Type 'Immediate' is not assignable to type 'number'.
[at-loader] ./app/src/ui/lib/list/list.tsx:624:22
    TS2345: Argument of type 'number' is not assignable to parameter of type 'Immediate'.
[at-loader] ./app/src/ui/diff/image-diffs/modified-image-diff.tsx:74:28
    TS2345: Argument of type 'number' is not assignable to parameter of type 'Immediate'.
[at-loader] ./app/src/ui/diff/image-diffs/modified-image-diff.tsx:77:11
    TS2322: Type 'Immediate' is not assignable to type 'number'.
[at-loader] ./app/src/ui/history/commit-summary.tsx:145:30
    TS2345: Argument of type 'number' is not assignable to parameter of type 'Immediate'.
[at-loader] ./app/src/ui/history/commit-summary.tsx:148:13
    TS2322: Type 'Immediate' is not assignable to type 'number'.
[at-loader] ./node_modules/@types/node/index.d.ts:82:15
    TS2451: Cannot redeclare block-scoped variable 'custom'.
[at-loader] ./node_modules/@types/node/index.d.ts:85:15
    TS2451: Cannot redeclare block-scoped variable 'custom'.
```

These seem to relate some upstream PRs where things have changed in 10.x but the `@types/node` package took time to catch up:

 - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29599
 - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34285

Given we're not seeing any runtime issues with the app currently, I've avoided updating the contents of `app/` in this PR and have pinned the version for now to focus on the CI changes.

I'd be happy to revisit this in a later situation, such as:

 - we update the app to Electron 5 #6888, which should be tested against Node 12 (and has the correct internal `@types/node` as a dependency)
 - we update `@types/node` manually to a later `10.x` and verify the code that needs to be updated

.circleci/config.yml
outofambit: maybe #7836 would work?

shiftkey:
@desktop/engineering I want to use this PR and verify we can create a test release before this gets merged. I'll report back when I can confirm this won't interfere with our upcoming beta release.

shiftkey:
This is good to be reviewed and merged again. There's some Cloudflare issues which mean nodejs.org is inaccessible, so some CI providers are unable to install the desired version of Node and failing the build, but that should be resolved today...

shiftkey:
@outofambit the 2.1.0 beta has been published, so feel free to merge this in whenever you're ready

Issue #7091, opened 2019-03-14
[Metrics] Commits to protected branches


billygriffin:
To understand the potential impact of #7023 (and to decide how to prioritize it), we'd like to understand the following information about commits to protected branches.

**For users who have committed to a repo with at least one protected branch**:

1. How many users fit the above category (people who have committed to a repo with at least one protected branch)
2. How many commits per user to a protected branch
3. How frequently it's occurring for people
4. % of people accidentally committing to a protected branch
5. On repos with protected branches, % of commits to those branches

(thanks to @nerdneha for making this way better than it started, and cc: @scottypate for visibility since we've discussed this)

shiftkey:
Closing this out as changes in the 1.7 release will provide insights into these questions, and we have the original issue  #7023 open to discuss things further

PR #7529, opened 2019-05-14
retrieve and cache protected branch information per-repository

shiftkey:
## Overview

**First part of #7091**

## Description

This PR lands the work required to retrieve the protected branch and store it in IndexedDB. I'd love some feedback about the table changes made (cc @niik @iAmWillShepherd) to ensure they make sense.

I also wired up a simple check to see if the tracking branch being committed to (they don't need to match up, so I'm deferring to Git rather than trusting the user is following conventions) is found in the list of cached protected branches.

If we're fine with the changes and landing them for 1.7, I'll follow up with subsequent PRs to refine the checks and record metrics to complete #7091.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -226,6 +226,11 @@ export interface IAPIRefStatus {
   readonly statuses: ReadonlyArray<IAPIRefStatusItem>
 }
 
+export interface IAPIBranch {
+  readonly name: string
+  readonly protected: boolean
+}
+
 interface IAPIPullRequestRef {
   readonly ref: string
   readonly sha: string
@@ -694,6 +699,15 @@ export class API {
     const path = `repos/${owner}/${name}/commits/${ref}/status`
     const response = await this.request('GET', path)
     return await parsedResponse<IAPIRefStatus>(response)
+  }
+
+  public async fetchProtectedBranches(
+    owner: string,
+    name: string
+  ): Promise<ReadonlyArray<IAPIBranch>> {
+    const path = `repos/${owner}/${name}/branches?protected=true`
+    const response = await this.request('GET', path)
+    return await parsedResponse<IAPIBranch[]>(response)
   }
 
   /**
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -226,8 +226,20 @@ export interface IAPIRefStatus {
   readonly statuses: ReadonlyArray<IAPIRefStatusItem>
 }
 
+/** Branch information returned by the GitHub API */
 export interface IAPIBranch {
+  /**
+   * The name of the branch stored on the remote.
+   *
+   * NOTE: this is NOT a fully-qualified ref (i.e. `refs/heads/master`)
+   */
   readonly name: string
+  /**
+   * Branch protection settings:
+   *
+   *  - `true` indicates that the branch is protected in some way
+   *  - `false` indicates no branch protection set
+   */
   readonly protected: boolean
 }
 
--- a/app/src/lib/databases/repositories-database.ts
+++ b/app/src/lib/databases/repositories-database.ts
@@ -23,6 +23,12 @@ export interface IDatabaseGitHubRepository {
   readonly lastPruneDate: number | null
 }
 
+export interface IDatabaseGitHubBranch {
+  readonly repoId: number
+  readonly name: string
+  readonly protected: boolean
+}
+
 export interface IDatabaseRepository {
   readonly id?: number | null
   readonly gitHubRepositoryID: number | null
@@ -40,6 +46,9 @@ export class RepositoriesDatabase extends BaseDatabase {
 
   /** The GitHub repositories table. */
   public gitHubRepositories!: Dexie.Table<IDatabaseGitHubRepository, number>
+
+  /** The branch lookup table for GitHub repositories table. */
+  public githubBranches!: Dexie.Table<IDatabaseGitHubBranch, number>
 
   /** The GitHub repository owners table. */
   public owners!: Dexie.Table<IDatabaseOwner, number>
@@ -76,6 +85,10 @@ export class RepositoriesDatabase extends BaseDatabase {
 
     this.conditionalVersion(5, {
       gitHubRepositories: '++id, name, &[ownerID+name], cloneURL',
+    })
+
+    this.conditionalVersion(6, {
+      githubBranches: '[repoId+name], repoId',
     })
   }
 }
--- a/app/src/lib/databases/repositories-database.ts
+++ b/app/src/lib/databases/repositories-database.ts
@@ -87,7 +87,7 @@ export class RepositoriesDatabase extends BaseDatabase {
     })
 
     this.conditionalVersion(6, {
-      protectedBranches: '[repoId+name]',
+      protectedBranches: '[repoId+name], repoId',
     })
   }
 }
--- a/app/src/lib/databases/repositories-database.ts
+++ b/app/src/lib/databases/repositories-database.ts
@@ -23,10 +23,9 @@ export interface IDatabaseGitHubRepository {
   readonly lastPruneDate: number | null
 }
 
-export interface IDatabaseGitHubBranch {
+export interface IDatabaseProtectedBranch {
   readonly repoId: number
   readonly name: string
-  readonly protected: boolean
 }
 
 export interface IDatabaseRepository {
@@ -48,7 +47,7 @@ export class RepositoriesDatabase extends BaseDatabase {
   public gitHubRepositories!: Dexie.Table<IDatabaseGitHubRepository, number>
 
   /** The branch lookup table for GitHub repositories table. */
-  public githubBranches!: Dexie.Table<IDatabaseGitHubBranch, number>
+  public protectedBranches!: Dexie.Table<IDatabaseProtectedBranch, number>
 
   /** The GitHub repository owners table. */
   public owners!: Dexie.Table<IDatabaseOwner, number>
@@ -88,7 +87,7 @@ export class RepositoriesDatabase extends BaseDatabase {
     })
 
     this.conditionalVersion(6, {
-      githubBranches: '[repoId+name], repoId',
+      protectedBranches: '[repoId+name]',
     })
   }
 }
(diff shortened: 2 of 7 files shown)


app/src/lib/databases/repositories-database.ts
```diff
@@ -23,6 +23,12 @@ export interface IDatabaseGitHubRepository {
   readonly lastPruneDate: number | null
 }
 
+export interface IDatabaseGitHubBranch {
+  readonly repoId: number
+  readonly name: string
+  readonly protected: boolean
```
shiftkey: For the moment we're just looking at the `protected` column because that was easy to compute, but there might come a day down the track where we store more information in this table. In this case I'd recommend just dropping and recreating the table with a newer schema.

tierninho:
@shiftkey Please confirm the location of the protected branch data in IndexedDB as I couldn't locate it. Do I have to delete and rebuild the database first?

shiftkey:
@tierninho you'll need to drop the database after switching away from this PR (because it increments the version), but check the `IndexedDB` -> `Database` -> `githubBranches` table for entries:

shiftkey:
> I assume we do not record `protected: false` for all other branches as that would be cumbersome?

I believe it's a waste to download and store these branches, and the API request excludes these:

https://github.com/desktop/desktop/blob/cf10301b2ab5a04c11a8563a964e22455d167f2f/app/src/lib/api.ts#L530

shiftkey:
This is ready for another round of review

app/src/lib/stores/repositories-store.ts
```diff
@@ -485,4 +510,38 @@ export class RepositoriesStore extends BaseStore {
...
+
+    const existing = this.branchProtectionCache.get(key)
+    if (existing !== undefined) {
+      return existing
+    }
+
+    const result = await this.db.protectedBranches
+      .where('[repoId+name]')
+      .equals([gitHubRepository.dbID, branchName])
+      .first()
```
niik: This could be simplified to a primary key lookup

```ts
const existing = this.db.protectedBranches.get([gitHubRepository.dbID, branchName])
const protected = existing !== undefined
```

app/src/lib/stores/app-store.ts
```diff
@@ -2373,6 +2373,27 @@ export class AppStore extends TypedBaseStore<IAppState> {
             }
           }
         }
+
+        if (
+          gitStore.tip.kind === TipState.Valid &&
+          gitStore.currentRemote !== null
+        ) {
```
niik: Given that this section doesn't do anything but write to the log at this time could we add a feature flag to ensure it's isolated from production while we work on this?

app/src/lib/stores/app-store.ts
```diff
@@ -3019,11 +3039,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return repository
     }
 
+    const branches = await api.fetchProtectedBranches(owner, name)
```
niik: Could we feature flag this as well and use an empty array when the feature flag isn't turned on?

app/src/lib/stores/repositories-store.ts
```diff
@@ -399,6 +404,26 @@ export class RepositoriesStore extends BaseStore {
           gitHubRepositoryID: updatedGitHubRepo.dbID,
         })
 
+        const repoId = updatedGitHubRepo.dbID!
+
+        await this.db.protectedBranches
```
niik: Don't we have to clear the cache at this point? Say that the last time we fetched `branch-a` was protected but now it no longer is we wouldn't update its entry in the cache since it won't be returned in the api response.

app/src/lib/databases/repositories-database.ts
```diff
@@ -23,6 +23,11 @@ export interface IDatabaseGitHubRepository {
   readonly lastPruneDate: number | null
 }
 
+export interface IDatabaseProtectedBranch {
```
niik: Can we add a tiny bit of documentation to explain what this does.

app/src/lib/databases/repositories-database.ts
```diff
@@ -23,6 +23,11 @@ export interface IDatabaseGitHubRepository {
...
+  readonly repoId: number
+  readonly name: string
```
niik: I'm fine with the name of this property but it could use a sprinkle of documentation to make it easy to understand that it's a branch name it's referring to (and also whether than branch name is fully qualified or not).

app/src/lib/api.ts
```diff
@@ -226,6 +226,11 @@ export interface IAPIRefStatus {
   readonly statuses: ReadonlyArray<IAPIRefStatusItem>
 }
 
+export interface IAPIBranch {
+  readonly name: string
```
niik: Could we document this please? In particular I'd like to see a description of whether this is fully qualified or not.

app/src/lib/databases/repositories-database.ts
```diff
@@ -41,6 +46,9 @@ export class RepositoriesDatabase extends BaseDatabase {
   /** The GitHub repositories table. */
   public gitHubRepositories!: Dexie.Table<IDatabaseGitHubRepository, number>
 
+  /** The branch lookup table for GitHub repositories table. */
```
niik:
```suggestion
  /** A table containing the names of protected branches per repository. */
```
❓

app/src/lib/stores/repositories-store.ts
```diff
@@ -399,6 +404,26 @@ export class RepositoriesStore extends BaseStore {
           gitHubRepositoryID: updatedGitHubRepo.dbID,
         })
 
+        const repoId = updatedGitHubRepo.dbID!
```
niik: Could we put this entire section behind a feature flag?

app/src/lib/stores/repositories-store.ts
shiftkey: That's a case I didn't think about initially - good call. The problem I have currently is the cache key is built off the repository id (known) and the branch name (unknown). I guess I could enumerate the keys and filter on everything starting with a certain pattern - will investigate.

app/src/lib/api.ts
shiftkey: Added in 3581f36e87387ae4fa8beba20ab9a0b0d90c630c

app/src/lib/databases/repositories-database.ts
shiftkey: Added in 3581f36e87387ae4fa8beba20ab9a0b0d90c630c

app/src/lib/databases/repositories-database.ts
shiftkey: Added in 3581f36e87387ae4fa8beba20ab9a0b0d90c630c

app/src/lib/stores/repositories-store.ts
shiftkey: This doesn't seem to work OOTB because it doesn't seem to like my usage of `dbID` (a number):

```
Argument of type '(string | number)[]' is not assignable to parameter of type '{ [key: string]: IndexableType; }'.
  Index signature is missing in type '(string | number)[]'.
```

I'll see whether that this is some typing issue with Dexie that I need to workaround ala `PullRequestKey`, but if you've got any extra insight here it'd be greatly appreciated:

https://github.com/desktop/desktop/blob/04cb3aaefed772f3d1657f20c87886727fdac998/app/src/lib/databases/pull-request-database.ts#L63-L70

app/src/lib/stores/app-store.ts
shiftkey: Addressed in dc8a86a12e42e4e90ded7054b4ef15d6ef372c01

app/src/lib/stores/repositories-store.ts
shiftkey: Addressed in e36df6dd69209202e433f28fb5aa9224cb0c03fb

app/src/lib/stores/repositories-store.ts
shiftkey: Addressed in 424bc9178618e5767956e29dbdf349ededb60743

app/src/lib/stores/repositories-store.ts
```diff
@@ -485,4 +530,43 @@ export class RepositoriesStore extends BaseStore {
...
+    }
+
+    const key = getKey(gitHubRepository.dbID, branchName)
+
+    const existing = this.branchProtectionCache.get(key)
+    if (existing !== undefined) {
+      return existing
+    }
+
+    const result = await this.db.protectedBranches.get({
```
niik: So the problem with this approach of using an object instead of an array is that it'll be the [exact same thing as doing a `.where().equals().first()`](https://github.com/dfahlander/Dexie.js/blob/v2.0.0/src/Dexie.js#L958-L959). I.e. it will not be a primary key lookup. I bet this is just another case of the Dexie typings not being perfect. Did you try casting the array to `Array<any>`?

app/src/lib/stores/repositories-store.ts
```diff
@@ -399,6 +405,45 @@ export class RepositoriesStore extends BaseStore {
...
+          // - update the underyling database state
+          //
+          // This should ensure any stale values are not being used, and avoids
+          // the need to query the database while the results are in memory.
+
+          const prefix = getKeyPrefix(repoId)
+
+          for (const key of this.branchProtectionCache.keys()) {
+            // invalidate any cached entries belonging to this repository
+            if (key.startsWith(prefix)) {
```
niik: Ordinarily I'd be hesitant with a linear search here but given that I expect that the number of protected branches will be tiny (i.e. very rarely more than 1 per repository and most of the time none at all) I think this is fine. Even if we 10x our expectations and 10x the average number of repositories this number should still be manageable.

app/src/lib/stores/repositories-store.ts
shiftkey: The problem here was the original database declaration wasn't updated from `Key` being `number` to being `[number, string]` to match the new constraints. Fixed in c64f057551ba0eb9327f3c2b5bd26f280abb251b.

shiftkey:
@tierninho would you be able to give this another pass today? I'd like to merge this Wednesday morning EST so I can focus on getting #7595 ready for the next beta.

tierninho:
@shiftkey Gave it another pass and the new `protectedBranches` table is reflected with all the protected branches matching up between remote and local. 👍

PR #7532, opened 2019-05-14
Ensure `updateStashEntryCountMetric` is updated during a long session of Desktop

iAmWillShepherd:
Addresses feedback left in #7470 ([comment](https://github.com/desktop/desktop/pull/7470/files#r283814821)) by moving where we do the stash size check.

cc @niik @billygriffin

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1141,7 +1141,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return Promise.resolve(null)
     }
 
-    this.updateStashEntryCountMetric(repository)
     setNumber(LastSelectedRepositoryIDKey, repository.id)
 
     // if repository might be marked missing, try checking if it has been restored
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -3580,20 +3583,33 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
   }
   private async updateStashEntryCountMetric(repository: Repository) {
-    const lastStashEntryCheck = await this.repositoriesStore.getLastStashCheckDate(
-      repository
-    )
+    let lastStashEntryCheck =
+      this.lastStashCheckCache.get(repository.id) || null
+    if (lastStashEntryCheck === null) {
+      lastStashEntryCheck = await this.repositoriesStore.getLastStashCheckDate(
```
niik: So this seems redundant to me given that we already have a caching layer for this exact thing in `repositoriesStore.getLastStashCheckDate`:

https://github.com/desktop/desktop/blob/e8eaaeab08eed4381e389ae11da04612f56fd20a/app/src/lib/stores/repositories-store.ts#L17-L18

and

https://github.com/desktop/desktop/blob/e8eaaeab08eed4381e389ae11da04612f56fd20a/app/src/lib/stores/repositories-store.ts#L289-L292

I believe the only thing this PR really needs to do is move the callsite for `this.updateStashEntryCountMetric` from `_selectRepository` to `_refreshOrRecoverRepository` (like what we've already done). The caching layer in `repositoriesStore` should take care of any performance concerns we might have by ensuring that we only hit the database once every 24 hours

outofambit: oh, i missed that `repositories-store.ts` already had caching for this same thing! +1 on @niik's feedback here.

PR #7541, opened 2019-05-16
Address Case of the Missing Monitor

### LINKED ISSUE #2107: Window hiding after using a second screen

### LINKED ISSUE #7418: update electron-window-state to get improved checks for missing monitor
I haven't kept up with `electron-window-state` for a while, but it looks like they've landed a fix for the problem we've seen about the app not being restored correctly when the secondary monitor is disconnected #2107 

https://github.com/mawie81/electron-window-state/pull/45

This is available in `5.0.3` on NPM, and I'd love some help from a person with multiple monitor to upgrade our version in `app/package.json` and test this correctly works when performing these steps:

 - launch Desktop, move it to the secondary monitor
 - close Desktop
 - disconnect the secondary monitor
 - launch Desktop

**Expected behaviour:** application is launched on primary monitor at default size

I expect this is relevant to both macOS and Windows users

say25:
## Overview

**Closes #7418**
**Closes #2107**

## Description

- Updates `electron-window-state` to `5.0.3`
- In my personal testing this package update alone did not resolve this issue on Windows yet it did not make the problem any worse. Maybe I didn't test the reproduction steps properly? Maybe the fact that both my monitors are 4k contribute to my ability to reproduce?
- This eliminates the need for `jsonfile@^2.2.3`

## Release notes

Notes: [Fixed] Application window positioned off-screen when using multiple monitors on Windows.


shiftkey:
> I've been able to reproduce the scenario outlined in #7418 on Windows (macOS seems to work just fine with the previous version as far as I can tell). 

This is great news, and I'm tentatively going to add #2107 to the list of `Fixes` issues on this PR so we can close that out too!

say25:
Shall remove the `any`s and retest tonight

say25:
@niik & @shiftkey changes made, I will try and test again on my Windows 10 machine soon

niik:
@say25 I'm gonna go ahead and bring this in since I've been able to reproduce the original issue and confirm that the update resolves it for me. I'd still love it if you could test it on your machine so that we can feed that data into either a stronger confirmation that the issue is resolved or a new issue with specific troubleshooting data.

say25:
Seems like I can reproduce now (I recently had an issue with my machine and had to get a new motherboard maybe that was it?)

PR #7555, opened 2019-05-16
fixing tests for 1.6.6 branch

outofambit:
resolved some shared state issues in the jest tests by moving setup from a `beforeEach` to each individual test. its not pretty, but it works and i don't think its worth spending more time on this.


outofambit:
merging to test that this builds successfully in azure pipelines on the release branch

PR #7557, opened 2019-05-17
Only pass what's actually needed for a commit message

niik:
## Overview

Spotted this tiny mistake when looking for problems related to reported performance problems in 1.6.6. Note that this has no bearing on the performance investigation, just something I happened on while investigating.

## Description

Way back in the day the state of the `CommitMessage` component consisted only of the summary and description properties, i.e. the shape of the state matches the `ICommitMessage` interface perfectly. Since then more fields have been added to the state shape but we were still passing the entire state to the dispatcher.

## Release notes

Notes: no-notes


tierninho:
@niik Can you confirm this is what we are supposed to be seeing now? I couldn't compare it to the prod version as my devtools was somehow missing the React feature.

Screen Shot 2019-05-17 at 12 06 25 PM

54f8c2347d, mac

shiftkey:
> I couldn't compare it to the prod version as my devtools was somehow missing the React feature.

@tierninho the React Devtools browser extension is only installed when working in development mode. That's why it's not available in the installed production or beta builds.

niik:
> @niik Can you confirm this is what we are supposed to be seeing now?

It is indeed!

PR #7558, opened 2019-05-17
more test fixing :/

outofambit:


app/src/lib/stores/app-store.ts
```diff
@@ -2530,11 +2530,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
       gitStore.updateLastFetched(),
       gitStore.loadStashEntries(),
       this.refreshAuthor(repository),
+      this.updateStashEntryCountMetric(repository),
```
shiftkey: So I think this is the "fix" for the CI issue - previously the function wasn't being awaited on anywhere, so moving this into here ensures awaiting `AppStore._refreshRepository`  will await on this new function.

shiftkey: Another thing to do could be to run `this.repositoriesStore.updateLastStashCheckDate(repository)` in the test setup so that the new code path would be skipped in tests. I'm going to try that out locally and see how it feels.

shiftkey:
Closing this out in favour of #7561

PR #7537, opened 2019-05-15
Release 1.6.6

outofambit:
~~🌵 **Blocked by intermittent test failures. See #7562** 🌵~~

This PR shows the differences between this branch and our development branch. Since this release is based off of the last production release please see the [diff between previous prod release and this branch](https://github.com/desktop/desktop/compare/release-1.6.5...releases/1.6.6) for an overview of what changes this branch brings over the last release.

**Update: Diff between 1.6.5 and 1.6.6:** https://github.com/desktop/desktop/compare/release-1.6.5...release-1.6.6

- [x] ✍️ finalize release notes
- [x] 🚀 ship to production
- [x] 🆙 update this branch from 'development'
- [x] 💯 merge!

PR #7562, opened 2019-05-17
disable flaky tests in 1.6.6 release branch

outofambit:
## Context

We've been dealing with intermittent test failures in #7537. A good amount of time has been spent trying to isolate and fix this, but we're still not out of the woods.

## The Problem

The `_finishConflictedMerge` tests trigger `appStore._refreshRepository`, which makes an asynchronous, but un-`await`-ed call to the stats store and database. Sometimes this results in an attempt to access the database after the test has ended, which throws an error and (rightfully) fails the test.

## Summary

To be clear, this isn't a bug in appStore's behavior, but it is evidence of a difficult-to-test architecture that requires more scaffolding/mocking/rigging in our tests for them to be reliable. This PR disables the tests in question so they can be revisited at some point in the future. (tracked over in #7563)

_____

### Even More Context

(#7558 looked into forcing these calls to complete before the test finished, but affects the behavior of the appStore in those situations, which could affect the performance of the app. and #7561 attempted to fix this by manually calling the stats store before the test to prevent it from getting triggered in the test cases.)

### DIFF
--- a/app/test/unit/app-store-test.ts
+++ b/app/test/unit/app-store-test.ts
@@ -179,7 +179,12 @@ describe('AppStore', () => {
       expect(state.localCommitSHAs).toHaveLength(0)
     })
   })
-  describe('_finishConflictedMerge', () => {
+
+  // skipping these tests because its not worth the time it would take
+  // to make them reliable. the underlying problem is that this scenario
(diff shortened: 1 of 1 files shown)


shiftkey:
I'm going to merge this in and ignore that Codecov warning because that's not important right now...

PR #7568, opened 2019-05-19
organize and update community section of README

### LINKED ISSUE #7564: Linux Shiftkey Linked
Hi,

I think the Shiftkey/Desktop linux version should be linked as it is unclear that is an option from this main page. Keep up the good work!

Maxwell

shiftkey:
## Overview

**Closes #7564**

## Description

Adding a section below the beta channel to make it clear where you can find community releases of the app.

## Release notes

Notes: no-notes


Avz7uy123:
Sorry I’m just trying to understand this message system. But to me my point of view is the county is our community. I will however stay away from other County’s. And to start I would not .com pursuit what I haven’t finished. If its within my area of infiltration, then it became part of our community. But I will take a step backwards and let the right department handle it. As for me if I don’t see result of ignoring my opinions, I will voluntarily walk away from the purpose of the main Threat. And my own world and back to a world of worries of my kids future.

Get Outlook for iOS<https://aka.ms/o0ukef>

Daniel-McCarthy:
@outofambit I most definitely agree that it should be made clear that it is not officially supported and that issues with the unofficial Linux support should be kept to the fork's GitHub repository.

However, I do think there is value in having some form of navigation to this. Discovery of the linux fork is not exactly the convenient and I believe can easily be missed entirely. When I switched from Windows to Linux, I looked at GitHub Desktop to see if I could use it on my new OS and did not see anything related on the website. Eventually though continuing to search for any information on it I discovered a very large Issue about lack of linux support. After reading through a large number of comments I just _very_ fortunately lucked out and saw a comment that linked to the fork where I was able to install it.

To me it is unfortunate that the ability to use GitHub Desktop on linux, even unofficially, could potentially be missed. To that end I feel there's great value in this, while also agreeing that it should be amended to with heavy clarification that it is not official.

shiftkey:
> I don't feel comfortable having the linux desktop fork and packages outside of the community section in our readme and under its own subheader since it is not supported by our team.

I made this move because this list combined supported and unsupported platforms, and I wanted to move the Linux platform out to its own section to be more explicit about where these are coming from.

outofambit:
thanks for your thoughts @Daniel-McCarthy! that all makes sense. @shiftkey what are your thoughts on adding some language to clarify linux isn't officially supported?

niik:
> adding some language to clarify linux isn't officially supported?

I agree wholeheartedly with @outofambit's suggestion here, in fact I'd say is an absolute must and I'd add that I believe the language should be very clear before we put it in our README.

shiftkey:
@niik @outofambit I've moved everything community-related down to below the beta channel.

Let me know if that feels like a better way to organize things.

Issue #7023, opened 2019-03-06
I accidentally commit locally to protected branches


billygriffin:
**Please describe the problem you think should be solved**

"I frequently commit locally to a branch that's protected on GitHub and one that I can therefore never push to. After I realize I committed to the wrong branch, then I have to either Undo however many commits I've made, switch branches, and do the commits again on a new branch or I have to use a different method to get those commits over to a different branch."

Branches on GitHub are protected for a reason (so there's a particular level of scrutiny before code makes its way onto them), and it doesn't make sense why GitHub Desktop shouldn't have knowledge of this central concept of GitHub's overall model.

A version of this was brought up in #1021 and we explored this a bit in #4647, but I'd like to refocus the discussion around the problem before we move toward solutions.

Reference GitHub protected branch docs: https://help.github.com/en/articles/about-protected-branches

**Open questions**

1. Is this only relevant for protected branches, or are there other instances where people never want to commit to a particular branch? One that has come up previously is the default (often `master`) branch, but I want to question that assumption.
2. If a branch is protected, is there *ever* a reason someone would want to commit to that branch locally? Our assumption thus far is no.

If you have thoughts about either of the open questions, we're really interested in understanding this better.

**Success metrics**

1. If our assumptions are correct and we succeed in solving this problem, we'd anticipate that the number of commits in Desktop to a branch that's protected would drop significantly (by at least half, or potentially even to zero).

shiftkey:
I found myself updating the `push` implementation in Desktop to support `--force-with-lease`. This required handling when `ProtectedBranchForcePush` is reported, to ensure its handled correctly and a meaningful error is reported to the user.

https://github.com/desktop/desktop/blob/35f72e065d684b2ddbff32410b216d07a0e57e35/app/src/lib/git/push.ts#L65-L71

This might  a chance to identify how often users `push` to a protected branch currently and see this error:

 - Pros
   - we don't need to be aware of the protected branches configured - this could ship independent of any exploratory work and give an interesting data point of the pain seen currently by users
   - supporting rebase flows might provide an uptick of users who encounter this problem

 - Cons
  - this issue is about `commit` time checks, which is earlier in the workflow than this area we might be able to instrument and get insights

ampinsk:
I did some thinking here and chatted with @billygriffin. We believe this problem would better served by not letting users commit to protected branches in the first place, rather than letting them move commits after the fact. Here's the flow we're proposing: 

## Proposed flow 

### Lock commit area
[...]

PR #7595, opened 2019-05-21
record metrics for protected branch scenarios due for 1.7

### LINKED ISSUE #7091: [Metrics] Commits to protected branches

shiftkey:
## Overview

**Second PR to address #7091**

## Description

- [x] add `commitsToProtectedBranch` and `commitWithBranchProtectionsEnabled` to help understand how protected branches are encountered in the wild
- [x] update metrics docs with this information
- [x] address gaps with `RepositoriesStore` not being aware of "branch protection is enabled" scenario
- [x] wire up metrics when committing
- [x] tune caching so that at worst only one DB call occurs when `RepositoriesStore.getBranchProtectionContext` runs
- [x] rebase on top of `development` once #7529 has been merged
- [x] open Central PR to consume and pass through these metrics - https://github.com/github/central/pull/434
- [x] open marketing site PR to document these metrics - https://github.com/github/desktop.github.com/pull/132

I believe this is feature-complete now, and will go out with the next beta. I'll follow up with a PR to enable it for production based on how further testing on beta looks.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -140,6 +140,9 @@ export interface IDailyMeasures {
   /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
   readonly commitsToProtectedBranch: number
 
+  /** A flag indicating the user made a commit to a repository with branch protections enabled */
+  readonly commitWithBranchProtectionsEnabled: boolean
+
   /** The number of times the user dismissed the merge conflicts dialog */
   readonly mergeConflictsDialogDismissalCount: number
 
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -137,6 +137,9 @@ export interface IDailyMeasures {
   /** The number of time the user made a commit to a repo hosted on Github.com */
   readonly dotcomCommits: number
 
+  /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
+  readonly commitMadeToProtectedBranch: number
+
   /** The number of times the user dismissed the merge conflicts dialog */
   readonly mergeConflictsDialogDismissalCount: number
 
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -140,8 +140,8 @@ export interface IDailyMeasures {
   /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
   readonly commitsToProtectedBranch: number
 
-  /** A flag indicating the user made a commit to a repository with branch protections enabled */
-  readonly commitWithBranchProtectionsEnabled: boolean
+  /** The number of times the user made a commit to a repository with branch protections enabled */
+  readonly commitsToRepositoryWithBranchProtections: number
 
   /** The number of times the user dismissed the merge conflicts dialog */
   readonly mergeConflictsDialogDismissalCount: number
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -138,7 +138,7 @@ export interface IDailyMeasures {
   readonly dotcomCommits: number
 
   /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
-  readonly commitsMadeToProtectedBranch: number
+  readonly commitsToProtectedBranch: number
 
   /** The number of times the user dismissed the merge conflicts dialog */
   readonly mergeConflictsDialogDismissalCount: number
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -138,7 +138,7 @@ export interface IDailyMeasures {
   readonly dotcomCommits: number
 
   /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
-  readonly commitMadeToProtectedBranch: number
+  readonly commitsMadeToProtectedBranch: number
 
   /** The number of times the user dismissed the merge conflicts dialog */
   readonly mergeConflictsDialogDismissalCount: number
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -134,7 +134,7 @@ export interface IDailyMeasures {
    */
   readonly enterpriseCommits: number
 
-  /** The number of time the user made a commit to a repo hosted on Github.com */
+  /** The number of times the user made a commit to a repo hosted on Github.com */
   readonly dotcomCommits: number
 
   /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2383,14 +2383,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
           if (upstreamWithoutRemote !== null) {
             const {
-              branchProtectionCount,
+              branchProtectionsFound,
               isProtected,
             } = await this.repositoriesStore.getBranchProtectionContext(
               repository.gitHubRepository,
               upstreamWithoutRemote
             )
 
-            if (branchProtectionCount > 0) {
+            if (branchProtectionsFound) {
               this.statsStore.recordCommitWithBranchProtectionsEnabled()
             }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -163,6 +163,7 @@ import {
   IMatchedGitHubRepository,
   matchGitHubRepository,
   repositoryMatchesRemote,
+  urlMatchesCloneURL,
 } from '../repository-matching'
 import {
   initializeRebaseFlowForConflictedRepository,
@@ -2381,7 +2382,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
           // look up if the tracked branch matches a protected remote branch
           const { upstreamWithoutRemote } = gitStore.tip.branch
 
-          if (upstreamWithoutRemote !== null) {
+          // ensure the remote associated with this branch is the one we
+          // are checking the branch protections against
+          const remoteURL = gitStore.currentRemote.url
+
+          if (
+            upstreamWithoutRemote !== null &&
+            urlMatchesCloneURL(remoteURL, repository.gitHubRepository)
+          ) {
             const {
               branchProtectionsFound,
               isProtected,
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2388,9 +2388,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
             )
 
             if (isProtected) {
-              log.debug(
-                `[_commitIncludedChanges] the upstream ref '${upstreamWithoutRemote}' is protected by the GitHub API`
-              )
+              this.statsStore.recordCommitMadeToProtectedBranch()
             }
           }
         }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2399,7 +2399,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
             )
 
             if (branchProtectionsFound) {
-              this.statsStore.recordCommitWithBranchProtectionsEnabled()
+              this.statsStore.recordCommitToRepositoryWithBranchProtections()
             }
 
             if (isRemoteBranchProtected) {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2382,10 +2382,17 @@ export class AppStore extends TypedBaseStore<IAppState> {
           const { upstreamWithoutRemote } = gitStore.tip.branch
 
           if (upstreamWithoutRemote !== null) {
-            const isProtected = await this.repositoriesStore.isBranchProtected(
+            const {
+              branchProtectionCount,
+              isProtected,
+            } = await this.repositoriesStore.getBranchProtectionContext(
               repository.gitHubRepository,
               upstreamWithoutRemote
             )
+
+            if (branchProtectionCount > 0) {
+              this.statsStore.recordCommitWithBranchProtectionsEnabled()
+            }
 
             if (isProtected) {
               this.statsStore.recordCommitToProtectedBranch()
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2388,7 +2388,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
             )
 
             if (isProtected) {
-              this.statsStore.recordCommitMadeToProtectedBranch()
+              this.statsStore.recordCommitToProtectedBranch()
             }
           }
         }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -163,7 +163,6 @@ import {
   IMatchedGitHubRepository,
   matchGitHubRepository,
   repositoryMatchesRemote,
-  urlMatchesCloneURL,
 } from '../repository-matching'
 import {
   initializeRebaseFlowForConflictedRepository,
@@ -238,6 +237,7 @@ import { IStashEntry, StashedChangesLoadStates } from '../../models/stash-entry'
 import { RebaseFlowStep, RebaseStep } from '../../models/rebase-flow-step'
 import { arrayEquals } from '../equality'
 import { MenuLabelsEvent } from '../../models/menu-labels'
+import { findBranchName } from './helpers/find-branch-name'
 
 /**
  * As fast-forwarding local branches is proportional to the number of local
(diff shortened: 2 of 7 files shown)


shiftkey:
@nerdneha @desktop/engineering if you've got time today/tomorrow, could you sanity check that these `commitsToProtectedBranch` and `commitWithBranchProtectionsEnabled` make sense with respect to the list of things we want to capture in the original issue?

Here's the original list, with my current thoughts about how this PR addresses them:

> 1. How many users fit the above category (people who have committed to a repo with at least one protected branch)

I think this is covered by `commitWithBranchProtectionsEnabled` - this will be set to `true` when the user makes a commit and the repository has branch protections. This will reset every day after the usage stats are submitted.

> 2. How many commits per user to a protected branch

This should be covered by `commitsToProtectedBranch` - this will increment whenever a commit is shown to belong to a protected branch.

> 3. How frequently it's occurring for people

We have various related commit metrics already - `commits` is the absolute count, and `dotcomCommits` and `enterpriseCommits` are the subset of these for known remotes. "Frequently" here suggests there's a time dimension to this, but I'm not sure if we have a good precedence for this.

To put this simply as a percentage value (so from "never" to "always") this could be computed by

 - `commitsToProtectedBranch / commits` to include all commits
 - `commitsToProtectedBranch / (dotcomCommits + enterpriseCommits)` is a bit more accurate (excludes commits for local-only and non-GitHub remotes)

> 4. % of people accidentally committing to a protected branch

I think this just represents all users with `commitsToProtectedBranch > 0`.

> 5. On repos with protected branches, % of commits to those branches

This feels like `#3` but filtered only to users where `commitWithBranchProtectionsEnabled` is `true`.

shiftkey:
Marking this as `ready-for-review` while we confirm the captured metrics satisfy the goals of the issue

mido1966:
Pls No Email mor pls Tank you



Von meinem Samsung Galaxy Smartphone gesendet.


-------- Ursprüngliche Nachricht --------
Von: Brendan Forster <notifications@github.com>
Datum: 22.05.19 16:03 (GMT+01:00)
An: desktop/desktop <desktop@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Betreff: Re: [desktop/desktop] record metrics for protected branch scenarios due for 1.7 (#7595)


Marking this as ready-for-review while we confirm the captured metrics satisfy the goals of the issue

app/src/lib/stats/stats-database.ts
```diff
@@ -134,9 +134,15 @@ export interface IDailyMeasures {
...
 
-  /** The number of time the user made a commit to a repo hosted on Github.com */
+  /** The number of times the user made a commit to a repo hosted on Github.com */
   readonly dotcomCommits: number
 
+  /** The number of times the user made a commit to a protected GitHub or GitHub Enterprise repository */
+  readonly commitsToProtectedBranch: number
+
+  /** A flag indicating the user made a commit to a repository with branch protections enabled */
+  readonly commitWithBranchProtectionsEnabled: boolean
```
outofambit: non-blocking, but perhaps `committedInRepoWithBranchProtections` could be more clear? to make it more clear that the repo has the branch protections and not that a commit was made on a protected branch.

(the doc for this is totally clear to me though)

app/src/lib/stores/repositories-store.ts
```diff
@@ -531,31 +557,100 @@ export class RepositoriesStore extends BaseStore {
...
-    const key = getKey(gitHubRepository.dbID, branchName)
+    const repoID = gitHubRepository.dbID
+
+    const branchProtectionsFound = this.branchProtectionSettingsFoundCache.get(
+      repoID
+    )
+
+    if (branchProtectionsFound === undefined) {
+      return this.loadAndCacheBranchProtection(repoID, branchName)
+    } else if (branchProtectionsFound) {
```
outofambit: this is redundant but made this easier for me to follow since it wasn't immediately apparent that `branchProtectionsFound` was a `boolean`. definitely non-blocking :)

```suggestion
    } else if (branchProtectionsFound === true) {
```

app/src/lib/stores/app-store.ts
```diff
@@ -2384,16 +2385,28 @@ export class AppStore extends TypedBaseStore<IAppState> {
           // look up if the tracked branch matches a protected remote branch
           const { upstreamWithoutRemote } = gitStore.tip.branch
 
-          if (upstreamWithoutRemote !== null) {
-            const isProtected = await this.repositoriesStore.isBranchProtected(
+          // ensure the remote associated with this branch is the one we
```
outofambit: should all of this be behind the `enableBranchProtectionWarning()` flag (on line `2381`)? currently that flag is set to `development` only, so i'm pretty sure this should be in a separate conditional or that the feature flag should be changed (but i don't think i fully understand the implications of turning that feature flag on). otherwise we won't collect metrics for this outside of `development`.

outofambit:
> if you've got time today/tomorrow, could you sanity check that these commitsToProtectedBranch and commitWithBranchProtectionsEnabled make sense with respect to the list of things we want to capture in the original issue?

these all look good to me, but i have questions about:

> How frequently it's occurring for people

is that how often they're committing to protected branches? or how often they try to push to a protected branch and get rejected? its pretty vague in #7091 but i'm leaning towards the former, which i believe is covered with the calculation @shiftkey outlined above (and thus by the implementation in this PR). @nerdneha can double check this in case we're reading this wrong?

app/src/lib/stores/repositories-store.ts
```diff
@@ -531,31 +557,100 @@ export class RepositoriesStore extends BaseStore {
...
+  /**
+   * Load the branch protection information for a repository from the database
+   * and cache the results in memory
+   */
+  private async loadAndCacheBranchProtection(
+    repoID: number,
+    branchName: string
+  ) {
+    // query the database to find any protected branches
+    const branches = await this.db.protectedBranches
```
outofambit: i'm guessing this is not super fast based on our other performance investigations. the caching should help a lot with that such that this should only slow down the first commit in each repository per run of the app, correct?

app/src/lib/stores/app-store.ts
shiftkey: > otherwise we won't collect metrics for this outside of `development`.

I was planning to follow up with a trivial "enable this for beta" PR after this landed and before we were ready to enable this for beta, just in case it interfered with something more urgent.

app/src/lib/stores/repositories-store.ts
shiftkey: Correct. Ideally this should already be cached when `RepositoriesStore.updateGitHubRepository` was run, but for the edge case of a user making a commit immediately after starting (before that API refresh kicks in), it may not be cached - so we'll need to check the database.

And I realise now in this case we don't cache the `branchProtectionsFound` boolean in memory here, just the branch entries found. I'll address that.

shiftkey: Addressed in 6f8c908a4

shiftkey: I poked at this using `nodejs/node` (which has 12 protected branches) and the time to run this query was only **3ms** on my macOS machine, so I'm fine with the overhead here when we need to do this.

But given I already had caching in from the previous PR it made sense, I think sticking to this pattern is fine given this needs to run when committing - and also be low-overhead.

Going to mark this as resolved but let me know if you have any other questions about this.

app/src/lib/stats/stats-database.ts
shiftkey: Gonna hold of on this potential rename until we can confirm we're not misunderstanding the requirements in https://github.com/desktop/desktop/pull/7595#issuecomment-494925379

shiftkey:
> is that how often they're committing to protected branches? or how often they try to push to a protected branch and get rejected? its pretty vague in #7091 but i'm leaning towards the former, 

This is also my feeling to, because the check for "the push was rejected" is much later in the flow, and was separate enough from the rest of the questions about commit time that I didn't focus on anything `push`-related.

I wrote up some notes about how we _could_ detect failing pushes being related to protected branches in https://github.com/desktop/desktop/issues/7023#issuecomment-471618918, if you want to understand the potential extra work here.

nerdneha:
> 1. How many users fit the above category (people who have committed to a repo with at least one protected branch)

We have `commitWithBranchProtectionsEnabled` set to true.

This should be fine since the `userId` is sent along in the package. If I understand correctly, from the consumption side, we could just count the number of users who have had this set to true in the last month. 

> 2. How many commits per user to a protected branch

We have `commitsToProtectedBranch` for the number of commits to a protected branch.

Would this be per repo or across the entire user's account? It might be helpful to also know how many repos that is across also but I'm not 100% sure here. 

> 3. How frequently it's occurring for people

We're probably looking for whether the amount of times people commit to a protected branch increases or decreases over time. I think having `commitsToProtectedBranch` is sufficient here, but I think distilling it down to a per repo basis might give us some extra granularity (see question for 4).

> is that how often they're committing to protected branches? or how often they try to push to a protected branch and get rejected? its pretty vague in #7091 but i'm leaning towards the former, which i believe is covered with the calculation @shiftkey outlined above (and thus by the implementation in this PR). @nerdneha can double check this in case we're reading this wrong?

i think this is "attempted commits" (aka committing to protected branches, as @outofambit said) -- so it covers commits and rejections

4. % of people accidentally committing to a protected branch

I agree, it's # of users with `commitsToProtectedBranch` > 0 / # of users with `commitsWithProtectedBranchesEnabled` = true.  (or potentially # of users period)

5. On repos with protected branches, % of commits to those branches

I think this is `#4` but on a repo level instead of a user level; we'd need to know the number of commits to a protected branch per user per repo to answer this.

shiftkey:
@nerdneha thanks for the extra context!

> > 2. How many commits per user to a protected branch
> 
> We have `commitsToProtectedBranch` for the number of commits to a protected branch.
>
> Would this be per repo or across the entire user's account? It might be helpful to also know how many repos that is across also but I'm not 100% sure here.

This metric is not broken down per-repository, and I'd love to unpack what "per repo" means here because I don't think I have a good grasp on how this should be computed. Could you elaborate?

> i think this is "attempted commits" (aka committing to protected branches, as @outofambit said) -- so it covers commits and rejections

Based on my understanding of what you're saying here, are you interesting in how may times a `git push` is rejected because of branch protection issues? Because that's not supported currently in this PR, and is a separate area to our current focus on commit authoring in #7023 and #7091.

app/src/lib/stores/app-store.ts
```diff
@@ -2384,16 +2385,28 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
               repository.gitHubRepository,
               upstreamWithoutRemote
             )
 
-            if (isProtected) {
-              log.debug(
-                `[_commitIncludedChanges] the upstream ref '${upstreamWithoutRemote}' is protected by the GitHub API`
-              )
+            if (branchProtectionsFound) {
+              this.statsStore.recordCommitWithBranchProtectionsEnabled()
```
niik: I believe that for this metric for be able to fulfil the objectives outlined in the issue, in particular the one quoted below, we're going to have to track this regardless of whether the current branch is published or not. The condition should simply be whether the repository has *any* protected branches or not.

> On repos with protected branches, % of commits to those branches

nerdneha:
@shiftkey thanks for the great points there. addressing the "per repo" piece here and will address the "attempted commits" piece in a subsequent comment. @niik and I took a look at the metrics you're collecting and have a change that can address the "per repo" piece elegantly. 

Ultimately, we want to know the `number of attempted commits to a protected branch` /  `commits overall to a repo with branch protected enabled`

`number of attempted commits to a protected branch`: this is `commitsToProtectedBranch`

`commits overall to a repo with branch protected enabled`: since we right now have `commitWithBranchProtectionsEnabled` as a boolean, if we turn this into an incremented number, that will provide this measurement and then, for the boolean case, we can just see if it's >0. 

Would this be too big of a change to make today?

nerdneha:
> i think this is "attempted commits" (aka committing to protected branches, as @outofambit said) -- so it covers commits and rejections

> Based on my understanding of what you're saying here, are you interesting in how may times a git push is rejected because of branch protection issues? Because that's not supported currently in this PR, and is a separate area to our current focus on commit authoring in #7023 and #7091.

English language is hard y'all. Sorry about the confusion, what I was trying to say is that we shouldn't have to worry about any case where a push is rejected. What we have is sufficient and no extra work is needed.

When we said `How frequently it's occurring for people`, honestly I think we're trying to measure the amount of wasted effort we can move the needle on and that happens the minute they click on the `commit` button, regardless of whether or not it gets eventually rejected (ex: someone could realize before they push but they still have to put in extra work). If we collect "attempted commits" aka, `commitsToProtectedBranch`, then I think we have the right metric to measure the problem and eventually the impact.

shiftkey:
@nerdneha 

> Would this be too big of a change to make today?

Not too big of a change, just wanted to understand the context as it wasn't familiar to me and make sure I didn't head down the wrong path.

> What we have is sufficient and no extra work is needed.

👍

nerdneha:
> Not too big of a change, just wanted to understand the context as it wasn't familiar to me and make sure I didn't head down the wrong path.

totally, this was a super valid clarification and I'm glad we could take a step back and avoid actually collecting metrics on a per repo basis for this feature! thank you!!! 🙏 🙇‍♀

app/src/lib/stores/app-store.ts
shiftkey: @niik I'll have a go at rejigging this now that @nerdneha also followed up. I encountered some scenarios where working in forks meant it felt more reliable to look at the tracking information, but I'll see if I can make this simpler to just rely on `repository.githubRepository`

shiftkey:
I've update this PR to add two commits: 

 - https://github.com/desktop/desktop/pull/7595/commits/6520ce892c86cd65434341e5c5bb056284b219be - this commit changes the previous flag to a counter now named `commitsToRepositoryWithBranchProtections` - hopefully that name is clear enough to differentiate it from the existing `commitsToProtectedBranch`
 - https://github.com/desktop/desktop/pull/7595/commits/006d484692792ac393e7a301df16259e6b92a5cb - this commit refactors the flow to always check whether branch protections are found for the current repository, and performs a simpler check on the current branch to see if it would be considered "protected"

I'm still refining the flow to confirm the logic is correct and it's caching as much as possible, but I wanted to get this out for other :eyes: while it's fresh.

The goal with this last commit was to relax the flow a bit to:

 - favour the GitHub API over whatever remote is found in application state
 - check the upstream as well as the fork if both are found in the GitHub API
 - use the local branch name if no tracking branch is found

Feel free to challenge those assumptions if you think there's a better flow here

app/src/lib/stats/stats-database.ts
shiftkey: After further discussions this has been renamed to `commitsToRepositoryWithBranchProtections` in https://github.com/desktop/desktop/pull/7595/commits/6520ce892c86cd65434341e5c5bb056284b219be

app/src/lib/stores/repositories-store.ts
```diff
@@ -531,33 +546,132 @@ export class RepositoriesStore extends BaseStore {
...
+
+    if (isBranchProtectionsFound) {
+      return true
+    }
 
-    const existing = this.branchProtectionCache.get(key)
-    if (existing !== undefined) {
-      return existing
+    if (
+      gitHubRepository.parent !== null &&
```
shiftkey: After thinking about this some more, I'm going to skip looking at the parent repository for branch protections for two reasons:

 - by checking both we're probably going to collect a lot of OSS projects that protect their default branch, but that's not relevant to people working in forks
 - when you're working in a fork, the typical workflow isn't to push to the upstream - you'd create a branch in your fork and submit that as a pull request

app/src/lib/stores/repositories-store.ts
```diff
@@ -531,33 +546,132 @@ export class RepositoriesStore extends BaseStore {
...
-    return value
+    if (isProtected) {
+      return true
+    }
+
+    if (
+      gitHubRepository.parent !== null &&
+      gitHubRepository.parent.dbID !== null
+    ) {
+      const parentID = gitHubRepository.parent.dbID
```
shiftkey: And I'm going to skip this case here because we don't actually fetch branch protections for the upstream branch, so unless you already have this populated because it's also tracked in the app, this is not going to be present in the database.

And this feels like it's not focused on the workflow we're interested in - **commits to a protected branch on the current remote**. If we start trying to be clever here by being aware of the upstream repository I think it's going to distort the numbers here.

shiftkey:
I think this is ready for another round of review 🤘

app/src/lib/stores/repositories-store.ts
outofambit: yep i think that's a reasonable assumption! 🍴

app/src/lib/stores/app-store.ts
outofambit: this is addressed in 006d484 correct?

shiftkey: Actually, I _think_ this is covered by 6520ce892c86cd65434341e5c5bb056284b219be

outofambit: ah yes, that's right! thank you!

app/src/lib/stores/helpers/find-branch-name.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Function to determine which branch name to use when looking for branch
+ * protection information.
+ *
+ * If the remote branch matches the current `githubRepository` associated with
+ * the repostiory, this will be used. Otherwise we will fall back to using the
+ * branch name as that's a reasonable approximation for what would happen if the
+ * user tries to push the new branch.
+ */
+export function findRemoteBranchName(
```
outofambit: ✨

app/src/lib/stores/repositories-store.ts
outofambit: thanks for the extra info!

shiftkey:
> @shiftkey should we flip `enableBranchProtectionWarning` to `beta` in this PR so it's good to go for tomorrow's beta?

@outofambit I'll push a commit to this branch now to enable this, and we can get a gut check in the morning from everyone else about how they feel for this to get into beta.

shiftkey:
Just updated the PR description to indicate this will be enabled for the next beta release, and I'll follow up with a PR to enable it for production if we're happy to turn it on for 1.7.

PR #6916, opened 2019-02-21
[Branch Pruning] Refactor to allow for easier testing of the branch pruner

iAmWillShepherd:
This PR was initially for @desktop/quality-assurance, but some of the changes introduced in the PR makes the code a bit more readable.

Changes in this PR:
* Clearer comments
* New menu item in the top menu
* Clearer logging
* Extracted function to make logic a bit easier to follow
* Branches ready for pruning are no longer deleted in dev mode and are instead logged to the console

screen shot 2019-02-21 at 1 58 24 pm

I think this is a good thing to ship since we may need to debug the pruner in the future and this definitely helps with that.

PR #7616, opened 2019-05-23
add check that remote ref does not exist when pruning

### LINKED ISSUE #7604: Branch pruner not using remote info to determine if a branch can be pruned
There is a currently a bug in the branch pruner that causes branches to be pruned when they have not been deleted from the remote. 

Screen Shot 2019-04-03 at 4 25 53 PM

**Reproduction steps**
1. Create a PR on dotcom
2. Merge the PR into the default branch, but **don't** delete the branch
3. Go back to Desktop and perform a fetch
4. Click Help -> Prune branches and observe the console
5. You should notice the branch that was recently merged is _marked_ as prune-able

**What I expected**
No branches should have been marked as prune-able since only 2 of 3 condition are met: 
* Branch is merged into default branch
* Branch hasn't been checked out in 14 days (this constraint is skipped when selecting Help -> Prune branches)

**Root cause**
This bug is caused by not checking to see if the remote still exists on the server. The fix would involve either using `git ls-remote` to get a list of all refs on the remote or using state that Desktop already has. I was planning to pair with @niik on implementing the latter of the two options.

_Originally posted by @iAmWillShepherd in https://github.com/desktop/desktop/pull/6916#issuecomment-479664139_

shiftkey:
## Overview

**Closes #7604**

This fix wasn't as trivial as I'd hoped it would be, but I'll walk you through the changes and how we ended up here.

 - [ ] gather feedback on proposed changes
 - [ ] tidy up commit history to make changes easier to follow
 - [ ] more testing of corner cases

## Description

The intended fix was described like this:

> The fix would involve either using `git ls-remote` to get a list of all refs on the remote or using state that Desktop already has.

As we want to avoid the network call associated with `ls-remote`, and the branches for pruning should be stale for at least two weeks, the application state should be safe enough to rely on here.

### Problem 1 - Where are my branches?

Being able to enumerate application state to find  the remote branch didn't work, because of how we merge together local and remote branches. I added tests to `GitStore` to confirm this behaviour in d66a271e2. I didn't want to change how `GitStore.mergeRemoteAndLocalBranches` works just to address this bug, so I fell back to invoking `getBranches` to look for this specific ref. I think this is fine for what we're trying to do here - we filter out plenty of conditions before we get to this point, so this should only run for a subset of the merged branches.

### Problem 2 - Now it's async

With that working, a different problem appeared - `Array.filter` (where all the checks are being performed) doesn't support `async/await`. I needed to change that to a `for` loop to build up the results. This now lives in `BranchPruner.filterBranch` and I've tried to preserve the flow of the previous rules as much as possible to avoid further confusion.

### Problem 3 - Testing Things

The feedback loop on testing this on `development` is not really available, and #6916 had some logic to make this easier to test in dev mode. I borrowed from that and tried to keep this as simple as possible, but I eventually came to the realization that this needs runtime flags to control the behaviour of the branch pruner, and ended up with the creatively-named `PruneRuntimeOptions` type to represent this:

```ts
/**
 * Behavior flags for the branch prune execution, to aid with testing and
 * verifying locally.
 */
type PruneRuntimeOptions = {
  /**
   * By default the branch pruner will only run every 24 hours
   *
   * Set this flag to `false` to ignore this check.
   */
  readonly enforcePruneThreshold: boolean
  /**
   * By default the branch pruner will also delete the branches it believes can
   * be pruned safely.
   *
   * Set this to `false` to keep these in your repository.
   */
  readonly deleteBranch: boolean
}
```

This is a bit more explicit than what we originally had in  #6916, but I hope this makes the overall component easier to follow. Here's the default options:

```ts
const DefaultPruneOptions: PruneRuntimeOptions = {
  enforcePruneThreshold: true,
  deleteBranch: true,
}
```

And here's what the test when running this in development mode does:

```ts
  public async testPrune(): Promise<void> {
    return this.pruneLocalBranches({
      enforcePruneThreshold: false,
      deleteBranch: false,
    })
  }
```

I resisted the urge to put these into one "flag" because while they feel similar, they represent two very distinct decisions (run the branch pruner checks, versus deleting the valid refs) and being able to switch one or the other off for testing feels like something worth supporting.

### Good news!

By going through the process of implementing this flow, I found a test was now impacted. Fearing I'd regressed something, I looked into it and found this scenario was now being handled correctly:

```ts
    const expectedBranchesAfterPruning = [
      'not-deleted-branch-1',
      'deleted-branch-1',
    ]

    for (const branch of expectedBranchesAfterPruning) {
      expect(branchesAfterPruning).not.toContain(branch)
    }
```

I remember talking with @iAmWillShepherd about this at some stage, but `not-deleted-branch-1` was an example of a merged branch with a valid remote ref, so now we correctly handle this in the branch pruner, which was Precisely The Bug™ :tada:

This test is now updated to reflect this change.

```ts
    expect(branchesAfterPruning).not.toContain('deleted-branch-1')
    expect(branchesAfterPruning).toContain('not-deleted-branch-1')
```

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/helpers/branch-pruner.ts
+++ b/app/src/lib/stores/helpers/branch-pruner.ts
@@ -158,7 +158,7 @@ export class BranchPruner {
 
     // Get list of branches that have been merged
     const { branchesState } = this.repositoriesStateCache.get(this.repository)
-    const { defaultBranch, allBranches } = branchesState
+    const { defaultBranch } = branchesState
 
     if (defaultBranch === null) {
       return
@@ -188,63 +188,12 @@ export class BranchPruner {
       [...recentlyCheckedOutBranches.keys()].map(formatAsLocalRef)
     )
 
-    // Create array of branches that can be pruned
-    const candidateBranches = mergedBranches.filter(
-      mb => !ReservedRefs.includes(mb.canonicalRef)
-    )
-
     const branchRefPrefix = `refs/heads/`
 
-    const branchesReadyForPruning = candidateBranches.filter(async mb => {
-      if (recentlyCheckedOutCanonicalRefs.has(mb.canonicalRef)) {
-        // branch was recently checked out, exclude it from pruning
-        return false
-      }
-
-      if (!mb.canonicalRef.startsWith(branchRefPrefix)) {
-        // branch does not match the expected conventions, exclude it
-        return false
-      }
-
-      const branchName = mb.canonicalRef.substr(branchRefPrefix.length)
-
-      const localBranch = allBranches.find(
-        b => b.type === BranchType.Local && b.name === branchName
-      )
-
-      if (localBranch === undefined) {
-        // unable to find the local branch in repository state, this feels
-        // sufficient concerning to indicate we shouldn't try and delete this
-        // reference
-        return false
-      }
-
-      if (localBranch.upstream === null) {
-        // no remote ref is being tracked for this branch, which means it
-        // most likely wasn't pushed to the remote, so we can include this
-        // in our prune list
-        return true
-      }
-
-      // we cannot use `allBranches` here because it merges together local and
-      // remote branches to de-dupe results, so instead we're going to make a
-      // Git call, one per candidate branch that reaches this point, and see if
-      // the remote ref exists
-      //
-      // As the app runs `git fetch --prune` regularly, this is a good enough
-      // indicator that the remote ref is gone especially when we're only
-      // filtering local branches that have not been used in the past 2 weeks
-      const remoteBranch = await getBranches(
-        this.repository,
-        `refs/remotes/${localBranch.upstream}`
-      )
-
-      if (remoteBranch.length === 0) {
-        return true
-      }
-
-      return false
-    })
+    const branchesReadyForPruning = await this.filterBranches(
+      mergedBranches,
+      recentlyCheckedOutCanonicalRefs
+    )
 
     log.info(
       `[Branch Pruner] pruning ${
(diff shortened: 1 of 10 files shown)


app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -28,6 +29,33 @@ const ReservedRefs = [
   'refs/heads/release',
 ]
 
+const BranchRefPrefix = `refs/heads/`
+
+/**
+ * Behavior flags for the branch prune execution, to aid with testing and
+ * verifying locally.
+ */
+type PruneRuntimeOptions = {
```
iAmWillShepherd: I like that this type makes the logic for testing locally a bit easier to digest.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -64,6 +92,13 @@ export class BranchPruner {
     this.timer = null
   }
 
+  public async testPrune(): Promise<void> {
```
iAmWillShepherd: This function is super convenient!

shiftkey:
@desktop/engineering I'd love your thoughts on two questions I have here before I go further:

 1. Given the changes are a bit more involved than anticipated, are you still interested in seeing this land for 1.7.0 so we can enable branch pruning for everyone?

 2. Alternatively, if I break this up into three separate PRs, do you feel this would reduce the risk and/or make it more appealing to land for 1.7.0?

I believe I can break this into three independent PRs that build on top of each other:
 
 - improve the dev-time experience for testing branch pruning
 - refactor the filtering of branches so it can use `async/await`
 - address the bug and the impacted tests

iAmWillShepherd:
> Given the changes are a bit more involved than anticipated, are you still interested in seeing this land for 1.7.0 so we can enable branch pruning for everyone?

While it would certainly be nice to have this ship in the next release. I don't feel that strongly that it _has_ to be in. I think holding off until post 1.7 would be useful given the additional risk involved. I'll let others chime in, but my vote given the context is to hold off on shipping this in 1.7.0.

outofambit:
Thanks for writing all of this up and sharing @shiftkey!

> Given the changes are a bit more involved than anticipated, are you still interested in seeing this land for 1.7.0 so we can enable branch pruning for everyone?

This is definitely more involved than I was expecting. I started researching an idea that might be simpler (and will share more about that tomorrow) but regardless I think we should wait to avoid extra risk for this release (and give us time to take a step back and evaluate some options, for example: should we investigate changing the behavior of `GitStore.mergeRemoteAndLocalBranches`?).

nerdneha:
> are you still interested in seeing this land for 1.7.0 so we can enable branch pruning for everyone?

I definitely think we have enough going on in 1.7.0, especially around branches, to put a hold on branch pruning so
a) we can decide what solution we want (ex: this PR, whatever @outofambit presents, something around addressing getBranches fundamentally, or something else)
b) give some air on beta to whatever we land on; preferably a month or so

@shiftkey unless there's any reason you feel otherwise, i think it's safe to remove this from the `1.7.0` milestone.

shiftkey:
Thanks for the input everyone. I've dropped this PR from the milestone in the meantime and here's my next set of action items:

 - close this PR, as now that I've gone through the work and wrapped my head around the impact I know there's a more optimal way to make these commits
 - see if @outofambit has an alternate proposal to address the original bug
 - open a PR to address "improve the dev-time experience for testing branch pruning" - this feels important enough to land independent of the bugfix to help with testing and verifying the branch pruning in development mode. Doesn't need to go in for 1.7.0 either, but having the API available as soon as practical would be great.

PR #7629, opened 2019-05-24
improve the dev-time experience of testing branch pruning

shiftkey:
## Overview

**Related to #7604 and #6916**

**Supersedes #6916**

## Description

This PR implements two runtime flags to `BranchPruner.pruneLocalBranches` to control how it behaves:

 - `enforcePruneThreshold` - whether to check and respect the prune threshold tracked in IndexedDB for the repository
 - `deleteBranch` - whether to actually remove the branch from the local repository

These are both set to `true` by default, but this PR also adds a menu item (visible when `__DEV__` is set) to be able to test the branch pruning logic against a live application.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/helpers/branch-pruner.ts
+++ b/app/src/lib/stores/helpers/branch-pruner.ts
@@ -4,7 +4,7 @@ import { Branch } from '../../../models/branch'
 import { GitStoreCache } from '../git-store-cache'
 import {
   getMergedBranches,
-  getCheckoutsAfterDate,
+  getBranchCheckouts,
   getSymbolicRef,
   IMergedBranch,
   formatAsLocalRef,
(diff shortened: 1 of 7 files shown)


app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -167,14 +210,16 @@ export class BranchPruner {
...
-        deleteLocalBranch(this.repository, branchName)
-      )
+        const isDeleted = await gitStore.performFailableOperation(() =>
+          deleteLocalBranch(this.repository, branchName)
+        )
 
-      if (isDeleted) {
-        log.info(`Pruned branch ${branchName} (was ${branch.sha})`)
+        if (isDeleted) {
+          log.info(`Pruned branch ${branchName} (was ${branch.sha})`)
```
iAmWillShepherd: I think we want to log a message if the branch _would have been_ deleted. This way, we can verify that the right branches are being tagged for deletion without having to rely on setting breakpoints and stepping through the code.

shiftkey: Check out how 0e820db2004b2b482e9f77aadbad6991be8207ee feels

Issue #6714, opened 2019-01-25
[Metrics] No local changes / suggested next steps usage


billygriffin:
We shipped suggested next steps in the "No local changes" view in 1.6, and it would be helpful to start to get a feel for how those are being used to help us optimize what we enable through that, and help us think about how we consider additional things to highlight and potentially things to remove in that area.

1. Number of times opening the repo in their external editor.
2. Number of times are users viewing the files in Finder/Explorer.
3. Number of times users are opening the repo page on GitHub in their browser.
4. Number of times users Create Pull Request via the suggestion (this feature is not yet in production).
5. Number of times users publish a repository to GitHub.
6. Number of times users publish a branch to GitHub.

Questionable (I'd love thoughts on this, but I'm leaning toward not wanting them captured distinctly from pushes and pulls more generally, assuming we have that?):

1. Pull from branch
2. Push to branch

Anything I missed?

Ideally, we'd capture metrics on each of these things through the suggested next steps, and separately capture the number of times for each of them via their respective keyboard shortcuts.

If possible, I think it would be great to get these metrics in *prior to* shipping the "Create PR" suggestion to production so we can get a feel for whether it helps people use the GitHub PR workflow more frequently by bringing it top of mind when it makes the most sense as a next step.

cc: @desktop/analytics @nerdneha @ampinsk @niik

billygriffin:
I'm comfortable closing this based on #7635 and I think we've gotten fantastic data that PR creation via Desktop has increased dramatically since shipping this, with zero complaints that it has worsened the experience qualitatively. 🎉

PR #7635, opened 2019-05-27
Suggested changes metrics

niik:
## Overview

**Addresses (but does not yet fully close): #6714**

This PR implements the critical piece of #6714, namely the `suggestedStepCreatePR` metric which will allow us to differentiate between how often users use the create pull request suggested next step vs how often they use the menu (the latter already exists in production as `createPullRequestCount`).

Additionally this will give us metrics to evaluate the relative use, and effectiveness, of the various suggested next steps.

I'm tentatively putting this into 1.7.0 since this would be nice to have but if we run up on the deadline this is not a critical piece.

## Description

The entirety of metrics added are as follows

### suggestedStepOpenInExternalEditor

The number of times the user has opened their external editor from the suggested next steps view

### suggestedStepOpenWorkingDirectory

The number of times the user has opened their repository in Finder/Explorer from the suggested next steps view

### suggestedStepViewOnGitHub

The number of times the user has opened their repository on GitHub from the suggested next steps view

### suggestedStepPublishRepository

The number of times the user has used the publish repository action from the suggested next steps view

### suggestedStepPublishBranch

The number of times the user has used the publish branch action branch from the suggested next steps view

### suggestedStepCreatePR

The number of times the user has used the Create PR suggestion in the suggested next steps view. Note that this number is a subset of `createPullRequestCount`. I.e. if the Create PR suggestion is invoked both `suggestedStepCreatePR` and `createPullRequestCount` will increment whereas if a PR is created from the menu or from a keyboard shortcut only `createPullRequestCount` will increment.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -220,7 +220,11 @@ export interface IDailyMeasures {
 
   /**
    * The number of times the user has used the Create PR suggestion
-   * in the suggested next steps view
+   * in the suggested next steps view. Note that this number is a
+   * subset of `createPullRequestCount`. I.e. if the Create PR suggestion
+   * is invoked both `suggestedStepCreatePR` and `createPullRequestCount`
+   * will increment whereas if a PR is created from the menu or from
+   * a keyboard shortcut only `createPullRequestCount` will increment.
    */
   readonly suggestedStepCreatePR: number
 }
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -192,37 +192,37 @@ export interface IDailyMeasures {
    * The number of times the user has opened their external editor from the
    * suggested next steps view
    */
-  suggestedStepOpenInExternalEditor: number
+  readonly suggestedStepOpenInExternalEditor: number
 
   /**
    * The number of times the user has opened their repository in Finder/Explorer
    * from the suggested next steps view
    */
-  suggestedStepOpenWorkingDirectory: number
+  readonly suggestedStepOpenWorkingDirectory: number
 
   /**
    * The number of times the user has opened their repository on GitHub from the
    * suggested next steps view
    */
-  suggestedStepViewOnGitHub: number
+  readonly suggestedStepViewOnGitHub: number
 
   /**
    * The number of times the user has used the publish repository action from the
    * suggested next steps view
    */
-  suggestedStepPublishRepository: number
+  readonly suggestedStepPublishRepository: number
 
   /**
    * The number of times the user has used the publish branch action branch from
    * the suggested next steps view
    */
-  suggestedStepPublishBranch: number
+  readonly suggestedStepPublishBranch: number
 
   /**
    * The number of times the user has used the Create PR suggestion
    * in the suggested next steps view
    */
-  suggestedStepCreatePR: number
+  readonly suggestedStepCreatePR: number
 }
 
 export class StatsDatabase extends Dexie {
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -187,11 +187,41 @@ export interface IDailyMeasures {
 
   /** The number of times the user opens the "Rebase current branch" menu item */
   readonly rebaseCurrentBranchMenuCount: number
+
+  /**
+   * The number of times the user has opened their external editor from the
+   * suggested next steps view
+   */
   suggestedStepOpenInExternalEditor: number
+
+  /**
+   * The number of times the user has opened their repository in Finder/Explorer
+   * from the suggested next steps view
+   */
   suggestedStepOpenWorkingDirectory: number
+
+  /**
+   * The number of times the user has opened their repository on GitHub from the
+   * suggested next steps view
+   */
   suggestedStepViewOnGitHub: number
+
+  /**
+   * The number of times the user has used the publish repository action from the
+   * suggested next steps view
+   */
   suggestedStepPublishRepository: number
+
+  /**
+   * The number of times the user has used the publish branch action branch from
+   * the suggested next steps view
+   */
   suggestedStepPublishBranch: number
+
+  /**
+   * The number of times the user has used the Create PR suggestion
+   * in the suggested next steps view
+   */
   suggestedStepCreatePR: number
 }
 
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -187,6 +187,12 @@ export interface IDailyMeasures {
 
   /** The number of times the user opens the "Rebase current branch" menu item */
   readonly rebaseCurrentBranchMenuCount: number
+  suggestedStepOpenInExternalEditor: number
+  suggestedStepOpenWorkingDirectory: number
+  suggestedStepViewOnGitHub: number
+  suggestedStepPublishRepository: number
+  suggestedStepPublishBranch: number
+  suggestedStepCreatePR: number
 }
 
 export class StatsDatabase extends Dexie {
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -226,7 +226,7 @@ export interface IDailyMeasures {
    * will increment whereas if a PR is created from the menu or from
    * a keyboard shortcut only `createPullRequestCount` will increment.
    */
-  readonly suggestedStepCreatePR: number
+  readonly suggestedStepCreatePullRequest: number
 }
 
 export class StatsDatabase extends Dexie {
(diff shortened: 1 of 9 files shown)


app/src/ui/app.tsx
```diff
@@ -324,21 +324,17 @@ export class App extends React.Component<IAppProps, IAppState> {
         return this.props.dispatcher.showPopup({ type: PopupType.Preferences })
       case 'open-working-directory':
         return this.openCurrentRepositoryWorkingDirectory()
-      case 'update-branch': {
```
niik: The changes in this file are all stylistic in nature. We don't use curly brackets in switch statements anywhere else in the app (that I know of) so this change is for consistency.

app/src/ui/lib/button.tsx
```diff
@@ -139,12 +139,12 @@ export class Button extends React.Component<IButtonProps, {}> {
...
-      event.preventDefault()
-    }
-
     if (this.props.onClick) {
       this.props.onClick(event)
     }
+
+    if (this.props.type !== 'submit') {
+      event.preventDefault()
+    }
```
niik: This change is necessary in order for the menu backed blankslate action to determine whether consumers of its `onClick` event wants to prevent launching the menu action.

I've gone through all the places in the app where we rely on the `defaultPrevented` property and none of them are related to the `Button` component so this feels like a safe change while at the same time following the conventions of other components.

app/src/ui/app.tsx
shiftkey: 👍 to tidying this up - these are mostly unnecessary...

> We don't use curly brackets in switch statements anywhere else in the app (that I know of)

For extra context, I know of at least these two places where we've needed to do this:

https://github.com/desktop/desktop/blob/82c188d68a83330a08feacec782e32fee3eed5a4/app/src/ui/app.tsx#L1560-L1586

https://github.com/desktop/desktop/blob/82c188d68a83330a08feacec782e32fee3eed5a4/app/src/ui/app.tsx#L1612-L1659

This is because local variables inside a `case` share the same scope. If you remove the curly brackets here you'll see the error that `tsc` complains about:

app/src/lib/stats/stats-database.ts
```diff
@@ -187,6 +187,46 @@ export interface IDailyMeasures {
...
+
+  /**
+   * The number of times the user has used the Create PR suggestion
+   * in the suggested next steps view. Note that this number is a
+   * subset of `createPullRequestCount`. I.e. if the Create PR suggestion
+   * is invoked both `suggestedStepCreatePR` and `createPullRequestCount`
+   * will increment whereas if a PR is created from the menu or from
+   * a keyboard shortcut only `createPullRequestCount` will increment.
+   */
+  readonly suggestedStepCreatePR: number
```
shiftkey: :art: I'm not if we have strong feels for avoiding abbreviations, but this would help it align better with `createPullRequestCount`:

```suggestion
  readonly suggestedStepCreatePullRequest: number
```

But we do have some existing keys that don't follow this pattern:

 - `prBranchCheckouts`
 - `repoWithIndicatorClicked`
 - `repoWithoutIndicatorClicked`

FYIL I haven't made the changes to the other places where this is used as I wanted to flag it for discussion

app/src/lib/stats/stats-database.ts
```diff
@@ -187,6 +187,46 @@ export interface IDailyMeasures {
 
   /** The number of times the user opens the "Rebase current branch" menu item */
   readonly rebaseCurrentBranchMenuCount: number
+
+  /**
+   * The number of times the user has opened their external editor from the
+   * suggested next steps view
+   */
+  readonly suggestedStepOpenInExternalEditor: number
```
shiftkey: :art: another consistency thought here is that we've used `Count` as a suffix in places for numeric metric. Should we stick to that here too?

```suggestion
  readonly suggestedStepOpenInExternalEditorCount: number
```

We've been less consistent with this pattern in the past. All of these are could be considered count-like metrics but aren't named as such:

 - `branchComparisons`
 - `coAuthoredCommits`
 - `commits`
 - `defaultBranchComparisons`
 - `divergingBranchBannerDismissal`
 - `divergingBranchBannerDisplayed`
 - `divergingBranchBannerInfluencedMerge`
 - `divergingBranchBannerInitatedMerge`
 - `divergingBranchBannerInitiatedCompare`,
 - `dotcomCommits`
 - `enterpriseCommits`
 - `errorWhenSwitchingBranchesWithUncommmittedChanges`
 - `mergesInitiatedFromComparison`
 - `partialCommits`
 - `prBranchCheckouts`
 - `repoWithIndicatorClicked`
  - `repoWithoutIndicatorClicked`
  - `stashEntriesCreatedOutsideDesktop`
  - `unattributedCommits`

app/src/lib/stats/stats-store.ts
```diff
@@ -1006,6 +1012,68 @@ export class StatsStore implements IStatsStore {
...
+  public async recordSuggestedStepOpenInExternalEditor(): Promise<void> {
+    return this.updateDailyMeasures(m => ({
+      suggestedStepOpenInExternalEditor:
+        m.suggestedStepOpenInExternalEditor + 1,
+    }))
+  }
+
+  /**
+   * Increment the number of times the user has opened their repository in
+   * Finder/Explorerfrom the suggested next steps view
```
shiftkey:
```suggestion
   * Finder/Explorer from the suggested next steps view
```

app/src/lib/stats/stats-store.ts
```diff
@@ -1006,6 +1012,68 @@ export class StatsStore implements IStatsStore {
     }))
   }
 
+  /**
+   * Increment the number of times the user has opened their external editor
+   * from the suggested next steps view
+   */
+  public async recordSuggestedStepOpenInExternalEditor(): Promise<void> {
```
shiftkey: None of these new functions use `await` inside the body, so the `async` can be omitted safely:

```suggestion
  public recordSuggestedStepOpenInExternalEditor(): Promise<void> {
```

app/src/lib/stats/stats-database.ts
niik: Honestly I feel like the `Count` suffix is redundant here. Almost all of our metrics are counts of one form or another. I think the pattern of suffixing with count stems from back when we used Graphite and had naming conventions that altered the behavior of how the metrics were persisted. Given that we don't have any consistency here and that we're unlikely to achieve consistency given the complexity of changing these I'm personally fine with leaving the suffix out but if you're hesitant on merging this PR without it I'll make the change.

app/src/lib/stats/stats-database.ts
niik: Aligned in 4149ad7ac

app/src/lib/stats/stats-database.ts
shiftkey: Nah, this isn't a blocker. Happy to leave it as-is.

niik:
Ready for another look

nerdneha:
@niik and I talked about the metrics and are 👍 on differentiating button vs. shortcut for creating a PR only.

shiftkey:
> `*` Note that although `suggestedStepPublishRepository` does increment when the `publish` button is clicked, the user can still cancel and not actually publish. Not sure if this use case matters.

Personally I'm fine with this. I think it's still interesting for us to see how often users gravitate towards that button, even if they don't complete the flow. I'll defer to @niik and @nerdneha here about whether there's interest in a more accurate metric about seeing users complete the publish flow.

niik:
> I'll defer to @niik and @nerdneha here about whether there's interest in a more accurate metric about seeing users complete the publish flow.

The same thing can be said about the create pr metrics. We don’t know if the user actually completes the PR flow. Such data is definitely interesting but for not for this PR. We’re looking for relative use between suggested next steps as well as success indicator for create PR specifically since that’s a new feature.

nerdneha:
lol @niik typed exactly what I was going to say. I think a great future step is seeing the "funnel" for all of these actions and whether they're actually completed but it isn't necessary for this iteration.

To further back up what @niik is saying, the problem we're trying to address is that people have to think about the next step they want to do and then find the button/menu item or remember the shortcut. The solution is to make those next steps more discoverable and make it such that the user is more likely to continue the workflow. 

So, these metrics are to make sure that process is more likely to be continued and then eventually we can make sure that the next steps are completed for each workflow we provide a next step for. 

Based on all of us agreeing, i think this is good to merge today!

shiftkey:
> Based on all of us agreeing, i think this is good to merge today!



Aiming to merge this in with the next 30mins

outofambit:
thanks everyone! i’ll plan to merge this in 15 minutes unless someone objects before then. :)

outofambit:
i’ll let @shiftkey do the honors since he called dibs!

PR #7640, opened 2019-05-28
Cleans up unnecessary async usage in Stat-Store

iAmWillShepherd:
Address feedback from https://github.com/desktop/desktop/pull/7257#discussion_r288300219.


shiftkey:
I've opened #7645 to track this because this PR was pulled into #7257, but the tip commit was reverted which was the important change, as I think we still have work to do.

Issue #5197, opened 2018-07-19
update electron-winstaller to help app clean up after itself


zero77:
## Description
The Squirrel dependency is out of date by a few versions also, it should automatically update itself alongside Git Desktop.

https://github.com/Squirrel/Squirrel.Windows/blob/master/docs/getting-started/1-integrating.md#basic-squirrelwindows-update-code

## Version
1.2.6 on windows 8

shiftkey:
@zero77 Thanks for the feedback!

This is a dependency managed by [`electron-winstaller`](https://github.com/electron/windows-installer), and the latest published version contains Squirrel.Windows `1.7.9`. We're on `1.5.2` currently.

Is there anything in particular we should be interested in from recent updates to `Squirrel.Windows`? Breaking the updater is the stuff of my eternal nightmares, so I want to ensure upgrading `electron-winstaller` gets a thorough set of testing before it goes out to everyone.

shiftkey:
Just updated the title on this to indicate the actionable work, but I'd still love to hear from @zero77 about Squirrel-specific improvements that might help out Desktop.

zero77:
Sorry for the very late reply, there’s been a few changes from 1.5.2 to 1.7.9. so, from memory I can only suggest one change currently.

The newer version cleans logs and cleans up after old versions that are left behind in AppData\Local\GitHubDesktop, because, this can quickly become filled up will unnecessary leftover files that take up disk space over time.

I can also look over the change log to find some of the major changes.

shiftkey:
#6486 is an interesting error about the upgraded app being in an invalid state on-disk that might be related to some `Squirrel.Windows` work that we should look at while upgrading this dependency.

shiftkey:
#7084 is an interesting problem about the start menu shortcut going missing between updates. We should investigate after the upgrade whether this can still be reproduced.

steveward:
#8204 was another request for updating Squirrel to help reduce leftover files after updates occur.

PR #7646, opened 2019-05-29
1.7.1-beta1

shiftkey:
## Overview

Last beta before we ship 1.7.1 next week. I hope 🙏 .

 - [x] approval for release notes
 - [x] publish to beta

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.7.0-beta10",
+  "version": "1.7.1-beta1",
   "main": "./main.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


app/package.json
```diff
@@ -3,7 +3,7 @@
   "productName": "GitHub Desktop",
   "bundleID": "com.github.GitHubClient",
   "companyName": "GitHub, Inc.",
-  "version": "1.7.0-beta9",
+  "version": "1.7.1-beta1",
```
shiftkey: We had to bump the build version here to `1.7.1` because of how Squirrel.Windows currently handles prerelease builds which are later than `-beta9`:

 - This has been fixed when generating releases in https://github.com/anaisbetts/NuGet/pull/1 and https://github.com/Squirrel/Squirrel.Windows/pull/1353 which are available in Squirrel.Windows 1.9.0
 - The outstanding pull request https://github.com/Squirrel/Squirrel.Windows/pull/1492 to address this in another part of Squirrel.Windows needs to be reviewed and published in a future update

We're looking into upgrading this dependency in https://github.com/desktop/desktop/issues/5197 and https://github.com/desktop/desktop/pull/7647, but we need to be mindful of these prereleases until we can get this fixed upstream...

PR #7654, opened 2019-05-30
Fix unmounted component set state

### LINKED ISSUE #5071: address unmounted component issue in CommitMessage
Found this error in the React Dev Tools while investigating #5040 and wanted to ensure it was captured and addressed:

```
Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in CommitMessage (created by ChangesList)
    in div (created by ChangesList)
    in ChangesList (created by ChangesSidebar)
    in div (created by ChangesSidebar)
    in ChangesSidebar (created by RepositoryView)
```

As the warning suggests, I think we need to detect when the `CommitMessage` component has been unmounted and avoid calling `setState` - we do this in a few places in `receiveProps`, which I think is the most likely culprit:

https://github.com/desktop/desktop/blob/1c443bcf02c59dbf0833975211c9cffaaaa489ea/app/src/ui/changes/commit-message.tsx#L125-L133

ADustyOldMuffin:
## Overview

**Closes #5071**

## Description

Inside of commit-message React was throwing an error stating that setState had been called after the component had been unmounted. @shiftkey suspected that this was happening inside of `receiveProps` which has since been changed to `componentWillReceiveProps`.

I have added a check `_isMounted` that is set to false to begin, and then set when the component mounts and unmounts. I did this check since `isMounted()` has been depreciated, when I looked for documentation on it this was the suggested work around.

I also added the check to a `setState` call that was suspected to throw the error and I have since not been able to reproduce the error.

I believe this fixes the issue as I have not been able to find any other places where it occurs, but I welcome others to test it out as it's very possible I missed something.

## Release notes

Notes: no-notes (I'm not sure if you'd mention this in notes to end users or not)


app/src/ui/changes/commit-message.tsx
```diff
@@ -93,6 +93,7 @@ export class CommitMessage extends React.Component<
   ICommitMessageProps,
   ICommitMessageState
 > {
+  private _isMounted: boolean = false
```
outofambit: Hi @ADustyOldMuffin and thanks for looking into this! 

This seems like a reasonable solution, I'd just like to make sure we capture this context for future readers of this code. Could we add a code comment above this with a little context for this and a link to https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html as a reference (i'm assuming that's the article you're referring to in the PR body)? Thank you!

ADustyOldMuffin: Certainly, I added a comment stating that `_isMounted` was being used instead of `this.isMounted()` because the latter is depreciated and then left a link for references as to why.

outofambit:
hey @ADustyOldMuffin thanks for this pull request! one thing i wanted to verify, were you able to reproduce this bug before these changes? i'm still looking into it, but i'm not sure that react would trigger `componentWillReceiveProps` after the component had been unmounted... 🤔

@shiftkey is there any context we're missing re: #5071?

shiftkey:
@outofambit @ADustyOldMuffin I was under the impression that I'd be able to reproduce this fairly easily, but I'm not able to reproduce this on 33aa131428 in development mode - so maybe the underlying issue has been fixed indirectly?

Unless you're able to reproduce this @ADustyOldMuffin I'd suggest we close this out as unable to reproduce, as I dislike introducing this sort of workaround for a problem we're not able to see.

ADustyOldMuffin:
Whoops! Lesson for the future, no I forgot to try and reproduce before I worked on it and I can't get it either. I'll close it.

shiftkey:
@ADustyOldMuffin no problem, and I think that's also on the maintainers to revisit old bugs like this and confirm they're still happening!

ADustyOldMuffin:
@shiftkey Well either way, one issue down. Back to backlog hunting with me!

PR #7663, opened 2019-06-01
Remove url and replaces with text links

leelena:
## Overview

Makes minor changes so that following the steps is intuitive

## Release notes

Notes: tidy up links in testing document


leelena:
testing webhooks

PR #7666, opened 2019-06-01
override port for development build using environment variable

### LINKED ISSUE #4675: starting in dev mode doesn't respect PORT value
Stumbled upon this as part of #3675 but it's actually an issue that's been around forever.

In theory, you can override a different port when launching the app in dev mode:

```
$ PORT=3001 yarn start
```

If you don't specify this environment variable, `3000` is used. That's this value:

https://github.com/desktop/desktop/blob/f7297021482886ceec53dfc414b8235525adc223/script/start#L33

However we also hard-code this value in [`webpack.development.js`](https://github.com/desktop/desktop/blob/master/app/webpack.development.js):

https://github.com/desktop/desktop/blob/f7297021482886ceec53dfc414b8235525adc223/app/webpack.development.js#L17-L26

Here's a rough patch to update this value, to get things started:

```diff
diff --git a/app/webpack.development.js b/app/webpack.development.js
index aab57ac31..1bd701afd 100644
--- a/app/webpack.development.js
+++ b/app/webpack.development.js
@@ -14,15 +14,29 @@ const askPassConfig = merge({}, common.askPass, config)
 const cliConfig = merge({}, common.cli, config)
 const highlighterConfig = merge({}, common.highlighter, config)
 
+function getPortOrDefault() {
+  const port = process.env.PORT
+  if (port != null) {
+    const result = parseInt(port)
+    if (isNaN(result)) {
+      throw new Error(`Unable to parse '${port}' into valid number`)
+    }
+    return result
+  }
+
+  return 3000
+}
+
+const port = getPortOrDefault()
+const rendererEntryPoint = `webpack-hot-middleware/client?path=http://localhost:${port}/__webpack_hmr`
+const publicPath = `http://localhost:${port}/build/`
+
 const rendererConfig = merge({}, common.renderer, config, {
   entry: {
-    renderer: [
-      'webpack-hot-middleware/client?path=http://localhost:3000/__webpack_hmr',
-      common.renderer.entry.renderer,
-    ],
+    renderer: [rendererEntryPoint, common.renderer.entry.renderer],
   },
   output: {
-    publicPath: 'http://localhost:3000/build/',
+    publicPath,
   },
   module: {
     rules: [
```

There's two things in here for a contributor to help improve the development experience:

First, confirm the fix above addresses the issue - these steps should launch the app:

```shellsession
$ PORT=3001 yarn build:dev
$ PORT=3001 yarn start
```

Second, choosing a different port when `yarn start` should also work:

```shellsession
$ PORT=3001 yarn build:dev
$ PORT=3002 yarn start
```

Currently this just shows the developer tools without launching the Desktop UI - this might need some familiarity with how we're using `webpack-dev-middleware` here to ensure we're correctly building the renderer config:

https://github.com/desktop/desktop/blob/f7297021482886ceec53dfc414b8235525adc223/script/start#L35-L40

sr98bj:
Fixed the issue with the hard-coded port, now you can pass the port via the command line argument.

## Overview
Changed the webpack.development.ts to remove the hard-coded port

**Closes #4675**

## Description
Followed the instructions in the issue but made one update - did not need this change afterall:

renderer: [rendererEntryPoint, common.renderer.entry.renderer],


app/src/ui/blank-slate/blank-slate.tsx
```diff
@@ -103,7 +103,7 @@ interface IBlankSlateState {
 export class BlankSlateView extends React.Component<
   IBlankSlateProps,
   IBlankSlateState
-> {
+  > {
```
shiftkey: @sr98bj thanks for starting this off! It looks like there's some formatting changes that have been introduced by the editor. To tidy these up, you can run `yarn lint:fix` locally and commit the changes.

sr98bj: Okay, Ill do that now and resubmit

shiftkey: @sr98bj you can just make new commits to the branch and push those to your fork - that will refresh this pull request without needing to submit a new one...

shiftkey:
Dropping `ready-for-review` due to inacitivty. @sr98bj when you're able to make the changes please add a comment in here and we can move this back in the review queue.

ADustyOldMuffin:
Hey @shiftkey if @sr98bj doesn't have time, I'd be interested in also helping this cross the finish line since it's basically there. Not sure what the policy for picking up old PR's are.

shiftkey:
@ADustyOldMuffin I'd like to give @sr98bj a chance to wrap this up, as I think it's very close to landing and just needs a bit of tidy up. I'll check back in some time next week to see if it's moved.

shiftkey:
@sr98bj @ADustyOldMuffin I pushed a commit to this branch to address the lint issues - no need to start a new PR or hand it over to a new contributor. I'm going to leave this for a different reviewer to pick up when they have some bandwidth.

shiftkey:
@iAmWillShepherd that's a good spot about how `PORT` needs to be the same between `build:dev` and `start`. I didn't think that'd get baked into the output from webpack but I guess it does.

The case I was focused on addressing in the original issue is a hypothetical where a different app also needs port `3000`. In that case, I'd set the `PORT` environment variable once for the shell session, so all commands use the same value, rather than pre-pend it to each command. Does that make sense?

iAmWillShepherd:
> In that case, I'd set the PORT environment variable once for the shell session, so all commands use the same value, rather than pre-pend it to each command. Does that make sense?

That does make sense and is in line with my initial expectations of how things would work. 

It was this bit (from https://github.com/desktop/desktop/issues/4675#issue-322672951) that changed my assumption of how this should work.

> Second, choosing a different port when yarn start should also work:
> 
> ```
> $ PORT=3001 yarn build:dev
> $ PORT=3002 yarn start
> ```

Since this change does work when `PORT` is set once in the session, I'll go ahead and approve this PR.

shiftkey:
@iAmWillShepherd ah, now I understand. I'm happy to leave that case unsupported - it looks like we may need to change our Webpack setup to support that case, and I don't think it's worth revisiting unless someone feels this pain.

PR #7670, opened 2019-06-02
Upgrade node-sass to 4.12.0

### LINKED ISSUE #7498: upgrade node-sass so we don't need to block Node 12 usage
`node-sass` had an update to `v4.12.0` recently which enables it to be used within Node 12: https://www.npmjs.com/package/node-sass/v/4.12.0

We use it here in Desktop to convert out `SCSS` files into CSS as part of the app bundle using [`sass-loader`](https://github.com/webpack-contrib/sass-loader):

https://github.com/desktop/desktop/blob/6974d6613a950d7bc2414e2f8699edea3d176818/package.json#L94

If we can verify the app builds and styles aren't affected by the version update, we can then kill off the workaround in #7399.

say25:
## Overview

**Closes #7498**

## Description

- Upgrades node-sass to 4.12.0, to a version that supports Node 12
- Reverts work around that blocks node 12

## Release notes

Notes: No notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">= 8.11 < 12.0.0",
+    "node": ">= 8.11",
     "yarn": ">= 1.9"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


shiftkey:
Tested and verified the change in macOS, Windows and Linux  🎉

PR #7680, opened 2019-06-04
[WIP] Fix wrapping in history list

### LINKED ISSUE #7679: Moving the selection up (in history) should not loop
Moving the selection up (in history), when on the topmost commit, should not loop around to the last commit.

Before pressing up:

> “When I run into a merge conflict, I don’t know where to go to resolve it. Anyone who works off of multiple branches will likely run into this problem at least occasionally.”

A clear and concise description of one or more solutions you think might solve the problem. Please include any considered drawbacks or tradeoffs, and how users might use your solution(s). Screenshots or mockups are helpful here!

sandstrom:
Don't wrap the history list.

This PR is WIP.

Need to adjust the code so that we only disable wrapping for history? (unless we want to disable it for all lists, I personally wouldn't mind.

Fixes #7679

### DIFF
--- a/app/src/ui/lib/list/list.tsx
+++ b/app/src/ui/lib/list/list.tsx
@@ -526,7 +526,7 @@ export class List extends React.Component<IListProps, IListState> {
 
     const newRow = findNextSelectableRow(
       this.props.rowCount,
-      { direction, row: lastSelection },
+      { direction, row: lastSelection, wrap: false },
       this.canSelectRow
     )
 


shiftkey:
> This PR is WIP.

@sandstrom i'm going to add a `WIP` prefix to the PR body, but feel free to remove that and/or comment here when you feel it's ready for review

sandstrom:
@shiftkey I'm not going to run with this. I merely added it as a response to #7679 where a colleague of yours said this was tricky to do.

I think it makes more sense for someone in Github Inc to continue working on this.

I don't know the codebase + I'm not paid to work on it 😄

shiftkey:
>@shiftkey I'm not going to run with this, I merely added it as a response to #7679 where a colleague of yours said this was tricky to do.

I had a look at this work, and while it does address the issue for the History list I encountered various  issues in the Branches list (you can lose focus on the list while just using keyboard navigation) as well as the Changes tab (pressing down to navigate the list eventually causes the list to jump and lose focus).

I don't believe those changes are related to this PR, but given how much we use lists in the app I want a more thorough test plan in place to ensure any fix for the reported issue doesn't regress other lists in the app.

@sandstrom thanks for taking a shot at this. I'm going to close this out because I don't have the bandwidth to pick this up currently, but we have #7679 to track the issue and discuss things further.

Issue #7643, opened 2019-05-29
detect and warn about unreachable code using eslint


shiftkey:
**Please describe the problem you think should be solved**

A small change to the diff of `app/src/ui/changes/no-changes.tsx` in https://github.com/desktop/desktop/commit/06b9016636f2cec67084700c76fbea611759e933 caught my eye, as it didn't look relevant to the changes in #7257.

<img width="495" alt="Screen Shot 2019-05-29 at 10 39 43 AM" src="https://user-images.githubusercontent.com/359239/58561810-63598c00-81fe-11e9-8d2d-c6a6fffd0bfe.png">

Turns out this was some unreachable code, and could be safely removed.

**[Optional] Do you have any potential solutions in mind?**

To catch any other cases like this, I'd recommend enabling [`no-unreachable`](https://eslint.org/docs/rules/no-unreachable) in our `.eslintrc.yml` file and verifying we have no more issues like this.

Rexogamer:
do you want this as a warning or an error?

Issue #7627, opened 2019-05-24
define a consistent import order for modules whenever possible


shiftkey:
**Please describe the problem you think should be solved**

Something that's been fermenting in my head for a while is how the tooling we use gives us the convenience to import functionality easily, but this tooling always adds it to the end of the list.

A great example of this is `AppStore`, where imports related to new features just accrue at the bottom of the list:

https://github.com/desktop/desktop/blob/cc6e4828893bd3b90ced8bbc98dd1fbc686bf7e4/app/src/lib/stores/app-store.ts#L1-L239

It's not easy to navigate the things this module requires, and given this is one of the more complex parts of the app it feels valuable here to have things better organized whenever we can.

**[Optional] Do you have any potential solutions in mind?**

I think we could leverage the [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import) plugin here, which has an `order` plugin to help enforce some consistency. 

It recommends this order:

```ts
// 1. node "builtin" modules
import fs from 'fs';
import path from 'path';
// 2. "external" modules
import _ from 'lodash';
import chalk from 'chalk';
// 3. "internal" modules
// (if you have configured your path or webpack to handle your internal paths differently)
import foo from 'src/foo';
// 4. modules from a "parent" directory
import foo from '../foo';
import qux from '../../foo/qux';
// 5. "sibling" modules from the same or a sibling's directory
import bar from './bar';
import baz from './bar/baz';
// 6. "index" of the current directory
import main from './';
```

By leveraging eslint we should be able to auto-fix most of the issues, but this also lets us disable it for specific cases where the import order is important and should not be changed, like inside `CodeMirrorHost`:

https://github.com/desktop/desktop/blob/cc6e4828893bd3b90ced8bbc98dd1fbc686bf7e4/app/src/ui/diff/code-mirror-host.tsx#L1-L20

PR #6912, opened 2019-02-21
Convert the custom TSLint rules to ESLint

j-f1:
## Overview

**Closes #6889**

## Description

- [x] Convert the ESLint rule to TypeScript
- Port:
  - [x] `button-group-order`
  - [ ] `react-no-unbound-dispatcher-props`
  - [ ] `react-proper-lifecycle-methods`
  - [x] `react-readonly-props-and-state`
- [ ] Clean up the mess in `eslint-rules/types.ts` (ref https://github.com/typescript-eslint/typescript-eslint/issues/63) (mostly fixed)
- [ ] Update Prettier to fix the output in CI

## Release notes

Notes: no-notes

PR #7681, opened 2019-06-04
update some eslint dependencies

shiftkey:
## Overview

I'd like to dust off #6912 at some point, but before that I'd like to address some low hanging fruit along the way around reducing our usage of `tslint` in the project, and moving more of it into our `eslint` setup.

This also touches on #7627 and #7643, but they are further down my list of priorities.

## Description

This is the first of a series of small cleanup PRs to move us along the path:

 - bring our `eslint` tooling up to date
 - port some included `tslint` rules over to their `eslint` equivalents (where appropriate)
 - replace some `tslint` rules that don't have an equivalent in `eslint` nor are worth keeping around (where appropriate)
 
## Release notes

Notes: no-notes


.eslintrc.yml
```diff
@@ -6,6 +6,10 @@ plugins:
   - react
   - json
 
+settings:
+  react:
+    version: '16.3'
```
shiftkey: This is a new feature of `eslint-plugin-react` and it's autodetection doesn't work because we use `app/package.json` to install react, so this needs to be hard-coded for the moment.

.eslintrc.yml
```diff
@@ -36,8 +40,6 @@ rules:
   # this rule now works but generates a lot of issues with the codebase
   # '@typescript-eslint/member-ordering': error
 
-  '@typescript-eslint/type-annotation-spacing': error
```
shiftkey: The update to `eslint-config-prettier` now indicates that this rule is unnecessary (because `prettier` handles this formatting).

ELOHIMSUPREMES:
Let us began

PR #7693, opened 2019-06-05
Additional debug data in crash reports

niik:
## Overview

We're seeing a number of crashes related to the most recent releases (2.x) and have realized we lack some key information about the state that the application is in which makes troubleshooting (and ultimately delivering a fix for affected users) slower than it needs to be.

## Description

This PR adds a number of fields to our existing crash reports which will hopefully help us troubleshoot rare crashes with no clear reproduction steps more easily.

The newly added fields are as follows

| Field | Explanation |
|-----|-----|
| time | The current time of the environment (including the time zone offset) |
| uptime | The number of seconds the application had been running when the error was encountered |
| currentBanner | The type of banner (if any) that the app was showing when the error was encountered |
| currentPopup | The type of popup (if any) that the app was showing when the error was encountered |
| selectedState | The current repository selection type, i.e. repository, missing repository, cloning repository (if any) |
| selectedRepositorySection | The selected tab in the repository view, i.e. changes or history |
| currentFoldout | The type of foldout (application drop down such as the repository list or branches list) (if any) that the app was showing when the error was encountered |
| inWelcomeFlow | Whether the app was in the welcome flow when the error was encountered |
| windowZoomFactor | The current zoom factor (if not the default zoom level) |
| activeAppError | The number of application errors currently shown in the app (if any) |
| lastUnhandledRejection | The last unhandled promise rejection seen in the app (see comments in code) |
| lastUnhandledRejectionTime | Time when the last unhandled promise rejection occurred |
| repositoryCount | The number of repositories tracked in the app at the time |
| windowState | The window state (minimized, maximized, etc) |
| accounts | The *number* of accounts currently signed in |
| automaticallySwitchTheme | Whether the app automatically switches themes based on the system configuration (macOS only) |

### DIFF
--- a/app/src/ui/index.tsx
+++ b/app/src/ui/index.tsx
@@ -35,7 +35,7 @@ import {
 } from '../lib/stores'
 import { GitHubUserDatabase } from '../lib/databases'
 import { URLActionType } from '../lib/parse-app-url'
-import { SelectionType } from '../lib/app-state'
+import { SelectionType, IAppState } from '../lib/app-state'
 import { StatsDatabase, StatsStore } from '../lib/stats'
 import {
   IssuesDatabase,
@@ -89,6 +89,8 @@ if (!process.env.TEST_ENV) {
   require('../../styles/desktop.scss')
 }
 
+let currentState: IAppState | null = null
+
 process.once('uncaughtException', (error: Error) => {
   error = withSourceMappedStack(error)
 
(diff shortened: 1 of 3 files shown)


app/src/ui/index.tsx
```diff
@@ -99,15 +103,94 @@ process.once('uncaughtException', (error: Error) => {
...
+    const extra: Record<string, string> = {
       osVersion: getOS(),
       guid: getGUID(),
-    })
+    }
+
+    try {
+      if (currentState) {
+        if (currentState.currentBanner !== null) {
+          extra.currentBanner = currentState.currentBanner.type
```
shiftkey: :art: Should we stringify this value like the others?

```suggestion
          extra.currentBanner = `${currentState.currentBanner.type}`
```

iAmWillShepherd: Won't these `type` properties return the enum number and not the human readable value?

niik: The currentBanner enum type is a string but the other are numbers. We'll have to do some work to map it ourselves. We can talk about mapping these to human readable strings in the future

iAmWillShepherd (approved):
My one comment is really more about convenience. I think this change is a giant leap in the right direction.

shiftkey (commented):
The changes seem reasonable enough to glean more context about the crash.

Just one question/suggestion.

app/src/ui/index.tsx
niik: This value is actually already a string. Anything else would have caused a compile error since `extra` is `Record<string, string>`

iAmWillShepherd: Yea, I only through it on that line since it was the first occurrence. I assumed most of our enums would be a number instead of a human-readable string.

Also, this is very much a non-blocker for me.

shiftkey: Nevermind: https://github.com/desktop/desktop/pull/7693#discussion_r290852181

shiftkey (approved):

shiftkey (approved):

nerdneha (approved):
reviewed PR content, general skimming of code, and changelog entry ✅

PR #7699, opened 2019-06-05
Fix account identity mismatch

niik:
## Overview

This should fix the source of our recent crashes after 2.0.

## Description

The API repositories store (the store responsible for loading and caching cloneable repositories for an account) was built on reference equality. In hindsight that was probably a bad idea but fixing that holistically is too much for tonight.

What happened was that when you sign in through the welcome wizard we get an `Account` from the sign-in process which we pass to the `AccountsStore` which creates a new `Account` instance. We would then tell the `ApiRepositoriesStore` to load the initial `Account` instance that we got, thereby creating a mismatch between the truth as IAppState saw it and how `ApiRepositoriesStore` saw it.

This was exacerbated, but not caused, by us calling `ensureRepositoriesForAccount` unconditionally from `componentDidUpdate` (relying on the ApiRepositoriesStore to prevent loading already loaded accounts)

The end result of all of this was that for each time the blank slate component got updated we would trigger another reload of all repositories for the account, exponentially driving up the number of requests to the API until eventually the app buckled under the load.

On macOS, however, there was another problem. When the app window was hidden (which can only happen on macOS by closing the window) [we would synchronously emit updates from AppStore](https://github.com/desktop/desktop/blob/c2dc85f8153f78bbca84baa012fe285909d2dae3/app/src/lib/stores/app-store.ts#L503-L509) which caused us to run into [a depth limit](https://reactjs.org/docs/error-decoder.html/?invariant=185) in react which ultimately caused the crashes on macOS.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -473,18 +473,14 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   private getDotComAccount(): Account | null {
-    const state = this.props.appStore.getState()
-    const accounts = state.accounts
-    const dotComAccount = accounts.find(
+    const dotComAccount = this.state.accounts.find(
       a => a.endpoint === getDotComAPIEndpoint()
     )
     return dotComAccount || null
   }
 
   private getEnterpriseAccount(): Account | null {
-    const state = this.props.appStore.getState()
-    const accounts = state.accounts
-    const enterpriseAccount = accounts.find(
+    const enterpriseAccount = this.state.accounts.find(
       a => a.endpoint !== getDotComAPIEndpoint()
     )
     return enterpriseAccount || null
--- a/app/src/ui/blank-slate/blank-slate.tsx
+++ b/app/src/ui/blank-slate/blank-slate.tsx
@@ -142,8 +142,17 @@ export class BlankSlateView extends React.Component<
     this.ensureRepositoriesForAccount(this.getSelectedAccount())
   }
 
-  public componentDidUpdate(prevProps: IBlankSlateProps) {
-    this.ensureRepositoriesForAccount(this.getSelectedAccount())
+  public componentDidUpdate(
+    prevProps: IBlankSlateProps,
+    prevState: IBlankSlateState
+  ) {
+    if (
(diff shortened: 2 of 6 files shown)


app/src/ui/app.tsx
```diff
@@ -473,18 +473,14 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   private getDotComAccount(): Account | null {
-    const state = this.props.appStore.getState()
-    const accounts = state.accounts
-    const dotComAccount = accounts.find(
+    const dotComAccount = this.state.accounts.find(
```
niik: These changes help ensure that we're all talking about the same instances of accounts during one render cycle.

app/src/ui/blank-slate/blank-slate.tsx
```diff
@@ -142,15 +142,24 @@ export class BlankSlateView extends React.Component<
...
-    this.ensureRepositoriesForAccount(this.getSelectedAccount())
+  public componentDidUpdate(
+    prevProps: IBlankSlateProps,
+    prevState: IBlankSlateState
+  ) {
+    if (
+      prevProps.dotComAccount !== this.props.dotComAccount ||
+      prevProps.enterpriseAccount !== this.props.enterpriseAccount ||
+      prevState.selectedTab !== this.state.selectedTab
+    ) {
```
niik: These changes aren't strictly necessary for the fix but it's good practice to only do work when things actually change that affect the work.

app/src/ui/blank-slate/blank-slate.tsx
```diff
@@ -142,15 +142,24 @@ export class BlankSlateView extends React.Component<
...
+      this.ensureRepositoriesForAccount(this.getSelectedAccount())
+    }
   }
 
   private ensureRepositoriesForAccount(account: Account | null) {
     if (account !== null) {
       const accountState = this.props.apiRepositories.get(account)
 
-      if (accountState === undefined || accountState.repositories === null) {
+      if (accountState === undefined) {
```
niik: This is just stylistic, `accountState.repositories` can't be null

tierninho:
Cannot reproduce the error anymore. ✅

PR #7708, opened 2019-06-06
ensure the current window is provided when launching  showOpenDialog

### LINKED ISSUE #7636: showOpenDialog should be bound to a given BrowserWindow
## Description
If you click outside the file dialog while adding a local repo, the dialog _appears_ to dismiss and the UI is unresponsive.

## Version
* GitHub Desktop: 1.6.6
* Operating system: macOS 10.14.4

## Steps to Reproduce
1. Cmd-O to add a local repo
2. Click on _Choose…_
3. Click on the underlying window, outside of the Open dialog
4. Click on _Choose…_ again

### Expected Behavior
Really, I'd expect the Open dialog to appear as a sheet on the window. That's the correct way to indicate the modality that's used here.

But it would also be fine if clicking on the window actually dismissed the Open dialog or if the buttons on the _Add Local Repository_ modal worked.

### Actual Behavior
The Open dialog goes _behind_ the window with no indication that it's there. I actually thought that the dialog _had_ been dismissed. I quit the app several times while reproducing this because I didn't realize that I could get back to the open dialog.

shiftkey:
## Overview

**Closes #7636**

## Description

- [x] test on macOS and verify the flow works as expected
- [x] test on Windows and verify the flow works as expected
- [x] test on Linux and verify the flow works as expected

## Release notes

Notes: `[Fixed] \"Open folder\" dialogs are not bound to the current window on macOS`


shiftkey:
@tierninho the other case is "Relocate Repository" - when a repository goes missing on disk, it can run the `showOpenDialog` function if the user knows where it now is, rather than cloning again...

billygriffin:
Just so I'm clear, is the intended impact for the user here that they are just forced to make a selection (to include "Cancel") while the file chooser is active, and can't click back to the Desktop window?

shiftkey:
@billygriffin that's correct. Without this change, the user can avoid the OS dialog (which is confusing) and because they're confused, try and launch the dialog again (which does nothing, because the app is waiting on a result from the first dialog)...

shiftkey:
@iAmWillShepherd reading over what [Sheets](https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/sheets/) represent in the HIG, this feels suitable for Desktop's case. The other example I found,  [App-Modal Dialogs](https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/dialogs/), feels less relevant, based on the confusion initially reported.

>  I compared to the default text editor that ships with macOS and it behaves the same way as Desktop in its current form.

Can you elaborate on the workflow you went through? TextEdit does support opening multiple documents, so I worry we're not comparing equivalent flows...

billygriffin:
Got it, thanks @shiftkey. This seems like a good choice to me with that understanding.

iAmWillShepherd:
@shiftkey I just did CMD+O on Texted it, Logic Pro X, and Sublime Text and they all opened a standalone window, so I wasn't sure if the sheet is conventional. Turns out it is, so I'm good 👍

shiftkey:
@iAmWillShepherd yeah, programs like editors that can open multiple windows/documents are a different scenario to us. Thanks for the extra context!

PR #7765, opened 2019-06-11
pass the current state when emitting show event from BrowserWindow

### LINKED ISSUE #7750: Maximized window won't "restore down" after restarting application
## Description

The maximize button in the top-right corner of the window does nothing if the application is ran maximized.

## Version

* GitHub Desktop: 2.0.3

* Operating system: Microsoft Windows [Version 10.0.17763.437]

## Steps to Reproduce

1. Run GitHub Desktop.
2. Maximize the window.
3. Close GitHub Desktop.
4. Run GitHub Desktop.
5. Try to use the maximize button in the top-right corner of the window.
### Expected Behavior

The maximize button should actually be the "restore down" button as the window is already maximized.
### Actual Behavior

Nothing. When you click the maximize button it doesn't maximize the window (because it's already maximized), nor does it do anything else or change to the "restore down" button. The only way to get the window un-maximized is to drag the title bar away from the top of the screen or press Windows+Left/Right/Down (any direction other than up) on  your keyboard.

## Additional Information

If this can't be reproduced, I'd be happy to make a gif showing it happening in GitHub Desktop but no other application.
### Logs

[2019-06-10.desktop.production.log](https://github.com/desktop/desktop/files/3272785/2019-06-10.desktop.production.log)

shiftkey:
## Overview

**Closes #7750**

## Description

The root cause of this issue is that the main process subscribes to the `show` event from `BrowserWindow` (to indicate when the window is first displayed) to then emit a message to the renderer process so that the UI is in the right state. This has always emitted `normal` despite the fact that the app could have been maximized when it was previously closed.

This PR now checks what the window state is at the time and emits that to the app. And because this is related to the launch state of the app, I'll ensure this behaviour is correct across all the platforms.

 - [x] test on Windows in development mode
 - [x] test on Windows in production mode
 - [x] test on macOS in development mode
 - [x] test on macOS in production mode
 - [x] test on Linux in development mode
 - [x] test on Linux in production mode

## Release notes

Notes: Launching app on Windows after being maximized does not restore correct window state

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -184,7 +184,11 @@ import {
   hasSeenUsageStatsNote,
 } from '../stats'
 import { hasShownWelcomeFlow, markWelcomeFlowComplete } from '../welcome'
-import { getWindowState, WindowState } from '../window-state'
+import {
+  getWindowState,
+  WindowState,
+  windowStateChannelName,
+} from '../window-state'
 import { TypedBaseStore } from './base-store'
 import { AheadBehindUpdater } from './helpers/ahead-behind-updater'
 import { MergeResult } from '../../models/merge'
@@ -412,7 +416,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
   private wireupIpcEventHandlers(window: Electron.BrowserWindow) {
     ipcRenderer.on(
-      'window-state-changed',
+      windowStateChannelName,
       (event: Electron.IpcMessageEvent, args: any[]) => {
         this.windowState = getWindowState(window)
         this.emitUpdate()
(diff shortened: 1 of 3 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -412,9 +416,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
   private wireupIpcEventHandlers(window: Electron.BrowserWindow) {
     ipcRenderer.on(
-      'window-state-changed',
-      (event: Electron.IpcMessageEvent, args: any[]) => {
-        this.windowState = getWindowState(window)
```
shiftkey: While investigating this bug I also noted that the IPC call contains the `WindowState` we need to work with, so don't need to call this function if we can trust the value from the main process.

app/src/ui/window/window-controls.tsx
```diff
@@ -52,9 +52,9 @@ export class WindowControls extends React.Component<{}, IWindowControlState> {
 
   private onWindowStateChanged = (
     event: Electron.IpcMessageEvent,
-    args: any
+    windowState: WindowState
   ) => {
-    this.setState({ windowState: args as WindowState })
```
shiftkey: The change above can also be justified by this usage of the IPC result - I just tidied it up here to avoid the `any` usage.

PR #7782, opened 2019-06-12
restore caches for CI providers

### LINKED ISSUE #7781: re-enable eslint cache for AppVeyor and Travis
See #7771 for more information - it's not clear to me just yet what the root cause is...

shiftkey:
## Overview

**Resolves #7781**

## Release notes

Notes: no-notes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ cache:
   timeout: 600
   directories:
     - $HOME/.electron
+    - .eslintcache
     - $HOME/.cache/electron-builder
 
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ environment:
 
 cache:
   - '%USERPROFILE%\.electron'
+  - .eslintcache
   - '%LOCALAPPDATA%\Yarn\Cache\v4'
 
 branches:


shiftkey:
@outofambit oops, I meant to brain dump what I found. I'll add this to #7781 now and add more context.

shiftkey:
@outofambit I believe this will be resolved after merging because we're no longer having to worry about multiple versions of eslint trying to manage this cache. 

The file itself is (on my machine) 240KB in size, so the overhead in caching is negligible, but I think having this information available to `eslint` at build time helps to speed things up.

Without the cache:

```
$ yarn eslint
yarn run v1.15.2
$ eslint --cache --rulesdir ./eslint-rules "./eslint-rules/**/*.js" "./{script,tslint-rules}/**
/*.ts{,x}" "./app/{src,typings,test}/**/*.{j,t}s{,x}" "./changelog.json"
Done in 14.62s.
```

With the cache:

```
$ yarn eslint
yarn run v1.15.2
$ eslint --cache --rulesdir ./eslint-rules "./eslint-rules/**/*.js" "./{script,tslint-rules}/**
/*.ts{,x}" "./app/{src,typings,test}/**/*.{j,t}s{,x}" "./changelog.json"
Done in 3.30s.
```

outofambit:
thanks @shiftkey!

leaving a note here in case someone looks at this later that we didn't remove CircleCI's eslint cache so we don't need to re-add it here.

shiftkey:
> leaving a note here in case someone looks at this later that we didn't remove CircleCI's eslint cache so we don't need to re-add it here.

This is because CircleCI has a configurable strategy for what it restores from cache:

https://github.com/desktop/desktop/blob/ce61834cf5f1c1db7b479853a3491ac071ea514a/.circleci/config.yml#L19-L23

It didn't get caught out by this because changed to the root lockfile (or use of a different branch) will invalidate the cache on build.

PR #7784, opened 2019-06-12
Add key property for all dialogs to ensure transitions

outofambit:
## Overview

While investigating a crash related to dialogs, I saw that most of our dialog components are missing a `key` property, which is needed to ensure that the component isn't created/destroyed more than once during transitions.

This PR adds `key`s to all dialog components that lack them.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1420,16 +1420,23 @@ export class App extends React.Component<IAppProps, IAppState> {
       case PopupType.RemoveRepository:
         return (
           <ConfirmRemoveRepository
+            key="confirm-remove-repository"
             repository={popup.repository}
             onConfirmation={this.onConfirmRepoRemoval}
             onDismissed={this.onPopupDismissed}
           />
         )
       case PopupType.TermsAndConditions:
-        return <TermsAndConditions onDismissed={this.onPopupDismissed} />
+        return (
+          <TermsAndConditions
+            key="terms-and-conditions"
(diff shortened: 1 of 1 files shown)


tierninho:
Anything to test here? Dialogs seem fine in dev mode. Tried to use React in dev tools to verify `key` but crashes periodically.

shiftkey:
> Tried to use React in dev tools to verify `key` but crashes periodically.

@tierninho what sort of crash? 🤔

iAmWillShepherd:
> Anything to test here? Dialogs seem fine in dev mode. Tried to use React in dev tools to verify `key` but crashes periodically.

Are you referring to React dev tools crashing or the Desktop application?

tierninho:
> Are you referring to React dev tools crashing or the Desktop application?

Yes, when I drill down into some content in the React tab, it crashes every time.

> what sort of crash?

```
TypeError: Cannot convert undefined or null to object
    at Function.getOwnPropertyNames (<anonymous>)
    at Bridge._inspectResponse (chrome-extension://react-developer-tools/build/backend.js:7696:17)
    at Bridge._handleMessage (chrome-extension://react-developer-tools/build/backend.js:7592:15)
    at listener (chrome-extension://react-developer-tools/build/backend.js:92:10)

Version: 2.0.4
OS: Mac OS 10.14.5
```

iAmWillShepherd:
@tierninho 

This change isn't causing that crash. I get those crashed pretty often as well when trying to inspect state. I think it's a bug with dev tools.

PR #7793, opened 2019-06-13
add rebaseCurrentBranchMenuCount metric to list

shiftkey:
## Overview

This measure was added in https://github.com/desktop/desktop/pull/6961 but wasn't documented as part of that PR.

## Release notes

Notes: no-notes


shiftkey:
Merging this in a bit early as I'd like to open a PR that builds upon it

Issue #7645, opened 2019-05-29
cleanup unnecessary async modifier usage in codebase


shiftkey:
**Please describe the problem you think should be solved**

Because of recent work in https://github.com/desktop/desktop/pull/7257#discussion_r288300219 and other PRs around adding new metrics, we found ourselves adding a lot more functions that are marked `async` but don't use `await` inside.

**[Optional] Do you have any potential solutions in mind?**

Clean them up 🗑

PR #7794, opened 2019-06-13
drop unnecessary async modifiers from public API for StatsStore

shiftkey:
## Overview

**Related to #7645**

## Description

In advance of some metrics-related work for 2.1, I figured I'd use this as a chance to cleanup this area before I do some testing in there.

I almost wanted to also drop the `Promise<void>` return type annotations in here, as I think they add noise given how each API infers and inherits from calling `updateDailyMeasures`. If you think this could also be tidied up while we're in here, let me know!

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -732,14 +732,14 @@ export class StatsStore implements IStatsStore {
   }
 
   /** Record that user dismissed diverging branch notification */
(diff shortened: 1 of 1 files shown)


shiftkey:
I avoided assigning myself and marking this as closing #7645 because I think there's other places in the codebase where we've followed this pattern incidentally - if I can find some automated way to detect these I'll update that issue, but this was some low-hanging fruit to kick it off...

PR #7799, opened 2019-06-13
Upgrade to Electron Five (5)

### LINKED ISSUE #6888: upgrade app to use Electron 5

outofambit:
## Overview

closes #6888 

## Description

- addresses breaking API changes


yarn.lock
```diff
@@ -316,10 +316,10 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-8.0.47.tgz#968e596f91acd59069054558a00708c445ca30c2"
   integrity sha512-kOwL746WVvt/9Phf6/JgX/bsGQvbrK5iUgzyfwZNcKVFcjAUVSpF9HxevLTld2SG9aywYHOILj38arDdY1r/iQ==
 
-"@types/node@^8.0.24":
-  version "8.10.40"
-  resolved "https://registry.yarnpkg.com/@types/node/-/node-8.10.40.tgz#4314888d5cd537945d73e9ce165c04cc550144a4"
-  integrity sha512-RRSjdwz63kS4u7edIwJUn8NqKLLQ6LyqF/X4+4jp38MBT3Vwetewi2N4dgJEshLbDwNgOJXNYoOwzVZUSSLhkQ==
+"@types/node@^10.12.18":
```
shiftkey: Can we also bump our `@types/node` dependency in here to a 10.x version?

shiftkey: Actually, let's not worry about that until we decide to move the CI builds to Node 10 or later

outofambit:
@niik could you take a look at this and let me know if it seems sensible? i annotated most of the changes to help give context!

outofambit:
also @tierninho if you could do a smoke test of some sort and make sure nothing obvious is broken? the plan is to have this on beta for a while, so we'll have time to find other things too. (i looked and didn't notice anything but 🤷‍♀.)

app/src/lib/menu-item.ts
```diff
@@ -17,7 +17,43 @@ export interface IMenuItem {
    * When specified the click property will be ignored.
    * See https://electronjs.org/docs/api/menu-item#roles
    */
-  readonly role?: string
+  readonly role?:
```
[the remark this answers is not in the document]
j-f1: Is there a way to pull this directly from Electron so it doesn’t get out of date? Perhaps `Electron.MenuItem['role']`?

app/src/lib/shell.ts
```diff
@@ -57,7 +57,7 @@ async function getRawShellEnv(): Promise<string | null> {
       cleanup()
     }, 5000)
 
-    const options = {
+    const options: ChildProcess.SpawnOptions = {
```
j-f1: Is there a reason this is pulled out into a constant rather than inlined directly into the `ChildProcess.spawn` call?

outofambit: i don't think so. moved! that does look nicer, thanks!

tierninho:
In the meantime, I'm gonna do a flyover on the open Electron 5.0x bugs from the [Electron repo](https://github.com/electron/electron/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+5+label%3A5-0-x+label%3A%22bug+%3Abeetle%3A%22+) to see if any might apply to us.

PR #7803, opened 2019-06-14
add measure for rebase completion without conflicts

shiftkey:
## Overview

Following on from #7793, there was an additional case that we don't currently have insights into - completing a rebase without encountering conflicts. This PR adds that measure, and gives me an opportunity to review what's there.

## Description

- [x]  add `rebaseSuccessWithoutConflictsCount` measure
- [x] increment `rebaseSuccessWithoutConflictsCount` and `rebaseSuccessAfterConflictsCount` as part of rebase flow
- [x] test flows and ensure measures are working as expected
- [x] setup central PR to consume measure https://github.com/github/central/pull/439
- [x] setup marketing site PR to list measure https://github.com/github/desktop.github.com/pull/137

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/updates/changes-state.ts
+++ b/app/src/lib/stores/updates/changes-state.ts
@@ -233,11 +233,11 @@ function performEffectsForRebaseStateChange(
   ) {
     const previousTip = prevConflictState.originalBranchTip
 
-    if (
+    const previousTipChanged =
       previousTip !== currentTip &&
       currentBranch === prevConflictState.targetBranch
-    ) {
-      statsStore.recordRebaseSuccessAfterConflicts()
+
+    if (previousTipChanged) {
     } else {
       statsStore.recordRebaseAbortedAfterConflicts()
     }
--- a/app/src/lib/stores/updates/changes-state.ts
+++ b/app/src/lib/stores/updates/changes-state.ts
@@ -237,8 +237,7 @@ function performEffectsForRebaseStateChange(
       previousTip !== currentTip &&
       currentBranch === prevConflictState.targetBranch
 
-    if (previousTipChanged) {
-    } else {
+    if (!previousTipChanged) {
       statsStore.recordRebaseAbortedAfterConflicts()
     }
   }
--- a/app/test/unit/stores/updates/update-conflict-state-test.ts
+++ b/app/test/unit/stores/updates/update-conflict-state-test.ts
@@ -7,14 +7,19 @@ import {
   ManualConflictResolution,
   ManualConflictResolutionKind,
 } from '../../../../src/models/manual-conflict-resolution'
+import { IStatsStore } from '../../../../src/lib/stats'
 
 describe('updateConflictState', () => {
-  const statsStore = {
-    recordMergeAbortedAfterConflicts: jest.fn(),
-    recordMergeSuccessAfterConflicts: jest.fn(),
-    recordRebaseAbortedAfterConflicts: jest.fn(),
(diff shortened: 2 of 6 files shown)


app/test/unit/stores/updates/update-conflict-state-test.ts
```diff
@@ -7,14 +7,19 @@ import {
...
 describe('updateConflictState', () => {
-  const statsStore = {
-    recordMergeAbortedAfterConflicts: jest.fn(),
-    recordMergeSuccessAfterConflicts: jest.fn(),
-    recordRebaseAbortedAfterConflicts: jest.fn(),
-    recordRebaseSuccessAfterConflicts: jest.fn(),
-  }
+  let statsStore: IStatsStore
+  beforeEach(() => {
+    statsStore = {
```
shiftkey: The original change in this test indicated that `statsStore.recordRebaseAbortedAfterConflicts` had been called twice (likely by other tests), so it would fail. This change is to ensure the mocks are not reused across tests, and we could explicitly assert here that other mocks were not called, but I felt that would make the tests overly verbose.

app/src/lib/stores/updates/changes-state.ts
```diff
@@ -233,12 +233,11 @@ function performEffectsForRebaseStateChange(
   ) {
     const previousTip = prevConflictState.originalBranchTip
 
-    if (
+    const previousTipChanged =
       previousTip !== currentTip &&
       currentBranch === prevConflictState.targetBranch
-    ) {
-      statsStore.recordRebaseSuccessAfterConflicts()
```
shiftkey: This check was originally intended to catch rebases where the user completes the rebase outside the app, but I don't think this case is interesting enough to blend it with how the user works within the app flow.

cc @billygriffin for :eyes: and 👍 that we don't care about this case

billygriffin: I think that makes sense.

docs/process/metrics.md
```diff
@@ -74,7 +74,8 @@ These are general metrics about feature usage and specific feature behaviors. Th
 | `rebaseConflictsDialogDismissalCount` | The number of times the rebase conflicts dialog is dismissed. | To understand how users respond when they encounter rebase conflicts. |
 | `rebaseConflictsDialogReopenedCount` | The number of times the rebase conflicts dialog is reopened. |To understand how users respond when they encounter rebase conflicts. |
 | `rebaseCurrentBranchMenuCount` | The number of times the `Branch -> Rebase Current Branch` menu item is used. | To understand how often users start a rebase in the app. |
-| `rebaseSuccessAfterConflictsCount` |  The number of times a successful rebase is detected in the app. | To understand whether users are able to complete the rebase conflicts flow. |
+| `rebaseSuccessAfterConflictsCount` |  The number of times a successful rebase is made in the app after the user resolved conflicts. | To understand whether users are able to complete the rebase conflicts flow. |
+| `rebaseSuccessWithoutConflictsCount` | The number of times a successful rebase is made in the app without the user needing to resolve conflicts. | To understand whether users are able to complete the rebase conflicts flow. |
```
billygriffin:
```suggestion
| `rebaseSuccessWithoutConflictsCount` | The number of times a successful rebase is made in the app without the user needing to resolve conflicts. | To understand how frequently people are rebasing compared to merging. |
```

app/src/lib/stores/updates/changes-state.ts
outofambit: yeah i think if we wanted to track this metric, it should be its own measure

PR #7809, opened 2019-06-14
Fix history compare branch selection

### LINKED ISSUE #7802: unable to scroll branch list in compare tab using keyboard
## Description

The compare tab should let you switch between the filter text and the list of filtered branches, but the keyboard navigation doesn't appear to work as expected currently.

## Version
* GitHub Desktop: 2.0.4
* Operating system: macOS and Windows

## Steps to Reproduce
1. Switch to the History tab for a repository
2. Use the <kbd>Tab</kbd> key to switch focus to the Branches list
3. Press <kbd>Down</kbd> to try and select a branch

### Expected Behavior
Keyboard moves to branches list, and user can scroll through the list of branches

### Actual Behavior
Filter textbox retains focus, and scrolling appears to be stuck at the top

### Additional information

@ADustyOldMuffin provided a GIF of the interaction in an earlier investigation of this area:

ADustyOldMuffin:
## Overview

**Closes #7802**

## Description

- A user would be unable to scroll through branches to compare with in the history tab with the arrow keys. It will also handle reaching the bottom of the list and going to the top as well as going from the top to the bottom. 

This just refactors several methods used to select the first item in a list to select next item which will handle all cases.

CompareNavigation

## Release notes

Notes: You can now navigate through the compare branches using the up and down arrow keys.


tierninho:
@ADustyOldMuffin Thanks for taking the time to fix this. I spotted some bugs but both were **pre-existing** so not sure if you need/want to fix them.

1) Missing tab over
- Make sure you have a long history... long enough to scroll off the page. 
- Select the last history item, scroll back up to the top
- Click into the Select branch filter
- Tab tab and the commit that was selected out of view will be skipped over to the diff

2) Cannot select compare branch
- Click into the Select branch filter
- Press down arrow key to get to compare branches
- Press `Enter` to select one
- You are taken to History list not the compare branch

ADustyOldMuffin:
@tierninho I can actually fix 2 as I thought it was a bit odd, but it existed before so I left it. I think that specifically is related to this in a way as it's the same function. The 1st one I'm not sure I understand.

ADustyOldMuffin:
So there was two ways to do it,

1. Check if filter is empty and selectedBranch is not null

2. Just get rid of enter to exit entirely

I went with the 2nd option as I think it's more user friendly as I've never known a search filter where enter is used to exit the search/filter. I'm also down for debate on if the enter to exit should stay.

app/src/ui/history/compare.tsx
```diff
@@ -401,35 +401,34 @@ export class CompareSidebar extends React.Component<
...
-            kind: HistoryTabMode.Compare,
-            comparisonMode: ComparisonMode.Behind,
-            branch,
-          })
-
-          this.props.dispatcher.updateCompareForm(this.props.repository, {
-            filterText: branch.name,
-          })
-        }
+        const branch = this.state.focusedBranch
```
shiftkey: :art: extract this variable outside of the `if` statement, so this becomes a bit more succinct:

```diff
diff --git a/app/src/ui/history/compare.tsx b/app/src/ui/history/compare.tsx
index 7fa6d96f5..0b8be3f7e 100644
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -401,11 +401,11 @@ export class CompareSidebar extends React.Component<
         event.preventDefault()
         return
       }
-      if (this.state.focusedBranch == null) {
+      const branch = this.state.focusedBranch
+
+      if (branch == null) {
         this.viewHistoryForBranch()
       } else {
-        const branch = this.state.focusedBranch
-
         this.props.dispatcher.executeCompare(this.props.repository, {
           kind: HistoryTabMode.Compare,
           comparisonMode: ComparisonMode.Behind,
```

app/src/ui/history/compare.tsx
```diff
@@ -401,35 +401,34 @@ export class CompareSidebar extends React.Component<
         event.preventDefault()
         return
       }
-
-      if (this.props.compareState.filterText.length === 0) {
-        this.handleEscape()
+      if (this.state.focusedBranch == null) {
```
shiftkey: :art: can you use `===` here instead of `==`? We control the value here, and it should never be `undefined`...

app/src/ui/history/compare.tsx
ADustyOldMuffin: Consider it done!

app/src/ui/history/compare.tsx
ADustyOldMuffin: Also done!

ADustyOldMuffin:
@shiftkey Alright I think that was everything requested, lemme know if you see anything else.

PR #7821, opened 2019-06-17
Add support for WSL shell in Windows

### LINKED ISSUE #4641: Add support for WSL Bash on Windows 10

ADustyOldMuffin:
## Overview

**Closes #4641**

## Description

Written by  @Jayschwa on #6420 

> The shell is invoked with `wsl.exe`, which uses whichever Linux distribution the user has configured as their default. [1](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#wsl-and-wsl-command)
> 
> The option is presented to the user when the following conditions are met:
> 
>     1. `wsl.exe` and `wslconfig.exe` are both found in the system.
> 
>     2. `wslconfig.exe /list` returns an exit code of zero.
> 
> 
> I verified that `wslconfig.exe /list` returns a non-zero exit code when the system has no Linux distributions installed. I decided to avoid the registry for detection since that has been discouraged by the WSL developers. [2](https://github.com/Microsoft/WSL/issues/3515)
> ## Release notes
> 
> Notes: Adds support for WSL shell in Windows.


ADustyOldMuffin:
I created this PR as the original author said he could no longer champion it so I wanted to help out and work on it (sorry if there's a better way to do this 😳). 

As per the original PR I added the exit code to the error message, along with an extra log when the wslPath returns null just for some extra visibility as to what happened. 

I also added a reject call when the process is spawned, but I need to test it and make sure the error is actually handled as [the documentation](https://nodejs.org/api/child_process.html#child_process_event_error) mentions that it's possible to have some weird behavior.

app/src/lib/shells/win32.ts
```diff
@@ -268,6 +284,38 @@ async function findCygwin(): Promise<string | null> {
...
+      `[WSL] found wsl.exe, but wslconfig.exe does not exist at '${wslConfigPath}'`
+    )
+    return null
+  }
+  const exitCode = new Promise<number>((resolve, reject) => {
+    const wslDistros = spawn(wslConfigPath, ['/list'])
+    wslDistros.on('error', reject)
+    wslDistros.on('exit', resolve)
+  })
+  if ((await exitCode) !== 0) {
```
shiftkey: I know I mentioned making this promise `reject` if the underlying list `spawn` failed, but I think this now means we need to handle the error here.

Can we wrap this in a `try-catch` block and log these two scenarios:

 - the error that gets raised when the `await` throws because of a rejected promise
 - the non-zero exit code returned when the promise resolves

Something like this:

```ts
try {
  const result = await exitCode
  if (result !== 0) {
    log.debug(
      `[WSL] found wsl.exe and wslconfig.exe, but no distros are installed. Error code: ${result}`
    )
    return null
  }

  return wslPath
} catch (err) {
    log.error(`[WSL] unhandled error when invoking 'wsl /list'`, err)
} 
```

Another benefit of this change is we'll fix how the `log.debug` tries to write the promise itself to the log, when we want the result of the promise instead.

ADustyOldMuffin: I added the suggestion above, and I also return null if we reach the end of the method as if we make it out of the try catch block then something went wrong anyway. I can put the return into the catch block if needed, but I figured this would handle if something weird happened.

ADustyOldMuffin:
I also managed to get up and running with WSL finally and tested it on the current WSL and everything seems to check out in Windows 10.

shiftkey:
Thanks for helping to get this across the line @ADustyOldMuffin!

And thanks @jayschwa for laying the groundwork!

shiftkey:
This is still feature-flagged for development, so I'm going to merge this in so we can do more testing and then figure out when we want to enable it for beta to gather more feedback.

PR #7822, opened 2019-06-17
introduce work tree APIs and tests to support better conflict detection

shiftkey:
## Overview

🌵🌵🌵**builds upon #7819, please review that first** 🌵🌵🌵 

**Related to #7445**

## Description

This PR adds scaffolding to manage temporary worktrees from in the app. The goal here is to support being able to properly detect merge/rebase conflicts, without touching the user's working directory.

We can do this with the work in this PR using these functions exported from 

 - `findOrCreateTemporaryWorkTree` - create or use an existing temporary worktree associated with a repository
 - `cleanupTemporaryWorkTrees` - once complete, tidy up any worktrees created for use in Desktop

The API is subject to change, but it helped me get something going quickly to support rebase conflict detection in #7445 so I think it's on the right track at least.

I also added a bunch of tests of this plumbing to ensure that it works as expected on macOS/Windows/Linux.

## Release notes

Notes: no-notes


shiftkey:

PR #7839, opened 2019-06-20
use remotes info in branch pruning

### LINKED ISSUE #7604: Branch pruner not using remote info to determine if a branch can be pruned

outofambit:
## Overview

closes #7604

## Description

* modify `getMergedBranches` to only return a canonical ref for each branch (we weren't using the `sha` anywhere and it made some logic i needed to add simpler)
* get list of local remotes and cross-reference it with the upstreams of candidate branches for pruning
* update tests from @shiftkey 's prior work

## Next Steps

* ~testing~
* ~quick/small performance optimizations~


shiftkey:
I'll look over the proposed changes later today but I wanted to add some context to this bit first:

> but the tests pass? 

Check out the change I made to this test in #7616 - https://github.com/desktop/desktop/pull/7616/files#diff-b72d22e30762c772413c941a2cfd888c - I think if you enable and the tests continue to pass that it'll help confirm we have the right fix.

The context for that is written up in the "Good news!" part of the PR body.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -190,8 +193,24 @@ export class BranchPruner {
       mb => !ReservedRefs.includes(mb.canonicalRef)
     )
 
+    // get the local remote ref name for the github repository attached to this repo
+    const defaultRemoteRefName = (await getRemotes(this.repository)).find(
+      r =>
+        r.url === gitHubRepository.cloneURL ||
+        r.url === gitHubRepository.cloneURL
```
shiftkey: These checks are the same?

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -190,8 +193,24 @@ export class BranchPruner {
...
+      this.repository,
+      `refs/remotes/${defaultRemoteRefName}`
+    )).map(b => formatAsLocalRef(b.name))
+
     const branchesReadyForPruning = candidateBranches.filter(
-      mb => !recentlyCheckedOutCanonicalRefs.has(mb.canonicalRef)
+      candidate =>
+        !recentlyCheckedOutCanonicalRefs.has(candidate.canonicalRef) &&
+        !remoteBranches.includes(candidate.canonicalRef)
+      // do i need to have the upstream of the merged branch to make this check?
```
shiftkey: I'd prefer this check looked at the `upstream` of the branch because the names don't have to match when tracking an upstream branch in Git.

app/src/lib/stores/helpers/branch-pruner.ts
outofambit: 🤦‍♀

app/src/lib/stores/helpers/branch-pruner.ts
outofambit: yeah, you are totally right. good shout.

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -193,12 +193,25 @@ export class BranchPruner {
...
+      `refs/remotes/`
+    )).map(b => formatAsLocalRef(b.name))
+
+    const branchesReadyForPruning = candidateBranches.filter(ref => {
+      if (recentlyCheckedOutCanonicalRefs.has(ref)) {
+        return false
+      }
+      const upstream = getUpstreamRefForLocalBranchRef(ref, allBranches)
+      if (upstream === undefined) {
+        return false
```
shiftkey: `getUpstreamRefForLocalBranchRef` looks like it has two conditions where it returns `undefined`:

 - unable to find the branch matching `ref` -> this seems a valid case to return `false` here
 - unable to find an upstream branch for `ref` -> should we return `true` here to ensure a branch without a local upstream branch is pruned here?

I worry we're conflating both of these cases and potentially not pruning some branches, but maybe I'm missing some context...

app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -225,7 +238,7 @@ export class BranchPruner {
 
         if (isDeleted) {
           log.info(
-            `[BranchPruner] Pruned branch ${branchName} (was ${branch.sha})`
```
shiftkey: We logged the explicit `sha` value here so that the user could find this value in the logs if they wished to restore this ref manually. Can we add that back in?

app/src/lib/stores/helpers/branch-pruner.ts
outofambit: in this context, unable to find an upstream ref means that we don't have upstream information for that branch, so the branch doesn't have an upstream (or for some reason we can't figure out what it is). in other words, its the just local ref for the upstream not the actual local branch (which may or may not actually exist).

line 213 is where we actually check to see if the upstream ref exists locally.

i've added some docs for `getUpstreamRefForLocalBranchRef` to clarify this logic. Does that sound right? (it took me more than a couple minutes to retrace my steps here in this logic, so please lmk if anything doesn't)

app/src/lib/stores/helpers/branch-pruner.ts
outofambit: ah, i was missing that context for why it was there. will get that back! :)

outofambit: fe05dd9 and 2b79154 should have us covered :)

outofambit:
@shiftkey this is ready for another round of review 🍎

app/src/lib/stores/helpers/branch-pruner.ts
shiftkey: Yeah, I'm still not clear on why it's fine to have the control flow like this. I wasn't clear on how else I could explain it, so I put together a test case that illustrates my thinking: #7860

outofambit:
@tierninho this is great, thank you!

> Also, attempted to prune the active branch. It was not pruned as expected. If this is not the correct logic, lemme know.

that's all correct 👍

shiftkey:
@outofambit there's a minor conflict in here that's blocking me from merging the PR - it'd be great to include this with the next beta

outofambit:
@shiftkey should be good now!

outofambit:
i spoke too soon :(

shiftkey:
@outofambit looks like there's a build error currently:

```
[at-loader] ./app/test/helpers/repository-builder-branch-pruner.ts:91:7 
    TS2554: Expected 3 arguments, but got 4.
```

outofambit:
🤞

PR #7840, opened 2019-06-20
fix: prevent doing some actions when committing on action

### LINKED ISSUE #3814: lock the file/line selection controls while a commit is running

### LINKED ISSUE #5934: File selection of disabled entries possible with space
## Description
I have a long-running pre-commit hook. After hitting *commit*, the checkboxes for file selection are getting disabled immediately and any clicking on these checkboxes are prevented (everything fine here).  
However, I can still mark a checkbox by hitting space on a highlighted entry/file, which is quite misleading. Luckily, this has no impact on the commit itself (nor the pre-commit hook), so it's only a bug in convenience which might be confusing for users, but without any impact on functionality.

## Version
* GitHub Desktop: 1.4.2
* Operating system: Microsoft Windows [Version 10.0.17134.345]

## Steps to Reproduce
1. Add a long-running pre-commit hook; for testing, try `sleep 30s`
2. Edit some files (at least 2)
3. Select one file and commit it
4. While the pre-commit hook runs, select the other file and hit space. You can also deselect the currently committed file

### Expected Behavior
Hitting space doesn't do anything until the commit has been done.

### Actual Behavior
Hitting space selects/deselects files, although they're disabled.

HashimotoYT:
## Overview

**Closes #3814**
**Closes #5934**

## Description

This PR adds prevent doing actions with the application when we making a commit. Please notice, this not fixes all actions what we able to do when we committing, this fixes only some part of these bugs. 🦄

GIFs and actions what prevented:

- Diff panel goes to read only
	diff-readonly

- Opening the context menu
	context-menu

- Clicking undo button
	dont try to undo

- File selections from the keyboard ('Enter' or 'Spacebar')
	file-selections
- 'Discard all changes' context menu
	discard-all-changes

## Release notes

Notes: 🤔 💭


tierninho:
@HashimotoYT Many thanks for the PR on this. 

I tested all the scenarios and all passed expect the `File selections from the keyboard ('Enter' or 'Spacebar')` from https://github.com/desktop/desktop/pull/7840/commits/a246352c14ca395ea64802ea4bce3ac568f8875d. The `enter` and `space` do not work anymore before/after committing.

I am going to consider other scenarios and circle back later. Thanks again!

tierninho:
> The enter and space do not work anymore before/after committing.

Looks good now. Thanks.

app/src/ui/changes/changes.tsx
```diff
@@ -41,7 +43,7 @@ export class Changes extends React.Component<IChangesProps, {}> {
             repository={this.props.repository}
             imageDiffType={this.props.imageDiffType}
             file={file}
-            readOnly={false}
+            readOnly={isCommitting ? true : false}
```
iAmWillShepherd:
```suggestion
            readOnly={isCommitting}
```
`this.props.isCommiting` is already a boolean.

app/src/ui/changes/changes.tsx
```diff
@@ -12,6 +12,7 @@ interface IChangesProps {
   readonly diff: IDiff
   readonly dispatcher: Dispatcher
   readonly imageDiffType: ImageDiffType
+  readonly isCommitting: boolean
```
iAmWillShepherd:
```suggestion

/** whether a committing is in progress */
  readonly isCommitting: boolean
```

📝

app/src/ui/lib/list/list.tsx
```diff
@@ -419,6 +419,10 @@ export class List extends React.Component<IListProps, IListState> {
       this.props.onRowKeyDown(rowIndex, event)
     }
 
+    if (event.defaultPrevented) {
```
iAmWillShepherd: I think this change is fine, but since this touches a core component, I'd like to get the thoughts of someone else from @desktop/engineering before giving this a full ✅

outofambit: i'm not sure this `if` statement is really necessary. it looks like similar logic (on [L430](https://github.com/desktop/desktop/pull/7840/files#diff-a6fd50f3585afc2eeb1a60931fbf01a8R430)) should already take care of this. @HashimotoYT is there more context here that i'm missing?

iAmWillShepherd:
@HashimotoYT, I've gone ahead and pushed the requested changes up to your branch so we can get your PR merged 👍

tierninho:
Took another pass at it. LGTM!

PR #7866, opened 2019-06-25
Text search in diffs

### LINKED ISSUE #7538: Search Within Diffs
Some diffs can get pretty large. A simple search feature on a diff would be helpful in just finding a term or a line of code.

Example:

> When I am looking at the changes I made, I want to hit `ctrl f` and search for a certain term or possible mistake.

Maybe something similar to google chrome's search as a ui? The search bar sticks out at the top right.
Screen Shot 2019-05-15 at 1 40 35 PM

niik:
## Overview

**Closes #7538**

## Description

This turns on the CodeMirror search plugin in order to add text searching functionality to our diffs in Desktop.  The plugin is coupled with a little sprinkle of our own logic to make the search box look and feel like a Desktop component.

Searching is triggered by focusing on (clicking) the diff and hitting <kbd>⌘F</kbd> on macOS and <kbd>Ctrl</kbd>+<kbd>F</kbd> on Windows. Once the user had typed their query into the search field and hit <kbd>Enter</kbd> all matching locations in the document are highlighted and the first match is selected. Subsequent <kbd>Enter</kbd> presses will select the next match and <kbd>Shift</kbd>+<kbd>Enter</kbd> will select the previous match.

The search plugin also supports (though we're not going to advertise that) regular expression matchin by pre- and postfixing the expression with forward slashes (i.e `/search(Field|Label)/`).

## Demo

Jun-25-2019 15-53-04

## Release notes

Notes: [Added] Cmd/Ctrl+F will trigger a text search while focused on a diff.

### DIFF
--- a/app/src/ui/diff/code-mirror-host.tsx
+++ b/app/src/ui/diff/code-mirror-host.tsx
@@ -19,6 +19,8 @@ if (__DARWIN__) {
   require('codemirror/addon/scroll/simplescrollbars')
 }
 
+require('codemirror/addon/search/search')
+
 interface ICodeMirrorHostProps {
   /**
    * An optional class name for the wrapper element around the
--- a/app/src/ui/diff/code-mirror-host.tsx
+++ b/app/src/ui/diff/code-mirror-host.tsx
@@ -19,7 +19,7 @@ if (__DARWIN__) {
   require('codemirror/addon/scroll/simplescrollbars')
 }
 
-require('codemirror/addon/search/search')
+import 'codemirror/addon/search/search'
 
 interface ICodeMirrorHostProps {
   /**
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -129,6 +129,25 @@ interface ITextDiffProps {
 
 const diffGutterName = 'diff-gutter'
 
+function showSearch(cm: Editor) {
+  cm.execCommand('findPersistent')
+  const wrapper = cm.getWrapperElement()
+
+  if (wrapper) {
+    const searchLabel = wrapper.querySelector('.CodeMirror-search-label')
+    const searchField = wrapper.querySelector('.CodeMirror-search-field')
+
+    if (
+      searchLabel instanceof HTMLElement &&
+      searchField instanceof HTMLInputElement
+    ) {
+      searchLabel.style.display = 'none'
+      searchField.placeholder = 'Search'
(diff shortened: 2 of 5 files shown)


tierninho:
Tested and works as described. 👍 

As it's a plugin and we are limited to enhancements and fixes, let me know if you'd like me to test it further.

app/src/ui/diff/code-mirror-host.tsx
```diff
@@ -19,6 +19,8 @@ if (__DARWIN__) {
   require('codemirror/addon/scroll/simplescrollbars')
 }
 
+require('codemirror/addon/search/search')
```
outofambit: is this a require because of something to do with how the codemirror plugin loads?

app/src/ui/diff/text-diff.tsx
```diff
@@ -137,7 +165,11 @@ const defaultEditorOptions: IEditorConfigurationExtra = {
   lineWrapping: true,
   mode: { name: DiffSyntaxMode.ModeName },
   // Make sure CodeMirror doesn't capture Tab (and Shift-Tab) and thus destroy tab navigation
-  extraKeys: { Tab: false, 'Shift-Tab': false },
+  extraKeys: {
+    Tab: false,
+    'Shift-Tab': false,
+    [__DARWIN__ ? 'Cmd-F' : 'Ctrl-F']: showSearch,
```
outofambit: i literally didn't know you could do this with a ternary

outofambit:
i was also able to use <kbd>⌘</kbd> + <kbd>G</kbd> as well as <kbd>⌘</kbd> + <kbd>Shift</kbd> + <kbd>G</kbd> (on macOS) to move through the found instances of my search term.

app/src/ui/diff/code-mirror-host.tsx
niik: Old habit on my end, it used to be that typescript would complain loudly if you tried to import something which didn't have a module declaration so we used require to get around that. Now it seems it only warns about it so we can switch to import instead. Resolved in a8be19d2f

app/src/ui/diff/text-diff.tsx
niik: It comes in pretty handy at times, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names

I was sure we used it elsewhere in the app but during my quick look for examples I didn't find anything.

PR #7869, opened 2019-06-25
2.1.0-beta1 release

shiftkey:
## Overview

Aiming to ship this to production around 2nd July, so this is an important beta to gather feedback and ensure things are robust.

## Description

- [x] draft changelog
- [x] get 👍 on changelog
- [x] ship to beta

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -1,5 +1,17 @@
 {
   "releases": {
+    "2.1.0-beta1": [
+      "[New] Commit form will prompt user when working on a protected branch - #7826",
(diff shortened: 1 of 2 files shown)


changelog.json
```diff
@@ -1,5 +1,17 @@
 {
   "releases": {
+    "2.1.0-beta1": [
+      "[New] Commit form will prompt user when working on a protected branch - #7826",
```
billygriffin: I'm wondering if warn is more accurate than prompt, given that we use the warning icon?
```suggestion
      "[New] Commit form will warn user when working on a protected branch - #7826",
```

shiftkey: I dig it

PR #7883, opened 2019-06-27
set initial state before potential state update

shiftkey:
## Overview

This change addresses a warning emitted by React when running the app in development mode. @tierninho pointed this out to me yesterday in chat.

```
warning.js:33 Warning: Can't call setState on a component that is not yet mounted. This is a no-op, but it might indicate a bug in your application. Instead, assign to `this.state` directly or define a `state = {};` class property with the desired state in the ChooseBranchDialog component.
printWarning    @    warning.js:33
warning    @    warning.js:57
warnNoop    @    react.development.js:124
enqueueSetState    @    react.development.js:193
module.exports../app/node_modules/react/cjs/react.development.js.Component.setState    @    react.development.js:238
updateRebaseStatus    @    choose-branch.tsx:115
ChooseBranchDialog.onBranchChanged    @    choose-branch.tsx:108
ChooseBranchDialog    @    choose-branch.tsx:91
constructClassInstance    @    react-dom.development.js:6801
...
```

## Description

The root cause of this message is that `this.state` was not set if `ChooseBranch.onBranchChanged` was run, which could result in `setState` firing inside `updateRebaseStatus`.

The fix is to set `this.state` before any potential `setState` calls occur.

## Release notes

Notes: no-notes


app/src/ui/rebase/choose-branch.tsx
```diff
@@ -87,15 +87,15 @@ export class ChooseBranchDialog extends React.Component<
...
-    }
-
     this.state = {
       selectedBranch,
       rebasePreview: null,
       filterText: '',
     }
+
+    if (selectedBranch !== null) {
+      this.onBranchChanged(selectedBranch)
```
niik: So this *might* still end up calling `setState` before the component is mounted. Given that we're using a `promiseWithMinimumTimeout` it seems very unlikely but still, should we just move this call to `componentDidMount`?

shiftkey: @niik works for me

shiftkey: Addressed in afce9ea7e

PR #7886, opened 2019-06-27
update registry-js to version that is compatible with Electron 5

shiftkey:
## Overview

Related to #7799 

## Description

This can be tested and merged before we get to the Electron 5 upgrade itself, but this doesn't need to go out in a hurry.

This is also a Windows-only dependency, but can be verified fairly quickly by checking the Editor and Shell listings under the Options dialog. If those aren't populating as expected, more investigating is needed.

## Release notes

Notes: no-notes


shiftkey:
I'm adding this PR to the 2.2 project board because it's a dependency for migrating to Electron 5, which is currently in scope #6888

PR #7899, opened 2019-07-02
Trigger text diff search plugin even when CodeMirror doesn't have focus

niik:
## Overview

We shipped #7866 in 2.1.0. One limitation of the implementation there is that since CodeMirror is responsible for handling the keyboard shortcut it will never trigger unless the editor is focused (this was described in #7866).

## Description

This PR adds a menu item under the Edit menu which dispatches a custom event (similar to how we deal with select-all). The `TextDiff` component then registers an event handler *on the document node* allowing it to catch CmdOrCtrl+F regardless of what element is in focus. If one or more components are active and subscribing to this event the first one to register will get the opportunity to handle it first. Right now there are only ever one text diff active in the app at any given time so that's okay.

In the future we might want to either de-scope the focus area or have other component do so but we'll cross that bridge if and when we get there.

## Release notes

Notes: no-notes


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -159,6 +159,13 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+A',
         click: emit('select-all'),
       },
+      separator,
+      {
+        id: 'find',
+        label: __DARWIN__ ? 'Find' : '&Find',
```
outofambit: should this say 'Find in Diff'? or something like that to indicate the scope of the find?

niik: That's a great question. I went back and forth a bit before I picked the `Find` label. The primary thing that tipped the balance for `Find` vs `Find in Diff` was platform familiarity/consistency. Most apps that I looked at on Mac and Windows used Find unless there were multiple separate things one could search and even when there were multiple things to search, such as in VS Code, the primary action was still labelled `Find` and the secondary `Find in Files` with a separate shortcut.

So I'm moderately in favor of keeping it as just `Find` for now but I obviously welcome more input here.

outofambit: thanks for the context @niik. sounds good to me!

PR #7921, opened 2019-07-04
Hover Shows on list item

### LINKED ISSUE #7910: Name disambiguation repository selector list
When adding multiple instances (on the file system) of a repository of the same name, there is no way to disambiguate these repositories in selector list.

I  could (and do) have two (or more) separate checkouts of repo "Foo", on my file system. I can add both these instances to github desktop client. I then click on the "current repository" drop down list to select from all my added repositories.

In this scenario, all that will display in the subsequent dropdown list is two instances of the word "Foo", with no indication about which repo on my file system that these particular entries correspond to.

I can imagine two rather sensible fixes....

1. The simplest approach is to have a tooltip popup that displays the full path when you hover over the entries in this list.

2. The best approach would be to always display the path to the repo as smaller text, under the repo name. This would be similar to how commits are displayed in the history view.

Thanks for your consideration.

say25:
## Overview

**Closes #7910**

## Description

- Hover shows for entire repository list item

## Release notes

Notes: Hover shows for entire repository list item

### DIFF
--- a/app/src/ui/repositories-list/repository-list-item.tsx
+++ b/app/src/ui/repositories-list/repository-list-item.tsx
@@ -76,7 +76,11 @@ export class RepositoryListItem extends React.Component<
       : 'repository-list-item'
 
     return (
-      <div onContextMenu={this.onContextMenu} className={className}>
+      <div
+        onContextMenu={this.onContextMenu}
+        className={className}
+        title={repoTooltip}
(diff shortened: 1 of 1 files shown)


say25:
@outofambit also shouldn’t interfere with the `has changes` and `remote is ahead` indicators

PR #7957, opened 2019-07-15
Adds help URL for macOS Keychain issues

tierninho:
Following conversation with a Desktop user, we decided to add further documentation to the current error messaging. 

The current message lacks detail on how to fix the issue:
Screen Shot 2019-07-15 at 10 12 11 AM

This PR adds the this link: https://github.com/desktop/desktop/blob/development/docs/known-issues.md#the-username-or-passphrase-you-entered-is-not-correct-error-after-signing-into-account---3263 to the error message.

Unsure where to add:
```
private onShowAccountsStoreExamples = () => {
    this.props.dispatcher.openInBrowser('https://github.com/desktop/desktop/blob/development/docs/known-issues.md#the-username-or-passphrase-you-entered-is-not-correct-error-after-signing-into-account---3263')
  }
```

❓ If someone can point to the file, that would be helpful.

### DIFF
--- a/app/src/lib/stores/accounts-store.ts
+++ b/app/src/lib/stores/accounts-store.ts
@@ -97,7 +97,11 @@ export class AccountsStore extends TypedBaseStore<ReadonlyArray<Account>> {
       if (__DARWIN__ && isKeyChainError(e)) {
         this.emitError(
           new Error(
-            `GitHub Desktop was unable to store the account token in the keychain. Please check you have unlocked access to the 'login' keychain.`
+            `GitHub Desktop was unable to store the account token in the keychain. 
+             Please check you have unlocked access to the 'login' keychain. Further documention can be found {' '}
+             <LinkButton onClick={this.props.onShowExamples}>
+               here.`
+             </LinkButton>
           )
         )
       } else {


app/src/lib/stores/accounts-store.ts
```diff
@@ -97,7 +97,11 @@ export class AccountsStore extends TypedBaseStore<ReadonlyArray<Account>> {
       if (__DARWIN__ && isKeyChainError(e)) {
         this.emitError(
           new Error(
-            `GitHub Desktop was unable to store the account token in the keychain. Please check you have unlocked access to the 'login' keychain.`
+            `GitHub Desktop was unable to store the account token in the keychain. 
+             Please check you have unlocked access to the 'login' keychain. Further documention can be found {' '}
+             <LinkButton onClick={this.props.onShowExamples}>
```
shiftkey: Because we're inside the `AccountsStore`, where we're not allowed to emit JSX like this, I think my preferred approach to address this would be something like this:

 - create a new error type to represent this, e.g `KeyChainError`
 - throw this error instead of a regular `Error`
 - add error handling to our error handling infrastructure to detect this 
 - if we find this `KeyChainError`, we can display a popup with more context, rather than the default popup that is shown currently

For reference, our `ExternalEditorError` feels similar to this flow, so I'll point out the relevant parts:

 - created a custom error type (not sure if you need any extra fields)

https://github.com/desktop/desktop/blob/fcbb95d9df8741ab48f0dd06625c8204aa743e7c/app/src/lib/editors/shared.ts#L50-L59

 - write an error handler to look for this error (and display a popup if found)

https://github.com/desktop/desktop/blob/fcbb95d9df8741ab48f0dd06625c8204aa743e7c/app/src/ui/dispatcher/error-handlers.ts#L173-L192

 - handle the popup value to display a dialog

https://github.com/desktop/desktop/blob/fcbb95d9df8741ab48f0dd06625c8204aa743e7c/app/src/ui/app.tsx#L1490-L1503

 - wire up the error handler in the app setup:

https://github.com/desktop/desktop/blob/fcbb95d9df8741ab48f0dd06625c8204aa743e7c/app/src/ui/index.tsx#L255

tierninho:
Closing and adding https://github.com/desktop/desktop/issues/8032 to the OSC.

PR #7977, opened 2019-07-17
Enhance incorrect clone URLs error handling

### LINKED ISSUE #7914: Enhance error messaging for incorrect clone urls
Currently the error messaging for the clone repo URL modal does not properly handle URLs correctly and includes more info than needed. 

For instance:
https://github.com/aws-samples/amazon-kinesis-learning/tree/learning-module-1 results in this error

```
`git -c credential.helper= clone --recursive --progress -- https://github.com/aws-samples/amazon-kinesis-learning/tree/learning-module-1 /Users/tierninho/github/tests/learning-module-1` exited with an unexpected code: 128.
stderr:
Cloning into '/Users/tierninho/github/tests/learning-module-1'...
fatal: repository 'https://github.com/aws-samples/amazon-kinesis-learning/tree/learning-module-1/' not found

(The error was parsed as 8: The repository does not seem to exist anymore. You may not have access, or it may have been deleted or renamed.)
```

when really we should recognize the url is too long and should be tapered to https://github.com/aws-samples/amazon-kinesis-learning

Current logic takes the user to an old Auth modal like so:
Screen Shot 2019-07-03 at 2 23 01 PM

Our error messaging should include this scenario and suggest to the user to check the url, not reroute them to an auth modal.

v2.1, Mac

ADustyOldMuffin:
## Overview

**Closes #7914**

## Description

- Error handling when a URL that is to long is entered currently is not correct as it asks the person for their credentials rather than actually stating the error. This will hopefully handle if the URL is to long, or state a more clear error message to the user.

Screen Shot 2019-08-07 at 2 46 19 PM

## Release notes

Notes: Enhances incorrect clone URLs error handling


ADustyOldMuffin:
@tierninho Sorry to bug you, but since you were the one who made the issue I wanted to get your input. I've added tests to assist in fixing the issue. My idea is to fix how the URL is parsed first as you suggested and I was curious if you think I'm headed in the right direction or if I should look at something else.

tierninho:
@ADustyOldMuffin I'd say you are on the right track. We should parse the URL and display an inline message to the user to check their URL. 

The messaging can also be persistent after the user has submitted the form. I do not think we should block the user from submitting the form.

ADustyOldMuffin:
So we should parse it as normal, but instead show a warning to the user if we detect that the URL is to long?
Or should we try and trim the fat off of the URL, and display the warning?

tierninho:
> So we should parse it as normal, but instead show a warning to the user if we detect that the URL is to long?

Let's start with this option. Thanks for checking in.

ADustyOldMuffin:
So I've added two possible solutions if someone wants to test more cases (both have worked for the ones I can come up with). 
The two fixes both do different things and I have a preferred one I think is better, but I'm open to discussion and suggestions.

First fix [is this](https://github.com/ADustyOldMuffin/desktop/blob/ce864fe613a5b0642c933ada12fa1c5f050ea2a5/app/src/ui/clone-repository/clone-repository.tsx#L622-L627)
```
    } else if (!account && identifier) {
      if (identifier.hostname && identifier.hostname.indexOf('/') >= 0) {
        //The hostname has extra paths to it so the URL must be to long
        return null
      }
    }
```
which will fix that the incorrect URL is being recognized as correct, but I'm not a huge fan as the error given to the user is pretty generic when we know the issue.

Second fix [is this](https://github.com/ADustyOldMuffin/desktop/blob/ce864fe613a5b0642c933ada12fa1c5f050ea2a5/app/src/ui/clone-repository/clone-repository.tsx#L655-L662)
```
    if (url.substring(url.length - 4) !== '.git') {
      const error = new Error(
        `There seems to be an issue with the repository URL. Check that it is correct and is not too long.`
      )
      this.setState({ loading: false })
      this.setSelectedTabState({ error })
      return
    }
```
which will see that the URL sent back didn't resolve correctly and append it with .git. I like this approach as it gives a better error message, but I don't think my validation of if the URL is correct and that there's a better way to validate the URL given back.

ADustyOldMuffin:
It might also be possible that we could return an error of some sort from the method used to get the account and resolve the URL and then pass that back if received instead of checking if there is an error in the calling method.

tierninho:
@ADustyOldMuffin Thanks for all the work thus far!! 

Possible to sum up what you changed? Looks like there are four scenarios to check against: https://github.com/desktop/desktop/pull/7977/commits/d5b18e774be1ec96d92f088601dff089d1652199

Also, is this PR still in-progress or `ready-for-review`?

ADustyOldMuffin:
@tierninho So I got rid of the tests and things you reference in that commit as the approach was for a different scenario. Essentially if the host is returned with more paths after the main one (I.E. github.com/something/somethingelse rather than just github.com) then it returns null which signals an error rather than returning the URL as it did before.

I think its `ready-for-review` personally.

tierninho:
So I tested this and it works fine for HTTPs, but SSH failed to time out the attempted cloning process. 

On the current production build if an incorrect SSH clone is copy/pasted, it immediately times out with an error like so:

Screen Shot 2019-08-07 at 2 41 53 PM

On your build (b736de2fd1) it just hangs indefinitely.  For example, I tried cloning `git@github.com:tierninho/ssdfsdf/ffff.git`, which should trigger your error messaging.

ADustyOldMuffin:
Since I removed the fork this was with, I'm going to reopen a PR with the new fork and branch. Sorry I dropped this, but I'm working on it now.

PR #7989, opened 2019-07-20
add new types for Git work trees

outofambit:
## Overview

supports #6960 
based on #7819

## Description

- adds base type for work trees
- incorporates work tree type into repository model (every repository has a main work tree)
- adds linked work tree type

_____

## About work trees

> A git repository can support multiple working trees, allowing you to check out more than one branch at a time. With git worktree add a new working tree is associated with the repository. This new working tree is called a "linked working tree" as opposed to the "main working tree" prepared by "git init" or "git clone". A repository has one main working tree (if it’s not a bare repository) and zero or more linked working trees.

_from https://git-scm.com/docs/git-worktree_

### DIFF
--- a/app/src/models/repository.ts
+++ b/app/src/models/repository.ts
@@ -15,21 +15,35 @@ function getBaseName(path: string): string {
   return baseName
 }
 
+/** Base type for a directory you can run git commands successfully */
+export type WorkingTree = {
+  readonly path: string
+}
+
 /** A local repository. */
 export class Repository {
   public readonly name: string
+  /** The main working tree's path
+   * (also what we commonly think of as the repository's working directory)
+   */
+  private readonly mainWorkTree: WorkingTree
 
   /**
    * @param path The working directory of this repository
    * @param missing Was the repository missing on disk last we checked?
    */
   public constructor(
-    public readonly path: string,
+    path: string,
     public readonly id: number,
     public readonly gitHubRepository: GitHubRepository | null,
     public readonly missing: boolean
   ) {
+    this.mainWorkTree = { path }
     this.name = (gitHubRepository && gitHubRepository.name) || getBaseName(path)
+  }
+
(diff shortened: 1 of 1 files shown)


app/src/models/repository.ts
```diff
@@ -15,23 +15,37 @@ function getBaseName(path: string): string {
...
 
+/** Base type for a directory you can run git commands successfully */
+export type WorkingTree = {
+  readonly path: string
+}
+
 /** A local repository. */
 export class Repository {
   public readonly name: string
+  /** The main working tree's path
```
shiftkey: :art: as this is a multi-line comment, let's bring the first comment down to the next line to let it breathe a bit

app/src/models/repository.ts
```diff
@@ -15,23 +15,37 @@ function getBaseName(path: string): string {
...
+   * (also what we commonly think of as the repository's working directory)
+   */
+  private readonly mainWorkTree: WorkingTree
```
shiftkey: @outofambit can you tell me a bit about how you envision this being used? It's `private` at the moment, and it'd be cool to think about what should be accessing this and why it might need to?

outofambit: honestly i'm not sure yet, which is why its marked `private` for now. mostly, i'd like our model to reflect the relationships between Git concepts, as much for accuracy as for making these things easier to understand for people coming into the codebase.

my first thought was to replace the use of `Repository.path` with `Repository.mainWorkTree.path` but that would be a much larger (though not complicated) refactor and wouldn't get us immediate material benefits. so having that internally with the same `.path` api seemed like a decent compromise to start from.

shiftkey: I'm a bit hesitant to move `path` under `mainWorkTree` and make callers aware of that, so what you've done here makes sense. Just some food for thought as you're making changes to these internals.

PR #8006, opened 2019-07-23
Git WorkTree APIs

outofambit:
## Overview

supports #6960
based on #7822

## Description

- adds functions for creating, listing, and removing git work trees
- adds [`jest-extended`](https://github.com/jest-community/jest-extended) for better assertions around strings (paths)


app/src/lib/git/worktree.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  return workTree
+}
+
+/** Cleanup the temporary worktree at a given location */
+export async function removeWorkTree(
+  repository: Repository,
+  path: string
+): Promise<true> {
+  await git(
```
iAmWillShepherd: If the worktree is not able to be removed, is returning `true` still cool?

app/src/lib/git/worktree.ts
```diff
@@ -0,0 +1,143 @@
...
+    // NodeJS doesn't seem to include this when we ask for the temporary
+    // directory for the OS
+    const directoryName = Path.basename(t.path)
+    return directoryName.startsWith(DesktopWorkTreePrefix)
+  })
+}
+
+/**
+ * Lookup or create a new temporary work tree for use in Desktop without
+ * touching the user's current repository state.
```
iAmWillShepherd: I think a note here that this function also updates `HEAD` due to a checkout would be nice.

app/src/lib/git/worktree.ts
outofambit: if any errors are occurred dugite should throw an error, in which case the caller would not get `true` returned so i think this is ok. does that sound right?

app/src/lib/git/worktree.ts
outofambit: done in 96a12fe3d146ef187689fa9eb891eedf018ec4b5

iAmWillShepherd:
@outofambit feel free to merge when ready 👍

PR #8018, opened 2019-07-25
swap request bot label to new one

tierninho:
The proper label is now `more-info-needed` so the bot created the old label `more-information-needed` each time it was evoked.

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -13,7 +13,7 @@ requestInfoReplyComment: >
 
   Thanks for understanding and meeting us halfway 😀
 
-requestInfoLabelToAdd: more-information-needed
+requestInfoLabelToAdd: more-info-needed
 
 requestInfoOn:
   pullRequest: false


shiftkey:
Merging this in now to prevent further mis-labelling

Rexogamer:
More info needed about this PR

shiftkey:
@Rexogamer this file represents the config we use for the [`request-info`](https://github.com/apps/request-info) GitHub App, which triages opened issues. If someone doesn't specify anything in the issue body, it will comment and add a label to indicate it's incomplete.

The change here is that we changed the label name in the issue tracker, but the config points at the old label.

Rexogamer:
> @Rexogamer this file represents the config we use for the [`request-info`](https://github.com/apps/request-info) GitHub App, which triages opened issues. If someone doesn't specify anything in the issue body, it will comment and add a label to indicate it's incomplete.
> 
> The change here is that we changed the label name in the issue tracker, but the config points at the old label.

whoooosh

PR #8021, opened 2019-07-25
update keychain known issue to indicate this is relevant to later macOS versions

jipexu:
still present with mojave
[#3263]


shiftkey:
I haven't been following along with this issue, but I suspect this is a reasonable change.

@desktop/support any thoughts on your end?

docs/known-issues.md
```diff
@@ -20,7 +20,7 @@ Please check the [open](https://github.com/desktop/desktop/labels/bug) and [clos
 
 ### 'The username or passphrase you entered is not correct' error after signing into account - [#3263](https://github.com/desktop/desktop/issues/3263)
 
-This seems to be caused by the Keychain being in an invalid state, affecting applications that try to use the keychain to store or retrieve credentials. Seems to be specific to macOS High Sierra (10.13).
+This seems to be caused by the Keychain being in an invalid state, affecting applications that try to use the keychain to store or retrieve credentials. Seems to be specific to macOS High Sierra (10.13) to macOS Mojave 10.14.5 (18F132).
```
steveward:
```suggestion
This seems to be caused by the Keychain being in an invalid state, affecting applications that try to use the keychain to store or retrieve credentials. This has been reported from macOS High Sierra (10.13) to macOS Mojave 10.14.5 (18F132).
```

steveward (changes requested):
@jipexu I've seen less reports of this, but if you have seen this on macOS 10.14.5 then this change makes sense to me. I changed some of the wording to clarify the range of versions.

docs/known-issues.md
jipexu: Thanks 
yes i have experimented both  ...  on and on mojave just recently when i post this ... so since high sierra to mojave is true ... and the process to solve stay the same lock and unlock ....

steveward: @jipexu Great, thanks. If you can go ahead and commit the suggested change I made we can get this merged in.

jipexu: @steveward sorry i dont know what i must do now .... ?

shiftkey: @jipexu if you are fine with the wording change proposed by @steveward, the "Commit suggestion" button above will apply this change to the PR



Then I think this will be good to merge

aguynamedben:
+1. Our Electron desktop app uses keytar and also sees this annoying issue still present in user with macOS 10.14.5 (18F132).

Any idea how macOS Keychain gets in this state? My users are non-technical so the workaround isn't very friendly.

This is presumed to be a bug with macOS Keychain itself, right? Not keytar?

shiftkey:
> Any idea how macOS Keychain gets in this state? My users are non-technical so the workaround isn't very friendly.
> 
> This is presumed to be a bug with macOS Keychain itself, right? Not keytar?

@aguynamedben it's an OS-specific issue with Keychain and not something we can really detect or repair from `keytar` when it happens.

aguynamedben:
@shiftkey Thanks for weighing in... and thanks for all the work you put into keytar 🙏

xsangi (dismissed):

shiftkey (approved):

shiftkey:
@steveward I plan to merge this in tomorrow unless you had more to add

steveward (approved):

PR #8026, opened 2019-07-25
git patch APIs

outofambit:
## Overview

supports #6960

## Description

- adds functions for generating patches and applying them
- adds `jest-extended` matchers (cherry picked commits from #8006)


app/test/unit/git/apply-test.ts
```diff
@@ -0,0 +1,49 @@
...
+    describe('on related repository without conflicts', () => {
+      let repository: Repository
+      let patch: string
+      beforeEach(async () => {
+        const originalRepo = await setupTwoCommitRepo()
+        repository = await cloneLocalRepository(originalRepo)
+        await makeCommit(originalRepo, {
+          entries: [{ path: 'just-okay-file', contents: 'okay' }],
+        })
+        ;({ stdout: patch } = await GitProcess.exec(
```
shiftkey: Can we make this variable assignment more readable? What about something like this?

```ts
  ['format-patch', '--stdout', 'HEAD~'],
  originalRepo.path
)
patch = result.stdout
```

app/src/lib/git/format-patch.ts
```diff
@@ -0,0 +1,26 @@
+import { git } from './core'
+import { revRange } from './rev-list'
+import { Repository } from '../../models/repository'
+
+/**
+ * Generate a patch containing the changes associated with this range of commits
```
shiftkey: :art: tightening this wording up a bit, feel free to tweak or decline it

```suggestion
 * Generate a patch representing the changes associated with a range of commits
```

app/test/unit/git/format-patch-test.ts
```diff
@@ -0,0 +1,39 @@
...
+    describe('applied in a related repo', () => {
+      let clonedRepository: Repository
+      beforeEach(async () => {
+        clonedRepository = await cloneLocalRepository(repository)
+        await makeCommit(clonedRepository, {
+          entries: [{ path: 'okay-file', contents: 'okay' }],
+        })
+      })
+      it('is valid', async () => {
+        const patch = await formatPatch(repository, 'HEAD~', 'HEAD')
```
shiftkey: Have you tried testing with a range of multiple commits (`HEAD~..HEAD` is a single commit)?

I'm curious whether `git diff` as it's currently used will cover all the cases that `git format-patch` supports.

outofambit: yeah i think you're suspicion is right. `diff` will create a single patch instead of a series of patches.

i made this change due to limitations of buffering output from a child process. if there were too many commits, the output would be truncated. i experimented with raising the output buffer limit, but i had to make it very large to accommodate some test cases i tried, which resulted in tripling desktop's usual heap size.

i'm going to investigate some more alternatives. piping output directly from one child process to another doesn't get us around the buffering issue. so we could
* write a potentially large patch file to disk and read it back in (probably not as performant)
* write a script/command that pipes the output of one git command to another thereby avoiding the buffering issue (might have cross-platform issues)
* abandon the patch approach and simply execute the expected rebase command in the working tree and check the output for conflicts (probably not as performant)

shiftkey: What about our current `spawn` APIs? This might be a bit more flexible for this use case:

https://github.com/desktop/desktop/blob/e71e497e7366f53820f408aa1bc49a7eb8ac6426/app/src/lib/git/spawn.ts#L13-L26

app/src/lib/git/format-patch.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns patch generated
+ */
+export async function formatPatch(
+  repository: Repository,
+  base: string,
+  head: string
+): Promise<string> {
+  const range = revRange(base, head)
+  const result = await git(
+    ['diff', '--unified=1', '--minimal', range],
```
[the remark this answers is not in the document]
outofambit: https://github.com/desktop/desktop/pull/8026/commits/4f992081cd4ec662bf5f17d4ba028c271d82fdab

app/test/unit/git/apply-test.ts
outofambit: cce48cf93d5ad31f9d8619ad7502381a3366ec1f

app/src/lib/git/apply.ts
```diff
@@ -67,3 +68,33 @@ export async function applyPatchToIndex(
 
   return Promise.resolve()
 }
+
+/**
+ * Test a patch to see if it will apply cleanly.
+ *
+ * @param workTree work tree which should be checked out to the right base commit
+ * @param patch the raw patch generated by Git to test (see `formatPatch`)
```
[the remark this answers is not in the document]
outofambit: ba1fb681c51cdf56f872cd167c72c74365a92f90

app/test/unit/git/format-patch-test.ts
outofambit: thanks @shiftkey! that did the trick! (d828453cf90b950c0289942a25d427e60bb00f87)

outofambit:
this is ready for another go-around @shiftkey! thanks!

app/test/unit/git/format-patch-test.ts
```diff
@@ -0,0 +1,72 @@
...
+    beforeEach(async () => {
+      const path = await setupFixtureRepository('repository-with-105-commits')
+      repository = new Repository(path, -1, null, false)
+      const { stdout } = await GitProcess.exec(
+        ['rev-list', '--max-parents=0', 'HEAD'],
+        path
+      )
+      firstCommit = stdout.trim()
+    })
+    it('resolves with a string', async () => {
```
shiftkey: :art: I think we can make this test name a bit more contextual:

```suggestion
    it('can create a series of commits from start to HEAD', async () => {
```

app/test/unit/git/format-patch-test.ts
```diff
@@ -0,0 +1,72 @@
...
+    })
+    describe('applied in a related repo', () => {
+      let clonedRepository: Repository
+      beforeEach(async () => {
+        clonedRepository = await cloneLocalRepository(repository)
+        await makeCommit(clonedRepository, {
+          entries: [{ path: 'okay-file', contents: 'okay' }],
+        })
+      })
+      it('is valid', async () => {
```
shiftkey: :art: to make this test a bit more descriptive of what Git is doing with the patch

```suggestion
      it('will be applied cleanly', async () => {
```

app/test/unit/git/apply-test.ts
outofambit: 🤦‍♀

outofambit: dab8b4815f91980e03221a9dda052f0edbbda71b

shiftkey:
@outofambit I was planning to merge this in tomorrow, but let me know if you'd like it sooner to unblock other work

outofambit:
@shiftkey thanks! yeah, sometime today would be great!

PR #8045, opened 2019-07-29
Hide Whitespace in Historical Commits

say25:
## Overview

**Addresses #1940**

## Description

- Adds a "Hide Whitespace" Checkbox to in the `History` tab

## Release notes

Notes:

### History Tab

- Checkbox has moved to the right since filming this `.gif`

historytab


app/src/lib/git/diff.ts
```diff
@@ -120,8 +124,10 @@ export async function getCommitDiff(
     args.push(file.status.oldPath)
   }
 
+  const argsToUse = args.filter(arg => arg !== '')
```
outofambit: instead of using a `.filter` here, could we conditionally `.splice` `'-w'` into `args` or something like that? we could make the `whiteSpaceArg` ternary return `[]` or `['-w']`and spread (`...`) that into `args` at creation time.

i'd like to avoid adding arguments that have to be removed later.

shiftkey:
@say25 @outofambit can we ensure this change is feature-flagged, so we can control how this is rolled out and give us the option to roll it back if we feel later it's not the right approach?

say25:
Sounds reasonable to me

say25:
@outofambit is it cool if we enable this in production release now?

outofambit:
@say25 yes! i’m happy to approve that PR and make sure it gets in!

Issue #7991, opened 2019-07-20
Menus disabled after branch rebase failure


immackay:
## Description

Rebasing from a branch to master with unstashed changes produces a dialog stating that the changes must be stashed or discarded. After this dialog is dismissed, all menu items under the `Repository` and `Branch` menus are disabled. They are reenabled upon restarting the application.

## Version

* GitHub Desktop: 2.1.0

* Operating system: MacOS 10.14.5

## Steps to Reproduce

1. Have unstashed or uncommitted changes in your working branch
2. Rebase current branch (Shift-Command-E on Mac)
3. Select master and press `Begin Rebase` (This may work rebasing to other branches as well, I have not tested)
4. An error dialog will show stating that there are unstaged changes - close this *immediately*
5. The `Repository` and `Branch` menus will be entirely disabled.

Interestingly enough, if you wait ~10 seconds before closing the error dialog this behaviour does not occur.

### Expected Behavior

`Repository` and `Branch` menus to not be disabled

### Actual Behavior

`Repository` and `Branch` menus are disabled

### Logs

Relevant lines:
```
2019-07-20T07:18:16.484Z - info: [ui] [rebase] starting rebase for ui-modification at f9c7799c6768accac71d6c158350f697a17e28ef
2019-07-20T07:18:16.485Z - info: [ui] [rebase] to restore the previous state if this completed rebase is unsatisfactory:
2019-07-20T07:18:16.485Z - info: [ui] [rebase] - git checkout ui-modification
2019-07-20T07:18:16.486Z - info: [ui] [rebase] - git reset f9c7799c6768accac71d6c158350f697a17e28ef --hard
2019-07-20T07:18:16.640Z - error: [ui] `git rebase master ui-modification` exited with an unexpected code: 1.
stderr:
Cannot rebase: You have unstaged changes.
Please commit or stash them.

2019-07-20T07:18:16.659Z - info: [ui] [rebase] completed rebase - got Error and on tip f9c7799c6768accac71d6c158350f697a17e28ef - kind Valid
2019-07-20T07:18:17.555Z - warn: [ui] [App] invalid state encountered - rebase flow should not be active when step is null
```

tierninho:
@immackay I was able to reproduce this bug, however my steps were slightly different.

Steps:
1. Have file that will trigger a merge conflict on your branch
2. Start a rebase
4. Abort the merge conflict dialog
5. The `Repository` and `Branch` menus will be entirely disabled.

Logs:

```
[...]

PR #8055, opened 2019-07-31
menu update does not invoke required Electron API to update application menu for macOS

shiftkey:
## Overview

**Related to #7991**

## Description

This adds an explicit `Menu.setApplicationMenu(menu)` call where we previously didn't had one.

This seems to have worked incidentally when the renderer emits a `'update-menu-state'` event, but I believe the upgrade to Electron 5 has changed this behaviour.

I also tweaked the flow to save calling `Menu.getApplicationMenu()` within the `for` loop, as this check would exit the update if it can't find a menu. This can be done outside the loop and save some cycles.

## Release notes

Notes: `[Fixed] menu state not updated for macOS after performing some actions`

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -396,6 +396,7 @@ app.on('ready', () => {
       }
 
       if (sendMenuChangedEvent && mainWindow) {
+        Menu.setApplicationMenu(currentMenu)
         mainWindow.sendAppMenu()
       }
     }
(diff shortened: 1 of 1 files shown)


tierninho:
> I've opened #8055 to close out this case, and I'll see if the original issue is also fixed by this change (I suspect not, because it's currently on production which is still Electron 3).

Tested this fix and menu seems fixed after small changes like deleting a branch, but as you suspected the rebase menu blackout is still an issue.

PR #8060, opened 2019-08-01
use checkout to resolve unmerged binary file conflicts

### LINKED ISSUE #8059: Binary conflict resolve doesn't work
## Description

When I pull repository from remote, there is conflict for an image. When I resolve the conflict with accept the remote branch in GitHub Desktop, and then commit and push, the result isn't expected, it's  still the image in my local, not the image of remote.
image

## Version

* GitHub Desktop: 2.1.0

## Steps to Reproduce

1. commit changes for 1.png
2. pull from remote which has changes for 1.png
3. show the conflict dialog for 1.png
4. chose "use the modified file from origin/xxx"
5. commit merge, and push

### Expected Behavior

the image is the remote one

### Actual Behavior

it's still the local one, didn't change

## Additional Information

Even for banch merge, such as merge master into branch1, the image conflict resolve doesn't work either.

outofambit:
## Overview

Closes #8059 

## Description

it looks like our prior implementation of manual conflict resolution never worked for files that were changed in both branches. this is because it would simply stage whatever state the file was in in the index (probably a weird conflicted state, or whatever was their before the merge/rebase was started).

this PR adds the step of explicitly `git checkout`-ing the desired version of the file before staging it, ensuring we choose the single version of the file the user requested in the UI.

we need to double check that this works for merging _and_ rebasing due to [this potential complication](https://git-scm.com/docs/git-checkout#Documentation/git-checkout.txt---theirs). (it might check out the opposite version in a rebasing flow, but that should be easy to address if that is the case.)

## Release notes

Notes: [Fixed] Manual conflict resolution always added the same file version

### DIFF
--- a/app/src/lib/git/stage.ts
+++ b/app/src/lib/git/stage.ts
@@ -54,13 +54,31 @@ export async function stageManualConflictResolution(
       )).exitCode
       break
     }
-    case GitStatusEntry.Added:
-    case GitStatusEntry.UpdatedButUnmerged: {
+    case GitStatusEntry.Added: {
       exitCode = (await git(
         ['add', file.path],
         repository.path,
         'addConflictedFile'
       )).exitCode
+      break
+    }
+    case GitStatusEntry.UpdatedButUnmerged: {
+      const choiceFlag =
+        manualResolution === ManualConflictResolutionKind.theirs
+          ? 'theirs'
+          : 'ours'
+      exitCode = (await git(
+        ['checkout', `--${choiceFlag}`, '--', file.path],
+        repository.path,
+        'checkoutConflictedFile'
+      )).exitCode
+      if (exitCode === 0) {
+        exitCode = (await git(
+          ['add', file.path],
+          repository.path,
+          'addConflictedFile'
+        )).exitCode
+      }
       break
     }
     default:
--- a/app/test/helpers/repositories.ts
+++ b/app/test/helpers/repositories.ts
@@ -183,13 +183,6 @@ export async function setupConflictedRepoWithMultipleFiles(): Promise<
   Repository
 > {
   const repo = await setupEmptyRepository()
-  const filePaths = [
-    Path.join(repo.path, 'foo'),
-    Path.join(repo.path, 'bar'),
-    Path.join(repo.path, 'baz'),
-    Path.join(repo.path, 'cat'),
-    Path.join(repo.path, 'dog'),
-  ]
 
   const firstCommit = {
     entries: [{ path: 'foo', contents: 'b0' }, { path: 'bar', contents: 'b0' }],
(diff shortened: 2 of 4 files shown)


app/src/lib/git/stage.ts
```diff
@@ -54,15 +54,33 @@ export async function stageManualConflictResolution(
...
+        ['checkout', `--${choiceFlag}`, '--', file.path],
+        repository.path,
+        'checkoutConflictedFile'
+      )).exitCode
+      if (exitCode === 0) {
+        exitCode = (await git(
+          ['add', file.path],
+          repository.path,
+          'addConflictedFile'
+        )).exitCode
```
shiftkey: :art: not a blocker, but seeing how we poke at the `exitCode` within a switch statement a few times I wonder if there's a better way to organize this logic.


 -   introduce a simple wrapper to de-duplicate the same work that each Git operation does in this function:

```ts
/**
 * Run a Git command and return whether the exit code indicated success
 *
 * This defers to the default error handling infrastructure inside if an error
 * is encountered.
 */
async function runGitCommand(
  args: string[],
  path: string,
  name: string
): Promise<boolean> {
  const { exitCode } = await git(args, path, name)
  return exitCode === 0
}
```

 - tidy up the simple case to make it a bit more readable:

```ts
    case GitStatusEntry.Deleted: {
      return await runGitCommand(
        ['rm', file.path],
        repository.path,
        'removeConflictedFile'
      )
    }
```

 - For this new case, we also get a bit more readability:

```
      const choiceFlag =
        manualResolution === ManualConflictResolutionKind.theirs
          ? 'theirs'
          : 'ours'
      const checkoutCompleted = await runGitCommand(
        ['checkout', `--${choiceFlag}`, '--', file.path],
        repository.path,
        'checkoutConflictedFile'
      )
      if (checkoutCompleted) {
        return await runGitCommand(
          ['add', file.path],
          repository.path,
          'addConflictedFile'
        )
      }
      return false
```

This should make the variable `exitCode` unnecessary, and you can `return assertNever` below to make the compiler happy.

shiftkey:
> (it might check out the opposite version in a rebasing flow, but that should be easy to address if that is the case.)

If you're not keen to tackle this case while you're in here, 👍 to tracking this in a fresh issue.

Any interest in capturing a test that verifies this behaviour for merge conflicts?

tierninho:
Testing results are as follows:

#### Rebasing
- Rebase conflict, select option 1 = unable to successfully verify outcome as no commit was shown with the proper selected file
- Rebase conflict, select option 2 = unable to verify outcome as no commit was shown with the proper selected file

Screen Shot 2019-08-01 at 8 41 02 AM

#### Merge conflict
- Merge conflict, select option 1 =  unable to successfully verify outcome as no commit was shown with the proper selected file
- Merge conflict, select option 2 =  unable to successfully verify outcome as no commit was shown with the proper selected file

Screen Shot 2019-08-01 at 8 59 55 AM

Perhaps I was looking for the wrong outcome. Please advise if so.

99dcd8b8ce9f35d934130f7f0011e3efd150a4a9, mac

kuychaco:
I'll be picking this work up and carrying it forward while @outofambit is out this week. Decided to open a new PR for concision rather than add more noise to this one, and keep the PR body of this one preserved for posterity. So I'm closing this as it's superseded by #8097

Issue #7617, opened 2019-05-23
Prevent concurrent appStore._checkoutBranch call from error handlers


iAmWillShepherd:
Original context from @outofambit and @niik in https://github.com/desktop/desktop/pull/7474#discussion_r286717265

A great shout @outofambit. `_checkoutBranch` does become re-entrant in this scenario. I think we should still await here though, not awaiting will not alter the re-entrant nature of the logic as `performFailableOperation` doesn't itself await the result of the error handlers. I think instead we should focus on your second idea and explore that a little bit more but I definitely agree with your assessment of risk in doing so.

I imagine that the fact that we have this disconnect in the awaitable chain will cause other problems such as the `StashAndSwitchBranch` progress not working properly. If you choose to move your changes to the next branch and hit Ok, the dialog progress will start spinning but if the checkout fails and the error handler takes over I would imagine that the dialog will close and the stash + checkout operation will continue in the background.

My gut feeling here is that a better approach would be to have `_checkoutBranch` catch this specific error inline and deal with it accordingly rather than waiting for the global error handler to pick it up. That, however, is a much bigger task that I don't think we'll have time to implement and test but definitely something worth investigating post 2.0.

A completely untested rough draft of how inline-management of this would work:

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index 05d219da5..5472d89d9 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2918,10 +2918,48 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const checkoutSucceeded =
       (await this.withAuthenticatingUser(repository, (repository, account) =>
         gitStore.performFailableOperation(
-          () =>
-            checkoutBranch(repository, account, foundBranch, progress => {
-              this.updateCheckoutProgress(repository, progress)
-            }),
+          async () => {
+            try {
+              await checkoutBranch(
+                repository,
+                account,
+                foundBranch,
+                progress => {
+                  this.updateCheckoutProgress(repository, progress)
+                }
+              )
+              return true
+            } catch (err) {
+              if (
+                !isLocalChangesWouldBeOverwrittenError(err) ||
+                stashToPop !== null
+              ) {
+                throw err
+              }
+
+              if (
+                !(await gitStore.performFailableOperation(() => {
+                  return createDesktopStashEntry(repository, foundBranch.name)
+                }))
+              ) {
+                return false
[...]

Issue #7609, opened 2019-05-22
extract stash creation in AppStore._checkoutBranch into its own method


outofambit:
This is a follow up refactor issue (and a subset of the work captured in #7607) to simplify some logic in `AppStore._checkoutBranch` and localize variable creation and assignment for 

https://github.com/desktop/desktop/blob/49d5d54c9096b7a8b9d9b0cc93664033f11c5a04/app/src/lib/stores/app-store.ts#L2868

The idea is to encapsulate the following code block into its own private function on AppStore that returns `Promise<StashEntry | null>` so we can assign that result directly to `stashToPop`.

https://github.com/desktop/desktop/blob/49d5d54c9096b7a8b9d9b0cc93664033f11c5a04/app/src/lib/stores/app-store.ts#L2885-L2916

And then check the `stashToPop` later in `_checkoutBranch` to see if we have something non-`null` to `pop`.

_Originally posted by @outofambit in https://github.com/desktop/desktop/pull/7474_

PR #8076, opened 2019-08-07
Handle when checkout branch fails

probablycorey:
## Overview

Addresses https://github.com/desktop/desktop/issues/7617

One of the concerns in the issue above was that Desktop wasn't awaiting calls to `checkoutBranch`

```ts
 const checkoutSucceeded =
  (await this.withAuthenticatingUser(repository, (repository, account) =>
    gitStore.performFailableOperation(
      () =>
        checkoutBranch(repository, account, foundBranch, progress => {
          this.updateCheckoutProgress(repository, progress)
  }),
```

It turns out it is already awaiting, because `checkoutBranch` returns a promise to `performFailableOperation` which returns a promise to `withAuthenticatingUser` which is awaited.

But there is another concern raised in the issue. When a checking out a branch fails (because of conflicts or some other evil) Desktop should stashing the changes and checkout the branch again. This was happening as a side effect in the error handlers, but now it is handled explicitly in the `AppStore._checkoutBranch` method.

Speaking of the _checkoutBranch method, it's pretty long and tricky to follow. I created this pseudo code story to help me understand what the method is trying to do:

```
Do you have changes and is “ask before stashing” set?
	**Yes**
		Show a that popup and exit before doing anything (the popup will end up calling checkoutBranch again)
	**No**
		continue	
Are you on a branch and you want to keep your changes on that current branch?
	Stash the changes for the current branch
Or do you want to move your changes to the branch you are checking out?
	Are there deleted files and there is no transientStashEntry
		Create a stash
Checkout the branch
	Was there an error (maybe a conflict or some other badness)
		Is there already a stash
			THROW ERROR
		Are there no local changes?
			THROW ERROR
		Stash the changes
		Checkout again
Is there a stash and you moving changes to a new branch?
        Apply the stash
etc...
```

Now that I understand how it works I think I'll be able to clean up the method a bit (like https://github.com/desktop/desktop/issues/7609 describes) in a follow up PR.

Closes https://github.com/desktop/desktop/issues/7617

## Description

Notes:
- Diff best viewed with `Hide whitespace changes` on.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2988,7 +2988,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     try {
       this.updateCheckoutProgress(repository, {
-        kind: "checkout",
+        kind: 'checkout',
         title: __DARWIN__ ? 'Refreshing Repository' : 'Refreshing repository',
         value: 1,
         targetBranch: foundBranch.name,
@@ -3068,7 +3068,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
           repository,
           branch.name
         )
-        await authenticatedCheckoutBranch(repository)  
+        await authenticatedCheckoutBranch(repository)
         return stash
       } else {
         return null
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2963,23 +2963,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     if (
       uncommittedChangesStrategy.kind ===
-        UncommittedChangesStrategyKind.MoveToNewBranch &&
-      stashToPop
+      UncommittedChangesStrategyKind.MoveToNewBranch
     ) {
-      // We increment the metric after checkout succeeds to guard
-      // against double counting when an error occurs on checkout.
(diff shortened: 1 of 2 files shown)


probablycorey:
I think there is a problem with some of my assumptions, specifically "Was there an error (maybe a conflict or some other badness)"

During my testing I was trying to get a checkout to fail, but I couldn't. I tried checking out a file with conflicts, and checking out a file with unstaged changes, but none of that caused `checkout` to fail. So I need to dig into this a little deeper.

probablycorey:
I moved the logic from @niik's into another method. The only major difference is I removed the `gitStore.performFailableOperation` so the error from `checkoutBranch` can be caught. But after looking at this with @kuychaco, we were wondering if this change was worth it.

The user-facing experience with this is pretty minimal. It would ensure that popups remain open while the reentrant checkoutBranch is run. In my testing this wasn't noticeable, but maybe if it is a giant stash or checkout it would be more noticeable.

One thing we did notice is how complicated `checkoutBranch` has become, so maybe it makes more sense to refactor more of the method before adding this additional error detection?

probablycorey:
Closing based on the findings described in https://github.com/desktop/desktop/issues/8099

Issue #7720, opened 2019-06-07
Cannot switch branch with untracked files


markp2:
## Description

If you have any untracked files you cannot switch branches in desktop because this new dialog appears. Previously this was not the case and you could switch branches and carry around untracked files.

![](https://user-images.githubusercontent.com/33937982/59070161-a1674780-88b1-11e9-836e-92e0331fd621.png)

## Version

GitHub Desktop: 2.03
Operating system: macOS 10.14.5 (18F132)

## Steps to Reproduce

Try and switch branch with untracked files.

### Expected Behavior

Ignore untracked files like the previous versions and the command line.

### Actual Behavior

Annoying dialog every time you switch branches.

shiftkey:
@markp2 thanks for the feedback!

> Previously this was not the case and you could switch branches and carry around untracked files.

This is now that second option in the list - "Bring my changes to master". If you choose that you'll complete the flow and have your changes available there.

The context behind this new dialog is outlined in more detail in https://github.com/desktop/desktop/issues/6107, and we decided that checkout was a good point to ask the user whether they wanted to bring the changes with them, or leave them stashed with their previous branch.

markp2:
@shiftkey 

Thankyou, I see that now. 

The question is, even if I confirm that changes to be brought to master, when I end up in master, whether these files are considered as potentially part of the changes to be committed? Normal git workflow distinguished between files that have been added, and files that are untracked. This new desktop change merges those two sets and therefore appears to be a regression in desktop.

Also, this is confusing new and additional step because these are not **changes**. Git does not track these as changes and so why is the user being asked to bring changes [which are not changes] into the new branch?

ChristophRob:
quite unhappy with this new feature as well, switching branches is very annoying right now. Please consider to put a new configuration setting, which removes this extra dialog

shiftkey:
@markp2 I'm going to close this out, but let us know if you have any other questions about this area

jmonrove:
I just mentioned this similar annoyance in [6107](https://github.com/desktop/desktop/issues/6107)

I really hope an option is added to be able to de/activate magic stashing as it slows down agile workflow.

shiftkey:
@jmonrove thanks for that input. I want to provide some extra context from your comment in #6107 that I think is overlooked:

> You travel to release and merge feature_1 (stash dialog)
> You travel to master and merge release (stash dialog if you didn't stash)
> You travel back to feature_1 (time to unstash)

[...]

PR #8085, opened 2019-08-09
Carry over untracked files during branch creation that results in a conflict

### LINKED ISSUE #8084: Stash error with conflicted file upon new branch creation due to untracked files
## Description

When switching branches and carrying over changes that produce a conflict there is an edge case when untracked files are involved in the conflict. Desktop stashes the untracked file, creates the new branch, and tries to apply the stashed changes on the new branch. But if the new branch already contains a file that corresponds to the stashed untracked file, Git throws an error and doesn't allow the changes to be applied. 

image

For reference -- I pulled this out as a separate issue from https://github.com/desktop/desktop/issues/7827#issuecomment-503361669. It's a distinct bug and pertains specifically to untracked files being stashed as part of checking out a new branch.

## Version

* GitHub Desktop: Desktop-dev a6a2bf7859

* Operating system: macOS High Sierra 10.13.6

## Steps to Reproduce

1. checkout `master`
1. create branch `branch-a`
1. create new file `file.txt` on master and commit
1. checkout `branch-a` and create new file `file.txt` with conflicting changes. do not stage or commit.
1. checkout new branch `branch-b` based off of master. select option to carry over changes. (note: you also hit the error by checking out `master` instead of creating a new branch)

### Expected Behavior

`branch-b` is checked out and `file.txt` shows conflicting changes

### Actual Behavior

Error appears (see screenshot above)

## Additional Information

N/A

### Logs

N/A

### LINKED ISSUE #8200: Swapping branches with changes
## Description

Wanted to bring uncommitted changes over to a new branch but instead just lost everything.

## Version

* GitHub Desktop: 2.1.3

* Operating system: macOS Mojave 10.14.6

## Steps to Reproduce

Doesn't happen all the time but...
1. Make changes in one branch
2. Change to new branch
3. When prompted to stash changes or bring over choose bring over choose bring over
4. You will see no new changes on the new branch and when you check old branch no stashed changes are there

### Expected Behavior

I expected my changes to be in the new branch

### Actual Behavior

They were erased

### Logs

[2019-09-03.desktop.production.log](https://github.com/desktop/desktop/files/3571550/2019-09-03.desktop.production.log)

kuychaco:
## Overview

Carry over untracked files during branch creation that results in a conflict

**Closes #8084**
**Closes #8200**

### The Problem
When switching branches and carrying over changes that produce a conflict there is an edge case when newly created untracked files are involved in the conflict. This issue crops up when a new file has been added AND has not yet been staged, so it is "untracked" by Git. 

Previously, Desktop stashed the untracked file (using the `--include-untracked` flag, which takes care to preserve the untracked nature of the file), created the new branch, and tried to apply the stashed changes on the new branch. But if the new branch _already contained an untracked file with the same name_, Git throws an error `untracked working tree files would be overwritten by checkout` and doesn't allow the changes to be applied. 

Presumably, this is because Git wants to be conservative and avoid mucking with files that it hasn't been explicitly told to track via staging or committing. So when you tell it to restore an untracked file from a stash, it tries to restore the file exactly as it exists, and doesn't try to combine it with conflicting changes on the new branch. 

### Proposed Solution
But we know that the user's intention is to carry over the change they've made (whether staged/tracked, or unstaged/untracked), and expects to deal with any conflicts using the usual merge conflict resolution flow. And since the Desktop UI doesn't actually distinguish between newly created untracked and tracked (staged) files, we probably don't need to be too concerned with preserving the untracked status of the user's working directory files. 

To better meet the user's expectations, we want to complete the checkout and present the user with a merge conflict to resolve for the newly created file. This PR does so by staging any untracked files prior to stashing them. Which means that Git has no problem popping the stash and generating merge conflicts since it's dealing with changes we told it to track. And we can drop the `--include-untracked` flag when stashing since we've just staged everything.

### Alternatives considered
@outofambit and I also noted a couple of other approaches, but both seemed needlessly complex compared to the solution outlined above. In any case, here's what we considered:
Alternative 1: manually generate the conflict using git merge plumbing commands to merge the stash entry with the working directory changes
Alternative 2: create a branch from the stash and merge that with the newly created branch

TODO: 
- [x] Check with @niik to see if we're missing context on why we might not want to stage before stashing. My guess is that there's no reason, and that we use the `--include-untracked` flag when stashing because that's the only way to get untracked files into the stash, not because we want to keep the untracked status.
- [x] Remove flag for stashing untracked files
- [x] See if checkout branch needs similar staging logic (moved logic into git/stash so we're good)
- ~~[ ] Add tests?~~

**Testing**: 
* Ensured that stash unit test for handling untracked files still passed
* Manually tested using repro steps from https://github.com/desktop/desktop/issues/8084#issue-478743904
* Decided (along with @outofambit) to not add an AppStore test considering how [challenging](https://github.com/desktop/desktop/blob/709276d8e62b9510d8799fbae97d060c9dae81df/app/test/unit/app-store-test.ts#L181-L184) it would be to add a reliable test and how edge-casey and rare this bug is

## Release notes

Notes: [Fixed] Carry over untracked files during branch creation that results in a conflict


kuychaco:
@niik before investing more energy into this it’d be great to get a 30% review to validate that my UX reasoning and technical approach are sound… I got a 👍 from @billygriffin on the UX, and wanted to check in with you as well on any historical technical context that might be relevant 💭 

@nerdneha pointed me to https://github.com/desktop/desktop/issues/7720#issuecomment-499877163 which seems to validate my proposed solution

niik:
@kuychaco This was a thoroughly enjoyable writeup to read, really well done!

> Check with @niik to see if we're missing context on why we might not want to stage before stashing. My guess is that there's no reason, and that we use the `--include-untracked` flag when stashing because that's the only way to get untracked files into the stash, not because we want to keep the untracked status.

The reason I pushed for us using `--include-untracked` instead of staging untracked files before stashing wasn't out of any known problems with the staging approach but rather a philosophical concern where we've previously been bitten by hard problems whenever we've strayed from the beaten path of Git. I'm always a bit anxious when we rely too much on our own understanding of Git's inner workings because they're almost always more complex than they initially appear.

That said, if I remember the conversation correctly from a few months back the only reason that staging before stashing was proposed was that it was tricky to produce a list of which files were in a stash when the stash commit had three parents. I still think that trying to stick with the purpose-built support for untracked files and spending the extra time figuring out the correct way to produce a file list for a stash was the right call.

The situation now, however, is different in that I don't see any good workaround that would let us stay on the beaten path. It's always a shame when we can't use the tooling that's so close to what we need but c'est la vie.

:+1: To continue down this path and see if there are any caveats.

app/src/lib/git/stash.ts
```diff
@@ -113,10 +111,21 @@ export function createDesktopStashMessage(branchName: string) {
...
+  untrackedFilesToStage: ReadonlyArray<string>
 ): Promise<true> {
+  // We must ensure that no untracked files are present before stashing
+  // See https://github.com/desktop/desktop/pull/8085
+  if (untrackedFilesToStage.length) {
+    await git(
+      ['add', ...untrackedFilesToStage],
+      repository.path,
+      'stageUntrackedFilesBeforeStashing'
+    )
```
kuychaco: A few notes: 

I'm intentionally choosing to do the staging of untracked files in this method (as opposed to in AppStore prior to calling this) because it's a necessary step of stashing correctly. I just deleted a bunch of code in this file that was handling the case of untracked files in the stash, so I want it to be nearly impossible to accidentally stash without first staging all untracked files. 

After discussing with @outofambit, I decided to pass untracked files in rather than shell out to Git and get this information. The reason is that stashing already takes a long time and requires several git calls, and we want to don't want to make it even slower with another git call since we can avoid it. 

I considered staging all files with a `git add .` for the sake of both simplicity and consistency. If we're gonna muck with the user's index by staging something without consulting them, it's probably better to do so in an all-or-nothing way where we stage all of their files so there's consistency and predictability wrt what's staged. But, once again, I decided to opt for the more performant approach and stage only the untracked files. Command line profiling shows that staging individual files is faster than `git add .`

And I decided not to use the `stageFiles` method in `update-index.ts` because it does a lot of unnecessary stuff, including multiple `git update-index` calls.

app/src/lib/stores/app-store.ts
```diff
@@ -5152,6 +5156,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
   }
 
+  private getUntrackedFiles(repository: Repository) {
```
outofambit: love this little helper function

app/src/lib/git/stash.ts
```diff
@@ -216,60 +225,3 @@ function extractBranchFromMessage(message: string): string | null {
...
   return match === null || match[1].length === 0 ? null : match[1]
 }
-
-/**
- * Get the files that were changed in the given stash commit.
- *
- * This is different than `getChangedFiles` because stashes
- * have _3 parents(!!!)_
- */
-export async function getStashedFiles(
```
niik: If we remove this method won't that mean we'll break the display of any existing stashes that contain untracked files? Could we perhaps keep this around for the time being and consider moving to the simpler approach at a later date?

We could also potentially figure out whether the stash entry includes untracked files ahead of time (from `%P` in the `log` call) but that seems a tad too complicated compared to just keeping this method around for now.

outofambit: oh yes that is a super good point @niik

app/src/lib/git/stash.ts
```diff
@@ -113,10 +111,21 @@ export function createDesktopStashMessage(branchName: string) {
  */
 export async function createDesktopStashEntry(
   repository: Repository,
-  branchName: string
+  branchName: string,
+  untrackedFilesToStage: ReadonlyArray<string>
```
outofambit: i'd prefer this be `ReadonlyArray<WorkingDirectoryFileChange>` so that its harder to pass a malformed path to this function. it looks like all the callers of this are using a helper to get those paths from a list of `WorkingDirectoryFileChange`s already. we already have a `Repository` argument that is really just used for the path so this would be pretty parallel to that

app/src/lib/git/stash.ts
outofambit: restored in 3651a9f !

outofambit:
@niik this is ready for another round of review! i tested and i think we're good to go!

app/src/lib/stores/app-store.ts
```diff
@@ -5488,3 +5507,11 @@ function isPullRequestAssociatedWithBranch(
     repositoryMatchesRemote(pr.head.gitHubRepository, remote)
   )
 }
+
+function getUntrackedFiles(
```
mislav: Nit: could this method potentially belong on the WorkingDirectoryStatus class?

app/src/lib/git/stash.ts
outofambit: documented in #8345

app/src/lib/git/stash.ts
outofambit: actually restored in 6b9604e and 9f2caa6dff512fcfdc62bb0fd5407b285f25ac25

app/src/lib/stores/app-store.ts
outofambit: hmm that's a good point. we have some similar helpers in `lib/status.ts`, so i've moved it there in 1e603eb

outofambit:
i also tested this with old stashes to make sure they behaved properly and found no issues.

outofambit:
will merge this for beta when the build goes green

app/src/lib/git/stash.ts
kuychaco: Turns out that using `stageFiles` is less heavy-handed than I originally thought. We call the `updateIndex` method a few times, but all but one will be a no-op because we're passing empty file arrays. So 👍 to using `stageFiles`. 

And thanks for providing the valuable context @niik! Glad you caught this and saved us future bites by the Window's command line argument length cap 😌 

Addressed in https://github.com/desktop/desktop/pull/8363

PR #8087, opened 2019-08-09
In the checkoutBranch method, move stash logic into new method.

probablycorey:
## Overview

Closes https://github.com/desktop/desktop/issues/7609

## Description

This is a straight forward extraction of some code in the _checkoutBranch method . 

- I ended up with a method name that is pretty long `_stashBasedOnUncommittedChangesStrategy` so if you have a better name I'm all ears 🌽! 
- I removed the `enableStashing` because the feature is globally enabled!

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2907,58 +2907,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
       repository
     )
 
-    let stashToPop: IStashEntry | null = null
-    if (enableStashing()) {
-      const hasChanges = changesState.workingDirectory.files.length > 0
-      if (
-        hasChanges &&
-        uncommittedChangesStrategy.kind ===
-          UncommittedChangesStrategyKind.AskForConfirmation
-      ) {
-        this._showPopup({
-          type: PopupType.StashAndSwitchBranch,
-          branchToCheckout: foundBranch,
-          repository,
-        })
-        return repository
-      }
-
-      const { tip } = branchesState
-      const currentBranch = tip.kind === TipState.Valid ? tip.branch : null
-      if (
-        currentBranch !== null &&
-        uncommittedChangesStrategy.kind ===
-          UncommittedChangesStrategyKind.StashOnCurrentBranch
-      ) {
-        await this._createStashAndDropPreviousEntry(
-          repository,
-          currentBranch.name
-        )
-        this.statsStore.recordStashCreatedOnCurrentBranch()
-      } else if (
-        uncommittedChangesStrategy.kind ===
-        UncommittedChangesStrategyKind.MoveToNewBranch
-      ) {
-        const hasDeletedFiles = changesState.workingDirectory.files.some(
-          file => file.status.kind === AppFileStatusKind.Deleted
-        )
-        if (
-          hasDeletedFiles &&
-          uncommittedChangesStrategy.transientStashEntry === null
-        ) {
-          const stashCreated = await gitStore.performFailableOperation(() => {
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -3037,6 +3000,50 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
   }
 
+  private async _stashBasedOnUncommittedChangesStrategy(
+    repository: Repository,
+    branch: Branch,
+    uncommittedChangesStrategy: UncommittedChangesStrategy = askToStash
+  ): Promise<IStashEntry | null> {
+    const {
+      changesState,
+      branchesState: { tip },
```
kuychaco: ✨oooh nested destructuring... I forget about this

app/src/lib/stores/app-store.ts
```diff
@@ -2907,59 +2907,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    if (hasChanges && uncommittedChangesStrategy.kind === askToStash.kind) {
+      this._showPopup({
+        type: PopupType.StashAndSwitchBranch,
+        branchToCheckout: foundBranch,
+        repository,
+      })
+      return repository
     }
 
+    let stashToPop: IStashEntry | null = await this._stashBasedOnUncommittedChangesStrategy(
```
kuychaco: With this extraction, it would be awesome to make this a `const` and minimize mutation if not completely necessary. Which I think we should be able to do with:

```
index 0d7d02af4..152dbc75a 100644
@@ -2958,9 +2958,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
       // call this method again once the working directory has been cleared.
       this.statsStore.recordChangesTakenToNewBranch()
 
-      stashToPop = stashToPop || uncommittedChangesStrategy.transientStashEntry
-      if (stashToPop !== null) {
-        const stashSha = stashToPop.stashSha
+      const stashEntry =
+        stashToPop || uncommittedChangesStrategy.transientStashEntry
+      if (stashEntry !== null) {
+        const stashSha = stashEntry.stashSha
         await gitStore.performFailableOperation(() => {
           return popStashEntry(repository, stashSha)
         })
```

This'll make it easier to scan the code and know with certainty what `stashToPop` is (since it's value is assigned in only one place) and we can ensure that we don't add logic in the future that changes the value in unexpected ways.

outofambit: yeah, i agree this would be a nice improvement too! also agree its non-blocking.

fwiw this line was only moved down from earlier in the function by this PR (see https://github.com/desktop/desktop/pull/8087/files#diff-d6b76b602a5c48740d6d10deda6f71ebL2910).

app/src/lib/stores/app-store.ts
probablycorey: I didn't know if this was too weird. It's kind of hard to follow and spitting it into two lines might be more readable.

probablycorey:
> My first small request is that we let the enableStashing check live on as a guard to not show the popup when there are changes...

I'm less worried about the organic decay of feature-flag methods, but I'm deferring to your knowledge on this one!

> I think we should look into having the stashBasedOnUncommittedChangesStrategy return the transientStashEntry if there is one

This was annoying me too. I switched it around to do what you suggested. `transientStashEntry` was confusing to me, so hiding it behind a cleaner stash interface makes sense to me.

> Lastly, while we don't require documentation for private methods I feel like this one warrants it,

I couldn't come up with a great new method name, so I updated the name to `stashToPopAfterBranchCheckout` and added a comment.

Issue #6101, opened 2018-11-05
Merge hint count doesn't match actual conflicted files


tierninho:
The merge hint indicated there will be more conflicted files than what was showed in the actual merge conflict modal. 

From four to three:
<img width="448" alt="screen shot 2018-11-05 at 9 12 43 am" src="https://user-images.githubusercontent.com/14828183/48014456-95c55f80-e0db-11e8-92e4-33bc7ff0578f.png">
<img width="507" alt="screen shot 2018-11-05 at 9 12 36 am" src="https://user-images.githubusercontent.com/14828183/48014454-95c55f80-e0db-11e8-83eb-675164731ef5.png">

Reproduction may be difficult. 

In command line:
```
Unmerged paths:
  (use "git add <file>..." to mark resolution)

	both added:      UnitTestProject1/UnitTestProject1.sln
	both added:      UnitTestProject1/UnitTestProject1/Properties/AssemblyInfo.cs
	both added:      UnitTestProject1/UnitTestProject1/UnitTestProject1.csproj
```

tested in 1.5.0-beta3, mac

shiftkey:
> Reproduction may be difficult.

Yeah, and I think this comes down to `git merge-tree` not being smart enough to use the same rules as `git merge`. We've got #5568 which tracks the same problem manifesting with renames, but a good sample of a public repository with the actual commits being merged will help with reproducing, like I did in #5568 by capturing the commits associated with each branch.

```
$ git rev-parse polish-merge-status
361f587830413e05b7299d9f9932cff36aee0127
$ git rev-parse branch-list-context-menu
cbdb0f78739d21193d42eb0f87aa0cded6e505b9
```

That means I can re-run the same `git merge-tree` result at a later stage, even if the branches themselves are cleaned up.

tierninho:
I have come across an instance where the merge hint says there are conflicted files, yet when I click to resolve them, the success banner is shown because the `git merge-tree` is not being smart enough.

outofambit:
this also manifests for conflicts without markers (similar to #6123)
![conflicts without markers](https://user-images.githubusercontent.com/964912/48276768-accac100-e3fd-11e8-98f8-b8a9526070b4.gif)

shiftkey:
@outofambit is the repository in https://github.com/desktop/desktop/issues/6101#issuecomment-437424265 something you're able to find again and share? Having repeatable reproductions of merges that aren't working as expected would be ideal here to ensure I'm looking at the correct thing.

[...]

PR #8096, opened 2019-08-13
Rebase Preview Conflict Detection

### LINKED ISSUE #6960: Warn if there will be conflicts prior to starting a rebase

### LINKED ISSUE #8049: Number of merge conflicts are incorrect in merge hint
We initially opened  #7445 as a proof-of-concept for using worktrees to detect conflicts as part of a rebase, and if we are happy with shipping this to address #6960 we should apply this same approach to our existing merge conflict detection.

This will address at least two known issues with our current `git merge-tree` approach:

 - #5664 - incorrectly reporting there will be conflicts
 - #5568 - incorrectly reporting the number of conflicted files

The worktree approach is favourable here because it lets us rely on Git's behaviour when merging and rebasing, rather than using an approximation. It should also let us perform the work while not affecting the user's working directory.

outofambit:
## Overview
 
closes #6960

## Description

This PR hooks up the work done in #8006 and #8026 to the existing rebase modal flow.

I used a new (to this repo) technical pattern inspired by [redux-saga](https://redux-saga.js.org/). Long story short is that I'm using a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*#Description) pattern so that we can cancel multi-git-operation tasks (at the `yield` moments between git operations). This is particularly useful for more expensive tasks that can be fired rapidly (in this case by keying through the branch list).

Doing this allowed me to move a lot of git logic out of the react component and into the generator. I was also able to move the re-entrant logic already in place in the component into AppStore.

### Known Issues

I've seen two issues in my own testing:

* **(fix in #8135)** `GitError: A lock file already exists in the repository, which blocks this operation from completing.` sometimes you can get an error from git worktree checkout. i believe this is a sort of mid-air collision between two generators trying to access the same Git worktree. it's a little finicky to reproduce because you have to get the timing just right.
* `error: [ui] 'git apply --check -' exited with an unexpected code: 128.` : sometimes the output from `git patch` isn't valid for `git patch apply`. i've only been able to reproduce this in [this repo](https://github.com/github/noodle-al-dente).

### Future Work

I can imagine this being expanded to also cancel in-flight git operations, but that requires more consideration.

## Release notes

Notes: [Improved] More thorough conflict preview for rebasing


tierninho:
Attempted to test this PR, however after clicking `Start Rebase` the application just hung indefinitely. No logs or error messages were present.

Also, the tooltip suggested no commits...

Screen Shot 2019-08-14 at 11 57 33 AM

13890a8ce8, mac

billygriffin:
Nice!

Two questions:

1. For the known issues, are you suggesting we just leave them be and not try to mitigate them? I'm curious what the impact is if we shipped to prod and folks ran into those.

2. Since this closes #8049 does it also fix the problem where the number of merge conflicts is improperly detected? It seems like yes, but I wanted to be sure.

outofambit:
> Attempted to test this PR, however after clicking Start Rebase the application just hung indefinitely.

@tierninho thanks for finding that! i've fixed that in my own testing now, could you give it another pass?

> For the known issues, are you suggesting we just leave them be and not try to mitigate them? I'm curious what the impact is if we shipped to prod and folks ran into those.

@billygriffin oh no, not at all. i think preventing the first error (`A lock file already exists in the repository`) is a must for shipping to production. the second one im not sure about yet since i haven't been able to reliably repro, but i'm leaning towards it not being a blocker.

> Since this closes #8049 does it also fix the problem where the number of merge conflicts is improperly detected? It seems like yes, but I wanted to be sure.

@spillygriffin thanks for double-checking this! this pr doesn't solve that issue—i linked it in error. will remove now

billygriffin:
@outofambit Thanks! My understanding was that the work required to add rebase conflict detection would also solve the merge conflict issue of mismatched files in #8049 and #6101. Would that happen in a follow-on PR from this one or how are you thinking about that?

Also agreed that the first error should likely be fixed before this ships. Are you looking for help figuring that one out as part of this review process?

outofambit:
closing this while we focus on the [2.2 release](https://github.com/desktop/desktop/projects/27). will reopen when i'm ready to take this the final few steps.

Issue #7520, opened 2019-05-13
Images comparison with 2-up method scales down small images too much


DavidPDev:
## Description
Images of 60x60 are shown too tiny when compared. Also other small images (72x72) are rendered very small. Example:
![image](https://user-images.githubusercontent.com/25801199/57617778-e97c9e00-7581-11e9-9149-78100465ae22.png)

Images below or equal 59x59 size are displayed and compared properly. Example:
![image](https://user-images.githubusercontent.com/25801199/57617808-f9947d80-7581-11e9-8b58-5805e3642bd1.png)

When the size of the image is bigger (for example 128x128) GitHub Desktop then scales down a bit the image, but it's ok.
Seems to have a "hardcoded" size of >=60 to scale the images when comparing them with the "2-up" method. The other types are working without problems (Swipe, Onion skin and Difference)

## Version
* GitHub Desktop: 1.6.5
* Operating system: Windows 7 SP1 (all patches up to date).

## Steps to Reproduce
1. `Create a 60x60 png image and a 59x59 png image`
2. `Commit both images`
3. `Modify both images`
4. `Compare images with GitHub Desktop`

### Expected Behavior
Both images should be rendered in a way to be able to compare their differences

### Actual Behavior
The 60x60 is displayed too small.

## Full images attached
![image](https://user-images.githubusercontent.com/25801199/57618418-9c99c700-7583-11e9-9779-f470e951b5f4.png)

![image](https://user-images.githubusercontent.com/25801199/57618425-a4596b80-7583-11e9-98ea-84ab6699a549.png)

![image](https://user-images.githubusercontent.com/25801199/57618431-acb1a680-7583-11e9-9f3c-089da58f8a26.png)

niik:
Thanks for reporting this @DavidPDev, I especially appreciate your efforts to provide solid reproduction steps and explanatory images.

I've reproduced this behavior and have tentatively labelled it as a priority 2 bug since I think it fits in that criteria:

> Bug that affects more than a few users in a meaningful way but doesn't prevent core functions

kkayri:
Merci mon ami qu'est ce que je dois faire

DavidPDev:
If it helps... The "issue" has to do with this source file:
https://github.com/desktop/desktop/blob/249c06d51332a7254de848e1c77a60e1ae52d587/app/src/ui/diff/image-diffs/two-up.tsx#L30

[...]

PR #8103, opened 2019-08-13
Fix problem with 2-up image rendering

probablycorey:
## Overview

Closes https://github.com/desktop/desktop/issues/7520

## Description

The "two up" image diff had some trouble. The first problem was if an image's height was 60px, a `max-height` would be set to 0. The second problem was images with a large height or width would overflow out of bounds. Both of these problems can be seen in this gif.

**Before**

**After**

I cleaned this up be removing the calculated maxWidth and maxHeight values and instead relying on flexbox to handle the sizing. Much like me, this fix is close to perfect but with a caveat. If you are viewing a very large image and resize the window, the images don't resize. This also was broken on development. I tried for too long to get this to work, but am being pragmatic and throwing in the towel.

## Release notes

- Fixes 2-up diff image rendering problem.

### DIFF
--- a/app/src/ui/diff/image-diffs/modified-image-diff.tsx
+++ b/app/src/ui/diff/image-diffs/modified-image-diff.tsx
@@ -168,9 +168,6 @@ export class ModifiedImageDiff extends React.Component<
         return (
           <TwoUp
             {...this.getCommonProps(maxSize)}
-            containerWidth={
-              (this.state.containerSize && this.state.containerSize.width) || 0
-            }
             previousImageSize={this.state.previousImageSize}
             currentImageSize={this.state.currentImageSize}
           />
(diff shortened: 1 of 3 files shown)


tierninho:
Tested this fix and all looks good. 👍 Ran into one minor issue where a tall image was cut off in full-screen mode. Not worth fixing, imo, and should be somewhat expected by the user. 

Screen Shot 2019-08-13 at 2 13 18 PM

919b45d8e6, mac

niik:
Super happy we're starting to look into the image diffs, there's plenty of wacky behavior in them that we've never gotten around to investigating. This change however, seems to have introduced a new behavior when the images being compared are different in size:

image

In the previous version it seems like the before and after images are always sized the same. Is this change intended?

I put up a repository to test some of these combinations here: http://github.com/niik/image-diff-tests

probablycorey:
@niik this wasn't explicitly intended. But I think it makes sense because the relative sizes of the two images stay the same in the new version. In the new version you can kind of tell because the first red square's edges are half the length of the green square. This makes it easier to see how the images changed.

If you prefer the previous behavior let me know and I can rework the changes to handle that.

PR #8118, opened 2019-08-17
Check the commit message dialog scroll state also on input events

### LINKED ISSUE #7994: Description shadow does not update when text changed
## Description

A shadow appears above the co-authors row in the description field when there is enough text to scroll. Scrolling all the way to the bottom causes the shadow to disappear, but updating the description text so that it no longer needs to scroll does not, and causes the shadow to get stuck.

## Version

* GitHub Desktop: 2.1.0

* Operating system: Windows 10 version 1903 [10.0.18362.239]

## Steps to Reproduce

1. Focus on commit description, type enough so that a shadow appears.
2. Scroll to the top of the commit description
3. Select and delete all text in commit description

### Expected Behavior

Shadow should disappear

### Actual Behavior

Shadow remains

## Additional Information

Here's a gif:
GIF

KarstenRa:
## Overview
**Closes #7994**

## Description
To determine wether a shadow appears between the commit message text box and the "add co-authors" field, not only scroll events but also input events have to be considered. Deleting content from the commit message may lead the scrollbar to disappear (no scroll-event fired) and therefore make the shadow supefluous.


tierninho:
Looks good to me! Thanks, @KarstenRa

Issue #3855, opened 2018-01-25
GitHub Desktop names repository incorrectly and opens wrong repository


jkafkaris:
### Description

GitHub Desktop names repository incorrectly and opens wrong repository. I have 2 repos on Github Enterprise, one named VertigoRealityVive and the other VertigoReality. If I clone both to my PC they are both named VertigoRealityVive. Uninstalling GitHub Desktop and removing localdata then had both repos receive the name VertigoReality.

### Version

1.0.12

**GitHub Desktop version:** [1.0.12]

**OS version:** [Windows 10]

### Steps to Reproduce

1. Clone repo VertigoRealityVive
1. Clone repo VertigoReality
1. both repose then have same name in GitHubDesktop - though it only happens with these 2 repos and initially VertigoRealityVive was named VertigoReality, but I changed it on GitHub Enterprise settings.

**Expected behavior:** 

I expect each repo to be named what they are on GitHub Enterprise, or at least give me a way to change it.

**Actual behavior:** 

Both are named the same. They do commit to correct repo, but selecting "View on GitHub" they go to the same repo - so one is correct and one is not.

Log attached

[2018-01-25.desktop.production.log](https://github.com/desktop/desktop/files/1662310/2018-01-25.desktop.production.log)

iAmWillShepherd:
Thanks for taking the time to open an issue! 🙇 

I've sent you an email with the same instructions in case you have data you want to remain private.

I need to know the state of your database, I have a hunch on what the issue could be, but I’d like confirmation.

To get to your database:

1. Select `View -> Toggle Developer Tools` in the application menu
1. Select the `Application` tab in Chrome Dev Tools _(the tab may be located in a popup menu identified the >> icon)
1. Expand the `IndexDB` tree node on the left-most column
1. Expand the `gitHubRepositories` object store
1. Select the item labeled `cloneURL`
1. Expand the value object (right-most column) for the two repositories you are having issues with
1. Take a screenshot

Additionally, select the item labeled name from the `gitHubRepositories` object store and take a screenshot of that as well. If you have multiple keys with the same name, expand all of their value objects.

![gitHubRepositories.cloneURL](https://user-images.githubusercontent.com/1715082/35451243-336f20c8-0288-11e8-9bf1-dc196ac1522d.png
![gitHubRepositories.name](https://user-images.githubusercontent.com/1715082/35451309-64eb09d2-0288-11e8-8c0b-677871d18170.png)

AmeeRamy:
[...]

PR #8132, opened 2019-08-19
Update the remote url if it changed.

### LINKED ISSUE #8590: Alert Desktop when Repository has been renamed

shows that if you rename a repository or your username, GitHub Desktop is not aware and never knows of any further pull requests. 

### Proposed solution

It stops users from having the run the command
```
$ git remote set-url origin https://github.com/KnowledgeableKangaroo/Borum.git
```
every time they rename the repository or account username. 

### Additional context

#7947 is the issue you should refer to.

probablycorey:
## Overview

Closes part of https://github.com/desktop/desktop/issues/3855
Closes #8590 (verify that this is true before merging)

## Description
The steps to reproduce this bug are found at https://github.com/desktop/desktop/issues/3855#issuecomment-522170778.

The tl;dr is – if you rename a repository on github.com the remote url isn’t updated in the .gitconfig. This isn’t a problem until you create new repo that has the original name and the remote urls are in conflict.

My attempt to fix this involves detecting when the repository is renamed (in `AppStore.repositoryWithRefreshedGitHubRepository`) and updating the remote url in  `.gitconfig` (via `gitStore.setRemoteURL`).

**Special notes**
- If the current remote url uses a protocol other than `ssh` we don’t update the remote url. I did this based on what Markus said in https://github.com/desktop/desktop/pull/8132#discussion_r316675400. The assumption being, if the user has specified a protocol other than Desktop’s default protocol then we should trust that they know what the remote url should be.
- If the `gitStore.defaultRemote.url` is the different than
`repository.gitHubRepository.cloneURL` we don’t update the remote url. Like above, this means the user has manually changed the remote url at some point so we trust that they know what the url should be and how to updated it.
- @outofambit suggested calling  `gitStore.loadRemotes()` after the remote is set. But I tried to verify that this was a problem in my tests and couldn’t.

## Release notes

- Fixes bug caused by changing repository names via github.com

cc @kuychaco


app/src/lib/stores/app-store.ts
```diff
@@ -3136,6 +3136,20 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      repository.gitHubRepository &&
+      repository.gitHubRepository.cloneURL !== apiRepo.clone_url
+    ) {
+      await this._setRemoteURL(
+        repository,
+        gitStore.currentRemote.name,
+        apiRepo.clone_url
+      )
+    }
+
```
probablycorey: Hey @niik and @outofambit, this is @kuychaco and my attempt to fix the bug described in this comment. https://github.com/desktop/desktop/issues/3855#issuecomment-522170778. It appears like what we have fixes that bug, but before we get deeper into testing and validating it we wanted to get a sanity check to the approach. 

1. Does it make sense to update the origin url in `.git/config` (using setRemoteUrl) when the remote url changes on the GitHub side?
2. Are there instances when using `gitStore.currentRemote.name` will cause problems? Should we use `gitStore.defaultRemote.name` instead?
3. Assuming we should update the remote url, Is this the right place to update the remote url?

outofambit: @kuychaco @probablycorey this looks like a good approach to me! 

1. yep! that's ideal
2. hmm i'm not sure we want to update non-default github remote urls... there might be multiple githubby remotes for a repo (eg `origin` and `upstream`). i think we have to make sure `gitStore.defaultRemote.url` is the same as `repository.gitHubRepository.cloneURL`, and update the url as you have here.
3. i think this is the right place to do it!
4. we might need to call [`gitStore.loadRemotes()`](https://github.com/desktop/desktop/blob/f96b6930a420887923db6e2b9b008a444f515c52/app/src/lib/stores/git-store.ts#L1075-L1099) after we the call to `_setRemoteUrl` to make sure other references to that remote in memory are updated to the new url. (maybe that call should be inside of `_setRemoteUrl`?)

outofambit: i also updated the docs for `repositoryWithRefreshedGitHubRepository` while i was looking into this. going to push that into this PR if that's ok!

niik: One thing that concerns me with this approach is that we'd potentially change the remote protocol as part of this. On GitHub Enterprise administrators are able to globally enforce that all repository `clone_url` use the ssh protocol and given how suboptimal our support for ssh is at the moment that makes me nervous. Vice versa some of our users are adamant about using the SSH protocol (and accept that they have to do some work of their own to make it work) and we could potentially end up switching them back to http.

> i think we have to make sure gitStore.defaultRemote.url is the same as repository.gitHubRepository.cloneURL

:+1: This has the added benefit of us never overwriting the remote URL if the user has intentionally changed it from the default.

app/test/unit/stores/updates/update-remote-url-test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const repository = await repositoriesStore.updateGitHubRepository(
+      blankRepo,
+      '',
+      apiRepository
+    )
+    await addRemote(repository, 'origin', remoteUrl)
+    gitStore = new GitStore(repository, shell)
+    await gitStore.loadRemotes()
+
+    return { repository, gitStore }
```
probablycorey: If there is an easier way to get the repo and gitstore setup for the test I'd love to hear about it!

outofambit: i don't think i know of one 😿

this looks like a more complex case since you're dealing with github remotes

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+    return
+  }
+  if (!repository.gitHubRepository) {
+    return
+  }
+
+  const remoteUrl = gitStore.currentRemote.url
+  const updatedRemoteUrl = apiRepo.clone_url
+  const isProtocolHttps = remoteUrl && remoteUrl.startsWith('https://')
+  const usingDefaultRemote =
```
niik: I'm not quite following the logic here. Based on the naming I'd assume that this variable is supposed to hold a value indicating whether the current branch is tracking a branch on the default remote but what it seems to be checking for is merely that the current repository has a default remote and that the default remote is the same as the clone URL of the GitHubRepository which I think it'll always be since the GitHubRepository instance is assigned based on the default remote URL:

https://github.com/desktop/desktop/blob/d8665f8417b9d7b931d97f56ccaed88a139b383e/app/src/lib/stores/app-store.ts#L3167-L3169

I would expect the logic to look more like 👇 

```ts
const usingDefaultRemote =
  gitStore.defaultRemote &&
  gitStore.currentRemote.name === gitStore.defaultRemote.name
```

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!gitStore.currentRemote) {
+    return
+  }
+  if (!repository.gitHubRepository) {
+    return
+  }
+
+  const remoteUrl = gitStore.currentRemote.url
+  const updatedRemoteUrl = apiRepo.clone_url
+  const isProtocolHttps = remoteUrl && remoteUrl.startsWith('https://')
```
niik: This is checking that the current remote protocol is `https`. What I was concerned about in https://github.com/desktop/desktop/pull/8132#discussion_r316675400 was mainly that we'd be switching protocols from https to ssh given our poor support of ssh. I think perhaps a safer logic here would be to check if the protocol of the current remote url matches that of the new clone url and to only update it if they do.

```ts
protocolEquals = url.parse(remoteUrl).protocol === url.parse(updatedRemoteUrl).protocol
```

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+  const usingDefaultRemote =
+    gitStore.defaultRemote &&
+    gitStore.defaultRemote.url === repository.gitHubRepository.cloneURL
+
+  if (isProtocolHttps && usingDefaultRemote && remoteUrl !== updatedRemoteUrl) {
+    await gitStore.setRemoteURL(gitStore.currentRemote.name, updatedRemoteUrl)
```
niik: Following on what I said above about the current remote versus the default remote I think this should probably be using the default remote

```suggestion
    await gitStore.setRemoteURL(gitStore.defaultRemote.name, updatedRemoteUrl)
```

app/src/lib/stores/updates/update-remote-url.ts
outofambit:
```suggestion
  const remoteUrlUnchanged =
```

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+  // I'm not sure when these early exit conditions would be met. But when they are
+  // we don't have enough information to continue so exit early!
+  if (!gitStore.currentRemote) {
+    return
+  }
+  if (!repository.gitHubRepository) {
+    return
+  }
+
+  const remoteUrl = gitStore.currentRemote.url
```
outofambit:
```suggestion
  const remoteUrl = gitStore.defaultRemote.url
```

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+import { GitStore } from '../git-store'
+
+export const updateRemoteUrl = async (
+  gitStore: GitStore,
+  repository: Repository,
+  apiRepo: IAPIRepository
+): Promise<void> => {
+  // I'm not sure when these early exit conditions would be met. But when they are
+  // we don't have enough information to continue so exit early!
+  if (!gitStore.currentRemote) {
```
outofambit:
```suggestion
  if (!gitStore.defaultRemote) {
```

app/src/lib/stores/updates/update-remote-url.ts
outofambit:
```suggestion
```

app/src/lib/stores/app-store.ts
outofambit: maybe we can frankenstein the URL path from the API onto the protocol+hostname already configured in the existing remote? this would preserve any user configured protocol or enforced ones from an admin. i'm pretty sure we don't need to worry about hostnames changing. from looking through our githubrepository store, it looks like if the enterprise hostname were to change, it wouldn't be updated anywhere in the desktop databases.

app/src/lib/stores/updates/update-remote-url.ts
outofambit: see review comment about preserving protocol and hostname while adding new URL path from the API.

app/src/lib/stores/updates/update-remote-url.ts
outofambit: we probably don't care about the `currentRemote` in this method since we're only interested in updating `origin`.

app/src/lib/stores/updates/update-remote-url.ts
outofambit: don't want to consider the `currentRemote` for changing the `defaultRemote`

app/src/lib/stores/updates/update-remote-url.ts
outofambit: don't want to consider the `currentRemote` for changing the `defaultRemote`

app/src/lib/stores/updates/update-remote-url.ts
outofambit: +1

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  const remoteUrl = gitStore.currentRemote.url
+  const updatedRemoteUrl = apiRepo.clone_url
+  const isProtocolHttps = remoteUrl && remoteUrl.startsWith('https://')
+  const usingDefaultRemote =
+    gitStore.defaultRemote &&
+    gitStore.defaultRemote.url === repository.gitHubRepository.cloneURL
+
+  if (isProtocolHttps && usingDefaultRemote && remoteUrl !== updatedRemoteUrl) {
```
outofambit:
```suggestion
  if (remoteUrlUnchanged && remoteUrl !== updatedRemoteUrl) {
```
^ this `!==` would have to be more involved if we only want to check that the URL path is now different that it was

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!repository.gitHubRepository) {
+    return
+  }
+
+  const remoteUrl = gitStore.currentRemote.url
+  const updatedRemoteUrl = apiRepo.clone_url
+  const isProtocolHttps = remoteUrl && remoteUrl.startsWith('https://')
+  const usingDefaultRemote =
+    gitStore.defaultRemote &&
+    gitStore.defaultRemote.url === repository.gitHubRepository.cloneURL
```
outofambit: ^ this !== would have to be more involved if we only want to check that the URL path is now different that it was

kuychaco:
@tierninho this should be ready enough for QA testing! Please follow steps at https://github.com/desktop/desktop/issues/3855#issuecomment-522170778 as well as https://github.com/desktop/desktop/issues/3855 🙏

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as URL from 'url'
+
+export async function updateRemoteUrl(
+  gitStore: GitStore,
+  repository: Repository,
+  apiRepo: IAPIRepository
+): Promise<void> {
+  // I'm not sure when these early exit conditions would be met. But when they are
+  // we don't have enough information to continue so exit early!
+  if (!gitStore.defaultRemote) {
```
outofambit: nitpick:
```suggestion
  if (gitStore.defaultRemote !== null) {
```

app/src/lib/stores/updates/update-remote-url.ts
```diff
@@ -0,0 +1,51 @@
...
+    return
+  }
+  if (!repository.gitHubRepository) {
+    return
+  }
```
outofambit: if we want we could avoid this check by changing the type signature for this method to use `RepositoryWithGitHubRepository` instead of `Repository`. this may push this check up to the caller but i'm in favor of doing that for logic that is truly only for GitHub repositories.

app/src/lib/stores/app-store.ts
outofambit: we abandoned this approach after @kuychaco found that the `git@github.com` ssh urls are are not parseable by `URL.parse`.

tierninho:
Confirming all looks good in OSX. Currently blocked on Windows testing for the time being. Will circle back in due time.

PR #8150, opened 2019-08-22
Ignore global gitconfig for unit tests

### LINKED ISSUE #8078: The autosetuprebase and conflictStyle settings in ~/.gitconfig can cause tests to fail
I was having trouble getting the test suite to pass on the main branch. This was because I had these lines in my ~/.gitconfig.

```
# If these lines exists a dozen tests will fail!
[branch]
  autosetuprebase = always
```

It looks like it is problematic to rely on the user's `.gitconfig` file. A possible solution would be to unset the `HOME` env var in the test scripts, but I'm not sure how well that would work on windows.

-----

## Proposed Solution

@kuychaco @probablycorey @billygriffin and @outofambit discussed this further and agreed upon the following plan for now:
* `conflictStyle = diff3` will be resolved once #8049 is finished (and its currently in progress), so we'll let that go for now
*  `autosetuprebase = always` is a value we should set to `false` in the setup of tests that will fail without it. (see #6972 for an example of this approach)

there's still the open question of using a standard config (or a series of them) for our tests, but i think we're punting on that for now!

vilmibm:
## Overview

**Closes #8078**

## Description

We have experienced test failures when a user's local `.gitconfig` file sets things our tests don't expect. This PR forces unit tests to ignore a user's `.gitconfig` by unsetting `HOME` and `USERPROFILE` (for mac/linux and windows respectively).

While the original issue suggested punting on this kind of decision we'd like to try it in the hopes of no longer playing whack-a-mole or experiencing surprise test failures.

## Release notes

Notes: no-notes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ app/node_modules/
 *.iml
 .envrc
 junit*.xml
+*.swp
--- a/app/test/unit-test-env.ts
+++ b/app/test/unit-test-env.ts
@@ -6,6 +6,8 @@ const environmentVariables = {
   GIT_COMMITTER_EMAIL: 'joe.bloggs@somewhere.com',
   // signalling to dugite to use the bundled Git environment
   TEST_ENV: '1',
+  HOME: '',
+  USERPROFILE: '',
 }
 
 process.env = { ...process.env, ...environmentVariables }


HippieSavag3:
The phuk

PR #8157, opened 2019-08-24
Add documentation for authenticating GHE from dev build

kuychaco:
## Description

"GitHub Desktop" is an authorized OAuth app for ghe.io allowing us to login from stable and beta instances of Desktop. However, "GitHub Desktop-dev" is not authorized. So, in order to sign into GitHub Enterprise Server from our development build, we must go through a few extra steps. The process is documented here. 

Credit goes to past-@shiftkey for clearly outlining these steps in Slack, and @steveward for pointing us to them. Thanks!

[Rendered](https://github.com/desktop/desktop/blob/kuychaco-patch-1/docs/contributing/github-enterprise-auth-from-dev-build.md)

## Release notes

Notes: no-notes

/cc @vilmibm @nerdneha @billygriffin


vilmibm:
my earlier comment was kind of confused and can be disregarded, i think my only useful feedback is that mentioning ghe.io specifically is unnecessary (as it is only relevant to GitHub employees). I think the first sentence of line 5 can be removed but things can otherwise stay the same.

docs/contributing/github-enterprise-auth-from-dev-build.md
```diff
@@ -0,0 +1,14 @@
...
+
+So you want to do some testing using GitHub Enterprise in your dev build of GitHub Desktop... here's how!
+
+"GitHub Desktop" is an authorized OAuth app for ghe.io allowing us to login from stable and beta instances of Desktop. However, "GitHub Desktop-dev" is not authorized. So, in order to sign into GitHub Enterprise Server from our development build, we must go through a few extra steps:
+
+1. Find or create an organization on the GHE instance that you are an owner of
+1. [Create a new OAuth app](https://developer.github.com/apps/building-oauth-apps/creating-an-oauth-app/) in the organization. Input `x-github-client://oauth` as the authorization callback URL:
+   <img width="558" alt="New_OAuth_Application" src="https://user-images.githubusercontent.com/7910250/63631725-2ffd8200-c5e0-11e9-92e4-b2e5b61d9c89.png">
+1. Insert the Client ID and Client Secret into the [`app-info.ts` file](https://github.com/desktop/desktop/blob/e3991a8c73ab10ca12fcad23f7e367707051d985/app/app-info.ts#L28-L31)
+1. Build the app
```
mislav: Since the next step is to be performed within the newly built Desktop-dev, perhaps this step should be worded like “Build and run the Desktop-dev app” to reduce ambiguity around which "app" are we talking about?

Alternatively, this step could say “Build the Desktop-dev app” and the next step could begin with “Within Desktop-dev, go to…”

docs/contributing/github-enterprise-auth-from-dev-build.md
```diff
@@ -0,0 +1,14 @@
...
+1. Go to "Preferences", click the "Accounts" tab, and sign into GitHub Enterprise Server. Click "Continue With Browser" to proceed with the OAuth flow
+1. Try publishing a new repository using Desktop-dev for the organization associated with the new OAuth app
```
mislav: I don't think this step needs to be listed, since the developer who is reading this documentation is likely already blocked on Desktop-dev interacting with GHE during a particular operation, and once they get unblocked by completing these steps, they will already know how to test if it worked.

docs/contributing/github-enterprise-auth-from-dev-build.md
```diff
@@ -0,0 +1,14 @@
...
+1. Celebrate and go about your day :vulcan_salute:
```
mislav: In the interest in condensing the how-to docs to their bare minimum, I think that this step can be dropped while retaining the doc's full value.

PR #8163, opened 2019-08-25
VSCodium support

### LINKED ISSUE #8000: support VS Codium as external editor
VS Codium is based on VS Code, the only differences being VS Codium is VS Code without MS branding/telemetry/licensing

Rexogamer:
## Overview

**Closes #8000**

## Description

-Adds support for [VSCodium](VSCodium.com) as an editor. I also renamed the Visual Studio Code vars to be more concise, and did some refactoring in relation to Atom.

## Release notes

Notes: Adds support for VSCodium as an external editor (#8000). Thanks @Rexogamer!


Rexogamer:
It seems CircleCI/Github doesn’t seem to recognise this commit. Infuriating.

Rexogamer:
It doesn’t contain any changes, even though it does

Rexogamer:
So it seems another line is always created. That’s why there’s an error. 


_sigh_

Rexogamer:
So I’m going through the CircleCI errors and fixing them, but the second error makes no sense. I defined the path in the same way as VSCode, but it says it’s essentially not defined?

Rexogamer:
Where is the Prettier error?

Issue #1879, opened 2017-06-02
Register a custom protocol handler just for authentication


niik:
Right now our oauth apps (dev and prod) use `x-github-client://` when terminating the web OAuth flow. This leads to a race where the last app launched (on Windows, undefined on Mac) wins making it tricky to run the dev build and prod build side-by-side while testing authentication.

We should register specialized protocol handlers just for auth that differ depending on whether you use the dev build or the prod build.

Something like `x-github-desktop-auth` and `x-github-desktop-dev-auth`. We should start registering the protocols as soon as possible and then make the switch in our oauth apps. Note that this will affect GHE as well so we'll have to keep supporting oauth callback over the generic protocol for quite some time.

PR #8166, opened 2019-08-26
Make Windows protocol launcher code less brittle

### LINKED ISSUE #8142: unable to login to enterprise server using github desktop
Hi,
 i am unable to login my enterprise server using github desktop.

i am using windows 10 and github version 2.1.2. 

i tried below steps :  

1) Clone Repository from the Internet

 2) select github enterprise server click on sign in  and enter company server address and continue

3) it is opening in browser and redirecting to github desktop app and hangs there.

please help me with the issue asap.

                                 3)

### LINKED ISSUE #8154: Protocol launcher code too brittle on Windows
## Description

Desktop is not properly parsing the arguments sent by Windows and augmented by Chromium when it is involved via a protocol handler.

In this block of code:

```typescript
  if (__WIN32__) {
    // Desktop registers it's protocol handler callback on Windows as
    // `[executable path] --protocol-launcher "%1"`. At launch it checks
    // for that exact scenario here before doing any processing, and only
    // processing the first argument. If there's more than 3 args because of a
    // malformed or untrusted url then we bail out.
    if (args.length === 3 && args[1] === '--protocol-launcher') {
      handleAppURL(args[2])
    }
  } else if (args.length > 1) {
    handleAppURL(args[1])
  }
}
```

( https://github.com/desktop/desktop/blob/709276d8e62b9510d8799fbae97d060c9dae81df/app/src/main-process/main.ts#L190-L202 )

we are hardcoding `3` as the expected length of a protocol launcher payload. That payload now has length 5 as updates to electron have added two additional parameters. While we tested a quick fix of checking for length `5` and pulling the app url from `args[4]`, the correct solution is to:

- assure that `--protocol-handler` is one of the passed args
- find the arg that starts with one of our expected protocols and pass that to `handleAppURL`. i only see us using `x-github-client` but noticed we additionally register `x-github-desktop-auth` and `x-github-desktop-dev-auth`. 

If we are concerned about malicious protocol handler URLs (as the existing comment suggests we ought to be) then we can additionally check to see if there are multiple args with our protocols in them and refusing to dispatch if so.

## Version

* GitHub Desktop:
2.1.2

* Operating system:
Windows 10

## Steps to Reproduce

1. Attempt to use OAuth
2. See Desktop hang upon completing browser flow and being sent back to Desktop

### Expected Behavior
See Desktop complete the OAuth flow after the browser portion finishes

### Actual Behavior
Desktop hangs because it does not dispatch the protocol handler

kuychaco:
## Overview

Fix bug breaking OAuth login flow on Windows by improving protocol launcher code to handle the fact that with the Electron 5 upgrade, Chromium now adds extra command line arguments for the `second-instance` event (see [docs](https://electronjs.org/docs/api/app#event-second-instance)).

**Closes #8154**
Fixes #8142
Fixes https://github.com/github/ce-desktop/issues/611

## Description

With the Electron 5 upgrade, Chromium now adds extra command line arguments for the [`second-instance` event](https://electronjs.org/docs/api/app#event-second-instance). This event gets emitted during our OAuth login flow. We are now seeing two additional arguments:

```
[
  executable_path,
  "--protocol-launcher",
  "--allow-file-access-from-files",                                         <---
  "--original-process-start-time=13211099044577424",    <---
  "x-github-client://oauth/..."
]
```

Previously, our protocol launcher code expected the app url passed to be the third argument. So with the Electron upgrade the OAuth broke because we were passing the wrong argument to `handleAppURL`. 

This PR introduces more robust protocol launcher code, which checks for arguments that match one of our registered protocols without relying on argument location. 

@desktop/core a couple of notes:
* I am lacking historical context on this code and made some educated guesses, so I left some comments with questions for folks who have been around longer than I have
* for the sake of expediency, please feel free to push code directly to this branch and carry forward the work (or change what I have entirely)

/cc @vilmibm 

## Release notes

Notes: Fixes bug breaking OAuth login flow on Windows due to Electron 5 upgrade

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -2,6 +2,7 @@ import '../lib/logging/main/install'
 
 import { app, Menu, ipcMain, BrowserWindow, shell } from 'electron'
 import * as Fs from 'fs'
+import * as URL from 'url'
 
 import { MenuLabelsEvent } from '../models/menu-labels'
 
@@ -70,6 +71,19 @@ function getExtraErrorContext(): Record<string, string> {
     uptime: getUptimeInSeconds().toFixed(3),
     time: new Date().toString(),
   }
+}
+
+const possibleProtocols = new Set(['x-github-client'])
+if (__DEV__) {
+  possibleProtocols.add('x-github-desktop-dev-auth')
+} else {
+  possibleProtocols.add('x-github-desktop-auth')
+}
+// Also support Desktop Classic's protocols.
+if (__DARWIN__) {
+  possibleProtocols.add('github-mac')
+} else if (__WIN32__) {
+  possibleProtocols.add('github-windows')
 }
 
 process.on('uncaughtException', (error: Error) => {
@@ -189,12 +203,20 @@ function handlePossibleProtocolLauncherArgs(args: ReadonlyArray<string>) {
 
   if (__WIN32__) {
     // Desktop registers it's protocol handler callback on Windows as
-    // `[executable path] --protocol-launcher "%1"`. At launch it checks
-    // for that exact scenario here before doing any processing, and only
-    // processing the first argument. If there's more than 3 args because of a
+    // `[executable path] --protocol-launcher "%1"`. Note that extra command
+    // line arguments might be added by Chromium
+    // (https://electronjs.org/docs/api/app#event-second-instance).
+    // At launch Desktop checks for that exact scenario here before doing any
+    // processing. If there's more than one matching url argument because of a
     // malformed or untrusted url then we bail out.
-    if (args.length === 3 && args[1] === '--protocol-launcher') {
-      handleAppURL(args[2])
+
+    const matchingUrls = args.filter(arg => {
+      const url = URL.parse(arg)
+      return url.protocol && possibleProtocols.has(url.protocol.slice(0, -1))
+    })
+
+    if (args[1] === '--protocol-launcher' && matchingUrls.length === 1) {
+      handleAppURL(matchingUrls[0])
     }
   } else if (args.length > 1) {
     handleAppURL(args[1])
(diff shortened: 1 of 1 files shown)


app/src/main-process/main.ts
```diff
@@ -189,12 +203,20 @@ function handlePossibleProtocolLauncherArgs(args: ReadonlyArray<string>) {
...
     // malformed or untrusted url then we bail out.
-    if (args.length === 3 && args[1] === '--protocol-launcher') {
-      handleAppURL(args[2])
+
+    const matchingUrls = args.filter(arg => {
+      const url = URL.parse(arg)
+      return url.protocol && possibleProtocols.has(url.protocol.slice(0, -1))
+    })
+
+    if (args[1] === '--protocol-launcher' && matchingUrls.length === 1) {
```
kuychaco: is it reasonable to expect that `--protocol-launcher` is always the second argument?

kuychaco: And as far as guarding against malformed or untrusted urls, is it enough to check that only one argument matches has the correct protocol? I'm curious what all of the possible scenarios of malformed or untrusted urls might be.

app/src/main-process/main.ts
```diff
@@ -72,6 +73,19 @@ function getExtraErrorContext(): Record<string, string> {
   }
 }
 
+const possibleProtocols = new Set(['x-github-client'])
+if (__DEV__) {
+  possibleProtocols.add('x-github-desktop-dev-auth')
+} else {
+  possibleProtocols.add('x-github-desktop-auth')
```
kuychaco: Are we actually using `x-github-desktop-dev-auth` and `x-github-desktop-auth` yet? #1879 suggests that we intended to switch over to using them at some point, but for now afaict we're still just using `x-github-client`

PR #8202, opened 2019-09-04
Add Visual Studio Code Exploration version support

pluwen:
Just try to add Visual Studio Code Exploration version support.


niik:
Hi @pluwen, this PR has sat for a long time with CI failing. If you're interested in picking this back up let us know but since we haven't heard any other requests for supporting Visual Studio Code Exploration and considering maintaining these integrations isn't effortless we won't be picking this up ourselves.

Issue #8149, opened 2019-08-22
Onboarding tutorial: tutorial repo steps


billygriffin:
Note: this is a follow-on issue in conjunction with #8148. 

### Problem:

Many users when they first download GitHub Desktop aren't certain what things they are able to do with it and how to do those things. We try to make primarily functionality in the app as discoverable as we can (and sometimes do it better than others), but that presumes a basic level of knowledge of Git.

The goal here is to help new users get more comfortable with working with GitHub repositories locally, understand how to use GitHub Desktop with GitHub, and give people muscle memory to repeat similar steps in a real repository.

### Onboarding tutorial:

The work in this issue assumes the user has created their tutorial repository and that it has been automatically published as a private repository to GitHub.com. And they are now in this tutorial repository.

When they arrive in the tutorial, the user should be provided some indication of what context they're in and what they should expect.

The tutorial itself should consist of six steps that we think will provide them with a core workflow, and also introduce the concept of what things are typically done in an editor compared to those done in GitHub Desktop:

1. Install an editor
2. Create a branch
3. Edit a file
4. Make a commit
5. Push to GitHub
6. Create a pull request

When you complete the final step, you should have a completion state that gives you reasonable options for what to do next.

#### Considerations:

1. Some people will use an editor that is unsupported by GitHub Desktop, and we shouldn't force them to download a supported editor just to complete Step 1. Therefore they'll need to be able to skip this step and let us know they already have an editor. 

2. If they don't have a detected editor, we can provide options for them to install (Visual Studio Code and Atom). Ideally, when they go through the steps to download and install an editor, it should not require restarting GitHub Desktop in order to detect it.

3. For the "Edit a file" step, we should account in our instructions for those who do have an editor detected in Desktop and those who do not.

4. The steps must be completed in order. The only steps that should be able to be skipped are step 1 (install an editor) and step 6 (create a pull request).

5. For the first version, the way to exit the tutorial is to navigate to a different respository (or add or clone one if the tutorial repo is the only one that exists). We may include a button or link to make this a bit simpler by opening the repositories dropdown.

6. Initially, users cannot make more than one tutorial repository.

We expect that this will be done in multiple feature iterations. I'll let @ampinsk respond with the actual flow for the first iteration of this.

ampinsk:
# Tutorial designs

This outlines the core tutorial steps and exiting behavior. 

[...]

PR #8207, opened 2019-09-04
use panel css for changes sidebar

outofambit:
## Overview

found while researching panels for #8149 

## Description

- reuse some existing css / reduce css duplication

## Release notes

Notes: no-notes


tanyadigman49:
Delete all my info threads etcc coding I've been hacked and the hacker is
coding my  devices accounts external storage.do not tip off the hacker

PR #8213, opened 2019-09-05
fix merge preview conflict count with worktree approach

### LINKED ISSUE #8049: Number of merge conflicts are incorrect in merge hint

outofambit:
## Overview

Closes #8049 

## Description

-

## Release notes

Notes: **forthcoming**


outofambit:
will reopen when we focus on the 2.3 release

Issue #4592, opened 2018-05-03
Submodule Changes not Being Discarded


say25:
## Description
Submodule Changes not being discarded.

## Version
* GitHub Desktop: 1.1.2-beta4
* Operating system: macOS Sierra

## Steps to Reproduce
1. Clone GitHub Desktop Repo
2. Make a change in the Gemoji folder
3. Attempt to discard the changes.

### Expected Behavior
Changes to discard.

### Actual Behavior
Nothing.

### Logs
[2018-05-03.desktop.beta.log](https://github.com/desktop/desktop/files/1971701/2018-05-03.desktop.beta.log)

shiftkey:
@say25 thanks for the report. It looks like the app isn't able to resolve the submodules listed in the repository, because `git-submodule` is generating a segmentation fault.

```
2018-05-03T05:41:28.688Z - error: [ui] `git submodule status --` exited with an unexpected code: 139.
 fe965f16f6f690d40a738e75a46fad3f41f5daa3 app/static/common/choosealicense.com (fe965f1)
 690a433b668348c33087d34fafe2155ebe8bb4cf app/static/common/gitignore (remotes/origin/reduce-noise-1009-g690a433)

/Applications/GitHub Desktop.app/Contents/Resources/app/git/libexec/git-core/git-submodule: line 979: 12490 Segmentation fault: 11  git ${wt_prefix:+-C "$wt_prefix"} ${prefix:+--super-prefix "$prefix"} submodule--helper status ${GIT_QUIET:+--quiet} ${cached:+--cached} ${recursive:+--recursive} "$@"
```

We last updated `dugite` in `1.1.2-beta2` but I tested the same steps provided in the same build and was able to successfully discard. `git-submodule` is just a loose script that hasn't changed [since January](https://github.com/git/git/commit/2e612731b55f1a83fb5b7f4ecb9391f0cba63cb2), so I'm puzzled why it's now failing on your setup.

shiftkey:
@say25 I'm still puzzled about how this might be failing. Has it been consistent to reproduce? Any clues about your setup?

say25:
Tried it on the latest beta, here is my the log from the work flow. I can consistently reproduce it.

Steps to Reproduce:
1. Clone GitHub Desktop Repo
1. Modify Gemoji/Readme.md in any way. (add a space or word at the end of the file for example)
1. Attempt to discard the changes from the "Changes" page by either right click discarding on 
the Gemoji submodule or the discard all. 

[...]

PR #8218, opened 2019-09-05
Enable discarding submodule changes

mislav:
## Overview

Fixes https://github.com/desktop/desktop/issues/4592

## Description

Without `--force`, `git submodule update -- path/to/submodule` is not enough to discard changes in the working copy under the git submodule since git will see that the currently checked out git commit matches the one that the superproject expects and the operation will be a no-op.

Potential TODO:

- [ ] The current approach discards changes under the submodule **without** moving the copies of changed files to Trash. Should we consider moving them to Trash to avoid discarding someone's work irrevocably?
- [ ] It might be a good idea to add a test that exercises discarding changes from multiple submodules at the same time.

## Release notes

Notes:

- Enables discarding changes to files under git submodules


app/src/lib/git/submodule.ts
```diff
@@ -80,11 +80,9 @@ export async function resetSubmodulePaths(
...
   paths: ReadonlyArray<string>
 ): Promise<void> {
-  for (const path of paths) {
-    await git(
-      ['submodule', 'update', '--recursive', '--', path],
-      repository.path,
-      'updateSubmodule'
-    )
-  }
+  await git(
```
outofambit: doing this in one shell command is super awesome ✨

outofambit:
> It might be a good idea to add a test that exercises discarding changes from multiple submodules at the same time.

:+1: on this if its not too time consuming to add. (i'm not sure if we have some repository with submodules fixtures. @vilmibm and i were looking at unit test setup stuff recently so he can probably answer questions about this too.)

outofambit:
quick note on this one:

> ❌ discarding a Genmoji file when the file name is changed

this is because `git submodule update` doesn't touch untracked files. not sure what we need to do to cover that case too.

that being said, this PR is still a marked improvement on the current state of things, so i'd be in favor of merging as-is and filing a new issue for this last case.

niik:
> * The current approach discards changes under the submodule **without** moving the copies of changed files to Trash. Should we consider moving them to Trash to avoid discarding someone's work irrevocably?

I think we absolutely should consider that but it does seem like this PR improves upon one scenario while not making the trash/restore situation any worse so I still think it's valuable to get this in. I'm gonna open an issue to track this and the fact that discarding a submodule using this method still doesn't clear the dirty state when the submodule contains untracked files.

PR #8220, opened 2019-09-05
use openExternal async

### LINKED ISSUE #8156: Fix `shell.openExternal` calls now that it's an async method
With the upgrade to Electron 5 `shell.openExternal` is now an async method (see https://github.com/electron/electron/pull/16176).

So we want to update the following accordingly... https://github.com/desktop/desktop/blob/e23328a063e3c7702993c41714bdfcaa5eaf3bbe/app/src/main-process/main.ts#L483-L484

As it is now, we prematurely send the 'open-external-result' event. And we won't be alerted if this operation fails, as we're not handling the promise rejection. From what I can tell we're not actually doing anything with `result`, so nothing is breaking due to the fact that its value is now a promise rather than a boolean, as was previously returned from `shell.openExternal`.

To fix this we can either use [`shell.openExternalSync`](https://electronjs.org/docs/api/shell#shellopenexternalsyncurl-options) or await the promise that is now returned from [`shell.openExternal`](https://electronjs.org/docs/api/shell#shellopenexternalurl-options). 

Note that strangely `shell.openExternalSync` returns a `boolean` whereas `shell.openExternal` returns `Promise<void>`. I'm generally in favor of using the async method, unless we rely on the `boolean` return value for subsequent logic in our code.

There are several other places that we use `shell.openExternal` in the code, so we'll want to do a thorough audit and make sure to use the appropriate method in each case.

/cc @vilmibm

vilmibm:
## Overview

Fixes #8156 

## Description

This PR corrects our use of Electron's `shell.openExternal` since it changed to become async by default. There is only one invocation of `openExternal` where we check the result and potentially show an error; this code path should function once more as of this PR.

## Release notes

Notes: no-notes


billygriffin:
Thanks @vilmibm! 

From #8156:
> There are several other places that we use shell.openExternal in the code, so we'll want to do a thorough audit and make sure to use the appropriate method in each case.

So am I understanding correctly that we use it elsewhere, but it's only in this one instance you found that we actually need to make the change?

vilmibm:
@billygriffin 

> So am I understanding correctly that we use it elsewhere, but it's only in this one instance you found that we actually need to make the change?

We use it in many places; however, in most places we just call it and discard the result. There is only one code path that _checks_ the result, and it didn't need changing; I had to change our lower level wrapper around Electron's `openExternal` to actually return a success/fail boolean so that that one existing code path can work again.

outofambit:
@vilmibm thank you! 🎉 so from what you saw we should be safe with async behavior in places we probably assumed were sync in the past?

vilmibm:
@outofambit yup! in most places we just called it as `openExternal(url)` which would have the desired behavior async or not. in just that one place I linked[1] were we actually checking for an error result.

[1] https://github.com/desktop/desktop/blob/f37e6bf6d9a51641f3bfdf1082a3a5b166b00e1e/app/src/ui/lib/open-file.ts#L8

outofambit:
@vilmibm awesome! pretty sure that was the case, but just wanted to be sure! ✨

vilmibm:
@kuychaco seems worth a follow-up issue so we can collect UX feedback if needed. would you like to open one or should I?

billygriffin:
> seems worth a follow-up issue so we can collect UX feedback if needed. would you like to open one or should I?

@vilmibm @kuychaco If we don't feel like it's a thing we should do immediately, I'd prefer not to open an issue about it unless it surfaces in the wild. Does that seem reasonable? If you think it's worth adding logging for that type of case before merging this PR, I'm totally fine with that. Just for something that stays open and isn't super likely to get prioritized, I'd rather see it surface through a user-reported problem instead of trying to over-optimize ahead of time.

billygriffin:
Hi @vilmibm and @kuychaco! Just wanted to follow up to see if y'all wanted to do additional logging here or if this PR should otherwise be merged?

vilmibm:
@billygriffin I'm comfortable merging as is.

kuychaco:
@billygriffin I'm totally cool to merge as is and get this fix in. I would like to see us fix up the other places we use Electron's `shell.openExternal` method (like append a `.catch` with logging), because I think it's a minor and risk-free change that could potentially save us some debugging time in the future. But that can happen in a follow-on PR. I'll defer to @vilmibm's preference. And I can maybe see about fixing up the other places when I've got a bit more bandwidth.

PR #8246, opened 2019-09-11
Log `shell.openExternal` errors

### LINKED ISSUE #8156: Fix `shell.openExternal` calls now that it's an async method

kuychaco:
## Overview

This is a follow-on PR to #8220, which had **closed #8156**

## Description

Electron's `shell.openExternal` changed to become async so in order to ensure that errors don't get lost in the async void we need to explicitly catch them.

I'm choosing to log them rather than throw because surfacing the error to the user doesn't add any value. They'll notice that something is wrong by virtue of the fact that their browser didn't open correctly.

The logging will be helpful for us developers when we're debugging what went wrong and save us time identifying the root cause.

## Release notes

Notes: no-notes


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -425,32 +425,40 @@ export function buildDefaultMenu({
   const submitIssueItem: Electron.MenuItemConstructorOptions = {
     label: __DARWIN__ ? 'Report Issue…' : 'Report issue…',
     click() {
-      shell.openExternal('https://github.com/desktop/desktop/issues/new/choose')
+      shell
+        .openExternal('https://github.com/desktop/desktop/issues/new/choose')
+        .catch(err => log('error', `Call to openExternal failed: '${err}'`))
```
niik: It seems this file erroneously imports the `log` method from `log.ts` which shadows the [globally available `log` object](https://github.com/desktop/desktop/blob/284e374294289fc44d45bd115031027b3ce1f136/app/src/lib/globals.d.ts#L100-L168) (installed [here for the main process](https://github.com/desktop/desktop/blob/284e374294289fc44d45bd115031027b3ce1f136/app/src/lib/logging/main/install.ts) and [here for the renderer](https://github.com/desktop/desktop/blob/284e374294289fc44d45bd115031027b3ce1f136/app/src/lib/logging/renderer/install.ts)).

Once you've removed that import you should get access to `log.error` which takes an error message as its first parameter and an optional exception as its second. By passing the exception separately we can globally decide how to format our error messages in the log file.

I'd be looking for something similar to this

```suggestion
        .catch(err => log.error('Failed opening issue creation page', err))
```

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -425,32 +425,40 @@ export function buildDefaultMenu({
...
     label: __DARWIN__ ? 'Show Keyboard Shortcuts' : 'Show keyboard shortcuts',
     click() {
-      shell.openExternal(
-        'https://help.github.com/en/desktop/getting-started-with-github-desktop/keyboard-shortcuts-in-github-desktop'
-      )
+      shell
+        .openExternal(
+          'https://help.github.com/en/desktop/getting-started-with-github-desktop/keyboard-shortcuts-in-github-desktop'
+        )
+        .catch(err => log('error', `Call to openExternal failed: '${err}'`))
```
mislav: Nit: since there are multiple `shell.openExternal` calls in this file and each handles errors in an identical way, would we sometimes opt to DRY up these calls by creating an intermediate method to handle errors, or would we rather keep repeat these and treat is as incidental duplication?

```js
const openExternal = (...args) =>
  shell.openExternal(...args).catch(err => log('...'))
```

outofambit:
took a quick pass at this to get it across the line after discussing briefly with @kuychaco on friday. i had misunderstood 

> Electron's `shell.openExternal` changed to become async so in order to ensure that errors don't get lost in the async void we need to explicitly catch them.

PR #8255, opened 2019-09-12
Adding Xcode as external editor

msztech:
## Overview

This is a work on issue #4825 and an improvement on [previous PR](https://github.com/desktop/desktop/pull/7369).

## Description

On Mac OS, `Xcode` can now be detected as an external editor. Whenever `Xcode` has been selected as the editor, Desktop checks if the repository directory contains either an Xcode project file (`*.xcodeproj`) or an Xcode workspace file (`*.xcworkspaces`). It will open the workspace file if both presents (reasoning in Notes); if there's a project file but not a workspace file, the project file will be opened; and if neither of these exists, an alert will be presented.

## Release notes

- Added Xcode as an external editor

Notes:

Almost every Xcode project contains a project file. There're cases when a user uses CocoaPods (which will generate a workspace file) or combines different project into one workspace. And usually workspace contains multiple projects a user is working on. So the function chooses to open a workspace file at first if there exists one.

And thanks for @shiftkey 's help!


app/src/lib/editors/launch.ts
```diff
@@ -13,6 +15,31 @@ export async function launchExternalEditor(
   editor: FoundEditor
 ): Promise<void> {
   const editorPath = editor.path
+
+  //Special launching for Xcode.
+  if (editor.editor === Darwin.Xcode) {
+    const projectFiles = await findFilesMatching('*.xcodeproj', fullPath)
+    const workspaces = await findFilesMatching('*.xcworkspace', fullPath)
```
mislav: Since this is a relatively simple glob expression and there are two queries, we could potentially avoid hitting the filesystem twice and also avoid having to pull in an extra "glob" dependency:

```js
const files = await FSE.readdir(fullPath)
const projectFiles = files.filter(f => f.endsWith('.xcodeproj'))
const workspaces = files.filter(f => f.endsWith('.xcworkspace'))
```

app/src/lib/editors/darwin.ts
```diff
@@ -217,6 +225,7 @@ export async function getAvailableEditors(): Promise<
     typoraPath,
     codeRunnerPath,
     slickeditPath,
+    xCodePath,
```
mislav: Minor, but could this be `xcodePath` to avoid "c" having any special capitalization rules?

app/src/lib/editors/launch.ts
msztech: Hi, @mislav ! These sound like some great suggestions! I'll work on these now and commit to this PR today.

msztech:
@mislav Please check out the new changes I made per the suggestions. Thanks! ;)

msztech:
Also will it be a good idea to show up a button `Open in Atom` if the user has Atom installed and Xcode couldn't open the folder?

Rodion8888:
$ ssh-keygen -t rsa -b 4096 -C "your_email@example.com"

tanyadigman49:
I want you to delete my info my device all of my info.someone has hacked my
accounts device.they are blocking me from using my own device and whatever
else hackers do.please delete do not respond to the person doing all the
coding or ibwill report anyone who helps them.thank you

tanyadigman49:
Oh they are also reading all my info from my device

msztech:
> When trying to open `README.md` in a project that _doesn't_ have `*.xcodeproj` or `*.xcworkspace`, I get this error:
> 
> ```
> ENOTDIR: not a directory, scandir '/path/to/myproject/README.md'
> ```
> 
> It definitely feels misleading to me. If we do expect an error here, perhaps the error message should be friendlier? However, is it not possible to open an individual file in Xcode even if it doesn't belong to a folder that has an associated Xcode project/workspace?

It is possible and that's a great suggestion. I can make sure Xcode opens up directly if the object is a file instead of a folder. Will commit soon. @mislav

mislav:
Thank you for your ongoing work! ✨

msztech:
@mislav Change submitted:

> Code logic now checks if the path refers to a file first. If it is, it triggers Xcode directly. Otherwise, the original folder logic will apply.

Please review. Thanks! ;)

app/src/lib/editors/launch.ts
```diff
@@ -18,7 +18,12 @@ export async function launchExternalEditor(
 
   //Special launching for Xcode.
   if (editor.editor === Darwin.Xcode) {
-    //Query the file system for xCode related files
+    //Directly open if the fullPath contains a file
+    if (lstatSync(fullPath).isFile()) {
```
mislav: Nit: could we use the async `await lstat` please? I don't think there should be a reason to use sync operations in an already `async` function.

msztech: Sure. Do you think this change will work?

`if (await lstat(fullPath).isFile())`

Will push the new commit today

msztech: Changes pushed @mislav

mislav:
@msztech I have a pushed a simplification to this PR. Please review if you have a chance

msztech:
@mislav What a great idea! I never knew `xed`before! Thanks! ;)

msztech:
@kuychaco @tierninho Thanks for reviewing this PR. Glad to be able to contribute to this ;)

msztech:
Thank you ;) @kuychaco

Issue #6666, opened 2019-01-18
Truncate file names more in conflict resolution modal


ampinsk:
![image_720](https://user-images.githubusercontent.com/10404068/51416658-4f569d00-1b2f-11e9-8592-3dac0f5498c3.png)

When there are long paths these file names can get a little hairy. It would be nice to follow similar truncation patterns we have elsewhere if we aren't already, such as: 

![screen shot 2019-01-18 at 2 43 52 pm](https://user-images.githubusercontent.com/10404068/51416710-84fb8600-1b2f-11e9-8d46-39f797446912.png)

cc @outofambit

SolanPy:
It would also be nice to display a hint with the full path to the file.

xt0rted:
The two issues I have with this dialog are the file paths are truncated way too soon for my resolution (4k and 150% scaling in win10), and there's no tool tip displaying the full value.

![image](https://user-images.githubusercontent.com/831974/63688774-42d9a780-c7d6-11e9-923a-43b7d9cf9605.png)

PR #8290, opened 2019-09-18
Fix file path truncation in merge conflicts dialog

mislav:
## Overview

Fixes https://github.com/desktop/desktop/issues/6666
Followup to https://github.com/desktop/desktop/pull/5809

## Description

Long file paths were incorrectly truncated within the merge conflicts dialog: the basename component was truncated with "..." (this was via CSS) instead of the directory component. Moreover, there was no tooltip on mouse hover which revealed the full path name.

This change ensures that the path truncation logic is re-executed once the dialog is shown so that the correct layout can be applied.

Screen Shot 2019-09-18 at 9 30 06 PM

## Release notes

Notes:
- Fixes display of long file paths within the merge conflicts dialog


billygriffin:
@mislav Thanks! 🎉 Just to clarify, does this PR also introduce the behavior where hovering on a truncated file path also shows a tooltip with the full path?

app/src/ui/lib/conflicts/unmerged-file.tsx
```diff
@@ -122,7 +122,7 @@ const renderResolvedFile: React.SFC<{
     <li key={props.path} className="unmerged-file-status-resolved">
       <Octicon symbol={OcticonSymbol.fileCode} className="file-octicon" />
       <div className="column-left">
-        <PathText path={props.path} availableWidth={200} />
```
mislav: Due to “Open in \<EDITOR>”  buttons on the right side, which are of variable width depending on which editor was selected, we don't really know how much space the PathText component has available. Therefore, we leave it to be calculated instead.

app/src/ui/dialog/dialog.tsx
```diff
@@ -202,9 +202,21 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
   }
 
   public componentDidMount() {
+    if (!this.dialogElement) {
```
niik: While I doubt that it's possible for the dialogElement to be null at this point I'm still not a fan of the sentiment of bailing on the whole method if it is. Could we instead wrap an `if (this.dialogElement !== null)` around the two lines that use it?

mislav: Sure, I can do that. The reason why I chose to bail on the entire method is that, if `dialogElement` is unavailable for whatever reason, the rest of the method doesn't make sense:

1. The old code (before this PR) would encounter a runtime error due to `undefined.showModal()`;
2. `this.setState({ isAppearing: true })` wouldn't make sense since nothing is "appearing";
3. `this.scheduleDismissGraceTimeout()` also wouldn't make sense.

To keep compatibility with old behavior, would a better approach be to manually raise an exception if `dialogElement` was unavailable?

niik: I don't feel strongly enough about this. Let's keep it as-is, it has no practical effect at all either way

PR #8298, opened 2019-09-20
[Tutorial] Update resolved editor at launch

niik:
## Overview

This is a follow up to the discussion in https://github.com/desktop/desktop/pull/8242 about whether we should be resolving the external editor during startup or not.

## Description

From https://github.com/desktop/desktop/pull/8242#pullrequestreview-289936574

> Additionally I found an unfortunate side-effect of moving to only resolving the external editor on focus which is that you might not trigger the resolve at all while going through the creation flow unless you explicitly blur and focus the app. See 👇 
>
> 
>
> We probably need to do some magic to trigger the resolve at least once initially.

From https://github.com/desktop/desktop/pull/8242#issuecomment-533342932

> I'm not sure why we don't do this during app startup if no editor is configured? For now, I've added it to AppStore._addTutorialRepository in 14175f5.

From https://github.com/desktop/desktop/pull/8242#pullrequestreview-291134338

> Yeah, me neither. I think we ought to. The fix you did works well for the first time you're going through the tutorial but not if you re-enter the tutorial later. I'm not gonna hold up this PR for that though but I'll open a PR to address that. I also noticed that it doesn't seem like the resolvedExternalEditor gets updated when the user changes their editor preference. I'll open a separate PR to address this such that we can get this merged.

---

This PR does two things, first it attempts to resolve the current editor at launch and second it notifies the tutorial assessor when a new editor has been resolved (if applicable). This in turn means that there's no longer any need for the conflicts dialogs to manually ask for a resolved editor.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5203,6 +5203,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const resolvedExternalEditor = match != null ? match.editor : null
     if (this.resolvedExternalEditor !== resolvedExternalEditor) {
       this.resolvedExternalEditor = resolvedExternalEditor
+
+      // Make sure we let the tutorial assesor know that we have a new editor
+      // in case it's stuck waiting for one to be selected.
+      if (this.currentOnboardingTutorialStep === TutorialStep.PickEditor) {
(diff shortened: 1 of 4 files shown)


niik:
Gonna merge this in preparation for the beta release today.

PR #8311, opened 2019-09-23
Prevent triggering multiple tutorial creation flows in parallel

### LINKED ISSUE #8288: [onboarding] Disable Continue button if tutorial repo is loading
**When the tutorial repo is loading**, we should either disable the `Continue` button or remove it altogether as the user can keep clicking it causing errors. Secondly, I am not sure `Cancel` really does anything valuable here as the repo is still created with an error in the console:

```
warning.js:33 Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in CreateTutorialRepositoryDialog (created by App)
    in CSSTransitionGroupChild (created by TransitionGroup)
```

Screen Shot 2019-09-17 at 2 02 24 PM

d9ff175937, mac

niik:
## Overview

**Closes #8288**

## Description

I forgot to disable the dialog while the tutorial repository is being created which resulted in the end user being able to trigger multiple tutorial repo creation flows in parallel as well as being able to cancel a non-cancellable flow 🤦‍♂ 

This also ensures that we don't end up calling `setState` after the dialog has been unmounted.

### DIFF
--- a/app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
+++ b/app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
@@ -195,6 +195,8 @@ export class CreateTutorialRepositoryDialog extends React.Component<
       this.props.onTutorialRepositoryCreated(path, account, repo)
       this.props.onDismissed()
     } catch (err) {
+      this.setState({ loading: false })
+
       if (err instanceof GitError) {
         this.props.onError(err)
       } else {
@@ -204,8 +206,6 @@ export class CreateTutorialRepositoryDialog extends React.Component<
           )
         )
       }
-    } finally {
-      this.setState({ loading: false })
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
```diff
@@ -204,8 +206,6 @@ export class CreateTutorialRepositoryDialog extends React.Component<
           )
         )
       }
-    } finally {
-      this.setState({ loading: false })
```
mislav: I take it that doing this explicitly is no longer needed?

outofambit: i think this is no longer needed because `this.props.onDismissed()` is now guaranteed to be called _after_ the repo has been successfully created. calling `this.props.onDismissed()` will close this popup (unmounting the react component).

PR #8327, opened 2019-09-24
Support enforced web flow authentication on Dotcom

niik:
## Overview

For GitHub Enterprise Server we've long supported the ability for server admins to enforce web flow authentication over basic authentication. This PR is the first step towards supporting the same for GitHub.com users.

## Description

Send the user to the browser to complete their sign-in when security policies dictate that they're required to.

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -865,6 +865,12 @@ export enum AuthorizationResponseKind {
   PersonalAccessTokenBlocked,
   Error,
   EnterpriseTooOld,
+  /**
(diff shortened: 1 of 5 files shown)


niik:
This is ready for review. The easiest way to test this is to add the following snipped to the top of the `createAuthorization` method in `api.ts`.

```ts
if (oneTimePassword !== null) {
  return { kind: AuthorizationResponseKind.WebFlowRequired }
}
```

That will simulate the response we'll be getting from the API. Here's a screen cap of the new flow from the welcome wizard and from the preferences dialog.

PR #8320, opened 2019-09-23
[Onboarding tutorial] Add "Welcome" and "You're done!" panes

kuychaco:
## Overview

**Closes #8232**

<details>
<summary>screenshots</summary>

**Welcome screen**
welcome screen

**Done screen**
done screen
</details>

## Description

Add "Welcome" and "You're done!" panes to the onboarding tutorial. When there are working directory changes, the usual diff pane is displayed. When the "History" tab is selected, the usual commit info pane is displayed.

We added an svg version of a couple octicons (for the done view) since some of the images are not part of the octicon icon set.

## Release notes

Notes: no-notes

PR #8331, opened 2019-09-25
[Onboarding tutorial] Implement functionality for exiting and resuming tutorial

### LINKED ISSUE #8231: [Onboarding tutorial] Exit tutorial
(breaking up #8149 into bite-sized issues)

> For the first version, the way to exit the tutorial is to navigate to a different respository (or add or clone one if the tutorial repo is the only one that exists). We may include a button or link to make this a bit simpler by opening the repositories dropdown.

(copied from #8149)

### No other repositories
If the user has no other repositories, we'll send them back to the Let's Get Started screen, with an option to resume the tutorial

exit-no-repo

### With other repositories
If the user has other repositories, we'll pop open the repositories panel where they can navigate to another repo, or add new ones. You can resume the tutorial by going back to the tutorial repo.

exit-with-repo

kuychaco:
## Overview

**Closes #8231**

### No other repositories

If the user has no other repositories, we'll send them back to the Let's Get Started screen, with an option to resume the tutorial. When resuming the tutorial, we re-evaluate what step the user is on in case their state has changed.

exit-tutorial-only-repo

### With other repositories

If the user has other repositories, we'll pop open the repositories panel where they can navigate to another repo, or add new ones. You can resume the tutorial by going back to the tutorial repo.

exit-tutorial-other-repos

TODO: 
- [x] style button
- [x] style popup
- [x] audit the code for places where we currently rely on the size of the repositories array as the sole indicator as to whether we're currently rendering the blank slate or not

## Release notes

Notes: no-notes

/cc @outofambit


niik:
@kuychaco :wave: Just so that we don't forget in the rush towards beta and prod releases could you please add a todo list item to the PR body to audit the code for places where we currently rely on the size of the repositories array as the sole indicator as to whether we're currently rendering the blank slate or not. I'm assuming there's a few spots.

app/src/ui/app.tsx
```diff
@@ -2469,6 +2520,14 @@ export class App extends React.Component<IAppProps, IAppState> {
       kind: HistoryTabMode.History,
     })
   }
+
+  private inNoRepositoriesBlankSlateState: () => boolean = () => {
```
niik: First, 😍 , second: I don't believe we're passing this anywhere such that it needs to be bound

```suggestion
  private inNoRepositoriesBlankSlateState() {
```

app/src/ui/app.tsx
```diff
@@ -2469,6 +2520,14 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+    return this.state.repositories.length === 0 || this.isTutorialPaused()
+  }
+
+  private isTutorialPaused: () => boolean = () => {
```
niik: The same suggestion as above here that I don't see this method ever getting passed around so I don't think it needs to be bound
```suggestion
  private isTutorialPaused() {
```

app/src/lib/stores/app-store.ts
```diff
@@ -441,6 +441,19 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
   }
 
+  public _resumeTutorial(repository: Repository) {
+    if (this.currentOnboardingTutorialStep === TutorialStep.Paused) {
```
niik: Should this be the other way around?

```suggestion
    if (this.currentOnboardingTutorialStep !== TutorialStep.Paused) {
```

Without this change I'm unable to resume the tutorial from the welcome flow.

app/src/ui/tutorial/confirm-exit-tutorial.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export class ConfirmExitTutorial extends React.Component<
+  IConfirmExitTutorialProps,
+  {}
+> {
+  public render() {
+    return (
+      <Dialog
+        title="Exit tutorial"
```
niik:
```suggestion
        title={ __DARWIN__ ? "Exit Tutorial" : "Exit tutorial" }
```

app/src/lib/stores/app-store.ts
outofambit: yep, good catch thank you!

outofambit:
@niik that's a great point about the assessor needing to keep track of paused state. i've made that change in 3d60d1d

app/src/lib/stores/helpers/tutorial-assessor.ts
```diff
@@ -143,5 +148,19 @@ export class OnboardingTutorialAssessor {
     localStorage.removeItem(skipInstallEditorKey)
     this.prStepComplete = false
     localStorage.removeItem(pullRequestStepCompleteKey)
+    this.tutorialPaused = false
+    localStorage.removeItem(tutorialPausedKey)
+  }
+
+  /** Call when the user pauses the tutorial */
+  public pauseTutorial = () => {
```
niik: I don't believe this needs to be bound?

```suggestion
  public pauseTutorial() {
```

app/src/lib/stores/helpers/tutorial-assessor.ts
```diff
@@ -143,5 +148,19 @@ export class OnboardingTutorialAssessor {
...
+    this.tutorialPaused = true
+    setBoolean(tutorialPausedKey, this.tutorialPaused)
+  }
+
+  /** Call when the user resumes the tutorial */
+  public resumeTutorial = () => {
```
niik: I don't believe this needs to be bound?
```suggestion
  public resumeTutorial() {
```

outofambit:
going to merge so i can troubleshoot styling with this in #8320

Issue #8329, opened 2019-09-24
removing repo wont delete folder


JandJ101:
Description:
When remove a repository in the GitHub Desktop app, the folder is not removed when some of it files are open in another program. GH Desktop seems to ignore open file deletion warnings.

## Version
2.1.3

* Operating system:
Microsoft Windows [Version 10.0.18362.356]

## Steps to Reproduce

1. Open currently cloned repos
2. Right click on desired repo and click remove

### Expected Behavior
Gives warning that folder or files are open

### Actual Behavior
removes repo from list but leaves folder and all files in place.

billygriffin:
Thanks for the issue @JandJ101! We'll let you know if we have other questions, but looks like @mislav already has a PR up, so hopefully we can get this fixed soon.

PR #8333, opened 2019-09-25
After failing to move a repository to Trash, show error and keep the repository listed in Desktop

mislav:
## Overview

Fixes https://github.com/desktop/desktop/issues/8329

## Description

When removing a repository with the "move the repository to Trash" option being checked, the Electron `moveItemToTrash()` operation could fail (e.g. if affected files are open in another application). This change ensures that the error is handled:
- the repository won't be removed from Desktop;
- the error will be shown in a modal dialog.

## Release notes

Notes:
- Handle case where moving a repository to Trash has failed


app/src/ui/app.tsx
```diff
@@ -1011,11 +1011,18 @@ export class App extends React.Component<IAppProps, IAppState> {
...
     deleteRepoFromDisk: boolean
   ) => {
-    this.props.dispatcher.removeRepositories([repository], deleteRepoFromDisk)
+    try {
+      await this.props.dispatcher.removeRepositories(
+        [repository],
+        deleteRepoFromDisk
+      )
+    } catch (error) {
+      this.props.dispatcher.presentError(error)
```
mislav: Is this the right way to ensure that the error is shown?

niik: I think you want the `postError` method instead. As per the `presentError` documentation:

https://github.com/desktop/desktop/blob/c080afb5777faa4c2d2a898023ddb56872955067/app/src/ui/dispatcher/dispatcher.ts#L555-L560

tierninho:
@mislav had difficulty triggering this fix. I opened a repo in terminal, finder and Atom, then removed it / trashed it, yet the repo was still removed with no warning. Am I missing something? Thanks.

mislav:
> had difficulty triggering this fix.

@tierninho I should have noted in my PR that I also never managed to reproduce this on macOS. I kept various files open in other apps, and yet deleting the repository always worked. Note that the person who reported this was running Windows 10, so it might be a platform-specific error.

I did notice that we were not handling the result of `moveItemToTrash()` at all, so this PR does exactly that, in anticipation that this method may result in an error for whatever reason. I've tested my work by manually replacing the invocation to that method with something that raises an error.

niik:
Hey @mislav, just checking in to see if there's anything I can help out with here? If you don't have the bandwidth to work on this right now I'd be happy to help usher it across the finish line.

niik:
Hey @mislav, I took the liberty of making those changes I outlined above. They're pretty minor and doesn't affect the functionality of the PR itself. I'm gonna see if I can try this out on Windows and then we should be good to go

niik:
So I wasn't able to trigger this on Windows. The docs for [moveItemToTrash](https://github.com/electron/electron/blob/a1eceee48fc20e591f1137c0a0af1c29a04a5c6b/docs/api/shell.md#shellmoveitemtotrashfullpath-deleteonfail) says that the return value will tell us whether we were able to delete or not 

> Returns Boolean - Whether the item was successfully moved to the trash or otherwise deleted.

I couldn't get it to ever return false though. Here's what I did

1. Create a repository
2. Open the Command Prompt (cmd.exe) and cd into the repo root
3. Try removing the repository using Explorer and verify it fails

image


4. Try deleting the repository and move it to trash in Desktop

## Expected result

See an error about the removal failing

### Actual result

Desktop silently suppresses the error but the folder isn't moved to trash.

@mislav I suspect there's something fishy going on in Electron here. https://github.com/electron/electron/issues/15370 and  https://github.com/microsoft/vscode/issues/43819 are likely related.

niik:
Finally got this to work after #10532 was merged

image

niik (approved):

Issue #8215, opened 2019-09-05
Version link in 'About GitHub Deskop' is not clickable


zeke:
## Description

`Version 2.1.3-beta` looks like a clickable link, but nothing happens when I click it. All the other links in the dialog work (some open the browser, others open another dialog in the app).

## Version

* GitHub Desktop: 2.1.3-beta1
* Operating system: macOS Mojave 10.14.6

## Screenshot

![Screen Shot 2019-09-05 at 1 34 16 PM](https://user-images.githubusercontent.com/2289/64378961-be133880-cfe2-11e9-8140-f8fb6c86fc77.png)

zeke:
Ah I think I see what's up. Looking in the inspector:

```html
<a class="link-button-component version-text" href="" title="Click to copy">Version 2.1.3-beta1</a>
```

It does indeed copy the version number to my clipboard, but this is not evident in the UI (unless I hover until the title text appears)

PR #1948, opened 2017-06-08
Make the version text copiable

j-f1:
Fixes #1935.

PR #8334, opened 2019-09-25
Make app's version string selectable text instead of a special link

mislav:
## Overview

Fixes https://github.com/desktop/desktop/issues/8215
Followup to https://github.com/desktop/desktop/pull/1948

## Description

The version string in the About dialog looks like other links in the same dialog (blue, underline on hover), but is not really a hyperlink. Instead, clicking on it puts the version number to clipboard.

This change makes the version text selectable instead and relies on the user to select it and copy to clipboard manually if they so desire. On macOS, for example, the text can be triple-clicked in order to be selected in full without having to make a precise selection with a mouse (this is OS-wide behavior).

This aims to make the About dialog a little more intuitive.

### After
Screen Shot 2019-09-25 at 4 13 18 PM

## Release notes

Notes:
- Make the application version string in the About dialog be selectable text

### DIFF
--- a/app/src/ui/about/about.tsx
+++ b/app/src/ui/about/about.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react'
-import { clipboard } from 'electron'
 
 import { Row } from '../lib/row'
 import { Button } from '../lib/button'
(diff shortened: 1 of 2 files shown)


zeke:
Thanks for the fix! ✨

PR #8335, opened 2019-09-25
Pre-fill user name and email in configure Git step in Welcome Wizard (again)

### LINKED ISSUE #8323: Git configuration fields in onboarding are not pre-filled for new GitHub users

This came up during onboarding usability tests for the tutorial, but is before the user gets to the tutorial page. It also may help improve things for #610, where many commits in Desktop are not attributed to a GitHub user.

When someone new signs up for a GitHub account and subsequently downloads Desktop, the "Configure Git" step in onboarding is highly confusing because it doesn't pre-fill either of the fields properly unless they've done more manual setup in their GitHub.com settings, which is highly unlikely. These are the fields:

Name:
Email:

Ideally, it would be nice to "just do the right thing" and commit with the right email address depending on which GitHub account the person is logged in with. However, for now, we could do something less elegant but still provide some sane defaults.

Name: (Pre-fill the user's GitHub handle if they have no real name in GitHub)
Email: (Pre-fill the user's noreply GitHub email)

One consideration here is how we handle Enterprise, so we'll have to think through that a bit more.

niik:
## Overview

**Closes #8323**

## Description

Due to a race condition introduced sometime after we initially shipped the welcome wizard the name and email fields in the Configure Git step in the welcome wizard were no longer pre-filled with information from the user's account.

This PR handles takes care of the race condition and expands the logic for determining the most likely candidate for preferred email address to also support stealth (aka no-reply/anonymous) email addresses for GitHub Enterprise Server.

### Testing notes

What I've been doing to test this is to explicitly unset the globally configured `user.name` and `user.email` configurations. I've been using this combo command `git config --global --unset user.name && git config --global --unset user.email`.

After clearing the config I make sure to sign out of all accounts in the application (usually by deleting the `users` entry in local storage) and then delete the `has-shown-welcome-flow` local storage setting in order to force the app into the welcome wizard again.

After signing in make sure that the name and email matches what you'd expect given your user profile settings on Dotcom.

## Release notes

Notes: [Fixed] Git configuration fields in onboarding were not pre-filled from user's profile

### DIFF
--- a/app/src/ui/lib/configure-git-user.tsx
+++ b/app/src/ui/lib/configure-git-user.tsx
@@ -84,7 +84,7 @@ export class ConfigureGitUser extends React.Component<
         }
 
         if (this.state.email.length === 0) {
-          const preferredEmail = lookupPreferredEmail(account.emails)
+          const preferredEmail = lookupPreferredEmail(account)
           if (preferredEmail) {
             this.setState({
               email: preferredEmail.email,
--- a/app/src/ui/lib/configure-git-user.tsx
+++ b/app/src/ui/lib/configure-git-user.tsx
@@ -65,7 +65,7 @@ export class ConfigureGitUser extends React.Component<
       globalUserEmail,
       name: prevState.name.length === 0 ? globalUserName || '' : prevState.name,
       email:
-        prevState.email.length === 0 ? globalUserEmail || '' : prevState.name,
+        prevState.email.length === 0 ? globalUserEmail || '' : prevState.email,
     }))
   }
 
--- a/app/src/ui/lib/configure-git-user.tsx
+++ b/app/src/ui/lib/configure-git-user.tsx
@@ -76,22 +76,25 @@ export class ConfigureGitUser extends React.Component<
     ) {
       if (this.props.accounts[0] !== prevProps.accounts[0]) {
         const account = this.props.accounts[0]
+        this.setDefaultValuesFromAccount(account)
+      }
+    }
+  }
 
-        if (this.state.name.length === 0) {
-          this.setState({
-            name: account.name || account.login,
-          })
-        }
+  private setDefaultValuesFromAccount(account: Account) {
+    if (this.state.name.length === 0) {
+      this.setState({
+        name: account.name || account.login,
+      })
+    }
 
-        if (this.state.email.length === 0) {
-          const preferredEmail = lookupPreferredEmail(account)
-          if (preferredEmail) {
-            this.setState({
-              email: preferredEmail.email,
-              avatarURL: this.avatarURLForEmail(preferredEmail.email),
-            })
-          }
-        }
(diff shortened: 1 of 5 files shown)


billygriffin:
Updated the release notes with a suggestion @niik, but not wedded to it if you'd like to change it at all - thank you! 🎉

app/src/ui/lib/configure-git-user.tsx
```diff
@@ -42,27 +45,64 @@ export class ConfigureGitUser extends React.Component<
...
+    this.state = {
+      globalUserName: null,
+      globalUserEmail: null,
+      name: '',
+      email: '',
+      avatarURL: null,
+    }
   }
 
   public async componentWillMount() {
```
outofambit: could the contents of this method go in the constructor as part of initializing state? `componentWillMount` is a deprecated react lifecycle method so it would be nice to remove it.

the `prevState` in `setState` here should just be whatever was set in the constructor.

app/src/ui/lib/configure-git-user.tsx
```diff
@@ -42,27 +45,64 @@ export class ConfigureGitUser extends React.Component<
...
+        prevState.email.length === 0 ? globalUserEmail || '' : prevState.email,
+    }))
+  }
 
-    const user = this.props.accounts[0]
-    if ((!name || !name.length) && user) {
-      name = user.name && user.name.length ? user.name : user.login
+  public componentDidMount() {
+    if (this.props.accounts.length > 0) {
+      this.setDefaultValuesFromAccount(this.props.accounts[0])
```
kuychaco: I'm curious about this accounts array... don't have time at the moment to investigate it, but what I'm wondering is how we can be so sure that the first element is what we want? Same for below in `componentDidUpdate`

niik: This is old logic but the thought is that unless you're fiddling with `localStorage` there's no way you can run the welcome wizard twice and the welcome flow is designed such that when you're done you're signed in to one account. It's technically possible to sign in to two accounts in the welcome flow but chances that you'd end up doing that is pretty slim so picking the first account seemed safe enough at the time and we haven't had any complaints that I know of regarding that logic.

app/src/ui/lib/configure-git-user.tsx
niik: Strictly speaking, no, they can't go into the constructor because of the async nature of the method. We can't use `setState` until the component is mounted and unless we implement another anti-pattern (isMounted) we won't know that. It would also be unsafe to update `this.state` again for the same reason. So it's a chicken and egg problem. Technically the same can be said about what we've got now though because we don't unsubscribe from the promises when the component gets unmounted but that's a problem we've got all over the code base.

Your point about `componentWillMount` is well taken though and I can update it to use `componentDidMount` instead.

PR #8340, opened 2019-09-26
Request additional workflow scope on GitHub.com

### LINKED ISSUE #7079: Can't push from GitHub desktop with Actions
We have recently started using Actions on our GitHub repository.
I just started trying out GitHub desktop and fell at the first hurdle

When I try to push I get an error ending with

>(refusing to allow an integration to create .github/main.workflow

I have made no changes to that file in my commit

niik:
## Overview

**Closes #7079**

## Description

It's not possible to push a commit which affects a workflow file using just the `repo` and `user` scopes which Desktop has requested in the past. This PR adds support for a new `workflow` scope which will grant permission to push workflow files.

## Release notes

Notes: TBD

cc @billygriffin


app/src/lib/api.ts
```diff
@@ -71,8 +71,14 @@ if (!ClientID || !ClientID.length || !ClientSecret || !ClientSecret.length) {
...
-/** The OAuth scopes we need. */
-const Scopes = ['repo', 'user']
+/** The OAuth scopes we want to request from GitHub.com. */
+const DotComOAuthScopes = ['repo', 'user', 'workflow']
+
+/**
+ * The OAuth scopes we want to request from GitHub
+ * Enterprise Server.
+ */
+const EnterpriseOAuthScopes = ['repo', 'user']
```
mislav: Could we try requesting the `workflow` scope for Enterprise too and fall back to `repo,user` if that results in an error? That would enable compatibility with future Enterprise versions that support the `workflow` scope, and we wouldn't have to maintain separate code paths for dotcom and Enterprise.

This is the HTTP 422 response when invalid scopes `ramen` and `cereal` were specified in addition to some valid scopes:

```json
{
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "OauthAccess",
      "code": "custom",
      "field": "scopes",
      "message": "scopes are invalid: ramen and cereal"
    }
  ],
  "documentation_url": "https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization"
}
```

If this is something that's better suiter for a followup PR, that's fine. I was just wondering whether fowards compatibility was planned. Thank you!

niik: I think, given that web flow is slated to become the primary auth mechanism soon (#8327) and given that Enterprise Server the launch post for GitHub Actions said that Enterprise support was slated for "next year" I don't think it's worthwhile to invest in, and increase the complexity of, the `createAuthorization` method for this release.

zeke:
Yay!!!

PR #8358, opened 2019-09-27
Don't say that the repository already exists unless we're sure about it

### LINKED ISSUE #8351: [onboarding] show a generic error popup for errors that occur during tutorial repo creation
the main idea here is to make sure we don't mislead people with an specific, but inaccurate error message.

i'm think a standard "something went wrong" popup with the message from whatever `Error` was thrown.

niik:
## Overview

**Closes #8351**

## Description

Only show an error claiming that the reason tutorial repository creation failed was due to the user already having a repository named `desktop-tutorial` if that's really the reason. See https://github.com/desktop/desktop/issues/8351#issuecomment-535977512

**Repository already exists**

image

**Any other error**

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
+++ b/app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
@@ -23,6 +23,7 @@ import {
   executionOptionsWithProgress,
 } from '../../lib/progress'
 import { Progress } from '../../models/progress'
+import { APIError } from '../../lib/http'
 
 interface ICreateTutorialRepositoryDialogProps {
   /**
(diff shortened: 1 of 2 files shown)


app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
```diff
@@ -114,12 +115,30 @@ export class CreateTutorialRepositoryDialog extends React.Component<
...
-          `"${name}" on your account at ${friendlyEndpointName(account)}.\n\n` +
-          'Please delete the repository and try again.'
-      )
+      if (
+        err instanceof APIError &&
+        err.responseStatus === 422 &&
+        err.apiError !== null
+      ) {
+        if (err.apiError.message === 'Repository creation failed.') {
+          if (
```
mislav: Could the 3 nested `if`s be combined into one large condition, or is it preferrable to keep them separate for readability?

kuychaco: [non-blocking] Perhaps we can assume that we will only get the `name already exists on this account` message if repository creation failed, so we can drop the outer conditional and simply rely on the inner conditional check?

kuychaco: Oh oops, got trigger happy and merged before giving @niik a chance to weigh in on this... Still curious to hear thoughts, and can make changes in a separate PR if we decide to

PR #8360, opened 2019-09-27
[Onboarding tutorial] Add metrics

### LINKED ISSUE #8233: [Onboarding tutorial] Incorporate metrics
Did the user create a tutorial repo?
Record the last/furthest step the user completed in the tutorial

/cc @billygriffin

kuychaco:
## Overview

**Closes #8233**

## Description

Add the following metrics (booleans that indicate a step has been completed): 
* tutorialStarted (button clicked in popup after user has confirmed that they want to create a tutorial repo)
* tutorialRepoCreated (repo successfully created and pushed)
* tutorialEditorInstalled
* tutorialBranchCreated
* tutorialFileEdited
* tutorialCommitCreated
* tutorialBranchPushed
* tutorialPrCreated
* tutorialCompleted (final step of tutorial has been completed)

## Release notes

Notes: no-notes

/cc @outofambit @billygriffin


app/src/lib/stats/stats-store.ts
```diff
@@ -912,7 +921,7 @@ export class StatsStore implements IStatsStore {
   /**
    * Increments the `createPullRequestCount` metric
    */
-  public recordCreatePullRequest(): Promise<void> {
+  public recordCreatedPullRequest(): Promise<void> {
```
billygriffin: This name change seems like it might just be cleanup for consistency? Just wanted to make sure it's not going to impact anything in the actual metrics before/after this change.

kuychaco: Oops! That was totally a mistake. Thanks for catching it @billygriffin! Fixed in 8ad060c and bfba4ef

kuychaco: Oops! That was totally a mistake. Thanks for catching it @billygriffin! Fixed in 8ad060c and bfba4ef

app/src/lib/stores/app-store.ts
```diff
@@ -438,10 +438,43 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
       this.currentOnboardingTutorialStep = currentStep
+      this.recordTutorialStepCompleted(currentStep)
       this.emitUpdate()
     }
   }
 
+  private recordTutorialStepCompleted(step: TutorialStep): void {
+    switch (step) {
+      case TutorialStep.PickEditor:
+        this.statsStore.recordTutorialRepoCreated()
```
niik: I was initially confused about why this method was called from here. It clicked eventually but I wonder if it would be more obvious and straightforward if we could call `recordTutorialRepoCreated` from `onNewTutorialRepository` rather than relying on this Step to be set.

app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
```diff
@@ -148,6 +151,8 @@ export class CreateTutorialRepositoryDialog extends React.Component<
   }
 
   public onSubmit = async () => {
+    this.props.dispatcher.recordTutorialStarted()
```
niik: Is the intention behind having both `tutorialStarted` and `tutorialRepositoryCreated` to be able to differentiate users where the tutorial repository creation fails? If it is I feel like a dedicated metric would be more understandable and if it isn't I wonder if we could make do with just one of them given that they both get set during creation.

Furthermore, if we could make do with just one of them I think we should be able to call that method from the `onNewTutorialRepository` method on the tutorial assessor and thereby avoid passing the dispatcher to the `CreateTutorialRepositoryDialog` component entirely (and also get rid of the two dispatcher methods).

app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
```diff
@@ -193,6 +198,7 @@ export class CreateTutorialRepositoryDialog extends React.Component<
 
       this.setProgress('Finalizing tutorial repository', 0.9)
       await this.props.onTutorialRepositoryCreated(path, account, repo)
+      this.props.dispatcher.recordTutorialRepositoryCreated()
```
niik: Is this redundant given that it's already set from the tutorial assessor?

app/src/lib/stores/app-store.ts
kuychaco: Yup, dropped `recordTutorialRepoCreated` here

app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
kuychaco: @niik yup, you've got the intention. My logic was if we have attrition between the user clicking the button and completing the first step, we'd want to know how many we're losing due to repo creation versus editor installation. I was imagining a sort of waterfall chart that shows the fall-off from clicking the button (tutorialStarted) to getting to the welcome page/install editor step (tutorialRepoCreated) to getting to the second step (tutorialEditorInstalled), etc. 

All that said, what your describing as an alternative also sounds reasonable -- a dedicated metric for repo creation failure plus just one of either tutorialStarted or tutorialRepoCreated, and dropping the dispatcher methods. From an implementation standpoint that might be cleaner. From a metric-interpreting standpoint, for some reason my mind finds the current implementation more straight-forward/easier to reason about. 

That said, I don't have a strong preference either way. Nor does @outofambit. We discussed things for a bit, and ultimately decided to leave the call up to you. Please feel free to make changes as you see fit. We thought about cutting a branch and guessing at the changes you suggested to give you options to choose from, but ran out of time to implement them, and figured it (hopefully) wouldn't take you too long to make the changes you want (and they'd be exactly what you want rather than our interpretation). Whatever you decide on, we support you :)

PR #8363, opened 2019-09-27
[Stashing] Use 'update-index' to stage untracked files

### LINKED ISSUE #8345: use `stageFiles` to stage untracked files for stashing
> And I decided not to use the `stageFiles` method in `update-index.ts` because it does a lot of unnecessary stuff, including multiple `git update-index` calls.

One of the reasons we use `update-index` instead of `add` in the app is because `update-index` lets us pass the paths to add/remove/etc over stdin as opposed to add which only lets us pass them as command line arguments. On Windows there's a hard cap for the combined length of command line arguments (see for example https://github.com/desktop/desktop/issues/1668) which we've been bitten by a number of times in the past leading to us doing our very best to avoid passing paths and such as command line args.

I think we can merge this without making that change given that we're close to the release but if we end up doing a beta release on Monday it would be swell to have this leverage `update-index` instead. I think we could just export the `updateIndex` method from `update-index.ts` and leverage that here if going through `stageFiles` feels too heavy-handed.

_Originally posted by @niik in https://github.com/desktop/desktop/pull/8085_

kuychaco:
## Overview

**Closes #8345**

## Description

As @niik kindly pointed out in #8085, on Windows there's a hard cap for the length of command line arguments, so we avoid using `git add` which requires us to pass file names as command line arguments. `git update-index` lets us pass files names as stdin so let's use that instead. 

See https://github.com/desktop/desktop/pull/8085/files/ceaeaa8c7a37202c9b3e099bed383e0ade224f16#discussion_r328512068 for reference.

## Release notes

Notes: no-notes


app/src/lib/git/stash.ts
```diff
@@ -119,14 +120,7 @@ export async function createDesktopStashEntry(
...
   // See https://github.com/desktop/desktop/pull/8085
-  if (untrackedFilesToStage.length > 0) {
-    const untrackedFilepaths = untrackedFilesToStage.map(uf => uf.path)
-    await git(
-      ['add', '--', ...untrackedFilepaths],
-      repository.path,
-      'stageUntrackedFilesBeforeStashing'
-    )
-  }
+  await stageFiles(repository, untrackedFilesToStage)
```
niik: I believe this will fail unless the user has explicitly checked the checkboxes for the untracked files. I think we need to tell `stageFiles` that we want to stage all contents of the untracked files.

```suggestion
  await stageFiles(
    repository,
    untrackedFilesToStage.map(x => x.withIncludeAll(true))
  )
```

If we don't the file will be treated as a partial commit (which I believe is a bug in and of itself but that's a separate story) and GitHub will attempt to apply a zero-length patch.

image

kuychaco: Oh my! Great catch @niik! Fixed in 85472ba and documented in 8f0073b 👍

niik:
Thanks a bunch for fixing this @kuychaco!

PR #8364, opened 2019-09-27
[Onboarding tutorial] Retain commit summary placeholder in tutorial repo

### LINKED ISSUE #8354: Refine language to make commit step more clear
**MVP method:**
Because people skip over the summary field, we should see if we can improve the language of the commit step to make this more clear

To do:
- [ ] Write new copy (@ampinsk)

**Optional but preferred method:**
Modify commit box logic to require commit message for change made during tutorial, and use default placeholder copy

kuychaco:
## Overview

TODO
- [ ] update tutorial commit step copy accordingly 

**Closes #8354**

## Description

Ordinarily, when a single file is selected, we use a default commit summary "Update README.md" and allow users to commit without supplying their own commit message.

GitHub_Desktop

However, @billygriffin uncovered during usability testing of the onboarding tutorial that people were skipping over the summary field when making a commit. 

To make it more things more clear to the user where to input their commit message, this PR makes it so that we do not add a default commit summary when in a tutorial repository. This allows us to point the user to the "summary" input box. The commit button will be disabled until the user enters a summary.

GitHub_Desktop-dev

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -530,9 +530,9 @@ export class ChangesList extends React.Component<
 
   private getPlaceholderMessage(
     files: ReadonlyArray<WorkingDirectoryFileChange>,
-    singleFileCommit: boolean
+    prepopulateCommitSummary: boolean
   ) {
-    if (!singleFileCommit) {
+    if (!prepopulateCommitSummary) {
       return 'Summary (required)'
     }
 
@@ -598,7 +598,8 @@ export class ChangesList extends React.Component<
     const filesSelected = workingDirectory.files.filter(
       f => f.selection.getSelectionType() !== DiffSelectionType.None
     )
-    const singleFileCommit = filesSelected.length === 1
+    const prepopulateCommitSummary =
+      filesSelected.length === 1 && !repository.isTutorialRepository
 
     return (
       <CommitMessage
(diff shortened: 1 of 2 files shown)


app/src/ui/changes/changes-list.tsx
```diff
@@ -530,9 +530,9 @@ export class ChangesList extends React.Component<
 
   private getPlaceholderMessage(
     files: ReadonlyArray<WorkingDirectoryFileChange>,
-    singleFileCommit: boolean
+    prepopulateCommitSummary: boolean
   ) {
-    if (!singleFileCommit) {
+    if (!prepopulateCommitSummary) {
```
mislav: Nit: this is the only place where I feel that the new change is a slight setback in terms of readability. The rest of this method body doesn't make sense unless there is only 1 file in the commit, so the original guard `if (!singleFileCommit) return` made perfect sense. Now that the condition is called `prepopulateCommitSummary`, I feel like the new name (by itself) doesn't communicate forward that this is just a more specific condition of `filesSelected.length === 1` at its core.

This is just me thinking out loud and shouldn't be taken as a blocker for this PR.  👍

outofambit:
> update tutorial commit step copy accordingly

let's address this todo in a separate PR so figuring out the copy doesn't delay merging this logic change. cc @ampinsk @billygriffin 

gonna merge this now

PR #8370, opened 2019-09-30
[onboarding] metrics revision + addition

outofambit:
## Overview

#8233 #8360

This PR revises some metrics logic in #8360 and adds an additional measure.

## Description

- https://github.com/desktop/desktop/commit/330eb741ce42feb69e45f01da4ea272ec4538261 refactors `recordTutorialStepCompleted` to ensure that the individual step booleans are marked off properly when a user completes more than a single step between runs of `updateCurrentTutorialStep`. unfortunately this implementation loses the exhaustive check of `assertNever`, but it does guarantee our metric data to be more accurate.
- https://github.com/desktop/desktop/commit/d601338b69dfb706b322b716be85e034bba2524b adds the additional measure `highestTutorialStepCompleted`, which is the index of the highest step completed. this captures essentially the same information as the `boolean`s from #8360, but will be much easier to query from the analytics side (no `AND`ing up to 6 booleans). **i'm proposing we include both.** if we change order of tutorial steps, `highestTutorialStepCompleted` will no longer provide a meaningful index to _which_ step the user last completed (for example, if step 3 becomes push branch instead of edit readme, `highestTutorialStepCompleted` == 3 would have different implications over time. so we could use the `highestTutorialStepCompleted` as a measure of attrition/retention over the tutorial, and the individual `boolean` measure to drill down for more information.

this PR is split into two commits so https://github.com/desktop/desktop/commit/d601338b69dfb706b322b716be85e034bba2524b can easily be reverted should we decide against including both kinds of metrics.

cc— @billygriffin @kuychaco @niik 

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -240,7 +240,11 @@ import { arrayEquals } from '../equality'
 import { MenuLabelsEvent } from '../../models/menu-labels'
 import { findRemoteBranchName } from './helpers/find-branch-name'
 import { findBranchesForFastForward } from './helpers/find-branches-for-fast-forward'
-import { TutorialStep } from '../../models/tutorial-step'
+import {
+  TutorialStep,
+  isValidTutorialStep,
+  orderedTutorialSteps,
+} from '../../models/tutorial-step'
 import { OnboardingTutorialAssessor } from './helpers/tutorial-assessor'
 import { getUntrackedFiles } from '../status'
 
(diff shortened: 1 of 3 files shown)


billygriffin:
The thinking behind including both makes sense to me @outofambit. Thank you for describing the logic here in such clear terms. ✨

app/src/lib/stores/app-store.ts
```diff
@@ -444,34 +448,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-        assertNever(step, 'Unaccounted for step type')
+    // if it's not a valid step, we don't need to record anything
+    if (!isValidTutorialStep(step)) {
+      return
+    }
+
+    const stepInd = orderedTutorialSteps.indexOf(step)
+    this.statsStore.recordHighestTutorialStepCompleted(stepInd)
+
+    if (stepInd > orderedTutorialSteps.indexOf(TutorialStep.PickEditor)) {
```
kuychaco: Non-blocking suggestion:

We could store off all of these indices as variables to reduce having to compute them every time we record steps as completed. For example:

```suggestion
    if (stepInd > PickEditorStepIndex) {
```

app/src/lib/stores/app-store.ts
```diff
@@ -444,34 +448,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      this.statsStore.recordTutorialEditorInstalled()
```
kuychaco: Non-blocking question: So it looks like we'll record this multiple times, right? For example, when the user progresses to step 2 we'll record it, as well as when they progress to steps 3, 4, 5, and 6. Which should be fine, just double-checking that it's intentional/desired. My sense is that we're doing it because it's the only way to account for multiple steps between runs of updateCurrentTutorialStep.

kuychaco:
Spoke with @outofambit and decided to keep `highestTutorialStepCompleted` to make it easier to analyze our metrics. So I'm merging this into the other metrics branch

PR #8371, opened 2019-09-30
Fixed strong contrast border colors.

MairwunNx:
This pull request corrects the **strong contrast** between the borders of some elements. The change is especially noticeable on the title-bar border color.

Box border color changed `#141414` -> `#191919`.
Title bar border color changed `#000000` -> `#181818`.

### Before

image

### After

image

### DIFF
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -6,7 +6,7 @@
 @import '~primer-support/lib/variables/color-system.scss';
 
 #desktop-app-title-bar {
-  border-bottom: 1px solid #181818!important;
+  border-bottom: 1px solid #181818 !important;
 }
 
 body.theme-dark {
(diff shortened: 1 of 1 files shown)


MairwunNx:
Why review failed, lol?

app/styles/themes/_dark.scss
```diff
@@ -5,6 +5,10 @@
 // Primer colors, see https://github.com/primer/primer-css/blob/master/modules/primer-support/lib/variables/color-system.scss
 @import '~primer-support/lib/variables/color-system.scss';
```
Lelum1986: [routeprint.txt](https://github.com/desktop/desktop/files/3673849/routeprint.txt)

mislav:
Thank you for your suggestions! I've reviewed the ideas with some of my teammates, but we've come to a conclusion that the reduction in contrast goes against some guidelines for minimum contrast in order to provide accessibility for visually impaired users.

I understand that these contrasting UI elements might be jarring for you, especially because this is dark mode which is supposed to be soothing on the eyes in low light environment. But we also have to balance being pleasing to the eyes with keeping sufficient contrast to ensure that Desktop remains usable for a wide variety of people.

PR #8372, opened 2019-09-30
Fixed strong contrast titlebar buttons.

MairwunNx:
I improved the color of the buttons in state `hover`, now they are **less contrasting** and **pleasing to the eyes**, see for yourself.

### Before

image

image

### After

image

image


mislav:
Thank you! Similar reasoning for closing as in https://github.com/desktop/desktop/pull/8371#issuecomment-537408845

Furthermore, this change only affect state of some buttons on hover, and we don't mind if these elements have hover styles with high contrast.

PR #8374, opened 2019-09-30
Onboarding copy updates

ampinsk:
This addresses some copy changes from [this column on the 2.2 project board](https://github.com/desktop/desktop/projects/27#column-6624222)

- Updates copy on the Commit step
- Updates copy on the Publish step
- Updates copy in the README contents

Open to feedback and suggestions on these!


app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
```diff
@@ -88,7 +88,7 @@ const InititalReadmeContents =
   `# Welcome to GitHub Desktop!${nl}${nl}` +
   `This is your README. READMEs are where you can communicate ` +
   `what your project is and how to use it.${nl}${nl}` +
-  `Make any change to this file, save it, and then head ` +
+  `Write your name on line 6, save it, and then head ` +
```
billygriffin: I love the specificity here! Will the line numbers be consistent across editors? In other words, can we rely on this "line 6" to make sense, or should we use something like, "Write your name on a new line after this one, save it..."?

app/src/ui/tutorial/tutorial-panel.tsx
```diff
@@ -221,9 +221,10 @@ export class TutorialPanel extends React.Component<
             onSummaryClick={this.onStepSummaryClick}
           >
             <p className="description">
-              A commit allows you to save sets of changes. In the bottom left
-              area, write a message that describes the changes you made. When
-              you’re done, click the blue Commit button to finish.
+              A commit allows you to save sets of changes. In the “summary“
+              field in the bottom left, write a message that describes the
```
billygriffin:
```suggestion
              field in the bottom left, write a short message that describes the
```

app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
ampinsk: I _think_ we can rely on "line 6" but maybe @outofambit can confirm?

tierninho:
For Step 6: I suggest we change `A pull request allows...` to `A pull request or "PR" allows..` to include common lingo. No strong feelings on this.

ampinsk:
Thanks for the feedback @tierninho! Since we have limited space, and PR is more casual lingo, I might not include that here, at least for this first version. Also surprisingly, I don't believe we use "PR" in the actual product! If we hear a lot of confusion I'm happy to revisit 😄

app/src/ui/blank-slate/create-tutorial-repository-dialog.tsx
outofambit: "line 6" should be consistent across text editors!

app/src/ui/tutorial/tutorial-panel.tsx
```diff
@@ -236,7 +237,8 @@ export class TutorialPanel extends React.Component<
           >
             <p className="description">
               Publishing will “push”, or upload, your commits to this branch of
-              your repository on GitHub.com.
+              your repository on GitHub.com. Publish using the third button in
```
niik: While I really like that we're using `GitHub.com` here it's possible that the repository will be published to a GitHub Enterprise Server and not to GitHub.com. I propose we make the simplest change possible here since we're coming up on a release and then we can open another low-risk PR that we can sneak into the production release which switches the language based on which endpoint we're publishing to.

```suggestion
              your repository on GitHub. Publish using the third button in
```

cc @billygriffin cause you'll have to approve this before the beta release.

PR #8389, opened 2019-10-02
Resolve harder

### LINKED ISSUE #8386: Error for failure to resolve editor after it has been uninstalled
Fullscreen_10_1_19__6_52_PM

**Repro steps**:
1. Install Atom and another valid editor (like VS Code)
1. Set Atom as your editor in Preferences
1. Uninstall Atom (on Mac, you can simply move to trash)
1. Reload GitHub Desktop

**Expected**: New editor (like VS Code) is set as editor. If in tutorial, the first step should be marked complete

**Actual**: Error and Desktop has not detected new editor. If in tutorial, the first step is not marked complete despite there being an installed editor

_[Edit] Also note that this bug prevents people from opening their editor to resolve merge conflicts if they've uninstalled their selected editor:_

GitHub_Desktop-dev

niik:
## Overview

**Closes #8386**

**Note that this targets the release branch**

## Description

In #8386 @kuychaco found that `resolveCurrentEditor` could throw an error if the preferred editor wasn't found. The error itself wouldn't cause any issues but the behavior could cause us to incorrectly state that the user had no useable editor installed. @kuychaco then opened   https://github.com/desktop/desktop/pull/8388 with the intention to refactor and solve this error in a proper way. This PR is not that and does not supersede or replace #8388.

Instead this PR is a quick and dirty (yet safe) workaround which ensures that the `resolveCurrentEditor` doesn't throw when the preferred editor isn't installed.

I think this might be worthwhile to include in the release for two reasons

1. It fixes the edge-case reported in #8386 
2. Having `_resolveCurrentEditor` throw when it wasn't intended to feels dangerous to me. Right now the only unintended side-effect of this that I've been able to prove is that background refresh might stop working while in the tutorial which doesn't matter but there could be other unforeseen consequences.

cc @kuychaco @outofambit @billygriffin.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5268,7 +5268,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   public async _resolveCurrentEditor() {
-    const match = await findEditorOrDefault(this.selectedExternalEditor)
+    const match = await findEditorOrDefault(this.selectedExternalEditor).catch(
+      () => findEditorOrDefault()
+    )
+
     const resolvedExternalEditor = match != null ? match.editor : null
     if (this.resolvedExternalEditor !== resolvedExternalEditor) {
       this.resolvedExternalEditor = resolvedExternalEditor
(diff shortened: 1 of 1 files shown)


outofambit:
unfortunately, this change introduces more inconsistencies into the UI. see the gif below for an example.

editor sadness

 @desktop/engineering has discussed this offline and we're going to close this pr and continue working on bigger solutions for the next release.

PR #8395, opened 2019-10-03
Fix [object Object] pre-filled url text in the clone dialogue when opened from BlankSlateView

### LINKED ISSUE #8394: Clone repo dialog shows `[object Object]` as repo url
Screen Shot 2019-10-02 at 2 18 23 PM

Repro steps:
1. be on "let's get started page"
2. click "clone a repo from the internet"
3. be on the URL tab and see the weird [Object object] prefill in the repo URL input field

_Originally reported by @nerdneha, repro steps provided by @billygriffin_

kuychaco:
## Overview

**Closes #8394**

## Description

A regression was introduced when we introduced the onboarding tutorial and made changes in the `BlankSlateView` component. We got rid of a wrapper method that prevented a click event object from being passed along to `showCloneRepo` where it is eventually treated as a string...

https://github.com/desktop/desktop/blob/ac48f98a14f9670f6689b6f900685322e9cb45b9/app/src/ui/app.tsx#L670

This PR restores the wrapper method, ensuring that when the "Clone" button is clicked, no arguments are passed to `showCloneRepo` and `cloneUrl` is `undefined`. I've added a comment in the code to ensure that this wrapper remains in place, as otherwise it is very easy to assume it's unnecessary because all it does is call a prop method. 

### Alternative approach
Before implementing ☝️I played around with other ways to guard against this bug from a code robustness standpoint (as opposed to relying on a comment within the code). We could verify that the `cloneUrl` is valid in `showCloneRepo` (something like 2f16869). The benefit would be that this makes the `showCloneRepo` method more robust (only valid urls get through). The drawback is that it adds a bit of extra logic and computation. If we like 2f16869 we can revert ac48f98 to get back to it. 

## Release notes

Notes: Fixes weird pre-filled url text in the clone dialogue when opened from BlankSlateView

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react'
 import { ipcRenderer, remote } from 'electron'
 import { CSSTransitionGroup } from 'react-transition-group'
-import * as URL from 'url'
 
 import {
   IAppState,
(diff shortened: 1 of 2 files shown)


tierninho:
✅ Verifying the bug is not reproducible with the fix. Nice job @kuychaco !

koobecafdiordna:
may I pull a request

PR #8405, opened 2019-10-04
Don't terminate the app immediately when the main window is closed

niik:
## Description

Some time recently (likely related to one of our Electron upgrades) the logic for when to terminate the application subtly changed to immediate termination as soon as no `BrowserWindow` instances were active.

Unfortunately this meant that our crash window broke and with it likely some percentage of all crashes ended up never getting shipped to central.

The crash handling process in Desktop (for a renderer crash) looks something like this (simplified)

1. Renderer catches an uncaughtException event
2. Renderer sticks some additional data onto the event and passes it to the main process
3. The main process closes the main window
4. The main process initiates an asynchronous network request to post the error to central
5. The main process launches the crash process

In between steps three and five we'd get interrupted by the application shutting down and therefore we're likely loosing quite a bit of needles from clients.

You can verify that the crash process doesn't work by clicking on `Help -> Crash renderer process` while in development mode or by running the following line of code in the DevTools console on the production process.

```
require('electron').remote.getCurrentWebContents().send('menu-event', { name: 'boomtown' })
```

## Release notes

Notes: TBD

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -89,6 +89,16 @@ if (__DARWIN__) {
   possibleProtocols.add('github-windows')
 }
 
+app.on('window-all-closed', () => {
+  // If we don't subscribe to this event and all windows are closed, the default
+  // behavior is to quit the app. We don't want that though, we control that
+  // behavior through the mainWindow onClose event such that on macOS we only
+  // hide the main window when a user attempts to close it.
+  //
+  // If we don't subscribe to this and change the default behavior we break
+  // the crash process window which is shown after the main window is closed.
+})
+
 process.on('uncaughtException', (error: Error) => {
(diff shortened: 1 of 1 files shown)


tierninho:
I am confirming that selecting the option `Help -> Crash renderer process` resulting in the app crashing, with the following output:

```
Error: Boomtown!
    at Immediate.<anonymous> (app.tsx:489:13)
    at processImmediate (internal/timers.js:443:21)

Version: 2.2.1-beta0
OS: Mac OS 10.14.6
```

and logs for the menu option and the line of code were:

```
debug: [main] Showing crash process window
Attempting to call a function in a renderer window that has been closed or released.
Function provided here: renderer.js:114163:23
Remote event names: will-quit
debug: [main] Got quit signal from crash process
```

6f3fe2ade7, mac

PR #8413, opened 2019-10-05
registry stuff

### LINKED ISSUE #8258: TortoiseGit steals GitHub Desktop links (closed error #4754)
The problem described in #4754 still exists.

and here's a possible solution:

I the problem is the missing registration of GitHubDesktop at HKEY_LOCAL_MACHINE\SOFTWARE\RegisteredApplications and HKEY_LOCAL_MACHINE\SOFTWARE\GitHubDesktop.
I described the solution here: https://stackoverflow.com/a/57905263/1997370
Please add the missing registration to the GitHubDesktop-installer

Rexogamer:
Closes #8258

## Description

- Sets some registry variables in a way that can’t be overridden (ie when opening certain links Desktop is always an option). Please see #8258.

## Release notes

Notes:

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -3,6 +3,7 @@ import '../lib/logging/main/install'
 import { app, Menu, ipcMain, BrowserWindow, shell } from 'electron'
 import * as Fs from 'fs'
 import * as URL from 'url'
+import { Registry } from 'rage-edit'
 
 import { MenuLabelsEvent } from '../models/menu-labels'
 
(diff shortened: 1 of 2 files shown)


Rexogamer:
Still need to add it to yarn.lock (which I cannot do) and there might be some other stuff to clean up

niik:
Hi @Rexogamer. Thank you for your efforts in trying to address this issue. As you know GitHub Desktop relies on the implementation of Electron's [setAsDefaultProtocolClient](https://electronjs.org/docs/api/app#appsetasdefaultprotocolclientprotocol-path-args) method to register the various protocols that we support. If that implementation isn't sufficient (as was suggested in the TortoiseGit issue) we'd like to see this resolved on the Electron side where it could benefit all consumers of Electron and where the change itself would get the amount of review and testing it requires.

I'd recommend opening an issue with Electron describing the problem you're facing and the solution you're proposing. For now I'm not comfortable making the suggested changes, in part because I'm not confident that we're able to foresee the effects across the versions of Windows we support, but I'd be delighted to see a discussion upstream about the pros and cons of the current implementation and possible solutions.

PR #8424, opened 2019-10-07
Fix exectuable path for launching vscodium on macos

akim8:
#8000 

## Description

Fix support for launching VSCodium. It was just a typo regarding the executable.

## Release notes

Notes: no-notes


Rexogamer:
I originally presumed they would’ve changed that. Oh well; LGTM.

PR #8438, opened 2019-10-09
Add important notes to warn against 🙀

kuychaco:
## Description

Just did a production release and learnt a couple of important (and scary) gotchas from @outofambit that would result inadvertently releasing things to production 🙀. So I'm documenting them in order to help prevent us from accidentally hitting them in the future. 

## Release notes

Notes: no-notes


Rexogamer:
🙀

PR #8463, opened 2019-10-14
update command line conflict error message 

### LINKED ISSUE #8061: Binary file conflict resolution error message
When resolving a conflict for a binary file, I noticed two things not consistent with the normal flow of a conflict when dealing with non-binary files.

We surface a message stating the conflict must be resolved via the command line, yet when you go to the conflict modal, you can select the file from source A or B and avoid the command line. 

Recommended text: "This file is in conflict and can be resolved via the command line or by returning to the conflict modal". Are there other binary file conflicts that cannot be resolved in the modal and that this message applies too?

bug

2.1.1-beta3, mac

Rexogamer:
Closes #8061 

## Description

- Changes the error message for when a conflict must be resolved via the command line or conflict modal to `This file is in conflict; the conflict can be resolved via the command line or by returning to the conflict modal.`

## Release notes

Notes: no-notes?

### DIFF
--- a/app/src/ui/diff/index.tsx
+++ b/app/src/ui/diff/index.tsx
@@ -197,7 +197,7 @@ export class Diff extends React.Component<IDiffProps, IDiffState> {
       ) {
         return (
           <div className="panel empty">
-            The file is in conflict and must be resolved via the command line.
+            This file is in conflict; the conflict can be resolved via the command line or by returning to the conflict modal.
           </div>
         )
       }


billygriffin:
Thanks @Rexogamer. Is it accurate that when this verbiage is shown, the user will *always* have the conflict modal available from the top notification area? I just want to be sure we wouldn't sometimes provide them something misleading or inconsistent here.

Rexogamer:
Not sure (sorry, but I can’t test this :/)

outofambit:
> Is it accurate that when this verbiage is shown, the user will always have the conflict modal available from the top notification area? 

@billygriffin @Rexogamer It's possible (but unlikely) to have a conflicted file without the banner present. I think the easiest way to trigger this is to start a conflicted merge, exit the modal, and commit some but not all of the conflicted (or subsequently resolved) files

billygriffin:
Thanks @tierninho and @Rexogamer. Given that the conflicts modal isn't always going to be present when this is shown, I think we ought to just leave it as is. This is likely not a frequent occurrence but I'd prefer not to add confusion for those folks by talking about a thing that doesn't exist. I'm also going to close the issue as well. Thanks for the PR.

PR #8470, opened 2019-10-15
Update SlickEdit support for 2019

texasaggie97:
## Description

- Linux/Mac: Update expected paths for SlickEdit Pro 2019
- Windows: Use App Path registry location to get path
    - This location is used for the Run dialog so application paths do not need to be added to the environment
    - This location is not per version or per edition
    - Hopefully will prevent needing to be updated every year

## Release notes

Notes: Add support for SlickEdit 2019


texasaggie97:
How do I find out what prettier has an issue with? All the output says is that `Code style issues found in the above file(s). Forgot to run Prettier?`

outofambit:
@texasaggie97 you can run `yarn lint:fix` to auto-fix prettier formatting issues!

texasaggie97:
Thank you for that command. I missed it when I looked and that took care of the issue in a specific file.

However, `yarn lint` passes for me now locally, yet the build is still failing. What else is wrong?

Rexogamer:
One of the CI providers had a Windows-specific issue, but they don’t give much other than “cmd.exe exited with code 1”.

app/src/lib/editors/win32.ts
```diff
@@ -218,6 +218,18 @@ function getRegistryKeys(
       ]
     case ExternalEditor.SlickEdit:
       return [
+        // 64-bit version of SlickEdit
+        {
+          key: HKEY.HKEY_LOCAL_MACHINE,
+          subKey:
+            'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\vs.exe',
```
niik: Hey @texasaggie97, would you mind elaborating on why the `Uninstall` path approach doesn't work for SlickEdit 2019? I don't know anything about the editor I'm afraid and it makes me a little bit nervous when we're diverging from an established pattern.

app/src/lib/editors/win32.ts
```diff
@@ -672,6 +721,7 @@ export async function getAvailableEditors(): Promise<
     results.push({
       editor: ExternalEditor.SlickEdit,
       path: slickeditPath,
+      usesShell: false,
```
niik: Can you provide some more context for this change? Has something changed in SlickEdit 2019 that necessitates the change and if so have we tested it for the previous versions of SlickEdit?

texasaggie97: This line was added for all the other editors but not SlickEdit, so I added it to be consistent with the others.

app/src/lib/editors/win32.ts
texasaggie97: The uninstall path works fine. I just don't want to have to go through this every time there is a new release and add the new GUIDs. This is my attempt at making this work for future releases as well.

app/src/lib/editors/win32.ts
niik: Looking into this it seems the default for this option is false so this change is great for consistency and doesn't change the logic.

app/src/lib/editors/win32.ts
niik: I definitely understand the frustration of having to update for each release but I'm not comfortable with diverging from the convention in favor of an app path entry. Had the executable's name been something more distinctive like `VisualSlickEdit.exe` or similar I think I may have felt more comfortable.

The whole editor resolution part of Github Desktop is due for an overhaul and I'd love it if we could find a more flexible approach than looking up these uninstall keys but we're a small team of engineers working on Github Desktop and consistency helps us have confidence in these changes where we're not able to install and test every editor out ourselves.

Would you mind changing this back to the uninstall key?

niik:
Hey @texasaggie97, since I haven't gotten a response to my questions above I'm assuming you're not interested or don't have the time to contribute to this PR at this time so I'm going to close it to keep our PR list under control.

Should you choose to pick this up again just let us know and we'll reopen this PR (or if you'd rather create a new one that's fine as well).

Thanks again for you work here and apologies for not getting around to it sooner.

PR #8477, opened 2019-10-17
Animation Tuning for Nudge Arrow

outofambit:
This PR is mostly a place to share this branch between myself and @ampinsk. Everyone else move along! This work will be folded into a separate PR with the appropriate application logic.

PR #8485, opened 2019-10-17
Onboarding Tutorial Nudge Arrows

### LINKED ISSUE #8487: Add Nudge Arrows to Some Onboarding Tutorial Steps
It can be difficult to find where in the interface to go to complete some steps of the onboarding tutorial. We'd like to add some kind of highlighting animation to the following steps:

* Create a new branch
* Publish new branch
* Make a commit

I will be (have been) working closely with @desktop/design to hone the animation for these.

outofambit:
supersedes #8477 (and includes its work)
closes #8487

## Description

- adds css for a hovering arrow (pointing left or pointing up)
- pipes existing state into components to add that css when necessary

### Screenshots

arrow 1

arrow 2

arrow 3

## Release notes

Notes: [Added] Helpful nudge animations in onboarding tutorial


tierninho:
Tested this an it works well functionally, however some matters which require further thought are:

- For the branch nudge, the text says "Go to the branch menu..." but the arrow directs you to another route -- the Branch dropdown. Perhaps they should match?
- Depending on your zoom and window sizing, the arrow disappears behind the tutorial
Screen Shot 2019-10-17 at 3 17 32 PM
- Publishing and push should not be coupled, as he help nudge only shows for publish and not push -- which I believe is inaccurate. Our real intention is to **push** the commits to GitHub and publishing is just a side effect. You can see we use push and publish interchangeably but Step 5 is not marked as complete unless you **push** the commit. Publishing the branch can technically be done at any time.  

I published the branch earlier and the push was not made:
Screen Shot 2019-10-17 at 3 23 33 PM


Build ec8381b4a9, mac

billygriffin:
Thanks @tierninho and @niik for the excellent feedback. ✨ 

> For the branch nudge, the text says "Go to the branch menu..." but the arrow directs you to another route -- the Branch dropdown. Perhaps they should match?

This is a known language problem that we're ok with for now. I also have a slight preference for "branch dropdown," but people are semi-confused about what menu and dropdown mean regardless, which is partially why the arrow is so useful.

> Depending on your zoom and window sizing, the arrow disappears behind the tutorial

Great spot! Thank you for testing on the smallest window size!

> Publishing and push should not be coupled, as he help nudge only shows for publish and not push -- which I believe is inaccurate. Our real intention is to push the commits to GitHub and publishing is just a side effect. You can see we use push and publish interchangeably but Step 5 is not marked as complete unless you push the commit. Publishing the branch can technically be done at any time.

This is also a known language limitation. Because we didn't want a separate step for publish and then another one for push, we're trying to combine them here. In usability testing, that seemed fine, and I don't think it'll cause much confusion. This step has also not seen a high dropoff rate, so I'm not concerned about it.

> the arrow is also the same color as the added+included diff gutter. Perhaps we could add a 2px white border around the arrow?

Fantastic spot. I don't think this is super likely to happen because the diff will likely be small, but solving for the case where it does since it's so lightweight would be 👍 as far as I'm concerned.

outofambit:
all fixed up! thanks @niik and @tierninho!

### Arrow shadow

arrow shadow light
arrow shadow

### z indexing
arrow on top

this required a reshuffle of some of our z-index css logic, so i did this test to make sure everything still looks ok:
z-index test

PR #8512, opened 2019-10-22
Add rebase conflict detection

### LINKED ISSUE #6960: Warn if there will be conflicts prior to starting a rebase

outofambit:
## Overview
 
closes #6960
rebased version of #8096 

## Description

This PR hooks up the work done in #8006 and #8026 to the existing rebase modal flow.

I used a new (to this repo) technical pattern inspired by [redux-saga](https://redux-saga.js.org/). Long story short is that I'm using a [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*#Description) pattern so that we can cancel multi-git-operation tasks (at the `yield` moments between git operations). This is particularly useful for more expensive tasks that can be fired rapidly (in this case by keying through the branch list).

Doing this allowed me to move a lot of git logic out of the react component and into the generator. I was also able to move the re-entrant logic already in place in the component into AppStore.


tsconfig.json
```diff
@@ -10,7 +10,14 @@
     "jsx": "react",
     "strict": true,
     "noEmit": true,
-    "outDir": "./out"
+    "outDir": "./out",
+    "lib": [
```
outofambit: since this is now specified explicitly, i've included what was the previous default, plus `esnext.asynciterable`

> Note: If --lib is not specified a default list of libraries are injected. The default libraries injected are:
► For --target ES5: DOM,ES5,ScriptHost
► For --target ES6: DOM,ES6,DOM.Iterable,ScriptHost

from https://www.typescriptlang.org/docs/handbook/compiler-options.html

outofambit:
talked with @niik about this a couple weeks ago, and we've decided to sideline this for now. we have some serious performance concerns for creating and using git work trees in this. we discussed that using https://github.com/nodegit/nodegit for this might be a better approach.

(we don't use nodegit in desktop currently because it is not performant for the majority of our use cases, but this would be an exception.)

PR #8514, opened 2019-10-22
Prevent temp worktree collisions

outofambit:
supports #8512 and #6960
originally documented in #8096

## The Problem

If you have concurrent calls to a the temp git work tree API for the same repo, they can easily clash with one another, resulting in explicit failures (and hard to debug race conditions). Our original implementation only allowed for one worktree to be created at most for each desktop user's local repo, so they would quickly get reused concurrently.

## The Fix

This PR removes that constraint by using unique paths for each work tree created by desktop. This guarantees that callers to the work tree API end up working in separate work trees.

We don't need to persist a work tree (or a path to it) beyond any single run of desktop, so we don't have to do as much work as we did with stashing feature to keep track of them.

## Release Notes

Notes: no-notes


app/src/lib/git/worktree.ts
```diff
@@ -3,6 +3,7 @@ import * as Path from 'path'
 import * as FSE from 'fs-extra'
 
 import { git } from './core'
+import { v4 as uuid } from 'uuid'
```
outofambit: I didn't use our internal wrapper for uuid because electron has since fixed the performance issues that motivated its creation.

app/src/lib/git/worktree.ts
```diff
@@ -110,38 +135,13 @@ async function findTemporaryWorkTrees(repository: Repository) {
   })
 }
 
-/**
- * Finds an existing temporary work tree for use in Desktop or creates a new one.
- * Won't modify the repository's working directory.
- * _The returned worktree will be checked out to the given commit._
- */
-export async function findOrCreateTemporaryWorkTree(
```
outofambit: since we no longer want to reuse worktrees automatically, this method is no longer needed.

Issue #8033, opened 2019-07-27
Desktop crashes if emoji pasted into Git fields


tierninho:
Desktop will crash and reload if you paste an emoji into the Git username or email fields in Preferences. Perhaps we need some validation on the fields?

Logs

```
info: [ui] Executing getConfigValueInPath: git config -z --global user.name (took 0.038s)
info: [ui] Executing getConfigValueInPath: git config -z --global user.email (took 0.015s)
info: [ui] Executing getConfigValueInPath: git config -z --global merge.tool (took 0.011s)
error: [main] Error: `value` must be a single character
    at Object.fatalError (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/src/lib/fatal-error.ts:3:9)
    at disallowedCharacter (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/src/ui/preferences/identifier-rules.ts:18:12)
    at Object.disallowedCharacters (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/src/ui/preferences/identifier-rules.ts:9:9)
    at Preferences.disallowedCharacterErrorMessage (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/src/ui/preferences/preferences.tsx:172:38)
    at Object.Preferences.onCommitterNameChanged [as onValueChanged] (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/src/ui/preferences/preferences.tsx:275:46)
    at TextBox.<anonymous> (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/src/ui/lib/text-box.tsx:165:20)
    at commitCallbacks (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/node_modules/react-dom/cjs/react-dom.development.js:6556:1)
    at commitLifeCycles (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/node_modules/react-dom/cjs/react-dom.development.js:9798:1)
    at commitAllLifeCycles (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/node_modules/react-dom/cjs/react-dom.development.js:11455:1)
    at HTMLUnknownElement.callCallback (/Users/tierninho/.Trash/desktop 10.16.03 AM/webpack:/app/node_modules/react-dom/cjs/react-dom.development.js:100:1)
[...]

Issue #3204, opened 2017-10-31
Block saving (.) as your git username or email


tierninho:
You are allowed to save `.` as your git credentials, however, it causes problems further down the line. When you attempt to commit or revert, there are two similar errors that surface, each saying you cannot save your same as `.`. 

<img width="453" alt="screen shot 2017-10-31 at 11 34 15 am" src="https://user-images.githubusercontent.com/14828183/32219887-fbd210ea-be2f-11e7-99c1-4ccf8a0a78c5.png">

We should block the user from saving their git name or email as a `.` in the `Preferences > Git` modal. Not sure if there are other special characters, et al, that would cause disruption as well.

<img width="401" alt="screen shot 2017-10-31 at 11 33 10 am" src="https://user-images.githubusercontent.com/14828183/32219886-fbafb9b4-be2f-11e7-8f17-bd2af3dbaddd.png">
<img width="603" alt="screen shot 2017-10-31 at 11 33 27 am" src="https://user-images.githubusercontent.com/14828183/32219888-fbfbdd1c-be2f-11e7-8426-0e898655facc.png">

1.0.7-beta0, mac

niik:
For the future implementor of this. It would be spectacular if we could find a way to leverage Git's own validation logic here rather than try to replicate it ourselves.

shiftkey:
@tierninho can you recall the flow to trigger that first error? The second one looks like committing, but I'm curious if we stumble upon it earlier...

tierninho:
@shiftkey took a sec to figure it out, but that error can be surfaced when you revert a commit... in this case the initial commit on an already published repo.

<img width="850" alt="screen shot 2017-11-02 at 8 34 01 am" src="https://user-images.githubusercontent.com/14828183/32314569-d485d0ec-bfa8-11e7-8d01-9a0960727db3.png">

nerdneha:
Looks like they use a function called `crud` to throw this error, see:
https://github.com/git/git/blob/e629a7d28a405e48fae6b064a781a10e885159fc/ident.c#L191-L203

shiftkey:
Nice find @nerdneha. While @niik above suggested that it'd be great if we could get this by invoking Git in some way, the rules themselves haven't changed in at least 9 years:

<img width="783" src="https://user-images.githubusercontent.com/359239/33863940-5852e928-df3e-11e7-87b9-18eb7baabff2.png">

I'd suggest extracting those rules into our app is a fair  compromise.

PR #8541, opened 2019-10-28
Don't crash when attempting to use multibyte characters as name

### LINKED ISSUE #8534: Setting an Emoji as Name triggers error on blur
### Describe the bug
I set my Commit name to an Emoji
and then when i go to edit my email it crashes

### Version & OS
Tested on OS X and Windows

1. Go to Github commit settings
2. Click on name and put in an emoji 🧙 
3.  down to click on email 

### Expected behavior
edit the email 

### Actual behavior

Error: `value` must be a single character at Object.n [as fatalError] (file:///app/renderer.js:1:1753) at o (file:///app/renderer.js:16:594244) at Object.t.disallowed Characters (file:///app/renderer.js:16:594543) at C.disallowed Character ErrorMessage (file:///app/renderer.js:16:605654) at Object.C.onCommitterNameChanged [as onValueChanged] (file:///app/renderer.js:16:602110) at d.<anonymous> (file:///app/renderer.js:11:26012) at Re (file:///app/renderer.js:27:20247) at commitLifeCycles (file:///app/renderer.js:27:44342) at L (file:///app/renderer.js:27:55807) at A (file:///app/renderer.js:27:54616)
A clear and concise description of what actually happened.

### Screenshots

### Logs

niik:
Closes #8534 
X-ref: https://github.com/desktop/desktop/issues/8033

## Description

The method we used to determine whether the `Name` text box in the preferences dialog consisted only of disallowed characters would throw an exception when encountering a multi-byte character. See https://github.com/desktop/desktop/issues/3204 for the backstory of why we do this at all.

The simplest fix here would have been to simply remove these lines...

https://github.com/desktop/desktop/blob/2ea8c4f7dd4989c20d8cd0edf8edfaed55ec79c3/app/src/ui/preferences/identifier-rules.ts#L17-L19

... as it shouldn't be considered a fatal error to have multibyte characters in the name. That said I took the liberty to do some cleanup while I was in there as the method was lacking in documentation and didn't follow the prevailing naming conventions of the app today.

## Release notes

Notes: [Fixed] Prevent crash when setting a name consisting of multibyte characters (such as emojis) in the preferences dialog.


tierninho:
I couldn't not trigger the error message with `😀`, but I was able to see the "Name is invalid..."  error message when pasting in a `.`. Is this the intended behavior?

niik:
> but I was able to see the "Name is invalid..." error message when pasting in a `.`. Is this the intended behavior?

It is, and you should see the same behavior (albeit with a different error message) in the latest production release as well. This PR does not (at least not intentionally) change the definition of what's considered valid. One way to test what Git considers valid is to run the following command

```
$ GIT_COMMITTER_NAME="." GIT_COMMITTER_EMAIL="foo@foo.com" git var GIT_COMMITTER_IDENT
fatal: name consists only of disallowed characters: .
GIT_COMMITTER_NAME="M.O" GIT_COMMITTER_EMAIL="foo@foo.com" git var GIT_COMMITTER_IDENT
M.O <foo@foo.com> 1572343411 +0100
```

PR #8555, opened 2019-10-30
add notarization for publishable macOS builds

### LINKED ISSUE #8506: Unable to use GitHub Desktop on MacOS Catalina 10.15
### I'm unable to install. 

I get this error message here.

Screen Shot 2019-10-21 at 5 50 59 PM

I've tried modifying for privacy, but I'm thinking it's versioning support from Mac?

outofambit:
closes #8506 

## Description

- look for (and use) credentials to notarize mac builds
- use a newer version of xcode (that supports notarization) on circleci
- due to electron/electron-notarize#12, i added a postinstall patch for our `electron-notarize` dependency using [`patch-package`](https://github.com/ds300/patch-package). i tried a lot of other solutions, but yarn resolutions failed and other options required us to jump far ahead in our electron packaging dependencies.
- this does slow down our mac build process unfortunately. to mitigate this, we'll need to revisit our build scripts and remove signing/notarizing/etc for CI builds that aren't going to be published
- enable the `hardenedRuntime` code-signing option (required for notarizing)
- add necessary code-signing entitlements for desktop run in the `hardenedRuntime` environment (https://github.com/electron/electron-osx-sign/issues/188#issuecomment-483778579)

## Release notes

Notes: no-notes


Rexogamer:
failed on CiricleCI

also is it just me or are a lot of CircleCI builds 404s for some reason

outofambit:
fyi: this build won't pass until CircleCI is configured correctly

Rexogamer:
question: what is notarisation?

tierninho:
Notarization: https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution

outofambit:
this build is currently failing due to https://github.com/electron/electron-notarize/issues/12

outofambit:
sadly, this is not working in my local testing. will update with more information next week (and hopefully a solution.)

outofambit:
tried the latest test build locally and it works great! @niik could you review and see if you have any concerns? thank you!

script/build.ts
```diff
@@ -171,7 +188,15 @@ function packageApp() {
     appBundleId: getBundleID(),
     appCategoryType: 'public.app-category.developer-tools',
     darwinDarkModeSupport: true,
-    osxSign: true,
+    osxSign: {
+      // @ts-ignore (the library's type definition is wrong)
+      hardenedRuntime: true,
+      identity: true,
```
niik: If I'm reading things correctly I _think_ we can drop this and rely on the default behavior rather than fight the type system. It seems that by setting this to true we're causing `electron-packager` to reset it to `null`:

https://github.com/electron/electron-packager/blob/fd237af51aefe3dbbdc3a84ed380eb468317db3b/mac.js#L367-L369

This gets sent to `electron-osx-sign` which does a truthy check for `.identity` and treats anything falsy (such as `null`) as asking for an automatic detection of identity.

https://github.com/electron/electron-osx-sign/blob/0aa699d20754c9b9fe076723871ca19aa6a255e4/sign.js#L233

script/build.ts
```diff
@@ -171,7 +188,15 @@ function packageApp() {
     appBundleId: getBundleID(),
     appCategoryType: 'public.app-category.developer-tools',
     darwinDarkModeSupport: true,
-    osxSign: true,
+    osxSign: {
+      // @ts-ignore (the library's type definition is wrong)
```
niik: This is a heavy hammer we're using here, if I'm not mistaken it's ignoring this entire subtype (as opposed to just the subsequent line).

This makes me a little nervous. Is there any chance we could make what we're overriding be explicit? Looking at the type it seems what's missing is the `hardenedRuntime` and the `type` properties. Could we extend the custom type we already have to include these? If we add the following to the `IPackageAdditionalOptions` type...

```ts
readonly osxSign: packager.ElectronOsXSignOptions & {
  readonly hardenedRuntime?:boolean
  readonly type?: string
}
```

... combined with my comment below about removing the `identity` property we should be okay?

outofambit:
>>    this does slow down our mac build process unfortunately. to mitigate this, we'll need to revisit our build scripts and remove signing/notarizing/etc for CI builds that aren't going to be published
>
> If I'm reading the code correctly you've already done this, right?

@niik not quite, sorry for not making this clearer. i've addressed removing notarizing from dev builds, so that shouldn't be impacted. But, since we use the prod build for our mac CI process, we are notarizing builds that are only for CI testing.

https://github.com/desktop/desktop/blob/development/.circleci/config.yml#L40

To fix that, we'll need to spend a bit more time re-configuring our test vs build vs deploy process on circleci.

script/build.ts
outofambit: i kept `identity: true` in hopes that it would signal similarly to `osxSign: true` from before. but i'm happy to remove it! (eb00ee4)

script/build.ts
outofambit: ah thanks, i missed our existing type override. the `type` field wasn't a problem, so i just added `hardenedRuntime`. (yesterday i also opened https://github.com/electron/electron-osx-sign/pull/216 for fix things upstream.) 

(also from what i've read, [`@ts-ignore` only ever ignores errors for the one line after it](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#suppress-errors-in-ts-files-using--ts-ignore-comments), but i agree its a heavy hammer when we can easily override the type.)

https://github.com/desktop/desktop/pull/8555/commits/3b8675f5cbde5c0375e0c4e5d8971713eab8f421

patches/electron-notarize+0.0.5.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/node_modules/electron-notarize/lib/index.js b/node_modules/electron-notarize/lib/index.js
```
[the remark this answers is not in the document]
outofambit: that's super fair!

in https://github.com/desktop/desktop/pull/8555/commits/51f0fb1399c269670b5e9a6da6d33258cb4b0ee6 i've added the patch for `index.js.map`.
in https://github.com/desktop/desktop/pull/8555/commits/306397a41f543e6f8a2c90a20d946502659ed054 i've added a comment to the patch file itself that points to [the commit i used](https://github.com/desktop/electron-notarize/commit/2795347201ab8d32d1240147cebdd68c4a94d8fc) to generate it.

niik:
> @niik not quite, sorry for not making this clearer. i've addressed removing notarizing from dev builds, so that shouldn't be impacted. But, since we use the prod build for our mac CI process, we are notarizing builds that are only for CI testing.

Hmm, I thought that the `isPublishableBuild` condition for attempting to pull the `notarizationCredentials` would take care of this. I was under the impression that `isPublishableBuild` will only return true if we're building on a release branch which should only happen as a result of a deployment.

script/build.ts
niik: > also from what i've read, @ts-ignore only ever ignores errors for the one line after it, but i agree its a heavy hammer when we can easily override the type

This is what I thought as well but when I tried removing the comment as well as the `hardenedRuntime` property I still got an error for the `identity` type mismatch. 🤷‍♂

outofambit:
> Hmm, I thought that the `isPublishableBuild` condition for attempting to pull the `notarizationCredentials` would take care of this. I was under the impression that `isPublishableBuild` will only return true if we're building on a release branch which should only happen as a result of a deployment.

@niik i thought something similar at first, but `isPublishableBuild` is not that sophisticated right now.

https://github.com/desktop/desktop/blob/f0f1404f9d7cbbf9ced04c9bf18f4cb9ee7340b5/script/build.ts#L49

`getReleaseChannel()` checks the branch name for the release prefix, but if it doesn't find it, it just falls back to `NODE_ENV`.

https://github.com/desktop/desktop/blob/f0f1404f9d7cbbf9ced04c9bf18f4cb9ee7340b5/script/dist-info.ts#L101-L109

since we use `yarn build:prod`, we set `NODE_ENV` to `production`

https://github.com/desktop/desktop/blob/f0f1404f9d7cbbf9ced04c9bf18f4cb9ee7340b5/package.json#L24

___

> we need to figure that out before landing this so that forked PRs aren't affected (since forked PRs don't have access to the environment variables).

@niik gotcha, good catch. i think i have a minimal fix we can add here.

outofambit:
@niik i've done some (light) refactoring to preserve our current logic but make our "isPublishable" logic more accurate. tried the test build locally on windows and mac and everything seems to be good!

PR #8556, opened 2019-10-30
remove travis config

outofambit:
We don't use the Travis build for anything and it creates noise in our CI statuses.

## Description

- removes travis config

## Release notes

Notes: no-notes


Rexogamer:
there are various references to travis that need to be removed with this pr

outofambit:
@Rexogamer i'm not really concerned about travis build tooling in our scripts. the focus here is to decrease noise in our CI statuses.

outofambit:
@niik @kuychaco i know we've discussed this previously. this is a super light weight approach i tried (while looking into notarization stuff). i'd like to merge this monday next week, unless there are objections before then. thanks!

cc @nerdneha

PR #8563, opened 2019-11-01
[Branch protection] Use new API endpoint

### LINKED ISSUE #7023: I accidentally commit locally to protected branches

kuychaco:
Closes #7023

## Description

Use new `push_control` endpoint introduced in https://github.com/github/github/issues/127973 to determine whether the user's current branch is protected.

Relevant work PRs that this builds on top of: #7826, #7595 and #7529

### Screenshots

GitHub_Desktop-dev_and_first_pass_at_UI_for_warning_about_protected_branches_by_shiftkey_·_Pull_Request__7826_·_desktop_desktop

### Testing
Manually tested the following cases on review-lab:
- owner has no warning if admin not included and review is required
- owner has warning if admin is included and review is required
- collaborator on allow list has no warning
- collaborator on team on allow list has no warning
- collaborator not on allow list has warning
- collaborator on allow list has warning if review is required

## Release notes

Notes: [Feature] Suggest moving to a branch prior to committing to a protected branch


app/src/lib/stores/app-store.ts
```diff
@@ -1474,6 +1495,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
     repository: Repository,
     previouslySelectedRepository: Repository | CloningRepository | null
   ): Promise<Repository | null> {
+    this.refreshBranchProtectionState(repository)
+
+    // This is now purely for metrics collection for `commitsToRepositoryWithBranchProtections`
+    // Understanding how many users actually contribute to repos with branch protections gives us
+    // insight into who our users are and what kinds of work they do
+    this.updateBranchProtectionsFromAPI(repository)
```
kuychaco: Now that we're only using the `repos/${owner}/${name}/branches?protected=true` endpoint to gather metrics about if a user is committing to a repository that has protected branches, we can afford to update the data less frequently (individual rule changes don't matter to us, only if the repo has a non-zero number of protected branches). So for now I'm opting to simply update it on repo change.

app/src/lib/stores/app-store.ts
```diff
@@ -825,19 +825,40 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        const {
+          allow_actor,
+          required_status_checks,
+          required_approving_review_count,
+        } = await api.fetchPushControl(owner, name, branchName)
+
+        const pushableByUser =
+          allow_actor &&
+          required_status_checks.length === 0 &&
+          required_approving_review_count === 0
```
outofambit: could this go into a helper method that lives somewhere in `lib/`? this is pretty crucial logic and i want to make sure its clearly encapsulated and documented.

i'm also a little unclear why we need to use the required status checks and review counts. are these enforced for pushing directly to a branch and not just for merges?

app/src/lib/stores/app-store.ts
outofambit: 💯!

app/src/lib/stores/repositories-store.ts
```diff
@@ -638,41 +638,6 @@ export class RepositoriesStore extends BaseStore {
 
     return branchProtectionsFound
   }
-
-  /**
-   * Check if the given branch for the repository is protected through the
-   * GitHub API.
-   */
```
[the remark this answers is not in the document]
outofambit: 🔥 😍

app/src/lib/api.ts
```diff
@@ -232,6 +232,41 @@ export interface IAPIRefStatus {
...
+  allow_actor: boolean
+
+  /**
+   * Currently unused properties
+   */
+  pattern: string | null
+  required_signatures: boolean
+  required_linear_history: boolean
+  allow_deletions: boolean
+  allow_force_pushes: boolean
```
outofambit: i like that we're including these so its easy to use them later if we need to

app/src/lib/api.ts
```diff
@@ -719,6 +754,33 @@ export class API {
     return await parsedResponse<IAPIRefStatus>(response)
   }
 
+  public async fetchPushControl(
```
outofambit: a little doc for this method please? i think its worth calling out that if it can't get the info, it defaults to assuming full access for the user (which i think is the right approach btw).

PR #8599, opened 2019-11-07
Platform specific dialog button order, take two

### LINKED ISSUE #2463: swap buttons in 'remove repository' and 'discard changes' dialog on Windows
### Description

I would like to suggest swapping the Cancel and Remove buttons in the confirmation dialog for removing a repo from Desktop. The reason is that in other dialogs, such as branch creation, cloning and creating a new repo, the button to proceed/confirm is always on the left side and colored blue.

image

For the sake of consistency, I think this should go for the repo removal confirmation as well. It's quite easy for a user who might not exactly be paying attention, to hit the wrong button, because they were expecting Cancel to be on the other side.

I guess it might be nitpicky, but this really struck me as an odd design decision. :smile: If this was done intentionally to draw the user's attention to the Cancel button (as removing something might be destructive), might I suggest instead coloring the Remove button red, like the 'danger zone' buttons on a repo's controls on the GH site?

Other dialog boxes, to illustrate my point: (Note that the "yes, continue"-like option is colored and is on the left.)

image
image

### Version

**GitHub Desktop version:** 0.7.2

**OS version:** Windows 10 Home

### LINKED ISSUE #5883: Document how we treat dialog boxes
We constantly run up against questions about how dialog boxes should be treated, and the differentiation between macOS and Windows makes it even trickier.

I'm planning to write up documentation to clarify the guidelines we use for macOS and for Windows and how they *should* operate in Desktop. Hopefully this will help identify places where that isn't happening correctly today, and for future implementations make it easier to figure out what to do.

### LINKED ISSUE #7186: Button order is flipped on Windows for Discard changes dialog
## Description

On Windows, it's "Ok Cancel", not "Cancel Ok", even if the default action here is Cancel. Other dialogs seem to be correct though, kudos!

image

## Version

* GitHub Desktop: 1.6.5

niik:
😅 **This looks like a massive PR but most of the changes are straightforward refactors** 😅 

Closes #7186
Closes #2463
Closes #5883

## Description

Ever since the initial implementation of dialogs in Desktop we've been aware of, and had support for, platform-specific button order in dialogs. A good summary of the difference can be found in https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/. Given how complicated our solution to this problem had become (see https://github.com/desktop/desktop/issues/7273) @billygriffin and I decided to take a look at solving this problem holistically.

Our solution to the problem up until now was a component called `ButtonGroup`, the usage of which looked something like this

```jsx
<ButtonGroup>
  <Button type="submit">Ok</Button>
  <Button>Cancel</Button>
</ButtonGroup>
```

The button order above matches the Windows convention so at runtime the button group would switch the order of the buttons if running on macOS. In order to ensure that we used the Windows convention at compile time we had a [static analysis rule](https://github.com/desktop/desktop/blob/f9d27338260fb0e464b4d7cc22c9d5b10cf1dba3/tslint-rules/buttonGroupOrderRule.ts) which would emit an error if the submit button was not the first child of the ButtonGroup.

This approach worked reasonably well for a while with the exception of it being hard to reason about what the order would end up being at runtime and the fact that there was nothing stopping us from using the submit button as the cancel button.

The real problem with our support had to do with the introduction of the concept of "destructive" choices in dialogs.

### Button order

We started out by defining the expected behavior and came up with the following graph.

button conventions

I.e, contrary to what we've done in the past the destructive property should not alter the order of the buttons, it should, however, affect which button is the primary (or "default").

Our proposed solution to this is a new component called `OkCancelButtonGroup` which does not let consumers alter the ordering of the buttons and thanks to it owning the rendering of the buttons it's also able to switch which of the buttons should be considered the primary based on whether the dialog is destructive or not.

### Dialog auto focus

A secondary, follow on, problem to fixing the ordering on Windows was that the left-most button could be the destructive action and since the dialog element automatically focuses the first keyboard-focusable element of the dialog we would end up with dialogs where the destructive button had the initial keyboard focus. That's obviously very dangerous as picking the destructive action is now only a space or enter key away. What we want is for the safe option to have keyboard focus.

To solve this I've introduced some new focus logic in the dialog, the intricacies of which is probably best explained by reading the comments in the `focusFirstSuitableChild` method of the dialog component.

### Review notes

A lot of the changes in this PR are boilerplate refactors. The most important areas to look at are as follows

 - the `focusFirstSuitableChild` method in `dialog.tsx`
 - the `OkCancelButtonGroup` component
 - the `DefaultDialogFooter` component

Other things to note:

 - A lot of custom focusing logic from individual dialogs are now gone thanks to the focus algorithm baked into the dialog itself.
 - The `AbortMergeWarning` and `ConfirmAbortDialog` dialogs were mimicking a warning dialog by including a warning octicon in the dialog contents rather than using the `type='warning'` dialog prop. This has been addressed.

### Screenshots

Artboard Copy 2

### TODO

 - [x] Finalize button order documentation (`button-order.md`) - @billygriffin 
 - [x] Finalize dialog documentation (`dialogs.md`) - @niik

## Release notes

Notes: [Improved] Dialog button order in destructive dialogs follows platform convention on Windows.


niik:
@billygriffin Would you say this also closes https://github.com/desktop/desktop/issues/5883?

billygriffin:
Nice spot @niik, wish I would've seen that issue when it was created so I'd have known about it! 😆 Adding to the body.

app/src/ui/dialog/dialog.tsx
```diff
@@ -206,9 +206,8 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
       return
     }
 
-    // This cast to any is necessary since React doesn't know about the
-    // dialog element yet.
-    ;(this.dialogElement as any).showModal()
+    this.dialogElement.showModal()
```
outofambit: 🥇

app/src/ui/dialog/dialog.tsx
```diff
@@ -221,9 +220,135 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
...
+    if (dialog === null) {
+      return
+    }
+
+    const selector = [
+      'input:not([type=hidden]):not(:disabled):not([tabindex="-1"])',
+      'textarea:not(:disabled):not([tabindex="-1"])',
+      'button:not(:disabled):not([tabindex="-1"])',
+      '[tabindex]:not(:disabled):not([tabindex="-1"])',
+    ].join(', ')
```
outofambit: this could be defined at the file-level, no? (this is a nit-pick and is non-blocking)

app/src/ui/dialog/dialog.tsx
```diff
@@ -221,9 +220,135 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
...
+      ':not([type=reset])',
+      ':not([type=hidden])',
+      ':not([type=checkbox])',
+      ':not([type=radio])',
+    ]
+
+    const inputSelector = `input${excludedInputTypes.join('')}, textarea`
+    const buttonSelector =
+      'input[type=button], input[type=submit] input[type=reset], button'
+    const submitSelector = 'input[type=submit], button[type=submit]'
```
outofambit: these could be defined at the file-level as well, i think? (this is a nit-pick and is non-blocking)

app/src/ui/dialog/dialog.tsx
```diff
@@ -221,9 +220,135 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
...
+   *     tab)
+   *
+   *  3. The first submit button. We use this as a proxy for what macOS HIG calls
+   *     "default button". It's not the same thing but for our purposes it's close
+   *     enough.
+   *
+   *  4. Any remaining button
+   *
+   */
+  private focusFirstSuitableChild() {
```
outofambit: i won't pretend to fully understand the internal logic of this method, but i'm a bit concerned that we're always overriding default focus behavior with it. is there a way we could only use this logic if the dialog is destructive?

app/src/ui/dialog/dialog.tsx
niik: They could but given that this method is called very rarely (once per dialog shown) performance isn't an issue and I think it's beneficial for readability if we keep it localized to the actual focus logic.

app/src/ui/dialog/dialog.tsx
niik: The intention is to always override the default focus behavior because the default logic doesn't fit our needs. The diff has a number of examples of where we've now been able to remove explicit focus calls by individual dialogs to work around shortcomings in the default focus logic (`acknowledgements.tsx` is a good example). With this we've got a consistent focus logic (albeit custom) that should hopefully suit our needs. Should we discover that a particular dialog suffers from this focus logic (which is certainly possible, I've tested a lot of them but it's possible I missed something) I think we should either see if we can fix the logic in such a way that it works for all dialogs and if that's not possible we should add an opt-out prop for the dialog component that can be explicitly set for that case.

All that said dialogs still have the ability to control focus manually after they've been mounted.

PR #8614, opened 2019-11-11
Prevent application crash when spawn fails

niik:
## Description

This PR bumps dugite to v1.87.3, a patch release containing https://github.com/desktop/dugite/pull/381. The dugite bump is necessary in order for Desktop to be able to properly handle errors related to process spawning.

If the operating system prevented us from launching a processes due to resource exhaustion (EAGAIN), permissions problems (EACCES), or other problems `GitProcess.exec` and `GitProcess.spawn` would throw an error due to the stdio streams on the launched child process not being available. This error would mask the actual, underlying error which would be [emitted asynchronously](https://github.com/nodejs/node/blob/75657ee41677d56a498e298b01d049aef5bf199f/lib/internal/child_process.js#L348) and since we hadn't subscribed to the `error` handler of the process yet (due to the masking exception) the exception would be raised as an unhandled exception and the app would crash.

In addition to fixing this "unpreventable" crash this PR audits the places in Desktop where we assume that the `process.std*` streams are always present and adds som extra debugging data for us to use instead of the generic `spawn /Users/.../git: EAGAIN` error thrown by Node.

## Release notes

Notes: [Fix] Prevent application crash when system is unable to launch Git


outofambit:
CI is failing here because we didn't update the lockfile with `package.json`. pushing that tiny change up now so build can go green and i can merge.

PR #8619, opened 2019-11-12
Audit exception messages

niik:
## Description

This PR audits our exception messages and, where possible, attempts to remove potentially sensitive information such as file contents and repository paths that we don't need for troubleshooting.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/no-repositories/create-tutorial-repository-dialog.tsx
+++ b/app/src/ui/no-repositories/create-tutorial-repository-dialog.tsx
@@ -184,7 +184,7 @@ export class CreateTutorialRepositoryDialog extends React.Component<
 
       if (await pathExists(path)) {
         throw new Error(
-          `The path ${path} already exists. Please move it ` +
+          `The path '${path}' already exists. Please move it ` +
             'out of the way, or remove it, and then try again.'
         )
       }
--- a/app/src/lib/diff-parser.ts
+++ b/app/src/lib/diff-parser.ts
@@ -230,7 +230,7 @@ export class DiffParser {
   private parseHunkHeader(line: string): DiffHunkHeader {
     const m = diffHeaderRe.exec(line)
     if (!m) {
-      throw new Error(`Invalid hunk header format: '${line}'`)
+      throw new Error(`Invalid hunk header format`)
     }
 
     // If endLines are missing default to 1, see diffHeaderRe docs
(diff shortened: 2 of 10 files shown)


app/src/ui/no-repositories/create-tutorial-repository-dialog.tsx
```diff
@@ -184,7 +184,7 @@ export class CreateTutorialRepositoryDialog extends React.Component<
 
       if (await pathExists(path)) {
         throw new Error(
-          `The path ${path} already exists. Please move it ` +
+          `The path '${path}' already exists. Please move it ` +
```
outofambit: I don't think this full path is very helpful, and will get sent to central as a non-fatal error. Does adding the `''` make it easier to scrub? i think we could also:
```suggestion
          `The path '${name}' already exists. Please move it ` +
```

niik: > I don't think this full path is very helpful

While I agree that it's not super helpful for us this exception message is presented as-is to the user and for them I'd argue it's quite necessary to see the full path. In fact, `name`, here is a constant (`'desktop-tutorial'`). Until we've built in support for overwriting, or forcing the repo creation elsewhere, we need the user to be able to go into explorer/finder and manually remove the tutorial folder.

> Does adding the '' make it easier to scrub?

I wan initially thinking we should add a generic scrubber for anything matchin `path '[^']'` but I've added a specific scrub for this pattern instead. For consistency though I think we should keep it, we usually put quotes around paths in exception messages.

outofambit: @niik Ah right. The path is useful for the user, but not the exception reporting.

> but I've added a specific scrub for this pattern instead

So that means this particular error message will be scrubbed in `central`? we could also add logic in the `catch` statement to remove the path (near line 226).

niik: > So that means this particular error message will be scrubbed in `central`?

Yep, that part is being deployed as we speak!

> we could also add logic in the `catch` statement to remove the path (near line 226).

Yeah, we could but I need to scrub it for versions existing already so the central side was needed regardless. Additionally I think we should discuss not sending these exceptions to the non-fatal bucket any more. We've captured enough data to understand what the problem is and I think we should treat the non-fatal bucket as just that a temporary resource to investigate a problem (rather than a catch-all bucket).

outofambit: sounds good @niik! thank you!

PR #8634, opened 2019-11-14
Remove Azure Pipelines CI

outofambit:
In #8614, Azure Pipelines reported a fail status due an install issue on the Linux job. I began by removing the Linux job, but then thought that we shouldn't maintain CI services we're not using (which is exactly what I was doing in that moment).

**So, this PR removes our Azure Pipelines config.** If we want to centralize our CI onto Azure Pipelines later, we can revert eff96e3 and pick up where we left off.

**I'd like to merge this on Monday, November 18th.** @desktop/engineering, please let me know if you have any concerns before then.


outofambit:
gonna yolo this. happy to revert later if others would like!

Issue #8509, opened 2019-10-22
Improve how Desktop handles forks


billygriffin:
This is a meta issue intended to clarify the landscape around problems related to forks of repositories. The primary use case for forks is to contribute to repositories where the person contributing does not have write access. These contributions are almost exclusively made via pull requests. This process today is confusing and Desktop can make it a lot simpler and help people get out of a bind more easily.

### Problem 1: Trying to push changes to a repo you don't have write access to

If you've cloned a repo and make some changes with the intent to contribute, you're basically stuck. You can't push those changes to the repo you've cloned because you don't have write access, and you can't bring them over to a fork that doesn't yet exist in a low-friction way.

Desktop could let you know you don't have write access and give you a nice pathway to fork and create a PR. This is conceptually similar to local commits to protected branches, where ideally we wouldn't allow you to do that (because you'd never be able to push those commits to that branch), and instead give you a nice path to somewhere where your work can eventually be pushed without issue.

A potential solution is outlined in https://github.com/desktop/desktop/issues/3918 by @say25.

### Problem 2: Staying up to date on your fork with changes on upstream
It's not at all straightforward how to pull in updates from the default branch of the upstream repo while you're working on a contributions on your forked repo. You don't necessarily know they exist, and even if you do it's not clear for people who don't do it frequently how to merge in those changes to get your branch up to date.

One way in which this manifests is when you're creating a new branch, and if there has been new work on the upstream repo, you're starting your branch out-of-date: https://github.com/desktop/desktop/issues/7762

We also have an existing feature (notification when diverging from default branch) that was intended to solve some of this, but it's currently not working as designed. A potential path forward on that specifically is described here: https://github.com/desktop/desktop/issues/6875#issuecomment-470302301

### Problem 3: Viewing pull requests for the upstream repo from your fork (including your own PR)
If you fork a repo and make a PR to contribute to the upstream repo, there's currently no way to view your PR in Desktop from your fork because the PR is associated with the upstream repo, not your fork of it.

Desktop could pull in PRs from both the forked repo and the upstream repo to ensure that you can easily see your changes. That's described here: https://github.com/desktop/desktop/issues/6383

### Problem 4: Cannot open a pull request from a GitHub.com PR page in your fork in Desktop

[...]

PR #8635, opened 2019-11-15
Store user's repo permissions

outofambit:
supports #8509 

## Description

- parse and store user's repo permissions (via existing API calls)
- a little cleanup in `GitHubRepository` model definition


app/src/lib/databases/repositories-database.ts
```diff
@@ -21,6 +21,8 @@ export interface IDatabaseGitHubRepository {
 
   /** The last time a prune was attempted on the repository */
   readonly lastPruneDate: number | null
+
+  readonly permissions: 'read' | 'write' | 'admin' | null
```
niik: I _think_ this needs to be `permissions?` since we're not doing any database migration to ensure that it's present on the existing database entries. If we don't I think we could end up having `GitHubRepository` instances where the `permissions` property is undefined even though the type says that it can only be one of the three strings or `null`.

```suggestion
  readonly permissions?: 'read' | 'write' | 'admin' | null
```

outofambit: you are correct! thanks for catching that. updated in https://github.com/desktop/desktop/pull/8635/commits/09e35e9ea63ef2cb62bf6a89d1c10217c5ea17ee

outofambit:
gonna merge this so we can start looking at dependent PRs more easily.

PR #8652, opened 2019-11-19
Support pull request actions from local forks

### LINKED ISSUE #8549: [Fork] "Create pull request" is still a suggestion even after I've created a PR for the current branch on my fork
_Extracted from https://github.com/desktop/desktop/issues/8509 (Problem 2) for tracking work on this particular issue_

### Problem: "Create pull request" is still a suggestion even after I've created a pull request for the current branch on my fork

After you create a PR from a branch of your fork that's targeting the upstream repo, the "Create pull request" suggestion is still displayed. This is very confusing because you already have a PR.

Example: I'm in billygriffin/desktop (a fork of desktop/desktop) on my branch new-work, and I make some commits and create a pull request targeting the development branch of desktop/desktop. In the app, I'm still shown the "Create pull request" suggested next step even though I've already created one.

outofambit:
Closes #8549

- Improves PR matching logic to match local branches with upstream PRs
- Fixes logic to navigate to correct PR webpage from local branch
- Hardens logic around opening a new PR from a fork

## Description

- store both upstream and regular PRs for cloned forks in our pr database
- add upstream remote to repository state cache
- and logic to handle multiple remotes while looking up PRs
- filter PR list to omit upstream in UI 

### Screenshots

gif

electron fork


app/src/lib/helpers/pull-request-matching.ts
```diff
@@ -0,0 +1,41 @@
+import { IRemote } from '../../models/remote'
+import { repositoryMatchesRemote } from '../repository-matching'
+import { Branch } from '../../models/branch'
+import { PullRequest } from '../../models/pull-request'
+
+/**
+ * Find the pull request for this branch.
+ *
+ * The upstream remote is only used
```
joshaber: I'm not sure I understand this comment? 🤔 Or I'm having a hard time figuring out what it means in relation to the arguments.

app/src/lib/stores/pull-request-store.ts
```diff
@@ -139,9 +142,16 @@ export class PullRequestStore {
     api: API,
     repository: GitHubRepository
   ) {
-    const { name, owner } = getNameWithOwner(repository)
-    const open = await api.fetchAllOpenPullRequests(owner, name)
-    await this.storePullRequestsAndEmitUpdate(open, repository)
+    const bothPrLists = await Promise.all([
+      fetchOpenPullRequests(api, repository),
+      repository.fork && repository.parent
+        ? fetchOpenPullRequests(api, repository)
```
joshaber: Maybe I'm missing something, but shouldn't this fetch the open PRs on `repository.parent` not `repository`?

app/src/lib/stores/app-store.ts
```diff
@@ -5105,16 +5115,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
 
-    const urlEncodedBranchName = escape(branch.nameWithoutRemote)
-    const baseURL = `${
-      gitHubRepository.htmlURL
-    }/pull/new/${urlEncodedBranchName}`
+    const { parent: parentGitHubRepository } = gitHubRepository
 
+    const urlEncodedBranchName = escape(branch.nameWithoutRemote)
+    const baseURL =
+      repository.isGitHubFork && parentGitHubRepository
```
joshaber: I think this should be:

```suggestion
      repository.isGitHubFork() && parentGitHubRepository
```

since `isGitHubFork` is a method.

app/src/lib/stores/pull-request-store.ts
```diff
@@ -139,9 +142,16 @@ export class PullRequestStore {
     api: API,
     repository: GitHubRepository
   ) {
-    const { name, owner } = getNameWithOwner(repository)
-    const open = await api.fetchAllOpenPullRequests(owner, name)
-    await this.storePullRequestsAndEmitUpdate(open, repository)
+    const bothPrLists = await Promise.all([
```
joshaber: You could destructure the returned array if you wanna be fancy, e.g.:

```suggestion
    const [repositoryPRs, parentPRs] = await Promise.all([
```

app/src/models/repository.ts
```diff
@@ -69,6 +69,11 @@ export class Repository {
   public get isTutorialRepository() {
     return enableTutorial() && this._isTutorialRepository === true
   }
+
+  /** Is this associated with a GitHub fork? */
+  public get isGitHubFork() {
```
joshaber: This is a bikeshed, but I wonder if "GitHub" is adding much to the method name?

app/src/lib/stores/pull-request-store.ts
```diff
@@ -318,7 +328,9 @@ export class PullRequestStore {
         return fatalError('PR cannot have a null base repo')
       }
 
-      const baseGitHubRepo = await upsertRepo(endpoint, pr.base.repo)
+      const baseGitHubRepo = repository.fork
+        ? repository
```
joshaber: I'm not sure I understand what this change is doing, could you walk me through it?

app/src/lib/stores/pull-request-store.ts
```diff
@@ -139,9 +142,16 @@ export class PullRequestStore {
...
-    const { name, owner } = getNameWithOwner(repository)
-    const open = await api.fetchAllOpenPullRequests(owner, name)
-    await this.storePullRequestsAndEmitUpdate(open, repository)
+    const bothPrLists = await Promise.all([
+      fetchOpenPullRequests(api, repository),
+      repository.fork && repository.parent
+        ? fetchOpenPullRequests(api, repository)
+        : Promise.resolve([]),
+    ])
+    await this.storePullRequestsAndEmitUpdate(
```
niik: `storePullRequestsAndEmitUpdate` was written with the intention that the PRs in the list would all target the same repository as the `repository` parameter is referring to and with this change that's no longer true. I'm finding it difficult to fully foresee the implications of this but it's making me nervous.

I think, for one thing, that it could break the delta updating algorithm since that relies on accurately keeping track of the most recently updated PR for a particular repository (that timestamp is then used in subsequent fetches). With this change I believe we could end up in a scenario where we'd be missing new or updated PRs from one of the repositories since we'd be storing the max of `updated_at` from either repository in the `pullRequestsLastUpdated` table for the fork repository.

I also think it will break our optimization to avoid updating the PR list when we get exactly one PR back: https://github.com/desktop/desktop/blob/2870bfb005e569d37c142d5c17f4f7138a44a20c/app/src/lib/stores/pull-request-store.ts#L380-L387

First of all there's always gonna be a minimum of two PRs in that list now (one for the fork and one for the parent) and even if that wasn't the case the `getPullRequest` call relies on the `repository` parameter to look up PRs targeting that repository and with this change it's not gonna find anything for the parent repository.

The last thing I thought of which is a minor thing is that I believe we could now be emitting an update event as a result of `storePullRequestsAndEmitUpdate` even if nothing has changed since `storePullRequests` could find a PR that has been updated for the parent that won't be returned in the `getAll` call.

app/src/lib/stores/app-store.ts
```diff
@@ -5008,20 +5014,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
 
   public async _showPullRequest(repository: Repository): Promise<void> {
-    const gitHubRepository = repository.gitHubRepository
+    const { gitHubRepository } = repository
 
     if (!gitHubRepository) {
       return
     }
 
+    const lookupGitHubRepository =
```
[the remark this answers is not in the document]
outofambit: ah you are totally right. fixed in feb82b1a39fa6f95172238117ac004950a2a6e72

app/src/lib/helpers/pull-request-matching.ts
outofambit: fixed this in d1d4c19b5

app/src/lib/stores/pull-request-store.ts
outofambit: hey @niik i took a run at fixing this up, let me know what you think!

app/src/lib/stores/pull-request-store.ts
```diff
@@ -249,6 +253,28 @@ export class PullRequestStore {
...
+      return []
+    }
+    if (repository.fork && repository.parent && repository.parent.dbID) {
+      const [upstreamRecords, forkRecords] = await Promise.all([
+        this.db.getAllPullRequestsInRepository(repository.parent),
+        this.db.getAllPullRequestsInRepository(repository),
+      ])
+      const records = new Array<IPullRequest>()
+      records.push(...upstreamRecords, ...forkRecords)
+      return records
```
niik: Could we get away with skipping the intermediate array creation here?
```suggestion
      return [...upstreamRecords, ...forkRecords]
```

app/src/lib/stores/app-store.ts
```diff
@@ -1503,6 +1509,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
       this.pullRequestStore.getAll(gitHubRepository).then(prs => {
         this.onPullRequestChanged(gitHubRepository, prs)
       })
+
+      // if there's an upstream, we want to add those pull requests, too
```
niik: I'm confused, wouldn't the parent repository's PRs already have been added by the previous statement given that `pullRequestStore.getAll` now returns PRs for the parent repository as well?

app/src/lib/stores/app-store.ts
```diff
@@ -5105,16 +5111,23 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    const { parent: parentGitHubRepository } = gitHubRepository
 
+    const urlEncodedBranchName = escape(branch.nameWithoutRemote)
+    const urlEncodedOwner = escape(gitHubRepository.owner.login)
+
+    const baseURL =
+      gitHubRepository.fork && parentGitHubRepository
+        ? `${
+            gitHubRepository.htmlURL
+          }/compare/${urlEncodedOwner}:${urlEncodedBranchName}`
```
niik: Could you elaborate on why we're using the `/compare` endpoint here for forks vs the old `/pull/new`?

outofambit:
Closing this PR and continuing this effort with a different strategy in #8825 (and on). Thanks all!

xXxrAinZyian:
Ok

lydz1209:
ktf De : xXxrAinZyianEnvoyé le :8 juillet 2020 20:52À : desktop/desktopCc : SubscribedObjet :Re: [desktop/desktop] Support pull request actions from local forks (#8652) Ok—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or unsubscribe.

PR #8678, opened 2019-11-22
Improve verbiage for deleting branch with PR

### LINKED ISSUE #8416: DX: misleading message when deleting a branch
when deleting a local branch which was already merged you face the following message:

image

I think the "DELETE" Button is misleading because as the message states its not 100% clear whether a button click only deletes the local branch and/or it deletes the remote branch

billygriffin:
Closes #8416 

## Description

The text for the dialog is misleading when you're deleting a branch that has or previously had a pull request associated with it. The problem is that the PR might have been merged in the mean time, so it's not necessarily true that it does have an open PR associated with it. Additionally, the previous verbiage did not make it clear necessarily what would actually happen when you hit "Delete."

### Screenshots

Before: 
Screen Shot 2019-11-22 at 1 39 55 PM

After:
Screen Shot 2019-11-22 at 2 19 51 PM

## Release notes
Notes: No notes


app/src/ui/delete-branch/delete-pull-request-dialog.tsx
```diff
@@ -29,13 +29,14 @@ export class DeletePullRequest extends React.Component<IDeleteBranchProps, {}> {
...
         <DialogContent>
-          <p>This branch has an open pull request associated with it.</p>
+          <p>This branch may have an open pull request associated with it.</p>
           <p>
             If{' '}
             <LinkButton onClick={this.openPullRequest}>
               #{this.props.pullRequest.pullRequestNumber}
             </LinkButton>{' '}
-            has been merged, you can also remove the remote branch on GitHub.
+            has been merged, you can also go to GitHub.com to delete the remote
```
niik: > GitHub.com

It's possible that the PR is stored on a GitHub Enterprise Server instance and in that case I think explicitly saying `GitHub.com` could add some confusion. We can either use a non-specific host terminology or we can change the text if the PR is on GHES, I'll leave that up to you.

outofambit:
```suggestion
            has been merged, you can also go to GitHub to delete the remote
```

PR #8693, opened 2019-11-24
changed wording to match git autocrlf behavior

### LINKED ISSUE #5741: Line endings change warning phrasing is backwards
## Description

The warning that line endings for a file have changed has it backwards, saying the file's line endings have change *from* "what it is now" *to* "what it was", instead of vice versa.  See screenshot below illustrating the problem.  This confused me for a bit when it first happened.

lf-crlf

## Version

* GitHub Desktop: 1.4.0

* Operating system: Windows 10 Pro x64 [10.0.17134.285]

## Steps to Reproduce

* Change the line endings of a file
* View diff in GitHub Desktop
* Note that change described in warning text is backwards

### Expected Behavior
The warning to say "Line endings have changed from 'CRLF' to 'LF'"

### Actual Behavior
The warning actually says "Line endings have changed from 'LF' to 'CRLF'"

jmhammock:
Closes #5741

## Description
Changed wording of autocrlf warning to match the actual behavior of autocrlf.
In Windows: 
If autocrlf = true, then a file with LF line endings will be changed to CRLF during the commit/checkout cycle. 
If autocrlf = input, then a file with CRLF line endings will be changed to LF during the commit/checkout cycle.
If autocrlf = false, then no line ending transformations take place.

Changing the messaging from "...have changed..." to "...will be changed..." is enough to accurately describe what will happen in the next commit.
-

### Screenshots

line-endings-warning

## Release notes

Notes:


niik:
Thanks for this @jmhammock! You can expect this to be part of the next release which is slated for mid December.

Dimas19812:
src/web/mjs/connectors/Neonime.mjs

Dimas19812:
src/web/mjs/connectors/Neonime.mjs

PR #8812, opened 2019-12-18
Update branch protection state for checkout outside of Desktop

### LINKED ISSUE #8790: Switching branches via Cmd line doesn't update branch protection
If a user is switching between branches and goes from a protected branch to a unprotected branch via the command line, then Desktop fails to update the protected branch warning.

Repro:
- Go to protected branch in Desktop
- Use cmd line to swap branches to unprotected branch
- Warning is intact

Note the situation can be reversed, ie if a user goes from an unprotected branch to protected branch that warning is missing.

Version 2.2.4, osx

cc @kuychaco @cheshire137

kuychaco:
Closes #8790

## Description

Previously we updated the branch protection state in the `_checkoutBranch` method of `AppStore`, which doesn't get called when a user checks out a branch on the command line. With this PR, we update the branch protection state when the branch tip in the `repositoryStateCache` is being changed.

## Release notes

Notes: [Fixed] Update branch protection warning when branch switch occurs outside of Desktop

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -762,7 +762,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
             state.tip.kind === TipState.Valid &&
             tip.branch.name !== state.tip.branch.name
           ) {
-            console.warn(tip.branch.name, state.tip.branch.name)
             this.refreshBranchProtectionState(repository)
           }
         }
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -756,6 +756,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
             const prs = state.openPullRequests
             currentPullRequest = findAssociatedPullRequest(branch, prs, remote)
           }
+
+          if (
+            tip.kind === TipState.Valid &&
+            state.tip.kind === TipState.Valid &&
+            tip.branch.name !== state.tip.branch.name
```
outofambit: nice!

PR #8813, opened 2019-12-18
Preferences redesign

### LINKED ISSUE #7746: Please allow disabling of the "Switch branch" dialog that offers to stash changes.
In a big port we're doing, very often we're making changes in whatever branch we're in - the delta matters but the parent doesn't, and we'll flip to master or another branch to actually commit those changes. Most often, this is to a new branch, off of master - so that our migration work continually flows, rather than a big bang at the end.

My steps are:
- Be in a branch (maybe `master`, maybe some random branch from the last change)
- Create a new branch for the changes to go in
- Commit changes to new branch
- Push new branch

Since 2.x, I get a prompt like this:
image

To not be annoyed by this, it forces a workflow on the user of defining their branch *before making their changes*. That's not how we work. If we find a problem, we solve the problem, then commit that work in a branch for PR review. We don't know what the problem is ahead of time. This was painless before, but now results in a prompt each time. If I need to change to master first for some reason, it's 2 prompts. I don't want GitHub Desktop to "help" here...it's an annoyance instead. And a mis-click means even more work. It only offers me opportunities to fail and cause more work. 

Please allow us to disable this prompt - can we add a preference here?
image

/cc @shiftkey

### LINKED ISSUE #8774: Preferences redesign
Since we decided to add the new preference in #7746, we realized we're running out of room in our current preferences dialog. To accommodate this new preference and future new ones, we've decided to redesign the navigation and rearrange some tabs. 

We noted this incorporated three big changes, not including the new stashing preference: 

1. New side navigation
2. Move integration preferences to its own new tab
3. Reword the confirmation dialog preferences for less redundancy

Here are all the future tab states:

### Accounts

Accounts@2x

### Integrations
Integrations@2x

### Git Config
Git@2x

### Appearance
Appearance@2x

### Advanced

Advanced@2x

This tab is where the new stashing preference will live.

cc @kuychaco

kuychaco:
Closes #8774
Closes #7746 

## Description

From @ampinsk 
> A few months ago, we introduced a feature that when you have uncommitted changes, we let you either bring the changes to your new branch or leave them on your current branch. Generally this has been really well received, but some folks have specific workflows where this introduces a lot of extra clicks and pain. 

> We've decided to introduce a preference for this, but our preferences are getting a little overloaded so I assessed a few different ways we could introduce this new setting, including a bigger redesign to our preferences pane

This PR implements the new preference, as well as the redesign of the preferences pane.

### Screenshots

**Before: Tab design**
GitHub_Desktop

**After: Sidebar design**
* New side navigation
* New switching branches preference
* Reworded the confirmation dialog preferences for less redundancy
* Moved integration preferences to its own new tab

GitHub_Desktop-dev

GitHub_Desktop-dev

## Release notes

Notes: Redesign preferences pane and add preference for stashing changes when switching branches.


kuychaco:
For when y'all get back from the holidays... 🙏 

@ampinsk would love a design review from you! Feel free to make modifications as you see fit (commits are welcome!)

@tierninho would appreciate your QA on this!

kuychaco:
@ampinsk is it desired/expected to have that salmon background color for the merge tool section? It's from a `brutalism` class

niik:
> @ampinsk is it desired/expected to have that salmon background color for the merge tool section? It's from a `brutalism` class

@kuychaco That style is one we used to apply in the early days when designs were not finalized yet. @joshaber dubbed it "pink brutalism" and the thinking was that by applying that style a developer could trigger something deep and unsettling within myself or @donokuda and said discomfort would result in us springing into action and finalizing the actual designs before anyone else would see the abhorrent injustice afflicted on the otherwise perfect piece of design that was the Desktop app.

In this case the merge tool selection is behind a dev feature flag because we never ended up shipping it. At some point it might be a good idea to unship all of #2802 since it's been superseded by the in-app conflict resolution work.

outofambit:
@niik @kuychaco just chiming in to say +1 for removing the merge tool ui from the codebase entirely!

donokuda:
> [...] discomfort would result in us springing into action and finalizing the actual designs before anyone else would see the abhorrent injustice afflicted on the otherwise perfect piece of design that was the Desktop app.

Poppin' in here to say that the plan backfired and I ended up embracing pink brutalism 😍

joshaber:
>>[...] discomfort would result in us springing into action and finalizing the actual designs before anyone else would see the abhorrent injustice afflicted on the otherwise perfect piece of design that was the Desktop app.
>
>Poppin' in here to say that the plan backfired and I ended up embracing pink brutalism 😍

Bloopin' in here to say that @donokuda is a creative genius without equal in this world.

kuychaco:
I'm delighted by all of the ploppin' and bloopin' 😆 

I've opened an issue to track removing the merge tool UI -- https://github.com/desktop/desktop/issues/8875

And I've fixed all of the issues that @tierninho found (thanks @tierninho!)

So this PR should now be ready for design review (@ampinsk) and another round of QA (@tierninho) 🙏

app/styles/ui/_preferences.scss
```diff
@@ -1,4 +1,12 @@
 #preferences {
+  .preferences-container {
+    display: flex;
+    .tab-container {
+      border-left: 1px solid var(--box-border-color);
```
outofambit: there's a variable for this already at https://github.com/desktop/desktop/blob/cdbfb04ff1afda377e59e8e5f158da9f446442d0/app/styles/_variables.scss#L198 that you can use

app/styles/ui/_tab-bar.scss
```diff
@@ -3,6 +3,15 @@
   flex-direction: row;
   height: var(--tab-bar-height);
 
+  &.vertical {
+    flex-direction: column;
+    align-items: stretch;
+    height: auto;
+    flex-shrink: 0;
+    padding-top: var(--spacing);
+    padding-bottom: var(--spacing);
```
outofambit: not important, but this could also be abbreviated
```scss
padding: var(--spacing) 0;
```

app/styles/ui/_tab-bar.scss
```diff
@@ -108,4 +117,45 @@
...
+  &.vertical &-item {
+    // Reset styles from global buttons
+    cursor: pointer;
+    border: none;
+    box-shadow: none;
+    border-radius: var(--border-radius);
+    color: var(--text-color);
+    background: var(--background-color);
+    font-family: var(--font-family-sans-serif);
+    font-size: var(--font-size);
```
outofambit: looks like these are copied from `&.switch &-item`, perhaps they could be shared between the two classes somehow? maybe with a shared base class or something?

ampinsk:
Thank you @kuychaco his is looking so good! 💕 A couple of tiny things I noticed that I wasn't sure how to fix on my own: 

Screen Shot 2020-01-07 at 1 07 47 PM

In this screenshot:
- Looks like the text isn't lining up on the left (you can see based on the red line)
- The non-selected tab icons should be gray 

Screen Shot 2020-01-07 at 1 40 51 PM

In this screenshot:
- The background color behind the sidebar tabs in dark mode is darker than the tabs, they should be the same dark gray


|Before|After|
|---|---|
|Screen Shot 2020-01-07 at 1 37 20 PM|Screen Shot 2020-01-07 at 1 39 18 PM|

- For advanced settings, is it possible to get the spacing closer to the screenshot on the right? I tightened the spacing between the row items and increased the spacing between the sections

Let me know if you want to pair on any of this! ✨

kuychaco:
Okay! Styling fixes complete! Thanks @outofambit and @ampinsk for your great input 🙏 

@ampinsk how does this look for the Advanced section... we can pair if you want to further tweak the spacing. Or you can make changes to the values I used in https://github.com/desktop/desktop/pull/8813/commits/5bbc79d9079aafead3b899b070a9cad0bf7f465a

|Current State|Your Design|
|---|---|
|GitHub_Desktop-dev|Screen Shot 2020-01-07 at 1 39 18 PM|

Otherwise I think this is ready for a full code review! /cc @outofambit

ampinsk:
That spacing looks great!! 

Just one last thing I noticed from your screenshot is the non-selected icons should be gray 😄 Otherwise this seems ready to go!

app/src/lib/stores/app-store.ts
```diff
@@ -1769,6 +1777,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
       askForConfirmationOnForcePushDefault
     )
 
+    const strategy = localStorage.getItem(
+      uncommittedChangesStrategyKindKey
+    ) as UncommittedChangesStrategyKind
```
outofambit: I'd really like to avoid using `as` here, though I'm not sure of a good alternative approach yet. `localStorage.getItem` can return `undefined` if the given key doesn't exist, and this implementation tells the compiler that 'undefined' is a valid `UncommittedChangesStrategyKind`, which will result in runtime errors.

app/src/lib/stores/app-store.ts
```diff
@@ -3034,20 +3052,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-          branchToCheckout: foundBranch,
+      if (hasChanges) {
+        if (uncommittedChangesStrategy.kind === askToStash.kind) {
+          this._showPopup({
+            type: PopupType.StashAndSwitchBranch,
+            branchToCheckout: foundBranch,
+            repository,
+          })
+          return repository
+        }
```
outofambit: 😋

app/src/models/uncommitted-changes-strategy.ts
```diff
@@ -20,3 +24,25 @@ export const askToStash: UncommittedChangesStrategy = {
...
+      return moveToNewBranch
+    case UncommittedChangesStrategyKind.StashOnCurrentBranch:
+      return stashOnCurrentBranch
+    default:
+      return assertNever(
+        kind,
+        `Unknown UncommittedChangesStrategyKind: ${kind}`
+      )
+  }
+}
```
outofambit: I have some concerns about this function, but I think the core of my concern is that it took me a while to figure out what it was doing. Perhaps documenting this to clarify that its primary purpose is to map a kind to an actual `UncommittedChangesStrategy`?

It feels to me like there's a simpler version of this overall approach, but I'm not sure what it is yet...Is the main reason this exists so that we can store an `UncommittedChangesStrategyKind` in `AppState` and convert it to an `UncommittedChangesStrategy` whenever we use it? perhaps we could store the result of this function in `AppState` and avoid doing this conversion in many places?

app/src/ui/branches/branches-container.tsx
```diff
@@ -45,6 +45,10 @@ interface IBranchesContainerProps {
   readonly isLoadingPullRequests: boolean
 
   readonly currentBranchProtected: boolean
+
+  readonly selectedUncommittedChangesStrategy: UncommittedChangesStrategy
+
+  readonly hasChangesAndStash: boolean
```
outofambit: non-blocking idea for a maybe more legible name:
```suggestion
  readonly couldOverwriteStash: boolean
```

app/src/ui/tab-bar.tsx
```diff
@@ -30,11 +33,17 @@ export class TabBar extends React.Component<ITabBarProps, {}> {
...
     return (
       <div
         className={
           'tab-bar ' +
-          (this.props.type === TabBarType.Switch ? 'switch' : 'tabs')
+          (type === TabBarType.Switch
+            ? 'switch'
+            : type === TabBarType.Vertical
+            ? 'vertical'
+            : 'tabs')
```
outofambit: would be nice to pull this out into a little `switch...case` with an `assertNever`, but not a big deal. (non-blocking)

app/src/ui/toolbar/branch-dropdown.tsx
```diff
@@ -45,6 +46,10 @@ interface IBranchDropdownProps {
 
   /** Whether this component should show its onboarding tutorial nudge arrow */
   readonly shouldNudge: boolean
+
+  readonly selectedUncommittedChangesStrategy: UncommittedChangesStrategy
+
+  readonly hasChangesAndStash: boolean
```
outofambit: same thing as above regarding variable name

PR #8825, opened 2019-12-20
[Fork] Initial PullRequestCoordinator (P)refactor

### LINKED ISSUE #8549: [Fork] "Create pull request" is still a suggestion even after I've created a PR for the current branch on my fork

outofambit:
supports #8549

## Background

While discussing #8652, @niik and I agreed we needed an layer between `AppStore` and the workings of `PullRequestStore` and `PullRequestUpdater`. These relatively simple classes were getting bloated with additional logic and it felt like it would be easier to manage if there was something to contain that logic, outside of `AppStore`.

## Description

This pull request begins this work by introducing this layer as `PullRequestCoordinator` (at least that's what I'm calling it for now) but retaining the current application's behavior.

### Small Changes

One notable change is that the PullRequestCoordinator communicates with the AppStore in reference to a `Repository`, not a `GitHubRepository` as it did previously. This small but significant change will allow the PullRequestCoordinator to compartmentalize concerns around matching remote GitHub repositories with local repositories.

AppStore manages state for a single "selected" repository (the one currently in view in the UI), so this is a better fit for AppStore's concerns regardless.

### Next Steps

Further changes to close #8549 will happen in future pull requests. They vast majority of these changes should be contained within the new `PullRequestCoordinator`.


app/src/models/repository.ts
```diff
@@ -77,6 +77,17 @@ export type LinkedWorkTree = WorkingTree & {
   readonly head: string
 }
 
+/** Identical to `Repository`, except it **must** have a `gitHubRepository` */
+export type RepositoryWithGitHubRepository = Repository & {
+  readonly gitHubRepository: GitHubRepository
+}
```
outofambit: One potentially controversial thing I've done is add a `RepositoryWithGitHubRepository` type which guarantees a `Repository` model has a `GitHubRepository` model in it. I did this primarily to make the null checks a little easier to read, since the `PullRequestCoordinator` is only be concerned with repositories that are connected to GitHub somehow, while `AppStore` and `RepositoriesStore` are concerned with all local repositories. This is kind of a band-aid for our domain model but that's a (long) story for another day.

app/src/lib/stores/app-store.ts
```diff
@@ -1624,36 +1613,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   private startPullRequestUpdater(repository: Repository) {
-    if (this.currentPullRequestUpdater) {
-      this.stopPullRequestUpdater()
```
niik: The removal of this `stop` operation from the `startPullRequestUpdater` method makes me a teeny bit nervous. I can' point towards anything that will break by removing this but given that `pullRequestCoordinator.stopPullRequestUpdater` is both cheap and idempotent could we maintain the current behavior just to be safe?

app/src/lib/stores/app-store.ts
```diff
@@ -1624,36 +1613,22 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-    }
-
     // We don't want to run the pull request updater when the app is in
```
niik: Suggested change for the comment above about maintaining existing behavior.

```suggestion
    this.stopPullRequestUpdater()

    // We don't want to run the pull request updater when the app is in
```

app/src/lib/stores/app-store.ts
outofambit: ah I see. even though we do this in `PullRequestCoordinator.startPullRequestUpdater`, it might not get called! fixed in https://github.com/desktop/desktop/pull/8825/commits/8fb41ce0e2e7201b1065487d3cae512c5287fe0c

outofambit:
Addressed @niik's concerns, going to merge so I can get the next PR in this series ready for review

PR #1834, opened 2017-05-30
Silently delete local branch if remote one is already deleted

terales:
See https://github.com/desktop/desktop/issues/1829

Tested manually:
* delete branch which is already deleted on GitHub.com,
* delete branch which was published and is available on GitHub.com.

Both cases works fine.

I want to create unit test, but I can't understand what fixture should I take into. Can you suggest something, please?

Issue #1829, opened 2017-05-30
Failing to delete a remote branch isn't error-worthy?


joshaber:
1. Create a new branch.
1. Go to dotcom.
1. Open a PR from branch
1. Merge PR and delete branch from dotcom.
1. Go back to Desktop.
1. Delete branch.
1. Get an error about how it couldn't delete the remote branch.

It makes sense that the remote branch deletion failed since we just deleted the branch from the Pull Request view. But it's not noteworthy for the user.

Of course the catch is that sometimes it _might_ be noteworthy. Maybe we should fetch + prune before trying to delete the remote branch?

terales:
See a record: https://youtu.be/xCVKG1d93Ow

It's really unclear. Looks like there was an error about removing *local* branch.

Maybe we can add a n indication that the local branch was already successfully deleted and a remote one was already deleted?

terales:
[Found](https://stackoverflow.com/a/30524983/1363799) a more solid solution. We can check if remote branch exists with
```
git ls-remote --heads git@github.com:desktop/desktop.git branch-name
```

If branch exists it would respond with zero and some data,
if there is no such branches it [would respond with zero and empty data](https://git-scm.com/docs/git-ls-remote.html#git-ls-remote---exit-code).

If there is no such branches we shouldn't run [`deleteBranch` command](https://github.com/desktop/desktop/blob/43be7fabe83f42b6e52db33fde8599169c127f06/app/src/lib/git/branch.ts#L49).

Should I close previous pull request and implement this explicit check?

shiftkey:
@terales that seems like a nicer solution than running a full fetch. This can also be tweaked to use the remote name instead of the full URL:

```
git ls-remote --heads origin branch-name
```

I'm happy for you to fold this into #1834 and see how that works.

terales:
@shiftkey I'll update pull request with another solution and remove skipped test in three days. Fell free to assign this issue to me.

shiftkey:
Fixed by #1834

PR #8836, opened 2019-12-27
Atomic remote branch delete and resilience

### LINKED ISSUE #8831: Problem with removing branch from branches list
Hi All, 
I found a small bug (or only inconvenience). When I try delete local branch with remote branch and this branch was deleted a moment earlier via github.com website, this branch is still on the branches list (GH Desktop) and remove action does not return any error / info message. 

This branch disappears from list only after fetching origin.

### Version & OS

GitHub Desktop 2.2.4
Windows 10 Pro

1. Go to github.com and delete branch
2. Go to GitHub Desktop and delete remote branch **without** fetching origin
3. Removing modal box disappears without any info / error message
4. Removed branch is still on the branches list

### Expected behavior

Information with removing error - "Branch does not exist on remote". Branch should be remove from branches list.

### Actual behavior

Removing modal box disappears without any info / error message. Removed branch is still on the branches list.

niik:
Closes #8831 

## Description

When a user deleted a branch in early versions of GitHub Desktop and indicated that they wanted to delete the remote branch as well we would attempt to delete the branch on the remote and show an error if that didn't work.

That behavior gave rise to https://github.com/desktop/desktop/issues/1829 which stipulated that showing an error when the remote branch we tried to delete was already deleted wasn't really the best UX. Following that https://github.com/desktop/desktop/pull/1834 introduced new logic that attempted to first verify that the branch did exist on the remote before trying to delete it and if it didn't exist then it would simply skip the remote branch deletion.

This approach has two shortcomings, firstly it introduces a second (fairly expensive) network operation (`ls-remote`) and second it fails to delete the `refs/remotes/xyz/name` ref if the `ls-remote` call determines that no such ref exists on the remote.

The approach introduced in this PR instead always attempts to delete the remote ref and then looks at the error message (if any) returned to figure out if the reason the push failed was because the remote ref doesn't exist. If the remote ref doesn't exist we now consider it safe to delete the `refs/remotes/xyz/name` ref as well which prevents the confusing behavior described in #8831. A nice side-effect of this is that remote ref deletes are now atomic (in the previous logic it was possible that the existence of a remote ref could change between the `ls-remote` and the `push` operation).

Depending on an error message in this scenario warrants test coverage to ensure that we detect if the underlying error message ever changes so I've added that as well.

## Release notes

Notes: [Fixed] Remote branch could remain in branch list after deleting it


app/test/helpers/repositories.ts
```diff
@@ -249,3 +250,16 @@ export async function setupTwoCommitRepo(): Promise<Repository> {
   await makeCommit(repo, secondCommit)
   return repo
 }
+
+/**
+ * Sets up a local fork of the provided repository
+ * and configures the origin remote to point to the
+ * local "upstream" repository.
+ */
+export async function setupLocalForkOfRepository(
```
outofambit: this is a great helper, thank you!

app/src/lib/git/update-ref.ts
```diff
@@ -36,17 +36,13 @@ export async function updateRef(
...
   repository: Repository,
   ref: string,
-  reason: string
-): Promise<true | undefined> {
-  const result = await git(
-    ['update-ref', '-m', reason, '-d', ref],
-    repository.path,
-    'deleteRef'
-  )
+  reason?: string
```
outofambit: What's the reasoning for making this argument optional? i don't think its critical, but seems like a nice bit of logging, just in case. (non-blocking)

app/src/lib/git/branch.ts
```diff
@@ -94,41 +85,26 @@ export async function deleteBranch(
...
     // If the user is not authenticated, the push is going to fail
     // Let this propagate and leave it to the caller to handle
-    await git(args, repository.path, 'deleteRemoteBranch', opts)
+    const result = await git(args, repository.path, 'deleteRemoteBranch', {
+      env: envForAuthentication(account),
+      expectedErrors: new Set<DugiteError>([DugiteError.BranchDeletionFailed]),
+    })
+
+    // It's possible that the delete failed because the ref has already
+    // been deleted on the remote. If we can identify that specific
```
outofambit: (non-blocking) small doc wording tweak to indicate that this is in fact what we're doing. for a minute i thought this was a speculative comment
```suggestion
    // been deleted on the remote. If we identify that specific
```

app/src/lib/git/update-ref.ts
niik: Adding a reflog entry when deleting a ref doesn't do much in terms of logging because the reflog file for that ref will get deleted as well.

The reason we have the `reason` parameter at all is because we previously only used it to delete the `HEAD` reference when undoing the initial commit which is a super special case because the HEAD reflog file will never get deleted so we're able to tack on a message as part of that operation.

I'll update the parameter documentation to make this clear

outofambit: gotcha, that makes sense! thank you!

PR #8906, opened 2020-01-13
Sign Windows installer using SHA256 digest

### LINKED ISSUE #8901: "Windows protected your computer" during installation
### Describe the bug

Windows 10 latest build shows protection warning screen whenever I try installing desktop after downloading from https://desktop.github.com/

### Version & OS

GitHub Desktop Version: 2.2.4

Winver: Windows 10 Home, 1909 with latest updates, OS Build 18363.418

1. Download the .exe from https://desktop.github.com/
2. Try opening it from the browser download panel
3. Windows moans about unknown publisher

### Expected behavior

Would expect the download to be digitally signed to avoid that warning.

### Actual behavior

Windows moans about unknown publisher.

### Screenshots

image

niik:
Closes #8901

## Description

According to https://www.digicert.com/code-signing/ev-authenticode-certificates.htm Windows may have stopped supporting SHA1 as the file digest algorithm. I haven't been able to find any definitive answer or authoritative document. The only Microsoft sources I've found seem to indicate that the 2020 date is only for executables signed and timestamped way earlier than ours. 

> Microsoft will support SHA1 Code Signing Certificates until Jan 1, 2020. Microsoft recommends using SHA-256 certificate/digest algorithm/timestamp for all applications. Microsoft has not yet released a SHA1 deprecation policy for drivers. For more information, refer to the Windows Enforcement of Authenticode Code Signing and Timestamping page.

While we're definitely using a SHA2 certificate and have been for a long time it seems we may be using SHA1 as the digest algorithm and it's possible that there's been some change to disallow that recently.

image

I'm opening this in order to do a test release so that we have some options should it turn out that we need to change our signing mechanism.

## Release notes

Notes:


tierninho:
> I haven't been able to find any definitive answer or authoritative document. 

Hopefully this is relevant:
https://techcommunity.microsoft.com/t5/windows-hardware-certification/starting-in-march-2018-packages-signed-using-a-sha-1-digest/ba-p/364916

niik:
@outofambit @kuychaco Luckily the originating issue here turned out to be a fluke and there's no immediate danger from us using SHA1. That said switching to SHA256 is still the recommended algorithm and both VS Code and Slack use it so I think we should land this in order to not have to scramble the day it finally goes away as a file digest algorithm as well.

I did a test release and verified on my Windows VM. Patched Windows 7 and up support SHA256 and MS has signed their own windows updates with SHA2 since July 2019 so it's well supported. I propose we include this in tomorrow's release.

PR #8929, opened 2020-01-15
[Fork] Open PR from fork in browser

### LINKED ISSUE #8928: Direct user to correct URL if PR lives on the upstream repo
When testing this issue https://github.com/desktop/desktop/issues/8549 with this scenario:

> Example: I'm in billygriffin/desktop (a fork of desktop/desktop) on my branch new-work, and I make some commits and create a pull request targeting the development branch of desktop/desktop. In the app, I'm still shown the "Create pull request" suggested next step even though I've already created one.

I noticed that we route the user to the forked repo if they click `Show Pull Request` (aka `cmd-R`) from the Repository menu after completing the above steps. 

Doing so results in a 404 page as the PR is located on the upstream repo.

Screen Shot 2020-01-15 at 12 33 16 PM

tested in 2.2.5-beta1, OSX

outofambit:
Closes #8928
(pulled out of #8652)

## Description

- Always use the pull request's base repo url to find the pull request online

## Release notes

Notes: [Fixed] Open fork pull requests in browser


outofambit:
> LGTM!
> 
> I _did_ wanted to surface two things, but I believe it is ok to leave them as-is:
> 
>     * The `http.ts:148 GET https://api.github.com/repos/tierninho/TestWebAppProject/commits/refs/pull/67/head/status 404 (Not Found)` error is still shown in the console logs.
> 
>     * The `#67` badge is shown in the nav header but not in the PR tab. Makes sense once you realize it is on the upstream repo.

@tierninho thanks! yeah the first issue is a pre-existing error that i haven't been able to connect with any user-facing problems (yet). the second issue i've seen as well and i think its worth us considering if we want to hide that badge. (this would be in a separate PR) @ampinsk any thoughts?

ampinsk:
> the second issue i've seen as well and i think its worth us considering if we want to hide that badge. (this would be in a separate PR) @ampinsk any thoughts?

@outofambit and I chatted through this in Slack and we think we should hide these upstream PR badges until we eventually start displaying upstream PRs in the PRs list. I'll open an issue!

PR #8941, opened 2020-01-18
Toggle co-authors icon

MartinBenninger:
## Description

- This adds a different icon for removing co-authors. Currently, the same icon is used for both adding and removing co-authors. I think this is counterintuitive so I made this change to toggle what icon is shown.

### Screenshots

Add co-authors (unchanged):  
Add co-authors

Remove co-authors before:  
Remove co-authors before

Remove co-authors after:  
Remove co-authors after

As you can see in the screenshots, the little blue icon has been changed from a plus to a minus.

## Release notes

Notes: Add a remove co-authors icon.


niik:
Hi @MartinBenninger, thanks so much for wanting to contribute! Thanks especially for providing detailed screenshots of your change, that always helps!

In this instance the choice of a single icon is an intentional one. I think between the difference in color, the altered description text, and the fact that the co-author text box represents such a big change in the UI when clicking the icon it should be clear what the active/inactive states of the component is though I can see how the plus sign could be perceived as indicating a toggle state on its own.

Given that we haven't heard this particular feedback until now I'm reluctant to make any changes in the iconography at this time but if we see more reports of this behavior being confusing we can always revisit and consider alternatives.

cc @ampinsk FYI since I know you have been looking at icons recently.

MartinBenninger:
Hi @niik, thanks for the information. I just assumed that this was unintentional 😊 

I agree that the other UI changes make it clear enough what the icon does.

Also, I think I just now understood the icon the way it was intended. I was thinking the plus symbol was for the "add" in "**Add** co-authors" but I see now that it (together with the user icon) is meant to represent the "co-author" part. 😆 

Silly me.

PR #8937, opened 2020-01-17
[Fork] Hide Branch's PR Badge if for Upstream PR

outofambit:
Closes #8936

## Description

- add feature flag for showing upstream PRs (since we plan to revisit this post-2.3 release)
- use flag in existing pr list filtering
- use flag and hide PR badge if PR is based on upstream repo

### Screenshots

Screen Shot 2020-01-17 at 9 43 48 AM

PR #8952, opened 2020-01-21
Lookup commit status using the PRs base repository

niik:
## Description

In https://github.com/desktop/desktop/pull/8929#pullrequestreview-343602548 @tierninho remarked that he was seeing error messages about failed fetches

> The `http.ts:148 GET https://api.github.com/repos/tierninho/TestWebAppProject/commits/refs/pull/67/head/status 404 (Not Found)` error is still shown in the console logs.

This is because the `<CommitStatus>` component (or rather consumers of that component) were built when there could only ever be one candidate repository for looking up a commit status. Since that's changed now with the fork flow it's time to update that assumption and rely on the pull requests base repository instead (since that's where any commit checks will run and apply).

Note that I expect the error raised by @tierninho is no longer reproducible as of https://github.com/desktop/desktop/pull/8937 since I think that removes the last instance where a PR badge or CommitStatus component is rendered for an upstream PR. Nevertheless we should future-proof this.

### DIFF
--- a/app/src/ui/branches/pull-request-list.tsx
+++ b/app/src/ui/branches/pull-request-list.tsx
@@ -176,7 +176,7 @@ export class PullRequestList extends React.Component<
         author={pr.author}
         matches={matches}
         dispatcher={this.props.dispatcher}
-        repository={this.props.repository}
+        repository={pr.base.gitHubRepository}
       />
     )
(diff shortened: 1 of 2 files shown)


O330oei:
now

PR #8958, opened 2020-01-22
Update error message regex when push fails due to missing workflow scope

niik:
Follow-up to https://github.com/desktop/desktop/pull/8357.

## Description

Today I learned from @riha that we no longer provide an easy route to adding a workflow scope when needed. He had an old token and tried to push a workflow file and instead of being greeted by the nice dialog from #8357 he got the generic push failed dialog with raw Git output.

Turns out github.com updated their error message to be more user friendly (and I applaud the intention). Unfortunately for us that meant that Desktop no longer detects the specific workflow error.

Ideally we'd have some sort of language neutral error code to rely on here but since we don't we can only adapt to the new state of affairs. Luckily this dialog will become less and less important as more users automatically acquire the workflow scope as part of signing in for the first time.

### Before

image

### After

image

## Release notes

Notes: [Fixed] Detect and inform when push fails due to missing workflow scope

### DIFF
--- a/app/src/ui/dispatcher/error-handlers.ts
+++ b/app/src/ui/dispatcher/error-handlers.ts
@@ -517,17 +517,9 @@ export async function refusedWorkflowUpdate(
     return error
   }
 
-  const rejectedPath = match[1]
-  const pathIsLikelyWorkflowFile =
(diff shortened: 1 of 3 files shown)


niik:
> thank you for jumping on this @niik! this seems worthwhile to include in `2.3.0`

I agree

PR #8960, opened 2020-01-23
add error handler for fork unready errors

### LINKED ISSUE #8949: [Fork] Detect failures with new fork due to async fork creation
raised in https://github.com/desktop/desktop/pull/8889#pullrequestreview-344482340 by @niik 

>Looking at https://developer.github.com/v3/repos/forks/#create-a-fork
>
> > Note: Forking a Repository happens asynchronously. You may have to wait a short period of time before you can access the git objects. If this takes longer than 5 minutes, be sure to contact GitHub Support.
>
> Should we perhaps make sure that the repository is all set up and ready for the user before we dismiss the fork dialog?

This issue is the first step towards this potential problem. Let's create a mechanism that will detect when/if this happens so we can evaluate addressing this in code.

outofambit:
Closes #8949 

fork-error-retry

## Description

- adds error handler to catch errors from created but not ready forks and log them
- show a dialog alerting the user to the issue and inviting them to try again momentarily
- if a `RetryAction` is available, dialog will offer a "Retry" button, otherwise just shows default "Ok" button

## Release Notes

Notes: [Added] Provide retry option if unable to reach a fork that's mid-creation


billygriffin:
Thanks @outofambit, this is great. My only concern is that I’m confused about when to click “Retry” because it says “in just a moment,” and that makes me uncertain about whether I can go ahead and click it or whether I need to wait some period of time, and if so how long? Can we make that clearer what their action should be and when?

app/src/ui/forks/fork-unready.tsx
```diff
@@ -0,0 +1,61 @@
...
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+        type="error"
+        id="fork-unready"
+      >
+        <DialogContent>
+          {`We were unable to reach `}
+          <strong>{this.props.repository.gitHubRepository.fullName}</strong>
+          {` because fork creation hasn't completed on GitHub.
+          Please retry in just a moment.`}
```
outofambit:
```suggestion
          Please try again.`}
```

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -627,6 +628,52 @@ export async function insufficientGitHubRepoPermissions(
   return null
 }
 
+const forkUnreadyErrorMessageRe = /fatal: remote error: access denied or repository not exported: (.+)/s
```
niik: This error is a pretty [generic one from Git's dameon process]( https://github.com/git/git/blob/228f53135a4a41a37b6be8e4d6e2b6153db4a8ed/daemon.c#L308-L314) that indicate (in our case) that something went wrong when attempting to read from the repo on disk. Looking at previous support requests at GitHub it appears that this isn't an uncommon error ot see when GitHub file servers are having availability issues for example.

As such I feel uncomfortable having a generic catch-all error handler react so specifically to this generic error.

I think there's a few options. Either we move the detection of a live fork to an in-line fetch loop in the create fork flow itself, or we find a way of determining in this handler whether the repository is recently forked, or we skip handling this all-together until we're certain it's a problem.

outofambit: Thanks @niik! I’m going to close this until we know it’s an issue. We can revisit then with more context about how this presents in desktop.

Issue #8971, opened 2020-01-23
[Metrics] Forking flow


nerdneha:
👋 We'd like to get a better idea of how helpful the Forking flow (#8509) is that we're creating and whether the solution we've proposed makes a difference in the problem that people experience when they have worked on a repo they don't have write permissions to and want to keep their work. 

Here are some initial thoughts about what we're trying to measure here:
 1. How many people are committing their work to a repo that they don't have write permissions for? How is that number changing each month? 

This will help us know whether this is a consistent problem that people run into.

 2. Of those people who try to push their work, how many people create a fork because of the functionality that we provide?

This will give us a signal on whether the solution we're providing is one that our users understand and want to use.

cc @outofambit @billygriffin

billygriffin:
> How many people are trying to push their work to a repo that they don't have write permissions for? How is that number changing each month?

What if this was "how many people are committing" instead of "how many people are trying to push"? In other words, are you doing work on a repo where you don't have write access? And how does that compare to the number who end up creating a fork in order to contribute instead of having that work effectively orphaned. Since there's a path to create a fork prior to actually attempting to push, I feel like the commit level might be a better level of granularity.

nerdneha:
@billygriffin that's a great point. i agree that we'd want to measure from the minute we expose a path out -- i've gone ahead and updated the description to reflect that!

billygriffin:
Awesome @nerdneha, thank you! I think your updated description accurately describes what @outofambit and I discussed as well. We want to be able to understand something along the lines of: "Last month, x% of Desktop users committed to [# of repos] repos that they didn't have permission to push to. This month, x% of those users forked [# of repos] using GitHub Desktop, bringing down the number of repos where someone committed work they couldn't push by x%."

To be able to understand something like this, we'll need to know:
1. total repos where a user without write access commits at least once
2. total number of commits in a repo where they don’t have write access
3. number of forks created from desktop (potentially split between the two paths to create a fork, though not necessarily)

@outofambit also mentioned wanting to know the conversion rate for the "push failed" dialog to fork creation. In other words, if people are hitting that dialog because they don't have the correct permissions, are they creating a fork via Desktop nearly all the time, or is there something else we might be missing?

PR #8978, opened 2020-01-24
Add Number Array Helpers for Local Storage

outofambit:
supports #8971 

## Description

- extracts some logic written for storing recent repositories as a list of numbers in Local Storage into helpers so we can use them elsewhere in the codebase (like #8980)

### DIFF
--- a/app/src/lib/local-storage.ts
+++ b/app/src/lib/local-storage.ts
@@ -87,3 +87,44 @@ export function getNumber(
 export function setNumber(key: string, value: number) {
   localStorage.setItem(key, value.toString())
 }
+
+/**
+ * Retrieve an array of `number` values from a given local
+ * storage entry, if found. The array will be empty if the
+ * key doesn't exist or if the values cannot be converted
+ * into numbers
+ *
+ * @param key local storage entry to read
+ */
+export function getNumberArray(key: string): ReadonlyArray<number> {
+  const numbersAsText = localStorage.getItem(key)
+  let values = new Array<number>()
+  if (numbersAsText !== null) {
+    try {
+      values = numbersAsText
+        .split(NumberArrayDelimiter)
+        .map(n => parseInt(n, 10))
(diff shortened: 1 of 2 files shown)


app/src/lib/local-storage.ts
```diff
@@ -87,3 +87,44 @@ export function getNumber(
...
+ * @param key local storage entry to read
+ */
+export function getNumberArray(key: string): ReadonlyArray<number> {
+  const numbersAsText = localStorage.getItem(key)
+  let values = new Array<number>()
+  if (numbersAsText !== null) {
+    try {
+      values = numbersAsText
+        .split(NumberArrayDelimiter)
+        .map(n => parseInt(n, 10))
```
niik: Given that we call this method get**Number**Array we should probably preserve the accuracy of the values. Numbers in javascript can be both floating point and integers but this method assumes integers.

Given that there's no type support specifically for integers I think we should make this method capable of handling all numbers. Also, while more academic it's possible to `.toString` an integer that won't be parseable by `parseInt`, such as `parseInt(Math.pow(10, 21).toString(), 10)`

```suggestion
        .map(parseFloat)
```

app/src/lib/local-storage.ts
```diff
@@ -87,3 +87,44 @@ export function getNumber(
...
+        .filter(n => !isNaN(n))
+    } catch {
+      // clear array, just in case
+      values.splice(-1)
```
niik: I'm not wild about this. There's no way the above block can throw and assign something to the values array. Since we're only using the value to return and since we're not doing any post processing I think this code can be made more clear by not allocating this mutable value. Something like

```ts
return (localStorage.getItem(key) || '')
  .split(NumberArrayDelimiter)
  .map(parseFloat)
  .filter(n => !isNaN(n))
```

None of these methods can throw so we can get rid of the try catch and there's always going to be an empty array to return if the values couldn't be parsed.

PR #8979, opened 2020-01-24
[Forks] Metrics, Part 1

outofambit:
first part of #8971 

## Description

Instruments two new metrics:

- `commitsInRepositoryWithoutWriteAccess` - How many commits did the user make in a repo they don't have `write` access to?
- `forksCreated` - How many forks did the user create from Desktop?

### DIFF
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -2245,7 +2245,9 @@ export class Dispatcher {
   }
 
   /**
-   * Fork created
+   * Increments the `forksCreated ` metric` indicating that the user has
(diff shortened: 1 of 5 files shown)


app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -2243,4 +2243,11 @@ export class Dispatcher {
   public recordTutorialRepoCreated() {
     return this.statsStore.recordTutorialRepoCreated()
   }
+
+  /**
+   * Fork created
```
niik:
```suggestion
   * Increments the `forksCreated ` metric` indicating that the user has
   * elected to create a fork when presented with a dialog informing
   * them that they don't have write access to the current repository.
```

Issue #8875, opened 2020-01-07
Remove Merge Tool UI


kuychaco:
Extracted from https://github.com/desktop/desktop/pull/8813#issuecomment-570256202

> the merge tool selection is behind a dev feature flag because we never ended up shipping it. At some point it might be a good idea to unship all of #2802 since it's been superseded by the in-app conflict resolution work.

PR #9007, opened 2020-01-29
2.3.2-beta1 Release

kuychaco:
- [x] finalize release notes
- [x] release!

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -1,5 +1,6 @@
 {
   "releases": {
+    "2.3.2-beta1": [],
     "2.3.1": [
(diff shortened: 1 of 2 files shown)


changelog.json
```diff
@@ -1,5 +1,6 @@
 {
   "releases": {
+    "2.3.2-beta1": [],
```
kuychaco: We use `-beta0` to indicate there are no new changes on top of what's currently on production. I'm using `-beta1` here because technically we have a change (#8875), but it's not user-facing so there is no changelog entry.

PR #9014, opened 2020-01-30
Account for peeps who maybe shouldn't create fork

### LINKED ISSUE #9013: Improve verbiage in fork dialog for lacking permissions

I can imagine a relatively common use case where people don't have write access to a repository being that they actually *should* have access but they haven't yet gotten the proper permissions. In these cases, nudging them to create a fork is actually not what we want.

### Proposed solution

I'm thinking that we should adjust the dialog verbiage slightly to indicate that they should either talk to an admin about their permissions if they should be able to push to the repository, or if not then create a fork.

billygriffin:
Closes #9013 

## Description

This adjusts the verbiage for the create a fork dialog to include guidance to contact a repo admin in the case where you probably *should* have permission to push. This is basically to try to prevent people creating forks who should just get their org permissions straightened out instead.

Note for reviewers: My implementation is probably super hacky with just using a couple line breaks explicitly. If there's a better way to do this, please feel free to suggest a better way or just change it.

### Screenshots
Before:
Screen Shot 2020-01-30 at 1 42 05 PM

After:
Screen Shot 2020-01-30 at 1 35 49 PM

## Release notes

Notes: No notes

### DIFF
--- a/app/src/ui/forks/create-fork-dialog.tsx
+++ b/app/src/ui/forks/create-fork-dialog.tsx
@@ -101,7 +101,7 @@ function renderCreateForkDialogContent(
       <DialogContent>
         {`It looks like you don’t have write access to `}
         <strong>{repository.gitHubRepository.fullName}</strong>
-        {`. If you should have permissions, please check with a repository administrator.`}
+        {`. If you should have permission, please check with a repository administrator.`}
         <br />
         <br />
         {` Do you want to create a fork of this repository at `}
--- a/app/src/ui/forks/create-fork-dialog.tsx
+++ b/app/src/ui/forks/create-fork-dialog.tsx
@@ -99,16 +99,18 @@ function renderCreateForkDialogContent(
   return (
     <>
       <DialogContent>
-        {`It looks like you don’t have write access to `}
-        <strong>{repository.gitHubRepository.fullName}</strong>
-        {`. If you should have permission, please check with a repository administrator.`}
-        <br />
-        <br />
-        {` Do you want to create a fork of this repository at `}
-        <strong>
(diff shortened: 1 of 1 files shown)


app/src/ui/forks/create-fork-dialog.tsx
```diff
@@ -101,7 +101,10 @@ function renderCreateForkDialogContent(
       <DialogContent>
         {`It looks like you don’t have write access to `}
         <strong>{repository.gitHubRepository.fullName}</strong>
-        {`. Do you want to create a fork of this repository at `}
+        {`. If you should have permissions, please check with a repository administrator.`}
```
billygriffin:
```suggestion
        {`. If you should have permission, please check with a repository administrator.`}
```

tierninho:
> After pic

👍 Looks good.

Also, the words "...to continue" seem unnecessary and a bit awkward.

app/src/ui/forks/create-fork-dialog.tsx
```diff
@@ -99,13 +99,18 @@ function renderCreateForkDialogContent(
...
-        <strong>{repository.gitHubRepository.fullName}</strong>
-        {`. Do you want to create a fork of this repository at `}
-        <strong>
-          {`${account.login}/${repository.gitHubRepository.name}`}
-        </strong>
-        {` to continue?`}
+        <p>
+          {`It looks like you don’t have write access to `}
+          <strong>{repository.gitHubRepository.fullName}</strong>
+          {`. If you should have permission, please check with a repository administrator.`}
```
outofambit:
```suggestion
          {`. If you should, please check with a repository administrator.`}
```
@billygriffin maybe we could shorten this a little?

Issue #7827, opened 2019-06-18
Stash error with conflicted file upon new branch creation


tierninho:
1. checkout `master` at sha A
2. create `branch-a` at sha A
3. commit to `master` a change to `config/routes.rb` so master is now at sha B
4. checkout `branch-a` (which is at sha A)
5. make a conflicting change to `config/routes.rb`
6. hit the "current branch" button and create a new branch (branch-b), based off master, bring changes with me
7. after a moment branch-a will be checked out, the changes are gone
8. but `git stash show` will show the changes
9. and the new branch (branch-b) will have been created, if you switch to it manually, the UI Shows the stashed changes

Step 9:
![image](https://user-images.githubusercontent.com/14828183/59715264-31a96480-91af-11e9-91d5-c31ab6f80500.png)

```
2019-06-18T11:22:55.114Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: github - latentflip (has token)
2019-06-18T11:22:56.120Z - error: [ui] `git -c credential.helper= -c protocol.version=2 checkout --progress live-update-new-notifications --` exited with an unexpected code: 1.
stderr:
error: Your local changes to the following files would be overwritten by checkout:
	config/routes.rb
Please commit your changes or stash them before you switch branches.
Aborting

(The error was parsed as 37: Unable to switch branches as there are working directory changes which would be overwritten. Please commit or stash your changes.)
2019-06-18T11:23:00.092Z - info: [ui] Executing createStashEntry: git stash push --include-untracked -m !!GitHub_Desktop<live-update-new-notifications> (took 3.968s)
2019-06-18T11:23:01.252Z - warn: [ui] failed fetching updated PRs for repository github/github
Error: got max pull requests, aborting
    at Object.continue (/webpack:/app/src/lib/api.ts:677:23)
    at k.fetchAll (/webpack:/app/src/lib/api.ts:755:5)
2019-06-18T11:23:02.251Z - info: [ui] Executing getBranches: git for-each-ref --format=%(refname)%00%(refname:short)%00%(upstream:short)%00%(objectname)%00%(objectname:short)%00%(author)%00%(committer)%00%(parent)%00%(symref)%00%(subject)%00%(body)%00%(trailers:unfold,only)%00%1F refs/heads refs/remotes (took 5.190s)
[...]

PR #9038, opened 2020-02-03
Race condition in re-entrant checkoutBranch can cause stash and move operations to fail silently

niik:
## Description

I stumbled upon this today as I was creating a new branch and wanted to bring along some changes I had made. After typing in the new branch name and choosing to bring along changes I was left on the branch I started on. The new branch had been created and the stash had been created for it but Desktop hadn't switched to the new branch or popped the transient stash entry.

In digging through this I discovered that this is due to a race condition due to the re-entrant nature of `_checkoutBranch`. I've been able to consistently reproduce this recently and I was able to narrow it down to the recent Git/dugite upgrade (#8920). I can't see any reason why the dugite upgrade should affect this behavior so I'm forced to assume that thanks to the performance improvements in Git 2.23 (which I've been able to measure) this race condition is easier to trigger on my machine.

In order to reproduce we need to engineer a scenario where https://github.com/desktop/desktop/pull/7474 comes into play. I.e. the initial checkout fails because it can't cleanly checkout the branch without overwriting local changes. See the "Steps to reproduce" section below for details on how I chose to do that.

The flow goes something like this.

1. The branch is created by `_createBranch` which returns a **branch model** (this will be important later on) that's created by shelling out to Git (i.e. the Git store is not updated).
1. `_checkoutBranch` is called with `UncommittedChangesStrategyKind.MoveToNewBranch` and a **branch model**
1. `_checkoutBranch` attempts to  checkout the branch which fails since it would overwrite local changes
https://github.com/desktop/desktop/blob/df6772f01029bb659cc31815fa19643476a8e121/app/src/lib/stores/app-store.ts#L3076-L3096
1. The `performFailableOperation` from :point_up: catches the `DugiteError.LocalChangesOverwritten` and emits that error and attaches the **branch name** (important) as extra metadata
1. The `localChangesOverwrittenHandler` error handler catches this specific error and calls the `_moveChangesToBranchAndCheckout` method with the **branch name** that it wants to checkout
https://github.com/desktop/desktop/blob/df6772f01029bb659cc31815fa19643476a8e121/app/src/ui/dispatcher/error-handlers.ts#L475
1. We eventually make it back into the `_checkoutBranch` method where this 👇 hack takes the **branch name** and attempts to resolve it to a **branch model**
https://github.com/desktop/desktop/blob/df6772f01029bb659cc31815fa19643476a8e121/app/src/lib/stores/app-store.ts#L3042-L3049
1. Whether or not it's able to resolve that branch model depends entirely on whether the original `checkoutBranch` method has had time to refresh the repository state since the `getLocalBranch` method depends on the cached branch state in the repository state cache.

### The real solution

I was never a big fan of the re-entrant error-handler approach as can be seen in my [comment from the initial PR](https://github.com/desktop/desktop/pull/7474#discussion_r286883784) and I think the best solution would be to move this logic inline instead of relying on an error handler.

We filed https://github.com/desktop/desktop/issues/7617 to track implementing an inline logic approach and it turns out @probablycorey already did most of the work in https://github.com/desktop/desktop/pull/8076 but it was ultimately closed. I think we should resurrect that work and bring it over the finish line. That's a decently sized chunk of work though which is why I'm opening this intermediate PR as a stop-gap measure.

### The intermediate solution

This PR doesn't implement the nice solution that I'm proposing in the previous section but rather plugs this hole by passing along the same branch model that was returned from `_createBranch` all the way through the error handler and back and removes the ability to call `_checkoutBranch` with a branch name.

#### Steps to reproduce

1. Create a new repository
2. Create, and commit, a file `README.txt` with some random contents in it
3. Create a new branch `branch-a` make some changes to `README.txt`, and commit them
4. Switch back to `master` and make some conflicting changes to `README.txt`
5. Switch back to `branch-a`
6. Make some changes to `README.txt` but **don't** commit them.
7. From within desktop choose to create a new branch based on `master`, call it `branch-b`. When asked if you want to bring your changes with you say yes

#### Expected result

`branch-b` is checked out and one conflicted file (`README.txt`) exists in your working directory

#### Actual result (sometimes, since this is a race condition)

`branch-a` is still checked out and you have no changes in your working directory. If you switch to `branch-b` the stashed changes show up.

### DIFF
--- a/app/src/lib/git-error-context.ts
+++ b/app/src/lib/git-error-context.ts
@@ -1,3 +1,5 @@
+import { Branch } from '../models/branch'
+
 type MergeOrPullConflictsErrorContext = {
   /** The Git operation that triggered the conflicted state */
   readonly kind: 'merge' | 'pull'
@@ -13,7 +15,7 @@ type CheckoutBranchErrorContext = {
   readonly kind: 'checkout'
 
   /** The branch associated with the current tip of the repository, "ours" in Git terminology */
-  readonly branchToCheckout: string
+  readonly branchToCheckout: Branch
 }
 
 /** A custom shape of data for actions to provide to help with error handling */
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5246,22 +5246,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
       await this._fetchRemote(repository, remote, FetchType.UserInitiatedTask)
 
-      const gitStore = this.gitStoreCache.get(repository)
-
       const localBranchName = `pr/${pullRequest.pullRequestNumber}`
-      const doesBranchExist =
-        gitStore.allBranches.find(branch => branch.name === localBranchName) !=
-        null
+      const existingBranch = this.getLocalBranch(repository, localBranchName)
 
-      if (!doesBranchExist) {
+      if (existingBranch === null) {
         await this._createBranch(
           repository,
           localBranchName,
           `${remoteName}/${head.ref}`
         )
+      } else {
+        await this._checkoutBranch(repository, existingBranch)
       }
-
-      await this._checkoutBranch(repository, localBranchName)
     }
 
     this.statsStore.recordPRBranchCheckout()
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3032,21 +3032,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
   /** This shouldn't be called directly. See `Dispatcher`. */
   public async _checkoutBranch(
     repository: Repository,
-    branch: Branch | string,
+    branch: Branch,
     uncommittedChangesStrategy: UncommittedChangesStrategy = getUncommittedChangesStrategy(
       this.uncommittedChangesStrategyKind
     )
   ): Promise<Repository> {
     const gitStore = this.gitStoreCache.get(repository)
     const kind = 'checkout'
-    const foundBranch =
-      typeof branch === 'string'
-        ? this.getLocalBranch(repository, branch)
-        : branch
-
-    if (foundBranch == null) {
-      return repository
-    }
 
     const { changesState, branchesState } = this.repositoryStateCache.get(
       repository
@@ -3059,7 +3051,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         if (uncommittedChangesStrategy.kind === askToStash.kind) {
           this._showPopup({
             type: PopupType.StashAndSwitchBranch,
-            branchToCheckout: foundBranch,
+            branchToCheckout: branch,
             repository,
           })
           return repository
@@ -3067,7 +3059,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
         stashToPop = await this.stashToPopAfterBranchCheckout(
           repository,
-          foundBranch,
+          branch,
           uncommittedChangesStrategy
         )
       }
@@ -3077,7 +3069,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       (await this.withAuthenticatingUser(repository, (repository, account) =>
         gitStore.performFailableOperation(
           () =>
-            checkoutBranch(repository, account, foundBranch, progress => {
+            checkoutBranch(repository, account, branch, progress => {
               this.updateCheckoutProgress(repository, progress)
             }),
           {
@@ -3089,7 +3081,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
             },
             gitContext: {
               kind: 'checkout',
-              branchToCheckout: foundBranch,
+              branchToCheckout: branch,
             },
           }
         )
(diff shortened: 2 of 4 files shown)


tierninho:
I followed your testing notes and ended up at the expected result outlined above. Here is my result:

Screen Shot 2020-02-03 at 11 54 16 AM

Subsequently, I resolved the conflict in my editor, got a green checkmark, and attempted to commit the file but ran into a commit error:

Screen Shot 2020-02-03 at 11 54 38 AM

Logs are
```
install.ts:23 `git commit -F -` exited with an unexpected code: 1.
stdout:
On branch branch-b

(The error was parsed as 17: There are no changes to commit.)
```


This bug reminds me of https://github.com/desktop/desktop/issues/7827

app/src/lib/git-error-context.ts
```diff
@@ -13,7 +15,7 @@ type CheckoutBranchErrorContext = {
   readonly kind: 'checkout'
 
   /** The branch associated with the current tip of the repository, "ours" in Git terminology */
-  readonly branchToCheckout: string
+  readonly branchToCheckout: Branch
```
qkienly: Duplicate of #

app/src/lib/stores/app-store.ts
```diff
@@ -5246,22 +5249,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      const existingBranch = this.getLocalBranch(repository, localBranchName)
 
-      if (!doesBranchExist) {
+      if (existingBranch === null) {
         await this._createBranch(
           repository,
           localBranchName,
           `${remoteName}/${head.ref}`
         )
+      } else {
```
kuychaco: This `else` is new. Just wanted to double check that the behavior is intended... my understanding is that previously if the branch didn't exist we would create it and then check it out. Now, with this `else` we only create the branch and _do not_ check it out? @niik

kuychaco (commented):
@niik did a first pass and left a few comments with questions and suggestions. Much appreciation for the thorough and informative writeup! It was super helpful in reviewing this PR, and I found myself feeling delighted as I read it 💖

app/src/lib/stores/app-store.ts
```diff
@@ -3089,7 +3081,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
             },
             gitContext: {
               kind: 'checkout',
-              branchToCheckout: foundBranch.name,
+              branchToCheckout: branch,
```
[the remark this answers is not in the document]
niik: No, we want to avoid using the branch name here because whatever is set in this property is what ends up getting passed to the second `_checkoutBranch` invocation from the error handler. Since `_checkoutBranch` no longer supports string branch names the definition for this property no longer includes the string union type.

https://github.com/desktop/desktop/blob/30305ab02a1efef6404f940e3443a5805c10c41c/app/src/lib/git-error-context.ts#L18

app/src/lib/stores/app-store.ts
```diff
@@ -5219,7 +5211,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
         }
       }
-      await this._checkoutBranch(repository, head.ref)
+      const branch = this.getLocalBranch(repository, head.ref)
+
+      // N.B: This looks weird, and it is. _checkoutBranch used
+      // to behave this way (silently ignoring checkout) when given
+      // a string and not a Branch model. When rewriting _checkoutBranch
+      // to remove the support for string branch names the behavior
+      // was moved up to this method to not alter the current behavior.
```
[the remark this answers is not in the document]
niik: I think your suggestion is much better, thanks!

app/src/lib/stores/app-store.ts
niik: Oh yeah, I should have documented this, sorry about that. If you look in `_createBranch` we already do a checkout as part of creating the branch

https://github.com/desktop/desktop/blob/30305ab02a1efef6404f940e3443a5805c10c41c/app/src/lib/stores/app-store.ts#L3000-L3006

In other words, the previous `_checkoutBranch` call was redundant.

niik:
> I followed your testing notes and ended up at the expected result outlined above. Here is my result:

Right, so the fact that you ended up at the _expected results_ means that you didn't hit the race condition. That's not terribly surprising as it's the nature of such problems to be timing dependent. It's possible that on your machine it's going to be very hard to reproduce the problem whereas on my machine it became straightforward to reproduce with the Git 2.23 upgrade.

> This bug reminds me of #7827

I agree. We should reopen that. The issue is that you're resolving a (stashed) conflict in such a way that the contents of the file match that of the contents already present in `branch-b` and as such there's nothing to commit.

image

And the diff view shows:

image

niik:
> I agree. We should reopen that. The issue is that you're resolving a (stashed) conflict in such a way that the contents of the file match that of the contents already present in `branch-b` and as such there's nothing to commit.

Spoke too soon. That issue is slightly different but this is still an issue. If you end up resolving the stash conflict in such a way that no changes were made (i.e. `--ours`) then we shouldn't let you commit that file.

kuychaco (approved):
LGTM! :shipit:

tierninho (approved):
Makes sense. Thanks for the clarification!

laro67:
Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/ghei36>

nigar9595 (commented):
9109

DurgadeviAvatarae:
Hk

PR #9052, opened 2020-02-04
Revamp draft release task

outofambit:
## Overview

updates our draft release task to be a little more helpful. it now

* no longer aborts at the start if there are uncommitted changes in your local repo
* updates the desktop version number for you in `app/package.json`
* places draft release notes in `changelog.json` for you (unless the version already has an entry, in which case if prints it out on the console like before)
* has updated instructional text to explain all of this!

### DIFF
--- a/script/draft-release/run.ts
+++ b/script/draft-release/run.ts
@@ -62,24 +62,27 @@ function parseChannel(arg: string): Channel {
  * @param entries release notes for the next release
  */
 function printInstructions(nextVersion: string, entries: Array<string>) {
-  const object: any = {}
-  object[nextVersion] = entries.sort()
-
   const baseSteps = [
     'Revise the release notes according to https://github.com/desktop/desktop/blob/development/docs/process/writing-release-notes.md',
+    'Lint them with: yarn draft-release:format',
     'Commit these changes (on a "release" branch) and push them to GitHub',
     'Read this to perform the release: https://github.com/desktop/desktop/blob/development/docs/process/releasing-updates.md',
   ]
   if (entries.length === 0) {
     printSteps(baseSteps)
   } else {
-    printSteps([
+    const object: any = {}
+    object[nextVersion] = entries.sort()
+    const steps = [
       `Concatenate this to the beginning of the 'releases' element in the changelog.json as a starting point:\n${format(
         JSON.stringify(object),
-        { parser: 'json' }
+        {
+          parser: 'json',
+        }
       )}\n`,
       ...baseSteps,
-    ])
+    ]
+    printSteps(steps)
   }
 }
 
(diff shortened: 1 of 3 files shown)


script/draft-release/run.ts
```diff
@@ -96,21 +111,81 @@ export async function run(args: ReadonlyArray<string>): Promise<void> {
...
-
-    if (channel === 'production') {
-      // make sure we only include entries since the latest production release
-      const existingChangelog = getChangelogEntriesSince(previousVersion)
-      const entries = [...existingChangelog]
-      printInstructions(nextVersion, entries)
-    } else if (channel === 'beta') {
-      const entries = [...changelogEntries]
-      printInstructions(nextVersion, entries)
+    return
```
kuychaco: Question: wondering if we want to wait to return out of the function until after we update the app version?

script/draft-release/run.ts
```diff
@@ -96,21 +111,81 @@ export async function run(args: ReadonlyArray<string>): Promise<void> {
...
+    Please manually set it to ${nextVersion} in app/package.json.`)
+  }
+
+  const currentChangelog: IChangelog = require(changelogPath)
+  // if it's a new production release, make sure we only include
+  // entries since the latest production release
+  const newEntries =
+    channel === 'production'
+      ? [...getChangelogEntriesSince(previousVersion)]
+      : [...(await convertToChangelogFormat(lines))]
```
kuychaco: Wondering if we want to explicitly check that the channel is `beta` here (like we did before these changes). Right now we're doing this anytime the channel is not `production` (so if it's undefined, or some other string, etc)

script/draft-release/run.ts
```diff
@@ -57,30 +62,40 @@ function parseChannel(arg: string): Channel {
...
 }
 
 export async function run(args: ReadonlyArray<string>): Promise<void> {
-  try {
-    await spawn('git', ['diff-index', '--quiet', 'HEAD'])
-  } catch {
-    throw new Error(
-      `There are uncommitted changes in the working directory. Aborting...`
-    )
-  }
```
kuychaco: > no longer aborts at the start if there are uncommitted changes in your local repo

@outofambit I'm curious about your thoughts on dropping this. I'm generally cool with doing so, and wondering if you have any specific scenarios in mind that prompted this. I can't quite think of any reason why there would be uncommitted changes on a release branch. 

The risk of dropping this does seem minimal -- overwriting changes to the package.json or changelog, or accidentally committing unrelated working directory changes. In both of these cases I expect issues to be caught in PR review. So I don't see a problem with this change, just want to get a sense of what your thoughts are on it.

outofambit: @kuychaco yes good question! the context here is that this check is pretty flakey and gives me (and others from what i hear) false rejections pretty often. and yep to your point it is quite uncommon to have any changes on a release branch. (now that i'm thinking about it, this check was added when we were less rigorous about using release branches, which might explain its initial utility better)

script/draft-release/run.ts
outofambit: good point. fixed in 6823bc683.

script/draft-release/run.ts
outofambit: mmm good point! i restructured things for this to work better in a39c3cfe101279d75595ac180d868a12db64605b

PR #9058, opened 2020-02-05
Update roadmap with latest ships

billygriffin:
This isn't great because it doesn't add the things we're working on and those that are upcoming except for proxies, but this at least moves us in the right direction. 😅


outofambit:
this is only a docs change so i'm going to skip waiting for CI

PR #9063, opened 2020-02-06
2.3.2-beta2 release

outofambit:
- [x] finalize release notes
- [x] release!

Fixes from upgrade to Electron 7:
Closes #8846 
Closes #9078

Issue #9048, opened 2020-02-04
[Fork] Clean up 2.3.1 hotfix


outofambit:
## Current state

In https://github.com/desktop/desktop/commit/aaf78b3236d4d1a74a920b3ae70bfea8b678fb26, we implemented a fix that erred on the side of duplicating some logic for the sake of safety.

https://github.com/desktop/desktop/commit/aaf78b3236d4d1a74a920b3ae70bfea8b678fb26#diff-d6b76b602a5c48740d6d10deda6f71ebR5566-R5578

Checks the current remotes before selecting the appropriate path: creating a new `upstream` remote, or updating the existing one. This is because git errors if you try to add a remote that already exists in the repository.

## Proposed Refactor

Add ability to detect the "remote already exists" error so that we can skip loading remotes and just attempt adding the `upstream` remote. if it fails with the "remote already exists" error, we just update the existing remote and move on.

This will reduce our number of git calls by one (1) in situations where the `upstream` remote already exists.

PR #9071, opened 2020-02-07
Update dugite to fix macOS notarization

outofambit:
- [x] release new dugite-native with desktop/dugite-native#319 fix
- [x] release new dugite with this fix as v1.88.2
- [x] update this branch to use dugite v1.88.2
- [ ] merge
- [ ] revisit #9063

### DIFF
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -367,6 +367,8 @@ function getDescriptionForError(error: DugiteError): string | null {
       // be used as the error message (or stdout if stderr is empty), i.e. the
       // same behavior as before the ConfigLockFileAlreadyExists was added
       return null
+    case DugiteError.RemoteAlreadyExists:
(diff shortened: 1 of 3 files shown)


app/src/lib/git/core.ts
```diff
@@ -367,6 +367,8 @@ function getDescriptionForError(error: DugiteError): string | null {
       // be used as the error message (or stdout if stderr is empty), i.e. the
       // same behavior as before the ConfigLockFileAlreadyExists was added
       return null
+    case DugiteError.RemoteAlreadyExists:
+      return null
```
outofambit: this is essentially a pass-through here ie doesn't change any behavior

kuychaco: Ah gotcha, this is in preparation for https://github.com/desktop/desktop/issues/9048

PR #9085, opened 2020-02-10
Scroll diffs with keyboard without moving the invisible cursor

### LINKED ISSUE #9074: Suggestion: Support Partial Scrolling through Arrow Keys

Support Scrolling Changes of any file with Arrow Keys.

### Proposed solution

How will it benefit Desktop and its users? It can help people like with broken scroll wheel to scroll a little, pressing page down works but that's too much to scroll. Perhaps, giving to scroll line-by-line would make more sense while reviewing changes before committing them? 

### Additional context

Add any other context like screenshots or mock ups are helpful, if applicable.
Modify any file, select any modified file, press up/down arrow keys to see 'em not working at all.

Thanks in advance ;)

niik:
Closes #9074

## Description

This came up in #9074. Turns out the <kbd>Up</kbd> and <kbd>Down</kbd> keys (as well as the <kbd>PageUp</kbd> and <kbd>PageDown</kbd>) move the cursor inside of a CodeMirror document even when the document is readonly and the cursor is hidden.

This affects accessibility since the scrolling behavior becomes confusing without a visible cursor to observe.

This PR alters the behavior to line-by-line (or rather 1 "normal" line height given that we wrap lines) scrolling.

Priority 3 since there's a workaround. Tentatively put up for inclusion in 2.4


tierninho:
I attempted to reproduce the new behavior, but I did not notice any changes. Should I be able to scroll the Changes > diff > line by line here? I was unable to do so with the up/down keys on a Mac.

Screen Shot 2020-02-10 at 10 32 18 AM

niik:
@tierninho that screenshot is a bit confusing because it doesn't look like there's a scroll bar in it from what I can tell. In general though you should be able to test this out by building this branch, finding, or creating, a large file or lots of changes so that a scrollbar appears to the right hand side of the diff

image

Once you see that scrollbar and are able to scroll using the trackpad/mouse then you should be able to *click* on the diff text  and then use the up/down key to move the scroll bar.

Issue #9054, opened 2020-02-05
After a pull, GitHub still says, there are changes to pull


sdiedrichsen:
### Describe the bug

After a pull, the window top bar refreshes but instead of "Fetch origin", it says "Pull" with the same number fo changes. 
Clicking on the tabs "Changes" and "History" resolves the issue eventually. 

 

### Version & OS

2.3.1 on macOS 10.15.3.

### Expected behavior

After  a pull. there should be a feedback, that the pull was successful and the right most button in the top bar should display "Fetch origin"

### Actual behavior

A clear and concise description of what actually happened.

### Screenshots

Add screenshots to help explain your problem, if applicable.

### Logs

Attach your logs by opening the `Help` menu and selecting `Show Logs...`, if applicable.

steveward:
Thanks for the report @sdiedrichsen. Could you upload a log file from GitHub Desktop so that I could get some more information about this issue? To access the log files go to the file menu in GitHub Desktop and select `Help` > `Show Logs in Finder (macOS) or Explorer (Windows)`.

The log files are created daily -- please upload a log file as an attachment from a day where you experienced the issue. Thanks!

sdiedrichsen:
Here's the log:
>>>>>

2020-02-06T07:08:23.288Z - info: [ui] [AppStore] loading 1 repositories from store
2020-02-06T07:08:23.290Z - info: [ui] [AppStore] found account: sdiedric (Steffan Diedrichsen)
2020-02-06T07:08:23.290Z - info: [ui] [AppStore] found account: sdiedrichsen (sdiedrichsen)
2020-02-06T07:08:23.394Z - info: [ui] [BranchPruner] Last prune took place in 2 hours - skipping
2020-02-06T07:08:23.936Z - info: [ui] launching: 2.3.1 (Mac OS 10.15.3)
2020-02-06T07:08:23.937Z - info: [ui] execPath: '/Applications/GitHub Desktop.app/Contents/Frameworks/GitHub Desktop Helper.app/Contents/MacOS/GitHub Desktop Helper'
2020-02-06T07:08:25.604Z - info: [ui] Executing getStatus: git --no-optional-locks status --untracked-files=all --branch --porcelain=2 -z (took 1.670s)
2020-02-06T07:08:25.605Z - info: [ui] Executing getStatus: git --no-optional-locks status --untracked-files=all --branch --porcelain=2 -z (took 2.252s)
[...]

Issue #9053, opened 2020-02-05
After pull, I get an alert "Cannot read property 'length' of undefined"


sdiedrichsen:
### Version & OS

Git Hub Desktop 2.3.1 with macOS 10.15,3

### Steps to reproduce the behavior

1. Trigger a Pull

### Expected behavior

A feedback, that the pull with rebase did succeed. That was the case in previous versions.

### Actual behavior

Now, I get an alert.

### Screenshots

![GitHubErrorAlert](https://user-images.githubusercontent.com/47812706/73823252-21a2cb80-47f8-11ea-9d6b-de549aa679b7.png)

### Logs

Attach your logs by opening the `Help` menu and selecting `Show Logs...`, if applicable.

steveward:
Thanks for the report @sdiedrichsen. Could you upload the log file from GitHub Desktop so that I could get some more information about this error? To access the log files go to the file menu in GitHub Desktop and select `Help` > `Show Logs in Finder (macOS) or Explorer (Windows)`.

The log files are created daily -- please upload a log file as an attachment from a day where you experienced the issue. Thanks!

Thanks,
Steve

sdiedrichsen:
Here's the log:

>>>>>>

020-02-06T07:12:29.397Z - info: [ui] Current tutorial step is NotApplicable
2020-02-06T07:12:41.714Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: MusicApps - sdiedric (has token)
2020-02-06T07:12:45.746Z - info: [ui] Executing checkoutBranch: git -c credential.helper= checkout --progress lhierl/TronTempoCompile -- (took 4.026s)
2020-02-06T07:12:47.006Z - info: [ui] Executing getStatus: git --no-optional-locks status --untracked-files=all --branch --porcelain=2 -z (took 1.259s)
2020-02-06T07:12:47.137Z - info: [ui] Current tutorial step is NotApplicable
2020-02-06T07:12:47.145Z - info: [ui] [Timing] Action 'checkout branch from list' for 'musicapps/MusicApps' took 5.413s
2020-02-06T07:12:47.587Z - warn: [ui] deserialize: invalid JSON found at 'https://github.aso.apple.com/api/v3/repos/musicapps/MusicApps/branches/lhierl/TronTempoCompile/push_control' - status: 500, length: '(missing)' id: '(missing)'
SyntaxError: Unexpected end of JSON input
    at o (/webpack:/app/src/lib/http.ts:68:5)
[...]

PR #9101, opened 2020-02-13
Guard against unexpected push control API response

niik:
## Description

See https://github.com/desktop/desktop/issues/9054#issuecomment-582768322 and https://github.com/desktop/desktop/issues/9053.

```
2020-02-06T07:08:44.310Z - error: [ui] Error performing periodic fetch for 'musicapps/MusicApps'
TypeError: Cannot read property 'length' of undefined
    at Object.t.isBranchPushable (/webpack:/app/src/lib/helpers/push-control.ts:37:28)
    at Ve.refreshBranchProtectionState (/webpack:/app/src/lib/stores/app-store.ts:887:9)
    at /webpack:/app/src/lib/stores/app-store.ts:4012:9
    at Ve.withPushPullFetch (/webpack:/app/src/lib/stores/app-store.ts:3609:7)
    at Ve.performFetch (/webpack:/app/src/lib/stores/app-store.ts:3974:5)
    at t.BackgroundFetcher.performAndScheduleFetch (/webpack:/app/src/lib/stores/helpers/background-fetcher.ts:97:9)
```

This appears to be coming from the `.length` property access in https://github.com/desktop/desktop/blob/ede708400a30be98b64cc7b844dfda9c137c7cd1/app/src/lib/helpers/push-control.ts#L37 and means that _at least_ the `required_status_checks` property is undefined (I suspect that's not the only property though).

While I absolutely think we should get to the bottom of what's actually going on (I have some theories but haven't been able to conclusively prove any of them yet) this is a stop-gap measure that should avoid the problem by being much more defensive when interpreting the data returned from the API.

Priority 2 since it's affecting users in a meaningful way occurs during critical flows (checkout, push, pull, fetch).

### DIFF
--- a/app/src/lib/helpers/push-control.ts
+++ b/app/src/lib/helpers/push-control.ts
@@ -43,5 +43,8 @@ export function isBranchPushable(pushControl: IAPIPushControl) {
   const noMergeRequirements =
     requiredStatusCheckCount === 0 && required_approving_review_count === 0
 
+  // We check for !== false so that if a future version of the API decides to
+  // remove or rename that property we'll revert to assuming that the user
+  // _does_ have access rather than assuming that they _don't_.
   return allow_actor !== false && noMergeRequirements
 }
--- a/app/src/lib/helpers/push-control.ts
+++ b/app/src/lib/helpers/push-control.ts
@@ -27,6 +27,13 @@ export function isBranchPushable(pushControl: IAPIPushControl) {
     required_approving_review_count,
   } = pushControl
 
+  // See https://github.com/desktop/desktop/issues/9054#issuecomment-582768322
+  // We'll guard against this being undefined until we can determine the
+  // root cause and fix that.
+  const requiredStatusCheckCount = Array.isArray(required_status_checks)
+    ? required_status_checks.length
+    : 0
+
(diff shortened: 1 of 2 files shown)


niik:
Alright, I've found one way to reproduce this error but it doesn't work if you've already got https://github.com/desktop/desktop/pull/9086.

1. Create a branch called `issue` and publish that
2. Create a branch called `issue#123` and publish that
3. :boom:

image

What happens is that because there's a `#` at the end of the branch GitHub Desktop will request `branches/issue` instead of `branches/issue#123/push_control` and since there's already a branch named `issue` the API will return that instead of a push_control hash.

Gonna go back and see if this could be the issue for some of the reports we've been getting.

tierninho:
Successfully created both `issue` and `issue#123` with this build, and the latter returned `OK 200` . However, I am unsure if #9086 was involved.

Screen Shot 2020-02-13 at 2 27 16 PM

tested in OSX, 0d868bcfcc

niik:
> However, I am unsure if #9086 was involved.

Yeah it would have been if you built from this branch. If you want to take a stab at reproducing it I'd suggest using the latest beta which doesn't have #9086 in it yet.

PR #9109, opened 2020-02-15
Checkout correct PR branch when “open this in GitHub Desktop” link is clicked on dotcom

### LINKED ISSUE #9108: 🐛 Clicking “open this in GitHub Desktop” in a PR on dotcom doesn't check out branch
### Describe the bug

Clicking “open this in GitHub Desktop” in a PR on dotcom no longer checks out branch associated with the PR

### Version & OS

As of 2.3.2-beta2 this feature has regressed (see below for context). On both Mac and Windows

1. Go to a PR on github.com and click “open this in GitHub Desktop”

### Expected behavior

The correct repo is checked out AND the branch associated with the PR is checked out

### Actual behavior

The correct repo is checked out. The PR branch is NOT checked out.

### Additional context

The branch checkout logic was refactored in #9038 such that we never get to the checkout logic because `localBranch` comes back as `undefined` (https://github.com/desktop/desktop/pull/9038/files#diff-795e95169b44990beac65ec8f1ea4d06R1559)

kuychaco:
Closes #9108 

## Description

The branch checkout logic was refactored in #9038 such that we never get to the checkout logic because `localBranch` comes back as `undefined` (https://github.com/desktop/desktop/pull/9038/files#diff-795e95169b44990beac65ec8f1ea4d06R1559).

This PR fixes the logic so that the appropriate branch is checked out.

## Release notes

Notes: [Fixed] Checkout correct PR branch when “open this in GitHub Desktop” link is clicked on dotcom


app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -1556,7 +1556,10 @@ export class Dispatcher {
         shouldCheckoutBranch = tip.branch.nameWithoutRemote !== branch
       }
 
-      const localBranch = branches.find(b => b.upstreamWithoutRemote === branch)
+      const localBranch = branches.find(
+        b =>
+          b.upstreamWithoutRemote === branch || b.nameWithoutRemote === branch
+      )
```
kuychaco: In some cases `upstream` is null (I did some investigation and am still not 100% clear on when this is the case). For example, in desktop the Branch object associated with https://github.com/desktop/dugite/pull/310 is `{name: "pr/310", upstream: null, tip: Commit, type: 0}` so there is no match for `upstreamWithoutRemote`. 

I've added a check to see if the branch name passed in matches with `nameWithoutRemote` and this seems to fix the issue. I'm now wondering in which cases we can expect to match against `upstreamWithoutRemote`, and relatedly, if we can rely solely on the `nameWithoutRemote` check.

/ cc @outofambit @niik

niik: Nice troubleshooting @kuychaco! It seems this was a lazy copy paste error on my part. The intention of https://github.com/desktop/desktop/pull/9038/commits/1b44dcbd615634b3d2c4d0abf954d4c6603c1162 was to exactly replicate the [functionality removed from `AppStore._checkoutBranch`](https://github.com/desktop/desktop/pull/9038/commits/0d1134d1289d1bb9bd5c8cafef47d72b855812fe#diff-d6b76b602a5c48740d6d10deda6f71ebL3042-L3049) to resolve the branch. The implementation of `AppStore.getLocalBranch` was simple enough: https://github.com/desktop/desktop/blob/222a5de9a8326a459f4dcb7487ddbd2212866255/app/src/lib/stores/app-store.ts#L3025-L3027

But as I was updating the `handleCloneInDesktopOptions` method I spotted the `.find` earlier in the method and figured it did the same thing as `getLocalBranch`:
https://github.com/desktop/desktop/blob/222a5de9a8326a459f4dcb7487ddbd2212866255/app/src/ui/dispatcher/dispatcher.ts#L1542-L1543

I didn't notice the subtle difference between the two, and used the same approach here.

All that said I therefore believe we should attempt to implement the original intention of the change from #9038 and _only_ find based on the `nameWithoutRemote` property in order to avoid any other possible subtle behavioral differences from using both `upstreamWithoutRemote` and `nameWithoutRemote `

```ts
const localBranch = branches.find(b => b.nameWithoutRemote === branch)
```

kuychaco: Fixed in 8d94da955!

PR #9126, opened 2020-02-17
Proxy (p)refactor of create tutorial repository dialog

niik:
## Description

As part of the proxy work I'm doing I've tentatively settled on an approach which means that all network Git operations are going to have to originate from within the App Store. The only network Git operation that doesn't originate from the AppStore today is the `push` operation within the `CreateTutorialRepositoryDialog` component.

In refactoring this I realized that the change was becoming rather large and thus a good candidate to be extracted and reviewed on its own.

Also, [@outofambit totes foresaw this happening](https://github.com/desktop/desktop/pull/8244#discussion_r323884417)

### Notes for the reviewer

This is in essence a refactor and no functionality should have changed. Evidence of altered logic are probably a mistake worth calling out in the review.

One notable exceptions is that the `CreateTutorialRepository` component used to have a separate `loading` state which was set to true when the create operation started and reset to false once it ended. This state determined whether the dialog would be enabled (i.e. whether the create/cancel buttons could be clicked), whether it was dismissable, and whether it should show the progress spinner or not. This state has been replaced by a variable derived from the presence of a progress property. Functionally to the user there should be no difference.
 
As a side effect of :point_up: I've slightly reordered the `createTutorialRepository` method to immediately set a progress upon commencing the creation.

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -1227,8 +1227,13 @@ export class StatsStore implements IStatsStore {
     }))
   }
 
-  /**
+  /*
    * Onboarding tutorial metrics
+   */
+
+  /**
+   * Onboarding tutorial has been started, the user has
+   * clicked the button to start the onboarding tutorial.
    */
   public recordTutorialStarted() {
     return this.updateDailyMeasures(() => ({
@@ -1236,6 +1241,9 @@ export class StatsStore implements IStatsStore {
     }))
   }
 
+  /**
+   * Onboarding tutorial has been successfully created
+   */
   public recordTutorialRepoCreated() {
     return this.updateDailyMeasures(() => ({
       tutorialRepoCreated: true,
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5587,8 +5587,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
       const apiRepository = await createTutorialRepository(
         account,
-        path,
         name,
+        path,
         this.setCreateTutorialProgress
       )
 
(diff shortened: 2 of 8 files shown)


app/src/lib/stats/stats-store.ts
```diff
@@ -1227,15 +1227,23 @@ export class StatsStore implements IStatsStore {
     }))
   }
 
-  /**
+  /*
```
outofambit: good catch

app/src/lib/stores/helpers/create-tutorial-repository.ts
```diff
@@ -0,0 +1,141 @@
...
+import { APIError } from '../../http'
+import {
+  executionOptionsWithProgress,
+  PushProgressParser,
+} from '../../progress'
+import { envForAuthentication } from '../../git/authentication'
+import { git } from '../../git'
+import { friendlyEndpointName } from '../../friendly-endpoint-name'
+
+const nl = __WIN32__ ? '\r\n' : '\n'
```
outofambit: [:blood-sob:](https://stackoverflow.com/questions/49968130/line-endings-also-known-as-newlines-in-js-strings)

app/src/lib/stores/app-store.ts
```diff
@@ -5568,6 +5577,70 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    description?: string
+  ) => {
+    this.setCreateTutorialRepositoryProps({
+      progress: { kind: 'generic', title, value, description },
+    })
+  }
+
+  private setCreateTutorialRepositoryProps(
+    state: Partial<ICreateTutorialRepositoryPopupProps>
+  ) {
```
outofambit: it looks like we're only using this to modify the `progress` property on the popup state. i believe with this typing though, you could use this method to modify the `account` property as well, which seems ill-advised. perhaps we could move `account` out of `ICreateTutorialRepositoryPopupProps` or narrow this method's typing to only modify `progress`?

(i like this as a general pattern for updating popups, just feels a little dangerous here)

niik: Good catch, I initially intended to port over the `loading` state to the popup props as well but I ended up deriving that state from the progress instead. I'll simplify this down to just progress.

PR #9139, opened 2020-02-19
Determine remote endpoint for Git network operations

niik:
*This is another continuation of the proxy support started in #9126 and #9127.*

Now that we're going to start looking up the proxy to use for Git network operations we're going to have to have an understanding of what the remote endpoint is gonna be for any given network operation.

Corporate proxies are often set up so that a script determines which proxy to use based on the url that the client wants to access (for https urls the script usually only gets the protocol and domain). See #9127 for details on that. 

Unfortunately for us there's not a 1:1 correlation between git command and url. Take the simplest case of `git clone URL` for example. While the initial request will be to `URL` it's possible that the repository could contain submodules pointing to other hosts. It's also possible that the repository is set up to use LFS in which case there may be subsequent requests to a dedicated LFS server.

While there might be multiple endpoints accessed by a single Git call we only have the ability to provide Git one proxy url per protocol (http/https) so we're gonna have do do a best-effort guess at a reasonable host. We're also gonna assume that the vast majority of repositories these days do all of their communications over http**s**.

It's also worth noting that the model we've used to provide Git with authentication details (username/password) has been based on the premise that the same credentials will work for any submodule and/or LFS instance accessed during the operation.

I've introduced a substitute method for `envForAuthentication` called `envForRemoteOperation`. This will now be the centralized location for determining the proxy url. The actual process of setting the required environment variables will come in a follow up pr.

## Notes for the reviewer

This PR touches sensitive parts of the codebase such as fetch, push, and pull. Please keep a lookout for anything that looks off or alters the behavior of either of these operations in a way not described below.

### Affected operations

**deleteBranch**

`deleteBranch` is also responsible for deleting the branch on the remote which means that it's essentially a `push` operation. Here I've had to introduce a new git operation to load the url of the branch remote. The risk of not being able to resolve the remote url is slim here but there's a fallback anyway

**Expected altered behavior**: One extra Git exec to lookup the remote url of the remote tracking branch

**checkout and revert**

While not a network operation in Git's eyes it's possible that the repository is set up to use LFS which might have to communicate with its server on checkout or revert. While we can't tell for certain what the LFS server url is (without shelling out to git lfs) it's a pretty safe bet that the account endpoint is a decent substitute given that that's what Git LFS will use to authenticate.

**Expected altered behavior**: None

**clone**

We use whatever url is provided as the remote url

**Expected altered behavior**: None

**fetch and fetchRefSpec**

We already had access to `IRemote` objects a few levels up in the call chain to the changes in here are just forwarding that information down rather than only using the `name` property of the `IRemote`

**Expected altered behavior**: None

**push**

This one got a bit complicated because I really didn't want to risk introducing any new behavior here. See the code for a long comment explaining the tradeoff.

**Expected altered behavior**: None

**Tutorial repository creation**

Simply a refactor to pass the known remote url to `envForRemoteOperation`

**Expected altered behavior**: None

### DIFF
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -3,10 +3,14 @@ import { getBranches } from './for-each-ref'
 import { Repository } from '../../models/repository'
 import { Branch, BranchType } from '../../models/branch'
 import { IGitAccount } from '../../models/git-account'
-import { envForAuthentication } from './authentication'
 import { formatAsLocalRef } from './refs'
 import { deleteRef } from './update-ref'
 import { GitError as DugiteError } from 'dugite'
+import { getRemoteURL } from './remote'
+import {
+  envForRemoteOperation,
+  getFallbackUrlForProxyResolve,
+} from './environment'
 
 /**
  * Create a new branch from the given start point.
@@ -73,22 +77,30 @@ export async function deleteBranch(
     await deleteLocalBranch(repository, branch.name)
   }
 
-  const remote = branch.remote
+  const remoteName = branch.remote
 
-  if (includeRemote && remote) {
+  if (includeRemote && remoteName) {
     const networkArguments = await gitNetworkArguments(repository, account)
+    const remoteUrl =
+      (await getRemoteURL(repository, remoteName).catch(err => {
+        // If we can't get the URL then it's very unlikely Git will be able to
+        // either and the push will fail. The URL is only used to resolve the
+        // proxy though so it's not critical.
+        log.error(`Could not resolve remote url for remote ${remoteName}`, err)
+        return null
+      })) || getFallbackUrlForProxyResolve(account, repository)
 
     const args = [
       ...networkArguments,
       'push',
-      remote,
+      remoteName,
       `:${branch.nameWithoutRemote}`,
     ]
 
     // If the user is not authenticated, the push is going to fail
     // Let this propagate and leave it to the caller to handle
     const result = await git(args, repository.path, 'deleteRemoteBranch', {
-      env: envForAuthentication(account),
+      env: envForRemoteOperation(account, remoteUrl),
       expectedErrors: new Set<DugiteError>([DugiteError.BranchDeletionFailed]),
     })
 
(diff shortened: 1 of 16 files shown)


tierninho:
> This PR touches sensitive parts of the codebase such as fetch, push, and pull. Please keep a lookout for anything that looks off or alters the behavior of either of these operations in a way not described below.

This may be premature, but I spent some time combing over the areas outlined above to search for potential inconsistencies, however nothing was found. Fetch, push, and pull, checkout, revert, clone, and the tutorial were tested, however LFS and submodules were not at this point. I will continue testing and report back if any out of the ordinary is noticed.

fred652:
Better

app/src/lib/git/environment.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @param account   The authentication information (if available) to provide
+ *                  to Git for use when connectingt to the remote
+ * @param remoteUrl The primary remote URL for this operation. Note that Git
+ *                  might connect to other remotes in order to fulfill the
+ *                  operation. As an example, a clone of
+ *                  https://github.com/desktop/desktop could containt a submodule
+ *                  pointing to another host entirely. Used to resolve which
+ *                  proxy (if any) should be used for the operation.
+ */
+export async function envForRemoteOperation(
```
outofambit: i'm assuming this `async` is anticipation of future work in this method since `envForAuthentication` itself isn't async.

app/src/lib/git/branch.ts
```diff
@@ -73,22 +77,30 @@ export async function deleteBranch(
...
+  if (includeRemote && remoteName) {
     const networkArguments = await gitNetworkArguments(repository, account)
+    const remoteUrl =
+      (await getRemoteURL(repository, remoteName).catch(err => {
+        // If we can't get the URL then it's very unlikely Git will be able to
+        // either and the push will fail. The URL is only used to resolve the
+        // proxy though so it's not critical.
+        log.error(`Could not resolve remote url for remote ${remoteName}`, err)
+        return null
+      })) || getFallbackUrlForProxyResolve(account, repository)
```
outofambit: just a thought but this _feels_ like a pattern we'll reuse elsewhere over time and might be a nice helper function to pull out sooner than later as we're going through proxy things.

did a quick search and am honestly surprised we don't use `git remote get-url` anywhere else in the codebase currently.

app/src/lib/git/environment.ts
niik: Yup! It becomes async in #9154 but making it async here meant a whole lot less changes to review in #9154

app/src/lib/git/branch.ts
niik: Yeah, I was on the fence about whether to extract it early but when I couldn't come up with a name for that method in under 10 seconds I bailed 😛

Issue #8989, opened 2020-01-27
Easy path to create an issue from GitHub Desktop


billygriffin:
### Describe the feature or problem you’d like to solve

In https://github.com/desktop/desktop/issues/2376, @TheGrimSilence proposed an idea to be able to create issues in GitHub Desktop. Generally we've felt like an integrated approach goes too far in the direction of in-app project management, but we do frequently hear that in the context of doing your work, if you see something that warrants creating a new issue, there's not a great path for that currently without causing considerable context switching and related cognitive overhead.

### Proposed solution

Our proposal here is to create a menu item and associated keyboard shortcut to direct people to the correct issue creation path on GitHub.com. This is dependent on a single path that accounts for repositories that use issue templates and those that do not, which doesn't currently exist but hopefully will in the near future.

This menu item should live in the `Repository` menu.

### Considerations

1. This is dependent on work on GitHub.com that provides a common path that satisfies the use case of repositories that use issue templates *and* those that do not. Currently the path would need to be /issues/new/choose for repositories using templates, and /issues/new for those not using templates.
2. We'll have to determine how to handle the cases where repositories do not have issues enabled at all (which is true for newly created forks by default - thanks @Rexogamer).

MakePixelsWork:
This still means you are redirected to another environment and have to switch back and forth. Having a full in-app (desktop) solution seems better to me... so you can keep it all in one central place and let the contents sync with the online Github environment and vice versa. One of the pillars of Git, right?

I'm not against a link that facilitates this from within the Desktop version, but a full integration seems better.

billygriffin:
Thanks @MakePixelsWork - we've talked about that several times, and though we agree that it would likely be useful for some folks, we don't think project management is core enough to the app's core functionality to warrant the prominence of an integrated experience. At least that's our working assumption to date, though we're not opposed to revisiting after this initial implementation.

Rexogamer:
another thing to note: forks by default don’t have issues enabled, so we’ll have to work around that

billygriffin:
> another thing to note: forks by default don’t have issues enabled, so we’ll have to work around that

This is a great point, thanks @Rexogamer! I was hoping we wouldn't need any information from the API for this and we could just rely on a common target URL, but it sounds like that edge case might present some difficulties. I added a "Considerations" section to the issue body so we sort those things out in implementation.

Kruithne:
[...]

PR #9141, opened 2020-02-20
use MenuIDs type in app menu building

outofambit:
While working on #8989, I noticed that we don't enforce the use of our MenuID type (which is a list of the menu ids we allow and use in the app) in the construction of our menu items. We use our MenuID type in many other places to ensure we use IDs that are consistent in different places in the app (event handling, etc).

This PR adds this type to our menu construction for (even) greater consistency.

### DIFF
--- a/app/src/models/app-menu.ts
+++ b/app/src/models/app-menu.ts
@@ -1,4 +1,5 @@
 import { assertNever } from '../lib/fatal-error'
+import { MenuIDs } from './menu-ids'
 
 /** A type union of all possible types of menu items */
 export type MenuItem =
@@ -16,7 +17,7 @@ export type ExecutableMenuItem = IMenuItem | ICheckboxMenuItem | IRadioMenuItem
  * Only useful for declaring the types, not for consumption
  */
 interface IBaseMenuItem {
-  readonly id: string
+  readonly id: MenuIDs
   readonly enabled: boolean
   readonly visible: boolean
   readonly label: string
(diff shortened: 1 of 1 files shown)


app/src/models/app-menu.ts
```diff
@@ -189,7 +190,7 @@ function menuItemFromElectronMenuItem(menuItem: Electron.MenuItem): MenuItem {
   // Our menu items always have ids and Electron.MenuItem takes on whatever
   // properties was defined on the MenuItemOptions template used to create it
   // but doesn't surface those in the type declaration.
-  const id: string | undefined = (menuItem as any).id
+  const id: MenuIDs | undefined = (menuItem as any).id
```
niik: I don't believe all `menuItem.id` will actually be MenuIDs here. We use `MenuIDs` for all the menu items that we're going to be manipulating from the renderer (setting enabled/disabled, changing text, etc).

For all the other menu items we [generate an automatic unique identified](https://github.com/desktop/desktop/blob/452c89d0de300878e2320c5da7bdeb42e949d765/app/src/main-process/menu/ensure-item-ids.ts#L16) based on its position, label, or role.

So while the `any` cast here is valid for strings, it's not valid for `MenuIDs`

outofambit: ah thanks @niik, i totally missed the autogenerated ids logic! seems like this PR is not what we want then, so i'll close it.

PR #9142, opened 2020-02-20
Add "create github issue" menu item

### LINKED ISSUE #8989: Easy path to create an issue from GitHub Desktop

outofambit:
Closes #8989 

## Description

- adds menu item that opens the "new issue" page on github for the current repo
- menu item is disabled when user is in a non-github repo
- feature flagged
- repos without issue templates will see a less than ideal web page (empty template list with a small link to open a blank issue at the bottom), but this is getting fixed on server-side github

### Screenshots

screenshot of create issue menu item

## Release notes

Notes: [Added] Add menu shortcut to create GitHub issues

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -323,6 +323,15 @@ export function buildDefaultMenu({
       },
       separator,
       {
+        id: 'create-issue-in-repository-on-github',
+        label: __DARWIN__
+          ? 'Create Issue on GitHub'
+          : 'Create &Issue on GitHub',
+        accelerator: 'CmdOrCtrl+Shift+I',
+        click: emit('create-issue-in-repository-on-github'),
+      },
+      separator,
+      {
         label: __DARWIN__ ? 'Repository Settings…' : 'Repository &settings…',
         id: 'show-repository-settings',
         click: emit('show-repository-settings'),
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -330,7 +330,7 @@ export function buildDefaultMenu({
         id: 'create-issue-in-repository-on-github',
         label: __DARWIN__
           ? 'Create Issue on GitHub'
-          : 'Create &Issue on GitHub',
+          : 'Create &issue on GitHub',
         accelerator: 'CmdOrCtrl+Shift+I',
         click: emit('create-issue-in-repository-on-github'),
         visible: enableCreateGitHubIssueFromMenu(),
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -5,7 +5,11 @@ import { truncateWithEllipsis } from '../../lib/truncate-with-ellipsis'
 import { getLogDirectoryPath } from '../../lib/logging/get-log-path'
 import { ensureDir } from 'fs-extra'
 import { openDirectorySafe } from '../shell'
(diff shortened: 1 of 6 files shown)


paulweldon:
On Feb 25, 2020 7:30 PM, Tiernan L <notifications@github.com> wrote:@tierninho commented on this pull request.

This feature looks good. Some things I tested were:
Keyboard shortcutLocal vs remote limitationsForked repo URLTemplate vs no-template UI
Note, with forked repos, the Issues tab is turned off by default, so we might want to tap grey out the menu item if Issues are disabled.

billygriffin:
@tierninho That's a good point for repos with issues disabled! The path we have outlined would take you here: https://github.com/ruby/ruby/issues/new/choose

Screen Shot 2020-02-26 at 8 50 59 AM

Then clicking the "Open a blank issue" gets you a 404 page. 😬 

@outofambit I don't think accounting for this needs to hold up shipping this for v1, but if it's relatively simple to detect, I like @tierninho's idea of graying out the menu item for repos where issues are disabled.

outofambit:
good catch @tierninho and @billygriffin. we can detect that and i'll add that logic in a follow up PR soon!

tierninho:
An archived repo cannot accept comments also, but a little of an edge case.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -322,6 +326,16 @@ export function buildDefaultMenu({
         click: emit('open-external-editor'),
       },
       separator,
+      {
+        id: 'create-issue-in-repository-on-github',
+        label: __DARWIN__
+          ? 'Create Issue on GitHub'
+          : 'Create &Issue on GitHub',
```
niik: I don't believe `issue` should be title cased on Windows.
```suggestion
          : 'Create &issue on GitHub',
```

app/src/ui/app.tsx
```diff
@@ -1100,6 +1102,19 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+  /**
+   * Opens a browser to the issue creation page
+   * of the current GitHub repository.
+   */
+  private createIssueInRepositoryOnGitHub() {
+    const url = this.getCurrentRepositoryGitHubURL()
+
+    if (url) {
+      this.props.dispatcher.openInBrowser(`${url}/issues/new/choose`)
+      return
```
niik: Can we live without this explicit return?
```suggestion
```

outofambit: we can and we will!

(fwiw i copy pasta-ed this from other url-opening methods in `app.tsx`, so maybe we can clean those up next time we mess with them.)

Issue #8748, opened 2019-12-05
Expose specific git http settings in UI


jasonmacgowan:
### Describe the feature or problem you’d like to solve

When using `schannel` in a corporate proxy environment, `git` operations may fail when attempting to check the TLS certificate of the remote server against a revocation list (see [known issues](https://github.com/desktop/desktop/blob/development/docs/known-issues.md#certificate-revocation-check-fails---3326)).

Implementing the workaround as described requires manually editing a `gitconfig` file which may be challenging for some users who are not used to modifying configuration files.  This file also expects unix-style line endings which may be difficult to do in a Windows environment.

### Proposed solution

Expose `git` configuration settings `http.sslBackend` and `http.schannelCheckRevoke` in GitHub Desktop settings, perhaps under the existing "git" or "advanced" tab.

niik:
Hi @jasonmacgowan, thanks for opening this.

As stated in the known issues document we don't recommend turning off certificate revocation checks except as a temporarily escape hatch. Certificate revocation is an important piece to ensuring secure communications which makes me a little bit hesitant to include a setting to toggle this in the preferences dialog. If we do decided to implement something like that I would like for us to think about cautioning the user when they turn it on to prevent users accidentally turning it on/off as a general network troubleshooting step.

I'd like to understand under what scenario having a corporate proxy would prevent certificate revocation checks. I was under the impression that as long as the system level proxy configuration (i.e. OS-level setting) was properly configured then Windows would ping the revocation servers using that proxy. [This SO answer](https://stackoverflow.com/a/53090169/2114) provided an interesting point of view and is an argument in favor of us making it easier to opt out of revocation checks.

We're currently discussing two H2 initiatives which relate to this issue. The first is an overall support for proxies (https://github.com/desktop/desktop/issues/2789), the second is a redesign of the preferences dialog. I'd like to tee up a discussion around this for the next Desktop product meeting (in two weeks) to get some more input before we move forward with this.

jasonmacgowan:
@niik I am still actively troubleshooting the nature of the revocation check failure during the clone process and do not have a great answer for why it fails in some proxy environments but not in others.  I agree in theory the revocation check would go through the proxy and should just work.  When I run it down, I'll update here.

The spirit of this issue is aligned with #2789, where a Desktop user needs to make an advanced git configuration change where it is difficult to do so.

[...]

PR #9188, opened 2020-02-27
Support disabling certificate revocation checks on Windows

### LINKED ISSUE #3326: Provide a better error message for schannel errors
We've had a good number of reports of schannel errors. We should parse these errors and present a more helpful dialog.

* https://github.com/desktop/desktop/issues/3300
* https://github.com/desktop/desktop/issues/3316
* https://github.com/desktop/desktop/issues/3260
* https://github.com/desktop/desktop/issues/3156
* https://github.com/desktop/desktop/issues/2920
* https://github.com/desktop/desktop/issues/2884
* https://github.com/desktop/desktop/issues/2471
* https://github.com/desktop/desktop/issues/2452
* https://github.com/desktop/desktop/issues/2414
* https://github.com/desktop/desktop/issues/2187

niik:
Closes #3326, #8748

## Description

A (hopefully temporary) workaround for the issues outlined in #9154 when using GitHub Desktop with a MITM proxy server which issues certificates without CRL Distribution Point attributes. This PR only affects Windows. See our [known issues document](https://github.com/desktop/desktop/blob/development/docs/known-issues.md#certificate-revocation-check-fails---3326) for more details.

In addition to the dialog allowing users to disable certificate revocation checks there’s also some seemingly unrelated changes in this PR that deals with ensuring that a dialog is positioned within the bounds of the application window. This change was necessary because when adding the new checkbox to the advanced section of the preferences dialog the dialog became so tall that when switching to the advanced section with the app window at its minimum size the save and cancel buttons would be positioned below the window bounds.

### Isn't this a terribly heavy hammer to use?

Yeah, it sure is. What we're ultimately striving towards if to have the same behavior as we do on macOS today, i.e. that we'll attempt to check the revocation status of a certificate but not error if the revocation check fails, i.e. a best-effort type of revocation check. This is coincidentally what all major browsers and operating systems do as well.

Getting to that point will require changes to several upstream projects and the cooperation of their maintainers so it's not something we can plan on achieving in time for the next release. @dscho has been absolutely wonderful in helping us get the ball rolling on this.

### Screenshots

Screen Shot 2020-02-27 at 15 50 06
Screen Shot 2020-02-27 at 15 51 26

### TODO

- [x] Hide this behind a feature flag
- [x] Add a way to toggle the behavior in the preferences dialog


billygriffin:
@niik Thanks! I'm curious whether this should live in "Advanced" or in "Git." It felt a bit odd to me that there would be a subheading of "Git" in the Advanced tab, as opposed to just putting it in the Git tab. But maybe we don't want to expose that unless you really should be changing it, and therefore Advanced makes sense.

Additionally, I assume this option just becomes available for everyone, or should we scope it to only people who it's relevant for?

onur1907 (dismissed):

niik:
> @niik Thanks! I'm curious whether this should live in "Advanced" or in "Git." It felt a bit odd to me that there would be a subheading of "Git" in the Advanced tab, as opposed to just putting it in the Git tab.

Yeah, I went back and forth on that a few times. FWIW it felt pretty weird having an "Advanced" section in the Git tab as well 😛

I tried doing a optional section in the Git tab

Screen Shot 2020-02-28 at 14 32 25

Screen Shot 2020-02-28 at 14 32 38

Ultimately though I felt like I wanted to emphasize the advanced nature of the preference and sticking it in the tab which deals with something so high-level as name and email felt... wrong I guess?

I'm not married to the current state but the above is the thought process leading up to it. If you feel strongly about not having it the advanced tab I'd be happy to revisit.

> Additionally, I assume this option just becomes available for everyone, or should we scope it to only people who it's relevant for?

So right now it's only visible for Windows users but it is visible for all of them. I considered hiding it unless they've explicitly chosen to opt out. Considering that we've both had the same thought I think we should probably try that direction. I'll make it so

billygriffin:
Thanks @niik. I'm good with everything you said and the current location of "Advanced" given the limited scope of people who this will be relevant for.

app/src/ui/dialog/dialog.tsx
```diff
@@ -143,9 +143,56 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
...
+      return
+    }
+
+    const padding = 10
+    const overflow = offsetTop + offsetHeight + padding - window.innerHeight
+
+    if (overflow > 0) {
+      const top = Math.max(titleBarHeight, offsetTop - overflow)
+      this.dialogElement.style.top = `${top}px`
+    }
```
outofambit: i would really like to solve this in our scss if possible. @niik is there something preventing us from writing roughly equivalent logic into our base dialog styles?

outofambit (commented):
This looks good. I've one question I'd like to talk about but I don't think its a blocker to getting this on beta.

app/src/ui/dialog/dialog.tsx
niik: > is there something preventing us from writing roughly equivalent logic into our base dialog styles?

Yeah, this is something we talked about at when we initially implemented the dialogs. The chromium logic takes care of centering the dialog when it's first opened but doesn't move the dialog after that. That's the behavior we want. If we were to use CSS to position all dialogs at the vertical midpoint they'd start jumping around when switching between tabs in dialogs or when the contents of the dialog changes for other reasons (error messages dynamically appearing/disappearing for example).

The problem we're running into here is when a dialog can have wildly different heights depending on its configuration which is the case for the preferences dialog where the entry tab is very short:

image

And the advanced tab is much much taller

image

I see this change in this PR as a safe-guard to ensure that we don't ever end up in scenarios where the dialog ends up extending beyond the bounds of the window. A much better solution here would be to rebalance the tabs in the preferences dialog so that each tab is more uniform in height. We could for example add the "save/cancel" footer to the accounts and theme tabs and we could consider extracting the "show a confirmation dialog when..." section of the advanced tab into its own "Prompts" tab.

TL;DR; I don't see this as a solution in any way shape or form. Our dialogs aren't intended to differ wildly in height and we should ensure that the preferences dialog doesn't do that either. It is, however, a safe-guard.

outofambit (approved):

mphmahla (commented):
Thanks

PR #9206, opened 2020-03-02
Add GitHubRepository test fixture utility

outofambit:
supports further work related to #8989 

## Description

We call the `GitHubRepository` constructor directly in many places in our unit tests. Since we enumerate the arguments separately, this means that anytime we update the `GitHubRepository` constructor with new fields, a bunch of tests break even if they don't use that field. I've run into this multiple times while working on forks and issues integration in desktop.

This PR adds a little factory function to wrap the GitHubRepository constructor to minimize churn in our unit tests.

### DIFF
--- a/app/test/helpers/github-repo-builder.ts
+++ b/app/test/helpers/github-repo-builder.ts
@@ -4,8 +4,7 @@ import { getDotComAPIEndpoint } from '../../src/lib/api'
 
 let id_counter = 0
 
-// TODO: rename me!
-export function plainGitHubRepo({
+export function gitHubRepoFixture({
   owner,
   name,
   parent,
--- a/app/test/helpers/github-repo-builder.ts
+++ b/app/test/helpers/github-repo-builder.ts
@@ -40,6 +40,9 @@ export function gitHubRepoFixture({
   endpoint,
   isPrivate,
 }: IGitHubRepoFixtureOptions): GitHubRepository {
+  const htmlUrl = `${
+    endpoint !== undefined ? endpoint : 'https://github.com'
+  }/${owner}/${name}`
   return new GitHubRepository(
     name,
     new Owner(
@@ -49,13 +52,9 @@ export function gitHubRepoFixture({
     ),
     id_counter++,
     isPrivate !== undefined ? isPrivate : null,
-    endpoint !== undefined
-      ? `${endpoint}/${owner}/${name}`
-      : `https://github.com/${owner}/${name}`,
+    htmlUrl,
     defaultBranch || 'master',
-    endpoint !== undefined
-      ? `${endpoint}/${owner}/${name}.git`
-      : `https://github.com/${owner}/${name}.git`,
+    `${htmlUrl}.git`,
     null,
     parent
   )
--- a/app/test/helpers/github-repo-builder.ts
+++ b/app/test/helpers/github-repo-builder.ts
@@ -4,19 +4,38 @@ import { getDotComAPIEndpoint } from '../../src/lib/api'
 
 let id_counter = 0
 
+/**
+ * Most of these fields are passed on to the
+ * GitHubRepository constructor directly.
+ *
+ * Notable exception: `endpoint`
+ */
+interface IGitHubRepoFixtureOptions {
+  owner: string
+  name: string
+  parent?: GitHubRepository
+  defaultBranch?: string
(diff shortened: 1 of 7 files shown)


app/test/helpers/github-repo-builder.ts
```diff
@@ -0,0 +1,62 @@
...
+    ),
+    id_counter++,
+    isPrivate !== undefined ? isPrivate : null,
+    endpoint !== undefined
+      ? `${endpoint}/${owner}/${name}`
+      : `https://github.com/${owner}/${name}`,
+    defaultBranch || 'master',
+    endpoint !== undefined
+      ? `${endpoint}/${owner}/${name}.git`
+      : `https://github.com/${owner}/${name}.git`,
```
kuychaco: Minor suggestion -- this is so similar to the lines above it that we could extract a variable like so:

```
const htmlUrl = endpoint !== undefined
      ? `${endpoint}/${owner}/${name}`
      : `https://github.com/${owner}/${name}`
```

And the `cloneURL` argument would be `${htmlUrl}.git`

app/test/unit/infer-comparison-branch-test.ts
```diff
@@ -35,29 +35,16 @@ function createTestGhRepo(
   defaultBranch: string | null = null,
   parent: GitHubRepository | null = null
 ) {
-  if (owner.indexOf('/') !== -1) {
-    throw new Error(
-      'Providing a slash in the repository name is no longer supported, please update your test'
-    )
-  }
```
kuychaco: Wondering if this check is still needed. I notice we're removing it and not incorporating a check elsewhere.

app/test/unit/infer-comparison-branch-test.ts
```diff
@@ -35,29 +35,16 @@ function createTestGhRepo(
...
+  return gitHubRepoFixture({
+    owner: owner,
+    name: 'my-cool-repo',
+    defaultBranch: `${
       defaultBranch !== null && defaultBranch.indexOf('/') !== -1
         ? defaultBranch.split('/')[1]
         : defaultBranch
     }`,
-    cloneURL,
-    'write',
```
kuychaco: Ah okay, looks like we default to having write permissions so it's not necessary to specify `write` 👍 -- https://github.com/desktop/desktop/blob/88e9777a0a847eeaff3aa8840e2d745130160bdd/app/src/models/github-repository.ts#L58-L65

app/test/unit/repositories-list-grouping-test.ts
```diff
@@ -3,28 +3,22 @@ import {
...
-        1
-      ),
+      gitHubRepoFixture({ owner: 'me', name: 'my-repo2' }),
       false
     ),
     new Repository(
       'repo3',
       3,
-      new GitHubRepository('my-repo3', new Owner('', '', null), 1),
+      gitHubRepoFixture({ owner: '', name: 'my-repo3', endpoint: '' }),
```
kuychaco: I'm curious about this case of `endpoint` being an empty string... is it particularly important for this test?

app/test/unit/infer-comparison-branch-test.ts
outofambit: yeah, i removed this because we don't have this check in any other places where we name a repository (and removing it didn't break anything). it was added in 3fe0747133274f97d92c3a9249c62f57c22e0fd3, presumably out of an abundance of caution as i can't find any evidence in the git history that we were ever using a `/` in our repository names.

app/test/helpers/github-repo-builder.ts
outofambit: ah nice! done in https://github.com/desktop/desktop/pull/9206/commits/44a9fb957c2059c7e67fad269ae19d29c9a15c7a

app/test/unit/repositories-list-grouping-test.ts
outofambit: thanks for the nudge, i wondered that too, but didn't dig in...until now. 😄

it is important, only in that we need it to be a non-github dot com endpoint (its supposed to be a repo on gh enterprise instead). i've updated the test to use a more obvious "dummy" endpoint in https://github.com/desktop/desktop/pull/9206/commits/21773bb76077eb19dba9dcde233754938b1815c2.

PR #9207, opened 2020-03-02
Disable creating issues from desktop on repos that have issues disabled

outofambit:
continues https://github.com/desktop/desktop/pull/9142#pullrequestreview-364568677

## Description

- store GitHub API result, which can include the `has_issues` field
- use this to decide if we should enable the "create issue on github" menu item or not
- some more test cleanup

## Testing

1. create a repo on github
1. disable issues in repo settings
1. clone repo into desktop
1. verify that "Create Issues" menu item is disabled when in that repo in desktop.

would be great to verify the same steps for a repo that is archived, too!

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -104,6 +104,7 @@ export interface IAPIRepository {
   readonly default_branch: string
   readonly pushed_at: string
   readonly parent?: IAPIRepository
+  readonly has_issues?: boolean
 
   /**
    * The high-level permissions that the currently authenticated
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -104,7 +104,7 @@ export interface IAPIRepository {
   readonly default_branch: string
   readonly pushed_at: string
   readonly parent?: IAPIRepository
-  readonly has_issues?: boolean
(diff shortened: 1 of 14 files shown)


tierninho:
After a short delay, toggling the Issues flag on Github.com is reflected in Desktop properly. ✅ 

I couldn't get an archived repo to properly disable the Menu item, despite being rejected (as expected) for trying to Push a commit.

app/src/lib/api.ts
```diff
@@ -104,6 +104,7 @@ export interface IAPIRepository {
   readonly default_branch: string
   readonly pushed_at: string
   readonly parent?: IAPIRepository
+  readonly has_issues?: boolean
```
niik: I was under the impression that `has_issues` is a mandatory field in API responses except for a few rare cases which we shouldn't have to concern ourselves with (public repository listing, checks, notifications etc). Do we know of an endpoint that we rely on to upsert GitHubRepositories which doesn't provide `has_issues`?

outofambit: ah, you are right! that was my mistake.
```suggestion
  readonly has_issues: boolean
```

PR #9248, opened 2020-03-09
Updated shell integration docs to reflect code

AndreiMaga:
## Description

- updated the shell integration docs to reflect the changes in code

## Release notes

Notes: no-notes

### DIFF
--- a/docs/technical/shell-integration.md
+++ b/docs/technical/shell-integration.md
@@ -29,7 +29,7 @@ These shells are currently supported:
  - [Hyper](https://hyper.sh/)
  - Git Bash (from [Git for Windows](https://git-for-windows.github.io/))
  - [Cygwin](https://www.cygwin.com/)
- - WSL
+ - [WSL](https://docs.microsoft.com/en-us/windows/wsl/about)
  - [Windows Terminal](https://github.com/microsoft/terminal)
 
 These are defined in an enum at the top of the file:
--- a/docs/technical/shell-integration.md
+++ b/docs/technical/shell-integration.md
@@ -28,6 +28,9 @@ These shells are currently supported:
  - [PowerShell Core](https://github.com/powershell/powershell/)
  - [Hyper](https://hyper.sh/)
  - Git Bash (from [Git for Windows](https://git-for-windows.github.io/))
+ - [Cygwin](https://www.cygwin.com/)
+ - WSL
+ - [Windows Terminal](https://github.com/microsoft/terminal)
 
 These are defined in an enum at the top of the file:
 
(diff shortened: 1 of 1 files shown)


docs/technical/shell-integration.md
```diff
@@ -28,6 +28,9 @@ These shells are currently supported:
  - [PowerShell Core](https://github.com/powershell/powershell/)
  - [Hyper](https://hyper.sh/)
  - Git Bash (from [Git for Windows](https://git-for-windows.github.io/))
+ - [Cygwin](https://www.cygwin.com/)
+ - WSL
```
say25:
```suggestion
 - [WSL](https://docs.microsoft.com/en-us/windows/wsl/about)
```

docs/technical/shell-integration.md
```diff
@@ -28,6 +28,9 @@ These shells are currently supported:
  - [PowerShell Core](https://github.com/powershell/powershell/)
  - [Hyper](https://hyper.sh/)
  - Git Bash (from [Git for Windows](https://git-for-windows.github.io/))
+ - [Cygwin](https://www.cygwin.com/)
+ - [WSL](https://docs.microsoft.com/en-us/windows/wsl/about)
+ - [Windows Terminal](https://github.com/microsoft/terminal)
```
outofambit: @AndreiMaga could we add a note that WSL is only on beta currently? thanks!

```suggestion
 - [Windows Terminal](https://github.com/microsoft/terminal) (on beta only)
```

say25: If I'm not mistaken, following the release of 2.4 this will be in "prod" the docs are not shipped so it seems to me like the beta only note is not necessary?

outofambit: @say25 i'm not aware of any discussion around flagging WSL detection for prod, but maybe i missed something? the feature flag is still beta-only:

https://github.com/desktop/desktop/blob/6c1800a06d2caaf9b6cebed8345cced1b32b4019/app/src/lib/feature-flag.ts#L84-L86

AndreiMaga: @outofambit In my opition "only in beta" is not a good term for it as u can enable the functionality with an environment variable in "prod". Maybe a little "flag" like the one for dark mode would suit it better?
```suggestion
 - [WSL](https://docs.microsoft.com/en-us/windows/wsl/about) (beta)
```

outofambit: ohhh good point @AndreiMaga! that sounds good to me!

outofambit: ah i also just realized i suggested on the wrong line 🙈 sorry!

say25: that was the confusion Windows Terminal vs WSL 😅

outofambit: ohhhhh! sorry @say25 !

Issue #9185, opened 2020-02-27
urls within commits get cut off and cannot get opened by the browser


staabm:
### Describe the bug

lookup a commits within the history of a project which contains a url.
the url - when displayed cut off - because of its length does not open properly in the browser

### Version & OS

Windows10x64
GHD Version 2.3.1

### Steps to reproduce the behavior

1. clone https://github.com/phpstan/phpstan-src
2. open the history
3. select the commit 7d2ecf9a1f911f4bc3d64373119097e927084221
4. click the url contained in the commit

![image](https://user-images.githubusercontent.com/120441/75432035-130a8880-594e-11ea-8020-aeb315fc61b5.png)

### Expected behavior

the url opens in the browser

### Actual behavior

the browser does not open the actual url from the commit `https://github.com/phpstan/phpstan/issues/3008` but it seem to open the cut-off version of it `https://github.com/phpstan/phpstan/i%E2%80%A6` -> which leads to a 404 http error page on github.com

opening the same url from within github.com works like expected, see https://github.com/phpstan/phpstan-src/commit/7d2ecf9a1f911f4bc3d64373119097e927084221

steveward:
Thanks for the detailed report @staabm! I'm going to label this as a bug.

halitogunc:
Mind if I fix that one ? I have one solution in my mind without breaking truncated summary. The thing is if I choose that way, I have to add an extra optional prop to `<RichText />` component, and also a new parameter for scanForHyperlink function. What do you think ?

steveward:
@halitogunc that'd be great. Pinging @desktop/engineering for thoughts on your question.

Issue #9061, opened 2020-02-06
List of "Changed files" shows only the last file


VKostenc:
### Describe the bug
"Changes" tab shows empty space and then the last(by order) changed file.

### Version & OS
GitHub Desktop 2.3.1 & MacOS Mojave 10.14.6

### Steps to reproduce the behavior

1. Make changes in several files
2. Open GitHub Desktop
3. Switch from "Changes" to "History" tab
4. Switch back to the "Changes" tab
5. See error

### Expected behavior

List of changed files should be represented.

### Actual behavior

List of changed files is missed. You can see the space where they should be located and only the last file.

### Screenshots
<img width="1044" alt="Screen Shot 2020-02-06 at 3 14 18 PM" src="https://user-images.githubusercontent.com/13135644/73940156-9dce0980-48f3-11ea-9d7a-a8bf0b7f7481.png">

### Additional context

The issue appears after update to the latest GitHub Desktop.
Tried to reboot the app.

steveward:
@VKostenc thanks for the report! This definitely looks like a bug. Are you able to consistently reproduce this issue across multiple repositories, or is it only happening in one specific repository?

tierninho:
This is the second report of this bug, with the other being: https://github.com/desktop/desktop/issues/8919

SmilinBrian:
I'm seeing this sporadically - same OS and GitHub Desktop version.

SmilinBrian:
I'm seeing this somewhat consistently with the same OS & GitHub Desktop versions.
Log enclosed in case it is helpful.
[2020-02-07.desktop.production.log](https://github.com/desktop/desktop/files/4173162/2020-02-07.desktop.production.log)

VKostenc:
Tried to reproduce with other repositories but without success. 
The issue was disappeared after reboot of the machine.
@steveward Should I close the issue? Or further investigation is needed because of appearing in other users?

slawkens:
Same error happened for me today.

Windows 10 Pro. Latest GitHub Desktop version.

The issue appears when I scroll through History Tab and then switch back to Changes Tab. When I scroll to the top of the History Tab then switch back to Changes Tab then issue disappears.

Screencast:
![github_error](https://user-images.githubusercontent.com/16687302/74373891-9d46de80-4ddd-11ea-83b1-97567aa682a3.gif)

tierninho:
@slawkens Many thanks for that video. I was able to reproduce this. Investigating now for a fix!

[...]

PR #9259, opened 2020-03-10
Use not truncated url as href to get opened by the browser 

halitogunc:
This PR fixes an issue that was created in #9185 

## Description

- Urls within commit summary get cut off and make clicking url unworkable to open on the browser. My plan was mentioned here https://github.com/desktop/desktop/issues/9185#issuecomment-594044174. So I have added a new prop called url to `<RichText>` component and used it through tokenization process to override default url if it is given. My main motivation was keeping to current structure if extra `url` prop is not provided, so it shouldn't break any prior logic. I have also added two simple test cases for github and non-github repositories. 

- To get url in a string, I used https://github.com/sindresorhus/get-urls lib. (new dependency)

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -35,6 +35,7 @@
     "fs-admin": "^0.12.0",
     "fs-extra": "^6.0.0",
     "fuzzaldrin-plus": "^0.6.0",
+    "get-urls": "^9.2.0",
     "keytar": "^5.0.0",
     "mem": "^4.3.0",
     "memoize-one": "^4.0.3",
--- a/app/src/lib/text-token-parser.ts
+++ b/app/src/lib/text-token-parser.ts
@@ -202,6 +202,7 @@ export class Tokenizer {
 
   private scanForHyperlink(
     text: string,
+    url: string,
     index: number,
     repository?: GitHubRepository
   ): LookupResult | null {
@@ -228,7 +229,7 @@ export class Tokenizer {
           const idText = issueMatch[1]
           this._results.push({
             kind: TokenType.Link,
-            url: maybeHyperlink,
+            url: url || maybeHyperlink,
             text: `#${idText}`,
           })
           return { nextIndex }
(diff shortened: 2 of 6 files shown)


tierninho:
@halitogunc Thanks for the PR. I link works in example you provided from the original issue, however when I try to recreate the situation, I end up with a blank commit:

Screen Shot 2020-03-10 at 3 21 21 PM

Recreate by:
- Make a change to a file
- In subject line, add some text, then past a URL
- Commit, and view commit in History

I tried this in our current production version and the bug was not reproducible, so it looks like it stems from your PR. Reach out if you need assistance, thanks!

halitogunc:
Hey! @tierninho. I tested it with different urls within commit summary and couldn't reproduce it. Can you please share your commit message fully ?

tierninho:
Sure! I used commit description: "fsdf sdfsd sdf sdf sdf sdf sfd sdf https://phpstan.org/r/16ff6510-48da-48de-a657-8674a666a688".

halitogunc:
desktop-app

still can't reproduce on my local. The only thing that comes to my mind is that the error you encountered may be related to https://github.com/desktop/desktop/issues/9061 or it was a fluke ?

tierninho:
Perhaps it was a fluke. I "resolved" it by mousing up on the commits and now it appears and works fine. 

Regardless, if the next reviewer can double check that would be great. 

Thanks for the work @halitogunc !

halitogunc:
Ping! @desktop/engineering
Can anyone review the changes please?

niik:
Hey @halitogunc, thanks for opening this and wanting to contribute! We really appreciate the effort you've put in here!

Thanks to us being able to look at your implementation we were able to identify several issues that we previously hadn't considered having to do with how to wrap links that are shortened (issue links), multiple links in a commit summary, emojis, etc that ultimately led us to realize that this problem can't be solved entirely within the `Tokenizer` class.

As such we've opted to work on an approach of a token-aware wrapper built into the `CommitSummary` class which lets us make decisions around what should be part of the commit summary and what should overflow into the body in a more holistic way.

I've opened #9931 to get some feedback on that approach from the rest of the team and I've added you as a co-author on the initial proof of concept as recognition for the hard work you've already done.

I will be closing this PR I favor of #9931 so please follow along there to see if and when we're able to solve this issue.

PR #9267, opened 2020-03-12
Pre-launch documentation and brain-dump for proxy support

niik:
As I'll be out when it's time to release GitHub Desktop 2.4 which will include the proxy support I worked on I wanted to do a bit of a brain dump and provide some thoughts around troubleshooting should anything go wrong with the proxy support once it hits production.

I likely won't be around to make any changes to this if there are review requests so I kindly ask y'all to use your best judgement and apply whatever changes you feel necessary in my absence.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -45,3 +45,5 @@ and how things work:
     to reduce the friction of getting started.
  - **[Building and Packaging Desktop](technical/packaging.md)** - Outlines how
     Desktop is currently packaged for all platforms 
+ - **[Automatic Git Proxy support](technical/proxies.md)** - A pre-launch overview
+    and troubleshooting guide for the Git automatic proxy support in GitHub Desktop.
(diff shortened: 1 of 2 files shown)


tierninho:
Great write up! If you are still around, here are two questions I had regarding remote endpoints. No worries if you can't get to them now.

1. If submodules or the LFS server is different from the host, do we have an error messaging in place that alerts to the user to check the server URL or auth?

2. If the user happens to be using http only, and we guess wrong, will they be alerted?

niik:
> If submodules or the LFS server is different from the host, do we have an error messaging in place that alerts to the user to check the server URL or auth?

No, and we’ve always had the assumption that multi-host repositories are very rare so chances are that since the same credentials work for these setups (if any exist) today then the same proxy ought to as well. The one hypothetical scenario I can think of would be a game development company which has their core logic on GitHub but use an on-prem LFS that _shouldn’t_ be accessed through the proxy. For those scenarios there’s the `NO_PROXY` environment variable as well as the option to set `http.proxy` for a specific host (`http.hostname.proxy`)

> If the user happens to be using http only, and we guess wrong, will they be alerted?

If they attempt to use http against GHES or GitHub.com they will be automatically redirected to use https. GitHub doesn’t support Git over vanilla http.

PR #9270, opened 2020-03-12
Default "create an issue" to create issue in parent repository

### LINKED ISSUE #9232: Default "create an issue" to create it in the parent repository

Currently we default to your *current* repository when creating an issue. However, based on #9170 I think we want to default to the parent repository instead.

### Proposed solution

If you're in a fork and you select "Create an issue," instead default the issue creation to the parent repository.

### Additional context

We know this isn't what *everyone* wants, but it does represent the vast majority of use cases, where the fork exists in order to allow someone to contribute to the upstream repository. Therefore, it makes more sense to default the create issue URL to the parent as well.

kuychaco:
Closes #9232

## Description

Currently clicking "Create Issue on GitHub" creates an issue on the current repository opened in Desktop. 

With this PR, if the current repo is a fork and the parent repo (upstream) has issues enabled, then clicking "Create Issue on GitHub" creates an issue on the parent/upstream repository. 

If there is a parent/upstream and issues are disabled (or the repo has been archived) then the "Create Issue on GitHub" menu item is disabled/greyed out. We believe that in the vast majority of use cases a fork exists in order to allow someone to contribute to the upstream repository, so we assume issue creation is also intended for the upstream repository.

And if there is no parent/upstream associated with the current repository, then clicking "Create Issue on GitHub" creates an issue on the current repository opened in Desktop (current behavior).

## Release notes

Notes:

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -162,7 +162,10 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
   let rebaseInProgress = false
   let branchHasStashEntry = false
   // check that its a github repo and if so, that is has issues enabled
-  const repoIssuesEnabled = getRepoIssuesEnabled(state)
+  const repoIssuesEnabled =
+    selectedState !== null &&
+    selectedState.repository instanceof Repository &&
+    getRepoIssuesEnabled(selectedState.repository)
 
   if (selectedState && selectedState.type === SelectionType.Repository) {
     repositorySelected = true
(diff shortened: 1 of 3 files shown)


app/src/lib/menu-update.ts
```diff
@@ -384,6 +379,30 @@ function getNoRepositoriesBuilder(state: IAppState): MenuStateBuilder {
   return menuStateBuilder
 }
 
+function getRepoIssuesEnabled(state: IAppState): boolean {
```
outofambit: 💯 this helper method

are there other helpers in this file that use all of `state`? if that's not an established pattern, then i'd prefer to only pass in `selectedState` or `selectedState.repository` instead of all of `state`, to help keep the scope of this helper constrained.

app/src/ui/app.tsx
```diff
@@ -1108,7 +1108,11 @@ export class App extends React.Component<IAppProps, IAppState> {
    * of the current GitHub repository.
    */
   private createIssueInRepositoryOnGitHub() {
-    const url = this.getCurrentRepositoryGitHubURL()
+    // Default to creating issue on parent repo
+    // See https://github.com/desktop/desktop/issues/9232 for rationale
+    const url =
+      this.getParentRepositoryGitHubURL() ||
+      this.getCurrentRepositoryGitHubURL()
```
outofambit: nice! :)

PR #9301, opened 2020-03-17
Set min-height of FilterList to make it scrollable under any situation

### LINKED ISSUE #9175: Resize layout problem with post welcome flow landing page
### Describe the bug

The post-welcome flow landing page on Windows doesn't respond as intended when being resized. Assuming this is related to #8791 or #8959 (i.e #8967).

Tentatively putting this in the 2.4 release as it appears to be a regression from what's on prod at the moment.

### Version & OS

Latest development build (452c89d0de300878e2320c5da7bdeb42e949d765). Windows 10 and macOS 10.15.3

1. Remove all repositories to get to the post-welcome flow landing page
2. Wait until the list of cloneable repositories has loaded
3. Select any repository from the list such that the blue clone button shows up at the bottom of the page

image

4. Resize the application window

### Expected behavior

That the ProTip box as well as the clone button to be anchored toward the bottom of the view

image

### Actual behavior

The clone button and the ProTip box are positioned outside of the bounds of the application Window.

image

### Screenshots

welcome-resize-bug

### Logs

N/A

rafeca:
Closes #9175

## Description

This PR adds the `min-height` property to the `.filter-list-container` selector. This allows it to shrink its height to a value smaller than its content size and show a scrollbar whenever there are too many items to fit on the screen.

### Alternative solutions

I've tested using `overflow: auto` instead and it also works completely fine, so a solution could be to just use that (which is a more explicit rule).

I've ended up choosing `min-height` since it feels less disruptive (specially since the `FilterList` container is used in many places), but I'm not attached to this decision and I would be happy to change it.

### Screenshots

demo

## Release notes

Notes:

- Fix layout bug on post-welcome screen for big displays.

### DIFF
--- a/app/styles/ui/_filter-list.scss
+++ b/app/styles/ui/_filter-list.scss
@@ -19,6 +19,9 @@
   &-container {
     display: flex;
     flex: 1;
+    // The filter-list-container div should be scrollable if its contents
+    // are larger than the available space.
+    min-height: 0;
   }
 }
 


billygriffin:
@rafeca Thanks! 

> I've ended up choosing min-height since it feels less disruptive (specially since the FilterList container is used in many places)

Did you by chance look at the other places where this is used to ensure it doesn't break any existing functionality that is working properly?

rafeca:
> Did you by chance look at the other places where this is used to ensure it doesn't break any existing functionality that is working properly?

I've checked in the most predominant places where the `FilterList` component is used:

1. Branch selector

    Screenshot 2020-03-17 at 17 43 17

2. Repo history:

    Screenshot 2020-03-17 at 17 43 32

I'm going to do an additional check in every callsite of `FilterList`.

PR #8307, opened 2019-09-22
Fix arrow key navigation of Changes and History lists (#6845)

JMcKiern:
## Overview

**Closes #6845** by reworking #6481.

## Description

With #6481, the scroll position would be set every time the sidebar was rendered.
In this pr, the scroll position is only updated when the sidebar is changed (determined by which sidebar is being rendered).
This allows the user to navigate through the changes and history commit lists using the arrow keys, addressing the issue brought up in #6845.

## Release notes

Notes: no-notes

PR #9306, opened 2020-03-17
Use the correct scroll variable from the state to persist scroll

### LINKED ISSUE #9061: List of "Changed files" shows only the last file

rafeca:
Closes #9061

## Description

This PR fixes an issue when switching from the History tab to the Changes tab after scrolling.

The fix consists on using the correct variable that stores the scroll position of the Changes tab instead of the scroll position of the History tab. The error was probably introduced by mistake in https://github.com/desktop/desktop/pull/8307

### Screenshots

Before:

before

After:

after

## Release notes

Notes:

- Fixed empty spaces shown on the "Changes" tab after switching from the "History" tab.

### DIFF
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -163,7 +163,7 @@ export class RepositoryView extends React.Component<
 
     const scrollTop =
       this.previousSection === RepositorySectionTab.History
-        ? this.state.compareListScrollTop
+        ? this.state.changesListScrollTop
         : undefined
     this.previousSection = RepositorySectionTab.Changes
 


rafeca:
I'm going to check how feasible is to add a test to not regress here...

rafeca:
> I'm going to check how feasible is to add a test to not regress here...

I haven't been able to create such automated test: from what I see unit tests are meant to test very specific/isolated JS logic and the integration test checks the launching behaviour. We would need some kind of end to end tests to be able to automatically verify this.

Issue #9312, opened 2020-03-18
Show the upstream repository on github when clicking on "view on github"


rafeca:
### Describe the feature or problem you’d like to solve

We're currently redirecting users to the forked repository when clicking on "View on GitHub", but based on #9170 I think we want to default to the parent repository instead.

### Proposed solution

When a user clicks `view on GitHub` while they’re on a fork, we eant to redirect them to the upstream repository on GitHub, not the fork.

### Additional context

In a similar fashion than #9232, this may not be what everyone wants, but based on https://github.com/desktop/desktop/issues/9170 it'll cover the vast majority of use cases, where the fork exists in order to allow someone to contribute to the upstream repository.

tierninho:
Thanks for the issue! @rafeca 

We happen to use an identical "View on GitHub" label for viewing single commits on GitHub.com.

![Screen Shot 2020-03-18 at 7 46 44 AM](https://user-images.githubusercontent.com/14828183/76990932-da2a5780-68ec-11ea-810e-fb5b5fb79197.png)

Although you are referencing repo-level buttons, the identical wording here may cause some confusion. Just wanted to point it out in case we want to rephrase it. I have no option.

rafeca:
Oh thanks for pointing this out @tierninho ! I just talked about this earlier today with @outofambit !

I think that ideally that menu item should also redirect to the upstream repository in normal conditions, but there's something tricky here: we would need to check if the commit is only pushed to the fork and redirect to the fork in that case.

Issue #9170, opened 2020-02-24
Clarify and hopefully align fork vs. upstream preferences


billygriffin:
### Describe the feature or problem you’d like to solve

Currently, there are several places in GitHub Desktop where how the interaction should work in the app when someone is on a fork is highly dependent on someone's relationship to the project. In other words, are they on the fork because they're just trying to contribute to the upstream repo? Is this a "long-running" fork that's intended to diverge entirely from upstream? Something else? In talking through this, we think there may be several connected elements that may operate similarly depending on the preference.

#### Creating an issue

Using fork to contribute to upstream: Create new issue in upstream repo (documented in #9232)
Using fork to contribute to fork: Create issue in fork repo

#### Creating a pull request

Using fork to contribute to upstream: Create PR from fork targeting upstream
Using fork to contribute to fork: Create PR from fork targeting fork (this represents a divergence from the default behavior of GitHub.com, which by default targets new PRs toward upstream)

#### Viewing list of pull requests in the PR tab of the branch list

Using fork to contribute to upstream: View pull requests for the upstream repo
Using fork to contribute to fork: View pull requests for the fork repo

#### Staying up to date

Using fork to contribute to upstream: Ability to easily stay up to date with latest from upstream repo's default branch
Using fork to contribute to fork: Ability to easily stay up to date with latest from fork

#### Any others?

These are the ones we've discussed, but there may be other things that likely depend on this preference about whether you're on a fork with the intention of contributing back to upstream, or on a fork because that's where you're contributing to.

---

My biggest question is whether these are always consistent across these scenarios. In other words, is there ever a time where someone would want to create a PR targeting upstream by default but want to create an issue in their fork (and any of the other combinations)? I think even if that is the case, we could make the defaults align along the lines described here, and give people flexibility to change them.

ampinsk:
@billygriffin and I chatted through this and landed in a much clearer place! So we're talking about two different kinds of forks, "Contribution Forks", where I am using this fork solely for contributing to the upstream parent project, and "Maintained Forks" a fork of a project that I use for my own purposes.

We realized that github.com pretty strongly _discourages_ Maintained Forks in several ways: 

1. When you create a PR from a fork, it always automatically defaults the target branch as the parent. GitHub.com redirects you from your fork's PR page (https://github.com/ampinsk/desktop/compare, for example) to the parent (https://github.com/desktop/desktop/compare)
2. When you create a fork, it has issues disabled automatically 
[...]

PR #9316, opened 2020-03-18
Redirect to upstream repo when clicking on "View on GitHub"

rafeca:
Closes https://github.com/desktop/desktop/issues/9312

## Description

This PR changes the redirection URL when clicking the "View on GitHub" main menu item to open the upstream repository when the current repository is a fork.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1119,7 +1119,9 @@ export class App extends React.Component<IAppProps, IAppState> {
   }
 
   private viewRepositoryOnGitHub() {
-    const url = this.getCurrentRepositoryGitHubURL()
+    const url =
+      this.getParentRepositoryGitHubURL() ||
+      this.getCurrentRepositoryGitHubURL()
 
     if (url) {
       this.props.dispatcher.openInBrowser(url)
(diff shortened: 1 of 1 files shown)


rafeca:
> However, I deleted the upstream in the cmd line and the View on GitHub button/menu item points to the Upstream repo. Same result if you were to change it to point elsewhere. Is this intended?

Good question! From what I see on the code, the upstream repository is calculated using the GitHub API (via the [`parent` field on the repository object](https://developer.github.com/v3/repos/#response-5)). This means that if the cloned repository is a GitHub fork both the "Create issue on GitHub" and "View on GitHub" menu items will redirect to the parent repository (no matter if the local `git` working copy has an `upstream` remote).

Based on https://github.com/desktop/desktop/issues/9170, I think that this is the expected behaviour, but looping in @billygriffin and @ampinsk to doublecheck.

outofambit:
> Based on #9170, I think that this is the expected behaviour,

@rafeca and @tierninho yep i can confirm this is expected behavior. we care more about the repos on github than local repos for this scenario.

outofambit:
appveyor doesn't seem to be picking this up after the base changed, but this change is safe enough that i'm going to just merge it.

Issue #6875, opened 2019-02-19
iterating on the NDDB prompt user experience


shiftkey:
**Please describe the problem you think should be solved**

We've had the "Notification of Diverging from Default Branch" in the app for a while, but some recent changes make me think there's a better way to surface this information.

![](https://user-images.githubusercontent.com/359239/52881834-2cbea080-313c-11e9-8a4d-39b2318341e3.png)

This came up in #6862 because the user wanted a convenient way to update their fork from the upstream, and was under the impression this didn't exist in the app.

The pain points I'd like to call out:

 - seeing this prompt while you're comparing branches feels unintuitive - why is the app distracting the user from what they're trying to do?
 - the compare tab is a very narrow space, limiting what information we can show to the users
 - it's not easy to manually trigger this information, because of how we implemented it at the time

**[Optional] Do you have any potential solutions in mind?**

The blank slate improvements in #6445 have provided real-estate for showing contextual tasks, like pulling from the tracked branch:

![](https://user-images.githubusercontent.com/1322749/52889826-a88b0e00-3179-11e9-9f05-71c735bfe8fe.png)

Seeing this in action - suggesting a task to the user, and then hiding it after it's been completed - feels like a similar flow to what the NDDB prompt was trying to achieve.

How I'd iterate on this feature:

  - extract the logic for detecting and prompting the user from the compare tab and into a service that runs in the background - this gives us the ability to control when it runs, and doesn't tie it to user interactions
  - design an action UI to show when the user should merge their default branch (either from a feature branch or from the upstream of a fork) or dismiss the action (because it's not important)

We've had previous discussions about adding in additional actions (e.g. #6674) but because this action is only shown when something needs to be done I don't think it's as important to do here.

j-f1:
I noticed this morning that making a change causes the suggestions to disappear, which makes it harder to open Atom. Maybe this blank slate should be a separate “home” screen so it’s always available? How about a new button next to the push/pull button that opens a dropdown with the available suggestions? It could even have a numeric badge for the more active suggestions (like merging the main branch or pushing to the remote).

borpin:
> This came up in #6862 because the user wanted a convenient way to update their fork from the upstream, and was under the impression this didn't exist in the app.

I think that part of my issue is that I do not see these prompts.  I suspect this is because the repositories were cloned with Desktop before the feature was introduced.  However, I have added the upstream via the CLI.

On Github I see:
[...]

Issue #5293, opened 2018-07-31
The New Feature for Alerting User When Base Branch Moves Ahead of Current is Inconsistent


JosephTLyons:
## Description
The new feature "Notification displayed in History tab when the base branch moves ahead of the current branch" seems inconsisent.  If I take a base branch, make a commit to it, flip to a new branch and move to the history tab, I would expect to see the alert jump out at me, but instead, I don't see anything.  I don't even see the blue dot on the history tab when in the changes tab.  I only trigger the alert the first time if I click the text box that lets us choose the branch to compare, then I immediately get the alert.  After the first time the alert is triggered, it then begins to work as expected.  If I flip back to a base branch that has the newer commit to it, then back to the non-base branch, the history tab has a blue indicator and clicking on the history tab immediately shows the alert.

## Version
* GitHub Desktop: v1.3.1
* Operating system: macOS 10.13.6

## Steps to Reproduce
1. Using a test repo with just a master branch, make a new branch
2. Add a commit to this new branch
3. Flip back to the master and add a commit to the master
4. Flip back to new branch and expect to see the alert showing that the new branch is behind master (but no alert shows)
5. Click text box to compare and compare with master, alert may show up now
6. Clear the text box for comparing a branch to another
7. Flip back to master, then back to new branch and see that alert pops up now as I would expect.

### Expected Behavior
I'd expect that the alert would show up in the same manner that it shows up during step 7: As soon as I flip to a branch that is behind a base branch, the blue dot should appear in the history tab and flipping to the history tab, the alert should show up without having to play around with the comparison text box.

### Actual Behavior
I get two different ways that the alert is triggered.  The first time it is triggered, which calls for clicking the comparison box and flipping back and forth on the changes and history tab, and every time after the first, which is it automatically coming up as I think it should.

Sorry if this is vague or unclear, it is hard to describe.  I will try to explain better if it is still not understood.

## Additional Information
![branch bug](https://user-images.githubusercontent.com/19867440/43438023-2183e5c4-945a-11e8-98d2-08cf44b77a01.gif)

billygriffin:
At the time, it seems like we thought this was relatively minor, but my read after further consideration is that this inconsistency is actually preventing people from being informed about this state the vast majority of the time. I wrote up some thoughts in #6875, but this feels like something worth buttoning up before we iterate on how to surface this information since it seems like it never worked as intended.

rafeca:
Closing since this has been already fixed (the fix will be available on the next release!)

PR #9319, opened 2020-03-19
Fetch the aheadBehind state when the info doesn't exist

rafeca:
Partially addresses #5293

## Description

This PR partially addresses https://github.com/desktop/desktop/issues/6875 by fixing the point 1. stated in https://github.com/desktop/desktop/issues/6875#issuecomment-600621130.

There are still 2 pending issues with the prompt that I'm planning to address as separate PRs:

1. The prompt sometimes disappears just after it gets displayed. This is related to the point 2. in https://github.com/desktop/desktop/issues/6875#issuecomment-600621130.
2. The base branch used for the comparison is not consistent: most of the times `upstream/master` is used, but sometimes it defaults to just `master`.

### Screenshots

Before. Prompt only shown after opening the "Select branch..." dropdown:

before

Now. Prompt shown immediately:

demo

## Release notes

Notes: [Fixed] Display more consistently the prompt to notify when the current branch is outdated.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1042,6 +1042,9 @@ export class AppStore extends TypedBaseStore<IAppState> {
           inferredBranch.tip.sha
         )
 
+        // In the case that the aheadBehindCache doesn't have the needed data, try to
+        // request it directly from AheadBehindUpdater. This usually happens on initial load
+        // before AheadBehindUpdater has run though all the branches.
         if (
           aheadBehindOfInferredBranch === null &&
           this.currentAheadBehindUpdater
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1036,24 +1036,11 @@ export class AppStore extends TypedBaseStore<IAppState> {
         compareState.aheadBehindCache
       )
 
-      if (inferredBranch !== null) {
-        aheadBehindOfInferredBranch = compareState.aheadBehindCache.get(
+      if (inferredBranch !== null && this.currentAheadBehindUpdater !== null) {
+        aheadBehindOfInferredBranch = await this.currentAheadBehindUpdater.executeAsyncTask(
           tip.branch.tip.sha,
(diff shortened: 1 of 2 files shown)


billygriffin:
Thanks for looking into this @rafeca! I added @ampinsk as a reviewer as I think we need to discuss whether we actually want to fix this so it works as it was originally intended or go a different direction.

rafeca:
> Thanks for looking into this @rafeca! I added @ampinsk as a reviewer as I think we need to discuss whether we actually want to fix this so it works as it was originally intended or go a different direction.

Thanks for pointing this out! I've talked to @ampinsk and she mentioned that for now it's worth fixing the current behaviour of the NDDB and learn from its usage before going deeper into a different direction.

app/src/lib/stores/app-store.ts
```diff
@@ -1041,6 +1041,19 @@ export class AppStore extends TypedBaseStore<IAppState> {
           tip.branch.tip.sha,
           inferredBranch.tip.sha
         )
+
+        // In the case that the aheadBehindCache doesn't have the needed data, try to
+        // request it directly from AheadBehindUpdater. This usually happens on initial load
+        // before AheadBehindUpdater has run though all the branches.
+        if (
+          aheadBehindOfInferredBranch === null &&
```
outofambit: I think we can remove this guard (and the lines 1040-1043 above) since if we're checking `this.comparisonCache` in `executeTaskAsync`, that should return the same result as would be found in the `compareState.aheadBehindCache` in this scope. i think that would remove an extra cache check and simplify the logic path here.

(the `aheadBehindUpdater` emits updates whenever it adds to its cache, which then gets propagated to the repository state's compare state)

rafeca: Addressed in https://github.com/desktop/desktop/pull/9319/commits/f150e1fa067e930ca2a3c1c5b30e70f841e3b5a2 !

rafeca:
Thanks everyone for the reviews! We're going to wait to merge this one until the v2.4 release is done since this should go to v2.5

PR #9320, opened 2020-03-19
More consistent logic for showing/hiding the NBBD prompt

rafeca:
Partially addresses #5293

This branch is based on https://github.com/desktop/desktop/pull/9319

## Description

This PR partially addresses #6875 by fixing the point 2. stated in https://github.com/desktop/desktop/issues/6875#issuecomment-600621130 .

In order to fix this, I've decided to add a third flag that controls whether the user has currently dismissed the banner (this way if the user dismisses the banner, it doesn't get shown again when the diverging logic gets executed afterwards).

In terms of behaviour, I've tried to follow what was stated in [the original NBBD issue](https://github.com/desktop/desktop/issues/4646). This is a quick summary:

* The blue nudge is shown whenever when the number of commits behind is different from the last time Desktop checked, and it gets hidden when the user clicks on the history tab.
* The prompt is shown whenever the current branch is behind the inferred branch. It stays visible until the user dismisses the prompt by clicking on the "x" button.

Something important to note is that this whole logic gets reset every time the user changes branches (or the app gets restarted). This means that if the user checks out a branch and goes back to the current branch, they'll see the blue nudge and the prompt again even if they have dismissed the prompt.

This is caused by the way the NBBD flags are stored, we could probably persist them in a different place to keep the state for longer.

cc/ @billygriffin, @nerdneha , @ampinsk 

### Screenshots

Before:

before

After:

after-2

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -643,6 +643,9 @@ export interface ICompareState {
   /** Show the diverging notification banner */
   readonly isDivergingBranchBannerVisible: boolean
 
+  /** Has the user dismissed the notification banner? */
+  readonly isDivergingBranchBannerDismissed: boolean
+
   /** Show the diverging notification nudge on the tab */
   readonly isDivergingBranchNudgeVisible: boolean
 
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -643,6 +643,9 @@ export interface ICompareState {
   /** Show the diverging notification banner */
   readonly isDivergingBranchBannerVisible: boolean
 
+  /** Show the diverging notification nudge on the tab */
+  readonly isDivergingBranchNudgeVisible: boolean
+
   /** The current state of the compare form, based on user input */
   readonly formState: IDisplayHistory | ICompareBranch
 
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -640,14 +640,8 @@ export interface ICompareBranch {
 }
 
 export interface ICompareState {
-  /** Show the diverging notification banner */
-  readonly isDivergingBranchBannerVisible: boolean
-
-  /** Has the user dismissed the notification banner? */
-  readonly isDivergingBranchBannerDismissed: boolean
-
-  /** Show the diverging notification nudge on the tab */
-  readonly isDivergingBranchNudgeVisible: boolean
+  /** The current state of the NBBD banner */
+  readonly divergingBranchBannerState: IDivergingBranchBannerState
 
   /** The current state of the compare form, based on user input */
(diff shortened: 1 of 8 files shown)


app/src/lib/app-state.ts
```diff
@@ -695,6 +695,17 @@ export interface ICompareState {
...
+export interface IDivergingBranchBannerState {
+  /** Show the diverging notification banner */
+  readonly isPromptVisible: boolean
+
+  /** Has the user dismissed the notification banner? */
+  readonly isPromptDismissed: boolean
+
+  /** Show the diverging notification nudge on the tab */
+  readonly isNudgeVisible: boolean
+}
```
rafeca: I've decided two put these 3 flags under the same object rather than having 3 separate flags starting with `divergingBranchBannerXXXX`. This also makes it slightly easier to update the state from the `app-store`.

app/src/lib/stores/app-store.ts
```diff
@@ -4617,24 +4638,39 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
   }
 
-  public _setDivergingBranchBannerVisibility(
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public _updateDivergingBranchBannerState(
```
rafeca: This method now can update any of the 3 NBBD flags by passing an object with the flags to update. This reduces some repetition and makes sure that only 1 update is emitted when changing multiple flags at once.

The main tradeoff is that the method is a bit more "dynamic" in the sense that it accepts a `Partial` of an interface as an argument (still, everything is still correctly covered by TypeScript).

rafeca:
I just changed the base branch of this PR to be `development`, since #9319 has already been merged

rafeca:
@outofambit I checked with @ampinsk and the PR looks good to her! ✨

app/src/lib/stores/app-store.ts
```diff
@@ -4604,24 +4624,39 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          : currentBannerState.isPromptVisible,
+      isNudgeVisible:
+        divergingBranchBannerState.isNudgeVisible !== undefined
+          ? divergingBranchBannerState.isNudgeVisible
+          : currentBannerState.isNudgeVisible,
+      isPromptDismissed:
+        divergingBranchBannerState.isPromptDismissed !== undefined
+          ? divergingBranchBannerState.isPromptDismissed
+          : currentBannerState.isPromptDismissed,
+    }
```
outofambit, line 4648: @rafeca we talked about this already, just recording this here.

I think we can replace this block with a couple object spreads— 
```suggestion
    const newBannerState = { ...currentBannerState, ...divergingBranchBannerState }
```

PR #9332, opened 2020-03-21
Add support for Android Studio on macOS

guidezpl:
## Description

- Add support for Android Studio on macOS

## Release notes

Notes:
- Add support for Android Studio on macOS

### DIFF
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -351,7 +351,10 @@ export async function getAvailableEditors(): Promise<
   }
 
   if (androidStudioPath) {
-    results.push({ editor: ExternalEditor.AndroidStudio, path: androidStudioPath })
+    results.push({
+      editor: ExternalEditor.AndroidStudio,
+      path: androidStudioPath,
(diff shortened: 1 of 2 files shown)


rafeca:
Hey @guidezpl, Thanks for your contribution! Adding Android Studio to the list of supported editors will be useful for many users.

The OSX part looks good codewise, but it would be good to also add support for Android Studio on Windows as part of this PR (we like to keep the list of editors consistent between operating systems when possible).

Adding an editor on Windows requires a bit more work since it requires reading stuff from the registry (see `app/src/lib/editors/win32.ts`), but if you have access to a Windows VM should be easy to do.

guidezpl:
I don't have access to a Windows VM or machine so I won't be able to work on it

rafeca:
> I don't have access to a Windows VM or machine so I won't be able to work on it

Understandable, I think it's still okay to merge this PR: it's better to offer an improvement only to OSX users than no users at all. I'm going to create an issue to add Windows support for Android Studio so maybe somebody from the community does it.

PR #8965, opened 2020-01-23
[Fork] PR list call to action

outofambit:
this is a revisiting of #8936 and #8937 

## Description

- adds a version of the PR list that includes a link to view the current PR online if its in an upstream repo
- the hope is that if someone sees the PR badge on the branch dropdown and they open the PR list looking for their PR, they'll see this link as a path forward
- its also worth nothing this is likely a temporary design while we work towards redesigning the PR list (cc @ampinsk)

### Screenshots

Screen Shot 2020-01-23 at 1 12 24 PM

## Release notes

Notes: [Added] Link in Pull Request list to view the current pull request on GitHub

## Testing notes

1. Select a fork in desktop
2. Open a pull request from your current branch that targets the upstream repository (aka make a PR from a branch in `outofambit/desktop` to be merged into `desktop/desktop`)
3. See pr badge in branch dropdown button
4. Open branch list to PR tab and see new message

PR #9138, opened 2020-02-19
Open a pull request from a GitHub.com PR page in your fork in Desktop

kuychaco:
Closes #9012 

## Description

This PR introduces behavior supportive of workflows where contributions are made from forks.

Previously, clicking the "open this in GitHub Desktop" link for a **PR made from a fork** on github.com would open the upstream repo, cloning if necessary. In order to make contributions locally from your fork, you would have to open your fork of the repo manually in the app.

With the changes introduced here, Desktop opens a user's fork repo when it makes sense to. We check the source of the PR, and if it is in Desktop (either as a fork or upstream) then we open the source repo and checkout the branch corresponding to the PR. If the source is not in Desktop, but the user has a fork in Desktop (ex: the source is another fork that doesn't belong to the user), we open the user's fork and checkout a branch corresponding to the PR. And in cases other than this we fallback to the current behavior of opening the upstream repo, cloning if needed. 

/ cc @outofambit @billygriffin @tierninho 

### Behavior and test cases

Clone the following repos and open them in Desktop accordingly based on the setup instructions for each scenario:
**Fork:** https://github.com/kuychaco/urban-potato
**Upstream:** https://github.com/not-kuychaco/urban-potato

Note: start out each scenario with a repo focused in Desktop that has nothing to do with the fork or upstream being tested (to test that Desktop correctly switches to the correct repo). 

**Only fork in Desktop, no upstream:**
Note: _The fork opened in Desktop should be the PR source (ex: for https://github.com/not-kuychaco/urban-potato/pull/2, https://github.com/kuychaco/urban-potato is the PR source so clone and open that fork in Desktop)._ 
Before testing each of the following scenarios, make sure master branch is checked out in fork repo, and non-fork repo is the focused repo in Desktop (this tests that Desktop correctly switches to the fork repo, and correctly switches from the default branch to the source branch)
* PR in upstream with fork as source —> opens fork in Desktop and source branch is checked out (https://github.com/not-kuychaco/urban-potato/pull/2) 
* PR in upstream with upstream as source —> opens fork in Desktop and source branch is checked out (https://github.com/not-kuychaco/urban-potato/pull/3)
* PR in upstream with someone else’s fork as source —> opens fork in Desktop and `pr/#` branch is checked out  (https://github.com/not-kuychaco/urban-potato/pull/5)

**Fork AND upstream in Desktop:**
[...]

PR #9340, opened 2020-03-23
Display PRs from upstream repository in forks

### LINKED ISSUE #6383: display PRs from upstream repository in your fork
I am the maintainer of Zebobo5/Vlc.DotNet. I mostly work on my fork, and make PR to integrate changes to the main repository.

To avoid pushing inadvertently bad code into Zebobo5's project, I cloned my fork (which is `origin`), and added an `upstream` remote to track changes to the main project (`git pull upstream develop`)

I've always done everything from CLI, but wanted to give GitHub desktop a try, to see if it could improve my workflow. It seems that GitHub desktop doesn't understand this workflow, and doesn't let me checkout the PR from the main project.

My question is: Is there a way to checkout a specific PR from the main project while on a fork?

Ideas:
- Show PR from this project + from the original fork in a separate category
- Show PR from all remotes configured

rafeca:
Closes #6383

## Description

This PR changes the filtering logic of the PRs list to display the pull requests made on the upstream repository when the current repository is a fork.

As part of this PR, I've removed the logic that adds a link to the list of PRs on the upstream repository  when there are no PRs (which was added in https://github.com/desktop/desktop/pull/8965). Now, the logic is not needed since the shown PRs are the ones in the upstream repository already.

Instead of removing that logic, we can also change it to show a link to the PRs page of the forked repository, but I feel that this would create more confusion than help users (since the most normal use case is to have PRs on the upstream repo).

### Screenshots

Screenshot 2020-03-23 at 14 05 47

### Test plan

On a forked repository:

- The PRs from the upstream repository should be shown.
- An empty screen should be shown when there are no PRs on the upstream repository.
- All repository names on the PRs screens should mention the upstream repository.

On a non-forked repository:

- The PRs from the origin repository should be shown.
- An empty screen should be shown when there are no PRs on the origin repository.
- All repository names on the PRs screens should mention the origin repository.

## Release notes

Notes: [Improved] Show the list of pull requests from the upstream repository on forks

### DIFF
--- a/app/src/lib/stores/pull-request-coordinator.ts
+++ b/app/src/lib/stores/pull-request-coordinator.ts
@@ -149,7 +149,7 @@ export class PullRequestCoordinator {
       gitHubRepository,
       this.repositories
     )
-    // mark all matching repos for parent as now loading
+    // mark all matching repos as now loading
     for (const match of matches) {
       this.emitIsLoadingPullRequests(match, true)
     }
--- a/app/src/lib/stores/pull-request-coordinator.ts
+++ b/app/src/lib/stores/pull-request-coordinator.ts
@@ -154,7 +154,6 @@ export class PullRequestCoordinator {
       this.emitIsLoadingPullRequests(match, true)
     }
 
-    // mark all matching repos as now loading
     await this.pullRequestStore.refreshPullRequests(gitHubRepository, account)
 
     // mark all matching repos as done loading
--- a/app/src/lib/stores/pull-request-coordinator.ts
+++ b/app/src/lib/stores/pull-request-coordinator.ts
@@ -3,6 +3,7 @@ import { PullRequest } from '../../models/pull-request'
 import {
   RepositoryWithGitHubRepository,
   isRepositoryWithGitHubRepository,
+  getNonForkGitHubRepository,
 } from '../../models/repository'
 import { PullRequestStore } from '.'
 import { PullRequestUpdater } from './helpers/pull-request-updater'
@@ -86,28 +87,14 @@ export class PullRequestCoordinator {
         }
 
         // find all related repos
-        const { matches, forks } = findRepositoriesForGitHubRepository(
+        const matches = findRepositoriesForGitHubRepository(
           ghRepo,
           this.repositories
         )
 
-        // emit updates for forks
-        for (const fork of forks) {
-          this.getPullRequestsFor(fork.gitHubRepository).then(prs =>
-            fn(fork, [...prs, ...pullRequests])
-          )
-        }
-
         // emit updates for matches
         for (const match of matches) {
-          if (match.gitHubRepository.parent) {
-            // matches that are forks need to include the PRs from upstreams
-            this.getPullRequestsFor(match.gitHubRepository.parent).then(prs =>
-              fn(match, [...prs, ...pullRequests])
-            )
-          } else {
-            fn(match, pullRequests)
-          }
+          fn(match, pullRequests)
         }
       }
     )
@@ -155,9 +142,11 @@ export class PullRequestCoordinator {
     repository: RepositoryWithGitHubRepository,
     account: Account
   ) {
+    const gitHubRepository = getNonForkGitHubRepository(repository)
+
     // get all matches for the repository to be refreshed
-    const { matches } = findRepositoriesForGitHubRepository(
-      repository.gitHubRepository,
+    const matches = findRepositoriesForGitHubRepository(
+      gitHubRepository,
       this.repositories
     )
     // mark all matching repos for parent as now loading
@@ -166,31 +155,7 @@ export class PullRequestCoordinator {
     }
 
     // mark all matching repos as now loading
-    await this.pullRequestStore.refreshPullRequests(
-      repository.gitHubRepository,
-      account
-    )
-    if (repository.gitHubRepository.parent !== null) {
-      // get all matches for the parent repository
-      const { matches: parentMatches } = findRepositoriesForGitHubRepository(
-        repository.gitHubRepository.parent,
-        this.repositories
-      )
-      // mark all matching repos for parent as now loading
-      for (const parentMatch of parentMatches) {
-        this.emitIsLoadingPullRequests(parentMatch, true)
-      }
-
-      await this.pullRequestStore.refreshPullRequests(
-        repository.gitHubRepository.parent,
-        account
-      )
-
-      // mark all matching repos for parent as done loading
-      for (const parentMatch of parentMatches) {
-        this.emitIsLoadingPullRequests(parentMatch, false)
-      }
-    }
+    await this.pullRequestStore.refreshPullRequests(gitHubRepository, account)
 
     // mark all matching repos as done loading
     for (const match of matches) {
@@ -204,24 +169,16 @@ export class PullRequestCoordinator {
    *
    * Since `PullRequestStore` stores these timestamps by
    * `GitHubRepository`, we get timestamps for this
-   * repo's `GitHubRepository` and its parent (if it has one)
-   * and return the _older one._
+   * repo's `GitHubRepository` or its parent (if it has one).
    *
-   * If neither timestamp is stored, returns `undefined`
+   * If no timestamp is stored, returns `undefined`
    */
   public getLastRefreshed(
     repository: RepositoryWithGitHubRepository
   ): number | undefined {
-    const ghr = repository.gitHubRepository
-    const lastRefresh = this.pullRequestStore.getLastRefreshed(ghr)
+    const ghr = getNonForkGitHubRepository(repository)
 
-    const parentLastRefresh = ghr.parent
-      ? this.pullRequestStore.getLastRefreshed(ghr.parent)
-      : undefined
-
-    return !lastRefresh || !parentLastRefresh
-      ? lastRefresh || parentLastRefresh
-      : Math.min(lastRefresh, parentLastRefresh)
+    return this.pullRequestStore.getLastRefreshed(ghr)
   }
 
   /**
@@ -231,15 +188,7 @@ export class PullRequestCoordinator {
   public async getAllPullRequests(
     repository: RepositoryWithGitHubRepository
   ): Promise<ReadonlyArray<PullRequest>> {
-    if (repository.gitHubRepository.parent !== null) {
-      const [prs, upstreamPrs] = await Promise.all([
-        this.getPullRequestsFor(repository.gitHubRepository),
-        this.getPullRequestsFor(repository.gitHubRepository.parent),
-      ])
-      return [...prs, ...upstreamPrs]
-    } else {
-      return await this.getPullRequestsFor(repository.gitHubRepository)
-    }
+    return this.getPullRequestsFor(getNonForkGitHubRepository(repository))
   }
 
   /** Start background pull request fetching machinery for this Repository */
@@ -307,23 +256,21 @@ export class PullRequestCoordinator {
 /**
  * Finds local repositories related to a GitHubRepository
  *
- * * Related repos include:
- *  * **matches** — the corresponding GitHub repo (the `origin` remote for
- *    the `Repository`)
- *  * **forks** — the parent GitHub repo, if the `Repository` has one (the
+ * * Related repos include the corresponding GitHub repo (the `origin` remote for
+ *    the `Repository`) or the parent GitHub repo, if the `Repository` has one (the
  *    `upstream` remote for the `Repository`)
  *
  * @param gitHubRepository
  * @param repositories list of repositories to search for a match
- * @returns two lists of repositories: **matches** and **forks**
+ * @returns the list of repositories.
  */
 function findRepositoriesForGitHubRepository(
   gitHubRepository: GitHubRepository,
   repositories: ReadonlyArray<RepositoryWithGitHubRepository>
-) {
+): Array<RepositoryWithGitHubRepository> {
   const { dbID } = gitHubRepository
-  const matches = new Array<RepositoryWithGitHubRepository>(),
-    forks = new Array<RepositoryWithGitHubRepository>()
+  const matches = new Array<RepositoryWithGitHubRepository>()
+
   for (const r of repositories) {
     if (r.gitHubRepository.dbID === dbID) {
       matches.push(r)
@@ -331,9 +278,9 @@ function findRepositoriesForGitHubRepository(
       r.gitHubRepository.parent !== null &&
       r.gitHubRepository.parent.dbID === dbID
     ) {
-      forks.push(r)
+      matches.push(r)
     }
   }
 
-  return { matches, forks }
+  return matches
 }
--- a/app/src/lib/stores/pull-request-coordinator.ts
+++ b/app/src/lib/stores/pull-request-coordinator.ts
@@ -266,20 +266,10 @@ export class PullRequestCoordinator {
 function findRepositoriesForGitHubRepository(
   gitHubRepository: GitHubRepository,
   repositories: ReadonlyArray<RepositoryWithGitHubRepository>
-): Array<RepositoryWithGitHubRepository> {
+): ReadonlyArray<RepositoryWithGitHubRepository> {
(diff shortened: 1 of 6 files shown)


rafeca:
@ampinsk any feedback on the sizing of the PR list header?

Screenshot 2020-03-23 at 14 12 16

A couple of considerations:
- It currently uses the same font size/margins as the branch list headers:
    Screenshot 2020-03-23 at 14 14 07
- Because of the scrolling logic of filtered lists, the header must have the same height as each of the items on the list (not sure if this can be changed though).

Things I think we could do:
- Make the header font slightly larger.
- Move the header text a bit down (keeping the height of the header), so the space between the header and the first item gets reduced.

ampinsk:
> @ampinsk any feedback on the sizing of the PR list header?

Thanks for the ping @rafeca! I actually think this looks totally fine as is 👍 I think the spacing makes sense in context, and if the header was any bigger it'd probably draw too much attention.

rafeca:
> The functionality works well, but I spotted a bug. Anytime you create/close a PR from the forked repo, the PR list in Desktop toggles between these two views:

Thanks for reporting!! I've been able to reproduce and looking into it!

> Also, just to be clear, for any forked repo on Desktop, we are only going to surface PRs from the upstream repo? No PRs will be shown in only on the forked repo on Github.com?

Exactly. This makes me think about what happens when the user doesn't have the parent repo as an `upstream` remote on their working copy, gonna test that scenario!

rafeca:
> The functionality works well, but I spotted a bug. Anytime you create/close a PR from the forked repo, the PR list in Desktop toggles between these two views:

After some investigation, it looks like this was an existing issue, which before this PR used to happen in the opposite way: when a PR was created/closed/updated on the upstream repository, the list of PRs went blank.

I'm looking into the logic that updates the list of PRs to find where the issue comes from.

rafeca:
@tierninho I've fixed the issue that you found in https://github.com/desktop/desktop/pull/9340/commits/f8f54b5cf063985e5463a85b532b773d172ceb12 ! That was a really good catch!!

rafeca:
In 6760ddc0eb245019561575ed0617379bc80b207e I've changed a bit the logic to fetch PRs to only fetch the pull requests from the upstream repository when working on a fork.

I found that by fetching both the upstream and origin PRs but only showing the upstream ones some issues caused some issues on the logic that tries to find the correspondant PR of a branch (if for some reason the same branch was used for both a PR on the upstream and on the fork repo the Ui went a bit nuts).

Also, with this change we're fetching a bit less data.

app/src/ui/branches/branches-container.tsx
```diff
@@ -208,13 +203,32 @@ export class BranchesContainer extends React.Component<
...
 
+  private getRepositoryName(): string {
+    const gitHubRepository = this.getGitHubRepository()
+
+    if (gitHubRepository === null) {
+      return this.props.repository.name
+    }
+
+    return gitHubRepository.fullName
+  }
```
outofambit: @rafeca is the primary purpose of this to fallback to the plain repository name if there's no GitHubRepository for it? how does this differ from `nameOf`, otherwise?

tierninho:
> I've fixed the issue that you found in 32a4222 ! That was a really good catch!!

Bug fix looks good ✅ 

> tomorrow to make sure we don't regress anything tangentially related elsewhere in the app, but i think it is likely a great solution.

Let me know if there are any vulnerable areas that may need testing post the meeting, as I'd be happy to explore more.

tierninho:
One minor issue that may be present in a small subset of users. It looks like we center the text if the repo name is too long. Should we wrap it?

Almost too long:
Screen Shot 2020-03-24 at 12 22 21 PM

Too long:
Screen Shot 2020-03-24 at 12 21 25 PM

app/src/ui/branches/branches-container.tsx
rafeca: The main difference with `nameOf` is that this function tries to use the parent repository if it exists (it's unfortunate that I couldn't use `nameOf`).

But now that I think of this, this function is only used for the pull requests view. This means that I can it to receive a `RepositoryWithGitHubRepository` argument instead of using the `repository` from the props and then I won't need the fallback for non-github repositories.

rafeca:
> One minor issue that may be present in a small subset of users. It looks like we center the text if the repo name is too long. Should we wrap or truncate it?

Good point! gonna check how do we do it in other places and follow the same pattern

app/src/ui/branches/branches-container.tsx
rafeca: I've ended up changing the strategy a little bit: Instead of start using a `repository` argument on each method that has the more specific type than `this.props.repository` (this felt like a code smell), I've ended up passing the `repository` object to the `<PullRequestList/>` component and moved all the pullrequest-specific logic to that component so it can freely use `this.props.repository` knowing that the object is a `RepositoryWithGitHubRepository`.

This has the added benefit of compartmentalizing a little bit more the logic (logic specific to the `<PullRequestList/>` list doesn't leak to `<BranchContainer/>`). We can do the same for `<BranchList/>` in a separate PR.

This has been done in e2597febe5b0116e9e58ddd692dbeb3285f3ea4a. If you think that the commit is out of the scope of this PR I can move it to a separate one.

rafeca:
> One minor issue that may be present in a small subset of users. It looks like we center the text if the repo name is too long. Should we wrap or truncate it?

I've seen that usually we truncate branch/repository names with ellipsis so I did the same in the latest commit:

Screenshot 2020-03-25 at 14 42 25

rafeca:
> Let me know if there are any vulnerable areas that may need testing post the meeting, as I'd be happy to explore more.

I've synced with @outofambit and she mentioned two areas of potential issues:

1. The integration with the "open this in GitHub Desktop" link on dotcom for PRs.
2. The logic that prompts the user to create a PR on the "No local changes screen".

I've been able to find one issue (actually it's a discrepancy) between the behaviour of "open this in GitHub Desktop" button and clicking on a PR from the PRs list in desktop:

When opening a PR created to the upstream repository from a branch that also lives on the upstream repository, the behaviour is different between the two:

- The dotcom integration creates a local branch with the same name as the original PR branch that's tracking the original branch on the `upstream` remote.
- The PR list functionality creates a local branch called `pr/<num_pr>` that tracks the original branch in a newly created remote that points to upstream.

So after using the "open this in GitHub Desktop" button and clicking on the same PR from the PRs list on Desktop, we end up with the following local branches on `git`:

```sh
$ git branch -vv
  pr/1             774a131 [github-desktop-rafeca/rafeca-patch-1] assaas
  rafeca-patch-1   774a131 [upstream/rafeca-patch-1] assaas
```

```sh
$ git remote -v
github-desktop-rafeca	https://github.com/rafeca/test-forks.git (fetch)
github-desktop-rafeca	https://github.com/rafeca/test-forks.git (push)
origin	                https://github.com/rafeca-fork/test-forks.git (fetch)
origin	                https://github.com/rafeca-fork/test-forks.git (push)
upstream	        https://github.com/rafeca/test-forks.git (fetch)
upstream        	https://github.com/rafeca/test-forks.git (push)
```

Based on this, I think that the behaviour of the "open this in GitHub Desktop" integration is the correct one (we don't need to duplicate the upstream remote), so I'm going to try to change the logic of the PRs list to match it.

rafeca:
I've been able to fix the issue described in https://github.com/desktop/desktop/pull/9340#issuecomment-603935866 via 0a33f905d3ed692c28537bf153eea22fc3fb973c 

Now the behaviour between the dotcom integration and the PRs list is consistent.

Before:

before

After:

after

tierninho:
@rafeca For above, possible to share the reproduction steps? I ran into a fast-forward error (I _think_ its an existing issue) and I wanted to make sure I was testing the steps in the right way.

Screen Shot 2020-03-25 at 2 41 01 PM

Note there was some recent PR work done in this area, see here: https://github.com/desktop/desktop/pull/9138 I do not believe we have released it yet.

rafeca:
> @rafeca For above, possible to share the reproduction steps? I ran into a fast-forward error (I think its an existing issue) and I wanted to make sure I was testing the steps in the right way.

I've used the following setup with a repository and two forks, where there are 3 PRs, each of these based on a branch from each of the repositories:

excalidraw-2020326142448 (1)

Then I add one of the forked repositories to GitHub Desktop and do the following actions for each of the PRs:

1. Click on the "open this in GitHub Desktop" link from dotcom.
2. Click on the PR from the lists of PRs inside GitHub Desktop.

I expect both actions to do exactly the same, no matter which of these gets done first.

The expected result is to checkout the branch from either the origin or upstream repository (in the case of `PR #1` and `PR #2`) or add the remote, fetch it and checkout the branch (in the case of `PR #3`). Does this makes sense to you?

>  I ran into a fast-forward error (I think its an existing issue) and I wanted to make sure I was testing the steps in the right way.

Do you remember under which conditions this happened? I haven't run into this yet

tierninho:
> Do you remember under which conditions this happened? I haven't run into this yet

Yes, this occurred when I clicked the `Open in Desktop` link from a PR in the forked repo. The PR wanted to merge commits into master (fork) from branchX (upstream). See here: https://github.com/kuychaco/urban-potato/pull/3. Upon clicking the link, Desktop opened the PR in the forked repo, and the error was surfaced. (Make sure you are on the master branch in Desktop before clicking the link.)

Based on the steps you outlined above, I will recreate the entire setup from scratch and retest.

tierninho:
> I expect both actions to do exactly the same, no matter which of these gets done first.

Confirming the steps you outlined above result in the expected actions. The "open this in GitHub Desktop" link and the clicking on the PR from inside GitHub Desktop led to the same result. ✅

rafeca:
> Yes, this occurred when I clicked the Open in Desktop link from a PR in the forked repo. The PR wanted to merge commits into master (fork) from branchX (upstream). See here: kuychaco/urban-potato#3. Upon clicking the link, Desktop opened the PR in the forked repo, and the error was surfaced. Note: The branch was not published. Also, make sure you are on the master branch in Desktop before clicking the link.)

I think I know what happened here: local branches for PRs that are started on forks are named `pr/<number_pr>` (e.g `pr/4`). If there are PRs on both the fork and the upstream repository and you try to checkout the same PR number from both repositories, the second one will give an error.

This happens with https://github.com/kuychaco/urban-potato/pull/3 and https://github.com/not-kuychaco/urban-potato/pull/3 . If you try to checkout both of these PRs from dotcom you'll probably get an error for the second one (I guess this even happens in production).

Right now this kind of problems are a little bit hard to fix, since the logic to checkout a PR is duplicated ([the logic for checking out a PR internally](https://github.com/desktop/desktop/blob/development/app/src/lib/stores/app-store.ts#L5268) is not the same as the one used [by the interceptor in dotcom](https://github.com/desktop/desktop/blob/development/app/src/ui/dispatcher/dispatcher.ts#L1430)).

I can try to unify this logic so it's going to be easier to make fixes (plus the functionality will be uniform between both).

app/src/lib/stores/app-store.ts
```diff
@@ -5297,6 +5304,26 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          const remoteUpstream = findUpstreamRemote(gitHubRepository, remotes)
+
+          if (remoteUpstream) {
+            await this._fetchRemote(
+              repository,
+              remoteUpstream,
+              FetchType.UserInitiatedTask
+            )
+          }
+        }
```
outofambit, line 5326: i have some questions about this flow, but it looks like @rafeca and @tierninho have already been discussing this. can you share the current understanding here? it sounds like the fork case is broken here, but also in production? is that right?

would it make sense to file a separate bug for this problem and close it with  #9382? (I'm having a little trouble following here distinguishing the bug and which environments its present in, and therefore whether it's a regression or not)

app/src/lib/stores/pull-request-coordinator.ts
```diff
@@ -300,33 +256,31 @@ export class PullRequestCoordinator {
...
       r.gitHubRepository.parent !== null &&
       r.gitHubRepository.parent.dbID === dbID
     ) {
-      forks.push(r)
+      matches.push(r)
     }
   }
 
-  return { matches, forks }
+  return matches
```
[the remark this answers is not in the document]
rafeca: Oh yes! good point!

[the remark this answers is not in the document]
rafeca: I've thought a bit about this and indeed I think that the current code will cause issues with grandforks, but I think that it should be changed to:

```js
return repositories.filter(
  repository => getNonForkGitHubRepository(repository).dbID === dbID
)
```

The reasoning is that this method is used to map ` GitHubRepository` object to any `Repository`  

used to calculate which `Repository` objects need to be updated after receiving an updated list of PRs. Since we fetch the list of PRs of a `Repository` by getting its `GitHubRepository` via `getNonForkGitHubRepository()`, we need to use the same method when trying to map back a `GitHubRepository` to the potential list of `Repository` objects. Does this make sense?

----

However, this raises a different question about which PRs to show when opening a grandfork. Currently if we have a `repo`, a `fork` and a `grandfork`, when opening `grandfork` on Desktop the PRs from `fork` will be shown.

I feel that it would make more sense to show the PRs from `repo`, but I don't have strong opinions about it. @billygriffin do you have any thoughts about this? (if we decide to change this we can do it in a separate PR).

app/src/lib/stores/app-store.ts
rafeca: I made this change in 0a33f905d3ed692c28537bf153eea22fc3fb973c to fix the issue described in https://github.com/desktop/desktop/pull/9340#issuecomment-603935866.

Since the issue that it fixes is not important (and is reproducible on production) I'm gonna revert this commit from this PR and move it to https://github.com/desktop/desktop/pull/9382, where it makes much more sense. This will also make this PR a bit smaller and focused on actually displaying PRs from upstream repos.

👍to create an issue for https://github.com/desktop/desktop/pull/9382, since I just realized that this whole thing got a bit out of control 😅

rafeca:
I've just moved 6b1d68dbec6ca467ebee3d6c1ca31e1b1d248ef0 to https://github.com/desktop/desktop/pull/9382 .

This PR now is only focused on displaying the PRs from upstream repositories when working on forks.

Anything to do with the consistency between checking out a PR in dotcom and in Desktop is fixed by https://github.com/desktop/desktop/pull/9382.

app/src/lib/stores/pull-request-coordinator.ts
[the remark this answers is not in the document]
outofambit: > used to calculate which `Repository` objects need to be updated after receiving an updated list of PRs. Since we fetch the list of PRs of a `Repository` by getting its `GitHubRepository` via `getNonForkGitHubRepository()`, we need to use the same method when trying to map back a `GitHubRepository` to the potential list of `Repository` objects. Does this make sense?

@rafeca oh yes, great point! i agree that this is the right solution for this method
> ```js 
> return repositories.filter(
>   repository => getNonForkGitHubRepository(repository).dbID === dbID
> )
> ```

thus far, we've explicitly tried to avoid allowing any repo to cause parent traversals more than one level (ie no repo should never have info/data from a repo that is a grandparent or higher), but also i don't think grandforks are a common enough use case to worry about beyond "will the app crash?". this logic maintains that line which i think is enough for now, but happy to reconsider more if you or @billygriffin have more thoughts!

app/src/lib/stores/app-store.ts
outofambit: awesome, thank you @rafeca!

billygriffin:
Thanks for the questions @rafeca! I don't think my comment here changes anything on this PR, but I just wanted to summarize the conversation I had with @outofambit about this for posterity in case we're revisiting this in the future or trying to understand it better.

Our assumption is that if you're on a fork, **we only care about the parent of the fork** - so we only go one level up, and never traverse through multiple forks. If your fork's parent repo also happens to be a fork, that's irrelevant to how things work for you in Desktop.

rafeca:
Thanks everyone for your comments! Gonna merge this one! 😃

PR #9343, opened 2020-03-23
Fix scroll position of history tab when switching repositories

### LINKED ISSUE #9341: Blank space in repository history
Sometimes, the latest version (2.3.2-beta7) on Windows 10 x64 gives the following problem:

sshot-2

The blank space cannot be clicked. After I restart the app, it's back to normal:

sshot-3

[2020-03-23.desktop.beta.log](https://github.com/desktop/desktop/files/4369618/2020-03-23.desktop.beta.log)

Experienced with multiple repositories, not very often, no exact steps to reproduce.

rafeca:
Closes #9341

## Description

This PR fixes the problem described in #9341 by resetting the scroll positions of the changes tab and the history tab when changing repositories.

The root cause of the issue was that the scroll position of the tabs is kept as part of the state of the `RepositoryView` component.

Since the component `RepositoryView` doesn't get unmounted when switching repositories, the scroll position from a different repository was maintained when switching repositories. This could cause issues when e.g switching from a repository where either list of changes or commits was very large to another repository with very little changes or commits (since the persisted scroll position could get out of range on the new repository).

### Screenshots

Video of the issue:

demo

## Release notes

Notes: [Fixed] Restored missing commits on the History tab when switching repositories.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -2510,6 +2510,9 @@ export class App extends React.Component<IAppProps, IAppState> {
 
       return (
         <RepositoryView
+          // When switching repositories we want to remount the RepositoryView
+          // component to reset the scroll positions.
+          key={selectedState.repository.hash}
           repository={selectedState.repository}
           state={selectedState.state}
           dispatcher={this.props.dispatcher}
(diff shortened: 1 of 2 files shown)


rafeca:
> Hey @rafeca this looks reasonable, but i'm curious if using a key for the RepositoryView component would be better for this case? We only use it in one place and I think it would be ok (or better?) to reset all the state in that component when the repository changes?

Oh yeah, that's much better!! thanks! Hadn't thought about using `key` for this, that's a very cool tip!

PR #9344, opened 2020-03-23
Allow shrinking for resizable lists with a footer

outofambit:
Fixes a regression due to our electron upgrade where vertical lists in vertically resizable containers would no longer shrink after growing.

### Before fix

broken-shrinking

### After fix

branches-fix
compare-fix


outofambit:
i'm going to merge so we can beta release it. its low risk.

rafeca:
Nice!!!

PR #9366, opened 2020-03-26
Add indicator for unpushed commits

### LINKED ISSUE #5873: Differentiating commits in history between local-only and remote
Currently the history view displays local commits the same as synced commits, minus 'View on Github' being greyed out. This isn't ideal as you can't tell at a glance what is local and needs to be pushed, and the 'undo' feature isn't accessible in the history view.

I'd suggest a few changes to improve how local commits are shown in the history tab:

1. Add a 'This commit has not been pushed' indicator somewhere, visually similar to the uncommitted changes indicator on the repository list. Possible implementations could be before the commit change message, at the end of it, or fixed in the upper right corner.

2. If the commit is local, and at the top of the commits list, swap 'Revert This Commit' with 'Undo This Commit' so that the undo feature is available on the history view. This is nicer than undoing commits from the changes tab where you don't know how many are available to undo. I can't think of a reason you'd want to revert the top-level local commit rather than just undoing it.

3. [Optional] This one is more work, but being able to edit the commit message of a local commit would be nice (I imagine under the covers that would be a `git rebase` which uses the reword option). I've definitely had cases where I have several local commits and I spot a typo in a commit message for one further back in the list. Using `git rebase` with the reword option is a simple solution, but it seems like GitHub Desktop could provide this functionality natively.

rafeca:
Closes #5873

## Description

This PR adds a small indicator for commits that haven't been pushed to any remote repository (local commits).

Some considerations:

- The indicator will be shown also for repositories with no upstream, meaning that all their commits will have the indicator. I think we should not show the indicator in that scenario, but don't have strong opinions on it (@billygriffin what do you think?).

### Screenshots

Normal mode:

Screenshot 2020-03-26 at 18 30 27
Screenshot 2020-03-26 at 18 30 33

Dark mode:

Screenshot 2020-03-26 at 18 30 55
Screenshot 2020-03-26 at 18 31 01

Long commit text:

Screenshot 2020-03-26 at 18 46 49

## Release notes

Notes: [Added] Add indicators to commits that haven't been pushed

### DIFF
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -21,7 +21,7 @@ interface ICommitProps {
   readonly onRevertCommit?: (commit: Commit) => void
   readonly onViewCommitOnGitHub?: (sha: string) => void
   readonly gitHubUsers: Map<string, IGitHubUser> | null
-  readonly isLocalRepository: boolean
+  readonly showUnpushedIndicator: boolean
 }
 
 interface ICommitListItemState {
@@ -91,7 +91,7 @@ export class CommitListItem extends React.Component<
   }
 
   private renderPushIndicator() {
-    if (this.props.isLocalRepository || !this.props.isLocal) {
+    if (!this.props.showUnpushedIndicator) {
       return null
     }
 
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -11,6 +11,7 @@ import { CommitAttribution } from '../lib/commit-attribution'
 import { IGitHubUser } from '../../lib/databases/github-user-database'
 import { AvatarStack } from '../lib/avatar-stack'
 import { IMenuItem } from '../../lib/menu-item'
+import { Octicon, OcticonSymbol } from '../octicons'
 
 interface ICommitProps {
   readonly gitHubRepository: GitHubRepository | null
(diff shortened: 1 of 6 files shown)


billygriffin:
@rafeca I agree that if there's no remote, it doesn't make a ton of sense to show this on every commit. I'm also interested in maybe doing a few usability tests when this gets out to beta to ensure people understand what this means. I'm a little concerned that it's not obvious that the indicator means it hasn't been pushed yet, but that might be unfounded and it is intuitive. @ampinsk thoughts on that?

ampinsk:
Also agree on the no remote scenario!

> I'm also interested in maybe doing a few usability tests when this gets out to beta to ensure people understand what this means. I'm a little concerned that it's not obvious that the indicator means it hasn't been pushed yet, but that might be unfounded and it is intuitive. @ampinsk thoughts on that?

Always happy to do usability tests! I hope it makes sense considering it borrows the same iconography from the repo indicators and the push/pull button. If it turns out to be super unclear to people I'm happy to rethink it as well. 

Also, we could add a tooltip similar to what we do for the repo indicators: 

Screen Shot 2020-03-26 at 12 02 27 PM

rafeca:
> Also, we could add a tooltip similar to what we do for the repo indicators:
>> I'm a little concerned that it's not obvious that the indicator means it hasn't been pushed yet.

I think a tooltip will help a lot here! I'll look into it

rafeca:
> Also, we could add a tooltip similar to what we do for the repo indicators:

Added in ced302a:

Screenshot 2020-03-27 at 09 36 27

rafeca:
> ⚠️ However, the indicator does not auto-update after the user clicked the Push origin button. Since the same indicator auto-updates on the Repository list, we should do the same here for consistency.

Nice catch!! I fixed this issue with the latest commits :)

app/src/ui/history/commit-list-item.tsx
```diff
@@ -79,12 +81,33 @@ export class CommitListItem extends React.Component<
...
       </div>
     )
   }
 
   public shouldComponentUpdate(nextProps: ICommitProps): boolean {
-    return this.props.commit.sha !== nextProps.commit.sha
+    return (
+      this.props.commit.sha !== nextProps.commit.sha ||
+      this.props.showUnpushedIndicator !== nextProps.showUnpushedIndicator
+    )
```
outofambit, line 93: i appreciate the restraint here :) or at least, i would have felt tempted to refactor this `shouldComponentUpdate` away, but i think its better not to in this PR

outofambit:
@tierninho has already tested this and I've approved it so I'm going to go ahead and this merge now! 🚀

PR #9383, opened 2020-03-27
Retreive and store git tags on the commit model

### LINKED ISSUE #9425: Add place to store tags in desktop
(not persistently, just in app state and caches)

we currently don't parse or store tags anywhere in desktop

_____
part of work for #9424

outofambit:
closes #9425 

## Description

- adds tags to `getCommits` git command
- adds parsing to extract that tag information
- adds field to commit model to store that information in
- adds unit tests to cover this codepath (including a new fixture git repo)
- display tags in commit summary UI (thanks @rafeca!)
  - no display on the commit list itself (that'll come in a later PR)
- feature flag for the UI portion of this

## Reviewing and Testing Notes

You can ignore all the added files in `app/test/fixtures`.

I think the only thing to test here is to make sure that all tags associated with a commit show up in the commit details UI.

Screen Shot 2020-04-02 at 3 20 03 PM


rafeca:
> It's possible that the tag information may be missing for some commits stored in state, because getBranches uses a git command that can't extract that information when it gets the tip of each branch. I've verified this is a problem and am looking into ways to address it.

I've created a new branch [`simplify-branch-tip`](https://github.com/desktop/desktop/tree/simplify-branch-tip) which tries to address this (relevant commit: 8339aa73e66f922c9f1a359634380394ad282293). 

What I've done is creating a new structure to store the branch tip which contains much less information than a commit (it only contains the sha and the commitIdentity, which is the required data to show on the branches UI). Good thing is that this allows us to request less data in the `getBranches()` method, which I guess will mean faster load times with repos with lots of branches.

I've done this in a separate branch because I wanted to get your opinion on this approach first

rafeca:
👋I've added 350dea1 to this branch which shows the list of tags on the commit view:

Screenshot 2020-03-31 at 16 32 38

(haven't added them yet on the history tab, since afaik the designs are not final)

app/test/unit/create-branch-test.ts
```diff
@@ -20,6 +20,7 @@ const stubTip = {
   authoredByCommitter: true,
   parentSHAs: [],
   trailers: [],
+  tags: null,
```
rafeca, line 23: Should this be an empty array?

```suggestion
  tags: [],
```

app/test/unit/git/log-test.ts
```diff
@@ -45,6 +45,18 @@ describe('git/log', () => {
...
+    it('parses tags', async () => {
+      const commits = await getCommits(repository, 'HEAD', 100)
+      expect(commits).toBeArrayOfSize(5)
+
+      expect(commits[0].tags).toIncludeSameMembers(['important'])
+      expect(commits[1].tags).toIncludeSameMembers([
+        'tentative',
+        'less-important',
+      ])
+      expect(commits[2].tags).toBeArrayOfSize(0)
```
rafeca, line 58: This kind of tests are super nice! 😍

app/test/unit/create-branch-test.ts
outofambit: ah good catch! this type is totally wrong since this is now an `IBranchTip` instead of a `Commit`. fixing!

outofambit:
> All tags retrieved/deleted in command line were successfully updated. All tags for Github.com were retrieved ok. I could not get deleted tags from Github.com to disappear from Desktop, but as the PR says "retrieve" I will add this to tests as the feature progresses.

@tierninho yes that's right!

> I take it this is WIP:
> Screen Shot 2020-04-02 at 2 34 03 PM

@tierninho oh that is not an intentional regression! fixing now

app/test/unit/create-branch-test.ts
outofambit: fixed in 4be73cf

outofambit:
@tierninho can you please verify the changed file list css is fixed now with 142a870?

app/test/unit/git/log-test.ts
me198930: **@[]()  **

rafeca:
> @tierninho oh that is not an intentional regression! fixing now

Ups sorry about that, I introduced this in https://github.com/desktop/desktop/pull/9383/commits/8988dce2423571423fbcdf67520e3adf4ecf6a25 🤡

PR #9412, opened 2020-03-30
Fix issue #7710: remove _hideStashedChanges invoke when switching sections

### LINKED ISSUE #7710: Multiple files selected after viewing History tab
## Description

If you click History and then Changes, it selects all your files. To reproduce:

1. Make some changes in a repo, in multiple files.
2. On the Changes tab, select one file to see the changes.
3. Click the History tab.
4. Click the Changes tab.

Now all the files are selected, instead of showing the file you were just looking at.

## Version

* GitHub Desktop: 2.0.3

* Operating system: macOS Mojava 10.14.5

TerryChan:
Because the _hideStashedChanges method will update the selectedFileIDs to all files, which causes all changed files seelcted when click Changes tab

Closes #7710

## Description
When we click History tab from Changes, the _hideStashedChanges  method will be invoked, and _hideStashedChanges  will refresh selectedFileIDs to all changed files, and set diff to null. Then when we click Changes back, all files will be selected, instead of the file we selected before, and the diff view doesn't show diff content.

This will also update the design below: 
when we click Changes tab, it always show the change list instead of the Stash, even the Stash was checked before. 
Now it will show Stash if the Stash was checked before. Maybe you need to verify whether it's ok.

### Screenshots
All changed files are selected when click Changes tab:
image

## Release notes

Notes:

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2178,8 +2178,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }))
     this.emitUpdate()
 
-    this._hideStashedChanges(repository)
-
     if (selectedSection === RepositorySectionTab.History) {
       return this.refreshHistorySection(repository)
     } else if (selectedSection === RepositorySectionTab.Changes) {


yufizan:
#9923 2.5.2

PR #9423, opened 2020-03-31
Simplify branch tip object

rafeca:
## Description

This PR changes the `tip` object of the `Branch` model to not store a full commit, but only the basic information that's needed by the UI (which is the `sha` of the tip and its `CommitIdentity`.

This allows https://github.com/desktop/desktop/pull/9383 to work fine without needing to have `Commit` objects with `null` tags (which can cause unexpected issues in the future).

See https://github.com/desktop/desktop/pull/9383#issuecomment-606515192

### Screenshots

N/A

## Release notes

Notes: no-notes


outofambit:
merging this—onward!

outofambit:
prefactor to support #9423

Issue #9364, opened 2020-03-26
Wrong keybind for Developer Tools, after Release 2.4.0


ruggi99:
### Describe the bug

With version 2.4.0, I'm not able to open Developer Tools with "Ctrl + Shift + I" keybind anymore.

### Version & OS

Version 2.4.0, OS: Windows 10

### Steps to reproduce the behavior

1. Open Github Desktop.
2. Use "Ctrl + Shift + I" keybind.

### Expected behavior

Open developer Tools.

### Actual behavior

Opens browser to create a new issue on Github Web.

### Additional information

Trying to open Developer Tools from "View" > "Toggle developer tools" menu, works
With the previous version, keybind worked

outofambit:
hi @ruggi99 thanks for reporting this! Sorry about this error, this is definitely a bug resulting from work in #9142. We'll find a different key binding for the "create an issue" action and be sure to fix this in the next release.

PR #9430, opened 2020-04-01
Resolved Accelerator Conflict

jfgordon2:
Fixes: https://github.com/desktop/desktop/issues/9364

## Description

On Windows and Linux, the keyboard shortcut `Ctrl+Shift+I` was assigned to two actions: 
1) open developer tools, and 
2) create issue in repository on GitHub.

Staying consistent with Chrome, retained `Ctrl+Shift+I` for opening developer tools, and
changed the shortcut to create an issue to `Ctrl+I` for just these platforms (MacOS retains `Cmd+Shift+I` since its shortcut to launch developer tools does not overlap).

**TODO**: Keyboard shortcuts on [help page](https://help.github.com/en/desktop/getting-started-with-github-desktop/keyboard-shortcuts-in-github-desktop) need updating.

## Release notes

Notes: The keyboard shortcut to launch developer tools on Windows now works as expected.
 Updated the keyboard shortcut to create a new issue in a repository to just `Ctrl+I` for Windows.

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -331,9 +331,7 @@ export function buildDefaultMenu({
         label: __DARWIN__
           ? 'Create Issue on GitHub'
           : 'Create &issue on GitHub',
-        accelerator: (() => {
-          return 'CmdOrCtrl+I'
-        })(),
+        accelerator: 'CmdOrCtrl+I',
         click: emit('create-issue-in-repository-on-github'),
         visible: enableCreateGitHubIssueFromMenu(),
       },
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -332,7 +332,7 @@ export function buildDefaultMenu({
           ? 'Create Issue on GitHub'
           : 'Create &issue on GitHub',
         accelerator: (() => {
-          return __DARWIN__ ? 'Cmd+Shift+I' : 'Ctrl+I'
+          return 'CmdOrCtrl+I'
         })(),
         click: emit('create-issue-in-repository-on-github'),
         visible: enableCreateGitHubIssueFromMenu(),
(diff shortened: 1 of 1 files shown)


jfgordon2:
@tierninho Thanks for testing this out so quickly!  I just rebuilt it and it seems to be working with the new keybindings on my end.

It's strange that yours still retained the old binding with that line removed -- I wonder if `Ctrl+I` has some sort of system-wide binding to italics or some other strange effect.

Do you see anything obvious on your end?

tierninho:
Thanks for the update @jfgordon2 

> It's strange that yours still retained the old binding with that line removed -- I wonder if Ctrl+I has some sort of system-wide binding to italics or some other strange effect.

I suspect that might be the case. For some reason, I cannot get it to work. 

cc @desktop/core for eyes on this as it is a bug fix for a production issue.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -331,7 +331,9 @@ export function buildDefaultMenu({
         label: __DARWIN__
           ? 'Create Issue on GitHub'
           : 'Create &issue on GitHub',
-        accelerator: 'CmdOrCtrl+Shift+I',
+        accelerator: (() => {
+          return __DARWIN__ ? 'Cmd+Shift+I' : 'Ctrl+I'
```
outofambit, line 335: I appreciate not changing more than we have to, but since this is a relatively new accelerator, lets keep them consistent across platforms! (And now we can drop the self-executing anonymous function here, too.)

```suggestion
          return 'CmdOrCtrl+I'
```

jfgordon2:
@tierninho I tested again with @outofambit's suggested change, and it's working successfully again on my side, but I'm curious about yours.

tierninho:
Thanks for the update @jfgordon2 ! Mac works and looks great now, but for some reason Windows keeps showing the old key bindings for me. I checked the tip of the branch and it is up-to-date, so not sure what is going on.

@outofambit any ideas or if you can confirm all is working on your windows env I am good with that? 🙇

tierninho:
Confirming the shortcut will be updated in our docs, as the necessary PR was submitted for the same.

say25:
@outofambit @tierninho @jfgordon2  can we update the release notes to reflect that the Mac shortcut has changed to `⌘ I`

outofambit:
> @outofambit @tierninho @jfgordon2 can we update the release notes to reflect that the Mac shortcut has changed to `⌘ I`

updated! thanks for catching that @say25 !

PR #9432, opened 2020-04-01
Add createTag() method to git library wrapper

### LINKED ISSUE #9426: Add lib function to create and list git tags
a thin wrapper around a dugite call that will address most of our create tag cases

and a separate method to list all tags in a repository, so we can proactively warn the user about naming collisions

see files in `app/src/lib/git/` for examples

_____
part of work for #9424

rafeca:
Closes #9426

This PR is based on https://github.com/desktop/desktop/pull/9383 since the new tests use the `tags` property on the `Commit` model.

## Description

This PR adds a `createTag()` method to the git library wrapper which is a requirement to end up adding a UI that allows tags to be created.

Things still pending:
- Decide on whether the created tags should be annotated or not.
- Update unit tests to assert that a special error code is thrown when trying to create a tag with an existing name (we'll be able to do this once https://github.com/desktop/dugite/issues/393 is done).

### Screenshots

N/A

## Release notes

Notes: no-notes


app/src/lib/git/tag.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param repository    The repository in which to get all the tags from.
+ */
+export async function getAllTags(
+  repository: Repository
+): Promise<ReadonlyArray<string>> {
+  const args = ['tag']
+
+  const tags = await git(args, repository.path, 'getAllTags')
+
+  return tags.stdout.split('\n').filter(Boolean)
```
outofambit, line 33: @rafeca is the `.filter(Boolean)` to filter out empty lines? we generally avoid boolean coercion in the codebase, so maybe this could be: 
```suggestion
  return tags.stdout.split('\n').filter(s => s !== '')
```

rafeca: Yes, it's to filter empty lines. Thanks for the suggestion!

PR #9461, opened 2020-04-06
upgrade dugite to 1.88.3

outofambit:
supports #9440


rafeca:
Based on the build error, it looks like we need to add a handler for the new error code in `getDescriptionForError()`.

outofambit:
sorry @rafeca i had fixed this yesterday but pushed my commit to the wrong remote 🤦‍♀

Issue #9467, opened 2020-04-07
Co-author auto-complete list is muddled 


tierninho:
I cannot read this as all the usernames are stacked on top of each other

![Screen Shot 2020-04-07 at 11 40 33 AM](https://user-images.githubusercontent.com/14828183/78723006-54377600-78c6-11ea-889e-e8de95600e62.png)

Repro:
- clone cli/cli
- go to co-author field and type `@`
- as you type my username, it will eventually go away

_OSX, v2.4.0_

PR #9474, opened 2020-04-08
Use flex-shrink on the author selector modal

rafeca:
Closes https://github.com/desktop/desktop/issues/9467

## Description

This PR fixes a layout bug that appeared after [upgrading to Electron v7](https://github.com/desktop/desktop/pull/8967) on the modal that appears when searching for co-authors.

### Screenshots

Now the modal is shown correctly:

Screenshot 2020-04-08 at 11 59 43

## Release notes

Notes: [Fixed] Fix layout of the co-author auto-complete list.

### DIFF
--- a/app/styles/ui/_autocompletion.scss
+++ b/app/styles/ui/_autocompletion.scss
@@ -18,6 +18,7 @@
   overflow-y: auto;
 
   li {
+    flex-shrink: 0;
     height: 29px;
     padding: 0 var(--spacing);
   }


rafeca:
> However, in Windows, I noticed that the scroll bar is missing for the co-author menu.

Based on [this code comment](https://github.com/desktop/desktop/blob/development/app/styles/ui/_autocompletion.scss#L25-L30) I think this is on purpose:

> We can't replicate the nice scrollbars that we have in
> the regular autocomplete popup since that's based on List
> and this CodeMirror stuff is just an unordered list.
> The autocomplete popup is not intended to be used with
> a pointer-device anyway so not being able to mouse-wheel
> through it shouldn't be a big deal

I can investigate though if we can re-add the scroll on Windows here...

Issue #9435, opened 2020-04-01
Add tags to existing push machinery


outofambit:
from our research, we've tentatively settled on the `--follow-tags` option.

_____

part of work for #9424

PR #9481, opened 2020-04-08
Push tags when pushing changes to repository

rafeca:
Partially addresses https://github.com/desktop/desktop/issues/9435

This PR is based on https://github.com/desktop/desktop/pull/9440 since I need to use the `enableGitTagsCreation()` feature flag.

## Description

This PR adds the `--follow-tags` argument to the push commands done on Desktop to be able to push the relevant annotated tags (note that non-annotated tags won't get automatically pushed).

See the [official docs](https://git-scm.com/docs/git-push#Documentation/git-push.txt---follow-tags) about the `--follow-tags` argument for more information.

An important note is that the docs are a bit misleading, since `--follow-tags` will push any annotated tag that is reachable from any ref available on the remote repository (not only the tags reachable from refs that are getting pushed).

Next steps (to be done in separate PRs):

-  [ ] Display the push/pull button as pushable when there are unpushed tags locally.

### Screenshots

N/A

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/push.ts
+++ b/app/src/lib/git/push.ts
@@ -69,7 +69,7 @@ export async function push(
 
   if (!remoteBranch) {
     args.push('--set-upstream')
-  } else if (options.forceWithLease) {
+  } else if (options.forceWithLease === true) {
     args.push('--force-with-lease')
   }
 
--- a/app/src/lib/git/push.ts
+++ b/app/src/lib/git/push.ts
@@ -15,6 +15,12 @@ import { IRemote } from '../../models/remote'
 import { envForRemoteOperation } from './environment'
 
 export type PushOptions = {
+  /**
+   * Force-push the branch without losing changes in the remote that
+   * haven't been fetched.
+   *
+   * See https://git-scm.com/docs/git-push#Documentation/git-push.txt---no-force-with-lease
+   */
   readonly forceWithLease: boolean
 }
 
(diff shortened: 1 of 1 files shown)


app/src/lib/git/push.ts
```diff
@@ -13,8 +13,15 @@ import { PushProgressParser, executionOptionsWithProgress } from '../progress'
 import { AuthenticationErrors } from './authentication'
 import { IRemote } from '../../models/remote'
 import { envForRemoteOperation } from './environment'
+import { enableGitTagsCreation } from '../feature-flag'
 
-export type PushOptions = {
+export interface IPushOptions {
```
outofambit, line 18: @rafeca why is this an interface now instead of a type? i'm not opposed i'm just not following 😕

rafeca: ups I changed this because I saw other options objects around the codebase using an interface, but now that I look more closely both `type` an `interface` are used similarly:

* `git grep 'interface [^\s]*Options' ':(exclude)app/test/fixtures/*' | wc -l` => 7 matches
* `git grep 'type [^\s]*Options' ':(exclude)app/test/fixtures/*' | wc -l` => 2 matches

I'm gonna remove this change from this PR and maybe revisit this in a separate PR.

Lazer39:
Thanks.

app/src/lib/git/push.ts
```diff
@@ -60,10 +69,14 @@ export async function push(
 
   if (!remoteBranch) {
     args.push('--set-upstream')
-  } else if (options !== undefined && options.forceWithLease) {
+  } else if (options.forceWithLease) {
```
outofambit, line 72: sorry, this is nit-picky
```suggestion
  } else if (options.forceWithLease === true) {
```

PR #9491, opened 2020-04-09
DO NOT MERGE: Keep list of remote tags and unpushed tags locally

rafeca:
Related issue: https://github.com/desktop/desktop/issues/9435

This PR is based on https://github.com/desktop/desktop/pull/9481

## Description

This PR is a prototype of a solution to handle un-pushed tags in GitHub Desktop.

## Background

`git` doesn't store locally the remote tags (as opposed as it does with branches), which means that it's not possible for Desktop to know (and display to the user) if there are tags to push without making a request to the remote repository.

## Solution

This PR attempts to solve the problem mentioned above by implementing some logic that gets the remote tags from a remote when fetching from that remote, and keeping these remote tags stored locally in the `IndexedDB` database.

## Benefits of this solution

This solution allows us to access/sync remote tags in a similar way that Desktop currently does for branches (by only doing network requests when performing fetches), which gives us consistency in terms of behaviour between branches and tags.

Additionally, this allows us to calculate in a performant way every single un-pushed tag, and this would allow displaying that information in the commit history list without issues.

## Main drawback of the solution

The main drawback here is that it makes it hard to calculate whether the push/pull button should be in a push state (which should be done by predicting whether a push would actually send any tag to the remote).

The problem lies on the fact that we're using `git push --follow-tags` in order to push the tags to the remote. The logic that `git` uses for `--follow-tags` is not simple: it first searches all local annotated tags and then for each of them it checks if it's an ancestor of an existing ref in the remote.

We can perform this same logic via `git` commands (see the `gitStore.refreshTagsToPush()`), but unfortunately this requires us to execute a `git` command N times (for each of the unpushed tags of the repository), which can cause performance bottlenecks on repositories with a lot of tags.

Because of this performance issue, it's desirable to predict the tags that are going to be pushed by performing a dry-run `git push`, which will directly output the tags that will get pushed.

## Conclusions

While this solution doesn't seem suitable to solve the whole problem, I think we can reuse most of it whenever we want to implement unpushed tag indicators on commits, since the part that calculates the list of tags that have not been pushed should be performant and sound.

If we ever want to stop using `--follow-tags` to push the local tags to the remote repository (maybe we could just push all unpushed annotated tags?), then this solution would be viable for everything.

In the meantime, I think it makes more sense (and it would me much simpler to implement) a solution that just executes `git push --follow-tags --dry-run` to predict whether any local tag will get pushed and based on that change the push/pull button.


rafeca:
Closing this PR since we're going to go through the solution implemented by https://github.com/desktop/desktop/pull/9492

We can always revisit this in the future

PR #9492, opened 2020-04-09
Add fetchTagsToPush() git method

rafeca:
Partially addresses #9435

This branch is based on https://github.com/desktop/desktop/pull/9481

## Description

This PR uses `git push --follow-tags --dry-run` to be able to predict whether any new tag will get pushed as part of the push command.

The next step here is to use the `fetchTagsToPush()` method to check whether the push/pull button should be shown as push.

We can avoid doing network requests when not necessary by only calling this method when no local commits need to be pushed (otherwise, the button will already be shown as "push").

### Screenshots

N/A

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/tag.ts
+++ b/app/src/lib/git/tag.ts
@@ -2,6 +2,7 @@ import { git, gitNetworkArguments } from './core'
 import { Repository } from '../../models/repository'
 import { IGitAccount } from '../../models/git-account'
 import { IRemote } from '../../models/remote'
+import { envForRemoteOperation } from './environment'
 
 /**
  * Create a new tag on the given target commit.
@@ -36,7 +37,7 @@ export async function getAllTags(
 }
 
 /**
- * Fetches the unpushed tags from the remote repository (it does a network request).
+ * Fetches the tags that will get pushed to the remote repository (it does a network request).
  *
  * @param repository  - The repository in which to check for unpushed tags
  * @param account     - The account to use when authenticating with the remote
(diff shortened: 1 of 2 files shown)


app/test/unit/git/tag-test.ts
```diff
@@ -70,4 +96,59 @@ describe('git/tag', () => {
...
+      await createTag(repository, 'my-new-tag', 'HEAD')
+
+      await push(repository, account, originRemote, 'master', null)
+
+      expect(
+        await fetchTagsToPush(repository, account, originRemote, 'master')
+      ).toEqual([])
+    })
+
+    it('does not return a tag created o a non-pushed branch', async () => {
```
outofambit, line 137:
```suggestion
    it('does not return a tag created on a non-pushed branch', async () => {
```

app/src/lib/git/tag.ts
```diff
@@ -32,3 +35,55 @@ export async function getAllTags(
...
+
+  const result = await git(args, repository.path, 'fetchTagsToPush', {
+    env: await envForRemoteOperation(account, remote.url),
+  })
+
+  const lines = result.stdout.split('\n')
+  let currentLine = 1
+  const unpushedTags = []
+
+  while (lines[currentLine] !== 'Done') {
```
outofambit, line 73: had to look this up for myself, could we record it in a comment?
```suggestion
  // the last line of this porcelain command is always 'Done'
  while (lines[currentLine] !== 'Done') {
```

PR #9498, opened 2020-04-10
Change deprecated getZoomFactor() with zoomFactor property

ruggi99:
Closes #[issue number] (no issue related)

## Description

Executing:
```
yarn start
```
there will be a warning in the log saying:
```
(electron) 'getZoomFactor function' is deprecated and will be removed. Please use 'zoomFactor property' instead.
```

Documentation:
https://github.com/electron/electron/blob/master/docs/api/modernization/property-updates.md

## Release notes

Notes:


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -631,12 +631,12 @@ function zoom(direction: ZoomDirection): ClickHandler {
...
     } else {
-      const rawZoom = webContents.getZoomFactor()
+      const rawZoom = webContents.zoomFactor
       const zoomFactors =
         direction === ZoomDirection.In ? ZoomInFactors : ZoomOutFactors
 
-      // So the values that we get from getZoomFactor are floating point
-      // precision numbers from chromium that don't always round nicely so
+      // So the values that we get from zoomFactor property are floating point
+      // precision numbers from chromium, that doesn't always round nicely, so
```
say25, line 639:
```suggestion
      // precision numbers from chromium, that don't always round nicely, so
```
I think this reads better

PR #9509, opened 2020-04-13
add explicit onDismissed handler for force push warn dialog

### LINKED ISSUE #9507: "Rebase will require force push" dialog window: Cancel button does not work
"Rebase will require force push" dialog window: Cancel button does not work
- https://i.ibb.co/LvTLDjx/github-desktop-rebase-will-require-force-push-cancel-does-not-work.png

outofambit:
Closes #9507

## Description

- this is a workaround for a bug introduced in #8599, where non-dismissable dialogs that use the OkCancelButtonGroup and don't pass an explicit onCancel handler will fail to dismiss
- i have an idea for a more permanent fix, but want to vet it a little more
- i also did a quick audit and didn't find any other popups that also have this problem

cc @tierninho @billygriffin


outofambit:
CircleCI is failing with a strange cloning the desktop repo error. will look into later if it persists

Issue #9427, opened 2020-03-31
Display tags in commit list and commit details


outofambit:
From @ampinsk:

We decided to prototype with this design and may adjust depending on how this looks with real data

<img width="267" alt="View-Tags-07@2x" src="https://user-images.githubusercontent.com/10404068/78723797-f3696700-78e0-11ea-94df-96e961df4f0e.png">

_____
part of work for #9424

PR #9513, opened 2020-04-14
Tweaks to displayed tags on commit view

rafeca:
Partially addresses https://github.com/desktop/desktop/issues/9427

## Description

This PR implements a few tweaks to the display of tags on the commit details view:

1. Better wrapping of text when there are a lot of tags (or very long ones).
2. Tags are now sorted alphabetically.
3. Now a tooltip is shown when hovering for a few seconds on the tag list to see the whole list of tags.

### Screenshots

Before:

Screenshot 2020-04-14 at 17 45 24

After:

Screenshot 2020-04-14 at 17 44 50

(notice that the author name is not wrapped anymore).

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/history/commit-summary.tsx
+++ b/app/src/ui/history/commit-summary.tsx
@@ -338,14 +338,20 @@ export class CommitSummary extends React.Component<
               />
             </li>
 
-            <li className="commit-summary-meta-item" aria-label="SHA">
+            <li
+              className="commit-summary-meta-item without-truncation"
+              aria-label="SHA"
+            >
               <span aria-hidden="true">
                 <Octicon symbol={OcticonSymbol.gitCommit} />
               </span>
               <span className="sha">{shortSHA}</span>
             </li>
 
-            <li className="commit-summary-meta-item" title={filesDescription}>
+            <li
+              className="commit-summary-meta-item without-truncation"
+              title={filesDescription}
+            >
               <span aria-hidden="true">
                 <Octicon symbol={OcticonSymbol.diff} />
               </span>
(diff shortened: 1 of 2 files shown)


app/src/ui/history/commit-summary.tsx
```diff
@@ -384,13 +396,15 @@ export class CommitSummary extends React.Component<
       return null
     }
 
+    const sortedTags = [...tags].sort(caseInsensitiveCompare)
+
     return (
-      <li className="commit-summary-meta-item">
+      <li className="commit-summary-meta-item" title={sortedTags.join('\n')}>
```
outofambit, line 402: i think a comma here will make this more useful for screenreaders
```suggestion
      <li className="commit-summary-meta-item" title={sortedTags.join(',\n')}>
```

app/src/ui/history/commit-summary.tsx
```diff
@@ -355,15 +362,20 @@ export class CommitSummary extends React.Component<
...
-              <Checkbox
-                label="Hide Whitespace"
-                value={
-                  this.props.hideWhitespaceInDiff
-                    ? CheckboxValue.On
-                    : CheckboxValue.Off
-                }
-                onChange={this.onHideWhitespaceInDiffChanged}
-              />
+              <li
```
outofambit, line 365: 👍🏼

app/src/ui/history/commit-summary.tsx
```diff
@@ -384,13 +396,15 @@ export class CommitSummary extends React.Component<
       return null
     }
 
+    const sortedTags = [...tags].sort(caseInsensitiveCompare)
```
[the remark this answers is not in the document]
rafeca: We haven't yet discussed this with @ampinsk  and @billygriffin 😅, this came up from @tierninho comments in https://github.com/desktop/desktop/pull/9440#pullrequestreview-391100749 .

>  i'm inclined to leave them in the order they are returned by git.

From my tests it looks like `git log` prints the tags on inverse alphabetical order (by looking at the code I think it returns the tags in the reverse order as `git for-each-ref`, but I'm not sure if `git for-each-ref` returns the same sorting on all operating systems).

> what's the reasoning for sorting tags alphabetically?

I think that alphabetical sorting may be more intuitive in some situations (e.g tags with names that are standard words), although on the other side inverse alphabetical sorting could be better for semver tags (this way greater versions would get displayed before smaller versions).

Still, the sorting of tags is only going to be relevant when there are multiple tags associated with the same commit, which I guess happens very rarely, so I don't think there's going to be a big difference between the different sortings.

app/src/ui/history/commit-summary.tsx
rafeca: I've used newlines to separate the tags in the tooltip, so commas won't play nice with that... I wonder if I can make screenreaders ignore the title (since it contains exactly the same as the text inside the `<li>`).

outofambit: looks like we're fine to leave as-is after reading https://developer.paciellogroup.com/blog/2013/01/using-the-html-title-attribute-updated/

> Takeaway: If you want to hide content from mobile and tablet users as well as assistive tech users and keyboard only users, use the title attribute.

Issue #9429, opened 2020-03-31
git tag creation popup flow


outofambit:
includes:
* context menu to launch popup
* new popup with tag form (and validation / sanitization)
* basic error handling for making tag

_____
main UI work for #9424

PR #9514, opened 2020-04-14
Disable undoing commits that have tags associated

rafeca:
Related issue: https://github.com/desktop/desktop/issues/9429

Related discussion: https://github.com/desktop/desktop/pull/9440#pullrequestreview-391100749

## Description

This PR removes the "undo" UI when the most recent local commit has any tags associated to it.

This prevents users from keeping "orphaned tags" pointing to commits that were meant to be deleted, which can cause confusion to users (specially since the deleted commits won't be visible from Desktop but the tag will still exist).

Another option we've considered to solve this is to still allow undoing commits with associated tags, but displaying a message warning the user that any associated tag will get deleted. We've finally decided to just prevent undoing commits with tags since this approach is much simpler to implement and this scenario seems quite rare (we'll reevaluate if we find otherwise in the future).

### Screenshots

N/A

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/changes/sidebar.tsx
+++ b/app/src/ui/changes/sidebar.tsx
@@ -309,7 +309,7 @@ export class ChangesSidebar extends React.Component<IChangesSidebarProps, {}> {
   private onUndo = () => {
     const commit = this.props.mostRecentLocalCommit
 
-    if (commit) {
+    if (commit && commit.tags.length === 0) {
       this.props.dispatcher.undoCommit(this.props.repository, commit)
     }
   }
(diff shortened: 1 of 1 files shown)


rafeca:
> However, I was able to undo the commit with a tag by doing the following hack in OSX:

I'm going to add a safety check to make sure that even if the user clicks on the undo button the commit doesn't get undone!

> In general, are there any issues the user could run into if they are editing the same file and one of the commits is tagged (and cannot be undone)? I couldn't think of any, but a dev might have some more insight.

I'm not sure I understand this scenario... is it about two commits that edit the same file?

PR #9519, opened 2020-04-14
Conform verbiage example to our expectations

billygriffin:
Update docs to make the example present tense. cc: @nerdneha

### DIFF
--- a/docs/process/writing-release-notes.md
+++ b/docs/process/writing-release-notes.md
@@ -26,7 +26,7 @@ We don't include release notes that don't impact the user's experience using Des
 
 #### Exceptions for Security Vulnerability Fixes
 
-We'll make exceptions for security vulnerability fixes, especially if they're high profile:
+We'll make exceptions for security vulnerability fixes, especially if they're high profile. These are typically general in nature and do not include the CVE for the vulnerability. For example:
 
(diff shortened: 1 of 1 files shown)


docs/process/writing-release-notes.md
```diff
@@ -29,7 +29,7 @@ We don't include release notes that don't impact the user's experience using Des
 We'll make exceptions for security vulnerability fixes, especially if they're high profile:
```
nerdneha, line 29: How do we feel about being explicit and saying "While we won't mention the CVE, we will mention where the security vulnerability was" -- too explicit? It's come up twice so I wondered if it'd be helpful to memorialize

billygriffin, line 29:
```suggestion
We'll make exceptions for security vulnerability fixes, especially if they're high profile. These are typically general in nature and do not include the CVE for the vulnerability:
```

billygriffin: How's that @nerdneha?

nerdneha: Fantastic! 🚢

billygriffin: Done!

PR #7050, opened 2019-03-10
Upgrade to React 16.8.4

say25:
## Overview

**Supports #6685**

## Description

- Upgrade to the latest version of `react`, `react-dom`, `@types/react` & `@types/react-dom`

## Release notes

Notes: No notes

PR #9542, opened 2020-04-16
Removed REACT_PERF

### LINKED ISSUE #6685: deprecate React Perf extension for React Profiler
Over in https://github.com/desktop/desktop/pull/6442#issuecomment-455675571 we noticed an issue with committing in development mode triggering an error that brought down the app:

```
error: [main] TypeError: Cannot read property 'stop' of undefined
    at a (chrome-extension://react-perf/page.bundle.js:1:2004)
debug: [main] Starting crash process
debug: [main] Crash process in startup
debug: [main] Crash process is ready
debug: [main] Crash process started
debug: [main] Showing crash process window
debug: [main] Got quit signal from crash process
```

To me, this reads as the React-Perf extension encountering some weird state. This is only installed when running the app in development mode:

https://github.com/desktop/desktop/blob/52ab4b91016f3e82e13361161e89cc1bffece2c5/app/src/main-process/main.ts#L503-L524

I'm not sure this is currently working, or whether it ever did, so instead I'm opening this issue to do two things:

 - remove the work that installs the `REACT_PERF` dev tools when launching the app in dev mode
 - investigate how to use the new React Profiler (requires upgrading to `react-dom` 16.5+)

Start with [this link](https://reactjs.org/docs/optimizing-performance.html#profiling-components-with-the-devtools-profiler) and then [this blog post](https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html) and see if this feels like something we can use in the app at development time.

ADustyOldMuffin:
Closes #6685

Depends on PR #7050 as it requires React and React-Dom 16.5+

Because this depends on that PR it's marked as a draft

## Description

- This just removes the old profiler and I've also included some gifs of the new profiler below this.

### Screenshots

You can start profiling by going to the new Profiler tab in the dev tools.

Start_Profiling

Once you're done and want to see the results you click the stop button

Stop_Profiling

You can then drill down into details of the commits that were collected during the profiling.

View_Commit

You can read more about what all you can see inside of this view [here.](https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html)

[Here is also a video if you're not the reading type.](https://www.youtube.com/watch?v=nySib7ipZdk&feature=emb_title)

## Release notes

Notes: None, I don't think this impacts users unless you activate for production. 

Right now this only works in development mode, but you can enable it for production by building with some flags, but they mention how it has a performance impact.


outofambit:
hey @ADustyOldMuffin thanks for opening this! #7050 is merged now, could you update from `development` and make this ready for review? thanks!

ADustyOldMuffin:
@outofambit Thanks for the notification! I've done so.

PR #9545, opened 2020-04-16
Always return URL of current repository

Raymo111:
Regardless if a parent exists (i.e. if the repository is forked)

## Description
This will allow "open on GitHub" to open the repository open in Desktop regardless of the existence of a parent, instead of opening the repo from which it was forked. (that is confusing).

## Release notes

Notes:
"Open on GitHub" opens the current repository, not the one from which it was forked.

### DIFF
--- a/app/src/models/repository.ts
+++ b/app/src/models/repository.ts
@@ -114,8 +114,7 @@ export function nameOf(repository: Repository) {
 }
 
 /**
- * Get the GitHub html URL for a repository, if it has one.
- * Will return the parent GitHub repository's URL if it has one.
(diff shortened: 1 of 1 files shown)


Raymo111:
Although this has no existing issue, I believe that the change is worth it.

billygriffin:
Hi @Raymo111, we require that any substantive PR have an issue where we agree on a solution. There is an issue for this in #9500 and this was an intentional decision, so I’m going to close this PR and you’re welcome to weigh in there. The more specifics you can provide would be helpful. Thanks!

PR #9554, opened 2020-04-17
Evidence Commits

### LINKED ISSUE #9381: Visually distinguish commits that are just on my branch in history

If I'm on a branch (not the default one), It's known that if I want to get all commits that are ahead or behind from default branch (or others, but default one in this example), it's sufficient to go into History tab and do a compare with that branch.

### Proposed solution

I propose to add a spacer (line in image below) to separate ahead commits from default branch, so it's easier focusing on ahead commits only. Moreover it's faster than doing a compare each time.

### Additional context

Cattura

ruggi99:
Closes #9381

## Description

This PR is created to distinguish commits that are just on a branch and so not yet in default branch.
Here "evidence" class is added to these commits in the list (History Tab).

### Screenshots
Cattura

The black commit is just on evidencecommits branch

## Release notes

Notes:
- I've not add styles. So maintainers can choose the best graphic solution
- I'm not able to create tests for it. Please provide them


tierninho:
@ruggi99 Many thanks for pull request!

> I've not add styles. So maintainers can choose the best graphic solution

Just wanted to confirm the black background styling (as seen in the pic) is not added to this PR, as I did not see it when testing this?

My steps:
- Add a new commit on a branch other than default.
- Commit should have UI treatment until it is on default.

ruggi99:
You are right.
I've added black background in developer tools only to take a screenshot. Maybe changing background it's not the best graphic solution. Maybe putting an indicator (like the unpushed indicator) could be better.
I've flagged "allow edits from maintainers". Hope it works

Rexogamer:
This might be confusing, but here’s my idea: what if there was a indicator with a branch that had a tooltip (eg “this commit has not been pushed to <default branch>”)? probably not the right direction but might be something to consider

niik:
Hey @ruggi99, thanks for opening this and demonstrating how a feature like this could be implemented!

That said, until we've had a chance to decide on whether this is something we'll want to address over in #9381 we won't be reviewing this PR. In order to try and keep our open PR list manageable I'm going to close this PR but don't worry, your code won't be lost and we can easily reopen it should we choose to pursue this feature.

PR #9576, opened 2020-04-20
Use the `fetchTagsToPush()` output to display the push/pull information

rafeca:
Partially addresses #9435

This branch is based on #9492

## Description

This PR plugs the `fetchTagsToPush()` method introduced in #9492 with the application state logic and the UI to change two components based on wether there are tags pending to be pushed:

1. The push/pull button on the toolbar.
   Screenshot 2020-04-17 at 22 29 20
2. The prompt to push displayed on the no changes view.
   Screenshot 2020-04-17 at 22 30 02

### Changes in the push/pull button

- Now the button will show the "Push" action when there are tags to be pushed (even if there are no commits to be pushed).
- The number that appears next to the up arrow is meant to show the number of items to push (where a new commit with a new tag is counted as one). This means that:
  - When there are commits + tags to be pushed, it'll be the number of commits + tags.
  - When there are only commits to be pushed, i't will be the number of commits.
  - When there are only tags to be pushed, it'll be the number of tags.
- No other changes are done to this component.

### Changes in the prompt to push changes

- Now the prompt to push is going to be shown when there are tags to be pushed (even if there are no commits to be pushed).
- The title/description of the prompt changes depending on whether there are tags or/and commits to be pushed:
  - Only commits:
    Screenshot 2020-04-20 at 10 18 33
  - Only tags:
    Screenshot 2020-04-20 at 10 15 37
  - Commits and tags:
    Screenshot 2020-04-20 at 10 17 28

## Technical considerations

There are two important factors that have driven to the current implementation:

1. The `fetchTagsToPush()` method needs to perform a network request, which makes it much slower than any other local git operation.
2. `fetchTagsToPush()` can return different values when any of these values change:
   - Remote repository to push (in the case the repository has multiple remotes).
   - Currently checked out branch (if the checked out branch changes, `--follow-tags` may push unpushed tags from the newly checked out branch).
   - Ancestor commits of the current branch (if another branch with unpushed tags is merged into the current branch).
   - Existing local tags.
   - Existing remote tags (a tag previously marked as unpushed can get pushed from the terminal).

In order to keep the UI updated when any of the previous values change, we try to fetch the remote tags every time that `dispatcher.refreshRepository()` is called. Additionally, we also call it after a new tag is created via the Desktop UI.

`dispatcher.refreshRepository()` is called very often (when most of the repository state attributes change), which on one side it solves our problem of keeping the UI updated but on the other side causes too many calls to `fetchTagsToPush()` which will cause network requests.

#### Caching to the resque

In order to minimize the number of network requests, we've introduced a memoization layer that takes care of only calling the real `fetchTagsToPush()` method only when one of the values that can alter its output changes.

The values are the following:

- remote repository.
- checked out branch name.
- existing local tags.
- SHA of the HEAD commit of the current branch.

#### When we do bypass cache?

With this caching system we make sure that any meaningful change in the local repository will cause an invalidation of the cache . Unfortunately, we don't have any way to control the changes that happen in the remote repository and invalidate the cache in these scenarios.

To fix this, there's a method to clear the cache for a specific remote repository. This is used to force network requests after the user fetches, pulls or pushes from the remote repository. This way we solve the potential case where tags are created (or removed) from the remote without Desktop knowing (Desktop will get updated after the user syncs with the remote).

## Known limitations

Due to the way that pushing/fetching tags is implemented in `git`, when a tag gets either deleted or modified in the remote repository, weird things can happen:

1. If a tag gets deleted from the remote repository, Desktop won't know whether thee tag missing from the remote has been deleted or is a local unpushed tag. Because of this, Desktop will push again the tag to the remote next time the user pushes their changes.
2. In a similar fashion, if a tag gets modified in the remote repository (it gets deleted and recreated pointing to a different commit), Desktop won't update the local tag when fetching changes. Instead it'll keep the local tag pointing to the old commit. Interestingly, when pushing Desktop won't give any error (`git push --follow-tags` ignore conflicting tags in this scenario).

Note that these two scenarios and behaviours are not Desktop-specific: the same applies when using the official `git` CLI.

Since these scenarios are quite rare, and because `git` by default behaves in a similar way, we've decided to acknowledge that these are known limitations and we're not currently going to address them (we'll revisit if we learn that this causes friction to many users).

## Screenshots

N/A

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/tag.ts
+++ b/app/src/lib/git/tag.ts
@@ -1,8 +1,9 @@
-import { git, gitNetworkArguments } from './core'
+import { git, gitNetworkArguments, GitError } from './core'
 import { Repository } from '../../models/repository'
 import { IGitAccount } from '../../models/git-account'
 import { IRemote } from '../../models/remote'
 import { envForRemoteOperation } from './environment'
+import { IGitResult } from 'dugite'
 
 /**
  * Create a new tag on the given target commit.
@@ -62,16 +63,28 @@ export async function fetchTagsToPush(
     '--porcelain',
   ]
 
-  const result = await git(args, repository.path, 'fetchTagsToPush', {
-    env: await envForRemoteOperation(account, remote.url),
-  })
+  let result: IGitResult | null = null
 
-  const lines = result.stdout.split('\n')
+  try {
+    result = await git(args, repository.path, 'fetchTagsToPush', {
+      env: await envForRemoteOperation(account, remote.url),
+      successExitCodes: new Set([0, 1]),
+    })
+  } catch (e) {
+    // Even if we receive an unexpected error code, we try to
+    // do the best effort to try to parse the stdout to find unpushed tags.
+    // If we can't even parse the stdout, we return an empty array of unpushed tags.
+    if (e instanceof GitError) {
+      result = e.result
+    }
+  }
+
+  const lines = result ? result.stdout.split('\n') : []
   let currentLine = 1
   const unpushedTags = []
 
   // the last line of this porcelain command is always 'Done'
-  while (lines[currentLine] !== 'Done') {
+  while (currentLine < lines.length && lines[currentLine] !== 'Done') {
     const line = lines[currentLine]
     const parts = line.split('\t')
 
--- a/app/src/lib/git/tag.ts
+++ b/app/src/lib/git/tag.ts
@@ -23,18 +23,43 @@ export async function createTag(
 }
 
 /**
- * Gets all the local tags.
+ * Gets all the local tags. Returns a Map with the tag name and the commit it points to.
  *
  * @param repository    The repository in which to get all the tags from.
  */
 export async function getAllTags(
   repository: Repository
-): Promise<ReadonlyArray<string>> {
-  const args = ['tag']
+): Promise<Map<string, string>> {
+  const args = ['show-ref', '--tags', '-d']
 
-  const tags = await git(args, repository.path, 'getAllTags')
+  const tags = await git(args, repository.path, 'getAllTags', {
+    successExitCodes: new Set([0, 1]), // when there are no tags, git exits with 1.
+  })
 
-  return tags.stdout.split('\n').filter(s => s !== '')
+  const tagsArray: Array<[string, string]> = tags.stdout
+    .split('\n')
+    .filter(line => line !== '')
+    .map(line => {
+      const [commitSha, rawTagName] = line.split(' ')
+
+      // Normalize tag names by removing the leading ref/tags/ and the trailing ^{}.
+      //
+      // git show-ref returns two entries for annotated tags:
+      // deadbeef refs/tags/annotated-tag
+      // de510b99 refs/tags/annotated-tag^{}
+      //
+      // The first entry sha correspond to the blob object of the annotation, while the second
+      // entry corresponds to the actual commit where the tag was created.
+      // By normalizing the tag name we can make sure that the commit sha gets stored in the returned
+      // Map of commits (since git will always print the entry with the commit sha at the end).
+      const tagName = rawTagName
+        .replace(/^refs\/tags\//, '')
+        .replace(/\^\{\}$/, '')
+
+      return [tagName, commitSha]
+    })
+
+  return new Map(tagsArray)
 }
 
 /**
--- a/app/src/lib/git/tag.ts
+++ b/app/src/lib/git/tag.ts
@@ -1,9 +1,8 @@
-import { git, gitNetworkArguments, GitError } from './core'
+import { git, gitNetworkArguments } from './core'
 import { Repository } from '../../models/repository'
 import { IGitAccount } from '../../models/git-account'
 import { IRemote } from '../../models/remote'
 import { envForRemoteOperation } from './environment'
-import { IGitResult } from 'dugite'
 
 /**
  * Create a new tag on the given target commit.
@@ -88,23 +87,12 @@ export async function fetchTagsToPush(
     '--porcelain',
   ]
 
-  let result: IGitResult | null = null
+  const result = await git(args, repository.path, 'fetchTagsToPush', {
+    env: await envForRemoteOperation(account, remote.url),
+    successExitCodes: new Set([0, 1]),
+  })
 
-  try {
-    result = await git(args, repository.path, 'fetchTagsToPush', {
-      env: await envForRemoteOperation(account, remote.url),
-      successExitCodes: new Set([0, 1]),
-    })
-  } catch (e) {
-    // Even if we receive an unexpected error code, we try to
-    // do the best effort to try to parse the stdout to find unpushed tags.
-    // If we can't even parse the stdout, we return an empty array of unpushed tags.
-    if (e instanceof GitError) {
-      result = e.result
-    }
-  }
-
-  const lines = result ? result.stdout.split('\n') : []
+  const lines = result.stdout.split('\n')
   let currentLine = 1
   const unpushedTags = []
 
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -114,6 +114,8 @@ export class GitStore extends BaseStore {
 
   private _defaultBranch: Branch | null = null
 
+  private _localTags: Set<string> | null = null
+
   private _allBranches: ReadonlyArray<Branch> = []
 
   private _recentBranches: ReadonlyArray<Branch> = []
@@ -262,8 +264,10 @@ export class GitStore extends BaseStore {
     return commits.map(c => c.sha)
   }
 
-  public async getAllTags(): Promise<ReadonlyArray<string>> {
-    return getAllTags(this.repository)
+  public async loadLocalTags() {
+    this._localTags = new Set(await getAllTags(this.repository))
+
+    this.emitUpdate()
   }
 
   public async createTag(
@@ -281,6 +285,7 @@ export class GitStore extends BaseStore {
       this.storeCommits([foundCommit], true)
     }
 
+    await this.loadLocalTags()
     this.fetchTagsToPush(account)
   }
 
@@ -291,6 +296,10 @@ export class GitStore extends BaseStore {
 
   public get tagsToPush(): ReadonlyArray<string> | null {
     return this._tagsToPush
+  }
+
+  public get localTags(): Set<string> | null {
+    return this._localTags
   }
 
   /** Load all the branches. */
@@ -393,24 +402,28 @@ export class GitStore extends BaseStore {
       return
     }
 
+    const localTags = this._localTags
+    if (localTags === null) {
+      this._tagsToPush = null
+      return
+    }
+
     if (this.tip.kind !== TipState.Valid) {
       this._tagsToPush = null
       return
     }
     const currentBranch = this.tip.branch
 
-    const tagsToPush = await this.performFailableOperation(async () => {
-      const localTags = await this.getAllTags()
-
-      return fetchTagsToPushMemoized(
+    const tagsToPush = await this.performFailableOperation(() =>
+      fetchTagsToPushMemoized(
         this.repository,
         account,
         currentRemote,
         currentBranch.name,
-        localTags,
+        Array.from(localTags),
         currentBranch.tip.sha
       )
-    })
+    )
     this._tagsToPush = tagsToPush !== undefined ? tagsToPush : null
 
     this.emitUpdate()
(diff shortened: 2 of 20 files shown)


app/src/lib/stores/helpers/fetch-tags-to-push-memoized.ts
```diff
@@ -0,0 +1,73 @@
...
+  repository: Repository,
+  account: IGitAccount | null,
+  remote: IRemote,
+  branchName: string,
+  _localTags: ReadonlyArray<string>, // only used for cache invalidation.
+  _currentTipSha: string, // only used for cache invalidation.
+  _options: { forceFetch: boolean } // only used for cache invalidation.
+) {
+  return fetchTagsToPush(repository, account, remote, branchName)
+}
```
rafeca, line 56: I'm going to check whether it makes more sense to use a standard cache via a `Map` object instead of `memoizeOne` here...

tierninho:
Some initial testing and comments before this is `ready for review`: 

- I suggest we change the following as using the word "with" instead of "and" implies that the two are associated with each other as a tag and a commit can be pushed for two different SHA.

  Current prompt language:
  “You have one local commit **with** one tag…”

  Suggestion:
  “You have one local commit **and** one tag…”

- Question for this logic:

  > * When there are commits + tags to be pushed, it'll be the number of commits.
  > * When there are only commits to be pushed, i't will be the number of commits.
  > * When there are only tags to be pushed, it'll be the number of tags.

  Are we counting tags created from the Desktop UI the same as those from the cmd line? Also, do commits + tags created before pushing have to be on the same commit?

- I switched branches before I decided to push new commits + tags and got an error:

  ```
  install.ts:23 `git -c credential.helper= -c protocol.version=2 push origin master --follow-tags --dry-run 
  --porcelain` exited with an unexpected code: 1.
  stdout:
  To https://github.com/tierninho/gfhfghfg.git
  *	refs/tags/DDD:refs/tags/DDD	[new tag]
  *	refs/tags/adf:refs/tags/adf	[new tag]
  !	refs/heads/master:refs/heads/master	[rejected] (fetch first)
  Done
  ```
  
- Lastly, can you confirm that we are pulling down remote tags in this PR? Or should I just focus on the push functionality?

rafeca:
Thanks for looking at this @tierninho !!

> I suggest we change the following as using the word "with" instead of "and" implies that the two are associated with each other as a tag and a commit can be pushed for two different SHA.

That's a good point! when chatting with @ampinsk we considered the scenario where unpushed tags point to unpushed commits, but as you mention there can be unpushed tags that point to commits that have already been pushed. Taking this into account I think we should change the text to `and`.

> Are we counting tags created from the Desktop UI the same as those from the cmd line?

We're only counting annotated tags, but they can have been created from either the Desktop UI or the command line (we have no way to differenciate them).

> Also, do commits + tags created before pushing have to be on the same commit?

I think that this is related to the first point right? A tag can be associated to a commit that has already been pushed, so maybe instead of `When there are commits + tags to be pushed, it'll be the number of commits.` we should do `When there are commits + tags to be pushed, it'll be the number of commits + tags.`

> I switched branches before I decided to push new commits + tags and got an error:

Gonna look at this one, haven't seen it yet! did you get the error in a modal or you saw it on the logs?

> Lastly, can you confirm that we are pulling down remote tags in this PR? Or should I just focus on the push functionality?

I'm pretty sure that GitHub Desktop have pulled down remote tags for some time (before we started working on tags). We haven't changed anything on the fetching logic so I don't think we need to do specific tests about fetching remote tags (maybe we only need to test that the existing fetching logic plays well with the rest of things).

rafeca:
> I switched branches before I decided to push new commits + tags and got an error:

I've been able to reproduce and fix this bug in the last commit, good catch!

rafeca:
This is ready for review!

I'm going to try to fix the following point in a separate PR, since it's quite isolated and will need some design/product input:

> In a similar fashion, if a tag gets modified in the remote repository (it gets deleted and recreated pointing to a different commit), Desktop won't update the local tag when fetching changes. Instead it'll keep the local tag pointing to the old commit and fail when trying to push.

app/src/lib/stores/helpers/fetch-tags-to-push-memoized.ts
rafeca: I've finally used a `Map`, since it gives us more flexibility to invalidate the cache and the code is simpler: f7e3244c504d12ef8b04d3198c4b4e6cbc55c1b4

ampinsk:
> Taking this into account I think we should change the text to `and`.

+1 on this!

rafeca:
> I did notice that git tags created in the cmd line were not getting accounted for in the prompt. The commit caused the prompt to increment, but the tag associated with it did not.

We finally decided to only push annotated tags, so in order to create a tag that will get pushed via Desktop you need to execute `git tag -a tagck 81e19c0`. Does it work this way?

tierninho:
> Does it work this way?

Yes, thank you!

app/src/lib/stores/git-store.ts
```diff
@@ -372,6 +385,37 @@ export class GitStore extends BaseStore {
...
+        account,
+        currentRemote,
+        currentBranch.name,
+        localTags,
+        currentBranch.tip.sha
+      )
+    })
+    this._tagsToPush = tagsToPush !== undefined ? tagsToPush : null
+
+    this.emitUpdate()
```
outofambit, line 416: do we need to emit an update if the list of tags to push doesn't change?

app/src/lib/stores/helpers/fetch-tags-to-push-memoized.ts
```diff
@@ -0,0 +1,77 @@
+import { fetchTagsToPush } from '../../git'
+import { Repository } from '../../../models/repository'
+import { IGitAccount } from '../../../models/git-account'
+import { IRemote } from '../../../models/remote'
+
+const cache = new Map<string, Map<string, ReadonlyArray<string>>>()
```
outofambit, line 6: one thing i think @niik would note here is that this cache can grow unbounded. i don't think that's likely to matter in this case though, as these arrays will generally be quite short or empty.

app/src/lib/stores/git-store.ts
```diff
@@ -372,6 +385,37 @@ export class GitStore extends BaseStore {
...
+    }
+
+    if (this.tip.kind !== TipState.Valid) {
+      this._tagsToPush = null
+      return
+    }
+    const currentBranch = this.tip.branch
+
+    const tagsToPush = await this.performFailableOperation(async () => {
+      const localTags = await this.getAllTags()
```
outofambit, line 403: this line will get called every time this method is called (via `_refreshRepository` for example). `_refreshRepository` is called from a _lot_ of different places, but i think we only need to load all the tags in the following cases:

* after a push/pull/fetch
* after the app is foregrounded
* when the user switches repos or branches

does that sound right? is it possible to only call this method in these cases?
_____
**more context:** `_refreshRepository` gets called from a lot of different places and its pretty expensive computationally. we've had performance issues in the past when we've added too many git calls (or other work) to it, so i'm being extra cautious here.

app/src/lib/git/tag.ts
```diff
@@ -62,16 +63,28 @@ export async function fetchTagsToPush(
...
+  } catch (e) {
+    // Even if we receive an unexpected error code, we try to
+    // do the best effort to try to parse the stdout to find unpushed tags.
+    // If we can't even parse the stdout, we return an empty array of unpushed tags.
+    if (e instanceof GitError) {
+      result = e.result
+    }
+  }
+
+  const lines = result ? result.stdout.split('\n') : []
```
outofambit, line 82: I'm not sure returning an empty list for error cases is a good idea with the cacheing in this PR. It seems like we should not cache an error result, or at least not cache error results that are empty lists.

@rafeca do we have a sense of what kind of errors we might encounter here, or is this a general precaution?

app/src/lib/stores/helpers/fetch-tags-to-push-memoized.ts
outofambit: on second thought, with the key specificity here, it might be worth being safe by removing the previous entry for a branch or sha when a new one is cached? i don't feel strongly about this and am happy to leave this alone until it becomes a problem, if it ever does.

app/src/lib/git/tag.ts
rafeca: Really good point! ✨  This is meant to silently handle at least two kind of expected errors (but there may be more):

1. Current branch is not fully in sync with the remote one (3rd point in @tierninho's [comment](https://github.com/desktop/desktop/pull/9576#issuecomment-616902677)).
2. Computer not connected to Internet.

I see the issue of caching an error result, which may get resolved over time while the cached result will be outdated with the empty list. Still, if we don't cache anything when we have these errors, it's possible that we end making too many dry-run pushes in these scenarios...

Good news is that the current cache handles scenario 1. (the cache key contains the HEAD commit sha, so once the branch gets synced the cache will get invalidated), but scenario 2 or other potential errors will result in outdated results. I see a few options here:

1. Don't cache any error result (this will cause a lot of operation when having persistent errors).
2. Keep caching error results (this will cause outdated results when having temporary errors).
3. Use a smarter cache (like [`node-lru-cache`](https://github.com/isaacs/node-lru-cache) which allows us to set a max age for entries. This won't completely solve any of the problems of the other two approaches but can be in a nice sweet spot to mitigate both quite reasonably. 

Given that 3. will also solve https://github.com/desktop/desktop/pull/9576/files#r416242898 I'm going to opt to implement that one, if you feel strongly about it we can reconsider it.

app/src/lib/stores/git-store.ts
rafeca: No we don't 😄 I'll change that 🤡

app/src/lib/stores/git-store.ts
rafeca: Thanks for the tip!

Yup I think it's possible to just load the tags in any of these situations. I'm going to add this commit: https://github.com/desktop/desktop/pull/9638/commits/b2319c584e8392b7969a261270986b8fd8b36402 to this PR so we keep the list of local tags in memory, this way I'll be able to call `loadLocalTags` independently from `fetchTagsToPush`.

app/src/lib/stores/helpers/fetch-tags-to-push-memoized.ts
rafeca: The growing of the cache gets a bit limited since it gets cleared for a repository every time there's a fetch/push/pull, but I agree that it can still grow unbounded on certain scenarios 👍

I'm going to try to solve this issue by using a LRU cache as described in https://github.com/desktop/desktop/pull/9576/files#r416450730.

rafeca:
@outofambit thanks for the review! I've added a few commits to address your comments and I think the overall solution is much more solid now! ✨ 

A short summary of the new commits to make it easier to re-review:

* ca908a75773fcb06ac79118b150a89655b775cd3, 1c327cc3f120f2c35000906eb42d8c7b08fb5f37, 17b7375dfa2505da78ee086624c3abc4772eaa26: Address https://github.com/desktop/desktop/pull/9576#discussion_r416247033 by keeping the list of local tags on the repository state and then calling `refreshTags()` only when needed.
* dc34c824998c2b3f8f858a01ae3b8726faf98ca4: Address https://github.com/desktop/desktop/pull/9576#discussion_r416242040 by comparing the previously calculated pushed tags with the new ones.
* b5997a453fcfe83a97a6bc19d1abe1e5017f9c43: Address  https://github.com/desktop/desktop/pull/9576#discussion_r416242898 by switching to `lru-cache` for the cache.
* b7db5840fb473d2973ca2a2afd1bb5d6a1f634f6: Address https://github.com/desktop/desktop/pull/9576#discussion_r416248725 by only caching for 5 minutes the results when there are git errors.

I've also added 80ece10e1a470346c288f9bb99b984c89bc34b3c and 6a938d5c537b8b28127c96c2b8b547914d7e4cb7 to solve an issue I've just discovered where the tag info was not updated when tags were created or deleted from outside of Desktop.

app/src/lib/stores/helpers/fetch-tags-to-push-memoized.ts
outofambit: ah i missed the part where the _entire_ cache gets cleared after a push/pull/fetch! i'm much less worried, but i think this is a good pattern for us to use in other places :)

app/src/lib/git/tag.ts
outofambit: this looks great! i haven't used `node-lru-cache` before and i'm excited to.

PR #9577, opened 2020-04-20
show proper LFS progress 

### LINKED ISSUE #3171: show proper LFS progress
We made a bunch of progress in https://github.com/desktop/desktop/pull/2355 to get the raw LFS progress information in the app. But it's currently just the raw progress, it'd be good to get this information flowing through to the progress side:

 - clone progress bar will fetch LFS assets after
 - checkout progress can include downloading LFS assets
 - fetch/push/pull also

AndreiMaga:
Closes #3171

## Description

- Passing the LFS output straight to the progress bar

### Screenshots

Capture
Will add more screenshots soon.
## Release notes

Notes: Better LFS output


AndreiMaga:
I just realized that by doing it this way, the only way to get the proper output would be by using ```git lfs ``` commands.

PR #9537, opened 2020-04-16
Display tags in commit list

outofambit:
Closes #9427
(redo of #9488)

### Screenshots

Screen Shot 2020-04-20 at 8 26 29 AM
Screen Shot 2020-04-20 at 8 25 38 AM
Screen Shot 2020-04-20 at 8 26 04 AM
Screen Shot 2020-04-21 at 1 49 51 PM

PR #9578, opened 2020-04-20
Refresh commit list after tagging

### LINKED ISSUE #9555: refresh history commit list after a tag is created in desktop

outofambit:
Closes #9555

_(this branch based off of #9537, please review that first!)_

## Description

- forces the commit list to re-render whenever the store of commit information passed to it changes in a way that would affect the list item rendering
- updated `key` prop for commit list items to allow such re-rendering even when the commit sha has not changed

### Screenshots

tag-now

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -270,9 +270,8 @@ export class GitStore extends BaseStore {
       return getCommit(this.repository, targetCommitSha)
     })
 
-    if (foundCommit != null) {
-      this.commitLookup.set(targetCommitSha, foundCommit)
-      this.emitNewCommitsLoaded([foundCommit])
+    if (foundCommit instanceof Commit) {
+      this.storeCommits([foundCommit], true)
     }
   }
 
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -166,5 +166,5 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
  * Makes a hash of the commit's data that will be shown in a CommitListItem
  */
 function commitListItemHash(commit: Commit): string {
-  return `${commit.sha} ${commit.summary} ${commit.tags}`
+  return `${commit.sha} ${commit.tags}`
 }
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -64,11 +64,7 @@ interface ICommitListProps {
 
 /** A component which displays the list of commits. */
 export class CommitList extends React.Component<ICommitListProps, {}> {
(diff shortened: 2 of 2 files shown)


app/src/lib/stores/git-store.ts
```diff
@@ -270,9 +270,8 @@ export class GitStore extends BaseStore {
       return getCommit(this.repository, targetCommitSha)
     })
 
-    if (foundCommit != null) {
-      this.commitLookup.set(targetCommitSha, foundCommit)
-      this.emitNewCommitsLoaded([foundCommit])
+    if (foundCommit instanceof Commit) {
+      this.storeCommits([foundCommit], true)
```
outofambit, line 274: this method does the same thing as the above lines ☺︎

app/src/ui/history/commit-list.tsx
```diff
@@ -143,10 +151,20 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
             commits: this.props.commitSHAs,
             gitHubUsers: this.props.gitHubUsers,
             localCommitSHAs: this.props.localCommitSHAs,
+            commitLookupHash: this.commitLookupHash([
+              ...this.props.commitLookup.values(),
+            ]),
```
rafeca, line 156: Can we use the commits stored in `this.props.commitSHAs` instead of the ones in `this.props.commitLookup` to invalidate the commit list?

I'm asking because `commitLookup` will contain all the commits that have been loaded in the repository, while `commitSHAs` will have the ones that get actually rendered by the list, so it may be less work.

The only downside is that we'll need to do a lookup for each of them (but that should be really fast since the lookup object is a `Map`)

rafeca, line 156: Can we use the commits stored in `this.props.commitSHAs` instead of the ones in `this.props.commitLookup` to invalidate the commit list?

I'm asking because `commitLookup` will contain all the commits that have been loaded in the repository, while `commitSHAs` will have the ones that get actually rendered by the list, so it may be less work.

The only downside is that we'll need to do a lookup for each of them (but that should be really fast since the lookup object is a `Map`)

app/src/ui/history/commit-list.tsx
```diff
@@ -143,10 +151,20 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
       </div>
     )
   }
 }
+
+/**
+ * Makes a hash of the commit's data that will be shown in a CommitListItem
+ */
+function commitListItemHash(commit: Commit): string {
+  return `${commit.sha} ${commit.summary} ${commit.tags}`
```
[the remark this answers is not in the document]
outofambit: oooo good catch!

app/src/ui/history/commit-list.tsx
outofambit: excellent idea! done in 8922172 :)

PR #8644, opened 2019-11-16
Home/End keys

AndreiMaga:
Closes #6295

## Description

- Added shortcuts for home/end keys
- ```Home``` will jump to the start of the diff
- ```End``` will jump to the end of the diff
- Added shortcuts for ```Cmd+ Up/Down```

## Release notes

Notes: <kbd>Home</kbd> and <kbd>End</kbd> buttons will now work in all lists, also <kbd>Cmd + Up/Down</kbd>will emulate <kbd>Home</kbd>/<kbd>End</kbd>.

PR #9582, opened 2020-04-20
Intuitive LFS Output

### LINKED ISSUE #3171: show proper LFS progress

AndreiMaga:
Created map with the files, and tracking the total progress

Closes #3171

## Description

- Intuitive LFS Output, as it's tracking all the progress from LFS and combining all the states to create the output
- This should be very easy to customize from now on
- Also, no more flickering

### Screenshots

download
I can't add more screenshots as I'm over the data quota
## Release notes

Notes: Intuitive LFS Output


niik:
Hi @AndreiMaga, thanks fo much for working on this! I tested this with https://github.com/niik/lfs-test and found that for pushing the new logic seemed to work really well but for cloning it seemed to produce some unintuitive results



I captured the combined output from LFS and Git clone at https://gist.github.com/niik/48ae2cf1a20f5ff399a12cffcd08d04e in order to take a closer look at what was going on and it seems like the occasional flickering between the indeterminate progress and the regular one is the result of progress lines from Git such as this one:

```
Filtering content:  33% (2/6), 6.00 MiB | 1.72 MiB/s
```

I'm fairly certain these didn't exist when we implemented the LFS support a long time ago.

The second problem is that the download text always says "0/1" which has to do with LFS not being great (at  least  in this case) about estimating the number of files that will get downloaded.

Initially while looking into this problem I figured the solution would be simple enough that I could just do it locally and then request the necessary changes but I'm afraid I got caught up in it and my local changes ended up being quite extensive. The end result, while not perfect, at least addresses the two issues I spotted initially


Would you mind if we took the same approach as in https://github.com/desktop/desktop/pull/8644 and author this fix together?

AndreiMaga:
Hey @niik ,
I don't mind at all if you take the same aproach as the other PR.
And I don't recall the '0/1' problem, maybe that was some bad testing on my part.

niik:
> I don't mind at all if you take the same aproach as the other PR.

Cheers! Here's [my changes](https://github.com/desktop/desktop/pull/9582/files/d401355be998d903208fb8960b13fd3003165ded..b6722d3f68555c236bb69fe41e3b1372c68d0a15), please take a look when you've got a moment and let me know what you think. We'll have to get @rafeca's blessing on this eventually but I'm afraid I've all but broken him with PRs pending review while he was out last week so to set expectations we should expect that it might take a while for this to get reviewed.

AndreiMaga:
I am more than impressed with your changes.
As for the review time, on my part is not a problem. I'm just glad I can learn from you and try to improve my own code.

niik:
Sorry @rafeca, need another review for the merge conflict

PR #9590, opened 2020-04-21
disable link navigation within desktop windows

### LINKED ISSUE #9551: Insecure design -- missing navigation control
_As suggested on H1, opening a public issue instead._

---

GitHub Desktop should implement proper navigation control and block attempts to redirect `BrowserWindow` contents to any other location to prevent RCE in case if the attacker somehow manages to insert a link into the view, for example.

Note that nothing even prevents loading not only remote resources inside the BrowserWindow, but even `http:` resources as opposed to `https:`.

The recommended solution here would be to do all of the following:

* intercept `will-navigate`, `new-window`, `web-contents-created`, and perhaps other events
* intercept requests via session API and consider restricting loading data from thirdparty servers and over insecure protocol

First of those is even in the documented Electron security checklist: https://www.electronjs.org/docs/tutorial/security#12-disable-or-limit-navigation, not sure how you missed that.

## Impact

Exploiting this requires another vuln, but it turns an otherwise low-impact problem (i.e. inserting a link to a thirdparty site) into a full-featured RCE.

## Screenshots

This:
github-sad-remote-0

Leads to this:
github-sad-remote-1

outofambit:
Closes #9551

## Description

prevents link navigation from executing within desktop windows, since we never show content that isn't rendered by desktop itself

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -26,6 +26,7 @@ import { now } from './now'
 import { showUncaughtException } from './show-uncaught-exception'
 import { IMenuItem } from '../lib/menu-item'
 import { buildContextMenu } from './menu/build-context-menu'
+import { sendNonFatalException } from '../lib/helpers/non-fatal-exception'
 
 app.setAppLogsPath()
 enableSourceMaps()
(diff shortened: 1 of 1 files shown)


app/src/main-process/main.ts
```diff
@@ -563,6 +563,12 @@ app.on('web-contents-created', (event, contents) => {
     event.preventDefault()
     log.warn(`Prevented new window to: ${url}`)
   })
+  // prevent link navigation within our windows
+  // see https://www.electronjs.org/docs/tutorial/security#12-disable-or-limit-navigation
+  contents.on('will-navigate', (event, url) => {
+    event.preventDefault()
+    log.warn(`Prevented navigation to: ${url}`)
```
rafeca, line 570: Is this automatically uploaded to sentry or somewhere? Would be good to know when this happens to be able to catch attackers doing nasty things...

outofambit: oo that's a good point! it will not be logged to sentry. i will add that!

outofambit: done in d397ef7!

Issue #9608, opened 2020-04-23
Losing focus while creating a new branch changes what desktop thinks is the default branch.


anarten:
### Describe the bug

While creating a branch from another base branch that is **not** the default branch, if GitHub desktop loses focus it changes the "Create a branch" window to appear like the branch I am basing off of is the default branch. 

### Version & OS

Version 2.4.2
Windows 10 Pro Version 1809 Build 17763

### Steps to reproduce the behavior

1. Make your current branch not the default branch. 
2. Go to menu Branch -> New Branch
3. Click on "Create branch based on..." **currently checked out branch.**
4. Click on another Windows (i.e. not GitHub Desktop) to lose focus from GitHub Desktop.
5. "Create a branch" sub-window in GitHub Desktop will change to state that your current branch is the default branch for the repository. 

A similar issue also occurs by reversing the order of 3 & 4:
1. Make your current branch not the default branch. 
2. Go to menu Branch -> New Branch
3. Click on another Windows (i.e. not GitHub Desktop) to lose focus from GitHub Desktop.
4. With the first click back on GitHub Desktop, hit "Create branch based on..." **currently checked out branch.**
5. "Create a branch" sub-window in GitHub Desktop will change to state that your current branch is the default branch for the repository. 

### Expected behavior

The appearance of what the default branch is should not change. 

### Actual behavior

The appearance of what the default branch is changes to the currently checked out branch. 

### Screenshots

Step 3:
![image](https://user-images.githubusercontent.com/31749093/80115541-18e1ac00-8542-11ea-83b8-59b910c2809a.png)

Step 4 - 5:
![image](https://user-images.githubusercontent.com/31749093/80115575-27c85e80-8542-11ea-98d8-9f7d4cdc59f3.png)

### Logs

N/A

### Additional context

N/A

rafeca:
Thanks for the detailed report @anarten! I've been able to reproduce this problem

PR #9611, opened 2020-04-23
Calculate the start point of the branch on the create branch dialog from the passed properties

rafeca:
Closes https://github.com/desktop/desktop/issues/9608

## Description

This fixes the issue described in https://github.com/desktop/desktop/issues/9608 by calculating the default branch using the passed props instead of the current dialog state.

An alternative would have been to completely get rid of the `defaultBranchAtCreateStart` state property and recalculate that value on every render, but I'm unsure if that could mess up things while the branch is being created, based on this comment on the code:

> Whether or not the dialog is currently creating a branch. This affects
> the dialog loading state as well as the rendering of the branch selector.
>
> When the dialog is creating a branch we take the tip and defaultBranch
> as they were in props at the time of creation and stick them in state
> so that we can maintain the layout of the branch selection parts even
> as the Tip changes during creation.
>
> Note: once branch creation has been initiated this value stays at true
> and will never revert to being false. If the branch creation operation
> fails this dialog will still be dismissed and an error dialog will be
> shown in its place.

### Screenshots

N/A

## Release notes

Notes: [Fixed] Solved issue on the create branch dialog when refocusing the Desktop window.

### DIFF
--- a/app/src/ui/create-branch/create-branch-dialog.tsx
+++ b/app/src/ui/create-branch/create-branch-dialog.tsx
@@ -121,10 +121,15 @@ export class CreateBranch extends React.Component<
     })
 
     if (!this.state.isCreatingBranch) {
+      const defaultStartPoint = getStartPoint(
+        nextProps,
+        StartPoint.UpstreamDefaultBranch
+      )
+
       this.setState({
         tipAtCreateStart: nextProps.tip,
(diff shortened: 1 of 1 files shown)


foofaa222:
> Nice work ✨ Works well on Mac/Win as the dialogue now remains intact.

PR #9637, opened 2020-04-27
Measure number of tags created from GitHub Desktop

rafeca:
Partially addresses https://github.com/desktop/desktop/issues/9612

This PR is based on https://github.com/desktop/desktop/pull/9576

## Description

This PR adds instrumentation of the number of tags that get created from the Desktop UI to let us learn how much is this feature used.

### Screenshots

N/A

## Release notes

Notes: no-notes

PR #9638, opened 2020-04-27
Measure number of created tags outside of Desktop

### LINKED ISSUE #9612: Add instrumentation to tags feature
What data do we want to instrument as part of the tags feature?

- [ ] Number of created tags from GitHub Desktop.
- [ ] Total number of created tags.

-------

part of work for #9424

rafeca:
Closes #9612

This PR is based on https://github.com/desktop/desktop/pull/9637

## Description

This PR adds a new measure to be logged which counts the number of tags that are being created (both from Desktop and from any external program).

This will help us know the percentage of people use the create tag feature so we can plan how to improve it in the future.

An important consideration/known limitation is that we're only going to be able to detect tags created for a repository when that repository is opened in Desktop (since we don't persist in disk the local tags). This means that the total tags measure will only actually contain the number of created tags while a repository is open in Desktop. Even with this limitation, I think that the measure can be useful since it'll give us data about whether users are going to the terminal to create a tag even with Desktop opened.

### Screenshots

N/A

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -99,8 +99,6 @@ const RecentBranchesLimit = 5
 
 /** The store for a repository's git data. */
 export class GitStore extends BaseStore {
-  private readonly shell: IAppShell
-
   /** The commits keyed by their SHA. */
   public readonly commitLookup = new Map<string, Commit>()
 
@@ -109,8 +107,6 @@ export class GitStore extends BaseStore {
   private _history: ReadonlyArray<string> = new Array()
 
   private readonly requestsInFight = new Set<string>()
-
-  private readonly repository: Repository
 
   private _tip: Tip = { kind: TipState.Unknown }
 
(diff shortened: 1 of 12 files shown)


app/src/lib/stores/git-store.ts
```diff
@@ -145,11 +142,12 @@ export class GitStore extends BaseStore {
 
   private _stashEntryCount = 0
 
-  public constructor(repository: Repository, shell: IAppShell) {
+  public constructor(
+    private readonly repository: Repository,
+    private readonly shell: IAppShell,
+    private readonly statsStore: StatsStore
```
rafeca, line 148: I've added the `statsStore` as a parameter to the `gitStore`. This has caused a bunch of changes in tests but couldn't find an easier way to record the added tags (an alternative would have been to recalculate the difference between the previous state and the new state on the `appStore`, but I prefered to keep that logic on a single point.

outofambit: ugh. yeah i think that's the best solution for this without refactoring the entire interface for the stats store.

the only other place i could think of would be [`AppStore.OnGitStoreUpdated`](https://github.com/desktop/desktop/blob/1f15e696163ecad0f3e5a797492317d6d5977d8f/app/src/lib/stores/app-store.ts#L744), but it would involve a whole comparison iteration thing and we don't do any stats updating there already, so 🚢 it.

outofambit:
rebasing this on development to fix appveyor reporting

Issue #9396, opened 2020-03-28
Desktop not compiling if directory path contains a space


ruggi99:
### Describe the bug

If path of Desktop repo contains a space, then Desktop is not compiled, returning an error.

### Version & OS

Github Desktop: development branch, OS: Windows 10

### Steps to reproduce the behavior

1. Compile Desktop with `yarn build:dev`

### Expected behavior

Compiling without error.

### Actual behavior

Generates an error (image below) during compile.

### Screenshots

![Cattura](https://user-images.githubusercontent.com/25064808/77831889-26b63400-7132-11ea-9283-146abb9d1289.PNG)

### Additional context

If path doesn't contain spaces (in my case) it compiles correctly.
If the error isn't repairable, I suggest to add a note in "docs/contributing/setup.md".
Or add the same note into troubleshooting section.

rafeca:
Thanks for the report! I've been able to reproduce the issue on my Window 10 VM

ruggi99:
How will this issue be fixed?

rafeca:
@ruggi99 I'm doing some quick investigation since the error comes from `electron-packager`: I'm checking whether this is something that we can fix in Desktop or it's a bug on `electron-packager`. I'll give more details soon.

rafeca:
@ruggi99 looks like this is a known problem with `electron-packager` (see https://github.com/electron/electron-packager/issues/1120).

The issue got fixed a few days ago by https://github.com/electron/electron-packager/pull/1139, once a new version gets published with the fix we'll upgrade to it.

ruggi99:
@rafeca thank you for investigating this issue.

ruggi99:
@rafeca how about opening a PR to update only the cross-zip dependency and close this issue? I know it's not the best practice to edit the `yarn.lock` file, but I think it won't break anything.

rafeca:
Sounds good to me as a temporary fix, so feel free to send the PR!

Note that `electron-packager` is moving away from `cross-zip` to `extract-zip` (see https://github.com/electron/electron-packager/pull/1139), but either way upgrading `cross-zip` to v3.1.0 should be fine now.

PR #9640, opened 2020-04-27
Windows: make it safer to Open in Command Prompt (in cmd.exe)

dscho:
Not associated with any open issue.

## Description

Just like https://github.com/git-for-windows/MINGW-packages/commit/82172388b, we would like a command prompt that runs `cmd.exe` _not_ to pick up any `git` executable/script in the current working directory.

At the same time, we also do not want `Open in Command Prompt` to execute a `cmd.bat` in the current directory, so we'll safeguard against that, too.

### Screenshots

image

Note that there is a freshly-built `git.exe` in my current working directory, and due to the missing `.dll` files (they are in `C:\git-sdk-64\mingw64\bin`, not in `C:\git-sdk-64\usr\src\git`), it does not even start, but instead fails with code `-1073741515` (which is `0xC0000135` in hex, translating to `STATUS_DLL_NOT_FOUND`, see [here](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55) for further details).

## Release notes

Notes:

### DIFF
--- a/app/src/lib/is-git-on-path.ts
+++ b/app/src/lib/is-git-on-path.ts
@@ -1,12 +1,13 @@
-import { spawn } from 'child_process'
+import { spawn, SpawnOptions } from 'child_process'
 import * as Path from 'path'
 
 function captureCommandOutput(
   command: string,
-  args?: string[] | undefined
+  args?: string[] | undefined,
+  options: SpawnOptions = {}
 ): Promise<boolean> {
   return new Promise<boolean>((resolve, reject) => {
-    const cp = spawn(command, args)
+    const cp = spawn(command, args, options)
 
     cp.on('error', error => {
       log.warn(`Unable to spawn ${command}`, error)
@@ -40,7 +41,7 @@ export function isGitOnPath(): Promise<boolean> {
     // is found, one per line, and return 0, or print an error and
     // return 1 if it cannot be found
     log.info(`calling captureCommandOutput(where git)`)
-    return captureCommandOutput(wherePath, ['git'])
+    return captureCommandOutput(wherePath, ['git'], { cwd: windowsRoot })
   }
 
   if (__LINUX__) {
--- a/app/src/lib/is-git-on-path.ts
+++ b/app/src/lib/is-git-on-path.ts
@@ -5,33 +5,38 @@ function captureCommandOutput(
   command: string,
   args?: string[] | undefined,
   options: SpawnOptions = {}
-): Promise<boolean> {
-  return new Promise<boolean>((resolve, reject) => {
+): Promise<string | undefined> {
+  return new Promise<string | undefined>((resolve, reject) => {
     const cp = spawn(command, args, options)
 
     cp.on('error', error => {
       log.warn(`Unable to spawn ${command}`, error)
-      resolve(false)
+      resolve(undefined)
+    })
+
+    const chunks = new Array<Buffer>()
+    let total = 0
+
+    cp.stdout.on('data', (chunk: Buffer) => {
+      chunks.push(chunk)
+      total += chunk.length
     })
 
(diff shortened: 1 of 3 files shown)


app/src/lib/is-git-on-path.ts
```diff
@@ -1,47 +1,75 @@
...
+    return Promise.resolve(true)
   }
 
-  return Promise.resolve(false)
+  return findGitOnPath().then(
+    (path: string | undefined) =>
+      new Promise<boolean>((resolve, reject) => {
+        resolve(path !== undefined)
+      })
+  )
```
outofambit, line 74: we use `async`/`await` pretty liberally in the desktop codebase, so i think this could be simplified to
```suggestion
  return (await findGitOnPath()) !== undefined
```

app/src/lib/is-git-on-path.ts
```diff
@@ -1,47 +1,75 @@
...
+  return Promise.resolve(undefined)
+}
 
-      // `which` will return 0 when the executable
-      // is found under PATH, or 1 if it cannot be found
-      process.on('close', function(code) {
-        resolve(code === 0)
-      })
-    })
+export function isGitOnPath(): Promise<boolean> {
```
outofambit, line 61:
```suggestion
export async function isGitOnPath(): Promise<boolean> {
```

app/src/lib/is-git-on-path.ts
dscho: Right, I would prefer that, too, but as I was only refactoring, I tried to keep the style as close to the original as possible: https://github.com/desktop/desktop/blob/1123057da3728015853bb8ac859e296575bc4ef7/app/src/lib/is-git-on-path.ts#L14-L31

If you are okay with changing the style while refactoring, I will be delighted to make that change!

dscho: Oh right, since you suggested that change, I figure this project does not want the commits to be rebased and squashed, so I'll just accept your suggestion. Thanks!

dscho:
> @dscho this looks good to me! left a couple code style suggestions.

Thank you! I accepted both suggestions as-are.

> @tierninho would you mind giving this a whirl to smoke test using PowerShill and Command Prompt on Windows? thanks!

I like the nickname "PowerShill" ;-)

@tierninho please let me know if I can help in any way?

tierninho:
Currently blocked from testing this due to https://github.com/desktop/desktop/issues/9396

dscho:
> Currently blocked from testing this due to #9396

Thank you for letting me know!

tierninho:
@dscho Got the Windows issue resolved. I am confirming that both PowerThrill and Commando Prompt are functional is Desktop when opening a repo directory. 

Let me know if there was a specific use case you needed tested. Thanks.

dscho:
Yes, try opening `cmd` or PowerShell (both with and without these here patches) when there is a `cmd.bat` or `powershell.bat` in the worktree. Thank you!

tierninho:
So here are the results, which look to be sufficient. Let me know otherwise.

When testing against this PR:
- Command Prompt successfully opened a repository without executing a `cmd.bat` script in the working directory.
- Powershell successfully opened a repository without executing a `powershell.bat` script in the working directory.

When testing against our current Desktop version:
- Command Prompt failed to open.
- Powershell successfully opened a repository but executed the `powershell.bat` script in the working directory.

dscho:
@outofambit excellent!

dscho:
> @outofambit excellent!

Whoops, I meant: @tierninho excellent!

dscho:
fireworks

outofambit:
@dscho welcome and congratulations! 😜 

welcome

PR #9652, opened 2020-04-28
Add CodeQL Analysis workflow

jhutchings1:
Adds a GitHub Actions workflow that runs CodeQL on every push, and on a daily schedule.

Code scanning looks for vulnerabilities, such as XSS, SQL injection, etc., in your code. If it finds any new vulnerabilities it surfaces them in the PR as check annotations, and blocks the build until they’re fixed or marked as false positives.

(This functionality is being tested on GitHub Desktop and is not available for all other repositories at this time.)


outofambit:
Hi @jhutchings1! Is this PR still in progress? If not, could we close it or mark it a draft until it's ready? Thank you!

jhutchings1:
@outofambit Nope, this is good to go if you're happy to merge it.

outofambit:
@jhutchings1 it looks like the build is failing due to linting, could you take a look at that please? should i be able to see the results of this CodeQL action somewhere? how do we verify its working? i don't see them in the check suite or in the actions tab.

niik:
Hey @jhutchings1, sorry for jumping in so late, I'm just getting up to speed here. I was looking through the action file and the build output from the code scanning action here and I was curious how exactly CodeQL was analyzing our TypeScript files, perhaps you could shed some light on that? From 👇 one might suspect that it's only running against our javascript files. The few JS files we have are all either config files or lint rules neither of which end up as part of the end user build.

image

Does the JavaScript codeql support also support TypeScript behind the scenes? If it doesn't and runs solely on the pure JS in our repository then I think we need to discuss the cost vs reward of adding this action.

If it does support TypeScript it would be fantastic if you could provide some sort of sentinel code/dangerous pattern that we could apply to our codebase to see if codeql picks it up.

jhutchings1:
> Does the JavaScript codeql support also support TypeScript behind the scenes? If it doesn't and runs solely on the pure JS in our repository then I think we need to discuss the cost vs reward of adding this action.

Yep, it does! 

> If it does support TypeScript it would be fantastic if you could provide some sort of sentinel code/dangerous pattern that we could apply to our codebase to see if codeql picks it up.

We've got a bunch of sample vulnerabilities in the codeql repository. You can definitely add one of those, although I'll note the examples are in vanilla JS. https://github.com/github/codeql/tree/master/javascript/ql/src/Security

PR #9655, opened 2020-04-28
no more branch filtering for appveyor

outofambit:
we have had a configuration setting in our appveyor.yml that tells appveyor to only build specific branches in the desktop repo. i have no idea why this has worked for us for so long, but it's clearly not anymore.

appveyor offers very few tools to debug your config, so this is the simplest thing that should just work. this may increase the number of appveyor builds beyond what it was before this broke, so we should keep an eye on that.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,13 +11,6 @@ cache:
   - .eslintcache
   - '%LOCALAPPDATA%\Yarn\Cache\v4'
 
-branches:
-  only:
-    - development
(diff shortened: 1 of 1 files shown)


outofambit:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FPWwC2oE1oipEs%2Fgiphy.gif&token=gohome" data-gitmeme-token="gohome"></a>

Issue #9231, opened 2020-03-06
GitHub Desktop authentication primarily through browser flow


billygriffin:
GitHub is [deprecating and removing the authorizations API](https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/), which handles how most users of GitHub Desktop sign into GitHub from the app. We've always had a secondary method of authenticating via the browser, but we'll need to make this the main authentication method going forward.

### Considerations

1. We know protocol handlers can be brittle depending on browser and OS, so we need to ensure people know that the sign in was successful and that they can return to the app. We're imagining this as an interstitial after a successful sign in on github.com.
2. To help discover potential pitfalls with moving to 100% browser authentication, we might start by swapping the prominent default initially so people have something to fall back to before we're forced to remove it altogether.
3. We need to make sure the next version of GitHub Enterprise uses both our protocol handlers (including the auth-specific one) so we get backward compatibility on Enterprise.

steven351:
make things work easier

gitwdel:
#9231

ampinsk:
# Designs 

## Transition state

For the period of time when we're transitioning, we'll need to emphasize the browser log in flow, while still keeping the normal login flow

### Welcome screen

This moves all the login options to the first screen, and emphasizes the GitHub.com option.

<img width="1440" alt="Transition-Welcome-Screen" src="https://user-images.githubusercontent.com/10404068/77484135-fe9f9b80-6de6-11ea-8d61-ae8ce6cedd11.png">

### Preferences
The initial login screen should stay the same: 

<img width="1440" alt="Transition-Preferences-1" src="https://user-images.githubusercontent.com/10404068/77484139-019a8c00-6de7-11ea-869e-80cd12d6eaed.png">

Then on this screen, I flipped the options and emphasized the GitHub.com option.

<img width="1440" alt="Transition-Preferences-2" src="https://user-images.githubusercontent.com/10404068/77484141-02332280-6de7-11ea-9de6-498124874a70.png">


## Final state

For when the transition is complete, and the *only* option is to login using your browser

### Welcome screen

After the transition is complete, we can remove the login fields

<img width="1440" alt="Final-Welcome-Screen" src="https://user-images.githubusercontent.com/10404068/77484142-02cbb900-6de7-11ea-9f49-25143d5428b2.png">

### Preferences

After the transition is complete, we can just surface the GitHub.com option directly on this screen.

[...]

PR #9663, opened 2020-04-29
Emphasize sign in via browser on welcome screen

rafeca:
Addresses the first part of https://github.com/desktop/desktop/issues/9231

## Description

This PR implements the transition phase outlined in https://github.com/desktop/desktop/issues/9231#issuecomment-603545527 from the user/pass authentication to the browser authentication.

After talking to @ampinsk, we decided to tweak a bit the designs from https://github.com/desktop/desktop/issues/9231#issuecomment-603545527 to make the implementation much easier. The main change is that instead of showing the user/password form directly on the welcome screen, we're showing a link to sign in via user/pass which takes the user to a separate screen where they can enter their credentials.

The other important change that has been made is changing the widths of the 2 columns of the welcome screen: they are currently 50/50 but in this PR I've changed it to 60/40 to acccomodate the extra horizontal space.

Here's is the difference between the two widths when opening Desktop for the first time (the initial window has a width of 960px).

**Before: **

**Now: **

@ampinsk what do you think about this change? maybe we can change a bit the right background image to fill the full height... (or keep the previous proportions and figure out a different solution).

## Screenshots

Welcome screen:

image

Loading welcome screen (after pressing the sign in via browser button:

image

Screen showing the soon-to-be-deprecated user/password authentication:

image

Sign in screen via the settings dialog:

image

## Release notes

Notes: [Improved] Emphasize the sign in via browser flow when opening Desktop for the first time.

### DIFF
--- a/app/src/ui/lib/authentication-form.tsx
+++ b/app/src/ui/lib/authentication-form.tsx
@@ -88,7 +88,6 @@ export class AuthenticationForm extends React.Component<
     const disabled = this.props.loading
     return (
       <>
-        <hr />
         <TextBox
           label="Username or email address"
           disabled={disabled}
@@ -139,8 +138,13 @@ export class AuthenticationForm extends React.Component<
   }
 
   private renderSignInWithBrowser() {
+    if (this.props.endpoint === getDotComAPIEndpoint()) {
+      return
+    }
+
     return (
       <>
+        {this.props.supportsBasicAuth && <hr />}
         {this.renderAuthIntroMessage()}
 
         <Button
--- a/app/src/ui/lib/authentication-form.tsx
+++ b/app/src/ui/lib/authentication-form.tsx
@@ -74,9 +74,8 @@ export class AuthenticationForm extends React.Component<
   public render() {
     return (
       <Form className="sign-in-form" onSubmit={this.signIn}>
-        {this.renderUsernamePassword()}
-
         {this.renderSignInWithBrowser()}
+        {this.renderUsernamePassword()}
       </Form>
     )
   }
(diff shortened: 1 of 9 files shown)


rafeca:
> When the browser authenticates, and as you head back to Desktop, this shows for a brief moment right before Desktop completes the login process.

Good catch! gonna fix it!

rafeca:
@tierninho fixed in 9b53f7d0d

app/src/ui/lib/authentication-form.tsx
```diff
@@ -139,37 +138,38 @@ export class AuthenticationForm extends React.Component<
...
-        <Octicon symbol={OcticonSymbol.linkExternal} />
-      </LinkButton>
-    )
-
-    const browserSignInButton = (
-      <Button type="submit" onClick={this.signInWithBrowser}>
-        Sign in using your browser
-      </Button>
-    )
+    if (this.props.endpoint === getDotComAPIEndpoint()) {
```
outofambit, line 141:
```suggestion
    // we don't render this here because the user will have already
    // had the option to sign in via the browser earlier in the sign-in flow
    if (this.props.endpoint === getDotComAPIEndpoint()) {
```

app/src/ui/lib/authentication-form.tsx
```diff
@@ -139,33 +138,28 @@ export class AuthenticationForm extends React.Component<
...
-        {basicAuth ? null : this.renderEndpointRequiresWebFlow()}
-
-        <div className="sign-in-footer">
-          {basicAuth ? browserSignInLink : browserSignInButton}
-          {basicAuth ? null : this.renderActions()}
-        </div>
-      </div>
+      <>
+        {this.props.supportsBasicAuth && <hr />}
+        {this.props.supportsBasicAuth && this.renderEndpointRequiresWebFlow()}
```
rafeca, line 150: This was wrong, the previous code did the opposite 🤦‍♂️

outofambit: omg 🙀

Issue #9424, opened 2020-03-31
[Tracking issue] Add, push, and view Git tags in Desktop


outofambit:
Allow a desktop user to:
* create a new git tag
* push tags to remote(s)
* view tags on commits

_more info to come_

original reported issue: https://github.com/desktop/desktop/issues/4829

Mishzs:
تعال ايمو ي عيني

billygriffin:
I think we can close this now that it's all shipped to beta! 🎉

PR #9667, opened 2020-04-29
Show unpushed indicator on commits that have unpushed tags

rafeca:
Related issue: https://github.com/desktop/desktop/issues/9424

## Description

This PR uses the data about unpushed tags to display the unpushed indicator on the commits that have unpushed tags (even if these commits are already in the remote).

Note that only annotated tags are taken into account for the unpushed indicator (since lightweight tags are never pushed to the remote by Desktop).

### Screenshots

demo

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -129,7 +129,7 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
     numUnpushedTags: number
   ) {
     if (isLocalCommit) {
-      return "This commit has not been pushed to the remote repository"
+      return 'This commit has not been pushed to the remote repository'
     }
 
     if (numUnpushedTags > 0) {
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -60,6 +60,9 @@ interface ICommitListProps {
 
   /* Whether the repository is local (it has no remotes) */
   readonly isLocalRepository: boolean
+
+  /* Tags that haven't been pushed yet. This is used to show the unpushed indicator */
+  readonly tagsToPush: ReadonlyArray<string> | null
 }
 
 /** A component which displays the list of commits. */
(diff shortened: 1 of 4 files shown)


rafeca:
Thanks for the finding @tierninho ! I'm gonna fix the tooltip and fix the indicator on a separate PR!

rafeca:
I've fixed the tooltip text in b95e4d8a11305e02743032bc6e7625a029dc7d3c: now it'll display the following messages:

* **This commit hasn't been pushed to the remote repository yet**: This is shown when the commit hasn't been pushed, regardless of whether it has unpushed tags
* **This commit has XXX tags to push**: Shown when the commit has been pushed but 1 or some of its tags haven't.

cc/ @billygriffin , @ampinsk , @nerdneha for these texts 👆

app/src/ui/history/commit-list.tsx
```diff
@@ -97,19 +97,23 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
+    ).length
 
     const showUnpushedIndicator =
-      (isLocal || hasUnpushedTags) && this.props.isLocalRepository === false
+      (isLocal || numUnpushedTags > 0) && this.props.isLocalRepository === false
 
     return (
       <CommitListItem
-        key={commitListItemHash(commit)}
+        key={commit.sha}
```
rafeca, line 109: @outofambit I've changed this `key` to just use the commit sha and modified the `shouldComponentUpdate` function on the `CommitListItem` component to take into account them.

I think it's better to just re-render the component when the tags change rather than unmounting/mounting.

Related to that, I think that we should remove the `shouldComponentUpdate` method from the `commitListItem` (it's not obvious that any new property that gets added to `commitListItem` has to be added there) and instead make the `CommitListItem` a `PureComponent` (since we use immutable data structures everywhere this should be fine). I can prepare a separate PR for that.

rafeca: Created https://github.com/desktop/desktop/pull/9672

app/src/ui/history/commit-list.tsx
```diff
@@ -110,6 +124,23 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
     )
   }
 
+  private getUnpushedIndicatorTitle(
+    isLocalCommit: boolean,
+    numUnpushedTags: number
+  ) {
+    if (isLocalCommit) {
+      return "This commit hasn't been pushed to the remote repository yet"
```
billygriffin, line 132:
```suggestion
      return "This commit has not been pushed to the remote repository"
```

rafeca: Thanks @billygriffin !

PR #9680, opened 2020-04-30
Add GitHub Actions

outofambit:
_[...]_


shiftkey:
@outofambit I noticed a similar issue on Azure Pipelines recently, and added https://github.com/shiftkey/desktop/commit/42f219c35bb2890236622eaff5778271e1129af3 to workaround the issue

app/test/__mocks__/electron.ts
```diff
@@ -15,6 +15,11 @@ export const remote = {
   autoUpdater: {
     on: jest.fn(),
   },
+  nativeTheme: {
+    addListeners: jest.fn(),
```
outofambit, line 19:
```suggestion
    addListener: jest.fn(),
```

amir4toshiba (dismissed):

rafeca:
I've started seeing build failures in the `yarn check-modified` step on GitHub Actions under Windows, it seems like the `app/static/common/gitignore` is causing a dirty tree due to the symlinks that it has:

After checkout, this is what we got:

```
$ git diff
diff --git a/app/static/common/gitignore b/app/static/common/gitignore
--- a/app/static/common/gitignore
+++ b/app/static/common/gitignore
@@ -1 +1 @@
-Subproject commit 2644536a24f37d3906c7ea56e57f6965b1e21f36
+Subproject commit 2644536a24f37d3906c7ea56e57f6965b1e21f36-dirty

$ cd app/static/common/gitignore && git status
HEAD detached at 2644536
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Clojure.gitignore
	modified:   Fortran.gitignore
	modified:   Global/Octave.gitignore
	modified:   Kotlin.gitignore

no changes added to commit (use "git add" and/or "git commit -a")
```

Note that the modified files correspond to the symlinks in the original repo: https://github.com/github/gitignore

Link to the Run: https://github.com/desktop/desktop/runs/897975189#step:7:40

I'm investigating why this didn't fail before (https://github.com/desktop/desktop/runs/854464821) but now it does... 🤔

rafeca:
Ok, looks like the Windows image has been changed and it's quite possible that an upgraded version of `git` or another dependency is causing this change.

* In 48fe311 (a build that passed) the virtual environment [was v20200630.0](https://github.com/desktop/desktop/runs/854464821#step:1:8) ([list of installed software](https://github.com/actions/virtual-environments/blob/win19/20200630.0/images/win/Windows2019-Readme.md)).
* In 00d8c32 (a build that failed) the virtual environment [was 20200714.1](https://github.com/desktop/desktop/runs/897975189#step:1:8) ([list of installed software](https://github.com/actions/virtual-environments/blob/win19/20200714.1/images/win/Windows2019-Readme.md)).

These are the changes between the two virtual environment versions: https://github.com/actions/virtual-environments/commit/f7c1a0264ff7112855c9c19c48dfbdb42526a7c1 and https://github.com/actions/virtual-environments/commit/95222e07e94ec86d400c670bdc2f92f5a2b20fd0 .

The only suspecting thing that I see is "Enabled Windows Subsystem for Linux", but even this seems unlikely to cause this problem, I'm going to keep investigating.

Anyways, I think that in order to keep our builds stable and consistent, we should pin to specific VMs and not use latest, going to investigate this as well.

rafeca:
To verify that the failure is caused by the change of the VM image that GitHub Actions is using I tried re-running a job on an old commit that used to have CI passing (f4aace4).

* Old run that passed with the previous VM: https://github.com/desktop/desktop/runs/854271314
* New run that failed with the current VM: https://github.com/desktop/desktop/runs/898140396

I cannot find any information about pinning our CI runs to a specific version of the VM (in the [official docs](https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#ip-addresses-of-github-hosted-runners) it mentions using either `windows-latest` or `windows-2019`, implying that both strings will use the latest available VM for Windows 2019.

rafeca:
Oh god, this was a tricky one to debug! turns out that `core.symlinks` was enabled by default in the VMs recently (see https://github.com/actions/virtual-environments/pull/1186). Since there was no reference to this in the VM release changelogs I couldn't find this by myself.

Long story short, it looks like Git for Windows has some issues when `core.symlinks` is enabled on repositories with submodules. I'm going to create an issue on that repository.

rafeca:
Ok, I've added a few commits that hopefully will make this PR green again:

- d57b390 and 033cb0c revert the failures that were added in purpose to test that CI was reporting errors correctly.
- ce6228b enables GitHub Actions on pushes to release branches.
- 9398db9 adds the missing secrets as environment variables for the publish step.
- 853dfea pins the VM version (only to the major version of the VM but that's better than nothing).
- 1d6b7e9 disables symlink support on the Windows runs (this is equivalent to what happened in AppVeyor).

With this, I hope that the build turns green 🤞

Two things to take into account:

- Would be good to enable symlinks support on Windows. Since we don't have it enabled there's a bug in Windows where selecting certain `gitignore` templates (like Kotlin) when creating a new repo in Windows results in wrong contents of the `.gitignore` file. This is a long-standing bug and I don't think we can fix it properly without a fix in Git for Windows.
- Would also be good to pin to a specific VM version to have more stable builds in CI. Unfortunately, after talking to the GitHub Actions team this is only possible when using our own self-hosted VMs for CI. We don't want to go through that path (at least for now) so we'll have to live with VMs that self-update automatically.

rafeca:
I've added a few more commits to fix publishing releases. With this, I've been able to release successfully `v2.5.4-test3`.

Some things I'd like to improve from the CI setup:

1. Run the tests before publishing the release: We currently run the tests afterwards, which doesn't make a lot of sense 😅 
2. Only run the `yarn publish` step on releases branch: we currently run it always and the code decides whether to do something or not.
3. Only pass the secrets when building a release: we currently always pass the secrets, so it would be better to have a dedicated workflow for releasing and only pass the secrets in that workflow.

rafeca:
> Long story short, it looks like Git for Windows has some issues when core.symlinks is enabled on repositories with submodules. I've created a small repository that reproduces these issues (https://github.com/rafeca/test-submodules-with-symlinks) and I'm planning to create an issue in the git-for-windows repository.

I've checked with the newly released v2.28.0 version of git for windows and these issues are fixed :tada:, this means that once the GitHub Actions VMs get updated with this new git version (which is happening very very soon: https://github.com/actions/virtual-environments/pull/1349 😃 ), we'll be able to revert 1d6b7e90483ee7c2548516116fae74a825ab670e.

app/test/__mocks__/electron.ts
VChernomYA: -    addListeners: jest.fn(),
+    addListener: jest.fn(),

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - development
+      - releases/*
```
rafeca, line 7: I don't think we need to explicitly build the `releases/*` branches, I added this while testing...

```suggestion
```

rafeca (dismissed):
LGTM

.github/workflows/ci.yml
niik: Yeah, that makes sense to me

niik (approved):

rafeca (approved):
LGTM! :shipit:

outofambit:
OH EM GEE

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FL0ZuUuWbOScpA3No8f%2Fgiphy.gif&token=oh_em_gee" data-gitmeme-token="oh_em_gee"></a>

SecretRose7 (commented):
FiireFox7_RoseHub Enterprise software systems are the best way to go

PR #9683, opened 2020-05-01
release 2.4.4-beta2

outofambit:
- [ ] notes
- [ ] release

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -1,5 +1,9 @@
 {
   "releases": {
+    "2.4.4-beta2": [
+      "[Fixed] Enable tags feature flags for beta - #9682",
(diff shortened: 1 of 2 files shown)


changelog.json
```diff
@@ -1,5 +1,9 @@
 {
   "releases": {
+    "2.4.4-beta2": [
+      "[Fixed] Enable tags feature flags for beta - #9682",
```
nerdneha, line 4: This is internal stuff right? So I think we can just nix this -- what do you think? otherwise we should just say [Fixed] show unpushed... so basically what we had intended to do but it's "fixed" from the user's perspective

outofambit:
```suggestion
      "[Fixed] Display tags in commit list and show tag creation option - #9682",
```

PR #9688, opened 2020-05-02
Fix Desktop Welcome 2FA Confirmation Step

### LINKED ISSUE #9686: 2fa not working with new login flow
When attempting to login with 2fa, the screen goes blank after entering in the 6-digit auth code from my Google Auth app. 401 and 410 errors in the console. Can reproduce in Mac/Win, v2.4.4-beta2.

- Start Desktop as a new user and select the user/pw link. 
- Enter credentials and then 2fa code.
- Screen goes partial blank and no browser prompt

Screen Shot 2020-05-01 at 10 59 06 AM

Note the 2fa flow in the Preferences works fine, but I still spot the same console errors:

Screen Shot 2020-05-01 at 11 28 10 AM

outofambit:
closes #9686

## Description

this was an interesting collision of conditions. the issue was not 2FA verification itself, it was the next step in the flow, which should have been a "you must sign in via the browser" message. that flow is restored here. There is also some refactoring of the logic within the auth form component here since that contributed significantly to this bug.

(this video is a little chopped up, because i skipped places where i entered sensitive account info)
welcome-ghd

## Testing

This required some build modifications (using non-dev client id and secrets) to reproduce and test without another beta, so I've verified this locally. @tierninho has agreed to test this flow more in the next beta release.

### DIFF
--- a/app/src/ui/lib/authentication-form.tsx
+++ b/app/src/ui/lib/authentication-form.tsx
@@ -72,17 +72,9 @@ export class AuthenticationForm extends React.Component<
   }
 
   public render() {
-    const content = this.props.supportsBasicAuth ? (
-      this.renderEndpointRequiresWebFlow()
-    ) : this.props.endpoint === getDotComAPIEndpoint() ? (
-      this.renderUsernamePassword()
-    ) : (
-      <>
-        {this.props.endpoint !== getDotComAPIEndpoint() &&
-          this.renderSignInWithBrowser()}
-        {this.renderUsernamePassword()}
-      </>
-    )
+    const content = this.props.supportsBasicAuth
+      ? this.renderEndpointRequiresWebFlow()
+      : this.renderSignInForm()
 
     return (
       <Form className="sign-in-form" onSubmit={this.signIn}>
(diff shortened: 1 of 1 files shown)


app/src/ui/lib/authentication-form.tsx
```diff
@@ -72,19 +72,18 @@ export class AuthenticationForm extends React.Component<
   }
 
   public render() {
+    const content = this.props.supportsBasicAuth
+      ? this.renderEndpointRequiresWebFlow()
+      : this.renderSignInForm()
```
rafeca, line 77: Shouldn't this be the opposite? (show the message requiring web flow auth when the endpoint doesn't support basic auth).

(I think I introduced this bug in https://github.com/desktop/desktop/pull/9663/files#r419732770 😅 🤡 ). 

```suggestion
    const content = this.props.supportsBasicAuth
      ? this.renderSignInForm()
      : this.renderEndpointRequiresWebFlow()
```

outofambit: fixed in ac2e8c4!

Issue #2402, opened 2017-08-05
Allow discarding of individual lines in a file


mikeabiezzi:
### Description

It would be convenient to discard individual lines in a file. Currently, I have to first commit the lines I want and then discard the whole file. Sometimes I'm not ready to commit lines, but I would like to get rid of unwanted lines while I'm reviewing the diff.

### Version

**GitHub Desktop version:** 0.7.2

**OS version:** macOS 10.12.6

joshaber:
Huh, I'm surprised we don't already have an issue for this. Thanks @mikeabiezzi!

JosephTLyons:
+1, Atom's GitHub package allows for lines or chunks to be discarded, would love to see that in GitHub Desktop.

Janpot:
@billygriffin As requested, adding these use cases to the ticket:

1.
Right before I commit I usually like to review all my changes in github desktop and basically do a code review of my own commit. Sometimes there are things in there, like console.log that I'd like to revert, or other things that I forgot to clean up. I do want them reverted instead of selected, because I still want to run tests or verify my code still works before actually committing the changes.

2.
Sometimes when working on a change, I notice I made some wrong assumptions about the code halfway, or figure out a better way to do things, and would like to revert some changes and redo a part. This could also be solved by deselecting, committing, reverting what's left and undoing the commit. But the annoying thing about that approach is that visually, those lines that I want reverted are still there, and that makes it hard to properly review the code I'd like to keep during the process.

carldunham:
Why is this closed? Has it been addressed? #3225 was closed referring dismissively to this issue, now it's closed. Huh?

Gama11:
@carldunham Not sure what you mean, [this issue is not closed](https://i.imgur.com/jtcHeqL.png).

carldunham:
Sorry, I had a dumb.

carldunham:
Ironically, having this same conversation about Gitkraken at https://stackoverflow.com/questions/55368879/how-to-discard-line-in-gitkraken/

billygriffin:
As an update, I think this addition makes a ton of sense, but is non-trivial in terms of design and implementation and is unlikely to be prioritized in the very near future. However, it's absolutely something we think would be a great improvement and we definitely want to support it at some point in the future.

mnaoumov:
https://github.com/gitextensions/gitextensions had this feature since forever. Just saying...

JohannSig:
[...]

PR #9399, opened 2020-03-29
Implementing Discard Unselected Changes

hottestchilipepper:
Closes #2402

## Description
I implemented 'Discard Unselected Changes'.
You can see the 'Discard unselected changes' menu in Right-click menu which is shown when you do Right-click a file in Changes View.
I knew that many people want to discard only some changed lines from their changed files. I think this implementing is not exactly same as the feature which is described at #2402 , but if you use this feature, result will be same as you want.
If you want to discard lines, select a changed file and select all lines what you want to remain, deselect lines what you want to discard. Then right-click the file in Changes View and click 'Discard unselected changes'.
Warning! Your discarding will not be able to recover! Please take care.

This is very useful. I did use also this new feature to make a commit for this implementing.

I'm first in TypeScript and React. So please review meticulously.
Any suggestions and advices can be helpful to me.
Thanks.
-

### Screenshots

K-025
K-030
K-028
K-029

## Release notes

Notes: Discard Unselected Changes Added

Issue #5847, opened 2018-10-08
toPlatformCase component helper


outofambit:
## `toPlatformCase`

In #5744 we added `toPlatformCase`, a helper function to change this pattern:

https://github.com/desktop/desktop/blob/3feeaac7ba184aacdeb21aadc85587b29e3f36f6/app/src/ui/merge-conflicts/merge-conflicts-warning.tsx#L31

to this:

https://github.com/desktop/desktop/blob/3feeaac7ba184aacdeb21aadc85587b29e3f36f6/app/src/ui/discard-changes/discard-changes-dialog.tsx#L69

**The idea is that `toPlatformCase` will case the string properly depending on the current platform.** macOS is title case and windows is sentence case _generally_.

The original pattern appears in about 85 places in Desktop's React components, according to a quick search of the codebase. (Currently `toPlatformCase` is only used in one component.)

## Limitations

There are limitations to this approach because there are a number of edge cases that extend beyond simply title casing or sentence casing a string.

1. Prepositions (https://github.com/desktop/desktop/pull/5841#pullrequestreview-162478757), which are lower case, even in title case.
2. Platform-specific terminology (#2103), like trash or recycling bin
3. Key shortcuts (#2103), which differ on platforms

## The Question

Options:
* keep `toPlatformCase` and spread in the codebase (where applicable)
* remove `toPlatformCase` and continue using string literals in our components
* refactor `toPlatformCase` to cover more edge cases
* something else?

cc @Daniel-McCarthy @j-f1 @say25 @iAmWillShepherd

outofambit:
### My proposal:

> * keep toPlatformCase and spread in the codebase (where applicable)

but additionally:

* make the limitations abundantly clear through code documentation
* in components where we can't use `toPlatformCase` create a helper function in that component that returns the correct wording and casing depending on platform to avoid the inline platform check syntax outlined aboe

say25:
I am personally in favor of options 2 or 3. I think for first time contributors, it is easier to call the `toPlatformCase` function and get what one might expect.

say25:
@j-f1 that would basically be what toPlatformCase would conditionally call correct?

j-f1:
Yep @say25! We could even define `toPlatformCase` as:

```ts
export const toPlatformCase = __DARWIN__ ? title : (s: string) => s
```

j-f1:
For edge cases, we could define a second parameter that’s used as-is on macOS when it’s passed instead of the title-cased version.

[...]

PR #9728, opened 2020-05-07
Discard individual selection of lines from file

rafeca:
Closes https://github.com/desktop/desktop/issues/2402

## Description

This PR allows users to discard local changes to specific lines or hunks from a file via the context menu. To do so, a new `git` method called `discardChangesFromSelection()` has been created.

The implementation of the `discardChangesFromSelection()` method differs from the one initially done in https://github.com/desktop/desktop/pull/9399 : instead of reading the whole file into memory and manually copying/deleting the discarded lines, we rely on `git` and its patching functionality to discard the changes. This should make the discard process much more performant for large files.

In order to do so, a new method `formatPatchToDiscardChanges()` has been created. This method generates a string containing the patch that needs to be applied to discard the selected changes from the diff.

### Screenshots

demo

## Release notes

Notes: [Added] Discard individual lines from a changed file


rafeca:
I'm going to add a bunch of unit tests to the `discardChangesFromSelection()` method, since I had to deal with many edge cases while implementing this.

@tierninho would be good to do quite a bit of testing of discarding lines/hunks from files on different scenarios. Some stuff I tested so far:

* Files marked as moved.
* Files with some changes added to the index.
* Deleted/added files.

There is probably more tricky scenarios (or combinations of these) that would be worth testing.

rafeca:
I've added 4675a23 which does a few changes:

1. Changes the label names to remove the line number from there (sometimes displaying the line number gets confusing since the diff tool display two different sets of line numbers - one for the original file and another for the modified one).
2. Disables the "Discard this line" menu item on hunks that have both added and removed lines. This is done to avoid an unexpected behaviour when discarding changes from modified lines.

As an example of 2., if a user tries to discard the added line L275 from the following chunk:

image

Then the expected behaviour would be to remove the comment from that line. Instead, what `git` will do is removing that added line and end up with this line completely removed in the file:

image

rafeca:
Thanks for testing this @tierninho ! ✨ 

> Right clicking in the "discard all lines" area of a file that has only one individual line says: “Discard these [added/deleted] lines”. If the file has only one line, should we change the language to singular?

Yup, the language should be in singular in this case. I'm trying to reproduce with a new created file with a single line but I'm seeing a correct message:

image

When do you see the wrong message?

> If a file is modified and the user opts to delete all the new lines, the entire contents of the file are deleted. For example if I click "discard all lines" in this file initiated from either the red or green area, then all file contents are deleted, including the white area. I think this ok but just want to point it out.

We shouldn't delete all file contents when discarding changes, that is quite wrong 😟 Are you sure that all the file contents were deleted when you did so (tried reproducing but couldn't)? Maybe it's just that the file had no longer any local change and then Desktop didn't show any changes on the diff view.

> I discarded some lines, edited the file in Atom again, and then tried to discard some more lines but the single-line discard option was greyed out. Why?

We've decided to disable discarding single lines on hunks that contain both added and removed lines, since discarding lines in these scenarios can cause unexpected results (see comment in https://github.com/desktop/desktop/pull/9728#issuecomment-625460839). Was it confusing that you couldn't discard single lines there?

> Do we want to leave the user in a state with an empty file as the user can delete all added/deleted lines?

Good point! Yup I think we should delete the file in that case, gonna implement that.

> If a user removes modified lines and then file disappears, is this expected?

Do you mean disappearing from the GitHub Desktop UI? If that's the case then yes: if a user discards all changed lines of a file, it'll disappear from the "changed files" list since it won't have local changes anymore.

> The blue highlighting was somewhat confusing at first. A line does not have to be selected blue in order to right-click. I understand why, it was just odd looking like this:

I agree, the fact that line numbers can be used both to select changes and discard lines can be a bit confusing at first. 

> "Undo" button would be nice, but I foresee problems if 100 lines get reinserted. The 'Are you sure?' popup is a good compromise.

I think we'll have to end up implementing an "undo" button at some point, but we need to find a good place in the current UI to fit it.

> I managed to discard lines from a conflicted file. I believe we are supporting this.

ups, I didn't test this 😅 did you see any unexpected result when doing so? I'll do some tests and if things weird happen we can disable that.

tierninho:
@rafeca Thanks for the answers! 

> Yup, the language should be in singular in this case. I'm trying to reproduce with a new created file with a single line but I'm seeing a correct message. When do you see the wrong message?

I cannot reproduce it now so let's ignore this for now.

> We shouldn't delete all file contents when discarding changes, that is quite wrong 😟 Are you sure that all the file contents were deleted when you did so (tried reproducing but couldn't)? Maybe it's just that the file had no longer any local change and then Desktop didn't show any changes on the diff view.

Yes, sorry that is what I meant --only the local changes were discarding. The existing file remained as-is, so we are ok here.

> We've decided to disable discarding single lines on hunks that contain both added and removed lines, since discarding lines in these scenarios can cause unexpected results (see comment in #9728 (comment)). Was it confusing that you couldn't discard single lines there?

Only somewhat confusing at first but I realized that I had made multiple rounds of changes and that must of been the cause. 

> Do you mean disappearing from the GitHub Desktop UI? If that's the case then yes: if a user discards all changed lines of a file, it'll disappear from the "changed files" list since it won't have local changes anymore.

Yes, from the UI. Thanks for the confirmation.

> ups, I didn't test this 😅 did you see any unexpected result when doing so? I'll do some tests and if things weird happen we can disable that.

All seemed to work fine when I tested it 👍 I'll report back if I can break it.

tierninho:
I was just testing merge conflicts again and this conversation👇 made me wonder:

> Do we want to leave the user in a state with an empty file as the user can delete all added/deleted lines?

> Good point! Yup I think we should delete the file in that case, gonna implement that.

Based on that, if this gets implemented should be still be deleting the file even if it is a conflicted file? I think that might cause some problems as a user might remove all lines to resolve a merge conflict and still expect to see the file checked green, like so:

Screen Shot 2020-05-11 at 9 40 31 AM

rafeca:
> Based on that, if this gets implemented should be still be deleting the file even if it is a conflicted file?

Good point, in the case of a conflicted file, it should still appear on the changed files list, even if the user discard all its changes.

app/src/lib/feature-flag.ts
```diff
@@ -185,3 +185,10 @@ export function enableGitTagsDisplay(): boolean {
 export function enableGitTagsCreation(): boolean {
   return true
 }
+
+/**
+ * Should we show the discard lines/hunks contet menu item?
```
Rexogamer, line 190: tiny typo:
```suggestion
 * Should we show the discard lines/hunks context menu item?
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -480,17 +499,138 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
+
+    const selection = file.selection
+      .withSelectNone()
+      .withRangeSelection(startLine, endLine - startLine + 1, true)
+
+    this.props.onDiscardChanges(this.props.diff, selection)
+  }
+
+  private getDiscardLabel(hunkType: HunkType, numLines: number): string {
+    return toPlatformCase(
```
niik, line 627: I'd rather we not use `toPlatformCase`. It has a checkered past in Desktop and I believe the last time we discussed it we were all in favor of removing it so adding more usage to it would be counterproductive.

For context the rules for title casing are much too complex to succinctly express in a helper function like this which is why you see these ternary statements scattered about the code base. See https://github.com/desktop/desktop/pull/5841#pullrequestreview-162478757 and https://github.com/desktop/desktop/issues/5847 for some of the back story.

In this scenario it's possible that it's doing the right thing but while I'm far far far from an expert in English grammar I wonder if the word "these" can be considered an article in this sentence and therefore should be excluded from uppercasing based on https://developer.apple.com/library/archive/documentation/FinalCutProX/Conceptual/FxPlugHIG/TextStyleGuidelines/TextStyleGuidelines.html

> In title-style capitalization, capitalize every word except:
> - Articles (a, an, the)
>

I propose we make this easy on us by rewording and refactor here 👇 

```ts
private getDiscardLabel(hunkType: HunkType, numLines: number): string {
  const s = numLines > 1 ? 's' : ''
  return __DARWIN__ ? `Discard Line${s}` : `Discard line${s}`
}
```

or, slightly more verbose

```ts
private getDiscardLabel(hunkType: HunkType, numLines: number): string {
  if (numLines > 1) {
    return __DARWIN__ ? 'Discard Lines' : 'Discard lines'
  } else {
    return __DARWIN__ ? 'Discard Line' : 'Discard line'
  }
}
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -773,7 +913,12 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
       return
     }
 
-    const { file, hunks, readOnly } = this.props
+    // Discard right click events
```
niik, line 916: This comment read to me as  "ignore right click events". Might I suggest the following rewording?

```suggestion
    // We only care about the primary button here, secondary
    // button clicks are handled by `onContextMenu`
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -401,6 +415,11 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
   private onDocumentMouseUp = (ev: MouseEvent) => {
     ev.preventDefault()
 
+    // Discard right click events
```
niik, line 418: This comment read to me as  "ignore right click events". Might I suggest the following rewording?

```suggestion
    // We only care about the primary button here, secondary
    // button clicks are handled by `onContextMenu`
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -797,7 +942,12 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
       return
     }
 
-    const { file, hunks, readOnly } = this.props
+    // Discard right click events
```
niik, line 945: This comment read to me as  "ignore right click events". Might I suggest the following rewording?

```suggestion
    // We only care about the primary button here, secondary
    // button clicks are handled by `onContextMenu`
```

app/src/lib/patch-formatter.ts
```diff
@@ -225,3 +225,106 @@ export function formatPatch(
...
+      }
+
+      // Context lines can always be let through, they will
+      // never appear for new files.
+      if (line.type === DiffLineType.Context) {
+        hunkBuf += `${line.text}\n`
+        oldCount++
+        newCount++
+      } else if (selection.isSelected(absoluteIndex)) {
+        // Reverse the change (if it was an added line, treat is as removed and viceversa).
```
niik, line 276: total nitpick

```suggestion
        // Reverse the change (if it was an added line, treat it as removed and vice versa).
```

app/src/lib/stores/git-store.ts
```diff
@@ -1497,6 +1499,21 @@ export class GitStore extends BaseStore {
     })
   }
 
+  public async discardChangesFromSelection(
+    filePath: string,
+    diff: ITextDiff,
+    selection: DiffSelection
+  ) {
+    await this.performFailableOperation(async () => {
+      await discardChangesFromSelection(
```
niik, line 1508: I think we can skip the inner await here since `performFailableOperation` will await it so it only creates another Promise closure 

```suggestion
      discardChangesFromSelection(
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -480,17 +499,138 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
+          action: () => this.onDiscardChanges(file, lineNumber),
+          enabled: hunkType !== HunkType.Modified,
+        },
+      ]
+    }
+
+    return null
+  }
+
+  private getHunkType(range: { from: number; to: number }): HunkType | null {
```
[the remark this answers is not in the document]
rafeca: Makes sense! I think the naming that you proposed is much better than the current one as well 😅 

re. checking each line, I tried to avoid having to do so because new added files (or removed ones) there's only going to be a single hunk (or range composition) and we'll need to iterate through all file lines. Anyways since the check is super simple and it's only executed when right clicking I don't think it should have any noticeable impact on performance (I'm going to do a quick check with a file with a few thousand lines just to be sure).

app/src/ui/diff/text-diff.tsx
rafeca: Good point! I just realized that this method was not used anywhere else in the app 🤦‍♂️ (don't know how I even discovered it).

If you don't mind I'll send another PR to completely remove that method from the codebase.

app/src/lib/stores/git-store.ts
rafeca: I don't think we can just remove the `await` here: we'll need to also return the promise that gets returned by `discardChangesFromSelection()` to make `performFailableOperation()` await that promise.

I can do that by removing the braces from the function and get advantage of the implicit return.

app/src/lib/git/discard-changes.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param repository The repository in which to update the working directory
+ *                   with information from the index
+ *
+ * @param filePath   The relative path in the working directory of the file to use
+ *
+ * @param diff       The diff containing the file local changes
+ *
+ * @param selection  The selection of changes from the diff to discard
+ */
+export async function discardChangesFromSelection(
```
rafeca, line 22: @niik I'm going to move this method to the `apply.ts` file, since as we discussed we use the first git argument as the name of the file for git methods.

app/src/ui/diff/text-diff.tsx
niik: > If you don't mind I'll send another PR to completely remove that method from the codebase.

I don't mind, quite the opposite 😄

app/src/lib/git/discard-changes.ts
niik: Awesome!

app/src/ui/diff/text-diff.tsx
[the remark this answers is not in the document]
rafeca: I've done some tests with out current `yarn.lock` file which is 11k lines (I think this could be a relatively common scenario, where a user wants to discard a `yarn.lock` file), and while the results are worse than I expected they're still not terribly bad.

It took between 2ms in the best case (probably with all codecaches warned in v8) and 20ms in the worse case (these results are on an average Intel i5 Windows laptop).

Something a bit worrying is that we have to call `diffLineForIndex()` for each line in the range, and this method iterates through all hunks on the file, which can potentially lead to an `O(n * m)` operation (it's a quite rare scenario, since this would require selecting a hunk with a ton of lines inside a file with a ton of hunks, but still something to be aware).

Main problem here is `getHunkType()` accepts a range, which can potentially contain lines from multiple hunks, although we know for sure (if I'm not wrong) that in the discard scenario you can only discard a hunk at a time. I'm going to try to change slightly the API of this method to be more restrictive and avoid this `O(n * m)` operation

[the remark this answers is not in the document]
rafeca: I've ended up addressing this in 754f85b49: by moving this calculation to the `findInteractiveDiffRange()` method we make it costless (at least in big o terms).

This could add (an unapreciable tbh) overhead to other methods using `findInteractiveDiffRange()`, but since this is only used for calculating selections on partial commits it should be fine.

rafeca:
@niik I've addressed your comments so this PR is ready for another review 😃

app/src/ui/diff/diff-explorer.ts
```diff
@@ -1,4 +1,27 @@
-import { DiffLine, DiffHunk } from '../../models/diff'
+import { DiffLine, DiffHunk, DiffLineType } from '../../models/diff'
+
+/**
+ * Indicate the type of changes that are included in the current range.
+ */
+export enum DiffRangeType {
+  /** Only contains added lines. */
+  Additions,
+  /** Only contains added lines. */
```
niik, line 9:
```suggestion
  /** Only contains deleted lines. */
```

app/src/ui/diff/index.tsx
```diff
@@ -64,6 +64,11 @@ interface IDiffProps {
 
   /** Hiding whitespace in diff. */
   readonly hideWhitespaceInDiff: boolean
+
+  readonly onDiscardChanges?: (
```
niik, line 68: Can we copy and paste the nice documentation you added for `onDiscardChanges` in `ITextDiffProps` here?

app/src/ui/diff/diff-explorer.ts
```diff
@@ -85,10 +115,52 @@ export function findInteractiveDiffRange(
...
   }
 
   const to =
     contextLineAfterIndex !== null ? contextLineAfterIndex : hunk.unifiedDiffEnd
 
-  return { from, to }
+  return { from, to, type: rangeType }
+}
+
+function getNextRangeType(
```
niik, line 128: Big fan of this approach ✨

app/src/ui/diff/diff-explorer.ts
```diff
@@ -1,4 +1,27 @@
...
+  /** Contains both added and removed lines. */
+  Mixed,
+}
+
+/**
+ * Helper object that represents a range of lines in a diff.
+ * Its type represents the type of interactive (added or deleted)
+ * lines that it contains, being null when it has no interactive lines.
+ */
+type DiffRange = {
```
niik, line 20: Love that you took the time to break this out into its own type. I think conventionally we'd use an interface here instead of a type but I'm absolutely fine with this, it's such an improvement over the anonymous tuple type.

nk250319:
Of note, the verbiage SourceTree uses is plural for everything: "Discard Selected Lines" in addition to "Discard Hunk" (if the lines are all grouped together).

app/src/ui/diff/text-diff.tsx
```diff
@@ -480,17 +505,133 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
+    let type = ''
+
+    if (rangeType === DiffRangeType.Additions) {
+      type = __DARWIN__ ? 'Added' : 'added'
+    } else if (rangeType === DiffRangeType.Deletions) {
+      type = __DARWIN__ ? 'Removed' : 'removed'
+    } else if (rangeType === DiffRangeType.Mixed) {
+      type = __DARWIN__ ? 'Modified' : 'modiifed'
+    } else {
+      assertNever(rangeType, `Invlaid range type: ${rangeType}`)
```
niik, line 621:
```suggestion
      assertNever(rangeType, `Invalid range type: ${rangeType}`)
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -480,17 +505,133 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
+
+  private getDiscardLabel(rangeType: DiffRangeType, numLines: number): string {
+    let type = ''
+
+    if (rangeType === DiffRangeType.Additions) {
+      type = __DARWIN__ ? 'Added' : 'added'
+    } else if (rangeType === DiffRangeType.Deletions) {
+      type = __DARWIN__ ? 'Removed' : 'removed'
+    } else if (rangeType === DiffRangeType.Mixed) {
+      type = __DARWIN__ ? 'Modified' : 'modiifed'
```
niik, line 619:
```suggestion
      type = __DARWIN__ ? 'Modified' : 'modified'
```

rafeca:
Thanks for the feedback @nk250319 and @niik ! I agree that "this"/"these" don't add much information in the label (although I'm not a native speaker so take my opinions with a grain of salt).

Let me summarize a couple of constrains/edge cases that we faced when designing this label, so we can take them into account if we want to iterate again through the text of the labels:

* Since discarding changes does the opposite as what the diff is showing (it will delete green/added lines and it will re-add red/removed lines), this can be a bit confusing to a user. Imagine the following example, where discarding these lines will re-add them to the file (this may be obvious for people :
  image
  This made us add the explicit "Added"/"Deleted" adjective on the label ([link to internal slack conversation](https://github.slack.com/archives/C17LP0XU3/p1588795624418900)).

* Since we have two sets of line numbers on a diff (the ones from the previous state and the ones from the next state), in some situations displaying the line numbers (or even the number of lines to discard) can be quite confusing for users. For example in this scenario: 
  image
  This made us remove the line numbers from the labels ([link to slack conversation](https://github.slack.com/archives/C17LP0XU3/p1588870600433400)).

I think we never discussed whether to use "this"/"these", it just came as some glue between the verb "Discard" and "Added Lines"/"Removed Lines".

I think that we can probably just ditch the "this" and just display "Discard Added Line"/"Discard Added Lines" (I don't have a strong opinion about which one I prefer), what do you think @billygriffin / @ampinsk ?

ampinsk:
> I think that we can probably just ditch the "this" and just display "Discard Added Line"/"Discard Added Lines" (I don't have a strong opinion about which one I prefer), what do you think @billygriffin / @ampinsk ?

I don't mind ditching the "this!" 👍

tierninho:
- Ran through another rounds of testing given the recent changes, and things are functioning as expected.

- I wanted to point out the distinction between discarding one line and multiple lines hard to recognize, as we are literally talking about 1 or 2 pixels apart and one letter away from each other (see pics and notice distance of the context menu). I suggest making it clearer, especially due to the destructive nature of the feature. 

  Screen Shot 2020-06-02 at 4 03 12 PM
  Screen Shot 2020-06-02 at 4 02 27 PM

  When discarding or ignoring files, we include the word "all", perhaps `Discard All Added Lines` is more clear.

rafeca:
Thanks for the feedback @tierninho ! I agree the difference it's too subtle, but I think "Discard All Added Lines" can be misleading since it can mean that it will discard all added lines from that file.

What about "Discard Multiple Added Lines"? it feels super verbose so any other alternative will be appreciated!

billygriffin:
I think I'm fine with where we're at after the removal of "this." There don't seem to be any perfect solutions, but where we're at is pretty close. Unless @ampinsk feels strongly about changing the language from here, I think we can get feedback and update verbiage if there's an opportunity to make it more clear once people are using it.

jabamaus:
Hi,

Thanks for the feature - very useful. Not fully working for me though. 'Discard added line' works as expected but 'Discard modified line' is always greyed out in the context menu. What am I missing?

Cheers.

Rexogamer:
> Hi,
> 
> Thanks for the feature - very useful. Not fully working for me though. 'Discard added line' works as expected but 'Discard modified line' is always greyed out in the context menu. What am I missing?
> 
> Cheers.

Hey! I think you might want to [create an issue for that](https://github.com/desktop/desktop/issues/new/choose) with more info (eg your OS, Github Desktop version and more info about your files/diffs).

PR #9729, opened 2020-05-07
Add confirmation dialog when discarding lines from a file

rafeca:
Closes https://github.com/desktop/desktop/issues/2402

This PR is based on https://github.com/desktop/desktop/pull/9728

## Description

This PR adds a confirmation dialog when discarding a selection from a file. The discard dialog has been reused (and tweaked) for this use case.

### Screenshots

image

## Release notes

Notes: no-notes


app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -57,19 +62,37 @@ export class DiscardChanges extends React.Component<
...
     }
-    return __DARWIN__ ? 'Discard changes' : 'Discard changes'
   }
 
   private getDialogTitle() {
-    if (this.props.discardingAllChanges) {
-      return __DARWIN__
-        ? 'Confirm Discard All Changes'
-        : 'Confirm Discard all changes'
+    switch (this.props.discardType) {
```
niik: If we grab this out of  props we can get the `assertNever` to fit nicely onto one line

```suggestion
    const { discardType } = this.props

    switch (discardType) {
```

app/src/ui/discard-changes/discard-changes-dialog.tsx
```diff
@@ -57,19 +62,37 @@ export class DiscardChanges extends React.Component<
...
+      case DiscardType.SomeFiles:
+      case DiscardType.Selection:
+        return __DARWIN__
+          ? 'Confirm Discard changes'
+          : 'Confirm Discard changes'
+      default:
+        return assertNever(
+          this.props.discardType,
+          'Invalid discardType property'
+        )
```
niik: We typically include the value that we thought was guaranteed to be exhaustively checked in case it happens to fail at runtime. None ever has thus far but theoretically we could be asserting on a value that originates from the API and if that then fails we want to know what new value they've introduced.

```suggestion
        return assertNever(discardType, `Unknown DiscardType: ${discardType}`)
```

rafeca:
Ups I was in the middle of updating this PR when your PR came in 😅

I usually re-request a review when the PR is ready to get review again, does this work? (another option would be to put the PR in draft mode, but this unsubscribes people from the PR).

I've done quite a few changes on the PR, basically I've created a separate dialog component for discarding a selection and got rid of the `discardType` enum.

What triggered me to do that is that I opted to not pass an `onSubmit` callback to the dialog via the `Popup` object (since as we talked before this can end up executing the discard on a wrong file/repository if props change while the dialog is opened).

At the point where I had to pass the `diff` and `diffSelection` objects via the `Popup` object I realized that reusing the dialog component was not worth it anymore (too many needed checks compared to the amount of code reuse).

app/src/ui/discard-changes/discard-selection-dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public render() {
+    const isDiscardingChanges = this.state.isDiscardingSelection
+
+    return (
+      <Dialog
+        id="discard-changes"
+        title={
+          __DARWIN__ ? 'Confirm Discard changes' : 'Confirm Discard changes'
```
niik, line 73:
```suggestion
          __DARWIN__ ? 'Confirm Discard changes' : 'Confirm discard changes'
```

app/src/ui/discard-changes/discard-selection-dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+    super(props)
+
+    this.state = {
+      isDiscardingSelection: false,
+      confirmDiscardSelection: false,
+    }
+  }
+
+  private getOkButtonLabel() {
+    return __DARWIN__ ? 'Discard changes' : 'Discard changes'
```
niik, line 63:
```suggestion
    return __DARWIN__ ? 'Discard Changes' : 'Discard changes'
```

niik:
> I usually re-request a review when the PR is ready to get review again, does this work? (another option would be to put the PR in draft mode, but this unsubscribes people from the PR).

I think this is a great discussion to have between the three of us (cc @outofambit). If there's time left at the end of our fortnightly maybe we could chat about it. We've gone through a few iterations of how to request/re-request reviews and we haven't really decided on how to deal with the new fancy functionality we've got like draft PRs and review re-requests. I think in this particular case just a comment to say that you were going to spend some more time working on this would have been great. I was under the impression that the work was all buttoned up and just needed the conflict resolved.

> What triggered me to do that is that I opted to not pass an `onSubmit` callback to the dialog via the `Popup` object (since as we talked before this can end up executing the discard on a wrong file/repository if props change while the dialog is opened).

I'll take a look but that sounds like it could simplify matters!

app/src/ui/discard-changes/discard-selection-dialog.tsx
rafeca: Ups thanks for noticing this!

I copy-pasted both strings from the `DiscardChanges` dialog: https://github.com/desktop/desktop/blob/development/app/src/ui/discard-changes/discard-changes-dialog.tsx#L63

I'm going to fix these as well!

PR #9737, opened 2020-05-08
use only custom rules in tslint

outofambit:
At this point, all of our tslint rules are also covered by eslint except for the custom ones (see #6912). This turns off the duplicate tslint rules so developers aren't notified twice for the same error or warning.

@niik and I had discussed this a few months ago. finally making good on this. :)


tslint.json
```diff
@@ -1,42 +1,9 @@
 {
   "extends": ["tslint-config-prettier"],
-  "rulesDirectory": [
-    "node_modules/tslint-react/rules",
-    "node_modules/tslint-microsoft-contrib/",
```
niik, line 5: I'm an absolute novice when it comes to eslint but I can't figure out how the rules from tslint-microsoft-contrib (like react-this-binding-issue, promise-must-complete etc) get picked up in the eslint-realm. Should we keep the rules from microsoft-contrib in tslint.json in addition to our own custom rules perhaps?

niik: @outofambit I know you were OOO when I asked this so I'm bumping it in case you didn't see it but please don't feel stressed about it, it's not high priority

outofambit: @niik ah i think you're right in that they won't get picked up. i did some research and i think we can use https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin-tslint to move our custom tslint rules and the tslint-microsoft-contrib to be run by eslint. (they still get run by tslint at the end of the day, its just a wrapper.) this seems like a good step to keep our rules without having to run multiple linters.

niik:
I picked up this work that @outofambit started and went through each TSLint rule that we were previously using in `tslint.json` and matched them up with their corresponding ESLint rules using the [migration guide](https://github.com/typescript-eslint/typescript-eslint/blob/16667b1d5e24d0c3d4a30ca43f7e9e388f8d1ca6/packages/eslint-plugin/ROADMAP.md).

I've left the rules which had no substitute or where the substitute was lacking and as I attempted to replicate the spirit of our previous rules I also had to fix some errors introduced by the new rules.

I've enabled the default rule set for typescript-eslint and opted out of the rules that we don't want or that would be to cumbersome to turn on now. That way we should get new rules deemed worthy for general consumption applied for free as we upgrade the library.

app/src/ui/rebase/rebase-flow.tsx
```diff
@@ -73,14 +73,6 @@ interface IRebaseFlowProps {
 
 /** A component for initiating and performing a rebase of the current branch. */
 export class RebaseFlow extends React.Component<IRebaseFlowProps> {
-  public constructor(props: IRebaseFlowProps) {
-    super(props)
-
-    this.state = {
-      userHasResolvedConflicts: false,
-    }
```
rafeca, line 81: wooottt!

Was wondering why this didn't fail and just stumbled across this TypeScript weirdness around empty objects:

https://www.typescriptlang.org/play/#code/DYUwLgBAZg9jEF4IG8C+BYAUF28kEYAmLHAVwDsBjMASxnOgsoAoAjAQwCcAuFVASj4lMUJs1z8yVZkUkixAcgVyp1Og1FVCbLrwDyrAFYhqgtMM2VtE1dtm3mSlfKq16jKgGYdPCDCMmYGYY2C6U3jZh3vZRjspYQA

Do the new linting rules complain about types being empty objects?

app/src/ui/ui-view.tsx
```diff
@@ -16,7 +14,7 @@ interface IUiViewProps extends React.HTMLProps<HTMLDivElement> {}
  * Examples of what's not a View include the Changes and History tabs
  * as these are contained within the <Repository /> view
  */
-export class UiView extends React.Component<IUiViewProps, {}> {
+export class UiView extends React.Component<React.HTMLProps<HTMLDivElement>> {
```
rafeca, line 17: Nice!

app/src/ui/rebase/rebase-flow.tsx
niik: I believe I caught this when I turned on `@typescript-eslint/ban-types` but I had to turn it off because it was begin triggered a lot. It's probably the rule I'd most want to turn on in the future.

app/src/ui/branches/branches-container.tsx
```diff
@@ -65,12 +71,27 @@ export class BranchesContainer extends React.Component<
...
+    if (state.currentPullRequest !== props.currentPullRequest) {
+      return {
+        currentPullRequest: props.currentPullRequest,
+        selectedPullRequest: props.currentPullRequest,
+      }
+    }
+
+    return null
+  }
+
```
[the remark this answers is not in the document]
niik: I can rip it out but this was done in response to the new lint rules actually catching this error (not using the state property, only writing to it). I'll rip it out but it'll look weird for a bit then.

Issue #1597, opened 2017-05-18
Spell checking in commit summary & description


xt0rted:
I'm a horrible speller so it'd be nice to see some spell checking in these boxes similar to the message box in Slack's desktop app.

![image](https://cloud.githubusercontent.com/assets/831974/26184136/04b163a8-3b51-11e7-8e93-583d9b0da677.png)

shiftkey:
Prior art (I'm pretty sure this is used in the Slack app): https://www.npmjs.com/package/electron-spellchecker

xt0rted:
That's the package listed in their acknowledgements list.

roryokane:
If you add spell-checking, make sure it can easily be disabled as a setting. I find that for writing about code, it often annoys me more than it helps.

For example, if I type “Refactor FooGlobber”, I don’t want “FooGlobber” to be marked as misspelled. Nor “refactor”, which is not in all dictionaries. The same with file paths – if I write “changed src/core/index.js”, I don’t want “src” to be marked as misspelled.

ahmed-taj:
I would like to give it a try

@roryokane when you said

> make sure it can easily be disabled as a setting

Can you clarify the above statement more!

@shiftkey anything I need to consider ?

roryokane:
@ahmed-taj I mean that somewhere, there should be an interface for disabling (and re-enabling) spell-checking. And when the user triggers that action to disable spell-checking, the program should remember it and not enable spell-checking if you relaunch the app and then type in another commit message again.

As for the interface for disabling spell-checking, given that spelling suggestions would be given as context menu options, then it would make sense to have the disable method be another menu item such as "✓ Check Spelling" that will uncheck itself when selected. It would be divided by a menu separator from the spelling suggestions.

ahmed-taj:
@roryokane great, thanks

xt0rted:
An `Add to Dictionary` option would be nice for the reasons mentioned above, but probably out of scope for the initial implementation. It also looks like there's some issues with making this work. https://github.com/electron-userland/electron-spellchecker/issues/36

shiftkey:
@ahmed-taj 

>  anything I need to consider ?

Not really. I'm not familiar with the `electron-spellchecker` project so I'm not sure what it supports in terms of things like "Add to Dictionary". Have a go at getting it wired up and working, and we can figure out the rest of the UX stuff either here or in a PR.

elan:
[...]

Issue #9710, opened 2020-05-05
Crash reports from Sentry - read ENOTCONN


billygriffin:
### Describe the bug

We're seeing a crash report with the error `read ENOTCONN`

### Version & OS

Seems to be specific to macOS and 2.4+

outofambit:
These are likely caused by https://github.com/nodejs/node/issues/26315, which we likely inherited when we moved to electron `v7.x`. This wasn’t fixed in nodejs until in the `v12.13.1` release, which isn’t included in electron v7.x. (They’ve only upgraded it to nodejs `v12.8.1`.) The next release of Electron (`v8.2.6`) will likely include a version of nodejs with the fix for this.

plivesey:
The app crashes consistently on launch for me. Is there a workaround to fix this locally? It suddenly started happening and I'm not sure what changed.

```
Error: read ENOTCONN
    at Socket._read (net.js:530:20)
    at Socket.Readable.read (_stream_readable.js:481:10)
    at new Socket (net.js:328:12)
    at Object.Socket (net.js:237:41)
    at createSocket (internal/child_process.js:311:14)
    at ChildProcess.spawn (internal/child_process.js:424:23)
    at spawn (child_process.js:553:9)
    at Object.execFile (child_process.js:237:17)
    at Object.module.<computed> [as execFile] (electron/js2c/asar.js:163:31)
    at file:///app/renderer.js:16:254915

Version: 2.4.3
OS: Mac OS 10.15.4```

plivesey:
Looks like for me it was because I had too many processes running. My whole computer was bricked: `fork: Resource temporarily unavailable`. I needed to go to Activity Monitor and kill a process which had 1000s of copies of itself.

The service is DTServiceHub which seems to be an Xcode related service. So I think my original cause was a bug in Xcode which bricks your computer.

outofambit:
> The next release of Electron (`v8.2.6`) will likely include a version of nodejs with the fix for this.

It did not, it turns out. but electron `v9.0.0` did. (it has nodejs `v12.14.1`)

billygriffin:
Since we're now on Electron 9, I think this should be fixed. Please let us know if you run into this issue again!

PR #9766, opened 2020-05-12
Chore/electron v8.4.0

say25:
⚠️ Builds off #9674 ⚠️ 

Fixes https://github.com/desktop/desktop/issues/10085
Fixes https://github.com/desktop/desktop/issues/10303

There is no larger issue for this. I can open one but this primarily "unlocks" the spell checking API exposed in Electron 8.

~~It may also fix a `read: ENOTCONN` error we're seeing more frequently on Electron 7 due to the node version.~~ *Edit (rafeca): The `ENOTCONN` error gets fixed by Electron v9 as discovered [by @outofambit](https://github.com/desktop/desktop/pull/9766#issuecomment-639114245))*

## Description

Updates the Electron from v7.1.8 to v8.4.0

## Breaking changes

**IPC between main and renderer processes now uses the Structured Clone Algorithm.**

We prepared our codebase for this in https://github.com/desktop/desktop/pull/10066

## Major changes

**Chromium got upgraded from v78 to v80.**

* [Release notes for v79](https://developers.google.com/web/updates/2019/12/nic79)
* [Release notes for v80](https://developers.google.com/web/updates/2020/02/nic80)

**Node.js got upgraded from v12.8.1 to v12.13.0.**

* [Release notes for v12.9.0](https://nodejs.org/en/blog/release/v12.9.0/)
* [Release notes for v12.10.0](https://nodejs.org/en/blog/release/v12.10.0/)
* [Release notes for v12.11.0](https://nodejs.org/en/blog/release/v12.11.0/)
* [Release notes for v12.12.0](https://nodejs.org/en/blog/release/v12.12.0/)
* [Release notes for v12.13.0](https://nodejs.org/en/blog/release/v12.13.0/)

**V8 got upgraded from v7.8 to v8.0**.

* [Release notes for v7.9](https://v8.dev/blog/v8-release-79)
* [Release notes for v8.0](https://v8.dev/blog/v8-release-80)

**Native support for optional chaining and nullish coallescendant syntax in v8 ([more info](https://v8.dev/blog/v8-release-80#javascript)).**

We have just started using them in our codebase, and we're now transpiling them. With this, we can remove the transpilation and save some bytes.

**Added spellchecker support ([more info](https://github.com/electron/electron/pull/20692)).**

This has been requested for Desktop in https://github.com/desktop/desktop/issues/1597. There's a WIP PR based on this one that enables this: https://github.com/desktop/desktop/pull/9767

**Programmatically modify traffic light positioning on macOS ([more info](https://github.com/electron/electron/pull/22533)).**

This will allow us to get rid of the title bar in macOS without making the traffic light buttons look uncentered (another thing is whether we want to do that 😅 ):

Screen Shot 2020-07-16 at 12 02 34

### Full changelog

Changelog of all major/minor Electron releases (these links may break when new Electron versions get published since the pagination in their changelog is not stable):

* [v7.2.0](https://www.electronjs.org/releases/stable?version=7&page=2#release-notes-for-v720)
* [v7.3.0](https://www.electronjs.org/releases/stable?version=7#release-notes-for-v730)
* [v8.0.0](https://www.electronjs.org/releases/stable?version=8&page=4#release-notes-for-v800)
* [v8.1.0](https://www.electronjs.org/releases/stable?version=8&page=3#release-notes-for-v810)
* [v8.2.0](https://www.electronjs.org/releases/stable?version=8&page=3#release-notes-for-v820)
* [v8.3.0](https://www.electronjs.org/releases/stable?version=8&page=2#release-notes-for-v830)
* [v8.4.0](https://www.electronjs.org/releases/stable?version=8&page=1#release-notes-for-v840)

## Release notes

Notes: No notes


billygriffin:
@outofambit I think you looked into #9710 further, and maybe this doesn't actually fix it?

outofambit:
@billygriffin correct i did look into this and later confirmed (https://github.com/desktop/desktop/issues/9710#issuecomment-634290004) that `v8.2.x` will not fix it. Neither will `v8.3.0`, which i just checked. unfortunately, the fix is only on the `v9.x` releases of electron.

say25:
If I might suggest, I think it would be easier to:
1. (Optional) Get to 7.X latest
1. Get to 8.X
1. Get to 9.X

kadirselcuk:
Merhaba buradaki çekme istediğim nasıl karşılanır?

rafeca:
@say25 thanks for working on this!

I'm going to fix the current conflicts and update the electron version [to v8.4.0](https://www.electronjs.org/releases/stable?version=8#release-notes-for-v840) so we can do some testing.

> If I might suggest, I think it would be easier to:
> 1. (Optional) Get to 7.X latest
> 2. Get to 8.X
> 3. Get to 9.X

I agree, and I think we can skip updating to 7.X latest, so the current plan is the following:

1. Get to 8.4.0 (this PR)
2. Get to 9.X (gonna send a different PR for that based on this one).

rafeca:
> I'm going to fix the current conflicts and update the electron version to v8.4.0 so we can do some testing.

Done! I've also updated the PR description with a list of potential interesting changes

rafeca:
I've added https://github.com/desktop/desktop/pull/9766/commits/7e6b6a6f075da9b369bb01833ea7e5d4d44d3686 to fix a regression caused by the electron upgrade when resizing the Desktop window:

demo

It looks like the Chromium flexbox implementation got even more stricter than in the previous version and we need to add more `min-height: 0` CSS attributes on some of our flexbox containers (similar to what we did when upgraded to Electron v7: https://github.com/desktop/desktop/pull/9344).

I haven't been able to identify the actual bugfix that caused this issue, but it has to be somehow related to the one that made our layouts break in Electron v7 (which was [this one](https://bugs.chromium.org/p/chromium/issues/detail?id=596743)).

Under other circumstances I would add this commit to a separate PR so it could get reviewed and merged in a quicker way, but since we're currently limited in terms of code reviews putting this commit in this PR will make things easier.

app/.npmrc
```diff
@@ -1,4 +1,4 @@
 runtime = electron
 disturl = https://atom.io/download/electron
-target = 7.1.8
+target = 8.2.5
```
say25, line 3: This needs to be 8.4.0

say25:
```suggestion
target = 8.4.0
```

app/src/main-process/main.ts
```diff
@@ -450,9 +450,13 @@ app.on('ready', () => {
...
         const menu = buildContextMenu(items, indices => resolve(indices))
-
         const window = BrowserWindow.fromWebContents(event.sender)
-        menu.popup({ window, callback: () => resolve(null) })
+
+        if (window != null) {
+          menu.popup({ window, callback: () => resolve(null) })
+        } else {
+          menu.popup()
+        }
```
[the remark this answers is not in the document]
rafeca: Good catch! I think that I caused this issue by mistake when dealing with some merge conflicts... 

> Could we simplify this to just coercing fromWebContents null return values to undefined? I.e:

👍

> I think we ought to do the same in execute-menu-item as well:

👍

app/src/main-process/main.ts
```diff
@@ -450,9 +450,13 @@ app.on('ready', () => {
     ): Promise<ReadonlyArray<number> | null> => {
       return new Promise(resolve => {
         const menu = buildContextMenu(items, indices => resolve(indices))
+        const window = BrowserWindow.fromWebContents(event.sender) || undefined
 
-        const window = BrowserWindow.fromWebContents(event.sender)
-        menu.popup({ window, callback: () => resolve(null) })
+        if (window != null) {
```
say25, line 455: I think we should remove the if block here as both paths are identical

niik:
Thanks for getting this started @say25! Also, spectacular PR description work here @rafeca!

PR #9772, opened 2020-05-12
move moment plugin setup

outofambit:
This PR moves the setup for `moment-duration-format` to the primary ui renderer file.


outofambit:
@rafeca dangit, i didn't look into that part. thank you! does this cleanup still seem worth merging? i've edited the PR body to...lower expectations for the effect this will have.

PR #9775, opened 2020-05-12
Stop using whitelist and blacklist in our code base

billygriffin:
This is a simple change, and I'd prefer us not to be a part of the problem with language perpetuating stereotypes of associating "white" with things that are safe, and "black" with things that are suspect or shouldn't be permitted.

This PR should not affect anything in terms of app behavior. If I missed other instances, please call those out as well.

Notes: No notes

### DIFF
--- a/app/src/lib/shell.ts
+++ b/app/src/lib/shell.ts
@@ -152,7 +152,7 @@ async function getEnvironmentFromShell(
  */
 function mergeEnvironmentVariables(env: IndexLookup) {
   for (const key in env) {
-    if (DenylistedNames.has(key)) {
+    if (ExcludedEnvironmentVars.has(key)) {
       continue
     }
 
--- a/app/src/lib/shell.ts
+++ b/app/src/lib/shell.ts
@@ -10,7 +10,7 @@ type IndexLookup = {
 /**
  * The names of any env vars that we shouldn't copy from the shell environment.
  */
-const BlacklistedNames = new Set(['LOCAL_GIT_DIRECTORY'])
+const DenylistedNames = new Set(['LOCAL_GIT_DIRECTORY'])
 
 /**
  * Inspect whether the current process needs to be patched to get important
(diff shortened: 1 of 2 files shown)


docs/known-issues.md
```diff
@@ -114,7 +114,7 @@ are unable to find another cygwin DLL.
 
 Enabling Mandatory ASLR affects the MSYS2 core library, which is relied upon by Git for Windows to emulate process forking.
 
-**Not supported:** this is an upstream limitation of MSYS2, and it is recommend that you either disable Mandatory ASLR or whitelist all executables under `<Git>\usr\bin` which depend on MSYS2.
+**Not supported:** this is an upstream limitation of MSYS2, and it is recommend that you either disable Mandatory ASLR or allowlist all executables under `<Git>\usr\bin` which depend on MSYS2.
```
niik, line 117:
```suggestion
**Not supported:** this is an upstream limitation of MSYS2, and it is recommend that you either disable Mandatory ASLR or explicitly allow all executables under `<Git>\usr\bin` which depend on MSYS2.
```

app/src/lib/shell.ts
```diff
@@ -10,7 +10,7 @@ type IndexLookup = {
 /**
  * The names of any env vars that we shouldn't copy from the shell environment.
  */
-const BlacklistedNames = new Set(['LOCAL_GIT_DIRECTORY'])
+const DenylistedNames = new Set(['LOCAL_GIT_DIRECTORY'])
```
niik, line 13:
```suggestion
const ExcludedEnvironmentVars = new Set(['LOCAL_GIT_DIRECTORY'])
```

app/src/lib/shell.ts
```diff
@@ -152,7 +152,7 @@ async function getEnvironmentFromShell(
  */
 function mergeEnvironmentVariables(env: IndexLookup) {
   for (const key in env) {
-    if (BlacklistedNames.has(key)) {
+    if (DenylistedNames.has(key)) {
```
niik, line 155:
```suggestion
    if (ExcludedEnvironmentVars.has(key)) {
```

app/src/lib/shell.ts
billygriffin: Totally cool with this.

outofambit:
i'll let @niik review this again before merging!

Issue #9776, opened 2020-05-12
Authentification error, infinite retry loop. 


manugd14:
Previously, when cloning a GitLab (or other external host) repo on GitHub desktop, aat some point the app prompted you for your gitab credentials, as it was not able to perform the task with the stored (github) account details.

After today's changes, now it tells you that there is an authentification error, tells you a few possible reasons and let you either retry or cancel. No more options to enter external credentials, so virtually impossible now to clone  private GitLab repo on github desktop.

EGGamer:
Yes, this is happening to me too. Private GitLab repositories just broke. ![Image](https://i.imgur.com/VTO7lDy.png)

billygriffin:
Thanks for letting us know @manugd14 and @EGGamer, and sorry this is causing problems for you. It looks like it may have been a result of https://github.com/desktop/desktop/pull/7080, and we're investigating the cause and potential fix. Apologies that things aren't working for you in the meantime.

manugd14:
Thanks so much for your response. Please let us know how to proceed once there is a solution in place. 
Best wishes

Manuel.

manugd14:
Excuse me, I see you closed the issue. But the issue is still persisting for us. Could you let us know if we are missing any steps to update our client? Thanks.

billygriffin:
@manugd14 Sorry for the confusion. We merged a PR that fixes the problem, but isn't yet released to production. We're going to ship this to beta in the next few days, and at that point you'll hopefully be able to use the beta version, and then in two weeks we're planning to release to production to fix this and a few other things.

manugd14:
@billygriffin thanks for the response. I have tried to install on my computer the beta channel but the installer (linked on the home page of the git repository) seems to install the release version I already have. How should I proceed? Thanks.

billygriffin:
We released this to beta today so downloading the beta channel version in the README should help with this until the next version is released to production. Apologies again!

PR #9777, opened 2020-05-12
exempts known git auth errors from clone retry error handler

outofambit:
Closes https://github.com/desktop/desktop/issues/9776

## Description

- exempts known git auth errors from getting handled by the retry clone handler
- reorders error handlers so that auth error handler is always checked first.

## Release notes

Notes: [Fixed] Restore option to enter credentials for non-GitHub clone operations


manugd14:
How can I update my client with this changes?. Thanks.

PR #9783, opened 2020-05-13
Fix for Enter key in Clone Repository Dialog

### LINKED ISSUE #5234: Unexpected behaviour pressing enter/return on Clone a Repository options dialog
## Description

When pressing enter/return on Clone a Repository options dialog the behaviour is unexpected and confusing 

## Version

* GitHub Desktop: 1.2.6

* Operating system: macOS 10.13.6 (17G65)

## Steps to Reproduce

1. _Clone_ a repo from GitHub website
2. Look at the _Clone a Repository_ dialog
3. Press _Enter_

### Expected Behavior

Enter should choose the default blue Clone button

### Actual Behavior

Enter activates the GitHub.com tab

## Additional Information

screen shot 2018-07-24 at 14 25 49

### LINKED ISSUE #8804: autoFocus not working for clone URL textbox when not signed into GitHub.com
### Describe the bug

See title.

### Version & OS

GHD 2.2.4 on Windows 10 Enterprise (build 17763)

1. On a new system, install GHD, but don't sign it into a GitHub.com (or Enterprise Server) account
1. Press `Ctrl`+`Shift`+`O` => Focus is on `GitHub.com` tab
1. Switch to `URL` tab & clone something
1. Press `Ctrl`+`Shift`+`O` again to clone something else
1. Check focus (blue rim)

### Expected behavior

Focus should be on the URL textbox.

### Actual behavior

Focus is on `URL` tab. The URL textbox is only focused, if the dialog opens in the `GitHub.com` tab and one manually switches to the `URL` tab.

### Screenshots

grafik

### Logs

Nothing about "focus" in there.

### Additional context

https://github.com/desktop/desktop/blob/bded16772fd3405aee4875930ed6a4432ed3a076/app/src/ui/clone-repository/clone-generic-repository.tsx#L37-L40

does not seem to be triggered when the app isn't signed into any account.

ruggi99:
Closes #5234 
Closes #8804

## Description

- This PR fixes unexpected behavior when trying to clone a repository from Github site and interact with Enter/Return key without using pointing devices.
When dialog is shown, focus is firstly put on "URL or username/repository" input because that input has autoFocus property set. But then, `focusFirstSuitableChild()` function (called on dialog component mounted) puts focus to tab bar button, because of it's higher precedence. Here Enter/Return key doesn't submit the form because they are simple buttons and there are no listeners on that button.
So I've added tabIndex={1} to the "URL or username/repository" input, which ha higher precedence of tab bar button. I think this should be the default action, also because that input has autoFocus property set.
- Only applies to Clone repository dialog so it's not a breaking change.
- Others dialog with the same wrong behaviour maybe fixed in other PR.

### Screenshots

None

## Release notes

Notes: PR ready for review

### DIFF
--- a/app/src/ui/tab-bar.tsx
+++ b/app/src/ui/tab-bar.tsx
@@ -160,7 +160,7 @@ class TabBarItem extends React.Component<ITabBarItemProps, {}> {
         onClick={this.onClick}
         role="tab"
         aria-selected={selected}
-        tabIndex={selected ? 0 : -1}
+        tabIndex={selected ? undefined : -1}
         onKeyDown={this.onKeyDown}
         type="button"
       >


ruggi99:
After opening this PR, I've found that putting tabIndex={1} breaks the correct flow of focusing elements with Tab key. I don't know how I didn't think about it.
So it's not the best practice.
I've tried to not call `focusFirstSuitableChild()` function but for some reason I don't know, autoFocus on the input doesn't work.
Works in Publish Repository too.
Other dialogs should not be affected because they have not got Tab bars.
So my propose is to give higher priority to simple inputs (like text) than to elements with tabIndex={0}.
What you think about?

ruggi99:
Hy @niik , thank you for your feedback.

I've tried your idea before opening this PR. It works as expected but the reason why I dumped it is that I've thought `tabIndex={selected ? 0 : -1}` maybe addresses another bug.
If not, the question is.. Why `tabIndex={selected ? undefined : -1}` hasn't been thought before?

If we are sure that this doesn't alter other things (such as other algorithms or focus flow or other things..) I think is the best option.

Otherwise I think making a small change to the algorithm for `focusFirstSuitableChild()` function isn't so bad. 

> By doing it this way we also preserve the ability to boost a particular component's priority by 
setting an explicit tab index.

I didn't understand this so much.
With the actual algorithm you have to pay attention at some troubles. If you want to boost a particular component's priority, you can set tabIndex=0 but you have to be sure your component is before any other component with tabIndex=0. Otherwise you can set a tabIndex>0 but this leads to a probably wrong focus flow. So you can't boost a particular component without worrying about it's place in the DOM.

niik:
> If not, the question is.. Why `tabIndex={selected ? undefined : -1}` hasn't been thought before?

I'm relatively certain that using `1 : -1` as opposed to `undefined : -1` is just an oversight on my part. The initial implementation of the tab bar used divs and not buttons and it's possible I still had that in mind when I worked on the keyboard logic.

It's equally possible that I just didn't think hard enough about it 😛 Regardless I think it's the right thing to do. The intent behind the logic is to disable keyboard tab stop for all but the selected tab button, not the other way around.

> If we are sure that this doesn't alter other things (such as other algorithms or focus flow or other things..) I think is the best option.

It should be tested obviously before we merge this PR but `focusFirstSuitableChild` only applies to dialogs and the number of dialogs that have tab bars is limited (Clone Repository, Preferences, Publish Repository, and Repository Settings). That feels manageable to exhaustively test.

The approach proposed here of changing the algorithm for `focusFirstSuitableChild` might very well be something we want to do as well but I propose we start by this smaller, more manageable change and see if that solves the two issues reported and then revisit the discussion if/when we find other issues with the dialog focus logic.

ruggi99:
> I'm relatively certain that using `1 : -1` as opposed to `undefined : -1` is just an oversight on my part. The initial implementation of the tab bar used divs and not buttons and it's possible I still had that in mind when I worked on the keyboard logic.
> 
> It's equally possible that I just didn't think hard enough about it 😛 Regardless I think it's the right thing to do. The intent behind the logic is to disable keyboard tab stop for all but the selected tab button, not the other way around.

Now it makes sense to me and now I also think it's the best option. I've reverted changes to `focusFirstSuitableChild()` function and modified tabIndex in Tab Bar.

Tested with Clone Repository and works both from File menu and from Add menu in Repository dropdown. Since Clone Repository dialog opens with the previously tab bar selected, I also tested it with the various scenario and it works. Same with the other dialogs with Tab Bars.
Not tested with GitHub Enterprise Server, as I've not got one.

> Think it should close #8804 issue

Should this PR close this issue?

gingerbeardman:
Thanks both!

katrinleinweber:
Thank you for this fix :-)

PR #9784, opened 2020-05-13
Remove old feature flags

rafeca:
## Description

This PR removes the checks for old feature flags that have been set to true for more than 6 months. The list of removed flags is the following:

* enableBranchPruning	
* enableNoChangesCreatePRBlankslateAction	
* enableGroupRepositoriesByOwner	
* enableRebaseDialog	
* enableStashing	
* enableBranchProtectionChecks	
* enableBranchProtectionWarningFlow	
* enableHideWhitespaceInDiffOption	
* enableTutorial	

In terms of functionality nothing should change (the logic should be the same). In order to reduce the chances of manual errors while deleting the checks I've done the process in two steps: first replaced every function call by a `true` and then perform the dead code elimination manually.

Since this PR changes a bunch of files I suggest waiting a few days to merge it (to minimize the risk of merging this before we need to release a potential emergency v2.5.1 hotfix).

### DIFF
--- a/app/src/ui/repositories-list/repositories-list.tsx
+++ b/app/src/ui/repositories-list/repositories-list.tsx
@@ -162,10 +162,7 @@ export class RepositoriesList extends React.Component<
     const label = this.getGroupLabel(identifier)
 
     return (
-      <div
-        key={identifier}
-        className="filter-list-group-header group-repositories-by-owner"
-      >
+      <div key={identifier} className="filter-list-group-header">
         {label}
       </div>
     )
--- a/app/src/ui/repositories-list/repositories-list.tsx
+++ b/app/src/ui/repositories-list/repositories-list.tsx
@@ -20,7 +20,6 @@ import { IMenuItem } from '../../lib/menu-item'
 import { PopupType } from '../../models/popup'
 import { encodePathAsUrl } from '../../lib/path'
 import memoizeOne from 'memoize-one'
-import { enableGroupRepositoriesByOwner } from '../../lib/feature-flag'
 
 const BlankSlateImage = encodePathAsUrl(__dirname, 'static/empty-no-repo.svg')
 
@@ -161,9 +160,8 @@ export class RepositoriesList extends React.Component<
   private renderGroupHeader = (id: string) => {
     const identifier = id as RepositoryGroupIdentifier
     const label = this.getGroupLabel(identifier)
-    const className = enableGroupRepositoriesByOwner()
-      ? 'filter-list-group-header group-repositories-by-owner'
-      : 'filter-list-group-header'
+    const className = 'filter-list-group-header group-repositories-by-owner'
+
     return (
       <div key={identifier} className={className}>
         {label}
@@ -193,7 +191,6 @@ export class RepositoriesList extends React.Component<
     )
 
     const groups =
-      enableGroupRepositoriesByOwner() &&
       this.props.repositories.length > recentRepositoriesThreshold
         ? [
             makeRecentRepositoriesGroup(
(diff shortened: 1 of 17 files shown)


app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -71,33 +70,14 @@ export class RepositoryListItem extends React.Component<
       prefix = `${gitHubRepo.owner.login}/`
     }
 
-    const className = enableGroupRepositoriesByOwner()
-      ? 'repository-list-item group-repositories-by-owner'
-      : 'repository-list-item'
+    const className = 'repository-list-item group-repositories-by-owner'
```
niik, line 73: I think we should take this opportunity to delete the `:not(.group-repositories-by-owner)` meta selector from `_repository_list.scss` since that's never going to be applied now. If you want to tackle that in a separate PR that's fine but I think that a part of removing the feature flags should be removing these feature flag specific css classes.

app/src/ui/repositories-list/repositories-list.tsx
```diff
@@ -161,9 +160,8 @@ export class RepositoriesList extends React.Component<
   private renderGroupHeader = (id: string) => {
     const identifier = id as RepositoryGroupIdentifier
     const label = this.getGroupLabel(identifier)
-    const className = enableGroupRepositoriesByOwner()
-      ? 'filter-list-group-header group-repositories-by-owner'
-      : 'filter-list-group-header'
+    const className = 'filter-list-group-header group-repositories-by-owner'
```
niik, line 163: I think we should take this opportunity to remove the `group-repositories-by-owner` meta class and roll up those rules into the `filter-list-group-group-header` class since that's always going to be applied now. If you want to tackle that in a separate PR that's fine but I think that a part of removing the feature flags should be removing these feature flag specific css classes.

app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -220,9 +200,8 @@ const renderAheadBehindIndicator = (aheadBehind: IAheadBehind) => {
 }
 
 const renderChangesIndicator = () => {
-  const classNames = enableGroupRepositoriesByOwner()
-    ? 'change-indicator-wrapper group-repositories-by-owner'
-    : 'change-indicator-wrapper'
+  const classNames = 'change-indicator-wrapper group-repositories-by-owner'
```
niik, line 203: I think we should take this opportunity to delete the `.change-indicator-wrapper:not(.group-repositories-by-owner)` meta selector from `_repository_list.scss` since that's never going to be applied now. If you want to tackle that in a separate PR that's fine but I think that a part of removing the feature flags should be removing these feature flag specific css classes.

app/src/ui/repositories-list/repository-list-item.tsx
rafeca: I agree and thanks for the pointer! since these changes didn't remove any CSS class I thought there would not be any styles that we could remove (didn't think about `:not()` lol).

app/src/ui/repositories-list/repositories-list.tsx
rafeca: I think I prefer to tackle this specific cleanup in a separate PR, since I'm not sure this cleanup can be as trivial as just removing a few CSS style declarations (maybe I'm wrong)

rafeca: Ok the cleanup turned up much easier than I anticipated 😜 so I've ended up including it in this PR as 844c10bab1802f862ffc9bc6d2667fcc334d876a

rafeca:
@niik I've ended up removing all instances of `group-repositories-by-owner` in the CSS or TS in this PR, thanks for the tip!

niik:
> @niik I've ended up removing all instances of `group-repositories-by-owner` in the CSS or TS in this PR, thanks for the tip!

Awesome, thanks for taking the time to really clean it up!

PR #9788, opened 2020-05-13
Add "Copy GitHub URL" to history commit right-click menu

tigpt:
Feature Request
## Description

This Commit add a "Copy GitHub URL" to the right-click on commits in history tab.
When working with GitHub Desktop client, after a commit I need to reference that commit in a issue tracker, it can be on the same GitHub repo, but it can be other GitHub repo or even other issue tracker like JIRA.

### Currently to get a Full URL to a commit I need to:
1. Right-Click a Commit
2. Click on "View on GitHub"
3. Wait for browser to open a new tab and load the commit URL
4. Double Click the browser URL
5. Type Copy Shortcut [Cmd+C] / [Ctrl+C]
6. Close the current tab [Cmd+W] / [Ctrl+W]

### With the new button we can get the full URL to a commit in the clipboard with:
1. Right-Click a Commit
2. Click on "Copy GitHub URL"

The URL is on the clipboard ready to paste on your issue tracker.

### Screenshots

Screenshot 2020-05-13 at 18 42 35
## Release notes
Notes: Add a new button to Copy GitHub URL to a commit.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -116,6 +116,7 @@ import { findUpstreamRemoteBranch } from '../lib/branch'
 import { GitHubRepository } from '../models/github-repository'
 import { CreateTag } from './create-tag'
 import { RetryCloneDialog } from './clone-repository/retry-clone-dialog'
+import { clipboard } from 'electron'
 
 const MinuteInMilliseconds = 1000 * 60
 const HourInMilliseconds = MinuteInMilliseconds * 60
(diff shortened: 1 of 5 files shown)


billygriffin:
Hi @tigpt, thanks for your desire to contribute! We only accept PRs for things that have already been discussed as an issue and the team indicates interest in a community PR to solve. For this one generally, we try to be really mindful of the principle that anything added (for example, to a context menu) detracts from the prominence of everything else. So generally that leads us to intentionally doing less. In this case, I'm not sure we've seen a lot of evidence that the current way of doing it is particularly painful, though I remember it coming up once before. With this context, if you think it's worthwhile you're welcome to still create a new issue and we can continue the discussion from there. Thanks again!

PR #9796, opened 2020-05-14
Add delete tags functionality

### LINKED ISSUE #9770: Delete tags that haven't yet been pushed

If you make a mistake in a tag, there's no way to delete it.

### Proposed solution

Allow people to delete tags that haven't yet been pushed to the remote. I'm not sure how the UX should work for this, but maybe just a context menu on the tag in the commit metadata and a delete option?

I'll defer to @ampinsk on how this should work.

rafeca:
Closes https://github.com/desktop/desktop/issues/9770

This PR is based on https://github.com/desktop/desktop/pull/9828

## Description

This PR adds the functionality to delete tags that haven't been pushed yet from GitHub Desktop.

In order to do so, a new context menu item has been added to commits that have unpushed tags. If the commit has one single unpushed tag, the menu iten will show a text like "Delete tag v1.0.0", but if there are more than one unpushed tags, the menu item will display a "Delete tags..." text and will show the tags that can get deleted via a submenu.

In order to implement that, the context menu behaviour has been changed to support submenus in context menus (see commit 55443cd).

@tierninho in order to test this PR you'll have to run `yarn build:dev` beforehand, since it modifies code from the main process.

### Screenshots

demo

## Release notes

Notes: [Added] Allow to delete tags that have not been pushed to the remote repository

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3071,8 +3071,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const gitStore = this.gitStoreCache.get(repository)
 
     await gitStore.deleteTag(name)
-
-    this.statsStore.recordTagDeleted()
   }
 
   private updateCheckoutProgress(
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3066,6 +3066,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this._closePopup()
   }
 
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _deleteTag(repository: Repository, name: string): Promise<void> {
+    const gitStore = this.gitStoreCache.get(repository)
+
+    await gitStore.deleteTag(name)
+  }
+
   private updateCheckoutProgress(
     repository: Repository,
     checkoutProgress: ICheckoutProgress | null
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3071,6 +3071,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const gitStore = this.gitStoreCache.get(repository)
 
     await gitStore.deleteTag(name)
+
+    this.statsStore.recordTagDeleted()
   }
 
   private updateCheckoutProgress(
--- a/app/src/ui/main-process-proxy.ts
+++ b/app/src/ui/main-process-proxy.ts
@@ -75,7 +75,7 @@ export function registerContextualMenuActionDispatcher() {
   ipcRenderer.on(
     'contextual-menu-action',
     (event: Electron.IpcRendererEvent, indices: number[]) => {
-      if (currentContextualMenuItems == null) {
+      if (currentContextualMenuItems === null) {
         return
       }
 
--- a/app/src/ui/main-process-proxy.ts
+++ b/app/src/ui/main-process-proxy.ts
@@ -74,16 +74,25 @@ let currentContextualMenuItems: ReadonlyArray<IMenuItem> | null = null
 export function registerContextualMenuActionDispatcher() {
   ipcRenderer.on(
     'contextual-menu-action',
-    (event: Electron.IpcRendererEvent, index: number) => {
-      if (!currentContextualMenuItems) {
-        return
-      }
-      if (index >= currentContextualMenuItems.length) {
+    (event: Electron.IpcRendererEvent, indices: number[]) => {
+      if (currentContextualMenuItems == null) {
         return
       }
 
-      const item = currentContextualMenuItems[index]
-      const action = item.action
+      let foundMenuItem: IMenuItem = {
+        submenu: currentContextualMenuItems,
+      }
+
+      // Traverse the submenus of the context menu until we find the appropiate index.
+      for (const index of indices) {
+        if (foundMenuItem == null || foundMenuItem.submenu == null) {
+          return
+        }
+
+        foundMenuItem = foundMenuItem.submenu[index]
+      }
+
+      const action = foundMenuItem.action
       if (action) {
         action()
         currentContextualMenuItems = null
--- a/app/src/ui/main-process-proxy.ts
+++ b/app/src/ui/main-process-proxy.ts
@@ -79,26 +79,34 @@ export function registerContextualMenuActionDispatcher() {
         return
       }
 
-      let foundMenuItem: IMenuItem = {
-        submenu: currentContextualMenuItems,
-      }
+      const menuItem = findSubmenuItem(currentContextualMenuItems, indices)
 
-      // Traverse the submenus of the context menu until we find the appropiate index.
-      for (const index of indices) {
-        if (foundMenuItem == null || foundMenuItem.submenu == null) {
-          return
-        }
-
-        foundMenuItem = foundMenuItem.submenu[index]
-      }
-
-      const action = foundMenuItem.action
-      if (action) {
-        action()
+      if (menuItem !== undefined && menuItem.action !== undefined) {
+        menuItem.action()
         currentContextualMenuItems = null
       }
     }
   )
+}
+
+function findSubmenuItem(
+  currentContextualMenuItems: ReadonlyArray<IMenuItem>,
+  indices: ReadonlyArray<number>
+): IMenuItem | undefined {
+  let foundMenuItem: IMenuItem | undefined = {
+    submenu: currentContextualMenuItems,
+  }
+
+  // Traverse the submenus of the context menu until we find the appropiate index.
+  for (const index of indices) {
+    if (foundMenuItem === undefined || foundMenuItem.submenu === undefined) {
+      return undefined
+    }
+
+    foundMenuItem = foundMenuItem.submenu[index]
+  }
+
+  return foundMenuItem
 }
 
 /** Show the given menu items in a contextual menu. */
--- a/app/src/lib/git/tag.ts
+++ b/app/src/lib/git/tag.ts
@@ -22,6 +22,21 @@ export async function createTag(
 }
 
 /**
+ * Delete a tag.
+ *
+ * @param repository        - The repository in which to create the new tag.
+ * @param name              - The name of the tag to delete.
+ */
+export async function deleteTag(
+  repository: Repository,
+  name: string
+): Promise<void> {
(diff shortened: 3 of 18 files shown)


tierninho:
This just randomly popped up in the console of this build (a4620407560193c316b6db8771c32b5def323380), however I do not see delete tags functionality despite running the `yarn build:dev`:

```
install.ts:23 `git -c credential.helper= push origin sdfsdf:sdfsdf --follow-tags --progress` exited with an unexpected code: 1.
stderr:
error: src refspec sdfsdf matches more than one
error: failed to push some refs to 'https://github.com/tierninho/Blah.git'
```

tierninho (dismissed):
- Functionally everything works great for OSX, however there is a 2-3 sec lag in getting the tag to be recognized by Desktop. For example, if I create a tag and immediately try to delete it, there is a lag which prohibits me from doing soon right away. Is this due to the dev env?

- Secondly, perhaps we should always show the `Delete tag` menu option, but just disable it if the logic does not apply? It would be consistent with the `View on Github` option here:
Screen Shot 2020-05-14 at 11 50 46 AM

  There is an edge case if we opt to go this route. If a pushed commit with an existing tag gets a new second tag, we would have to enable Delete tag option, but in the secondary context menu, only show the new tag as enabled.

--

In Windows after running `yarn build:dev` I am getting this error and I cannot successfully spin up:

Screen Shot 2020-05-14 at 12 03 08 PM

ruggi99:
@tierninho see #9396

rafeca:
> Secondly, perhaps we should always show the Delete tag menu option, but just disable it if the logic does not apply? It would be consistent with the View on Github option here

Thanks for pointing to the "View on GitHub" option, I agree we should be consistent here (I remember @ampinsk also mentioning that to me for the submenu, so we seem to all agree), so I'll change this.

> There is an edge case if we opt to go this route. If a pushed commit with an existing tag gets a new second tag, we would have to enable Delete tag option, but in the secondary context menu, only show the new tag as enabled.

Yes I think that this should be fine, do you think this can cause problems/confussion?

> Functionally everything works great for OSX, however there is a 2-3 sec lag in getting the tag to be recognized by Desktop. For example, if I create a tag and immediately try to delete it, there is a lag which prohibits me from doing soon right away. Is this due to the dev env?

Yup, this delay is caused by the network request that is done to check for unpushed tags. This delay will probably go away once https://github.com/desktop/desktop/issues/9800 is fixed.

app/src/ui/main-process-proxy.ts
```diff
@@ -74,16 +74,25 @@ let currentContextualMenuItems: ReadonlyArray<IMenuItem> | null = null
...
-    (event: Electron.IpcRendererEvent, index: number) => {
-      if (!currentContextualMenuItems) {
+    (event: Electron.IpcRendererEvent, indices: number[]) => {
+      if (currentContextualMenuItems == null) {
         return
       }
-      if (index >= currentContextualMenuItems.length) {
-        return
+
+      let foundMenuItem: IMenuItem = {
```
niik, line 82: This is one of the places where TypeScript not only won't have our backs, it'll even stab us in it when we're not looking.

As a reader of this code I look at this statement and think cool, foundMenuItem will never be `undefined`  or `null`. But that's not necessarily true since `foundMenuItem.submenu[index]` doesn't come with any guarantees that the index we're looking for exists even though the type system fools us into believing that.

So in a worst-case scenario here where the `currentContextualMenuItems` global variable ends up gets swapped at an inopportune time we _could_ (most likely we wouldn't though) end up with `foundMenuItem` that's `undefined` and :boom:

Given that we don't anticipate ever having an unbounded depth of submenus (because that'd be 🍌:banana:) I wonder if we can both categorically guarantee that it won't be undefined and at the same time uphold out [immutability](https://github.com/desktop/desktop/blob/development/docs/contributing/engineering-values.md#immutability-is-good) preference by adding a recursive helper function. Something like this perhaps

```ts
const foundMenuItem = findSubmenuItem(
  { submenu: currentContextualMenuItems },
  indices
)

if (foundMenuItem !== undefined && foundMenuItem.action !== undefined) {
  foundMenuItem.action()
  currentContextualMenuItems = null
}
```

```ts
/**
 * Recursively traverse the submenu(s) of the given menu item following
 * the indices provided until the desired submenu item could be found.
 *
 * Returns undefined if menu could not be traverse in accordance with
 * the indices parameter.
 */
function findSubmenuItem(
  menuItem: IMenuItem | undefined,
  indices: ReadonlyArray<number>,
  depth: number = 0
): IMenuItem | undefined {
  if (
    menuItem === undefined ||
    menuItem.submenu === undefined ||
    depth >= indices.length
  ) {
    return undefined
  }

  return depth === indices.length - 1
    ? menuItem.submenu[indices[depth]]
    : findSubmenuItem(menuItem.submenu[indices[depth]], indices, depth + 1)
}
```

niik (changes requested):
Looking good @rafeca, a few nitpicks from me and one theoretical concern, happy to discuss that when we sync up later today

app/src/ui/main-process-proxy.ts
rafeca: Good catch! You concern is real though and I think it's quite possible that this could happen under some race condition (although it would be a very rare situation). 

This makes we wonder whether apart from fixing this specific `undefined` issue we should think about adding some logic to warranty that the `currentContextualMenuItems` object doesn't change between the time when the context menu gets opened and the time that an item gets clicked: e.g it would also be very bad if a user clicks on a tag to delete it and we end up deleting a different one because a new tag appears on the list at the appropiate moment.

Regarding your proposed solution, I'm not convinced about changing a simple for loop by a recursive function (I'd rather add an `if` check when accessing the array index), since making the function recursive doesn't solve the root problem of TypeScript [not returning `undefined` for index lookups](https://github.com/microsoft/TypeScript/issues/13778) and imho makes the logic harder to reason.

I agree about prefering immutabilty over mutabilty (and requiring it on structures that are shared between functions), but I think having a local mutation sometimes is a more pragmatic approach (see: https://twitter.com/dan_abramov/status/1099777642885189633).

Maybe I can move the `for` loop to its own method so the variable with a mutable assignment has less scope (even if the original method grows in size)? something like:

```ts
function findSubmenuItem(
  currentContextualMenuItems: ReadonlyArray<IMenuItem>,
  indices: ReadonlyArray<number>
): IMenuItem | undefined {
  let foundMenuItem: IMenuItem | undefined = {
    submenu: currentContextualMenuItems,
  }

  // Traverse the submenus of the context menu until we find the appropiate index.
  for (const index of indices) {
    if (foundMenuItem === undefined || foundMenuItem.submenu === undefined) {
      return undefined
    }
    foundMenuItem = foundMenuItem.submenu[index]
  }

  return foundMenuItem
}
```

niik: > (I'd rather fix the type of foundMenuItem and add an additional if check

Yeah, that was my first thought as well but TS type narrowing bites us in the behind here as it'll always infer the type of `foundMenuItem` to be `IMenuItem` since the last executable statement of the `for` narrows it to `IMenuItem`. So even explicitly typing the variable as `IMenuItem | undefined` won't afford us any compile-time guarantees in case someone were to attempt to remove the added `if (foundMenuItem !== undefined)` check.

> Maybe I can move the for loop to its own method so the variable with a mutable assignment has less scope?

Yes, I think that's a lovely compromise!

tierninho:
I opened this PR in Windows, but every time I try to create or delete a tag it crashes with this error:

```
TypeError: indices is not iterable
    at EventEmitter.<anonymous> (http://localhost:3000/build/renderer.js:125188:29)
    at EventEmitter.emit (events.js:203:13)
    at Object.onMessage (electron/js2c/renderer_init.js:2399:16)

Version: 2.5.1-beta1
OS: Windows 10.0.18362
```

niik:
@tierninho You’ll have to rebuild the main process for this to work. `yarn build:dev` before your `yarn start` should do the trick but let me know if it doesn’t

app/src/ui/main-process-proxy.ts
rafeca: >  it'll always infer the type of foundMenuItem to be IMenuItem since the last executable statement of the for narrows it to IMenuItem

Luckily in this specific case, since the block inside a for loop is not guaranteed to be executed (e.g empty array) TS can't narrow the type, but I agree that this could bite us in any other scenario... Unfortunately this is something I don't expect in the short-mid term to get fixed in TS since it would cause tons of errors on existing apps 😟

niik (changes requested):
Just a tiny question remaining. I'm fine with not addressing it but it felt like such a small change

app/src/lib/stores/app-store.ts
```diff
@@ -3066,6 +3066,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this._closePopup()
   }
 
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _deleteTag(repository: Repository, name: string): Promise<void> {
+    const gitStore = this.gitStoreCache.get(repository)
+
+    await gitStore.deleteTag(name)
+
+    this.statsStore.recordTagDeleted()
```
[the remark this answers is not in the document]
rafeca: Really good point! I keep forgetting that git failures don't throw...

tierninho (commented):
Confirming that functionality works great on both OSX/Win however I spotted this edge case:

- User can delete a tag while commits are pushed.
  - Add a tag and push
  - While the tag is being right-click to delete the tag
  - Wait until push finishes in background (Delete modal should still be open).
  - Notice that at this point, a user should not be able to delete a tag
  - Click `Delete` and it is removed
  - Click `Fetch` to bring it back

Also, a suggestion:
- Enabling `Delete tags...` in this instance was confusing as I thought my menu was broken.
Screen Shot 2020-05-26 at 2 29 52 PM
  - Should we disable the `Delete tags...` menu option if all tags cannot be deleted? It would match a single tag menu item like so:
Screen Shot 2020-05-26 at 2 30 15 PM
and we would only make the menu intact when a fresh tag is added like so:
Screen Shot 2020-05-26 at 2 31 43 PM

niik (approved):
Awesome, for the sake of  getting us set up for a beta I'm going to merge this as-is and leave the suggestions @tierninho made to be considered separately

afsaneh751 (changes requested):
[]bc1qhzktlv392tq37cwtsv0wfy32rcq67lcagay40z

Issue #9679, opened 2020-04-30
Provide a way to more easily work with a maintained fork


billygriffin:
### Describe the feature or problem you’d like to solve

The 2.4 release improved the way forks work in Desktop assuming people are using their fork to contribute to upstream. We discussed the divergence between people who are using a fork to contribute vs. those who are contributing to a fork independent of the upstream repo in #9170. We were likely too heavy-handed in optimizing the workflow for people using a fork primarily to contribute to upstream, and broke the workflow for those who were contributing to a maintained fork independent of upstream.

The goal with this issue is to ensure people working on forks to contribute are able to benefit from the improvements we made, but also account for those who experienced a regression from the recent release.

### Proposed solution

We're proposing a repo setting that allows users to specify how they're working with their fork, and several things shift based on that:
1. When branches are created, which repo's default branch they're based off of.
2. Which repo's page "View on GitHub" opens in the browser.
3. Which PRs are shown in the PR list, based on which repo the PRs are targeting.
4. Which repo new issues are created in.

We want to ensure that people have the opportunity to make this decision and we aren't making as many default assumptions, so some kind of prompt is likely warranted when they're first opening a fork, and also potentially when cloning a fork into Desktop or creating a fork in Desktop.

### Additional context

This is related to #9670 and #9500.

ampinsk:
## Prompt to ask about fork type

<img width="474" alt="Fork_Decision" src="https://user-images.githubusercontent.com/10404068/80765995-36da7e00-8af9-11ea-9c47-56cfa474ed3d.png">

## Repository setting

<img width="618" alt="Fork_Setting" src="https://user-images.githubusercontent.com/10404068/80765998-380bab00-8af9-11ea-9a92-3919ccd3b611.png">


## Setting behavior

**Contribution Fork:** A fork being used solely to contribute to the upstream repo
**Maintained Fork:** A fork being used for its own purposes

||Contribution Fork|Maintained Fork|
|---|---|---|
|Creating a new branch from default|<img width="400" alt="Fork-New-Branch-Parent" src="https://user-images.githubusercontent.com/10404068/80765694-7b194e80-8af8-11ea-99c8-f5cce61a2992.png">New branch will be created off of the parent repo's default branch|<img width="400" alt="Fork-New-Branch_01" src="https://user-images.githubusercontent.com/10404068/80765691-79e82180-8af8-11ea-85f8-9caa1b949bc0.png">New branch will be created off of the current repo's default branch|
[...]

PR #9805, opened 2020-05-16
add storage for fork repo workflow settings

outofambit:
Supports #9679 

## Description

- add a database persisted collection of repo workflow settings to the `Repository` model
- includes a setting for whether a fork is mean to contribute upstream or to itself

### DIFF
--- a/app/src/lib/databases/repositories-database.ts
+++ b/app/src/lib/databases/repositories-database.ts
@@ -1,5 +1,6 @@
 import Dexie from 'dexie'
 import { BaseDatabase } from './base-database'
+import { WorkflowPreferences } from '../../models/workflow-preferences'
 
 export interface IDatabaseOwner {
   readonly id?: number | null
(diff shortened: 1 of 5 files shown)


rafeca:
I've added 2 commits to this PR to make the `WorkflowSettings` object non-nullable: this makes it more ergonomic to consume in the UI

app/src/lib/databases/repositories-database.ts
```diff
@@ -48,6 +49,8 @@ export interface IDatabaseRepository {
   /** The last time the stash entries were checked for the repository */
   readonly lastStashCheckDate: number | null
 
+  readonly workflowPreferences: WorkflowPreferences
```
niik, line 52: I think this needs to be optional since we're not running any database upgrade/migration to ensure that it exists on every IDatabaseRepository

```suggestion
  readonly workflowPreferences?: WorkflowPreferences
```

rafeca: Thanks for the suggestion! this is something I messed up in Evelyn's PR 🤡 , gonna change it in a separate commit since it requires more changes outside of this file

PR #9499, opened 2020-04-10
Git Repo Config

say25:
Closes #9449

## Description

- Add a place to modify Git Config values for a particular repository.

### Screenshots

Screen Shot 2020-08-08 at 4 47 53 PM

## Release notes

Notes: [Added] Ability to modify git config on a per repository basis.

PR #9806, opened 2020-05-16
use fork workflow preferences if possible

outofambit:
~~👋🏼 _based on https://github.com/desktop/desktop/pull/9805, please review that first!_~~ - Done!

Supports #9679

## Description

- Modify logic in `getNonForkGitHubRepository` to respect the new workflow setting

## Notes

- maintains previous logic default if the workflow setting hasn't been set
- @rafeca and i consolidated the majority of, if not all, fork determining logic in desktop into this method

### DIFF
--- a/app/src/lib/branch.ts
+++ b/app/src/lib/branch.ts
@@ -18,7 +18,7 @@ import {
 export function findDefaultUpstreamBranch(
   repository: RepositoryWithGitHubRepository,
   branches: ReadonlyArray<Branch>
-) {
+): Branch | null {
   const githubRepository = getNonForkGitHubRepository(repository)
 
   // This is a bit hacky... we checked if the result of calling
@@ -33,9 +33,11 @@ export function findDefaultUpstreamBranch(
     return null
   }
 
-  return branches.find(
+  const foundBranch = branches.find(
     b =>
       b.type === BranchType.Remote &&
       b.name === `${UpstreamRemoteName}/${githubRepository.defaultBranch}`
   )
+
+  return foundBranch !== undefined ? foundBranch : null
 }
--- a/app/src/lib/branch.ts
+++ b/app/src/lib/branch.ts
@@ -1,25 +1,41 @@
 import { Branch, BranchType } from '../models/branch'
 import { UpstreamRemoteName } from './stores'
+import {
+  RepositoryWithGitHubRepository,
+  getNonForkGitHubRepository,
+} from '../models/repository'
 
 /**
- * Finds the remote branch for a branch in the upstream repository
+ * Finds the default branch of the upstream repository of the passed repository.
  *
- * For example:
- * If official/funnel has the branch `development`, then running
- * this on the branches from the fork outofambit/funnel with
- * `development` as the specified branch name will find the
- * branch `remotes/upstream/development`
+ * When the passed repository is not a fork or the fork is configured to use itself
+ * as the ForkContributionTarget, then this method will return null. Otherwise it'll
+ * return the default branch of the upstream repository.
  *
- * @param branchName short name of the branch in the upstream repo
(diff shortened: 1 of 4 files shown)


rafeca:
I've added a few commits to this PR, first one that uses the `getForkContributionTarget()` method to infer the default value of `ForkContributionTarget` when it's undefined (this way we can reuse this method in the UI).

Then a few commits to make sure that we use everywhere the `getNonForkGitHubRepository()` method:

* To calculate the inferred branch for the NDDB (even if it's currently disabled, it's better to fix this now, since we may rely on that logic in the future).
* The default branch used for the create branch dialog.

app/src/lib/branch.ts
```diff
@@ -1,25 +1,41 @@
...
+  const githubRepository = getNonForkGitHubRepository(repository)
+
+  // This is a bit hacky... we checked if the result of calling
+  // getNonForkGitHubRepository() is the same as the associated
+  // gitHubRepository. When this happens, we know that either the
+  // repository is not a fork or that it's a fork with
+  // ForkContributionTarget=Self.
+  //
+  // TODO: Make this method return the default branch of the
+  // origin repository instead of null in this scenario.
```
rafeca, line 31: This works fine but it's not elegant... I'll fix this by refactoring this method in a separate commit. I started doing that but I need to change quite a few parts of the application since I need to make this method async.

Ideally, we should be able to unify the logic in `infer-comparison-branch.ts` with this one and create a generic function that always returns the default branch of a `Repository` object.

say25:
With regards to #9499, I would assume I'll have to rebase off this at some point?

rafeca:
> With regards to #9499, I would assume I'll have to rebase off this at some point?

Yup I think so, but let's wait until this PR (and specially https://github.com/desktop/desktop/pull/9842) get merged (which hopefully will happen very soon).

This will also change the layout of the tabs to be vertical, giving more space for new tabs

niik:
Merged #9805, retargeted this PR, and resolved conflicts

app/src/lib/branch.ts
```diff
@@ -1,25 +1,41 @@
...
+  if (githubRepository === repository.gitHubRepository) {
+    return null
```
niik, line 33: Is there value in being able to differentiate between this scenario and when the default branch could not be found? I.e. we return `null` here but `undefined` below from the `.find`. In general I find I'm often confused when I encounter methods with tri-state returns like this so if there's value in callers being able to differentiate between this then I think it should be clearly documented in the function documentation, if not we should return `undefined` here or `null` below.

```suggestion
    return undefined
```

rafeca: Thanks for the suggestion! yup there's no value doing so, it was just muscle memory of returning `null` without realizing that lower down the function there's `find()` that can return `undefined`.

I think in this specific case I'd rather leave this `return null` statement and add a check on `find()` to always return `null` (since anyways the only user of this method is casting the `undefined` to `null`). 

I think these kind of issues could get reduced substantially if we enforced typing the return type of functions (this could also potentially highlight some issues with our types where sometimes TypeScript infers `any` somewhere and this gets returned in a function.

niik: > I think in this specific case I'd rather leave this return null statement and add a check on find() to always return null (since anyways the only user of this method is casting the undefined to null).

Yep, that's fine with me!

> I think these kind of issues could get reduced substantially if we enforced typing the return type of functions

I think you're right but it's also a tradeoff, most of the time it's really convenient to be able to rely on the inference. I wouldn't want to mandate explicit typing everywhere but there's certainly lots of places where adding it as a form of codified expectation of the method is a great idea and I've never rejected a contribution for using explicitly typed return params.

PR #9828, opened 2020-05-19
Avoid performing network operations to calculate unpushed tags

### LINKED ISSUE #9800: Perform check for unpushed tags without doing a network request
## Problem

In order to find out which tags need to be pushed, GitHub Desktop calls `git push --follow-tags --dry-run` and parses its response searching for tags that will get pushed (see https://github.com/desktop/desktop/pull/9576 for details).

This approach gives very reliable information about the tags that need to get pushed and aligns well with the standard way of pushing tags via `git` (which is by using the `--follow-tags` argument), but since it does a network request it has a few drawbacks:

- It's much slower than any other `git` operation (this is why we needed to add a caching layer).
- Doesn't work when the user is offline.

Apart from this, we've discovered an edge case that degrades drastically the experience when using a corporate git server that enforces 2fa on pushes (see https://github.com/desktop/desktop/issues/9787#issuecomment-628676921).

This last scenario is quite severe, since there's no workaround to it and causes major invonveniences to the affected users.

## Proposed solution

A quite simple solution proposed by @niik to avoid doing network requests is to downscope the problem to only track the unpushed tags that get created from Desktop. This way we can just store in a simple list the tags that have been created and treat the tags in that list as unpushed tags.

Then, once a `push` operation finishes correctly, we can clear the list of unpushed tags for that remote.

With this solution, though, we won't be able to rely solely on the `--follow-tags` functionality when pushing, and instead we'll need to modify the `push` command to also include each of the unpushed tags so we can be sure that all of them get pushed.

(this is because the list of unpushed tags will contain every single tag that has not been pushed to the remote, while `--follow-tags`, depending on the circumstances, will only push a subset of them).

## Advantages of this solution

* It doesn't require any network operation (or even any `git` command), so it's way faster.
* Because of the previous point, we won't need any caching layer so it's a much simpler approach.
* Doesn't add potential side-effects to the most common Desktop flows (e.g a user that does not create tags).

## Potential drawbacks of this solution

* We're not going to be able to know if a tag created outside of Desktop has been pushed (we're going to treat all of them as pushed).
* It's possible that the list of unpushed tags gets inconsistent/out of sync (when a push operation fails but some the tags could get pushed, when a user pushes from outside of Desktop, when a tag gets deleted from the remote, ...).
* We're going to possibly have to handle the scenario where there's a conflict when pushing a tag that already exists on the remote.

## Conclusions

Overall, the proposed solution replaces some of the tradeoffs of the current solution that could affect any user (whether they are creating tags or not) by another set of tradeoffs that only affect users when they are actually creating tags. Since creating tags is not a very common action in `git` (a lot of users don't ever create tags and even users who create tags do much more rarely  than e.g creating commits), this seems like a good direction to move.

Another important point is that the identified drawbacks from this solution are mostly cosmetic (an outdated list of unpushed tags won't affect the main behaviour of Desktop, it'll just cause a small temporary inconvinience to users), while with the currently implemented solution we have some issues that are quite severe (like https://github.com/desktop/desktop/issues/9787#issuecomment-628676921).

## Other potential solutions

Another proposed solution was to modify the `refspec` that `git` uses to do the fetches to be able to store the tags for each remote separately (e.g in `refs/remotes/origin/tags/xxx`).

On paper, this solution will solve the tradeoffs of the currently implemented approach while avoiding the new set of tradeoffs from the proposed approach.

The main risk here though is that modifying the `refspec` used for every single `git fetch` command can cause another set of problems and edge cases to users with custom refspecs, and these problems will again affect the users quite severily whether they create tags or not. Therefore, in order to implement this, we would need to spend some time investigating potential edge cases and fix them, so we're instead leaning forward implementing the proposed approach first (which is much simpler), and only considering this alternate solution in the future if we get complains from users.

rafeca:
Closes https://github.com/desktop/desktop/issues/9800

## Description

This PR addresses the issues described in https://github.com/desktop/desktop/issues/9800 by storing locally a list of tags that have been created but not yet pushed to any remote repository. Once a push operation finishes, the list of unpushed tags gets resetted.

With this we avoid doing network operations to calculate the tags that need to be pushed.

### Technical considerations

* The list of unpushed tags gets persisted in local storage using the repository hash as key. Since the data is just an array of strings and we don't expect it grow bigger than than a few elements (at most a few tens of elements), this simpler approach should not degrade performance and it's very simple to implement.
* The push command has been changed to explicitly push all the unpushed tags instead of using `--follow-tags`. This gives us a much more predictable outcome of the push operation and it's safe to perform since we're only telling `git` to push the tags that Desktop tracks.

### Next steps

As an improvement on top of this PR, we can parse the `push` operation output to know if there was a failure pushing any of the tags. With this we can infer the type of error and show better description errors.

## Screenshots

N/A

## Release notes

Notes: [Improved] Do not perform background network operations to calculate tags that need to be pushed.


app/src/lib/git/push.ts
```diff
@@ -67,16 +69,15 @@ export async function push(
...
+  }
   if (!remoteBranch) {
     args.push('--set-upstream')
   } else if (options.forceWithLease === true) {
     args.push('--force-with-lease')
   }
 
-  if (enableGitTagsCreation()) {
-    args.push('--follow-tags')
-  }
```
rafeca, line 78: I removed the check for the feature flag, since if the feature flag is disabled the array of tags will get null

app/src/lib/git/push.ts
```diff
@@ -53,6 +54,7 @@ export async function push(
   remote: IRemote,
   localBranch: string,
   remoteBranch: string | null,
+  tagsToPush: ReadonlyArray<string> | null,
```
rafeca, line 57: As opposed to what we talked @niik , I've ended up adding an independent argument to indicate the tags to push.

I've changed my mind because on L76 we set the `--set-upstream` argument when `remoteBranch` is `null`.

If we changed the signature of this method to accept a single array of tuples that would englobe both branches and tags (e.g `ReadonlyArray<[string, string | null]>` where the first element of the tuple is the local ref and the second one is the remote ref), then we wouldn't be able to know when exactly we should set the `--set-upstream` param (we could set it when the remote reference was `null`, but then it would mean setting `--set-upstream` every time we push tags).

Anyways, at first glance it doesn't seem like adding the `--set-upstream` arg when pushing tags could cause issues, but I don't think that the marginal improvement on cleanliness for this method is worth the risk

app/src/lib/stores/git-store.ts
```diff
@@ -148,6 +146,9 @@ export class GitStore extends BaseStore {
     private readonly statsStore: StatsStore
   ) {
     super()
+
+    const rawTags = localStorage.getItem(getTagsToPushKey(repository))
+    this._tagsToPush = rawTags ? JSON.parse(rawTags) : []
```
rafeca, line 151: I've opted to store the list of tags in JSON format since this will be a short list of strings. We could also have just joined the array of tags using whitespace since tags can't have whitespace, but I'm erring on the safe side here.

tierninho (dismissed):
Took this PR for a spin and I did not notice anything wonky. After creating and pushing a tag, I can confirm I did not see any network requests nor the `--follow-tags` flag, as expected. 

~~Still blocked on Windows testing until the Electron package issue is resolved.~~ Issue was resolved and WIN was tested 👍

ruggi99:
> Still blocked on Windows testing until the Electron package issue is resolved.

https://github.com/feross/cross-zip/issues/23. That bug has been resolved in cross-zip@3.1.0. You can upgrade cross-zip dependence and it should work.

tierninho:
Thanks @ruggi99 Any chance you got it working? I have had no luck. Still results in the same error for me.

ruggi99:
@tierninho yes. Of course, you can move Desktop to a path with no spaces, eventually creating a new partition. Apart from this there is no official support from yarn for updating the indirect dependencies (cross-zip of electron-packager in this case) but you can get around it. You can simply delete the cross-zip dependency in `yarn.lock` file, as if it had never been installed and then install cross-zip which will be installed to the latest version in the semver range. Finally run `yarn` or `yarn install`. This has worked for me.

niik (changes requested):

niik (changes requested):
Just writing down the things I can remember from our Zoom.

- [x] Investigate removing the tags to push local storage entry when removing a repository
- [x] Consider extracting getter/setter/delete? methods for the localStorage entry into separate methods that can be documented and reused (if we chose to remove the entry along with the repository model)

tierninho:
Many thanks @ruggi99 It works now!

rafeca:
I've created the abstraction layer to access localStorage in fe3edc6 and implemented logic to remove the tags from localStorage when the repository is removed in 22016b6

niik (approved):
Thanks for all your hard work here @rafeca! I really appreciate the care in documenting the new exported methods.

PR #9842, opened 2020-05-21
Add tab for fork settings in repository settings

rafeca:
This PR is based on https://github.com/desktop/desktop/pull/9806 

Supports #9679

## Description

This PR adds the fork settings tab in the repository settings dialog and implements the needed logic to update the workflow preferences of a repository when the user changes it from the UI.

A new feature flag has been created to be able to show/hide this feature.

**Note**: This PR doesn't yet add the dialog when cloning a forked repository, this will be done in followup commits or as a separate PR.

### Screenshots

image

## Release notes

Notes: no-notes


app/src/models/repository.ts
```diff
@@ -88,12 +90,37 @@ export type RepositoryWithGitHubRepository = Repository & {
...
+ * RepositoryWithForkedGitHubRepository if it returns true.
+ */
+export function isRepositoryWithForkedGitHubRepository(
+  repository: Repository
+): repository is RepositoryWithForkedGitHubRepository {
+  return (
+    isRepositoryWithGitHubRepository(repository) &&
+    repository.gitHubRepository.parent !== null
+  )
+}
```
rafeca, line 122: This helper method allows us to be able to pass `RepositoryWithForkedGitHubRepository ` objects around and not have to check whether `parent` is `null` when we now it isn't.

app/src/ui/repository-settings/fork-settings.tsx
```diff
@@ -0,0 +1,96 @@
...
+          <input
+            type="radio"
+            id={RadioButtonId.Self}
+            value={ForkContributionTarget.Self}
+            checked={
+              this.props.forkContributionTarget === ForkContributionTarget.Self
+            }
+            onChange={this.onForkContributionTargetChanged}
+          />
+          <label htmlFor={RadioButtonId.Self}>To use for my own purposes</label>
```
Rexogamer, line 51: The grammar here is a bit messy, what about something like:
```suggestion
          <label htmlFor={RadioButtonId.Self}>For my own purposes</label>
```

tierninho:
I spun up a8b36aa85dc0db0274f88d37f8399be65a1e42a7 on OSX with `yarn && yarn build:dev && yarn start` and am seeing a missing tab in the settings:

Screen Shot 2020-05-21 at 8 34 40 AM

rafeca:
@tierninho you're right! thanks for finding this! just fixed this in the latest commit

Note that now the repository setting only appears for repositories that are forks

tierninho:
Thanks for the fix! Here are some issues I found so far:

- On a non-fork, open the modal and the click the Remote tab, press down arrow a few times and the settings disappear like so:

  Screen Shot 2020-05-21 at 9 21 22 AM

  A side-effect also prevents me from tabbing back/forth to the `Cancel` button.

- If the Repo Settings are opened to the Fork tab, and a user leaves Desktop to go to another repo on Github.com, and then clicks `Open in Desktop`, the Repo settings for the fork remain in the foreground but the background changes to the new repo. Should we close the settings modal? Any chances made in the settings will still take effect.

  Screen Shot 2020-05-21 at 9 28 38 AM

rafeca:
> On a non-fork, open the modal and the click the Remote tab, press down arrow a few times and the settings disappear like so:

Thanks! I've just fixed this one on the latest commit 😃 

> If the Repo Settings are opened to the Fork tab, and a user leaves Desktop to go to another repo on Github.com, and then clicks Open in Desktop, the Repo settings for the fork remain in the foreground but the background changes to the new repo. Should we close the settings modal? Any chances made in the settings will still take effect.

Wow! such an edge case! nice catch! I've checked and this issue seems to happen with other modals (like the create branch modal), I'm gonna create an issue for it and fix it across all of them since it seems related to how modals work.

app/src/ui/repository-settings/fork-settings.tsx
rafeca: Thanks for the suggestion!

Since I'm not a native speaker I'll leave this to @ampinsk and @billygriffin .

Context of this text:

image

ampinsk: "For my own purposes" makes sense to me!

tierninho:
Did spot this error while testing, but not sure if its related. Seems the api failed?

Screen Shot 2020-05-27 at 8 05 27 AM

rafeca:
> Did spot this error while testing, but not sure if its related. Seems the api failed? Not sure what caused it...

It seems unlikely to me that this could be related to this PR, but if you notice the error again while testing it let me know and I'll investigate!

app/src/models/github-repository.ts
```diff
@@ -52,6 +52,10 @@ export class GitHubRepository {
   }
 }
 
+export type ForkedGitHubRepository = GitHubRepository & {
+  parent: GitHubRepository
+}
```
outofambit, line 57: i think `parent` should be readonly here too.

we could also explicitly say that `fork` will return `true` in the typing here. i don't think typescript will figure that out on its own

```suggestion
export type ForkedGitHubRepository = GitHubRepository & {
  readonly parent: GitHubRepository
  readonly fork: true
}
```

app/src/ui/repository-settings/repository-settings.tsx
```diff
@@ -198,6 +233,19 @@ export class RepositorySettings extends React.Component<
       }
     }
 
+    if (
```
outofambit, line 236: took me a minute to read this `if` statement correctly
```suggestion
    // only update this if it will be different from what we have stored
    if (
```

app/src/ui/tab-bar.tsx
```diff
@@ -56,9 +56,9 @@ export class TabBar extends React.Component<ITabBarProps, {}> {
     direction: 'next' | 'previous',
     index: number
   ) => {
-    const children = this.props.children as ReadonlyArray<JSX.Element> | null
+    const children = React.Children.toArray(this.props.children)
```
outofambit, line 59: nice!

app/src/models/github-repository.ts
rafeca: Good catch! I was thinking about whether we can remove the `fork` property now and use everywhere the new `isRepositoryWithForkedGitHubRepository` method (which is preferred since it will narrow down the type), but in the meantime it's good to have `fork: true` in the type!

app/src/ui/tab-bar.tsx
rafeca: For the future reader, the changes in `<TabBar/>`  were needed to fix [the issues](https://github.com/desktop/desktop/pull/9842#issuecomment-632272679) that @tierninho reported.

They were caused by having a child in `<TabBar/>` that is `false`, which is quite common pattern in JSX when wanting to conditionally render stuff:

```jsx
<TabBar>
  <span>Item 1</span>
  {showSecondItem && (<span>Item 2</span>)}
</TabBar>
```

`React.Children.toArray()` filters out nullable children, plus it converts the structure to an actual array (`this.props.children` is actually an array-like object, and does not implement all `Array.prototype` methods).

app/src/ui/repository-settings/fork-settings.tsx
```diff
@@ -0,0 +1,96 @@
+import * as React from 'react'
+import { DialogContent } from '../dialog'
+import { ForkContributionTarget } from '../../models/workflow-preferences'
+import { RepositoryWithForkedGitHubRepository } from '../../models/repository'
+
+interface IGitIgnoreProps {
```
niik, line 6:
```suggestion
interface IForkSettingsProps {
```

app/src/ui/repository-settings/fork-settings.tsx
```diff
@@ -0,0 +1,96 @@
...
+  ) => void
+}
+
+enum RadioButtonId {
+  Parent = 'ForkContributionTargetParent',
+  Self = 'ForkContributionTargetSelf',
+}
+
+/** A view for creating or modifying the repository's gitignore file */
+export class ForkSettings extends React.Component<IGitIgnoreProps, {}> {
```
niik, line 20:
```suggestion
export class ForkSettings extends React.Component<IForksSettingProps, {}> {
```

though I bet you that'll need a prettify run

app/src/models/github-repository.ts
```diff
@@ -52,6 +52,11 @@ export class GitHubRepository {
   }
 }
 
+export type ForkedGitHubRepository = GitHubRepository & {
```
niik, line 55: I would love it if we could document this similar to how we've documented `RepositoryWithGitHubRepository`

PR #9864, opened 2020-05-25
Use check runs API for PR status

### LINKED ISSUE #9819: Use Checks API for PR combined status

GitHub Desktop displays head commit combined status next to PR informations (in the PR list or on the branch selector button). But it only requests the [status API](https://developer.github.com/v3/repos/statuses/), and some services (such as GitHub Actions) uses the [checks API](https://developer.github.com/v3/checks/), and they don't appear in the interface.

### Proposed solution

We could also request the checks API in addition to the status API, and display a combined status as it's already the case, so no UI change. This way all users will be able to see the combined PR status regardless of the used CI services.

### Additional context

Both endpoints are independent, and the checks API does not provide a "combined status" out of the box as the status API does. Therefore, I think we will have to determine by ourselves what the combined status is, based on both requests results.

If you think this could be a great addition, I would gladly work on it! 🙌  (with a little bit of help though 😇 )

Raul6469:
Closes #9819 

## Description

This will display a combined PR status from both the [status API](https://developer.github.com/v3/repos/statuses/) and the [checks API](https://developer.github.com/v3/checks/runs/)

As discussed in the issue, the checks API is still in public beta, but I don't know if it's still evolving frequently.

This is my first substantial contribution, so I mark it as a draft as I am unfamiliar with the codebase. That's also why I did not add tests yet, in case the whole thing is wrong 😇 

### Screenshots

No visible UI changes, the usual PR status badge and icons will appear when there are checks on the PR

image


niik:
:wave:  @Raul6469, first off, thanks for wanting to contribute to Desktop!

I don't want to forego the discussion that needs to happen in #9819 before we can decide if this is something we are able to tackle now but I did want to jump in here real quick and say that should we choose to implement this now we need to consider GitHub Enterprise Server. Actions hasn't launched there yet so there API doesn't exist and even if/when it does we'll need to support a diverse set of versions of enterprise server.

Raul6469:
Ok, I perfectly understand! 👍 
It's true that I did not think of GitHub Enterprise servers. I'm closing the PR as it's far too early for making changes

niik:
Hey @Raul6469, don’t sweat the enterprise stuff, that’s what we’re here for. We can help you make the necessary changes and do the compatibility testing when we come to that.

I hope my comment didn’t come off as negative towards the change itself. The little I saw of the code looked great and supporting checks is something we’re gonna have to get around to supporting sooner or later IMO.

My intention was to convey to the team that my comments here shouldn’t be interpreted as me taking a stance in the discussion about whether to prioritize this work now while still leaving a note for future reviewers to bear the enterprise caveat in mind.

Once we’ve ironed out the details and prioritization in the issue we’d be happy to review your PR

niik:
Hi @Raul6469, sorry for the long wait, hope you had a lovely summer!

We've finally dug ourselves out from underneath a pile of work and are ready and excited to take a look at this. We're experimenting with building GitHub Desktop itself using GitHub Actions in #9680 so we have a great opportunity to test this work against desktop/desktop itself.

I'm going to reopen this and bring it up to date with a merge from development and take a look at what work we need to do in  order to support GitHub Enterprise so you'll likely see a couple of commits from me here before we can get @rafeca to do a review.

3koonce:
I really have no idea what I am doing but I believe all my accounts have
been compromised

niik:
@rafeca This is ready for an initial look. I went down the path of treating both the legacy commit statuses and the modern check runs as check runs to simplify the logic a bit and I think that turned out quite well. You'll notice that I haven't used `assertNever` here and that's because the checks API is still in preview and it's not unlikely that we could see additions in the conclusions type for example. So instead of being strict here I opted for a solution which would be resilient (albeit not necessarily accurate) against API changes.

app/src/lib/stores/commit-status-store.ts
```diff
@@ -261,13 +292,46 @@ export class CommitStatusStore {
...
+      this.cache.set(key, { check: null, fetchedAt: new Date() })
+      return
+    }
+
+    let status: APICheckStatus
+    let conclusion: APICheckConclusion | null = null
+
+    if (checks.some(isIncompleteOrFailure)) {
+      status = 'completed'
+      conclusion = 'failure'
```
rafeca, line 323: Super-small-nitpick: can we move the logic in `isIncomplete()` into the `isFailure()` method and have only that method?

While I understand the distinction between the two, I think here it creates more confusion than clarity, since `isIncomplete` can easily be confused with the `complete` status on the API (when reading this I initially thought that it made no sense to set the combined status to "complete" when there's any incomplete step 😆).

app/src/ui/branches/ci-status.tsx
```diff
@@ -83,43 +84,87 @@ export class CIStatus extends React.PureComponent<
...
-    default:
-      return assertNever(state, `Unknown state: ${state}`)
+    case 'cancelled':
+      return OcticonSymbol.stop
+    case 'action_required':
+      return OcticonSymbol.alert
+    case 'skipped':
+      return OcticonSymbol.skip
+    case 'stale':
+      return OcticonSymbol.issueReopened
```
rafeca, line 129: It's very nice to have specific icons and messages for all the potential conclusions, but if I understand the code correctly the combined conclusion currently can only be `success`, `failure` or `null`.

Is the plan to make the logic that calculates the combined state a bit smarter to make it more granular? (I think this can be done in a separate PR)

rafeca (dismissed):
Looks good to me! I've added one nitpick comment and one question, but none of them are blocking so I think this PR is already in shippable state!

app/src/lib/stores/commit-status-store.ts
niik: I absolutely agree but in this case I've chosen to mirror the logic on the dotcom side so I'd prefer to keep the terminology and logic in sync in order to more easily be able to adapt should they decide to support more states in the future.

app/src/ui/branches/ci-status.tsx
niik: Whoopsie, great catch! The intention here was to support all the different conclusions when there's only a single check but I forgot about that along the way. Change incoming momentarily. A medium term goal for us is to allow peeking into a combined status to see exactly what's going on similar to how dotcom does it which is why the checks array is kept in the combined check to begin with.

> Is the plan to make the logic that calculates the combined state a bit smarter to make it more granular? (I think this can be done in a separate PR)

Same thing here as with the incomplete/failure distinction. The logic is mirroring that of how dotcom does check rollups AFAICT. There's room for improvement but if we change it here we should do so on the dotcom side as well.

rafeca: Thanks for the context! Makes sense!

app/src/lib/stores/commit-status-store.ts
rafeca: > I absolutely agree but in this case I've chosen to mirror the logic on the dotcom side so I'd prefer to keep the terminology and logic in sync in order to more easily be able to adapt should they decide to support more states in the future.

I'm not sure I follow, you mean there's also an `isIncomplete()` method on dotcom that only returns `true` when status=='complete'`? If that's the case then I agree about keeping it that way (note that my suggestion was only stylistic and didn't imply changing the logic).

rafeca (dismissed):
Approving this PR, don't want to block it because of a nitpick 😃

app/src/lib/stores/commit-status-store.ts
niik: > I'm not sure I follow, you mean there's also an `isIncomplete()` method on dotcom that only returns `true` when status=='complete'`?

Right, an incomplete check run is a check run which has status 'complete' but doesn't have a definite conclusion (success, failure, neutral) because it's been cancelled, timed out, or marked as stale. So an incomplete check is different from an in progress check or a queued check. I've added a comment to hopefully clarify this in the code.

niik:
Sorry @rafeca, gonna need another review here because of the duplicate check run elimination we discussed in chat.

rafeca (dismissed):
This is so nice! Let's merge it!

rafeca (approved):
Now?? 🤣

niik:
Thank you so much @Raul6469! This will unfortunately not make it into the next production release but we should be able to ship it to our beta channel in a week or so if you'd like to try it out there.

PR #9865, opened 2020-05-25
Perform refreshTags operation inside refreshRepository()

rafeca:
## Description

As suggested by @niik, we prefer to do all refresh-related tasks inside the `refreshRepository()` method so in the future we have the assurance that calling `refreshRepository()` will update all the repo data.

Initially I opted out to call `refreshTags()` independently to be able to choose exactly when we need to refresh the list of tags: there are not many situations where we need to do that, while `refreshRepository()` is called in many more scenarios where we know that the list of tags won't change (e.g when creating a commit, switching branches, performing a merge, etc). Anyways, we tradeoff a really tiny bit of performance when executing these operations to gain predictability when calling `refreshRepository()`.

If in the future we find that the `refreshRepository()` becomes a bottleneck we can always try to make that method smarter by only updating the stuff that needs to be updated based on the action that triggered the call (we don't seem to need that for now though).

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/index.tsx
+++ b/app/src/ui/index.tsx
@@ -305,7 +305,7 @@ document.body.classList.add(`platform-${process.platform}`)
 
 dispatcher.setAppFocusState(remote.getCurrentWindow().isFocused())
 
-ipcRenderer.on('focus', async () => {
+ipcRenderer.on('focus', () => {
   const { selectedState } = appStore.getState()
 
   // Refresh the currently selected repository on focus (if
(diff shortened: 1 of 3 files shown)


app/src/ui/index.tsx
```diff
@@ -314,8 +314,7 @@ ipcRenderer.on('focus', async () => {
     selectedState &&
     !(selectedState.type === SelectionType.CloningRepository)
   ) {
-    await dispatcher.refreshTags(selectedState.repository)
-    await dispatcher.refreshRepository(selectedState.repository)
+    dispatcher.refreshRepository(selectedState.repository)
```
rafeca, line 317: I've left this method exactly as it was before adding the `refreshTags()` call (which means that it doesn't await `refreshRepository()` to finish before calling `setAppFocusState(true)`).

PR #9869, opened 2020-05-26
Show popup for fork settings when cloning a fork

rafeca:
This PR is based on https://github.com/desktop/desktop/pull/9842

Supports #9679

## Description

This PR adds an additional dialog that appears when a user clones a forked GitHub repository that shows the fork contribution settings.

The dialog is shown after the clone process finishes, since up until that time Desktop does not call the GitHub API and doesn't have any information about whether the repository is a fork or not.

We could refactor the cloning logic to do an eager call to the API and then be able to display the dialog earlier (so the user can make the choice while the repository is in the background), but this would require non-trivial work.

Another improvement that we can do as a separate PR would be to show a checkbox (e.g "Choose always this option when cloning a forked repository") to remember the selected option and not show the dialog during forks again (this would only affect the default setting of new forked repos and would not impede users from customizing the fork behaviour of a specific repository from the repo settings).

### Screenshots

demo

## Release notes

Notes: no-notes


niik:
:wave: I haven't reviewed this, I just looked at the GIF and wanted to leave a quick note with a request to make the VerticalSegmentedControl expand the full width of the dialog, i.e 👇 

image

I thought it would do it automatically but maybe the `<Row>` it's in is interfering or something.

rafeca:
> 👋 I haven't reviewed this, I just looked at the GIF and wanted to leave a quick note with a request to make the VerticalSegmentedControl expand the full width of the dialog, i.e 👇

This control is designed to not currently occupy the whole width of the container but the space that its content needs, for example in the theme settings:

image

But I agree with you: it'll look better everywhere if it uses all the width available  (ccing @ampinsk to confirm).

app/src/ui/repository-settings/fork-contribution-target-description.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react'
+import { ForkContributionTarget } from '../../models/workflow-preferences'
+import { RepositoryWithForkedGitHubRepository } from '../../models/repository'
+
+interface IForkSettingsDescription {
+  readonly repository: RepositoryWithForkedGitHubRepository
+  readonly forkContributionTarget: ForkContributionTarget
+}
+
+export function ForkSettingsDescription(props: IForkSettingsDescription) {
```
outofambit, line 10: 💯

tierninho:
@rafeca So this popup appeared and replaced by Preferences modal after I cloned a fork. I think we should not show it in this case? Because of a slight lag on my Windows, I cloned a forked repo and immediately went to make a change in Preferences. Right as I was about to click `Save`, the fork preferences modal popped up and I lost all my changes. Happy to create an issue if needed.

Repro:
- Clone a fork repo on Windows VM (or a slower machine)
- Immediately open Preferences, use the cmd to make it faster
- Wait for the Fork modal to appear

cc @billygriffin for thoughts on which method we ought to go with

billygriffin:
Thanks @tierninho! That certainly sounds not ideal, but I'm also not certain this is very likely to surface organically. I think based on what we've seen, most frequently when people are cloning they're waiting for the clone to finish, or leaving the app and coming back during the cloning process. Not typically engaging with other parts of the app. Just my two cents, I *think* (though I don't feel strongly about this) I'd prefer to solve this only when we see it occur for one or more users.

PR #9870, opened 2020-05-26
Refactor VerticalSegmentedControl to use unique values instead of indices

rafeca:
## Description

This PR implements a small refactor to the `<VerticalSegmentedControl />` to use generic values for items instead of their indices.

The motivation for this change is that while implementing https://github.com/desktop/desktop/pull/9869 I found that the current API for the component is not very ergonomic.

The main reason for that is that in the vast majority of cases, the actual index of the item just serves as a mapping to an actual value, so in order to use this component the parent component needs to do two mappings, one from a value to an index when rendering it and another one from an index to a value when the selected item changes:

```tsx
function MyComponent(props) {
  const selectedIndex =
    props.forkContributionTarget === ForkContributionTarget.Parent
      ? 0
      : 1;

  const items = [
    { title: 'To contribute to the parent project' },
    { title: 'For my own purposes' },
  ];

  return (
    <VerticalSegmentedControl
      items={items}
      selectedIndex={selectedIndex}
      onSelectionChanged={onSelectionChanged}
    />
  )
}

function onSelectionChanged(index: number) {
  const forkContributionTarget =
    index === 0
      ? ForkContributionTarget.Parent
      : ForkContributionTarget.Self

  // (...)
}
```

This not only adds some boilerplate but also adds unnecessary coupling between the mapping logic and the rendering order of the items , which can potentially cause issues when changing the order and forgeting to update some of the mapping logic.

With this PR, the values of the items get encoded on the same items array (in a similar way that a standard `<select/>` HTML component does, which removes the concept of indices from the API:

```tsx
function MyComponent(props) {
  const items = [
    { title: 'To contribute to the parent project', value: ForkContributionTarget.Parent },
    { title: 'For my own purposes', value: ForkContributionTarget.Self },
  ];

  return (
    <VerticalSegmentedControl
      items={items}
      selectedValue={props.forkContributionTarget}
      onSelectionChanged={onSelectionChanged}
    />
  )
}

function onSelectionChanged(value: ForkContributionTarget) {
  // (...)
}
```

Every component that currently renders `<VerticalSegmentedControl />` is doing these mappings, so this change benefits all current usages of the component.

One interesting case is `<StashAndSwitchBranch />`, where the mapping was done implictly by using [an enum](https://github.com/desktop/desktop/blob/development/app/src/ui/stash-changes/stash-and-switch-branch-dialog.tsx#L17-L20) with their entries in the same order as the items on the list, and taking advantage that TypeScript enums by default take autoincrementing values starting from 0 😅 . While this approach reduces the boilerplate of having to do the manual mappings, it's also quite fragile because it depends on the order of the enum and the items being always in sync.

### What if we need value-less items?

Something worth noting is that if we ever need the previous API (maybe to render an indeterminate amount of items), we can always create a higher order component that uses indices by doing something like:

```tsx
function VerticalSegmentedControlWithIndices({
  items,
  selectedIndex,
  onSelectionChanged,
}) {
  const itemsWithValue = items.map((item, index) => ({
    ...item,
    value: index
  })

  return (
    <VerticalSegmentedControl
      items={itemsWithValue}
      selectedValue={selectedIndex}
      onSelectionChanged={onSelectionChanged}
    />
  )
}
```

### Tradeoffs

The most obvious tradeoff is that since the values are used to identify an item (via the `selectedValue` prop), the parent component must pass unique values for the items. As a safenet the component right now filters out any item with a duplicated value (we could add a console warning when this happens).

## Release notes

Notes: no-notes


rafeca:
> Overall I dig it! I'm not 100% sold on the getUniqueItems part though. Did you add that just as a precaution or do you know of a place where we could run into duplicate item values already?

Yup I just added it as a precaution. Looking back to it I agree that we can achieve something similar with a different method.

> If it's former I wonder if we could instead document the fact that values need to be unique and also rename the value property to key to further indicate the need for uniqueness.

I like the idea of renaming it to `key`. In addition to that (and to keep it aligned with React keys), we could limit them to be either `number` or `string`. I've checked and we're currently passing either a number or a string everywhere anyways.

With this, we could even pass they actual item key to React when [rendering each item](https://github.com/desktop/desktop/blob/development/app/src/ui/lib/vertical-segmented-control/vertical-segmented-control.tsx#L118).

This will make the reconciliation logic more performant for React when re-rendering the component with added/removed items (don't think it'll mean much for us though), but the good thing is that it would give us the nice React errors if somebody uses the same key twice:

Screen Shot 2020-06-24 at 16 28 10

niik:
> but the good thing is that it would give us the nice React errors if somebody uses the same key twice:

Oooh, I hadn't thought of that, that's a great benefit!

Issue #9540, opened 2020-04-16
"Open in GitHub Desktop" doesn't work open PR branch when you don't have repo cloned already


billygriffin:
### Describe the bug

When you don't have the repo already cloned in Desktop, clicking "Open this pull request in Desktop" asks you to clone the repo (this seems reasonable), but then doesn't actually check out the PR branch. It instead checks out the default branch.

### Version & OS

Desktop Version 2.4.2-beta1
macOS 10.15.3

### Steps to reproduce the behavior

1. Go to a pull request for a repo you do not already have cloned in Desktop
2. In the merge area, click "Open this pull request in Desktop"
3. Clone the repo into Desktop
4. 👇 

### Expected behavior

I expected to be routed to the branch for the PR that I was on after cloning.

### Actual behavior

I was on the repo's default branch.

PR #9873, opened 2020-05-26
Wait until the repository is cloned in order to checkout a PR

rafeca:
Closes https://github.com/desktop/desktop/issues/9540

## Description

This PR fixes the issues described in https://github.com/desktop/desktop/issues/9540 by making sure that the `_completeOpenInDesktop` promise doesn't get resolved when a different modal gets closed.

### Context

When opening the clone repository modal, Desktop tries to close a few popups ([source](https://github.com/desktop/desktop/blob/713f8e878bdf535bd3186cecae153de9ebbf6e3a/app/src/lib/stores/app-store.ts#L2095-L2096) even if they're not actually opened.

If these calls got executed after the clone repository modal got opened, they caused the `_completeOpenInDesktop` promise to get resolved prematurely (before the user actually cloned the repo) and then the logic that tries to checkout the PR branch after cloning would fail.

The fix consists on moving the condition that checks if the dialog to close is the current opened dialog to be executed before resolving the promise.

## Release notes

Notes: [Fixed] Fix "Open in GitHub Desktop" when the repository has not been cloned yet.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2960,12 +2960,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
-    if (currentPopup.type === PopupType.CloneRepository) {
-      this._completeOpenInDesktop(() => Promise.resolve(null))
-    }
-
     if (popupType !== undefined && currentPopup.type !== popupType) {
       return
+    }
+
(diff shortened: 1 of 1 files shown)


silocoder:
Hi everyone I am using Version 2.5.3. on Mac OS X.
Is this issue related to opening the master branch instead of the branch (e.g. features) when using Desktop Github?

PR #9893, opened 2020-05-29
What's a little space between friends?

niik:
Happy Friday everyone!

Did y'all know there's [no consensus in the English language](https://english.stackexchange.com/a/3284) for whether to include a space in between the number and the percent sign (i.e. `100 %` vs `100%`)?

While I myself like a little bit of social distance every now again that's obviously not everyone's cup of tea. In fact, some people, like our very own @joshaber feels pretty strongly about this.

> image

[According to Wikipedia](https://en.wikipedia.org/wiki/Percent_sign#Spacing) there seems to be more style guides recommending no space than there are recommending a space

> English style guides prescribe writing the percent sign following the number without any space between (e.g. 50%).[1][2][3][4][5][6][7] However, the International System of Units and ISO 31-0 standard prescribe a space between the number and percent sign

Now, I'm an International System of Units kind of a kid but since Git itself will report percentages without spaces and there's some evidence for a convergence towards no space in English I say we let consistency win while at the same time throwing a bone to our old friend, I'm sure he could use one to close the week.

Are there any other strong feelings about spacing?

### DIFF
--- a/app/src/ui/toolbar/branch-dropdown.tsx
+++ b/app/src/ui/toolbar/branch-dropdown.tsx
@@ -142,7 +142,7 @@ export class BranchDropdown extends React.Component<IBranchDropdownProps> {
 
       if (checkoutProgress.value > 0) {
         const friendlyProgress = Math.round(checkoutProgress.value * 100)
-        description = `${description} (${friendlyProgress} %)`
+        description = `${description} (${friendlyProgress}%)`
       }
 
       progressValue = checkoutProgress.value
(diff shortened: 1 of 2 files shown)


joshaber:
giphy-4

nk250319:
No space please. Thanks !

(see how weird that looks?)

Rexogamer:
While the timing feels ironic considering global events, I am certainly a fan of this change!

niik:
Appreciate all the feedback y'all!

Issue #9899, opened 2020-05-30
Cant sign in to multiple accounts on git desktop


JohnnieBlinx:
### Describe the bug

Cant log in to any other account. Browser is opened and auto signs me in with no page loading for me to change the account

### Version & OS

Windows 10
Git Desktop 2.5.0

### Steps to reproduce the behavior

1. Already logged in to git account
2. Sign out
3. Enter details of 2nd account and click sign in

### Expected behavior

Account signs in

### Actual behavior

Chrome opens even though i didnt click sign in with browser, a blank page opens with a giturl with no UI to interact with. Alert pops up (Open git desktop)

Click open git desktop and it signs me back into account #1 not the one ive just entered

codenameone:
This is happening to me too. Retrying (signing in again) produces the desired result. Notice that this only happens if account 2 is not the one signed in on the browser. So now if I go back to account 1 it works the first time around. 

This seems to be related to #9231 

Honestly, I think the best solution is to finally support multiple account signin so we won't have to type the password over and over. If that would work it would be the best way.

johnsiilver:
I'm having the a similar issue.  

If I attempt to sign out and back in with another account, strange things occur (never the desired behavior).

Sometimes after I try and sign in, it opens in Edge beta (default browser) with a URL and then a blank page.  If I try to auth with the browser, does the same thing.  If I restart the app, then I can auth sometimes (hit or miss).

But, if I get  signed in, one of two things will happen (neither a good result):

- While it shows the right account, it doesn't let me access my repos, always an auth failure
- Even though I logged in with one account, it will actually log me back into my other account (and display that name).

What I think is happening is that it really never changes the user context.  No matter what it shows if I auth, I am in my other auth context.

JohnnieBlinx:
I have to open chrome, go to git, sign out and then change accounts in git desktop.

I dont see why youd have the option 'sign in with browser' because when you DONT click that it still opens the browser to sign in

billygriffin:
Sorry for the confusion. I think @codenameone's hypothesis might be correct that this could be related to the work in #9231. @rafeca Do you know if anything might have changed that would affect this?

Also for those in this thread, have you seen this specific issue before? The primary sign in method used to be username/password, but we're shifting to browser-based authentication because username/password auth will be deprecated on the API side. So I'm curious if this problem always existed but wasn't surfaced because everyone was just typing in username/password directly and only surfaced because more people are using the browser auth flow now, or whether this is something entirely new.

[...]

PR #9920, opened 2020-06-02
Put browser authorization button in its own form

rafeca:
Closes https://github.com/desktop/desktop/issues/9899

## Description

This PR fixes https://github.com/desktop/desktop/issues/9899 by creating a separate `<form />` element that isolates the "Sign in via browser" button from the rest of the form.

Since that button (for styling purposes) has a `submit` type, Chromium inferred that this was the main button for the `Dialog` form, which meant that:

1. When clicking on the button, the dialog form was submitted.
2. When pressing enter on the user/password inputs, the button got "pressed".

The first point was worked around by adding a `preventDefault` call to the button click event handler, but the second point was never worked around and has caused  https://github.com/desktop/desktop/issues/9899

### Alternate solution

There's another valid solution to this issue which consists on removing the `type="submit"` attribute from the button. Unfortunately, the current styleguide does not allow to display a button as "primary" (with a blue background) without having it with a `submit` type. We could fix that by creating a custom `buttom-component-primary` which had the same styles as `button[type='submit']`.

I think both the alternate solution and the chosen solution are valid, but I've leaned towards this solution since it doesn't require adding more CSS and it feels slightly more semantically correct to make that button of type `submit` in its own form (since it's a separate action from the main dialog purpose).

## Release notes

Notes: [Fixed] Pressing "enter" on the user/password inputs when authenticating won't open the browser flow.

### DIFF
--- a/app/src/ui/sign-in/sign-in.tsx
+++ b/app/src/ui/sign-in/sign-in.tsx
@@ -212,17 +212,14 @@ export class SignIn extends React.Component<ISignInProps, ISignInState> {
     return (
       <DialogContent>
         <Row className="sign-in-with-browser">
-          <form>
-            <Button
-              className="button-with-icon"
-              type="submit"
-              onClick={this.onSignInWithBrowser}
-              disabled={disableSubmit}
-            >
-              Sign in using your browser
-              <Octicon symbol={OcticonSymbol.linkExternal} />
-            </Button>
-          </form>
+          <Button
+            className="button-with-icon button-component-primary"
+            onClick={this.onSignInWithBrowser}
+            disabled={disableSubmit}
+          >
+            Sign in using your browser
+            <Octicon symbol={OcticonSymbol.linkExternal} />
+          </Button>
         </Row>
 
         <div className="horizontal-rule">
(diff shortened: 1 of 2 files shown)


rafeca:
Thanks for your review! You're right, this is technically not allowed 😅 

In that case I'd rather implement the alternate approach (aka pragmatic one), since a redesign of that flow would take more time to implement and I'm not certain that we're going to be able to reuse much of it later (since when we disable user/pass authentication we can just reuse the whole flow that's used currently for enterprise instances that do not support user/pass authentication).

niik:
> In that case I'd rather implement the alternate approach (aka pragmatic one), since a redesign of that flow would take more time to implement

Perfect!

>[...] that's used currently for enterprise instances that do not support user/pass authentication).

Not that it matters at all for the sake of this discussion but for context some GitHub Enterprise Server instances do support username and password authentication, it's dependent on how they've configured their instance and given that we try to support at least a few old versions of enterprise we'll likely have to keep the username/password flow for a bit longer even after it's unshipped from dotcom.

rafeca:
> In that case I'd rather implement the alternate approach (aka pragmatic one), since a redesign of that flow would take more time to implement

Done in 7a7c310

PR #9921, opened 2020-06-02
Avoid resetting changes state when not showing stash

niik:
References #7710

## Description

This is a follow up to #9412 which fixed the problem described in #7710. This PR addresses a conceptual issue with regards to `_hideStashedChanges`.

After #9412 `_hideStashedChanges`  Prior to this PR `_hideStashedChanges` would always reset the changes state to show the working directory changes (selecting all of the changes in the working directory since it didn't have the context that the `<Changes>` component has about previous selection). This small change makes the method a noop when there's no stash visible.

See https://github.com/desktop/desktop/pull/9412#pullrequestreview-419442826

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2342,6 +2342,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   public _hideStashedChanges(repository: Repository) {
     const { changesState } = this.repositoryStateCache.get(repository)
 
+    // makes this safe to call even when the stash ui is not visible
     if (changesState.selection.kind !== ChangesSelectionKind.Stash) {
       return
     }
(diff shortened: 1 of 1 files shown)


tierninho:
Ran through some stashing scenarios as well as the issue from #7710 and all appears to be functioning as normal. 👍

app/src/lib/stores/app-store.ts
```diff
@@ -2340,6 +2340,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   public _hideStashedChanges(repository: Repository) {
+    const { changesState } = this.repositoryStateCache.get(repository)
+
+    if (changesState.selection.kind !== ChangesSelectionKind.Stash) {
```
outofambit, line 2345: just a little comment please!
```suggestion
    // makes this safe to call even when the stash ui is not visible
    if (changesState.selection.kind !== ChangesSelectionKind.Stash) {
```

niik: Great shout, thanks!

PR #9929, opened 2020-06-03
Diffs, sans flicker

niik:
## Description

When switching between changed files in the changes tab (or as in the case of the gif below changed files in a commit) there's a noticeable flicker in the diff view for each file change that's bugged me ever since the first prototype of diff support in Desktop.

This is because when switching the selected file we [clear the diff text](https://github.com/desktop/desktop/blob/067f01f844e8cb35cf1de1adfcabb441ce7ca814/app/src/lib/stores/app-store.ts#L929) and the load that asynchronously. So in between each file change we need to unmount the TextDiff component, mount the `No File Selected ` component (which is weird in and of itself), and when the diff loading is complete we remount the TextDiff, reinitialize CodeMirror, restart the highlighters and render the diff.

image

I've had hypothesis for a long time that since we're talking about so little time in between the switch happening and the diff being loaded we should be able to cheat a little and keep the last displayed diff around for a 100-150 ms or so and just update CodeMirror with a new document rather than unmounting and remounting the component.

Here's the same GIF from above but manually edited to remove the "No File Selected" frames, looks much better now, doesn't it?

(null) 2020-05-27 19_52_00

This is obviously a fake and a little bit unrealistic since it doesn't account for the time it actually takes to get the new diff but to the user's eye but the reality isn't too far off from that.

Should the Diff takes more than a 100 ms to load we gradually (over another 100ms) fade out the diff and fade in a loading spinner.

image

## Approach

I settled on creating a new component (`SeamlessDiffSwitcher`) which takes care of keeping the old diff around and swapping it out for the new when that's available. That way we keep the logic in the Diff component focused only on rendering a Diff instead of complicating it even further.

I found one bug in our use of CodeMirror that I could reproduce without the diff switcher but which became much more prevalent now that we're keeping CodeMirror instances around for longer. The good news is I've seen this bug in the wild (you focus the app and all of a sudden the diff gutter is gone) and now that I could reliably replicate it I was able to narrow it down and solve it.

The Diff component previously took a `Dispatcher` instance in its props which made it really hard to reason about what actions it could end up doing. I catalogued the places where it was used substituted it for specific callbacks. This allows the `SeamlessDiffSwitcher` to swap out any actions the diff could theoretically do for no-op functions while the diff is transitioning.

## Release notes

Notes: [Improved] A smoother transition between diffs when switching between changed files


tierninho:
Very nice, much softer on the eyes! 😎

outofambit:
tried this out locally because i wanted to see how it dealt with quick, successive paging down a list of diffs. it worked great! didn't do a full code review, so leaving this as a comment!
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FXbxZ41fWLeRECPsGIJ%2Fgiphy.gif&token=me_likey" data-gitmeme-token="me_likey"></a>

niik:
Thanks y'all! I'm excited about this, please keep an eye out for any glitches in the diffs just in case I missed something but I'm feeling pretty confident given that we've supported in-place reloading of TextDiff's in other scenarios (refreshing after app focus, after partial commits, when toggling display of white space, etc).

PR #2575, opened 2017-08-29
Commit summary truncation

niik:
This fixes #1742 by truncating and wrapping commit summaries over 72 characters in the same manner that dotcom currently does it. We're able to render 72 (non-fancy-unicodey) characters in the minimum window width and even if we didn't it'd just wrap like it currently does.

This follows what @donokuda and I talked about and is a fairly simple fix we think will work alright for now. Note that this might cause links to break if we truncate them in the middle. The solution for this would be to do a pre-pass with the tokenizer and truncate using tokens instead but that's a bigger task that I'm not convinced that we need right now.

### Before
image

### After
image

PR #9931, opened 2020-06-03
Token-aware text wrapping of commit summaries

### LINKED ISSUE #9185: urls within commits get cut off and cannot get opened by the browser

niik:
Closes #9185 

## Description

As described in https://github.com/desktop/desktop/issues/9185 our naive text wrapping efforts implemented in https://github.com/desktop/desktop/pull/2575 suffer from a major problem when commit summaries contains links that gets wrapped.

image

Since the wrapper operates on the raw text of the commit summary before the `Tokenizer` class (used by the `RichText` component) has had a chance to work it happily hard wraps in the middle of a url. When the `RichText` component then gets a chance to work on it it creates a link out of the link that remains.

This was a limitation that was called out in #2575

> [...] a fairly simple fix we think will work alright for now. Note that this might cause links to break if we truncate them in the middle. The solution for this would be to do a pre-pass with the tokenizer and truncate using tokens instead but that's a bigger task that I'm not convinced that we need right now.

In reviewing #9259 it became clear that that assumption no longer held true and we needed to look into a robust way of doing wrapping that was Token aware.

In this PR I've made it possible to provide the `RichText` component with a list of tokens as well as a string for the component to tokenize. I've also updated the `CommitSummary` to tokenize both the commit summary and the body before providing them to the `RichText` component.

Finally I've implemented a token-aware wrapper function that hopefully covers most of the many edge-cases involved in doing this type of wrapping. I've set up https://github.com/niik/commit-summary-wrap-tests/commits/master as a list of commits with weird behaviors that should be accounted for and intend to keep adding commits there as I encounter them.

### Screenshots

Here's a few before and afters

**Before**
image

**After**

image

Here the wrapper takes into account the fact that the link has been resolved as an issue link and thereby shortened to just two characters.

**Before**

image

**After**

image

**Before**

image

This  illustrates the main problem described in #9185.

**After**

image

While the link is hard-wrapped both parts are clickable and go to the same URL.

**Before**

image

Not aware of emojis, splits after 72 characters (counting ':one:` as five characters)

**After**

image

Emojis are counted as two regular characters so we're  able to fit a lot of them in  the commit summary.

## Notes

This does not solve (but it doesn't make the situation any worse either) the problem of multibyte unicode characters (native emoji). They can still get wrapped in the middle. I don't intend to address that as part of this PR.

## Release notes

Notes: [Fixed] Links are no longer broken when text wrapping needs to take place in commit summaries

### DIFF
--- a/app/src/lib/text-token-parser.ts
+++ b/app/src/lib/text-token-parser.ts
@@ -263,9 +263,7 @@ export class Tokenizer {
     }
   }
 
-  private tokenizeNonGitHubRepository(
-    text: string
-  ): ReadonlyArray<TokenResult> {
+  private tokenizeNonGitHubRepository(text: string): TokenResult[] {
     let i = 0
     while (i < text.length) {
       const element = text[i]
@@ -294,7 +292,7 @@ export class Tokenizer {
   private tokenizeGitHubRepository(
     text: string,
     repository: GitHubRepository
-  ): ReadonlyArray<TokenResult> {
+  ): TokenResult[] {
     let i = 0
     while (i < text.length) {
       const element = text[i]
@@ -338,7 +336,7 @@ export class Tokenizer {
    *
    * @returns an array of tokens representing the scan results.
    */
-  public tokenize(text: string): ReadonlyArray<TokenResult> {
+  public tokenize(text: string): TokenResult[] {
     this.reset()
 
     if (this.repository) {
--- a/app/src/lib/text-token-parser.ts
+++ b/app/src/lib/text-token-parser.ts
@@ -263,7 +263,9 @@ export class Tokenizer {
     }
   }
 
-  private tokenizeNonGitHubRepository(text: string): TokenResult[] {
+  private tokenizeNonGitHubRepository(
+    text: string
+  ): ReadonlyArray<TokenResult> {
     let i = 0
     while (i < text.length) {
       const element = text[i]
@@ -292,7 +294,7 @@ export class Tokenizer {
   private tokenizeGitHubRepository(
     text: string,
     repository: GitHubRepository
-  ): TokenResult[] {
+  ): ReadonlyArray<TokenResult> {
     let i = 0
     while (i < text.length) {
       const element = text[i]
@@ -336,7 +338,7 @@ export class Tokenizer {
    *
    * @returns an array of tokens representing the scan results.
    */
-  public tokenize(text: string): TokenResult[] {
+  public tokenize(text: string): ReadonlyArray<TokenResult> {
     this.reset()
 
     if (this.repository) {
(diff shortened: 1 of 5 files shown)


app/src/lib/text-token-parser.ts
```diff
@@ -338,7 +336,7 @@ export class Tokenizer {
    *
    * @returns an array of tokens representing the scan results.
    */
-  public tokenize(text: string): ReadonlyArray<TokenResult> {
+  public tokenize(text: string): TokenResult[] {
```
rafeca, line 339: Since this is a public method and it's used from React components I'm not super happy about making the output array mutable (with these current changes it's still safe but I'm worried that at some point somebody will pass this mutable around as a Prop to a React component).

Can we keep this as a readonly array? (see my comment below)

app/src/lib/wrap-rich-text-commit-message.ts
```diff
@@ -0,0 +1,117 @@
...
+      body.unshift({ kind: TokenType.Text, text: `…` }, ...overflow, {
+        kind: TokenType.Text,
+        text: '\n\n',
+      })
+    } else {
+      body.unshift({ kind: TokenType.Text, text: `…` }, ...overflow)
+    }
+  }
+
+  return { summary, body }
```
rafeca, line 116: Maybe we can avoid mutating the output of the tokenizer by using early returns:

```suggestion
  if (overflow.length === 0) {
    return { summary, body }
  }

  summary.push({ kind: TokenType.Text, text: '…' })
  overflow.unshift({ kind: TokenType.Text, text: `…` })

  if (body.length === 0) {
    return { summary, body: overflow }
  }

  return {
    summary,
    body: [...overflow, { kind: TokenType.Text, text: `…` }, ...body],
  }
```

(another option would be to just clone the output array and keep the current logic).

app/src/lib/text-token-parser.ts
niik: > at some point somebody will pass this mutable around as a Prop to a React component

You sure know how to nerd-snipe me. See https://github.com/desktop/desktop/pull/9939

> Can we keep this as a readonly array? (see my comment below)

Yep, I took a stab at reworking it, I took a slightly different approach to yours but I think we achieve the same end-result.

app/src/lib/wrap-rich-text-commit-message.ts
niik: Took a stab at this, feel free to have another look

niik:
Thanks y'all!

PR #9944, opened 2020-06-05
Move retry clone dialog into AppError

niik:
## Description

@rafeca ran into an edge case where a clone failed due to some refspec tomfoolery he was up to and noticed that the dialog presenting the error extended way beyond the window dimensions.

image

That's certainly not great UX. Upon seeing the image I was surprised that the git error output wasn't monospaced as I'd done some work to ensure that it was in https://github.com/desktop/desktop/pull/8964. Turns out that's because it's a custom dialog invoked through an error handler and as such doesn't enjoy any of the improvements we've made to app errors recently.

It also meant that the error text wasn't selectable, and that the error message didn't overflow using a capped max-height like in the AppError case.

What I've done now instead is to remove the custom error handler and move the detection logic for this particular error into the `AppError` dialog similar to how we've had custom support for auth errors there.

**Note:** this does complicate the logic of the `AppError` component in a fairly nasty way since it means we'll have to check for this error in three places (when determining the title, when rendering the contents, and when rendering the footer). This is a limitation that me and @rafeca discussed and that we think we can get around with some refactoring of the `AppError` component but not something I want to tackle in this PR.

### Before
Screen Shot 2020-06-05 at 13 08 20

### After

Screen Shot 2020-06-05 at 13 06 29

## Release notes

Notes: [Improved] Clearer presentation of error message when a clone fails


app/src/ui/app-error.tsx
```diff
@@ -198,9 +286,17 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
...
 
-    if (e instanceof GitError) {
+      if (metaDataFooter) {
+        return metaDataFooter
+      }
+    }
+
+    const e = getUnderlyingError(error)
+
+    if (isGitError(e)) {
```
rafeca, line 299: It took me a while to understand this logic, it seems like L297-L303 is doing the same as L105-L109 when the error is `ErrorWithMetadata`, but in that case if I understand the logic correctly `metaDataFooter` will not be `null` (it can never be `null`) and this code won't get executed.

If I understood the whole thing correctly, I think we can simplify this whole thing by removing the `renderErrorWithMetaDataFooter()` method and changing this method to something like:

```jsx
  private renderFooter(error: Error) {
    if (isCloneError(error)) {
      return this.renderRetryCloneFooter() // The renderRetryCloneFooter() doesn't need any param
    }

    const e = getUnderlyingError(error)

    if (isGitError(e)) {
      return this.renderGitErrorFooter(e)
    }

    return this.renderDefaultFooter()
  }
```

app/src/ui/app-error.tsx
```diff
@@ -91,6 +94,51 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
...
+    }
+
+    if (isGitError(error.underlyingError)) {
+      return this.renderGitErrorFooter(error.underlyingError)
+    }
+
+    return this.renderDefaultFooter()
+  }
+
+  private renderRetryCloneFooter(retryAction: RetryAction) {
```
rafeca, line 112: This function is not using the `retryAction` param so we can get rid of it

app/src/ui/app-error.tsx
```diff
@@ -91,6 +94,51 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
...
+    let retryTitle = 'Retry'
+
+    if (this.isCloneError) {
```
rafeca, line 115: Can we get rid of this check and assume that this function is only called when there is a clone error? (its name is  `renderRetryCloneFooter()` so it should be a safe assumption).

app/src/ui/clone-repository/retry-clone-dialog.tsx
```diff
@@ -1,76 +0,0 @@
-import * as React from 'react'
```
rafeca, line 1: So good to delete this file!!! 🌶️

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -188,49 +188,6 @@ export async function gitAuthenticationErrorHandler(
   return null
 }
 
-/**
- * Handle git clone errors to give chance to retry error.
- * Doesn't handle auth errors.
- */
-export async function gitCloneErrorHandler(
```
rafeca, line 195: yay!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FwkKRo7N0T1ONO%2Fgiphy.gif&token=delete" data-gitmeme-token="delete"></a>

app/src/ui/app-error.tsx
```diff
@@ -132,6 +184,28 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
...
+      if (
+        e.metadata.retryAction !== undefined &&
+        e.metadata.retryAction.type === RetryActionType.Clone
+      ) {
+        return true
+      }
+    }
+
+    return false
+  }
```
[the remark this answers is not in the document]
niik: Yeah I don't think so, I started out this way hoping that I would be able to share it with other methods but that didn't happen (due to the need for type narrowing)

app/src/ui/app-error.tsx
niik: Good catch, this method was initially called `renderRetryFooter` and I had intended it to be the entry point for all retry actions but then I renamed it at the last minute and didn't update the assumptions within.

app/src/ui/app-error.tsx
niik: Yeah, this was another misguided attempt at a generalization of the AppError component to handle even more errors in the future. I've backed away from that now  given that thus far there's only two special kinds of errors and when we add more we'll revisit proper refactoring

app/src/ui/app-error.tsx
```diff
@@ -132,6 +125,14 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
     return <p className={className}>{e.message}</p>
   }
 
+  private getTitle(error: Error) {
+    if (isErrorWithMetaData(error) && isCloneError(error)) {
```
rafeca, line 129: Super-small nit: I don't think we need to check whether the error has metadata here (since `isCloneError()` already does:

```suggestion
    if (isCloneError(error)) {
```

niik: Good catch!

PR #9945, opened 2020-06-05
Remove redundant progress text from Git error messages

niik:
~~:cactus: This depends on #9944, please review that first 🌵~~

## Description

When looking at #9944 I was struck once again by how bad our error messages from Git commands that show some sort of progress are:

image

While somewhat improved in https://github.com/desktop/desktop/pull/8964...

image

There's still a whole lot of redundant text in there that users don't care about and which often times ends up making it really hard to find the actual error message.

The reason for all of this is that Git (and many other CLI tools) leverage an old trick from the typewriter era called the carriage return character to make it seem like they're just updating a single character in the users' terminal.

giphy

When the users sees something like `Progress: 1%` that then changes to `Progress: 2%` what Git actually does is it writes `Progress: 1%` followed by a carriage return character (`\r`) which sends the "carriage" back to the leftmost position and then it overwrites the existing text with `Progress: 2%`. Since we didn't handle carriage returns we got, and displayed each of those updates which for clones and fetches of large repositories could be enormous.

This PR changes that by introducing a parser that emulates how a terminal would interpret carriage returns so now we only show the last progress line for each type of progress reported.

#### Before

Screen Shot 2020-06-05 at 13 06 29

#### After

image

## Release notes

Notes: [Improved] Removed redundant progress text in Git error messages

### DIFF
--- a/app/src/lib/parse-carriage-return.ts
+++ b/app/src/lib/parse-carriage-return.ts
@@ -25,7 +25,7 @@ export function parseCarriageReturn(text: string) {
   }
 
   const lines = new Array<string>('')
-  const crLfOrEnd = /(.*?)([\r\n$])/gm
+  const crLfOrEnd = /(.*?)([\r\n]|$)/gm
 
   let col = 0
   let match
--- a/app/src/lib/parse-carriage-return.ts
+++ b/app/src/lib/parse-carriage-return.ts
@@ -34,7 +34,6 @@ export function parseCarriageReturn(text: string) {
   const crLfOrEnd = /([\s\S]*?)([\r\n]|$)/g
   const lines = new Array<string>('')
 
-  let col = 0
   let match
 
   while ((match = crLfOrEnd.exec(text)) !== null) {
@@ -48,15 +47,11 @@ export function parseCarriageReturn(text: string) {
 
     if (match[1].length > 0) {
       const line = lines[lines.length - 1]
-      const before = line.substring(0, col)
-      const after = line.substring(col + match[1].length)
-      col += match[1].length
-      lines[lines.length - 1] = `${before}${match[1]}${after}`
+      const remainder = line.substring(match[1].length)
+      lines[lines.length - 1] = `${match[1]}${remainder}`
     }
 
-    if (match[2] === '\r') {
-      col = 0
-    } else if (match[2] === '\n') {
+    if (match[2] === '\n') {
       lines.push('')
     }
   }
--- a/app/src/lib/parse-carriage-return.ts
+++ b/app/src/lib/parse-carriage-return.ts
@@ -24,12 +24,12 @@ export function parseCarriageReturn(text: string) {
     return text
   }
 
-  // In JavaScript `.` will never match a newline character since
-  // there's no Regex modifier to allow that which is why the
-  // first group doesn't have to be lazy. The second group matches
-  // the next following \r or \n character or the end of the string.
-  // Matching the end of the string lets us avoid dealing with
-  // leftover characters outside of the matching loop.
+  // `.` will never match a newline character unless we set
+  // the dotAll flag (/s) which is why the first group doesn't
+  // have to be lazy. The second group matches the next following
+  // \r or \n character or the end of the string. Matching the
+  // end of the string lets us avoid dealing with leftover
+  // characters outside of the matching loop.
   const crLfOrEnd = /(.*)([\r\n]|$)/gm
   const lines = new Array<string>('')
 
--- a/app/src/lib/parse-carriage-return.ts
+++ b/app/src/lib/parse-carriage-return.ts
@@ -31,6 +31,14 @@ export function parseCarriageReturn(text: string) {
   let match
 
   while ((match = crLfOrEnd.exec(text)) !== null) {
+    // If we match the $ (end of string) we'll get a zero
+    // with match, this is a known problem in JS so we'll
+    // need to bump the regexp cursor to ensure it fails to
+    // match on the next round
+    if (match.index === crLfOrEnd.lastIndex) {
+      crLfOrEnd.lastIndex++
+    }
+
     if (match[1].length > 0) {
       const line = lines[lines.length - 1]
       const before = line.substring(0, col)
--- a/app/src/lib/parse-carriage-return.ts
+++ b/app/src/lib/parse-carriage-return.ts
@@ -1,3 +1,22 @@
+/**
+ * Parses carriage returns the same way a terminal would, i.e by
+ * moving the cursor and (potentially) overwriting text.
+ *
+ * Git (and many other CLI tools) use this trick to present the
+ * user with nice looking progress. When writing something like...
+ *
+ * 'Downloading: 1%  \r'
+ * 'Downloading: 2%  \r'
+ *
+ * ...to the terminal he user is gonna perceieve it as if the 1 just
+ * magically changes to a two.
+ *
+ * The carriage return character for all of you kids out there
+ * that haven't yet played with a manual typewriter refers to the
+ * "carriage" which held the character arms, see
+ *
+ *  https://en.wikipedia.org/wiki/Carriage_return#Typewriters
+ */
 export function parseCarriageReturn(text: string) {
   // Happy path, there are no carriage returns in
   // the text, making this method a noop.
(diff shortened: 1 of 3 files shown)


app/src/lib/parse-carriage-return.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Parses carriage returns the same way a terminal would, i.e by
+ * moving the cursor and (potentially) overwriting text.
+ *
+ * Git (and many other CLI tools) use this trick to present the
+ * user with nice looking progress. When writing something like...
+ *
+ * 'Downloading: 1%  \r'
+ * 'Downloading: 2%  \r'
+ *
+ * ...to the terminal he user is gonna perceieve it as if the 1 just
```
rafeca, line 11:
```suggestion
 * ...to the terminal the user is gonna perceive it as if the 1 just
```

app/src/lib/parse-carriage-return.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export function parseCarriageReturn(text: string) {
+  // Happy path, there are no carriage returns in
+  // the text, making this method a noop.
+  if (text.indexOf('\r') < 0) {
+    return text
+  }
+
+  const lines = new Array<string>('')
+  const crLfOrEnd = /(.*?)([\r\n$])/gm
```
rafeca, line 28: Do we need the question mark on the RegExp?

```suggestion
  const crLfOrEnd = /(.*)([\r\n$])/gm
```

app/src/ui/app-error.tsx
```diff
@@ -122,7 +125,7 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
 
     const className = monospace ? 'monospace' : undefined
 
-    return <p className={className}>{e.message}</p>
+    return <p className={className}>{this.formatGitErrorMessage(e.message)}</p>
```
rafeca, line 128: question: do we want to call the format function for any kind of error or only for error messages that come from the output of `git` (since we only use monospaced fonts in that scenario)?

app/src/lib/parse-carriage-return.ts
niik: Technically no since `.` doesn't match newline characters in javascript, but in other languages we would have had to in order to make the expression lazy to ensure it matches as few times as possible. If we made it greedy the first group could end up eating up one or `\r` or `\n` if they are adjacent to other newline characters.

I think being explicit about the need for the first expression to be lazy here is a good thing though so I'd like to keep it that way.

app/src/ui/app-error.tsx
niik: Ah, yeah good call. It shouldn't matter since the only added cost would be one `.indexOf` call but I agree with you that we should keep these two concepts aligned.

niik: Fixed in https://github.com/desktop/desktop/pull/9945/commits/ce210cacf8d86e478fff32f1e2ef46236580ed3a

app/src/lib/parse-carriage-return.ts
rafeca: Thanks for the context!! Super useful!

> but in other languages we would have had to in order to make the expression lazy

I thought that in all RegExp implementations newlines were not matched by `.` 😅  (too much time doing JS 🙃).

I agree about keeping it but I would add a comment about why it's there (otherwise I can see my future self removing it LOL).

BTW I just noticed the `$` sign in the last group, I think that it should be moved to the end of the expression since afaik right now it will match the dollar character.

```suggestion
  const crLfOrEnd = /(.*?)([\r\n])$/gm
```

niik: > I thought that in all RegExp implementations newlines were not matched by . 😅 (too much time doing JS 🙃).

And I learned that my knowledge was outdated here as well, turns out [`.` matching newlines](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) has been supported in most browser for [a good while now](https://caniuse.com/#search=dotall).

> BTW I just noticed the $ sign in the last group, I think that it should be moved to the end of the expression since afaik right now it will match the dollar character.

Wow, I don't know how I missed that, and even more depressing is that I hadn't written a test to catch it. I've added both a test and changed the logic now.

> otherwise I can see my future self removing it LOL

Soooo... as I was researching the dotAll stuff I happened to see that JavaScript `.` matches on unicode line and paragraph breaks as well 😱. I feel like I knew this at the back of my head but have never really had to care about it. Now we have to though because hypothetically the remote could send either of those characters so I've had added a test specifically for that and updated the RegExp to use `[\s\S]` as a more explicit substitute for `.` i.e. we want it to match _EVERYTHING_ (but lazily)

rafeca: Wow! Really good catch and thanks for adding the super detailed comment!

image

niik:
@rafeca I realized when looking at the code with a fresh set of eyes that the `${before}` part of our line merging wasn't necessary, it was alway an empty string which makes sense since the `\r` character always moves the cursor to the first column. That let me get rid of the `col` var and one unnecessary `substring` call.

I also spotted one more opportunity for optimization to avoid the last `substring` in the majority use case and I took the time to add a few more tests to validate these assumptions and to test both the happy path and the non-happy-path. Sorry for the extra review request!

rafeca:
Thanks for doing these optimizations! From your explanations and by looking at the code they make sense.

I'm getting a little bit worried that using a RegExp makes the logic of this method a bit hard to understand due to a couple of smart tricks (overriding the `lastIndex` property and the RegExp definition with `[\s\S]` and the lazy matches). The comments that you added are really great and help a lot, but it still took me a bit to understand what's going on.

Have you considered just splitting the string manually to avoid using a RegExp? It will be less memory efficient but I don't think it would be a big deal in this scenario since we already have the whole string in memory... I'm thinking about something similar to:

```tsx
  const lines = text.split('\n')

  const cleanLines = lines
    .map(line => {
      const sublines = line.split('\r')
      let out = ''

      for (const subline of sublines) {
        if (subline.length >= out.length) {
          // Happy path, the new line is equal to or longer
          // than the previous, we can just use the new one
          // without creating any new strings.
          out = subline
        } else {
          const remainder = out.substring(subline.length)
          out = `${subline}${remainder}`
        }
      }

      return out
    })

    return cleanLines.join('\n')
```

app/src/lib/parse-carriage-return.ts
```diff
@@ -0,0 +1,67 @@
...
+  // Happy path, there are no carriage returns in
+  // the text, making this method a noop.
+  if (text.indexOf('\r') < 0) {
+    return text
+  }
+
+  // JavaScript `.` includes unicode line and paragraph break
+  // characters which is why we use [\s\S] instead of '.'. The
+  // first group matches lazily (as few times as possible) and
+  // tthe second group matches the next following \r or \n
```
rafeca, line 30:
```suggestion
  // the second group matches the next following \r or \n
```

app/src/lib/parse-carriage-return.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  let match
+
+  while ((match = crLfOrEnd.exec(text)) !== null) {
+    // If we match the $ (end of string) we'll get a zero width
+    // match, so we'll need to bump the regexp cursor to ensure
+    // it fails to match on the next round
+    if (match.index === crLfOrEnd.lastIndex) {
+      crLfOrEnd.lastIndex++
+    }
```
rafeca, line 45: I'm not 100% sure why this is needed... 🤔  is it because without increasing the index do we get into an infinite loop because we get empty matches?

niik: Yup, exactly. `$` matches end of string which is a zero length match. See http://www.regexguru.com/2008/04/watch-out-for-zero-length-matches/

niik:
> Have you considered just splitting the string manually to avoid using a RegExp? It will be less memory efficient but I don't think it would be a big deal in this scenario since we already have the whole string in memory... I'm thinking about something similar to:

Yes, I played around with something similar to that early on but I wasn't feeling great about all the intermediate arrays but I agree that this solution has incrementally evolved into something much more complex than I had originally envisioned. 

While there are cases when the clone output become ridiculously long we do have a cap at 10Mb in dugite for `stderr` so even if the split/map approach would consume more peak memory (not saying it does, I haven't measured and substrings are usually cheap in v8) it should be at a cost we'd be willing to pay for less complexity.

I took your proposed change and tweaked it just a little bit since I think a reduce operation fits in really nicely here.

app/test/unit/parse-carriage-return-test.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(parseCarriageReturn('foo\r')).toBe('foo')
+    expect(parseCarriageReturn('foo\nbar\r')).toBe('foo\nbar')
+  })
+
+  it('handles strings terminating without newline or CR', () => {
+    expect(parseCarriageReturn('foo\rbar')).toBe('bar')
+    expect(parseCarriageReturn('foo\r\r\r\r\rbar')).toBe('bar')
+  })
+
+  it('treats unicode line separator and paragraph separator as flow text', () => {
```
niik, line 93: @rafeca This test is kinda pointless now that we're not using a regex but I think it's a fair test still in case we alter course in the future and turn it into a regex again. I'll drop it if you want though

PR #9984, opened 2020-06-11
Promise helper cleanup

niik:
## Description

Happened to open this file today while looking the generic `delay()` promise helper that I knew we had lying around here somewhere and was caught trying to understand the `promiseWithMinimumTimeout` function.

Am I missing something or is it needlessly complicated for what it's trying to do?

### DIFF
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -46,9 +46,5 @@ export function promiseWithMinimumTimeout<T>(
  * @param timeout the time to wait before resolving the promise (in milliseconds)
  */
 export async function timeout(timeout: number): Promise<void> {
-  return new Promise((resolve, reject) => {
-    window.setTimeout(() => {
-      resolve()
-    }, timeout)
-  })
+  return new Promise(resolve => window.setTimeout(resolve, timeout))
 }
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -7,35 +7,13 @@
  *
  *
  * @param action the promise work to track
- * @param timeout the minimum time to wait before resolving the promise (in milliseconds)
+ * @param timeoutMs the minimum time to wait before resolving the promise (in milliseconds)
  */
 export function promiseWithMinimumTimeout<T>(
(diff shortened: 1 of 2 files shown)


niik:
I should point out that we [do have tests for this function](https://github.com/desktop/desktop/blob/0c3fe5fbf53c23fe83baa2d9acbbd5ec095f0b95/app/test/unit/promise-with-timeout-test.ts).

app/src/lib/promise.ts
```diff
@@ -7,35 +7,13 @@
...
   action: () => Promise<T>,
-  timeout: number
+  timeoutMs: number
 ): Promise<T> {
-  return new Promise<T>((resolve, reject) => {
-    let timeoutExpired = false
-    let result: T | null = null
-
-    const resolveIfBothDone = () => {
-      if (result != null && timeoutExpired) {
```
niik, line 21: Looking at this diff now I'm pretty sure this is a bug and if the `action` returns `null` or `undefined` then `promiseWithMinimumTimeout` will never resolve. Gonna add a test to confirm

app/src/lib/promise.ts
```diff
@@ -7,35 +7,13 @@
  *
  *
  * @param action the promise work to track
- * @param timeout the minimum time to wait before resolving the promise (in milliseconds)
+ * @param timeoutMs the minimum time to wait before resolving the promise (in milliseconds)
  */
 export function promiseWithMinimumTimeout<T>(
   action: () => Promise<T>,
-  timeout: number
+  timeoutMs: number
```
niik, line 14: This rename is to avoid the name clash with the `timeout` function below

app/src/lib/promise.ts
rafeca: You're right!  🙈

app/src/lib/promise.ts
```diff
@@ -7,35 +7,13 @@
  *
```
rafeca, line 7: It's not related to this PR, but the comment in L6 seems to be incomplete, maybe we can fix it while we're modifying this file.

Issue #9698, opened 2020-05-04
Show in Finder/Explorer no longer works


Clancey:
### Describe the bug

Show in Finder/Explorer used to take you inside the cloned folder where the .git folder is located.  Now it takes you to the parent projects. i.e, `~/Projects/MyRepo`, now it takes you to `~/Projects/`.  When you have a Projects folder with hundreds of projects this is less than helpful.

### Version & OS
Version 2.4.3
Windows and Mac

### Steps to reproduce the behavior

1. Click on Repository Menu Item
2. Click on Show in Finder/Explorer
3. Folder now opens to parent directory instead of the repository directory 

### Expected behavior

I would see a window to `~/Projects/MyRepo`

### Actual behavior

I am taken to `~/Projects/`

steveward:
Thanks for the report @Clancey. I was able to reproduce this issue as well, but only on macOS. Can you confirm you were able to reproduce this on Windows?

Clancey:
I can confirm it is mac only. I thought I saw it on windows.

PR #2968, opened 2017-10-05
just open the log directory

shiftkey:
`showItemInFolder` is designed for revealing a file in the file manager, by opening the parent directory and focusing on it. This isn't great for our case with opening the logs directory - we should open that directory instead of making this flow confusing.

As `openExternal` expects a URL, this also uses the `Url.format` helper function to generate a safe representation of the folder.

PR #2861, opened 2017-09-24
show contents of directory if asked, not parent folder

shiftkey:
Fixes #2798 

I'd rather make these IPC events specific to the Desktop app, so that these events:

```ts
  ipcMain.on(
    'open-external',
    (event: Electron.IpcMessageEvent, { path }: { path: string }) => {
        // code
    }
  )

  ipcMain.on(
    'show-item-in-folder',
    (event: Electron.IpcMessageEvent, { path }: { path: string }) => {
      // code      
    }
  )
```

Looks more like this in the main process:

```ts
  ipcMain.on(
    'open-url',
    (event: Electron.IpcMessageEvent, { path }: { path: string }) => {
        // code
    }
  )

  ipcMain.on(
    'open-file',
    (event: Electron.IpcMessageEvent, { path }: { path: string }) => {
      // code      
    }
  )

  ipcMain.on(
    'open-folder',
    (event: Electron.IpcMessageEvent, { path }: { path: string }) => {
      // code      
    }
  )
```

And we move and complexity into the renderer, where we know more about the context for making the call. But I'll spin that off to a new `tech-debt` task if we're :thumbsup: on going down that path.

PR #9987, opened 2020-06-11
Fix logic to open item in Finder/explorer

rafeca:
Closes https://github.com/desktop/desktop/issues/9698

## Description

This PR fixes the behaviour of opening a repository in Finder (in macOS) to open the actual repository directory instead of its parent folder.

The bug was introduced in https://github.com/desktop/desktop/commit/edde78b3e05ccb2fb0ee2238bd3656e02367285f (I cannot find its associated PR, so maybe it was committed directly to `master`).

`openDirectorySafe()` was created in https://github.com/desktop/desktop/pull/2968 as a refactor of https://github.com/desktop/desktop/pull/2861, and from both the description of the PRs and my understanding of its logic, that function should be always called (both for macOS and Windows, since the function already has logic to handle both platforms).

More specifically, based on [this comment](https://github.com/electron/electron/issues/10095#issuecomment-317285391) we should always use `openItem()` to open a directory, we just happen to have some specific code in `openDirectorySafe()` to open the folder and focus Finder.

## Release notes

Notes: [Fixed] Open the repository folder when clicking on "Show in Finder" in macOS


say25:
With this in mind, I think this will solve the issue I was running into on #9499. Ill try to get that up-to-date in the next few days.

niik:
@outofambit Since I wrote a lot of this way back when it would be wonderful if you could do a review of this before the end of the month. Let me know if you don't think you'll have the time and me and @rafeca can do a paired review

app/src/lib/app-shell.ts
```diff
@@ -7,8 +7,31 @@ export interface IAppShell {
   readonly moveItemToTrash: (path: string) => boolean
   readonly beep: () => void
   readonly openExternal: (path: string) => Promise<boolean>
+  /**
+   * Reveals the specified file using the operating
+   * system default application.
+   * Do not use this method with non-validated paths.
+   *
+   * @param path - The path of the file to open
```
alayoon, line 15: 15

niik:
@rafeca I started reviewing this and found some things that weren't ideal and started down another path of being even more cautious. I figure we could pair review this tomorrow if you have time.

niik:
@tierninho Would love for you to take another spin with this now.

Issue #9988, opened 2020-06-11
Default checkbox of Discard line confirmation dialogue 


tierninho:
When discarding a line item, the "Don't show again" checkbox is set at true by default and keeps reverting to the checked state even if set otherwise in the Preferences. 

It should remain unchecked unless toggled by the user in the modal or in Preferences. This would match the functionality of other similar features.

![Screen Shot 2020-06-11 at 9 36 29 AM](https://user-images.githubusercontent.com/14828183/84432019-ed855000-abc7-11ea-8eb3-4d5601b97f2a.png)

v2.5.3-beta1, OSX

PR #9989, opened 2020-06-12
Default value discard changes modal

rafeca:
Closes https://github.com/desktop/desktop/issues/9988

## Description

When creating the new discard selection modal in https://github.com/desktop/desktop/pull/9729 I've copy-pasta'd it from the `<DiscardChangesDialog />` which has a property called `confirmDiscardSelection`.

In the new `<DiscardSelectionDialog />` I decided to remove this props since it's not needed for the new dialog and set a default value for the state to `false`. Unfortunately I made a mistake and put the default value to `false` when it should be `true` (since the value of the checkbox is the negation of this state property).

A bit unrelated but to give some context, in the `<DiscardChangesDialog />` passing the `confirmDiscardSelection` is currently needed because that dialog could get opened even when the `confirmDiscardSelection` was set to `false` (this is because when discarding all files the dialog is shown even if the user chose to not show the discard changes dialog again).

## Release notes

Notes: [Fixed] Set default value of checkbox of Discard line confirmation dialogue to false.

### DIFF
--- a/app/src/ui/discard-changes/discard-selection-dialog.tsx
+++ b/app/src/ui/discard-changes/discard-selection-dialog.tsx
@@ -50,13 +50,9 @@ export class DiscardSelection extends React.Component<
   IDiscardSelectionProps,
   IDiscardSelectionState
 > {
-  public constructor(props: IDiscardSelectionProps) {
-    super(props)
-
-    this.state = {
-      isDiscardingSelection: false,
-      confirmDiscardSelection: true,
-    }
+  public state: IDiscardSelectionState = {
+    isDiscardingSelection: false,
+    confirmDiscardSelection: true,
   }
 
   private getOkButtonLabel() {
--- a/app/src/ui/discard-changes/discard-selection-dialog.tsx
+++ b/app/src/ui/discard-changes/discard-selection-dialog.tsx
@@ -50,9 +50,13 @@ export class DiscardSelection extends React.Component<
   IDiscardSelectionProps,
   IDiscardSelectionState
 > {
-  public state: IDiscardSelectionState = {
-    isDiscardingSelection: false,
-    confirmDiscardSelection: true,
+  public constructor(props: IDiscardSelectionProps) {
+    super(props)
+
+    this.state = {
+      isDiscardingSelection: false,
+      confirmDiscardSelection: true,
+    }
   }
 
   private getOkButtonLabel() {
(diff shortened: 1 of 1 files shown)


app/src/ui/discard-changes/discard-selection-dialog.tsx
```diff
@@ -50,13 +50,9 @@ export class DiscardSelection extends React.Component<
...
   IDiscardSelectionState
 > {
-  public constructor(props: IDiscardSelectionProps) {
-    super(props)
-
-    this.state = {
-      isDiscardingSelection: false,
-      confirmDiscardSelection: false,
-    }
+  public state: IDiscardSelectionState = {
```
niik, line 53: It feels really weird to me to declare state as public like this. Likely it's just a remnant of my c# days but given that this is a brand new pattern that we don't use anywhere else in the code base (that I know of) could we keep this PR focused on the flipping of the property and maybe discuss the pros/cons of switching patterns at our next sync?

rafeca: Oh yes! Didn't realize it's not used anywhere else in this codebase 😅  (I got confused because we're already using instance fields for other object fields like refs and instance methods for event handlers that need to be autobound, but you're right that the state it's currently always initialized inside the constructor).

For the upcoming discussion, a couple of points:
* Behind the scenes both forms are exactly equal. I agree that seeing the state being declared as `public` looks weird at first sight, but that's how it's actually typed in the React type declarations so this is just making it more obvious...
* I'm not a big fan of defining a constructor on React components: their usefulness is quite limited (they should only be used to initialize instance fields or instance methods - see the [React docs](https://reactjs.org/docs/react-component.html#constructor) -, which can be done via the instance field syntax). Any other logic should instead be included in the `componentDidMount()` method. Having components with an already defined constructor makes it easier to keep adding logic to it and end up with side effects there. I've found a few examples of components with side effects in the constructor in our codebase ([one of them](https://github.com/desktop/desktop/blob/development/app/src/ui/dialog/dialog.tsx#L158-L160)), so this is something that definitely ends up happening.
* As a con, in some components we currently have not-so-trivial-logic to calculate the state based on props ([example](https://github.com/desktop/desktop/blob/development/app/src/ui/add-repository/create-repository.tsx#L98-L119)). In these cases we would have to choose whether we keep the constructor or we create a `createInitialState()` method like we have in other components.

PR #10012, opened 2020-06-16
Header, License

ricksentam:
Added Header to let users/ client to access the source code. 

Also, I think we need to discuss about License. Shouldn‘t we change it to Open source instead of MIT. As previous version mentioned MIT License is not for GitHub. Therefore temporary empty it, for us to discuss what we should adapt a suitable license for this GitHub Desktop.

Closes #[issue number]

## Description

-

## Release notes

Notes:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-# [GitHub Desktop](https://desktop.github.com)
-
+# [../](https://github.com/desktop/desktop/blob/testing-123)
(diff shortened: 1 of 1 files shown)


niik:
Thanks for your submission.

It appears that you've created a pull request that doesn't add functionality to the GitHub Desktop project so we're going to go ahead and close this. If it was intentional, please let us know what you were intending and we can see about reopening it. We are not looking to change the license of GitHub Desktop at this time but feel free to open an issue to discuss it further

Thanks again!

PR #10013, opened 2020-06-16
Username must store as No-Case characters to IndexDB (as in 'login' field)

### LINKED ISSUE #9890: Usernames in the repository list view are shown with full lowercase characters
### Describe the bug

Usernames in the repository list view are shown with full lowercase characters; in other words, usernames in the view have their capitalization stripped.

With the exception of #9997, usernames are correctly cased and displayed elsewhere in the application.

### Version & OS

2.5.1-beta2 (Mac OS 10.14.6)

1. Open a repository belonging to a user with at least one uppercase character in their GitHub username (e.g., "RA80533")
2. Click on the **Current Repository** dropdown on the left-hand side
3. Note how the capitalization in the username has been stripped

### Expected behavior

The username should have the same character casing as it does elsewhere in the application.

### Actual behavior

The username is displayed in full lowercase characters.

> Usernames in the repository list view are shown with full lowercase characters; in other words, usernames in the view have their capitalization stripped.

### Screenshots

Screen Shot 2020-05-28 at 9 47 36 PM

### Logs

```console
$ cat "~/Library/Application Support/GitHub Desktop/logs/2020-05-29.desktop.beta.log"
2020-05-29T02:06:52.434Z - info: [ui] [AppStore] loading 1 repositories from store
2020-05-29T02:06:52.436Z - info: [ui] [AppStore] found account: RA80533 (RA80533)
2020-05-29T02:06:52.904Z - info: [ui] launching: 2.5.1-beta2 (Mac OS 10.14.6)
2020-05-29T02:06:52.905Z - info: [ui] execPath: '/Applications/GitHub Desktop.app/Contents/Frameworks/GitHub Desktop Helper (Renderer).app/Contents/MacOS/GitHub Desktop Helper (Renderer)'
2020-05-29T02:06:52.963Z - info: [ui] [BranchPruner] No branches to prune.
```

### Additional context

Related: #9997

rahul3v:
We show user's usernames as NO-Charecter-Casing. So to avoid that we must store username as NO-Case  in IndexDB (inside 'login ' field in this case)
could fix #9890

### DIFF
--- a/app/src/lib/stores/github-user-store.ts
+++ b/app/src/lib/stores/github-user-store.ts
@@ -532,6 +532,6 @@ function userWithLowerCase(user: IGitHubUser): IGitHubUser {
   return {
     ...user,
     email: user.email.toLowerCase(),
-    login: user.login,    // Not LowerCaseing to avoid username casing
+    login: user.login,    // Not Lowercasing to avoid username casing
   }
 }
--- a/app/src/lib/stores/github-user-store.ts
+++ b/app/src/lib/stores/github-user-store.ts
@@ -532,6 +532,6 @@ function userWithLowerCase(user: IGitHubUser): IGitHubUser {
   return {
     ...user,
     email: user.email.toLowerCase(),
-    login: user.login.toLowerCase(),
+    login: user.login,    // Not LowerCaseing to avoid username casing
   }
 }
--- a/app/src/lib/stores/repositories-store.ts
+++ b/app/src/lib/stores/repositories-store.ts
@@ -385,7 +385,6 @@ export class RepositoriesStore extends TypedBaseStore<
   }
 
   private async putOwner(endpoint: string, login: string): Promise<Owner> {
-    login = login.toLowerCase()
 
     const existingOwner = await this.db.owners
       .where('[endpoint+login]')
@@ -412,7 +411,7 @@ export class RepositoriesStore extends TypedBaseStore<
       parent = await this.putGitHubRepository(endpoint, gitHubRepository.parent)
     }
 
-    const login = gitHubRepository.owner.login.toLowerCase()
+    const login = gitHubRepository.owner.login
     const owner = await this.putOwner(endpoint, login)
 
     const existingRepo = await this.db.gitHubRepositories


niik:
Hi @rahul3v, thanks for wanting to contribute to GitHub Desktop!

I took a quick look at your proposed change and while the build failure should be easy enough to fix (seems like you [accidentally removed an entire line](https://github.com/desktop/desktop/pull/10013/files#diff-ec18b0d3b956763a4b5751f63da459d4L388) where only dropping the `.toLowerCase` would have sufficed) I'm on the fence on whether we might have to go down the route of migrating the database here.

Since IndexedDB is case-sensitive and since the login field is already part of a [compound index](https://github.com/desktop/desktop/blob/development/app/src/lib/databases/repositories-database.ts#L100) which we use to look up repositories this change would cause us to create new owner records whenever we encounter an owner that's not all-lowercase. That in turn would leave a number of owner objects dangling and there would be a (temporary) mismatch between which ownerId each GitHubRepository uses.

That might be an acceptable tradeoff, I'm not sure yet. I'll discuss it with @rafeca when he's back from vacation but I wanted to let you know why we haven't reviewed this PR yet.

rahul3v:
hello @niik

Well, I can understand the concern about the mismatch problem, but we can do one thing then, on initial update we have to include feature to check for loweCaseMatch and then update that key with the actual one, that would prevent this issue. I think for that i have to dig a little deeper.

Taking about the  **accidentally removed an entire line** as you mentioned. if am not wrong you are pointing the **lineNO-388**, well in that case, just removing `.toLowerCase` lefts with `login=login` , which don't make sense.

And let me know what you thinking about how you wanna handle the issue finally. (I do have suggested a solution above),  so that I can see for the change to make

VineGaming:
> We show user's usernames as NO-Charecter-Casing. So to avoid that we must store username as NO-Case in IndexDB (inside 'login ' field in this case)
> could fix #9890

iabooo553 (dismissed):

app/src/lib/stores/github-user-store.ts
```diff
@@ -532,6 +532,6 @@ function userWithLowerCase(user: IGitHubUser): IGitHubUser {
   return {
     ...user,
     email: user.email.toLowerCase(),
-    login: user.login.toLowerCase(),
+    login: user.login,    // Not LowerCaseing to avoid username casing
```
say25, line 535:
```suggestion
    login: user.login,    // Not Lowercasing to avoid username casing
```

sa-orlov (approved):

live125:
adım aynur
müzik ve sarkı soylemeyi çok seviyorum ve ünlü bir şarkıcı olmak istiyorum

Griffithjosh (approved):

michelkk96 (commented):
short

niik:
Thanks for wanting to contribute here @rahul3v, we discussed this in a recent sync and decided that we needed a database migration to solve this. I've opened #13706 to address this and I'll close this PR in favor of it.

rahul3v:
Hello @niik
Well, I believe it might not needed a database to solve, 
as an example  'user12a' or 'user12A', are not really any different users, it just the way GitHub shows or might save as 
case-insensitive usernames, if am not wrong.

so there is no chance of a user to be the same as it's case insensitive,

**As there is one more way you can solve this**:
- You can just make it true for both ( 'user12a' or 'user12A') case-insensitive usernames, but where you show usernames on the interface, just show it as "case-insensitive" everywhere 

Let me know what you think ! :)

PR #10020, opened 2020-06-16
Detect whether GitHub.com supports API username password auth

### LINKED ISSUE #9231: GitHub Desktop authentication primarily through browser flow

niik:
~🌵 Based on #10021, please review that first 🌵~

Closes #9231

## Description

As per #9231 we're now taking the `verifiable_password_authentication` API meta property into account when determining whether GitHub.com supports signing in with username and password.

If we haven't yet received a response from the GitHub.com API or if we fail to do so we fall back to honoring the [deprecation timeline](https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/)

### Today

image

### When the API says nope

image

## Release notes

Notes: TBD

### DIFF
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -12,7 +12,7 @@ export function promiseWithMinimumTimeout<T>(
   action: () => Promise<T>,
   timeoutMs: number
 ): Promise<T> {
-  return Promise.all([action(), timeout(timeoutMs)]).then(x => x[0])
+  return Promise.all([action(), sleep(timeoutMs)]).then(x => x[0])
 }
 
 /**
@@ -22,6 +22,6 @@ export function promiseWithMinimumTimeout<T>(
  *
  * @param timeout the time to wait before resolving the promise (in milliseconds)
  */
-export async function timeout(timeout: number): Promise<void> {
+export async function sleep(timeout: number): Promise<void> {
   return new Promise(resolve => window.setTimeout(resolve, timeout))
 }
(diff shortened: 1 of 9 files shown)


niik:
> this all looks good to me, but i think [sign-in.tsx](https://github.com/desktop/desktop/blob/development/app/src/ui/sign-in/sign-in.tsx) needs to incorporate this logic before we can close #9231, right?

Ah, I should have addressed this explicitly in the PR body, sorry about that. `sign-in.tsx` [already has support for web flow only endpoints](https://github.com/desktop/desktop/blob/56e1a34ed8d90baf8d991e30e58be4dedfa86a30/app/src/ui/sign-in/sign-in.tsx#L183). It was added to support the SAML-enforced web flow in https://github.com/desktop/desktop/pull/8327.

6TUKATA:
heloooooooooooo

PR #10021, opened 2020-06-16
Promise timeout helper

niik:
## Description

Extracted from #10020 to help lessen the review burden. We needed a promise timeout function, i.e. a helper method that would return a fallback value should the original promise take too long to  resolve.

While we could have implemented this by composing the old `timeout` method in a race like this

```ts
Promise.race([promise, timeout(500).then(() => fallbackValue)])
```

It felt like a useful enough helper method that we wanted to avoid keeping the timer (and thereby the closure encompassing `fallbackValue`) alive for any longer than necessary. Imagine this super contrived scenario:

```ts
timeout(someExpensivePromise, 3600000, someGinormousDiffOrSomething)
```

If we composed like above then `someGinormousDiffOrSomething` would be kept alive for an hour even if `someExpensivePromise` resolves immediately.

### Rename

We already had a promise helper method called `timeout`. After spending a few minutes trying to come up with a different name for this new helper we concluded that renaming the old helper to `sleep` would both rid us of the need to come up with a creative name as well as make it more clear what that method did.

### ES2018

In order to implement this timeout method elegantly we needed `Promise.prototype.finally` which is part of ES2018. After investigating https://kangax.github.io/compat-table/es2016plus/ we were confident that the version of Chrome that we run on (78) supports all of ES2018. Nevertheless we didn't want to take the step to upgrade to ES2018 in this PR and opted instead to only specifically include that part of ES2018 in our TSConfig. We'd be happy to entertain the thought of moving us to ES2018 fully in a separate PR.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/promise.ts
+++ b/app/src/lib/promise.ts
@@ -12,7 +12,7 @@ export function promiseWithMinimumTimeout<T>(
   action: () => Promise<T>,
   timeoutMs: number
 ): Promise<T> {
-  return Promise.all([action(), timeout(timeoutMs)]).then(x => x[0])
+  return Promise.all([action(), sleep(timeoutMs)]).then(x => x[0])
 }
 
 /**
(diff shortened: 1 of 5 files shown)


outofambit:
i'm going to go ahead and merge this as theirs no one else to review it after me (and its a low risk change).

Issue #9735, opened 2020-05-07
upgrade to typescript 3.7


outofambit:
## Benefits

[TypeScript 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) introduces some very nice syntax additions:

- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

We work around these cases with far more verbose code quite often in our project.

There's also better [typechecking for functions that return `never`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#better-support-for-never-returning-functions) (which we also use in many places) and [the addition of assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) (which @rafeca and I have previously discussed as being a useful tool for us.)

## Risks

There are [some breaking changes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#37-breaking-changes) in TypeScript 3.7. None of these should affect us significantly, and can all be caught and fixed at compile-time. Therefore the risk here is fairly small.

## Dependencies

We'll need the following things upgraded first or in tandem to keep our tooling compatible:

- [ ] prettier to >= `v1.19`
- [ ] typescript-eslint to >= `2.4.0`

rafeca:
yay! I'm specially looking forward for assertions (to be able to implement an `invariant()` function) and optional chaining (to remove a few of the `null` checks).

niik:
> yay!

:point_up:

This is an upgrade of TypeScript that I'm excited about. I think it has the potential of significantly increasing readability by letting us get rid of many explicit type narrowing gates.

I'm in favor of moving us to this version sooner rather than later but even though no-one is even suggesting it I want to caution against the urge of rewriting all the things (even though it's so incredibly incredibly tempting).

I think we should take a slow, gradual approach towards leveraging these new tools so that we don't inadvertently introduce subtle behavioral changes and so that we have time to find any unforeseen nuances (if any).

rafeca:
TypeScript has already been upgraded in https://github.com/desktop/desktop/pull/10025, we can close this issue! 🎉

PR #10025, opened 2020-06-16
Upgrade to TypeScript 3.9.5

amcasey:
## Description

- Newer versions of TypeScript have more sophisticated control-flow analysis, so a bunch of dummy returns/throws/asserts became provably unreachable.

### Screenshots

I don't expect any change to the runtime behavior, but someone more familiar with the project should confirm.

## Release notes

Notes: no-notes


amcasey:
I'm seeing a test failure locally, but I suspect it's user error.  Investigating.

```
$ cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe --auto-servernum -- jest --config ./app/jest.integration.config.js
 FAIL  app/test/integration/launch-test.ts (18.336s)
  App
    × opens a window on launch (9286ms)

  ● App › opens a window on launch
```

amcasey:
I thought I was so clever, cleaning up all the whitespace changes in otherwise untouched code...

amcasey:
This new warning doesn't affect CI:
```
WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.
SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.6.0
YOUR TYPESCRIPT VERSION: 3.9.5
```

Do you want me to fix it by upgrading @typescript-eslint?

amcasey:
[Moved out of description]

Ulterior motive: this is a cool project and it'll be easier for the TypeScript team to use it for performance investigations and backward-compatibility checks if it's on the latest version of TypeScript.

app/src/ui/app.tsx
```diff
@@ -395,8 +395,6 @@ export class App extends React.Component<IAppProps, IAppState> {
       case 'find-text':
         return this.findText()
     }
-
-    return assertNever(name, `Unknown menu event name: ${name}`)
```
rafeca, line 399: We used `assertNever` here not only to catch errors whenever the switch statement did not handle all potential values of `name` but also to throw a runtime error in the case we receive an event with an unknown `name`.

I see that now with the upgraded TypeScript we're going to be able to catch the first issue but do you know if there's a way to keep these runtime checks (or similar ones) without TypeScript complaining about unreachable code and without setting the global `allowUnreachableCode` config to true?

amcasey:
> Thanks @amcasey for the contribution! This is something we were planning to do in #9735, so this PR is very appreciated :)

Oops, it didn't even occur to me that you might have an existing work item. :smile:

> I've added a few comments: I see that you tried to not modify the current logic when fixing the new TypeScript errors that have appeared after upgrading, but I think we can take this opportunity to fix some of the logic that was wrong and you discovered thanks to this upgrade.

If a domain expert (hint hint) wants to weigh in, I'm happy to update the code as appropriate.

> Yesterday I discussed with @niik that we should try to keep the `assertNever()` calls even after this upgrade, since `assertNever()` also adds a runtime protection in case some wrongly casted value slips through one of these switch cases.

I removed the calls not because they were unnecessary but because they're now erroneous (since they're unreachable).  I believe there's a pattern for enforcing `switch` exhaustiveness, but I can't remember it off the top of my head.  Let me know if you'd like me to dig it up.

> > Do you want me to fix it by upgrading @typescript-eslint?
> 
> Thanks for the offer! I can handle this in a separate PR (unless you're interested in doing the upgrade yourself).

I've actually already done the work, but didn't want to force it on you.  If you're interested, I can add it to this PR.

> That's really nice! Happy to be able to help the TypeScript project, we love it ❤️

:heart:

app/src/ui/app.tsx
amcasey: I'm pretty sure there is - let me get back to you.

amcasey: There is, but it looks awfully similar to what you're doing now: https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript

Maybe I misread the error and it was actually saying you were missing a case?

amcasey: I can't say that I find this answer particularly satisfying, but it works if you [move ](https://www.typescriptlang.org/play/#code/KYOwrgtgBAwg9gGzgJygbwLACgq6gJWABMAabPKAcWWFDJzwCEExhsBfbbAMzBAGMALgEs4IKAEMAzlODJBAVRA0J-ABYSARgmAAKAB4AuKCGAA3OQEpjpi6kwNcgtcjgB3E8A8BRZK+S6AEQAIsJEIADkglDA+gAOwEJQgnBQAObA0WpywIGWANwcXFi8AiJi6ZnwSMgAchIQevzG1SjWUFKCyMIgaejkeFJuwoLquvyW-Y4UUPzSwLCIKAB0hESGAzMzNIJgyOIRNEQRhdMUc7KLNcvUtCAbZ1tQO3sHaTSgJ5szFwutyMtmKwHk9tplXlAItpWF9HrgiMBuBIwAhBCDQXhpLJ5EoVOotDpxgVvpwsOwgA)the assert into the `default` case.  That was probably the only way we could distinguish switch exhaustiveness checks from actually unreachable code.

app/src/lib/shells/shared.ts
```diff
@@ -118,7 +118,7 @@ function addErrorTracing(
   cp: ChildProcess,
   onError: (error: Error) => void
 ) {
-  cp.stderr.on('data', chunk => {
+  cp.stderr!.on('data', chunk => {
```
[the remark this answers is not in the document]
amcasey: This will be a syntax error for estree without the upgrade, so I guess I'll go ahead with that change.

app/src/ui/app.tsx
amcasey: I've restored the `assertNever`s but am holding off on pushing until I figure out how to make prettier happy.

amcasey:
The prettier/eslint update required a bunch of changes.  If you'd rather hold off, we can just change `cp.stderr?.on` to `cp.stderr && cp.stderr.on`.  The TypeScript team doesn't benefit from any changes other than the removal of the unreachable code (and, I suppose, the new Node types), so that's the most interesting bit for me.

amcasey:
CI has a lint failure that I'm not seeing locally.  Investigating.

amcasey:
Looks like circleci (why did I have to make an account?) runs an older version of node than appveyor:

```
@typescript-eslint/eslint-plugin@3.3.0: The engine "node" is incompatible with this module. Expected version "^10.12.0 || >=12.0.0". Got "10.11.0"
 ```

File: app/src/ui/diff/text-diff.tsx [hunk 3]
```diff
@@ -187,7 +187,7 @@ function showSearch(cm: Editor) {
   ) {
     searchLabel.style.display = 'none'
     searchField.placeholder = 'Search'
-    searchField.style.width = null
+    searchField.style.width = null!
```
amcasey, line 190: @rafeca I kept the behavior as it was, but this seemed like another place you might want to make a more positive change.

rafeca: I saw that one as well but honestly I don't know what can we replace it with 😅 

I'm in favor of leaving it like this for now 🙃

rafeca:
> The prettier/eslint update required a bunch of changes. If you'd rather hold off, we can just change cp.stderr?.on to cp.stderr && cp.stderr.on. The TypeScript team doesn't benefit from any changes other than the removal of the unreachable code (and, I suppose, the new Node types), so that's the most interesting bit for me.

Yup I agree. Let's remove the upgrades to `eslint`/`prettier` and friends from this PR since that can end up being a quite big rabbit hole.

I also realized that since we're using Electron v7 (which ships with Chromium v78) so we don't have native support for optional chaining and this means we also need to add new transformers to our babel config... So yup let's not use it for now.

> Looks like circleci (why did I have to make an account?) runs an older version of node than appveyor:
> @typescript-eslint/eslint-plugin@3.3.0: The engine "node" is incompatible with this module. Expected version "^10.12.0 || >=12.0.0". Got "10.11.0"

Yup! I actually saw that the other day as well! I've just created https://github.com/desktop/desktop/pull/10045 which should upgrade Node.js to v12 everywhere. I've also added some of the commits from this PR to that one (the commits that upgrade the `@types/node` definitions), so that PR upgrades all Node.js, but I'm fine if we keep the Node.js type definitions upgrade here.

Anyways, if you remove the `eslint`/`prettier` changes from this PR, then Node.js v12 should not be needed.

amcasey:
@rafeca Sounds good - I'll drop all the extra package upgrades.  Do you want me to open a draft PR or something for when you're ready to move ahead?

app/src/ui/app.tsx
```diff
@@ -397,8 +397,6 @@ export class App extends React.Component<IAppProps, IAppState> {
       case 'find-text':
         return this.findText()
```
rafeca, line 398: It seems like here there should be an `assertNever()` right?

app/src/highlighter/tsconfig.json.stdoutS
```diff
@@ -0,0 +1,26 @@
+Files:                         24
```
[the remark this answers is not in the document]
amcasey: You know the most exasperating thing about this?  I did notice and fix this, but the fix was in the prettier change that we dropped...

amcasey:
Done.  Thanks, @rafeca!

Issue #8197, opened 2019-09-02
Allow users to stash changes without switching branches


maxloh:
Every time I wanted to stash changes in GitHub Desktop, I need to switch to another branch, choose the "Leave my changes on ..." option in pop-up, and switch back to the branch I am working on.

It would be better if there is a "stash" button in the right click menu, or in the top menu bar.

billygriffin:
Hi @maxloh, thanks for the idea. We’re open to potentially revisiting this in the future but for now we’re going to keep the current flow for stashing, as that has been by far the most common use case we’ve heard from folks. I’m going to close this issue to convey that we don’t intend to support this in the near term but if it’s something we continue to hear from others and see workflows where it would be useful to a broader cross-section of people, we’re not opposed to adding something like this down the road. Thanks again!

tierninho:
Duplicate requests, Ref #8775 and #8931

JosephTLyons:
This would be extremely beneficial to me.

thesyntaxinator:
I was looking for this feature as well and couldn't find it. Would be helpful especially if it isn't too difficult to implement.

billygriffin:
Hi folks, I'm reopening this issue to indicate that we would like to support this. Our thinking is to support stashing changes on a branch that then leverages the existing UI for viewing, deleting, and restoring a stash on a branch.

Our proposal is to do this through the context menu in the header of the changes list, and also as a menu item in "Branch" under "Discard all changes".

cc: @shieldgenerator7 @adamwych @Alvin-Seville

JosephTLyons:
@billygriffin  Excellent!  I’m so happy to hear this!

Will this only allow us to use one stash, or multiple?

billygriffin:
@JosephTLyons Awesome! It'll continue to use the same paradigm that exists today, where you can't have multiple stashes per branch. That's a good point though that we'll need to consider in implementation - if you already have a stash, we'll need to also warn that stashing again will overwrite the existing stash, just like we do in the switching branches stashing flow.

davidbarton:
I would like to have stash option available inside error dialog when trying to pull with unstaged changes. As a user this error is very frustrating and does not offer any way out. The only current workaround for this situation is to switch to different branch and back and then try to pull again.

Would even suggest pushing this out as a separate patch before you decide on how to manage stash in other scenarios.

[...]

PR #10032, opened 2020-06-17
Allow users to stash changes without switching branches

rafeca:
Closes https://github.com/desktop/desktop/issues/8197

Paired with @niik 

## Description

This PR adds the option to stash all local changes without having to checkout a different branch by adding two context menu items (one in the application menu, on the "Branch" submenu and the other on the context menu that appears when right-clicking on the changed files header:

Screen Shot 2020-06-17 at 18 51 43

Screen Shot 2020-06-17 at 18 51 07

This PR reuses the logic that keeps the changes on the current branch when checking out another PR, which means:

* The stash will get associated to the currently checked out branch (it will only be accessible from this branch within Desktop).
* If a stash already exists on the currently checked out branch, when stashing again it will get "overwritten" (a dialog will appear to inform the user about this).

### Product considerations

* The stash menu items will get disabled state when any of the following conditions is true:
  * There are no local changes.
  * The current HEAD is in a detached state (no branch currently checked out).
  * There is an unresolved git conflict.
* If there's an existing stash entry on the current branch, the following dialog will get shown (it's the same that gets shown when this scenario happens while switching branches):
* The <kbd>Cmd + Shift + S</kbd> keyboard shortcut has been added for this feature. It's not used anywhere in Desktop and by doing a quick search it doesn't seem to be used by global shortcuts in either macOS or Windows.

Screen Shot 2020-06-17 at 19 33 14

## Release notes

Notes: [Added] Allow users to stash changes without switching branches

### DIFF
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -2325,6 +2325,24 @@ export class Dispatcher {
     return this.commitStatusStore.subscribe(repository, ref, callback)
   }
 
+  /**
+   * Creates a stash for the current branch. Note that this will
+   * override any stash that already exists for the current branch.
+   *
+   * @param repository
+   * @param showConfirmationDialog  Whether to show a confirmation
+   *                                dialog if an existing stash exists.
+   */
+  public createStashForCurrentBranch(
+    repository: Repository,
+    showConfirmationDialog: boolean = true
+  ) {
+    return this.appStore._createStashForCurrentBranch(
+      repository,
+      showConfirmationDialog
+    )
+  }
+
   /** Drops the given stash in the given repository */
   public dropStash(repository: Repository, stashEntry: IStashEntry) {
     return this.appStore._dropStashEntry(repository, stashEntry)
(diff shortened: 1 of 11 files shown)


outofambit:
> remove the `...` from the references as there is no secondary action needed by the user, as it would imply.

that one's tricky, because the user _might_ get a popup asking if they want to replace the previous stash (if one already exists for that branch in desktop)

maste197:
Ä``

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -2325,6 +2325,24 @@ export class Dispatcher {
...
 
+  /**
+   * Creates a stash for the current branch. Note that this will
+   * override any stash that already exists for the current branch.
+   *
+   * @param repository
+   * @param showConfirmationDialog  Whether to show a confirmation dialog if an
+   *                                existing stash exists (defaults to true).
+   */
+  public createStashForCurrentBranch(
```
ayed42, line 2336: .742@![Uploading 90313977...]

ayed42: we he w

PR #10034, opened 2020-06-18
Disable stash menu item when there are conflicts and make ellipsis optional

niik:
## Description

While testing #10032 @tierninho [found two issues](https://github.com/desktop/desktop/pull/10032#pullrequestreview-432808924) that we decided to address in a separate PR.

> remove the ... from the references as there is no secondary action needed by the user, as it would imply.

This is a great shout. We'll modify the label based on whether there's a current stash entry or not.

> If any of the files are conflicted, then the Stash feature does not work after selecting it and no error message is shown.

There are cases when there are files marked as conflicted in the index without there being a merge or rebase action in progress (like the revert commit case we found yesterday for example). So in addition to checking for a conflicted state we should check to see if there are any conflicted files in the working directory.

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -9,7 +9,7 @@ import { CloningRepository } from '../models/cloning-repository'
 import { TipState } from '../models/tip'
 import { updateMenuState as ipcUpdateMenuState } from '../ui/main-process-proxy'
 import { AppMenu, MenuItem } from '../models/app-menu'
-import { AppFileStatusKind } from '../models/status'
+import { hasConflictedFiles } from './status'
 
 export interface IMenuItemState {
   readonly enabled?: boolean
@@ -207,9 +207,7 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
     rebaseInProgress = conflictState !== null && conflictState.kind === 'rebase'
     hasConflicts =
       changesState.conflictState !== null ||
-      workingDirectory.files.some(
-        x => x.status.kind === AppFileStatusKind.Conflicted
-      )
+      hasConflictedFiles(workingDirectory)
     hasChangedFiles = workingDirectory.files.length > 0
   }
 
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -9,6 +9,7 @@ import { CloningRepository } from '../models/cloning-repository'
 import { TipState } from '../models/tip'
 import { updateMenuState as ipcUpdateMenuState } from '../ui/main-process-proxy'
 import { AppMenu, MenuItem } from '../models/app-menu'
+import { AppFileStatusKind } from '../models/status'
 
 export interface IMenuItemState {
   readonly enabled?: boolean
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -204,7 +204,11 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
     const { conflictState, workingDirectory } = selectedState.state.changesState
 
     rebaseInProgress = conflictState !== null && conflictState.kind === 'rebase'
-    hasConflicts = changesState.conflictState !== null
+    hasConflicts =
+      changesState.conflictState !== null ||
+      workingDirectory.files.some(
+        x => x.status.kind === AppFileStatusKind.Conflicted
+      )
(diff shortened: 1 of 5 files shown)


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -369,7 +376,9 @@ export function buildDefaultMenu({
         click: emit('discard-all-changes'),
       },
       {
-        label: __DARWIN__ ? 'Stash All Changes…' : '&Stash all changes…',
+        label: askForConfirmationWhenStashingAllChanges
+          ? confirmStashAllChangesLabel
+          : stashAllChangesLabel,
```
rafeca, line 381: I noticed that we're not updating the label of the app menu after it's initially set.

It would be cool to fix but I also don't think it's a big deal (specially since I don't see that we update labels for any other app menu items in `menu-update.ts`).

Anyways, I think that we decide to not update the label here we should have a fixed label (e.g the one with ellipsis) so it remains always static and it does not vary depending on the initial state when Desktop is launched

niik: > I noticed that we're not updating the label of the app menu after it's initially set.

That's odd, it's updating for me. Are you sure you rebuilt the main process before testing?

Screen Shot 2020-06-18 at 14 05 53
Screen Shot 2020-06-18 at 14 06 09

rafeca: Ups I didn't test it, just looked at the code 🤡 

I thought that `menu-update.ts` needed to be called in order to update menu items... I'm gonna recheck the logic to get a better understanding of how it works :D

rafeca: Ah ok now I saw it! Both the `build-default-menu.ts` logic and the `menu-update.ts` logic gets executed when `refreshRepository()` is called.

The main difference that I see is that `menu-update.ts` is also called every time the AppStore emits a change (which I assume it happens way more often).

Have we considered having a single entrypoint to update the menu that only gets executed on `refreshRepository()`? Or there are scenarios where we want to update enabled value of an item where `refreshRepository()` is not called?

app/src/lib/menu-update.ts
```diff
@@ -204,7 +205,11 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
     const { conflictState, workingDirectory } = selectedState.state.changesState
 
     rebaseInProgress = conflictState !== null && conflictState.kind === 'rebase'
-    hasConflicts = changesState.conflictState !== null
+    hasConflicts =
+      changesState.conflictState !== null ||
+      workingDirectory.files.some(
+        x => x.status.kind === AppFileStatusKind.Conflicted
+      )
```
outofambit, line 212:
```suggestion
      hasConflictedFiles(workingDirectory)
```
i think this could use this helper (which does the same thing) https://github.com/desktop/desktop/blob/6287ef907070a6a205f241a8066b3f5e1f71e278/app/src/lib/status.ts#L62-L66

app/src/ui/changes/changes-list.tsx
```diff
@@ -340,6 +340,19 @@ export class ChangesList extends React.Component<
     }
 
     const hasLocalChanges = this.props.workingDirectory.files.length > 0
+    const hasStash = this.props.stashEntry !== null
+    const hasConflicts =
+      this.props.conflictState !== null ||
+      this.props.workingDirectory.files.some(
+        x => x.status.kind === AppFileStatusKind.Conflicted
+      )
```
outofambit, line 348: same thing here
```suggestion
      hasConflictedFiles(this.props.workingDirectory.files)
```

app/src/main-process/menu/build-default-menu.ts
niik: > Have we considered having a single entrypoint to update the menu that only gets executed on `refreshRepository()`? Or there are scenarios where we want to update enabled value of an item where `refreshRepository()` is not called?

I think one scenario where we want to is when there's no repository selected. Now that you mention it though it seems odd that we update stuff in `refreshRepository`. I was under the impression that `refreshRepository` could be called for a repository which isn't currently selected  (i.e. when refreshing repository indicators).

As for the single entry point for both name and enabledness I think what happened is I wrote the stuff for toggling menu item enabledness and then the label stuff happened separately from that so chances are it's not a conscious decision but rather something that's evolved over time.

tierninho:
Changes look good. Thanks for fixing!

PR #10053, opened 2020-06-22
Show specific dialog when we get stashed files git errors

### LINKED ISSUE #8197: Allow users to stash changes without switching branches

rafeca:
Closes https://github.com/desktop/desktop/issues/8197

~~⚠️ This PR is based on https://github.com/desktop/desktop/pull/10077. Please do not merge before https://github.com/desktop/desktop/pull/10077 has been reviewed and merged.~~

> Steps to test this PR locally before a new `dugite` version with https://github.com/desktop/dugite/pull/403 gets released:~~
>  * Clone the `dugite` repository
>  * checkout https://github.com/desktop/dugite/pull/403
> * `npm install && npm run build`
>  * `yarn link`
>  * Go back to the folder where Desktop is checked out and execute `yarn link dugite` from the `app/` folder.
> 
> (this PR is expected to fail until https://github.com/desktop/dugite/pull/403 gets shipped).

**edit**: The info above ☝️ is not relevant anymore, since a new `dugite` version has already been shipped.

## Description

This PR adds a new error handler for 3 dugite errors (`LocalChangesOverwritten`, `MergeWithLocalChanges` and `RebaseWithLocalChanges`) that get triggered when a git command cannot be executed due to existing local changes.

When this error gets intercepted a generic modal gets displayed with an option to "Stash the Changes and Continue", which internally will stash the local changes and retry the action that caused the error:

Screen Shot 2020-06-22 at 15 02 37

### retryAction

In order to retry the action, this PR makes use of the retry mechanism that's incorporated in some `git` calls. Since not all `git` calls currently have retry logic, this modal will only get shown whenever the error contains retry logic (in any other case the generic error modal will get shown, as it was done before).

Currently, Merge operation don't have retry logic so we can add it to them in order to get this dialog also when merging branches.

### Existing stash

If there's already an existing stash, the dialog won't offer the option to stash changes:

Screen Shot 2020-06-22 at 15 18 38

I think that we can improve this by showing a clear warning message in this scenario that tells the user that the current stash will get overwritten and allow the user to stash the changes. I initially tried to trigger the standard overwrite stash warning popup when this happens, but turns out that opening a popup within another popup creates a few edges cases.

I'm going to experiment with different ways to show that warning within the modal (cc @ampinsk and @billygriffin if you have any suggestion here).

## Release notes

Notes: [Added] Suggest to stash changes when trying to do an operation that requires a clean working directory.


app/src/ui/app.tsx
```diff
@@ -1300,6 +1313,8 @@ export class App extends React.Component<IAppProps, IAppState> {
       return null
     }
 
+    const onDismissed = this.getOnPopupDismissedFn(popup.type)
+
```
rafeca, line 1317: As discussed offline with @niik, we're creating a memoized version of `onPopupDismissed` that passes the `popupType` that gets opened, so when dismissing a modal it will only close itself and not other modals that could have got opened afterwards.

For now I've decided to only use this method on the new modal. It would be good to use it for every other modal but that would require extensive testing since it's possible that some modals are relying on the fact that currently `onDismissed` closes any other modal.

tierninho:
> (this PR is expected to fail until desktop/dugite#403 gets shipped).

I plan to wait to test this PR unless you suggest otherwise. Thanks!

niik:
> I plan to wait to test this PR unless you suggest otherwise. Thanks!

@tierninho Feel free to test to your heart's content!

app/src/ui/dispatcher/error-handlers.ts
```diff
@@ -690,6 +690,59 @@ export async function schannelUnableToCheckRevocationForCertificate(
...
+    return error
+  }
+
+  const dugiteError = gitError.result.gitError
+  if (dugiteError === null) {
+    return error
+  }
+
+  if (
+    dugiteError !== DugiteError.LocalChangesOverwritten &&
```
niik, line 717: It's not immediately obvious tot me why we'd be including the `LocalChangesOverwritter` error in this handler. Wouldn't that get caught by the `localChangesOverwrittenOnCheckoutHandler`?

app/src/ui/local-changes-overwritten/local-changes-overwritten-dialog.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <Dialog
+        title="Error"
+        loading={this.state.loading}
+        onDismissed={this.props.onDismissed}
+        onSubmit={this.onSubmit}
+        type="error"
+      >
+        <DialogContent>
+          <p>
+            Unable to perform this action when changes are present on your
```
niik, line 53: "this action" is a bit too generic for my taste. Could we attempt to look at the retryAction type and say something like "Unable to merge when changes..." and fall back to "perform this action" only if the retryAction isn't something we're expecting?

niik (changes requested):
Overall the code looks great. I noticed that when I attempt to rebase with local changes I'm presented with a generic dialog 


image

It's not clear to me whether this is something that this PR is intending to solve but from looking at the code it seems like it should. It seems the only reason the dialog isn't shown when rebasing is because we don' have a retry action for rebase. Could/should we add one?

app/src/ui/dispatcher/error-handlers.ts
rafeca: `localChangesOverwrittenOnCheckoutHandler()` ignores errors that don't get triggered by a checkout, see [its comment](
 https://github.com/desktop/desktop/blob/development/app/src/ui/dispatcher/error-handlers.ts#L466-L475):

> This indicates to us whether the action which triggered the
> LocalChangesOverwritten was the AppStore _checkoutBranch method.
> Other actions that might trigger this error such as deleting
> a branch will not provide this specific gitContext and that's
> how we know we can safely move the changes to the destination
> branch.

There may not be many situations where this happens, but I thought it would still be nice to catch them (I'm going to add a comment in the code).

app/src/ui/local-changes-overwritten/local-changes-overwritten-dialog.tsx
rafeca: I like it! Going to do that :)

rafeca:
> It's not clear to me whether this is something that this PR is intending to solve but from looking at the code it seems like it should. It seems the only reason the dialog isn't shown when rebasing is because we don' have a retry action for rebase. Could/should we add one?

Good point! I tried the merge workflow but totally missed that there's a rebase workflow (I thought rebases were only done when pulling based on the `pull.rebase` config option). I'm going to address that

app/src/ui/app.tsx
```diff
@@ -123,6 +123,8 @@ import { CreateTag } from './create-tag'
 import { DeleteTag } from './delete-tag'
 import { ChooseForkSettings } from './choose-fork-settings'
 import { DiscardSelection } from './discard-changes/discard-selection-dialog'
+import { LocalChangesOverwrittenDialog } from './local-changes-overwritten/local-changes-overwritten-dialog'
+import * as memoizeOne from 'memoize-one'
```
rafeca, line 127: I'm using `import * from memoizeOne` here to overcome the limitation that jest uses the commonJS version of npm dependencies, which in most cases (like `memoize-one`, `classnames` or `moment`) don't expose a default attribute to make it interoperable with default imports.

While this technically is not ES module compliant, it's a quick fix that's been already widely used in this codebase (for `moment` and `classnames`).

I'm going to create a follow-up PR to fix these in a way that makes our codebase compatible with ES6 module syntax.

niik: Looks like the star-import trick didn't quite do the trick here (pun intended) 😢 
> ```
> [at-loader] ./app/src/ui/app.tsx:196:35 
>    TS2349: This expression is not callable.
>  Type 'typeof import("/Users/distiller/Desktop/desktop/node_modules/@types/memoize-one/index")' >  has no call signatures.
> ```

niik (dismissed):
The changes look great but unfortunately it seems the build is still broken

app/src/ui/app.tsx
rafeca: Damn it! It looks like the TypeScript definitions for `memoize-one` are stricter than for the other libraries and don't allow this trick.

I'm going to remove the memoization from this module for now and create a follow-up PR to try to fix the madness when importing npm packages that expose both commonJS and ES6 modules

tierninho:
Happy to test this in beta as it already has an approval and I can avoid the yarn errors after running `yarn && yarn build:dev && yarn start`:

```
✖ ｢atl｣: Checking finished with 1 errors
(node:53756) Warning: a promise was rejected with a non-error: [object Object]
[WEBPACK] Build failed after 33.396 seconds
(node:53756) Warning: a promise was rejected with a non-error: [object Object]
(node:53756) Warning: a promise was rejected with a non-error: [object Object]
[WEBPACK] Errors building ask-pass.js
[at-loader] ./app/src/ui/app.tsx:196:35
    TS2349: This expression is not callable.
  Type 'typeof import("/Users/tierninho/github/desktop/node_modules/@types/memoize-one/index")' has no call signatures.
✖ ｢atl｣: Checking finished with 1 errors
➜  desktop git:(prompt-to-stash) ✖ ｢atl｣: Checking finished with 1 errors
✖ ｢atl｣: Checking finished with 1 errors
✖ ｢atl｣: Checking finished with 1 errors
```

Feel free to merge when ready.

app/src/ui/app.tsx
```diff
@@ -186,6 +187,15 @@ export class App extends React.Component<IAppProps, IAppState> {
...
 
+  /**
+   * Returns a memoized instance of onPopupDismissed() bound to the
+   * passed popupType, so it can be used in render() without creating
+   * multiple instances when the component gets re-rendered.
+   */
+  private getOnPopupDismissedFn = (popupType: PopupType) => {
+    return () => this.onPopupDismissed(popupType)
+  }
+
```
niik, line 198: Should we drop this in favor of reintroducing it properly in the memoization PR?
```suggestion
```

app/src/ui/app.tsx
```diff
@@ -1989,6 +2001,24 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+            ? selectedState.state.changesState.stashEntry
+            : null
+
+        return (
+          <LocalChangesOverwrittenDialog
+            repository={popup.repository}
+            dispatcher={this.props.dispatcher}
+            hasExistingStash={existingStash !== null}
+            retryAction={popup.retryAction}
+            onDismissed={this.getOnPopupDismissedFn(popup.type)}
```
niik, line 2019:
```suggestion
            onDismissed={this.onPopupDismissed}
```

niik (changes requested):
Can we clean up the remaining pieces of the memoized specific popup dismissal in favor of the forthcoming PR?

app/src/ui/app.tsx
rafeca: Without this check, in some situations the dismissal of this popup will cause other popups to get closed: I've just tested it and when merging the conflicts modal will get automatically closed after stashing the changes.

Anyways, I'm planning to send a PR with the memoization very soon, so in either case we'll fix the bug or the performance issue very soon 😃

niik: Ah, that's great context that I was missing. Sounds great, we can live with the performance issue for a brief period of time I think.

niik (approved):

WALLOUD:
***

PR #10058, opened 2020-06-22
Update prettier and eslint for optional chaining support

amcasey:
Follow up to #10025.  This should make it possible to use features like `?.`.


amcasey:
FYI @rafeca

rafeca:
Awesome! Now that both the TypeScript and Node.js PRs have been upgraded if you rebase this PR hopefully CI will pass 🤞

package.json
```diff
@@ -172,4 +172,4 @@
     "electron-packager": "^14.2.1",
     "electron-winstaller": "4.0.0"
   }
-}
+}
```
rafeca: Let's keep the end line here to not have it changed back in another PR

```suggestion
}

```

.prettierrc.yml
```diff
@@ -2,6 +2,8 @@ singleQuote: true
 trailingComma: es5
 semi: false
 proseWrap: always
+endOfLine: auto
```
rafeca, line 5: Does keeping the default value for `endOfLine` produce lots of changes in the PR? If not, I prefer to have the default value (`lf`) to not have mixed end of line flavors.

```suggestion
```

.eslintrc.yml
```diff
@@ -24,10 +24,18 @@ rules:
...
+  '@typescript-eslint/naming-convention':
     - error
-    - always
-  '@typescript-eslint/no-angle-bracket-type-assertion': error
+    - selector: interface
+      format:
+        - PascalCase
+      custom:
+        regex: '^I[A-Z]'
+        match: true
```
rafeca, line 34: Nice! Much more explicit now!

package.json
amcasey: I have no preference - I leave it in whatever state it's in after `lint:fix` runs.

.prettierrc.yml
amcasey: Yes, the diff was huge, so I set it back to the old value.

.eslintrc.yml
amcasey: Stolen from the typescript-eslint docs. :wink:

.prettierrc.yml
amcasey: If I'm reading your comment correctly, it sounds like you're on board with saving this change for a future PR.

package.json
amcasey: And it looks like `lint:fix` took care of it in the second commit. :smile:

amcasey:
@rafeca Maintainer edits are enabled, so please feel free to make whatever changes you see fit.  From my perspective, things look good.

rafeca:
Thanks for the response! I think we can merge this one then 😃

package.json
rafeca: Ah cool, nevermind then 😄

PR #10066, opened 2020-06-23
Refactor logic to open context menus

rafeca:
This PR does a small refactor to the logic to address two separate things:

1. Potential race conditions that could potentially be caused if a "open contextMenu" action happens at roughly the same time as the user clicks on an item of an already opened context menu. See [this comment](https://github.com/desktop/desktop/pull/9796#discussion_r429218437) and [this other one](https://github.com/desktop/desktop/pull/9796#discussion_r429241288) for more information:
    > This makes we wonder whether apart from fixing this specific undefined issue we should think about adding some logic to warranty that the currentContextualMenuItems object doesn't change between the time when the context menu gets opened and the time that an item gets clicked: e.g it would also be very bad if a user clicks on a tag to delete it and we end up deleting a different one because a new tag appears on the list at the appropriate moment (we can do this in a separate PR!).

2. It removes the `action` property from the menu items before sending them via IPC. This is not necessary right now but it will prepare us for an upcoming [breaking change](https://github.com/electron/electron/blob/master/docs/breaking-changes.md#behavior-changed-values-sent-over-ipc-are-now-serialized-with-structured-clone-algorithm) in Electron v8, which disallows to send functions or other non-serializable values.

## Context

In order to fix 1), I've chosen to use the Electron [`invoke` API](https://www.electronjs.org/docs/api/ipc-renderer#ipcrendererinvokechannel-args) which allows to do Request/Response communication between the renderer process and the main process.

With this, and the fact that now Electron can notify when a context menu gets closed, we can safely implement a cross-process async method that opens the context menu and resolves whenever the user clicks on an item or discards the menu.

## Alternative solutions

I considered two other solutions to solve 1):

* Creating the `Menu` directly from the renderer process (as explained in the [Electron docs](https://www.electronjs.org/docs/api/menu#render-process)). At first sight this seems to be the simplest and most robust approach, but I discarded it since as I understand this would cause sync calls to the main process when creating the menu items which could potentially block the UI (based on [this doc](https://www.electronjs.org/docs/api/remote)). If my assumptions are wrong I think we can reconsider this approach.
* Adding a softer protection that ensures that click events correspond to the stored `currentContextualMenuItems` array (by generating unique random ids each time we open a context menu). This would prevent the problem by ignoring these click events but make the solution even more complex than it is right now.

## Potential drawbacks of this solution

The only drawback that I could find is that in the case that Electron for some reason doesn't emit the correspondant callback when the menu in closed, this could lead to a small leak of memory, since the `Promise` returned on the `invoke` call would never get fulfilled. Even if this happened the amount of memory allocated would be very small (I think it would be 1 Promise on the main process, another Promise on the renderer process and the event object used for the invocation).

I've done some tests on macOS by opening/closing the context menu multiple times with very fast transitions and couldn't find cases where the close menu event doesn't get fired, so I'm not concerned about this potential issue.

### DIFF
--- a/app/src/lib/menu-item.ts
+++ b/app/src/lib/menu-item.ts
@@ -25,6 +25,9 @@ export interface IMenuItem {
   readonly submenu?: ReadonlyArray<this>
 }
 
+/**
+ * A menu item data structure that can be serialized and sent via IPC.
+ */
 export interface ISerializableMenuItem extends IMenuItem {
   readonly action: undefined
 }
--- a/app/src/lib/menu-item.ts
+++ b/app/src/lib/menu-item.ts
@@ -22,7 +22,11 @@ export interface IMenuItem {
   /**
    * Submenu that will appear when hovering this menu item.
    */
-  readonly submenu?: ReadonlyArray<IMenuItem>
+  readonly submenu?: ReadonlyArray<this>
+}
+
+export interface ISerializableMenuItem extends IMenuItem {
+  readonly action: undefined
 }
 
 /**
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -435,15 +435,25 @@ app.on('ready', () => {
     }
   )
 
-  ipcMain.on(
+  /**
+   * Handle the action to show a contextual menu.
+   *
+   * It responds an array of indices that maps to the path to reach
+   * the menu (or submenu) item that was clicked or null if the menu
+   * was closed without clicking on any item.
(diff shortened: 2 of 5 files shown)


tierninho:
> I've done some tests on macOS by opening/closing the context menu multiple times with very fast transitions and couldn't find cases where the close menu event doesn't get fired, so I'm not concerned about this potential issue.

👍  Played around with it too and didn't notice anything odd.

PR #10073, opened 2020-06-24
Strip some ampersands

niik:
## Description

This fixes some false-positives reported in code scan. While they're false positives this is technically the more correct approach anyway which is why we're fixing them rather than ignoring the warning.


niik:
✅ Verified these changes on Windows which is the only OS affected

PR #10074, opened 2020-06-24
Exclude the vendor directory from code scanning

niik:
## Description

We're getting Code Scanning issues reported by our [vendored yarn](https://github.com/desktop/desktop/blob/d9279df6ff51f96d34e4a491390e326a26e074cd/vendor/). There's no risk of that file creating security vulnerabilities in GitHub Desktop so in order to reduce the noise we're gonna ignore it.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,5 +1,4 @@
 name: 'GitHub Desktop CodeQL config'
 
 paths-ignore:
-  - '/vendor/'
-  - 'vendor/yarn*.js
+  - 'vendor/yarn*.js'
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,4 +1,4 @@
 name: 'GitHub Desktop CodeQL config'
 
 paths-ignore:
-  - 'vendor/yarn*.js'
+  - 'vendor/**'
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -2,3 +2,4 @@ name: 'GitHub Desktop CodeQL config'
 
 paths-ignore:
   - '/vendor/'
+  - 'vendor/yarn*.js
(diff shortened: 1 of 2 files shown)


rafeca:
Doesn't seem like the code scanning skipped the yarn vendor file (at least based on the open reports for this branch: https://github.com/desktop/desktop/security/code-scanning?query=ref%3Arefs%2Fheads%2Fcodeql-config).

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,4 @@
+name: 'GitHub Desktop CodeQL config'
+
+paths-ignore:
+  - '/vendor/'
```
rafeca, line 4: Maybe this needs to be something like this?

```suggestion
  - './vendor/'
```

niik: I have no clue, it  was copied straight from https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning but https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths seems to suggest it's supposed to be a pattern. Maybe something like this?

```suggestion
  - '/vendor/'
  +'/vendor/yarn*.js'
```

niik: Actually, looking at https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet it seems like the patterns are relative to the repository root.

```suggestion
  - '/vendor/'
  - 'vendor/yarn*.js
```

niik:
I've also added a config option here so that we only build on pushes to `development`. Since development is a protected branch that we can only push to via PRs it's redundant to build for both pushes and PRs

rafeca (approved):
security

niik:
Gonna merge this early to clean up our code scanning results

.github/workflows/codeql.yml
```diff
@@ -2,6 +2,8 @@ name: 'Code scanning - action'
 
 on:
   push:
+    branches:
+      - development
```
mathieuedwards, line 6: .github/workflows/codeql.yml

mathieuedwards: .github/workflows/codeql.yml

Issue #10022, opened 2020-06-16
Upgrade git


rafeca:
This issue tracks the investigation to upgrade the version of `git` that GitHub Desktop is currently using.

## Current git version

As of now (GitHub Desktop `v2.5.2`), we're using `dugite` `v1.88.6` which includes `git` `v2.23.3` on macOS and `v2.23.0.windows.5` on Windows.

## Candidates to upgrade

I've considered 2 different versions to use for the upgrade:

1. `v2.27.0` (released on June 1st 2020).
1. `v2.26.2` (released on April 20th 2020, while `v2.26.0` was released on March 23rd 2020).

## Interesting factors to consider

* We've historically avoided upgrading to the latest `git` version to avoid potential instabilities.
* As mentioned by @dscho in [Slack](https://github.slack.com/archives/C17LP0XU3/p1591272441422800), "Git for Windows v2.27.0 seems to be not quite as robust as the median Git for Windows release". Given that there's currently no other `v2.27.x` version of Git for Windows, it seems fair to avoid `v2.27.0`.
* `v2.25.0` contains a fix for https://github.com/desktop/desktop/issues/9904.
* `v2.27.0` contains a fix for https://github.com/desktop/desktop/issues/9597 on macOS.
* `v2.26.0` enabled by default the transport protocol version 2. Given that this got reverted on `v2.27.0` due to "some remaining rough edges" ([more info](https://github.com/git/git/commit/3697caf4b962f65430fdcd6f0d41b7fe1f579e18)), we should consider changing the default if we upgrade to `v2.26.2`

## Recommendation

❗ **My recommendation is to upgrade to `git` `v2.26.2`** ❗ 

Regarding the default protocol version, after talking to @niik we found a couple of ways to override its default value to be v1 (the same value as older or newer `git` versions):
1. Override the system gitconfig to set the appropriate default protocol version. This can be easily done for Windows (we already do it for [other config settings](https://github.com/desktop/dugite-native/blob/master/script/build-win32.sh#L81-L85) but it would need some patching of the `git` sourcecode for macOS (basically applying [this commit](https://github.com/git/git/commit/11c7f2a30b9dadcccc7bde66a34e0cb0cb5cf52c#diff-6da6886a4c67f48a449b698adc074a05)), since there doesn't seem to be a system-wide gitconfig available.
[...]

Issue #9904, opened 2020-05-31
Cloning: error: filename in tree entry contains backslash" '\'


nikolapeja6:
### Describe the bug

When cloning the [MicrosoftDocs/azure-docs](https://github.com/MicrosoftDocs/azure-docs) repository though the GitHub Desktop app, it failed. The logs showed the following error:

```
error: [ui] `git -c credential.helper= -c protocol.version=2 clone --recursive --progress -- https://github.com/MicrosoftDocs/azure-docs.git C:\_repos\azure-docs` exited with an unexpected code: 128.
...
Receiving objects:  80% (2707986/3353572), 8.32 GiB | 17.28 MiB/s
error: filename in tree entry contains backslash: '\media'
fatal: Not all child objects of cfb022b5e95f43eb589592e5c3633fc41f36dba4 are reachable
fatal: index-pack failed
```
From the logs, the command that executed the clone was this:

```
git -c credential.helper= -c protocol.version=2 clone --recursive --progress -- https://github.com/MicrosoftDocs/azure-docs.git C:\_repos\azure-docs
```

However, when I tried to manually execute the command it passed. Not sure what is causing the problem.

### Version & OS

GitHub Desktop Version 2.5.0 on Windows 10 (also tried on Windows Server, failed there as well).

### Steps to reproduce the behavior

Clone the repo `https://github.com/MicrosoftDocs/azure-docs` form the GitHub Desktop application.

### Expected behavior

Cloning should be successful.

### Actual behavior

Cloning fails.

steveward:
@nikolapeja6 thanks for the report. It looks like you are running into https://github.com/git-for-windows/git/issues/2435. When we [update the version of Git](https://github.com/desktop/dugite/pull/399) that ships with GitHub Desktop it should resolve this issue. Here's the relevant bit from the release notes:

> * An earlier update to Git for Windows declared that a tree object is
   invalid if it has a path component with backslash in it, which was
   overly strict, which has been corrected.  The only protection the
   Windows users need is to prevent such path (or any path that their
   filesystem cannot check out) from entering the index.
   (merge 224c7d70fa js/mingw-loosen-overstrict-tree-entry-checks later to maint).

[Full Release Notes](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.25.0.txt)

steveward:
A workaround for this bug can be found in this [comment in issue #10004](https://github.com/desktop/desktop/issues/10004#issuecomment-644075555) by @niik.

PR #10077, opened 2020-06-24
Upgrade dugite to v1.91.3

rafeca:
Closes https://github.com/desktop/desktop/issues/10022
Closes https://github.com/desktop/desktop/issues/9904

## Description

This PR upgrades the current `dugite` version to [`v1.91.3`](https://github.com/desktop/dugite/releases/tag/v1.91.3) which includes a new version of `git` (`v2.26.2`).

These are the changes from to the currently used version `v1.88.6` ([full list of commits](https://github.com/desktop/dugite/compare/v1.88.6...v1.91.3)):

* Upgraded `git` from `v2.23.3` to `v2.26.2` (full release notes for [`v2.24.0`](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.24.0.txt), [`v2.25.0`](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.25.0.txt) and [`v2.26.0`](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.26.0.txt). Summary of relevant changes in https://github.com/desktop/desktop/issues/10022).
* Added support for Windows on ARM via the `npm_config_arch` config  (more info in https://github.com/desktop/dugite/pull/391).
* Added parsing for errors related to local changes conflicting with merge and rebase actions (more info in https://github.com/desktop/dugite/pull/403).

## Release notes

Notes: no-notes


rafeca:
There are some tests failures so we'll need to fix these before...

say25:
If I am not wrong this should also close - #9904

rafeca:
It seems like the build failures are caused by the following breaking changes on `git` `v2.26`:

> "git rebase" uses a different backend that is based on the 'merge' machinery by default.  There are a few known differences in the behaviour from the traditional machinery based on patch+apply.
>
> If your workflow is negatively affected by this change, please report it to git@vger.kernel.org so that we can take a look into it. After doing so, you can set the 'rebase.backend' configuration variable to 'apply', in order to use the old default behaviour in the meantime.

So far I've found 3 differences that cause the tests breakages:

1. The folder where the rebase information is stored has changed from `.git/rebase-apply` to `.git/rebase-merge` (we use this to calculate the current [rebase data](https://github.com/desktop/desktop/blob/development/app/src/lib/git/rebase.ts#L94)).
2. Now the error message that gets output when there are rebase conflicts is the same as when there are merge conflicts, so the separation of both errors [in `dugite`](https://github.com/desktop/dugite/blob/master/lib/errors.ts#L72-L74) does not work/is not needed.
3. Now, when executing `git rebase --continue`, `git` will launch the editor to allow the user to change the commit message. I haven't found a way to disable this behaviour ([more info](https://git-scm.com/docs/git-rebase#_commit_rewording).

There are more differences between the two backend, listed [here](https://git-scm.com/docs/git-rebase#_behavioral_differences).

say25:
> It seems like the build failures are caused by the following breaking changes on git v2.26

I wonder if tests around this should exist in Dugite itself?

rafeca:
> I wonder if tests around this should exist in Dugite itself?

Yup, some of them for sure! e.g the ones that detect the error messages, which weren't caught since a lot of tests on `dugite` hardcode the error message that `git` should output 🤦‍♂️

Other errors are more related to the logic that's implemented in Desktop (e.g the whole rebase flow and how the different subcommands play with each other), so it's going to be harder to move them to `dugite` without moving more logic to it.

rafeca:
I've spoken with @dscho and he suggested setting the `GIT_EDITOR` env variable to `:` in order to tell `git` to use the default commit message when executing `git rebase --continue`.

I've added a few commits to this PR to check if this works well in all OSes (good thing we have a test for the rebase workflow). 

If this works fine I think that's going to be a good solution (@tierninho would be good to do a bunch of manual tests on the rebase workflow to verify that the new backend doesn't cause other issues).

Something we have to decide is whether we want to keep supporting doing rebases with the `apply` backend (in the case that the user has it configured in its `.gitconfig`) or we want to set the `rebase.backend=merge` config param on every rebase command (or git command that can cause a rebase). Supporting both should be quite easy at this moment but may become tricky if the features of both backends keep diverging before the `apply` backend gets completely deprecated.

dscho:
> 3\. Now, when executing `git rebase --continue`, `git` will launch the editor to allow the user to change the commit message. I haven't found a way to disable this behaviour ([more info](https://git-scm.com/docs/git-rebase#_commit_rewording)).

As discussed elsewhere, `GIT_EDITOR=:` should be the way to go forward with this.

Explanation: the [`:` command](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_16) only expands the command-line arguments, but does nothing other than return with exit code 0 (i.e. success). Now, the `:` command is not necessarily present (Git for Windows ships without it, for example), but the POSIX shells typically handle it internally. Besides, Git special-cases `GIT_EDITOR=:` and does not even spawn a process: https://github.com/git/git/blob/c9c318d6bf26bcecdca5b6f31683b9d5887a83ee/editor.c#L56. In other words, this is designed to work everywhere, even on Windows.

dscho:
@rafeca sorry, I didn't see your comment before sending mine...

The reason why the default backend changed is actually because Git would like to deprecate the `apply` one eventually. So maybe it'll be better to switch instead of supporting both?

rafeca:
I've created https://github.com/desktop/dugite/pull/407 to fix point 2) from https://github.com/desktop/desktop/pull/10077#issuecomment-648951880.

Once https://github.com/desktop/dugite/pull/407 is merged and released, we should be able to revert https://github.com/desktop/desktop/pull/10077/commits/c3e0379166264e0bc5739d171f9ddb6b9a6d3e6c and keep all the tests passing.

rafeca:
> The reason why the default backend changed is actually because Git would like to deprecate the apply one eventually. So maybe it'll be better to switch instead of supporting both?

Cool, I'm going to keep this PR as it is without adding compatibility to the old backend. I've also added https://github.com/desktop/desktop/pull/10077/commits/60c50dbfe16b72ee486a046694ab97c4580ee1cd to this PR to force the `merge` backend on `rebase` and `pull` commands even if the user has the `apply` backend configured. Do you know of any other `git` command that can trigger a rebase? (I prefer to add that override explicitly on the commands that can trigger it rather than adding it to every single git invocation).

dscho:
> Do you know of any other `git` command that can trigger a rebase?

There is `git p4 rebase` and `git svn rebase`. Not sure whether you care, AFAIR GitHub Desktop does not support Perforce and Subversion remotes, right?

There is also `git submodule update --rebase`, though, that _might_ be something you'll need to take care of because it can be configured via `submodule.<name>.update=rebase` in either `.gitmodules` or the repository config. But then, I think that there is actually no way to tell `git submodule` which rebase backend to use...

PR #10084, opened 2020-06-25
Fix mentioning of users from a fork

niik:
## Description

Pulling this out of my hack day work since it can stand on its own. 

If you clone a fork today and try to autocomplete users or issues those issues and mentionable users will always be derived from the fork which for the majority case means that users will not be able to autocomplete (forks are overwhelmingly used only to contribute to the upstream).

Since we introduced the workflow preferences in https://github.com/desktop/desktop/pull/9842 we've let users tell us how they want to work with their fork.

This PR takes that preference into account so that if say @billygriffin attempts to autocomplete issues or users in https://github.com/billygriffin/desktop with the "To contribute to the parent repository" setting selected he'll now see the issues and mentionable users from https://github.com/desktop/desktop instead.

### Screenshots

image

## Release notes

Notes: [Improved] Autocompleting issues or users in forks now respect configured fork behavior

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -45,6 +45,7 @@ import {
   Repository,
   isRepositoryWithGitHubRepository,
   RepositoryWithGitHubRepository,
+  getNonForkGitHubRepository,
 } from '../../models/repository'
 import {
   CommittedFileChange,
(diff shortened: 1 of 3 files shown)


niik:
> Should we include a line item for this here or too minor of a change? https://github.com/desktop/desktop/blob/forks-doesnt-need-to-be-this-lonely/app/src/ui/repository-settings/fork-contribution-target-description.tsx#L19

That's a great question, I don't feel super strongly about it but maybe it should. I wasn't part of the design discussions around that view but it seems like it's pretty exhaustive right now so I've added this there as well.

image

PR #6308, opened 2018-11-28
Add metrics for capturing unattributed commits

iAmWillShepherd:
Fixes #6194 

This PR adds 3 new metrics:
1. `unattributedCommits`: number of commits that will go unattributed (this only affects those who are logged into Desktop and we do not differentiate between Dotcom and GHE here)
1. `gheCommits`: number of commits that were made to a GHE hosted repo
1. `dotcomCommits`: number of commits that were made to a repo hosted on dotcom

To whoever ends up reviewing this, please make sure to review and merge the PRs targeting github/central and github/desktop.github.com 

**Todo**
- [x] Request that github/analytics update tables - https://github.com/github/analytics/issues/788
- [x] Update central - https://github.com/github/central/pull/416 (will need to be reviewed and merged **with** this PR)
- [x] Update marketing site - https://github.com/github/desktop.github.com/pull/121 (will need to be reviewed and merged **with** this PR)
- [x] Update metrics doc - a01b548a9c3736f531efc1a38e240e5d6f93f649

PR #10092, opened 2020-06-26
Ensure both legacy and modern format stealth email addresses are counted for attribution

niik:
# Description

This is another break out from my hack-day project which can stand on its own and isn't related to the hack.

Ever since https://github.com/desktop/desktop/pull/6308 we've kept track of the number of commits that we believe will end up not being attributed to the currently signed in user once they're pushed to the remote.

We deduce this by loading all emails (including private) for the user from the API and comparing those emails to the one about to be used when committing.

This PR  fixes a small discrepancy in that logic where if one where to look at my account's email addresses from the API you'd get something like this

```
[
  {
    "email": "[MY PRIVATE  EMAIL]",
    "primary": true,
    "verified": true,
    "visibility": "public"
  },
  {
    "email": "[MY WORK EMAIL],
    "primary": false,
    "verified": true,
    "visibility": null
  },
  {
    "email": "[AN OLD EMAIL]",
    "primary": false,
    "verified": false,
    "visibility": null
  },
  {
    "email": "niik@users.noreply.github.com",
    "primary": false,
    "verified": true,
    "visibility": null
  }
]
```

That last email address is the anonymous (or "stealth" as it's referred to internally) email address. It is however in a legacy format. Modern stealth emails include the user id so as to be able to preserve attribution when renaming an account. In my case the modern version would look like this: `634063+niik@users.noreply.github.com`

Both the legacy format and the modern end up getting counted for attribution though so on the off chance that someone has configured their Git config with a version newer or older than the one attached to their account currently this will catch that.

The second part of this is that if the API doesn't return any email addresses we'll fall back to a stealth email (I don't believe this is a possibility with how the API works today but it's possible this behavior exists on some legacy GHES instance idk). This lets us simplify some logic by being able to count on having an email address for the API. This is currently only used to pre-fill a suggested email address in the welcome flow (or preferences).

### Notes to reviewer

You might find it easier to review this with whitespace changes turned off.

cc @ampinsk and @billygriffin since we discussed this tiny edge case in our last meeting


app/src/lib/email.ts
```diff
@@ -9,36 +9,35 @@ import { Account } from '../models/account'
  *
  *  - the primary email if it's publicly visible
  *  - the first public email
```
rafeca, line 11: Not related to this PR, but if I understand the logic correctly, we don't seem to be looking for the first public email, so maybe we can take the opportunity to change the logic or update the comment.

niik: Sounds good 👇 

```suggestion
```

PR #3492, opened 2017-12-03
port registry usage to native Node module

shiftkey:
Fixes #3105
Fixes #3405

Supersedes #3209 

- [x] port registry usage over to new library
- [x] update `package.json` so that non-`win32` platforms can compile
- [x] test and benchmark performance differences (should be significantly quicker)
- [x] confirm `registry-js`returns UTF-8 strings (avoiding the need for any active code page shenanigans)

- [x] confirm app still works fine with group policy enabled to block registry editing tools
- [x] test on Windows 7 VM
  - [x] VSCode Insiders install (32-bit) found
  - [x] fresh install of Atom found
  - [x] PowerShell default path found
- [x] test on Windows 8 VM
  - [x] VSCode install (64-bit) found
  - [x] Hyper shell found
  - [x] PowerShell default path found
- [x] test on Windows 10 VM
  - [x] Sublime Text found
  - [x] Git Bash found
  - [x] PowerShell default path not found
- [x] cleanup changelog and version changes

PR #1925, opened 2017-06-07
Commander of lines

joshaber:
~~_⚠️ Dependent on https://github.com/desktop/desktop/pull/1924 ⚠️_~~

Fixes #696 

On Windows, we'll automatically insert our `bin` directory into the user's `PATH`, like Atom and VS Code do. On Mac, the user will need to explicitly choose to install the tool.

While I was mucking around with Squirrel.Windows, I also fixed a few bugs:

Fixes #1875
Fixes #1759 
Fixes #1863

screen shot 2017-06-08 at 10 33 37 am

screen shot 2017-06-08 at 3 06 19 pm

PR #10110, opened 2020-06-30
Read PATH environment variable using registry-js

### LINKED ISSUE #10100: Path variable is overwritten rather than appended when GitHub Desktop is installed or updated
### Describe the bug

When GitHub Desktop is installed or updated, it overwrites the User Path variable when it includes the GitHubDesktop folder in the Path rather than appending to it.

### Version & OS

Version 2.5.3 (though I updated to it which caused the problem in addition to the first time when I installed GitHub Desktop (though I didn't realize it was GitHub Desktop that caused it and it was only after installing this update that I realized that it happened again, and the only path that remained was the GitHubDesktop\bin folder. 

OS: Windows 10 Pro 1909.

1. Either reinstall GitHub Desktop or wait for an update. After installing it, the Path variable will lose all other 

### Expected behavior

The GitHubDesktop\bin folder should be appended to the Path variable (or ignored if it already exists when an update is being installed)

### Actual behavior

The Path variable is overwritten and all other Path variables are lost, breaking functionality of other apps and scripts.

niik:
References, and may close #10100

## Description

Excerpt from https://github.com/desktop/desktop/issues/10100#issuecomment-651730371

> We currently load the `PATH` env var through powershell:
> 
> https://github.com/desktop/desktop/blob/d6f7518f43bd7f9f55e56737e2845474fe2c7ebb/app/src/lib/process/win32.ts#L20-L38
> 
> This was introduced in https://github.com/desktop/desktop/pull/1925/commits/9820f7f988e04ea4b268f1462f6dbe6bb6d8c473#diff-fc992a6c0da9f6047bd328ec09982ec2R111 and copied from Atom https://github.com/atom/atom/pull/10326/files.
> 
> The reason for this odd-looking way of loading the path variable is that the previous method of using `reg.exe query HKCU\Environment` could fail due to a ban on registry modification in a group policy on Windows even though reg.exe was only reading a value, not modifying it. This peculiarity was the reason behind implementing our own registry library (https://github.com/desktop/desktop/pull/3492).

This PR removed the powershell workaround in favor of using `registry-js` to read the PATH env var.

## Testing

@tierninho This isn't a high priority but when you've got time I'd appreciate if you could test this out. It only affects Windows but ideally we'd test it on both Windows 7 and Windows 10 and with a user that contains non-ascii characters and preferably a space as well. If you haven't got one I've been using `Håkan Bråkan`. I'm building a one-off installer for this and I'll update once the build is done.

Rough test plan

1. Inspect the `PATH` environment variable: (see https://www.architectryan.com/2018/03/17/add-to-the-path-on-windows-10/)
1. Install GitHub Desktop
1. Verify that a `...\AppData\Local\GitHubDesktop\bin` entry was appended to the path list and no other changes occurred
1. Uninstall GitHub Desktop
1. Verify that the `...\AppData\Local\GitHubDesktop\bin` entry was removed from the path list and no other changes occurred

And obviously any other ideas you have on how to break this!

### DIFF
--- a/app/src/lib/process/win32.ts
+++ b/app/src/lib/process/win32.ts
@@ -17,16 +17,13 @@ function isStringRegistryValue(rv: RegistryValue): rv is RegistryStringEntry {
 
 /** Get the path segments in the user's `Path`. */
 export function getPathSegments(): ReadonlyArray<string> {
-  const HKCU = HKEY.HKEY_CURRENT_USER
-  const value = enumerateValues(HKCU, 'Environment\\Path').find(
-    isStringRegistryValue
-  )
-
-  if (value === undefined) {
-    throw new Error('Could not find PATH environment variable')
+  for (const value of enumerateValues(HKEY.HKEY_CURRENT_USER, 'Environment')) {
+    if (value.name === 'Path' && isStringRegistryValue(value)) {
+      return value.data.split(';').filter(x => x.length > 0)
+    }
   }
 
-  return value.data.split(';').filter(x => x.length > 0)
+  throw new Error('Could not find PATH environment variable')
 }
 
 /** Set the user's `Path`. */
--- a/app/src/lib/process/win32.ts
+++ b/app/src/lib/process/win32.ts
@@ -1,41 +1,29 @@
 import { spawn as spawnInternal } from 'child_process'
 import * as Path from 'path'
+import {
+  HKEY,
+  RegistryValueType,
+  RegistryValue,
+  RegistryStringEntry,
+  enumerateValues,
+} from 'registry-js'
+
+function isStringRegistryValue(rv: RegistryValue): rv is RegistryStringEntry {
+  return rv.type === RegistryValueType.REG_SZ
+}
 
 /** Get the path segments in the user's `Path`. */
-export async function getPathSegments(): Promise<ReadonlyArray<string>> {
-  let powershellPath: string
-  const systemRoot = process.env.SystemRoot
-  if (systemRoot != null) {
-    const system32Path = Path.join(systemRoot, 'System32')
-    powershellPath = Path.join(
-      system32Path,
-      'WindowsPowerShell',
-      'v1.0',
-      'powershell.exe'
-    )
-  } else {
-    powershellPath = 'powershell.exe'
+export function getPathSegments(): ReadonlyArray<string> {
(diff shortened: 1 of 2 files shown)


tierninho:
Thank you for the testing notes! 

I have a few questions:
- I should be verifying the `...\AppData\Local\GitHubDesktop\bin` entry under the user-level path, not the system path?
- How do I spin up this PR in a dev environment and simultaneously uninstall/install Desktop without killing the dev environment?

niik:
> I should be verifying the ...\AppData\Local\GitHubDesktop\bin entry under the user-level path, not the system path?

A great question! Yes, the user-level path is what we care about.

> How do I spin up this PR in a dev environment and simultaneously uninstall/install Desktop without killing the dev environment?

Sorry, I got sidetracked and forgot to post the link. I've made a special build of this PR with an installer available at https://s3.amazonaws.com/github-desktop/releases/2.5.4-test1-1584ba94/GitHubDesktopSetup.exe. You can download that installer and install/uninstall as many times as you'd like, it will replace the prod/beta version if you've got either installed and it will never auto-update (so you'll have to remember to install the beta/prod version once you're done testing).

niik:
> but my attempt to include non-ascii characters failed as it didn't update inside the path

Could you elaborate on what you mean by this? In particular the "didn't update inside the path" part?

Regardless I tested this with my Swedish fake user and the path was inserted correctly. For some reason the path never got removed though but that was the case for the production installer as well so I'm assuming this has something to do with my VM

tierninho:
> Could you elaborate on what you mean by this? In particular the "didn't update inside the path" part?

This part wouldn't update after changing the username of my computer:
Screen Shot 2020-07-01 at 9 37 06 AM
Screen Shot 2020-07-01 at 9 40 00 AM

> Regardless I tested this with my Swedish fake user and the path was inserted correctly. 

👍

niik:
> This part wouldn't update after changing the username of my computer:

Ah, no I imagine it wouldn't. I should have been more specific in that you would need to create another user account for that test. But no biggie since I already had one laying around!

niik:
> Regarding the following comment, will `registry-js` also fail when there's such ban or will it work fine?

It will work fine, that was the main motivation behind building `registry-js` IIRC (see https://github.com/desktop/desktop/pull/3492)

PR #3513, opened 2017-12-06
fallback to Gravatar for rendering avatar icons

shiftkey:
Fixes #821 

This changes up the strategy for rendering avatars in a few ways, because who knew rendering an image on screen could be so _fascinating_.

 - [x] initially render Hubot avatar while fetching an image
 - [x] if no GitHub avatar found, generate and try for a Gravatar avatar
 - [x] transform these into object URLs that the browser handles and can cache fine
 - [x] handle cache failures instead of retrying each time
 - [x] test against GHE setup and confirm the fallbacks make sense
   - [x] unauthenticated GHE requests will redirect and return something like `text/html` - handle this and skip it
   - [x] blacklist GHE avatars for now so we don't `HTTP 429` ourselves
   - [x] commit form user should change to use gravatar, instead of falling back to Hubot
   - [x] accounts list should render _something_ for the avatar
 - [x] test against non-GitHub repository
   - [x] don't show Hubot, do show Gravatars

PR #10112, opened 2020-06-30
Render avatars without resolving to a GitHubUser

niik:
~~🌵 Based on #10092, please review that first 🌵~~

## Description

This is another extraction out of my hack day project which can stand on its own in order to make reviewing a little bit easier.

If you've ever looked at the network tab when GitHub Desktop loads you'll have seen entries like these:

image

Those first three requests are to fetch an avatar (in this case my avatar), and the subsequent three `blob://` urls are the result of us taking the response of those requests and creating [object URLs](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) from them.

https://github.com/desktop/desktop/blob/d6f7518f43bd7f9f55e56737e2845474fe2c7ebb/app/src/ui/lib/avatar-in-memory-cache.ts#L30-L37

This is done from the `InMemoryAvatarCache` which has a number of flaws.

1. It calls createObjectURL but never [revokeObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL) which means that any time an avatar gets loaded it'll stay in memory for the lifetime of the app (and to make matters worse we might store up to five copies of the same image due to issue number two)
1. It doesn't handle concurrent operations. While it does have a cache that gets reused it will issue new fetches for each method call to [fetchAndCache](https://github.com/desktop/desktop/blob/d6f7518f43bd7f9f55e56737e2845474fe2c7ebb/app/src/ui/lib/avatar-in-memory-cache.ts#L25) until the fetch promise has completed for the given URL.

### Why do we have an InMemoryCache?

Great question. We absolutely don't need one, it was a misguided attempt at providing a way for us to load avatars from GitHub Enterprise Server where all avatars are hidden behind what's called "private mode" (i.e calling the avatar endpoint requires a browser session). The thought was (I believe) that by using `fetch` as opposed to an image tag we could inject custom headers that would let us bypass private mode. When it became apparent that we weren't going to be able to do so (because the work on GHES to open that endpoint up) stalled we never revisited our implementation and instead implemented a layer higher up which replaced all GHES avatar urls with gravatar URLs (https://github.com/desktop/desktop/pull/3513).

### What are we going to do instead

We're gonna use `` tags! I've reworked the `Avatar` component to [generate a list of fallback avatar URLs](https://github.com/desktop/desktop/blob/c968adfef5c8eec05c0f91a3ad65378b8369f395/app/src/ui/lib/avatar.tsx#L53) based on an `IAvatarUser` instance. We'll try each of those URLs out by setting them as the `src` attribute on an `img` tag and listening to the `onError` event. If the image fails to load we'll try the next avatar URL in line. If all avatars fail to load (like if you don't have an internet connection and are faced with a previously uncached, by chromium, avatar) we'll fall back to a default image just like we did before.

### Speaking of fallback images

Unlike what we did before though we're not gonna use the hubot image as the fallback any more

image

I suspect very few of our users know what hubot is anyway so we're gonna switch that to a more generic silhouette.

image

This being an SVG image it has the added benefit of working nicely in both themes

image

image

### What's this about removing the `getUsersForRepository` stuff all about?

Prior to this PR whenever we needed to display an avatar for a commit we would attempt to resolve the commit email to a GitHubUser in order to use its `avatar_url` property. This meant that we were holding user objects from the API in order to always be able to resolve an email to a user. This lead to us carrying around a giant map of emails to `IGitHubUser` in memory for each repository in perpetuity. Since this was literally the only use case for it I've removed it in favor of requesting an avatar based on the email address itself. My actual hack day project which I've been working my way towards in #10092, #10084 and finally this PR will be removing a lot more thanks to this rather simple change.

### Special case handling of stealth email addresses

If you've been paying attention you would have noticed that commits made by @billygriffin through the web flow didn't show his usual friendly face in the history view

image

This is because Billy has configured his GitHub.com account to use a stealth email address

```
Author: Billy Griffin <5091167+billygriffin@users.noreply.github.com>
```

Unfortunately we can't use that email address to look up the avatar since it's a "fake" email. But we can look up an avatar based on either the user id or the user login and as luck would have it we have either both or at least the login available to us in the email address so we'll detect this special format and then create an avatar URL based on either the id or the login. By doing so we can all see Billy's lovely face again

image

## Release notes

Notes: [Fixed] Show correct avatar for commits made using anonymous GitHub.com emails


app/src/models/avatar.ts
```diff
@@ -13,46 +10,30 @@ export interface IAvatarUser {
   readonly email: string
 
   /** The user's avatar URL. */
-  readonly avatarURL: string
+  readonly avatarURL: string | undefined
```
rafeca, line 13: I'm wondering if after these changes we even need the `avatarURL` property... Or even the whole `IAvatarUser` object and instead get the emails on the `<Avatar />` component from from the `CommitIdentity`/`GitAuthor` objects.

From what I see we're only populating `avatarURL` when rendering the avatar on the account settings, so we could probably remove that.

If in the future we need to display avatars for GitHub accounts (e.g when showing PRs), we could have a specific `Avatar` component that just uses the `avatarURL` property that comes from the API.

Anyways, this can/should be done in a separate PR 😃

app/src/ui/lib/avatar.tsx
```diff
@@ -15,74 +15,134 @@ interface IAvatarProps {
...
+const StealthEmailRegexp = /^(?:(\d+)\+)?(.+?)@users\.noreply\.github\.com$/i
+
+/**
+ * Produces an ordered iterable of avatar urls to attempt to load for the
+ * given user.
+ */
+function* getAvatarUrlCandidates(
+  user: IAvatarUser | undefined,
+  size = 64
+): Iterable<string> {
```
rafeca, line 60: I'm not a big fan of using a generator here: since we're anyways always converting the result of `getAvatarUrlCandidates()` into an array we don't get the main benefit of generators which is laziness execution, and as a tradeoff we use a non-common language feature that may be harder to understand to some future readers (in this case I personally find this harder to understand than creating an empty array and pushing to it instead of yielding).

Is there any other reason I'm missing to use a generator here?

app/src/ui/lib/avatar.tsx
```diff
@@ -15,74 +15,134 @@ interface IAvatarProps {
...
-  public async componentWillReceiveProps(nextProps: IAvatarProps) {
-    const dataUrl = await fetchAvatarUrl(DefaultAvatarURL, nextProps.user)
+  public constructor(props: IAvatarProps) {
+    super(props)
 
-    // https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html
-    // We're basically doing isMounted here. Let's look at better ways
-    // in the future
-    if (!this.cancelFetchingAvatar) {
-      this.setState({ dataUrl })
```
rafeca, line 78: image

app/src/ui/lib/avatar.tsx
```diff
@@ -105,19 +165,38 @@ export class Avatar extends React.Component<IAvatarProps, IAvatarState> {
     return 'Unknown user'
   }
 
+  private onImageError = (e: React.SyntheticEvent<HTMLImageElement>) => {
+    if (this.state.candidates.length > 0) {
+      this.setState({ candidates: this.state.candidates.slice(1) })
+    }
+  }
```
rafeca, line 172: I really 💟 this approach, thanks for cleaning all this!!

app/src/ui/lib/avatar.tsx
```diff
@@ -15,74 +15,134 @@ interface IAvatarProps {
...
     }
-
-    return false
   }
 
-  public async componentWillMount() {
-    const dataUrl = await fetchAvatarUrl(DefaultAvatarURL, this.props.user)
+  // The /u/e endpoint above falls back to gravatar (proxied)
+  // so we don't have to add gravatar to the fallback.
+  yield `${avatarEndpoint}/u/e?email=${encodeURIComponent(email)}&s=${size}`
```
rafeca, line 119: I think we should still add the gravatar fallback for non-github or local repositories (in the case that our endpoint does not proxy to gravatar for non-existant users, which may be quite probable), what do you think?

(I think ideally we should not do the call to our endpoint on non-github repositories but keep it for local repositories, but that may be hard to do without access to the whole `Repository` object).

app/src/ui/lib/avatar.tsx
niik: > Is there any other reason I'm missing to use a generator here?

No, I just really like to yield. Iterators/generators were a huge part of the classic GitHub Desktop codebase and I was happy that ECMAScript added them to the core language.

I agree that it could look intimidating to readers of the code base though and while I think we're going to be using generators (and async generators) increasingly in the future this particular hill is not one I'm remotely willing to die on so I'll convert it to an array.

app/src/models/avatar.ts
niik: You're right, functionality-wise we don't need it, no. The reason I kept it is that if we've got it then it's a good source of the canonical URL to a user's avatar as opposed to our (albeit well educated) guesses so why not use it?

> Anyways, this can/should be done in a separate PR 😃

:+1: Let's discuss this elsewhere, I was going back and forth on this.

app/src/ui/lib/avatar.tsx
niik: Given that GitHub's avatar endpoint will always fall back to gravatar (and if that fails it'll return a generic identicon) I think adding one more fallback is only gonna add to the loading time before we show the no-avatar-avatar.

> (I think ideally we should not do the call to our endpoint on non-github repositories but keep it for local repositories, but that may be hard to do without access to the whole `Repository` object).

I agree with that reasoning but just as you say, it's hard to do without access to the list of remotes.

rafeca: Cool, let's leave it like this then!

PR #10119, opened 2020-07-01
Repository indicator improvements

niik:
## Description

This is a work-in-progress PR with the intent of reviewing our current logic for repository indicators, improving upon it where necessary, and ultimately making the entire feature itself optional as per our Q1/H1 initiatives

In additional to general cleanup this PR does the following

- Extracts all scheduling logic for refreshing repository sidebar indicators to `RepositoryIndicatorUpdater`
- Ensures that we don't refresh repository sidebar indicators when the app is in the background
- Gets rid of the simplistic `setInterval` in favor of a proper scheduler which can be paused and which will only schedule its next run after having refreshed all indicators (preventing the possibility of multiple concurrent refreshes if a refresh takes longer than the interval)
- Defers loading of sidebar indicators on initial launch until 2 minutes post launch or the first time the user  opens the repository sidebar
- Prevents redundant double fetches when refreshing repository indicators by fixing a bug in inferLastPushForRepository
- Removes inferLastPushForRepository's misplaced responsibility to ensure that we have an up-to-date `lastFetched` date for a repository


app/src/lib/infer-last-push-for-repository.ts
```diff
@@ -46,7 +46,7 @@ export async function inferLastPushForRepository(
     }
   }
 
-  if (repository.gitHubRepository !== null) {
+  if (lastPushDate === null && repository.gitHubRepository !== null) {
```
niik, line 49: This change makes little sense when viewed like this but if you expand the diff upwards you'll see that this is a local variable that's set when we've fetched the repository associated with the current remote. In other words not having this check is what's been causing us to do two requests for the same repository in rapid succession when refreshing indicators

image

It doesn't have any measurable effect on performance since the previous request will be cached and but it's nevertheless redundant and fills our network log with unnecessary requests.

niik: Worked around this using early returns instead

app/src/lib/stores/helpers/background-fetcher.ts
```diff
@@ -90,7 +92,12 @@ export class BackgroundFetcher {
       return
     }
 
-    const shouldFetch = this.shouldPerformFetch(this.repository)
+    const shouldFetch = await this.shouldPerformFetch(this.repository)
+
+    if (this.stopped) {
```
niik, line 97: Had to add the because of the newly introduced async call above. We could have been stopped in between calling that method and it returning.

rafeca: Nice catch!

app/src/lib/stores/helpers/repository-indicator-updater.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      this.paused = true
+    }
+  }
+
+  public resume() {
+    if (this.paused) {
+      log.debug('[RepositoryIndicatorUpdater] Resuming')
+
+      this.pausePromise = Promise.resolve()
```
rafeca, line 159: Why do we need to reassign `pausePromise` here to a resolved promise if we're already resolving the existing `pausePromise` promise?

rafeca (dismissed):
LGTM!

only thing thing to note is that this will cause the repository updater to run when focusing the application after long periods of inactivity, but I don't think that' an issue since the updater should be quite lightweight in terms of CPU usage (it's actually a nice feature, since we then make sure that we update the repos when the user actually wants to use Desktop and not during inactivity periods).

app/src/lib/stores/helpers/repository-indicator-updater.ts
niik: I don't think we do, I believe this was a leftover from a previous attempt of mine. Nice catch!

niik:
> only thing thing to note is that this will cause the repository updater to run when focusing the application after long periods of inactivity, but I don't think that' an issue since the updater should be quite lightweight in terms of CPU usage (it's actually a nice feature, since we then make sure that we update the repos when the user actually wants to use Desktop and not during inactivity periods).

Yeah, this is a tough change to make. We don't want to add much work on focus but the way we had it was pretty terrible where it would constantly fetch and call status in the background. I think this is a step in the right direction but we'll need more work here as well as an explicit opt-out (which will come in a later PR). Ideally I'd want to defer refresh on focus for a while but the logic is complex enough now that I'd rather way until we know it's a problem.

niik:
@rafeca I made one additional change which was to not refresh repository indicators for the currently selected repository as that's redundant. We already refresh the current repo on focus.

rafeca (approved):
Awesome!

David8166 (commented):
Better

PR #10120, opened 2020-07-01
Remove the users table from the GitHubUsers database

niik:
~~🌵 Based on #10112, please review that first 🌵~~

## Description

Finally, after #10112, #10092, and #10084 we get to the grand finale of my hack day project, removing the users table!

### What even is a users table?

The users table maintains a per-repository mapping of email addresses to GitHubUser objects. Think of GitHubUser objects as the information we'd get from a  user through the API, i.e. their display name, user id, login, avatar url, etc.

We used to use this for resolving an avatar url based on an email address that we'd get from a commit but as of #10112 we're able to look up an avatar based solely on an email address which removes the purpose of the users table entirely.

### Where did we even get the email to user mapping from

Through hoarding basically. If we found a commit with an email address we didn't know about we'd use the API to retrieve that commit and thereby get the user object attached to that response. If we failed doing that we would even fall back to the search API trying to find a user based on the email address. The search API is heavily rate-limited so we would frequently get rate-limited when trying to look up too many email addresses.

### Was there any other legitimate use case for the users table

Kinda sorta yes. When autocompleting co-author users we need to have a name and an email address for that user. In the case when we don't have an email address now we'll simply construct a stealth email address (i.e. `niik@users.noreply.github.com`). This has the added benefit of never accidentally using another user's real email address when they've specified in their preferences that they only want to be attributed by a stealth email address.

### Mentionable improvements

The mentionables table which is what we use to source the data for user autocompletion was previously just a mapping table requiring one scan of the mentionables table and then one primary key lookup (for lack of better terms when it comes to IndexedDB) for each hit in the users table. I've de-normalized this into a single table and incorporated better cache persistence. The result being that searching for users when autocompleting is substantially faster now.

```
before: getAutocompletionItems: 100.529052734375ms
after: getAutocompletionItems: 17.7509765625ms
```

:point_up: is for a worst-case scenario with 1000 mentionable users, a roughly 5-6x improvement.


app/src/lib/stores/github-user-store.ts
```diff
@@ -75,369 +33,82 @@ export class GitHubUserStore extends BaseStore {
...
-            user = { ...user, id: match.id }
-          } else {
-            user = { ...user, id: match.id, email: match.email }
-          }
-        }
-
-        const id = await this.database.users.put(user)
-        return this.database.users.get(id)
-      }
+      cacheEntry?.eTag
```
rafeca, line 80: yay! 🎉

app/src/lib/api.ts
```diff
@@ -950,14 +907,14 @@ export class API {
...
-    etag: string | null
+    etag: string | undefined
   ): Promise<IAPIMentionablesResponse | null> {
     // NB: this custom `Accept` is required for the `mentionables` endpoint.
     const headers: any = {
       Accept: 'application/vnd.github.jerry-maguire-preview',
     }
 
-    if (etag) {
+    if (etag !== undefined) {
```
rafeca, line 917: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FALWsTtMzfjai4%2Fgiphy.gif&token=better" data-gitmeme-token="better"></a>

app/src/lib/stores/github-user-store.ts
```diff
@@ -75,369 +33,82 @@ export class GitHubUserStore extends BaseStore {
...
-    return await this.cacheUser(user, false)
   }
 
   /** Update the mentionable users for the repository. */
   public async updateMentionables(
     repository: GitHubRepository,
     account: Account
   ): Promise<void> {
-    const api = API.fromAccount(account)
+    assertPersisted(repository, this.updateMentionables.name)
```
rafeca, line 62: That's nice! :)

PR #10127, opened 2020-07-02
Dialog now clears filters on esc and closes on double esc

### LINKED ISSUE #6369: Double ESC should be enabled for the every filter field

AndreiMaga:
Closes #6369

## Description

- <kbd>ESC</kbd> clears the filters in the dialog (only if there is something in the filters, otherwise it closes it)
- Double <kbd>ESC</kbd> closes the dialog
Also implemented the changes requested by @niik in #6531
### Screenshots
clone

## Release notes

Notes:

### DIFF
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -436,7 +436,7 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
     }
   }
 
-  private onDialogCancel = (e: Event) => {
+  private onDialogCancel = (e: Event | React.SyntheticEvent) => {
     e.preventDefault()
     this.onDismiss()
   }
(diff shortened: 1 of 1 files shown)


app/src/ui/dialog/dialog.tsx
```diff
@@ -526,19 +526,20 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
...
     }
 
     this.dialogElement = e
   }
 
-  private onKeyDown = (event: KeyboardEvent) => {
+  private onKeyDown = (event: React.KeyboardEvent) => {
     const shortcutKey = __DARWIN__ ? event.metaKey : event.ctrlKey
     if ((shortcutKey && event.key === 'w') || event.key === 'Escape') {
+      if (event.defaultPrevented) {
```
niik, line 540: Could we move this to the top of the method such that we return early if the event has already been handled? While it doesn't affect the logic at all it would follow the pattern we use elsewhere in the app

https://github.com/desktop/desktop/blob/21529147666b7beb6102594c3167ccea024b9c9c/app/src/ui/app.tsx#L901-L904

https://github.com/desktop/desktop/blob/21529147666b7beb6102594c3167ccea024b9c9c/app/src/ui/repository.tsx#L474-L477

etc

AndreiMaga: Sorry I didn't catch that sooner, already done in 69dbe27.

app/src/ui/dialog/dialog.tsx
```diff
@@ -526,17 +526,18 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
     if (!e) {
       if (this.dialogElement) {
         this.dialogElement.removeEventListener('cancel', this.onDialogCancel)
-        this.dialogElement.removeEventListener('keydown', this.onKeyDown)
       }
     } else {
       e.addEventListener('cancel', this.onDialogCancel)
-      e.addEventListener('keydown', this.onKeyDown)
```
niik, line 533: So happy to get rid of this!

AndreiMaga: I bet, too bad cancel and others are nowhere near that easily done

PR #10141, opened 2020-07-06
Reuse existing search dialogs in text diffs

### LINKED ISSUE #9563: Both the Desktop Search and Code Mirror Default Search Show
### Describe the bug

Both the Desktop Search and Code Mirror Default Search show in app.

### Version & OS

2.4.2-beta1, macOS 10.14.6

1. Go to changed files tab.
2. Hit `CmdOrControl+F` three times in a row
3. See both the Desktop and Code Mirror Default Search

### Screenshots

GitHubSearch

### Additional context

Should we have our own custom search box at all? Maybe we should just use the built in search.

niik:
Closes #9563

## Description

Turns out the CodeMirror search extension doesn't guarantee that there will only be one dialog open at a time, instead it relies on the `onBlur` event to close the existing dialog when a new one opens. Unfortunately for us this doesn't work when invoked through the app menu `find-text` event.

So to work around that we'll ensure that we don't open a new search dialog if we can find an existing one in the codemirror instance.

Also includes some small cleanup items.

## Release notes

Notes: [Fixed] Never show more than one search field at a time in text diffs

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -198,7 +198,7 @@ function showSearch(cm: Editor) {
   ) {
     searchLabel.style.display = 'none'
     searchField.placeholder = 'Search'
-    searchField.style.width = null!
+    delete searchField.style.width
   }
 }
 
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -170,7 +170,7 @@ function showSearch(cm: Editor) {
   // focus there instead of opening another dialog since CodeMirror
   // doesn't auto-close dialogs when opening a new one.
   const existingSearchField = wrapper.querySelector(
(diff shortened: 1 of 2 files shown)


app/src/ui/diff/text-diff.tsx
```diff
@@ -164,30 +164,36 @@ interface ITextDiffProps {
...
 
-  if (
-    searchLabel instanceof HTMLElement &&
-    searchField instanceof HTMLInputElement
-  ) {
-    searchLabel.style.display = 'none'
+  if (searchField instanceof HTMLInputElement) {
     searchField.placeholder = 'Search'
-    searchField.style.width = null!
+    searchField.style.removeProperty('width')
```
rafeca, line 196: yay! thanks for fixing this!

ed126:
niik merged commit 7c41224 into development 5 days ago

PR #10142, opened 2020-07-06
Don't wrap PathText in Monospaced component

niik:
## Description

Stumbled upon this in my DevTools console just now

Warning validateDOMNesting( ) div cannot appear as a descendant of

Turns out this had become quite the widespread pattern in the app so I went through the instances I could find and removed them. After I'd done that there was only two spots left where we used the component and given that I'm not a big fan of mandating such specific design properties through the component structure I figured I might as well just remove it entirely.

Along the way I found some oddities (like the repeated `ul` in the oversized files warning dialog) so I've attempted to ensure that the lists in the various dialogs match up.

## Screenshots

image

image

image

image

image

image

image

### DIFF
--- a/app/src/ui/changes/oversized-files-warning.tsx
+++ b/app/src/ui/changes/oversized-files-warning.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react'
 import { Dialog, DialogContent, DialogFooter } from '../dialog'
 import { LinkButton } from '../lib/link-button'
-import { Monospaced } from '../lib/monospaced'
 import { PathText } from '../lib/path-text'
 import { Dispatcher } from '../dispatcher'
 import { Repository } from '../../models/repository'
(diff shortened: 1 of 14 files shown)


app/src/ui/changes/oversized-files-warning.tsx
```diff
@@ -64,13 +63,13 @@ export class OversizedFiles extends React.Component<IOversizedFilesProps> {
   private renderFileList() {
     return (
       <div className="files-list">
-        {this.props.oversizedFiles.map(fileName => (
-          <ul key={fileName}>
```
rafeca, line 68: 🙈

app/src/ui/discard-changes/discard-selection-dialog.tsx
```diff
@@ -80,12 +79,13 @@ export class DiscardSelection extends React.Component<
         type="warning"
       >
         <DialogContent>
-          <p>
-            Are you sure you want to discard the selected changes to
-            <Monospaced>
+          <p>Are you sure you want to discard the selected changes to</p>
```
rafeca, line 82: Just added a colon to match the other discard changes dialog:

image


```suggestion
          <p>Are you sure you want to discard the selected changes to:</p>
```

PR #10160, opened 2020-07-07
Release 2.5.4-beta1

niik:
## Description

Looking for the PR for the upcoming v2.5.4-beta1 beta release? Well you've just found it, congratulations! :tada:

## Release checklist

Release checklist:

- [x] ~~Check to see if there are any errors in Sentry that have only occurred since the last production release~~


niik:
This has shipped!

niik:
That was not the button I was intending to press 🤦

PR #10168, opened 2020-07-08
Autocompletion tweaks, consistency between providers

niik:
🌵 Depends upon #10121, please review that first 🌵 

## Description

Continuation of #10121. Attempts to provide a consistent logic between issues, user, and emoji autocompletion providers. Introduces an exact replica of the cache implemented for the user autocompletion in #10121.

### DIFF
--- a/app/src/lib/stores/github-user-store.ts
+++ b/app/src/lib/stores/github-user-store.ts
@@ -116,10 +116,7 @@ export class GitHubUserStore extends BaseStore {
       response.etag
     )
 
-    if (
-      this.queryCache !== null &&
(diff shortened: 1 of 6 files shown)


app/src/lib/stores/github-user-store.ts
```diff
@@ -185,6 +182,7 @@ export class GitHubUserStore extends BaseStore {
       .sort(
         (x, y) => compare(x.ix, y.ix) || compare(x.user.login, y.user.login)
       )
+      .slice(0, maxHits)
```
rafeca, line 185: yay! I think this is way better than the previous logic, which could discard items that would have appeared at the top of the list 😀

PR #10182, opened 2020-07-10
Configure TypeScript to add an interoperability layer between ES6 imports and commonjs modules

rafeca:
This PR configures TypeScript compiler with the [`esModuleInterop` option](https://www.typescriptlang.org/docs/handbook/compiler-options.html) to add a small layer of interoperability to allow using ES6 imports to import commonJS modules.

## Context

Desktop currently have two different build pipelines: the "WebPack pipeline" (used to compile all user-facing code) and the "Node pipeline" (used to compile the development tooling: scripts that get run via `yarn` and tests).

These two pipelines share some logic (the compilation step, which is performed by TypeScript) but differ on the bundling step (the WebPack pipeline bundles the code using WebPack, while the Node pipeline doesn't bundle any code and relies on the Node.js require implementation).

Because of this, there's some discrepancy between the code that gets executed on the unit tests VS the code that gets executed on the actual app.

While some discrepancy here is normally fine (unit tests are not meant to replicate 1:1 the actual application), our current setup currently causes big differences that make some code work fine on the app but fail when executed on a test. An example of such code is:

```ts
// memoize-one is an npm package that exposes both a commonJS and ESM modules
import memoize from 'memoize-one';

memoize(xxx);
```

To understand why this code causes an issue, let's look at the two build pipelines, starting by the compilation process.

### Compilation process

Our build pipeline is currently configured in a way that TypeScript "downgrades" our ESM imports/exports to the commonJS equivalents (this is done via the `"module": "commonjs"` compiler option).

The above code, after being compiled by TypeScript looks like:

```js
const memoize_1 = require("memoize-one");

memoize_1.default(() => {});
```

Here, TypeScript does the straightforward/naive thing: since we used a default import, it considers `memoize-one` to be an ESM module that has get downgraded to commonJS. When that's the case, its exports contents will look like:

```ts
module.exports = {
  __esModule: true /** indicates that this is an ESM module **/
  default: memoizeOneSource /** the actual code of memoizeOne **/
}
```

And everything will work correctly.

### Bundling process -WebPack

When WebPack runs, it bundles these commonJS modules together and adds the needed sugar (via its own require implementation) to make the requires work well together.

Part of the bundling process deals with resolving files. Here, WebPack differs a bit with Node.js and when resolving `npm` packages it'll try to use the ESM version of a package if it exists ([more info](https://webpack.js.org/concepts/module-resolution/)).

In the code above, WebPack will use the ESM version of `memoize-one` since this package has a published [ESM version](https://unpkg.com/memoize-one@5.1.1/dist/memoize-one.esm.js) which is referenced by the `"module"` attribute of its [`package.json`](https://unpkg.com/memoize-one@5.1.1/package.json).

This file will then get downgraded to commonJS by WebPack (WebPack has also this capability), and their exports will look like what TypeScript was expecting:

```js
module.exports = {
  __esModule: true /** indicates that this is an ESM module **/
  default: memoizeOneSource /** the actual code of memoizeOne **/
}
```

This will make the code work fine in the application.

### Bundling process -Node.js

When running the compiled code on Node.js, though, things will change. In this case, Node.js will resolve `memoize-one` using the `"main"` field of its `package.json`, which returns the [commonJS version](https://unpkg.com/memoize-one@5.1.1/dist/memoize-one.cjs.js) of that library.

In that case, the exports look like (this is because `npm` package authors expect people to use `require()` calls to consume the commonJS version of their packages):

```js
module.exports = memoizeOneSource;
```

This will cause a TypeError, since `memoize_1.default` is not a function.

## Current solution

In order to fix this, I've enabled the `esModuleInterop` compiler option on the TypeScript config. This option is meant to be used exactly for this: to be able to use `import` statements against commonJS modules in a transparent way. This will also change slightly the compiler output (and its type system rules) to add some helpers that allow `import` statements to play nice with commonJS modules.

In the previous example, the compiler will generate the following code that works fine when bundled both with WebPack and Node.js:

```js
const {__importDefault} = require('tslib')
const memoize_1 = __importDefault(require("memoize-one"));

memoize_1.default(() => {});
```

The implementation of `__importDefault` is quite trivial, it just detects whether the imported file is an ES6 module and uses its default exported key in that case:

```js
function __importDefault(mod) {
    return (mod && mod.__esModule) ? mod : { default: mod };
}
```

## Alternative solutions

The most straightforward way to avoid this problem would have been to change our codebase to not use imports and fallback to `require` calls when importing commonJS modules. This unfortunately would be insufficient for modules like `memoize-one` which export both a commonJS and ESM modules, so we would have to deal with them (we could configure WebPack to always resolve to commonJS versions of packages, but that's not ideal since we lose some optimizations that can be done on ESM modules thanks to their import/exports being statically analyzable).

Another potential solution to this problem could have been to only change the TypeScript configuration to add the interop layer for tests and scripts that run on Node.js. While this would have fixed the problem above, it would create different problems on other scenarios.

* First problem would have been present when default importing a commonJS module: in this case the tests would now pass thanks to the interop layer but the application code would fail since we wouldn't have an interop layer there.
* Second problem would have been present when trying to call directly the object returned by "import star" on a commonJS module:
  ```ts
  import * as classnames from 'classnames'

  classnames('x')
  ```

  This code (which is invalid based on the ESM standard) would work on the application (TypeScript without the interop layer allows this since it just puts the output of `module.exports` in `classnames`) but it would fail in the tests (With the interop layer, that code becomes invalid since `classnames` is always going to be an object with a `default` key that would contain the contents of `module.exports`) 

Because of this, I've opted to add the interop layer on all our TypeScript configurations, so we have more consistency in behaviour between tests and application code.

## Tradeoffs of this change

Since this change affects our compilation process, it has some impact on our production bundles.

First, it have some small effect on the size of our JavaScript bundles. It affects mostly the main and the renderer process which get ~8KB bigger each one:

**Renderer process:**
* Before: 1783116 bytes
* After: 1792845 bytes

**Main process:**
* Before: 170331 bytes
* After: 178165 bytes

The second change has to do with how modules get required. When using either an "import default" or "import star" statement, an additional function call will be made:

```diff
-const chalk_1 = __webpack_require__(/*! chalk */ "./app/node_modules/chalk/index.js");
-const Path = __webpack_require__(/*! path */ "path");
+const tslib_1 = __webpack_require__(/*! tslib */ "./app/node_modules/tslib/tslib.es6.js");
+const chalk_1 = tslib_1.__importDefault(__webpack_require__(/*! chalk */ "./app/node_modules/chalk/index.js"));
+const Path = tslib_1.__importStar(__webpack_require__(/*! path */ "path"));
```

An additional function call per se is not going to have a big impact in our startup time, but I'm slightly more worried about what both `__importDefault` and `__importStar` do behind the scenes: both of these functions create a new object every time we import a module (even if we import the same module twice):

https://github.com/microsoft/tslib/blob/23642901c0cfc63dff8da27c4c8c71acd265445a/tslib.es6.js#L202-L212

Good thing is that first we don't use default imports/import star very often (Desktop modules always use named exports, so we use that mostly for third party dependencies) and second we don't inline our import calls, which means that each import call is only going to be executed once and we won't allocate a ton of objects in hot paths (e.g we can't have an `import` call in a for loop).

We currently have 1199 import star calls and 60 import default calls, so this means that this change will allocate ~1200 more objects during startup. I haven't measured the impact of this but my hypothesis is that it's not going to be appreciable in terms of startup time.

## Next steps

In order to revert the negative impact of this change, there's one clear next step (which I prefer to tackle separately from this PR but give it high priority):

We should change our TypeScript configuration to not "downgrade" ESM imports to commonJS requires, and leave that responsibility to WebPack. This will not only get rid of the increased bundle size caused by this PR and the intermediate interop methods created by TypeScript, but also unblock [Tree Shaking](https://webpack.js.org/guides/tree-shaking/) optimizations that WebPack can perform thanks to being able to statically analyze ES6 imports.

Unfortunately, this change required some non-trivial changes (and some more changes in some of our modules), so I prefer to tackle it separately to not make this PR even bigger.


script/licenses/license-overrides.ts
```diff
@@ -44,4 +44,23 @@ COPYRIGHT HOLDERS WILL NOT BE LIABLE FOR ANY DIRECT, INDIRECT, SPECIAL OR CONSEQ
 
 The name and trademarks of copyright holders may NOT be used in advertising or publicity pertaining to the work without specific, written prior permission. Title to copyright in this work will at all times remain with copyright holders.`,
   },
+
+  'tslib@2.0.0': {
+    repository: 'git+https://github.com/microsoft/tslib',
+    license: 'BSD',
```
rafeca, line 50: I had to add this because `tslib` has a 0-BSD license, but not sure (see https://github.com/microsoft/tslib/blob/master/LICENSE.txt), but not sure if we should instead add 0-BSD to the [`legal-eagle` package](https://github.com/atom/legal-eagle)

app/src/lib/git/rebase.ts
```diff
@@ -2,7 +2,7 @@ import * as Path from 'path'
 import { ChildProcess } from 'child_process'
 import * as FSE from 'fs-extra'
 import { GitError } from 'dugite'
-import * as byline from 'byline'
+import byline from 'byline'
```
rafeca, line 5: I've had to change all the `import *` statements that were invalid based on the ESM standard (which mandates that the output of import star is an object with the exported values of the object).

Good thing is that with the interop flag, TypeScript now complains about this 😃

Issue #10187, opened 2020-07-11
Cannot "show in explorer" in some projects


khiav223577:
### Describe the bug

The "show in explorer" command raise errors in some of my projects while "open in Command Prompt" and "open in Sublime Text" works.

It will work fine if I rename the folder from `echo` to `echo123`, but it will raise the same errors after I rename it back to original name.

### Version & OS

Version 2.5.3
Windows 10 Build 19041.329

### Steps to reproduce the behavior

![image](https://user-images.githubusercontent.com/4011729/87220527-09327080-c397-11ea-8029-5d61e5ad3023.png)

### Expected behavior

Open the exploror

### Actual behavior

Raise an error dialog. See the screenshot below.

### Screenshots

![image](https://user-images.githubusercontent.com/4011729/87220437-48ac8d00-c396-11ea-8378-0acf6c9ccb16.png)

### Logs

[2020-07-11.desktop.production.log](https://github.com/desktop/desktop/files/4906521/2020-07-11.desktop.production.log)

rafeca:
Hey @khiav223577!

I've been able to reproduce the problem in v2.5.3 by doing the following:

1. Create a random folder (name doesn't matter).
2. Copy any Windows executable to that folder (e.g `c:/windows/system32/cmd.exe`).
3. Create a repository with the same name as the executable.

After these steps, I get the following error when trying to "Show in Explorer":

![image](https://user-images.githubusercontent.com/408035/87295582-baefaf80-c505-11ea-984c-f8deb6b4c96b.png)

@khiav223577 do you have by any chance any `echo.exe` file (or other similar files) inside your `回音 git 版/` folder?

I'm going to check whether https://github.com/desktop/desktop/pull/9987 fixes this problem, since it refactors a bit that same logic.

rafeca:
> I'm going to check whether #9987 fixes this problem, since it refactors a bit that same logic.

Update: that PR doesn't solve this issue so we need to look for another solution to this.

khiav223577:
> @khiav223577 do you have by any chance any `echo.exe` file (or other similar files) inside your `回音 git 版/` folder?

It does have. And the bug only happens when there is `echo.exe` inside `回音 git 版/` folder. :fearful:

rafeca:
Thanks for the information @khiav223577 !

[...]

PR #10195, opened 2020-07-13
Add backslash when trying to open folders in Windows

rafeca:
Closes https://github.com/desktop/desktop/issues/10187

This PR depends on https://github.com/desktop/desktop/pull/9987, to avoid conflicts when merging since the open folder logic has been refactored in https://github.com/desktop/desktop/pull/9987.

## Description

This PR adds a final slash to the path used in Windows to open a folder to prevent Windows from opening unintended files that have the same name as the folder.

## Context

In Windows, if there's a file and a directory with the same name, like:

```
C:\MyFolder\foo        # Directory
C:\MyFolder\foo.exe    # Random file
```

Then electron's `shell. openItem()` API will try to open `foo.exe` when called with `C:\MyFolder\foo`. 

I haven't doublechecked (not familiar enough with the Windows APIs) but it seems like this is caused by Windows itself, since in that same situation, running `start C:\MyFolder\foo` from `cmd` will cause `foo.exe` to get executed. 

The workaround to make sure that the folder is always open is to add a trailing slash, which seems to make Windows happy (both from Electron's API and from the cmd).

## Release notes

Notes: [Fixed] Fix "Show in Explorer" in Windows when there are other files with the same name as the project folder.


app/src/main-process/shell.ts
```diff
@@ -27,6 +28,15 @@ export function UNSAFE_openDirectory(path: string) {
...
   } else {
-    shell.openItem(path)
+    // Add a trailing slash to the directory path.
+    //
+    // In Windows, if there's a file and a directory with the
+    // same name (e.g `C:\MyFolder\foo` and `C:\MyFolder\foo.exe`),
+    // when executing shell.openItem(`C:\MyFolder\foo`) then the EXE file
+    // will get opened.
+    // We can avoid this by adding a final backslash at the end of the path.
+    const pathname = path[path.length - 1] !== Path.sep ? path + Path.sep : path
```
niik, line 38: I wonder if we should future-proof this by ensuring we only add a trailing backslash on Windows in case we end up supporting more platforms. Perhaps something like this
```suggestion
    const pathname = __WIN32__ && !path.endsWith('\\') ? `${path}\\` : path
```

app/src/main-process/shell.ts
```diff
@@ -27,6 +28,15 @@ export function UNSAFE_openDirectory(path: string) {
       .openExternal(directoryURL)
       .catch(err => log.error(`Failed to open directory (${path})`, err))
   } else {
-    shell.openItem(path)
+    // Add a trailing slash to the directory path.
+    //
+    // In Windows, if there's a file and a directory with the
```
niik, line 33:
```suggestion
    // On Windows, if there's a file and a directory with the
```

app/src/main-process/shell.ts
```diff
@@ -27,6 +28,15 @@ export function UNSAFE_openDirectory(path: string) {
...
+    // same name (e.g `C:\MyFolder\foo` and `C:\MyFolder\foo.exe`),
+    // when executing shell.openItem(`C:\MyFolder\foo`) then the EXE file
+    // will get opened.
+    // We can avoid this by adding a final backslash at the end of the path.
```
niik, line 37: Love this explanatory comment!

app/src/main-process/shell.ts
rafeca: Good point! 👍

Issue #10202, opened 2020-07-14
Improvements to the build pipeline


rafeca:
This issue tracks the improvements made to the Desktop build pipeline to reduce the bundle size of the production assets and the startup time of the Desktop app.

### Add interop layer to import commonJS modules

* **Impact**: This will fix the issues we currently have on unit tests where the code imports a commonJS dependency (see [this comment](https://github.com/desktop/desktop/pull/10053#discussion_r450302353)) and allows to remove the nasty hacks that mix imports with requires ([example](https://github.com/desktop/desktop/blob/development/script/generate-octicons.ts#L13)).
* **PR**: https://github.com/desktop/desktop/pull/10182

### Enable Tree shaking

* **Impact**: This will reduce the bundle size and improve the startup time
* **PR**: https://github.com/desktop/desktop/pull/10185

### Use terser as the minifier

* **Impact**: This will reduce build times, reduce the bundle size and improve the startup time
* **PR**: https://github.com/desktop/desktop/pull/10196

niik:
I believe all the core pieces of this has been merged now, right @rafeca?

rafeca:
> I believe all the core pieces of this has been merged now, right @rafeca?

Yup, https://github.com/desktop/desktop/pull/10244 is just a nice improvement

PR #10204, opened 2020-07-14
Replace require calls by import calls

rafeca:
Related issue: https://github.com/desktop/desktop/issues/10202

~~⚠️  *This PR is based on https://github.com/desktop/desktop/pull/10196, please review that one first*~~

## Description

This PR cleans a bit the codebase by replacing all `require` calls that are in the top level scope of a module and are static.

As part of this process, I've added TS type definitions to the modules that were missing them (for some of them I could install the type definition package from `npm`, and for other I had to upgrade the library since the types were included in a newer version of that library.

This PR doesn't have any substantial user-facing value, but provides some devX benefits for us:

- Now we have a consistent way of importing files (we should always use `import` statements).
- Now we have better type definitions for most of our dependencies, which will make it easier to upgrade them.

## Next steps

A potential next step is to standarize how do we import external libraries, e.g `import * as moment from 'moment'` or `import moment from 'moment'`.

This, though, is a bit tricky since it depends on how each library exports their API (and whether they export commonJS modules, ES6 modules or both). I'm going to do a lightweight exploration to see if there's anything we can improve there.


app/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "app-path": "^2.2.0",
+    "app-path": "^3.2.0",
```
rafeca, line 20: There are no big changes between these two version of the library: it just uses newer JS syntax and some internal changes ([more info](https://github.com/sindresorhus/app-path/compare/v2.2.0...v3.2.0)).

app/package.json
```diff
@@ -30,7 +30,7 @@
     "dugite": "1.91.3",
     "electron-window-state": "^5.0.3",
     "event-kit": "^2.0.0",
-    "file-uri-to-path": "0.0.2",
+    "file-uri-to-path": "^2.0.0",
```
rafeca, line 33: It's hard to tell based on [the diff](https://github.com/TooTallNate/file-uri-to-path/compare/0.0.2...2.0.0) between the two versions (since the codebase got converted to TS), but based on [the diff between v0.0.2 and v1.0.0](https://github.com/TooTallNate/file-uri-to-path/compare/0.0.2...1.0.0), the changelog (that mentions that v2.0.0 only converted the codebase to TS), and [the list of commits](https://github.com/TooTallNate/file-uri-to-path/compare/1.0.0...2.0.0) between v1.0.0 and v2.0.0, I'm quite certain that the risk of this update is very low.

app/package.json
```diff
@@ -43,7 +43,7 @@
     "mri": "^1.1.0",
     "p-limit": "^2.2.0",
     "primer-support": "^4.0.0",
-    "queue": "^4.4.2",
+    "queue": "^5.0.0",
```
rafeca, line 46: Based on [the diff](https://github.com/jessetane/queue/compare/4.4.2...5.0.1) between v4.4.2 and v5.0.0, the only change that could affect the logic of this module is https://github.com/jessetane/queue/pull/52 .

Based on that, my assessment is that this upgrade is very low risk.

app/src/lib/source-map-support.ts
```diff
@@ -39,9 +38,9 @@ function retrieveSourceMap(source: string) {
         return { url: Path.basename(path), map: xhr.responseText }
       }
     } catch (error) {
-      return
+      return null
```
rafeca, line 41: This is enforced by the TS type definitions of `source-map-support` (with the `require` statement, `source-map-support` was typed as `any`).

Based on the module internal logic, it should change things when returning here `null` instead of `undefined`, since there's a loose check ([source](https://github.com/evanw/node-source-map-support/blob/v0.4.15/source-map-support.js#L168)), so this change should be safe.

app/src/lib/stores/helpers/ahead-behind-updater.ts
```diff
@@ -77,15 +56,15 @@ export class AheadBehindUpdater {
...
+        error !== null ? reject(error) : resolve(result || null)
       )
     })
   }
 
   private executeTask = (
     from: string,
     to: string,
-    callback: (error: Error | null, result: IAheadBehind | null) => void
+    callback?: (error?: Error, result?: IAheadBehind) => void
```
rafeca, line 67: The TS definitions of the `queue` package are quite low quality, since they have a few discrepancies with the actual logic (plus they don't use generics so the result is typed as `any`). These are some of the discrepancies:

- The actual `callback` function is always passed to each job ([source](https://github.com/jessetane/queue/blob/5.0.1/index.js#L143)) but the typings say that it can be `undefined`.
- The type of the `error` argument is `Error | undefined`, when the convention in Node.js is to use `null` for error arguments in callbacks ([source](https://nodejs.org/en/knowledge/errors/what-are-the-error-conventions/)). It also [internally uses `null`](https://github.com/jessetane/queue/blob/5.0.1/index.js#L146) when it throws an error itself (this doesn't matter much because the library does a [loose null comparison](https://github.com/jessetane/queue/blob/5.0.1/index.js#L106) for the error object).
- The `result` type is `Object | undefined`, while the `undefined` part makes sense for when we only throw an error, forcing a `Object` type prevents us from returning `null` (since `null` is not a subtype of `Object`).

I'm considering fixing their types and sending a PR to that project, but for now I've decided to make very small and low-risk changes to this logic (return `undefined` instead of `null`) to workaround the type issues.

rafeca: Edit: I've ended up creating 1020e07c08176ed2064d3442ab062c0d57d31749 to change a bit the logic and move the jobs to return a `Promise`. With this I can bypass some of the issues on the types of `queue` 😅

01044273055:
심3

app/src/lib/stores/helpers/ahead-behind-updater.ts
```diff
@@ -70,38 +49,33 @@ export class AheadBehindUpdater {
...
+  private executeTask = async (
     from: string,
-    to: string,
-    callback: (error: Error | null, result: IAheadBehind | null) => void
-  ) => {
+    to: string
+  ): Promise<IAheadBehind | null> => {
     if (this.comparisonCache.has(from, to)) {
-      return
+      return null
```
niik, line 64: This seems really weird (and I know it's not your fault). Shouldn't this be `return this.comparisonCache.get(from, to)` ?

```suggestion
      return this.comparisonCache.get(from, to)
```

And once we've done that is there even a need for the `executeTask` method or could we just roll up the logic into `executeAsyncTask`.

The previous logic seems super buggy, why would we just skip calling the callback?

app/src/lib/uuid.ts
```diff
@@ -1,13 +1,5 @@
 import { randomBytes as nodeCryptoGetRandomBytes } from 'crypto'
-
-// Work around synchronously seeding of random buffer in the v1
-// version of uuid by explicitly only requiring v4. As far as I'm
-// aware we cannot use an import statement here without causing webpack
-// to load the v1 version as well.
-//
-// See
-//  https://github.com/kelektiv/node-uuid/issues/189
```
niik, line 9: Was this comment removed because we're certain it's not an issue anymore?

rafeca: Oh yes, I should have added a comment! this was solved in electron some versions ago (see https://github.com/electron/electron/issues/2073), I remember suffering this same issue in Atom 😅

We could now change the import to `import { v4 as uuid } from 'uuid';`, but the version of `node-uuid` that we're using does not export ES6 modules so that would cause a bundle size regression.

We should upgrade the `node-uuid` version since newer ones export a ESM version of the library

app/src/lib/stores/helpers/ahead-behind-updater.ts
rafeca: Yes, I think returning the value of the `comparisonCache` it's much better. I thought it didn't matter much since the queue results are not directly used: there's only a not-null check on the `success` callback but if I understand the logic correctly, it won't matter if we return `null`.

> The previous logic seems super buggy, why would we just skip calling the callback?

Uhm good point! that seems like a potential serious issue 😅 

The whole logic of this module is quite weird... what I think makes it hard to reason is that the queue is only used to populate the cache object which then gets exported outside of this module.

In one of my first PRs in Desktop I had to deal with an issue with this approach and I introduced the `executeAsync` method to be able to access the aheadBehind information of a single `from/`to` combination without having to execute the whole queue (see https://github.com/desktop/desktop/pull/9319).

Issue #10200, opened 2020-07-14
Annoying error message


kevincappuccino:
If we drag 'n drop a file that is not a git repo, it displays this message:
![image](https://user-images.githubusercontent.com/56762529/87371860-9ffc5a00-c55c-11ea-90e6-6932cded03b1.png)

The issue is when accidentally we drag and drop 100 files, 'cause that message will pop up for each one of those 100 files.
### Proposed solution
Simply add a "close all (n)" option

How will it benefit Desktop and its users?
UX will be improved

rafeca:
Thanks for the feedback @kevdeveloper!

Just to confirm, did you drag these files by mistake and or there's any scenario where you would want to drag multiple files to Desktop (where some/most of them are not git repositories)?

I think fixing this should be quite easy, so I'll take a look at it.

kevdeveloper:
I dragged them by mistake ;)

rafeca:
I've just created https://github.com/desktop/desktop/pull/10211 which should only show a single modal with all the invalid paths in this situation

PR #10211, opened 2020-07-15
Show invalid path error only once when adding multiple repositories

rafeca:
Closes https://github.com/desktop/desktop/issues/10200

## Description

This PR changes the handling of errors when adding repositories to Desktop to show a single error modal instead of multiple ones when the user drags multiple invalid folders.

In order to limit the amount of items that get displayed if e.g the user drags thousands of files to Desktop, we only display the first 3 invalid paths that were invalid:

Screen Shot 2020-07-15 at 11 53 40

(would be cool to display the full list of paths when hovering over the "more" text, but unfortunately it's not easy to add rich text in error traces).

I haven't implemented any smart trimming of paths, since the wrapping doesn't look terrible and this is a quite rare error:

Screen Shot 2020-07-15 at 11 55 51

## Release notes

Notes: [Improved] Show only one error dialog when dragging multiple invalid folders to Desktop


dawg79:
Ok winner

app/src/lib/stores/app-store.ts
```diff
@@ -4933,9 +4937,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        .slice(0, MaxInvalidFoldersToDisplay)
+        .map(path => `- ${path}`)
+        .join('\n')}${
+        invalidPaths.length > MaxInvalidFoldersToDisplay
+          ? `\n\n(and ${invalidPaths.length - MaxInvalidFoldersToDisplay} more)`
+          : ''
+      }`
+
+      if (invalidPaths.length === 1) {
+        errorMessage = `${invalidPaths} isn't a git repository.`
```
niik, line 4955: Nitpicking deluxe here but could we refactor this so that we only go down the map/join path if there are multiple invalid paths? Additionally we should capitalize `Git` in `git repositories`

```suggestion
      let errorMessage

      if (invalidPaths.length === 1) {
        errorMessage = `${invalidPaths} isn't a Git repository.`
      } else {
        errorMessage = `The following paths aren't Git repositories:\n\n${invalidPaths
          .slice(0, MaxInvalidFoldersToDisplay)
          .map(path => `- ${path}`)
          .join('\n')}${
          invalidPaths.length > MaxInvalidFoldersToDisplay
            ? `\n\n(and ${
                invalidPaths.length - MaxInvalidFoldersToDisplay
              } more)`
            : ''
        }`
```

This could be make much nicer with a little helper method for producing  the error message and using early returns

rafeca: Thanks for the suggestion! I'm going to create a separate method since I agree it's going to be much nicer!

PR #10239, opened 2020-07-20
Make sure that `onPopupDismissed` only dismisses the relevant popup

rafeca:
~~⚠️ This PR is based on https://github.com/desktop/desktop/pull/10182, since without it we can import the `memoize-one` package from `app.tsx`.~~

## Description

This is a follow-up of https://github.com/desktop/desktop/pull/10053 (see [this comment](https://github.com/desktop/desktop/pull/10053#discussion_r450722148)) that memoizes the `onPopupDismissed` function and makes sure that all the other dialogs in Desktop use the memoized version of `dismissPopup`.

This will fix instances where a dialog calls `onPopupDismissed` and it ends up closing a different dialog that got opened afterwards.

While the new behaviour is definitely better than the old one, it's possible that in some instances the logic was relying on the fact that `onPopupDismissed` closed any popup, so this PR has a small risk of causing some unexpected behaviour.

Since the risk is actually quite low, and the impact of a potential issue rather small (a popup would not get automatically closed), I think it's fine to live with that risk and fix any potential unexpected behaviour afterwards if we stumble across it.


giovih:
wow

PR #10224, opened 2020-07-17
Upgrade octoicons

rafeca:
Closes https://github.com/desktop/desktop/issues/10223

## Description

This PR updates the `@primer/octoicons` package to v10.0.0, which includes the refreshed icons that have recently been shipped on GitHub.com

### Technical details

Something to take into account is that with the new icons release, icons can now have multiple SVGs depending on the size. 

This PR, though, will discard additional sizes and will only take into account the `16` size (or the first one available if `16` does not exist) for all the icons.

The reasoning behind that is that currently we don't have a way to not include the icons that are not being used in the JS bundle, so including all sizes for each icon would increase the bundle size (and anyways we're rendering the icons with a height of `16px` most of the times).

I'm planning to refactor a bit the script that generates the octoicons data to make the `app/src/ui/octicons/octicons.generated.ts` file export each icon separately, so we can take advantage of WebPack's tree shaking capabilities (once https://github.com/desktop/desktop/pull/10185 gets merged) to not include in the bundle icons or sizes that are not used in the app.

### Screenshots

image

## Release notes

Notes: [Added] Refresh the design of the app icons.

PR #10244, opened 2020-07-21
Tree shake octoicons

rafeca:
~~⚠️ This PR is based on https://github.com/desktop/desktop/pull/10196 since we need the updated version of WebPack to get Tree Shaking when using `import star`.~~

Related issue: https://github.com/desktop/desktop/issues/10202

## Description

This PR addresses the following comment in the summary of https://github.com/desktop/desktop/pull/10224:

> I'm planning to refactor a bit the script that generates the octoicons data to make the app/src/ui/octicons/octicons.generated.ts file export each icon separately, so we can take advantage of WebPack's tree shaking capabilities (once #10185 gets merged) to not include in the bundle icons or sizes that are not used in the app.

With this PR, we make sure that only the used octicons gets included in the bundle, allowing the production bundles to be smaller and therefore improve startup time and memory consumption.

In terms of bundle size, this is the impact of this PR

### Renderer process bundle

branch in https://github.com/desktop/desktop/pull/10196: 1504698 bytes
this PR: 1453814

⏬ Bundle size **3.4% smaller** (**51KiB** less)

### Crash window bundle

branch in https://github.com/desktop/desktop/pull/10196: 206558 bytes
this PR: 144982 bytes

⏬ Bundle size **30% smaller** (**61KiB** less)

*(This PR has no effect on the main process bundle since we're not including the octicons there).*

## Technical considerations

I would have loved to not have to add https://github.com/desktop/desktop/commit/585691e4d63600207f1be80e4f1b81c1655f0525 and https://github.com/desktop/desktop/commit/4e9b06deb1fea261b00e27109f9726b44dc8a0b4 in this PR to keep the list of changes in this PR much smaller (and because I think it looks nicer to import all the octicon-related stuff from a single module), but unfortunately WebPack v4 Tree Shaking capabilities have some limitations when reexporting everything from another module:

1. Doing `import * as Foo from './foo'; export {Foo}` prevents deleting unused exports from `./foo` (see https://github.com/webpack/webpack/issues/9607)
2. Doing `export * as Foo from './foo' (which is equivalent as the code above, and was added to the standard [recently](https://github.com/tc39/proposal-export-default-from)) is not even supported by WebPack v4 (it's only [been added](https://github.com/webpack/webpack/pull/11117) to WebPack v5). 

I've also had to change the octicon symbols to be standard JS objects instead of constructing our custom `OcticonSymbol` object. This is to facilitate the dead code elimination: when some code has a function call (even if it's a constructor) the minifier cannot easily know if that function could have side effects so it acts defensively and it keeps the code even if it's not used anywhere. There are ways to tells WebPack that a module is side effect free but I opted to use regular objects instead to not mess with WebPack's configuration.


niik:
There's one or two conflicting files here @rafeca 😛

Twixes:
There's a few more now LOL. So I suppose this will not go in?

sergiou87:
I just updated the branch and measured the bundle size:

### Renderer process bundle

development: 1896933 bytes
this PR: 1813571

⏬ Bundle size **4.4% smaller** (**81KiB** less)

### Crash window bundle

development: 276161 bytes
this PR: 171471 bytes

⏬ Bundle size **38% smaller** (**102KiB** less)

tidy-dev (approved):
✨ Looks like a nice improvement.

sergiou87:
Thanks for your work on this @rafeca!! I learned a lot 🤓 ❤️

PR #10260, opened 2020-07-24
Create <RadioButton/> component

rafeca:
Related issue: https://github.com/github/desktop/issues/642

## Description

This PR adds a new `<RadioButton />` component that can be used to create radio buttons with a label:

Screen Shot 2020-07-24 at 10 35 28

This new component doesn't have much logic (nor anything complex to do), but it still simplifies a couple of things that had to be done every time a radio button is used:

1. Create a unique id for the radio button to be able to link the label to it (we didn't always use a globally unique id since creating and releasing unique ids is quite tedious).
2. Pass directly the `value` of the radio button on the `onSelected` callback. Until now we were reading the value from the event and manually casting it to the desired type, now we benefit from TypeScript generics avoiding the potential bugs caused by manually casting values.

## Release notes

Notes: no-notes

### DIFF
--- a/app/styles/ui/_radio-button.scss
+++ b/app/styles/ui/_radio-button.scss
@@ -1,0 +1,12 @@
+.radio-button-component {
+  & + .radio-button-component {
+    margin-top: var(--spacing-half);
+  }
+
+  label {
+    margin: 0;
+
+    // Only add a right margin if there's a label attached to it
+    margin-left: var(--spacing-half);
+  }
+}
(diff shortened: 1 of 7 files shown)


app/styles/ui/_radio-button.scss
```diff
@@ -0,0 +1,12 @@
+.radio-button-component {
+  & + .radio-button-component {
+    margin-top: var(--spacing-half);
+  }
+
+  label {
+    margin: 0;
+
+    // Only add a right margin if there's a label attached to it
```
niik, line 9: I  think this comment is probably redundant now that it's on `label`
```suggestion
```

PR #10261, opened 2020-07-24
Create <RefNameTextBox/> component

rafeca:
Related issue: github/desktop#642

~~⚠️ This PR is based on https://github.com/desktop/desktop/pull/10260 (it does not depend on it, but since the followup PR is going to depend on both I've opted to create a linear PR dependency).~~

## Description

This PR creates a new component called `<RefNameTextBox/>` that handles the sanitization needed when asking users to enter a name for a git reference (branch name, tag name, etc).

This component does 3 things that used to be done every time we wanted to ask the user for a ref name:

1. Sanitize the user-entered value to a ref name compatible with git.
2. Display a warning if the sanitized value is different than the value that the user provided.
3. Display a different warning if the santization results in an empty string.

Screen Shot 2020-07-24 at 12 12 13

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/ref-name-text-box.tsx
+++ b/app/src/ui/lib/ref-name-text-box.tsx
@@ -80,7 +80,6 @@ export class RefNameTextBox extends React.Component<
         <TextBox
           label={this.props.label}
           value={this.state.proposedValue}
-          autoFocus={this.props.autoFocus}
           onValueChanged={this.onValueChange}
           onBlur={this.onBlur}
         />
--- a/app/src/ui/lib/ref-name-text-box.tsx
+++ b/app/src/ui/lib/ref-name-text-box.tsx
@@ -92,12 +92,11 @@ export class RefNameTextBox extends React.Component<
 
   private onValueChange = (proposedValue: string) => {
     const sanitizedValue = sanitizedRefName(proposedValue)
+    const previousSanitizedValue = this.state.sanitizedValue
 
     this.setState({ proposedValue, sanitizedValue })
 
-    // If the sanitized value didn't change we don't need to
-    // call the onValue change prop.
-    if (sanitizedValue === this.state.sanitizedValue) {
+    if (sanitizedValue === previousSanitizedValue) {
       return
     }
 
--- a/app/src/ui/lib/ref-name-text-box.tsx
+++ b/app/src/ui/lib/ref-name-text-box.tsx
@@ -60,7 +60,7 @@ export class RefNameTextBox extends React.Component<
     const proposedValue = props.initialValue || ''
 
     this.state = {
-      proposedValue: proposedValue,
+      proposedValue,
       sanitizedValue: sanitizedRefName(proposedValue),
     }
   }
--- a/app/src/ui/lib/ref-name-text-box.tsx
+++ b/app/src/ui/lib/ref-name-text-box.tsx
@@ -111,13 +111,8 @@ export class RefNameTextBox extends React.Component<
     if (this.props.onBlur !== undefined) {
       // It's possible (although rare) that we receive the onBlur
       // event before the sanitized value has been committed to the
-      // state so we need to check for that condition and sanitize
-      // the value ourselves if that's the case.
-      const sanitizedValue =
-        proposedValue === this.state.proposedValue
-          ? this.state.sanitizedValue
-          : sanitizedRefName(proposedValue)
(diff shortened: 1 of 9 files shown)


app/src/ui/lib/ref-name-text-box.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { sanitizedRefName } from '../../lib/sanitize-ref-name'
+import { TextBox } from './text-box'
+import { Octicon, OcticonSymbol } from '../octicons'
+import { Ref } from './ref'
+
+interface IRefNameProps {
+  /**
+   * The initial value for the ref name.
+   *
+   * Note that updates to this prop will be ignored.
```
niik, line 12: Gosh this irks me a bit but I'm not gonna hold up the PR for it. I can see how it'd be messy to try and have the parent component keep track of both the sanitized and the "raw" value but it sure doesn't feel idiomatic. I think I'd  feel a tad bit better if we at least reacted to changes of this prop but :shrug:

app/src/ui/lib/text-box.tsx
```diff
@@ -83,7 +83,7 @@ export interface ITextBoxProps {
   /**
    * Callback used when the component loses focus.
    */
-  readonly onBlur?: () => void
+  readonly onBlur?: (value: string) => void
```
niik, line 86: Can we add just a tiny note about what `value` is here since `value` is such a generic term?

app/src/ui/lib/ref-name-text-box.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const sanitizedValue = sanitizedRefName(proposedValue)
+
+    this.setState({
+      proposedValue,
+      sanitizedValue,
+    })
+
+    // If the sanitized value didn't change we don't need to
+    // call the onValue change prop.
+    if (sanitizedValue === this.state.sanitizedValue) {
```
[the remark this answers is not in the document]
rafeca: Fair point, I agree the intention will be more clear!

app/src/ui/lib/ref-name-text-box.tsx
```diff
@@ -0,0 +1,176 @@
...
+      // It's possible (although rare) that we receive the onBlur
+      // event before the sanitized value has been committed to the
+      // state so we need to check fot that condition and sanitize
+      // the value ourselves if that's the case.
+      const sanitizedValue =
+        proposedValue === this.state.proposedValue
+          ? this.state.sanitizedValue
+          : sanitizedRefName(proposedValue)
+
+      this.props.onBlur(sanitizedRefName(sanitizedValue))
```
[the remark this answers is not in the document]
rafeca: Yup, `sanitizedValue` is pure, so let's always do the sanitization 😃

app/src/ui/lib/ref-name-text-box.tsx
rafeca: The main problem about reacting to changes to this prop is that then we would have two sources of truth for the state: one that comes from the upwards props and another that comes from what the user types.

This could cause problems if the component gets re-rendered while/after the user has changed the input value: the props won't match the state anymore and whatever the user has typed will get lost (this whole thing is way better explained [in this article](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#common-bugs-when-using-derived-state)).

The other option would be to make this component controlled and move the state handling to its parents. As you mention this would add some additional boilerplate there and would leak the abstraction around the sanitization (this wouldn't be a big deal but I think that the current solution is slightly better).

Also, if we did that, note that we would just move this pattern to each of this component parents, making them uncontrolled (in fact, they are already uncontrolled, e.g `<CreateBranchDialog />` that accepts a `initialName` prop which gets ignored in subsequent re-renders ([source](https://github.com/desktop/desktop/blob/development/app/src/ui/create-branch/create-branch-dialog.tsx#L45)).

Anyways, if you feel strongly about it I can move the state that controls the raw value upwards, but imho this will only move the "problem" to a few places instead of having it in a single component (although is not a big change so I'm open to do it to unblock this)

niik: Yeah, I understand the tradeoffs and no, I don't feel strongly enough about it to hold up this PR. It's a pragmatic solution to a pre-existing condition.

PR #10262, opened 2020-07-24
Add setting to customize the default branch for new repositories

rafeca:
Related issue: github/desktop#642

~~⚠️ This PR depends on https://github.com/desktop/desktop/pull/10261 and https://github.com/desktop/desktop/pull/10260 since it uses the new components created in these PRs (`<RadioButton />` and `<RefNameTextBox />`).~~

## Description

This PR consists in 3 main things:

1. Add the ability to retrieve/store the default branch used by git by reading/writing the git global config key `init.defaultBranch` (which is the same config option name that `git` [is going to use](https://github.com/git/git/blob/4a0fcf9f760c9774be77f51e1e88a7499b53d2e2/Documentation/config/init.txt#L5-L7) in the upcoming `v2.28` release).
2. Use the configured default branch name when creating a new repository to use that branch as the default one (we do this manually by executing `git checkout -b <default_branch>` just after `git init`).
3. Add a new setting in the git tab on the preferences dialog to allow user to configure the default branch.

## Considerations

In order to align with [the GitHub plan](https://github.com/github/renaming/#what-were-changing-pencil2) to rename the default branch from `master`, this PR implements the first point of the plan ("a configurable default for new repositories").

This means that:

* If the user doesn't change the default branch from the settings dialog (or from their global git config), Desktop will keep using the same default branch as previously (which is `master`).
* We're offering `main` as an easy-to-choose alternative via a radio button, but we're allowing users to select any name they want for the default branch.

In order to prepare for the next step of the default branch rename, we're not storing any default branch unless the user explicitly changes their chosen default branch. This will allow us in the near future to set the default to `main` for any new user or existing user that hasn't chosen a default branch.

## Screenshots

demo

## Release notes

Notes: [New] Allow to configure the default branch used in new repositories.


tierninho:
Tested this and it looks good. However, if a user want to name the branch "maintenance" or "masterP" they cannot type it in as it looks like the user is trying to type `main` or `master` . You _can_ paste any word into the `other` field, but the user should be able to type it as it is more intuitive. 

Also, I did not see the setting in the global gitconfig. Was this implemented?

tierninho:
Also filed a related issue here: https://github.com/desktop/desktop/issues/10293

Lastly, will any setting in the global git config override Desktop? Or will it show up in the `Other` field pre-populated?

rafeca:
> Tested this and it looks good. However, if a user want to name the branch "maintenance" or "masterP" they cannot type it in as it looks like the user is trying to type main or master . You can paste any word into the other field, but the user should be able to type it as it is more intuitive.

Good one! didn't think about this, I'm going to fix it since that's quite annoying 😅

> Also, I did not see the setting in the global gitconfig. Was this implemented?

Yes, we only save it when the user clicks on "Save". The setting should look like:

```
[init]
        defaultBranch = main
```

> Also filed a related issue here: #10293

👍 gonna answer that one there!

> Lastly, will any setting in the global git config override Desktop? Or will it show up in the Other field pre-populated?

It should show up in the "Other" field pre-populated if it's different that `main` or `master` (it should show up with `main` or `master` selected if the value is one of these).

rafeca:
> Good one! didn't think about this, I'm going to fix it since that's quite annoying 😅

Addressed in 206b860, now we only execute the logic that autoselects `main` or `master` when the user leaves the text input, so a user can type `masterA` without issues (if they type `master` and leave the text input, the `master` option will get selected).

app/src/lib/feature-flag.ts
```diff
@@ -133,3 +133,10 @@ export function enableForkSettings(): boolean {
 export function enableDiscardLines(): boolean {
   return true
 }
+
+/**
+ * Should we allow to change the default branch when creating new repositories?
+ */
+export function enableDefaultBranchSetting(): boolean {
+  return enableDevelopmentFeatures()
```
niik, line 141: As discussed in the async sync today we can ship this to beta now.

```suggestion
  return enableBetaFeatures()
```

app/src/ui/preferences/git.tsx
```diff
@@ -29,7 +54,82 @@ export class Git extends React.Component<IGitProps, {}> {
...
+
+    const defaultBranchIsOther = !SuggestedBranchNames.includes(
+      this.props.defaultBranch
+    )
+
+    return (
+      <>
+        <p className="default-branch-title">
+          Default branch on new repositories
+        </p>
```
niik, line 95: We use `h2` in the advanced tab for section titles, let's mirror that there and I'd suggest we say "Default branch **for** new..." instead of **on**

```suggestion
        <h2>
          Default branch for new repositories
        </h2>
```

app/styles/ui/_preferences.scss
```diff
@@ -59,6 +59,21 @@
       color: var(--text-secondary-color);
     }
   }
+
+  .default-branch-title {
+    margin-top: var(--spacing-double);
+    margin-bottom: var(--spacing-half);
+  }
```
niik, line 66: We can use `h2` instead.
```suggestion
```

rafeca:
I think a little bit more margin between the name/email fields and the default branch field won't hurt:

Screen Shot 2020-08-17 at 19 14 42

rafeca:
In 6568f47 I've tweaked the margins to make it look more similarly than the advanced preferences section:

Screen Shot 2020-08-17 at 19 22 50

rafeca:
> type something and hit Enter (instead of clicking on the Save button) the change doesn't get persisted. I imagine this is because we update the branch name on blur but the blur happens after the dialog gets dismissed.

Good catch! Damm it I didn't think about this edge case 😟  gonna look into how can I fix this easily and come back

rafeca:
> gonna look into how can I fix this easily and come back

@niik I've fixed this by using the `onValue` handler instead of the `onBlur` one and creating some local state to know whether the current branch is `Other` or not.

This local state gets calculated on component mount and when the user clicks on any of the radio buttons, but it doesn't get updated when the value of the `Other` textbox changes (to prevent the issue described in https://github.com/desktop/desktop/pull/10262#issuecomment-665829488).

I think we can argue the same as in https://github.com/desktop/desktop/pull/10261#discussion_r466649375 about this approach (we're not reacting on prop updates), but I haven't found a better way to do it without complicating things too much (any suggestion is welcome 😃 ).

niik:
> @niik I've fixed this by using the `onValue` handler instead of the `onBlur` one and creating some local state to know whether the current branch is `Other` or not.

I'm not sure if it matters or not but with this change we won't select the `main` or `master` radio buttons if the user types `main` or `master` and then tabs out of the control. I believe we could get both behaviors working if we in addition to listening for `onValueChanged` keep listening to `onBlur` but only use `onBlur` to update the local state. I.e something like this:

```diff
diff --git a/app/src/ui/preferences/git.tsx b/app/src/ui/preferences/git.tsx
index 4f6a8f62c..63ca30197 100644
--- a/app/src/ui/preferences/git.tsx
+++ b/app/src/ui/preferences/git.tsx
@@ -132,6 +132,7 @@ export class Git extends React.Component<IGitProps, IGitState> {
           <RefNameTextBox
             initialValue={this.props.defaultBranch}
             renderWarningMessage={this.renderWarningMessage}
+            onBlur={this.onRefNameBlur}
             onValueChange={this.props.onDefaultBranchChanged}
             ref={this.defaultBranchInputRef}
           />
@@ -155,10 +156,17 @@ export class Git extends React.Component<IGitProps, IGitState> {
    * @param defaultBranch string the selected default branch
    */
   private onDefaultBranchChanged = (defaultBranch: string) => {
+    this.updateDefaultBranchIsOther(defaultBranch)
+    this.props.onDefaultBranchChanged(defaultBranch)
+  }
+
+  private onRefNameBlur = (defaultBranch: string) => {
+    this.updateDefaultBranchIsOther(defaultBranch)
+  }
+
+  private updateDefaultBranchIsOther(defaultBranch: string) {
     this.setState({
       defaultBranchIsOther: !SuggestedBranchNames.includes(defaultBranch),
     })
-
-    this.props.onDefaultBranchChanged(defaultBranch)
   }
 }
```

rafeca:
> I'm not sure if it matters or not but with this change we won't select the main or master radio buttons if the user types main or master and then tabs out of the control. I believe we could get both behaviors working if we in addition to listening for onValueChanged keep listening to onBlur but only use onBlur to update the local state. I.e something like this:

Apart from the issue you mentioned above, there's another (very small) issue with the `onBlur()` approach which happens when the user writes either `master` or `main` and then clicks on the "Save" button (which will happen if we add the changes you suggested):

demo

Since we update DOM during the blur event, it happens very often that the click event is not triggered and therefore ignored, since when the mouseup event is emitted, the mouse is not over the button anymore.

It's a very small issue anyways, but I think that getting the UI updated when unfocusing does not matter enough to introduce this small nuance.

niik:
> It's a very small issue anyways, but I think that getting the UI updated when unfocusing does not matter enough to introduce this small nuance.

Definitely not, thanks for explaining that!

niik (approved):
Great work @rafeca!

Issue #5093, opened 2018-07-03
upgrade react-transition-group to 2.4.x


shiftkey:
As part of the transition to async rendering in React, we need to upgrade `react-transition-group` to a version that works with the new component APIs.

`v2.4.0` just landed last week with that support, but we're still using`1.x` and the transition process, while [documented pretty well](https://github.com/reactjs/react-transition-group/blob/v2.4.0/Migration.md), is not trivial.

There's a fair bit of animations in the app, so it should be easy to identify we've got things right, so this seems ideal for someone who knows the library a bit and can help with the migration.

Issue #9875, opened 2020-05-26
Make Desktop pass React's strict mode


rafeca:
## Motivation

The benefits of making a React app run in strict mode are well described in [the official docs](https://reactjs.org/docs/strict-mode.html), but here's a summary of the most important benefits for us:

* Stop using the deprecated lifecycle methods that throw a warning on latest React versions.
* Make Desktop ready for [concurrent mode](https://reactjs.org/docs/concurrent-mode-intro.html) once React ships it. This will allow us to use Suspense
* It's a good opportunity to clean up some components logic where the state handling logic has become very complicated or fragile.
* Fix some potential subtle issues due to the inconsistency on behaviour of the deprecated lifecycle methods (I don't think this is a big factor, since it's possible that some of these issues have already been worked around or don't affect users at all).

## Steps involved

There are 2 main steps required for the migration:

1. Update our components code to not throw errors in strict mode.
2. Update third party dependencies to versions that do not throw errors in strict mode.

The second point may involve more investigation/uncertain, since it's unclear the total number of dependencies to upgrade and the amount of work needed to upgrade them.

So far I have identified the following third party libraries that need to get updated:

* `react-transition-group`

## Migration tips and tricks

These are some trickt to help us get rid of the deprecated React methods: 

### `componentWillReceiveProps`

* When used to calculate some state with the main purpose of caching a calculated value, it's usually safe to create a memoized function with `memoizeOne` and call it on every render. You can notice this one when the same logic is applied on the constructor/`componentWillMount` and these are the only places where state is updated ([example](https://github.com/desktop/desktop/blob/3c31e8e9b202f379f740d25454b1bbcf7a013594/app/src/ui/changes/changes-list.tsx#L210-L222)).
* In rare situations, the calculation of the value may depend on previous state/data ([example](https://github.com/desktop/desktop/blob/development/app/src/ui/branches/pull-request-list.tsx#L113-L121)).  In these cases keeping the local state may be required, so `getDerivedStateFromProps` can be used.
  * Note that having local state that gets updated based on props [in most of the cases is an anti-pattern](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html), but refactoring current components to avoid this could be specially tricky.
[...]

PR #10319, opened 2020-08-04
Upgrade react-transition-group npm package

rafeca:
Addresses https://github.com/desktop/desktop/issues/9875
Closes https://github.com/desktop/desktop/issues/5093

## Description

This PR updates the `react-transition-group` npm package from v1.2.0 to v4.4.1.

The currently used version [is deprecated](https://github.com/reactjs/react-transition-group#react-transition-group-) and is not going under any maintenance.

Apart from that, the v1 version of the library still uses the deprecated `componentWillMount()`, `componentWillUpdate()`, etc lifecycle methods, which are deprecated and [render warnings](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) when used on newer versions of React. There was an attempt of replacing these methods in the v1 branch (see [this PR](https://github.com/reactjs/react-transition-group/pull/355)) but the maintainers decided to not publish a new v1 version with these changes (see [this comment](https://github.com/reactjs/react-transition-group/pull/355#issuecomment-400827723)):

> the logic in the PR is not the same as the old code. cDU is called more often than cWRP did, and the timing is also very different. I'm not very comfortable releasing this, as we don't really support v1 anymore (though haven't offically deprecated it). Practically maintainers don't use v1 anymore and I don't feel very comfortable in the v1 codebase anymore.

(note that this was 2 years ago, now v1 is officially deprecated).

The changes made in this PR have been done following [the migration guide from v1](https://github.com/reactjs/react-transition-group/blob/master/Migration.md), which has been quite helpful.

### Test plan

I've tested that the following animations look the same as in production:

- Modal popups getting opened/closed.
- The undo commit indicator appearing.
- The full screen information badge.
- The zoom in/zoom out information badge.

## Release notes

Notes: no-notes


app/webpack.common.ts
```diff
@@ -59,7 +59,6 @@ const commonConfig: webpack.Configuration = {
   ],
   resolve: {
     extensions: ['.js', '.ts', '.tsx'],
-    modules: [path.resolve(__dirname, 'node_modules/')],
```
rafeca, line 62: Without this fix this PR would not build, since the newer version of `react-transition-group` depends on `dom-helpers@^5.0.1`, which has a different API than `dom-helpers@3.x`.

With the current dependency versions, v3 gets installed in `app/node_modules/dom-helpers`, while v5 gets installed in `app/node_modules/react-transition-group/node_modules/dom-helpers`. Because of this WebPack config, the v3 version was always getting used by WebPack.

Note that this change can potentially have an impact on our bundle size, since we used to (wrongly) deduplicate npm dependencies by always using the version installed in the root `node_modules` folder.

I'm going to measure the impact on bundle size

say25:
This should also close https://github.com/desktop/desktop/issues/5093 and make progress towards the issue it references https://github.com/desktop/desktop/issues/4369 as well as the milestone it is a part of https://github.com/desktop/desktop/projects/13#column-3110838

app/webpack.common.ts
rafeca: Results comparing bundle sizes of development with this branch:


| | development | this branch | difference | % |
|-|-|-|-|-|
| Main process | 253924 | 262277 | 🔼 8.3KiB | 🔼 3.12%  |
| Renderer process | 1794308 |1812048  | 🔼 17.7KiB | 🔼 0.97% |

We can mitigate substantially this increase by deduplicating compatible dependencies from our `yarn.lock`, like the following:

https://github.com/desktop/desktop/blob/3cab784e0cbaed1fa342ed969777b720c414cb49/app/yarn.lock#L615-L623

I did a quick test deduplicating dependencies with [`yarn-deduplicate`](https://github.com/atlassian/yarn-deduplicate) and with that the main process bundle size on this PR was the same as in master and the renderer process bundle size on this PR was only 7.3KiB bigger than master.

app/src/ui/window/full-screen-info.tsx
```diff
@@ -93,9 +93,20 @@ export class FullScreenInfo extends React.Component<
...
-      </div>
+      <CSSTransition
+        classNames="toast-animation"
+        appear={true}
+        enter={false}
+        exit={true}
+        timeout={{
+          appear: transitionAppearDuration,
+          exit: transitionLeaveDuration,
+        }}
```
niik, line 104: Can we extract this into a const please? Perhaps `toastTransitionTimeout`?

app/src/ui/app.tsx
```diff
@@ -2505,14 +2508,16 @@ export class App extends React.Component<IAppProps, IAppState> {
...
-        transitionEnterTimeout={500}
-        transitionLeaveTimeout={400}
-      >
-        {banner}
-      </CSSTransitionGroup>
+      <TransitionGroup>
+        {banner && (
+          <CSSTransition
+            classNames="banner"
+            timeout={{ enter: 500, exit: 400 }}
```
niik, line 2515: Can we extract this into a const please? Perhaps `bannerTransitionTimeout`?

PR #10334, opened 2020-08-05
Fix keyboard navigation in PR list

niik:
## Description

This was caught as part of #9737, we have a `selectedBranch` state property for the PR list but it's only ever written to and never used for the actual selection. The consequences of this is that you can't navigate the PR list using the up/down keys.

The PR list receives the currently "selected" (i.e. checked out) pull request via props but we need to keep track of what they user has currently selected in the list but not necessarily checked out. By leveraging `getDerivedStateFromProps` we can reset the selection should the currently checked out branch in a repo change while still keeping track of the currently selected/focused/highlighted PR in the list.

## Release notes

Notes: [Fixed] Pull request list is now keyboard accessible

### DIFF
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -55,8 +55,14 @@ interface IBranchesContainerProps {
 }
 
 interface IBranchesContainerState {
-  readonly selectedBranch: Branch | null
+  /**
+   * A copy of the last seen currentPullRequest property
+   * from props. Used in order to be able to detect when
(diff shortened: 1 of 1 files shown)


niik:
@tierninho That's weird, I'm not able to reproduce your results

pr-list-keyboard-win

Looking at your screenshot the "Branches" and "Pull requests" tabs look squished. I wonder if there's something else going on with your environment. Could you try again but with a full `yarn clean-slate && yarn build:dev && yarn start` ?

PR #9767, opened 2020-05-12
Add spell checking to GitHub desktop

say25:
⚠️ Warning: Builds off #9766 ⚠️ 

Closes #1597 

## Description

- Add Spelling warnings to GitHub Desktop text boxes

### Screenshots

spell-checking-typo-image

## Release notes

Notes: Add Spelling warnings to GitHub Desktop text boxes

PR #10398, opened 2020-08-17
Disable spellchecking for the time being

niik:
## Description

I noticed some squiggly red lines under a typo of mine in the latest beta and was surprised because I thought @say25's PR #9767 that dealt with enabling spell checking hadn't been merged. Turns out is hasn't but Electron enabled spellchecking by default in Electron 9.

image

Unfortunately it only marks unknown words as misspelled but doesn't let the user pick a substitute or add a false-positive to the dictionary. There are [ways of enabling suggestions and opt-outs](https://www.electronjs.org/docs/all#how-do-i-put-the-results-of-the-spellchecker-in-my-context-menu) in the context menu but it'll require a bit more work for us due to our custom context menus for the commit message.

So instead of shipping a half-baked solution I'm turning off spell checking now so that we can use #9767 to discuss a more holistic solution to spellchecking in Desktop including the ability to turn it on/off in preferences.

### DIFF
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -43,6 +43,7 @@ export class AppWindow {
         disableBlinkFeatures: 'Auxclick',
         nodeIntegration: true,
         enableRemoteModule: true,
+        spellcheck: false,
       },
       acceptFirstMouse: true,
     }
--- a/app/src/main-process/crash-window.ts
+++ b/app/src/main-process/crash-window.ts
@@ -37,6 +37,7 @@ export class CrashWindow {
         // See https://developers.google.com/web/updates/2016/10/auxclick
         disableBlinkFeatures: 'Auxclick',
         nodeIntegration: true,
+        spellcheck: false,
       },
     }
 


say25:
@niik out of curiosity, would specifying the language locale improve the UX with this? I can understand waiting to enable this feature but it would definitely help me out with my day to day work.

niik:
> @niik out of curiosity, would specifying the language locale improve the UX with this?

I don't think so. In fact I believe Electron already defaults to using the OS locale, See https://github.com/electron/electron/blob/master/docs/tutorial/spellchecker.md#how-to-set-the-languages-the-spellchecker-uses

> By default the spellchecker will enable the language matching the current OS locale.

The problem here is that we don't surface suggested corrections and that we don't allow users to add words to the dictionary. As @billygriffin outlined in https://github.com/desktop/desktop/pull/9767#issuecomment-635494581 it's not uncommon for folks to use domain specific language that wouldn't be part of a standard dictionary.

> I can understand waiting to enable this feature but it would definitely help me out with my day to day work.

You and me both, my keyboard is acting up and I'm typoing all over the place. Nevertheless I feel that we must take the time to make sure we do it right.

PR #10415, opened 2020-08-21
Allow to rename the default branch

rafeca:
## Description

This PR removes the check that prevents the default branch of a repository to be renamed. This action will work fine for local repositories, non-GitHub repositories and GitHub repositories but it causes some minor side effects:

### Local repositories and non-GitHub repositories

When the default branch of a local repository is different than the globally configured default branch in Desktop (see https://github.com/desktop/desktop/pull/10262), then it won't be detected as the default branch:

* The unused branch pruner won't be executed since it won't find the default branch of the repo.
* In the different places where the branches list is displayed, we won't show the default branch at the top of the list.
* The popup asking whether to create a new branch from the current branch or the default branch won't be shown.

Note that none of these side-effects are exclusive to renaming the default branch: This same behaviour will be visible when a user changes its default branch setting in GitHub Desktop and opens an old local repository that has a different default branch than the currently configured.

Also, these side-effects look quite minor to me and taking into account that the main use case of GitHub Desktop is to use it with GitHub repositories I don't think these should block us from shipping this.

### GitHub repositories

When renaming the default branch of a GitHub repository, an expected popup warning the user that the branch also exists on the remote is displayed:

Screen Shot 2020-08-21 at 09 45 36

After renaming the branch, the local branch with the old name tracks the remote branch that has the previous name (as expected):

```sh
$ git branch -vv
* main   0641c9c [origin/master] some commit
```

This seems to work nicely for all the git operations, but as a side effect GitHub Desktop is not able to recognize the renamed branch as the default branch. This seems to be caused by the fact that now the local branch name does not match the default branch name returned by the GitHub API.

We could fix that by tweaking the logic that calculates the default branch on GitHub repositories and take into account branch tracking information in `git` to do a mapping between local and remote branches.

## Release notes

Notes: [Added] Allow to rename the default branch of a repository

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -242,7 +242,7 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
 
     menuStateBuilder.setEnabled(
       'rename-branch',
-      onNonDefaultBranch && !branchIsUnborn && !onDetachedHead
+      !branchIsUnborn && !onDetachedHead
     )
     menuStateBuilder.setEnabled(
       'delete-branch',
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -192,6 +192,8 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
     if (tip.kind === TipState.Valid) {
       if (defaultBranch !== null) {
         onNonDefaultBranch = tip.branch.name !== defaultBranch.name
+      } else {
+        onNonDefaultBranch = true
       }
 
       hasPublishedBranch = !!tip.branch.upstream
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -244,7 +244,9 @@ function getRepositoryMenuBuilder(state: IAppState): MenuStateBuilder {
 
     menuStateBuilder.setEnabled(
       'rename-branch',
-      !branchIsUnborn && !onDetachedHead
+      (onNonDefaultBranch || !hasPublishedBranch) &&
+        !branchIsUnborn &&
+        !onDetachedHead
     )
     menuStateBuilder.setEnabled(
       'delete-branch',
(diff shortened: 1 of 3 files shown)


rafeca:
> We could fix that by tweaking the logic that calculates the default branch on GitHub repositories and take into account branch tracking information in git to do a mapping between local and remote branches.

This can be achieved by changing the logic that finds the local default branch, but I'm not 100% sure whether this could cause any other side effect:

```diff
diff --git a/app/src/lib/stores/git-store.ts b/app/src/lib/stores/git-store.ts
index c25437b34..34dccb0dc 100644
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -465,7 +465,11 @@ export class GitStore extends BaseStore {
     // priority to local branches by sorting them before remotes
     this._defaultBranch =
       this._allBranches
-        .filter(b => b.name === defaultBranchName)
+        .filter(
+          b =>
+            b.name === defaultBranchName ||
+            b.upstreamWithoutRemote === defaultBranchName
+        )
         .sort((x, y) => compare(x.type, y.type))
         .shift() || null
   }
```

rafeca:
> I did not notice one potential side-effect. If you rename the default branch to pick-anything and then create a new branch with the same name as the default branch setting, like main, it will make that new branch the new default branch.

Thanks @tierninho for testing this!

I'm going to disable default branch renaming when the branch has already been pushed, so we can limit the amount of edge cases. The principal use case to rename the default branch from Desktop is just after creating a new repository and realizing that the default branch is not what the user expected to be (e.g when we switch the default branch from `master` to `main`).

There's a small improvement that we can do in the near future once GitHub launches the default branch rename functionality: do not disable the menu item to rename the default branch on GitHub repositories, and when the user clicks on it show a modal with instructions about how to rename the branch from GitHub.

To your point, renaming the default branch in a local repository has the limitation that we're no longer able to detect it as the default branch. Note that the concept of a default branch is Desktop/GitHub specific, so we cannot calculate it only based on the `git` project but we need to store it somewhere (github.com) or assume its value (so far we assumed it was `master`, now we assume it's the configured value of the default branch setting). With this, there's no easy way for GitHub Desktop to know which of its branches is the default one.

niik (approved):
Great incremental improvement here @rafeca!

PR #10410, opened 2020-08-20
Add AspNet and Diff/Patch Syntax Highlights

say25:
Closes #[issue number]

## Description

- Added syntax highlighting for some Asp.Net Files
   - Extension Examples
       -  [.props](https://github.com/PostgreSQLCopyHelper/PostgreSQLCopyHelper/blob/master/Directory.Build.props)
       - [.targets](https://github.com/npgsql/npgsql/blob/dev/Directory.Build.targets)
       - [.aspx](https://github.com/aspnet/samples/blob/master/samples/aspnet/WebApi/WebFormSample/WebForm/About.aspx)
       - [.cshtml](https://github.com/aspnet/samples/blob/master/samples/aspnet/WebApi/Todo/Todo.Web/Views/Home/Index.cshtml)
  - Added syntax highlighting for diff files

### Related PRs

- https://github.com/desktop/highlighter-tests/pull/41

## Release notes

Notes: Added syntax highlight for .props, .targets, .aspx, .cshtml and .patch/.diff files

PR #10407, opened 2020-08-19
Release 2.5.4-beta4

niik:
## Description

Looking for the PR for the upcoming v2.5.4-beta4 beta release? Well you've just found it, congratulations! :tada:

## Release checklist

Release checklist:

- [x] ~~Check to see if there are any errors in Sentry that have only occurred since the last production release~~
- [x] Verify that all feature flags are flipped appropriately
  One new feature flag enabled for beta: https://github.com/desktop/desktop/blob/c83318799c6899f5153713b236a652a2c8985710/app/src/lib/feature-flag.ts#L137-L142
- [x] If there are any new metrics, ensure that central and desktop.github.com have been updated

PR #10434, opened 2020-08-24
improved detection for publishable builds when running CI

shiftkey:
:wave: I hope you're all doing well. Congrats on a shiny new release today - can't wait to try it out myself! Also, no rush on this PR - I can workaround it in the meantime, but seeing this project leverage Actions will make maintaining the fork much easier for me. Kudos!

This is an infrastructure fix for an issue I've started to encounter over in `shiftkey/desktop` since https://github.com/desktop/desktop/pull/9680 was merged into `development`. 

### Description

When I pushed `development` to my fork I see that the CI build on macOS tries to setup the keychain, which it doesn't have access to, and errors.

https://github.com/shiftkey/desktop/runs/1022917605#step:8:403

```
Setting up keychain…
...
security: SecKeychainItemImport: Unknown format in import.
security: SecItemCopyMatching: The specified item could not be found in the keychain.

    at checkExecSyncError (child_process.js:621:11)
    at Object.execSync (child_process.js:657:15)
...
```

I believe it's related to this new check inside `script/build-platforms.ts`:

https://github.com/desktop/desktop/blob/ac3b9e8b871ca57187bcfe5f6332ceaad546be8b/script/build-platforms.ts#L66-L72

My testing indicates that `GITHUB_HEAD_REF` is not always present, and my current working theory is that it's only available for `pull_request` events. I poked at the default branch in my fork to add some tracing output, triggered a `push` event and got these values for the two important environment variables.

https://github.com/shiftkey/desktop/runs/1023228953?check_suite_focus=true#step:6:396

```
Found environment variables GITHUB_REPOSITORY='shiftkey/desktop' and GITHUB_HEAD_REF=''
```

As discussed with @niik below, this PR removes the `isRunningOnFork()` check in favour of the more explicit `isPublishable()` which uses the branch name to determine whether this build should be published (and thus we need to set the keychain).

This PR should trigger a `pull_request` workflow, and avoid setting up the keychain, as these recent PRs indicate is the desired behaviour:

 - `pull_request` event from fork - https://github.com/desktop/desktop/pull/10410 - [[failing console message]](https://github.com/desktop/desktop/runs/1008518373#step:8:409)
 - `pull_request` event from local branch - https://github.com/desktop/desktop/pull/10407 - [[failing console message]](https://github.com/desktop/desktop/runs/1002520685#step:8:407)

There's also some `isCircleCI()` checks in here which are no longer necessary - I won't remove them all, just the ones relevant to this PR.

TODO:

 - [x] confirm CI does not try and setup keychain by default
 - [x] confirm changes work on `shiftkey/desktop`  - https://github.com/shiftkey/desktop/runs/1026751693#step:6:402

## Release notes

Notes: improved detection for publishable builds when running CI

### DIFF
--- a/script/build-platforms.ts
+++ b/script/build-platforms.ts
@@ -32,48 +32,6 @@ export function getSha() {
   )
 }
 
-export function isRunningOnFork() {
-  if (isCircleCI() && process.env.CIRCLE_PR_USERNAME != null) {
-    return true
-  }
-
-  if (
-    isAppveyor() &&
-    process.env.APPVEYOR_PULL_REQUEST_NUMBER != null &&
-    process.env.APPVEYOR_PULL_REQUEST_HEAD_REPO_NAME !== 'desktop/desktop'
-  ) {
-    return true
-  }
-
-  if (
-    isTravis() &&
-    process.env.TRAVIS_PULL_REQUEST_SLUG != null &&
-    // empty string denotes a `push` build
-    process.env.TRAVIS_PULL_REQUEST_SLUG !== '' &&
-    process.env.TRAVIS_PULL_REQUEST_SLUG !== 'desktop/desktop'
-  ) {
-    return true
-  }
-
-  if (
-    isAzurePipelines() &&
-    process.env.SYSTEM_PULLREQUEST_ISFORK != null &&
-    process.env.SYSTEM_PULLREQUEST_ISFORK === 'True'
-  ) {
-    return true
-  }
-
-  if (
-    isGitHubActions() &&
-    process.env.GITHUB_HEAD_REF !== undefined &&
-    process.env.GITHUB_HEAD_REF.length > 0
-  ) {
-    return true
-  }
-
-  return false
-}
-
 export function isTravis() {
   return process.platform === 'linux' && process.env.TRAVIS === 'true'
 }
--- a/script/build.ts
+++ b/script/build.ts
@@ -75,7 +75,7 @@ moveAnalysisFiles()
 if (
   (isCircleCI() || isGitHubActions()) &&
   process.platform === 'darwin' &&
-  isPublishable()
+  isPublishableBuild
 ) {
   console.log('Setting up keychain…')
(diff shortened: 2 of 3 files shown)


shiftkey:
> **If we should be marking all `pull_request` events as being "from a fork" that might make this change simpler. Does that make sense from your perspective?**

I ended up going down this path and not needing to check the `GITHUB_HEAD_REF` environment variable, which simplified the checks.

niik (changes requested):
Hey @shiftkey, good to see you in this repository again! We're still learning the ins and outs of actions so chances are we're not using these env vars correctly.

That said I took a look at the offending method `isRunningOnFork` and I see that it's only used two spots

https://github.com/desktop/desktop/blob/45f006ec96fcd0e5a38de4080c96a7722321ce32/script/build.ts#L75-L82

and

https://github.com/desktop/desktop/blob/56dc6a59fafe335bbe445a9ebbf6c5ffee5bfa0b/script/test-setup.ts#L10-L23

The first case is installing the certificates needed for code signing and the second one is verifying that the binaries got signed correctly. I don't see how either step would be necessary unless we're building with the intent of publishing so we should be able to remove the `isRunningOnFork` method entirely and substitute its callsites with `isPublishable`...

https://github.com/desktop/desktop/blob/ea6b809b63e336c84bce6f42276682ea659dac06/script/dist-info.ts#L103-L108

...which we're already using for things like notarization...

https://github.com/desktop/desktop/blob/ea6b809b63e336c84bce6f42276682ea659dac06/script/build.ts#L165

...and obviously publishing...

https://github.com/desktop/desktop/blob/ea6b809b63e336c84bce6f42276682ea659dac06/script/publish.ts#L6-L9

shiftkey:
👍 on swapping it out for `isPublishable()` - will put this back to draft and take a shot at porting those usages over

niik (approved):
This looks great to me. Gonna merge it and try it out in #10439 in a bit. Thanks!

LisaSante (approved):
done

PR #10452, opened 2020-08-26
refactor: Rename MergeResult to MergeTreeResult

niik:
## Description

This type is only used to communicate the status or result of a `git merge-tree` operation. This rename will open us up to be able to use MergeResult to communicate the result of a `git merge` operation. I'm opening this ahead of another PR in order to minimize the clutter of all the renames in that PR.

**There should be no logical effects on this change whatsoever, it's straight up a rename of a type and a method**

## Release notes

Notes: no-notes


rafeca (approved):
LGTM!

PR #10458, opened 2020-08-27
Show loading indicator when committing a merge

niik:
## Description

Ran into this yesterday while working on a bug fix for electron (https://github.com/electron/electron/pull/25124). The electron project makes heavy use of git hooks and after merging from their default branch and resolving the conflicts I went to commit from the merge conflict resolution dialog which seemingly did nothing.

Turns out the commit hook ran for 30+ seconds before completing and we didn't  provide any feedback. The commit button was still enabled making it possible to launch another git commit process while the first one was running.

#### Before

image

#### After

image

## Release notes

Notes: [Improved] Show progress spinner while waiting merge commit to be created after resolving conflicts


rafeca (approved):
Nice one!

PR #10459, opened 2020-08-27
Avoid potentially expensive and potentially network dependent findAccountForRemoteURL in push

niik:
## Description

Ran into this while reviewing another PR. In https://github.com/desktop/desktop/pull/5050 we added metrics to see how many  times a day users  were pushing to either GitHub.com, GitHub Enterprise Server, or other Git hosts.

The `recordPush` method was set up to take an `Account` instance which we used to determine which type of host the push was targeting based on its `endpoint` property. This is all find but since the `_push` app store method didn't have easy access to the `Account` but only `IGitAccount` the decision was made to leverage `findAccountForRemoteURL` to resolve the correct account based on the remote url.

`findAccountForRemoteURL` is a tricky beast though. In some scenarios (likely not in the one we're looking at in _push though) it can start issuing API requests to differentiate between a GHES and a GitHub.com account. Instead we'll leverage the synchronous `getAccountForRepository` method which is what `withAuthenticatingUser` is using to provide the `IGitAccount` to `_push`:

https://github.com/desktop/desktop/blob/47d420132dbc04d76bf5d159ed9f6e531cb89aba/app/src/lib/stores/app-store.ts#L5020-L5062

In fact we were so inclined we could straight up use the `IGitAccount` instance and do something like `recordPush(account instanceof Account ? account : null)` but that's relying on implementation details that we absolutely shouldn't be relying on in this method.

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -21,6 +21,7 @@ import {
   setNumberArray,
 } from '../local-storage'
 import { PushOptions } from '../git'
+import { IGitAccount } from '../../models/git-account'
 
 const StatsEndpoint = 'https://central.github.com/api/usage/desktop'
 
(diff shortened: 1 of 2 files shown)


rafeca (approved):
LGTM!

Issue #10479, opened 2020-09-01
Have to left click and hold to "select branch to compare" in Windows UI - single or double click not working


DennisWormald:
Have to left click and hold to "select branch to compare" in Windows UI - single or double click not working

### Version 2.5.4 on Windows 10 most recent update - this problem didn't happen with the last version.

Open 'About GitHub Desktop' menu to see the Desktop version. Also include what operating system you are using.

### Steps to reproduce the behavior

1. Go to Windows UI in repository with 2 or more branches that are currently identical
2. Make a change to the active branch 
3. Click "Commit to $branch_name"
4. On the "Current branch" menu, click and select any other branch
5. Click on "history" on the left tab/pane
6. Click on "Select branches to compare"
7. Left click (or double click) on the previous branch you were just on (under "Recent branches")
8. See error

![error](https://user-images.githubusercontent.com/29453368/91788505-b61ac280-ec4f-11ea-8964-29a864b39507.jpg)

### Expected behavior

You should be able to see the merge into "other branch name" button at the bottom of the left pane
![Expected](https://user-images.githubusercontent.com/29453368/91788501-b450ff00-ec4f-11ea-8597-92167b2b03a0.jpg)

### Actual behavior

You can't see the merge into "other branch name" button at the bottom of the left pane

### Additional context

If you click and hold you can sometimes get it to work, but its unpredictable.

rafeca:
Thanks for the report @DennisWormald ! I've been able to reproduce it on macOS as well.

It's quite possible that the issue got introduced by https://github.com/desktop/desktop/pull/9977. Interestingly, I can't see the issue on my development version of Desktop, only in the production one, so it may be some caused by some race condition of events.

rafeca:
Update: I've been able to reproduce in my development version (it just not reproduces 100% of times there), and tried reverting #9977 and is still reproducible, so the issue may come from a different change

dhunmoon:
@rafeca This is not a new bug I have experienced something similar in the past as well. it goes way deep. At that point in UI if you double click instead on the single click it goes to the point where you can see `Behind` and `Ahead` changes.
 #10473 This bug is also along the same line. 
 Not just that it goes a little bit deeper and there are many other small hiccups I generally face while comparing branches. I think it has something to do with how the state has been managed(I could be completely wrong though)

I will create a new Issue regarding what I have faced.(and maybe add it to here also)

Issue #10473, opened 2020-08-29
Select Branch to Compare Bugs


JosephTLyons:
### Describe the bug

There seem to be bugs associated with Select Branch to Compare.  I have attached a GIF that shows a few of them.  In the video, I try to compare my `master` branch with a remote branch `main` in the [Nushell](https://github.com/nushell/nushell) project.  When I go to select the `main` branch, it filters out some branches, but still shows a few others, instead of just directly going into the comparison view of `master` and `main`.  Then, I try clicking the `main` branch, and it won't let me.

### Version & OS

macOS 10.15.6
GitHub Desktop v2.5.4

### Steps to reproduce the behavior

See attached GIF

### Expected behavior

I expected that the first time I click on a branch to compare, that it goes directly to the comparison view between that branch and the currently checked out one.

### Actual behavior

Some branches are filtered out, while some are not.

### Screenshots

Add screenshots to help explain your problem, if applicable.

![Select Branch to Compare Bugs](https://user-images.githubusercontent.com/19867440/91642122-bbc99b80-e9f6-11ea-9868-5f096c964bba.gif)

### Logs

Attach your logs by opening the `Help` menu and selecting `Show Logs...`, if applicable.

### Additional context

I reported some other bug with Select Branch to Compare previously [here](https://github.com/desktop/desktop/issues/9971); I wonder if this is a regression or if these are bugs that are related to that fix.

dhunmoon:
@JosephTLyons Yes, it does happen, If you single-click the list will not respond you should try double-clicking that will select the branch every time. Regardless that is an Issue and should not happen.

Other than that it's showing some other branches even if there is `main` in the filter because string `main` might be a substring in other branch names(that's my guess).

niik:
Hi @JosephTLyons, thanks for opening this. I _believe_ this was fixed by #10485 which we shipped in yesterday's patch release (GitHub Desktop 2.5.5). I'd really appreciate it if you could upgrade to the latest version of GitHub Desktop and let us know if you're still able to reproduce this. In the meantime I'm going to assume it's fixed and close this but let us know if it isn't and we'll reopen.

JosephTLyons:
[...]

Issue #10463, opened 2020-08-27
Comparing branches in full screen mode does not show the "merge into" information


robbiegod:
### Describe the bug

Hi, I recently updated Github Desktop to version 2.5.4 on Windows 10.  When I compare a branch to another branch and want to merge the branches together, I am not seeing the button to merge them together when I GD open in full screen mode.  The messages would appear above the "Merge into myBranch" blue button in the left column.

I can get the information to display by exiting full screen mode.

I have also tried running GD in Windowed mode, but still the message does not appear.  Only when I toggle the full screen mode will it appear.

### Version & OS

Github Desktop to version 2.5.4
Windows 10 Pro

### Steps to reproduce the behavior

1. Create two branches. 
2. Make some changes in one of the branches and commit those changes.
3. Switch to the other branch.
4. From the Branch menu, click Compare to Branch.

### Expected behavior

You should see the message in the left column of the window above the Merge into blue button that says something like "it is safe to merge branch_one into branch_two.

### Actual behavior

The message does not appear.  It only appears after I toggle full screen mode.

niik:
Hey @robbiegod, thanks for opening this. Do you think you could share a screenshot of what you're seeing? I tried reproducing it on Windows 10 running in both full screen and windowed mode but I couldn't.

<img width="844" alt="image" src="https://user-images.githubusercontent.com/634063/91456176-412c3f00-e883-11ea-9608-9c07a62bf618.png">

robbiegod:
Basically, where you see "There will be 4 conflicted files when merging origin/diff-too-large into repo-trash-error, I don't see that message until I toggle our of the full screen mode or the blue button if there is a message.

If there is nothing to merge, i also don't see the "there is nothing to merge message" and the "greyed out button.

I can try re-installing the application, perhaps it is something that went wrong with the update. I'll get you a screenshot asap.

robbiegod:
![Github-Desktop---no-merge-into-button-panel](https://user-images.githubusercontent.com/3711317/91459023-2f7f7300-e854-11ea-857c-ca7569f4604b.jpg)

See the attached screenshot. @niik

niik:
@robbiegod What happens if you click on the branch listed under "Other branches"? From looking at that screenshot it seems you've filtered the list but not yet selected the branch

robbiegod:
@niik  This was difficult to screenshot.  The missing button would appear when I clicked off of Github Desktop to open the snip tool.  I had to print screen and crop it for you...

[...]

PR #10485, opened 2020-09-02
Only focus compare branch input when the compareState changes

rafeca:
Closes https://github.com/desktop/desktop/issues/10479

Potentially addresses https://github.com/desktop/desktop/issues/10473 and https://github.com/desktop/desktop/issues/10463

## Description

This PR addresses the problem where clicking on a branch to compare re-focus the branch input and prevents the compare view from appearing.

It's quite probable that this has existed for a long time, but somehow may be only reproducible when the mouse events and the update from the store come in a certain order, and recent changes may have caused that the ordering has subtly changed and it's more reproducible now.

This seems to have been introduced in https://github.com/desktop/desktop/commit/e7295f44795988c8566be9ea308ac79f1329c32b, where a small refactor on the focus logic did not keep the checks to only do the focus when the compareState changes.

By testing this locally this fix solves the problem, but I wonder if we even need to do the focus on `componentDidUpdate()` (it looks like this was done to allow focusing the input when clicking "Compare Branches..." from the application menu, but the focus seems to be done regardless of this logic).

## Release notes

Notes: [Fixed] Clicking on a branch to compare now opens the compare view

### DIFF
--- a/app/src/ui/history/compare.tsx
+++ b/app/src/ui/history/compare.tsx
@@ -129,6 +129,10 @@ export class CompareSidebar extends React.Component<
   public componentDidUpdate(prevProps: ICompareSidebarProps) {
     const { showBranchList } = this.props.compareState
 
+    if (showBranchList === prevProps.compareState.showBranchList) {
+      return
+    }
+
     if (this.textbox !== null) {
       if (showBranchList) {
         this.textbox.focus()


niik (approved):
Thanks for looking into this @rafeca 

> By testing this locally this fix solves the problem, but I wonder if we even need to do the focus on `componentDidUpdate()` (it looks like this was done to allow focusing the input when clicking "Compare Branches..." from the application menu, but the focus seems to be done regardless of this logic).

You're right but the focus that happens when this `.focus` is removed is due to another "bug" or at least misunderstanding. I've been looking at fixing that in a separate branch which will mean that it will stop stealing focus. As such I think we should be explicit here and call `.focus` for now and then move the `.focus` to only be called when the app menu shortcut is used.

PR #10507, opened 2020-09-04
Don't discard all submodules when no submodules are specified

### LINKED ISSUE #10469: Discarding Changes to a File also Discards Unrelated Reverts Submodule Commits
### Describe the bug

When using submodules, if changes to a file completely unrelated to the submodule (e.g. Readme.md) are discarded, the submodule is randomly checked out to the last committed SHA (detached head state), even though the submodule should not have been touched.

### Version & OS

- Windows 10 (2004)
- Github Desktop Version 2.5.4

1. Open repository (Foo) which includes a submodule (Bar). Bar is currently at, say `c01dbed5`.
2. Make commits to the submodule (Bar), its head is now at, say `deadbeef`.
    Foo repository now shows Bar as changed - showing -`c01dbed5` +`deadbeef` - this is expected behaviour.
3. Edit some other file in Foo - lets say README.md.
    Foo now shows README.md as also being changed - again expected behaviour
4. Discard changes to **only** README.md in Foo.
5. Bar is now back at `c01dbed5` - even though it was **not** discarded! Changes in the submodule go irrevocably missing.

### Expected behavior

After discarding the unrelated file (README.md), the submodule should be untouched - it should have remained at `deadbeef` and not reverted, it was not asked to be discarded.

### Actual behavior

Changes to the submodule get discarded - it gets reverted from `deadbeef` to `c01dbed5` even though it was not asked to be discarded.

### Additional context

This has started happening since the latest update (2.5.4) - it may therefore be related to changes for #8218.

niik:
Closes #10469

## Description

https://github.com/desktop/desktop/pull/8218/commits/4cad2f5efa7d9578a0f1dbef7c0294fde212e13d  (which was part of #8218) introduced an unfortunate behavioral change

```diff
diff --git a/app/src/lib/git/submodule.ts b/app/src/lib/git/submodule.ts
index 497889aae0..2d15ed3b7b 100644
--- a/app/src/lib/git/submodule.ts
+++ b/app/src/lib/git/submodule.ts
@@ -80,11 +80,9 @@ export async function resetSubmodulePaths(
   repository: Repository,
   paths: ReadonlyArray<string>
 ): Promise<void> {
-  for (const path of paths) {
-    await git(
-      ['submodule', 'update', '--recursive', '--', path],
-      repository.path,
-      'updateSubmodule'
-    )
-  }
+  await git(
+    ['submodule', 'update', '--recursive', '--', ...paths],
+    repository.path,
+    'updateSubmodule'
+  )
 }
```

The intent behind this change was to leverage Git's own functionality for discarding changes in multiple submodules instead of repeatedly executing `git submodule update --recursive --` for each submodule. Unfortunately the removal of the for-loop removed the short-circuit of not calling Git at all if no submodules was provided. Turns out that if `git submodule update` is called without any paths it will enumerate all submodules and update them.

Reviewed in isolation the change looked good assuming that the caller was responsible for not calling the method if it didn't have any specific submodules to update but unfortunate that wasn't the case here:

https://github.com/desktop/desktop/blob/d7e7f602cf49556ef5162bdd2cb8c6be62091424/app/src/lib/stores/git-store.ts#L1493-L1500

Users will not notice this unless the submodule update fails which it might if the submodule isn't checked out and the clone fails (for example).

## Release notes

Notes: [Fixed] Discarding changed could fail due to incorrect attempt to update unrelated submodules

### DIFF
--- a/app/src/lib/git/submodule.ts
+++ b/app/src/lib/git/submodule.ts
@@ -80,6 +80,10 @@ export async function resetSubmodulePaths(
   repository: Repository,
   paths: ReadonlyArray<string>
 ): Promise<void> {
+  if (paths.length === 0) {
+    return
+  }
+
   await git(
     ['submodule', 'update', '--recursive', '--force', '--', ...paths],
     repository.path,
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -1491,7 +1491,10 @@ export class GitStore extends BaseStore {
     // 3. Checkout all the files that we've discarded that existed in the previous
     //    commit from the index.
     await this.performFailableOperation(async () => {
-      await resetSubmodulePaths(this.repository, submodulePaths)
+      if (submodulePaths.length > 0) {
+        await resetSubmodulePaths(this.repository, submodulePaths)
+      }
+
       await resetPaths(
         this.repository,
         GitResetMode.Mixed,


rafeca (approved):
lgtm! thanks for fixing that!

Tabby85:
Your welcome

PR #10519, opened 2020-09-07
Fix keyboard submit in merge and rebase dialog

### LINKED ISSUE #10511: Support keyboard to confirm on "Merge into branch" dialog
When clicking Branch -> Merge into Current Branch I'm presented with this dialog:

When the dialog appears, I can immediately type "master" and then use <cursor down> to select the upstream/master branch:

Screenshot 2020-09-05 at 11 55 45

However, I can't press <enter>/<space> (with or without any modifier key as far as I can tell) and have to use the mouse to perform the merge. It would be much better if I could do this using only the keyboard.

niik:
Closes #10511 

## Description

This allows users to initiate a merge or rebase by pressing Enter while focused on the list of branches or the filter text box (assuming there's a match) in the merge and rebase dialogs.

Since the only way to start a merge or rebase prior to this PR was clicking the button I've also had to rearrange the validation logic a bit to ensure that we don't simply rely on the button's enabled state to determine whether we can proceed or not.

## Release notes

Notes: [Improved] Pressing Enter in the merge or rebase dialog will now start the merge or rebase

### DIFF
--- a/app/src/ui/merge-branch/merge.tsx
+++ b/app/src/ui/merge-branch/merge.tsx
@@ -335,6 +335,10 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
   }
 
   private merge = () => {
+    if (!this.canMergeSelectedBranch()) {
+      return
+    }
+
     const branch = this.state.selectedBranch
     if (!branch) {
       return
--- a/app/src/ui/merge-branch/merge.tsx
+++ b/app/src/ui/merge-branch/merge.tsx
@@ -264,7 +264,12 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
       this.state.mergeStatus === null ||
       this.state.mergeStatus.kind !== ComputedAction.Invalid
 
-    return !selectedBranchIsCurrentBranch && isBehind && canMergeBranch
+    return (
+      selectedBranch !== null &&
+      !selectedBranchIsCurrentBranch &&
+      isBehind &&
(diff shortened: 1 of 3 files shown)


app/src/ui/merge-branch/merge.tsx
```diff
@@ -103,6 +104,19 @@ export class Merge extends React.Component<IMergeProps, IMergeState> {
...
 
+  private onItemClick = (branch: Branch, source: ClickSource) => {
+    if (source.kind !== 'keyboard' || source.event.key !== 'Enter') {
+      return
+    }
+
+    source.event.preventDefault()
+    const { selectedBranch } = this.state
+
+    if (selectedBranch !== null && selectedBranch.name === branch.name) {
```
rafeca, line 115: Is a race condition possible here where the clicked branch is different that the selected branch or this is a defensive check to make sure that we don't mess up?

rafeca (approved):
LGTM! just one question that doesn't need any change in this PR!

app/src/ui/merge-branch/merge.tsx
niik: The latter, perhaps too defensive since the only way to trigger a _keyboard_ event for a list row is to first move selection to it. For click events there's a chance of a race condition though.

tierninho (approved):
Confirming the functionality enacted is working as intended in both OSX/Windows. 👍

PR #10527, opened 2020-09-08
Flip default branch name preference feature flag and update default value

niik:
## Description

This is in advance of the next production release and enables the default branch name preference flow (https://github.com/desktop/desktop/pull/10262) for production as well as updates the default value for new repositories created in Desktop.

## Release notes

Notes: [Added] New settings to customize the default branch name to use for new repositories
[New] Repositories are created with 'main' as the initial branch name

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -138,5 +138,5 @@ export function enableDiscardLines(): boolean {
  * Should we allow to change the default branch when creating new repositories?
  */
 export function enableDefaultBranchSetting(): boolean {
-  return enableBetaFeatures()
+  return true
 }
--- a/app/src/lib/helpers/default-branch.ts
+++ b/app/src/lib/helpers/default-branch.ts
@@ -1,9 +1,22 @@
 import { getGlobalConfigValue, setGlobalConfigValue } from '../git'
 import { enableDefaultBranchSetting } from '../feature-flag'
 
+/**
+ * The default branch name that Desktop's embedded version of Git
+ * will use when initializing a new repository.
+ */
 export const DefaultBranchInGit = 'master'
+
+/**
+ * The default branch name that GitHub Desktop will use when
+ * initializing a new repository.
+ */
 export const DefaultBranchInDesktop = 'main'
 
+/**
+ * The name of the Git configuration variable which holds what
+ * branch name Git will use when initializing a new repository.
+ */
 const DefaultBranchSettingName = 'init.defaultBranch'
 
 /**
--- a/app/src/lib/helpers/default-branch.ts
+++ b/app/src/lib/helpers/default-branch.ts
@@ -23,7 +23,7 @@ const DefaultBranchSettingName = 'init.defaultBranch'
  * The branch names that Desktop shows by default as radio buttons on the
  * form that allows users to change default branch name.
  */
-export const SuggestedBranchNames: ReadonlyArray<string> = ['master', 'main']
+export const SuggestedBranchNames: ReadonlyArray<string> = ['main', 'master']
 
 /**
  * Returns the configured default branch when creating new repositories
--- a/app/src/lib/helpers/default-branch.ts
+++ b/app/src/lib/helpers/default-branch.ts
@@ -2,6 +2,7 @@ import { getGlobalConfigValue, setGlobalConfigValue } from '../git'
 import { enableDefaultBranchSetting } from '../feature-flag'
 
 export const DefaultBranchInGit = 'master'
+export const DefaultBranchInDesktop = 'main'
 
 const DefaultBranchSettingName = 'init.defaultBranch'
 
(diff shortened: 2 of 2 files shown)


rafeca (approved):
yay! 🚢

app/src/lib/helpers/default-branch.ts
```diff
@@ -1,15 +1,29 @@
...
+ * branch name Git will use when initializing a new repository.
+ */
 const DefaultBranchSettingName = 'init.defaultBranch'
 
 /**
  * The branch names that Desktop shows by default as radio buttons on the
  * form that allows users to change default branch name.
  */
-export const SuggestedBranchNames: ReadonlyArray<string> = ['master', 'main']
+export const SuggestedBranchNames: ReadonlyArray<string> = ['main', 'master']
```
say25, line 26:
```suggestion
export const SuggestedBranchNames: ReadonlyArray<string> = [DefaultBranchInDesktop, DefaultBranchInGit]
```
Not sure if this reduces "magic strings" or not

adamreisnz:
May I ask why the change from `master` to `main` as the default branch name for new repositories?

niik:
@adamreisnz GitHub selected `main` as a replacement for `master` after watching many other communities proactively rename the default branch of their existing repositories. We also like it because it’s short, it keeps your muscle memory intact, and it translates well across most languages. There’s more info in https://github.com/github/renaming, if you’d like.

adamreisnz:
Right, got it, thanks.

laurentksh:
It is also completely unnecessary, confusing for those learning git and will probably break a lot of scripts.
I'm fine with letting people change the default branch name but setting it to "main" by default to everyone is ridiculous.

adamreisnz:
You may not realise, as I ignorantly at first didn't, that the reason to do
this is not a random whim, but related to the black lives matter movement
and global calls to replace offensive wording associated with slavery and
racism (master / slave).

Change like this has to start somewhere, and I for one applaud Github for
introducing this new naming convention.

Technology changes rapidly anyway. What you read online now is going to be
outdated in 2 years. I'm sure that this particular change will not confuse
anyone for longer than 5 minutes.

laurentksh:
I know very well that this is related to the BLM movement.
But there is literally no link between the term master used in the git culture and the master / slave concept. Master in git means main, not a superior to a slave or anything like that. Also I spoke with multiple persons of colors on the internet and in real life and literally none of them wanted this change to happen, some of them even said that it was offensive.
And yes that is true, technology changes rapidly, but in general when something tech-related changes, it's backward compatible in some way.

adamreisnz:
You are of course entitled to your own opinion.

> But in general when something tech-related changes, it's backward compatible in some way.

I disagree with this statement. There is a reason semantic versioning exists to introduce breaking changes. Some open source packages that I use change major versions faster than I can keep up with. Software frameworks change every 2-3 years, introducing many breaking changes in order to drive improvement and enforce best practices of the day. Web technology is far from static, and also far from generally being backwards compatible.

Besides – how is this name change a breaking change anyway? This only applies to new repositories you create via the app, and you can toggle the setting off to set it back to master being the default name if you like?

Boca36:
#10735 ``

laurentksh:
@adamreisnz  It's a breaking change because we can't expect anymore that all newly created repos have a master branch. It's a stupid reason to be honest but i'm sure there's a lot of legacy apps that won't support that change 🤷
But yeah I used that example for the sake of arguing. To me the main problem still is that we shouldn't change something that stayed the same for 20 years for no reason. I think the neutral option that would have satisfied both sides is that master stays default, but people are allowed to change the default.

bony-0690 (commented):
#10749

apps-ms-vss-build-web-ci-designer-hub (commented):
https://github.com/desktop/desktop/pull/10527/files#diff-79c61ec51f073cf0b10003bd7b9eb190

Somnamboo:
I'll be attentive.

Somnamboo:
little worm... Yes, I thought a worm was just a emoji wormy. Meanwhile, the problem really started to increase. So, when I can Leav the

Somnamboo:
No, no... it`s nothing...

Somnamboo:
Let the bubbles be with all of as.

may365-media:
[@
```suggestion

```
**_image_**]()

app/src/lib/feature-flag.ts
```diff
@@ -138,5 +138,5 @@ export function enableDiscardLines(): boolean {
  * Should we allow to change the default branch when creating new repositories?
  */
 export function enableDefaultBranchSetting(): boolean {
-  return enableBetaFeatures()
```
B-the-vibe, line 141: `**amazing stuff, really mate, you got **### props yeah** !!!!**`

olemay:
What is that ?

olemay:
Who is that

Magicacilme1:
- [ ]

PR #10533, opened 2020-09-09
Bump registry-js to 1.12.0

niik:
## Description

[registry-js](https://github.com/desktop/registry-js) is our own package for accessing the registry on Windows without relying on `reg.exe`. This bump will give us [prebuilt binaries for electron 9](https://github.com/desktop/registry-js/pull/180), [a much smaller package](https://github.com/desktop/registry-js/pull/188), and hopefully shaves a few seconds on the Windows build time. Additionally (and as a minor secondary objective), it will allow us to upgrade a few dependencies that dependabot is complaining about.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -51,7 +51,7 @@
     "react-dom": "^16.8.4",
     "react-transition-group": "^4.4.1",
     "react-virtualized": "^9.20.0",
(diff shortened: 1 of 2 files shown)


rafeca (approved):
yay!

Issue #10549, opened 2020-09-10
Conflict details


hadeel-:
### Describe the feature or problem you’d like to solve

I use GitHub for desktop, and in previous versions, when I try to (Pull Origin) in case of conflict --> usually I see conflict files so I know how to handle them, but now after last update, I got msg about conflict but I don't know which files, I have to go to github website and check files changed in last commit to be able to find conflict files

### Proposed solution

keep it as previous versions and mention conflict files

### Additional context

Add any other context like screenshots or mockups are helpful, if applicable.
![image](https://user-images.githubusercontent.com/7837795/92754121-39948c00-f393-11ea-98eb-6924519f72e9.png)

niik:
Hi @hadeel-, thanks for opening this! I believe this is the same issue as https://github.com/desktop/desktop/issues/10441 so I'm going to close this in favor of it. If you believe I'm mistaken please let me know and we can reopen this.

PR #10550, opened 2020-09-10
Show which files are blocking a merge-like operation

### LINKED ISSUE #10441: Pulling without a clean directory doesn't show which files need to be stashed or discarded

When pulling and there are files already modified it doesn't show anymore which files need to be discarded or stashed. It now only displays a button to 'Stash and Continue'

### Proposed solution

Sometimes we don't really need to stash the whole directory for it to pull. It would be great to display which files are triggering the requested stash and continue option (similar to how it was before).

### Additional context

Reference: before the 'stash and continue' update.

niik:
Closes #10441

## Description

In #10053 we added a custom error dialog for when a pull, checkout, merge, or rebase fails due to there being changes in the working directory. In the dialog we added a quick way to stash all changes and continue the operation.

Since then we've been made aware (in #10441, and #10549) that some users depended on the raw error message from Git to know which specific files in their working directory was preventing the operation.

In this PR we parse those details out of the raw error message and incorporate them into the dialog, hopefully producing a satisfying result for all.

#### Before #10053 

image

#### In production today (2.5.5)

image

#### After this PR

image

## Release notes

Notes: [Improved] Show which uncommitted changes prevented a merge, pull, or checkout operation from completing

### DIFF
--- a/app/test/unit/parse-files-to-be-overwritten-test.ts
+++ b/app/test/unit/parse-files-to-be-overwritten-test.ts
@@ -28,4 +28,74 @@ describe('parseFilesToBeOverwritten', () => {
     expect(result.gitError).toBe(DugiteError.MergeWithLocalChanges)
     expect(parseFilesToBeOverwritten(result.stderr)).toEqual(['a', 'b'])
   })
+
+  it("isn't able to parse files from pull rebase error", async () => {
+    const parent = await setupEmptyRepository()
+    await writeFile(join(parent.path, 'a'), '1')
+    await writeFile(join(parent.path, 'b'), '2')
+    await git(['add', 'a', 'b'], parent.path, 'add')
+    await git(['commit', '-m', 'initial'], parent.path, 'add')
+
+    await writeFile(join(parent.path, 'a'), '3')
+    await writeFile(join(parent.path, 'b'), '4')
+    await git(['add', 'a', 'b'], parent.path, 'add')
+    await git(['commit', '-m', 'second'], parent.path, 'add')
+
+    const fork = await cloneLocalRepository(parent)
+    await git(['reset', 'HEAD^'], fork.path, 'reset')
+    const result = await git(['pull', '--rebase'], fork.path, 'pull', {
+      expectedErrors: new Set([DugiteError.RebaseWithLocalChanges]),
+    })
+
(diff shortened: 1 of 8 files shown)


tierninho:
So I keep getting the `In production today (2.5.5)` version on build ac41183003. 

These are my steps:
- Edit a file on local
- Edit same file on github.com
- Pull

Are these steps accurate? Thanks!

niik:
@tierninho That's weird, could you provide a log dump? The only thing I can think of is if you've configured the app to do rebase pulls because that error message doesn't contain the information we need to present the files list, only regular pull merge does that.

tierninho (approved):
Good catch! I had `pull rebase = true` in the config. 

Works now with single and multiple files, and also if an existing stash is present (although you cannot override it) ✨

app/test/unit/parse-files-to-be-overwritten-test.ts
```diff
@@ -0,0 +1,101 @@
+import { setupEmptyRepository } from '../helpers/repositories'
+import { writeFile } from 'fs-extra'
+import { join } from 'path'
+import { git } from '../../src/lib/git'
+import { cloneLocalRepository } from '../helpers/repository-scaffolding'
+import { GitError as DugiteError } from 'dugite'
+import { parseFilesToBeOverwritten } from '../../src/ui/lib/parse-files-to-be-overwritten'
+
+describe('parseFilesToBeOverwritten', () => {
+  it('parses files from pull error', async () => {
```
rafeca, line 10: These tests are awesome! Thanks for taking care of calling the actual git commands here ❤️

rafeca (approved):
This is so nice! Thanks for following up on this!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FKJ1f5iTl4Oo7u%2Fgiphy-downsized-medium.gif&token=thanks" data-gitmeme-token="thanks"></a>

app/test/unit/parse-files-to-be-overwritten-test.ts
niik: Thanks, it'd be nice if this parsing (and testing) would live in dugite instead but we'll save that for another day

PR #10551, opened 2020-09-10
Catch linting errors on Windows

niik:
## Description

In #10550 I accidentally introduced a linting error. When looking at the [failed check suite](https://github.com/desktop/desktop/pull/10550/checks?check_run_id=1098327995) I noticed that only the macOS check had failed and not the Windows one.

The Windows log showed that the error itself had been picked up

image

But it didn't fail the action step.

It turns out this is [a known limitation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference) when using the `cmd` shell on Windows to run multiple commands like we are doing for the lint step

https://github.com/desktop/desktop/blob/260d1987c31d14536f3fbb31923bdf27e2cb3979/.github/workflows/ci.yml#L49-L53

Only the exit code of the last command is considered. The most straightforward solution for us here would be to either create a new yarn script which runs those three actions chained with `&&` or to break that step into multiple steps. I've opted for the latter in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,11 +46,12 @@ jobs:
 
       - name: Install and build dependencies
         run: yarn
-      - name: Pre-build checks (Linting)
-        run: |
-          yarn lint
-          yarn validate-changelog
-          yarn check-modified
+      - name: Lint
+        run: yarn lint
+      - name: Validate changelog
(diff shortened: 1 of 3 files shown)


rafeca (approved):
Sounds fair!

PR #1972, opened 2017-06-09
Full screen ahead

donokuda:
From https://github.com/desktop/desktop/issues/1809

> Second, I think a notification similar to that of the zoom notification that just landed in #1720 could help out as well.

This PR adds a notification popup whenever a user toggles on full screen mode informing how to get out of full screen mode.

When the user is on macOS, the keys displayed are `⌃⌘F` and on Windows, the keys displayed are `F11`.

#### Couple of gifs demonstrating this feature:

**Windows**
full-screen

**macOS**
output

@niik I'm having an issue trying to get the CSS animation to work when the notification is supposed to go away. Currently it just disappears all at once. I'm open to any suggestions on how to get that working.

PR #10554, opened 2020-09-10
Toggle Fullscreen Info Popup

### LINKED ISSUE #7916: Exit fullscreen message is displayed too often

I run GitHub Desktop maximised, and since the last update whenever I switch back to the app it displays the "Press [⌃⌘F] to exit fullscreen" message for a few seconds.

image

Only display the message once, when entering fullscreen.

Sam-Spencer:
Closes #7916

## Description

- Users now have the option to switch OFF the `Press [key command] to exit fullscreen` popup that appears when:
  - Expanding GitHub to fullscreen
  - and when macOS users switch back to fullscreen GitHub from Mission Control
 - The preference is ON by default (the current behavior)
 - Opt-out from the Preferences -> Appearance menu. 
 - @tierninho explained in issue #7916 that a viable solution to the issues presented by others in that discussion would be preference to disable this feature. This implementation attempts to address the issue with one of the recommended solutions. This PR may also be of interest to @ampinsk.
 - Preferences copy may need to be tweaked for clearer understanding of what this toggle does

### Screenshots

Screen Shot 2020-09-10 at 3 57 52 PM

## Release notes

Notes: [Added] Option in Appearance Preferences to disable the fullscreen info popup.

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -216,6 +216,9 @@ export interface IAppState {
   /** Whether we should automatically change the currently selected appearance (aka theme) */
   readonly automaticallySwitchTheme: boolean
 
+  /** Whether we should display a toast notification when the user enters fullscreen */
+  readonly displayFullscreenInfoToast: boolean
+
   /**
    * A map keyed on a user account (GitHub.com or GitHub Enterprise)
    * containing an object with repositories that the authenticated
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -72,6 +72,8 @@ import {
   setPersistedTheme,
   getAutoSwitchPersistedTheme,
   setAutoSwitchPersistedTheme,
+  getDisplayFullscreenInfoToast,
+  setDisplayFullscreenInfoToast,
 } from '../../ui/lib/application-theme'
 import {
   getAppMenu,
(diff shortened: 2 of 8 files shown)


niik:
Hey @Sam-Spencer, thanks so much for opening this and wanting to contribute. I took a look at your code and it looks great, you did a splendid job of keeping with the conventions of the code base and I especially appreciate your efforts in documenting variables and functions.

That said I think we need to go back to first principles here and ask ourselves where the problem really lies. GitHub Desktop puts great effort into being deliberate about what preferences we surface to users and whenever possible we try to avoid the user having to make a choice. In this case I believe adding a preference isn't the best solution as the core behavior is wrong. I think we'd be better of fixing the underlying issue.

I very rarely use the app in full screen which is why I probably hadn't noticed the issue before but now that I see it I agree that it's annoying. When we initially implemented the full screen prompt in https://github.com/desktop/desktop/pull/1972 this wasn't the case so something must have changed since. Looking at what happens when going in and out of mission control or when switching between virtual desktops the window state toggles between `full-screen` and `hidden`.

This is due to our decision early on to merge a window's visibility state with the window state in https://github.com/desktop/desktop/blob/260d1987c31d14536f3fbb31923bdf27e2cb3979/app/src/lib/window-state.ts#L11-L23

For the purposes of the full screen notification we don't care about the visibility of the window though, just whether the window was full-screened or not. As such I think the solution here is to ignore transitions from any window state to 'hidden' and only look at transitions between minimized, normal, maximized, and full-screen.

I've opened #10561 and made you co-author of it, I hope that's alright with you. Please take a look at that PR and let me know what you think. I believe it's the best approach here. I'm going to let #10561 supersede this PR and I hope that doesn't discourage you from contributing in the future.

Sam-Spencer:
Thank you so much for the generous feedback @niik. I really appreciate it. I agree with you that, from a design perspective, the user shouldn't need to deal with excessive preferences. Addressing the root of the problem makes much more sense. I'll take a look at #10561 and comment further there! Thanks again!

PR #10561, opened 2020-09-11
Only show full screen toast notification when making the app window full-screen

### LINKED ISSUE #7916: Exit fullscreen message is displayed too often

niik:
Closes #7916 

## Description

@Sam-spencer opened #10554 and made me aware of a macOS bug where we show the full screen notification every time the user switches back to the app while it's in full screen mode.

image

When looking at the window state changes that happen when switching back to a full-screened app on macOS we see the following

user full-screens the app: `normal -> hidden ->  full-screen`
user moves focus to another app: `full-screen -> hidden`
user moves focus back to desktop: `hidden -> full-screen`.

This is due to our decision early on to merge a window's visibility state with the window state in https://github.com/desktop/desktop/blob/260d1987c31d14536f3fbb31923bdf27e2cb3979/app/src/lib/window-state.ts#L11-L23

For the purposes of the full screen notification we don't care about the visibility of the window though, just whether the window was full-screened or not. As such I think the solution here is to ignore transitions from any window state to 'hidden' and only look at transitions between minimized, normal, maximized, and full-screen.

This PR makes the `FullScreenInfo` component ignore the `hidden` state and I've also taken the time to modernize it to remove the deprecated `componentWillReceiveProps` lifecycle method and favoring `getDerivedStateFromProps` (to react to the prop change) and `componentDidUpdate` to schedule the timers to remove the notification after a set time.

## Release notes

Notes: [Fixed] Full screen notification would show on macOS when focusing an already full-screen Window

### DIFF
--- a/app/src/ui/window/full-screen-info.tsx
+++ b/app/src/ui/window/full-screen-info.tsx
@@ -42,7 +42,11 @@ export class FullScreenInfo extends React.Component<
     }
 
     if (state.computedWindowState !== props.windowState) {
-      return { computedWindowState: props.windowState }
+      return {
+        computedWindowState: props.windowState,
+        renderInfo: props.windowState === 'full-screen',
+        renderTransitionGroup: props.windowState === 'full-screen',
+      }
     }
 
     return null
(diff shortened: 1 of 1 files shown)


tierninho (approved):
Use case changes look good. Confirming I no longer see the notification when coming out of a hidden state.

Sam-Spencer:
Great! This looks like the right fix for the issue reported in #7916. It seems to work properly on my MacBook in the scenarios I previously saw the bug.

rafeca (approved):
💅

hubgit:
I've just tried this in the beta release (2.5.6-beta2, on macOS 10.15.6) and it's much better now - if the app window is maximised after launching the notification is only displayed once.

However there's an issue if the app is launched with the window already maximised - the notification is displayed and never disappears!

niik:
thank you so much for pointing that out @hubgit, I've opened https://github.com/desktop/desktop/pull/10644 which should fix the problem and we'll make sure that's included in the next beta release as well as the next production release.

PR #10588, opened 2020-09-16
Repository indicator optimizations, round two

niik:
## Description

This is a follow-up to #10119 and a precursor to being able to disable background updates for repository indicators.

Prior to this PR the repository indicator updater would call `AppStore.performFetch` when fetching repositories. `performFetch` does a fair bit more work than just a fetch, such as a full repository refresh, branch protection updates, and branch fast forwards.

https://github.com/desktop/desktop/blob/5f2038cac8b3ba26f31790d46089afa3b6b2decd/app/src/lib/stores/app-store.ts#L4179-L4192

Since all we need for the repository indicators is an ahead/behind status for the current branch (which we'll get from both [status](https://github.com/desktop/desktop/blob/5f2038cac8b3ba26f31790d46089afa3b6b2decd/app/src/lib/stores/git-store.ts#L1050), and from [running fetch](https://github.com/desktop/desktop/blob/5f2038cac8b3ba26f31790d46089afa3b6b2decd/app/src/lib/stores/git-store.ts#L951)) we can take a shortcut here. We're never run repository indicator refreshes on the currently selected repository and we'll run a full repository refresh when a repository is selected so we're not missing anything.

The second change here is about updating the repository indicator for a repository as soon as we get the results back from `status` instead of waiting for the fetch. That's just a nice little ux improvement.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2841,7 +2841,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
     )
 
     if (await this.shouldBackgroundFetch(repository, lastPush)) {
-      await this._fetch(repository, FetchType.BackgroundTask)
+      await this.withAuthenticatingUser(repository, (repository, account) => {
+        const isBackgroundTask = true
+
+        return this.withPushPullFetch(repository, account, () =>
+          gitStore.fetch(account, isBackgroundTask, p =>
+            this.updatePushPullFetchProgress(repository, p)
+          )
+        )
+      })
     }
 
     lookup.set(repository.id, {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2834,18 +2834,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
+    this.updateSidebarIndicator(repository, status)
+    this.emitUpdate()
+
     const lastPush = await inferLastPushForRepository(
       this.accounts,
       gitStore,
       repository
     )
     const shouldFetch = await this.shouldBackgroundFetch(repository, lastPush)
-
-    lookup.set(repository.id, {
-      aheadBehind: gitStore.aheadBehind,
-      changedFilesCount: status.workingDirectory.files.length,
-    })
-    this.emitUpdate()
 
     if (!shouldFetch) {
       return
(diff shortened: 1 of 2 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -3750,7 +3778,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
   private async withPushPullFetch(
     repository: Repository,
-    account: IGitAccount | null,
```
niik, line 3753: I just stumbled upon the fact that this parameter was unused so I removed it which let me keep a newly introduced call to this method on one line. It's unrelated to this PR but it's small enough that I'm hoping we can just sneak it in.

app/src/lib/stores/app-store.ts
```diff
@@ -3785,7 +3812,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     repository: Repository,
     account: IGitAccount | null
   ): Promise<void> {
-    return this.withPushPullFetch(repository, account, async () => {
+    return this.withPushPullFetch(repository, async () => {
```
niik, line 3815: Removes the unused `account` parameter. No logical difference

app/src/lib/stores/app-store.ts
```diff
@@ -4140,7 +4167,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     fetchType: FetchType,
     remotes?: IRemote[]
   ): Promise<void> {
-    await this.withPushPullFetch(repository, account, async () => {
+    await this.withPushPullFetch(repository, async () => {
```
niik, line 4170: Removes the unused `account` parameter. No logical difference

app/src/lib/stores/git-store.ts
```diff
@@ -949,9 +949,14 @@ export class GitStore extends BaseStore {
...
         this._aheadBehind = await getAheadBehind(this.repository, range)
-        this.emitUpdate()
+      } else {
+        this._aheadBehind = null
       }
+    } else {
+      this._aheadBehind = null
     }
+
+    this.emitUpdate()
```
niik, line 959: We haven't run into any problems with not resetting `aheadBehind` in fetch before since all fetches are followed by a `status` call which itself refreshes (and resets) the `aheadBehind` but now that we're running fetch stand-alone it makes sense for us to be more diligent here.

rafeca (approved):
LGTM!

rafeca (approved):

PR #10590, opened 2020-09-17
Update rebase progress parser to parse output from the merge rebase backend

niik:
## Description

In #10077 we upgraded to Git v2.26.2 in which the default rebase backend changed from `apply` to `merge`. Unbeknownst to us at the time was that switching the backend would also affect rebase progress parsing. The result of which meant that our rebase dialogs were stuck on the initial commit for the duration of the rebase.

image
 
Unfortunately there weren't any tests that would have caught this so I've taken the time to add some which hopefully lets us catch any change in behavior in the rebase space in the future.

## Release notes

Notes: [Fixed] Show each step and the overall progress in while applying a rebase

### DIFF
--- a/app/src/models/popup.ts
+++ b/app/src/models/popup.ts
@@ -65,7 +65,6 @@ export enum PopupType {
   CreateTag,
   DeleteTag,
   LocalChangesOverwritten,
-  RebaseConflicts,
   ChooseForkSettings,
   ConfirmDiscardSelection,
 }
--- a/app/test/unit/git/rebase/progress-test.ts
+++ b/app/test/unit/git/rebase/progress-test.ts
@@ -1,7 +1,6 @@
 import {
   IStatusResult,
   continueRebase,
-  git,
   getStatus,
 } from '../../../../src/lib/git'
 import {
--- a/app/test/unit/git/rebase/progress-test.ts
+++ b/app/test/unit/git/rebase/progress-test.ts
@@ -25,8 +25,7 @@ const featureBranchName = 'this-is-a-feature'
 describe('git/rebase', () => {
   describe('skips a normal repository', () => {
     it('returns null for rebase progress', async () => {
-      const repository = await setupEmptyDirectory()
-
+      const repository = setupEmptyDirectory()
       const progress = await getRebaseSnapshot(repository)
 
       expect(progress).toEqual(null)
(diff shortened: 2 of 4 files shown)


app/src/models/popup.ts
```diff
@@ -65,7 +65,6 @@ export enum PopupType {
   CreateTag,
   DeleteTag,
   LocalChangesOverwritten,
-  RebaseConflicts,
```
niik, line 68: This is unrelated to this PR but I stumbled upon it trying to find the right dialog and it confused the heck out of me so I figured we should just remove it

app/test/unit/git/rebase/progress-test.ts
```diff
@@ -100,6 +127,38 @@ describe('git/rebase', () => {
...
+
+      expect(progress.length).toEqual(10)
+      expect(progress[9]).toEqual({
+        currentCommitSummary: 'Feature Branch Tenth Commit!',
+        kind: 'rebase',
+        rebasedCommitCount: 10,
+        title: 'Rebasing commit 10 of 10 commits',
+        totalCommitCount: 10,
+        value: 1,
+      })
```
rafeca, line 159: Something I find quite useful on situations where there is a large object to assert is to use [jest snapshots](https://jestjs.io/docs/en/snapshot-testing).

There's some controversy around jest snapshots and I think it's relatively easy to abuse them (e.g by having huge snapshots with tons of data that can't easily be reviewed and changes often), but when I saw this assertion I thought about them :)

To be clear, I'm not suggesting changing the PR to use them here (there are no prior art of snapshots in Desktop so probably there needs to be a more in-depth discussion/investigation before introducing them), just want to showcase them just in case you don't know about them already 😃

rafeca (approved):
Ups, didn't catch this when I worked on the git upgrade 😳 Thanks for fixing that!

tierninho (approved):
Confirming that rebases were successful in OSX/WIN with and without merge conflicts.

app/test/unit/git/rebase/progress-test.ts
niik: Yeah, I did consider snapshots here and I have nothing against them personally but I opted not to for the exact reason you mentioned, i.e. lack of prior art.

PR #10593, opened 2020-09-17
Support disabling repository indicators

niik:
~~🌵 This is based on #10588, please review and merge that first 🌵~~

## Description

This is the final piece (for now) of the repository indicator work (see #10119, #10588) aimed to reduce the resource consumption of background updates in the app.

Background updates of repository indicators powers the repository list feature to at-a-glance see whether your repositories have any uncommitted changes or if they're out of sync with their remote without having to select each repository (see https://github.com/desktop/desktop/pull/4770) for more details on that. In the code base we call this feature **repository indicators**

This used to check each repository every 15 minutes to see if it the time since it hadn't been fetched in the last 30 minutes. Additionally it would check the GitHub.com API to see if anything had been pushed to the repository since the last time the repository was fetched in order to minimize the amount of times the repository was fetched.

Regardless of whether the repository was fetched or ot it would be refreshed (which, among other things, includes running a `git status`). For users who had a lot of tracked repositories or repositories where running `git status` was prohibitively expensive this could cause significant CPU usage.

In this PR we're adding the ability to disable background updates of repository indicators. We'll still show the last known status in the repository list *if the user has selected the repository in the past* but we will not automatically update the status without the user selecting the repository.

#### The setting in preferences
image

#### The new preferences tab

In order to make room in the advanced tab for this setting something had to give. I chose to move the confirmation prompts settings out of advanced and into their own 'Prompts' tab. I figured they're not really "advanced" features, it's more that we didn't have another place to put them in the past. cc @ampinsk on thoughts here. The new prompts tab contents look a little sad but I think we can live with that for the time being.

image

## Release notes

Notes: [Added] Periodic background fetch and status updates for all repositories can now be disabled in the Advanced section of Preferences/Options.


tierninho:
@niik Is there a way to trigger the background fetch job instead of waiting around for it to execute? Thanks.

niik:
> @niik Is there a way to trigger the background fetch job instead of waiting around for it to execute? Thanks.

Hmm.. Not exactly but if you've disabled it and then restart the app it will run soon after (max 30 seconds) turning it back on and saving it again. You can filter the developer tools console for `RepositoryIndicatorUpdater` to see when it starts.

app/src/ui/preferences/advanced.tsx
```diff
@@ -134,32 +103,21 @@ export class Advanced extends React.Component<
...
-            value={
-              this.state.confirmForcePush ? CheckboxValue.On : CheckboxValue.Off
-            }
-            onChange={this.onConfirmForcePushChanged}
+            onChange={this.onRepositoryIndicatorsEnabledChanged}
           />
+          <p className="git-settings-description">
+            Background updates allows Desktop to highlight repositories that
+            have uncommitted changes or are out of sync with their remote in the
+            repository list. May adversely affect performance.
```
rafeca, line 119: Not 100% sure about the "May adversely affect performance." sentence: while it's true it feels strange to have this disclaimer for a setting that's enabled by default.

I'd rather omit this until the setting is changed to be disabled by default (if that's the plan), or turn it around (e.g "Disabling this may improve performance when there are a lot of repositories to track").

rafeca (dismissed):
Added one comment about the description, but I don't think it's blocking for merging this (or even releasing a beta)

niik:
> Added one comment about the description, but I don't think it's blocking for merging this (or even releasing a beta)

Good comment, I'll noodle on it. I propose we see if @ampinsk has time to look at that paragraph and the other design changes today and if she doesn't well go ahead and get it to beta and adjust in a follow-up PR

ampinsk:
Hey, sorry for the delay just saw this! We could address this in a follow up but here's my loose thoughts: 

It feels a little odd to me having a "Prompts" category, I'm not sure as a user I'd understand what was in there. I'm honestly not sure what I'd do otherwise without some more thought, so if we do have this section, I think it makes sense for the "Switching branches" preference to go in there since that also has to do with prompting. 

I'm also curious if we could cut down the length of the explanatory copy for background updates, I can take a stab at it!

app/src/ui/preferences/advanced.tsx
niik: I dig it, it didn't make any difference in terms of vertical space consumed (it was still three lines) but removing the leading `This` got it down to two lines

image

niik:
> I think it makes sense for the "Switching branches" preference to go in there since that also has to do with prompting.

Maybe? What is it more, an advanced preference or a preference related to prompting? I don't feel like there's a clear cut answer here. I think the stash preference is perhaps the most fitting part of the advanced tab because it's a setting where the default makes the most amount of sense and only people with specific workflows should tweak it and I feel like its place hints at that.

The prompts section is weird but they never really belonged in 'Advanced' either IMO, it was just an easy place to dump them. I've seen some apps who doesn't allow controlling individual confirmation prompts and instead have a `Reset all confirmation dialogs` action in preferences. If we're aiming for discoverability then the usage section should live under a privacy tab and the stash preference under a `Stash` tab or perhaps even under the `Git` tab.

I'm thinking out loud here, I don't feel like I have the answer but I agree that the prompts section feels a bit weird (though mostly the contents of it).

@ampinsk Do you have a strong preference on the path forward?

ampinsk:
All fair points @niik. I'm fine to keep it as is in this PR and we can revisit as we use it more or as we add new things to preferences 👍

niik:
> All fair points @niik. I'm fine to keep it as is in this PR and we can revisit as we use it more or as we add new things to preferences 👍

Sounds good, mind writing your opinion in stone with a ✅ PR review?

ampinsk (approved):
Looks great! I'm a little iffy on the `Prompt` section but happy to ship as is and look at it again later.

koenpunt:
Would be great if there's a similar setting added for the background fetcher! (https://github.com/desktop/desktop/issues/1128)

niik:
@koenpunt ha, what a coincidence, I was just writing up a response in that issue as  well as opening a new issue specifically dealing with that, see #10687

PR #9382, opened 2020-03-27
Unify logic to checkout a pull request

rafeca:
Closes #9439
Closes #9593

This PR is based off https://github.com/desktop/desktop/pull/9340

## Description

This PR contains a small refactor to unify the logic to checkout a PR from GitHub.

Currently, there are two ways to checkout a PR:

1. From the "Pull Requests" list tab that appears on the GitHub Desktop app.
    Screenshot 2020-03-27 at 15 42 47
2. From github.com when navigating to a Pull Request page:
    Screenshot 2020-03-27 at 15 43 15

Each of these entry points has its own logic, which causes some inconsistencies between them (as we found in https://github.com/desktop/desktop/pull/9340#issuecomment-603953203).

## Test cases

I've tried to follow the test cases that were written in https://github.com/desktop/desktop/pull/9138, but couldn't check every single one of them.

## Release notes

Notes: no-notes

PR #10598, opened 2020-09-18
Reuse local branch if already checked out when checking out a PR

niik:
## Description

I was trying to check out one of my own PRs (#10593) in Desktop just now and noticed some funky behavior. Instead of simply checking out the local branch that I had used to create the PR in the first place desktop created a new fork-like remote (but with the same URL as desktop/desktop) and checked out a new branch using the `pr/number` format that we use for checking out contributor PRs.

The problem was that when attempting to locate the remote branch in `_getPullRequestHeadBranchInRepo` Desktop only looked for branches whose name exactly matched `origin/disable-repository-indicators` but since I had already checked that branch out there's no such branch in `gitStore.allBranches`, instead there's the local branch `disable-repository-indicators` which has an `.upstream` property of `origin/disable-repository-indicators`.

Likely related to #9382. Note that this regression luckily hasn't made it to production yet, but it does exist on the beta channel.

## Release notes

Notes: [Fixed] Checking out a PR attempts to locate an existing branch tracking the PR base branch

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5504,7 +5504,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const gitStore = this.gitStoreCache.get(repository)
 
     // Find a remote branch matching the given name or a local branch
-    // whose upstream tracking branch matches the given name (i.e someon
+    // whose upstream tracking branch matches the given name (i.e someone
     // has already checked out the remote branch)
     const findBranch = (name: string) =>
       gitStore.allBranches.find(branch =>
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -5503,8 +5503,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const gitStore = this.gitStoreCache.get(repository)
 
+    // Find a remote branch matching the given name or a local branch
+    // whose upstream tracking branch matches the given name (i.e someon
```
rafeca, line 5507:
```suggestion
    // whose upstream tracking branch matches the given name (i.e someone
```

rafeca (approved):
omg! 🤦‍♂️ good catch!!

rafeca (approved):

PR #10599, opened 2020-09-18
Remove duplicated logic in PR checkout logic

niik:
~~🌵 Based on #10598, please review that first 🌵~~

## Description

When opening #10598 I spotted [this comment of mine](https://github.com/desktop/desktop/pull/9382#pullrequestreview-476640503) in the original PR for unifying the PR logic:

> I've got [..] some changes that I've made as I've been reviewing that I intend to open as a separate PR once this has merged so that we can review them on their own.

Turns out I had a branch lying around that I never got around to pushing so here it is. Rebased on top of #10598. 

The diff may look a bit intimidating but I recommend disabling whitespace changes to make it a bit easier to see what's really changed. The motivation behind this work was to remove the duplicated logic in `_getPullRequestHeadBranchInRepo` by refactoring that into `findPullRequestHeadInRemote`. In order to do so I also had to expose the `upstreamRemote` property on the `GitStore`.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5551,8 +5551,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
       remote = gitStore.upstreamRemote
     }
 
-    // If we don't have a remote here, it's probably going
-    // to just crash and burn on checkout, but that's okay
     if (remote !== null) {
       return this.findPullRequestHeadInRemote(repository, remote, headRefName)
     }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -268,10 +268,7 @@ import { parseRemote } from '../../lib/remote-parsing'
 import { createTutorialRepository } from './helpers/create-tutorial-repository'
 import { sendNonFatalException } from '../helpers/non-fatal-exception'
 import { getDefaultDir } from '../../ui/lib/default-dir'
-import {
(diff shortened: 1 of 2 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -5541,12 +5539,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+    // Determine whether the ref is in the current repository or the parent
+    if (headCloneURL === cloneURL) {
+      remote = gitStore.defaultRemote
+    } else if (headCloneURL === parent?.cloneURL) {
+      remote = gitStore.upstreamRemote
+    }
+
+    // If we don't have a remote here, it's probably going
+    // to just crash and burn on checkout, but that's okay
```
rafeca, line 5555: Maybe I'm missing something, but if I understand the logic correctly having a null `remote` here is a possible scenario: it'll just trigger the logic in L5504 where it tries to find the fork remote repo or add it to the current repo.

In that case, we can get rid of this comment:

```suggestion
```

rafeca (approved):
Yay, thanks for cleaning this up, much better and easier to understand now!

app/src/lib/stores/app-store.ts
niik: Yeah, this comment came from the earlier logic and was specifically related to not having a _default_ remote. I'm 👍 on removing it.

rafeca (approved):

PR #10532, opened 2020-09-09
Bump electron to 9.3.1

niik:
Closes #10267 

## Description

Electron 9.3.0 contains [the fix for shell.moveItemToTrash](https://github.com/electron/electron/pull/25124) that's been blocking #8333 from getting merged as well as a [number of other fixes](https://www.electronjs.org/releases/stable?version=9). No new functionality though, this is a patch release for the 9 series of Electron and hopefully doesn't contain any surprises. I've done a quick review of the changes from 9.1.2 to 9.3.0 and I don't see anything scary. The plan is to get this onto beta soon-ish to get some mileage on it before the next production release.

PR #10601, opened 2020-09-18
Electron 9.3.1

say25:
## Description

- Something we missed in #10532

## Release notes

Notes: no-notes


niik (approved):
Whoops, great catch @say25, thank you!

PR #10602, opened 2020-09-18
Typo fixes and some minor readability tweaks

say25:
## Description

- Split from #6818, minor cleanup I did while implementing

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -242,9 +242,9 @@ const defaultEditorOptions: IEditorConfigurationExtra = {
     // custom search UI.
     [__DARWIN__ ? 'Cmd-F' : 'Ctrl-F']: showSearch,
     // Disable all other search-related shortcuts so that they
-    // don't interfer with global app shortcuts.
     [__DARWIN__ ? 'Cmd-G' : 'Ctrl-G']: false, // findNext
     [__DARWIN__ ? 'Shift-Cmd-G' : 'Shift-Ctrl-G']: false, // findPrev
+    // don't interfere with global app shortcuts.
     [__DARWIN__ ? 'Cmd-Alt-F' : 'Shift-Ctrl-F']: false, // replace
     [__DARWIN__ ? 'Shift-Cmd-Alt-F' : 'Shift-Ctrl-R']: false, // replaceAll
     Down: scrollEditorVertically(1, 'line'),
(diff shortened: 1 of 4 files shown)


app/src/ui/diff/text-diff.tsx
```diff
@@ -240,11 +240,11 @@ const defaultEditorOptions: IEditorConfigurationExtra = {
...
     // custom search UI.
-    [__DARWIN__ ? 'Cmd-F' : 'Ctrl-F']: showSearch,
+    'CmdOrCtrl+F': showSearch,
     // Disable all other search-related shortcuts so that they
-    // don't interfer with global app shortcuts.
-    [__DARWIN__ ? 'Cmd-G' : 'Ctrl-G']: false, // findNext
-    [__DARWIN__ ? 'Shift-Cmd-G' : 'Shift-Ctrl-G']: false, // findPrev
+    // don't interfere with global app shortcuts.
+    'CmdOrCtrl+G': false, // findNext
+    'Shift+CmdOrCtrl+G': false, // findPrev
```
niik, line 247: I don't believe any of these changes produce the desired results. To my knowledge CodeMirror doesn't support `CmdOrCtrl`, that's an electron/desktop-ism. We rely on CodeMirror's [default keymap](https://github.com/codemirror/CodeMirror/blob/c74a1cafc01a7e34af1b19dd4c82ff821c2e1442/src/input/keymap.js#L19-L47) so the entries in `extraKeys` need to match exactly what's in the `pc` or `mac` keymap, hence the inline conditionals.

I think we need to revert these changes save for the typo.

niik (changes requested):
Thanks for extracting this work @say25, I just had one request to remove the codemirror changes.

app/src/ui/diff/text-diff.tsx
```diff
@@ -242,9 +242,9 @@ const defaultEditorOptions: IEditorConfigurationExtra = {
     // custom search UI.
     [__DARWIN__ ? 'Cmd-F' : 'Ctrl-F']: showSearch,
     // Disable all other search-related shortcuts so that they
-    // don't interfer with global app shortcuts.
     [__DARWIN__ ? 'Cmd-G' : 'Ctrl-G']: false, // findNext
     [__DARWIN__ ? 'Shift-Cmd-G' : 'Shift-Ctrl-G']: false, // findPrev
+    // don't interfere with global app shortcuts.
```
niik, line 247: Seems this comment got accidentally split up when removing the changes here

niik (changes requested):
Just a tiny comment that needs to be moved

niik (approved):
Thanks @say25!

PR #10617, opened 2020-09-22
Side by side diffs

rafeca:
## Description

This is an initial implementation of side by side diffs on GitHub Desktop.

## UX considerations

I've been heavily inspired on the dotcom design/UX to implement this feature, trying to mimic it behaviour as much as possible. This includes a few considerations:

* We only highlight the differences on modified lines when the number of added and deleted lines in a hunk is the same. As an example, in the next screenshot the first hunk has diff highlighting while the second one does not have it:
  image
* I've tried to follow as closely as possible the colors/background/borders of dotcom for the different elements (line number boxes, empty spaces, hunk headers, etc) on light mode. Since there's no dark mode in GitHub, I had to go a bit on my own there 🙃 .

In some cases, in order to reuse already existing logic in Desktop, the behaviour is a bit different than in GitHub.com. The most visible instance of this is differences highlighting, where the logic that we use in Desktop differs a bit from the dotcom one (I think that the one in GitHub.com is slightly better, but I don't consider it a big issue to worry about for now):

Desktop:
image

GitHub.com:
image

### Partial committing

In order to have a nicer interface for partial committing, I've decided to move the line numbers of the left pane to the right, this way we have both columns of line numbers together.

This is quite unorthodox, but it solves the problem of selecting multiple lines by drag/drop plus it allows us to use the spacing between the two columns as a gutter that allows selecting the whole hunk (in a kind of similar way as it's done for unified diffs).

I think I like this approach, but I would love to have a more conventional way to solve it if it exists (it's also strange to have the line numbers in the middle in one view and on the left on another). Any suggestion/idea that I haven't considered is welcome here :) 

demo

## Technical considerations

The meaty part of the implementation of this feature is in the `getDiffRowsFromHunk()` method: this function converts a `DiffHunk` object (that contains the different lines of the diff for that hunk) into an array of `DiffRow` objects (which is a new object that represents a line that gets displayed on screen).

This conversion is done by merging groups of consecutive deleted/added lines together in the same `DiffRow` so they can get displayed side by side. Other types of lines are directly converted from a `DiffLine` object to a `DiffRow` object.

Ideally, I'd like to not have `DiffRow` objects, but I found that by creating them with a much more specific TypeScript types than `DiffLine`, the rendering of rows (done in `<SideBySideDiffRow/>` gets quite simpler. A follow-up could be to make the `DiffLine` objects more strict in terms of typing (no nullable fields) which would allow to remove the conversion from `DiffRow` to `DiffLine` objects.

Not using CodeMirror allows us to have more control on the rendering, which makes it much easier to handle click interactions on the lines (e.g we can pass a hunk identifier to each row so we can easily know which hunk to select when clicking on the hunk handler, instead of having to calculate that based on the line clicked).

Also, I could appreciate a performance improvement when quickly scrolling through long diffs: it's quite possible that the overhead that CodeMirror adds is superior to the overhead of React (this is more appreciable when building in prod mode, since in dev mode side by side diffs are penalized by React debugging logic):

Unified diffs (48ms spent in scroll handler):
Screen Shot 2020-09-02 at 20 26 05

Side By Side diffs (14ms spent in scroll handler):
Screen Shot 2020-09-02 at 20 26 48

## Next steps

There are a few things that I haven't implemented yet:

1. Auto-scroll the screen when moving the cursor up/down while selecting changes to allow selecting lines that are not shown in the view.

A future next step that I think can be done once/if this PR is merged is to implement unified diffs re-using this same logic. I've taken a quick spike on it and the implementation is relatively straightforward (I only had to change one line in `<SideBySideDiff />` to prevent creating modified rows, but `<SideBySideDiffRow />` and the css needed a few more changes to adjust the layout accordingly).

With this, we could use split diffs to stress test this diff viewer that does not have CodeMirror, and once is stable enough we could enable it for unified diffs behind a feature flag to end up replacing completely the current diff viewer.

## Screenshots

image

image

image

## Release notes

Notes: [New] Side by side diffs


say25:
This seems like it would collide with https://github.com/desktop/desktop/pull/6818 :( I don't mind rebasing that branch after this work is done.

tierninho:
Some notes while testing this:

- Currently, the History tab has the flag to toggle this feature on/off. It seems like it should be on the Changes tab too for obvious reasons.
- Should we add some minimal limitation on how narrow you can make this pane, else we end up with this?
Screen Shot 2020-09-22 at 9 14 19 AM
- In the split mode, under Changes, I cannot discard multiple lines like so:
Screen Shot 2020-09-22 at 10 26 21 AM

niik (changes requested):
I'm slowly working my way through this. Overall I think it's fantastic as I've already said in our internal review of this. I'll have some more nits and questions but for now I happened to note that the side by side diffs doesn't contain the diff line type indicators (i.e. `+`, `-`, or ` `).

### Side by side
image

### GitHub.com
image

I think this is because the side by side diff component uses the `.content` property of DiffLine

https://github.com/desktop/desktop/blob/d0b745a9931fb474f240e1fd4edecf378a0696df/app/src/ui/diff/side-by-side-diff.tsx#L579

Which explicitly omits the commit line type marker

https://github.com/desktop/desktop/blob/d0b745a9931fb474f240e1fd4edecf378a0696df/app/src/models/diff/diff-line.ts#L33-L36

rafeca:
> Currently, the History tab has the flag to toggle this feature on/off. It seems like it should be on the Changes tab too for obvious reasons.

Yup, completely agree. I can add the toggle on the right of the topbar in a similar way that @say25 did in https://github.com/desktop/desktop/pull/6818 for the whitespace.

Something worth discussing is how we're going to surface this whole new diff viewer... I think we could have it as some kind of experimental feature (and only show the toggle then when the "experimental diff viewer" is enabled), but maybe there are better options.

> Should we add some minimal limitation on how narrow you can make this pane, else we end up with this?

Good point! I think this would be nice to have, specially since the side by side mode needs more horizontal real estate, but I don't think it's a blocker (given the fact that we have a similar problem currently with normal diffs).

> In the split mode, under Changes, I cannot discard multiple lines like so:

I'm not able to reproduce it, when clicking on the gutter between the line numbers I'm able to discard them (maybe it's confusing that the gutter is in the middle?):

image

> I happened to note that the side by side diffs doesn't contain the diff line type indicators (i.e. +, -, or ).

Good catch! I'm thinking about adding them manually so I can make them non-selectable by the cursor (in a similar way that dotcom does).

tierninho:
> I'm not able to reproduce it, when clicking on the gutter between the line numbers I'm able to discard them (maybe it's confusing that the gutter is in the middle?):

Ah, I see it now. Thanks for the reminder.

rafeca:
> I happened to note that the side by side diffs doesn't contain the diff line type indicators (i.e. +, -, or ).

I've added the add/delete indicators in https://github.com/desktop/desktop/commit/2e10ce4f7dcae85f57bab0e0cd0fc36fa13ff336, which is not part of this PR because I did this on top of the `diff-engine-everywhere` branch which changes a bit the CSS of diffs.

I can merge that branch into this one if this makes the review easier.

cprovatas:
So Exciting!!!! Thank you @rafeca!

JosephTLyons:
Extremely stoked for this.  I can't test this myself but I'm optimistically hoping it gets added in!

tierninho:
> I can merge that branch into this one if this makes the review easier.

At least for me, 👍 

I noticed the gutters are half the normal width when split view in enabled. Very hard to click that thin line -- they should both be doubled in size to match current gutters.

Screen Shot 2020-09-24 at 2 40 44 PM

Optional: We could add a right-click context menu in the diff header (on the Changes view side) that enables the user to turn on/off split view. This would remove the need for an awkward checkbox. UI for History tab remains as-is.

rafeca:
Thanks for the feedback @tierninho !

> I noticed the gutters are half the normal width when split view in enabled. Very hard to click that thin line -- they should both be doubled in size to match current gutters.

Some of the gutters of split diffs are 4px wide (e.g the one in the top of the image you pasted) while some of them are 2px wide. I'm going to try to make all of them 4px wide which is also the width of the current gutters.

Regarding clicking on gutters, the click area of them is much larger (20x20 pixels), which is similar to the current behaviour:

image

niik:
> I can merge that branch into this one if this makes the review easier.

I took a quick look at the diff of that branch and in general I'd be in favor of pulling it into this but it appeared to me as if that branch made the codemirror-based diff viewer obsolete in its entirety. I think what I'd like to see here is a feature flag to enable the side by side diffs like we have in this PR and then an additional feature flag for whether to use the new diff engine for unified diffs.

It's your call if you want to update that branch with those changes or only port the CSS for the line type indicators over here

rafeca:
> I took a quick look at the diff of that branch and in general I'd be in favor of pulling it into this but it appeared to me as if that branch made the codemirror-based diff viewer obsolete in its entirety. I think what I'd like to see here is a feature flag to enable the side by side diffs like we have in this PR and then an additional feature flag for whether to use the new diff engine for unified diffs.

👍 done that and merged that branch into this one!

Some stuff I improved since yesterday:

- Improved context menus to make them look the same as the current diff viewer (same labels to discard lines and added a context menu to copy text).
- Added the +/- signs next to added/removed lines.
- Improved the gutters to select lines/hunk for partial committing (now they're all 4px wide):
    image
    image
- Added a simple checkbox to enable the side by side mode in the topbar of the changes tab (probably needs redesign, but it can come later).

app/styles/ui/_side-by-side-diff.scss
```diff
@@ -0,0 +1,284 @@
...
+      }
+    }
+
+    &.added .after {
+      width: calc(50% + 4px);
+
+      .line-number {
+        border-left-width: 4px;
+      }
+    }
```
niik, line 205: These changes seem to have caused a 1px misalignment

image

niik (changes requested):

rafeca:
> These changes seem to have caused a 1px misalignment

Good catch! just fixed it!

I've also did what we talked about the small refactor to the logic that calculates the columns data which has allowed me to improve the selection logic in split diffs to be more similar to selecting by drawing:

demo

app/src/lib/equality.ts
```diff
@@ -77,3 +77,33 @@ export function arrayEquals<T>(x: ReadonlyArray<T>, y: ReadonlyArray<T>) {
...
+    if (!y.has(key)) {
+      return false
+    }
+  }
+
+  for (const key of y.keys()) {
+    if (!x.has(key)) {
+      return false
+    }
+  }
```
niik, line 106: I doubt it has a practical performance impact but isn't this redundant given that we short-circuit if the size differs?  Shouldn't it be enough to just check that all the keys of `x` exists in `y`?

```suggestion

  for (const key of y.keys()) {
    if (!x.has(key)) {
      return false
    }
  }
```

app/src/ui/changes/changed-file-details.tsx
```diff
@@ -25,6 +32,14 @@ export class ChangedFileDetails extends React.Component<
         <PathLabel path={this.props.path} status={this.props.status} />
         {this.renderDecorator()}
 
+        <Checkbox
```
niik, line 35: Can we wrap this in a `enableSideBySideDiffs` feature conditional?

app/src/ui/diff/index.tsx
```diff
@@ -238,6 +243,22 @@ export class Diff extends React.Component<IDiffProps, IDiffState> {
   }
 
   private renderTextDiff(diff: ITextDiff) {
+    if (enableExperimentalDiffViewer() || this.props.showSideBySideDiff) {
```
niik, line 246: Should this be `(enableExperimentalDiffViewer() || (enableSideBySideDiffs() && this.props.showSideBySideDiff))`?

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -0,0 +1,967 @@
...
+    if (!this.props.showSideBySideDiff) {
+      return
+    }
+
+    // We need to use the event target since the current target will
+    // always point to the container.
+    const target = event.target as HTMLDivElement
+
+    const isSelectingBeforeText = target.closest('.before')
+    const isSelectingAfterText = target.closest('.after')
```
niik, line 427: This makes me a little uneasy. We don't know that the thing the  user happened to click on is a `div` do we? And it doesn't seem like we care either, we just want the `.closest` function. Could we do something like this instead?

```suggestion
    if (!(event.target instanceof HTMLElement)) {
      return
    }

    // We need to use the event target since the current target will
    // always point to the container.
    const isSelectingBeforeText = event.target.closest('.before')
    const isSelectingAfterText = event.target.closest('.after')
```

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -0,0 +1,967 @@
...
+  ) => {
+    this.setState({
+      temporarySelection: {
+        from: { row, column },
+        to: { row, column },
+        isSelected: select,
+      },
+    })
+
+    document.addEventListener('mouseup', this.onEndSelection, { once: true })
```
niik, line 449: I know that there's a `once` in there but could we add a corresponding `removeEventListener` in `componentWillUnmount` anyway? I've chased enough weird errors due to use-after-unmount issues in the past.

niik (changes requested):
I know we're coming right up on when you need to stop working on this @rafeca so don't feel like you have to address these pieces of feedback, I can do that. But if you see a faulty assumption from me I'd appreciate it if you pointed it out

app/src/lib/equality.ts
rafeca: Oh yes you're right!

app/src/ui/diff/index.tsx
rafeca: It should be equivalent since it should not be possible to have `showSideBySideDiff` to true when `enableSideBySideDiffs()` returns false, but yes adding a double check won't hurt us

niik:
> * Improved the gutters to select lines/hunk for partial committing (now they're all 4px wide):

I've had to revert this change (i.e https://github.com/desktop/desktop/pull/10617/commits/3c32a6a7c2f910f21f03c61edd80ace894c1946e). While https://github.com/desktop/desktop/pull/10617/commits/d38716d8ad87a79493b327ce63297a8e229d9012 fixed the initial misalignment there were still problems with misaligned columns in other zoom levels. I figured it's not worth holding up this entire PR for and we can take a look at it in a separate PR.

niik (approved):
This is sufficiently isolated and tested that I'm comfortable bringing it to development. Once #10756 has merged we'll see about releasing this to beta.

Absolutely stellar work here @rafeca.

LeonardLaszlo:
When are we going to see this in the official release?

niik:
> When are we going to see this in the official release?

We're hoping to ship this in the next production release in early to mid November

victor871129:
Two features that Sublime Merge has and Github Desktop does not was side-by-side diff and viewing all the files at once, modified files and staged files, all files at once. Awesome that you implemented side-by-side diff. Can you implement a setting to view all the modified-staged files at once?

MarcelloTheArcane:
This is fantastic!  If it's a new file, it would make sense to show the old single diff, do you think?

image

JosephTLyons:
> This is fantastic!  If it's a new file, it would make sense to show the old single diff, do you think?
> 
> 
> 
> image
> 
> 

I think that makes sense.  I suppose it would imply that if it were a single file delete, it would show the old single file view as well?

MarcelloTheArcane:
@JosephTLyons  Yep, I forgot deleting files too 😄

ariooo100 (commented):
delete

mohamaddartomi (commented):
holle

killmen136 (commented):
Ed

PR #10396, opened 2020-08-17
Adopt new primer border radius

niik:
~~🌵 This is based on #10224, please review that first 🌵~~

## Description

In primer v15 the default border radius was increased from 3px to 6px (https://github.com/primer/css/commit/2b54e867a4006ca0fffd9b01c9c1ab689c9a34f9) along with a slew of other visual changes. As GitHub.com upgraded to primer v15 not to long ago the discrepancy between Desktop and GitHub.com has become more noticeable.

This PR takes us one tiny step towards aligning (where aligning makes sense) to Primer v15 and beyond.

### Screenshots

image (1)
image (2)
image (3)
image

## Release notes

Notes: no-notes

Issue #10223, opened 2020-07-17
Upgrade octicons


rafeca:
The latest release of the `octicons` package (v10.0.0) includes the updated icons refresh that was shipped in GitHub.com during 2020 ([more internal info](https://github.com/github/design-systems/issues/711)).

This issue tracks the work needed to update Desktop to use the refreshed icons.

PR #10639, opened 2020-09-24
Increase base font size slightly

niik:
## Description

As part of the minor visual refresh we're doing (see #10223, #10396) we've come to realize that our base font size of 12px doesn't align with the default operating font size or that of other popular electron applications such as VS Code. As such we've decided to increase it ever so slightly to 13px and the diff font size to 12px (up from 11px).

#### Before

Screen Shot 2020-09-24 at 18 33 27

#### After

Screen Shot 2020-09-24 at 18 30 33


rafeca (dismissed):
A subtle change but makes things look nicer!

niik:
We've decided to punt on this for now. An increase in font size would carry along with it a large number of other spacing changes that we don't have the bandwidth to tackle right now.

PR #10646, opened 2020-09-25
Align resolve button text and size icon appropriately

niik:
## Description

In #10224 we upgraded the version of octicons that we ship and this little fella missed our careful audit of changes. We had an explicit height and width set for the arrow down octicon that didn't align with the dimensions of the new icon causing the text in the button to be misaligned.

#### Before
Screen Shot 2020-09-25 at 14 23 11

#### After
Screen Shot 2020-09-25 at 14 22 59

## Release notes

Notes: no-notes


rafeca (approved):
LGTM! ✨

Issue #2745, opened 2017-09-15
Column widths are limited — would love to remove those widths


tjdraper:
### Description

(forgive me if this is an issue already, I looked and couldn't find it)

The various columns in the application are constrained to certain max widths — which makes using the app on larger screens frustrating. Particularly when being able to expand a column to see all the data would be very helpful. The application should not place a width limit on columns so that the user of the application can determine the column widths desired at any time.

<img width="974" alt="github-columns" src="https://user-images.githubusercontent.com/3803475/30488007-bea92c48-99fa-11e7-8e1c-12c8c9f66f39.png">

### Version

**GitHub Desktop version:** 0.8.2

**OS version:** macOS Sierra 10.12.6

### Steps to Reproduce

1. Try to resize a column in the app to any width wider than the limit

**Expected behavior:** Column should resize to any size user would like

**Actual behavior:** Columns have a max-width limit

joshaber:
Thanks @tjdraper! We've done some work to increase those limits, but I agree it'd be nice if they'd be entirely dependent on the window size.

tjdraper:
I’d argue they should be independent of window size and be entirely user defined, but I’d certainly accept them being wider when the window is wider. 🙂

niik:
While this issue pertains to all resizable panes I think the answer I wrote in https://github.com/desktop/desktop/issues/1588#issuecomment-329405096 applies here as well.

arfordweb:
From @niik 's other comment and subsequent responses, it seems the max-width was a temporary solution.  The real solution would be to handle the case for when the panes that currently have max-widths on them have been made wide by the user, then the window is made narrower than the wide column can accommodate.  The widths of those panels need to be recomputed, which sounds to be non-trivial.

I would, though, like to throw in that I'd love to see the column of file path/names expandable as large as we'd like to expand it, in the hopes that maybe it gets prioritized.  It's by far my biggest UX frustration with this otherwise beautiful and intuitive app.

glennreid:
Rather than enforcing MAX constraints, you can just enforce MIN constraints.  The only reason to have a MAX is so you can't a column so wide as to reduce its neighbor to a size that's too small.  Enforcing a MIN limit fixes this, and lets people decide how they want to apportion the window width across columns.

The fewer hard limits the better.  That's a tenet of all programming.

richardthombs:
[...]

Issue #1588, opened 2017-05-18
Ability to make the file list wider


xt0rted:
### Description

I'm running on a 27" 4k @ 150% scaling and the electron version of the app generally feels smaller than the old native version. For the most part this isn't an issue but for things like the file list the maximum width is half that of the native version. I'd like to be able to make this wider now that the file paths are no longer shown in the wider portion of the window.

![image](https://cloud.githubusercontent.com/assets/831974/26182158/35fa2d12-3b44-11e7-9850-0226ec2230ea.png)

According to the dev tools this is set to  `350px`, I'd like to see a max width closer to `600px`. I'm also noticing the file sorting isn't the same between both versions, not that that really matters.

![image](https://cloud.githubusercontent.com/assets/831974/26182312/64a7498c-3b45-11e7-84de-5b4bb3208b29.png)

### Version

**GitHub Desktop version:** 0.5.4
**OS version:** Win10 Creators Update (10.0.15063)

shiftkey:
> I'm also noticing the file sorting isn't the same between both versions, not that that really matters.

Happy to look into this in a separate issue if you want to put together a repro of it.

xt0rted:
Comparing the two versions shows that the old version scaled that list. So if the window was up to 50% the width of my screen it would max out at ~350px, but as the window got wider it would also get wider until at full width it was at ~600px. Having it be able to auto size like that would be awesome.

joshaber:
I've actually wanted this too 👍

niik:
For anyone who'd like to take a stab at this the `Resizable` component for the file list lives in [repository.tsx](https://github.com/desktop/desktop/blob/432738e426f83c309375291f0d5da86132f167ae/app/src/ui/repository.tsx#L111-L122) and it [supports overriding the maximum width](https://github.com/desktop/desktop/blob/432738e426f83c309375291f0d5da86132f167ae/app/src/ui/resizable/resizable.tsx#L7-L11) making for an excellent first contribution!

rlabrecque:
Any downside to having the maximumWidth set to window.innerWidth / 2?
That feels more natural to me. Two problems when trying it out: If you resize the window then it doesn't resize the Resizable until the window loses focus, and the `sidebar-section` toolbar button doesn't update until you manually resize the Resizable.

shiftkey:
> Any downside to having the maximumWidth set to window.innerWidth / 2?

I'd relax this even more, like VSCode does. Here's how far you can drag their sidebar:

[...]

PR #10706, opened 2020-10-01
Dynamically set max width of resizable component

marzepani:
Related to #2745 (and #1588)

## Description

The max-width limit of the sidebar and file list is too small on wide screens. This PR dynamically sets the max-width of every resizable component to one-third of the main window. It does not listen to the window resize event, but it recalculates on every drag start to keep it simple.

## Release notes

Notes:
I don't think the width of the sidebar and file list should be automatically resized when the main window is resized. Auto-resizing can be quite annoying. Al least to me, it is.


Luckylukee (approved):
#10706

ibtisam167 (approved):

paul-szczepanek-arm (approved):
Why is this PR stuck? Looks great!

Shahryarmohammadie:
good

marzepani:
It has been one year now that I opened this PR. Is anyone interested in reviewing, accepting, rejecting, or commenting?

sergiou87:
Sorry about the radio silence @marzepani 😓 

We discussed about this PR a few weeks ago, and we liked the idea but were afraid it could break anything without us noticing, so we wanted to test it thoroughly but so far we haven't found the time yet 😭

danny007in:
better use percentage for width.

adamgit:
This appears to be a trivial change that takes an *already broken* part of the UI and makes it less broken. How is there a risk here that 'something else' might get broken? It's barely changing 8 lines of code.

Would it help if I submitted a PR that simply removes the offending code, the max-size limit that should never have been there in the first place, as per glennreid's comment - now 3 years old - https://github.com/desktop/desktop/issues/2745#issuecomment-957914585 ?

niik:
Thanks so much for wanting to contribute here @marzepani, we really appreciate it! I'm going to close this in favor of #13439 which hopefully will let us relax the constraints of the resizable components without running into the risks that @sergiou87 alluded to.

PR #10720, opened 2020-10-03
Spike primer utility classes

garg-manish:
Closes #[issue number]

## Description

-

## Release notes

Notes:


niik:
Thanks for your submission.

It appears that you've created a pull request that based on an old, defunct branch. This is a pattern we often see from automated scripts so we're going to go ahead and close this. If it was intentional, please let us know what you were intending and we can see about reopening it.

Thanks again!

PR #10746, opened 2020-10-07
Adding pointer for tab bar items

Kevin-Kayser:
Adding visual pointer to help show that tab-bar-items are clickable. 

This PR is not related to a reported issue.

## Description

- Adds css property to use mouse pointer when focusing 

### Screenshots

tab-item-pointer

## Release notes

Notes: Adding visual pointer to help show that tab-bar-items are clickable.

### DIFF
--- a/app/styles/ui/_no-repositories.scss
+++ b/app/styles/ui/_no-repositories.scss
@@ -121,6 +121,7 @@
 
       .tab-bar-item {
         border-bottom: 0;
+        cursor: pointer;
       }
     }
     .clone-selected-repository {


niik:
@Kevin-Kayser thanks for opening this. I feel like we haven't been very consistent here in the past but the original intention was for us not to use the pointer as that's a not a pattern that desktop applications typically employ instead favoring patterns like hover states (https://github.com/desktop/desktop/pull/6310). That said we do use it in a few places in the app already:

- The close `x` button in dialogs
- Link buttons
- Hunk and line handles in partial commits
- Co-Author toggle icon in the commit message
- Image diff mode selection

I'd like to come up with a guideline here before we move ahead with this PR so I'm going to tag @ampinsk for when she's back from vacation.

Kevin-Kayser:
> @Kevin-Kayser thanks for opening this. I feel like we haven't been very consistent here in the past but the original intention was for us not to use the pointer as that's a not a pattern that desktop applications typically employ instead favoring patterns like hover states (#6310). That said we do use it in a few places in the app already:
> 
> * The close `x` button in dialogs
> * Link buttons
> * Hunk and line handles in partial commits
> * Co-Author toggle icon in the commit message
> * Image diff mode selection
> 
> I'd like to come up with a guideline here before we move ahead with this PR so I'm going to tag @ampinsk for when she's back from vacation.

Thanks @niik, I appreciate the response and the general feedback provided. Since this short PR was submitted I also noticed additional places (that you mentioned too). That being said, I'd be happy to help implement and/or discuss implementation details with @ampinsk later on.

ampinsk:
After playing with this a bit, I think the only place we should really have pointer cursors is on text links, so we should actually  do a pass and remove the pointer cursors from most of those spots 👍 Not using pointer cursors helps make the app feel more native.

billygriffin:
I'm going to close this PR based on @ampinsk's feedback, but @Kevin-Kayser please let us know if you'd like to create a new PR for removing the pointers in other areas of the app! Thanks!

PR #10756, opened 2020-10-07
Diff options popover

niik:
## Description

Adds a button with a gear icon to the changes and history view that toggles a popover containing diff options. This replaces the checkbox we currently have for hiding/showing whitespace changes and the diff mode.

For discoverability reasons we show a little 'New' callout bubble next to the button. Both the popover and the new callout are new patterns in the app but I think they fit in nicely. The popover parts of the diff options component could and should be extracted into its own reusable container but I'm punting on that until we want to reuse it.

To a large extend this copies the GitHub.com UX for diff settings

image

### Screenshots

image

image

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -33,6 +33,7 @@
     "file-metadata": "^1.0.0",
     "file-uri-to-path": "^2.0.0",
     "file-url": "^2.0.2",
+    "focus-trap-react": "^8.1.0",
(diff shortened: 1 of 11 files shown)


niik:
@ampinsk Feel free to do an after-the-fact review of this, I'd love your input. Since you're out for the week however I'm going to merge this as-is to get it into the next beta.

PR #10762, opened 2020-10-08
Include stdout and stderr in push error messages

### LINKED ISSUE #5439: include the output from a pre-push commit hook failure
I'm running `tsc` and `tslint` in a pre-push hook, and GitHub Desktop shows a message when the hook finds a problem: "error Command failed with exit code 1".

It would be useful if this message included the output from the git hook, to help with diagnosing the problem.

clarkd:
Closes #5439

## Description

Unlike when committing, when a push operation fails the full error is not shown. This is useful when using prepush hooks or similar as, without the error the user is left confused or having to run the prepush hook manually to figure out what went wrong. 

Have matched similar code achieving the same for commits in `app/src/lib/git/commit.ts`.

#### BEFORE
image

#### AFTER
image

## Release notes

Notes: Push errors now include the full error message

### DIFF
--- a/app/src/lib/git/push.ts
+++ b/app/src/lib/git/push.ts
@@ -131,19 +131,12 @@ export async function push(
       const stderr = e.result.stderr.trim()
       const stdout = e.result.stdout.trim()
 
-      let message = ''
-      if (stderr.length > 0) {
-        // append stderr
-        message += stderr
-      }
-      if (stdout.length > 0) {
-        // append stdout
-        message += stdout
-      }
+      const combined = stderr.concat(stdout)
+      const message = combined.length > 0 ? `, with output: '${combined}'` : ''
 
       const exitCode = e.result.exitCode
       const error = new Error(
-        `Push failed - exit code ${exitCode} received ${message}`
+        `Push failed - exit code ${exitCode} received${message}`
       )
       error.name = 'push-failed'
       throw error
--- a/app/src/lib/git/push.ts
+++ b/app/src/lib/git/push.ts
@@ -120,9 +120,35 @@ export async function push(
     })
   }
 
-  const result = await git(args, repository.path, 'push', opts)
+  try {
+    const result = await git(args, repository.path, 'push', opts)
 
-  if (result.gitErrorDescription) {
-    throw new GitError(result, args)
+    if (result.gitErrorDescription) {
+      throw new GitError(result, args)
+    }
+  } catch (e) {
+    if (e instanceof GitError) {
+      const stderr = e.result.stderr.trim()
+      const stdout = e.result.stdout.trim()
+
+      let message = ''
+      if (stderr.length > 0) {
+        // append stderr
+        message += stderr
+      }
+      if (stdout.length > 0) {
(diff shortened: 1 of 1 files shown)


clarkd:
Is there anything else needed from me here, or is it just a case of waiting till this reaches the top of the (very busy) pile? @niik?

niik:
>is it just a case of waiting till this reaches the top of the (very busy) pile? @niik?

Yeah, pretty much I'm afraid. We're very strapped for resources at the moment (this week it's just me) so there's a risk it'll take a while for us to get to this.

One thing for you to noodle on until I can do an actual review is how this implementation could lead to out-of-order messages compared to how it would appear to a terminal user since it's buffering all stderr and stdout output individually and then presenting one after the other. Ideally we'd join the stderr and stdout streams to produce something more akin to how a terminal user would perceive it.

clarkd:
> > is it just a case of waiting till this reaches the top of the (very busy) pile? @niik?
> 
> Yeah, pretty much I'm afraid. We're very strapped for resources at the moment (this week it's just me) so there's a risk it'll take a while for us to get to this.
> 
> One thing for you to noodle on until I can do an actual review is how this implementation could lead to out-of-order messages compared to how it would appear to a terminal user since it's buffering all stderr and stdout output individually and then presenting one after the other. Ideally we'd join the stderr and stdout streams to produce something more akin to how a terminal user would perceive it.

Thanks for taking the time to reply. I hadn't thought about the two streams like that, I'll give it some thought and see if there's something we can do.

clarkd:
> > > is it just a case of waiting till this reaches the top of the (very busy) pile? @niik?
> > 
> > 
> > Yeah, pretty much I'm afraid. We're very strapped for resources at the moment (this week it's just me) so there's a risk it'll take a while for us to get to this.
> > One thing for you to noodle on until I can do an actual review is how this implementation could lead to out-of-order messages compared to how it would appear to a terminal user since it's buffering all stderr and stdout output individually and then presenting one after the other. Ideally we'd join the stderr and stdout streams to produce something more akin to how a terminal user would perceive it.
> 
> Thanks for taking the time to reply. I hadn't thought about the two streams like that, I'll give it some thought and see if there's something we can do.

@niik I've taken a brief look at this, but from what I can see stdin and stdout are both strings at the point we have access to them - I can't see that we have access to the buffer/stream. I may well have missed something though so feel free to point me in the right direction if you've got an idea.

clarkd:
@niik Is there anything I can do on my side to help this along?

niik:
Hey @clarkd, sorry  for the delay in getting back to you. We've had a lot on our plate for a while. @sergiou87 took a stab at a solution for this problem which combines the stderr and stdout streams by line over in https://github.com/desktop/desktop/pull/11368. Take a look at that and see what you think.

As previously stated we'd rather try to find a solution which merges the streams rather than concatenates so I'm going to close this PR in favor of https://github.com/desktop/desktop/pull/11368 but I want to stress that we really appreciate your contribution and we hope this won't dissuade you from contributing in the future!

PR #10793, opened 2020-10-11
chore: fix typos in comments

dsanders11:
## Description

Fixes typos in comments in the code. Since they're only in comments should be no risk of unintended side effects. Just needs a check to make sure I didn't mess any of them up.

## Release notes

Notes: no-notes


niik (approved):
Thank you so much @dsanders11!

dsanders11:
@niik, happy to do it, I get an odd amount of satisfaction from cleaning up typos.

PR #10807, opened 2020-10-12
chore: fix typos

dsanders11:
## Description

A few fixes for typos in the code. They're only in log statements or variable names, so should be safe of any side effects.

@niik, last PR on typos, that's pretty much all I found. The only other two had potential side-effects of changing, so I left them as is, which are [this one (Initated => Initiated)](https://github.com/desktop/desktop/blob/d6176b9b9f5873f9a4814296a863fc5c5d1be0d4/app/src/lib/stats/stats-store.ts#L789), and [this one (Clonable => Cloneable)](https://github.com/desktop/desktop/blob/d6176b9b9f5873f9a4814296a863fc5c5d1be0d4/app/src/ui/clone-repository/group-repositories.ts#L9).

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/stage.ts
+++ b/app/src/lib/git/stage.ts
@@ -67,6 +67,6 @@ export async function stageManualConflictResolution(
       break
     }
     default:
-      assertNever(chosen, 'unnacounted for git status entry possibility')
+      assertNever(chosen, 'unaccounted for git status entry possibility')
   }
 }
--- a/app/src/lib/stores/helpers/create-tutorial-repository.ts
+++ b/app/src/lib/stores/helpers/create-tutorial-repository.ts
@@ -14,7 +14,7 @@ import { IRemote } from '../../../models/remote'
 import { envForRemoteOperation } from '../../git/environment'
 
 const nl = __WIN32__ ? '\r\n' : '\n'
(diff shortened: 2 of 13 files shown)


niik:
> this one (Initated => Initiated)

Yeah, we'll have to live with that one since changing it would mean altering our metrics pipeline in our backend.

> [this one (Clonable => Cloneable)](https://github.com/desktop/desktop/blob/d6176b9b9f5873f9a4814296a863fc5c5d1be0d4/app/src/ui/clone-repository/group-repositories.ts#L9).

That one, however, is just an interface so if you'd like you can go ahead and change that without risking side-effects. Note that you'll need to change it in more than just that one file though.

dsanders11:
> Yeah, we'll have to live with that one since changing it would mean altering our metrics pipeline in our backend.

Yea, I figured once I saw it was a metric that it was probably baked in at this point, heh.

> That one, however, is just an interface so if you'd like you can go ahead and change that without risking side-effects. Note that you'll need to change it in more than just that one file though.

Cool, I'll go ahead and add that one. I'll be sure to get all the instances.

niik (approved):
Awesome, thanks a bunch @dsanders11!

PR #10868, opened 2020-10-19
Fix no-response bot

steveward:
Fixes our no-response bot which has been broken for a while due to a label change.

## Description

- Updated the `more-information-needed` label to `more-info-needed` since we changed that at some point.
- Reduced the days until close to 7 since we've been manually been closing stale `more-info-needed` issues in that time frame.


niik (approved):
Thanks @steveward! I also noticed that the no-response app wasn't installed in the org. I've been searching slack to see if I could find a reason why it was removed. I did realize though that we need to go through the currently open issues tagged with `more-info-needed` before we unleash this bot

Issue #5639, opened 2018-09-13
I can't push my commits to my private repository.


Yeadh:
When I am trying to push my commit it is showing an error message every time.It says 
"cannot push these commits as they contain an email address marked as private on GitHub"
![screenshot_3](https://user-images.githubusercontent.com/8848247/45509460-7421b900-b7b9-11e8-8697-cdc9b8c31ff0.png)

Please help me how can I solve this problem?

DanielHix:
You're likely pushing a commit that is signed using your email address, and in your settings on Github. You can change this behavior by either going to github -> settings -> email and unchecking the box that says Block command line pushes that expose my email, make your email not private (right above that option), or sign your commits with a public email.

Yeadh:
@DanielHix I have unchecked the box that says Block command line pushes that expose my email and also made my email not private by unchecked the box and now I have the same problem.

Also, I sign in with a public email but still not working.

DanielHix:
Can you give steps that you do when this problem occurs? Such as, are you creating a new repo and this happens or is to any repo?

Yeadh:
This problem occurs when I create a new repo and in my old repo when I push, it says "last fetched just now" but into the repo, it shows nothing update.
![screenshot_4](https://user-images.githubusercontent.com/8848247/45532259-36527e00-b815-11e8-88fc-2ba5037bbb90.png)

shiftkey:
@Yeadh thanks for the report!

Could you attach the full log files? The **Help** | **Show Logs in Explorer** menu items should open the logs folder, and inside those are the log files generated by day. This will provide more information to help troubleshoot.

Yeadh:
It is working now.

wasteproduct:
@DanielHix Which github do you mean by "going to github -> settings ?" I can't find the email part.

assidiwassim:
![image](https://user-images.githubusercontent.com/29416254/66218168-ee74f200-e6c8-11e9-91a7-57810b2a48c7.png)

going to github -> settings -> email and unchecking this box.
view more https://help.github.com/en/articles/setting-your-commit-email-address

Saleca:
> You're likely pushing a commit that is signed using your email address, and in your settings on Github. You can change this behavior by either going to github -> settings -> email and unchecking the box that says Block command line pushes that expose my email, make your email not private (right above that option), or sign your commits with a public email.

[...]

PR #10932, opened 2020-10-27
Add workaround to "email address marked as private" error message

steveward:
Related to https://github.com/desktop/desktop/issues/5639.

## Description

Users can run into a scenario where they have commits that have been made with a non-private email address. If they attempt to push these commits and have the "Keep my email address private" option checked in https://github.com/settings/emails the push will fail. This PR updates the error message to provide the workaround of uncheck -> push -> re-check.

## Release notes

N/A 

## Notes:

Question for @niik: is it possible to have URLs be rendered as clickable links in error messages? 

For anyone: Is the last sentence necessary or is it implied? I don't want the error message to be unnecessarily lengthy.

### DIFF
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -335,7 +335,7 @@ function getDescriptionForError(error: DugiteError): string | null {
     case DugiteError.CannotMergeUnrelatedHistories:
       return 'Unable to merge unrelated histories in this repository.'
     case DugiteError.PushWithPrivateEmail:
-      return 'Cannot push these commits as they contain an email address marked as private on GitHub.'
+      return 'Cannot push these commits as they contain an email address marked as private on GitHub. To workaround this go to https://github.com/settings/emails and uncheck the "Keep my email address private" checkbox. Then switch back to GitHub Desktop to push your commits. You can then enable the checkbox again.'
     case DugiteError.LFSAttributeDoesNotMatch:
       return 'Git LFS attribute found in global Git configuration does not match expected value.'
     case DugiteError.ProtectedBranchDeleteRejected:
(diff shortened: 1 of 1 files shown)


app/src/lib/git/core.ts
```diff
@@ -335,7 +335,7 @@ function getDescriptionForError(error: DugiteError): string | null {
     case DugiteError.CannotMergeUnrelatedHistories:
       return 'Unable to merge unrelated histories in this repository.'
     case DugiteError.PushWithPrivateEmail:
-      return 'Cannot push these commits as they contain an email address marked as private on GitHub.'
+      return 'Cannot push these commits as they contain an email address marked as private on GitHub. To workaround this go to https://github.com/settings/emails and uncheck the "Keep my email address private" checkbox. Then switch back to GitHub Desktop to push your commits. You can then enable the checkbox again.'
```
billygriffin, line 338:
```suggestion
      return 'Cannot push these commits as they contain an email address marked as private on GitHub. To push anyway, visit https://github.com/settings/emails, uncheck "Keep my email address private", then switch back to GitHub Desktop to push your commits. You can then enable the setting again.'
```

billygriffin (commented):
Minor suggestions for brevity and clarity (not wedded to this), but as you alluded to I'm on the fence about the final sentence.

niik (approved):
> Question for @niik: is it possible to have URLs be rendered as clickable links in error messages?

It is possible, but not straightforward. As it's architected right now you'd have to create a custom error handler that detect just this error and presents a custom dialog. I think there are ways we could make it easier but for now I think this is a worthwhile improvement in and of itself.

While not ideal at least the URL is selectable for copy-pasting: 
image


Ideally we'd use the url provided by the GitHub error message rather than hardcoding GitHub.com here but since GHES doesn't currently support private emails I think we can live with the hardcoded URL.

PR #10934, opened 2020-10-28
Realign the diff options component

niik:
## Description

The  changes header assumed that there would only be one octicon in it and therefore applied styling that interfered with the new diff options component (#10756).

#### Before

image

image

#### After

image

image

## Release notes

Notes:


billygriffin (approved):
Looks great, thanks @niik! So much nicer.

PR #9691, opened 2020-05-02
Add Windows ARM64 + Apple Silicon support

dennisameling:
Closes #9034 

## Description

**Requirements for cross-compilation from x64 to arm64:**
- [x] (✔️ merged) Let Dugite download 32-bit Dugite Native (including Git for Windows) as Git isn't available natively on Windows arm64 https://github.com/desktop/dugite/pull/392
- [x] (✔️ 7.1.0 released which contains the fix) Add prebuild Electron arm64 to Keytar https://github.com/atom/node-keytar/pull/319 + fix native arm64 build on Windows https://github.com/atom/node-keytar/pull/325
- [x] (✔️ merged) Switch to using the dart version of sass instead of node-sass https://github.com/desktop/desktop/pull/11019
- [x] (✔️ merged) Update Electron-winstaller Squirrel version to v2+ for arm64 support https://github.com/electron/windows-installer/pull/361 

### Building for Windows ARM64
```
Run the following in Windows Command Prompt (NOT PowerShell):
set npm_config_arch=arm64
set TARGET_ARCH=arm64
yarn
yarn build:prod
```

### Running tests
In my [test build](https://github.com/desktop/desktop/pull/9691#issuecomment-721156225), all tests have passed: yarn test:unit:cov, yarn test:script:cov, yarn test:integration.

### Getting the installer to work
Squirrel has been updated to support Windows arm64, now we're only waiting for electron-winstaller to approve the PR that updates Squirrel: https://github.com/electron/windows-installer/pull/361 

### Screenshots

No UI changes, but just to show that building + launching + login on Windows ARM works:

image

## Release notes

Notes: Add Windows ARM build support

PR #10949, opened 2020-10-31
Update node-sass to v5.0.0

dennisameling:
Closes #N/A

## Description

- This PR updates [node-sass to v5.0.0](https://github.com/sass/node-sass/releases/tag/v5.0.0). It contains some important dependency updates (like node-gyp) which are needed for https://github.com/desktop/desktop/pull/9691.

### Screenshots

N/A

## Release notes

Notes:
- Update node-sass to v5.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "klaw-sync": "^3.0.0",
     "legal-eagle": "0.16.0",
     "mini-css-extract-plugin": "^0.4.0",
-    "node-sass": "^4.13.1",
+    "node-sass": "^5.0.0",
     "parallel-webpack": "^2.3.0",
     "prettier": "2.0.5",
     "request": "^2.72.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1250,6 +1250,16 @@ ajv@^6.10.0, ajv@^6.10.2, ajv@^6.12.0:
     json-schema-traverse "^0.4.1"
     uri-js "^4.2.2"
 
+ajv@^6.12.3:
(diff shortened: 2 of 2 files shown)


niik:
Hey @dennisameling, thanks for opening this! I'd be happy to merge this as-is but I'm wondering if we should just switch out node-sass in favor of [sass](https://www.npmjs.com/package/sass) and avoid the native compilation step altogether. It seems that's what [webpack is recommending these days](https://webpack.js.org/loaders/sass-loader/)

> ℹ️ We recommend using Dart Sass.

I believe we'd need to swap `node-sass` for `sass` in addition to bumping `sass-loader` to the latest version.

What do you think?

dennisameling:
Moving to `sass` makes sense I think, but unfortunately I don't have the bandwidth currently to look into a possible migration path. I'm very close to having a working GH Desktop build process on native Windows ARM64, which will allow me to verify if there are any failing tests on this platform 🚀

niik:
> Moving to `sass` makes sense I think, but unfortunately I don't have the bandwidth currently to look into a possible migration path

That's not a problem. Let me see if I can find some time to do it in the next couple of days (I believe I might have done it already on a local branch somewhere) and if not then we can just merge this. Are you able to work around this problem currently or are you blocked until one of the two solutions lands on development?

dennisameling:
I'm not blocked currently as I updated `node-sass` to v5 in my local branch. Thanks!!

niik:
Gonna close this in favor of #11019 which incorporates this branch

PR #10957, opened 2020-11-02
Refactor checkout branch and stash logic

### LINKED ISSUE #7617: Prevent concurrent appStore._checkoutBranch call from error handlers

### LINKED ISSUE #8099: Refactor appStore._checkoutBranch
appStore._checkoutBranch is a 150 line method with a lot going on. It also has a reentrant problem described in https://github.com/desktop/desktop/issues/7617 (See below for more notes.)

For maintainability, it makes sense to extract this method into a few methods (or possibly into its own class). This would make it easier to maintain and open up the possibility for it to be unit tested more easily.

_notes on #7617_

#7617 describes why checkoutBranch becomes reentrant. Basically, if the checkout fails, it emits an error, exits the method, the error handler then takes over and stashes the current changes and calls checkoutBranch again. This is pretty hard to follow and #7617’s goal was to solve this. But when I tried to solve it in  https://github.com/desktop/desktop/pull/8076 it felt like I was introducing more complexity to an already complex method. So I closed that PR and opened up this one. Now you are all caught up!

### LINKED ISSUE #9297: 2.3.2 beta has a UI Loop when checkout in isn't possible
### Describe the bug
When trying to checkout a PR, it goes into a fun loop retrying the checkout but not presenting an error (screenshot below).

At the console, here's what the error was:
```
[19:45]  ❯ git checkout pr/1389
error: Your local changes to the following files would be overwritten by checkout:
        tests/RedisConfigs/Sentinel/sentinel-26379.conf
        tests/RedisConfigs/Sentinel/sentinel-26380.conf
Please commit your changes or stash them before you switch branches.
Aborting

[19:46]  ❯ git status
On branch master
Your branch is up to date with 'origin/master'.

```

Note that the 2 files in question do _not_ show as pending differences, because they have both been:
```
git update-index --assume-unchanged <file>
```
(this is because Redis stores state in the config, and we shouldn't be updating master with that state constantly)

This combination seems to confuse the UI and it doesn't present an error...it just gets in a fun loop. Closing the app is the only way I've found to escape it.

### Version & OS
OS: Windows 10
Client: 2.3.2-beta5

See above description...unfortunately not trivial to reproduce. But it's possible this repros without any PR needed on any branch checkout.

### Expected behavior
User gets an error indicating why checkout isn't possible.

### Actual behavior
Infinite merry-go-round loop of fun!

### Screenshots

GHD-Loop

### Logs
Logs have sensitive data from our GHE instance, but the only related bits are this in a loop:
```
2020-03-16T23:45:28.667Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: StackExchange.Redis - NickCraver (has token)
2020-03-16T23:45:29.139Z - error: [ui] `git -c credential.helper= -c protocol.version=2 checkout --progress pr/1389 --recurse-submodules --` exited with an unexpected code: 1.
stderr:
error: Your local changes to the following files would be overwritten by checkout:
	tests/RedisConfigs/Sentinel/sentinel-26379.conf
	tests/RedisConfigs/Sentinel/sentinel-26380.conf
Please commit your changes or stash them before you switch branches.
Aborting

(The error was parsed as 37: Unable to switch branches as there are working directory changes which would be overwritten. Please commit or stash your changes.)
2020-03-16T23:45:29.671Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: StackExchange.Redis - NickCraver (has token)
2020-03-16T23:45:30.127Z - error: [ui] `git -c credential.helper= -c protocol.version=2 checkout --progress pr/1389 --recurse-submodules --` exited with an unexpected code: 1.
stderr:
error: Your local changes to the following files would be overwritten by checkout:
	tests/RedisConfigs/Sentinel/sentinel-26379.conf
	tests/RedisConfigs/Sentinel/sentinel-26380.conf
Please commit your changes or stash them before you switch branches.
Aborting

(The error was parsed as 37: Unable to switch branches as there are working directory changes which would be overwritten. Please commit or stash your changes.)
2020-03-16T23:45:30.596Z - info: [ui] [AppStore.withAuthenticatingUser] account found for repository: StackExchange.Redis - NickCraver (has token)
2020-03-16T23:45:31.046Z - error: [ui] `git -c credential.helper= -c protocol.version=2 checkout --progress pr/1389 --recurse-submodules --` exited with an unexpected code: 1.
stderr:
error: Your local changes to the following files would be overwritten by checkout:
	tests/RedisConfigs/Sentinel/sentinel-26379.conf
	tests/RedisConfigs/Sentinel/sentinel-26380.conf
Please commit your changes or stash them before you switch branches.
Aborting

(The error was parsed as 37: Unable to switch branches as there are working directory changes which would be overwritten. Please commit or stash your changes.)
```

Hope this helps!

### LINKED ISSUE #10956: GitHub Desktop may overwrite existing stash when "Always stash and leave my changes" is selected
### Describe the bug

When checking out a branch using the pull request list or by following a "Checkout in Desktop" link on GitHub.com while having the "Always stash and leave my changes on the current branch" option selected in preferences (under Advanced) GitHub Desktop fails to prompt the user for confirmation before overwriting their existing stash.

### Version & OS

I believe this has been true since https://github.com/desktop/desktop/pull/8813 which shipped in 2.3.0. OS-independent

1. Ensure your current branch has a stash associated with it.
2. Ensure you have uncommitted changes in your working directory
3. Ensure you've selected the "Always stash and leave my changes on the current branch" option in preferences
4. Attempt to check out a pull request (note: not a branch, there's different logic involved here)

### Expected behavior

GitHub Desktop should show a confirmation prompt like the one below

image

### Actual behavior

The PR branch is checked out and the stash on the previous branch is overwritten with a newly created stash containing your uncommitted changes.

### Additional context

I've got a fix ready for this, I just wanted to ensure we have a record of this being a bug that we can refer back to. The dropped stash can be restored by looking through the GitHub Desktop logs and searching for a message with the following format:

```
 info: [ui] Dropped stash '0e95c933c4c6ba4403564407f2c8a7fbe829fb3d' associated with main
```

Taking that stash SHA one could restore the changes by running `git stash apply 0e95c933c4c6ba4403564407f2c8a7fbe829fb3d`

niik:
Closes #9297 - Infinite loop when checking out branch
Closes #10956 - GitHub Desktop may overwrite existing stash
Closes #8099 - Reduce complexity of checkout logic
Closes #7617 - Rewrite checkout logic to keep logic inline as opposed to relying on error handlers and continuations.

## Description

I set out to understand the cause of #9297 and ended up having to significantly rework our stashing and checkout logic.

### What was the problem with the previous logic

Aside from the two serious bugs of an infinite loop (#9297) and the potential for users to loose work by Desktop overwriting their existing stash (#10956) the checkout logic has been a source of much confusion and complexity for a long time.  #8099 does a good job of providing an overview of the problem:

> appStore._checkoutBranch is a 150 line method with a lot going on. It also has a reentrant problem described in https://github.com/desktop/desktop/issues/7617 (See below for more notes.)
> 
> For maintainability, it makes sense to extract this method into a few methods (or possibly into its own class). This would make it easier to maintain and open up the possibility for it to be unit tested more easily.
> 
> _notes on #7617_
> 
> #7617 describes why checkoutBranch becomes reentrant. Basically, if the checkout fails, it emits an error, exits the method, the error handler then takes over and stashes the current changes and calls checkoutBranch again. This is pretty hard to follow and #7617’s goal was to solve this. But when I tried to solve it in  https://github.com/desktop/desktop/pull/8076 it felt like I was introducing more complexity to an already complex method. So I closed that PR and opened up this one. Now you are all caught up!

### In what ways does the new logic improve things

First of all it removes the re-entrant problem described above by [explicitly detecting when changes would be overwritten](https://github.com/desktop/desktop/blob/0b4d244e610dd639e76429aa47ac037532985c46/app/src/lib/stores/app-store.ts#L3259-L3286) and acting upon it as it happens rather than relying on an error handler. I've already fixed one issue due to the re-entrant nature (https://github.com/desktop/desktop/pull/9038) and #9297 proved there's at least one more serious issue. My [comment from the original implementation](https://github.com/desktop/desktop/pull/7474#discussion_r286883784) which introduced this unfortunately proved to be well founded

> `_checkoutBranch` does become re-entrant in this scenario. I think we should still await here though, not awaiting will not alter the re-entrant nature of the logic as `performFailableOperation` doesn't itself await the result of the error handlers. I think instead we should focus on your second idea and explore that a little bit more but I definitely agree with your assessment of risk in doing so.
>
> I imagine that the fact that we have this disconnect in the awaitable chain will cause other problems such as the `StashAndSwitchBranch` progress not working properly. If you choose to move your changes to the next branch and hit Ok, the dialog progress will start spinning but if the checkout fails and the error handler takes over I would imagine that the dialog will close and the stash + checkout operation will continue in the background.
>
> My gut feeling here is that a better approach would be to have `_checkoutBranch` catch this specific error inline and deal with it accordingly rather than waiting for the global error handler to pick it up

Furthermore it ensures all different ways of checking out a branch (regular branch checkout, PR checkout, web links, etc) use the same consolidated logic.

Whether it achieves its goal of being easier to understand (and therefore also to maintain) is subjective but I absolutely think so. What was previously scattered across several flows in the application spanning hundreds of lines of code has now been reduced to a single linear flow that's easily debuggable. I think this is best illustrated by the top level logic:

https://github.com/desktop/desktop/blob/0b4d244e610dd639e76429aa47ac037532985c46/app/src/lib/stores/app-store.ts#L3192-L3196

The three different strategies for checking out a branch has now been extracted into three distinct methods that no longer share unrelated logic making it easier to reason about each individual one

https://github.com/desktop/desktop/blob/0b4d244e610dd639e76429aa47ac037532985c46/app/src/lib/stores/app-store.ts#L3207-L3212

### Needless stashing 

Thanks to this restructuring we're now able to only create a stash when it's really needed (i.e. when `git checkout branch` fails due to local changes being overwritten) whereas before there were scenarios (such as if there were deleted files in the working directory) where we'd always create a stash regardless of whether it was necessary.

## Release notes

Notes:
```
[Fixed] Let the user know when a checkout fails due to use of assume-unchanged or skip-worktree - #9297
[Fixed] Always show confirmation prompt before overwriting existing stash entry - #10956 
```


FelipeF1198 (commented):
FF

sakariye233:
#10957 
#11455 
#11269 
#9297
#8099

PR #10973, opened 2020-11-05
Release v2.5.7

niik:
## Description

This is a hotfix release on top of 2.5.6 which bumps dugite in order to get the latest Git LFS ([CVE-2020-27955](https://github.com/git-lfs/git-lfs/security/advisories/GHSA-4g4p-42wc-9f3m))

We're targeting a release on November 6th.

## Release checklist

- [x] Check to see if there are any errors in Sentry that have only occurred since the last production release
- [x] Verify that all feature flags are flipped appropriately
 - no new feature flags
- [x] If there are any new metrics, ensure that central and desktop.github.com have been updated
 - no new metrics

PR #9431, opened 2020-04-01
Use the non-fork repository on the commit message TokenParser

rafeca:
Closes #9334

This PR is based on https://github.com/desktop/desktop/pull/9340 since it uses the `getNonForkGitHubRepository` method created on that PR

## Description

This PR changes the `GitHub` repository object that's used by the commit message tokenizer to be able to link to PRs/issues on the upstream repository when working with a fork.

I've verified this by opening a fork of the `react` project and clicking on a PR link from a commit message. e.g:

Screenshot 2020-04-01 at 13 43 14

### Screenshots

N/A

## Release notes

Notes: [Fixed] Link to upstream repository from commit messages on forks.

PR #10977, opened 2020-11-06
Fix broken issues links in release notes

niik:
As I was upgrading to  2.5.7 (#10973) I noticed that the issue number in the release notes dialog wasn't hyperlinked as I would have expected it to be.

image

Turns out this was a bug due to us using an object conformant to a `GitHubRepository` which TypeScript is absolutely fine with but we were checking for an actual instance of `GitHubRepository`.

See https://github.com/desktop/desktop/blob/ef939e870341001c23f7ec439519c705b287b3a4/app/src/models/repository.ts#L110
See https://github.com/desktop/desktop/blob/ef939e870341001c23f7ec439519c705b287b3a4/app/src/lib/text-token-parser.ts#L66-L68

This was inadvertently broken in https://github.com/desktop/desktop/pull/9431


yvettep321:
_thank you _

PR #11006, opened 2020-11-10
Remove basic auth support for GitHub.com

niik:
## Description

In https://github.com/desktop/desktop/pull/10020 we added support for detecting whether GitHub.com supports API username password auth. This was done in preparation for the  [authorizations API brownouts](https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/). This Friday the deprecation timeline will move to its final phase and the authorizations endpoint will be removed entirely.

This PR reverts most the work done in #10020 as we no longer need it and hard-codes the GitHub.com endpoint such that the app always assumes that GitHub.com doesn't support username and password auth.

## Release notes

Notes: ???


billygriffin:
I'm not 100% sure it's necessary, but if it is, maybe something like:

`[Removed] Sign in to GitHub with username/password is no longer supported`

PR #11017, opened 2020-11-12
Increase title bar height on Big Sur

### LINKED ISSUE #10980: Big Sur Title bar
Minor issue but the new title bar on Big Sur looks a little out of place
Screenshot 2020-11-06 at 13 32 32

niik:
Closes #10980 

## Description

Seems macOS Big Sur (which is launching today) has a taller title bar making our traffic lights seem misaligned. Only judging from the screenshot in #10980 it looks like the title bar height increased by 4px so I've adjusted our title bar accordingly. Note that I haven't had a chance to actually test this on Big Sur yet.

While I was looking at this I noticed some needlessly complex logic to determine the current platform version which could be simplified thanks to https://github.com/electron/electron/pull/16599.

## Release notes

Notes: [Fixed] Adjust title bar height on Big Sur to match new platform convention

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -57,7 +57,6 @@
     "strip-ansi": "^4.0.0",
     "textarea-caret": "^3.0.2",
     "tslib": "^2.0.0",
-    "ua-parser-js": "^0.7.12",
     "untildify": "^3.0.2",
     "username": "^2.3.0",
     "uuid": "^3.0.1",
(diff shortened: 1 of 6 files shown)


niik:
I was able to install Big Sur in a VM and that was a good thing because it turns out Apple will lie about the version number to Electron saying it's 10.16 rather than 11.0 since electron (or chromium) isn't compiled against the Big Sur SDK, see https://github.com/electron/electron/issues/26419.

We're able to work around that though and here's the results (the dimmed traffic lights are from the production app)

image

I'm happy with that

PR #11019, opened 2020-11-12
Switch to using the dart version of sass instead of node-sass

niik:
Superseedes #10949

## Description

In #10949 @dennisameling graciously updated our version of node-sass to v5 which is required to move us forward with the Windows/arm64 work. This PR takes a slightly different approach by switching the engine out entirely in favor of the "pure JS" version of the sass compiler. In theory this could affect build performance negatively but if anything it appears to be faster from the few tests I've done though that could be related to the upgraded sass-loader as well.

See https://github.com/desktop/desktop/pull/10949#issuecomment-720538204

> Hey @dennisameling, thanks for opening this! I'd be happy to merge this as-is but I'm wondering if we should just switch out node-sass in favor of [sass](https://www.npmjs.com/package/sass) and avoid the native compilation step altogether. It seems that's what [webpack is recommending these days](https://webpack.js.org/loaders/sass-loader/)
> 
> > ℹ️ We recommend using Dart Sass.
> 
> I believe we'd need to swap `node-sass` for `sass` in addition to bumping `sass-loader` to the latest version.

This will rid us of one more native module which will be helpful with both the Windows ARM64 and the upcoming macOS ARM64 builds.


dennisameling (approved):
Thanks! Can confirm this works on arm64 Windows 🚀

PR #11115, opened 2020-11-24
Parse enums using runtime enum type defintion

niik:
## Description

Enums,  unlike a lot of TypeScript features, are available at runtime. We can leverage this to get rid of some tedious manual parsing logic when converting string enums to and from strings persisted in localStorage.


yvettep321:
_thank you _

PR #11124, opened 2020-11-25
Fix image swipe regression

### LINKED ISSUE #11120: Image Difference Swipe mode bugged
### Describe the bug

The image difference swipe mode is supposed to let you have a moveable, hard cutoff point on either side of which you can view the differences in an image. The current version locks the left edge of the "old version" of the image to the slider position, rather than the left edge being locked to the left edge of the viewing area (so the images can actually be compared).

### Version & OS

Version 2.6.0
Windows 10 Pro

1. Find an image in a commit that was changed (not added or deleted).
2. Select "Swipe" mode from the bottom right
3. Move the slider (the dot on the blue bar), you will see that the previous version of the image moves around with the slider instead of being locked to the left of the viewing area.

### Expected behavior

The image should not move around with the slider, instead, the previous and current versions of the image should occupy the same space in the viewing area, with the slider only hiding and revealing portions of each.

### Actual behavior

The current version of the image remains locked in place (as it should), but the previous version of the image moves around with its left edge locked to the slider's position.

### Screenshots
image
image
image

### Logs

### Additional context

I restarted github desktop and the problem persisted. I assume the error occurs for all supported image formats, but in my particular case it occurs with a .bmp file.

niik:
Closes #11120 

## Description

As part of fixing the long standing image diff scaling problems in #10904 I probably (not verified that it originates from #10904 but it's very likely) introduced a regression for the swipe mode of image diffs which was reported in #11120. 

This fixes the regression and also aligns the swipe mode even further with GitHub.com's swipe mode by making the slider start at the leftmost position and by having the current version of the image sit on top of the previous version.

### Before

20201125-before

### After

20201125-after

## Release notes

Notes: [Fixed] The image diff swipe mode overlays the two versions on top of each other again

### DIFF
--- a/app/src/ui/diff/image-diffs/swipe.tsx
+++ b/app/src/ui/diff/image-diffs/swipe.tsx
@@ -25,6 +25,14 @@ export class Swipe extends React.Component<
       width: this.props.maxSize.width,
     }
 
+    const swiperWidth = this.props.maxSize.width * (1 - this.state.percentage)
+
+    const currentStyle: React.CSSProperties = {
+      height: this.props.maxSize.height,
+      width: this.props.maxSize.width,
+      left: -(this.props.maxSize.width - swiperWidth),
+    }
+
     const maxSize: React.CSSProperties = {
       maxHeight: this.props.maxSize.height,
       maxWidth: this.props.maxSize.width,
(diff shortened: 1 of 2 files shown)


billygriffin (approved):
❤️

swipe5 (commented):
I

PR #11133, opened 2020-11-27
Refresh branches after creating a new branch

### LINKED ISSUE #11125: New branches aren't shown until the next refresh event
### Describe the bug

Unless an update occurs, a newly created local branch is not displayed within Desktop Branches drop down menu.

### Version & OS

Desktop 2.6.0,
MacOS  10.12.6

### Steps to reproduce the behaviour

1. Create a staged change that's awaiting commit
2. Open the branches drop down, and select `New Branch`
3. Provide name, and confirm you want to `Create Branch`
4. When prompted about your stage, hit cancel; don't switch branch.

Not switching to the new branch prevents the refresh required to display it in the branches menu.

### Expected behaviour

The branch now exists locally, but is not checked out. It shows in the branch menu, confirming to the user that it was created.

### Actual behavior

The branch now exists locally; it's visible in the CLI, and Desktop recognises it's presence if you attempt to create it a second time.

However, it's not visible within the branches menu until the user performs another action, such as switching repositories or removing focus from and returning focus to the Desktop window, which causes a refresh.

### Additional

This has the possibility to confuse the user by lack of creation confirmation. After pressing 'cancel' and seeing that the branch isn't in the branches list, it's easy to draw the conclusion that the branch wasn't created; yet you're informed it already exists when you try again.

niik:
Closes #11125

## Description

After creating a branch we now refresh the list of branches immediately rather than relying on checkoutBranch to do it for us as part of its `.refreshRepository`.

Since we always refresh the list of branches after creating a branch now we can just use that list rather than issuing a redundant for-each-ref command inside of the `createBranch` Git method.

## Release notes

Notes: [Fixed] Ensure newly created branches are shown even if the checkout is cancelled - #11125

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -340,16 +340,19 @@ export class GitStore extends BaseStore {
     startPoint: string | null,
     noTrackOption: boolean = false
   ) {
-    const branch =
-      (await this.performFailableOperation(() =>
-        createBranch(this.repository, name, startPoint, noTrackOption)
-      )) ?? null
+    const createdBranch = await this.performFailableOperation(async () => {
+      await createBranch(this.repository, name, startPoint, noTrackOption)
+      return true
+    })
 
-    if (branch !== null) {
+    if (createdBranch === true) {
       await this.loadBranches()
+      return this.allBranches.find(
+        x => x.type === BranchType.Local && x.name === name
+      )
     }
 
-    return branch
+    return undefined
   }
 
   public async createTag(name: string, targetCommitSha: string) {
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -66,6 +66,7 @@ import {
   getAllTags,
   deleteTag,
   MergeResult,
+  createBranch,
 } from '../git'
 import { GitError as DugiteError } from '../../lib/git'
 import { GitError } from 'dugite'
@@ -332,6 +333,23 @@ export class GitStore extends BaseStore {
     }
 
     this.storeCommits(commitsToStore)
+  }
+
+  public async createBranch(
+    name: string,
+    startPoint: string | null,
+    noTrackOption: boolean = false
+  ) {
+    const branch =
+      (await this.performFailableOperation(() =>
+        createBranch(this.repository, name, startPoint, noTrackOption)
+      )) ?? null
+
+    if (branch !== null) {
+      await this.loadBranches()
+    }
+
+    return branch
   }
 
   public async createTag(name: string, targetCommitSha: string) {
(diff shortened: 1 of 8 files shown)


app/src/lib/stores/git-store.ts
```diff
@@ -334,6 +335,23 @@ export class GitStore extends BaseStore {
...
+    if (branch !== null) {
+      await this.loadBranches()
+    }
+
+    return branch
+  }
+
   public async createTag(name: string, targetCommitSha: string) {
     const result = await this.performFailableOperation(async () => {
       await createTag(this.repository, name, targetCommitSha)
```
kuku3031, line 360: 3

kuku3031:
[Unicode whitespace](@) is a sequence of one |   | [Unicode whitespace](@) is a sequence of one |  
-- | -- | -- | --
  | or more [Unicode whitespace characters]. |   | or more [Unicode whitespace characters]. |  
  |   |   |   |  
  | A [space](@) is `U+0020`. |   | A [space](@) is `U+0020`. |  
  |   |   |   |  
  | A [non-whitespace character](@) is any character |   | A [non-whitespace character](@) is any character |  
  | that is not a [whitespace character]. |   | that is not a [whitespace character]. |  
  |   |   |   |  
  | An [ASCII punctuation character](@) |   | An [ASCII punctuation character](@) |  
  | is `!`, `"`, `#`, `$`, `%`, `&`, `'`, `(`, `)`, |   | is `!`, `"`, `#`, `$`, `%`, `&`, `'`, `(`, `)`, |  
 
  | `*`, `+`, `,`, `-`, `.`, `/`, `:`, `;`, `<`, `=`, `>`, `?`, `@`, |   | `*`, `+`, `,`, `-`, `.`, `/` (U+0021–2F), |  
  | `[`, `\`, `]`, `^`, `_`, `` ` ``, `{`, `\|`, `}`, or `~`. |   | `:`, `;`, `<`, `=`, `>`, `?`, `@` (U+003A–0040), |  
  |   |   | `[`, `\`, `]`, `^`, `_`, `` ` `` (U+005B–0060), |  
  |   |   | `{`, `\|`, `}`, or `~` (U+007B–007E). |  
  |   |   |   |  
  | A [punctuation character](@) is an [ASCII |   | A [punctuation character](@) is an [ASCII |  
  | punctuation character] or anything in |   | punctuation character] or anything in |  
  | the general Unicode categories  `Pc`, `Pd`, `Pe`, `Pf`, `Pi`, `Po`, or `Ps`. |   | the general Unicode categories  `Pc`, `Pd`, `Pe`, `Pf`, `Pi`, `Po`, or `Ps`. |  
  |   |   |   |  
  | ## Tabs |   | ## Tabs |  
  |   |   |   |  
  | Tabs in lines are not expanded to [spaces].  However, |   | Tabs in lines are not expanded to [spaces].  However, |  
  | in contexts where whitespace helps to define block structure, |   | in contexts where whitespace helps to define block structure, |  
  | tabs behave as if they were replaced by spaces with a tab stop |   | tabs behave as if they were replaced by spaces with a tab stop |  
  |   |   |   |

PR #11160, opened 2020-12-02
Fix invalid symmetric difference expression errors

niik:
## Description

If you take a look at your Developer Tools console for an instance of Desktop that's been running for a bit chances are you'll see a bunch of errors like these:

image

Turns out this is an incredibly common error but due to poor error management in the AheadBehindUpdater these errors end up not being handled and wind up as unhandled promise rejections. Given that we want to eventually make unhandled promise rejections act like regular uncaught errors and crash the app we need to address this.

Looking into the problem I found that the reason for these errors is that the AheadBehindUpdater gets confused about which repository it's calculating ahead/behind for. So when you're switching repositories we often end up trying to run `git rev-list X...Y` where `X` and `Y` are commits in the previously selected repository.

The design of the `AheadBehindUpdater` proved to unwieldy to fix the source of the problem in a succinct way so I took a note from the `CommitStatusStore` (https://github.com/desktop/desktop/pull/7163) and replicated that pattern here. What this means is that ahead/behind statuses are now requested by the individual branch list rows in the compare list _when they are mounted_. For a repository with hundreds of branches this means that we'll only calculate the ahead/behind status for the branches that are currently visible and we'll do so on demand. Since the branch you're looking to compare against is usually either the default branch or one of the branches in the recent branches group we'll avoid a lot of unnecessary computation this way which will be especially noticeable on Windows where these calls are rather expensive.

There's also a few performance improvements in here and low-hanging fruit fixes like not shelling out to `rev-list` when the shas of two branches match.

## Screenshots

ahead-behind-after

## Release notes

Notes: no-notes


sergiou87 (approved):
LGTM! Thanks for that last change, it's much clearer now 😌 

:shipit:

damecia:
WHY AM IM GETTING ALL THESE MESSAGES?


Sent from the all new Aol app for iOS


On Tuesday, December 22, 2020, 9:29 AM, Sergio Padrino <notifications@github.com> wrote:




@sergiou87 approved this pull request.

LGTM! Thanks for that last change, it's much clearer now 😌

PR #11162, opened 2020-12-02
Fix starting as maximized

### LINKED ISSUE #5631: Dark Theme Splash Screen is White
With dark theme selected, I think it'd be best to not have a blinding white light in fullscreen when opening the app. It only appears for a second, and I'm aware this isn't a bug, it's just an unimplemented feature. Below is a screenshot as is encouraged in issue tracking, however, as you can see, it's just white; not much to see...

image

Then after it opens I quickly get back to my nice dark theme:

image

But not until after I've been blinded. If it's possible to change this, I'm sure many of the dark theme users would appreciate it.

Thank you for listening and let me know if there's anything else I can do to report the problem accurately or fix the problem.

AndreiMaga:
Closes #5631

## Description

- The problem itself was kinda weird, if the window was saved as maximized, it would always load as a "splash screen" with the default color ```#fff``` before the other components were done loading.
- So the problem is actually ```this.savedWindowState.manage(this.window)``` that will call ```window.maximize()``` if the saved state of ```isMaximized``` and the option ```maximize``` (which is true by default) are true. That in itself wouldn't be a problem if ```window.maximize()``` would not also call ```window.show()```.
- And by maximizing only when the window should show solves the issue.

## Release notes

Notes: Full screen won't start as a white screen anymore

### DIFF
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -20,20 +20,20 @@ export class AppWindow {
   private minWidth = 960
   private minHeight = 660
 
-  private savedWindowState: windowStateKeeper.State | null = null
+  private shouldMaximizeOnShow: boolean = false
 
   public constructor() {
-    this.savedWindowState = windowStateKeeper({
+    const savedWindowState = windowStateKeeper({
       defaultWidth: this.minWidth,
       defaultHeight: this.minHeight,
       maximize: false,
     })
 
     const windowOptions: Electron.BrowserWindowConstructorOptions = {
-      x: this.savedWindowState.x,
-      y: this.savedWindowState.y,
-      width: this.savedWindowState.width,
-      height: this.savedWindowState.height,
+      x: savedWindowState.x,
+      y: savedWindowState.y,
+      width: savedWindowState.width,
+      height: savedWindowState.height,
       minWidth: this.minWidth,
       minHeight: this.minHeight,
       show: false,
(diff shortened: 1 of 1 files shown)


app/src/main-process/app-window.ts
```diff
@@ -204,7 +208,11 @@ export class AppWindow {
 
   /** Show the window. */
   public show() {
-    this.window.show()
+    if (this.shouldMaximizeOnShow) {
+      this.window.maximize() // maximize will also call show()
+    } else {
+      this.window.show()
+    }
```
niik: Looking at the documentation for `maximize` it says 

> Maximizes the window. This will also show (**but not focus**) the window if it isn't being displayed already.

And the docs for show says

> Shows **and gives focus** to the window.

Which makes me think we should always show and sometimes maximize the window. What do you think about something like this?

```suggestion
    this.window.show()
    if (this.shouldMaximizeOnShow) {
      this.window.maximize()
    }
```

niik (changes requested):
Thanks a bunch @AndreiMaga, I left one question which we need to look at before we can get this in but it's a nice improvement. Nice find!

app/src/main-process/app-window.ts
AndreiMaga: Oh, you're right it does not actually call show() but "shows" the window. And your fix looks great, I will commit it right now

app/src/main-process/app-window.ts
```diff
@@ -20,10 +20,13 @@ export class AppWindow {
   private minWidth = 960
   private minHeight = 660
 
+  private shouldMaximizeOnShow: boolean = false
```
niik: We don't typically annotate types that are easily inferable like this one. Let's also add a little pointer for our future selves.

```suggestion
  // See https://github.com/desktop/desktop/pull/11162
  private shouldMaximizeOnShow = false
```

niik (changes requested):
Thanks a lot @AndreiMaga. Gonna merge this with just a teeny change and a pointer back to this PR for future reference.

niik (approved):

PR #11169, opened 2020-12-03
Change revert commit message

### LINKED ISSUE #10515: "Revert" is ambiguous
Let's say I have the following commits (most recent at the top)

C. Broke stuff
B. More good stuff
A. Good stuff

I want to get my repo back to the state it was at B.

The current UI makes this confusing. Do I click on C and choose "revert" (meaning "I want to undo this bad commit) or do I click on B and choose "revert" (meaning "I want to go back to this point in time")?

I think it would be clearer if either:

1. the wording was changed to "revert to here" and the user was expected to click on B
or 2. The wording was changed to "revert the changes in this commit" and the user was expected to click on C.

I prefer option 1.

halitogunc:
Closes #10515 

### Screenshots
image

### DIFF
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -157,7 +157,9 @@ export class CommitListItem extends React.PureComponent<
 
     const items: IMenuItem[] = [
       {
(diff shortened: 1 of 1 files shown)


niik (approved):
Thanks a bunch @halitogunc!

PR #11195, opened 2020-12-07
Release v2.6.1-beta3

niik:
## Description

Looking for the PR for the upcoming third beta of the v2.6.1 series? Well you've just found it, congratulations! :tada:

## Release checklist

- [x] ~~Check to see if there are any errors in Sentry that have only occurred since the last production release~~
- [x] Verify that all feature flags are flipped appropriately
 - no feature flags
- [x] If there are any new metrics, ensure that central and desktop.github.com have been updated
  - no new metrics

PR #11198, opened 2020-12-08
Always wrap return values in promises when inside a Dexie transaction

### LINKED ISSUE #11192: Transaction committed too early
Extracting https://github.com/desktop/desktop/issues/11176#issuecomment-739507650 into its own issue

---

@hubgit: I'm getting a slightly different error after downloading the latest beta from the link above (which I have been using, not sure if I'd accidentally switched back to stable):

image

This is macOS 11.0.1, and I have been using GitHub Desktop for quite a long time.

Nothing's logged when running the script above in Dev Tools, so I guess it's not an undefined `parentID`.

It's working ok again after removing `~/Library/Application\ Support/GitHub\ Desktop`.

niik:
Closes #11192

## Description

Due to what I suspect is a bug in Dexie we must wrap our return values or we're more likely to run out of Dexies "micro-awaits" and cause a PrematureCommitError.

I was able to reliably reproduce #11192 after adding a lot of (50) local-only repositories to the app. I also tried upgrading to the latest version of Dexie where I couldn't reproduce it anymore but after digging into it some more it turns out the underlying problem wasn't solved, but the max "micro-awaits" had been lifted from 7 to 100 (https://github.com/dfahlander/Dexie.js/commit/3046e3684cb6092d31c634e36521364070a9f87b).

Problem was still reproducible running dexie@latest with ZONE_ECHO_LIMIT manually patched back down to 7.

There's a decent chance more of these instances exist in the app but this should at least get us back to a state better than or equal to 2.6.0.

## Release notes

Notes: [Fixed] Desktop would fail to launch for a small percentage of users that had several non-GitHub repositories in the app

### DIFF
--- a/app/src/lib/stores/issues-store.ts
+++ b/app/src/lib/stores/issues-store.ts
@@ -160,7 +160,8 @@ export class IssuesStore {
   ): Promise<ReadonlyArray<IIssueHit>> {
     const issues =
       this.queryCache?.repository.dbID === repository.dbID
-        ? this.queryCache?.issues
+        ? // Dexie gets confused if we return without wrapping in promise
+          await Promise.resolve(this.queryCache?.issues)
         : await this.getAllIssueHitsFor(repository)
 
     this.setQueryCache(repository, issues)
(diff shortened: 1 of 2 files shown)


niik:
Gonna land this quickly and get it into #11195.

PR #11222, opened 2020-12-13
Remove unused devtron dependency

roma2:
Closes #[issue number]

## Description

-

## Release notes

Notes:
It was added in https://github.com/desktop/desktop/pull/914 and made obsolete in https://github.com/desktop/desktop/pull/2457 but not properly cleaned up

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -66,7 +66,6 @@
     "winston-daily-rotate-file": "^1.4.5"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


niik:
Thanks for your submission.

It appears that you've created a pull request that based on an old, defunct branch. This is a pattern we often see from automated scripts so we're going to go ahead and close this. If it was intentional, please let us know what you were intending and we can see about reopening it.

Thanks again!

PR #11230, opened 2020-12-14
Replace "Enterprise Server" with "Enterprise"

billygriffin:
Since GitHub is launching additional variants of GitHub Enterprise (GitHub AE in particular) and AE should work with the existing implementation of GitHub Enterprise Server in Desktop, we want to avoid confusion by changing references to Enterprise Server to just Enterprise.

We're also adding AE to a few places where it's potentially helpful in explaining that it's also valid for AE in addition to Enterprise Server. For example:

Screen Shot 2020-12-16 at 3 49 50 PM

Notes: No notes (open to including a note if we think it's useful)

### DIFF
--- a/app/src/lib/email.ts
+++ b/app/src/lib/email.ts
@@ -44,7 +44,7 @@ export function lookupPreferredEmail(account: Account): string {
  */
 function isEmailPublic(email: IAPIEmail): boolean {
   // If an email doesn't have a visibility setting it means it's coming from an
-  // older Enterprise which doesn't have the concept of visibility.
+  // older Enterprise version which doesn't have the concept of visibility.
   return email.visibility === 'public' || !email.visibility
 }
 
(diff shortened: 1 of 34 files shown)


say25:
@billygriffin I think the Linter needs to be run

app/src/lib/email.ts
```diff
@@ -44,7 +44,7 @@ export function lookupPreferredEmail(account: Account): string {
  */
 function isEmailPublic(email: IAPIEmail): boolean {
   // If an email doesn't have a visibility setting it means it's coming from an
-  // older Enterprise Server which doesn't have the concept of visibility.
+  // older Enterprise which doesn't have the concept of visibility.
```
niik: Just to avoid any confusion for a reader let's make it a bit more explicit that it's not "an enterprise's email" but the version of Enterprise that they were running.

```suggestion
  // older Enterprise version which doesn't have the concept of visibility.
```

niik (changes requested):
Just one teeeny tiiiiiny suggestion

app/src/lib/email.ts
billygriffin: Makes total sense to me, thanks!

niik (approved):
✨

maya-ross:
@billygriffin is anything more needed to deploy this change out to customers?

PR #11241, opened 2020-12-16
Fix pull request list not updating

niik:
## Description

I noticed recently that the PR list for a repository was missing the most recent PR or two. It only happened intermittently but I was recently able to reproduce it and found that due to a bug in the `PullRequestCoordinator` (which has been there since the beginning) it would fail to emit an update when the PR list for a repository changed. Story time.

In #11063 we significantly reduced the number of times the repository store emits an update signaling to the rest of the app that the list of repositories has changed (i.e. a repository was added, removed, or its information was updated). See https://github.com/desktop/desktop/commit/e14db8639654da898f355ca9cc79a5f6d700c022#diff-efb525e0e38f508020a700c360ea248af1b2a6f0c5cb44f728830acbecbccbe3 for one such example. 

When the `PullRequestStore` refreshes the list of PRs for a repository it emits an event

https://github.com/desktop/desktop/blob/cf603e84a6dcf5516eac7b27df330aedf64ad110/app/src/lib/stores/pull-request-store.ts#L221

Which is then caught by the `PullRequestCoordinator` which figures out which `Repository` models needs to be updated as a result of the PR Store refreshing PRs for a `GitHubRepository` (several `Repository` models may share the same `GitHubRepository` either by being duplicate clones of the same repository or by one being a forked repo and the other being the parent).

https://github.com/desktop/desktop/blob/cf603e84a6dcf5516eac7b27df330aedf64ad110/app/src/lib/stores/pull-request-coordinator.ts#L82-L96

The problem here turned out to be that `this.repositories` was an empty array. The `repositories` field is updated whenever the `RepositoriesStore` emits an update:

https://github.com/desktop/desktop/blob/cf603e84a6dcf5516eac7b27df330aedf64ad110/app/src/lib/stores/pull-request-coordinator.ts#L50-L60

And here's the crux of the issue. Since #11063 made it that much rarer for the `RepositoriesStore` to emit an update you would need something to happen (add/remove/change a repository) before the `PullRequestCoordinator` kicked into gear and started forwarding these events properly.

I've addressed this by ensuring that the list of repositories is loaded when the coordinator initiates. I've also fixed a minor bug that I happened to note while debugging this where a non-primitive object (repo.owner) was coerced into a string in order to form the equality comparison hash.

## Release notes

Notes: [Fixed] Addressed an issue where pull requests could fail to update until the user switched repositories


app/src/lib/stores/pull-request-coordinator.ts
```diff
@@ -53,10 +51,18 @@ export class PullRequestCoordinator {
...
     // register an update handler for the repositories store
     this.repositoriesStore.onDidUpdate(allRepositories => {
-      this.repositories = allRepositories.filter(
-        isRepositoryWithGitHubRepository
+      this.repositories = Promise.resolve(
+        allRepositories.filter(isRepositoryWithGitHubRepository)
       )
     })
+
+    this.repositories = this.repositoriesStore
```
sergiou87, line 62: Is it worth adding a comment to indicate this triggers the initial load of repositories? 🤔

app/src/lib/stores/pull-request-coordinator.ts
```diff
@@ -53,10 +51,18 @@ export class PullRequestCoordinator {
...
+      .getAll()
+      .then(x => x.filter(isRepositoryWithGitHubRepository))
+      .catch(e => {
+        log.error(`PullRequestCoordinator: Error loading repositories`)
+        return []
+      })
```
sergiou87, line 68: **DISCLAIMER:** this comment is heavily biased by my lack of JS/TS experience and how I used to work with asynchronous code in previous codebases. Consider it just some simple "food for thought" where no actions are actually required.

While I'm amazed by the magic of Promises/Futures 🤩, it feels a bit misleading to me having a property that looks like an immediate value but holds a future value instead (and therefore you need to either `Promise.resolve` when setting immediate values, and `await` when reading its value).

I think my first approach would've been to keep the `repositories` property as an immediate value, and then store it here and then update all the PRs:
```typescript
    this.repositoriesStore
      .getAll()
      .then(x => {
        this.repositories = x.filter(isRepositoryWithGitHubRepository))
      })
      .catch(e => {
        log.error(`PullRequestCoordinator: Error loading repositories`)
        this.repositories = []
      })
      .finally(() => {
        this.handleRepositoriesUpdatedOrSomethingPlz()
      })
```

But of course `handleRepositoriesUpdatedOrSomethingPlz` lacks of some data that the `onPullRequestsChanged` handler has when it's invoked. Trying to compensate/fix that lack of data would involve some extra complexity (and of course there are many other scenarios where some race conditions would need to be handled), so I see how `await` solves the problem in a simpler and more elegant way. My questions are…
1. Is it worth renaming `repositories` to something more explicit of its non-immediate nature? (nope, I can't think of good name suggestions 😂)
1. Is this whole "use what seems to be a regular property as an async value" a common pattern in JS/TS or this codebase? 😳

sergiou87 (commented):
I always love a good story, thanks for the interesting read! ☕🤓

I've noticed one extra file got added to the repo accidentally, is that right?

Other than that, and a more philosophical question/comment I added, LGTM!

sergiou87 (approved):
Thanks for adding that comment! Looks much clearer now ❤️ 

:shipit:

app/src/lib/stores/pull-request-coordinator.ts
niik: We talked about this in zoom. It's a common practice in JS-land albeit a confusing one to reuse promises. Promises are always eagerly executed so they act as a memoization constructs when multiple consumers await them

PR #11244, opened 2020-12-16
Just use openPath on all platforms

niik:
## Description

The issue described in https://github.com/desktop/desktop/pull/2968#issuecomment-334629549 doesn't appear to exist any more and the openExternal solution fails when opening paths that ends with a backslash which I happened to notice when I accidentally clones a repository to a path ending with `\`

I'd bet the issue we were seeing  was  https://github.com/electron/electron/issues/10477 which has been long since resolved.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/shell.ts
+++ b/app/src/main-process/shell.ts
@@ -1,4 +1,3 @@
-import * as Url from 'url'
 import { shell } from 'electron'
 
 /**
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
LGTM, tested it and works like a charm here too 😌 :shipit:

PR #11263, opened 2020-12-21
Add detection of Sublime Text 4 and 2 as editor on macOS.

yurikoles:
Add detection of Sublime Text 4 and 2 as editor on macOS.
    
Signed-off-by: Yurii Kolesnykov <root@yurikoles.com>

## Description

Sublime Text Dev/4 is now being detected and recognized as editor.
I followed same code paths as for ST3, the only change is bundle id, which is `com.sublimetext.4` and I named it `Sublime Text Dev`, which is disputable, we may call it `Sublime Text 4`, or `Sublime Text Dev 4`

Sublime Text 4 is even unannounced on offifcial site yet, but some hackers have it installed already either from [offical ST Discord chat server](https://discord.gg/D43Pecu) or via my brew tap: 

```
brew install yurikoles/yurikoles/sublime-text-dev-4
```

### Screenshots

UI isn't changed

## Release notes

Notes:

### DIFF
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -47,7 +47,7 @@ function getBundleIdentifiers(editor: ExternalEditor): ReadonlyArray<string> {
     case ExternalEditor.VSCodium:
       return ['com.visualstudio.code.oss']
     case ExternalEditor.SublimeText:
-      return ['com.sublimetext.3']
+      return ['com.sublimetext.4', 'com.sublimetext.3', 'com.sublimetext.2']
     case ExternalEditor.BBEdit:
       return ['com.barebones.bbedit']
     case ExternalEditor.PhpStorm:


niik:
Thanks so much for getting the ball rolling on this @yurikoles!

> we may call it `Sublime Text 4`

Yeah, I think that's the better option here. Do you know roughly when Sublime 4 is estimated to ship?

I'm also wondering if we should follow the same approach as that of SlickEdit here in order to support multiple versions:

https://github.com/desktop/desktop/blob/f66cfe1795a225554c2072a50300f196f4077dc0/app/src/lib/editors/darwin.ts#L71-L77

If we added the new bundle id to the already existing SublimeText entry we'd automatically start launching Sublime 4 when users install it which seems like the expected action to me.

yurikoles:
Hi @niik, 

> If we added the new bundle id to the already existing SublimeText entry we'd automatically start launching Sublime 4 when users install it which seems like the expected action to me.

Good catch, I'm sorry that I didn't figure out this possibility. I applied your suggestion and added fallback to ST2.

niik (approved):
This looks great, thanks a bunch @yurikoles!

PR #11266, opened 2020-12-21
Prevent pruning remotes with branches available locally

sergiou87:
## Description

When a pull request from a fork is checked out, a remote prefixed with `github-desktop-` is added locally for the PR's branch. To prevent a clutter of remotes, the app will periodically prune old remotes from forks that are, apparently, not used anymore: if all PRs from a given remote have been closed, that remote will be removed.

However, it's possible that one of those remotes doesn't have any open PR, but branches from them are still available locally. In that case, the remote should be preserved.

This PR makes sure remotes from forked repos are not removed if a branch is available locally. It also adds some unit tests for the function in charge of deciding which remotes must be pruned.

It's important to know that local branches from merged PRs are also removed if they haven't been checked out in more than 2 weeks.

## Release notes

Notes: [Fixed] Remotes referencing forked repos would be removed even when branches from them were available locally


sergiou87:
The work here should be done, but @niik found an issue with `allBranches` from `GitStore`: it doesn't actually have *all* branches, because `getBranches` from `for-each-ref.ts` filters out branches from forked repos:

https://github.com/desktop/desktop/blob/510087a4d9b84ff5a2875a1e61481a94f311ba21/app/src/lib/git/for-each-ref.ts#L100-L106

I'll leave this as a draft until that is fixed, @niik is working on it! ❤️

niik:
> I'll leave this as a draft until that is fixed, @niik is working on it! ❤️

https://github.com/desktop/desktop/pull/11267 :sparkles:

app/src/lib/stores/git-store.ts
```diff
@@ -1647,18 +1648,15 @@ export class GitStore extends BaseStore {
...
-    for (const pr of openPRs) {
-      if (pr.head.gitHubRepository.cloneURL !== null) {
-        prRemotes.add(pr.head.gitHubRepository.cloneURL)
-      }
-    }
+    const remotesToPrune = findForkedRemotesToPrune(
+      remotes,
+      openPRs,
+      this.allBranches
+    )
```
niik: If we're being a little sneaky and grabbing a reference to allBranches here we can tighten this up a bit.

```suggestion
    const branches = this.allBranches
    const remotesToPrune = findForkedRemotesToPrune(remotes, openPRs, branches)
```

app/src/lib/stores/git-store.ts
```diff
@@ -1647,18 +1648,15 @@ export class GitStore extends BaseStore {
...
 
-    for (const r of remotes) {
-      if (r.name.startsWith(ForkedRemotePrefix) && !prRemotes.has(r.url)) {
-        await removeRemote(this.repository, r.name)
-      }
+    for (const r of remotesToPrune) {
+      await removeRemote(this.repository, r.name)
```
niik: Here on the other hand I think we can afford to be a bit more verbose. We typically use 1 char variables in anonymous functions but not so much in 
```suggestion
    for (const remote of remotesToPrune) {
      await removeRemote(this.repository, remote.name)
```

app/src/lib/stores/helpers/find-forked-remotes-to-prune.ts
```diff
@@ -0,0 +1,29 @@
+import { Branch } from '../../../models/branch'
+import { PullRequest } from '../../../models/pull-request'
+import { ForkedRemotePrefix, IRemote } from '../../../models/remote'
+
+/**
+ * Function to determine which remotes are not referenced anymore (by pull
```
niik: We should probably be clear that it's not all remotes, just Desktop's fork remotes that will be pruned.

app/test/unit/find-forked-remotes-to-prune-test.ts
```diff
@@ -0,0 +1,126 @@
...
+  gitHubRepository: GitHubRepository,
+  userName: string,
+  baseBranchName: string
+) {
+  return new PullRequest(
+    new Date(),
+    'desktop',
+    1,
+    {
+      ref: 'refs/heads/master',
```
niik: This won't matter for these tests but just in case this gets reused the format for the head and base ref is not qualified (i.e. it's 'main' rather than 'refs/heads/main'). And we try to avoid using `master` in new code (see https://github.com/desktop/desktop/pull/10833) (I know this was just copy-pasted from elsewhere but we might as well clean it up along the way).

```suggestion
      ref: 'main',
```

app/test/unit/find-forked-remotes-to-prune-test.ts
```diff
@@ -0,0 +1,126 @@
...
+      sha: 'deadbeef',
+      gitHubRepository,
+    },
+    {
+      ref: `refs/heads/${baseBranchName}`,
```
niik:
```suggestion
      ref: baseBranchName,
```

niik (changes requested):
Just a few small requests here but overall this looks great!

niik (approved):
✨ Thanks for picking this up!

PR #11267, opened 2020-12-21
Fix "Couldn't find branch" error when checking out a PR from a fork

niik:
## Description

When checking out a pull request from a fork prior to #11242 we attempted to look up whether a local branch existed already by looking for a branch named `pr/NUMBER`. This was brittle as it's possible that the user had already checked out the remote branch under another name, or conversely, that the user had a branch named "pr/NUMBER" which did not track the corresponding remote branch (@sergiou87 is fixing on the cause of the latter in https://github.com/desktop/desktop/pull/11266).

Unfortunately my testing of #11242 didn't uncover the fact that the `allBranches` property on `gitStore` didn't actually contain all branches. Forked remote branches were [intentionally excluded](https://github.com/desktop/desktop/blob/49a876e2250b208dc63fd9b7bbf272fffc1491b5/app/src/lib/git/for-each-ref.ts#L100-L106) in order not to clutter the interface. Not cluttering the interface is great but it's applied at too low of a level here.

In this PR I've lifted the filtering of remote fork branches to the presentation layer and audited all the dependents of `getBranches`.

A notable refactor here is the rename of `allBranches` to `branches` in `ICompareState`, that was done because `allBranches` was misleading given that it filtered not only remote fork branches but also the current branch.

I've also had to update the logic of the ahead-behind-updater so that it won't start calculating ahead/behind information for forked remote branches which aren't displayed in the UI (note that this will all be obsolete once #11160 lands).

## Release notes

Notes: [Fixed] Checking out a pull request from a fork for the first time now correctly identifies the remote branch to create the branch from.

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -672,9 +672,10 @@ export interface ICompareState {
   /** The SHAs of commits to render in the compare list */
   readonly commitSHAs: ReadonlyArray<string>
 
-  /** A list of branches (remote and local) except for Desktop fork remote
-   * branches (see `Branch.isDesktopForkRemoteBranch`) and the default branch
-   * currently in the repository. */
+  /**
+   * A list of branches (remote and local) except the current branch, and
+   * Desktop fork remote branches (see `Branch.isDesktopForkRemoteBranch`)
+   **/
   readonly branches: ReadonlyArray<Branch>
 
   /**
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -672,8 +672,10 @@ export interface ICompareState {
   /** The SHAs of commits to render in the compare list */
   readonly commitSHAs: ReadonlyArray<string>
 
-  /** A list of all branches (remote and local) currently in the repository. */
-  readonly allBranches: ReadonlyArray<Branch>
+  /** A list of branches (remote and local) except for Desktop fork remote
+   * branches (see `Branch.isDesktopForkRemoteBranch`) and the default branch
+   * currently in the repository. */
+  readonly branches: ReadonlyArray<Branch>
 
   /**
    * A list of zero to a few (at time of writing 5 but check loadRecentBranches
--- a/app/src/lib/stores/helpers/ahead-behind-updater.ts
+++ b/app/src/lib/stores/helpers/ahead-behind-updater.ts
@@ -104,7 +104,7 @@ export class AheadBehindUpdater {
     const newRefs = new Set<string>()
 
     const addBranch = (b: Branch | null) => {
-      if (b && b.isDesktopForkRemoteBranch && !cache.has(from, b.tip.sha)) {
+      if (b && !b.isDesktopForkRemoteBranch && !cache.has(from, b.tip.sha)) {
         newRefs.add(b.tip.sha)
       }
     }
--- a/app/src/lib/stores/helpers/ahead-behind-updater.ts
+++ b/app/src/lib/stores/helpers/ahead-behind-updater.ts
@@ -100,25 +100,22 @@ export class AheadBehindUpdater {
     this.clear()
 
     const from = currentBranch.tip.sha
+    const cache = this.comparisonCache
+    const newRefs = new Set<string>()
 
-    const filterBranchesNotInCache = (branches: ReadonlyArray<Branch>) => {
-      return branches
-        .map(b => b.tip.sha)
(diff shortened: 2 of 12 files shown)


app/src/lib/app-state.ts
```diff
@@ -672,8 +672,10 @@ export interface ICompareState {
   /** The SHAs of commits to render in the compare list */
   readonly commitSHAs: ReadonlyArray<string>
 
-  /** A list of all branches (remote and local) currently in the repository. */
-  readonly allBranches: ReadonlyArray<Branch>
+  /** A list of branches (remote and local) except for Desktop fork remote
+   * branches (see `Branch.isDesktopForkRemoteBranch`) and the default branch
+   * currently in the repository. */
```
sergiou87: > except … the **default** branch

Is it the **default** branch or the **current** branch? This code is using the **current** branch the array: https://github.com/desktop/desktop/pull/11267/files#diff-b238c18a99c08ab800e422715cc02105067f5f1394027f489e52d7fb15d6ca74R1030

🤔

app/src/lib/stores/helpers/ahead-behind-updater.ts
```diff
@@ -100,25 +100,22 @@ export class AheadBehindUpdater {
...
     const from = currentBranch.tip.sha
+    const cache = this.comparisonCache
+    const newRefs = new Set<string>()
 
-    const filterBranchesNotInCache = (branches: ReadonlyArray<Branch>) => {
-      return branches
-        .map(b => b.tip.sha)
-        .filter(to => !this.comparisonCache.has(from, to))
+    const addBranch = (b: Branch | null) => {
+      if (b && b.isDesktopForkRemoteBranch && !cache.has(from, b.tip.sha)) {
```
sergiou87: If I'm following this PR's logic correctly, shouldn't this filter *out* Desktop fork remote branches? 🤔 
```suggestion
      if (b && !b.isDesktopForkRemoteBranch && !cache.has(from, b.tip.sha)) {
```

app/src/lib/stores/helpers/ahead-behind-updater.ts
```diff
@@ -100,25 +100,22 @@ export class AheadBehindUpdater {
...
 
-    const otherBranches = [...recentBranches, ...allBranches]
-
-    const branches =
-      defaultBranch !== null ? [defaultBranch, ...otherBranches] : otherBranches
-
-    const newRefsToCompare = new Set<string>(filterBranchesNotInCache(branches))
+    addBranch(defaultBranch)
+    recentBranches.forEach(addBranch)
+    allBranches.forEach(addBranch)
```
sergiou87, line 114: Assuming I understand the intent of this, and given I needed to look for more implementation details, it's probably worth commenting here that we want recent branches to be scheduled first (since they show up first in the UI), and it's not a problem scheduling them again later with `allBranches` because they'll be in cache already and no extra work will be done.

Does that sound good?

app/src/lib/app-state.ts
niik: Yeah you're absolutely right, should be the **current branch**

app/src/lib/stores/helpers/ahead-behind-updater.ts
niik: > because they'll be in cache already and no extra work will be done.

And also because the `newRefs` is a set consisting of the tip SHA of each branch they won't actually be scheduled to be refreshed. I'm proposing we don't bother too much with this and instead focus on reviewing and merging #11160 which will remove the ahead-behind-updater entirely.

app/src/lib/stores/helpers/ahead-behind-updater.ts
niik: One day I'll learn how to computer. Great catch @sergiou87!

app/src/lib/stores/helpers/ahead-behind-updater.ts
sergiou87: Yep, I'm reviewing that one as we speak!

sergiou87 (approved):
Thanks for those changes! LGTM now :shipit: 

Should we wait until #11160 is merged first?

niik:
> Thanks for those changes! LGTM now :shipit:

thank-you-gif

Feels so good to have someone review my code and catch my mistakes again :heart:!

> Should we wait until #11160 is merged first?

Nah, let's get this in. We'll have a few easy-to-resolve conflicts either way. I'll land this and merge development into #11160

PR #11268, opened 2020-12-21
Change how initial path is set when launching PowerShell (Core) terminals

sergiou87:
## Description

This PR makes a couple of changes around setting the initial path when launching PowerShell (Core) terminals. Up until now, the app basically ran `Set-Location -LiteralPath` inside the new shell to switch to the initial directory. From now on…
- For PowerShell just rely on `cwd`.
- For PowerShell Core `cwd` is not enough, but we can use the `-WorkingDirectory` argument.

## Release notes

Notes: no-notes


niik (approved):
Love it!

PR #11280, opened 2020-12-23
Add dimension to track app launches from the Applications folder

sergiou87:
## Description

We suspect some weird issues on the macOS app happen because the app is not launched from the Applications folder. This PR adds a new measure to track how many macOS users actually launch the app from somewhere else.

**UPDATE:** after a discussion with @niik, I changed the code and the text above ☝️ to just report whatever `remote.app.isInApplicationsFolder` returns, and get that info directly from the `StatsStore` resulting in a much cleaner implementation 😌 

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -139,7 +139,6 @@ const DefaultDailyMeasures: IDailyMeasures = {
   diffOptionsViewedCount: 0,
   repositoryViewChangeCount: 0,
   unhandledRejectionCount: 0,
-  launchedFromDownloadsFolder: null,
 }
 
 interface IOnboardingStats {
@@ -1373,14 +1372,6 @@ export class StatsStore implements IStatsStore {
   public recordUnhandledRejection() {
     return this.updateDailyMeasures(m => ({
       unhandledRejectionCount: m.unhandledRejectionCount + 1,
-    }))
-  }
-
-  public recordLaunchedFromDownloadsFolder(
-    launchedFromDownloadsFolder: boolean
-  ) {
-    return this.updateDailyMeasures(m => ({
-      launchedFromDownloadsFolder,
     }))
   }
 
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -488,9 +488,9 @@ export class StatsStore implements IStatsStore {
     ).length
     const diffMode = getShowSideBySideDiff() ? 'split' : 'unified'
 
-    const launchedFromApplicationsFolder = __DARWIN__
-      ? remote.app.isInApplicationsFolder()
-      : null
+    // isInApplicationsFolder is undefined when not running on Darwin
+    const launchedFromApplicationsFolder =
+      remote.app.isInApplicationsFolder?.() ?? null
 
     return {
       eventType: 'usage',
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -139,6 +139,7 @@ const DefaultDailyMeasures: IDailyMeasures = {
   diffOptionsViewedCount: 0,
   repositoryViewChangeCount: 0,
   unhandledRejectionCount: 0,
+  launchedFromDownloadsFolder: null,
 }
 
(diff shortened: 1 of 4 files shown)


app/src/ui/app.tsx
```diff
@@ -295,6 +295,15 @@ export class App extends React.Component<IAppProps, IAppState> {
...
     log.info(`execPath: '${process.execPath}'`)
+
+    const launchedFromDownloadsFolder = this.isLaunchedFromDownloadsFolder(
+      process.execPath
+    )
+    if (launchedFromDownloadsFolder !== null) {
+      this.props.dispatcher.recordLaunchedFromDownloadsFolder(
+        launchedFromDownloadsFolder
+      )
+    }
```
sergiou87: Is this the best place to do this? 🤔

damecia:
STOP with these messages I DONT WANT TO RECEIVE THEM 


Sent from the all new Aol app for iOS


On Wednesday, December 23, 2020, 10:08 AM, Sergio Padrino <notifications@github.com> wrote:




@sergiou87 commented on this pull request.

In app/src/ui/app.tsx:
> +    const launchedFromDownloadsFolder = this.isLaunchedFromDownloadsFolder(
+      process.execPath
+    )
+    if (launchedFromDownloadsFolder !== null) {
+      this.props.dispatcher.recordLaunchedFromDownloadsFolder(
+        launchedFromDownloadsFolder
+      )
+    }

Is this the best place to do this? 🤔

SonyViao (approved):

app/src/lib/stats/stats-store.ts
```diff
@@ -481,6 +488,10 @@ export class StatsStore implements IStatsStore {
     ).length
     const diffMode = getShowSideBySideDiff() ? 'split' : 'unified'
 
+    const launchedFromApplicationsFolder = __DARWIN__
+      ? remote.app.isInApplicationsFolder()
+      : null
```
sergiou87: I wondered if I should check `isInApplicationsFolder` too, like this:
```suggestion
    const launchedFromApplicationsFolder = __DARWIN__ && remote.app.isInApplicationsFolder
      ? remote.app.isInApplicationsFolder()
      : null
```

But decided not to do it because it feels "too defensive". Being on macOS and that method `undefined` should just not happen. Does that sound 👍 ?

sergiou87: Also, is it ok to just import and use `remote.app` from here directly? I come from a codebase where we had dependency injection EVERYWHERE, so this "feels bad"… but, of course, it's very convenient 😂  Thoughts?

niik: > I wondered if I should check isInApplicationsFolder too, like this:

Eh, I'd approve either option but I'd add a comment so that future readers know why we're special casing. If you wanted to you could use optional chaining and nullish coalescing here:

```suggestion
    // isInApplicationsFolder is undefined when not running on Darwin
    const launchedFromApplicationsFolder =
      remote.app.isInApplicationsFolder?.() ?? null
```

> Also, is it ok to just import and use `remote.app` from here directly? I come from a codebase where we had dependency injection EVERYWHERE, so this "feels bad"… but, of course, it's very convenient 😂 Thoughts?

Yup! Totally okay! We don't use DI. We used to in the classic app but it got so abstract and caused so much boilerplate that we felt heavily constrained by it and  for the electron app we just haven't seen any compelling reason to (yet).

niik (changes requested):
This looks great.  I'll let you decide on whether to check if `isInApplicationsFolder` is defined or not (which really should be reflected in the electron type definitions IMO) but I'd like to add that comment regardless.

app/src/lib/stats/stats-store.ts
sergiou87: > If you wanted to you could use optional chaining and nullish coalescing

Coalescing undefined to null? 🤯  What a time to be alive 😂 

> We used to in the classic app but it got so abstract and caused so much boilerplate that we felt heavily constrained by it

I totally know that feel 😐 🤗

Thanks for the feedback! ❤️

niik (approved):
:sparkles:

PR #11332, opened 2021-01-05
Update repository after changing fork behavior

### LINKED ISSUE #11327: Unable to set 'Fork behaviour' to 'For my own purposes'
### Describe the bug

I have forked a public repository and cloned to GitHub Desktop, when prompted with the 'Fork behaviour' dialog, I selected 'For my own purposes'.

However, if I then go to 'Repository -> Fork behaviour' the 'To contribute to the parent repository' option is selected. If I then select the 'For my own purposes' option, click 'Save', and then re-open the dialog, the incorrect 'To contribute to the parent repository' option is selected again.

### Version & OS

Microsoft Windows 10 Home 10.0.19041
GitHub Desktop Version 2.6.1

1. Go to 'Repository -> Fork behaviour'
2. Click on 'For my own purposes'
3. Click 'Save'
4. Go to 'Repository -> Fork behaviour'
5. See that the incorrect option is selected

### Expected behaviour

I would expect the 'For my own purposes' option to be selected, or an error message if there was a failure with the selection.

### Actual behaviour

The 'To contribute to the parent repository' option is selected.

### Screenshots

### Logs

### Additional context

Forking the https://github.com/ARMmbed/mbed-os repository.

sergiou87:
Closes #11327

## Description

Not sure if this was broken in a refactor or what happened exactly, but from what I've seen, when the user changes the fork behavior (either right after cloning a fork repo or from the `Repository Settings` dialog), the app updates the database as expected… but doesn't update the "in-memory" repository instance with the new data.

This means the app needs to be restarted in order to get a `Repository` instance with the new fork behavior.

When the `RepositoryStore` emits an update, the `AppStore` checks if the current repository has changed, and if so it reloads all the repository info. However, in order to detect a repository change, it only checks a change in the hash or a change from no-repo selected to some repo selected and viceversa.

The changes in this PR make sure the app also checks for fork behavior changes to trigger a reload of the repo info.

**UPDATE**: thanks to @niik ❤️ I learned that those workflow preferences were already taken into account in the hash (which is checked to detect repository changes), however due to an overlook, the `workflowPreferences` object was converted to a `[Object object]` string, which is not very useful.

My final changes basically implement a suggestion from @niik where the array of items used to build the hash is typed to prevent the use of non-primitive objects (that would end up as `[Object object]`).

## Release notes

Notes: [Fixed] Fork behavior changes are now reflected in the app immediately.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1817,6 +1817,31 @@ export class AppStore extends TypedBaseStore<IAppState> {
     })
   }
 
+  private hasRepositoryChanged(
+    oldRepository: Repository | CloningRepository | null,
+    newRepository: Repository | CloningRepository | null
+  ) {
+    if (oldRepository && newRepository) {
+      // If this is a GitHub repository, check if the workflow preferences
+      // changed too.
+      if (
+        oldRepository instanceof Repository &&
+        isRepositoryWithGitHubRepository(oldRepository) &&
+        newRepository instanceof Repository &&
+        isRepositoryWithGitHubRepository(newRepository) &&
+        oldRepository.workflowPreferences !== newRepository.workflowPreferences
+      ) {
+        return true
+      }
+
+      return oldRepository.hash !== newRepository.hash
+    }
+
+    return (
+      (oldRepository && !newRepository) || (!oldRepository && newRepository)
+    )
+  }
+
   private updateRepositorySelectionAfterRepositoriesChanged() {
     const selectedRepository = this.selectedRepository
     let newSelectedRepository: Repository | CloningRepository | null = this
@@ -1844,13 +1869,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       }
     }
 
-    const repositoryChanged =
-      (selectedRepository &&
-        newSelectedRepository &&
-        selectedRepository.hash !== newSelectedRepository.hash) ||
-      (selectedRepository && !newSelectedRepository) ||
-      (!selectedRepository && newSelectedRepository)
-    if (repositoryChanged) {
+    if (this.hasRepositoryChanged(selectedRepository, newSelectedRepository)) {
       this._selectRepository(newSelectedRepository)
       this.emitUpdate()
     }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1822,8 +1822,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
     newRepository: Repository | CloningRepository | null
   ) {
     if (oldRepository && newRepository) {
-      // If both are GitHub repositories, check if the workflow preferences
-      // are different too.
+      // If this is a GitHub repository, check if the workflow preferences
+      // changed too.
       if (
         oldRepository instanceof Repository &&
         isRepositoryWithGitHubRepository(oldRepository) &&
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1817,31 +1817,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     })
   }
 
-  private hasRepositoryChanged(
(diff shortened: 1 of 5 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -1817,6 +1817,31 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    newRepository: Repository | CloningRepository | null
+  ) {
+    if (oldRepository && newRepository) {
+      // If this is a GitHub repository, check if the workflow preferences
+      // changed too.
+      if (
+        oldRepository instanceof Repository &&
+        isRepositoryWithGitHubRepository(oldRepository) &&
+        newRepository instanceof Repository &&
+        isRepositoryWithGitHubRepository(newRepository) &&
```
sergiou87: Is there a cleaner way of doing this repo downcasting? 🤔

app/src/models/hash.ts
```diff
@@ -0,0 +1,5 @@
+type HashableType = number | string | boolean | undefined | null
+
+export function createHash(items: HashableType[]) {
```
sergiou87: I was wondering if there is an easy way of specifying a type (that is not a class) to enforce that Hashes are created with this function.

Something like `type Hash = string` doesn't work because you can still create a `string` however you want and use it instead. Does TypeScript have anything to help with this? 🤔

app/src/models/hash.ts
```diff
@@ -0,0 +1,5 @@
+type HashableType = number | string | boolean | undefined | null
```
niik:
```suggestion
/**
  Types which can safely be coerced to strings without loosing information
   As an example `1234.toString()` doesn't loose any information whereas
   `({ foo: bar }).toString()` does (`[Object object]`)
**/
type HashableType = number | string | boolean | undefined | null
```

app/src/models/hash.ts
niik: Thanks so much for doing this @sergiou87. Should prevent us from similar mistakes in the future. Two small thoughts. 1) Could we make the argument list dynamic to get rid of the `[]` brackets?


And second, could we add a little documentation to the method? We try to do so for all exported or otherwise public methods.

```suggestion
/**
  Creates a string representation of the provided arguments.
  
  This is a helper function used to create a string representation of
  an object based on its properties for the purposes of simple equality
  comparisons.
**/
export function createHash(...items: HashableType[]) {
```

Lastly, I wonder if we should call this method `createEqualityHash` or something to differentiate it from [createHash](https://nodejs.org/api/crypto.html#crypto_crypto_createhash_algorithm_options).

niik (changes requested):
Thanks for fixing this  @sergiou87, just a couple of requests

app/src/models/hash.ts
niik: No, not that I'm aware of. In TypeScript types are considered equal if they're semantically equivalent which leads to hacks like https://github.com/desktop/desktop/blob/d008892180353fd8e4417cc5c3c8478b9a6aafa3/app/src/lib/api.ts#L112-L122 in order to prevent an `IAPIRepository` from being passed as an `IAPIFullRepository`.

This is particularly easy to get tripped up on when dealing with types containing only optional fields since they're essentially assignable  by any object which shares at least one compatible field signature.

sergiou87: Oh, wow, that's a good one!! 🤯 Very interesting, thanks!

app/src/models/hash.ts
sergiou87: Is the use of "loosing" and "loose" intentional or you meant "losing" and "lose"? 🤔

sergiou87:
@niik I made the changes you suggested (assuming you wanted to say lose instead of loose) and renamed `hash.ts` to `equality-hash.ts` for consistency. Happy to change it back if you prefer `hash.ts`!

niik:
> assuming you wanted to say lose instead of loose

yuuuup!

niik (approved):
Thanks for picking this up. Should hopefully save us from similar mistakes in the future

PR #11345, opened 2021-01-08
Fix summary & description fields deleting user's text (sometimes) after a commit

### LINKED ISSUE #7251: Commit Message Clears When Committing in Stages
## Description
Commit Message Clears When Committing

## Version
* GitHub Desktop: 1.7.0-beta2 (tested this still happens in 1.6.5)

## Steps to Reproduce
1. Change a number of files.
2. Commit a subset of files.
3. Start to commit more files and start writing a commit message and see that it is cleared.

### Expected Behavior
It doesn't clear my commit messages.

### Actual Behavior
It clears my commit messages.

## Additional Information
This generally happens more often during high system CPU and memory load.

sergiou87:
Closes #7251

## Description

This PR tackles a problem where the component `CommitMessage` could clear the summary and description textfields a few seconds after the commit completed, when the user might already have written something for their next commit.

The problem was `CommitMessage` relied on the `repository.state.isCommitting` attribute to enable/disable the textfields, but wouldn't clear the textfields until the commit + some additional tasks were finished. So the timeline would look something like this:
1. Set `repository.state.isCommitting = true` (**textfields are now disabled**)
1. Do commit work
1. Set `repository.state.isCommitting = false` (**textfields are now enabled**)
1. Do additional commit work (refreshing repository and change list).
1. Clear the textfields.

This bug surfaces when step (4) takes too long, which is when the user might be preparing their next commit message, just to see how it gets deleted in step (5).

With this PR, the timeline of events looks like this:
1. Set `repository.state.isCommitting = true` (**textfields are now disabled**)
1. Do commit work
1. Do additional commit work (refreshing repository and change list).
1. Set `repository.state.isCommitting = false` (**textfields are now enabled**)
1. Clear the textfields.

Now the the textfields will be cleared right after enabling them again.

## Release notes

Notes: [Fixed] Commit messages won't be cleared while the user is writing them.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2376,90 +2376,106 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const gitStore = this.gitStoreCache.get(repository)
 
-    const result = await this.isCommitting(repository, () => {
-      return gitStore.performFailableOperation(async () => {
+    return this.withIsCommitting(repository, async () => {
+      const result = await gitStore.performFailableOperation(async () => {
         const message = await formatCommitMessage(repository, context)
-        return createCommit(repository, message, selectedFiles)
+        return createCommit(repository, message, selectedFiles) !== undefined
       })
+
+      if (result) {
+        await this._recordCommitStats(
+          gitStore,
+          repository,
+          state,
+          context,
+          files
+        )
+
+        await this._refreshRepository(repository)
+        await this.refreshChangesSection(repository, {
+          includingStatus: true,
+          clearPartialState: true,
+        })
+      }
+
+      return result || false
     })
+  }
 
-    if (result) {
-      this.statsStore.recordCommit()
+  private async _recordCommitStats(
+    gitStore: GitStore,
+    repository: Repository,
+    repositoryState: IRepositoryState,
+    context: ICommitContext,
+    files: readonly WorkingDirectoryFileChange[]
+  ) {
+    this.statsStore.recordCommit()
 
-      const includedPartialSelections = files.some(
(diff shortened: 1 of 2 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -3562,14 +3578,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
     })
   }
 
-  private async isCommitting(
+  private async withIsCommitting(
     repository: Repository,
-    fn: () => Promise<string | undefined>
-  ): Promise<boolean | undefined> {
+    fn: () => Promise<boolean>
+  ): Promise<boolean> {
```
sergiou87, line 3584: Aside from renaming this function as @niik suggested, I also changed the return type to (hopefully) simplify the code, specially after moving things around.

app/src/ui/changes/changes-list.tsx
```diff
@@ -776,7 +776,10 @@ export class ChangesList extends React.Component<
           selectedRows={this.state.selectedRows}
           selectionMode="multi"
           onSelectionChanged={this.props.onFileSelectionChanged}
-          invalidationProps={this.props.workingDirectory}
+          invalidationProps={{
+            workingDirectory: this.props.workingDirectory,
+            isCommitting: this.props.isCommitting,
+          }}
```
sergiou87, line 782: After the changes in this PR, I noticed the checkboxes of the files in the change list stayed disabled. Putting the `_refreshDirectory` and `refreshChangesSection` inside the `withIsCommitting` block meant that the `workingDirectory` prop would change *before* `isCommitting` is set to `false`. And when that happens, it wouldn't trigger a render on the list.

niik (approved):
high-five

PR #11368, opened 2021-01-11
Improve git error API to show the combined stdout and stderr outputs

### LINKED ISSUE #6403: Use monospace font for reporting errors from git hooks
Usually the error message thrown by git hooks was mean to shown on terminal with monospace font, or we will break the text layout:

photo_2018-12-12_20-18-32

sergiou87:
Closes #6403 #5439

## Description

This PR changes the git error API in a way that allows us to show the full output of the git command on errors. Thanks to this, git hooks will now also show the full output in monospace font, making them easier to read.

With these changes, `logCommitError` is also unnecessary, which let us clean up the `createCommit` and `createMergeCommit` functions a bit.

In order to parse the output of git commands without losing the `\r` characters, we needed to use [split2](https://www.npmjs.com/package/split2), since [byline](https://github.com/jahewson/node-byline) (already used in the project), [uses `\r` as line separator too](https://github.com/jahewson/node-byline/blob/master/lib/byline.js#L103).

## Release notes

Notes: [Improved] Show full output of git hooks on errors


app/src/lib/stores/app-store.ts
```diff
@@ -2379,7 +2379,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return this.withIsCommitting(repository, async () => {
       const result = await gitStore.performFailableOperation(async () => {
         const message = await formatCommitMessage(repository, context)
-        return createCommit(repository, message, selectedFiles) !== undefined
+        await createCommit(repository, message, selectedFiles)
+        return true
```
sergiou87: Since `createCommit` can't be `undefined` anymore, I simplified this to just return `true` when it doesn't fail (otherwise, `result` will be `undefined` thanks to `performFailableOperation`). Any suggestions to make this cleaner?

app/test/unit/git/tag-test.ts
```diff
@@ -166,7 +166,7 @@ describe('git/tag', () => {
 
       await checkoutBranch(repository, account, branch!)
       const commitSha = await createCommit(repository, 'a commit', files)
-      await createTag(repository, 'my-new-tag', commitSha!)
```
sergiou87, line 169: Since `createCommit` doesn't return `undefined` anymore, this is not needed, right? :_D

niik: Right!

app/src/ui/app-error.tsx
```diff
@@ -107,14 +107,23 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
...
+    const e = getUnderlyingError(error)
 
     if (e instanceof GitError) {
-      // See getResultMessage in core.ts
-      // If the error message is the same as stderr or stdout then we know
-      // it's output from git and we'll display it in fixed-width font
-      if (e.message === e.result.stderr || e.message === e.result.stdout) {
+      // If the error message is just the raw git output, display it in
+      // fixed-width font
+      if (e.isRawMessage) {
```
niik: It'd be nice if we could share the logic in isRawGitError between this method and the render method so that it don't accidentally drift apart over time, similar to how `isErrorWithMetaData`,`isGitError`, and `isCloneError` works.

app/src/lib/stores/app-store.ts
niik: You could return `createCommit` directly (without even awaiting), and then check if result is not undefined (as opposed to the truthy check that's in place right now. That might be a bit cleaner.

niik (changes requested):
This is looking great @sergiou87, just a few small comments

app/src/lib/stores/app-store.ts
sergiou87: > and then check if result is not undefined (as opposed to the truthy check that's in place right now

~~That's what we had before (since #11345, more specifically). But with the changes in this PR, `createCommit` will never return `undefined`, since that could only happen when throwing an exception and `performFailableOperation` would handle that one.~~

~~I changed it because it felt wrong / dishonest, meaning it could lead you to think it returns `undefined` but it doesn't.~~

~~Anyway, happy to go ahead with the `!== undefined` if you find it cleaner!~~

I read your comment a couple of times more and understood what you meant, sorry 😂  Changing…

sergiou87: Hopefully this looks ok! f7efae756426670ffcea26981d0580d4ce2f8a51

app/src/lib/stores/app-store.ts
```diff
@@ -2379,7 +2379,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return this.withIsCommitting(repository, async () => {
       const result = await gitStore.performFailableOperation(async () => {
         const message = await formatCommitMessage(repository, context)
-        return createCommit(repository, message, selectedFiles) !== undefined
+        return createCommit(repository, message, selectedFiles)
       })
 
       if (result) {
```
niik: Since result is a string (`string | undefined`) here now we should be explicit or else an empty string and `undefined` will be treated the same.

```suggestion
      if (result  !== undefined) {
```

app/src/lib/stores/app-store.ts
```diff
@@ -2398,7 +2398,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         })
       }
 
-      return result || false
+      return !!result
```
niik: Like above, explicit comparisons when there's ambiguity in the truthy/falsy checks

```suggestion
      return result !== undefined
```

niik (changes requested):
This is coming together really well @sergiou87! 

Before we merge this could you please update the logic in `scrollToBottomOfGitErrorMessage` to use `isRawGitError` as well so that we auto-scroll to the bottom of git errors?

sergiou87:
Done @niik! Thanks for the thorough review! ❤️ 🙇‍♂️🙇‍♂️🙇‍♂️

niik (approved):
well_done_sir

PR #11278, opened 2020-12-23
Contain focus in dropdowns using FocusTrap

sergiou87:
## Description

This PR uses a `FocusTrap` to make sure the focus is contained within dropdowns when pressing tab.

It's a trap

### Screenshots

Before this PR:
ezgif com-gif-maker

After this PR:
ezgif com-gif-maker (1)

*theme change not included

## Release notes

Notes: [Fixed] Pressing tab multiple times in a dropdown doesn't focus on outside elements anymore

PR #11371, opened 2021-01-11
Fix crash clicking menus on Windows: disable FocusTrap for menus

sergiou87:
## Description

With #11278 I introduced a regression in the Windows app where clicking on a menu would crash the app.

The problem is those menus are also `ToolbarDropdown` elements but the items are not "focusable" (at least [not as `FocusTrap` expects](https://github.com/focus-trap/tabbable/blob/master/src/index.js#L1-L13)). Since those menus have their own custom "focus trap", I've disabled the one used in other dropdowns.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/app-menu/app-menu-bar-button.tsx
+++ b/app/src/ui/app-menu/app-menu-bar-button.tsx
@@ -193,6 +193,11 @@ export class AppMenuBarButton extends React.Component<
         dropdownState={dropDownState}
         onDropdownStateChanged={this.onDropdownStateChanged}
         dropdownContentRenderer={this.dropDownContentRenderer}
+        // Disable the dropdown focus trap for menus. Items in the menus are not
+        // "tabbable", so the app crashes when this prop is set to true and the
+        // user opens a menu (on Windows).
+        // Besides, we use a custom "focus trap" for menus anyway.
+        focusTrapActive={false}
         showDisclosureArrow={false}
         onMouseEnter={this.onMouseEnter}
         onKeyDown={this.onKeyDown}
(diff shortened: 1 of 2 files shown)


app/src/ui/toolbar/dropdown.tsx
```diff
@@ -75,6 +75,9 @@ export interface IToolbarDropdownProps {
   /** The button's style. Defaults to `ToolbarButtonStyle.Standard`. */
   readonly style?: ToolbarButtonStyle
 
+  /** Wether the dropdown will trap focus or not. Defaults to true. */
+  readonly focusTrapActive?: boolean
```
sergiou87: I'm open to suggestions for better names 😂

app/src/ui/app-menu/app-menu-bar-button.tsx
```diff
@@ -193,6 +193,11 @@ export class AppMenuBarButton extends React.Component<
         dropdownState={dropDownState}
         onDropdownStateChanged={this.onDropdownStateChanged}
         dropdownContentRenderer={this.dropDownContentRenderer}
+        // Disable the dropdown focus trap for menus. Items in the menus are not
+        // "tabbable", so the app crashes when this prop is set to true and the
+        // user opens a menu (on Windows).
+        // Besides, we use a custom "focus trap" for menus anyway.
```
niik, line 199: Love love love this explanation for the future reader!

app/src/ui/toolbar/dropdown.tsx
niik: We typically use `enableX` or `disableX`, see https://github.com/desktop/desktop/blob/1c488e94f56671535ea5de2dbcccd5d13d0c434f/app/src/ui/app-menu/app-menu.tsx#L27-L32 for an example.

So probably
```suggestion
  readonly enableFocusTrap?: boolean
```

niik (changes requested):

niik (approved):
Nice fix!

PR #11387, opened 2021-01-13
Branch fast-forwarding improvements: round 1

sergiou87:
## Description

This PR is the first step to improve the branch auto fast-forwarding mechanism in the app.

Before the changes in this PR, the only criteria to decide which branches we should _try_ to fast-forward looks for local branches with upstream that are not the current branch. If there are more than 20 branches matching that criteria in the whole repo, we would just filter the recent branches by that criteria. With such a "light" condition to be eligible, it was easy to get more than 20 branches, where some of them might be actually "fast-forwardable" but could be left out due to the app falling back to just processing the recent branches.

With this PR, the criteria gets a bit smarter: we use `git for-each-ref` to look for local branches whose SHA differs from their remote branches' SHA. This gives us a much narrower list of eligible branches, where only those which have diverged or are ahead of their remote won't be fast-forwarded.

A good way to create an environment to test these changes is running this:
```bash
git clone https://github.com/ARMmbed/mbed-os.git
cd mbed-os

# Move ref some commits behind
git checkout feature-bluetooth-traces
git reset --hard dd4eca1b693c253626481fb400a9451458f1f161

# Move ref some commits behind
git checkout feature-cmake
git reset --hard 66f65c0e89dea555993416376dd980df23d66d48

# Move ref to another ref (X commits behind and Y commits ahead)
git checkout feature-i2c
git reset --hard feature-wisun-1.0.0

# Create branch without upstream
git branch no-upstream-branch 61f2097223060b6cebd2246dde3318fa0a127a5f

# Check out branch and don't move it
git checkout feature-hal-spec-sai
```

This will clone a repo and move some branches around, so it should end up fast-forwarding `feature-bluetooth-traces` and `feature-cmake`.

## Release notes

Notes: [Fixed] When fetching a repository, the app will fast-forward all possible branches (except the current branch)


app/src/lib/git/for-each-ref.ts
```diff
@@ -101,3 +101,104 @@ export async function getBranches(
...
+): Promise<ReadonlyArray<Branch>> {
+  const format = [
+    '%(refname)',
+    '%(refname:short)',
+    '%(objectname)', // SHA
+    '%(upstream)',
+    '%(symref)',
+  ].join('%00')
+
+  const prefixes = ['refs/heads', 'refs/remotes']
```
sergiou87, line 127: Aside from copy&pasting 😆  it felt clearer to have this in a separated array. Happy to write these inline though.

app/src/lib/git/for-each-ref.ts
```diff
@@ -101,3 +101,104 @@ export async function getBranches(
...
+ * @param currentBranchName Name of the current branch.
+ */
+export async function getBranchesDifferingFromUpstream(
+  repository: Repository,
+  allBranches: ReadonlyArray<Branch>,
+  currentBranchName: string | null
+): Promise<ReadonlyArray<Branch>> {
+  const format = [
+    '%(refname)',
+    '%(refname:short)',
```
sergiou87, line 120: I added the name without `refs/heads` to the output for convenience, instead of having to juggle with it in the code afterwards.

app/src/lib/git/for-each-ref.ts
```diff
@@ -101,3 +101,104 @@ export async function getBranches(
...
+      continue
+    }
+
+    if (ref.startsWith('refs/head')) {
+      if (!upstream.length) {
+        // Exclude local branches without upstream
+        continue
+      }
+
+      localBranches.push({ name, ref, sha, upstream })
```
sergiou87, line 169: Is this ok? I figured that just creating an ad-hoc object was ok since TypeScript is nice enough to still infer types and all that 😍

app/src/lib/git/for-each-ref.ts
```diff
@@ -101,3 +101,104 @@ export async function getBranches(
...
+ * forwarded.
+ *
+ * @param repository Repository to get the branches from.
+ * @param allBranches All known branches in the repository.
+ * @param currentBranchName Name of the current branch.
+ */
+export async function getBranchesDifferingFromUpstream(
+  repository: Repository,
+  allBranches: ReadonlyArray<Branch>,
+  currentBranchName: string | null
```
sergiou87: Not 100% confident about passing these here vs doing the actual `Branch` filtering outside, tbh. What do you think?

app/src/lib/stores/helpers/find-branches-for-fast-forward.ts
```diff
@@ -1,46 +0,0 @@
...
-  // we don't have to worry about this being a duplicate, because recent branches
-  // never include the default branch (at least right now)
-  const shortListBranches =
-    defaultBranch !== null ? [...recentBranches, defaultBranch] : recentBranches
-
-  const eligibleShortListBranches = shortListBranches.filter(b =>
-    eligibleForFastForward(b, currentBranchName)
-  )
-  return eligibleShortListBranches
-}
```
sergiou87, line 46: I thought the implementation of this function after removing the threshold stuff was simple enough to write it inline in `app-store.ts`.

6068 (approved):

app/src/lib/git/for-each-ref.ts
niik: I think given the name of the method I wouldn't expect it do filter out the current branch. That seems like a responsibility which should live with the caller.

I suspect tho that as we start looking at part 2 of this we could use the `%(HEAD)` format argument to `for-each-ref` to get determine if a branch is the current HEAD or not and not involve our internal list of branches at all.

app/src/lib/git/for-each-ref.ts
niik: Love it

app/src/lib/git/for-each-ref.ts
```diff
@@ -101,3 +101,104 @@ export async function getBranches(
...
+  // - For local branches with upstream: name, ref, SHA and the upstream.
+  // - For remote branches we only need the sha (and the ref as key).
+  for (const line of lines) {
+    const pieces = line.split('\0')
+
+    const ref = pieces[0]
+    const name = pieces[1]
+    const sha = pieces[2]
+    const upstream = pieces[3]
+    const symref = pieces[4]
```
niik: We can clean this up a tad by deconstructing

```suggestion
    const [ref, name, sha, upstream, symref] = line.split('\0')
```

app/src/lib/git/for-each-ref.ts
```diff
@@ -101,3 +101,104 @@ export async function getBranches(
...
+
+    if (symref.length > 0 || name === currentBranchName) {
+      // Exclude symbolic refs and the current branch
+      continue
+    }
+
+    if (ref.startsWith('refs/head')) {
+      if (!upstream.length) {
```
niik: We try to avoid truthy/falsy comparisons

```suggestion
      if (upstream.length === 0) {
```

app/src/lib/git/for-each-ref.ts
niik: Super okay since we're not leaking it outside of the method

niik (changes requested):
This is looking great @sergiou87! Just a few notes

app/src/lib/git/for-each-ref.ts
sergiou87: That sounds good! ❤️

sergiou87:
@niik thanks for the feedback! I think I addressed all, so it's ready for a second review 😄

sergiou87:
@niik I just added a unit test for this new method, with a fixture that should become handy for the next round of ff improvements 😄

app/test/unit/git/for-each-ref-test.ts
```diff
@@ -58,4 +61,36 @@ describe('git/for-each-ref', () => {
...
+      const branchNames = branches.map(branch => branch.name)
+      expect(branchNames).toHaveLength(3)
+
+      // All branches that are behind and/or ahead must be included
+      expect(branchNames).toContain('branch-behind')
+      expect(branchNames).toContain('branch-ahead')
+      expect(branchNames).toContain('branch-ahead-and-behind')
+
+      // `master` is the current branch, and shouldn't be included
+      expect(branchNames).not.toContain('master')
```
niik: Is the local `master` branch behind its upstream? I.e. are we sure we're testing that it gets filtered out because it's the current branch and not just because it's up to date?

Also, I know this is a bit of a pain but do you think you could rename the branch in this fixture to `main` instead? We're trying to phase out the `master` references and have gotten rid of almost all of them save for tests.

niik (changes requested):
> @niik I just added a unit test for this new method, with a fixture that should become handy for the next round of ff improvements 😄

:heart: I was just about to ask for a test here given the potential impact if we get this wrong (especially if we accidentally included the current branch in the list).

Just one last question and one last ask and then I think this is good to go!

app/test/unit/git/for-each-ref-test.ts
sergiou87: > Is the local `master` branch behind its upstream?

Yes!

I updated the repo with the branch rename and also added some info to its `README.md` to clarify that the `main` branch is the current branch and is behind its upstream.

niik (approved):
Wonderful work here @sergiou87!

UncleStoney2273 (commented):
- Not to smart

PR #11400, opened 2021-01-15
Clean up editors code

### LINKED ISSUE #5409: Refactor editor handling
Currently, the editor handling is [rather duplicative](https://github.com/desktop/desktop/blob/master/app/src/lib/editors/darwin.ts). I‘ve realized that instead of having `ExternalEditor` be an enum, we can instead have an `editors` object with an object containing any necessary information about the editor. Since these are unique objects, they can still be equality-checked (i.e. `editor === editors.Atom`). Also, it saves on lots of duplication with `switch` statements, and reduces the required contribution to add a new editor to just five lines of code in one place for each platform.

<details><summary>A hopefully-working version of <code>editors/darwin.ts</code></summary>

```ts
function parse(label: string): ExternalEditor | null {
  for (const key of Object.keys(editors)) {
    const value = editors[key]
    if (value.name === label) {
      return value
    }
  }
  return null
}

async function findApplication(editor: ExternalEditor): Promise<string | null> {
  for (const identifier of editor.bundleIdentifiers) {
    try {
      const installPath = await appPath(identifier)
      const path = Path.join(installPath, ...editor.bundleIdentifiers)
      const exists = await pathExists(path)
      if (exists) {
        return path
      }

      log.debug(`Command line interface for ${editor} not found at '${path}'`)
    } catch (error) {
      log.debug(`Unable to locate ${editor} installation`, error)
    }
  }

  return null
}

async function getAvailableEditors() {
  const paths = await Promise.all(allEditors.map(findApplication))

  return paths.reduce((results, path, i) => {
    if (path) {
      results.push({ editor: editors[i], path })
    }
    return results
  }, new Array<IFoundEditor<ExternalEditor>>())
}

interface ExternalEditor {
  name: string
  bundleIdentifiers: ReadonlyArray<string>
  /** path segments after the found path of the bundle */
  path: ReadonlyArray<string>
}

const editors: { [editor: string]: ExternalEditor } = {
  Atom: {
    name: 'Atom',
    bundleIdentifiers: ['com.github.atom'],
    path: ['Contents', 'Resources', 'app', 'atom.sh'],
  },
  MacVim: {
    name: 'MacVim',
    bundleIdentifiers: ['org.vim.MacVim'],
    path: ['Contents', 'MacOS', 'MacVim'],
  },
  VisualStudioCode: {
    name: 'Visual Studio Code',
    bundleIdentifiers: ['com.microsoft.VSCode'],
    path: ['Contents', 'Resources', 'app', 'bin', 'code'],
  },
  VisualStudioCodeInsiders: {
    name: 'Visual Studio Code (Insiders)',
    bundleIdentifiers: ['com.microsoft.VSCodeInsiders'],
    path: ['Contents', 'Resources', 'app', 'bin', 'code'],
  },
  SublimeText: {
    name: 'Sublime Text',
    bundleIdentifiers: ['com.sublimetext.3'],
    path: ['Contents', 'SharedSupport', 'bin', 'subl'],
  },
  BBEdit: {
    name: 'BBEdit',
    bundleIdentifiers: ['com.barebones.bbedit'],
    path: ['Contents', 'Helpers', 'bbedit_tool'],
  },
  PhpStorm: {
    name: 'PhpStorm',
    bundleIdentifiers: ['com.jetbrains.PhpStorm'],
    path: ['Contents', 'MacOS', 'phpstorm'],
  },
  RubyMine: {
    name: 'RubyMine',
    bundleIdentifiers: ['com.jetbrains.RubyMine'],
    path: ['Contents', 'MacOS', 'rubymine'],
  },
  TextMate: {
    name: 'TextMate',
    bundleIdentifiers: ['com.macromates.TextMate'],
    path: ['Contents', 'Resources', 'mate'],
  },
  Brackets: {
    name: 'Brackets',
    bundleIdentifiers: ['io.brackets.appshell'],
    path: ['Contents', 'MacOS', 'Brackets'],
  },
  WebStorm: {
    name: 'WebStorm',
    bundleIdentifiers: ['com.jetbrains.WebStorm'],
    path: ['Contents', 'MacOS', 'WebStorm'],
  },
}
const allEditors = Object.keys(editors).map(k => editors[k])

// deps
function assertNever(x: never, message: string): never {
  throw new Error(message)
}

declare const log: any

declare const Path: {
  join(...args: string[]): string
}
declare const appPath: (bundleId: string) => Promise<string>
declare function pathExists(path: string): Promise<boolean>

interface IFoundEditor<T> {
  readonly editor: T
  readonly path: string
}
```

</details>

sergiou87:
Closes #5409

## Description

The changes in this PR are an attempt to simplify the code that integrates the app with external editors. Main changes:
- Gather all the info relative to external editors into entries of an array. This should make it easier and clearer adding support for new editors, since now all that info and logic won't be scattered.
- Removed the `ExternalEditor` enum and use a string instead. Several reasons for this:
  1. No code was leveraging the advantages of `ExternalEditor` as an enum, other than in `editors/darwin.ts`, `editors/win32.ts` and `editors/linux.ts`. Everywhere else its value was passed around, and finally stored/read as a string in/from the app settings.
  1. There weren't references to specific values of the enum outside of the mentioned files.
  1. With the new list of editors, keeping the enum around was just redundant. In the end, only editors inside the list AND detected in the user's machine will be available and used (🐛s aside).
- macOS editors are now opened using `open`, which only needs us to provide the bundle identifier of the editor, so we don't need to keep hardcoding paths to specific executables inside the editors' bundle.

Remaining work in this PR:
- ~~Make the same changes in the `editors/linux.ts` file.~~
- ~~Update the `docs/technical/editor-integration.md` document to reflect these new changes.~~
- ~~Other cleanup!~~

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -129,7 +129,7 @@ async function findApplication(
         return installPath
       }
 
-      log.debug(`App instalation for ${editor} not found at '${installPath}'`)
+      log.debug(`App installation for ${editor} not found at '${installPath}'`)
     } catch (error) {
       log.debug(`Unable to locate ${editor} installation`, error)
     }
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -1,7 +1,5 @@
-import * as Path from 'path'
 import { pathExists } from 'fs-extra'
 import { IFoundEditor } from './found-editor'
-import { assertNever } from '../fatal-error'
 import appPath from 'app-path'
 import { parseEnumValue } from '../enum'
 
@@ -30,145 +28,127 @@ export enum ExternalEditor {
   Nova = 'Nova',
 }
 
+interface IDarwinExternalEditor {
+  readonly name: ExternalEditor
+  readonly bundleIdentifiers: string[]
+}
+
+export const editors: IDarwinExternalEditor[] = [
+  {
+    name: ExternalEditor.Atom,
+    bundleIdentifiers: ['com.github.atom'],
+  },
+  {
+    name: ExternalEditor.MacVim,
+    bundleIdentifiers: ['org.vim.MacVim'],
+  },
+  {
+    name: ExternalEditor.VSCode,
+    bundleIdentifiers: ['com.microsoft.VSCode'],
+  },
+  {
+    name: ExternalEditor.VSCodeInsiders,
+    bundleIdentifiers: ['com.microsoft.VSCodeInsiders'],
+  },
+  {
+    name: ExternalEditor.VSCodium,
+    bundleIdentifiers: ['com.visualstudio.code.oss'],
+  },
+  {
+    name: ExternalEditor.SublimeText,
+    bundleIdentifiers: [
+      'com.sublimetext.4',
+      'com.sublimetext.3',
+      'com.sublimetext.2',
+    ],
+  },
+  {
+    name: ExternalEditor.BBEdit,
+    bundleIdentifiers: ['com.barebones.bbedit'],
+  },
+  {
+    name: ExternalEditor.PhpStorm,
+    bundleIdentifiers: ['com.jetbrains.PhpStorm'],
+  },
+  {
+    name: ExternalEditor.PyCharm,
+    bundleIdentifiers: ['com.jetbrains.PyCharm'],
+  },
+  {
+    name: ExternalEditor.RubyMine,
+    bundleIdentifiers: ['com.jetbrains.RubyMine'],
+  },
+  {
+    name: ExternalEditor.IntelliJ,
+    bundleIdentifiers: ['com.jetbrains.intellij'],
+  },
+  {
+    name: ExternalEditor.TextMate,
+    bundleIdentifiers: ['com.macromates.TextMate'],
+  },
+  {
+    name: ExternalEditor.Brackets,
+    bundleIdentifiers: ['io.brackets.appshell'],
+  },
+  {
+    name: ExternalEditor.WebStorm,
+    bundleIdentifiers: ['com.jetbrains.WebStorm'],
+  },
+  {
+    name: ExternalEditor.Typora,
+    bundleIdentifiers: ['abnerworks.Typora'],
+  },
+  {
+    name: ExternalEditor.CodeRunner,
+    bundleIdentifiers: ['com.krill.CodeRunner'],
+  },
+  {
+    name: ExternalEditor.SlickEdit,
+    bundleIdentifiers: [
+      'com.slickedit.SlickEditPro2018',
+      'com.slickedit.SlickEditPro2017',
+      'com.slickedit.SlickEditPro2016',
+      'com.slickedit.SlickEditPro2015',
+    ],
+  },
+  {
+    name: ExternalEditor.Xcode,
+    bundleIdentifiers: ['com.apple.dt.Xcode'],
+  },
+  {
+    name: ExternalEditor.GoLand,
+    bundleIdentifiers: ['com.jetbrains.goland'],
+  },
+  {
+    name: ExternalEditor.AndroidStudio,
+    bundleIdentifiers: ['com.google.android.studio'],
+  },
+  {
+    name: ExternalEditor.Rider,
+    bundleIdentifiers: ['com.jetbrains.rider'],
+  },
+  {
+    name: ExternalEditor.Nova,
+    bundleIdentifiers: ['com.panic.Nova'],
+  },
+]
+
 export function parse(label: string): ExternalEditor | null {
   return parseEnumValue(ExternalEditor, label) ?? null
 }
 
-function getBundleIdentifiers(editor: ExternalEditor): ReadonlyArray<string> {
-  switch (editor) {
-    case ExternalEditor.Atom:
-      return ['com.github.atom']
-    case ExternalEditor.MacVim:
-      return ['org.vim.MacVim']
-    case ExternalEditor.VSCode:
-      return ['com.microsoft.VSCode']
-    case ExternalEditor.VSCodeInsiders:
-      return ['com.microsoft.VSCodeInsiders']
-    case ExternalEditor.VSCodium:
-      return ['com.visualstudio.code.oss']
-    case ExternalEditor.SublimeText:
-      return ['com.sublimetext.4', 'com.sublimetext.3', 'com.sublimetext.2']
-    case ExternalEditor.BBEdit:
-      return ['com.barebones.bbedit']
-    case ExternalEditor.PhpStorm:
-      return ['com.jetbrains.PhpStorm']
-    case ExternalEditor.PyCharm:
-      return ['com.jetbrains.PyCharm']
-    case ExternalEditor.RubyMine:
-      return ['com.jetbrains.RubyMine']
-    case ExternalEditor.IntelliJ:
-      return ['com.jetbrains.intellij']
-    case ExternalEditor.TextMate:
-      return ['com.macromates.TextMate']
-    case ExternalEditor.Brackets:
-      return ['io.brackets.appshell']
-    case ExternalEditor.WebStorm:
-      return ['com.jetbrains.WebStorm']
-    case ExternalEditor.Typora:
-      return ['abnerworks.Typora']
-    case ExternalEditor.CodeRunner:
-      return ['com.krill.CodeRunner']
-    case ExternalEditor.SlickEdit:
-      return [
-        'com.slickedit.SlickEditPro2018',
-        'com.slickedit.SlickEditPro2017',
-        'com.slickedit.SlickEditPro2016',
-        'com.slickedit.SlickEditPro2015',
-      ]
-    case ExternalEditor.Xcode:
-      return ['com.apple.dt.Xcode']
-    case ExternalEditor.GoLand:
-      return ['com.jetbrains.goland']
-    case ExternalEditor.AndroidStudio:
-      return ['com.google.android.studio']
-    case ExternalEditor.Rider:
-      return ['com.jetbrains.rider']
-    case ExternalEditor.Nova:
-      return ['com.panic.Nova']
-    default:
-      return assertNever(editor, `Unknown external editor: ${editor}`)
-  }
-}
-
-function getExecutableShim(
-  editor: ExternalEditor,
-  installPath: string
-): string {
-  switch (editor) {
-    case ExternalEditor.Atom:
-      return Path.join(installPath, 'Contents', 'Resources', 'app', 'atom.sh')
-    case ExternalEditor.VSCode:
-    case ExternalEditor.VSCodeInsiders:
-      return Path.join(
-        installPath,
-        'Contents',
-        'Resources',
-        'app',
-        'bin',
-        'code'
-      )
-    case ExternalEditor.VSCodium:
-      return Path.join(
-        installPath,
-        'Contents',
-        'Resources',
-        'app',
-        'bin',
-        'code'
-      )
-    case ExternalEditor.MacVim:
-      return Path.join(installPath, 'Contents', 'MacOS', 'MacVim')
-    case ExternalEditor.SublimeText:
-      return Path.join(installPath, 'Contents', 'SharedSupport', 'bin', 'subl')
-    case ExternalEditor.BBEdit:
-      return Path.join(installPath, 'Contents', 'Helpers', 'bbedit_tool')
-    case ExternalEditor.PhpStorm:
-      return Path.join(installPath, 'Contents', 'MacOS', 'phpstorm')
-    case ExternalEditor.PyCharm:
-      return Path.join(installPath, 'Contents', 'MacOS', 'pycharm')
-    case ExternalEditor.RubyMine:
-      return Path.join(installPath, 'Contents', 'MacOS', 'rubymine')
-    case ExternalEditor.TextMate:
-      return Path.join(installPath, 'Contents', 'Resources', 'mate')
-    case ExternalEditor.Brackets:
-      return Path.join(installPath, 'Contents', 'MacOS', 'Brackets')
-    case ExternalEditor.WebStorm:
-      return Path.join(installPath, 'Contents', 'MacOS', 'WebStorm')
-    case ExternalEditor.IntelliJ:
-      return Path.join(installPath, 'Contents', 'MacOS', 'idea')
-    case ExternalEditor.Typora:
-      return Path.join(installPath, 'Contents', 'MacOS', 'Typora')
-    case ExternalEditor.CodeRunner:
-      return Path.join(installPath, 'Contents', 'MacOS', 'CodeRunner')
-    case ExternalEditor.SlickEdit:
-      return Path.join(installPath, 'Contents', 'MacOS', 'vs')
-    case ExternalEditor.Xcode:
-      return '/usr/bin/xed'
-    case ExternalEditor.GoLand:
-      return Path.join(installPath, 'Contents', 'MacOS', 'goland')
-    case ExternalEditor.AndroidStudio:
-      return Path.join(installPath, 'Contents', 'MacOS', 'studio')
-    case ExternalEditor.Rider:
-      return Path.join(installPath, 'Contents', 'MacOS', 'rider')
-    case ExternalEditor.Nova:
-      return Path.join(installPath, 'Contents', 'SharedSupport', 'nova')
-    default:
-      return assertNever(editor, `Unknown external editor: ${editor}`)
-  }
-}
-
-async function findApplication(editor: ExternalEditor): Promise<string | null> {
-  const identifiers = getBundleIdentifiers(editor)
-  for (const identifier of identifiers) {
+async function findApplication(
+  editor: IDarwinExternalEditor
+): Promise<string | null> {
+  for (const identifier of editor.bundleIdentifiers) {
     try {
       const installPath = await appPath(identifier)
-      const path = getExecutableShim(editor, installPath)
-      const exists = await pathExists(path)
+      const exists = await pathExists(installPath)
       if (exists) {
-        return path
+        return installPath
       }
 
-      log.debug(`Command line interface for ${editor} not found at '${path}'`)
+      log.debug(`App instalation for ${editor} not found at '${installPath}'`)
     } catch (error) {
       log.debug(`Unable to locate ${editor} installation`, error)
     }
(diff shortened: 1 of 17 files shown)


app/src/lib/editors/shared.ts
```diff
@@ -1,32 +1,11 @@
-import * as Darwin from './darwin'
-import * as Win32 from './win32'
-import * as Linux from './linux'
-
-export type ExternalEditor = Darwin.ExternalEditor | Win32.ExternalEditor
```
sergiou87, line 5: Does removing `ExternalEditor` feel like we're losing _something valuable_?

As I mentioned in the PR description, it doesn't feel like we lost anything from a "compiler help" perspective. But maybe the code was clearer with it… maybe it's worth making `type ExternalEditor = string`? 🤔

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,592 @@ import {
...
+  readonly appInformationExtractor?: AppInformationExtractor
+
+  /**
+   * Function to check if the found installation matches the expected identifier
+   * details.
+   *
+   * @param displayName The display name as listed in the registry
+   * @param publisher The publisher who created the installer
+   */
+  readonly expectedInstallationChecker: ExpectedInstallationChecker
```
sergiou87, line 56: I'm not particularly proud of these properties or their names or anything, but I'm open to suggestions 😂 

This felt like the best way to avoid scattering specific editor stuff all over this file 🤔

app/src/lib/editors/win32.ts
```diff
@@ -705,147 +637,28 @@ async function findApplication(editor: ExternalEditor): Promise<string | null> {
...
-  if (webstormPath) {
-    results.push({
-      editor: ExternalEditor.Webstorm,
-      path: webstormPath,
-    })
-  }
+  const editorPaths = await Promise.all(
+    editors.map(editor =>
+      findApplication(editor).then(path => {
+        return { editor, path }
```
sergiou87: Ad-hoc type + TypeScript compiler help = 😍 

But does it "feel" right for a TypeScript coder? 😳

DannyBhoy82 (approved):

DannyBhoy82 (dismissed):

app/src/lib/editors/darwin.ts
```diff
@@ -1,174 +1,135 @@
...
-      const path = getExecutableShim(editor, installPath)
-      const exists = await pathExists(path)
+      const exists = await pathExists(installPath)
       if (exists) {
-        return path
+        return installPath
       }
 
-      log.debug(`Command line interface for ${editor} not found at '${path}'`)
+      log.debug(`App instalation for ${editor} not found at '${installPath}'`)
```
niik:
```suggestion
      log.debug(`App installation for ${editor} not found at '${installPath}'`)
```

app/src/lib/editors/win32.ts
niik: > But does it "feel" right for a TypeScript coder? 😳

Absolutely!

app/src/lib/editors/win32.ts
```diff
@@ -705,147 +640,28 @@ async function findApplication(editor: ExternalEditor): Promise<string | null> {
...
-    })
-  }
-
-  if (webstormPath) {
-    results.push({
-      editor: ExternalEditor.Webstorm,
-      path: webstormPath,
-    })
-  }
+  const editorPaths = await Promise.all(
```
niik: I know that we used Promise.all in the previous implementation but I think we can make this method much more readable and compact by moving the call to `findApplication` inside the loop below. Since our registry-js calls are synchronous (unfortunately) it shouldn't alter the behavior in any noticeable way.

app/src/lib/editors/win32.ts
```diff
@@ -705,147 +640,28 @@ async function findApplication(editor: ExternalEditor): Promise<string | null> {
...
-  if (riderPath) {
-    results.push({
-      editor: ExternalEditor.Rider,
-      path: riderPath,
-    })
+    if (path) {
+      results.push({
+        editor: editor.name,
+        path: path,
+        usesShell: editor.usesShell ?? false,
```
niik: If we infer this from the path we can ditch  the `usesShell` from the `IFoundEditor` type.

```suggestion
        usesShell: path.endsWith('.cmd')
```

app/src/lib/editors/win32.ts
```diff
@@ -705,147 +640,28 @@ async function findApplication(editor: ExternalEditor): Promise<string | null> {
...
-
-  if (riderPath) {
-    results.push({
-      editor: ExternalEditor.Rider,
-      path: riderPath,
-    })
+    if (path) {
+      results.push({
+        editor: editor.name,
+        path: path,
```
niik: If the name of the local variable matches that of the field you're trying to assign to you can omit the right hand side.

```suggestion
        path,
```

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,595 @@ import {
...
+    name: 'Sublime Text',
+    registryKeys: [
+      {
+        key: HKEY.HKEY_LOCAL_MACHINE,
+        subKey:
+          'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\Sublime Text 3_is1',
+      },
+    ],
+    executableShimPath: ['subl.exe'],
+    appInformationExtractor: keys => {
```
niik: If we change the `expectedInstallationChecker` for this entry to check whether displayName starts with `Sublime Text` we should be able to remove this `appInformationExtractor` method

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,595 @@ import {
...
+      },
+      // Webstorm 2020.1
+      {
+        key: HKEY.HKEY_LOCAL_MACHINE,
+        subKey:
+          'SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\WebStorm 2020.1',
+      },
+    ],
+    executableShimPath: ['bin', 'webstorm.exe'],
+    appInformationExtractor: keys => {
```
niik: Since the `expectedInstallationChecker` for this entry already checks whether displayName starts with `WebStorm` we should be able to remove this `appInformationExtractor` method, right?

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,595 @@ import {
...
+      },
+      // PhpStorm 2020.1
+      {
+        key: HKEY.HKEY_LOCAL_MACHINE,
+        subKey:
+          'SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\PhpStorm 2020.1',
+      },
+    ],
+    executableShimPath: ['bin', 'phpstorm.exe'],
+    appInformationExtractor: keys => {
```
niik: Similar to the `WebStorm` entry, this method looks to be redundant?

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,595 @@ import {
...
+      },
+    ],
+    executableShimPath: [],
+    appInformationExtractor: keys => {
+      const displayName = getKeyOrEmpty(keys, 'DisplayName')
+      const publisher = getKeyOrEmpty(keys, 'Publisher')
+      const installLocation = getKeyOrEmpty(keys, 'DisplayIcon')
+
+      return { displayName, publisher, installLocation }
+    },
```
niik: If we leverage the default extractor here we can make it clear that we're only overriding the `installLocation`
```suggestion
    appInformationExtractor: keys => ({
      ...defaultAppInformationExtractor(keys),
      installLocation: getKeyOrEmpty(keys, 'DisplayIcon'),
    }),
```

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,595 @@ import {
...
+    registryKeys: [
+      // Rider 2019.3.4
+      {
+        key: HKEY.HKEY_LOCAL_MACHINE,
+        subKey:
+          'SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\JetBrains Rider 2019.3.4',
+      },
+    ],
+    executableShimPath: ['bin', 'rider64.exe'],
+    appInformationExtractor: keys => {
```
niik: Similar to `WebStorm` and `PhpStorm` we should be able to get rid of this method

app/src/lib/editors/win32.ts
```diff
@@ -10,663 +10,595 @@ import {
...
-async function findApplication(editor: ExternalEditor): Promise<string | null> {
-  const registryKeys = getRegistryKeys(editor)
+const defaultAppInformationExtractor: AppInformationExtractor = keys => {
+  const displayName = getKeyOrEmpty(keys, 'DisplayName')
+  const publisher = getKeyOrEmpty(keys, 'Publisher')
+  const installLocation = getKeyOrEmpty(keys, 'InstallLocation')
+  return { displayName, publisher, installLocation }
+}
 
+async function findApplication(
```
niik: I know this isn't part of your refactor but upon reading this method again it sure feels weird that we're just grabbing the first set of keys and then not continuing down the fallback chain if the `expectedInstallationChecker`  fails.

Seems like this method should be rewritten such that we test the validity of each set of keys. Something like this:

```ts
async function findApplication(editor: IWindowsExternalEditor) {
  const getAppInfo =
    editor.appInformationExtractor ?? defaultAppInformationExtractor

  for (const { key, subKey } of editor.registryKeys) {
    const keys = enumerateValues(key, subKey)
    if (keys.length === 0) {
      continue
    }

    const { displayName, publisher, installLocation } = getAppInfo(keys)

    if (!editor.expectedInstallationChecker(displayName, publisher)) {
      log.debug(`Unexpectted registry entries for ${editor.name}`)
      continue
    }

    const path = Path.join(installLocation, ...editor.executableShimPath)
    const exists = await pathExists(path)
    if (!exists) {
      log.debug(`Executable for ${editor.name} not found at '${path}'`)
      continue
    }

    return path
  }

  return null
}
```

app/src/lib/editors/win32.ts
niik: I'm fine with these names but I would probably have called the `appInformationExtractor` method something like `getAppInfo` or `getUninstallInfo`. Your call tho

app/src/lib/editors/darwin.ts
```diff
@@ -182,150 +143,27 @@ async function findApplication(editor: ExternalEditor): Promise<string | null> {
...
-
-  if (riderPath) {
-    results.push({ editor: ExternalEditor.Rider, path: riderPath })
-  }
-
-  if (novaPath) {
-    results.push({ editor: ExternalEditor.Nova, path: novaPath })
+  const results: Array<IFoundEditor<string>> = []
+
+  const editorPaths = await Promise.all(
```
niik: Similar comment here as with the win32 `getAvailableEditors`. I think we should move the call to `findApplication` into the loop below.

app/src/lib/editors/darwin.ts
```diff
@@ -182,150 +143,27 @@ async function findApplication(editor: ExternalEditor): Promise<string | null> {
...
+  )
+
+  for (const editorPath of editorPaths) {
+    const { editor, path } = editorPath
+
+    if (path) {
+      results.push({
+        editor: editor.name,
+        path: path,
+      })
```
niik: :art: 
```suggestion
      results.push({ editor: editor.name, path })
```

niik (changes requested):
This is such a fantastic improvement @sergiou87, thanks for working on this!

We'll also need to review and possibly update [our documentation](https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md) for adding a new external editor.

I also think there's potential to reduce duplication in the win32 file by introducing a few helper functions to construct these registry keys. Something like:

```ts
const registryKey = (key: HKEY, ...subKeys: string[]) =>
  ({ key, subKey: Path.win32.join(...subKeys) })

const uninstallSubKey =
  'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall'

const wow64UninstallSubKey =
  'SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall'

const CurrentUserUninstallKey = (subKey: string) =>
  registryKey(HKEY.HKEY_CURRENT_USER, uninstallSubKey, subKey)

const LocalMachineUninstallKey = (subKey: string) =>
  registryKey(HKEY.HKEY_LOCAL_MACHINE, uninstallSubKey, subKey)

const Wow64LocalMachineUninstallKey = (subKey: string) =>
  registryKey(HKEY.HKEY_LOCAL_MACHINE, wow64UninstallSubKey, subKey)
```

What do you think about that?

It'd cut down the VS code entry to just

```ts
    name: 'Visual Studio Code',
    registryKeys: [
      // 64-bit version of VSCode (user) - provided by default in 64-bit Windows
      CurrentUserUninstallKey('{771FD6B0-FA20-440A-A002-3B3BAC16DC50}_is1'),
      // 32-bit version of VSCode (user)
      CurrentUserUninstallKey('{D628A17A-9713-46BF-8D57-E671B46A741E}_is1'),
      // 64-bit version of VSCode (system) - was default before user scope installation
      LocalMachineUninstallKey('{EA457B21-F73E-494C-ACAB-524FDE069978}_is1'),
      // 32-bit version of VSCode (system)
      Wow64LocalMachineUninstallKey('{F8A2A208-72B3-4D61-95FC-8A65D340689B}_is1'),
    ],
```

app/src/lib/editors/win32.ts
sergiou87: I thought about this, but I wasn't 100% sure this would always be the case for future editors we might support. Anyway, happy to make the change!

app/src/lib/editors/win32.ts
sergiou87: Nice catch!

sergiou87:
Thank you for the thorough review and awesome feedback!! 🙇‍♂️ 🙇‍♂️ ❤️ 

> We'll also need to review and possibly update [our documentation](https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md) for adding a new external editor.

Yup! That (and `linux.ts`) is in my todo list. I wanted to confirm this approach with you before updating the documentation 😄 

> I also think there's potential to reduce duplication in the win32 file by introducing a few helper functions to construct these registry keys. Something like:
> …

I absolutely love it 😍 

After your suggestions to remove many of the `appInformationExtractor` definitions, and given I still don't like its name 😂 , I realized we don't really need it anymore. Only Notepad++ has a different logic, and it's just it looks for the install location in a different registry key. So I just replaced the `appInformationExtractor` function with an optional `installLocationRegistryKey` that defaults to `'InstallLocation'` and is overridden for Notepad++ only.

I'll make the changes in `linux.ts`, in the docs, test everything again everywhere I can, and let you know when it can be re-reviewed 🤓

sergiou87:
@niik I think I addressed all your feedback and also did some additional cleanup. Please, take a look at my comment above for more details 😌 

I also tested the few editors I have on Windows, Linux and macOS and everything worked as expected. On Linux I had to change locally a few executable paths though, since they didn't match what I had in my Ubuntu 🤷‍♂️

app/src/lib/editors/darwin.ts
```diff
@@ -1,174 +1,135 @@
...
 }
 
-async function findApplication(editor: ExternalEditor): Promise<string | null> {
-  const identifiers = getBundleIdentifiers(editor)
-  for (const identifier of identifiers) {
+/**
+ * This list contains all the external editors supported on macOS. Add a new
+ * entry here to add support for your favorite editor.
+ **/
+const editors: IDarwinExternalEditor[] = [
```
niik, line 21: Given how simple this type is you could simplify it to be a Record type keyed on the name if you wanted.  Something like

```ts
const editors: Record<string, ReadonlyArray<string>> = {
  Atom: ['com.github.atom'],
  MacVim: ['org.vim.MacVim'],
  'Visual Studio Code': ['com.microsoft.VSCode'],
  'Visual Studio Code (Insiders)': ['com.microsoft.VSCodeInsiders'],
  VSCodium: ['com.visualstudio.code.oss'],
  'Sublime Text': [
    'com.sublimetext.4',
    'com.sublimetext.3',
    'com.sublimetext.2',
  ],
  BBEdit: ['com.barebones.bbedit'],
  PhpStorm: ['com.jetbrains.PhpStorm'],
  PyCharm: ['com.jetbrains.PyCharm'],
  RubyMine: ['com.jetbrains.RubyMine'],
  TextMate: ['com.macromates.TextMate'],
  Brackets: ['io.brackets.appshell'],
  WebStorm: ['com.jetbrains.WebStorm'],
  Typora: ['abnerworks.Typora'],
  CodeRunner: ['com.krill.CodeRunner'],
  SlickEdit: [
    'com.slickedit.SlickEditPro2018',
    'com.slickedit.SlickEditPro2017',
    'com.slickedit.SlickEditPro2016',
    'com.slickedit.SlickEditPro2015',
  ],
  IntelliJ: ['com.jetbrains.intellij'],
  Xcode: ['com.apple.dt.Xcode'],
  GoLand: ['com.jetbrains.goland'],
  'Android Studio': ['com.google.android.studio'],
  Rider: ['com.jetbrains.rider'],
  Nova: ['com.panic.Nova'],
}
```

Can't get much more compact than that eh? Totally up to you. Then you can use `Object.entries` to iterate over them

```ts
  for (const [editor, bundleIdentifiers] of Object.entries(editors)) {
```

app/src/lib/editors/win32.ts
```diff
@@ -10,458 +10,279 @@ import {
...
+      LocalMachineUninstallKey('{7CC0E567-ACD6-41E8-95DA-154CEEDB0A18}'),
+    ],
+    executableShimPath: ['win', 'vs.exe'],
+    expectedInstallationChecker: (displayName, publisher) =>
+      displayName.startsWith('SlickEdit') && publisher === 'SlickEdit Inc.',
+  },
+  {
+    name: 'JetBrains Webstorm',
+    registryKeys: [
+      // Webstorm 2018.3
```
niik: Given that the uninstall key match the comment exactly I don't feel like these comments add much value

```suggestion
```

app/src/lib/editors/win32.ts
```diff
@@ -10,458 +10,279 @@ import {
...
+      Wow64LocalMachineUninstallKey('WebStorm 2020.1'),
+    ],
+    executableShimPath: ['bin', 'webstorm.exe'],
+    expectedInstallationChecker: (displayName, publisher) =>
+      displayName.startsWith('WebStorm') && publisher === 'JetBrains s.r.o.',
+  },
+  {
+    name: 'JetBrains Phpstorm',
+    registryKeys: [
+      // PhpStorm 2019.2
```
niik: Similar to Webstorm. Let's just drop these comments here

app/src/lib/editors/win32.ts
```diff
@@ -10,458 +10,279 @@ import {
...
+    ],
+    executableShimPath: [],
+    installLocationRegistryKey: 'DisplayIcon',
+    expectedInstallationChecker: (displayName, publisher) =>
+      displayName.startsWith('Notepad++') && publisher === 'Notepad++ Team',
+  },
+  {
+    name: 'JetBrains Rider',
+    registryKeys: [
+      // Rider 2019.3.4
```
niik: It's already in the subKey
```suggestion
```

app/src/lib/editors/win32.ts
```diff
@@ -471,381 +292,65 @@ function getKeyOrEmpty(
...
-  if (editor === ExternalEditor.Phpstorm) {
-    let displayName = ''
-    let publisher = ''
-    let installLocation = ''
+): IWindowsAppInformation {
+  const displayName = getKeyOrEmpty(keys, 'DisplayName')
+  const publisher = getKeyOrEmpty(keys, 'Publisher')
+  const installLocation = getKeyOrEmpty(
+    keys,
+    editor.installLocationRegistryKey ?? 'InstallLocation'
```
niik, line 291: 😍

niik (changes requested):
image

😻 this is such a wonderful cleanup @sergiou87! I had just a few tiny suggestions to give but I'm happy to ship this as is if you don't agree with them!

app/src/lib/editors/darwin.ts
sergiou87: That's a good one! 🤩 

However I'm tempted to leave the current approach I have 😳 , since I think it makes the code (and the documentation in `editor-integration.md`) more consistent with Linux and Windows. What do you think?

niik: Fine by me!

niik (approved):
Feels so good to have all of this cleaned up. It's been collecting dust for much to long now!

PR #11424, opened 2021-01-20
On hover show full message

### LINKED ISSUE #11375: On hover show full message

Complete message of current operation is not visible.

### Proposed solution

On hover show complete message.

### Additional context

A gif of current problem : 

Screencast from মঙ্গলবার 12 জানু 2021 12 35 48 +06

maifeeulasad:
Closes #11375 

## Description
 - Added title to div, so that on hover it shows full message. Basic tooltip.

### Screenshots
On hover show full message

## Release notes

Notes:


say25:
I sorta think this should just wrap to multiple lines

billygriffin:
I'm fine with it wrapping. Would you mind updating the PR and adding another gif so we can see that before reviewing and merging?

maifeeulasad:
I've used `lint:fix` this time, so all checks have been passed.

And here is another GIF showing the result : 
https___github com_maifeeulasad_desktop_tree_enhance_clone_hover_full_message

say25:
Maybe I wasn't super clear, I think the `Receiving objects: ...`  message should be wrap to multi-lines

maifeeulasad:
```
  /* Limit the progress text to one line. Without
   * this the entire container will jump around in
   * y-pos due to the change in height */
   .details {
    max-width: 100%;
    @include ellipsis;
  }
```
[source](https://github.com/maifeeulasad/desktop/blob/42a5be758c2ff19ae6ea44721db3556b8c16ae7c/app/styles/ui/_cloning-repository-view.scss#L69_L71)

Should I try and change it ?

say25:
Ahh that makes sense then, nevermind 😄

say25 (dismissed):

sergiou87:
@maifeeulasad I think it's worth changing that, or trying at least 😂 

Maybe we can move the progress bar to the center of the screen and keep it fixed. Then that `details` text goes below the progress bar, and takes as many lines as needed (up to some reasonable number of lines or height?).

Does that make sense?

maifeeulasad:
I've tried line-clamping, but it didn't work for me.
My tries :
```
    display: -webkit-box;
    -webkit-box-orient: vertical;
    -webkit-line-clamp: 3;
```
And I've also tried 
```
    text-align: center;
    width: calc(100%);
    max-width: calc(100%);
```

tidy-dev:
@maifeeulasad I think something like this should do the trick. It gives the details all the room it would want to wrap while using height to center the message. Mind giving it a try and testing it out?
image

maifeeulasad:
@tidy-dev really appreciate your help, and I've tried it. Here is the result : 
on overflow show full message; when tries to set line number to display, it gets overfitted


I surely can use **js** to calculate height of that line(as font size may vary ?) and set the height. Will that work ?

*Asking, although I haven't seen this type of code in `desktop` repository*

tidy-dev:
@maifeeulasad I would not be inclined to solve this with js. We prefer to manage view problems with css (if possible). It appears from your last screen capture that we are just missing the css to force a line-break on long non-breaking sections of text when needed. 

Can you add: 
overflow-wrap: break-word;

to the .details class?

image

maifeeulasad:
@tidy-dev this doesn't solve the line count problem. 
And I'm quite sure, *I need* to use `js` to do so. Maybe there is some way around with `css`, but I'm not that good in `css`.

youholemy:
Preschool

youholemy:
Dims

sergiou87:
> this doesn't solve the line count problem

@maifeeulasad I'm sorry, what's that problem exactly? It seems to me that @tidy-dev's last suggestion should solve the problem of long paths. Isn't that right? 🤔

maifeeulasad:
> (up to some reasonable number of lines or height?).

as mentioned https://github.com/desktop/desktop/pull/11424#issuecomment-804208358, or should I just implement it, from @tidy-dev 's solution

tidy-dev:
@maifeeulasad If the concern is allowing it to go all the way down to the bottom of the screen, we can force some margin. This doesn't limit to a specific line count, but I think the goal should just be to allow it to wrap as many times as it needs without it starting to look weird by going out the bottom of the screen.. and potentially pushing the text up. 

For that... 
```
.details {
    max-width: 100%;
    height: 30vh;
    margin-bottom: 20vh;
    overflow-wrap: anywhere;
    overflow: hidden;
    text-overflow: ellipsis;
}
```

Below I use zoomed in as much as allowed and shorted the screen as much as allowed while provide an extreme amount of text. This still looks fine to me :) 
image


Here is what that looks back on 100% zoom and 1100px height.
image

The one pitfall here is that the text-overflow:ellipsis doesn't apply most of the time in this scenario... But, my assumption is that most use cases will have 2 and at most 3 lines here and will never hit the max-height here.

app/styles/ui/_cloning-repository-view.scss
```diff
@@ -66,12 +66,13 @@
     overflow: hidden;
```
memodrnn, line 66: …

maifeeulasad: hello @memodrnn , i couldn't get you. a little explanation would be helpful

app/src/ui/cloning-repository.tsx
```diff
@@ -27,7 +27,9 @@ export class CloningRepositoryView extends React.Component<
           <div className="title">Cloning {this.props.repository.name}</div>
         </div>
         <progress value={progressValue} />
-        <div className="details">{this.props.progress.description}</div>
+        <div title={this.props.progress.description} className="details">
```
vidalvenegas2, line 30:
```suggestion
        <div title={this.props.progress.description} className="details">
```

vidalvenegas2 (approved):

app/styles/ui/_cloning-repository-view.scss
sergiou87: Don't worry about that comment @maifeeulasad, it was just a spammy message 😓

tidy-dev:
@maifeeulasad Thank you for all the time you have put into this! 🙇  Do you think you could also update the branch to the latest that is in development? After that we will can look at getting this merged. 🙂

maifeeulasad:
@tidy-dev updated branch from upstream/development.

tidy-dev:
@maifeeulasad Hmm.. seems something went a little wrong with the merge -> It is showing 291 file changes. Think you can try creating a new branch off latest development and open a new pr with just your changes. :)

tidy-dev:
Closing this in favor of #12035

PR #11434, opened 2021-01-21
Add branch context menu

### LINKED ISSUE #5803: Add context menu to branch list

### LINKED ISSUE #10432: Delete Branch from Branches View

I would like to be able to delete branches from the branches selection menu, along with PRs.

### Proposed solution

Add a little trash can icon next to the branch you want to delete. When you click it, it will come up with a notice saying what it will do. If the user clicks Confirm, the branch will be deleted. If the user clicks cancel, the branch will not be deleted. The `Master` Branch cannot be deleted from it.

### Additional context

Mock ups

GitHubDesktop_ObeyntEbIh
GitHubDesktop_ccFZESo31v

tidy-dev:
Closes #5803
Closes #10432

## Description

Allow user to delete and rename a branch from the branch dropdown list.

#### Delete
Local Only - does NOT have check box for deleting remote
Local and Remote - does have check box for deleting remote
Delete Context Menu

Delete Remote Only 
image

#### Rename
Remote Only - disabled
Local Only - does NOT have remote won't changed warning.
Local and Remote - Includes remote won't change warning.
Rename Via Context Menu

## Release notes

[Added] Add branch context menu with rename and delete to branch dropdown list  - #5803 

Notes:


niik:
> Add branch context menu - IN PROGRESS

@tidy-dev in the future feel free to just make the PR a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) until you're ready to have it be looked at. It makes it a bit easier to see which PRs in the list require review.

tidy-dev:
@niik :D Thank you - no idea that existed!

app/src/ui/branches/branch-list-item.tsx
```diff
@@ -18,10 +20,61 @@ interface IBranchListItemProps {
...
+      this.props.onRenameBranch === undefined &&
+      this.props.onDeleteBranch === undefined
+    ) {
+      return
+    }
+
+    // TODO: find out - does branch menu have "Rename..." because ... stand
+    // for branch name? If so is it needed here? (same for delete)
+    // Also, branch menu has & in front for not mac, what is that?
+    // Should it be disabled for current branch? - note from prev PR
```
tidy-dev: @niik This todo was questions I add after reviewing the code for the top menu in the build-default-ment.ts
image

I went without the ... presuming it was indicating the branch name which is clear when right clicking  on it.. but is it some convention I should following in the context menu?

Also what is the & for?

niik: The `...` suffix is a UX convention which means that the action in question isn't just gonna happen when you click on the item, i.e. there'll be some form of confirmation or other action needed before said action actually happens.

Here's a good example where we conditionally add the ellipsis depending on whether the user has chosen to see a confirmation prompt when stashing changes or not

https://github.com/desktop/desktop/blob/42a5be758c2ff19ae6ea44721db3556b8c16ae7c/app/src/ui/changes/changes-list.tsx#L354-L359

The ampersand is a Windows specific thing which indicates that the following character is the [access key](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/bb226831(v=vs.85)) meaning that a user can press <kbd>Alt</kbd> and that key instead of clicking on the item. We typically only use access keys in the top level menu so you can go  ahead and remove it from the context menu.

app/src/lib/git/branch.ts
```diff
@@ -78,7 +78,13 @@ export async function deleteBranch(
...
   }
 
-  const remoteName = branch.remote
+  let remoteName = branch.remote
+
+  if (branch.type === BranchType.Remote) {
+    remoteName = branch.name
+      .replace(branch.nameWithoutRemote, '')
+      .replace('/', '')
+  }
```
sergiou87: Feels a bit odd having this logic here in addition to the `let remoteName = branch.remote` 🤔  I think I'd do one of these:
1. Adding a comment explaining why we need this (i.e. explain `branch.remote` doesn't give us the remote of remote branches).
2. Renaming the property `remote` of `Branch` to something more explicit, like `upstreamRemote`, and then encapsulate this logic in a new property, maybe something like `remotePrefix` (to be consistent with https://github.com/desktop/desktop/blob/development/app/src/lib/remove-remote-prefix.ts/#L9) and maybe a new `remote` property in `Branch` that chooses between `remotePrefix` and `upstreamRemote` depending on their availability and/or type of branch?

I think I'd go with (2). At least the part of renaming `remote` to `upstreamRemote`. I think just calling it `remote` is misleading because you would expect to get _something_ from remote branches too 😢

app/src/lib/git/branch.ts
```diff
@@ -78,7 +78,13 @@ export async function deleteBranch(
     await deleteLocalBranch(repository, branch.name)
```
sergiou87: I think the comment/documentation for `deleteBranch` should be updated to reflect the new behavior, right? (Maybe after considering the rest of the feedback in this PR)

(Sorry, I couldn't post this comment in the lines of the function documentation, this is as far as I could go 😆 )

app/src/lib/git/branch.ts
```diff
@@ -78,7 +78,13 @@ export async function deleteBranch(
...
-  const remoteName = branch.remote
+  let remoteName = branch.remote
+
+  if (branch.type === BranchType.Remote) {
+    remoteName = branch.name
+      .replace(branch.nameWithoutRemote, '')
+      .replace('/', '')
+  }
 
   if (includeRemote && remoteName) {
```
sergiou87: The use of `includeRemote` when you're deleting specifically a remote branch feels odd too (from an API consumer perspective). Like, if I want to delete a remote branch I must call `deleteBranch`, pass the remote branch as a parameter **and** set `includeRemote` to `true`. Shouldn't `includeRemote` be irrelevant once I pass a remote branch? 🤔 

Options I see here:
1. Make `includeRemote` optional, ignore it when the branch is remote, and assume it's `false` when omitted for local branches.
2. Extract the logic to remove remote branches to a `deleteRemoteBranch`, then `deleteBranch` will delete only the local branch or both local and remote when needed, and `DeleteRemoteBranch` would use `deleteRemoteBranch` directly (through the dispatcher, of course, and renaming the previous). More in detail, this is what I'd do for this option:
   1. Inline the implementation of `deleteLocalBranch` in `deleteBranch`.
   2. **BONUS:** remove the `if (branch.type === BranchType.Local)` check since, unless I'm missing something, it wasn't needed before this PR (it was only used with local branches) and it wouldn't be needed after making the changes in this option (2)?
   3. Remove `deleteLocalBranch`
   4. Rename `deleteBranch` to `deleteLocalBranch`. Do the same with `deleteBranch` in the dispatcher and `_deleteBranch` in `AppStore`.
   5. Extract (and reuse) the logic to delete a remote branch to a new `deleteRemoteBranch` function.
   6. Add new `deleteRemoteBranch` in the dispatcher and `_deleteRemoteBranch` in `AppStore` to use this new function from the previous step.

IMO (2) feels much cleaner than (1), but I understand it might be a bit overkill too given the changes we need in the dispatcher and the app store.

What do you think?

app/src/lib/stores/app-store.ts
```diff
@@ -3351,34 +3352,19 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-            branchToCheckout = branchesState.recentBranches[i]
-            break
-          }
-          i++
-        }
-      }
+      const branchToCheckout = this.getBranchToCheckoutAfterDelete(
+        branch,
+        branchesState
+      )
```
sergiou87: Nice cleanup with this `getBranchToCheckoutAfterDelete`! 😍

app/src/ui/branches/branch-list-item.tsx
```diff
@@ -18,10 +20,51 @@ interface IBranchListItemProps {
...
+    const { onRenameBranch, onDeleteBranch, name, isLocal } = this.props
+    if (onRenameBranch === undefined && onDeleteBranch === undefined) {
+      return
+    }
+
+    const items: Array<IMenuItem> = []
+
+    if (onRenameBranch !== undefined) {
+      items.push({
+        label: 'Rename...',
```
sergiou87: Nit: prefer the use of `…` over `...`?
```suggestion
        label: 'Rename…',
```

app/src/ui/branches/branch-list-item.tsx
```diff
@@ -18,10 +20,51 @@ interface IBranchListItemProps {
...
+        action: () => onRenameBranch(name),
+        enabled: isLocal,
+      })
+    }
+
+    if (onDeleteBranch !== undefined) {
+      items.push({
+        label: 'Delete...',
```
sergiou87: Same nit 😳 
```suggestion
        label: 'Delete…',
```

app/src/ui/branches/branches-container.tsx
```diff
@@ -261,4 +267,46 @@ export class BranchesContainer extends React.Component<
   ) => {
     this.setState({ selectedPullRequest })
   }
+
+  private onRenameBranch = (branchName: string) => {
+    const branch: Branch | undefined = this.props.allBranches.find(
+      branch => branch.name === branchName
+    )
```
sergiou87: Nit: maybe extract this to a `getBranch` or `getBranchWithName` function and reuse in `onDeleteBranch`? 🤔

sergiou87 (commented):
Looks great overall, awesome job!! I only have some nits and then a few concerns/questions in `app/src/lib/git/branch.ts`. I'd love to ~~hear~~ read what you think!

app/src/lib/stores/app-store.ts
tidy-dev: Thank you!

app/src/ui/branches/branches-container.tsx
```diff
@@ -261,4 +267,46 @@ export class BranchesContainer extends React.Component<
   ) => {
     this.setState({ selectedPullRequest })
   }
+
+  private getBranchWithName(branchName: string): Branch | undefined {
+    return this.props.allBranches.find(branch => branch.name === branchName)
+  }
+
+  private onRenameBranch = (branchName: string) => {
+    const branch: Branch | undefined = this.getBranchWithName(branchName)
```
sergiou87: I think you can drop the variable type here now, right? 😌 
```suggestion
    const branch = this.getBranchWithName(branchName)
```

app/src/ui/branches/branches-container.tsx
```diff
@@ -261,4 +267,46 @@ export class BranchesContainer extends React.Component<
...
+
+    this.props.dispatcher.showPopup({
+      type: PopupType.RenameBranch,
+      repository: this.props.repository,
+      branch: branch,
+    })
+  }
+
+  private onDeleteBranch = (branchName: string) => {
+    const branch: Branch | undefined = this.getBranchWithName(branchName)
```
sergiou87: Same here
```suggestion
    const branch = this.getBranchWithName(branchName)
```

app/src/lib/git/branch.ts
tidy-dev: I definitely agree; however I am hesitant to change what remote is as it could be fairly impactful. I did add a upstreamRemote and add documentation to remote explaining what happens when you use it for a remote branch.

app/src/lib/git/branch.ts
tidy-dev: I like option 2, but kind of like to keep `deleteBranch` existing to be a controller as to determine which of `deleteLocalBranch` and `deleteRemoteBranch` to use.

As for point ii. -> I think there was some defensive programming happening here to prevent the unintentional deletion of remote branches. Because if you had passed in a remote branch to this, it would simply return true and do nothing.

app/src/lib/git/branch.ts
sergiou87: Yeah, I understand how it's a big change 😕 

The reason why I suggested to rename the current `remote` (in the class and wherever is used) to `upstreamRemote` is because it describes its behavior more accurately: it doesn't always return the remote of the branch, it only does when the branch is local and has an upstream.

The new `upstreamRemote` you just added is returning the upstream remote for local branches, which is perfect 👍, but also the branch remote for remote branches (what a 👅 🌪️  🤣), which technically don't have an upstream.

That's why I think their names it would be semantically more accurate if the names of those properties were swapped. What do you think?

app/src/lib/git/branch.ts
sergiou87: > but kind of like to keep `deleteBranch` existing to be a controller as to determine which of `deleteLocalBranch` and `deleteRemoteBranch` to use.

Yeah I considered that too, but thinking about it, it didn't feel like that function had any use with remote branches: its main purpose is to delete local branches, and it would delete remote branches if and only if `includeRemote` is `true`. Technically, it should _always_ delete the local branch (except in completely unexpected situations, hence the defensive programming).

> I think there was some defensive programming happening here

Interesting… I wonder if that check should be switched to a "fast return" and send a non-fatal error to see if it actually happens out in the wild (/cc @niik), and if it doesn't happen just remove the check. In the end `deleteLocalBranch` would only delete the remote branch locally, but that wouldn't impact the remote repo 🤔

tidy-dev: I looked into doing this.. and the further I got the more I didn't want to do it. :P 

1) If we step back to app-store and look at its `deleteBranch`, It is heavy in promises.. thus splitting it into 2 methods of remote versus local would either be really duplicated functions or passing in a function to call which to me is starting defeat the purpose of being cleaner cause passing functions inside functions that already pass functions hard to read. The `deleteBranch` in branches.ts is very simple and easy to see that it is just a decision maker.

2) `branch-pruner` uses the current deleteLocalBranch.. thus potential to impact it.

sergiou87: > 1) If we step back to app-store and look at its `deleteBranch`…

I don't think duplication there would be _that_ bad (in remote branches you don't need to "checkout after delete", because a remote branch can't be the current branch, so the new `_deleteRemoteBranch` method only needs to delete the branch), specially knowing deleting local vs remote branches is a bit different, but I agree all that makes the decision a bit less easy. I'll leave it up to you!

> 2) `branch-pruner` uses the current deleteLocalBranch.. thus potential to impact it.

Nice catch! Seems like inlining that function is not a (good) option then :doh:

However, that wouldn't stop _me_ from renaming `deleteBranch` to `deleteLocalBranch`, and renaming that `deleteLocalBranch` to something more specific like `deleteLocalBranchWithName`?

> The `deleteBranch` in branches.ts is very simple and easy to see that it is just a decision maker.

For me the problem is that you _need_ to look into the implementation/documentation because the function signature is misleading/confusing now 😕 ("what's `includeRemote` if my branch is remote already?"), and I don't think it's hard to make the intent of the function clearer.

Anyway, I personally think the change I'm suggesting would lead to a cleaner result, but I agree it might not be a huge improvement overall vs the amount of changes, so I don't want to block this PR and I'll leave it up to you!

(I'd also love to know what's @niik 's take on this 😆 Sorry!)

app/src/lib/git/branch.ts
```diff
@@ -64,9 +64,64 @@ export async function deleteLocalBranch(
...
+      // proxy though so it's not critical.
+      log.error(`Could not resolve remote url for remote ${remoteName}`, err)
+      return null
+    })) || getFallbackUrlForProxyResolve(account, repository)
+
+  const args = [
+    ...networkArguments,
+    'push',
+    remoteName,
+    `:${branch.nameWithoutRemote}`,
```
niik: I think this is incorrect. And it's likely that it's been incorrect since the dawn of time so nothing to do with your changes but still, seems incorrect.

Since we could be dealing with a local branch here we could (in the best case scenario) fail to delete the remote branch and in the worst case scenario delete the wrong remote branch (since a local branch name doesn't have to match the remote branch name).

If we create a local branch named foo which tracks a remote branch called bar this method will attempt to delete `foo` on the remote
```
git checkout -b foo
git push -u origin foo:bar
```

At best this will run `git push origin :foo` which won't match anything but if there's already a branch on the remote called `foo` that'll get wiped as well. We should fix this by looking at `upstreamWithoutRemote` if the branch is local and `nameWithoutRemote` if it's remote

app/src/models/branch.ts
```diff
@@ -68,6 +68,20 @@ export class Branch {
...
+    if (this.type === BranchType.Local) {
+      return null
+    }
+
+    const pieces = this.name.match(/(.*?)\/.*/)
+    if (!pieces || pieces.length < 2) {
+      return null
+    }
+
+    return pieces[1]
```
niik: I don't think we can assume that the name of a remote branch will always start with the remote name. `name` is derived from `%(refname:short)` which attempts to give you the shortest possible **non-ambiguous** name. As a convoluted example if you've got a remote branch `refs/remotes/origin/foo` and a local branch called `origin/foo` the remote branch name would be `remotes/origin/foo` as that's the shortest non-ambiguous name.

I think we need to have access to the canonical ref here. We extract it from `for-each-ref` but we don't pass it to the branch currently.

app/src/lib/stores/app-store.ts
```diff
@@ -3344,49 +3346,87 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      while (i < branchesState.recentBranches.length) {
+        if (branchesState.recentBranches[i].name !== branchToDelete.name) {
+          branchToCheckout = branchesState.recentBranches[i]
+          break
+        }
+        i++
+      }
+    }
+
+    if (branchToCheckout === null) {
```
niik: I know this isn't your doing but am I missing something or is this a very convoluted way of saying

```suggestion
    // If the default branch is null, use the most recent branch excluding the branch
    // the branch to delete as the branch to checkout.
    const branchToCheckout =
      branchesState.defaultBranch ??
      branchesState.recentBranches.find(x => x.name !== branch.name)
    
    if (branchToCheckout === undefined) {
```

niik (dismissed):
Hey friends, I'm trying to get up to speed with this PR in order to be able to have an informed opinion on the discussion y'all are having. I need to jump into a meeting but in the meantime here's a few thoughts I had while reading through the changes.

app/src/lib/git/branch.ts
niik: > (I'd also love to know what's @niik 's take on this 😆 Sorry!)

Hey, sorry about the delay, took me a time to get up to speed here. Great discussion already here :heart: and I think both directions has merit.

What speaks in favor of keeping the deleteBranch method as-is is that I'm a little concerned about the amount of change we're introducing in order to support the branch context menu. While some refactors were inevitable and necessary I wonder if we could have extracted some of the more architectural changes into separate PRs. My gut instinct would have been to try to minimize the surface area where we could introduce bugs as much as possible. I think I would probably just have introduced a new method for deleting a remote branch that could have been used by the new dialog and kept the other changes to a bare minimum to be able to reason about the impact of this change more easily.

On the other hand, I'll say that in general I'm very much in favor of method doing as little as possible (i.e. deleteLocalBranch  and deleteRemoteBranch vs a deleteBranch with a  parameter to include the remote). This goes double for methods in the `git` folder where we should aim for keeping them very focused the git work as opposed to app logic.

So in this case I draw the line at the appStore. In an ideal world (read: this isn't at all the case at the moment and that's fine) app store methods should represent UX actions. I.e. if we show the user a dialog asking them to confirm deletion of a local branch and offer them a choice to also delete the remote branch I think it's absolutely fine for that app store method to take an `includeRemote` parameter as it encapsulates the user's intended operation. That method can then call both `deleteLocalBranch` and `deleteRemoteBranch` which each encapsulate git-level operations.

The presence of a `deleteLocalBranch` in the git layer which optionally deletes the remote branch seems counterintuitive as there's no such operation in git. So my preference would be two dedicated git methods for deleting a local and a remote branch. An added benefit of this is that we no longer need to be passing these methods `Branch` objects as they'll be so much simpler now. The `deleteLocalBranch` can just take a local branch name and the deleteRemoteBranch can take a remote name and a branch to delete on that remote. That way we should get rid of a lot of the trickery of checking whether the branch object it local or not and leave that to the app store method which, if I understand this correctly, is the only caller of these methods where it matters.

I'm like 99% sure I'm just muddying the waters of this discussion by contributing here but those are my initial thoughts. I hope they're helpful but I trust the both of you to find a solution you're both happy with.

sergiou87: Thanks for chiming in!! That was a really insightful read ❤️ 

> **So in this case I draw the line at the appStore**. In an ideal world (read: this isn't at all the case at the moment and that's fine) app store methods should represent UX actions. I.e. if we show the user a dialog asking them to confirm deletion of a local branch and offer them a choice to also delete the remote branch I think it's absolutely fine for that app store method to take an includeRemote parameter as it encapsulates the user's intended operation. That method can then call both deleteLocalBranch and deleteRemoteBranch which each encapsulate git-level operations.

I personally like this idea.

> On the other hand, I'll say that in general I'm very much in favor of method doing as little as possible (i.e. deleteLocalBranch and deleteRemoteBranch vs a deleteBranch with a parameter to include the remote). This goes double for methods in the git folder where we should aim for keeping them very focused the git work as opposed to app logic.

And this is a very nice takeaway 😍 

So **ideally** the logic in `deleteBranch` would be moved up to the app store, while in `git` land we'd have the much simpler `deleteLocalBranch` and `deleteRemoteBranch`. I don't think that should be too problematic… except for the fact that we'd lose some valuable tests from `branch-test.ts`.

So I think the best trade-off right now would be to have only one `_deleteBranch` function in app store as you suggest to reflect the UX action, move the current `deleteBranch` from `branch.ts` to its own file (for a clear separation between core `git` stuff vs higher level policies) and be more specific about what it does (`deleteLocalBranchAndUpstream`? `deleteLocalAndTrackingBranch`? `deleteLocalAndRemoteBranch`?), while keeping its tests. Then the app store `_deleteBranch` function will either:
1. Call the shiny new `deleteLocalBranchAndUpstream` (or whatever) for local branches.
2. Call `deleteRemoteBranch` for remote branches.

That said, I agree this might make more sense in a different PR, to keep the refactor part away from the actual feature 😅

app/src/lib/git/branch.ts
tidy-dev: Nice! 

Since this is a change to the original functionality, I will open a PR just to address it (probably as an on top of branch.ts refactor branch) and test it out including the scenario about mistakenly deleting the wrong remote branch.

app/src/models/branch.ts
tidy-dev: To play with this I made a branch called `origin/test` along with my `test` and current code doesn't even seem to know what to do with it to the point where it errors with popup. 

I will make a PR to introduce the conical ref, and I think there may already be an underlying issue with how local names are parsed... thus this might be useful for both.... but needs further investigation.

app/src/lib/git/branch.ts
tidy-dev: Awesome discussions! 

I am definitely in favor of moving this refactor out to it's own PR. 

Taking in to consideration @niik's other comments. I think I will create a separate PR to address adding the canonical ref to the branch and making the remoteName function there to provide this functionality separately. 

Additionally, seems the fix needed in the comment about possibly being able to delete non intended remote branches would be good on top of the delete functions refactor PR. 

Going to move this PR back to draft. After those issue get addressed, I will merge into the add-context-menu and go from there.  Thank you both for your great input!

app/src/lib/git/branch.ts
tidy-dev: PR https://github.com/desktop/desktop/pull/11483 now uses upstreamWithoutRemote when deleting remotes.

app/src/models/branch.ts
tidy-dev: PR https://github.com/desktop/desktop/pull/11490 to resolve this

app/src/lib/git/branch.ts
tidy-dev: I believe this is resolved now

app/src/lib/stores/app-store.ts
```diff
@@ -3346,17 +3346,29 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        }
+
+        await gitStore.performFailableOperation(() =>
+          deleteRemoteBranch(
+            r,
+            account,
+            branch.remoteName!,
+            branch.nameWithoutRemote
+          )
+        )
```
sergiou87: I think that while the use in `!` is kind of acceptable in tests, we shouldn't use it in production code.
If I understand correctly, you need it because the `if` above is checking on a computed property that might change its value. If so, I've seen this pattern in other parts of the code to prevent the use of `!`:

```suggestion
        const remoteName = branch.remoteName
        if (remoteName === null) {
          // This is based on the branches ref. It should not be null for a
          // remote branch
          throw new Error(
            `Could not determine remote name from: ${branch.ref}.`
          )
        }

        await gitStore.performFailableOperation(() =>
          deleteRemoteBranch(
            r,
            account,
            remoteName,
            branch.nameWithoutRemote
          )
        )
```

sergiou87: Not sure why the diff is so bad 😞 basically I extracted the `remoteName` to a constant, and thanks to that the compiler will help you out to avoid `!`.

app/src/lib/stores/app-store.ts
```diff
@@ -3390,7 +3402,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /**
-   * Deletes the local branch. If the parameter `includeRemote` is true, the
+   * Deletes the local branch. If the parameter `includeUpstream?` is true, the
```
sergiou87: Nit: typo?
```suggestion
   * Deletes the local branch. If the parameter `includeUpstream` is true, the
```

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -686,14 +686,24 @@ export class Dispatcher {
...
+  }
+
+  /**
+   * Delete the remote branch.
+   */
+  public deleteRemoteBranch(
+    repository: Repository,
+    branch: Branch
+  ): Promise<void> {
+    return this.appStore._deleteBranch(repository, branch)
```
sergiou87, line 706: As you know, I'm completely ok with having `deleteLocalBranch` and `deleteRemoteBranch` here (since I'm the one who suggested it in the first place 😂 ), but what I understood from [Markus' last comment](https://github.com/desktop/desktop/pull/11434#discussion_r566194963) is that the dispatcher would also keep its only `deleteBranch` method, and we would only have local/remote versions at the `app/lib/git/branch` level.

I'll leave this up to you 😄

sergiou87 (changes requested):
Looks great! 😍 

I only have a few comments about "style", but other than that it's perfect!

app/src/ui/dispatcher/dispatcher.ts
tidy-dev: I went back and forth on this. I already had it from one of past revisions and now couldn't decide if I want to revert it back. 😄 

We are splitting it out into two concepts already via the `delete-branch-dialog` and `delete-remote-branch-dialog`. 

Because they are so simple, I am inclined to keep the distinction.

app/src/lib/stores/app-store.ts
tidy-dev: Ah - didn't realize using the ! was not ideal. Thanks! I updated per your suggestion.

app/src/ui/dispatcher/dispatcher.ts
sergiou87: Sounds good!

sergiou87 (approved):
Looks perfect now, thanks for all those changes! ❤️ :shipit:

PR #11445, opened 2021-01-22
Add beta release notes to area of README where we suggest downloading it 

billygriffin:
Closes #11444

## Description

This adds a link to the download portion of the README so folks are able to see what has shipped in recent beta releases.

PR #11452, opened 2021-01-24
Beta Release Notes Url

### LINKED ISSUE #3385: Beta Release Notes Don't Point to the Right Place

say25:
Closes #3385

## Description

Inspired by https://github.com/desktop/desktop/pull/11445, thanks @billygriffin

- Beta Release Notes Point to the Right Place

## Release notes

Notes: Beta Release Notes Point to the correct place

### DIFF
--- a/app/src/ui/lib/releases.ts
+++ b/app/src/ui/lib/releases.ts
@@ -1,1 +1,3 @@
-export const ReleaseNotesUri = 'https://desktop.github.com/release-notes/'
+export const ReleaseNotesUri = __RELEASE_CHANNEL__ === 'beta'
+  ? 'https://desktop.github.com/release-notes/?env=beta'
+  : 'https://desktop.github.com/release-notes/'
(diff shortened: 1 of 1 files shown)


billygriffin (dismissed):
Oh wow this is cool! I'd love another reviewer but conceptually this is great. ❤️

niik (changes requested):
This is great @say25 thanks so much for fixing this. The only ask I have is if we could just check if `__RELEASE_CHANNEL__ === 'beta'` inside of `releases.ts`? I'm not super fond of overloading the feature-flag method for this as it's not conceptually what we want to be checking for.

It would also match how we determine which release notes to pull down to show in the release-notes dialog: 
https://github.com/desktop/desktop/blob/6b65413b83f5d035576e860ab6023177c19b6e0c/app/src/lib/release-notes.ts#L86

niik (approved):
Thank you @say25!

PR #11463, opened 2021-01-26
Branch fast-forwarding improvements: round 2

sergiou87:
## Description

This work is a continuation of #11387 which now focuses on the actual action of fast-forwarding the branches.

Instead of using `update-ref` with each branch, the changes in this PR switch the implementation to just use `fetch` on all fast-forwardable branches, making it faster since only one `git` invocation is needed to fast-forward all the eligible branches.

Again, a good way to create an environment to test these changes is running this:
```bash
git clone https://github.com/ARMmbed/mbed-os.git
cd mbed-os

# Move ref some commits behind
git checkout feature-bluetooth-traces
git reset --hard dd4eca1b693c253626481fb400a9451458f1f161

# Move ref some commits behind
git checkout feature-cmake
git reset --hard 66f65c0e89dea555993416376dd980df23d66d48

# Move ref to another ref (X commits behind and Y commits ahead)
git checkout feature-i2c
git reset --hard feature-wisun-1.0.0

# Create branch without upstream
git branch no-upstream-branch 61f2097223060b6cebd2246dde3318fa0a127a5f

# Check out branch and don't move it
git checkout feature-hal-spec-sai
```

However, a proper "unit" test has been included in the changeset.

## Release notes

Notes: [Improved] Automatic branch fast-forwarding is now faster.


app/src/lib/stores/app-store.ts
```diff
@@ -3778,37 +3776,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-          localRef,
-          branch.tip.sha,
-          upstreamRef,
-          'pull: Fast-forward'
-        )
-      }
+      await fastForwardBranches(repository, eligibleBranches)
+    } catch (e) {
+      log.error(`Branch fast-forwarding failed (${e})`)
+      sendNonFatalException('fastForwardBranches', e)
```
sergiou87, line 3846: As discussed, if there are issues with the command line length when running `git fetch` to fast-forward branches, we'll report it here as non-fatal.

app/src/models/branch.ts
```diff
@@ -21,6 +21,14 @@ export interface ICompareResult extends IAheadBehind {
   readonly commits: ReadonlyArray<Commit>
 }
 
+/** Basic data about a branch. */
+export interface IBranchBasicInfo {
```
sergiou87: I'm open to suggestions for better names 😂

app/src/lib/git/for-each-ref.ts
```diff
@@ -160,7 +158,7 @@ export async function getBranchesDifferingFromUpstream(
       continue
     }
 
-    if (ref.startsWith('refs/head')) {
+    if (ref.startsWith('refs/heads')) {
```
niik, line 166: 🤦 How did I miss that?

app/src/models/branch.ts
niik: How about `ITrackingBranch`? Seeing how this differs from a regular `Branch` object in that it's guaranteed to always have an `upstreamRef` and `upstreamSha`

app/src/lib/git/for-each-ref.ts
```diff
@@ -109,12 +109,10 @@ export async function getBranches(
...
  */
 export async function getBranchesDifferingFromUpstream(
-  repository: Repository,
-  allBranches: ReadonlyArray<Branch>
-): Promise<ReadonlyArray<Branch>> {
+  repository: Repository
+): Promise<ReadonlyArray<IBranchBasicInfo>> {
   const format = [
     '%(refname)',
     '%(refname:short)',
```
niik: Looks like we're not using the shortname anymore. Can we drop it?

app/src/lib/git/fetch.ts
```diff
@@ -128,3 +129,37 @@ export async function fetchRefspec(
...
+      GIT_REFLOG_ACTION: 'pull',
+    },
+  }
+
+  const refPairs = branches.map(branch => `${branch.upstreamRef}:${branch.ref}`)
+
+  await git(
+    [
+      '-c',
+      'fetch.output=full',
```
niik: This is in preparation of step 3 right? Could we drop this until we get to actually parsing the output?

app/src/lib/git/fetch.ts
```diff
@@ -128,3 +129,37 @@ export async function fetchRefspec(
...
+      'fetch',
+      '.',
+      '--show-forced-updates',
+      '-v',
+      // TODO: Once we upgrade to git 2.29 we can use --stdin to pass the refs,
+      // avoiding hitting any shell limitations related to the length of the
+      // command.
+      ...refPairs,
```
niik: I think we need to explicitly check that the `refPairs` (or `branches`) array is not empty before spreading here or else we could end up running `git fetch .` without any refs which would be redundant at best and will, if nothing else, clear `FETCH_HEAD`

niik (changes requested):
This is looking great @sergiou87. One thing that I know we talked about was whether we should use `--append` for `FETCH_HEAD` (or more specifically figuring out exactly what `FETCH_HEAD` really does). Were you able to look into that as part of this PR?

app/src/lib/git/fetch.ts
sergiou87: 😱 😱 😱 😱 😱 😱

sergiou87:
I totally forgot about the `FETCH_HEAD` thing 😖 I'll take a look tomorrow, thanks for the review!! ❤️

app/src/lib/git/for-each-ref.ts
sergiou87: Nice catch! TypeScript didn't complain when I stopped using `name` in the `localBranches` array below 😖

sergiou87:
Thanks to yesterday's git bump to 2.29.2, I was able to make two important changes:
- Pass the refs via stdin instead of as arguments: that'll circumvent line length limitations in some shells (looking at you, Windows 👀 )
- Add [the shiny new `--no-write-fetch-head` argument](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---no-write-fetch-head) to our `git fetch` invocation to address the `FETCH_HEAD` concern you had 😌  I also added a unit test to verify it doesn't touch `FETCH_HEAD` ✨ 

Ready for another review @niik !! 😄

app/src/lib/git/fetch.ts
```diff
@@ -128,3 +129,44 @@ export async function fetchRefspec(
...
+      GIT_REFLOG_ACTION: 'pull',
+    },
+    stdin: refPairs.join('\n'),
+  }
+
+  await git(
+    [
+      'fetch',
+      '.',
+      '-v',
```
niik: Nitpick here but since we're note parsing the output I'm guessing we can drop this? At least until we get to phase 3?

app/src/lib/git/fetch.ts
```diff
@@ -128,3 +129,44 @@ export async function fetchRefspec(
...
+      // Make sure we don't try to update branches that can't be fast-forwarded
+      // even if the user disabled this via the git config option
+      // `fetch.showForcedUpdates`
```
niik, line 161: Love this explanation (and the other explanations of the command line args here :heart:

app/src/lib/git/for-each-ref.ts
```diff
@@ -153,45 +150,41 @@ export async function getBranchesDifferingFromUpstream(
...
 
-      localBranches.push({ name, ref, sha, upstream })
+      localBranches.push({ ref, sha, upstream })
     } else {
       remoteBranchShas.set(ref, sha)
     }
   }
 
-  const eligibleBranchNames = new Set<String>()
+  const eligibleBranches: ITrackingBranch[] = []
```
niik: These are semantically equivalent but we typically use the 👇  syntax nowadays

```suggestion
  const eligibleBranches = new Array<ITrackingBranch>()
```

app/src/lib/stores/app-store.ts
```diff
@@ -3778,37 +3776,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-          repository,
-          localRef,
-          branch.tip.sha,
-          upstreamRef,
-          'pull: Fast-forward'
-        )
-      }
+      await fastForwardBranches(repository, eligibleBranches)
+    } catch (e) {
+      log.error(`Branch fast-forwarding failed (${e})`)
```
niik: The `log.*` methods accept an error object as the second argument and formats it nicely for us
```suggestion
      log.error('Branch fast-forwarding failed', e)
```

app/src/lib/stores/app-store.ts
```diff
@@ -3778,37 +3776,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      )
 
-      const { ahead, behind } = aheadBehind
-      // Only perform the fast forward if the branch is behind it's upstream
-      // branch and has no local commits.
-      if (ahead === 0 && behind > 0) {
-        // At this point we're guaranteed this is non-null since we've filtered
-        // out any branches will null upstreams above when creating
-        // `eligibleBranches`.
-        const upstreamRef = branch.upstream!
```
niik, line 3861: Yessss, one less `!` assertion in favor of true type safety!

app/src/lib/git/fetch.ts
```diff
@@ -128,3 +129,44 @@ export async function fetchRefspec(
...
+  branches: ReadonlyArray<ITrackingBranch>
+): Promise<void> {
+  if (branches.length === 0) {
+    return
+  }
+
+  const refPairs = branches.map(branch => `${branch.upstreamRef}:${branch.ref}`)
+
+  const opts: IGitExecutionOptions = {
+    successExitCodes: new Set([0, 1]),
```
niik, line 146: Could we add a small note about this for the future reader? Perhaps something like
```suggestion
    // Fetch exits with an exit code of 1 if one or more refs failed to update
    // which is what we expect will happen
    successExitCodes: new Set([0, 1]),
```

niik (dismissed):
Just nitpicks left now, this is looking fantastic!

soexcited

app/test/unit/git/for-each-ref-test.ts
```diff
@@ -71,26 +71,21 @@ describe('git/for-each-ref', () => {
...
-      const allBranches = await getBranches(repository)
+      const branches = await getBranchesDifferingFromUpstream(repository)
 
-      const branches = await getBranchesDifferingFromUpstream(
-        repository,
-        allBranches
-      )
-
-      const branchNames = branches.map(branch => branch.name)
+      const branchNames = branches.map(branch => branch.ref)
```
tidy-dev: Nit: Suggest to rename this variable to branchRefs

sergiou87: Nice catch!

tidy-dev (dismissed):

tidy-dev (approved):

PR #11466, opened 2021-01-26
Rename branch.remote to branch.upstreamRemote

tidy-dev:
## Description

Rename `Branch` property `remote` to `upstreamRemote` as is a more accurate description.


niik (dismissed):
:sparkles:

sergiou87 (commented):
Looks good! But if we're later reintroducing `remote` as `remoteName`, shouldn't this one be `upstreamRemoteName`? 😳

tidy-dev:
> Looks good! But if we're later reintroducing `remote` as `remoteName`, shouldn't this one be `upstreamRemoteName`? 😳

Makes sense

sergiou87 (approved):
Thanks for the change! LGTM now 😄

niik (approved):

PR #11469, opened 2021-01-27
Update dugite to bump Git to 2.29.2

tidy-dev:
Update dugite to 1.96 for git 2.29.2

### DIFF
--- a/app/test/unit/git/worktree-test.ts
+++ b/app/test/unit/git/worktree-test.ts
@@ -128,13 +128,13 @@ describe('git/worktree', () => {
     it('creates worktree at temporary path', async () => {
       const workTree = await createTemporaryWorkTree(repository, 'HEAD')
       const tmpDir = Os.tmpdir()
-      const directories = await FSE.readdir(realpathSync(tmpDir))
-
-      expect(workTree.head).toBe(currentHeadSha)
       // we use realpathSync here because git and windows/macOS report different
       // paths even though they are the same folder
-      const path = realpathSync(workTree.path).split(/\/|\\/)
-      const folder = path[path.length - 1]
+      const directories = await FSE.readdir(realpathSync(tmpDir))
+      const folder = Path.basename(realpathSync(workTree.path))
+
+      expect(workTree.head).toBe(currentHeadSha)
+
       // We are checking if the last folder on worktree path exists in tmpDir
       // because tmpDir on windows uses 8.3 short names
       // and workTree path uses long names; thus, we cannot simply compare paths.
--- a/app/test/unit/git/worktree-test.ts
+++ b/app/test/unit/git/worktree-test.ts
@@ -127,18 +127,17 @@ describe('git/worktree', () => {
 
     it('creates worktree at temporary path', async () => {
       const workTree = await createTemporaryWorkTree(repository, 'HEAD')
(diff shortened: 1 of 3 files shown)


app/test/unit/git/worktree-test.ts
```diff
@@ -128,11 +128,14 @@ describe('git/worktree', () => {
...
       const workTree = await createTemporaryWorkTree(repository, 'HEAD')
       const tmpDir = Os.tmpdir()
+      const directories = await FSE.readdir(realpathSync(tmpDir))
 
       expect(workTree.head).toBe(currentHeadSha)
       // we use realpathSync here because git and windows/macOS report different
       // paths even though they are the same folder
-      expect(realpathSync(workTree.path)).toStartWith(realpathSync(tmpDir))
+      const path = realpathSync(workTree.path).split(/\/|\\/)
+      const folder = path[path.length - 1]
```
sergiou87: I think you can use `Path.basename` for this:
```suggestion
      const folder = Path.basename(workTree.path)
```

sergiou87: Also probably worth writing a comment about why are we doing this instead of the `toStartWith` 🤔  Specially now that the test could start giving false  positives.

sergiou87 (changes requested):

app/test/unit/git/worktree-test.ts
tidy-dev: Done - Great input!

app/test/unit/git/worktree-test.ts
```diff
@@ -128,11 +128,17 @@ describe('git/worktree', () => {
...
-      expect(realpathSync(workTree.path)).toStartWith(realpathSync(tmpDir))
+      const directories = await FSE.readdir(realpathSync(tmpDir))
+      const folder = Path.basename(realpathSync(workTree.path))
+
+      expect(workTree.head).toBe(currentHeadSha)
+
+      // We are checking if the last folder on worktree path exists in tmpDir
+      // because tmpDir on windows uses 8.3 short names
+      // and workTree path uses long names; thus, we cannot simply compare paths.
+      expect(directories).toContain(folder)
```
sergiou87: Nit that I overlooked in my previous review, sorry 😖  `directories` sounds a bit odd when it's referring only to the tmp dir, so maybe just use `tmpDir` directly? 🤔 
```suggestion
      const tmpDir = await FSE.readdir(realpathSync(Os.tmpdir()))
      const folder = Path.basename(realpathSync(workTree.path))

      expect(workTree.head).toBe(currentHeadSha)

      // We are checking if the last folder on worktree path exists in tmpDir
      // because tmpDir on windows uses 8.3 short names
      // and workTree path uses long names; thus, we cannot simply compare paths.
      expect(tmpDir).toContain(folder)
```

sergiou87 (dismissed):
Only one nit that I think would be nice to change if you agree, otherwise LGTM! Thanks for the fix 😄

app/test/unit/git/worktree-test.ts
tidy-dev: Always good to be more precise in naming :)

sergiou87 (approved):
Thanks for that last change! ❤️ :shipit:

PR #11477, opened 2021-01-29
Update instructions to stop a release mid-flight

sergiou87:
We don't use Appveyor nor CircleCI anymore, only GitHub Actions 😌 💪


niik (approved):
:sparkles:

PR #11483, opened 2021-01-29
Refactor delete branch methods

tidy-dev:
## Description

Based on discussions on PR https://github.com/desktop/desktop/pull/11434 about adding delete context menus to branches. We want to refactor the delete branch methods in branches to be more singular in responsibility to hopefully improve readability and clarify the intent of the code. This also creates a clearer separation from UI layer (handling user actions such as "I also want to delete the remote branch.") and the more api type layer handling commands such as `deleteLocalBranch` and `deleteRemoteBranch`.

- Removing `deleteBranch` from `branch.ts` as it was too broad of scope in deleting local and remote branches as well as deciding between those.
- Add `deleteRemoteBranch` to `branch.ts` to handle deleting remote branches.
- Refactor `app-store.ts` to handle the `includeRemote` logic from the user interface and choose when to use `deleteLocalBranch` and `deleteRemoteBranch`.
- During refactor, I changed the reference for getting the branch name when deleting remotely to use `upstreamWithoutRemote` as opposed to `nameWithoutRemote`

I confirmed what @niik pointed out -> that for that last bullet point (prior to this fix) that if I 

1. Created a branch `testA`, pushed it to remote, deleted it locally (`testA` only exists remotely)
image
2. Created a branch `testB`, pushed it to remote, renamed it locally to `testA`
image
image
3. Deleted `testA` and checked to include also delete it remotely.
image
4. Wallah! I now deleted remote `origin/testA` and am left with `origin/testB`
image

Albeit that the above scenario is unlikely... it is a one variable name change to delete the actual upstream or tracking branch to what the user selected... which "should" be what the user intends... and not using this as a round about way of deleting a remote branch without checking it out :D

Notes: no-notes

### DIFF
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -65,6 +65,7 @@ export async function deleteLocalBranch(
 
 /**
  * Deletes a remote branch
+ *
  * @param remoteName - the name of the remote to delete the branch from
  * @param remoteBranchName - the name of the branch on the remote
  * @param branchTipSha - the branch's sha is logged for reference, more useful
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -68,15 +68,12 @@ export async function deleteLocalBranch(
  *
  * @param remoteName - the name of the remote to delete the branch from
  * @param remoteBranchName - the name of the branch on the remote
- * @param branchTipSha - the branch's sha is logged for reference, more useful
- * for a remote only branch if the user wants to undo this delete.
  */
 export async function deleteRemoteBranch(
   repository: Repository,
   account: IGitAccount | null,
   remoteName: string,
-  remoteBranchName: string,
-  branchTipSha: string
+  remoteBranchName: string
 ): Promise<true> {
   const networkArguments = await gitNetworkArguments(repository, account)
   const remoteUrl =
@@ -96,10 +93,6 @@ export async function deleteRemoteBranch(
     env: await envForRemoteOperation(account, remoteUrl),
     expectedErrors: new Set<DugiteError>([DugiteError.BranchDeletionFailed]),
   })
-
-  if (branchTipSha !== undefined) {
-    log.info(`Deleted branch ${remoteBranchName} (was ${branchTipSha})`)
-  }
 
   // It's possible that the delete failed because the ref has already
   // been deleted on the remote. If we identify that specific
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -103,7 +103,7 @@ export async function deleteRemoteBranch(
 
   // It's possible that the delete failed because the ref has already
   // been deleted on the remote. If we identify that specific
-  // error we can safely remote our remote ref which is what would
+  // error we can safely remove our remote ref which is what would
   // happen if the push didn't fail.
   if (result.gitError === DugiteError.BranchDeletionFailed) {
     const ref = `refs/remotes/${remoteName}/${remoteBranchName}`
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -1,6 +1,6 @@
 import { git, gitNetworkArguments } from './core'
 import { Repository } from '../../models/repository'
-import { Branch, BranchType } from '../../models/branch'
+import { Branch } from '../../models/branch'
 import { IGitAccount } from '../../models/git-account'
 import { formatAsLocalRef } from './refs'
 import { deleteRef } from './update-ref'
@@ -53,8 +53,7 @@ export async function renameBranch(
 }
 
 /**
- * Delete the branch locally, see `deleteBranch` if you're looking to delete the
- * branch from the remote as well.
+ * Delete the branch locally.
  */
 export async function deleteLocalBranch(
   repository: Repository,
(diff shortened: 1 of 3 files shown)


app/src/lib/git/branch.ts
```diff
@@ -65,54 +64,50 @@ export async function deleteLocalBranch(
...
-    // been deleted on the remote. If we identify that specific
-    // error we can safely remote our remote ref which is what would
-    // happen if the push didn't fail.
-    if (result.gitError === DugiteError.BranchDeletionFailed) {
-      const ref = `refs/remotes/${remoteName}/${branch.nameWithoutRemote}`
-      await deleteRef(repository, ref)
-    }
+  // It's possible that the delete failed because the ref has already
+  // been deleted on the remote. If we identify that specific
+  // error we can safely remote our remote ref which is what would
```
sergiou87: Nit: this is not your typo, but is still a typo 😂 
```suggestion
  // error we can safely remove our remote ref which is what would
```

app/src/lib/git/branch.ts
```diff
@@ -65,54 +64,50 @@ export async function deleteLocalBranch(
...
+
+  // If the user is not authenticated, the push is going to fail
+  // Let this propagate and leave it to the caller to handle
+  const result = await git(args, repository.path, 'deleteRemoteBranch', {
+    env: await envForRemoteOperation(account, remoteUrl),
+    expectedErrors: new Set<DugiteError>([DugiteError.BranchDeletionFailed]),
+  })
+
+  if (branchTipSha !== undefined) {
+    log.info(`Deleted branch ${remoteBranchName} (was ${branchTipSha})`)
```
sergiou87: Looking at the rest of the codebase, it doesn't look like we're leaving many traces for debugging, so I think I would remove this since:
1. It's logging a successful case for a very probable scenario (I see more useful logging errors in this case, which we are probably already doing in the git invocation).
2. It forces us to add `branchTipSha` as a parameter, just for logging purposes 😕 

Is there any good reason to keep this log line and "pollute" the function signature with `branchTipSha`? 😳

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,89 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-          i++
-        }
-      }
+      const branchToCheckout = this.getBranchToCheckoutAfterDelete(branch, r)
+      const gitStore = this.gitStoreCache.get(r)
 
-      if (branchToCheckout === null) {
-        throw new Error(
-          `It's not possible to delete the only existing branch in a repository.`
+      if (branchToCheckout != null) {
```
sergiou87: Shouldn't this be `!==`? I thought the linter would fail on these ones 🤔 
```suggestion
      if (branchToCheckout !== null) {
```

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,89 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
         )
       }
 
-      const nonNullBranchToCheckout = branchToCheckout
-      const gitStore = this.gitStoreCache.get(r)
+      await gitStore.performFailableOperation(() => {
+        if (branch.type === BranchType.Remote) {
+          // Note: deleting a remote branch implementation not needed, yet.
+          return Promise.resolve(true)
+        }
```
sergiou87: Since this PR doesn't do anything with remote branches, maybe this should be at the beginning of the `_deleteBranch` method.

In theory, you don't need to switch branches or anything when deleting a remote branch anyway.

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,89 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          includeRemote
+        )
+      })
+
+      return this._refreshRepository(r)
+    })
+  }
+
+  /**
+   * Deletes the local branch. If the branch `includeRemote` is true, the
```
sergiou87: Should `branch` be `parameter` here?
```suggestion
   * Deletes the local branch. If the parameter `includeRemote` is true, the
```

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,89 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+   * upstream branch will be deleted also.
+   */
+  private async deleteLocalBranchAndUpstreamBranch(
+    repository: Repository,
+    branch: Branch,
+    account: IGitAccount | null,
+    includeRemote?: boolean
+  ): Promise<true> {
```
sergiou87: Does that `true` give us any value? If its return value is always the same, I'd say we stick to return `void` for clarity's sake 🤔 
```suggestion
  ): Promise<void> {
```

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,89 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  private getBranchToCheckoutAfterDelete(
+    branchToDelete: Branch,
+    repository: Repository
+  ): Branch | null {
+    const { branchesState } = this.repositoryStateCache.get(repository)
+    const tip = branchesState.tip
+    const currentBranch = tip.kind === TipState.Valid ? tip.branch : null
+    // if current branch is not the branch being deleted, no need to switch
+    // branches
+    if (currentBranch != null && branchToDelete.name !== currentBranch.name) {
```
sergiou87: Hmmm `!=` again… is that a mistake too or it's intended and I'm missing some Javscript/TypeScript magic here? 😳 
```suggestion
    if (currentBranch !== null && branchToDelete.name !== currentBranch.name) {
```

From what I see, `currentBranch` can only be either a `Branch` or `null`, so checking for `!=` shouldn't be necessary (it can never be `undefined`). Am I right? 🤔

app/test/unit/git/branch-test.ts
```diff
@@ -186,74 +187,111 @@ describe('git/branch', () => {
...
       expect(localBranch).not.toBeUndefined()
+      expect(localBranch.upstreamRemoteName).not.toBeNull()
+      expect(localBranch.upstreamWithoutRemote).not.toBeNull()
+
+      if (
+        localBranch.upstreamRemoteName === null ||
+        localBranch.upstreamWithoutRemote === null
+      ) {
+        return
+      }
```
sergiou87: I think this is a leftover and should be removed, right?

app/test/unit/git/branch-test.ts
```diff
@@ -186,74 +187,111 @@ describe('git/branch', () => {
...
```
sergiou87: Another leftover?

sergiou87 (changes requested):
LGTM overall, just a few comments!

app/src/lib/git/branch.ts
tidy-dev: I agree it does pollute the function signature but it does have valid purpose (maybe there is a better approach). 

If a user deletes a remote branch on github.com, they can undo it. @niik informed me during pairing session, that a user can also restore a remote branch if they know the branches sha. By providing the sha in the log, we provide a way that the user can restore a mistakenly deleted remote branch, which is also an effort to provide similar functionality to github.com.

I believe it is important to keep it. It also seems logical to place in the `deleteRemoteBranch` to ensure it is captured with that action, but another option for us is to back it out to the `app-store.ts` `deleteBranch` method. What do you think about that?

app/test/unit/git/branch-test.ts
tidy-dev: The setup up for the test should ensure these are not null. But, the linter doesn't know that; those these are here to appease it. I added the expect not nulls to error out before this in the case that something would ever change making these actually be null because then the test would not make sense.

app/test/unit/git/branch-test.ts
tidy-dev: same.

app/src/lib/stores/app-store.ts
tidy-dev: True, I will move it - just wasn't focusing on that logic in this PR.

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,98 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      const gitStore = this.gitStoreCache.get(r)
+
+      // If soley a remote branch, there is no need to checkout a branch.
+      if (branch.type === BranchType.Remote) {
+        await gitStore.performFailableOperation(() => {
+          // Note: deleting a remote branch implementation not needed, yet.
+          return Promise.resolve()
+        })
+
+        return this._refreshRepository(r)
```
sergiou87: AFAIK none of this is needed:
- The `performFailableOperation` is only used to wrap `git` calls (which is not happening here).
- `_refreshRepository` should have nothing to refresh since we're not doing anything at all for remote branches.

Am I missing anything? 😳 
```suggestion
        return
```

sergiou87: Oh, unless you're just leaving a "placeholder" for when the implementation for remote branches is added 😅  (I'd still prefer to remove it and add the implementation in the right PR, but I don't have a strong opinion about it since that new PR is almost done already 😆 ).

app/test/unit/git/branch-test.ts
sergiou87: Ohhh, ok! In other tests I saw we just used `!` in situations like that:
```ts
      await deleteRemoteBranch(
        mockLocal,
        null,
        localBranch.upstreamRemoteName!,
        localBranch.upstreamWithoutRemote!,
        localBranch.tip.sha
      )
```

I'm happy with both approaches, although if you prefer to keep the `if`, I'd add a comment to explain why it's there (in case someone wonders the same I did 🤣 ).

app/src/lib/git/branch.ts
sergiou87: > but another option for us is to back it out to the `app-store.ts` `deleteBranch` method. What do you think about that? 

I think that would bother me less, yeah 😳

tidy-dev: Done :)

app/src/lib/stores/app-store.ts
tidy-dev: Just leaving as placeholder - Since I am planning merging this into the add context menu one today. I would prefer to keep it.

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,102 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-        while (i < branchesState.recentBranches.length) {
-          if (branchesState.recentBranches[i].name !== branch.name) {
-            branchToCheckout = branchesState.recentBranches[i]
-            break
-          }
-          i++
-        }
+      const gitStore = this.gitStoreCache.get(r)
+
+      // If soley a remote branch, there is no need to checkout a branch.
```
sergiou87: Nit: typo? (you can fix it in the other PR if nothing else comes up in this one, don't worry 😆 )
```suggestion
      // If solely a remote branch, there is no need to checkout a branch.
```

app/src/lib/stores/app-store.ts
```diff
@@ -3347,44 +3348,102 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+  /**
+   * Deletes the local branch. If the parameter `includeRemote` is true, the
+   * upstream branch will be deleted also.
+   */
+  private async deleteLocalBranchAndUpstreamBranch(
+    repository: Repository,
+    branch: Branch,
+    account: IGitAccount | null,
+    includeRemote?: boolean
+  ): Promise<void> {
```
sergiou87: Nit: just "thinking out loud", but feels weird calling it **upstream** branch in the function name, but then the parameter is `includeRemote` instead of `includeUpstream`. Maybe we should use the same name in both places? Either remote or upstream (upstream is probably less confusing, IMO).

What do you think?

If you agree, this can be done in your other PR, don't worry about that!

sergiou87 (approved):
Beautiful ✨  Thanks for the changes! :shipit:

PR #11486, opened 2021-01-29
git for-each-ref and log parser

niik:
## Description

@sergiou87 Just tossing all of this code over the wall as I said I would. Do something with it or close this PR and forget all about it. Either way is absolutely fine with me but I didn't want this to just bitrot on my machine until after my parental leave.


sergiou87:
I fixed a typo and changed a couple of lines for better readability, and then used the new parser in `getBranchesDifferingFromUpstream`. I love it!!!! 😍 😍 😍 😍 

I think this is now ready for another review from @tidy-dev 👀

app/src/lib/git/log.ts
```diff
@@ -106,58 +104,26 @@ export async function getCommits(
...
-      throw new Error(`Couldn't parse author identity for '${shortSha}'`)
-    }
-
-    const committer = CommitIdentity.parseIdentity(committerIdentity)
-
-    if (!committer) {
-      throw new Error(`Couldn't parse committer identity for '${shortSha}'`)
-    }
 
     return new Commit(
```
tidy-dev, line 115: This is beautiful! 😍

tidy-dev (dismissed):
This is just so pretty! 

image

app/src/lib/git/log.ts
tidy-dev: ... meant to grab a set of lines here -> the refactor of log is what I was referring too. 🤣

tidy-dev (approved):

PR #11489, opened 2021-01-29
Feature/add discard all changes option on branch switch

### LINKED ISSUE #11491: Add a "Discard all changes" option on branch switch attempt

Attempting to switch branches while changes haven't been handled gives currently 2 options:

image

In the case where you want to discard those changes, to continue switching branches unhindered -

You'd need to exit the dialog and do the action in the `Branch` tab.

(To discard all changes you can currently use the `Branch` tab option `discard all changes`.)

### Proposed solution

Adding a 3rd option in the switch branch dialog, to discard all changes.

The proposed change is also already pretty fitting for this dialog -

Since the associated action has already been related to the `stash` action, in the `Branch` tab.

Which is already an option in this dialog.

Example for how the solution would look:

image

SHOSH-NI:
Closes #11491

## Description

- Allows switching branches by discarding all changes on current branch.

### Screenshots
Screenshot 2021-01-29 202511

## Release notes

Notes:

 - Adds a `Discard All` option to submit in the `Switch branch` action Dialog

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3155,7 +3155,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     repository: Repository,
     branch: Branch,
     account: IGitAccount | null
-  ){
+  ) {
     const repositoryState = this.repositoryStateCache.get(repository)
     const { workingDirectory } = repositoryState.changesState
     const { tip } = repositoryState.branchesState
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3067,7 +3067,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return this.checkoutAndLeaveChanges(repository, branch, account)
     } else if (strategy === UncommittedChangesStrategy.MoveToNewBranch) {
       return this.checkoutAndBringChanges(repository, branch, account)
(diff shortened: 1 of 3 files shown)


billygriffin:
Hi @STRONG-MAD - it looks like you've done quite a bit of work here, so thank you for that. However, we don't accept PRs for features that haven't had the opportunity to be discussed in an issue and where someone from the Desktop team gives the 👍 to go ahead and create a PR for it. You are certainly welcome to create an issue for this, and we can go from there. I would ask you to share more context on how the existing behavior isn't working for you, and how you think about the tradeoff of adding yet another option here with the convenience it might provide.

My gut instinct is that the additional option introduces more confusion and probably shouldn't be in this flow, given that you're easily able to discard changes at any point on either side of switching branches. So I'm going to close this for now, but if you'd like to make a case for why you think this would be broadly beneficial to Desktop users, please feel free to open an issue. Thanks again!

SHOSH-NI:
Hi @billygriffin -I opened an issue.

Would it be ok until it gets a 👍 that we keep this PR open as WIP and link it to the new issue?

billygriffin:
Totally fine to link to this PR from the issue! But I'm going to keep it closed until we decide what we want to do here. Thank you! ❤️

javablockef (commented):
@regAHawlery

PR #11490, opened 2021-01-29
Add ref to branch

tidy-dev:
## Description

Add `ref` to the `branch` class to allow us to more accurately obtain a remote branch's remote. 

From @niik on PR https://github.com/desktop/desktop/pull/11434
> I don't think we can assume that the name of a remote branch will always start with the remote name. `name` is derived from `%(refname:short)` which attempts to give you the shortest possible **non-ambiguous** name. As a convoluted example if you've got a remote branch `refs/remotes/origin/foo` and a local branch called `origin/foo` the remote branch name would be `remotes/origin/foo` as that's the shortest non-ambiguous name.

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -1111,7 +1111,7 @@ export class GitStore extends BaseStore {
           status.currentUpstreamBranch || null,
           branchTipCommit,
           BranchType.Local,
-          ''
+          `refs/heads/${currentBranch}`
         )
         this._tip = { kind: TipState.Valid, branch }
       } else if (currentTip) {
--- a/app/src/lib/stores/git-store.ts
+++ b/app/src/lib/stores/git-store.ts
@@ -1110,7 +1110,8 @@ export class GitStore extends BaseStore {
           currentBranch,
           status.currentUpstreamBranch || null,
           branchTipCommit,
-          BranchType.Local
+          BranchType.Local,
+          ''
         )
         this._tip = { kind: TipState.Valid, branch }
       } else if (currentTip) {
--- a/app/src/models/branch.ts
+++ b/app/src/models/branch.ts
@@ -45,7 +45,7 @@ export class Branch {
    * @param upstream The remote-prefixed upstream name. E.g., `origin/main`.
    * @param tip Basic information (sha and author) of the latest commit on the branch.
    * @param type The type of branch, e.g., local or remote.
-   * @param ref The connical ref of the branch
+   * @param ref The canonical ref of the branch
    */
   public constructor(
     public readonly name: string,
--- a/app/src/models/branch.ts
+++ b/app/src/models/branch.ts
@@ -77,7 +77,7 @@ export class Branch {
     }
 
     const pieces = this.ref.match(/^refs\/remotes\/(.*?)\/.*/)
-    if (!pieces || pieces.length < 2) {
+    if (!pieces || pieces.length === 2) {
       // This shouldn't happen, the remote ref should always be prefixed
       // with refs/remotes
       throw new Error(`Remote branch ref has unexpected format: ${this.ref}`)
--- a/app/src/models/branch.ts
+++ b/app/src/models/branch.ts
@@ -45,12 +45,14 @@ export class Branch {
    * @param upstream The remote-prefixed upstream name. E.g., `origin/main`.
    * @param tip Basic information (sha and author) of the latest commit on the branch.
    * @param type The type of branch, e.g., local or remote.
+   * @param ref The connical ref of the branch
    */
   public constructor(
     public readonly name: string,
     public readonly upstream: string | null,
     public readonly tip: IBranchTip,
-    public readonly type: BranchType
+    public readonly type: BranchType,
+    public readonly ref: string
   ) {}
 
   /** The name of the upstream's remote. */
--- a/app/src/models/branch.ts
+++ b/app/src/models/branch.ts
@@ -70,6 +70,20 @@ export class Branch {
     return pieces[1]
   }
 
+  /** The name of remote for a remote branch. If local, will return null. */
+  public get remoteName(): string | null {
+    if (this.type === BranchType.Local) {
+      return null
+    }
+    const remoteRefPrefix = 'refs/remotes/'
+    const remoteBranchName = this.ref.replace(remoteRefPrefix, '')
+    const pieces = remoteBranchName.match(/(.*?)\/.*/)
+    if (!pieces || pieces.length < 2) {
+      return null
+    }
+    return pieces[1]
+  }
+
   /**
    * The name of the branch's upstream without the remote prefix.
    */
--- a/app/src/models/branch.ts
+++ b/app/src/models/branch.ts
@@ -75,11 +75,12 @@ export class Branch {
     if (this.type === BranchType.Local) {
       return null
     }
-    const remoteRefPrefix = 'refs/remotes/'
-    const remoteBranchName = this.ref.replace(remoteRefPrefix, '')
-    const pieces = remoteBranchName.match(/(.*?)\/.*/)
+
+    const pieces = this.ref.match(/^refs\/remotes\/(.*?)\/.*/)
     if (!pieces || pieces.length < 2) {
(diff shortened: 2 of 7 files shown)


app/src/models/branch.ts
```diff
@@ -45,12 +45,14 @@ export class Branch {
    * @param upstream The remote-prefixed upstream name. E.g., `origin/main`.
    * @param tip Basic information (sha and author) of the latest commit on the branch.
    * @param type The type of branch, e.g., local or remote.
+   * @param ref The connical ref of the branch
```
sergiou87: Typo:
```suggestion
   * @param ref The canonical ref of the branch
```

app/src/lib/stores/git-store.ts
```diff
@@ -1110,7 +1110,8 @@ export class GitStore extends BaseStore {
           currentBranch,
           status.currentUpstreamBranch || null,
           branchTipCommit,
-          BranchType.Local
+          BranchType.Local,
+          ''
```
sergiou87: **If** we accept it's valid not having a proper `ref`, I'd say we state that clearly by making `ref` be of type `string | null`.
Therefore this should be…
```suggestion
          null
```

And we should mention that in `ref`s documentation.

app/src/models/branch.ts
```diff
@@ -45,12 +45,14 @@ export class Branch {
...
    */
   public constructor(
     public readonly name: string,
     public readonly upstream: string | null,
     public readonly tip: IBranchTip,
-    public readonly type: BranchType
+    public readonly type: BranchType,
+    public readonly ref: string
```
sergiou87, line 55: Per my suggestion above:
> **If** we accept it's valid not having a proper `ref`, I'd say we state that clearly by making `ref` be of type `string | null`.
```suggestion
    public readonly ref: string | null
```

app/src/models/branch.ts
```diff
@@ -68,6 +70,20 @@ export class Branch {
...
+  /** The name of remote for a remote branch. If local, will return null. */
+  public get remoteName(): string | null {
+    if (this.type === BranchType.Local) {
+      return null
+    }
+    const remoteRefPrefix = 'refs/remotes/'
+    const remoteBranchName = this.ref.replace(remoteRefPrefix, '')
+    const pieces = remoteBranchName.match(/(.*?)\/.*/)
+    if (!pieces || pieces.length < 2) {
+      return null
```
sergiou87: I'm tempted to say this should throw an error instead of returning `null`. If it's a remote branch, and it has a `ref` (which it should), it should conform to the expected format (`/refs/remotes/<remote-name>/<branch-name>`), otherwise bad things could happen so we better abort the operation and crash / display an error.
```suggestion
      throw new Error(`Remote branch ref has unexpected format: ${this.ref}`)
```

app/src/models/branch.ts
```diff
@@ -68,6 +70,20 @@ export class Branch {
...
   }
 
+  /** The name of remote for a remote branch. If local, will return null. */
+  public get remoteName(): string | null {
+    if (this.type === BranchType.Local) {
+      return null
+    }
+    const remoteRefPrefix = 'refs/remotes/'
+    const remoteBranchName = this.ref.replace(remoteRefPrefix, '')
+    const pieces = remoteBranchName.match(/(.*?)\/.*/)
```
sergiou87: I think this could be safer:
```suggestion
    const pieces = this.ref.match(/^refs\/remotes\/(.*?)\/.*/)
```
It makes sure the ref _actually has_ `refs/remotes` as a prefix (in your `replace`, nothing guaranteed the remote prefix was present nor that it was at the beginning).

app/src/models/branch.ts
```diff
@@ -68,6 +70,20 @@ export class Branch {
...
+    if (!pieces || pieces.length < 2) {
```
sergiou87: I think we're expecting the length to be exactly 2, right? I'd probably state that clearly too:
```suggestion
    if (!pieces || pieces.length === 2) {
```

app/test/unit/create-branch-test.ts
```diff
@@ -29,6 +29,8 @@ const defaultBranch: Branch = {
   upstreamWithoutRemote: null,
   nameWithoutRemote: 'my-default-branch',
   isDesktopForkRemoteBranch: false,
+  ref: '',
```
sergiou87, line 32: If you agree with my suggestion above:
```suggestion
  ref: null,
```

app/test/unit/create-branch-test.ts
```diff
@@ -42,6 +44,8 @@ const someOtherBranch: Branch = {
   upstreamWithoutRemote: null,
   nameWithoutRemote: 'some-other-branch',
   isDesktopForkRemoteBranch: false,
+  ref: '',
```
sergiou87, line 47: Same: if you agree with my suggestion above:
```suggestion
  ref: null,
```

app/test/unit/create-branch-test.ts
```diff
@@ -29,6 +29,8 @@ const defaultBranch: Branch = {
   upstreamWithoutRemote: null,
   nameWithoutRemote: 'my-default-branch',
   isDesktopForkRemoteBranch: false,
+  ref: '',
+  remoteName: '',
```
sergiou87, line 33: Shouldn't this be deleted? 🤔 
```suggestion
```

app/test/unit/create-branch-test.ts
```diff
@@ -42,6 +44,8 @@ const someOtherBranch: Branch = {
   upstreamWithoutRemote: null,
   nameWithoutRemote: 'some-other-branch',
   isDesktopForkRemoteBranch: false,
+  ref: '',
+  remoteName: '',
```
sergiou87, line 48: Same: shouldn't this be deleted? 🤔 
```suggestion
```

app/test/unit/find-forked-remotes-to-prune-test.ts
```diff
@@ -42,7 +42,7 @@ function createSampleBranch(name: string, upstream: string | null) {
     author,
   }
 
-  return new Branch(name, upstream, branchTip, BranchType.Local)
+  return new Branch(name, upstream, branchTip, BranchType.Local, '')
```
sergiou87, line 45: Same: if you agree with my suggestion above:
```suggestion
  return new Branch(name, upstream, branchTip, BranchType.Local, null)
```

sergiou87: Check the rest of the changes, because there would be more `null`s to add and `remoteName: '',` to remove 😅

sergiou87 (changes requested):
I have some concerns about having non-valid `ref` values (either `''` or `null`) 🤔 

Maybe it makes sense to just add it manually where it's missing. Specially in `loadStatus`, since it's the only part of your changes that affects the app: `refs/heads/${currentBranch}`

This way we make sure the branch info is _always_ consistent.

Thoughts?

app/src/models/branch.ts
tidy-dev: Makes sense

app/test/unit/create-branch-test.ts
tidy-dev: Decided to leave as string and as such want ref to be an empty string on unit tests to not add something that appears meaningful.

app/test/unit/create-branch-test.ts
tidy-dev: Typing gets mad if it doesn't exist.

app/test/unit/create-branch-test.ts
tidy-dev: Same

app/src/lib/stores/git-store.ts
tidy-dev: Updated here to have a valid ref

app/test/unit/create-branch-test.ts
sergiou87: 🤯

sergiou87 (approved):
Looks great, thank you for addressing my feedback! 🙇‍♂️ :shipit:

PR #11510, opened 2021-02-01
Implement a TCP-based trampoline mechanism

sergiou87:
## Description

This PR adds support for a replacement of our current `GIT_ASKPASS` trampoline script, that not only is a bit slow, but also [causes some issues on Windows](https://github.com/desktop/desktop/blob/development/docs/known-issues.md#ask-pass-trampolinebat-errors).

With this new trampoline, we get replace the old script with:
1. A generic "data forwarder" that is invoked by git, takes the command line arguments passed by git and the relevant environment variables and sends them to the GitHub Desktop app. This is the [trampoline client](https://github.com/desktop/desktop-trampoline).
2. Specific "command handlers" that run inside and are part of the GitHub Desktop app. These will compute an output that will be sent back to the trampoline client, which will print it in `stdout` for git to consume it.

image

### Trampoline server

Regarding the server side of this trampoline mechanism:
- Starts listening to incoming TCP connections lazily. This not only makes sure the server is not running until it's needed, but it also makes easier handling connection errors: if, for some reason, the server fails and shuts itself down, it will be automagically restarted the next time it's needed.
Right now, "it's needed" means whenever the user performs a remote git operation that needs authentication (i.e. that needs `GIT_ASKPASS`).
- The port used to listen for these TCP connections is chosen by the OS. There is no hardcoded port to prevent running into `EADDRINUSE` errors when the port is already in use.
- In order to increase security, the server side will generate short-lived tokens that will be used to make sure no third party binary is trying to impersonate the trampoline. These tokens will be passed via environment variable (`DESKTOP_TRAMPOLINE_TOKEN`) to the trampoline client, which will have to send it back to the server along with the rest of the info. The server will verify the token, compute the command output, and then revoke the token so it can't be reused.

### Trampoline client

The [trampoline client](https://github.com/desktop/desktop-trampoline) basically receives the server port in a `DESKTOP_PORT` environment variable, connects to that TCP port, and then sends the following (in the same order):
1. The number of command line arguments (`argc - 1`).
2. The command line arguments (`argv`, except `argv[0]` which is the program name).
3. The number of [valid](https://github.com/desktop/desktop-trampoline/blob/main/src/desktop-trampoline.c#L21) environment variables to send.
4. The (valid) environment variables found in the previous step.

After that, the client will read all the data coming from the server until the connection is closed. Whatever is read at this point will be printed in `stdout`, so git can consume it.

For a more exhaustive review of this work, I'd recommend to take a look at the specific desktop-trampoline implementation, more specifically the files in https://github.com/desktop/desktop-trampoline/tree/main/src

## Release notes

Notes: [Improved] Remote Git operations are now faster and less prone to errors on Windows.


app/src/lib/git/branch.ts
```diff
@@ -100,10 +98,18 @@ export async function deleteBranch(
 
     // If the user is not authenticated, the push is going to fail
     // Let this propagate and leave it to the caller to handle
-    const result = await git(args, repository.path, 'deleteRemoteBranch', {
-      env: await envForRemoteOperation(account, remoteUrl),
-      expectedErrors: new Set<DugiteError>([DugiteError.BranchDeletionFailed]),
-    })
+    const result = await withTrampolineEnvForRemoteOperation(
```
sergiou87: All these files under `app/src/lib/git` just change a bit because of a design decision: just like we do in other parts of the app (`withPushPullFetch` or `withIsCommitting`), `withTrampolineEnvForRemoteOperation` will wrap a git operation that might need to use the trampoline. The purpose of `withTrampolineEnvForRemoteOperation` is to generate a trampoline token before invoking git, and revoking it right after that.

app/src/lib/git/checkout.ts
```diff
@@ -105,7 +100,17 @@ export async function checkoutBranch(
...
   )
 
-  await git(args, repository.path, 'checkoutBranch', opts)
+  await withTrampolineEnvForRemoteOperation(
+    account,
+    getFallbackUrlForProxyResolve(account, repository),
+    env => {
+      return git(args, repository.path, 'checkoutBranch', {
+        ...opts,
+        env: merge(opts.env, env),
```
sergiou87, line 109: Commands that also have a `progressCallback` need this additional step of merging the `env` provided by `withTrampolineEnvForRemoteOperation` and the one coming from that progress callback.

app/src/lib/trampoline/trampoline-askpass-handler.ts
```diff
@@ -0,0 +1,29 @@
...
+      return undefined
+    }
+
+    const key = getKeyForEndpoint(endpoint)
+    const token = await TokenStore.getItem(key, username)
+    return token ?? undefined
+  }
+
+  return undefined
+}
```
sergiou87, line 29: The logic in this file is basically the same we have in https://github.com/desktop/desktop/blob/development/app/src/ask-pass/ask-pass.ts

script/build.ts
```diff
@@ -326,6 +326,23 @@ function copyDependencies() {
...
+  fs.removeSync(desktopTrampolineDir)
+  fs.mkdirSync(desktopTrampolineDir)
+  fs.copySync(
+    path.resolve(
+      projectRoot,
+      'app/node_modules/desktop-trampoline/build/Release',
+      desktopTrampolineFile
+    ),
+    path.resolve(desktopTrampolineDir, desktopTrampolineFile)
+  )
```
sergiou87, line 344: This snippet here makes sure the desktop-trampoline binary is copied from the `node_modules` folder to the app bundle.

app/src/lib/trampoline/trampoline-server.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    this.listeningPromise = new Promise((resolve, reject) => {
+      // Observe errors while trying to start the server
+      this.server.on('error', error => {
+        reject(error)
+        this.close()
+      })
+
+      this.server.listen(0, '127.0.0.1', async () => {
```
mislav, line 66: Will the TCP server be accessible to other computers on the network?

script/build.ts
mislav: Will the exe binary need to be signed on Windows in order to be operational?

app/src/lib/trampoline/trampoline-server.ts
sergiou87: We definitely don't want it to be, and that `'127.0.0.1'` should restrict it to connections in the same machine (or that's my hope, at least 😂 ).

script/build.ts
sergiou87: Not sure if it's necessary, but definitely something we want (and it's required on macOS). It's in my TODO list for tomorrow 😄

tidy-dev (dismissed):
Looks great! ✨

script/build.ts
sergiou87: FYI @mislav I can confirm the new binary is signed for both macOS and Windows without any extra effort 😌 🎉

sergiou87:
@tidy-dev I updated the branch with the latest `development` and also bumped `desktop-trampoline` to v0.9.2 to include some safe guard options: https://github.com/desktop/desktop-trampoline/pull/1

tidy-dev (approved):

app/src/lib/trampoline/trampoline-server.ts
Xyncgas: Let me test and see if it's really keeping the connections to local or if I can use vpn from my company to connect

supertry55 (commented):
Submit feedback approve merging these changes #11510

sogan66 (approved):
Submit feedback approve merging these changes

UpUpnAwaaaay:
ell

UpUpnAwaaaay:
> ``

PR #11511, opened 2021-02-01
Update about modal logo 

ampinsk:
This updates the about modal logo from the old icon to the new icon!

|Before|After|
|---|---|
|Screen Shot 2021-02-01 at 11 30 10 AM|Screen Shot 2021-02-01 at 11 35 26 AM|

Question: This will update for both Windows and Mac now, and in the future, will have the Mac icon, instead of the Windows icon. How do we want to approach this?


billygriffin:
Thanks @ampinsk and @tidy-dev! I love the idea of differentiating by platform and now that Big Sur has been out for some time, I think I'm comfortable shipping the updated About icon for macOS alongside the new dock icon.

sergiou87 (approved):
LGTM! 😍

Issue #1685, opened 2017-05-22
Feature Request: Cherry-Picking commits from another branch


kiuru:
### Description

Would be great if I could cherry-pick commits from another branch. That would be very useful and often used feature in our company.

### Version

**GitHub Desktop version:** 0.5.7
**OS version:** Microsoft Windows [Version 10.0.14393]

### Steps to Reproduce

For example, I would like to select "cherry-pick..." and then it opens window where I could tag specific commits from another branches to my active branch.

And same thing step-by-step:

1. Select "Branch" tab from client's toolbar and select "cherry-pick...".
1. New window opens where I can tag specific commits from another branches to my active branch.

nathanpitman:
Something I have requested many times for the old desktop client so adding my vote here to see this become a reality in the new GitHub Desktop. 👍

kiuru:
@nathanpitman Do you have any good advice how to reproduce this new feature to project contributors? Is my example good enough or will come to mind some better one? I thought that was the best one.

nathanpitman:
Here's how it worked in GitBox (my previous fave OS X Git client):

![screen shot 2017-05-24 at 10 56 13](https://cloud.githubusercontent.com/assets/694545/26397909/b74405ca-406f-11e7-972f-9b54ea21f039.png)

1. Pick branch you want to cherry pick from on the right in the title bar (development).
2. Commits which are not in the branch on the left (master) are denoted with a blue marker in the commit list
3. Right click on a commit with a blue marker and then choose 'Cherry-pick' from the context menu
4. This commits the 'commit' you cherry picked to the local branch
5. Switch branch on the right back to origin/master and push changes
6. Done

Pretty quick and easy. :)

kiuru:
Sorry for my lack of understanding, but when I choose "origin/development" on the right in the title bar, does it means commits on below (in the middle of screen) will update according to this branch?

That function is not available at this moment on GitHub Desktop client. Is there anything else good reasons than cherry-pick why they should also add this function to client?

I mean I trying to figure out the most easily way to implement cherry-pick to the current version of client. :)

nathanpitman:
I don't really care how it's implemented, so long as there is a low friction way to cherry pick a commit and merge it into another branch. :)

loukia:
As an feature in UI, this could also be a improvement on current feature "Branch / Merge into current branch".

[...]

PR #11549, opened 2021-02-09
Cherry pick git no confict resolution

tidy-dev:
Part of work on #1685
## Description

Adds basic cherry pick git logic without conflict resolution

Notes: no-notes

### DIFF
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -1,0 +1,47 @@
+import { Repository } from '../../models/repository'
+import { git, IGitResult } from './core'
+
+/** The app-specific results from attempting to cherry pick commits*/
+export enum CherryPickResult {
+  /**
+   * Git completed the cherry pick without reporting any errors, and the caller can
+   * signal success to the user.
+   */
+  CompletedWithoutError = 'CompletedWithoutError',
+
+  /**
+   * An unexpected error as part of the cherry pick flow was caught and handled.
+   *
+   * Check the logs to find the relevant Git details.
+   */
+  Error = 'Error',
+}
+
+/**
+ * A stub function to initiate cherry picking in the app.
+ *
+ * @param revisionRange - this could be a single commit sha or could be a range
+ * of commits like sha1..sha2 or inclusively sha1^..sha2
+ */
+export async function cherryPick(
+  repository: Repository,
+  revisionRange: string
+): Promise<CherryPickResult> {
+  const result = await git(
+    ['cherry-pick', revisionRange],
+    repository.path,
+    'cherry pick'
+  )
+
+  return parseCherryPickResult(result)
+}
+
+function parseCherryPickResult(result: IGitResult): CherryPickResult {
+  if (result.exitCode === 0) {
+    return CherryPickResult.CompletedWithoutError
+  }
+
+  // TODO: handle known exceptions
+
+  throw new Error(`Unhandled result found: '${JSON.stringify(result)}'`)
+}
--- a/app/test/helpers/repository-builder-cherry-pick-test.ts
+++ b/app/test/helpers/repository-builder-cherry-pick-test.ts
@@ -1,0 +1,54 @@
+import { Repository } from '../../src/models/repository'
+import { setupEmptyRepository } from './repositories'
+import { makeCommit, switchTo, createBranch } from './repository-scaffolding'
+
+/**
+ * Creates a test repository to be used as the branch for testing rebase
+ * behaviour with:
+ *  - two commits on default branch,
+ *  - one commit on `featureBranchName`, which has a commit on in addition to default to cherry pick
+ *  - one commit on `targetBranchName`, which is also based on default branch
+ */
+export async function createRepository(
+  featureBranchName: string,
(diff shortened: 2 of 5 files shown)


app/src/lib/git/cherry-pick.ts
```diff
@@ -0,0 +1,47 @@
...
+  const result = await git(
+    ['cherry-pick', revisionRange],
+    repository.path,
+    'cherry pick'
+  )
+
+  return parseCherryPickResult(result)
+}
+
+function parseCherryPickResult(result: IGitResult): CherryPickResult {
```
sergiou87, line 39: I'm sure you already know, but just in case: you can probably use one of the new parsers introduced in #11486

app/test/helpers/repository-builder-cherry-pick-test.ts
```diff
@@ -0,0 +1,45 @@
+import { Repository } from '../../src/models/repository'
+import { setupEmptyRepositoryDefaultMain } from './repositories'
+import { makeCommit, switchTo, createBranch } from './repository-scaffolding'
+
+/**
+ * Creates a test repository to be used for testing cherry pick behaviour with:
+ *  - one commits on default branch,
```
sergiou87: Nit: typo
```suggestion
 *  - one commit on default branch,
```

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -0,0 +1,250 @@
...
+      featureBranch.tip.sha
+    )
+
+    const commits = await getCommits(repository, targetBranch.ref, 3)
+    expect(commits.length).toBe(2)
+    expect(commits[0].summary).toBe(cherryPickedCommit!.summary)
+    expect(result).toBe(CherryPickResult.CompletedWithoutError)
+  })
+
+  it('successfully cherry pick a commit with empty message', async () => {
```
sergiou87, line 59: Is this correct? I just learned cherry-picking commits with empty messages was a special scenario, so I looked into the documentation and seems like we need a `--allow-empty-message` parameter when calling `git cherry-pick`: https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt---allow-empty-message

However, that parameter is not passed to `git cherry-pick` in this PR. Am I missing anything? 😳

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -0,0 +1,250 @@
...
+          contents: '# HELLO WORLD! \nTHINGS GO HERE\n',
+        },
+      ],
+    }
+    await makeCommit(repository, featureBranchCommitThree)
+
+    featureBranch = await getBranchOrError(repository, featureBranchName)
+    await switchTo(repository, targetBranchName)
+
+    const commitRange = `${firstCommitSha}^..${featureBranch.tip.sha}`
```
sergiou87: Not asking you to do anything here, but I'm just wondering if we should add a function for this like `revRange` and `revSymmetricDifference` that we have in `app/src/lib/git/rev-list.ts`.

That specific notation is only used in this file so 🤷‍♂️ I'll leave it up to you 😄

sergiou87 (commented):
Great work!! I love those tests 😍 

I only have a nit that you can ignore, a suggestion that you can ignore (😂 ) and a question about one of your tests.

app/src/lib/git/cherry-pick.ts
tidy-dev: Thanks for the reminder! I  believe that may be more prudent when I add the progress callback in - there is a bit more parsing on that - it pulls a few more attributes out to monitor progress. All this method is responsible for is mapping the dugite/git exit/error code to a cherry pick result (following the pattern established in rebase) to allow for the app store to translate that into a user friendly response. But, I will keep it in mind in case there is better way to do both now :)

app/test/unit/git/cherry-pick-test.ts
tidy-dev: Nope. I was also a little perplexed by that. (I actually still have it in my write up to the contrary) I originally had it grouped with the failures. But, I confirmed that this is adding a commit with an empty message and it does in fact cherry pick it. Thus, I am not sure why the documentation indicates that it would fail.

app/test/unit/git/cherry-pick-test.ts
tidy-dev: :D - I was going to look to see if we had something like this when I went to implement range cherry picking ui.  I will go ahead and add it since I will need for that later PR anyways.

app/test/unit/git/cherry-pick-test.ts
tidy-dev: I just added a couple more lines to the test to verify before cherry picking that the commit I am cherry picking does have an empty message. 🤷‍♀️

sergiou87: Hmmm maybe the behavior changed in a more recent version of git 🤔
Anyway, that additional check you just added is 💯 , thank you!

sergiou87 (approved):
LGTM, thanks for those changes! 😍 :shipit:

PR #11559, opened 2021-02-10
Refactor getCommitsInRange to more generically accept a range parameter

tidy-dev:
Part of work on #1685
## Description
Refactor getCommitsInRange to accept a revisionRange as the parameter instead of two commits so I can send an inclusive range in for cherry picking.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/rev-list.ts
+++ b/app/src/lib/git/rev-list.ts
@@ -119,13 +119,25 @@ export async function getBranchAheadBehind(
  * Returns `null` when the rebase is not possible to perform, because of a
  * missing commit ID
  */
-export async function getCommitsInRange(
+export async function getCommitsBetweenCommits(
   repository: Repository,
   baseBranchSha: string,
   targetBranchSha: string
 ): Promise<ReadonlyArray<CommitOneLine> | null> {
   const range = revRange(baseBranchSha, targetBranchSha)
 
+  return getCommitsInRange(repository, range)
+}
+
+/**
+ * Get a list of commits inside the provided range.
+ *
+ * Returns `null` when it is not possible to perform because of a bad range.
+ */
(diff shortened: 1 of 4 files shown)


app/src/lib/git/rev-list.ts
```diff
@@ -119,13 +119,25 @@ export async function getBranchAheadBehind(
  * Returns `null` when the rebase is not possible to perform, because of a
  * missing commit ID
  */
-export async function getCommitsInRange(
+export async function getCommitsBetweenCommits(
```
sergiou87, line 123: Given there are two methods to retrieve commits, maybe it makes sense to add a note to the comment above to specify that `getCommitsBetweenCommits` doesn't include the "from" commit. What do you think?

sergiou87 (commented):
LGTM! Only a question to improve the old (now renamed) function documentation.

app/src/lib/git/rev-list.ts
tidy-dev: This is the `getCommitsBetweenCommits`  comment
>* Get a list of commits from the target branch that do not exist on the base
>* branch, ordered how they will be applied to the base branch. `

The "from" parameter is the `baseBranchSha`. The comment does seem to cover that it only includes commits on the target branch that don't exist on the base branch. This is the same as saying it would not have the commit referenced by the `baseBranchSha` as it would be on the base branch.

tidy-dev: Added a line in addition 
> Therefore, this will not include the baseBranchSha commit.

To hopefully be more clear.

sergiou87 (approved):
Looks good, thanks for that change!! ❤️ :shipit:

Issue #610, opened 2016-11-08
Improper commit attribution


joshaber:
Problem statements:

1. If the email in my Git config doesn't match any emails on my account in GitHub.com, commits I push are not properly attributed to my GitHub account.

2. Commits in my repository history are not attributed to a GitHub account, so I don't have confidence or peace of mind about who actually made the commit.

3. As a beginner to Git and GitHub, I don't understand why they're different or why my Git info might not match my GitHub info, and why it matters. This is just confusing. (from #4002)

4. If I sign out of my account and back in with a different GitHub account on the same machine, I may see an unexpected avatar based on my underlying Git email that doesn't match the GitHub account I'm signed in with (from https://github.com/desktop/desktop/issues/6973)

5. I want to commit in some repositories with my personal email, and in other repositories with my work email (from #9218). This is a slightly different version of (4) I think.

Considerations:

- Unexpected mismatches most frequently occur when people have a work email on their GitHub account and a personal email in their Git config, or visa versa.
- This can be solved by adding additional emails to your GitHub.com account to ensure that whatever is in your Git config matches one of them.
- We don't seem to currently update any changes from the remote in how we're rendering accounts locally. See: https://github.com/desktop/desktop/issues/3066

From @niik at https://github.com/desktop/desktop/pull/574#pullrequestreview-7378302:

>I was thinking about this one above-and-beyond thing we could do in the configuration step. If a user enters an email that's not associated with their GitHub account (assuming they've signed it) we could show a friendly warning and maybe a link to either add it or suggest one of the email addresses they have associated.

shiftkey:
#3876 was raised by @LitenUpNet as the user had switched accounts but `user.email` was an email address from the old account, which lead to some confusion. This is a separate flow to the Welcome Wizard, but it'd be nice to have a similar check at that point too.

shiftkey:
Noodling on this idea some more, we could change the commit avatar to indicate there's a problem needing attention:

<img width="383" alt="screen shot 2018-07-25 at 10 37 10 am" src="https://user-images.githubusercontent.com/359239/43204294-df0e02b8-8ff6-11e8-9a6d-e1ceb6e460e2.png">

(Yes, I'm going to continue to abuse ⚠️ until someone arrests me or comes along with better design skills)

Clicking on this could show a dialog which then explains the situation and helps the user resolve this, e.g: "The email address you are using for this GitHub repository isn't associated with your GitHub account. Want to add it to your GitHub account now?"

[...]

PR #11561, opened 2021-02-11
Improve onboarding to suggest emails from the user account

sergiou87:
## Description

This PR is part of #610 work. It introduces a few changes in the onboarding flow, in the "Configure Git" screen, to help the user choosing an email address that will get their commits properly attributed.

Additionally, I fixed a race condition that would replace the user's global git config with the name and email from the GitHub account under some circumstances.

There is only one known issue with these changes: a visual glitch that makes the email dropdown look a bit smaller than the textbox. I haven't seen this effect in other places of the app, but I'll investigate it separately and fix it in a different PR.

### Screenshots

image

image

image

## Release notes

Notes: [Improved] Suggest email addresses from the GitHub account and warn about misattributed commits during the onboarding

### DIFF
--- a/app/src/ui/lib/git-email-not-found-warning.tsx
+++ b/app/src/ui/lib/git-email-not-found-warning.tsx
@@ -1,0 +1,69 @@
+import * as React from 'react'
+import { Account } from '../../models/account'
+import { LinkButton } from './link-button'
+import { getDotComAPIEndpoint } from '../../lib/api'
+
+interface IGitEmailNotFoundWarningProps {
+  /** The account the commit should be attributed to. */
+  readonly accounts: ReadonlyArray<Account>
+
+  /** The email address used in the commit author info. */
+  readonly email: string
+}
+
+/**
+ * A component which just displays a warning to the user if their git config
+ * email doesn't match any of the emails in their GitHub (Enterprise) account.
+ */
+export class GitEmailNotFoundWarning extends React.Component<
+  IGitEmailNotFoundWarningProps
+> {
+  private get accountEmails(): ReadonlyArray<string> {
+    // Merge email addresses from all accounts into an array
+    return this.props.accounts.reduce<ReadonlyArray<string>>(
+      (previousValue, currentValue) => {
+        return previousValue.concat(currentValue.emails.map(e => e.email))
+      },
+      []
+    )
+  }
+
+  public render() {
+    if (this.accountEmails.includes(this.props.email)) {
+      return null
+    }
+
+    return this.props.accounts.length === 1
+      ? this.renderForSingleAccount(this.props.accounts[0])
+      : this.renderForMultipleAccounts()
+  }
+
+  private renderForSingleAccount(account: Account) {
+    const accountType =
+      account.endpoint === getDotComAPIEndpoint()
+        ? 'GitHub'
+        : 'GitHub Enterprise'
+
+    return (
+      <div>
+        ⚠️ This email address doesn't match your {accountType} account, so your
+        commits will be wrongly attributed.{' '}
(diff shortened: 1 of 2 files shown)


app/src/ui/lib/git-email-not-found-warning.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </LinkButton>
+      </div>
+    )
+  }
+
+  private renderForMultipleAccounts() {
+    return (
+      <div>
+        ⚠️ This email address doesn't match either of your GitHub.com nor GitHub
+        Enterprise accounts, so your commits will be wrongly attributed.{' '}
```
sergiou87: > doesn't match either of your GitHub.com nor GitHub Enterprise accounts

Does this sound ok for a native English speaker? 😂

tidy-dev: It doesn't stick out as odd to me :). I think I feel this way of it sounding off whenever I use the word "nor".. but I believe it perfectly correct here. :)

app/src/ui/lib/git-email-not-found-warning.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    return this.props.accounts.length === 1
+      ? this.renderForSingleAccount(this.props.accounts[0])
+      : this.renderForMultipleAccounts()
+  }
+
+  private renderForSingleAccount(account: Account) {
+    const accountType =
+      account.endpoint === getDotComAPIEndpoint()
+        ? 'GitHub'
```
tidy-dev: Below in the `renderForMultipleAccounts()`, you use "GitHub.com" and here you use "GitHub". Just a suggestion, but maybe they should be the same.

app/src/ui/lib/git-email-not-found-warning.tsx
```diff
@@ -0,0 +1,69 @@
...
+        ⚠️ This email address doesn't match your {accountType} account, so your
+        commits will be wrongly attributed.{' '}
+        <LinkButton uri="https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user">
+          Learn more.
+        </LinkButton>
+      </div>
+    )
+  }
+
+  private renderForMultipleAccounts() {
```
tidy-dev: Seems to be a bit of repeated wording here. I wonder it it would be simpler to do something like:
image

tidy-dev (commented):
Just a couple small nit like comments. But, other than that, I pulled it down and tested it out. The wrongly attributed warning is slick! 

I looked very briefly at why the name text box looked bigger than the email dropdown and it would seem there is some welcome screen styles being applied to the textbox input.

image

(If I uncheck the height: auto - then it becomes similarly skinny to the email.. tho I would think the fix should try to make the email similarly thick to the input)

app/src/ui/lib/git-email-not-found-warning.tsx
sergiou87: Good point! I'm not sure what's the correct thing here, because I see we use `GitHub.com` only when it's shown along with `GitHub Enterprise`, and then we use just `GitHub` when it's shown independently 🤔

app/src/ui/lib/git-email-not-found-warning.tsx
sergiou87: Love it, thank you!! 😍

sergiou87:
Ohhh, thanks for looking into that! I'll take note and fix it with the PR with the Preferences screen changes, which also includes an email dropdown (and so far works well).

app/src/ui/lib/git-email-not-found-warning.tsx
tidy-dev: Ah! Well if there is already a convention of doing so then I would think to just leave it as is.

tidy-dev (approved):

PR #11566, opened 2021-02-11
In Preferences -> Git, allow using emails from the signed in accounts

sergiou87:
## Description

The work in this PR is also part of #610. In this case, showing a dropdown to users in Preferences -> Git to suggest using one of the email addresses from their account(s), also giving them the ability to select a custom email address.

In case of using a custom email address, a warning will be displayed if the provided address doesn't match any of the account email addresses, which could lead to misattributed commits.

As a bonus, it also fixes a small visual glitch mentioned in #11561 🎉 

### Screenshots

image
image
image

## Release notes

Notes: [Improved] Suggest email addresses from the GitHub account and warn about misattributed commits from Preferences

### DIFF
--- a/app/src/ui/lib/git-config-user-form.tsx
+++ b/app/src/ui/lib/git-config-user-form.tsx
@@ -59,17 +59,6 @@ export class GitConfigUserForm extends React.Component<
   }
 
   public render() {
-    const dotComEmails =
-      this.props.dotComAccount?.emails.map(e => e.email) ?? []
-    const enterpriseEmails =
-      this.props.enterpriseAccount?.emails.map(e => e.email) ?? []
-
-    const shouldShowAccountType =
-      this.props.dotComAccount !== null && this.props.enterpriseAccount !== null
-
-    const dotComSuffix = shouldShowAccountType ? '(GitHub.com)' : ''
-    const enterpriseSuffix = shouldShowAccountType ? '(GitHub Enterprise)' : ''
-
     return (
       <div>
         <Row>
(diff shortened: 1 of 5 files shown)


app/styles/ui/_welcome.scss
```diff
@@ -62,15 +66,21 @@
...
   input {
     padding: var(--spacing-half);
   }
 
+  select {
+    // We need to tweak the horizontal padding a bit here to fit correctly with
+    // the bigger size
+    padding: 0 1px;
+  }
+
```
sergiou87, line 83: The changes in this file are what fix the visual glitch from #11561 

The problem was that the bigger size of text fields and buttons relied entirely on a bigger font size (`--font-size-md` instead of the default `--font-size`) and `auto` height. For _reasons_, just increasing the font size doesn't impact the `select`'s height the same way, so it needs some help 😆 

Since we already have some nice height variables that were used by buttons, inputs and selects, I basically overrode them in the `welcome` element with the value we need, `29px`.

Additionally, the `select` element requires a small tweak in the horizontal padding to make it match the one from the text box.

To be honest, I can't tell which approach is hackier: what we have now or what we had before 😂  Maybe they're both about the same level of hackiness?? 🧙

app/src/ui/lib/git-config-user-form.tsx
```diff
@@ -0,0 +1,180 @@
...
+interface IGitConfigUserFormState {
+  /**
+   * True if the selected email in the dropdown is not one of the suggestions.
+   * It's used to display the "Other" text box that allows the user to
+   * enter a custom email address.
+   */
+  readonly emailIsOther: boolean
+}
+
+export class GitConfigUserForm extends React.Component<
```
sergiou87, line 38: @say25 this component right here + #9499 = 🤤 💯

app/styles/ui/_welcome.scss
tidy-dev: Hackiness is the world of html components and css sometimes... 😂

tidy-dev (approved):
Nice! ✨ ✨ ✨   I like the approach of using an "Other" option in the email dropdown.

Mahmed21:
Mahmed1212

PR #11570, opened 2021-02-12
Fix welcome screen on Windows

sergiou87:
## Description

While testing #11564 on Windows, I noticed something slightly odd: the text in buttons and textboxes in the Welcome screen weren't properly aligned. It was a regression introduced in #11561 where a fixed hardcoded height was set for those elements.

The problem is the specific height chosen, 29px, was only a good fit for macOS but not for Windows, which needs 31.5px.

This PR also bumps `node-abi`, since it includes a few fixes trying to build the app after an Electron/node upgrade. The original issue doesn't affect clean environments (hence why we were able to build the Windows app from CI).

## Release notes

Notes: no-notes

### DIFF
--- a/app/styles/ui/_welcome.scss
+++ b/app/styles/ui/_welcome.scss
@@ -1,7 +1,8 @@
 #welcome {
-  // Override the height of some components to make them bigger
-  --text-field-height: 29px;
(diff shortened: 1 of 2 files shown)


tidy-dev (approved):
✨

PR #11571, opened 2021-02-12
Extract popover to a reusable component

sergiou87:
## Description

This PR is part of #610 work. It extracts the popover we have for the diff options into a reusable component. It'll be used to warn users about the possibility of misattributed commits in a future PR.

In order to make that work easier, it also adds the possibility to position the popover caret in different positions (the 2 we need, so far 🤣 ).

The only thing this refactor might regress is the diff options popover in both History and Changes views. It should behave the same way when clicking inside, outside, on the ⚙️  button, etc.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/popover.tsx
+++ b/app/src/ui/lib/popover.tsx
@@ -1,0 +1,54 @@
+import * as React from 'react'
+import FocusTrap from 'focus-trap-react'
+import { Options as FocusTrapOptions } from 'focus-trap'
+
+interface IPopoverProps {
+  readonly onClickOutside: () => void
+}
+
+export class Popover extends React.Component<IPopoverProps> {
+  private focusTrapOptions: FocusTrapOptions
+  private divRef = React.createRef<HTMLDivElement>()
+
+  public constructor(props: IPopoverProps) {
+    super(props)
+
+    this.focusTrapOptions = {
+      allowOutsideClick: true,
+      escapeDeactivates: true,
+      onDeactivate: this.props.onClickOutside,
+    }
+  }
+
+  public componentDidMount() {
+    document.addEventListener('mousedown', this.onDocumentMouseDown)
+  }
+
+  public componentWillUnmount() {
+    document.removeEventListener('mousedown', this.onDocumentMouseDown)
+  }
+
+  private onDocumentMouseDown = (event: MouseEvent) => {
+    const { current: ref } = this.divRef
+    const { target } = event
+
+    if (
+      ref !== null &&
+      ref.parentElement !== null &&
+      target instanceof Node &&
+      !ref.parentElement.contains(target)
+    ) {
+      this.props.onClickOutside()
+    }
+  }
+
+  public render() {
+    return (
+      <FocusTrap active={true} focusTrapOptions={this.focusTrapOptions}>
+        <div className="popover-component" ref={this.divRef}>
+          {this.props.children}
+        </div>
+      </FocusTrap>
+    )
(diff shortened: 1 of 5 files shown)


app/src/ui/lib/popover.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react'
+import FocusTrap from 'focus-trap-react'
+import { Options as FocusTrapOptions } from 'focus-trap'
+
+/**
+ * Position of the caret relative to the pop up. It's composed by 2 dimensions:
+ * - The first one is the edge on which the caret will rest.
+ * - The second one is the alignment of the caret within that edge.
+ *
+ * Example: TopRight means the caret will be in the top edge, on its right side.
```
tidy-dev, line 10: Very helpful documentation!

tidy-dev: Think it might be useful to add documentation explaining that this will add a class onto the popupover and to add more position options requires adding new classes to the popover.scss?

app/src/ui/lib/popover.tsx
```diff
@@ -0,0 +1,72 @@
...
+  private divRef = React.createRef<HTMLDivElement>()
+
+  public constructor(props: IPopoverProps) {
+    super(props)
+
+    this.focusTrapOptions = {
+      allowOutsideClick: true,
+      escapeDeactivates: true,
+      onDeactivate: this.props.onClickOutside,
+    }
```
tidy-dev, line 37: Popover being a more generic component. What do you think about having focusTrapOptions be an optional input? so later we could have the flexibility to specify one of the other options like 'onActivate'?

app/src/ui/lib/popover.tsx
```diff
@@ -0,0 +1,72 @@
...
+  LeftTop = 'left-top',
+}
+interface IPopoverProps {
+  readonly onClickOutside: () => void
+  readonly caretPosition: PopoverCaretPosition
+}
+
+export class Popover extends React.Component<IPopoverProps> {
+  private focusTrapOptions: FocusTrapOptions
+  private divRef = React.createRef<HTMLDivElement>()
```
tidy-dev: Took me a minute to realize what this was for. Suggest a comment or maybe a name such as focusedDivRef?

tidy-dev (changes requested):
Great lift and shift of reusable functionality! My comments are very minor - mostly food for thought.

app/src/ui/lib/popover.tsx
sergiou87: I considered that, yeah, but I didn't add it to this specific comment block because that's public and the class thing is an implementation detail. But I think it's worth commenting it somewhere/somehow regardless!

app/src/ui/lib/popover.tsx
sergiou87: That sounds good, although I usually prefer not adding stuff until we need it (unless we know we're gonna need it 100%). An example of that are the caret positions: I only added the two we know we need (one for the diff options, already in use, and the other one for the new popover I'm working on). We keep the API simpler and we prevent adding complexity before we need it.

In the future, when (if) we need it, we can add that option.

At least that's how I see it 😅  Does that sound ok to you?

app/src/ui/lib/popover.tsx
sergiou87: Yes, really good point!

tidy-dev (approved):

PR #11572, opened 2021-02-12
Cherry pick resolve conflicts

tidy-dev:
Part of work on #1685
## Description
Cherry picking git conflict resolution logic and bump to dugite 1.98.0 (which has cherry picking unresolved conflict error regex update)

## Release notes

Notes: no-notes


app/src/lib/git/cherry-pick.ts
```diff
@@ -8,7 +17,22 @@ export enum CherryPickResult {
    * signal success to the user.
    */
   CompletedWithoutError = 'CompletedWithoutError',
-
+  /**
+   * The cherry pick encountered conflicts while attempting to cherry pick and
+   * need to be resolved by the user can continue.
```
sergiou87: Nit: I think this is what you meant? 😳 
```suggestion
   * need to be resolved before the user can continue.
```

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -264,4 +271,126 @@ describe('git/cherry-pick', () => {
...
+          {
+            path: 'THING.md',
+            contents: '# HELLO WORLD! \n CREATING CONFLICT! FUN TIMES!\n',
+          },
+        ],
+      }
+      await makeCommit(repository, conflictingCommit)
+    })
+
+    it('successfully detect cherry pick with conflicts', async () => {
```
sergiou87: Nit: This is like the final boss of nits, I'm super super sorry 😳   But in theory these verbs should be in 3rd person because `it` is supposed to be the subject in the sentence:
```suggestion
    it('successfully detects cherry pick with conflicts', async () => {
```

The same applies to the rest of test cases.

Feel free to ignore my comment for this PR, but I think it'd be nice to follow that convention like we do in other test cases 😳

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -264,4 +271,126 @@ describe('git/cherry-pick', () => {
     }
     expect(result).toBe(null)
   })
+
+  describe('cherry picking with conflicts', () => {
```
sergiou87, line 275: TIL you can nest `describe`s 🤯

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -264,4 +271,126 @@ describe('git/cherry-pick', () => {
...
+    beforeEach(async () => {
+      // In the 'git/cherry-pick' `beforeEach`, we call `createRepository` which
+      // adds a commit to the feature branch with a file called THING.md. In
+      // order to make a conflict, we will add the same file to the target
+      // branch.
```
sergiou87, line 280: Thanks for this comment!! 😍

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -264,4 +271,126 @@ describe('git/cherry-pick', () => {
...
+      const conflictingCommit = {
+        commitMessage: 'Conflicting Commit!',
+        entries: [
+          {
+            path: 'THING.md',
+            contents: '# HELLO WORLD! \n CREATING CONFLICT! FUN TIMES!\n',
```
sergiou87, line 286: > CREATING CONFLICT! FUN TIMES!

Some men just want to watch the world burn

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -264,4 +271,126 @@ describe('git/cherry-pick', () => {
...
+        Path.join(repository.path, 'THING.md'),
+        '# HELLO WORLD! \nTHINGS GO HERE\nFEATURE BRANCH UNDERWAY\n'
+      )
+
+      const diffCheckAfter = await GitProcess.exec(
+        ['diff', '--check'],
+        repository.path
+      )
+
+      expect(diffCheckAfter.exitCode).toEqual(0)
```
sergiou87: Just curious: what is this `git diff --check` for? Should we add a comment to explain _why_ this was added? 🤔

app/src/lib/git/cherry-pick.ts
```diff
@@ -41,7 +70,100 @@ function parseCherryPickResult(result: IGitResult): CherryPickResult {
...
+
+/**
+ * Proceed with the current cherry pick operation and report back on whether it completed
+ *
+ * It is expected that the index has staged files which are cleanly cherry
+ * picked onto the base branch, and the remaining unstaged files are those which
+ * need manual resolution or were changed by the user to address inline
+ * conflicts.
+ *
+ */
```
sergiou87, line 93: Could you add a `@param` explaining what `files` is exactly? 😳

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -264,4 +271,126 @@ describe('git/cherry-pick', () => {
...
+      expect(files.length).toBe(2)
+
+      result = await continueCherryPick(repository, files)
+      expect(result).toBe(CherryPickResult.CompletedWithoutError)
+
+      // Only UNRELATED_FILE.md should be in working directory
+      // THING.md committed with cherry pick
+      const status = await getStatusOrThrow(repository)
+      expect(status.workingDirectory.files[0].path).toBe('UNRELATED_FILE.md')
+    })
```
sergiou87: Instead of saying "files unrelated to conflicts", maybe we should be more specific and say untracked files?

I ask because you could change an existing tracked file, and that one would be included in the cherry pick, right? What this test is checking only affects untracked files.

```suggestion
    it('successfully continue cherry pick with new untracked files', async () => {
      result = await cherryPick(repository, featureBranch.tip.sha)
      expect(result).toBe(CherryPickResult.ConflictsEncountered)

      // resolve conflicts by writing files to disk
      await FSE.writeFile(
        Path.join(repository.path, 'THING.md'),
        '# HELLO WORLD! \nTHINGS GO HERE\nFEATURE BRANCH UNDERWAY\n'
      )

      // add new untracked file
      await FSE.writeFile(
        Path.join(repository.path, 'UNTRACKED_FILE.md'),
        '# HELLO WORLD! \nUNTRACKED FILE STUFF IN HERE\n'
      )

      const statusAfterCherryPick = await getStatusOrThrow(repository)
      const { files } = statusAfterCherryPick.workingDirectory

      // THING.MD and UNTRACKED_FILE.md should be in working directory
      expect(files.length).toBe(2)

      result = await continueCherryPick(repository, files)
      expect(result).toBe(CherryPickResult.CompletedWithoutError)

      // Only UNTRACKED_FILE.md should be in working directory
      // THING.md committed with cherry pick
      const status = await getStatusOrThrow(repository)
      expect(status.workingDirectory.files[0].path).toBe('UNTRACKED_FILE.md')
    })
```

sergiou87 (changes requested):
I only found a few things worth commenting. Other than that, it looks good to me, awesome job!! I love those test cases 😍

app/test/unit/git/cherry-pick-test.ts
tidy-dev: Wow.. I don't think I have seen anyone pose it like that... but it just makes perfect sense. 😂

tidy-dev: Done!

app/test/unit/git/cherry-pick-test.ts
tidy-dev: It is just a way to check if conflicts exist. It was being used in rebase testing for conflicts so I just followed the practice. (After having the same question as you tho so good point adding documentation of it) https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---check

app/test/unit/git/cherry-pick-test.ts
tidy-dev: Makes sense -> The term untracked files is a little foreign to me (still coming up to speed with official git terminology). Thus, I was just using words that I was familiar with. 😄

sergiou87 (approved):
All good now! ✨ 

Thank you for making those changes ❤️ 

🚀 🌔

PR #11591, opened 2021-02-15
Show warning when the user is about to commit with the wrong email

sergiou87:
## Description

The work in this PR is part of #610.

When the user has configured the wrong email address that could result in misattributed commits in the current repository, they will get a warning badge on their avatar:

image

When the user clicks that badge, the app will show a popover (requires #11571 to be merged) describing the problem and suggesting a solution:

image

The user can now choose one of their emails found in the GitHub (Enterprise) account relevant for the current repository:

image

After that, clicking on `Save` will update the `user.email` git config to solve the issue.

image

However, clicking on `Ignore` will close the popup and remove the warning badge for this repository. The email address will remain unchanged.

## Release notes

Notes: [Improved] Suggest emails from GitHub accounts and warn about misattributed commits in the commit message area.

### DIFF
--- a/app/src/ui/changes/commit-message-avatar.tsx
+++ b/app/src/ui/changes/commit-message-avatar.tsx
@@ -70,7 +70,9 @@ export class CommitMessageAvatar extends React.Component<
   private shouldShowMisattributedCommitWarning() {
     if (
       this.props.email === undefined ||
-      this.props.repositoryAccount === null
+      this.props.repositoryAccount === null ||
+      this.props.repository === null ||
+      this.props.repository.ignoreWrongUserEmail
     ) {
       return false
     }
@@ -189,10 +191,18 @@ export class CommitMessageAvatar extends React.Component<
   private onIgnoreClick = (event: React.MouseEvent<HTMLButtonElement>) => {
     event.preventDefault()
     this.closePopover()
+
+    const repository = this.props.repository
+    if (repository === null) {
+      return
+    }
+
+    this.props.dispatcher.updateRepositoryIgnoreWrongUserEmail(repository, true)
   }
 
   private onSaveClick = async (event: React.MouseEvent<HTMLButtonElement>) => {
     event.preventDefault()
+    this.closePopover()
 
     const repository = this.props.repository
     if (repository === null) {
@@ -203,8 +213,6 @@ export class CommitMessageAvatar extends React.Component<
       await setGlobalConfigValue('user.email', this.state.accountEmail)
       this.props.dispatcher.refreshAuthor(repository)
     }
-
-    this.closePopover()
   }
 
   private onSelectedGitHubEmailChange = (
--- a/app/src/ui/changes/commit-message-avatar.tsx
+++ b/app/src/ui/changes/commit-message-avatar.tsx
@@ -15,10 +15,10 @@ import { Dispatcher } from '../dispatcher'
 import { setGlobalConfigValue } from '../../lib/git'
 
 interface ICommitMessageAvatarState {
-  isPopoverOpen: boolean
+  readonly isPopoverOpen: boolean
 
   /** Currently selected account email address. */
-  accountEmail: string
+  readonly accountEmail: string
 }
 
 interface ICommitMessageAvatarProps {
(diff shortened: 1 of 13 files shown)


app/src/ui/changes/commit-message-avatar.tsx
```diff
@@ -0,0 +1,226 @@
...
+      >
+        <h3>This commit will be misattributed</h3>
+        <Row>
+          <div>
+            The email in your{' '}
+            <span className="highlighted-text">git config</span> (
+            {this.props.email}) doesn't match your{' '}
+            <span className="highlighted-text">
+              GitHub{accountTypeSuffix} account
+            </span>
```
sergiou87: I think having these chunks of text with the same color as a link might be misleading. Specially since there is an actual link in the same popover. I picked this text style from the mocks, but maybe we should do something else… 🤔 

What do you think, @ampinsk?

image

tidy-dev: :D - I was going to comment this.. I tried to click them...

app/src/ui/changes/commit-message-avatar.tsx
```diff
@@ -0,0 +1,226 @@
...
+            .{' '}
+            <LinkButton uri="https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user">
+              Learn more.
+            </LinkButton>
+          </div>
+        </Row>
+        <Row>
+          <Select
```
tidy-dev, line 138: image

Looks like we may need some styling on the email. It doesn't like long emails.

tidy-dev (commented):
Great work!

Here are some thoughts that came to mind when reviewing and interacting with it. Not necessarily a call to change if you don't feel the same.

1. My first reaction to the component is that it felt heavy for a small ui display component. Following the smart/dumb component approach, it seems like having an icon have access to the repository and the dispatcher feels very smart for a display component.  My thoughts drifted to wondering if we could just pass emails to it and relay the action of ignore/save to the parent. (Also was wondering if you were encapsulating this in case we used this elsewhere.. in which case it would make sense to have it heavier so we don't have to duplicate parent/smart logic elsewhere)
2. I hit ignore and I didn't realize the button had more to it than just dismissing the popup information... also once I hit ignore I couldn't get back to that.. so if I hit ignore to dismiss instead now I may not know how I can address it later in the settings. Maybe be more explicit? "Ignore for this repository".. ugg long name. Since misattributed commits are something we are trying to prevent.. is it really that bad to always show the warning icon in this scenario? If we are aware of a subset of users who would be particularly annoyed by not being able to ignore it, could that be a settings thing and the other button be "View In Settings"?
3.  I am given a list of emails and a save button. But.. not exactly sure what that means.. especially if I don't want to click learn more.. which rebel me just wants to commit and not go learn stuff.  😄   Would save be clearer as "Update Git Config"? Would it be too much clutter to add a heading to the email dropdown?

app/src/ui/changes/commit-message-avatar.tsx
sergiou87: Ouch! Nice catch 😄 Probably worth truncating the text (assuming the dropdown will show the full content when expanded, but I can try and see how it feels).

sergiou87:
Thank you for your thorough review @tidy-dev 🙇‍♂️ 🙇‍♂️ 🙇‍♂️ ❤️ 

Regarding your comments:

1. I think that's a very good point. I see other components also taking this kind of objects as input and computing the rest. I'll try the approach you suggest and let you know 😄 
2. Yes, that's something I wanted to mention in today's sync meeting. IMHO the behavior I implemented makes sense, but I agree we might need to be more specific with the wording, and also I think we might need a checkbox somewhere (Repository Settings?) to re-enable that popup.
Other than that, if you just want to dismiss the popover (but not clear the warning), the behavior is the same as in the diff options popup: just click outside of the popover. The `Ignore` button is supposed to be used when just don't want to be bothered with the warning again (for that repo, at least).
3. All good questions… I think `Save` was chosen because it was more "git agnostic" and concise. But probably worth asking later today too 😄

sergiou87:
I made the change you suggested in (1) and I think it's much better indeed, thank you! 😍 

I also addressed the issue with long emails by truncating the dropdown contents as I mentioned. When the user clicks on the dropdown and the whole list is displayed, the emails are not truncated.

sergiou87:
@tidy-dev I already implemented also the other changes agreed with the team, it's ready for another review when you have a chance! 😊

tidy-dev (approved):
Nice! ✨  Thank you for all the changes. 🙂

PR #11595, opened 2021-02-15
Allow users to modify git config on a per repository basis

### LINKED ISSUE #9449: Allow Changing Git Config Per Repository

I have multiple git configurations (due to multiple git accounts/configs).
 
I have a global default one of:

```bash
git config --global user.name Steven Yeh
git config --global user.name Steven.Yeh@vauto.com
```

and this account my public one

```bash
git config user.name say25
git config user.email say25@cornell.edu
```

### Proposed solution

Add a tab in the repository setting menu to change user.name and user.email without a command line.

repo-settings

#### Related 
#6534 - `Display better message when adding a repository with malformed .git/config`
#9255 - `2.3.2 beta 4 changed my git name to my github username`

sergiou87:
## Description

The work in this PR is part of #610

Closes #9449

It includes the changes from #9499 with a few fixes and additions:
- Now it uses the new reusable user name/email form component that allows you to choose an email from your account (introduced in #11566)
- When Global config is selected, it **removes** the `user.email` and `user.name` values from the local git config, making sure git will actually use the global values.
- Added a message to the misattributed commit popover suggesting the user to fix the email problem only for that record, linking to the repository settings. Requires #11591 
- The popover mentioned above has been moved to grow towards the top and right sides of the screen (vs bottom and right), because its height is bigger now and might go out of the window bounds.

### Screenshots

image

image

image

## Release notes

Notes: [Added] Ability to modify git config on a per repository basis. Thanks @say25!


say25:
This should close https://github.com/desktop/desktop/issues/9449 if I'm not mistaken

app/src/lib/git/config.ts
```diff
@@ -128,3 +125,104 @@ export async function getGlobalConfigPath(env?: {
...
+    flags.push('--global')
+  }
+
+  flags.push('--replace-all', name, value)
+
+  await git(flags, path || __dirname, 'setConfigValueInPath', options)
+}
+
+/** Remove the local config value by name. */
+export async function removeConfigValue(
```
sergiou87, line 182: To save you jumping through the code, this is used when the user switches from local user info (name and email) to global user info, to delete the local user info.

app/src/models/popup.ts
```diff
@@ -102,7 +103,11 @@ export type Popup =
       repository: Repository
       branch?: Branch
     }
-  | { type: PopupType.RepositorySettings; repository: Repository }
+  | {
+      type: PopupType.RepositorySettings
+      repository: Repository
+      initialSelectedTab?: RepositorySettingsTab
```
sergiou87, line 109: This is used to allow opening the repository settings dialog from a specific tab.

app/src/ui/changes/commit-message-avatar.tsx
```diff
@@ -121,7 +123,7 @@ export class CommitMessageAvatar extends React.Component<
 
     return (
       <Popover
-        caretPosition={PopoverCaretPosition.LeftTop}
+        caretPosition={PopoverCaretPosition.LeftBottom}
```
sergiou87, line 126: I had to do this because the popup was too big and it would go beyond the boundaries of the window, and getting its content truncated.

app/src/ui/changes/commit-message-avatar.tsx
```diff
@@ -146,6 +148,15 @@ export class CommitMessageAvatar extends React.Component<
             ))}
           </Select>
         </Row>
+        <Row>
+          <div className="secondary-text">
+            You can also choose an email only for this repository from the{' '}
+            <LinkButton onClick={this.onRepositorySettingsClick}>
+              repository settings
+            </LinkButton>
+            .
```
sergiou87: Open to wording suggestions 😂  Although Amanda seemed happy with this 👍

app/src/ui/lib/select.tsx
```diff
@@ -11,6 +11,9 @@ interface ISelectProps {
   /** The default value of the select control. */
   readonly defaultValue?: string
 
+  /** Whether or not the select control is disabled. False by default. */
+  readonly disabled?: boolean
```
sergiou87, line 15: Needed to disable de email dropdown when the user chooses the "Use global git config" option.

app/src/ui/repository-settings/repository-settings.tsx
```diff
@@ -18,17 +18,32 @@ import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
...
   readonly repository: Repository
+  readonly repositoryAccount: Account | null
   readonly onDismissed: () => void
 }
 
-enum RepositorySettingsTab {
+export enum RepositorySettingsTab {
   Remote = 0,
   IgnoredFiles,
+  GitConfig,
```
sergiou87, line 46: I added this tab before `ForkSettings` because `ForkSettings` is optional (only shows up for forks). Adding a new non-optional tab after an optional tabs would require us some special treatment of tab indices.

app/styles/ui/_commit-message-avatar.scss
```diff
@@ -41,6 +44,10 @@
     justify-content: flex-end;
   }
 
+  .secondary-text {
+    color: var(--text-secondary-color);
+  }
```
sergiou87, line 49: Suggestion from Amanda to make the new text in the popover less prominent.

app/src/ui/changes/commit-message-avatar.tsx
```diff
@@ -146,6 +148,15 @@ export class CommitMessageAvatar extends React.Component<
             ))}
           </Select>
         </Row>
+        <Row>
+          <div className="secondary-text">
+            You can also choose an email only for this repository from the{' '}
```
tidy-dev:
```suggestion
            You can also choose an email specific to this repository from the{' '}
```

tidy-dev: Don't know why but "only for" feels funny to me.. but could just be me 🙂  - I don't have any strong feelings on this so feel free to ignore the suggestion.

app/styles/ui/_commit-message-avatar.scss
tidy-dev: I like this! ✨  ✨

tidy-dev (commented):
Code and logic looks great! I really liked all the documentation throughout - made it easy to understand.
 
I have some thoughts on  semantics. In the popup, we have
image

We use more git agnostic phrases: 
- "The email in your git config ... doesn't match your"  (global setting)
- "also choose an email only for this repository" (local setting)

However in the repository settings:
image

We use git language
- "global Git config"
- "local Git config" 

These two components say the same thing in different ways... maybe leading a beginner user not to know they are referring to the same things?

I would suggest stepping the git language up in the pop over for internet search-ability as well as tying the two application areas together.
- "The email in your global git config ... doesn't match your"
- "also choose a local email only for this repository"

@ampinsk I would love your input on this? Maybe I am over thinking this 🙂

sergiou87:
I think you make a very good point @tidy-dev, thank you! ❤️ 

IMHO saying `The email in your global git config ... doesn't match your` sounds good, but adding `local` to the other one sounds a bit odd to me and kind of redundant when you already have `only for this repository` (or as you suggested `specific to this repository`).

Also, I noticed I should probably change the popover to say `email in your Git config` instead of `email in your git config` 🤣

tidy-dev:
Other thought -> "also choose an email local to this repository"

billygriffin:
Great suggestions @tidy-dev!

For the reasons @tidy-dev shared and at the risk of slightly less clear language, I prefer:

"The email in your global Git config ... doesn't match your..."

and 

"You can also choose an email local to this repository"

But I'm happy with @ampinsk being the final decision maker if she disagrees. 😄

ampinsk:
I like the language @billygriffin suggested! 👍

say25:
Could just be me but having most of the commits link to 610 just makes the issue a little messy. Looked over the code and it seems that most everything have been covered by the rest of the team.

sergiou87:
@say25 yes, you're right, sorry, I linked all the commits from previous PRs to #610 as they are part of that work, but it's a bit odd doing the same with this one.

tidy-dev (approved):
Thank you for the changes! :) LGTM!

Issue #6674, opened 2019-01-20
Add option - [show in git shell]


AlanCPSC:
In the latest update, I noticed GitHub Desktop now has a variety of convenience buttons. I have found myself frequently using the [show in explorer] function. I was wondering if it is possible to add a similar button [show in git shell] that opens a command prompt in the directory of the repository.

![suggestion](https://user-images.githubusercontent.com/22120384/51434743-b9993b80-1c1c-11e9-80a2-c2cff9fcbefe.png)

shiftkey:
Marking this one as `user-research` because I'm sure there was some thinking around this and why `Open in Shell` might not have made the cut for this first iteration of the UI.

cc @desktop/product

billygriffin:
Thanks for the idea! I would love @niik’s perspective on this as well before we make a call one way or the other. I don’t want this blank slate to be a catch all for anything and everything, but rather to thoughtfully enable workflows that make people more productive.

@alanCPSC and anyone else who is interested in this - Out of curiosity, do you frequently use the existing keyboard shortcut for this in the Repository menu? And at what times?

say25:
I personally feel like encouraging “opening in shell” from the empty state has value. Currently doing so supports rebasing, tagging and some other workflows we currently don’t support. I also find that people that use Desktop aren’t aware of this shortcut. I’m a large proponent of this tool at work and whenever I’m pairing with someone and I open the shell with the shortcut I get a “how did you do that”

billygriffin:
Thanks for the input y'all. We're going to regroup on this a bit more internally before deciding which way to go on this one. Generally, we want to be very intentional about how we use the "No local changes" area to ensure that it best enables the things we think make people most productive. One thing we're doing to validate that is to measure usage of the existing items (that's TBD for me to make issues for, but should happen hopefully today). And then potentially if one of them is not getting much usage at all, we might consider a swap as opposed to an addition.

billygriffin:
Thanks for the issue @AlanCPSC and for sharing your follow up thoughts. Just to add a bit more color, it's not about minimalist design, but rather the idea that everything added is diminishing the importance of everything that was already there. So generally we prefer to be very deliberate in assessing the value proposition of something prior to implementing it. For "open in explorer/finder," the value proposition has been super clear from many of our user interviews that it's often something people are confused about and want an easier way to know how to easily get there.

[...]

PR #11602, opened 2021-02-16
Added "Open the repository in your shell" to no-changes.tsx

SupertigerDev:
## Description
- Adds open in shell option to the default NoChanges screen so it is easier to access.

### Screenshots
screenshot

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1274,7 +1274,7 @@ export class App extends React.Component<IAppProps, IAppState> {
       case PopupType.RenameBranch:
         const stash =
           this.state.selectedState !== null &&
-            this.state.selectedState.type === SelectionType.Repository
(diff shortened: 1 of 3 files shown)


sergiou87:
Hey @supertiger1234, thank you for opening this PR :smile: 

We find this addition very useful, but we are afraid we could end up bloating that screen very easily with more options. There have been some [other discussions about this in the past](https://github.com/desktop/desktop/issues/6674#issuecomment-459905998), and decided we had to draw the line somewhere. Not having this specific action in the No Changes screen also shouldn't be too problematic, since there is a keyboard shortcut for it.

On a more general note, we don't usually accept PRs with features that haven't been discussed in an issue and approved by the team.

I'm very sorry, we really appreciate your contribution and hope to see more of them coming :smile:

SupertigerDev:
How about if I add an option in settings to toggle which options you can show?

PR #11624, opened 2021-02-19
Cherry picking branch dialog

tidy-dev:
Part of work on #1685
## Description

First pass at choose branch dialog for cherry picking flow. It also includes some other general cherry picking setup details for managing the cherry picking state and flow in the app.

2nd pass to come will include preview message.

### Screenshots

image

Notes: no-notes

### DIFF
--- a/app/src/models/cherry-pick.ts
+++ b/app/src/models/cherry-pick.ts
@@ -13,7 +13,7 @@ export interface ICherryPickSnapshot {
 /** Union type representing the possible states of the cherry pick flow */
 export type CherryPickFlowStep = ChooseTargetBranchesStep
 
-export const enum CherryPickStep {
+export const enum CherryPickStepKind {
   /**
    * The initial state of a cherry pick.
    *
@@ -26,7 +26,7 @@ export const enum CherryPickStep {
 
 /** Shape of data needed to choose the base branch for a cherry pick  */
 export type ChooseTargetBranchesStep = {
-  readonly kind: CherryPickStep.ChooseTargetBranch
+  readonly kind: CherryPickStepKind.ChooseTargetBranch
   readonly defaultBranch: Branch | null
   readonly currentBranch: Branch
   readonly allBranches: ReadonlyArray<Branch>
--- a/app/src/models/cherry-pick.ts
+++ b/app/src/models/cherry-pick.ts
@@ -1,3 +1,4 @@
+import { Branch } from './branch'
 import { CommitOneLine } from './commit'
 import { ICherryPickProgress } from './progress'
 
@@ -7,4 +8,27 @@ export interface ICherryPickSnapshot {
   readonly remainingCommits: ReadonlyArray<CommitOneLine>
   /** The progress of the operation */
   readonly progress: ICherryPickProgress
+}
+
+/** Union type representing the possible states of the cherry pick flow */
+export type CherryPickFlowStep = ChooseTargetBranchesStep
+
+export const enum CherryPickStep {
+  /**
+   * The initial state of a cherry pick.
+   *
+   * This is where the user picks which is the target of the cherry pick.
+   * This step will be skipped when cherry pick is initiated through
+   * drag and drop onto a specific branch.
+   */
+  ChooseTargetBranch = 'ChooseTargetBranch',
+}
+
+/** Shape of data needed to choose the base branch for a cherry pick  */
+export type ChooseTargetBranchesStep = {
+  readonly kind: CherryPickStep.ChooseTargetBranch
+  readonly defaultBranch: Branch | null
+  readonly currentBranch: Branch
+  readonly allBranches: ReadonlyArray<Branch>
+  readonly recentBranches: ReadonlyArray<Branch>
 }
--- a/app/src/ui/cherry-pick/choose-target-branch.tsx
+++ b/app/src/ui/cherry-pick/choose-target-branch.tsx
@@ -113,15 +113,9 @@ export class ChooseTargetBranchDialog extends React.Component<
 
     const { selectedBranch } = this.state
     if (selectedBranch !== null) {
-      // the amount of characters to allow before we truncate was chosen
-      // arbitrarily and matches rebase choose branch dialog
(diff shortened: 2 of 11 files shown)


app/src/models/cherry-pick.ts
```diff
@@ -8,3 +9,26 @@ export interface ICherryPickSnapshot {
...
+   * This is where the user picks which is the target of the cherry pick.
+   * This step will be skipped when cherry pick is initiated through
+   * drag and drop onto a specific branch.
+   */
+  ChooseTargetBranch = 'ChooseTargetBranch',
+}
+
+/** Shape of data needed to choose the base branch for a cherry pick  */
+export type ChooseTargetBranchesStep = {
+  readonly kind: CherryPickStep.ChooseTargetBranch
```
sergiou87: Since `CherryPickStep` is used as a "kind" of step, maybe it should be renamed to `CherryPickStepKind` for clarity?
```suggestion
export const enum CherryPickStepKind {
  /**
   * The initial state of a cherry pick.
   *
   * This is where the user picks which is the target of the cherry pick.
   * This step will be skipped when cherry pick is initiated through
   * drag and drop onto a specific branch.
   */
  ChooseTargetBranch = 'ChooseTargetBranch',
}

/** Shape of data needed to choose the base branch for a cherry pick  */
export type ChooseTargetBranchesStep = {
  readonly kind: CherryPickStepKind.ChooseTargetBranch
```

app/src/ui/cherry-pick/choose-target-branch.tsx
```diff
@@ -0,0 +1,173 @@
...
+            cancelButtonVisible={false}
+          />
+        </DialogFooter>
+      </Dialog>
+    )
+  }
+
+  private startCherryPick = async () => {
+    const { selectedBranch } = this.state
+    if (!selectedBranch) {
```
sergiou87: I think it's preferred to use `!== null` explicitly (at least that's what Markus told me a few times 🤣 )
```suggestion
    if (selectedBranch !== null) {
```

app/src/ui/cherry-pick/choose-target-branch.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  private renderOkButtonText() {
+    const okButtonText = 'Cherry pick commit'
+
+    const { selectedBranch } = this.state
+    if (selectedBranch !== null) {
+      return (
+        <>
+          {okButtonText} to <strong>{selectedBranch.name}</strong>…
+        </>
```
sergiou87, line 117: What is this sorcery??? 😂 
I've been wrapping things in `div`s, how is this better?

app/src/ui/cherry-pick/choose-target-branch.tsx
```diff
@@ -0,0 +1,173 @@
...
+  private onEnterPressed = (branch: Branch, source: ClickSource) => {
+    if (source.kind !== 'keyboard' || source.event.key !== 'Enter') {
+      return
+    }
+
+    source.event.preventDefault()
+
+    const { selectedBranch } = this.state
+
+    if (selectedBranch !== null && selectedBranch.name === branch.name) {
```
sergiou87, line 89: This should also check if the selected branch is the current branch, right?
```suggestion
    if (
      selectedBranch !== null &&
      selectedBranch.name === branch.name &&
      this.canCherryPickOntoSelectedBranch()
    ) {
```

sergiou87 (changes requested):
Looking great and I learned a few things, only found one real issue. Awesome job!! ❤️

app/src/models/cherry-pick.ts
tidy-dev: Makes sense. Thanks!

app/src/ui/cherry-pick/choose-target-branch.tsx
tidy-dev: No. Current branch is the branch that is currently checked out that we are cherry picking from. This method is when the user hits enter. It is for the flow such that a user clicks on a "target-branch" to cherry pick to and  then hits enter to start the cherry pick. The branch that is a parameter should be the selected branch; this is just a fail safe in case of a timing issue of user clicks branch A and then clicks branch B at the same time they hit enter. Enter branch would be B but state has not updated from A to B yet. (forcing them to hit enter again when the state has actually updated and thus the ui truly reflects the selected branch)

app/src/ui/cherry-pick/choose-target-branch.tsx
tidy-dev: 😄 I don't know if it is better per say. It is called a fragment. It is specifically for the purpose of needing JSX but not needing any attributes or specific html tag.https://reactjs.org/docs/fragments.html 
 https://medium.com/@yurachoi/react-js-fragments-175290da6435

app/src/ui/cherry-pick/choose-target-branch.tsx
tidy-dev: Absolutely! I blame habit. 😸  Markus called me out on that too. 🙂  Thank you. Actually now that you point it out, I made this method before I made the "canCherryPickOntoSelectedBranch" method which would be a more complete failsafe here.

app/src/ui/cherry-pick/choose-target-branch.tsx
tidy-dev: Benefit is it doesn't add to the DOM unnecessarily - had to think about it for a second to remember why I use it. 🙂

Degesan150 (commented):
____

app/src/ui/cherry-pick/choose-target-branch.tsx
sergiou87: Yes, but still you have to check the user didn't press enter in the current branch before calling `startCherryPick`, right?

I see you have added that check inside of `startCherryPick` which should be 👍 

Thank you!

sergiou87 (approved):
Everything looks good now ✨ ❗ Thank you!! :shipit:

PR #9016, opened 2020-01-31
Add Support for Windows Automatic Dark Mode

say25:
~⚠️Depends on #8967⚠️~

Closes #9015

## Description

- Add automatic switching between Dark Mode and Light Mode on Windows.

### Screenshots

WindowsDarkMode

- Also see [original macOS implementation](https://github.com/desktop/desktop/pull/6051) for details.

## Release notes

Notes: Add Support Automatic Dark Mode Switching on Windows

PR #11629, opened 2021-02-20
Use System Theme as Default

### LINKED ISSUE #9959: Following system theme should be a default
The macOS version of GitHub Desktop apparently has an appearance option (which I'm recently finding out and is the inspiration to this issue) which let's us use a dark theme and an automatic switching feature that follows the system theme along with the other apps. But the default of following the system theme doesn't come enabled by default causing people to not even know about the feature (if you're using the program for a long time and haven't come across with the dark/light option on the introduction page).

The best and most logical solution would be to enable the following of the system theme by default.

Important note: Not sure about the latest behavior of the app (the installation wizard, defaults etc.) and suggesting this to have a discussion on it. Don't hesitate to correct any of my errors, as at the end this is a suggestion to improve the app.

Screen Shot 2020-06-08 at 01 35 54

say25:
Closes #9959

## Description

- Uses System Theme as Default
- Moves to use [nativeTheme](https://www.electronjs.org/docs/api/native-theme#nativetheme) as its been built into Electron [since 7.0.0-beta3](https://www.electronjs.org/docs/api/native-theme/history) (after the original implementation reacting to System Theme changes)
- Renames method from `supportsDarkMode` to `supportsSystemThemeChanges` to reflect reality (we support Dark mode prior to Mojave and Windows 1809, we just don't support reacting to System Theme changes)
- Re-Adds analytics for Windows (missed when https://github.com/desktop/desktop/pull/9016 was implemented)

### Screenshots

image

## Release notes

Notes:
- Uses System theme as default


app/src/ui/lib/dark-theme.ts
```diff
@@ -4,7 +4,10 @@ import {
   isWindows10And1809Preview17666OrLater,
 } from '../../lib/get-os'
 
-export function supportsDarkMode() {
+/**
```
say25: I'm not sure if this file should be renamed to something like `system-theme.ts` as well

say25: part of me wants to merge this file into `app/src/ui/lib/application-theme.ts`

app/src/ui/lib/dark-theme.ts
```diff
@@ -4,7 +4,10 @@ import {
...
+ * Whether or not the current OS supports System Theme Changes
+ */
+export function supportsSystemThemeChanges() {
```
say25: Something of note, this doesn’t have to be here. From what Electron shows we can use system theme in older OS versions but of course setting it as such is effectively the same as Light mode. I think for that reason it might be worth keeping this around. I sorta suspect that Chrome Dev tools have System Theme as a possible setting on older OSes despite the fact that the value would never change. I’m not sure what the older OS would do as I don’t have a machine on an older OS to test this.

daddy62:
rr

app/src/ui/lib/dark-theme.ts
tidy-dev: Agree. It does feel arbitrarily separated from application-theme given that supportsDarkMode just became more generic.

tidy-dev (commented):
Nice refactor! The theme settings look so much nicer as another vertical selection option as opposed to the checkbox. 

As I was testing it out, I noticed that when I toggled light/dark mode on my system preferences that VS Code, Chrome, and my Production Desktop immediately picked up the change. For Desktop running in dev mode, I had to make a selection change in the preferences or restart Desktop. This appears to be a change from production. Are you seeing that behavior (or is it something on my machine)?

app/src/lib/stores/app-store.ts
```diff
@@ -1694,20 +1690,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-        ? ApplicationTheme.Dark
-        : ApplicationTheme.Light
-      setPersistedTheme(this.selectedTheme)
-    } else {
-      this.selectedTheme = getPersistedTheme()
-    }
+    this.selectedTheme = getPersistedTheme()
+    this.automaticallySwitchTheme =
+      this.selectedTheme === ApplicationTheme.System
+    setPersistedTheme(this.selectedTheme)
```
sergiou87: Setting the persisted theme right after reading it, feels a bit odd to me 🤔 

After looking at the implementation I understood why, but maybe it should be documented with a comment.

app/src/lib/stores/app-store.ts
```diff
@@ -5411,23 +5400,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
    */
   public _setSelectedTheme(theme: ApplicationTheme) {
     setPersistedTheme(theme)
+    this.automaticallySwitchTheme = theme === ApplicationTheme.System
```
sergiou87: Maybe this should be inside of `setPersistedTheme`? I think it makes sense 🤔

app/src/ui/lib/dark-theme.ts
sergiou87: Sounds good to me!

app/src/ui/preferences/appearance.tsx
```diff
@@ -27,66 +33,36 @@ const themes: ReadonlyArray<ISegmentedItem<ApplicationTheme>> = [
...
-        isDarkModeEnabled() ? ApplicationTheme.Dark : ApplicationTheme.Light
-      )
+    if (
+      this.props.selectedTheme === ApplicationTheme.System &&
+      !supportsSystemThemeChanges()
+    ) {
+      selectedTheme = isDarkModeEnabled()
+        ? ApplicationTheme.Dark
+        : ApplicationTheme.Light
     }
```
sergiou87: If `supportsSystemThemeChanges` is `false`, then `isDarkModeEnabled` would never be true, right?
```suggestion
    if (
      this.props.selectedTheme === ApplicationTheme.System &&
      !supportsSystemThemeChanges()
    ) {
      selectedTheme = ApplicationTheme.Light
    }
```

sergiou87 (commented):
I haven't tested it yet, but I'll give it a try after the first batch of comments is addressed 😄 

Thanks a lot for your work @say25 !!! ✨

app/src/ui/preferences/appearance.tsx
say25: My understanding is no, even though your system can not emit events for Electron to pick up and automatically change with your system, you are still able to switch from DarkMode to LightMode. The reason for it is the default setting is System now. Users on older systems will need this case.

say25: Edit: Ahh, I see what you mean, I think I didn't test that case (my OS is too new) but effectively we need to just return `remote.nativeTheme.shouldUseDarkColors` no matter what now.

say25:
I think I need to re-visit something, there seems to be a mistake in my implementation

say25:
@sergiou87 @tidy-dev sorry for the delay here. Took a look with fresh eyes and noticed a few issues that I believe are resolved now (basically I need to split `selectedTheme` from `currrentTheme` to account for a few different scenarios).

I'm not sure how you guys feel about f4d393f in particular. Not super thrilled with `ApplicationTheme.Light | ApplicationTheme.Dark` everywhere so I created `ApplicableTheme` I don't have strong feelings either way.

app/src/ui/lib/application-theme.ts
```diff
@@ -1,76 +1,93 @@
...
 /**
  * A set of the user-selectable appearances (aka themes)
  */
 export enum ApplicationTheme {
   Light,
   Dark,
+  System,
 }
 
+export type ApplicableTheme = ApplicationTheme.Light | ApplicationTheme.Dark
```
tidy-dev, line 16: @say25  I really like this! It makes the code much more readable.

tidy-dev (dismissed):
I pulled it down and tested it out. My concern about not automatically picking up the system theme change is addressed and working as it previously was.

app/src/ui/index.tsx
```diff
@@ -176,9 +180,10 @@ const sendErrorWithContext = (
         extra.windowState = currentState.windowState
         extra.accounts = `${currentState.accounts.length}`
 
-        if (__DARWIN__) {
```
say25, line 179: Forgot to add this comment when I opened the PR: This check should have been removed before. We support this on Windows as of https://github.com/desktop/desktop/pull/9016

app/src/ui/lib/theme-change-monitor.ts
```diff
@@ -15,27 +19,26 @@ class ThemeChangeMonitor implements IDisposable {
...
   }
 
-  private onThemeNotificationFromOS = (event: string, userInfo: any) => {
-    const darkModeEnabled = isDarkModeEnabled()
-
-    const theme = darkModeEnabled
+  private onThemeNotificationUpdated = (event: string, userInfo: any) => {
+    const theme = getCurrentlyAppliedTheme()
       ? ApplicationTheme.Dark
       : ApplicationTheme.Light
```
sergiou87: Shouldn't these two lines be removed entirely?

sergiou87 (changes requested):
Only one change needed before merging! Thanks for your great work! 😍

app/src/ui/lib/theme-change-monitor.ts
say25: Yes, not sure how I missed that 🤔

sergiou87 (approved):
Thanks for that fix, LGTM now! :shipit:

app/src/ui/lib/theme-change-monitor.ts
say25: Looking deeper, this worked because enum Dark (1) vs Light (0) became the proper true false for the statement.

PR #11642, opened 2021-02-23
Cherry picking Continue After Conflicts

tidy-dev:
Part of work on #1685

## Description

Added to continue cherry pick logic to allow for manual resolutions as well as handling the scenario of if a user resolves all conflicts by selecting the current's branches resolution resulting in an empty commit.

User can now cherry pick a commit resulting in conflicts, resolve conflicts, and continue the cherry pick for a successful cherry pick. 

continueCherryPick

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -277,7 +277,7 @@ export async function getCherryPickSnapshot(
   }
 
   const count = commits.length - remainingShas.length
-  const commitSummaryIndex = (count > 0 ? count : 1) - 1
+  const commitSummaryIndex = count > 0 ? count - 1 : 0
   return {
     progress: {
       kind: 'cherryPick',
@@ -316,13 +316,13 @@ export async function continueCherryPick(
   // apply conflict resolutions
   for (const [path, resolution] of manualResolutions) {
     const file = files.find(f => f.path === path)
-    if (file !== undefined) {
-      await stageManualConflictResolution(repository, file, resolution)
-    } else {
+    if (file === undefined) {
       log.error(
         `[continueCherryPick] couldn't find file ${path} even though there's a manual resolution for it`
       )
+      continue
     }
+    await stageManualConflictResolution(repository, file, resolution)
   }
 
   const otherFiles = trackedFiles.filter(f => !manualResolutions.has(f.path))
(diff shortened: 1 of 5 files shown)


app/src/lib/git/cherry-pick.ts
```diff
@@ -275,14 +277,15 @@ export async function getCherryPickSnapshot(
   }
 
   const count = commits.length - remainingShas.length
+  const commitSummaryIndex = (count > 0 ? count : 1) - 1
```
sergiou87: Kind of a nit, but I find this easier to read (since my brain is already wired for 0-index 🤣 )
```suggestion
  const commitSummaryIndex = count > 0 ? count - 1 : 0
```

Feel free go ignore this, though!

app/src/lib/git/cherry-pick.ts
```diff
@@ -302,13 +305,28 @@ export async function getCherryPickSnapshot(
...
+  // apply conflict resolutions
+  for (const [path, resolution] of manualResolutions) {
+    const file = files.find(f => f.path === path)
+    if (file !== undefined) {
+      await stageManualConflictResolution(repository, file, resolution)
+    } else {
+      log.error(
+        `[continueCherryPick] couldn't find file ${path} even though there's a manual resolution for it`
+      )
+    }
```
sergiou87: As a general rule I prefer to avoid `else` and extra nesting when handling these kinds of checks:
```suggestion
    if (file === undefined) {
      log.error(
        `[continueCherryPick] couldn't find file ${path} even though there's a manual resolution for it`
      )
      continue
    }

    await stageManualConflictResolution(repository, file, resolution)
```

Basically, I usually try to turn error checks into "fast returns" (or fast `continue`s in this case 🤣 ), leaving the "happy path" in a clearer state.

What do you think?

sergiou87 (commented):
LGTM! I only have a couple of nits/comments, let me know what you think about them.

app/src/lib/git/cherry-pick.ts
tidy-dev: Yep. I do too.. :) copy/pasta from rebase and merge. Random side note, on my last team, I got the opposite PR comments to undo fast returns. This comment makes me happy! 😄

app/src/lib/git/cherry-pick.ts
tidy-dev: lol .. now that I come back to it.. that is harder to read.. it was a result of doing A then modifying to B without rethinking it.

sergiou87 (approved):
✨ :shipit: ✨

PR #11644, opened 2021-02-23
Cherry picking launch conflicts dialog on open

tidy-dev:
Part of work on #1685

## Description

Launches conflicts dialog on desktop open with conflicted cherry picking state

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -353,7 +353,7 @@ export function isRebaseConflictState(
 }
 
 /**
- * Conflicts can occur during a rebase, merge, or cherry pick
+ * Conflicts can occur during a rebase, merge, or cherry pick.
  *
  * Callers should inspect the `kind` field to determine the kind of conflict
  * that is occurring, as this will then provide additional information specific
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -353,7 +353,7 @@ export function isRebaseConflictState(
 }
 
 /**
- * Conflicts can occur during a rebase, merge, or cherry pick.
+ * Conflicts can occur during a rebase, merge, or cherry pick
  *
  * Callers should inspect the `kind` field to determine the kind of conflict
  * that is occurring, as this will then provide additional information specific
--- a/app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
+++ b/app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
@@ -38,7 +38,7 @@ interface ICherryPickConflictsDialogProps {
   readonly workingDirectory: WorkingDirectoryStatus
 
   // For display in manual resolution context menu
-  readonly sourceBranchName: string
+  readonly sourceBranchName: string | null
 
   readonly onDismissed: () => void
   readonly onContinueCherryPick: (step: ShowConflictsStep) => void
@@ -123,13 +123,18 @@ export class CherryPickConflictsDialog extends React.Component<
       repository,
       dispatcher,
       step,
-      sourceBranchName: ourBranch,
+      sourceBranchName,
     } = this.props
 
     const {
       manualResolutions,
       targetBranchName: theirBranch,
     } = step.conflictState
+
+    let ourBranch: string
+    if (sourceBranchName !== null) {
+      ourBranch = sourceBranchName
+    }
 
     return (
       <ul className="unmerged-file-statuses">
--- a/app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
+++ b/app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
@@ -131,11 +131,6 @@ export class CherryPickConflictsDialog extends React.Component<
       targetBranchName: theirBranch,
     } = step.conflictState
 
-    let ourBranch: string
-    if (sourceBranchName !== null) {
-      ourBranch = sourceBranchName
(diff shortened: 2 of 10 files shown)


app/src/lib/app-state.ts
```diff
@@ -353,7 +353,7 @@ export function isRebaseConflictState(
 }
 
 /**
- * Conflicts can occur during a rebase, merge, or cherry pick.
+ * Conflicts can occur during a rebase, merge, or cherry pick
```
sergiou87: Nit (and LOL): I liked the `.` 😂  (I guess you deleted that `.` by accident, but if not feel free to ignore my dumb commit 🤣 )

app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
```diff
@@ -123,14 +123,19 @@ export class CherryPickConflictsDialog extends React.Component<
...
 
     const {
       manualResolutions,
       targetBranchName: theirBranch,
     } = step.conflictState
 
+    let ourBranch: string
+    if (sourceBranchName !== null) {
+      ourBranch = sourceBranchName
+    }
```
sergiou87: Doesn't this code make the app fail when `ourBranch` is used a few lines below? Because if `sourceBranchName` happens to be `null`, `ourBranch` would be uninitialized 🤔 

image

app/src/ui/cherry-pick/cherry-pick-flow.tsx
```diff
@@ -154,7 +159,7 @@ export class CherryPickFlow extends React.Component<ICherryPickFlowProps> {
             openFileInExternalEditor={openFileInExternalEditor}
             resolvedExternalEditor={resolvedExternalEditor}
             openRepositoryInShell={openRepositoryInShell}
-            sourceBranchName={sourceBranch.name}
+            sourceBranchName={sourceBranch ? sourceBranch.name : null}
```
sergiou87: I think Markus preferred to avoid truthy/falsey checks:
```suggestion
            sourceBranchName={sourceBranch !== null ? sourceBranch.name : null}
```

sergiou87 (changes requested):
Just a few things, but again looks awesome!

app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
tidy-dev: No. Actually you can test this scenario too! Because it will be null if you start a conflicted cherry pick outside of our app. The `renderUnmergedFile` is made to gracefully handle it .. maybe not the most optimal, but you do not get the name of the source on that dropdown for which side of the conflict you want in the conflict dialog.. but you still get the target branch name for opposing.. thus, in this unique scenario you can still differentiate between target and source even if not in the most clear manner.

I looked into what it would take to determine the source branch in this scenario... and it is technically possible and I have in my cherry picking notes to look into it further to make this one scenario better. But, I was comfortable as this is usable and there are paths that fail much harder in cherry picking that I think warrant more attention (starting without a clean working tree for example).

app/src/lib/app-state.ts
tidy-dev: Lol.. I guess that period just made me mad one day.. 😄

app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
tidy-dev: image

app/src/ui/cherry-pick/cherry-pick-flow.tsx
tidy-dev: Yep. Sorry. in grained habit thanks for the catch.

app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
sergiou87: I'm sorry, I wasn't referring at the possibility of not having a source branch. You explained that scenario perfectly in one of the comments (and I'm very grateful for that 🙏 ❤️ ).

In this case I'm referring strictly to the code side of things:

```
    let ourBranch: string // declared, but not initialized!!!
    if (sourceBranchName !== null) {
      ourBranch = sourceBranchName // initialized 😌 
    }
    // what happens at this point if the sourceBranchName was null?
    // ourBranch would still be uninitialized, which should fail at compile time as I showed in my screenshot, no?
```

I will check out this branch later and look at this 👆  if I have time before you get back to it 😅

tidy-dev: This was intentional. In fact, I just updated the code to show that intention better. Js being runtime.. doesn't compile really and so it is ok for things to be `undefined`.. it is similar behavior to `null`.

As you can see below, the function `renderUnmergedFIle` is the only place I use the `ourBranch` and on the other side of the split you can see that `ourBranch` has the ? as an optional property. If I send in `undefined`, it will be same as if I did not send it at all. This is what I want, I want to send it if the sourceBranch exists, but I don't want to send it if I don't have anything to send.

image
image

sergiou87: Ohhh!! I totally missed that `?`, but the new code makes it even clearer, thank you!

sergiou87 (approved):
All looking good now! Thank you! 💖 :shipit:

PR #11683, opened 2021-03-03
Support deploying arm64 builds

sergiou87:
## Description

This PR prepares our build scripts to deploy arm64 binaries to Central. Depends on github/central#500

Changes:
- Add the new builds to the README file. macOS only for now.
- Add new checks for the current architecture, trying to guess the best build we should try to update to (e.g. if we're running the macOS on Rosetta, we know we should try to update to the arm64 binary).
- New S3 keys to create a subfolder for each architecture.
- Take into account the new contexts created (`darwin-arm64` and `win32-arm64`).

## Release notes

Notes: no-notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,7 @@ jobs:
       - name: Publish production app
         run: yarn run publish
         env:
+          npm_config_arch: ${{ matrix.arch }}
           DESKTOPBOT_TOKEN: ${{ secrets.DESKTOPBOT_TOKEN }}
           WINDOWS_CERT_PASSWORD: ${{ secrets.WINDOWS_CERT_PASSWORD }}
           KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ uses [React](https://reactjs.org/).
 Download the official installer for your operating system:
 
  - [macOS](https://central.github.com/deployments/desktop/desktop/latest/darwin)
- - [macOS (M1 Chip)](https://central.github.com/deployments/desktop/desktop/latest/darwin-arm64)
+ - [macOS (Apple Silicon)](https://central.github.com/deployments/desktop/desktop/latest/darwin-arm64)
  - [Windows](https://central.github.com/deployments/desktop/desktop/latest/win32)
  - [Windows machine-wide install](https://central.github.com/deployments/desktop/desktop/latest/win32?format=msi)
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -27,6 +28,7 @@ Want to test out new features and get fixes before everyone else? Install the
 beta channel to get access to early builds of Desktop:
 
  - [macOS](https://central.github.com/deployments/desktop/desktop/latest/darwin?env=beta)
+ - [macOS (M1 Chip)](https://central.github.com/deployments/desktop/desktop/latest/darwin-arm64?env=beta)
```
say25:
```suggestion
 - [macOS (Apple Silicon)](https://central.github.com/deployments/desktop/desktop/latest/darwin-arm64?env=beta)
```

I tend to think this is the "best way" to highlight these things, I think a future Mac could very well be shipping with an M2 or M1X (or so my sources say)

sergiou87: Good point! I chose `M1` because that's how [Apple showcases it](https://www.apple.com/mac/m1/) and I don't see any references to `Apple Silicon`, but I agree it's probably more accurate and future-proof 😄 Thanks!!

parishshani (approved):

parishshani (approved):

parishshani (approved):

korakit171129 (dismissed):

rwidmark:
Question, if m1 user already has PS (intel) installed trough brew, will this (PS m1 support) uninstall and install m1 version when they update?

If not, then I think that's a good ide to have.

parishshani (approved):

parishshani (approved):

.github/workflows/ci.yml
```diff
@@ -93,6 +93,7 @@ jobs:
       - name: Publish production app
         run: yarn run publish
         env:
+          npm_config_arch: ${{ matrix.arch }}
```
sergiou87, line 96: This is what's used to tell our scripts about the target architecture.

app/src/ui/lib/update-store.ts
```diff
@@ -160,10 +161,24 @@ class UpdateStore {
...
+    // on an arm64 machine), we need to tweak the update URL here to point at
+    // the arm64 binary.
+    if (
+      enableUpdateFromRosettaToARM64() &&
+      remote.app.runningUnderRosettaTranslation === true
+    ) {
+      const url = new URL(updatesURL)
+      url.searchParams.set('architecture', 'arm64')
+      updatesURL = url.toString()
+    }
```
sergiou87, line 176: This is needed because `__UPDATES_URL__` is set at compile time, so we can't set it to look for arm64 for users who downloaded the x64 build to then run it on an Apple Silicon machine.

tidy-dev (approved):
Makes sense to me. ✨

PR #11695, opened 2021-03-04
Release 2.6.5

sergiou87:
## Description

Looking for the PR for the upcoming v2.6.5 production release? Well you've just found it, congratulations! :tada:

## Release checklist

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -1,5 +1,6 @@
 {
   "releases": {
+    "2.6.5": ["[Fixed] Crash sometimes when performing remote Git operations"],
     "2.6.4": [
       "[Added] Allow users to rename and delete branches via a new context menu on branches in the branches list - #5803 #10432",
       "[Fixed] Allow users to modify git config on a per repository basis - #9449. Thanks @say25!",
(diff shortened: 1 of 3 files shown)


changelog.json
```diff
@@ -1,5 +1,6 @@
 {
   "releases": {
+    "2.6.5": ["[Fixed] Crash sometimes when performing remote Git operations"],
```
tidy-dev:
```suggestion
    "2.6.5": ["[Fixed] Inform user of errors encountered when performing remote Git operations"],
```

tidy-dev (changes requested):
Just a wording suggestion. For some reason, I had to read the other one a couple times to comprehend. Hopefully this is clearer.

changelog.json
sergiou87: Not sure what to think, I'm 👍  with improving the wording, but I think we shouldn't remove the "crash" part from the release note, since I'm guessing users will be looking for it to make sure that's what we fixed. What do you think? Maybe something like this?
```suggestion
    "2.6.5": ["[Fixed] Performing remote Git operations could crash the app"],
```

tidy-dev: I like it!

billygriffin (dismissed):
This seems reasonable but let's chat for a minute before we ship it.

tidy-dev (approved):

PR #11702, opened 2021-03-04
Prompt people to download beta in ask-pass part of known issues doc

billygriffin:
This adds a note at the top of the ask-pass issue in the known issues doc that prompts people to download the beta.


steveward (approved):
👍🏻

PR #11712, opened 2021-03-05
Cherry picking multiple commits

tidy-dev:
Part of #1685

## Description

User can now cherry pick multiple commits from the context menu!

### Screenshots

https://user-images.githubusercontent.com/75402236/110111690-b9b61000-7d7e-11eb-9b1f-27ed42c7f921.mp4

## Release notes
Notes: no-notes


app/src/ui/lib/list/list.tsx
```diff
@@ -1065,6 +1066,14 @@ export class List extends React.Component<IListProps, IListState> {
             event,
           })
         }
+      } else if (
```
tidy-dev, line 1069: @sergiou87 You may be rightly weary of changes to this generalized `List` component that is the basis for most if not all lists in the app.  Just for some piece of mind I am adding this screen shot to show that selectionMode of range is only used for the commit list.  Thus, this 'should' not impact other lists... (tho out of caution I did regression test the changes list as it makes significant use of the methods I modified here) 

image

(Also in case it is not readily clear.. tho I hope my comments indicate it. The reason for this change is that if a user had multiple commits selected and then pressed down on one to start dragging the multiple commits, it would invoke a single selection again. Tho the state at the time of the drag still had the multiple commits so it "worked" it appeared like they were only dragging one)

app/src/ui/history/commit-list-item.tsx
```diff
@@ -227,7 +239,25 @@ export class CommitListItem extends React.PureComponent<
...
+    const pluralize = count === 1 ? '' : 's'
+    const countOut = count === 1 ? '' : ` ${count} `
+    if (enableCherryPicking()) {
+      items.push({
+        label: __DARWIN__
+          ? `Cherry Pick${countOut}Commit${pluralize}…`
+          : `Cherry pick${countOut}commit${pluralize}…`,
+        action: this.onCherryPick,
+      })
+    }
```
sergiou87: Unless I'm missing something, we shouldn't need to check the `count` here, right? In this method there will always be `count > 1` commits. Otherwise, I think would have a weird string with `count === 1` unless we add a space to the `countOut` string.
```suggestion
    const pluralize = count === 1 ? '' : 's'
    const countOut = count === 1 ? ' ' : ` ${count} `
    if (enableCherryPicking()) {
      items.push({
        label: __DARWIN__
          ? `Cherry Pick${countOut}Commit${pluralize}…`
          : `Cherry pick${countOut}commit${pluralize}…`,
        action: this.onCherryPick,
      })
    }
```

app/src/ui/history/commit-list.tsx
```diff
@@ -194,9 +218,10 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
         <List
           rowCount={this.props.commitSHAs.length}
           rowHeight={RowHeight}
-          selectedRows={[this.rowForSHA(this.props.selectedSHA)]}
+          selectedRows={this.props.selectedSHAs.map(sha => this.rowForSHA(sha))}
```
sergiou87, line 271: Not saying you have to do this, but you could, right? Again, nothing to do in this PR, just curious to know what is your opinion 🤣 
```suggestion
          selectedRows={this.props.selectedSHAs.map(this.rowForSHA)}
```

app/src/ui/history/commit-list.tsx
```diff
@@ -194,9 +218,10 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
           rowRenderer={this.renderCommit}
-          onSelectedRowChanged={this.onSelectedRowChanged}
+          onSelectedRangeChanged={this.onSelectedRangeChanged}
+          selectionMode="range"
```
sergiou87: This should be feature-flagged, right?

sergiou87 (changes requested):
Looks good overall, I only have a couple of comments and one actual request.

I also tested it and found a couple of things:
- The first one feels like a regression: using the keyboard to go up and down (with the arrow keys or with the Page Up and Page Down keys).
- Also if you select multiple commits, let's say range A...D, but then while those are selected you drag F, it will actually be dragging (and cherry-picking) A...D. I don't think this is a major issue, but probably worth at least thinking about it.

None of this need to be fixed in this PR, but the first one definitely needs to be fixed before the beta, I think, since it'd be a regression.

app/src/ui/history/commit-list-item.tsx
tidy-dev: Absolutely.. good catch - from before I split it out into two methods.

app/src/ui/history/commit-list.tsx
tidy-dev: It doesn't work.. I tried that first. 😄  Scoping of "this"
image

But, for future reference, if it is just a helper method. (something like just transforms.. agnostic of scope). I am all for this syntax.

app/src/ui/history/commit-list.tsx
tidy-dev: Absolutely. Thank you

tidy-dev: Absolutely. Thank you

tidy-dev:
> The first one feels like a regression: using the keyboard to go up and down (with the arrow keys or with the Page Up and Page Down keys).

 I am sorry I am not seeing it? EDIT: I added the "onSelectionChange" handler back for feature flag fix before looking at this.. and that fixed it. Updating comment on that function to signify its importance here.

https://user-images.githubusercontent.com/75402236/110364132-67d8e880-8011-11eb-8290-7718a31d1745.mp4

tidy-dev:
> Also if you select multiple commits, let's say range A...D, but then while those are selected you drag F, it will actually be dragging (and cherry-picking) A...D. I don't think this is a major issue, but probably worth at least thinking about it.

This one actually bothered me more. 😸  A great find!  I updated the logic delaying the selection on mouse down of multiple to only delay new selection if user is holding down on one of the selected rows. Other wise, dragging a commit outside of the selection will incur a new selection and with adding the onSelectionChange back in will correctly invoke that selection the cherry pick.

sergiou87 (approved):
All is working perfectly now!! Thanks for those changes ❤️

sergiou87:
Well, only one nit about that new image in the intro callout: it takes a bit to load and causes an annoying glitch. Nothing major to block this PR, but maybe worth mentioning 🤣 

Not sure if there is any better solution to this, but since it's a static image we control, we could set a fixed width and height to the `img` element, and at least that would prevent the callout resizing. What do you think @tidy-dev ?

https://user-images.githubusercontent.com/1083228/110445998-02433580-80bf-11eb-8f79-d190de1719bd.mov

tidy-dev:
> Well, only one nit about that new image in the intro callout: it takes a bit to load and causes an annoying glitch. Nothing major to block this PR, but maybe worth mentioning 🤣

Thank you for that attention to detail. This picture might change depending on what our drag ghost ends up like so I am holding off on tweaking it until that is finalized.. also in other places in the app we have svg so I may look into what is involved in getting this as an svg (which should help with better rendering too)

PR #11713, opened 2021-03-05
Cherry Picking Graceful Fail

tidy-dev:
Part of #1685

## Description

Adds logic to abort any ongoing cherry pick in the event of an unhandled failure. (hopefully there won't be any.. but in case) If user reaches a failure resulting in a generic error dialog, trying to cherry pick after closing it will just result in more errors due to attempting to cherry pick with a ongoing cherry pick. This will clear this bad state.. so maybe they can try cherry picking another way..

## Release notes
Notes: no-notes


sergiou87 (approved):
LGTM ✨ :shipit:

PR #11716, opened 2021-03-05
Cherry Picking Undo Cherry Picked Commits

tidy-dev:
Part of #1685

## Description

Allows a user to undo the cherry picked commits.  

Note: We may expand upon this to revert back to the source branch of the cherry pick. But this part is needed either way.

### Screenshots

https://user-images.githubusercontent.com/75402236/110150535-5f807380-7dad-11eb-96c5-df3a4f5e9352.mp4

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -755,6 +755,13 @@ export interface ICherryPickState {
    * cherry pick.
    */
   readonly userHasResolvedConflicts: boolean
+
+  /**
+   * The sha of the target branch tip before cherry pick initiated.
+   *
+   * This will be set to null if no cherry pick has been initiated.
+   */
+  readonly targetBranchUndoSha: string | null
 }
 
 /**
--- a/app/src/ui/banners/successful-cherry-pick.tsx
+++ b/app/src/ui/banners/successful-cherry-pick.tsx
@@ -1,32 +1,45 @@
 import * as React from 'react'
+import { LinkButton } from '../lib/link-button'
 import { Octicon, OcticonSymbol } from '../octicons'
 import { Banner } from './banner'
 
-export function SuccessfulCherryPick({
-  targetBranchName,
-  countCherryPicked,
-  onDismissed,
-}: {
+interface ISuccessfulCherryPickBannerProps {
   readonly targetBranchName: string
   readonly countCherryPicked: number
   readonly onDismissed: () => void
-}) {
-  const pluralized = countCherryPicked === 1 ? 'commit' : 'commits'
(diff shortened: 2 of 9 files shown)


MADTechMarket (commented):
says user has resolved conflicts im reviewing steps

app/src/lib/app-state.ts
```diff
@@ -755,6 +755,13 @@ export interface ICherryPickState {
    * cherry pick.
    */
   readonly userHasResolvedConflicts: boolean
+
+  /**
+   * The sha of the target branch tip before cherry pick initiated.
+   *
+   * This will be set to null if no cherry pick has been initiated.
```
sergiou87, line 762: If I'm reading this PR correctly, this only works if the cherry-pick was initiated from GitHub Desktop, right?

So maybe we should specify that here, if there is no way to get the original sha in that scenario.
```suggestion
   * This will be set to null if no cherry pick has been initiated, or if it was
   * initiated from outside of Desktop.
```

app/src/ui/banners/successful-cherry-pick.tsx
```diff
@@ -1,32 +1,45 @@
...
+        onDismissed={onDismissed}
+      >
+        <div className="green-circle">
+          <Octicon className="check-icon" symbol={OcticonSymbol.check} />
+        </div>
+        <div className="banner-message">
+          <span>
+            Successfully copied {countCherryPicked} {pluralized} to{' '}
+            <strong>{targetBranchName}</strong>.{' '}
+            <LinkButton onClick={this.undo}>Undo</LinkButton>
```
sergiou87, line 39: If we can't get the original sha for scenarios where cherry-picking was initiated outside of our app, we should probably remove this link, right?

sergiou87 (changes requested):

app/src/lib/app-state.ts
tidy-dev: Very good point! Thank you! And fortunately, that sha is stored in the sequencer during an active cherry pick. I just added it to the setting progress based on state method.

app/src/ui/banners/successful-cherry-pick.tsx
tidy-dev: (see answer to other comment)

app/src/lib/app-state.ts
sergiou87: Yay!! 🎉

sergiou87 (approved):
LGTM! :shipit:

PR #11717, opened 2021-03-05
Cherry Picking allow redundant and empty commits

tidy-dev:
Part of #1685

## Description

Allows user to cherry pick redundant or empty commits. This will result in a commit added to the target branches history that is empty but has the same summary of the cherry picked commit. 

Some discussion about this took place. Decision was we wouldn't want to stop user flow for these scenarios.
Especially in the case of a user selecting 5 commits and one in the middle was redundant/empty.

However, it does open the user up to make "weird" empty commits as many times as they would like.
May warrant further discussion. (Alternative in the same vein of not stopping flow that wasn't discussed is "skipping" these commits)

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -161,8 +161,12 @@ export async function cherryPick(
     )
   }
 
+  // --keep-redundant-commits follows pattern of making sure someone cherry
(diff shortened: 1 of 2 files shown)


sergiou87 (approved):
LGTM

PR #11725, opened 2021-03-07
import refactor

angad777:
Closes #[issue number]

## Description

-

## Release notes

Notes:


sergiou87:
Thank you for your contribution! ❤️ 

However, right now we prefer the other style with `import * as React from 'react'` which we also use in the rest of the codebase, sorry! 😅

PR #11747, opened 2021-03-09
Add cherry picking stats

tidy-dev:
Part of #1685

## Description
Adding stats to track cherry picking.

Questions attempting to shed light on:
1. What % of users cherry pick successfully?
2. What % of users use drag and drop?
3. What % of users use context menu?
4. What % of users start a drag and drop cherry pick but cancel it?
5. What % of users enter into conflicts and succeed after?
6. What % of users undo a cherry pick?
7. What % of users cherry pick multiple commits?

Notes:
1. Test our hypothesis that this is a desired feature.

2/3 - Test our hypothesis that users will prefer the drag/drop UI.

4. We are concerned users could start dragging and not know where to drag the commits. This will track if a user starts to drag and releases without starting a cherry pick. (This could indicate erroneous drags as well as not sure where to land drags).

5. Are conflicts common/are we providing a clear path to resolution?
6. Is undo a used pattern in this situation?
7. Multiple commits selection is not readily apparent - do we need more info on it?

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -559,6 +559,8 @@ export class RepositoryView extends React.Component<
   private onDragCommitEnd = async () => {
     this.props.dispatcher.closeFoldout(FoldoutType.Branch)
 
+    // On drag end fires before the drop event, we need to wait here and see if
+    // the drop event results in a change of cherry picking state.
     await sleep(10)
     const { cherryPickState } = this.props.state
     if (
--- a/app/src/ui/repository.tsx
+++ b/app/src/ui/repository.tsx
@@ -14,6 +14,7 @@ import {
   IRepositoryState,
   RepositorySectionTab,
   ChangesSelectionKind,
+  FoldoutType,
 } from '../lib/app-state'
 import { Dispatcher } from './dispatcher'
 import { IssuesStore, GitHubUserStore } from '../lib/stores'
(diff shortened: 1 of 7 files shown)


app/src/ui/repository.tsx
```diff
@@ -545,4 +549,27 @@ export class RepositoryView extends React.Component<
...
+   * dragged.
+   *
+   * Currently only used for cherry picking, but this could be more generic.
+   */
+  private onDragCommitEnd = async () => {
+    this.props.dispatcher.closeFoldout(FoldoutType.Branch)
+
+    // On drag end fires before the drop event, we need to wait here and see if
+    // the drop event results in a change of cherry picking state.
+    await sleep(10)
```
sergiou87, line 564: This is very scary IMO. If it takes any longer to update the state this logic would break, no?

Can't we listen to the drop event instead?

sergiou87 (changes requested):

app/src/ui/repository.tsx
sergiou87: @tidy-dev and I discussed about this. We need a slightly bigger refactor to get rid of this `sleep`, so we're gonna go ahead and merge this PR, and then tackle that problem separately and ASAP!!

sergiou87 (approved):
Other than the `sleep`, that we agreed will address separately since it doesn't break anything in case it's too short, everything else LGTM.

:shipit:

PR #11753, opened 2021-03-09
Cherry picking merge commits error

tidy-dev:
Part of #1685 
## Description

Adds a friendly error message for when a user tries to cherry pick a merge commit.

If source branch known, it takes user back to source branch after error. (Following convention set in undo logic of "take user back to where they were")

Bumps dugite as latest version parses the merge commit git error.

No longer throws generic error, because we now expect that cherry pick can throw an error.

### Screenshots

image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -6088,13 +6088,28 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
-  public async _clearCherryPickingHead(repository: Repository): Promise<void> {
+  public async _clearCherryPickingHead(
+    repository: Repository,
+    sourceBranch: Branch | null
+  ): Promise<void> {
     if (!isCherryPickHeadFound(repository)) {
       return
     }
 
     const gitStore = this.gitStoreCache.get(repository)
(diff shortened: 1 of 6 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -6088,13 +6088,28 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      return this._refreshRepository(repository)
+    }
+
+    await this.withAuthenticatingUser(repository, async (r, account) => {
+      await gitStore.performFailableOperation(() =>
+        checkoutBranch(repository, account, sourceBranch)
+      )
+    })
+
+    return this._refreshRepository(repository)
```
sergiou87: As fan as I am of early returns, maybe it's better here to do the opposite?
```suggestion

    if (sourceBranch !== null) {
      await this.withAuthenticatingUser(repository, async (r, account) => {
        await gitStore.performFailableOperation(() =>
          checkoutBranch(repository, account, sourceBranch)
        )
      })
    }

    return this._refreshRepository(repository)
```

sergiou87 (changes requested):
Aside from the need to update the branch with `development`, I only have one suggestion. Other than that LGTM 😄

sergiou87 (approved):
Nice cleanup!! ✨ :shipit:

PR #11764, opened 2021-03-10
Cherry picking refactor drop cancelled logic

tidy-dev:
Part of #1685

## Description

Refactor cherry pick drop cancelled logic

## Release notes

Notes: no-notes


sergiou87 (approved):
I still wonder if there is a better way to handle this drag&drop madness, but for now this is good and it's definitely much much better than the `sleep` approach, as this is 100% deterministic ❤️ 

Thank you for taking the time to refactor this!!!! 🙇‍♂️ :shipit:

PR #11790, opened 2021-03-14
Cherry picking ghosts

tidy-dev:
Part of #1685 

## Description

Refactor Drag/Drop logic to a custom solution that allows for customized drag ghost.

### Screenshots

Mac
image

Windows
image

## Release notes

Notes: no-notes


app/src/ui/branches/branches-container.tsx
```diff
@@ -155,7 +158,9 @@ export class BranchesContainer extends React.Component<
       this.props.currentBranch,
       this.onRenameBranch,
       this.onDeleteBranch,
-      this.onDropOntoBranch
+      this.onDropOntoBranch,
+      this.props.onDropOntoCurrentBranch,
+      this.props.isCherryPickInProgress
```
sergiou87, line 163: This shouldn't cause issues, but it's a bit misleading since a cherry-pick could be in progress but not be a drag&drop cherry-pick.

I was testing the PR and found an issue related to this, but also to the cherry-pick state not cleared properly: if you right-click on a commit, select cherry-pick and then dismiss the popup by clicking outside (and not in the X), you can't cherry pick some commits, and branches are highlighted like when you drag something over them.

The latter is what I think is related to this `this.props.isCherryPickInProgress` along with the cherry pick state not being cleared properly after dismissing the popup.

https://user-images.githubusercontent.com/1083228/111141920-5e58fe80-8584-11eb-88ec-53a2185a2395.mov

app/src/ui/history/commit-list-item.tsx
```diff
@@ -304,21 +326,155 @@ export class CommitListItem extends React.PureComponent<
...
+
+      // We must be over a branch.
+      if (branchListItem !== null) {
+        ghost.classList.add('over-branch')
+        // Grabbing branch name for copy label on windows implementation
+        const branchNameElement = branchListItem.querySelector('.name')
+        if (branchNameElement !== null && copyMessageLabelElement) {
+          copyMessageLabelElement.innerHTML = branchNameElement.innerHTML
+        }
+      } else if (branchListItem == null) {
```
sergiou87: Probably a leftover, but not needed anymore, right? Or does the compiler complain?
```suggestion
      } else {
```

sergiou87 (changes requested):
LGTM overall. I don't know if the approach using CSS styles to query/check the right elements are being dragged over and all that, but I understand how that's a simple solution to this problem.

Something I _think_ I noticed, though, which might be related to that, is how performance gets worse when you're dragging and you open the branch dropdown. At that point the mouse cursor and the ghost seem more sluggish, while it works perfectly fine when the dropdown is hidden away:

https://user-images.githubusercontent.com/1083228/111147128-73389080-858a-11eb-8fc5-0e5be802ee38.mov

Is that just me or does it actually get worse? There is even a short hiccup right on the moment where the dropdown is shown 🤔

app/src/ui/branches/branches-container.tsx
tidy-dev: Yeah. I caught this and is fixed in my other PR. You should never be able to cherry pick if one is in progress at all.. but the bug was that I didn't clear cherry pick state on x of choose branch dialog.

app/src/ui/history/commit-list-item.tsx
tidy-dev: Nope just left over from where I was comparing against null only before. Thx

tidy-dev:
> LGTM overall. I don't know if the approach using CSS styles to query/check the right elements are being dragged over and all that, but I understand how that's a simple solution to this problem.

Yep.. I agree.. using the query selectors is not my preferred approach.. but, in order to have a customized ghost. It is required.. however, if I succeed in refactoring to draggable component with a drag state. I think I can refactor some of it out anyways. 

> Something I _think_ I noticed, though, which might be related to that, is how performance gets worse when you're dragging and you open the branch dropdown. At that point the mouse cursor and the ghost seem more sluggish, while it works perfectly fine when the dropdown is hidden away:
> 
>  cherry-pick-perf-issues.mov 
> Is that just me or does it actually get worse? There is even a short hiccup right on the moment where the dropdown is shown 🤔

I believe you are right.. however, I believe it is still better for the user than what we had before in terms of being able to indicate multiple commits in the drag. When the dropdown is open, I am modifying it as well as modifying the branches state; guessing this is why there is lag.

Native Drag/Drop Saga (if your interested): The reason this doesn't happen for the native drag and drop is because it somehow takes a snapshot of the element you provide it as the ghost and doesn't actually drag an element around making it more performant - at least that is my assumption. Because it is quite annoying otherwise... when I implemented the count bubble in the native drag/drop solution it would take a snapshot of the commit right above the commit I was dragging when trying to take a snapshot of the actual commit html... since the count bubble was hovering over it. Also.. that bubble would stay on the commit list for the entire drag.. and if I made a duplicate commit to throw away.. it couldn't take the screenshot if I had it out of view and wouldn't show up if I tried to throw it away quickly (which was annoyingly hacky anyways)... thus.. here I am with non-optimal solution number 2. 😂

sergiou87 (approved):
As discussed in Slack, the performance issue happens **after** the branch dropdown is visible, not before.

I _think_ the problem is not the ghost itself, but the CSS queries that are run when the dropdown is visible.

However, there is a refactor in progress for the drag&drop stuff that might address the performance issues. In any case, we agree it's worth tackling it separately.

:shipit:

PR #11795, opened 2021-03-15
Cherry picking mutliple commits text in branch dialog and [New] callout for history tab

tidy-dev:
Part of #1685

## Description

When multiple commits selected via context menu, have branch dialog indicate that multiple commits are selected.
(Also noticed a bug that if you opened branch modal and then dismissed, cherry pick state wasn't cleared)

When user has not been introduced to cherry picking yet, have [New] call out on history tab. It is removed once user hits got it on cherry pick intro pop over.

### Screenshots
image

image

## Release notes

Notes: no-notes


sergiou87 (changes requested):
LGTM overall, there is an edge case though where empty repos still show the `(New)` badge, but you can't get rid of it yet:

image

sergiou87 (approved):
Looks great, thank you!!

PR #11800, opened 2021-03-15
Cherry picking ghosts take two

tidy-dev:
Part of #1685

## Description

Refactor drag and drop logic into a react friendly, easier to understand/maintain, draggable component. Appears to also be more performant than last attempt.

## Release notes
Notes: no-notes


app/src/lib/app-state.ts
```diff
@@ -39,6 +39,7 @@ import { IStashEntry } from '../models/stash-entry'
 import { TutorialStep } from '../models/tutorial-step'
 import { UncommittedChangesStrategy } from '../models/uncommitted-changes-strategy'
 import { CherryPickFlowStep } from '../models/cherry-pick'
+import { DragElement } from '../models/dragElement'
```
sergiou87: I don't think we have any file in the code base with camel casing, instead I think it should be…
```suggestion
import { DragElement } from '../models/drag-element'
```

app/src/ui/app.tsx
```diff
@@ -2184,6 +2187,45 @@ export class App extends React.Component<IAppProps, IAppState> {
...
 
+  private renderDragElement() {
+    return <div id="dragElement">{this.currentDragElement()}</div>
+  }
+
+  /**
+   * Render the current drag element based on it's type. Used in conjunction
+   * with the `Draggable` component.
+   */
+  private currentDragElement(): JSX.Element | null {
```
sergiou87: Nit: maybe this should be renamed to `renderCurrentDragElement` to avoid confusion with the `currentDragElement` property in state?

app/src/ui/lib/draggable.tsx
```diff
@@ -0,0 +1,138 @@
...
+  /** Callback to remove a drag element inside the #dragElement */
+  readonly onRemoveDragElement: () => void
+
+  /** Whether dragging is enabled */
+  readonly isEnabled: boolean
+
+  /** An array of css selectors for elements that are valid drop targets. */
+  readonly dropTargetSelectors: ReadonlyArray<string>
+}
+export class Draggable extends React.Component<IDraggableProps> {
```
sergiou87, line 32: Ultra nit: missing blank line before the class declaration.

sergiou87 (changes requested):
LGTM! I still see some performance issues. I've looked into it, and the problem comes, I think, when we update the drag element the moment the user hovers a branch while dragging commits.

That ends up in the app emitting an update, which results in the app stuttering. If you just drag your commits around out of the branch dropdown (even when it's visible), performance is not affected.

sergiou87 (approved):
Performance is SO MUCH BETTER with the latest changes, thank you!!

As we discussed in Slack there might be a ~~cleaner~~ different approach to encapsulate all the drag&drop madness which could make everything easier to understand and manage, but that can be implemented at some point later.

Great work!!! 💖 :shipit:

PR #11815, opened 2021-03-16
Preserve and Convert Legacy Theme

### LINKED ISSUE #11814: System Theme Update defaults to System Theme even if you had selected something else
### Describe the bug

If you had your system theme set to dark, but the app theme set to light. Then, you pull the 2.7.0-beta3 update. You will be set onto the system theme. This may be an unavoidable outcome, but wanted to document it here so we can remember to figure out if we can keep users previous settings.

### Version & OS

2.7.0-beta3

Before install of 2.7.0-beta3.
Set your system to dark mode.
Set Desktop to light mode.
Install 2.7.0-beta3.

### Expected behavior

Remember my manually set theme preference.

### Actual behavior

Changes my theme to dark mode and setting set as system theme.

say25:
Closes #11814

## Description

- Fixes small regression where user theme settings are not persisted to the 2.7.0+ upgrade

## Release notes

Notes: No notes

### DIFF
--- a/app/src/ui/lib/application-theme.ts
+++ b/app/src/ui/lib/application-theme.ts
@@ -129,7 +129,7 @@ export function supportsSystemThemeChanges(): boolean {
     return isMacOSMojaveOrLater()
   } else if (__WIN32__) {
     // Its technically possible this would still work on prior versions of Windows 10 but 1809
-    // was released October 2nd, 2018 that the feature can just be "attained" by upgrading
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
Looks good to me! Thanks for taking care of this so quickly 🙌

tidy-dev (approved):
Thank you @say25! I tested it out and it works great.

PR #11818, opened 2021-03-17
Fix race condition writing/reading git config

sergiou87:
## Description

Recently I noticed we got more CI builds failing randomly. After a closer inspection, seems like some of the changes in #11595 introduced race conditions by not waiting for some async functions to complete (or by not returning their promise).

This race condition affects places where a config value is set/removed, and then read immediately after (like we do in `config-test.ts`).

The only areas I see could be affected by something like that are the recently added improvements to fix misattributed commits. More specifically when the git config email is changed from anywhere inside the app, and then the UI is immediately refreshed to reflect the change in the avatar of the committer.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/config.ts
+++ b/app/src/lib/git/config.ts
@@ -135,7 +135,7 @@ export async function setConfigValue(
     HOME: string
   }
 ): Promise<void> {
-  setConfigValueInPath(name, value, repository.path, env)
+  await setConfigValueInPath(name, value, repository.path, env)
 }
 
 /** Set the global config value by name. */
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
Nice find!

app/src/lib/git/config.ts
```diff
@@ -196,7 +196,7 @@ export async function removeGlobalConfigValue(
     HOME: string
   }
 ): Promise<void> {
-  removeConfigValueInPath(name, null, env)
+  await removeConfigValueInPath(name, null, env)
```
say25, line 199: I could be wrong but shouldn't these basically be `return`s instead of awaits. This way we aren't "double awaiting" assuming you have awaits on the "caller"s

PR #11826, opened 2021-03-18
Handle the Fact that themeSource isn't sticky

say25:
## Description

- @sergiou87 pointed to me that ThemeSource is actually not persisted to disk in any way https://github.com/electron/electron/blob/14acf00ba9ecdd8103828659b8a4f9d405fd81ec/patches/chromium/feat_add_set_theme_source_to_allow_apps_to.patch#L42-L50. This fixes theme to now persist to disk again

## Release notes

Notes: No Notes

### DIFF
--- a/app/src/ui/lib/application-theme.ts
+++ b/app/src/ui/lib/application-theme.ts
@@ -78,11 +78,14 @@ function migrateAutomaticallySwitchSetting(): string | null {
 // in localStorage.
 const applicationThemeKey = 'theme'
 
+/**
+ * Returns User's theme preference or 'system' if not set or parsable
+ */
 function getApplicationThemeSetting(): 'light' | 'dark' | 'system' {
   const themeSetting = localStorage.getItem(applicationThemeKey)
 
   if (themeSetting === null) {
-    return remote.nativeTheme.themeSource
+    return 'system'
   }
 
   if (
@@ -99,7 +102,7 @@ function getApplicationThemeSetting(): 'light' | 'dark' | 'system' {
 /**
  * Load the name of the currently selected theme
  */
- export function getCurrentlyAppliedTheme(): ApplicableTheme {
+export function getCurrentlyAppliedTheme(): ApplicableTheme {
   return isDarkModeEnabled() ? ApplicationTheme.Dark : ApplicationTheme.Light
 }
 
@@ -110,14 +113,10 @@ export function getPersistedThemeName(): string {
   const setting = migrateAutomaticallySwitchSetting()
 
   if (setting === 'system') {
-    remote.nativeTheme.themeSource = setting
     return setting
   }
 
-  const themeSetting = getApplicationThemeSetting()
-  remote.nativeTheme.themeSource = themeSetting
-
-  return themeSetting
+  return getApplicationThemeSetting()
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tidy-dev (changes requested):
@say25 Thanks for looking at this! I was testing it out and am running into some odd behavior. 

I spun up my Desktop-dev. I was in dark mode, which is my system theme setting.. but I have been setting it to light for cherry picking testing. (But due to the bug this pr is fixing, it doesn't remember that and goes back to Dark on each reload)

I then go to appearance. I expected to be set on system but I was on Dark. I figure this is just because I was last on dark before the update.

I then set my theme to light. It switches to light great! Then I close out Desktop-dev and spin it back up again. I expect it to remember my light mode selection. Sadness... it did not persist my selection. It is Dark again. However... I go to the preferences and light is selected just the light theme is not applied! I think this PR is heading the right direction, but needs one more step to apply locally stored theme. 

image

sergiou87:
Could that be some kind of race condition? Maybe we need to set the `themeSource` before/after some specific event 🤔

app/src/ui/lib/application-theme.ts
```diff
@@ -78,32 +78,48 @@ function migrateAutomaticallySwitchSetting(): string | null {
...
+  }
+
+  return 'system'
 }
 
 /**
  * Load the name of the currently selected theme
  */
-export function getCurrentlyAppliedTheme(): ApplicableTheme {
+ export function getCurrentlyAppliedTheme(): ApplicableTheme {
```
sergiou87: Nit (sorry 😆 ): there is a space before `export`
```suggestion
export function getCurrentlyAppliedTheme(): ApplicableTheme {
```

app/src/ui/lib/application-theme.ts
```diff
@@ -78,32 +78,48 @@ function migrateAutomaticallySwitchSetting(): string | null {
...
+  if (setting === 'system') {
+    remote.nativeTheme.themeSource = setting
+    return setting
+  }
+
+  const themeSetting = getApplicationThemeSetting()
+  remote.nativeTheme.themeSource = themeSetting
+
+  return themeSetting
+}
```
sergiou87, line 120: I'm personally not comfortable having a `get` function that also changes things 😕 

This might be redundant, but I'd rather have in `app-store.ts` line 1471 this:
```
    this.selectedTheme = getPersistedTheme()
    // Make sure the persisted theme is applied
    setPersistedTheme(this.selectedTheme)

    this.currentTheme = getCurrentlyAppliedTheme()
```

Or instead of `setPersistedTheme` add a `applyTheme` function which changes `remote.nativeTheme.themeSource` but doesn't persist it? (if we wanted to decouple persistence from applying the theme).

What do you think?

sergiou87 (changes requested):

app/src/ui/lib/application-theme.ts
say25: I think setPersistedTheme is a little cleaner, going with that

sergiou87 (approved):
LGTM and seems to work fine now 😌 

Thanks for taking care of this! ❤️ 

:shipit:

tidy-dev (approved):

PR #11827, opened 2021-03-19
Cherry picking overlay drag prompt

tidy-dev:
Part of #1685 

## Description

Improves discoverability of cherry pick drag target of branch menu by having an overlay of the main content area. Also, after 2.5 seconds of not dragging to the target gives a tool tip type prompt.

### Screenshots
EDIT: The text-align left to center regression that can be seen in multiple selection has been fixed since this video.

EDIT: updated tool tip look since video.
image

https://user-images.githubusercontent.com/75402236/111718562-853b5d00-8830-11eb-931f-10bcab0d5823.mp4

## Release notes
Notes: Improve discoverability of cherry pick drag target of branch menu.

### DIFF
--- a/app/src/ui/drag-overlay.tsx
+++ b/app/src/ui/drag-overlay.tsx
@@ -2,8 +2,10 @@ import * as React from 'react'
 import { dragAndDropManager } from '../lib/drag-and-drop-manager'
 import { PopoverCaretPosition } from './lib/popover'
 
+// time till we prompt the user about where to drag in seconds
+const dragPromptWaitTime = 2500
 interface IDragOverlayProps {
-  readonly dropZoneDescription: string
+  readonly dragZoneDescription: string
 }
 
 interface IDragOverlayState {
@@ -31,8 +33,8 @@ export class DragOverlay extends React.Component<
   }
 
   /** If drop zone is entered, hide drag prompts */
-  private dropZoneEntered = (dropZoneDescription: string) => {
-    if (this.props.dropZoneDescription === dropZoneDescription) {
+  private dragZoneEntered = (dropZoneDescription: string) => {
+    if (this.props.dragZoneDescription === dropZoneDescription) {
       this.clearDragPromptTimeOut()
       this.setState({ showDragPrompt: false })
     }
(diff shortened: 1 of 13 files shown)


tidy-dev:
@say25 Hey! @sergiou87 is out today. If you are around/available, I would appreciate your review on this from a technical standpoint as I would like to get it into beta today. Also don't feel at all obligated to do so - totally understand if you don't have time.

app/styles/ui/_popover.scss
```diff
@@ -40,6 +40,31 @@
...
+    border: 8px solid transparent;
+    border-bottom-color: var(--box-border-color);
+  }
+
+  &::after {
+    top: -14px;
+    margin-right: -8px;
+    border: 7px solid transparent;
+    border-bottom-color: var(--background-color);
+  }
```
say25, line 65: Part of me thinks we could unify a lot of this between the caret variants but I also see the benefits of keeping them separated

ampinsk:
This looks great! Could we just decrease the padding in the tooltip a bit? This is how it looks in the designs for reference: 

Screen Shot 2021-03-19 at 10 01 37 AM

app/src/ui/drag-overlay.tsx
```diff
@@ -0,0 +1,73 @@
...
+      this.setState({ showDragPrompt: true })
+    }, dragPromptWaitTime)
+    dragAndDropManager.onEnterDragZone(this.dragZoneEntered)
+  }
+
+  public componentWillUnmount = () => {
+    this.clearDragPromptTimeOut()
+  }
+
+  private renderDragPrompt() {
```
say25:
```suggestion
  private renderDragPrompt(): JSX.Element | null {
```

might be worth adding types

say25 (dismissed):

app/src/ui/drag-overlay.tsx
tidy-dev: Thank you! I much prefer typing the methods.. tho seems an inconsistent convention in this code base.

app/styles/ui/_popover.scss
tidy-dev: Yeah. I see what you are getting at; But, I think it would require potentially creating popover-caret class as distinct from the placement class and I am not inclined to tear that apart with this pr.

billygriffin (approved):

ampinsk (approved):

PR #11834, opened 2021-03-19
Cherry Picking - Allow switching tab in branch menu during drag

tidy-dev:
Part of #1685

## Description

During usability testing, we caught a scenario where if a user was just on the pr tab in the branch menu and then dragged up to open the branch menu, it would be on the pr tab. There was nowhere for the user to drop the commits on. 

This allows for the user to be able to switch to the branches tab by hovering over the tab selector for ~1.5s.~ 500ms.

### Screenshots

 500ms

https://user-images.githubusercontent.com/75402236/111973375-cce21300-8ad4-11eb-9e0c-c4b56da49057.mp4

## Release notes
Notes: [Fixed] Prevent user from being trapped on PR tab during cherry picking drag.

### DIFF
--- a/app/src/lib/drag-and-drop-manager.ts
+++ b/app/src/lib/drag-and-drop-manager.ts
@@ -9,9 +9,13 @@ import { Disposable, Emitter } from 'event-kit'
  * drag event.
  */
 export class DragAndDropManager {
-  public isDragInProgress: boolean = false
+  private _isDragInProgress: boolean = false
 
   protected readonly emitter = new Emitter()
+
+  public get isDragInProgress(): boolean {
+    return this._isDragInProgress
+  }
 
   public emitEnterDropTarget(targetDescription: string) {
     this.emitter.emit('enter-drop-target', targetDescription)
@@ -32,11 +36,11 @@ export class DragAndDropManager {
   }
 
   public dragStarted(): void {
-    this.isDragInProgress = true
+    this._isDragInProgress = true
   }
 
   public dragEnded() {
-    this.isDragInProgress = false
+    this._isDragInProgress = false
   }
 
   public emitEnterDragZone(dropZoneDescription: string) {
--- a/app/src/lib/drag-and-drop-manager.ts
+++ b/app/src/lib/drag-and-drop-manager.ts
@@ -9,6 +9,8 @@ import { Disposable, Emitter } from 'event-kit'
  * drag event.
  */
 export class DragAndDropManager {
+  public isDragInProgress: boolean = false
+
   protected readonly emitter = new Emitter()
 
   public emitEnterDropTarget(targetDescription: string) {
@@ -28,4 +30,14 @@ export class DragAndDropManager {
   public onLeaveDropTarget(fn: () => void): Disposable {
     return this.emitter.on('leave-drop-target', fn)
   }
+
+  public dragStarted(): void {
+    this.isDragInProgress = true
+  }
+
+  public dragEnded() {
+    this.isDragInProgress = false
+  }
 }
+
+export const dragAndDropManager = new DragAndDropManager()
--- a/app/src/ui/tab-bar.tsx
+++ b/app/src/ui/tab-bar.tsx
@@ -1,5 +1,9 @@
 import * as React from 'react'
 import classNames from 'classnames'
+import { dragAndDropManager } from '../lib/drag-and-drop-manager'
+
+/** Time to wait for drag element hover before switching tabs */
+const dragTabSwitchWaitTime = 1500
 
 /** The tab bar type. */
 export enum TabBarType {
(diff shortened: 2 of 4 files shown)


app/src/ui/tab-bar.tsx
```diff
@@ -22,6 +26,9 @@ interface ITabBarProps {
 
   /** The type of TabBar controlling its style */
   readonly type?: TabBarType
+
+  /** Navigate via drag over */
+  readonly allowDragOverSwitching?: boolean
```
say25, line 31: Is there a reason to make this nullable? Seems like false might suffice as a default?

say25 (dismissed):

app/src/ui/tab-bar.tsx
tidy-dev: Reason I have it optional is so I do not have to specify it 13 places it is used, and AFAIK you can't set a default value on an interface.

ampinsk:
This makes sense to me! The time to hover on the branch tab could be shorter, maybe 500ms?

Abhishek5772705 (approved):

Abhishek5772705 (dismissed):
...

app/src/lib/drag-and-drop-manager.ts
```diff
@@ -9,6 +9,8 @@ import { Disposable, Emitter } from 'event-kit'
  * drag event.
  */
 export class DragAndDropManager {
+  public isDragInProgress: boolean = false
```
sergiou87: If this is public, that means it can also be changed from outside, right?

Maybe we should have a private mutable property, and a public getter?

app/src/ui/tab-bar.tsx
```diff
@@ -88,6 +96,31 @@ export class TabBar extends React.Component<ITabBarProps, {}> {
...
 
+  /**
+   * If something is being dragged, this allows for tab selection by hovering
+   * over a tab for a few seconds (dragTabSwitchWaitTime)
+   */
+  private onMouseEnter = (index: number) => {
+    if (
+      index === this.props.selectedIndex ||
+      !dragAndDropManager.isDragInProgress ||
+      this.props.allowDragOverSwitching === undefined ||
```
sergiou87, line 107: I wonder if this should be `true` by default. That way the user could learn they can switch between tabs from anywhere. If I dragged the commits from the History tab over to the Changes tab and it didn't work, maybe I wouldn't try to do the same inside the branch dropdown.

What do you think?

tidy-dev:
> This makes sense to me! The time to hover on the branch tab could be shorter, maybe 500ms?

@ampinsk Done, and definitely easier to use, but not so easy enough that I think you would do it on accident.

sergiou87 (changes requested):
Code looks good! I only have a couple of comments 😄

app/src/ui/tab-bar.tsx
tidy-dev: :) I had the exact opposite thought. I originally had it open for anywhere, but then I was like.. we would not want the user to be able to navigate somewhere that they can't do anything with commits (Changes tab) (similar concept to instituting the drag overlay - narrow the users possibilities to their target). However, to your point... I wondered if it would make any sense to style the tabs differently to indicate dragging over them does something... or highlight them when you drag over them.. but that may feel like you can drop on them.

@ampinsk thoughts?

app/src/lib/drag-and-drop-manager.ts
tidy-dev: done

app/src/ui/tab-bar.tsx
billygriffin: I think I agree with @tidy-dev on this one. There's nothing for you to do in the "Changes" tab, so you shouldn't be able to drag commits there. I think it would be jarring and unexpected to end up there, and one of our core tenets of Desktop is to help people get to where they need to go and avoid unintended mishaps.

sergiou87: @billygriffin I totally agree with that. My concern is more about the learning process: trying to switch to the "Changes" tab while you're dragging commits (because at that point you don't know where to drop them), you see it doesn't work but don't know why, so you might think that's something you can't do anywhere in the app. Specially since you need to keep the mouse over the element for a few milliseconds: if you see it doesn't work in one place and you're not patient enough when you try in the Branches/PRs tab in the dropdown, you might think it doesn't work there either.

IMHO, it also makes the UI feel inconsistent (in terms of behavior): I can't switch to "Changes" because there is nothing for me to do there, but I can switch from "Branches" to "Pull Requests"? (that's how the current implementation works, if I read it correctly)

> I wondered if it would make any sense to style the tabs differently to indicate dragging over them does something... or highlight them when you drag over them.. but that may feel like you can drop on them.

I think _any_ kind of positive/negative feedback would work: if we disable switching tabs to "Changes" but we somehow inform the user, they might understand it's something specific to that tab.

sergiou87 (approved):
Thanks for those changes! I think after our discussion today this should be good to go 😄 

:shipit:

PR #11846, opened 2021-03-22
Suggest macOS users to move the app to `/Applications` to prevent issues

sergiou87:
## Description

We've noticed ~20% of our macOS users are launching the app from a path different than `/Applications`, and we believe that is causing issues when they try to sign in, where the browser is unable to call the app back with the user's token.

In order to prevent that, this PR uses [some Electron APIs](https://www.electronjs.org/docs/api/app#appmovetoapplicationsfolderoptions-macos) to detect this scenario and even move the app from wherever it is to the `/Application` folder, immediately restarting the app from its new location.

### Screenshots

image

## Release notes

Notes: [Improved] Suggest macOS users to move the app to the Applications folder to prevent issues signing in

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -134,6 +134,7 @@ import { DragElementType } from '../models/drag-element'
 import { CherryPickCommit } from './drag-elements/cherry-pick-commit'
 import classNames from 'classnames'
 import { dragAndDropManager } from '../lib/drag-and-drop-manager'
+import { MoveToApplicationsFolder } from './move-to-applications-folder'
 
 const MinuteInMilliseconds = 1000 * 60
 const HourInMilliseconds = MinuteInMilliseconds * 60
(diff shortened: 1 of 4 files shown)


app/src/ui/app.tsx
```diff
@@ -309,6 +310,11 @@ export class App extends React.Component<IAppProps, IAppState> {
 
     log.info(`launching: ${getVersion()} (${getOS()})`)
     log.info(`execPath: '${process.execPath}'`)
+
+    // Only show the popup in beta/production releases
```
tidy-dev:
```suggestion
    // Only show the popup in beta/production releases and mac machines
```

tidy-dev (commented):
The comment I left is just a minor nit because when I first read that bit of code it confused me for a sec.. (didn't see the `? ` and was like what about windows machines)

tidy-dev (dismissed):

tidy-dev (approved):
Looks good!

PR #11857, opened 2021-03-24
Cherry-picking non existant file error

tidy-dev:
## Description

Parse and reformat conflict (rename/modify) error to a friendly message. This error occurs if a user attempts to cherry-pick a commit with modifications to files that do not exist on the target branch.

### Screenshots

image

## Release notes
Notes: [Improved] Provide a friendly error when user cherry-picks modifications to files non-existent target branch.

### DIFF
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -190,6 +190,7 @@ function parseCherryPickResult(result: IGitResult): CherryPickResult {
   }
 
   switch (result.gitError) {
+    case GitError.ConflictModifyDeletedInBranch:
     case GitError.MergeConflicts:
       return CherryPickResult.ConflictsEncountered
     case GitError.UnresolvedConflicts:
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -19,6 +19,10 @@ import byline from 'byline'
 import { ICherryPickSnapshot } from '../../models/cherry-pick'
 import { ManualConflictResolution } from '../../models/manual-conflict-resolution'
 import { stageManualConflictResolution } from './stage'
+import {
+  parseConflictModifyDeleteCommitSummary,
+  parseConflictModifyDeleteFiles,
+} from '../../ui/lib/parse-conflict-modify-delete-files'
 
 /** The app-specific results from attempting to cherry pick commits*/
 export enum CherryPickResult {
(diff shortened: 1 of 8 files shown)


app/src/ui/lib/parse-conflict-modify-delete-files.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return files
+}
+
+/**
+ * Parse CONFLICT (modify/delete) errors to get commit summary of the commit
+ * that caused the error
+ */
+export function parseConflictModifyDeleteCommitSummary(stdOut: string): string {
+  const versionRegex = /Version\s.*\((.*)\)/
```
sergiou87, line 34: I think this regex causes issues if the commit summary has `()` in it:

image

I think this should work:
```suggestion
  const versionRegex = /Version\s[^(]*\((.*)\)/
```

https://regex101.com/ is very useful to test this stuff 😄 

Another issue I found is if the commit summary has a `Version whatever (...` in it, since that commit summary is also printed _before_ the `Version` you're using as reference:
image

I think we'll need a more complex logic for this parser 😰

app/src/lib/git/cherry-pick.ts
```diff
@@ -187,11 +194,23 @@ function parseCherryPickResult(result: IGitResult): CherryPickResult {
...
 }
 
+function parseConflictModifyDeleteFilesError(result: IGitResult): string {
+  const files = parseConflictModifyDeleteFiles(result.stdout)
+  const commitSummary = parseConflictModifyDeleteCommitSummary(result.stdout)
+  let error = `Cherry-pick failed. The commit with summary '${commitSummary}' modifies one or more files that does not exist on the target branch.`
+  files.forEach(file => {
+    error = error + '\n' + file
+  })
+  return error
```
sergiou87, line 212: If we assume at this point `files` won't be empty (that's the whole reason for this PR, right? 🤔 ) I guess we can simplify the logic like this:

```suggestion
  let error = `Cherry-pick failed. The commit with summary '${commitSummary}' modifies one or more files that does not exist on the target branch.`
  return error + '\n' + files.join('\n')
```

sergiou87 (changes requested):
I found a couple regex-related bugs, and also posted a suggestion.

I think these two new functions are good candidates to be unit tested (and use that to try and protect ourselves against bugs like the ones I pointed out).

app/src/ui/lib/parse-conflict-modify-delete-files.ts
tidy-dev: Awesome catches and yep I love that site.

PR #11866, opened 2021-03-25
Cherry picking non existant file conflict

tidy-dev:
Part of #1685

## Description

Make it so when user encounters conflict error when modifying a file that doesn't exist on the target branch, we give the user the option to either add the modified file or not include the file.

### Screenshots
image

## Release notes
Notes: [Improved] Show conflict dialog when user cherry-picks modifications to files non-existent target branch.

### DIFF
--- a/app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
+++ b/app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
@@ -128,7 +128,7 @@ export class CherryPickConflictsDialog extends React.Component<
 
     const {
       manualResolutions,
-      targetBranchName: theirBranch,
+      targetBranchName: ourBranch,
     } = step.conflictState
 
     return (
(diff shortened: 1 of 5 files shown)


app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
```diff
@@ -128,7 +128,7 @@ export class CherryPickConflictsDialog extends React.Component<
 
     const {
       manualResolutions,
-      targetBranchName: theirBranch,
+      targetBranchName: ourBranch,
```
sergiou87, line 131: Is this change correct in all cases? I ask because my brain always 🤯  when I try to think about their vs our branch in git conflicts 😂

sergiou87 (commented):
Only one question, other than that is quite an elegant approach to the original issue. Great job!!! 💖

app/src/ui/cherry-pick/cherry-pick-conflicts-dialog.tsx
tidy-dev: :D - Yes. I had them backwards 🤦  from the start because rebase requires them opposite and I was following it as a pattern and was a little confused by them so I thought I just didn't understand them at the time.

tidy-dev: Btw, AFAIK, the only other cherry-picking scenario is a regular conflict and I did test that the wording made sense for the outcome that followed. :)

sergiou87 (approved):
LGTM! :shipit:

PR #11875, opened 2021-03-26
Prevent quick click bad drag state during cherry-picking

tidy-dev:
Part of #1685

## Description

This is to mitigate the timing bug that could occur if the user clicked between commits quickly invoking the drag technically twice, but due to timing the mouse up event was not clearing out the  mouse move ent.

This also just improves feel of drag as it doesn't invoke drag immediately. User must drag for 100 ms to show intention to drag before seeing drag indicators. (before user could just click with a little bit of mouse movement and see drag indicators flash)

(video of bug that hopefully is resolved.. if not resolved it is reduced)
https://user-images.githubusercontent.com/75402236/112670929-eddd9780-8e37-11eb-8458-5555a9e4183e.mov

## Release notes

Notes: Prevent quick click bad state perpetual drag state

### DIFF
--- a/app/src/ui/lib/draggable.tsx
+++ b/app/src/ui/lib/draggable.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react'
 import { dragAndDropManager } from '../../lib/drag-and-drop-manager'
 import { mouseScroller } from '../../lib/mouse-scroller'
+import { sleep } from '../../lib/promise'
 
 interface IDraggableProps {
   /**
@@ -40,6 +41,7 @@ export class Draggable extends React.Component<IDraggableProps> {
   // dragElement then elemBelow will always return the dragElement and cannot
   // detect drop targets or scroll elements.
   private verticalOffset: number = __DARWIN__ ? 32 : 15
+  private hasMouseUpOccurred = false
 
   public componentDidMount() {
     this.dragElement = document.getElementById('dragElement')
@@ -66,14 +68,19 @@ export class Draggable extends React.Component<IDraggableProps> {
    * - clears variables from last drag
    * - sets up mouse move and mouse up listeners
    */
-  private onMouseDown = (event: React.MouseEvent<HTMLDivElement>) => {
+  private onMouseDown = async (event: React.MouseEvent<HTMLDivElement>) => {
     if (!this.canDragCommit(event)) {
       return
     }
-    this.initializeDrag()
-
-    document.addEventListener('mousemove', this.onMouseMove)
+    this.hasMouseUpOccurred = false
     document.onmouseup = this.onMouseUp
+    await sleep(100)
+    if (this.hasMouseUpOccurred) {
+      return
+    }
+
+    this.initializeDrag()
+    document.addEventListener('mousemove', this.onMouseMove)
   }
 
   /**
@@ -84,6 +91,10 @@ export class Draggable extends React.Component<IDraggableProps> {
    * just clicking a draggable element.
    */
   private onMouseMove = (moveEvent: MouseEvent) => {
+    if (this.hasMouseUpOccurred) {
+      this.onDragEnd()
+      return
+    }
     // start drag
     if (!this.dragStarted) {
       this.props.onRenderDragElement()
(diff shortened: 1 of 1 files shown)


app/src/ui/lib/draggable.tsx
```diff
@@ -66,14 +68,19 @@ export class Draggable extends React.Component<IDraggableProps> {
...
+    this.hasMouseUpOccurred = false
+    document.onmouseup = this.onMouseUp
+    await sleep(100)
+    if (this.hasMouseUpOccurred) {
+      return
+    }
 
+    this.initializeDrag()
     document.addEventListener('mousemove', this.onMouseMove)
-    document.onmouseup = this.onMouseUp
```
sergiou87, line 76: I think what you're looking for here is a `debounce`. There should be one already accessible in the codebase thanks to `lodash`.

This implementation you have here basically ignores all quick clicks until the user keeps the mouse button down for more than 100ms. That's something super easy to model with `debounce`, I think, so you don't need to implement it yourself.

No need to address this right now or even for the upcoming beta/release, but I think it'd be nice if we explored that option and add that tool to our toolbox!

Basically I think it'd be something like this:
```typescript
private debouncedInitializeDrag = null
...
private onMouseDown = (event: React.MouseEvent<HTMLDivElement>) => { ```
    if (!this.canDragCommit(event)) {
      return
    }

	this.debouncedInitializeDrag = debounce(() => this.initializeDrag(), 100)
    document.addEventListener('mousemove', this.onMouseMove)
    document.onmouseup = this.onMouseUp
}
...
  private onMouseMove = (moveEvent: MouseEvent) => {
    if (this.debouncedInitializeDrag !== null) {
      this.debouncedInitializeDrag.cancel()
      this.debouncedInitializeDrag = null
      return
    }
    ...
  }
...
  private onMouseUp = (moveEvent: MouseEvent) => {
    if (this.debouncedInitializeDrag !== null) {
      this.debouncedInitializeDrag.cancel()
      this.debouncedInitializeDrag = null
      return
    }
    ...
  }
...
  private initializeDrag() {
    if (this.debouncedInitializeDrag !== null) {
      this.debouncedInitializeDrag.cancel()
      this.debouncedInitializeDrag = null
      return
    }
    ...
  }
```

app/src/ui/lib/draggable.tsx
```diff
@@ -84,6 +91,10 @@ export class Draggable extends React.Component<IDraggableProps> {
    * just clicking a draggable element.
    */
   private onMouseMove = (moveEvent: MouseEvent) => {
+    if (this.hasMouseUpOccurred) {
+      this.onDragEnd()
+      return
+    }
```
sergiou87, line 97: Is this necessary? Right now `document.addEventListener('mousemove', this.onMouseMove)` is invoked _after_ the `await sleep(100)`

sergiou87 (commented):
LGTM overall! I have a suggestion and a question.

app/src/ui/lib/draggable.tsx
tidy-dev: I didn't think so... but it seems to help. :/

tidy-dev: It was the first thing I did to basically say if mouse move event is happening after mouse down, then close it out. Then I did the debounce thing and thought I shouldn't need it.. but was still able to reproduce.. so I added it back and couldn't reproduce again.. But.. that said it was hard to reproduce anyways.

tidy-dev: I think it is because the 100ms is just delaying the bug timing. Thus, the bug could still be there if you hold your mouse right around 100ms before doing another click.

app/src/ui/lib/draggable.tsx
tidy-dev: Interesting - I will look into it.

tidy-dev: I tried this.. and it doesn't seem to work.. debugger shows that it never invokes the method.
I originally had `_.debounce(()=>this.initializeDrag, 100)` as I think it should be for scoping only switched to picture to make sure that wasn't something weird. I double checked the wait time is in milliseconds. 

image

tidy-dev: Also that this.debouncedDrag.cancel() crashes the app

sergiou87: Sorry, I'm not familiar with how `debounce` works in JS/TS, only with the operation itself since I used it a lot in previous codebases, I just took a look at the API and it seemed the right way to use it 😅 

Again, that doesn't need to go in this PR, so don't worry!! 😄

sergiou87 (approved):
Tested it and I couldn't break it! :shipit:

PR #11877, opened 2021-03-26
Cherry-picking use main-line 1 for merge commits

tidy-dev:
Part of #1685 

## Description
Merge commits do not always show their history inline on the branch. Thus, in order to cherry-pick something that was merged in and history not shown. User must cherry-pick a merge commit to get those changes. If the commits were in the branches history, the merge commit just becomes an empty commit following the pattern of conflicts or redundant commits. This also means, that if a user has to cherry-pick multiple things that were merged, they can do them all at the the same time instead of selecting only non-merge commits.

### Screenshots
cherry-pick-merge-commits

## Release notes
Notes: [Add] User can cherry-pick merge commits

### DIFF
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -167,9 +167,14 @@ export async function cherryPick(
   }
 
   // --keep-redundant-commits follows pattern of making sure someone cherry
-  // picked commit summaries appear in target branch history even tho they may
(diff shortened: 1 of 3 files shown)


sergiou87 (approved):
LGTM! I needed some help in Slack to understand all this crazy sorcery, but all good now ✨ 🤣 :shipit:

PR #11899, opened 2021-03-30
Diff expansion, part I: add dummy hunk at the end of the file

sergiou87:
## Description

Another step towards diff expansion! This PR adds a hunk header at the bottom of the diff that the user will be able to click to expand the diff from the bottom. In order to do that, and to allow diff expansion properly, I needed to add the diff to the state of the text-diff component. All this is hidden by a feature flag.

As a bonus, I also fixed an annoying warning where we had checks for the existence of a function `onCopy` that is always defined.

### Screenshots

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/syntax-highlighting/index.ts
+++ b/app/src/ui/diff/syntax-highlighting/index.ts
@@ -28,8 +28,8 @@ interface ILineFilters {
 
 interface IFileContents {
   readonly file: ChangedFile
-  readonly oldContents: Buffer
-  readonly newContents: Buffer
+  readonly oldContents: string
+  readonly newContents: string
 }
 
 interface IFileTokens {
@@ -120,7 +120,11 @@ export async function getFileContents(
     }),
   ])
 
-  return { file, oldContents, newContents }
+  return {
+    file,
+    oldContents: oldContents.toString('utf8'),
+    newContents: newContents.toString('utf8'),
+  }
 }
 
 /**
@@ -184,7 +188,7 @@ export async function highlightContents(
 
   const [oldTokens, newTokens] = await Promise.all([
     highlight(
-      oldContents.toString('utf8'),
+      oldContents,
       Path.basename(oldPath),
       Path.extname(oldPath),
       tabSize,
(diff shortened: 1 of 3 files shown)


app/src/ui/diff/syntax-highlighting/index.ts
```diff
@@ -120,7 +120,11 @@ export async function getFileContents(
     }),
   ])
 
-  return { file, oldContents, newContents }
+  return {
+    file,
+    oldContents: oldContents.toString('utf8'),
+    newContents: newContents.toString('utf8'),
```
sergiou87, line 126: This conversion from `Buffer` to `utf8` string was done anyway, I just moved it here for convenience.

app/src/ui/diff/text-diff.tsx
```diff
@@ -340,44 +350,84 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
+
+    const currentDiff = this.state.diff
+    const newDiff = enableTextDiffExpansion()
+      ? getTextDiffWithBottomDummyHunk(
+          currentDiff,
+          currentDiff.hunks,
+          oldContentLines.length,
+          newContentLines.length
+        )
+      : null
```
sergiou87, line 414: This is the interesting part: `getTextDiffWithBottomDummyHunk` will make a new diff with an extra hunk at the end which only has 1 line, needed to create a clickable area that the user can use to expand the diff from the bottom.

That function returns `null` if the current diff already reaches the bottom of the old/new file.

app/src/ui/diff/text-diff.tsx
```diff
@@ -529,16 +579,14 @@ export class TextDiff extends React.Component<ITextDiffProps, {}> {
...
+      this.onCopy(instance, event)
     }
 
     const items: IMenuItem[] = [
       {
         label: 'Copy',
         action,
-        enabled: this.onCopy && isTextSelected,
+        enabled: isTextSelected,
       },
```
sergiou87, line 590: `onCopy` is always defined, hence why we don't need these checks.

tidy-dev (approved):
✨

PR #11902, opened 2021-03-30
Diff expansion, part III: expand diffs for real now :_)

sergiou87:
## Description

This PR glues every previous PR together and makes diff expansion actually happen :_D

**Important:** in this PR, selecting changes per line doesn't work when the diff is expanded, but it shouldn't affect production/beta releases. That problem is already fixed, I'll submit a PR with the required changes Soon™.

### Screenshots

https://user-images.githubusercontent.com/1083228/112471903-c66dc880-8d6c-11eb-82b8-c8c90dd78359.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -47,6 +47,7 @@ import {
 } from '../../lib/feature-flag'
 import { canSelect } from './diff-helpers'
 import {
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
This is awesome! ✨  Great work! :shipit:

PR #11903, opened 2021-03-30
Create branch during cherry-picking via the context menu

tidy-dev:
Adding to #1685

## Description

Add ability to create a branch during cherry-picking via the context menu.

### Screenshots

https://user-images.githubusercontent.com/75402236/113047818-c9532980-916f-11eb-880d-6c64467e68c7.mp4

## Release notes

Notes: [Added]  User can create a branch during cherry-picking via the context menu.


billygriffin (dismissed):
From a product perspective, I think this is just 💯  Great work!

app/src/lib/stores/app-store.ts
```diff
@@ -3009,14 +3009,17 @@ export class AppStore extends TypedBaseStore<IAppState> {
     repository: Repository,
     name: string,
     startPoint: string | null,
-    noTrackOption: boolean = false
-  ): Promise<void> {
+    noTrackOption: boolean = false,
+    checkoutBranch: boolean = false
```
sergiou87: I couldn't see where this new parameter is used. Is this a leftover?

app/src/lib/stores/app-store.ts
```diff
@@ -3442,7 +3446,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
       // If a local branch, user may have the branch to delete checked out and
       // we need to switch to a different branch (default or recent).
-      const branchToCheckout = this.getBranchToCheckoutAfterDelete(branch, r)
+      const branchToCheckout =
+        toCheckout || this.getBranchToCheckoutAfterDelete(branch, r)
```
sergiou87: Nit: Should we use `??` here? I understand they should be equivalent in this specific case, but `||` does a falsy/truthy check which, in theory, we should avoid 🤔 
```suggestion
        toCheckout ?? this.getBranchToCheckoutAfterDelete(branch, r)
```

sergiou87 (changes requested):
Only found a couple of minor issues. Other than that it looks good and works beautifully well!! 😍

app/src/lib/stores/app-store.ts
tidy-dev: Had to look that one up I had always used/seen `||` before - and yeah `??` makes more sense.

app/src/lib/stores/app-store.ts
tidy-dev: Ugg.. good catch.. that should default to true..

tidy-dev: And start cherry-pick with target branch name in dispatcher should set to false.  (in other words this was a regression)

sergiou87 (approved):
LGTM now!! Thanks for those changes 😄  :shipit: 🚀

sirpenev (commented):
T

PR #11906, opened 2021-03-31
Diff expansion, part IV: fix selection on expanded diffs

sergiou87:
## Description

This PR makes a few changes in the way the app handles line selection. Instead of just using the line number of the change in the current diff (which can be expanded), the app will track the line number in the original diff, and keep it during the multiple expansions of the diff.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #11912, opened 2021-04-01
Ability to change spell check language to English

### LINKED ISSUE #11589: Ability to change spell check language

Currently, the spell-checker seems to use the OS language to set the spellcheck language... but I do use English in my commits, instead of German, which kind of makes the spell-checker useless in my case.

### Proposed solution

A new option (File -> Options) to change the spell-checker language.

tidy-dev:
Closes #11589 

## Description

Add option for a non-macOS users who's system language is not English to set the applications spell checker to English (and also toggle back to the system language).

Note: MacOS natively detects multiple languages and thus does not support manually setting the language.

### Screenshots

Windows with system language set to German.

https://user-images.githubusercontent.com/75402236/113306740-66c66e80-92d2-11eb-984d-e89e4b0657ec.mov

Windows with system language set to English (No need for options)
Screen Shot 2021-04-01 at 9 59 39 AM

MacOS
MacOS natively detects multiple languages. As seen in screen shot, it doesn't show English or German word as misspelled and the English is set as the system language. Thus, no need for another option.
Screen Shot 2021-04-01 at 10 06 18 AM

## Release notes
Notes: [Improved] For non macOS users with system language of not English, user can toggle between English and system language.

### DIFF
--- a/app/src/ui/main-process-proxy.ts
+++ b/app/src/ui/main-process-proxy.ts
@@ -150,11 +150,11 @@ function getSpellCheckLanguageMenuItem(
     return null
   }
 
-  const currentLanguageCodes = session.getSpellCheckerLanguages()
+  const spellcheckLanguageCodes = session.getSpellCheckerLanguages()
 
   const languageCode =
-    currentLanguageCodes.includes(englishLanguageCode) &&
-    !currentLanguageCodes.includes(userLanguageCode)
+    spellcheckLanguageCodes.includes(englishLanguageCode) &&
+    !spellcheckLanguageCodes.includes(userLanguageCode)
       ? userLanguageCode
       : englishLanguageCode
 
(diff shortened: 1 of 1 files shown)


billygriffin:
This seems reasonable to me! I'd like to take another crack at that release note prior to shipping in a release but the approach seems 👍

art1-svg:
well done

sergiou87 (approved):
Code LGTM!! ✨ :shipit:

abstractspoon:
Now we just need an option to control which version of English to use: AUS or UK or USA!

AaronPedersen-QUT:
> Now we just need an option to control which version of English to use: AUS or UK or USA!

It's been 2 years, and they still haven't done anything about this.

Clicking `Set spellcheck to system language` doesn't change the spellcheck to `English (Australia)` (my system language on my Windows OS), but instead keeps it as American English.

Please give us this option, I'm begging. 🙏

PR #11916, opened 2021-04-01
Fix cherry-pick not ended on drop to current branch

### LINKED ISSUE #11915: Cherry picking by drag and drop to the same branch breaks file and diff panes
### Describe the bug

When trying to cherry pick a commit with the newly added drag and drop functionality (#1685) if you drop the commit onto the branch it came from, the files and diff panes break.

### Version & OS

Version 2.7.1,  Windows 10 20H2

1. Create a repo
2. Make some commits
3. Go to History View
4. Drag a commit from history onto branch selector and drop it on current branch
4. The commit details view is now dimmed and cannot be interacted with (red 🚫 cursor)

### Expected behavior

You should be told that the commit you are trying to cherry pick is already on this branch OR that this commit comes from the branch your are picking from.

### Actual behavior

You can't interact with files that were changed in the commit and you cannot interact with diff view because they are now under dimming overlay that blocks mouse interaction. Switching the branch or repo solves the problem.

### Screenshots

image

### Logs

### Additional context

Action after which this but occurs is a bit nonsensical. I discovered the bug because I had only one branch in my repro and I wanted to see what will happen when I drop the commit on it. Somebody could perform this move by accident thus mildly braking the program. 

This is my first bug report ever, sorry if I'm doing something wrong.

tidy-dev:
Closes #11915

## Description

When user cherry-picks onto the current branch, it should do nothing - it is not a valid drop target. 

### Screenshots

https://user-images.githubusercontent.com/75402236/113318948-f6721a00-92de-11eb-8a49-ae46fa4a29d0.mov

## Release notes

Notes: [Fixed] Dropping cherry-pick on current branch ends cherry-picking drag.

### DIFF
--- a/app/src/ui/branches/branch-list-item.tsx
+++ b/app/src/ui/branches/branch-list-item.tsx
@@ -98,10 +98,19 @@ export class BranchListItem extends React.Component<IBranchListItemProps, {}> {
   }
 
   private onMouseUp = () => {
-    const { onDropOntoBranch, name, isCurrentBranch } = this.props
+    const {
+      onDropOntoBranch,
+      onDropOntoCurrentBranch,
+      name,
+      isCurrentBranch,
+    } = this.props
 
     if (onDropOntoBranch !== undefined && !isCurrentBranch) {
       onDropOntoBranch(name)
+    }
+
+    if (onDropOntoCurrentBranch !== undefined && isCurrentBranch) {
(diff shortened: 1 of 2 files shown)


billygriffin (approved):
Logic looks good to me and building the app from this branch it also seems fixed, so 🎉

PR #11921, opened 2021-04-01
Fix cannot undo remote branch cherry-pick

tidy-dev:
## Description

Undo after cherry-picking onto a remote branch did nothing because it was referencing remote branch name and checking against local. Now, it will undo.

### Screenshots

https://user-images.githubusercontent.com/75402236/113360348-9ba8e480-9317-11eb-91e8-1786f502455c.mov

## Release notes

Notes: [Fixed] User can undo a cherry-pick onto a remote branch.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5816,7 +5816,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.emitUpdate()
   }
 
-  private async _cherryPick(
+  /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _cherryPick(
     repository: Repository,
     commits: ReadonlyArray<CommitOneLine>
   ): Promise<CherryPickResult> {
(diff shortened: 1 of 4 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -6093,7 +6093,11 @@ export class AppStore extends TypedBaseStore<IAppState> {
   ): Promise<boolean> {
     const { branchesState } = this.repositoryStateCache.get(repository)
     const { tip } = branchesState
-    if (tip.kind !== TipState.Valid || tip.branch.name !== targetBranchName) {
+    if (
+      tip.kind !== TipState.Valid ||
+      (tip.branch.name !== targetBranchName &&
+        tip.branch.upstream !== targetBranchName)
```
sergiou87: Maybe instead of this, the `targetBranchName` should be the branch name after checking out? 🤔

tidy-dev: Hmm.. I refactored to this approach.. not really sure it is better tho. 🙂  It technically opens us up to the very slight possibility that if someone switches branches (with command line) at the exact same time cherry-pick is completing.. we could do a reset on the wrong branch... it would also show wrong branch in success bar... and the only way to prevent that scenario would to be do the exact same check I am trying not to do in a different place. 😛  Also.. not sure if it is better or not, but success bar now shows `branchName` as opposed to `origin/branchName`.  Thoughts?

tidy-dev: Based on more feedback.. I refactored again 😄 .... maybe this is more along the lines of what you are suggesting via slack. I like this solution better as it halts the whole cherry-pick process if something is wrong with the target branch being checked out. Also, separates concerns which is nice.

Technically still has the scenario of someone checking out another branch during the cherry-pick via command line..but this risk was already present and just means they would cherry-pick onto a different branch (which they could undo as opposed to hard reseting commits on the wrong branch which could be irrevocable loss of work).

sergiou87: I personally prefer `branchName` over `origin/branchName` 🤔

sergiou87 (approved):
LGTM, thanks for those changes! :shipit:

yuu1111yuu (commented):
...

PR #11926, opened 2021-04-02
Split Whitespace Settings

say25:
Follow up from #6818

## Description

- Decouples the hide whitespace changes in Changes and History Tabs

## Release notes

Notes: No notes (covered in #6818)

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4663,14 +4663,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return Promise.resolve()
   }
 
-  public async _setHideWhitespaceInChangesDiff(
+  public _setHideWhitespaceInChangesDiff(
     hideWhitespaceInDiff: boolean,
     repository: Repository
   ): Promise<void> {
     setBoolean(hideWhitespaceInChangesDiffKey, hideWhitespaceInDiff)
     this.hideWhitespaceInChangesDiff = hideWhitespaceInDiff
 
-    await this.refreshChangesSection(repository, {
+    return this.refreshChangesSection(repository, {
       includingStatus: true,
       clearPartialState: true,
     })
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4670,12 +4670,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
     setBoolean(hideWhitespaceInChangesDiffKey, hideWhitespaceInDiff)
     this.hideWhitespaceInChangesDiff = hideWhitespaceInDiff
 
-    if (enableHideWhitespaceInDiffOption()) {
-      await this.refreshChangesSection(repository, {
-        includingStatus: true,
-        clearPartialState: true,
-      })
-    }
+    await this.refreshChangesSection(repository, {
+      includingStatus: true,
+      clearPartialState: true,
+    })
   }
 
   public async _setHideWhitespaceInHistoryDiff(
@@ -4685,13 +4683,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
   ): Promise<void> {
     setBoolean(hideWhitespaceInHistoryDiffKey, hideWhitespaceInDiff)
     this.hideWhitespaceInHistoryDiff = hideWhitespaceInDiff
-
-    if (enableHideWhitespaceInDiffOption()) {
-      await this.refreshChangesSection(repository, {
-        includingStatus: true,
-        clearPartialState: true,
-      })
-    }
 
     if (file === null) {
       return this.updateChangesWorkingDirectoryDiff(repository)
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -316,8 +316,10 @@ const externalEditorKey: string = 'externalEditor'
 const imageDiffTypeDefault = ImageDiffType.TwoUp
 const imageDiffTypeKey = 'image-diff-type'
 
-const hideWhitespaceInDiffDefault = false
-const hideWhitespaceInDiffKey = 'hide-whitespace-in-diff'
+const hideWhitespaceInChangesDiffDefault = false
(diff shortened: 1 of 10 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -316,8 +316,10 @@ const externalEditorKey: string = 'externalEditor'
 const imageDiffTypeDefault = ImageDiffType.TwoUp
 const imageDiffTypeKey = 'image-diff-type'
 
-const hideWhitespaceInDiffDefault = false
-const hideWhitespaceInDiffKey = 'hide-whitespace-in-diff'
+const hideWhitespaceInChangesDiffDefault = false
+const hideWhitespaceInChangesDiffKey = 'hide-whitespace-in-changes-diff'
+const hideWhitespaceInHistoryDiffDefault = false
+const hideWhitespaceInHistoryDiffKey = 'hide-whitespace-in-diff'
```
say25, line 322:
```suggestion
const hideWhitespaceInHistoryDiffKey = 'hide-whitespace-in-history-diff'
```

I did want to do something like this but, then you'd loose your setting value when upgrading

app/src/ui/history/selected-commit.tsx
```diff
@@ -193,7 +193,7 @@ export class SelectedCommit extends React.Component<
   private onHideWhitespaceInDiffChanged = async (
```
say25: Maybe this function name should change but I sorta think not 🤔

app/src/ui/changes/changes.tsx
```diff
@@ -130,7 +130,7 @@ export class Changes extends React.Component<IChangesProps, {}> {
   private onHideWhitespaceInDiffChanged = async (
```
say25: Maybe this function name should change but I sorta think not 🤔

say25:
Other thoughts, should I change the the toggle to be for Beta instead of Development as part of this PR?

app/src/lib/stores/app-store.ts
tidy-dev: Makes sense

app/src/ui/changes/changes.tsx
tidy-dev: Makes sense to me to leave it since we are in the `changes.tsx` .. `inChanges` would be redundant.

app/src/ui/history/selected-commit.tsx
tidy-dev: Ditto

app/src/lib/stores/app-store.ts
```diff
@@ -4652,13 +4663,28 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
 
-  public async _setHideWhitespaceInDiff(
+  public async _setHideWhitespaceInChangesDiff(
+    hideWhitespaceInDiff: boolean,
+    repository: Repository
+  ): Promise<void> {
+    setBoolean(hideWhitespaceInChangesDiffKey, hideWhitespaceInDiff)
+    this.hideWhitespaceInChangesDiff = hideWhitespaceInDiff
+
+    if (enableHideWhitespaceInDiffOption()) {
```
tidy-dev: Curious if we need the feature flag check here. Wouldn't this only have impact if the user has access to the setting in the UI (which is feature flagged)?

app/src/lib/stores/app-store.ts
```diff
@@ -2198,7 +2209,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const diff = await getWorkingDirectoryDiff(
       repository,
       selectedFileBeforeLoad,
-      enableHideWhitespaceInDiffOption() && this.hideWhitespaceInDiff
+      enableHideWhitespaceInDiffOption() && this.hideWhitespaceInChangesDiff
```
tidy-dev, line 2212: Won't `hideWhitespaceInChangesDiff` be false if feature flag is false? (Is check needed here since we feature flag the entry point in UI?)

tidy-dev (changes requested):
This looks great! I just have a couple of nits.

tidy-dev:
> Other thoughts, should I change the the toggle to be for Beta instead of Development as part of this PR?

Yep. Go ahead. :)

app/src/lib/stores/app-store.ts
say25: This is the case, but the reason for the flag check here is to allow us to rollback the flag if needed.

app/src/lib/stores/app-store.ts
say25: yea you shouldn't be able to call this function without the UI, I'll remove this check

app/src/lib/stores/app-store.ts
tidy-dev: Ah ok. Cool

app/src/ui/history/selected-commit.tsx
```diff
@@ -193,7 +193,7 @@ export class SelectedCommit extends React.Component<
...
     hideWhitespaceInDiff: boolean
   ) => {
-    await this.props.dispatcher.onHideWhitespaceInDiffChanged(
+    await this.props.dispatcher.onHideWhitespaceInHistoryDiffChanged(
```
sergiou87: Should/Could this be `return` instead of `await`?

app/src/ui/changes/changes.tsx
```diff
@@ -130,7 +130,7 @@ export class Changes extends React.Component<IChangesProps, {}> {
...
     hideWhitespaceInDiff: boolean
   ) => {
-    await this.props.dispatcher.onHideWhitespaceInDiffChanged(
+    await this.props.dispatcher.onHideWhitespaceInChangesDiffChanged(
```
sergiou87: Should/Could this be `return` instead of `await`?

sergiou87 (commented):
LGTM! Only a couple of questions 😄

sergiou87 (dismissed):
Looking good! Thanks for the great work! ❤️

sergiou87 (approved):

tidy-dev (approved):

PR #11935, opened 2021-04-05
Diff expansion, part V: refactor how we compute hunk expansion info

sergiou87:
## Description

This is a refactor needed to make it easier (and hopefully faster) the integration of diff expansion in the split diff mode. Instead of computing whether or not a hunk can be expanded in any direction at render time, we'll do that at parsing (and expanding) time, and store that info within the hunk.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #11936, opened 2021-04-05
Diff expansion, part VI: refactor split diff to use data from state instead of props

sergiou87:
## Description

The changes in this PR are quite simple, they just refactor the side-by-side diff view to grab the diff data from the state instead of doing so from its props. Thanks to this change, it can (in upcoming PRs) modify the diff by expanding it.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -70,7 +70,7 @@ interface ISideBySideDiffProps {
   /** The file whose diff should be displayed. */
   readonly file: ChangedFile
 
-  /** The diff that should be rendered */
+  /** The initial diff */
   readonly diff: ITextDiff
 
   /**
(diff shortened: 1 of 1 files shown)


app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -303,25 +316,41 @@ export class SideBySideDiff extends React.Component<
   }
 
   private async initDiffSyntaxMode() {
-    const { file, diff, repository } = this.props
+    const { file, repository } = this.props
+    const { diff } = this.state
 
-    // Store the current props to that we can see if anything
+    // Store the current props and state to that we can see if anything
```
tidy-dev:
```suggestion
    // Store the current props and state so that we can see if anything
```

tidy-dev (changes requested):
Just a typo that was already there 😸

tidy-dev (approved):
✨

PR #11937, opened 2021-04-05
Diff expansion, part VII: add dummy hunk to allow expanding diffs from the bottom in split view

sergiou87:
## Description

This change adds the dummy hunk at the bottom of the diff so that it can be expanded there in split view.

I noticed that we display `@ @@` instead of `@@ @@`. That's a bug existing in prod that will be fixed in an upcoming PR.

### Screenshots

image

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #11938, opened 2021-04-05
Diff expansion, part VIII: render hunk expansion handlers in split view

sergiou87:
## Description

These changes just add the hunk expansion handlers to the split view. It also fixes the bug I mentioned in the previous PR about a character being deleted in the hunk headers.

### Screenshots

https://user-images.githubusercontent.com/1083228/113584619-6956fa00-962b-11eb-9919-40bbc3c079c3.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -48,7 +48,10 @@ import { getTokens } from './diff-syntax-mode'
 import { DiffSearchInput } from './diff-search-input'
 import { escapeRegExp } from '../../lib/helpers/regex'
 import { enableTextDiffExpansion } from '../../lib/feature-flag'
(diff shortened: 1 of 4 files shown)


app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -916,8 +925,9 @@ function getDiffRowsFromHunk(
     if (line.type === DiffLineType.Hunk) {
       rows.push({
         type: DiffRowType.Hunk,
-        content: line.content,
+        content: line.text,
```
sergiou87, line 928: This is what fixes the issue with the first `@` missing. `line.content` refers to the content of the line in a diff (which means the initial `+` or `-` is removed).

tidy-dev (approved):
✨

PR #11947, opened 2021-04-05
Diff expansion, part X: fix selection of expanded diffs in split view

sergiou87:
## Description

This PR fixes the selection of changes from a expanded diff in split view. Basically, instead of using the line number in the expanded diff, it uses the line number in the original diff (which all added/deleted lines should have).

## Release notes

Notes: no-notes


sergiou87:
This is the last part of the series 🤣 

From now on just fixes, I think!

tidy-dev (approved):
✨

PR #11949, opened 2021-04-05
Use a save dialog instead of an open dialog when choosing where to clone a repo

j-f1:
Closes [N/A]

## Description

Since cloning a repo creates a directory, display a save dialog instead of an open dialog when picking a location to clone to.

### Screenshots

Screenshot_2021-04-05 17 25 02

## Release notes

Notes: Improves the dialog box shown when picking a location to clone a repository to

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -165,3 +165,8 @@ export function enableCherryPicking(): boolean {
 export function enableTextDiffExpansion(): boolean {
   return enableDevelopmentFeatures()
 }
+
+/** Should we allow using the save dialog when choosing where to clone a repo */
+export function enableSaveDialogOnCloneRepository(): boolean {
+  return enableBetaFeatures()
+}
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -24,6 +24,7 @@ import { IAccountRepositories } from '../../lib/stores/api-repositories-store'
 import { merge } from '../../lib/merge'
 import { ClickSource } from '../lib/list'
 import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
+import { enableSaveDialogOnCloneRepository } from '../../lib/feature-flag'
 
 interface ICloneRepositoryProps {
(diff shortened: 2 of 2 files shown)


tidy-dev (commented):
@j-f1 Thank you for this improvement. I tested it out in windows and mac. Looks great! Particularly liked how on my windows testing, it sets the filepath to the path displayed (sans the new folder of course) unlike before where it was setting me to some last known unrelated directory.  I am wondering if it would be easy to also update the "Create a New Repository" to follow suit?

image

j-f1:
Good point! It would probably make sense to put the last component of the selected path in the “Name” field in that case.

j-f1:
I tried that out, but it makes it harder to open a directory to start off a Git repo for an existing project. Do you think that’s worth the tradeoff?

tidy-dev:
@j-f1 Thank you for trying that out. We discussed this a bit and agree that it probably isn't worth the trade-off to add it to the create repository dialog at this time. 

However, we would really like to to get what you have for clone repository out on beta, but we would like to feature-flag it just to be sure we don't get any unexpected behaviors before releasing to prod. Would you be interested and feature-flagging it? (If not, I can totally do it.)

lildixk1000 (dismissed):

j-f1:
I’ve never done a feature-flagged thing before so if you have time to do it, that would be great!

tidy-dev (approved):
✨

sergiou87 (approved):
I love it, thank you!!! 🙇‍♂️ :shipit:

sergiou87:
It looks like these changes caused a regression on Windows: #12812 😢 

We should revisit this and decide whether we just keep the Save dialog for macOS or use Open in both.

X3A3 (commented):
# @ #
Duplicate of #

> [Xnxx](xnx.com)

X3A3:
Xnxx"http

SharakPL:
> Notes: Improves the dialog box shown when picking a location to clone a repository to

@j-f1 I'm afraid it improves nothing. Actually it breaks setting up the path for the repo #12812. Please revert it to _Browse for folder_ prompt, at least on Windows 🙏

PR #11954, opened 2021-04-07
Fix some issues expanding diffs

sergiou87:
## Description

This PR fixes two issues:
1. Right now the app is avoiding to load and highlight the old/new file contents when there are only added lines or only removed lines, but not both kinds of lines. However, for diff expansion we need to at least **load** both file contents, and keep the new file contents around for expansion.
2. When 2 hunks are merged expanding down, the "adjacent" hunk is not updated so its `unifiedDiffEnd` was wrong, and we were using it for the new merged hunk 😓  Here we change the logic to one that should always work.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/text-diff-expansion.ts
+++ b/app/src/ui/diff/text-diff-expansion.ts
@@ -49,12 +49,17 @@ function mergeDiffHunks(hunk1: DiffHunk, hunk2: DiffHunk): DiffHunk {
     false
   )
 
+  const newHunkLines = [
+    newFirstHunkLine,
+    ...allHunk1LinesButFirst,
+    ...allHunk2LinesButFirst,
+  ]
+
   return new DiffHunk(
     newHunkHeader,
-    [newFirstHunkLine, ...allHunk1LinesButFirst, ...allHunk2LinesButFirst],
+    newHunkLines,
     hunk1.unifiedDiffStart,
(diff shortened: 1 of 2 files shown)


app/src/ui/diff/text-diff-expansion.ts
```diff
@@ -49,12 +49,17 @@ function mergeDiffHunks(hunk1: DiffHunk, hunk2: DiffHunk): DiffHunk {
...
+    ...allHunk2LinesButFirst,
+  ]
+
   return new DiffHunk(
     newHunkHeader,
-    [newFirstHunkLine, ...allHunk1LinesButFirst, ...allHunk2LinesButFirst],
+    newHunkLines,
     hunk1.unifiedDiffStart,
-    // This -1 represents the header line of the second hunk that we removed
-    hunk2.unifiedDiffEnd - 1,
```
sergiou87, line 57: https://github.com/desktop/desktop/blob/8daf7ede719fd0d013084ff8ef5bac1930bfe8c9/app/src/ui/diff/text-diff-expansion.ts#L311-L335

This `hunk2` is the `adjacentHunk` you see in line 334. However, whereas `hunk1` had the right `unifiedDiffStart` and `unifiedDiffEnd` after expansion (see lines 315-316), the `adjacentHunk` is never updated like that.

tidy-dev (approved):
:sparkles:

PR #11955, opened 2021-04-07
Allow expanding whole files in diff views

sergiou87:
## Description

This PR adds the ability to expand diffs completely, and then collapse them back to their previous state. All this from the context menu.

### Screenshots

https://user-images.githubusercontent.com/1083228/113877074-a1d80e80-97b8-11eb-9d58-2a939cf3b450.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/syntax-highlighting/index.ts
+++ b/app/src/ui/diff/syntax-highlighting/index.ts
@@ -29,8 +29,8 @@ interface ILineFilters {
 
 interface IFileContents {
   readonly file: ChangedFile
-  readonly oldContents: string
-  readonly newContents: string
+  readonly oldContents: string | null
+  readonly newContents: string | null
 }
 
 interface IFileTokens {
@@ -41,12 +41,12 @@ interface IFileTokens {
 async function getOldFileContent(
   repository: Repository,
   file: ChangedFile
-): Promise<Buffer> {
+): Promise<Buffer | null> {
   if (
     file.status.kind === AppFileStatusKind.New ||
     file.status.kind === AppFileStatusKind.Untracked
   ) {
-    return Buffer.alloc(0)
+    return null
   }
 
   let commitish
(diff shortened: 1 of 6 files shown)


sergiou87:
I think it's better if #11954 is merged before going through this PR 😅

tidy-dev (changes requested):
It appears to work great in history diff.. but not so much in changes diff? Is that coming in another PR (It sometimes collapses in changes tho)? 

All I did was add the word test to end of our readme file to do this. 

https://user-images.githubusercontent.com/75402236/113886572-cae3a980-978e-11eb-8c64-fa74fc65e64a.mov

sergiou87:
Hmmm while working on it, I used the Changes tab all the time. I'll take a look, thank you!

sergiou87:
@tidy-dev fixed! The problem was specifically when the diff only had 1 hunk at the bottom of the file 🤦‍♂️ 

I also fixed another issue where the "diff to restore" was kept around when switching files 😐

tidy-dev (dismissed):
✨ 
Suggestions for further work in this area (if you think they make sense 🙂 )
1 - Disable "Expand Whole File" when nothing to expand (Example: new file creation)
2 - Have "Collapse Expand Line" available if any expanding has happened and/or if user expanded the whole file manually.

sergiou87:
Good point about (1)!!
(2) is a just bit harder to implement. From what I see, dotcom doesn't even support it. If you manually expand a file completely, it will hide the expand/collapse button and that's it. Maybe I can try that out after doing the refactor we discussed yesterday 😄

sergiou87:
@tidy-dev I just implemented (1) and fixed a bug when you delete a file, the app would still allow you to expand it :doh:

app/src/ui/diff/syntax-highlighting/index.ts
```diff
@@ -74,9 +74,9 @@ async function getOldFileContent(
 async function getNewFileContent(
   repository: Repository,
   file: ChangedFile
-): Promise<Buffer> {
+): Promise<Buffer | null> {
   if (file.status.kind === AppFileStatusKind.Deleted) {
-    return Buffer.alloc(0)
```
sergiou87, line 79: Because of this we would get an empty string even when the file was deleted, which counted as a 1-line file that messed up with other calculations and made the app think the file could be expanded 😅

tidy-dev (approved):
Sweet! I like it! Thanks for adding (1). Makes sense on (2).

PR #11958, opened 2021-04-07
Show alert when moving the app to /Applications fails

sergiou87:
## Description

Just what the description says! 😛 

### Screenshots

https://user-images.githubusercontent.com/1083228/113901763-77de1680-97cf-11eb-9d1f-34ae8f8bd5be.mov

## Release notes

Notes: [Improved] Show an alert when the app fails to move itself to the Applications folder on macOS

### DIFF
--- a/app/src/ui/move-to-applications-folder.tsx
+++ b/app/src/ui/move-to-applications-folder.tsx
@@ -57,13 +57,14 @@ export class MoveToApplicationsFolder extends React.Component<
     )
   }
 
-  private onSubmit = async () => {
+  private onSubmit = () => {
+    this.props.onDismissed()
+
     try {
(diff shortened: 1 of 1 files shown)


app/src/ui/move-to-applications-folder.tsx
```diff
@@ -57,13 +57,14 @@ export class MoveToApplicationsFolder extends React.Component<
     )
   }
 
-  private onSubmit = async () => {
+  private onSubmit = () => {
+    this.props.onDismissed()
+
     try {
       this.props.dispatcher.moveToApplicationsFolder()
```
sergiou87, line 64: You can replace this line with `throw new Error('some message')` if you want to test it 😄

tidy-dev (approved):
✨

PR #11964, opened 2021-04-08
Cherry picking snapshot bug

tidy-dev:
## Description

The cherry-picking snapshot/progress did not always accurately reflect the operation during cherry-picks of multiple commits - most noticeable against an old branch and when a user restarted the app during a the cherry-pick.

For those of you who would like to take a trip down mind blown lane after the fact (seems less mind blowing now)...

The goal of cherry-picking snapshot is to be able to provide data for the progress dialog , the success banner, and obtain the sha to which we would reset the target branch too on undo. The information we need to parse from the sequencer is the count of all commits for the operation, how many have been cherry-picked, and the summaries of commits remaining to be cherry-picked. It is used in two places:  on continue and on opening the app during an ongoing cherry-pick. Both of these instances result in a pause in the progress and discontinuity in user interaction (meaning we can just keep the information in state and we need to parse it). Also, based on which one of those, we use different info from it.. thus something that appears to work for one doesn't mean it works for the other or all cases (why this bug is not readily apparent).

On first attempt, I erroneously thought that the commit sha stored in `abort-safety` was the commit you would reset to on abort. (This appears to work in basic scenarios) That is not the case, it is simply the sha of the last commit on the target branch. During a cherry-pick this starts off as the target branch head and then is updated each time a cherry-picked commit is committed. 

Fortunately, I found that the sha stored in `head` is the head of the target branch before the cherry-pick operation and does not change. This is what we want to abort too or undo too. 

Of note: The `todo` file was correctly understood in that it lists all the remaining shas left to cherry-pick in the operation with the next in line being the first line in the file. 

Thus, first major change is switching to that understanding. 

With that .. we now know that doing a `rev-list` (`getCommitInRange`) from the `abort-safety` to last todo sha made no sense to obtain all the commits for the operation as the `abort-safety` was changing with each commit. Just on that our total count would get smaller with each continued cherry-pick.

But to really throw us for a loop... If those to commit sha's are very different in terms of files changed and time in relation to the branch being cherry-picked onto. The `rev-list` may collect a bunch of commits that make all the changes to the files modified in the commit sha's given make sense. Therefore, cherry-picking from a relatively new branch to an old branch would have lots of file changes (and lots of commits in relation to those file changes) to make the changes in the referenced commit make sense on the target branch. Thus, during `rev-list` during cherry-picking in reference to sha's not yet cherry-picked sometimes resulted in a very wrong total commits count.
.... This is my understanding from toying messing with it (maybe there is a more technical understanding of this out there)...

I realized.. the reason I was using `rev-list` was to get all the commits for the operation and get the summaries for the remaining commits. In other words, get (1) the commits already cherry-picked (for total count) and (2) the commits remaining to be cherry-picked (for summaries). (2) I already had in the `todo` file - it provides the shas and summarys of all the remaining commits to be cherry-picked. I just need to update my parser to get the summary along with the shas. Yay! 

Now, I was left with getting the already cherry-picked commits. This I can use `rev-list` for from the target branch's `head` to the latest cherry-picked commit aka sha in `abort-safety`. From my understanding, this should always be accurate because we just committed these commits so there is no worry of comparison against an old branch/timeline.

### Screenshots
After fix:

https://user-images.githubusercontent.com/75402236/114029753-d8109f00-9847-11eb-9540-7bc6a62ec42d.mov

Before fix (in case you are curious... no 819 commits were not copied 😸 )

https://user-images.githubusercontent.com/75402236/114029615-af88a500-9847-11eb-9d0b-4d01b4ecaff3.mov

## Release notes
Notes: [Fixed] Cherry-pick progress no longer shows wrong number of commits in some circumstances.

### DIFF
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -83,7 +83,7 @@ class GitCherryPickParser {
       kind: 'cherryPick',
       title: `Cherry-picking commit ${this.count} of ${this.commits.length} commits`,
       value: round(this.count / this.commits.length, 2),
-      cherryPickCommitCount: this.count,
+      position: this.count,
       totalCommitCount: this.commits.length,
       currentCommitSummary: this.commits[this.count - 1]?.summary ?? '',
     }
@@ -319,21 +319,21 @@ export async function getCherryPickSnapshot(
   }
 
   const commits = [...commitsCherryPicked, ...remainingCommits]
-  const count = commitsCherryPicked.length + 1
+  const position = commitsCherryPicked.length + 1
 
   return {
     progress: {
       kind: 'cherryPick',
-      title: `Cherry-picking commit ${count} of ${commits.length} commits`,
-      value: round(count / commits.length, 2),
-      cherryPickCommitCount: count,
+      title: `Cherry-picking commit ${position} of ${commits.length} commits`,
+      value: round(position / commits.length, 2),
+      position,
       totalCommitCount: commits.length,
       currentCommitSummary: remainingCommits[0].summary ?? '',
     },
     remainingCommits,
     commits,
     targetBranchUndoSha: headSha,
-    countCherryPicked: commitsCherryPicked.length,
+    cherryPickedCount: commitsCherryPicked.length,
   }
 }
 
@@ -413,7 +413,7 @@ export async function continueCherryPick(
       options,
       snapshot.commits,
       progressCallback,
-      snapshot.countCherryPicked
+      snapshot.cherryPickedCount
     )
   }
 
--- a/app/src/lib/git/cherry-pick.ts
+++ b/app/src/lib/git/cherry-pick.ts
@@ -240,7 +240,7 @@ export async function getCherryPickSnapshot(
   // Each line of .git/sequencer/todo holds a sha of a commits to lined up to be
   // cherry-picked. These sha are in historical order starting oldest commit as
   // the first line and newest as the last line.
-  const remainingShas: string[] = []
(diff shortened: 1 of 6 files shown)


app/src/lib/git/cherry-pick.ts
```diff
@@ -262,49 +287,53 @@ export async function getCherryPickSnapshot(
...
       totalCommitCount: commits.length,
-      currentCommitSummary: commits[commitSummaryIndex].summary ?? '',
+      currentCommitSummary: remainingCommits[0].summary ?? '',
     },
-    remainingCommits: commits.slice(count, commits.length),
+    remainingCommits,
     commits,
-    targetBranchUndoSha: firstSha,
+    targetBranchUndoSha: headSha,
+    countCherryPicked: commitsCherryPicked.length,
```
sergiou87: Total nit, but the properties above are named `cherryPickCommitCount` and `totalCommitCount`, so maybe this one should be `cherryPickedCount`?

sergiou87 (changes requested):
LGTM and makes sense! Thanks for such a detailed explanation 🙇‍♂️ 

I only have one nit, but I think it makes sense looking at other identifiers in this PR 😳

app/src/lib/git/cherry-pick.ts
tidy-dev: Done.. and also updated the cherryPickCommitCount to position.. because it really isn't a count. It represents which commit in a selection of commits is being cherry-picked at them moment.

sergiou87 (approved):
LGTM, thanks for those changes! ❤️ :shipit:

PR #11972, opened 2021-04-09
Fix off-by-one error merging diff hunks

sergiou87:
## Description

Because of how Codemirror works, I have only seen the consequences of this bug in the unified diff 🤦‍♂️ 

The problem was we calculated the new end line of a new merged hunk with a +1. For example, if a hunk starts at 0 and has 47 lines… its end line should be 46, not 47.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #11975, opened 2021-04-09
Cherry-pick Snapshot Refactor didn't account for when no sequencer files

tidy-dev:
## Description

If user cherry-picks one conflicting commit, cherry-pick operation does not use sequencer files (not multiple commits). But, we still need to grab a snapshot for on continue to display info in progress dialog. That info can be grabbed from the cherry-picking head which contains the sha of the commit being cherry-picked.

## Release notes
Notes: no-notes


sergiou87 (approved):
LGTM! It's interesting that it uses a different method when it's only one commit 🤯

PR #11976, opened 2021-04-09
Cherry-picking refactor to use commit list instead of range for multiple commits

tidy-dev:
## Description

We found there were certain scenarios in which using the rev range argument `sha1^..sha4` did git math in a way that may not be expected by the user and resulted in cherry-picking more or less commits then shown in their selection.

This PR refactors cherry-picking to use each sha of the commits in the user's selection as arguments (`sha1 sha2 sha3 sha4`). This forces the cherry-pick to evaluate them individually and does not go and find more related commits/changes.

### Screenshots
Prior to this change -> the following cherry-pick would exclude the middle commit. As it was doing git math across sha1 and sha3 with respect to new target branch (which was 4 month old). sha1 and sha3 were not related to changes in sha2.. so git math said it didn't matter for the target branch/rev range. 😛 

https://user-images.githubusercontent.com/75402236/114194505-93573780-991d-11eb-8611-2a584faa2487.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/test/unit/git/cherry-pick-test.ts
+++ b/app/test/unit/git/cherry-pick-test.ts
@@ -4,6 +4,7 @@ import * as Path from 'path'
 import {
   getCommit,
   getCommits,
+  getCommitsInRange,
   merge,
   MergeResult,
   revRangeInclusive,
@@ -55,7 +56,8 @@ describe('git/cherry-pick', () => {
   })
 
   it('successfully cherry-picked one commit without conflicts', async () => {
-    result = await cherryPick(repository, featureBranch.tip.sha)
+    const featureTip = await getCommitOneLine(repository, featureBranch.tip.sha)
+    result = await cherryPick(repository, [featureTip])
     const cherryPickedCommit = await getCommit(
       repository,
       featureBranch.tip.sha
@@ -88,7 +90,8 @@ describe('git/cherry-pick', () => {
     )
     expect(emptyMessageCommit?.summary).toBe('')
 
-    result = await cherryPick(repository, featureBranch.tip.sha)
+    const featureTip = await getCommitOneLine(repository, featureBranch.tip.sha)
+    result = await cherryPick(repository, [featureTip])
 
     const commits = await getCommits(repository, targetBranch.ref, 5)
     expect(commits.length).toBe(2)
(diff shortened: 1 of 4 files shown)


app/test/unit/git/cherry-pick-test.ts
```diff
@@ -494,23 +530,23 @@ describe('git/cherry-pick', () => {
...
 
-    it('errors when given invalid revision range', async () => {
+    it('errors when given no commits', async () => {
       const progress = new Array<ICherryPickProgress>()
-      result = await cherryPick(repository, 'INVALID REF', p =>
-        progress.push(p)
-      )
+      result = await cherryPick(repository, [], p => progress.push(p))
       expect(result).toBe(CherryPickResult.UnableToStart)
     })
```
sergiou87: Isn't this the same as https://github.com/desktop/desktop/pull/11976/files#diff-ce9ae9235d8e18f9cc6b0b7868c17378800f12f5cacf604cff2eb0e4354e833fR194 ? I think this specific test case made sense with the previous implementation, because `UnableToStart` was only returned when `cherryPick` had a progress callback, but I think we can remove this now, no? 🤔

sergiou87 (changes requested):
All looks good! Only have one comment.

I think this change will make cherry-picking much easier to understand and aligned with what users expect 😄

app/test/unit/git/cherry-pick-test.ts
tidy-dev: Yeah. Makes sense

sergiou87 (approved):
:shipit:

billygriffin:
This makes sense to me too, thanks!

app/test/unit/git/cherry-pick-test.ts
```diff
@@ -599,6 +635,11 @@ describe('git/cherry-pick', () => {
   })
 })
 
+async function getCommitOneLine(repository: Repository, commitSha: string) {
+  const { sha, summary } = (await getCommit(repository, commitSha))!
+  return { sha, summary }
+}
+
 async function addThreeMoreCommitsOntoFeatureBranch(repository: Repository) {
```
natasha1007, line 643: hello mur

PR #11986, opened 2021-04-12
Replace non-fatal exception with error log

sergiou87:
## Description

We're getting a lot of non-fatal exceptions for something that we confirmed is not a problem at all anymore, so this PR replaces that non-fatal exception with a plain console log.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #11988, opened 2021-04-12
Disable diff expansion for big files

sergiou87:
## Description

Files longer than 256KB (-1B) will have diff expansion disabled for performance reasons. We would also not be able to highlight those files effectively either.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #12000, opened 2021-04-14
Allow creating an alias for repositories

### LINKED ISSUE #7856: Rename repository in the UI (or via a Workaround)
I have a couple of repositories checked out in multiple folders. When I add them to `GitHub Desktop` I see the same name of each of the repositories a couple of times. I can distinguish between them only by hovering with the cursor on their names.

I'd be very useful if `GitHub Desktop` allowed to rename what is shown in the UI. 

---

But maybe until this is (ever) implemented there is a workaround like modifying its settings in the `Roaming` folder?

sergiou87:
This is part of #7856

## Description

This PR adds initial support to create alias for repositories:
- New context menu actions to create, change or remove the alias of a repository from the repository list. Any character is valid.
- Filtering repositories now takes the alias into account.
- The original repository name is still visible by hovering over it, in the tooltip.

### Screenshots

https://user-images.githubusercontent.com/1083228/114741641-40c9b100-9d4b-11eb-8b83-a43270e293b3.mov

## Release notes

Notes: [New] Add an alias to your repos and name them however you want!

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1451,6 +1451,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
     previousRepositoryId: number | null,
     currentRepositoryId: number
   ) {
+    // No need to update the recent repositories if the selected repository is
+    // the same as the old one (this could happen when the alias of the selected
+    // repository is changed).
+    if (previousRepositoryId === currentRepositoryId) {
+      return
+    }
+
     const recentRepositories = getNumberArray(RecentRepositoriesKey).filter(
       el => el !== currentRepositoryId && el !== previousRepositoryId
     )
@@ -3411,7 +3418,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   /** This shouldn't be called directly. See `Dispatcher`. */
   public async _changeRepositoryAlias(
     repository: Repository,
-    newAlias: string
+    newAlias: string | null
   ): Promise<void> {
     return this.repositoriesStore.updateRepositoryAlias(repository, newAlias)
   }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -3409,6 +3409,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   /** This shouldn't be called directly. See `Dispatcher`. */
+  public async _changeRepositoryAlias(
+    repository: Repository,
+    newAlias: string
+  ): Promise<void> {
+    return this.repositoriesStore.updateRepositoryAlias(repository, newAlias)
+  }
+
+  /** This shouldn't be called directly. See `Dispatcher`. */
   public async _renameBranch(
     repository: Repository,
     branch: Branch,
--- a/app/src/lib/stores/repositories-store.ts
+++ b/app/src/lib/stores/repositories-store.ts
@@ -278,7 +278,7 @@ export class RepositoriesStore extends TypedBaseStore<
    */
   public async updateRepositoryAlias(
     repository: Repository,
-    alias: string
+    alias: string | null
   ): Promise<void> {
     await this.db.repositories.update(repository.id, { alias })
 
--- a/app/src/lib/stores/repositories-store.ts
+++ b/app/src/lib/stores/repositories-store.ts
@@ -132,6 +132,7 @@ export class RepositoriesStore extends TypedBaseStore<
         ? await this.findGitHubRepositoryByID(repo.gitHubRepositoryID)
         : await Promise.resolve(null), // Dexie gets confused if we return null
       repo.missing,
+      repo.alias,
       repo.workflowPreferences,
       repo.isTutorialRepository
     )
@@ -194,6 +195,7 @@ export class RepositoriesStore extends TypedBaseStore<
         return await this.db.repositories.put({
           ...(existingRepo?.id !== undefined && { id: existingRepo.id }),
           path,
+          alias: null,
           gitHubRepositoryID: ghRepo.dbID,
           missing: false,
           lastStashCheckDate: null,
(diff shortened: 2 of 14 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -1451,6 +1451,13 @@ export class AppStore extends TypedBaseStore<IAppState> {
     previousRepositoryId: number | null,
     currentRepositoryId: number
   ) {
+    // No need to update the recent repositories if the selected repository is
+    // the same as the old one (this could happen when the alias of the selected
+    // repository is changed).
+    if (previousRepositoryId === currentRepositoryId) {
+      return
+    }
```
sergiou87, line 1459: Without this, creating/changing/removing the alias of the current repo, would move the current repo into the Recents list, which doesn't make sense.

app/src/lib/stores/repositories-store.ts
```diff
@@ -132,6 +133,7 @@ export class RepositoriesStore extends TypedBaseStore<
         ? await this.findGitHubRepositoryByID(repo.gitHubRepositoryID)
         : await Promise.resolve(null), // Dexie gets confused if we return null
       repo.missing,
+      enableRepositoryAliases() ? repo.alias : null,
```
sergiou87, line 136: This prevents the alias from being loaded from persistence, so we can forget about checking the feature flag in many other places 😅

app/src/ui/app.tsx
```diff
@@ -2367,8 +2377,9 @@ export class App extends React.Component<IAppProps, IAppState> {
     let icon: OcticonSymbol
     let title: string
     if (repository) {
+      const alias = repository instanceof Repository ? repository.alias : null
       icon = iconForRepository(repository)
-      title = repository.name
+      title = alias ?? repository.name
```
sergiou87, line 2382: This is the title of the drop down that shows the current repository.

app/src/ui/repositories-list/group-repositories.ts
```diff
@@ -89,7 +89,7 @@ export function groupRepositories(
       const { aheadBehind, changedFilesCount } =
         localRepositoryStateLookup.get(r.id) || fallbackValue
       const repositoryText =
-        r instanceof Repository ? [r.name, nameOf(r)] : [r.name]
+        r instanceof Repository ? [r.alias ?? r.name, nameOf(r)] : [r.name]
```
sergiou87, line 92: This basically makes sure the alias is used to look for matches when the user filters the repository list.

app/src/ui/repositories-list/group-repositories.ts
```diff
@@ -147,11 +149,13 @@ export function makeRecentRepositoriesGroup(
 
     const { aheadBehind, changedFilesCount } =
       localRepositoryStateLookup.get(id) || fallbackValue
+    const repositoryAlias =
+      repository instanceof Repository ? repository.alias : null
     const repositoryText =
       repository instanceof Repository
-        ? [repository.name, nameOf(repository)]
+        ? [repositoryAlias ?? repository.name, nameOf(repository)]
```
sergiou87, line 156: This is also for the repository list filter.

app/src/ui/repositories-list/group-repositories.ts
```diff
@@ -132,8 +132,10 @@ export function makeRecentRepositoriesGroup(
   for (const id of recentRepositories) {
     const repository = repositories.find(r => r.id === id)
     if (repository !== undefined) {
-      const existingCount = names.get(repository.name) || 0
-      names.set(repository.name, existingCount + 1)
+      const alias = repository instanceof Repository ? repository.alias : null
+      const name = alias ?? repository.name
+      const existingCount = names.get(name) || 0
+      names.set(name, existingCount + 1)
```
sergiou87, line 138: This list of names is used to detect duplicates in the `Recents` group, to know when we need to disambiguate showing the owner as a prefix (e.g. `github/desktop`).
In this change I'm just adding the alias instead of the repo name, when available.

app/src/ui/repositories-list/group-repositories.ts
```diff
@@ -147,11 +149,13 @@ export function makeRecentRepositoriesGroup(
...
       localRepositoryStateLookup.get(id) || fallbackValue
+    const repositoryAlias =
+      repository instanceof Repository ? repository.alias : null
     const repositoryText =
       repository instanceof Repository
-        ? [repository.name, nameOf(repository)]
+        ? [repositoryAlias ?? repository.name, nameOf(repository)]
         : [repository.name]
-    const nameCount = names.get(repository.name) || 0
+    const nameCount = names.get(repositoryAlias ?? repository.name) || 0
```
sergiou87, line 158: This is where we actually check if the repo alias or name is duplicated and need disambiguation.

app/src/ui/repositories-list/repository-list-item.tsx
```diff
@@ -144,9 +156,37 @@ export class RepositoryListItem extends React.Component<
...
       },
     ]
+
     showContextualMenu(items)
   }
 
+  private buildAliasMenuItems(): ReadonlyArray<IMenuItem> {
+    const repository = this.props.repository
+
+    if (!(repository instanceof Repository) || !enableRepositoryAliases()) {
```
tidy-dev, line 166: Just curious.. is there a time that it is not an instance of a Repository?

tidy-dev: Hmmm.. just realized there is a type of Repositoryish. Which made me laugh... But I see that evidently it can be cloning one.

tidy-dev (approved):
✨  This is awesome! Though it is a pretty straight forward PR. I really appreciate all your PR comments along the way - makes reading through it a breeze. Tested it out and it is slick.

Only thing I came across which I don't think causes any harm - just a UI quirk. I can make the alias as long as I want, but I can only ever see so many characters and when I saw the ellipsis, I automatically tried to hover over it to see the rest of the alias. This likely not something people will do tho.. and if they do.. they can change it. 🤷 

image

sergiou87:
Thank you for your feedback! ❤️ 

The thing about the long aliases and not being able to see it in the tooltip is a good point, but also true that they can just change it 😆  I'll raise it in the next sync meeting.

wshaver:
Thank you thank you thank you thank you <3<3<3

PR #12001, opened 2021-04-14
Cherry-pick: create branch during drag n drop

tidy-dev:
## Description
Create a branch during cherry-picking via drag and drop!

### Screenshots

https://user-images.githubusercontent.com/75402236/114746671-039afa00-9d1e-11eb-826f-f2c4e73f7dd5.mov

Windows ghost difference: (where it says "a new branch", it normally displays the branch name)
image

## Release notes
Notes: [Added] Create a branch during cherry-picking via drag and drop!

### DIFF
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -2984,7 +2984,9 @@ export class Dispatcher {
     repository: Repository,
     targetBranchName: string
   ): Promise<void> {
-    const { branchesState } = this.repositoryStateManager.get(repository)
+    const { branchesState, cherryPickState } = this.repositoryStateManager.get(
+      repository
+    )
     const { defaultBranch, allBranches, tip } = branchesState
 
     if (tip.kind === TipState.Unknown) {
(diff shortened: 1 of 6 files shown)


ampinsk:
Can we add some margin above the new branch item? It's looking a little squished at the top there 😄

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -3014,7 +3016,26 @@ export class Dispatcher {
...
+      cherryPickState.step.kind !== CherryPickStepKind.CommitsChosen
+    ) {
+      // Started from context menu, cherry pick flow popup already open
+      return
+    }
+
+    const sourceBranch = tip.kind === TipState.Valid ? tip.branch : null
+
+    // If invoked from drag/drop, we need to show the cherry pick flow popup
+    this.showPopup({
```
sergiou87, line 3033: I just saw `showPopup` returns a Promise, so I guess we should `return` here. And there are a bunch of other calls to `showPopup` that are not awaited/returned 😱 

But then I follow `showPopup` implementation, get to https://github.com/desktop/desktop/blob/37d162de570e7d210b65c0ce078fe7939f51d412/app/src/lib/stores/app-store.ts#L2959 and see that the only reason to make `showPopup` async is not being awaited 😂 

And if I keep digging into it, I get to https://github.com/desktop/desktop/blob/37d162de570e7d210b65c0ce078fe7939f51d412/app/src/lib/stores/app-store.ts#L3012-L3023 where there is no reason for `_closeFoldout` to be async nor return a Promise 😅  So there is no need for `showPopup` to be async at all.

sergiou87: Of course, no need to say, this ☝️ was just me rambling on stuff that requires no action in this PR, sorry 😅

sergiou87 (approved):
Code looks great, and I tested and worked beautifully 😍 :shipit:

PR #12018, opened 2021-04-18
add stylelint

danny007in:
Removed many duplicate css properties 

like 

firefox_FLR8HV3aA4
firefox_ncCvfcqPnr

and classes

like

firefox_haVly1v4I6

Alos fixed with standard


tidy-dev:
Thank you @danny007in for wanting to improve the code-base. Even tho improving the style linter is great, we are quite cautious on it as this could have many unintended consequences, adds a lot of dependencies, and would require a lot of effort to correct the build processes.

PR #12029, opened 2021-04-20
Remove unused feature flags

sergiou87:
## Description

This PR removes some feature flags of features that were enabled long ago and are now part of the app 😄 

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/changes/changed-file-details.tsx
+++ b/app/src/ui/changes/changed-file-details.tsx
@@ -4,7 +4,6 @@ import { AppFileStatus } from '../../models/status'
 import { IDiff, DiffType } from '../../models/diff'
 import { Octicon, OcticonSymbol, iconForStatus } from '../octicons'
 import { mapStatus } from '../../lib/status'
-import { enableSideBySideDiffs } from '../../lib/feature-flag'
 import { DiffOptions } from '../diff/diff-options'
 import { RepositorySectionTab } from '../../lib/app-state'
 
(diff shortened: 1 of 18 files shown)


app/src/ui/history/commit-summary.tsx
```diff
@@ -357,42 +345,23 @@ export class CommitSummary extends React.Component<
...
-                  label="Hide Whitespace"
-                  value={
-                    this.props.hideWhitespaceInDiff
-                      ? CheckboxValue.On
-                      : CheckboxValue.Off
-                  }
-                  onChange={this.onHideWhitespaceInDiffChanged}
-                />
-              </li>
-            )}
```
sergiou87, line 375: This was here because before side-by-side diff, the "Hide Whitespace" option was a checkbox, that then was replaced by the ⚙️  Diff Options button/popup.

app/src/ui/history/commit-summary.tsx
```diff
@@ -157,13 +152,6 @@ export class CommitSummary extends React.Component<
     }
   }
 
-  private onHideWhitespaceInDiffChanged = (
-    event: React.FormEvent<HTMLInputElement>
-  ) => {
-    const value = event.currentTarget.checked
-    return this.props.onHideWhitespaceInDiffChanged(value)
-  }
```
sergiou87, line 165: This was used in `Hide Whitespace Changes` when it was a checkbox inline in the diff header, but not needed when it was moved into the `Diff Options` popover.

billygriffin (approved):
🪓

app/src/ui/changes/changed-file-details.tsx
```diff
@@ -43,18 +42,16 @@ export class ChangedFileDetails extends React.Component<
         <PathLabel path={this.props.path} status={this.props.status} />
         {this.renderDecorator()}
 
-        {enableSideBySideDiffs() && (
```
tidy-dev, line 46: TIL you can do this in react logical render in react.. 😁  It makes total sense; I had just always seen the pattern of call out to a method that would render based on some if.

tidy-dev (approved):
❌  😄

app/src/ui/changes/changed-file-details.tsx
sergiou87: I was more surprised about the `enableSideBySideDiffs() || …` to prevent a component from being rendered 😂

PR #12032, opened 2021-04-20
Cherry-pick multiple non sequential commits

### LINKED ISSUE #12030: Cherry Pick multiple commits via Ctrl button

The new cherry-pick feature is fantastic!  The one issue I have is that it currently only allows to select multiple commits with the Shift button, so it is impossible to select commits which are not listed side by side, 

### Proposed solution

Add support for multiple choice via Ctrl key, which will allow users to select several commits that aren't side by side.

tidy-dev:
Closes #12030

## Description

Allow cherry-picking multiple non sequential commits. 

A workflow decision made in this PR is that if a user selects commits in the order of 1, 5, 3. This will cherry pick in the order of 5,  3, 1 (5 being the oldest in history). This is generally cherry-picking best practice as it prevents accidentally cherry-picking a commit without changes/files needed in a previous commit.  If a user is attempting to use cherry-picking to reorder commits, they will have to explicitly do so with multiple cherry-pick operations.

### Screenshots

Below demos non sequential commits as well as shows some regression testing (range selection, up/down/paging, and single commit). Ensuring chronological ordering despite non chronological selection was also tested.

https://user-images.githubusercontent.com/75402236/115442450-05fac980-a1e0-11eb-8ec8-c383f35b1094.mov

## Release notes

Notes: [Improved] Cherry-pick multiple non-sequential commits at one time.

### DIFF
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -194,7 +194,7 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
     // cherry-picking results. If user wants to use cherry-picking for
     // reordering, they will need to do multiple cherry-picks.
     // Goal: first commit in history -> first on array
-    const sorted = rows.map(r => r).sort((a, b) => b - a)
+    const sorted = [...rows].sort((a, b) => b - a)
 
     const selectedShas = sorted.map(r => this.props.commitSHAs[r])
     const selectedCommits = this.lookupCommits(selectedShas)
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -3,7 +3,7 @@ import memoize from 'memoize-one'
 import { GitHubRepository } from '../../models/github-repository'
 import { Commit, CommitOneLine } from '../../models/commit'
 import { CommitListItem } from './commit-list-item'
-import { List, SelectionSource } from '../lib/list'
+import { List } from '../lib/list'
 import { arrayEquals } from '../../lib/equality'
 import { Popover, PopoverCaretPosition } from '../lib/popover'
 import { Button } from '../lib/button'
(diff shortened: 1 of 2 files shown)


app/src/ui/history/commit-list.tsx
```diff
@@ -188,22 +188,16 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
-      .slice(rangeStart, rangeEnd)
-      .reverse()
-    const selectedCommits = this.lookupCommits(commitSHARange)
+  private onSelectionChanged = (rows: ReadonlyArray<number>) => {
+    // Multi select can give something like 1, 5, 3 depending on order that user
+    // selects. We want to ensure they are in chronological order for best
+    // cherry-picking results. If user wants to use cherry-picking for
+    // reordering, they will need to do multiple cherry-picks.
+    // Goal: first commit in history -> first on array
+    const sorted = rows.map(r => r).sort((a, b) => (a - b) * -1)
```
sergiou87: Why not just this? 🤔 
```suggestion
    const sorted = rows.map(r => r).sort((a, b) => b - a)
```

sergiou87: Also, why the `rows.map(r => r)` part? 🤔 is that to create a copy?

sergiou87 (commented):
Only have a couple of style-related nits, other than that LGTM 😄

app/src/ui/history/commit-list.tsx
tidy-dev: `b - a` because that is not how my brain thinks inversely 😁 

`rows.map(r => r)` does copy it but for my purposes converts it into a normal not readonly array that `sort` can mutate. I could have modified the `List` component to just pass a non-readonly array, but didn't think my specific problem warranted that change.

Muhammadariesta (approved):

app/src/ui/history/commit-list.tsx
sergiou87: For the copy, I was thinking about something more explicit like a `rows.slice()` like I've seen in other places of the codebase. I wish there was an ever more explicit `rows.mutableCopy()`-like method, but well 😂

sergiou87 (approved):
Thanks for the changes! :shipit:

app/src/ui/history/commit-list.tsx
tidy-dev: For context, discussed the copy and landed on using the spread operator. I am of the camp that `slice` is not more explicit than `map`. After research of some options, we landing on using the spread operator `[...rows]`.  

Also notes for future reference, efficiency was taking into consideration. According to [this stack overflow](https://stackoverflow.com/questions/3978492/fastest-way-to-duplicate-an-array-in-javascript-slice-vs-for-loop)  the spread operator is the fastest. Also to note `slice` is 2nd fastest far outpacing `map`. There is also `Array.From` which is similar performance to `slice` and `...`. Seems it could also be considered a good choice.

Muhammadariesta (approved):

PR #12045, opened 2021-04-23
Add the architecture as path component of the updates URL

sergiou87:
## Description

As part of the work to support arm64 I learned a bit about how Squirrel.Windows works and what it expects. So far we tried:
1. Using query parameters to indicate the architecture of the updates we want: this query parameter is only sent when Squirrel.Windows hits the `/RELEASES` endpoint, but not when it's gonna download the nupkg file with the update.
2. Adding an architecture suffix to the nupkg files listed in the `/RELEASES` endpoint, so we know which build to provide: this doesn't work because for some reason Squirrel.Windows is getting confused with versions and it just doesn't update.

In this PR we try a third approach which consists on changing the base path of all endpoints used by Squirrel (for Mac and Windows), including the architecture as a path component. Thanks to that we will know the architecture in both the `/RELEASES` and nupkg endpoints, without the need of adding any kind of suffix to the nupkg files.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/update-store.ts
+++ b/app/src/ui/lib/update-store.ts
@@ -171,7 +171,10 @@ class UpdateStore {
       remote.app.runningUnderRosettaTranslation === true
     ) {
       const url = new URL(updatesURL)
-      url.searchParams.set('architecture', 'arm64')
+      url.pathname = url.pathname.replace(
+        /\/desktop\/desktop\/(x64\/)?latest/,
+        '/desktop/desktop/arm64/latest'
+      )
       updatesURL = url.toString()
     }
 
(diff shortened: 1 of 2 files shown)


app/src/ui/lib/update-store.ts
```diff
@@ -171,7 +171,10 @@ class UpdateStore {
       remote.app.runningUnderRosettaTranslation === true
     ) {
       const url = new URL(updatesURL)
-      url.searchParams.set('architecture', 'arm64')
+      url.pathname = url.pathname.replace(
+        /\/desktop\/desktop\/(x64\/)?latest/,
+        '/desktop/desktop/arm64/latest'
+      )
```
sergiou87, line 177: Maybe I was overly cautious here and just looking for `desktop\/(x64\/)?latest` is enough, but 🤷‍♂️

tidy-dev (approved):
✨

PR #12067, opened 2021-04-26
Enable Hide White Space In Diff Option

tidy-dev:
## Description

Flip enable hide white space in diff option for production.

## Release notes
Notes: no-notes


sergiou87 (approved):
:shipit:

PR #12071, opened 2021-04-27
Change the diff gutter width based on number of lines in diff

### LINKED ISSUE #2102: Adjust width of line number column based on number of lines in file
Why are the line number columns so wide?  Minor I know, but annoying to look it.  Can we make it adjust to the size needed or let the user define it?

ADustyOldMuffin:
Closes #2102

## Description

- Adjusts the number column width in the diff view to size based on the number of lines.

### Single line

image

### Double digit lines

image

### Triple digit lines

image

This continues for any number of lines, this has the potential to get excessive since there is no limit but in my opinion I think you'd have other issues if someone is loading a document with more than a million line difs.

## Release notes

Notes: Changed line number column in diff view to scale in size based on the number of lines in the diffs.


ADustyOldMuffin:
I'm looking for feedback, I know the solution isn't super elegant but I'd like to know if this is on the right track/any suggestions.

app/src/ui/diff/text-diff.tsx
```diff
@@ -1018,6 +1021,29 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
       : canSelect(file) && file.selection.isSelected(index)
   }
```
salihalgul: ok

salihalgul (approved):

salihalgul (approved):

salihalgul:
ok

tidy-dev:
@ADustyOldMuffin Thanks for picking this up! It does seem that the solution would need to be along these lines of calculating a width to apply for each line.  

Couple of things to consider: 
- The unified diff and the split diff use a different renderer. Thus, the complete solution will need something for both, where what you have currently is only for unified. (not assuming that wasn't your intention just voicing it for clarity)
- We just shipped expanding diffs, thus, when a file is really long there will be expansion handles at the top bottom and sometimes in the middle of the diffs number columns. Just need to ensure they are not negatively impacted on both unified/split view. 
Screen Shot 2021-04-28 at 7 37 33 AM

live125 (approved):

ADustyOldMuffin:
@tidy-dev Thanks for pointing out the split view (I had no idea), and after going back and checking I was using the incorrect value when trying to calculate how wide the div should be.

I changed it to use the line number since that's what we're trying to display differently anyway (whoops), and I also added the variable sizing to the split view. The expanding hunk works fine, and actually will make the line number div larger if the numbers add a digit.

Here are some screen shots, they're only the split view but if I need to redo the original screenshots I can

**Single**
SplitSingle

**Double**
SplitDouble

**Triple**
SplitTriple

**Quad**
SplitQuad

**Before Expand**
SplitBeforeExpand

**After Expand**
SplitAfterExpand

**Lots of changes!**
SplitSideBySideMultiHunk

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -14,6 +14,7 @@ import { narrowNoNewlineSymbol } from './text-diff'
 import { shallowEquals, structuralEquals } from '../../lib/equality'
 import { DiffHunkExpansionType } from '../../models/diff'
 import { DiffExpansionKind } from './text-diff-expansion'
+import { String } from 'aws-sdk/clients/xray'
```
tidy-dev: Looks like you are just passing in a regular string - no need to include `String` from this package.

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -40,6 +41,11 @@ interface ISideBySideDiffRowProps {
   /** Whether or not whitespace changes are hidden. */
   readonly hideWhitespaceInDiff: boolean
 
+  /**
+   * Whether to display the rows side by side.
```
tidy-dev: Update comment please.

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -384,18 +409,20 @@ export class SideBySideDiffRow extends React.Component<
...
    *
+   * @param divWidth The width the line number component should be.
    * @param lineNumbers Array with line numbers to display.
    * @param isSelected  Whether the line has been selected.
    *                    If undefined is passed, the line is treated
    *                    as non-selectable.
    */
   private renderLineNumbers(
     lineNumbers: Array<number | undefined>,
+    divWidth: String,
```
tidy-dev: I am not sure this actually needs passed in from the render function since it is on the `props`. I think you should be able to access it direction in this method with `this.props.lineNumberWidth`.

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -427,12 +455,17 @@ export class SideBySideDiffRow extends React.Component<
...
    *
    * @param lineNumber  Line number to display.
+   * @param divWidth The width the line number component should be.
    * @param isSelected  Whether the line has been selected.
    *                    If undefined is passed, the line is treated
    *                    as non-selectable.
    */
-  private renderLineNumber(lineNumber?: number, isSelected?: boolean) {
-    return this.renderLineNumbers([lineNumber], isSelected)
+  private renderLineNumber(
```
tidy-dev: See comment on `renderLineNumbers`

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -337,6 +340,54 @@ export class SideBySideDiff extends React.Component<
     )
   }
 
+  private getRowWidth = (rows: readonly SimplifiedDiffRow[]): string => {
+    const largestBeforeLineNum = Math.max(
```
tidy-dev: I am not super familiar with the diff logic.. but I would suspect that the rows are ordered. If this is true, we could more efficiently just grab the last value from the rows array, correct?

@sergiou87 Maybe you are more familiar with this to weigh in.

app/src/ui/diff/text-diff.tsx
```diff
@@ -1019,6 +1022,48 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
+      ...totalHunks.map(function (o) {
+        return Math.max(
+          ...o.lines.map(function (val) {
+            return val.newLineNumber ?? 0
+          })
+        )
+      })
+    )
+
+    const maxLinesDigitAmount =
```
tidy-dev: Seems like duplicate logic for calculating width at the end of this method and the end of `getRowWidth` for side by side implementation, suggest it could be moved to a util/lib function.

tidy-dev (changes requested):
@ADustyOldMuffin Thanks for all your work on this so far. This is looking like it is on the right track. However, I have a few suggestions I would like for you to take a look at.

billygriffin (dismissed):
Thanks for this PR! In addition to the things @tidy-dev mentioned in her review, I agree with the initial proposal in the issue to define a minimum width, and I also think a minimum that's equivalent to 3 numbers feels appropriate. To make that more concrete, if you're on line 4 (single digit) or line 22 (two digits), we would still make the gutter wide enough to fit three digits. Excited to see this come together!

app/src/ui/diff/text-diff.tsx
ADustyOldMuffin: I moved this to diff-helper.tsx, but I'm not sure if this is the correct location for it. Let me know if I should move it elsewhere.

app/src/ui/diff/side-by-side-diff.tsx
ADustyOldMuffin: After poking through several repos and files it seems like the 2nd max call can be simplified as you said as the final value is always the last line in the diff. This makes me feel better because I didn't like the nested max statements, thank you!

ADustyOldMuffin:
Thank you @billygriffin I've made the changes requested by @tidy-dev and set the default to 3 (10 * digit number + 5) which gives 10 pixels to each digit and 5 pixels of padding which I think looks pretty good.

This can also work with any number of digits, but I didn't know if you wanted to possibly put an upper cap for edge cases? I would think there would be other problems if you have files larger than a couple million lines.

billygriffin:
Thanks @ADustyOldMuffin! I'm seeing a strange gutter area to the right of the line numbers, but only in unified diffs. In split diffs, things seem to work as I expected. To your other question about capping this at some upper number, I'm indifferent about that and agree that there are likely other problems if we run into that constraint.

Screen Shot 2021-05-12 at 12 47 14 PM

ADustyOldMuffin:
@billygriffin So after looking at it, this seems to be the Code Mirror as the `diff-gutter` class is set to a size of `125px` in `_diff.scss` (I'd link the line if I knew how). I added a dynamic update to this style at the end of the diff creation, but it needs a call to `cm.refresh()`  to move the editor into place. I mention this because right above where I did this there's some code doing kind of the same thing, but being picky about if there was a change or not. I'm not sure if I should also add more code to be picky, or just refresh it every time as I'm not sure of impact.

billygriffin:
@ADustyOldMuffin Thanks for the context! @tidy-dev or @sergiou87 would one of y'all mind weighing in on that.

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: Sorry I completely missed this conversation 😱 

As @tidy-dev says, you shouldn't need to go through all rows. Just going to the last line of the last hunk (unless it's a "dummy" hunk just used for expansion in the UI) and going backwards looking for the first (last) instance of a new/old line, should be enough and much more efficient.

app/src/ui/diff/text-diff.tsx
```diff
@@ -1004,6 +1007,11 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
     if (batchedOps.length > 0) {
       cm.operation(() => batchedOps.forEach(x => x()))
     }
+
+    // We have to set the size of the gutter, and because it might have changed we should refresh to fit the editor.
+    const gutterElement = document.getElementsByClassName('diff-gutter')[0]
+    gutterElement.setAttribute('style', `width: ${diffSize * 2}px;`)
```
sergiou87, line 1021: Instead of changing the style directly, hardcoding a width, I think it'd be cleaner (and more consistent with the rest of the codebase) to have a set of "diff gutter width" classes depending on the size we need. For example:
```css
--diff-gutter-width-per-char: 15px;

// Width for 1-digit line numbers
.diff-gutter-width-1:
  width: calc(var(--diff-gutter-width-per-char) + 5);

// Width for 2-digit line numbers
.diff-gutter-width-2:
  width: calc(var(--diff-gutter-width-per-char) * 2);

// Width for 3-digit line numbers
.diff-gutter-width-2:
  width: calc(var(--diff-gutter-width-per-char) * 3);

...
```

Then, you apply a class to the diff gutter depending on the number of digits. Does that make sense?

app/src/ui/diff/text-diff.tsx
```diff
@@ -1004,6 +1007,11 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
+    cm.refresh()
```
sergiou87, line 1022: Maybe there is some other place where you can apply the style changes and you don't need to call `refresh()`? Maybe in the `render` method? 🤔

app/src/ui/diff/text-diff.tsx
```diff
@@ -1019,6 +1027,28 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
+      ...totalHunks.map(function (o) {
+        return o.lines[o.lines.length - 1].oldLineNumber ?? 0
+      })
+    )
+
+    const largestNewLineNum = Math.max(
+      ...totalHunks.map(function (o) {
+        return o.lines[o.lines.length - 1].newLineNumber ?? 0
+      })
+    )
```
sergiou87: As mentioned above, this way of obtaining the largest line numbers seems it could be inefficient, and just look backwards from the end of the diff should work.

Another option that might be more efficient is to carry that information in the `ITextDiffData` interface, and obtain it when the diff is parsed in https://github.com/desktop/desktop/blob/21e071f5eff5e533afe4d4ef39ef7d1e4d7e9f9b/app/src/lib/diff-parser.ts#L386 and then keep it updated when the user expands the diff in https://github.com/desktop/desktop/blob/21e071f5eff5e533afe4d4ef39ef7d1e4d7e9f9b/app/src/ui/diff/text-diff-expansion.ts#L173

That way, this info has always been pre-calculated and available.

What do you think?

sergiou87 (changes requested):

app/src/ui/diff/text-diff.tsx
sergiou87: Also /cc @tidy-dev to get her input, I understand the convenience of just hardcoding the exact width depending on the number of digits, but it feels "wrong" 😳

tidy-dev: Agree using css is generally optimal when applying styles as you are letting the css control the view. Doing so would necessitate having an upper limit, because you are limited to the number of classes you make. IMHO.. the diff viewer is a unique case because it inherently must calculates things dynamically for the view. Thus, it doesn't feel that wrong to me. I think due to the nature of how the lines are rendered we are stuck with a non-optimal solution of calculating it where in an ideal world these things would be in columns that we could just add padding and it would adjust itself to the widest contents. 

Long story short: My 2 cents is I feel either solution would be acceptable.

ADustyOldMuffin: I don't know if my input makes a difference, but I will say that hard coding css values would work since the upper limit is ignored. The difference is the dynamic version will allow the large numbers to work with the amount of space for text possibly looking wonky, but the other will just truncate the numbers since it'll only get so large.

I'd personally prefer dynamic.

app/src/ui/diff/text-diff.tsx
ADustyOldMuffin: That sounds a lot better since I believe this is ran several times when the view is refreshed, but I wasn't sure of a better place.

A long with the render method suggestion I'll try and scoot stuff around and get it working.

app/src/ui/diff/text-diff.tsx
sergiou87: 👍  let's go ahead with the dynamic approach then!

app/src/ui/diff/text-diff.tsx
ADustyOldMuffin: This works excellently and is much better than calculating it on every single render of the page. Thank you!

app/src/ui/diff/text-diff.tsx
ADustyOldMuffin: With the `render` method the issue is the element doesn't exist and it's a generated div so I can't grab it before hand. I tried `componentDidMount` (granted we'd still have to call refresh), but the issue of navigating to a new diff with say 4 digits when the initial had 3 would make the gutter cover text. 

I've gone ahead and left the refresh where it's at as I'm not sure I can get around it since `CodeMirror` has to be the one to correct the text editor size. with that though if it's okay I would remove the call right above these lines since we're refreshing regardless.

app/src/models/diff/diff-data.ts
```diff
@@ -53,6 +53,8 @@ interface ITextDiffData {
   readonly hunks: ReadonlyArray<DiffHunk>
   /** A warning from Git that the line endings have changed in this file and will affect the commit */
   readonly lineEndingsChange?: LineEndingsChange
+  /** The number of digits in the largest line number of the diff  */
+  readonly maxLineNumberDigitCount: number
```
sergiou87: I personally feel like this field name is too UI-related for a model, so I think I'd just store here the max line number that we found during parsing. Then, do the final step of `.toString().length` right where it's needed at the UI level.

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -352,3 +352,32 @@ export function canSelect(
...
+  const finalHunkIndex =
+    fromExpansion && hunks.length > 2 ? hunks.length - 2 : hunks.length - 1
+  const lastLine =
+    hunks[finalHunkIndex].lines[hunks[finalHunkIndex].lines.length - 1]
+  const largestLine =
+    ((lastLine.newLineNumber ?? 0) > (lastLine.oldLineNumber ?? 0)
+      ? lastLine.newLineNumber
+      : lastLine.oldLineNumber) ?? 0
+  return largestLine.toString().length
+}
```
sergiou87: I think the logic when `fromExpansion === true` is wrong 🤔 

If I understand correctly the logic around `hunks.length > 2 ? hunks.length - 2 : hunks.length - 1` is to avoid the dummy hunk used for expansion at the end of the diff, however that might not exist if the last hunk of the (real) diff is already located at the end.

A more generic approach to this would be a loop from the last line of the last hunk, going backwards, until you find a line that is not of `Hunk` type (and therefore has either `newLineNumber`, or `oldLineNumber`, or both). Sounds a bit excessive, but I think it's cleaner than having to take into account, explicitly, if the last hunk is one of these dummy hunks and/or if there are more than 2 or 1 hunks. Also, I think the intent of the resulting code would be clearer too 😄

sergiou87: Also, given how cheap this implementation seems to be… maybe there is no need to do it on both expansion and parsing and we can just do it when needed as you had it before?

Sorry about that change of mind 😅  To be honest, when I suggested to do this on expansion and parsing, I had something different in mind: we would change the max line number as we expand a hunk or parse the diff line by line, instead of looking at the last line of the resulting diff after parsing/expanding.

If this is fast enough, I'd rather remove it from the expansion/parsing code 😳

app/src/ui/diff/text-diff.tsx
sergiou87: Ohhh ok, that makes sense!

Re obtaining the `diff-gutter` element: can you use `cm.getGutterElement()` instead? That gives you the parent of the element your code is getting now. If setting the `width` to that doesn't work, you can at least use that element instead of `document` to call `getElementsByClassName('diff-gutter')` which I'd hope would be more efficient as it should narrow the search.

I wonder if there is a way to encapsulate this logic inside `CodeMirrorHost`, so that the `refresh` is not needed (for example, adding another prop to `CodeMirrorHost` with the gutter width and apply it to the gutter element in the `updateDoc` method?).

sergiou87 (changes requested):
This is looking much better, thank you for all your work!! ❤️ 

I found an issue getting the max number of lines, and also left a couple of suggestions/thoughts 😄

app/src/ui/diff/diff-helpers.tsx
ADustyOldMuffin: If you'll take a look I changed the logic for grabbing the last line number and parsing hunks. It seems to work alright and feels a lot cleaner than what I was doing before since it had a bit of code smell to me when I was skipping the last hunk or two to dodge the dummy.

As for cheapness I have to say that putting the `largestLineNumber` in the diff data and calculating it when parsing them feels a lot cleaner as the data feels like it belongs there. It also only runs once when the diffs load (sometimes for 2 different diffs when first loading, but I'm assuming that's when it's parsing every hunk in that diff) and then only when expanded. 

I personally think the approach you originally suggested seems a lot cleaner and efficient than what I was doing before as the logic doesn't seem to belong in the individual diffs and also replicates code a bit.

That's just my 2 cents though, so after looking at it, if you still want to swap it back I can.

app/src/ui/diff/text-diff.tsx
ADustyOldMuffin: So I swapped it to use the getGutterElement and that works fine. I've tried the method you mentioned for doing it inside of the updateDoc method, but even if a different size is passed to updateDoc and it's set on the correct element the size of the editor never changes. I might be doing it incorrectly or not understanding, but I didn't see anything inside the updateDoc method that would refresh code mirror to resize itself.

ADustyOldMuffin:
@sergiou87 or @billygriffin Is there anything else that needs to be done with this? Just wanted to make sure it wasn't waiting on me.

billygriffin:
@ADustyOldMuffin Sorry for the delay! We're working toward our next release later this week, which is pretty big and required a lot of work to get it together over the past several weeks. @sergiou87 should hopefully be able to take another look at this by next week. ❤️

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -352,3 +353,52 @@ export function canSelect(
...
+  }
+
+  return diffSize
+}
+
+/** Utility function for getting the digit count of the largest line number in an array of diff hunks */
+export function getLargestLineNumber(hunks: DiffHunk[]): number {
+  let isValidLine = false
+  let currentHunkIndex = hunks.length - 1
+  let currentLineIndex = hunks[currentHunkIndex].lines.length - 1
```
sergiou87: Shouldn't we have checks for the length of those arrays?
```suggestion
  if (hunks.length === 0) {
    return 0
  }
  
  let currentHunkIndex = hunks.length - 1
  
  if (hunks[currentHunkIndex].lines.length === 0) {
    return 0
  }
  
  let currentLineIndex = hunks[currentHunkIndex].lines.length - 1
```

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -367,17 +368,37 @@ export function getLineWidthFromDigitCount(digitAmount: number): number {
...
+      // We've managed to go through all the hunks and there are no valid lines
+      // I don't think this could ever happen, but just in case let's return
+      if (currentHunkIndex === -1) {
+        return 0
+      }
+
+      currentLineIndex = hunks[currentHunkIndex].lines.length - 1
+    }
+  }
+  return largestLineNumber
```
sergiou87: Maybe I'm missing something, but isn't this logic too complicated? 🤔 

I had in mind something like this:
```typescript
for (let i = hunks.length -1; i >= 0; i--) {
  const hunk = hunks[i]

  for (let j = hunk.lines.length - 1; j >= 0; j--) {
    const line = hunk.lines[j]

    if (line.type === DiffLineType.Hunk) {
      continue
    }

    const newLineNumber = line.newLineNumber ?? 0
    const oldLineNumber = line.oldLineNumber ?? 0
    return newLineNumber > oldLineNumber ? newLineNumber : oldLineNumber
  }
}

return 0
```

This might require some linter love, sorry 😅

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -302,7 +302,7 @@ export class SideBySideDiff extends React.Component<
     }
 
     const lineNumberWidth = `${getLineWidthFromDigitCount(
-      this.state.diff.maxLineNumberDigitCount
+      this.state.diff.maxLineNumber.toString().length
```
sergiou87: From what I read, there are faster ways of getting how many digits a number has. Haven't looked into it with much detail (lack of time today) but in case you want to take a look: https://stackoverflow.com/a/14879700

This applies to other instances where you do the same `.toString().length` dance 😄

sergiou87 (changes requested):
Sorry again for the delay @ADustyOldMuffin!! As Billy said, this release had plenty of things and kept us busy 😅 

I love where this is going, hopefully we can get it merged soon 🤞  Thanks for all your hard work on it!! ❤️

ADustyOldMuffin:
I completely understand! I just wanted to make sure it wasn't waiting on me.

app/src/ui/diff/diff-helpers.tsx
ADustyOldMuffin: I removed most of this since the below logic does away with a lot of this anyway.

app/src/ui/diff/diff-helpers.tsx
ADustyOldMuffin: I like this idea much more since I wasn't a fan of the while loop I used. Thanks!

app/src/ui/diff/side-by-side-diff.tsx
ADustyOldMuffin: That seems to be the fastest for positive integers, which we should only ever have.

sergiou87 (dismissed):
Thanks for all your changes @ADustyOldMuffin !!! It looks good to me now 😄 

I'd love to get a final 👍  from @tidy-dev before merging it, but I think it looks great. I tested it and worked as expected in both split and unified views 😄

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -352,3 +353,45 @@ export function canSelect(
...
+export function getLineWidthFromDigitCount(digitAmount: number): number {
+  let diffSize: number
+
+  if (digitAmount <= 3) {
+    diffSize = 35
+  } else {
+    diffSize = 10 * digitAmount + 5
+  }
+
+  return diffSize
```
tidy-dev: Minor Suggestion - We try to make use of constants where possible and follow a pattern of tertiary operators to do so. Also at top of method you could have a `const minLineWidth = 35` to make it just a bit easier to read at a glance. Thus this function could be something like `return digitAmount <= 3 ? minLineWidth : 10 * digitAmount + 5`.

app/src/ui/diff/text-diff.tsx
```diff
@@ -1004,6 +1009,17 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
     if (batchedOps.length > 0) {
       cm.operation(() => batchedOps.forEach(x => x()))
     }
+
+    const diffSize = getLineWidthFromDigitCount(
+      this.state.diff.maxLineNumber.toString().length
```
tidy-dev: Why not use your `getNumberOfDigits` here just for consistency sake?

tidy-dev (changes requested):
Just a couple of minor suggestions.

app/src/ui/diff/diff-helpers.tsx
sergiou87: or…
```typescript
return Math.max(digitAmount, 3) * 10 + 5
```

app/src/ui/diff/text-diff.tsx
sergiou87: 🤦  I missed this one. I did good asking for a second pair of 👀  Thank you!!

ADustyOldMuffin:
Thanks @tidy-dev and @sergiou87, sorry for the delay but the requested changes should be in now. Let me know if you see anything else!

tidy-dev (approved):
✨   Thanks for those changes!

sergiou87 (approved):
Looks good!!! 😍

sergiou87:
Thank you again for your contribution @ADustyOldMuffin !!!! 🙇‍♂️ 🙇‍♂️ 🙇‍♂️ 🙇‍♂️

PR #11969, opened 2021-04-08
Temporary fix for keytar on Apple Silicon

dennisameling:
Related to https://github.com/desktop/desktop/pull/9691#issuecomment-815999757

There are issues with Electron arm64 builds on Apple Silicon when using keytar. Related issue: https://github.com/atom/node-keytar/issues/346.

Keytar will (likely quite soon) switch to N-API prebuilds which eliminates the needs for them to publish prebuilds for every Node + Electron version (currently at 92 (!) prebuilds https://github.com/atom/node-keytar/releases/tag/v7.6.0), see https://github.com/atom/node-keytar/pull/331. I've created a temporary Keytar release based on that PR: https://www.npmjs.com/package/@dennisameling/keytar-temp

The alternatives to merging this PR are:
- Waiting for the upstream PR to be merged (switch to N-API) https://github.com/atom/node-keytar/pull/331
- Upgrading to Electron 12 (which for some weird reason doesn't have this issue, see https://github.com/atom/node-keytar/pull/376)

A release for testing on Apple Silicon can be found here: https://github.com/dennisameling/desktop/releases/tag/2.7.3-test

PR #12083, opened 2021-04-27
Bump dependencies in order to support arm64 builds for macOS

sergiou87:
## Description

This PR bumps keytar, fs-admin, registry-js and desktop-trampoline to builds with N-API support, so that we can finally build a macOS beta with [_some_](https://github.com/desktop/desktop/pull/11969#issuecomment-827317892) arm64 support 😄 

## Release notes

Notes: no-notes


.nvmrc
```diff
@@ -1 +1 @@
-v14
+v14.15.4
```
sergiou87, line 1: This is the version we use in our CI scripts. Using something lower than 14.1 will fail when building the app due to https://github.com/maxogden/extract-zip/issues/94

tidy-dev (approved):
✨

PR #12087, opened 2021-04-28
Initial squash git logic

tidy-dev:
## Description

First bit of logic for squashing:
- adds interactive rebasing building upon existing rebase methods/data structures
- adds method for squashing utilizing introduced interactive rebase
- one unit test - POC
   - next PR to expand unit tests to include multiple, non-sequential/reorder, and conflict unit tests
- Also modified log to allow for no limit on getCommits as we don't know how far back in history someone might try to squash

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/squash.ts
+++ b/app/src/lib/git/squash.ts
@@ -9,11 +9,10 @@ import { rebaseInteractive, RebaseResult } from './rebase'
  * Squashes provided commits by calling interactive rebase
  *
  * @param toSquash - commits to squash onto another commit
- * @param squashOnto  - commit to squash the `squash` commits onto
- * @param previousCommit -  sha of commit before all commits in squash operation
+ * @param squashOnto  - commit to squash the `toSquash` commits onto
+ * @param lastRetainedCommitRef -  sha of commit before all commits in squash operation
  * @param commitSummary - summary of new commit made by squash
  * @param commitDescription - description of new commit made by squash
- * @returns - true if successful, false if failed
  */
 export async function squash(
   repository: Repository,
--- a/app/src/lib/git/squash.ts
+++ b/app/src/lib/git/squash.ts
@@ -1,0 +1,87 @@
+import * as Path from 'path'
+import * as FSE from 'fs-extra'
+import { getCommits, git, IGitExecutionOptions } from '.'
+import { CommitOneLine } from '../../models/commit'
+import { Repository } from '../../models/repository'
+
+/**
+ * Initiates an interactive rebase and squashes provided commits
+ *
+ * @param squash - commits to squash onto another commit
+ * @param squashOnto  - commit to squash the `squash` commits onto
+ * @param previousCommit -  sha of commit before all commits in squash operation
+ * @param commitSummary - summary of new commit made by squash
+ * @param commitDescription - description of new commit made by squash
+ * @returns - true if successful, false if failed
+ */
+export async function squash(
+  repository: Repository,
+  squash: ReadonlyArray<CommitOneLine>,
+  squashOnto: CommitOneLine,
+  logIndex: number,
+  commitSummary: string,
+  commitDescription: string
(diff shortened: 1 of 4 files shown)


app/src/lib/git/squash.ts
```diff
@@ -0,0 +1,69 @@
...
+    ...(await getCommits(repository, revRange(lastRetainedCommitRef, 'HEAD'))),
+  ]
+
+  let todoOutput = ''
+  commits.forEach(c => {
+    if (toSquash.map(sq => sq.sha).includes(c.sha)) {
+      return
+    }
+
+    if (c.sha === squashOnto.sha) {
```
sergiou87: This should be pretty straightforward, but I think I'd add some comments above these ifs to explain a bit what/why it does what it does:
```suggestion
    // Ignore commits to squash because those are written right next to the target commit
    if (toSquash.map(sq => sq.sha).includes(c.sha)) {
      return
    }

    // If it's the target commit, write a `pick` command for it, and immediately after a `squash` line for every commit to squash
    if (c.sha === squashOnto.sha) {
```

What do you think? 😳

app/src/lib/git/squash.ts
```diff
@@ -0,0 +1,69 @@
...
+        todoOutput += `squash ${sq.sha} ${sq.summary}\n`
+      })
+      return
+    }
+
+    todoOutput += `pick ${c.sha} ${c.summary}\n`
+  })
+
+  const todoPath = await getTempFilePath('squashTodo')
+  await FSE.writeFile(todoPath, todoOutput)
```
sergiou87: Nothing to worry about in this PR (and probably ever 😂 ), but if the list of commits was too long, having the whole content of the `squashTodo` in memory before writing it to a file could kill the app.

As an alternative, we could write a line at a time directly instead of building the string.

Again, not suggesting you have to do it, I'm just thinking out loud 😆

app/src/lib/git/squash.ts
```diff
@@ -0,0 +1,69 @@
...
+    repository,
+    todoPath,
+    lastRetainedCommitRef,
+    'squash',
+    `cat "${messagePath}" >`
+    // TODO: add progress
+  )
+
+  FSE.remove(todoPath)
+  FSE.remove(messagePath)
```
sergiou87: Should this be in the `finally` of a `try…catch` maybe? 🤔

sergiou87 (commented):
LGTM! Great work 😄  I only have a couple of comments/ideas, let me know what you think!

app/src/lib/git/squash.ts
tidy-dev: :D That was my original solution.. but async troubles.. But honestly didn't look into it too long. Probably easiest to just move it from a foreach to a regular for loop to make async not be mean.

app/src/lib/git/squash.ts
tidy-dev: Good point :P Went ahead and wrapped the whole thing.. cause if there was an error anywhere.. we would want to just return a error result and close out the squash action.

app/src/lib/git/squash.ts
tidy-dev: Updated to for loop/appendFile. May not have needed `for loop`... I was using `writeFile`.. when I needed to use `appendFile` originally 🤦, However, the `for loop` is also significantly more efficient which would be good under the circumstances of crazy numerous commits and this prevents having to convert commits into a modifiable array and in my next pr I realized I need to traverse the commits in reverse order so added that here and the for loop allows me to do that with out calling reverse.. which again is much more efficient. 

Also.. noticed I had unnecessarily had logic to write the pick lines twice... so refactored that too.

Basically the guts of the squash method are completely refactored.. :P

sergiou87 (approved):
LGTM! Thanks for those changes 😄 :shipit:

PR #12089, opened 2021-04-29
Fix crash reporting errors

sergiou87:
## Description

Due to a recent change, the app is crashing when the app crashes and tries to report the crash. Yes.

The problem is `remote` is `undefined` in the main process, so calling `remote.app` would crash.

### Screenshots

image

## Release notes

Notes: [Fixed] The app will send error reports when it crashes


tidy-dev (approved):
✨

PR #12104, opened 2021-04-30
Enable `remote` for the crash window

sergiou87:
## Description

This PR fixes a problem introduced when we upgraded from Electron v9 to v11: since that upgrade, a crash in the application would kill the main window but wouldn't show the alert dialog with the stack trace and other information. Instead, a crash would happen in the context of that alert dialog because `remote` there is `undefined`.

The problem was it's now a requirement to specify `enableRemoteModule: true` if you want `remote` to be available in the context of that new window.

### Screenshots

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/crash-window.ts
+++ b/app/src/main-process/crash-window.ts
@@ -38,6 +38,7 @@ export class CrashWindow {
         disableBlinkFeatures: 'Auxclick',
         nodeIntegration: true,
         spellcheck: false,
+        enableRemoteModule: true,
       },
     }
 


tidy-dev (approved):

boontee99 (commented):
n

app/src/main-process/crash-window.ts
```diff
@@ -38,6 +38,7 @@ export class CrashWindow {
         disableBlinkFeatures: 'Auxclick',
         nodeIntegration: true,
         spellcheck: false,
+        enableRemoteModule: true,
```
RandinAk, line 41:
```suggestion
        enableRemoteModule: true,
```

RandinAk: BFF

RandinAk (commented):
Yeah I'll get back

PR #12133, opened 2021-05-04
Thank you Part 1 - Thank you dialog

tidy-dev:
## Description

Adds the dialog for displaying the thank you message, release notes, confetti.

Does not yet invoke it - breaking work into a couple PRs for easier review.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/thank-you/thank-you.tsx
+++ b/app/src/ui/thank-you/thank-you.tsx
@@ -1,0 +1,154 @@
+import * as React from 'react'
+
+import { encodePathAsUrl } from '../../lib/path'
+
+import { ReleaseNote } from '../../models/release-notes'
+
+import { Dialog, DialogContent } from '../dialog'
+
+import { RichText } from '../lib/rich-text'
+import { Repository } from '../../models/repository'
+import { getDotComAPIEndpoint } from '../../lib/api'
+import { GitHubRepository } from '../../models/github-repository'
+import { Owner } from '../../models/owner'
+
+// HACK: This is needed because the `Rich`Text` component
+// needs to know what repo to link issues against.
+// Since release notes are Desktop specific, we can't
+// rely on the repo info we keep in state, so we've
+// stubbed out this repo
+const desktopOwner = new Owner('desktop', getDotComAPIEndpoint(), -1)
+const desktopUrl = 'https://github.com/desktop/desktop'
+const desktopRepository = new Repository(
+  '',
+  -1,
+  new GitHubRepository('desktop', desktopOwner, -1, false, desktopUrl),
+  true
+)
+
+const ReleaseNoteHeaderLeftUri = encodePathAsUrl(
+  __dirname,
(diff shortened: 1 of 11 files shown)


app/src/ui/thank-you/thank-you.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+import { Dialog, DialogContent } from '../dialog'
+
+import { RichText } from '../lib/rich-text'
+import { Repository } from '../../models/repository'
+import { getDotComAPIEndpoint } from '../../lib/api'
+import { GitHubRepository } from '../../models/github-repository'
+import { Owner } from '../../models/owner'
+
+// HACK: This is needed because the `Rich`Text` component
```
tidy-dev: Directly copied form the release note component - tho this is noted as a Hack. It seems there isn't a better alternative.

app/src/ui/thank-you/thank-you.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    return (
+      <div className="section">
+        <ul className="entries">{options}</ul>
+      </div>
+    )
+  }
+
+  private renderConfetti(): JSX.Element | null {
```
tidy-dev, line 49: I am not 100% sure. But, we would likely see better performance if we did this outside the renderer. Grab a div ref and pump the confetti in. Currently, if user moves their mouse during confetti, there can sometimes be a noticeable lag. 

Edit.. tried didn't help.. not sure what is causing the mouse move/lag thing unless there is a mouse move event trigger taking priority... oddly when running the performance profiler it behaves. 😛

Edit again: I cut the confetti down to 750 pieces in last PR and performance is much better and still looks good.

app/src/ui/thank-you/thank-you.tsx
```diff
@@ -0,0 +1,154 @@
...
+)
+
+const ReleaseNoteHeaderLeftUri = encodePathAsUrl(
+  __dirname,
+  'static/release-note-header-left.svg'
+)
+const ReleaseNoteHeaderRightUri = encodePathAsUrl(
+  __dirname,
+  'static/release-note-header-right.svg'
+)
```
say25: Would it be weird if we extracted this (duplicated in `release-notes-dialog.tsx`) to a common place?

sergiou87: Yeah I agree that'd be nice if possible! 👍

sergiou87 (commented):
Looks great overall!! Please consider @say25 suggestion, and :shipit:

app/src/ui/thank-you/thank-you.tsx
tidy-dev: Yep. Good point!

sergiou87 (approved):
Looks perfect, thanks for all those changes!! 🙏 🙇‍♂️ :shipit:

PR #12135, opened 2021-05-04
Thank you Part 3 - Parse thank you release notes separately

tidy-dev:
## Description

Logic to start parsing the thank you notes as their own category/type.

The logic just says that the release note must contain ' Thanks @'. Therefore, we are open to a false-positive if we ever write a release note with ' Thanks @' inside of it. But, seeing as we input the release notes, that is an unlikely scenario. Also, in that unlikely scenario there would also have to be the logged in user's handle in the release note for it to trigger the thank you logic. And, even in that case, it wouldn't the end of the world to accidenlty thank that user.. seeing as we put the word "Thanks" and their handle the same message. 😛 

This PR is also independent of part 1 and 2.

## Release notes
Notes: no-notes


sergiou87 (approved):
:shipit: ✨

PR #12136, opened 2021-05-04
Thank you Part 4 - Logic for storing when we have thanked a user

tidy-dev:
## Description

This is the app state and local storage logic for remembering when we have thanked a user and they have thrown away their card. 

Caveats.. If a user jumps between versions, they can get thanked multiple times. That would mean install version x and then installing version y that is older and in both version x and version y the user had a contribution. I am comfortable with assuming that is a rare/non-existent scenario and worst case is they just get thanked again (which is just a banner they don't have to open the card). 

Also, this one is not dependent on part 1, 2, 3, 4. 

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -269,6 +269,15 @@ export interface IAppState {
    * Whether or not the user has been introduced to the cherry pick feature
    */
   readonly hasShownCherryPickIntro: boolean
+
+  /**
+   * Record of what logged in users have been checked to see if thank you is in
+   * order for external contributions in latest release.
+   *
+   * String of the form [version-number, user-login, user-login, user-login].
+   * This will be reset for each new version.
+   */
+  readonly versionAndUsersOfLastThankYou: ReadonlyArray<string>
 }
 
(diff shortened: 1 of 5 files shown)


sergiou87 (commented):
I wonder if it'd be better to store that info as an object (with fields `version` and `userLogins`) instead of an array of strings. What do you think?

Other than that LGTM 👍

tidy-dev:
> I wonder if it'd be better to store that info as an object (with fields `version` and `userLogins`) instead of an array of strings. What do you think?
> 
> Other than that LGTM 👍

Originally, I just threw it in a string array to test the concept and move on during hack day because it was quick and easy. When going back over to prep for PR, I considered an object, but didn't think it was really worth the extra time. 

However, if you have a strong preference for this, I don't mind to refactor.

sergiou87:
Only to get rid of the `whatever[0]` and `whatever.slice(1)` you will need to extract the version or user logins I think it might be worth it, yeah 🤔

sergiou87 (approved):
Thanks for that refactor! :shipit:

PR #12137, opened 2021-05-04
Thank you

tidy-dev:
## Description

Has the logic to invoke the Thank you banner and card.

This will retroactively thank contributors in the last 10 releases (if on beta, last 10 beta, if on prod, last 10 prod)

If a user has no local storage showing they have ever been shown a thank you card (or checked to see if they made contributions), the thank you card will not specify the current version and will show all the release note they are tagged in for the last 10 releases (limitation of our release note retrieval). 

If they are pulling a new update and had been checked for contributions they made (something exists in local storage), then it will only show contributions for the current release.

### Screenshots
Below demo is hard-coded to @say25 cause he is awesome contributor and therefore has great test data. 😄 

In video, I step through:
1) As tho I have not seen a card before and have we never have checked for a card before.
2) As tho I have seen (or have been checked for) on the previous version.
3) Showing that it goes away and doesn't come back if user throws it away.

https://user-images.githubusercontent.com/75402236/117081950-fbc3e980-ad0e-11eb-8b0c-97e6cb7f9e84.mov

Video after some updates from other PRs and showing performance is much better: I can move my mouse like crazy with no trouble!

https://user-images.githubusercontent.com/75402236/117162566-e1caeb00-ad90-11eb-95bb-fff17306175e.mov

## Release notes
Notes: [Added] Thank external contributors for their work.

### DIFF
--- a/app/src/ui/thank-you/thank-you.tsx
+++ b/app/src/ui/thank-you/thank-you.tsx
@@ -1,0 +1,154 @@
+import * as React from 'react'
+
+import { encodePathAsUrl } from '../../lib/path'
+
+import { ReleaseNote } from '../../models/release-notes'
+
+import { Dialog, DialogContent } from '../dialog'
+
+import { RichText } from '../lib/rich-text'
+import { Repository } from '../../models/repository'
+import { getDotComAPIEndpoint } from '../../lib/api'
+import { GitHubRepository } from '../../models/github-repository'
+import { Owner } from '../../models/owner'
+
+// HACK: This is needed because the `Rich`Text` component
+// needs to know what repo to link issues against.
+// Since release notes are Desktop specific, we can't
+// rely on the repo info we keep in state, so we've
+// stubbed out this repo
+const desktopOwner = new Owner('desktop', getDotComAPIEndpoint(), -1)
+const desktopUrl = 'https://github.com/desktop/desktop'
+const desktopRepository = new Repository(
+  '',
+  -1,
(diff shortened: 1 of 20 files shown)


app/src/ui/thank-you/thank-you.tsx
```diff
@@ -49,7 +49,7 @@ export class ThankYou extends React.Component<IThankYouProps, {}> {
   private renderConfetti(): JSX.Element | null {
     const confetti = new Array<JSX.Element>()
 
-    const howMuchConfetti = 1500
+    const howMuchConfetti = 750
```
sergiou87, line 52: Shouldn't this also be changed?

https://github.com/desktop/desktop/blob/bf8c4170a368cc5726ee538224f34010d527151f/app/styles/ui/dialogs/_thank-you.scss#L126

sergiou87 (changes requested):
Only one thing that I _think_ needs to be changed, other than that LGTM!

app/src/ui/thank-you/thank-you.tsx
tidy-dev: Great catch! This wouldn't have hurt anything - but makes sense for them to match.

sergiou87 (approved):
🎊  :shipit: 🎉

kornbik (commented):
#12137
https://github.com/desktop/desktop/pull/12137#pullrequestreview-836590494

PR #12138, opened 2021-05-05
Fix hover highlighting for expansion handlers

sergiou87:
## Description

Due to a regression introduced in #12129, the expansion handlers wouldn't highlight anymore on hovering. A new "hoverable" class was introduced, but these handlers didn't have it.

### Screenshots

image

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #12143, opened 2021-05-05
Allow users to stash existing changes before undoing a commit

sergiou87:
## Description

If changes in the working directory overlap changes in a commit the user is going to undo, Desktop will allow the user to stash those changes before undoing the commit.

If the user already has stashed changes, Desktop will warn the user that those stashed changes will be replaced (and therefore lost).

I'd love to hear your feedback about the wording inside those confirmation dialogs.

### Screenshots

https://user-images.githubusercontent.com/1083228/117131839-4773a300-ada2-11eb-9b92-25a952baf177.mov

With no stashed changes:
image

With other stashed changes already:
image

## Release notes

Notes: [Improved] Current changes will be stashed before undoing a commit in case of conflicts

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4127,11 +4127,59 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return this._refreshRepository(repository)
   }
 
+  public async _stashAndUndoCommit(
+    repository: Repository,
+    commit: Commit
+  ): Promise<void> {
+    const gitStore = this.gitStoreCache.get(repository)
+    const repositoryState = this.repositoryStateCache.get(repository)
+    const { changesState } = repositoryState
+
+    // Stash the current changes if there are file changes in the working directory.
+    if (
+      changesState.workingDirectory.files.length > 0 &&
+      gitStore.tip.kind === TipState.Valid
+    ) {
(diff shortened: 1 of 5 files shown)


sergiou87:
I converted this PR back to a draft, because talking to @billygriffin we came up with a scenario that is still a bit painful: if a user tries to use Undo as a way to amend a previous commit, it's possible that the user ends up stashing the change that they wanted to include in the previous commit. Once it's stashed, they cannot get it out until the working directory has no changes 😞 

One idea could be making the stash step optional, so the user can: 1) cancel the Undo action, 2) stash before undo and 3) just undo directly.

image

image

I still think users will find this hard to grasp, because there are many technical limitations that they don't (need to) know…

billygriffin:
@sergiou87 In the second image when there's not an existing stash, I think it's still a bit confusing. The button should say "Undo" or "Continue" (I'm on the fence and can't remember what we do in other similar situations).

And the verbiage "Do you want to stash your changes and undo the commit?" also seems awkward with the new checkbox style. I can take a look at some possible language later today but wanted to give you feedback quickly.

caanoent (commented):
Ok

sergiou87:
We'll go with another approach here, and just warn the user about the potential data loss, but not suggest them to stash anything.

PR #12148, opened 2021-05-05
Get all previous releases to retroactively thank contributors

tidy-dev:
## Description

Make use of modified change log call to get more than 10 previous release when retroactively thanking past contributors.

### Screenshots
Check out all those PRs @say25!!!

https://user-images.githubusercontent.com/75402236/117182141-fbc1f900-ada3-11eb-8e91-c5e2137a7456.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -88,18 +88,14 @@ export async function getChangeLog(
 ): Promise<ReadonlyArray<ReleaseMetadata>> {
   const changelog =
     'https://central.github.com/deployments/desktop/desktop/changelog.json'
-  const params = []
+  const params = new URLSearchParams()
   if (__RELEASE_CHANNEL__ === 'beta') {
-    params.push('env=beta')
+    params.append('env', 'beta')
   }
   if (limit !== undefined) {
-    params.push(`limit=${limit}`)
+    params.append('limit', limit.toString())
   }
-  let query = ''
-  params.forEach(param => {
-    const prefix = query === '' ? '?' : '&'
-    query += `${prefix}${param}`
-  })
+  const query = params.toString() !== '' ? `?${params.toString()}` : ''
 
   const response = await fetch(`${changelog}${query}`)
   if (response.ok) {
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -86,18 +86,19 @@ export function getReleaseSummary(
 export async function getChangeLog(
   limit?: number
 ): Promise<ReadonlyArray<ReleaseMetadata>> {
-  const changelog =
+  const changelogURL = new URL(
     'https://central.github.com/deployments/desktop/desktop/changelog.json'
-  const params = new URLSearchParams()
-  if (__RELEASE_CHANNEL__ === 'beta') {
-    params.append('env', 'beta')
-  }
-  if (limit !== undefined) {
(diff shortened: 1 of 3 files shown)


app/src/lib/release-notes.ts
```diff
@@ -83,10 +83,23 @@ export function getReleaseSummary(
...
+    params.push('env=beta')
+  }
+  if (limit !== undefined) {
+    params.push(`limit=${limit}`)
+  }
+  let query = ''
+  params.forEach(param => {
+    const prefix = query === '' ? '?' : '&'
+    query += `${prefix}${param}`
+  })
```
say25: I sorta think this might be better off using [URLSearchParams, see MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

```suggestion
  const params = new URLSearchParams()
  if (__RELEASE_CHANNEL__ === 'beta') {
    params.append('env', 'beta')
  }
  if (limit !== undefined) {
    params.append('limit', limit)
  }
  let query = `?{params.toString()}'
```

tidy-dev: Dude! Did not know we had this URLSearchParams (bad assumption based on what it was before)! Thank you for pointing it out.

tidy-dev: Refactored! Thanks again

say25 (dismissed):

app/src/lib/release-notes.ts
```diff
@@ -83,10 +83,19 @@ export function getReleaseSummary(
...
+  const params = new URLSearchParams()
+  if (__RELEASE_CHANNEL__ === 'beta') {
+    params.append('env', 'beta')
+  }
+  if (limit !== undefined) {
+    params.append('limit', limit.toString())
+  }
+  const query = params.toString() !== '' ? `?${params.toString()}` : ''
 
   const response = await fetch(`${changelog}${query}`)
```
sergiou87: I personally prefer to handle URLs as I did in https://github.com/desktop/desktop/blob/bbe45dbcb676aad7505cbf7af8f9ba7144b8d8b1/app/src/ui/lib/update-store.ts#L173-L178

It would look something like this here:

```typescript
  const changelog =
    new URL('https://central.github.com/deployments/desktop/desktop/changelog.json')
  const params = new URLSearchParams()
  if (__RELEASE_CHANNEL__ === 'beta') {
    changelog.searchParams.set('env', 'beta')
  }
  if (limit !== undefined) {
    changelog.searchParams.set('limit', limit.toString())
  }

  const response = await fetch(changelog.toString())
```

This way you don't have to worry about concatenating `?` or any of that.

sergiou87 (commented):
Looks good! I only have a suggestion about how to handle URLs, let me know what you think 😄

app/src/lib/release-notes.ts
tidy-dev: :D See I did a brief google search for something like that when I learned of URLSearchParams.. since i figured if there was that there should be a way to for it to know about the `?` and `&`. I will refactor to that tomorrow.

say25: @sergiou87 this is even better, today I learned 🤓

tidy-dev: Done! Much cleaner - thank you @sergiou87 and @say25!

sergiou87 (approved):
😍 :shipit:

PR #12153, opened 2021-05-05
Minor changes to thank you card

billygriffin:
This just slightly changes the wording of the thank you card.

1. I removed "recent" from the banner text that lets people know they have a card available since we're also sending cards for retroactive contributions.
2. I updated the layout and verbiage slightly on the card itself. Here's how it looks with this PR.

After:
Screen Shot 2021-05-05 at 12 50 13 PM

Before:
Screen Shot 2021-05-05 at 12 55 43 PM

NOTE: I didn't have time to tackle the differentiation for the description when the Desktop version number is included. @tidy-dev Would you mind taking a crack at that? I'll add a self-review note where I'm referring to.

### DIFF
--- a/app/src/ui/thank-you/thank-you.tsx
+++ b/app/src/ui/thank-you/thank-you.tsx
@@ -104,7 +104,7 @@ export class ThankYou extends React.Component<IThankYouProps, {}> {
       >
         <DialogContent>
           <div className="container">
-            <div className="thank-you-note">{thankYouNote}.</div>
+            <div className="thank-you-note">{thankYouNote}</div>
             <div className="contributions-heading">You contributed:</div>
             <div className="contributions">
               {this.renderList(this.props.userContributions)}
--- a/app/src/ui/thank-you/thank-you.tsx
+++ b/app/src/ui/thank-you/thank-you.tsx
@@ -80,9 +80,6 @@ export class ThankYou extends React.Component<IThankYouProps, {}> {
               renderUrlsAsLinks={true}
             />
           </div>
-          <div className="thank-you-note">
-            The Desktop team wants to thank you personally.
-          </div>
         </div>
       </div>
     )
--- a/app/src/ui/thank-you/thank-you.tsx
+++ b/app/src/ui/thank-you/thank-you.tsx
@@ -84,7 +84,7 @@ export class ThankYou extends React.Component<IThankYouProps, {}> {
       </div>
     )
 
-    const thankYou = 'Thank you for all your hard work on GitHub Desktop'
+    const thankYou = "Thanks so much for all your hard work on GitHub Desktop. We're so grateful for you helping to make the app better for everyone!"
     let thankYouNote
     if (this.props.latestVersion === null) {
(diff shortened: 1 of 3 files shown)


app/src/ui/thank-you/thank-you.tsx
```diff
@@ -80,14 +80,11 @@ export class ThankYou extends React.Component<IThankYouProps, {}> {
...
           </div>
-          <div className="thank-you-note">
-            The Desktop team wants to thank you personally.
-          </div>
         </div>
       </div>
     )
 
-    const thankYou = 'Thank you for all your hard work on GitHub Desktop'
+    const thankYou = "Thanks so much for all your hard work on GitHub Desktop. We're so grateful for you helping to make the app better for everyone!"
```
billygriffin: This is the TODO mentioned in the PR body, where adding the second sentence messed up the version number. I think basically we just need to inject something like the result of the if statement below this into the end of the first sentence before the period.

app/src/ui/thank-you/thank-you.tsx
```diff
@@ -107,7 +104,7 @@ export class ThankYou extends React.Component<IThankYouProps, {}> {
       >
         <DialogContent>
           <div className="container">
-            <div className="thank-you-note">{thankYouNote}.</div>
```
billygriffin, line 110: This existed previously to facilitate the conditional addition of the version number, so I removed it here because with the second sentence we'll need to do it a bit differently.

tidy-dev:
After adapting for when only showing latest version (and some minor css tweaks to dialog header)
image

tidy-dev:
After latest tweaking
image

say25:
Just downloaded and installed `2.8.2-beta1`, I opened my Thank You card and read it but found it slightly annoying that I had to hit the "throw it away" button to actually get it to dismiss the banner, is this intentional?

billygriffin:
That's a good point @say25! How would you imagine getting it to dismiss?

jjlittlezz (approved):

say25:
There is an `X` in the Thank You card (its a touch hard to see with the fancy stars around it) but I sorta think clicking that should dismiss the banner (I kinda get not dismissing banner when clicking outside of the card to prevent accidental miss-clicks hiding your Than You)

tidy-dev:
Hmm... Reason we chose to have the throw away was so a user had the option to view it more than once.

tidy-dev (approved):
✨

tidy-dev (approved):

say25:
I guess that’s fair. Seems to disappear when you do “more actions” pull branch, switch repos etc so I think this is less of a deal than I previously thought

sergiou87 (approved):
💯 :shipit: 

One nit that can be solved in a different PR: should we remove the `Thanks @user!` part of every release note? Seems kind of redundant in this specific context 😆

jjlittlezz:
Hey bro do U Smoke I want a New Smoke Buddy

ghost (commented):
enter

PR #12159, opened 2021-05-06
FIx overflow behavior for tab-containers and select components

sergiou87:
## Description

This PR makes a few changes in the (global) style of `Select` components and tab containers.

The original problem was that very long items inside of a `Select` component would cause the tab container to grow beyond its boundaries:

image

Making sure the `select` element doesn't exceed the boundaries of its `Select` parent component, and that `tab-container` elements don't allow overflow, fixed this issue.

These changes fix the mentioned issue in:
- The Git config subscreen in the `Preferences` dialog (the one from the screenshot above)
- The Git config subscreen in the `Repository Settings` dialog
- The warning for misattributed commits (which was fixed in a different way)

Other sections that could've regressed (but didn't see anything in my testing):
- Any view where `tab-container` is used. So far that's only the `Preferences` and `Repository Settings` dialogs.
- Any view where a `Select` component is used:
	- New repo dialog
	- Publish repo dialog
	- Git config screen in Welcome flow
	- `Integrations` screen (external editor and shell)

### Screenshots

image

image

image

## Release notes

Notes: [Fixed] Long emails are truncated in the Git config

### DIFF
--- a/app/styles/ui/_select.scss
+++ b/app/styles/ui/_select.scss
@@ -21,5 +21,6 @@
 
     // Don't include the default padding from textboxish, it's too much
     padding: 0;
+    width: 100%;
   }
 }
--- a/app/styles/ui/_tab-bar.scss
+++ b/app/styles/ui/_tab-bar.scss
@@ -3,6 +3,10 @@
   flex-direction: row;
   height: var(--tab-bar-height);
 
+  .tab-container {
+    overflow: hidden;
+  }
+
   &.vertical {
     flex-direction: column;
     align-items: stretch;
--- a/app/styles/ui/_tab-bar.scss
+++ b/app/styles/ui/_tab-bar.scss
@@ -3,10 +3,6 @@
   flex-direction: row;
   height: var(--tab-bar-height);
 
-  .tab-container {
-    overflow: hidden;
-  }
-
   &.vertical {
     flex-direction: column;
     align-items: stretch;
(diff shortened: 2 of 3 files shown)


app/styles/ui/_select.scss
```diff
@@ -21,5 +21,6 @@
 
     // Don't include the default padding from textboxish, it's too much
     padding: 0;
+    width: 100%;
```
sergiou87, line 24: I think this makes sense since `select` elements inside of our `Select` components are supposed to take the whole horizontal space. Does that sound 👍 ?

app/styles/ui/_tab-bar.scss
```diff
@@ -3,6 +3,10 @@
   flex-direction: row;
   height: var(--tab-bar-height);
 
+  .tab-container {
+    overflow: hidden;
+  }
```
sergiou87: The reason for this is that, so far, we're giving fixed widths to our all the dialogs (`Preferences` and `Repository Settings`) where a tab-bar is used, and in any case it doesn't make sense to allow them to grow beyond the boundaries of their parents.

What do you think?

sergiou87: Actually, I tested again without this change and seems to work just fine, so I removed it. Nothing to see here… 👀

tidy-dev (approved):
✨  Code makes sense.  I pulled this down and did quick bounce to a few  places we use the `Select` component and tab container to do a quick gut check for impact and every things looks good.

PR #12160, opened 2021-05-06
Allow creating branches from a specific commit

### LINKED ISSUE #3474: Create a new branch from a previous commit
-Ability to checkout any previous commit, sometimes I want to go back and test something for legacy product support or to find out where a bug first started. For example in the history tab right click on a commit and select "set code to this commit" which would set the code to the way it was at the time of that commit without making a new commit that reverts all of the commits above it.

### Description

Feature request: "checkout previous commit"

sergiou87:
## Description

Fixes #3474 

This PR just adds a new action to the commit context menu to allow creating branches from a specific commit.

### Screenshots

https://user-images.githubusercontent.com/1083228/117316984-6436d600-ae89-11eb-896b-e7f148969c38.mov

## Release notes

Notes: [Added] Create branches from any commit in your repository

### DIFF
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -202,8 +202,8 @@ export class CommitListItem extends React.PureComponent<
 
     items.push({
       label: __DARWIN__
-        ? 'Create Branch From This Commit'
-        : 'Create branch from this commit',
+        ? 'Create Branch from Commit'
+        : 'Create branch from commit',
       action: () => {
         if (this.props.onCreateBranch) {
           this.props.onCreateBranch(this.props.commit.sha)
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -21,7 +21,7 @@ interface ICommitProps {
   readonly isLocal: boolean
   readonly onRevertCommit?: (commit: Commit) => void
   readonly onViewCommitOnGitHub?: (sha: string) => void
-  readonly onCreateBranch?: (targetCommitSha: string) => void
+  readonly onCreateBranch?: (commit: CommitOneLine) => void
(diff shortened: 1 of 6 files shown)


app/src/ui/history/commit-list-item.tsx
```diff
@@ -199,6 +200,17 @@ export class CommitListItem extends React.PureComponent<
       },
     ]
 
+    items.push({
+      label: __DARWIN__
+        ? 'Create Branch From This Commit'
```
tidy-dev:
```suggestion
        ? 'Create Branch from This Commit'
```
Where mac uses title case and `from` is a preposition, it should be lowercase.

tidy-dev: After testing it out, 
I noticed the revert and cherry-pick menu item  does not have the word 'This' in it.  Just a suggestion, but for consistency we could leave `This` out. 

image

tidy-dev (approved):
✨ Code looks great! Tested it out and works like a charm. I am excited to use this feature! 

A suggestion: (and if you think it is worth it - could be a separate PR)
We have easy access to the commit summary at this point. It may be more user friendly to show the summary and the sha? 

We do this on the progress dialogs for rebase/cherry-picking so I also think it would be more consistent. (Below is just a quick mock up, there may be a better way to put it to be friendly to long summaries and maybe styling should use a bold/emphasized instead of in quotes)
image

Side note:  "from your repository" may be redundant?

tidy-dev (changes requested):
... context switched while reviewing this.. and forgot about the title case comment, I think that should be updated before merging. :)

sergiou87:
I implemented both of your suggestions, I think they make a lot of sense, so thank you!! ❤️ 

image

tidy-dev (approved):
✨  Thanks for the changes!

PR #12165, opened 2021-05-06
Allow creating a branch by clicking `Enter` in the text box

### LINKED ISSUE #12154: Create a new branch from filtering by just hitting `Enter`

This request is via @cheshire137. To create a new branch in the branch filter process, you have to click the blue button with your mouse or tab to it. 

### Proposed solution

It would be nice to just be able to hit `Enter` to pop the dialog to create the branch after you've typed the new branch name. I'm not super worried about lowering the friction because you still have the dialog to verify whether it's what you meant to do.

### Additional context

Screen Shot 2021-05-05 at 3 28 24 PM

tsvetilian-ty:
Closes #12154

## Description

- Create a new branch on `Enter` clicked

### Screenshots
Animation

### DIFF
--- a/app/src/ui/branches/branch-list.tsx
+++ b/app/src/ui/branches/branch-list.tsx
@@ -81,6 +81,9 @@ interface IBranchListProps {
   /** Callback to fire when the filter text is changed */
   readonly onFilterTextChanged: (filterText: string) => void
 
+  /** Callback to fire when the filter text is changed */
+  readonly onEnterClicked?: (text: string) => void
+
   /** Can users create a new branch? */
   readonly canCreateNewBranch: boolean
 
@@ -188,6 +191,7 @@ export class BranchList extends React.Component<
         renderGroupHeader={this.renderGroupHeader}
         onItemClick={this.onItemClick}
         onSelectionChanged={this.onSelectionChanged}
+        onEnterClicked={this.props.onEnterClicked}
         groups={this.state.groups}
         invalidationProps={this.props.allBranches}
         renderPostFilter={this.onRenderNewButton}
--- a/app/src/ui/branches/branch-list.tsx
+++ b/app/src/ui/branches/branch-list.tsx
@@ -81,9 +81,6 @@ interface IBranchListProps {
   /** Callback to fire when the filter text is changed */
   readonly onFilterTextChanged: (filterText: string) => void
 
-  /** Callback to fire when the filter text is changed */
-  readonly onEnterClicked?: (text: string) => void
-
   /** Can users create a new branch? */
   readonly canCreateNewBranch: boolean
 
@@ -191,7 +188,7 @@ export class BranchList extends React.Component<
         renderGroupHeader={this.renderGroupHeader}
         onItemClick={this.onItemClick}
         onSelectionChanged={this.onSelectionChanged}
-        onEnterClicked={this.props.onEnterClicked}
+        onEnterPressedWithoutFilteredItems={this.onCreateNewBranch}
         groups={this.state.groups}
         invalidationProps={this.props.allBranches}
         renderPostFilter={this.onRenderNewButton}
--- a/app/src/ui/lib/filter-list.tsx
+++ b/app/src/ui/lib/filter-list.tsx
@@ -113,7 +113,7 @@ interface IFilterListProps<T extends IFilterListItem> {
   ) => void
 
   /** Called when the Enter key is pressed in field of type search */
-  readonly onEnterClicked?: (text: string) => void
+  readonly onEnterPressedWithoutFilteredItems?: (text: string) => void
 
   /** The current filter text to use in the form */
   readonly filterText?: string
@@ -253,7 +253,7 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
         placeholder={this.props.placeholderText || 'Filter'}
         className="filter-list-filter-field"
         onValueChanged={this.onFilterValueChanged}
-        onEnterClicked={this.onEnterClicked}
+        onEnterPressed ={this.onEnterPressed}
         onKeyDown={this.onKeyDown}
         value={this.props.filterText}
         disabled={this.props.disabled}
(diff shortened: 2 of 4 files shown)


app/src/ui/lib/filter-list.tsx
```diff
@@ -112,6 +112,9 @@ interface IFilterListProps<T extends IFilterListItem> {
     event: React.KeyboardEvent<HTMLInputElement>
   ) => void
 
+  /** Called when the Enter key is pressed in field of type search */
+  readonly onEnterClicked?: (text: string) => void
```
cheshire137: I could see calling this `onEnter`, `onEnterDown` (like `onKeyDown`), or `onEnterPressed`. 'Clicked' doesn't seem right for a keyboard key.

app/src/ui/lib/text-box.tsx
```diff
@@ -271,6 +274,14 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
           this.inputElement.blur()
         }
       }
+    } else if (
+      this.props.type === 'search' &&
+      event.key === 'Enter' &&
+      value !== undefined &&
+      value !== '' &&
+      this.props.onEnterClicked !== undefined
```
cheshire137: Would just checking `this.props.onEnterClicked` be sufficient without the `!== undefined`?

app/src/ui/lib/text-box.tsx
```diff
@@ -271,6 +274,14 @@ export class TextBox extends React.Component<ITextBoxProps, ITextBoxState> {
           this.inputElement.blur()
         }
       }
+    } else if (
+      this.props.type === 'search' &&
+      event.key === 'Enter' &&
+      value !== undefined &&
```
cheshire137, line 280: Should this be `typeof value !== 'undefined'` or is that not a TypeScript thing?

cheshire137 (commented):
Thank you for doing this! 🙇‍♀️

app/src/ui/lib/text-box.tsx
tidy-dev: Thank you for jumping in and reviewing this! Our team has a code convention of not using falsy comparisons. Technically, it would work without the `!== undefined` but we prefer to be explicit. The only time we do that is if it is a boolean that type-wise cannot be undefined.

tidy-dev (changes requested):
@tsvetilian-ty Thank you for working on this! I tested it out and it works great! I concur with @cheshire137's comment on the preferred naming. If you could update that, I would love to get this merged!

app/src/ui/lib/filter-list.tsx
sergiou87: Agreed! I'd say we use `onEnterPressed` as we do in https://github.com/desktop/desktop/blob/28a2ee45a220dc87592e5c8e4666c4a6006b4f2b/app/src/ui/cherry-pick/choose-target-branch.tsx#L95

app/src/ui/branches/branch-list.tsx
```diff
@@ -81,6 +81,9 @@ interface IBranchListProps {
   /** Callback to fire when the filter text is changed */
   readonly onFilterTextChanged: (filterText: string) => void
 
+  /** Callback to fire when the filter text is changed */
+  readonly onEnterClicked?: (text: string) => void
```
sergiou87: I don't think we need to expose this: `BranchList` already has a method to create a branch: https://github.com/desktop/desktop/blob/28a2ee45a220dc87592e5c8e4666c4a6006b4f2b/app/src/ui/branches/branch-list.tsx#L281-L285

app/src/ui/lib/filter-list.tsx
```diff
@@ -372,6 +376,13 @@ export class FilterList<T extends IFilterListItem> extends React.Component<
     }
   }
 
+  private onEnterClicked = (text: string) => {
+    const rows = this.state.rows.length
+    if (rows === 0 && this.props.onEnterClicked) {
```
sergiou87: We prefer to use explicit checks for `undefined`, and I think we should also check that `text` has something in it:
```suggestion
    if (rows === 0 && text.trim().length > 0 && this.props.onEnterClicked !== undefined) {
```

app/src/ui/lib/filter-list.tsx
sergiou87: Thinking more about it, this event will be propagated not only when <kbd>Enter</kbd> is pressed, but also when the list is empty after filtering the results, so maybe we should rename it to something more generic?

Something like… `onEnterPressedWithoutFilteredItems`?

app/src/ui/lib/text-box.tsx
sergiou87: Not a TypeScript expert, but I think that's not needed and `value !== undefined` is just fine 😄

sergiou87 (changes requested):

app/src/ui/lib/filter-list.tsx
tidy-dev: For more context, enter does does check out a branch when there are results. I agree with @sergiou87, clarification in the code would be nice for future devs.

DUNYAM01 (dismissed):

tsvetilian-ty:
Thank you for your suggestions, everyone. 🥳

app/src/ui/lib/text-box.tsx
cheshire137: Good deal, thank you @tidy-dev!

sergiou87 (dismissed):
It's looking good to me now, thanks for taking the tame to address our comments!! 🙇‍♂️ 

I also tested it and it's 😍

sergiou87 (approved):

tidy-dev (approved):
✨  Looking good! Thank you for the updates.

jwsch51:
text[](url)

1972chris (commented):
Enter

PR #12177, opened 2021-05-08
Better text token parsing of new line

### LINKED ISSUE #12105: URL clickable also in description in github Desktop
### URL clickable also in description in github Desktop like the title

image
image

### How will it benefit Desktop and its users?
Save times to not copy/paste

tsvetilian-ty:
Closes #12105

### Description

- If the new line starts with a link or a mention, the text token parser can't recognize them as such.

## Before
before

## After
after

## Tokenization result BEFORE changes
image

## Tokenization result AFTER changes
image

### DIFF
--- a/app/src/lib/text-token-parser.ts
+++ b/app/src/lib/text-token-parser.ts
@@ -180,7 +180,7 @@ export class Tokenizer {
     // to ensure this isn't part of an email address, peek at the previous
     // character - if something is found and it's not whitespace, bail out
     const lastItem = this.getLastProcessedChar()
-    if (lastItem && lastItem !== ' ') {
+    if (lastItem && !/\s/.test(lastItem)) {
       return null
     }
 
@@ -212,7 +212,7 @@ export class Tokenizer {
     // to ensure this isn't just the part of some word - if something is
     // found and it's not whitespace, bail out
     const lastItem = this.getLastProcessedChar()
-    if (lastItem && lastItem !== ' ') {
+    if (lastItem && !/\s/.test(lastItem)) {
       return null
     }
 
--- a/app/src/lib/text-token-parser.ts
+++ b/app/src/lib/text-token-parser.ts
@@ -279,6 +279,12 @@ export class Tokenizer {
             this.scanForHyperlink(text, i)
           )
           break
+        
+        case '\n':
+          this.append(element)
+          this.append(' ')
+          i++
+          break
 
         default:
           this.append(element)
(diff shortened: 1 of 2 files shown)


app/src/lib/text-token-parser.ts
```diff
@@ -280,6 +280,12 @@ export class Tokenizer {
           )
           break
 
+        case '\n':
+          this.append(element)
+          this.append(' ')
```
sergiou87: @tsvetilian-ty can you explain a bit how these changes fix the issue? At first sight, I don't see how adding this space fixes the link parsing 🤔

sergiou87: Oh I just saw why (I think):
https://github.com/desktop/desktop/blob/2a4d2ef3ea73af86fc5a3ad2732f4a08899d2cc9/app/src/lib/text-token-parser.ts#L215

and https://github.com/desktop/desktop/blob/2a4d2ef3ea73af86fc5a3ad2732f4a08899d2cc9/app/src/lib/text-token-parser.ts#L183

I think it would be more appropriate to just check for all whitespace characters instead of only space, using for example `/\s/.test(lastItem)` instead of `lastItem !== ' '`, or even `lastItem.trim() === ''`

sergiou87 (changes requested):

sergiou87:
Hey @tsvetilian-ty! 👋  Just curious to know if you have time to address the changes above. If you don't have time or whatever, just let us know and we'll take care of the rest!

tsvetilian-ty:
> Hey @tsvetilian-ty! 👋  Just curious to know if you have time to address the changes above. If you don't have time or whatever, just let us know and we'll take care of the rest!

Hello @sergiou87 Tomorrow, I'll have time to address the changes. Sorry for the delayed response 😞

sergiou87 (changes requested):
Thanks @tsvetilian-ty!!

I'm sorry I wasn't very clear in my response 😭  But with those changes you added to check for `/\s/.test…`, the rest of the changes in the PR shouldn't be necessary: you can remove both the `case '\n':` blocks and the change in the test.

sergiou87 (approved):
Beautiful, thank you!!! ❤️

tsvetilian-ty:
> Thanks @tsvetilian-ty!!
> 
> I'm sorry I wasn't very clear in my response 😭 But with those changes you added to check for `/\s/.test…`, the rest of the changes in the PR shouldn't be necessary: you can remove both the `case '\n':` blocks and the change in the test.

You were clear I just forgot to remove them 😢 Because I already changed the tests and they passed, so I didn't notice that I forgot to remove the new line token 🤦

PR #12187, opened 2021-05-10
Commit message dialog

tidy-dev:
## Description

This is a commit message dialog that makes use of the current commit message component to be used for squashing (and potentially other things..)

Making use of the existing component lets us automatically have nice features like:
- co authors
- no write access warning
- branch protection warning
- issue auto completion for GitHub repo
- emoji's
- author avatar
- (Also any existing commit message protections like null checks and such)

A few things were intentionally modified to be optional and not included in the dialog implementation such as:
- repositoryAccount -> commit misattribution: Due to avoiding dialog on dialog type behavior and this is planned to be used for niche type scenarios (squashing/amending) not as the rule for committing
- tutorial nudging: This currently isn't planned to be used during tutorial repo
- isCommitting: For current use (squashing), we will likely be moving to a progress dialog so not needed. (tho simply exposing it on the dialog and passing to the commit message component would allow for it's use)
- anyFilesSelected: Since opening the dialog is likely indicating you have something to commit, we don't need to check this for disabling the dialog's commit button based on files selected in the changes list. This also doesn't make sense in the context of squashing.

Other notes:
Currently another built in feature of the commit-message component is making some dispatcher calls to modify the changes state. This is for things such as remembering the coAuthor selection or typed in commit message when navigating between screens. In this PR, I did not disable this... may end up disabling it during squashing implementation.. but for now wasn't sure if I wanted to just make squashing be aware and reset these things. (cause what if this was a desired feature.. 🤔 )

### Screenshots
image

## Release notes
Notes: no-notes


app/src/lib/autocompletion-providers.ts
```diff
@@ -0,0 +1,50 @@
+import {
```
sergiou87: Maybe this file should be located in `app/src/ui/autocompletion` like the providers themselves?

app/src/ui/changes/commit-message.tsx
```diff
@@ -387,7 +391,7 @@ export class CommitMessage extends React.Component<
       action: this.onToggleCoAuthors,
       enabled:
         this.props.repository.gitHubRepository !== null &&
-        !this.props.isCommitting,
+        !(this.props.isCommitting === true),
```
sergiou87: This should work too, right?
```suggestion
        this.props.isCommitting !== true,
```

app/src/ui/changes/commit-message.tsx
```diff
@@ -572,30 +576,32 @@ export class CommitMessage extends React.Component<
     }
   }
 
-  public render() {
+  private renderSubmitButton() {
+    const { isCommitting } = this.props
     const isSummaryWhiteSpace = this.state.summary.match(/^\s+$/g)
     const buttonEnabled =
-      this.canCommit() && !this.props.isCommitting && !isSummaryWhiteSpace
+      this.canCommit() && !(isCommitting === true) && !isSummaryWhiteSpace
```
sergiou87: Is this the same?
```suggestion
      this.canCommit() && isCommitting !== true && !isSummaryWhiteSpace
```

app/src/ui/commit-message/commit-message-dialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { Dispatcher } from '../dispatcher'
+import { Repository } from '../../models/repository'
+import { Dialog, DialogContent } from '../dialog'
+import { ICommitContext } from '../../models/commit'
+import { CommitIdentity } from '../../models/commit-identity'
+import { ICommitMessage } from '../../models/commit-message'
+import { IAutocompletionProvider } from '../autocompletion'
+import { IAuthor } from '../../models/author'
+import { CommitMessage } from '../changes/commit-message'
+interface ICommitMessageDialogProps {
```
sergiou87, line 12: Nit: should we leave a blank line above this one?
```suggestion

interface ICommitMessageDialogProps {
```

sergiou87 (commented):
Only a few minor comments, other than that LGTM! 😄

app/src/ui/commit-message/commit-message-dialog.tsx
tidy-dev: I always think there is a space when I see this. :P 
image

sergiou87 (approved):
Code looks good now, and I tested it and couldn't find regressions in the commit message component 😄 

:shipit: ❗

Issue #12041, opened 2021-04-22
Conditional includes in .gitconfig not supported


ferenczy:
### Describe the bug

The GitHub Desktop client doesn't support [.gitconfig's Conditional includes](https://git-scm.com/docs/git-config#_conditional_includes).

I'm using a single machine for both work and private development. Each of those requires a distinct Git user (name and email). I don't want to configure the Git user for each and every repository individually (as I usually forget to do so and notice it when there're several commits under a wrong user already). That's why I'm using the _Conditional includes_. That allows me to have an **individual configuration for a specific directory and all its subdirectories**, for example:

- `~/development/` for personal projects with a personal user name/email
- `~/work/` for projects I'm working on in my job

My base `~/.gitconfig` contains something like the following:

```gitconfig
# this is a global configuration valid for the whole machine
[user]
    name = My Private Name With Nickname
    email = private@email

# the following section tells Git to use (actually merge to this config) the config file `~/.gitconfig-work` 
# for all repositories located in the directory `~/work/` and all its subdirectories
[includeIf "gitdir/i:~/work/**"]
    # include the content of `~/.gitconfig-work` here, if it contains any directives which have been already defined, it will overwrite them
    path = ~/.gitconfig-work
```

The content of `~/.gitconfig-work`:

```gitconfig
[user]
    name = My Work Name
    email = corporate@email
```

And that's it. All repositories under the directory `~/work/` will be using the configuration from the base `~/.gitconfig` merged with `~/.gitconfig-work` (which overwrites anything already defined in the base config). Repositories located anywhere outside of that directory will be using just the configuration from the base config.

And Git reflects that. When I execute `git config user.email` under the directory `~/work/` or its subdirectories, it prints the e-mail configured in `~/.gitconfig-work`.

The problem is that GitHub Desktop is always showing the configuration from the base config. When I open the _Repository_ menu > _Repository settings..._ and select _Git config_, the option _Use my global Git config_ is selected and the name and the email from the base config are shown there. And, of course, it's also used as a commit author.

### Version & OS

GitHub 2.7.2
Windows 10 Pro 64-bit

### Steps to reproduce the behavior

1. Create two configuration files as described above, one included in the other one
2. Create a new repository under the directory you have specified in the `includeIf` directive (`gitdir:...`)
3. Check if the included config is correctly applied by executing `git config user.email` in your repository
4. Check the _Git config_ section in the _Repository settings..._ dialog in GitHub Desktop

### Expected behavior

[...]

PR #12203, opened 2021-05-13
Python dependency upgraded to version 3.x

ferenczy:
## Description

According to the [node-gyp's GitHub page](https://github.com/nodejs/node-gyp), it officially requires Python 3.x (3.6 to 3.9) instead of 2.x on all platforms. So unless Python 2.x is a dependency of anything else, it shouldn't be needed anymore.

I have updated the documentation (development environment set up) to reflect that.


docs/contributing/setup-windows.md
```diff
@@ -74,16 +74,19 @@ Open a shell and run this command:
...
-don't have Python installed. You can install Python 2.7 from the
+don't have Python installed. You can install Python 3.9 from the
 [Python website](https://www.python.org/downloads/windows/).
 
 **Python installation notes:**
 
- - Let Python install into the default suggested path (`c:\Python27`), otherwise
+ - Let Python install into the default suggested path (`c:\Python39`), otherwise
    you'll have to configure `node-gyp` manually to look at a different path.
+ - In the case you already have or need to have Python on a different path, set the `PYTHON` environment variable or the npm's `python` config key to your Python's executable path:
```
Rexogamer: Minor wording changes:
```suggestion
 - In case you already have or need to have Python on a different path, set the `PYTHON` environment variable or npm's `python` config key to your Python's executable path:
```

Rexogamer:
One other thing: there's also [these](https://github.com/desktop/desktop/blob/development/.python-version) [two](https://github.com/desktop/desktop/blob/development/.tool-versions) files that will likely need updating

sergiou87:
Good point, thanks for updating the docs! ❤️

docs/contributing/setup-windows.md
ferenczy: Thank you @Rexogamer! 💛 I'm aware that I'm overusing articles, probably in a hope that at least some of them are actually placed correctly, since I have never properly learned the rules how to use them. 😊 Disclaimer: English is not my mother tongue.

Off topic: I would appreciate if you can suggest a good resource to learn how to use the English articles! 🙂

ferenczy:
> One other thing: there's also [these](https://github.com/desktop/desktop/blob/development/.python-version) [two](https://github.com/desktop/desktop/blob/development/.tool-versions) files that will likely need updating

Good catch @Rexogamer, thanks! Files updated.

> Good point, thanks for updating the docs! ❤️

Thanks @sergiou87! Docs are easy. ;) I just struggle a bit with building that app, but that's for another issue (#12041).

staabm:
Does this change has any performance implications?

Rexogamer:
> Does this change has any performance implications?

As far as I can tell, probably not - if anything this is just a docs change, as most people probably run node-gyp with py3 anyway

ferenczy:
@staabm I don't expect it may have any performance implications. As @Rexogamer wrote, on the GitHub Desktop's side, it's just a docs and a dependency manager configuration (files `.tool-versions` and `.python-version`) update.

Python should be only used by [node-gyp](https://github.com/nodejs/node-gyp) to build native Node.js modules. Node-gyp officially requires Python 3 and it's the only supported version now. The Python 2 support has been [removed in February 2021](https://github.com/nodejs/node-gyp/commit/a78b584236e92a9469f72916c55ba83e9819ddea#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

say25:
I believe this is build time only ~at the same time though, part of me thinks we should upgrade node-gyp in this PR otherwise we are just saying use Python3 for almost no reason~

Edit: I think node-gyp is pulled transitively so I guess unless we explicitly upgrade it or upgrade what pulls it in this is probably just fine IMHO

sergiou87 (approved):

docs/contributing/setup-windows.md
BONDROVER: 🤝🏽⚡️

PR #12204, opened 2021-05-13
Squashing: Update squash logic to respect log history in order to reduce conflicts

tidy-dev:
## Description

Updated to squash commits in the order that they appear on the log. 

This may feel trivial at first glance, but given that a user adds a line to to the bottom of a file in commit A and then adds another line to the bottom in commit B. If the user drags A onto B indicating they want to squash A onto B. The original logic would reorder to B than A and squash resulting in a conflict as would now modify the last line of the file with B's changes which would now be the same line as in A.  However, had the user dragged B onto A, it would have maintained the order of A to B and no conflicts would arise. We are assuming that the user's goal was just to merge A and B and did not want any reordering to happen, they logically just want them to be one commit where 2 lines were added. Thus, this update will always maintain squashing commits order in respect to their history on the log.

Note: Reordering will still happen in the scenario a user has commits A, B, and C and the drag A onto C. This will respect the users choice of direction so it will be come B, A-C as opposed to if they dragged C onto A which will become A-C, B. But the order of A and C to each other will still be maintained based on history to reduce conflicts. 

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/squash.ts
+++ b/app/src/lib/git/squash.ts
@@ -66,20 +66,21 @@ export async function squash(
     const toReplayAfterSquash = []
     // Traversed in reverse so we do oldest to newest (replay commits)
     for (let i = commits.length - 1; i >= 0; i--) {
-      if (toSquashShas.has(commits[i].sha)) {
-        // If it is a toSquash commit and we have not found the squashOnto
-        // commit yet we want to keep track of them in the order of the log.
-        // Thus, we use a new `toReplayAtSquash` array and not trust that what
-        // was sent is in the order of the log.
+      const commit = commits[i]
+      if (toSquashShas.has(commit.sha)) {
+        // If it is toSquash commit and we have found the squashOnto commit, we
+        // can go ahead and squash them (as we will hold any picks till after)
         if (foundSquashOntoCommitInLog) {
           await FSE.appendFile(
             todoPath,
-            `squash ${commits[i].sha} ${commits[i].summary}\n`
+            `squash ${commit.sha} ${commit.summary}\n`
           )
         } else {
-          // However, if we have found the squashOnto commit, we can go ahead
-          // and squash them (as we will hold any picks till after)
-          toReplayAtSquash.push(commits[i])
+          // However, if we have not found the squashOnto commit yet we want to
+          // keep track of them in the order of the log. Thus, we use a new
+          // `toReplayAtSquash` array and not trust that what was sent is in the
+          // order of the log.
+          toReplayAtSquash.push(commit)
         }
 
         continue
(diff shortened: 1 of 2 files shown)


app/test/unit/git/squash-test.ts
```diff
@@ -101,47 +101,50 @@ describe('git/cherry-pick', () => {
 
   it('squashes multiple commit non-sequential commits (reorders, non-conflicting)', async () => {
```
tidy-dev, line 102: This refactor did not break this test. I just decided to enhance it to verify that the change made in the refactor does in fact result in the squashing commits being order with respect to the log.

app/test/unit/git/squash-test.ts
```diff
@@ -101,47 +101,50 @@ describe('git/cherry-pick', () => {
...
+      ''
     )
 
     expect(result).toBe(RebaseResult.CompletedWithoutError)
 
     // From oldest to newest, log should look like:
     // - initial commit - log[2]
-    // - the squashed commit 'Test Summary' - log[1]
-    // - 'third' commit - log[0]
+    // - the squashed commit 'first third fifth` - order by log history
```
tidy-dev, line 131: Before this refactor, the squashed commit would have been 'third first fifth' as third was the specified squashOnto commit and the toSquash array was ordered as first, fifth. Now the test correctly verifies that it maintains the log history in the operation.

tidy-dev: Edit: Updated the test to send in toSquash to be in opposite log order to show even more that squash is smart enough to order it correctly.

app/src/lib/git/squash.ts
```diff
@@ -49,28 +62,70 @@ export async function squash(
...
     todoPath = await getTempFilePath('squashTodo')
     let foundSquashOntoCommitInLog = false
-    // need to traverse in reverse so we do oldest to newest (replay commits)
-
+    const toReplayAtSquash = []
+    const toReplayAfterSquash = []
+    // Traversed in reverse so we do oldest to newest (replay commits)
     for (let i = commits.length - 1; i >= 0; i--) {
-      // Ignore commits toSquash because those are written right after the squashOnto commit
       if (toSquashShas.has(commits[i].sha)) {
```
sergiou87: I just noticed we're referencing `commits[i]` multiple times. Not super costly, but probably worth extracting it to a `const commit = commits[i]` local variable?

app/src/lib/git/squash.ts
```diff
@@ -49,28 +62,70 @@ export async function squash(
...
+        if (foundSquashOntoCommitInLog) {
+          await FSE.appendFile(
+            todoPath,
+            `squash ${commits[i].sha} ${commits[i].summary}\n`
+          )
+        } else {
+          // However, if we have found the squashOnto commit, we can go ahead
+          // and squash them (as we will hold any picks till after)
+          toReplayAtSquash.push(commits[i])
+        }
```
sergiou87: Shouldn't these comments be the other way around?

sergiou87 (changes requested):
Looking good overall! Only found a couple of nits 😄

app/src/lib/git/squash.ts
tidy-dev: Ha! Yep!

sergiou87 (approved):
Thanks for those changes! 😍 :shipit: ✨

PR #12207, opened 2021-05-13
Add detection for x64 emulation on Windows ARM

sergiou87:
## Description

This PR adds a [new module](https://github.com/desktop/node-detect-arm64-translation) that tells us if an x64 build is being emulated on an ARM64 device. With it, we will get metrics of how many users are nowadays running (and crashing) the app on Windows ARM, and will allow them, eventually, to automagically switch to arm64 updates :_D

We'll get rid of this module once Electron implements the same functionality (I intend to submit a PR Soon™ 🤞 ).

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #12217, opened 2021-05-14
Squash by context menu

tidy-dev:
## Description

- Add squashing via the context menu which is feature-flagged to development.
- Makes persisting the coauthor and commit message to the changes state in the commit-message optional. Otherwise if you squash something and then navigate to changes tab, the commit message input is updated to what was done in the dialog.
- Added ability to get the commit's body without coauthor trailers so they are not duplicated during squashing and commit description is less cluttered.
- Also handled the scenario where user squashes the root commit of a branch which cannot be rebased using traditional commit ref syntax of `sha^` or `HEAD~n`.  In doing so, I needed to make the getCommits method flexible enough to not specify a rev-range (get all commits in the branches history). 

### Screenshots
https://user-images.githubusercontent.com/75402236/118318043-d4b3a780-b4c6-11eb-944f-aa1022399f80.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/unique-coauthors-as-authors.ts
+++ b/app/src/lib/unique-coauthors-as-authors.ts
@@ -1,0 +1,18 @@
+import _ from 'lodash'
+import { IAuthor } from '../models/author'
+import { Commit } from '../models/commit'
+import { GitAuthor } from '../models/git-author'
+
+export function getUniqueCoauthorsAsAuthors(
+  commits: ReadonlyArray<Commit>
+): ReadonlyArray<IAuthor> {
+  const allCommitsCoAuthors: GitAuthor[] = _.flatten(
+    commits.map(c => c.coAuthors)
+  )
+
+  const uniqueCoAuthors = _.uniqBy(allCommitsCoAuthors, a => a.email && a.name)
+
+  return uniqueCoAuthors.map(ca => {
+    return { name: ca.name, email: ca.email, username: null }
+  })
+}
--- a/app/src/lib/unique-coauthors-as-authors.ts
+++ b/app/src/lib/unique-coauthors-as-authors.ts
@@ -10,7 +10,10 @@ export function getUniqueCoauthorsAsAuthors(
     commits.map(c => c.coAuthors)
   )
 
-  const uniqueCoAuthors = _.uniqBy(allCommitsCoAuthors, a => a.email && a.name)
+  const uniqueCoAuthors = _.uniqWith(
+    allCommitsCoAuthors,
+    (a, b) => a.email === b.email && a.name === b.name
+  )
 
   return uniqueCoAuthors.map(ca => {
     return { name: ca.name, email: ca.email, username: null }
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -60,6 +60,13 @@ interface ICommitListProps {
   /** Callback to fire to cherry picking the commit  */
   readonly onCherryPick: (commits: ReadonlyArray<CommitOneLine>) => void
 
+  /** Callback to fire to squashing commits  */
+  readonly onSquash: (
+    toSquash: ReadonlyArray<Commit>,
+    squashOnto: Commit,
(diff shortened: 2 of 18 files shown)


app/src/lib/unique-coauthors-as-authors.ts
```diff
@@ -0,0 +1,18 @@
...
+import { GitAuthor } from '../models/git-author'
+
+export function getUniqueCoauthorsAsAuthors(
+  commits: ReadonlyArray<Commit>
+): ReadonlyArray<IAuthor> {
+  const allCommitsCoAuthors: GitAuthor[] = _.flatten(
+    commits.map(c => c.coAuthors)
+  )
+
+  const uniqueCoAuthors = _.uniqBy(allCommitsCoAuthors, a => a.email && a.name)
```
sergiou87: How is this `uniqBy` supposed to work? 🤔  If I understand it correctly, the way is written it will always compare by name, no?

Unless I'm missing something, I _think_ we want to use `uniqWith` with a comparator function that checks the email & name between two author instances.

app/test/unit/unique-coauthors-as-authors-test.ts
```diff
@@ -0,0 +1,94 @@
...
+  })
+
+  it('can get multiple coauthors on multiple commits', async () => {
+    const first_email = 'tidy-dev@github.com'
+    const first_name = 'tidy-dev'
+    const firstTrailers = [buildTestCoAuthorTrailer(first_email, first_name)]
+
+    const second_email = 'sergiou87@github.com'
+    const second_name = 'Sergio'
+    const secondTrailers = [buildTestCoAuthorTrailer(second_email, second_name)]
```
sergiou87, line 93: Probably worth adding a case where there are commits with same name but different email (and vice-versa)? Given I _think_ `uniqBy` doesn't work as intended, that would help covering those scenarios.

app/src/ui/history/commit-list.tsx
```diff
@@ -171,6 +179,19 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
+  private onSquash = (toSquash: ReadonlyArray<Commit>, squashOnto: Commit) => {
+    const indexes = [...toSquash, squashOnto].map(v =>
+      this.props.commitSHAs.findIndex(sha => sha === v.sha)
+    )
+    const maxIndex = Math.max(...indexes)
+    const lastIndex = this.props.commitSHAs.length - 1
+    /* If the commit is the first commit in the branch, you cannot reference it
+    using the sha, you must use the --root flag */
+    const lastRetainedCommitRef =
+      maxIndex === lastIndex ? '--root' : `${this.props.commitSHAs[maxIndex]}^`
```
sergiou87: What is `--root` exactly? Is it something git specific? If so, is it ok to "leak" that at this point or is there a better way to indicate this (like a `null`)? 🤔

sergiou87 (changes requested):

app/src/lib/unique-coauthors-as-authors.ts
tidy-dev: Hmm.. I have not used lodash functions before and was copying an example... But I can see why this shouldn't make sense :D

tidy-dev: Done.

sergiou87 (approved):
Thanks for those changes ✨  Awesome job!!! 😍 :shipit:

app/src/ui/history/commit-list.tsx
tidy-dev: Here is some documentation on the `--root flag` for reference. https://git-scm.com/docs/git-rebase/2.19.0#Documentation/git-rebase.txt---root

Refactored to `null` - I thought of doing that but thought specifying `--root` was more direct/readable.. But, can see where we don't want to pass `git` specific flags around and `null` does make sense because the lastRetainedCommitRef is the commit before the last commit in squash operation which does not exist in this context so `null` make sense.

PR #12226, opened 2021-05-17
Drag/Drop Bubble Up Not Needed

tidy-dev:
## Description

This bubble serves no purpose now that we have the drag and drop manager that we can inject on the branch. This bubble up was a remnant of other drag and drop designs.

## Release notes
Notes: no-notes


sergiou87 (approved):
✨ :shipit: ✨

PR #12232, opened 2021-05-18
Refactor storing drag data into drag and drop manager

tidy-dev:
## Description

Moving the tracking of data associated with a current drag into the drag and drop manager and using the drag and drop manager solely for determining when a drag is taking place as opposed to a cherry-pick state variable.

### Screenshots
Notes: no-notes

### DIFF
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -26,6 +26,7 @@ import { renderDefaultBranch } from './branch-renderer'
 import { IMatches } from '../../lib/fuzzy-find'
 import { startTimer } from '../lib/timing'
 import { dragAndDropManager } from '../../lib/drag-and-drop-manager'
+import { DragType } from '../../models/drag-drop'
 
 interface IBranchesContainerProps {
   readonly dispatcher: Dispatcher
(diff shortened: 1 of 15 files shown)


app/src/ui/branches/branches-container.tsx
```diff
@@ -229,15 +219,23 @@ export class BranchesContainer extends React.Component<
...
       return
     }
 
-    this.props.dispatcher.setCherryPickCreateBranchFlowStep(
+    await this.props.dispatcher.setCherryPickCreateBranchFlowStep(
       this.props.repository,
       ''
     )
+
+    this.props.dispatcher.showPopup({
```
tidy-dev, line 233: This popup call was in the dispatcher `setCherryPickCreateBranchFlowStep` and was conditional on if it was invoked by a drag event as it can be invoked for either context menu or drag flow. Since we know this is the drag flow, we can just call it here since it is just another dispatcher call.

app/src/ui/branches/branches-container.tsx
```diff
@@ -397,11 +394,18 @@ export class BranchesContainer extends React.Component<
...
         this.props.repository,
         branch
       )
     }
   }
+
+  private onDropOntoCurrentBranch = () => {
+    if (dragAndDropManager.isDragOfType(DragType.Commit)) {
+      this.props.dispatcher.endCherryPickFlow(this.props.repository)
+      this.props.dispatcher.recordCherryPickDragStartedAndCanceled()
```
tidy-dev, line 407: Hmm.. note for later.. this statistics name should be refactored to just `recordDragStartedAndCanceled`

sergiou87 (approved):
Nice refactor! 😍 :shipit:

PR #12240, opened 2021-05-19
Reusable conflicts dialog for multi commit operations - merging, rebasing, cherry-picking, etc..

tidy-dev:
## Description

This is simply the creation of a conflict modal that has all the shared parts of the three existing conflict modals.
To be followed with a PR for refactoring merging, rebasing, and cherry-picking to use it (which may result in tweaking it.. particularly I think the callbacks may need parameter adjustments).

A first step in creating Multi Commit Operation components/models to reduce a lot of redundant code.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/multi-commit-operation/conflicts-dialog.tsx
+++ b/app/src/ui/multi-commit-operation/conflicts-dialog.tsx
@@ -39,12 +39,12 @@ interface IConflictsDialogProps {
     repository: Repository,
     workingDirectory: WorkingDirectoryStatus,
     ourBranch: string,
-    theirBranch: string
-  ) => void
+    theirBranch?: string
+  ) => Promise<void>
   readonly onAbort: (
     workingDirectory: WorkingDirectoryStatus,
     manualResolutions: Map<string, ManualConflictResolution>
-  ) => void
+  ) => Promise<void>
   readonly onDismissed: () => void
   readonly openFileInExternalEditor: (path: string) => void
   readonly openRepositoryInShell: (repository: Repository) => void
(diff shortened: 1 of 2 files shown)


app/src/ui/multi-commit-operation/conflicts-dialog.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   *  Invokes abort callback and dismisses modal
+   */
+  private onAbort = async (e: React.MouseEvent<HTMLButtonElement>) => {
+    e.preventDefault()
+
+    const { workingDirectory, manualResolutions } = this.props
+    this.setState({ isAborting: true })
+    this.props.onAbort(workingDirectory, manualResolutions)
```
sergiou87: Shouldn't this be `awaited`? Otherwise I don't really see the need for the `isAborting` state.

I've checked the current implementation for for cherry-pick and rebase, and their abort methods both are async (as one would expect) but they're not awaited O_o

Am I missing something or you think my assessment makes sense?

tidy-dev: :D - I did see that and was going to investigate further when I refactored to use this one.. cause I was a little confused by that as well.

tidy-dev: The props methods were not async.... so I figured likely a previous oversight and they would need refactored when I get to them.

tidy-dev: Had to fix a build error - went ahead and added the await since it is likely heading that direction.

sergiou87 (approved):
:shipit: ❗

PR #12249, opened 2021-05-20
Update Merging to use new conflicts dialog

tidy-dev:
## Description

Updates merging to use new conflicts dialog. This one probably deserves more scrutiny since it is less of a swap as I had to move logic from the original merge conflicts dialog out into the app state.

I tested the following as I did on cherry-picking and rebase and I have not seen any regression:
- Opens conflicts dialog and all headers, contents, buttons appear as expected.
- Verify the manual resolution dropdown has correct our's and their's branch (target branch first option, source branch second option)
- Verified open in shell, open in external, and open in default work as expected.
- Verified that on dialog dismiss (click backdrop or x), conflicts exist banner appears.
- Verified that if you resolve a conflict and then abort it prompts for confirmation

## Release notes
Notes: no-notes


app/src/ui/app.tsx
```diff
@@ -1731,19 +1738,53 @@ export class App extends React.Component<IAppProps, IAppState> {
...
 
+        const { repository, ourBranch, theirBranch } = popup
+        const submit = __DARWIN__ ? 'Commit Merge' : 'Commit merge'
+        const abort = __DARWIN__ ? 'Abort Merge' : 'Abort merge'
+        const headerTitle = (
+          <span>
+            {`Resolve conflicts before merging `}
+            {theirBranch !== undefined && <strong>{theirBranch}</strong>}
+            {` into `}
+            <strong>{ourBranch}</strong>
```
sergiou87: Is it possible that `theirBranch` is actually `undefined`? If so the final text will have an extra space: `Resolve conflicts before merging  into`

sergiou87 (commented):
Looks good!! 😍  Only one question 😄

app/src/ui/app.tsx
tidy-dev: Good catch!. I updated it and here is what it looks like:
their undefined
image

their not undefined
image

sergiou87 (approved):
Looks good! I tested it and all worked as expected 😄 

I found a regression with drag&drop + switching branches (already mentioned in Slack: if you drag&drop a commit somewhere where the app doesn't do anything and then switch to another branch, the app tries to cherry-pick the commit to that branch), but doesn't seem related to this PR, so… :shipit:

PR #12251, opened 2021-05-20
Fix off-by-one issue expanding whole files

### LINKED ISSUE #12237: "Expand whole file" feature omits or duplicates display of lines in certain cases
### Describe the bug

When viewing file differences in either the Changes or History tab, selecting "Expand whole file" mode sometimes omits and sometimes duplicates display of lines.  Omitted lines can be detected by non-contiguous line numbers.  Duplicated lines contain the same "before change" line number, the same "after change" line number, and the same line content.

### Version & OS

GitHub Desktop v2.8.1
Windows 10 Pro v1909

1. In either the Changes or History tab, select an edited file that has differences.
2. In the differences display, right-click and select "Expand whole file".
3. Carefully examine at least five lines of context around each highlighted change.
4. Look before each highlighted change section for non-contiguous line numbers, and look after each highlighted change for duplicated line numbers (and line content).

### Expected behavior

In "Expand whole file" display mode, all lines should be displayed exactly once each and in line number order... that is, no lines should be omitted, and no lines should be duplicated.

### Actual behavior

In "Expand whole file" display mode, single lines are sometimes omitted prior to a highlighted change, and single lines are sometimes duplicated after a highlighted change

### Screenshots

Omitted line before a highlighted change:  
image

Duplicated line after a highlighted change:  image

### Logs

Not applicable.

### Additional context

None.

sergiou87:
Closes #12237 

## Description

The problem was an off-by-one error (classic) that only happened when a hunk is merged with the adjacent when it's expanded **down**. That doesn't happen when the user manually expands individual hunks, because when the gap between hunks is too short and will result in both hunks to be merged, the hunk below is expanded **up** with its adjacent above 🤯 

However, when a whole file is expanded, the algorithm just expands the first hunk first up and then down as many times as needed.

## Release notes

Notes: [Fixed] Expanding files doesn't show duplicated lines anymore

### DIFF
--- a/app/src/ui/diff/text-diff-expansion.ts
+++ b/app/src/ui/diff/text-diff-expansion.ts
@@ -232,7 +232,7 @@ export function expandTextDiffHunk(
       // which is effectively taking all the undiscovered file contents and
       // would prevent us from expanding down the diff.
       if (isAdjacentDummyHunk === false) {
-        const downLimit = adjacentHunk.header.newStartLine - 1
+        const downLimit = adjacentHunk.header.newStartLine
         to = Math.min(to, downLimit)
         shouldMergeWithAdjacent = to === downLimit
       }
--- a/app/test/unit/text-diff-expansion-test.ts
+++ b/app/test/unit/text-diff-expansion-test.ts
@@ -5,6 +5,7 @@ import { GitProcess } from 'dugite'
 import { DiffParser } from '../../src/lib/diff-parser'
 import {
   expandTextDiffHunk,
+  expandWholeTextDiff,
   getTextDiffWithBottomDummyHunk,
 } from '../../src/ui/diff/text-diff-expansion'
 import { ITextDiff, DiffType } from '../../src/models/diff/diff-data'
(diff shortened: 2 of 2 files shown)


app/src/ui/diff/text-diff-expansion.ts
```diff
@@ -232,7 +232,7 @@ export function expandTextDiffHunk(
       // which is effectively taking all the undiscovered file contents and
       // would prevent us from expanding down the diff.
       if (isAdjacentDummyHunk === false) {
-        const downLimit = adjacentHunk.header.newStartLine - 1
+        const downLimit = adjacentHunk.header.newStartLine
```
sergiou87, line 235: You can see a few lines below, in https://github.com/desktop/desktop/pull/12251/files#diff-d94718d927ac565787ca7ba29877a25a9262e95ec2e56dd9bd5976a97761c4faR244, that we're already subtracting 1 to the `to` value, exactly when it's needed:

```typescript
  const newLines = newContentLines.slice(
    Math.max(from - 1, 0),
    Math.min(to - 1, newContentLines.length)
  )
```

app/test/unit/text-diff-expansion-test.ts
```diff
@@ -184,8 +185,43 @@ describe('text-diff-expansion', () => {
 
     const firstHunk = expandedDiff!.hunks[0]
     expect(firstHunk.header.oldStartLine).toBe(8)
-    expect(firstHunk.header.oldLineCount).toBe(14)
+    expect(firstHunk.header.oldLineCount).toBe(16)
     expect(firstHunk.header.newStartLine).toBe(8)
-    expect(firstHunk.header.newLineCount).toBe(16)
+    expect(firstHunk.header.newLineCount).toBe(18)
```
sergiou87, line 190: This test merging hunks was just wrong 🤦

tidy-dev (approved):
✨ Nice unit test for this bug!

PR #12257, opened 2021-05-21
Add a reusable confirm abort dialog

tidy-dev:
## Description

Adds a reusable confirm abort dialog to use with multi commit operation flow.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/multi-commit-operation/confirm-abort-dialog.tsx
+++ b/app/src/ui/multi-commit-operation/confirm-abort-dialog.tsx
@@ -1,0 +1,80 @@
+import * as React from 'react'
+
+import { Dialog, DialogContent, DialogFooter } from '../dialog'
+import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
+interface IConfirmAbortDialogProps {
+  readonly operation: string
+
+  readonly onReturnToConflicts: () => void
+  readonly onConfirmAbort: () => Promise<void>
+}
+
+interface IConfirmAbortDialogState {
+  readonly isAborting: boolean
+}
+
+export class ConfirmAbortDialog extends React.Component<
+  IConfirmAbortDialogProps,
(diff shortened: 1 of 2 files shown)


app/src/ui/multi-commit-operation/confirm-abort-dialog.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react'
+import { Dialog, DialogContent, DialogFooter } from '../dialog'
+import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
+
+interface IConfirmAbortDialogProps {
+  readonly operation: string
```
sergiou87, line 14: From what I see above, it's expected this string to be capitalized, right? Probably worth to either (a) suggest that in a comment or (b) capitalize it correctly when it's used for macOS. What do you think?

sergiou87 (changes requested):

app/src/ui/multi-commit-operation/confirm-abort-dialog.tsx
tidy-dev: The way I am using it is by passing in this enum value. 
image

And since it's noted to be in sentence case.. I figured that would be enough (which I know you couldn't have seen on this PR).

tidy-dev: >(b) capitalize it correctly when it's used for macOS

because I will just be passing an enum and this is not likely something to be used for random input.. I don't think it is worth it to programmatically fix it.

tidy-dev: Added note in case we use it elsewhere.. maybe merge.. but I am halfway thinking merge could be refactored to use the multi-commit flow...

sergiou87 (approved):

Issue #5874, opened 2018-10-11
Using Undo On Local Commits Can Lead to Data Loss


dsanders11:
## Description

Data loss can happen if you use the 'undo' functionality on a reverted local commit. Noticed this when playing with #5873.

## Version

* GitHub Desktop: 1.4.2

* Operating system: macOS 10.13.6

## Steps to Reproduce

1. Make a change to a file.
2. Make a commit for this change, but do not push, so that it stays local.
3. Go to History tab, right-click on the local commit, 'Revert This Commit'. There are now 2 local commits that haven't been pushed.
4. Go to the Changes tab. Undo both commits.
5. View file on disk, the change made in step 1 should be gone.

### Expected Behavior

The change from step 1 should still be in the file. This is the behavior that happens if you simply skipped step 3 and used undo on the commit from step 2.

### Actual Behavior

When you undo the revert commit, the Changes tab shows the change from step 1 for that file, as if it's not committed by the commit from step 2, and the change is available to commit. I think this is the crux of the issue, as this is when things diverge between GitHub Desktop and the file on disk. If you only undo the revert commit, then view the file on disk, you'll see that it's back to the `HEAD` version, without the change from the commit in step 2, *but* the change is visible under Changes. So the Changes tab at this point is not reflecting the file on disk. Undoing the original commit removes the changes from the Changes tab, so now it reflects what's on disk, and now there's data loss, as the change from step 1 is lost.

outofambit:
@dsanders11 thanks for the detailed write up. this sounds like a bug to me! I've marked this as `investigation-needed` so we can reproduce it on our side and start to look into fixes. thanks again!

steveward:
@dsanders11 I was able to reproduce this behavior -- thanks again for the detailed steps!

dsanders11:
@steveward, ran into a similar issue which I think is probably the same root cause as this one, or at least a similar root cause. However, it's a 'data addition' bug instead of a 'data loss' bug.

1. Get a repo that has a pending commit to fetch
2. Make a commit, don't push, so it's local
3. Fetch the pending commits (should show 1 up, 1 down before clicking)
4. There are now two local commits, the second being the merge commit
5. Undo the merge commit

This leads to any new files in the merged commit being added to your working directory, making a real mess. That's why I called it a data addition bug, it adds stuff that wasn't there before.

enioluwa23:
[...]

PR #12258, opened 2021-05-21
Track stats when users undo a commit

sergiou87:
## Description

This PR starts tracking stats when users undo commits. It directly records when the user clicks on the `Undo` button and the commit is effectively undone.

Is there any other metric that could help us understand better how users use `Undo`? For example:
- Whether or not they undid a commit while having changes in the working directory.
- If there were changes… whether or not they conflicted with the changes in the commit.
- ???

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -41,6 +41,9 @@ export interface IDailyMeasures {
   /** The number of commits created with one or more co-authors. */
   readonly coAuthoredCommits: number
 
+  /** The number of commits undone by the user. */
+  readonly commitsUndone: number
+
   /** The number of times a branch is compared to an arbitrary branch */
   readonly branchComparisons: number
 
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -41,8 +41,11 @@ export interface IDailyMeasures {
   /** The number of commits created with one or more co-authors. */
   readonly coAuthoredCommits: number
 
-  /** The number of commits undone by the user. */
-  readonly commitsUndone: number
+  /** The number of commits undone by the user with a dirty working directory. */
+  readonly commitsUndoneWithChanges: number
+
+  /** The number of commits undone by the user with a clean working directory. */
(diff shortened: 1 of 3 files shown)


tidy-dev (dismissed):
✨

tidy-dev:
> Whether or not they undid a commit while having changes in the working directory.
> If there were changes… whether or not they conflicted with the changes in the commit.

For what is is worth, I do the top one frequently and have not run into a conflicts yet. When wanting to rewrite my history to make grouped logical sense. Example: Commit 1, Commit 2, and Commit 3 have related changes and I committed them as I did them.. but Commit 3 has a modification that goes logically with my Commit 1. 

I only do this if the changes are not massive things I have to look through or are changes that I may want to cherry-pick off and submit as a separate PR.... Also random work flow I have now found with Undo and cherry-pick is in that same scenario if Commit 2 had a ton of unrelated changes and I want to cherry-pick stuff form Commit 1 and Commit 3 but not all of it. I will cherry-pick them to a new branch and then undo them both and just commit back the stuff I want and discard the other stuff.

sergiou87:
> For what is is worth, I do the top one frequently and have not run into a conflicts yet.

FWIW, and this might be my fault for using the word "conflict", but when you undo/reset, there are no conflicts as you could have in a cherry-pick/merge/rebase/etc. `Undo`, since it's just a `git reset` under the hood, will just keep the current state of the repository files and only move back the `HEAD` of the repository to whatever commit you want (in the case of `Undo`, to the previous one).

What I mean with "conflicts" in that context is that your working directory and the commit you're undoing have some files (or even lines inside those files!) in common, which could potentially lead to losing changes: #5874

tidy-dev:
Ah.. Thank you for the context. I haven't run into that either - but that would be annoying.

billygriffin:
This is great, thanks @sergiou87! 

If this is trivial to add, I would be hugely in favor of it:
> Whether or not they undid a commit while having changes in the working directory.

sergiou87:
Done @billygriffin! Please @tidy-dev take another look when you have a chance 🙏

tidy-dev (approved):
✨ Looks Good!

PR #12268, opened 2021-05-22
use app.hide() instead of Menu.sendActionToFirstResponder()

kaatt:
## Description

Uses `app.hide()` instead of `Menu.sendActionToFirstResponder()` which is cleaner

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -81,7 +81,7 @@ export class AppWindow {
       this.window.on('close', e => {
         if (!quitting) {
           e.preventDefault()
-          Menu.sendActionToFirstResponder('hide:')
+          app.hide()
         }
       })
     }


tidy-dev:
Needs further investigation - But FWIW, I did a quick google about the difference on these and found [this thread on getting focusing other apps when switching between them](https://github.com/electron/electron/issues/5495) on MacOS. Tho I think the syntax look much cleaner, need to do some regression testing for this scenario.

tidy-dev (approved):
✨ I did a little bit of regression testing and can't see any change in behavior between app switching/closing. Tested on macOs and windows tho it seems to be the previous complaint was in mac os.

@sergiou87 mind providing your input?

sergiou87 (approved):
I think the change makes sense (or I can't think why it wouldn't) looking at Electron's implementation 🤔 

I didn't notice any regressions either, so :shipit:

PR #12274, opened 2021-05-24
Refactor on drag end

tidy-dev:
## Description

Another drag and drop refactor that is inline with other drag/drop refactors. Since the app is no longer in a cherry-picking state during a drag, there is no need to bubble up the drag end event until we reach the repo state variable and since we are using event emitters for the other  "global" event anyways - it makes since as a emitter and not as a bubble up just to reach a dispatcher.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/lib/draggable.tsx
+++ b/app/src/ui/lib/draggable.tsx
@@ -152,9 +152,9 @@ export class Draggable extends React.Component<IDraggableProps> {
     mouseScroller.clearScrollTimer()
     this.props.onRemoveDragElement()
     if (this.props.onDragEnd !== undefined) {
-      this.props.onDragEnd(this.isLastElemBelowDropTarget())
+      this.props.onDragEnd(this.getLastElemBelowDropTarget())
     }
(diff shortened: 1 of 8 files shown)


app/src/ui/lib/draggable.tsx
```diff
@@ -147,25 +151,25 @@ export class Draggable extends React.Component<IDraggableProps> {
...
+    dragAndDropManager.dragEnded(this.isLastElemBelowDropTarget())
   }
 
   /**
    * Compares the last element that the mouse was over during a drag with the
    * css selectors provided in dropTargetSelectors to determine if the drag
    * ended on target or not.
    */
-  private isLastElemBelowDropTarget = (): boolean => {
+  private isLastElemBelowDropTarget = (): DropTargetSelector | undefined => {
```
sergiou87: Given this function does not return a `boolean` anymore, maybe it's worth renaming it?
```suggestion
  private lastElemBelowDropTarget = (): DropTargetSelector | undefined => {
```

sergiou87 (changes requested):

sergiou87 (approved):
:shipit: ❗

PR #12279, opened 2021-05-25
Bump Electron to v11.3.0

sergiou87:
## Description

Newer versions of Electron 11 [have a bug that empties the lists in Desktop after switching between apps](https://github.com/electron/electron/issues/29261), so until that problem is solved we will stay at v11.3.0, where we couldn't reproduce the issue.

## Release notes

Notes: [Improved] Upgrade to Electron 11.3.0


tidy-dev (approved):
✨

PR #12280, opened 2021-05-25
Adding Multi Commit Operation models

tidy-dev:
## Description

I created a `MultiCommitOperationState` that is a combination of the things tracked by Cherry-picking, Rebasing, and Interactive rebasing (Squashing, Reordering). I also briefly looked over merging to see if merging could be easily refactored to use this too and it looks pretty similar, but right now going to focus on rebasing (so interactive rebase will play nice) and then cherry-picking since it is already following the pattern. We will see if time permits to bring merge in so we have consistency through out these kind of operations.

Some considerations and things that may change as I continue to develop this. 
- I have thought of having an OperationDetails property that would be of type OperationDetail = CherryPickDetails | RebaseDetails | InteractiveRebaseDetails. Thus, allowing a little more flexibility and not such as flat architecture where we have properties for all operations whether they are needed or not. 
 - Examples: 
       -  Cherry-pick and Rebase need to track a sourceBranch and squashing does not.
       -  Squashing/Reordering tracks a targetCommit and lastRetainedRef and cherry-pick and rebase do not.

ConflictState - Currently Merge, Rebase, and CherryPick track different sets of things under the conflictState.. However many of those things are part of the whole operation such as `targetBranch`, `sourceBranch`, `originalBranchTip` (undo sha), etc. I moved to flatten these to `multiCommitOperationState` and only have purely conflict state related things there and I was left with `manualResolutions`, `ours`, `theirs` as `ours` and `theirs` are different depending on your operation and only used in the conflicts modal. I also flattened that in the steps.. but I am considering wrapping those back under the `MultiCommitOperationConflictState`.. The final landing of this is to be determined when I refactor rebase under the multi commit operation (meaning the update conflict state logic will need to be updated in the load status logic) to ensure I have everything needed for all the use cases. 

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -4,7 +4,7 @@ import { IDiff, ImageDiffType } from '../models/diff'
 import { Repository, ILocalRepositoryState } from '../models/repository'
 import { Branch, IAheadBehind } from '../models/branch'
 import { Tip } from '../models/tip'
-import { Commit, CommitOneLine } from '../models/commit'
+import { Commit, CommitOneLine, ICommitContext } from '../models/commit'
 import { CommittedFileChange, WorkingDirectoryStatus } from '../models/status'
 import { CloningRepository } from '../models/cloning-repository'
 import { IMenu } from '../models/app-menu'
@@ -21,6 +21,7 @@ import {
   ICheckoutProgress,
   ICloneProgress,
   ICherryPickProgress,
+  IMultiCommitOperationProgress,
 } from '../models/progress'
 import { Popup } from '../models/popup'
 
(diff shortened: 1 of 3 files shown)


app/src/lib/app-state.ts
```diff
@@ -838,3 +847,112 @@ export function isCherryPickConflictState(
...
+   */
+  readonly commitContext: ICommitContext | null
+
+  /**
+   * For use with interaction rebase operations
+   */
+  readonly lastRetainedCommitRef: string | null
+
+  /**
+   * This is the commit ID of the HEAD of the in-flight operation used to compare
```
sergiou87, line 910: With commit ID you mean SHA, right?

sergiou87 (approved):
💯 :shipit:

app/src/lib/app-state.ts
tidy-dev: Yes. I copied the description from elsewhere; but can definitely update it since that is what it is. :D

PR #12282, opened 2021-05-25
Refactoring Multi Commit logic to use null at the top and details property

tidy-dev:
## Description

Refactor state logic with comments I made on #12280... decided to just go ahead and do this and I will merge into my other work... as this is pretty foundational.
- make repository state have null instead of all the children properties
- add a operationDetails property for operation specific properties
- move our/their under the conflict state

## Release notes
Notes: no-notes


sergiou87 (dismissed):
LGTM! I only wonder if you need to export `ISourceBranchDetails` and `IInteractiveRebaseDetails` since they're only used internally.

sergiou87 (approved):

PR #12285, opened 2021-05-25
Multi commit operation base class

tidy-dev:
## Description

This class is an abstract class that any operation we add can extend so any operation can use this shared logic but can also have logic specific to that operation. 

Note: Dispatcher method for showing conflict dialog banner will be further updated in later PR's. More a less a partial place holder for this class.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -6376,25 +6376,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     targetBranch: Branch,
     commits: ReadonlyArray<Commit>
   ): void {
-    this.repositoryStateCache.updateMultiCommitOperationState(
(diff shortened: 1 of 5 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -6376,25 +6376,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     targetBranch: Branch,
     commits: ReadonlyArray<Commit>
   ): void {
-    this.repositoryStateCache.updateMultiCommitOperationState(
```
tidy-dev, line 6379: Have to refactor to use initializer method - done in a different PR; just removing for now.

sergiou87 (approved):
Looks good! Awesome job! 🙌

PR #12286, opened 2021-05-25
Refactor Rebase progress to generic to allow squash to use rebase progress parser

tidy-dev:
## Description

This makes rebase use the new `IMultiCommitOperationProgress` anywhere it used `IRebaseProgress` or `GitRebaseProgress` which all have the same properties just that `rebasedCount` becomes `position`. Also make the `title` property optional for the `IProgress` since it was not actually used for  Rebase, Cherry-pick, nor the new Multi Commit Operation models.

This makes it so that squash can just use the rebase progress parser in the interactive rebase to process progress.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -5,18 +5,14 @@ import { GitError } from 'dugite'
 import byline from 'byline'
 
 import { Repository } from '../../models/repository'
-import {
-  RebaseInternalState,
-  RebaseProgressOptions,
-  GitRebaseProgress,
-} from '../../models/rebase'
-import { IRebaseProgress } from '../../models/progress'
+import { RebaseInternalState, RebaseProgressOptions } from '../../models/rebase'
+import { IMultiCommitOperationProgress } from '../../models/progress'
 import {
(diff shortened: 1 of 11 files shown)


app/src/lib/git/rebase.ts
```diff
@@ -546,9 +541,10 @@ export async function rebaseInteractive(
   repository: Repository,
   pathOfGeneratedTodo: string,
   lastRetainedCommitRef: string | null,
-  action: string = 'interactive rebase',
+  action: string = 'Interactive rebase',
   gitEditor: string = ':',
-  progressCallback?: (progress: IRebaseProgress) => void
+  progressCallback?: (progress: IMultiCommitOperationProgress) => void,
+  commits?: ReadonlyArray<Commit>
```
tidy-dev, line 547: Decided to pass commits in because in the squash method and likely for reorder we have to obtain this list of commits to generate  the todo list. -> no need to do the git command again if we can just pass in the commits.

sergiou87 (approved):
The refactor looks great, and I tested it and I couldn't find regressions rebasing! Great work 😄 ❤️

PR #12288, opened 2021-05-25
Conflict management methods and rebase refactor for squash

tidy-dev:
Dependent on model change in #12285

## Description
These method updates make so conflicts dialog can work for multi commit operation - some spot particularly so squash can use the continueRebase method.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -288,6 +288,7 @@ import { getTipSha } from '../tip'
 import {
   MultiCommitOperationKind,
   MultiCommitOperationStep,
+  MultiCommitOperationStepKind,
 } from '../../models/multi-commit-operation'
 
 const LastSelectedRepositoryIDKey = 'last-selected-repository-id'
@@ -1956,6 +1957,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     this.updateRebaseFlowConflictsIfFound(repository)
     this.updateCherryPickFlowConflictsIfFound(repository)
+    this.updateMultiCommitOperationConflictsIfFound(repository)
 
     if (this.selectedRepository === repository) {
       this._triggerConflictsFlow(repository)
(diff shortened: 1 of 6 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -2030,9 +2032,50 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    )
+
+    if (isRebaseConflictState(conflictState)) {
+      const { currentTip } = conflictState
+      this.repositoryStateCache.updateMultiCommitOperationState(
+        repository,
+        () => ({ currentTip })
+      )
+    }
+  }
```
sergiou87, line 2071: Could we merge these two in one step?

sergiou87: Doesn't the `isRebaseConflictState` check apply to cherry-picking too? 🤔  Or we only rely on the `sourceBranch` for that?

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -1168,6 +1179,7 @@ export class Dispatcher {
         originalBranchTip
       )
     }
+    return
```
sergiou87: I guess this can be removed, no?
```suggestion
```

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -1157,7 +1168,7 @@ export class Dispatcher {
 
       this.statsStore.recordRebaseSuccessAfterConflicts()
 
-      await this.completeRebase(
+      this.completeRebase(
```
sergiou87: Shouldn't this be returned or awaited?

Checking where this method is called, it seems to end up in the `ShowProgressStep`'s `rebaseAction` property, which returns a promise.

Am I missing something? 🤔

sergiou87 (changes requested):

app/src/lib/stores/app-store.ts
tidy-dev: This is some of more of the stuff I would like to refactor (but haven't delved into how hard it will be.. it is conflict state touches many files). When we do a load status, we update a property on the changes state called the conflictState which currently has 3 types, Merge, CherryPick, and Rebase. In a refactor I would like it just to update the multiCommitOperationState and have generic conflict state for other app purposes. All we need from those typically is the manualResolutions for the dialog box, but in order to continue a rebase we need that currentTip. Thus for now, if it is a rebase conflict I just call it out and copy it. Cherry-pick does not need anything else from the conflict state.

tidy-dev: Also to note, there is a similar method for cherry-pick and rebase specifically as well.. thus these changes "should" not impact the rebase and cherry-pick yet because I am not moving from those implementations yet.

app/src/ui/dispatcher/dispatcher.ts
tidy-dev: Weirdness.. It looks like I somehow cut the return statement and pasted it below..

app/src/lib/stores/app-store.ts
sergiou87: 👍 Thanks for clarifying!

sergiou87 (approved):
✨

app/src/ui/dispatcher/dispatcher.ts
tidy-dev: :D It would have been mad at that because this made all paths return something -> But put the return on the complete and switch and it is happy.

PR #12296, opened 2021-05-26
Multi commit operation popup

tidy-dev:
## Description

Logic to be able to open a multi commit operation popup.

## Release notes

Notes: no-notes


app/src/ui/multi-commit-operation/multi-commit-operation-manager.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
+import { assertNever } from '../../lib/fatal-error'
+import { MultiCommitOperationKind } from '../../models/multi-commit-operation'
+import { Squash } from './squash'
+import { IMultiCommitOperationProps } from './multi-commit-operation'
+
+/** A component for initiating and performing a rebase of the current branch. */
+export class MultiCommitOperationManager extends React.Component<
```
sergiou87: Using the suffix `Manager` for a view/component feels odd… what about `MultiCommitOperationView` or `MultiCommitOperationWrapper` since it wraps subclasses of `MultiCommitOperation` 🤔 

Another option… this is a `MultiCommitOperation` and the existing `MultiCommitOperation` becomes `BaseMultiCommitOperation`?

sergiou87 (commented):
Looks good overall! My only request is around the name of the new Manager class 😅

app/src/ui/multi-commit-operation/multi-commit-operation-manager.tsx
tidy-dev: On another project I worked on, we had the same flow for something similar and called it `Manager` (and another team called it `Process`). 😛  Thus, this feels natural to me. But.. I do like `BaseMultiCommitOperation`. It is clearer what the intent of it is. Refactored to  the second option. 😄

sergiou87 (approved):
Thank you for that renaming!! This feels much better to me 😳 

:shipit:

PR #12298, opened 2021-05-26
Squash with progress/conflict dialogs and force push

tidy-dev:
## Description
YAY! Get to use all that refactoring for something! Squashing has a complete flow through progress and conflicts management.

One unhappy feels: I am using the rebase progress parser since squash is a rebase.. however the rebase is actually aware of all the commits up to the one before the rebase action. Thus as can be seen in the video, even tho I am only squashing 2 commits... the rebase progress parser gets a total of all 5 that it has to reorder/pick back on to the branch during the operation.

Also there is an additional progress updater that is not implemented at this point... but really am unsure if we will ever get a similar feel to rebase/cherry-pick that is doing a one at time action easy to follow where this is a change of 5 into 3 action.

But there are higher priority things to focus on at the moment such as:
- Handling when changes exist
- Handling merge commits
- Disabling squashing in compare view where it doesn't make sense
- Implementing squash merge.
- Metric tracking

After that stuff will circle back to cosmetic feels/behind the scenes refactors. 

### Screenshots
https://user-images.githubusercontent.com/75402236/119686374-44edf180-be14-11eb-9897-ff595dad3156.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -561,7 +561,6 @@ export async function rebaseInteractive(
       return RebaseResult.Error
     }
 
-    // TODO: pass in a rebase action for parser - Rebase, Squash, etc..
     options = configureOptionsForRebase(baseOptions, {
       commits,
       progressCallback,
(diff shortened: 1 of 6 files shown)


sergiou87 (approved):
I love it!! 😍 :shipit:

PR #12326, opened 2021-05-31
Fix drag & drop warnings

sergiou87:
## Description

Some of the drag & drop events we're observing in the `CommitDragElement` and `DragOverlay` were being triggered after the component is unmounted, causing a warning because we tried to change the component's state.
This PR makes sure we unobserve them after the component is unmounted, hopefully being also a good citizen.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨ Was not aware of `Disposable`. Nice!

PR #12331, opened 2021-06-01
Throw error if merge commit exists in squash

tidy-dev:
## Description

Throws an error before starting squash if squash would fail due to merge commits being in the interactive rebase history.

Some context:
An interactive rebase by default will rewrite history keeping the commits merged in and drop the actual merge commit. We felt this would feel strange if a user dragged a commit onto another commit (whether either of those commits had to do with the merge) and the merge commit just disappearing from their history. Also if a user chose to actually squash a merge commit that had say 3 commits in the history associated with it, all 4 (3 + merge) of those commits would disappear as they would be squashed. Thus, we chose to error in this scenario and not allow users to do this as we are assuming it to be an uncommon/edge case that would require more explanation to the user that a drag/drop could not provide.

### Screenshots
image
Notes: no-notes

### DIFF
--- a/app/src/lib/git/log.ts
+++ b/app/src/lib/git/log.ts
@@ -15,6 +15,7 @@ import {
 } from './interpret-trailers'
 import { getCaptures } from '../helpers/regex'
 import { createLogParser } from './git-delimiter-parser'
+import { revRange } from '.'
 
 /**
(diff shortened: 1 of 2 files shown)


sergiou87 (approved):
Love it! :shipit:

PR #12332, opened 2021-06-01
Enable cross auto-update to arm64 on macOS production builds

sergiou87:
## Description

The feature flag will be `true` only on macOS, meaning Windows builds will still not auto-update to arm64 in production.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #12335, opened 2021-06-01
Add stash for local changes on squash

tidy-dev:
## Description

When user has local changes, require them to stash to continue. 

Notes: I had originally thought about doing this on behalf of the user. But changed my mind because it breaks our apps convention  compared to other similar functionality (merge, rebase, cherry-pick). Also, squashing can result in conflicts which when resolved result in conflicts with the stash; therefore, I think it would feel more logical to the user to handle these conflicts as separate actions of "conflicts while squashing" and "conflicts that are preset because of popping stash".

### Screenshots
https://user-images.githubusercontent.com/75402236/120373433-589ec980-c2e6-11eb-87fc-2e61ed34d3a5.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -6390,7 +6390,16 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const {
       branchesState,
       squashState: { undoSha, squashBranchName },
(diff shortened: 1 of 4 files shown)


sergiou87 (approved):
:shipit:

PR #12341, opened 2021-06-02
Show warning before undoing commit if it conflicts changes in working directory

### LINKED ISSUE #4596: Hitting "undo" is a little too easy: a saga of odd merge behavior
## Description

Today a tragedy occurred: I accidentally hit "undo" twice, and one of the commits I inadvertently undid was a master merge. I re-committed right away and pushed up, and imagine my surprise when my PR diff ballooned to include all of the code from master I was supposed to have merged 😭 
 (https://github.com/github/github/pull/89415)

So while the UX could maybe be improved with "maybe we shouldn't undo multiple times, or warn people first", I think there might be a larger issue around undoing a merge of another branch.

## Version

* GitHub Desktop: 1.1.1
* Operating system: Mac OS 10.12.6

## Steps to Reproduce

1. Merge master into an existing branch with commits
2. Make some changes
3. Commit those changes
4. Hit undo (your changes)
5. Hit undo again (your merge)
6. Commit your merge again
7. Push up and BE SAD

### Expected Behavior

I expected the merge to be performed as usual without attaching all that code to my PR.

### Actual Behavior

EXTREME SADNESS

It wasn't actually a merge as I would have expected 😬 - all the commits appeared in the diff.

## Additional Information

PR linked above, you can see it internally 👀 

### Logs

Attached!

[2018-05-03.desktop.production.log](https://github.com/desktop/desktop/files/1972685/2018-05-03.desktop.production.log)

### LINKED ISSUE #5874: Using Undo On Local Commits Can Lead to Data Loss

### LINKED ISSUE #6043: preserve commit messages when undoing several commits

After creating a lot of small commits, I sometimes want to rework the history before pushing the branch. I can do this by rebasing in the command line, but a quick way to do it while avoiding rebase flows would be to just undo all my local commits.

Currently the app will replace the commit summary/description with the details of the previous commit, which is nice if you want to just amend that commit. But if I want to undo multiple commits, the app will replace the contents with the previous commit, so I lose that context and need to poke at the reflog to see what my old history looked like.

One potential solution would be to append the commit summary and/or description to the current commit form. Here's an example of how this might work:

> I have three local commits - A, B and C - that I would like to undo
> C <---- my branch
> B
> A
> X <---- my default branch
>
> 1. I undo the first commit (C) and the commit form becomes:
> 
> *Summary:* C
> *Description:*
> Description related to C
> 
> 2. I undo the second commit (B) and the commit form becomes:
>
> *Summary:* B
> *Description:*
> Description related to B
> 
> C
> Description related to C
>
> 3. I undo the first commit (A) and the commit form becomes:
>
> *Summary:* A
> *Description:*
> Description related to A
>
> B
> Description related to B
> 
> C
> Description related to C

### LINKED ISSUE #9286: Undo does not stage partial files
### Describe the bug

When the uncommit button is used, partially selected files are fully unselected (i.e. they have no selected lines anymore).

### Version & OS

Version: 2.3.1
OS: Windows 10 x64

1. Partially select a file (i.e. select some lines in it to commit)
2. Commit
3. Uncommit

### Expected behavior

The partially selected file has it's staged lines reselected.

### Actual behavior

The entire file is unselected.

sergiou87:
Closes #4596
Closes #9286
Closes #5874
Closes #6043

## Description

This PR checks if there are changes in the working directory that could be overwritten by the changes in the commit to be undone. Basically, the warning is shown only if any of the files changed in the working directory was changed in the commit to undo.

### Screenshots

https://user-images.githubusercontent.com/1083228/120466412-5482b280-c39f-11eb-81f0-253594e7527d.mov

## Release notes

Notes: [Improved] Warn users before undoing commits when changes can be lost in the process

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4139,11 +4139,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
       changesState.workingDirectory.files.length === 0
 
     // Warn the user if there are changes in the working directory
-    if (showConfirmationDialog && !isWorkingDirectoryClean) {
+    if (
+      showConfirmationDialog &&
+      (!isWorkingDirectoryClean || commit.isMergeCommit)
+    ) {
       return this._showPopup({
         type: PopupType.WarnLocalChangesBeforeUndo,
         repository,
         commit,
+        isWorkingDirectoryClean,
       })
     }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4135,29 +4135,17 @@ export class AppStore extends TypedBaseStore<IAppState> {
     const gitStore = this.gitStoreCache.get(repository)
     const repositoryState = this.repositoryStateCache.get(repository)
     const { changesState } = repositoryState
-
-    if (showConfirmationDialog) {
-      // Compare file paths in working directory and in the commit that will be undone
-      const workingDirectoryFiles = changesState.workingDirectory.files.map(
-        f => f.path
-      )
-      const commitFiles = (await getChangedFiles(repository, commit.sha)).map(
-        f => f.path
-      )
-
-      // Warn the user if there is any overlap between commit files and working
-      // directory files
-      if (workingDirectoryFiles.some(f => commitFiles.includes(f))) {
-        return this._showPopup({
-          type: PopupType.WarnLocalChangesBeforeUndo,
-          repository,
-          commit,
-        })
-      }
-    }
-
     const isWorkingDirectoryClean =
       changesState.workingDirectory.files.length === 0
+
+    // Warn the user if there are changes in the working directory
+    if (showConfirmationDialog && !isWorkingDirectoryClean) {
+      return this._showPopup({
+        type: PopupType.WarnLocalChangesBeforeUndo,
+        repository,
+        commit,
+      })
+    }
 
     await gitStore.undoCommit(commit)
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4129,12 +4129,33 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
   public async _undoCommit(
     repository: Repository,
-    commit: Commit
+    commit: Commit,
+    showConfirmationDialog: boolean
   ): Promise<void> {
     const gitStore = this.gitStoreCache.get(repository)
+    const repositoryState = this.repositoryStateCache.get(repository)
+    const { changesState } = repositoryState
 
-    const currentState = this.repositoryStateCache.get(repository)
-    const { changesState } = currentState
+    if (showConfirmationDialog) {
+      // Compare file paths in working directory and in the commit that will be undone
+      const workingDirectoryFiles = changesState.workingDirectory.files.map(
+        f => f.path
+      )
+      const commitFiles = (await getChangedFiles(repository, commit.sha)).map(
+        f => f.path
+      )
+
+      // Warn the user if there is any overlap between commit files and working
+      // directory files
+      if (workingDirectoryFiles.some(f => commitFiles.includes(f))) {
+        return this._showPopup({
+          type: PopupType.WarnLocalChangesBeforeUndo,
+          repository,
+          commit,
+        })
+      }
+    }
+
     const isWorkingDirectoryClean =
       changesState.workingDirectory.files.length === 0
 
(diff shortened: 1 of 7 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -4129,12 +4129,33 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      // Compare file paths in working directory and in the commit that will be undone
+      const workingDirectoryFiles = changesState.workingDirectory.files.map(
+        f => f.path
+      )
+      const commitFiles = (await getChangedFiles(repository, commit.sha)).map(
+        f => f.path
+      )
+
+      // Warn the user if there is any overlap between commit files and working
+      // directory files
```
sergiou87: @billygriffin I kept this behavior that we discussed when the idea was to suggest stashing the changes. Does it still make sense to you? Or should we warn the user even when the current changes don't overlap the commit changes?

tidy-dev (dismissed):
✨  Looks great!

app/src/ui/undo/warn-local-changes-before-undo.tsx
```diff
@@ -0,0 +1,71 @@
...
+        title={title}
+        loading={this.state.isLoading}
+        disabled={this.state.isLoading}
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+      >
+        <DialogContent>
+          <Row>
+            You have changes on this branch. Undoing the commit might lose some
+            of these changes. Do you want to continue anyway?
```
billygriffin:
```suggestion
            You have changes in progress. Undoing the commit might result in some
            of these changes being lost. Do you want to continue anyway?
```

billygriffin (dismissed):
I made a minor wording suggestion in the dialog.

To your question about when we warn, I think there are potentially a few combinations of possibilities.
1. No changes in working directory and undoing normal commit.
2. No changes in working directory and undoing a merge commit.
3. Changes in working directory and undoing a normal commit that doesn't have any files that conflict with the working directory changes.
4. Changes in working directory and undoing a normal commit that does have files that conflict with the working directory changes.
5. Changes in working directory and undoing a merge commit.

I'm wondering if we should just warn anytime there are changes present *and* when you're undoing a merge commit. Warning when there are changes present helps avoid a situation where someone accidentally undoes a commit and then has to figure out which files they have selected to include in the commit. And warning when you're undoing a merge commit seems to help avoid the accidental weird behavior when that happens. I almost wonder if we should even show the Undo button when the last commit is a merge commit.

Thoughts?

sergiou87:
> I'm wondering if we should just warn anytime there are changes present

That definitely feels more consistent, because right now it's hard to understand when the warning is shown and when it isn't.

> when you're undoing a merge commit

That's a good idea! Seems like just running `git cat-file -p <commit SHA>` and looking for more than 1 parent is enough to tell if it's a merge commit 👍

app/src/ui/undo/warn-local-changes-before-undo.tsx
```diff
@@ -0,0 +1,112 @@
...
+      this.props.commit.isMergeCommit &&
+      !this.props.isWorkingDirectoryClean
+    ) {
+      return (
+        <>
+          You have changes in progress. Undoing the merge commit might result in
+          some of these changes being lost.
+          <br />
+          <br />
+          Also, undoing a merge commit will apply the changes from the merge
```
tidy-dev: Nit: maybe we could put the large repeated part of the merge commit warning into a variable at the top so if it ever needs modified we don't have to double check this chunk of text twice.

tidy-dev (changes requested):
Looks great! Just one suggestion on the repeated text.

app/src/ui/undo/warn-local-changes-before-undo.tsx
sergiou87: Fixed!

tidy-dev (approved):
✨ Thanks for the changes!

PR #12344, opened 2021-06-02
Add "Undo Commit" action to the context menu

sergiou87:
## Description

Depends on #12341 

Adds an `Undo Commit` in the context menu of the commit list, for individual commits.

### Screenshots

https://user-images.githubusercontent.com/1083228/120476668-230fe400-c3ab-11eb-8e00-f8db8f3124d5.mov

## Release notes

Notes: [Improved] Allow to undo the latest commit from the context menu in the History tab

### DIFF
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -252,7 +252,7 @@ export class CommitListItem extends React.PureComponent<
 
     if (this.props.canBeUndone) {
       items.push({
-        label: __DARWIN__ ? 'Undo Commit' : 'Undo commit',
+        label: __DARWIN__ ? 'Undo Commit…' : 'Undo commit…',
         action: () => {
           if (this.props.onUndoCommit) {
             this.props.onUndoCommit(this.props.commit)
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -26,6 +26,8 @@ interface ICommitProps {
   readonly selectedCommits: ReadonlyArray<Commit>
   readonly emoji: Map<string, string>
   readonly isLocal: boolean
+  readonly canBeUndone: boolean
+  readonly onUndoCommit?: (commit: Commit) => void
   readonly onRevertCommit?: (commit: Commit) => void
   readonly onViewCommitOnGitHub?: (sha: string) => void
   readonly onCreateBranch?: (commit: CommitOneLine) => void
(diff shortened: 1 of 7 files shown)


app/src/ui/history/commit-list-item.tsx
```diff
@@ -243,19 +245,31 @@ export class CommitListItem extends React.PureComponent<
...
-    const items: IMenuItem[] = [
-      {
-        label: __DARWIN__
-          ? 'Revert Changes in Commit'
-          : 'Revert changes in commit',
+    const items: IMenuItem[] = []
+
+    if (this.props.canBeUndone) {
+      items.push({
+        label: __DARWIN__ ? 'Undo Commit' : 'Undo commit',
```
tidy-dev: Should `Undo Commit` have the `...`?

tidy-dev (changes requested):
I am on your #12353 amend pr branch, which I realized was based off this one so I went ahead and tested this out there instead of switching to this branch.. But the undo does this weird flicker thing.  I was expecting it to work like Amend and bounce me back to the changes view.. but looking at the code, is that not the intention?  

If it is not supposed to bounce to the changes tab, then I still think we need to resolve the flicker.

https://user-images.githubusercontent.com/75402236/120651787-22c82500-c44d-11eb-82b7-45f476f584e2.mov

sergiou87:
@tidy-dev Fixed! It seems the selection was cleared twice: once in the `_undoCommit` method, and then again in the `_refreshRepository` method, causing some weird UI glitches to happen 😄 

https://user-images.githubusercontent.com/1083228/120655847-5c1b8780-c483-11eb-9e86-bdd5138e7eb8.mov

tidy-dev (commented):
Awesome! Thank you for that fix. It still feels strange to me tho that amend takes you to the changes tab and undo doesn't. Wouldn't a user more than likely need to go to the changes tab after undoing to fix what they are undoing? 

 (or do we think that is assuming the users intention too much.. I get that too)

sergiou87:
I don't have a strong opinion in this case, since you can also undo from the changes tab. With Amend, I found it difficult to understand the change in the app state if we just left the user in the History tab.

I could buy the "consistency" argument though 😄  @billygriffin any strong opinion about it?

billygriffin:
I agree with @tidy-dev that Undo should take you to the Changes tab as well, especially since it populates the changes with the contents. Or at the very least, I'd like to see it and make sure it doesn't feel too unreasonable.

sergiou87:
Fixed! Thanks for the feedback! 🙌 ❤️ 

https://user-images.githubusercontent.com/1083228/120684409-efae8180-c49e-11eb-8c3a-3a313baf45d9.mov

tidy-dev:
It maybe feels a little more jarring than amend since it populates the diff after switching... but I still think it is what I expected. Also I think if we do move to the new screen `Undo Commit` needs to be `Undo Commit...` to indicate you are going somewhere.

sergiou87:
Good point @tidy-dev!! Done 😄

tidy-dev (approved):

PR #12345, opened 2021-06-02
Dropdown select button

tidy-dev:
Waiting on feedback from @ampinsk on clipping issue if app window height is too short to show all the options.

## Description

This is the component for the dropdown select button. Demo shows it being implemented in merge dialog but this pull request does not implement it. 

### Screenshots
Latest result after design input: 

https://user-images.githubusercontent.com/75402236/120541843-658bed80-c3b8-11eb-92c4-1677328f4d10.mov

On windows.. just to show verification that it looks the same.. sometimes windows view tweaks things alittle.
image

Updated for dynamic popover:
https://user-images.githubusercontent.com/75402236/120532797-117c0b80-c3ae-11eb-8ad1-9efb7083e8bd.mov

Outdated scrolling:
https://user-images.githubusercontent.com/75402236/120485832-791c6180-c382-11eb-90ea-763e2d41bab6.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/dropdown-select-button.tsx
+++ b/app/src/ui/dropdown-select-button.tsx
@@ -14,7 +14,8 @@ export interface IDropdownSelectButtonOption {
 }
 
 interface IDropdownSelectButtonProps {
-  readonly options: IDropdownSelectButtonOption[]
+  /** The selection button options */
+  readonly options: ReadonlyArray<IDropdownSelectButtonOption>
 
   /** The selection option value */
   readonly selectedValue?: string
--- a/app/src/ui/dropdown-select-button.tsx
+++ b/app/src/ui/dropdown-select-button.tsx
@@ -96,7 +96,7 @@ export class DropdownSelectButton extends React.Component<
 
   private onSelectionChange = (selectedOption: IDropdownSelectButtonOption) => {
     return (_event: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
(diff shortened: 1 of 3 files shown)


app/src/ui/dropdown-select-button.tsx
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  private renderSplitButtonOptions() {
+    if (!this.state.showButtonOptions) {
+      return
+    }
+
+    const { options } = this.props
+    const { optionsPositionBottom: bottom } = this.state
+    const openClass = bottom !== undefined ? 'open-top' : 'open-bottom'
```
sergiou87, line 141: I think `optionsPositionBottom` and `bottom` are a bit misleading in this case, because I was expecting this condition to be the other way round 😂  
```suggestion
    const openClass = bottom !== undefined ? 'open-bottom' : 'open-top'
```

sergiou87 (approved):
Looks great, awesome job!! 😍

app/src/ui/dropdown-select-button.tsx
tidy-dev: Yeah. Definitely confusing.. but the definition of a bottom position style means that we are setting the 'bottom' style property of the options so that it will open to the top. I may tweak this logic (later) to store a boolean and calculate the bottom at the time usage to make reading it clearer.

Issue #12299, opened 2021-05-26
Reorder commits in commit history


billygriffin:
### Describe the feature or problem you’d like to solve

When we've gotten feature requests for interactive rebase, one of the things people say they're using it for is to reorder commits. So let's enable that!

### Proposed solution

Now that we have a drag-drop paradigm from cherry-picking, we can use that to let people drag commits to wherever they would like in the commit history.

Note: This should play well with squashing since we'll be enabling that with drag-drop functionality as well.

Kurt-von-Laven:
This would be great! It's one of a small handful of lingering issues keeping our team from adopting the tool, because we rebase a lot. Right now we rely on GitLens for this graphical interactive rebases, but I believe it is buggy.

PR #12358, opened 2021-06-03
Initial drag&drop support to insert items into a list

sergiou87:
Part of #12299 

## Description

This PR adds an overlay used to enable dropping items between rows of a list, which we will leverage in an upcoming PR to reorder commits in the `History` tab. In this PR, however, it's not possible to do so yet.

A few architecture details:

- When a list has this drag&drop "reordering" / "insertion" capability enabled, it will wrap every row with a `ListItemInsertionOverlay` component.
- This new component has:
	- Two drop areas/targets where items can be dropped.
	- Indicators for those two areas to give feedback to the user when they can drop the items.

It still requires some UX love for those indicators on both macOS and specially Windows, but that's for a future PR.

### Screenshots

This video shows the feature but it can't be seen in this branch yet.

https://user-images.githubusercontent.com/1083228/120681432-a9a3ee80-c49b-11eb-93f2-db4b281a3a48.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/models/drag-drop.ts
+++ b/app/src/models/drag-drop.ts
@@ -29,12 +29,14 @@ export type DragElement = {
 export enum DropTargetType {
   Branch,
   Commit,
+  ListInsertionPoint,
 }
 
 export enum DropTargetSelector {
   Branch = '.branches-list-item',
   PullRequest = '.pull-request-item',
   Commit = '.commit',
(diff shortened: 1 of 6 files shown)


tidy-dev (approved):
✨  Awesome! I was curious how you made that happen!

PR #12362, opened 2021-06-04
Refactor merge to multi commit operation

tidy-dev:
## Description

Refactors merge flow to use multi commit operation in preparation for squash-merge/rebase using same choose branch dialog.

### Screenshots

Things of note: 
- The choose branch dialog, merge conflict dialog, and abort merge warning are all now pointing to the multi-commit-operation reusable components. At this point that this has resulted in some slight changes from original merge in terms of copy. 
- Choose branch dialog:
    -  The button is the select dropdown button so it no longer says "Merge `branch-a` into `branch-b`"
    -  The select dropdown currently only has one option, but will eventually have squash and merge and rebase.
- Merge conflict dialog
   - The header no longer says resolve before merging `branch-a` into `branch-b`... This is do to using a reusable component.. however if we like the "resolve before "some operation about some branches" we can easily expand this dialog to have a title input for us. (future pr/discussion as I know when talking with Amanda before there was pain points with showing the branch names everywhere especially for long branch names)
- Abort warning when already resolved some conflicts now has much more generic copy that communicates the same thing that rebase, cherry-pick and merge did. 

https://user-images.githubusercontent.com/75402236/120831919-0fd95180-c52e-11eb-8bcb-6f1de8d6afc2.mov

## Release notes

Notes: no-notes


app/src/lib/stores/app-store.ts
```diff
@@ -2154,18 +2160,37 @@ export class AppStore extends TypedBaseStore<IAppState> {
   /** starts the conflict resolution flow, if appropriate */
   private async showMergeConflictsDialog(
     repository: Repository,
-    conflictState: MergeConflictState
+    conflictState: MergeConflictState,
+    multiCommitOperationState: IMultiCommitOperationState | null,
+    tip: Tip
   ) {
+    if (multiCommitOperationState === null && tip.kind === TipState.Valid) {
```
sergiou87, line 2167: I'd say this `if` deserves a comment explaining why merge is "special" in this regard and it needs to be "initialized" here.

sergiou87 (changes requested):
Nice cleanup!! Only have one request 😳

app/src/lib/stores/app-store.ts
tidy-dev: It isn't special actually it is just the only one of merge, cherry-picking, rebase we have implemented so far. Any of them once using the multi commit operation will require initialization here.

sergiou87: Cool! I thought this was merge-specific. ✅

sergiou87 (commented):
Code LGTM! I tested it and all looks good, with one small difference: if the dialog is open, when you leave the app in background and switch back to it, the dialog is closed.

sergiou87 (dismissed):
Discussed that regression with @tidy-dev in Slack and decided to tackle it in a different PR :shipit:

sergiou87 (approved):
:shipit:

PR #12380, opened 2021-06-07
Modifying merge conflict detection/handling to support --squash

tidy-dev:
## Description

There is a pretty fundamental change to getting the conflict state that is saying that there must be conflicted files or it just returns null. Before, it only checked against there being a merge head, rebase git files, or cherry picking head and it would return a conflict state whether there was conflicts or not (because if those are present it means the operation has stopped which occurs during conflicts). Fairly sure you should have always have conflicted files to have a conflicted state, and was required to add check for it to detect conflict state for merge --squash because if conflicts occur after a merge --squash there is no longer a merge head and only thing available to indicate there was a merge is the SQUASH_MSG file. Thus, I use that to set a Merge conflict state. However, if you were to discard the changes (with desktop or not) that are in conflict, SQUASH_MSG will remain and I so I need to check for conflicted files to actually put it in conflict state.  

TLDR; This change requires regression testing conflicts during merging, rebasing, and cherry-picks and I did not see a change in behavior after adding the additional constraint (which should be true for them) to the logic setting the conflict state.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2196,13 +2196,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return
     }
 
-    const possibleTheirsBranches = await getBranchesPointedAt(
+    let possibleTheirsBranches = await getBranchesPointedAt(
       repository,
       'MERGE_HEAD'
     )
     // null means we encountered an error
     if (possibleTheirsBranches === null) {
-      return
+      possibleTheirsBranches = []
+      // return
     }
     const theirBranch =
       possibleTheirsBranches.length === 1
--- a/app/src/lib/git/merge.ts
+++ b/app/src/lib/git/merge.ts
@@ -147,5 +147,18 @@ export async function abortMerge(repository: Repository): Promise<void> {
  */
 export async function isMergeHeadSet(repository: Repository): Promise<boolean> {
   const path = Path.join(repository.path, '.git', 'MERGE_HEAD')
-  return FSE.pathExists(path)
+  return await FSE.pathExists(path)
+}
+
+/**
+ * Check the `.git/SQUASH_MSG` file exists in a repository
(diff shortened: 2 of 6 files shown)


Ahmadi53:
در تاریخ دوشنبه ۷ ژوئن ۲۰۲۱،‏ ۴:۴۷ tidy-dev ***@***.***> نوشت:

app/src/lib/stores/app-store.ts
```diff
@@ -2196,13 +2196,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-    const possibleTheirsBranches = await getBranchesPointedAt(
+    let possibleTheirsBranches = await getBranchesPointedAt(
       repository,
       'MERGE_HEAD'
     )
     // null means we encountered an error
     if (possibleTheirsBranches === null) {
-      return
+      possibleTheirsBranches = []
+      // return
```
sergiou87, line 2206: I guess this `return` can be removed?
```suggestion
```

sergiou87 (commented):
Haven't tested it _yet_. The change makes sense to me, but I'm wondering if this will cause any performance issue with repos with many changes that are not conflicts. Not sure how crazy things can get in someone's repository 😂 

Maybe there is a way to get this info before hand in the `git status` call? As we parse the output, get that info and keep it stored 🤔

app/src/lib/stores/app-store.ts
tidy-dev: ... meant to put a future PR mention here. Right now for squash merge I don't have a way to detect the our branch for squash merge (no merge head) so I need to add a little bit here to check if multi commit state is present and get it from there if so... TLDR: going to leave for now as it was reminder to myself to refactor the logic here.

sergiou87 (approved):
LGTM! I tested it and all didn't notice any regressions with rebase, merge, squash and cherry-pick 🤞 

:shipit:

PR #12357, opened 2021-06-03
Initial support for the git-based reorder implementation

sergiou87:
Part of #12299 

## Description

This PR just implementes a `reorder` function with some tests.

### Screenshots

image

Pretty, ain't it?

## Release notes

Notes: no-notes

PR #12384, opened 2021-06-07
Enable flow to reorder commits

### LINKED ISSUE #12299: Reorder commits in commit history

sergiou87:
Closes #12299 

## Description

This PR enables the flow to reorder commits, handling conflicts and undo. I refactored a few things around undoing squash commits, just to be able to reuse that code for reordering.

However, there are other functions that seemed less easy to reuse, so for now I duplicated them and we'll look for opportunities to refactor later (I'm open to suggestions of course! 😄 )

### Screenshots

https://user-images.githubusercontent.com/1083228/120996944-13b3e080-c787-11eb-9f6e-03f13a503e31.mov

## Release notes

Notes: [New] Reorder commits from the History tab

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -471,8 +471,8 @@ export interface IRepositoryState {
   /** State associated with a cherry pick being performed */
   readonly cherryPickState: ICherryPickState
 
-  /** State associated with a squash operation */
-  readonly squashState: ISquashState
+  /** Undo state associated with a multi commit operation operation */
+  readonly multiCommitOperationUndoState: IMultiCommitOperationUndoState
 
   /** State associated with a multi commit operation such as rebase,
    * cherry-pick, squash, reorder... */
@@ -808,21 +808,24 @@ export interface ICherryPickState {
   readonly branchCreated: boolean
 }
 
-/** State associated with a cherry pick being performed on a repository */
-export interface ISquashState {
+/**
+ * Undo state associated with a multi commit operation being performed on a
+ * repository.
+ */
+export interface IMultiCommitOperationUndoState {
   /**
-   * The sha of the tip before squash was initiated.
+   * The sha of the tip before operation was initiated.
    *
-   * This will be set to null if no squash has been initiated.
+   * This will be set to null if no multi commit operation has been initiated.
    */
   readonly undoSha: string | null
 
   /**
-   * The name of the branch the squash operation applied to
+   * The name of the branch the operation applied to
    *
-   * This will be set to null if no squash has been initiated.
+   * This will be set to null if no multi-commit operation has been initiated.
    */
-  readonly squashBranchName: string | null
+  readonly branchName: string | null
 }
 
 /**
--- a/app/src/lib/git/reorder.ts
+++ b/app/src/lib/git/reorder.ts
@@ -86,7 +86,7 @@ export async function reorder(
       // appeared on the log to reduce potential conflicts.
       if (beforeCommit !== null && commit.sha === beforeCommit.sha) {
         foundBaseCommitInLog = true
-        toReplayBeforeBaseCommit.unshift(commit)
+        toReplayAfterReorder.push(commit)
 
         for (let j = 0; j < toReplayBeforeBaseCommit.length; j++) {
           await FSE.appendFile(
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -291,6 +291,7 @@ import {
   MultiCommitOperationStep,
   MultiCommitOperationStepKind,
 } from '../../models/multi-commit-operation'
+import { reorder } from '../git/reorder'
 
 const LastSelectedRepositoryIDKey = 'last-selected-repository-id'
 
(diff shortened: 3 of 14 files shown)


app/src/lib/app-state.ts
```diff
@@ -808,21 +808,24 @@ export interface ICherryPickState {
   readonly branchCreated: boolean
 }
 
-/** State associated with a cherry pick being performed on a repository */
-export interface ISquashState {
+/**
+ * Undo state associated with a multi commit operation being performed on a
+ * repository.
+ */
+export interface IMultiCommitOperationUndoState {
```
sergiou87, line 815: Not super happy with the name since it's only used for squash and reordering, but maybe it can be used for others? I haven't spent much time looking into that yet 😳

app/src/lib/git/reorder.ts
```diff
@@ -86,7 +86,7 @@ export async function reorder(
       // appeared on the log to reduce potential conflicts.
       if (beforeCommit !== null && commit.sha === beforeCommit.sha) {
         foundBaseCommitInLog = true
-        toReplayBeforeBaseCommit.unshift(commit)
+        toReplayAfterReorder.push(commit)
```
sergiou87, line 89: I made a mistake here with my logic in #12357 which was putting the "base" commit before the commits you're moving. I adapted the tests in consequence to reflect the actual expected behavior (moved commits should be placed _before_ the base commit).

app/src/lib/stores/app-store.ts
```diff
@@ -6482,8 +6526,27 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        break
+      case MultiCommitOperationKind.Rebase:
+      case MultiCommitOperationKind.CherryPick:
+      case MultiCommitOperationKind.Merge:
+        throw new Error(
+          `Unexpected multi commit operation kind to undo ${kind}`
+        )
+      default:
+        assertNever(kind, `Unsupported multi operation kind to undo ${kind}`)
+    }
```
sergiou87, line 6546: Again, maybe we can use this for rebase/cherry-picking/merge? Haven't looked into it yet! 😳

app/src/models/banner.ts
```diff
@@ -92,6 +94,18 @@ export type Banner =
...
+      /** number of commits reordered */
+      readonly count: number
+      /** callback to run when user clicks undo link in banner */
+      readonly onUndo: () => void
+    }
+  | {
+      readonly type: BannerType.ReorderUndone
+      /** number of commits reordered */
+      readonly commitsCount: number
+    }
```
sergiou87, line 108: Maybe we could have just one banner for all the multi commit operations too? 🤔

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -3187,14 +3246,19 @@ export class Dispatcher {
     )
   }
 
-  private logHowToRevertSquash(tip: IValidBranch): void {
+  private logHowToRevertMultiCommitOperation(
```
sergiou87, line 3257: I tried to reuse this one for rebase but I saw some differences, so I stopped there since I didn't want to break anything (else) 😂

app/src/ui/history/commit-list.tsx
```diff
@@ -384,8 +384,21 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
+      this.props.commitSHAs.length === 0 ||
+      (baseCommitIndex !== null &&
+        baseCommitIndex > this.props.commitSHAs.length)
+    ) {
+      return
+    }
+
     const baseCommitSHA =
-      row < this.props.commitSHAs.length ? this.props.commitSHAs[row] : null
+      baseCommitIndex === null ? null : this.props.commitSHAs[baseCommitIndex]
```
sergiou87, line 401: This is also part of the mistake in the logic I mentioned earlier. This logic was wrong, but since the `reorder.ts` logic was wrong too, everything seemed to work… until you try to put a commit at the beginning or the end of the history 😅

app/src/lib/app-state.ts
tidy-dev: Yeah. It should be able to be used for cherry-picking when it is refactored to multi-commit-operation. I like it!

app/src/lib/app-state.ts
```diff
@@ -808,21 +808,24 @@ export interface ICherryPickState {
...
 
   /**
-   * The name of the branch the squash operation applied to
+   * The name of the branch the operation applied to
    *
-   * This will be set to null if no squash has been initiated.
+   * This will be set to null if no multi-commit operation has been initiated.
    */
-  readonly squashBranchName: string | null
+  readonly branchName: string | null
```
tidy-dev, line 828: The `null` discussion again.. since we refactored `MultiCommitOperation` to be | null should we do that with `multiCommitOperationUndoState`

app/src/lib/stores/app-store.ts
tidy-dev: Yes for cherry-picking.. rebase and merge currently don't support undo.. but maybe they could...

app/src/models/banner.ts
tidy-dev: Yeah. I think I was heading that way.. but didn't get there on the refactor. :P  I am all for it.  (Maybe a future PR tho)

tidy-dev (approved):
✨  Awesome! - Love how you changed the squashing stuff to multi commit (where my mind was going with it, but wasn't sure yet if those things were reusable) Tested it out and works great!

I had one commit about changing the `multiCommitOperationUndoState` to a  `| null` like we did for `multiCommitOperation` ... but this would involve creating an initializer function (and maybe a clear state function) which since this state only has 2 properties may be this is just as "messy" as the null checks. Thus, I good either way with that.

sergiou87:
I thought about it but didn't actually do it for some reason 🤔  I'll raise another PR with it if you don't mind 😄

PR #12364, opened 2021-06-04
Drag & drop commits inside of the History tab

sergiou87:
Part of #12299

(Blocked by #12358)

## Description

This PR wires the work from #12358 with the code needed to reorder commits, leaving a placeholder for the endpoint in the `Dispatcher`.

### Screenshots

https://user-images.githubusercontent.com/1083228/120681432-a9a3ee80-c49b-11eb-93f2-db4b281a3a48.mov

## Release notes

Notes: no-notes

PR #12385, opened 2021-06-07
Fix for dragging commits clicking from anywhere in them

sergiou87:
## Description

After my work in #12364 I noticed that you cannot start dragging commits from anywhere in the list items anymore, there are some areas that don't seem clickable.

The reason was the new "insertion overlay" was being rendered when the user is not dragging, and the areas where you can drop the commits to reorder them have pointer events disabled.

This PR observes when the app starts dragging and starts rendering those areas only when needed.

### Screenshots

https://user-images.githubusercontent.com/1083228/121003260-a0619d00-c78d-11eb-9b36-76b77b1b92cf.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/drag-and-drop-manager.ts
+++ b/app/src/lib/drag-and-drop-manager.ts
@@ -40,6 +40,10 @@ export class DragAndDropManager {
     return this.emitter.on('leave-drop-target', fn)
   }
 
+  public onDragStarted(fn: () => void): Disposable {
+    return this.emitter.on('drag-started', fn)
+  }
+
   public onDragEnded(
     fn: (dropTargetSelector: DropTargetSelector | undefined) => void
   ): Disposable {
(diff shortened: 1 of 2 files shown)


app/src/lib/drag-and-drop-manager.ts
```diff
@@ -48,11 +52,12 @@ export class DragAndDropManager {
 
   public dragStarted(): void {
     this._isDragInProgress = true
+    this.emitter.emit('drag-started', {})
   }
 
   public dragEnded(dropTargetSelector: DropTargetSelector | undefined) {
-    this.emitter.emit('drag-ended', dropTargetSelector)
     this._isDragInProgress = false
+    this.emitter.emit('drag-ended', dropTargetSelector)
```
sergiou87, line 60: Hopefully this change in the order (which I think makes more sense: notify drag has ended when internally is marked as such) doesn't affect anything. I've tested cherry-picking and squash and all seemed to work as expected.

tidy-dev: Makes sense to me

tidy-dev (approved):
✨  Nice UX catch!

PR #12386, opened 2021-06-07
Don't allow reordering commits when merge commits are involved

sergiou87:
## Description

Just like what we do with squash since #12331, we'll throw an error if we try to reorder commits when merge commits are in the history.

### Screenshots

https://user-images.githubusercontent.com/1083228/121008915-cdb14980-c793-11eb-9702-aa13c9d9c6a8.mov

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #12393, opened 2021-06-07
Reset to commit

sergiou87:
Closes TBD

## Description

Initial work to reset HEAD to a previous commit, up to the first non-local commit. When you use it, it resets the HEAD of the repo to the commit you selected, keeping all the changes in the working dir and in the commits in between. As a consequence, undoing the last commit is equivalent to resetting to the second to last commit, with the only difference that undoing the commit also keeps the commit summary and description.
You can see how you can reset up to the first non-local commit (which would, effectively, “undo” all local commits at once).

I'll add metrics in a new PR.

### Screenshots

https://user-images.githubusercontent.com/1083228/121159693-b170e380-c84b-11eb-9c9a-f6f69897649a.mov

## Release notes

Notes: [New] Reset to a previous commit

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -117,3 +117,8 @@ export function enableAmendingCommits(): boolean {
 export function enableCommitReordering(): boolean {
   return enableDevelopmentFeatures()
 }
+
+/** Should we allow resetting to a previous commit? */
+export function enableResetToCommit(): boolean {
+  return enableDevelopmentFeatures()
+}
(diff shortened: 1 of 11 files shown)


tidy-dev (changes requested):
✨  Code looks good and tried it out. Awesome work!

Should we put this behind a feature flag? 

Amanda suggested putting a separator in to make the dropdown easier to scan.  The top four seem like a logical grouping to me. They are kind of like undos/modifies. 

image

sergiou87:
Sounds good! I still see it a bit excessive, but it's definitely an improvement 😂 

Oh and I'll hide it behind a feature flag, I totally forgot about that since I started with just a PoC 🤦

sergiou87:
Done! @tidy-dev

tidy-dev (approved):
✨  Thanks for those changes! The separator does help.

Dan88271988:
> ✨ Cod

PR #12394, opened 2021-06-07
Drop empty root commit after an interactive rebase

sergiou87:
## Description

Something I noticed while working on reordering is that if you move a commit to the beginning of the branch, and that operation causes conflicts, git creates a completely empty commit. I haven’t seen a way to prevent this or to set a message for it… and right now the app doesn’t render it nicely :cry: The app is expecting that commits always have a summary/description.

This branch cleans that commit after the reorder/squash operation is completed.

However…
1. The empty commit is still visible while the user deals with the conflicts.
2. If, for whatever reason, we wrongly decide the root commit is empty and we drop it… it might cause user data loss.

So, personally, I prefer to make sure we render those empty commits properly and avoid destructive operations as much as possible.

### Screenshots

https://user-images.githubusercontent.com/1083228/121057854-02d39100-c7c0-11eb-8679-333effd57e3b.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -28,6 +28,8 @@ import { stageFiles } from './update-index'
 import { getStatus } from './status'
 import { getCommitsBetweenCommits } from './rev-list'
 import { Branch } from '../../models/branch'
+import { getCommits } from '.'
+import { getTempFilePath } from '../file-system'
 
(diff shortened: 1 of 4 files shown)


sergiou87:
We decided not to do this, and instead just make those commits look as nice as we reasonably can 😄

PR #12403, opened 2021-06-08
Add force-push warning to reorder and squash operations

sergiou87:
## Description

This PR adds the new dialog to warn the user when a squash or reorder operation will require force-pushing the current branch.

The way it detects this situation is by looking for commits in common between the remote and the "last retained commit ref" (meaning, the oldest commit involved in the rebase operation). If there are commits in common, it means the user will require a force-push after the interactive rebase.

### Screenshots

https://user-images.githubusercontent.com/1083228/121173465-2565b880-c859-11eb-9320-d54d899e822d.mov

## Release notes

Notes: no-notes


tidy-dev (approved):
✨  Looks and works great! Thank you for implementing the squash part!

PR #12407, opened 2021-06-08
Downgrade Electron to 11.1.1

sergiou87:
Since I was able to reproduce https://github.com/electron/electron/issues/29261 on v11.3.0, we decided that the safest thing to do is to go back to v11.1.1 for the upcoming release.


tidy-dev (approved):
✨ : sadness.. but yeah.

PR #12408, opened 2021-06-08
Squash and squash merge metrics

tidy-dev:
## Description

This PR starts tracking for metrics around squashing commits and squash merging. Squashing follows similar pattern to cherry-picking, rebasing, and reordering.  Squash-merging I did not recreate all the merge tracking metrics as they looked to be monitoring the merge conflicts dialog process pretty tightly.  That information I do not think is needed at a squash merging granular level.

Squashing Commits:
- squashConflictsEncounteredCount
- squashMultipleCommitsInvokedCount
- squashSuccessfulCount
- squashSuccessfulWithConflictsCount
- squashViaContextMenuInvokedCount:
- squashViaDragAndDropInvokedCount
- squashUndoneCount

Merge Squashing:
- squashMergeInvokedCount
- squashMergeIntoCurrentBranchMenuCount
   - useful to see if users will invoke squash merge via branch menu or via merge dialog select dropdown button
- squashMergeSuccessfulCount
- squashMergeSuccessfulWithConflictsCount

Also refactors 
`cherryPickDragStartedAndCanceledCount` to  `dragStartedAndCanceledCount`

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -161,6 +161,16 @@ const DefaultDailyMeasures: IDailyMeasures = {
   reorderSuccessfulWithConflictsCount: 0,
   reorderMultipleCommitsCount: 0,
   reorderUndoneCount: 0,
+  squashConflictsEncounteredCount: 0,
+  squashMultipleCommitsInvokedCount: 0,
+  squashSuccessfulCount: 0,
+  squashSuccessfulWithConflictsCount: 0,
+  squashViaContextMenuInvokedCount: 0,
+  squashViaDragAndDropInvokedCount: 0,
+  squashUndoneCount: 0,
+  squashMergeIntoCurrentBranchMenuCount: 0,
+  squashMergeSuccessfulWithConflictsCount: 0,
+  squashMergeSuccessfulCount: 0,
 }
 
 interface IOnboardingStats {
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -172,6 +172,7 @@ const DefaultDailyMeasures: IDailyMeasures = {
   squashMergeIntoCurrentBranchMenuCount: 0,
   squashMergeSuccessfulWithConflictsCount: 0,
   squashMergeSuccessfulCount: 0,
+  squashMergeInvokedCount: 0,
 }
 
 interface IOnboardingStats {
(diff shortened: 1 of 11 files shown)


billygriffin (dismissed):
I'm good with these!

app/src/lib/stats/stats-store.ts
```diff
@@ -1525,12 +1543,149 @@ export class StatsStore implements IStatsStore {
...
+      case MultiCommitOperationKind.Squash:
+        return this.recordSquashConflictsEncountered()
+      case MultiCommitOperationKind.Reorder:
+        return this.recordReorderConflictsEncountered()
+      case MultiCommitOperationKind.CherryPick:
+      case MultiCommitOperationKind.Rebase:
+      case MultiCommitOperationKind.Merge:
+        throw new Error(
+          `[recordOperationConflictsEncounteredCount] - Operation not supported: ${kind}`
+        )
```
sergiou87: Isn't throwing an error too much for this? 😳 

I think I'd just log them. We don't want to throw an error and crash the app or bother the user if, by our mistake, the app reaches these stats-related lines.

app/src/lib/stats/stats-store.ts
```diff
@@ -1525,12 +1543,149 @@ export class StatsStore implements IStatsStore {
...
+  }
+
+  public recordOperationConflictsEncounteredCount(
+    kind: MultiCommitOperationKind
+  ): Promise<void> {
+    switch (kind) {
+      case MultiCommitOperationKind.Squash:
+        return this.recordSquashConflictsEncountered()
+      case MultiCommitOperationKind.Reorder:
+        return this.recordReorderConflictsEncountered()
```
sergiou87, line 1603: Just throwing an idea… I think it's not compile-time safe (or maybe it is with the proper TypeScript knowledge 😂 ), but maybe we can leverage the string values behind the `MultiCommitOperationKind` to compose the key we want to update in the daily measures structure.

Something like…
```
const key = `${kind.toLowerCase()}ConflictsEncounteredCount`
return this.updateDailyMeasures(m => ({
      `${key}`: m[key] + 1,
}))
```

sergiou87 (changes requested):
LGTM! Only have a request about those `throw Error` in the stats code, I'd rather have them logged instead 😳

app/src/lib/stats/stats-store.ts
tidy-dev: :D - I thought of that.. but I have always been told it is kind of bad practice since you are kind skipping typing..

sergiou87 (approved):
LGTM! :shipit:

PR #12410, opened 2021-06-08
Remove cherry-picking drag prompt

tidy-dev:
## Description

Remove cherry-picking drag prompt used when introducing drag and drop. We no longer think it is necessary as drag/drop discoverability doesn't appear to be a problem and now is not the only drag and drop operation.

### Screenshots
**REMOVED** tooltip/prompt shown in below image:
image
## Release notes

Notes: [Removed] Cherry-pick drag prompt removed.


billygriffin (dismissed):
Sounds good! 🪓

app/src/ui/history/selected-commit.tsx
```diff
@@ -276,7 +275,7 @@ export class SelectedCommit extends React.Component<
       return null
     }
 
-    return <DragOverlay dragZoneDescription="branch-button" />
+    return <div id="drag-overlay"></div>
```
sergiou87: Why didn't you remove this function (and the `showDragOverlay` prop) altogether? 🤔 

Just keeping changes to a minimum?

sergiou87 (dismissed):
LGTM, but I'm curious to know why not removing all related to the `renderDragOverlay` method 🤔

app/src/ui/history/selected-commit.tsx
tidy-dev: We still want to only show the drag overlay when this.props.showDragOverlay. I think I am misunderstanding you..

sergiou87 (approved):
:shipit:

PR #12412, opened 2021-06-09
Fix update from default branch not being in merge state

tidy-dev:
## Description

Fix regression that broke update from default branch.

Thanks again @cheshire137!

### Screenshots
https://user-images.githubusercontent.com/75402236/121274829-54038380-c899-11eb-97f4-c709d0e0e927.mov

## Release notes

Notes: [Fixed] Update from default branch works as expected now.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4528,7 +4528,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       opState === null ||
       opState.operationDetail.kind !== MultiCommitOperationKind.Merge
     ) {
-      log.error('[mergeBranch] - Not in merge oerpation state')
+      log.error('[mergeBranch] - Not in merge operation state')
       return
     }
 
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -974,6 +974,38 @@ export class Dispatcher {
     mergeStatus: MergeTreeResult | null,
     isSquash: boolean = false
   ): Promise<void> {
+    const {
+      multiCommitOperationState,
+      branchesState: { tip },
+    } = this.repositoryStateManager.get(repository)
+
+    if (
+      multiCommitOperationState === null ||
+      multiCommitOperationState.operationDetail.kind !==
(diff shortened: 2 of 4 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -4522,7 +4522,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       opState === null ||
       opState.operationDetail.kind !== MultiCommitOperationKind.Merge
     ) {
-      log.error('[mergeBranch] - Not in merge oerpation state')
+      log.error('[mergeBranch] - Not in merge operation state')
```
cheshire137, line 4531: I had grown rathoer fond of this typo. 😢😄

cheshire137 (commented):
Thank you for the fix!

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -970,6 +970,38 @@ export class Dispatcher {
...
+    const {
+      multiCommitOperationState,
+      branchesState: { tip },
+    } = this.repositoryStateManager.get(repository)
+
+    if (
+      multiCommitOperationState === null ||
+      multiCommitOperationState.operationDetail.kind !==
+        MultiCommitOperationKind.Merge
+    ) {
```
sergiou87: Feels weird having this kind of check here knowing `mergeBranch` is also used in other merge entry points to do the actual merge (whereas the multi commit operation was initialized somewhere else). I think I'd rather have a new method in the dispatcher for this that `updateBranch` in app.tsx would call, but this is ok.

In any case, I think it'd be good adding a comment right above this `if` to explain why sometimes there is an ongoing Merge operation, and sometimes there isn't 🤔

sergiou87 (commented):
LGTM! Only have a comment/suggestion.

app/src/ui/dispatcher/dispatcher.ts
tidy-dev: Yeah. I like that better.. this actually broke in 2 places. updateBranch and in the compare mode.

sergiou87 (dismissed):
Thanks for those changes, looks much better now! 😍

sergiou87 (approved):

Issue #62, opened 2016-06-20
Should we hard wrap commit messages?


joshaber:
Enter a commit message and commit.

jacobbednarz:
+1 on making the client hard wrap the commits to be consistent with GitHub & good practices from Git itself.

redhotvengeance:
I found this issue after being dismayed that the new GH Desktop app _doesn't_ wrap, which I was expecting it to do since that's what the previous version did.

At the core of GitHub is Git, and I think it makes sense for GitHub to encourage people to follow Git's best practices. It is widely accepted that Git commit messages should:

- Have a title no longer than 50 chars
- Have an optional body wrapped at 72 chars

Therefore, I think it makes sense for GitHub to enforce these practices at well. Or, at the very least, have their tools optionally enable/disable them. While I primarily interact with Git via the CLI, I enjoy using GitHub Desktop to check diffs, do line-based commits, etc. If GitHub Desktop doesn't properly wrap, I'll have to bounce back to the CLI to compose all commits with a body.

At the end of the day, Git commits should:

- Follow best community practices
- Look good in the terminal (where Git lives)

danny-andrews:
Any update on this? The lack of line length indicators on the desktop app makes it almost unusable.

danny-andrews:
Having a twitter-like counter would be a good solution, I think.

agatac:
> The lack of line length indicators on the desktop app makes it arduous to use as I have to copy and paste commit messages into an editor to check line lengths.

I came here to suggest just that. Length indicator would be useful. So much easier to follow Git's best practices.

shiftkey:
We probably need to think about how small the current commit form is before we can show any fancy UI here. Related issue: #1646

dhoulb:
I'd prefer no hard wrapping, i.e. that GitHub Desktop had a forward-looking approach to Git where wrapping is handled when the message is displayed, not by stealthily modding user input.

I also can't see how it'd be implemented without _disabling_ the current softwrap on that textarea so lines don't wrap, which would be unexpected behaviour for most newbie users. So it probably needs to be enabled by an advanced setting checkbox.

Feels like a more complete solution would be to include this in some potential future functionality that also implements commit templates, so maintainers of projects could specify how they want commits formatted, wrapped, and structured.

(However: I would love a soft character-limit reminder on the title! Shorter titles significantly improve the readability of commit lists. Just painting the excess characters in red or warning orange etc would be enough!)

[...]

Issue #2055, opened 2017-06-19
Warn after typing more than 50 characters for a commit message


tobinibot:
At some point in the distant past, the OS X GitHub Desktop app greyed out any characters past 50 and displayed a little notice about 50 characters being a suggested character limit for commit messages. I’d love to see something like that come back. 

I’m not married to the idea of text getting greyed out (although that was great), I’m more interested in the concept coming back in whatever form would work best in the current app.

I do realize this is a pretty minor feature, but it’d be nice to have on the todo list at some point.

j-f1:
I like the way Atom does it:

<img width="698" alt="screen shot 2017-06-19 at 10 21 40" src="https://user-images.githubusercontent.com/25517624/27290837-6ec82264-54dc-11e7-845d-c770b64493f2.png">
<img width="381" alt="screen shot 2017-06-19 at 10 22 32" src="https://user-images.githubusercontent.com/25517624/27290839-6ee21192-54dc-11e7-8853-49237a266f7c.png">

christophfriedrich:
I think this is already implemented:
![git 50 plus characters greyed out](https://user-images.githubusercontent.com/17781682/27512517-58a7dde2-5943-11e7-9d7e-263461ead543.PNG)
Extremely subtle, but it's there

j-f1:
That’s the old version of the app @christophfriedrich.

ericdrobinson:
I would like to suggest that this gets a full design pass that is focused on communicating the **why** behind the suggested limits. Even [this GitHub blog entry](https://github.com/blog/926-shiny-new-commit-styles) is relatively vague on details as to what the GitHub website does with commit messages in building the visual representation. As far as I can tell, there's nothing in the [GitHub Help Pages](https://help.github.com/) that describes the formatting either.

Currently, the GitHub website will truncate the first line of a commit message to roughly 72 characters. The extended commit message contents show up "below the fold" when you click the "..." button (along with the rest of the first-line [aka summary] in case you went over 72 characters).

[...]

PR #12421, opened 2021-06-10
Add formatter function for commit descriptions

JKRhb:
## Description

This PR adds a formatter regex that wraps commit descriptions word-wise at the 72 character limit. This way you can simply enter a description and don't have to worry about line lengths to avoid a long one-liner description.

It addresses (in part) issue #62. 

#### Before

Description is on one line

grafik

#### After

Description is split across multiple lines

grafik

## Release notes

Notes: Add word-wise wrap to commit descriptions

### DIFF
--- a/app/src/lib/format-commit-message.ts
+++ b/app/src/lib/format-commit-message.ts
@@ -21,6 +21,13 @@ export async function formatCommitMessage(
 ) {
   const { summary, description, trailers } = context
 
+  // Apply word wrap and limit the lines of the description
+  // to 72 characters
+  const formattedDescription = description?.replace(
+    /(?![^\n]{1,72}$)([^\n]{1,72})\s/g,
+    '$1\n'
+  )
+
   // Git always trim whitespace at the end of commit messages
   // so we concatenate the summary with the description, ensuring
   // that they're separated by two newlines. If we don't have a
@@ -28,7 +35,10 @@ export async function formatCommitMessage(
   // all get trimmed away and replaced with a single newline (since
   // all commit messages needs to end with a newline for git
   // interpret-trailers to work)
-  const message = `${summary}\n\n${description || ''}\n`.replace(/\s+$/, '\n')
+  const message = `${summary}\n\n${formattedDescription || ''}\n`.replace(
+    /\s+$/,
+    '\n'
+  )
 
   return trailers !== undefined && trailers.length > 0
     ? mergeTrailers(repository, message, trailers)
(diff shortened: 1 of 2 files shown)


JKRhb:
I fixed a small bug in the code and ran the CI on my fork now with passing tests :) Let me know if there is anything that should be changed.

AmdrewAEddy (dismissed):

JKRhb:
I now improved the function a bit to also cover cases where newlines are already present in the description. In this case the seperate lines are kept and the counter is reset once a "regular line" is reached again. An example for this can be seen in this screenshot where (in the lower part)  only the overly long line is split at the 72 character limit:

grafik

I also added a unit test that asserts this behavior.

JKRhb:
I've updated this now to a much cleaner solution that should be ready for review.

JKRhb:
@billygriffin Sorry to ping you like this, but could you maybe have a look on this PR? I think it would be awesome to format descriptions like this as the need to format descriptions manually to keep their lines within the 72 character limit is actually the main reason I mostly do not add descriptions at all.

JKRhb:
I replaced the function now with a regex and fixed a bug while doing so.

sergiou87:
Hey @JKRhb!! 👋 

Thank you for your contribution ❤️   and sorry it took us this long to get back to you  😳 

We don't think this is the best approach, since it changes the text the user entered without their permission. We were hoping there was a more "indulgent" approach where we suggest the user that the line is too long. For example, something like what you can see in https://github.com/desktop/desktop/issues/2055#issuecomment-310871285

image

But instead of the summary textfield, applied to every line of the description text area, if that makes sense. This is just an idea, we can't promise that would be merged until we see it live and evaluate how it feels 😅 

I hope this sheds some light, sorry we can't give you a better answer 😳

JKRhb:
> Hey @JKRhb!! 👋
> 
> Thank you for your contribution ❤️   and sorry it took us this long to get back to you 😳
> 
> We don't think this is the best approach, since it changes the text the user entered without their permission. We were hoping there was a more "indulgent" approach where we suggest the user that the line is too long. For example, something like what you can see in [#2055 (comment)](https://github.com/desktop/desktop/issues/2055#issuecomment-310871285)
> 
> image
> 
> But instead of the summary textfield, applied to every line of the description text area, if that makes sense. This is just an idea, we can't promise that would be merged until we see it live and evaluate how it feels 😅
> 
> I hope this sheds some light, sorry we can't give you a better answer 😳

Thank you for your reponse! Your reasoning is totally understandable, I had to notice myself that my approach was a bit too "invasive" and could also be irritating if you hit the undo button. I think some kind of helper that suggests adding a linebreak is a good idea :) Thank you for dealing with this issue!

PR #12422, opened 2021-06-10
Check when the drag&drop for reordering can be ignored

sergiou87:
## Description

Before this PR, the app always attempted to move the commits when they're dropped to any position in the `History` tab.

However, sometimes that's unnecessary because the commits are dropped to a position where reordering has no effects at all.

This PR addresses that by checking if the commits dropped are consecutive, and they're dropped in a position that falls within the range of the dropped commits.

### Screenshots

https://user-images.githubusercontent.com/1083228/121500035-3855c580-c9de-11eb-9e43-7251000ac122.mov

## Release notes

Notes: no-notes


billygriffin (approved):
This makes sense to me and really nice improvement. Doing the reorder with no effect is strange.

PR #12429, opened 2021-06-10
Clear merge operations state after successful merge so rebases can start

sergiou87:
## Description

This PR fixes a regression in the upcoming release, where completing a merge wouldn't clear the state of the app, so new operations like a rebase wouldn't start.

### Screenshots

Before the fix:

https://user-images.githubusercontent.com/1083228/121554281-50473c80-ca12-11eb-9a52-0d6a2c81d0d4.mov

After the fix: In the video, you can see how after a successful merge with `development`, if I try to start a rebase, the dialog pops up as expected. In the latest beta, that doesn't happen and the app grays out all options in the Branch menu, making it impossible to proceed with any other operation.

https://user-images.githubusercontent.com/1083228/121554497-82589e80-ca12-11eb-949e-5b85d1d44ed9.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4587,6 +4587,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         // successfully after conflicts in `dispatcher.finishConflictedMerge`.
         this.statsStore.recordSquashMergeSuccessful()
       }
+      this._endMultiCommitOperation(repository)
     } else if (
       mergeResult === MergeResult.AlreadyUpToDate &&
       tip.kind === TipState.Valid
@@ -4596,6 +4597,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         ourBranch: tip.branch.name,
         theirBranch: sourceBranch.name,
       })
+      this._endMultiCommitOperation(repository)
     }
 
     return this._refreshRepository(repository)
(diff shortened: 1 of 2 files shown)


app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -3361,7 +3361,7 @@ export class Dispatcher {
       tip
     )
 
-    this.processMultiCommitOperationRebaseResult(
+    return this.processMultiCommitOperationRebaseResult(
```
sergiou87, line 3364: This one is for free 😆 

To make sure whoever `await`s for this function, is also `await`ing for `processMultiCommitOperationRebaseResult`. You can see how we do exactly this for `reorderCommits` too in https://github.com/desktop/desktop/pull/12429/files#diff-c8d5cceaa76fd0fe1490c522f46fa1c4b804fffe63ca13acaf0b272578a40b59R3258

app/src/lib/stores/app-store.ts
```diff
@@ -4596,6 +4597,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         ourBranch: tip.branch.name,
         theirBranch: sourceBranch.name,
       })
+      this._endMultiCommitOperation(repository)
```
sergiou87, line 4600: These two `this._endMultiCommitOperation(repository)` lines make sure the flow ends when the merge finishes and there are no conflicts. If there are conflicts, the multi commit operation will be ended after the conflicts are resolved.

billygriffin (approved):
Woof, that's a bad one. I don't understand the logic but tested and seems good!

Issue #1644, opened 2017-05-19
Enhancement: Amend last commit


1pete:
### Description

Amend last commit feature is very useful and exists in many git clients, but not in GitHub Desktop.
It would be nice if we have it in GitHub Desktop too.

Current workaround is click `Undo` and commit again but I've to carefully check what I want to commit again.

### Version
**GitHub Desktop version:** 0.5.5

Thanks.

xt0rted:
I typically commit the missed/extra changes and then rebase & squash, this would certainly speed that process up.

joshaber:
Thanks for the suggestion!

This idea is interesting for the future, but this is beyond the scope of our [current roadmap](https://github.com/desktop/desktop/blob/master/docs/process/roadmap.md).

I'm going to add the `future-proposal` label to this issue and close it out, so that we don't get overwhelmed with our backlog! We'll revisit these as part of our next planning session.

levinit:
useful function

meyer:
@joshaber do you all accept PRs for features outside the current roadmap?

j-f1:
@desktop/design Does this need some design 💭?

If not, feel free to go ahead, @meyer.

shiftkey:
Please hold off on this. We haven't had a chance to evaluate these proposals, and I'm pretty sure this needs some design thought. @meyer if you have thoughts on how this should look/feel/behave feel free to share them in here as a starting point.

DanTup:
I find myself really often committing things with the message "f" as a reminder to `rebase -i` and mark as `fixup` later. Sometimes I forget and I push these changesets :(

Would be good to see this - maybe a tickbox near the commit button for "Amend last commit" - ticking it should allow committing without entering a summary/description (or ticking it could pre-fill the previous messages if the fields are blank, for easier typo-correction).

Oh, and show a warning if the previous changeset may have been pushed ;-) (I'm not sure if Git has any equiv of Mercurial's phases?)

JaneSmith:
For what it's worth, the three major reasons I use SourceTree instead of GitHub Desktop whenever possible are:
1. Multi-file selection (for discarding, staging, etc.).
2. Branch relationship visualisation.
3. Amend last commit.

The first is apparently being worked on already in a pull request, and hopefully will be in GitHub Desktop soon. The second one is kind of on the roadmap, although perhaps not in the form I want (a visual graph). That leaves the last one. Being able to quickly and easily amend the last commit is very useful to me.

[...]

PR #12464, opened 2021-06-16
 Add Amend button next to Undo for the last unpushed commit

Twixes:
This provides an enhancement to #12353 and adds an even more convenient UX solving #1644.

## Description

This is a feature I've been missing for quite some time.
I was happy to see #12353 going live, but it's not quite as handy as I'd like – the convenience of Undo in the sidebar is specifically the reason why I've been using that button _a lot_. But, fortunately now with Amend being easy to dispatch, it was pleasant to implement a useful Amend button right next to Undo, making amending a breeze.

Also increased `defaultSidebarWidth` a bit from 250 to 275, so that there's still comfortable room the for commit message (and time).

### Screenshots

Amend

## Release notes

Notes: [New] Amend the last local commit handily from the staging area

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -303,7 +303,7 @@ const RecentRepositoriesKey = 'recently-selected-repositories'
  */
 const RecentRepositoriesLength = 3
 
-const defaultSidebarWidth: number = 250
+const defaultSidebarWidth: number = 275
 const sidebarWidthConfigKey: string = 'sidebar-width'
 
 const defaultCommitSummaryWidth: number = 250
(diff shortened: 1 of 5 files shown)


billygriffin:
Thanks @Twixes, but we chose not to just add it next to undo purposely, at least initially, as we don't want to confuse people who are uncertain what the differences between the two are. I totally understand how it might be more convenient to have it right there as well, but it will also create confusion. So at least for now we're going to leave this as is, and likely iterate in some manner down the road as we see how each of them are used.

Twixes:
I understand that perspective, though it is a bit of a shame. As much as I appreciate the simplicity of GitHub Desktop, it would be nice to have some slightly more advanced options (like this one) at hand – maybe optionally.

PR #12465, opened 2021-06-16
Alert when commit summary exceeds cutoff limit

### LINKED ISSUE #2055: Warn after typing more than 50 characters for a commit message

Twixes:
Closes #2055.

## Description

This adds visual feedback regarding commit summary length, using the textbox's border:
- When ideal length (50 characters or less), the textbox is blue as usual. All is well.
- When a bit longer, but still below or at cutoff length (72 characters or less), the textbox is calmly yellow. This is still an okay length, though getting near the limit.
- When above cutoff length (72 characters or more), the textbox is alarmingly red. Since GitHub Desktop will messily cut the message off right at the 72 character mark, the user should be very deliberate about exceeding this.

### Screenshots

Commit summary length feedback in action

## Release notes

Notes: [New] Visual feedback on commit summary length

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -35,6 +35,10 @@ import { setGlobalConfigValue } from '../../lib/git/config'
 import { PopupType } from '../../models/popup'
 import { RepositorySettingsTab } from '../repository-settings/repository-settings'
 import { isAccountEmail } from '../../lib/is-account-email'
+import {
+  IdealSummaryLength,
+  MaxSummaryLength,
+} from '../../lib/wrap-rich-text-commit-message'
 
 const addAuthorIcon = new OcticonSymbol(
   18,
@@ -731,6 +735,8 @@ export class CommitMessage extends React.Component<
 
     const summaryInputClassName = classNames('summary-field', 'nudge-arrow', {
       'nudge-arrow-left': this.props.shouldNudge === true,
+      'near-limit': this.state.summary.length > IdealSummaryLength,
+      'over-limit': this.state.summary.length > MaxSummaryLength,
     })
 
     return (
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -43,11 +43,11 @@ const addAuthorIcon = new OcticonSymbol(
   18,
   13,
   'M14 6V4.25a.75.75 0 0 1 1.5 0V6h1.75a.75.75 0 1 1 0 1.5H15.5v1.75a.75.75 0 0 ' +
-    '1-1.5 0V7.5h-1.75a.75.75 0 1 1 0-1.5H14zM8.5 4a2.5 2.5 0 1 1-5 0 2.5 2.5 0 0 1 5 ' +
-    '0zm.063 3.064a3.995 3.995 0 0 0 1.2-4.429A3.996 3.996 0 0 0 8.298.725a4.01 4.01 0 0 ' +
-    '0-6.064 1.91 3.987 3.987 0 0 0 1.2 4.43A5.988 5.988 0 0 0 0 12.2a.748.748 0 0 0 ' +
-    '.716.766.751.751 0 0 0 .784-.697 4.49 4.49 0 0 1 1.39-3.04 4.51 4.51 0 0 1 6.218 ' +
-    '0 4.49 4.49 0 0 1 1.39 3.04.748.748 0 0 0 .786.73.75.75 0 0 0 .714-.8 5.989 5.989 0 0 0-3.435-5.136z'
+  '1-1.5 0V7.5h-1.75a.75.75 0 1 1 0-1.5H14zM8.5 4a2.5 2.5 0 1 1-5 0 2.5 2.5 0 0 1 5 ' +
+  '0zm.063 3.064a3.995 3.995 0 0 0 1.2-4.429A3.996 3.996 0 0 0 8.298.725a4.01 4.01 0 0 ' +
+  '0-6.064 1.91 3.987 3.987 0 0 0 1.2 4.43A5.988 5.988 0 0 0 0 12.2a.748.748 0 0 0 ' +
+  '.716.766.751.751 0 0 0 .784-.697 4.49 4.49 0 0 1 1.39-3.04 4.51 4.51 0 0 1 6.218 ' +
+  '0 4.49 4.49 0 0 1 1.39 3.04.748.748 0 0 0 .786.73.75.75 0 0 0 .714-.8 5.989 5.989 0 0 0-3.435-5.136z'
 )
 
 interface ICommitMessageProps {
@@ -761,7 +761,7 @@ export class CommitMessage extends React.Component<
             onContextMenu={this.onAutocompletingInputContextMenu}
             disabled={this.props.isCommitting === true}
             spellcheck={this.props.commitSpellcheckEnabled}
-            warningMessage={charactersOverLimit > 0 ? { title:'This commit summary is too long',description: <>The summary is limited to {MaxSummaryLength} characters, which you've exceeded. <strong>{charactersOverLimit} {charactersOverLimit === 1 ? 'character':'characters'}</strong> going over the limit will be cut off.</> }: undefined}
+            warningMessage={charactersOverLimit > 0 ? { title: 'This commit summary is very long', description: <>The summary is limited to {MaxSummaryLength} characters, which you've exceeded. <strong>{charactersOverLimit} {charactersOverLimit === 1 ? 'character' : 'characters'}</strong> going over the limit will be moved into the commit description.</> } : undefined}
           />
         </div>
 
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -36,7 +36,6 @@ import { PopupType } from '../../models/popup'
 import { RepositorySettingsTab } from '../repository-settings/repository-settings'
 import { isAccountEmail } from '../../lib/is-account-email'
 import {
-  IdealSummaryLength,
   MaxSummaryLength,
 } from '../../lib/wrap-rich-text-commit-message'
 
@@ -354,7 +353,7 @@ export class CommitMessage extends React.Component<
     const accountEmails = repositoryAccount?.emails.map(e => e.email) ?? []
     const email = commitAuthor?.email
 
-    const warningBadgeVisible =
+    let warningBadgeVisible =
       email !== undefined &&
       repositoryAccount !== null &&
       repositoryAccount !== undefined &&
@@ -734,10 +733,10 @@ export class CommitMessage extends React.Component<
     })
 
     const summaryInputClassName = classNames('summary-field', 'nudge-arrow', {
-      'nudge-arrow-left': this.props.shouldNudge === true,
-      'near-limit': this.state.summary.length > IdealSummaryLength,
-      'over-limit': this.state.summary.length > MaxSummaryLength,
+      'nudge-arrow-left': this.props.shouldNudge === true
     })
+
+    const charactersOverLimit = this.state.summary.length - MaxSummaryLength
 
     return (
       <div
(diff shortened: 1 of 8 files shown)


Twixes:
Bump?

nxcco:
Could one review that one real quick?

nelsoncardenas:
Simple and elegant solution. I hope this feedback to be applied soon

tidy-dev (dismissed):
✨  This looks like a great warning system! Would love our designer to get a look at it though before moving forward on it.

billygriffin (changes requested):
Thanks @Twixes for taking a stab at this! After talking through this internally, we have a few concerns:

1. Red feels incorrect because its use in traditional forms indicates that the form cannot be submitted, where that's not the case here.
2. There's no indication of why people are seeing yellow and red, and they're not very subtle so it feels too heavy handed.

Would you mind trying out something similar to the previous version of the app, where we just used a slightly lighter grey in the text after 50 characters? This would make it apparent for those who really care about this, but be less of a thing in the way for people who are just trying to get things done.

billygriffin:
Hi @Twixes, just wanted to follow up and see if you're open to adjusting this as I mentioned previously. I'll go ahead and close the PR if not and we can always reopen if/when you want to pick it back up and take it in a bit of a different direction. Thanks!

nxcco:
@Twixes Do you think you could make those changes? The world is needing this feature.

nelsoncardenas:
I agree with @nxcco, this small feature has been requested for more than 4 years, and I will prefer the @Twixes solution over nothing

Twixes:
Sure! I've been on vacation in France for the past 2 weeks so didn't have time for this PR, but I'll work a slightly different solution out this weekend for you.

billygriffin:
@Twixes No super rush, just wanted to make sure you still wanted to do it with a different direction. Hope you had a wonderful vacation!

Twixes:
After looking into this, I see three ways to go about this:

1. Implementing your suggestion, that is colored characters in the input field. Unfortunately that would be kind of hacky technically, as the `input` element does not allow this – a way out would be to set the `input` element's text color to transparent and show the typed characters using extra `span`s underneath it (which _would_ allow coloring of overflow characters).
1. Keeping the current proposed style, but also adding an info icon when the border is non-blue, that would explain first-time users what this indicates, and making the "over limit" indication orange.
1. Keeping the current proposed style, but ditching the "near limit" indication and making the "over limit" indication orange. To be honest, this would make me perfectly happy as I only care about GitHub Desktop _cutting my message off_ (72+ chars case), and I believe this would be simple enough to be intuitive as is (only notifying the user in case of a definite threat to their commit message).

What's your take on this? I believe ALL of the above solutions (and also the current one as is) are much better than what the product does currently, which is silently cutting messages off without warning. I currently have to manually test if I'm exceeding the cutoff length every time I make a commit with a longer message, and if it is, I have to amend the commit with a shorter message – an annoying user experience!

nxcco:
Personally, I think the best choice is number three, as it's intuitive and every serious Git user will understand it. If it's not a serious Git user, then they will google it or ask what this orange border means on Stack. So in the WORST CASE you have to do what you do anyway as a software dev: looking it up.
@Twixes Did you still indicate subject lines longer than 50 chars?

Twixes:
Hey GitHub team, what do you think? :) Happy to implement a solution, let's just first choose which option, to avoid changing direction and wasted work later.

sergiou87:
Thanks @Twixes for your patience! We discussed today about it and we think we're gonna go ahead with this:
- Nothing up to 72 characters
- Yellow border after 72 characters, along with a warning icon of the same color. Something like this:

image

When you hover over the icon (or when you click it), the app will show a message explaining why the yellow color (i.e. explaining users why >72 characters is bad).

I think it could mimic what we do for misattributed commits #11591:

image

That is already using a warning icon (`OcticonSymbol.alert`) and displays a `Popover` component with a message when it's clicked.

That shouldn't be crazy to implement (🤞 ) and also gives the user better feedback about _why_ the textbox turns yellow.

Does that sound good?

Twixes:
That's a great proposal! I've pushed an implementation of it.

I didn't end up replicating the "warning icon _inside_ input box" design you linked, as it's not really possible to do it without significant reengineering of `AutocompletingTextInput` without breaking input text display (i.e. it could be done by just _overlaying_ the warning icon, but then that would bypass input layout and so text would be covered by the icon, which is exactly what we don't want in this situation) – or it'd be kind of hacky, which I'd rather avoid.

I did use `Popover` (hover-based), and also borrowed the warning badge. Here's what this looks like in action:
Alert! Alert!

sergiou87:
Thank you @Twixes, we love that approach!!! 😍 

The only thing that requires some changes is the wording inside the popover, since those extra characters will actually be cut off, right?

Twixes:
Thanks. Hm, I don't quite get what about the wording should change, since it does say how many characters will actually be cut off (unless you delete enough for none to be cut off, in which case the popover disappears).

sergiou87:
Sorry I wasn't clear enough 😞 

We think the text is misleading because the whole text will make it to the commit even if it's longer than 72 characters. It will just be truncated in GitHub Desktop (and maybe other apps) and move the rest of characters to the description text: #2575

I'll discuss about this with the team and see if we can come up with a better wording. Of course, we're more than open to suggestions! 😄

app/src/ui/changes/commit-message.tsx
```diff
@@ -756,6 +761,7 @@ export class CommitMessage extends React.Component<
             onContextMenu={this.onAutocompletingInputContextMenu}
             disabled={this.props.isCommitting === true}
             spellcheck={this.props.commitSpellcheckEnabled}
+            warningMessage={charactersOverLimit > 0 ? { title:'This commit summary is too long',description: <>The summary is limited to {MaxSummaryLength} characters, which you've exceeded. <strong>{charactersOverLimit} {charactersOverLimit === 1 ? 'character':'characters'}</strong> going over the limit will be cut off.</> }: undefined}
```
billygriffin: Thoughts on something like this? We don't want to indicate that it's going to delete something that's over 72 characters, so just trying to be more clear about that.
```suggestion
            warningMessage={charactersOverLimit > 0 ? { title:'This commit summary is longer than recommended',description: <>The summary will be truncated after {MaxSummaryLength} characters, which you've exceeded. <strong>{charactersOverLimit} {charactersOverLimit === 1 ? 'character':'characters'}</strong> going over the limit will be cut off from the summary and wrap into the description.</> }: undefined}
```

billygriffin (commented):
Left a suggestion for making the message clearer, but not wedded to this and curious what others think.

app/src/ui/changes/commit-message.tsx
sergiou87: After looking into the origin of this feature request, I think it'll make more sense to have something like… "It is recommended to keep the commit summary under 50 characters, which you have exceeded by X characters. Learn more.", where "Learn more" is a link to https://git-scm.com/docs/git-commit#_discussion

And then, of course, use a different limit constant of 50 (leave that 72 for UI purposes when long summaries are wrapped into the commit description) to match the official recommendation in Git's documentation.

What do you think?

Twixes: Thanks for clarifying, I see your point. My thoughts on solutions are:
- The popover's title needs to be short enough to fit in without a line break, otherwise it takes up an awkward amount of space. Hence "This commit summary is longer than recommended" isn't great, as it's super verbose and would require at least two lines. I went for "This commit summary is very long" instead – somewhat less authoritative than before, but still really concise.
- It's a great point that the extra characters actually go into the description and don't just disappear. Also wanting to avoid saying "the summary" twice, I went for this: "The summary is limited to 72 characters, which you've exceeded. **n** characters going over the limit will be moved into the commit description." It's also reasonably short for the amount of information, but it is factual: the summary CAN'T be longer than 72 in GitHub Desktop, so it is a hard limit over which the commit description comes into play.
- I think 50 characters is _a_ guideline, but personally I find it debatable and including it in the app would have low impact – it's maybe a nice-to-have. The actually painful thing is summaries getting truncated without warning, so I believe shipping this warning is a lot higher priority than a "nudge" towards 50, I'd rather move forward without it for now.

Twixes:
Bump?

billygriffin:
Hey @Twixes, thanks for your patience and apologies for such a long back and forth. We're debating this a bit more internally and trying to figure out how heavy handed we really want to be here. Feel free to nudge us again if you haven't heard back in a couple weeks, and apologies again.

niik:
Hey @Twixes, thanks for sticking with this for so long!

We discussed this at a recent design review and have taken another look at how we'd like to see this implemented. For future reference we did investigate a grayed-out approach as was mentioned before and have a proof of concept of that working by utilizing CodeMirror as the text input similar to how the author input component work. Ultimately we decided against that approach due to the risk involved in swapping out our most frequently used input for a (in Desktop) unproved approach.

Instead we went back to Here's a quick mock

commitsummary

The key differences here being

- Not using warning symbology to convey a hint, instead opting for the `Tip!` icon (lightbulb). Warnings should be reserved for potentially dangerous operations which this isn't. Additionally we feel like using a warning (or colors often associated with warnings) isn't subtle enough for the users who choose not to adhere to the 50 character convention.
- Aligning our messaging with GitHub.com (and showing the lightbulb at 50+ characters)

image

- Inline hint (I'll address the [trickiness](https://github.com/desktop/desktop/pull/12465#issuecomment-904123562) of implementing this below)
- Primer.style tooltips (I'm working on this separately so using a `title` attribute for now is just fine)

> I didn't end up replicating the "warning icon _inside_ input box" design you linked, as it's not really possible to do it without significant reengineering of `AutocompletingTextInput` without breaking input text display (i.e. it could be done by just _overlaying_ the warning icon, but then that would bypass input layout and so text would be covered by the icon, which is exactly what we don't want in this situation)

I think we could render the warning outside of the `AutocompletingTextInput` (rendering popovers and handling subject-specific length doesn't seem like a concern that component should have). I absolutely agree that we shouldn't cover the icon, but I think we can avoid that by using padding on the input element and positioning the icon on top of the textbox itself. I did a super quick proof of concept in https://github.com/desktop/desktop/compare/inline-overflow-hint-spike.

How would you feel about taking a stab at what I've described above?

Twixes:
Hey! Thanks for getting back to this.
I think that's some really good thoughts on UI and I'm happy to use the proposed design in the implementation.

It seems however that we've confused end goals here:

- I am putting focus on the 72 characters hard limit, which _is_ a potentially dangerous operation in GitHub Desktop, as it simply **cuts off the commit message**. This should be given a warning ASAP, in my opinion as someone who uses GitHub Desktop _every day_ and commits with messages of various lengths.
- You are putting focus on the 50 characters suggestion, which is merely a guideline that can be mentioned. I see the value in alignment with GitHub.com, but – as you pointed out – that is just **not** a dangerous situation.

Eventually I'd love both to be implemented (and I'm happy to help out with this too), but the design needs to incorporate the 72+ chars case _before_ the 50+ chars one.

niik:
> * **cuts off** the commit message

GitHub Desktop will *wrap* the commit subject after roughly 72 characters[^1] to ensure it can be read in its entirety. This is not a hard wrap though, it's only a presentation concern. 

In an ideal situation GitHub Desktop wouldn't be wrapping at a set number of characters to begin with but instead wrap when necessary depending on the dimensions of the window. The only reason we don't do that now is that the implementation of that is... complicated to say the least.

> * I am putting focus on the 72 characters hard limit, which _is_ a potentially dangerous operation [...] This should be given a warning ASAP, in my opinion

While I can see getting your subject wrapped in Desktop being a source of irritation our definition of what counts as a dangerous operation differs here. No information is lost when the subject is wrapped.

Given that the wrapping occurs based on _rendered characters_[^1] and not the actual number of characters in the textbox the notification about wrapping can't be accurate all the time, that and the fact that GitHub.com doesn't do any warning for subject wrapping makes me feel like we should stick to a hint about the message starting to become lengthy.

[^1]: Saying that we wrap at 72 characters isn't especially accurate as we do tokenization of commit messages. As an example a commit message with a link to an issue (i.e. `https://github.com/desktop/desktop/issues/12345`) will actually render just the issue number (i.e. `#12345`) which is significantly shorter than what we would use to count characters in this implementation (there are other examples such  as `:smile:` which counts as two characters when rendered in the history view). We could use tokenization before counting but that's both much too complicated for what we're trying to achieve and also too tightly coupled to Desktop's current rendering implementation.

Twixes:
Thanks, I looked more into how Desktop wraps things and made use of your design. This is now superseded by https://github.com/desktop/desktop/pull/13075 – let me know what you think about that. I built it on top of https://github.com/desktop/desktop/compare/inline-overflow-hint-spike.

PR #12479, opened 2021-06-18
Fix: Missing right border in FilterList and in BranchesContainer tab bar

### LINKED ISSUE #12478: Right border missing in FilterList and in BranchesContainer tab bar
### Describe the bug

There is no right border in FilterList and in BranchesContainer tab bar. 

### Version & OS

Version 2.9.0 (x64)

1. Open application
2. Click on `Current repository` button or `Current branch` button
3. See that there is no right border which is inconsistent with the rest of the app UI

### Screenshots

Example of UI with border

Examples of UI with missing border:

meJevin:
Right border was missing in FilterList and BranchesContainer.
Closes #12478

I'm looking at other parts of the UI and this little issue is driving my OCD crazy so I decided to fix it

## Description

Modified a little bit of CSS to fix the underlying issue.

I've used the same css varible for border as you use everywhere else.

### Screenshots

Example of UI with border:

Examples of UI with missing border:

### DIFF
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -165,7 +165,7 @@ body.theme-dark {
 
   --toolbar-button-active-color: var(--text-color);
   --toolbar-button-active-background-color: var(--background-color);
-  --toolbar-button-active-border-color: var(--box-border-color);
+  --toolbar-button-active-border-color: var(--background-color);
 
   --toolbar-button-progress-color: #{$gray-800};
   --toolbar-button-focus-progress-color: #{$gray-700};
--- a/app/styles/ui/_branches.scss
+++ b/app/styles/ui/_branches.scss
@@ -8,7 +8,6 @@
 
   & > .tab-bar {
     border-top: var(--base-border);
-    border-right: var(--base-border);
   }
 
   .branches-list {
--- a/app/styles/ui/_branches.scss
+++ b/app/styles/ui/_branches.scss
@@ -8,6 +8,7 @@
 
   & > .tab-bar {
     border-top: var(--base-border);
+    border-right: var(--base-border);
   }
 
   .branches-list {
--- a/app/styles/ui/_filter-list.scss
+++ b/app/styles/ui/_filter-list.scss
@@ -23,6 +23,8 @@
     // are larger than the available space.
     min-height: 0;
   }
+
+  border-right: var(--base-border);
 }
 
 .filter-list-group-header {
@@ -45,6 +47,7 @@
     display: flex;
     flex-direction: column;
     border-top: var(--base-border);
+    border-right: var(--base-border);
     padding: var(--spacing);
   }
 
--- a/app/styles/ui/_filter-list.scss
+++ b/app/styles/ui/_filter-list.scss
@@ -23,8 +23,6 @@
     // are larger than the available space.
     min-height: 0;
   }
-
-  border-right: var(--base-border);
 }
 
 .filter-list-group-header {
@@ -47,7 +45,6 @@
     display: flex;
     flex-direction: column;
     border-top: var(--base-border);
-    border-right: var(--base-border);
     padding: var(--spacing);
   }
 


meJevin:
Attaching screenshots of UI with fixed right border issue:
fixed_1
fixed_2

meJevin:
Dang, I just noticed that I didn't create a separate branch for this PR. Tell me if that's OK or not I'll fix it right up and open up a new PR if needed.

sergiou87 (changes requested):
OMG I didn't notice this and now I can't unsee it 😱 

I think these changes don't actually fix the underlying issue. I just checked and it works fine with the light theme:

image

If you zoom in, there are some subtle differences that make the effect that triggered your OCD:
image image

In the light theme, the right border of the button in the top bar disappears, so I think we should probably make sure the same happens in the dark theme.

Other than that, your changes feel too specific to the repository and branch lists, while this should require a more generic fix either in the `.foldout` or `toolbar-button` classes 🤔

sergiou87:
> Dang, I just noticed that I didn't create a separate branch for this PR. Tell me if that's OK or not I'll fix it right up and open up a new PR if needed

Don't worry about that! As long as it's in another repo is ok for us 😄

meJevin:
> Other than that, your changes feel too specific to the repository and branch lists, while this should require a more generic fix either in the `.foldout` or `toolbar-button` classes 🤔

Sure, I'll take a look.

meJevin:
> In the light theme, the right border of the button in the top bar disappears, so I think we should probably make sure the same happens in the dark theme.

Setting border of `.foldout`. Here's what's happening when I set `border-right` on `.foldout` 
image

For some reason not the entire foldout is covered with the right border. Only the upper part is covered with the border..

meJevin:
@sergiou87 or are you saying that we should ditch the borders and just disable the right border for `Current repository` button when it's open?

meJevin:
In the case of latter I guess setting `border-right: none` in `ui/toolbar/_dropdown.scss` could work. There's a CSS selector for when it's open and it's a toolbar button.

Here's how it looks:
Dark:
image
image
Light:
image
image

sergiou87:
@meJevin yes and no 😂 

Basically what I had this in mind: 
1. Understand why the button appearance is different on Dark and Light themes.
2. Remove whatever the difference is.

I just took a quick look at the CSS and found why they're different:
- Dark theme: uses the regular border color when it's active https://github.com/desktop/desktop/blob/9630999f5e32f9aef4c3a0ae54ba88153aa6b588/app/styles/themes/_dark.scss#L168
- Default (Light theme): uses the background color when it's active https://github.com/desktop/desktop/blob/9630999f5e32f9aef4c3a0ae54ba88153aa6b588/app/styles/_variables.scss#L229

Since the Light theme is the original one, I assumed (maybe incorrectly) that's the correct behavior, but looking at your screenshots of the Dark theme without the border, it also looks kind of wrong since there is much less contrast with the other button than in the Light theme. Maybe @ampinsk can chime in and tell us what she thinks the right approach is!

meJevin:
@sergiou87 sure, let's wait for @ampinsk and we'll move on from there.

ampinsk:
Hey, thanks for this thorough investigation y'all! I do think the fix is what @sergiou87 found: there's a misalignment between the dark and light themes on that variable. The "foldouts" aren't intended to have borders, since they sit "on top" of the interface. I think they end up looking like they have borders on one side because they sit on top of elements that do have a border. But that line being jagged like that is definitely annoying, I like your approach @sergiou87 to align that button border across the themes!

meJevin:
> I like your approach @sergiou87 to align that button border across the themes

So we're choosing this variation then? https://github.com/desktop/desktop/pull/12479#issuecomment-864001693 

In that case we don't wanna get rid of the border like I did in the CSS, but rather match the border color of active button to it's background. The result will look the same as in the comment I mentioned, but the change will be the following: 

Change this: 
https://github.com/desktop/desktop/blob/9630999f5e32f9aef4c3a0ae54ba88153aa6b588/app/styles/themes/_dark.scss#L168

To: `--toolbar-button-active-border-color: var(--background-color);`

meJevin:
I've committed the changes so you can see the result.

meJevin:
How it looks on Mac:
image
image

sergiou87 (approved):
This is great! Thank you @meJevin for your contribution!!! ❤️ 

Can't wait to have this in a build so I can effectively unsee the bug 😂 😂

app/styles/themes/_dark.scss
```diff
@@ -165,7 +165,7 @@ body.theme-dark {
 
   --toolbar-button-active-color: var(--text-color);
   --toolbar-button-active-background-color: var(--background-color);
-  --toolbar-button-active-border-color: var(--box-border-color);
```
get-bent-gaming, line 168: --toolbar-button-active-border-color: var(--background-color);

PR #12485, opened 2021-06-20
Added *.tsbuildinfo, .npm and .node_repl_history

Walledgarden:
*.tsbuildinfo => Build info from TypeScript Builds
.npm => Node Package Manager Cache
.node_repl_history => Repl history of node


sergiou87:
Thank you for your PR! But after discussing it, we'll only accept changes in `.gitignore` for tools that we use (and support directly). Sorry!

PR #12492, opened 2021-06-22
Add a menu item to view a committed file change on GitHub

haykam821:
This pull request brings consistency to context menus by adding the 'View on GitHub' menu item to files in the committed file change list.

### Screenshots

Context menu with 'View on GitHub' menu item

Clicking this menu item will open a browser window to the commit on GitHub with the selected file highlighted.

## Release notes

Notes:

- [Added] "View on GitHub" menu item in the context menu for committed file changes


tidy-dev:
@haykam821 Thanks for the PR. Just curious if there is a workflow that you have that makes this feature particularly helpful to you.

haykam821:
For commits with large amounts of changes, I'd rather get the link from GitHub Desktop instead of using the comparatively clunky GitHub web UI.

tidy-dev:
Thanks for the feedback. We are a bit apprehensive to add context menus/features without clear value impact (this may be great for you, but maybe clutter to many other users) . Sorry but do you mind elaborating. Since you are requesting to open the file on github.com, I believe I am understanding that the clunky Github web UI refers to GitHub Desktop code changes viewer? Is there a particular feature from .com you are looking for?

Rexogamer:
From how I'm understanding it, they meant that instead of going through the list of commits on .com and finding the specific file diff there, they wanted a direct link?

app/src/ui/history/selected-commit.tsx
```diff
@@ -350,9 +350,40 @@ export class SelectedCommit extends React.Component<
...
+      gitHubRepository.endpoint !== getDotComAPIEndpoint()
+    ) {
+      viewOnGitHubLabel = 'View on GitHub Enterprise'
+    }
+
+    items.push({
+      label: viewOnGitHubLabel,
+      action: () => this.onViewOnGitHub(file),
+      enabled:
+        !this.props.isLocalRepository &&
```
sergiou87: I assume this is supposed to mimic the `this.props.isLocal` check from https://github.com/desktop/desktop/blob/5d230b88021428960271180dc6ef392577fd204d/app/src/ui/history/commit-list-item.tsx#L371

However, that `this.props.isLocal` checks if the specific commit is local (i.e. not pushed) or not, since `View on GitHub` doesn't make sense for commits that only exist locally.

I just tested the current implementation and you can try and open in GitHub commit files that haven't been pushed:

2021-07-26 at 18 20

2021-07-26 at 18 21

sergiou87 (changes requested):
Thank you for this @haykam821!! We decided to go ahead with this feature, but first we'd need you to address the bug I mention in the comments.

Other than that, the implementation looks good to me 😄

app/src/ui/history/selected-commit.tsx
haykam821: Fixed in b054452d87a5fed833844015e682af2c29302589

app/src/ui/repository.tsx
```diff
@@ -400,6 +400,11 @@ export class RepositoryView extends React.Component<
     const selectedCommit =
       sha != null ? this.props.state.commitLookup.get(sha) || null : null
 
+    const isLocal =
+      selectedCommit == null
+        ? false
+        : this.props.state.localCommitSHAs.includes(selectedCommit.sha)
```
sergiou87: I personally find this a bit cleaner and less confusing:

```typescript
const isLocal = selectedCommit != null && this.props.state.localCommitSHAs.includes(selectedCommit.sha)
```

(beware of formatting if you copy&paste it, or our CI will complain 😅 )

sergiou87 (commented):
I haven't tested it again (yet), but found a nit while reviewing it. I'll let you know when I've tested it 😄  Thanks for your work on this! ❤️

app/src/ui/repository.tsx
haykam821: Fixed in 2feae3e1447fd2bc0d18e4838e17b769a66a7b5f

sergiou87 (approved):
Works like a charm!! Love it 😍  Thank you!

PR #12529, opened 2021-06-28
Improve time committing

sergiou87:
## Description

This PR tries to address an issue where committing _seems_ to take too long, specially in large repos.

The problem is the commit time is short, but right after it we're doing some work to prepare the UI for that new commit. During that time, the commit button is disabled and shows a "loading" indicator.

This work consists of: 1) update the list of changes and 2) refreshing the whole repository.

We think (1) is essential to complete before we let the user make another commit, otherwise the changes list might show stuff that has been committed already.

(2) however might be necessary, but there is no need to block the user until that's finished. This PR keeps triggering that repository status update but makes it asynchronous. Thanks to that we've seen improvements of up to 50%.

## Release notes

Notes: [Improved] Reduced time needed to make a commit

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -2638,11 +2638,14 @@ export class AppStore extends TypedBaseStore<IAppState> {
           }
         })
 
-        await this._refreshRepository(repository)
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨

PR #12540, opened 2021-06-30
Add file name to file exceeds size limit error

### LINKED ISSUE #10242: File size restriction error message should tell what file is causing the issue

When performing a push, you may risk getting the "File exceeds GitHub's file size restriction (of 100MB)" error. It does not, in fact, mention what file it is.

### Proposed solution

It will benefit greatly to immediately know what file is causing the issue without having to perform further checks and could improve workflow even just a little bit.

For people with slow internet connections, it would benefit greatly if GitHub Desktop would check all file sizes before uploading files, to make sure no time is wasted upon the developer.

### Additional context

Image of the error message:
image

ADustyOldMuffin:
Closes #10242 

## Description

- This will put the offending file name in the error message 

### Screenshots

image

## Release notes

Notes: Added offending file name to the file exceeds size limit error.


ADustyOldMuffin:
So this is just a first stab (and IMO a kind of ugly one) but I was hoping to get feedback and maybe some direction on if this could be handled better.

I couldn't find a better place to get the offending file's name as the only place it was located was in the error messages returned.

I also know that this will not work with multiple files (see screenshot), but I wanted to wait on feedback before bothering with it since there might be an easier way to do it.

image

app/src/lib/helpers/regex.ts
```diff
@@ -51,3 +51,19 @@ export function getMatches(text: string, re: RegExp): Array<RegExpExecArray> {
 export function escapeRegExp(expression: string) {
   return expression.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&')
 }
+
+export function getFileFromExceedsError(error: string): string | null {
```
sergiou87: Could you add an example of error message in the comments of this function?

Also I think it'd be better if this returned a list of files, instead of an already formatted string.

app/src/lib/helpers/regex.ts
```diff
@@ -51,3 +51,19 @@ export function getMatches(text: string, re: RegExp): Array<RegExpExecArray> {
...
+  const endRegex = /(\sis\s\d+.\d+ [a-zA-Z]+;\sthis\sexceeds\sGitHub's\sfile\ssize\slimit\sof\s100.00\sMB)/gm
+  const beginRegex = /(^remote:\serror:\sFile\s)/gm
+  const beginMatch = beginRegex.exec(error)
+  const endMatch = endRegex.exec(error)
+
+  if (beginMatch && beginMatch.index && endMatch && endMatch.index) {
```
sergiou87: FYI nowadays we prefer specific checks for `null` / `undefined`:
```suggestion
  if (beginMatch !== null && beginMatch.index && endMatch !== null && endMatch.index) {
```

What are the `.index` checks for exactly? 🤔

sergiou87 (changes requested):
I like the idea! I think there is another error or something that just shows a list of files below the message, but I can't find it 🤔

app/src/lib/helpers/regex.ts
ADustyOldMuffin: I just really like conditional statements... 

I was using `matches` before instead of `exec` in which the index might be undefined. I'll update it while I make the other changes.

app/src/lib/helpers/regex.ts
ADustyOldMuffin: I added an example of the return, and also changed it to return an array of filenames

ADustyOldMuffin:
Current result, I tried to add some styling such as bold Files: but since it's just text I didn't have any luck. I use just newlines and tabs for moving the text around.

image

app/src/lib/git/core.ts
```diff
@@ -230,6 +231,17 @@ export async function git(
 
   log.error(errorMessage.join('\n'))
 
+  if (gitError && gitError === DugiteError.PushWithFileSizeExceedingLimit) {
```
sergiou87: You don't really need the `gitError && ` part here, right?
```suggestion
  if (gitError === DugiteError.PushWithFileSizeExceedingLimit) {
```

app/src/lib/git/core.ts
```diff
@@ -230,6 +231,17 @@ export async function git(
...
+    const result = getFileFromExceedsError(errorMessage.join())
+    if (result) {
+      const files = result.join('\n\t\t')
+
+      if (files && files !== '') {
+        gitResult.gitErrorDescription += '\n\nFiles:\n\t\t' + files
+      }
+    }
```
sergiou87: Since `result` is always an array, this can probably be simplified like this:
```suggestion
    const files = result.join('\n\t\t')

    if (files !== '') {
      gitResult.gitErrorDescription += '\n\nFiles:\n\t\t' + files
    }
```

sergiou87 (changes requested):
Thanks for addressing those issues! I have other 2 suggestions (sorry 😳 ).

Just food for thought (**not a requirement**), but I think it could be nice to also extract the size of those files from the error message, and show it like `MyBigFile.txt (235.00MB)`

sergiou87:
> I think there is another error or something that just shows a list of files below the message, but I can't find it 🤔

@ADustyOldMuffin I found it! It's the confirmation dialog that shows up when you select multiple files in the `Changes` tab, right-click them and select `Discard changes`

image

You will find it in `discard-changes-dialog.tsx`, and key is in the `PathText` component: https://github.com/desktop/desktop/blob/f09e9ec53e771aa90787ff8e8cb058bf6437f5bc/app/src/ui/discard-changes/discard-changes-dialog.tsx/#L128-L151

ADustyOldMuffin:
> > I think there is another error or something that just shows a list of files below the message, but I can't find it 🤔
> 
> @ADustyOldMuffin I found it! It's the confirmation dialog that shows up when you select multiple files in the `Changes` tab, right-click them and select `Discard changes`
> 
> image
> 
> You will find it in `discard-changes-dialog.tsx`, and key is in the `PathText` component: https://github.com/desktop/desktop/blob/f09e9ec53e771aa90787ff8e8cb058bf6437f5bc/app/src/ui/discard-changes/discard-changes-dialog.tsx/#L128-L151

Sorry for the super late reply! (Life) I'm curious on any tips as to how I might go about doing something along those lines since this is returned as just text. I'll continue to look into something of this nature.

ADustyOldMuffin:
Also in regards to your want of the file size I've added it, I also ran into some odd behavior in that the error message returned from git only ever returns 1 offending file (even if there is more than 1).

image

ADustyOldMuffin:
Since I took to long my candy is gone. I'll be making a new PR since I seem to have messed up this one.

PR #12554, opened 2021-07-02
Check invalidated tokens and suggest to sign in again

sergiou87:
## Description

With the changes in this PR, the app will now check for 401 errors coming from API requests. If those happen, the app will immediately sign out the user and warn them about this, while suggesting them to sign in again.

If they decided to sign in again, they'll be led directly to the sign in flow of the specific account whose token was invalidated (i.e. if the dotcom account was invalidated, they'll sign in that account directly, if the GHE account was invalidated, the account's endpoint will be used to sign in again in that GHE account).

### Screenshots

https://user-images.githubusercontent.com/1083228/124268896-74da9400-db3a-11eb-92eb-42b853ef1ef2.mov

## Release notes

Notes: [Improved] Detect when tokens are invalidated and suggest users to sign in again

### DIFF
--- a/app/src/ui/invalidated-token/invalidated-token.tsx
+++ b/app/src/ui/invalidated-token/invalidated-token.tsx
@@ -1,0 +1,62 @@
+import * as React from 'react'
+import { Dialog, DialogContent, DialogFooter } from '../dialog'
+import { Dispatcher } from '../dispatcher'
+import { Row } from '../lib/row'
+import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
+import { Account } from '../../models/account'
+import { getDotComAPIEndpoint } from '../../lib/api'
+
+interface IInvalidatedTokenResetProps {
+  readonly dispatcher: Dispatcher
+  readonly account: Account
+  readonly onDismissed: () => void
+}
+
+/**
+ * Dialog that alerts user that there are uncommitted changes in the working
+ * directory where they are gonna be resetting to a previous commit.
+ */
+export class InvalidatedToken extends React.Component<
+  IInvalidatedTokenResetProps
+> {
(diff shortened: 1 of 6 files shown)


billygriffin:
As discussed in our sync, this is a great improvement but we need to mitigate the problem where this dialog would replace the dialog that you're already in and therefore lose your state prior to merging this.

UpUpnAwaaaay:
Ok I just got back on and was reading your emails. I trust you on what needs dine. You are the pro. Lol. Thanks you so much for helping and replying

⁣Get TypeApp for Android ​

UpUpnAwaaaay (dismissed):
Thank goodness for you and your services are very helpful

sergiou87:
@billygriffin no worries! That was already addressed in https://github.com/desktop/desktop/pull/12554/commits/c4eb1a6f3f4d3e764a9b65259317043d42394600 😄

app/src/ui/invalidated-token/invalidated-token.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface IInvalidatedTokenResetProps {
+  readonly dispatcher: Dispatcher
+  readonly account: Account
+  readonly onDismissed: () => void
+}
+
+/**
+ * Dialog that alerts user that there are uncommitted changes in the working
+ * directory where they are gonna be resetting to a previous commit.
```
tidy-dev: Think this maybe some copy/pasta comments that should be updated. :)

tidy-dev (changes requested):
Looks and works great. Only nit is updating the copied comments on the new component. :)

tidy-dev (approved):
✨  Thanks for the update!

PR #12711, opened 2021-08-02
Switch deployment publish script to Azure

tidy-dev:
## Description

Swaps AWS storage upload functionality to Azure storage.

To note: Making the storage url, storage account name, and container as "secrets" was not required as they are public facing, but it is better practice to have them as environment variables since they are based on azure configurations that we could decide change (thus, now we wouldn't require a code change in that case). Same could be said for "releases", but as this is just a subfolder and not an azure configuration I let be.

## Release notes
Notes: no-notes

### DIFF
--- a/script/publish.ts
+++ b/script/publish.ts
@@ -25,7 +25,7 @@ if (!currentTipSHA.toUpperCase().startsWith(releaseSHA!.toUpperCase())) {
 
 import * as Fs from 'fs'
 import { execSync } from 'child_process'
-import * as AWS from 'aws-sdk'
+import * as Azure from 'azure-storage'
 import * as Crypto from 'crypto'
 import request from 'request'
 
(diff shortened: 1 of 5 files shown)


script/publish.ts
```diff
@@ -163,6 +165,37 @@ function upload(assetName: string, assetPath: string) {
...
+      publicAccessLevel: 'blob',
+    },
+    error => {
+      if (error !== undefined) {
+        console.error(
+          `Unable to ensure azure blob container - ${process.env.AZURE_BLOB_CONTAINER}`
+        )
+      }
+    }
+  )
```
sergiou87: Do we need to handle this asynchronously? According to https://azure.github.io/azure-storage-node/BlobService.html#createContainerIfNotExists__anchor the last parameter is a callback saying whether the result was successful or not.

From what I'm reading, sounds like we don't care about the result, just logging the error to give us a clue but not halting the publish process. Is that correct?

script/publish.ts
```diff
@@ -118,30 +118,25 @@ interface IUploadResult {
...
+  const azureBlobService = getAzureBlobService()
+
+  if (azureBlobService === null) {
+    throw new Error(
+      'Unable to obtain an azure blob service. Deployment aborting...'
+    )
   }
 
+  const container = process.env.AZURE_BLOB_CONTAINER || ''
+  const cleanAssetName = assetName.replace(/ /g, '')
```
sergiou87, line 123: Just curious, why not using `replaceAll` with a string instead of a regex?

```suggestion
  const cleanAssetName = assetName.replaceAll(' ', '')
```

script/publish.ts
```diff
@@ -151,9 +146,16 @@ function upload(assetName: string, assetPath: string) {
           hash.setEncoding('hex')
           const input = Fs.createReadStream(assetPath)
 
+          const uploadResult: IUploadResult = {
+            name: assetName,
+            url,
+            size: stats['size'],
+            sha: '',
+          }
```
sergiou87: Couldn't you put this inside the `hash.on('finish'…` callback? That way you don't need to re-set the `uploadResult.sha`.

sergiou87 (changes requested):
Overall LGTM! Only a few comments/suggestions 😄

script/publish.ts
tidy-dev: Yes. I wasn't super convinced that this was all that valuable to our solution (based my solution on some examples and they all  had it). Because really if this azure container didn't exist, it would fail on the upload anyways... and likely there would be bigger problems.. Would you prefer just to remove this error logging?

script/publish.ts
tidy-dev: Cause that is what was there already in the S3 implementation. :D Do you have a reason to want to change it?

script/publish.ts
tidy-dev: Yep.

script/publish.ts
sergiou87: Nope, I think the error is interesting! Maybe posting it as a warning?

And also a comment stating that we're intentionally not waiting for the operation to complete.

script/publish.ts
sergiou87: Ohh sorry, didn't see it. I just tend to avoid regex when possible: slower and more error prone. It's just a habit 😆 

Happy to keep it if it gives you peace of mind minimizing changes 🤣

script/publish.ts
tidy-dev: Decided since the next call would fail if this one did.. probably didn't make sense to go ahead with it. Refactored this to a promise so we don't bother attempting upload if there is no container to upload to.

sergiou87 (approved):
Looks great, thanks for all the changes!! 💖 :shipit:

PR #12714, opened 2021-08-03
Show number of lines changed in a commit

### LINKED ISSUE #11656: Add added and deleted count

When you have committed something, you will see the number of lines that have been changed (added and removed) on the site. I think it would be nice if you could also see this in the desktop app before you put this online.
I myself work with different JSONs and in this way I can easily check whether too much has been deleted.

### Proposed solution

More clarity about the changes in the files.

sergiou87:
Closes #11656

## Description

This PR adds the number of lines added/removed in a commit. This is only shown if there are line changes of any kind. If all files in a commit are binary files, those counters won't be shown at all.

I hid this behind a feature flag, in case having the extra `git log` causes any harm anywhere I haven't seen.

### Screenshots

image

## Release notes

Notes: [Added] See how many lines were added or deleted in a commit.

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -574,7 +574,7 @@ export interface ICommitSelection {
   /** The commits currently selected in the app */
   readonly shas: ReadonlyArray<string>
 
-  /** The list of files associated with the current commit */
+  /** The changeset data associated with the selected commit */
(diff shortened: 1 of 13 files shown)


tidy-dev (approved):
✨  Looks great!

PR #12734, opened 2021-08-05
Allow users to use Windows' OpenSSH binaries when available

### LINKED ISSUE #5641: Desktop does not use OpenSSH on Windows if running, favours embedded SSH
There seems to be no support for the Native Windows 10 Authorisation using the built-in ssh-agent.

Windows 10 now provides it's own built-in implementation of openssh. And it provides its own ssh-agent as a Windows Service for storing PassPhrases.

BUT I cannot get GitHub Desktop to work using the Keys setup with the Windows ssh-agent, so Pushing always fails to any private repositories on servers when using SSH.

When will the NATIVE Windows ssh-agent service be supported in GitHub Desktop application?

sergiou87:
Closes #5641

## Description

This PR introduces a new setting for users on Windows versions 10.0.17134 and newer, where there is a system OpenSSH available. New Windows installs will have this setting enabled by default, while existing installs will have this setting disabled in order to persist the existing behavior and avoid causing issues to users that already have everything set up and running.

Regarding the implementation, it just adds the `-c core.sshCommand="C:\\Windows\\System32\\OpenSSH\\ssh.exe"` parameter to all git calls.
I haven't managed to make this works on Windows 11 for ARM, it's like Desktop doesn't see that `ssh.exe`. I still have to test this on Windows 11 for x64 to see if it works there or if it's a Windows 11 issue.

**EDIT:** I just tested it and it doesn't work on ARM, but it does work on Windows 11 for x64 🎉 

### Screenshots

image

## Release notes

Notes: [Improved] Windows users can now rely on the system OpenSSH for their git repositories.

### DIFF
--- a/app/src/lib/ssh/ssh.ts
+++ b/app/src/lib/ssh/ssh.ts
@@ -1,0 +1,28 @@
+import * as fse from 'fs-extra'
+import memoizeOne from 'memoize-one'
+
+const WindowsOpenSSHPath = 'C:\\Windows\\System32\\OpenSSH\\ssh.exe'
+
+export const isWindowsSSHAvailable = memoizeOne(
+  async (): Promise<boolean> => {
+    if (!__WIN32__) {
+      return false
+    }
+
+    return await fse.pathExists(WindowsOpenSSHPath)
+  }
+)
+
+/**
+ * Returns the git arguments related to SSH depending on the current context
+ * (OS and user settings).
+ */
+export async function getSSHArguments() {
+  const useWindowsSSH = await isWindowsSSHAvailable()
+  if (!useWindowsSSH) {
+    return []
+  }
+
+  // Replace git sshCommand with Windows' OpenSSH executable path
+  return ['-c', `core.sshCommand="${WindowsOpenSSHPath}"`]
+}
--- a/app/src/lib/ssh/ssh.ts
+++ b/app/src/lib/ssh/ssh.ts
@@ -35,15 +35,17 @@ function isWindowsOpenSSHUseEnabled() {
 }
 
 /**
- * Returns the git arguments related to SSH depending on the current context
- * (OS and user settings).
+ * Returns the git environment variables related to SSH depending on the current
+ * context (OS and user settings).
  */
-export async function getSSHArguments() {
+export async function getSSHEnvironment() {
   const canUseWindowsSSH = await isWindowsOpenSSHAvailable()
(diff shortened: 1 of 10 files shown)


app/src/lib/ssh/ssh.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const UseWindowsOpenSSHKey: string = 'useWindowsOpenSSH'
+
+export const isWindowsOpenSSHAvailable = memoizeOne(
+  async (): Promise<boolean> => {
+    if (!__WIN32__ || !enableWindowsOpenSSH()) {
+      return false
+    }
+
+    return await fse.pathExists(WindowsOpenSSHPath)
```
sergiou87, line 22: For some reason this is `true` on Windows 11 for ARM, but later `git` can't find it 🤔

app/src/lib/stores/app-store.ts
```diff
@@ -484,6 +487,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     this.showWelcomeFlow = !hasShownWelcomeFlow()
 
+    // If the user never selected whether to use Windows OpenSSH or not, use it
+    // by default if we have to show the welcome flow (i.e. if it's a new install)
+    if (__WIN32__ && getBoolean(UseWindowsOpenSSHKey) === undefined) {
+      this._setUseWindowsOpenSSH(this.showWelcomeFlow)
+    }
```
sergiou87: This is what I used to check whether it's a first install or not: whether the welcome flow has been shown or not. If there is no stored value for this setting, set it to `true` only if we still have to show the welcome flow (i.e. if it's the first time the user opens the app).

ghost (commented):
Review please

sergiou87:
Just found some issues, I'll try another approach…

sergiou87:
Hopefully should be good now. I noticed some issues with the persisted setting not being read when the app starts 🤦‍♂️ 

I also noticed there is a `GIT_SSH_COMMAND` env variable that we can use instead of `core.sshCommand`, which should keep our logs cleaner. On the other hand, our logs won't tell us when git is using a custom ssh set by us, not sure if we should worry much at this point…

tidy-dev (approved):
✨  Looks good and tested well. 

_(Definitely doesn't show up in mac. 😄   I tried there first without thinking about it being windows only   🤦  )_

PR #12756, opened 2021-08-11
Improve SSH support with passphrase and host prompts

### LINKED ISSUE #2859: Support non-GitHub hosted git repositories connection with ppk
SourceTree supports connection with ppk

grabilla em4836

while GitHub desktop/windows will throw "invalid format":

```
Load key "/.ssh/id_rsa.ppk": invalid format
Permission denied (publickey,keyboard-interactive).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

(The error was parsed as 2: Authentication failed. You may not have permission to access the repository. Open options and verify that you're signed in with an account that has permission to access this repository.)
2017-09-19T19:12:02.433Z - info: [ui] Executing getStatus: git status --untracked-files=all --branch --porcelain=2 -z (took 1.164s)
2017-09-19T21:00:59.688Z - info: [ui] Executing fetch: git -c credential.helper= fetch --progress --prune origin (took 1.716s)
2017-09-19T21:00:59.692Z - error: [ui] `git -c credential.helper= fetch --progress --prune origin` exited with an unexpected code: 128.

Load key "/.ssh/id_rsa.pub": invalid format
Permission denied (publickey,keyboard-interactive).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

(The error was parsed as 2: Authentication failed. You may not have permission to access the repository. Open options and verify that you're signed in with an account that has permission to access this repository.)
2017-09-19T21:04:29.701Z - error: [ui] `git -c credential.helper= fetch --progress --prune origin` exited with an unexpected code: 128.
```

### LINKED ISSUE #3457: Github Desktop does not provide a UI to accept the public key of a new remote host
### Description

When using the Github Desktop (Windows) client

### Version

**GitHub Desktop version:** 1.0.9

**OS version:** Windows 10

### Example strings

HTTPS string: https://github.com/desktop/desktop.git
SSH string: git@github.com:desktop/desktop.git

### Notes

I've chosen to use Github.com as the example here, but the below applies to any host you try interact with (eg, a private gitlab instance) via the desktop client

### Steps to Reproduce

**0. Make a copy of your ~/.ssh/known_hosts file and delete it prior to testing the below issue**

1. Install Github Desktop client
2. Configure Github.com to accept your SSH public key (After following the SSH keygen process, or using a pregenerated key)
3. Add (clone) a repository using the SSH method instead of the HTTPS method (notes below regarding this, and see next step to deal with this behavior)
4. Reconfigure the repository through the UI via Repository>Repository settings to use the ssh remote origin (**Potentially unexpected behavior:**  Even when using the SSH string method the github desktop client will swap to using the HTTPS string, so you need to change it so it actually uses the SSH string)
5. Once you've confirmed the repository settings are using the SSH string (i.e. git@github.com:desktop/desktop.git) try and fetch origin (it will fail)

### Steps to "fix"

1. use a git shell to git pull on a repository, this will trigger the normal prompt:

```
The authenticity of host 'github.com (192.30.253.113)' can't be established.
RSA key fingerprint is SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8.
Are you sure you want to continue connecting (yes/no)? yes
```

2. This adds an entry to ~/.ssh/known_hosts and future interaction with that host will be OK.

**Expected behavior:** The UI should prompt to approve/deny the new remote SSH server key

**Actual behavior:** An Authentication error is spawned, with no actions available (or debugging information)

#### Additional Information

Whilst the github desktop client changes the repository settings (at least, for github.com repositories) to use HTTPS, if a user has specified the SSH string, the client should not automatically change this to be the HTTPS string (seems like unexpected behavior / a feature that could be misleading to the user)

For external / private repositories not hosted on github.com, the client doesn't seem to try change these to HTTPS strings.

### LINKED ISSUE #8761: GitHub desktop should be able to prompt for SSH key passphrase
When we followed the instructions for setting up SSH keys, that involves making a passphrase. Then when we started using Github desktop, it will not clone or pull the repos. It gives error message about authentication failure. It suggested logging in again and that didn't help. It seems like the desktop app does not prompt for SSH key passphrase when it should.

Then we created a new SSH key without passphrase and the problem was solved.

We know that there could be some Windows ssh-agent workarounds that we could be missing but it would be cool if the app just prompts for passphrase and do the right thing. 😄

sergiou87:
Closes #3457
Closes #8761

## Description

This PR sits on top of #12734 

The changes in this PR leverage the new TCP-based askpass trampoline from #11510, by providing that trampoline as the askpass program for SSH too.

That way, we can intercept when SSH prompts stuff to the user (like the SSH key passphrase or adding a host to the list of known hosts).

So far this is only available on Windows because macOS makes it a bit difficult to provide a custom SSH_ASKPASS program. That'll be tackled in a different PR.

### Screenshots

Note: the video is kind of a lie, since github.com is added as a host automatically. Users will be prompted about other hosts, though.

https://user-images.githubusercontent.com/1083228/129027589-def08dcc-e986-48aa-9c00-c15a2480848f.mov

## Release notes

Notes: [Fixed] Prompt SSH info to Windows users, like the key passphrase or adding a host to the list of known hosts

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -132,3 +132,8 @@ export function enableLineChangesInCommit(): boolean {
 export function enableWindowsOpenSSH(): boolean {
   return enableBetaFeatures()
 }
+
+/** Should we use SSH askpass? */
+export function enableSSHAskPass(): boolean {
+  return __WIN32__ && enableBetaFeatures()
+}
--- a/app/src/lib/file-system.ts
+++ b/app/src/lib/file-system.ts
@@ -117,7 +117,7 @@ export async function getFileHash(
   path: string,
   type: 'sha1' | 'sha256'
 ): Promise<string> {
-  return new Promise(resolve => {
+  return new Promise((resolve, reject) => {
     const hash = Crypto.createHash(type)
     hash.setEncoding('hex')
     const input = FSE.createReadStream(path)
@@ -125,6 +125,9 @@ export async function getFileHash(
     hash.on('finish', () => {
       resolve(hash.read() as string)
     })
+
+    input.on('error', reject)
+    hash.on('error', reject)
 
     input.pipe(hash)
   })
--- a/app/src/lib/file-system.ts
+++ b/app/src/lib/file-system.ts
@@ -4,6 +4,7 @@ import * as Path from 'path'
 import { Disposable } from 'event-kit'
 import { Tailer } from './tailer'
 import byline from 'byline'
+import * as Crypto from 'crypto'
 
 /**
  * Get a path to a temp file using the given name. Note that the file itself
@@ -109,5 +110,22 @@ export async function readPartialFile(
       })
       .on('error', reject)
       .on('end', () => resolve(Buffer.concat(chunks, total)))
+  })
+}
+
+export async function getFileHash(
+  path: string,
+  type: 'sha1' | 'sha256'
+): Promise<string> {
+  return new Promise(resolve => {
+    const hash = Crypto.createHash(type)
+    hash.setEncoding('hex')
+    const input = FSE.createReadStream(path)
+
+    hash.on('finish', () => {
+      resolve(hash.read() as string)
+    })
+
+    input.pipe(hash)
   })
 }
--- a/app/src/lib/git/authentication.ts
+++ b/app/src/lib/git/authentication.ts
@@ -8,13 +8,16 @@ import { TrampolineCommandIdentifier } from '../trampoline/trampoline-command'
 
 /** Get the environment for authenticating remote operations. */
 export function envForAuthentication(auth: IGitAccount | null): Object {
+  const askPassPath = enableDesktopTrampoline()
+    ? getDesktopTrampolinePath()
+    : getAskPassTrampolinePath()
+
   const env = {
     DESKTOP_PATH: process.execPath,
     DESKTOP_ASKPASS_SCRIPT: getAskPassScriptPath(),
     DESKTOP_TRAMPOLINE_IDENTIFIER: TrampolineCommandIdentifier.AskPass,
-    GIT_ASKPASS: enableDesktopTrampoline()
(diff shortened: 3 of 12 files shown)


app/src/lib/feature-flag.ts
```diff
@@ -132,3 +132,8 @@ export function enableLineChangesInCommit(): boolean {
 export function enableWindowsOpenSSH(): boolean {
   return enableBetaFeatures()
 }
+
+/** Should we use SSH askpass? */
+export function enableSSHAskPass(): boolean {
+  return __WIN32__ && enableBetaFeatures()
```
tidy-dev, line 138: Nit: Should the windows check be in the feature-flag? Just figured that once this goes in and someday we remove the feature flag.. will have to move the Windows check to inside the code somewhere.

tidy-dev (approved):
✨  Tested it out and it works great! On little nit that you can ignore if you like.

app/src/lib/feature-flag.ts
sergiou87: Yeah, I'm expecting to have something for macOS soon and remove that Win32 check, so it should be ok 🤞

the-real-nps:
I'm using the latest version (2.9.3) and this functionality isn't there: when I try to clone using SSH URL and I don't have an appropriate entry in the known_hosts file it ends with the "Authentication failed" error. If I have the appropriate entry in known_hosts then cloning works fine. But I have to add the entry by other means (e.g. "ssh <URL>" in the command line), Github Desktop doesn't ask if I want to add it as in the video.

ByteToHex:
Hi, would just like to chime in- as of 6 Feb 2025, Github Desktop encounters an infinite loop of 'cloning repository' when a SSH key with passphrase is used. Unsure if there are any workarounds- no prompts are appearing on my end.

I am on Windows.

PR #12749, opened 2021-08-10
4. Add choose branch functionality to new cherry-pick

tidy-dev:
## Description

Adds the choose branch (which includes create branch) functionality to new cherry-pick class (copied from cherry-pick flow but refactored to use multi commit operation state variables).

## Release notes
Notes: no-notes

PR #12757, opened 2021-08-11
5. Use multi commit op for cherry pick

tidy-dev:
Built on top of #12749

## Description

Refactors cherry-pick to use multi commit operation and removes cherry-pick specific logic/objects.

Scenarios tested:
- Cherry-pick drag and drop
- Cherry-pick context menu
- Cherry-pick multiple commits
- Cherry-pick to new branch context menu and drag and drop
- Cherry-pick with conflicts
    - Ensured the our's and their's branches make sense in manual conflict resolution dropdown.
- Cherry-pick with conflicts aborted
- Cherry-pick with conflicts started, close app, open app continue cherry-pick
- Cherry-pick undo
   - undoes cherry-pick
   - redirects to source branch
   - deletes target branch if created during cherry pick
  
## Release notes
Notes: no-notes


app/src/lib/stats/stats-store.ts
```diff
@@ -1670,6 +1671,7 @@ export class StatsStore implements IStatsStore {
       case MultiCommitOperationKind.Reorder:
         return this.recordReorderUndone()
       case MultiCommitOperationKind.CherryPick:
+        return this.recordCherryPickUndone()
```
sergiou87, line 1674: After this I assume `recordCherryPickUndone` can be made `private`?

app/src/lib/stats/stats-store.ts
```diff
@@ -1630,6 +1630,7 @@ export class StatsStore implements IStatsStore {
       case MultiCommitOperationKind.Reorder:
         return this.recordReorderSuccessful()
       case MultiCommitOperationKind.CherryPick:
+        return this.recordCherryPickSuccessful()
```
sergiou87, line 1633: Same `recordCherryPickSuccessful` can now be `private`, right?

app/src/lib/stores/app-store.ts
```diff
@@ -6215,21 +6150,36 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          },
+        }
       }
-    })
+    )
 
     this.emitUpdate()
   }
 
+  private getMCOProgressCallBack(repository: Repository) {
```
sergiou87: Not a fan of `MCO`, but I understand why 😂 

However keeping `MCO` could make reading the code a bit harder for someone who doesn't know about this? (poor Markus 🤣 )

app/src/lib/stores/app-store.ts
```diff
@@ -6404,56 +6327,110 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+
+    // are we already in the conflicts flow?
+    const alreadyInFlow =
+      this.currentPopup !== null &&
+      this.currentPopup.type === PopupType.MultiCommitOperation &&
+      multiCommitOperationState !== null &&
+      (multiCommitOperationState.step.kind ===
+        MultiCommitOperationStepKind.ShowConflicts ||
+        multiCommitOperationState.step.kind ===
+          MultiCommitOperationStepKind.ConfirmAbort)
```
sergiou87, line 6398: Personally I prefer checking the contents of an array instead: `[MultiCommitOperationStepKind.ShowConflicts, MultiCommitOperationStepKind.ConfirmAbort].includes(multiCommitOperationState.step.kind)` (maybe extracting that array into a constant first)

What do you think?

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -3509,57 +3456,74 @@ export class Dispatcher {
...
+          },
+        }
         break
-      case MultiCommitOperationKind.Rebase:
       case MultiCommitOperationKind.CherryPick:
+        const targetBranchName = targetBranch !== null ? targetBranch.name : ''
+        banner = {
+          type: BannerType.SuccessfulCherryPick,
+          targetBranchName,
+          countCherryPicked: count,
```
sergiou87: Maybe I'm missing something, but wouldn't be simpler to rename `countCherryPicked` to `count` and leave the old code here? 🤔

sergiou87 (changes requested):
LGTM overall… only have a few comments/suggestions before I test it

app/src/ui/dispatcher/dispatcher.ts
tidy-dev: Cherry-pick still has a different form because it requires the targetBranchName so I didn't see it worth bothering to change.

app/src/lib/stores/app-store.ts
tidy-dev: Yeah... I know.. I will switch it back. :D

app/src/lib/stats/stats-store.ts
tidy-dev: Good catch

app/src/lib/stores/app-store.ts
tidy-dev: :D - I actually considered just making a helper function for this annoyingly long check since it is used twice. Once in cherry-pick method and once in merge method. However, there are a couple of methods that say "cherryPick" in them in the app-store and maybe dispatcher that I think can be refactored out because we are now using the multi-commit-operation logic and this is one of them. That is what my next PR will be doing. Ok with going ahead with this for now since this code may be removed anyways (if it doesn't I can address it in next PR)?

sergiou87: SGTM!

app/src/ui/dispatcher/dispatcher.ts
sergiou87: Ouch, didn't notice that, sorry!

sergiou87 (dismissed):
Changes look good, tested as thoroughly as I could and all worked fine! :shipit:

sergiou87 (approved):

PR #12762, opened 2021-08-12
Refactor around trampoline environment variables

sergiou87:
## Description

This PR sits on top of #12756 

While working on #12756, I noticed how there are some scenarios where the ASKPASS env variables weren't set. We were only setting those for specific `git` commands, but some others could need it too.

This PR moves all the logic around setting those env variables to a more reasonable place, and then those are set for every git command. That meant I had to revert some of the changes from #11510.

I recommend reviewing this PR hiding the whitespace changes.

As a bonus, I also renamed the token storage key we use to persist SSH key passphrases to something less cryptic, as we have in https://github.com/desktop/desktop/blob/7665cf24c50e287493f52395fce6955fb8f38c56/app/src/lib/auth.ts#L9-L14

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/authentication.ts
+++ b/app/src/lib/git/authentication.ts
@@ -1,45 +1,21 @@
-import * as Path from 'path'
-
 import { GitError as DugiteError } from 'dugite'
 import { IGitAccount } from '../../models/git-account'
-import { enableDesktopTrampoline, enableSSHAskPass } from '../feature-flag'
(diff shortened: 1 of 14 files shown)


tidy-dev (approved):
✨  So pretty to see all the `withTrampolineEnvForRemoteOperation` get trimmed down to just passing in that env! ❤️

PR #12765, opened 2021-08-12
Workaround for Electron issue with setImmediate

sergiou87:
## Description

This PR includes a workaround for https://github.com/electron/electron/issues/29261 that basically replaces calls to `setImmediate` with `setTimeout(…, 0)`.

Given it will need some time in testing, I created a very original abstraction with `setAlmostImmediate` and wrote an eslint rule to make sure we don't use `setImmediate` (@shiftkey I'd love if you could take a look at it and share any feedback… but only if you have time!!!)

After testing it, I haven't noticed any issues, and of course I wasn't able to repro the aforementioned issue after bumping Electron to one of the affected versions, so we could say this temporary workaround is actually useful 😌 

## Release notes

Notes: no-notes

### DIFF
--- a/eslint-rules/set-almost-immediate.js
+++ b/eslint-rules/set-almost-immediate.js
@@ -35,7 +35,8 @@ module.exports = {
     const sourceCode = context.getSourceCode()
 
     /**
-     * Check each member of the interface body and ensure it is marked `readonly`.
+     * Check if a type annotation contains any references to NodeJS.Immediate
+     * and report them to be changed to AlmostImmediate.
      *
      * @param {TSTypeAnnotation} node
      * @param {TypeNode} typeAnnotation
--- a/eslint-rules/set-almost-immediate.js
+++ b/eslint-rules/set-almost-immediate.js
@@ -1,0 +1,105 @@
+// @ts-check
+
+/**
+ * set-almost-immediate
+ *
+ * This custom eslint rule is highly specific to GitHub Desktop and attempts
+ * to prevent using setImmediate since, under some circumstances, could not work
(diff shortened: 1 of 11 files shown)


eslint-rules/set-almost-immediate.js
```diff
@@ -0,0 +1,105 @@
...
+      nodeJSImmediateForbidden: `NodeJS.Immediate cannot be used, use AlmostImmediate instead`,
+    },
+    fixable: 'code',
+    schema: [],
+  },
+  create: function (context) {
+    const sourceCode = context.getSourceCode()
+
+    /**
+     * Check each member of the interface body and ensure it is marked `readonly`.
```
tidy-dev: Copy/Pasta comment

eslint-rules/set-almost-immediate.js
```diff
@@ -0,0 +1,105 @@
+// @ts-check
+
+/**
+ * set-almost-immediate
```
tidy-dev, line 4: These custom rules are neat!

tidy-dev (changes requested):
✨  Seems like a reasonable work around. I like the name. 😸  

One comment need updated from copy/paste of previous lint rule.

shiftkey:
@sergiou87 ESLint rules look fine to me :shipit:

eslint-rules/set-almost-immediate.js
sergiou87: Ouch! Fixed now, thanks 😄

tidy-dev (approved):
✨

PR #12777, opened 2021-08-16
Enhancing tag creation with autocompletion

### LINKED ISSUE #12509: Need create tag name more friendly
Hey.

 I think, when we create new tags name, need show all tag history like

image

it usefull way for create new tag with version or check duplicate tag or simple select new copy

mahezsh:
This PR Fixes #12509 

## Description

This shows the last 3 tags within the **Create Tag** component. 
It also filters the shown tags as the user type the tag name. (as suggested by @sergiou87)

### Screenshots

screen shot

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/create-tag/create-tag-dialog.tsx
+++ b/app/src/ui/create-tag/create-tag-dialog.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import _ from 'lodash'
 
 import { Repository } from '../../models/repository'
 import { Dispatcher } from '../dispatcher'
@@ -28,6 +29,7 @@ interface ICreateTagState {
    * shown in its place.
    */
   readonly isCreatingTag: boolean
+  readonly previousTags: Array<string> | null
 }
 
 const MaxTagNameLength = 245
@@ -43,12 +45,14 @@ export class CreateTag extends React.Component<
     this.state = {
       tagName: props.initialName || '',
       isCreatingTag: false,
+      previousTags: this.getExistingTagsFiltered(),
     }
   }
 
   public render() {
     const error = this.getCurrentError()
     const disabled = error !== null || this.state.tagName.length === 0
+    const lastThreeTags = _.takeRight(this.state.previousTags, 3)
 
     return (
       <Dialog
@@ -67,6 +71,17 @@ export class CreateTag extends React.Component<
             initialValue={this.props.initialName}
             onValueChange={this.updateTagName}
           />
+
+          {this.state.previousTags && (
+            <>
+              {lastThreeTags.length > 0 && <p>Previous Tags</p>}
+              {lastThreeTags.map((item: string, index: number) => (
+                <>
+                  <Ref key={index}>{item}</Ref>{' '}
+                </>
+              ))}
+            </>
+          )}
         </DialogContent>
 
         <DialogFooter>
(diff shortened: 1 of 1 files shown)


sergiou87:
Thank you @matrixnl!! We love it 😍 

The only thing that feels jarring is when there are no matches and the whole `Previous Tags` section just disappears. We think it might be nice to keep the `Previous Tags` section with something like `No matches found for '7'` or something similar.

(Aside from that, I still have to review the code itself, I'll give it a first pass tomorrow 😄 )

app/src/ui/create-tag/create-tag-dialog.tsx
```diff
@@ -67,6 +71,17 @@ export class CreateTag extends React.Component<
             initialValue={this.props.initialName}
             onValueChange={this.updateTagName}
           />
+
+          {this.state.previousTags && (
```
sergiou87: Nit, we try to avoid falsy/truthy checks
```suggestion
          {this.state.previousTags !== null && (
```

app/src/ui/create-tag/create-tag-dialog.tsx
```diff
@@ -99,9 +114,21 @@ export class CreateTag extends React.Component<
...
+    const previousTags = []
+    if (this.props.localTags) {
+      for (const item of this.props.localTags.keys()) {
+        previousTags.push(item)
+      }
+      return previousTags.filter(item => item.includes(filter))
+    }
+    return null
+  }
+
```
sergiou87: We prefer checks + early returns, and also I think you can create an array from `localTags.keys()` just using `Array.from`

```suggestion
  private getExistingTagsFiltered(filter: string = ''): Array<string> | null {
    if (this.props.localTags === null) {
      return null
    }
    const previousTags = Array.from(this.props.localTags.keys())
    return previousTags.filter(item => item.includes(filter))
  }

```

sergiou87 (changes requested):
Looks good overall! Only a few suggestions 😄

mahezsh:
Thanks @sergiou87, added **No matches found** and updated according to the suggestions 🙂

Screen Shot

sergiou87 (approved):
I love it, thank you!!! ❤️

app/src/ui/create-tag/create-tag-dialog.tsx
```diff
@@ -67,6 +71,23 @@ export class CreateTag extends React.Component<
             initialValue={this.props.initialName}
             onValueChange={this.updateTagName}
           />
+
+          {this.state.previousTags !== null && (
+            <>
+              <p>Previous Tags</p>
+              {lastThreeTags.length === 0 ? (
+                <>
+                  <Ref>{`No matches found for '${this.state.tagName}'`}</Ref>
```
sergiou87, line 80: Just noticed in your screenshot (thanks for sharing that) something I overlooked while reviewing it.

I think for this specific scenario, we don't want to use a `Ref`, just normal text 😅 

Sorry about that!

sergiou87 (changes requested):

Lulu-Lucinka (approved):

mahezsh:
Created a new PR for this issue at #12966. Have cherry-picked the first 2 commits and addressed @sergiou87's comment.

sergiou87:
Thanks, @mahezsh! I don't really understand why creating a new PR, but it's ok 😄  I guess we can close this one.

PR #12566, opened 2021-07-05
Added support for IntelliJ IDEA and IDEA CE

Cool-Katt:
Added support for IntelliJ IDEA and IDEA Community Edition as an external editor on Windows

Addresses  #9453

## Description

- Added support for IntelliJ IDEA and IDEA Community Edition as an external editor on Windows

## Release notes

Notes: Tried adding support for IntelliJ IDEA and IDEA Community Edition, versions 2020.2.4 & 2021.1.3, as external editors.

Issue #9453, opened 2020-04-04
Add support for JetBrains IDEs


van800:
There are multiple problems with current approach:

1. Each of ~15 JetBrains IDEs has to be manually added to ExternalEditor enum.
2. For Windows each new version of each JetBrains IDE requires listing its specific reg key in the list
https://github.com/desktop/desktop/blob/4c68dbc943d3c118a4ec6db883f1fb2f52314a17/app/src/lib/editors/win32.ts#L384
3. For Windows only HKEY_LOCAL_MACHINE installations are detected. HKEY_CURRENT_USER are not.
4. JetBrains Toolbox installations are not detected at all.
More details about detecting Toolbox installations https://github.com/JetBrains/resharper-unity/blob/net201/unity/EditorPlugin/RiderPathLocator.cs

related to https://github.com/desktop/desktop/issues/9365, https://github.com/desktop/desktop/issues/6077, https://github.com/desktop/desktop/issues/7195, https://github.com/desktop/desktop/issues/7065

steveward:
Request for IntellijIDEA in #9853.

steveward:
Requested again in #10101.

steveward:
From #11379:

>Nearly the whole family of JetBrains IDE flavors seems to be supported for integration/as editor - except CLion.
Is there a specific reason fro not supporting C/C++ IDE from JetBrains?

Cool-Katt:
Hi there,
Tried adding support for IntelliJ IDEA & IDEA Community edition in #12566. Waiting for approval to run workflows however and a review.

AmdrewAEddy:
> There are multiple problems with current approach:
> 
> 
> 
> 1. Each of ~15 JetBrains IDEs has to be manually added to ExternalEditor enum.
> 
> 2. For Windows each new version of each JetBrains IDE requires listing its specific reg key in the list
> 
> https://github.com/desktop/desktop/blob/4c68dbc943d3c118a4ec6db883f1fb2f52314a17/app/src/lib/editors/win32.ts#L384
> 
> 3. For Windows only HKEY_LOCAL_MACHINE installations are detected. HKEY_CURRENT_USER are not.
> 
> 4. JetBrains Toolbox installations are not detected at all.
> 
> More details about detecting Toolbox installations https://github.com/JetBrains/resharper-unity/blob/net201/unity/EditorPlugin/RiderPathLocator.cs
> 
> 
> 
[...]

PR #12778, opened 2021-08-16
Improved support for JetBrains IDEs

sergiou87:
## Description

This PR is part of #9453 and supersedes #12566.

It changes a bit the way we look for installations of JetBrains IDEs: while it still looks for specific versions, it generates those version numbers combining year and release number ([according to JetBrains](https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html?from=jetbrains.org#intellij-platform-based-products-of-recent-ide-versions)) in the last 2 years.

It might not be perfect, but should be much better than the current approach. I _think_ it doesn't cover yet JetBrains Toolbox.

Aside from the hacky approach, I can't tell if there is a performance penalty with this approach, since I could only test it in a VM.

## Release notes

Notes: [Improved] Added support for more versions of JetBrains IDEs

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -93,6 +93,7 @@ const registryKeysForJetBrainsIDE = (
     for (let release = 1; release <= maxReleasesPerYear; release++) {
       const key = `${product} ${year}.${release}`
       result.push(Wow64LocalMachineUninstallKey(key))
+      result.push(CurrentUserUninstallKey(key))
     }
   }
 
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨

app/src/lib/editors/win32.ts
```diff
@@ -32,13 +34,13 @@ interface IWindowsExternalEditor {
    * Some tools (like VSCode) may support a 64-bit or 32-bit version of the
    * tool - we should use whichever they have installed.
```
MelnikovAG, line 35: …

MelnikovAG: …

MelnikovAG (commented):
…

PR #12785, opened 2021-08-17
Refactor rebase to be a multi-commit-operation

tidy-dev:
## Description

Refactors Rebase to use multi commit operation. There are a few more broad refactors.
- made a base rebase class for rebase, squash, and reorder
- merge, merge squash, and rebase options in choose branch modal no longer needed endflow methods
- triggering conflicts when app is opened mid operation 

Basically all multi commit operations should be retested. :/

## Release notes

Notes: no-notes


app/src/lib/multi-commit-operation.ts
```diff
@@ -0,0 +1,57 @@
...
+import { IMultiCommitOperationState, IRepositoryState } from './app-state'
+
+/**
+ * Setup the multi commit operation state when the user needs to select a branch as the
+ * base for the operation.
+ */
+export function getMultiCommitOperationChooseBranchStep(
+  state: IRepositoryState,
+  initialBranch?: Branch | null
+) {
```
sergiou87: Maybe make the return value explicit?
```suggestion
): ChooseBranchStep {
```

app/src/lib/multi-commit-operation.ts
```diff
@@ -0,0 +1,57 @@
...
+  const initialState: ChooseBranchStep = {
+    kind: MultiCommitOperationStepKind.ChooseBranch,
+    defaultBranch,
+    currentBranch,
+    allBranches,
+    recentBranches,
+    initialBranch: initialBranch !== null ? initialBranch : undefined,
+  }
+
+  return initialState
```
sergiou87: If the return type is explicit as suggested in the comment above, we can now just return the object directly I think:
```suggestion
  return {
    kind: MultiCommitOperationStepKind.ChooseBranch,
    defaultBranch,
    currentBranch,
    allBranches,
    recentBranches,
    initialBranch: initialBranch !== null ? initialBranch : undefined,
  }
```

app/src/ui/multi-commit-operation/base-rebase.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const { operationDetail, originalBranchTip } = state
+
+    const targetBranchName =
+      conflictState !== null && isRebaseConflictState(conflictState)
+        ? conflictState.targetBranch
+        : null
+    const baseBranchName =
+      conflictState !== null && isRebaseConflictState(conflictState)
+        ? conflictState.baseBranch || ''
+        : ''
```
sergiou87: Just a note: I think this doesn't work yet ([but works in a newer version of TypeScript](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#cfa-aliased-conditions)), at some point we can extract part of the condition to a constant and TS will infer the types as expected :_)
```suggestion
    const isRebaseConflictState = conflictState !== null && isRebaseConflictState(conflictState)
    const targetBranchName =
      isRebaseConflictState
        ? conflictState.targetBranch
        : null
    const baseBranchName =
      isRebaseConflictState
        ? conflictState.baseBranch || ''
        : ''
```

app/src/models/multi-commit-operation.ts
```diff
@@ -216,3 +224,14 @@ export type MultiCommitOperationDetail =
   | ICherryPickDetails
   | IRebaseDetails
   | IMergeDetails
+
+export function instanceOfIBaseRebaseDetails(
+  object: any
+): object is IBaseInteractiveRebaseDetails {
+  return 'commits' in object && 'currentTip' in object
```
sergiou87: Not a fan of this check 😕 

This is not much better, but what if we declare a dummy object of type `IBaseInteractiveRebaseDetails` and check if `object` contains at least all the properties of that dummy object?

That way, the compiler will force us to include all the properties from `IBaseInteractiveRebaseDetails` in the dummy object in case we add/remove any:

```suggestion
  const dummy: IBaseInteractiveRebaseDetails = {
    commits: [],
    currentTip: 'main',
  }

  return Object.keys(dummy).every(key => key in object)
```

sergiou87 (commented):
On a first pass it looks good to me! I will give it another pass next week and then test it as much as possible, but in general feels like a nice cleanup :_)

Great work!

app/src/models/multi-commit-operation.ts
tidy-dev: Yeah. This is kind of `meh` but I like your much better. But to note, it is the `): object is IBaseInteractiveRebaseDetails {` part that makes typescript happy.  I could put `return true` in there for all typescript cares.

sergiou87: That's good to know, but it's better if the implementation matches the signature as closely as possible 😂

tidy-dev: Absolutely!

app/src/ui/multi-commit-operation/reorder.tsx
```diff
@@ -3,6 +3,7 @@ import { BaseRebase } from './base-rebase'
 
 export abstract class Reorder extends BaseRebase {
   protected conflictDialogOperationPrefix = 'reordering commits on'
+  protected rebaseKind = MultiCommitOperationKind.Reorder
```
tidy-dev, line 6: The reorder and squash "worked" without this, but in the conflict dialog after a rebase has been continued ones (2nd time hitting rebase dialog) you wouldn't see the "Reorder Commit" or "Squash Commit".

sergiou87:
Thanks for the cleanup!! I'm in the middle of testing it, and noticed this:

When a rebase is successful, it now says `Successfully rebased <branch> onto <branch>`, while before this PR it just said `Successfully rebased <branch>`. Is that expected?
image

Also, if I reopen the app while dealing with conflicts, after solving all the conflicts I get this instead:
image

sergiou87 (changes requested):
Tested all operations, only found 2 issues. One was a regression from an previous PR (already fixed in #12817) and the other was added as a comment, related to successful rebases after reopening the app in the middle of the rebase.

Great work!!!

tidy-dev:
TLDR: I have fixed the "open app in ongoing rebase state" success banner and the expected success banner behavior is to show both target and base branch if base branch is known, but prod has a bug.

I am using the same rebase success banner as in prod.
It's message is: 

```
const message =
    baseBranch !== undefined ? (
      <span>
        {'Successfully rebased '}
        <strong>{targetBranch}</strong>
        {' onto '}
        <strong>{baseBranch}</strong>
      </span>
    ) : (
      <span>
        {'Successfully rebased '}
        <strong>{targetBranch}</strong>
      </span>
    )

```
So the goal is to have a message as:
> Successfully rebased `targetBranch` onto `baseBranch`.

In prod today, it will do this if there are not conflicts. :D It should technically do it too if there are conflicts but I didn't dig into why it looses track of the baseBranch after conflicts. This is a current prod bug.

Thus when mine does say:
> Successfully rebased `targetBranch` onto `baseBranch`.

That is expected behavior. :D 

But... There is the other scenario you pointed out that is when a rebase is already started and you open the app (or close and reopen during). At this point we don't have a way to recover the base branch easily and current prod doesn't either. Thus the expected message is:

> Successfully rebased `targetBranch`

My code was not obtaining the target branch at this point nor was it correctly sending undefined for the base branch in the banner. (Corrected in latest commit)

sergiou87 (approved):
It's looking good now, thanks!! :shipit:

Issue #12767, opened 2021-08-13
Suggesting change to wording of squash warning box


hughlilly:
Hi,

I'd like to suggest that "force-push" (and "force-pushing") should be hyphenated here:
![image](https://user-images.githubusercontent.com/12046008/129293588-910bb8b6-11f4-4770-8564-c93c50aa29ee.png)

This wording is contained in two files: [app/src/ui/multi-commit-operation/warn-force-push-dialog.tsx](https://github.com/desktop/desktop/blob/18238077aa1e4851e710cfa66587cd4fd597bacd/app/src/ui/rebase/warn-force-push-dialog.tsx#L56-L59) and [app/src/ui/rebase/warn-force-push-dialog.tsx](https://github.com/desktop/desktop/blob/18238077aa1e4851e710cfa66587cd4fd597bacd/app/src/ui/rebase/warn-force-push-dialog.tsx#L56-L59).

Also, as you can see from the above screenshot, something in the code is causing an apostrophe to appear in the title of this dialog box.

Thanks
Hugh

Windows 10.0.19042.1052; Github Desktop 2.9.0 (x64)

tidy-dev:
@hughlilly Thank you for the suggestion - I opened a PR to correct the extra apostrophe. However, I did a quick google search of 'force-push' and 'force push' and  I believe adding the hyphenation is no more industry practice than having it without. In fact, I think I found more references to without it. Will close this issue when the apostrophe fix goes in. Thanks again!

li1r:
> 
> 
> Hi,
> 
> I'd like to suggest that "force-push" (and "force-pushing") should be hyphenated here:
> ![image](https://user-images.githubusercontent.com/12046008/129293588-910bb8b6-11f4-4770-8564-c93c50aa29ee.png)
> 
> This wording is contained in two files: [app/src/ui/multi-commit-operation/warn-force-push-dialog.tsx](https://github.com/desktop/desktop/blob/18238077aa1e4851e710cfa66587cd4fd597bacd/app/src/ui/rebase/warn-force-push-dialog.tsx#L56-L59) and [app/src/ui/rebase/warn-force-push-dialog.tsx](https://github.com/desktop/desktop/blob/18238077aa1e4851e710cfa66587cd4fd597bacd/app/src/ui/rebase/warn-force-push-dialog.tsx#L56-L59).
> 
> Also, as you can see from the above screenshot, something in the code is causing an apostrophe to appear in the title of this dialog box.
> 
> Thanks
[...]

PR #12792, opened 2021-08-17
Update warn-force-push-dialog.tsx

tidy-dev:
## Description

As pointed out in #12767,  we have an random apostrophe in this title. The PR removes it.

## Release notes
Notes: no-notes


billygriffin (approved):
Sad to lose the apostrophe but all good things come to an end.

sergiou87 (approved):

PR #12782, opened 2021-08-17
Enable SSH improvements on macOS

sergiou87:
## Description

This PR is a continuation of #12756 , where macOS support was left out since it required some additional work.

The important work to enable these improvements on macOS is in the desktop-trampoline package, more specifically in https://github.com/desktop/desktop-trampoline/pull/13

This PR just bumps `desktop-trampoline` to a version which includes the new `ssh-wrapper`, and makes sure the wrapper is copied in builds and git uses it on macOS.

### Screenshots

Note: the video is kind of a lie, since github.com is added as a host automatically. Users will be prompted about other hosts, though.

https://user-images.githubusercontent.com/1083228/129715245-47370730-8359-4273-b878-f26a14845c36.mov

## Release notes

Notes: [Fixed] Prompt SSH info to macOS users, like the key passphrase or adding a host to the list of known hosts

Issue #2579, opened 2017-08-29
Reduce the pain of consuming SSH remotes


shiftkey:
There's a collection of issues spread across multiple issues which touch on situations involving SSH remotes:

 - a user isn't clear on how to setup their machine to consume SSH repositories: #2453 #2847

> I looked around, and I cannot see anywhere, either in the desktop app settings, nor in the settings for any repository, nor in my GitHub profile settings, where I would go about doing SSH remote configuration.

- cloning a repo via SSH (and not defaulting back to https) - #2085 

 - setting up an SSH key and registering it on GitHub on behalf of a user - #2577

 - a repository with SSH submodules cannot be cloned because of a vague authentication failure: #2344 #2338

 - an SSH private may be protected with a passphrase - this can be handled by `ssh-agent` or PuTTY #2859

 - No way to trust unknown host keys - #3457 

- Using plink causes pushes to hang - #8765 

- No option for a default protocol (https://github.com/desktop/desktop/issues/11885)

- No support for hardware keys (https://github.com/desktop/desktop/issues/12216) 

- Desktop does not use OpenSSH on Windows if running, favours embedded SSH (#5641)

The classic apps addressed some of this (at least on the Windows side):

 - creating a new SSH key if not found and adding it to GitHub
 - ensuring `ssh-agent` was running when launching the app or Git Shell
 - setting the right environment variables when performing Git operations

joshaber:
The Classic Mac app did essentially the same as Windows, minus the `ssh-agent` stuff since that was unnecessary.

mcmonkey4eva:
Does current GitHub Desktop support SSH remotes at all?
I ran ssh-agent, did ssh-add on my keys (all on a random git bash instance) and it was all successful.
Github Desktop then failed to fetch or pull an SSH repo.
For reference, old Github-Windows client pulls the same fine.

Fair note that I've never use ssh-agent manually before, so I may have messed up somewhere.

So two things:

- Full support for making this easy and automated in the new Desktop client.
- How would I set it up under current limitations to where I can pull SSH repos in the new Desktop client?

shiftkey:
> How would I set it up under current limitations to where I can pull SSH repos in the new Desktop client?

[...]

PR #12800, opened 2021-08-18
SSH passphrase fixes and improvements

sergiou87:
## Description

This PR addresses an issue discovered in https://github.com/desktop/desktop/issues/2579#issuecomment-900938792

It sits on top of #12782 so that I could test it properly, because doing so on Windows would be very painful for me 😆 

Basically, the problem with the implementation prior to this PR is that we were storing the SSH key passphrase immediately after the user entered it, regardless of whether the git operation succeeded or not. Because of that, if the user entered the wrong passphrase, that would be stored immediately and never asked again, resulting in a loop of despair where Desktop would automagically read the (bad) passphrase from the Keychain and giving it to git to just fail miserably.

This problem was a bit challenging to fix, because the code handling the prompt (`trampoline-askpass-handler.ts`) and the code handling the git operation (`core.ts` or `spawn.ts`) are mostly disconnected, and we must be sure to store password A for operation X only when operation X succeeds, but not for operations Y or Z.

In order to solve this, I've leveraged the use of unique tokens for the trampoline: for the time a git operation is running, we create a unique token that the trampoline will use to make it harder to exploit for malicious purposes. That token, effectively, identifies that specific git operation uniquely. Thanks to that, we can do this:
1. When we get the prompt from the trampoline askpass handler, prompt the user for their SSH key passphrase.
2. If the user checked the `Remember passphrase` checkbox, keep the passphrase they entered **in memory**, associated to the unique trampoline token (and the hash of the SSH key file, of course).
3. Pass the passphrase to git so it can continue the operation.
4. Once the git operation finished, look for the passphrase associated to its trampoline token and store it in the keychain / credential manager.
5. Remove the passphrase from memory.

The main caveat of this approach is that if the user enters the right passphrase but git fails for whatever other reason, the passphrase won't be remembered and the user will need to enter it in future attempts. An alternative would be not storing the passphrase **only** if the error returned by the git operation was an authentication error. However, in that case, if the password was wrong but the git operation failed in some other way before noticing that (hence not reporting an authentication failure), the bad password might be stored in the keychain and would be imposible to fix unless the user manually deleted the stored password.

### Screenshots

The video below shows how I:
1. Enter the wrong password 3 times with `Remember passphrase` unchecked. And, of course, it doesn't remember the passphrase ✅ 
1. Enter the wrong password 3 times with `Remember passphrase` checked. And it still doesn't remember the passphrase ✅ 
1. Enter the right password with `Remember passphrase` unchecked. And, of course, it doesn't remember the passphrase but the command works ✅  
1. Enter the right password with `Remember passphrase` checked. It does remember the passphrase and the command works ✅  
1. I fetch again and it just works because it remembered the password 🎉 

https://user-images.githubusercontent.com/1083228/129895330-2fea89c0-e0f6-4399-87a6-d0716eed5f27.mp4

## Release notes

Notes: [Fixed] Wrong SSH key passphrases are not stored.

### DIFF
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -18,7 +18,6 @@ import { isErrnoException } from '../errno-exception'
 import { ChildProcess } from 'child_process'
 import { Readable } from 'stream'
 import split2 from 'split2'
-import { getSSHEnvironment } from '../ssh/ssh'
 import { merge } from '../merge'
 import { withTrampolineEnv } from '../trampoline/trampoline-environment'
 
@@ -160,9 +159,8 @@ export async function git(
     combineOutput(process.stdout)
   }
 
-  const result = await withTrampolineEnv(async env => {
-    const sshEnvironment = await getSSHEnvironment()
-    const combinedEnv = merge(opts.env, merge(env, sshEnvironment))
+  return withTrampolineEnv(async env => {
+    const combinedEnv = merge(opts.env, env)
 
     // Explicitly set TERM to 'dumb' so that if Desktop was launched
     // from a terminal or if the system environment variables
@@ -172,7 +170,7 @@ export async function git(
 
     const commandName = `${name}: git ${args.join(' ')}`
 
-    return GitPerf.measure(commandName, () =>
+    const result = await GitPerf.measure(commandName, () =>
       GitProcess.exec(args, path, opts)
     ).catch(err => {
       // If this is an exception thrown by Node.js (as opposed to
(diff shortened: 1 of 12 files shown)


app/src/lib/git/core.ts
```diff
@@ -160,9 +159,8 @@ export async function git(
     combineOutput(process.stdout)
   }
 
-  const result = await withTrampolineEnv(async env => {
-    const sshEnvironment = await getSSHEnvironment()
```
sergiou87, line 164: Since I added some SSH knowledge to the trampoline, I moved this SSH env variables to `withTrampolineEnv`, cleaning up the code a bit here and in `spawn.ts` 😄

app/src/lib/git/core.ts
```diff
@@ -184,64 +182,66 @@ export async function git(
 
       throw err
     })
-  })
 
-  const exitCode = result.exitCode
+    const exitCode = result.exitCode
```
sergiou87, line 186: All the code from this file below this line has just been indented. There is a very important (and long) comment in `trampoline-environment.ts` that explains why, but just a summary/spoiler: we want `withTrampolineEnv` to get any result returned by `git`/`spawnAndComplete`, and catch any error thrown by them.

That way, `withTrampolineEnv` can tell when the git operation succeeded or not and, according to that, store any SSH key passphrase associated to the git operation or not.

app/src/lib/git/spawn.ts
```diff
@@ -39,13 +38,7 @@ export async function spawnAndComplete(
       commandName,
       () =>
         new Promise<ProcessOutput>(async (resolve, reject) => {
-          const sshEnv = await getSSHEnvironment()
```
sergiou87, line 42: Same cleanup I mentioned in `core.ts` 😍

app/src/lib/ssh/ssh-key-passphrase.ts
```diff
@@ -0,0 +1,78 @@
...
+/** Retrieves the passphrase for the SSH key in the given path. */
+export async function getSSHKeyPassphrase(keyPath: string) {
+  try {
+    const fileHash = await getHashForSSHKey(keyPath)
+    return TokenStore.getItem(SSHKeyPassphraseTokenStoreKey, fileHash)
+  } catch (e) {
+    log.error('Could not retrieve passphrase for SSH key:', e)
+    return null
+  }
+}
```
sergiou87, line 20: These are the same functions that previously lived in `ssh.ts`, just brought them here to live with other passphrase-related functions that I needed to add.

app/src/lib/ssh/ssh-key-passphrase.ts
```diff
@@ -0,0 +1,78 @@
...
+  keyPath: string,
+  passphrase: string
+) {
+  try {
+    const keyHash = await getHashForSSHKey(keyPath)
+    SSHKeyPassphrasesToStore.set(operationGUID, { keyHash, passphrase })
+  } catch (e) {
+    log.error('Could not store passphrase for SSH key:', e)
+  }
+}
```
sergiou87, line 59: This chunk is important!

app/src/lib/ssh/ssh.ts
```diff
@@ -73,38 +71,3 @@ export async function getSSHEnvironment() {
...
-    const fileHash = await getHashForSSHKey(keyPath)
-    await TokenStore.setItem(
-      SSHKeyPassphraseTokenStoreKey,
-      fileHash,
-      passphrase
-    )
-  } catch (e) {
-    log.error('Could not store passphrase for SSH key:', e)
-  }
-}
```
sergiou87, line 110: Just stuff I moved to `ssh-key-passphrase.ts` (except for `storeSSHKeyPassphrase`, which was removed).

app/src/lib/trampoline/trampoline-askpass-handler.ts
```diff
@@ -62,7 +66,14 @@ async function handleSSHKeyPassphrase(
...
   }
 
-  const passphrase = await trampolineUIHelper.promptSSHKeyPassphrase(keyPath)
+  const {
+    passphrase,
+    storePassphrase,
+  } = await trampolineUIHelper.promptSSHKeyPassphrase(keyPath)
+
+  if (passphrase !== undefined && storePassphrase) {
+    keepSSHKeyPassphraseToStore(operationGUID, keyPath, passphrase)
```
sergiou87, line 75: Another important bit. Now the UI prompt we show to the user just tells us if the user wanted to remember the passphrase, and if so we just "keep it around" in memory, associated to the `operationGUID` (which is just the trampoline token).

app/src/lib/trampoline/trampoline-command-parser.ts
```diff
@@ -126,8 +126,17 @@ export class TrampolineCommandParser {
...
+      'DESKTOP_TRAMPOLINE_TOKEN'
+    )
+
+    if (trampolineToken === undefined) {
+      throw new Error(`The trampoline token is missing`)
+    }
+
     return {
       identifier,
+      trampolineToken,
```
sergiou87, line 139: I included the trampoline token to the trampoline commands because it was very convenient.

app/src/lib/trampoline/trampoline-environment.ts
```diff
@@ -21,18 +28,40 @@ export async function withTrampolineEnv<T>(
...
+    // The code below assumes a few things in order to manage SSH key passphrases
+    // correctly:
+    // 1. `withTrampolineEnv` is only used in the functions `git` (core.ts) and
+    //    `spawnAndComplete` (spawn.ts)
+    // 2. Those two functions always thrown an error when something went wrong,
+    //    and just return a result when everything went fine.
+    //
+    // With those two premises in mind, we can safely assume that right after
+    // `fn` has been invoked, we can store the SSH key passphrase for this git
+    // operation if there was one pending to be stored.
```
sergiou87, line 43: This is very very important!!!

app/src/lib/trampoline/trampoline-environment.ts
```diff
@@ -21,18 +28,40 @@ export async function withTrampolineEnv<T>(
...
+        DESKTOP_TRAMPOLINE_IDENTIFIER: TrampolineCommandIdentifier.AskPass,
+
+        ...sshEnv,
+
+        // Env variables specific to the old askpass trampoline
+        DESKTOP_PATH: process.execPath,
+        DESKTOP_ASKPASS_SCRIPT: getAskPassScriptPath(),
+      })
+
+      await storePendingSSHKeyPassphrase(token)
```
sergiou87, line 58: If the app got to this point, it meant the git operation didn't fail, and therefore we can safely store any pending SSH key passphrase associated to it 😌

tidy-dev (approved):
✨  Makes sense - For my own summary, you are saying "User gave me passphrase and asked me to store it. Let's try it! If it works, store it. If not, delete it so we can ask for a new one."

PR #12804, opened 2021-08-19
Remove pending SSH Key passphrases when users don't want to remember them

sergiou87:
## Description

This PR fixes a more convoluted scenario where a user could end up with a wrong SSH key passphrase stored in the Keychain / Credential Manager. The problem was that the logic in #12800 did not account for these multiple attempts properly, so what happened was…
1. The user tries to perform a git operation requiring SSH key passphrase
2. The user enters the wrong passphrase **and** checks `Remember the passphrase`. **The app will keep around this wrong password to store it in the Keychain if the git command succeeds**
3. Since the passphrase was wrong, the app prompts the passphrase again.
4. The user enters the right passphrase **but** does **NOT** check `Remember the passphrase`.
5. The git command succeeds.
6. **The app checks if we kept any passphrases around to save when this command succeeded…** so the wrong passphrase entered in (2) is remembered 🤦 🤦 🤦 🤦 🤦 

What this PR changes is step (4), by also respecting the **not** to `Remember the passphrase`. When that checkbox is not checked, the app will explicitly delete any passphrases that are kept around to remember if this git operation succeeded.

## Release notes

Notes: [Fixed] Wrong SSH key passphrases are not stored after multiple failed attempts and then one successful

### DIFF
--- a/app/src/lib/trampoline/trampoline-askpass-handler.ts
+++ b/app/src/lib/trampoline/trampoline-askpass-handler.ts
@@ -2,6 +2,7 @@ import { getKeyForEndpoint } from '../auth'
 import {
   getSSHKeyPassphrase,
   keepSSHKeyPassphraseToStore,
+  removePendingSSHKeyPassphraseToStore,
 } from '../ssh/ssh-key-passphrase'
 import { TokenStore } from '../stores'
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨ Works great!

PR #12833, opened 2021-08-24
Customizable high contrast theme

tidy-dev:
## Description

This started as a hack day project with two goals. 
1) Providing a high-contrast theme for users who have vision impairments or otherwise prefer high contrast screens.
2) Provide a customizable theme. 
-- A lot of times high-contrast themes are customizable as people who benefit from them sometimes need to adjust them to be optimal for their setup. 
-- Just to provide customizable themes for those who prefer it. 

However... with customizability brings ability for a user to reduce legibility to the point of making the app unusable. Thus, the high contrast theme and the customizability of it are independently feature flagged, so we can at least provide a more accessible app by providing a non-customizable high-contrast if we determine the customizable option to be too unwieldy.

Other input received that could be added:
- Adding stroke/shadow/changed color or some form of contrast effect to icons so they show regardless of customized background color. Here is example of poor contrast of the modified icon.
image
- Updating the diff js code to produce start and end classes on addition and deletion chunks so we can add a border around entire sections of additions and deletions instead of individual lines.
- Detect if a users choose a background and text color that do not contrast with each other and provide a warning with a quick fix button.
- The color picker could be made to remain open to allow a user to "look" for a color and then have a means to confirm that selection so that it doesn't just dismiss as soon as you click it. Also, it doesn't dismiss if you click outside it. Also, this color picker library has other color pickers.. maybe there is a more ideal one - designer input/review?
- The color swatches and color picker styles could be prettier - maybe get a designer to review?
- Add Statistics... how many users use high-contrast/how many users customized their theme. I would expect < 1% of users, but they are users who are hindered by the lack of a high-contrast option. 

Random other thoughts:
- Since this is a json object of colors... we could make customized themes shareable. 😄 

### Screenshots
image

image

## Release notes

Notes: [Added] Users can choose a customizable high-contrast appearance option.


app/src/lib/feature-flag.ts
```diff
@@ -123,6 +123,16 @@ export function enableLineChangesInCommit(): boolean {
...
+/** Should we allow high contrast theme option */
+export function enableHighContrastTheme(): boolean {
+  return true
+}
+
+/** Should we allow customizing a theme */
+export function enableCustomizeTheme(): boolean {
+  return true
+}
+
```
sergiou87: Shouldn't these be enabled only for beta/dev builds? 😳

app/src/ui/app-theme.tsx
```diff
@@ -44,19 +58,128 @@ export class AppTheme extends React.PureComponent<IAppThemeProps> {
...
+    )
+    body.appendChild(styles)
   }
 
   private clearThemes() {
     const body = document.body
 
-    for (const className of body.classList) {
+    const classList = [...body.classList]
+    for (const className of classList) {
```
sergiou87: Why is this change needed? 🤔

app/src/lib/stores/app-store.ts
```diff
@@ -5864,6 +5871,20 @@ export class AppStore extends TypedBaseStore<IAppState> {
   public _setSelectedTheme(theme: ApplicationTheme) {
     setPersistedTheme(theme)
     this.selectedTheme = theme
+    if (theme === ApplicationTheme.HighContrast) {
+      this.currentTheme = theme
+    }
+    this.emitUpdate()
+
+    return Promise.resolve()
```
sergiou87, line 5879: Not something for this PR, just food for thought, but I see the logic of the current/selected/applied/custom/whatever theme getting more and more complicated 😂  Maybe we should consider at some point encapsulating all this stuff in its own class/module, out of the app store…

app/src/ui/app-theme.tsx
```diff
@@ -44,19 +58,128 @@ export class AppTheme extends React.PureComponent<IAppThemeProps> {
...
+        --diff-delete-border: 1px solid crimson;
+        --tab-bar-hover-border: 2px solid ${border} !important;
+        --tab-bar-item-border: 2px solid ${background};
+        --foldout-border: 1px solid ${border};
+        --horizontal-bar-active-color: ${activeItem};
+        --horizontal-bar-active-text-color: ${activeText};
+    `
+
+    styles.appendChild(
+      document.createTextNode(
```
sergiou87: TIL you can add custom CSS rules like this 😄

sergiou87 (commented):
Looks great overall!! Just have a few comments, questions 😄

app/src/lib/feature-flag.ts
tidy-dev: :D - Woops... Yes. I had to set them for true to put out a test build.

app/src/ui/app-theme.tsx
tidy-dev: This is one of those .. I don't know why it didn't work ... it only worked for our app before because we only had `-dark` or `-light`.. now we have (`-dark` or `-light`) and `-custom` (two theme classes at a time) and I found that when switching between dark and light and high contrast that it would only remove one of those classes..

My theory is that `body.classList` is not a regular array (it some html dom construct) and the iterator didn't work like a normal (again no idea why) array.  Using the `[...body.classList]` forces it into a "normal" array... and therefore, it will actually clear all the themes.  Other thoughts.. when using `for (const className of body.classList)` maybe the classList is being changed on each iteration and causing a fluke..

app/src/ui/app-theme.tsx
tidy-dev: To note... This feels a bit hacky to me because you lose the context of the css syntax. However, from what research I did, it seemed like about the only way to go about programmatically changing theme variables from js. 

In fact.. now that I think about it, it may be nice to pull the meat of this style generating code out into a library function that outputs the stylesheet string so we can run "lint type checking" unit tests against it.

app/src/ui/app-theme.tsx
sergiou87: 🤯

sergiou87:
@tidy-dev I tested it and looks good! But it feels weird that in the color selector any click dismisses the popup, so I can't explore different colors from the sliders and so on. You can't easily see it in the video, but I'm trying to drag while clicking in order to see different colors:

https://user-images.githubusercontent.com/1083228/131841470-3474973b-8e11-4f9c-ba55-f3cd9777889e.mov

tidy-dev:
@sergiou87 Yep! Completely agree. :D That is one of my bullet points up top of things that could be added/made better... it definitely annoys me that was the default behavior.

sergiou87:
LOL sorry I read it the first time I reviewed it but I forgot about it today when I finally tested 😂

sergiou87 (approved):
:shipit: ✨

PR #12834, opened 2021-08-24
Fix (or vastly improve) performance issue /w dynamic gutter lines

### LINKED ISSUE #12830: 2.9.1 causes large diffs to flicker and lag
### Describe the bug
After upgrading from 2.9.0 to 2.9.1, the diff pane flickers or stutters while rendering.

When scrolling, most of the diff view is replaced by the window background color as it attempts to render during scroll. Worse yet, it now scrolls very slowly because of the lag... what was once a single swipe between pages, becomes several swipes moving 20 lines at a time.

Could it be due to the dynamic gutter width change? #12071

### Version & OS

GitHub Desktop Version 2.9.1 (x64)
macOS Big Sur 11.5.2 (20G95)

Also occurs on Windows: #12831

- Click a file with a large diff (300+ lines)
- Scroll the diff view up and down

I had to split this lorem ipsum into one line per sentence because otherwise, a diff of the same size but only containing a dozen or fewer changed lines has no issue, suggesting the variable that changes the render speed is the number of lines with changes, not the vertical height of the content.

### Expected behavior

It used to be very fast and fluid with no flicker in 2.9.0.

### Actual behavior

Render issues cause the scroll to stutter and the content to be replaced by the window background color.

### Screenshots

https://user-images.githubusercontent.com/704098/130614300-782ca662-da0e-4319-82c2-041033e26c93.mov

tidy-dev:
Closes #12830

## Description

In 2.9.1, we released an improvement on the gutter diff displays that dynamically set the width of the line number gutters to prevent clipping - PR #12071.

Unfortunately, it was found to cause performance issues for large diffs. This PR aims to fix or vastly improve that performance issue by only applying the dynamic styles for the gutter lines when they change (not continuously as the user scrolls).

## Release notes
Notes: [Fixed] Fix scrolling performance for large diffs.

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -1016,10 +1016,15 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
 
     const gutterParentElement = cm.getGutterElement()
     const gutterElement = gutterParentElement.getElementsByClassName(
-      'diff-gutter'
+      diffGutterName
     )[0]
-    gutterElement.setAttribute('style', `width: ${diffSize * 2}px;`)
-    cm.refresh()
+
+    const newStyle = `width: ${diffSize * 2}px;`
+    const currentStyle = gutterElement.getAttribute('style')
+    if (newStyle !== currentStyle) {
+      gutterElement.setAttribute('style', newStyle)
+      cm.refresh()
+    }
   }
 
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
LGTM, tested and worked nicely! :shipit: 

Just a nit, the release note should say `[Fixed]` instead of `[Fix]`

PR #12863, opened 2021-08-30
Fix Notepad++ and RStudio integration on Windows

### LINKED ISSUE #12841: Notepad++ External Editor support regression
### Describe the bug
I no longer have the option to open files in Notepad++, which I had configured as my External Editor.
GitHub Desktop doesn't find any suitable editors and prompts me to install VS Code.

### Version & OS
GitHub Desktop version 2.92.
Microsoft Windows [Version 10.0.19041.1165]

1. Go to the Changes tab.
2. Check the main pane (on the right).

### Expected behavior

The main pane should show an option to "Open the repository in your external editor", just above the option to "View the files of your repository in Explorer".

### Actual behavior

Only the Explorer option is shown.

Also, the File, Options, Integrations dialog now says "External editor.  No editors found.  Install Visual Studio Code?"
but it should offer a way to choose Notepad++.

### Additional context

I've observed this on two different computers, both Windows 10.
Notepad++ support was working well in version 2.90 as shown in the screenshot.

### Screenshots

ghd290-nppp

sergiou87:
Closes #12841

## Description

The regression was introduced in #12778, where the `executableShimPath` changed from a list of path components, to a list of lists of path components (meaning, there could be multiple executable paths to check): RStudio and Notepad++ don't need specific paths because for those we just check the `DisplayIcon` entry in the registry, and just specifying `[]` broke the logic we have for those, because that meant there weren't executable files for those editors. Instead, we needed to use `[[]]` (meaning, there is only 1 executable path which is whatever was in the `DisplayIcon`).

## Release notes

Notes: [Fixed] Fix integration of Notepad++ and RStudio on Windows

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -298,7 +298,7 @@ const editors: IWindowsExternalEditor[] = [
       // 32-bit version of Notepad++
       Wow64LocalMachineUninstallKey('Notepad++'),
     ],
-    executableShimPaths: [],
+    executableShimPaths: [[]],
     installLocationRegistryKey: 'DisplayIcon',
     expectedInstallationChecker: (displayName, publisher) =>
       displayName.startsWith('Notepad++') && publisher === 'Notepad++ Team',
@@ -314,7 +314,7 @@ const editors: IWindowsExternalEditor[] = [
   {
     name: 'RStudio',
     registryKeys: [Wow64LocalMachineUninstallKey('RStudio')],
-    executableShimPaths: [],
+    executableShimPaths: [[]],
     installLocationRegistryKey: 'DisplayIcon',
     expectedInstallationChecker: (displayName, publisher) =>
       displayName === 'RStudio' && publisher === 'RStudio',


niik (approved):
This looks like it would work just fine. I do think that it'd be best if we altered the code such that an empty array didn't trigger this issue since it seems quite likely we'll make this error in the future. We could even make the executableShimPaths optional.

Perhaps something like

```ts
    const shimPaths = editor.executableShimPaths !== undefined && editor.executableShimPaths.length > 0
      ? editor.executableShimPaths
      : [[]]

    for (const executableShimPath of shimPaths) {
```

Either way this seems like a fine fix for the problem at hand.

sergiou87:
Yup! That definitely crossed my mind, but decided to go with the simpler fix for now. My intention (today) was to also write some tests to automatically check, every time we add a new entry, that we haven't broken anything basic. And then use some TypeScript magic to enforce proper editor entries 🤔 

However, my initial approach became more messy than I anticipated, so I decided to go ahead with this 😂  I'll try to have some initial tests soon…

PR #12872, opened 2021-08-31
Tidy up log messages around scanning for installed editors

niik:
## Description

Noticed these log messages in my console while debugging and they struck me as both unnecessary and uninformative

image

The `[Object object]` part here used to say which editor wasn't being found and the log messages were a lot less annoying when there were only a few editors.

Note that this only affects debug log messages (which aren't persisted to disk) and only on macOS.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -131,9 +131,20 @@ async function findApplication(
 ): Promise<string | null> {
   for (const identifier of editor.bundleIdentifiers) {
     try {
-      const installPath = await appPath(identifier)
-      const exists = await pathExists(installPath)
-      if (exists) {
+      // app-path not finding the app isn't an error, it just means the
+      // bundle isn't registered on the machine.
+      // https://github.com/sindresorhus/app-path/blob/0e776d4e132676976b4a64e09b5e5a4c6e99fcba/index.js#L7-L13
+      const installPath = await appPath(identifier).catch(e =>
+        e.message === "Couldn't find the app"
+          ? Promise.resolve(null)
+          : Promise.reject(e)
+      )
+
+      if (installPath === null) {
+        return null
+      }
+
+      if (await pathExists(installPath)) {
         return installPath
       }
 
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨  Nice title. 😄

app/src/lib/editors/darwin.ts
```diff
@@ -131,15 +131,28 @@ async function findApplication(
...
   for (const identifier of editor.bundleIdentifiers) {
     try {
-      const installPath = await appPath(identifier)
-      const exists = await pathExists(installPath)
-      if (exists) {
+      // app-path not finding the app isn't an error, it just means the
+      // bundle isn't registered on the machine.
+      // https://github.com/sindresorhus/app-path/blob/0e776d4e132676976b4a64e09b5e5a4c6e99fcba/index.js#L7-L13
+      const installPath = await appPath(identifier).catch(e =>
+        e.message === "Couldn't find the app"
```
sergiou87, line 138: Looking at the source code, wouldn't be better to check `e.code === 2`? If I'm not wrong, that 2 comes from `ENOENT` so you can probably even use the `os.constants.errno.ENOENT` constant from Node.

I want to think that would be a bit less magic and a bit less coupled to the implementation of app-path than checking the "user-facing" string 😳  What do you think?

sergiou87 (commented):
Nice cleanup, thanks! I just learned why I've never seen these logs myself: I always use the default levels in Dev Tools 🤣 

Only left one suggestion

app/src/lib/editors/darwin.ts
niik: Although ENOENT is likely what inspired it, it seems the return value from the app-path executable doesn't use the constant and is just as magic. See https://github.com/sindresorhus/app-path/blob/0e776d4e132676976b4a64e09b5e5a4c6e99fcba/main.m#L21-L23

So I think either option is tightly coupled with the implementation of app-path although matching the string is perhaps more readable? I'm like 2 out of 10 on the care-scale here though so I'm happy to switch to checking `code` instead if you'd like but I'm not sure using the constant variable is merited, what do you think?

sergiou87: Ohhh that's sad… I'm ok with leaving it as it is!

sergiou87 (approved):
:shipit:

PR #12874, opened 2021-08-31
Windows editor integration improvements

sergiou87:
## Description

The reason behind this refactor is to make our editor integration for Windows a bit less error-prone and also (in future commits) cover it with some basic automated tests.

Within the refactor there are two main changes:
1. The `executableShimPaths` is now an optional value. Its requirement will depend on the value of `installLocationRegistryKey`, which will be discriminant property in this tagged union. Thanks to this, when `installLocationRegistryKey` is `DisplayIcon`, the property `executableShimPaths` must **not** be provided.
2. I removed the `ExpectedInstallationChecker` and just replaced it with the `displayNamePrefix` and `publisher`. The main change is that for some editors we currently check the whole `DisplayName`, and with this new implementation we would always check its prefix. However, given how we look for specific registry keys and how we check the publisher to be the same as in the registry, it should be safe (and simpler/cleaner).

What do you think?

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -23,8 +23,30 @@ type ExpectedInstallationChecker = (
 
 type RegistryKey = { key: HKEY; subKey: string }
 
+type WindowsExternalEditorPathInfo =
+  | {
+      /**
+       * Registry key with the install location of the app. If not provided,
+       * 'InstallLocation' will be used.
+       **/
+      readonly installLocationRegistryKey?: 'InstallLocation'
+
+      /**
+       * List of lists of path components from the editor's installation folder to
+       * the potential executable shims. Only needed when the install location
+       * registry key is `InstallLocation`.
+       **/
+      readonly executableShimPaths: ReadonlyArray<ReadonlyArray<string>>
+    }
+  | {
+      /**
+       * Registry key with the install location of the app.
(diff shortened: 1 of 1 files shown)


app/src/lib/editors/win32.ts
```diff
@@ -16,15 +16,32 @@ interface IWindowsAppInformation {
...
+       * registry key is `InstallLocation`.
+       **/
+      readonly executableShimPaths: ReadonlyArray<ReadonlyArray<string>>
+    }
+  | {
+      /**
+       * Registry key with the install location of the app.
+       **/
+      readonly installLocationRegistryKey: 'DisplayIcon'
+    }
```
sergiou87, line 41: This addresses the issue mentioned in https://github.com/desktop/desktop/pull/12863#pullrequestreview-741609918

Maybe it's a bit overkill? But I think it's more honest about the problem at hand… 🤔

niik (approved):
I'm into this, the needlessly nervous part of me says that we might be constraining ourselves too much here but we can always adapt should some odd editor come along that needs a more customizable approach.

One thing that gave me pause was the double-use of the discriminating property. I might have opted for a more verbose approach of using a `kind` property to discriminate. If only for the reason that it would follow a common pattern.

PR #12882, opened 2021-09-01
Small fixes/improvements in diff expansion

sergiou87:
## Description

This PR makes two changes in diff expansion to be consistent with dotcom's experience:
1. Use the right icon in short diff hunks.
2. Use empty lines for the dummy hunks at the bottom.

### Screenshots

Before:
image

After:
image

## Release notes

Notes: [Fixed] Use unfold icon for diff expansion

### DIFF
--- a/app/src/ui/diff/diff-syntax-mode.ts
+++ b/app/src/ui/diff/diff-syntax-mode.ts
@@ -1,4 +1,4 @@
-import { DiffHunk, DiffLine } from '../../models/diff'
+import { DiffHunk, DiffLine, DiffLineType } from '../../models/diff'
 import * as CodeMirror from 'codemirror'
 import { diffLineForIndex } from './diff-explorer'
 import { ITokens } from '../../lib/highlighter/types'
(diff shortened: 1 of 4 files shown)


app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -112,10 +118,22 @@ export class DiffSyntaxMode {
     return { diffLineIndex: 0, previousHunkOldEndLine: null }
   }
 
-  // Should never happen except for blank diffs but
-  // let's play along
   public blankLine(state: IState) {
```
sergiou87, line 121: Took me a while to find that empty lines went through this function 🤣

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -131,13 +149,17 @@ export class DiffSyntaxMode {
         state.diffLineIndex++
       }
 
-      const token = index ? TokenNames[index] : null
+      if (index === null) {
+        return null
+      }
+
+      const token = TokenNames[index] ?? null
```
sergiou87, line 156: This is unrelated to the purposes of this PR, but I noticed the `token` ended up being `undefined` instead of `null` sometimes.

niik (approved):
Looks great to me! 👍

PR #12887, opened 2021-09-01
Use same width for hunk expansion handles as we do for line numbers

niik:
## Description

Noticed that the diff expansion handle gutters didn't use the same size as line numbers in the unified diff view

### Before
image

### After
image

## Release notes

Notes: no-notes


sergiou87 (approved):
Looks good to me! I just noticed that in split mode we show the line numbers at the left (of each half) for commits, but in the center for the Changes tab 🤯 

What's the reasoning behind that?

Anyway, this can be merged :shipit: Thanks for taking care of it ❤️

niik:
> What's the reasoning behind that?

It's so that you can drag to select multiple lines in both columns in the changes view which isn't needed in the history view. It's a UX problem that dotcom hasn't needed to solve 😄

Djamma1986:
``Document
Edit

PR #12892, opened 2021-09-02
Release 2.9.4-beta1

tidy-dev:
## Description

Looking for the PR for the upcoming first beta of the v2.9.4 series? Well you've just found it, congratulations!

## Release checklist

image
---- No new metrics

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -4,7 +4,7 @@
       "[Added] Add a menu item to view a committed file change on GitHub - #12492. Thanks @haykam821!",
       "[Added] Customizable high contrast theme - #12833",
       "[Fixed] App no longer crashes on successful merge after conflicts resolved - #12888",
-      "[Improved] Use same width for hunk expansion handles as we do for line numbers - #12887",
+      "[Fixed] Use same width for hunk expansion handles as we do for line numbers - #12887",
       "[Fixed] Use the correct icon for diff expansion - #12882"
     ],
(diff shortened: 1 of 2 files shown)


changelog.json
```diff
@@ -1,5 +1,12 @@
 {
   "releases": {
+    "2.9.4-beta1": [
+      "[Added] Add a menu item to view a committed file change on GitHub - #12492. Thanks @haykam821!",
+      "[Added] Customizable high contrast theme - #12833",
+      "[Fixed] App no longer crashes on successful merge after conflicts resolved - #12888",
+      "[Improved] Use same width for hunk expansion handles as we do for line numbers - #12887",
```
sergiou87: This sounds more like a fix to me, but in any case @niik tagged this with `no-notes`… since this is an user-facing fix, I'd say we add the release note anyway.

changelog.json
```diff
@@ -1,5 +1,12 @@
...
+      "[Improved] Small fixes/improvements in diff expansion - #12882"
```
sergiou87: Maybe we can be a bit more specific about this one? Also I think it should be a fix.
```suggestion
      "[Fixed] Use the correct icon for diff expansion - #12882"
```

billygriffin (dismissed):

sergiou87 (approved):

billygriffin (approved):

PR #12893, opened 2021-09-02
Diff expansion optimizations

sergiou87:
## Description

In a conversation with @niik, he pointed out that the app was loading the file contents several times since diff expansion was added. It was especially bad that the app read the files (old and new) twice in a row the first time a diff is displayed.

After discussing some alternatives, we decided to cache the file contents in memory, when the diff can be expanded. That will happen when the files, old and new, are smaller than 256KB. That means, we would only cache up to 0.5MB.

### Screenshots

Before:

https://user-images.githubusercontent.com/1083228/131876584-7a63f60f-fcc2-4d6b-b3b6-3c133cbae40a.mov

After:

https://user-images.githubusercontent.com/1083228/131873777-48fe982e-3bf6-4b1c-8a0b-afbd04a11dd8.mov

## Release notes

Notes: [Improved] Better performance displaying diffs

### DIFF
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -392,9 +392,9 @@ export class SideBySideDiff extends React.Component<
     }
 
     const newContentLines =
-      contents.newContents === null ? [] : contents.newContents.split('\n')
+      contents.newContents === null ? [] : contents.newContents
     const oldContentLines =
-      contents.oldContents === null ? [] : contents.oldContents.split('\n')
+      contents.oldContents === null ? [] : contents.oldContents
 
     const currentDiff = this.state.diff
     const shouldEnableDiffExpansion =
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -12,6 +12,7 @@ import {
   getLineFilters,
   getFileContents,
   highlightContents,
+  IFileContents,
 } from './syntax-highlighting'
 import { ITokens, ILineTokens, IToken } from '../../lib/highlighter/types'
 import {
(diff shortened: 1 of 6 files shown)


app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -114,10 +114,29 @@ interface ISideBySideDiffProps {
...
   /** The diff that should be rendered */
   readonly diff: ITextDiff
 
+  /** Contents of the old and new files for this diff. */
+  readonly contents: ISideBySideTextDiffDocumentContents | null
+}
+
+interface ISideBySideDiffState {
+  /** Represents a diff and its associated files (old and new) */
+  readonly document: ISideBySideTextDiffDocument
```
sergiou87, line 128: Not a fan of the `document` word, since CodeMirror also uses it, but I couldn't come up with a better name 🤣

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -114,10 +115,18 @@ interface ISideBySideDiffProps {
   readonly showSideBySideDiff: boolean
 }
 
-interface ISideBySideDiffState {
+interface ISideBySideTextDiffDocument {
   /** The diff that should be rendered */
   readonly diff: ITextDiff
 
+  /** Contents of the old and new files for this diff. */
+  readonly contents: IFileContents | null
```
sergiou87, line 123: Not super proud of exposing and using `IFileContents` here, but we basically needed everything inside of that interface (except `file`), so… 😕

app/src/ui/diff/side-by-side-diff.tsx
tidy-dev: Would `diffDocument` work just to give it that differentiation?

tidy-dev (approved):
✨  Nice! Tested it out and works great.

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: That feels a bit redundant, but definitely better! In any case, after some chat with @niik about how to improve diff rendering performance, I might get rid of this structure entirely 😅

PR #12902, opened 2021-09-03
Update error messages

ghost:
## Description

This updates two error messages defined in `app-store.ts`.

### Message 1 (line 5273)
- _Fail_ goes with _to_ and the infinitive rather than with the gerund. Here’s an example from the _Cambridge Dictionary_: “John failed _to turn up_ for football practice yesterday” (emphasis added).
- _Repository directory_ needs an article.
- The construction “a common reason&nbsp;.&nbsp;.&nbsp;. is if” is unsuitable here. The conjuction _if_ attaches to _is_, the sentence suggesting that an undescribed reason occurs under a certain condition. “A common reason&nbsp;.&nbsp;.&nbsp;. is that” will fulfill the function needed.
- I believe the second sentence should be referring to the repository directory or a part of it, not any file or directory at all.

### Message 2 (lines 5758–5760)
See the third point above. I’ve also replaced _cause_ with _reason_ for consistency.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5270,7 +5270,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         if (!deleted) {
           this.emitError(
             new Error(
-              `Failed moving repository directory to ${TrashNameLabel}.\n\nA common reason for this is if a file or directory is open in another program.`
+              `Failed to move the repository directory to ${TrashNameLabel}.\n\nA common reason for this is that the directory or a part of it is open in another program.`
             )
           )
           return
@@ -5756,7 +5756,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       this.emitError(
         new Error(
           `Couldn't find branch '${headRefName}' in remote '${remote.name}'. ` +
-            `A common cause for this is if the PR author has deleted their ` +
+            `A common reason for this is that the PR author has deleted their ` +
             `branch or their forked repository.`
         )
       )
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -5270,7 +5270,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         if (!deleted) {
           this.emitError(
             new Error(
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -5270,7 +5270,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         if (!deleted) {
           this.emitError(
             new Error(
-              `Failed moving repository directory to ${TrashNameLabel}.\n\nA common reason for this is if a file or directory is open in another program.`
+              `Failed to move the repository directory to ${TrashNameLabel}.\n\nA common reason for this is that the directory or a part of it is open in another program.`
```
tidy-dev: I think we loose some context if we take the word `file` out of it. How about?

```suggestion
              `Failed to move the repository directory to ${TrashNameLabel}.\n\nA common reason for this is that the file or its directory is open in another program.`
```

mahamadmansor (commented):
Good

app/src/lib/stores/app-store.ts
ghost: Thanks for the review! The determiner _the_ would suggest that a definite file is in question, but neither the context nor the error message specifies a file. My understanding is that GitHub Desktop fails to delete a repository directory when said directory or _any_ of its files is open in another program. (Do let me know if I’m mistaken.) I hence suggest the following: 
 
> A common reason for this is that the directory or one of its files is open in another program.

I look forward to hearing your thoughts.

tidy-dev: Yes. That sounds good. I think a common issue is that a user has a file from the repo open in another application such as a code editor.

ghost: Thank you for the feedback! Commit a3e1ba0 updates the error message as we have agreed, which resolves this conversation.

tidy-dev (approved):
✨ Thank you for the much better verbiage!

ghost:
Happy to help!

Hakim77pixel:
Pada Jum, 3 Sep 2021, 10:55 PTG Patryk Michalik ***@***.***>
menulis:

son3ofSatan (commented):
please

PR #12903, opened 2021-09-03
Improve diff render and load

sergiou87:
## Description

This PR brings MOAR optimizations discussed with @niik. After #12893, we also wanted to get rid of some of the glitches that made the experience jarring, like lines changing height or the hunk at the bottom to expand the diff popping up a few milliseconds:

https://user-images.githubusercontent.com/1083228/132034431-cabef7a3-96a2-4599-870d-c5424a35cbe8.mov

The problem was that for diff expansion we needed to load the file contents before we know whether or not the file can be expanded (also from the bottom). Once we know, we can start showing the nice unfold icons, and for that we might need to tweak the height of the affected lines.

These contents were loaded a few milliseconds after the diff is rendered for the first time, so we've moved here that part to the `SeamlessDiffSwitcher`. That way, we can leverage its logic to show the old diff for as long as we need before both the diff and the file contents are available, replacing that old diff with a spinner if the load takes too long, and finally switching to the new diff once everything is ready for perfect first render 😍 

### Screenshots

This is how it looks on my machine when I switch quickly between files:

https://user-images.githubusercontent.com/1083228/132034982-4cef499a-9ea1-48fd-9bd5-b61e1078aab2.mov

And then if I simulate that loading the file contents takes longer, we show the loading feedback as expected…

https://user-images.githubusercontent.com/1083228/132035096-069bb742-c376-4e6c-ac4d-59dc2a907f35.mov

## Release notes

Notes: [Fixed] Fixed glitches when diffs are shown for the first time

### DIFF
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -84,8 +84,8 @@ interface ISideBySideDiffProps {
   readonly diff: ITextDiff
 
   /**
-   * Contents of the old and new files related to the current diff. Only needed
-   * for diffs that can be expanded.
+   * Contents of the old and new files related to the current text diff. Must
+   * be null for diffs that cannot be expanded.
    */
   readonly fileContents: IFileContents | null
 
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -50,7 +50,6 @@ import { showContextualMenu } from '../main-process-proxy'
 import { getTokens } from './diff-syntax-mode'
 import { DiffSearchInput } from './diff-search-input'
 import { escapeRegExp } from '../../lib/helpers/regex'
-import { enableTextDiffExpansion } from '../../lib/feature-flag'
 import {
   expandTextDiffHunk,
   DiffExpansionKind,
@@ -752,7 +751,6 @@ export class SideBySideDiff extends React.Component<
     const contents = this.props.fileContents
     const { diff } = this.state
     if (
-      !enableTextDiffExpansion() ||
       contents === null ||
       !contents.canBeExpanded ||
       contents.newContents.length === 0
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -238,16 +238,22 @@ export class SideBySideDiff extends React.Component<
     }
   }
 
-  public render() {
+  private canExpandDiff() {
     const contents = this.props.fileContents
+    return (
+      contents !== null &&
+      contents.canBeExpanded &&
+      contents.newContents.length > 0
+    )
+  }
+
+  public render() {
     const { diff } = this.state
 
     const rows = getDiffRows(
       diff,
       this.props.showSideBySideDiff,
-      contents !== null &&
-        contents.canBeExpanded &&
-        contents.newContents.length > 0
+      this.canExpandDiff()
     )
     const containerClassName = classNames('side-by-side-diff-container', {
       'unified-diff': !this.props.showSideBySideDiff,
@@ -298,14 +304,11 @@ export class SideBySideDiff extends React.Component<
   }
 
   private renderRow = ({ index, parent, style, key }: ListRowProps) => {
-    const contents = this.props.fileContents
     const { diff } = this.state
     const rows = getDiffRows(
       diff,
       this.props.showSideBySideDiff,
-      contents !== null &&
-        contents.canBeExpanded &&
-        contents.newContents.length > 0
+      this.canExpandDiff()
     )
     const row = rows[index]
 
@@ -542,14 +545,11 @@ export class SideBySideDiff extends React.Component<
     rowNumber: number,
     column: DiffColumn
   ): number | null {
-    const contents = this.props.fileContents
     const { diff } = this.state
     const rows = getDiffRows(
       diff,
       this.props.showSideBySideDiff,
-      contents !== null &&
-        contents.canBeExpanded &&
-        contents.newContents.length > 0
+      this.canExpandDiff()
     )
     const row = rows[rowNumber]
 
@@ -748,13 +748,8 @@ export class SideBySideDiff extends React.Component<
   }
 
   private buildExpandMenuItem(): IMenuItem | null {
-    const contents = this.props.fileContents
     const { diff } = this.state
-    if (
-      contents === null ||
-      !contents.canBeExpanded ||
-      contents.newContents.length === 0
-    ) {
+    if (!this.canExpandDiff()) {
       return null
     }
 
@@ -779,11 +774,7 @@ export class SideBySideDiff extends React.Component<
     const contents = this.props.fileContents
     const { diff } = this.state
 
-    if (
-      contents === null ||
-      !contents.canBeExpanded ||
-      contents.newContents.length === 0
-    ) {
+    if (contents === null || !this.canExpandDiff()) {
       return
     }
 
@@ -910,7 +901,6 @@ export class SideBySideDiff extends React.Component<
   private onSearch = (searchQuery: string, direction: 'next' | 'previous') => {
     let { selectedSearchResult, searchResults: searchResults } = this.state
     const { showSideBySideDiff } = this.props
-    const contents = this.props.fileContents
     const { diff } = this.state
 
     // If the query is unchanged and we've got tokens we'll continue, else we'll restart
@@ -930,9 +920,7 @@ export class SideBySideDiff extends React.Component<
         diff,
         showSideBySideDiff,
         searchQuery,
-        contents !== null &&
-          contents.canBeExpanded &&
-          contents.newContents.length > 0
+        this.canExpandDiff()
       )
       selectedSearchResult = 0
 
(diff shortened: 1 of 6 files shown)


app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -382,20 +379,6 @@ export class SideBySideDiff extends React.Component<
...
-    if (
-      !highlightParametersEqual(
-        this.props,
-        propsSnapshot,
-        this.state,
-        stateSnapshot
-      )
-    ) {
-      return
-    }
```
niik, line 397: Could you elaborate for a bit on why this isn't needed anymore? It seems to me it could still be useful to bail here and not kick off another set of highlighter workers?

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -370,8 +362,13 @@ export class SideBySideDiff extends React.Component<
   }
 
   private async initDiffSyntaxMode() {
-    const { file, repository } = this.props
-    const { contents: currentContents, diff: currentDiff } = this.state.document
+    const contents = this.props.fileContents
+
+    if (contents === null) {
+      return
+    }
```
niik, line 374: I believe there's a change in behavior here. Previously we'd always syntax highlight the first 256kb of file contents and then bail if the file was longer. With this it looks like we skip on highlighting entirely if the file cannot be expanded (meaning it's 256kb  - 1b or longer).

I don't think whether a file can be expanded should be coupled to syntax highlighting

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -822,10 +788,7 @@ export class SideBySideDiff extends React.Component<
     this.diffToRestore = diff
 
     this.setState({
-      document: {
-        ...this.state.document,
-        diff: updatedDiff,
-      },
+      diff: updatedDiff,
```
niik: :art: Could we roll this up into just one line now please?

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -835,10 +798,7 @@ export class SideBySideDiff extends React.Component<
     }
 
     this.setState({
-      document: {
-        ...this.state.document,
-        diff: this.diffToRestore,
-      },
+      diff: this.diffToRestore,
```
niik: :art: Could we roll this up into just one line now please?

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -1017,10 +979,7 @@ export class SideBySideDiff extends React.Component<
     }
 
     this.setState({
-      document: {
-        ...this.state.document,
-        diff: updatedDiff,
-      },
+      diff: updatedDiff,
```
niik: :art: Could we roll this up into just one line now please?

app/src/ui/diff/text-diff.tsx
```diff
@@ -389,8 +381,13 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
       return
     }
 
-    const { file, repository } = this.props
-    const { contents: currentContents, diff: currentDiff } = this.state.document
+    const contents = this.props.fileContents
+
+    if (contents === null) {
```
niik, line 385: Same comment here as in side-by-side about not coupling expansion to highlighting

app/src/ui/diff/text-diff.tsx
```diff
@@ -402,20 +399,6 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
     const tsOpt = this.codeMirror.getOption('tabSize')
     const tabSize = typeof tsOpt === 'number' ? tsOpt : 4
 
-    const contents =
-      currentContents ?? (await getFileContents(repository, file, lineFilters))
-
-    if (
-      !highlightParametersEqual(
```
niik, line 409: Same question here about why this isn't needed anymore

app/src/ui/diff/text-diff.tsx
```diff
@@ -653,10 +610,7 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
     }
 
     this.setState({
-      document: {
-        ...this.state.document,
-        diff: updatedDiff,
-      },
+      diff: updatedDiff,
```
niik: :art: Could we roll this up into just one line now please?

app/src/ui/diff/text-diff.tsx
```diff
@@ -817,28 +771,25 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
 
-    this.diffToRestore = this.state.document.diff
+    this.diffToRestore = this.state.diff
 
     this.setState({
-      document: {
-        ...this.state.document,
-        diff: updatedDiff,
-      },
+      diff: updatedDiff,
```
niik: :art: Could we roll this up into just one line now please?

app/src/ui/diff/text-diff.tsx
```diff
@@ -849,10 +800,7 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
     }
 
     this.setState({
-      document: {
-        ...this.state.document,
-        diff: this.diffToRestore,
-      },
+      diff: this.diffToRestore,
```
niik: :art: Could we roll this up into just one line now please?

niik (changes requested):
I'm so so happy with this change, it feels so snappy now! :heart: Just one comment about highlighting and a few nitpicks about style but overall I'm ecstatic

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: Ouch! That's completely true, I'll fix it 😳 
Nice catch!

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: Unless I'm missing something, that was useful only because we loaded the file contents here, right?

I mean, that `highlightParametersEqual` was supposed to bail if the diff changed while the app was loading the file contents. With this PR, the file contents are not loaded here anymore, but in `SeamlessDiffSwitcher`, and that's where we're bailing: https://github.com/desktop/desktop/pull/12903/files#diff-c771a816631b2d7e1340655981ed97791b724216730e5353a624356ee692f5b0R245-R247

Does that make sense? 😳

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: @niik the last set of commits should fix this. Basically this check here makes sense, the problem is I was only passing down the `fileContents` when the diff can be expanded. Changing that logic required changing a few other checks here and there.

I also realized this new approach broke syntax highlighting for huge files (> 10MB), so I fixed that too. The only caveat of this is that we'll load the 0.5MB of file contents for huge files before the user clicked on `Show Diff`, but I think that should be fine since it won't be a very common scenario and, again, it's only 0.5MB. Does that sound reasonable to you?

app/src/ui/diff/side-by-side-diff.tsx
niik: > Does that make sense? 😳

It sure does, thanks!

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -1042,7 +993,7 @@ function highlightParametersEqual(
     (newProps === prevProps ||
       (newProps.file.id === prevProps.file.id &&
         newProps.showSideBySideDiff === prevProps.showSideBySideDiff)) &&
-    newState.document.diff.text === prevState.document.diff.text
+    newState.diff.text === prevState.diff.text
```
niik: I _think_ this now needs to be updated to return false when the `fileContents` change. As it stands now highlighting will not be applied for the first mount since at mount time the `fileContents` will be `null` and when it later gets loaded the check in [componentDidUpdate](https://github.com/desktop/desktop/blob/448c788f762d9ed49ef838912a83cae878e2b3fa/app/src/ui/diff/side-by-side-diff.tsx#L235-L240) will prevent `initDiffSyntaxMode` from running again. Perhaps something like this?

```suggestion
    newState.diff.text === prevState.diff.text &&
    prevProps.fileContents?.file.id === newProps.fileContents?.file.id
```

app/src/ui/diff/text-diff.tsx
```diff
@@ -87,7 +84,7 @@ function highlightParametersEqual(
 ) {
   return (
     (newProps === prevProps || newProps.file.id === prevProps.file.id) &&
-    newState.document.diff.text === prevState.document.diff.text
+    newState.diff.text === prevState.diff.text
```
niik: Similarly to the `highlightParametersEqual` method in side-by-side-diff

```suggestion
    newState.diff.text === prevState.diff.text &&
    prevProps.fileContents?.file.id === newProps.fileContents?.file.id
```

niik (changes requested):
Getting close, just one comment about the initial mount

Hakim77pixel:
Pada Sel, 7 Sep 2021, 8:46 PTG Markus Olsson ***@***.***>
menulis:

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: You're totally right, nice catch again! ❤️

niik (approved):
I love it :shipit:

Hakim77pixel:
Pada Sel, 7 Sep 2021, 10:03 PTG Markus Olsson ***@***.***>
menulis:

PR #12894, opened 2021-09-02
Add host key verification to auth issue list

ghost:
## Description

Recently I encountered an issue where I deleted my `.ssh/known_hosts` file and afterward was unable to fetch and push from a repository I was previously able to use with GitHub Desktop. The app showed an "Authentication failure".
Once I verified the host's fingerprint running, e.g. `ssh git@github.com`, everything worked fine.

The dialog tries to be helpful and expansive, but it did not lead me to the resolution.
Adding this helps people save some time; however, handling host key verification in-app in the future may be a helpful idea.

I hope you find this valuable, too.

### Screenshots

Screenshot of updated dialog

## Release notes

Notes: no-notes

PR #12919, opened 2021-09-07
Add unit tests for our SSH host prompt parser

sergiou87:
## Description

After the work done in #12894 I think it'd be useful to keep around some test cases in case future versions of OpenSSH change this prompt, make sure we don't introduce regressions.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #12922, opened 2021-09-07
Improves added and deleted line borders for high contrast mode

tidy-dev:
## Description

Improves added and deleted line borders for high contrast mode. Previously high-contrast mode put a border around each line. Goal was to have borders around each group of deleted or added lines.

### Screenshots

Split:
image

Unified:
image

## Release notes
Notes: [Improved] Use a border around groups of added and deleted lines in high-contrast mode

### DIFF
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -393,3 +393,30 @@ export function getNumberOfDigits(val: number): number {
  * GitHub.com's behavior.
  **/
 export const MaxIntraLineDiffStringLength = 1024
+
+/**
+ * Used to obtain classes applied to style the row if it is the first or last of
+ * a group of added, deleted, or modified rows in the unified diff.
+ **/
+export function getFirstAndLastClassesUnified(
+  tokenIndex: string,
+  prevTokenIndex: string | undefined,
+  nextTokenIndex: string | undefined
+): string[] {
+  const addDeleteTokens = ['+', '-']
+  if (!addDeleteTokens.includes(tokenIndex)) {
+    return []
+  }
+
+  const classNames = []
+
+  if (prevTokenIndex !== tokenIndex) {
+    classNames.push('is-first')
+  }
+
+  if (nextTokenIndex !== tokenIndex) {
+    classNames.push('is-last')
+  }
+
+  return classNames
+}
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -395,6 +395,39 @@ export function getNumberOfDigits(val: number): number {
 export const MaxIntraLineDiffStringLength = 1024
 
 /**
+ * Used to obtain classes applied to style the row as first or last of a group
+ * of added or deleted rows in the side-by-side diff.
+ **/
+export function getFirstAndLastClassesSideBySide(
+  row: SimplifiedDiffRow,
+  previousRow: SimplifiedDiffRow | undefined,
+  nextRow: SimplifiedDiffRow | undefined,
+  addedDeleted: DiffRowType
+): ReadonlyArray<string> {
+  const classes: Array<string> = []
+  const typesToCheck = [addedDeleted, DiffRowType.Modified]
+
+  // Is the row of the type we are checking? No. Then can't be first or last.
+  if (!typesToCheck.includes(row.type)) {
+    return []
+  }
+
+  // Is the previous row exist or is of the type we are checking?
+  // No. Then this row must be the first of this type.
+  if (previousRow === undefined || !typesToCheck.includes(previousRow.type)) {
+    classes.push('is-first')
+  }
+
+  // Is the next row exist or is of the type we are checking?
+  // No. Then this row must be last of this type.
+  if (nextRow === undefined || !typesToCheck.includes(nextRow.type)) {
+    classes.push('is-last')
+  }
+
+  return classes
+}
+
+/**
  * Used to obtain classes applied to style the row if it is the first or last of
  * a group of added, deleted, or modified rows in the unified diff.
  **/
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -402,7 +402,7 @@ export function getFirstAndLastClassesSideBySide(
   row: SimplifiedDiffRow,
   previousRow: SimplifiedDiffRow | undefined,
   nextRow: SimplifiedDiffRow | undefined,
-  addedOrDeleted: DiffRowType
+  addedOrDeleted: DiffRowType.Added | DiffRowType.Deleted
 ): ReadonlyArray<string> {
   const classes = new Array<string>()
   const typesToCheck = [addedOrDeleted, DiffRowType.Modified]
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -404,7 +404,7 @@ export function getFirstAndLastClassesSideBySide(
   nextRow: SimplifiedDiffRow | undefined,
   addedOrDeleted: DiffRowType
 ): ReadonlyArray<string> {
-  const classes: Array<string> = []
+  const classes = new Array<string>()
   const typesToCheck = [addedOrDeleted, DiffRowType.Modified]
 
   // Is the row of the type we are checking? No. Then can't be first or last.
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -437,10 +437,7 @@ export function getFirstAndLastClassesUnified(
   prevToken: DiffSyntaxToken | null,
   nextToken: DiffSyntaxToken | null
 ): string[] {
-  const addedOrDeletedTokens = [
-    DiffSyntaxToken.DiffAdded,
-    DiffSyntaxToken.DiffDeleted,
-  ]
+  const addedOrDeletedTokens = [DiffSyntaxToken.Add, DiffSyntaxToken.Delete]
   if (!addedOrDeletedTokens.includes(token)) {
     return []
   }
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -10,6 +10,7 @@ import {
 } from '../../models/status'
 import { DiffHunk, DiffHunkExpansionType } from '../../models/diff/raw-diff'
 import { DiffLineType } from '../../models/diff'
+import { DiffSyntaxToken } from './diff-syntax-mode'
 
 /**
  * DiffRowType defines the different types of
@@ -432,22 +433,25 @@ export function getFirstAndLastClassesSideBySide(
  * a group of added, deleted, or modified rows in the unified diff.
  **/
 export function getFirstAndLastClassesUnified(
-  tokenIndex: string,
-  prevTokenIndex: string | undefined,
-  nextTokenIndex: string | undefined
+  token: DiffSyntaxToken,
+  prevToken: DiffSyntaxToken | null,
+  nextToken: DiffSyntaxToken | null
 ): string[] {
-  const addDeleteTokens = ['+', '-']
-  if (!addDeleteTokens.includes(tokenIndex)) {
+  const addedOrDeletedTokens = [
+    DiffSyntaxToken.DiffAdded,
+    DiffSyntaxToken.DiffDeleted,
+  ]
+  if (!addedOrDeletedTokens.includes(token)) {
     return []
   }
 
   const classNames = []
 
-  if (prevTokenIndex !== tokenIndex) {
+  if (prevToken !== token) {
     classNames.push('is-first')
   }
 
-  if (nextTokenIndex !== tokenIndex) {
+  if (nextToken !== token) {
     classNames.push('is-last')
   }
 
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -402,10 +402,10 @@ export function getFirstAndLastClassesSideBySide(
   row: SimplifiedDiffRow,
   previousRow: SimplifiedDiffRow | undefined,
   nextRow: SimplifiedDiffRow | undefined,
-  addedDeleted: DiffRowType
+  addedOrDeleted: DiffRowType
 ): ReadonlyArray<string> {
   const classes: Array<string> = []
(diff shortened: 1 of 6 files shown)


app/src/ui/diff/diff-helpers.tsx
```diff
@@ -387,3 +387,32 @@ export function getLargestLineNumber(hunks: DiffHunk[]): number {
 export function getNumberOfDigits(val: number): number {
   return (Math.log(val) * Math.LOG10E + 1) | 0
 }
+
+export function getFirstAndLastClasses(
+  row: SimplifiedDiffRow,
+  previousRow: SimplifiedDiffRow | undefined,
+  nextRow: SimplifiedDiffRow,
```
sergiou87: This one could also be `undefined`
```suggestion
  nextRow: SimplifiedDiffRow | undefined,
```

sergiou87 (dismissed):
I love it!! The only problem I see with this in unified diffs is that the changes in `diff-syntax-mode` rely on having the diff hunks. However, those are not available on the first pass, because the `text-diff.tsx` does something like this:
1. Render diff with a partially initialized `DiffSyntaxMode` instance (without diff)
2. Get a reference to the Code Mirror instance
3. Parse the original file for syntax highlighting
4. Create a new `DiffSyntaxMode`, now with the diff hunks and also the tokens for syntax highlighting
5. Render with all that info

The consequence of this is a short flickering between the render of (1) (without these new borders) and the render of (5) (with them).

I see two options here to address that… 
1. Somehow try to make CodeMirror use the diff hunks in the first render
2. Not rely on diff hunks to detect groups of added/deleted lines. You can use a new field in `IState` to keep track of the type of the previous line, and then `stream.lookAhead(1)` to get the next line. However, I see `lookAhead` in the [documentation](https://codemirror.net/doc/manual.html#modeapi) and in the [code](https://github.com/codemirror/CodeMirror/blob/bd1b7d2976d768ae4e3b8cf209ec59ad73c0305a/src/util/StringStream.js#L80) but not in the [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/codemirror/index.d.ts) 🤔 

I think (1) would be ideal. In any case I'm not against merging this as it is right now and fix that in a different PR if needed for whatever reason!

tidy-dev:
Been attempting to reduce the flicker. Looked briefly into (1) but seems it is because we are waiting on the first run through of the code mirror render for the hunks?  

As for number (2), in the debugger, I can see that the `Codemirror.StringStream` has a `lineOracle` on it that has the `lookAhead` method. I attempted use that but it always came back `undefined`. However, `lineOracle` has a `line` property and `doc` on it which has a `getLine` method on it.  Unfortunately, like you said those do not appear to have typing or types in  the `CodeMirror` library in order to properly expose them. So.. I made my own interface that extends `Codemirror.StringStream` and then just casting the `Codemirror.StringStream` sent in to that so I can access those properties. This works... but feels really hacky and potentially fragile if Codemirror would change their implementation to not have that `LineOracle` property on it anymore.

Additionally, I noticed we already have flicker because the syntax highlighting for keywords and such doesn't get applied till the second render either - so.. is the hack worth it?

I have pushed that solution so you can take a look.

sergiou87:
I think we have the hunks in time for (1), the problem is we don't get a reference to the Code Mirror instance until it's too late 😢 

Regarding (2), I don't think we need to expose the `lineOracle` property, just the `lookAhead` is enough, at least I see it from the debugger:
image

But instead of hacking it, I think I'd just submit a PR to https://github.com/DefinitelyTyped/DefinitelyTyped/ and once it's done I'd use that method here. Could you try that `lookAhead` with your hack and see if that works?

tidy-dev:
Done and [PR to DefinitelyTypes](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55661) is open

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -387,3 +387,32 @@ export function getLargestLineNumber(hunks: DiffHunk[]): number {
 export function getNumberOfDigits(val: number): number {
   return (Math.log(val) * Math.LOG10E + 1) | 0
 }
+
+export function getFirstAndLastClasses(
+  row: SimplifiedDiffRow,
+  previousRow: SimplifiedDiffRow | undefined,
+  nextRow: SimplifiedDiffRow | undefined,
+  addedDeleted: DiffRowType
```
niik: It took me a while to understand this parameter. Could we rename this to `addedOrDeleted`? I think that would have made it easier for me to grasp

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -42,6 +42,11 @@ const TokenNames: { [key: string]: DiffSyntaxToken | null } = {
 interface IState {
   diffLineIndex: number
   previousHunkOldEndLine: number | null
+  prevLineTokenIndex: string | null
+}
+
+type StringStreamWithLookahead = CodeMirror.StringStream & {
+  lookAhead: (n: number) => string | undefined
```
niik: Upon closer inspection of the codemirror source it seems this is either gonna be:

- a string
- `false` (https://github.com/codemirror/CodeMirror/blob/22e9d4f6ce451196803898be1bad859567df6107/src/line/highlight.js#L25-L28)
- `undefined` (https://github.com/codemirror/CodeMirror/blob/22e9d4f6ce451196803898be1bad859567df6107/addon/runmode/runmode.js#L64)

For our use case it's likely only ever going to be a string or undefined though. I wonder if it would make sense to invert our checks though to look for strings specifically.

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -166,7 +205,19 @@ export class DiffSyntaxMode {
         return null
       }
 
-      let result = getBaseDiffLineStyle(token)
+      // Hack to get the next line. We found that the StringStream has an extra
+      // property of lookAhead at run time that has what we need to access the
+      // next line. This just exposes it so we can use it.
+      const nextLine = (stream as StringStreamWithLookahead).lookAhead(1)
+      const nextLineTokenIndex = nextLine !== undefined ? nextLine[0] : null
```
niik: This type is lying a bit, it's defined as `string | null` but it can actually be `string | null | undefined`. I think it would make sense to redefine the type to be `string | undefined` and change this to

```suggestion
      const nextLineTokenIndex =
        typeof nextLine === 'string' ? nextLine[0] : undefined
```

app/styles/ui/_diff.scss
```diff
@@ -269,13 +269,32 @@
 .CodeMirror-linebackground {
   &.diff-add {
     background: var(--diff-add-background-color);
-    border: var(--diff-add-border);
-    margin-top: 1px;
+    border-right: var(--diff-add-border);
+    border-left: var(--diff-add-border);
```
niik: Similarly  to what we talked about re: high contrast theme borders being defined only for the high contrast theme I think we should do the same here. One approach  would be to add a mixin similar to https://github.com/desktop/desktop/blob/f9178938f1190dd289d62d4c36fb12f101c9c615/app/styles/mixins/_platform.scss#L9-L13

Then  we could do something like
```suggestion
    @include high-contrast-theme {
      border-right: var(--diff-add-border);
      border-left: var(--diff-add-border);
    }
```

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -387,3 +387,32 @@ export function getLargestLineNumber(hunks: DiffHunk[]): number {
 export function getNumberOfDigits(val: number): number {
   return (Math.log(val) * Math.LOG10E + 1) | 0
 }
+
+export function getFirstAndLastClasses(
```
niik: Woah, this function is infinitely better than the hack of nested if statements I set you up for, nicely done!

niik (changes requested):
Looks great, I really like the abstraction method here!

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -166,7 +179,17 @@ export class DiffSyntaxMode {
         return null
       }
 
-      let result = getBaseDiffLineStyle(token)
+      const nextLine = stream.lookAhead(1)
+      const nextLineTokenIndex =
+        typeof nextLine === `string` ? nextLine[0] : undefined
```
niik: Not that it really matters in practice but we can just use single quotes here since we're not interpolating. 

```suggestion
        typeof nextLine === 'string' ? nextLine[0] : undefined
```

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -138,8 +148,15 @@ export class SideBySideDiffRow extends React.Component<
...
-
+    const {
+      row,
+      showSideBySideDiff,
+      beforeClassNames,
+      afterClassNames,
+    } = this.props
+
+    const beforeClasses = classNames('before', ...beforeClassNames)
+    const afterClasses = classNames('after', ...afterClassNames)
```
niik, line 159: I think we can just se the array here, no need to spread it
```suggestion
    const beforeClasses = classNames('before', beforeClassNames)
    const afterClasses = classNames('after', afterClassNames)
```

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -393,3 +393,63 @@ export function getNumberOfDigits(val: number): number {
...
+ * Used to obtain classes applied to style the row as first or last of a group
+ * of added or deleted rows in the side-by-side diff.
+ **/
+export function getFirstAndLastClassesSideBySide(
+  row: SimplifiedDiffRow,
+  previousRow: SimplifiedDiffRow | undefined,
+  nextRow: SimplifiedDiffRow | undefined,
+  addedDeleted: DiffRowType
+): ReadonlyArray<string> {
+  const classes: Array<string> = []
```
niik:
```suggestion
  const classes = new Array<string>()
```

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -166,7 +179,17 @@ export class DiffSyntaxMode {
         return null
       }
 
-      let result = getBaseDiffLineStyle(token)
+      const nextLine = stream.lookAhead(1)
+      const nextLineTokenIndex =
```
niik: What is the `index` suffix referring to here and in the other variables ending with `index`? I would expect an `index` variable to be numeric but here they're all strings.

app/src/ui/diff/diff-helpers.tsx
```diff
@@ -393,3 +393,63 @@ export function getNumberOfDigits(val: number): number {
...
+
+/**
+ * Used to obtain classes applied to style the row as first or last of a group
+ * of added or deleted rows in the side-by-side diff.
+ **/
+export function getFirstAndLastClassesSideBySide(
+  row: SimplifiedDiffRow,
+  previousRow: SimplifiedDiffRow | undefined,
+  nextRow: SimplifiedDiffRow | undefined,
+  addedOrDeleted: DiffRowType
```
niik: I wonder if we could make this even more clear by explicitly stating the types we allow here

```suggestion
  addedOrDeleted: DiffRowType.Added | DiffRowType.Deleted
```

niik (changes requested):
Looking great, just a few questions and then I'll take a more thorough look tomorrow

app/src/ui/diff/side-by-side-diff-row.tsx
tidy-dev: This is because they're readonly.
image

app/src/ui/diff/diff-syntax-mode.ts
tidy-dev: I think I was following the existing word choice of `index` on line 166. Here index refers to  looking for `+`, `-`, `@`,  or ` ` as they are indexes (maybe more accurately described as `key`) for the `TokenNames` map.

tidy-dev: After looking at this, I decided to use what we define in this code as a `DiffSyntaxToken` instead of using the line markers aka token keys. Then while I was doing that, I updated the type of DiffSyntaxToken to an enum .. because I think they make things more readable.

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -30,18 +31,24 @@ export interface IDiffSyntaxModeSpec extends IDiffSyntaxModeOptions {
...
 
-type DiffSyntaxToken = 'diff-add' | 'diff-delete' | 'diff-hunk' | 'diff-context'
+export enum DiffSyntaxToken {
+  Add = 'diff-add',
+  Delete = 'diff-delete',
+  Hunk = 'diff-hunk',
+  Context = 'diff-context',
+}
 
 const TokenNames: { [key: string]: DiffSyntaxToken | null } = {
```
niik: I don't know how this mistake of mine has lived for so long but if you could please change this to token or undefined instead please? Null is just a lie

```ts
const TokenNames: Record<string, DiffSyntaxToken | undefined> = {
```

app/src/ui/diff/diff-syntax-mode.ts
```diff
@@ -150,27 +168,38 @@ export class DiffSyntaxMode {
...
+      const token = TokenNames[tokenKey] ?? null
 
       if (token === null) {
         return null
       }
 
-      let result = getBaseDiffLineStyle(token)
+      const nextLine = stream.lookAhead(1)
+      const nextLineToken =
+        typeof nextLine === 'string' ? TokenNames[nextLine[0]] : null
```
niik: If you inspect the type of `nextLineToken` here TypeScript will tell you it's `string | null` which is a lie since an empty line will produce `undefined` (`""[0] === undefined`) and so will a line that starts with a character that doesn't exist in the TokenNames object.

Either we change the type to be `string | undefined` by means of changing the `: null` to `: undefined`

```ts
typeof nextLine === 'string' ? TokenNames[nextLine[0]] : undefined
```

or we keep the current type by converting undefined to null

```ts
typeof nextLine === 'string' ? TokenNames[nextLine[0]] ?? null : null
```

app/src/ui/diff/side-by-side-diff-row.tsx
niik: Ah, that's a bummer. Thanks for checking!

niik (changes requested):
Really close now @tidy-dev!

app/src/ui/diff/diff-syntax-mode.ts
tidy-dev: Went with everything could be undefined.

niik (approved):
✨ So sorry this dragged out for so long @tidy-dev, this looks great :shipit:

PR #12935, opened 2021-09-09
Show add repo dialog when opening repo from CLI

niik:
## Description

Showing the add repo dialog when running `github .`, let users double check that they're adding the intended repo. Does not affect the behavior if the repository is already added in the app.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨ Works great!

billygriffin (commented):
🎉

PR #12944, opened 2021-09-10
Display CI checks 

tidy-dev:
## Description

This PR is the logic to display the PR check runs. This PR is to get the UI display feature under review - feature flagged to development.

There are business logic semantics to determine like what should be displayed in the description.. and if we should use the app names for the headers. Additionally, we may get some UI tweaks form the designer in terms of specific css/layout.
Thus, this can be thought of "in development" first draft in terms of final appearance/text, but the underlying showing the checks in the UI logic will likely be what is merged with this PR and is why I want to go ahead and get this PR up. 

### Screenshots
(button isn't blue outlined, my screenshot tool keeps doing that)

image

(These are just what are already in the description fields)
image

image

## Release notes
Notes: [Added] Show PR CI check runs.


tidy-dev:
Notes.. When investigating how to rerun all the checks. I learned about workflows, workflow runs, and jobs. To my current understanding, check run is some sort of thing in between that has been built based on those things... but for the purposes of showing logs and rerunning we are actually are talking about workflow runs and a workflow run jobs. Thus, tho I will reuse a lot of the UI in this PR. The guts about where data comes from is going to change... Thus, moving this to a draft.

tidy-dev:
>Notes.. When investigating how to rerun all the checks. I learned about workflows, workflow runs, and jobs. To my current understanding, check run is some sort of thing in between that has been built based on those things... but for the purposes of showing logs and rerunning we are actually are talking about workflow runs and a workflow run jobs. Thus, tho I will reuse a lot of the UI in this PR. The guts about where data comes from is going to change... Thus, moving this to a draft.

Nvm... :D More understanding..  TLDR; We must use Checks API and Actions API - Checks API for non-Actions check run app and Actions for actions apps.

A check run is a check that is run and resides in a check-suite (a set of checks that are run). These check runs can be run by any app that is programmed to interface with the check runs api. GitHub Actions is such an app and it calls its things -  workflow, workflow runs and jobs. However, we want to display check runs whether or not the PR uses Actions. Thus, this PR was on the right track. 

I was hopeful that maybe we didn't have to make calls to the Actions API if the rerequest call in the checks api is opened up for our use ([pending issue)](https://github.com/github/c2c-actions-checks/issues/94). However... logs. 

Currently when an app interacts with the check runs api, it can decide what it returns in an `output` property. For apps, that are not Actions, this is our only option for obtaining logs. Unfortunately, Actions does not utilize it for sending logs. 

Therefore, I have another branch going that will build on top of this PR. That will call the Actions api to retrieve the workflows for the PR and because both contain a check suite id. I can match up the ones from the checks api that come from Actions and use actions to provide logs and re-run actions check runs (which are called workflow runs in Actions).

app/src/lib/stores/commit-status-store.ts
```diff
@@ -389,33 +390,121 @@ function apiStatusToRefCheck(apiStatus: IAPIRefStatusItem): IRefCheck {
...
+  durationSeconds?: number
+): string {
+  if (status !== APICheckStatus.Completed || conclusion === null) {
+    return 'In progress'
+  }
+
+  if (status !== APICheckStatus.Completed) {
+    return 'In progress'
+  }
+
```
sergiou87: This is a leftover, right?
```suggestion
```

app/src/ui/branches/ci-status.tsx
```diff
@@ -36,11 +40,15 @@ export class CIStatus extends React.PureComponent<
...
     this.state = {
-      check: props.dispatcher.tryGetCommitStatus(
-        this.props.repository,
-        this.props.commitRef
-      ),
+      check,
+    }
+    if (this.props.onCheckChange !== undefined) {
+      this.props.onCheckChange(check)
     }
```
sergiou87: I think we could use this to clean up those `!== undefined` checks. What do you think?
```suggestion
    this.props.onCheckChange?.(check)
```

app/src/ui/branches/ci-status.tsx
```diff
@@ -86,6 +94,10 @@ export class CIStatus extends React.PureComponent<
   }
 
   private onStatus = (check: ICombinedRefCheck | null) => {
+    if (this.props.onCheckChange !== undefined) {
+      this.props.onCheckChange(check)
+    }
```
sergiou87, line 97: Same
```suggestion
    this.props.onCheckChange?.(check)
```

sergiou87 (changes requested):
I tested it a bit and noticed a few things…
- Not sure why but the popover is not correctly aligned for me as it is in your screenshots:
image

- Clicking on the badge while the popover is visible doesn't close the popover, just makes it "flicker"
Kapture 2021-09-22 at 12 39 16

- You can show the popover while the branch selector is open, which looks a bit crowded
image

- I tried a PR with lots of jobs (https://github.com/git/git/pull/1094) and there is no way to scroll the list because the popover is taller than the window 😳 

image

- It also seems the contents of each item in the list of check runs is not properly vertically aligned:
image

app/src/lib/stores/commit-status-store.ts
tidy-dev: :D Yeah

app/src/ui/branches/ci-status.tsx
tidy-dev: I am totally good with that.. but I thought we were avoiding non-direct checks. :/

tidy-dev:
> Not sure why but the popover is not correctly aligned for me as it is in your screenshots:

Made some UI input changes to width and text layout and forgot to adjust for the caret. Thanks

tidy-dev:
> I tried a PR with lots of jobs (git/git#1094) and there is no way to scroll the list because the popover is taller than the window 😳

Yes. This I was aware of.. but we had some discussions about making the headers accordion and possibly logs behave the same way. I know these things will interact  and so I stopped going forward on final UI for this PR until I figure out definitely UI endpoint on that front.

tidy-dev:
> It also seems the contents of each item in the list of check runs is not properly vertically aligned:
Are you referring to the icon? Because it used to be, but I received feedback from Ramil that it should be aligned with the top text. Therefore I updated it.

app/src/ui/branches/ci-status.tsx
sergiou87: What do you mean? 😳

sergiou87:
> Are you referring to the icon? Because it used to be, but I received feedback from Ramil that it should be aligned with the top text. Therefore I updated it.

I'm actually referring to all of it because the block of text (title + subtitle) seems to be closer to the top. Not sure if that's because of the specific strings used there, but it's definitely how it feels to me 😳

tidy-dev:
> You can show the popover while the branch selector is open, which looks a bit crowded

This one I wasn't sure if we should close it or not.. but now that I did it. It does feel better. Thanks!

> Clicking on the badge while the popover is visible doesn't close the popover, just makes it "flicker"

I had noticed this and had just added on my working branch, but went ahead and added to this PR since it is more UI related. Thanks. :)

tidy-dev:
Adjusted the text.. not sure if the status indicator now feels too high. :/ 
image



Not pushed.. but here is what it would look like if status indicator was dropped a pixel.
image

tidy-dev:
Decided to go with dropping status indicator a pixel (See last comment). If someone feels strongly another direction, I don't mind updating it.

sergiou87:
Yes! That looks much better now, thanks 😄

sergiou87 (commented):
It works much better now :_D Thanks!!

Last thing I've noticed is that the caret of the popover is now overlapping the badge (whereas in your screenshots it's right below them):

image

sergiou87 (approved):
As discussed with @tidy-dev in Slack, we think this is in good shape to be merged since (1) this is only enabled for dev builds and (2) the only issue is that the caret of the popover is not positioned correctly. That issue can be addressed in a new PR.

sergiou87:
Found a last-minute issue, but still can be fixed later:
Kapture 2021-09-23 at 16 38 57

PR #12952, opened 2021-09-13
Prevent styles from high-contrast theme from leaking

tidy-dev:
## Description

There was some bleed from high-contrast border styles into non-high contrast themes. This moves high-contrast styles into their own css file and only applies them through use of a mixin instead of css variables. Making it so that when not on a high-contrast theme, there should be no new styles/variables applied.

### Screenshots

https://user-images.githubusercontent.com/75402236/133092469-7a8f73e5-8402-4417-a214-ed73a18c6f88.mov

## Release notes
Notes: [Fixed] High-contrast theme styles no longer bleed into regular themes.

### DIFF
--- a/app/styles/themes/_high-contrast.scss
+++ b/app/styles/themes/_high-contrast.scss
@@ -16,7 +16,7 @@ body.theme-high-contrast {
   --text-secondary-color: var(--hc-text-color);
 
   //border
-  --hc-hover-border: 2px solid var(--hc-border-color) !important;
+  --hc-hover-border: 2px solid var(--hc-border-color);
 
   /// tab-bar
   --tab-bar-active-color: var(--hc-active-item-color);
--- a/app/styles/themes/_high-contrast.scss
+++ b/app/styles/themes/_high-contrast.scss
@@ -1,73 +1,68 @@
 body.theme-high-contrast {
   // A small set of colors used to flatten ux and improve contrast
-  --hc-background: #090c11;
-  --hc-secondary-background: #323946;
-  --hc-border: #7a828e;
-  --hc-text: #f0f3f6;
-  --hc-active-item: #9ea7b3;
-  --hc-secondary-active-item: #596069;
-  --hc-active-text: #090c11;
+  --hc-background-color: #090c11;
+  --hc-secondary-background-color: #323946;
+  --hc-border-color: #7a828e;
+  --hc-text-color: #f0f3f6;
+  --hc-active-item-color: #9ea7b3;
+  --hc-secondary-active-item-color: #596069;
+  --hc-active-text-color: #090c11;
+  --hc-diff-add-border-color: green;
+  --hc-diff-delete-border-color: crimson;
 
   // background/text
-  --background-color: var(--hc-background);
-  --text-color: var(--hc-text);
-  --text-secondary-color: var(--hc-text);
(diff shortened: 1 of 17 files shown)


app/styles/themes/_high-contrast.scss
```diff
@@ -0,0 +1,73 @@
+body.theme-high-contrast {
+  // A small set of colors used to flatten ux and improve contrast
+  --hc-background: #090c11;
+  --hc-secondary-background: #323946;
+  --hc-border: #7a828e;
+  --hc-text: #f0f3f6;
+  --hc-active-item: #9ea7b3;
+  --hc-secondary-active-item: #596069;
+  --hc-active-text: #090c11;
```
niik: Absolutely fine with punting this to another PR but these should all be suffixed with `--color` to match the convention of the other variables.

niik (changes requested):
I think this is a great step in the right direction @tidy-dev! One thing I fear could be confusing though is the use of custom variables in the `_high_contrast.scss` file. When looking at a rule like this

https://github.com/desktop/desktop/blob/bcc96d22ac774ec085894a62deb1fb10a1d4673f/app/styles/ui/_tab-bar.scss#L64-L71

As a casual reader of the code there's no easy way for me to know that `--tab-bar-hover-border` isn't globally available (as in it's not defined in `_variables.scss`) whereas `--tab-bar-hover-background-color` is.

So I think we should leave `_high_contrast.scss` to **only redefine variables which are already defined** (like in _dark.scss) but not to introduce new variables unless necessary (the `--hc-*` variables are a good example of this) and in cases like the above either set the border to `2px solid var(--hc-border)` or  introduce a `--hc-hover-border` variable if it's going to be used frequently. Having a `--hc` prefix coupled with the mixin will make it clearer that it's a one-off thing specific to high contrast.

The thinking around themes when we first introduced them was that `_variables.scss` would contain fallback variables for all themes.

app/styles/themes/_high-contrast.scss
tidy-dev: done

app/styles/themes/_high-contrast.scss
```diff
@@ -0,0 +1,68 @@
...
+  --hc-diff-add-border-color: green;
+  --hc-diff-delete-border-color: crimson;
+
+  // background/text
+  --background-color: var(--hc-background-color);
+  --text-color: var(--hc-text-color);
+  --text-secondary-color: var(--hc-text-color);
+
+  //border
+  --hc-hover-border: 2px solid var(--hc-border-color) !important;
```
niik: I'd rather we didn't codify specificity into the variables. And in this case I don't believe it's necessary? If it is we should move it to the specific place where it is or we'll taint all rules which include this variable with the highest specificity and then it's an `!important` war

```suggestion
  --hc-hover-border: 2px solid var(--hc-border-color);
```

niik (changes requested):
Awesome work @tidy-dev once tiny request and then I think we're good!

app/styles/themes/_high-contrast.scss
tidy-dev: Makes sense. Seems I remember needed it for the tab bars.. but now I can't reproduce needing it so I just removed it.

niik (approved):
✨

PR #12964, opened 2021-09-16
Reload diff when app regains focus

### LINKED ISSUE #12962: View not refreshing when active file changes
### Describe the bug
- The "diff" view is not refreshing when the file being displayed is altered on disk.
- The "diff" view is not refreshing after selected (cherry-picked) lines are committed.

### Version & OS

2.9.4-beta2 (arm64) on macOS 11.6

1. Select a file so it appears in the "diff" view.
2. Go to another app, edit, and save changes.
3. Return to GitHub Desktop.

-- or --

2. Select some changed lines in the "diff" view for cherry-picking.
3. Create a commit with the selected lines.

### Expected behavior

- In the first case, the "diff" view should update to show the new changes.
- In the second case, the "diff" view should update to hide the committed changes.

### Actual behavior

- The view stays the same, so you must select a different file and then switch back to force an update.

### Additional context

- This has only started to occur in this specific update, which fixed the layout issues.

niik:
Closes #12962

## Description

As noted in #12962 the otherwise wonderful changes we made in https://github.com/desktop/desktop/pull/12903 had the unintended side-effect of not reloading diffs when the application loses and then regains focus. This PR makes us consider changed between the previous and current diff (the diff between the diff and the diff or the diff diff if you will) as a condition for diff reload in the seamless diff switcher.

## Release notes

Notes: [Fixed] Refresh diffs when application receives focus

### DIFF
--- a/app/src/ui/diff/seamless-diff-switcher.tsx
+++ b/app/src/ui/diff/seamless-diff-switcher.tsx
@@ -120,6 +120,15 @@ function isSameFile(prevFile: ChangedFile, newFile: ChangedFile) {
   return prevFile === newFile || prevFile.id === newFile.id
 }
 
+function isSameDiff(prevDiff: IDiff, newDiff: IDiff) {
+  return (
+    prevDiff === newDiff ||
+    (isTextDiff(prevDiff) &&
+      isTextDiff(newDiff) &&
+      prevDiff.text === newDiff.text)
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
Looks great to me!! Thanks for the quick fix ❤️

PR #12965, opened 2021-09-16
Only consider invalid tokens when 401s come from GH(E)

### LINKED ISSUE #12943: GitHub Desktop forgets the GitHub Enterprise server URL
### Describe the bug
Our GitHub Enterprise server is behind the company firewall and only accessible via VPN.

With the recent versions of GH Desktop (Since 2-3 weeks ago approx), every time I log off of VPN, GH desktop forgets my enterprise URL and I have to type it again.

While its expected that GH won't be able to communicate with the enterprise server when its not accessible, having to manually type in the URL every time I reconnect to VPN is tedious - And was not required with previous versions of the desktop tool.

### Version & OS

Version 2.9.3 (x64)

Mac OS Big Sur

1. Confirm GH Enterprise server is configured in preferences:

image

2. Disconnect from VPN so that GH Enterprise is not available to GH desktop.

3. Try to fetch origin, error message is shown (correctly) since GH Enterprise is not reachable.

4. You get an error that the account token is invalidated. This is actually not entirely accurate - Account token is still valid, just that the service is not reachable:

image
 
5. Click "No" (Not on VPN now, no need to try again)

6. Connect to VPN again - GH Enterprise is reachable

7. Click "Fetch Origin" again - Error message shows up. Click "Open Preferences"

8. GH Enterprise is completely forgotten, need to type the URL again to use GH desktop:

image

image

### Expected behavior

I should be able to use GH Desktop with and without access to GH Enterprise for local development, and not have to re-type the Enterprise URL every time.

### Actual behavior

Need to provide the GH enterprise URL on every connection to VPN.

### Screenshots

Screenshots added above.

### Logs
[2021-09-10.desktop.production.log](https://github.com/desktop/desktop/files/7145152/2021-09-10.desktop.production.log)

sergiou87:
Closes #12943 

## Description

It seems company gateways/proxies might give a 401 error in certain circumstances when users attempt to access GHE instances (in the issue above, when the user tries to access GHE repos from out of the company VPN).

This PR adds an additional check to make sure the 401 we received comes from GH(E) by looking for the `X-GitHub-Request-Id` header, and only in that scenario the app will consider the token invalid.

## Release notes

Notes: [Fixed] Tokens are not considered invalid when accessing GitHub Enterprise from outside of the company's VPN

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1058,7 +1058,13 @@ export class API {
       options.reloadCache
     )
 
-    if (response.status === 401) {
+    // Only consider invalid token when the status is 401 and the response has
+    // the X-GitHub-Request-Id header, meaning it comes from GH(E) and not from
+    // any kind of proxy/gateway. For more info see #12943
+    if (
+      response.status === 401 &&
+      response.headers.has('X-GitHub-Request-Id')
+    ) {
       API.emitTokenInvalidated(this.endpoint)
     }
 
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1061,9 +1061,12 @@ export class API {
     // Only consider invalid token when the status is 401 and the response has
     // the X-GitHub-Request-Id header, meaning it comes from GH(E) and not from
     // any kind of proxy/gateway. For more info see #12943
+    // We're also not considering a token has been invalidated when the reason
+    // behind a 401 is the fact that any kind of 2 factor auth is required.
     if (
       response.status === 401 &&
-      response.headers.has('X-GitHub-Request-Id')
+      response.headers.has('X-GitHub-Request-Id') &&
+      !response.headers.has('X-GitHub-OTP')
     ) {
       API.emitTokenInvalidated(this.endpoint)
     }


app/src/lib/api.ts
```diff
@@ -1058,7 +1058,13 @@ export class API {
       options.reloadCache
     )
 
-    if (response.status === 401) {
+    // Only consider invalid token when the status is 401 and the response has
+    // the X-GitHub-Request-Id header, meaning it comes from GH(E) and not from
+    // any kind of proxy/gateway. For more info see #12943
+    if (
+      response.status === 401 &&
+      response.headers.has('X-GitHub-Request-Id')
```
niik: This is slightly outside of the scope of this PR but I wonder if we should exclude the OTP response here as that's the only other response which will return a 401 while not for that matter meaning that the token should be cleared. See

https://github.com/desktop/desktop/blob/3da974a2ee54c76d5a8931caf9b42f90ede155fc/app/src/lib/api.ts#L1226-L1227

```suggestion
      response.headers.has('X-GitHub-Request-Id') &&
      !response.headers.has('X-GitHub-OTP')
```

niik (changes requested):
This is great @sergiou87, I think this will fix a lot of false positives! Just one question which  you can feel free to ignore and tackle in a separate PR/issue if you want

app/src/lib/api.ts
sergiou87: Sounds great, thanks!! I don't think it's completely mandatory, since when this happens there is no account to sign in again, and this check will prevent the prompt entirely: https://github.com/desktop/desktop/blob/3da974a2ee54c76d5a8931caf9b42f90ede155fc/app/src/lib/stores/app-store.ts#L539-L543

However, it'd be semantically more correct, so I think we should include it!

niik (approved):
:sparkles:

PR #12966, opened 2021-09-16
Tag creation enhancement

### LINKED ISSUE #12509: Need create tag name more friendly

mahezsh:
This PR Fixes #12509 

## Description

This shows the last 3 tags within the **Create Tag** component. 
It also filters the shown tags as the user type the tag name. (as suggested by @sergiou87)

### Screenshots

screenshot

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/create-tag/create-tag-dialog.tsx
+++ b/app/src/ui/create-tag/create-tag-dialog.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import _ from 'lodash'
 
 import { Repository } from '../../models/repository'
 import { Dispatcher } from '../dispatcher'
@@ -28,6 +29,7 @@ interface ICreateTagState {
    * shown in its place.
    */
   readonly isCreatingTag: boolean
+  readonly previousTags: Array<string> | null
 }
 
 const MaxTagNameLength = 245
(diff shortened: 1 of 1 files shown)


chrisgdavies (approved):

app/src/ui/create-tag/create-tag-dialog.tsx
```diff
@@ -43,12 +45,14 @@ export class CreateTag extends React.Component<
...
       tagName: props.initialName || '',
       isCreatingTag: false,
+      previousTags: this.getExistingTagsFiltered(),
     }
   }
 
   public render() {
     const error = this.getCurrentError()
     const disabled = error !== null || this.state.tagName.length === 0
+    const lastThreeTags = _.takeRight(this.state.previousTags, 3)
```
say25, line 55:
```suggestion
    const lastThreeTags = this.state.previousTags.slice(0, 3)
```
LoDash isn't needed for this operation from what I can tell

sergiou87: But `slice` will take the elements from the left, no? `_.takeRight` will return the 3 elements at the end, while `.slice(0,3)` will return the first 3 elements 🤔

say25: I agree but a couple things
1. Why do we want the last 3 items to begin with? Is this because they are the oldest? Should we sort by something like creation time?
2. We can always use `.slice(-3)` to get the last 3 items

sergiou87: 1. They are supposed to be the last 3 (the newest)
2. Good point!

I think I'll merge this tomorrow, address the `.slice` thing and hide this behind a feature flag 😄  The team agreed today that this should be ready to go 🚀

sergiou87 (approved):

PR #12979, opened 2021-09-17
Show popover explaining why partial commits isn't working

### LINKED ISSUE #12862: Inform user that selecting individual lines to commit isn't possible.

Enabling `Hide Whitespace Changes` prevents you from selecting individual lines to commit.
There is information about this in the UI where the feature is enabled, but that is not enough in my opinion.
There is a big chance you don't remember that little piece of information, and also there is no indicator that `Hide Whitespace Changes` is enabled.

### Proposed solution

Show a popup or similar (with a `Sorry, you can't select individual lines while Hide Whitespace Changes is enabled.` message) when trying to select individual lines & hide whitespace is enabled.

niik:
Closes https://github.com/desktop/desktop/issues/12862

## Description

Show a popover when users attempt to toggle lines in the changes view while having the option to hide whitespace changes turned on.

## TODO

- [x] Unified diffs
- [x] What to do about positioning when the diff container is tiny (i.e. one line files) or when the gutter is at the very bottom of the file

## Release notes

Notes: [Improved] Show a message explaining why line selection is disabled when hiding whitespace #12862


tidy-dev (changes requested):
When I was testing it out, I noticed that the side-by-side diff can sometimes cause the scroll height to suddenly jump when clicking one near the bottom of the view port. Was wondering if it would feasible to do what you did for unified diffs and flip the popup?

https://user-images.githubusercontent.com/75402236/134358162-36aaec9e-fca4-412b-9a36-6023ae053fec.mov

.

app/styles/ui/_popover.scss
```diff
@@ -12,105 +12,173 @@
...
+
+  &.appear-shake {
+    animation: popover-shake 0.15s both;
+    transform: translate3d(0, 0, 0);
+
+    // This is a fairly jarring animation so let's make sure we don't
+    // run it for people who don't want excess motion.
+    @media (prefers-reduced-motion) {
+      animation: none;
+    }
```
sergiou87, line 63: I love this ❤️  (also TIL)

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -548,6 +599,7 @@ export class SideBySideDiffRow extends React.Component<
 
   private onClickHunk = () => {
     if (this.props.hideWhitespaceInDiff) {
+      this.setState({ showWhitespaceHint: DiffColumn.Before })
```
sergiou87: Not sure how hard is it to fix this, but if the hunk is all in the `After` column, the popover looks a bit weird 😅 

image

app/src/ui/diff/text-diff.tsx
```diff
@@ -1381,10 +1390,88 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
+      // Offset down by 10px to align the popover arrow.
+      container.style.top = `${lineY - 10}px`
+
+      // If the line is further than 50% down the viewport we'll flip the
+      // popover to point upwards so as to not get hidden beneath (or above)
+      // the scroll boundary.
+      if (lineY - scroller.scrollTop > scroller.clientHeight / 2) {
+        caretPosition = PopoverCaretPosition.LeftBottom
+        style.bottom = -35
+      }
```
sergiou87, line 1441: This is a nice touch! I see it's not in the side-by-side diff, which instead pushes the scroll down:
Kapture 2021-09-22 at 17 37 33

I don't think we _need_ to add complexity for such an edge case but just wanted to let you know!

app/src/ui/diff/text-diff.tsx
```diff
@@ -1381,10 +1390,88 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
+    }
+
+    if (this.whitespaceHintContainer !== null) {
+      ReactDOM.unmountComponentAtNode(this.whitespaceHintContainer)
+    }
+
+    if (this.whitespaceHintContainer !== null) {
+      this.whitespaceHintContainer.remove()
+      this.whitespaceHintContainer = null
+    }
```
sergiou87, line 1476: Is there a reason why this is in 2 different `if`? 🤔 
```suggestion
    if (this.whitespaceHintContainer !== null) {
      ReactDOM.unmountComponentAtNode(this.whitespaceHintContainer)
      this.whitespaceHintContainer.remove()
      this.whitespaceHintContainer = null
    }
```

sergiou87 (commented):
Looks awesome overall! Great job! 😍 

Only found a few nits I don't feel super strongly about, let me know what you think!

app/src/ui/diff/text-diff.tsx
niik: > Is there a reason why this is in 2 different `if`?

Yes! And the fault is mine, I should have documented it 🤦 So `unmountComponentAtNode` can actually (or will even) end up calling `unmountWhitespaceHint` since the `componentDidUnmount` callback in `Popover` will be invoked and in turn call `onDismissed`. So depending on who ended up calling `unmountWhitespaceHint` this can end up being reentrant which means that `this.whitespaceHintContainer` will be `null` the `unmountComponentAtNode` call.

I'll add some comments here to explain this weirdness

app/src/ui/diff/text-diff.tsx
niik: Yeah, I looked into this and I'll reply to both your and @tidy-dev's comments on this here. Due to the way the `List` component works (hiding overflow) we can't easily have the popover extend above the diff so there would always be scenarios where we'd extend the height of the diff (single line/smalls diffs) so I figured we'll live with this annoyance for now and see if we can't fix it by means of a more flexible popover system in the future.

app/src/ui/diff/side-by-side-diff-row.tsx
niik: Agreed, that does look weird. I've updated the logic to prefer the left hand side except in the case of added lines like the one above in https://github.com/desktop/desktop/pull/12979/commits/e8d57a25e5b56c31a6e9b249c89d1b4d18d486c8

app/src/ui/diff/text-diff.tsx
niik: https://github.com/desktop/desktop/pull/12979/commits/6e6a1f0aaf443b10d505a57bd7b2981b400484fa

sergiou87 (approved):
Yay!! Thanks for those changes and comments 😄 

I tested it again and works wonderfully, great work!!! 💖 :shipit:

tidy-dev (approved):
✨

PR #12982, opened 2021-09-18
Expand/Collapse Organisations in "Clone Repository" List

### LINKED ISSUE #11908: Selectively Ignore Organisations in "Clone Repository" List

When a user is a member of multiple organisations each with lots of repositories, the list shown in the "Clone Repository" window on the GitHub.com tab can become very large. I realise it is possible to filter this by searching, but when there are lots of repositories it can take a while to populate the list to search within.

There are times when it is desireable to ignore certain organisations in this list, either permanently, or temporarily. This would allow reducing the amount of respositories which need to be listed.

### Proposed solution

I envisage a solution whereby rather than having a flat list which shows all repositories from all organisations, instead the list is more of a tree-view style.

Still grouped by organisations, a setting could be added to choose whether to expand the list fully by default (current behaviour), or have organisations collapsed by default (or even remember which ones were collapsed and which ones were expanded).

Then loading of the list need only to populate repositories in groups which are expanded. If a collapsed organisation needs to be viewed, it can be expanded and the list populated on demand.

UI-wise it would be a relatively minor change, something like the doodle below. The grouping by organisation already exists, so it would just be a matter of making it collapsable.

image

### Additional context

My main example of this is when using GitHub Classrooms for teaching. I'm a member of two organisations (different modules) which are used for storing student repositories. With the large class sizes involved and multiple assignments, this has lead to there being an excessively large number of repositories which are parsed by the clone repository window. In my case there are about 1400 repositories listed!

In the GitHub Desktop window for cloning repositories, it takes about 2 minutes for this list of repositories to be populated as a result, because it needs to query and list thousands of student repositories, which then clutter up the list. It would be much easier if I could selectively disable certain organisations from being shown in the list.

TCWORLD:
Closes #11908 

## Description

When a user is a member of multiple organisations each with lots of repositories, the list shown in the "Clone Repository" window on the GitHub.com tab can become very large. I realise it is possible to filter this by searching, but when there are lots of repositories it can take a while to populate the list to search within - in cases with thousands of repos, this can take several minutes making the window practically unusable.

There are times when it is desireable to ignore certain organisations in this list, either permanently, or temporarily. This would allow reducing the amount of respositories which need to be listed.

As the repositories are already grouped by organisation, from a UI standpoint this becomes a matter of making each group collapsable, showing initially only the users own repositories and collapsed groups for their organisation membership. This reduces the loading time considerably because organisations are not queried.

Once a user expands an organisation group by clicking on it (or using space/enter with it selected as per standard UI), then if this is the first time the organisation is viewed, it will only then query Github for the repo list for that org. Once queried the user can then expand/collapse an organisation at will without requerying the list (nice to keep the list free from clutter).

### Screenshots

The PR makes a minor change to the UI on the clone repository tab to accomadate the collapsing/expansion of each organisation group. The only real change is that organisations now have a plus or dash Octicon next to them depending on whether expanded or contracted. This is shown in the image below.

Clone repos tab with collapsable groups

## Release notes

Notes:


sergiou87:
Hi @TCWORLD 👋 

Thank you for all your work on this. However, we don't accept PRs for features that haven't had the opportunity to be discussed in an issue and where someone from the Desktop team gives the 👍  to go ahead and create a PR for it. It'd be better to continue gathering feedback and discussing this in #11908 before working on any specific implementation.

Sorry about that! 😳

TCWORLD:
Well given somebody at GitHub suggested they had the issue, feel free to go about discussing it.
In the mean time, the work is done, and the branch will just sit there. Myself and three other people I work with will simply have to use the version I've built.

PR #12998, opened 2021-09-21
Update CodeQL workflow using best practices and reduce token permissions

niik:
## Description

I noticed this warning while looking at our workflow runs

To address that I've updated the CodeQL workflow based on the sample in https://github.com/github/codeql-action/blob/main/README.md#usage.

Additionally I found out about the `permissions` key which lets us reduce the power of the `GITHUB_TOKEN`. Since we don't use it for any write actions (that I know of) I'm turning it into a read only token that we can make more permissive should we need to.

## Release notes

Notes: no-notes


sergiou87 (approved):
LGTM! :shipit:

PR #13004, opened 2021-09-22
Fix missing initial syntax highlighting for simple unified diffs

niik:
## Description

Spotted this while working on #12979. For non-expandable diffs 

### Steps to reproduce

1. Create a straightforward diff which will be the first selected in the repository, like a new file called `a.js` containing `function() {}`.
2. Reload or launch the app

Expected results
The file is syntax highlighted

Actual results
It's not because the file contents are loaded asynchronously by the `SeamlessDiffSwitcher` but there's not logic in `TextDiff` to update syntax highlighting when file contents change (in this case "change" means "are loaded").

## Release notes

Notes: [Fixed] Syntax highlighting is once again applied when viewing a small change for the first time

### DIFF
--- a/app/src/ui/diff/text-diff.tsx
+++ b/app/src/ui/diff/text-diff.tsx
@@ -623,6 +623,7 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
 
   private getAndStoreCodeMirrorInstance = (cmh: CodeMirrorHost | null) => {
     this.codeMirror = cmh === null ? null : cmh.getEditor()
+    this.initDiffSyntaxMode()
   }
 
   private onContextMenu = (instance: CodeMirror.Editor, event: Event) => {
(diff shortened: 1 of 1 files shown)


app/src/ui/diff/text-diff.tsx
```diff
@@ -623,6 +623,7 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
 
   private getAndStoreCodeMirrorInstance = (cmh: CodeMirrorHost | null) => {
     this.codeMirror = cmh === null ? null : cmh.getEditor()
+    this.initDiffSyntaxMode()
```
niik, line 626: This is technically unrelated to the bug at hand but when debugging I noticed that `onSwapDoc` would trigger before we got the editor ref, making its call to `initDiffSyntaxMode` a noop.

sergiou87 (approved):
Nice catch!

PR #13010, opened 2021-09-23
Scroll diff to top if we switched to a new file

### LINKED ISSUE #12980: Scroll position is shared across files
### Describe the bug
A repository often has several changed files. When I click on a changed file in order to see its diff, the scrollbar position varies according to the previous diff I was looking at.

### Version & OS

Version 2.9.3, Arch Linux x64.

1. Have a repository with several changed files, where the changes in each file are big enough that the diff needs to use scroll bars.
2. Click on a changed file and scroll the diff down to the bottom.
3. Click on another changed file.

### Expected behavior

The diff should be scrolled back to the top after clicking on a different changed file. Ideally (but I can live without it): if that diff had already been viewed before, the scroll position should be restored from that time.

### Actual behavior

The second diff is scrolled near the bottom because the previous one also was scrolled to the bottom.

### Screenshots

1. First diff (scrollbar is at the top).
Screenshot_20210917_230500
2. Scroll it to the bottom.
Screenshot_20210917_230522
3. After clicking on a different changed file, the diff is still scrolled to the bottom (instead of being at the top).
Screenshot_20210917_230537
### Logs

N/A.

### Additional context

None.

sergiou87:
Closes #12980

## Description

This PR scrolls the diff to the top when switching between files. I tried to save the scroll position per file and restore it when needed, but it was a minefield, especially given we have 2 completely different implementations of the diff viewer. For example, not only it's resulting difficult to restore the scroll position after the cells have been resized and rendered, but there are also some situations (resizing and expanding diffs) where the stored position won't be useful.

Because of that, I decided to just fix the most annoying part for now 😅 

## Release notes

Notes: [Fixed] Diff are scrolled to the top when switching between files.

### DIFF
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -236,6 +236,14 @@ export class SideBySideDiff extends React.Component<
         diff: this.props.diff,
       })
     }
+
+    // Scroll to top if we switched to a new file
+    if (
+      this.virtualListRef.current !== null &&
+      this.props.file.id !== prevProps.file.id
+    ) {
+      this.virtualListRef.current.scrollToPosition(0)
+    }
   }
 
   private canExpandDiff() {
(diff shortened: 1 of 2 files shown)


tidy-dev (approved):
✨  Makes sense. Works well

PR #13024, opened 2021-09-27
Filesize error improvement

### LINKED ISSUE #10242: File size restriction error message should tell what file is causing the issue

ADustyOldMuffin:
Is a redo of #12540 since I ruined the git history.

Closes #10242 

## Description

- This will put the offending file name in the error message 

### Screenshots

image

## Release notes

Notes: Added offending file name to the file exceeds size limit error.

### DIFF
--- a/app/src/lib/git/core.ts
+++ b/app/src/lib/git/core.ts
@@ -18,6 +18,7 @@ import { isErrnoException } from '../errno-exception'
 import { ChildProcess } from 'child_process'
 import { Readable } from 'stream'
 import split2 from 'split2'
+import { getFileFromExceedsError } from '../helpers/regex'
 import { merge } from '../merge'
 import { withTrampolineEnv } from '../trampoline/trampoline-environment'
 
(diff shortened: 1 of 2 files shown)


ADustyOldMuffin:
@sergiou87 👋

sergiou87:
Hey @ADustyOldMuffin! 👋  Thanks for updating the branch!

Replying to your question from the [previous PR](https://github.com/desktop/desktop/pull/12540#issuecomment-927505916):
> Sorry for the super late reply! (Life) I'm curious on any tips as to how I might go about doing something along those lines since this is returned as just text. I'll continue to look into something of this nature.

Hmmm I see what you mean… one option is to add a `files` to the `GitError` class, and then print that separately. That'd be tricky too because you'd need to pass not only the filename and path, but also the file size.

I don't think it's worth the effort, so we can probably go ahead with your current implementation 😄

sergiou87 (approved):
Looks good! Tested it and worked as expected, thanks for this!! ❤️

billygriffin:
Nice improvement, thank you!

PR #13028, opened 2021-09-27
Check Run Logs

tidy-dev:
## Description

The focus on this PR is about retrieving and being able to output the logs (utilizing jzip to unpack the zip file). In doing so I also added the collapsibility of the check run app title bar and the View in GitHub button.  

This is still feature flagged to development as the final UI is still to be determined. Also, there is a security concern in current implementation as the html from non-actions logs are not sanitized.

Known things for upcoming PRs:
- Create a log output component 
  - Sanitize html from non-actions logs
  - Parse actions logs for better reading/scanning
  - Pretty up log output designs.. make consistent log appearance, view on GitHub button less prominent.
- Be nice to get loading image for logs (not copy PR one)
- Bug: Currently relative position impacts other toolbar components, but absolute positioning needs additional logic due to fullscreen mode changing relative height to top of screen.
- Bug: When I reload in a local Desktop, the PR badge sometimes doesn't show the check run status even tho the check runs call has been made. Similarly, if the status wasn't there and then loads, the check run dropdown still won't open sometimes... I think I may have a timing/race condition to figure out. :/ 

## Release notes

Notes: no-notes


app/src/lib/api.ts
```diff
@@ -361,6 +365,48 @@ export interface IAPIRefCheckRuns {
   readonly check_runs: IAPIRefCheckRun[]
 }
 
+interface IAPIWorkflowRuns {
+  readonly total_count: number
+  readonly workflow_runs: IAPIWorkflowRun[]
```
sergiou87: Any reason for not using `ReadonlyArray`?
```suggestion
  readonly workflow_runs: ReadonlyArray<IAPIWorkflowRun>
```

app/src/lib/api.ts
```diff
@@ -361,6 +365,48 @@ export interface IAPIRefCheckRuns {
...
+  readonly jobs_url: string
+  readonly logs_url: string
+  readonly name: string
+  readonly rerun_url: string
+  readonly check_suite_id: number
+}
+
+interface IAPIWorkflowJobs {
+  readonly total_count: number
+  readonly jobs: IAPIWorkflowJob[]
```
sergiou87, line 389: Any reason for not using `ReadonlyArray`?
```suggestion
  readonly jobs: ReadonlyArray<IAPIWorkflowJob>
```

app/src/lib/api.ts
```diff
@@ -960,6 +1006,77 @@ export class API {
...
+    const path = `repos/${owner}/${name}/actions/runs?event=pull_request&branch=${branchName}`
+    const customHeaders = {
+      Accept: 'application/vnd.github.antiope-preview+json',
+    }
+    const response = await this.request('GET', path, { customHeaders })
+    try {
+      return await parsedResponse<IAPIWorkflowRuns>(response)
+    } catch (err) {
+      log.debug(
+        `Failed fetching workflow runs for  ${branchName} (${owner}/${name})`
```
sergiou87: Nit affecting my OCD 😆 
```suggestion
        `Failed fetching workflow runs for ${branchName} (${owner}/${name})`
```

app/src/lib/api.ts
```diff
@@ -960,6 +1006,77 @@ export class API {
...
+    const response = await this.request('GET', logsUrl, {
+      customHeaders,
+    })
+
+    try {
+      const zipBlob = await response.blob()
+      return new JSZip().loadAsync(zipBlob)
+    } catch (e) {
+      // Sometimes a workflow provides a log url, but still returns a 404
+      // because a log file doesn't makes sense for the workflow. Thus, we just
```
sergiou87: Nit: typo
```suggestion
      // because a log file doesn't make sense for the workflow. Thus, we just
```

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
       }
     })
   }
+
+  /**
+   * Retrieve GitHub Actions workflows, jobs, and logs for the branch and apply
+   * logs to matching check run output.
+   */
+  public async getLatestPRWorkflowRunsLogsForCheckrun(
```
sergiou87: Nit: typo
```suggestion
  public async getLatestPRWorkflowRunsLogsForCheckRun(
```

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+    )
+
+    if (latestWorkflowRuns.length === 0) {
+      return checkRuns
+    }
+
+    const logHash = new Map<string, JSZip>()
+    const mappedCheckRuns = new Array<IRefCheck>()
+    for (let i = 0; i < checkRuns.length; i++) {
+      const cr = checkRuns[i]
```
sergiou87: Why not just a "for each"? I don't see `i` being used 🤔 
```suggestion
    for (const cr of checkRuns.length) {
```

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+      const matchingJob = workFlowRunJobs?.jobs.find(j => j.name === cr.name)
+      if (matchingJob === undefined) {
+        mappedCheckRuns.push(cr)
+        continue
+      }
+
+      // One workflow can have the logs for multiple check runs.. no need to
+      // keep retrieving it. So we are hashing it.
+      const logZip =
+        logHash.get(matchingWR.name) ??
```
sergiou87: Is `name` safe enough to use here as a unique identifier? I've seen workflow runs have an `id` field so maybe it's better to use that? 🤔 

Or even better, why not use `logs_url` directly? It should always be the same, right?

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+      const cr = checkRuns[i]
+      const matchingWR = latestWorkflowRuns.find(
+        wr => wr.check_suite_id === cr.checkSuiteId
+      )
+      if (matchingWR === undefined) {
+        mappedCheckRuns.push(cr)
+        continue
+      }
+
+      const workFlowRunJobs = await api.fetchWorkflowRunJobs(matchingWR)
```
sergiou87: Should we cache this per check suite id? At least in this function call, as you do with `logHash`.

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+    )
+
+    if (allBranchWorkflowRuns === null) {
+      return []
+    }
+
+    // When retrieving Actions Workflow runs it returns all present and past
+    // workflow runs for the given branch name. For each workflow name, we only
+    // care about showing the latest run.
+    allBranchWorkflowRuns.workflow_runs.forEach(wr => {
```
sergiou87: Given you don't need the index of each element, is there a reason for not using `for( … of …)` here again? In theory [it should perform better than `forEach`](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/#arrayforeach-vs-for-and-forof) and it is, IMHO, more readable.

Also if you make this change, beware of the `return` you have a few lines below that must be changed to `continue` if I'm not wrong.
```suggestion
    for (const wr of allBranchWorkflowRuns.workflow_runs) {
```

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+      const storedWR = wrMap.get(wr.name)
+      if (storedWR === undefined) {
+        wrMap.set(wr.name, wr)
+        return
```
sergiou87: If you 👍  my suggestion above, this needs to be changed to `continue`
```suggestion
        continue
```

EDIT: I just noticed I wrote `break` here instead of `continue` as I did in my previous message 🤦

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+
+    if (allBranchWorkflowRuns === null) {
+      return []
+    }
+
+    // When retrieving Actions Workflow runs it returns all present and past
+    // workflow runs for the given branch name. For each workflow name, we only
+    // care about showing the latest run.
+    allBranchWorkflowRuns.workflow_runs.forEach(wr => {
+      const storedWR = wrMap.get(wr.name)
```
sergiou87: Same question as in the method above: is `name` the best field to cache by? Better than `id`?

app/src/lib/stores/commit-status-store.ts
```diff
@@ -380,6 +417,154 @@ export class CommitStatusStore {
...
+): Promise<ReadonlyArray<IAPIWorkflowJobStep>> {
+  try {
+    const jobFolder = logZip.folder(job.name)
+    if (jobFolder === null) {
+      return job.steps
+    }
+
+    const stepsWLogs = new Array<IAPIWorkflowJobStep>()
+    for (let j = 0; j < job.steps.length; j++) {
+      const step = job.steps[j]
```
sergiou87:
```suggestion
    for (const step of job.steps) {
```

app/src/ui/branches/ci-check-run-list.tsx
```diff
@@ -40,16 +43,47 @@ export class CICheckRunList extends React.PureComponent<
...
+      loadingLogs: true,
+    }
+
+    this.onStatus(combinedCheck)
+  }
+
+  public componentDidUpdate(prevProps: ICICheckRunListProps) {
+    // Re-subscribe if we're being reused to show a different status.
+    if (
+      this.props.repository !== prevProps.repository ||
```
sergiou87: Shouldn't we check the hash here? 🤔 
```suggestion
      this.props.repository.hash !== prevProps.repository.hash ||
```

sergiou87 (changes requested):
Looks good overall! I mainly found a few nits, and also a couple of things that look suspicious but I might be missing something.

app/src/lib/api.ts
tidy-dev: Original solution I added jobs to existing object - but refactor to not do that. Thanks!

app/src/ui/branches/ci-check-run-list.tsx
tidy-dev: Hmm. Yes. I copied that from the badge output component.. Likely should be the hash there as well.

app/src/lib/stores/commit-status-store.ts
tidy-dev: Done. Good point. Thanks!

app/src/lib/stores/commit-status-store.ts
GetBentGaming: Ok

app/src/ui/branches/ci-check-run-list.tsx
GetBentGaming: I'm really not sure

sergiou87 (approved):
Thanks for those changes! 🙌  ❤️ 

I tested it and all worked as expected 😄 :shipit:

MadMena (approved):

GetBentGaming:
Im New to Github But Smart, Somewhat, Any help i greatly appreciate

PR #13064, opened 2021-10-05
Moving the retrieval of actions workflows sooner for better performance

tidy-dev:
## Description
For logs output, there are 2 sources. The check runs call has an `output` property for non-actions check runners and if a check run is also an action's workflow. We must additionally call the jobs endpoint and the logs endpoint to obtain the logs. We already do the check run call for each PR so we can show the check runs overall status in the pull request dropdown list and in the pull request badge when a pull request branch is checked out. 

The only way to know if a check run has an action workflow is to retrieve the action workflows for a given branch (the pull request branch). Since we don't want to have the expense of this call for all pull requests when the user may not even care about most of the PR logs, my current implementation waits until the user clicks on the PR badge and then does the actions workflow retrieve for that pr branch and and if the check runs have actions workflows it subsequently does the jobs and logs retrieve... But, this means that check runs that don't have action workflows, show loading when we have retrieved all the actions workflows, jobs, and logs when we in reality already have their output.

In this PR, when we subscribe to the check runs in the PR Badge, I set a `checkedOutRef` property. Thus, when we loop through PRs to retrieve the check runs. I can go ahead and do the actions workflow call for the PR (and just get the logs_url and jobs_url) that is currently checked out. Technically this really is the last checked out PR because I do not bother to clear this when switching to a non-pr branch, but this could be added if we feel this call is two high a cost. 

Now, if you open the PR Badge, it will start retrieving the logs for each of the PR's with action workflow and go ahead and show the `output` for those that are not check-runs. Resulting in a smoother experience on the non-action check runs.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/commit-status-store.ts
+++ b/app/src/lib/stores/commit-status-store.ts
@@ -18,6 +18,7 @@ import {
 import { IDisposable, Disposable } from 'event-kit'
 import { setAlmostImmediate } from '../set-almost-immediate'
 import JSZip from 'jszip'
+import { enableCICheckRuns } from '../feature-flag'
 
 /**
  * A Desktop-specific model closely related to a GitHub API Check Run.
@@ -38,6 +39,8 @@ export interface IRefCheck {
   readonly checkSuiteId: number | null // API status don't have check suite id's
   readonly output: IRefCheckOutput
   readonly htmlUrl: string | null
+  readonly jobs_url?: string
+  readonly logs_url?: string
 }
 
 /**
@@ -186,6 +189,17 @@ export class CommitStatusStore {
 
   private backgroundRefreshHandle: number | null = null
   private refreshQueued = false
+
+  /**
+   * The repositories currently checked out ref and branch name. This is used to
+   * determine if we retrieve the action workflows to see if they exist for the
+   * checkruns, which we currently only do if the we are retrieve a status for a
+   * ref that is currently checked out - not for all the pull requests.
+   */
+  private checkedOutRef: {
+    ref: string
+    branchName: string
+  } | null = null
 
   /**
    * A map keyed on the value of `getCacheKey` containing one object
@@ -350,7 +364,25 @@ export class CommitStatusStore {
       checks.push(...latestCheckRunsByName.map(apiCheckRunToRefCheck))
     }
 
-    const check = createCombinedCheckFromChecks(checks)
+    let checksWithActionsInfo: ReadonlyArray<IRefCheck> | undefined
+    if (
+      enableCICheckRuns() &&
+      this.checkedOutRef !== null &&
+      this.checkedOutRef.ref === ref
+    ) {
+      const actionWorkflows = await this.getLatestPRWorkflowRuns(
+        api,
+        owner,
+        name,
+        this.checkedOutRef.branchName
+      )
+      checksWithActionsInfo = this.getCheckRunWithActionsJobAndLogURLs(
+        checks,
+        actionWorkflows
+      )
+    }
+
+    const check = createCombinedCheckFromChecks(checksWithActionsInfo ?? checks)
     this.cache.set(key, { check, fetchedAt: new Date() })
     subscription.callbacks.forEach(cb => cb(check))
   }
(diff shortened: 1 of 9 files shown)


app/src/ui/branches/ci-check-list-item-logs.tsx
```diff
@@ -0,0 +1,179 @@
+import * as React from 'react'
```
tidy-dev, line 1: This really isn't new stuff - I just moved the log display portion of the check-list-item component into it's own component.

app/src/lib/stores/commit-status-store.ts
```diff
@@ -187,6 +190,17 @@ export class CommitStatusStore {
...
+  /**
+   * The repositories currently checked out ref and branch name. This is used to
+   * determine if we retrieve the action workflows to see if they exist for the
+   * checkruns, which we currently only do if the we are retrieve a status for a
+   * ref that is currently checked out - not for all the pull requests.
+   */
+  private checkedOutRef: {
+    ref: string
+    branchName: string
+  } | null = null
```
niik, line 202: Style nit, we can keep this compact

```suggestion
    private checkedOutRef: { ref: string; branchName: string } | null = null
```

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -2411,8 +2411,13 @@ export class Dispatcher {
   public subscribeToCommitStatus(
     repository: GitHubRepository,
     ref: string,
-    callback: StatusCallBack
+    callback: StatusCallBack,
+    branchName?: string
```
niik, line 2415: At this point we've lost the context of what the `branchName` really is, i.e. the name of the branch that the `ref` refers to if (and only if) that branch happens to be the currently checked out branch in the current repository. Either we find a better name for it or we document the crap out of it.

app/src/ui/branches/pull-request-badge.tsx
```diff
@@ -16,6 +16,9 @@ interface IPullRequestBadgeProps {
 
   /** The GitHub repository to use when looking up commit status. */
   readonly onBadgeClick: () => void
+
+  /** The current branch name. */
+  readonly branchName: string
```
niik, line 21: If it's the "current" branch name I think the name should reflect that, like `currentBranchName` or even `currentlyCheckedOutBranchName`

app/src/ui/branches/ci-status.tsx
```diff
@@ -26,6 +26,12 @@ interface ICIStatusProps {
 
   /** A callback to bubble up whether there is a check displayed */
   readonly onCheckChange?: (check: ICombinedRefCheck | null) => void
+
+  /**
+   * The branch name of the commit ref. This can be optionally supplied when
+   * fetching actions workflow info is required for the status.
+   */
+  readonly branchName?: string
```
niik, line 34: The documentation here is glossing over the fact that this should only ever be provided when the ref is resolved from currently checked out branch. We wouldn't want consumers to add a branchname just because they have it handy.

niik (changes requested):
I can see the appeal of preloading this data for the current branch but I think the current approach is problematic. The commit status store is agnostic as to what the current repository is (dealing only with GitHubRepository instances) so I fear introducing a repository-level concern is going to be awkward.

One example of this logical mismatch can be seen in this PR by first opening the list of PRs (causing a bunch of CIStatus components to render and subscribe to the store), then checking out one of those PRs. When opening the PR badge for said PR there's not going to be any logs since the prerequisite data for those logs will only be loaded in `refreshSubscription` (which only happens once every 3 minutes).

I'm sure we could get around that limitation by some means of forcing a refresh for a subscription that we've identified as warranting loading workflow runs but I'm not certain that loading that data continually in the background even when the PR badge is collapsed is justifiable when we consider how frequently we'll want to check the overall status of a ref vs how often a user will want to drill into logs (likely mostly when the overall status has failed).

I left a few smaller comments as well but I think we need to question whether preloading the data provides enough of a UX boost to warrant the complexity we'd be getting ourselves into.

tidy-dev:
Closing this one in favor of better approach in #13096

PR #13075, opened 2021-10-06
Signal when a commit summary is getting long

### LINKED ISSUE #2055: Warn after typing more than 50 characters for a commit message

Twixes:
Closes #2055. Supersedes #12465.

## Description

This adds visual feedback regarding commit summary length, based on the design from https://github.com/desktop/desktop/pull/12465#issuecomment-933432339.

#### Ideal length case (`length <= 50`) – no change

under-50

#### Excessive length case (`50 < length <= 72`) – lightbulb hint with a tooltip on hover

over-50-out
over-50-hover

#### Overflowing length case (`length > 72`) – alert hint with a tooltip on hover

over-72-hover

## Release notes

Notes: [New] Signal when a commit summary is getting long


Twixes:
Actually looked into the wrapping mechanism more and made use of it for _true_ ahead-of-time detection of wrapping. Now emojis and URLs are taken into account properly, instead of a naive character count.

Screen Shot 2021-10-06 at 20 51 39
Screen Shot 2021-10-06 at 20 51 51

Twixes:
Citing Adele – "Hello from the other side"

Twixes:
Bump?

sergiou87:
Sorry @Twixes, we've been busy and forgot about this 😢 

I hope we can take a look at this next week. Thank you for your patience!

niik:
Hey @Twixes, my apologies for the delay. I've been out for a few weeks but I'm back now. Thanks for all the work you've put into this.

Now that our [custom tooltips PR](https://github.com/desktop/desktop/pull/13116) has been merged now we should be able to replace the `Hint` component with a `Tooltip` to align ourselves with the other tooltips and reduce the surface area of this PR.

As per https://github.com/desktop/desktop/pull/12465#issuecomment-935628317 we'd also like to focus only on the 50 character hint for now in order to align with the messaging of GitHub.com. We can continue discussing a secondary threshold separately but for this work we're looking for product alignment.

Twixes:
Alright, rescoped this PR to only show the 50 characters suggestion, using the new `Tooltip` component. Take a look.
tooltip

app/styles/ui/window/_tooltips.scss
```diff
@@ -160,4 +160,12 @@ body > .tooltip,
       margin-right: var(--spacing);
     }
   }
+
+  .title {
+    white-space: pre;
+  }
+
+  .description {
+    opacity: 0.7;
```
niik: Let's use our themed colors here instead of relying on opacity
```suggestion
    color: var(--text-secondary-color);
```

app/styles/ui/window/_tooltips.scss
```diff
@@ -160,4 +160,12 @@ body > .tooltip,
       margin-right: var(--spacing);
     }
   }
+
+  .title {
```
niik: Once we move the tooltip out of the autocompleting text input we can give it a specific class name like `overflow-hint` or something and move these styles inside of that like we do with `user-info` above.

niik (changes requested):
Thanks for making those changes @Twixes, we're getting really close now!

Could you move the hint logic out of the `AutoCompletingTextInput` component and into the `CommitMessage` component as per https://github.com/desktop/desktop/pull/12465#issuecomment-933432339

> I think we could render the warning outside of the AutocompletingTextInput (rendering popovers and handling subject-specific length doesn't seem like a concern that component should have). I absolutely agree that we shouldn't cover the icon, but I think we can avoid that by using padding on the input element and positioning the icon on top of the textbox itself. I did a super quick proof of concept in https://github.com/desktop/desktop/compare/inline-overflow-hint-spike.

app/styles/ui/window/_tooltips.scss
Twixes: I gave this a shot but it was extremely unreadable in light mode (where it's `$gray-500`). I also tried out out `$gray-400`, but that was still a bit darker than it should. And then `$gray-300`, but sadly there is a huge jump in brightness from 400 to 300 and that was barely distinguishable from white. I'd love something like `$gray-350`, but not having that, opacity 0.7 works equally well across dark and light modes, and is in the visibility/dimming sweet spot.

#### Original design
135830820-98603c30-21c5-4027-8e31-f3c0748070a1

#### Opacity 0.7
Screen Shot 2021-12-08 at 01 10 33

#### Gray 300
Screen Shot 2021-12-08 at 00 47 21

#### Gray 400
Screen Shot 2021-12-08 at 01 08 17

#### Gray 500
Screen Shot 2021-12-08 at 00 43 41

app/src/ui/autocompletion/autocompleting-text-input.tsx
```diff
@@ -39,6 +40,12 @@ interface IAutocompletingTextInputProps<ElementType> {
   /** Indicates if input field applies spellcheck */
   readonly spellcheck?: boolean
 
+  /** Optional hint icon on the right side of the component, with details shown on hover. */
+  readonly hint?: {
+    symbol: OcticonSymbolType
+    tooltip: JSX.Element | string
+  }
+
```
niik: We can remove this now, right?

```suggestion
```

app/src/ui/autocompletion/autocompleting-text-input.tsx
```diff
@@ -16,6 +16,7 @@ interface IRange {
 
 import getCaretCoordinates from 'textarea-caret'
 import { showContextualMenu } from '../main-process-proxy'
+import { OcticonSymbolType } from '../octicons'
```
niik: We should be able to remove this now that we've moved the hint out of this component.

```suggestion
```

app/styles/ui/changes/_commit-message.scss
```diff
@@ -202,3 +223,7 @@
     }
   }
 }
+
+.length-hint-tooltip {
+  max-width: none !important; // The i
```
niik: Looks like the comment here got cut off?
```suggestion
  max-width: none !important; // The i
```

Secondarily, if we still need this max-width we should move it to the `_tooltips.scss` declaration for this class which should also negate the need for the `!important` modifier.

Twixes:
Bump?

Twixes:
Hello?

app/src/ui/changes/commit-message.tsx
```diff
@@ -738,6 +742,30 @@ export class CommitMessage extends React.Component<
...
+        delay={0}
+        tooltip={
+          <>
+            <span className="title">
+              Great commit summaries contain fewer than 50 characters
+            </span>
+            <br />
+            <span className="description">
+              Place extra information in the description field.
+            </span>
```
niik: Could we use block elements here and get rid of the `br` please
```suggestion
            <div className="title">
              Great commit summaries contain fewer than 50 characters
            </div>
            <div className="description">
              Place extra information in the description field.
            </div>
```

niik (changes requested):

Twixes:
Good points, addressed all!

niik (approved):
Thank you so much for all of your hard work here @Twixes! Keep an eye on the beta channel if you'd like to try this out as soon as it becomes available.

sandstrom:
Awesome!! @Twixes @niik

callumtw:
As this is a GitHub™ thing, could we please have the option to disable this, or at least adjust the character limit, for repos that aren't hosted on github?

I and those I work with do not subscribe to the arbitrary cutoff of 50 characters as a criterion for a "good" summary, and it is distracting seeing this pop up now, which is obviously desirable for those who want the feature, not so great for those of us who don't.

PR #13076, opened 2021-10-06
Sandboxed Markdown Parser

tidy-dev:
## Description

In this PR, output summaries are displayed using this new sandboxed markdown parser.  This PR is meant just for the implementation details of parsing of markdown and the sandboxing of that parsing. Thus, if you are testing it out, you may notice that the css will need some work to make it prettier. 

What it does:
- Builds an iframe with the content of some inline styles and a single script that is the only one allowed via a content security policy.
- Component reads in the `marked.min.js` file  and the `markdown.css` provided in the `app/static` folder and provides it in a string to the iframe via the`srcdoc` property.
  - If we need to update to a newer `marked` version. Will require manual step of updating the file in the static folder.
  - `markdown.css` may feel disjoint, but better than in the js where we don't get help on our css syntax.
- Uses an event listener on the iframe's `contentDocument` to intercept link clicks and attempting to parse them as URL's/verifying they are `https` links. 

**Security Notes:**
- sandbox - "allow-scripts" means any script is allowed. The content security policy added to the iframe is what is preventing the running of other scripts.
- `<a>` tag href parsing - Href hijacking attempts by a malicious output string will occur inside the sandboxed iframe. Thus, by the time a link is clicked, it "should" be benign. Nevertheless, the url parser will also reject things that don't look like urls. After successfull url parsing, we are disallowing non `https` links

Other notes:
- Looking further into [dotcom's Markdown parser](https://github.com/github/github/blob/c64a19a5173834d776cc67f21fd7cf141e5fd2ab/lib/github/goomba.rb#L129.). It appears that they have several custom filters applied to get true gfm flavored markdown parsing. For an initial implementation, I think this js markdown parser is fine. But, it may be nice to add parsing for GitHub.com links, emojis, commit links, etc. These are all used in GitHub.com's code scanning output.

## Release notes
Notes: no-notes


sergiou87 (dismissed):
Looks good to me! Great work 😄

tidy-dev:
UPDATE: 
We were able to use the `src` attribute as opposed to the `srcdoc` attribute which was the thing preventing us from using the iframes `csp` attribute for the content security policy.

It also turns out that using `srcdoc` may not be the most secure anyways as it executes first in the context of the parent document.. which we were intending to isolated to the iframe. Per this [website tests](https://csplite.com/csp/test188/):
>  Content of srcdoc is part of the parent document, so such content is governed by the CSP of the parent document. That is, in fact, the contents of srcdoc are first executed in the context of the parent document, and then inserted into the <iframe>.

Additionally, added a baseHref property for relative links functionality.

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,222 @@
...
+                    });`
+
+    const src = `
+      <html>
+      <head>
+        ${
+          this.props.baseHref !== null
+            ? `<base href="${this.props.baseHref}" />`
+            : ''
+        }
```
niik: This is likely not an actual risk since we're currently only using it for endpoints but since we might end up using this component for more purposes than this one I'd rather we properly escape strings we're mixing with the markup. How about something like this?

```suggestion
        ${base ? base.outerHTML : ''}
```

Coupled with

```ts
  let base: HTMLBaseElement | null = null
  if (this.props.baseHref !== null) {
    base = document.createElement('base', {})
    base.href = this.props.baseHref
  }
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,222 @@
...
+    // We used this `Buffer.toString('base64')` approach because `btoa` could not
+    // convert non-latin strings that existed in the markedjs.
+    const b64src = Buffer.from(src, 'utf8').toString('base64')
+
+    // We are using `src` and data uri as opposed to an html string in the
+    // `srcdoc` property because the `srcdoc` property renders the html in the
+    // parent dom and therefore does not utilize the content security policy
+    // provided in the iframe's csp attribute. Additionally, we want all rendering
+    // to be isolated to our sandboxed iframe - not in the parents dom.
+    // -- https://csplite.com/csp/test188/
```
niik: Love these comments :sparkles:

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * We still want to be able to navigate to links provided in the markdown.
+   * However, we want to intercept them an verify they are valid links first.
+   */
+  private setupLinkInterceptor(frameRef: HTMLIFrameElement): void {
+    frameRef.onload = () => {
+      if (frameRef.contentDocument === null) {
+        return
+      }
+      const linkTags = frameRef.contentDocument.querySelector('a')
```
niik: This will only work for the first link in the document. That might be okay for the purposes of a proof of concept but the code looks well thought out enough that I suspect the intention is for this to continue working? If so perhaps we could use something like this

```ts
frameRef.contentDocument.addEventListener('click', ev => {
  const a = ev.target instanceof Element ? ev.target.closest('a') : null

  if (a !== null) {
    ev.preventDefault()
    if (/^https?:/.test(a.protocol)) {
      this.props.onMarkdownLinkClicked?.(a.href)
    }
  }
})
```

Where `.closest` is perhaps overzealous but it is required if we want to accurately capture links which themselves have clickable elements (i.e. `<a href="foo"><button>bar</button></a>`)

app/static/common/markdown.css
```diff
@@ -0,0 +1,60 @@
...
+  background-color: var(--code-background-color);
+  border-radius: 6px;
+}
+
+table {
+  display: block;
+  width: 100%;
+  width: -webkit-max-content;
+  width: -moz-max-content;
+  width: max-content;
```
niik, line 25: We don't need vendor-specific prefixes, we know which browser we're on

```suggestion
  width: max-content;
```

niik (changes requested):
This is great @tidy-dev! I added a few comments. I was also curious about whether we could experiment with grabbing the minified marked file from npm instead of including it in the static folder? It looks like it is included: https://unpkg.com/browse/marked@3.0.7/

If not then I think we should probably drop the marked-dependency from package.json since we're not using it

app/static/common/markdown.css
tidy-dev: :D copy/pasta -  Good point. Like to also point out that this css was all just a quick copy/pasts from dotcom and likely in use we will add/modify it. For example, when working my hack day project, I found it was useful for image widths to be set to 100% so they would respect an outer width set to the iframe.

wmgl5593:
> ## Description
> In this PR, output summaries are displayed using this new sandboxed markdown parser. This PR is meant just for the implementation details of parsing of markdown and the sandboxing of that parsing. Thus, if you are testing it out, you may notice that the css will need some work to make it prettier.
> 
> What it does:
> 
> * Builds an iframe with the content of some inline styles and a single script that is the only one allowed via a content security policy.
> * Component reads in the `marked.min.js` file  and the `markdown.css` provided in the `app/static` folder and provides it in a string to the iframe via the`srcdoc` property.
>   
>   * If we need to update to a newer `marked` version. Will require manual step of updating the file in the static folder.
>   * `markdown.css` may feel disjoint, but better than in the js where we don't get help on our css syntax.
> * Uses an event listener on the iframe's `contentDocument` to intercept link clicks and attempting to parse them as URL's/verifying they are `https` links.
> 
> **Security Notes:**
> 
> * sandbox - "allow-scripts" means any script is allowed. The content security policy added to the iframe is what is preventing the running of other scripts.
> * `<a>` tag href parsing - Href hijacking attempts by a malicious output string will occur inside the sandboxed iframe. Thus, by the time a link is clicked, it "should" be benign. Nevertheless, the url parser will also reject things that don't look like urls. After successfull url parsing, we are disallowing non `https` links
> 
> Other notes:
> 
> * Looking further into [dotcom's Markdown parser](https://github.com/github/github/blob/c64a19a5173834d776cc67f21fd7cf141e5fd2ab/lib/github/goomba.rb#L129.). It appears that they have several custom filters applied to get true gfm flavored markdown parsing. For an initial implementation, I think this js markdown parser is fine. But, it may be nice to add parsing for GitHub.com links, emojis, commit links, etc. These are all used in GitHub.com's code scanning output.
> 
> ## Release notes
> Notes: no-notes

add-markdown-parser

app/src/ui/branches/ci-check-run-list.tsx
```diff
@@ -210,13 +210,18 @@ export class CICheckRunList extends React.PureComponent<
   }
 
   private renderList = (checks: ReadonlyArray<IRefCheck>) => {
+    const baseHref =
+      this.props.repository.htmlURL !== null
+        ? new URL('/', this.props.repository.htmlURL).href
+        : null
```
niik: Thinking about this some more I think this is probably the more idiomatic way to get the host address

```suggestion
    const baseHref = getHTMLURL(this.props.repository.endpoint)
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Builds a <base> tag for cases where markdown has relative links
+   */
+  private getBaseTag(baseHref: string | null): string {
+    if (baseHref == null) {
+      return ''
+    }
+
+    const base = document.createElement('base', {})
```
niik: This is my bad I believe but we don't need that second parameter here
```suggestion
    const base = document.createElement('base')
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+    // provided in the iframe's csp attribute. Additionally, we want all rendering
+    // to be isolated to our sandboxed iframe - not in the parents dom.
+    // -- https://csplite.com/csp/test188/
+    this.frameRef.src = `data:text/html;charset=utf-8;base64,${b64src}`
+    this.setupLinkInterceptor(this.frameRef)
+  }
+
+  public render() {
+    return (
+      <iframe className="markdown-iframe" sandbox="" ref={this.onFrameRef} />
```
niik: Typically we use the `x-component` syntax for reusable component classnames
```suggestion
      <iframe className="sandboxed-markdown-component" sandbox="" ref={this.onFrameRef} />
```

app/styles/ui/_sandboxed-markdown.scss
```diff
@@ -0,0 +1,9 @@
+.markdown-iframe {
```
niik:
```suggestion
.sandboxed-markdown-component {
```

app/static/common/markdown.css
niik: Seems like this was marked as resolved without the change being made?

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    // We used this `Buffer.toString('base64')` approach because `btoa` could not
+    // convert non-latin strings that existed in the markedjs.
+    const b64src = Buffer.from(src, 'utf8').toString('base64')
+
+    // We are using `src` and data uri as opposed to an html string in the
+    // `srcdoc` property because the `srcdoc` property renders the html in the
+    // parent dom and therefore does not utilize the content security policy
+    // provided in the iframe's csp attribute. Additionally, we want all rendering
+    // to be isolated to our sandboxed iframe - not in the parents dom.
```
niik:
```suggestion
    // parent dom and we want all rendering to be isolated to our sandboxed iframe.
```

Since we're not using the CSP attribute we might as well drop the specific mention of it?

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+      return ''
+    }
+
+    const base = document.createElement('base', {})
+    base.href = baseHref
+    return base.outerHTML
+  }
+
+  /**
+   * Generates an iframe with a csp allowed script of a markdown parser.
```
niik: This seems like it's survived a few iterations of change?

```suggestion
   * Populates the mounted iframe with HTML generated from the provided markdown
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    const sanitizedHTML = DOMPurify.sanitize(parsedMarkdown)
+
+    const src = `
+      <html>
+        <head>
+          ${this.getBaseTag(this.props.baseHref)}
+        </head>
+        <body>
+          ${styleSheet}
```
niik: Can we move this into the `<head>` element please?

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+          <div id="content">${sanitizedHTML}</div>
```
niik: Do we need this to be encapsulated?

```suggestion
          ${sanitizedHTML}
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const b64src = Buffer.from(src, 'utf8').toString('base64')
+
+    // We are using `src` and data uri as opposed to an html string in the
+    // `srcdoc` property because the `srcdoc` property renders the html in the
+    // parent dom and therefore does not utilize the content security policy
+    // provided in the iframe's csp attribute. Additionally, we want all rendering
+    // to be isolated to our sandboxed iframe - not in the parents dom.
+    // -- https://csplite.com/csp/test188/
+    this.frameRef.src = `data:text/html;charset=utf-8;base64,${b64src}`
+    this.setupLinkInterceptor(this.frameRef)
```
niik: We should set up this interceptor when the iframe mounts, as it is now we'll add one more event listener for each time the sandboxed markdown component gets reused with new markdown

niik (changes requested):
This is looking great @tidy-dev, nearly there!

app/src/ui/lib/sandboxed-markdown.tsx
tidy-dev: TIL

tidy-dev:
@niik Incorporated the PR comments and retested. :)

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+    const base = document.createElement('base')
+    base.href = baseHref
+    return base.outerHTML
+  }
+
+  /**
+   * Populates the mounted iframe with HTML generated from the provided markdown
+   */
+  private mountIframeContents = async (): Promise<void> => {
```
niik: Looks like this doesn't have to be a bound method?

```suggestion
  private async mountIframeContents() {
```

script/build.ts
```diff
@@ -253,7 +253,7 @@ function copyStaticResources() {
   if (fs.existsSync(platformSpecific)) {
     fs.copySync(platformSpecific, destination)
   }
-  fs.copySync(common, destination, { overwrite: false })
+  fs.copySync(common, destination, { overwrite: true })
```
niik: Do you recall why this change was necessary?

niik (changes requested):
One tiny nitpicking detail and one question and then I'm happy, promise!

script/build.ts
tidy-dev: It wasn't copying the files I added... but that may have been due to something else - I was trying a few things just trying to get it to work.. I will see if still works if I remove it

tidy-dev: Seems to be picking up style changes without the change. reverted

niik (dismissed):
Great work @tidy-dev!

niik (approved):

PR #13088, opened 2021-10-07
Add checks summary header

tidy-dev:
## Description

Adds a header to the check runs so the rerun button has a home. Additionally provides an overall summary/meta data.

### Screenshots

image

There are multiple conclusion types. Technically, the summary could get very lengthy.. tho probably very unlikely.:
```
if (conclusion === null) {
    return 'In progress'
  }

  switch (conclusion) {
    case APICheckConclusion.ActionRequired:
      return 'Action required'
    case APICheckConclusion.Canceled:
      return 'Canceled'
    case APICheckConclusion.TimedOut:
      return 'Timed out'
    case APICheckConclusion.Failure:
      return 'Failed'
    case APICheckConclusion.Neutral:
      return 'Neutral'
    case APICheckConclusion.Success:
      return 'Successful'
    case APICheckConclusion.Skipped:
      return 'Skipped'
    case APICheckConclusion.Stale:
      return 'Marked as stale'
  }
  ```
Here is what it would look like if there were multiple of conclusion types:
image

UPDATE:
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/branches/ci-check-run-list.tsx
+++ b/app/src/ui/branches/ci-check-run-list.tsx
@@ -265,9 +265,9 @@ export class CICheckRunList extends React.PureComponent<
         <div className="ci-check-run-list-header">
           <div className="ci-check-run-list-title-container">
             <div className="title">Checks Summary</div>
-            {this.renderRerunButton()}
+            <div className="check-run-list-summary">{checkRunSummary}</div>
           </div>
-          <div className="check-run-list-summary">{checkRunSummary}</div>
+          {this.renderRerunButton()}
         </div>
         {checkRuns.length !== 0 ? checkLists : 'Unable to load checks runs.'}
       </div>
--- a/app/src/ui/branches/ci-check-run-list.tsx
+++ b/app/src/ui/branches/ci-check-run-list.tsx
@@ -3,6 +3,7 @@ import { GitHubRepository } from '../../models/github-repository'
 import { IDisposable } from 'event-kit'
 import { Dispatcher } from '../dispatcher'
 import {
+  getCheckRunConclusionAdjective,
   ICombinedRefCheck,
   IRefCheck,
 } from '../../lib/stores/commit-status-store'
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by tidy-dev** [comment]
@sergiou87 I also fixed the inside square contents popping out of the popover at the corners in this PR. Thanks for pointing that out. 🙂

**Review by sergiou87 — COMMENTED** [review]
Thanks for that fix 😄 

The code looks good to me overall, what feels a bit off is the location of the re-run jobs button, especially with a long summary:

image

It's like the margins of the button are uneven (top vs right), and the long summary is too close to the button.

I wonder if it makes sense to move the button down a bit (to make margin.top = margin.right) and then wrap the summary when it reaches the button. Not perfect but maybe better / less jarring?

I don't have a strong opinion about what to do, nor whether or not it should be done in this PR, but it definitely feels weird to me 😆

**Comment by tidy-dev** [comment]
Yeah. I think it looked funny too.. but wasn't sure about clipping it at the button due to that could cause pushing the summary to three lines quicker.. but again likely edge cases. I will give that a shot and see what it looks like.

File: app/src/ui/branches/ci-check-run-list.tsx [hunk 1]
```diff
@@ -147,6 +155,36 @@ export class CICheckRunList extends React.PureComponent<
...
+      summaryArray.push(`${count} ${conclusion}`)
+    }
+
+    if (summaryArray.length > 1) {
+      return `${summaryArray.slice(0, -1).join(', ')}, and ${summaryArray.slice(
+        -1
+      )} checks`
+    }
+
+    return `${summaryArray[0]} check`
```
sergiou87: I just noticed that this should be plural the first and only `count` is > 0, otherwise…
image

sergiou87 (changes requested):

app/src/ui/branches/ci-check-run-list.tsx
tidy-dev: Good catch. Thanks!

darck-storm:
hello

darck-storm:
hello

sergiou87 (approved):
Looks good! Thanks for that change 😄 

:shipit:

PR #13093, opened 2021-10-08
Rerun button will rerequest all check suite's

tidy-dev:
## Description

This adds the logic to trigger rerunning the check run suites on the rerun all jobs button. 

Currently only prod OAuth app id is allowed via checks api. Thus, to test this you must enable to the feature-flag and do a build:prod and run it from your dist folder.

@gimenete  Already has us a PR going to get the dev app id added so running build:prod is not a permanent thing.
https://github.com/github/github/pull/196291

Next up we need to retrieve the check runs immediately after to show they are pending.

ADDED:
Also note, currently this naively attempts to retry all check suites but some of them are not re-requestable and result in a 403 error. [Looking into a way to prevent trying to re-request those ones.
](https://github.com/github/c2c-actions-checks/issues/94#issuecomment-938797794)
## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1082,6 +1082,29 @@ export class API {
   }
 
   /**
+   * Triggers GitHub to rerequest an existing check suite, without pushing new
+   * code to a repository.
+   */
+  public async rerequestCheckSuite(
+    owner: string,
+    name: string,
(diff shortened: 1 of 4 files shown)


sergiou87 (approved):
:shipit: ❗

PR #13096, opened 2021-10-08
Better performance for non actions output

tidy-dev:
## Description
Approach from #13064 was undesirable from a code maintenance/concept coupling stand point.

In this PR, I still wait until the user clicks the PR badge. However, instead of waiting for the actions workflow call, jobs call, and logs call/parsing to complete. I split it into two waitings. We have a boolean that is for loading only the actions workflow call where we can determine whether a check run has actions jobs/logs associated with it. Then, we have the original boolean for when the jobs/logs api call and parsing is complete.

The actions workflow call is very lightweight in comparison to the logs retrieval/parsing. Thus, even tho this approach is not as fast #13064 in terms of performance for non-actions, I only saw the loading image for a couple of seconds for non actions check runs when my network setting was for slow 3g ... when compared 20-30s for the action logs. On my regular settings, I was hard-pressed to ever see the loading image for them. Thus, I think this is just as useable a solution.

Side note: We hope to subscribe a websocket for action logs which would likely mean they would become individualized as opposed to currently waiting on all the logs for all the check runs - thus much faster.

### Screenshots

https://user-images.githubusercontent.com/75402236/136603603-e84b6329-6fed-4a88-8ffd-42d1c1089efb.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1040,20 +1040,18 @@ export class API {
    * List workflow run jobs for a given workflow run
    */
   public async fetchWorkflowRunJobs(
-    workflowRun: IAPIWorkflowRun
+    jobs_url: string
   ): Promise<IAPIWorkflowJobs | null> {
     const customHeaders = {
       Accept: 'application/vnd.github.antiope-preview+json',
     }
-    const response = await this.request('GET', workflowRun.jobs_url, {
+    const response = await this.request('GET', jobs_url, {
       customHeaders,
     })
     try {
       return await parsedResponse<IAPIWorkflowJobs>(response)
     } catch (err) {
-      log.debug(
-        `Failed fetching workflow jobs for workflow run named: ${workflowRun.name}`
-      )
+      log.debug(`Failed fetching workflow jobs: ${jobs_url}`)
     }
     return null
   }
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -371,13 +371,13 @@ interface IAPIWorkflowRuns {
 }
 // NB. Only partially mapped
 export interface IAPIWorkflowRun {
+  readonly id: number
   /**
    * The workflow_id is the id of the workflow not the individual run.
    **/
   readonly workflow_id: number
   readonly cancel_url: string
   readonly created_at: string
-  readonly jobs_url: string
   readonly logs_url: string
   readonly name: string
   readonly rerun_url: string
@@ -1040,18 +1040,23 @@ export class API {
    * List workflow run jobs for a given workflow run
    */
   public async fetchWorkflowRunJobs(
-    jobs_url: string
+    owner: string,
+    name: string,
+    workflowRunId: number
   ): Promise<IAPIWorkflowJobs | null> {
+    const path = `repos/${owner}/${name}/actions/runs/${workflowRunId}/jobs`
     const customHeaders = {
       Accept: 'application/vnd.github.antiope-preview+json',
     }
-    const response = await this.request('GET', jobs_url, {
+    const response = await this.request('GET', path, {
       customHeaders,
     })
     try {
       return await parsedResponse<IAPIWorkflowJobs>(response)
     } catch (err) {
-      log.debug(`Failed fetching workflow jobs: ${jobs_url}`)
+      log.debug(
+        `Failed fetching workflow jobs (${owner}/${name}) workflow run: ${workflowRunId}`
+      )
     }
     return null
   }
--- a/app/src/ui/branches/ci-check-list-item.tsx
+++ b/app/src/ui/branches/ci-check-list-item.tsx
@@ -24,7 +24,10 @@ interface ICICheckRunListItemProps {
   readonly checkRun: IRefCheck
 
   /** Whether call for actions logs is pending */
-  readonly loadingLogs: boolean
+  readonly loadingActionLogs: boolean
+
+  /** Whether tcall for actions workflows is pending */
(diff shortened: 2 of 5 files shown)


app/src/lib/api.ts
```diff
@@ -1040,20 +1040,18 @@ export class API {
    * List workflow run jobs for a given workflow run
    */
   public async fetchWorkflowRunJobs(
-    workflowRun: IAPIWorkflowRun
+    jobs_url: string
```
niik: I realize this isn't a new change in this PR but I would like to see if there's a way for us to pass a repo name, owner, and id here instead of an opaque url? The request-method was intended to take a relative url and construct the absolute url and I'd rather we have more control over which urls we make requests for. There are exceptions to this in the current code base though, see [fetchAll](https://github.com/desktop/desktop/blob/7f164be1198622441996218cbb39f226c4dcbb86/app/src/lib/api.ts#L1174-L1198) as an example.

Unless the jobs_url is impossible/hard for us to construct?

app/src/ui/branches/ci-check-list-item.tsx
```diff
@@ -158,13 +161,21 @@ export class CICheckRunListItem extends React.PureComponent<
     )
   }
 
+  private hasActionsWorkflowLogs = (): boolean => {
```
niik: It doesn't appear as if we're calling this from an unbound context anywhere so we can probably just make it a regular class field

```suggestion
  private hasActionsWorkflowLogs() {
```

app/src/ui/branches/ci-check-run-list.tsx
```diff
@@ -106,17 +108,42 @@ export class CICheckRunList extends React.PureComponent<
...
+      a check run does not have action logs to retrieve/parse.
+    */
+    const checkRunsWithActionsUrls = await this.props.dispatcher.getCheckRunActionsJobsAndLogURLS(
+      this.props.repository,
+      this.getCommitRef(this.props.prNumber),
+      this.props.branchName,
+      statusChecks
+    )
+
+    this.setState({
```
niik, line 146: Given that these methods can end up taking a bit of time it's pretty easy to get into a scenario where we call setState after unmount

image

Could we either add a flag that we toggle on unmount (i.e. the [good ol' so-called anti-pattern](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html#:~:text=An%20easy%20migration%20strategy%20for%20anyone%20upgrading%20their%20code%20to%20avoid%20isMounted()%20is%20to%20track%20the%20mounted%20status%20yourself.%20Just%20set%20a%20_isMounted%20property%20to%20true%20in%20componentDidMount%20and%20set%20it%20to%20false%20in%20componentWillUnmount%2C%20and%20use%20this%20variable%20to%20check%20your%20component%E2%80%99s%20status.)) or could we use `this.subscription` as a proxy for that?

```suggestion
  if (this.subscription === null) {
    return
  }

  this.setState({
```

If nothing else it'll prevent us from calling `getActionsWorkflowRunLogs` unnecessarily

app/src/ui/branches/ci-check-run-list.tsx
```diff
@@ -106,17 +108,42 @@ export class CICheckRunList extends React.PureComponent<
...
+      loadingActionWorkflows: false,
+    })
+
+    const checkRuns = await this.props.dispatcher.getActionsWorkflowRunLogs(
+      this.props.repository,
+      this.getCommitRef(this.props.prNumber),
+      checkRunsWithActionsUrls
+    )
+
+    this.setState({ checkRuns, loadingActionLogs: false })
```
niik, line 164: Same question here about calling setState after unmount

niik (changes requested):
I like this approach much better @tidy-dev :sparkles:

While I was testing out #13106 I disabled the cache and happened to see that we're making multiple consecutive requests to the same jobs endpoint (for the same id). This is certainly not a huge problem when caching is enabled but it is typically indicative of potential areas of improvement as even if the resource is cached there's still parsing overhead and likely logical overhead on our end as well.

I haven't dug into why it's happening so it's absolutely possible that the complexity of maintaining our own high-level cache isn't worth it but since I saw mentions of a logCache in there already I figured I should at least raise the issue and understand it a little bit better.

niik:
> While I was testing out #13106 I disabled the cache and happened to see that we're making multiple consecutive requests to the same jobs endpoint (for the same id). This is certainly not a huge problem when caching is enabled but it is typically indicative of potential areas of improvement as even if the resource is cached there's still parsing overhead and likely logical overhead on our end as well.

Just realized I forgot to include any more information on this. My bad, I had recorded a small GIF to illustrate but forgot to attach it.

app/src/lib/api.ts
tidy-dev: Sure can. This would also apply to the logs_url - but that one I just found out I was doing in a more complicated way than I needed to because I had overlooked where I believe I can get individual job log in the API. Thus, I am going to open a separate PR to improve that anyways and I will make build the url then. :)

tidy-dev:
Surprised on the multiple calls to same jobs url id.. cause I am doing a cache to try to prevent just this.. :/

EDIT: Evidently I didn't every set anything in that cache or I accidentally deleted it at some point. 🤦

tidy-dev:
Fixed duplicate job calls:

https://user-images.githubusercontent.com/75402236/137484221-6788d4fe-302e-4509-b685-193c78e469f4.mov

niik (approved):
:sparkles:

PR #13106, opened 2021-10-11
Make log file request without auth token

tidy-dev:
## Description

This PR:
1) Adds the redirect: 'manual' to all our http requests so that no-one in the future can unintentionally send an auth token to an external provided url via fetches default behavior to follow the location header with the initial requests headers. This setting cancels the redirect call. This means it will show in the network tab in red. 

_Another other option would be to set it to 'error' and add a parameter such as isRedirectManual. This would provided more explicit code for future devs and still disallow the default option of 'follow'_

2) Utilizes electron session's `webRequest.onBeforeRedirect` to grab the redirect url and then makes a request to it without providing the auth token.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -15,6 +15,7 @@ import username from 'username'
 import { GitProtocol } from './remote-parsing'
 import { Emitter } from 'event-kit'
 import JSZip from 'jszip'
+import { remote } from 'electron'
 
 const envEndpoint = process.env['DESKTOP_GITHUB_DOTCOM_API_ENDPOINT']
 const envHTMLURL = process.env['DESKTOP_GITHUB_DOTCOM_HTML_URL']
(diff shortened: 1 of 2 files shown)


niik:
Thanks for opening this @tidy-dev! I think I would prefer a more overarching approach here. While this solves the issue for this specific endpoint I'm also concerned about future API endpoints we might end up calling that leverage this redirect technique. I took a stab at an approach in https://github.com/desktop/desktop/pull/13108 which should protect us from cross-domain requests preserving tokens regardless of where they're initiated from. Let me know what you think.

app/src/lib/api.ts
```diff
@@ -1064,13 +1065,60 @@ export class API {
...
-      customHeaders,
+    const zipURL: string | null = await new Promise(resolve => {
+      const customHeaders = {
+        Accept: 'application/vnd.github.antiope-preview+json',
+      }
+
+      remote.session.defaultSession.webRequest.onBeforeRedirect(details => {
+        if (
+          details.responseHeaders === undefined ||
+          details.responseHeaders.location.length === 0
```
niik, line 1076: This did end up triggering a crash for me because the responseHeaders aren't automatically lowercased so `details.responseHeaders.location` was undefined whereas `details.responseHeaders.Location` wasn't

tidy-dev:
> Thanks for opening this @tidy-dev! I think I would prefer a more overarching approach here.

Your overarching solution is sweet!  ✅ 

> while this solves the issue for this specific endpoint I'm also concerned about future API endpoints we might end up calling that leverage this redirect technique. 

I was thinking from the vantage point of disabling all future attempts (redirect: manual) to redirect and then if someone needed it - they would have to be intentional about it.  That said I much prefer yours where we don't get that error state network requests.. and the "Just Works" feel of it.

niik:
> I was thinking from the vantage point of disabling all future attempts (redirect: manual) to redirect and then if someone needed it - they would have to be intentional about it.

Ah, that's a great point. And I imagine we would have been able to create an abstraction method to reduce the boilerplate but then we'd still have the issue of not being able to have two of these intercept mechanisms alive concurrently (due to webRequest only allowing one filter at a time)

PR #13108, opened 2021-10-12
Prevent cross-domain redirects from preserving auth headers

niik:
Refs: #13106

## Description

GitHub Desktop uses the [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) web API for all of our API requests. When fetch is used in a browser and it encounters an http redirect to another origin domain CORS policies will apply to prevent submission of [credentials](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch).

In our case however there's no concept of same-origin (and even if there were it'd be problematic because we'd be making cross-origin request constantly to GitHub.com and GHE instances) so the `credentials: same-origin` setting won't help us.

This is normally not a problem until http redirects get involved. When making an authenticated request to an API endpoint which in turn issues a redirect to another domain fetch will happily pass along our token to the second
domain and there's no way for us to prevent that from [happening](https://github.com/whatwg/fetch/issues/763) using the vanilla fetch API.

In this PR I'm introducing a filter which attempts to identify any cross-origin redirect and sanitize potentially sensitive headers from the subsequent request.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -1,6 +1,6 @@
 import '../lib/logging/main/install'
 
-import { app, Menu, ipcMain, BrowserWindow, shell } from 'electron'
+import { app, Menu, ipcMain, BrowserWindow, shell, session } from 'electron'
 import * as Fs from 'fs'
 import * as URL from 'url'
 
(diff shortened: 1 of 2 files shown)


tidy-dev (approved):
✨ Nice! (So much better than mine having all the error state network requests)

PR #13115, opened 2021-10-13
Fix missing CSS variables

niik:
## Description

Noticed that the inactive selected state for list items has disappeared in the light theme.

This is how it looks

image

And this is how it should look

image

This was accidentally removed and not caught in review in https://github.com/desktop/desktop/pull/12952

Additionally I noticed that the hover state for vertical segmented controls didn't work in either dark or light theme. That was accidentally removed by myself in https://github.com/desktop/desktop/commit/2476c49f8655ed7c5a9e07550eb62c21fb609432 and apparently hasn't worked since 2017(!)

## Release notes

Notes: no-notes

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -113,6 +113,11 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
   --box-border-accent-color: #{$blue};
 
   /**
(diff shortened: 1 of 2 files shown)


sergiou87 (approved):
Nice catch!! 🤯 

:shipit:

PR #13116, opened 2021-10-13
Custom tooltips

### LINKED ISSUE #6384: Implement our own tooltip style
Right now, when we need to use a tooltip, we've been utilizing the ability to show a title/alt tag on hover. This doesn't show up immediately, or very prominently, so it would be helpful to implement our own tooltips that can show more context on hover when needed

Currently we do use this, which shows up after a delay: 

screen shot 2018-12-10 at 2 12 48 pm

And it would be awesome to implement something like this, which shows up immediately:

tooltip-custom

niik:
Closes #6384 

## Description

The day has finally come. We've been talking about better tooltips for years now and hopefully this is it. We've done attempts before but have been thwarted either by complexity (positioning tooltips without overflowing the bounds of the window is tricky) or simplicity (using a primer-style CSS-only solution will only let us have text tooltips).

In designing this I had a few objectives that I hoped to achieve

1. Tooltips should not affect the element tree of the target (i.e. don't force me to wrap my button in a div)
2. Tooltips should be usable inside of our lists and should not be affected by `overflow: hidden` (i.e. absolute positioning relative to window[^1]). Additionally they should work transparently within a dialog (i.e. we can't rely entirely on z-index).
3. Tooltips should be accessible
4. We should allow interactive tooltips with images, buttons, you name it

I'm happy to say I think this PR achieves the objectives or at least comes pretty close. The accessibility part is tricky as I'm not an expert in it. I have followed the accessibility guidelines for custom tooltips as best I've could though.

This turned into a rather large PR, the majority of the changes being additions of tooltips, or removals of tooltips that shouldn't be tooltips (a tooltip with the same content as the button to which it's referring to doesn't add much value). I'll try to describe a few concepts and changes and hope that will alleviate some of the review burden but I'm also **happy to pair-review on this**

#### createObservableRef

React has two recommended ways of [capturing references](https://reactjs.org/docs/refs-and-the-dom.html) to component instances or underlying DOM elements. Either [createRef](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) or callback refs. You typically use `createRef` for its advantageous syntax and callback refs if you care about when the ref change.

For `Tooltip` I wanted a combination of both. I wanted the ease of use for consumers of `createRef` but I needed a way to get notified when the Ref changed. That's what `createObservableRef` does. For all intents and purposes it can be thought of as a normal Ref except that it also allows consumers to subscribe to change events.

#### Tooltip placement

When a tooltip is about to be displayed it'll look for the closest ancestor of the target element which is considered a suitable host for tooltips. For GitHub Desktop that's either the `body` or `dialog` element in case the target lives within a modal dialog.

#### Disabled buttons aren't disabled any more

One change of note in this PR is how we're handling disabled buttons. Previously we just slapped the `disabled` attribute on 'em and called it a day. Unfortunately that means no pointer events will trigger on the button which in turn means we have no way to know when to show the tooltip. Since showing tooltips for disabled buttons was called out even in the initial issue for this as being useful I had to take a different approach.

Instead of `disabled` we're now using `aria-disabled` to let a11y tools know what's up and we then disable the pointer events in react.

A side-effect of this is that disabled buttons are now reachable by keyboard navigation. This is a change in behavior but one that I think it beneficial from an accessibility standpoint.

See https://a11y-101.com/development/aria-disabled which recommends this approach over `disabled`

## Release notes

Notes:

[^1]: Rather closest modal container so window or dialog


app/src/ui/branches/pull-request-list-item.tsx
```diff
@@ -48,10 +48,20 @@ export interface IPullRequestListItemProps {
   readonly onDropOntoPullRequest: (prNumber: number) => void
 }
 
+interface IPullRequestListItemState {
+  readonly isDragInProgress: boolean
```
niik, line 52: All of the changes related to this come from https://github.com/desktop/desktop/pull/13116/commits/0c0675843f54e9d4f897cbb034cd712141544de0 where we previously disabled all pointer events for all PR list items we only do so now when a drag operation is in progress.

app/src/ui/branches/branch-list-item.tsx
```diff
@@ -119,33 +144,36 @@ export class BranchListItem extends React.Component<IBranchListItemProps, {}> {
...
-    const name = this.props.name
-
-    const date = lastCommitDate ? moment(lastCommitDate).fromNow() : ''
+    const { lastCommitDate, isCurrentBranch, name } = this.props
     const icon = isCurrentBranch ? OcticonSymbol.check : OcticonSymbol.gitBranch
-    const infoTitle = isCurrentBranch
-      ? 'Current branch'
-      : lastCommitDate
-      ? lastCommitDate.toString()
-      : ''
```
niik, line 132: This is handled by the `RelativeTime` component now instead

app/src/ui/branches/ci-check-list-item.tsx
```diff
@@ -79,7 +79,6 @@ export class CICheckRunListItem extends React.PureComponent<
                 `ci-status-${getClassNameForCheck(step)}`
               )}
               symbol={getSymbolForCheck(step)}
-              title={step.name}
```
niik: This didn't add information which wasn't already present

app/src/ui/diff/side-by-side-diff-row.tsx
```diff
@@ -446,9 +445,6 @@ export class SideBySideDiffRow extends React.Component<
           hover: this.props.isHunkHovered,
         })}
         style={{ width: this.props.lineNumberWidth }}
-        title={
-          this.props.hideWhitespaceInDiff ? HideWhitespaceWarning : undefined
-        }
```
niik, line 451: This is sufficiently covered by the whitespace hint popover on click now

app/src/ui/lib/avatar-stack.tsx
```diff
@@ -29,7 +29,7 @@ export class AvatarStack extends React.Component<IAvatarStackProps, {}> {
         elems.push(<div key="more" className="avatar-more avatar" />)
       }
 
-      elems.push(<Avatar key={`${i}`} user={users[i]} title={null} />)
```
niik, line 32: This is a change in behavior. We want to show tooltips for avatars even inside of an avatar stack

app/src/lib/is-empty-or-whitespace.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Indicates whether the given string is null, undefined, empty, or consists
+ * only of white-space characters.
+ */
```
sergiou87: I guess `null` and `undefined` are not options anymore, right?
```suggestion
/**
 * Indicates whether the given string is empty or consists only of white-space
 * characters.
 */
```

app/src/ui/lib/tooltip.tsx
```diff
@@ -0,0 +1,533 @@
...
+import { createUniqueId, releaseUniqueId } from './id-pool'
+import classNames from 'classnames'
+import { assertNever } from '../../lib/fatal-error'
+import { rectEquals, rectContains } from './rect'
+
+export type TooltipDirection = 'n' | 'ne' | 'e' | 'se' | 's' | 'sw' | 'w' | 'nw'
+const DefaultTooltipDelay = 400
+const InteractiveTooltipHideDelay = 150
+
+// Curse you SVGSVGElement
```
sergiou87: 🤣 

What happened here? 😂  Is it worth explaining in the comment?

app/src/ui/lib/observable-ref.ts
```diff
@@ -0,0 +1,32 @@
+type RefCallback<T> = (instance: T | null) => void
+export type ObservableRef<T> = {
+  current: T | null
+  subscribe: (cb: RefCallback<T>) => void
+  unsubscribe: (cb: RefCallback<T>) => void
+  (instance: T): void
```
sergiou87, line 6: TIL about [Call Signatures](https://www.typescriptlang.org/docs/handbook/2/functions.html#call-signatures)

app/src/ui/lib/tooltip.tsx
```diff
@@ -0,0 +1,533 @@
...
+   */
+  readonly direction?: TooltipDirection
+
+  /**
+   * An optional additional class name to set on the tooltip in order to be able
+   * to apply specific styles to the tooltip
+   */
+  readonly className?: string
+
+  readonly onlyWhenOverflowed?: boolean
```
sergiou87, line 77: Could we have a comment for this one too?

app/src/ui/lib/tooltip.tsx
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    // We already know it won't fit
+    candidates.delete(direction)
+  } else {
+    // Placement based on mouse position, prefer south east, north east
+    candidates = new Set(['se', 'ne', ...candidates])
+  }
+
+  for (const candidate of candidates) {
```
sergiou87, line 489: I was surprised to see this logic based on a Set, because in my experience with other languages sets were always unordered. TIL JavaScript uses the insertion order when you iterate over its entries!

app/src/ui/lib/tooltip.tsx
```diff
@@ -0,0 +1,533 @@
...
+  private onTargetFocusIn = (event: FocusEvent) => {
+    if (this.state.target?.matches(':focus-visible')) {
+      this.beginShowTooltip()
+    }
+  }
+
+  private onTargetFocusOut = (event: FocusEvent) => {
+    if (!this.mouseOverTarget) {
+      this.beginHideTooltip()
+    }
```
sergiou87: Shouldn't this (or even better, `beginHideTooltip`) take into account not only whether or not the mouse is over the target, but also if it's over the tooltip (when it's interactive)?

app/src/ui/lib/tooltip.tsx
```diff
@@ -0,0 +1,533 @@
...
+      prevProps.target.unsubscribe(this.onTargetRef)
+      this.props.target.subscribe(this.onTargetRef)
+    }
+
+    if (target !== prevState.target) {
+      if (target === null) {
+        this.removeTooltip(prevState.target)
+      } else {
+        this.installTooltip(target)
+      }
```
sergiou87: Shouldn't we always remove the tooltip from the previous target?
```suggestion
      this.removeTooltip(prevState.target)
      if (target !== null) {
        this.installTooltip(target)
      }
```

app/src/ui/branches/branch-list-item.tsx
```diff
@@ -119,33 +144,36 @@ export class BranchListItem extends React.Component<IBranchListItemProps, {}> {
...
-        </div>
+        </TooltippedContent>
+        {lastCommitDate && (
+          <RelativeTime
+            className="description"
+            date={lastCommitDate}
+            onlyRelative={true}
+          />
+        )}
       </div>
```
sergiou87, line 179: Not sure if I'm in the wrong place, but the app is not showing these specific tooltips (branch name and relative time) for me 🤔

app/src/ui/lib/path-text.tsx
```diff
@@ -296,31 +294,35 @@ export class PathText extends React.PureComponent<
...
-      >
+      <div className="path-text-component" ref={this.pathElementRef}>
         <span ref={this.onPathInnerElementRef}>
           {directoryElement}
           <span className="filename">{this.state.fileText}</span>
         </span>
+        {truncated && (
+          <Tooltip
+            target={this.pathElementRef}
+            interactive={true}
```
sergiou87, line 307: Not for this PR, but since this is interactive, it'd be nice to have a copy button too 😄

app/styles/_variables.scss
```diff
@@ -294,23 +294,24 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
   --win32-scroll-bar-size: 10px;
 
   /**
-   * The z-index for popups. Nothing should be higher, but other z-indexes can
+   * The z-index for tooltips. Nothing should be higher, but other z-indexes can
    * be defined in terms of it.
    *
    * 10 seems arbitrary (because it is) but we have to go that high because
    * codemirror is all over the place with their z-indexes and uses at least
    * 0-6.
```
sergiou87: This comment needs to be updated, right?

app/styles/_variables.scss
```diff
@@ -294,23 +294,24 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
...
-  --drag-overlay-z-index: calc(var(--popup-z-index) - 3);
+  --tooltip-z-index: 20;
+  --popup-z-index: calc(var(--tooltip-z-index) - 1);
+  --popup-overlay-z-index: calc(var(--tooltip-z-index) - 2);
+  --foldout-z-index: calc(var(--tooltip-z-index) - 3);
+  --nudge-arrow-z-index: calc(var(--tooltip-z-index) - 4);
+  --drag-overlay-z-index: calc(var(--tooltip-z-index) - 5);
   /**
    * This is 7 to make sure it appears on top code mirror
    * but behind popups, overlays, and foldouts
```
sergiou87: Same about this one? 😆

sergiou87 (changes requested):
Looks great overall!! This approach feels super clean 😍 

Awesome job! Reviewing this was very educational too 🙇‍♂️

app/src/ui/branches/branch-list-item.tsx
niik: My bad: 639752c3ac200713fdc3ec12744d353d6424e7d8

app/src/ui/lib/path-text.tsx
niik: Yeah, I was going back and forth on that one. The reason I didn't add one was that these tooltips only show when the path overflows and I feel like if we add a nice copy-button it could lead to confusion when users aren't able to copy that way on all paths. I'm still not convinced either way though

app/src/ui/lib/tooltip.tsx
niik: Lol, yeah probably: bd9d939b4539361405add50424a4dfebcf9658b4

app/src/ui/lib/tooltip.tsx
niik: You bet: db5482e4eca219bd8e902066dc4fbacbf2e0cc76

app/src/ui/lib/tooltip.tsx
niik: Yeah, good catch!

app/src/ui/lib/tooltip.tsx
niik: Yup! Same goes for `Map`, very surprising but super handy

app/src/ui/lib/tooltip.tsx
niik: It sure should, nice catch: 7e7ed3121c870ff2d6a763a13228ed693c7ecccf

app/styles/_variables.scss
niik: Yeah, let's fix these once and for all: https://github.com/desktop/desktop/pull/13116/commits/3199e81d23b5192a51cbecc1cd04b118414d394d

sergiou87 (dismissed):
Thanks for those changes, it looks great now in my setup!! 😍 

The only thing I noticed (but can be done in a separate PR if we think it's worth it) is that you need to move the mouse "fast enough" when moving from the tooltipable element to the (interactive) tooltip or the tooltip will be hidden.

That means I expect people to randomly get the tooltip hidden when they try to interact with interactive tooltips. You can see in the following video how the tooltip disappears if my mouse lands anywhere between the tooltipable element and the tooltip itself:

https://user-images.githubusercontent.com/1083228/138680684-76c2004b-5970-47e0-97eb-35e2d925da0d.mov

Feel free to ignore this in this PR if you want, because otherwise it's ready for :shipit: ❗

niik:
> The only thing I noticed (but can be done in a separate PR if we think it's worth it) is that you need to move the mouse "fast enough" when moving from the tooltipable element to the (interactive) tooltip or the tooltip will be hidden.

Yeah, I've been tweaking that delay a few times but I probably ended up making it too fast. I've bumped the delay by another 100ms in https://github.com/desktop/desktop/pull/13116/commits/9b45db402c47a2b1fc3d1c16c83a9203c051ab54, let me know how that feels

sergiou87 (approved):
It feels much better, thanks! ❤️ 

Although to be honest I was expecting some kind of way to extend the "hoverable" area of the tooltip to overlap/reach the area of the tooltipable element.

That said, this approach definitely works good enough with 250ms, in my experience at least, so I'm more than happy to go ahead with it!

niik:
> Although to be honest I was expecting some kind of way to extend the "hoverable" area of the tooltip to overlap/reach the area of the tooltipable element.

Yeah, it's a bit of a hack. I did play around with an extended hit area but it got tricky when using pointer position placement (i.e. when not specifying a direction) as the hitarea could end up extending over the pointer itself. By no means an unsolvable problem but let's try the hack and see how it feels before trying something else. At the very least we could try extending the hitarea to encompass the tip of the tooltip.

app/src/ui/branches/pull-request-list-item.tsx
jmayes41: > All of the changes related to this come from https://github.com/desktop/desktop/pull/13116/commits/0c0675843f54e9d4f897cbb034cd712141544de0 where we previously disabled all pointer events for all PR list items we only do so now when a drag operation is in progress.

PR #13150, opened 2021-10-19
Actions Log Parser

tidy-dev:
## Description

The logs that are provided for action workflow check runs are generated by Azure devops. Therefore, they follow the [Azure pipeline logging format](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#formatting-commands)

Fortunately for us, a parsing script for this was already available in typescript in use on GitHub.com. With permission, I have stolen the parser.ts and log-parser.ts file. However, these files were using a npm package called lit-html for the html rendering. Since we want to user react to render our, I have modified the approach to remove that dependency and simply provide as a parsed template object that we can use to render similar html.

In the dotcom approach, the parser.ts is the meat of the parsing for tokens such as warning, error, notice, colors and log-parser was a ui log manager/template generator. The exception to this is that the parsing of timestamps and detail groups was done in the log-parser. The log ui management was mostly for the concept of toggling display of timestamps and search highlighting. For our current implementation, we don't want to provide log searching (users can go to dotcom for full log experience). 

Thus, in this PR, I have stripped the templating, lit-html referencing, and searching logic. This means modifying the parsing to return an object instead of lit-html templates. I also broke down the regions in the original file into other files to try to make the parser file just about the parser.

Below are things I left in that could be further stripped... but I am not so sure we won't want them and they don't add much complexity.
- Storing the timestamps. (right click or hover for timestamp in case it is important to user?)
- Providing a base url for line number links. (another way to jump to dotcom? or copy to send to coworker?)

### Screenshots

**THIS PR IS JUST THE PARSER - No ui changes ** but below is a screen shot for POC that it works. 😛 
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/actions-log-parser/actions-log-parser-objects.ts
+++ b/app/src/lib/actions-log-parser/actions-log-parser-objects.ts
@@ -1,0 +1,114 @@
+export const enum NodeType {
+  Plain = 0,
+  Command = 1,
+  Debug = 2,
+  Error = 3,
+  Info = 4,
+  Section = 5,
+  Verbose = 6,
+  Warning = 7,
+  Group = 8,
+  EndGroup = 9,
+  Icon = 10,
+  Notice = 11,
+}
+
+// Set max to prevent any perf degradations
+export const maxLineMatchesToParse = 100
+export const unsetValue = -1
+export const newLineChar = '\n'
+export const hashChar = '#'
+export const commandStart = '['
+export const commandEnd = ']'
+
+interface IGroupInfo {
+  lineIndex: number
+  nodeIndex: number
+}
+
+export interface ISectionFind {
+  line: number
+}
+
+export interface IParseNode {
+  type: NodeType
+  /**
+   * Index of the node inside ILine
+   */
+  index: number
+  /**
+   * Index of the ILine this node belongs to
+   */
+  lineIndex: number
+  /**
+   * Starting index of content
+   */
+  start: number
+  /**
+   * Ending index of content
+   */
+  end: number
+  /**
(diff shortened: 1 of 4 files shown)


app/src/lib/actions-log-parser/actions-log-parser-objects.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+export interface ILogLine extends IParseNode {
+  text?: string
+}
+
+export interface ILogLineTemplateData {
+  className: string
+  lineNumber: number
+  lineContent: IParsedContent[]
```
sergiou87:
```suggestion
  lineContent: ReadonlyArray<IParsedContent>
```

app/src/lib/actions-log-parser/actions-log-parser-objects.ts
```diff
@@ -0,0 +1,114 @@
...
+   */
+  group?: IGroupInfo
+  /**
+   * If this is a group, this would be set
+   */
+  isGroup?: boolean
+}
+
+export interface ILine {
+  nodes: IParseNode[]
```
sergiou87:
```suggestion
  nodes: ReadonlyArray<IParseNode>
```

sergiou87 (commented):
Since this code was taken from somewhere else, there are a bunch of discrepancies with our code style, but not sure if it's worth upgrading it… depending on how often we expect to "steal" upgrades to the original code or if it's gonna be extracted into a package for easier reuse.

Other than that, this should be ready to :shipit:

sergiou87 (approved):

PR #13154, opened 2021-10-19
Action Logs congruent with dotcom styling

tidy-dev:
## Description

- Moved action logs output display into it's own component.
- Updated UI elements/icons/CSS to be congruent with dotcom's dark terminal like approach (whether app is in light or dark mode)
- Allow user to open any step log output
- Reverted to simple Loading... text for loading state.. Seems it might fit better for logs.. but TBD/designer input needed.

To be continued...
- Add add azure log format parser to the actual log output. (With this will come more styles and html)
- Move the View on GitHub button to an external link icon in the corner.
- Resizable log area
- Remove hover effect for skipped headers.. surprisingly difficult with my current css knowledge... there must be something I don't understand.. :/ 

### Screenshots

image

image

Light theme broader context:
image

Dark theme broader context:
image

## Release notes
Notes: no-notes


app/src/lib/stores/commit-status-store.ts
```diff
@@ -778,6 +779,27 @@ export function getCheckDurationInSeconds(
...
+    return `${duration}s`
+  }
+
+  const minutes = Math.floor(duration / 60)
+  const seconds = duration - minutes * 60
+  if (seconds === 0) {
+    return `${minutes}m`
+  }
+
+  return `${minutes}m ${seconds}s`
```
niik: Looking at what .com does it seems you'd want to support more than just minutes and hours. I think something like this would match pretty closely

```suggestion
  return moment
    .duration(getCheckDurationInSeconds(checkRun), 'seconds')
    .format('d[d] h[h] m[m] s[s]', { largest: 4 })
```

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -0,0 +1,110 @@
...
+      stepOpenState.set(index, !(stepOpenState.get(index) === true))
+      this.setState({ stepOpenState: new Map(stepOpenState) })
+    }
+  }
+
+  public render() {
+    const { output, loadingLogs } = this.props
+
+    if (loadingLogs) {
+      return <>Loading...</>
```
niik:
```suggestion
      return <>Loading…</>
```

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -0,0 +1,110 @@
...
+interface ICICheckRunActionLogsProps {
+  /** The check run to display **/
+  readonly output: IRefCheckOutput
+
+  /** Whether call for actions logs is pending */
+  readonly loadingLogs: boolean
+}
+
+interface ICICheckRunActionLogsState {
+  readonly stepOpenState: Map<number, boolean>
```
niik: First off I'd like it if we could make this a readonly map to ensure we don't accidentally mutate the map instead of replacing it? (as mutating won't cause us to rerender)

But, after thinking about it some more a `Map<T, boolean>` can usually be represented equally as well if not better with a `Set<T>`. In this case you only care about which sections _are open_, not which are explicitly closed.

```suggestion
  readonly openSections: ReadonlySet<number>
```

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export class CICheckRunActionLogs extends React.PureComponent<
+  ICICheckRunActionLogsProps,
+  ICICheckRunActionLogsState
+> {
+  public constructor(props: ICICheckRunActionLogsProps) {
+    super(props)
+    this.state = {
+      stepOpenState: new Map<number, boolean>(),
```
niik:
```suggestion
      openSections: new Set<number>(),
```

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+  public toggleOpenState = (index: number) => {
+    return () => {
+      const { stepOpenState } = this.state
+      stepOpenState.set(index, !(stepOpenState.get(index) === true))
+      this.setState({ stepOpenState: new Map(stepOpenState) })
```
niik:
```suggestion
            const openSections = new Set(this.state.openSections)
      if (openSections.has(index)) {
        openSections.delete(index)
      } else {
        openSections.add(index)
      }

      this.setState({ openSections })
```

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    if (output.type !== RefCheckOutputType.Actions) {
+      // This shouldn't happen, should only be provided actions type
+      return <>Unable to load logs.</>
+    }
+
+    return output.steps.map((step, i) => {
+      const isSkipped = step.conclusion !== 'skipped'
+      const showLogs = this.state.stepOpenState.get(i) === true && isSkipped
```
niik:
```suggestion
      const showLogs = this.state.openSections.has(i) && isSkipped
```

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -0,0 +1,110 @@
...
+      return <>Loading...</>
+    }
+
+    if (output.type !== RefCheckOutputType.Actions) {
+      // This shouldn't happen, should only be provided actions type
+      return <>Unable to load logs.</>
+    }
+
+    return output.steps.map((step, i) => {
+      const isSkipped = step.conclusion !== 'skipped'
```
niik: The name and condition seems to mismatch here?

app/src/ui/check-runs/ci-check-run-logs.tsx
```diff
@@ -132,13 +91,7 @@ export class CICheckRunLogs extends React.PureComponent<ICICheckRunLogsProps> {
...
 
   private renderLoadingLogs = () => {
-    return (
-      <div className="loading-logs">
-        <img src={BlankSlateImage} className="blankslate-image" />
-        <div className="title">Hang tight</div>
-        <div className="loading-blurb">Loading the logs as fast as I can!</div>
-      </div>
-    )
+    return <div className="no-logs-to-display">Loading</div>
```
niik:
```suggestion
    return <div className="no-logs-to-display">Loading…</div>
```

app/styles/ui/check-runs/_ci-check-run-logs.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  &.actions {
+    color: var(--check-run-header-label-text-color);
+    background-color: var(--check-run-bg-color);
+    border-color: var(--check-run-border-color);
+
+    .ci-check-run-log-step {
+      // TODO: Figure out why hover only does middle part and not padding when using classes
+      // Currently can't figure out how to not hover skipped steps. :/
```
niik, line 28: It doesn't appear as-if you're adding any class name for skipped steps at the moment?

niik (dismissed):
Looking great! 😍

app/styles/ui/check-runs/_ci-check-run-logs.scss
tidy-dev: Yeah. I beat my head on it for an hour trying to do it === 'skipped' with classes and doing it !== 'skipped'.. and then gave up and removed work on that and put the todo comment in. .... but evidently left my !== in there.. which looks like I need to put back. I will fix that but would like to table figuring out  the skipped hover state for this PR (maybe we can pair on it later so I can show you the css problem I am running into)

app/src/lib/stores/commit-status-store.ts
tidy-dev: 😻

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
tidy-dev: Flipped back.... Skipped css rules were not playing nice so I was flipped the logic to try it that way (which did not work either.. which makes sense it would behave the same.. but don't understand it's current behavior... so tried it. 😛).

sergiou87 (approved):
Looks beautiful!! 😍 :shipit:

PR #13169, opened 2021-10-24
Added styled scrollbar

Piterden:
## Description

Added styles for scrollbar

### Screenshots

image

instead of 

image

### DIFF
--- a/app/styles/_globals.scss
+++ b/app/styles/_globals.scss
@@ -186,3 +186,20 @@ kbd {
     margin-right: 2px;
   }
 }
+
+::-webkit-scrollbar {
+  width: 10px;
+}
+
+::-webkit-scrollbar-thumb {
+  background-image: linear-gradient(to bottom, #8290AD85, #68769385);
+  box-shadow: 4px 0 13px 0 rgba(53, 147, 238, 0.43);
+}
+
+::-webkit-scrollbar-track {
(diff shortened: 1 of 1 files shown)


niik:
Hi @Piterden, thanks for opening this. Could you provide a bit more information on your system. The scrollbar from your first screenshot looks like it's a Windows scrollbar but on Windows we already use custom, overlying, scrollbars, see:

image

By the looks of the screenshots you're posting the font also differs from what I'd expect. Have you made any other changes to the styling which could have affected the layout?

Piterden:
I have manjaro linux with awesome as a WM. But I guess there should be exactly the same scrollbars for any OS, if you define scrollbar styles directly, like I propose to do it in this PR. 

Electron uses WebView, which will display almost any custom styled element the same for any OS. Only default (not-styled) elements can be different depending on OS

niik:
> I have manjaro linux with awesome as a WM.

Ah, got it, that was the piece I was missing here. I would suggest opening an issue in the [linux fork](https://github.com/shiftkey/deskop) as it seems like this problem was supposed to be addressed in https://github.com/shiftkey/desktop/pull/240 but might have regressed since?

Since the official GitHub Desktop repository doesn't contain the linux-specific styling I'm going to close this, hope you're able to resolve this with the linux fork!

PR #13178, opened 2021-10-25
Replace dialog animations with fade when user prefers reduced motion

### LINKED ISSUE #2768: Github Desktop not obeying disabling of UI grow animations.

### LINKED ISSUE #12976: Dialog animations are bouncy when the user prefers reduced motion
By default, dialogs appear and disappear with a slight bounce. This is not affected by the user's reduced motion preference.

### Version & OS

Open the 'About GitHub Desktop' menu to see the GitHub Desktop version. Also include what operating system you are using.

- **GitHub Desktop:** 2.9.4-beta2
- **Operating System:** macOS Monterey 12.0 beta

### Expected behavior

The dialog uses an alternate animation such as a fade-in when the user prefers reduced motion.

### Actual behavior

The dialog uses the same animation when the user prefers reduced motion.

Relates to #4544

niik:
Fixes #12976
Fixes #2768

## Description

Since we're now able to make use of the `prefers-reduced-motion` media query I took the opportunity to update the dialog animation based on it.

Note that the presence of `prefers-reduced-motion` does not imply blanket removal of animations but rather to avoid excessive motion. In the case of dialog appear/exit the scale transform is merely replaced with a fade in/out transition instead. In other places like the [whitespace hint popover](https://github.com/desktop/desktop/blob/e3e2a4dd087d152038474b1d5928fc817553a18c/app/styles/ui/_popover.scss#L55-L64) it makes more sense to remove the animation altogether.

## Release notes

Notes:

### DIFF
--- a/app/styles/ui/_dialog.scss
+++ b/app/styles/ui/_dialog.scss
@@ -50,6 +50,22 @@ dialog {
   min-width: 400px;
   max-width: 600px;
 
+  // Substitute the appear/exit animations with a fade in fade out effect if the
+  // user prefers reduced motion.
+  //
+  // https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
+  @media (prefers-reduced-motion) {
+    transition: opacity 100ms ease-in !important;
+    transform: none !important;
+
+    &.modal-enter {
+      opacity: 0 !important;
+    }
+    &.modal-enter-active {
+      opacity: 1 !important;
+    }
+  }
+
   // The modal class here is the transition name for the react css transition
   // group which allows us to apply an animation when the popup appears.
   &.modal {


sergiou87 (approved):
This is nice! 😌 :shipit:

PR #13207, opened 2021-10-29
Checks failed dialog

sergiou87:
## Description

More work on "checks failed" notification: this PR has the new dialog that will be shown when a notification like that is received.

One change worth noting in this PR is that I made the state of showing or not the CI status popover part of the global app state. That will allow us to easily show/hide it when the user clicks on a notification, depending on whether or not the user is in the very same repository and PR referenced by the notification.

There is [a commit](https://github.com/desktop/desktop/commit/925355de177f136d17a767c6823988c7e1f95bcb) in this branch that must not be merged because it's here just for testing purposes: if you click on the CI badge in the branch dropdown button, the app will show the dialog for the current Pull Request.

### Screenshots

https://user-images.githubusercontent.com/1083228/142204248-04534d73-e8cd-4bf6-8158-11b2f296602b.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -28,6 +28,12 @@ interface ICICheckRunListItemProps {
   /** Whether or not to show the action workflow event in the title */
   readonly showEventInTitle: boolean
 
+  /** Whether the list item can be selected */
+  readonly selectable: boolean
+
+  /** Whether the list item is selected */
+  readonly selected: boolean
+
   /** Callback for when a check run is clicked */
   readonly onCheckRunExpansionToggleClick: (checkRun: IRefCheck) => void
 
@@ -35,7 +41,7 @@ interface ICICheckRunListItemProps {
   readonly onViewCheckExternally: (checkRun: IRefCheck) => void
 
   /** Callback to open a job steps link on dotcom*/
-  readonly onViewJobStep: (
+  readonly onViewJobStep?: (
     checkRun: IRefCheck,
     step: IAPIWorkflowJobStep
   ) => void
@@ -54,7 +60,7 @@ export class CICheckRunListItem extends React.PureComponent<
   }
 
   private onViewJobStep = (step: IAPIWorkflowJobStep) => {
-    this.props.onViewJobStep(this.props.checkRun, step)
+    this.props.onViewJobStep?.(this.props.checkRun, step)
   }
 
   private renderCheckStatusSymbol = (): JSX.Element => {
@@ -74,9 +80,9 @@ export class CICheckRunListItem extends React.PureComponent<
   }
 
   private renderCheckJobStepToggle = (): JSX.Element | null => {
-    const { checkRun, isCheckRunExpanded } = this.props
+    const { checkRun, isCheckRunExpanded, selectable } = this.props
 
-    if (checkRun.actionJobSteps === undefined) {
+    if (checkRun.actionJobSteps === undefined || selectable) {
       return null
     }
 
@@ -119,7 +125,10 @@ export class CICheckRunListItem extends React.PureComponent<
     return (
       <>
         <div
-          className="ci-check-list-item list-item"
+          className={classNames('ci-check-list-item list-item', {
+            selected: this.props.selected,
+          })}
+          tabIndex={0}
           onClick={this.toggleCheckRunExpansion}
         >
           {this.renderCheckStatusSymbol()}
--- a/app/src/ui/check-runs/ci-check-run-list.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list.tsx
@@ -6,6 +6,7 @@ import {
   isFailure,
 } from '../../lib/ci-checks/ci-checks'
 import { CICheckRunListItem } from './ci-check-run-list-item'
+import { FocusContainer } from '../lib/focus-container'
 
 interface ICICheckRunListProps {
   /** List of check runs to display */
@@ -17,11 +18,17 @@ interface ICICheckRunListProps {
   /** Whether loading workflow  */
   readonly loadingActionWorkflows: boolean
 
+  /** Whether check runs can be selected. Default: false */
+  readonly selectable?: boolean
+
   /** Callback to opens check runs target url (maybe GitHub, maybe third party) */
   readonly onViewCheckDetails: (checkRun: IRefCheck) => void
 
+  /** Callback when a check run is clicked */
+  readonly onCheckRunClick?: (checkRun: IRefCheck) => void
+
   /** Callback to open a job steps link on dotcom*/
-  readonly onViewJobStep: (
+  readonly onViewJobStep?: (
     checkRun: IRefCheck,
     step: IAPIWorkflowJobStep
   ) => void
@@ -54,8 +61,7 @@ export class CICheckRunList extends React.PureComponent<
     props: ICICheckRunListProps,
     currentState: ICICheckRunListState | null
   ): ICICheckRunListState {
-    let checkRunExpanded =
-      currentState !== null ? currentState.checkRunExpanded : null
+    let checkRunExpanded = currentState?.checkRunExpanded ?? null
 
     if (currentState === null || !currentState.hasUserToggledCheckRun) {
       // If there is a failure, we want the first check run with a failure, to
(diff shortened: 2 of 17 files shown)


app/src/ui/notifications/pull-request-checks-failed.tsx
```diff
@@ -0,0 +1,301 @@
...
+  IPullRequestChecksFailedState
+> {
+  private checkRunsLoadCancelled: boolean = false
+
+  public constructor(props: IPullRequestChecksFailedProps) {
+    super(props)
+
+    const { checks } = this.props
+
+    const selectedCheck = checks.find(check => isFailure(check)) ?? checks[0]
```
tidy-dev:
```suggestion
    const selectedCheck = checks.find(isFailure) ?? checks[0]
```
Nit: Just a short hand that can make it slightly more readable. :P

app/src/ui/notifications/pull-request-checks-failed.tsx
```diff
@@ -0,0 +1,301 @@
...
+      <div className="ci-check-rerun">
+        <Button onClick={this.rerunJobs} disabled={checks.length === 0}>
+          <Octicon symbol={syncClockwise} /> Re-run jobs
+        </Button>
+      </div>
+    )
+  }
+
+  private rerunJobs = () => {
+    // Get unique set of check suite ids
```
tidy-dev: Hmm.. would you mind creating a dispatcher method that we pass the checks and the githubrepository into to handle this duplicated logic? 

This method will evolve some with the actions team adding the property to determine which checks can be retried and then being able to only rerun failed jobs. It would be nice that if we update the dispatcher method it will update both implementations that I think we would expect similar behavior from. :)

app/src/ui/notifications/pull-request-checks-failed.tsx
```diff
@@ -0,0 +1,301 @@
...
+    if (this.checkRunsLoadCancelled) {
+      return
+    }
+
+    this.setState({
+      checks: checkRunsWithActionsUrls,
+      loadingActionWorkflows: false,
+    })
+
+    const checks = await getLatestPRWorkflowRunsLogsForCheckRun(
```
tidy-dev, line 331: Just thoughts no change needed... For some reason and now I can't remember where I picked this up, but I was working under the assumption that we use the dispatcher > a store > api as a code convention/pattern for getting stuff from api.. but not sure if there is a reason for this (unless you need to update a store/state). Just noticing the difference in approach of your dialog vs my popover. Thinking that it could be better if we shared approaches just for future maintainability/maybe we could share code for doing the same things. I have no preferences on it one way or another.. only reason I think going through commit store could be good for my case anyways is this opens it up for caching in the commit store.. currently if a user closes and opens popover they reload all the checks and logs again... which is annoying when there are no changes.. and maybe exp could be improved with a thought out caching scheme. Also.. streaming logs.. may play a role in this a well...

tidy-dev (changes requested):
Just a couple of small considerations. But overall looking great! Making me want to get the action log parser in so both of our logs will be more readable. 😁

app/src/ui/notifications/pull-request-checks-failed.tsx
sergiou87: Oops! This was `!isSuccess` before, when I made the change to `isFailure` I didn't realize about this 😂  Thanks!

app/src/ui/notifications/pull-request-checks-failed.tsx
sergiou87: Agree with all that! It probably makes sense to at least hide access to the API behind the dispatcher, but I think we'll ultimately want to see what the implementation looks like in the end for both the popover (already implemented) and all the "checks failed" notification stuff (for which I have only implemented a PoC but will need many changes after we use an actual real-time approach) before we decide what's the best way to keep everything reusable and encapsulated.

sergiou87:
I'm gonna move this back to draft until I update the branch with the latest changes from `development` and we clarify what amount of detail we want to show here.

/cc @adrianmg

app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -116,7 +122,10 @@ export class CICheckRunListItem extends React.PureComponent<
     return (
       <>
         <div
-          className="ci-check-list-item list-item"
+          className={classNames('ci-check-list-item list-item', {
+            selected: this.props.selected,
+          })}
+          tabIndex={0}
```
sergiou87, line 131: This is needed to make the `FocusContainer` work.

app/src/ui/check-runs/ci-check-run-list.tsx
```diff
@@ -53,8 +60,7 @@ export class CICheckRunList extends React.PureComponent<
     props: ICICheckRunListProps,
     currentState: ICICheckRunListState | null
   ): ICICheckRunListState {
-    let checkRunExpanded =
-      currentState !== null ? currentState.checkRunExpanded : null
+    let checkRunExpanded = currentState?.checkRunExpanded ?? null
```
sergiou87, line 64: This should be equivalent!

app/src/ui/check-runs/ci-check-run-popover.tsx
```diff
@@ -186,6 +188,11 @@ export class CICheckRunPopover extends React.PureComponent<
       loadingActionWorkflows: false,
     })
 
+    if (1 !== NaN) {
+      this.props.onCheckRunsLoaded(checkRunsWithActionsUrls)
+      return
+    }
```
sergiou87: This won't be merged, of course 😄

sergiou87:
@tidy-dev I finally updated my changes basing them on your latest (great) work on the popover, please take another look when you have a chance 🙏

tidy-dev (dismissed):
Works great! ✨ 

Some thoughts - but not a hinderance to this PR: 

Wondering if maybe it would benefit from getting some UI/UX designer input?

This area particularly...
image

This top part feels hard to digest. 
-  If this popped up, I don't think the main header tells me what I am looking at/ I don't know which bold thing to look at.
- The way the things line up feels disjoint.
- The commit message I think feels like in-actionable extra info that draws my attention immediately... but maybe I don't understand the context as to why we are including it. 
- I think " x checks failed in our pull request" is the key take away but is hard to see quickly.

A quick play around with css in debugger.. thoughts?
image

sergiou87:
I like that! We were already considering changes in this UI, but didn't have anything specific. I like your suggestion, I'll run it by Adrian and try to get a 👍  from him 😄 

Meanwhile I'd like to merge this as it is (I'll remove 925355de177f136d17a767c6823988c7e1f95bcb so the dialog is not reachable)

tidy-dev (approved):

adrianmg:
Some quick feedback (sorry if it's too late!)

image

- Dividers are making things harder to scan. They also make it less clear that there's a parent selected and some children on its side. It feels like they are at the same level, not within it.
- `Re-run jobs` button needs consistent spacing as we have for the left side of the dialog
- The original dialog had a better hierarchy of the information: you are looking at this PR, here is some information. Now it reads like some information you don't know where is coming from, and here's the PR where it's coming from.
- Avoid having iconography of the same size sharing almost the same space (proximity). It dilutes the purpose of the overall state and makes things less clear.

tidy-dev:
> The original dialog had a better hierarchy of the information: you are looking at this PR, here is some information. Now it reads like some information you don't know where is coming from, and here's the PR where it's coming from.

Just my 2 cents. We are not looking at a PR in this dialog and is why I find using it as the header title confusing. We are looking at the checks that failed for a PR.

tidy-dev:
Also definitely not too late! I believe Sergio is planning a follow up PR as needed for the UI/UX and that screenshot was not implemented in any way :D just a quick moving of things and adjusting css in my debugger - so placement and such was not thought out... more so a What if? trying to help communicate the goal of the dialog which is to bring to the user the failed check runs on a PR that they are not currently on so they can decid what to do about it (rerun jobs, switch pr, or nothing). 

> make it less clear that there's a parent selected

The layout of left hand "menu style" to the right content was originally planned with displaying logs in mind.. so maybe there is a better way to show this info in this dialog knowing we don't need as much screen real estate for logs? I don't have any ideas and we probably do want it to feel similar to the popover list so user connect the two concepts easier.

Tho this does kind of fall in line with how our preferences dialog works. :/ 
image

PR #13230, opened 2021-11-01
Implementing the actions parser

tidy-dev:
## Description

Adding html rendering based on action log parser objects. TLDR; Still needs css work. This PR is getting the action log parser objects into rendered html. 

As can be seen in the screenshot.. I have some css to figure out.. spent some time on it.. but the details open and close thing is competing with the showing the lines as they are and not to wrap - use x for overflow (goal is to have the popover resizable if user would like to view more with out scrolling)... I figured I could go ahead and get the logic/html portion reviewed and then toil away at css prettifying in a later PR as there are a few CSS type things that need looked at.

### Screenshots
https://user-images.githubusercontent.com/75402236/139723723-574f93cb-84e2-44de-ac21-9022a6a5ee07.mov

## Release notes
Notes: no-notes


app/src/ui/check-runs/ci-check-run-actions-logs.tsx
```diff
@@ -49,23 +61,176 @@ export class CICheckRunActionLogs extends React.PureComponent<
...
+    const logParser = new ActionsLogParser(log, '')
+    const logLinesData = logParser.getParsedLogLinesTemplateData()
+    this.calculateLineLogNumberWidth(logLinesData)
+
+    const logLineJSX: JSX.Element[] = []
+
+    let logGroup: ILogLineTemplateData[] = []
+    for (let i = 0; i < logLinesData.length; i++) {
+      const lineData = logLinesData[i]
+      const isNextLineDataInGroup = logLinesData[i + 1]?.inGroup
```
sergiou87, line 97: Ultra minor nit: if we want to avoid using `!` for falsey/truthy conditions, we should provide a default `false` value for this:
```suggestion
      const isNextLineDataInGroup = logLinesData[i + 1]?.inGroup ?? false
```

Feel free to ignore it if you want 😅

sergiou87 (approved):
Looks good! Tested it and works as expected :_) :shipit:

app/src/ui/check-runs/ci-check-run-actions-logs.tsx
tidy-dev: :) I will go ahead and merge but catch it in another PR. :

PR #13255, opened 2021-11-03
Implement markdown parser

tidy-dev:
## Description
Notes to test this - you will have to run build:dev because the markdown.css is copied from the static files the build.

This PR implements the new sandboxed markdown parser for all content that could have markdown on it for check runs. That is non-action logs and the optional summary output.

Major changes:
- Combined summary and action log output when both exist so we only have to deal with one iframe.
- Wired up the links from markdown to open in a browser.
- Implemented a dynamic height on the iframe div container and styled the iframe such that it absolutely fits it's container div so that we can use the containing divs scrollbars to hopefully prevent double scrollbars from happening. This also gives a more inline and less of a bulky iframe feel.
- Downloaded a copy of the markdown.css from the [primer repo](https://github.com/primer/css/) to use as the base and vast majority of they styling we have for markdown to mimic dotcom feel.
  - I kept the '.markdown-body' class even tho that is functionally the body tag in the iframe mainly for search-ability in the primer repo.
  - The color variable names are modified to match our variable names but they were copied from the light and dark theme in the same primer repo so their core names are searchable. 
  -  dotcom has the luxury of real estate and not working within an iframe. Thus, I did have to do some tweaking of that css for the pre tag and the tables so we wouldn't get multiple horizontal scroll bars.
  - Since dotcom implementation is not in an iframe, it does rely on its base css for some things such as some href styling and table styling that I have added. Thus, It could be such that there are other styles we may run into that I have not thought of/missed. (For example, I haven't had a check run with an image yet... may have to force that scenario to see how images look).

### Screenshots

The repos I have access to that use this component are private so skipping the screenshots.
Goal is to feel like dotcom implementation.
Feel free to kick the tires on it. :) 

## Release notes
Notes: no-notes


app/styles/ui/_sandboxed-markdown.scss
```diff
@@ -1,9 +1,22 @@
...
-  border: 0;
-  padding: 0;
-  margin: 0;
-  box-shadow: none;
+.sandboxed-markdown-iframe-container {
+  position: relative;
+  min-height: 50px;
+
+  .sandboxed-markdown-component {
+    // With compbination of setting height of containing div.
```
sergiou87:
```suggestion
    // With combination of setting height of containing div.
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -161,11 +204,16 @@ export class SandboxedMarkdown extends React.PureComponent<
...
   public render() {
     return (
-      <iframe
-        className="sandboxed-markdown-component"
-        sandbox=""
-        ref={this.onFrameRef}
-      />
+      <div
+        className="sandboxed-markdown-iframe-container"
+        ref={this.onFrameContainingDivRef}
```
sergiou87, line 209: If you don't plan on doing anything when the reference is set, I think you can just pass the ref itself here:
```suggestion
        ref={this.frameContainingDivRef}
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -88,22 +103,50 @@ export class SandboxedMarkdown extends React.PureComponent<
...
-          if (a !== null) {
-            ev.preventDefault()
+  /**
+   * Iframes without much styling help will act like a block element that has a
+   * predetermiend height and width and scrolling. We want our iframe to feel a
+   * bit more like a div. Thus, we want to capture the scroll height, and set
+   * the container div to that height and with some additional css we can
+   * achieve a inline feel.
+   */
+  private setupFrameScrollHeightListener(frameRef: HTMLIFrameElement): void {
```
sergiou87: Is this actually a listener? 🤔  Seems like it only changes the div height once, right?

sergiou87 (changes requested):
LGTM overall! Only a couple of things called my attention

app/src/ui/lib/sandboxed-markdown.tsx
tidy-dev: technically anytime the iframe loads..

tidy-dev: Which we expect to be only once.. but it happens anytime the iframe onload event fires.

tidy-dev: Actually... that may fire again if you change the markdown .. which we would want to happen

app/src/ui/lib/sandboxed-markdown.tsx
tidy-dev: I was excited to use this... but when I tried it said it needed to be type LegacyRef<HTMLDivElement | undefined> When I set it to that and I tried to use it, it didn't have the style could be type 'string'.. So evidently using the method does some type casting for me that I would like to keep.

app/src/ui/lib/sandboxed-markdown.tsx
tidy-dev: 😜  Tho I think your point that the method name is misleading just flew over my head. I renamed it.

sergiou87: > 😜 Tho I think your point that the method name is misleading just flew over my head. I renamed it.

Yes 😂  Either it was misleading or I was missing something 😳

sergiou87 (approved):
LGTM and works as expected! 😄  :shipit:

local-developer:
800e25528d4f5c46634a3d531525514e2ff4f758

PR #13260, opened 2021-11-03
Make the check run "View on GitHub" button not so prominent. 

tidy-dev:
## Description

Made the "View on Github" button not so prominent - also increased log viewing max-height.

Thanks for the feedback @sergiou87 and @billygriffin about hover over to display.

### Screenshots

https://user-images.githubusercontent.com/75402236/140186329-db77102d-13b0-469f-8336-c5a0ca4f1a2e.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -77,7 +77,6 @@ export class CICheckRunListItem extends React.PureComponent<
             baseHref={baseHref}
             loadingActionLogs={loadingActionLogs}
             loadingActionWorkflows={loadingActionLogs}
-            onViewOnGitHub={this.onViewOnGitHub}
             onMarkdownLinkClicked={this.props.onMarkdownLinkClicked}
           />
         ) : null}
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -57,6 +57,14 @@ export class CICheckRunListItem extends React.PureComponent<
     this.props.onViewOnGitHub(this.props.checkRun)
   }
 
+  private onMouseOverLogs = () => {
+    this.setState({ isMouseOverLogs: true })
+  }
+
+  private onMouseLeaveLogs = () => {
+    this.setState({ isMouseOverLogs: false })
+  }
+
   public render() {
     const { checkRun, showLogs, loadingActionLogs, baseHref } = this.props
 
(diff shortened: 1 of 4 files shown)


app/styles/ui/check-runs/_ci-check-run-list-item.scss
```diff
@@ -30,4 +30,16 @@
...
+    display: none;
+    padding: var(--spacing);
+    padding-bottom: var(--spacing-half);
+    color: var(--text-secondary-color);
+  }
+
+  &:hover .view-on-github,
+  .view-on-github.show {
+    display: block;
+  }
```
sergiou87, line 48: It feels weird that when you hover over the icon/button to view it on GitHub it doesn't highlight too 🤔  This feels better to me:

```suggestion
  &:hover .view-on-github,
  .view-on-github.show {
    display: block;
  }
  
  .view-on-github:hover {
    color: var(--text-primary-color);
  }
```

app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -63,21 +85,32 @@ export class CICheckRunListItem extends React.PureComponent<
...
+            className={classNames('view-on-github', {
+              show: this.state.isMouseOverLogs,
+            })}
+            onClick={this.onViewOnGitHub}
+          >
+            <Octicon
+              symbol={OcticonSymbol.linkExternal}
+              title="View on GitHub"
+            />
+          </div>
```
sergiou87, line 104: Setting up the tooltip only for the octicon feels weird (especially if you add the `:hover` color change to `.view-on-github` as I suggest in my other comment) because it's smaller than the parent `div`, and clicking on that `div` but outside of the `Octicon` will open GitHub even if you don't see the tooltip.

I tried just replacing the `div` with a `TooltippedContent` element but then the tooltip doesn't appear from the center of the `div`/octicon 😢  It would probably require some extra work to make all this look 100% perfect… what do you think?

sergiou87 (changes requested):
I love it!!! Only left a couple of things that can be improved 😄

app/src/ui/check-runs/ci-check-run-list-item.tsx
tidy-dev: I don't know.. The tooltip feels a bit aggressive to me already.. I even considered not having it. IMO, it would feel weird for it to popup when I wasn't on the icon (even tho the click area is a bit bigger which makes it easy to hit).

tidy-dev: After discussion, have changed the text color change to when you are over the icon so the text color changes feels in conjunction with the tool tip.

sergiou87 (approved):
Works great now! 😍 

🐸 🕴️

PR #13261, opened 2021-11-03
Non Actions log output content share scrollbar

tidy-dev:
Note: This is built on top #13260 - just split it out into another pull request for easier review.

## Description

Make all non-actions log output to be combined in the iframe so we don't get weird disjoint scrollbar.

### Screenshots

image

## Release notes
Notes: no-notes


sergiou87 (approved):
This is much better!! 🐸 🕴️

PR #13271, opened 2021-11-04
Open and scroll to first Error check run when opening a failed check run popover.

tidy-dev:
## Description

To help get he user to the likely information they need, we will open the first app header that has a check run with an error and then open the first error check run and scroll to the first check run step with an error.

### Screenshots
https://user-images.githubusercontent.com/75402236/140426513-4d94eb39-f4a8-47a4-bfe3-96ce22203fb7.mov

## Release notes
Notes: no-notes


sergiou87 (approved):
Nice touch!! :shipit:

PR #13272, opened 2021-11-04
Clip long check run names so we don't have wonkiness on hover.

tidy-dev:
## Description
Noticed that for long check run names.. we get some ugly movement/wrapping with new hover to View on Github.

### Screenshots
I found it via the last check run on Central repo (one in Other group).. didn't post screenshots since it is a private repo.

But, here is a forced finished result with text clipping. (Before, it would wrap to new line..)
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/styles/ui/check-runs/_ci-check-run-list-item.scss
+++ b/app/styles/ui/check-runs/_ci-check-run-list-item.scss
@@ -12,7 +12,7 @@
   .ci-check-list-item-detail {
     flex: 1;
     margin: auto;
-    width: 80%;
+    overflow-x: hidden;
   }
 
   .ci-check-name {
--- a/app/styles/ui/check-runs/_ci-check-run-list-item.scss
+++ b/app/styles/ui/check-runs/_ci-check-run-list-item.scss
@@ -12,10 +12,14 @@
   .ci-check-list-item-detail {
     flex: 1;
     margin: auto;
+    width: 80%;
   }
 
   .ci-check-name {
     font-weight: var(--font-weight-semibold);
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
   }
 
   .ci-check-description {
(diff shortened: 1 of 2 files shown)


app/styles/ui/check-runs/_ci-check-run-list-item.scss
```diff
@@ -12,10 +12,14 @@
   .ci-check-list-item-detail {
     flex: 1;
     margin: auto;
+    width: 80%;
```
sergiou87: I think we should avoid setting the size of elements to a % when it's not 100%, unless we know the value we use will work with any width of the container. For example, if we made the popover wider, would 80% still be a good value? Probably not since the check status icon and the "View On GitHub" icon have fixed widths. We want the titles to take the rest.

For me, this change worked (and I think it makes more sense):
```suggestion
    overflow-x: hidden;
```

app/styles/ui/check-runs/_ci-check-run-list-item.scss
```diff
@@ -32,10 +36,14 @@
   }
 
   .view-on-github {
-    display: none;
-    padding: var(--spacing);
-    padding-bottom: var(--spacing-half);
+    margin: var(--spacing);
     color: var(--text-secondary-color);
+    min-height: var(--spacing);
+    min-width: var(--spacing-double);
```
sergiou87: It feels odd to use variables that are related to spacing to calculate height and width in situations when they're not directly related to spacing 😕 

I tried out removing these two lines and I personally liked the result: the title would expand to use all the width available and then, when you hover over the item (or its logs), the icon would show up:

https://user-images.githubusercontent.com/1083228/140714396-89342ae9-e36c-4bbf-99c2-e54689dbbb0e.mov

What do you think?

sergiou87 (changes requested):
Just a couple of notes. Also, maybe we should set the `title` for those elements because if the title is too long, I'm not getting a tooltip with the full title 🤔

tidy-dev:
Added tooltip for on hover.
image

sergiou87 (approved):
Awesome! Thanks for those changes 🙇‍♂️  🐸 🕴️

PR #13273, opened 2021-11-04
Prevent unnecessary log retrieval when check run popover is left open.

tidy-dev:
## Description
Before adding this, if you were to open the popover and expand to view some action logs and wait long enough for the pull requests status refresh to happen (or navigate to different app and return after a minute). The action logs would suddenly go away and show "Unable to display logs" until all the logs are retrieved again and user may need to click on it. This was because the status refresh causes a new retrieval of the logs and didn't flip the loading boolean. This PR does 2 things. It adds some detection to see if the check runs actually changed since that last retrieve to prevent lots of unnecessary retrievals. But if they do differ, it sets the loading flags to be true, so the user will see loading indicator until logs retrieval completes.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-popover.tsx
+++ b/app/src/ui/check-runs/ci-check-run-popover.tsx
@@ -9,9 +9,10 @@ import {
 } from '../../lib/ci-checks/ci-checks'
 import { Octicon, syncClockwise } from '../octicons'
 import { Button } from '../lib/button'
-import { getHTMLURL } from '../../lib/api'
+import { APICheckConclusion, getHTMLURL } from '../../lib/api'
 import { Popover, PopoverCaretPosition } from '../lib/popover'
 import { CICheckRunList } from './ci-check-run-list'
+import _ from 'lodash'
 
 interface ICICheckRunPopoverProps {
   readonly dispatcher: Dispatcher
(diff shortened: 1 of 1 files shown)


app/src/ui/check-runs/ci-check-run-popover.tsx
```diff
@@ -81,6 +83,12 @@ export class CICheckRunPopover extends React.PureComponent<
   }
 
   public componentDidMount() {
+    const combinedCheck = this.props.dispatcher.tryGetCommitStatus(
```
tidy-dev, line 86: This actually should have been here already.. because onStatus calls setState.. which will error if the component isn't mounted yet. Before.. it just so happened it was mounted because the api call took long enough. (Now that I have set state earlier for loaders.. it was causing the error)

app/src/ui/check-runs/ci-check-run-popover.tsx
```diff
@@ -108,6 +116,31 @@ export class CICheckRunPopover extends React.PureComponent<
...
+      // Already loading, it is the first iteration and we need to continue to retrieve the logs.
+      !loadingActionWorkflows &&
+      // This is for typing. If check is null, state will be set in next statement.
+      check !== null &&
+      // If check run conclusion is null, then it is pending and we don't want to stop retrieval of latest logs
+      check.conclusion !== null &&
+      // If not same conclusion, then need to get updated logs.
+      check.conclusion === combinedConclusion &&
+      // If same conclusion, we see if checkrun ids are the same in the two arrays.
+      _.xor(
```
sergiou87, line 136: TIL about `_.xor` 😄

sergiou87 (approved):
Thanks for this! 🙇‍♂️ :shipit:

PR #13289, opened 2021-11-08
Spinner loader on check run header selection (similar to dotcom loading)

tidy-dev:
## Description

On dotcom, when a user clicks on an actions step workflow header, the user gets a spinner in place of the dropdown chevron until the logs can load. In Desktop, since our logs all are loaded at the check name level (one step higher in the grouping). I implemented a similar loader at that level instead, but the same spinner to mimic dotcom feel. Just like dotcom, you only see a spinner if you attempt to open one and the logs are being retrieved.

### Screenshots
When a failure exists, it is going to already be in a loading state as we want to open it as soon as it loads. On a success state, you won't see the spinner till you try to click on one that is loading... this way you don't see numerous loaders on the screen at once

https://user-images.githubusercontent.com/75402236/140754548-d18be9ba-567a-40eb-a1cc-58e8c51abeca.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -122,7 +122,7 @@ export class CICheckRunListItem extends React.PureComponent<
   }
 
   public render() {
-    const { checkRun, showLogs, loadingActionLogs, baseHref } = this.props
+    const { checkRun, showLogs, baseHref } = this.props
 
     return (
       <>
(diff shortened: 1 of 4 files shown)


app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -67,25 +67,70 @@ export class CICheckRunListItem extends React.PureComponent<
...
+      >
+        <g strokeWidth="2">
+          <circle cx="8" cy="8" r="7" stroke="#959da5"></circle>
+          <path
+            d="m12.9497 3.05025c1.3128 1.31276 2.0503 3.09323 2.0503 4.94975 0 1.85651-.7375 3.637-2.0503 4.9497"
+            stroke="#fafbfc"
+          ></path>
+        </g>
+      </svg>
+    )
```
sergiou87, line 104: It'd be nice to have this as a reusable component, but I guess we can wait until we want to actually reuse it

sergiou87 (commented):
Looks and works great! 😍 ✨ 

Only one nit, what do you think?

app/src/ui/check-runs/ci-check-run-list-item.tsx
tidy-dev: Yeah. I would definitely like to wait.. seeing as this really is only for waiting on logs.. and that may not be a thing anyways.

sergiou87 (approved):

PR #13299, opened 2021-11-09
No logs display for ci checks

tidy-dev:
## Description
This may change depending on further feedback on what expanding check runs is going to look like... but I already had this going and want to get it in so we have something that looks more usable.

image

## Release notes

Notes: no-notes


sergiou87 (approved):
Code looks good and works as expected! ✨

PR #13303, opened 2021-11-10
CI check run popover positioning

tidy-dev:
## Description

1) Check run popover is now absolutely positioned so that it doesn't cause the Publish/Fetch button to move on open. Added logic to watch for change in position caused by mac full screen - same approach used for branch and repo dropdowns.

2) Check run popover will now expand to take up the whole height of the app when viewing a repo with numerous checks.

### Screenshots
(Shows that publish/fetch button isn't moved)
image

## Release notes

Notes: no-notes


sergiou87 (approved):
Looks and works good! :shipit:

PR #13306, opened 2021-11-10
Flatten check run heirarchy

tidy-dev:
## Description

This PR flattens the check hierarchy and changes the naming scheme to have a more consistent feel to the merged check list on a PR conversation tab of dotcom.

Also while testing this, I realized that PRs from forked repos were not showing action steps when dotcom was.. and that was because I was using Desktop's local branch name which is a locally generated name in the case of forks. Thus, switched to use the upstream branch name.

Notes for future PRs:
1) Requires us to wait for first actions check run retrieve and therefore we will need to add a loading at the top level and I don't think spinner I had on other fits here.. leaning towards a similar loader as shown in PR list.

2) This required some removal of old log viewing code that was just feature flagged before, but it is not meant to exhaustively removal log logic (have another PR for that).

### Screenshots

image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -336,7 +336,6 @@ export interface IAPIRefCheckRun {
   readonly status: APICheckStatus
   readonly conclusion: APICheckConclusion | null
   readonly name: string
-  readonly output: IAPIRefCheckRunOutput
   readonly check_suite: IAPIRefCheckRunCheckSuite
   readonly app: IAPIRefCheckRunApp
   readonly completed_at: string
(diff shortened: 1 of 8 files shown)


sergiou87 (commented):
LGTM but when I tested I noticed this (which AFAIK is not mentioned in the PR description):

2021-11-12 at 11 10

You can fix it in this PR or in another one, I don't think that's a problem 😄  Up to you!

EDIT: The app also crashed when I clicked on one of those steps 😱 

image

tidy-dev:
I am going to have another PR to address the steps. As per the UI feedback, I am going to switch it to look like a desktop list with just a slightly darker background tone (same as header) instead of the super dark dotcom steps. Since they are not associated with showing logs anymore, we don't really need the terminal look or muscle memory benefit from making them look like dotcom.

sergiou87 (dismissed):

sergiou87 (approved):

jeanp1209 (commented):
Fgb

jeanp1209:
> Fgb

PR #13317, opened 2021-11-12
Refactor job steps/header toggle/view on github 

tidy-dev:
## Description

 - With removal of logs, refactored job steps to use a desktop themed list. 
 - Moved check run toggle to end.
 - Made list item a split hover interaction with tooltips. 
   - Clicking on main details with view check run online.
   - Non-actions check runs will not have a dropdown toggle
- This also removes a lot of the logs code but again not exhaustively.

### Screenshots
https://user-images.githubusercontent.com/75402236/141520059-eaf01851-5e36-4991-910a-39f60ff48873.mov

## Release notes
Notes: no-notes


app/src/ui/check-runs/ci-check-run-popover.tsx
```diff
@@ -217,8 +217,18 @@ export class CICheckRunPopover extends React.PureComponent<
...
+    checkRun: IRefCheck,
+    step: IAPIWorkflowJobStep
+  ): void => {
+    const url = checkRun.htmlUrl
+      ? `${checkRun.htmlUrl}/#step:${step.number}:1`
+      : `${this.props.repository.htmlURL}/pull/${this.props.prNumber}`
+    if (url === null) {
+      // The repository should have a htmlURL.
+      return
+    }
```
sergiou87: This is impossible, right? Since we're using template strings, `url` can't be null.
```suggestion
```

sergiou87 (changes requested):
Code LGTM overall, only one minor thing!

In any case, I'm not sure how to feel about having so many tooltips. I *think* it'd work better if the default action when you click an element is for expanding/collapsing the steps, and then on hover show an additional icon to view on GitHub. What do you think?

app/src/ui/check-runs/ci-check-run-popover.tsx
tidy-dev: Yep. I think that null check was left over from a different approach. Thanks! I updated both instances of using this.

tidy-dev:
> In any case, I'm not sure how to feel about having so many tooltips. I think it'd work better if the default action when you click an element is for expanding/collapsing the steps, and then on hover show an additional icon to view on GitHub. What do you think?

Yeah. On the [private issue ](https://github.com/github/desktop/issues/710), I sort mention that . It definitely feels odd, but I am not sure how what you suggest would work with the visually showing expand/collapse chevrons.. which distinguish the expandable vs non expandable rows.

tidy-dev:
@sergiou87 Based on today's discussion, going to update that list item interaction. But, would like to go ahead and merge this and build on top of it.

sergiou87 (approved):
LGTM!

PR #13326, opened 2021-11-15
Fix click area to select hunk on both sides of diff

### LINKED ISSUE #13325: Click area to select hunk on both sides of diff is wrong
### Describe the bug
In split diff display, the click area to select a hunk on both sides of a diff is wrong (limited to the left side).
 
### Version & OS

2.9.4 (x64), Windows 10.

1. Make a change to some file in a repository.
2. Select it in the Changes tab.
3. Make sure you have "Split" diff display selected in the settings under the gear icon in the top right corner.
4. Hover over the area between the line numbers of the changed lines.

### Expected behavior

The click area is stretched from the left to the right side of the center area.

### Actual behavior

The click area is limited to the *left* side of the center area.

### Screenshots

To select a hunk on both sides of a diff, I can perform a click right in the center of the diff: 
image

In practice, however, the click area is limited to a small space between the line numbers on the left and the center:
image

### Additional context

The reason for this bug is that the click area (the `.hunk-handle` element), while being positioned correctly, is obstructed by the `.after` element. A possible solution is to set a `z-index` to the `.hunk-handle` element or move it after the `.after` element.
image

jwbth:
Closes #13325

## Description

Add a `z-index` to elements responsible for selecting a hunk on the both sides of a diff, so that the click area is above the row and not obstructed by its right-side part.

### Screenshots

Restored click area:
image

### Gifs
Before: 
w7EtEmWaCW
After:
tT91g5D0vv

## Release notes

Notes:

### DIFF
--- a/app/styles/ui/_side-by-side-diff.scss
+++ b/app/styles/ui/_side-by-side-diff.scss
@@ -138,6 +138,7 @@
     width: 20px;
     left: calc(50% - 10px);
     top: 0;
+    z-index: 1;
 
     &.hoverable {
       cursor: pointer;
--- a/app/styles/ui/_side-by-side-diff.scss
+++ b/app/styles/ui/_side-by-side-diff.scss
@@ -138,7 +138,6 @@
     width: 20px;
     left: calc(50% - 10px);
     top: 0;
-    z-index: 1;
 
     &.hoverable {
       cursor: pointer;
--- a/app/src/ui/diff/side-by-side-diff-row.tsx
+++ b/app/src/ui/diff/side-by-side-diff-row.tsx
@@ -204,12 +204,12 @@ export class SideBySideDiffRow extends React.Component<
               {this.renderContentFromString('')}
               {this.renderWhitespaceHintPopover(DiffColumn.Before)}
             </div>
-            {this.renderHunkHandle()}
             <div className="after">
               {this.renderLineNumber(lineNumber, isSelected)}
               {this.renderContent(row.data)}
               {this.renderWhitespaceHintPopover(DiffColumn.After)}
             </div>
+            {this.renderHunkHandle()}
           </div>
         )
       }
@@ -241,12 +241,12 @@ export class SideBySideDiffRow extends React.Component<
               {this.renderContent(row.data)}
               {this.renderWhitespaceHintPopover(DiffColumn.Before)}
             </div>
-            {this.renderHunkHandle()}
             <div className="after">
               {this.renderLineNumber()}
               {this.renderContentFromString('')}
               {this.renderWhitespaceHintPopover(DiffColumn.After)}
             </div>
+            {this.renderHunkHandle()}
           </div>
         )
       }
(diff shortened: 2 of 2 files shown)


app/styles/ui/_side-by-side-diff.scss
```diff
@@ -138,6 +138,7 @@
     width: 20px;
     left: calc(50% - 10px);
     top: 0;
+    z-index: 1;
```
sergiou87: We'd like to avoid setting the `z-index` because it could end up leading to complex situations where we have to assign random z-index number to different elements to make everything work as expected.

Instead, I've looked into this particular bug, and it seems we can fix it by removing this line…
```suggestion
```
…and reordering the hunk handle when it's rendered. You can apply these changes to fix it in all cases:
```diff
diff --git a/app/src/ui/diff/side-by-side-diff-row.tsx b/app/src/ui/diff/side-by-side-diff-row.tsx
index f1decce05..bd94942b4 100644
--- a/app/src/ui/diff/side-by-side-diff-row.tsx
+++ b/app/src/ui/diff/side-by-side-diff-row.tsx
@@ -204,12 +204,12 @@ export class SideBySideDiffRow extends React.Component<
               {this.renderContentFromString('')}
               {this.renderWhitespaceHintPopover(DiffColumn.Before)}
             </div>
-            {this.renderHunkHandle()}
             <div className="after">
               {this.renderLineNumber(lineNumber, isSelected)}
               {this.renderContent(row.data)}
               {this.renderWhitespaceHintPopover(DiffColumn.After)}
             </div>
+            {this.renderHunkHandle()}
           </div>
         )
       }
@@ -241,12 +241,12 @@ export class SideBySideDiffRow extends React.Component<
               {this.renderContent(row.data)}
               {this.renderWhitespaceHintPopover(DiffColumn.Before)}
             </div>
-            {this.renderHunkHandle()}
             <div className="after">
               {this.renderLineNumber()}
               {this.renderContentFromString('')}
               {this.renderWhitespaceHintPopover(DiffColumn.After)}
             </div>
+            {this.renderHunkHandle()}
           </div>
         )
       }
@@ -259,12 +259,12 @@ export class SideBySideDiffRow extends React.Component<
               {this.renderContent(before)}
               {this.renderWhitespaceHintPopover(DiffColumn.Before)}
             </div>
-            {this.renderHunkHandle()}
             <div className="after" onMouseEnter={this.onMouseEnterLineNumber}>
               {this.renderLineNumber(after.lineNumber, after.isSelected)}
               {this.renderContent(after)}
               {this.renderWhitespaceHintPopover(DiffColumn.After)}
             </div>
+            {this.renderHunkHandle()}
           </div>
         )
       }
```

One of the changes affect added lines, which already worked… except for this annoying glitch:
Kapture 2021-11-16 at 13 24 56

With the diff above, it will keep the same cursor over the area of the whole hunk handle 😄

sergiou87 (changes requested):
Nice catch @jwbth !! Thank you very much for reporting and fixing this 😄 

However, we think there is a better fix for this bug 🤞  I hope you don't mind applying the suggested changes.

Thanks for your patience!

app/styles/ui/_side-by-side-diff.scss
jwbth: Thanks, that was the other possible solution that I was thinking. I've applied the suggested changes 🙂

sergiou87 (approved):
Thank you very much for your contribution!!! ❤️

PR #13327, opened 2021-11-15
Revert split action list item - Use title as link

tidy-dev:
Based on #13317 

## Description

The click list item to view online did not quite fit the Desktop UI/felt a bit unexpected. Switched back to the list item expand/collapsing and hover of title appears as link with blue text.

Got more UI feedback for using links as opposed to link external icon on private issue since team sync.

### Screenshots

https://user-images.githubusercontent.com/75402236/141828886-1c6d1c1e-fb99-4ed0-b04f-5111877c6429.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/lib/tooltip.tsx
+++ b/app/src/ui/lib/tooltip.tsx
@@ -7,6 +7,16 @@ import { assertNever } from '../../lib/fatal-error'
 import { rectEquals, rectContains } from './rect'
 
 export type TooltipDirection = 'n' | 'ne' | 'e' | 'se' | 's' | 'sw' | 'w' | 'nw'
+export enum TooltipDirectionDef {
+  NORTH = 'n',
+  NORTH_EAST = 'ne',
+  EAST = 'e',
+  SOUTH_EAST = 'se',
+  SOUTH = 's',
+  SOUTH_WEST = 'sw',
+  WEST = 'w',
+  NORTH_WEST = 'nw',
+}
 const DefaultTooltipDelay = 400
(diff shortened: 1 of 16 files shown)


app/src/ui/lib/tooltip.tsx
```diff
@@ -7,6 +7,16 @@ import { assertNever } from '../../lib/fatal-error'
...
+export enum TooltipDirectionDef {
+  NORTH = 'n',
+  NORTH_EAST = 'ne',
+  EAST = 'e',
+  SOUTH_EAST = 'se',
+  SOUTH = 's',
+  SOUTH_WEST = 'sw',
+  WEST = 'w',
+  NORTH_WEST = 'nw',
+}
```
sergiou87: Why do we need two of these? Can't we just use the `enum`?

Or, at least, maybe build the `TooltipDirection` type using the enum values, no? 🤔

sergiou87 (commented):
Code LGTM! Only one minor thing.

I also tested it and it feels sooo much better than adding the icon that we discussed, I like it! 😍

app/src/ui/lib/tooltip.tsx
tidy-dev: .... Fear of regression. I just pushed a commit that updates 11 files to do this. :) I do prefer the enum everywhere.

sergiou87 (approved):
Thanks for those changes! They look good, and I tested it and didn't find regressions and worked great too 🐸 🕴️ ❗

PR #13343, opened 2021-11-18
Warn hidden bidi characters in diffs

sergiou87:
## Description

This PR adds a file-level warning when hidden bidi characters are found in diffs. For this, I've used the same text and style used in dotcom.

### Screenshots

https://user-images.githubusercontent.com/1083228/142419897-38148e72-0522-4908-b79e-1b0fd713e941.mov

With dark theme:
image

## Release notes

Notes: [Improved] Warn users when files contain bidirectional Unicode text

### DIFF
--- a/app/src/lib/diff-parser.ts
+++ b/app/src/lib/diff-parser.ts
@@ -375,15 +375,13 @@ export class DiffParser {
       throw new Error('Malformed diff, empty hunk')
     }
 
-    const hunk = new DiffHunk(
+    return new DiffHunk(
       header,
       lines,
       linesConsumed,
       linesConsumed + lines.length - 1,
       getHunkHeaderExpansionType(hunkIndex, header, previousHunk)
     )
-
-    return hunk
   }
 
   /**
--- a/app/src/lib/diff-parser.ts
+++ b/app/src/lib/diff-parser.ts
@@ -24,7 +24,7 @@ const diffHeaderRe = /^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@/
 
 /**
  * Regular expression matching invisible bidirectional Unicode characters that
- * may be interpreted or compiled differently than what it appears. See more:
+ * may be interpreted or compiled differently than what it appears. More info:
  * https://github.co/hiddenchars
  */
 export const HiddenBidiCharsRegex = /[\u202A-\u202E]|[\u2066-\u2069]/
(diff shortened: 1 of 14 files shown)


tidy-dev:
Imho the warning is sort of misleading. It says “This file contains…“, but it only shows when those characters are in the current diff (which I understand is what we have access to run the regex on at the moment). But… with those words as an unknowing user, I would be expecting it to run on the whole file and when you expand/collapse it feels weird that it goes away.  Wondering if we can’t do it on the whole file, if the verbiage should change to say "This diff contains..."

sergiou87:
Sounds good!

app/src/lib/diff-parser.ts
```diff
@@ -22,6 +22,8 @@ import { getLargestLineNumber } from '../ui/diff/diff-helpers'
 // in which case s defaults to 1
 const diffHeaderRe = /^@@ -(\d+)(?:,(\d+))? \+(\d+)(?:,(\d+))? @@/
 
+export const HiddenBidiCharsRegex = /[\u202A-\u202E]|[\u2066-\u2069]/
```
niik, line 30: Could we please add a little explanatory comment here since it's being exported?

app/src/lib/diff-parser.ts
```diff
@@ -438,6 +461,7 @@ export class DiffParser {
         hunks,
         isBinary: headerInfo.isBinary,
         maxLineNumber: getLargestLineNumber(hunks),
+        hasHiddenBidiChars,
```
niik: Couldn't we just run the regex here instead?

```suggestion
        hasHiddenBidiChars: HiddenBidiCharsRegex.test(text),
```

niik (changes requested):
Great work @sergiou87. I'm curious though as to why we're matching on each line instead of simply running the regex on the whole diff text which we receive in the parse function? What is it that I'm missing?

sergiou87:
> why we're matching on each line instead of simply running the regex on the whole diff text which we receive in the parse function? What is it that I'm missing?

I'd say lack of proper sleep, sorry 😓   It's mainly a leftover of my initial approach (that we discussed on Monday) of looking for these characters only in added lines, then after our chat on Monday I switched to show the warning whenever those chars are found anywhere in the diff, but I wanted to try and show _something_ per line too. Finally, I realized it wasn't worth the effort and started cleaning up my mess but I guess I stopped early 🤦‍♂️

I'll switch to that approach!

sergiou87:
@niik all fixed in my latest commits. Thank you! 🙇‍♂️

niik (approved):
💖

PR #13344, opened 2021-11-18
Differentiate check runs by event

tidy-dev:
## Description
During the beta release, I was watching the release branch with the new check runs component and noticed that it did not show the job steps for the check runs like it normally would.. and on dotcom the check runs were listed twice one with "(push)" after the name and one with "(pull_request)" after the name.

Found out that release branches... tho seems I can't replicate on a test release... :/ can have check-runs that are triggered by `push` events from a generated branch during release.

Our existing check run logic will retrieve all check runs for a pull request git reference from the checks api. This means if a pull request check runs have been rerun due to additional commits or just being rerun, we get all of them back. Those types are `pull_request` events. Thus, we were filtering to the ones of the same name with the highest check run id (most recent). Unfortunately, ones that are run with a `push` also have the same name but are treated separately on dotcom's list as they have different origination.. not exactly just a re-run. Since, the `push` events occurred on deployment of the release branch, they were the most recent ones. Thus, those are the ones we filtered to and kept to show in the check run list. 

First problem is now we only show each one once where dotcom shows them twice. This PR addresses that by also taking into account if a check run has a pull request associated with it as that means it is the `pull_request` event type... a little hacky but check runs from the checks api do not have a concept of event type as that is an actions construct and check runs on checks api is check runner agnostic.

Second problem is the check runs job steps do not show for `push` event check runs. This was because we make a call to the actions api to get all the workflow runs for the branch name. Unfortunately, those `push` events on the release pull request were not actually made by that same branch; it was made with another generated branch but presumedly for the same commit/git ref (how it gets associated with the PR)... Thus, the workflow runs call does not pull the workflow runs associated with the `push` event check runs back in this case. Thus I have opened [an issue](https://github.com/github/c2c-actions-checks/issues/177) with the actions team requesting a way to retrieve these workflow runs (either using the git ref like check runs api does or by check suite id since that is what we are ultimately matching check runs to action workflows with). 

_Other notes: After thinking on it more.. we technically could address the second problem as we can retrieve action workflows runs by date and we have a create date on the  check run. Thus, we could retrieve all workflow runs with the same date. But that would retrieve them for any other pull request/workflow run from that date.. which for some repos may mean retrieving numerous unrelated workflow runs that have nothing to do with the that desktop user and then also having to deal with pagination limitations... definitely not ideal._

### Screenshots
... Shouldn't have merged the beta release...  after that my testing came from just hard coding the different types in as I couldn't reproduce on a test release.. must have a different flow.

This is how it will look at this point in time for push events (duplicate rows without the expansion arrow):
image

This is how it will look when we can differentiate between push and pull_request events if action teams makes it so we can get the workflow runs for push events. Note: I only append the event if there are more than one kind of event since this is not a common. If I get sticky headers working, will probably have multiple headers for different events like. CI (push) and CI (pull_request).
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list.tsx
@@ -77,8 +77,7 @@ export class CICheckRunList extends React.PureComponent<
     )
     return {
       checkRunExpanded,
-      hasUserToggledCheckRun:
-        currentState !== null ? currentState.hasUserToggledCheckRun : false,
+      hasUserToggledCheckRun: currentState?.hasUserToggledCheckRun || false,
       checkRunsHaveMultipleEventTypes: checkRunEvents.size > 1,
     }
   }
--- a/app/src/ui/check-runs/ci-check-run-list.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list.tsx
@@ -30,6 +30,7 @@ interface ICICheckRunListProps {
 interface ICICheckRunListState {
   readonly checkRunExpanded: string | null
   readonly hasUserToggledCheckRun: boolean
+  readonly checkRunsHaveMultipleEventTypes: boolean
 }
 
 /** The CI Check list. */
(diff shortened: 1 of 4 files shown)


app/src/ui/check-runs/ci-check-run-list.tsx
```diff
@@ -69,11 +70,16 @@ export class CICheckRunList extends React.PureComponent<
...
+        .map(c => c.actionsWorkflow?.event)
+        .filter(c => c !== undefined && c.trim() !== '')
+    )
     return {
       checkRunExpanded,
-      hasUserToggledCheckRun: currentState
-        ? currentState.hasUserToggledCheckRun
-        : false,
+      hasUserToggledCheckRun:
+        currentState !== null ? currentState.hasUserToggledCheckRun : false,
```
sergiou87: Not sure if this is better or worse, but is shorter and should be equivalent 😂 
```suggestion
      hasUserToggledCheckRun: currentState?.hasUserToggledCheckRun || false,
```

sergiou87 (dismissed):
Looks good to me!! Nice catch 😱

app/src/ui/check-runs/ci-check-run-list.tsx
tidy-dev: I like it!

sergiou87 (approved):

PR #13348, opened 2021-11-19
CI Check run Sticky Headers

tidy-dev:
## Description
Add group headers for the different types of action workflow app types to remove redundant name convention of {app name} / {check run name}

Follow up PRs:
- Need to adjust the "find first failed checkrun" logic as now it finds it and then does the grouping/sorting so the first one may not be first anymore. :)
- Add event type groupings in the headers.  Example: If you have under CI, Windows x64 (pull request) and Windows x64 (push), we will split this into groups of CI (pull request) and CI (push) with check runs of Windows x64 underneath. Expecting there would be multiple in each group.

### Screenshots

https://user-images.githubusercontent.com/75402236/142648080-0ab6eca7-45cc-406d-b84a-de8c7761d138.mov

## Release notes
Notes: [Improved] Add group headers to CI Check run list

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -120,7 +120,7 @@ export class CICheckRunListItem extends React.PureComponent<
       sticky: isCheckRunExpanded,
     })
     return (
-      <>
+      <div className="ci-check-list-item-group">
         <div className={classes} onClick={this.toggleCheckRunExpansion}>
           {this.renderCheckStatusSymbol()}
           {this.renderCheckRunName()}
@@ -132,7 +132,7 @@ export class CICheckRunListItem extends React.PureComponent<
             onViewJobStep={this.onViewJobStep}
           />
         ) : null}
-      </>
+      </div>
     )
   }
 }
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -116,12 +116,12 @@ export class CICheckRunListItem extends React.PureComponent<
   public render() {
     const { checkRun, isCheckRunExpanded } = this.props
 
+    const classes = classNames('ci-check-list-item', 'list-item', {
+      sticky: isCheckRunExpanded,
+    })
     return (
       <>
-        <div
-          className="ci-check-list-item list-item"
-          onClick={this.toggleCheckRunExpansion}
-        >
+        <div className={classes} onClick={this.toggleCheckRunExpansion}>
           {this.renderCheckStatusSymbol()}
           {this.renderCheckRunName()}
           {this.renderCheckJobStepToggle()}
--- a/app/src/ui/check-runs/ci-check-run-list.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list.tsx
@@ -1,10 +1,7 @@
+import _ from 'lodash'
 import * as React from 'react'
 import { IAPIWorkflowJobStep } from '../../lib/api'
-import {
-  getCheckRunDisplayName,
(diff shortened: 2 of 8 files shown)


app/src/ui/check-runs/ci-check-run-list.tsx
```diff
@@ -104,24 +101,88 @@ export class CICheckRunList extends React.PureComponent<
...
-        getCheckRunDisplayName(
-          a,
-          this.state.checkRunsHaveMultipleEventTypes
-        ).localeCompare(
-          getCheckRunDisplayName(b, this.state.checkRunsHaveMultipleEventTypes)
-        )
-      )
+  private getCheckRunsGroupedByWorkflowAction = (): _.Dictionary<
+    IRefCheck[]
+  > => {
```
sergiou87, line 106: Is `_.groupBy` the reason for not using a `Map<string, IRefCheck[]>` instead? AFAICT we don't use `_.Dictionary` anywhere else 🤔 

This is not an issue, it just called my attention 😅

app/src/ui/check-runs/ci-check-run-list.tsx
```diff
@@ -104,24 +101,88 @@ export class CICheckRunList extends React.PureComponent<
...
+    const groupNameSet = new Set<string>()
+
+    for (const checkRun of this.props.checkRuns) {
+      let name = checkRun.actionsWorkflow?.name || 'Other'
+      if (name === 'Other' && checkRun.appName === 'GitHub Code Scanning') {
+        name = 'Code scanning results'
+      }
+      if (groupNameSet.has(name)) {
+        continue
+      }
```
sergiou87, line 148: Do you need this given it's a `Set`? There should be no duplicates 🤔

app/styles/ui/check-runs/_ci-check-run-list-item.scss
```diff
@@ -4,6 +4,13 @@
   border-bottom: var(--base-border);
   height: auto;
 
+  &.sticky {
+    position: sticky;
```
sergiou87, line 8: TIL 🤯

app/styles/_variables.scss
```diff
@@ -309,6 +309,9 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
   --nudge-arrow-z-index: 16;
   --drag-overlay-z-index: 15;
   --side-panel-z-index: 14;
+  // We want second level to fall under top level when scrolling
+  --list-sticky-header-top-level: 11;
+  --list-sticky-header-second-level: 10;
```
sergiou87: Nit: should we suffix these with `-z-index` for more clarity? Or mention z-index in the comment above?

sergiou87 (commented):
Only a few questions/nits, otherwise looks and works great! 😍

app/src/ui/check-runs/ci-check-run-list.tsx
tidy-dev: Well in the next PR.. I get rid of it since the group logic got more complicated. And Yes, that lodash returns the lodash type of dictionary.. which is basically a map.

sergiou87 (approved):
✨

app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -124,10 +124,15 @@ export class CICheckRunListItem extends React.PureComponent<
 
     const classes = classNames('ci-check-list-item', 'list-item', {
       sticky: isCheckRunExpanded,
+      selected: this.props.selected,
     })
     return (
       <div className="ci-check-list-item-group">
-        <div className={classes} onClick={this.toggleCheckRunExpansion}>
```
ewabor441: Fdjb

PR #13360, opened 2021-11-22
Select actual first failed check run

tidy-dev:
## Description

Since updates to grouping/sorting methods the first failed check runs logic didn't correctly get the first failed check run as it was searching on the unsorted list.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/ci-checks/ci-checks.ts
+++ b/app/src/lib/ci-checks/ci-checks.ts
@@ -615,7 +615,7 @@ export function getCheckRunsGroupedByActionWorkflowNameAndEvent(
 
   sortedGroupNames.forEach(gn => {
     const group = groups.get(gn)
-    if (group) {
+    if (group !== undefined) {
       const sortedGroup = group.sort((a, b) => a.name.localeCompare(b.name))
       groups.set(gn, sortedGroup)
     }
--- a/app/src/lib/ci-checks/ci-checks.ts
+++ b/app/src/lib/ci-checks/ci-checks.ts
@@ -574,6 +574,7 @@ export function getCheckRunStepURL(
 /**
  * Groups check runs by their actions workflow name and actions workflow event type.
  * Event type only gets grouped if there are more than one event.
+ * Also sorts the check runs in the groups by their names.
  *
  * @param checkRuns
  * @returns A map of grouped check runs.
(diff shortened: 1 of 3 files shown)


app/styles/ui/check-runs/_ci-check-run-list.scss
```diff
@@ -1,6 +1,7 @@
 .ci-check-run-list {
   overflow-y: auto;
   overflow-x: hidden;
+  scroll-padding-top: 35px;
```
tidy-dev, line 4: Found out that the scroll-margin-top on the job step list items for some reason doesn't work if the first failed step is in the first workflow action group... This makes it work... but unfortunately, does push the non-first groups ones down a step (making a non-failed step visible right before it). I may revisit this to see if there is a solution to make it uniform (first failed step being always the first one seen). But at least this way the first failed step is visible in both cases.

app/src/lib/ci-checks/ci-checks.ts
```diff
@@ -610,6 +611,16 @@ export function getCheckRunsGroupedByActionWorkflowNameAndEvent(
     groups.set(group, newGroup)
   }
 
+  const sortedGroupNames = getCheckRunGroupNames(groups)
+
+  sortedGroupNames.forEach(gn => {
+    const group = groups.get(gn)
+    if (group) {
```
sergiou87: Nit: explicit check?
```suggestion
    if (group !== undefined) {
```

app/src/ui/check-runs/ci-check-run-list.tsx
```diff
@@ -52,30 +53,46 @@ export class CICheckRunList extends React.PureComponent<
...
-      // If there is a failure, we want the first check run with a failure, to
-      // be opened so the user doesn't have to click through to find it.
-      // Otherwise, just open the first one. (Only actions type can be expanded.)
-      const firstFailure = props.checkRuns.find(
+    // If there is a failure, we want the first check run with a failure, to
+    // be opened so the user doesn't have to click through to find it.
+    let firstFailure: IRefCheck | undefined
+    checkRunExpanded = null
+    for (const group of checkRunGroups.values()) {
+      firstFailure = group.find(
```
sergiou87: Not sure if I'm missing anything, but is there a reason to move this out of the `for`? AFAICT `firstFailure` is not being changed outside of the `for` so it can still be `const`, right? 🤔

sergiou87 (changes requested):

app/src/ui/check-runs/ci-check-run-list.tsx
tidy-dev: Yep - Sorry left over from old approach.

sergiou87 (approved):
LGTM! ✨

PR #13369, opened 2021-11-23
Rerun check run dialog

tidy-dev:
## Description

A dialog to communicate to users which check runs will be rerun as not all check runs can be.

Something left to figure out.. Currently, if you have a check job steps expanded and then hit rerun. It manually updates the check statuses to pending, but this triggers a check job step retrieval.. but it is too fast and just retrieves the stale job steps.

### Screenshots

image

## Release notes
Notes: [Improved] Added rerun check dialog to know which checks will be rerun.

### DIFF
--- a/app/src/lib/stores/commit-status-store.ts
+++ b/app/src/lib/stores/commit-status-store.ts
@@ -19,7 +19,7 @@ import {
   getLatestCheckRunsByName,
   apiStatusToRefCheck,
   getLatestPRWorkflowRunsLogsForCheckRun,
-  getCheckRunActionsJobsAndLogURLS,
+  getCheckRunActionsWorkflowRuns,
 } from '../ci-checks/ci-checks'
 
 interface ICommitStatusCacheEntry {
@@ -401,10 +401,10 @@ export class CommitStatusStore {
   }
 
   /**
-   * Retrieve GitHub Actions workflows and populates the job and log url if
-   * applicable to the checkruns
+   * Retrieve GitHub Actions workflows and maps them to the check runs if
+   * applicable
    */
-  public async getCheckRunActionsJobsAndLogURLS(
+  public async getCheckRunActionsWorkflowRuns(
     repository: GitHubRepository,
     ref: string,
     branchName: string,
@@ -423,7 +423,7 @@ export class CommitStatusStore {
     }
 
     const api = API.fromAccount(account)
-    return getCheckRunActionsJobsAndLogURLS(
+    return getCheckRunActionsWorkflowRuns(
       api,
       owner,
       name,
--- a/app/src/lib/stores/commit-status-store.ts
+++ b/app/src/lib/stores/commit-status-store.ts
@@ -4,7 +4,7 @@ import QuickLRU from 'quick-lru'
 import { Account } from '../../models/account'
 import { AccountsStore } from './accounts-store'
 import { GitHubRepository } from '../../models/github-repository'
-import { API, getAccountForEndpoint } from '../api'
+import { API, getAccountForEndpoint, IAPICheckSuite } from '../api'
 import { IDisposable, Disposable } from 'event-kit'
 import {
   ICombinedRefCheck,
@@ -427,5 +427,19 @@ export class CommitStatusStore {
 
     const api = API.fromAccount(account)
     return api.rerequestCheckSuite(owner.login, name, checkSuiteId)
+  }
+
+  public async fetchCheckSuite(
+    repository: GitHubRepository,
+    checkSuiteId: number
+  ): Promise<IAPICheckSuite | null> {
+    const { owner, name } = repository
+    const account = getAccountForEndpoint(this.accounts, repository.endpoint)
+    if (account === null) {
+      return null
+    }
+
+    const api = API.fromAccount(account)
+    return api.fetchCheckSuite(owner.login, name, checkSuiteId)
   }
 }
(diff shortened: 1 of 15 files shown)


app/src/lib/stores/commit-status-store.ts
```diff
@@ -236,6 +241,44 @@ export class CommitStatusStore {
...
+
+      updatedChecks.push({
+        ...check,
+        status: APICheckStatus.InProgress,
+        conclusion: null,
+        actionJobSteps: undefined,
+      })
+    }
+
+    const check = createCombinedCheckFromChecks(updatedChecks)
```
sergiou87, line 277: Could this belong to `ci-checks.ts`? I'd say we shouldn't use this dialog for my "checks failed" dialog, but in any case I think it makes sense to have all the generic CI-checks stuff there. What do you think?

app/src/ui/check-runs/ci-check-run-list.tsx
```diff
@@ -63,7 +66,10 @@ export class CICheckRunList extends React.PureComponent<
   ): ICICheckRunListState {
     let checkRunExpanded = currentState?.checkRunExpanded ?? null
 
-    if (currentState === null || !currentState.hasUserToggledCheckRun) {
+    if (
+      (currentState === null || !currentState.hasUserToggledCheckRun) &&
+      !this.props.notExpandable
+    ) {
```
sergiou87: Nit: Alternative? Is this better? 🤷  (might require linting)
```suggestion
    if (currentState?.hasUserToggledCheckRun !== true && !this.props.notExpandable) {
```

app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+          : null}
+      </Row>
+    )
+  }
+
+  public render() {
+    return (
+      <Dialog
+        id="rerun-check-runs"
+        title={__DARWIN__ ? 'Re-run checks' : 'Re-run checks'}
```
sergiou87:
```suggestion
        title={__DARWIN__ ? 'Re-run Checks' : 'Re-run checks'}
```

app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+        loading={this.state.loadingCheckSuites || this.state.loadingRerun}
+      >
+        <DialogContent>{this.renderRerunnableJobsList()}</DialogContent>
+        <DialogFooter>
+          {this.renderRerunInfo()}
+          <OkCancelButtonGroup
+            okButtonText={__DARWIN__ ? 'Re-run checks' : 'Re-run checks'}
```
sergiou87:
```suggestion
            okButtonText={__DARWIN__ ? 'Re-run Checks' : 'Re-run checks'}
```

app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+        <Octicon symbol={OcticonSymbol.alert} />
+
+        {this.state.rerunnable.length === 0
+          ? `There are no checks that can be rerun. `
+          : `There ${verb} ${this.state.nonRerunnable.length} ${pluralize} that cannot be rerun. `}
+
+        {this.state.nonRerunnable.length > 0
+          ? `A check run cannot be rerun if the check is more than one month old,
+          the check has not completed, or the check is not configured to be
+          rerun.`
```
sergiou87: Should we use re-run or rerun? (the dialog title, submit button and this message use different spelling).
```suggestion
          ? `There are no checks that can be re-run. `
          : `There ${verb} ${this.state.nonRerunnable.length} ${pluralize} that cannot be re-run. `}

        {this.state.nonRerunnable.length > 0
          ? `A check run cannot be re-run if the check is more than one month old,
          the check has not completed, or the check is not configured to be
          re-run.`
```

sergiou87 (changes requested):

app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
tidy-dev: Good point... and I know from reading stuff on actions... there is debate on should it be.. re-rerun, retry, or rerequest.. :P  But, dotcom uses Re-run so I think it does make sense to follow suite. :)

app/src/lib/stores/commit-status-store.ts
tidy-dev: That makes sense to me. I am going to do another PR to better handle the job steps status not being reflective of actual state. In that one, I will move any check logic (like this) into ci-checks.ts... cause I think I am going to move more of it out of the list component logic closer to the status retrieve/caching anyways.

sergiou87: SGTM!

sergiou87 (approved):
I wasn't able to expand any of the jobs, not sure if that's expected, despite they show the expand/collapse arrow.

Also, a nit: since the dialog starts being super small, when the data is loaded and it is expanded, it appears misplaced and not centered.

I think all these can be tackled in a different PR, but I thought they were worth mentioning 😄

PR #13390, opened 2021-11-26
Checks failed dialog UX improvements

sergiou87:
## Description

This PR introduces some UX improvements to the checks failed dialog suggested in https://github.com/desktop/desktop/pull/13207#pullrequestreview-810016263 and https://github.com/desktop/desktop/pull/13207#issuecomment-974272620

Things I'm not 100% convinced about:
- Removing the commit SHA and message. I think that might be important info when looking at checks that failed.
- The alignment of the check icons feels somehow uneven since I had to add a margin to align them with the dialog icon.
- The re-run jobs button is displaced to the left to make room for the loading indicator when needed.

From an implementation-wise perspective, I change a few things in the `dialog-header` and `dialog-content` classes that feel too hacky, maybe I should adapt the Dialog API to accommodate these new needs.

### Screenshots

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list.tsx
@@ -167,8 +167,6 @@ export class CICheckRunList extends React.PureComponent<
   }
 
   public render() {
-    return (
-      <div className="ci-check-run-list-container">{this.renderList()}</div>
-    )
+    return <div className="ci-check-run-list">{this.renderList()}</div>
   }
 }
--- a/app/src/ui/notifications/pull-request-checks-failed.tsx
+++ b/app/src/ui/notifications/pull-request-checks-failed.tsx
@@ -19,7 +19,6 @@ import * as OcticonSymbol from '../octicons/octicons.generated'
 import { Button } from '../lib/button'
 import { RepositoryWithGitHubRepository } from '../../models/repository'
 import { CICheckRunActionsJobStepList } from '../check-runs/ci-check-run-actions-job-step-list'
-import { truncateWithEllipsis } from '../../lib/truncate-with-ellipsis'
 import { LinkButton } from '../lib/link-button'
 import { encodePathAsUrl } from '../../lib/path'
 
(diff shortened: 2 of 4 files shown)


tidy-dev:
> Removing the commit SHA and message. I think that might be important info when looking at checks that failed.

Just out of curiosity - because I haven't used/looked for this information when doing this action myself.. maybe I am missing out on something :D. When/why do you use or like having this info? 

Could have another line under the pr title line for it?  
Maybe making header too busy, but if it is important info.. seems it could be a logical place. This way the header lines read:
- "X checks that failed on a pr" (What is this dialog about?) 
- "PR title" (What PR?) 
-  "commit - commit title" (What change caused this? - who do we blame!!!! :P)

sergiou87:
> Just out of curiosity - because I haven't used/looked for this information when doing this action myself.. maybe I am missing out on something :D. When/why do you use or like having this info?

Sometimes I'm looking at a check run that is old and doesn't belong to the latest commit I pushed (for example) and waste a few minutes trying to think what went wrong 😆 

> Could have another line under the pr title line for it?

As you say that would make it too busy, so in a first attempt I just picked the previous "sub-header" for the commit sha/message and and put it as a row between the dialog header and the list of checkruns (and their steps). It didn't look too bad, but I guess it made hard to scan the data structure as Adrian said.

> who do we blame!!!! :P

It's always me!

image

tidy-dev (dismissed):
✨  Looks good to me.  

I see what you are saying on the icons looking disjoint... but can't offer any help.. I feel like it is a fall out of the headers being disjoint. But not sure if aligning the headers makes sense either. 🤷  .. sorry no help.

image

Different note, if there is a lack of scroll bar on check list (small number of checks), the differentiation of the two lists feels a little blurred to me. But may just be personal preference.

tidy-dev:
I went ahead and approved - figured if you decided to add the commit message back in somewhere. I can re-approve. :)

sergiou87:
Thanks! I'd say we merge it and we can iterate later when we have more feedback (you will need to reapprove it though, once I remove the commit I added for fast testing, sorry 😅 ).

tidy-dev (approved):

PR #13405, opened 2021-11-29
Show jobs as pending after checks are rerun

tidy-dev:
## Description

TLDR; Timing and cache issues caused job steps to not show pending; moved job steps into cache so they would show pending. Also, addressed toggle indicator showing when it shouldn't.

An issue with current rerun dialog is that after you rerun it, I manually update the checks to be pending because an immediate re-retrieve will be stale as the rerun has not triggered yet and we don't know when the rerun will fire exactly.

However, the job steps were not stored in the cache they were retrieved on the fly when the user opens the popover. Thus, I could not modify their state such that it would reflect in the dialog.

Therefore, I moved jobs steps also to be store in the commit status store cache. This was problematic before because we do not want to retrieve jobs steps (and previously logs) for all pull requests when the user may not even open the check run popover for more than the one they are working on. Thus, I have set an optional branch name for the subscriptions such that it only retrieve them if the branch name is set (something required to retrieve them anyways). Thus, by default, it won't be set, but when the user opens the popover, we resubscribe for that pull request and when doing so update the subscription to have the branch name and go get the job steps.

This is a two for one improvement as now the job steps are cached and no longer have to be retrieved every time the user opens the popover. But, for the point of this PR, we can set the job steps in the cache to pending until the next cycle of commit status store retrieve.

Unfortunately... there is still a timing issue here to note - if user hits rerun at the same time as the commit-status-store cycles they could end up with stale data in the popover. But, I think this is not terribly likely, and not as bad as trying to set a timer to guess when the rerun happens on the server.

Another not so nice about this is, the links for the check runs in the popover or for the run before the pending one, so at this point if the user clicks on the link they will go to one on dotcom that tells them they are one attempt behind... but we can't remedy this until we get a fresh retrieve back.. only option would be to disable them.. but I think that is worse. :/

Also addressed some UI feedback from Sergio in last PR -> toggle indicators hidden (I had that and then changed approach and didn't realize those came back.) and found out the job ids.. do match check run ids so I could match on an id and not a name.. I believe this may have changed with recent actions update.

### Screenshots

https://user-images.githubusercontent.com/75402236/143910354-a0a076f5-47d7-4085-9afd-4029be5230e0.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/commit-status-store.ts
+++ b/app/src/lib/stores/commit-status-store.ts
@@ -342,8 +342,7 @@ export class CommitStatusStore {
     if (
       existingChecks !== undefined &&
       existingChecks.check !== null &&
-      existingChecks.check.checks.find(c => c.actionsWorkflow !== undefined) !==
-        undefined &&
+      existingChecks.check.checks.some(c => c.actionsWorkflow !== undefined) &&
       _.xor(
         existingChecks.check.checks.map(cr => cr.id),
         checks.map(cr => cr.id)
@@ -408,6 +407,11 @@ export class CommitStatusStore {
       if (cache !== undefined) {
         this.cache.set(key, {
           ...cache,
+          // The commit status store is set to only retreive on a refresh
+          // trigger if the subscription has not been fetched for 60 minutes
+          // (cache/api limit). This sets this sub back to 61 so that on next
+          // refresh triggered, it will be reretreived, as this time, it will be
+          // different given the branch name is provided.
           fetchedAt: moment(new Date()).subtract(61, 'minutes').toDate(),
         })
       }
--- a/app/src/lib/stores/commit-status-store.ts
+++ b/app/src/lib/stores/commit-status-store.ts
@@ -4,12 +4,7 @@ import QuickLRU from 'quick-lru'
 import { Account } from '../../models/account'
 import { AccountsStore } from './accounts-store'
 import { GitHubRepository } from '../../models/github-repository'
-import {
-  API,
-  APICheckStatus,
-  getAccountForEndpoint,
-  IAPICheckSuite,
-} from '../api'
+import { API, getAccountForEndpoint, IAPICheckSuite } from '../api'
 import { IDisposable, Disposable } from 'event-kit'
 import {
   ICombinedRefCheck,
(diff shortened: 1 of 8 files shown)


app/src/lib/stores/commit-status-store.ts
```diff
@@ -331,11 +316,56 @@ export class CommitStatusStore {
...
+  ): Promise<ReadonlyArray<IRefCheck> | null> {
+    const existingChecks = this.cache.get(key)
+
+    // If the checks haven't changed since last status refresh, don't bother
+    // retrieving actions workflows and they could be stale if this is directly after a rerun
+    if (
+      existingChecks !== undefined &&
+      existingChecks.check !== null &&
+      existingChecks.check.checks.find(c => c.actionsWorkflow !== undefined) !==
+        undefined &&
```
sergiou87: I think this should do the same, no?
```suggestion
      existingChecks.check.checks.some(c => c.actionsWorkflow !== undefined) &&
```

app/src/lib/stores/commit-status-store.ts
```diff
@@ -345,18 +375,44 @@ export class CommitStatusStore {
...
+
+      const withBranchName = { ...subscription, branchName }
+      this.subscriptions.set(key, withBranchName)
+      const cache = this.cache.get(key)
+      if (cache !== undefined) {
+        this.cache.set(key, {
+          ...cache,
+          fetchedAt: moment(new Date()).subtract(61, 'minutes').toDate(),
+        })
+      }
```
sergiou87: I think it'd be nice to have a comment explaining this.

sergiou87 (changes requested):

app/src/lib/stores/commit-status-store.ts
tidy-dev: Good call - I always forget about `some`

sergiou87 (approved):
Thanks for those changes! 🐸 🕴️

app/src/lib/stores/commit-status-store.ts
hamzawytsco: كمل

PR #13414, opened 2021-11-30
Check run interaction stats

tidy-dev:
## Description
A few questions that could be answered:

If we find that users that open the check runs popover and also are viewing the checks or job steps online at a high rate, it would indicate that the users do want to see the logs or more detail of some kind. I would think > 25% of popover opens result in viewing online would be highly indicative. Even >5% would be interesting.. I would expect if it is not important, it would be hardly at all < 1%. If so, would result, it revisiting the idea of viewing logs in Desktop.

If there is a low user count (< 1%?) on opening the check runs popover, it could indicate that the feature is not desired or that it is not easily discoverable.  If so, may consider a "Did you know?"  one time popup like we did for cherry-picking.

If there is a low user count (< 5%) of users who open the check runs popover and also rerun jobs, that would indicate to me that being able to rerun jobs is not a desirable feature from Desktop. -> Maybe not worth investing time in rerunning of individual actions jobs or only failed checks. (Currently is not available via api and/or may entail rerunning an entire check suite)

Stats added:
- opensCheckRunsPopover
- viewsCheckOnline (This applies to action and non action checks)
- viewsCheckJobStepOnline (If user is viewing this frequently, it would indicate they would like to see actions logs).
- rerunsChecks 

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/toolbar/branch-dropdown.tsx
+++ b/app/src/ui/toolbar/branch-dropdown.tsx
@@ -276,8 +276,6 @@ export class BranchDropdown extends React.Component<
       return null
     }
 
-    this.props.dispatcher.recordCheckRunsPopoverOpened()
-
     return (
       <CICheckRunPopover
         prNumber={pr.pullRequestNumber}
--- a/app/src/ui/toolbar/branch-dropdown.tsx
+++ b/app/src/ui/toolbar/branch-dropdown.tsx
@@ -276,6 +276,8 @@ export class BranchDropdown extends React.Component<
       return null
     }
 
+    this.props.dispatcher.recordCheckRunsPopoverOpened()
+
     return (
       <CICheckRunPopover
         prNumber={pr.pullRequestNumber}
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -451,6 +451,18 @@ export interface IDailyMeasures {
 
   /** The number of times the user reset to a previous commit. */
   readonly resetToCommitCount: number
+
+  /** The number of times the user opens the check run popover. */
+  readonly opensCheckRunsPopover: number
+
+  /** The number of times the user clicks link to view a check online */
+  readonly viewsCheckOnline: number
+
+  /** The number of times the user clicks link to view a check job step online */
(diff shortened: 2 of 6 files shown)


sergiou87 (dismissed):
LGTM! I think this popover will be hard to discover, but I'm hesitant about adding yet another popup for it 🤔 

Mainly because new users will have 9234798234 popups to close before they start using the app 😂

app/src/ui/toolbar/branch-dropdown.tsx
```diff
@@ -276,6 +276,8 @@ export class BranchDropdown extends React.Component<
       return null
     }
 
+    this.props.dispatcher.recordCheckRunsPopoverOpened()
```
sergiou87: This is a render method, so might be invoked many times after the popover is opened. Maybe this should be in `AppStore::_setShowCIStatusPopover` method when `showCIStatusPopover` is `true`?

sergiou87 (changes requested):
Sorry about previous approval 😆

app/src/ui/toolbar/branch-dropdown.tsx
tidy-dev: Good point!

tidy-dev:
>Mainly because new users will have 9234798234 popups to close before they start using the app

Exactly why I was not initially recommending it.. 😛 

It had got me thinking about our approach with this -  tho I like the niceness of pointing to something on the screen like a tour mode... it does get busy... especially for first time users.. ugg.. I think an optional first time user guide would be nice and then a more informative consistent way of on update - here is a flashy new feature notification.  

On my budgeting software, after an update, if there is a significant feature change (not just bug fixes), they have a single popup which quickly describes the new feature(s).. It popups up in the bottom right hand corner like a toastr message and goes a way after a minute if I don't interact with it. One time, they even had a button in it "would you like a tour of the new features" - big change on that one. We sort of do this with the hint to look at release notes in our banner... but that requires the user to interact to see what the new features are.

Rambling thoughts. :D

sergiou87 (approved):
LGTM now, thanks for that last change!

PR #13423, opened 2021-12-01
Check run tidying

tidy-dev:
## Description

This incorporates the remaining feedback from @adrianmg. 

- Creates a consistent completeness indicator and header to dotcom's pr merged status list.
- Removes borders from job list items to help with distinguishing hierarchy.

 I did not incorporate the sticky "View on GitHub" footer (that opens the pr) as I am not sure of its value. This feels a little misleading as it would not be taking you to view the check runs on GitHub. Also, I am not exactly sure what the use case for it is. That said, open to discussing this if someone thinks it is important. :) 

### Screenshots

https://user-images.githubusercontent.com/75402236/144276771-c15cec4b-4a42-4cdc-9a78-efa0eaedc368.mov

Showing all failed:
Screen Shot 2021-12-01 at 11 34 59 AM

Showing what canceled ones look like:
Screen Shot 2021-12-01 at 11 40 42 AM

## Release notes

Notes: [Improved] Show visual completeness indicator for check runs.


app/src/ui/octicons/donut.tsx
```diff
@@ -0,0 +1,122 @@
+import * as React from 'react'
```
sergiou87: Only a nit about this file: should it live in `app/src/ui` instead of `app/src/ui/octicons`?

sergiou87 (changes requested):
I haven't tested it yet but other than the nit mentioned with the location of `donut.tsx`, the only thing that calls my attention from the video is that the loading image is not properly aligned anymore:
image

app/src/ui/octicons/donut.tsx
tidy-dev: :D I kept going back and forth about that.... because I was thinking if I saw that on the screen, I would think of it like our other icons.. and go looking around the octicons.. but.. also agree it isn't one. :P Since, it felt weird there to you I will gladly move it. :)

tidy-dev:
Fixed (loader forced on in this image so I could get screen shot :P)
image

tidy-dev:
Other thing about loading state.  I have the spinner icon in the donut place while loading.. which since it is so big, feels very prominent to me. Agree/suggestions?

tidy-dev:
Other idea for loader - using same spin class for animation, but not as contrasting with background.

Donut loader:

https://user-images.githubusercontent.com/75402236/144431822-a941405e-99f0-41ca-b1ab-f307825a2a5f.mov

Current loader (already used in other places in app - just not so big):

https://user-images.githubusercontent.com/75402236/144431867-48e2083c-d33d-4654-8632-5e8970da8852.mov

tidy-dev:
I realized that the information provided in the header will not change on retrieval of action workflow (what the loader is for). I updated to add that if loading that info, go ahead and use the cached status for the header. This should pretty much eliminate seeing a loader flicker in the header. Thus, I think the spinner loader is ok as there "should" always be a cache value if you are opening the popover.

sergiou87 (approved):
LGTM! My only doubt is which of these two looks better, because seeing the re-run button not vertically centered is a bit odd.

In any case, it can be tackled at any other moment if needed.

2021-12-03 at 16 41

vs

2021-12-03 at 16 41

tidy-dev:
Both feel natural to me. Going to go ahead and merge as is. But, more than willing to switch it.

PR #13426, opened 2021-12-01
Upgrade to Electron v13

### LINKED ISSUE #12978: Upgrade to Electron v13
In an attempt to upgrade GitHub Desktop to Electron v12 and then v13, we ran into a performance issue affecting unsigned builds on Big Sur (and potentially Monterey too). More specifically, using `spawn` becomes very slow.

The problem seems to be in [libuv](https://github.com/libuv/libuv/issues/3050), and [a first attempt to fix it](https://github.com/libuv/libuv/pull/3064) made some Node.js tests fail, so the change [was reverted](https://github.com/libuv/libuv/pull/3107). This affects other apps like VS Code, and they decided to apply the changes to Electron, ignoring those broken Node.js tests, but [they found other problems derived from this](https://github.com/electron/electron/issues/26143#issuecomment-879569238) that they had to deal with.

Right now we don't need anything from Electron v12 or v13, so we're gonna wait until there is more progress around this. Our hopes are in a [new PR trying to fix this](https://github.com/libuv/libuv/pull/3257), but the new fix led to a kernel bug that has been reported to Apple.

Related issues:
- Issue in libuv: https://github.com/libuv/libuv/issues/3050
- First report in the Electron repo: https://github.com/electron/electron/issues/26143
- Same issue in VS Code: https://github.com/microsoft/vscode/issues/132654
- VS Code uses a fork with the first fix but it causes its own issues: https://github.com/electron/electron/issues/26143#issuecomment-879569238
- First attempt o fix the issue: https://github.com/libuv/libuv/pull/3064
- Second attempt to fix the issue: https://github.com/libuv/libuv/pull/3257

sergiou87:
Closes #12978 

## Description

This PR bumps Electron to v13. It has some changes needed to adapt to the new API.

## Release notes

Notes: no-notes


app/src/lib/git/spawn.ts
```diff
@@ -101,7 +101,7 @@ export async function spawnAndComplete(
             // the process terminates
             const exitCodes = successExitCodes || new Set([0])
 
-            if (exitCodes.has(code) || signal) {
+            if ((code !== null && exitCodes.has(code)) || signal) {
```
sergiou87, line 104: More context here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50262

tidy-dev (dismissed):
✨  Looks good - runs good on my machine 😸

app/src/lib/stores/app-store.ts
```diff
@@ -5266,9 +5266,11 @@ export class AppStore extends TypedBaseStore<IAppState> {
   ): Promise<void> {
     try {
       if (moveToTrash) {
-        const deleted = shell.moveItemToTrash(repository.path)
+        try {
+          await shell.moveItemToTrash(repository.path)
+        } catch (error) {
+          log.error(error)
```
niik: Could we add a little message here to differentiate errors in the log?

```suggestion
          log.error('Failed moving repository to trash`, error)
```

niik (dismissed):
Looks great, just one small nit

sergiou87:
@niik @tidy-dev I had to upgrade `spectron` to support Electron 13 and fix the integration test, ready for review/re-approve again!

niik (approved):
Sorry about the quotation mismatch 😄  We need to get compilation errors into the suggested code editor!

PR #13432, opened 2021-12-02
Cherry-pick regression: On stash, we clear multiCommitOperation, we need to reset on retry

### LINKED ISSUE #13419: cherry picking freezes if uncommitted changes are stashed
### Describe the bug
I had uncommited changes on a branch and I wanted to cherry pick some commit by dragging them to another branch. Desktop proposed to stash the changes (which I did) but then freezed. I could close the popup window asking for stashing. The stash was made, then I could effectively cherry pick what I wanted.

### Version & OS

Version 2.9.4 (x64) on OSX 10.14.6 (18G9323)

steps given in the description.

### Expected behavior

Should have stashed then proceeded to the cherry-picking operation.

### Actual behavior

Froze after stashing.

tidy-dev:
Closes #13419

## Description

Repro steps:
1. Have uncommitted changes
2. Attempt to cherry-pick a commit. Note there are three entry points. 1) drag and drop to an existing branch 2). drag to new branch and 3) context menu. 
3. On popup, hit stash and continue 

As reported in the bug, the app would freeze during cherry-pick stashing, because on attempt to run retry method the multi commit operation would be null. This is because the Local Changes Overridden dialog is not part of the multi commit operation flow and yet dismissing it would technically end the flow. Thus, we end the flow on the dialog opening so that it is not in a state of cherry-pick if the user dismisses. Unfortunately, during refactor the initialization of the mutli comitt operation was moved farther back then the retry method. Thus, this PR adds an if clause in the retry methods (there are 2 since there are 3 entry points to cherry-pick) to re initialize if state does not exist.

## Release notes
Notes: [Fixed] App no longer freezes on stash dialog if cherry-picking with uncommitted changes present.

### DIFF
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -2695,7 +2695,7 @@ export class Dispatcher {
   }
 
   /** Initializes multi commit operation state for cherry pick if it is null */
-  public resetMultiCommitOperationStateCherryPick(
+  public initializeMultiCommitOperationStateCherryPick(
     repository: Repository,
     targetBranch: Branch,
     commits: ReadonlyArray<CommitOneLine>,
@@ -2731,7 +2731,7 @@ export class Dispatcher {
   ): Promise<void> {
     // If uncommitted changes are stashed, we had to clear the multi commit
     // operation in case user hit cancel. (This method only sets it, if it null)
-    this.resetMultiCommitOperationStateCherryPick(
+    this.initializeMultiCommitOperationStateCherryPick(
       repository,
       targetBranch,
       commits,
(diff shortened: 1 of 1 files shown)


app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -2694,13 +2694,50 @@ export class Dispatcher {
     log.info(`[cherryPick] - git reset ${beforeSha} --hard`)
   }
 
+  /** Initializes multi commit operation state for cherry pick if it is null */
+  public resetMultiCommitOperationStateCherryPick(
```
niik: Calling this method `reset` makes me think it also deals with clearing out any in-progress state but reading this method it will be a no-op if there's already a multi commit operation in progress. Could/should we rename it to just call it `initialize`?

Looking at `initializeMultiCommitOperationIfConflictsFound` it seems it matches this pattern pretty closely (i.e. no-op if operation in progress).

```suggestion
  public initializeMultiCommitOperationStateCherryPick(
```

niik (changes requested):
Just a small question but I'm not familiar with this code yet so take it for what it's worth

niik:
Sorry, I didn't set you up for success with that suggested change, we need to update the call sites as well

tidy-dev:
> Sorry, I didn't set you up for success with that suggested change, we need to update the call sites as well

Yeah. I knew that was going to happen - was already heading to my vs code to address it. :D

niik (approved):

PR #13436, opened 2021-12-03
Move items to trash from main process

### LINKED ISSUE #13433: Unable to Discard Files
### Describe the bug
Write a clear and concise description of what the bug is.

### Version & OS

Version 2.9.6-beta1 (x64)
Windows 10

1. Make a change
2. Click right click discard changes on file
3. See application hang

### Expected behavior

File is discarded

### Actual behavior

App hangs, you are unable to cancel discard changes as well.

### Screenshots

image

### Logs

### Additional context

This is the first version with new Electron and there are Electron changes around file discards

sergiou87:
Closes #13433

## Description

With the upgrade to Electron 13 in #13426 we introduced a regression in the app when running on Windows where `electron.shell.trashItem` would fail with `Failed to create FileOperation instance`. According to https://github.com/electron/electron/issues/29598 the solution (at least for now) is to invoke that `trashItem` method from the main process.

On macOS, this doesn't seem to be necessary. However, since this approach works too, I preferred to keep the same implementation for both OS.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/app-shell.ts
+++ b/app/src/lib/app-shell.ts
@@ -35,6 +35,9 @@ export interface IAppShell {
 }
 
 export const shell: IAppShell = {
+  // Since Electron 13, shell.trashItem doesn't work from the renderer process
+  // on Windows. Therefore, we must invoke it from the main process. See
+  // https://github.com/electron/electron/issues/29598
   moveItemToTrash: path => {
     return new Promise<void>((resolve, reject) => {
       ipcRenderer.once(
--- a/app/src/lib/app-shell.ts
+++ b/app/src/lib/app-shell.ts
@@ -39,36 +39,11 @@ export const shell: IAppShell = {
   // on Windows. Therefore, we must invoke it from the main process. See
   // https://github.com/electron/electron/issues/29598
   moveItemToTrash: path => {
-    return new Promise<void>((resolve, reject) => {
-      ipcRenderer.once(
-        'move-to-trash-result',
-        (
-          event: Electron.IpcRendererEvent,
-          { error }: { error: Error | null }
-        ) => {
-          if (error === null) {
-            resolve()
-          } else {
-            reject(error)
-          }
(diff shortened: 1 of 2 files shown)


app/src/lib/app-shell.ts
```diff
@@ -35,7 +35,28 @@ export interface IAppShell {
...
+        ) => {
+          if (error === null) {
+            resolve()
+          } else {
+            reject(error)
+          }
+        }
+      )
+
+      ipcRenderer.send('move-to-trash', { path })
```
niik: Could we use [ipcRenderer.invoke](https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrendererinvokechannel-args) instead of send here to avoid setting up our own single use listener? Seems like the this approach could suffer problems should we happen to call it before the previous invocation completed.

niik (changes requested):

app/src/lib/app-shell.ts
sergiou87: Absolutely! FWIW I used `.send` because (1) I had no idea about that and (2) the rest of the methods below in this very same file also use `.send`. Should we change all of those too? (If not now, at some point)

niik: > the rest of the methods below in this very same file also use .send. Should we change all of those too? (If not now, at some point)

Send is fine when you don't need a result and the operation can't fail (or we don't care that it fails) but yeah, we should definitely change that `openExternal` to use `invoke` instead. I imagine `invoke` didn't exist at the time that was written or we just didn't catch it.

sergiou87: @niik fixed in 6ea87e0bb09ab2ec8c73f78b468d24427ef9fc1d

Took me a while to rebuild and retest everything, sorry! But seems to work now on macOS and Windows 😌

niik (approved):

PR #13439, opened 2021-12-03
Dynamic constraints for resizable components

### LINKED ISSUE #2745: Column widths are limited — would love to remove those widths

niik:
Closes #2745 

## Description

Since the very first versions of GitHub Desktop we've enforced a max width for our resizable components (the changes/history list, and the file list in the stashed changes view and in the selected commit view).

These constraints were put in place to ensure that you couldn't resize these components to a point where you could no longer use parts of the UI because they were now being rendered outside of the window bounds. 

The constraints we set where based on running the app in the minimum window size such that it wouldn't be possible to maximize the app, resize the panes, and then restore the window size and find yourself in a situation where parts of the UI were being rendered outside of the window bounds.

This was always intended to be a temporary solution until we could get around to implementing a dynamic max-width that got recalculated based on the layout of the app and the window size (see https://github.com/desktop/desktop/issues/1588#issuecomment-329405096). There's been several attempts by community contributors to address this issue that we've unfortunately had to reject since they focused on removing/increasing the constraints, not making the constraints themselves adapt to the window dimensions.

In this PR we're making the resizable panes remember their "desired" value and dynamically calculate their max width depending on the width of other resizable panes that affect the available space and the windows dimensions while making sure the the UI is usable.

By keeping track of the desired width we can support resizing the application in a consistent way where maximizing the app, resizing the panes then restoring its size will cause the panes to shrink to fit their constraints but (unless modified again) they will inflate to their desired size again once the app is maximized again.

### Screenshots

image
image
image

## Release notes

Notes:

image

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1873,21 +1873,50 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.accountsStore.refresh()
   }
 
+  /**
+   * Calculate the constraints of our resizable panes whenever the window
+   * dimensions change.
+   */
   private updateResizableConstraints() {
+    // Start with all the available width
     let available = window.innerWidth
-    this.sidebarWidth = { ...this.sidebarWidth, min: 200, max: available - 460 }
+
+    // The combined width of the branch dropdown and the push pull fetch button
+    const toolbarButtonsWidth = 460
+
+    // Working our way from left to right (i.e. giving priority to the leftmost
+    // pane when we need to constrain the width)
+    this.sidebarWidth = {
+      ...this.sidebarWidth,
+      // 220 is the smallest width that will still fit the placeholder text in
+      // the branch selector textbox of the history tab
+      min: 220,
+      // Since the repository list toolbar button width is tied to the width of
+      // the sidebar we can't let it push the branch, and push/pull/fetch
+      // buttons off screen.
+      max: available - toolbarButtonsWidth,
+    }
+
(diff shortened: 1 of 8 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -6826,3 +6886,11 @@ function isLocalChangesOverwrittenError(error: Error): boolean {
     error.result.gitError === DugiteError.LocalChangesOverwritten
   )
 }
+
+function constrain(
+  value: IConstrainedValue | number,
+  min = -Infinity,
```
tidy-dev, line 6892: TIL: You can specify infinity as a number 🤯

tidy-dev (approved):
✨  Looks great and tested it out and feels oddly freeing 😸

app/src/lib/stores/app-store.ts
niik: It's the negative infinity that cracks me up! :smile:

Issue #11517, opened 2021-02-02
What could a better pull request experience in Desktop be?


billygriffin:
As pull requests have become a larger part of the regular workflow for more and more projects on GitHub, I think there's a lot of opportunity for Desktop to incorporate them in a more first-class way. This is a first stab at laying out some of the problems and opportunities I see, but I'm sure this isn't exhaustive, so feedback welcome.

### Problem 1: I can't see the delta between my PR branch and the branch that it's proposed to be merged into.

The idea here is for this to serve as the equivalent of the "Files changed" tab on dotcom, where you can see the entirety of the diff between the two branches. Currently, Desktop only gives you the diff by commit, which isn't really what you want in the case of considering whether to merge one branch into another.

<img width="1268" alt="Screen Shot 2021-02-02 at 12 38 12 PM" src="https://user-images.githubusercontent.com/5091167/106653034-869c1900-6553-11eb-89b7-d19a4a243cc9.png">

This is somewhat captured in https://github.com/desktop/desktop/issues/6757, though that's a request for a broader diff not just excluded to PRs. @rafeca and I also did a hack day exploration of this, which resulted in this branch: https://github.com/desktop/desktop/compare/hack/diff-prs It's half-baked for sure but provides a starting point for when we start working on this.

### Problem 2: If I check out a PR to build it and review it locally, I have to then go back to GitHub.com to actually add my review

There's nowhere in Desktop today to just approve a PR or request changes, despite being able to check out a PR very easily. This could optionally involve some incorporation of actual code review comments and/or suggested changes, but I'm more ambivalent about the question of whether that belongs in Desktop or not.

### Problem 3: It's challenging to actually get to the PR on dotcom from Desktop

There's a keyboard shortcut to view the PR on GitHub, but that's not very discoverable, and it's strange that there's not a natural place in the app to get to the PR page as it's something I do regularly (though the need for this might be partially mitigated if we solve some of the other problems, because there's less need to go back and forth if you can do more with PRs in Desktop).

This is captured in https://github.com/desktop/desktop/issues/8152

### Problem 4: I can't see any additional context about checks other than whether the entire suite passed, or if something failed (for example, which checks failed)

It's nice to see the top level green check or red x for everything passing or something failed, but it's not sufficient if I want to know anymore detail about what failed, or potentially re-run a failing check.

Here's a design of a possible solution for showing more granularity in checks:
[...]

PR #13441, opened 2021-12-03
Pull Request Quick View Component - Getting started

tidy-dev:
Part of #11517

This is feature flagged to development.

## Description

As part of the work to bring pull requests into the app, we need a way to view pull requests details. 

This PR is to create the first stab/draft of the component for the pull request quick view component modeled after these designs:
image

Work to be built on top of this PR:
- Dynamic status (draft/open) #13480
- View on github button works #13478 
- Base href isn't hard coded. #13481
- Currently when switching between different PR's, the height of the containing div of the markdown parser is not dynamically shrinking and growing as expected and sometimes gives a console error (seems there is a timing issue) #13482
- Responsive positioning/sizing/look feel
  - The popover should move to align with the hovered over pr list item (This may mean align top for upper list items and align bottom for lower list items) #13491
  - The popover width could be dynamic for larger screens.
  - We could have an arrow pointing from the modal to the list item to make it obvious which one is being displayed (This especially needed when we have a pr checked out and therefore one of the list items is in a selected state.. which is confusing when it is not the one you are looking at in the card) #13498 
- Actions Footer; Dynamic Actions needed header
- Markdown Parser GitHub Custom filters (Things like commit links and emojis)
- More meta data - such labels/assignee

Bugs found during dev:
- Sometimes branch/pr tabs get squished on screen resizing since adding pr quick view card.

### Screenshots

https://user-images.githubusercontent.com/75402236/144881726-82cb324b-ca2e-49c5-b47b-623469c51057.mov

## Release notes
Notes: [Added] User can hover over a pull request to see a quick view of that pull request.

### DIFF
--- a/app/src/lib/databases/pull-request-database.ts
+++ b/app/src/lib/databases/pull-request-database.ts
@@ -23,6 +23,9 @@ export interface IPullRequest {
   /** The title. */
   readonly title: string
 
+  /** The body of the PR - This is markdown. */
+  readonly body: string
+
   /** The string formatted date on which the PR was created. */
   readonly createdAt: string
 
@@ -121,6 +124,15 @@ export class PullRequestDatabase extends BaseDatabase {
        * and unless the draft PRs get updated at some point in the future we'll
        * never pick up on it so we'll clear the db to seed it with fresh data
        * from the API.
+       */
+      tx.table('pullRequests').clear()
+      tx.table('pullRequestsLastUpdated').clear()
+    })
+
+    this.conditionalVersion(9, {}, async tx => {
+      /**
+       * We're introducing the `body` property on PRs in version 8 in order
+       * to be able to display the body of the pr.
        */
       tx.table('pullRequests').clear()
       tx.table('pullRequestsLastUpdated').clear()
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -121,11 +121,9 @@ export class BranchesContainer extends React.Component<
   }
 
   private onMouseLeavePullRequestQuickView = () => {
-    /*
     this.setState({
       pullRequestBeingViewed: null,
     })
-    */
   }
 
   private renderMergeButtonRow() {
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -80,9 +80,7 @@ export class BranchesContainer extends React.Component<
     return null
   }
 
-  private pullRequestQuickViewRef: React.RefObject<
-    HTMLDivElement
-  > = React.createRef()
+  private pullRequestQuickViewTimerId: number | null = null
 
   public constructor(props: IBranchesContainerProps) {
     super(props)
@@ -94,6 +92,10 @@ export class BranchesContainer extends React.Component<
       branchFilterText: '',
       pullRequestBeingViewed: null,
     }
+  }
+
+  public componentWillUnmount = () => {
+    this.clearPullRequestQuickViewTimer()
   }
 
   public render() {
@@ -117,18 +119,23 @@ export class BranchesContainer extends React.Component<
 
     return (
       <PullRequestQuickView
-        ref={this.pullRequestQuickViewRef}
         dispatcher={this.props.dispatcher}
         pullRequest={this.state.pullRequestBeingViewed}
+        onMouseEnter={this.onMouseEnterPullRequestQuickView}
         onMouseLeave={this.onMouseLeavePullRequestQuickView}
       />
     )
+  }
+
+  private onMouseEnterPullRequestQuickView = () => {
+    this.clearPullRequestQuickViewTimer()
   }
 
   private onMouseLeavePullRequestQuickView = () => {
     this.setState({
       pullRequestBeingViewed: null,
     })
+    this.clearPullRequestQuickViewTimer()
   }
 
   private renderMergeButtonRow() {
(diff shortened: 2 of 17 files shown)


app/src/lib/databases/pull-request-database.ts
```diff
@@ -23,6 +23,9 @@ export interface IPullRequest {
   /** The title. */
   readonly title: string
 
+  /** The body of the PR - This is markdown. */
```
sergiou87, line 26: Sorry, this came to mind while reading your comment, I couldn't help it 😳 🤣 

5xopk7

app/src/ui/branches/branches-container.tsx
```diff
@@ -275,10 +303,37 @@ export class BranchesContainer extends React.Component<
...
+    this.setState({ pullRequestBeingViewed })
+  }
+
+  private onMouseLeavePullRequestListItem = async (
+    event: React.MouseEvent<HTMLDivElement, MouseEvent>
+  ) => {
+    // If we leave a list item, onto the pull request quick view, we don't want
+    // to close the quick view
+    const { relatedTarget } = event
+    const prQuickView = document.getElementById('pull-request-quick-view')
```
niik: Could we solve this by using a Ref instead of relying on a global id?

If we add a prop like this to the QuickView component

```
  readonly rootElementRef?: React.Ref<HTMLDivElement>
```

We can pass that to the `ref` attribute of the root div inside the quickview and use that. In "modern" react you'd use `React.forwardRef` (https://reactjs.org/docs/forwarding-refs.html)

app/src/ui/pull-request-quick-view.tsx
```diff
@@ -0,0 +1,76 @@
...
+  readonly onMouseLeave: () => void
+}
+
+export class PullRequestQuickView extends React.Component<
+  IPullRequestQuickViewProps,
+  {}
+> {
+  private baseHref = 'https://github.com/'
+
+  private renderHeader = (): JSX.Element => {
```
niik: Do we need to specify a return type here or can we rely on inference?

```suggestion
  private renderHeader = () => {
```

app/src/ui/pull-request-quick-view.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <div className="action-needed">Review requested</div>
+        <Button className="button-with-icon">
+          View on GitHub
+          <Octicon symbol={OcticonSymbol.linkExternal} />
+        </Button>
+      </header>
+    )
+  }
+
+  private renderPR = (): JSX.Element => {
```
niik: Do we need to specify a return type here or can we rely on inference?
```suggestion
  private renderPR = () => {
```

niik (changes requested):
Looking great so far @tidy-dev, sorry for taking so long to review

app/src/ui/pull-request-quick-view.tsx
tidy-dev: Curious, I think you are the first person I have met to ask me to remove a return type. :D Do you just find that it clutters things/ do you think it just isn't useful for things like render which are not typically reusable methods?

app/src/ui/branches/branches-container.tsx
tidy-dev: Neat! I will try that out.

app/src/ui/pull-request-quick-view.tsx
niik: Hah, yeah, I can see how it runs a bit counter to my strong preference for static typing in general but we don't gain anything by including it here and explicitly typing the return value as JSX.Element could even cause us to have to change it down the line should we want to return a string for example (or make it optional, i.e. able to return a falsy value).

I think specifying explicit return types is great when we either want to codify an external requirement or when we want to make absolutely certain that we always return an element (though this is a bigger issue for codebases which allow implicit returns).

When I see explicit return types like this it makes me wonder what the special constraint is that caused us to add them and can make me concerned about changing the return type. All that said the primary reason I rarely write them myself is just because I'm lazy and because the compiler knows better what the return type is than I do.

app/src/lib/databases/pull-request-database.ts
Upinstuff: 😂🤣😂🤣👏👏👏👍🥰🌞✊🙏🇺🇸🌎🌎🤪😁

tidy-dev:
Sorry for last minute change (last commit was going to be a separate pr) I explored fixing the scrollbar hover issue and adding delay to hover and it negated the need of where I added `React.forwardRef`. I figured it made more sense to go ahead and add to this pr instead of another pr undoing the last commit in this one.

app/src/ui/pull-request-quick-view.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private onMouseLeave = () => {
+    this.props.onMouseLeave()
+  }
+
+  public render() {
+    return (
+      <div
+        id="pull-request-quick-view"
```
niik: Could we get rid of the id now?
```suggestion
        className="pull-request-quick-view"
```

app/styles/ui/_pull-request-quick-view.scss
```diff
@@ -0,0 +1,53 @@
+#pull-request-quick-view {
```
niik:
```suggestion
.pull-request-quick-view {
```

niik (changes requested):

niik (approved):

PR #13460, opened 2021-12-07
Flag check rerun button to non-GHE repos

tidy-dev:
## Description

Since there were api changes made to the Checks api, we cannot support the rerun function for GHE repos as those new changes are not available on GHE yet.

### Screenshots
https://user-images.githubusercontent.com/75402236/145045335-9b5c53ba-eed9-4c3d-968e-2bd510315b27.mov

## Release notes
Notes: [Fixed] Remove rerun button from GHE repositories.

### DIFF
--- a/app/src/models/github-repository.ts
+++ b/app/src/models/github-repository.ts
@@ -58,6 +58,11 @@ export class GitHubRepository {
   public get fork(): boolean {
     return !!this.parent
   }
+
+  /** Is the repository a GHE repository */
+  public get isGHE(): boolean {
+    return this.owner.endpoint.startsWith('https://ghe.io/')
+  }
 }
 
 /**
--- a/app/src/models/github-repository.ts
+++ b/app/src/models/github-repository.ts
@@ -58,11 +58,6 @@ export class GitHubRepository {
   public get fork(): boolean {
     return !!this.parent
   }
-
-  /** Is the repository a GHE repository */
-  public get isGHE(): boolean {
-    return this.owner.endpoint.startsWith('https://ghe.io/')
-  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


app/src/models/github-repository.ts
```diff
@@ -58,6 +58,11 @@ export class GitHubRepository {
   public get fork(): boolean {
     return !!this.parent
   }
+
+  /** Is the repository a GHE repository */
+  public get isGHE(): boolean {
```
niik: So `ghe.io` is our own internal GitHub Enterprise Server (GHES) installation. For a customer the endpoint would typically be somethine like `https://github.company.com/api/v3`. So what we typically do everywhere we need to differentiate GitHub.com from non GitHub.com endpoints (which up until the advent of GHAE was only GHES) is to check if the endpoint matches .com's API endpoint and if it doesn't we'll know that it's either GHAE or GHES.

app/src/ui/check-runs/ci-check-run-popover.tsx
```diff
@@ -218,6 +218,10 @@ export class CICheckRunPopover extends React.PureComponent<
 
   private renderRerunButton = () => {
     const { checkRuns } = this.state
+    if (this.props.repository.isGHE) {
```
niik: While it'd be nice to have a more reusable approach here I'm proposing we stick with our typical way of selectively targeting GHES here and potentially consider a more generalized approach separately.

```suggestion
    // GHES doesn't support rerunning checks yet
    if (this.props.repository.endpoint !== getDotComAPIEndpoint()) {
      return null
    }
```

niik (changes requested):

niik (approved):

ghost:
thx

PR #13521, opened 2021-12-13
Take alias into account when sorting repositories

### LINKED ISSUE #13429: Sort repositories by alias instead of folder name when using alias
### Describe the bug
I set an alias on every repository and prepend them with the project name because otherwise many of the repositories would have the same names and it's hard to differentiate which project they belong to. However their sort order seems to be based on the folder names even when using an alias. It would make sense to order them by their displayed name i.e. alias if used or folder name otherwise.

### Version & OS
Version 2.9.4 (x64)
MacOS 11.6

Add a couple of repositories to the list and set aliases. The sort order doesn't change to be based on the displayed name, but instead still be based on their folder name.

### Expected behavior

Repositories should be ordered by their displayed name, not by folder name when using alias.

### Actual behavior

Repositories seem to be ordered by their folder name; aliases don't override it.

### Screenshots

All repositores here have aliases.
image

niik:
Closes #13429

## Description

Take repository aliases into account when sorting the repository list.

## Release notes

Notes: [Improved] Repository aliases are taken into account when sorting the repository list

### DIFF
--- a/app/src/ui/repositories-list/group-repositories.ts
+++ b/app/src/ui/repositories-list/group-repositories.ts
@@ -83,7 +83,9 @@ export function groupRepositories(
       names.set(repository.name, existingCount + 1)
     }
 
-    repositories.sort((x, y) => caseInsensitiveCompare(x.name, y.name))
+    repositories.sort((x, y) =>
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
:shipit:

PR #13528, opened 2021-12-13
Update ChooseLicense submodule to the latest version

agisilaos:
## Description

With this PR I updated ChooseLicense submodule to the latest version

## Release notes

Notes: no-notes

PR #13527, opened 2021-12-13
Update gitignore submodule to the latest version

agisilaos:
Closes #13410

## Description

Updated gitignore submodule to the latest version

## Release notes

Not sure what we should put here since with this update we're fixing an open issue.

PR #13529, opened 2021-12-13
Update gemoji submodule to the latest version

agisilaos:
## Description

With this PR I updated gemoji submodule to the latest version.

Notes: no-notes


niik:
@agisilaos Is there an issue or documentation of a problem with the existing version of our gemoji submodule? Taking a casual look at the diff on the gemoji side it seems it's no longer shipping the emoji images nor the [custom emoji](https://github.com/github/gemoji/pull/158) that we rely on in the app (like :shipit:). I haven't tested this locally but I would be surprised if this change _didn't_ break the emoji support in the app?

agisilaos:
@niik Not sure. Tbh, I opened this PR because as I checked #13527 and except gitignore, #13528 (Choose License) needed an update as well as gemoji.

niik:
> @niik Not sure. Tbh, I opened this PR because as I checked #13527 and except gitignore, #13528 (Choose License) needed an update as well as gemoji.

Given that CI is failing and given the upstream changes of removing custom emoji and images I think we can be fairly certain this change will break our emoji support in-app. As such I'll close this out but let us know if this is something you'd like to investigate and improve upon further.

PR #13201, opened 2021-10-28
IGNORE: Fail CI tasks immediately

sergiou87:
This is a test PR, :+1: :+1:  don't look! 🙈 🙈  :+1: :rocket: :zzz:

| Syntax      | Description |
| ----------- | ----------- |
| Header       | Title             |
| Paragraph  | Text  :+1:           |

 :astonished:

```
This will be code tag in a pre tag. -> What is here :astonished:
```
:no_mobile_phones:
<pre> just a pre tag :no_mobile_phones: tag </pre>

:computer:

` In some code, :computer:`

PR #13548, opened 2021-12-16
Custom Markdown Emoji filter POC

tidy-dev:
Part of #11517

<a href="https://github.com/desktop/desktop/pull/13553123">hmmm</a>.

Note: for testing, run `yarn build:dev` otherwise new markdown emoji css won't be applied and emoji's will be big

## Description

One of our current sandboxed markdown limitations is that even tho we use the MarkedJS library that does have a GitHub Flavored Markdown (gfm) flag; dotcom still has several custom markdown filters it applies on top of more generic markdown. Some of them that are pretty noticeable are emoji's, issue references, username references, and commit references.

This PR is a POC showing we can emulate the approach taken by dotcom to apply these custom filters. That is, after we parse the markdown via the MarkedJS library and apply the DomPurify sanitizer to obtain a markdown html string, we create DOM Document inside our renderer and filter the contents based on its nodes. Then, we will convert it back to a html string to inject into the sandboxed iframe.

Benefits of this approach:
- We should be able to closely mimic dotcom's logic for our markdown parser to reduce likely hood of different behavior for our parser versus dotcom.
- Using Node traversal gives us a quick way to narrow nodes down to the nodes applicable to  currently being applied custom filters.

Possible Cons:
-  We are instantiating  user provided parsed html in Desktop's DOM renderer. This should have the same security implications as using the marked js / sanitizer at this level (which as passed a [security review](https://github.com/github/security-reviews/issues/464))
- Not sure if it is faster than trying to make custom filters on the markdown html string.

### Screenshots

image

#13201 (<- If you want to see how dotcom renders it)

## Release notes
Notes: [Improved] Sandbox markdown parser parses emojis


app/src/lib/markdown-filters/node-filter.ts
```diff
@@ -0,0 +1,97 @@
...
+  const replacementMap = new Map<Node, ReadonlyArray<Node>>()
+  while (textNode !== null) {
+    const replacementNodes = await nodeFilter.filter(textNode)
+    if (replacementNodes !== null) {
+      replacementMap.set(textNode, replacementNodes)
+    }
+    textNode = walker.nextNode()
+  }
+
+  for (const [toReplace, replacements] of replacementMap.entries()) {
```
tidy-dev: May wonder why not do this replacement in the above loop. I tried that at first, but if I replace the current node, the walker fails to get the next node. Maybe there is a better way of doing this still yet - open to thoughts.... maybe it would work if I keep a `lastNode` ref and on each iteration replace the prev node stuf.. but I am not sure that won't break the walker either.

tidy-dev:
Paired with Markus on this today:
- [x] Found a better way to replace the nodes (thing I commented on)
- [x] Going to implement a better way to get the base64 of the emoji images use FSE reader
- [x] Going to cache base 64 uri's of emojis so we don't have to reconvert same emoji's for repeat emoji's.
- [x] Will look into possibility of reusing the "maybe is emoji" scanning logic from `text-token-parser.ts` instead of really large build regex. See #13556

app/src/lib/markdown-filters/node-filter.ts
tidy-dev: Paired with Markus - found [a better way](https://github.com/desktop/desktop/pull/13548/commits/10dd400dbc9412278bc06cd9bb86477d182d6880). :)

app/static/common/markdown.css
```diff
@@ -295,7 +295,9 @@ changes from the original, just a note that this will not match 1-1)
 .markdown-body .emoji {
 	max-width: none;
 	vertical-align: text-top;
-	background-color: transparent
+	background-color: transparent;
+	height: var(--font-size);
+	width: var(--font-size);
```
sergiou87: Just checked how dotcom handles this, and saw a bunch of properties. Is any of this useful for us here?

image

app/src/lib/markdown-filters/emoji-filter.ts
```diff
@@ -0,0 +1,138 @@
...
+      if (emojiPath === undefined) {
+        continue
+      }
+
+      const emojiPosition = text.indexOf(emojiMatches[0])
+      const textBeforeEmoji = text.slice(0, emojiPosition)
+      const textNodeBeforeEmoji = document.createTextNode(textBeforeEmoji)
+      nodes.push(textNodeBeforeEmoji)
+
+      const emojiImg = await this.createEmojiNode(emojiPath)
```
niik: In our previous uses of emoji we just give the url to an `<image>` tag and let it deal with any potential errors in loading. Here we run the risk of crashing the app should we fail to read the image from disk. Could we catch here and just ignore the emoji substitution if it fails?

app/src/lib/markdown-filters/emoji-filter.ts
```diff
@@ -0,0 +1,138 @@
...
+      return null
+    }
+
+    let text = node.textContent
+    const emojiMatches = text.match(this.emojiRegex)
+    if (emojiMatches === null) {
+      return null
+    }
+
+    const nodes: Array<Text | HTMLImageElement> = []
```
niik: Our convention is typically to use the `new Array<T>` format vs the `: Array<T> = []` format

```suggestion
    const nodes = new Array<Text | HTMLImageElement>()
```

app/src/lib/markdown-filters/node-filter.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Builds an array of node filters to apply to markdown html. Referring to it as pipe
+ * because they will be applied in the order they are entered in the returned
+ * array. This is important as some filters impact others.
+ *
+ * @param emoji Map from the emoji shortcut (e.g., :+1:) to the image's local path.
+ */
+export function buildCustomMarkDownNodeFilterPipe(
+  emoji: Map<string, string>
+): ReadonlyArray<INodeFilter> {
+  return [new EmojiFilter(emoji)]
```
sergiou87: Can we cache this? Maybe with `memoize` or something (if it works well with a `Map<string, string>`). You can even set an expiration time IIRC.

If I'm not wrong, the emojis shouldn't change while the app is running, so having to create all these all the time feels like a waste of resources if the sandboxed markdown component is recreated very frequently 😳

sergiou87 (changes requested):
Looks good overall! Just one question/suggestion

app/static/common/markdown.css
niik: dotcom use native unicode emoji characters IIRC whereas we still use PNGs so most of the rules here are not applicable. That said I'm not so sure it's a good idea to mix font sizes with width/height like this. I would prefer if we specified the height relative to the font size using a relative unit and that we leave the width to be automatically determined (in case of non-uniform emoji images)

So something like

```css
height: 1em;
width: auto;
```

app/src/lib/markdown-filters/emoji-filter.ts
```diff
@@ -0,0 +1,138 @@
...
+   * Builds a regular expression that is looking for all group of characters
+   * that represents any emoji ref (or map key) in the provided map.
+   *
+   * @param emoji Map from the emoji ref (e.g., :+1:) to the image's local path.
+   */
+  private buildEmojiRegExp(emoji: Map<string, string>): RegExp {
+    const emojiGroups = [...emoji.keys()]
+      .map(emoji => escapeRegExp(emoji))
+      .join('|')
+      .slice(0, -1)
```
niik: This would remove the last `:` from the last emoji and would cause us to match `:zzz` instead of `:zzz:`. I suggest we remove it

```suggestion
```

app/src/lib/markdown-filters/emoji-filter.ts
```diff
@@ -0,0 +1,138 @@
...
+      fatalError(
+        'Emoji filter requires text nodes; otherwise we may inadvertently replace non text elements.'
+      )
+    }
+
+    if (node.textContent === null || !node.textContent.includes(':')) {
+      return null
+    }
+
+    let text = node.textContent
```
niik: Do we really need to throw for this scenario or could we simply return null here?

Perhaps we could combine a few of these assertions into one:

```suggestion
    let text = node.textContent
    if (node.nodeType !== node.TEXT_NODE || text === null) {
      return null
    }
```

app/src/lib/markdown-filters/emoji-filter.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    const nodes: Array<Text | HTMLImageElement> = []
+    for (let i = 0; i < emojiMatches.length; i++) {
+      const emojiKey = emojiMatches[i]
+      const emojiPath = this.emojiFilePath.get(emojiKey)
+      if (emojiPath === undefined) {
+        continue
+      }
+
+      const emojiPosition = text.indexOf(emojiMatches[0])
```
niik:
```suggestion
      const emojiPosition = text.indexOf(emojiMatches[i])
```

Or else it'll get weird when there are multiple emojis in the same text node with differing lengths

niik (changes requested):
This is looking really good @tidy-dev! I 😍 all of this documentation as well, a pleasure reading this code!

app/src/lib/markdown-filters/emoji-filter.ts
tidy-dev: I was thinking along the lines of if a dev in the future used this `filter` method without being in conjunction with it's tree walker creation method, then, the fatalError would explain to them quickly as to why giving this a non text node could be problematic. But, I moved that warning up into the method doc which should be sufficient for a scenario that "should be" unlikely. :D

app/src/lib/markdown-filters/node-filter.ts
tidy-dev: Added a caching of the entire custom markdown node filter pipeline as the filters instantiated in it should not change once instantiated. It is a pretty naive cache without checking to see if parameters changed... but I don't think it can/should during run time - but this could need updated as more filters are added.

For others following along, Sergio and I discussed the emoji storing/caching as is done currently. Right now, that main emoji map is generated on app start and then stored in main app state and passed down everywhere we need it. We could move this into a global constant to use as needed via importing. As that would be a pretty decent refactor (probably warrants more discussion),  I believe it is out of scope for this PR.

app/src/lib/markdown-filters/node-filter.ts
```diff
@@ -33,7 +40,11 @@ export interface INodeFilter {
 export function buildCustomMarkDownNodeFilterPipe(
   emoji: Map<string, string>
 ): ReadonlyArray<INodeFilter> {
-  return [new EmojiFilter(emoji)]
+  if (customMarkDownNodeFilterPipe === undefined) {
+    customMarkDownNodeFilterPipe = [new EmojiFilter(emoji)]
+  }
+
+  return customMarkDownNodeFilterPipe
 }
```
niik: I think we can get rid of the intermediate `customMarkdownFillterPipe` by leveraging `memoizeOne` here.

```ts
export const buildCustomMarkDownNodeFilterPipe = memoizeOne(
  (emoji: Map<string, string>): ReadonlyArray<INodeFilter> => [
    new EmojiFilter(emoji),
  ]
)
```

This will also ensure that we'll build a new pipe for each emoji map (by reference equality) should we happen to use multiple in the future.

niik (changes requested):
Just one little suggestion that offers no real practical difference so if you want feel free to ignore it, just let me know and I'll approve either way

niik (approved):
Nicely done @tidy-dev! :sparkles:

sergiou87 (approved):
🐸 🕴️ ❗

PR #13552, opened 2021-12-17
PR Quick View: position quick view relative to list item shown (take 2)

tidy-dev:
Part of #11517

(take 2 - other branch got accidentally messed up when I tried to update it...)

## Description

This is to position the quick view card relative to the the position of the pr list item. It will align top if there is height enough and otherwise align bottom if there is height enough. Otherwise, it will attempt to center the quick view card on the pr list view item with a min top position of the top of the list and a min bottom position of visible on screen. 

### Screenshots
https://user-images.githubusercontent.com/75402236/145462507-ad67a560-bbb1-4f91-b53c-2673b7dde324.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/branches/branches-container.tsx
+++ b/app/src/ui/branches/branches-container.tsx
@@ -58,7 +58,10 @@ interface IBranchesContainerState {
   readonly selectedPullRequest: PullRequest | null
   readonly selectedBranch: Branch | null
   readonly branchFilterText: string
-  readonly pullRequestBeingViewed: PullRequest | null
+  readonly pullRequestBeingViewed: {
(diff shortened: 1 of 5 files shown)


sergiou87 (approved):
Not super fan of having constants referencing heights that depend on CSS rules but… not sure if there's a better way of handling that. Other than that LGTM and works well!

tidy-dev:
> Not super fan of having constants referencing heights that depend on CSS 

Agree. But, they really just act as "default" positions until the component mounts. Thus, not as concerned as if we actually had to rely on it.

dehghon270 (commented):
Github

PR #13553, opened 2021-12-17
PR QuickView: Add arrow indicating which pr list item is being viewed

tidy-dev:
Part of #11517

## Description

This adds an arrow from the quick view card pointing to the pull request item that it refers to. 

### Screenshots
https://user-images.githubusercontent.com/75402236/145591014-f174a4ea-d8cb-4508-b64b-21e22409effe.mov

## Release notes
Notes: no-notes


app/src/ui/pull-request-quick-view.tsx
```diff
@@ -121,6 +125,40 @@ export class PullRequestQuickView extends React.Component<
     return { top: clamp(alignedMiddle, minTop, maxTop) }
   }
 
+  private getPointerPosition(
+    position: React.CSSProperties | undefined,
+    quickViewHeight: number
+  ): React.CSSProperties | undefined {
+    // 23 = half a pr list item.
```
niik: This comment confuses me, what is it in relation to given that `23` isn't used as a constant in this method?

app/src/ui/pull-request-quick-view.tsx
```diff
@@ -121,6 +125,40 @@ export class PullRequestQuickView extends React.Component<
...
+    }
+
+    const { top, bottom } = position
+    const quickViewTopZero = this.getTopPRList()
+    const prListItemTopWRTQuickViewTopZero =
+      this.props.pullRequestItemTop - quickViewTopZero
+
+    let quickViewTop
+    if (top !== undefined) {
+      quickViewTop = parseInt(top.toString())
```
niik: The fact that we have to coerce this to a string and then parse it suggests to me that we're not leveraging the storage type properly. Looking at `state.position` it's a `CSSProperties` type but it seems we only ever use it to set the `top` property?

Could we change the state type to something like

```ts
interface IPullRequestQuickViewState {
  /** The absolutely positioned top of the quickview element in pixels  */
  readonly top: number | undefined
}
```

app/src/ui/pull-request-quick-view.tsx
```diff
@@ -121,6 +125,40 @@ export class PullRequestQuickView extends React.Component<
...
+    } else if (bottom !== undefined) {
+      const bottomAsNum = parseInt(bottom.toString())
+      quickViewTop = bottomAsNum - quickViewHeight + quickViewTopZero
+    }
```
niik: This seems unused? I.e. there doesn't seem to be a case where bottom can be set?

niik (changes requested):

app/src/ui/pull-request-quick-view.tsx
tidy-dev: I had like 3-4 prs this was built on top of and I had tons of merge conflicts... and I appear to be left with an older version of this method.  Sorry for the confusion/not really ready pr. (I used to use a top/bottom and then found a simpler solution)

app/src/ui/pull-request-quick-view.tsx
tidy-dev: Sorry was left over from a different approach

niik (dismissed):
Looks like you've got a merge conflict there but other than that I think this is good to go!

sergiou87 (approved):

PR #13554, opened 2021-12-17
Endpoint capabilities based on version number

niik:
## Description

I paired with @tidy-dev on being able to automagically "light up" areas of the application for GitHub Enterprise users as their instances of GHES gets upgraded to a version which allows us to perform the API requests we need for a certain feature.

This is what we came up with. We wanted an API which optimized for ease-of-use for developers adding capabilities checks which is why we settled on a declarative approach using semver ranges as opposed to writing if statements yourself. In other words:

👇 
```ts
export const supportsRerunningChecks = endpointSatisfies({
  dotcom: true,
  ae: '>= 3.4.0',
  es: '>= 3.4.0',
})
```
but not 👇 
```ts
export const supportsRerunningChecks = (endpoint: string) => {
  return isDotCom(endpoint) || semver.satisfies(getEndpointVersion(endpoint), ">= 3.3.0")
}
```

Unfortunately (but for good reasons) GitHub AE doesn't publish a version number so we'll just have to treat is as if it's running the most recent stable version of GHES.

Similarly, for GHES there's a risk that corporate proxies and such would remove the version header for obfuscation reasons so if that happens we'll assume their running the oldest still supported version of GHES.

These versions will need to be updated manually every now and then which is a bummer but that's the case with our static checks today as well where we only check if an endpoint is dotcom with the purpose of removing that check once GHES version uptick is high enough. This at least will let us handle capabilities in one place similar to how we deal with feature flags.

## Release notes

Notes: no-notes


app/src/lib/endpoint-capabilities.ts
```diff
@@ -0,0 +1,155 @@
...
+  // Version number constraint
+  assertNonNullable(epVersion, `Need to provide a version to compare against`)
+  return epMatchesType && semver.satisfies(epVersion, epConstraint)
+}
+
+/**
+ * Returns a predicate which verifies whether a given endpoint matches the
+ * provided constraints.
+ *
+ * Note: NOT ment for direct consumption, only exported for testability reasons.
```
sergiou87: Nit:
```suggestion
 * Note: NOT meant for direct consumption, only exported for testability reasons.
```

app/test/unit/endpoint-capabilities-test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    it('assumes GHAE versions', () => {
+      expect(testGHAE('>= 3.3.1')).toBeFalse()
+      expect(testGHAE('>= 3.3.0')).toBeTrue()
+    })
+
+    it('assumes GHES versions', () => {
+      expect(testGHES('>= 3.1.1')).toBeFalse()
+      expect(testGHES('>= 3.1.0')).toBeTrue()
+    })
```
sergiou87: Maybe it's worth adding a comment explaining what do they mean about "assuming the GHAE/GHES versions", where those constants live, and mentioning that these tests will need to be updated when those constants change.

sergiou87 (changes requested):
Looks great! I love that API 😄 

Only found a nit and left some suggestions.

sergiou87 (dismissed):
Thanks for those changes! ❤️  :shipit:

sergiou87 (approved):

PR #13555, opened 2021-12-17
Bump `username` to 5.1.0

sergiou87:
## Description
This fixes the `mem` vulnerability reported in: https://github.com/desktop/desktop/security/dependabot/app/yarn.lock/mem/open

It'd be nice to test the app in general looking for regressions. FWIW, the `username` package was only used in API requests.

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -63,7 +63,7 @@
     "textarea-caret": "^3.0.2",
     "tslib": "^2.0.0",
     "untildify": "^3.0.2",
-    "username": "^6.0.0",
+    "username": "^5.1.0",
     "uuid": "^3.0.1",
     "wicg-focus-ring": "^1.0.1",
     "winston": "^2.3.1",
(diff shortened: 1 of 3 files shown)


sergiou87:
First I tried bumping it to 6.0.0, but I got build issues on CI because it seems to require a higher version of node to build (how could it work on my machine though?!?!?!). Instead of doing that, I decided to bump it to 5.1.0, since the only difference is https://github.com/sindresorhus/username/commit/6c6ea197d0c97f53a90f2a810980095d76642af8 (which actually introduces this problem for us).

Maybe in future Electron upgrades we can afford to safely upgrade node too, and therefore this package. But for now, this should be fine.

(DISCLAIMER: at the time I posted this comment, the CI jobs are still in progress… I assume they will succeed this time, but they might not 😂 )

niik (approved):

PR #13556, opened 2021-12-17
Use maybe emoji regex

tidy-dev:
Warning - Compared to #13548

## Description

TLDR; This change mostly a coding preference and is divergent from dotcom approach. Up for thoughts? 

The Emoji filter in #13548 uses the same approach as dotcom and that is to build a regexp with all the 894 emoji keys. This works and have not noticed a performance issue nor other browser constraint. However, @niik pointed out we already have a emoji filtering approach in Desktop that would likely be more efficient. In this approach, instead of regexp that exactly matches all the emoji possibilities, it does a search for anything that looks like an emoji.  `:somecharacters:` Then, it checks the matches against our emoji map.

https://github.com/desktop/desktop/blob/bc195212a6bfda18a078f80c3408fdc0cde22fe0/app/src/lib/text-token-parser.ts#L116-L126

I modified the original regexp as the above is only scanned upon things that already looks like words -- contiguous chunks of characters. Thus, the start/end anchors don't apply and I needed to use the `\S` to specify non whitespace characters are allow between the colons. Also, used a `+` as one or more instead of `*` as zero or more.

This is not a necessary change and does pose the risk of not being the same as dotcom and therefore could have a different behavior. A benefit is if we would like to update our current emoji filtering logic for stuff like the rich text component to use this emoji filter class to share logic, we would be less likely to regress our current in app emoji behavior.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/markdown-filters/emoji-filter.ts
+++ b/app/src/lib/markdown-filters/emoji-filter.ts
@@ -1,7 +1,6 @@
 import { fatalError } from '../fatal-error'
 import { INodeFilter } from './node-filter'
 import * as FSE from 'fs-extra'
-import { escapeRegExp } from '../helpers/regex'
 import uri2path from 'file-uri-to-path'
 
 /**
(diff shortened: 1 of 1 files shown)


Majkasgurunu:
maybe-emoji-regex

niik:
I did some measuring here and there seems to be no upside performance-wise to using the approach I suggested. I can't help but feel bad about constructing a 10kb+ regex at runtime but 🤷 Unless we find an issue with it let's go with what we've already got. I'll close this out. Thanks for humoring me!

PR #13569, opened 2021-12-21
Update main.ts

ghost:
Just added a dialogbox if one tries to open another instance, then close that another instance


app/src/main-process/main.ts
```diff
@@ -165,6 +165,7 @@ if (!handlingSquirrelEvent) {
   })
 
   if (isDuplicateInstance) {
+    dialog.showMessageBox("Another instance of Github Desktop is already running");
```
ItzLevvie, line 168:
```suggestion
    dialog.showMessageBox("Another instance of GitHub Desktop is already running");
```

PR #13573, opened 2021-12-21
Issue Mention Filter

tidy-dev:
Built on top of https://github.com/desktop/desktop/pull/13548

## Description

Based on dotcom's issue mention filter, it matches for issue references such as #1234, gh-1234, /issues/1234, /pull/1234, or /discussions/1234 and replaces them with anchor tags linking to the issue, pull request, or discussion in the parent repository. (Issues and pull-requests are the same thing) Those issue mentions can also be prefixed with an owner/repo which allows linking to a non-parent repo such as desktop/dugite#1.

Questions:
1 - I couldn't find an GitHub Rest Api for discussion links.. So I used graph ql api.. are there cons to this asides from it feeling much messier?

2. How to address lag?
Current approach appears very slow... I have added a filter cache, but that will only apply to multiple api requests for same issue reference in a single PR (not switching between pr quick views). Additionally, I do check the issue store before making an api call. But.. still noticeable lag.

Thoughts on speeding things up:
1. Since, we are going out to a dispatcher anyways, we could cache these api calls in a store outside of the filter so that at least if a user has opened the pr quick view once, it should be relatively quick the second time.

2. Dotcom has some of it's filters split into two parts (this being one of them), first is a synchronous filter that just goes and looks for issue references and marks them with a unique tag and the second asynchronous filter that matches for those tags and does the check against the api. I am not 100% sure how this works.. But, I think it goes ahead and shows the user the parsed markdown and filters that are asynchronous will update the output as they resolve. This is more feasible since dotcom is not waiting on filters before putting it into a sandboxed iframe and it is manipulating dom nodes.... But.. maybe we could do something similar by keeping a DOM version that as filters resolve we update them and then pass down into the iframe.

3.  We don't hit the api and have more false positives... if something says `#99999999`, we simply make it refer to `current repo owner/current repo/99999999` whether that issue exists or not. Note... dotcom has popovers over issues links that we are not currently doing, which makes this more feasible. But, does mean a different behavior than in dotcom. (This is currently our approach in our ~~autocompletion~~ text-token-parser logic)

### Screenshots
Testing against - https://github.com/tidy-dev/foo-really-not/pull/2

https://user-images.githubusercontent.com/75402236/147777747-d9785ec1-b042-4425-bc00-3d26ebe50cc2.mov

## Release notes
Notes: Adding an issue mentions node filter for markdown parsing.

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -287,6 +287,7 @@ export interface IAPIIssue {
   readonly title: string
   readonly state: 'open' | 'closed'
   readonly updated_at: string
+  readonly html_url: string
 }
 
 /** The combined state of a ref. */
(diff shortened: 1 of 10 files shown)


niik:
> 3\. We don't hit the api and have more false positives... if something says `#99999999`, we simply make it refer to `current repo owner/current repo/99999999` whether that issue exists or not. Note... dotcom has popovers over issues links that we are not currently doing, which makes this more feasible. But, does mean a different behavior than in dotcom. (This is currently our approach in our autocompletion logic)

This is my preferred approach. We don't have the luxury of being able to resolve like dotcom does. I think an "if it looks like it's an issue it's probably an issue" approach will work just fine for us. Given how rare cross-repo references are I could see us go for a hybrid approach where we consult our issues/pull stores (in a preloaded fashion probably) but chances are that would produce more false negatives (where we can't find an issue in the store because it's recently created) so it's likely not worth it.

cercia00:
I

AilMilk1100:
ail12121@icloud.com

Danindu-2005-24:
I am don't know how to program .

PR #13610, opened 2021-12-29
Fix commit message dialog inherited styles

### LINKED ISSUE #13606: Background color on “create a fork” prompt is wrong in squash dialog
### Describe the bug
When squashing two commits on a repo that you don’t have write access to, the background for the no write access warning is the wrong color.

### Version & OS

Version 2.9.6 (arm64)
macOS 12.1 (21C52)

1. Clone a repo you can’t push to
2. In the history tab, drag a commit onto another to squash them
4. See the behavior below

### Expected behavior

The warning has a white background 

### Actual behavior

Screenshot showing gray background

tidy-dev:
Closes #13606 

## Description
My use case is different from the bug ticket, but it is the same div, just a different warning.

### Screenshots

image

## Release notes
Notes: [Improved] Commit message dialog warning background styles congruent with dialog.

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -755,7 +755,6 @@ export class CommitMessage extends React.Component<
 
     return (
       <div
-        id="commit-message"
         role="group"
         aria-label="Create commit"
         className={className}
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -740,7 +740,7 @@ export class CommitMessage extends React.Component<
   }
 
   public render() {
-    const className = classNames({
+    const className = classNames('commit-message-component', {
       'with-action-bar': this.isActionBarEnabled,
       'with-co-authors': this.isCoAuthorInputVisible,
     })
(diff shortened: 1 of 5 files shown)


niik (commented):
The fact that these are `id`s makes me uncomfortable. Could we address that in the same PR by making both `commit-message` and `commit-warning` classes instead following the typical pattern we use (i.e. `commit-message-component` and `commit-warning-component`)?

tidy-dev:
Updated to use classes:
image

And showing main commit message area unaffected:
image

(I also tested in dark mode tho I wouldn't have expected anything there)

niik (changes requested):
Thanks a bunch for making those changes @tidy-dev, could we also remove the `id` attricute from the `CommitMessage` div wrapper now that we've turned it into a class?

niik (approved):

Mc480:
> Closes #13606
> 
> ## Description
> My use case is different from the bug ticket, but it is the same div, just a different warning.
> 
> ### Screenshots
> image
> 
> ## Release notes
> Notes: [Improved] Commit message dialog warning background styles congruent with dialog.

PR #13640, opened 2022-01-07
Request the workflow scope on GitHub Enterprise Server

niik:
## Description

Way back in https://github.com/desktop/desktop/pull/8340 we added support for the (then) new `workflow` scope with automatic scope upgrade for existing sessions. At the time GitHub Enterprise Server did not support GitHub Actions so we only implemented support for GitHub.com with the intention of coming back and adding support for GHES.

This PR does just that and in the process I've also opened a [docs PR](https://github.com/github/docs/pull/13807) to fix an omission where we don't show the 'workflow' scope in the documentation for GHES/GHAE

### Screenshots

https://user-images.githubusercontent.com/634063/148573872-c75efc8a-0cab-4455-8085-b54f52da147b.mov

## Release notes

Notes: [Added] Support pushing workflow files for GitHub Actions to GitHub Enterprise Server

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1534,7 +1534,7 @@ export async function fetchMetadata(
 
 /** The note used for created authorizations. */
 async function getNote(): Promise<string> {
-  let localUsername = await username().catch(_ => undefined)
+  let localUsername = await username()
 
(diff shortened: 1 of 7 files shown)


sergiou87 (approved):
Looks great ✨

PR #13641, opened 2022-01-07
Fetch before opening branch from url

Bestra:
Part of https://github.com/github/special-projects/issues/804

## Description
Currently, if a user creates a new branch on github.com, the GitHub Desktop client will be unable to open a link to the branch without manually fetching first. 

When we parse a link to a PR, we'll automatically try to fetch from the relevant remote here: https://github.com/desktop/desktop/blob/b81cdc419e8adebf30a4f55c26866331c37a050a/app/src/lib/stores/app-store.ts#L5832-L5841

This PR makes the desktop client fetch from the default remote (`origin` or the first one in alphabetical order if `origin` isn't configured) before switching branches. 

### Screenshots

https://user-images.githubusercontent.com/2043348/148781727-acccfdc8-b97c-4bdf-a359-75a70bf07489.mp4

## Release notes
- GitHub desktop will now fetch before trying to follow a URL link to a specific branch

Notes:

- I'll need help seeing if there's an idiomatic way to unit test any of this.
- Does fetching seem like a reasonable default, or do we want to hide this behind a query parameter in the URL?

### DIFF
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -1690,7 +1690,7 @@ export class Dispatcher {
     // if the repo has a remote, fetch before switching branches to ensure
     // the checkout will be successful
     if (isRepositoryWithGitHubRepository(repository)) {
-      await this.appStore._fetchDefaultRemote(repository)
+      await this.appStore._fetch(repository, FetchType.UserInitiatedTask)
     }
 
     await this.checkoutLocalBranch(repository, branchName)
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -1687,6 +1687,12 @@ export class Dispatcher {
     // up-to-date before performing the "Clone in Desktop" steps
     await this.appStore._refreshRepository(repository)
 
+    // if the repo has a remote, fetch before switching branches to ensure
+    // the checkout will be successful
+    if (isRepositoryWithGitHubRepository(repository)) {
+      await this.appStore._fetchDefaultRemote(repository)
+    }
+
     await this.checkoutLocalBranch(repository, branchName)
 
     return repository
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -1688,10 +1688,8 @@ export class Dispatcher {
     await this.appStore._refreshRepository(repository)
 
     // if the repo has a remote, fetch before switching branches to ensure
-    // the checkout will be successful
-    if (isRepositoryWithGitHubRepository(repository)) {
-      await this.appStore._fetch(repository, FetchType.UserInitiatedTask)
-    }
+    // the checkout will be successful. This operation could be a no-op.
+    await this.appStore._fetch(repository, FetchType.UserInitiatedTask)
 
     await this.checkoutLocalBranch(repository, branchName)
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -286,7 +286,6 @@ import { DragAndDropIntroType } from '../../ui/history/drag-and-drop-intro'
 import { UseWindowsOpenSSHKey } from '../ssh/ssh'
 import { isConflictsFlow } from '../multi-commit-operation'
 import { clamp } from '../clamp'
-import { findDefaultRemote } from './helpers/find-default-remote'
 
 const LastSelectedRepositoryIDKey = 'last-selected-repository-id'
 
@@ -5778,25 +5777,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
     if (prBranch !== undefined) {
       await this._checkoutBranch(repository, prBranch)
       this.statsStore.recordPRBranchCheckout()
-    }
-  }
-
-  /**
-   *
-   * Fetches from "origin" or the first alphabetical remote if "origin" doesn't exist.
-   * This function is a no-op if no remote is configured.
-   */
-  public async _fetchDefaultRemote(
-    repository: RepositoryWithGitHubRepository
-  ): Promise<void> {
-    const remotes = await getRemotes(repository)
-    const defaultRemote = findDefaultRemote(remotes)
-    if (defaultRemote) {
-      await this._fetchRemote(
-        repository,
-        defaultRemote,
-        FetchType.UserInitiatedTask
-      )
     }
   }
 
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -286,6 +286,7 @@ import { DragAndDropIntroType } from '../../ui/history/drag-and-drop-intro'
 import { UseWindowsOpenSSHKey } from '../ssh/ssh'
 import { isConflictsFlow } from '../multi-commit-operation'
 import { clamp } from '../clamp'
+import { findDefaultRemote } from './helpers/find-default-remote'
 
 const LastSelectedRepositoryIDKey = 'last-selected-repository-id'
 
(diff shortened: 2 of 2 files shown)


niik:
:wave: @Bestra thanks so much for picking this up! I took a look at it and it looks great but I'm afraid we might have to make this logic a bit more complicated because of how `openOrCloneRepository` works.

`openOrCloneRepository` will try to find a repository which is either the repository specified in the URL **or a fork of that repository**. This means that I suspect always using the default remote can get confusing.

Say that a user has a fork of a repository and uses the tool you're building to create a branch. Would that branch then be created in the user's fork? I suspect it would and if that's the case this logic would fail to checkout that branch given that it only fetches the default remote.

So there's two options here, the easy route where we just fetch all remotes or the slightly more complex route where we duplicate the logic of `openPullRequestFromUrl` which attempts to check out a branch from a specific remote. Just reasoning about this in my head didn't get me anywhere so I did a spike here for you to take a look at and see if I understand your use case well enough: https://github.com/Bestra/desktop/compare/development...desktop:niik/specific-remote-branch

Bestra:
> Say that a user has a fork of a repository and uses the tool you're building to create a branch. Would that branch then be created in the user's fork?

@niik you are totally correct! 

Using a hypothetical PR to `desktop/desktop`, this is the scenario I'd imagined.
image

I was thinking that the link we'd give in this situation would be:
```
x-github-client://openRepo/https://github.com/bestra/desktop?branch=some-branch
```

I'm probably misunderstanding how desktop works with forks 😅. I thought we'd have to clone the fork into a completely new repo from the upstream, and the fork itself would still be `origin` for the local copy. Am I missing something?

[edit] I think I've been looking at my specific use case, but you're properly considering the more general cases that opening a repo from a URL would need to account for.

niik:
@Bestra That's super helpful, thanks a lot, chances are I'm just overthinking this 😄 

I'm guessing no, but is there ever a scenario where a user might have a fork but would still get a link to the upstream repo? Like in my case where I happen to have a fork of desktop but I exclusively do work in the upstream repo?

> I'm probably misunderstanding how desktop works with forks 😅. I thought we'd have to clone the fork into a completely new repo from the upstream, and the fork itself would still be `origin` for the local copy. Am I missing something?

Desktop will try (though not very hard) to have just one repository representing both the upstream and the fork (by having the fork be the `origin` remote and the `upstream` remote be the upstream). So it should be relatively rare that you have both your fork and the upstream repository added as separate repositories.

To complicate things somewhat (for you and me that is, it's certainly only meant to simplify things for users :smile:) we let users choose how they wish to work within their fork

image
image

Where the default setting is the former.

Anyway, for your use case, if the fork is created as part of the workflow then Desktop would currently set up a new repository regardless of whether the parent repo is already added or not. Ideally that wouldn't be the case and we'd recognize that we've encountered a fork owner by the current user and offer to convert the existing clone to a fork instead but c'est la vie

Bestra:
> I'm guessing no, but is there ever a scenario where a user might have a fork but would still get a link to the upstream repo? Like in my case where I happen to have a fork of desktop but I exclusively do work in the upstream repo?

I could see that being possible.  I'll repeat the scenario back to make sure I'm getting it:
1. You have a local fork of desktop, `niik/desktop`
2. You create a branch in `desktop/desktop` for some feature, and you follow the link `x-github-client://openRepo/https://github.com/desktop/desktop?branch=niiks-awesome-feature`
3. Now we'd want to fetch the branch from `upstream` since `origin` is your fork, right?

If that's right, it seems at least plausible enough that it'd be nice to account for.  I owe you a huge thanks for following up! I'd happily pull the changes from https://github.com/Bestra/desktop/compare/development...desktop:niik/specific-remote-branch into this branch so we can take another look.

niik:
> I'll repeat the scenario back to make sure I'm getting it:

Yup, that's the one, but specifically in relation to the feature you're building. The scenario is possible today without your feature and we haven't had any real problems with that limitation so I don't want to pile onto your PR to solve a problem that's already existing in Desktop today. If your feature makes the above scenario more likely, however, then we should address it. So I guess the question is are users always going to have their branch created in their fork or will that depend on whether or not they have write access to the repository?

If not then I'd be content with just switching out that call to `fetchDefaultRemote` and using `AppStore#_fetch` instead which will fetch both the default and the upstream (if there is one) and consider the remote-specific work separately from this.

Bestra:
> So I guess the question is are users always going to have their branch created in their fork or will that depend on whether or not they have write access to the repository?

Up next for us will be https://github.com/github/special-projects/issues/782, and that'll allow the user to create a branch in any repo they have write access to.  I'd imagine we'd default to the repo for the Issue they're currently viewing (as long as they have write access).  Up until now I don't think we'd even specifically considered forks (🙈).  I think the most common use cases are going to be:
* people have write access to the Issue's repo and they're going to do their work there directly
* people are tracking work in a totally different repo than the one for the branch

My hypothesis is that people will use this feature less if they have to work in a fork,  but it'll be some time before we get data to figure that out.  

We've just [released this feature to private beta](https://github.com/github/feedback/discussions/3441#discussioncomment-1926157).  I've added mention to https://github.com/github/special-projects/issues/782 that we should also collect stats to see if people are creating their branches in the base repo or a fork if possible.  I'm inclined to go with `AppStore#_fetch` for now, and follow up with our users to see if they're running into any issues.

niik:
> I'm inclined to go with `AppStore#_fetch` for now, and follow up with our users to see if they're running into any issues.

That sounds good to me 👍  The lax nature of `checkoutLocalBranch` coupled with fetching all remotes will likely mean that it'll "just work" for most of users even if they target a fork vs parent since `checkoutLocalBranch` will pick the first branch it can find where the short name is equal.

Get that change in whenever you've got a moment and I'll review it again tomorrow and continue thinking about whether this is a scenario we need to build support for separately.

Thanks for letting me pester you for a bit 😛

Bestra:
As expected `_fetch` works perfectly, and now we're down to 6 added lines 🎉 

https://user-images.githubusercontent.com/2043348/149007277-f26176b7-4750-4cd0-89d6-b458aff1622a.mp4

Bestra:
Hi @niik, could I get a second (last 🤞 ) look at this PR this week? Thanks again!

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -1687,6 +1687,12 @@ export class Dispatcher {
     // up-to-date before performing the "Clone in Desktop" steps
     await this.appStore._refreshRepository(repository)
 
+    // if the repo has a remote, fetch before switching branches to ensure
+    // the checkout will be successful
+    if (isRepositoryWithGitHubRepository(repository)) {
+      await this.appStore._fetch(repository, FetchType.UserInitiatedTask)
+    }
```
niik: So this is rather nitpicky but `isRepositoryWithGitHubRepository` checks whether the Repository is hosted on GitHub.com/GHES/GHAE vs some other Git server. Is there a compelling reason to limit this to just GitHub repositories?

```suggestion
    await this.appStore._fetch(repository, FetchType.UserInitiatedTask)
```

niik (changes requested):
Sorry about the wait @Bestra, this looks great. Just one last question.

app/src/ui/dispatcher/dispatcher.ts
Bestra: @niik I don't think that's a nitpick at all 😄 ! We dont' have any reason to limit this to just GitHub repositories.  I think I modeled the type signature of [fetch_default_remote](https://github.com/desktop/desktop/pull/13641/commits/ffc6d887a9b9259b78f649a0361dbce5f6039475#diff-b238c18a99c08ab800e4227) (from my original commit) off of a function that dealt with a pull request and conflated "fetching" with "remotes". 

I'm glad you pointed it out, and thanks for reading carefully.   I've just added a commit to eliminate the 'if' check.

niik: Awesome, and thanks for tweaking that comment there as well, I love it!

niik (approved):
Thank you so much for this @Bestra!

Dawidaco (commented):
> @jonmagic

PR #13655, opened 2022-01-10
Initial support for high-signal notifications

sergiou87:
## Description
This PR contains the changes needed to start supporting high-signal notifications, including some basic infrastructure and the first kind of notification. Initially, the only notification supported will be when the checks of a PR you created failed which will show the dialog implemented in #13207

It also adds a new Advanced setting that allows users to disable high-signal notifications altogether. Since we intend these notifications to be very relevant to the user, this new setting will be enabled by default.

Some work still needs to be done on the dotcom side to make this work, but given everything should be properly feature-flagged, it should be safe to merge.

## Release notes

Notes: no-notes


app/src/lib/stores/alive-store.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (endpointSession === null || channelInfo === null) {
+      return
+    }
+
+    const subscription = {
+      subscriber: this,
+      topic: {
+        name: channelInfo.channel_name,
+        signed: channelInfo.signed_channel,
+        offset: '', // TODO: do we need to set an offset?
```
sergiou87: I still have to look into what to do about this, more in upcoming PRs…

app/src/lib/stores/alive-store.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  private notify = (subscribers: Iterable<AliveStore>, event: AliveEvent) => {
+    console.log('Alive event received:', event)
+
+    if (event.type !== 'message') {
+      return
+    }
+
+    // TODO: parse this safely
+    const desktopEvent = event.data as any
```
sergiou87: @niik is there a proper/cleaner way of handling this in TypeScript?

Technically, we could receive anything in that `data`. According to the type definitions, it should actually conform to a specific interface with these fields:

```typescript
    timestamp: number;
    wait: number;
    gid?: string;
```

But I don't see any of them being used in the `@github/alive-client` package, so I don't think it's required for us (I have to double check).

Thoughts?

niik: There are sadly no ways (that I know of) of taking a runtime object and checking its conformity to a compile-time type. At some point you need to tell typescript that the object can be anything and then decide how much you trust the service on the other end. You can certainly type check (at runtime) all the properties to build up a new object which is guaranteed to conform. Or you can toss all safety to the wayside and say "trust me" which is essentially what you're doing here.

So, in decreasing order of safety:

## Testing everything

```ts
const data = event.data as any
if (data.type === ('pr-checks-failed' as const)) {
  const {
    timestamp,
    owner,
    repo,
    pull_request_number,
    check_suite_id,
    commit_sha,
  } = data

  assert(typeof timestamp === 'number', 'Expected timestamp to be numeric')
  assert(typeof owner === 'string', 'Expected owner to be a string')
  assert(typeof repo === 'string', 'Expected repo to be a string')
  assert(typeof pull_request_number === 'number', 'Expected pull_request_number to be a string')
  assert(typeof check_suite_id === 'number', 'Expected check_suite_id to be a string')
  assert(typeof commit_sha === 'string', 'Expected commit_sha to be a string')

  const desktopEvent: IDesktopChecksFailedAliveEvent = {
    type: 'pr-checks-failed',
    timestamp,
    owner,
    repo,
    pull_request_number,
    check_suite_id,
    commit_sha,
  }

  this.emitter.emit(this.ALIVE_EVENT_RECEIVED_EVENT, desktopEvent)
}
```

## Test that the type matches what we expect it to and then trust the rest

```ts
const data = (event.data as any) as DesktopAliveEvent
if (data.type === 'pr-checks-failed') {
    this.emitter.emit(this.ALIVE_EVENT_RECEIVED_EVENT, data)
}
```

## YOOOOLOOOO

```ts
this.emitter.emit(this.ALIVE_EVENT_RECEIVED_EVENT, event.data as any)
```


My vote would be on number 2 here.

app/src/lib/stores/alive-store.ts
sergiou87: It seems this argument should be optional and we can ignore it safely 😌

app/test/unit/app-store-test.ts
```diff
@@ -78,6 +80,15 @@ describe('AppStore', () => {
...
     const apiRepositoriesStore = new ApiRepositoriesStore(accountsStore)
 
+    const aliveStore = new AliveStore(accountsStore)
+
+    const notificationsStore = new NotificationsStore(
+      accountsStore,
+      aliveStore,
+      pullRequestCoordinator
+    )
+    notificationsStore.setNotificationsEnabled(false)
```
sergiou87, line 90: @niik is there a better way of disabling notifications for tests?

app/src/lib/app-state.ts
```diff
@@ -294,6 +294,11 @@ export interface IAppState {
    * Whether or not the CI status popover is visible.
    */
```
NITTYG, line 295: [](url)

tidy-dev (changes requested):
It makes sense. I built and tested out that the setting worked as expected. Can't wait to test out the notifications part!

The only things to note is the chooselicense and gitignore submodule are changed. (which I know you are aware/going to clean up)

Also, seems there are some console.logs such as `console.log('Unubscribed from Alive channel!')` which look intentional data points -> wondering if we should use the IDesktopLogger for those or if those are going to go away eventually? But, I figured this is pretty minor so if that is to change, could be updated in a future pr.

sergiou87:
@tidy-dev yes, sorry, forgot to mention that I'm keeping those logs mostly for testing the feature, but my plan is to remove them (or use our logger if I think they're valuable) in a future PR.

Other than that, I reverted the changes in the submodules and all should be fine now 😌

tidy-dev (approved):
✨  Thanks for those changes and makes sense about console.log.

firebirdarkgit:
start

app/src/lib/stores/alive-store.ts
firebirdarkgit: S

PR #13662, opened 2022-01-12
Issue Mention Filter - Assume issues exist

tidy-dev:
Part of #11517

## Description

Based on dotcom's issue mention filter, it matches for issue references such as #1234, gh-1234, /issues/1234, /pull/1234, or /discussions/1234 and replaces them with anchor tags linking to the issue, pull request, or discussion in the parent repository. Those issue mentions can also be prefixed with an owner/repo which allows linking to a non-parent repo such as desktop/dugite#1. 

Additionally dotcom's issue filter searches for auto tagged urls such as `https://github.com/desktop/desktop/issues/1` and replaces the inner text with a uniform issue mention. This PR does that but I separated it into it's own filter since the logic was not really similar.

My first stabs #13573 and #13617 at this attempted to follow dotcom's approach tightly and verify the issue links through api calls. However, I found that to be very sluggish. This PR is trimming all the api call logic out and working on the assumption that issue mentions exist. This creates 2 scenarios for false positives, but they are both not likely/it doesn't benefit a user to do them anyways. Those 2 scenarios are linking to a non-existing issue such as #99999 and a non-exist or private repos that the user doesn't have access to external issue such as private/private#1. In both cases, the user attempts to follow the links, they will get a 404 not found page.

### Screenshots

Testing against - tidy-dev/foo-really-not#2

https://user-images.githubusercontent.com/75402236/149148530-87c6cb60-9286-4e6d-808b-7dd72f7725dd.mov

## Release notes
Notes: [Improved] Sandbox markdown parser parses issue mentions and auto tag issue links.

### DIFF
--- a/app/src/lib/markdown-filters/issue-mention-filter.ts
+++ b/app/src/lib/markdown-filters/issue-mention-filter.ts
@@ -1,0 +1,245 @@
+import { GitHubRepository } from '../../models/github-repository'
+import { getHTMLURL } from '../api'
+import { INodeFilter } from './node-filter'
+
+/**
+ * The Issue Mention filter matches for text issue references. For this purpose,
+ * issues, pull requests, and discussions all share reference patterns and
+ * therefore are all filtered.
+ *
+ * Examples:  #1234, gh-1234, /issues/1234, /pull/1234, or /discussions/1234,
+ * desktop/dugite#1, desktop/dugite/issues/1234
+ *
+ * Each references is made up of {ownerOrOwnerRepo}{marker}{number} and must be
+ * preceded by a non-word character.
+ *   - ownerOrOwnerRepo: Optional. If both owner/repo is provided, it can be
+ *              used to specify an issue outside of the parent repository.
+ *              Redundant references will be trimmed. Single owners can be
+ *              redundant, but single repo names are treated as non-matches.
+ *
+ *              Example: When viewing from the tidy-dev/foo repo,
+ *                  a. tidy-dev/foo#1 becomes linked as #1.
+ *                  b. tidy-dev#1 becomes linked as #1,
+ *                  c. foo#1 is not linked and is a non-match.
+ *                  d. desktop/desktop#1 is linked and stays desktop/desktop#1
+ *
+ *   - marker: Required #, gh-, /issues/, /pull/, or /discussions/
+ *   - number: Required and must be digits followed by a word bounding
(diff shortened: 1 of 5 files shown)


app/src/lib/markdown-filters/issue-mention-filter.ts
```diff
@@ -0,0 +1,245 @@
...
+  public async filter(node: Node): Promise<ReadonlyArray<Node> | null> {
+    const { textContent: text } = node
+    const markerRegexp = new RegExp(this.marker, 'i')
+    if (
+      node.nodeType !== node.TEXT_NODE ||
+      text === null ||
+      !markerRegexp.test(text)
+    ) {
+      return null
+    }
```
sergiou87: Maybe it's worth doing the `nodeType` and `text` checks before creating the `RegExp` object, since it's expensive to create. It'd be even better if we could cache it.
```suggestion
    if (
      node.nodeType !== node.TEXT_NODE ||
      text === null
    ) {
      return null
    }
    const markerRegexp = new RegExp(this.marker, 'i')
    if (!markerRegexp.test(text)) {
      return null
    }
```

app/src/lib/markdown-filters/issue-mention-filter.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const ownerAndRepo = ownerOrOwnerRepo.split('/')
+    // Invalid - This shouldn't happen based on the regex, but would mean
+    // something/something/something/#1 which isn't an issue ref.
+    if (ownerAndRepo.length > 3) {
+      return null
+    }
+
+    // Invalid - If it is only something/#1 and that `something` isn't the
+    // current repositories owner login, then we is not an actual, 'relative to
```
sergiou87: Nit
```suggestion
    // current repositories owner login, then it is not an actual, 'relative to
```

sergiou87 (dismissed):
Everything makes sense to me! Tested and seems to work well 😄 

Only found a couple of nits.

app/src/lib/markdown-filters/issue-mention-filter.ts
tidy-dev: I refactored slightly to make it so we just use the `marker` regex as defined on the filter instantiation. That way with the node filter being cached, the regex should only be created once.

sergiou87 (approved):
Looks great!! Thank you for those changes 😄

PR #13663, opened 2022-01-12
Use breaks: true option of marked.js

tidy-dev:
## Description
While working on the issue mention filter, I noticed that even with same styles applied to markdown on dotcom and in desktop. Where ever a newline character`\n` existed, we were not getting a line break in Desktop. Inspection on dotcom showed that it's markdown parser was inserting `<br>` where those were. I looked up the specs and it is part of GitHub Flavored Markdown (GFM), but there is a flag to enable in [Marked.js](https://marked.js.org/using_advanced). I enabled that flag. 

### Screenshots
Before 
Screen Shot 2022-01-12 at 8 31 29 AM

After:
image

## Release notes
Notes: [Improved] Sandbox Markdown Parser inserts newline characters with br tags.


sergiou87 (approved):
Makes total sense! :shipit:

PR #13678, opened 2022-01-16
Add link to open fork settings when creating new branch on a forked repository

### LINKED ISSUE #12534: When forked, "New Branch" is confusing for students
## Describe the feature or problem you’d like to solve

_Two related problems:_

- **As a** University Lecturer  
  **I want** students to make weekly branches and merges within a student fork
  **So that** students can learn a standard development process, I can review their code frequently, they can begin from a starter template, they avoid fiddly repo creation or fork decoupling actions, they are able to diff against my solution code.  
**=> problem: re-branch is not possible in GHD**

- **As a** Campus Advisor  
**I want** to promote a simple & intuitive Git Client for teaching  
**So that** more teachers are able to integrate source control, code review, CICD, etc into their classes  
**=> problem: I feel uncomfortable promoting GHD when it cannot do something fundamental that every other client in the industry can**

## Proposed solution

- **Solution:** Let the user change the source for new branches (i.e. "change" link next to the current source).
- **Benefits:** Without this feature, my students are forced to leave GitHub Desktop to use the CLI or another Git Client. At that point there is no point using GitHub Desktop. IMO this is a deal-breaker preventing adoption of GHD. With this feature, GHD is my first choice for teaching.

### My Use Case

I am a lecturer and our university's GitHub Campus Advisor. I'm trying to promote using GitHub in the classroom. I want to use industry tools & standard development processes as much as possible because I think the best learning environment is a real environment. I don't want to teach anything artificial that our students have to unlearn later.

I picked GitHub Desktop because it's no fuss and it "just works". It seems to me that education should be a major market for GitHub Desktop for this reason.

My students fork their repos from my template. **I want to use a fork so that they can easily diff against my solution code.** Each week they create a branch and PR their work (to their fork, not upstream). This process teaches them a normal development workflow, facilitates feedback, and allows them to compare their code to mine.

_Note: When cloning, my students explicitly select an option to make PRs against their forks, not the upstream._

I accept that there are almost certainly work-arounds; but this does not address problem (2) -- **in industry the expectation is that I'm able to branch; I don't want to break that in my classroom environment.**

### This is not about UX (default source)

IMO defaulting to the upstream is incredibly unintuitive, especially when my forked `develop` & `origin/develop` have commits that aren't in `upstream/develop`. I would have preferred a warning dialog "your branch is out of date", and perhaps a checkbox "always pull before branching". However, this is besides the point!

**I don't care what the default is. I just want to be able to change it, without leaving GHD.**

### Past Discussions

There is some existing discussion here: https://github.com/desktop/desktop/issues/9384. I tried to pick out some key quotes:

> This was an intentional change in #7762 based on the vast majority of forks being used to contribute to the upstream branch. So the previous behavior often would mean people working from forks in Desktop were starting new branches way out of date from the latest commits, and this was intended to help people ensure they're up to date when starting on a new branch.

> The reason why we've chosen not to provide a toggle in this particular case is that we think it's such a limited edge case

**I am proposing education as a not-so-limited edge-case. But I accept that other universities may have different processes or be more open to work-arounds (I have a particular aversion to diverging from industry), so I don't know how big this part of the market is.**

geophilusd:
Closes #12534 

## Description

- Added link to open fork settings when creating a new branch on a forked repository.

### Screenshots

Screenshot 2022-02-18 at 12 50 05 PM
Screenshot 2022-02-18 at 12 30 21 PM

## Release notes

Notes:
Added link to open fork settings when creating a new branch on a forked repository.


app/src/ui/create-branch/create-branch-dialog.tsx
```diff
@@ -422,8 +430,28 @@ export class CreateBranch extends React.Component<
...
+      )
+    }
+  }
 
-      return this.renderOptions(items, selectedValue)
+  private renderForkLink = () => {
+    if (isRepositoryWithForkedGitHubRepository(this.props.repository)) {
+      return (
+        <p>
+          Fork behaviour can be changed from {' '}
```
haykam821: This should probably match the American English spelling used in repository settings:

```suggestion
          Fork behavior can be changed from {' '}
```

villabellissima (commented):
Hello😶

tidy-dev (changes requested):
@geophilusd Thank you for wanting to improve the fork new branch interaction! 

I have a few small requests.

In other cases where we want to inform the user of a reason, but it may not be something they actually use regularly, we put the note in less eye catching text. 

As done in the commit message warning, "You can also choose..."
image

Here is a link to the code for that, 
https://github.com/desktop/desktop/blob/17afca94bb98a4edca6b21ccf539074945553133/app/src/ui/changes/commit-message-avatar.tsx#L154

Other thoughts are that "Fork behavior can be changed" does indicate that something on this dialog is impacted, but doesn't really tell us what is being impacted by it. What do you think of changing the verbiage "Your default branch source is determined by your `<link>` fork behavior settings`</link>`."

Also, usually we place these sort of notes directly underneath the input they impact. In this case, I think it should be placed underneath the branch selection. 

Similarly, I think this kind of note needs to be placed in the dialog when there is not a branch selection as well. (see first if block of the `renderForkBranchSelection` method)

geophilusd:
@tidy-dev I have made the suggested changes. Updated the screenshots in the PR description. Please check and let me know if I need to change something.

app/src/ui/create-branch/create-branch-dialog.tsx
```diff
@@ -348,12 +351,13 @@ export class CreateBranch extends React.Component<
...
     if (defaultBranch === null || defaultBranch.name === currentBranchName) {
       return (
-        <p>
+        <div>
           Your new branch will be based on your currently checked out branch (
           <Ref>{currentBranchName}</Ref>
           ). <Ref>{currentBranchName}</Ref> is the {defaultBranchLink} for your
           repository.
-        </p>
+          {this.renderForkLink()}
```
tidy-dev: I think we can be a little more concise in these "non-selection" instances by conditionally adding prefix to the existing informative text of: ```
As determined by your `<link>` for behavior settings `</link>`, your new branch will be based on your currently checked out branch. ```  Just got to be mindful of the 'y' vs 'Y' depending on if it is a fork or not.

app/src/ui/create-branch/create-branch-dialog.tsx
```diff
@@ -395,12 +404,13 @@ export class CreateBranch extends React.Component<
...
     if (currentBranchName === upstreamDefaultBranch.nameWithoutRemote) {
       return (
-        <p>
+        <div>
           Your new branch will be based on{' '}
           <strong>{upstreamRepositoryFullName}</strong>
           's {defaultBranchLink} (
           <Ref>{upstreamDefaultBranch.nameWithoutRemote}</Ref>).
-        </p>
+          {this.renderForkLink()}
```
tidy-dev: Same as comment on `renderRegulartBranchSelection` - Use prefix instead of additional sentence.. probably a good place for another method like `renderForkLinkPrefix`

app/src/ui/create-branch/create-branch-dialog.tsx
```diff
@@ -422,8 +432,28 @@ export class CreateBranch extends React.Component<
...
+      )
+    }
+  }
 
-      return this.renderOptions(items, selectedValue)
+  private renderForkLink = () => {
+    if (isRepositoryWithForkedGitHubRepository(this.props.repository)) {
+      return (
+        <div className="secondary-text">
+          Your default branch source will be determined by your{' '}
```
tidy-dev:
```suggestion
          Your default branch source is determined by your{' '}
```

app/src/ui/create-branch/create-branch-dialog.tsx
```diff
@@ -422,8 +432,28 @@ export class CreateBranch extends React.Component<
...
+          <LinkButton onClick={this.onForkSettingsClick}>
```
tidy-dev, line 448: This link is not displaying inline and seems to wrap oddly. Think we may want to add a css rule to the `.link-button-component` to be `display: inline;` when used in this dialog.

tidy-dev (changes requested):
@geophilusd Thank you for those changes! The secondary text under the options definitely feels more consistent to the rest of app.

However, for the instances where there are not branch options, I think it ends up being a little too verbose. I have commented a few suggestions to make it more concise and hopefully easier to digest at a glance.

app/src/ui/create-branch/create-branch-dialog.tsx
geophilusd: I have changed it to a suffix instead. Does that work?

app/src/ui/create-branch/create-branch-dialog.tsx
geophilusd: changed

app/src/ui/create-branch/create-branch-dialog.tsx
geophilusd: added 👍🏻 
Now the wrapping seems to be fine.

geophilusd:
@tidy-dev I have pushed some changes based on your suggestions. Please have a look.

tidy-dev (changes requested):
@geophilusd This is so close! 

However, one state that feels confusing is when we see (fork settings is non-contributing)
image
It seems to indicate that development is the default branch because of the fork settings, but that is not the case. This is just an informational sentence about default branches.

It is the first sentence that is influenced by the setting, we need: 
```
Your new branch will be based on your currently checked out branch (development) as determined by your fork settings. development is your default branch.
```

geophilusd:
@tidy-dev done

tidy-dev (approved):
✨  Thank you @geophilusd for making forking a little easier for new fork contributors in desktop!

PR #13682, opened 2022-01-17
Alive fixes and improvements

sergiou87:
## Description
This PR includes a number of fixes and improvements around Alive support:
- Fixed issue where Alive tried to reconnect but we were passing a deallocated instance of the `API` class.
- Fixed an issue where the app would subscribe twice to the Alive channel of the user's account.
- Fixed an issue where Alive would reject a WebSocket connection due to having the wrong `Origin` header. Still have to discuss with the Alive team about the best value for this `Origin` header.
- Clean up logging a bit.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/alive-store.ts
+++ b/app/src/lib/stores/alive-store.ts
@@ -1,6 +1,6 @@
 import { AccountsStore } from './accounts-store'
 import { Account, accountEquals } from '../../models/account'
-import { API } from '../api'
+import { API, getDotComAPIEndpoint } from '../api'
 import { AliveSession, AliveEvent, Subscription } from '@github/alive-client'
 import { Emitter } from 'event-kit'
 import { enableHighSignalNotifications } from '../feature-flag'
@@ -192,6 +192,11 @@ export class AliveStore {
   }
 
   private subscribeToAccount = async (account: Account) => {
+    // For now, Alive is only available for GitHub.com accounts
+    if (account.endpoint !== getDotComAPIEndpoint()) {
+      return
+    }
+
     const endpointSession = await this.createSessionForAccount(account)
     const api = API.fromAccount(account)
     const channelInfo = await api.getAliveDesktopChannel()
--- a/app/src/lib/stores/alive-store.ts
+++ b/app/src/lib/stores/alive-store.ts
@@ -195,7 +195,7 @@ export class AliveStore {
 
     endpointSession.session.offline()
 
-    console.log('Unubscribed from Alive channel!')
+    log.info(`Unubscribed '${account.login}' from Alive channel`)
   }
 
   private subscribeToAccount = async (account: Account) => {
@@ -223,11 +223,11 @@ export class AliveStore {
       subscription,
     })
 
-    console.log('Subscribed to Alive channel!')
+    log.info(`Subscribed '${account.login}' to Alive channel`)
   }
 
   private notify = (subscribers: Iterable<AliveStore>, event: AliveEvent) => {
-    console.log('Alive event received:', event)
+    console.log('Alive event received:', event) // TODO: delete this line after testing
 
     if (event.type !== 'message') {
       return
--- a/app/src/lib/stores/alive-store.ts
+++ b/app/src/lib/stores/alive-store.ts
@@ -1,9 +1,10 @@
 import { AccountsStore } from './accounts-store'
 import { Account, accountEquals } from '../../models/account'
-import { API, getDotComAPIEndpoint } from '../api'
+import { API } from '../api'
 import { AliveSession, AliveEvent, Subscription } from '@github/alive-client'
 import { Emitter } from 'event-kit'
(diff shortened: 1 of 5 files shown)


sanechek1 (commented):
bbb

app/src/main-process/install-web-request-filters.ts
```diff
@@ -53,7 +55,20 @@ export function installSameOriginFilter(webRequest: WebRequest) {
...
+
+    // If it's a WebSocket Secure request directed to a github.com subdomain,
+    // probably related to the Alive server, we need to override the `Origin`
+    // header with a valid value.
+    if (protocol === 'wss:' && /(^|\.)github\.com$/.test(host)) {
+      return cb({
+        requestHeaders: {
+          ...details.requestHeaders,
+          // TODO: discuss with Alive team a good Origin value to use here
+          Origin: 'https://desktop.github.com',
```
sergiou87, line 68: As mentioned in today's meeting, this is something that still needs to be discussed to make sure we use the right `Origin` before going to beta/production, but I think this should be good to go now that the feature is enabled only for dev builds.

app/src/lib/stores/alive-store.ts
```diff
@@ -83,18 +84,33 @@ export class AliveStore {
     this.subscribeToAccounts(accounts)
   }
 
-  private unsubscribeFromAllAccounts() {
+  private async unsubscribeFromAllAccounts() {
+    // Wait until previous (un)subscriptions finish
+    await this.accountSubscriptionPromise
```
niik, line 90: Awaiting null, what a world!

app/src/lib/stores/alive-store.ts
```diff
@@ -133,9 +149,16 @@ export class AliveStore {
...
+    // session. This function will be called multiple times if the session
+    // needs to reconnect.
+    const getRetryAliveURL = () => {
+      return api.getAliveWebSocketURL()
+    }
+
     const aliveSession = new AliveSession(
       webSocketUrl,
-      api.getAliveWebSocketURL,
+      getRetryAliveURL,
```
niik: > Fixed issue where Alive tried to reconnect but we were passing a deallocated instance of the API class.

It's not that it's deallocated, it's that javascript uses [runtime binding](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) so the alive clients ends up calling `getAliveDesktopChannel` with a different `this` (in this case it's actually `undefined`) than the API instance. So when `getAliveDesktopChannel` calls `this.request(` it's actually calling `undefined.request(`. So you could either use bind

```ts
api.getAliveWebSocketURL().bind(api),
```
or, an arrow function

```ts
() => api.getAliveWebSocketURL()
```

which is even shorter. Semantically these are the same as your current solution but this would be a little bit less code. Your call!

niik (dismissed):
Just one suggestion but I'm approving regardless of whether you apply it or not :smile:

app/src/lib/stores/alive-store.ts
sergiou87: Yup! From your question, I understand it is a new thing? 🤣 

However, [one small trade-off I learned](https://stackoverflow.com/a/55263084) is that even if the promise is `null`, `await` will yield to the event loop anyway. I don't think we should worry about that but I can do `if (this.accountSubscriptionPromise !== null)` to prevent that if you feel strongly about it.

niik: > Yup! From your question, I understand it is a new thing? 🤣

Just an old ghost from my C# days 😄 I'm perfectly happy with us doing so here, I don't think the yield will have any noticeable effects.

sergiou87:
@niik aside from your suggestion to wrap the websocket URL function, I added a couple of extra changes:
- I disabled this feature for non-GH.com accounts.
- The app will not show notifications when looking for the failed checks we get 0 (example: if new checks start running since the app received the Alive event and the failed checks are fetched)

Please take a look at those 🙏

app/src/lib/stores/alive-store.ts
```diff
@@ -192,6 +192,11 @@ export class AliveStore {
   }
 
   private subscribeToAccount = async (account: Account) => {
+    // For now, Alive is only available for GitHub.com accounts
+    if (account.endpoint !== getDotComAPIEndpoint()) {
```
niik: Do you think we could add an endpoint capability function for this instead? My hope is that we'll move away from this pattern. Something like

```ts
export const supportsAliveSessions = endpointSatisfies({
  dotcom: true,
  ae: false,
  es: false,
})
```

In `endpoint-capabilities.ts` and then invoke that instead?

niik (changes requested):

app/src/lib/stores/alive-store.ts
sergiou87: True! Sorry, I already forgot about it, I'm the worst 😂 

Changed in b7653f1

niik (approved):

PR #13685, opened 2022-01-18
Fix close button in full screen mode on macOS

### LINKED ISSUE #12838: Close button doesn't work while fullscreen on OSX 11.5.2
### Describe the bug
While running GitHub Desktop in fullscreen clicking on the red close button doesn't close the program. It only hides the menu bar again but GitHub Desktop stays open on screen as before.

If it's not running fullscreen the close button works as expected.

### Version & OS

I am currently running GitHub Desktop Version 2.9.2 (arm64) on OSX 11.5.2.
Note: I have been having this issue for about the last year. It was also present on my old MacBook, running the Intel based architecture. After switching to the Arm based MacBook and still seeing the same problem I decided to report the bug.

1. Open GitHub Desktop
2. Enter Fullscreen Mode
3. Click on the red close button

### Expected behavior

I expect the current GitHub Desktop window to close and keep running in the taskbar.

### Actual behavior

Nothing happens. The window stays open and in focus. You have to either exit fullscreen mode and then press the close button or quit the program using CMD + Q.

niik:
Closes #12838

## Description

Ran into this issue today and I can't for the life of me understand when it broke or if it's possibly never worked properly (since adding app hiding instead of close at least).

The problem is that `App.hide()` doesn't work when the window is in full screen mode. `BrowserWindow.hide()` does hide the window but leaves the user stuck with a blank screen in full screen mode. Taking the window out of full screen mode and then hiding it is the least controversial fix I could come up with here. The only downside is that users closing the window in full screen mode will have to wait for the macOS transition effect before the window gets hidden.

## Release notes

Notes: [Fixed] Closing the window using the title bar buttons on macOS now hides the window

### DIFF
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -83,7 +83,13 @@ export class AppWindow {
       this.window.on('close', e => {
         if (!quitting) {
           e.preventDefault()
-          app.hide()
+          // https://github.com/desktop/desktop/issues/12838
+          if (this.window.isFullScreen()) {
+            this.window.setFullScreen(false)
+            this.window.once('leave-full-screen', () => app.hide())
+          } else {
+            app.hide()
+          }
         }
       })
     }


sergiou87 (approved):
LGTM! Tested it and all good too :shipit:

PR #13689, opened 2022-01-18
Upgrade to Electron v14.2.3

tidy-dev:
## Description

This PR bumps Electron to the latest version of v14. With the bump to 14, the electron team moved the `remote` module into it's own package. This upgrade requires install `@electron/remote` and updating all of the `remote` imports to import from that package. To use this `remote` package, we have to invoke the `initialize` method in the main processor and invoke the `enable` method on the window `webContents` (both app and crash windows). This "should" be a lift and shift and not impact functionality of any of the places we utilize remote, but I have tested those places since they are most likely regression areas.

UPDATE: Tho the build and testing it out was working. Unit test were failing `electron_1.ipcRenderer.sendSync` type errors. I am assuming there should be a way to just tell the unit tests that that type is ok.. but in the instances that those occurred, it could be resolved by removing the dependency on the `remote` module. That is something we want to do in all cases `remote` so I went ahead with that route.

Also, the setting of [`worldSafeExecuteJavaScript` was removed](https://www.electronjs.org/docs/latest/breaking-changes#removed-worldsafeexecutejavascript) and defaults to true - we do not use either of the methods that are specified as being impacted by this change.

Here is link to version[ 14 change log.](https://www.electronjs.org/releases/stable?version=14)

## Release notes
Notes: [Improved] Upgrade to Electron 14.2.3


app/src/lib/stores/app-store.ts
```diff
@@ -406,7 +402,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
   private commitSummaryWidth = constrain(defaultCommitSummaryWidth)
   private stashedFilesWidth = constrain(defaultStashedFilesWidth)
 
-  private windowState: WindowState
+  private windowState: WindowState = 'normal'
```
sergiou87: Should we initialize this to `null` or `undefined` instead? Checking the usages of this variable, I didn't see how setting 'normal' here could ever be a problem. However, I think using `null` or `undefined` would be more honest in case it actually matters.

app/src/ui/window/window-controls.tsx
```diff
@@ -34,7 +30,7 @@ interface IWindowControlState {
  */
 export class WindowControls extends React.Component<{}, IWindowControlState> {
   public componentWillMount() {
-    this.setState({ windowState: getWindowState(remote.getCurrentWindow()) })
+    this.setState({ windowState: 'normal' })
```
sergiou87: Same about my other comment: maybe this should be initialized to `null` / `undefined`

app/src/ui/lib/update-store.ts
```diff
@@ -182,11 +189,13 @@ class UpdateStore {
...
-    } catch (e) {
-      this.emitError(e)
+    const checkedResults = await ipcRenderer.invoke(
+      'check-for-updates',
+      updatesURL
+    )
+
+    if (checkedResults !== null && checkedResults !== undefined) {
+      this.emitError(checkedResults)
     }
```
sergiou87: Probably worth renaming this to `error`?
```suggestion
    const error = await ipcRenderer.invoke(
      'check-for-updates',
      updatesURL
    )

    if (error !== null && error !== undefined) {
      this.emitError(error)
    }
```

app/src/main-process/app-window.ts
```diff
@@ -309,4 +328,66 @@ export class AppWindow {
...
+
+  public quitAndInstallUpdate() {
+    autoUpdater.quitAndInstall()
+  }
+
+  public disposeAutoUpdater() {
+    autoUpdater.removeAllListeners()
+  }
+
+  public subscribeNativeThemUpdated() {
```
sergiou87: Typo here:
```suggestion
  public subscribeNativeThemeUpdated() {
```

sergiou87 (changes requested):

app/src/lib/stores/app-store.ts
```diff
@@ -504,16 +500,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
       error => this.emitError(error)
     )
 
-    const browserWindow = remote.getCurrentWindow()
-    this.windowState = getWindowState(browserWindow)
-
-    this.onWindowZoomFactorChanged(browserWindow.webContents.zoomFactor)
```
niik, line 507: Why isn't this necessary any more? I would think we'd have to get a hold of the zoomFactor before it changes (like if the app starts with a zoomFactor other than 1)

app/src/lib/stores/app-store.ts
```diff
@@ -504,16 +500,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
       error => this.emitError(error)
     )
 
-    const browserWindow = remote.getCurrentWindow()
-    this.windowState = getWindowState(browserWindow)
```
niik, line 505: Wouldn't we want to initialize this (even if not synchronously?)

app/src/main-process/app-window.ts
```diff
@@ -145,6 +159,11 @@ export class AppWindow {
 
     this.window.webContents.on('did-finish-load', () => {
       this.window.webContents.setVisualZoomLevelLimits(1, 1)
+      this.window.webContents.send(
+        'zoom-factor-changed',
+        this.window.webContents.zoomFactor
+      )
+      sendWindowStateEvent(this.window, getWindowState(this.window))
```
niik: I don't think this is safe as we can't know for certain that the AppStore is set up to receive these events when `did-finish-load` happens, can we? `did-finish-load` is an Electron event that signals that the page has loaded. `renderer-ready` would be a better approach but I think the safest approach would be to request these states from the AppStore as we did before, only asynchronously this time round.

niik (changes requested):
Thanks for tackling this best of an upgrade @tidy-dev! I had a few questions.

In general I'd like for us to try to minimize `ipcRenderer.invoke(` calls all over the codebase and would prefer it if followed the existing convention of proxy functions in `main-process-proxy.ts`. Now that we're rapidly expanding the number of IPC calls and events we might also want to contemplate sharing the IPC event name and possibly even the type signature between the main process and the renderer so that we don't end up introducing bugs due to a typo in the event name.

app/src/lib/stores/app-store.ts
tidy-dev: Seeing your other comment, I guess you saw that I am initializing this as I was sending these values on app-window load. Tho I didn't know the concern about that possibly loading before the app-store since it tested out well.  But, no matter, working to use the invoke/handle to request the values from here (2nd stab at that.. was running into issue the first time I tried that).

app/src/lib/stores/app-store.ts
tidy-dev: Ditto to other comment.

app/src/main-process/app-window.ts
tidy-dev: I tried the approach of going from the app-store first, but I had other conflicting issue at the time.. so going to try again. Thanks for explaining the timing of 'did-finish-load' - I did test it out and the values were being set as I expected on the 'did-finish-load' but it sounds like that may not be an guaranteed case.

say25:
Don’t forget to update this file as well https://github.com/desktop/desktop/blob/main/app/.npmrc

tidy-dev:
There is one unit test that I was having trouble getting this branch to pass -> it is the one we usually see - the flaky one. But I think it was exasperated with a typing error now (which adding to the mock file didn't appear to help).  When I looked into it, it is the one we want to remove when we upgrade to 16 anyways because we are going to remove the spectron dependency. Thus, I went ahead and removed `spectron` and removed the integration test file (and related references - removed integration test call from ci.yml). Evidently `spectron` was bringing us `lodash` typing so I had to add that in separately.

app/src/ui/main-process-proxy.ts
```diff
@@ -46,6 +46,79 @@ export const showFolderContents = sendProxy('show-folder-contents')
...
+export function onAutoUpdaterUpdateDownloaded(eventHandler: () => void) {
+  ipcRenderer.on('auto-updater-update-downloaded', eventHandler)
+}
+
+/** Tell the main process to subscribe to native theme updated listener */
+export const subscribeNativeThemeUpdated = sendProxy(
+  'subscribe-native-theme-updated'
+)
+
+/** Subscribes to the native theme updated ecent originating from the main process */
```
niik:
```suggestion
/** Subscribes to the native theme updated event originating from the main process */
```

app/src/main-process/app-window.ts
```diff
@@ -307,4 +319,86 @@ export class AppWindow {
...
+
+  public disposeAutoUpdater() {
+    autoUpdater.removeAllListeners()
+  }
+
+  public subscribeNativeThemeUpdated() {
+    nativeTheme.addListener('updated', (event: string, userInfo: any) => {
+      ipcWebContents.send(this.window.webContents, 'native-theme-updated')
+    })
+  }
```
niik: This pattern lends itself to excessive subscriptions if we find another place in the renderer where we'd like to get notified of theme changes. I think we should either always subscribe to the event in the main process and send it along to the renderer such that the subscription in the renderer becomes the only deciding factor. Alternatively we can keep the current approach but make the method idempotent

```ts
  private subscribed = false
  
  public subscribeNativeThemeUpdated() {
    if (!this.subscribed) {
      nativeTheme.addListener('updated', this.onNativeThemeUpdated)
      this.subscribed = true
    }
  }

  private onNativeThemeUpdated = () =>
    ipcWebContents.send(this.window.webContents, 'native-theme-updated')
```

app/src/main-process/app-window.ts
```diff
@@ -74,6 +85,7 @@ export class AppWindow {
     })
 
     ipcMain.on('will-quit', event => {
+      nativeTheme.removeAllListeners()
```
niik: Given that the `native-theme-updated` event is scoped to the main window only I think it would make more sense to have `mainWindow` handle the unsubscription. The `will-quit` was an acceptable approach when using the renderer as we could rely on the remote module unsubscribing ipc events when the process exited.

```ts
this.window.on('closed', () => {
  nativeTheme.removeListener('updated', this.onNativeThemeUpdated)
})
```

```suggestion
```

app/src/ui/lib/update-store.ts
```diff
@@ -55,25 +62,20 @@ class UpdateStore {
...
+      onAutoUpdaterError(this.onAutoUpdaterError)
+      onAutoUpdaterCheckingForUpdate(this.onCheckingForUpdate)
+      onAutoUpdaterUpdateAvailable(this.onUpdateAvailable)
+      onAutoUpdaterUpdateNotAvailable(this.onUpdateNotAvailable)
+      onAutoUpdaterUpdateDownloaded(this.onUpdateDownloaded)
+      setupAutoUpdater()
+
+      window.addEventListener('beforeunload', () => {
+        disposeAutoUpdater()
+      })
```
niik: This should be the responsibility of the main process, `beforeunload` won't trigger if we destroy the window from the main process (which we do in the event of an uncaught exception for example).

Similar to the `nativeTheme` unloading above we should handle this in the `closed` event on the app window.

app/src/ui/lib/update-store.ts
```diff
@@ -55,25 +62,20 @@ class UpdateStore {
...
+    } catch (e) {
+      this.emitError(e)
+    }
```
niik: What kind of exception are we expecting here? It seems like wiring up event listeners should be fairly safe right?

app/src/main-process/main.ts
```diff
@@ -431,6 +433,50 @@ app.on('ready', () => {
...
+    mainWindow?.maximizeWindow()
+  })
+
+  ipcMain.on('unmaximize-window', () => {
+    mainWindow?.unmaximizeWindow()
+  })
+
+  ipcMain.on('close-window', () => {
+    mainWindow?.closeWindow()
+  })
```
niik: We can tighten all of these (and several of the other as well) up by using the short hand arrow function syntax
```suggestion
  ipcMain.on('minimize-window', () => mainWindow?.minimizeWindow())
  ipcMain.on('maximize-window', () => mainWindow?.maximizeWindow())
  ipcMain.on('unmaximize-window', () => mainWindow?.unmaximizeWindow())
  ipcMain.on('close-window', () => mainWindow?.closeWindow())
```

package.json
```diff
@@ -8,12 +8,11 @@
     "cli": "ts-node --require ./app/test/globals.ts --require ./app/src/cli/dev-commands-global.ts app/src/cli/main.ts",
     "check:eslint": "tsc -P eslint-rules/",
     "test:eslint": "jest eslint-rules/tests/*.test.js",
-    "test:integration": "cross-env TEST_ENV=1 ELECTRON_NO_ATTACH_CONSOLE=1 xvfb-maybe --auto-servernum -- jest --testLocationInResults --config ./app/jest.integration.config.js",
```
niik, line 11: I think we can remove the `xvfb-maybe` dependency now as well, right?

sergiou87 (dismissed):
Thanks for making those changes! LGTM 😄

niik (changes requested):

app/src/ui/lib/update-store.ts
tidy-dev: I don't know - That was copied from original renderer code. I didn't dig into why it was. Happy to remove it - especially since the auto updates appears to capture it errors and send them as event.

app/src/main-process/app-window.ts
tidy-dev: I like the subscribe in main process - less messaging code to follow.

app/src/ui/lib/update-store.ts
tidy-dev: Also, subscribed to the auto updater on load like the native theme.

package.json
tidy-dev: Yep. Good catch!

app/src/lib/window-state.ts
```diff
@@ -56,7 +56,7 @@ export function registerWindowStateChangedEvents(
  * Short hand convenience function for sending a window state change event
  * over the window-state-changed channel to the render process.
  */
-function sendWindowStateEvent(
+export function sendWindowStateEvent(
```
niik: Did we need this export?

```suggestion
function sendWindowStateEvent(
```

niik (dismissed):
One tiny question but feel free to ignore it and merge anyway. Great work on this @tidy-dev!

niik (approved):

PR #13702, opened 2022-01-19
Shared IPC type contract between renderer and main process

niik:
## Description

This relates to #13689 which we discussed this synchronously today. Our current IPC approach relies on copy-pasted code between the renderer and the main process where we hope that the channel names and the type signature match between the two.

This attempts to reduce that brittle approach by sharing a common type contract between the renderer and the main process.

## Release notes

Notes:

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -162,6 +162,11 @@ rules:
           message:
             "Please use 'import * as ipcRenderer' from 'ipc-renderer' instead to
             get strongly typed IPC methods."
+        - name: electron
+          importNames: ['ipcMain']
+          message:
+            "Please use 'import * as ipcMain' from 'ipc-main' instead to get
+            strongly typed IPC methods."
         - name: electron/main
           importNames: ['ipcMain']
           message:
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -149,6 +149,24 @@ rules:
     - error
     - global
   no-buffer-constructor: error
+  no-restricted-imports:
+    - error
+    - paths:
+        - name: electron
+          importNames: ['ipcRenderer']
+          message:
+            "Please use 'import * as ipcRenderer' from 'ipc-renderer' instead to
+            get strongly typed IPC methods."
+        - name: electron/renderer
+          importNames: ['ipcRenderer']
+          message:
+            "Please use 'import * as ipcRenderer' from 'ipc-renderer' instead to
+            get strongly typed IPC methods."
+        - name: electron/main
+          importNames: ['ipcMain']
+          message:
+            "Please use 'import * as ipcMain' from 'ipc-main' instead to get
+            strongly typed IPC methods."
 
   ###########
   # SPECIAL #
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -25,6 +25,7 @@ rules:
   react-no-unbound-dispatcher-props: error
   react-readonly-props-and-state: error
   react-proper-lifecycle-methods: error
+  no-loosely-typed-webcontents-ipc: error
 
   ###########
   # PLUGINS #
--- a/app/src/lib/logging/renderer/install.ts
+++ b/app/src/lib/logging/renderer/install.ts
@@ -1,8 +1,9 @@
-import { ipcRenderer } from 'electron'
 import { LogLevel } from '../log-level'
 import { formatLogMessage } from '../format-log-message'
+import { sendProxy } from '../../../ui/main-process-proxy'
 
 const g = global as any
+const ipcLog = sendProxy('log')
 
 /**
  * Dispatches the given log entry to the main process where it will be picked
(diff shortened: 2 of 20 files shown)


tidy-dev:
Approach looks great to me! Thank you for doing this typing wizardry to make things so much prettier!

I like the names on `sendProxy`, and `invokeProxy`, since they directly correlate with the IPC `send` and `invoke` methods and because of that the `onIPCMainRequest` feel off.. part of me wants it to say `onIPCMainHandleRequest` or have `onProxy` and `handleProxy` for consistency....and maybe all of those could have `IPC` in them. `sendIPCProxy`, `invokeIPCProxy`, `onIPCProxy`, `handleIPCProxy`. Just my random thoughts. :) 

What do you all think about also at this time adding a IPCChannelName enum since we have to use the channel names in three places (renderer, main, and channel definition)?

Also, in the electron upgrade, I refactored the autoUpdater logic to use the main process proxy and added "on" methods there that took a function as the param to move all the ipc channel references to the proxy. (In the #13689, you can see what I am talking about in the `update-store.ts` changes). Thoughts on adding this to the approach?

niik:
> I like the names on `sendProxy`, and `invokeProxy`, since they directly correlate with the IPC `send` and `invoke` methods and because of that the `onIPCMainRequest` feel off.. part of me wants it to say `onIPCMainHandleRequest` or have `onProxy` and `handleProxy` for consistency....and maybe all of those could have `IPC` in them. `sendIPCProxy`, `invokeIPCProxy`, `onIPCProxy`, `handleIPCProxy`. Just my random thoughts. :)

I love your random thoughts. I'll give this some more thought. Maybe we could try duplicating the ipcXYZ modules. Something like 

```ts
import * as ipcMain from 'ipc-main.ts'

ipcMain.on(...)
ipcMain.handle(...)
```

> What do you all think about also at this time adding a IPCChannelName enum since we have to use the channel names in three places (renderer, main, and channel definition)?

I'm not sure I see what an enum would give us that the current approach doesn't. The strings are effectively constants already as you can't pass in something that's not defined in the channel types and you already get the same level of autocompletion that you'd get by using an enum. They only thing I can think of is that they would let us put docstrings on the enum values but I doubt we would. As for the duplication we would still have to use the channel name _enum value_ in three places (renderer, main, and channel definition)?

> Also, in the electron upgrade, I refactored the autoUpdater logic to use the main process proxy and added "on" methods there that took a function as the param to move all the ipc channel references to the proxy. (In the #13689, you can see what I am talking about in the `update-store.ts` changes). Thoughts on adding this to the approach?

Yeah definitely, adding a type safe `.on` and `.handle` for the renderer as well as the main process is certainly on my list!

tidy-dev:
> I'm not sure I see what an enum would give us 

😸  Being able to say `IPCChannelName.` and visual studio code will help me finish it, so I don't have to worry about my ineptness of copy and paste. But like you said the build:dev yells at me when I do this, so either way is good.

tidy-dev:
> Maybe we could try duplicating the ipcXYZ modules. 

I like the sound of that.

niik:
> Being able to say `IPCChannelName.` and visual studio code will help me finish it, so I don't have to worry about my ineptness of copy and paste. But like you said the build:dev yells at me when I do this, so either way is good.

In the current case you only have to type a single `'` and you'll get the exact same level of autocompletion though.

https://user-images.githubusercontent.com/634063/150398186-ba215de0-aabd-4b8f-9336-7e23eb8150a4.mov

tidy-dev:
Ah yeah! I forgot that the new typing thing will give us that too! This things is so cool!

sergiou87:
Sorry I didn't have much time to look at this yesterday, but I like it overall, and the idea of "replacing"  the IPC objects sounds great! 😄

niik:
Alright,this is ready for a little lookiloo again. I've tried the approach of matching the API (albeit only a subset of it) and I dig it.

app/src/lib/logging/renderer/install.ts
```diff
@@ -1,20 +1,17 @@
-import { ipcRenderer } from 'electron'
 import { LogLevel } from '../log-level'
 import { formatLogMessage } from '../format-log-message'
+import { sendProxy } from '../../../ui/main-process-proxy'
 
 const g = global as any
+const ipcLog = sendProxy('log')
```
tidy-dev, line 6: Nit: For consistency with other calls like `moveToTrash`, seems like this could be in the `main-process-proxy` and import `ipcLog` from there instead of `sendProxy`.

tidy-dev (approved):
✨  Looks great! I pulled and tested it out. Also been working with it on my 16 upgrade branch when refactor remote references and so far it has worked and been fantastic.

One little nit, but feel free to ignore.

.eslintrc.yml
```diff
@@ -149,6 +150,29 @@ rules:
...
+        - name: electron
+          importNames: ['ipcMain']
+          message:
+            "Please use 'import * as ipcMain' from 'ipc-main' instead to get
+            strongly typed IPC methods."
+        - name: electron/main
+          importNames: ['ipcMain']
+          message:
+            "Please use 'import * as ipcMain' from 'ipc-main' instead to get
+            strongly typed IPC methods."
```
sergiou87, line 175: TIL about `no-restricted-imports` 😍

sergiou87 (approved):
I love it!! Awesome work 😍

PR #13719, opened 2022-01-21
Support avatars on GitHub Enterprise Server

niik:
## Description

Up until now we've relied on Gravatar to render avatars on GHES/GHAE since we haven't been able to load avatars from behind GHES private mode. As of GHES 3.0 we have access to an internal API that I built specifically for Desktop.

This PR leverages that API and in order to do so it adds one more web request filter which inserts the appropriate auth header for URLs matching the API endpoints. The reason we need a filter to begin with is that there's no way to specify auth headers in an `` tag which is ultimately what we're relying on to render these avatars.

You might ask why we don't use `fetch` or something instead and the reason is we used to and it was a mess. I'm sure we could have gotten it to work eventually but by relying on `` we get a lot of browser-level logic for free.

Given that we already had two web request filters (the same origin filter and the new alive filter from #13682) I decided to not overload our onBeforeRequest further and opted instead to create `OrderedWebRequest` which lets us install multiple filters which are then run in order.

## Release notes

Notes:


juggalettesx7:
@niik https://user-images.githubusercontent.com/57433155/150632809-030e11f8-d4a8-454d-bf41-5c7a2a94c16d.mp4

# [**_∆_****
∆**∆~~**∆*~~*†](s://OHHKACISTARZ.CLUB/)
https://user-images.githubusercontent.com/57433155/150632970-bfbca51c- -fe64-4313-b483-81e5d69ffe92.mp4
> - x```
``
```

File: app/src/main-process/ordered-webrequest.ts [hunk 1]
```diff
@@ -0,0 +1,254 @@
...
+  }
+}
+
+/**
+ * A utility class allowing consumers to apply more than one WebRequest filter
+ * concurrently into the main process.
+ *
+ * The WebRequest class in Electron allows us to intercept and modify web
+ * requests from the renderer process. Unfortunately it only allows one filter
+ * to be installed forcing consumers to build monolitchi filters. Using
```
tidy-dev:
```suggestion
 * to be installed forcing consumers to build monolithic filters. Using
```

tidy-dev: Nit: Typo

tidy-dev (changes requested):
✨  Makes sense. The Ordered Web request is slick. 

Nit: Just a typo I think.. monlitchi sounds like a cool word tho.

app/src/lib/stores/app-store.ts
```diff
@@ -696,6 +698,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     this.accountsStore.onDidUpdate(accounts => {
       this.accounts = accounts
+      const endpointTokens = accounts.map<EndpointToken>(
+        ({ endpoint, token }) => ({ endpoint, token })
+      )
```
sergiou87, line 708: Can't you avoid this `map` if you make `EndpointToken` an interface? 🤔 

You'd be passing the account instances, but only the `endpoint` and `token` fields would be accessible.

Just an idea, no need to do anything!

app/src/main-process/ordered-webrequest.ts
```diff
@@ -0,0 +1,254 @@
...
+          if (response.requestHeaders !== undefined) {
+            // I have no idea why there's a discrepancy of types here.
+            // details.requestHeaders is a Record<string, string> but
+            // BeforeSendResponse["requestHeaders"] is a
+            // Record<string, (string) | (string[])>. Chances are this was done
+            // to make it easier for filters but it makes it trickier for us as
+            // we have to ensure the next filter gets headers as a
+            // Record<string, string>
+            const requestHeaders = flattenHeaders(response.requestHeaders)
+            details = { ...details, requestHeaders }
```
sergiou87, line 187: 🤯

sergiou87 (dismissed):
LGTM! Just one suggestion/idea, feel free to ignore it.

app/src/lib/stores/app-store.ts
niik: I think it would have compiled even without making it an interface but given that `accounts` contain instances of `Account` classes I wanted to explicitly convert this to dumb objects before passing the IPC serialization boundary similar to https://github.com/desktop/desktop/blob/ef31b214980582a077bb2036002bc92daec6ee9b/app/src/ui/main-process-proxy.ts#L252-L258

It just gives us a little more explicit control over what we're sending is all.

tidy-dev (dismissed):
✨  Yay! Avatars!

say25 (dismissed):
Very excited for this feature. I'm more than happy to test this out as we are using a GHE server at my workplace

sergiou87 (dismissed):
:shipit:

tidy-dev (dismissed):

tidy-dev (approved):

PR #13735, opened 2022-01-24
Workaround for Windows notifications

sergiou87:
## Description
This PR has the following changes:
- It extracts the code used to show notifications, and changes the implementation to rely on the HTML5 Notification API instead of Electron's Notification module.
- It gives the focus to the app's window when a notification is clicked.
- It adds a new `Help` -> `Show notification` menu to aid development and debugging. When that notification is clicked, the `About` dialog is shown in the app.
- It adds a workaround for `click` events for notifications on Windows.

About that issue with notifications on Windows… the problem was that `click` events weren't reported back to the app in production builds. The reason is [Squirrel.Windows generates shortcuts with a "toast activator CLSID"](https://github.com/Squirrel/Squirrel.Windows/blob/76c87af6f389bea69576c738ef46574b2945b0ba/src/Squirrel/UpdateManager.ApplyReleases.cs#L182-L201) that Windows uses to decide which app to report these events back. In order for this CLSID to be useful, the app needs to implement a COM activator with that CLSID. That is not supported right now by Electron or GitHub Desktop, so as a temporary workaround, the app just goes through those shortcuts and removes the toast activator CLSID.

The main drawback of this approach (vs implementing a proper COM-based or protocol-based activator) is that the app won't get `click` events from notifications that were hidden away to the Windows Activation Center (see https://github.com/electron/electron/issues/29461).

## Release notes

Notes: no-notes


app/src/lib/fix-windows-shortcuts.ts
```diff
@@ -0,0 +1,57 @@
...
+      'AppData',
+      'Roaming',
+      'Microsoft',
+      'Windows',
+      'Start Menu',
+      'Programs',
+      'GitHub, Inc',
+      'GitHub Desktop.lnk'
+    ),
+    path.join(os.homedir(), 'Desktop', 'GitHub Desktop.lnk'),
```
sergiou87, line 28: I haven't seen a better way of getting these paths (at least not without additional dependencies).

app/src/ui/app.tsx
```diff
@@ -494,6 +497,18 @@ export class App extends React.Component<IAppProps, IAppState> {
...
 
+  private testShowNotification() {
+    if (!__DEV__) {
+      return
+    }
+
+    showNotification(
+      'Test notification',
+      'Click here! This is a test notification',
+      () => this.props.dispatcher.showPopup({ type: PopupType.About })
```
sergiou87, line 508: It's showing the about dialog because it's the easiest one and I didn't wanna create another dialog just for this 🤣 

I also didn't want to use `alert` because I wanted to make sure that clicking on the notification focused on the app's window.

app/src/lib/stores/helpers/show-notification.ts
```diff
@@ -0,0 +1,20 @@
...
+  title: string,
+  body: string,
+  onClick: () => void
+) {
+  const notification = new Notification(title, {
+    body,
+  })
+
+  notification.onclick = () => {
+    remote.getCurrentWindow().focus()
```
sergiou87, line 17: This should be replaced with something better other than `remote`. Probably worth doing the change after #13702 is merged.

tidy-dev: I will catch this with the remote 16 upgrade refactors (probably will already be making a proxy method for it)

tidy-dev (approved):
✨  Bummer as this does feel hacky. But, I tested it out and it worked as expected (Click doesn't work in notification center, but does when notification first appears)

app/src/lib/fix-windows-shortcuts.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    console.info(
+      `Found previous shortcut with toast activator CLSID in ${shortcutPath}. Fixing it...`
+    )
+
+    shell.writeShortcutLink(shortcutPath, 'replace', {
+      ...shortcutDetails,
+      toastActivatorClsid: '',
+    })
```
sergiou87, line 53: Forgot to mention that ideally this should be fixed on Squirrel.Windows. Probably make it an optional parameter where you can tell Squirrel to skip setting this CLSID.

sergiou87:
Closed in favor of #13767

PR #13741, opened 2022-01-25
Refactor `remote.app.isInApplicationFolder` to the main process

tidy-dev:
## Description

This refactors the `remote` module uses of `remote.app.isInApplicationFolder` method to use IPC messaging to the main process instead.

### Screenshots
On Windows, it correctly returned undefined - so no action taken.

On mac - correctly asked me to move to app folder,
image

## Release notes
Notes: no-notes


app/src/lib/stats/stats-store.ts
```diff
@@ -514,7 +515,7 @@ export class StatsStore implements IStatsStore {
 
     // isInApplicationsFolder is undefined when not running on Darwin
     const launchedFromApplicationsFolder =
-      remote.app.isInApplicationsFolder?.() ?? null
+      (await isInApplicationFolder()) ?? null
```
niik: It'd be great if we could avoid the IPC roundtrip and event loop dispatch here when we're not on macOS.
```suggestion
      ? (await isInApplicationFolder()) ?? null
      : null
```

app/src/ui/app.tsx
```diff
@@ -308,7 +312,7 @@ export class App extends React.Component<IAppProps, IAppState> {
     if (
       __DEV__ === false &&
       this.state.askToMoveToApplicationsFolderSetting &&
-      remote.app.isInApplicationsFolder?.() === false
+      (await isInApplicationFolder()) === false
```
niik, line 315:
```suggestion
      __DARWIN__ &&
      (await isInApplicationFolder()) === false
```

niik (changes requested):

app/src/lib/stats/stats-store.ts
tidy-dev: Makes sense. I also made the ipc call return null instead of undefined so we don't need the ?? anymore.

app/src/main-process/main.ts
```diff
@@ -532,15 +532,53 @@ app.on('ready', () => {
...
-   * Note: This will be undefined when not running on Darwin
+   * Note: This will return null when not running on Darwin
    */
   ipcMain.handle('is-in-application-folder', async () => {
-    return app.isInApplicationsFolder?.()
+    if (app.isInApplicationsFolder === undefined) {
+      return null
+    }
+
+    return app.isInApplicationsFolder()
```
niik: We had a comment to this effect at some point but it probably got lost, could we add it back so that we don't accidentally trust the type system and remove the undefined check? I think the shorthand optional chaining is neat but I'll accept either syntax

```suggestion
    // Contrary to what the types tell you the `isInApplicationsFolder` will be undefined
    // when not on macOS
    return app.isInApplicationsFolder?.() ?? null
```

niik (changes requested):

niik (approved):

PR #13742, opened 2022-01-25
Refactor `remote.dialog.showOpenDialog` to main process

tidy-dev:
## Description

This refactors the `remote` module use of `remote.dialog.showOpenDialog` method to use IPC messaging to the main process instead.

Impacts:
- Adding, cloning, creating a repo
- Relocating a missing repo

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -307,4 +307,17 @@ export class AppWindow {
   public destroy() {
     this.window.destroy()
   }
+
+  /**
+   * Method to show the open dialog and return the first file path it returns.
+   */
+  public async showOpenDialog(options: Electron.OpenDialogOptions) {
+    const { filePaths } = await dialog.showOpenDialog(this.window, options)
+
+    if (filePaths.length === 0) {
+      return null
+    }
+
+    return filePaths[0]
+  }
 }
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -531,6 +531,20 @@ app.on('ready', () => {
       UNSAFE_openDirectory(path)
     }
   })
+
+  /**
+   * An event sent by the renderer asking to show the open dialog
+   */
+  ipcMain.handle(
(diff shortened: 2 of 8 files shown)


app/src/main-process/main.ts
```diff
@@ -531,6 +531,20 @@ app.on('ready', () => {
...
+  ipcMain.handle(
+    'show-open-dialog',
+    async (_, options: Electron.OpenDialogOptions) => {
+      if (mainWindow === null) {
+        return null
+      }
+
+      return mainWindow.showOpenDialog(options)
+    }
+  )
```
niik, line 547: Let's see if we can't get this a little bit more succinct

```suggestion
  ipcMain.handle('show-open-dialog', async (_, options) => {
    return mainWindow?.showOpenDialog(options) ?? null
  })
```

app/src/main-process/app-window.ts
```diff
@@ -307,4 +307,17 @@ export class AppWindow {
...
+   * Method to show the open dialog and return the first file path it returns.
+   */
+  public async showOpenDialog(options: Electron.OpenDialogOptions) {
+    const { filePaths } = await dialog.showOpenDialog(this.window, options)
+
+    if (filePaths.length === 0) {
+      return null
+    }
+
+    return filePaths[0]
```
niik, line 321:
```suggestion
    return filePaths[0] ?? null
```

niik (approved):
Looks great @tidy-dev! Two purely stylistic requests but you can feel free to ignore them

app/src/main-process/main.ts
tidy-dev: I will have a pull request for the `show-save-dialog` -> I will update this on that PR since it will follow a similar pattern.

app/src/main-process/app-window.ts
tidy-dev: Ditto to last comment

PR #13745, opened 2022-01-26
Strict parameter count in IPC proxy methods

niik:
## Description

This fixes an issue where we could potentially encounter runtime errors using our new (#13702) IPC proxy methods. Given that it's possible to pass a `() => void` method (such as `getAppMenu`) to an event handler which will call it as `(e: Event) => void` we could end up unnecessarily marshaling values over the IPC boundary at best and attempting to marshal types incompatible with the structured clone algorithm at worst.

I don't know of a way to enforce this in a way that's completely transparent but by leveraging some type inference we can at least guarantee that it's not possible to create an ipc proxy and specify that it takes more or fewer parameters than the shared contract specifies.

## Release notes

Notes:


tidy-dev (dismissed):
✨  Looks great! Awesome typescript wizardry. 🧙‍♂️

sergiou87 (dismissed):
I love this 😍  Thank you!!!

niik:
@tidy-dev @sergiou87 could I please get a re-review of this post-merge?

tidy-dev (approved):

PR #13748, opened 2022-01-26
Refactor remote.dialog.showSaveDialog to main process

tidy-dev:
## Description

This refactors the `remote` module use of `remote.dialog.showSaveDialog` method to use IPC messaging to the main process instead.

Impacts: cloning dialog

## Release notes
Notes: no-notes


app/src/main-process/app-window.ts
```diff
@@ -317,16 +317,23 @@ export class AppWindow {
...
+  /**
+   * Method to show the save dialog and return the first file path it returns.
+   */
+  public async showSaveDialog(options: Electron.SaveDialogOptions) {
+    const { canceled, filePath } = await dialog.showSaveDialog(
+      this.window,
+      options
+    )
+
+    return !canceled ? filePath ?? null : null
```
niik: The docs for this method says that `filePath` will be undefined if the dialog was cancelled and it doesn't look like we actually care about the cancelled bool. Could we simplify this (I'm suggesting this knowing that you're refactoring and not changing legacy behavior, nevertheless it'd be a nice change I think)

```suggestion
    const { filePath } = await dialog.showSaveDialog(this.window, options)
    return filePath ?? null
```

app/src/main-process/app-window.ts
```diff
@@ -317,16 +317,23 @@ export class AppWindow {
...
    */
   public async showOpenDialog(options: Electron.OpenDialogOptions) {
     const { filePaths } = await dialog.showOpenDialog(this.window, options)
-
-    if (filePaths.length === 0) {
-      return null
-    }
-
-    return filePaths[0]
+    return filePaths[0] ?? null
```
niik: I know I suggested this approach but it _just_ dawned on me that since TypeScript doesn't think that filePath[0] can ever be undefined the return type of `showOpenDialog` will be `Promise<string>` instead of `Promise<string | null>`.

```suggestion
    return filePaths.length > 0 ? filePaths[0] : null
```

niik (changes requested):

app/src/main-process/app-window.ts
tidy-dev: Sounds good to me!

tidy-dev:
@niik I put the canceled back because the filepath returned as an empty string when you hit cancel. Which with our other current logic, cleared out the filepath input on the clone repository dialog.  (I supposed I could have just checked for empty string.. but the canceled is a bit more clear when reading it.)

niik:
> @niik I put the canceled back because the filepath returned as an empty string when you hit cancel. Which with our other current logic, cleared out the filepath input on the clone repository dialog. (I supposed I could have just checked for empty string.. but the canceled is a bit more clear when reading it.)

🤦 Never trust documentation I guess. Good on you for testing!

niik (approved):

PR #13750, opened 2022-01-26
Refactor `remote.getCurrentWindow().isFocused` to main process

tidy-dev:
## Description
This refactors the `remote` module use in `remote.getCurrentWindow().isFocused)` method to use IPC messaging to main process instead.

Impacts: 

- Index.tsx -> initial app state setup
- dispatcher.ts -> On mac, after oauth, used to determine if app needs refocused.

### Screenshots
With window in focus,
image

With window not in focus,
image

## Release notes
Notes: no-notes


sergiou87 (approved):
Looks good! I tested both scenarios you described and it works as expected in both 😌

PR #13761, opened 2022-01-27
Refactor `remote.app.getVersion()` to main process

tidy-dev:
## Description
This refactors the `remote` module uses of `remote.app.getVersion()` method to use IPC messaging to the main process instead.

Impacts: (Some are easier to test if you build:prod)
- About Dialog
- Acknowledgments Dialog
- Crash App
- Thank you logic
- Http request headers - user agent header
- Stats store

### Screenshots

- About Dialog
Screen Shot 2022-01-27 at 11 30 54 AM

- Acknowledgments Dialog
image

- Crash App
Screen Shot 2022-01-27 at 11 30 54 AM

- Thank you logic
image

- Http request headers - user agent header
image

- Stats store
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/crash/crash-app.tsx
+++ b/app/src/crash/crash-app.tsx
@@ -33,6 +33,9 @@ interface ICrashAppState {
    * The current state of the Window, ie maximized, minimized full-screen etc.
    */
   readonly windowState: WindowState
+
+  /** The version of the application */
+  readonly appVersion: string | null
 }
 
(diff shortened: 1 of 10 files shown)


niik:
The conversion of these methods to async caused a lot of change so I opened #13773 to embed the values at compile time. Take a look and see if you think that's a suitable replacement for this approach.

PR #13768, opened 2022-01-28
Upgrade to TypeScript 4.5

niik:
## Description

This upgrades TypeScript to 4.5 while steering well clear of upgrading Webpack which is another beast entirely. In order to avoid upgrading webpack I had to use ts-loader 8.x instead of 9.x.

Additionally this does not leverage the unknown type for errors in catch blocks which we should investigate turning on separately.

## Release notes

Notes: no-notes


app/src/lib/globals.d.ts
```diff
@@ -41,57 +41,6 @@ declare const __UPDATES_URL__: string
  */
 declare const __PROCESS_KIND__: 'main' | 'ui' | 'crash' | 'highlighter'
 
-/**
```
tidy-dev, line 58: With upgrade in typescript, do these things just exist and is why we don't need them in globals.d.ts anymore?

tidy-dev (dismissed):
✨  This seems to all make sense. Tho, I with my limited experience in this sort of thing, I am not sure I know what to look out for. :P

app/src/lib/globals.d.ts
sergiou87: Yup!

It seems like TypeScript goes a bit behind the latest API standard, and updates their [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts) later than we get new features from Electron/Chromium updates.

sergiou87 (dismissed):
Awesome job! Looks great to me 😄

niik:
Need a lil' re-review. Nothing has changed that hasn't already been reviewed though. We just merged https://github.com/desktop/desktop/pull/13807 into this branch

tidy-dev (approved):

PR #13767, opened 2022-01-28
Notification improvements

sergiou87:
## Description

This PR has the following changes:
- It extracts the code used to show notifications, and changes the implementation to rely on the HTML5 Notification API instead of Electron's Notification module.
- It gives the focus to the app's window when a notification is clicked.
- It adds a new `Help` -> `Show notification` menu to aid development and debugging. When that notification is clicked, the `About` dialog is shown in the app.
- It fixes getting notifications on Windows without a nice title and icon when they're shown from dev builds.

## Release notes

Notes: no-notes

PR #13769, opened 2022-01-28
Confirm re-running checks from "checks failed" dialog

sergiou87:
## Description

This PR sits on top of #13767 and adds the confirmation dialog introduced in #13369 to the "checks failed" notification dialog.

### Screenshots

https://user-images.githubusercontent.com/1083228/151540476-fff72e6e-336e-40fc-881e-1b101b2b8432.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/models/pull-request.ts
+++ b/app/src/models/pull-request.ts
@@ -1,5 +1,9 @@
 import { GitHubRepository } from './github-repository'
 
+export function getPullRequestCommitRef(pullRequestNumber: number) {
+  return `refs/pull/${pullRequestNumber}/head`
+}
+
 export class PullRequestRef {
   /**
    * @param ref The name of the ref.
--- a/app/src/models/pull-request.ts
+++ b/app/src/models/pull-request.ts
@@ -1,5 +1,6 @@
 import { GitHubRepository } from './github-repository'
 
+/** Returns the commit ref for a given pull request number. */
 export function getPullRequestCommitRef(pullRequestNumber: number) {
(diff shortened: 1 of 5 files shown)


app/src/models/pull-request.ts
```diff
@@ -1,5 +1,10 @@
 import { GitHubRepository } from './github-repository'
 
+/** Returns the commit ref for a given pull request number. */
+export function getPullRequestCommitRef(pullRequestNumber: number) {
+  return `refs/pull/${pullRequestNumber}/head`
+}
```
sergiou87, line 6: Happy to move this somewhere else, but this file seemed like a good fit?

tidy-dev (approved):
✨  Works great! Makes sense!

Only thing that comes to mind is that on cancel of rerun.. should the other dialog come back? On the check run popover because it is a popover not a dialog, cancel returns user to the popover. I know that is not our current dialog convention, but, I just hit cancel was like.. how do I get back to those failed checks? Seeing as this is really a design concern, I don't think it should hold up this PR. :)

sergiou87:
Yeah, I was thinking the same. Before this PR, the dialog was (supposed to) dismiss immediately after hitting `Rerun checks` (it didn't because of a bug 🤦 🤣 ), so in that regard this doesn't regress that behavior: once you click on `Rerun checks`, the other dialog is lost.

Anyway, before making more complex logic I'd rather gather more feedback and see what response we get to this behavior 🤞

PR #13773, opened 2022-01-28
Embed product name and version at compile time

niik:
Superseedes #13761

## Description

Product name and version are determined at build time so we might as well embed them as global constants so avoid having to do any IPC roundtrip to get them.

## Release notes

Notes: no-notes


tidy-dev (approved):
This is awesome! Way better than changing everything to async. :D

PR #13775, opened 2022-01-28
Refactor `remote` usage in title to main process

tidy-dev:
## Description
This refactors the `remote` module usage  in the `title-bar.tsx` to the main process instead.

Impacts: On mac, when you double click the title bar, it should minimize/maximize the app depending on user's system preference. 

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/ipc-shared.ts
+++ b/app/src/lib/ipc-shared.ts
@@ -100,4 +100,8 @@ export type RequestResponseChannels = {
   'show-open-dialog': (
     options: Electron.OpenDialogOptions
   ) => Promise<string | null>
+  'is-window-maximized': () => Promise<boolean>
+  'get-apple-action-on-double-click': () => Promise<
+    Electron.AppleActionOnDoubleClickPref
+  >
 }
--- a/app/src/main-process/app-window.ts
+++ b/app/src/main-process/app-window.ts
@@ -399,6 +399,10 @@ export class AppWindow {
     this.window.close()
   }
 
+  public isMaximized() {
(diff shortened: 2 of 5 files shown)


sergiou87 (commented):
Code looks good! Just a comment (maybe beyond the scope of this PR): I couldn't test this properly as double-clicking the title bar did nothing to me.

I checked and it's because `getAppleActionOnDoubleClick()` returned `""` for me. Once I went to System Preferences -> Dock & Menu Bar and toggled the mentioned setting, I wouldn't get anything.

After a few tests, I think it's just that `""` is the default value and is equivalent to `Maximize`, so probably worth doing returning `Maximize` in that case from https://github.com/desktop/desktop/blob/430695699e14e38cebec0ab0665361a8d6762763/app/src/lib/globals.d.ts#L217-L223

A good way to test this is running `defaults delete NSGlobalDomain AppleActionOnDoubleClick` to return this setting to the default value.

What do you think? Since your PR doesn't regress the previous behavior, I'm ok with merging it, but it'd be nice to fix this 😅

tidy-dev:
That makes sense. I first attempted to just check if empty string and if so return `Maximize`, but the typescript type thinks it can't be anything but `Maximize`, `Minimize`, or `None`. Thus, I handled it in the switch statement, which thinks that the default is `Maximize` only, but now handles the `` as well (Of course pitfall is also any other setting that could be added in the future, but we would likely have had to adapt for that anyways).

sergiou87 (approved):
Thanks for that change!

Works great now 😄

PR #13790, opened 2022-01-31
Refactor `remote.app.runningUnderRosettaTranslation` to main process

tidy-dev:
## Description
This refactors the `remote` module usage of `remote.app.runningUnderRosettaTranslation` to the main process instead.

Impacts: 
- The update store checks against this and and Arm64Translation to determine if it should request an arm64 update. 

I tested this locally by debugger and verified I received the same value as before which was false.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/lib/update-store.ts
+++ b/app/src/ui/lib/update-store.ts
@@ -1,10 +1,10 @@
-import * as remote from '@electron/remote'
 const lastSuccessfulCheckKey = 'last-successful-update-check'
 
 import { Emitter, Disposable } from 'event-kit'
 
 import {
   checkForUpdates,
+  isRunningUnderRosettaTranslation,
   onAutoUpdaterCheckingForUpdate,
   onAutoUpdaterError,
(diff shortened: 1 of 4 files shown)


app/src/ui/lib/update-store.ts
```diff
@@ -160,7 +160,7 @@ class UpdateStore {
     // the arm64 binary.
     if (
       enableUpdateFromEmulatedX64ToARM64() &&
-      (remote.app.runningUnderRosettaTranslation === true ||
+      ((await isRunningUnderRosettaTranslation()) === true ||
         isRunningUnderARM64Translation() === true)
```
sergiou87, line 164: FYI Electron 15 adds my contribution of the [`runningUnderARM64Translation`](https://www.electronjs.org/docs/latest/api/app#apprunningunderarm64translation-readonly-macos-windows) property which unifies these two checks.

sergiou87 (approved):
Looks good to me!

PR #13792, opened 2022-01-31
Refactor spell check menu remote usage to main process

tidy-dev:
## Description
This refactors the `remote` module usage for building the spell check menu to the main-process.

The reasoning behind moving the whole build the spell check menu logic over to the main process side is because there are several instances of `remote` being used so would be multiple ipc channels made for something that doesn't have a need to be in the renderer. I also believe the code is a little easier to reason by separating the merging/building part. 

I also moved all the menu logic in the `main-process-proxy` to the `menu-item` file. It seemed to me the menu logic was cluttering the `main-process-proxy` and was not directly relevant in the file.

Impacts: Potentially all context menus... 

## Release notes
Notes: no-notes


app/src/lib/menu-item.ts
```diff
@@ -72,3 +74,58 @@ export function getPlatformSpecificNameOrSymbolForModifier(
...
+      menuItem.action()
+    }
+  }
+}
+
+/**
+ * Remove the menu items properties that can't be serializable in
+ * order to pass them via IPC.
+ */
+function serializeMenuItems(
```
tidy-dev, line 101: `serializeMenuItems` and `findSubmenuItem` are pretty much lift and shift (just a documentation change on on).

app/src/main-process/menu/build-spell-check-menu.ts
```diff
@@ -0,0 +1,112 @@
...
+    dom.
+  */
+  return new Promise(resolve => {
+    window.webContents.once('context-menu', (event, params) =>
+      resolve(getSpellCheckMenuItems(event, params, window.webContents))
+    )
+  })
+}
+
+function getSpellCheckMenuItems(
```
tidy-dev, line 22: `getSpellCheckMenuItems` is the moved version of `mergeDeferredContextMenuItems`. But it doesn't merge items any more, but the logic for building them should be the same (asides from making MenuItems as opposed to IMenuItems now)

app/src/main-process/main.ts
```diff
@@ -450,14 +451,24 @@ app.on('ready', () => {
...
+   * process as opposed to the renderer.
    */
-  ipcMain.handle('show-contextual-menu', (event, items) => {
-    return new Promise(resolve => {
-      const menu = buildContextMenu(items, indices => resolve(indices))
+  ipcMain.handle('show-contextual-menu', (event, items, addSpellCheckMenu) => {
+    return new Promise(async resolve => {
       const window = BrowserWindow.fromWebContents(event.sender) || undefined
 
+      const spellCheckMenuItems = addSpellCheckMenu
```
tidy-dev, line 462: Something I wasn't sure of with this approach is that the buildContextMenu takes a call back with the menu items indices so it can send back to the renderer which item was clicked (and that is how spell check menu items were invoked previously). However, if we send it items with roles like edit-menu, it adds menu items such as cut, copy, paste and when those are clicked null is returned. I was unsure if the spell check menu items had to specifically have a call back sending back null to the renderer since it is no longer responsible for invoking the action, but it appears they are handled just as if you are clicking off the menu or one of the electron role based menu items and `showContextMenu` does get back a null - I am assuming that is the `menu.popoup... resolve(null)` being invoked.

app/src/main-process/menu/build-spell-check-menu.ts
```diff
@@ -0,0 +1,112 @@
+import { app, BrowserWindow, MenuItem } from 'electron'
+
+export async function buildSpellCheckMenu(
+  window: BrowserWindow | undefined
+): Promise<ReadonlyArray<MenuItem> | undefined> {
+  if (window === undefined) {
+    // window.webContents.off('context-menu', mergeDeferredContextMenuItems)
```
sergiou87: This used to be a thing before this PR. Unless I'm missing something, it looks it's not needed anymore so we can probably remove this line.

sergiou87 (changes requested):
The code looks good, I don't see anything obviously wrong at least (only one nit about a line that can be removed).

Aside from that, I tested this on macOS and Windows and all menus seemed to work fine (I also tried spell checking from Windows when it's set in Spanish).

app/src/main-process/menu/build-spell-check-menu.ts
tidy-dev: Ah thanks, I forgot to remove that.  I copied that over in case I would need it, but I am not sure why it was there in the first place since we are using `.once` and it seems to work correctly without it... Only scenario I thought of was if you opened another context menu very quickly (before the `context-menu` event fires), which in the original logic flow could mean merging with the wrong context menu since we were storing it. Which again means we shouldn't need this with the new approach.

sergiou87 (approved):

PR #13794, opened 2022-01-31
Refactor remote usage in app's theme settings to main process

tidy-dev:
## Description
This refactors the remaining `remote` module usage in the app's theme settings to the main-process.
` remote.nativeTheme.themeSource = themeName`
`remote.nativeTheme.shouldUseDarkColors`

Impacts: Theme setting
1) Setting whether dark, light, or system is the native theme source. This is important if system is set to dark, but user is set to light in desktop. It uses light.
2) Used to determine if the system is using light or dark.

## Release notes
Notes: no-notes


app/src/ui/lib/application-theme.ts
```diff
@@ -34,14 +37,14 @@ export type ApplicableTheme =
   | ApplicationTheme.Dark
   | ApplicationTheme.HighContrast
 
+export type ThemeSource = 'light' | 'dark' | 'system'
```
niik: Given that this is being used by both the main process and the renderer now could we lift it out into its own file? Chances are it'll never be a problem but when we're importing renderer-specific files into the main process we always run the risk of running code that's only meant for the renderer (like if there was a `localStorage.get` call triggered by just importing for example).

app/src/ui/preferences/appearance.tsx
```diff
@@ -50,7 +54,45 @@ const themes: ReadonlyArray<ISegmentedItem<ApplicationTheme>> = [
...
+  }
+
+  public async componentDidUpdate(prevProps: IAppearanceProps) {
+    if (prevProps.selectedTheme === this.props.selectedTheme) {
+      return
+    }
+
+    const usePropTheme =
+      this.props.selectedTheme !== ApplicationTheme.System ||
+      supportsSystemThemeChanges()
```
niik, line 82: I recognize that this is the exact same logic from before, only in its inversed form but I can't understand it. Why do we want to `usePropTheme` when `supportsSystemThemeChanges` is true?

niik (changes requested):

app/src/ui/preferences/appearance.tsx
tidy-dev: I think the logic says,
If user is light or dark and their system doesn't support theme changes, we just use light/dark as that is what came from the local store. (and will be their only options)

If user is set to system or support for system themes, go get what is applied via the system.

Possibly this logic may be able to be simplified if this check was done outside of this component.

tidy-dev: Hmm.. no it doesn't. Just if it is set to system theme go get what is set or  it does not support system themes.. go get what is set? That does seem strange.. but I guess our possibilities are 
1) light/dark and supports system changes
2) light/dark and doesn't supports system changes
3) system and supports system changes.. cause it can't be system if your system doesn't support it. :D

tidy-dev: Based on my last reasoning/comment, I am not sure if we even need to check if it supportsSystemThemeChanges here...

tidy-dev: Meh.. I am confused.. it would seem the only time this check matters is if we have it set as system and system is not supported (How would that happen?).. cause then it would go get light or dark..

tidy-dev: Ah.. we default to system. :D Ok.. Definitely this logic would be less confusing if we did this `supportsSystemThemeChanges` up at the app store and set it to System or result of `getCurrentlyAppliedTheme` :) But... I think that could be a refactor for another PR

niik: Thanks for digging into this @tidy-dev!

niik (approved):

PR #13797, opened 2022-01-31
Fix Windows notifications with `desktop-notifications`

sergiou87:
## Description
Since Electron notifications don't Just Work™ on Windows because of [the lack of a COM-based activator](https://github.com/electron/electron/issues/29461), we decided to write our own module that we'll keep simple and will give more control over what we can/want to support: [`desktop-notifications`](https://github.com/desktop/desktop-notifications)

This PR adds this new library as a dependency, and picks [the toast activator CLSID generated by Squirrel](https://github.com/Squirrel/Squirrel.Windows/blob/7396fa50ccebf97e28c79ef519c07cb9eee121be/src/Squirrel/UpdateManager.ApplyReleases.cs#L258) for the shortcuts created during the app's installation.

### Screenshots

https://user-images.githubusercontent.com/1083228/151840416-5ab578ba-ae41-4ddc-9be5-ebb6a29426dc.mov

## Release notes

Notes: no-notes


Attila180 (commented):
...

tidy-dev (changes requested):
~~Discussed this with Sergio, just for other reading, when testing this, I had only the Beta app installed on my windows vm and when I attempted to show a notification it crashed.  I installed Prod app and it worked. (issue of where the shortcut is looked for).~~ Seems to have been a fluke? Moved back to beta app and it works fine. 

Other thoughts is may be better to use our desktop logger as opposed to `console.error`. (unless those are just there temp for development)

Other notes from that conversation was that this could be improved to not crashing the app in the case of not finding the shortcut - just not show notifications and log error out.

sergiou87:
@tidy-dev I just fixed the `console.` logs, I missed that yesterday 🤦 

Ready for re-review now!

sergiou87:
I forgot to mention: https://github.com/desktop/desktop-notifications/commit/8f10ebe78a6fb4efc23eaff26418af23ad2081e8 is the commit that fixes the crash when notifications aren't initialized.

tidy-dev:
Ok.. I am probably convolutely testing this (not a likely real scenario.. but I wanted to poke at it's edges). 

1. I currently have beta app only installed.
2. I deleted my desktop and start menu shortcut.
3. I tried the notification - based on debugger, it couldn't find shortcut/clsid. (didn't crash yay!)
4. It shows a notification without the GitHub Desktop icon.
Results..
1. if I click the notification immediately it, it does as expected and opens the about modal.
2. if I click on the notification in the windows notification center, it does nothing 
   -  I figured this may be expected due to not having the clsid, but wasn't sure so I am noting it.
3. (This is sporadic.. maybe just dev build fluke/related to web socket not windows notifications?) If I click on the x of the notification in the notification center, my dev app goes all white. Here is what is in console.
image


After that, I added the start menu short cut back and expected it to work correctly. However, still not showing Desktop icon and not working from notification center. In the debugger, it does show finding the Clsid.

After that, I created a new shortcut for the desktop, and now it does show the Github Desktop icon and works from the notification center.  

Got the white screen again for hitting x on a notification after now getting the desktop icon back, just it happens to be triggering it. 

Other notes: It does not look like the notification shown in your screen cast. (different OS? I am on Windows 10)
image


TLDR; 3 things, 1) no desktop or start menu shortcut - it doesn't work (thinking expected as we rely on shortcuts). 2) no desktop shortcut, it finds a clsid, but still doesn't work as expected. 3) random white screen for dismissing notifications in notification center

I think 1) and 3) are unlikely things and 3 not directly related to this PR.  2) does kind of concern me because I know some people do not like having desktop shortcuts.. and seems we shouldn't need to rely on this. But, I would still not hold up this PR for that since I know this is still under development.

tidy-dev (dismissed):
✨  Code looks good.. Read my other comment for testing results and you can decide if that should impact this PR merging.

sergiou87:
Thanks for such a thorough testing!

I haven't seen the random white screen and not sure how that could be possible, since `desktop-notifications` only sends `click` events back to the app 🤯  Of course, I haven't been able to repro that 😕 

> After that, I added the start menu short cut back and expected it to work correctly. However, still not showing Desktop icon and not working from notification center

This sounds surprising, I'd expect that to work too. It's not a big concern, but interesting. Did you restart (not reload) after restoring the shortcut? Windows requires us to set the AUMID at the beginning of the app, so maybe it requires us to also have the shortcut set up by then 🤔 

> Other notes: It does not look like the notification shown in your screen cast. (different OS? I am on Windows 10)

Yep! I have Windows 11.

I'm surprised (and worried) also about the rate limit that you hit. I don't think that's related to the Alive socket but it might be. In theory once it connects, it doesn't hit that endpoint again. And the cadence of retries is the same used on GitHub.com… 🤞 

In any case, this is probably safe to merge (once I fix the conflicts, of course 😩 ) and deploy to beta 🎉 🚀

niik:
> I'm surprised (and worried) also about the rate limit that you hit. I don't think that's related to the Alive socket but it might be.

Even more confusing is that that error is specifically about unauthenticated requests. When making requests to the API without a token (for the endpoints which allow that) the rate limit is only 60 request per hour IIRC. I can't understand how we'd send unauthenticated requests for `alive_internal/websocket_url` though and I don't understand how we'd get a response back without having an authenticated user either so I'm wondering if this could be a dotcom rate limit issue?

tidy-dev (approved):

PR #13803, opened 2022-02-01
Refactor `remote.app.getAppPath()` to main process

tidy-dev:
## Description
This refactors the `remote` module usage in `remote.app.getAppPath()` to the main-process.

Impacts:
- Loading emoji's is now async - being a deferred independent action, seems like it is reasonable to be async not be waited upon
- Acknowledgements: loading the licences.json is now async

Note: This is one of several PR's to refactor our usage of the `remote` module to instead use the IPC messaging with the main process so that we can remove our [undesired](https://nornagon.medium.com/electrons-remote-module-considered-harmful-70d69500f31) `remote` module dependency (supporting ultimate goal of upgrading to electron 16).

## Release notes
Notes: no-notes

PR #13804, opened 2022-02-01
Refactor `remote.app.get.Path()` to main process

tidy-dev:
## Description
This refactors the `remote` module usage in `remote.app.getPath(pathname)` to the main-process.

Impacts:
For the following, obtaining this path becomes asynchronous. 
- Clone Repository - is used to determine the default clone path
- Create Repository - is used to determine the default create path
- Creating Tutorial Repo - used to determine repo path

## Release notes
Notes: no-notes


app/src/ui/lib/app-proxy.ts
```diff
@@ -60,14 +79,13 @@ export function getUserDataPath(): string {
  *
  * This is preferable to using `remote` directly because we cache the result.
  */
-export function getDocumentsPath(): string {
+export async function getDocumentsPath(): Promise<string> {
   if (!documentsPath) {
-    const app = getApp()
```
sergiou87, line 52: Now that #13803 is merged, I think you can also remove the `getApp` function entirely, right?

sergiou87 (commented):
It looks and works great! Just a comment/suggestion that you might tackle in this PR.

app/src/ui/lib/app-proxy.ts
tidy-dev: Yep! That was my plan - not sure it would have even built after merging in dev.

sergiou87 (approved):
✨

PR #13807, opened 2022-02-01
TypeScript 4.5 linting

niik:
## Description

This change was intended to be part of the TS 4.5 PR (#13768) but the changes grew large enough that I felt it warranted a separate PR in order to make it easier to review.

The `tslint-microsoft-contrib` package didn't have a compatible peer dependency with TS 4.5 and while it still worked (kinda) the package has been deprecated and archived since long ago. So I set out to try and find replacements for as many of the rules that I could and remove any obsolete rules.

I'll make some comments in the diff below to make it easier to review.

## Release notes

Notes:


.eslintrc.yml
```diff
@@ -100,9 +99,7 @@ rules:
   '@typescript-eslint/no-explicit-any': off
   '@typescript-eslint/no-inferrable-types': off
   '@typescript-eslint/no-empty-function': off
-
-  # Babel
-  babel/no-invalid-this: error
```
niik, line 105: I don't know what was happening here but as soon as I bumped one of the `@typescript-eslint` packages this started reporting errors all over the place so I had to disable it. I can't remember a single time this rule has saved us so I'm not concerned about it going away.

.eslintrc.yml
```diff
@@ -114,6 +111,8 @@ rules:
   react/jsx-uses-vars: error
   react/jsx-uses-react: error
   react/no-unused-state: error
+  react/no-unused-prop-types: error
```
niik, line 114: This isn't actually working at the moment but we want it as soon as it does. I've opened https://github.com/yannickcr/eslint-plugin-react/pull/3195 to discuss possible solutions

app/src/ui/tab-bar-item.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react'
```
niik, line 1: This file and others like is are here because of the `no-use-before-declare` rule being unhappy with our previous state of affairs.

niik:
With this we're loosing the `promise-must-complete` (https://github.com/desktop/desktop/blob/development/tslint.json#L5) rule which was kinda nice but I haven't found an alternative and I don't think that should prevent us from upgrading and getting rid of tslint.

.eslintrc.yml
sergiou87: And it's been merged already! Yay!

sergiou87 (approved):
🎉

tidy-dev (approved):
✨

PR #13813, opened 2022-02-02
Enable manual "check for updates" on test builds

tidy-dev:
## Description
This PR enables manual updates for the test channel. 

Release 2.7.1-test21 is a release with these changes. 

The behaviors should be that - it should not check for updates on load. I checked that by putting a debugger here:
image

Then, when you go to the About dialog, you should see a update state and "Check for Updates" button.
image

Pressing "Check for Updates" should hit the "onCheckingForUpdate" method in the "update-store".
image

Now that I have test 2.7.1-test22, it should download that update. Note: test22 does not have this manual check for updates change it. It is the electron 16 upgrade.

As for my prev mention that "MacOs it doesn't and gets the prod release", I forgot that the central test release changes were rolled back when Sergio's notifications metric branch was deployed. I redeployed central with the test build changes and now everything works as expected from Desktop and Central.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -90,8 +90,8 @@ export async function getChangeLog(
     'https://central.github.com/deployments/desktop/desktop/changelog.json'
   )
 
-  if (__RELEASE_CHANNEL__ === 'beta') {
-    changelogURL.searchParams.set('env', 'beta')
+  if (__RELEASE_CHANNEL__ === 'beta' || __RELEASE_CHANNEL__ === 'test') {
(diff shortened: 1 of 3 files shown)


niik (approved):
This looks good to me but I'd love it if @sergiou87 could also look it over just so that we've got three :eyes: on changes to auto-update logic. I've had to deal with the outcome of accidentally nerfed auto-updates before 😛

sergiou87 (approved):
It looks good to me too! Should be safe to merge 🤞

Issue #2540, opened 2017-08-24
utf-8


Cirbles:
### Description

[Description of the issue]

### Version

**GitHub Desktop version:** [version here]

**OS version:** [version here]

### Steps to Reproduce

1. [First step]
1. [Second step]
1. [and so on]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

**Reproduces how often:** [What percentage of the time does it reproduce?]

tierninho:
@Cirbles Please fill out the template so we have more info to help you. Thanks.

PR #13814, opened 2022-02-03
Upgrade Electron 16

tidy-dev:
Note: This has the changes from #13794 included (merge that first). 

## Description

This PR bumps Electron 16.0.8 and node to 16.9.1.

Notes: [^1]
1. Had to add `NODE_OPTIONS='--max_old_space_size=4096'` due our unit test runs as it was causing actions workflow failures.
2. Minor typing change `Electron.session` became `Electron.Session`

Here is link to version[ 16 change log.](https://www.electronjs.org/releases/stable?version=16)

## Release notes
Notes: [Improved] Upgrade to Electron 16.0.8

[^1]: NodeJS team [released the windows arm64](https://github.com/nodejs/build/issues/2540#issuecomment-1030991902) build!  Outdated Note: Node version 16.2.0 doesn't match Electron 16's node version on 16.9.1, but using 16.2.0 does not seem to have any ill side effects and not necessary for this upgrade. 
 -- The reason for this is that node windows arm64builds are not released with every node build, and 16.2.0 is the closest previous version with it. Currently, creating these builds is manual process for the Node.js team, I have requested 16.9.1 build on the [#2540](https://github.com/nodejs/build/issues/2540). But, it looks like [they are close to it not being manual](https://github.com/nodejs/build/pull/2860), so I am not sure when/if releasing a manual build for 16.9.1 will be get done and is a blocker for us anyways. If they get a non-manual process for it, likely next electron upgrade, this won't be a thing. :)


package.json
```diff
@@ -8,7 +8,7 @@
     "cli": "ts-node --require ./app/test/globals.ts --require ./app/src/cli/dev-commands-global.ts app/src/cli/main.ts",
     "check:eslint": "tsc -P eslint-rules/",
     "test:eslint": "jest eslint-rules/tests/*.test.js",
-    "test:unit": "cross-env ELECTRON_RUN_AS_NODE=1 ./node_modules/.bin/electron ./node_modules/jest/bin/jest --detectOpenHandles --silent --testLocationInResults --config ./app/jest.unit.config.js",
+    "test:unit": "cross-env ELECTRON_RUN_AS_NODE=1 NODE_OPTIONS='--max_old_space_size=4096' ./node_modules/.bin/electron ./node_modules/jest/bin/jest  --detectOpenHandles --silent --testLocationInResults --config ./app/jest.unit.config.js",
```
tidy-dev: After electron upgrade, the actions workflows would sometimes get a ["Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory"](https://github.com/desktop/desktop/runs/5051290822?check_suite_focus=true) on running unit tests. Adding the node-options to increase memory allocation like we do in the `compile:prod` seems to resolve the issue.

sergiou87: Biggest (or smallest) nit ever: you added an extra space, sorry 😂 
```suggestion
    "test:unit": "cross-env ELECTRON_RUN_AS_NODE=1 NODE_OPTIONS='--max_old_space_size=4096' ./node_modules/.bin/electron ./node_modules/jest/bin/jest --detectOpenHandles --silent --testLocationInResults --config ./app/jest.unit.config.js",
```

sergiou87 (commented):
Top quality code review. You're welcome…


😅

sergiou87 (approved):
Code LGTM! And I've tested it on macOS (x64 and arm64) and Windows (arm64 only) and didn't find any issues, everything worked as expected.

niik (approved):

PR #13835, opened 2022-02-07
Surround `api.getAliveWebSocketURL` in try-catch

sergiou87:
## Description

Probably harmless, but still cleaner this way. I ran into this while I tried to repro the "white screen" issue mentioned in https://github.com/desktop/desktop/pull/13797#issuecomment-1027983548

It's worth mentioning that we only need to catch errors in this specific usage of `api.getAliveWebSocketURL`, but not in the lambda passed to the `AliveSession`. That `AliveSession` will catch any exceptions `getAliveWebSocketURL` throws and handle them accordingly.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/alive-store.ts
+++ b/app/src/lib/stores/alive-store.ts
@@ -144,7 +144,14 @@ export class AliveStore {
     }
 
     const api = API.fromAccount(account)
-    const webSocketUrl = await api.getAliveWebSocketURL()
+    let webSocketUrl = null
+
+    try {
+      webSocketUrl = await api.getAliveWebSocketURL()
+    } catch (e) {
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨  Makes sense. 

I looked at `api.getAliveWebSocketURL` and saw that the try/catch block, `log.warns` then bubbles the error.  Do we need it to bubble the error or could this change be simplified by changing `api.getAliveWebSocketURL try/catch block to `log.error`/return null (we do that in other api calls)?

sergiou87:
> Do we need it to bubble the error … ?

Yup! We need to bubble the error because `AliveSession` is expecting that. It expects the URL if all goes 👍 , `null` on 404, and an exception in other cases. I just forgot to handle that exception in the other only place where we use that API 😓

Issue #13816, opened 2022-02-03
Clone Error - Seems we should have better clone path validation


tidy-dev:
### Describe the bug
Found that you can provide any single string into the clone path and it will allow you to hit clone, but will fail.

### Version & OS

2.9.7-beta1 
Mac Os

### Steps to reproduce the behavior

1. Go to `Clone a Repository` dialog
2. Go to URL tab
3. Enter a valid url or username/repository in top input. - May need to be a github repo. I was using desktop/desktop.
4. Input a word  or even just a space in the `Local Path` input.
5. Hit Clone.

### Expected behavior
Inform me that this is not a valid local path.

### Actual behavior

Attempts to clone, but ultimately receive an error in the dialog.
![image](https://user-images.githubusercontent.com/75402236/152352202-5ccbf8f4-294c-4c90-b2f4-004c2a799e69.png)

### Screenshots
https://user-images.githubusercontent.com/75402236/152352330-4576bdec-f2d8-4485-9ffb-89c2918f1e98.mov

### Logs
![image](https://user-images.githubusercontent.com/75402236/152351716-70c80437-1e65-4590-9760-b774b4bbc2e1.png)

### Additional context
I found this when testing the `remote` module usage refactor that impacted the cloning. This refactor is not in beta 2.9.7-beta1 - It should not be an outcome of current work, but a existing edge case bug. Should be tested in a prior version and in windows.

LunarBlacksmith:
I'm only new to GitHub and version control stuff, but when updating the Desktop app today I made my last commit to main branch and then restarted for update to Version 2.9.13 and when re-opening I was greeted with a message saying that my repository that I had been using for 2 weeks wasn't a repository!
![error](https://user-images.githubusercontent.com/79879930/163174500-07e66cda-bd8f-4bca-a4dd-5601bbc08385.png)

Following this, I tried to locate it manually twice with no luck, same error. 
Then I tried deleting the local repo and cloning from the internet again. Once it finished cloning it came up and gave the same error! Tried it twice with no luck, making sure the folder was completely deleted prior to cloning.
I then tried to manually download from the cloud and choose to _add a local repository from HDD_ but it would simply fail and say that it wasn't a repository. Eventually I found the console log for the app and it read as follows:
![console_log](https://user-images.githubusercontent.com/79879930/163175212-90964137-bade-4160-a6ce-a963b191e1af.png)

[...]

PR #13844, opened 2022-02-07
Don't throw when repo couldn't be added

niik:
Refs #13816

## Description

This doesn't fix the problem described in #13816 but it gets rid of an exception that should never happen when we attempt to access an element out of bounds of the returned array from `addRepositories`.

### Screenshots

image

## Release notes

Notes:


tidy-dev (approved):
✨  Makes sense

PR #13847, opened 2022-02-08
Replace `node-detect-arm64-translation` dependency with `app.runningUnderARM64Translation`

sergiou87:
## Description

Now that we upgraded to Electron 16, we can use `app.runningUnderARM64Translation` instead of our [node-detect-arm64-translation](https://github.com/desktop/node-detect-arm64-translation) package. While testing this, I noticed a change in the behavior of IsWow64Process2 since I made the package, which is what both `app.runningUnderARM64Translation` and [node-detect-arm64-translation](https://github.com/desktop/node-detect-arm64-translation) use: it used to return a non-success value when the current process wasn't being translated. Now, it returns a success value even when the process is arm64. This means that the current implementation of both Electron's API and our package will return `true` when the CPU is arm64.

So I tried to make some changes to stick to the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2), and look for `nativeMachine == IMAGE_FILE_MACHINE_ARM64 && processMachine != IMAGE_FILE_MACHINE_UNKNOWN`

However, on the latest Windows 11 build (22000.469) for arm, `processMachine` is always `IMAGE_FILE_MACHINE_UNKNOWN`, meaning we think it's not emulated whenever the app is running on an arm64 device, no matter what the architecture of the process is.

That's why this PR takes the current implementation of `app.runningUnderARM64Translation` and also checks `process.arch == 'x64'` to determine whether or not we're running a x64 build of the app on arm64 devices.

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -28,7 +28,6 @@
     "deep-equal": "^1.0.1",
     "desktop-notifications": "^0.1.6",
     "desktop-trampoline": "desktop/desktop-trampoline#v0.9.8",
-    "detect-arm64-translation": "https://github.com/desktop/node-detect-arm64-translation#v1.0.4",
     "dexie": "^2.0.0",
     "dompurify": "^2.3.3",
     "double-ended-queue": "^2.1.0-0",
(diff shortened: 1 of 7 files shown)


niik (approved):

niik:
Thanks for the thorough explanation @sergiou87! :sparkles:

PR #13852, opened 2022-02-09
Loosen detection of web flow committer on GHES

### LINKED ISSUE #13848: GitHub Enterprise User Showing
### Describe the bug
GitHub Enterprise User Showing in commit history

### Version & OS

Version 2.9.7-beta2 (x64) - Windows 10

1. On a GHE repo, got to commit history

### Expected behavior

GitHub History matches GHES webpage

### Actual behavior

GitHub History does not match GHES webpage

### Screenshots

image

### Logs

### Additional context

This seems to have been introduced with the  changes to add Avatars from GHES 3.0+. https://github.com/desktop/desktop/pull/13719

We are currently on GHES 3.1.14

niik:
Closes #13848

## Description

Turns out the email address used for the web flow committer on GHES is admin-configurable so we can't rely on it matching the domain of the GHES instance.

I did briefly consider trying match on a substring of the domain but while likely rare it is possible for a company to have their GHES hosted on `domainA.com` while having their noreply emails originating from `domainB.com`.

## Release notes

Notes: [Fixed] Hide committer for commits made using the web flow on GitHub Enterprise Server

### DIFF
--- a/app/src/lib/web-flow-committer.ts
+++ b/app/src/lib/web-flow-committer.ts
@@ -31,9 +31,19 @@ export function isWebFlowCommitter(
     return true
   }
 
-  if (name === 'GitHub Enterprise') {
-    const host = new URL(endpoint).host.toLowerCase()
-    return email.endsWith(`@${host}`)
+  if (endpoint !== getDotComAPIEndpoint() && name === 'GitHub Enterprise') {
+    // We can't assume that the email address will match any specific format
+    // here since the web flow committer email address on GHES is the same as
+    // the noreply email which can be configured by domain administrators so
+    // we'll just have to assume that for a GitHub hosted repository (but not
+    // GitHub.com) a commit author of the name 'GitHub Enterprise' is the web
+    // flow author.
+    //
+    // Hello future contributor: Turns out the web flow committer name is based
+    // on the "flavor" of GitHub so it's possible that you're here wondering why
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
I think this is the safest option even if we have to update it with newer flavors of GHE as you mention in the comment 😅 

🚀

app/src/lib/web-flow-committer.ts
```diff
@@ -31,9 +31,19 @@ export function isWebFlowCommitter(
     return true
   }
 
-  if (name === 'GitHub Enterprise') {
-    const host = new URL(endpoint).host.toLowerCase()
-    return email.endsWith(`@${host}`)
+  if (endpoint !== getDotComAPIEndpoint() && name === 'GitHub Enterprise') {
```
say25, line 34: I think this is safe, we could I'm not sure if email starts with `ghe-noreply` would also work or be better/worse?

say25 (approved):

app/src/lib/web-flow-committer.ts
niik: It looks like that part is configurable as well unfortunately.

say25: 🤯

say25 (approved):

PR #13853, opened 2022-02-09
Upgrade code mirror typings

tidy-dev:
## Description
 A while back I [opened a PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55661) to add the `lookAhead` method to the StringStream type. This pr is an upgrade to `@types/codemirror` to the version that has this so I can get the [stale high contract pr](https://github.com/desktop/desktop/pull/12922) merged.

However this is kind of worrisome because we need to go from `0.0.76` -> `5.60.4`. (It looks like a bigger jump than it is because version numbering scheme was changed a few years ago, but still 35+ versions) Luckily, there were only 3 build errors as a result.

1) The `CodeMirrorHost.onChanges` callback typing, this shouldn't impact us since we don't even use this callback.
2) Per the docs, `mode` now must be set to 'null' if we want to explicitly not use the first mode loaded. Notes, this was introduced post the version of codemirror we are using - thus should be compatable.
3) We can remove the `editor-configuration-extra.ts` as it was extending the `Codemirror.EditorConfiguration` with properties that now exist on that type. 
     - However, before the property of `scrollbarStyle` was of type `'native' | 'simple'` , and` 'simple'` doesn't exist by default as it is part of an add on. This type needs explicitly imported from the addon typing to support our usage for it on macOs.
      - Additionally, the `lineSeparator` is now went from `string | undefined` to `string | null | undefined`, but the doc editor still expects `string | undefined` so we null coalesce to `string | undefined`

## Release notes
Notes: no-notes


app/src/ui/diff/text-diff.tsx
```diff
@@ -58,6 +57,7 @@ import * as OcticonSymbol from '../octicons/octicons.generated'
 import { WhitespaceHintPopover } from './whitespace-hint-popover'
 import { PopoverCaretPosition } from '../lib/popover'
 import { HiddenBidiCharsWarning } from './hidden-bidi-chars-warning'
+import 'codemirror/addon/scroll/simplescrollbars'
```
niik: I didn't have time to think this through to completion when we chatted earlier but I think with this change we should be able to substitute the `require` in `code-mirror-host` with this and remove it from here.

No need to conditionally include on darwin since we're already going to import it on both platforms here.

niik (dismissed):
Looks great, just one potential cleanup

sergiou87 (commented):
It all makes sense to me! I also tested it and all went good, so :shipit: 

There is only thing I found while testing this PR, but it's not related. I added lots of co-authors to the textfield, and the app eventually crashed because `startScroll` in https://cs.github.com/codemirror/CodeMirror/blob/18c12c99b7ab8446eba7759889d8bda6b55bde4b/addon/hint/show-hint.js#L336 is `undefined`. Not sure why this happens, but also happens in `development`.

I'll just comment this since @niik already requested some changes in his review 😄

sergiou87 (approved):
✨

PR #13854, opened 2022-02-09
Cherry-pick: Show target branch is success banner when cherry-pick through context menu

### LINKED ISSUE #13849: Cherrypick success banner confusing from & to
### Describe the bug
After successful cherry pick, the banner shows the wrong branch.

### Version & OS

x64 2.9.6 (OSX 10.14.6)

1. select branch A,
2. select commits to cherry pick and choose branch B,
3. after successful cherry pick notice the branch info in banner shown:
4. see the mention of source branch A, not the target branch B in the banner text.

(see screen shots)

### Expected behavior

`Successfully copied {countCherryPicked} to {targetBranchName}` telling me the target branch, not the source branch.

### Actual behavior

The source branch A is displayed instead of the target branch B as the wording and code would imply.

### Screenshots

Screenshot 2022-02-09 at 00 48 24
_(select from pub/•, apply to rel/•…)_

Screenshot 2022-02-09 at 00 48 47
_(applied to rel/•…)_

Screenshot 2022-02-09 at 00 49 27
_(banner showing the source branch pub/• as "to", even though applied to current rel/•…)_

### Logs

[2022-02-08-last.desktop.production.log](https://github.com/desktop/desktop/files/8028279/2022-02-08-last.desktop.production.log)

### Additional context

I originally thought it was a problem in successful-cherry-pick.tsx mixing up to/from, but the actual naming of the variable sounds alright, it must be the source of its value that mixes up things.

tidy-dev:
Closes #13849

## Description
I think this may have been a regression from a different bug fix. :/ But, when initializing the target branch for a context menu, it was setting the source and target to the current branch. Target should be null at this point, because the user has not yet selected a target branch. Then, upon selecting the target branch the cherry pick state needs to be updated to reflect that.

### Screenshots
https://user-images.githubusercontent.com/75402236/153201748-af7a9675-27e9-4936-855d-aa15f0def677.mov

## Release notes
Notes: [Fixed] Cherry-pick success message always correctly reflects target branch.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -2874,7 +2874,7 @@ export class App extends React.Component<IAppProps, IAppState> {
         branchCreated: false,
         commits,
       },
-      tip.branch,
+      null,
       commits,
       tip.branch.tip.sha
     )
--- a/app/src/ui/dispatcher/dispatcher.ts
+++ b/app/src/ui/dispatcher/dispatcher.ts
@@ -3192,6 +3192,19 @@ export class Dispatcher {
     return this.appStore._setMultiCommitOperationStep(repository, step)
   }
 
+  /** Set the multi commit operation target branch */
+  public setMultiCommitOperationTargetBranch(
+    repository: Repository,
+    targetBranch: Branch
+  ): void {
+    this.repositoryStateManager.updateMultiCommitOperationState(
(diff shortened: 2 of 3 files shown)


sergiou87 (approved):
Makes sense and works as expected!

PR #13856, opened 2022-02-09
Select Integration Tab Automatically when Changing Default Editor 

uttiya10:
## Description

image
Currently, selecting `Options` to change your default editor will open the options menu but wouldn't select the Integrations tab automatically, which is what most users would want when changing their editor.

This PR will have the options menu open with the Integrations Tab.

## Release notes

Notes: no-notes


sergiou87 (approved):
I love this, thank you so much for your contribution! ❤️

PR #13866, opened 2022-02-11
Load `desktop-notifications` module lazily

sergiou87:
## Description

This PR bumps `desktop-notification` to v0.1.7, which loads the native binary of the module lazily the first time it's used: https://github.com/desktop/desktop-notifications/pull/7

This will prevent seeing crashes when the app starts, which are harder to track down.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨  Tested on windows and notifications work as expected. :)

PR #13868, opened 2022-02-11
Implement retrieving Actions Workflow Runs via check suite id

tidy-dev:
## Description

Currently, for retrieving GitHub Action info/steps about check runs for a PR, we use an Action's api query to get all the Action Workflow Runs for a repo filtered on the PR branch name. This works for most cases. But, we found in the case of a release branch that might look like `releases/some.version` and when we release that branch, an auto generated `releases_random number stuff` is made that has the check runs that actually publish the release. Thus, in that case, we would not get information about the release check runs. This will result in these check runs being under a group header `Other` and not having the Action Workflow steps displayed. 

Other code maintenance pitfalls about this approach is that when we query with branch name and the PR Action workflows have been rerun many times for multiple commits, this will retrieve all of them and thus we have to client side filter down to the latest one before then matching the Action workflow runs to the check runs based on their `check_suite_id`. 

Fortunately, the Actions team helped us out and [added an additional query parameter](https://github.com/github/github/pull/205075) to the same Action's api query to filter by the `check_suite_id` that we ultimately match the check runs by. Thank you @soelinn!

This PR switches our approach to using the new `check_suite_id` query parameter. However, this new api query will not be on GHE until next version; thus, I added an GHE endpoint capability flag and I have kept the original approach for those users 

### Screenshots

Screen Shot 2022-02-11 at 9 35 39 AM

## Release notes
Notes: [Improved] The check runs list for pull requests with multiple branches displays all actions workflow steps and headers.

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1066,7 +1066,7 @@ export class API {
    * List workflow runs for a repository filtered by branch and event type of
    * pull_request
    */
-  public async fetchPRWorkflowRuns(
+  public async fetchPRWorkflowRunsByBranchName(
     owner: string,
     name: string,
     branchName: string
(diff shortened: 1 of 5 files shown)


sergiou87 (approved):
Looks good!! We'll enable this for GHES and GHAE once we're sure about the version in which the new endpoint is available. :shipit:

PR #13870, opened 2022-02-12
Get rid of dispatcher in CommitMessage component

### LINKED ISSUE #13850: Co-Author stealing focus from empty commit message
### Describe the bug
Trying to type into commit message box while adding co-authors steals focus after every single key press.

### Version & OS

x64 2.9.6 (OSX 10.14.6)

1. select commits to squash (might be present in other use cases, this is my repro case)
2. empty the commit message (summary)
3. try to add co-authors
4. then try to type in the commit message, and see how co-author input steals the focus

### Expected behavior

The active input should stay active.

### Actual behavior

After every single character, the focus jumped back to co-author input.

### Screenshots

https://user-images.githubusercontent.com/1784648/153098007-5d3ad6fb-9764-4453-a017-704014f59e5b.mov

_(video demonstrating trying to type "fooo" into the summary)_

### Logs

_(no log created for today 🤷‍♂️ …)_

niik:
Closes #13850

## Description

Prior to us adding squashing to Desktop the `CommitMessage` component was only used in one spot. Given that we could get away with using the dispatcher directly from the component. The downside of this being that not all aspects of the CommitMessage component makes sense when embedded in a dialog.

Most notably we don't want to persist the commit message and any selected co-authors when used from a Dialog. Currently this is avoided by the use of a `persistCommitMessage` and `persistCoAuthors` prop which causes the component to persist in its own state.

With this change we're instead making the component behave more like a typical Rect component by lifting storage of commit message and co authors to the parent components and by removing all direct dispatcher calls in favor of prop callbacks.

## Release notes

Notes:

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -127,6 +127,7 @@ interface ICommitMessageProps {
 
   readonly onShowPopup: (popup: Popup) => void
   readonly onShowFoldout: (foldout: Foldout) => void
+  readonly onCommitSpellcheckEnabledChanged: (enabled: boolean) => void
 }
 
 interface ICommitMessageState {
@@ -497,8 +498,7 @@ export class CommitMessage extends React.Component<
       : 'Disable commit spellcheck'
     return {
       label: isEnabled ? disableLabel : enableLabel,
-      action: () =>
-        this.props.dispatcher.setCommitSpellcheckEnabled(!isEnabled),
+      action: () => this.props.onCommitSpellcheckEnabledChanged(!isEnabled),
     }
   }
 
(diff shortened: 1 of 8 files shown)


app/src/ui/changes/commit-message.tsx
```diff
@@ -163,21 +177,13 @@ export class CommitMessage extends React.Component<
...
-      return
-    }
-    // We're unmounting, likely due to the user switching to the history tab.
-    // Let's persist our commit message in the dispatcher.
-    this.props.dispatcher.setCommitMessage(this.props.repository, {
-      summary: this.state.summary,
-      description: this.state.description,
-    })
+    const { props, state } = this
+    props.onPersistCommitMessage?.(pick(state, 'summary', 'description'))
```
sergiou87, line 186: Nice!

sergiou87 (approved):
Code LGTM, and I think I've tested every scenario and didn't find issues.

Maybe it'd be good if @tidy-dev tested it too, since she did the original refactor and fixed a few regressions after that, so she knows better what could break and what not!

In any case I'll approve it if you want to go ahead and merge it anyway 😄

tidy-dev (approved):
✨   I tested out the regressions we had before about the commit author persisting incorrectly; all good there.

PR #13871, opened 2022-02-12
Escape Special Characters When Adding a File to .gitignore

### LINKED ISSUE #10253: Fix ignoring file names with [Brackets] in dir path
When right clicking a file in the gui and selecting ignore file github desktop won't use escape characters to escape special characters like brackets([])

### Proposed solution
Just a check for special characters in the file dir and use the escape char(\) in conjunction?

So instead of writing 
`resources/[3dtext]/log.txt`
it should write
`resources/\[3dtext]/log.txt`
into the .gitignore file.

uttiya10:
Closes #10253

## Description

Currently, Github Desktop will not escape special characters when adding a file to `.gitignore`, causing certain files to not be ignored.
### Example:
File to Ignore:
`resources/[3dtext]/log.txt`

Expected Entry in .gitignore file:
`resources/\[3dtext\]/log.txt`

Actual Entry in .gitignore file:
`resources/[3dtext]/log.txt`

### Implementation:
The PR is based on a Python Library called `pathspec`. [Reference Here](https://github.com/cpburnz/python-path-specification/blob/7ca27e2300ced6364a3816820eb2532c89c7d01c/pathspec/patterns/gitwildmatch.py#L338)

## Release notes

Notes:
- [FIXED] File Paths' with special characters are now properly escaped when added to .gitignore


app/test/unit/git/gitignore-test.ts
```diff
@@ -129,6 +131,14 @@ describe('gitignore', () => {
 
       expect(files).toHaveLength(0)
     })
+
+    it('escapes string with special git characters', async () => {
+      const unescapedFilePath = '[never]\\!gonna*give#you?_.up'
+      const escapedFilePath = '\\[never\\]\\\\!gonna\\*give\\#you\\?_.up'
```
sergiou87, line 137: 🕺 😂

tsconfig.json
```diff
@@ -15,7 +15,14 @@
...
     "strict": true,
     "outDir": "./out",
-    "lib": ["ES2017", "DOM", "DOM.Iterable", "ES2018.Promise", "ES2020.string"],
+    "lib": [
+      "ES2017",
+      "DOM",
+      "DOM.Iterable",
+      "ES2018.Promise",
+      "ES2020.string",
+      "ES2021.String"
```
sergiou87: Why is this necessary?

sergiou87 (commented):
Looks great overall, thank you!! Just one question and then I will test it before merging it 😄

tsconfig.json
uttiya10: This is to access the new `.replaceAll` method for Strings which I used [here](https://github.com/desktop/desktop/pull/13871/files#diff-dd3aa4a2c0644b9dc56b14f01db04cf75f4d7266e96fbdd34a168834f34bcf90R113-R115)

sergiou87: Hopefully not needed after #14077 is merged 🎉

uttiya10: Resolving since libraries are not used anymore! 🎉

sergiou87 (approved):
I'm sorry it took me a while to get back to this. I just tested it and works as expected, thank you! ❤️

PR #13885, opened 2022-02-15
Fix duplicated first character on Windows when dealing with submodules or repos-within-repos

### LINKED ISSUE #12314: Commit bug, duplicating the first letter
### Describe the bug

I have had problems with Github Desktop when 'commit', whenever I create a new project, the first letter is duplicated, and this bug happens and I can't commit, I've seen some tutorials on 'commit' in the terminal to try to solve this problem, but it didn't work.

### Version & OS

Version 2.8.2 (x64) - Windows 10 Version 21H1

On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	react-native/chapter-2/

nothing added to commit but untracked files present (use "git add" to track)

### Expected behavior

I expected to click 'commit to main' and everything would work, I didn't do anything for it to stop working, I just created a new folder with expo init gofinances and this error started to appear, but when I went to create a new project, this error also appeared .

### Actual behavior

I have had problems with Github Desktop when 'commit', whenever I create a new project, the first letter is duplicated, and this bug happens and I can't commit, I've seen some tutorials on 'commit' in the terminal to try to solve this problem, but it didn't work.

### Screenshots

image

### Logs

[2021-05-29.desktop.production.log](https://github.com/desktop/desktop/files/6563987/2021-05-29.desktop.production.log)

### Additional context

I don't know if it's a bug, or if I'm doing something wrong, but there is still a bug where it is duplicating the first letter of the project.

niik:
Closes #12314

## Description

When running `git status` on a repository with untracked submodules (or folders containing `.git` folders, i.e. repositories within repositories) Git will report the path with a trailing slash (i.e. `repo-a/` instead of `repo-a`). This caused our path text component to mistakenly treat the first character of the path as the directory name

https://github.com/desktop/desktop/blob/ab48d526da6feb6a47fa9a4f4a77eaaf9293cc35/app/src/ui/lib/path-text.tsx#L151-L155

There was an attempt to solve this in https://github.com/desktop/desktop/pull/785 but unfortunately that PR didn't account for the fact that Windows uses `\` as its path separator character and not `/`.

This PR updates the code to be platform agnostic by leveraging `Path.sep`.

Note that users will still not be able to commit these repositories within repositories but that's a separate issue concerning our support for submodules and how best to handle repositories within repositories.

The fix for this is contained in 4f33cd66ea2ec312e0bf74f8cfe8b5a5aea8eefc but I also took the opportunity to swap out our uses of `.substr` (which is now deprecated) in favor of `.substring`

### Screenshots

image

## Release notes

Notes: [Fixed] Repositories containing untracked submodules no longer display an duplicated first character on Windows

### DIFF
--- a/app/src/ui/lib/path-text.tsx
+++ b/app/src/ui/lib/path-text.tsx
@@ -141,9 +141,9 @@ export function truncatePath(path: string, length: number) {
 export function extract(
   normalizedPath: string
 ): { normalizedFileName: string; normalizedDirectory: string } {
-  // for directories the status entry is returned as a path with a trailing
-  // which causes the directory to be trimmed in a weird way below. let's try
-  // and resolve this here
+  // for untracked submodules the status entry is returned as a path with a
+  // trailing path separator which causes the directory to be trimmed in a weird
+  // way below. let's try to resolve this here
   normalizedPath = normalizedPath.endsWith(Path.sep)
     ? normalizedPath.substring(0, normalizedPath.length - 1)
     : normalizedPath
--- a/app/src/ui/lib/path-text.tsx
+++ b/app/src/ui/lib/path-text.tsx
@@ -86,8 +86,8 @@ export function truncateMid(value: string, length: number) {
   }
 
   const mid = (length - 1) / 2
-  const pre = value.substr(0, Math.floor(mid))
-  const post = value.substr(value.length - Math.ceil(mid))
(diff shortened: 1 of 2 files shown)


sergiou87 (approved):
:shipit:

PR #13752, opened 2022-01-27
Refactor `remote.app.moveToApplicationFolder` to main process

tidy-dev:
## Description

This refactors the `remote` module use of `remote.app.moveToApplicationFolder` method to use IPC messaging to the main process instead.

Impacts: on mac os, if app is not in applications folder, you are presented with a dialog and when you click "move and restart", this event should fire to move and restart app.

Note: This is one of several PR's to refactor our usage of the `remote` module to instead use the IPC messaging with the main process so that we can remove our [undesired](https://nornagon.medium.com/electrons-remote-module-considered-harmful-70d69500f31) `remote` module dependency (supporting ultimate goal of upgrading to electron 16).

## Release notes
Notes: no-notes

PR #13886, opened 2022-02-15
Await moveToApplicationsFolder invocation on main process

sergiou87:
## Description

This PR fixes a regression introduced in https://github.com/desktop/desktop/pull/13752, where moving the invocation of `app.moveToApplicationsFolder` to the main process prevented the renderer process from catching exceptions.
It required to change the proxy `moveToApplicationsFolder` function from a "send" type to an "invoke" type, and then make sure the `Promise` is bubbled up and awaited from the `MoveToApplicationsFolder` dialog.

## Release notes

Notes: [Fixed] Fix crash when attempting to move the app to the /Applications folder on macOS


niik (approved):
Looks great!

tidy-dev (approved):
✨

PR #13887, opened 2022-02-15
Legacy stealth email attribution

### LINKED ISSUE #13879: Misattribution notice more tolerant for GH private emails

The warning triangle for author misattribution notice is present even though my git config email is fine, and has always been. It's just in the old format before the `id+…` addition. I'd prefer not to change the committer info in my git settings as I've always had this one and would prefer to keep it … and at the same time get rid of the warning triangle that covers my avatar.

### Proposed solution

Please consider silently tolerating `example@users.noreply.github.com` in git config even for `12345+example@users.noreply.github.com` account as that might have been "the correct" form back in the day.

### Additional context

Screenshot 2022-02-14 at 18 45 21

niik:
Closes #13879

## Description

When considering whether an commit made by a certain email address would be attributed to the user we only looked at the list of email addresses returned from the API whereas GitHub.com (and GHES) will attribute correctly for both the legacy and modern stealth email address regardless of whether the user has ever turned on the private email option.

## Release notes

Notes: [Fixed] Treat the old and new format of private email address equally when showing commit attribution warning


app/src/lib/email.ts
```diff
@@ -99,27 +99,26 @@ export function getStealthEmailForUser(
...
+export const isAttributableEmailFor = (account: Account, email: string) => {
+  const { id, login, endpoint, emails } = account
+  const needle = email.toLowerCase()
 
-  return [...uniqueEmails]
+  return (
+    emails.some(({ email }) => email.toLowerCase() === needle) ||
+    getStealthEmailForUser(id, login, endpoint).toLowerCase() === needle ||
+    getLegacyStealthEmailForUser(login, endpoint).toLowerCase() === needle
+  )
```
sergiou87, line 123: TIL about this whole file 🤯

sergiou87 (approved):
This is great! I just learned about all those nuances with different email addresses and the whole existence of `app/src/lib/email.ts` 🤦 

✨  :shipit: ✨

PR #13894, opened 2022-02-15
Bump Webpack to 5.68.0

sergiou87:
## Description
Our build stack has fallen behind with an old Webpack 4 version, and we had trouble trying to upgrade to TS 4.5 and using some other modules. This should bring us to 2022 and make things a bit easier from now on if we need to integrate with new modules.

Only one relevant highlight about these changes: the plugin `CleanWebpackPlugin` was removed. It was supposed to delete the `out` folder _after_ each build, but it wouldn't do so. I even found some `askpass.js` files from July 2021 🤣  The replacement is now [Webpack's `output.clean`](https://webpack.js.org/configuration/output/#outputclean), but that one would delete the contents of `out` before each module is built (main, renderer, crash, highlighter…), meaning they would delete each other files resulting in a build with random files missing.

If we really want to delete the `out` folder, we should find a proper way of doing so, but since this PR doesn't regress the existing behavior, we can tackle that in a different changeset.

Other than that, we should probably upgrade these libraries (I skipped them since I got Webpack v5 working with just this) that still rely on old versions of Webpack:
> warning " > css-loader@2.1.0" has incorrect peer dependency "webpack@^4.0.0".
> warning " > file-loader@2.0.0" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".
> warning " > parallel-webpack@2.3.0" has incorrect peer dependency "webpack@^1.12.9 || ^2.2.0 || ^3.x || ^4.x".
> warning "style-loader > schema-utils@0.4.5" has incorrect peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".

## Release notes

Notes: no-notes


app/webpack.development.ts
```diff
@@ -58,6 +58,9 @@ const rendererConfig = merge({}, common.renderer, config, {
       },
     ],
   },
+  infrastructureLogging: {
+    level: 'error',
+  },
```
sergiou87, line 63: This used to be the `logLevel` property from `devMiddleware` (https://github.com/desktop/desktop/pull/13894/files#diff-404124522cf8054ccbb64a424ba4e5de30e9c059c05b8cfaab3ad4339633f6c9L55)

script/start.ts
```diff
@@ -52,7 +52,6 @@ if (process.env.NODE_ENV === 'production') {
         message,
         u(message, u(message, rendererConfig).output).publicPath
       ),
-      logLevel: 'error',
```
sergiou87, line 55: This has been moved to `infrastructureLogging.level` in [webpack.development.ts](https://github.com/desktop/desktop/pull/13894/files#diff-bcf75b74c707fb68169331415b4cce71a5a14e41c81834ed8f8104df4467fce1R61-R63)

tidy-dev (approved):
✨  The changes looks like they parallel and I pulled it down and it builds. Looks great! - especially for these awesome confidence oozing commits. 😸 
image

sergiou87:
Me working on those commits:
image

Issue #13888, opened 2022-02-15
Application hangs when trying to discard changes


palhal:
### Describe the bug
When trying to discard changes, the application becomes unresponsive.

### Version & OS
2.9.7 (x64)
Windows 10 21H2 (19044.1526)

### Steps to reproduce the behavior
1. Change a file in any repository
2. Open the Changes tab
3. Right click on the applicable file
4. Click _Discard changes_ in the context menu
5. Click _Discard changes_ in the confirmation dialog

### Expected behavior
* The changes are discarded

### Actual behavior
* A spinner icon appears in the dialog title bar
* The whole application becomes unresponsive
* Nothing happens when clicking Discard changes or Cancel
* The changes are not discarded

### Screenshots
![github-discard](https://user-images.githubusercontent.com/56862114/154061580-7030fd63-ce1c-4ac3-a40f-061ca6e6505f.png)

### Logs
Not much relevant content in the logs. This is the only line written:
```log
2022-02-15T11:32:53.506Z - info: [ui] No submodules found. Skipping "git submodule status"
```

### Additional context
* I have waited about 10 minutes for the operation to finish
* The only solution is to terminate the application from Task Manager
* I have checked that there are no other processes using the file
* I have tried to discard multiple file types (.js, .html, .json)
* I have tried multiple repositories. All repositories are non-GitHub.
* I have tried deleting _%AppData%\GitHub Desktop_ to start "fresh" 
* Downgrading to 2.9.6 fixes the problem
* Not directly related, but the application seems slower to start in 2.9.7

I was unsure whether I should open a new issue or comment on #13433 or #13661, but those appear closed by #13436.

tidy-dev:
Thanks for the report! Curious if it does discard the changes and just that the application hangs?

tidy-dev:
I know there were not errors, but would you mind uploading the log anyways? - We might be able to at least see app progression to see if it is not reaching something we expect. :)

palhal:
Thanks for the prompt response. No, the changes are not discarded. That's why I thought there might be a file-already-in-use problem, but I'm quite certain that's not the case.

I didn't think older logs were relevant and unfortunately I deleted the whole logs folder when the problem started. I have the logs however for when I downgraded to 2.9.6 and it was working until I upgraded again. I also included a log for a "fresh" start with the _%AppData%\GitHub Desktop_ folder removed. I'm not sure if any of that helps.

[...]

PR #13899, opened 2022-02-15
Discarding changes displays error

tidy-dev:
Partially addresses #13888

## Description

An issue reported in 2.9.7 which is a result of the Electron upgrade is that discarding changes may now fail if the user has their recycling bin set to "Don't move files to the Recycle Bin. Remove files immediately when deleted". It is a bad failure case in that the application just hangs and must be force quitted. This PR captures the error thrown my the electron `shell.trashItem()` method and informs the user of the failure.

I marked this as partially addressing #13888. This is because in prior versions of Desktop on an Electron version prior to 13, the trashItem method did not error and simply did not move the files to the Recycle Bin (silently failing). This a bug because we expected the files to be moved to the recycle bin and we inform users of this on the discard pop up. However, to only display error here feels like a regression as we are preventing users from discarding changes when they used to be able to. 

### Screenshots
image

## Release notes

Notes: [Fixed] App does not hang when discarding changes in some scenarios.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4360,7 +4360,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
       this.emitError(
         new Error(
-          `Failed to discard changes to ${TrashNameLabel}.\n\nA common reason for this is that the directory or one of its files is open in another program or the ${TrashNameLabel} is set to delete items immediately.`
+          `Failed to discard changes to ${TrashNameLabel}.\n\nA common reason for this is that the ${TrashNameLabel} is set to delete items immediately.`
         )
       )
     }
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -4360,7 +4360,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
       this.emitError(
         new Error(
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -4352,7 +4352,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
-    await gitStore.discardChanges(files)
+
+    try {
+      await gitStore.discardChanges(files)
+    } catch (error) {
+      log.error('Failed discarding changes', error)
+
+      this.emitError(
+        new Error(
+          `Failed to discard changes to ${TrashNameLabel}.\n\nA common reason for this is that the directory or one of its files is open in another program or the ${TrashNameLabel} is set to delete items immediately.`
```
tidy-dev: Notes.. I am not sure if the "A common reason for this is that the directory or one of its files is open in another program" applies here as I think that is a problem more so for directory level deletions..

tidy-dev:
```suggestion
          `Failed to discard changes to ${TrashNameLabel}.\n\nA common reason for this is that the ${TrashNameLabel} is set to delete items immediately.`
```

tidy-dev: Nvm.. I removed it as I think this only applies to directories..

sergiou87 (approved):
TIL about that setting 🤯 

Looks good to me!

PR #13904, opened 2022-02-16
Discard changes with retry on error moving to recycling bin

### LINKED ISSUE #13888: Application hangs when trying to discard changes

tidy-dev:
Closes #13888

## Description
A possible solution to #13888 see #13899 for more details. This allows a user to retry discarding changes if it fails due to inability to move items to the recycling bin. Prior to 2.9.7, Desktop silently does this making those changes unknowingly unrecoverable. After Electron 16 upgrade, we must at least catch the error (or else bug seen in #13888). Another solution would be to just silent capture it, but this is misleading since our warning popup says you can recover changes in the recycling bin. 

### Screenshots

https://user-images.githubusercontent.com/75402236/154281170-9db4eee9-eee3-43a6-85dc-ef3ca84d9084.mov

## Release notes
Notes: [Fixed] App does not hang when discarding changes in some scenarios.


app/src/lib/stores/git-store.ts
```diff
@@ -1459,13 +1460,21 @@ export class GitStore extends BaseStore {
...
         // as many animation frames as it needs to.
-        await this.shell.moveItemToTrash(
-          Path.resolve(this.repository.path, file.path)
-        )
+        try {
+          await this.shell.moveItemToTrash(
+            Path.resolve(this.repository.path, file.path)
+          )
+        } catch (e) {
+          throw new ErrorWithMetadata(e, {})
```
tidy-dev: Using  `new ErrorWithMetadata(e, {})` as hacky way to distinguish an error thrown by the `moveItemToTrash` method somewhere else in this function...  needs more thought into this approach.... Either moving the `moveItemToTrash` up to the app-store (but then it's not in that queue) and handling the error there.. or some sort of direct return type indicating this specific failure.

app/src/ui/app-error.tsx
```diff
@@ -239,6 +248,28 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
     )
   }
 
+  private renderRetryFooter(error: Error) {
```
tidy-dev: Was sort of surprised we didn't already have a means to have a retry-able error without making a custom error dialog. Seems like the retry clone stuff here could have been more generic.

app/src/lib/stores/git-store.ts
tidy-dev: (Ignore from first stab)

app/src/ui/app-error.tsx
tidy-dev: (ignore from first stab)

app/src/ui/discard-changes/discard-changes-retry-dialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  private renderFooter() {
+    return (
+      <DialogFooter>
+        <OkCancelButtonGroup
+          okButtonText={
+            __DARWIN__
+              ? 'Retry Discard Changes and Continue'
+              : 'Retry discard changes and continue'
```
sergiou87: I think I prefer more explicit wording in these actions, something like: 'Discard Changes Forever' or 'Discard Changes Anyway' and 'Cancel'  instead of 'Close'.

app/src/ui/discard-changes/discard-changes-retry-dialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+          }
+          okButtonTitle={`This will discard changes and they will be unrecoverable.`}
+          cancelButtonText="Close"
```
sergiou87: We should add a `destructive={true}` attribute here probably 🤔 
```suggestion
          cancelButtonText="Close"
          destructive={true}
```

sergiou87 (changes requested):
As mentioned in my comment, I'd rather have a more explicit wording in the buttons (especially on the destructive one). Also, it seems like users with these Recycle Bin settings will always get this new popup, right? Shouldn't we add a "Don't bother me again" kind of checkbox for it?

palhal:
Sorry to interfere, but did anyone complain about the lack of warning in 2.9.6? I understand that the error was silently ignored, but wasn't that a good thing? If I choose to disable the recycle bin, it's because I don't care about recovery. I just want the changes gone, without additional interactions.

Another approach could be to rephrase the initial text with something like "Changes can _usually_ be restored by ..." or "Unless the recycle bin is disabled, changes can be restored by ...". (I know that none of them are optimal.)

Anyway thanks for a very quick follow up on this!

tidy-dev:
@palhal Thank you for the suggestions! We are try our best to help users not step into painful situations easily. Here that would be accidentally permanently discarding a change. Thus, we want to make sure the user is aware of what going forwards means. That said, I am sure this popup is annoying for users who have intentionally set their system settings this way. To help with that, I have just added the "Don't show this again" check box so users can choose not to receive this warning again.

app/src/ui/discard-changes/discard-changes-retry-dialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+          okButtonText={
+            __DARWIN__
+              ? 'Permanently Discard Changes and Continue'
+              : 'Permanently discard changes and continue'
+          }
+          okButtonTitle={`This will discard changes and they will be unrecoverable.`}
+          cancelButtonText="Cancel"
+          destructive={true}
+        />
+      </DialogFooter>
```
sergiou87: I find that button too verbose 😅  Can we at least remove the `and continue` part?
image

sergiou87 (commented):
Code looks great, I tested it and worked as expected! Only have one suggestion/comment about the retry/confirmation dialog, which doesn't look great with that button title, IMHO 😅

app/src/ui/discard-changes/discard-changes-retry-dialog.tsx
tidy-dev: done!

palhal:
@tidy-dev I get your point. I thought I could save you and the code reviewers some time 😄

By the way, I think the message _These changes will be unrecoverable from the {TrashNameLabel}_ is a bit verbose. If the changes are unrecoverable, I guess it doesn't matter from where?

sergiou87 (approved):
:shipit:

PR #13914, opened 2022-02-16
Tokenize unicode emoji

### LINKED ISSUE #13911: Emojis appear as white icon.
### Describe the bug
Emojis appear that in commit as white icon.

### Version & OS

Version 2.9.7 (x64)
Windows 11

1. Go to 'History Tab'

### Expected behavior

Emojis appear that in commit normally like previous version.

### Actual behavior

Emojis appear that in commit as white icon in current version.

### Screenshots

image

### Logs

-

### Additional context

-

niik:
Closes #13911 

## Description

This works around the Chromium bug described in https://github.com/desktop/desktop/issues/13911#issuecomment-1041360884 by wrapping all unicode emoji sequences in a `span` with the `unicode-emoji` class on it. That way we can select all unicode emoji ranges in CSS and explicitly set their font weight.

## Release notes

Notes: [Fixed] Unicode emoji on Windows no longer render as monochrome outlines

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -33,6 +33,7 @@
     "double-ended-queue": "^2.1.0-0",
     "dugite": "^1.104.0",
     "electron-window-state": "^5.0.3",
+    "emoji-regex": "^10.0.0",
     "event-kit": "^2.0.0",
     "file-metadata": "^1.0.0",
     "file-uri-to-path": "^2.0.0",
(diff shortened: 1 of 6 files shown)


sergiou87 (approved):
Looks great!! However, do you think is worth having these changes knowing the next Electron release fixes the issue? If you do, do you think we should revert these after we upgrade? 🤔

niik:
> Looks great!! However, do you think is worth having these changes knowing the next Electron release fixes the issue? If you do, do you think we should revert these after we upgrade? 🤔

Yeah, given that we don't intend to upgrade to Electron 17 right now we definitely want a workaround, it looks pretty bad as-is. And yeah, I think we can just revert this once we go to 17.

tidy-dev (approved):

PR #13916, opened 2022-02-16
Handle trampoline data weirdness more gracefully

sergiou87:
## Description

We've seen many app crashes due to connections to the trampoline that didn't have a command identifier (which is required). Instead of crashing the app, this PR will just log the error, send it to Sentry as a non-fatal error with additional debug info (which env variables -but not their values- were passed to the trampoline), and then just close the socket and do nothing.

## Release notes

Notes: [Fixed] Fix random crash under random circumstances??? 🤣  I seriously have no idea how this could happen.


tidy-dev (approved):
✨  Makes sense and hopefully will give a more info around this random crashes.

niik (approved):

PR #13917, opened 2022-02-16
Optimize validating repository when adding existing repository

### LINKED ISSUE #13901: Adding/Opening an Existing Local Repo from Disk Results in Folder Path Entry Issues (lost keystrokes)
### Describe the bug
When opening a local repository via CTRL+O and then typing in a path, characters appear, and then disappear as though either being erased, or the keystroke undo buffer is being triggered (as it sometimes affects more than 1 character).

### Version & OS

Version 2.9.7 (x64) on Windows 10

1. Open GitHub Desktop
2. CTRL + O to Add a local repository...
3. Begin typing the path in the Local path text field provided
4. Usually starts about the time the first path separator (backslash) is entered (e.g. c:\) -- the character will be displayed, then removed and possibly redisplayed (if typing is paused). 

### Expected behavior

When entering text for the path, don't alter the input.

If checking the path validity to highlight the Add repository button, ensure user input is not discarded.

### Actual behavior

When entering the path, it is necessary to pause regularly to figure out what the textbox is going to do. Sometimes it will swallow one or more characters, but then redisplay them. Other times, it will swallow and not redisplay the characters, forcing re-entry.

### Screenshots

add-local-repository-dialog

### Logs

No logs seem to be applicable. Opening the Developer Tools to watch the console during the process did not show anything at all (good or bad) until the repository had been selected, and was opened.

### Additional context

I am not using GitHub for these repos (or this install of GitHub Desktop). It is currently only operating on local repositories hosted on an internal service.

EDIT: fixed backslash disappearing from issue report (replaced with word)

niik:
Closes #13901

## Description

This takes care of the race condition in add-existing-repository where we would update the state with the new path only after having received the response from `git rev-parse`.

This also optimizes the validation process by checking if the path exists before spawning Git (significantly faster on Windows) and by leveraging `rev-parse` to both check if the repository exists _and_ if it's a bare repository instead of doing it twice.

Finally it reuses the logic to update and validate the path which was previously duplicated between the different methods of updating path (component mount, file picked, and manual input)

## Release notes

Notes:

### DIFF
--- a/app/src/lib/directory-exists.ts
+++ b/app/src/lib/directory-exists.ts
@@ -4,7 +4,11 @@ import { stat } from 'fs/promises'
  * Helper method to stat a path and check both that it exists and that it's
  * a directory.
  */
-export const directoryExists = (path: string) =>
-  stat(path)
-    .catch(null)
-    .then(s => s?.isDirectory() ?? false)
+export const directoryExists = async (path: string) => {
+  try {
+    const s = await stat(path)
+    return s.isDirectory()
+  } catch (e) {
+    return false
+  }
+}
--- a/app/src/lib/directory-exists.ts
+++ b/app/src/lib/directory-exists.ts
@@ -1,0 +1,10 @@
+import { stat } from 'fs/promises'
+
+/**
+ * Helper method to stat a path and check both that it exists and that it's
+ * a directory.
+ */
+export const directoryExists = (path: string) =>
+  stat(path)
+    .catch(null)
+    .then(s => s?.isDirectory() ?? false)
--- a/app/src/ui/add-repository/add-existing-repository.tsx
+++ b/app/src/ui/add-repository/add-existing-repository.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react'
 import * as Path from 'path'
 import { Dispatcher } from '../dispatcher'
-import { isGitRepository } from '../../lib/git'
-import { isBareRepository } from '../../lib/git'
+import { getRepositoryType } from '../../lib/git'
 import { Button } from '../lib/button'
 import { TextBox } from '../lib/text-box'
 import { Row } from '../lib/row'
@@ -15,7 +14,6 @@ import { OkCancelButtonGroup } from '../dialog/ok-cancel-button-group'
 
 import untildify from 'untildify'
 import { showOpenDialog } from '../main-process-proxy'
-import { stat } from 'fs-extra'
 
 interface IAddExistingRepositoryProps {
   readonly dispatcher: Dispatcher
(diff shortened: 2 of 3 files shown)


app/src/ui/add-repository/add-existing-repository.tsx
```diff
@@ -165,9 +175,10 @@ export class AddExistingRepository extends React.Component<
   }
 
   private onPathChanged = async (path: string) => {
-    const isRepository = await isGitRepository(path)
-
-    this.setState({ path, isRepository })
+    if (this.state.path !== path) {
+      this.setState({ path, isRepository: false })
+      this.validatePath(path)
```
tidy-dev:
```suggestion
      this.updatePath(path)
```

tidy-dev:
Seems an unintended error thrown/not caught

On windows vm tho not sure it matters.

Steps:
1. Hit Choose.. 
2. Select a valid repository folder
3. remove a character from the folder name in `Local path` input. 
4. Receive console error. (tho does correctly disable `Add Repository` button)

image


Also just tried to manually enter a directory:
Got all the uncaught ENOENT: errors on keypress.
image

app/src/lib/directory-exists.ts
```diff
@@ -0,0 +1,10 @@
+import { stat } from 'fs/promises'
+
+/**
+ * Helper method to stat a path and check both that it exists and that it's
+ * a directory.
+ */
+export const directoryExists = (path: string) =>
+  stat(path)
+    .catch(null)
```
sergiou87: I think this should be a function, right? 🤔 
```suggestion
    .catch(() => null)
```

sergiou87: I see `catch` accepts `null` as a parameter, and the compiler didn't complain with your `s?` below, even though `s` is never `null` or `undefined` in your implementation 🤯

sergiou87 (changes requested):

sergiou87:
I'm also testing it adding/removing characters multiple times from a valid path, and at some point it stops showing/hiding the warning as one would expect 🤔

sergiou87:
I have no idea why exactly, but only replacing `directoryExists` with the snippet below (which uses `await` and `try…catch` instead of `catch…then`) it works consistently well O_o

```typescript
export const directoryExists = async (path: string) => {
  try {
    const s = await stat(path)
    return s.isDirectory()
  } catch (e) {
    return false
  }
}
```

niik:
> I have no idea why exactly, but only replacing `directoryExists` with the snippet below (which uses `await` and `try…catch` instead of `catch…then`) it works consistently well O_o

Was that before or after you tried the `.catch(() => null)`? Either way I'm happy doing it the explicit async/await way too.

sergiou87:
@niik after `() => null`, I stopped seeing the uncaught exceptions @tidy-dev mentioned, but `directoryExists` sometimes never returned 🤯

app/src/ui/add-repository/add-existing-repository.tsx
sergiou87: @niik not super important but this small cleanup looks 👍  too, right?

sergiou87 (dismissed):
Works great!! Consider including https://github.com/desktop/desktop/pull/13917/files#r808039978 and then :shipit:

tidy-dev (dismissed):
✨  Working great now!

app/src/ui/add-repository/add-existing-repository.tsx
niik: Ah yup!

tidy-dev (approved):

sergiou87 (approved):
Still working great :shipit:

PR #13928, opened 2022-02-17
Fix "Create new repository" using the wrong path

### LINKED ISSUE #13909: New local repository has wrong local path
### Describe the bug
When adding a new local repository from a local path the `Local Path` changes to something else entirely.
No matter what folder I choose it always changes the `Local Path` to the same (different) folder. The folder that's chosen doesn't even exist - some path and a folder with my GitHub username as the final folder.

### Version & OS

Version 2.9.7 (x64)
Windows 10 (x64).

1. `File`
2. `Add Local Repository`
3. `(choose)`
4. `create a repository`
5. Wrong `Local Path`

sergiou87:
Closes #13909

## Description

This PR fixes a subtle regression introduced in https://github.com/desktop/desktop/pull/13804

The problem is we were always picking the default app path even when the dialog was created with an initial path.

## Release notes

Notes: [Fixed] "Create New Repository" dialog preserves the path set from "Add Local Repository" dialog

### DIFF
--- a/app/src/ui/add-repository/create-repository.tsx
+++ b/app/src/ui/add-repository/create-repository.tsx
@@ -116,7 +116,10 @@ export class CreateRepository extends React.Component<
       isRepository: false,
       readMeExists: false,
     }
-    this.initializePath()
+
+    if (path === null) {
+      this.initializePath()
+    }
   }
 
   public async componentDidMount() {


DavidTs93:
Ok, I have no idea why, but right after it finished doing all the checks the problem disappeared... :expressionless:

tidy-dev (approved):
✨  Good catch!

Fcus88 (commented):
v

LunaF28:
I am having this issue again. Is there any current fix?

PR #13957, opened 2022-02-20
Add tooltip to show types of file changes in a commit

uttiya10:
## Description
Currently, the tooltip shown when hovering over the number of changed files just repeats the same information:
image

This PR changes the tooltip to show the types of files changes in the commit:
image

## Release notes

Notes:

### DIFF
--- a/app/src/ui/history/commit-summary.tsx
+++ b/app/src/ui/history/commit-summary.tsx
@@ -418,20 +418,29 @@ export class CommitSummary extends React.Component<
     const filesLongDescription = (
       <>
         {filesAdded > 0 ? (
-          <span className="files-added">
-            <Octicon symbol={OcticonSymbol.diffAdded} />
+          <span>
+            <Octicon
+              className="files-added-icon"
+              symbol={OcticonSymbol.diffAdded}
+            />
             {filesAdded} added
           </span>
         ) : null}
         {filesModified > 0 ? (
-          <span className="files-modified">
-            <Octicon symbol={OcticonSymbol.diffModified} />
+          <span>
+            <Octicon
+              className="files-modified-icon"
+              symbol={OcticonSymbol.diffModified}
+            />
             {filesModified} modified
           </span>
         ) : null}
         {filesRemoved > 0 ? (
-          <span className="files-deleted">
-            <Octicon symbol={OcticonSymbol.diffRemoved} />
+          <span>
+            <Octicon
+              className="files-deleted-icon"
+              symbol={OcticonSymbol.diffRemoved}
+            />
             {filesRemoved} deleted
           </span>
         ) : null}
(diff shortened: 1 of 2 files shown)


haykam821:
I think that this tooltip should include the icons for each status kind instead of just labels. If implemented, these icons would make the information more glanceable than a simple list.

uttiya10:
image
Are you thinking of something like this?

I would like to try making the text colored (green for added, yellow for modified, etc) but can't seem to get css styling to work inside the tooltip😞

tidy-dev (changes requested):
@uttiya10 I like the direction you are going with the icons and agree coloring would be desired. There is a property of `tooltipClassName` that you can add to TooltippedContent component. We have been adding css rules to the `_tooltips.scss`. See the `length-hint-tooltip` class for an example. 

Also, need to address what happens on an empty commit. I would suggest to not show a tooltip in this case -> send undefined for the `tooltip` property.
image

app/src/ui/history/commit-summary.tsx
```diff
@@ -401,6 +393,48 @@ export class CommitSummary extends React.Component<
...
+          break
+        case AppFileStatusKind.Deleted:
+          filesRemoved += 1
+          break
+      }
+    }
+
+    const filesLongDescription = (
+      <>
+        {filesAdded ? <li>{filesAdded} added</li> : null}
```
tidy-dev: We prefer explicit checks - (same applies to next two lines)

```suggestion
        {filesAdded > 0 ? <li>{filesAdded} added</li> : null}
```

uttiya10:
image
Updated the design to use icons and coloured text 🎉

tidy-dev:
@uttiya10 This is looking great! One small suggestion. I think it would be more consistent with other tooltips and the changes list if only the icons were colored. 

https://user-images.githubusercontent.com/75402236/155343259-9be2671b-f20f-4d49-a475-9d5cb06828d8.mov

.

ghost:
Como faço parar de receber esses emails? Não fiz cadastro nisso, e não aguento mais receber isso a casa 5 mim!!!!! Att, Kellen VidalEm 20 de fev. de 2022 20:47, haykam821 ***@***.***> escreveu:
I think that this tooltip should include the icons for each status kind instead of just labels. If implemented, these icons would make the information more glanceable than a simple list.

uttiya10:
On a completely unrelated note, there seems to be a bug where any errors occurred when trying to trigger a re-run of checks do not appear to the user:

https://user-images.githubusercontent.com/56562649/155421881-af4cb625-b396-410d-ae5c-6df9bdbcdaa7.mp4

tidy-dev (approved):

tidy-dev:
> On a completely unrelated note, there seems to be a bug where any errors occurred when trying to trigger a re-run of checks do not appear to the user

If you haven't already, would you mind opening an issue for that bug? Also if when you hit rerun, please include in the issue whether if it does trigger a re-run for any of the checks. :)

1766701652 (commented):
w

PR #13962, opened 2022-02-21
Update Winston and use our own, custom, console transport

niik:
## Description

We recently saw a _very_ large number of crashes being reported by a _very_ small number of users. The cause turned out too be an infinite loop where logging to Winston's `Console` transport caused an asynchronous unhandled exception which caused us to log that we had encountered an unhandled exception which caused the chain of events to start over again.

The cause of the crash stems from Winston's use of `process.stderr` and `process.stdout` (or console._stderr/_stdout in the old version of Winston). If either output stream is closed (for whatever reason) then the `write()` function will throw an error which goes unhandled. Since we don't need Winston's ability to dynamically switch between `stderr` and `stdout` based on the log level we can just leverage `console.*` directly which takes [care of any potential exceptions](https://github.com/nodejs/node/blob/916227b3ed041ff13d588b02f98e9be0846a5a7c/lib/internal/console/constructor.js#L277-L295).

In this PR we upgrade Winston which by itself doesn't solve the problem but it lets us (easier) build our own transport for writing to the Console.

## Release notes

Notes:

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -68,8 +68,8 @@
     "username": "^5.1.0",
     "uuid": "^3.0.1",
     "wicg-focus-ring": "^1.0.1",
-    "winston": "^2.3.1",
-    "winston-daily-rotate-file": "^1.4.5"
+    "winston": "^3.6.0",
+    "winston-daily-rotate-file": "^4.6.1"
   },
   "devDependencies": {
     "devtron": "^1.4.0",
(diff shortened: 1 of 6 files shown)


sergiou87 (approved):
LGTM! It seems to work great too :shipit:

myldn (commented):
Duplicate of #

PR #13963, opened 2022-02-21
Never report more than one fatal error

niik:
## Description

Related to #13962, this PR ensures we never spam `central` with more than one fatal error per application session. Our non-fatal errors already have rate-limiting built-in (see https://github.com/desktop/desktop/blob/development/app/src/lib/helpers/non-fatal-exception.ts).

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/exception-reporting.ts
+++ b/app/src/main-process/exception-reporting.ts
@@ -5,6 +5,8 @@ const ErrorEndpoint = 'https://central.github.com/api/desktop/exception'
 const NonFatalErrorEndpoint =
   'https://central.github.com/api/desktop-non-fatal/exception'
 
+let hasSentFatalError = false
+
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
Looks great! As a side note, we talked yesterday about why https://github.com/desktop/desktop/blob/7bda8f2f8aed3fae363b989fa5710f28f08815b0/app/src/main-process/main.ts/#L124-L128 had to be `on` instead of `once` [as in the renderer](https://github.com/desktop/desktop/blob/7bda8f2f8aed3fae363b989fa5710f28f08815b0/app/src/ui/index.tsx/#L195-L198).

Maybe it's a good opportunity to leave a comment there with the reasoning.

sergiou87:
I'll merge this for now to deploy a beta, we can add that comment later.

PR #13965, opened 2022-02-21
24-Hour time format support

### LINKED ISSUE #13943: Option to display time/date with 24h clock

Such as when getting tooltip for a commit, it displays a date, but for some obscure reason it's displaying it with 12h clock which is not used in my locale.

### Proposed solution

Either detect the locale correctly or provide option to swap to 24h clock. Both would be preferable so people can choose which they prefer.

### Additional context

12h clock is not used where I live.

tsvetilian-ty:
Closes #13943 

## Description
- Allow visualization of 24-Hour time format

Note: Defaults to 12-Hour time format

### Screenshots

image

image

## Release notes

Added: Support for 24-Hour Time Format

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -299,6 +299,11 @@ export interface IAppState {
    * Whether or not the user enabled high-signal notifications.
    */
   readonly notificationsEnabled: boolean
+
+  /**
+   * Whether or not to use 24-Hour date formating.
+   */
+  readonly is24hourFormat: boolean
 }
 
 export enum FoldoutType {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -350,6 +350,7 @@ const InitialRepositoryIndicatorTimeout = 2 * 60 * 1000
 
 const MaxInvalidFoldersToDisplay = 3
 
+export const timeFormatKey = '24hour-time-format'
 const hasShownCherryPickIntroKey = 'has-shown-cherry-pick-intro'
 const dragAndDropIntroTypesShownKey = 'drag-and-drop-intro-types-shown'
 const lastThankYouKey = 'version-and-users-of-last-thank-you'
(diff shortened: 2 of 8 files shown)


uttiya10:
Might be nice if theres was third (default) option to automatically use the user's device settings. Similar to what we have for dark/light mode.

To get the user's locale, you can do `window.navigator.language`, and pass it to moment.

tsvetilian-ty:
@uttiya10 that was my intent, to use `navigator.language` or `app.getLocale()`, the issue would occur when the locale is passed to "moment." In `webpack.common.ts,` all locales were ignored because they increased the size of the library. Even the example in webpack docs (https://webpack.js.org/plugins/ignore-plugin/#example-of-ignoring-moment-locales) is with "moment." Because of that, I decided NOT to implement system detection 😢 and let the user pick

uttiya10:
Ahh yes, I see that 🤔. Considering the app is already 143mb, I wonder if an extra couple hundred kilobytes is a problem.

I'm just worried that there might be places that use 24-hour time but do **not** use the `Day, Month, Year` format.

tsvetilian-ty:
@uttiya10, great point! When I picked English (UK) as the default for a 24-hour format, I didn't think of that. Now I checked in [moment's page](https://momentjs.com/) to see if there are such locations (not matching 24-hour and `Day, Month, Year` format), and it looks like all locales with 24-hour time format are `Day, Month, Year.`

uttiya10:
Well in that case, this should be fine, unless theres a edge case I’m missing

sergiou87:
Thanks for your contribution @tsvetilian-ty! I like the idea behind this PR, but we're not very fond of new settings unless there is no other way. If using the system locale with `moment` only takes a few extra KB, I'd say that's the preferred option. However, I'd be cautious about what happens if the system is in a different language, in case the date would be translated to that language. For example, if my OS is in Spanish, I wouldn't want to see all texts on GitHub Desktop in English, and then the dates in Spanish.

sergiou87:
Ok, so… we've been talking about this today and we're gonna give this another thought. Internationalization is a really complex task and Electron might not have the best APIs right now for this. Ideally, we would like to add better/some support for i10n in general, sooner rather than later.

Since we'd rather not add new settings unless strictly needed, we're gonna close it for now, sorry!

PR #13980, opened 2022-02-23
Clone repo dialog: update local path after the initial one is retrieved

### LINKED ISSUE #13971: Local path no longer pre-filled with repo name
### Describe the bug

When cloning a repo from GitHub it will launch GitHub Desktop and open the "Clone a repository" window correctly, but the "Local path" option doesn't seem to be pre-filled anymore.

I'm positive in earlier versions that it would automatically create a folder or pre-populate the Local path input with the repo name. 

But now it's always just my default folder with an error saying "This folder contains files. Git can only clone to empty folders". So now I have to copy-paste the repo name from the Repository URL field down to the end of the Local path. I've never had to do this before, it used to happen automatically.

### Version & OS

Version 2.9.9 (x64) - Windows 11.

1. Go to any project on GitHub
2. Click on Clone - Open with GitHub Desktop
3. See how the Local path is no longer populated

### Expected behavior

I expect for it to either automatically append the repo name to the Local Path (which I _think_ is what it used to do?), or for you to be able to specify any folder as the Local path and when you click 'Clone' for it to automatically create a folder based on the name of the repo.

### Screenshots

image

sergiou87:
Closes #13971

## Description

The changes in #13804 introduced another regression when the `getDefaultPath` function was made async. In this case, it would manifest when clicking on a link like `x-github-client://openRepo/https://github.com/desktop/askpass-trampoline`. In this case, the URL would be ready before the initial path, and when the initial path is retrieved, it wouldn't get the repo name added to it.

## Release notes

Notes: [Fixed] Pre-fill clone path with repository name

### DIFF
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -213,6 +213,11 @@ export class CloneRepository extends React.Component<
       enterpriseTabState,
       urlTabState,
     })
+
+    // Update the local path based on the current url now that we have an
+    // initial path
+    const selectedTabState = this.getSelectedTabState()
+    this.updateUrl(selectedTabState.url)
   }
 
   public componentWillUnmount() {
@@ -580,9 +585,15 @@ export class CloneRepository extends React.Component<
     const tabState = this.getSelectedTabState()
     const lastParsedIdentifier = tabState.lastParsedIdentifier
 
+    // If there is no path yet, just update the url
+    if (tabState.path === null) {
+      this.setSelectedTabState({ url }, this.validatePath)
+      return
+    }
+
     let newPath: string
 
-    const dirPath = tabState.path ?? ''
+    const dirPath = tabState.path
     if (lastParsedIdentifier) {
       if (parsed) {
         newPath = Path.join(Path.dirname(dirPath), parsed.name)


app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -580,9 +585,15 @@ export class CloneRepository extends React.Component<
     const tabState = this.getSelectedTabState()
     const lastParsedIdentifier = tabState.lastParsedIdentifier
 
+    // If there is no path yet, just update the url
+    if (tabState.path === null) {
+      this.setSelectedTabState({ url }, this.validatePath)
+      return
```
sergiou87, line 591: It's important not to set the `lastParsedIdentifier: parsed` in the tab state before we have the path. Otherwise, next time the path is changed, the app will assume the last path component is the repo name and can be changed when the URL is updated.

For example, if the initial path is `/Users/sergiou87/Documents/GitHub` and the `lastParsedIdentifier` is not `null`, when I change the URL to point to a different repository, it will remove that `GitHub` and put the repo name in there (`/Users/sergiou87/Documents/askpass-trampoline`) instead of appending it (`/Users/sergiou87/Documents/GitHub/askpass-trampoline`).

niik (approved):
Looks great!

PR #14006, opened 2022-02-24
Prevent crash when console.log fails as part of logging to stdout/stderr

niik:
## Description

Continuation of #13962, apparently `console.log` in Node isn't entirely safe from exceptions either.

## Release notes

Notes: no-notes


sergiou87 (approved):
✨

PR #14007, opened 2022-02-24
Leverage comspec when opening protocol handler in CLI

niik:
## Description

This matches Node's behavior when spawning a process with `shell: true` with the exception that we don't account for other shells than `cmd.exe`.

## Release notes

Notes: no-notes


sergiou87 (approved):
Looks good!

PR #14022, opened 2022-02-28
Use built-in fileURLToPath

niik:
## Description

Stumbled upon this while looking at something completely unrelated. Turns out there's a built-in function for converting file URIs to platform-specific paths. Might as well use that and save us a dependency!

@tidy-dev This function was introduced in Node 10 so I guess it was there when I asked you to use the package :shrug:

## Release notes

Notes: no-notes


tidy-dev (approved):
✨  Makes sense and works as well.

PR #14023, opened 2022-02-28
Flip expanded state icon in CheckRunListItem

niik:
## Description

Noticed that the icon used to signal whether the check run list item is expanded or not was flipped compared to how we traditionally render exanded/collapsed state (like in our toolbar dropdowns)

#### Before
image

#### After
image

## Release notes

Notes: no-notes (not released)

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -97,8 +97,8 @@ export class CICheckRunListItem extends React.PureComponent<
         <Octicon
           symbol={
             isCheckRunExpanded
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨  When I first looked at this, my reaction was "This feels wrong". This is because I was thinking of it as a state icon, so when it is expanded it should be pointing down to reflect its state.

But.. then I went out to see what common convention was.. and TIL when the icon is before the content  on the left it is usually state reflective, but when it is on the far right it is toggle action reflective.  (However, my mind will always bulk at this unless the icon is clearly a button)

niik:
I'll merge this even though it doesn't have to go into the next prod release since it's zero risk

PR #14024, opened 2022-02-28
Fix discarding submodules in paths containing spaces

niik:
## Description

As I was working on removing our use of `substr` in the app (due to it being deprecated) I noticed something that didn't look right in our submodule status parsing logic:

https://github.com/desktop/desktop/blob/b6d027924a37a22acaf69346dd3b72908e4359a3/app/src/lib/git/submodule.ts#L66

This does not account for the case of submodules with spaces in their path.

```
Markuss-MBP:submodule-test markus$ git submodule status --
 39078e1b2fa43e4b32c80eba9a87053c5e2b6e4a and another submodule (heads/main)
 0334f75648d6254988b4e0378c4fafe394bc4ae9 look at me i am a submodule (remotes/origin/HEAD)
```

The end result of this oversight is that submodules residing in paths with spaces in them will have the entire submodule folder tossed in the trash when discarding changes instead of running `git submodule update` on them

https://github.com/desktop/desktop/blob/b6d027924a37a22acaf69346dd3b72908e4359a3/app/src/lib/stores/git-store.ts#L1466 

## Release notes

Notes: [Fixed] Discarding submodules with spaces in their relative path now correctly updates the submodule instead of moving it to Trash

### DIFF
--- a/app/src/lib/git/submodule.ts
+++ b/app/src/lib/git/submodule.ts
@@ -53,7 +53,7 @@ export async function listSubmodules(
   // about it if you want to learn more:
   //
   // https://git-scm.com/docs/git-describe
-  const statusRe = /^.([a-f0-9]{40}) (.+?) \((.+?)\)$/gm
+  const statusRe = /^.([^ ]+) (.+) \((.+?)\)$/gm
 
   for (const [, sha, path, describe] of stdout.matchAll(statusRe)) {
     submodules.push(new SubmoduleEntry(sha, path, describe))
--- a/app/src/lib/git/submodule.ts
+++ b/app/src/lib/git/submodule.ts
@@ -21,56 +21,42 @@ export async function listSubmodules(
   // We don't recurse when listing submodules here because we don't have a good
   // story about managing these currently. So for now we're only listing
   // changes to the top-level submodules to be consistent with `git status`
-  const result = await git(
+  const { stdout, exitCode } = await git(
     ['submodule', 'status', '--'],
     repository.path,
     'listSubmodules',
-    {
-      successExitCodes: new Set([0, 128]),
-    }
+    { successExitCodes: new Set([0, 128]) }
   )
 
-  if (result.exitCode === 128) {
+  if (exitCode === 128) {
     // unable to parse submodules in repository, giving up
     return []
   }
 
   const submodules = new Array<SubmoduleEntry>()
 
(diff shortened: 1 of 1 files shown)


app/src/lib/git/submodule.ts
```diff
@@ -21,56 +21,42 @@ export async function listSubmodules(
...
+  // then the 40-character SHA represents the current commit
+  //
+  // then the path to the submodule
+  //
+  // then the output of `git describe` for the submodule in braces
+  // we're not leveraging this in the app, so go and read the docs
+  // about it if you want to learn more:
+  //
+  // https://git-scm.com/docs/git-describe
+  const statusRe = /^.([a-f0-9]{40}) (.+?) \((.+?)\)$/gm
```
sergiou87: Not an actual problem, just a comment/thought: I usually try to avoid such specific regexp like the one for commit SHAs when it's not needed, in case the format changes somehow (shorter/longer shas now or in the future… whatever), and just look for non-whitespace characters, non-space characters in this case:
```suggestion
  const statusRe = /^.([^ ]+) (.+?) \((.+?)\)$/gm
```

Feel free to ignore this comment though (specially if my regexp is wrong 😂 )

sergiou87: Also it seems it doesn't work for folders with `()` in them:

```javascript
Array.from(' 1eaabe34fc6f486367a176207420378f587d3b48 this is a long (folder name) (v2.16.0-rc0)'.matchAll(/^.([a-f0-9]{40}) (.+?) \((.+?)\)$/gm))
```
image

Maybe it's not even worth to worry about this, though, but I think just making the middle part greedy fixes it:
```suggestion
  const statusRe = /^.([^ ]+) (.+) \((.+?)\)$/gm
```

image

What do you think?

sergiou87 (dismissed):
Looks good! Just a couple of comments/thoughts that can be ignored, so I'll approve if you want to merge this right away.

app/src/lib/git/submodule.ts
niik: I think both your suggestions are A+! :sparkles:

sergiou87 (approved):
Thanks for those changes! Tested and works like a charm 🔥 :shipit:🔥

ammmora:
Omg you work on that

PR #14025, opened 2022-02-28
Stop using substr, part one

niik:
## Description

[`substr`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) has been deprecated in favor of [`substring`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) and now VSCode is letting us know by showing our substr code in ~~strikethrough~~ which I can't stand.

This PR updates all of our `.substr` uses with `substring` (except for `submodule.ts` which is addresses in #14024).

There's three commits in this PR and my recommendation to anyone who reviews this PR is to review them separately, and in order. The first (09f16b4184a54da4eea9ae40743a1dbe188a7f8a) contains only mechanical translations following these assumptions (verify my assumptions pls).

```
Any call using .substr(X) can be substituted for .substring(X)
Any call using .substr(0, X) can be substituted for .substring(0, X)
```

So while this commit will look like it's introducing a lot of changes a reviewer only needs to verify that the changes within it correspond to the two rules above.

The second (1f72120cb8247ea2f1325ecd723192fcabb29b53) and third (2f70c3c470043346610d00776f98475b74cb2773) commits deal with the cases where mechanical translation wasn't possible and require a little more review since they're "hand written"

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/rebase.ts
+++ b/app/src/lib/git/rebase.ts
@@ -104,7 +104,7 @@ export async function getRebaseInternalState(
     )
 
     if (targetBranch.startsWith('refs/heads/')) {
-      targetBranch = targetBranch.substr(11).trim()
+      targetBranch = targetBranch.substring(11).trim()
     }
 
     baseBranchTip = await FSE.readFile(
--- a/app/src/lib/git/submodule.ts
+++ b/app/src/lib/git/submodule.ts
@@ -63,7 +63,7 @@ export async function listSubmodules(
     //
     // https://git-scm.com/docs/git-describe
 
-    const [path, describeOutput] = entry.substr(42).split(/\s+/)
+    const [path, describeOutput] = entry.substring(42).split(/\s+/)
 
     // if the submodule has not been initialized, no describe output is set
     // this means we don't have a submodule to work with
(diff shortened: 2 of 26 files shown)


tidy-dev:
Hmm.. In verifying these assumptions, I checked [the docs.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr)

I think 
```
Any call using .substr(0, X) can be substituted for .substring(0, X)
```
is wrong because, in `.substr(0,X)`, X should be representative of an index where ` .substring(0, X)`, X should be representative of a character count. 

This could also make the `.substr(0,X)` to ` .substring(X)` possibly off by one character

That said.. I have not done any testing on this yet... to verify this.

niik:
Extra :sparkles: for looking so thoroghly at this.

> is wrong because, in `.substr(0,X)`, X should be representative of an index where ` .substring(0, X)`, X should be representative of a character count.

That's right, but not the last index to include but rather "The index of the **first character to exclude** from the returned substring" which ought to make it analog to length (as long as the start index is zero obviously). I.e it's lower inclusive, upper exclusive.

```
substring: The index of the first character to exclude from the returned substring.
substr: The number of characters to extract.
```

tidy-dev:
> as long as the start index is zero obviously

Thanks for saying that 🤦 .  When I first read the assumption, my mind glanced over the fact that it was zero based. So yeah it makes sense.

tidy-dev (dismissed):
✨  Looks great and I pulled it down and spot tested it.

tidy-dev (approved):

PR #14034, opened 2022-03-01
Avoid spawning from main process on Catalina

### LINKED ISSUE #13974: GitHub Desktop quit unexpectedly
### Describe the bug
After opening GitHub Desktop a system alert says _"GitHub Desktop quit unexpectedly"_. The alert shows up every single time GitHub Desktop is opened. However, there is no impact on using the program, which so far works normally (it's just a minor nuisance). 

### Version & OS

GitHub Desktop Version 2.9.9 (x64)
Mac OS Catalina (version 10.15.7)

1. Just open GitHub Desktop

If GitHub Desktop is already running:

1. Close it (Command + Q)
2. Reopen it

### Expected behavior

No error alert.

### Actual behavior

An error alert shows up (however, there's no impact on the program, which runs normally).

### Screenshots

Screen Shot 2022-02-23 at 11 57 02 am

### Logs

This is the log created by Catalina:
[log.txt](https://github.com/desktop/desktop/files/8121345/log.txt)

The logs from GitHub Desktop seem to be normal (I can share those if requested).

niik:
Closes #13974

## Description

We're tracking a bug that causes a harmless but annoying crash report dialog to appear when launching the app and when clicking `Show in Finder` on macOS Catalina (or lower). While we don't know the root cause of the problem yet we know that spawning from the main process is what triggers it (see @sergiou87's excellent repro case at https://github.com/electron/electron/issues/32718#issuecomment-1050009071).

As such we're trying to mitigate this by auditing the places in the main process where we spawn and relocating them where possible to the renderer.

Fortunately for us there seems to be only two cases to worry about. The first being the patching of the process enviroment and the second being our use of the `file-metadata` package which will spawn `mdls`.

Patching the main process environment is likely not necessary at all but in order to minimize the surface area of this change we (@sergiou87 and I) opted to only disable it on the affected platforms for now.

## Release notes

Notes: [Fixed] Prevent crash report dialog from appearing when launching on macOS Catalina or earlier

### DIFF
--- a/app/src/lib/get-os.ts
+++ b/app/src/lib/get-os.ts
@@ -21,6 +21,11 @@ function systemVersionGreaterThanOrEqualTo(version: string) {
   return sysver === undefined ? false : compare(sysver, version, '>=')
 }
 
+function systemVersionLessThan(version: string) {
+  const sysver = getSystemVersionSafe()
+  return sysver === undefined ? false : compare(sysver, version, '<')
+}
+
 /** Get the OS we're currently running on. */
 export function getOS() {
   const version = getSystemVersionSafe()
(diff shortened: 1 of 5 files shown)


taurus228 (commented):
you-can-crash-if-you-want-to-you-can-leave-your-friends-behind

sergiou87 (commented):
I'll let @tidy-dev approve this one, since I "helped" writing it 😆 

In any case, I re-reviewed the code and all (still) looked good to me!

tidy-dev (approved):
✨  Looks good and works great!

PR #14051, opened 2022-03-02
Replace custom spawn implementations with promisified exexFile

niik:
## Description

I found that we were maintaining two different spawn-based implementations intended to simplify executing a command and returning its stdout. This, turns out, is an excellent job for [`execFile`](https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback) and by promisifying it we can simplify our own logic quite a bit.

As such, this PR introduces two new methods, one promisified execFile used in the renderer and/or main process and one `sh` method which is used in our script files. The `sh` file replaces pre-existing logic in our scripts and executes using the system shell which means that we can't pass untrusted output to it (we weren't so there's no change there, just calling it out).

One different between the `captureCommandOutput` method and the new `execFile` is that the old version didn't have any limits on how large the captured output could be whereas `execFile` by default caps the output length to 1 mb. For our use case (locating Git on Path) this limit is more than sufficient.

## Release notes

Notes: no-notes


ghost:
use-the-execfile-luke

sergiou87 (approved):
Nice cleanup! ✨

tidy-dev (approved):
✨  Pretty
image

PR #14055, opened 2022-03-02
Replace file-metadata with custom execFile implementatation

niik:
## Description

When attempting to upgrade file-metadata (in order to bump its plist dependency) we ran into trouble with it having been converted to [an ESM package](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). We currently can't load pure ESM packages (though #13894 is moving us in the right direction there).

So instead of upgrading I looked at what the file-metadata package really does which is in the end pretty simple and decided to replicate that instead. Since we don't need full plist parsing capabilities we can just look for our keywords in the output from `mdls`.

#### How to test this

- Copy an application bundle to somewhere where you've got write-access to it
`cp -r /System/Applications/Calculator.app .`

- Create a git repository inside the bundle
`cd Calculator.app && git init .`

- Add the repository to GitHub Desktop (the file picker won't let you so you have to enter the path manually) or use the `github` CLI tool
- Click on the `Repository` -> `Show in Finder` menu item and verify that a Finder window opens with the Calculator app selected. If this new logic doesn't work it'll launch the Calculator instead.

We can simplify this a tiny bit more once #14051 lands.

## Release notes

Notes: no-notes


sergiou87 (changes requested):
LGTM! Although we can probably delete `app/src/lib/helpers/file-metadata.js` in this PR too 🔥

sergiou87 (approved):
✨

sergiou87:
@tidy-dev since @niik and I worked together on this, do you mind taking an extra look at it before we merge it? 🙏

tidy-dev (approved):
✨  Looks good and tested!

PR #14060, opened 2022-03-02
[DO NOT MERGE] Test PR to Reproduce Github Desktop Issue when Rerunning Checks

uttiya10:
## Description
I will use this PR in a new issue to describe an bug I have experienced with re-running checks from Github Desktop while merging [#13957](https://github.com/desktop/desktop/pull/13957#issuecomment-1049289434). 

This issue seems to only occur with repositories where I do not have permission to manually rerun checks, which I why I am opening this PR here, and not on a personal fork.

Edit: Relevant issue is #14061


tidy-dev:
This was used to reproduce of a bug to create a screen capture, but is not needed anymore (or at least until there is a fix to test). Going to close

PR #13706, opened 2022-01-19
Preserve login case

niik:
Fixes: #9890
Superseedes: #10013

## Description

Adds a migration for our owners table to add a case-insensitive unique key so that we can preserve case in the `login` field.

This migration will not magically fix the cases where we've lost information by lowercasing logins but it will mean that we'll gradually swap out incorrectly cased logins for the accurate ones as we receive fresh data from the API. We typically refresh the repository API information when switching repositories and since there's potentially multiple repositories pointing to the same owner one repository refresh should sort of the login case for all repositories owned by that user.

## Release notes

Notes:

PR #14065, opened 2022-03-03
Fix error storing repository owners in database

sergiou87:
## Description

Introduced in #13706 (and I suspect after upgrading Electron/Chromium), the IndexDB doesn't like when we pass an `undefined` id, because it tries to store it. Instead, we shouldn't even set that `id: undefined` if we don't have an `id` at all and it will generate one by incrementing the last one in 1.

## Release notes

Notes: no-notes (not released to prod yet)

### DIFF
--- a/app/src/lib/stores/repositories-store.ts
+++ b/app/src/lib/stores/repositories-store.ts
@@ -383,7 +383,8 @@ export class RepositoriesStore extends TypedBaseStore<
     // login with the proper case.
     if (existingOwner === undefined || existingOwner.login !== login) {
       id = existingOwner?.id
-      id = await this.db.owners.put({ id, key, endpoint, login })
+      const existingId = id !== undefined ? { id } : {}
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨  Makes sense for getting rid of updating an undefined id..

Just curious cause I don't know exactly how this works with the local db.. but when we do this, does the db automatically assign an id?

niik (approved):
Thanks for taking care of that @sergiou87! 💖 

For future reference it seems this would not be necessary if we were on the most recent version of Dexie since they've introduced a workaround which removes undefined primary keys before passing it to IndexedDB: https://github.com/dexie/Dexie.js/commit/4abacd080f2033b775fba196fae2e7eb8b5ee8ee

sergiou87:
@tidy-dev exactly, because of this `++id` https://github.com/desktop/desktop/blob/1ae05a079cc98ad78d5df5fb4b1f6bea3cf6e72c/app/src/lib/databases/repositories-database.ts/#L103

PR #14078, opened 2022-03-04
fs/promises: readFile

niik:
## Description

`readFile` from fs-extra is just [a promisified version](https://github.com/jprichardson/node-fs-extra/blob/9.0.1/lib/fs/index.js#L30) of the readFile from `fs` so we can safely replace it with `fs/promises` instead.

Our custom `readFile` method is a promise-wrapper over an already promised-based FSE.readFile offering no added benefits (the default options [match the default options of fsPromises](https://nodejs.org/api/fs.html#fspromisesreadfilepath-options))

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/file-system.ts
+++ b/app/src/lib/file-system.ts
@@ -53,39 +53,6 @@ export function tailByLine(
 }
 
 /**
- * Asynchronous readFile - Asynchronously reads the entire contents of a file.
- *
- * @param fileName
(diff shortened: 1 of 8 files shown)


sergiou87 (approved):
I think I'm gonna love this series of refactors 🔥 🍿

tidy-dev (approved):
✨

PR #14079, opened 2022-03-04
fs/promises: mkdtemp, createReadStream, writeFile, appendFile, readlink, symlink, unlink, readdir, and realpath

niik:
## Description

Based on #14078, this replaces our use of the following `fs-extra` functions: `mkdtemp`, `createReadStream`, `writeFile`, `appendFile`, `readlink`, `symlink`, `unlink`, `readdir`, and `realpath`.

`createReadStream` is the exact same function as that from `fs` and the others are [promisified versions](https://github.com/jprichardson/node-fs-extra/blob/7b12b058e27df560ba777756f38f977662c23750/lib/fs/index.js#L8-L40) of their fs counterparts and can be safely replaced.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/add-repository/gitignores.ts
+++ b/app/src/ui/add-repository/gitignores.ts
@@ -1,6 +1,6 @@
 import * as Path from 'path'
-import { readdir, writeFile } from 'fs-extra'
-import { readFile } from 'fs/promises'
+import { readdir } from 'fs-extra'
+import { readFile, writeFile } from 'fs/promises'
 
 const GitIgnoreExtension = '.gitignore'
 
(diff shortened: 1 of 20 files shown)


app/src/ui/add-repository/gitignores.ts
```diff
@@ -1,6 +1,6 @@
 import * as Path from 'path'
-import { readdir, writeFile } from 'fs-extra'
-import { readFile } from 'fs/promises'
+import { readdir } from 'fs/promises'
+import { readFile, writeFile } from 'fs/promises'
```
sergiou87: All in one import maybe?
```suggestion
import { readdir, readFile, writeFile } from 'fs/promises'
```

sergiou87 (commented):
Looks great!! Only found one minor thing 😄

app/src/ui/add-repository/gitignores.ts
niik: Love it!

sergiou87 (approved):
:shipit: ❗

PR #14080, opened 2022-03-04
fs/promises: pathExists

niik:
Closes #[issue number]

## Description

Based on #14079 this PR replaces the `pathExists` method from `fs-extra` with our own implementation. As you can see it matches the `fs-extra` implementation closely albeit with a slightly different syntax for brevity

https://github.com/jprichardson/node-fs-extra/blob/7b12b058e27df560ba777756f38f977662c23750/lib/path-exists/index.js#L5-L7

```ts
function pathExists (path) {
  return fs.access(path).then(() => true).catch(() => false)
}
```

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/path-exists.ts
+++ b/app/src/ui/lib/path-exists.ts
@@ -1,0 +1,9 @@
+import { access } from 'fs/promises'
+import { constant } from 'lodash'
+
+/**
+ * Returns a value indicating whether or not the provided path exists (as in
+ * whether it's visible to the current process or not).
+ */
+export const pathExists = (path: string) =>
+  access(path).then(constant(true), constant(false))
--- a/app/src/ui/main-process-proxy.ts
+++ b/app/src/ui/main-process-proxy.ts
@@ -3,6 +3,7 @@ import { RequestResponseChannels, RequestChannels } from '../lib/ipc-shared'
 import * as ipcRenderer from '../lib/ipc-renderer'
 import { stat } from 'fs/promises'
 import { isApplicationBundle } from '../lib/is-application-bundle'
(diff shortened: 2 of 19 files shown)


app/src/ui/main-process-proxy.ts
```diff
@@ -90,7 +91,7 @@ export const focusWindow = sendProxy('focus-window', 0)
 const _showItemInFolder = invokeProxy('show-item-in-folder', 1)
 
 export const showItemInFolder = (path: string) =>
-  stat(path)
+  pathExists(path)
     .then(() => _showItemInFolder(path))
     .catch(err => log.error(`Unable show item in folder '${path}'`, err))
```
haykam821, line 96: Wouldn't this catch handler never be called, since the result is being passed through a boolean in the fulfilled handler instead?

niik: You're correct that `pathExists` doesn't throw. `_showItemInFolder` could still throw though and its exception would travel over the IPC boundary and be caught here.

app/src/ui/lib/path-exists.ts
```diff
@@ -0,0 +1,9 @@
+import { access } from 'fs/promises'
+import { constant } from 'lodash'
+
+/**
+ * Returns a value indicating whether or not the provided path exists (as in
+ * whether it's visible to the current process or not).
+ */
+export const pathExists = (path: string) =>
+  access(path).then(constant(true), constant(false))
```
sergiou87, line 9: TIL about `constant`

sergiou87 (approved):
🤩

tidy-dev (approved):

PR #14081, opened 2022-03-04
fs/promises: ensureDir

niik:
Closes #[issue number]

## Description

Based on #14080, this PR substitutes our use of `ensureDir` from `fs-extra` for mkdir with the `recursive` option from `fs/promises`.

Note that `ensureDir` and `mkdirp` are both aliases for [fs-extras internal makeDir](https://github.com/jprichardson/node-fs-extra/blob/c8815e3ccf130422b427484007f73029075b56b6/lib/mkdirs/index.js#L10-L13).

Our implementation here differs from that of `fs-extra` in that it doesn't resolve the path to an absolute path before passing it to `mkdir`, see https://github.com/jprichardson/node-fs-extra/blob/c8815e3ccf130422b427484007f73029075b56b6/lib/mkdirs/make-dir.js#L49

As far as I can tell that shouldn't be necessary except on Windows. On posix systems Node will call [mkdir(2)](https://github.com/nodejs/node/blob/master/deps/uv/src/unix/fs.c#L1717) which handles relative paths just fine and on Windows Node will call [toNamespacedPath](https://github.com/nodejs/node/blob/b35ee26f03021eb196fff0dec47291b0bb2f1c3a/lib/internal/fs/promises.js#L665) which will [resolve on Windows](https://github.com/nodejs/node/blob/b35ee26f03021eb196fff0dec47291b0bb2f1c3a/lib/path.js#L623) (toNamespacedPath is noop on unix).

There's also no mention about special care or treatment of relative paths in the Node docs for mkdir so I'm inclined to not add any extra behavior until we know we need it.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/helpers/create-tutorial-repository.ts
+++ b/app/src/lib/stores/helpers/create-tutorial-repository.ts
@@ -1,8 +1,7 @@
 import * as Path from 'path'
 
 import { Account } from '../../../models/account'
-import { ensureDir } from 'fs-extra'
-import { writeFile } from 'fs/promises'
+import { mkdir, writeFile } from 'fs/promises'
 import { API } from '../../api'
 import { APIError } from '../../http'
 import {
@@ -120,7 +119,7 @@ export async function createTutorialRepository(
   const branch = repo.default_branch ?? (await getDefaultBranch())
   progressCb('Initializing local repository', 0.2)
 
-  await ensureDir(path)
+  await mkdir(path, { recursive: true })
 
   await git(
     ['-c', `init.defaultBranch=${branch}`, 'init'],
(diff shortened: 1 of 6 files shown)


sergiou87 (approved):
Looks good! I kind of liked the semantics of `ensureDir` to express the intent vs invoking `mkdir` recursively, but it's such a simple thing that I think we can live with the `recursive: true` bit 😄

niik:
> Looks good! I kind of liked the semantics of `ensureDir` to express the intent vs invoking `mkdir` recursively, but it's such a simple thing that I think we can live with the `recursive: true` bit 😄

I'm happy to reintroduce a custom ensureDir if you'd like, I agree that it got kind messy with the options all of the plalce

sergiou87:
I'm kind of used to do `mkdir -p` in scripts to ensure a directory exists, so maybe that's why I find it normal. I wonder what @tidy-dev thinks (assuming she's not used to `mkdir -p` or similar given she had more experience with Windows in the past).

tidy-dev:
Vast majority of my exp is pure web apps - not electron (even tho using windows to build them).. which means niether of them jump out to me as normal. Either one is good to me. Only reason I lean towards what we have is I could easily see that recursive flag and google it to see what it does. :D

tidy-dev (approved):

PR #14082, opened 2022-03-04
fs/promises: ensureFile

niik:
## Description

Based on #14081 this removes our dependency on the `ensureFile` method of fs-extra. The [ensureFile method](https://github.com/jprichardson/node-fs-extra/blob/c8815e3ccf130422b427484007f73029075b56b6/lib/ensure/file.js#L8-L41) takes care of creating both the file and the directory (recursively). Luckily for us we don't need that as `mkdtemp` (in getTempFilePath) will already have taken care of creating the directory.

There's a behavioral change here in that this new method will throw if the file already exists rather than overwriting its contents. If the file exists then mkdtemp has been broken somehow and we can't trust it to not give us a location that's dangerous to overwrite.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/progress/lfs.ts
+++ b/app/src/lib/progress/lfs.ts
@@ -1,12 +1,18 @@
-import * as FSE from 'fs-extra'
 import { getTempFilePath } from '../file-system'
 import { IGitProgress, IGitProgressInfo, IGitOutput } from './git'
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
I think that makes total sense!

tidy-dev (approved):

PR #14083, opened 2022-03-04
fs/promises: remove

niik:
## Description

Based on #14082 this PR substitutes the `remove` method from `fs-extra` for the `rm` method in fs/promises. This is exactly what the `remove` method does currently: https://github.com/jprichardson/node-fs-extra/blob/c8815e3ccf130422b427484007f73029075b56b6/lib/remove/index.js#L9

## Release notes

Notes: no-notes

PR #14084, opened 2022-03-04
fs/promises: synchronous methods

niik:
## Description

Based on #14083 this PR replaces all synchronous fs-extra methods (ie. in our scripts) with their fs counterparts with the exception of `copySync` which will recursively copy a directory and for which `fs-extra` still seems to be the [recommended option](https://stackoverflow.com/a/64498004/2114).

Note that we'll still be able to remove `fs-extra` from our app, this only affects the root level package.json which governs the build process.

## Release notes

Notes: no-notes

### DIFF
--- a/script/build.ts
+++ b/script/build.ts
@@ -3,7 +3,6 @@
 
 import * as path from 'path'
 import * as cp from 'child_process'
-import * as fs from 'fs-extra'
 import * as os from 'os'
 import packager, {
   OfficialArch,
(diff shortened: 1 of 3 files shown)


sergiou87 (approved):
It's a shame we can't get rid of `copySync` from `fs-extras` but well 😢  Other than that looks great!

niik:
> It's a shame we can't get rid of `copySync` from `fs-extras` but well 😢 Other than that looks great!

I briefly considered just shelling out to `cp` and `xcopy` (on Windows) but xcopy doesn't support not overwriting existing files. Oh well, it's something we can tackle later

tidy-dev (approved):

PR #14088, opened 2022-03-04
Simplify updateEnvironmentForProcess

niik:
## Description

This is a cleanup me and @sergiou87 discovered while working on #14034. We have no reason to suspect that we need the env process to run detached so we can simplify things a lot by leveraging the promise based execFile implementation introduced in #14051

As I was writing this up I took a look at [Atom's current implementation](https://github.com/atom/atom/blob/d016186110ffbe588a45d594803f9f9078804085/src/update-process-env.js) (that's where we got this from initially) and realized that they had run into an issue with environment variables containing newlines and had switched to using `awk` instead of `env` for that reason. So I figured we could do the same.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/shell.ts
+++ b/app/src/lib/shell.ts
@@ -1,12 +1,8 @@
 /* eslint-disable no-sync */
 
-import * as ChildProcess from 'child_process'
-import * as os from 'os'
+import { ExecFileOptions } from 'child_process'
+import { execFile } from './exec-file'
 import { isMacOSCatalinaOrEarlier } from './get-os'
-
-type IndexLookup = {
-  [propName: string]: string
-}
 
 /**
  * The names of any env vars that we shouldn't copy from the shell environment.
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):

app/src/lib/shell.ts
```diff
@@ -36,144 +30,39 @@ export function shellNeedsPatching(process: NodeJS.Process): boolean {
...
- * @param updateEnvironment a callback to fire if a valid environment is found
+ * Update the current process's environment variables using environment
+ * variables from the user's shell, if they can be retrieved successfully.
  */
-async function getEnvironmentFromShell(
-  updateEnvironment: (env: IndexLookup) => void
-): Promise<void> {
-  if (__WIN32__) {
+export async function updateEnvironmentForProcess(): Promise<void> {
+  if (!__DARWIN__) {
```
sergiou87, line 38: Shouldn't we keep the previous check? It seems [Atom also runs this code for Linux](https://github.com/atom/atom/blob/d016186110ffbe588a45d594803f9f9078804085/src/update-process-env.js#L11). While we don't actively support Linux, it'd be a pity if we "broke" it or make it more difficult for the fork to keep up 🤔 

What do you think?
```suggestion
  if (__WIN32__) {
```

sergiou87 (commented):
Nice cleanup! Only have a question/suggestion about Linux, other than that LGTM!

app/src/lib/shell.ts
niik: I don't think we would break it as we only run this method if `shellNeedsPatching` returns true and that method has always required darwin:

https://github.com/desktop/desktop/blob/24dc211c6fdda0a2d5844a806ffaf3ca7db52009/app/src/lib/shell.ts#L26-L37

sergiou87: Ohhh! Right, all good then 😄

sergiou87 (approved):
🔥

PR #14109, opened 2022-03-07
Use our own slightly more type safe pick

niik:
## Description

We (as in myself mainly) recently started using lodash's `pick` method and I kinda like it. One thing I don't like about it though is that it accepts property paths as well as key names. I.e. you can do

```ts
const a = { foo: { bar: true } }
pick(a, 'foo.bar')
```
This means that we can't get autocompletion on property names like we would if we used a `keyof` parameter. It also means that typos runs the risk of going unnoticed as it'll happily accept any string as a property name.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨  Nice!

PR #5186, opened 2018-07-18
improve status parsing with this one weird trick

shiftkey:
I uncovered in https://github.com/desktop/desktop/pull/5053#issuecomment-405968534 that's there's a known O(N) issue with using `Array.shift` to return the first element from an array. This means `parsePorcelainStatus` can spend a significant amount of it's time churning on an array that's 100000 items or so (which can easily happen when users have a lot of untracked files by mistake).

The V8 team do have some tricks for optimizing this, but [this discussion](https://bugs.chromium.org/p/v8/issues/detail?id=3059) shows that it shouldn't be relied upon:

> `Array.shift()` is an expensive operation, as it requires all array elements to be moved. 

So I stumbled upon [`double-ended-queue`](https://github.com/petkaantonov/deque) which talks about this in more detail, and offers an easy way to convert an array to a queue.

## Benchmarking

This is more paper-napkin-esque timings using the live app than any sort of structured benchmarking, but here's where I added timings to the `getStatus` call so I can see how time is spent:

```diff
diff --git a/app/src/lib/git/status.ts b/app/src/lib/git/status.ts
index 7390d855e..cbfcaebe9 100644
--- a/app/src/lib/git/status.ts
+++ b/app/src/lib/git/status.ts
@@ -64,11 +64,13 @@ function convertToAppStatus(status: FileEntry): AppFileStatus {
 export async function getStatus(
   repository: Repository
 ): Promise<IStatusResult | null> {
+  console.time(`status git invocation`)
   const result = await spawnAndComplete(
     ['status', '--untracked-files=all', '--branch', '--porcelain=2', '-z'],
     repository.path,
     'getStatus'
   )
+  console.timeEnd(`status git invocation`)
 
   if (result.output.length > MaxStatusBufferSize) {
     log.error(
@@ -81,6 +83,7 @@ export async function getStatus(
 
   const stdout = result.output.toString('utf8')
 
+  console.time(`status parsing`)
   // Map of files keyed on their paths.
   // Note, map maintains insertion order
   const files = new Map<string, WorkingDirectoryFileChange>()
@@ -90,7 +93,11 @@ export async function getStatus(
   let currentTip: string | undefined = undefined
   let branchAheadBehind: IAheadBehind | undefined = undefined
 
-  for (const entry of parsePorcelainStatus(stdout)) {
+  console.time(`parse porcelain`)
+  const status = parsePorcelainStatus(stdout)
+  console.timeEnd(`parse porcelain`)
+
+  for (const entry of status) {
[...]

PR #14112, opened 2022-03-07
Remove double-ended-queue dependency

niik:
## Description

I'm doing some bundle analysis following our removal of fs-extra and have identified a few candidates for easy removal. This particular change is an easy win, swapping out `double-ended-queue` for a simple loop over the array.

The introduction of `double-ended-queue` was done in https://github.com/desktop/desktop/pull/5186 as a way to mitigate performance problems with `Array.shift`. Swapping out the array for a double ended queue certainly solved that particular performance problem but simply iterating over the array eliminates one step (and one dependency).

I've done performance regression testing on these two using a 50kb status output and they're equivalent with the loop implementation usually scoring a little better but not anything worth shouting from the rooftops about.

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -30,7 +30,6 @@
     "desktop-trampoline": "desktop/desktop-trampoline#v0.9.8",
     "dexie": "^2.0.0",
     "dompurify": "^2.3.3",
-    "double-ended-queue": "^2.1.0-0",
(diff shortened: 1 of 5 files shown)


sergiou87 (approved):
This makes total sense 🔥

PR #14115, opened 2022-03-07
Markdown Mention filter

tidy-dev:
## Description
This markdown filter is meant to take username mentions such as @tidy-dev and turn them into a link to that user's profile. This filter assumes that the user exists and builds a link of `https://github.com/tidy-dev`. For enterprise users, it will use their enterprise base url to form the link. This is based on the repo passed into the filter.

This filter deviates from dotcom in that dotcom verifies the user exists and then uses the user's profile url from pulling back the users data. Thus, this filter will have false positives for users that don't exist. It shouldn't be common and worse case user clicks a link to a 404.

## Release notes
Notes: [Improved] Sandbox markdown parser parses username mentions and auto tag issue links.

### DIFF
--- a/app/src/lib/markdown-filters/mention-filter.ts
+++ b/app/src/lib/markdown-filters/mention-filter.ts
@@ -1,3 +1,5 @@
+import { GitHubRepository } from '../../models/github-repository'
+import { getHTMLURL } from '../api'
 import { INodeFilter } from './node-filter'
 
 /**
@@ -22,7 +24,7 @@ export class MentionFilter implements INodeFilter {
   private readonly beginStringNonWord = /(^|[^a-zA-Z0-9_`])/
 
   // @username and @username_emu for emu support
-  private readonly userName = /@(?<username>[a-z0-9][a-z0-9-]*_[a-zA-Z0-9]+|[a-z0-9][a-z0-9-]*)/
+  private readonly userNameRef = /(?<userNameRef>@[a-z0-9][a-z0-9-]*_[a-zA-Z0-9]+|@[a-z0-9][a-z0-9-]*)/
 
   // without a trailing slash
   private readonly withoutTrailingSlash = /(?!\/)/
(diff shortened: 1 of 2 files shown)


app/src/lib/markdown-filters/mention-filter.ts
```diff
@@ -0,0 +1,147 @@
...
+ * Becomes three nodes:
+ * 1) "That is great "
+ * 2) <link src="github.com/tidy-dev">@tidy-dev</link>
+ * 3) "! Good Job!"
+ */
+export class MentionFilter implements INodeFilter {
+  // beginning of string or non-word, non-` char
+  private readonly beginStringNonWord = /(^|[^a-zA-Z0-9_`])/
+
+  // @username and @username_emu for emu support
```
niik: What's `emu` here?

app/src/lib/markdown-filters/mention-filter.ts
```diff
@@ -0,0 +1,147 @@
...
+    'ig'
+  )
+
+  /** The parent github repository of which the content the filter is being
+   * applied to belongs  */
+  private readonly repository: GitHubRepository
+
+  public constructor(repository: GitHubRepository) {
+    this.repository = repository
+  }
```
niik:
```suggestion
  public constructor(
    /** The repository which the markdown content originated from */
    private readonly repository: GitHubRepository
  ) {}
```

app/src/lib/markdown-filters/mention-filter.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      const { userNameRef } = match.groups
+      if (userNameRef === undefined) {
+        continue
+      }
+
+      const link = this.createLinkElement(userNameRef)
+      if (link === null) {
+        continue
+      }
```
niik: Doesn't look like `createLinkElement` will ever return null?
```suggestion
```

niik (changes requested):
Looking good @tidy-dev, just a few notes!

app/src/lib/markdown-filters/mention-filter.ts
tidy-dev: :D Thanks! - following my pattern from issue mention filter.

app/src/lib/markdown-filters/mention-filter.ts
tidy-dev: I always forget you can do this...

app/src/lib/markdown-filters/mention-filter.ts
tidy-dev: Enterprise managed user. There was another "emu" usage below (copied from dotcom's notes on their regex) that I swapped for the full usage.

niik (approved):

PR #14119, opened 2022-03-08
toMilliseconds helper function

niik:
## Description

I'm investigating what it would take to get rid of our moment.js dependency both for the sake of our bundle size but also for the sake of internationalization. One thing we use moment.js for is for date calculations such as `moment().subtract(24, 'hours')`. This isn't :rocket: 👩‍🔬 but it's a bit hard to read all of the `x * 60 * 60 * 1000` inline calculations everywhere. Therefore I wrote this little helper method that let's us do things like `Date.now() - toMilliseconds(24, 'hours')`.

Extracted from my WIP branch to reduce the review burden on the final thing.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/helpers/branch-pruner.ts
+++ b/app/src/lib/stores/helpers/branch-pruner.ts
@@ -16,6 +16,7 @@ import {
 import { fatalError } from '../../fatal-error'
 import { RepositoryStateCache } from '../repository-state-cache'
 import moment from 'moment'
+import { toMilliseconds } from '../../to-milliseconds'
 
 /** Check if a repo needs to be pruned at least every 4 hours */
 const BackgroundPruneMinimumInterval = 1000 * 60 * 60 * 4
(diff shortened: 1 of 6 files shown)


app/src/lib/stores/helpers/branch-pruner.ts
```diff
@@ -183,7 +183,7 @@ export class BranchPruner {
     }
 
     // Get all branches checked out within the past 2 weeks
-    const twoWeeksAgo = moment().subtract(2, 'weeks').toDate()
+    const twoWeeksAgo = new Date(Date.now() - toMilliseconds(14, 'days'))
```
tidy-dev, line 186: Seems this pattern is really common. Wondering if a helper function would make the code a little more readable?
Like dateAgo(value, unit) => new Date(Date.now() - toMilliseconds(value, unit)) 😄  Tho maybe there is a better name.. 'beforeNow'.. 'fromNow'

tidy-dev (approved):
✨  Makes sense.

I had one suggestion of making a helper function for getting a date x units in the past to make that pattern more readable. But, just a stylistic suggestion so feel fee to ignore. :)

app/src/lib/stores/helpers/branch-pruner.ts
niik: Yeah, I was thinking about it but I had trouble coming up with a succinct name. I'm super open to it but lets deal with that in a separate PR as an improvement on top of this.

I think your suggestion of `fromNow` is great but only for future dates obviously and `beforeNow` doesn't have quite the same ring to it. Maybe `offsetFromNow(-14, 'days')` or something and we could have an `offsetFrom(date, 24, 'housr')` for when we don't want to use now()?

sergiou87: Given all examples we have here use `Date.now()`, I'd rather have something like `offsetFromNow(-14, 'days')`

niik: Cool, I'll open a new PR to simplify things with an offsetFromNow function, thanks y'all!

PR #14122, opened 2022-03-08
Dialogs overflow should be unset

### LINKED ISSUE #13944: 'Create a merge commit' Dropdown expansion
### Describe the bug
When opening the 'Create a merge commit' dropdown, it expands inside the parent, making the element scrollable.
Before this version it was opening like a "normal" dropdown menu.

### Version & OS
Version 2.9.9 (x64)
Windows 10 64bits
  
Version 2.9.9 (arm64)
MacOS Monterey v12.0.1

1. Go to 'Current Branch' 
2. Click into 'Choose a branch to merge into ...'
3. Click into 'Create a merge commit' arrow
4. See the issue

### Expected behavior

Its expecting to open as regular dropdown

### Actual behavior

The dropdown panel opens inside the parent container expanding it and making it scrollable.

### Screenshots
Expected behavior
image

Current behavior
image

### Additional context

I've noticed this happens if my Display is not scaled (100%) 
If I manually set a different scale (125%) it sorts the issue.
Both on MAC and Windows the behavior is the same.
image

tidy-dev:
Closes #13944

## Description
With the electron upgrade comes chromium upgrade which evidently means updated user agent stylesheets!

Dialog User Agent Stylesheet Before:
image

Dialog User Agent Stylesheet After:
image

The previous user agent style sheet did not set an `overflow` property for `dialog` elements and the new one sets it to `overflow: auto`. As can be seen in the [issue](https://github.com/desktop/desktop/issues/13944), this makes it so content overflowing the set height/width now scrolls. 

I have noticed this a few other instance too like in preferences if you click into the high contrast mode you get a scroll bar and I believe I have see a weird scrollbar appear on other dialogs when an error/warning is added after initial view.

This PR simply sets dialog elements `overflow: unset` so it is back they way we expected. 

Other notes: on looking at the differences in the two screenshots above it is surprising the positioning styles don't appear to interfere with anything.. but good to be aware of in case we see any other weird style changes.

## Release notes
Notes: [Fixed] Dialogs do not scroll overflow content.

### DIFF
--- a/app/styles/ui/_dialog.scss
+++ b/app/styles/ui/_dialog.scss
@@ -25,6 +25,7 @@
 // allow easy layout using generalized components and elements such as <Row>
 // and <p>.
 dialog {
+  overflow: unset;
   // These are the 24px versions of the alert and stop octicons
   // from oction v10.0.0
   //


sergiou87 (approved):
Nice catch! I tested a bunch of dialogs and didn't run into any regressions (and saw how these issues were fixed compared to prod builds).

🐸 🕴️ ❗

PR #14140, opened 2022-03-10
Fix height issues with Markdown component

sergiou87:
## Description

Right now we were adding 50px to the calculated height because depending on the content, it would still be cut off. This PR makes a few (hacky) changes in an attempt to make our Markdown component properly adjust its height to the content. I honestly can't tell why most of the things in this PR work but they all seem to be necessary to make any effect 😂 

I've tested it with a bunch of PR descriptions and comments, and haven't found any issues with the component height so far 🤞 

## Release notes

Notes: no-notes


tidy-dev (approved):
✨  Nice! Appears to work great on all the pr's I tested.

PR #14142, opened 2022-03-10
Make dark scrollbar in dark theme

### LINKED ISSUE #11174: Make changes list scrollbar consistent with the diff viewer scrollbar

The scrollbar on the diff view is dope, but the ones on the changes and history lists ain't great. Having them look consistent would certainly add to the polish of the UI.

### Proposed solution

Make all the scrollbars in the app look consistent. 
* **Changes/History**
	* The scroll bar thumb thingy for changes and history component should be made darker
	* ☝️ The active state should be even darker
	* The scroll bar track should be the same color as the one in the diff viewer
* **Diff View**
	* Add an active state to the scroll bar thumb thingy

### Additional context

The light-gray scrollbar track on the left doesn't fit with the dark theme, especially when the right scrollbar's track is dark gray.

Screen Shot 2020-12-04 at 2 13 19 PM

I noticed these classes control Windows behavior, but I couldn't find any macOS specific styles in the ten mins I was willing to dedicate to this.

https://github.com/desktop/desktop/blob/6e51ed68797cacb8c6ca405e46b21cf7d8cf5df5/app%2Fstyles%2Fthemes%2F_dark.scss#L39-L53

#### Diff viewer
Notice how there isn't an active/hover state?

### LINKED ISSUE #14144: White scrollbar when dark mode is active
### Describe the bug
The scrollbar in the list of changes stays white, although GH Desktop and my OS are in dark mode.

### Version & OS
2.9.12-beta1 & MacOS 12.2.1

### Steps to reproduce the behaviour
1. Switch GH Desktop to dark mode

### Expected behaviour
The whole UI (including all scrollbars) turns dark

### Actual behaviour
The whole UI, except the scrollbar in the list of changed files, turns dark

### Screenshots
Screen Shot 2022-03-10 at 21 50 35 PM

### Logs
[2022-03-10.desktop.beta.log](https://github.com/desktop/desktop/files/8227139/2022-03-10.desktop.beta.log)

### Additional context
It seems to be irrelevant, if GitHub follows the OS theme, or is set to dark mode by hand

rakleed:
Closes #11174
Closes #14144

## Description

- Make a dark scrollbar in a dark theme. Although the implementation method is not the best, since the https://github.com/desktop/desktop/blob/development/app/styles/themes/_dark.scss file contains CSS variables, judging by its description.

### Screenshots

<details><summary>Before</summary>

Screenshot 2022-03-10 at 20 51 45

</details>

<details><summary>After</summary>

Screenshot 2022-03-10 at 20 51 02

</details>

## Release notes

Notes: [Improved] Render native elements such as scrollbars and checkboxes in dark or light mode depending on theme where supported

### DIFF
--- a/app/src/ui/app-theme.tsx
+++ b/app/src/ui/app-theme.tsx
@@ -63,7 +63,15 @@ export class AppTheme extends React.PureComponent<IAppThemeProps> {
     ) {
       this.clearThemes()
       body.classList.add(newThemeClassName)
+      this.updateColorScheme()
     }
+  }
+
+  private updateColorScheme = () => {
+    const isDarkTheme = document.body.classList.contains('theme-dark')
+    const rootStyle = document.documentElement.style
+
+    rootStyle.colorScheme = isDarkTheme ? 'dark' : 'light'
   }
 
   /**
@@ -102,6 +110,7 @@ export class AppTheme extends React.PureComponent<IAppThemeProps> {
     styles.appendChild(document.createTextNode(customThemeStyles))
 
     body.appendChild(styles)
+    this.updateColorScheme()
   }
 
   private clearThemes() {
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -10,6 +10,10 @@
 // See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
 // https://www.electronjs.org/docs/api/native-theme#nativethemethemesource
 
+:root {
+  color-scheme: dark;
+}
+
 body.theme-dark {
   --color-new: #{$green};
   --color-deleted: #{$red};
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -10,10 +10,6 @@
 // See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
 // https://www.electronjs.org/docs/api/native-theme#nativethemethemesource
 
-:root {
-  color-scheme: dark;
-}
-
 body.theme-dark {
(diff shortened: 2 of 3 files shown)


niik (dismissed):
Thanks so much for this @rakleed, today I learned about color-scheme! I could have sworn I looked for something like this when I did the dark theme but I see that support for it was added to Chromium very recently so maybe that's why.

Could you add the same approach to our `_high-contrast.scss` theme please? That theme is still in beta but it's based on the dark theme.

rakleed:
> Could you add the same approach to our `_high-contrast.scss` theme please? That theme is still in beta but it's based on the dark theme.

Done. But I was unable to enable the high contrast theme in GitHub Desktop on macOS to test the change.

niik:
Thanks so much for this @rakleed, this is exactly what we need. I did realize that we can't set color scheme on the root element in the theme CSS directly since the dark stylesheet is always loaded. Doing so would mean that we'd render dark scrollbars, checkboxes, etc even in light mode.

I took the liberty of updating your PR to set the `color-scheme` property dynamically as part of the app's theme update logic. I'll ask one of my colleagues to take another look here now that I've made changes but I'm really excited about gettitng this in.

@tidy-dev @sergiou87 Removing myself from this PR since I've meddled with it. This PR will ensure that we render not only scrollbars properly according to the theme but also elements that we haven't styled specifically such as checkboxes, radiobuttons, etc.

image

image

tidy-dev (approved):
✨  Makes sense and tested out well!

rakleed:
@tidy-dev @niik I did not find the source code of the https://desktop.github.com site in the public domain. Can you also add the same property on this site as in my PR, or send such a request to the site developers?

PR #14148, opened 2022-03-11
Custom formatter for "precise" durations

niik:
## Description

Following up on #14123, this PR introduces a custom formatter for "precise" (that term is lifted directly from .com's code dealing with check run duration formatting) to replace moment when formatting check run durations.

## Release notes

Notes: no-notes

PR #14149, opened 2022-03-11
Replace moment relative time formatting with custom formatRelative

niik:
## Description

Based on #14148 this PR replaces all of our uses of moment's relative time formatting in favor of a custom (based on GitHub.com's [relative-time.ts](https://github.com/github/time-elements/blob/428b02c98d1b6f6743acdfae27a32dccd52b866a/src/relative-time.ts) functions) relative time formatter.

One noteworthy, user facing change, in this PR is the removal of our ultra-compact (or so called "abbreviated") relative time format in favor of the regular format

#### Before
image

#### After
image

This is in line with what GitHub.com does these days. Commits, PR lists, etc, all use the regular format instead of the micro format.

With this we're able to unship `moment-duration-format` and we're inches away from unshipping all of moment

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -41,7 +41,6 @@
     "mem": "^4.3.0",
     "memoize-one": "^4.0.3",
     "moment": "^2.24.0",
-    "moment-duration-format": "^2.3.2",
     "mri": "^1.1.0",
     "p-limit": "^2.2.0",
     "primer-support": "^4.0.0",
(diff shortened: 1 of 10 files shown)


tidy-dev (dismissed):
:sparkles: Looks good and change in dates makes sense!

niik:
@tidy-dev I ran into a yarn.lock merge conflict here thanks to #14153, could I get another ✅ please?

tidy-dev (approved):

PR #14152, opened 2022-03-11
Unship moment.js from the renderer bundle

niik:
## Description

Based on #14149, this PR removes the last callsites using moment() and removes our dependency on the same. Unfortunately that doesn't mean we've unshipped moment completely. We still depend on packages which in turn depend on moment (stop-build in / and winston-daily-rotate-file in /app). We'll address those two separately though.

## Release notes

Notes: no-notes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -40,7 +40,6 @@
     "marked": "^4.0.10",
     "mem": "^4.3.0",
     "memoize-one": "^4.0.3",
-    "moment": "^2.24.0",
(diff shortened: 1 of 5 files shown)


tidy-dev (approved):
✨

niik:
@tidy-dev Follow-on merge conflict from #14149 here, another ✅ please?

tidy-dev (approved):

PR #14153, opened 2022-03-11
Remove stop-build in favor of running Git directly

niik:
## Description

I found this during my attempts to remove moment.js. Turns out that our `stop-build` dependency depends on not only moment but tons and tons of other packages (see yarn.lock o_O). `stop-build` ensures that we don't introduce working directory changes during CI builds (which could for example be indicative of an out-of-sync yarn.lock). It's purpose is solely to check for working directory changes. It does this using ggit

https://github.com/bahmutov/stop-build/blob/6bb9eac0524b63a8cb1ba745308683002e9eb5b4/src/index.js#L31

Which in turn calls `git diff --name-status`

https://github.com/bahmutov/ggit/blob/2e85b0316d9c272e067dca5679d39e302f4cb9cd/src/changed-files.js#L29

We should be able to cal this directory with the `--exit-code` flag to have the build fail.

## Release notes

Notes: no-notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,8 @@ jobs:
         run: yarn lint
       - name: Validate changelog
         run: yarn validate-changelog
+      - name: Dirty up the working directory for testing
+        run: echo ohaimark > README.md
       - name: Ensure a clean working directory
         run: git diff --name-status --exit-code
       - name: Build production app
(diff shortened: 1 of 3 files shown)


sergiou87 (approved):
🤯 :shipit: 🔥 🔥 🔥 🔥

tidy-dev (approved):
😮

PR #14154, opened 2022-03-11
Remove winston-daily-rotate-file in favor of custom implementation

niik:
## Description

As noted in #14152 our attempts to remove `moment.js` got thwarted at the last minute when we realized that `winston-daily-rotate-file` depended on it.

This is an attempt to rip out `winston-daily-rotate-file` in favor of a custom file rotator.

N.B if this lands there's very little of winston's logic left so it's worth thinking about simplifying even further

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/desktop-file-transport.ts
+++ b/app/src/main-process/desktop-file-transport.ts
@@ -6,7 +6,6 @@ import { EOL } from 'os'
 import { readdir, unlink } from 'fs/promises'
 import { escapeRegExp } from '../lib/helpers/regex'
 import { offsetFromNow } from '../lib/offset-from'
-import { omit } from 'lodash'
 
 type DesktopFileTransportOptions = TransportStreamOptions & {
   readonly logDirectory: string
@@ -39,8 +38,9 @@ export class DesktopFileTransport extends TransportStream {
   private logDirectory: string
 
   public constructor(opts: DesktopFileTransportOptions) {
-    super(omit(opts, 'logDirectory'))
-    this.logDirectory = opts.logDirectory
+    const { logDirectory, ...rest } = opts
+    super(rest)
+    this.logDirectory = logDirectory
   }
 
   public async log(info: any, callback: () => void) {
--- a/app/src/main-process/desktop-file-transport.ts
+++ b/app/src/main-process/desktop-file-transport.ts
@@ -13,7 +13,9 @@ type DesktopFileTransportOptions = TransportStreamOptions & {
 }
 
 const fileSuffix = `.desktop.${__RELEASE_CHANNEL__}.log`
-const pathRe = new RegExp('(d{4}-d{2}-d{2})' + escapeRegExp(fileSuffix) + '$')
+const pathRe = new RegExp(
+  '(\\d{4}-\\d{2}-\\d{2})' + escapeRegExp(fileSuffix) + '$'
+)
 
 /**
  * A re-implementation of the winston-daily-rotate-file module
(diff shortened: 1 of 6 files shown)


app/src/main-process/desktop-file-transport.ts
```diff
@@ -0,0 +1,94 @@
...
+    const draining: boolean = s.write(chunk, e =>
+      e ? reject(e) : resolve(draining)
+    )
+  })
+
+const getFilePrefix = (d = new Date()) => d.toISOString().split('T', 1)[0]
+const getFilePath = (p: string) => join(p, `${getFilePrefix()}${fileSuffix}`)
+
+const pruneDirectory = async (p: string) => {
+  const treshold = offsetFromNow(-14, 'days')
```
tidy-dev:
```suggestion
  const threshold = offsetFromNow(-14, 'days')
```

app/src/main-process/desktop-file-transport.ts
```diff
@@ -0,0 +1,90 @@
...
+  return undefined
+}
+
+/**
+ * A re-implementation of the winston-daily-rotate-file module
+ *
+ * winston-daily-rotate-file depends on moment.js which we're trying to unship
+ * so we're using this instead.
+ *
+ * Please note that this is in no way a general purpose transpot like
```
sergiou87: Nit
```suggestion
 * Please note that this is in no way a general purpose transport like
```

app/src/main-process/desktop-file-transport.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  for (const f of files ?? []) {
+    const m = pathRe.exec(f)
+    const d = m ? Date.parse(m[1]) : NaN
+
+    if (!isNaN(d) && d < threshold) {
+      await unlink(join(p, f)).catch(error('unlink'))
+    }
+  }
+}
```
sergiou87: Unless I'm missing something, I think this introduces an important change in behavior: the previous implementation kept up to 14 log files, no matter how old they were, while this one deletes log files older than 14 days, no matter how many there are. So if I install and open the app today, generate a log file, and then don't open the app in 15 days… the next day I open it, my first log file will be deleted.

Not the end of the world, but I think it's better to have as much info as possible when users report issues. Who knows if there's something valuable in older logs… 😅

app/test/unit/desktop-file-transport-test.ts
```diff
@@ -0,0 +1,75 @@
...
+    try {
+      expect(await readdir(d)).toBeArrayOfSize(0)
+      for (let i = 1; i < 20; i++) {
+        const date = `${i}`.padStart(2, '0')
+        globalDate.now = () => Date.parse(`2022-03-${date}T10:00:00.000Z`)
+        globalDate.prototype.toISOString = () => `2022-03-${date}T10:00:00.000Z`
+        await info(t, 'heyo')
+      }
+
+      expect(await readdir(d)).toBeArrayOfSize(14)
```
sergiou87: Maybe worth also checking that it deletes the older files, and not the newer ones, as that's critical behavior?

sergiou87 (commented):
Awesome work!!! I think it all makes sense, and after a thorough review, I don't see any obvious flaws 🤞 

I only noticed a change in behavior that I think is worth reverting (pruning logs older than 14 days vs pruning older logs when the user has more than 14), but other than that… 🤩

app/test/unit/desktop-file-transport-test.ts
niik: Great point!

app/src/main-process/desktop-file-transport.ts
niik: I would argue that retaining based on age is a better approach here, we rarely think of retention in any other metric than time. That said I think it's important to not introduce more change than we have to here given the delicate nature of this logic so I've changed it to retain 14 files, regardless of age.

sergiou87 (approved):
Looks great! I will play a bit with it tomorrow 😄  But I'll approve it now anyway in case you're in a hurry to merge it.

sergiou87 (dismissed):
Tested and seems to work as expected 🤞  :shipit: ❗

sergiou87 (approved):
Approving again after resolving conflicts…

PR #14168, opened 2022-03-14
Custom Video Markdown filters

tidy-dev:
## Description
This pr adds two markdown filters for managing video tags. 
1) The video tag filter looks for video tags and, if their src does not look like a GitHub user asset video url, then we remove the video tag.
2) The video link tag looks for where a user may have pasted a GitHub user asset url into markdown. In this case, the markdown parser will autotag the link and put it in a paragraph tag. This filter searches for that pattern and replaces it with a video tag.

Additionally, this pr added a css style to max-width a video element so it doesn't extend out of view of scrollable area.

This filter deviates from dotcom in that dotcom verifies the GitHub user asset url actually has an asset. For performance (not having to hit api), this filter looks for things that start with "https://user-images.githubusercontent.com" and end with a video extension. This means we could have a false positive which will result in a video tag that does not render anything, where dotcom would simply show the invalid video link. This would be unlikely (probably from user copy/paste error) and still must have a the safe github cdn host so users could not use this to view externally hosted videos.

## Release notes
Notes: [Improved] Sandbox markdown parser parses video tag to remove externally linked videos and video url links to display valid uploaded videos.


app/src/lib/markdown-filters/node-filter.ts
```diff
@@ -4,6 +4,8 @@ import { EmojiFilter } from './emoji-filter'
 import { IssueLinkFilter } from './issue-link-filter'
 import { IssueMentionFilter } from './issue-mention-filter'
 import { MentionFilter } from './mention-filter'
+import { VideoLinkFilter } from './video-link-filter'
```
sihamitas40, line 7: استيراد  {  VideoLinkFilter  }  من  "./video-link-filter"

app/src/lib/markdown-filters/video-link-filter.ts
```diff
@@ -0,0 +1,82 @@
...
+  public createFilterTreeWalker(doc: Document): TreeWalker {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const filter = this
+    return doc.createTreeWalker(doc, NodeFilter.SHOW_ELEMENT, {
+      acceptNode: function (el: Element) {
+        return filter.getGithubVideoLink(el) === null
+          ? NodeFilter.FILTER_SKIP
+          : NodeFilter.FILTER_ACCEPT
+      },
+    })
```
niik: I'd prefer if we used a bound method here instead of disabling the eslint rule. We try to avoid capturing `this` if possible

```suggestion
    return doc.createTreeWalker(doc, NodeFilter.SHOW_ELEMENT, {
      acceptNode: (el: Element) =>
        this.getGithubVideoLink(el) === null
          ? NodeFilter.FILTER_SKIP
          : NodeFilter.FILTER_ACCEPT,
    })
```

app/src/lib/markdown-filters/video-link-filter.ts
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * Example:
+   * <p>
+   *  <a href="https://user-images.githubusercontent.com/7559041/1234.mp4">
+   *    https://user-images.githubusercontent.com/7559041/1234.mp4
+   *  </a>
+   * </p>
+   *
+   * Output = [
+   * <videosrc="https://user-images.githubusercontent.com/7559041/1234.mp4"></video>
```
niik:
```suggestion
   * <video src="https://user-images.githubusercontent.com/7559041/1234.mp4"></video>
```

app/src/lib/markdown-filters/video-tag-filter.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * If the url for the src does NOT match the pattern of a github user asset, we
+ * remove the video tag.
+ */
+export class VideoTagFilter implements INodeFilter {
+  /**
+   * Video link filter matches on video tags that src does not match a github user asset url.
+   */
+  public createFilterTreeWalker(doc: Document): TreeWalker {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
```
niik: Doesn't seem to be necessary here?

```suggestion
```

app/src/lib/markdown-filters/video-tag-filter.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Takes a video element who's src host is not a github user asset url and removes it.
+   */
+  public async filter(node: Node): Promise<ReadonlyArray<Node> | null> {
+    if (
+      !(node instanceof HTMLVideoElement) ||
+      githubAssetVideoRegex.test(node.src)
+    ) {
+      // If it is video element with a valid source, we return null to leave it alone.
+      // This is different then dotcom that regenerates a video tag because it
```
niik:
```suggestion
      // This is different than dotcom which regenerates a video tag because it
```

app/src/lib/markdown-filters/video-url-regex.ts
```diff
@@ -0,0 +1,14 @@
...
+const user_images_cdn_url = 'https://user-images.githubusercontent.com'
+
+// List of common video formats obtained from
+// https://developer.mozilla.org/en-US/docs/https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs/Media/Formats/Video_codecs
+// The MP4, WebM, and Ogg formats are supported by HTML standard.
+const videoExtensionRegex = /(mp4|webm|ogg|mov|qt|avi|wmv|3gp|mpg|mpeg|)$/
+
+/** Regex for checking if a url is a github asset cdn video url */
+export const githubAssetVideoRegex = new RegExp(
+  escapeRegExp(user_images_cdn_url) + '.+' + videoExtensionRegex.source,
```
niik: This needs to be anchored to the start or else a malicious user could craft a url like `https://evil.com/https://user-images.githubusercontent.com/evil.mp4`

```suggestion
  '^' + escapeRegExp(user_images_cdn_url) + '.+' + videoExtensionRegex.source,
```

niik (changes requested):
Just a few requests here @tidy-dev

app/src/lib/markdown-filters/video-url-regex.ts
tidy-dev: 🤦  Thank you!

app/src/lib/markdown-filters/video-url-regex.ts
```diff
@@ -0,0 +1,14 @@
+import { escapeRegExp } from 'lodash'
```
niik: Can we use the `escapeRegExp` function from `regex.ts` instead? I'd rather we made a conscious decision to move over to lodash's method across the code base than to mix

```suggestion
import { escapeRegExp } from '../helpers/regex'
```

niik (changes requested):
Just a tiny tiny nit here @tidy-dev

niik (approved):

PR #14173, opened 2022-03-15
Formatting of relative and absolute dates using Intl.* APIs

niik:
Closes #13943 

## Description

This adds ~~region-aware~~ formatting for relative and absolute dates using the [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) namespace. This builds on #14149, #14123 and [app.getLocaleCountryCode](https://www.electronjs.org/docs/latest/api/app#appgetlocalecountrycode)

Note that this is not about localizing strings, the interface will remain US-english, ~~it will however apply the correct (hopefully) regional settings for formatting numbers. Most notably right now this means that users like myself who have their OS configured to display 24-hour format will now.~~

**Note that this is feature-flagged and currently only enabled for dev**.

cc @tsvetilian-ty as author of #13965 

### Screenshots

image

For reference these are my regional settings on Windows and macOS (they resolve to `en-SE` on both platforms)

image

image

## Release notes

Notes:

PR #14174, opened 2022-03-15
Number formatting using Intl.* APIs

### LINKED ISSUE #1245: Format the ahead/behind count in a locale-aware manner

niik:
Closes #1245

## Description

~~Back in 2017 I wrote *"I'm leaning towards trying to solve this properly."* in response to our first attempt  at fixing #1245 (https://github.com/desktop/desktop/pull/1257). Now, 5-ish years later technology has finally caught up to where we're able to do this with a relatively small amount of support logic :tada:~~

**Turns out ☝️ was a little bit premature, see https://github.com/desktop/desktop/pull/14173#issuecomment-1072293321**

This builds upon #14173 and adds region-aware formatting of numbers.

I also took this opportunity to render the number of commits ahead/behind in a compact format (i.e number > 1000 will render as 1K). Hovering will get you the exact number but if you're behind by 1000 or more you likely don't care about the exact number.

There's a good chance I've missed a number of (pun intended) places where we present numbers to the user but that's okay, we can fix that incrementally.

### Screenshots

| Before | en-US | en-SE |
|--------|-------|-------|
|  image | image | image |
| image | image | image |

## Release notes

Notes: no-notes


app/src/lib/format-number.ts
```diff
@@ -0,0 +1,37 @@
+import mem from 'mem'
+import QuickLRU from 'quick-lru'
+import { getFormattingLocales } from './formatting-locale'
+
+// Initializing a nuumber formatter is expensive but formatting is relatively
```
tidy-dev:
```suggestion
// Initializing a number formatter is expensive but formatting is relatively
```

tidy-dev (changes requested):
Just a little spelling nit. But, also, a place missed is line 654 of `compare.tsx` for the squashing dialog. 

As for it working, I switched my region to Sweden, Spain, and France. It worked as expected for Sweden, but not for Spain or France. But, I figure the is the same issue that Sergio mentioned in slack so I am assuming not something for this PR.

niik:
So this will also be affected by https://github.com/desktop/desktop/pull/14173#issuecomment-1072293321.

Given that we're switching to `en-US-POSIX` by default there will be no number grouping for now.

@sergiou87 @tidy-dev One thing that will change with this PR though is the compact format of commits > 1000. I think that's okay though even though that'll use `.` as the decimal separator. Our plan is definitely to try and support non-US region formats soon and the number of users pushing/pulling more than 1000 commits should be relatively low.

tidy-dev (approved):
:sparkles: Makes sense. :)

Eminealtug:
.

PR #14185, opened 2022-03-16
Fix react_dev_tools

### LINKED ISSUE #14183: There is no React Dev Tool when developing the app
### Describe the bug
There's no React Dev Tool in the inspect panel. Looking at the code it should have though, so I'm reporting this as a bug.

### Version & OS

I'm on the latest development branch, using MacBookPro with M1 Max.

1. yarn
2. yarn build:dev
3. yarn start

### Expected behavior

There should be a React Component Tab

image

### Actual behavior

There is not
image

### Logs

While there's an error in the log saying the extensions encounter an error: `Cannot read properties of undefined (reading 'instance')`
The error goes away once I upgrade electron to `"electron": "^17.1.2",`

Here the log after the upgrade:
[2022-03-16.desktop.development.log](https://github.com/desktop/desktop/files/8259635/2022-03-16.desktop.development.log)
- There are no errors now, but the app still doesn't have React Dev Tool

I even tried manually loading the dev tools as suggested here:
https://www.electronjs.org/docs/latest/tutorial/devtools-extension

Still no Dev Tool, and also no error/warning in the log

yuchien302:
Closes #14183

## Description
React Dev Tools is not showing up in the development build

### Screenshots

Check that React Dev Tools shows up:
image

### Notes

This is just for you guys to reference and make a proper PR. 

I can't get `yarn build:dev` to succeed unless I delete a whole bunch of code that requires registry-js and desktop-notifications.js

I believe the right fix is to just update `electron`, `electron-debug`, and `electron-devtools-installer` to latest, and add the following option:
image

## Release notes

Notes: no-notes


niik:
> This is just for you guys to reference and make a proper PR.

@yuchien302 Opening a PR with lots of changes only creates additional work for us maintainers. React devtools has been broken in Desktop dev environments for a long time but it's not a priority for us to fix. If you'd like to open a PR which actually fixes it we'll consider it. If not you can fill out an issue and include as much detail as possible.

PR #14187, opened 2022-03-16
Fix `Not Now` button in "initialize LFS" dialog

sergiou87:
## Description

With #14131 I made the dialog non-dismissable, but the `Not Now` button relied on the `onDismiss` callback. This PR changes that by using `onCancelButtonClick` instead of `onDismiss`.

## Release notes

Notes: no-notes (regression only in beta)


tidy-dev (approved):
✨  Looks great!

PR #14189, opened 2022-03-16
Fix permission issue when loading extension

### LINKED ISSUE #14183: There is no React Dev Tool when developing the app

tsvetilian-ty:
Closes #14183

## Description
- Update `electron-devtools-installer` dependency
- Applied allowFileAccess option to `installExtension()`

### Screenshots
image

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -66,7 +66,7 @@
   "devDependencies": {
     "devtron": "^1.4.0",
     "electron-debug": "^3.1.0",
-    "electron-devtools-installer": "^3.1.1",
+    "electron-devtools-installer": "^3.2.0",
     "temp": "^0.8.3",
     "webpack-hot-middleware": "^2.10.0"
   }
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -703,7 +703,9 @@ function createWindow() {
 
     for (const extension of extensions) {
       try {
-        installExtension(extension)
+        installExtension(extension, {
+          loadExtensionOptions: { allowFileAccess: true },
+        })
       } catch (e) {}
     }
   }
--- a/app/yarn.lock
+++ b/app/yarn.lock
@@ -459,13 +459,14 @@ electron-debug@^3.1.0:
     electron-is-dev "^1.1.0"
     electron-localshortcut "^3.1.0"
 
-electron-devtools-installer@^3.1.1:
-  version "3.1.1"
-  resolved "https://registry.yarnpkg.com/electron-devtools-installer/-/electron-devtools-installer-3.1.1.tgz#7b56c8c86475c5e4e10de6917d150c53c9ceb55e"
-  integrity sha512-g2D4J6APbpsiIcnLkFMyKZ6bOpEJ0Ltcc2m66F7oKUymyGAt628OWeU9nRZoh1cNmUs/a6Cls2UfOmsZtE496Q==
+electron-devtools-installer@^3.2.0:
+  version "3.2.0"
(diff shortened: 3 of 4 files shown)


tsvetilian-ty:
`electron-devtools-installer` uses additional dependency to [microsoft/tslib](https://github.com/microsoft/tslib), which uses [BSD Zero Clause License](https://github.com/microsoft/tslib/blob/main/LICENSE.txt), but [atom/legal-eagle](https://github.com/atom/legal-eagle) doesn't support 0BSD as permissive licenses.

tsvetilian-ty:
Update: Opened PR (atom/legal-eagle#19) in [atom/legal-eagle](https://github.com/atom/legal-eagle) to add Zero-Clause BSD as permissive license

sergiou87 (changes requested):
Thanks for this @tsvetilian-ty!! ❤️ 

I just checked about the tslib license, and I think you don't need to update legal-eagle for this. Instead, you can change the `script/licenses/license-overrides.ts` file to add a new entry at the bottom for `tslib@2.3.1` (you'll notice we already have one for 2.0.0, which can be removed).

You can probably create an object with the license data and reuse it, something like…
```typescript
const tslibLicenseOverride = {
  repository: 'git+https://github.com/microsoft/tslib',
  license: 'BSD',
  source:
    'https://github.com/microsoft/tslib/blob/74f1ff794985b33657b599b611e82937e3074617/LICENSE.txt',
  sourceText: `Copyright (c) Microsoft Corporation.

Permission to use, copy, modify, and/or distribute this software for any
purpose with or without fee is hereby granted.

THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
PERFORMANCE OF THIS SOFTWARE.`,
}
...
  'tslib@2.0.0': tslibLicenseOverride,
  'tslib@2.3.1': tslibLicenseOverride,
```

tsvetilian-ty:
@sergiou87 I didn't catch that there is an option for license override 😵‍💫 😄

sergiou87 (approved):
Looks good and works like a charm, thank you!! 🙇‍♂️

PR #14192, opened 2022-03-16
Persist GUID in file instead of LocalStorage, and send it with crashes from main process

sergiou87:
## Description

Up until now, we sent a GUID along with errors reported to Sentry. However, that GUID was generated in the renderer process, persisted across sessions in the LocalStorage (only accessible from the renderer process), and therefore never sent along with errors reported from the main process.

This PR changes that so the GUID is generated from the main process and persisted in a `.guid` file, so that the main process can restore it when the app is launched again. Access to this GUID by the renderer process is done via IPC to prevent concurrent access to the `.guid` file. Only the main process will access this file.

Also, in order to keep around GUIDs generated from the renderer process of existing users, the app will look at launch for a GUID in the Local Storage and persist it in the `.guid` file.

## Release notes

Notes: no-notes


app/src/ui/index.tsx
```diff
@@ -132,7 +134,6 @@ const sendErrorWithContext = (
   } else {
     const extra: Record<string, string> = {
       osVersion: getOS(),
-      guid: getGUID(),
```
sergiou87, line 135: I moved this to the main process.

sergiou87:
I tested this using 2.9.12-test1 and saw all the errors reported correctly, from both the renderer and main process 🥲  and it also preserves the existing guid from LocalStorage if any. I also checked and the guid is properly reported in stats too 👍

app/src/lib/get-main-guid.ts
```diff
@@ -0,0 +1,45 @@
...
+let cachedGUID: string | null = null
+
+/** Get the GUID for the Main process. */
+export async function getMainGUID(): Promise<string> {
+  if (!cachedGUID) {
+    cachedGUID = await readGUIDFile()
+  }
+
+  return cachedGUID
+}
```
niik: There's two approaches we can take here to clean this up a little. Either we leverage the new `??=` opertor

```ts
let cachedGUID: string | null = null

/** Get the GUID for the Main process. */
export const getMainGUID = async () => (cachedGUID ??= await readGUIDFile())
```

I did note, however that our version of prettier doesn't understand logical assignment operators. I'll fix that separately. We could also leverage memoizeOne in which case we don't need the cachedGUID at all.

```suggestion
import memoizeOne from 'memoize-one'

/** Get the GUID for the Main process. */
export const getMainGUID = memoizeOne(readGUIDFile)
```

The `memoizeOne` method will cache errors as well though whereas the previous method will keep retrying `readGUIDFile()` on each call but it looks like `readGUIDFile` isn't intended to throw ever?

app/src/lib/get-main-guid.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (guid === undefined) {
+    guid = uuid()
+    await saveGUIDFile(guid)
+  }
+
+  return guid
+}
+
+/** Saves the GUID to the .guid file */
+export async function saveGUIDFile(guid: string) {
```
niik, line 43: Doesn't this need to set/unset the cached value as well?

app/src/lib/get-main-guid.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  return cachedGUID
+}
+
+/** Reads the persisted GUID from the .guid file or generates a new one */
+export async function readGUIDFile(): Promise<string> {
+  let guid = undefined
+
+  try {
+    guid = await readFile(getGUIDPath(), 'utf8')
```
niik: Can we trim the read contents first? Just so that a newline introduced by curious folks opening the file and saving it doesn't invalidate it?

niik (changes requested):

app/src/lib/get-main-guid.ts
sergiou87: I love that `memoizeOne` approach! As you say, `readGUIDFile` isn't supposed to throw errors (happy change that if you have a good reason for it, of course).

app/src/lib/get-main-guid.ts
sergiou87: I don't think it's necessary, but while writing the explanation about it, I realized the renderer should be getting the GUID from `getMainGUID` (via IPC, of course) instead of using `readGUID` 🤦  that way the renderer will use whatever the main process had cached or will populate its cache at the same time.

In any case… the reason why I think it's not necessary is because `saveGUIDFile` is used:
1. When the main process tries to populate its cache, so the cached value is already set by `getMainGUID`
2. When the renderer process migrates an existing GUID from the LocalStorage. In that case, the main process would only have a GUID if it crashed before the migration. If the main process didn't crash (which is the only way to get to the migration step?), it shouldn't have anything cached and all should be good.

Does that make sense? 😳

sergiou87: Actually, maybe it makes sense since I'm caching the migrated GUID on the renderer process even when persistence fails. I'm gonna change logic a bit:
1. Don't remove existing GUID from the LocalStorage if persistence failed.
2. Update cached GUID in main process on migration

sergiou87:
@niik I think this is ready for a second pass, I made the following changes:
- If there is a GUID to migrate from the renderer process…
   - …it won't be deleted from the LocalStorage unless the main process succeeded to persist it to `.guid`.
   - …it will be cached on the main process despite of whether persisting it to `.guid` succeeds or not.
- The contents of `.guid` are trimmed before being cached.
- The renderer process will use/populate the main process cache when `getRendererGUID` is invoked.

I've tested these extensively with 2.9.12-test2, including creating a `.guid` file owned by `root` (to cause write issues), and all scenarios worked as expected 🤞

niik (approved):
:shipit:

PR #14196, opened 2022-03-17
Fix Markdown height issue (round 2)

sergiou87:
## Description

#14140 attempted to fix the issue in a hacky i-dont-know-what-im-doing.gif way, and it seemed to work for a bunch of PRs/comments I ran into, but not for all.

This PR presents a more deterministic attempt to fix the problem: I found that part of the CSS was wrong (probably due to a refactor or to wrapping stuff into the `<div class="content">` element). I also added a `ResizeObserver` to react to whenever the `#content` div is resized due to content changes.

This approach seems to work in many situations but I see it still presents a scrollbar in other like #13201 , I will investigate…

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/sandboxed-markdown.tsx
+++ b/app/src/ui/lib/sandboxed-markdown.tsx
@@ -100,6 +100,10 @@ export class SandboxedMarkdown extends React.PureComponent<
     }
   }
 
+  public componentWillUnmount() {
+    this.resizeObserver.disconnect()
+  }
+
   /**
    * Since iframe styles are isolated from the rest of the app, we have a
    * markdown.css file that we added to app/static directory that we can read in
@@ -171,6 +175,7 @@ export class SandboxedMarkdown extends React.PureComponent<
     ) as HTMLDivElement
 
     if (this.contentDivRef !== null) {
+      this.resizeObserver.disconnect()
       this.resizeObserver.observe(this.contentDivRef)
     }
   }
--- a/app/src/ui/lib/sandboxed-markdown.tsx
+++ b/app/src/ui/lib/sandboxed-markdown.tsx
@@ -44,6 +44,36 @@ export class SandboxedMarkdown extends React.PureComponent<
 > {
   private frameRef: HTMLIFrameElement | null = null
   private frameContainingDivRef: HTMLDivElement | null = null
+  private contentDivRef: HTMLDivElement | null = null
+
+  /**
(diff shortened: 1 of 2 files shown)


app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -44,6 +44,36 @@ export class SandboxedMarkdown extends React.PureComponent<
...
+   * Resize observer used for tracking height changes in the markdown
+   * content and update the size of the iframe container.
+   */
+  private readonly resizeObserver: ResizeObserver
+  private resizeDebounceId: number | null = null
+
+  public constructor(props: ISandboxedMarkdownProps) {
+    super(props)
+
+    this.resizeObserver = new ResizeObserver(this.scheduleResizeEvent)
```
niik, line 59: Just a drive-by comment noting that it doesn't look like we `.disconnect` or `.unobserve` on this ResizeObserver. We should at the very least call `.disconnect` when this component unmounts. We could also `.unobserve(contentDivRef)` in `setupContentDivRef` if `contentDivRef` isn't null prior to assigning it.

Haven't reviewed anything else in here tho

tidy-dev:
> This approach seems to work in many situations but I see it still presents a scrollbar in other like https://github.com/desktop/desktop/pull/13201 , I will investigate

https://user-images.githubusercontent.com/75402236/159015809-8a5b0994-dd60-4c8c-9f12-951a05f4e8c4.mov

Just note that I think the `code` pre tag may need some css adjustments to match dotcom behavior of showing a scrollbar individually instead of making the whole area scroll.

^ Not saying that needs resolved as part of this PR.. just maybe not a concern inside of the issue of the vertical height problem.

tidy-dev (changes requested):
Testing it out -  it appears to work great! I definitely like this approach better than the previous "solutions". :D 

Would approve - but I think Markus's drive by comment needs addressed. :)

app/src/ui/lib/sandboxed-markdown.tsx
sergiou87: Done in [6a967a7](https://github.com/desktop/desktop/pull/14196/commits/6a967a7af74b45947cf0a379ec217bd2d937c73e), thanks!!

tidy-dev (approved):
:sparkles: Thank you for delving into this further and figuring out a better way!

PR #14070, opened 2022-03-03
Linted the files according to the .prettierrc.yml

Harry-Hopkinson:
Closes #N/A

## Description

Notes:
prettier --config ./.prettierrc.yml  --write "./**/*.ts"
prettier --config ./.prettierrc.yml  --write "./**/*.tsx"

PR #14199, opened 2022-03-17
Update menu item to allow quick merge of upstream

### LINKED ISSUE #14145: Add option to quickly merge upstream branch into current branch

There are many times I want to merge the upstream default branch into my current branch, such as when working in a PR with conflicts. 

Github Desktop currently offers a quick way to merge the default branch, but this branch refers to your fork, instead of the upstream repo:
image

### Proposed solution

When the user's fork behaviour settings is "to contribute to the parent repository", this menu item should offer to merge from the upstream repo, instead of your fork.

uttiya10:
Closes #14145 

## Description

If the user is contributing to the parent repo, we now choose the upstream's default branch as the merging branch rather then the fork's default branch.

### Screenshots

**BEFORE:**
image
The app suggests to merge my local fork's default branch, even though I'm contributing to the main repo.

**AFTER:**
image
The app will now suggest to merge the upstream's default branch.

## Release notes

Notes:


app/src/main-process/menu/menu-event.ts
```diff
@@ -16,7 +16,7 @@ export type MenuEvent =
   | 'show-preferences'
   | 'choose-repository'
   | 'open-working-directory'
-  | 'update-branch'
+  | 'update-branch-with-contribution-target-branch'
```
uttiya10, line 19: Theses id's seems to be duplicated in `app/src/models/menu-ids.ts`. Just curious if there's a reason for this.

app/src/models/menu-ids.ts
```diff
@@ -5,7 +5,7 @@ export type MenuIDs =
   | 'discard-all-changes'
   | 'stash-all-changes'
   | 'preferences'
-  | 'update-branch'
+  | 'update-branch-with-contribution-target-branch'
```
uttiya10, line 8: Refer to comment above about duplicated id's.

uttiya10:
Seems like the build is failing due to an outdated prettier dependency not knowing how to interpret the syntax. Updating this dependency causes **ALOT** of other files to be modified/linted, so I guess we'll have to wait for #14070 to be merged.

app/src/lib/branch.ts
```diff
@@ -41,3 +44,29 @@ export function findDefaultUpstreamBranch(
...
+  let contributionTargetDefaultBranch: Branch | null
+    if (isRepositoryWithGitHubRepository(repository)) {
+      contributionTargetDefaultBranch = findDefaultUpstreamBranch(
+        repository,
+        branchesState.allBranches
+      )
+    }
+    contributionTargetDefaultBranch ??= branchesState.defaultBranch
+    return contributionTargetDefaultBranch
+}
```
niik: > Seems like the build is failing due to an outdated prettier dependency not knowing how to interpret the syntax. Updating this dependency causes ALOT of other files to be modified/linted, so I guess we'll have to wait for https://github.com/desktop/desktop/pull/14070 to be merged.

Yeah, we're looking at that in #14205. In the meantime though I think you can pretty easily avoid the `??=` operator here. We typically prefer avoiding temporary mutable variables like that anyway (see https://github.com/desktop/desktop/blob/24d89a800755705c0e9c3f1239d3b3ac224721f5/docs/contributing/engineering-values.md#immutability-is-good)

```suggestion
export function findContributionTargetDefaultBranch(
  repository: Repository,
  { allBranches, defaultBranch }: IBranchesState
): Branch | null {
  return isRepositoryWithGitHubRepository(repository)
    ? findDefaultUpstreamBranch(repository, allBranches) ?? defaultBranch
    : defaultBranch
}
```

niik (changes requested):

app/src/lib/branch.ts
uttiya10: You can destructure objects *in the function declaration itself*??? 😮

niik (changes requested):
Hey @uttiya10, sorry for the delay. It looks like this logic is only looking at whether the current repository has a parent repository or not (i.e. whether it's a fork or not).

We also have to take into account whether the user has indicated that they want to use the fork for contributing to the parent repository or for their own purposes:

image

I believe we can just use `findDefaultUpstreamBranch` for this, it should take care of resolving the right branch for you.

uttiya10:
I am a bit confused. I am using `findDefaultUpstreamBranch`, to determine the appropriate branch based on user settings. Can you please clarify?

niik (approved):
Hey @uttiya10, another apology here for taking so long to come back to this review!

> I am a bit confused. I am using `findDefaultUpstreamBranch`, to determine the appropriate branch based on user settings. Can you please clarify?

You're absolutely right, I must have misread while reviewing. This looks perfect

> Theses id's seems to be duplicated in `app/src/models/menu-ids.ts`. Just curious if there's a reason for this.

Yeah, they're two different things but in this case they end up being duplicated. There's menu item ids (defined in menu-ids.ts) which is needed for menu items which we need to lookup at runtime (if we want to change whether it's enabled or not or if we need to change its label). Then there's ids for menu items which need to be proxied to the renderer process. There are menu items where the id differs from the event emitted (such as the `toggle-stashed-changes` menu item which emits either `hide-stashed-changes` or `show-stashed-changes`). There used to be more of a difference in the past and I'm sure an argument could be made for cleaning this up but given that it's only a minor nuisance where the type system has our backs it's far down on our list of priorities.

PR #14204, opened 2022-03-18
Add include and exclude functionality to selected files

### LINKED ISSUE #7790: Handle multiple files for check / uncheck
- currently we can only check changed files one-by-one
- i would like to select multiple file and check/uncheck them all
- clicking on checkboxes should not invalidate the current multiple selection

have all selected file be checked/unchecked when multiple files are selected
-if some mixed checkd/unchecked status for the files, sohuld get the check state from the line the user is currently checking

[x] fileA
[  ] fileB
[x] fileC

=> if user select those 3 files, and click on the checkbox of fileB, mark all as checked, but if clicking on fileA or C, uncheck all (then  check all on second click)

tsvetilian-ty:
Closes #7790

## Description
Added two new options for selected files
- Include Selected Files
- Exclude Selected Files

### Screenshots
scree-rec

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -534,6 +534,33 @@ export class ChangesList extends React.Component<
         })
       })
 
+    if (paths.length > 1) {
+      const includeFilesOptions: IMenuItem[] = [
+        { type: 'separator' },
+        {
+          label: __DARWIN__
+            ? 'Include Selected Files'
+            : 'Include selected files',
+          action: () => {
+            selectedFiles.map(file =>
+              this.props.onIncludeChanged(file.path, true)
+            )
+          },
+        },
+        {
+          label: __DARWIN__
+            ? 'Exclude Selected Files'
+            : 'Exclude selected files',
+          action: () => {
+            selectedFiles.map(file =>
+              this.props.onIncludeChanged(file.path, false)
+            )
+          },
+        },
+      ]
+      items.push(...includeFilesOptions)
+    }
+
     const enabled = status.kind !== AppFileStatusKind.Deleted
     items.push(
       { type: 'separator' },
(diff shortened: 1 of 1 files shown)


app/src/ui/changes/changes-list.tsx
```diff
@@ -534,6 +534,33 @@ export class ChangesList extends React.Component<
...
+            : 'Exclude selected files',
+          action: () => {
+            selectedFiles.map(file =>
+              this.props.onIncludeChanged(file.path, false)
+            )
+          },
+        },
+      ]
+      items.push(...includeFilesOptions)
+    }
```
tidy-dev: Just a nit, but I don't believe we gain anything by declaring the variable and then pushing it; rather than just going ahead and pushing the items. :) 

Note.. below suggestion will likely break the linter.

```suggestion
    if (paths.length > 1) {
    items.push(
        { type: 'separator' },
        {
          label: __DARWIN__
            ? 'Include Selected Files'
            : 'Include selected files',
          action: () => {
            selectedFiles.map(file =>
              this.props.onIncludeChanged(file.path, true)
            )
          },
        },
        {
          label: __DARWIN__
            ? 'Exclude Selected Files'
            : 'Exclude selected files',
          action: () => {
            selectedFiles.map(file =>
              this.props.onIncludeChanged(file.path, false)
            )
          },
        },
      )
    }
```

tidy-dev (changes requested):
Thanks for adding this!  One little nit and I think we can merged. :)

Other notes:
We had a bit of an internal discussion on whether we think this just adds context menu clutter for a cross os system convention (using space bar to toggle check boxes); but, decided to keep it since we have had quite a bit of user feedback indicating that the spacebar toggle is not well known.

tidy-dev (approved):
✨  Thanks for those changes!

zeke:
Nice!

PR #14206, opened 2022-03-18
Custom Markdown Team Mention Filter

tidy-dev:
## Description
This pr adds a team mention filter to our markdown parser.  It looks for mentions like @org/team, example`@desktop/core`, and replaces them with a link to that teams dashboard page. 

Dotcom verifies the team exists before linking, we assume they exist for performance. That leads to false positives. We can fortunately reduce the false positives because these references need only be linked if the repo is owned by the referenced org. This means that the we can compare the repo owner to the org and if they do not match, we don't link. However, this still introduces false positives for a repo not owned by organizations. For example, if in my repo I did @tidy-dev/something, it would attempt to link to an org of tidy-dev. This would just return a 404.

Notes: going to look into for a follow up PR to add an organization attribute to our repo objects to prevent linking in non-org owned repos. From a quick look, we already pull this information in the api call we do for repos.

### DIFF
--- a/app/src/lib/markdown-filters/mention-filter.ts
+++ b/app/src/lib/markdown-filters/mention-filter.ts
@@ -36,7 +36,7 @@ export class MentionFilter implements INodeFilter {
   private readonly dotsAtEndOfLine = /\.+$/
 
   // non-word character except dot or `
-  private readonly nonWordExceptDotOrBackTick = /[^0-9a-zA-Z_.`]/
+  private readonly nonWordExceptDotOrBackTickOrHyphen = /[^0-9a-zA-Z_.`-]/
 
   // Pattern used to extract @mentions from text
   // Looking for @user or @user_user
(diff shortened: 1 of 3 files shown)


app/src/lib/markdown-filters/mention-filter.ts
```diff
@@ -36,7 +36,7 @@ export class MentionFilter implements INodeFilter {
   private readonly dotsAtEndOfLine = /\.+$/
 
   // non-word character except dot or `
-  private readonly nonWordExceptDotOrBackTick = /[^0-9a-zA-Z_.`]/
+  private readonly nonWordExceptDotOrBackTickOrHyphen = /[^0-9a-zA-Z_.`-]/
```
tidy-dev, line 40: For the purposes of a username, the hyphen is a word character. This is intending to exclude things that could be in a username characters (signifying the end of username characters). I ran into it because I made an org called tidy-dev-org. Then when I referenced @tidy-dev-org/the-a-team, the @tidy-dev part would be caught and linked by our mention filter. It shouldn't because the regex is meant not to catch things that look like @something/. 
image


Also to note, I originally put the call order of the TeamMentionFilter after the Mention Filter and simply swapping this prevented this problem because the team mention was caught first. Thus, I decided to put the TeamMentionFilter first in case there were any other regex fallacies.

sergiou87 (approved):
✨

PR #14208, opened 2022-03-18
Adding organization property to github repos for team mention filter

tidy-dev:
## Description

This is based on #14206, in which I introduced a TeamMentionFilter for detecting @org/team references in markdown. Since we do not want to incur the performance hit of verifying the referenced team exists, we want to check if the markdown content owning repository is owned by the org being referenced. Currently, I am just checking that the repo owner matches the org in the ref. However a repo owner may be an organization or may be a user, thus this results in a false positive. Example if I put @tidy-dev/some-team-name, our team mention filter would link it even tho tidy-dev is not an org.

~When we retrieve a repository and it is owned by an organization, it has an optional organization property on it. This PR adds that property to the github repo's we store locally.~ Then, changes my comparison from agains the repo's owner to the repo's organization. Reducing false positives to non existent teams.

Update: changed to storing the owner account type (either 'Organization' or 'User') on the database owner and owner models.


app/src/lib/databases/repositories-database.ts
```diff
@@ -131,6 +133,13 @@ export class RepositoriesDatabase extends BaseDatabase {
 
     this.conditionalVersion(8, {}, ensureNoUndefinedParentID)
     this.conditionalVersion(9, { owners: '++id, &key' }, createOwnerKey)
+
+    this.conditionalVersion(10, {}, async tx => {
+      /**
+       * We're introducing the `organization` property on PRs in version 10.
+       */
+      tx.table('gitHubRepositories').clear()
```
tidy-dev: I am not 100% sure this is all I need since now the repository tables will reference a GitHub table id that doesn't exist. During development, I saw weird behavior of repositories disappearing from the repo list.. but I also had manually cleared the db.. so going to do some more testing on that... But, if you happen to know that I need more feel free to let me know. :D

sergiou87: I think we shouldn't do this, it definitely sounds like it'd break things as you describe 😕 

Instead, the `organization` field for existing records in the  table `gitHubRepositories` should be `null` (or `undefined`, whatever it is the default), and then we should update the info when we have a chance.

tidy-dev: Yeah. Seems updating the hash is all I needed. :) It seems there is a little bit of lag in the db updating to it taking effect in the markdown. But, I think it will just be a one time blip that won't really be noticed.

app/src/lib/databases/repositories-database.ts
```diff
@@ -33,6 +34,7 @@ export interface IDatabaseGitHubRepository {
   readonly isArchived?: boolean
 
   readonly permissions?: 'read' | 'write' | 'admin' | null
+  readonly organization?: IAPIOrganization
```
niik: We shouldn't be storing raw API objects in the database. They often contain a lot more dat than we need and we can't trust that their shape stays consistent across GHE/GHES/GHAE versions. In the case of Organization it's not too bad:

```json
"organization": {
		"login": "desktop",
		"id": 13171334,
		"node_id": "MDEyOk9yZ2FuaXphdGlvbjEzMTcxMzM0",
		"avatar_url": "https://avatars.githubusercontent.com/u/13171334?v=4",
		"gravatar_id": "",
		"url": "https://api.github.com/users/desktop",
		"html_url": "https://github.com/desktop",
		"followers_url": "https://api.github.com/users/desktop/followers",
		"following_url": "https://api.github.com/users/desktop/following{/other_user}",
		"gists_url": "https://api.github.com/users/desktop/gists{/gist_id}",
		"starred_url": "https://api.github.com/users/desktop/starred{/owner}{/repo}",
		"subscriptions_url": "https://api.github.com/users/desktop/subscriptions",
		"organizations_url": "https://api.github.com/users/desktop/orgs",
		"repos_url": "https://api.github.com/users/desktop/repos",
		"events_url": "https://api.github.com/users/desktop/events{/privacy}",
		"received_events_url": "https://api.github.com/users/desktop/received_events",
		"type": "Organization",
		"site_admin": false
	},
```

Regardless I think a better approach here is to add an optional string "type" property to `IDatabaseOwner`. If you look at the API response for say desktop/desktop you can determine that the owner is an organization by looking at `owner.type`. See

```json
{
	...
	"name": "desktop",
	"full_name": "desktop/desktop",
	...
	"owner": {
		"login": "desktop",
		...
		"type": "Organization",
		...
	},
```

It'll be a little bit more work but it'll be worth it. I recently did some work on saving owners with correct casing so I've got this pretty fresh in my memory if you want to pair on adding it.

niik (changes requested):

app/src/lib/stores/repositories-store.ts
```diff
@@ -381,15 +395,26 @@ export class RepositoriesStore extends TypedBaseStore<
...
+        endpoint,
+        login,
+        type: ownerType,
+      })
     } else {
       id = forceUnwrap('Missing owner id', existingOwner.id)
     }
 
-    return new Owner(login, endpoint, id)
+    return new Owner(login, endpoint, id, ownerType ?? existingOwner?.type)
```
tidy-dev, line 417: In the scenario of a skeleton repo being added because the repo details couldn't be retrieved, this will add an undefined owner type. The negative consequence is that if the owner was an organization type, we won't link to team references. But.. the likely hood that they could retrieve markdown to view for a repo and not the repo details would be slim.

app/src/lib/databases/repositories-database.ts
tidy-dev: > shouldn't be storing raw API objects in the database. They often contain a lot more dat tha

:flushed: That totally makes sense... now that you point it out. 

I didn't know about the owner having the type on it (follow dotcom logic, the repo was checked for org). This follows more logically with what I am trying to accomplish. 🙇 

I have updated to change to adding an optional owner type so it can be added as the repo is refreshed. Even if there is a case where a pr is viewed before this happens, missing a link isn't isn't a big let down and should only happen once.

app/src/lib/markdown-filters/team-mention-filter.ts
```diff
@@ -90,7 +95,8 @@ export class TeamMentionFilter implements INodeFilter {
         org === undefined ||
         team === undefined ||
         // Team references are only added when the repository owner is the org to prevent linking to a team outside the repositories org.
-        org.slice(1) !== this.repository.owner.login
+        org.slice(1).toLocaleLowerCase() !==
+          this.repository.owner.login.toLocaleLowerCase()
```
niik: we should only use locale-aware string methods when dealing with data that is to be presented to the user (and currently not even that given that we don't localize the app) such as for sorting.

While using `localeLowerCase` _on both sides_ of a comparison like this is _probably_ safe there's no reason for us to bring locales into this and it can have unintended consequences. The typical example is Turkish where the lower case of `I` isn't `i`, but `ı`. Localization is super hard as I've once again learned this week and we don't want application logic to change (potentially) depending on a user's locale.

```suggestion
        caseInsensitiveEquals(org.slice(1), this.repository.owner.login)
```

app/src/lib/stores/repositories-store.ts
niik: Yeah, the skeleton ones are always going to be funky. The skeleton repositories are super weird. This seems like a great tradeoff tho

app/src/models/owner.ts
```diff
@@ -6,7 +8,8 @@ export class Owner {
   public constructor(
     public readonly login: string,
     public readonly endpoint: string,
-    public readonly id: number
+    public readonly id: number,
+    public readonly type?: GitHubAccountType
```
niik, line 12: Looks like we're not using the `hash` property on `Owner` any more which is great but we should either add this property to `.hash` or remove `.hash` entirely to stay consistent with the intent of `.hash`

niik (changes requested):
Really happy with this direction @tidy-dev!

app/src/models/owner.ts
tidy-dev: Decided to remove

niik (approved):

PR #14175, opened 2022-03-15
Pull request review notifications

sergiou87:
## Description

This PR adds support for notifications when someone submits a review to a Pull Request where the GitHub Desktop user is the author. It includes:
- New PR review dialog to display information about commented, approved and "changes requested" reviews.
- The app listens now to PR review submit Alive events and displays a notification with them. Clicking on these notifications will open the aforementioned dialog.

This is most of the work, but it still requires a few tweaks that I plan to tackle in different PRs:
- There still seems to be some layout issues with our `SandboxedMarkdown` component:
   - When the dialog is opened from a notification, the review body is rendered empty for around 1 second before its content is rendered properly.
   - ~~The height of the `SandboxedMarkdown` component is taller than it should, which is not very apparent when the comment has a lot of content, but it is with 1-line bodies.~~ Fixed for most of the times in #14196 
- ~~We'll want to change what the dialog looks like when a review doesn't have a `body`, in which case it might or might not have inline comments.~~ Updated with new designs :_D
- ~~Add some metrics around this new notification~~ Done in #14215 

### Screenshots

https://user-images.githubusercontent.com/1083228/159535964-0ac01ffb-37c5-484c-926b-166acc7f26f5.mov

image
image

## Release notes

Notes: [Added] Show notifications when Pull Requests are reviewed

PR #14215, opened 2022-03-21
Pull request review notification metrics

sergiou87:
## Description

This PR is based on #14175 

It adds metrics for our new Pull Request review notifications:
- Track when this kind of notifications are displayed to users.
- Track when users click on this kind of notifications.
- Track when users click on "Switch to Pull Request" from the PR review dialog shown for this kind of notifications.

In order to reuse some code and at the same time keep our code compile-time safe, I've used [some TypeScript magic](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html). I can't say if it's too much magic or not enough, though 😂  I'll leave some inline comments for more detail.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -484,6 +484,42 @@ export interface IDailyMeasures {
    * failed" dialog.
    */
   readonly checksFailedDialogRerunChecksCount: number
+
+  /** The number of "approved PR" notifications the user received */
+  readonly pullRequestReviewApprovedNotificationCount: number
+
+  /** The number of "approved PR" notifications the user clicked */
+  readonly pullRequestReviewApprovedNotificationClicked: number
+
+  /**
+   * The number of times the user decided to switch to the affected pull request
+   * from the "approved PR" dialog.
+   */
+  readonly pullRequestReviewApprovedDialogSwitchToPullRequestCount: number
+
+  /** The number of "commented PR" notifications the user received */
+  readonly pullRequestReviewCommentedNotificationCount: number
+
+  /** The number of "commented PR" notifications the user clicked */
+  readonly pullRequestReviewCommentedNotificationClicked: number
+
+  /**
+   * The number of times the user decided to switch to the affected pull request
+   * from the "commented PR" dialog.
+   */
+  readonly pullRequestReviewCommentedDialogSwitchToPullRequestCount: number
+
+  /** The number of "changes requested" notifications the user received */
+  readonly pullRequestReviewChangesRequestedNotificationCount: number
+
+  /** The number of "changes requested" notifications the user clicked */
+  readonly pullRequestReviewChangesRequestedNotificationClicked: number
+
+  /**
+   * The number of times the user decided to switch to the affected pull request
+   * from the "changes requested" dialog.
+   */
+  readonly pullRequestReviewChangesRequestedDialogSwitchToPullRequestCount: number
 }
 
 export class StatsDatabase extends Dexie {
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -27,6 +27,20 @@ import { MultiCommitOperationKind } from '../../models/multi-commit-operation'
 import { getNotificationsEnabled } from '../stores/notifications-store'
 import { isInApplicationFolder } from '../../ui/main-process-proxy'
 import { getRendererGUID } from '../get-renderer-guid'
+import { ValidNotificationPullRequestReviewState } from '../valid-notification-pull-request-review'
+
+type PullRequestReviewStatFieldInfix =
+  | 'Approved'
+  | 'ChangesRequested'
+  | 'Commented'
+
+type PullRequestReviewStatFieldSuffix =
+  | 'NotificationCount'
+  | 'NotificationClicked'
+  | 'DialogSwitchToPullRequestCount'
+
+type PullRequestReviewStatField =
+  `pullRequestReview${PullRequestReviewStatFieldInfix}${PullRequestReviewStatFieldSuffix}`
 
 const StatsEndpoint = 'https://central.github.com/api/usage/desktop'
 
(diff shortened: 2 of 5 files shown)


app/src/lib/stats/stats-store.ts
```diff
@@ -28,6 +28,18 @@ import { MultiCommitOperationKind } from '../../models/multi-commit-operation'
...
+  | 'Approved'
+  | 'ChangesRequested'
+  | 'Commented'
+
+type PullRequestReviewStatFieldSuffix =
+  | 'NotificationCount'
+  | 'NotificationClicked'
+  | 'DialogSwitchToPullRequestCount'
+
+type PullRequestReviewStatField = `pullRequestReview${PullRequestReviewStatFieldInfix}${PullRequestReviewStatFieldSuffix}`
```
sergiou87: This is basically the "magic" I mentioned. Basically, using the [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to build all the relevant stat field names into a type, the rest is in the `recordBlahBlah` methods that use this type to make sure we're building the right metric key.

app/src/lib/stats/stats-database.ts
```diff
@@ -484,6 +484,42 @@ export interface IDailyMeasures {
...
+  readonly pullRequestReviewChangesRequestedNotificationCount: number
+
+  /** The number of "changes requested" notifications the user clicked */
+  readonly pullRequestReviewChangesRequestedNotificationClicked: number
+
+  /**
+   * The number of times the user decided to switch to the affected pull request
+   * from the "changes requested" dialog.
+   */
+  readonly pullRequestReviewChangesRequestedDialogSwitchToPullRequestCount: number
```
sergiou87, line 522: I wondered if I should've used the type `PullRequestReviewStatField` to build the specific interface for those metrics, something like…
```typescript
type IDailyMeasures = {
...
} & {
  [K in PullRequestReviewStatField]: number
}
```

It feels a bit hacky, but OTOH it's just 3 lines 😂

app/src/lib/stats/stats-store.ts
```diff
@@ -1777,6 +1798,54 @@ export class StatsStore implements IStatsStore {
...
+    if (reviewType === 'APPROVED') {
+      infix = 'Approved'
+    } else if (reviewType === 'CHANGES_REQUESTED') {
+      infix = 'ChangesRequested'
+    } else {
+      infix = 'Commented'
+    }
+
+    return `pullRequestReview${infix}${suffix}`
+  }
```
sergiou87: Not very fond of `IAPIPullRequestReview['state']`, not only because it means bringing API-related stuff here (and some other interfaces), but also because that field has more values than we need for notifications.

Maybe I should write something like a `ValidNotificationPullRequestReview` type which is like `IAPIPullRequestReview` but where state is either `APPROVED`, `CHANGES_REQUESTED` or `COMMENTED`, and a type guard to ensure this 🤔  I'll give that approach a try.

app/src/lib/stats/stats-store.ts
```diff
@@ -1777,6 +1798,54 @@ export class StatsStore implements IStatsStore {
...
+  // Generic method to record stats related to Pull Request review notifications.
+  private recordPullRequestReviewStat(
+    reviewType: IAPIPullRequestReview['state'],
+    suffix: PullRequestReviewStatFieldSuffix
+  ) {
+    const statField = this.getStatFieldForRequestReviewState(
+      reviewType,
+      suffix
+    )
+    return this.updateDailyMeasures(m => ( { [statField]: m[statField] + 1 } as any ))
```
sergiou87: This is the ugliest part, because I had to add `as any` here, otherwise TS expects us to include all fields in `PullRequestReviewStatField` in the new object.

It shouldn't be bad because `m[statField]` already makes sure `statField` is a key of `IDailyMeasures` (i.e., that `PullRequestReviewStatField extends keyof IDailyMeasures`), but it'd be nice if I could get rid of it!

app/src/lib/stats/stats-store.ts
sergiou87: It seems prettier doesn't like template literal types right now 😅  https://github.com/desktop/desktop/runs/5626068009?check_suite_focus=true/#step:8:1

app/src/lib/stats/stats-store.ts
sergiou87: Done in 6156fe9cd2ce16472ded4194ca76d7c727f692eb...2ae88489687a033235759d17aa5d2784a95f1945 😍

sergiou87: I moved that cleanup to #14175

app/src/lib/stats/stats-store.ts
```diff
@@ -27,6 +27,20 @@ import { MultiCommitOperationKind } from '../../models/multi-commit-operation'
...
+  | 'ChangesRequested'
+  | 'Commented'
+
+type PullRequestReviewStatFieldSuffix =
+  | 'NotificationCount'
+  | 'NotificationClicked'
+  | 'DialogSwitchToPullRequestCount'
+
+type PullRequestReviewStatField =
+  `pullRequestReview${PullRequestReviewStatFieldInfix}${PullRequestReviewStatFieldSuffix}`
```
tidy-dev, line 43: 🤯  This is wild looking, but so cool. :)

app/src/lib/stats/stats-database.ts
tidy-dev: Only pause it gives me is that it is currently very readable. (and following convention.. but if there is a better convention :D)

tidy-dev (approved):
:sparkles: Nice reduction of redundancy! I feel like the rebase/cherrypick/squashing and probably more could be improved in this way.

app/src/lib/stats/stats-database.ts
tidy-dev: Only pause it gives me is that it is currently very readable. (and following convention.. but if there is a better convention :D)

sergiou87:
Thanks for the review @tidy-dev 🙇‍♂️ 

I will go ahead and merge it to get all this ready for release, although I'd love to get feedback from our resident TypeScript wizard @niik and I'll be happy to address any concerns he has in separate PRs.

app/src/lib/stats/stats-store.ts
```diff
@@ -1772,6 +1795,55 @@ export class StatsStore implements IStatsStore {
...
+
+  // Generic method to record stats related to Pull Request review notifications.
+  private recordPullRequestReviewStat(
+    reviewType: ValidNotificationPullRequestReviewState,
+    suffix: PullRequestReviewStatFieldSuffix
+  ) {
+    const statField = this.getStatFieldForRequestReviewState(reviewType, suffix)
+    return this.updateDailyMeasures(
+      m => ({ [statField]: m[statField] + 1 } as any)
+    )
```
niik, line 1823: This cast here is a bummer but I'm not sure what the best way to get rid of it would be. Copying and assigning manually works but isn't as succinct

```ts
return this.updateDailyMeasures(m => {
  const copy = { ...m }
  copy[statField] = copy[statField] + 1
  return copy
})
```

niik (commented):
> [...] our resident TypeScript wizard @niik

You give me way too much credit, there are actual type theorists lurking in Desktop's history.

Either way I love it. I would love to have gotten rid of that `as any` but I won't make you create a new PR for is unless you want to. Especially given that the suggested solution isn't super pretty

app/src/lib/stats/stats-store.ts
sergiou87: Yep I think I'm gonna choose succinctness here, also because I'd like to avoid that extra copy. Thanks for looking into this and for your feedback! 🙇‍♂️  Glad to know I'm learning useful stuff from those type challenges 😂

PR #14220, opened 2022-03-22
Swap escapeRegExp for lodash method

tidy-dev:
## Description
I keep running into where I use the `escapeRegExp` method and my auto importer chooses the lodash method. I looked into what was the difference between our method and the one in lodash and it seems to be that our regex escape also escapes the hyphen. That should be inconsequential as the hyphen is only a special character when used as a range inside brackets and brackets will be escaped.  Example: `[a-z]`  escaped become `\[a-z\]` so in a regex it literally searches for `[`, `a` `-` `z` `]` as opposed to any lowercase letter; thus, the hyphen does not need escaped.

Our Regex: (linked to regex 101 tester)
[`/[.*+\-?^${}()|[\]\\]/g`](https://regex101.com/r/lHktx8/1)

Lodash Regex: (linked to regex 101 tester)
[`/[\\^$.*+?()[\]{}|]/g`](https://regex101.com/r/SUYxUW/1)

As it seems there should be no impact with removing our custom method in favor of lodash. That is what this PR does.

Current usages:
- Markdown filters (still flagged to dev): emoji-filter, issue-link-filter, and video-url filter. Technically some of these had hypen so feel more "directly" to be potentially impacted, but I did some spot testing appeared good.
- The desktop-file-transport: Escapes `.desktop.${__RELEASE_CHANNEL__}.log` which doesn't have hyphens currently.
- Side-by-side diff uses it in the calcSearchTokens for the find feature of the diff. This one is escaping user input and feels the mostly likely to be able to be impacted.. but with the reasoning above and some spot testing.. I think it is safe.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/helpers/regex.ts
+++ b/app/src/lib/helpers/regex.ts
@@ -49,7 +49,8 @@ export function getMatches(text: string, re: RegExp): Array<RegExpExecArray> {
  * example return [ "LargeFile.exe (150.00 MB)", "AlsoTooLargeOfAFile.txt (1.00 GB)" ]
  */
 export function getFileFromExceedsError(error: string): string[] {
-  const endRegex = /(;\sthis\sexceeds\sGitHub's\sfile\ssize\slimit\sof\s100.00\sMB)/gm
+  const endRegex =
+    /(;\sthis\sexceeds\sGitHub's\sfile\ssize\slimit\sof\s100.00\sMB)/gm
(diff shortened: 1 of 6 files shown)


sergiou87 (approved):
I think it makes sense, and [even the regex linked in the `escapeRegExp` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Escaping) excludes `-` from it.

PR #14226, opened 2022-03-23
Add caller info on unexpected git exit codes

sergiou87:
## Description

Sometimes we get non-fatal errors for unexpected git exit codes and the stack trace doesn't show which specific git command was. This adds our custom identifiers to the error description.

## Release notes

Notes: no-notes


tidy-dev (approved):
:sparkles: Makes sense

PR #14228, opened 2022-03-23
Use open dialog when cloning on windows

### LINKED ISSUE #12812: Clone Selecting Directory Requires a file instead of a folder
### Describe the bug
The bug is inside the clone repository window. When I want to select a directory to clone the repo, It's actually a file dialog and it requires a file instead of a folder

### Version & OS

Version 2.9.0 (x64) on Windows 10

1. Click on File
2. Choose Clone Repository
3. Click on "Choose..." button

### Expected behavior

It should be a directory dialog so I can choose a folder

### Actual behavior

It's a file dialog so I'm not able to select a folder

tidy-dev:
Closes #12812

## Description
Reverted clone dialog to from save to open dialog on windows.

### Screenshots
https://user-images.githubusercontent.com/75402236/159717300-078251d4-0e98-4b44-8e96-c580dc29fd11.mov

## Release notes
Notes: [Fixed] Clone dialog "choose" button uses an open dialog for directory selection on mac.

### DIFF
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -78,11 +78,6 @@ export function enableUpdateFromEmulatedX64ToARM64(): boolean {
   return enableBetaFeatures()
 }
 
-/** Should we allow using the save dialog when choosing where to clone a repo */
-export function enableSaveDialogOnCloneRepository(): boolean {
-  return false
-}
-
 /** Should we allow setting repository aliases? */
 export function enableRepositoryAliases(): boolean {
   return true
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -80,7 +80,7 @@ export function enableUpdateFromEmulatedX64ToARM64(): boolean {
 
 /** Should we allow using the save dialog when choosing where to clone a repo */
 export function enableSaveDialogOnCloneRepository(): boolean {
-  return true
+  return __DARWIN__
 }
 
 /** Should we allow setting repository aliases? */
--- a/app/src/lib/feature-flag.ts
+++ b/app/src/lib/feature-flag.ts
@@ -80,7 +80,7 @@ export function enableUpdateFromEmulatedX64ToARM64(): boolean {
 
 /** Should we allow using the save dialog when choosing where to clone a repo */
 export function enableSaveDialogOnCloneRepository(): boolean {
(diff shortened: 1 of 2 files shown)


billygriffin (commented):
Awesome! Just a question - do we use the feature flag pattern for differentiating between Windows and macOS anywhere else? I'm wondering if this will be clear for others who come upon this in the future.

tidy-dev:
> Awesome! Just a question - do we use the feature flag pattern for differentiating between Windows and macOS anywhere else? I'm wondering if this will be clear for others who come upon this in the future.

Not really - we do have one other feature flag with arm64 updating that is flagged to macOs.  This save vs open dialog is more of unique case because we are providing different feature per OS, where usually macOS vs windows is either wording or because some underlying process (like spawn) doesn't work the same.

That said, I moved it into the code so that it would be more visible to a future devs that we are doing something different per OS and added some documentation as to why since it is a unique case.

Other notes: I took out the feature flag since it isn't really providing anything at this point and we could easily add back since we are still flagging by os if we so choose.

app/src/lib/feature-flag.ts
```diff
@@ -80,7 +80,7 @@ export function enableUpdateFromEmulatedX64ToARM64(): boolean {
 
 /** Should we allow using the save dialog when choosing where to clone a repo */
 export function enableSaveDialogOnCloneRepository(): boolean {
```
j-f1: Is this function used anymore? If not, should it be removed?

j-f1 (dismissed):
This makes a lot of sense — especially after seeing the new dialog on Windows, it much better matches the current macOS experience!

app/src/lib/feature-flag.ts
tidy-dev: Nope! Good catch. I thought I did remove it, but didn't save that file. 😳

Nonso15:
Legit company to invest and trade your Bitcoin and earn your profits daily safely withdrawal instantly.
🎲🎲🎲🎲🎲🎲🎲🎲🎲
https://Jomo-trade.com
https://Jomo-trade.com

Legit✅✅
No withdrawal fee✅✅
Instant withdraw ✅✅
https://Jomo-trade.com
https://Jomo-trade.com

sergiou87 (approved):
🐸 🕴️ ❗

PR #14231, opened 2022-03-24
Add the option to view the current branch on GitHub

### LINKED ISSUE #14224: Make "View on GitHub" point to the current branch or PR
### Search for existing feature requests

- [X] I have searched the issue tracker and confirmed that I was unable to find existing issues for the feature I am requesting.

### The feature request

I feel like the main "View on GitHub" button on the right-hand side should open the current branch or the open PR if we are looking at a branch that has one.

It is a lot easier to go back to the main page of the repo if that's what we want than going to these two pages, which is a very common flow.

### Proposed solution

Simply re-routing the "View on GitHub" button to:

1. the PR page if the branch we are on has one
2. the code page for the branch if it does not

### Additional context

Happy to help contribute / QA. And thanks for the amazing work!

tsvetilian-ty:
Closes #14224

## Description
Add option to "Branch" to open the current branch on GitHub.

Points to get feedback on before converting draft to standard PR
- [x] Position in the Branch menu
- [x] Should the option have shortcut/accelerator


tidy-dev:
> Position in the Branch menu

Where you put it makes good sense.

> Should the option have shortcut/accelerator

Definitely think it should... however, I am not sure there is a shortcut that just "makes sense". Maybe CtrlOrCmd+Alt+B

Other notes: Would you be willing to update the "Show Pull Request" to "View Pull Request on GitHub" to be consistent with everywhere else in the app as part of this?

tsvetilian-ty:
@tidy-dev, thanks for your feedback 💯
I will mark the PR as ready for review

tidy-dev (dismissed):
Thanks you for contributions @tsvetilian-ty!

(Approving - but also going to give some time before merging so other team members can weigh in on it. 🙂 )

sergiou87 (dismissed):
I like the consistency in wording and also the new menu item to view the current branch on GH 😍  Thank you!

ericboucher:
Wow that was quick, thanks a ton @tidy-dev and @tsvetilian-ty ❤️

app/src/ui/app.tsx
```diff
@@ -612,6 +614,29 @@ export class App extends React.Component<IAppProps, IAppState> {
     this.props.dispatcher.openInBrowser(compareURL)
   }
 
+  private showBranchOnGitHub() {
+    const htmlURL = this.getCurrentRepositoryGitHubURL()
```
ericboucher: Most of the logic in here could be shared with `compareBranchOnDotcom()`

tidy-dev: Good catch! @tsvetilian-ty think you could update this such that the two actions share a method?

tsvetilian-ty: @ericboucher great catch 🔝 @tidy-dev 👍

app/src/ui/app.tsx
```diff
@@ -387,7 +387,9 @@ export class App extends React.Component<IAppProps, IAppState> {
       case 'view-repository-on-github':
         return this.viewRepositoryOnGitHub()
       case 'compare-on-github':
-        return this.compareBranchOnDotcom()
+        return this.openBranchOnGitub(true)
```
ericboucher: Sorry for being picky but I would call out "compare=true/false" explicitly for clarity

app/src/ui/app.tsx
```diff
@@ -608,8 +610,15 @@ export class App extends React.Component<IAppProps, IAppState> {
       return
     }
 
-    const compareURL = `${htmlURL}/compare/${branchTip.branch.upstreamWithoutRemote}`
-    this.props.dispatcher.openInBrowser(compareURL)
+    let url = htmlURL
```
tidy-dev: We try to avoid using `let` if possible and I think we could simplify this function a bit by passing in a variable of `view: 'tree' | 'compare'` and then const url = `${htmlURL}/${view}/${branchTip.branch.upstreamWithoutRemote}`

tidy-dev (changes requested):
@tsvetilian-ty Thank you for updating that. I think we can make it a tad simpler and easier to read/clarity on action. See my suggestion. :)

tidy-dev (approved):
✨  Thank so much for those changes!

PR #14232, opened 2022-03-24
Update currently supported editors list on Windows

tsvetilian-ty:
## Description
Update docs information for supported editors on Windows

 - [JetBrains IntelliJ Idea](https://www.jetbrains.com/idea/)
 - [JetBrains RubyMine](https://www.jetbrains.com/rubymine/)
 - [JetBrains GoLand](https://www.jetbrains.com/go/)
 - [Android Studio](https://developer.android.com/studio)
 - [Brackets](http://brackets.io/)


tidy-dev (approved):
:sparkles: Good catch! Thanks for the update! 

Notes for future self: Updating this is something we probably ought to require when new editor pr's come in.

Issue #78, opened 2016-06-20
Setup gpg signing


joshaber:
We should explore setting up commit signing with GPG in Desktop. This could take two forms (or both):

1. Automatic commit signing using GitHub's GPG key in the same way that dotcom does it today, if the user is signed in to GitHub from Desktop.
2. Allow people to set up and utilize their own GPG key to sign commits in Desktop, and make that visible to users during the commit flow.

foss-:
Any news on this? Essential functionality and has been open for over a year.

niik:
> Any news on this? Essential functionality and has been open for over a year.

Hi @foss-, this is something that we're interested in but it's not part of our immediate [roadmap](https://github.com/desktop/desktop/blob/master/docs/process/roadmap.md). The best way to stay up to date would be to subscribe to this issue. I've heard whispers about it being possible to set up even though we currently lack first class support for it in the app (@shiftkey?) so if someone has got it working please feel free to share your workarounds here.

j-f1:
Bringing this over from the duplicate:

> It would be even better if the app integrated with [Keybase](https://keybase.io) so it could automatically obtain the GPG key.

shiftkey:
We've fixed some reported issues about our commit step failing when a user has signing enabled and configured already (see #1391 and #1578 for example) and if you're seeing errors like those in the app we'd like to fix it, but @niik is right in that we don't have inbuilt support for configuring commit signing in the app.

shiftkey:
#1974 indicates that we don't prompt for the user to enter the passphrase associated with the GPG key, we may need to support this flow for users who don't want to use `gpg-agent` to provide this transparently.

jccguimaraes:
@shiftkey thanks for the info shared! Keep up the good work!

bopoh13:
The problem of the impossibility of passing the *passphrase* exists on all OS? You would at least write, how works `gpg-agent` (in the package did not find) :scream_cat:
``` bash
$ type gpg-agent
bash: type: gpg-agent: not found
```
In Google additionally there is a solution `echo "no-tty" >> ~/.gnupg/gpg.conf` 
& `git config --global gpg.program $(which gpg)`. But in Windows [it](//help.github.com/articles/signing-commits-using-gpg/) does not work.
To resolve the [error](//lists.gnupg.org/pipermail/gnupg-users/2003-May/018492.html) must:
``` bash
sed -i '/^no-tty/d' ~/.gnupg/gpg.conf
[...]

PR #14234, opened 2022-03-24
Add option to configure signing key in Preferences & Repository settings

wojtekmaj:
## Description

Adds 3rd, after Name and Email, option to GitConfigUserForm. After GPG key is generated, user is able to easily configure Signing key both on global and repository level.

### Screenshots

Couldn't run the app on Node 17. :(

## Release notes

Notes: Add option to configure GPG signing key in Preferences & Repository settings

### DIFF
--- a/app/src/ui/lib/configure-git-user.tsx
+++ b/app/src/ui/lib/configure-git-user.tsx
@@ -36,6 +36,7 @@ interface IConfigureGitUserProps {
 interface IConfigureGitUserState {
   readonly globalUserName: string | null
   readonly globalUserEmail: string | null
+  readonly globalSigningKey: string | null
 
   readonly manualName: string
   readonly manualEmail: string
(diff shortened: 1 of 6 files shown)


sergiou87:
Thank you for your contribution @wojtekmaj! ❤️ 

However, we don't usually accept PRs of new features without previous consensus 😅  This in particular is something we have discussed in the past and never got to a solution we're happy with. We'd like to make it super simple to completely setup the user's environment for commit signing (vs, for example, only setting the signing key), but it's quite complicated, so we haven't prioritized it yet.

I'd recommend you to make your argument in https://github.com/desktop/desktop/issues/78. We revisit this topic every few months, so we appreciate al feedback we get. In the meantime, I'll close this PR. I'm sorry about that and thanks again for your effort 🙇‍♂️

PR #14235, opened 2022-03-24
Fix group header overflow when text is too long

### LINKED ISSUE #14233: Text out of range when group name is long
### Search existing issues

### The problem

When I click the Current Repository tab, the long-name group is out of range in the drop-down menu, so I can't read the name clearly.

### Release version

Version 2.9.12 (x64)

### Operating system

macOS Monterey

1. Click on 'Current Repository'
2. See error (If you have long -name group)

### Log files

_No response_

### Screenshots

스크린샷 2022-03-24 오후 8 50 12

### Additional context

_No response_

tsvetilian-ty:
Closes #14233

### Screenshots
Before the fix
scree-rec

After the fix
scree-rec

### DIFF
--- a/app/src/ui/repositories-list/repositories-list.tsx
+++ b/app/src/ui/repositories-list/repositories-list.tsx
@@ -20,6 +20,7 @@ import { showContextualMenu } from '../../lib/menu-item'
 import { IMenuItem } from '../../lib/menu-item'
 import { PopupType } from '../../models/popup'
 import { encodePathAsUrl } from '../../lib/path'
+import { TooltippedContent } from '../lib/tooltipped-content'
 import memoizeOne from 'memoize-one'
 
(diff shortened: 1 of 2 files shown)


tidy-dev (changes requested):
Thanks @tsvetilian-ty! This is looking great! 

We have introduced a `TooltippedContent` component that we use instead of the `title` attribute to give us more flexibility and consistency. I believe how we use it in the [branch-list-item](https://github.com/desktop/desktop/blob/f27e037d75c1627a98b1814ecf154cf3c943c316/app/src/ui/branches/branch-list-item.tsx#L160) is probably close to what we need here. If you could swap the div out here for that, I think this would be good to go. 🙂

tidy-dev (approved):

PR #14239, opened 2022-03-25
Fix PRs appearing to be created in the future

### LINKED ISSUE #14238: PR relative dates are in the future
### Search existing issues

### The problem

Relative dates on the PR tab are in the future. Instead of `in x days` and `tomorrow` it should `x days ago` and `yesterday`.

### Release version

2.9.12

### Operating system

Windows 10

_No response_

### Log files

_No response_

### Screenshots

image

### Additional context

_No response_

niik:
Closes #14238

## Description

Fixes the issue with pull requests, and one log message (skipping background fetch), being shown as happening in the future due to me mistakenly switching the sign order.

## Release notes

Notes: [Fixed] Pull requests adhere to temporal laws again 

(we should come up with a better changelog entry)

### DIFF
--- a/app/src/lib/format-relative.ts
+++ b/app/src/lib/format-relative.ts
@@ -17,7 +17,7 @@ const getRelativeFormatter = mem(
 export function formatRelative(ms: number) {
   const formatter = getRelativeFormatter('en-US', { numeric: 'auto' })
 
-  const sign = ms < 0 ? -1 : 0
+  const sign = ms < 0 ? -1 : 1
 
   // Lifted and adopted from
   // https://github.com/github/time-elements/blob/428b02c9/src/relative-time.ts#L57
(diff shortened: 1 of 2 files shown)


app/src/lib/format-relative.ts
```diff
@@ -17,7 +17,7 @@ const getRelativeFormatter = mem(
 export function formatRelative(ms: number) {
   const formatter = getRelativeFormatter('en-US', { numeric: 'auto' })
 
-  const sign = ms < 0 ? 1 : -1
+  const sign = ms < 0 ? -1 : 0
```
tidy-dev: Setting the sign here to 0 seems strange to me.. but I guess that is to follow the assumption that we should not have future dates.

tidy-dev (dismissed):
✨  Appears to work as expected and spot checked a couple other places. 

Also, that release note is awesome. 🚀

app/src/lib/format-relative.ts
niik: No, it's just straight up strange. As in wrong. As in attempting to squeeze in a fix on a Friday afternoon is a terrible idea. I'll fix it.

sergiou87 (approved):

tidy-dev (approved):

PR #14245, opened 2022-03-28
Release 2.9.13-beta1

sergiou87:
## Description
Looking for the PR for the upcoming first beta of the v2.9.13 series? Well, you've just found it, congratulations!

## Release checklist

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -4,7 +4,7 @@
       "[New] Add support for notifications of pull request reviews - #14175",
       "[Fixed] Pull requests adhere to temporal laws again - #14238",
       "[Fixed] Fix repository group header overflow when text is too long - #14233. Thanks @tsvetilian-ty!",
-      "[Fixed] Clone dialog \"Choose\" button uses an open dialog for directory selection on macOS - #12812"
+      "[Fixed] Clone dialog \"Choose\" button uses an open dialog for directory selection on Windows - #12812"
     ],
     "2.9.12": [
       "[Added] Add support for Brackets Editor on Windows - #14069. Thanks @tsvetilian-ty!",
--- a/changelog.json
+++ b/changelog.json
@@ -4,8 +4,7 @@
       "[New] Add support for notifications of pull request reviews - #14175",
       "[Fixed] Pull requests adhere to temporal laws again - #14238",
       "[Fixed] Fix repository group header overflow when text is too long - #14233. Thanks @tsvetilian-ty!",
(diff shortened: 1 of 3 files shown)


changelog.json
```diff
@@ -1,5 +1,12 @@
 {
   "releases": {
+    "2.9.13-beta1": [
+      "[New] Add support for notifications of pull request reviews - #14175",
+      "[Fixed] Pull requests adhere to temporal laws again - #14238",
```
sergiou87, line 5: Not sure if there is precedent of funny release notes like this, but I personally love it 😂

changelog.json
```diff
@@ -1,5 +1,12 @@
...
+      "[Fixed] Fix repository group header overflow when text is too long - #14233. Thanks @tsvetilian-ty!",
+      "[Fixed] Clone dialog \"Choose\" button uses an open dialog for directory selection on macOS - #12812",
```
tidy-dev: :D - my bad on writing the original note.. but is windows that is changing not mac.

```suggestion
      "[Fixed] Clone dialog \"Choose\" button uses an open dialog for directory selection on windows - #12812",
```

changelog.json
```diff
@@ -1,5 +1,12 @@
...
+      "[Improved] Add link to beta download page in about dialog - #14237"
```
say25:
```suggestion
```
Technically this doesn’t show in the betas but I don’t recall how release notes do or don’t build on each other 

Also this should really be a Added vs Improve

```suggestion
      "[Added] Link to beta download page in about dialog - #14237"
```

sergiou87: You're right it shouldn't be in these release notes at all, but I think `Improved` is the right category according to https://github.com/desktop/desktop/blob/241c4a2cfda89dc60d2ee7a094657ad349de7274/docs/process/writing-release-notes.md/#L129-L134

I'll remove it, thanks!! ❤️

tidy-dev (changes requested):

changelog.json
tidy-dev: .... So I thought I left this comment yesterday when the pr was opened.. but it was still pending. :face-palm:

bayleev33me:
Ok having a bad network trying to open site I have even I have data

David bayleev

changelog.json
sergiou87: Fixed in [ee57829](https://github.com/desktop/desktop/pull/14245/commits/ee5782916f73a517a71762b9ed1909873fbfcbd5)!

tidy-dev (approved):
✨

PR #14253, opened 2022-03-29
Markdown Commit Mention filter

tidy-dev:
## Description
This pr adds a commit mention filter to our markdown parser.

It looks for three patterns:
1) SHA (7-40 hex characters): e6c39f9ffdabc8370e2f04822968d4f7de63ede3
2) SHA...SHA: e6c39f9ffdabc8370e2f04822968d4f7de63ede3...e6c39f9ffdabc8370e2f04822968d4f7de63ede3
3) user/repo@SHA: tidy-dev/foo-really-not@e6c39f9ffdabc8370e2f04822968d4f7de63ede3

Then replaces those with a link and formats the sha ref by shortening it and putting it in a `<tt>` tag.

For the plain commits and sha ranges, dotcom will verify the commits exists within the markdown repo context. Thus this filter will have more false-positives. Additionally, this allows dotcom to have a more accurate short sha for those verified commits as opposed to this one which will follow the dotcom pattern of trimming unverified commits to 7 characters if the sha >= 30 characters in length.

To note, dotcom does not verify commit existence when the owner and repo is specified outside of the markdown's repo's context. Thus, both this filter and dotcom will have false positives that lead to 404's. However, dotcom does verify the existence of the owner/repo pair and this filter does not leading to more false positives.

The false-positives from this filter is not expected to be a likely scenario. Tho, a future iteration could introduce an async approach that after the fact verifies commits locally to reduce them.

Another differentiation from the dotcom approach is that if their filter is run in a context that the repo is undefined, they do not bother searching for anything other than user/repo@SHA. Since we are not verifying our commits existence anyways, we don't get any value from doing this.

For another PR:
- Dotcom's filter also searches for `https://github.com/desktop/desktop/commit/e6c39f9ffdabc8370e2f04822968d4f7de63ede3`. This link pattern to be directly pasted and if found shortens/formats it like, https://github.com/desktop/desktop/commit/e6c39f9ffdabc8370e2f04822968d4f7de63ede3.

### DIFF
--- a/app/src/lib/markdown-filters/commit-mention-filter.ts
+++ b/app/src/lib/markdown-filters/commit-mention-filter.ts
@@ -1,0 +1,336 @@
+import { GitHubRepository } from '../../models/github-repository'
+import { getHTMLURL } from '../api'
+import { INodeFilter } from './node-filter'
+import { resolveOwnerRepo } from './resolve-owner-repo'
+
+/**
+ * The Commit Mention Filter matches for sha patterns and replaces them with
+ * links to sha and concatenates long ones.
+ *
+ * There are three sha patterns:
+ *   1) SHA (7-40 hex characters)
+ *   2) SHA...SHA
+ *   3) user/repo@SHA
+ *
+ * Notes:
+ *  1) When no user/repo is provided, the link defaults to the provided repo
+ *     owner and/or repo name.
+ *  2) Notable difference from dotcom approach is that, it does not verify
+ *     commit exists in the given repo context. (To note, dotcom doesn't verify
+ *     for repo's outside of the markdown context.) This improves performance at
+ *     the cost of false-positives. Additionally, all commit shas are trimmed to
+ *     7 characters, if >= 30 characters, unlike dotcom that obtains the git
+ *     short sha for shas in the markdown context.
+ *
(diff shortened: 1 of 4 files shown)


Hovemeyer5 (commented):
(ignore commented for testing something)

app/src/lib/markdown-filters/commit-mention-filter.ts
```diff
@@ -0,0 +1,334 @@
...
+   * Method to create a commit mention link element.
+   *
+   * If for a range, it links to a 'compare' view
+   * If for a commit, it links to a 'commit' view (default)
+   **/
+  private createCommitMentionLinkElement(
+    ref: string,
+    view: 'commit' | 'compare' = 'commit',
+    repoOwner: string = this.repository.owner.login,
+    repoName: string = this.repository.name,
```
sergiou87, line 315: TIL you can default to instance variables through `this` 🤯 🤩

sergiou87 (approved):
It all looks good! I tested it and worked as described 👌

bosniltd-a11y (commented):
003df3a324b8a23f9a9ad474fd94cb5cc676ae82

PR #14256, opened 2022-03-30
Fix CI check status popover not closing when clicking on PR badge

sergiou87:
## Description
When the CI check status popover is open and the user clicks on the PR badge, it flashes for a bit and is reopened again:

https://user-images.githubusercontent.com/1083228/160805700-356e27c7-ff68-4d4c-971e-66e8ecbd6bfd.mov

The problem is the `onClickOutside` callback from our `Popover` component is invoked on `mousedown` events on the document. That means the `mousedown` event would close the popover (because the user clicked outside of it), and then the app would handle the 'click' event on the PR badge, which would open the popover again.

This PR changes the `Popover` component to listen to `click` events as well. With this change, the `onClickOutside` will be invoked (and therefore popovers will be dismissed) when the user releases the mouse button to complete the `click` event.

## Release notes

Notes: [Fixed] Pull request checks popover is closed when the pull request number badge is clicked

### DIFF
--- a/app/styles/ui/check-runs/_ci-check-run-popover.scss
+++ b/app/styles/ui/check-runs/_ci-check-run-popover.scss
@@ -5,6 +5,10 @@
     width: 440px;
     top: 60;
     margin-left: -302px;
+
+    &::after {
+      border-bottom-color: var(--box-alt-background-color);
+    }
   }
 
   .ci-check-run-list-header {
--- a/app/src/ui/lib/popover.tsx
+++ b/app/src/ui/lib/popover.tsx
@@ -51,14 +51,14 @@ export class Popover extends React.Component<IPopoverProps> {
   }
 
   public componentDidMount() {
-    document.addEventListener('mousedown', this.onDocumentMouseDown)
+    document.addEventListener('click', this.onDocumentClick)
   }
 
   public componentWillUnmount() {
-    document.removeEventListener('mousedown', this.onDocumentMouseDown)
+    document.removeEventListener('click', this.onDocumentClick)
   }
 
-  private onDocumentMouseDown = (event: MouseEvent) => {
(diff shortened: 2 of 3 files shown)


sergiou87:
I just noticed this fixes the issue but it's due to the `click` handler of the badge being invoked before the "click outside" handler of the popover. That's not a problem, but maybe it makes sense to create a transparent click area in the whole window to capture clicks, so that they don't go through the rest of the app, and therefore disabling clicks on elements behind the popover while the popover is visible. Thoughts?

app/styles/ui/check-runs/_ci-check-run-popover.scss
```diff
@@ -5,6 +5,10 @@
     width: 440px;
     top: 60;
     margin-left: -302px;
+
+    &::after {
+      border-bottom-color: var(--box-alt-background-color);
+    }
```
sergiou87, line 11: While recording the video for this PR, I noticed the tip of the popover looked wrong on the dark theme, and this change fixes it:
image

tidy-dev (approved):
✨  Makes sense - works and tested the diff setting popover and works great too. :)

tidy-dev:
> I just noticed this fixes the issue but it's due to the click handler of the badge being invoked before the "click outside" handler of the popover. That's not a problem, but maybe it makes sense to create a transparent click area in the whole window to capture clicks, so that they don't go through the rest of the app, and therefore disabling clicks on elements behind the popover while the popover is visible. Thoughts?

The first thing that comes to mind is that it is treating it like a dialog where users have the expectation that they are "locked" into that view and clicking off may be required to dismiss it. Whereas the popover feels more integrated and if I had it opened, I would not expect that I couldn't click on something else.

PR #14258, opened 2022-03-30
Fix layout of re-run checks dialog

### LINKED ISSUE #14246: Re-run checks visual bug
### Search existing issues

### The problem

Re-run checks window has a visual bug where the list of checks overflows the container. Possibly due to a longer than predicted list of checks and lack of scrolling?

### Release version

2.9.12 (x64)

### Operating system

macOS Monterey

_No response_

### Log files

_No response_

### Screenshots

image

### Additional context

_No response_

sergiou87:
Closes #14246

## Description

This PR fixes two issues in the re-run dialog:
- When all jobs are re-runnable, we'd show a `There are 0 jobs that cannot be re-run` warning.
- When the list of jobs is too long, they would overlap with the footer even go beyond the boundaries of the dialog.

## Release notes

Notes:
- [Fixed] Re-run Checks dialog doesn't show warning when all checks can be re-run.
- [Fixed] Re-run Checks dialog can be scrolled for long lists of checks.

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
+++ b/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
@@ -131,20 +131,12 @@ export class CICheckRunRerunDialog extends React.Component<
     return (
       <div className="ci-check-run-list check-run-rerun-list">
         {this.state.rerunnable.length > 0 ? (
-          <>
-            <CICheckRunList
-              checkRuns={this.state.rerunnable}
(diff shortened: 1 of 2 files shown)


tidy-dev (approved):
:sparkles: Thank you for grabbing this. 

My bad on that poor wording for 0 nonRerunnable checks.  On the overflow, I feel like I checked that :/ - wonder if another fall out of chromium css change - but could have been an oversight since we don't have that many rerunnable checks.

PR #14261, opened 2022-03-30
Fix relative time of pull request reviews

sergiou87:
## Description

The recent fix of #14239 uncovered wrong math introduced in #14175 for the time at which PR reviews were submitted. This PR fixes that math 😅 

## Release notes

Notes: [Fixed] Pull request review notifications adhere to temporal laws again too


tidy-dev (approved):
✨  Makes sense. Maybe we can get all of Desktop back into this time loop.

PR #14282, opened 2022-04-04
Markdown commit mention link filter

tidy-dev:
## Description
Follow up to #14253, this markdown filter looks for pasted commit link patterns and abbreviates them.

Example:
`https://github.com/desktop/desktop/commit/5c8db58657e3a4b827d9f78c5a6e436f7a1348b6/test/test/test/test`
https://github.com/desktop/desktop/commit/5c8db58657e3a4b827d9f78c5a6e436f7a1348b6

Differences from dotcom: 
1. Dotcom verifies existence of owner/name. We would just have more false positives.
2. Dotcom verifies commits existence within current repo context for plain commit links and therefore may use more accurate short shas and less false positives.
3. Dotcom does not verify commits exist for compare path nor do they verify they look like a sha pattern. In filter, I added the extra step to verify the sha's do look like sha's to reduce false positives in compare links.

Other notes that also apply to #14253 is that plain commit links on dotcom also have a popover with commit information. That is not done here, as would require either a git command or an api call to non-local repositories -> meaning a performance hit. Something that could possibly be considered for local repos in an async manner, but not worth investing at this time for our scaled down version.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/markdown-filters/commit-mention-link-filter.ts
+++ b/app/src/lib/markdown-filters/commit-mention-link-filter.ts
@@ -1,0 +1,290 @@
+import { escapeRegExp } from 'lodash'
+import { GitHubRepository } from '../../models/github-repository'
+import { getHTMLURL } from '../api'
+import { INodeFilter } from './node-filter'
+
+/**
+ * The Commit mention Link filter matches the target and text of an anchor element that
+ * is an commit mention link and changes the text to a uniform
+ * reference.
+ *
+ * Types of commit mention links:
+ * - Plain Single Commit: https://github.com/desktop/desktop/commit/6fd794543af171c35cc9c325f570f9553128ffc9
+ * - Compare a range of Commits: https://github.com/desktop/desktop/compare/6fd794543...6fd794543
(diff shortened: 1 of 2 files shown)


app/src/lib/markdown-filters/commit-mention-link-filter.ts
```diff
@@ -0,0 +1,293 @@
...
+    const match = path.match(this.pullCommitPath)
+    if (match === null || match.groups === undefined) {
+      return null
+    }
+
+    return this.trimCommitSha(match.groups.sha)
+  }
+
+  /**
+   * Commit action path's are formatted nor shortened.
```
tidy-dev:
```suggestion
   * Commit action path's are not formatted nor shortened.
```

sergiou87 (approved):
It all makes sense and worked as expected ✨  :shipit:

fernado13 (commented):
@tidy-dev

app/src/lib/markdown-filters/commit-mention-link-filter.ts
fernado13: @sergiou87

PR #14284, opened 2022-04-05
Enable code signing on macOS for dev builds

### LINKED ISSUE #13639: `spawn` is slow on dev builds for macOS running on Apple silicon devices
### Describe the bug
After upgrading to Electron 13 (#13426), dev builds of the app take much longer to run `git` commands on macOS when running on Apple silicon devices. It still happens when building the app with Electron 15.

Normally, spawning git commands on a Mac Mini M1 in a given small repo would take around 0.0015s. But now, dev builds take 0.15-0.20s for the very same git commands.

### Version & OS

v2.9.6-dev
macOS 11.6+
Apple silicon

1. Build the app for development. Follow the instructions in https://github.com/desktop/desktop/blob/development/docs/contributing/setup.md#building-desktop
2. Run the app
3. Explore a repository
4. Keep an eye on the Developer Tools, where you can see how long it takes to run `git` commands.
5. Compare with the latest release build https://central.github.com/deployments/desktop/desktop/latest/darwin-arm64

### Expected behavior

Times should be similarly low in both dev and release builds.

### Actual behavior

Times are 10-100x slower on dev builds.

### Additional context

As mentioned in #12978 , on Electron 13 using `spawn` becomes very slow on the aforementioned conditions.

The problem seems to be in [libuv](https://github.com/libuv/libuv/issues/3050), and [a first attempt to fix it](https://github.com/libuv/libuv/pull/3064) made some Node.js tests fail, so the change [was reverted](https://github.com/libuv/libuv/pull/3107). This affects other apps like VS Code, and they decided to apply the changes to Electron, ignoring those broken Node.js tests, but [they found other problems derived from this](https://github.com/electron/electron/issues/26143#issuecomment-879569238) that they had to deal with.

Our hopes are in a [new PR trying to fix this](https://github.com/libuv/libuv/pull/3257), but the new fix led to a kernel bug that has been reported to Apple.

Related issues:
- Issue in libuv: https://github.com/libuv/libuv/issues/3050
- First report in the Electron repo: https://github.com/electron/electron/issues/26143
- Same issue in VS Code: https://github.com/microsoft/vscode/issues/132654
- VS Code uses a fork with the first fix but it causes its own issues: https://github.com/electron/electron/issues/26143#issuecomment-879569238
- First attempt o fix the issue: https://github.com/libuv/libuv/pull/3064
- Second attempt to fix the issue: https://github.com/libuv/libuv/pull/3257

sergiou87:
Closes #13639

## Description
As part of some improvements I wanted to do for notifications using the latest native APIs for macOS, I realized I could only use them on dev builds if it was signed, so I enabled code signing as Xcode does to run the app locally.

Not only this worked, but I also noticed a performance improvement on dev builds running on Apple Silicon, which got degraded when we moved to Electron 13 due to an issue with spawn on Big Sur and newer (see #13639 for more info).

#### Before signing
image

#### After signing
image

## Release notes

Notes: no-notes

### DIFF
--- a/script/entitlements-dev.plist
+++ b/script/entitlements-dev.plist
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <key>com.apple.security.automation.apple-events</key>
+    <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
+  </dict>
+</plist>
--- a/script/build.ts
+++ b/script/build.ts
@@ -56,13 +56,14 @@ import {
 } from 'fs'
 import { copySync } from 'fs-extra'
 
-const projectRoot = path.join(__dirname, '..')
-const entitlementsPath = `${projectRoot}/script/entitlements.plist`
-const extendInfoPath = `${projectRoot}/script/info.plist`
-const outRoot = path.join(projectRoot, 'out')
-
 const isPublishableBuild = isPublishable()
 const isDevelopmentBuild = getChannel() === 'development'
+
+const projectRoot = path.join(__dirname, '..')
+const entitlementsSuffix = isDevelopmentBuild ? '-dev' : ''
+const entitlementsPath = `${projectRoot}/script/entitlements${entitlementsSuffix}.plist`
+const extendInfoPath = `${projectRoot}/script/info.plist`
+const outRoot = path.join(projectRoot, 'out')
 
 console.log(`Building for ${getChannel()}…`)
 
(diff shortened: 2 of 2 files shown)


script/entitlements-dev.plist
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+  <dict>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
+    <key>com.apple.security.automation.apple-events</key>
+    <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <true/>
```
sergiou87, line 10: This was required (**for dev builds only**) to make macOS happy. More info here: https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-library-validation

niik (approved):
Looks great 👍 Could you please just do a test channel deploy before merging this to make sure we haven't accidentally changed the behavior of prod builds?

sergiou87:
Tested it with 2.9.13-test1 and all worked good! 🎉

PR #14288, opened 2022-04-05
Add link to the OS notification settings

sergiou87:
## Description

This PR adds a link under our "Enable notifications" settings to the system notification settings. We plan to improve this so that we show this link only if we detect the app has no permission to display notifications.

### Screenshots

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/preferences/advanced.tsx
+++ b/app/src/ui/preferences/advanced.tsx
@@ -7,6 +7,7 @@ import { UncommittedChangesStrategy } from '../../models/uncommitted-changes-str
 import { RadioButton } from '../lib/radio-button'
 import { isWindowsOpenSSHAvailable } from '../../lib/ssh/ssh'
 import { enableHighSignalNotifications } from '../../lib/feature-flag'
+import { isWindows10OrLater } from '../../lib/get-os'
 
 interface IAdvancedPreferencesProps {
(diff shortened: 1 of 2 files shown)


stephdnisagamer83:
~~app/src/ui/preferences/advanced.tsx~~

app/src/ui/preferences/advanced.tsx
```diff
@@ -188,6 +188,10 @@ export class Advanced extends React.Component<
       return null
     }
 
+    const notificationSettingsURL = __DARWIN__
+      ? 'x-apple.systempreferences:com.apple.preference.notifications'
+      : 'ms-settings:notifications'
```
tsvetilian-ty: @sergiou87 I'm not sure, but I think the `ms-settings` protocol is supported by Windows 10 and above, so for all versions below that, the link to 'Notifications Settings' won't work 😢

sergiou87: Oh, thanks! 🙇‍♂️  Shouldn't be a huge issue since we don't have many users in older versions of Windows, but I will take a look 🤞

tsvetilian-ty:
@sergiou87 love the changes 🔝

tidy-dev (approved):
:sparkles: Makes sense and tested on macOs well. :)

PR #14289, opened 2022-04-05
Markdown Close Keyword Filter

tidy-dev:
## Description
This pr adds a markdown filter for the close keyword.
 
It looks for patterns like Closes #1234 and wraps the Closes keyword in a span that can be styled and has the title attribute.

Notable differences from dotcom:
1. Does not verify the mentioned issue exists (no api call - increased performance) - leading to false positives.
2. Because no api call is done about the issue references, we can not determine if it is an issue or a pull request. Thus the tooltip in dotcom will say "This pull request closes **issue** `#1234`" where this filter omits the "issue" descriptor. Additionally, we just make use of the issue reference as it is given and do no owner/repo verification.

To follow up on:
1) The markdown is sandboxed out of the scope our our react components. Thus, I could not easily make use of our tooltip component. This means these tool tips are the default system styles for the `title` attribute. It would be nice to explore the effort to make them consistent with the rest of the app.
2) Our issue mention filter does do some minimal owner/repo verification against the markdown context repo. This could be replicated here.

### Screenshots

image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/markdown-filters/close-keyword-filter.ts
+++ b/app/src/lib/markdown-filters/close-keyword-filter.ts
@@ -26,7 +26,7 @@ export function isIssueClosingContext(markdownContext: MarkdownContext) {
  * Closes keywords are close, closes, closed, fix, fixes, fixed, resolve,
  * resolves, and resolved.
  *
- * Issue reference can be plain test like #1234 or can bea  pasted issue link
+ * Issue reference can be plain test like #1234 or can be a pasted issue link
  * like https://github.com/owner/repo/issues/1234.
  *
  * Example:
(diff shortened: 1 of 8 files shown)


app/src/lib/markdown-filters/close-keyword-filter.ts
```diff
@@ -0,0 +1,190 @@
...
+ * reference. It replaces the closes keywords it with a span to be styled and
+ * provide a tooltip indicating the markdown will close the referenced issue.
+ *
+ * Markdown that can have these keywords are pull request bodies and commit
+ * messages.
+ *
+ * Closes keywords are close, closes, closed, fix, fixes, fixed, resolve,
+ * resolves, and resolved.
+ *
+ * Issue reference can be plain test like #1234 or can bea  pasted issue link
```
sergiou87: Nit
```suggestion
 * Issue reference can be plain test like #1234 or can be a pasted issue link
```

sergiou87 (dismissed):
Only found a typo in a comment, everything else looks great and it works as expected 😄 :shipit:

sergiou87 (dismissed):
✨

sergiou87 (approved):
🐸 🕴️

PR #14304, opened 2022-04-07
Re-run Individual/Failed action jobs - UI (non-functional)

tidy-dev:
## Description
This is the first of a few PR's to introduce the ability to rerun individual and failed action checks. This PR is simply to introduce the UI interaction for the functionality. Thus, it is feature flagged to development. Future PR's to add the logic to actually rerun individual/vs failed jobs. The little bit of logic changed in the Re-run dialog will evolve with future PRs.

Also, added a new dev/test menu item to show a mock Pull Request Checks Failed Dialog for easier testing to make sure I don't introduce regressions there. 

### Screenshots

https://user-images.githubusercontent.com/75402236/162270751-c6a719e9-ec4b-4d75-a741-4cd051a854b8.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-list-item.tsx
+++ b/app/src/ui/check-runs/ci-check-run-list-item.tsx
@@ -49,7 +49,7 @@ interface ICICheckRunListItemProps {
 }
 
 interface ICICheckRunListItemState {
-  readonly mouseOver: boolean
+  readonly mouseEnter: boolean
 }
 
 /** The CI check list item. */
@@ -59,7 +59,7 @@ export class CICheckRunListItem extends React.PureComponent<
 > {
   public constructor(props: ICICheckRunListItemProps) {
     super(props)
-    this.state = { mouseOver: false }
+    this.state = { mouseEnter: false }
   }
 
   private toggleCheckRunExpansion = () => {
@@ -83,14 +83,14 @@ export class CICheckRunListItem extends React.PureComponent<
     this.props.onRerunJob?.(this.props.checkRun)
   }
 
-  private onMouseOver = () => {
-    if (!this.state.mouseOver) {
-      this.setState({ mouseOver: true })
+  private onMouseEnter = () => {
+    if (!this.state.mouseEnter) {
+      this.setState({ mouseEnter: true })
     }
   }
 
   private onMouseOut = (e: React.MouseEvent) => {
-    this.setState({ mouseOver: false })
+    this.setState({ mouseEnter: false })
   }
 
   private renderCheckStatusSymbol = (): JSX.Element => {
@@ -162,10 +162,10 @@ export class CICheckRunListItem extends React.PureComponent<
 
   private renderJobRerun = (): JSX.Element | null => {
     const { checkRun, onRerunJob } = this.props
-    const { mouseOver } = this.state
+    const { mouseEnter } = this.state
 
     if (
-      !mouseOver ||
+      !mouseEnter ||
       onRerunJob === undefined ||
       !enableReRunFailedAndSingleCheckJobs()
     ) {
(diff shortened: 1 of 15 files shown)


app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
```diff
@@ -168,10 +182,15 @@ export class CICheckRunRerunDialog extends React.Component<
...
+    const failed = this.props.failedOnly
+      ? __DARWIN__
+        ? 'Failed '
+        : 'failed '
+      : ''
     return (
       <Dialog
         id="rerun-check-runs"
-        title={__DARWIN__ ? 'Re-run Checks' : 'Re-run checks'}
+        title={__DARWIN__ ? `Re-run ${failed}Checks` : 'Re-run ${failed}checks'}
```
tidy-dev:
```suggestion
        title={__DARWIN__ ? `Re-run ${failed}Checks` : `Re-run ${failed}checks`}
```

app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -58,6 +74,25 @@ export class CICheckRunListItem extends React.PureComponent<ICICheckRunListItemP
...
+  private rerunJob = (e: React.MouseEvent) => {
+    e.stopPropagation()
+    if (this.props.checkRun.actionJobSteps === undefined) {
+      return
+    }
+
+    this.props.onRerunJob?.(this.props.checkRun)
+  }
+
+  private onMouseOver = () => {
```
niik: Could we use `mouseEnter` here instead of mouseOver?

niik (changes requested):

app/src/ui/check-runs/ci-check-run-list-item.tsx
tidy-dev: I will give it a shot, but I was thinking mouse enter only triggers on the div and mouse over also triggers on child elements.. but maybe it won't matter.

tidy-dev: Seems to work fine. For some reason I was thinking that it might not trigger all the time because of child elements taking up space, but this works just as well (and less events :)).

niik: ~~Oh you're probably right. I was thinking mouseenter would bubble but it doesn't. Chances are `mouseover` will work just fine here but it's probably better to use your initial approach. Sorry for the distraction~~

Edit: Typed without thinking

app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -58,6 +74,25 @@ export class CICheckRunListItem extends React.PureComponent<ICICheckRunListItemP
...
+    this.props.onRerunJob?.(this.props.checkRun)
+  }
+
+  private onMouseEnter = () => {
+    if (!this.state.mouseEnter) {
+      this.setState({ mouseEnter: true })
+    }
+  }
+
+  private onMouseOut = (e: React.MouseEvent) => {
```
niik: Can we rename this to `onMouseLeave` to match the event please?

app/src/ui/check-runs/ci-check-run-list-item.tsx
```diff
@@ -42,10 +43,25 @@ interface ICICheckRunListItemProps {
     checkRun: IRefCheck,
     step: IAPIWorkflowJobStep
   ) => void
+
+  /** Callback to rerun a job*/
+  readonly onRerunJob?: (checkRun: IRefCheck) => void
+}
+
+interface ICICheckRunListItemState {
+  readonly mouseEnter: boolean
```
niik: I think I would prefer a more descriptive state name here like `isMouseOver`

niik (changes requested):

niik (approved):

PR #14310, opened 2022-04-08
Add ability to re-run a single job and failed jobs

tidy-dev:
Based on #14304

## Description
This adds the api calls and links them up to the re-request dispatcher call to add the functionally to rerun an individual job or failed jobs.  This is still flagged to development.

Remaining work:
- Rework the wording on the re-run dialog for single/failed jobs - probably follow dotcom wording.
- If all checks are successful, don't have dropdown button for failed option.
- If a check is pending, disable individual rerun button for all checks in that workflow run. On dotcom it is actually just not shown, but since we show it muted with tooltip for other non-rerunnable circumstances, I think it makes sense to do it here as well. Maybe with a custom message -> "Cannot rerun while other checks in this workflow are pending".

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1237,7 +1237,7 @@ export class API {
    * Re-run all of the failed jobs and their dependent jobs in a workflow run
    * using the id of the workflow run.
    */
-  public async rerunFailedJob(
+  public async rerunFailedJobs(
     owner: string,
     name: string,
     workflowRunId: number
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1234,6 +1234,45 @@ export class API {
   }
 
   /**
+   * Re-run all of the failed jobs and their dependent jobs in a workflow run
+   * using the id of the workflow run.
+   */
+  public async rerunFailedJob(
+    owner: string,
+    name: string,
+    workflowRunId: number
+  ): Promise<boolean> {
+    const path = `/repos/${owner}/${name}/actions/runs/${workflowRunId}/rerun-failed-jobs`
+    const response = await this.request('POST', path)
+    try {
+      return response.ok
+    } catch (err) {
+      log.debug(
+        `Failed to rerun failed workflow jobs for (${owner}/${name}): ${workflowRunId}`
+      )
+    }
+    return false
+  }
+
+  /**
+   * Re-run a job and its dependent jobs in a workflow run.
+   */
+  public async rerunJob(
+    owner: string,
+    name: string,
+    jobId: number
+  ): Promise<boolean> {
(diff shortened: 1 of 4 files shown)


app/src/lib/api.ts
```diff
@@ -1233,6 +1233,45 @@ export class API {
...
+  ): Promise<boolean> {
+    const path = `/repos/${owner}/${name}/actions/runs/${workflowRunId}/rerun-failed-jobs`
+    const response = await this.request('POST', path)
+    try {
+      return response.ok
+    } catch (err) {
+      log.debug(
+        `Failed to rerun failed workflow jobs for (${owner}/${name}): ${workflowRunId}`
+      )
+    }
```
niik: Response.ok is just a plain old property, accessing it can never throw, I'm guessing you meant to guard again `this.request` throwing?

```suggestion
return this.request('POST', path)
      .then(x => x.ok)
      .catch(err => {
        log.debug(
          `Failed to rerun failed workflow jobs for (${owner}/${name}): ${workflowRunId}`,
          err
        )
        return false
      })
```

app/src/lib/api.ts
```diff
@@ -1233,6 +1233,45 @@ export class API {
...
+   */
+  public async rerunJob(
+    owner: string,
+    name: string,
+    jobId: number
+  ): Promise<boolean> {
+    const path = `/repos/${owner}/${name}/actions/jobs/${jobId}/rerun`
+    const response = await this.request('POST', path)
+    try {
+      return response.ok
```
niik: Same comment here about this never throwing

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -2522,15 +2523,49 @@ export class Dispatcher {
...
+
+      // There could still be failed ones that are not action and only way to
+      // rerun them it rerun their whole check suite
+      if (cr.checkSuiteId !== null) {
+        checkSuiteIds.add(cr.checkSuiteId)
+      }
+    }
+
+    for (const id of workflowRunIds) {
+      if (id === null) {
```
niik: According to the type of `workflowRunIds` (`Set<number>`). This can't ever be null. If it can be null we need to update the type of `id` in `IAPIWorkflowRun` and update the set accordingly.

niik (changes requested):

app/src/ui/dispatcher/dispatcher.ts
tidy-dev: Shouldn't be able to be null -> Left over from previous solution and didn't realize new solution didn't need it.

niik (approved):

roysterbrock23 (commented):
…

PR #14320, opened 2022-04-11
Better markdown tooltips w niik ideas

tidy-dev:
## Description
This is a follow up to the closed keyword markdown filter (#14289). It made use of the title attribute in order to show the "This pull request closes x" tooltip.  Thanks to @niik, this pull request adds a way for us to make use of Desktop's beautiful custom tooltips in our markdown.

### Screenshots

image

(before)
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/lib/sandboxed-markdown.tsx
+++ b/app/src/ui/lib/sandboxed-markdown.tsx
@@ -9,6 +9,9 @@ import {
 } from '../../lib/markdown-filters/node-filter'
 import { GitHubRepository } from '../../models/github-repository'
 import { readFile } from 'fs/promises'
+import { Tooltip } from './tooltip'
+import { createObservableRef } from './observable-ref'
+import { getObjectId } from './object-id'
 
 interface ISandboxedMarkdownProps {
   /** A string of unparsed markdown to display */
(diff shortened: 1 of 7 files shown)


app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -159,10 +183,32 @@ export class SandboxedMarkdown extends React.PureComponent<ISandboxedMarkdownPro
...
   }
 
+  private setupTooltips(frameRef: HTMLIFrameElement) {
+    if (frameRef.contentDocument === null) {
+      return
+    }
+
+    const tooltipElements = new Array<HTMLElement>()
+
+    for (const e of frameRef.contentDocument.querySelectorAll('[aria-label]')) {
```
tidy-dev, line 198: I am wondering if we should be more specific on this (like search for a`tooltip` class  or data attribute and then look for an aria-label), but I think aria-label does make sense for the context and witout a competing need for it at the moment... I am inclined to leave it be.

Hovemeyer5:
Testing a comment

Hovemeyer5 (commented):
(Ignore.. testing review dialog in beta)

niik:
Given that I played a part in writing this I'm gonna request a review from @sergiou87 for when he's back from vacation

sergiou87 (approved):
Really nice! I learned a lot from this PR 🤓  And it works great too :shipit:

PR #14322, opened 2022-04-11
Update re run dialog

tidy-dev:
Based on #14310

## Description
This PR updates the re-run dialog to have consistent copy to dotcom about re-running failed and single checks. Dotcom however is only dealing with action jobs and we are dealing with any kind of check; thus, our copy is not identical.

Additionally, dotcom action view uses the circle status octicons in the actions view and the non-circle in the pull request view. Since, our view more closely aligns with the pull-request view, we maintained the non-circle octicons in our lists. I chose to carry that forward to the re-run dialog (but, if someone has strong feelings that this re-run dialog's icons should match dotcoms, I am more than willing to switch). 

Also added in this PR:
- Added enterprise support flagging - right now just flagged to dotcom as we don't know what version this is to be release to enterprise yet.
- Made re-run button not be a dropdown if all checks are successful.

Things to follow up on:
- Not ~~sure if it was introduced as part of this work, but there is clearly a bug on the count of "non-rerunnable" jobs as can be seen when I look at rerunning all failed jobs it says there are 4 non-rerunnable failed jobs and it can rerun 4 failed jobs and there are only 4, not 8~~.
   - Went ahead and addressed on this PR since it was a one line change
- ~~The loader flashing when rerunning one individual job is annoying..~~ It is less annoying now..
- ~~Limited windows testing~~

### Screenshots
https://user-images.githubusercontent.com/75402236/162829574-33c3ecc8-2748-4d0e-9b43-fe4b5ff3e1ac.mov

## Release notes
Notes: [Improved] Reduced noise on the rerun dialog user interface.

### DIFF
--- a/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
+++ b/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
@@ -160,7 +160,7 @@ export class CICheckRunRerunDialog extends React.Component<
       ) : (
         'these workflows'
       )
-    const dependentAdj = this.props.checkRuns.length === 1 ? "it's" : 'their'
+    const dependentAdj = this.props.checkRuns.length === 1 ? "its" : 'their'
 
     return (
       <div className="re-run-dependents-message">
--- a/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
+++ b/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
@@ -193,7 +193,7 @@ export class CICheckRunRerunDialog extends React.Component<
         <Octicon symbol={OcticonSymbol.alert} />
 
         {`${warningPrefix}. A check run cannot be re-run if the check is more than one month old,
-          the check or it's dependent has not completed, or the check is not configured to be
+          the check or its dependent has not completed, or the check is not configured to be
           re-run.`}
       </div>
     )
--- a/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
+++ b/app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
@@ -12,7 +12,6 @@ import {
 } from '../../lib/api'
 import { Octicon } from '../octicons'
 import * as OcticonSymbol from './../octicons/octicons.generated'
-import { Row } from '../lib/row'
 import { encodePathAsUrl } from '../../lib/path'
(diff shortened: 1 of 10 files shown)


app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
```diff
@@ -123,95 +122,135 @@ export class CICheckRunRerunDialog extends React.Component<
...
+      return null
+    }
+
+    const name =
+      this.props.checkRuns.length === 1 ? (
+        <strong>{this.props.checkRuns[0].name}</strong>
+      ) : (
+        'these workflows'
+      )
+    const dependentAdj = this.props.checkRuns.length === 1 ? "it's" : 'their'
```
niik: Should this be [its](https://www.grammarly.com/blog/its-vs-its/)?

```suggestion
    const dependentAdj = this.props.checkRuns.length === 1 ? "its" : 'their'
```

app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
```diff
@@ -123,95 +122,135 @@ export class CICheckRunRerunDialog extends React.Component<
...
+            this.props.failedOnly ? 'failed ' : ''
+          }${pluralize} that cannot be re-run`
     return (
-      <Row className="non-re-run-info warning-helper-text">
+      <div className="non-re-run-info warning-helper-text">
         <Octicon symbol={OcticonSymbol.alert} />
 
         {`${warningPrefix}. A check run cannot be re-run if the check is more than one month old,
-          the check has not completed, or the check is not configured to be
+          the check or it's dependent has not completed, or the check is not configured to be
```
niik:
```suggestion
          the check or its dependent has not completed, or the check is not configured to be
```

niik (changes requested):

niik (approved):

nattha36 (commented):
#

app/src/ui/check-runs/ci-check-run-rerun-dialog.tsx
nattha36: > Should this be [its](https://www.grammarly.com/blog/its-vs-its/)?
> 
> 
> 
> ```suggestion
> 
>     const dependentAdj = this.props.checkRuns.length === 1 ? "its" : 'their'
> 
> ```

PR #14335, opened 2022-04-13
Release 2.9.14-beta1

niik:
## Description
Looking for the PR for the upcoming beta of the v2.9.14 series? Well, you've just found it, congratulations!

This builds on 2.9.13 but contains dugite 1.106.0 which includes Git 2.35.2 and Git LFS 3.1.2

## Release checklist

None
No Changes

### DIFF
--- a/app/test/setup-test-framework.ts
+++ b/app/test/setup-test-framework.ts
@@ -3,3 +3,6 @@ jest.setTimeout(10000)
 
 import 'jest-extended'
 import 'jest-localstorage-mock'
+
+delete process.env['LOCAL_GIT_DIRECTORY']
+delete process.env['GIT_EXEC_PATH']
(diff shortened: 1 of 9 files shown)


tidy-dev (dismissed):

app/test/setup-test-framework.ts
```diff
@@ -3,3 +3,6 @@ jest.setTimeout(10000)
 
 import 'jest-extended'
 import 'jest-localstorage-mock'
+
+delete process.env['LOCAL_GIT_DIRECTORY']
```
tidy-dev, line 7: Why is this needed? I figured tests must be breaking.. but just curious.

niik: Yeah, see https://github.com/desktop/desktop/pull/14335/commits/2e80e8bc952366d654ead095381d6b0a15f4c6d0

> I had launched VS Code via Desktop and it had inherited these causing tests to use the prod version dugite

niik: So this change isn't necessary on CI but it's gonna bite me again one of these days so I'd rather fix it now

tidy-dev (approved):

Issue #14336, opened 2022-04-13
"Potentially unsafe" repository / Cannot Trust Repository after latest update


wghughes95:
### The problem

The latest update stopped recognising my repos as git repositories. Tried re-adding them as existing repos and also a full reboot of my machine but it just complains that they don't appear to be git repos and if I would like to create a git repo there instead. My repositories are located within my Ubuntu WSL2 filesystem, and are able to be picked up with Sourcetree. I prefer the GitHub Desktop UI but I'm having to use Sourcetree as of now because of this issue. 

### Release version

2.9.13 (x64)

### Operating system

Windows 10 (WSL 2 - Ubuntu 20.04)

weihong19971222:
+1

niik:
Hey @wghughes95, sorry to hear you're having problems. The latest version of Desktop shipped an update of Git containing a fix for a security vulnerability. I imagine that's what's causing issues for you here. 

Could you upload a log file from GitHub Desktop so that I could get some more information about this issue? To access the log files go to the file menu in GitHub Desktop and select `Help` > `Show Logs in Explorer`.

It would also be very helpful if you could ensure you're running one of the patched Git versions (v2.30.3, v2.31.2, v2.32.1,
v2.33.2, v2.34.2, and v2.35.2) in your WSL shell to see if you're encountering the same problem there.

You should be able to work around this by explicitly configuring your repository as a "safe directory".

```
git config --global --add safe.directory [PATH TO YOUR REPO]
```

cc @dscho @vdye For visibility as we troubleshoot

aftabdess:
Same here...

aftabdess:
[2022-04-13.desktop.production.log](https://github.com/desktop/desktop/files/8480100/2022-04-13.desktop.production.log)

niik:
@aftabdess Thanks for that log file. In your case it seems that you're not using WSL and that the vulnerability protection is working as intended:

```
fatal: unsafe repository ('G:/Official/DessBrowserApp' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory G:/Official/DessBrowserApp
```

Can you try adding the directory as per the instructions above?

wghughes95:
[2022-04-13.desktop.production.log](https://github.com/desktop/desktop/files/8480125/2022-04-13.desktop.production.log)

Here is my latest log. Just updated my git version WSL from 2.25.1 to 2.35.2 but that hasn't solved it yet.

[...]

PR #14344, opened 2022-04-13
Alert users of potentially unsafe directories

### LINKED ISSUE #14352: Must detect ownership issues and offer to resolve them
@TalicZealot For you it would appear the vulnerability protection is working as intended. Could you make sure that the `D:/Programming/Khaos-Bot-Script` folder is owned by your user or if that's not an option you can add it as an exception by running `git config --global --add safe.directory "D:/Programming/Khaos-Bot-Script"`

_Originally posted by @niik in https://github.com/desktop/desktop/issues/14336#issuecomment-1097847383_

GitHub Desktop can detect such issues, should provide a correct diagnostic, and it should offer to resolve them for us

niik:
Related to #14336 
Closes #14352

## Description

In the latest production release we upgraded Git to 2.32.1 which contained a protections against a security vulnerability when a repository was owned by a different user.

This PR introduces support in GitHub Desktop for detecting this specific condition and letting users chose to add the repository to an allow-list just like the CLI does:

```
fatal: unsafe repository ('C:/' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory C:/
```

### Screenshots

image

## Release notes

Notes: [Improved] Surface Git's warning about unsafe directories and provide a way to trust repositories not owned by the current user

### DIFF
--- a/app/src/lib/git/config.ts
+++ b/app/src/lib/git/config.ts
@@ -149,6 +149,14 @@ export async function setGlobalConfigValue(
   return setConfigValueInPath(name, value, null, env)
 }
 
+/** Set the global config value by name. */
+export async function addGlobalConfigValue(
+  name: string,
+  value: string,
+): Promise<void> {
+  await git(['config', '--global', '--add', name, value], __dirname, 'addGlobalConfigValue')
+}
+
 /**
  * Set config value by name
  *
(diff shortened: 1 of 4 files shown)


tidy-dev (approved):

bompus:
While providing the automatic fix, please see https://github.com/desktop/desktop/issues/14336#issuecomment-1098287711 regarding how I needed to use `%(prefix)/` for UNC paths ( paths starting with // ) to prevent Git GUI from erroring when loading the repo. I'm pretty sure the error came from the CLI and GUI was just displaying it, but not certain.

PR #14368, opened 2022-04-14
Add UNC path prefix before adding to safe directory list

niik:
## Description

This matches the fix for Git CLI over at https://github.com/git-for-windows/git/pull/3791. Whenever we encounter a UNC path (i.e path starting with `//`) from Git's unsafe error warning we'll prepend `%(prefix)/` to get Git to ultimately resolve it to the same path as the repository.

So when we encounter an error message like this:

```
unsafe repository ('//wsl/something' is owned by someone else)
To add an exception for this directory, call:
    git config --global --add safe.directory %(prefix)//wsl/something
```

We'll extract `//wsl/something` from the first line and prepend it with `%(prefix)/`. The reason we're not just plucking the path from the last line (i.e. the "example") is that that path will be quoted for use in a shell whereas we'll rely on Node to do our quoting for us.

Note that this will not have any effect until we're able to upgrade to a version of Git which supports `%(prefix)` (2.34.0 or higher). We're working with the fine folks on the Git side to hopefully get our hands on a Git for Windows version we can sign and ship today.

## Release notes

Notes: [Fixed] Support adding repositories located on network drives (such as NAS, WSL, SMB etc) to the list of safe directories in Git

cc @vdye @derrickstolee

### DIFF
--- a/app/src/lib/git/config.ts
+++ b/app/src/lib/git/config.ts
@@ -161,6 +161,21 @@ export async function addGlobalConfigValue(
   )
 }
 
+/**
+ * Adds a path to the `safe.directories` configuration variable if it's not
+ * already present. Adding a path to `safe.directory` will cause Git to ignore
+ * if the path is owner by a different user than the current.
+ */
+export async function addSafeDirectory(path: string) {
+  // UNC-paths on Windows need to be prefixed with `%(prefix)/`, see
+  // https://github.com/git-for-windows/git/commit/e394a16023cbb62784e380f70ad8a833fb960d68
+  if (__WIN32__ && path[0] === '/') {
+    path = `%(prefix)/${path}`
+  }
+
+  addGlobalConfigValueIfMissing('safe.directory', path)
+}
+
 /** Set the global config value by name. */
 export async function addGlobalConfigValueIfMissing(
   name: string,
--- a/app/src/ui/add-repository/add-existing-repository.tsx
+++ b/app/src/ui/add-repository/add-existing-repository.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react'
 import * as Path from 'path'
 import { Dispatcher } from '../dispatcher'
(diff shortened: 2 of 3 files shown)


app/src/lib/git/config.ts
```diff
@@ -161,6 +161,21 @@ export async function addGlobalConfigValue(
...
+ */
+export async function addSafeDirectory(path: string) {
+  // UNC-paths on Windows need to be prefixed with `%(prefix)/`, see
+  // https://github.com/git-for-windows/git/commit/e394a16023cbb62784e380f70ad8a833fb960d68
+  if (__WIN32__ && path[0] === '/') {
+    path = `%(prefix)/${path}`
+  }
+
+  addGlobalConfigValueIfMissing('safe.directory', path)
+}
```
derrickstolee, line 177: Chiming in from Git's perspective to say that this looks right to me. Thanks!

tidy-dev (approved):
✨  From my understanding of following the threads, this makes sense. 🙂

mdell-seradex:
I was just looking at commit [862f427](https://github.com/desktop/desktop/pull/14368/commits/862f427b60b162b48ea761a36a1525eb0359f2ef).
It looks to have a minor mistake in the comment to me.
`if the path is owner by a different user than the current`  Specifically the **is owner** part.

fabiofdsantos:
Thaaaanks 👍🏻

raywhite714:
[19dENFt4wVwos6xtgwStA6n8bbA57WCS58_en_usd_2020-12-18_2022-03-29_1.pdf](https://github.com/desktop/desktop/files/8546480/19dENFt4wVwos6xtgwStA6n8bbA57WCS58_en_usd_2020-12-18_2022-03-29_1.pdf)

mdell-seradex:
@raywhite714 I don't think you wanted to post that.

PR #14374, opened 2022-04-14
Cloning into, and adding multiple, unsafe locations

niik:
## Description

This is a follow up to https://github.com/desktop/desktop/pull/14368 and deals with alternate code paths to adding a repository. In #14344 we added support for surfacing Git's "unsafe directory" warning in the "add-local-repo" dialog and the "missing-repository" view.

Unfortunately we had checks in place both post-clone and when adding repositories via drag-and-drop that attempted to verify that the path was a valid repository. These checks were now confused by the unsafe error.

In order to address this I've extended the `getRepositoryType` PR to cover the role of `validatedRepositoryPath` so that we can have one unified code path for probing the type of repository inside a path.

If we detect an "unsafe" path when adding a repository (either after clone or through drag/drop) we'll skip the normal verifications and add it regardless but with the `missing` flag set to true from the beginning. This will cause repositories added this way to show up in the repository list but when selected users will see the same warning as in #14344.

## Release notes

Notes: no-notes (I figure this can go underneath the log entry for #14344)


app/test/unit/git/rev-parse-test.ts
```diff
@@ -36,61 +35,25 @@ describe('git/rev-parse', () => {
...
 
-    it('should return null when not run inside a working directory', async () => {
-      const result = await getTopLevelWorkingDirectory(os.tmpdir())
-      expect(result).toBeNull()
-    })
-
-    it('should resolve top level directory run inside the .git folder', async () => {
-      const p = path.join(repository.path, '.git')
-      const result = await getTopLevelWorkingDirectory(p)
-      expect(result).toBe(p)
```
niik, line 93: This test was removed because it was incorrect to begin with. See how it's saying that it "should resolve top level directory run inside the .git" but then in the `expect()` we check for it to equal the `.git` path, not the working directory path.

tidy-dev (approved):
✨  Makes sense

PR #14376, opened 2022-04-14
Update create repository dialog to understand unsafe paths

niik:
## Description

This is hopefully the last piece to the unsafe path handling fixes started in (#14344, #14368, and #14374).

This deals with the Create Repository dialog and makes it aware of the fact that directories can now be considered unsafe. It also fixes a bug where we would only consider the path, not the path + name when checking for a repository. That fix also applies to the "Add existing repository" shortcut which previously would only send along the base path, not the full path.

### Screenshots

image

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/add-repository/create-repository.tsx
+++ b/app/src/ui/add-repository/create-repository.tsx
@@ -541,15 +541,15 @@ export class CreateRepository extends React.Component<
   }
 
   private onAddRepositoryClicked = () => {
-    if (this.state.path === null) {
-      // Shouldn't be able to even get here if path is null.
-      return
-    }
+    const { path, name } = this.state
 
-    return this.props.dispatcher.showPopup({
-      type: PopupType.AddRepository,
-      path: this.state.path,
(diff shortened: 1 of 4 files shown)


app/src/ui/add-repository/create-repository.tsx
```diff
@@ -146,27 +163,40 @@ export class CreateRepository extends React.Component<
...
   }
 
+  private async updateIsRepository(path: string, name: string) {
+    const fullPath = Path.join(path, sanitizedRepositoryName(name))
+    const isRepository = await isGitRepository(fullPath)
+
+    // Only update isRepository if the path is still the same one we were using
+    // to check whether it looked like a repository.
+    this.setState(state =>
+      state.path === path && state.name === name ? { isRepository } : null
```
tidy-dev, line 196: TIL about optionally setting the state this way.

tidy-dev (approved):
✨  Looks good.

Issue #14255, opened 2022-03-29
Ability to customize "commit message length" nag


dangerismycat:
### Search for existing feature requests

- [X] I have searched the issue tracker and confirmed that I was unable to find existing issues for the feature I am requesting.

### The feature request

I'd like the ability to customize the nag about the commit message length. I'd like to be able to specify the number of characters or turn it off completely.

### Proposed solution

Different repos have different standards around things like commit message length. This is the sort of thing we can configure with linters and code formatters; we should allow per-repo configuration of this nag as well.

strughle333:
yes ready

steveward:
Requested in #14722:

>This may be "right" or "wrong" either way it's way to aggressive to always come up and no option to disable it.
Maybe our team thinks 49 or 51 characters is the magic number... don't force this on us.

DanielRouxSA:
I'd like to request this as well. In our team's case, we have no reason to limit the length of our commit messages, so the message is superfluous to us.

steveward:
From https://github.com/desktop/desktop/issues/16456:

>Currently the app is showing lightbulb icon when 50 characters met but sometimes you do need a bit more and it's nice to know the point where the commit message will be cutted (I guess it's at 72 symbols) and the rest of it will go to commit description.

steveward:
Closing this out since https://github.com/desktop/desktop/pull/16643 will soon be merged to remove the commit length hint entirely.

niik:
Hey @dangerismycat, as of https://github.com/desktop/desktop/pull/17370 it's possible to disable the commit message length warning and it's off by default for new installs.

![Image](https://github.com/user-attachments/assets/2fc2a9c5-17c4-4def-a298-10bb34255193)

Given that the ability to turn it off was one of your possible solutions I'm going to close this out.

PR #14406, opened 2022-04-19
Allow disabling commit summary length hint

lhvy:
Partially addresses #14255

## Description

- Adds a new section to the `Prompts` preferences tab to allow disabling the commit summary length hint that appears when a commit summary is over 50 characters

#14255 also mentioned customising this option per-repository, and setting a custom character limit for the hint. I'm unsure of the value of a custom number since the hint is pretty explicit about the 50 character limit. In terms of the per-repo configuration, is this useful and has it been done before in the codebase? That seems more difficult and I may need some assistance to start that change.

### Screenshots

Here is a photo of the hint in question:

Screenshot of the hint

Here is the updated `Prompts` tab which allows disabling the hint:

image

## Release notes

Notes: [IMPROVED] Add an option to disable the commit summary length hint

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -195,6 +195,9 @@ export interface IAppState {
   /** Should the app prompt the user to confirm a force push? */
   readonly askForConfirmationOnForcePush: boolean
 
+  /** Should the app show a hint when the commit summary is over 50 characters? */
+  readonly showSummaryLengthHint: boolean
+
   /** How the app should handle uncommitted changes when switching branches */
   readonly uncommittedChangesStrategy: UncommittedChangesStrategy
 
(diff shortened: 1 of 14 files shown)


lhvy:
If this PR is suffice for the mentioned issue, let me know and I'll edit my PR to resolve the issue. If anything can be changed or added, please let me know! :)

sergiou87:
Thank you for your contribution @lhvy, but unfortunately, we try to avoid adding settings unless strictly necessary 😕  In this case, we believe this hint is subtle enough to not justify the new setting.

We're sorry if this feels like a waste of your time 😢  In general, I think it's better to make your argument in an issue (like https://github.com/desktop/desktop/issues/14255 in this case) and wait until you get a 👍  from the team to implement it.

PR #14458, opened 2022-04-26
Show All release notes since last update.

tidy-dev:
## Description
Every so often we will release an update and the quickly determine we need to release a hotfix. What this means for release notes is if a user is on version x.x.1 and we update to x.x.2 with a cool new feature. The release notes will be for x.x.2 and talk about the cool new feature. However if we hotfix to x.x.3 before a user updates to x.x.2 updates, they get a release note about the fix and not about the cool new feature.

This PR attempts to combat that by checking the users current version and merging the release notes set between their current version in the latest. 

It limits releases to the past 10 releases or past 90 days (which ever is sooner). This working upon the assumption that, because of our auto updates, it is very unlikely scenario for someone to be 90 days or 10 versions behind and, even if so, there would be so many release notes, a user likely would not bother scrolling through them anyways. 

Other notes:
- Plan to investigate adding the ability to showcase by having the ability to add a markdown pretext to release notes.
- Plan to remove old "showcase" logic around drag/drop features and split views.

Testing this: 
In Dev environment, there is Help > Show popup > Release notes to assist in testing (I updated this to use actually release notes instead of the dummy ones so it would be closer to actual testing)

Adding that `x.x.x ?? ` in the semver construction is the easiest way up date what version the release notes thinks the app is running under. But you can also update the package.json and run yarn:build, to actually change the version the app believes it is running under.
image

### Screenshots
Last 90 days of releases:
Multiple versions

One release is fairly unchanged (max-width/height a bit more lax):
image

## Release notes

Notes: [Improved] User can see all releases notes between their current version and the latest update.

### DIFF
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -6,6 +6,7 @@ import {
 } from '../models/release-notes'
 import { getVersion } from '../ui/lib/app-proxy'
 import { formatDate } from './format-date'
+import { offsetFromNow } from './offset-from'
 import { encodePathAsUrl } from './path'
 
 // expects a release note entry to contain a header and then some text
@@ -111,13 +112,20 @@ export async function getChangeLog(
 export async function generateReleaseSummary(): Promise<
   ReadonlyArray<ReleaseSummary>
 > {
-  const releases = await getChangeLog()
+  const lastTenReleases = await getChangeLog()
   const currentVersion = new semver.SemVer(getVersion())
-  const releaseSinceCurrentVersion = releases.filter(r =>
-    semver.gte(new semver.SemVer(r.version), currentVersion)
+  const recentReleases = lastTenReleases.filter(
+    r =>
+      semver.gte(new semver.SemVer(r.version), currentVersion) &&
+      new Date(r.pub_date).getTime() > offsetFromNow(-90, 'days')
   )
 
-  return releaseSinceCurrentVersion.map(getReleaseSummary)
+  // We should only be pulling release notes when a release just happened, so
+  // there should be one within the past 90 days. Thus, this is just precaution
+  // to ensure we always show at least the last set of release notes.
+  return recentReleases.length > 0
+    ? recentReleases.map(getReleaseSummary)
+    : [getReleaseSummary(recentReleases[0])]
 }
 
 export const ReleaseNoteHeaderLeftUri = encodePathAsUrl(
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -116,7 +116,7 @@ export async function generateReleaseSummary(): Promise<
   const currentVersion = new semver.SemVer(getVersion())
   const recentReleases = lastTenReleases.filter(
     r =>
-      semver.gte(new semver.SemVer(r.version), currentVersion) &&
+      semver.gt(new semver.SemVer(r.version), currentVersion) &&
       new Date(r.pub_date).getTime() > offsetFromNow(-90, 'days')
   )
 
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -106,10 +106,12 @@ export async function getChangeLog(
   }
 }
 
-export async function generateReleaseSummary(): Promise<ReleaseSummary> {
+export async function generateReleaseSummary(): Promise<
+  ReadonlyArray<ReleaseSummary>
+> {
   const releases = await getChangeLog()
   const latestRelease = releases[0]
-  return getReleaseSummary(latestRelease)
+  return [getReleaseSummary(latestRelease)]
 }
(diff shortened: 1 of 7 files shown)


app/src/lib/release-notes.ts
```diff
@@ -106,10 +109,23 @@ export async function getChangeLog(
...
+      semver.gte(new semver.SemVer(r.version), currentVersion) &&
+      new Date(r.pub_date).getTime() > offsetFromNow(-90, 'days')
+  )
+
+  // We should only be pulling release notes when a release just happened, so
+  // there should be one within the past 90 days. Thus, this is just precaution
+  // to ensure we always show at least the last set of release notes.
+  return recentReleases.length > 0
+    ? recentReleases.map(getReleaseSummary)
+    : [getReleaseSummary(recentReleases[0])]
```
sergiou87: Shouldn't this be `lastTenReleases`?
```suggestion
    : [getReleaseSummary(lastTenReleases[0])]
```

sergiou87 (changes requested):

app/src/lib/release-notes.ts
tidy-dev: Yep.. I renamed variables and overlooked here. Thanks!

sergiou87 (dismissed):
Looks good!! Great work 😄  :shipit:

tidy-dev:
When testing the pretext change log entries, I realized this introduces a bug. In the below, I am upgrading to the "latest" test version. In this scenario, I should only show one latest release.
image

(Resolved -> I switched the comparison from gte to gt because the a user would be one behind what they haven't seen upon opening the release notes, because they have not upgraded yet)

sergiou87 (dismissed):
✨ 🚀 ✨

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -82,11 +82,47 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
...
 
+  /**
+   * If there is just one release, it returns it. If multiple, it merges the release notes.
+   */
+  private getDisplayRelease = () => {
+    if (this.props.newReleases.length === 1) {
+      return this.props.newReleases[0]
+    }
+
+    const earliestRelease = this.props.newReleases.slice(-1)[0]
```
niik: Shouldn't this be `oldestRelease`? While I guess an older release happened earlier than a later release it confused the heck out of me since we're comparing with `latestRelease` below.

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -82,11 +82,47 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
...
+
+    let enhancements = new Array<ReleaseNote>()
+    let bugfixes = new Array<ReleaseNote>()
+
+    for (const r of this.props.newReleases) {
+      enhancements = enhancements.concat(r.enhancements)
+      bugfixes = bugfixes.concat(r.bugfixes)
+    }
```
niik: Given how few releases we'll be iterating over here I think it's worth mapping to get rid of the `let`s and `for`

```suggestion
    const enhancements = this.props.newReleases.flatMap(x => x.enhancements)
    const bugfixes = this.props.newReleases.flatMap(x => x.bugfixes)
```

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -82,11 +82,47 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
     )
   }
 
+  /**
+   * If there is just one release, it returns it. If multiple, it merges the release notes.
+   */
+  private getDisplayRelease = () => {
+    if (this.props.newReleases.length === 1) {
```
niik: With how often we access `this.props.newReleases` in this method I think we should probably destructure it from `this.props` to clean things up a little

niik (changes requested):
Thanks for working on this @tidy-dev! A few small suggestions

app/src/ui/release-notes/release-notes-dialog.tsx
tidy-dev: :D - To me it is relays the same meaning so happy to change it.

app/src/ui/release-notes/release-notes-dialog.tsx
tidy-dev: I forgot flatMap existed!

app/src/ui/release-notes/release-notes-dialog.tsx
niik: Thank you, it's probably just a language barrier thing but `oldest` makes my brain happy!

niik (approved):

PR #14462, opened 2022-04-27
Exploring Pretext Changelog type for showcasing new features

tidy-dev:
Based on #14458

## Description
Something we have done on some releases is call out exciting new feature in very custom ways. For instance, we had a custom popover style information blurb about our recent drag/drop features (cherry-picking, re-ordering, squashing). But, currently we do not have an easy way to more generically showcase new features. This PR explores the possibility of having a release note of type "pretext" in the changelog in which the message would be markdown that is displayed first in the release notes dialog. By being markdown, means we can go as far as to implement inline styles to customize as desired. 

This doesn't preclude us from still making custom call outs as this will only be seen if a user clicks on the "what's new" in the update banner. The idea behind pretext would only be added for very splashing new features.

Here is an example of where I injected the following random markdown:

image

pretext release note

Some thoughts:
1. Even tho this pretext concept really isn't a release note, using a new "pretext" changelog type is a very simple and readily available. Also easily remotely updated.
2. Alternatives:
   a. we could create a new database field/table for our releases.
      - Pro: Database field is cleaner (not usurping the change log purpose). 
      - Con: Database field would require more building out of a way to modify it.
      
   b. we could go non-markdown route and build custom react pretext components. 
      - Pro: Ability to have call-to-actions that deep links within the app. Example: "Cherry-picking -> go to history to check it out now" or triggering a mini tutorial or learn more dialog.
      - Pro: Not likely that much more work than crafting a change log pretext/, but more flexible.
      - Con: Not remotely updatable. 
      - Con: heavy if we don't actually take advantage of in-app linking.
      - Con: Requires pruning (could be automated tho)
3. Could we have more than one thing to highlight/would more than one thing lead to the desire to expand this to a paginated style thing (multiple pretext release notes)? (thinking cherry-pick, reorder, squashing pagination we did)
4. Should showcasing be more prominent? 
   a. Interacted solely through clicking the "what's new" on a release
        -  giving user the choice on whether to learn more (not annoying them)
        -  The what new goes away once updated which can happen from opening/closing the app. Forgoing the chance to click on "what's new". 
       
   b. Make a permanent "what's new" call-to-action somewhere until a user interacts with it or sometime has passed since new release.
       - If we went route of non-markdown custom dialogs, this could be the start of a standard practice of having "call to action" bubbles anywhere that when clicked open a show case dialog. (again eventually prunned - probably having an expiration date on them) Thus, there could be a "What new" bubble button for show cases with new release and/or just "new" bubble buttons that are near the UI of the new feature.
       
   c. Popup a showcase dialog on update (force user to acknowledge)
     - From a discoverability standpoint, this is useful.
     - Dislike forced marketing. :D 
   

Just technical stuff/thoughts to determine (easy to resolve once definite path is known)
1. Do we have any db limitations on release notes that would hinder using markdown/would it need to be escaped?
2. What would the process for adding one of these look like? Currently, to preview/edit, one would have to programmatically inject a pretext release note to view in the dev build. Figured if we continue on this route, we could add logic to have our release notes dialog read in our changelog.json to preview on a dev build.
3. We would want images.. stored in our releases store, or stored as a static asset with the build? Seems indefinitely adding images to the build for things that are only seen once by the user is undesirable.
  
## Release notes
Notes: [Improved] Added ability to have showcasing of features through release notes.

### DIFF
--- a/app/src/lib/release-notes.ts
+++ b/app/src/lib/release-notes.ts
@@ -56,7 +56,14 @@ function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
 export function parseReleaseEntries(
   notes: ReadonlyArray<string>
 ): ReadonlyArray<ReleaseNote> {
-  return notes.map(n => parseEntry(n)).filter(notEmpty)
+  const pretext =
+    '<img style="width: 100%" src="https://user-images.githubusercontent.com/75402236/165329268-4650b6f1-a891-4f26-b123-e52de21f703c.png"/>' +
+    'Adding that `x.x.x ?? ` in the semver construction is the easiest way up date what version the release notes thinks the app is running under. ' +
+    '<div style="color:red">But you can also update the package.json and run yarn:build, to actually change the version the app believes it is running under.</div>'
+
+  const injectPretext = [...notes, '[pretext] ' + pretext]
+  console.log(injectPretext)
+  return injectPretext.map(n => parseEntry(n)).filter(notEmpty)
 }
 
 export function getReleaseSummary(
@@ -70,14 +77,14 @@ export function getReleaseSummary(
   const bugfixes = entries.filter(e => e.kind === 'fixed')
   const other = entries.filter(e => e.kind === 'removed' || e.kind === 'other')
   const thankYous = entries.filter(e => e.message.includes(' Thanks @'))
+  const pretext = entries.filter(e => e.kind === 'pretext')
 
   return {
     latestVersion: latestRelease.version,
     datePublished: formatDate(new Date(latestRelease.pub_date), {
       dateStyle: 'long',
     }),
-    // TODO: find pretext entry
-    pretext: undefined,
+    pretext,
     enhancements,
     bugfixes,
     other,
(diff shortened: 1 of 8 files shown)


app/src/lib/release-notes.ts
```diff
@@ -56,7 +56,14 @@ function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
 export function parseReleaseEntries(
   notes: ReadonlyArray<string>
 ): ReadonlyArray<ReleaseNote> {
-  return notes.map(n => parseEntry(n)).filter(notEmpty)
+  const pretext =
```
tidy-dev, line 59: The temporarily injected pretext note being markdown is failing unit test - but that is not a permanent change.

tidy-dev:
In case anyone finds this and is curious of decisions made:
1. Moving forward with using a change log entry of pretext to show case splashy new features in the release note dialog as shown in gif above.
  - Primary reasons: remotely updatable, adaptable to use as marketing material, and readily available to implement
  - Biggest drawbacks: Doesn't lend itself to tutorial style showcasing and removes ability for deep in app links. 
     - Decided that if those were something we wanted for a specific release we could do make a custom showcase in addition to the change log pretext.
2. More prominence: Decided to make a call to action that expires when a pretext exists to enable users to open the release notes after update. (Exact UI for this TBD)
3. We can use a GitHub generated image link. 
4. Marketing site: going to update the marketing site to also show pretext change logs in the release notes. 

Some things TBD:
- What the process for drafting one of these may be. Markdown ready strings change log may not be easy to focus on and edit for proofing - so having a markdown file or issue for drafting purposes may be ideal.
- Determine size requirements for pretext images + plus text so that they fit nicely in release note dialog and show a bit of the release notes below. (Similar to Hero image concept on blogs)
- Should we pre-process the markdown on release notes retrieval so that we reduce any display lag. (particularly thinking of larger gifs).

PR #14471, opened 2022-04-28
Do not crash the app when the trampoline server receives garbage

sergiou87:
## Description

Looking into some of our Sentry errors, we decided to downgrade this crash to just a logged error, since it'd be possible for some random app to send garbage data to random open ports, and that shouldn't make Desktop crash.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨  Makes sense. 🙂

PR #14481, opened 2022-04-29
Add support for syntax highlighting .mjs/.cjs (and similar) extensions

j-f1:
Closes [N/A]

## Description

This PR formats `.cjs`, `.mjs`, `.cts`, `.mts`, `.cjsx`, `.mjsx`, `.ctsx`, and `.mtsx` file extensions as JS/TS/JSX/TSX as appropriate. Note that I’m not sure that all tools support the JSX variants but I feel like it’s better to get ahead of adding them (especially since they don’t really have any other uses). But I think it would be reasonable to remove all the JSX extensions before merging this PR.

### Screenshots

Screenshot_2022-04-29 15 58 33

## Release notes

Notes: Add support for highlighting .mjs/.cjs/.mts/.cts files as JavaScript/TypeScript

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ app/node_modules/
 .awcache
 .idea/
 .vs/
+.vscode/*.log
 .eslintcache
 *.iml
 .envrc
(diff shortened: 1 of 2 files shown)


niik (approved):
Thank you so much @j-f1!

PR #14493, opened 2022-05-03
Add way to generate pretext notes from draft

tidy-dev:
## Description
This adds a way that if we create a pretext draft in the `pretext-draft.md`, as part of a release branch, we can add a `--pretext flag` to the draft release command to generate the note for us. 

In addition to making it easier to test, I added the ability to use the draft-release command for drafting in the test channel. (However, we add the actual tags during the deployment scripts so this is 1 of 2 pieces required to fully support it) 

## Release notes
Notes: no-notes

### DIFF
--- a/docs/process/releasing-updates.md
+++ b/docs/process/releasing-updates.md
@@ -71,17 +71,19 @@ Run the script below (which relies on the your personal access token being set),
 For `production` and `beta` releases, run:
 
 ```shellsession
-$ yarn draft-release (production|beta)
+$ yarn draft-release (production|beta|test)
 ```
 
 If you are creating a new beta release, the `yarn draft-release beta` command will help you find the new release entries for the changelog.
 
 If you are create a new `production` release, you should just combine and sort the previous `beta` changelog entries.
 
-(For `test` releases, follow the directions in the steps below to update `app/package.json`'s `version` to a higher version and add a changelog entry. The script does not support test yet.)
+(For `test` releases, it will do a naive attempt to update the `app/package.json`'s `version` and add a changelog entry. You may have to adjust the test version and add a changelog entry to the `changelog.json`.)
 
 The script will output a draft changelog, which covers everything that's been merged, and probably needs some love.
 The output will then explain the next steps:
+
+If you have pretext release note drafted in `app/static/common/pretext-draft.md`, you can add the `--pretext` flag to generate a pretext change log entry it. Example: `yarn draft-release test --pretext`
 
 ```shellsession
 Here's what you should do next:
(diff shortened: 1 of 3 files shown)


script/draft-release/version.ts
```diff
@@ -73,7 +73,33 @@ export function getNextVersionNumber(
...
+    case 'test':
+      console.warn(
+        `WARNING: Since we do not tag test versions, the version provided will always be the first of test series respective of latest prod or beta.
+         You may need to adjust the -testX according to how many test releases already exist.`
+      )
 
+      if (isTestTag(semanticVersion)) {
+        /*Since we do not tag test versions, it "should be" impossible to reach
+        this first if, but added just for the case of future proofing the ideal
+        path. */
```
sergiou87: Maybe we should tag them? In the end the code is public, there is a track record of them from Central. I don't think it's that bad, and it's always kind of annoying if someone else has been pushing test builds you have to take a look at the latest one and +1 whatever build number that has.

Another option is to maybe allow specifying a specific "test build number" as a parameter for test builds.

sergiou87 (commented):
LGTM but maybe we can make test builds easier to track (and easier to bump their versions too).

script/draft-release/version.ts
tidy-dev: I went ahead and investigated what else needs to be done to fully support the test channel. I have added the changes here, but for full support we need to create the tags which is done during deployment. I have added the necessary change to [here](https://github.com/github/central/pull/575).

docs/process/releasing-updates.md
```diff
@@ -71,18 +71,20 @@ Run the script below (which relies on the your personal access token being set),
...
-$ yarn draft-release (production|beta)
+$ yarn draft-release (production|beta|test)
 ```
 
 If you are creating a new beta release, the `yarn draft-release beta` command will help you find the new release entries for the changelog.
 
 If you are create a new `production` release, you should just combine and sort the previous `beta` changelog entries.
 
-(For `test` releases, follow the directions in the steps below to update `app/package.json`'s `version` to a higher version and add a changelog entry. The script does not support test yet.)
+(For `test` releases, it will do a naive attempt to update the `app/package.json`'s `version` and add a changelog entry. You may have to adjust the test version and add a changelog entry to the `changelog.json`.)
```
sergiou87: This is not true anymore, right? With your latest changes it should always work (or at least as much as it does for beta and production builds).

tidy-dev:
```suggestion
```
Remove this now

sergiou87 (approved):
✨

PR #14495, opened 2022-05-04
Upgrade Typescript to 4.6

niik:
## Description

Upgrading to [Typescript 4.6](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/) lets us target ES2022 (all ES2022 _language features_ are already [implemented](https://kangax.github.io/compat-table/es2016plus/#chrome98) in Chrome 98 and Node 16.13 which we're using) which in turn gives us access to the handy [`Array.prototype.at`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) function as well as [Object.hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn).

The `.at` method is typed as `T | undefined` whereas indexed access of an array is by default typed as `T`. While we don't have to use it everywhere it does add a nice extra guard to let us know we can't expect `arr[arr.length - 1]` to exist. Hopefully one day there will be enough compiler cleverness for us to be able to turn on [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) but that's way to big of a headache for us to deal with now.

I've had to make a change to our Dexie database subclasses since ES2022 has added support for class fields (https://inspiredwebdev.com/everything-new-in-es2022/) which caused our previous approach of declaring, but not initializing, properties to fail du to them being emitted now. See the docs for [TS 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier) for some more detail on this phenomenon. 

## Release notes

Notes: no-notes


package.json
```diff
@@ -9,13 +9,12 @@
     "check:eslint": "tsc -P eslint-rules/",
     "test:eslint": "jest eslint-rules/tests/*.test.js",
     "test:unit": "cross-env ELECTRON_RUN_AS_NODE=1 NODE_OPTIONS='--max_old_space_size=4096' ./node_modules/.bin/electron ./node_modules/jest/bin/jest --detectOpenHandles --silent --testLocationInResults --config ./app/jest.unit.config.js",
-    "test:unit:cov": "yarn test:unit --coverage",
+    "test:unit:cov": "yarn test:unit",
     "test:script": "jest --silent --config ./script/jest.config.js",
-    "test:script:cov": "yarn test:script --coverage",
+    "test:script:cov": "yarn test:script",
```
sergiou87: Can we also remove these `:cov` scripts entirely?

sergiou87 (commented):
Nice cleanup!! And great to see we can use `at` now :_D

Aside from testing it, I only have one question about removing the `:cov` scripts from our package.json.

package.json
niik: Absolutely, I thought I had already!

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -87,14 +87,14 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
    */
   private getDisplayRelease = () => {
     const { newReleases } = this.props
-    const latestRelease = newReleases[0]
 
-    if (newReleases.length === 1) {
+    const latestRelease = newReleases.at(0)
+    const oldestRelease = newReleases.at(-1)
+
+    if (!latestRelease || !oldestRelease || latestRelease === oldestRelease) {
```
tidy-dev: Minor nit - we usually try to avoid "truthy/falsey" comparisons. :)

app/src/ui/release-notes/release-notes-dialog.tsx
```diff
@@ -107,6 +107,11 @@ export class ReleaseNotes extends React.Component<IReleaseNotesProps, {}> {
 
   public render() {
     const release = this.getDisplayRelease()
+
+    if (!release) {
```
tidy-dev: Same as above.

tidy-dev (changes requested):
✨  That `.at` will be so handy!  

One minor nit (and mostly just cause it is a nit I have gotten on my prs 😸  so upkeeping convention)

tidy-dev (approved):

sergiou87 (approved):
😗👌 :shipit:

PR #14496, opened 2022-05-04
Notifications improved

sergiou87:
## Description

This PR makes improvements in the notifications support by leveraging the changes in `desktop-notifications` introduced in https://github.com/desktop/desktop-notifications/pull/13 like:
- Handling notifications from previous app sessions (right now clicking on those will do nothing at all). In order to achieve this, Desktop will attach the Alive event to the notification (via its `userInfo` parameter), which will be recoverable at the time the notification is clicked, even if it's in a different app session. About this, I'm not particularly proud of keeping the entire Alive event, but seemed to be the easier approach, and doing a custom structure that would end up storing most if not all the fields of the original Alive event… sounded overkill.
   - **Important:** this can only be tested with actual notifications (PR reviews or checks failed), not with `Help -> Show Notification`
- Checking whether or not the user gave Desktop permission to display notifications, and even requesting them for those permissions (macOS only), which might be handy from our Advanced settings dialog, as a hint/suggestion from the checkbox that enables/disables the notifications.
   - Only show the notification settings hint when the notifications are enabled within the app.
   - If the user never granted nor denied permission to display the notifications, and it's possible to actively request so to the user (macOS only), provide a quick link to do so. Otherwise, just show a link to the notification settings if available.
   - If the user explicitly denied permission to display the notifications (yet they're enabled within Desktop), warn the user they need to enable it in the system settings.
- ~~Not sure if possible, but it'd be nice also to handle notifications while the app is closed (which would effectively open the app **and then** handle the notification).~~

An important part of this PR consists on displaying the notifications from the main process (required for macOS…) and handling their events from the renderer process.

### Screenshots

https://user-images.githubusercontent.com/1083228/167612530-ee4af24e-d633-4421-aed3-22a002d33049.mov

https://user-images.githubusercontent.com/1083228/167612579-187712a9-ee92-475c-9846-087d2002b0b3.mov

## Release notes

Notes: TBD


app/src/lib/notifications/show-notification.ts
```diff
@@ -0,0 +1,39 @@
...
+  onClick: () => void
+}
+
+/**
+ * Shows a notification with a title, a body, and a function to handle when the
+ * user clicks on the notification.
+ */
+export async function showNotification(options: IShowNotificationOptions) {
+  if (!supportsNotifications()) {
+    const notification = new Notification(options.title, {
```
tidy-dev, line 22: I am probably forgetting something (or lack of sleep is making it not click..) but, this reads as "If we don't support notifications, make a new notification".  Is this a dual support of a kind that is notified by renderer vs main process?

app/src/ui/preferences/advanced.tsx
```diff
@@ -203,25 +218,81 @@ export class Advanced extends React.Component<
...
+    if (notificationSettingsURL === null) {
       return null
     }
 
-    const notificationSettingsURL = __DARWIN__
-      ? 'x-apple.systempreferences:com.apple.preference.notifications'
-      : 'ms-settings:notifications'
+    if (warnNotificationsDenied) {
+      return (
+        <p>
```
tidy-dev: Console is angry at this. :D 
image

tidy-dev (changes requested):
Everything works great!

Just be nice to address the nested `<p>` tag when showing the notifications permission denied warning. (doesn't actually cause a problem, just annoying)

Other thoughts. I think if I was a user and I had somehow turned off the system notifications and I came here to "turn them on".. That I might not see  this warning. It feels really subtle... wondering if it would be better to disable that checkbox in this scenario so that it is clear something is "wrong".. and maybe show the warning above it or above the description or maybe with some color inflection like orange warning?

image

(If you think that is valid, could easily be addressed in another PR)

app/src/lib/notifications/show-notification.ts
sergiou87: Yes, this would be HTML5 notification API. I'll write a comment to make it less confusing 😳  Sorry about that!

sergiou87:
Thanks for your feedback! 🙇‍♂️ 

I definitely agree that the warning might be hard to spot. However, I couldn't think of a solution I liked better 😕 

> That I might not see this warning. It feels really subtle...

I was counting on the "popping" effect + the ⚠️  emoji when the user enables them while permission has been denied to make it more visible in that specific scenario 🤔 

> wondering if it would be better to disable that checkbox in this scenario so that it is clear something is "wrong"

With this PR, we only show that warning when permission has been denied **and notifications are enabled**. We don't show the warning when notifications aren't enabled to avoid noise for a user who hasn't enabled them, and that feels like the right approach to me 🤔  Disabling the checkbox while notifications are enabled would be odd.

> maybe show the warning above it or above the description or maybe with some color inflection like orange warning?

Moving it above and displacing the checkbox description also feels odd to me 😕  Making the warning orange is something that I like better, but I reused the icon we had to warn misattributed commits in the `Git` settings section. This is what orange would look like in both settings panels and the commit message area.
image image
image

They don't look too bad I think? Thoughts?

app/src/ui/preferences/advanced.tsx
sergiou87: Fixed in e532ecf614c504e94de0c35c35fc6d22695cab2a

app/src/lib/notifications/show-notification.ts
sergiou87: Fixed in 5d62f75c6dc028cea5bc6ce7cf2aaa2dd99717f1

app/src/lib/notifications/show-notification.ts
```diff
@@ -16,6 +16,8 @@ interface IShowNotificationOptions {
  * user clicks on the notification.
  */
 export async function showNotification(options: IShowNotificationOptions) {
+  // `supportNotifications` checks if `desktop-notifications` is supported by
+  // the current platform. Otherwise, we'll rely on the HTML5 notification API.
```
tidy-dev, line 20: ❤️

tidy-dev:
Yeah. I like that and, what do you think about adding a line of white space between the description and the warning to make them feel more separate?

> I was counting on the "popping" effect

I didn't even realize that if I uncheck and check it, the warning goes away and comes back. Mainly cause I came and saw that it was checked so my brain was like "This should be working?" and I would not have thought to uncheck it and check it again. Thus, it totally makes sense from the perspective of someone who has it disabled and unchecked. Because you would get that popping effect.  

Happy to leave it with just the orange (and maybe the white space) to make it less subtle. But one other idea is to have the checkbox always reflect the apps behavior (I think that is the disconnect my brain is having).. thus if it is disable by system settings, also have it it unchecked on setting being opened. Thus, when the user attempts to check it, they get the popping effect. But still "behind the scenes" be remembering the last user selected Desktop setting for if the user updates the system settings.

tidy-dev (dismissed):
Approved in case you want to handle the cosmetic/ux interaction on another PR.

sergiou87:
Adding an extra blank line sounds good to me 👍 

> Happy to leave it with just the orange (and maybe the white space) to make it less subtle

I'd love to know what @niik thinks since that's a bigger change.

> But one other idea is to have the checkbox always reflect the apps behavior

Totally understand how that's confusing, but I think it's better to be "honest" about the current state of the app vs external factors. If you see that checkbox disabled, then you fix permissions in macOS and the checkbox magically gets checked could be 🤯  too 🤔

niik:
> I'd love to know what @niik thinks since that's a bigger change.

I'm :+1: on colorizing the warning icon

sergiou87:
Thank you all for your feedback! I've changed the icon color and added an extra blank line above of the warning:

image

tidy-dev (approved):
:sparkles: Thanks for those changes! :)

PR #13727, opened 2022-01-23
Removed unnecessary punctuation

Pinzauti:
Closes #13715

## Description

- Removed unnecessary period according to the [guidelines](https://primer.style/design/foundations/content#labels-and-buttons).

## Release notes

Notes: fixed punctuation typo.

PR #14511, opened 2022-05-06
Revert "Removed unnecessary punctuation"

Alfre1352:
Reverts desktop/desktop#13727


sergiou87:
I think the punctuation is fine and consistent with the rest of the options, but thank you 😄

Issue #4214, opened 2018-03-10
Warn user of unmerged commits before deleting branch


rock-hopper:
## Description

The user is not warned when deleting a branch which contains commits not yet merged into the master branch.

## Version

* GitHub Desktop: 1.1.0

* Operating system: Windows 10 / Windows 7

## Steps to Reproduce

1. create new branch
2. make one or more commits in the new branch
3. delete the branch

### Expected Behavior

A warning that not all commits have been merged into the master branch
### Actual Behavior

A warning message is displayed stating that the operation cannot be undone. However, the message doesn't mention that there are commits in this branch which have not been merged into master branch

a2937:
How far along was @iAmWillShepherd's work on the issue and to clarify were there any additional test cases not discussed in the original pull request?

tidy-dev:
> How far along was @iAmWillShepherd's work on the issue and to clarify were there any additional test cases not discussed in the original pull request?

This was never implemented. I believe on the basis that not everything in https://github.com/desktop/desktop/pull/4217#issuecomment-372131743 was addressed.

PR #14512, opened 2022-05-06
Handle Invalid Clone Path & Uncommitted Changes on Branch

WilliamStar007:
**Issue #13816 Clone Error**
**Issue #4214 Warn User of Unmerged Commit**

## Description
**Issue #13816:**
Set the default directory for cloning to be the root, preventing invalid absolute paths.  

**Issue #4214:** 
Warn users of unmerged commits before deleting a branch.
New feature to Merge and Delete.

## Release notes
Changed the default directory for cloning to the root directory to avoid invalid input directories.
Warn users of merged commits on a branch before deleting and allow "Merge and Delete".

### DIFF
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -151,7 +151,7 @@ export class CloneRepository extends React.Component<
     super(props)
 
     // Initialize initial path with a '/' to avoid invalid paths
-    const defaultDirectory = '/'
+    const defaultDirectory = __WIN32__? '\\' : '/'
 
     const initialBaseTabState: IBaseTabState = {
       error: null,
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -150,7 +150,8 @@ export class CloneRepository extends React.Component<
   public constructor(props: ICloneRepositoryProps) {
     super(props)
 
-    const defaultDirectory = null
+    // Initialize initial path with a '/' to avoid invalid paths
+    const defaultDirectory = '/'
 
     const initialBaseTabState: IBaseTabState = {
       error: null,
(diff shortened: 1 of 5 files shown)


app/src/ui/delete-branch/delete-branch-dialog.tsx
```diff
@@ -36,6 +40,43 @@ export class DeleteBranch extends React.Component<
...
 
   public render() {
+    const aheadBehind = this.props.aheadBehind
+    return aheadBehind !== null && aheadBehind.ahead > 0
+      ? this.renderDeleteBranchWithUnmergedCommits()
+      : this.renderDeleteBranch()
+  }
+
+  private renderDeleteBranchWithUnmergedCommits() {
+    const unmergedCommits = this.props.aheadBehind!.ahead
```
sergiou87, line 50: This makes the app crash if the branch hasn't been pushed and is only local.

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -150,7 +150,8 @@ export class CloneRepository extends React.Component<
   public constructor(props: ICloneRepositoryProps) {
     super(props)
 
-    const defaultDirectory = null
+    // Initialize initial path with a '/' to avoid invalid paths
+    const defaultDirectory = '/'
```
sergiou87, line 154: Unless I'm missing something, I don't see how this fixes the issue with paths 🤔

sergiou87 (changes requested):
Thank you for your contribution @WilliamStar007 ❤️ 

Unfortunately, the fix for #13816 doesn't seem to work. You can follow the same steps as in the video in that issue and it will behave exactly the same way.

Regarding the fix for #4214, aside from the crash for branches that haven't been pushed and only exist locally, I don't think is what #4214 describes: the approach in this PR checks if there are commits that haven't been pushed to the remote, but not whether or not they have been merged to the main branch.

If you want to try and solve this issues, do you mind tackling them in separate pull requests so they're easier to review and test?

PR #14054, opened 2022-03-02
Added Android Studio To External Editors

detherminal:
Added Android Studio To "Open External Editor" Options

Closes #14044

## Description

I added Android Studio To "Open External Editor" Options

## Release notes

Notes: Added Android Studio To "Open External Editor" Options

PR #14523, opened 2022-05-09
Specify 64-bit exe path for Android Studio

### LINKED ISSUE #14519: I cannot open files into Android Studio from GitHub.
### The problem

I cannot open files into Android Studio from GitHub.

image

### Release version

3.0.0 (x64)

### Operating system

Windows 10

Open project in GitHub
Select Repository > Open in Android Studio
Get error message.

### Log files

_No response_

### Screenshots

_No response_

### Additional context

_No response_

sergiou87:
Closes #14519 

## Description
This PR makes sure Desktop looks for the 64bit version of the Android Studio executable and runs that one. Otherwise, the registry keys point to the 32-bit which won't be able to launch the 64-bit versions of the JVM DLLs.

These changes make Desktop pick the installation folder from the `UninstallString` registry key (which points to something like `C:\Program Files\Android\Android Studio\uninstall.exe`, and then append `..\bin\studio64.exe`.

Not sure when this broke because I tested #14054 and worked as expected. Unless... this wasn't a problem on Windows for ARM, which is where I tested it 🤔 

## Release notes

Notes: [Fixed] Fix opening files with Android Studio

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -22,7 +22,7 @@ type WindowsExternalEditorPathInfo =
   | {
       /**
        * Registry key with the install location of the app. If not provided,
-       * 'InstallLocation' will be used.
+       * 'InstallLocation' or 'UninstallString' will be assumed.
        **/
       readonly installLocationRegistryKey?:
         | 'InstallLocation'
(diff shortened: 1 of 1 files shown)


tsvetilian-ty:
@sergiou87 I tested it (on Windows 11 x86-64) when #14054 was open, and I can confirm that the issue was not present!

tidy-dev (approved):
:sparkles:  Looks good and tested well on my Windows 10 vm.

sergiou87:
@tsvetilian-ty thanks for confirming I'm not (too) crazy 😂

PR #14601, opened 2022-05-11
Fix PR review notification dialog dashed lines

sergiou87:
## Description

Recently I noticed in a screenshot @niik shared that the dashed timeline lines in PR review notifications weren't actually dashed. It's a simple fix, but the problem seems to happen on high-density screens (happens on my MacBook but not on my external display).

### Screenshots

Before:
image

After:
image

## Release notes

Notes: no-notes

### DIFF
--- a/app/styles/ui/_pull-request-review.scss
+++ b/app/styles/ui/_pull-request-review.scss
@@ -52,7 +52,6 @@
         line {
           stroke: var(--pr-timeline-line-color);
           stroke-width: 1px;
-          stroke-linecap: square;
         }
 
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
:sparkles: Looks good.

... This screenshot is good reminder that I need to double check what I type before hitting enter for my tendency towards malapropism less it be used as examples. :D

sergiou87:
TIL `malapropism`, and I just learned that we have `malapropismo` in Spanish too 👀 😂

PR #14620, opened 2022-05-16
Pretext/Showcase Banner

tidy-dev:
## Description
This PR modifies the update banner such that, if there is a pretext in a release within the last 15 days, it will display a showcase banner where the user can open the release notes to see the pretext. 

Like the update banner, this will not be present if any other banner is open. This will also not be present if there is a pending update (as it will show the regular update banner). Additionally, if a user clicks the "what's new" to view the release notes while the update is pending, that will prevent user from seeing this showcase banner as it would be redundant. 

Other notes: as can be seen in this screen recording.. need to do some css work when displaying only one column of release notes under a showcase, but that is not related to this and I will get it in a follow up PR.

### Screenshots
https://user-images.githubusercontent.com/75402236/168608260-92cd053e-7e80-47f3-91fe-cab1b3b32735.mov

## Release notes
Notes: [Improved] Display a banner when we have a pretext release note to highlight the new feature.

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -180,7 +180,7 @@ export interface IAppState {
   readonly isUpdateAvailableBannerVisible: boolean
 
   /** Whether there is an update to showcase */
-  readonly isUpdateShowCaseVisible: boolean
+  readonly isUpdateShowcaseVisible: boolean
 
   /** Whether we should ask the user to move the app to /Applications */
   readonly askToMoveToApplicationsFolderSetting: boolean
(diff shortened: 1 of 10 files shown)


tidy-dev:
Just added a one line css change to address the funny looking alignment when there is only one column of release notes.
Below show all the different possible combos.

Pretext (one and two columns):
image
image

No pretext (one and two columns)
(before these release notes change, the one column would center itself. I am not sure that looks any better than keeping it left aligned as is shown here. 
image


image

sergiou87 (approved):
Code looks good! I tested a bunch of things from 3.0.1-test5 and all looked good.

:shipit:

PR #14622, opened 2022-05-16
Move markdown parsing to exportable/reusable method

tidy-dev:
## Description
This is an initial step in process to make markdown be able to be parsed ahead of render time to reduce user wait time of markdown parsing. It just moves methods around to make them separated/reusable.

... next thing will be to handle markdown as a promise or maybe an observable type thing... 

## Release notes
Notes: none


sergiou87 (approved):
:shipit:

PR #14625, opened 2022-05-17
Sandboxed Markdown Parser accepts an emitter

tidy-dev:
## Description
This refactors the markdown parser into an asynchronous process.  (But UI/UX wise doesn't noticeably change the experience)

Reason to do this:
1. We can decided to go ahead and parse markdown content ahead of time. For example, if a release note has "pretext" note. We have the ability to go ahead and initiate the parsing such that when the users actually sees it there is not render time lost to markdown parsing. This would likely prevent the need to see a load image.
2. The custom markdown filters are now performed individually in an async manner. This means the user will be able to see the vast majority of the markdown as parsed by markedjs and then get debounced updates to it as the custom markdown filters are applied (the slower part). This is nice to reduce load time for users to see actionable info when markdown is not pre-parsed as suggested in bullet 1. Tho it may create a visible updating (maybe this can be softened with animation?)... Also give us the ability to add to custom markdown filters that may take longer (like requiring network calls). 

Next PR to actually implement pre parsing on release notes and pr previews.

Can see usage in #14630. 

## Release notes
Notes: no-notes


app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -105,8 +118,27 @@ export class SandboxedMarkdown extends React.PureComponent<
...
+    }
+    const { emoji, repository, markdownContext } = this.props
+    this.markdownEmitter =
+      typeof this.props.markdown !== 'string'
+        ? this.props.markdown
+        : parseMarkdown(this.props.markdown, {
+            emoji,
+            repository,
+            markdownContext,
+          })
```
tidy-dev, line 133: Note.. this was supposed to be part of my last PR on this (but didn't cherry pick that commit).. and since it wasn't the markdown parsing is regressed to not having custom markdown filters on development. If we were to decide not to merge this PR, I would at least need to open a PR to provide this options **before next beta.**

tidy-dev: Opened [this PR](https://github.com/desktop/desktop/pull/14626) in case this is not merged before next beta.

tidy-dev: (Other PR merged so ignore this)

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -105,8 +118,27 @@ export class SandboxedMarkdown extends React.PureComponent<
     this.frameContainingDivRef = frameContainingDivRef
   }
 
+  private initializeMarkdownEmitter = async () => {
```
sergiou87: Does this need to be `async`? I don't see any `await` 🤔 
```suggestion
  private initializeMarkdownEmitter = () => {
```

app/src/ui/lib/sandboxed-markdown.tsx
```diff
@@ -105,8 +118,27 @@ export class SandboxedMarkdown extends React.PureComponent<
...
+          })
+
+    this.markdownEmitter.onMarkdownUpdated((markdown: string) => {
+      this.onMarkdownUpdated(markdown)
+    })
+  }
+
   public async componentDidMount() {
-    this.mountIframeContents()
+    await this.initializeMarkdownEmitter()
```
sergiou87: If, as mentioned in my previous comment, `initializeMarkdownEmitter` is not `async`, this doesn't need `await`, right?

app/src/lib/markdown-filters/node-filter.ts
```diff
@@ -104,15 +106,21 @@ export const buildCustomMarkDownNodeFilterPipe = memoizeOne(
...
+
+  const mdDoc = new DOMParser().parseFromString(
+    markdownEmitter.latestMarkdown,
+    'text/html'
+  )
 
   for (const nodeFilter of nodeFilters) {
     await applyNodeFilter(nodeFilter, mdDoc)
+    markdownEmitter.emit(mdDoc.documentElement.innerHTML)
   }
```
sergiou87: Just a suggestion: since this can be kind of expensive depending on the markdown… maybe we could ad a way of breaking out of this loop if the emitter is disposed?

app/src/ui/lib/sandboxed-markdown.tsx
tidy-dev: Thanks good catch -> I had moved stuff around and forgot to remove that.

app/src/lib/markdown-filters/node-filter.ts
tidy-dev: Great suggestion. I figured also to put one at the top since this is hit after the markdown emitter is created during marked js parsing. That ensures that the markdown emitter at-least should be viable when we hit the first filter application.

app/src/lib/markdown-filters/node-filter.ts
```diff
@@ -104,15 +106,24 @@ export const buildCustomMarkDownNodeFilterPipe = memoizeOne(
...
+  const mdDoc = new DOMParser().parseFromString(
+    markdownEmitter.latestMarkdown,
+    'text/html'
+  )
 
   for (const nodeFilter of nodeFilters) {
     await applyNodeFilter(nodeFilter, mdDoc)
+    if (markdownEmitter.isDisposed) {
+      break
+    }
```
sergiou87, line 124: @tidy-dev  I just tested this by adding a `await sleep(5000)` right above the `if` and setting a breakpoint in the `break` and it didn't stop there. From the console I checked, and it seems the emitter has a `disposed` property, but not `isDisposed`.

From the looks of it, the API changed so I bumped `@types/event-kit` and updated all the required references (including this `disposed`) in eba06ef67a0779a08f087dc0e265bedede92604a and now all works as expected 🎉 

Feel free to revert/remove/change anything from that commit if needed!

tidy-dev: Looks good to me.

sergiou87 (approved):
Code looks good! As mentioned in https://github.com/desktop/desktop/pull/14625#discussion_r882548421 I found something not working and fixed it. I'll approve this, and feel free to merge it if you're comfortable with my changes 😄

PR #14626, opened 2022-05-17
Fix regression from moving custom markdown filters

tidy-dev:
## Description
In https://github.com/desktop/desktop/pull/14622, I moved the parsing out into is own file and in doing so created custom markdown filter options.. which I neglected to add to that PR.. 🤦 Thus development is regressed not to apply the custom markdown filters at the moment. If https://github.com/desktop/desktop/pull/14625 is not merged, then this needs to be merged before a new release is shipped.

## Release notes
Notes: no-notes


sergiou87 (approved):
🚀

PR #14630, opened 2022-05-18
Pre-parse PR Quickview markdown

tidy-dev:
This is based on #14625

## Description

This makes use of the markdown emitter functionality in #14625 and pre-parses the PR bodies displayed in the quick view markdown. The quick view is still feature-flagged to development.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/models/pull-request.ts
+++ b/app/src/models/pull-request.ts
@@ -1,3 +1,8 @@
+import { enablePullRequestQuickView } from '../lib/feature-flag'
+import {
+  MarkdownEmitter,
+  parseMarkdown,
+} from '../lib/markdown-filters/markdown-filter'
 import { GitHubRepository } from './github-repository'
 
 /** Returns the commit ref for a given pull request number. */
@@ -20,6 +25,8 @@ export class PullRequestRef {
 }
 
 export class PullRequest {
+  public parsedMarkdownBody?: MarkdownEmitter
+
   /**
    * @param created The date on which the PR was created.
    * @param status The status of the PR. This will be `null` if we haven't looked up its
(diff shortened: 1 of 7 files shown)


app/src/models/pull-request.ts
```diff
@@ -20,6 +25,8 @@ export class PullRequestRef {
 }
 
 export class PullRequest {
+  public parsedMarkdownBody?: MarkdownEmitter
```
sergiou87, line 28: Not a big fan of having this in the "raw" Pull Request model, to be honest 🤔 Once it's initialized, it'll stay in memory until it's removed from the PullRequestStore, right?

Maybe it makes more sense to put it in the PullRequestListItem? 🤔

sergiou87 (commented):
It looks good overall but I'm not convinced about having the parsed markdown in the raw PR model 😕

app/src/models/pull-request.ts
tidy-dev: Discussed in slack a bit with @sergiou87, would appreciates @niik input from exp managing data in Desktop. :)

Reason I went with this over prListItem state was to reduce multiple parsing of same PR.

Good points against current approach:
- A second copy of the PR body maybe kept around longer than needed - feels inefficient. However, parsing is only done once (or as pr is updated)
- Having this added to the pr model close to the UI layer feels weird - weird to have something that sometimes exists.

Other approaches:
- Putting it on the prListItem state (bubble up on mouse over):
   - solves it only being around when the user might use it and not a second copy on pr model
   - concern: markdown will be re-parsed during scrolling (may cause scrolling lag on larger repos) and when the user closes and reopens the branch/pr dropdown. 
- Add this to the PR model at PR model creation:
    - Solves the weirdness of adding to a pr data model close to the UI layer. 
    - Concern: Keeping second copy around all the time.
    - Positive: Parsing only done once (or as pr is updated)
    - Other notes: This means passing necessary markdown parser options such as the emoji map to the pr store... may have to do some refactoring of when the emoji map is generated vs the pr store is... or make a pr store deferred action that is called during the app.tsx deferred launch actions method.
- Create PR Markdown Map: (likely add to it same time as now - list item render)
   - Positive: Keeps markdown copy separate from pr model - no adding to pr model weirdness
   - Positive: Parsing only done once (or as pr is updated)
   - Concern: Keeping second copy around all the time.
   - Other notes: We would need a clear mechanism if a PR is updated. 


Also open to other non-discussed approaches :D

tidy-dev:
Seeing as we are not looking to ship the PR previewer at this time, I am going to close this until we prioritize this work again.

PR #14633, opened 2022-05-18
Escape branch name

### LINKED ISSUE #14631: branch name with '#' doesn't open when "View branch of Github"
### The problem

branch name with '#' doesn't open when "View branch of Github"

My branch name: `ab#4176-use-badge-style-for-job-status-job-list`
> it links me to: 
> github.com/.../ab#4176-use-badge-style-for-job-status-job-list

> the right link:
> github.com/.../ab%234176-use-badge-style-for-job-status-job-list

### Release version

3.0.0

### Operating system

Windows 10

1. Name branch with '#'
2. From GitHub desktop from menu click Branch > View Branch in Github

### Log files

_No response_

### Screenshots

_No response_

### Additional context

_No response_

tsvetilian-ty:
Closes #14631

## Description
As an author of #14231 I had to fix this one 😄 

- Fix for unencoded branch name that causes issues with "View Branch on GitHub" and "Compare on GitHub".

### Screenshots
test-branch-name-fix

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react'
 import * as crypto from 'crypto'
-import { escape } from 'querystring'
 import { TransitionGroup, CSSTransition } from 'react-transition-group'
 import {
   IAppState,
@@ -691,7 +690,9 @@ export class App extends React.Component<IAppProps, IAppState> {
       return
     }
 
-    const urlEncodedBranchName = escape(branchTip.branch.upstreamWithoutRemote)
+    const urlEncodedBranchName = encodeURIComponent(
+      branchTip.branch.upstreamWithoutRemote
+    )
 
     const url = `${htmlURL}/${view}/${urlEncodedBranchName}`
     this.props.dispatcher.openInBrowser(url)
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import * as crypto from 'crypto'
+import { escape } from 'querystring'
 import { TransitionGroup, CSSTransition } from 'react-transition-group'
 import {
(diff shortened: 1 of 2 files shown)


tidy-dev (dismissed):
:sparkles: Thanks for that quick fix @tsvetilian-ty!

app/src/ui/app.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import * as crypto from 'crypto'
+import { escape } from 'querystring'
```
niik: The `querystring` module is marked as legacy in the Node docs. Could you please use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) instead? It won't have any practical effect but that's what we're using elsewhere and it's well documented.

niik (changes requested):
Echoing @tidy-dev's thanks here @tsvetilian-ty. I just had one small request that I'm hoping we could get in before merging

app/src/ui/app.tsx
tidy-dev: If you update that, could you also remove it's its usage in  `_openCreatePullRequestInBrowser` in the `app-store.tsx`? 

That was the reason I didn't realize it was different then rest of app, cause I saw it used there and assumed it was what we used.. even tho in the back of my mind I was thinking there was something else. Thanks @niik!

tsvetilian-ty: Nice catch @niik! I changed both instances of `escape` to `encodeURIComponent`. 

In the original PR, I used the encoding in `_openCreatePullRequestInBrowser` as a reference. That's why I picked `escape` in the first place.

niik (approved):

niik:
Thanks @tsvetilian-ty!

PR #14646, opened 2022-05-20
Correctly handle creating new branches when the default branch points to the upstream

j-f1:
Closes #N/A

## Description

I’ve been using a workflow where I delete the `main` branch of my forked repo and update it to point to `upstream/main`. However, GitHub desktop will no longer suggest creating a branch off of `upstream/main` in this setup.

This PR now looks for a local branch with an upstream of e.g. `upstream/main`, and will create the new branch off of that branch (if requested by the user).

One limitation of this approach is that the UI currently shows e.g. `development` instead of `upstream/development` in these cases, which can be confusing. I tried to remedy this but couldn’t figure out the best way to plumb through the information.

I also included a commit (9e4b42c4b365aa7e84a76df692d84cbc25c54243) where I tried to fix a UI bug where creating this type of local default branch would bypass the warning about lacking write access. However, I discovered that the GitHub API does not return permission info for the parent of a fork, so these changes wouldn’t work without further efforts to fetch that missing data (for example, by changing the API call that fetches the repository metadata to use the GraphQL API, which makes it easy to get the permission for the parent at the same time as the fork.

### Screenshots

I don’t think there’s anything to show, but here’s a step-by-step for setting up this usecase locally:

1. Clone a forked repository
2. Check out a non-default branch
3. Delete the default branch
4. Check out <code>upstream/*default branch name*</code> to recreate the default branch locally, except that now it points to the `upstream` remote

## Release notes

Notes: Improve support for forked repositories where the default branch points to the parent

(*I don’t think this wording is clear enough but I can’t think of anything better right now*)

### DIFF
--- a/app/src/lib/branch.ts
+++ b/app/src/lib/branch.ts
@@ -33,10 +33,12 @@ export function findDefaultUpstreamBranch(
     return null
   }
 
+  const upstreamName = `${UpstreamRemoteName}/${githubRepository.defaultBranch}`
+
   const foundBranch = branches.find(
     b =>
-      b.type === BranchType.Remote &&
-      b.name === `${UpstreamRemoteName}/${githubRepository.defaultBranch}`
+      (b.type === BranchType.Remote && b.name === upstreamName) ||
+      b.upstream === upstreamName
   )
 
   return foundBranch !== undefined ? foundBranch : null
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1042,7 +1042,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
       // test for this scenario and show a message specifically
       // about write access before showing a branch protection
       // warning.
-      if (!hasWritePermission(gitHubRepo)) {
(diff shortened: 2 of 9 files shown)


kayal28 (commented):
💙

niik:
Hey @j-f1, super sorry for us letting this sit for so long before coming back to it. Unfortunately thanks to us not checking in here a lot of things have changed in this area and I'm not sure that the original case still exists.

It's probably just because it's towards the end of my day and I haven't had lunch but I'm having a bit of a rough time understanding the problem to begin with. We have a big PR hopefully landing soon which will update our logic such that upstream (remote) default branches can be considered default branches as well (depending on your workflow preference). I wonder if that resolves this issue or if some change that went in before it did.

Either way given how much has changed since this was opened I'm inclined to close this and kindly ask you to check if this is still an issue (and something you're interested in spending time on fixing) and thereafter either opening an issue so we can all get on the same page on what the problem is or opening a PR with an updated fix that we can discuss.

Once again, my apologies!

j-f1:
I’m excited to see those changes. I think they will fix the issue I was having, so this can stay closed.

PR #14662, opened 2022-05-23
Added support for Visual Studio Community, Professional, Enterprise 2017 (v15.2^), 2019, 2022

### LINKED ISSUE #12946: Visual Studio Community Integration Needed
Feature needed to be implemented
Integration provides no feature for visual studio community as an editor.

Windows 10

image

Nico-Connor-Personal:
Closes [#12946](https://github.com/desktop/desktop/issues/12946)

## Description
Implementing Visual Studio support required an unconventional approach since Visual Studio doesn't use a unique registry key like most other IDEs.

Thankfully as of Visual Studio 2017 v15.2, Microsoft includes a CLI tool called [vswhere](https://github.com/microsoft/vswhere) with the visual studio installer to help locate the path to their IDE.  The installer has a unique registry key that we can use to locate its install path which will then allow us to query VSWhere and find the install paths for various visual studio installs.

### Screenshots

VS Github

image

## Release notes

Added IDE support for:
- Visual Studio Community 2022
- Visual Studio Professional 2022
- Visual Studio Enterprise 2022
- Visual Studio Community 2019
- Visual Studio Professional 2019
- Visual Studio Enterprise 2019
- Visual Studio Community 2017 (only version 15.2 and up)
- Visual Studio Professional 2017 (only version 15.2 and up)
- Visual Studio Enterprise 2017 (only version 15.2 and up)

Notes:

I don't have access to a 32 bit computer and therefore couldn't test my change on a 32 bit computer.  The registry location for the Visual Studio Installer might be different on a 32 bit computer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # [GitHub Desktop](https://desktop.github.com)
-test
+
 [GitHub Desktop](https://desktop.github.com/) is an open source [Electron](https://www.electronjs.org/)-based
 GitHub app. It is written in [TypeScript](http://www.typescriptlang.org) and
 uses [React](https://reactjs.org/).
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # [GitHub Desktop](https://desktop.github.com)
-
+test
 [GitHub Desktop](https://desktop.github.com/) is an open source [Electron](https://www.electronjs.org/)-based
 GitHub app. It is written in [TypeScript](http://www.typescriptlang.org) and
 uses [React](https://reactjs.org/).
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -142,7 +142,6 @@ const executableShimPathsForJetBrainsIDE = (
   ]
 }
 
-
 const visualStudioInstaller : WindowsVisualStudioInstaller = {
   name: 'Microsoft Visual Studio Installer',
   registryKeys: [LocalMachineUninstallKey('{6F320B93-EE3C-4826-85E0-ADF79F8D4C61}')],
@@ -610,8 +609,7 @@ export async function getAvailableEditors(): Promise<
         })
       }
     }
-  }
-  
+  } 
 
   return results
 }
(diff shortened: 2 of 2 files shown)


AraHaan:
32 bit computers are about 3% or less of modern developers nowadays so it’s a minor issue on not being able to test 32 bit.

Nico-Connor-Personal:
Turns out, according to the VSWhere documentation:
> vswhere.exe is installed in %ProgramFiles(x86)%\Microsoft Visual Studio\Installer (use %ProgramFiles% in a 32-bit program prior to Windows 10). This is a fixed location that will be maintained.
 
This means we don't need to do a registry search to get the visual studio installer path.  We can safely assume the installer will always be located at `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer`.   

`vswhere` provides us the ability to get the unique visual studio registry key.  As such, we can use use a more conventional approach to adding Visual Studio support with github desktop.

I started working on a second implementation here: https://github.com/desktop/desktop/pull/14664 , let me know if you would I prefer to focus on the new PR instead of this one.

The only downside about the new PR is that I have to make use of `execFileSync` to call `vswhere`.  The problem with `execFileSync(vswhere)` is that when we call vswhere, it creates a new process and won't return until the new process exits.  If the new process hangs, `execFileSync` won't return and will make github desktop hang.  If we decide to go with the new PR, I'll need to figure out how to more safely use `execFileSync` so it doesn't hang github desktop.

Nico-Connor-Personal:
I think the second implementation will be preferred: https://github.com/desktop/desktop/pull/14664

AraHaan:
> The only downside about the new PR is that I have to make use of `execFileSync` to call `vswhere`. The problem with `execFileSync(vswhere)` is that when we call vswhere, it creates a new process and won't return until the new process exits. If the new process hangs, `execFileSync` won't return and will make github desktop hang. If we decide to go with the new PR, I'll need to figure out how to more safely use `execFileSync` so it doesn't hang github desktop.

In the .NET World we would offload that into an task then use ``Task.Run`` for an fire-and-forget call that will not hang.

If all else fails, one can make an .NET 6 wrapper program (would require the .NET 6 runtime to be installed) to execute the program in a much safer way (also System.Diagnostics.Process is a good friend there).

Also a good thing about .NET is that it would also require the .NET SDK be installed so one can use ``dotnet run <dll path of the program to run>`` to run it however it should not be much of an issue I think (.NET 6 also is in the registry too I believe, however can be probed for in ``Program Files(x86)``, ``Program  Files``, and w/e Program Files is for arm64 as well).

PR #14664, opened 2022-05-23
Added support for Visual Studio Community, Professional, Enterprise - 2017 (v15.2^), 2019, 2022 on Windows 64-bit

### LINKED ISSUE #12946: Visual Studio Community Integration Needed

Nico-Connor-Personal:
Closes [#12946](https://github.com/desktop/desktop/issues/12946)

## Description
Implementing Visual Studio support required an unconventional approach since Visual Studio doesn't use a unique registry key like most other IDEs.

Thankfully as of Visual Studio 2017 v15.2, Microsoft includes a CLI tool called [vswhere](https://github.com/microsoft/vswhere) with the visual studio installer to help locate the unique registry of their IDE.  

### Screenshots

VS Github

image

## Release notes

Added IDE support for Windows 64 bit installations of:
- Visual Studio Community 2022
- Visual Studio Professional 2022
- Visual Studio Enterprise 2022
- Visual Studio Community 2019
- Visual Studio Professional 2019
- Visual Studio Enterprise 2019
- Visual Studio Community 2017 (only version 15.2 and up)
- Visual Studio Professional 2017 (only version 15.2 and up)
- Visual Studio Enterprise 2017 (only version 15.2 and up)

Notes:

I don't have access to a 32 bit computer and therefore couldn't explore what the root path of Visual Studio is on 32 bit computers.

It would be nice to have someone test this code with any version of visual studio 2017 15.2 and up and have someone test this with any supported version of visual studio enterprise.

A different implementation approach that I tried:
- https://github.com/desktop/desktop/pull/14662 <- was my first go at trying to implement support but I had to add too many customizations for visual studio.

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -86,46 +86,52 @@ const LocalMachineUninstallKey = (subKey: string) =>
 const Wow64LocalMachineUninstallKey = (subKey: string) =>
   registryKey(HKEY.HKEY_LOCAL_MACHINE, wow64UninstallSubKey, subKey)
 
-  const registryKeysForVisualStudioIDE = (
+
+/* 
+* This function generates registry keys for a given Visual Studio product
+* Currently, we only support 64 bit versions of Visual Studio
+*
+* To find the product ID, please consult the following link:
+* https://docs.microsoft.com/en-us/visualstudio/install/workload-and-component-ids?view=vs-2022 
+* (change the year at the end of the url to the year you want)
+* 
+* To find the version, please consult the following link:
+* https://docs.microsoft.com/en-us/visualstudio/install/visual-studio-build-numbers-and-release-dates?view=vs-2022
+* (change the year at the end of the url to the year you want)
+*/
+const registryKeysForVisualStudioIDE = (
   version: number,
   productId: string
 ): ReadonlyArray<RegistryKey> => {
 
   const result = new Array<RegistryKey>()
+  /*
+  * vswhere is a tool that is part of the Visual Studio Installer. It is used
+  * to find the path to the Visual Studio installation. 
+  * vswhere is included with the installer as of Visual Studio 2017 version 15.2 and later
+  * https://github.com/microsoft/vswhere
+  * 
+  * According to the documentation, vswhere is installed in the following fixed location location:
+  * %ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe for 64 bit
+  * %ProgramFiles%\Microsoft Visual Studio\Installer\vswhere.exe for 32 bit
+  * -https://github.com/Microsoft/vswhere/wiki/Installing
+  */
   let vswherePath = undefined;
   if(process.env['ProgramFiles(x86)']){
-    const path = Path.join(process.env['ProgramFiles(x86)'], 'Microsoft Visual Studio', 'Installer', 'vswhere.exe');
-    log.debug('Checking for vswhere.exe at ' + path);
-    if(existsSync(path)){
-      vswherePath = path;
-    }
+    vswherePath = Path.join(process.env['ProgramFiles(x86)'], 'Microsoft Visual Studio', 'Installer', 'vswhere.exe');
   } 
-  
-  if(process.env['ProgramFiles']){
-    const path = Path.join(process.env['ProgramFiles'], 'Microsoft Visual Studio', 'Installer', 'vswhere.exe');
-    log.debug('Checking for vswhere.exe at ' + path);
-    if(existsSync(path)){
-      vswherePath = path;
-    }
-  }
-  log.debug(`program file 86 path: ${process.env['ProgramFiles(x86)']}`);
-  log.debug(`program file 86 path: ${process.env['ProgramFiles']}`);
-  if(vswherePath !== undefined){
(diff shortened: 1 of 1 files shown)


app/src/lib/editors/win32.ts
```diff
@@ -82,6 +86,54 @@ const LocalMachineUninstallKey = (subKey: string) =>
...
+  * %ProgramFiles%\Microsoft Visual Studio\Installer\vswhere.exe for 32 bit
+  * -https://github.com/Microsoft/vswhere/wiki/Installing
+  */
+  let vswherePath = undefined;
+  if(process.env['ProgramFiles(x86)']){
+    vswherePath = Path.join(process.env['ProgramFiles(x86)'], 'Microsoft Visual Studio', 'Installer', 'vswhere.exe');
+  } 
+  if(vswherePath !== undefined && existsSync(vswherePath)){
+    try {
+      const test = execFileSync(vswherePath, ['-version', '['+String(version)+','+String(version+1)+')', '-products', productId, '-property', 'instanceId'],{timeout: 200, killSignal: 'SIGKILL'});
```
Nico-Connor, line 125: Is 200 milliseconds too long for the timeout?

tsvetilian-ty:
@Nico-Connor nice to see someone tackling the unstable keys of Visual Studio 🥳 I tested the branch, and it works like a charm 🎉 

I have a few suggestions on how to improve the PR:
- Make sure you run the `lint` script to check if there are issues with prettier
- You can update the list with supported editors in windows https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md

app/src/lib/editors/win32.ts
```diff
@@ -82,6 +86,54 @@ const LocalMachineUninstallKey = (subKey: string) =>
...
+  * According to the documentation, vswhere is installed in the following fixed location location:
+  * %ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe for 64 bit
+  * %ProgramFiles%\Microsoft Visual Studio\Installer\vswhere.exe for 32 bit
+  * -https://github.com/Microsoft/vswhere/wiki/Installing
+  */
+  let vswherePath = undefined;
+  if(process.env['ProgramFiles(x86)']){
+    vswherePath = Path.join(process.env['ProgramFiles(x86)'], 'Microsoft Visual Studio', 'Installer', 'vswhere.exe');
+  } 
+  if(vswherePath !== undefined && existsSync(vswherePath)){
```
tsvetilian-ty, line 123: You can use `pathExists` instead of `existsSync`.

app/src/lib/editors/win32.ts
```diff
@@ -82,6 +86,54 @@ const LocalMachineUninstallKey = (subKey: string) =>
...
+    try {
+      const vswhereStdOut = execFileSync(vswherePath, ['-version', '['+String(version)+','+String(version+1)+')', '-products', productId, '-property', 'instanceId'],{timeout: 2000, killSignal: 'SIGKILL'});
```
tsvetilian-ty, line 125: - Instead of `execFileSync` you can utilize `execFile` from `../exec-file` (it returns a Promise). It will make it consistent with other usages across GitHub Desktop
- Perhaps you can use template literal for `['+String(version)+','+String(version+1)+')`
- I'm not sure that's a problem, but the format of `['+String(version)+','+String(version+1)+')` look a bit odd... it opens with square brackets, and it ends with closing parenthesis?

sergiou87:
Hey @Nico-Connor 👋 

Thank you very much for your contribution ❤️  However, there is a major issue with it, related to how Desktop creates that list of editors. Right now, that list of editors is built when the app starts, which is kind of ok as long as that happens very quick. However, your PR requires running external processes to determine the location/existence of Visual Studio installations, which will definitely slow down the app start.

We've been discussing about it and we think it's time to refactor all this code so that Desktop does all this work asynchronously and on-demand when the app needs it (i.e. when the user wants to open the repo on their selected editor, or when the user opens `Settings` -> `Integrations` to select a new one). I'll file a new issue with that. And we could also take that chance to make it a bit more flexible, instead of depending exclusively on the registry. For example, we could use `vswhere` directly to get the location of the VS executable, instead of using it to build the registry keys…

Because of all of this, I'm afraid we cannot merge your PR without having that refactor first, sorry 😞

Nico-Connor-Personal:
That is very sensible reasoning!  I look forward to the completion of the refactoring.  Would you be able to link the new issue when it's up?  I wouldn't mind following the progress of the new issue 😉.

sergiou87:
Here it is: #14666 

Thank you for being so understanding 🙇‍♂️ ❤️

app/src/lib/editors/win32.ts
AraHaan: Also on x86 I believe ``Program Files (x86)`` does not exist as it is named ``Program Files`` if I remember from my 32 bit windows days.

app/src/lib/editors/win32.ts
AraHaan: https://github.com/microsoft/vswhere/blob/develop/src/vswhere.lib/VersionRange.cpp#L105-L108 take a look at this to know why the ``)`` is at the end.

app/src/lib/editors/win32.ts
sergiou87: @AraHaan that shouldn't be a problem since we only offer 64bit builds, right?

PR #14667, opened 2022-05-23
Add copy paths when multiple files are selected

### LINKED ISSUE #14638: Allow "Copy Relative File Path" menu option to copy multiple file paths
### The feature request

Currently, even if multiple files are selected in the list of changes, right clicking and selecting "Copy Relative Filepath" will only copy the filepath of the item which is clicked, not all the items which are selected.

This is a problem, because one use case for the copy feature is to paste these paths into a git command in the terminal. When paths are long and there are many of them, the user may want to do something like `git some-command -- file1 file2 file3` but will have to copy each path separately.

### Proposed solution

Clicking this menu command should copy a space-delimited list of all selected filenames. The space-delimitation will allow users to use these lists directly in git commands in the terminal.

### Additional context

Relates to #12589 

image

tsvetilian-ty:
Closes #14638

## Description
- Add "Copy Paths" and "Copy Relative Paths" when multiple files are selected

The paths are space-delimited as suggested in the feature request!

### Screenshots
multiple-path-copy

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -437,7 +437,7 @@ export class ChangesList extends React.Component<
         const fullPaths = files.map(file =>
           Path.join(this.props.repository.path, file.path)
         )
-        clipboard.writeText(fullPaths.join(' '))
+        clipboard.writeText(fullPaths.join('\r\n'))
       },
     }
   }
@@ -449,7 +449,7 @@ export class ChangesList extends React.Component<
       label: CopySelectedRelativePathsLabel,
       action: () => {
         const paths = files.map(file => Path.normalize(file.path))
-        clipboard.writeText(paths.join(' '))
+        clipboard.writeText(paths.join('\r\n'))
       },
     }
   }
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -27,6 +27,8 @@ import {
   RevealInFileManagerLabel,
   OpenWithDefaultProgramLabel,
   CopyRelativeFilePathLabel,
+  CopySelectedPathsLabel,
+  CopySelectedRelativePathsLabel,
 } from '../lib/context-menu'
 import { CommitMessage } from './commit-message'
 import { ChangedFile } from './changed-file'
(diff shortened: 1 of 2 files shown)


sergiou87:
I like this! @tfsJoe is there a reason why you suggested those paths to be separated by spaces instead of new lines?

I see VS Code separates them in lines, so I'd rather stick to some behavior that is considered standard.

tfsJoe:
From the issue:

 > the user may want to do something like `git some-command -- file1 file2 file3`

This is the use case that led me to write the ticket. So I'm trying to avoid an extra step to put those on one line.

sergiou87:
Thank you! I'll discuss it with the team and let you know 😄

sergiou87:
Hey! Sorry for the late response. We just discussed about this, and we think it's better to go with the standard approach of using new lines.

While we agree it might seem very convenient to use spaces for the use case mentioned above, it's also true that in order to make those paths "command line ready" we could need more than just separating them by spaces. For example, if the paths being copied have spaces, those must be escaped or the whole path quoted. That might also vary depending on the specific shell you use.

Using new lines is 1) a broadly adopted convention and 2) forces users to do proper treatment of the copied text before pasting it in a command line, depending on that terminal needs.

tsvetilian-ty:
@sergiou87 the separator is updated to a new line

app/src/ui/changes/changes-list.tsx
```diff
@@ -426,6 +428,32 @@ export class ChangesList extends React.Component<
...
+  private getCopySelectedPathsMenuItem = (
+    files: WorkingDirectoryFileChange[]
+  ): IMenuItem => {
+    return {
+      label: CopySelectedPathsLabel,
+      action: () => {
+        const fullPaths = files.map(file =>
+          Path.join(this.props.repository.path, file.path)
+        )
+        clipboard.writeText(fullPaths.join('\r\n'))
```
sergiou87: Instead of hardcoding `\r\n`, you can [`import { EOL } from 'os'` instead which will change depending on the platform](https://nodejs.org/api/os.html#oseol)

app/src/ui/changes/changes-list.tsx
```diff
@@ -426,6 +428,32 @@ export class ChangesList extends React.Component<
...
+  }
+
+  private getCopySelectedRelativePathsMenuItem = (
+    files: WorkingDirectoryFileChange[]
+  ): IMenuItem => {
+    return {
+      label: CopySelectedRelativePathsLabel,
+      action: () => {
+        const paths = files.map(file => Path.normalize(file.path))
+        clipboard.writeText(paths.join('\r\n'))
```
sergiou87: Same!

sergiou87 (changes requested):

sergiou87 (approved):
Looks great now, thank you! ❤️

PR #14668, opened 2022-05-23
Use diff not log for multi commit diffing

tidy-dev:
## Description
This is first step in the direction to support multi commit diffing for that we need to use `git diff`. This PR is to setup to use `git diff` as opposed to `git log` when more than one commit is selected. This works for one commit as well but due to this being a very impactful change if there are regressions as diff generation is core to the app, I am leaving `git log` in place for single commits. We may decide to remove it once confidence in this approach is built through implementing multi commit diffing. 

Scenarios considered/tested for a single commit: 
- adding
- changing
- renaming
- images
- initial commit
- squashing onto initial commit

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -181,13 +181,12 @@ export async function getCommitRangeDiff(
   // initial commit of a branch) and therefore `SHA^` is not a valid reference.
   // In which case, we will retry with the null tree sha.
   if (result.gitError === GitError.BadRevision && useNullTreeSHA === false) {
-    const useNullTreeSHA = true
     return getCommitRangeDiff(
       repository,
       file,
       commits,
       hideWhitespaceInDiff,
-      useNullTreeSHA
+      true
     )
   }
 
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -27,7 +27,7 @@ import { getOldPathOrDefault } from '../get-old-path'
 import { getCaptures } from '../helpers/regex'
 import { readFile } from 'fs/promises'
 import { forceUnwrap } from '../fatal-error'
-import { git } from '.'
+import { git } from './core'
 import { NullTreeSHA } from './diff-index'
 import { GitError } from 'dugite'
 
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -193,7 +193,7 @@ export async function getCommitRangeDiff(
   if (result.gitError !== null) {
     // This shouldn't happen...
     throw new Error(
-      `getCommitRangeDiff: Error in diffing the commit range of: ${oldestCommitRef} to ${commits[0]}`
+      `getCommitRangeDiff: Error in diffing the commit range of: ${oldestCommitRef} to ${commits[0]} - ${result.combinedOutput}`
     )
   }
 
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -190,13 +190,6 @@ export async function getCommitRangeDiff(
     )
   }
 
-  if (result.gitError !== null) {
-    // This shouldn't happen...
-    throw new Error(
-      `getCommitRangeDiff: Error in diffing the commit range of: ${oldestCommitRef} to ${commits[0]} - ${result.combinedOutput}`
-    )
-  }
-
   return buildDiff(
     Buffer.from(result.combinedOutput),
     repository,
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -151,11 +151,12 @@ export async function getCommitRangeDiff(
     throw new Error('No commits to diff...')
   }
 
-  const oldestCommitRef = useNullTreeSHA ? NullTreeSHA : `${commits.at(-1)}^`
+  const oldestCommitRef = useNullTreeSHA ? NullTreeSHA : `${commits[0]}^`
+  const latestCommit = commits.at(-1) ?? '' // can't be undefined since commits.length > 0
   const args = [
     'diff',
     oldestCommitRef,
-    commits[0],
+    latestCommit,
     ...(hideWhitespaceInDiff ? ['-w'] : []),
     '--patch-with-raw',
     '-z',
@@ -194,7 +195,7 @@ export async function getCommitRangeDiff(
     Buffer.from(result.combinedOutput),
     repository,
     file,
-    oldestCommitRef
+    latestCommit
   )
 }
 
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -27,6 +27,7 @@ import { getOldPathOrDefault } from '../get-old-path'
 import { getCaptures } from '../helpers/regex'
 import { readFile } from 'fs/promises'
 import { forceUnwrap } from '../fatal-error'
+import { git } from '.'
 
 /**
  * V8 has a limit on the size of string it can create (~256MB), and unless we want to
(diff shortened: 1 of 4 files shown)


app/src/lib/git/diff.ts
```diff
@@ -133,6 +133,59 @@ export async function getCommitDiff(
...
+  // hacky? We could derive - git hash-object -t tree /dev/null
+  const emptyTreeHash = '4b825dc642cb6eb9a060e54bf8d69288fbee4904'
+  const commitsToUse = selectionContainsInitialCommit
+    ? [...commits, emptyTreeHash]
+    : commits
+  const commitish =
+    commitsToUse.length === 1
+      ? `${commitsToUse.at(0)}^..${commitsToUse.at(0)}`
+      : `${commitsToUse.at(-1)}..${commitsToUse.at(0)}`
+  const oldestCommit = `${commitsToUse.at(-1)}`
```
tidy-dev: This bit is the major downside to `git diff` over `git log`. Thoughts? 
1) Could leave as is and use this constant.. some googling says it has always been not likely to change.
2) Use `git hash-object -t tree /dev/null` (save the result after first call).. a little future proofing in case this were to ever change.

app/src/lib/git/diff.ts
```diff
@@ -133,6 +133,59 @@ export async function getCommitDiff(
...
+  file: FileChange,
+  hideWhitespaceInDiff: boolean = false,
+  selectionContainsInitialCommit: boolean
+): Promise<IDiff> {
+  if (commits.length === 0) {
+    throw new Error('No commits to diff...')
+  }
+
+  // hacky? We could derive - git hash-object -t tree /dev/null
+  const emptyTreeHash = '4b825dc642cb6eb9a060e54bf8d69288fbee4904'
```
niik: Using the null tree is fine, in fact we've already defined this as a constant in `diff-index.ts` (NullTreeSHA). Might as well pull it out of there and into this file and export it along with a little blurb of documentation.

app/src/ui/dispatcher/dispatcher.ts
```diff
@@ -3552,7 +3558,16 @@ export class Dispatcher {
...
           // than one, just take the most recent. (not likely?)
-          await this.changeCommitSelection(repository, [status.currentTip])
+          await this.changeCommitSelection(
+            repository,
+            [status.currentTip],
+            status.currentBranch !== undefined
+              ? (await getCountCommitsInBranch(
+                  repository,
+                  status.currentBranch
+                )) === 1
```
niik: Counting commits in a repository can be very expensive for large repositories. Instead of using `rev-list --count` here we should look at the commit in question and see if it has any parents. If we have to know whether this commit is the first commit here a specialized method to run `git log -1 --format=%P SHA --` would be my preference

app/src/ui/history/commit-list.tsx
```diff
@@ -272,16 +275,26 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
 
     const selectedShas = sorted.map(r => this.props.commitSHAs[r])
     const selectedCommits = this.lookupCommits(selectedShas)
-    this.props.onCommitsSelected(selectedCommits)
+    const indexes = selectedCommits.map(v =>
+      this.props.commitSHAs.findIndex(sha => sha === v.sha)
+    )
+    const selectionContainsInitialCommit =
+      this.getLastRetainedCommitRef(indexes) === null
```
niik: Is this really doing what we expect it to? Looking at the `getLastRetainedCommitRef` it seems to look at the last commit in the `commitSHAs` list is among the selected ones. That doesn't _necessary_ mean that the first commit of the branch is selected since `commitSHAs` is loaded incrementally.

So if you go to the history tab and hit <kbd>Cmd</kbd>+<kbd>A</kbd> to select all commits the `selectionContainsInitialCommit` will be true and we will diff against the null tree without having selected the initial commit.

I think what we'd want to do here is something like

```suggestion
    const selectionContainsInitialCommit =
      selectedCommits.at(-1)?.parentSHAs.length === 0

**Review by niik — CHANGES_REQUESTED** [review]
This is a tricky one @tidy-dev but I'm wondering if, given how rare it is to diff the initial commit, we should optimize for the case where the initial commit is not part of the selection and get rid of the `selectionContainsInitialCommit` state keeping entirely.

Instead we could attempt to run the diff and detect if it fails due to an invalid revision and in that case diff against the null tree instead.

> ⚠️ Warning. This is a very impactful change if there are regressions as diff generation is core to the app. 😟 Thus, open for any suggestions for different scenarios to test. (Also, is feature flagged)

We could keep using the `log` approach for single-commit diffs for the foreseeable future and only switch to the diff based approach once we've gained enough confidence in it. This has the added benefit of us not having to solve the problem with `selectionContainsInitialCommit` at all for the time being. The more I think about it the more I like this approach, we avoid a lot of edge cases and get two specialized methods, one for single commit diffs, and one for commit ranges (endpoints really).

File: app/src/ui/dispatcher/dispatcher.ts (hunk 3 above)
Comment (tidy-dev) [reply app/src/ui/dispatcher/dispatcher.ts]: :D One of the reasons I split this off into its own PR so early is I wanted to get your all feedback about stuff like this! That makes total sense

File: app/src/ui/history/commit-list.tsx (hunk 4 above)
Comment (tidy-dev) [reply app/src/ui/history/commit-list.tsx]: Rendered moot since changed approach to try, check for bad revision, and then retry with null tree sha.

**Comment by tidy-dev** [comment]
After some discussion, I have updated this PR to not keep track of a state variable to determine if the range is over an initial commit. Switched to @niik suggested approach of trying it and if it fails with a bad revision to retry it with the null tree sha. 

Additionally, added a clause to only use this new diff code if there is more than one sha being diffed (which at this point is not programmatically possible). **Thus, just pulling this PR to test is not enough to run this logic, you must remove the shas.length > 1 ( add in https://github.com/desktop/desktop/pull/14668/commits/772cbfc80d3f45cfaaed5988c0c7cc30231c3ccd) clause after the feature flag check, if you want to try using it to see diffs for a single commit**

File: app/src/ui/dispatcher/dispatcher.ts (hunk 3 above)
Comment (tidy-dev) [reply app/src/ui/dispatcher/dispatcher.ts]: Moot since don't need to track this state

File: app/src/lib/git/diff.ts [hunk 5]
```diff
@@ -133,6 +136,75 @@ export async function getCommitDiff(
...
+  const result = await git(args, repository.path, 'getCommitsDiff', {
+    maxBuffer: Infinity,
+    expectedErrors: new Set([GitError.BadRevision]),
+  })
+
+  // This should only happen if the oldest commit does not have a parent (ex:
+  // initial commit of a branch) and therefore `SHA^` is not a valid reference.
+  // In which case, we will retry with the null tree sha.
+  if (result.gitError === GitError.BadRevision && useNullTreeSHA === false) {
+    const useNullTreeSHA = true
```
tidy-dev: @niik We discussed adding another git call here to verify that the bad revision was caused by the oldest sha not having a parent and doing a non fatal error to track and make sure this isn't a thing, then removing the check in a month or so if we don't get any reports of it. 

I am leaning to this not being worth the effort since it "shouldn't" happen as Desktop is getting the sha's from the `git log` call and therefore they should exist and we shouldn't be get bad revisions (except in this case)

For anyone else following along, the negative consequence of not doing this would be that we could unintentionally attempt a diff for a commit range using the null tree sha that would be very large if the bad revision was due to the oldest commit reference being corrupted. (imagine going from desktop's latest commit in development to the initial commit :D)

niik: That all sounds great. One comment on the line itself though, this is shadowing the original parameter by defining a new const over it. I'd rather we didn't do that as it's easily overlooked and can make it difficult to debug down the line once this is refactored enough. My preference would be to just pass `true` as the last argument to the `getCommitRangeDiff` call below. The comment above is clear enough about what's happening that we don't need the variable.

app/src/lib/git/diff.ts
```diff
@@ -27,6 +27,9 @@ import { getOldPathOrDefault } from '../get-old-path'
 import { getCaptures } from '../helpers/regex'
 import { readFile } from 'fs/promises'
 import { forceUnwrap } from '../fatal-error'
+import { git } from '.'
```
niik: Let's import it straight from the file rather than the index to stay consistent with other imports. The index file is poorly maintained and I wonder if we shouldn't just get rid of it tbh.

```suggestion
import { git } from './core'
```

niik (changes requested):

niik (approved):

PR #14674, opened 2022-05-24
Release 3.0.1-beta2

tidy-dev:
## Description
Looking for the PR for the upcoming second beta of the v3.0.1 series? Well, you've just found it, congratulations!

## Release checklist

### DIFF
--- a/changelog.json
+++ b/changelog.json
@@ -1,7 +1,7 @@
 {
   "releases": {
     "3.0.1-beta2": [
-      "[Fixed] Stream output from merge-tree - #14485",
+      "[Fixed] Prevent crash when encountering a large number of conflicts while checking for ability to merge branch - #14485",
       "[Fixed] Url encode branch names when 'Viewing Branch in Github' is selected - #14631. Thanks @tsvetilian-ty!"
     ],
     "3.0.1-beta1": [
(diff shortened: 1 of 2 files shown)


changelog.json
```diff
@@ -1,5 +1,9 @@
 {
   "releases": {
+    "3.0.1-beta2": [
+      "[Fixed] Stream output from merge-tree - #14485",
```
sergiou87: Not sure if this is the best wording (any suggestions @niik?), but definitely related to what the users experienced.
```suggestion
      "[Fixed] Fix crash determining mergeability of a branch with many conflicts - #14485",
```

sergiou87 (changes requested):

changelog.json
niik: Yeah, that's a great improvement. How about something like

```suggestion
      "[Fixed] Prevent crash when encountering a large number of conflicts while checking for ability to merge branch - #14485",
```

A bit wordy but uses the terminology of desktop and .com rather than my (possibly) made up word "mergeability"

sergiou87: Sold! I'd say we can deploy this one already

sergiou87 (approved):

PR #14679, opened 2022-05-25
Use diff to get changed files for commit range

tidy-dev:
## Description
Same premise as #14668, another step in the direction to support multi-commit-diffing. This PR is to setup to use `git diff` instead of `git log` to obtain a list of changed files when more than one commit is selected.

This works for one commit as well, but due to it being a quite impactful change if there are regressions, I am leaving the `git log` approach in place for single commits. If you would like to test the logic for one commit,  you can modify the `currentSHAs.length > 1` check in the `_loadChangedFilesForCurrentSelection` method in `app-store.ts` to always be true. 

Other notes: In `getChangedFiles` for a single commit, we parse the files changed and added/deleted lines with 2 repeated git calls. I believe this was necessary for feature flagging and reducing potential impact, but that feature flag has been enabled for 9 months and we have not had any complaints. I figured since this will be new diffing anyways a bug would not be a regression, but a bug in multi commit diffs (not as impactful), I made a parser to get the files changed and add/deleted lines in one call git instead. 

Upcoming Prs/things to figure out:
- Refactor UI to show multi commit diff when multiple commits selected (instead of blank slate)
- Refactor UI to have a multi commit summary above diff
- Delve into multi parent scenarios and how we want to handle that.. _merge commits always have to ruin the party..._
- Delve into how the ahead feature in the branch comparison should work for all commits selected. Goal is showing a PR diff if it makes sense.

## Release notes
Notes: no-notes


sergiou87 (dismissed):
Looks great! 🤩

sergiou87 (approved):

PR #14682, opened 2022-05-26
Lets see a multi commit diff

tidy-dev:
This PR builds on #14679

## Description
This PR namely refactors the `SelectCommit` component (that displays the files list, diff, and commit summary in the history view) to accept a selection of multiple commits. Then displays the diff for multiple commits. Additionally, I had to allow to open up some single diff checks in the `app-store.ts` to allow multi commit selections.

This PR also contains a bug fix in the method `updateOrSelectFirstCommit` of `app-store.ts`. The expected behavior is that when you switch branches, this method should select the first commit in the history. I found that if a single commit was selected before branch switching, it just cleared the selection. If multi-commit was selected, it still shows the multi commit view even tho no commits are selected. The later part needed updated as part of this PR anyways.

As you may note from the screenshot, this pr does not include creating a multi-commit summary. (next pr)

Things to yet to figure out...
- merge commit in middle of range.
- non-contiguous selection.

### Screenshots
May-26-2022 14-08-40

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/history/selected-commit.tsx
+++ b/app/src/ui/history/selected-commit.tsx
@@ -34,14 +34,15 @@ import { IChangesetData } from '../../lib/git'
 import { IConstrainedValue } from '../../lib/app-state'
 import { clamp } from '../../lib/clamp'
 import { pathExists } from '../lib/path-exists'
+import { enableMultiCommitDiffs } from '../../lib/feature-flag'
 
-interface ISelectedCommitProps {
+interface ISelectedCommitsProps {
   readonly repository: Repository
   readonly isLocalRepository: boolean
   readonly dispatcher: Dispatcher
   readonly emoji: Map<string, string>
-  readonly selectedCommit: Commit | null
-  readonly isLocal: boolean
+  readonly selectedCommits: ReadonlyArray<Commit>
+  readonly localCommitSHAs: ReadonlyArray<string>
   readonly changesetData: IChangesetData
   readonly selectedFile: CommittedFileChange | null
   readonly currentDiff: IDiff | null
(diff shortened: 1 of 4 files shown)


app/src/ui/history/selected-commit.tsx
```diff
@@ -383,20 +385,21 @@ export class SelectedCommit extends React.Component<
 
     items.push({
       label: viewOnGitHubLabel,
-      action: () => this.onViewOnGitHub(file),
+      action: () => this.onViewOnGitHub(selectedCommits[0].sha, file),
       enabled:
-        !this.props.isLocal &&
+        // TODO: maybe we can assume to use last commit in multi commit
+        // scenario?
+        selectedCommits.length === 1 &&
```
tidy-dev: Thoughts? I believe this should be true. Commits A, B, C and a diff between them (assuming C is the latest), if you look at the file as it is in C, it should be as it is displayed in the diff. The only thing is that we would have to make the localCommitShas verify all the commits in the selection are not local to know it is the same as reflected on dotcom.

tidy-dev: Nvm... I tried it.. since we are specifying the commit in the url.. then we can only reference files in that commit. Thus, if commit A has file X and commit B has file Y. Then, you can't to view file X on GitHub given commit B that has no file X.

tidy-dev: We could throw them to the compare view.. but doesn't appear we can just single out a file that way... so I think it makes sense to just leave this disabled for a multi commit.. or remove altogether

sergiou87 (approved):
I love this! 😍

nikissanez (commented):
Teen

Rs4178 (commented):
Comment

AliIbrahim12:
…

app/src/ui/history/selected-commit.tsx
AliIbrahim12: Refactor SelectedCommit to be SelectedCommits for multi commit diffing

PR #14701, opened 2022-05-31
Fix commit shortcut (Ctrl/Cmd + Enter)

### LINKED ISSUE #14689: Commit button doesn't work with kbd shortcut when an item in the changes list has focus
### The problem

When trying to commit using <kbd>CMD + Enter</kbd>, I just noticed that a change in the changes list is selected/deselected instead of the commit being made.

### Release version

3.0

### Operating system

macOS 12.4

1. Stage a change
2. Select the change from the changes list
3. Try to commit using <kbd>CMD + Enter</kbd> 
4. Note that the selection is toggled on and off

### Log files

_No response_

### Additional context

I can work around this by manually bringing focus to the commit form, but this is annoying, to say the least 😇

tsvetilian-ty:
Closes #14689

## Description
- Fix commit shortcut (Ctrl/Cmd + Enter)

> Note: With this implementation, the shortcut is still coupled to the lifecycle of `CommitMessage` component. 

### Screenshots
fix-commit-shotcut

### DIFF
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -343,8 +343,8 @@ export class CommitMessage extends React.Component<
 
     const isShortcutKey = __DARWIN__ ? event.metaKey : event.ctrlKey
     if (isShortcutKey && event.key === 'Enter' && this.canCommit()) {
-      const openedDialogs = document.getElementsByTagName('dialog')
-      if (openedDialogs.length === 0) {
+      const openDialogs = document.getElementsByTagName('dialog')
+      if (openDialogs.length === 0) {
         this.createCommit()
         event.preventDefault()
       }
--- a/app/src/ui/changes/commit-message.tsx
+++ b/app/src/ui/changes/commit-message.tsx
@@ -183,6 +183,11 @@ export class CommitMessage extends React.Component<
   public componentWillUnmount() {
     const { props, state } = this
     props.onPersistCommitMessage?.(pick(state, 'summary', 'description'))
+    window.removeEventListener('keydown', this.onKeyDown)
+  }
+
+  public componentDidMount() {
+    window.addEventListener('keydown', this.onKeyDown)
   }
 
   /**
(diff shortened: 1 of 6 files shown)


app/src/ui/changes/commit-message.tsx
```diff
@@ -331,15 +336,18 @@ export class CommitMessage extends React.Component<
...
+  private onKeyDown = (event: React.KeyboardEvent<Element> | KeyboardEvent) => {
     if (event.defaultPrevented) {
       return
     }
 
     const isShortcutKey = __DARWIN__ ? event.metaKey : event.ctrlKey
     if (isShortcutKey && event.key === 'Enter' && this.canCommit()) {
-      this.createCommit()
-      event.preventDefault()
+      const openDialogs = document.getElementsByTagName('dialog')
```
niik: We need to check for open foldouts (like the sidebar) here as well. You could check for the presence of an element with id `foldout-container` but I'd rather we proxied `isShowingModal` and `isShowingFoldout` from Repository down to the CommitMessage component and use them instead of looking at the DOM. There's no way for us to know at compile time that this breaks if we inspect the DOM but with a prop we'll know if we attempt to remove it and it won't be tied to the specific markup we choose to use to render dialogs and foldouts.

https://github.com/desktop/desktop/blob/172a9714e17a0365e3e51f3929a7b064d1e9849e/app/src/ui/repository.tsx#L57-L67

niik (changes requested):

app/src/ui/changes/commit-message.tsx
tsvetilian-ty: Thanks for your feedback @niik. I made the changes and hope that now it's ok 🙏

app/src/ui/lib/list/list.tsx
```diff
@@ -537,8 +540,11 @@ export class List extends React.Component<IListProps, IListState> {
   }
 
   private onFocusContainerKeyDown = (event: React.KeyboardEvent<any>) => {
+    const isSpecialShortcut = __DARWIN__ ? event.metaKey : event.ctrlKey
```
niik: Same as above comment

app/src/ui/lib/list/list.tsx
```diff
@@ -524,11 +524,14 @@ export class List extends React.Component<IListProps, IListState> {
       this.props.onRowKeyDown(rowIndex, event)
     }
 
+    const isSpecialShortcut = __DARWIN__ ? event.metaKey : event.ctrlKey
```
niik: Instead of checking whether the keyboard shortcut _isn't_ `CmdOrCtrl+Enter` we should check that it is 'Enter' or 'Space' without any modifiers. The first approach implies that the List component is aware of a specific keyboard shortcut used globally and avoids it whereas the second approach makes the List component specifically look for modifier-less `Enter` or `Space`

I'm working on a PR which will allow us to use Electron-style accelerators but for now checking that no modifiers are set should suffice.

```suggestion
    const hasModifier =  event.altKey || event.ctrlKey || event.metaKey || event.shiftKey
```

niik (changes requested):
Sorry for taking so long to come back to this @tsvetilian-ty. Just a minor request now, this is looking great.

niik (approved):
This is looking great @tsvetilian-ty, sorry again for the wait!

PR #14702, opened 2022-05-31
Non contiguous multi commit selections

tidy-dev:
## Description
The underlying git command to do a multi-commit diff does so by comparing the last commit in a range to the parent of the first. Thus, a selection of multiple commits that is not contiguous does not make sense in this context. This PR provides an updated multiple commit blank slate for that scenario. 

Also, I noticed that our blank screen doesn't mention the squashing/reordering but does cherry-picking. Seems like we should have that hint here too, but I will remove it if others disagree.

Other notes: From what I can tell, you may be able to finagle a "cherry-picked" diff view of non-contiguous commits by finding the files in the diffs of the individual commits and then doing the diff and limiting it to those files, but this would not be equivalent to one git command and may have other pitfalls (merged commits come to mind). Not worth exploring at this point.

### Screenshots
Before:
Screen Shot 2022-05-31 at 2 32 40 PM

After:
Screen Shot 2022-05-31 at 2 28 11 PM

## Release notes
Notes: no-notes


app/src/ui/history/selected-commit.tsx
```diff
@@ -302,11 +302,20 @@ export class SelectedCommit extends React.Component<
         <div className="panel blankslate">
           <img src={BlankSlateImage} className="blankslate-image" />
           <div>
-            <p>Unable to display diff when multiple commits are selected.</p>
+            <p>
+              Unable to display diff when multiple{' '}
+              {enableMultiCommitDiffs() ? 'non-adjacent ' : ' '}commits are
```
tidy-dev: This wording felt more common than `non-contiguous`; but completely open to feedback if there is a clearer way to communicate this. :)

sergiou87: `non-adjacent` works for me, but `non-consecutive` feels more natural maybe (said the non-English speaker 😂 )

tidy-dev: I like that better!

app/src/ui/history/selected-commit.tsx
```diff
@@ -296,11 +302,20 @@ export class SelectedCommits extends React.Component<
...
+              Unable to display diff when multiple{' '}
+              {enableMultiCommitDiffs() ? 'non-consecutive ' : ' '}commits are
+              selected.
+            </p>
             <div>You can:</div>
             <ul>
-              <li>Select a single commit to view a diff.</li>
+              <li>
+                Select a single commit{' '}
+                {enableMultiCommitDiffs() ? 'or a range of commits ' : ' '}to
```
sergiou87: Should we mention the commits should be consecutive here?
```suggestion
                {enableMultiCommitDiffs() ? 'or a range of consecutive commits ' : ' '}to
```

app/src/ui/history/selected-commit.tsx
```diff
@@ -244,10 +247,13 @@ export class SelectedCommits extends React.Component<
...
-    const { selectedCommits } = this.props
+    const { selectedCommits, isContiguous } = this.props
 
-    if (selectedCommits.length > 1 && !enableMultiCommitDiffs()) {
-      return this.renderMultipleCommitsSelected()
+    if (
+      selectedCommits.length > 1 &&
+      (!isContiguous || !enableMultiCommitDiffs())
+    ) {
+      return this.renderMultipleCommitsBlankSlate()
```
sergiou87, line 256: With that new condition, the new wording is gonna show up for users who make non-contiguous selection **even when the multi-commit diff feature flag is disabled**, right? 🤔 

We should probably make sure the feature flag is enabled inside of `renderMultipleCommitsBlankSlate` before showing the multi-commit stuff.

sergiou87 (changes requested):

app/src/ui/history/selected-commit.tsx
tidy-dev: Hmm.. I thought for sure I tested the dev flag on false..

tidy-dev: Feature flag false: 
image

tidy-dev: I think the logic is just confusing, 

It says, when feature flag is false, always render blank slate for multiple commits. But, only show non-contigous msg when feature flag is true.. because multiple commit blank slate only shows when feature flag is true if there are non-contiguous commits.

Or inversely, if feature flag is true, render blank slate only when selection is not contiguous.

sergiou87: Ohhh you're right, my bad!

sergiou87 (approved):
Changes look good and works as expected! :shipit:

app/src/ui/history/selected-commit.tsx
```diff
@@ -244,10 +247,13 @@ export class SelectedCommits extends React.Component<
   }
 
   public render() {
-    const { selectedCommits } = this.props
```
q2wrr2dfmt-maker: ivy

PR #14710, opened 2022-06-01
Fix Markdown highlighting

sergiou87:
## Description

I was pretty annoyed by this error for a while:
image

It only showed up with Markdown files. I can't tell if it every worked looking at the code, but it seems [the Markdown mode relied on `StringStream.lookAhead`](https://github.com/codemirror/CodeMirror/blob/74832f9ccf151f64c1eea53c305f60575cc41d76/mode/markdown/markdown.js#L256), which basically depended on the context/oracle object passed as 3rd parameter of the `StringStream` constructor. This is how it's initialized from codemirror: https://github.com/codemirror/CodeMirror/blob/7e2466f1653faf3832302fdee508ad886d5431e7/addon/runmode/runmode.js#L63-L67

This PR adds the missing `lookAhead` function required by the Markdown mode to our highlighter when it creates the `StringStream` instance.

### Screenshots

Before:
image

After:
image

## Release notes

Notes: [Fixed] Fix Markdown syntax highlighting

### DIFF
--- a/app/src/highlighter/globals.d.ts
+++ b/app/src/highlighter/globals.d.ts
@@ -19,6 +19,7 @@ declare namespace CodeMirror {
   interface StringStreamContext {
     lines: string[]
     line: number
+    lookAhead: (n: number) => string
   }
 
   /**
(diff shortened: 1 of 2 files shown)


tidy-dev (approved):
:sparkles: This is great!

Also totally make sense it is markdown related. I am making test branches with markdown being edited to test out merges with multi commit diffing and am seeing this left and right.

PR #14720, opened 2022-06-02
Handle merged commits in multi commit diffing

tidy-dev:
Build on #14702

## Description

One of the major pitfalls to attempting this multi commit diffing via user selecting a range of diffs is that the git diff command is `oldestsha^..latestSha` in regards to the user selection. This means it is comparing the parent of the oldest sha's tree to the latestSha's tree.  When we merge branch X into branch Y , the commits from branch X show in the history of Y in chronological order. Thus, you could have a history of Y1, X1, Y2, X2, Merge. And if we select Y2 to Y1. We will get the diff of Y1^..Y2 and the commit X1 will not be part of that tree and there fore changes from it will not be displayed. This could be disconcerting to a gui user who looks at this like "I selected Y1, X1, Y2, but X1's changes are not there?". 

This PR uses rev-list to determine the shas that are reachable by following the parent links from the parent of the oldest sha and the latest sha in the users selection. Then inform all the commits in their selection will be diffed.

Further iterations.... going to discuss this in the next team sync, we may want to 1) make the warning more prominent or 2) treat this like a non-contiguous selection and show a blank slate saying something along the lines of the diff of that range is not possible. Either way, the rev-list logic in this PR is necessary. If this doesn't introduce regressions, I would like to merge it and then iterate from here. 

### Screenshots
https://user-images.githubusercontent.com/75402236/171682767-782b9d87-b9d9-4a8b-83bc-f43784631b1d.mov

## Release notes
Notes: no-notes


tidy-dev:
Going with interstitial screen approach instead

PR #14727, opened 2022-06-03
Improve draft-release script

sergiou87:
## Description

This PR makes a few improvements to our `draft-release` script that I've wanted to see for a while:
- It grabs the `Notes: ...` release note from the PRs if it exists.
- If that `Notes:` is set to `no-notes`, that PR will be skipped in the changelog.
- PRs from releases will be skipped (checking the `releases/` prefix).
- When we run `yarn draft-release`, it will create the release branch for us (with the `releases/` prefix)

## Release notes

Notes: no-notes

### DIFF
--- a/script/draft-release/run.ts
+++ b/script/draft-release/run.ts
@@ -51,6 +51,20 @@ async function getLatestRelease(options: {
   return latestTag instanceof SemVer ? latestTag.raw : latestTag
 }
 
+async function createReleaseBranch(version: string): Promise<void> {
+  try {
+    const versionBranch = `releases/${version}`
+    const currentBranch = (
+      await sh('git', 'rev-parse', '--abbrev-ref', 'HEAD')
+    ).trim()
+    if (currentBranch !== versionBranch) {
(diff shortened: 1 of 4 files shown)


script/draft-release/run.ts
```diff
@@ -115,6 +129,10 @@ export async function run(args: ReadonlyArray<string>): Promise<void> {
   })
   const nextVersion = getNextVersionNumber(previousVersion, channel)
 
+  console.log(`Creating release branch for "${nextVersion}"...`)
+  createReleaseBranch(nextVersion)
+  console.log(`Done!`)
```
sergiou87, line 134: I think this is the most controversial change from this PR. What do you think about it?

tidy-dev: I am all for scripting this repetitive step!

tidy-dev (approved):
✨  Thank you! This will be so handy. :)

script/draft-release/run.ts
niik: I love it :shipit:

PR #14740, opened 2022-06-06
#14469 Low res images

### LINKED ISSUE #14469: Low resolution images are downscaled by border
It can be difficult to see minor changes in low resolution images (e.g. pixel art) when they're being scaled down by the inset borders.

### Proposed solution

Adding "box-sizing:content-box;" to the .image-wrapper styles seems to fix it without any issues.

a2937:
Add the "box-sizing:content-box;"  to the image-wrapper styles

Closes #14469

## Description

All I did was edit the image-wrapper in the diff information to make the images to be a tiny bit bigger. 

 https://github.com/a2937/desktop/blob/2ea8cd128e3c9f64ccd560a2aa7b9d9a8b2f387d/app/styles/ui/_diff.scss#L496

### Screenshots

Before: 
Before

After: 
After
## Release notes

Notes:

Edit: Copied the wrong imgur link originally

### DIFF
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
@@ -493,6 +493,7 @@
     min-width: 0;
     min-height: 0;
     text-align: center;
+    box-sizing: content-box;
   }
 
   .image-diff-previous {


a2937:
While the images are not low-resolution it demonstrates the changes.

niik (changes requested):
I'm hesitant to make changes to the image diffs without and looking at the pros and cons, and alternative approaches. We've seen so much churn in the image diffs in the past with each change causing subtle issues. What we've got now has been mostly stable for more than a year and generally does its job. #14469 doesn't go into much detail on the issue itself (nor the suggested fix or how it was reached) which makes it harder to review this.

Taking a quick look at the `img-diffs` branch of https://github.com/desktop/diff-tests/tree/img-diffs shows some small differences in all the image diffing modes which surround the image with a border (all but `difference`)

### Before

image

### After

image

Image size increased, margin to the image metadata decreased.

### Before

image

### After

image

Bottom green border cut off, image sizes changed.

It seems we either need to consider alternative solutions here or add more styles to offset the changes made by switching to content box.

a2937:
Yeah it's unfortunate that they don't go into deep details and I should do more research into pros and cons/alternative solutions. I am going to close this pull request with that. But thank you for your time.

PR #14744, opened 2022-06-06
Remember zoom level on update

### LINKED ISSUE #5315: Zoom level is not retained after updating
I use 125% zoom, and so far every time I've updated the software (with the "restart to update" feature), the zoom level resets, and I have to zoom back in to 125% manually again. It's not a huge deal, just a minor annoyance. (Especially when there are a lot of minor updates after a release)

tidy-dev:
Closes #5315

## Description
This pr fixes the zoom level being reset on each update for windows users. Technically, it is chromium/electron issue as Desktop simply retrieves the zoom level from the electron web contents api. Unfortunately, it seems that is behaving as expected in that it stores the zoom preference in a preferences setting that is based on the applications file path. For windows users, that file path contains the Desktop app version. 

Thus, to fix this, as I can tell, we have three options:
1) Request chromium change where they store their preferences.. not likely I would think...
2) Read the preferences file ourselves and search for the latest desktop version and apply. 
3) Keep track of it in local storage. 

I chose 3 over 2 out of simplicity sake, we can more so trust that local storage retrieval won't change as opposed to a chromium controlled preference file, and just local storage usage convention in the Desktop app. 

Notes:  
- This approach only sets the zoom to the locally stored value when it receives the default value and the locally stored value differs. Thus, most of the time, it will just rely on the current functionality. 
- If user clears their local storage or locally stored files, this will reset. 

## Release notes
Notes: [Fixed] On Windows, remember the apps zoom level on update.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -577,11 +577,10 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   private initializeZoomFactor = async () => {
-    let zoomFactor = await getCurrentWindowZoomFactor()
+    const zoomFactor = await this.getWindowZoomFactor()
     if (zoomFactor === undefined) {
       return
     }
-    zoomFactor = this.checkZoomFactorLocalStorage(zoomFactor)
     this.onWindowZoomFactorChanged(zoomFactor)
   }
 
@@ -592,7 +591,8 @@ export class AppStore extends TypedBaseStore<IAppState> {
    * Thus, on every update, the users set zoom level gets reset as there is not
    * defined value for the current app version.
    * */
-  private checkZoomFactorLocalStorage(zoomFactor: number) {
+  private async getWindowZoomFactor() {
+    const zoomFactor = await getCurrentWindowZoomFactor()
     // One is the default value, we only care about checking the locally stored
     // value if it is one because that is the default value after an
     // update
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -585,23 +585,26 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.onWindowZoomFactorChanged(zoomFactor)
   }
 
-  /** 
+  /**
    * On Windows OS, whenever a user toggles their zoom factor, chromium stores it
    * in their `%AppData%/Roaming/GitHub Desktop/Preferences.js` denoted by the
    * file path to the application. That file path contains the apps version.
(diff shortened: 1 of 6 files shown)


app/src/ui/main-process-proxy.ts
```diff
@@ -155,6 +155,9 @@ export const getCurrentWindowZoomFactor = invokeProxy(
   0
 )
 
+/** Tell the main process to obtain the current window's zoom factor */
```
sergiou87:
```suggestion
/** Tell the main process to set the current window's zoom factor */
```

app/src/main-process/app-window.ts
```diff
@@ -416,6 +416,10 @@ export class AppWindow {
     return this.window.webContents.zoomFactor
   }
 
+  public setWindowZoomFactor(zoomFactor: number) {
+    return (this.window.webContents.zoomFactor = zoomFactor)
```
sergiou87: No need to return anything here, right? 🤔 
```suggestion
   this.window.webContents.zoomFactor = zoomFactor
```

app/src/lib/stores/app-store.ts
```diff
@@ -575,13 +577,41 @@ export class AppStore extends TypedBaseStore<IAppState> {
   }
 
   private initializeZoomFactor = async () => {
-    const zoomFactor = await getCurrentWindowZoomFactor()
+    let zoomFactor = await getCurrentWindowZoomFactor()
     if (zoomFactor === undefined) {
```
sergiou87, line 581: Shouldn't we check the zoom factor stored locally even if Chromium gives us no zoom factor? 🤔

sergiou87 (changes requested):
Looks good overall, just a couple of comments

app/src/lib/stores/app-store.ts
tidy-dev: Technically it "shouldn't" be undefined. The undefined comes in to play because our browser window in the main process starts out as null. But, by the time this is called, it "should" always be available and therefore this should always return a value. Thus, the check is really just for typing sake. 

However.. after you said that, I realized this bit of logic could be slightly more succinct if I move that electron zoom factor fetching call into my checking method, and we get the double benefit of returning locally store if in some odd ball case in the future it did return undefined.

app/src/main-process/app-window.ts
tidy-dev: Haha.. yep sry copy/pasta remnant

sergiou87 (approved):
Looks great! Thanks for those changes ✨🤩✨

PR #14748, opened 2022-06-07
Fix issue with some repositories not being properly persisted

sergiou87:
## Description
I noticed that _sometimes_ Desktop wasn't able to move repositories to the right owner in the list (leaving them in `Other`) under some circumstances, along with this console error:

image

It turns out Dexie doesn't like returning non-promise values from async functions used in transactions: https://dexie.org/docs/DexieErrors/Dexie.PrematureCommitError.html

This is something we've run into in the past: https://github.com/dexie/Dexie.js/issues/1186

This PR wraps the resulting `GitHubRepository` instance in a `Promise.resolve(…)` to make Dexie happy.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/repositories-store.ts
+++ b/app/src/lib/stores/repositories-store.ts
@@ -124,7 +124,7 @@ export class RepositoriesStore extends TypedBaseStore<
       )
     }
 
-    return new GitHubRepository(
+    const ghRepo = new GitHubRepository(
       repo.name,
       owner,
       repo.id,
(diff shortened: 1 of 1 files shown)


niik (approved):
I can't say that I understand how this would affect the outcome of Dexie's promise counting but if we can reliably show that the error doesn't happen with this change I can't see any downside to it either. I would also really like to know what changed in Dexie to cause this, the micro await limit has been increase as I understand it so it's something more subtle than that then.

tidy-dev (approved):
✨ I think it make sense.. tho the whole dexie promise thing I haven't wrapped my head around yet. :D

PR #14754, opened 2022-06-07
Put the commit sha in the history

### LINKED ISSUE #14611: Use Standard Date/Time stamp and Commit ID for GUI
### The feature request

I love the History, but what is very useless in the gui is how the History of anything is presented.  Currently they are using the "Friendly" date format, which is vague and ambiguous, ie: "yesterday, few minutes ago, 2hours ago".
This is for both GitHub Desktop and also for the web application.
Additionally, in the Desktop, it would be great if we could also see the Commit ID in the list view.  It would be an additional bonus if we could maybe see the parent commits links

In order to see the time I made the commit/change, I have to try to find the "sweet spot" for where to hover over (the whole list item is not triggering the hover...)
image

When something goes wrong, we (developers) usually have a pretty good idea on the time we made a particular change, and if we do commit frequently, in order to find the right history to revert to: we literally have to go through a lot of list items that will have the identical friendly date.
If I could see the actual time stamp, it would drastically cut down on how many items I need to review to find the right item.

### Proposed solution

Instead of using the friendly date format, either make it customizable where the user could set a preference for date format, or just change the friendly date format to Standard date format.
I would love it if I could choose to see any date related data in the GUI in this format: (in **BLACK** font and  NOT some light color font that barely separates from the background...)
**05/10/2022 22:44:17**
Can you please consider making this change or add a settings where we could choose date format?
I think tons of us, end-users would really appreciate some ability to customize the interface to what works for us.
image

### Additional context

_No response_

a2937:
Closes #14611

## Description
- Adds the commit id next to the date-time of the commit. Rather than the parent id listed , it would be more beneficial just to scroll down. Considerations were made for the commit id Octicon ,however it seemed very difficult if not impossible for it to look right. Additionally I am not going to attempt to implement the long date feature toggle. It seemed to be a design feature that would cause more harm than good , especially since people can just hover over the friendly date format to get the actual time. 

### Screenshots

Commit history 

## Release notes

Notes:


tidy-dev:
This is being discussed but will likely hang out here for a bit as the decision is being made in relation to some ongoing feature work. We are not sure the additional clutter in an already cramped commit list item is ideal. Tho the request is understandable; a user can see this info upon clicking the commit in the commit diff summary.

tidy-dev:
Thank you for wanting to improve GitHub Desktop! However, we discussed this as a team and think the commits list item is already two cluttered when viewing the minimum constraints (small width plus tags). Additionally, a user can view the commit sha in a the commit summary and copy the commit sha via the context menu if this info is needed.

Qf4ctor:
> Additionally I am not going to attempt to implement the long date feature toggle. It seemed to be a design feature that would cause more harm than good , especially since people can just hover over the friendly date format to get the actual time.

My personal experience is that I didn't know I could hover over it to know the date. Believe it or not, I learned about it in this post! I don't think it is intuitive. On the other hand displaying the day in the 2024-03-01 format is short and clear. It is easy to see it happened yesterday or months ago by a quick look at the date in this format. I would only optionally display the hour,minute and second of the commit.

It is also worth comparing the length of the following strings:
"2024-03-01 23:59"
"23 minutes ago"
"10 months ago"
(The first string packs a hell lot of more information in just 2 more characters)

PR #14762, opened 2022-06-08
Add Actions no-response workflow

steveward:
## Description

Our Probot no-response bot has been broken -- this PR replaces that with a GitHub Actions workflow using [lee-dohm/no-response](https://github.com/lee-dohm/no-response).

I used mostly the same messaging, but added a note that the issue will automatically reopen if the original author replies. I kept the `daysUntilClose` at seven days.

## Release notes

Notes:


sergiou87 (commented):
Thank you for this!! ❤️ 

Seems like you must run the linter in order to make the checks mass, but other than that looks good to me 🤩

sergiou87 (approved):

PR #14781, opened 2022-06-10
Remove explicit opt-in to protocol version 2

niik:
## Description

Git protocol version 2 was [made the default in Git 2.26](https://github.blog/2020-03-22-highlights-from-git-2-26/) so we can remove our explicit opt-in which we first implemented (behind a feature flag) in https://github.com/desktop/desktop/pull/6142.

This saves us one `git config` process per network operation (so hardly signifiant performance wise) and lets us clean up some dead code.

## Release notes

Notes: no-notes


sergiou87 (approved):
Nice cleanup! 🙌

PR #14782, opened 2022-06-10
Avoid shelling out for trailer separator characters if able

niik:
## Description

Trailers returned from `git log`'s `%(trailers:unfold,only)` will always format the trailers with `:` as the separator. I don't believe this was always the case or perhaps it just wasn't documented when I wrote this initially. Either way, we can skip one `git config` call for each time we load commits.

Additionally we can optimize the restoring of commit message after undo slightly to only load separator chars when we encounter a co-authored-by trailer. Unfortunately `interpret-trailers` doesn't support providing a specific separator. If it did there's plenty more we could optimize.

Finally this fixes a potential problem where triple dashes (`---`) are [treated as the end of the commit message](https://github.com/git/git/commit/1688c9a4894df517241026c7a3848bdc84607986). I only encountered this when I was reading up on the documentation for `interpret-trailers` but it makes sense for us to tell `interpret-trailers` that what we've got is a commit message and not an email.

## Release notes

Notes: no-notes


sergiou87 (approved):
Looks good! :shipit:

PR #14791, opened 2022-06-12
Fix commit description overflowing with 3 lines when it shouldn't

HeCorr:
## Description
If the commit description has exactly 3 lines, it is detected as being overflown when it shouldn't be, and clicking the Expand button doesn't do anything.

This PR improves the CSS code which not only fixes the issue described above but also makes it easy to change the maximum amount of lines shown in the future (as demonstrated in the second video).

This was only tested on Windows with DevTools because I don't have different systems to test this on and don't have an Electron environment set up on my machine.

Thanks to @sergiou87 for the CSS code.

### Before

https://user-images.githubusercontent.com/75134774/177019862-816692b1-bf40-4518-be99-31578b0b9a10.mp4

### After

https://user-images.githubusercontent.com/75134774/177019870-38cc6a29-40f7-46fa-ace9-88a0eceae69d.mp4

## Release notes

Notes: Fixed commit description with 3 lines overflowing when it shouldn't


tidy-dev:
@HeCorr Thank you for wanting to improve GitHub Desktop! However, this is actually by design. The idea behind it is is that in the case where the description is long enough to overflow (3+ lines), we show a hint of the next line to indicate that there is more to see. I understand where this feels weird for the 3 line case, but there needs to be cut off somewhere. 

One could argue that we only do this on the 3rd line when there are 4+ lines, but that would require a different approach to the issue (likely refactoring the way this works to use js to calculate overflow as opposed to css).

HeCorr:
I think you misunderstood the issue here. There were only 3 lines there, there's nothing else to Expand to, so it doesn't make sense to click Expand just to reveal nothing. It's just visually distracting when you only have three lines.

This fix will still show the Expand button when there are **more** than 3 lines, it didn't remove it.

I do understand this is a temporary solution, but is a solution nonetheless. I rather have it fixed with CSS now because who knows when someone will get to refactor this.

Could you please re-evaluate your decision? If not I would like a second opinion from a different maintainer.

HeCorr:
@tidy-dev

tidy-dev:
> I do understand this is a temporary solution, but is a solution nonetheless. I rather have it fixed with CSS now because who knows when someone will get to refactor this.

I suppose that was my point, we would rather accept a full solution than a partial one. :) 

> If not I would like a second opinion from a different maintainer.

The GitHub Desktop maintainers are a team. Thus, requesting another maintainer likely won't get you very far. :)  If you would like to continue contributing, I would suggest you read up on respectful disagreement in our [Code of Conduct](https://github.com/desktop/desktop/blob/development/CODE_OF_CONDUCT.md).

However, I decided to talk with the team about the issue and it has been discussed in the past. Here https://github.com/desktop/desktop/issues/5700  and then addressed here https://github.com/desktop/desktop/pull/5711

Looking at https://github.com/desktop/desktop/pull/5711, it should work as I described where 3 lines does not show expand but 4 lines not only show the expand, but also show shadow on 3 lines and expand.  (the part I was concerned about)

Going to reopen your PR and test it out and see if the shadow shows as expected at 4 lines. If so, we will see about putting it through. :)

HeCorr:
> I suppose that was my point, we would rather accept a full solution than a partial one. :)

I respect that, but unfortunately I'm not able to provide a full solution, hence this simple 1 pixel change which should last for another year or two as long as the UI layout doesn't change much.

> If you would like to continue contributing, I would suggest you read up on respectful disagreement in our Code of Conduct.

I personally don't believe I was disrespectful in any way. I just politely asked for a second opinion if possible.
Due to the simplicity of the fix, it didn't make sense to me that it got rejected so quickly.
But if you felt offended, my apologies.

> (...) it should work as I described where 3 lines does not show expand but 4 lines not only show the expand, but also show shadow on 3 lines and expand.

That's **exactly** what this fix accomplishes.
I thought I had mentioned this on my original message but if you read my commit you can see that it mentions the [original fix](https://github.com/desktop/desktop/commit/b8ee7244c5248ec22e31e04b4b9c9ba1264fdf37) by @aryyya (which was bumping a CSS value, exactly like mine).
The layout has certainly changed since 2018 and the old fix stopped working. That's all this is, a simple "re-fix" if you will.

> Going to reopen your PR

Thank you very much!

app/styles/ui/history/_commit-summary.scss
```diff
@@ -104,7 +104,7 @@
   &-description-scroll-view {
     overflow: hidden;
     // The extra pixel makes the space align up with the commit list.
-    max-height: 69px;
+    max-height: 70px;
```
sergiou87: I've been investigating and I think this set of CSS properties would make the solution work in case we change the layout again:
```suggestion
    display: -webkit-box;
    -webkit-box-orient: vertical;
    -webkit-line-clamp: 3;
```
No more need for pixel perfect updates!

sergiou87 (changes requested):

app/styles/ui/history/_commit-summary.scss
HeCorr: I just tried out these changes in DevTools (because I don't have Electron set up) and it doesn't seem to actually expand when I click the button. Let me know if I'm doing it wrong.

https://user-images.githubusercontent.com/75134774/174136876-6ea87d4d-3265-442e-99ba-17f36721cec3.mp4

sergiou87: Weird, it worked when I tested it 🤔 

In any case, l played a bit more with it, and you need to make this other change to the `expanded` class:
```diff
diff --git a/app/styles/ui/history/_commit-summary.scss b/app/styles/ui/history/_commit-summary.scss
index 704ba0762..4ad369c05 100644
--- a/app/styles/ui/history/_commit-summary.scss
+++ b/app/styles/ui/history/_commit-summary.scss
@@ -20,20 +20,21 @@
     svg {
       vertical-align: text-top;
       margin-right: var(--spacing-half);
     }
   }

   &.expanded {
     .commit-summary-description-scroll-view {
       max-height: 400px;
       overflow: auto;
+      display: revert;

       &:before {
         content: none;
       }
     }
   }

   &.hide-description-border {
     .commit-summary-description-container {
       border-bottom: none;
```

This will revert the `display: -webkit-box;` from the other set of changes and will effectively expand the text:

https://user-images.githubusercontent.com/1083228/174261558-0930f87e-7b6d-474a-9335-64ad107aa038.mov

HeCorr: ~~I'm confused about what lines you actually changed, because your diff only states `display: revert` but the rest of the attributes are completely different on my system (i.e. `max-height: 400px` on `#commit-summary.expanded` which is `none` here).~~

~~I did my best to make my CSS as close as possible to yours and it still doesn't work.~~

~~Tomorrow I'll setup a dev environment and try editing the source .scss files instead of using DevTools to see if it will work.~~

HeCorr: Sorry, I just tried again and it worked. Please disregard the comment above.

I'll be doing more tests tomorrow and committing the new changes.

HeCorr:
My sincere apologies for not working on this when I said I was going to.
I've been very busy lately but I'll finish it **right now**.

HeCorr:
@tidy-dev I've implemented a more permanent solution and updated the PR's description, please take a look at it when you can!

HeCorr:
@sergiou87 could you please review the latest changes?

tidy-dev (approved):
✨ This works great! Thank you for your patience.

sergiou87 (approved):
Looks great!! Thank you 🤩

HeCorr:
Thank you very much!

PR #14807, opened 2022-06-15
[WIP] Text selection is too tall in diff view’s shrunk code element

### LINKED ISSUE #14760: Text selection is too tall in diff view’s shrunk code element
### The problem

In diff view’s shrunk code element (the one with blue background; see screenshot), text selection spans across the entire height of the element, while it should only cover the actual text.

### Release version

3.0.1

### Operating system

macOS 12.4

1. Make a commit with changes in two places of a single file, distant from each other;
2. Open the commit in History;
3. Select some text in the pane that appears between changed regions and allows to expand code (“shrunk code element”).

### Log files

_No response_

### Additional context

_No response_

biel-correa:
Closes #14760

## Description

- Reduced the height of diff view’s shrunk code element so when the text is selected it doesn`t get a big area

#### Before:
image

#### After:
image

## Release notes

Notes:

### DIFF
--- a/app/styles/ui/_diff.scss
+++ b/app/styles/ui/_diff.scss
@@ -33,8 +33,8 @@
     padding: 2px 0 2px var(--spacing-half);
 
     &.diff-hunk-expandable-both {
-      padding-top: 14px;
-      padding-bottom: 14px;
+      padding-top: 7px;
+      padding-bottom: 7px;
     }
 
     &.diff-hunk:not(.diff-hunk-expandable-both) {


niik:
Hey @biel-correa, seems this has been marked as WIP for a while and there doesn't seem to be consensus in the original issue on how to address this properly. That coupled with our efforts to get rid of CodeMirror (the responsible component for unified diff mode) leads the to think we should close this out for now. Hopefully this will get resolved as soon as we're able to switch over to the new rendering engine.

PR #14814, opened 2022-06-15
Adding new history tab mode for diffing commits

tidy-dev:
## Description
This PR adds a new history tab mode and thus a new compare view to the app state. It is not implemented; just the logic to support a new the commit range diff view.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -55,7 +55,7 @@ interface ICommitProps {
   readonly showUnpushedIndicator: boolean
   readonly unpushedIndicatorTitle?: string
   readonly unpushedTags?: ReadonlyArray<string>
-  readonly isCherryPickInProgress?: boolean
+  readonly disableCherryPicking?: boolean
   readonly disableSquashing?: boolean
 }
 
@@ -391,6 +391,7 @@ export class CommitListItem extends React.PureComponent<
           ? `Squash ${count} Commits…`
           : `Squash ${count} commits…`,
         action: this.onSquash,
+        enabled: !this.props.disableSquashing,
       })
     }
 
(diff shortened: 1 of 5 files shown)


app/src/ui/history/commit-list-item.tsx
```diff
@@ -391,19 +391,19 @@ export class CommitListItem extends React.PureComponent<
           ? `Squash ${count} Commits…`
           : `Squash ${count} commits…`,
         action: this.onSquash,
+        enabled: !this.props.disableSquashing,
```
tidy-dev, line 394: Noticed when testing the availability of squashing/reordering/cherry-picking that squashing via the context menu was not being disabled - only through drag and drop was being disabled. 🤦  This is a fix. :)

app/src/ui/history/commit-list-item.tsx
```diff
@@ -391,19 +391,19 @@ export class CommitListItem extends React.PureComponent<
...
   }
 
   private canCherryPick(): boolean {
-    const { onCherryPick, isCherryPickInProgress } = this.props
+    const { onCherryPick, disableCherryPicking } = this.props
     return (
       onCherryPick !== undefined &&
       this.onSquash !== undefined &&
-      isCherryPickInProgress === false
-      // TODO: isSquashInProgress === false
```
tidy-dev, line 406: >TODO: isSquashInProgress === false 

Need to revisit this.. separate PR. I tested it and it is a bug.. but not as ugly as a bug as I expected.. the clicking a multi commit action while another is in progress (during a merge conflict) just doesn't work/do anything.

app/src/ui/history/commit-list-item.tsx
Hacker00008: > Noticed when testing the availability of squashing/reordering/cherry-picking that squashing via the context menu was not being disabled - only through drag and drop was being disabled. 🤦 This is a fix. :)

`[
Duplicate of #](url)`

Laith994: كاش

PR #14818, opened 2022-06-16
 Fix crash when user's locale is unsupported by the spellchecker

### LINKED ISSUE #14817: Application crashes upon setting spellcheck to system language
### The problem

When I accidentally clicked `Set spellcheck to system language`, the whole application crashed, yielding the following error:
```
2022-06-16T09:22:55.962Z - error: [main] Error: Invalid language code provided: "zh-CN" is not a valid language code
    at click (C:\Users\<username>\AppData\Local\GitHubDesktop\app-3.0.1\webpack:\[name]\app\src\main-process\menu\build-spell-check-menu.ts:109:26)
    at MenuItem.click (node:electron/js2c/browser_init:73:1908)
    at EventEmitter.a._executeCommand (node:electron/js2c/browser_init:81:2453)
```

I've tested this multiple times, and every attempt resulted in a crash.

The problem probably has something to do with this:
https://github.com/desktop/desktop/blob/2389b82ba1058dbe28d45c293647362514bb0eb6/app/src/main-process/menu/build-spell-check-menu.ts#L109

### Release version

Version 3.0.2 (x64)

### Operating system

Windows 10

1. Set system language to Simplified Chinese
2. Right-click on a misspelled word
3. Select `Set spellcheck to system language`
4. (Crashes)

### Log files

_No response_

### Screenshots

image

---

image

### Additional context

_No response_

tsvetilian-ty:
Closes #14817

## Description
Added an additional check to make sure the user's locale is part of the `availableSpellCheckerLanguages` ([docs](https://www.electronjs.org/docs/latest/api/session#sesavailablespellcheckerlanguages-readonly))


tidy-dev:
Beat me to it - I was just starting to take a look at a fix for this. :D

tidy-dev (changes requested):
Thanks for picking this up!

However, I tested this logic by simply hard coding the user language as `zh-CN` and this still crashes. I believe that is do the this check to return null being a && when it needs to be an ||.

However.. I was starting down the path of making this function unit testable since I can't easily test it on my mac and feel like since there are at least 4 code paths in it, it would be nice to have. Thus, need to make this function accept the things it retrieves from the session as parameters and then added spell-checker-menu-test.ts to the test/main-process directory. Could you update your PR for that? (If not, I can forge on with what I started :)) 

Here are the 4 scenarios I was starting test for off the top of my head:
- spell checker language is english && user language is english -> return null
- spell checker language is english && user language is not english && user language is supported -> return `set spellcheck to system language` menu item
- spell checker language is english && user language is not english && user language is not supported -> return null
- spell checker is not english, return `Set spellcheck to English` menu item.

tsvetilian-ty:
@tidy-dev nice catch!

I changed the function to be more testable and added the four cases you suggested.
I had an issue while running the tests with the return type `MenuItem` `TypeError: electron_1.MenuItem is not a constructor ` I didn't spend too much time on it and ended up using a more generic approach. This way, the function is completely decoupled from any electron dependencies. 

Please share if you think there is a better way of doing it.

tidy-dev (dismissed):
✨  Thanks for those changes!  I tested it on my windows vm. Set my system language to Spanish and then to the zh-ch that the user reported with. Received the menu items as expected for Spanish and did not receive a menu item for the Chinese (and Chinese defaulted to English per the library electron is using). 

Random thoughts: I am thankful for system iconography. I know enough Spanish to recognize enough words to bounce around.. but Chinese I was lost. :D

app/src/main-process/menu/build-spell-check-menu.ts
```diff
@@ -104,8 +127,8 @@ function getSpellCheckLanguageMenuItem(
       ? 'Set spellcheck to English'
       : 'Set spellcheck to system language'
 
-  return new MenuItem({
-    label,
-    click: () => session.setSpellCheckerLanguages([languageCode]),
-  })
+  return {
+    label: label,
```
sergiou87: Nit:
```suggestion
    label,
```

app/src/main-process/menu/build-spell-check-menu.ts
```diff
@@ -60,35 +60,58 @@ function getSpellCheckMenuItems(
...
+ * If system language is english or it's not part of the available languages,
+ * it returns null. If spellchecker is not set to english, it returns options
+ * that can set it to English. If spellchecker is set to english, it returns
+ * the options that can set it to their system language.
  */
-function getSpellCheckLanguageMenuItem(
-  session: Electron.Session
-): MenuItem | null {
-  const userLanguageCode = app.getLocale()
+export function getSpellCheckLanguageMenuItemOptions(
```
sergiou87, line 113: Since this is now exported, we'd rather have its parameters also documented in the comment above 😳

app/test/unit/main-process/spell-checker-menu-test.ts
```diff
@@ -0,0 +1,58 @@
...
+import { getAvailableSpellcheckerLanguages } from '../../helpers/menus/available-spellchecker-languages-helper'
+
+describe('spell-checker-menu', () => {
+  it('spell checker language is english && user language is english', () => {
+    const spellcheckerLanguage = ['en-US']
+    const userLanguageCode = 'en-US'
+
+    const spellcheckerMenuItem = getSpellCheckLanguageMenuItemOptions(
+      userLanguageCode,
+      spellcheckerLanguage,
```
sergiou87: Given `userLanguageCode` and `spellcheckerLanguage` are not used anywhere else, you can just get rid of them:
```suggestion
    const spellcheckerLanguage = ['en-US']
    const userLanguageCode = 'en-US'

    const spellcheckerMenuItem = getSpellCheckLanguageMenuItemOptions(
      'en-US',
      ['en-US'],

```
(You might need to run the linter if you accept these changes).

This also applies to the other tests

app/test/unit/main-process/spell-checker-menu-test.ts
```diff
@@ -0,0 +1,58 @@
+import { getSpellCheckLanguageMenuItemOptions } from '../../../src/main-process/menu/build-spell-check-menu'
+import { getAvailableSpellcheckerLanguages } from '../../helpers/menus/available-spellchecker-languages-helper'
+
+describe('spell-checker-menu', () => {
+  it('spell checker language is english && user language is english', () => {
```
sergiou87: For the test description, we use sentences that describe how the subject under test should behave (where `it` is the subject under test):
```suggestion
  it('returns null menu item options when both spellchecker language and user language are english', () => {

```

app/test/unit/main-process/spell-checker-menu-test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const spellcheckerMenuItem = getSpellCheckLanguageMenuItemOptions(
+      userLanguageCode,
+      spellcheckerLanguage,
+      getAvailableSpellcheckerLanguages()
+    )
+
+    expect(spellcheckerMenuItem?.label).toBe(
+      'Set spellcheck to system language'
+    )
```
sergiou87: Not extremely concerned about this, but I think it's better if we don't use specific user-facing strings in tests. Maybe we could do something like…
```suggestion
    expect(spellcheckerMenuItem?.label).not.toBeFalsy()
```

That will check the label is not undefined, nor null, nor an empty string.

sergiou87 (changes requested):
Thanks for your contribution! Just a few things I spotted 😄

app/test/unit/main-process/spell-checker-menu-test.ts
tidy-dev: I am not sure I agree with this one. I get the idea - but, the label is the result we are checking for. There could be 2 different labels or null. If we want the test to be agnostic, we could put the labels in an exportable constant used by the implementation and the test.

tsvetilian-ty: I refactor the tests and impl to rely on constants as @tidy-dev suggested!

app/test/unit/main-process/spell-checker-menu-test.ts
tsvetilian-ty: @sergiou87 initially I implemented the tests without `userLanguageCode` and `spellcheckerLanguage`. I added them after that to increase readability because the parameters looked too similar even though they are different types.

app/test/unit/main-process/spell-checker-menu-test.ts
sergiou87: Sounds good, thank you both! ❤️

sergiou87 (approved):
Thanks for those changes 🙌 :shipit:

PR #14819, opened 2022-06-16
Verify sender of IPC messages

sergiou87:
## Description

Following [Electron's security documentation](https://github.com/electron/electron/blob/main/docs/tutorial/security.md#17-validate-the-sender-of-all-ipc-messages), we should validate the sender of IPC messages.

## Release notes

Notes: no-notes


app/src/main-process/ipc-main.ts
```diff
@@ -40,10 +45,22 @@ export function once<T extends keyof RequestChannels>(
...
-    event: IpcMainInvokeEvent,
-    ...args: Parameters<RequestResponseChannels[T]>
-  ) => ReturnType<RequestResponseChannels[T]>
+  listener: RequestResponseChannelListener<T>
 ) {
-  ipcMain.handle(channel, (event, ...args) => listener(event, ...(args as any)))
+  ipcMain.handle(channel, safeListener(listener))
+}
+
+const validSenderUrl = encodePathAsUrl(__dirname, 'index.html').toString()
```
niik: This probably needs to be a bit more lax to account for us sending IPC messages from the crash process as well?

app/src/main-process/ipc-main.ts
```diff
@@ -40,10 +45,22 @@ export function once<T extends keyof RequestChannels>(
...
+  return (event: IpcMainEvent | IpcMainInvokeEvent, ...args: any[]) => {
+    if (event.senderFrame.url !== validSenderUrl) {
+      log.error(
+        `IPC message received from invalid sender: ${event.senderFrame.url}`
+      )
+      return
+    }
+
+    return listener(event, ...args)
+  }
```
niik: You could do something like this to at least enforce that the listener takes an IPC event as its first parameter. Not much type safety but a tad bit more at least

```suggestion
function safeListener<E extends IpcMainEvent | IpcMainInvokeEvent, R>(
  listener: (event: E, ...a: any) => R
) {
  return (event: E, ...args: any) => {
    if (event.senderFrame.url !== validSenderUrl) {
      log.error(
        `IPC message received from invalid sender: ${event.senderFrame.url}`
      )
      return
    }
    return listener(event, ...args)
  }
```

niik (changes requested):
Looking good. I wonder if we could be more relaxed about the source url perhaps. I'm a little concerned about users running Desktop in non-standard locations, behind symlinks, network drives and what not that would end up having validSenderUrl not match senderFrame.url even though they're actually the same path.

What's the source url for a markdown frame for example? I'm guessing it's some opaque origin. Could we get away with just checking that the source is under the file protocol?

app/src/main-process/ipc-main.ts
sergiou87: Love it, thanks! ❤️

sergiou87:
Thanks for your feedback, I think you made a great point 🙇‍♂️  Using the `webContents.id` of each window (crash and app) should allow us to identify the crash and renderer processes and therefore make sure the IPC messages are sent from our own code.

I followed that approach in my latest changes although I don't like the hacky `export` of the main and crash windows… 😕 I'm trying to think of a better way of doing this without making things much more complicated but so far… 🤔

niik:
> I followed that approach in my latest changes although I don't like the hacky `export` of the main and crash windows… 😕 I'm trying to think of a better way of doing this without making things much more complicated but so far… 🤔

Could we use a glorified global variable to hide the ugliness?

`trusted-ipc-sender.ts`

```ts
const trustedSenders = new Set<number>()

export const addTrustedIPCSender = (wc: WebContents) => {
  trustedSenders.add(wc.id)
  wc.on('destroyed', () => trustedSenders.delete(wc.id))
}

export const isTrustedIPCSender = (wc: WebContents) => trustedSenders.has(wc.id)
```

And then let `app-window.ts` and `crash-window.ts` register the window on creation?

sergiou87:
I loved that, thank you!! I just implemented those changes in [15aee6d](https://github.com/desktop/desktop/pull/14819/commits/15aee6d037d7c1108c636ef33a643cd82e44bea2)

app/src/main-process/crash-window.ts
```diff
@@ -51,11 +52,16 @@ export class CrashWindow {
...
+    addTrustedIPCSender(this.window.webContents)
 
     this.error = error
     this.errorType = errorType
   }
 
+  public get webContents(): WebContents {
+    return this.window.webContents
+  }
+
```
niik: We can get rid of this now right?

```suggestion
```

app/src/main-process/app-window.ts
```diff
@@ -211,6 +214,10 @@ export class AppWindow {
     return !!this.loadTime && !!this.rendererReadyTime
   }
 
+  public get webContents(): WebContents {
+    return this.window.webContents
+  }
+
```
niik: We can get rid of this now right?

```suggestion
```

niik (dismissed):
This looks great @sergiou87, I just had two small comments but I'm approving this since I'll be out on vacation when you see this. Feel free to merge as-is or apply my suggestions, I'm happy either way!

sergiou87:
@tidy-dev could you re-approve this? I just removed two unnecessary properties in [0276dcf](https://github.com/desktop/desktop/pull/14819/commits/0276dcfa55b1c6d13acf45f704f62e7bb8baf4a9) since Markus' approval 🙏

niik (approved):
One last little review :)

shfog4449 (commented):
- [ ] [
Opera Snapshot_2022-07-07_084946_eloan sidf gov sa
](url)

Issue #14666, opened 2022-05-23
Refactor editor integration to be more asynchronous and flexible


sergiou87:
Context: https://github.com/desktop/desktop/pull/14664#issuecomment-1134820723

As of today, GitHub Desktop builds the list of [compatible](https://github.com/desktop/desktop/blob/b5c8511d3fb7e21f05500bc422e4363587152811/app/src/lib/editors/win32.ts#L137-L138) editors right when the app starts. This is ok as long as that task is quick, but some integrating some editors might require the use of external processes to find out their location, as can be shown in #14664, which would slow the app start down.

Another thing one can realize from #14664 is how using Registry keys to detect editors restricts us, since in that PR we could directly use `vswhere` to determine the location of the executable to open, instead of using it to determine the Registry keys where we can find that location.

With all that in mind, we could do the following refactors:
1. When the app starts, only have a lightweight list with the compatible editors, and defer checking which ones are actually available until they're needed. 
2. While keeping the Registry approach as the default one, provide more flexible ways to look for editors if needed or more convenient.

herboren:
There are multiple editors, give people a list box of programs to choose from that might be present on their system and have the GUI automatically detect the process they are using on their system, then update the GitHub UI accordingly based on running process. Then just restart the GitHub interface for the visual studio community to appear as an option after the entry has been updated inside a config.

**Edit:**

Others mentioning vswhere:
```
PS C:\Users\Daddy\Documents\Library\PackageCache\com.unity.ide.visualstudio@2.0.3\Editor\VSWhere> .\vswhere.exe | findstr 'productPath:'
productPath: C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\devenv.exe
```

Seems to do the trick. I think this would greatly help and save the team a ton of programming. Just extract the output of vswhere.exe. Althought, it doesn't target registry location

AraHaan:
Also cant vswhere be used to locate VSCode? If so, that would be an option to remove checking the registry entirely.

AraHaan:
Oh look it looks like vswhere is ok to be bundled with github desktop as well 🥳.
<img width="849" alt="image" src="https://user-images.githubusercontent.com/15173749/169879350-71668fde-d9db-445c-a8bf-fcf1e089a68a.png">

tsvetilian-ty:
There are several types of issues that I've faced myself or seen another dev face with the current setup:

[...]

Issue #7033, opened 2019-03-07
Desktop doesn't cleanly integrate with my editor (user feedback desired)


billygriffin:
**Please describe the problem you think should be solved**
This is an exploratory issue trying to understand the problem a bit better.

If a GitHub Desktop user doesn't use [one of the editors Desktop detects](https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md), many of the workflows in the app aren't nearly as clean. This used to be less of an issue as we didn't closely tie very many workflows in Desktop to a user's editor, but as we've learned from our users in many cases a tighter relationship between the two tools makes for a much more streamlined workflow.

There are a few issues where this has manifested:
Selected editor doesn't save when a previously selected editor is uninstalled: #7463 
Manual editor setup/detection in preferences #6114 
Merge conflicts flow when no editor is detected #6386 
Different versions of Atom aren't properly detected #6324 
After installing an editor (VS Code), Desktop didn't immediately detect it #7024 
Desire for custom defined editor integration #4465 and #5083

We also know that a few of the most widely used editors including Visual Studio and [XCode](https://github.com/desktop/desktop/issues/4825) are not detected by Desktop. The reason comes down to the inability of many of the editors we don't support to open a directory in the editor (see https://github.com/desktop/desktop/pull/2039 for example, where we made the decision not to support Visual Studio because it wasn't very clean to where we'd direct the user). 

**Impact**

This started out as a convenience to be able to quickly open any file in your editor as you're looking at the diff in Desktop, but the merge conflict resolution process expanded this, and with rebasing and stashing soon to be a part of the product and both having the potential to also introduce conflicts, this will likely be increasingly affecting more and more users.

From our metrics, we've learned that more than 25% of users have no editor detected currently, which is much larger than I expected.

It's unclear to me given the huge % of people that don't have an editor detected by Desktop *why* we haven't heard about this as a larger pain point. If you have thoughts about that, I'd be super interested to hear more.

**Potential next steps**

There seem to be two areas of potential focus:
1. Improving the "no editor detected" experience to either onramp folks to editors that we do detect, or more likely improve the usability of the app in cases where they use an editor we don't detect (for example: #6386).
2. Broaden the scope of editors that are detected. If we're going to improve this process, it seems like we need to do the work up front to segregate the editors that are able to open a directory from those that are not, and have a reasonable design in place for how the handoff interactions will work when the user's primary editor is unable to open a directory.

[...]

Issue #5083, opened 2018-07-03
Allow adding custom editor commands as "External Editor" option


kbauer:
**Is your feature request related to a problem? Please describe.**
Editor-specific implementation like https://github.com/desktop/desktop/issues/4785 aside, users may have specific expecations of how their favorite editor is invoked. Their needs can never be fulfilled by adding explicit support for editors.

In the case of emacs for instance,
- A user with a lightweight Emacs configuration will prefer to use the ``emacs`` executable, or ``runemacs.exe`` on Windows.
- An advanced Emacs user with a heavy configuration, causing the startup process to take a significant amount of time, will likely prefer some invocation of ``emacsclient``, but depending on operating system and preferences the details may vary, e.g.
    - ``emacsclient -a emacs <dirname>`` to handle cases when no emacs session is running a server yet.
    - ``emacsclient -a "emacs -Q" <dirname>`` if the user prefers quick startup over complete startup, if no emacs server is running yet.
    - ``<some wrapper script> <dirname>`` if the user has a more complicated setup.

**Describe the solution you'd like**
Adding custom command strings as *External Editor* option. In the command string, a pattern like ``%d`` should be replaced by the full directory path of the repository.

**Describe alternatives you've considered**
None really. Switching editors is not viable, and only a customizable command can fit my usage.

**Teachability, Documentation, Adoption, Migration Strategy**
![A mockup says more than a thoudsand words.](https://user-images.githubusercontent.com/7498468/42209879-151b7e88-7eb0-11e8-9753-68cdfcfae640.png)

shiftkey:
@kbauer thanks for opening that request. I was looking for more data points on how others would like to tweak their editor config - this is some helpful context.

I'll also cross-reference #4465 which was a similar request for a different editor and has some other discussion.

steveward:
https://github.com/desktop/desktop/issues/15532 was a request for the ability to launch the default editor to a specific line in a file.

jonlepage:
on my editor i use this to open script at line in vscode 
```ts
	public gotoVscode(searchString: string, path: string) {
		const FS = require('fs') as typeof import('fs');
		const fullpath = `C:\\Users\\jonle\\Documents\\DEV\\projets\\ANFTECS\\${path}`;
		const fileContent = FS.readFileSync(fullpath, 'utf8');
		const line = searchString ? findLineof(fileContent, searchString) : 0
		const { exec } = require("child_process") as typeof import('child_process');
[...]

Issue #4465, opened 2018-04-18
experiment with supporting advanced editor flow through config


Shinrai:
## Description
Portable versions of Atom are not supported. Even though with a little bit of work it works just fine.

## Version
* GitHub Desktop: 1.1.1
* Operating system: Windows 10 Pro x64

## Steps to Reproduce
N/A

### Expected Behavior
The ability to edit the path for the atom editor.

### Actual Behavior
If it's not a default install it won't show the atom editor

## Additional Information
I've written what I did to get it to work. But seems like a huge hassle compared to simply giving the user the ability to change the path for atom.
See: https://gist.github.com/Shinrai/27500b67563f7bc1d6968ba8e73cc59b

shiftkey:
@Shinrai thanks for the feedback.

We don't currently support pointing to arbitrary locations for an editor, and there's some more context about this that I've added to an earlier thread in https://github.com/desktop/desktop/issues/2487#issuecomment-347375512

shiftkey:
Marking this as `help wanted` because I'd like someone from the community to explore this further. 

The component for configuring your editor is in [`app/src/ui/preferences/advanced.tsx`](https://github.com/desktop/desktop/blob/master/app/src/ui/preferences/advanced.tsx), and I also have some questions that need to be thought about as part of tackling this:

 - this dialog currently works by detecting installed versions of the tools in the registry - what does the UX look like for adding a portable version of Atom?
 - Desktop currently [throws an error](https://github.com/desktop/desktop/blob/21da8674bbf2bd5348c611f48a3c6203564066a1/app/src/lib/editors/launch.ts#L18-L24) when it can't find the editor it's expecting - is that a suitable experience for portable tools that might not be present when using Desktop?
 - Desktop [caches the editors it finds](https://github.com/desktop/desktop/blob/21da8674bbf2bd5348c611f48a3c6203564066a1/app/src/lib/editors/lookup.ts#L13-L18) for performance reasons the first time the list is needed - a portable tool may become available later, meaning it won't exist in this list until restarting Desktop. What should we do about that?

Shinrai:
@shiftkey
Yes I’ve read that entire issue.

As to the three questions you mentioned. Let’s step back and realize for the most part the only people using these tools are developers. Granted you want to make it as easy as possible for newcomers. But some fine grain control would also be nice for advanced users.

[...]

PR #14843, opened 2022-06-20
Improved handling of external editors

tsvetilian-ty:
## Description
The PR introduces improvements in the handling of external editors. The implementation is limited by a feature flag (`enableCustomExternalEditor`), and it's within the scope of development features.

With these improvements, GitHub Desktop will be able to: 
- handle previously unsupported editors and IDEs (unstable registry key or not using the registry at all)
- handle portable versions of editors
- have same-day support for new editors

### Issues with feedback regarding external editor
- https://github.com/desktop/desktop/issues/7033
- https://github.com/desktop/desktop/issues/14666
- https://github.com/desktop/desktop/issues/4465
- https://github.com/desktop/desktop/issues/5083

### Performance
The implementation doesn't require any additional computation for the detection of editors. It only uses local storage to store a boolean with information if a custom external editor is in use and the custom external editor entry as `FoundEditor` object.

### UI/UX changes
The PR brings three new fields in `Options/Preferences > Integrations`

- Checkbox `Use external editor of your choice` - if checked, shows the following two fields
- `Custom External Editor` - path to custom external editor
- `Launch arguments` - arguments used when launching the custom editor

> The text `%REPO_PATH% - reference to the current repository folder path or a file path.`  appears if `%REPO_PATH%` is not present in the launch arguments.

### Possible feature improvements
- The user can specify different arguments based on the type of path (file or directory)

### Notes
- Currently, there are no updates to the documentation in this PR, but as the discussion progress and the fields are confirmed, I will add it!
- Should we send a boolean in the usage report with information if the user's external editor is custom or one of the defaults?

###  Eclipse integration
eclipse-example

###  Visual Studio integration
visual-studio-example

### VS Code with custom arguments
vscode-arg-example

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -47,6 +47,7 @@ import {
   MultiCommitOperationStep,
 } from '../models/multi-commit-operation'
 import { IChangesetData } from './git'
+import { FoundEditor } from './editors/shared'
 
 export enum SelectionType {
   Repository,
@@ -202,6 +203,12 @@ export interface IAppState {
 
   /** The external editor to use when opening repositories */
   readonly selectedExternalEditor: string | null
+
+  /** Wheter or not the current editor is executable picked by the user */
+  readonly useExternalCustomEditor: boolean
+
+  /** The custom editor picked by the user */
+  readonly customExternalEditor: FoundEditor | null
 
   /** Whether or not the app should use Windows' OpenSSH client */
   readonly useWindowsOpenSSH: boolean
--- a/app/src/lib/editors/found-editor.ts
+++ b/app/src/lib/editors/found-editor.ts
@@ -2,4 +2,5 @@ export interface IFoundEditor<T> {
   readonly editor: T
   readonly path: string
   readonly usesShell?: boolean
+  readonly launchArgs?: string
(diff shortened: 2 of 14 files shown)


bycruz:
Amazing work. This would be great for me considering github desktop is not detecting my install of VSCodium..

Hope this can be reviewed soon

Nico-Connor-Personal:
Amazing work!!

Using [VsWhere](https://github.com/microsoft/vswhere), on a separate PR, I could add a `click here to locate Visual Studio` to this page instead of manually having to find the executable:

image

I just can't think of a way to do that in an elegant way from a design perspective.

tidy-dev:
@tsvetilian-ty Thank you for wanting to contribute. Sorry for the time it took to get back to you on this. This is a really big change and feature with many concerns and we really need to think through it thoroughly. Unfortunately we can't commit to a time frame right now. Instead of letting this sit open, we are going to close it and as soon as we are able to pick this up again, we will open a new PR or revisit this one.

For future references and when we pick this up again, some of the concerns we have around this feature are:
1. Security implications on string injection on the launch arguments
2. Support of opening a folder versus file
3. Optimal design/user interface

PR #14847, opened 2022-06-21
Unreachable commits info message

tidy-dev:
## Description
This PR shows an information message when the diff does not include all the commits in the users selection. When the user hovers over the message or the count of commits in the diff, they will see them highlighted. Additionally, the commit summary is streamlined to not show authors, commit ref, or tags as they become unwieldy with multiple commits.

Next PR:
- On clicking the count of commits the diff or the count unreachable: User will be presented with a dialog showing the respective commits of each kind and maybe a short explanation.

### Screenshots
https://user-images.githubusercontent.com/75402236/176023574-2a06c57e-6d5c-4a0a-9a80-dfe6499b8c08.mov

## Release notes
Notes: no-notes


app/src/ui/lib/list/list.tsx
```diff
@@ -225,6 +225,8 @@ interface IListProps {
    * where to scroll do on rendering of the list.
    */
   readonly setScrollTop?: number
+
+  readonly forceGridUpdate?: boolean
```
tidy-dev: This feels hacky.. Other thought was to emulate the grid.forceUpdate pattern and add a public method of forceUpdate to the list component and do a List ref and List.forceUpdate() up a level. Or... open to other ideas. :)

sergiou87: I think we should try to use the `invalidationProps` prop to simulate this if we can. Just add a `shasNotInDiffHighlighted` prop to the list itself, and then pass it to the `invalidationProps` so that when that prop changes it will update the grid too. Did you try that? Maybe I'm missing something 😳

tidy-dev: Nope - I just didn't know what the invalidationProps was. :D I will give that try

tidy-dev: Works like a charm.

sergiou87: Yay! 🎉

app/src/ui/lib/list/list.tsx
```diff
@@ -77,6 +77,11 @@ interface IListProps {
    */
   readonly selectedRows: ReadonlyArray<number>
 
+  /**
+   * Used to attach special classes to specific rows
+   */
+  readonly rowCustomClassNameMap?: Map<string, ReadonlyArray<number>>
```
tidy-dev, line 83: I went for a flexible approach here (seemed to make more sense  to me with the prev iteration).. but this could be simpler if we just have a `highlightedRows` array much like selectedRows. Any preference?

tidy-dev: (I see pros to both, flexible rowCustomClass opens the door if we every have any list where specific rows need custom styles that isn't highlighting; but doing highlighted rows is simpler and we can set a reusable precedence for how rows should be styled to highlight them)

sergiou87: While I don't have any specific issues with this approach, was there any reason for not applying the `highlighted` in `renderCommit` directly? (in the `CommitList` component) 🤔

app/src/ui/history/commit-summary.tsx
```diff
@@ -316,17 +339,145 @@ export class CommitSummary extends React.Component<
...
+
+    return (
+      <div
+        className="commit-unreachable-info"
+        onMouseOver={this.onHighlightShasNotInDiff}
+        onMouseOut={this.onRemoveHighlightOfShas}
+      >
+        <Octicon symbol={OcticonSymbol.info} />
+        <LinkButton onClick={this.showUnreachableCommits}>
+          {excludedCommitsCount} unreachable commits
```
sergiou87: We should handle plural vs singular here right?
image

app/src/ui/history/commit-summary.tsx
```diff
@@ -316,17 +339,145 @@ export class CommitSummary extends React.Component<
...
+    return (
+      <div className={summaryClassNames}>
+        Showing changes from{' '}
+        {commitsNotInDiff > 0 ? (
+          <LinkButton
+            onMouseOver={this.onHighlightShasInDiff}
+            onMouseOut={this.onRemoveHighlightOfShas}
+            onClick={this.showUnreachableCommits}
+          >
+            {numInDiff} commits
```
sergiou87: This specific case (but not the one below) could also be plural vs singular:
2022-06-30 at 11 28

app/styles/ui/history/_commit-list.scss
```diff
@@ -52,6 +52,16 @@
...
+  &.has-highlighted-commits {
+    .list-item {
+      transition: opacity 0.6s;
+
+      &:not(.highlighted) {
+        opacity: 30%;
+      }
+    }
+  }
+
```
sergiou87: I personally think we should also make a faster transition when nothing is highlighted:

https://user-images.githubusercontent.com/1083228/176643427-0e46afd4-dbaf-4191-b3d1-7da258dc9af8.mov

These are the changes I applied:
```suggestion
  &.has-highlighted-commits {
    .list-item {
      transition: opacity 0.5s;

      &:not(.highlighted) {
        opacity: 30%;
      }
    }
  }

  &:not(.has-highlighted-commits) {
    .list-item {
      transition: opacity 0.25s;
    }
  }

```

app/src/ui/history/commit-list.tsx
```diff
@@ -350,19 +354,42 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
-        <div className="panel blankslate">{this.props.emptyListMessage}</div>
-      )
+    const { commitSHAs, selectedSHAs, shasToHighlight, emptyListMessage } =
+      this.props
+    if (commitSHAs.length === 0) {
+      return <div className="panel blankslate">{emptyListMessage}</div>
     }
 
+    const classes = classNames({
+      'has-highlighted-commits': shasToHighlight.length > 1,
```
sergiou87: I ran into a case where only 1 commit was reachable, and this was preventing it from being highlighted:
```suggestion
      'has-highlighted-commits': shasToHighlight.length > 0,
```

sergiou87 (changes requested):
Looks great overall!! And I love how the animations feel for highlighting commits 😍

app/src/ui/history/commit-summary.tsx
tidy-dev: Yes. 🤦

tidy-dev:
> was there any reason for not applying the highlighted in renderCommit directly? (in the CommitList component

That was where I originally had it, but then the opacity only applies to the words/inner component and not the background color which is at the list level.

sergiou87 (approved):
Thank you for those changes!! 🙇 :shipit:

PR #14871, opened 2022-06-25
skip notifications setup for Linux build

shiftkey:
## Description

I haven't had the time to dig into what an equivalent "advanced notifications" story would look like on Linux, so for now I've found the two places where this appears in the app and skip them for the Linux build.

Also fixed a typo in one of the error messages

### Screenshots

This is what the advanced settings page looks like when running the Linux build:

## Release notes

Notes: no-notes


app/src/ui/preferences/advanced.tsx
```diff
@@ -200,6 +200,11 @@ export class Advanced extends React.Component<
   }
 
   private renderNotificationsSettings() {
+    if (__LINUX__) {
+      // notifications not currently supported
+      return null
+    }
+
```
sergiou87: Thank you for this! But I think this specific function should still be rendered, no? Notifications using `desktop-notifications` (which is what you disabled in `initializeDesktopNotifications`) shouldn't work, but I'd hope that notifications using HTML5 API should still work 🤔 

An easy way to test this is using Help -> Show Notification in a dev/test build

shiftkey: > An easy way to test this is using Help -> Show Notification in a dev/test build

Let me test that and report back

shiftkey: Thanks for the review here @sergiou87. This does work as you expected on Linux so I rebased the first commit to ensure the setting is not hidden for Linux builds.

sergiou87: Yay!! 🎉  Thanks for checking 🙇

sergiou87 (approved):
✨ :shipit: ✨

PR #14883, opened 2022-06-29
Allow bypassing staggered releases

sergiou87:
## Description

This PR makes a small change to allow users to bypass the (upcoming) staggered updates and get an update right away (unless they're completely disabled). The purpose of this will be mainly internal testing but also to help specific users to get an update if we're certain it will help them.

They way this works is by pressing <kbd>Alt</kbd> (or <kbd>Option</kbd> on a Mac keyboard) when clicking `Check for Updates` in the About dialog.

As macOS used to do in the Displays settings dialog, I've made the updates button to change when the <kbd>Alt</kbd> button is pressed, showing `Check for Updates (Forced)` instead (I'm open to suggestions 😂 about the wording)

### Screenshots

https://user-images.githubusercontent.com/1083228/176401369-70ab31ea-565a-4d72-aae8-e40b5e01ca82.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/lib/update-store.ts
+++ b/app/src/ui/lib/update-store.ts
@@ -150,10 +150,13 @@ class UpdateStore {
   /**
    * Check for updates.
    *
-   * @param inBackground - Are we checking for updates in the background, or was
+   * @param inBackground  - Are we checking for updates in the background, or was
    *                       this check user-initiated?
+   * @param skipGuidCheck - If true, don't check the GUID. If true, this will
+   *                       effectively disable the staggered releases system and
+   *                       attempt to retrieve the latest available deployment.
    */
-  public async checkForUpdates(inBackground: boolean, staggered: boolean) {
+  public async checkForUpdates(inBackground: boolean, skipGuidCheck: boolean) {
     // An update has been downloaded and the app is waiting to be restarted.
     // Checking for updates again may result in the running app being nuked
     // when it finds a subsequent update.
@@ -161,7 +164,7 @@ class UpdateStore {
       return
     }
 
-    const updatesUrl = await this.getUpdatesUrl(staggered)
+    const updatesUrl = await this.getUpdatesUrl(skipGuidCheck)
 
     if (updatesUrl === null) {
       return
@@ -176,7 +179,7 @@ class UpdateStore {
     }
   }
 
-  private async getUpdatesUrl(staggered: boolean) {
+  private async getUpdatesUrl(skipGuidCheck: boolean) {
     let url = null
 
     try {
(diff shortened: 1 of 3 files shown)


app/src/ui/lib/update-store.ts
```diff
@@ -189,6 +189,11 @@ class UpdateStore {
     // Send the GUID to the update server for staggered release support
     url.searchParams.set('guid', await getRendererGUID())
 
+    if (!staggered) {
+      // If omitted, staggered releases will be enabled by default
+      url.searchParams.set('staggered', '0')
```
sergiou87: Not a fan of using parameters that by default, when omitted, are `true`, but I also didn't want to use negative logic (like `nonStaggered=1`)

In any case, open to suggestions 😄

tidy-dev: What about `forced`, `immediate`, `bypassStagger`? imho having parameter default true is more confusing than even having `nonStaggered`.  (in fact, when reviewing the central pr, I was a little confused by it)

tidy-dev: or `skipGuidCheck` ... having parameter to skip things is a pattern I have seen elsewhere

sergiou87: `forced` was my first option but it's like… "what's it forcing exactly?" (which is also why the `(Forced)` suffix in the button doesn't convince me 😂 )

And verb+`stagger` didn't sound very well (grammatically) in my head 😂 `skipGuidCheck` could be closer to what we want, but might be confusing since there is a special case where no GUID is sent.

Maybe I shouldn't overcomplicate things and just use `forced` 🤣

tidy-dev: Yeah. I think (Forced) on the button has context enough that it doesn't feel as funny to me and can't think user friendly/simple alternative. But.. .passing just the word "forced" as the parameter doesn't seem to me like it has enough context.. I lean more towards `skipGuidCheck` for the param name. The special case doesn't bother me because if there isn't a guid it makes sense that skipGuidCheck doesn't apply in that special scenario.

sergiou87: Sounds good!

niik: I think your initial inclination about the word choice here is worth listening to. Forced is a word with mostly negative connotations. It's hard to find a drop in replacement so I'm wondering if we instead could change the button label entirely. Perhaps to something like "Ensure latest version" coupled with a tooltip on hover which explains what's going to happen and how it differs from the regular button?

sergiou87: I like that! 🤩 I can't think of a better wording for the button… maybe @tidy-dev has some idea? (or other feedback about your suggestion)

tidy-dev: I like "Ensure Latest Version" -> agree that is much more user friendly then "forced". The tooltip is great too.

tidy-dev (approved):
Works great!

Safar25tory (commented):
[CodeQL](https://github.com/SBoudrias/Inquirer.js/actions/workflows/codeql-analysis.yml)

Safar25tory (commented):
[CodeQL](https://github.com/SBoudrias/Inquirer.js/actions/workflows/codeql-analysis.yml)

Safar25tory (commented):
[CodeQL](https://github.com/SBoudrias/Inquirer.js/actions/workflows/codeql-analysis.yml)

Safar25tory (commented):
[CodeQL](https://github.com/SBoudrias/Inquirer.js/actions/workflows/codeql-analysis.yml)

Issue #14676, opened 2022-05-25
GitHub Desktop doesn't support repositories from SSH servers with password


ndpetwar:
### The problem

My git shell is working well fetching ,pushing ,pulling and doing all other required operations. I am also able to do all operations using eclipse->team in my repository but GitHub desktop showing authentication error. 
these are my logs
2022-05-25T06:16:32.822Z - info: [ui] [RepositoryIndicatorUpdater]: Refreshing sidebar indicators for 4 repositories took 2.4s of which 0.0s paused, total 2.4s
2022-05-25T06:23:06.835Z - error: [ui] `git -c credential.helper= fetch --progress --prune origin` exited with an unexpected code: 128.
stderr:
Permission denied, please try again.
Permission denied, please try again.
git@10.64.28.130: Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

(The error was parsed as 2: Authentication failed. Some common reasons include:

- You are not logged in to your account: see File > Options.
- You may need to log out and log back in to refresh your token.
- You do not have permission to access this repository.
- The repository is archived on GitHub. Check the repository settings to confirm you are still permitted to push commits.
- If you use SSH authentication, check that your key is added to the ssh-agent and associated with your account.
- If you use SSH authentication, ensure the host key verification passes for your repository hosting service.
- If you used username / password authentication, you might need to use a Personal Access Token instead of your account password. Check the documentation of your repository hosting service.)

### Release version

3.0.0

### Operating system

windows 10

### Log files

[2022-05-24.desktop.production.log](https://github.com/desktop/desktop/files/8768481/2022-05-24.desktop.production.log)

sergiou87:
@ndpetwar I assume this is not a repository hosted on GitHub and it was cloned with SSH, is that right? If so, have you added your SSH key to ssh-agent? Take a look at this docs: https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#adding-your-ssh-key-to-the-ssh-agent

ndpetwar:
yes sir i have already added ssh key to ssh agent and thats why able to connect to the server repo via git bash  but not with github desktop @sergiou87

ndpetwar:
@sergiou87 take a look at the command line results
D:\GSTM_PROJ_NEW\utility-project>git config --list
[...]

PR #14884, opened 2022-06-29
Remove label link prop from TextBox

niik:
## Description

This was introduced in https://github.com/desktop/desktop/pull/981 and is only used in one location, 2FA input. As 2FA input in app is exceedingly rare these days with mandatory web flow login we can remove it and at the same time solve a layout issue with very long labels which would previously overflow (see https://github.com/desktop/desktop/issues/14676#issuecomment-1168828321).

## Release notes

Notes: no-notes


sergiou87 (approved):
Changes look good and I tested it and works as expected 🎉 

Thanks for taking care of this 🙇‍♂️ :shipit:

Patientdame (commented):
cancel

PR #14889, opened 2022-06-29
Unreachable commits dialog

tidy-dev:
This is based off of #14847

## Description
This pr adds a dialog to hopefully clarify which commits/changes are present in the history diff of multiple commits where not all changes from commits will be present.

### Screenshots

https://user-images.githubusercontent.com/75402236/176927178-d7a18581-79c9-4062-a1a6-91a3ee30c9d9.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -379,9 +379,7 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
     if (commitSHAs.length === 0) {
       return (
         <div className="panel blankslate">
-          {emptyListMessage === undefined
-            ? 'No commits to list'
-            : emptyListMessage}
+          {emptyListMessage ?? 'No commits to list'}
         </div>
       )
     }
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -57,7 +57,7 @@ interface ICommitListProps {
   readonly onResetToCommit?: (commit: Commit) => void
 
   /** Callback to fire to revert a given commit in the current repository */
-  readonly onRevertCommit?: ((commit: Commit) => void) | undefined
+  readonly onRevertCommit?: (commit: Commit) => void
 
   readonly onAmendCommit?: (commit: Commit, isLocalCommit: boolean) => void
 
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -51,7 +51,7 @@ interface ICommitListProps {
   readonly onScroll?: (start: number, end: number) => void
 
   /** Callback to fire to undo a given commit in the current repository */
-  readonly onUndoCommit?: ((commit: Commit) => void) | undefined
+  readonly onUndoCommit?: (commit: Commit) => void
 
   /** Callback to fire to reset to a given commit in the current repository */
   readonly onResetToCommit?: (commit: Commit) => void
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -24,13 +24,13 @@ interface ICommitListProps {
   readonly selectedSHAs: ReadonlyArray<string>
 
   /** Whether or not commits in this list can be undone. */
-  readonly canUndoCommits: boolean
+  readonly canUndoCommits?: boolean
 
   /** Whether or not commits in this list can be amended. */
-  readonly canAmendCommits: boolean
+  readonly canAmendCommits?: boolean
 
   /** Whether or the user can reset to commits in this list. */
-  readonly canResetToCommits: boolean
+  readonly canResetToCommits?: boolean
 
   /** The emoji lookup to render images inline */
   readonly emoji: Map<string, string>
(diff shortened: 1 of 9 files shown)


tsvetilian-ty:
@tidy-dev the highlighting of the unreachable commits on hover is an excellent idea🔝 As a suggestion - it might be more precise if the title in the dialog is something like `Commits reachability`, considering that the dialog contains both unreachable and reachable tabs 🤔

app/src/ui/history/commit-list.tsx
```diff
@@ -39,42 +39,42 @@ interface ICommitListProps {
...
     isContiguous: boolean
   ) => void
 
   /** Callback that fires when a scroll event has occurred */
-  readonly onScroll: (start: number, end: number) => void
+  readonly onScroll?: (start: number, end: number) => void
 
   /** Callback to fire to undo a given commit in the current repository */
-  readonly onUndoCommit: ((commit: Commit) => void) | undefined
+  readonly onUndoCommit?: ((commit: Commit) => void) | undefined
```
sergiou87: I think we can remove `| undefined`, no?
```suggestion
  readonly onUndoCommit?: (commit: Commit) => void
```

app/src/ui/history/commit-list.tsx
```diff
@@ -39,42 +39,42 @@ interface ICommitListProps {
...
 
   /** Callback to fire to reset to a given commit in the current repository */
-  readonly onResetToCommit: (commit: Commit) => void
+  readonly onResetToCommit?: (commit: Commit) => void
 
   /** Callback to fire to revert a given commit in the current repository */
-  readonly onRevertCommit: ((commit: Commit) => void) | undefined
+  readonly onRevertCommit?: ((commit: Commit) => void) | undefined
```
sergiou87: I think we can remove `| undefined`, no?
```suggestion
  readonly onRevertCommit?: (commit: Commit) => void
```

app/src/ui/history/commit-list.tsx
```diff
@@ -189,9 +189,11 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
...
         gitHubRepository={this.props.gitHubRepository}
         isLocal={isLocal}
-        canBeUndone={this.props.canUndoCommits && isLocal && row === 0}
-        canBeAmended={this.props.canAmendCommits && row === 0}
-        canBeResetTo={this.props.canResetToCommits && isResettableCommit}
+        canBeUndone={this.props.canUndoCommits === true && isLocal && row === 0}
+        canBeAmended={this.props.canAmendCommits === true && row === 0}
+        canBeResetTo={
+          this.props.canResetToCommits === true && isResettableCommit
+        }
```
sergiou87, line 196: Do we need these changes? We avoided truthy/falsy comparisons for strings or numbers, but not for booleans IIRC (at least that's what we discussed a few months ago).

app/src/ui/history/commit-list.tsx
```diff
@@ -377,11 +377,18 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
     const { commitSHAs, selectedSHAs, shasToHighlight, emptyListMessage } =
       this.props
     if (commitSHAs.length === 0) {
-      return <div className="panel blankslate">{emptyListMessage}</div>
+      return (
+        <div className="panel blankslate">
+          {emptyListMessage === undefined
+            ? 'No commits to list'
+            : emptyListMessage}
```
sergiou87: Maybe we can use this instead? (Warning: this suggestion might require linting)
```suggestion
          {emptyListMessage ?? 'No commits to list'}
```

app/src/ui/history/commit-list.tsx
```diff
@@ -397,7 +404,7 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
           selectionMode="multi"
           onScroll={this.onScroll}
           insertionDragType={
-            this.props.reorderingEnabled ? DragType.Commit : undefined
+            this.props.reorderingEnabled === true ? DragType.Commit : undefined
```
sergiou87, line 405: Same as before: is this really necessary?

app/src/ui/history/unreachable-commits-dialog.tsx
```diff
@@ -0,0 +1,149 @@
...
+export enum UnreachableCommitsTab {
+  Unreachable,
+  Reachable,
+}
+
+interface IUnreachableCommitsDialogProps {
+  /** The shas of the currently selected commits */
+  readonly selectedShas: ReadonlyArray<string>
+
+  /** The shas of the shas showed in the diff */
```
sergiou87: Nit:
```suggestion
  /** The shas of the commits showed in the diff */
```

sergiou87 (changes requested):
It looks great! ✨ 

Only have a few suggestions

app/src/ui/history/commit-list.tsx
tidy-dev: Yes. Not for booleans if we require the boolean (thus it can only be true or false). But those props became boolean | undefined. I think I was requested before to make it === true so that it is clear that you must say true for this to be true; thus false or undefined both definitely mean false.

app/src/ui/history/commit-list.tsx
tidy-dev: Probably.. I think I just was thinking to make them optional and didn't notice that they were "forced optional" already.  Not sure the reasoning behind these being `| undefined`. I will test it out just to make sure there isn't something I don't understand going on here.

sergiou87 (approved):
Thanks for those changes! ✨

PR #14907, opened 2022-07-05
fix: not able to detech latest yarn version in postinstall

henit-chobisa:
- Fixed issue in findYarnVersion while building from source. 
WhatsApp Image 2022-07-05 at 2 51 18 PM

### DIFF
--- a/script/post-install.ts
+++ b/script/post-install.ts
@@ -21,9 +21,8 @@ function findYarnVersion(callback: (path: string) => void) {
 
     // this ensures the paths returned by glob are sorted alphabetically
     files.sort()
-
     // use the latest version here if multiple are found
-    callback(forceUnwrap('Missing vendored yarn', files.at(-1)))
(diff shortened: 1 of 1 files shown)


niik:
Hey @HENIT0885, thanks for reaching out! It looks like you may be using an older version of Node.js than what's required. Could you please double check your version of node (`node --version`) and make sure it matches what's in [`.node-version`](https://github.com/desktop/desktop/blob/development/.node-version) (i.e 16.13.0). We've begun using [Array.prototype.at](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) in a number of places already and we expect to use it more and more so we're not keen on reverting those changes.

niik:
Thank you for your issue!

We haven't gotten a response to our questions.
With only the information that is currently in the issue, we don't have enough
information to take action. I'm going to close this but don't hesitate to reach
out if you have or find the answers we need, we'll be happy to reopen the
issue.

PR #14919, opened 2022-07-06
Summer cleanup: feature flags

niik:
## Description

It's a slow week and I'm doing some cleanup. This time removing all currently enabled feature flags. This is a pretty mechanical change but I'd appreciate a thorough review just in case I missed a boolean flip somewhere.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -135,7 +135,6 @@ import { ReleaseNote } from '../models/release-notes'
 import { CommitMessageDialog } from './commit-message/commit-message-dialog'
 import { buildAutocompletionProviders } from './autocompletion'
 import { DragType, DropTargetSelector } from '../models/drag-drop'
-import { enableSquashing } from '../lib/feature-flag'
 import { dragAndDropManager } from '../lib/drag-and-drop-manager'
 import { MultiCommitOperation } from './multi-commit-operation/multi-commit-operation'
 import { WarnLocalChangesBeforeUndo } from './undo/warn-local-changes-before-undo'
@@ -2402,7 +2401,7 @@ export class App extends React.Component<IAppProps, IAppState> {
       'commit-being-dragged': isCommitBeingDragged,
       // 'squashing-enabled' is due to feature flagging. If feature flag is
       // removed, we can just delete this line with adjustment to the css file
-      'squashing-enabled': isCommitBeingDragged && enableSquashing(),
+      'squashing-enabled': isCommitBeingDragged,
     })
   }
 
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -2399,9 +2399,6 @@ export class App extends React.Component<IAppProps, IAppState> {
       currentDragElement !== null && currentDragElement.type === DragType.Commit
     return classNames({
       'commit-being-dragged': isCommitBeingDragged,
-      // 'squashing-enabled' is due to feature flagging. If feature flag is
-      // removed, we can just delete this line with adjustment to the css file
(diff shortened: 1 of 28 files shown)


app/src/ui/app.tsx
```diff
@@ -2402,7 +2401,7 @@ export class App extends React.Component<IAppProps, IAppState> {
       'commit-being-dragged': isCommitBeingDragged,
       // 'squashing-enabled' is due to feature flagging. If feature flag is
       // removed, we can just delete this line with adjustment to the css file
-      'squashing-enabled': isCommitBeingDragged && enableSquashing(),
```
tidy-dev: We can remove this line and remove the squashing-enabled class bracket (but leave the sub css rules). This makes it so when you drag commits over a non selected commit the non-selected commit highlights blue to provide visual indication that the user can drop them there. We just didn't want it to do this when squashing was disabled. Technically, the `isCommmitBeingDragged` here didn't need to be here this whole time because this css rule is already under the `commit-being-dragged` class. (sorry for confusion).

app/src/ui/diff/diff-options.tsx
```diff
@@ -131,13 +130,6 @@ export class DiffOptions extends React.Component<
   }
 
   private renderHideWhitespaceChanges() {
-    if (
-      this.props.sourceTab === RepositorySectionTab.Changes &&
```
tidy-dev, line 135: Do we not need the check for sourceTab === Changes?

tidy-dev (changes requested):
Mostly looks good to me. Just a couple of requests.

app/src/ui/diff/diff-options.tsx
sergiou87: I can't tell but given the `&& !enableHideWhitespaceInDiffOption()` part, which was always `false`, it was irrelevant. Just checked the comment of the `enableHideWhitespaceInDiffOption` function and it only affected the "Changes" tab, so I think this is correct!

app/src/ui/app.tsx
sergiou87: Do you mind doing that change you suggest? I'm not entirely sure about the exact change and I don't want to break anything 😂 (and probably worth merging this before Markus comes back!)

sergiou87 (dismissed):
LGTM overall, tested it and works as expected!!

Only had a couple of suggestions but I addressed them myself in [d9ed147](https://github.com/desktop/desktop/pull/14919/commits/d9ed147d07333854755bb9270237207a526fbcdb) and [a6a9a60](https://github.com/desktop/desktop/pull/14919/commits/a6a9a600e82fb5be290f19adc9157da2449e91fd)

sergiou87:
@tidy-dev do you mind taking a look at those two commits I pushed? (can wait until next week, of course 😅 )

app/src/ui/app.tsx
tidy-dev: done

app/src/ui/diff/diff-options.tsx
tidy-dev: Ah ok. That makes sense.

tidy-dev (approved):
✨ Reviewed both Markus's and Sergio's changes. Looks good!

tidy-dev:
@sergiou87 I removed the `squashing-enabled` stuff. Mind reviewing [ that commit](https://github.com/desktop/desktop/pull/14919/commits/b99f2ec45311b880b8727efd96ff2eb0efc87ee1)?

sergiou87 (approved):
Looks great, yep!! :shipit:

PR #14921, opened 2022-07-06
Retrieve changed files and diff stat in one Git call

niik:
🌵 This is based on #14919, please review that first 🌵 

## Description

While removing feature flags in #14919 I stumbled over `getChangedFiles` which performed two `git log` calls, one to get the list of changed files and their status (i.e. added, deleted, renamed, etc) and another to get the number of added and removed lines in said files.

Curiosity got the best of me and I decided to investigate whether we could do with just one git call and it turns out we can. By combining `--raw` with `--numstat` and changing the parser a bit we can get rid of that second call. Once I had figured that out and written a parser I realized that @tidy-dev had already come to the same conclusion as myself and I had written an almost verbatim equal solution as her.

So I decided to combine our two implementations (I picked mine as the base only because it so happened to be a bit more terse but I'm happy to go either way here).

I then found the `getChangedFilesWithinStash` method which was a duplicate implementation of `getChangedFiles` with a TODO note to eventually combine it with `getChangedFiles`. Looking at its implementation it too executed Git twice, one to get the list of stashed tracked files and one to get the list of untracked stashed files (the third parent of the stash). Digging into this a little it turns out we can leverage [`git stash show`](https://github.com/git/git/blob/30cc8d0f147546d4dd77bf497f4dec51e7265bd8/builtin/stash.c#L925-L931) and it's ability to accept diff display options to not have to make that second Git call here either.

Reducing the number of times we shell out to Git has proven to be a successful strategy to improving performance (especially on Windows) and both of these are frequently called upon method (especially getChanges) so I'm real happy about this change.

## Release notes

Notes: no-notes


tidy-dev (dismissed):
✨ Makes sense. ⏩ ⏩

sergiou87 (approved):
LGTM after fixing conflicts with `development` ✅

PR #14930, opened 2022-07-08
Dynamic width menu panes on Windows

### LINKED ISSUE #14868: [usability]the long menu item name is not displayed completely
### The problem

image

### Release version

3.0.2

### Operating system

win10

_No response_

### Log files

_No response_

### Screenshots

_No response_

### Additional context

_No response_

niik:
Closes #14868 

## Description

The app menu on Windows is rendered in-app using the AppMenuBar component which in turn uses the MenuPane component. The MenuPane renders its list items using the `List` component. I picked this because it seemed like a good idea to share all the selection, and keyboard navigation logic but it does come with a decent amount of downsides, chief among them that we can't (easily) do dynamic width menu panes on Windows which leads us to having to hard code the width of each menu. This works fairly well but we frequently forget to update these widths and it falls apart for dynamic menu labels such as "Open in ${editor}".

The `List` component's primary purpose is enabling efficient rendering of large lists using virtualisation. That's not a concern for our app menu so in this PR I'm exploring what it would take to rip out the List component and just render the menus regularly.

So far it seems to be working just fine without much added complexity ~~but it requires a bit more testing which is why I'm opening this as a draft before heading out on summer vacation.~~

## Release notes

Notes: [Improved] Avoid truncating long menu item labels on Windows


app/src/ui/app-menu/menu-list-item.tsx
```diff
@@ -34,6 +34,35 @@ interface IMenuListItemProps {
...
+  ) => void
+
+  /** Called when the user's pointer device clicks on the list item */
+  readonly onClick?: (
+    item: MenuItem,
+    event: React.MouseEvent<HTMLDivElement>
+  ) => void
+
+  /**
+   * Whether the liste item should steal focus when selected. Defaults to
```
sergiou87: Nit
```suggestion
   * Whether the list item should steal focus when selected. Defaults to
```

sergiou87 (dismissed):
Looks good!! I tested it as thoroughly as I could, and the only thing that was different was sub menu rendering, which I think we only have for dev builds (`Show popup`) and renders like this:

image

If this is the only place, I think we're good to go with this PR and, if we wanted, we could fix that in a different PR.

niik:
> the only thing that was different was sub menu rendering, which I think we only have for dev builds (Show popup) and renders like this:

Yeah, I was a bit shocked to see that submenus were implemented at all on Windows, I don't remember doing that. Either way I'd say that this new state is arguably better than the previous one where the expanded sub menu ends up at the top:

image

Accurately positioning the submenu feels like more work than it's worth unless we decide to add sub menus to production menu items.

sergiou87 (approved):
`splelling` 💯 😂 

:shipit: ❗

PR #14971, opened 2022-07-14
Fix support for unsafe repositories

### LINKED ISSUE #14970: Can't add existing repository on file share
### The problem

Where repositories are located on WSL2 and exposed to windows via a mapped network drive.
Existing repositories continue to work, however I can no longer add a new repository

image

on linux

image

github desktop used to ask to make this exception as part of the add dialog, however this is no longer the case in the new version.

image

After adding the exception manually from the command line, then I was able to add the repository to github desktop

related to #14966 (probably)

### Release version

Version 3.0.4 (x64)

### Operating system

Windows 11

sergiou87:
Closes #14970

## Description
With #14958 we introduced a regression around the detection of unsafe repositories. With the upgrade to Git 2.35.4, the string we had to look for changed but we introduced a typo when we introduced it in our regex to detect those unsafe repositories.

This PR fixes that typo and adds a test to make sure this doesn't happen again 😓 

## Release notes

Notes: [Fixed] Surface again Git's warning about unsafe directories and provide a way to trust repositories not owned by the current user

### DIFF
--- a/app/src/lib/git/rev-parse.ts
+++ b/app/src/lib/git/rev-parse.ts
@@ -37,7 +37,7 @@ export async function getRepositoryType(path: string): Promise<RepositoryType> {
     }
 
     const unsafeMatch =
-      /fatal: detected dubious ownership in repository at '(.+)\' /.exec(
+      /fatal: detected dubious ownership in repository at '(.+)'/.exec(
         result.stderr
       )
     if (unsafeMatch) {
--- a/app/test/unit/git/rev-parse-test.ts
+++ b/app/test/unit/git/rev-parse-test.ts
@@ -101,5 +101,13 @@ describe('git/rev-parse', () => {
         kind: 'missing',
       })
     })
+
+    it('returns unsafe for unsafe repository', async () => {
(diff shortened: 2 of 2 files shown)


tsvetilian-ty (approved):
Works like a charm ✨ 

image

sergiou87:
Thank you for checking it! 🙇‍♂️  💖 I'm gonna go ahead and merge it then, since the rest of the team is OOO for the rest of the week.

sergiou87:
Oh, I cannot unless I force the merge. I'm not in such a rush so I'll just wait 😂

tsvetilian-ty:
We can point people to the beta release for now, so no pressure 😃

sergiou87:
Instead I'll be deploying a release to production, since IIRC this problem caused many issues to people when the problem arose the first time 😓

tidy-dev (approved):
✨  Thank you for fixing this and adding the test! Sorry for the bad regex...

sergiou87:
No worries, I didn't spot it either 😖

PR #14988, opened 2022-07-18
Multi-commit-diffing: Catch error when path not in ref

tidy-dev:
## Description
For our syntax highlighter, we grab what the a file's contents in the current commit and the commit before to create appropriate highlighting. For the commit to be diffed, we know if a file was added/untracked, modified, or deleted in that commit. For highlighting, we are only interested in modified and deleted; if it is new, there is nothing to highlight. Thus, if a file is new or untracked, we do not try to look for it in the commit before. This works great when comparing only one commit at time.

However, with multi-commit diffing, given a history of A, B, C and we pick a range of B - C,  we want to compare changes from B and C against the commit A. If a file is added in B and modified in C, when doing a diff of B..C the status will show modified for that file. Thus, our current logic will attempt to find the file contents of that file in A that do not exist -> resulting in a git error. 

This PR returns null if that error is present (which is what the parent method would return if the file status was new or untracked).

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/git/diff.ts
+++ b/app/src/lib/git/diff.ts
@@ -244,7 +244,11 @@ export async function getCommitRangeChangedFiles(
     return getCommitRangeChangedFiles(repository, shas, useNullTreeSHA)
   }
 
-  return parseChangedFilesAndNumStat(result.combinedOutput, oldestCommitRef)
+  return parseChangedFilesAndNumStat(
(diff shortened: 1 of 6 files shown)


sergiou87 (approved):
Looks great and works as expected! :shipit: ✨

PR #15010, opened 2022-07-21
Apple silicon covers more than M1

say25:
Closes #[issue number]

## Description

- With the release of M2 devices, we should make it clear this covers all Apple silicon devices

## Release notes

Notes:


sergiou87 (approved):
That is true! Thank you 😄

Issue #13736, opened 2022-01-24
Improve Options documentation


SetTrend:
After installing *GitHub Desktop* I couldn't fail to notice that there are two input boxes in *Options > Git* that are not getting explained nor are they self-explanatory:

![GitHub Desktop options](https://user-images.githubusercontent.com/9283914/150856636-f6615d53-50d4-46c0-9088-67c5ffbe7a6b.png)

As a user, I wonder: what is supposed to be entered into these two input fields?

Is it, for example:

```
First name

Surname
```

… or …

```
Repository name

Subtitle
```

Please clarify.

### Proposed solution

I propose to add placeholder text to the two text input elements.

Moreover, I suggest do add a documentation web page for each of the Options tabs on GitHub. (Next to [this documentation page](https://docs.github.com/en/desktop/installing-and-configuring-github-desktop/configuring-and-customizing-github-desktop/setting-a-theme-for-github-desktop).)

tidy-dev:
@SetTrend Thank you for the bug report! At first I was unable to reproduce this as I uninstalled it and reinstalled it and during open then app I logged into my github account. Thus, this screen appropriately shows Name = tidy-dev and the second box had the label of Email and was dropdown showing my accounts emails.

Then I realized if you didn't have a name there, it must mean you are not signed in to an account. If you go up to the accounts tab and sign in and then come back to the Git tab, the boxes will be labeled appropriately.

However, I am still labeling this as a bug, as we should have a better UX when person is not logged in (I believe before a recent change, the second input displayed "Email" as the label)

SetTrend:
Thank you for your valuable efforts, @tidy-dev!

Yes, your observation is right. I'm currently preparing a VM image for myself, to be cloned and used with different clients of mine. So I cannot log in yet.

Joshua698:
Yeh

sergiou87:
This was fixed by @andymckay in https://github.com/desktop/desktop/pull/15011 and the fix will be available soon in beta builds 😄 

Thanks for your patience!

PR #15011, opened 2022-07-21
Show email label on the preferences form if email isn't defined

andymckay:
This assists with #13736 but I'm not sure this fixes it, because the documentation is a little unclear.

## Description

The form has a couple of states.

When you are logged into GitHub you can choose a possible email address:

Screen Shot 2022-07-21 at 1 40 02 PM

In this first case, the input text field isn't shown at all, just the drop down 👍🏾 

When you are logged into GitHub and would like to use a different email address:

Screen Shot 2022-07-21 at 1 40 12 PM

In this case the input text field is shown, but there's no need for an `Email` label, its on the field above 👍🏾 

And in the case referenced in the original issue  #13736, where you are not logged into GitHub and no other information (such as no `~/.gitconfig`) is available. As noted in the original issue, you get an odd form with no label above the `Email` field 😢 

I believe this is because we are using a boolean field to determine if we should show the label, but there are three possible states, as mentioned as above. In this last case when no other information is available `this.props.email` is going to be `null`. If that's the case we probably shouldn't take a look inside `accountEmails` to see if it's there, we can assume it's not other.

The result is we see the form label:

Screen Shot 2022-07-21 at 1 53 33 PM

I didn't mark this as fixed because there were other questions about this form and I'm not expert on this flow and this was more just a quick drive by. But hopefully having the label `Email` above the field, makes it a little bit clearer.

I couldn't spot any unit tests for this particular form.

## Release notes

Notes:

### DIFF
--- a/app/src/ui/lib/git-config-user-form.tsx
+++ b/app/src/ui/lib/git-config-user-form.tsx
@@ -47,7 +47,7 @@ export class GitConfigUserForm extends React.Component<
     super(props)
 
     this.state = {
-      emailIsOther: !this.accountEmails.includes(this.props.email),
+      emailIsOther: this.props.email ? !this.accountEmails.includes(this.props.email) : false
     }
   }
 
--- a/app/src/ui/lib/git-config-user-form.tsx
+++ b/app/src/ui/lib/git-config-user-form.tsx
@@ -47,7 +47,9 @@ export class GitConfigUserForm extends React.Component<
     super(props)
 
     this.state = {
-      emailIsOther: this.props.email ? !this.accountEmails.includes(this.props.email) : false
+      emailIsOther:
+        this.accountEmails.length > 0 &&
+        !this.accountEmails.includes(this.props.email),
     }
   }
 
@@ -67,7 +69,9 @@ export class GitConfigUserForm extends React.Component<
     // from the user, to prevent annoying UI glitches.
     if (prevProps.email !== this.props.email && !isEmailInputFocused) {
       this.setState({
-        emailIsOther: this.props.email ? !this.accountEmails.includes(this.props.email) : false,
+        emailIsOther:
+          this.accountEmails.length > 0 &&
+          !this.accountEmails.includes(this.props.email),
       })
     }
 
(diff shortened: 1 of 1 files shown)


app/src/ui/lib/git-config-user-form.tsx
```diff
@@ -47,7 +47,7 @@ export class GitConfigUserForm extends React.Component<
     super(props)
 
     this.state = {
-      emailIsOther: !this.accountEmails.includes(this.props.email),
+      emailIsOther: this.props.email ? !this.accountEmails.includes(this.props.email) : false
```
sergiou87: I think this would be the right check:
```suggestion
      emailIsOther:
        this.accountEmails.length > 0 &&
        !this.accountEmails.includes(this.props.email),
```

app/src/ui/lib/git-config-user-form.tsx
```diff
@@ -67,7 +67,7 @@ export class GitConfigUserForm extends React.Component<
     // from the user, to prevent annoying UI glitches.
     if (prevProps.email !== this.props.email && !isEmailInputFocused) {
       this.setState({
-        emailIsOther: !this.accountEmails.includes(this.props.email),
+        emailIsOther: this.props.email ? !this.accountEmails.includes(this.props.email) : false,
```
sergiou87: Same:
```suggestion
        emailIsOther:
          this.accountEmails.length > 0 &&
          !this.accountEmails.includes(this.props.email),
```

sergiou87 (changes requested):
Thank you for your contribution @andymckay !!

I checked and the problem is `emailIsOther` is meant to indicate when (1) the dropdown is visible and (2) it has `Other` selected. However, if the dialog was shown with no user logged in (meaning there is no list of emails and therefore no dropdown), `emailIsOther` would be initialized to `true`.

The changes I suggested should fix that, by checking there is a list of emails.

andymckay:
Thank you for the suggestions @sergiou87 and taking a look at it. I gave it a quick test locally and those work just fine, so let's go with those. Cheers 🙇🏾

sergiou87 (approved):
Yay! :shipit:

PR #15028, opened 2022-07-25
Add a way of adding licenses to existing projects

### LINKED ISSUE #12222: Adding License e.g. Apache License 2.0 to existing projects!
I would like to Add a License to an Older/Existing Git, I see no way to do that - Only when a new Git is created I get that option!
You seem to have most of the code - Just make it available to append it to an Existing Git!

a2937:
There's some stuff that's not as good as it can be , but it's at least functionality.

Closes #12222
## Description
A tab under repository settings was added that allows repository owners to quickly add licenses to their projects. 
-

### Screenshots
Showing active license

No license found

## Release notes

Notes:

### DIFF
--- a/app/src/ui/repository-settings/licenses.tsx
+++ b/app/src/ui/repository-settings/licenses.tsx
@@ -1,0 +1,123 @@
+import * as React from 'react'
+import { ILicense, getLicenses } from '../add-repository/licenses'
+import { Select } from '../lib/select'
+import { Row } from '../lib/row'
+import { DialogContent } from '../dialog'
+import { LinkButton } from '../lib/link-button'
+import { Ref } from '../lib/ref'
+/** The sentinel value used to indicate no license template has been selected. */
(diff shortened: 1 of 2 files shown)


a2937:
I need to get the methods to be asynchronous and wanted to open this pull request as a draft.

tidy-dev:
Thank you for your contribution @a2937! The GitHub Desktop team has not had a chance to discuss this feature to determine if this is something we want to move forward with. We briefly discussed this PR today in our team sync, and currently, don't believe this feature would have widespread usage, and, if we were to consider it, we would not want it to be implemented as such a prominent feature such as being permanent fixture in the repository settings. 

We're sorry if this feels like a waste of your time 😢 In general, it's better to make your argument in the issue and wait until you get a 👍 from the team to implement it (or we place a help-wanted label on it).

a2937:
On future issues I will try to justify why it should be added. Thank you for your time.

PR #15030, opened 2022-07-25
Pull-request Preview POC WIP

tidy-dev:
## Description
This is a POC for what a Pull-request preview flow could look like in Desktop. 

Things to do something about:
- The view shows changes from all commits and when user selects any of the commits, it shows changes from all of them. Currently looking to make all the commits selected, and disable interaction with the commit list as the simplest solution... but if we made it so user could click through these, should we have something in this view to make it more obvious that you are not in the regular history or compare view/ would a user intuitively know to reselect all to get back to "looking at branch to branch" view?
- Getting here from the compare tool? I originally approached this from the compare tool, by adding a new dropdown to the merge button at the bottom of the compare behind section; but this is a little confusing because it would only make sense if you switched to your default branch and compared the feature branch your want to create the pr for.. usually developers are on their feature branch and then want to move to a pr. But.. we could add a button to bottom of "Ahead" section, so that when on a feature branch, you could compare to development (or other branch) and "Preview the PR" from there -> this would hide the compare view tabs (maybe not super clear you switched to a "new view". Other thoughts was compare tool gets a new tab of "Pull Request Preview". That is this new view.
- Should merge/squash and merge, and rebase be a thing here? Like if you don't need to do a pr - no collaboration, but want to use this simply to see the pr comparison before merging? I currently have merge and squash/merge added, but an issue with it is that in order to do so, we need to checkout the target branch... and if we abort the action, should we return to the pr preview (would require some more state remembering. 

Other functionality to add:
- Currently if you change the compare branch, it resets you to a "compare view", be better if this just changed to pull request view base branch. If so "create a pull request", functionality needs to be updated to take in a target branch instead of assuming repo default branch.

### Screenshots

https://user-images.githubusercontent.com/75402236/180788928-3fb3ee77-bbf8-45c0-a3e2-2fbebe1156a0.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -727,7 +727,7 @@ export interface IPullRequestPreview {
   readonly kind: HistoryTabMode.PullRequestPreview
 
   /** The branch to diff --merge-base against */
-  readonly comparisonBranch: Branch
+  readonly mergeBaseBranch: Branch
 }
 
 export interface ICompareState {
@@ -799,7 +799,7 @@ export interface ICompareToBranch {
 
 export interface IPreviewPullRequest {
   readonly kind: HistoryTabMode.PullRequestPreview
-  readonly branch: Branch
+  readonly mergeBaseBranch: Branch
 }
 
 /**
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -680,6 +680,7 @@ export interface IChangesState {
 export enum HistoryTabMode {
   History = 'History',
   Compare = 'Compare',
+  PullRequestPreview = 'PullRequestPreview',
 }
 
 /**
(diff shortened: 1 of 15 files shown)


billygriffin:
So excited for this!! One piece of feedback. It's not obvious to me when you click into the PR flow that you're viewing the combined diff of all the commits. It feels like it should be a more first class component that is obviously different from the individual commit list, and conveys that this is the entirety of what you're proposing to merge. But I'm just a rando in the wilderness, and I know y'all are thinking about this stuff way harder than me. Hope that's helpful feedback, and thanks for all your great work!! ❤️

tidy-dev:
> It's not obvious to me when you click into the PR flow that you're viewing the combined diff of all the commits. 

Thanks for the feedback! Confirmation of the conclusion we had-> got a third POC in progress. :)

PR #15103, opened 2022-08-12
Remove mention of Node 12

niik:
## Description

It's been a while since we were using Node 12.x.

## Release notes

Notes:


tidy-dev (approved):
✨

PR #15104, opened 2022-08-12
Fix some TS 4.8-related issues

RyanCavanaugh:
Closes #[issue number]

## Description

- Fixes some issues uncovered by TypeScript 4.8

## Release notes

Notes: Depends on https://github.com/desktop/dugite/pull/488


app/src/lib/git/core.ts
```diff
@@ -161,7 +161,7 @@ export async function git(
     // from a terminal or if the system environment variables
     // have TERM set Git won't consider us as a smart terminal.
     // See https://github.com/git/git/blob/a7312d1a2/editor.c#L11-L15
-    opts.env = { TERM: 'dumb', ...combinedEnv } as Object
+    opts.env = { TERM: 'dumb', ...combinedEnv } as object
```
RyanCavanaugh, line 164: Note -- 99% of the time, `object` is the type you want. `Object` is effectively a TS implementation detail that shouldn't be used in user code

app/src/lib/merge.ts
```diff
@@ -1,5 +1,5 @@
 /** Create a copy of an object by merging it with a subset of its properties. */
-export function merge<T, K extends keyof T>(obj: T, subset: Pick<T, K>): T {
+export function merge<T extends {}, K extends keyof T>(obj: T | null | undefined, subset: Pick<T, K>): T {
```
RyanCavanaugh: This fixes some typing issues with both the implementation and callers when `obj` is possibly null or undefined

app/src/lib/stores/helpers/repository-indicator-updater.ts
```diff
@@ -125,7 +125,7 @@ export class RepositoryIndicatorUpdater {
 
   private clearRefreshTimeout() {
     if (this.refreshTimeoutId !== null) {
-      window.clearTimeout()
+      window.clearTimeout(this.refreshTimeoutId)
```
RyanCavanaugh, line 128: Somehow this wasn't a TS error before

niik: Woah! How did I miss this, it's pretty terrible. I feel like we had an issue describing a case when the repository updater was running even though it was turned off in preferences but now I can't find it. Either way, thanks a bunch!

niik:
Waiting on a release of https://github.com/desktop/dugite in order to get CI passing here

niik (approved):
Thanks so much @RyanCavanaugh!

Issue #14197, opened 2022-03-17
Only the latest version of Sublime Text is recognized as available external editor


vbwx:
### Search existing issues

- [X] I have searched the issue tracker and confirmed that I was unable to find existing issues for the problem I am experiencing.

### The problem

I have Sublime Text 3 installed on my system but it's not in the list of external editors available to GitHub Desktop. Out of curiosity, I installed Sublime Text 4, and, voila, it's in the drop-down menu.

I believe the problem can be traced back to a function in `darwin.ts` that tries to find a path for every bundle identifier of a specified editor object but I think it aborts after searching for the first identifier, regardless of the result.

https://github.com/desktop/desktop/blob/a6a324da4bb11bff4e457e4a118d9968fd634345/app/src/lib/editors/darwin.ts#L137-L149

I am not familiar enough with TypeScript (and _promises_ in particular) to actually fix the bug without introducing new ones but I would suggest removing lines 143–145 as a starting point, and changing line 147 to `if (installPath && await pathExists(installPath)) {`.

### Release version

2.9.11

### Operating system

macOS 12.2.1

### Steps to reproduce the behavior

**The following workaround fixes the issue without having to install Sublime Text 4.**

Sublime Text 3 has to be installed.

1. Open Preferences > Integration
    * Sublime Text is not listed in the External Editor drop-down menu
2. Quit GitHub Desktop
3. Open `/Applications/GitHub Desktop.app/Contents/Resources/app/renderer.js` in an editor
4. Search for and delete `'com.sublimetext.4',`
5. Save the file and open GitHub Desktop
6. Open Preferences > Integration
    * Sublime Text should now be listed in the drop-down menu

Essentially, by removing line 45, the app can detect Sublime Text 3.
https://github.com/desktop/desktop/blob/a6a324da4bb11bff4e457e4a118d9968fd634345/app/src/lib/editors/darwin.ts#L42-L49

### Additional context

Multiple issues regarding Sublime Text support have already been submitted, but the exact cause of GitHub Desktop not recognizing this editor has not been found, as far as I can tell.

steveward:
Thanks for the detailed report @vbwx! If you're up for submitting a PR I'm sure someone from the team could assist with any specific questions. We also have a detailed editor integration guide here:

https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md

steveward:
Reported again in https://github.com/desktop/desktop/issues/16039.

steveward:
[...]

PR #15117, opened 2022-08-16
Search for all bundle identifier entries instead of just the first one

vbwx:
This should fix a bug that caused GitHub Desktop to only find the latest version of supported text editors.

The issue has been described in #14197.

## Description
It should be possible to set the default text editor to older versions of editors officially supported by GitHub Desktop, but that did not work when a major version bump also caused the bundle identifier of the text editor to change. (For example, `com.sublimetext.3` and `com.sublimetext.4` are identifiers representing different versions of the same editor.)

## Release notes

Notes: [Fixed] Older versions of Sublime Text and SlickEdit are also recognized as external editors.

### DIFF
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -148,7 +148,7 @@ async function findApplication(
           : Promise.reject(e)
       )
 
-      if (installPath && await pathExists(installPath)) {
+      if (installPath && (await pathExists(installPath))) {
         return installPath
       }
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
Nice catch! LGTM and works as expected, thank you for this fix! 🙇‍♂️🙇‍♂️🙇‍♂️🙇‍♂️

vbwx:
Happy to help!

PR #15128, opened 2022-08-17
Prevent virtualization of diff rows that are part of document selection

### LINKED ISSUE #13701: Only visible code stays selected when selecting a large chunk of code in a diff
### Describe the bug
When trying to select code in a diff and scrolling through it (I wanted to copy a chunk of code that's bigger than the viewport), only the selected code within the viewport stays selected.

### Version & OS
2.9.6 (Windows 10)

1. Open a file with a diff that's big enough to scroll
2. Start mouse selection at the top of the diff
3. Drag down and let the page scroll, or use the scroll wheel
3. Let go and scroll up to the start of the selection

### Expected behavior

I expected all code to be selected from the top of the diff to where I released the mouse button.

### Actual behavior

Only the code within the viewport is selected.

### Screenshots

https://user-images.githubusercontent.com/72231755/150154138-99d43cdd-0483-471b-b79b-4b6794890a79.mp4

niik:
Closes #13701

## Description

The biggest reason why we haven't started experimenting with using the new diff engine (which powers side-by-side diffs) for unified diffs as well is that it doesn't play well with text selection.

The side-by-side diff component is built on a react-virtualized List which will only render into the DOM the lines that are currently visible plus a few extra to make scrolling feel smooth (this is called overscan, we'll get back to that in a bit). The problem with this is if you select outside of those visible lines the DOM rows get removed and the selection collapses to contain the visible nodes.

Aug-17-2022 14-35-59

Previously our thought on how to solve this was by virtualizing the selection as well so that any time the list was scrolled we would update the Selection API and have it appear as if the selection was there all along. Unfortunately that approach proved complex and produced odd artefacts when attempting to modify an active selection (i.e. modify the selection while the user is still dragging).

This PR instead takes a different, hackier, but less complex approach. When we detect that the whole, or parts of the diff is contained within the document selection we keep track of which row the selection "started" from and which row it "ends" on. When we know this we can increase the List overscan to not recycle those rows while scrolling. Effectively killing virtualization for the part of the diff which is selected. For small-ish files this is no biggie but it has performance impacts when for example attempting to select all text in a huge diff.

Judging by the issues we've received about this and my own gut feeling I suspect that the majority use-case here is selecting parts of a diff (or copying all of a smaller sized diff perhaps) that extend past the bounds of the window. Not copying multi-megabyte diffs.

## Release notes

Notes: [Fixed] Text selection is preserved while scrolling diffs in split mode


app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -635,6 +776,31 @@ export class SideBySideDiff extends React.Component<
...
+
+  private onSelectAll = (ev: Event | React.SyntheticEvent<unknown>) => {
+    ev.preventDefault()
+
+    // Expand the overscan to infinity (i.e. render it all) before selecting
+    // all of the diff contents.
+    this.textSelectionStartRow = 0
+    this.textSelectionEndRow = Infinity
+
+    this.virtualListRef.current?.forceUpdate(() => {
```
sergiou87: Not sure why (I spent some time debugging but didn't find the cause), but this `forceUpdate` doesn't result in `overscanIndicesGetter` being invoked, so it doesn't overscan the remaining cells and therefore not all the content is copied 😕

sergiou87 (changes requested):
Looks good overall! I think it all makes sense, however when testing it, I found that `overscanIndicesGetter` wasn't being called on macOS when I pressed Cmd+A to select all (the `forceUpdate` method is called on the list, though 🤔 )

app/src/ui/diff/side-by-side-diff.tsx
niik: Fascinating. So the way I tested this was to hit Cmd+A and then scroll to verify that the selection extended to the end. I now see that it's the scrolling which forces the re-render and therefore the selection. So I'm guessing that the browser auto-extends the selection as we add in lines in-between the focus and anchor points. I'll take a look and see if we can trick it somehow.

niik:
@sergiou87 This is ready for another look. I tossed the idea of overscanning when selecting all as it just destroyed the performance. Instead I try to keep track of whether or not the entire diff is selected and hijack the copy event to fill it with the entire diff instead.

This got a little wonky when hitting Ctrl/Cmd+A outside of the diff since the selection API doesn't let us (easily) figure out what's copy-able and not so I made it such that text selection in the diff is disabled until it (or its descendants) are focused.

sergiou87 (dismissed):
The described changes make sense! I've tested it as thoroughly as I could and couldn't break it. It's nice to see that even when all text is selected, if you scroll, you see the virtualization going on but keeping every new row still selected! 👏 

Awesome work! :shipit:

niik:
@sergiou87 Can I bother you for a re-review after that merge conflict resolution?

sergiou87 (approved):
✨

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -75,6 +77,21 @@ export interface ISelection {
...
+    const rowIndex =
+      row.ariaRowIndex !== null ? parseInt(row.ariaRowIndex, 10) : NaN
+    return isNaN(rowIndex) ? undefined : rowIndex
+  }
+
+  return undefined
+}
+
 interface ISideBySideDiffProps {
   readonly repository: Repository
```
Gdub662, line 96: readonly repository: Repository

PR #15139, opened 2022-08-19
Fix org/repo#12345 links

### LINKED ISSUE #8403: Username/Repository# References In Commits Displayed Incorrectly
## Description

Autolinked references in commit messages of the form `Username/Repository#` are not displayed correctly in the app, but are on GitHub's website.
## Version

* GitHub Desktop: 2.1.3

* Operating system: 10.14.6

## Steps to Reproduce

1. Make a commit which references an issue on another repository, using the form `Username/Repository#`, e.g. "Testing desktop/desktop#9999999"

### Expected Behavior

Should show a clickable link in the repo history to the issue/PR on the **other** repository.

### Actual Behavior

Shows a clickable link in the repo history to the issue/PR on the **same** repository.

thinkyhead:
Closes #8403

## Description
Compose links of the form org/repo#12345 correctly. The link still only appears on the issue number. More disruptive changes would be required to apply the link to the entire `org/repo#12345` string, since `tokenizeGitHubRepository` follows a linear scanning methodology.

### Screenshots
n/a

## Release notes
- Issue numbers preceded by org/project prefix are now linked to the correct destination.

### DIFF
--- a/app/src/lib/text-token-parser.ts
+++ b/app/src/lib/text-token-parser.ts
@@ -162,7 +162,14 @@ export class Tokenizer {
       return null
     }
 
-    const url = `${repository.htmlURL}/issues/${id}`
+    // handle an issue link to another repo
+    let repoUrl = repository.htmlURL;
+    const orgRepo = text.slice(0,index).match(/(\w+\/\w+)$/);
+    if (orgRepo) {
+      repoUrl = `https://github.com/${orgRepo[0]}`
+    }
+
+    const url = `${repoUrl}/issues/${id}`
     this._results.push({ kind: TokenType.Link, text: maybeIssue, url })
     return { nextIndex }
   }


tidy-dev (changes requested):
@thinkyhead Thank you for wanting to improve GitHub Desktop! However, some of our users repositories are not based at `https://github.com` and therefore using the base from repository.htmlURL is important. We recently added `issue-mention-filter.ts` which will be used for doing this kind of filtering on markdown. It has a `createLinkElement` method for accomplishing the same thing while preserving the correct base href. (It also takes into account  user preceded issue refs `tidy-dev/#1234`). Additionally, there is a more thorough `IssueOwnerOrOwnerRepo` regex in that filter that would be more ideal to use for checking if the id is preceded by an owner/repo pattern. If you could make use of that logic, this would be nice improvement.

tidy-dev:
@thinkyhead It has been a while since we have heard from you on this. Thus, we are going to close it. However, feel free to open another PR if you would like to continue this work. Thanks again for wanting to improve GitHub Desktop.

PR #15161, opened 2022-08-24
Enable x64-to-arm64 immediate updates in prod builds

sergiou87:
Closes #14998

## Description
This PR enables in prod the feature flag created for #14998, so that users of prod builds will be prompted to upgrade x64 binaries to arm64 on macOS when running on Apple silicon devices.

## Release notes

Notes: [Improved] On Apple silicon devices running unoptimized builds, auto-update on first run to an optimized build


tidy-dev (approved):
✨

PR #15176, opened 2022-08-26
Improved submodule support

sergiou87:
## Description
This PR improves submodule support on GitHub Desktop by making sure it renders something useful to users instead of the `Subproject commit a901a6a0d95895e34c031c7ae8d3ab6d345a67fd`.

### Screenshots

https://user-images.githubusercontent.com/1083228/186927974-2bea1f17-8c00-4099-95dd-4ecc20370f46.mov

In the video you can see:
1. A new submodule with non-committed changes (`desktop-notifications`): you can partially select it because you can commit its addition to the repo, but not the non-committed changes within it
1. An existing submodule that has both a change in the commit it points to, and non-committed changes within it (`desktop-trampoline`): you can partially select it because you can commit the change in the commit it points to, but not the nont-committed changes within it
1. An existing submodule that only has a change in the commit it points to (`sp-ios-tools`): can be selected and committed normally
1. An existing submodule that only has non-committed changes within it (`test-repo`): it can’t be selected at all because it only has non-committed changes :cry: 

For 1, 2 and 4, you will get a tooltip if you hover over the checkbox, explaining why selection is either disabled or partial

## Release notes

Notes: [Improved] Improve submodule support

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -308,9 +308,9 @@ export class ChangesList extends React.Component<
       isCommitting || rebaseConflictState !== null || isUncommittableSubmodule
 
     const checkboxTooltip = isUncommittableSubmodule
-      ? 'This submodule cannot be checked in because it contains changes have not been committed.'
+      ? 'This submodule change cannot be added to a commit in this repository because it contains changes that have not been committed.'
       : isPartiallyCommittableSubmodule
-      ? 'Only changes that have been committed within the submodule will be checked in in this repository. You need to commit any other modified or untracked changes in the submodule before including them in this repository.'
+      ? 'Only changes that have been committed within the submodule will be added to this repository. You need to commit any other modified or untracked changes in the submodule before including them in this repository.'
       : undefined
 
     return (
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -278,6 +278,18 @@ export class ChangesList extends React.Component<
 
     const file = workingDirectory.files[row]
     const selection = file.selection.getSelectionType()
+    const { submoduleStatus } = file.status
+
+    const isUncommittableSubmodule =
+      submoduleStatus !== null &&
+      file.status.kind === AppFileStatusKind.Modified &&
+      !submoduleStatus.commitChanged
+
+    const isPartiallyCommittableSubmodule =
+      submoduleStatus !== null &&
+      (submoduleStatus.commitChanged ||
+        file.status.kind === AppFileStatusKind.New) &&
+      (submoduleStatus.modifiedChanges || submoduleStatus.untrackedChanges)
 
     const includeAll =
       selection === DiffSelectionType.All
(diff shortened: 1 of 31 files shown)


app/src/ui/changes/changes-list.tsx
```diff
@@ -286,22 +298,31 @@ export class ChangesList extends React.Component<
...
+      : rebaseConflictState !== null
+      ? file.status.kind !== AppFileStatusKind.Untracked
+      : includeAll
+
+    const disableSelection =
+      isCommitting || rebaseConflictState !== null || isUncommittableSubmodule
 
-    const disableSelection = isCommitting || rebaseConflictState !== null
+    const checkboxTooltip = isUncommittableSubmodule
+      ? 'This submodule cannot be checked in because it contains changes have not been committed.'
```
tidy-dev:
```suggestion
      ? 'This submodule cannot be checked in because it contains changes that have not been committed.'
```

Other thoughts are is the terminology "checked in" ideal - can't articulate what it is but it feels off to me - maybe it it is because in the UI we just use the word commit? Maybe this should be "This submodule change cannot be added to a commit in this repository, because..." or "This submodule change cannot be committed, because..."  

Similarly for next tooltip, "Only changes that have been committed within the submodule will be added to this repository" or "committed to this repository"

app/src/ui/lib/tooltipped-commit-sha.tsx
```diff
@@ -0,0 +1,51 @@
...
+interface ITooltippedCommitSHAProps {
+  readonly className?: string
+  /**
+   * Short SHA to display in the tooltipped element, or the tooltipped element
+   * itself.
+   */
+  readonly shortSHA: string | JSX.Element
+
+  /** Long SHA to display within the tooltip. */
+  readonly longSHA: string
```
tidy-dev, line 15: Just a thought since this is a formatting component -> feel free to ignore. We could pass in just longSHA: string | Commit. If is a string, we shorten it here using shortenSha method and if it is a commit we use commit.shortSha and commit.sha.

tidy-dev: (Or make longSHA? and commit? properties and use the defined one :D)

app/src/ui/changes/changes-list.tsx
sergiou87: Yeah I used `checked in` in an attempt to differentiate the actions in the main repo vs the actions in the submodule repo. I think "This submodule change cannot be added to a commit in this repository, because…" should be clear enough 🤔  Thanks for your suggestions!!

app/src/ui/lib/tooltipped-commit-sha.tsx
sergiou87: Well, the problem is the short SHA is either a `string` or a `JSX.Element`, if you need something custom (like we do in the submodule interstitial to use a `Ref` component to wrap the SHA).

tidy-dev: That could be done either as another component extending this or as boolean `asRef` and still keep all the formatting to his component or components.

tidy-dev (approved):
✨  Tested it out (on my mac) and everything worked as expected. Thanks for the tooltip changes and adding the feature flag. 

(Feel free to ignore the tool tip commit sha component suggestions)

app/src/ui/diff/submodule-diff.tsx
```diff
@@ -0,0 +1,166 @@
...
+          {this.renderCommitChangeInfo()}
+          {this.renderSubmodulesChangesInfo()}
+          {this.renderOpenSubmoduleAction()}
+        </div>
+      </div>
+    )
+  }
+
+  private renderSubmoduleInfo() {
+    // TODO: only for GH submodules?
```
tidy-dev, line 55: Just remembered - I meant to ask if this is needed - does it work just fine for non GH submodules?

sergiou87: I only tested this with GH and bitbucket repos and it worked fine, but maybe it doesn't with other hosting services? 🤔

sergiou87:
Thanks for the quick review! I'll explore the tooltipped component suggestion in another PR, and also submit some unit tests for the new submodule status stuff.

PR #15190, opened 2022-08-29
Stop showing login prompts from GCM for git operations

### LINKED ISSUE #15163: persistent login requests - does not stay logged in
### The problem

since the latest update, i am prompted to sign in online, use my token, etc - i do (either way), and it prompts again at a later time - this did not happen before

### Release version

3.0.6

### Operating system

windows 10

1. open github desktop
2. wait
3. when prompted sign in via browser or token
4. wait
5. get prompted again
6. repeat steps 3-5

### Log files

[2022-08-24.desktop.production.log](https://github.com/desktop/desktop/files/9417501/2022-08-24.desktop.production.log)

### Screenshots

_No response_

### Additional context

was ready to move to new version as last version stopped working on the create pr and was hoping this would have the fix

sergiou87:
Closes #15163

## Description

This PR fixes a regression introduced in #14985 : one of the new git calls introduced performs a network call that might require authentication. If the user has any kind of credential helper installed (like [Git Credential Manger](https://github.com/GitCredentialManager/git-credential-manager)), they get a prompt asking to login on GitHub or other hosting services. The problem is we didn't disable external credential helpers, since GitHub Desktop should handle those authentication requirements using its [askpass trampoline](https://github.com/desktop/desktop-trampoline).

## Release notes

Notes: [Fixed] Do not show login prompt when repositories are fetched

### DIFF
--- a/app/src/lib/git/remote.ts
+++ b/app/src/lib/git/remote.ts
@@ -6,6 +6,7 @@ import { IRemote } from '../../models/remote'
 import { envForRemoteOperation } from './environment'
 import { IGitAccount } from '../../models/git-account'
 import { getSymbolicRef } from './refs'
+import { gitNetworkArguments } from '.'
 
 /**
  * List the remotes, sorted alphabetically by `name`, for a repository.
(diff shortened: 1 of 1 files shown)


niik (approved):

Rosso555 (commented):
update

PR #15191, opened 2022-08-29
Bump Electron to 19.0.13

mamantoha:


sergiou87:
Thank you for your PR! However, we require PRs to follow the template. In any case, bumping Electron is something we have to do carefully and after a ton of testing, since those upgrades tend to introduce regressions.

If you have any reason to upgrade Electron to 19.0.13, we'd love to hear it, but for now we're not going beyond 19.0.0.

mamantoha:
@sergiou87 got it. Electron v19.0.9 fixes empty `app_id` when running under Wayland - https://github.com/electron/electron/pull/34855. Without this GitHub Desktop doesn't have an icon in Task Switcher.

sergiou87:
Oh, thanks for clarifying @mamantoha 🙇‍♂️ 

Unfortunately that doesn't seem urgent enough to force this kind of (dangeours) upgrade, especially since we don't officially support Linux (sorry!!).

In any case, we'll have to upgrade eventually, and you'll get that fix in.

Sorry I don't bring better news 😓

PR #15159, opened 2022-08-23
Adding Accessibility Linting

tidy-dev:
## Description
Adding [recommended accessibility linting](https://thehub.github.com/epd/engineering/dev-practicals/frontend/accessibility/readiness-routine/linting-recommendations/#linting-in-ci).

Related Issue: https://github.com/github/accessibility/issues/1673

## Release notes
Notes: no-notes

PR #15204, opened 2022-08-30
Markdown Linting fixed and added to lint call

tidy-dev:
Builds on #15159 by addressing through our markdown linting errors. It is mostly line length errors, but some heading, list, and code fence formatting. This also adds the markdown lint call to our regular lint call so that future markdown written will be flagged in our ci checks.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,44 +1,69 @@
 # Installing GitHub Desktop
 
-GitHub Desktop currently supports Windows 7 (or higher) and macOS 10.9 (or higher).
+GitHub Desktop currently supports Windows 7 (or higher) and
+macOS 10.9 (or higher).
 
-### macOS
+## macOS
 
-Download the `GitHub Desktop.zip`, unpack the application and put it wherever you want.
+Download the `GitHub Desktop.zip`, unpack the application and put it wherever
+you want.
 
-### Windows
+## Windows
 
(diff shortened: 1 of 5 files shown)


docs/installation.md
```diff
@@ -1,44 +1,69 @@
...
 ## Data Directories
 
-GitHub Desktop will create directories to manage the files and data it needs to function. If you manage a network of computers and want to install GitHub Desktop, here is more information about how things work.
+GitHub Desktop will create directories to manage the files and data it needs to
+function. If you manage a network of computers and want to install GitHub
+Desktop, here is more information about how things work.
 
-### macOS
- - `~/Library/Application Support/GitHub Desktop/` - this directory contains user-specific data which the application requires to run, and is created on launch if it doesn't exist. Log files are also stored in this location.
+On macOs,
```
haykam821:
```suggestion
On macOS,
```

tidy-dev:
Not sure why - locally it runs without error.. but in actions it found several hundred more errors. This needs some more looking into..

tidy-dev:
Going to close this for now. I don't think the markdown linting is directly related to accessibility linting (doesn't show on the dashboard) and looks like quite a bit more effort to add it. (But, if after accessibility auditing we find we need this, we can resurrect this branch).

PR #15231, opened 2022-09-05
Fix tooltip positioning in dialogs

sergiou87:
## Description
This PR addresses a problem displaying tooltips in dialogs, because they're actually rendered relative to the window, in the wrong place.

Unless I'm wrong, it looks like `position: fixed` expects coordinates relative to the window, and we're passing coordinates relative to the `tooltip-host` element (the `dialog` in this case), so we have two options:
1. Use `position: absolute` so that positioning is relative to the parent of the tooltip (the dialog itself).
1. Keep using `position: fixed`, but use coordinates relative to the window to position the tooltip (i.e. delete these two lines: https://github.com/desktop/desktop/blob/5a9963c30e44e4cc91533673f29a2c759e088b04/app/src/ui/lib/tooltip.tsx#L619-L620)

I chose the second one, because it seems the change from `absolute` to `fixed` was intentional in https://github.com/desktop/desktop/commit/b54339c0baeb8157dfa08ed7ed757be1373f1b49, although it'd be great to have more context from @niik before merging this.

### Screenshots

| Before | After |
|--------|--------|
| image | image |

## Release notes

Notes: [Fixed] Fix misplaced tooltips from dialogs

### DIFF
--- a/app/src/ui/lib/tooltip.tsx
+++ b/app/src/ui/lib/tooltip.tsx
@@ -133,9 +133,6 @@ interface ITooltipState {
   /** The size and position of the target element relative to the window */
   readonly targetRect: DOMRect
 
-  /** The size and position of the tooltip parent relative to the window */
-  readonly hostRect: DOMRect
-
   /** The size of the window */
   readonly windowRect: DOMRect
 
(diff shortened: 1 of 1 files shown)


sergiou87:
After talking to @niik about this, it seems there is something wrong in my machine to cause this issue, since we weren't able to repro it anywhere else. I'll investigate further (mainly because those misplaced tooltips are driving me crazy 😂 ) and will reopen if needed with the right fix.

PR #15237, opened 2022-09-06
More submodule improvements

### LINKED ISSUE #7691: syntax highlighting error when rendering submodule diff
## Description

Identified while looking into a React issue, but I don't believe this is a regression in 2.0

## Version
* GitHub Desktop: 2.0.0
* Operating system: macOS 10.14.5

## Steps to Reproduce
1. Move a submodule inside a repository to a different commit
2. View the working directory changes

3. Check the DevTools for errors

### Expected Behavior
Syntax highlighting worker process detects and gracefully avoids rendering subnmodule diff

### Actual Behavior

Two errors logged each time it tries to update the diff:

```
Could not load new contents for syntax highlighting
Error: EISDIR: illegal operation on a directory, read
```

```
Could not load old contents for syntax highlighting
Error: Git returned an unexpected exit code '128' which should be handled by the caller.'
    at ChildProcess.s.on (/webpack:/app/src/lib/git/spawn.ts:93:15)
    at ChildProcess.emit (events.js:182:13)
    at maybeClose (internal/child_process.js:961:16)
    at Socket.stream.socket.on (internal/child_process.js:380:11)
    at Socket.emit (events.js:182:13)
    at Pipe._handle.close [as _onclose] (net.js:596:12)
```

sergiou87:
Closes #7691

## Description
While looking into #7691 I realized we could also render submodules nicely in the Commit/History view could be useful. This PR does exactly that. Main changes/differences:
- While for active changes we use `git status`, for commit changes we use `git log`, and the output is different. In order to detect submodules, we must look for the file mode `160000`.
- In commits, submodules cannot have untracked or modified changes.
- Verbs change in the submodule diff screen, reflecting active vs past change.
- Sometimes, a submodule referenced within a commit doesn't exist anymore and cannot be opened, therefore the `Open Repository` quick action must be removed.
- Suggestions also change: in active changes context we suggest the user which changes can be committed and which cannot. That doesn't make sense within a commit.
- Added info about the commit a submodule is pointing at for added/deleted submodules.

### Screenshots

image
image
image
image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/log.ts
+++ b/app/src/lib/git/log.ts
@@ -6,6 +6,7 @@ import {
   CopiedOrRenamedFileStatus,
   UntrackedFileStatus,
   AppFileStatus,
+  SubmoduleStatus,
 } from '../../models/status'
 import { Repository } from '../../models/repository'
 import { Commit } from '../../models/commit'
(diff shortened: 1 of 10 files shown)


app/src/lib/git/log.ts
```diff
@@ -232,7 +268,19 @@ export function parseRawLogWithNumstat(
   for (let i = 0; i < lines.length - 1; i++) {
     const line = lines[i]
     if (line.startsWith(':')) {
-      const status = forceUnwrap('Invalid log output', line.split(' ').at(-1))
+      const lineComponents = line.split(' ')
+      const srcMode = forceUnwrap(
+        'Invalid log output (srcMode)',
+        lineComponents[0].replace(':', '')
```
niik: Shouldn't we use `?.` here? If we don't this will throw for `[0]` being undefined instead of the `Invalid log output` error from forceUnwrap
```suggestion
        lineComponents[0]?.replace(':', '')
```

niik (changes requested):
Looks fantastic @sergiou87! Just a small question

app/src/lib/git/log.ts
sergiou87: That's right! 😱  Nice catch ❤️

niik (approved):

PR #15256, opened 2022-09-08
PR View: Add state property and methods to repository model

tidy-dev:
## Description
This PR adds a `pullRequestState` to the `repositoryState` as well as methods to update it. This seems most logical to me since pull requests exist per repository - but always open to feedback/recommendations. Likely this state will expand to house the pull request body properties as well. 

Another thought was to make a `pullRequestStateCache` similar to `repositoryState` and `repository-state-cache`. The repository state cache works by holding a map of `IRepositoryState` -> an entry for each repository. So far, we just keep appending onto that for something that is a property of the `IRepositoryState`. For example, we have `updateCompareState`, `updateChangesState`, `updateCommitSelection` and so on and so forth.  Thus, instead of pullRequestState being another child property of repositoryState and appending yet another set of methods (as done in this PR). We could have a `pullRequestStateCache` that also has state map by repository as well and it now has it's 'update' method and 'updateChildMethod'. This could be beneficial for these models with their own children like pull requests and multi commit operations where we have these null checks on each update. Thus, app-store would get another property instead. This would house everything to do with managing PR's. (Separating concerns a little bit). Other down falls is having to pass down the pullRequestState down in the repository view instead of just being able to pull it off the repository state (not sure how this works with the emit functionality)

Other places I looked for this state to live is we have a the `PullRequestStore`, but this felt more about managing existing pull requests/the pr database and not app state management. Tho, I could see that being a logical extension. 

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -419,6 +419,16 @@ export interface IRepositoryState {
   readonly selectedSection: RepositorySectionTab
 
   /**
+   * The state of the current pull request view in the repository.
+   *
+   * It will populated when a user initiates a pull request. It may have
+   * content to retain a users pull request state if they navigate
+   * away from the current pull request view and then back. It is returned
+   * to null after a pull request has been opened.
+   */
+  readonly pullRequestState: IPullRequestState | null
+
+  /**
    * The name and email that will be used for the author info
    * when committing barring any race where user.name/user.email is
    * updated between us reading it and a commit being made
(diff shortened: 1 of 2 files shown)


app/src/lib/app-state.ts
```diff
@@ -418,6 +418,16 @@ export interface IRepositoryState {
   readonly compareState: ICompareState
   readonly selectedSection: RepositorySectionTab
 
+  /**
+   * The state of the current pull request view in the repository.
+   *
+   * It will populated when a user initiates a pull request. It may have
```
sergiou87: Nit:
```suggestion
   * It will be populated when a user initiates a pull request. It may have
```

app/src/lib/app-state.ts
```diff
@@ -418,6 +418,16 @@ export interface IRepositoryState {
...
+   * content to retain a users pull request state if they navigate
+   * away from the current pull request view and then back. It is returned
+   * to null after a pull request has been opened.
```
sergiou87, line 427: …or the dialog is closed, right?

app/src/lib/app-state.ts
```diff
@@ -928,3 +938,28 @@ export interface IConstrainedValue {
...
   readonly min: number
 }
+
+/**
+ * The state of the current pull request view in the repository.
+ */
+export interface IPullRequestState {
+  /**
+   * The base branch of a a pull request - the branch the currently checked out
+   *  branch would merge into
```
sergiou87: Nit:
```suggestion
   * branch would merge into
```

sergiou87 (dismissed):
Looks good!! I think this approach is better, at least to begin with. I only found a few nits in the comments, feel free to ignore them if you want 😄

app/src/lib/app-state.ts
tidy-dev: Maybe... what if the user starts a PR and fills out the body but then gets a notification about another PR.. maybe we want to remember where they are at and let them get back to it?

sergiou87: Yeah in that case, it'd be nice to keep it.

sergiou87 (approved):
✨

PR #15259, opened 2022-09-09
PR View: Pull request view app store methods

tidy-dev:
## Description
Adds app-store methods to set the pr state. One to set the initial state and one to update the selected file in the pr state. Along the way I realized I needed a method to obtain the commits of the PR at this point - that is added to the git-store.

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7156,16 +7156,20 @@ export class AppStore extends TypedBaseStore<IAppState> {
     )
 
     const commitSHAs = pullRequestCommits.map(c => c.sha)
-    const firstCommit = commitSHAs[0]
 
-    const changesetData = await gitStore.performFailableOperation(() =>
-      getBranchMergeBaseChangedFiles(
-        repository,
-        defaultBranch.name,
-        currentBranch.name,
-        firstCommit
-      )
-    )
+    // A user may compare two branches with no changes between them.
+    const emptyChangeSet = { files: [], linesAdded: 0, linesDeleted: 0 }
+    const changesetData =
+      commitSHAs.length > 0
+        ? await gitStore.performFailableOperation(() =>
+            getBranchMergeBaseChangedFiles(
+              repository,
+              defaultBranch.name,
+              currentBranch.name,
+              commitSHAs[0]
+            )
+          )
+        : emptyChangeSet
 
     if (changesetData === undefined) {
       return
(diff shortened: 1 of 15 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -7136,6 +7138,121 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      return
+    }
+
+    const currentBranch = tip.branch
+    const gitStore = this.gitStoreCache.get(repository)
+
+    const pullRequestCommits = await gitStore.getCommitsBetweenBranches(
+      defaultBranch,
+      currentBranch
+    )
```
sergiou87, line 7156: We should probably check if there are no commits at all, right?
```suggestion
    const pullRequestCommits = await gitStore.getCommitsBetweenBranches(
      defaultBranch,
      currentBranch
    )
    
    if (pullRequestCommits.length === 0) {
      return
    }
```

app/src/lib/stores/app-store.ts
```diff
@@ -7136,6 +7138,121 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+      (await this.gitStoreCache
+        .get(repository)
+        .performFailableOperation(() =>
+          getBranchMergeBaseDiff(
+            repository,
+            file,
+            baseBranch.name,
+            currentBranch.name,
+            this.hideWhitespaceInHistoryDiff,
+            commitSHAs[0]
```
sergiou87, line 7244: Up to this point, `commitSHAs` shouldn't be empty, but maybe we should check it and not generate a diff in that case?

sergiou87 (changes requested):
Looks good overall, just found a few edge cases that we might want to check 🤔

app/src/lib/stores/app-store.ts
tidy-dev: Yeah. Tho, it "shouldn't" happen, it wouldn't hurt. :D

app/src/lib/stores/app-store.ts
tidy-dev: Very good point that I intentionally left out at the moment,  because I didn't know if we would like to do something like:
image

I think we will want this because we are going to let a user change the base branch in the modal and we wouldn't want it to just close the dialog. :D However, I think I may refactor the initialization logic a little to allow a pr state without a change set. I was planning to do that in a later PR after discussion with group on possibility of a no changes dialog. But now that I think on it more, I may want to just go ahead and plan for it.

tidy-dev: Added an empty change set for this case.

sergiou87: Yeah, your change makes much more sense then. I was concerned about the `commitSHAs[0]` which would be `undefined` in that case, but now it's all good 🙌

sergiou87 (approved):
✨ :_D

PR #15272, opened 2022-09-12
PR View: Pull request dialog header

tidy-dev:
## Description
Add header to pull request dialog. Later to make the base branch into a dropdown. 

### Screenshots

image

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -2250,17 +2250,22 @@ export class App extends React.Component<IAppProps, IAppState> {
         }
 
         const { state: repoState, repository } = selectedState
-        const { pullRequestState } = repoState
-        if (pullRequestState === null) {
+        const { pullRequestState, branchesState } = repoState
+        if (
+          pullRequestState === null ||
+          branchesState.tip.kind !== TipState.Valid
+        ) {
           return null
         }
+        const currentBranch = branchesState.tip.branch
 
         return (
           <OpenPullRequestDialog
             key="open-pull-request"
+            currentBranch={currentBranch}
             dispatcher={this.props.dispatcher}
-            repository={repository}
             pullRequestState={pullRequestState}
+            repository={repository}
             onDismissed={onPopupDismissedFn}
           />
         )
--- a/app/src/ui/open-pull-request/open-pull-request-header.tsx
+++ b/app/src/ui/open-pull-request/open-pull-request-header.tsx
@@ -1,9 +1,13 @@
 import * as React from 'react'
+import { Branch } from '../../models/branch'
 import { DialogHeader } from '../dialog/header'
 import { createUniqueId } from '../lib/id-pool'
 import { Ref } from '../lib/ref'
 
 interface IOpenPullRequestDialogHeaderProps {
+  /** The base branch of the pull request */
+  readonly baseBranch: Branch
+
   /**
    * Event triggered when the dialog is dismissed by the user in the
    * ways described in the dismissable prop.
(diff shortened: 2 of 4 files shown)


app/src/ui/open-pull-request/open-pull-request-header.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <DialogHeader
+        title={title}
+        titleId={createUniqueId(`Dialog_${title}_${title}`)}
+        dismissable={true}
+        onDismissed={onDismissed}
+      >
+        <div className="break"></div>
+        <div className="base-branch-details">
+          Merge {commits} into <Ref>{baseBranch.name}</Ref> from{' '}
+          <Ref>{currentBranch.name}</Ref>.
```
tidy-dev, line 47: I used the `<Ref>` component to highlight the branch names like it is on a pull request on dotcom. But, our merge branch dialog bolds branch names so I am happy to change it if anyone feels that bolding is more inline with Desktop and thus we should do that. (I am more inclined to think we should use the `<Ref>` component everywhere in desktop for branch names and commits when in a sentence).

image

app/src/ui/open-pull-request/open-pull-request-header.tsx
```diff
@@ -41,7 +44,7 @@ export class OpenPullRequestDialogHeader extends React.Component<
         <div className="break"></div>
         <div className="base-branch-details">
           Merge {commits} into <Ref>{baseBranch.name}</Ref> from{' '}
-          <Ref>feature-branch</Ref>.
+          <Ref>{currentBranch.name}</Ref>.
         </div>
```
yasoo655, line 48: I want to withdraw my money out 100 dollar 💵

yasoo655 (commented):
>

app/src/ui/app.tsx
```diff
@@ -2250,17 +2250,22 @@ export class App extends React.Component<IAppProps, IAppState> {
         }
 
         const { state: repoState, repository } = selectedState
-        const { pullRequestState } = repoState
-        if (pullRequestState === null) {
+        const { pullRequestState, branchesState } = repoState
+        if (
+          pullRequestState === null ||
+          branchesState.tip.kind !== TipState.Valid
+        ) {
```
sergiou87, line 2257: I wonder if the `popup` object should have a snapshot of all this data in it, instead of having to grab it at render time 🤔 

That also makes me wonder if the `return null` pattern is correct in this method (I see it's used in other `case`s too). What I mean is, when we create the `popup` object it should already have all the info needed for that popup, and therefore there shouldn't be this kind of "invalid popup" situations. But that's just food for thought 😳

app/src/ui/open-pull-request/open-pull-request-header.tsx
sergiou87: Since in this case the branch names are not in the title but in the subtitle, I _think_ it makes sense to show them within `Ref` components. Or, at least, it doesn't feel that odd 😅

sergiou87 (approved):
I think it looks great!! I left a comment/question, but feel free to ignore and merge if you want :shipit:

app/src/ui/app.tsx
tidy-dev: Hmm.. I suppose and maybe it is worth throwing an error if we somehow how got to this state.. because otherwise the app just does nothing on the user invoking this popup (of course shouldn't happen..). However, I think it would just be putting these checks somewhere else (dispatcher or app-store) also we will need quite a few things from the repoState to pass down into the popup for diffing... 

I will look at this in my PR for introducing the diff view and see if it feels better in the dispatcher or app-store/ then I should know what all needs pass in anyways. I could see it having less checks in (or already checked) in the app-store. Thanks for the input.

PR #2381, opened 2017-08-01
(almost) always be updating

shiftkey:
Fixes #2373 by ignoring requests to check for updates when the app has downloaded and applied an update and is ready for restarting.

Issue #2373, opened 2017-07-31
update cleaning up old git.exe when it shouldn't


shiftkey:
### Description

@NickCraver reported on Twitter that he encountered this Git error, which is a generic and questionable message when we can't resolve the packaged Git:

![](https://user-images.githubusercontent.com/359239/28799857-99b793b4-7620-11e7-9376-0a71f545a8bf.jpg)

However after this he hasn't been able to launch Desktop at all. No errors, no logs, just silence.

![](https://cloud.githubusercontent.com/assets/24247/18013839/02da86b0-6b8e-11e6-9202-a98eef7cce6f.gif)

### Version

**GitHub Desktop version:** 0.7.0 -> 0.7.2
**OS version:** Windows 10 I guess

### Steps to Reproduce

TBC, seems to be something like this:

 - Open app running 0.7.0
 - Update app to 0.7.x (latest version) (in background? or manually?)
 - Leave app open, keep doing Git things
 - :boom:

**Expected behavior:** The current version's bits are still on disk until the app is closed, and cleaned up In The Future™

**Actual behavior:** Files related to the running app were cleaned up, and now the app doesn't launch. Not sure if these are related, let's assume so.

**Reproduces how often:** lol I have no idea

### Logs

`SquirrelSetup.log` is interesting:

![setup](https://user-images.githubusercontent.com/359239/28800030-4b842206-7621-11e7-835b-14311c49c67e.jpg)

It excluded `0.7.1` and `0.7.2` when cleaning up old versions, but not the running `0.7.0` version. Could this be related to us rolling out delta updates? [Full log is here](https://gist.github.com/NickCraver/5fbeedfaa316960b800e12cacdab22ce#file-squirrelsetup-log)

NickCraver:
I captured the state of `\AppData\Local\GitHubDesktop` in a .7z (it's 217MB). If that's useful at all, happy to upload somewhere. Also of course happy to leave it broken and debug locally for a bit to prevent it from happening to others, just ping me!

shiftkey:
I hate it when my psychic debugger is right:

<img width="671"  src="https://user-images.githubusercontent.com/359239/28800201-0ba78668-7622-11e7-9ea2-850dc2e7bc40.png">

<img width="653" src="https://user-images.githubusercontent.com/359239/28800214-19accbba-7622-11e7-9680-c33d055e5feb.png">

So delta updates maybe don't track the current running app? I'd need to try and reproduce this before I go down that rabbit hole.

[...]

PR #15275, opened 2022-09-13
Don't check for updates if there's one ready

sergiou87:
## Description
With staggered releases, we're starting to see some crashes from people who:
1. Download an update automatically with GitHub Desktop, but…
2. …don't restart in a long time.
3. Then we do another release but either it's disabled or not enabled to all users, so…
4. …if users don't have access to the new version, GitHub Desktop will eventually check for new updates again and will find nothing.
5. Then, at some point after that, the user will finally hit `Restart now`  or `Quit and Install Update` buttons.
6. 💥 

While looking into this, I ran into [an old code](https://github.com/desktop/desktop/blob/7bf9b1bb55b83e5338ca8915911e732647ae462b/app/src/ui/lib/update-store.ts#L184-L191) introduced in https://github.com/desktop/desktop/pull/2381 that disabled checking for updates when an update was already downloaded and ready to install, but only on Windows. Instead of doing anything different, we'll do the same on macOS. That way, users who have already downloaded an update will be able to, at least, upgrade to that version, even if newer ones are disabled. 

## Release notes

Notes: [Fixed] Fix crash installing updates that were downloaded a long time ago on macOS

### DIFF
--- a/app/src/ui/lib/update-store.ts
+++ b/app/src/ui/lib/update-store.ts
@@ -184,8 +184,10 @@ class UpdateStore {
   public async checkForUpdates(inBackground: boolean, skipGuidCheck: boolean) {
     // An update has been downloaded and the app is waiting to be restarted.
     // Checking for updates again may result in the running app being nuked
-    // when it finds a subsequent update.
-    if (__WIN32__ && this.status === UpdateStatus.UpdateReady) {
+    // when it finds a subsequent update on Windows, or the "Quit and Update"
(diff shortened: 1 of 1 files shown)


tidy-dev (approved):
✨  Makes sense - curious why this was limited to windows before.

sergiou87:
I only know what's explained in https://github.com/desktop/desktop/pull/2381 and https://github.com/desktop/desktop/issues/2373, it might be something specific about how Squirrel.Windows works (its updates are kind of crazy, IMHO)

PR #14813, opened 2022-06-15
Bump Electron to v19

sergiou87:
## Description

This PR upgrades Electron to v19. Luckily, no API changes affected us 🎉  so it'll be easy to revert/cherry-pick later if needed.

I've tested it as much as I could, but it'd be nice if other people did the same before merging it or deploying to beta 😨  Thanks!

## Release notes

Notes: no-notes

PR #15276, opened 2022-09-13
Enable workaround for DevTools crash in beta/prod builds

sergiou87:
## Description
When we upgraded to Electron v19 in #14813 we introduced a regression that was mostly annoying for us devs: https://github.com/electron/electron/issues/34350

In theory that bug was gonna be resolved in a newer version of Electron 19.x, but it didn't. The crash is solved in v20 instead. Since we released Electron v19 to prod in v3.0.8 (#15239), we think users that like to poke around with the DevTools might be affected too, so we decided to bring the workaround to beta/prod/test builds too.

## Release notes

Notes: no-notes


tidy-dev (approved):
✨

PR #15277, opened 2022-09-13
PR View: Branch select component

tidy-dev:
## Description
Adds a branch select component utilizing the desktop branch list and popover component. 

Decisions that I could go either way:
1. Filter text is remembered if you close and open the popup. Reason I kept it was if you filter and accidentally click off it, you don't lose it. But, I think it could also be expected that this should clear it.
2. Used the popover component, but removed the carrot. I think this feels more like a select component that way.
3. Other thought I had was, should this dropdown have more visual cue that it is a dropdown besides the down arrow? (Not sure what that would be, on dotcom when you hit edit the base branch looks more like a button) 

### Screenshots
https://user-images.githubusercontent.com/75402236/189933915-b77973bc-8286-48dc-9872-d1586f791cc4.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/branches/branch-select.tsx
+++ b/app/src/ui/branches/branch-select.tsx
@@ -5,8 +5,8 @@ import { Octicon } from '../octicons'
 import * as OcticonSymbol from '../octicons/octicons.generated'
 
 interface IBranchSelectProps {
-  /** The selected branch. */
-  readonly selectedBranch: Branch
+  /** The initially selected branch. */
+  readonly branch: Branch
 
   /** Called when the user changes the selected branch. */
   readonly onChange?: (branch: Branch) => void
--- a/app/src/ui/branches/branch-select.tsx
+++ b/app/src/ui/branches/branch-select.tsx
@@ -3,6 +3,7 @@ import { IMatches } from '../../lib/fuzzy-find'
 import { Branch } from '../../models/branch'
 import { Button } from '../lib/button'
 import { ClickSource } from '../lib/list'
+import { Popover, PopoverCaretPosition } from '../lib/popover'
 import { Ref } from '../lib/ref'
 import { Octicon } from '../octicons'
 import * as OcticonSymbol from '../octicons/octicons.generated'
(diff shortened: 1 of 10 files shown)


app/styles/ui/_branch-select.scss
```diff
@@ -0,0 +1,73 @@
...
+    width: 365px;
+    padding: 0;
+    margin-top: 25px;
+
+    &.popover-component {
+      &::before,
+      &::after {
+        display: none;
+      }
+    }
```
sergiou87: Could we handle this via a new `PopoverCaretPosition.None` enum value? (happy to wait until upcoming PRs to address this if you want)

app/src/ui/branches/branch-select.tsx
```diff
@@ -0,0 +1,191 @@
...
+      </Popover>
+    )
+  }
+
+  public render() {
+    return (
+      <div className="branch-select-component">
+        <Button
+          onClick={this.toggleBranchDropdown}
+          onButtonRef={this.onInvokeButtonRef}
```
sergiou87, line 178: If you just want to store it, I think you can do just this:
```suggestion
          onButtonRef={this.invokeButtonRef}
```

sergiou87 (commented):
Looks good overall! Just had a couple of suggestions

sergiou87 (dismissed):
Oops! Let me approve the PR in case you wanna address my comments in upcoming PRs, since none of them is a blocker.

app/styles/ui/_branch-select.scss
tidy-dev: That reminds me that I meant to make the caret property optional (cause me providing it in the branch select component doesn't make much sense). I will go ahead and take care of that on this PR.

app/src/ui/branches/branch-select.tsx
tidy-dev: I just verified that this does not work here because it is not a true `ref` as the `onButtonRef` is a callback for the `ref` on the `button` (lowercase) down in the wrapper `Button` component (uppercase);

sergiou87 (approved):
Thanks for those changes! Ready to merge 🙌 :shipit:

PR #15294, opened 2022-09-16
Submodule diff stats

sergiou87:
## Description

This PR tracks some basic stats for the new submodule diffs:
- When a submodule diff is viewed from the Changes list
- When a submodule diff is viewed from the History tab (from inside a commit)
- When the user clicks on `Open Repository` from a submodule diff

It'll be nice to see how much users actually run into submodule diffs, and how often they open the given submodules from this new view.

Happy to hear feedback about whether or not you think these stats are useful, or if there are any other metrics you consider valuable.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -532,6 +532,12 @@ export interface IDailyMeasures {
 
   /** The number of times the user opens the unreachable commits dialog */
   readonly multiCommitDiffUnreachableCommitsDialogOpenedCount: number
+
+  /** The number of times the user opens a submodule diff from the changes list */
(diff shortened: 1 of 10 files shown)


tidy-dev (approved):
✨  Those stats make sense to me. :) 

Thoughts: First I was sort of taken a back by passing the stat store into the repository cache.. like stats stuff didn't fit there? But, I can see this feels cleaner than adding this check everywhere one of those methods is called that changes a files selection. Only last thought was I couldn't make up my mind if the stat checking logic makes more sense in the state-cache or in the stat-store. :/  So, there is my wishy-washy comment to say it felt off but I can't think of a better way. :D

PR #15319, opened 2022-09-20
PR View: Pull request files changed component (start of)

tidy-dev:
## Description

This PR adds a component for viewing the files changed and adds the seamless diff viewer to it. Thus, adds the properties that must be passed down for diffing. Will follow by adding the file list, commit dropdown, and diff setting.

Additionally, worked in [a suggestion](https://github.com/desktop/desktop/pull/15272#discussion_r974489908) from previous PR for passing needed properties on the popup model to avoid repeated null checks.  

### Screenshots

image

## Release notes
Notes: no - notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7194,6 +7194,35 @@ export class AppStore extends TypedBaseStore<IAppState> {
         changesetData.files[0]
       )
     }
+
+    const { pullRequestState } = this.repositoryStateCache.get(repository)
+    if (pullRequestState === null) {
+      // This shouldn't happen.. we just initialized it.
+      sendNonFatalException(
+        'startPullRequest',
+        new Error(
+          'Failed to start pull request because pull request state was null'
+        )
+      )
+      return
+    }
+
+    const { allBranches, recentBranches } = branchesState
+    const { imageDiffType, hideWhitespaceInHistoryDiff, showSideBySideDiff } =
+      this.getState()
+
+    this._showPopup({
+      type: PopupType.StartPullRequest,
+      allBranches,
+      currentBranch,
+      defaultBranch,
+      hideWhitespaceInHistoryDiff,
+      imageDiffType,
+      pullRequestState,
+      recentBranches,
+      repository,
+      showSideBySideDiff,
+    })
   }
 
   public async _changePullRequestFileSelection(
--- a/app/src/models/popup.ts
+++ b/app/src/models/popup.ts
@@ -14,7 +14,7 @@ import { Commit, CommitOneLine, ICommitContext } from './commit'
 import { IStashEntry } from './stash-entry'
 import { Account } from '../models/account'
 import { Progress } from './progress'
-import { ITextDiff, DiffSelection } from './diff'
+import { ITextDiff, DiffSelection, ImageDiffType } from './diff'
 import { RepositorySettingsTab } from '../ui/repository-settings/repository-settings'
 import { ICommitMessage } from './commit-message'
 import { IAuthor } from './author'
@@ -22,6 +22,7 @@ import { IRefCheck } from '../lib/ci-checks/ci-checks'
 import { GitHubRepository } from './github-repository'
 import { ValidNotificationPullRequestReview } from '../lib/valid-notification-pull-request-review'
 import { UnreachableCommitsTab } from '../ui/history/unreachable-commits-dialog'
+import { IPullRequestState } from '../lib/app-state'
 
 export enum PopupType {
   RenameBranch = 1,
@@ -362,4 +363,13 @@ export type Popup =
     }
   | {
       type: PopupType.StartPullRequest
+      allBranches: ReadonlyArray<Branch>
+      currentBranch: Branch
+      defaultBranch: Branch | null
+      hideWhitespaceInHistoryDiff: boolean
+      imageDiffType: ImageDiffType
+      pullRequestState: IPullRequestState
+      recentBranches: ReadonlyArray<Branch>
+      repository: Repository
+      showSideBySideDiff: boolean
     }
--- a/app/src/ui/open-pull-request/pull-request-diff.tsx
+++ b/app/src/ui/open-pull-request/pull-request-diff.tsx
@@ -1,8 +1,29 @@
 import * as React from 'react'
-import { IDiff } from '../../models/diff'
+import { IDiff, ImageDiffType } from '../../models/diff'
+import { Repository } from '../../models/repository'
+import { CommittedFileChange } from '../../models/status'
+import { SeamlessDiffSwitcher } from '../diff/seamless-diff-switcher'
+import { Dispatcher } from '../dispatcher'
+import { openFile } from '../lib/open-file'
 
 interface IPullRequestDiffProps {
+  readonly repository: Repository
+  readonly dispatcher: Dispatcher
+
+  /** The file whose diff should be displayed. */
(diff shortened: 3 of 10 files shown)


app/styles/ui/_pull-request-diff.scss
```diff
@@ -0,0 +1,4 @@
+.pull-request-diff-viewer {
+  height: 500px;
```
tidy-dev, line 2: This static height is likely temporary and just to make the diff not expand past the dialog during development of this component. I will have to eventually determine the logic for dynamically setting this height based on the users window size so that the dialog can use as much room as is available within whatever min/max constraints we figure out for the diff (similarly for width).

app/styles/ui/dialogs/_open-pull-request.scss
```diff
@@ -15,4 +15,9 @@
       padding: var(--spacing-half);
     }
   }
+
+  .content {
+    height: 500px;
```
tidy-dev, line 20: Also subject to change (see comment on pull-request-diff.scss)

app/src/ui/open-pull-request/pull-request-diff.tsx
```diff
@@ -0,0 +1,96 @@
...
+        hideWhitespaceInDiff={hideWhitespaceInDiff}
+        showSideBySideDiff={showSideBySideDiff}
+        onOpenBinaryFile={this.onOpenBinaryFile}
+        onChangeImageDiffType={this.onChangeImageDiffType}
+        onHideWhitespaceInDiffChanged={this.onHideWhitespaceInDiffChanged}
+      />
+    )
+  }
+
+  public render() {
```
tidy-dev: My plan is that this component will house the file list, commit dropdown, diff setting dropdown for the diff as well (_maybe I should rename it to pull-request-files-changed_) . I think the empty change set may need to be handled higher up - more of a "This PR doesn't have changes" blank state for the dialog because when we get to having tabs, we won't want to show the body input tab either in this scenario.

app/src/ui/open-pull-request/pull-request-diff.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { IDiff, ImageDiffType } from '../../models/diff'
+import { Repository } from '../../models/repository'
+import { CommittedFileChange } from '../../models/status'
+import { SeamlessDiffSwitcher } from '../diff/seamless-diff-switcher'
+import { Dispatcher } from '../dispatcher'
+import { openFile } from '../lib/open-file'
+
+interface IPullRequestDiffProps {
+  readonly repository: Repository
+  readonly dispatcher: Dispatcher
```
tidy-dev: I am not a huge fan of passing the dispatcher down far into the component structure but there were a few dispatcher methods that would only pertain to diffing and felt muddy at the dialog level (which I am trying to reserve for the purpose of managing the tab state). Since we don't have a strict convention to not passing the dispatcher down, I think it is cleaner to go ahead and pass it down here.

app/src/ui/open-pull-request/pull-request-diff.tsx
tidy-dev: ... changed the component name. :D..

app/src/lib/stores/app-store.ts
```diff
@@ -7194,6 +7194,35 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        'startPullRequest',
+        new Error(
+          'Failed to start pull request because pull request state was null'
+        )
+      )
+      return
+    }
+
+    const { allBranches, recentBranches } = branchesState
+    const { imageDiffType, hideWhitespaceInHistoryDiff, showSideBySideDiff } =
```
tidy-dev, line 7211: >  hideWhitespaceInHistoryDiff

There is one of these for changes, and one for history... maybe there should be one for pr diff or maybe it should be remember on the pr dialog state (not app state). Thoughts?

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -0,0 +1,96 @@
...
+   * to change the diff presentation mode.
+   */
+  private onChangeImageDiffType = (imageDiffType: ImageDiffType) => {
+    this.props.dispatcher.changeImageDiffType(imageDiffType)
+  }
+
+  private renderDiff() {
+    const { diff, selectedFile } = this.props
+
+    if (diff === null || selectedFile === null) {
```
tidy-dev, line 65: > diff === null

Didn't realize till I was testing the file switcher..., but the SeamlessDiffSwitcher interprets a `null` here as loading. This diff === null check will be removed in next PR so that between file selections we will see a loader

app/src/models/popup.ts
```diff
@@ -362,4 +363,13 @@ export type Popup =
     }
   | {
       type: PopupType.StartPullRequest
+      allBranches: ReadonlyArray<Branch>
+      currentBranch: Branch
+      defaultBranch: Branch | null
+      hideWhitespaceInHistoryDiff: boolean
+      imageDiffType: ImageDiffType
+      pullRequestState: IPullRequestState
```
tidy-dev, line 371: Next PR is going to end up removing pullRequestState but rest of variables are simply inputs so they will should remain on the popup model

app/src/lib/stores/app-store.ts
sergiou87: Saving another pref sounds overkill, but maybe makes sense 😭 Another option is to make it match one of the existing settings when the dialog is opened, and then keep whatever the user does locally within the dialog?

I don't know… I think just having a third setting is the option I hate the least (but still hate it) 🤔

QUGL-98-06-15 (commented):
Submit general feedback without explicit approval.

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -0,0 +1,96 @@
...
+    openFile(fullPath, this.props.dispatcher)
+  }
+
+  /** Called when the user changes the hide whitespace in diffs setting. */
+  private onHideWhitespaceInDiffChanged = (hideWhitespaceInDiff: boolean) => {
+    const { selectedFile } = this.props
+    return this.props.dispatcher.onHideWhitespaceInHistoryDiffChanged(
+      hideWhitespaceInDiff,
+      this.props.repository,
+      selectedFile as CommittedFileChange
```
sergiou87, line 50: Do you need to cast this? `selectedFile` should already be `CommittedFileChange` (or `null`, but that's accepted too).
```suggestion
      selectedFile
```

sergiou87 (approved):
Looks good to me! I just found an unnecessary casting (unless I'm missing something), so I'm gonna approve it in case you want to ignore that or change it in another PR.

app/src/lib/stores/app-store.ts
tidy-dev: I added a third option in the diffOptions PR .. I landed on it feeling wierd to use changes/history since neither of them were the same. And it was just easier to make the the third option like the others instead of managing in dialog state. But, I am not strong on that decision. (We could restructure for this and default as false for dialog initialization and then pass it up for all future git calls (every time user changes a file selection or changes base branch).

app/src/ui/open-pull-request/pull-request-files-changed.tsx
tidy-dev: I copied this from selected commit and didn't question it (should have 😞). I am guessing at one point null wasn't handled.. I will catch this in another PR.

PR #15325, opened 2022-09-21
PR View: Add files list

tidy-dev:
Built on https://github.com/desktop/desktop/pull/15319

## Description
This adds the file list to the changes file tab diff. It also sets the width of the dialog to a static 850px that fits in the min app window size. Plan to investigate making the modal size responsive to the app window size, but for now I just needed a "default" minimum to make files list pane resizing start working.

### Screenshots

image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -170,6 +170,9 @@ export interface IAppState {
   /** The width of the files list in the stash view */
   readonly stashedFilesWidth: IConstrainedValue
 
+  /** The width of the files list in the pull request files changed view */
+  readonly pullRequestFilesListWidth: IConstrainedValue
+
   /**
    * Used to highlight access keys throughout the app when the
    * Alt key is pressed. Only applicable on non-macOS platforms.
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7195,18 +7195,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
       )
     }
 
-    const { pullRequestState } = this.repositoryStateCache.get(repository)
-    if (pullRequestState === null) {
-      // This shouldn't happen.. we just initialized it.
-      sendNonFatalException(
-        'startPullRequest',
-        new Error(
-          'Failed to start pull request because pull request state was null'
-        )
-      )
-      return
-    }
-
     const { allBranches, recentBranches } = branchesState
     const { imageDiffType, hideWhitespaceInHistoryDiff, showSideBySideDiff } =
       this.getState()
@@ -7218,7 +7206,6 @@ export class AppStore extends TypedBaseStore<IAppState> {
       defaultBranch,
       hideWhitespaceInHistoryDiff,
       imageDiffType,
-      pullRequestState,
       recentBranches,
       repository,
       showSideBySideDiff,
@@ -7252,6 +7239,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         diff: null,
       })
     )
+
     this.emitUpdate()
 
     if (commitSHAs.length === 0) {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7196,8 +7196,12 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
 
     const { allBranches, recentBranches } = branchesState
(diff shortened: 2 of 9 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -2077,6 +2087,41 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.stashedFilesWidth = constrain(this.stashedFilesWidth, 100, filesMax)
   }
 
+  /**
+   * Calculate the constraints of the resizable pane in the pull request dialog
+   * whenever the window dimensions change.
+   */
+  private updatePullRequestResizableConstraints() {
+    // TODO: Get width of PR dialog -> determine if we will have default width
```
tidy-dev, line 2095: At this point in time, the width of the dialog is static; thus, solving for this is not useful. I will address updating this logic when I solve the responsive width of the dialog challenge... if for some reason we deem having a responsive dialog unrealistic, this method likely isn't needed and we can just initialize the min/max constraints.

app/src/lib/app-state.ts
```diff
@@ -170,6 +170,9 @@ export interface IAppState {
   /** The width of the files list in the stash view */
   readonly stashedFilesWidth: IConstrainedValue
 
+  /** The width of the files list in the pull request files changed view */
+  readonly pullRequestFilesListWidth: IConstrainedValue
```
tidy-dev, line 174: I debated whether it made sense to keep track of this at the app level or since it is not used for anything outside that put it on the dialog state. Reason I chose to keep it here was just for consistency of where we keep all the other file width tracking logic and that we seem to have a convention of accessing local storage here and then adding it to the app state. Unless we move accessing/updating local storage down to the dialog, we would still have the same amount of update methods in the dispatcher/app-store. Completely open to moving this if anyone feels it makes more sense to scope it down to the dialog.

app/styles/ui/dialogs/_open-pull-request.scss
```diff
@@ -1,4 +1,7 @@
 .open-pull-request {
+  width: 850px;
```
tidy-dev, line 2: Fits in minimum window constraints -> may change this when solving for a responsive height/width design.

app/src/lib/app-state.ts
sergiou87: It _feels_ like it belongs to the dialog scope, but at the same time it's a very specific and "global" dialog (like we switch to a different app mode) so I'm not too bothered about it.

app/src/lib/stores/app-store.ts
```diff
@@ -323,6 +323,9 @@ const commitSummaryWidthConfigKey: string = 'commit-summary-width'
 const defaultStashedFilesWidth: number = 250
 const stashedFilesWidthConfigKey: string = 'stashed-files-width'
 
+const defaultPullRequestFileListWidth: number = 250
+const pullRequestFileListConfigKey: string = 'stashed-files-width'
```
sergiou87:
```suggestion
const pullRequestFileListConfigKey: string = 'pull-request-files-width'
```

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -89,8 +121,112 @@ export class PullRequestFilesChanged extends React.Component<
...
+      { type: 'separator' },
+      {
+        label: CopyFilePathLabel,
+        action: () => clipboard.writeText(fullPath),
+      },
+      {
+        label: CopyRelativeFilePathLabel,
+        action: () => clipboard.writeText(Path.normalize(file.path)),
+      },
+      { type: 'separator' },
```
sergiou87, line 215: Should we remove this separator?
```suggestion
```

sergiou87 (changes requested):
Looks great overall! Only found an issue with the key to store the file list width.

app/src/lib/stores/app-store.ts
tidy-dev: Oomph.. bad copy/pasta. Thanks

app/src/ui/open-pull-request/pull-request-files-changed.tsx
tidy-dev: Sort of.. Actually I needed to add one more context menu here so that it will match the history context menu

sergiou87 (approved):
Thanks for those changes 🙌  :shipit:

PR #15326, opened 2022-09-21
PR View: Files changed header

tidy-dev:
## Description
This PR adds the header for the "Showing changes from: x commits" and where the diff config icon will live.

### Screenshots

image
## Release notes

Notes: no-notes

### DIFF
--- a/app/styles/ui/_pull-request-diff.scss
+++ b/app/styles/ui/_pull-request-diff.scss
@@ -3,7 +3,7 @@
   border-radius: var(--border-radius);
 
   .files-changed-header {
-    padding: var(--spacing-half);
+    padding: var(--spacing);
     border-bottom: var(--base-border);
   }
 
--- a/app/styles/ui/_pull-request-diff.scss
+++ b/app/styles/ui/_pull-request-diff.scss
@@ -1,4 +1,14 @@
-.pull-request-diff-viewer {
-  height: 500px;
-  display: flex;
+.pull-request-files-changed {
+  border: var(--base-border);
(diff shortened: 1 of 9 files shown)


app/styles/ui/_pull-request-diff.scss
```diff
@@ -1,4 +1,14 @@
-.pull-request-diff-viewer {
-  height: 500px;
-  display: flex;
+.pull-request-files-changed {
+  border: var(--base-border);
+  border-radius: var(--border-radius);
+
+  .files-changed-header {
+    padding: var(--spacing-half);
```
tidy-dev: Based on the [figma](https://www.figma.com/file/PUqR1Esg6fVZbg1sa8vsko/Pull-Request-Exploration?node-id=5540%3A22270), this heading was meant to be narrow I think to try not to take more vertical space than needed. Thus, I used ```--spacing-half```, but realized it makes the text in this header not line up with other text in the dialog.
Screen Shot 2022-09-21 at 12 23 21 PM



Then I tried --spacing-half for just vertical padding and it feels a little off to me. 
``` var(--spacing-half) var(--spacing) ```
image


Thus, I think I like regular padding, but didn't know if it felt like too much to others?

```var(--spacing)```
image

gavinmn: I think using the regular padding (final screenshot) makes the most sense. It better visually separates it as a heading rather than letting it blend in as a file.

sergiou87 (approved):
Great work!! 🤩 

:shipit:

PR #15339, opened 2022-09-22
PR View: Popover Dropdown component

tidy-dev:
Based on #15326 

## Description
This PR separates out the popover dropdown logic of the branch select component into a new component that I can reuse for the commit list popover.  

I also updated the styles to match the figma better.. however, based on our accessibility meeting recently, I figured the link look would be recommended to be underlined to get the required contrast to be accessible, so I added that. But, now that I see it, I think that the design of this component should be revisited. Maybe since this is a button, it should look like a button. It looks like a button on dotcom (when editable) and we do have a dropdown button to context menu elsewhere in our app (repository add button). But, that I will leave to another PR, this PR is to add the dropdown to popover functionality so I can add the commit selection dropdown. 

cc: @gavinmn

### Screenshots

https://user-images.githubusercontent.com/75402236/191793871-4f24015e-9222-4e41-b88a-630a91c8106c.mov

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/branches/branch-select.tsx
+++ b/app/src/ui/branches/branch-select.tsx
@@ -1,18 +1,11 @@
 import * as React from 'react'
 import { IMatches } from '../../lib/fuzzy-find'
 import { Branch } from '../../models/branch'
-import { Button } from '../lib/button'
 import { ClickSource } from '../lib/list'
-import { Popover } from '../lib/popover'
-import { Ref } from '../lib/ref'
-import { Octicon } from '../octicons'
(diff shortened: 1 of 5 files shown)


gavinmn:
Looking good.

> But, now that I see it, I think that the design of this component should be revisited. Maybe since this is a button, it should look like a button. It looks like a button on dotcom (when editable) and we do have a dropdown button to context menu elsewhere in our app (repository add button).

I agree that we should change this and we can introduce it in a separate PR.

sergiou87 (approved):
Looks great! 🤩 :shipit:

PR #15342, opened 2022-09-22
PR View: Diff options

tidy-dev:
## Description
Add diff options to the diffing view. 

As discussed, this pr was changed from following convention of changing the split/united app state in favor of managing it at the files changed component state. (Pass in current app state - prevents changes happening behind dialog)

### Screenshots
image

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -352,7 +352,8 @@ const hideWhitespaceInChangesDiffKey = 'hide-whitespace-in-changes-diff'
 const hideWhitespaceInHistoryDiffDefault = false
 const hideWhitespaceInHistoryDiffKey = 'hide-whitespace-in-diff'
 const hideWhitespaceInPullRequestDiffDefault = false
-const hideWhitespaceInPullRequestDiffKey = 'hide-whitespace-in-pull-request-diff'
+const hideWhitespaceInPullRequestDiffKey =
+  'hide-whitespace-in-pull-request-diff'
 
 const commitSpellcheckEnabledDefault = true
 const commitSpellcheckEnabledKey = 'commit-spellcheck-enabled'
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7242,23 +7242,16 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
 
     const { allBranches, recentBranches } = branchesState
-    const {
-      imageDiffType,
-      hideWhitespaceInHistoryDiff,
-      showSideBySideDiff,
-      selectedExternalEditor,
-    } = this.getState()
+    const { imageDiffType, selectedExternalEditor } = this.getState()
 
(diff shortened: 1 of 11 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -351,6 +351,8 @@ const hideWhitespaceInChangesDiffDefault = false
 const hideWhitespaceInChangesDiffKey = 'hide-whitespace-in-changes-diff'
 const hideWhitespaceInHistoryDiffDefault = false
 const hideWhitespaceInHistoryDiffKey = 'hide-whitespace-in-diff'
+const hideWhitespaceInPullRequestDiffDefault = false
+const hideWhitespaceInPullRequestDiffKey = 'hide-whitespace-in-diff'
```
sergiou87:
```suggestion
const hideWhitespaceInPullRequestDiffKey = 'hide-whitespace-in-pull-request-diff'
```

app/src/ui/diff/diff-options.tsx
```diff
@@ -4,14 +4,13 @@ import { Octicon } from '../octicons'
...
-import { RepositorySectionTab } from '../../lib/app-state'
 
 interface IDiffOptionsProps {
-  readonly sourceTab: RepositorySectionTab
+  readonly isInteractiveDiff: boolean
   readonly hideWhitespaceChanges: boolean
   readonly onHideWhitespaceChangesChanged: (
     hideWhitespaceChanges: boolean
-  ) => Promise<void>
+  ) => Promise<void> | void
```
sergiou87: This is odd 🤔  I think this should just return `void`, since nothing else is awaiting for this, no?
```suggestion
  ) => void
```

sergiou87 (changes requested):

app/src/lib/stores/app-store.ts
tidy-dev: 🤦

app/src/ui/diff/diff-options.tsx
tidy-dev: Yeah. I am not sure why this is set up as promise in diff options and it required a promise to be returned. I will see if there is a cleaner way to handle it.

tidy-dev: Traced it back to the checkbox originating it and I can't find a reason that it was a Promise. We are not waiting on that to return before doing anything. It seems to work just fine too (reason I didn't change it was I didn't want to impact the changes/history view).

app/src/lib/stores/app-store.ts
tidy-dev: Random thoughts.. It would be totally cool if after accepting suggestions, I could press a button (or automatically run) on a PR to run some prettifier setup for the repo so I don't have to bounce back to local dev environment to run it. :D

sergiou87 (approved):
Looks good now!! (sorry about the linter thing 😅 )

:shipit:

PR #15377, opened 2022-09-28
PR View: No changes message

tidy-dev:
## Description
This PR adds a message for when there are no changes between the current branch and the selected base branch.

Inspiration was drawn from dotcom's no changes state. But, the context there is that it is an arbitrary comparison where our context is encouraging a pull request into your repositories default branch. Thus, I think using "compare" words are too broad. But, I am completely open to verbiage (or any other design change) to communicate the best here. 

Other thoughts:
-  "There isn't anything to merge" .. but merge is not the only way to submit a PR. 

image

### Screenshots
image

## Release notes
Notes: no-notes


sergiou87 (approved):
Looks great! I agree with the change of wording you suggest 💯

PR #15391, opened 2022-09-30
Add auto formatting

tidy-dev:
## Description
This PR adds a GitHub Action that will run `yarn prettier --write` and then commit the changes. This will happen on pull request and on pushes.

The reason behind wanting this is to automate a common workflow of after someone suggest a change on a pull request and it is committed, that change unfortunately fails the linter check, and then, you have to navigate to that PR locally just to run prettier and commit those changes and push them. 

Additionally, we sometimes get external PR that fail the linter that this would prevent the "ask to fix the linter" comment. 

## Release notes
Notes: no-notes

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -1,4 +1,5 @@
-name: Auto-format with Prettier
+name: Auto-format
+
 on:
   push:
     branches: ['development']
@@ -18,10 +19,10 @@ jobs:
           node-version: '16.13.0'
       - name: Install and build dependencies
         run: yarn
-      - name: Run prettier
-        run: yarn prettier-write
-      - name: Commit changes
+      - name: Run Lint fix
+        run: yarn lint:fix
+      - name: Commit linter changes
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          commit_message: Apply prettier formatting changes
+          commit_message: Apply auto-formatting changes
           branch: ${{ github.head_ref }}
--- a/.github/workflows/prettify.yml
+++ b/.github/workflows/prettify.yml
@@ -1,0 +1,27 @@
+name: Auto-format with Prettier
+on:
+  push:
+    branches: ['development']
+  pull_request:
+    branches: ['development']
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.13.0'
+      - name: Install and build dependencies
+        run: yarn
+      - name: Run prettier
+        run: yarn prettier-write
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Apply prettier formatting changes
+          branch: ${{ github.head_ref }}
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "lint:fix": "yarn prettier --write && yarn lint:src:fix",
     "markdownlint": "markdownlint **/*.{md,mdx} --config .markdownlint.js --rules node_modules/@github/markdownlint-github --ignore node_modules --ignore gemoji",
     "prettier": "prettier --check \"./**/*.{ts,tsx,js,json,jsx,scss,html,yaml,yml}\"",
-    "prettier-write": "prettier --write \"./**/*.{ts,tsx,js,json,jsx,scss,html,yaml,yml}\"",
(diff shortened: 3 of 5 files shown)


.github/workflows/prettify.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.13.0'
+      - name: Install and build dependencies
+        run: yarn
+      - name: Run prettier
+        run: yarn prettier-write
```
tidy-dev: We run `prettier --check `as our linter step `"lint": "yarn prettier && yarn lint:src",` in the ci.yml.  Would it make sense to remove that from the linter now? Would it make sense to just make this a step in the ci.yml?

sergiou87: Good question… I guess it makes sense to stop using it in our CI jobs, since we're fixing it anyway 🤔

package.json
```diff
@@ -27,6 +27,7 @@
     "lint:fix": "yarn prettier --write && yarn lint:src:fix",
     "markdownlint": "markdownlint **/*.{md,mdx} --config .markdownlint.js --rules node_modules/@github/markdownlint-github --ignore node_modules --ignore gemoji",
     "prettier": "prettier --check \"./**/*.{ts,tsx,js,json,jsx,scss,html,yaml,yml}\"",
+    "prettier-write": "prettier --write \"./**/*.{ts,tsx,js,json,jsx,scss,html,yaml,yml}\"",
```
sergiou87: Can't we just use `yarn lint:fix` instead? 🤔

tidy-dev: Good idea! I didn't realize that was there.. I usually just fix these by resaving the offending file locally.

.github/workflows/prettify.yml
tidy-dev: Just curious what happens when the `yarn lint:fix` fails. I removed a eslint rule disable line (lint:fix can fix that :D).  It [behaves as we would expect](https://github.com/desktop/desktop/pull/15391/checks?sha=ea571aa6ab2551d84db657e2c9e41de01ea260ac) and causes that action to fail  (meaning checks fail/stopping merge).

tidy-dev: Thus.. I think it will work to remove the linter step because if the `yarn lint:fix` errors due to a non-fixable error, it will have the same result as the `yarn lint` does now.

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,28 @@
+name: Auto-format
+
+on:
+  push:
+    branches: ['development']
+  pull_request:
+    branches: ['development']
```
sergiou87, line 7: Shouldn't this be for all branches and only in PRs? I also wonder if this will get too annoying if you're working on a draft PR and start needing to pull commits, but I guess we'll find out 😂 
```suggestion
  pull_request:
    branches: *
```

tidy-dev: Ok. I really have no idea how this part of the yaml works. But,  it worked on my fork with a branch that wasn't development. Does what you suggested mean a pull request into any branch and mine say a pull request into only development? Also,  does pull_request mean it runs anytime there is a change to a pull request.. I was thinking the "Push" was there to also run on new pushes to a pull request?

sergiou87: Basically what you had means (if I'm not wrong) that this job will run for every commit pushed to `development`, and for every PR created that target `development` (and any push to those PRs' branches). If you create PRs that sit on top of other PRs (like we usually do), those won't trigger this job.

Looking at [ci.yml](https://github.com/desktop/desktop/blob/426c834a117311f255cc303de28ac9c041c1242b/.github/workflows/ci.yml#L3-L9) I think we only need `pull_request:` (with no `branches:` line at all) and that's it.

sergiou87:
Just a thought: if we'd rather not have something 100% automatic and still want some control, we could also use a bot that reacts to comments in a PR to trigger the CI job that fixes the lint errors.

Jimimaku:
> ## Description
> This PR adds a GitHub Action that will run `yarn prettier --write` and then commit the changes. This will happen on pull request and on pushes.
> 
> The reason behind wanting this is to automate a common workflow of after someone suggest a change on a pull request and it is committed, that change unfortunately fails the linter check, and then, you have to navigate to that PR locally just to run prettier and commit those changes and push them.
> 
> Additionally, we sometimes get external PR that fail the linter that this would prevent the "ask to fix the linter" comment.
> 
> ## Release notes
> Notes: no-notes

tidy-dev:
Haven't been able to make time to get back to this. Going to close for now, I will reopen when I get a chance to revisit

PR #15410, opened 2022-10-04
PR View: Mergeability checking

tidy-dev:
## Description
Adds mergeability checking and messaging to PR View dialog. The messaging was copied from dotcom's mergeability message component so that it would match when you are Desktop and when you finish opening it on dotcom. The colors are colors existing in Desktop and do not match dotcom identically. 

Other notes: 
- I re-used the mergability checks from the compare branch feature. In doing so, I pulled out part of the logic into a method to reuse (also switched the finally logic to use finally). It "shouldn't" impact the compare branch feature and I tested it out...but when you change something there is chance for impact. 
- For the invalid state for compare branch feature, we say something about not being able to merge unrelated histories and for here I put the copy from dotcom which is basically that an error happened .. I don't know how to make an error happen that also happens on dotcom... thus, I haven't tested that path yet. Edit: figured out how to make an orphaned branch ([see below comment](https://github.com/desktop/desktop/pull/15410#issuecomment-1267451078))

### Screenshots

Clean:
Screen Shot 2022-10-04 at 1 52 33 PM

Loading:
Screen Shot 2022-10-04 at 2 25 38 PM

Merge conflicts exist:
Screen Shot 2022-10-04 at 2 01 47 PM

Error:
Screen Shot 2022-10-04 at 2 01 24 PM

## Release notes
Notes: [Improved] Add mergability info and warnings to pull request preview.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -1442,34 +1442,21 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
 
     if (tip.kind === TipState.Valid && aheadBehind.behind > 0) {
-      const mergeTreePromise = promiseWithMinimumTimeout(
-        () => determineMergeability(repository, tip.branch, action.branch),
-        500
-      )
-        .catch(err => {
-          log.warn(
-            `Error occurred while trying to merge ${tip.branch.name} (${tip.branch.tip.sha}) and ${action.branch.name} (${action.branch.tip.sha})`,
-            err
-          )
-          return null
-        })
-        .then(mergeStatus => {
+      this.currentMergeTreePromise = this.setupMergabilityPromise(
+        repository,
+        tip.branch,
+        action.branch,
+        mergeStatus => {
           this.repositoryStateCache.updateCompareState(repository, () => ({
             mergeStatus,
           }))
 
           this.emitUpdate()
-        })
-
-      const cleanup = () => {
-        this.currentMergeTreePromise = null
-      }
-
-      // TODO: when we have Promise.prototype.finally available we
-      //       should use that here to make this intent clearer
-      mergeTreePromise.then(cleanup, cleanup)
-
-      this.currentMergeTreePromise = mergeTreePromise
+        },
+        () => {
+          this.currentMergeTreePromise = null
+        }
+      )
 
       return this.currentMergeTreePromise
     } else {
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7254,18 +7254,18 @@ export class AppStore extends TypedBaseStore<IAppState> {
       currentBranch
     )
 
-    const commitSHAs = pullRequestCommits.map(c => c.sha)
+    const commitsBetweenBranches = pullRequestCommits.map(c => c.sha)
 
     // A user may compare two branches with no changes between them.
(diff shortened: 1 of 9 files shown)


tidy-dev:
Alright.. No idea what the `Error` state is for dotcom but unrelated histories is handled like:
image

1) Do we want to follow suit or make a new mergability warning at the bottom that says what our compare branch tool does combined with our [current empty state](https://github.com/desktop/desktop/pull/15377)?

image


2) And do we want to disable the "Create pull request" button.. or allow the user to open this on dotcom? 
 -> Seeing as the create pull request button hopefully will actually create a pull request at some point -> I think it make sense just to disable it (but can see it either way).


(Any which way this can be handled in a follow up PR as it should be pretty rare)

cc: @gavinmn

app/src/lib/stores/app-store.ts
```diff
@@ -1476,6 +1463,32 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+          err
+        )
+        return null
+      })
+      .then(mergeStatus => {
+        onLoad(mergeStatus)
+      })
+      .finally(() => {
+        cleanup?.()
+      })
```
niik: This feels a little odd to me given that we're already returning a promise. The caller of this could tack on `.finally()` on the returned promise instead which would feel better to me given that the caller is the one introducing the side effect which needs to be reset.

Same goes for the `onLoad` callback, I think it would be more idiomatic to tack on a `.then` on the returned promise and have the `setupMergeabilityPromise` only concern itself with the reusable parts of promiseWithMinimumTimeout and error management.

niik (changes requested):
One small note

gavinmn:
> Do we want to follow suit or make a new mergability warning at the bottom that says what our compare branch tool does combined with our current empty state

I think we should use this existing string `Unable to merge unrelated histories in this repository`

> And do we want to disable the "Create pull request" button.. or allow the user to open this on dotcom?
-> Seeing as the create pull request button hopefully will actually create a pull request at some point -> I think it make sense just to disable it (but can see it either way).
(Any which way this can be handled in a follow up PR as it should be pretty rare)

Agree — lets disable the button

app/src/lib/stores/app-store.ts
tidy-dev: Yep. That totally makes sense. I originally pulled out a lot more logic that was in that method than whittled it down to return the promise.. as soon as I read this I was like.. should have whittled it down farther. :D

app/src/lib/stores/app-store.ts
```diff
@@ -7263,7 +7263,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
         shas: commitSHAs,
         shasInDiff: commitSHAs,
         isContiguous: true,
-        changesetData,
+        changesetData: changesetData !== null ? changesetData : emptyChangeSet,
```
niik:
```suggestion
        changesetData: changesetData ?? emptyChangeSet,
```

niik (dismissed):
Looks great, one small stylistic suggestion

tidy-dev:
Went ahead and added the unrelated histories error message in that our compare feature has. However, it seemed strange for our no changes view to say one branch was "up to date" with the other in this scenario, so I made the no changes view say the same as dotcom when in this state.

image

niik (approved):

PR #15427, opened 2022-10-07
PR View: Responsive dialogs

tidy-dev:
## Description
This PR makes the width of the pr dialog responsive. Then, it calculate the height of the diff container based on the height of app window.

### Screenshots

Recorded on Windows VM because windows had more clipping issues than mac.

https://user-images.githubusercontent.com/75402236/196246098-724daa42-e02b-4cf3-9c06-7476658062ce.mov

## Release notes
Notes: [Improved] Pull request preview dialog width and height is responsive.

### DIFF
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -112,7 +112,8 @@ interface IDialogProps {
    * This is used to make a dialog width's responsive.
    *
    * On window resize, the dialog will adjust to fill the app window width sans
-   * padding within the constrains of the this max width and a css defined min.
+   * padding within the constrains of the this max width and a css defined
+   * minimum.
    */
   readonly maxWidth?: number
 }
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -111,8 +111,8 @@ interface IDialogProps {
    * An optional max width for the dialog.
    * This is used to make a dialog width's responsive.
    *
-   * On window resize, the dialog will adjust to fill the app window width sans padding within the
-   * constrains of the min and max.
+   * On window resize, the dialog will adjust to fill the app window width sans
+   * padding within the constrains of the this max width and a css defined min.
    */
   readonly maxWidth?: number
 }
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -3,8 +3,8 @@ import classNames from 'classnames'
 import { DialogHeader } from './header'
 import { createUniqueId, releaseUniqueId } from '../lib/id-pool'
 import { getTitleBarHeight } from '../window/title-bar'
-import { IConstrainedValue } from '../../lib/app-state'
 import { clamp } from '../../lib/clamp'
+import { infinity } from '../octicons/octicons.generated'
 
 /**
  * The time (in milliseconds) from when the dialog is mounted
@@ -108,14 +108,13 @@ interface IDialogProps {
   readonly loading?: boolean
 
   /**
-   * An option width for the dialog given as value, min and max.
+   * An optional max width for the dialog.
    * This is used to make a dialog width's responsive.
    *
-   * On component mounting, it will ge set at the value. On window resize, the
-   * dialog will adjust to fill the app window width sans padding within the
+   * On window resize, the dialog will adjust to fill the app window width sans padding within the
    * constrains of the min and max.
    */
-  readonly width?: IConstrainedValue
+  readonly maxWidth?: number
 }
 
 interface IDialogState {
@@ -192,10 +191,9 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
       return
     }
 
-    if (this.props.width !== undefined) {
-      const { min, max } = this.props.width
+    if (this.props.maxWidth !== undefined) {
       const widthSansPadding = window.innerWidth - 40
-      const newWidth = clamp(widthSansPadding, min, max)
+      const newWidth = clamp(widthSansPadding, -infinity, this.props.maxWidth)
       this.dialogElement.style.width = `${newWidth}px`
     }
 
@@ -263,10 +261,6 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
   public componentDidMount() {
     if (!this.dialogElement) {
       return
-    }
-
-    if (this.props.width !== undefined) {
-      this.dialogElement.style.width = `${this.props.width.value}px`
     }
 
     this.dialogElement.showModal()
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -3,6 +3,8 @@ import classNames from 'classnames'
 import { DialogHeader } from './header'
 import { createUniqueId, releaseUniqueId } from '../lib/id-pool'
 import { getTitleBarHeight } from '../window/title-bar'
+import { IConstrainedValue } from '../../lib/app-state'
+import { clamp } from '../../lib/clamp'
 
 /**
  * The time (in milliseconds) from when the dialog is mounted
@@ -104,6 +106,16 @@ interface IDialogProps {
    * of the loading operation.
    */
   readonly loading?: boolean
+
+  /**
+   * An option width for the dialog given as value, min and max.
+   * This is used to make a dialog width's responsive.
+   *
+   * On component mounting, it will ge set at the value. On window resize, the
+   * dialog will adjust to fill the app window width sans padding within the
+   * constrains of the min and max.
+   */
+  readonly width?: IConstrainedValue
 }
 
 interface IDialogState {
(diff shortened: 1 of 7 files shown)


app/src/ui/open-pull-request/open-pull-request-dialog.tsx
```diff
@@ -208,6 +208,7 @@ export class OpenPullRequestDialog extends React.Component<IOpenPullRequestDialo
         className="open-pull-request"
         onSubmit={this.onCreatePullRequest}
         onDismissed={this.props.onDismissed}
+        width={{ value: 900, max: 1100, min: 850 }}
```
tidy-dev, line 211: > max: 1100

Arbitrary... but no idea what we would make since as a maximum.

> min: 850

Not sure about this either.. 850 was chosen based on min window with being 960 - 100 for some margins. However, I am inclined to put this as -infinity as the logic should bear this out seeing as the resize function won't be called for less the the app width.

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -70,12 +70,71 @@ export class PullRequestFilesChanged extends React.Component<
...
+  }
+
+  private getPullRequestFooterHeight() {
+    const query = document.querySelectorAll('.open-pull-request .dialog-footer')
+    const elHeight = query.length > 0 ? query.item(0).clientHeight : 0
+    return elHeight > 0 ? elHeight : 67
+  }
+
+  private getDiffContainerHeightOffset() {
+    const dialogMargin = 100
```
tidy-dev: I do not know exactly why.. but need at least 80 or the footer starts clipping

app/src/ui/dialog/dialog.tsx
```diff
@@ -104,6 +106,15 @@ interface IDialogProps {
...
   readonly loading?: boolean
+
+  /**
+   * An optional max width for the dialog.
+   * This is used to make a dialog width's responsive.
+   *
+   * On window resize, the dialog will adjust to fill the app window width sans
+   * padding within the constrains of the this max width and a css defined min.
+   */
+  readonly maxWidth?: number
```
tidy-dev, line 118: I originally had an IContrainValue, but determined it was useless. :P We need to have a minimum as low as the css set minimum of 400 to support being able to close a dialog when zoomed in all the way. Therefore, the maxWidth is the only determining factor.

app/src/ui/open-pull-request/open-pull-request-dialog.tsx
```diff
@@ -208,6 +208,7 @@ export class OpenPullRequestDialog extends React.Component<IOpenPullRequestDialo
         className="open-pull-request"
         onSubmit={this.onCreatePullRequest}
         onDismissed={this.props.onDismissed}
+        maxWidth={2000}
```
tidy-dev: Arbitrary choice based on something that looked ok my monitor. :D Any feedback regarding a way to determine a good max width? Should we just go with infinity?

tidy-dev: Chose infinity for PR modal - since with other diffing we stretch for whole app.

app/src/ui/dialog/dialog.tsx
```diff
@@ -3,6 +3,8 @@ import classNames from 'classnames'
 import { DialogHeader } from './header'
 import { createUniqueId, releaseUniqueId } from '../lib/id-pool'
 import { getTitleBarHeight } from '../window/title-bar'
+import { clamp } from '../../lib/clamp'
+import { infinity } from '../octicons/octicons.generated'
```
sergiou87: I think this is not the `infinity` you were looking for

These are not the droids you're looking for

app/src/ui/dialog/dialog.tsx
```diff
@@ -180,6 +192,12 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
       return
     }
 
+    if (this.props.maxWidth !== undefined) {
+      const widthSansPadding = window.innerWidth - 40
```
sergiou87: Is that 40 specified/used somewhere else? Just making sure we reuse the constant somehow instead of hardcoding it multiple times.

app/src/ui/dialog/dialog.tsx
```diff
@@ -180,6 +192,12 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
...
+      const newWidth = clamp(widthSansPadding, -infinity, this.props.maxWidth)
```
sergiou87: I'm surprised this builds with the `infinity` Octicon 🤯 
```suggestion
      const newWidth = clamp(widthSansPadding, -Infinity, this.props.maxWidth)
```

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -70,12 +70,72 @@ export class PullRequestFilesChanged extends React.Component<
...
+    return (
+      this.getPullRequestHeaderHeight() +
+      this.getPullRequestFillDiffHeaderHeight() +
+      this.getPullRequestFooterHeight() +
+      dialogMargin +
+      diffMargin
+    )
+  }
+
+  private setfileDiffContainerHeight = () => {
```
sergiou87:
```suggestion
  private setFileDiffContainerHeight = () => {
```

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -70,12 +70,72 @@ export class PullRequestFilesChanged extends React.Component<
...
+
   public constructor(props: IPullRequestFilesChangedProps) {
     super(props)
 
     this.state = { showSideBySideDiff: props.showSideBySideDiff }
   }
 
+  public componentDidMount() {
+    this.setfileDiffContainerHeight()
+    window.addEventListener('resize', this.setfileDiffContainerHeight)
```
sergiou87:
```suggestion
    this.setFileDiffContainerHeight()
    window.addEventListener('resize', this.setFileDiffContainerHeight)
```

app/src/ui/open-pull-request/pull-request-files-changed.tsx
```diff
@@ -70,12 +70,72 @@ export class PullRequestFilesChanged extends React.Component<
...
+  }
+
+  public componentWillUnmount() {
+    window.removeEventListener('resize', this.setfileDiffContainerHeight)
```
sergiou87:
```suggestion
    window.removeEventListener('resize', this.setFileDiffContainerHeight)
```

sergiou87 (changes requested):
Looks good overall! 

Although I have to say that so many hardcoded heights/widths/paddings make me a bit nervous, thinking about the day we need to start moving things around, if those numbers are duplicated some of our CSS or whatever. But I have no idea if there's a better way of share those values, or even if it's worth doing so.

Anyway, I don't intend to block the PR on that, it's just food for thought 😄

app/src/ui/dialog/dialog.tsx
tidy-dev: Lol. And it works..

app/src/ui/dialog/dialog.tsx
tidy-dev: haha. Yes. Surprised it worked..

app/src/ui/dialog/dialog.tsx
tidy-dev: Made a constant

tidy-dev:
I retested on mac os (regular and full screen mode) and windows vm and everything still looks good with recent changes.

app/src/ui/open-pull-request/pull-request-files-changed.tsx
tidy-dev: Figured out that the app title bar needed to be part of the calculation as well because the dialog height is limited to the `desktop-app-contents` which is the `window.innerHeight` minus the `desktop-app-title-bar` height. Therefore, the dialog will always be guaranteed a minimum margin of the height of the title-bar since the dialog is absolutely positioned and margin set to auto to center it.

app/src/ui/lib/spacial-constants.ts
```diff
@@ -0,0 +1,22 @@
+/*
+These constants are put in place to mirror the styles in our variables.scss
+so we have centralized place to update them if the css would change.
+
+NB - Preferably we should dynamically find heights and widths of html elements
+when possible to avoid the need to use static heights and widths.
+*/
+
+/** The app width for standard padding as defined in the 'variables.sccs' */
+export const spacing = 10
```
sergiou87, line 10: The next step would be having _something_ sharing this directly with our CSS (some kind of variable replacement like what we do with `__DARWIN__` and such), but that's a bit more complicated.

sergiou87 (approved):
LGTM!! You might want Markus' input on this given he's much more knowledgeable about CSS tricks and, of course, the Desktop codebase.

app/src/ui/lib/spacial-constants.ts
tidy-dev: Yeah.. I thought about that, and also feel like that is a better use of time that if we get to the point where we find we actually need to change these to implement something more sophisticated and not over engineer for a what if that feels small right now.

niik:
:wave: Sorry for not getting to this sooner. I tried building this and the PR dialog height didn't change when resized the window which I thought was what we wanted it to do? Either way I am vary of adding this much layout logic outside of CSS. There are a few places in the codebase (mostly related to virtualised lists) where we have to (or have had to historically at least, there are plenty new tricks around like ResizeObserver, Grid, etc that may render even those places obsolete in the future) but generally they are painful to maintain and easy to get wrong.

If what we're trying to achieve here (and I suspect this is where I'm missing some context) is a dialog that fills the window (sans some padding) when it resizes then I think we can accomplish that with just CSS. I took a quick stab at a proof of concept here: https://github.com/desktop/desktop/compare/dynamic-size-pr-dialog

Sorry if I'm just wasting y'alls time here because I'm missing what we're aiming for.

PR #15444, opened 2022-10-11
Enable "experimental" diff viewer for unified diffs

sergiou87:
## Description

This PR enables **in beta builds** the new diff viewer used for side-by-side diffs in unified diffs, replacing CodeMirror. This is something we intended to do for a long time, since we expect our implementation to be easier to make accessible, but we were blocked by our inability to select text properly in long files. Something that was solved by @niik in #15128 

In this PR, aside from enabling it, I made a few CSS changes to ensure it looks as good as the current version. I also found a bug introduced in #15128 regarding which `diff` to use when copying the text (`props.diff` vs `state.diff`… not sure how it worked to begin with 🤔 because the `diff` from `props` is the original diff, while the one from `state` is the expanded diff -or the original diff if it was never expanded by the user).

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/ui/diff/side-by-side-diff.tsx
+++ b/app/src/ui/diff/side-by-side-diff.tsx
@@ -256,7 +256,7 @@ export class SideBySideDiff extends React.Component<
         : [DiffLineType.Add, DiffLineType.Context]
       : [DiffLineType.Add, DiffLineType.Delete, DiffLineType.Context]
 
-    const contents = this.props.diff.hunks
+    const contents = this.state.diff.hunks
       .flatMap(h =>
         h.lines
           .filter(line => lineTypes.includes(line.type))
--- a/app/styles/ui/_side-by-side-diff.scss
+++ b/app/styles/ui/_side-by-side-diff.scss
@@ -338,7 +338,7 @@
           border-color: var(--diff-hunk-gutter-background-color);
           align-self: stretch;
           align-items: center;
-          
+
           &.selectable:hover {
             border-color: var(--diff-hover-background-color);
           }
--- a/app/styles/ui/_side-by-side-diff.scss
+++ b/app/styles/ui/_side-by-side-diff.scss
@@ -322,13 +322,22 @@
         flex-direction: row;
       }
 
+
       .hunk-handle {
         left: 100px;
       }
 
-      &.hunk-info .line-number {
-        background: var(--diff-hunk-gutter-background-color);
-        border-color: var(--diff-hunk-border-color);
+      &.hunk-info {
+        .line-number {
+          background: var(--diff-hunk-gutter-background-color);
+          border-color: var(--diff-hunk-border-color);
+        }
+        .hunk-expansion-handle {
+          background: var(--diff-hunk-gutter-background-color);
+          border-right-width: 4px;
+          border-right-style: solid;
+          border-color: var(--diff-hunk-gutter-background-color);
+        }
       }
 
       .line-number {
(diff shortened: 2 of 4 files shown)


app/styles/ui/_side-by-side-diff.scss
```diff
@@ -322,13 +322,32 @@
...
+          border-color: var(--diff-hunk-gutter-background-color);
+          &.selectable:hover {
+            border-color: var(--diff-hover-background-color);
+          }
+          &:not(.selectable) span {
+            // This must be the height of an octicon, so that when a hunk handle
+            // of any type is replaced by an "empty" hunk handle (e.g. when a
+            // file is completely expanded), the height of the whole line
+            // remains the same.
+            height: 16px;
```
sergiou87: Feels like a hack, but couldn't find a better solution with my current CSS skills 🤣 😭

app/styles/ui/_side-by-side-diff.scss
```diff
@@ -326,9 +326,27 @@
...
+          }
+        }
```
tidy-dev:
```suggestion
        .hunk-expansion-handle {
          background: var(--diff-hunk-gutter-background-color);
          border-right-width: 1px;
          border-right-style: solid;
          border-color: var(--diff-hunk-gutter-background-color);
          align-self: stretch;
          align-items: center;
          
          &.selectable:hover {
            border-color: var(--diff-hover-background-color);
          }
        }
```

I think this will do what you need... I tested in the case of an expanded file and not expanded file and I think it behaves correctly.

tidy-dev: The `align-self: stretch` makes it take up it's parents height  and then the `align-items: center` centers the span inside like div's parent was doing to this div to achieve centering.

tidy-dev: (And I think this works because the parent is a flex box itself... but I am a little fuzzy on that).

app/src/ui/diff/side-by-side-diff.tsx
```diff
@@ -256,7 +256,7 @@ export class SideBySideDiff extends React.Component<
         : [DiffLineType.Add, DiffLineType.Context]
       : [DiffLineType.Add, DiffLineType.Delete, DiffLineType.Context]
 
-    const contents = this.props.diff.hunks
+    const contents = this.state.diff.hunks
```
tidy-dev, line 259: Just curious why this changed? (It works, I just didn't understand).

tidy-dev (dismissed):
:sparkles: Looks good to me and it tested out well on my mac.

I added a suggestion for a way to avoid the octicon 16px usage. But, also didn't want to block this on account of that. :)

app/src/ui/diff/side-by-side-diff.tsx
sergiou87: The diff from the `props` is the original diff, while the diff from the `state` is the original + extended hunks, so it's the one that must be used.

app/styles/ui/_side-by-side-diff.scss
sergiou87: Thank you!! It's not perfect as the CodeMirror behavior (it still gets a bit shorter), but it's the same behavior displayed in the split diff. We can address that "glitchy" behavior in a separate PR.

tidy-dev (approved):

sybrew:
Does this PR address the middle-mouse issue of #15675? I'm using 3.1.6, and I still face the issue.

sergiou87:
I'm sorry @sybrew , this is only in beta so far. There is no ETA of when this will hit production, as it's a big change in the app (I mean replacing the old diff viewer "engine" with the new one).

PR #15458, opened 2022-10-14
Add shortcut to toggle Diff Display mode and Hide Whitespace Changes

### LINKED ISSUE #15284: Feature request: add a shortcut to change diff display mode
### The feature request

At the moment, changing the diff display mode requires three clicks:
1. Click on the gear icon
2. Select the "unselected" display mode (unified/merge)
3. Click outside this panel to close it (or click again on the previous option)

I am doing a big refactoring for which I find helpful switching between the two modes to check complex commits, and I'd really appreciate a faster way to switch this option.

### Proposed solution

Add a shortcut to toggle between "unified" and "split" diff options.

### Additional context

image

angusdev:
Closes #15284 

## Description
Add shortcut to toggle Diff Display mode (Cmd/Ctrl+D) and Hide Whitespace Changes (Cmd/Ctrl+Alt+D)

### Screenshots
toggle

toggle

## Release notes

Notes: IMPROVED Add shortcut to toggle Diff Display mode (Cmd/Ctrl+D) and Hide Whitespace Changes (Cmd/Ctrl+Alt+D) #15284

### DIFF
--- a/app/src/ui/diff/diff-options.tsx
+++ b/app/src/ui/diff/diff-options.tsx
@@ -4,6 +4,7 @@ import { Octicon } from '../octicons'
 import * as OcticonSymbol from '../octicons/octicons.generated'
 import { RadioButton } from '../lib/radio-button'
 import { Popover, PopoverCaretPosition } from '../lib/popover'
+import { getPlatformSpecificNameOrSymbolForModifier } from '../../lib/menu-item'
 
 interface IDiffOptionsProps {
   readonly isInteractiveDiff: boolean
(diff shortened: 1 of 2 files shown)


palhal:
I'd love to see this PR merged. I'm surprised this feature isn't already built-in. Can someone please review this?

niik:
Hey @angusdev, sorry for us leaving you hanging here. We chose not to implement a keyboard shortcut here because our telemetry suggest that switching between modes is very rare, i.e users typically settle into a preferred mode and rarely change it.

That said I don't have anything against a keyboard shortcut but we'd prefer to implement it via the application menu for discoverability, improved accessibility with the added benefit of making it easier for us to ensure we don't assign the same keyboard shortcut twice. I'm going to close this PR now since it's not quite the approach we're looking for but let us know if you're interesting in working on that and we can either reopen or you can open a new PR.

Again, our apologies for not getting around to this sooner!

palhal:
I rarely change mode because it requires moving the mouse and then clicking two or three times. If we had a shortcut I would change modes more often. I don't mind having it in the application menu though, as long as I can use the keyboard.

PR #15462, opened 2022-10-15
Fix incorrect tooltip position if mouse is not moved

### LINKED ISSUE #13636: Misplaced tooltip
### Describe the bug
Sometimes the tooltip appears in the top-left of the window, amusingly behind the stop-light buttons.

### Version & OS

Version 2.9.6 (arm64) on macOS 12.1 (21C52)

1. Hover over a file title

### Expected behavior

Tooltip appears in the location of the pointer.

### Actual behavior

Tooltip appears in the top-left corner.

### Screenshots

image

angusdev:
Closes #13636

## Description
The tooltip position is referenced to mouse position set in mousemove event. If the tooltip is triggered by events other than mousemove (e.g. scroll the list with mousewheel or keyboard arrow keys, or click the Fetch Origin button without moving the mouse), the mouse position is not set (0,0).  The fix is to set the mouse position in mouseenter event as well

### Screenshots
https://user-images.githubusercontent.com/34896/196004801-467ebdf9-fc8b-4730-8c14-08015a6018b2.mov

## Release notes
Notes: FIXED Tooltips are positioned properly if mouse is not moved

### DIFF
--- a/app/src/ui/lib/tooltip.tsx
+++ b/app/src/ui/lib/tooltip.tsx
@@ -299,8 +299,12 @@ export class Tooltip<T extends TooltipTarget> extends React.Component<
     }
   }
 
-  private onTargetMouseEnter = (event: MouseEvent) => {
+  private rememberMousePosition = (event: MouseEvent) => {
     this.mouseRect = new DOMRect(event.clientX - 10, event.clientY - 10, 20, 20)
+  }
+
+  private onTargetMouseEnter = (event: MouseEvent) => {
+    this.rememberMousePosition(event)
 
     this.mouseOverTarget = true
     this.cancelHideTooltip()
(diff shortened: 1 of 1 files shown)


JohSteph890 (commented):
…

app/src/ui/lib/tooltip.tsx
```diff
@@ -300,6 +300,8 @@ export class Tooltip<T extends TooltipTarget> extends React.Component<
   }
 
   private onTargetMouseEnter = (event: MouseEvent) => {
+    this.mouseRect = new DOMRect(event.clientX - 10, event.clientY - 10, 20, 20)
```
sergiou87, line 303: Maybe this can be extracted in to a function called from both `onTargetMouseEnter` and `onTargetMouseMove`?

sergiou87 (changes requested):
Wow!! Nice catch 👏

app/src/ui/lib/tooltip.tsx
angusdev: Refactor to rememberMousePosition.  Feel free to suggest a better name

sergiou87: I'd just call it `updateMouseRect`, thanks!

angusdev: Updated

sergiou87 (approved):
Thank you for that! Works like a charm 💪

PR #15468, opened 2022-10-17
Consistently disable multi commit operations

tidy-dev:
## Description
A spam comment reminded me that I had seen a todo comment that needed addressed and forgot to circle back to it earlier this year.

The bad behavior: If you started a reorder, squash, or cherry-pick operation that results in conflicts, you could attempt to squash or reorder again. Squashing menu items were enabled but did nothing and since reordering is only through drag/drop, it showed dragging/dropping cues but did nothing.  After a few tries, it did seem I might be able to start another reorder again (I got the stash warning dialog). 

This PR disables them while any multi commit operation is in progress. This also includes merge and rebase, but I would not think we would want to allow these to be starting during those either - > squash, reorder, and cherry-pick are just interactive rebases anyways.

## Release notes
Notes: [Fixed] Disable reorder, squashing, cherry-picking while an action of this type is in progress.

### DIFF
--- a/app/src/ui/history/commit-list-item.tsx
+++ b/app/src/ui/history/commit-list-item.tsx
@@ -51,8 +51,8 @@ interface ICommitProps {
   readonly showUnpushedIndicator: boolean
   readonly unpushedIndicatorTitle?: string
   readonly unpushedTags?: ReadonlyArray<string>
-  readonly isCherryPickInProgress?: boolean
   readonly disableSquashing?: boolean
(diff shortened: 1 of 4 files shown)


sergiou87 (approved):
Changes look good, I tested it and works as expected! :shipit: 🚀

GolubokUK (commented):
app/src/ui/history/commit-list.tsx

babarmeer (commented):
[README.md](https://github.com/desktop/desktop/files/10460032/README.md)

PR #15475, opened 2022-10-18
Fix repository change indicator not visible if selected and in focus

### LINKED ISSUE #7651: "changed files" indicator in repository list can be hidden with keyboard focus
## Description

This is not a regression for 1.7, but something I noticed while testing today and wanted to write up because the layout changes in #7623 make this more noticeable and cause the other indicators to appear to be floating.

## Version
* GitHub Desktop: `1.7.1-beta1`

## Steps to Reproduce
1. Add a few repositories to the app so that you can switch between then
2. Add some changes to the list so that the indicator appears (at least the "changed files" indicator, but the "ahead/behind" indicator here also helps to understand the impact)
3. Expand the repositories list and use your keyboard to move up/down until you highlight a repository that has changed files. Don't switch to it, just linger there.

### Expected Behavior
The "changed files" indicator is visible

### Actual Behavior
The "changed files" indicator is the same colour as the background when the list has keyboard focus, so it becomes invisible. And because it's after the ahead/behind indicator, it now looks misaligned.

indicator-hidden-light-theme-keyboard-focus

I think, in this situation, the indicator should be "inverted" like we do with the icon, to ensure it's visible.

When you do not have keyboard focus, the background for the list reverts to a light grey, which shows the indicator.

indicator-light-no-keyboard-focus

This is also visible when using the dark theme:

indicator-hidden-when-keyboard-focus

And here's how it looks without the keyboard focus:

indicator-no-keyboard-focus

angusdev:
Closes #7651

## Description

The repository change indicator (the blue dot) is not visible if selected and in focus

### Resolution

Use ```var(--text-color)```  (white) when selected and in focus

### Screenshots

**Light Mode - With Focus**

light

**Light Mode - Without Focus**

light-nofocus

**Dark Mode - With Focus**
dark-focus

**Dark Mode - Without Focus**

dark-nofocus

## Release notes

Notes: FIXED Fix repository change indicator not visible if selected and in focus

### DIFF
--- a/app/styles/ui/_repository-list.scss
+++ b/app/styles/ui/_repository-list.scss
@@ -193,6 +193,12 @@
           background: var(--list-item-selected-active-badge-background-color);
           color: var(--list-item-selected-active-badge-color);
         }
+
+        .change-indicator-wrapper {
+          .octicon {
+            color: var(--text-color);
+          }
+        }
       }
     }
 


sergiou87 (approved):
This is great, thank you! 🙌  I've never navigated that list with the keyboard so never noticed this bug 😳

Issue #15422, opened 2022-10-06
Notifications occur for every check if any fail


ofek:
### The problem

Notifications occur for every PR check if any fail, see https://github.com/DataDog/integrations-core/pull/13053

I think Codecov temporarily failed while results were being uploaded for every job then went green. Maybe that's happening multiple times, rapidly? I'm getting dozens of failed message pop-ups but not hundreds

### Release version

Version 3.1.1 (x64)

### Operating system

Windows 10

ofek:
> but not hundreds

Actually it is...

sergiou87:
😱😱😱

In theory that should only happen for every PR where you're the author, and only when the CI jobs run and fail. I wouldn't expect anything like what you describe unless you have hundreds of PRs and they all initiate CI jobs simultaneously (or rapidly) and all fail.

Do you think that's what happened to you?

ofek:
Nope just 1 PR w/ hundreds of checks for AZP

sergiou87:
It should only send 1 notification per check suite 😱 And from what I see, your PR should only have 1 check suite with many many check runs. Is it possible that check suite affects multiple different PRs?

ofek:
No

ofek:
Please feel free to modify https://github.com/DataDog/integrations-core/blob/master/datadog_checks_base/datadog_checks/base/__about__.py in a PR to test

ofek:
btw this is new behavior, like within the last month or 2

sergiou87:
Thanks for the info! If you've seen this recently (we haven made changes in notifications on the GitHub Desktop side for a long time), maybe the behavior changed on the backend side. I will investigate and, if I can't find the reason, I will create a test PR in that repo as you suggest. Thank you again!!

ofek:
Did that reproduce?

sergiou87:
Nope, I didn't get a notification at all 😞 Which might be a different bug/issue.

However I tried with this other one #15451 and managed to get multiple notifications for it. More specifically I got 1 notification per check suite (group of checks), per individual commit that got their checks running. In this case, there were two check suites, one with only 1 check and the other one with 4 checks:

![image](https://user-images.githubusercontent.com/1083228/195356116-589e70fd-024f-4d06-b963-a7fc667d2296.png)

I pushed 10 commits **individually** that made both of them fail, so I got 10 * 2 = 20 notifications:
[...]

PR #15487, opened 2022-10-19
Check if notification matches current repository

sergiou87:
## Description

While investigating #15422, I noticed the app isn't using the repo owner/name to decide whether or not it should show the notification 🤦‍♂️  That check probably got lost due to a previous refactor. This PR brings that check back, and takes fork behavior into account.

## Release notes

Notes: [Fixed] Notifications are shown only when they are relevant to the current repository

### DIFF
--- a/app/src/lib/stores/notifications-store.ts
+++ b/app/src/lib/stores/notifications-store.ts
@@ -2,7 +2,9 @@ import {
   Repository,
   isRepositoryWithGitHubRepository,
   RepositoryWithGitHubRepository,
+  isRepositoryWithForkedGitHubRepository,
 } from '../../models/repository'
+import { ForkContributionTarget } from '../../models/workflow-preferences'
 import { PullRequest } from '../../models/pull-request'
 import { API, APICheckConclusion } from '../api'
 import {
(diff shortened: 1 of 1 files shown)


tidy-dev:
I thought that if I opened a PR in desktop/desktop and then was working in desktop/dugite. I should get a notification about desktop/desktop PR failing so I could switch to it. Is that not so?

sergiou87:
@tidy-dev the idea was not bothering users with notifications from repositories other than the currently selected one. The notification dialog, however, has a "Switch to repository" button because you might receive the notification, and then switch to another repo before you interact with the notification. In that case, the notification will suggest you to switch to the repo it belongs to.

Does that make sense? Happy to discuss about this behavior with the team 😄

tidy-dev (approved):
✨  

> idea was not bothering users with notifications from repositories

Ah ok. I just didn't realize that is how it was intended to work.

angusdev:
`idea was not bothering users with notifications from repositories`

Is it good? For me, notification is to notify me that something I need to know but I'm not currently focusing on, and it's how it works in mobile phone.  Also there shouldn't be very frequent for normal user. Or open source maintainers will get many notifications?

sergiou87:
@angusdev yeah the intention is to have "high-signal notifications", as in things we are 99.99% sure are relevant to you. You could be in the same repo working on a different branch, and want to know about things that happen in your previous branch.

In any case, after discussing with the team, we'll gather some metrics about how many users get events from the current repo vs from other recent repos, and use that data to decide if we change this behavior in the future.

Thanks for your feedback! ❤️

PR #15494, opened 2022-10-20
Enable fetching and force-pushing more easily

### LINKED ISSUE #7805: I don't seem to have a way to force a fetch
If GitHub Desktop knows I've got remote changes, I can't seem to tell it to fetch, because the pull button replaces the fetch button:

image

and as near as I can tell Desktop doesn't have another way to fetch (at least not an intuitive one). There isn't even an option in the repository context menu:

image

When working with a large repo, I sometimes don't want to pull if I'm somewhat close to remote/master. But I don't know that if it hasn't fetched recently. In the above example, it's only been 20 minutes, but earlier today, the last fetch was over a day prior. That might have been large enough to justify a pull, but I had no way to tell since my only option was to pull.

I can manually do a command-line fetch, of course, but it would be very nice to have the option to force a fetch within GitHub desktop, even if I've got remote changes.

### LINKED ISSUE #15211: Add a "Force Push" menu item
### The feature request

After performing an action like `git reset` outside GitHub Desktop, I'd like to be able to force push the current branch.

### Proposed solution

Add a "Force Push" menu item to the Repository menu.

### Additional context

Since https://github.com/desktop/desktop/pull/13384 the easiest workaround is to right-click the latest commit in the History tab and amend the commit message. When the commit is amended, the "Push origin" button changes to "Force push origin".

sergiou87:
Closes #15211
Closes #7805

## Description

This PR makes two changes:
1. Adds a `Fetch` item to the `Repository` menu, so users can fetch at any time. This is useful if you're in a state where the `Fetch` button has been replaced by the `Pull` or `Push` button but you still need to fetch. For example, if you have a branch that hasn't been pushed yet, you want to rebase it on top of another branch that doesn't exist or that is not updated with its upstream. **NOTE:** This option might go away since I implemented it on a whim, without discussing it with the team, after needing it a few times in the past.
2. Changes the logic of the `Force push` item under the `Repository` menu, to always show up when the current branch has diverged from its upstream. The rest of places where we usually suggest to force-push will keep the existing behavior: they will only show `Force push` when GitHub Desktop knows the branch diverged after the user performed an operation that is usually followed by a force-push (like a rebase or amending an already pushed commit).

I'm still considering "hiding" a `Force push` state in the pull/push/fetch button, when its state is `Pull` with ahead and behind commits (i.e. in a diverged branch). Maybe pressing <kbd>Alt</kbd> (or <kbd>Option</kbd> on macOS) while hovering the pull/push/fetch button in that state 🤔 

### Screenshots

https://user-images.githubusercontent.com/1083228/197004471-397680f5-bbbd-4af3-a81e-823c1a0cda23.mov

## Release notes

Notes: [Improved] Add new menu option for fetching at any time, and allow force-pushing diverged branches from menu

### DIFF
--- a/app/src/lib/rebase.ts
+++ b/app/src/lib/rebase.ts
@@ -3,6 +3,26 @@ import { IAheadBehind } from '../models/branch'
 import { TipState } from '../models/tip'
 import { clamp } from './clamp'
 
+/** Represents the force-push availability state of a branch. */
+export enum ForcePushBranchState {
+  /** The branch cannot be force-pushed (it hasn't diverged from its upstream) */
+  NotAvailable,
+
+  /**
+   * The branch can be force-pushed, but the user didn't do any operation that
+   * we consider should be followed by a force-push, like rebasing or amending a
+   * pushed commit.
+   */
+  Available,
+
+  /**
+   * The branch can be force-pushed, and the user did some operation that we
+   * consider should be followed by a force-push, like rebasing or amending a
+   * pushed commit.
+   */
+  Recommended,
+}
+
 /**
  * Format rebase percentage to ensure it's a value between 0 and 1, but to also
  * constrain it to two significant figures, avoiding the remainder that comes
@@ -16,17 +36,23 @@ export function formatRebaseValue(value: number) {
  * Check application state to see whether the action applied to the current
  * branch should be a force push
  */
-export function isCurrentBranchForcePush(
+export function getCurrentBranchForcePushState(
   branchesState: IBranchesState,
   aheadBehind: IAheadBehind | null
-) {
+): ForcePushBranchState {
   if (aheadBehind === null) {
     // no tracking branch found
-    return false
+    return ForcePushBranchState.NotAvailable
+  }
+
+  const { ahead, behind } = aheadBehind
+
+  if (behind === 0 || ahead === 0) {
+    // no a diverged branch to force push
+    return ForcePushBranchState.NotAvailable
   }
 
   const { tip, forcePushBranches } = branchesState
-  const { ahead, behind } = aheadBehind
 
   let canForcePushBranch = false
   if (tip.kind === TipState.Valid) {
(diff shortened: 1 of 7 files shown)


tidy-dev:
> Maybe pressing Alt (or Option on macOS) while hovering the pull/push/fetch button in that state

Just 2 cents with what I have been learning from accessibility. We wouldn't want to hide functionality behind a hover state; tho, I maybe it is ok given there is an alternate path -> the main menu/keyboard shortcut. However, maybe if there is a way to have a visible control while there is a divergent state that would be more ideal -> like I think there was talk of it being a dropdown button? And not solely from an accessibility stand point, it would be more discoverable too.

anafebacus:
Read me

Noong Biy, Okt 21, 2022, 2:49 AM, isinulat ni tidy-dev <
***@***.***> ang:

sergiou87:
Yes, I was counting on that being just an extra "mouse-based shortcut", and the accessibility would come from the menu item and its keyboard shortcut, exactly where it's always been (but now available more frequently) so the force push feature is not hidden.

I think the <kbd>Option</kbd> + something else is a common pattern in Apple's apps for these kinds of "mouse-based shortcuts", hence why I was suggesting it.

The dropdown sounds great in terms of accessibility and discoverability, if we find a nice way of fitting that behavior in the UI 🤔

tidy-dev:
>  exactly where it's always been

Touch back to discoverability. I had no idea it was there till this PR. :D I always used the force push option in the interface.

sergiou87:
Just went ahead and did the <kbd>Option</kbd> thing to switch between the `Pull` and `Force push` for diverged branches. It works even when the mouse is not on the button.

https://user-images.githubusercontent.com/1083228/197418643-915bf0a0-041f-4b45-947a-4098e707fb56.mov

sergiou87:
Superseded by https://github.com/desktop/desktop/pull/15520 and https://github.com/desktop/desktop/pull/15521

We'll explore other options to make this new force-push (or even the fetch) available from the push-pull-fetch button (e.g. turning it into a dropdown)

PR #15496, opened 2022-10-20
Stacked popups

tidy-dev:
## Description
This is one of of a couple PR's to introduce the idea of stacking popups. It does so through introducing a popup manager that works as a service introduced where needed.

Next PR, refactor the app errors to be part of the other popups so they are stacked as well.

Currently, this just shows one popup at a time, thus, on open and close the popup gets re-rendered and it's state effectively reset. This could be a problem for a popup with a more complicated state that we want to remember. A future pr will explore possibility of having the stack rendered by only one popup visible at a time or any other method for remembering states. 

I am attempting to not change the current app behavior regarding popups. For example, we have cases where we close the popup without a type -> I preserved that capability by removing the latest in the stack when no type is provided. Also, the `_completeOpenInDesktop` feels out of place here.. maybe it should be moved to the cloning components or whatever method calls the close popup in coordination with cloning. Thus, these things could be improved as part of this work, but I would like to isolate them to other PRs to be able to regression test them separately. 

Open question:
1. Add a close all?
2. Add a way to close popup type x when popup type y is closed? Like if a popup was open and an error popup arises rendering that popup in a bad state such that we want to close it on error popup closing. I don't have a real scenario in mind yet tho, so maybe wait till it is needed.

### Screenshots

https://user-images.githubusercontent.com/75402236/197042804-a3c40e99-9cd5-4721-8fab-1b56f481eded.mov

## Release notes

Notes: [Improved] Opening a new popup up does not close the existing one.


tidy-dev:
For context...
Discussed this approach with Markus, decided to move away from this being a singleton so that we can maintain our approach using the app-store as our controlling entity for app-wide state management. Thus, I am going to refactor slightly to make it so that all app interaction with the popup manager goes through the app-store.

app/src/lib/popup-manager.ts
```diff
@@ -0,0 +1,118 @@
...
+   */
+  public get isAPopupOpen(): boolean {
+    return this.currentPopup !== null
+  }
+
+  /**
+   * Returns an array of all popups in the stack of the provided type.
+   **/
+  public getPopupsOfType(popupType: PopupType): ReadonlyArray<Popup> {
+    return [...this.popupStack.filter(p => p.type === popupType)]
```
sergiou87: Why not this? 🤔
```suggestion
    return this.popupStack.filter(p => p.type === popupType)
```

app/src/lib/popup-manager.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Returns whether there are any popups of a given type in the stack.
+   */
+  public arePopupsOfType(popupType: PopupType): boolean {
```
sergiou87: Nit? (Feel free to ignore it)
```suggestion
  public areTherePopupsOfType(popupType: PopupType): boolean {
```

app/src/lib/popup-manager.ts
```diff
@@ -0,0 +1,118 @@
...
+      ...this.popupStack.slice(index + 1),
+    ]
+    return popupToUpdate
+  }
+
+  /**
+   * Removes a popup based on it's id.
+   */
+  public removePopup(popup: Popup) {
+    if (popup.id === null) {
```
sergiou87: Shouldn't this be `undefined`?
```suggestion
    if (popup.id === undefined) {
```

app/src/models/popup.ts
```diff
@@ -24,351 +24,530 @@ import { ValidNotificationPullRequestReview } from '../lib/valid-notification-pu
...
+  externalEditorLabel?: string
+  imageDiffType: ImageDiffType
+  recentBranches: ReadonlyArray<Branch>
+  repository: Repository
+  nonLocalCommitSHA: string | null
+  showSideBySideDiff: boolean
 }
 
 export type Popup =
-  | { type: PopupType.RenameBranch; repository: Repository; branch: Branch }
```
sergiou87, line 93: I assume you made all these changes so that all popup types had the common `id` property.

If you want to keep the changes to a minimum, I think you can write something like this:

```typescript
type PopupDetail = PopupType1 | PopupType2 | ...
export type Popup = { readonly id?: string } & PopupDetail
```

This is an example of using `&` to "join" types

sergiou87 (commented):
Looks good overall! Just a few comments. I'm a bit afraid of the regressions this new might introduce 😂

Maybe we could use a feature flag to make the PopupManager switch between using the stack or use the current single-popup approach? 🤔

app/src/lib/popup-manager.ts
tidy-dev: Habit from other angular/react stacks were we always copied arrays like this; but, the reasoning is not applicable here (I don't think anyways - has to do with defining new memory location and such).

app/src/lib/popup-manager.ts
tidy-dev: Yes... and I thought I intentionally looked for doing this.. because my original solution said they could be null but then realized be better to be optionally defined so as to not change everywhere in the app. 

Other side note.. I attempted to make a unit test file for this and it was giving me errors todo with making the uuid.. Googling around and it said updating jest to 27.0.0 would fix it... but I naively tried that and got a whole other host of errors. I may revisit making unit test.. but probably would have to setup a way to mock getting the unique id for the tests.

app/src/lib/popup-manager.ts
tidy-dev: I have no preference so I will change it but funnily enough I have been on a team where `are` would be preferred over `areThere`

app/src/models/popup.ts
tidy-dev: I think my druthers would be to have done interfaces.. but I have no real reasoning asides from it looks/feels nicer which is probably just "what I am used to" so not really valid :D ; and in this case I love not having to change as much existing stuff.

tidy-dev:
> Maybe we could use a feature flag to make the PopupManager switch between using the stack or use the current single-popup approach? 

Totally right -> I totally forgot to feature flag this..

app/src/models/popup.ts
sergiou87: Yeah I think deciding between types and interfaces is tough. I'd probably had done the same as you… for a new file. For an existing long file like this I'd rather keep the changes to a minimum. But that still is a personal preference! 😂

app/src/lib/popup-manager.ts
sergiou87: I just sounded odd to me. Maybe even `hasPopupsOfType` is more common for this kind of methods? Anyway, nothing to worry much about IMHO 😄

sergiou87 (dismissed):
Thanks for those changes!

sergiou87 (approved):
Now tested and seems to work as expected, with both the feature flag enabled and disabled 🙌

PR #15500, opened 2022-10-21
PR View: Dynamic Dialogs

tidy-dev:
## Description
This PR makes the width and height of the pr dialog responsive. Thanks to @niik's css wizardry we could discard my [ over engineering js madness](https://github.com/desktop/desktop/pull/15427) and use this purely css approach.

### Screenshots
Here is showing it in Windows. 

https://user-images.githubusercontent.com/75402236/197210539-9a2e96ca-335e-451b-9ab8-9fc1c5709453.mov

## Release notes
Notes: [Improved] Pull request preview dialog width and height is responsive.

### DIFF
--- a/app/src/ui/diff/code-mirror-host.tsx
+++ b/app/src/ui/diff/code-mirror-host.tsx
@@ -164,6 +164,12 @@ export class CodeMirrorHost extends React.Component<ICodeMirrorHostProps, {}> {
 
     CodeMirrorHost.updateDoc(this.codeMirror, this.props.value)
     this.resizeObserver.observe(this.codeMirror.getWrapperElement())
+
+    document.addEventListener('dialog-appeared', this.onDialogAppeared)
+  }
+
+  private onDialogAppeared = () => {
+    requestAnimationFrame(this.onResized)
   }
 
   private onSwapDoc = (cm: Editor, oldDoc: Doc) => {
@@ -199,6 +205,7 @@ export class CodeMirrorHost extends React.Component<ICodeMirrorHostProps, {}> {
     }
 
     this.resizeObserver.disconnect()
+    document.removeEventListener('dialog-show', this.onDialogAppeared)
   }
 
   public componentDidUpdate(prevProps: ICodeMirrorHostProps) {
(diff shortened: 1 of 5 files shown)


app/src/ui/diff/code-mirror-host.tsx
```diff
@@ -164,6 +164,12 @@ export class CodeMirrorHost extends React.Component<ICodeMirrorHostProps, {}> {
 
     CodeMirrorHost.updateDoc(this.codeMirror, this.props.value)
     this.resizeObserver.observe(this.codeMirror.getWrapperElement())
+
+    document.addEventListener('dialog-appeared', this.onDialogAppeared)
```
tidy-dev: For context, Codemirror was rendering before the dialog was fully mounted and in the scenario of having no scroll bars on either the file list and the diff, the diff would not render well the first time. This event listener is to make code mirror rerender after dialog has appeared and size figured out. You may see a quick flicker when opening in unified mode. This is a quick solution to make code mirror work. When we switch to the split diff engine, this won't be needed anymore.

sergiou87: From what I see, this event is received by CodeMirror instances that are not inside a dialog, which I guess we don't need/want (it will trigger the `refresh()` of all rendered CodeMirror instances when _any_ dialog appears).

Not sure if there is an easy/performant way of narrowing down this event to only child elements of a dialog, other than replacing that `document` with the closest `dialog` ancestor.

sergiou87 (commented):
Looks and works great!! Nice CSS magic 🪄  Only have one comment related to performance, but I don't think it's critical at all, but it'd be nice to address if you agree.

app/src/ui/diff/code-mirror-host.tsx
sergiou87: > When we switch to the split diff engine, this won't be needed anymore.

Oh lol I guess I spent too long writing my review and missed this 😂

tidy-dev: Yeah. But, you still have a valid point - I am out of time for now, but will see if I can do like you said and narrow it to code mirror instances inside a dialog.

sergiou87 (approved):
:shipit:

PR #15514, opened 2022-10-25
Improve keyboard shortcut in Changes and History tab

### LINKED ISSUE #9496: Keyboard shortcut for focusing Changes and History views

While my focus is on the commit message (⌘G) I want to return focus to the Changes or History view but I can't.

### Proposed solution

My proposal is to make ⌘1 and ⌘2 focus on the views after showing them.

#### How will it benefit Desktop and its users?
Github Desktop is the simplest and quickest git client for the most common git tasks. Changing the focus from the commit message to the Changes and History views is a very common task. The new behavior will make a very common task even quicker. This is why I believe this feature is a step towards GitHub Desktop values.

angusdev:
Closes #9496

## Description
When switch between Changes and History tab (click tab buttons, ctrl+tab, cmd+1/2), focus the list so can use keyboard up / down arrow keys to navigate the changed files.

Also in History tab, use left / right arrow to switch between file list and commit list.

#### Implementation ####

In order to grab the focus for keyboard navigation, it needs to focus the actual DOM element of `Grid`. So there will be a chain of calls from `Repository` (handle the `keydown` event) to `List` which can access the `Grid` DOM element as shown in below hierarchy.  Not sure if this is the best way to do so.

```
Repository
├── ChangesSidebar
│   └── ChangesList
│       └── List 
│           └── Grid
├── CompareSidebar
│   └── CommitList
│       └── List
│           └── Grid
└── SelectedCommits
    └── FileList
        └── List
            └── Grid
```

### Screenshots

https://user-images.githubusercontent.com/34896/197655888-9df7721e-67fc-4c69-a5a3-e5cd7e84e174.mov

## Release notes

Notes: IMPROVED Improve keyboard shortcut in Changes and History tab

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -240,6 +240,7 @@ export class ChangesList extends React.Component<
   IChangesState
 > {
   private headerRef = createObservableRef<HTMLDivElement>()
+  private list: List | null = null
 
   public constructor(props: IChangesListProps) {
     super(props)
@@ -259,6 +260,12 @@ export class ChangesList extends React.Component<
       )
     ) {
       this.setState({ selectedRows: getSelectedRowsFromProps(nextProps) })
+    }
+  }
+
+  public focus() {
+    if (this.list !== null) {
+      this.list.focus()
     }
   }
 
@@ -903,6 +910,10 @@ export class ChangesList extends React.Component<
     return
   }
 
+  private onListRef = (ref: List | null) => {
+    this.list = ref
+  }
+
   public render() {
     const { workingDirectory, rebaseConflictState, isCommitting } = this.props
     const { files } = workingDirectory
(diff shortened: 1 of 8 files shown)


angusdev:
Consider below as a bug?

Press up arrow key at first item normally will wrap the selection to the last item. But since the Change lists is partially loaded, more items will be appended so the selection will be in the middle of the list.  This is same as pressing the end key.  This behaviour is more or less expected for end key but not for up key.

To solve it need to disable the selection wrapping (add a `prop` in `List` and pass from `CommitList`)

github

niik:
Hey @angusdev, sorry for leaving you hanging here. With https://github.com/desktop/desktop/pull/16463 we're now moving focus when using the keyboard shortcuts which I believe would satisty the original issue here. Given that this PR has grown stale with conflicts I'm going to close it out to attempt to keep our PR list manageable and hopefully reduce the risk of PRs getting lost like this one.

PR #15520, opened 2022-10-25
Allow fetching at any time from menu

### LINKED ISSUE #7805: I don't seem to have a way to force a fetch

sergiou87:
Closes #7805

## Description

This PR supersedes #15494 in terms of allowing users to fetch the current repo at any time via the `Repository` menu.

### Screenshots

2022-10-25 at 11 21

## Release notes

Notes: [Improved] Add menu option to Fetch the current repository at any time

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -286,6 +286,12 @@ export function buildDefaultMenu({
         click: emit('pull'),
       },
       {
+        id: 'fetch',
+        label: __DARWIN__ ? 'Fetch' : '&Fetch',
+        accelerator: 'CmdOrCtrl+Shift+T',
+        click: emit('fetch'),
+      },
+      {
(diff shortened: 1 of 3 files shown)


tidy-dev (approved):
✨  Nice!

PR #15522, opened 2022-10-25
Metrics for notifications from recent repositories

sergiou87:
## Description
As discussed yesterday, and related to https://github.com/desktop/desktop/pull/15487#issuecomment-1284274142, this PR intends to gather metrics of how many events (potential notifications) users get from repositories other than the selected one.

Initially, we agreed to collect those metrics only for the recent repos, but it feels odd not to also keep track of events for non-recent repos. That "non-recent repos" thing is in a separate commit, so there is no problem if I have to get rid of that. Thoughts? cc @niik since you suggested focusing on recent repos!

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-database.ts
+++ b/app/src/lib/stats/stats-database.ts
@@ -473,6 +473,12 @@ export interface IDailyMeasures {
    */
   readonly checksFailedNotificationFromRecentRepoCount: number
 
+  /**
+   * The number of "checks failed" notifications the user received for a
+   * non-recent repository other than the selected one.
+   */
+  readonly checksFailedNotificationFromNonRecentRepoCount: number
+
   /** The number of "checks failed" notifications the user clicked */
   readonly checksFailedNotificationClicked: number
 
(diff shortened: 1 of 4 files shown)


tidy-dev (approved):
✨ This makes sense to me. 

My take away from the conversation was that we want to track when a notification occurs in any repository that is not the currently checked out one. This is to answer the question "How many notifications are we not showing? How much noise are we cancelling - is it that bad? Are there certain users this would necessarily harm/make notifications so noisy they are useless?" 

I think the assumption is we may have users with lots of repos and thus notifications for all repos could get really noisy for those users. The above metric would tell if that is true. If it is true, but we still want to open up more notifications  for the majority of our users that only have a couple repos,  the the next cutoff could be recent repos. Thus, we also want to track that to see if that would be "not noisy" enough for those users where opening the notifications on all repos would be harmful.

Edit: I also think recent repos follows the same thought pattern of "I pushed a PR, now I am going to work elsewhere, I want to be notified about the pr **I just pushed recently**. " whether my elsewhere is another branch in the same repo or in another repo.

sergiou87:
Agreed @tidy-dev , that's why I thought tracking _also_ notification to non-recent repos was also relevant. I like your point about recent repo = repo where I recently submitted a PR, I think that makes sense!

PR #15523, opened 2022-10-25
Stacked Popups -> Errors be popups

tidy-dev:
Based on #15496 

## Description
Continuation of stacked popups work. Now app errors are handled by the popup manager like other popups. 

Key logic: 
1. Error popups are shown before other popups.
2. We can have multiple popups of type error.

Other notes:
- We seem to have divergent approaches to error handling. When an error is captured, we parse it through the custom error handlers defined in `error-handlers.ts` with it lastly being handled by the `defaultErrorHandler` which is when the error ends up in the `AppError` component.. which then again does some custom error handling for cloning and auth errors. It seems to me the custom error handling done in the `AppError` component should be moved to an errorhandler method in the `error-handling.ts`. Unfortunately, any popup there is defined as it's own popup is not an error type popup so it does not get the benefit of being shown before other popup types. This could have the unintended effect of the stack being [error, (not really error), error]. The two non custom errors are displayed first.... however, since interacting with the app is so limited when an error is displayed.. I would suspect this scenario is unlikely.. But, it is event driven so suppose could happen. I think a cleaner solution.. would be to have the `AppError` component run the error handlers and categorize errors into types to display as a switch statement.. thus, all the error would be error popups and all the error logic would be under the `AppError` component. Tho not sure it is worth a refactor at this point.
- Other notes.. the `AppError` sometimes dispatches to open preferences... technically as the code stands if there are 2 app errors and the second has the option to open preferences. If the user hits to open the preferences, they will be shown the other error before the preferences is displayed... This feels wrong.. but not sure there is better option as the error may be prudent..

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/app-error.tsx
+++ b/app/src/ui/app-error.tsx
@@ -9,7 +9,6 @@ import {
 import { dialogTransitionTimeout } from './app'
 import { GitError, isAuthFailureError } from '../lib/git/core'
 import { Popup, PopupType } from '../models/popup'
-import { TransitionGroup, CSSTransition } from 'react-transition-group'
 import { OkCancelButtonGroup } from './dialog/ok-cancel-button-group'
 import { ErrorWithMetadata } from '../lib/error-with-metadata'
 import { RetryActionType, RetryAction } from '../models/retry-actions'
@@ -18,14 +17,11 @@ import memoizeOne from 'memoize-one'
 import { parseCarriageReturn } from '../lib/parse-carriage-return'
 
 interface IAppErrorProps {
-  /** The list of queued, app-wide, errors  */
-  readonly errors: ReadonlyArray<Error>
+  /** The error to be displayed  */
+  readonly error: Error
 
-  /**
-   * A callback which is used whenever a particular error
-   * has been shown to, and been dismissed by, the user.
-   */
-  readonly onClearError: (error: Error) => void
+  /** Called to dismiss the dialog */
+  readonly onDismissed: () => void
   readonly onShowPopup: (popupType: Popup) => void | undefined
   readonly onRetryAction: (retryAction: RetryAction) => void
 }
@@ -53,13 +49,13 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
   public constructor(props: IAppErrorProps) {
     super(props)
     this.state = {
-      error: props.errors[0] || null,
+      error: props.error,
       disabled: false,
     }
   }
 
   public componentWillReceiveProps(nextProps: IAppErrorProps) {
-    const error = nextProps.errors[0] || null
+    const error = nextProps.error
 
     // We keep the currently shown error until it has disappeared
     // from the first spot in the application error queue.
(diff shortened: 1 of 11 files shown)


app/src/ui/app.tsx
```diff
@@ -1393,18 +1398,24 @@ export class App extends React.Component<IAppProps, IAppState> {
...
-    }
-
     const popup = this.state.currentPopup
 
     if (!popup) {
       return null
     }
 
-    const onPopupDismissedFn = this.getOnPopupDismissedFn(popup.type)
+    if (popup.id === undefined) {
```
sergiou87, line 1407: Unless I'm missing something, this could happen with the feature flag disabled and therefore we'll send many non-fatal errors to Sentry. Could we check the feature flag here too?
```suggestion
    if (popup.id === undefined && enableStackedPopups()) {
```

(If you accept this suggestion you will need to import the feature flag at the top)

app/src/ui/app.tsx
```diff
@@ -1393,18 +1398,24 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+      sendNonFatalException(
+        'PopupNoId',
+        new Error(
+          `Attempted to open a popup of type '${popup.type}' without an Id`
+        )
+      )
+      return null
+    }
+
+    const onPopupDismissedFn = this.getOnPopupDismissedFn(popup.id)
```
sergiou87, line 1418: This wouldn't work with the feature flag disabled then, right? Maybe when we use only one dialog (non-stacked) we should just generate an ID and be done with it? 🤔

app/src/ui/app-error.tsx
```diff
@@ -53,13 +49,13 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
   public constructor(props: IAppErrorProps) {
     super(props)
     this.state = {
-      error: props.errors[0] || null,
+      error: props.error,
       disabled: false,
     }
   }
 
   public componentWillReceiveProps(nextProps: IAppErrorProps) {
```
sergiou87, line 57: We probably don't need this method anymore, and can remove the `error` from the state, right? Unless I understood it wrong, this popup will behave like any other, and will have an instance of it in the stack for each error.

app/src/ui/app-error.tsx
```diff
@@ -69,18 +65,7 @@ export class AppError extends React.Component<IAppErrorProps, IAppErrorState> {
   }
 
   private onDismissed = () => {
```
sergiou87: Maybe this `onDismissed` can be removed entirely and just use `this.props.onDismissed` wherever is needed?

sergiou87 (changes requested):
Looks good overall! Just a few things to look at.

I wasn't aware of the small differences between errors and other dialogs until now 🤯

app/src/ui/app.tsx
tidy-dev: I am going to resolve this by just making all popups now have an id (whether flag is enabled or not)

app/src/ui/app-error.tsx
tidy-dev: You may very well be right. I was trying to be the least impactful to this component... and didn't know exactly the bug this state management was supposed to prevent - maybe some scenario I don't know to test. I assume it happened when a user had back to back errors.. which can still happen here.  I am kind of keen on leaving this alone unless i have time to really investigate what that change might mean as I don't think it is needed for this work.

app/src/ui/app.tsx
tidy-dev: ^ Fixed by making all of them have an id.

sergiou87 (approved):
Looks and works great! :shipit:

PR #15530, opened 2022-10-26
Add default update error message

tidy-dev:
## Description
When attempting to install the latest beta on my windows vm, I would click check for updates in the about dialog and it would show the downloading updates with a spinner and would eventually open an error dialog with no message.  I found out the download/update was failing since my windows VM didn't have enough room on the hard drive.

Screen Shot 2022-10-26 at 12 35 17 PM

Screen Shot 2022-10-26 at 12 37 09 PM

## Release notes
Notes: [Improved] Always show a message to error when update fails.

### DIFF
--- a/app/src/ui/app.tsx
+++ b/app/src/ui/app.tsx
@@ -284,7 +284,14 @@ export class App extends React.Component<IAppProps, IAppState> {
     updateStore.onError(error => {
       log.error(`Error checking for updates`, error)
 
-      this.props.dispatcher.postError(error)
+      // It is possible to obtain an error with no message. This was found to be
+      // the case on a windows instance where there was not space on the hard
+      // drive to download the installer. In this case, we want to override the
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
🙌 :shipit:

JonayJonay:
Leave

JPTbohs:
Block all public account and devices 
> When attempting to install the latest beta on my windows vm, I would click check for updates in the about dialog and it would show the downloading updates with a spinner and would eventually open an error dialog with no message. I found out the download/update was failing since my windows VM didn't have enough room on the hard drive.
> 
> Screen Shot 2022-10-26 at 12 35 17 PM
> 
> Screen Shot 2022-10-26 at 12 37 09 PM
> 
> ## Release notes
> Notes: [you can’t ] Always show a message to error when update fails.

PR #15535, opened 2022-10-27
"Failed checks" notification fixes and improvements

### LINKED ISSUE #15422: Notifications occur for every check if any fail

sergiou87:
Closes #15422

## Description

As part of my work investigating the issues with notifications reported in #15422, this PR includes a few fixes and improvements:
- For forks, the wrong ref was used to retrieve the PR checks, using the PR branch name instead of the generic `refs/pull/XX/head` ref. Basically, this meant this notification didn't work for PRs submitted from forks.
- In line with that, make sure we're requesting the checks to the target repository of the PR (if it's a fork, that'd be the parent repository).
- When a "check suite failed" event is received for a particular commit, for which multiple check suites are still running, only process and send a notification for this first check suite and ignore the rest if any of them failed too.
- When retrieving the PR checks in preparation for a "checks failed" notification, skip any cached responses of the API endpoints. That way, the notification will always show accurate info about the current state of the PR checks (prior to this change it could show outdated info).

## Release notes

Notes: [Fixed] Notifications when checks of a Pull Request fail are displayed for forked repositories

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -1053,11 +1053,14 @@ export class API {
   public async fetchCombinedRefStatus(
     owner: string,
     name: string,
-    ref: string
+    ref: string,
+    reloadCache: boolean = false
   ): Promise<IAPIRefStatus | null> {
     const safeRef = encodeURIComponent(ref)
     const path = `repos/${owner}/${name}/commits/${safeRef}/status?per_page=100`
-    const response = await this.request('GET', path)
+    const response = await this.request('GET', path, {
+      reloadCache,
+    })
 
     try {
       return await parsedResponse<IAPIRefStatus>(response)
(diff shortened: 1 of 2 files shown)


tidy-dev (approved):
✨  Makes sense. 

Sry for the wait.. I read through this one before, but must have forgot to finish reviewing it.

PR #15549, opened 2022-11-01
Fix Shift + Down Arrow not selecting properly in the Commit List

### LINKED ISSUE #15536: Shift+DownArrow cannot select multiple commits in History tab
### The problem

In History tab, can use Shift+UpArrow to select multiple continous items.  But Shift+DownArrow can only select 1 or 2 items.  See the screenshot video for details.

### Release version

3.1.2

### Operating system

darwin-arm64

**Steps**
1. Go to History tab
2. Select first item
3. Press Shift+DownArrow
4. Press Shift+DownArrow again

***Expected***
Items 1-3 are selected

***Actual***
Only Item 2 is selected

### Log files

_No response_

### Screenshots

https://user-images.githubusercontent.com/34896/198418693-f2c8bb8c-bcab-4ca3-a557-8c025e8cdb7a.mov

### Additional context

`List` use `selectedRows` to determine the `selectionOrigin` and `lastSelection`. However `selectedRows` is sorted by decending order in `CommitList` in favour of cherry-pick handling. 

For example, when select row 5 and press up-arrow, `selectedRows = [5, 4]`. Then further press up-arrow will see `selectionOrigin = 5`, `lastSelection = 4`, `direction = 'up'`, so the next selected row = 3.

However, when select row 5 and press down-arrow, `selectedRows = [6, 5]` since it is sorted.  Then further press down-arrow wil see `selectedOrigin = 6`, `lastSelection = 5`, so the next selected row = 6, which the new `selectedRows = [6]`, then `[7, 6]`, `[7]`, `[8, 7]` etc.

https://github.com/desktop/desktop/blob/e6465802a25887c42747acbb07b1902af24e728f/app/src/ui/lib/list/list.tsx#L603-L611

https://github.com/desktop/desktop/blob/e6465802a25887c42747acbb07b1902af24e728f/app/src/ui/history/commit-list.tsx#L277-L287

There are four solutions:
1. Do not use `selectedRows` to determine `selectionOrigin` and `lastSelection`.  Instead explicit store these two values in states.  It is based on assumption that `selectRows` is passed by parent and not guaranteed in order of selection made. In this case needs to handle different keyboard and mouse events.  
2. Do not sort `selectedRows` in `CommitList` and handle the ordering of commits in function applying cherry-pick.  It is based on assumption that the order `selectedRow` should not be altered.
3. Remove `selectedRows` from `List.props`.  `selectedRows` should be internal state of `List` only. May add a `initialSelectedRows` for initialization only.
4. In `CommitList`, keep a separate unsorted copy of `selectedRows` and pass in `render()` instead of using the sorted `selectedSHAs`.  This is the simplest solution to solve this particular issue, but `selectedRows` may still be misused by other components.

https://github.com/desktop/desktop/blob/e6465802a25887c42747acbb07b1902af24e728f/app/src/ui/history/commit-list.tsx#L401-L408

tidy-dev:
Closes #15536

## Description
During cherry-picking development, I needed the selected commits for a cherry-pick to be ordered chronologically. Thus, I relied on the Commit List's row index ordering to sort it and then stored the sorted array in the apps `commitSelection.shas` property which is the propagated back down to the Commit List. This is not great because the `List` components methods for determining the how to modify the users selection when they use a keyboard command like `Shift` + `ArrowDown` rely on the first and last index on the `selectedRows` properties that is sent in. Thus, sorting as soon as a new selection changed resulted in the bug of never being able to expand your selection past 2 commits. 

This PR moves the logic for sorting up to the `cherry-pick` method where it more logically belongs as it is the only thing that actually needed it to be sorted (not the whole app). It also adds a note to the `selectedRows` property to help future devs from misstepping. Tho in hindsight, I should have realized that this logic belonged in the cherry-pick method anyways and found this other way to implement it.

Technically, interactive rebasing (squashing and rebasing) need this as well but due to how the interactive rebase git commands must be written, the sorting was happening at the point of the call anyways. Thus, this change should not impact them. See [reorder.ts](https://github.com/desktop/desktop/blob/7b34ee6afcb325b5c207cae2faff5bdbd56b7ec4/app/src/lib/git/reorder.ts) and [squash.ts](https://github.com/desktop/desktop/blob/development/app/src/lib/git/squash.ts) if you would like to double check my assessment.

Other comments.. it bothers me that the state from the apps `commitSelection.shas` is so coupled to the `List` component selection behaviors. But, potential impact to refactor the `List` component at this point is likely not worth it as the `List` component is used in many places throughout the app. 

### Screenshots

https://user-images.githubusercontent.com/75402236/199256238-6a392bc2-d7be-47ab-80ed-7f04fb8102d1.mov

## Release notes
Notes: [Fixed] Using the key command of 'Shift' + 'ArrowDown' adds the next commit below the current selection to the selection.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -6590,6 +6590,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
     }
   }
 
+  /**
+   * Multi selection on the commit list can give an order of 1, 5, 3 if that is
+   * how the user selected them. However, we want to main chronological ordering
+   * of the commits to reduce the chance of conflicts during interact rebasing.
+   * Thus, assuming 1 is the first commit made by the user and 5 is the last. We
+   * want the order to be, 1, 3, 5.
+   */
+  private orderCommitsByHistory(
+    repository: Repository,
+    commits: ReadonlyArray<CommitOneLine>
+  ) {
+    const { compareState } = this.repositoryStateCache.get(repository)
+    const { commitSHAs } = compareState
+
+    return [...commits].sort(
+      (a, b) => commitSHAs.indexOf(b.sha) - commitSHAs.indexOf(a.sha)
+    )
+  }
+
   /** This shouldn't be called directly. See `Dispatcher`. */
   public async _cherryPick(
     repository: Repository,
@@ -6600,13 +6619,15 @@ export class AppStore extends TypedBaseStore<IAppState> {
       return CherryPickResult.UnableToStart
     }
 
+    const orderedCommits = this.orderCommitsByHistory(repository, commits)
+
     await this._refreshRepository(repository)
 
     const progressCallback =
       this.getMultiCommitOperationProgressCallBack(repository)
     const gitStore = this.gitStoreCache.get(repository)
     const result = await gitStore.performFailableOperation(() =>
-      cherryPick(repository, commits, progressCallback)
+      cherryPick(repository, orderedCommits, progressCallback)
     )
 
     return result || CherryPickResult.Error
--- a/app/src/ui/history/commit-list.tsx
+++ b/app/src/ui/history/commit-list.tsx
@@ -275,15 +275,9 @@ export class CommitList extends React.Component<ICommitListProps, {}> {
   }
 
   private onSelectionChanged = (rows: ReadonlyArray<number>) => {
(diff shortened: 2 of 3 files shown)


sergiou87 (approved):
Your assessment makes total sense to me, and this works perfectly AFAICT!!! :_D

PR #15585, opened 2022-11-09
VSCodium Registry Key Update

voidei:
## Description

- Updated the codium UninstallKey values for x32, x64 and ARM64 as well as updated the publisher key to reflect change in the registry value.

## Release notes

Notes:

### DIFF
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -245,6 +245,28 @@ const editors: WindowsExternalEditor[] = [
     publisher: 'Microsoft Corporation',
   },
   {
+    name: 'Visual Studio Codium (Insiders)',
+    registryKeys: [
+      // 64-bit version of VSCodium - Insiders (user)
+      CurrentUserUninstallKey('{20F79D0D-A9AC-4220-9A81-CE675FFB6B41}_is1'),
+      // 32-bit version of VSCodium - Insiders (user)
+      CurrentUserUninstallKey('{ED2E5618-3E7E-4888-BF3C-A6CCC84F586F}_is1'),
+      // ARM64 version of VSCodium - Insiders (user)
+      CurrentUserUninstallKey('{2E362F92-14EA-455A-9ABD-3E656BBBFE71}_is1'),
+      // 64-bit version of VSCodium - Insiders (system)
+      LocalMachineUninstallKey('{B2E0DDB2-120E-4D34-9F7E-8C688FF839A2}_is1'),
+      // 32-bit version of VSCodium - Insiders (system)
+      Wow64LocalMachineUninstallKey(
+        '{EF35BB36-FA7E-4BB9-B7DA-D1E09F2DA9C9}_is1'
+      ),
+      // ARM64 version of VSCodium - Insiders (system)
+      LocalMachineUninstallKey('{44721278-64C6-4513-BC45-D48E07830599}_is1'),
+    ],
+    executableShimPaths: [['bin', 'codium-insiders.cmd']],
+    displayNamePrefix: 'VSCodium (Insiders)',
+    publisher: 'VSCodium',
+  },
+  {
     name: 'Sublime Text',
     registryKeys: [
       // Sublime Text 4 (and newer?)
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -214,7 +214,7 @@ const editors: WindowsExternalEditor[] = [
       // 32-bit version of VSCodium (system)
       Wow64LocalMachineUninstallKey(
         '{763CBF88-25C6-4B10-952F-326AE657F16B}_is1'
-        ),
+      ),
       // ARM64 version of VSCodium (system)
       LocalMachineUninstallKey('{67DEE444-3D04-4258-B92A-BC1F0FF2CAE4}_is1'),
     ],
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -12,7 +12,7 @@ import { IFoundEditor } from './found-editor'
 
 interface IWindowsAppInformation {
   displayName: string
-  publisher: string | string[]
+  publisher: string
   installLocation: string
 }
 
@@ -59,7 +59,7 @@ type WindowsExternalEditor = {
   readonly displayNamePrefix: string
 
   /** Value of the Publisher registry key that belongs to this editor. */
-  readonly publisher: string | string[]
+  readonly publishers: string[]
 
   /**
    * Default shell script name for JetBrains Product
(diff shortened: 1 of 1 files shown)


app/src/lib/editors/win32.ts
```diff
@@ -206,21 +206,21 @@ const editors: WindowsExternalEditor[] = [
...
+      CurrentUserUninstallKey('{57FD70A5-1B8D-4875-9F40-C5553F094828}_is1'),
       // 64-bit version of VSCodium (system)
-      LocalMachineUninstallKey('{D77B7E06-80BA-4137-BCF4-654B95CCEBC5}_is1'),
+      LocalMachineUninstallKey('{88DA3577-054F-4CA1-8122-7D820494CFFB}_is1'),
       // 32-bit version of VSCodium (system)
       Wow64LocalMachineUninstallKey(
-        '{E34003BB-9E10-4501-8C11-BE3FAA83F23F}_is1'
-      ),
+        '{763CBF88-25C6-4B10-952F-326AE657F16B}_is1'
+        ),
```
tsvetilian-ty: Thanks for the updates @voidei 🥳 

I've tested it on Windows x64 (both system and user installers), and it works as expected 💯 
The only issue I see is that the `Lint` part of the CI checks will potentially fail because of the formatting on line 217.
You could run `yarn prettier` and see if there are any warnings and run `yarn lint:fix` to fix the formatting.

voidei: Ah my bad, didn't see that when I was checking it. Thanks for the reminder, i've run prettier and lint:fix with 0 linting errors now.

sergiou87 (dismissed):
Thanks for your contribution! 🙇‍♂️  

It worked for me on Windows arm64. I'm a bit unsure about these changes since they could break the integration for existing users, but I think is the kind of app one would keep mostly updated, and I don't see traces of keeping older versions in the current implementation 🤔

tsvetilian-ty:
Suppose we try to combine older/legacy and newer versions. In that case, we will have an issue with the `publisher` since its value is also changed. The other alternative is to have two entries in the `editors` array with different values for `publisher`. Perhaps an implementation of a function that returns a `WindowsExternalEditor` with different values based on publisher param 🤔

sergiou87:
Yep, that's an option @tidy-dev and I were discussing and it might be a bit overkill for something that is supposed to be updated frequently (many times auto-updated), so maybe it's not worth supporting older versions in this case 🤔

srwrr (commented):
15585

voidei:
> Suppose we try to combine older/legacy and newer versions. In that case, we will have an issue with the `publisher` since its value is also changed. The other alternative is to have two entries in the `editors` array with different values for `publisher`. Perhaps an implementation of a function that returns a `WindowsExternalEditor` with different values based on publisher param 🤔

As a sort of way to deal with this, I've added the old keys and publisher ID (under the same outward name), excluding the x64 user key as that apparently change.

As well as that I've added the registry keys and values for VSCodium - Insiders, which seem to not have changed since they've been included in the codium source (tho i'm not entirely sure, would require digging through glame). Builds and functions on my end, picking up on both regular codium and insiders build.

sergiou87 (changes requested):
Sorry it took me a while to test this 😥 

I checked and this approach you suggest basically means that user who upgrade (or have upgraded!) from an older version with a different publisher, will now get two different entries in GitHub Desktop (both valid, though):

image

It looks like the VSCodium installers leave all the registry keys behind 😕  So I think we should either change `publisher` to `publishers` (and therefore hold an array of valid publishers) or keep only the newer version. Do we know when did VSCodium start using the `VSCodium` publisher?

voidei:
From what I could find, VSCodium started to use the new keys and publisher on [this commit](https://github.com/VSCodium/vscodium/pull/1246), with the earliest release after that being [1.71.1.22256](https://github.com/VSCodium/vscodium/releases/tag/1.71.1.22256). So the change has been a thing since early-mid September 2022. Working on a way to potentially get the 'publishers' idea working, ~~but there's issues w/ changing it from `string` to `string | string[]` - or simply to just `string[]` as other editors rely on things like `startsWith`.~~

Edit: turns out they don't use `startsWith`. I just don't know how to read

voidei:
Just updated it after testing that it worked for both `Microsoft Corporation` and `VSCodium` as the publishers and it worked fine on my end (would need testing on other platforms) - only thing to note is that since the registry keys for all but x64 User install are different, then those still using an older version of codium wouldn't have theirs picked up. I did try a few ways of getting it to read the registry key as a `string[]` instead of a `string` but I couldn't get it to work without needing to drastically redefine the type.

app/src/lib/editors/win32.ts
```diff
@@ -12,7 +12,7 @@ import { IFoundEditor } from './found-editor'
 
 interface IWindowsAppInformation {
   displayName: string
-  publisher: string
+  publisher: string | string[]
```
sergiou87: This is what Desktop retrieves from the registry, so it should _always_ be just one publisher.
```suggestion
  publisher: string
```

app/src/lib/editors/win32.ts
```diff
@@ -59,7 +59,7 @@ type WindowsExternalEditor = {
   readonly displayNamePrefix: string
 
   /** Value of the Publisher registry key that belongs to this editor. */
-  readonly publisher: string
+  readonly publisher: string | string[]
```
sergiou87: Since this can be one or more, we should just use `string[]` and then 1-item arrays for those which only have 1 known publisher.
```suggestion
  readonly publishers: string[]
```

app/src/lib/editors/win32.ts
```diff
@@ -216,21 +216,43 @@ const editors: WindowsExternalEditor[] = [
...
-      LocalMachineUninstallKey('{D77B7E06-80BA-4137-BCF4-654B95CCEBC5}_is1'),
+      LocalMachineUninstallKey('{88DA3577-054F-4CA1-8122-7D820494CFFB}_is1'),
       // 32-bit version of VSCodium (system)
       Wow64LocalMachineUninstallKey(
-        '{E34003BB-9E10-4501-8C11-BE3FAA83F23F}_is1'
+        '{763CBF88-25C6-4B10-952F-326AE657F16B}_is1'
       ),
       // ARM64 version of VSCodium (system)
-      LocalMachineUninstallKey('{D1ACE434-89C5-48D1-88D3-E2991DF85475}_is1'),
+      LocalMachineUninstallKey('{67DEE444-3D04-4258-B92A-BC1F0FF2CAE4}_is1'),
```
sergiou87: Shouldn't this have all known keys (the old and the new ones) instead of replacing ones with the others?

sergiou87 (changes requested):
I'm sorry but I don't see how those changes would work with multiple publishers 🤔 

[Line 496 of this editors/win32.ts](https://github.com/voidei/desktop/blob/f745df411ce995df66c281f9a2d33330b552473c/app/src/lib/editors/win32.ts#L496) should check if the `publisher` is not in the `editor.publishers` array:
```typescript
!editor.publishers.includes(publisher)
```

app/src/lib/editors/win32.ts
voidei: Yeah that's my bad, forgot it was an array not an object - should be good now

sergiou87 (approved):
Changes look good and it worked as expected! Thank you so much for your contribution (and patience) 🙇‍♂️ ❤️

PR #15588, opened 2022-11-09
PR Preview: Suggested Next Action

tidy-dev:
## Description
This PR adds a suggested next action Preview Pull Request by modifying the Create Pull Request suggested next action to have the two options of Preview Pull Request and Create Pull Request.  By default, it will show Preview Pull Request to the user; however, if the user makes a selection, it will remember that selection for the next PR. 

Testing Notes:
- I modified the dropdown-select-button styles to use flex box instead of percentages to be more flexible. I regression tested the use of it in the Merge Branch modal and the Merge option in the compare tool. 
- Edgecase testing: made the screen vary narrow to make sure we didn't get any really ugly wrapping.

### Screenshots
https://user-images.githubusercontent.com/75402236/200891028-c761619f-b2ff-44b1-91b3-1cf66cbe5420.mov

## Release notes
Notes: [Improved] User can preview a pull request from the suggested next actions.

### DIFF
--- a/app/src/ui/changes/no-changes.tsx
+++ b/app/src/ui/changes/no-changes.tsx
@@ -695,38 +695,40 @@ export class NoChanges extends React.Component<
       return null
     }
 
-    const createPullRequestAction: IDropdownSuggestedActionOption = {
-      title,
-      label: buttonText,
-      description,
-      value: PullRequestSuggestedNextAction.CreatePullRequest,
-      menuItemId: 'create-pull-request',
-      discoverabilityContent:
-        this.renderDiscoverabilityElements(createMenuItem),
-      disabled: !createMenuItem.enabled,
-      onClick: this.onCreatePullRequestClicked,
-    }
+    const createPullRequestAction: IDropdownSuggestedActionOption<PullRequestSuggestedNextAction> =
+      {
+        title,
+        label: buttonText,
+        description,
+        value: PullRequestSuggestedNextAction.CreatePullRequest,
+        menuItemId: 'create-pull-request',
+        discoverabilityContent:
+          this.renderDiscoverabilityElements(createMenuItem),
+        disabled: !createMenuItem.enabled,
+        onClick: this.onCreatePullRequestClicked,
+      }
 
-    const previewPullRequestAction: IDropdownSuggestedActionOption = {
-      title: `Preview the Pull Request from your current branch`,
-      label: 'Preview Pull Request',
-      description: (
-        <>
-          The current branch (<Ref>{tip.branch.name}</Ref>) is already published
-          to GitHub. Preview the changes this pull request will have before
-          proposing your changes.
-        </>
-      ),
-      value: PullRequestSuggestedNextAction.PreviewPullRequest,
-      menuItemId: 'start-pull-request',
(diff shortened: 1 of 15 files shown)


app/src/ui/suggested-actions/dropdown-suggested-action.tsx
```diff
@@ -0,0 +1,168 @@
...
+  /** The possible suggested next actions to select from
+   *
+   * This component assumes this is not an empty array.
+   */
+  readonly suggestedActions: ReadonlyArray<IDropdownSuggestedActionOption>
+
+  /** The value of the selected next action to initialize the component with */
+  readonly selectedActionValue?: string | undefined
+
+  readonly onSuggestedActionChanged: (action: string) => void
```
sergiou87: Could this be generic? Maybe we can add a `<T extends string = string>` (I think that syntax is correct 😆 ) to `IDropdownSuggestedActionProps` and `IDropdownSelectButtonOption` and use it here like this:

```suggestion
  readonly onSuggestedActionChanged: (action: T) => void
```

That way we don't need to use a `string` and cast it to our enum when the callback is invoked.

app/src/ui/changes/no-changes.tsx
```diff
@@ -657,17 +670,71 @@ export class NoChanges extends React.Component<
...
+      ),
+      value: PullRequestSuggestedNextAction.PreviewPullRequest,
+      menuItemId: 'start-pull-request',
+      discoverabilityContent: this.renderDiscoverabilityElements(startMenuItem),
+      disabled: !createMenuItem.enabled,
+    }
+
+    const pullRequestActions: ReadonlyArray<IDropdownSuggestedActionOption> = [
+      previewPullRequestAction,
+      createPullRequestAction as IDropdownSuggestedActionOption,
```
sergiou87: I think this cast is not needed? 🤔 
```suggestion
      createPullRequestAction,
```

(You might require to run the linter if you commit this change)

app/src/ui/suggested-actions/dropdown-suggested-action.tsx
sergiou87: This would also affect `selectedActionValue`:
```typescript
readonly selectedActionValue?: T | undefined
```

tidy-dev: Nice suggestion! I had to type the component for this to work (or I didn't know how not too) but I think that makes sense anyways to say what kind of dropdown suggested next action it is.

sergiou87 (dismissed):
Looks great!! I noticed a few things regarding the types that I think could be improved: some explicit types are not necessary, and it's probably better not to default `T` to `string` (sorry about that suggestion). Removing the `= string` forces the compiler to help us narrow down some other types.

Since changes are scattered over a few files and I needed to play a bit and see how much harm I would cause 😂  I made the changes in this commit: e558ec70137a3b5d78afd32723a5bcfc85ce17d6

Feel free to cherry-pick it if you like the changes in it (I do!). In any case, I won't block this PR on that so I'll approve this, since testing went great.

sergiou87 (approved):
🚀

PR #15590, opened 2022-11-09
PR Preview: Menu Item Availability

tidy-dev:
## Description
This PR makes updates the menu items availability and renames it to "Preview" as opposed to "Start" to have consistent referencing. 

Disabled when:
- another popup is open
- repository not active
- branch tip state is unborn or detached

Available changes:
- unpublished repo (as opposed to creating a pr)
- there is a PR open for the branch (as opposed to creating a pr)

## Release notes
Notes: [Improved] Preview Pull Request menu item is availability is consistent with other menu items.

### DIFF
--- a/app/src/main-process/menu/build-default-menu.ts
+++ b/app/src/main-process/menu/build-default-menu.ts
@@ -434,12 +434,12 @@ export function buildDefaultMenu({
     },
   ]
 
-  if (!hasCurrentPullRequest && enableStartingPullRequests()) {
+  if (enableStartingPullRequests()) {
     branchSubmenu.push({
-      label: __DARWIN__ ? 'Start Pull Request' : 'Start pull request',
-      id: 'start-pull-request',
+      label: __DARWIN__ ? 'Preview Pull Request' : 'Preview pull request',
+      id: 'preview-pull-request',
       accelerator: 'CmdOrCtrl+Alt+P',
-      click: emit('start-pull-request'),
+      click: emit('preview-pull-request'),
     })
   }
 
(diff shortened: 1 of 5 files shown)


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -434,12 +434,12 @@ export function buildDefaultMenu({
     },
   ]
 
-  if (!hasCurrentPullRequest && enableStartingPullRequests()) {
```
tidy-dev, line 437: I originally had this `!hasCurrentPullReques` to follow the convention for the create pull request item which switches to an open pull request item in this scenario. But, I don't think there is a good reason to preview a user from seeing the pull request preview of an open PR.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -434,12 +434,12 @@ export function buildDefaultMenu({
...
+  if (enableStartingPullRequests()) {
     branchSubmenu.push({
-      label: __DARWIN__ ? 'Start Pull Request' : 'Start pull request',
-      id: 'start-pull-request',
+      label: __DARWIN__ ? 'Preview Pull Request' : 'Preview pull request',
```
tidy-dev, line 439: After doing the suggested next actions changes, I realized that "Start Pull Request" doesn't exactly make sense to differentiate from creating a pull request. Decided to make consistent verbiage of "Preview Pull Request". But, I am completely open to discussion about what terminology would communicate the intention the best.

app/src/main-process/menu/build-default-menu.ts
sergiou87: Wouldn't it be confusing? If there is already a PR, seeing a "Preview PR" could lead you to think there isn't one already 🤔

sergiou87 (approved):
LGTM! Only have one question, but not gonna block the PR on that 😄

app/src/main-process/menu/build-default-menu.ts
tidy-dev: Create pr would be disabled. This would just be for viewing the PR Diff which I found myself wanting to do after making some changes to existing PR. .. If you think it would be confusing, we could switch the verbage to "View Pull Request Diff" in this scenario. Another thought not necessarily for this PR, but the create button could change to a push button.

Issue #2957, opened 2017-10-04
Scrolling in sidebar, selected item loses focus if out of view


NiklasBr:
### Description

When you have one item selected in the item/source list for changed files and scroll the item out of view it will lose focus and not regain its focus once returned into view. See attached screen recording below.

### Version

**GitHub Desktop version:** 1.0.3

**OS version:** macOS 10.12.6 (16G29)

### Steps to Reproduce

1. Have enough changes so that Desktop will allow you to scroll.
1. Select one item.
1. Scroll so that the item is no longer visible.
1. Scroll back so that the item becomes visible again.

![2017-10-04 10 41 01](https://user-images.githubusercontent.com/279826/31167213-9531cac2-a8f1-11e7-880f-097691e3d57c.gif)

**Expected behavior:** The item should always keep its focus.

**Actual behavior:** Focus is lost.

**Reproduces how often:** 100%

### Logs

n/a

#### Additional Information

n/a

tierninho:
Looks like we have a main and secondary color. @desktop/design can shed some light on this.

Can reproduce with clone repo list as well.

<img width="451" alt="screen shot 2017-10-04 at 10 00 37" src="https://user-images.githubusercontent.com/14828183/31167763-1728d144-a8eb-11e7-94f4-f99a650db487.png">
<img width="519" alt="screen shot 2017-10-04 at 10 00 20" src="https://user-images.githubusercontent.com/14828183/31167762-172690be-a8eb-11e7-9908-25e187104df6.png">

NiklasBr:
It's not just the colour. When the item has focus, the focus can be moved up and down to the previous/next items. But when losing the focus (and it it using the secondary grey colour) focus can no longer be moved up or down using the arrow keys.

shiftkey:
This is an unfortunate side-effect of using virtualized lists in many places. @niik talked a bit more about this limitation in https://github.com/desktop/desktop/issues/1669#issuecomment-314668885. I'll let him follow up with anything else, but maybe we do need to track this as a separate task.

niik:
Thanks for opening this @NiklasBr and nice find @shiftkey. I was scouring through issues because I knew I'd written something about this before but I couldn't find it. I definitely think this is worth tracking but as I said in that comment the fix is non-trivial.

danny007in:
can u make video,
ill try to solve

NiklasBr:
@danny007in  video is available in the first comment in this issue as well as in #1669

[...]

PR #15604, opened 2022-11-11
Virtualized list items get refocussed when recreated

markhicken:
Closes #[issue 2957](https://github.com/desktop/desktop/issues/2957)

## Description

As described in the issue, virtualized list items lose focus when scrolled out of view (because they are unmounted).

See comments with screenshots below...

### Screenshots

1. When scrolled back into view, this PR refocuses them.

    FocusRestored

2. We only want to restore focus if it was unintentionally lost. We'll keyboard tab away from the list to validate this.

    FocusNotRestored

3. Note that up and down keys still work as expected after refocussing.

    UpDownFocusRetained

## Release notes

Notes:
- Fixes an issue where selected files lost focus when scrolled out of view and then back into view.

### DIFF
--- a/app/src/ui/lib/list/list-row.tsx
+++ b/app/src/ui/lib/list/list-row.tsx
@@ -54,7 +54,7 @@ interface IListRowProps {
 
 export class ListRow extends React.Component<IListRowProps, {}> {
   private onRef = (element: HTMLDivElement | null) => {
-    this.props.onRef?.(element, this.props.rowIndex);
+    this.props.onRef?.(element, this.props.rowIndex)
   }
 
   private onRowMouseOver = (e: React.MouseEvent<HTMLDivElement>) => {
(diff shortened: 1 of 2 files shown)


markhicken:
@niik would you mind kicking the workflows for me or let me know who can?

niik:
Hey @markhicken, sorry for the late response. Thanks so much for wanting to contribute a fix to this long standing and annoying issue! While reviewing your changes I wanted to see if there was a way for us to combine this work with a fix so that keyboard navigation continues to work, that's the way the List component was intended to work and the reason why the Grid component has a tabIndex (for programmatic focus).

I took a stab at this in #15641 and added you as a co-author, please take a look at those changes and see if they make sense to you. In the meantime I'll close this PR as superseded by #15641 but if that work doesn't make it for some reason we can always reopen this.

PR #15605, opened 2022-11-11
[draft]: clone cancel

### LINKED ISSUE #2082: Canceling the cloning of a repo?
If a repo is in the middle of being cloned and Desktop is quit (accidentally I presume), the partially downloaded cloned repo remains on your local, while at the same time you are "blocked" from cloning the repo again unless you remove the files locally or clone it elsewhere. 

Should we implement the option to click a `cancel` button once the cloning process begins? Use cases: the user cloned the wrong repo or the repo is too big for their system and needs to be stopped. Of course all relevant files must be removed locally too, whether cancelled by the user or if the app is quit.

maifeeulasad:
Closes #2082

## Description

-

## Release notes

Notes:

### DIFF
--- a/app/src/lib/shells/darwin.ts
+++ b/app/src/lib/shells/darwin.ts
@@ -11,6 +11,7 @@ export enum Shell {
   PowerShellCore = 'PowerShell Core',
   Kitty = 'Kitty',
   Alacritty = 'Alacritty',
+  Tabby = 'Tabby',
   WezTerm = 'WezTerm',
   Warp = 'Warp',
 }
@@ -35,6 +36,8 @@ function getBundleID(shell: Shell): string {
       return 'net.kovidgoyal.kitty'
     case Shell.Alacritty:
       return 'io.alacritty'
+    case Shell.Tabby:
+      return 'org.tabby'
     case Shell.WezTerm:
       return 'com.github.wez.wezterm'
     case Shell.Warp:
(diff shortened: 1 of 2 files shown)


maifeeulasad:
@tidy-dev, @sergiou87 Hello,
 I tried to implement this cancel clone feature with the new `execGitTask` API. But I guess I'm missing something. Can you please take a look at it? I'm ready to do more experiments with it and doing right now.

sergiou87:
I recently learned about [`AbortController` and `AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal), which are used by other async APIs like [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#aborting_a_fetch) to cancel those async tasks.

Maybe we could use something like that? When the user initiates cloning, we store an instance of `AbortController` and pass its `signal` down to the clone function. The git function will also receive a new `abortSignal` parameter (from the `clone` function), and will listen to its events: when the signal is emitted, the task is cancelled.

Does that make sense? If this works well, maybe we should _also_ change the dugite API to use an `AbortSignal` given this is a more conventional approach (sorry about my initial suggestion of that `cancel` method 😢 ). However, that can be done later in a future PR.

IcebergVerse (commented):
app/src/lib/git/clone.ts

PR #15620, opened 2022-11-14
Make dropdown select component keyboard navigable

tidy-dev:
## Description
While working adding a suggested next actions dropdown for the PR Preview, I noticed that the `DropdownSelectButton` component was not keyboard navigable at all. This PR refactors the options to be buttons which naturally gives us navigation via the tab key and enter key. However, select components are also typically navigable via the up and down keys. Thus, I added in key handling for that as well as closing the dropdown on escape. Additionally, I made the hover/focus state use our "active" styles as this is inline with what dotcom does for its similar component and I think likely is so to meet the contrast accessibility requirements for "selected" state. 

### Screenshots

https://user-images.githubusercontent.com/75402236/201783749-d4316f2a-4100-4070-86c1-ea70dba9300b.mov

## Release notes
Notes: [Improved] The dropdown selection component is keyboard navigable.

### DIFF
--- a/app/src/ui/dropdown-select-button.tsx
+++ b/app/src/ui/dropdown-select-button.tsx
@@ -99,6 +99,11 @@ export class DropdownSelectButton extends React.Component<
 
   private onKeyDown = (event: KeyboardEvent) => {
     const { key } = event
+    if (this.state.showButtonOptions && key === 'Escape') {
+      this.setState({ showButtonOptions: false })
+      return
+    }
+
     if (
       !this.state.showButtonOptions ||
       !['ArrowUp', 'ArrowDown'].includes(key)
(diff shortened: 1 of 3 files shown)


app/src/ui/dropdown-select-button.tsx
```diff
@@ -145,6 +200,16 @@ export class DropdownSelectButton<
     )
   }
 
+  private renderOption = (o: IDropdownSelectButtonOption<T>) => {
+    return (
+      <Button key={o.value} onClick={this.onSelectionChange(o)}>
```
tidy-dev, line 205: Buttons might feel weird here as list item, but they naturally give us the accessibility features of being a selectable input without having to implement custom logic to focus and put keypress/onclicks on li or divs which then means we have to ignore linter rules. Also, it is what dotcom does as well.

sergiou87 (approved):
Code makes sense, I tested it and works as expected! 👏 :shipit:

PR #15630, opened 2022-11-16
Fix issue with notifications not displayed when a fork contributes to its parent

sergiou87:
## Description

While testing the latest beta, @tidy-dev noticed she didn't get notifications from PRs submitted from a fork to its parent. It seems `repository.workflowPreferences.forkContributionTarget` is `undefined` by default and it means "contributing to parent". In my tests, I switched to "Contribute to parent" manually, so I didn't have `undefined` there, hence why I didn't notice it… Luckily, we already had a `getForkContributionTarget` that looks for `undefined` values and gets us the right thing.

**BONUS FIX:** while reproducing this bug, I noticed an undesired behavior introduced in #15535. Since that PR, the app was ignoring subsequent notifications of check suites from the same PR in order to avoid flooding the user with notifications, at least until the user pushed more changes to their branch and that triggered new checks. However, that also disabled the notifications for check re-runs. I made a change to filter notifications by check runs instead of check suites, and seems to work as expected 🥲 

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/stores/notifications-store.ts
+++ b/app/src/lib/stores/notifications-store.ts
@@ -75,7 +75,7 @@ export class NotificationsStore {
     null
   private cachedCommits: Map<string, Commit> = new Map()
   private skipCommitShas: Set<string> = new Set()
-  private skipCheckSuites: Set<number> = new Set()
+  private skipCheckRuns: Set<number> = new Set()
 
   public constructor(
     private readonly accountsStore: AccountsStore,
@@ -216,10 +216,6 @@ export class NotificationsStore {
       return
     }
 
-    if (this.skipCheckSuites.has(event.check_suite_id)) {
-      return
-    }
-
     const pullRequests = await this.pullRequestCoordinator.getAllPullRequests(
       repository
     )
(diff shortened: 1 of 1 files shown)


app/src/lib/stores/notifications-store.ts
```diff
@@ -283,11 +290,11 @@ export class NotificationsStore {
       return
     }
 
-    // Ignore any remaining notification for check suites that started along
+    // Ignore any remaining notification for check runs that started along
     // with this one.
     for (const check of checks) {
       if (check.checkSuiteId !== null) {
```
tidy-dev: Should this `check.checkSuiteId !== null` be changed to `check.checkSuiteId === event.check_suite_id`?

sergiou87: Nope, because we're trying go skip notifications from _other_ check suites. However, maybe we can even remove that `if` and just add all `check.id` to the set 🤔

sergiou87: Done in 5d4fa4f2bbbecbdeb4c60a61bde3ee7fb39b567b !

tidy-dev (approved):
✨  Looks good!

Issue #10446, opened 2020-08-25
Add editor support for VSCodium


jebjordan:
### Describe the feature or problem you’d like to solve

A clear and concise description of what the feature or problem is.
I cannot select VSCodium as my editor even though I have it installed.
Can we not simply let the user select an executable to open the things in?

### Proposed solution

Put an option to simply browse for an executable to attempt to open the things in.

How will it benefit Desktop and its users?
more fine control, not being limited.

tierninho:
@jebjordan Thanks for the request. We do not code the integrations with each editor ourselves as it mean little room for development on features. Please see this document for directions on our to support VSCodium: 

https://github.com/desktop/desktop/blob/1de8d45fc6dc3da6b7099c68f0d4a97c311bf972/docs/technical/editor-integration.md

Labeling this issue as `help-wanted` in case others would like to contribute.

niik:
Hi @jebjordan. Support for VSCodium was added in https://github.com/desktop/desktop/pull/8163 and this is the first report we've gotten that it's not working. What OS are you on and what version of VSCodium are you using?

Cc @Rexogamer

niik:
Thank you for your issue!

We haven't gotten a response to our questions.
With only the information that is currently in the issue, we don't have enough
information to take action. I'm going to close this but don't hesitate to reach
out if you have or find the answers we need, we'll be happy to reopen the
issue.

KallePM:
The problem is that for VSCodium `isExpectedInstallation` checks the installation with this: 
`return displayName === 'Visual Source Codium' && publisher === 'VSCodium'` ([here](https://github.com/desktop/desktop/blob/7a7388edd069cc8e97ec6d3cdbe4d2747af5fbc4/app/src/lib/editors/win32.ts#L424))

To fix this, change the line mentioned above to:
```
return (
        displayName.startsWith('VSCodium') &&
        publisher === 'Microsoft Corporation'
      )
```
This will check for the correct values and fix the issue of VSCodium not showing up as an editor.

Rexogamer:
This is a pretty simple fix (although it seems a bit weird that Microsoft is still considered the publisher?), although I unfortunately wouldn't be able to test this currently. However, since it's one line (and seemingly you've tested this locally) do you wanna take this one Kalle?

mrrezaie:
Hi, 

In my case (Github Desktop 3.1.2 x64), there are two issues:

1.  the key doesn't match
[...]

PR #15639, opened 2022-11-18
Update VSCodium key and publisher

mrrezaie:
https://github.com/desktop/desktop/issues/10446#issuecomment-1319892688

However, I'm not sure the key remains identical for all versions.


tsvetilian-ty:
@mrrezaie Thank you for your interest! However, an active PR (https://github.com/desktop/desktop/pull/15585) aims to fix the registry key mismatch in the newer versions of VSCodium. 

I will close this PR, but if you think you can contribute more to GitHub Desktop, feel free to do it 🙌

PR #15640, opened 2022-11-18
Rerender diffs when newline are added/removed from the end of file

### LINKED ISSUE #15638: Desktop doesn't automatically refresh for a newline change. I have to select another file and back to the same file for checking that change
### The problem

The red icon at the end of the file doesn't get changed when I delete or add the newline from the end of the file.
I have to select another file and then come back to this file to check if there are any changes.

image

### Release version

3.1.2

### Operating system

Windows 10

_No response_

### Log files

_No response_

### Screenshots

_No response_

### Additional context

_No response_

niik:
Closes #15638

## Description

Our short-circuits to avoid unnecessary re-renders when a reloaded diff is equal to the previous diff didn't take EOF warnings into account

## Release notes

Notes: [Fixed] Update diff after adding or removing trailing newlines to a file

### DIFF
--- a/app/src/ui/diff/diff-helpers.tsx
+++ b/app/src/ui/diff/diff-helpers.tsx
@@ -9,7 +9,7 @@ import {
   CommittedFileChange,
 } from '../../models/status'
 import { DiffHunk, DiffHunkExpansionType } from '../../models/diff/raw-diff'
-import { DiffLineType } from '../../models/diff'
+import { DiffLineType, ILargeTextDiff, ITextDiff } from '../../models/diff'
 import { DiffSyntaxToken } from './diff-syntax-mode'
 
 /**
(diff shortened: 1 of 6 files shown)


app/src/ui/diff/diff-helpers.tsx
```diff
@@ -454,3 +454,39 @@ export function getFirstAndLastClassesUnified(
...
+    // This is a performance optimization which lets us avoid iterating over all
+    // lines (deep equality on all hunks). We're already comparing the diff text
+    // above so the only thing that can change with the diff text staying the
+    // same is whether or not the last line is followed by a trailing newline.
+    // That information is encodeded in the noTrailingNewLine property which
+    // exists on all lines but is only ever set on lines in the last hunk
+    return (
+      x.hunks.length === 0 ||
+      x.hunks[x.hunks.length - 1].equals(y.hunks[y.hunks.length - 1])
+    )
```
sergiou87, line 488: Given the whole diff could be one hunk, couldn't we just compare the last line of both diffs?

app/src/ui/diff/text-diff.tsx
```diff
@@ -1498,17 +1501,15 @@ export class TextDiff extends React.Component<ITextDiffProps, ITextDiffState> {
...
         // the viewport.
-        if (this.props.diff.text === prevProps.diff.text) {
+        if (isSameDiff) {
           this.updateViewport()
         }
       }
     }
 
-    if (this.props.diff.text !== prevProps.diff.text) {
+    if (isSameDiff) {
```
sergiou87: I think a `!` is missing here:
```suggestion
    if (!isSameDiff) {
```

sergiou87 (changes requested):

app/src/ui/diff/diff-helpers.tsx
niik: That would likely cover almost all cases yes. At least for "normal" use but it is possible to have two diffs where the text is equal, neither (or both) has the `noTrailingNewLine` property set on the last line but yet they are not equal. Here's one such example:

```diff
diff --git a/README.md b/README.md
index bab18e6..ffcaf82 100644
--- a/README.md
+++ b/README.md
@@ -1,2 +1,3 @@
 Foo
-Bar
\ No newline at end of file
+BarBar
+Baz
```

That change removed the last line of the file and replaced it with `BarBar\nBaz\n`. If the previous file did have a trailing newline the diff would look like this and the text would be equal.

```diff
diff --git a/README.md b/README.md
index bab18e6..ffcaf82 100644
--- a/README.md
+++ b/README.md
@@ -1,2 +1,3 @@
 Foo
-Bar
+BarBar
+Baz
```

Would you ever encounter this in real world usage? I don't know, and given how this only affects presentation we would likely get away with it even if it did but my head hurt too much to reason about it that I figured it was just easier to take the safer but teensy bit slower approach.

app/src/ui/diff/text-diff.tsx
niik: 🤦

sergiou87 (approved):
LGTM and works as expected!

PR #15661, opened 2022-11-21
Remember stacked popups state by "hiding" them

tidy-dev:
## Description
This is another iteration on top of the [recent stacked popup](https://github.com/desktop/desktop/pulls?q=is%3Apr+is%3Aclosed+stacked+popup+in%3Atitle) work.

This PR is utilizing react context for setting the `isVisible` property of a dialog. This is useful here as the `app.tsx` and the `dialog.tsx` component (not the popup components in between) are the only ones that care about the dialog visibility. 

Alternatively, this PR has 2 commits - this first commit users prop drilling ->  [wip](https://github.com/desktop/desktop/commit/24271d73c44b8f78b18c6cf10ffac39357626750) was my first attempt using prop drilling (and also doesn't use `dialog.close` which is necessary if popup stack reordering were to happen). 

Other thoughts:
1.  `React.Context` makes sense here because the `Dialog` component is the only one that needs to know whether it is visible or not. It is nice in that it "auto-magically" knows/works for all dialogs with minimal code change - we don't have to remember to pass this down on all future popup types.  On the other hand, we only have to prop drill from `App` to a `Popup` component to the `Dialog` component and this doesn't introduce a "new" code path for data in the App. 
2. Never have used `React.Context` before so I could be doing something stupid. :D It seems similar to the concept of having service providers in Angular that I am familiar with - same sort of pitfalls (globally variables, potential of  forcing of unexpected rendering, seemingly different sources of data.)
3. Could be a good time to shift all this popup logic that is housed in the `App` component into its own component. (I dabbled a little with this on this branch `app-popup-component`). But, this would just be for logical separation of concern sake..
5. Pondering if there is a more performant means of doing this -> Currently, closing or opening a popup re-renders all popups.. but we really only need to re-render the next one in the stack.
6. Not sure if it applies here, but it is recommended to memoize objects in context to prevent necessary pre-rendering.

### Screenshots

https://user-images.githubusercontent.com/75402236/203115010-5cccefc0-4039-419f-9ed2-488df9cd937e.mov

## Release notes
Notes: [Improved] Stacked popups remember their state.

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -119,6 +119,7 @@ export interface IAppState {
   readonly showWelcomeFlow: boolean
   readonly focusCommitMessage: boolean
   readonly currentPopup: Popup | null
+  readonly currentPopups: ReadonlyArray<Popup>
   readonly currentFoldout: Foldout | null
   readonly currentBanner: Banner | null
 
--- a/app/src/lib/globals.d.ts
+++ b/app/src/lib/globals.d.ts
@@ -162,6 +162,8 @@ interface Window {
 
 interface HTMLDialogElement {
   showModal: () => void
+  close: (returnValue?: string | undefined) => void
+  open: boolean
 }
 /**
  * Obtain the number of elements of a tuple type
--- a/app/src/lib/popup-manager.ts
+++ b/app/src/lib/popup-manager.ts
@@ -42,13 +42,24 @@ export class PopupManager {
   public constructor(private readonly popupLimit = defaultPopupStackLimit) {}
 
   /**
+   * Returns all the popups in the stack
+   *  If there are error popups, it returns on the top of the stack.
+   */
+  public get currentPopups(): ReadonlyArray<Popup> {
+    const errorPopups = this.getPopupsOfType(PopupType.Error)
+    const nonErrorPopups = this.popupStack.filter(
+      p => p.type !== PopupType.Error
+    )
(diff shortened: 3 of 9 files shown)


tidy-dev:
Just a recap of meeting about this:

Decision: Move forward with `React.Context`
- We want the solution well documented and well labeled as a one-off implementation specific to this use-case of React.Context and not something we want to introduce as a common pattern.

Thoughts:
- Main concerns:
   - Breaking well established pattern of prop drilling; adds mental load to external contributors understanding where data comes from.
   - Has similar detriments as a global variable; feels hacky/dangerous. We prefer the well established pattern of passing props in vast vast majority of cases so that bugs are easier to trace/ we know where data/changes comes from.
- React.Context make sense in this use case, because the dialog component at the bottom of the stack and is the only one that needs to know which dialog in the stack is currently available which is only known up at the app state.  It is also possible that a few (but, key mostly not) popup components could also need this information for knowing when to add/remove event listeners and would have access to it via the context.
- If we find this to be a mistake, we can use the alternative of passing in the props the dialog needs to all the 60+ popup components and passing that into the dialog props from those 60+ popups 
- Other thoughts around alternatives were putting the custom popup logic inside one PopupDialog component; or some dialog base class. (Invert the current dialog to custom popup approach) This would require major refactoring of all current popups. Still may find that not to work as some popups need a higher level of control over the dialog output.

PR #15668, opened 2022-11-22
PR Preview: Introduce react context for managing open/closing stacked popups

tidy-dev:
## Description
This is another iteration on top of the [recent stacked popup](https://github.com/desktop/desktop/pulls?q=is%3Apr+is%3Aclosed+stacked+popup+in%3Atitle) work in order for a closed popup to remember it's state while closed.

This PR is utilizing `React.Context` to add management of data around the stacked popups and the `Dialog` component. Currently, it only houses one property of `isTopMost` that is a boolean indicating whether the dialog is the top most on the stack and therefore should be the visible one of which the the use can interact. Based on this value, the `Dialog` component calls `showModal` or `close` on the dialog element. The `showModal` method adds an attribute of `open` as well as an invisible click shield, the click shield is why we cannot simply manage this by a css class. The `close` method inversely removes the click shield `open` attribute. Additionally, the dialog component as well as three other popup components subscribe and unsubscribe to event handlers base on whether the dialog is the top most. 

I utilized `memoizeOne` per @niik suggestion so that the update method is only run when the `isTopMost` value changes for the one or two dialog affected as opposed to running it in every dialog.

The `About`, `CreateRepository`, and `CloneRepository` dialogs all had `window.addEventListeners`. Thus, this pr also makes those components consume the context and adds/removes those listeners based on whether they are the top most popup. I regression tested this by:
- opening those dialogs
- placing a debugger in the listener methods 
- ensure they were working on the event listened for
- opened the test error dialog 
- ensured they were no longer being listened too
- closed the error dialog
- ensured they were being listened to again. 

[Read more](https://github.com/desktop/desktop/pull/15661#issuecomment-1323835749) about discussion of using React.Context.

### Screenshots

https://user-images.githubusercontent.com/75402236/203398951-a004343c-8c0d-4cac-9f20-4f69473c484f.mov

## Release notes
Notes: [Improved] Stacked popups remember their state when hidden due to another popup opening.


sergiou87 (commented):
LGTM overall!

I see a pattern with all the `checkWhetherDialogIsTopMost` and it occurs to me we could reuse some of that logic by creating a `TopMostDialogListener` component (or something with a better name) that doesn't render anything, and just has `checkWhetherDialogIsTopMost`, and invokes it in the `componentDid...` methods, and provides two props for `onDialogIsTopMost` and `onDialogIsNotTopMost` that clients of the component must implement.

Does that make sense? Do you think that'd be useful? I think we could get rid of a bunch of duplicated boilerplate.

tidy-dev:
> Does that make sense? Do you think that'd be useful? I think we could get rid of a bunch of duplicated boilerplate.

Totally does. Thank you for pointing that out. 

I named it `DialogStackContextConsumer` because in the world of `React.Context` the terminology is `Provider` and `Consumer` and in order to do anything with the `isTopMost` property, we need to consume the whole `DialogStackContext` which currently is just `isTopMost`, but if we added something to the stack context later.. then `TopMostConsumer` wouldn't make sense? .. But.. naming is hard and I can also see this just being a `TopMostDialogConsumer` since all it cares about right now is the `TopMost` property.. or maybe the DialogStackContext should be `TopMostDialogContext` until we actually need to expand upon it. :/

niik (changes requested):
Having the components inherit the `DialogStackContextConsumer` here doesn't feel idiomatic. React has had a pretty strong preference towards [composition over inheritance](https://reactjs.org/docs/composition-vs-inheritance.html) and I think we should stick to it. Requiring components to call `super.component*` seems easy to miss in a code review.

I think we should leverage the [built-in Consumer component](https://reactjs.org/docs/context.html#contextconsumer) such that it provides the `isTopMost` value to components who need it via their props. Something like this: https://github.com/desktop/desktop/commit/33521a068ce7787a6da76ff67e79ad7fe97fd257. This will also make things easier for us should we choose to start writing functional components in the future.

Zooming out a bit I think that only the About dialog really needs to care about whether it's top most or not (https://github.com/desktop/desktop/commit/84d4ee26f92605aaa59432b731176bc8e48c53a3). The Clone and Create Repository dialogs react to window focus because it's possible that the user has switched from the app to correct a file system error (i.e. created a path, moved something away) and I believe the dialogs should refresh their view of the outside world regardless of whether they're the topmost dialog or not. If we don't do that then we need to react to those dialogs _becoming_ the top most dialog and trigger a refresh then which seems like more work than it's worth :shrug:

Lastly, in its current state the dialogs doesn't unsubscribe from the focus/keyboard events on unmount, so we always need to add an explicit unmount, the memoized function won't take care of that for us (unlike a hook). I took a stab at adding a `WindowEvents` component for us to use so that we could subscribe to Window events just like they were synthetic React events. I dig it but given that we ended up at only one component needing to care about unsubscribing when not TopMost it doesn't seem worth it right now.

sergiou87:
> Having the components inherit the `DialogStackContextConsumer` here doesn't feel idiomatic

Yep, sorry I wasn't explicit about this. The `TopMostDialogListener` component I was suggesting would be rendered (and not inherited) by those who want to reuse/remove the boilerplate. In any case, that `Consumer` component looks much better 🤩

tidy-dev:
> Yep, sorry I wasn't explicit about this. The TopMostDialogListener component I was suggesting would be rendered (and not inherited) by those who want to reuse/remove the boilerplate. In any case, that Consumer component looks much better

Ah yeah. I didn't follow the render part.. but even so it feels very odd to me to have a component render but it doesn't actually render anything.. just being used to run functionality.

If we don't want inherit this logic, then using a prop as @niik suggested would probably be the next thing to avoid the  context property boiler plate.. but it doesn't really remove any boiler plate, just moves it around. Instead of having the context properties, it wraps the dialog component and sends it in as a prop. Then, you have the same amount of logic for managing the value in each dialog.. it just comes from a prop instead of the context. 

This feels like a messier/harder to follow use of context to me... This use of context would feel more logical if say we had a couple components or inline jsx usage that we wanted to wrap to consume the context values, Or we were already at least a component level deep from the provider.  We have the provider and consumer both in the `app.tsx`.. seems like if we want to pass it as a prop, we simply pass it in as a prop instead of jumping through consumer hoop here. 

So.. I have added another commit that I think is a hybrid of both your suggestions.
1) I pass `isTopMost` down from the `app.tsx` to the dialogs that need it.
2) I created a `isTopMostService` to have a consistent way of consuming that `isTopMost` value.

Lastly :) `Create-Repository` and `Clone-Repository` dialogs I think should still remove listeners to events when not in focus. In my opinion, it seems like unless a listener is actively needed in the background we shouldn't be running logic not related to present focus. 

> If we don't do that then we need to react to those dialogs becoming the top most dialog and trigger a refresh then which seems like more work than it's worth

I believe all I need to do is in the,  `onDialogIsTopMost` method to not only add a listener, but also invoke whatever method that listener will invoke to accomplish this.

sergiou87:
> Ah yeah. I didn't follow the render part.. but even so it feels very odd to me to have a component render but it doesn't actually render anything.. just being used to run functionality.

But isn't that exactly what we're already doing with this?
```typescript
<DialogStackContext.Provider
  key={popup.id}
  value={{
    isTopMost,
  }}
>
  {this.popupContent(popup, isTopMost)}
</DialogStackContext.Provider>
```

In the end is a hierarchy of components, some of them emit html some of them not, but all of them are tied to the same lifecycle, which is very convenient. I think it's a very common pattern and at least I've used it on iOS: with child `ViewController`s that only have behavior but don't render anything and are in the `ViewController` hierarchy so they can listen to lifecycle events and perform reusable logic. In any case, I think the `Consumer` approach makes more sense here (but in the end is the same: another React component that doesn't emit html).

I agree with your assessment of `Create-Repository` and `Clone-Repository`. Sounds reasonable to me, unless I'm missing something.

> So.. I have added another commit that I think is a hybrid of both your suggestions.

I still think using the `Consumer` is more idiomatic, within the React conventions around `Context`, and will end up in a cleaner/simpler implementation (I might be wrong about all of that, but it's my gut feeling 😂)

tidy-dev:
> I still think using the Consumer is more idiomatic

I guess I don't understand. To me, why use something we don't need to use - why add complexity we don't need? Props is what we use throughout the rest of the app and very easy to follow where this data is coming from. Adding the Context overhead only benefits us for situations like the Dialog component where we otherwise would be prop drilling for the sake of prop drilling and would have to pass props to all the 50+ popovers to do so. Thus, I think just passing props is the most easy to follow solution especially considering our code base.

> isn't that exactly what we're already doing with this?

Yes and no. :D I think that this is a unique use case for Context and it is only setting a state to subscribe (not doing custom functionality or UI). Every where else we use component composition, it is to wrap some child HTML output in some repeated html output like banners or dialogs. 

>  I think it's a very common pattern and at least I've used it on iOS

Very possibly so :P  -> I guess where I come from Angular background. We always tried to separate these concerns. If it was for UI, it was a component. If it was for reused logic, it was service (or module or reused method) or we used inheritance where it made sense. Maybe that is too ingrained in me. I also didn't realize inheritance was such a React no no.. since I was accustomed to it in past lives. 

Since we are passing the prop these components need... I think it makes easy to follow logic to do what we have done for all our other components which is to call logic in the lifecycle methods of the components that are enacting it. That is what I am doing here. The only thing different is using a class for the repeated check against the `isTopMost` which I also don't think breaks our patterns of exporting reusable methods across components. This I think makes sense to help other developers who may want to subscribe to this follow a similar pattern.

sergiou87:
So I think the reason we're using `Context` is precisely to avoid passing props to every single dialog needing it, and instead just let the dialogs take them from their implementation, so it feels kind of weird having both things: a `Context` to avoid passing props down, and then passing a prop down 🤔 

> We always tried to separate these concerns.

That is a good point. In this case, we're doing so _to some extent_ by encapsulating the "top most" logic into a component and exposing the minimum API, but I totally understand it's still a mix within the `render` method.

> I also didn't realize inheritance was such a React no no.. since I was accustomed to it in past lives.

I think that's a broader principle/approach: "favor composition over inheritance"
- https://wiki.c2.com/?CompositionInsteadOfInheritance
- https://en.wikipedia.org/wiki/Composition_over_inheritance

Of course, as everything, it depends on the specific scenario and codebase.

In any case, and just to clarify, to me this is more of a philosophical discussion, where I'd rather leverage the component system to listen to `isTopMost` changes and reuse lifecycle-dependant logic, but I still recognize this is a one-off exception (hopefully temporary) and I don't intend to block the PR on that, not at all!

app/src/lib/popup-manager.ts
```diff
@@ -47,8 +47,19 @@ export class PopupManager {
...
+   * Returns all the popups in the stack. If there are error popups, it returns
+   *  them on the top of the stack (the end of the array -> last on, last off).
+   */
+  public get allPopups(): ReadonlyArray<Popup> {
     const errorPopups = this.getPopupsOfType(PopupType.Error)
-    return errorPopups.at(-1) ?? this.popupStack.at(-1) ?? null
+    const nonErrorPopups = this.popupStack.filter(
+      p => p.type !== PopupType.Error
+    )
+    return [...nonErrorPopups, ...errorPopups]
```
niik: Given that this getter is called for each AppStore emit I'd rather we avoided doing array mutation here if we can avoid it, lots of unnecessary GC pressure. I'm not saying we need to fix this in this PR. If you have time I promise I'll re-review it tomorrow morning but if not it'd be great if we could put this on a TODO list somewhere and change this such that we constructs allPopups when the underlying array mutates.

app/src/ui/app.tsx
```diff
@@ -1397,13 +1399,37 @@ export class App extends React.Component<IAppProps, IAppState> {
...
+    return (
+      <>
+        {allPopups.map(popup => {
+          const isTopMost = this.state.currentPopup?.id === popup.id
+          return (
+            <DialogStackContext.Provider
+              key={popup.id}
+              value={{
+                isTopMost,
+              }}
```
niik:
```suggestion
              value={{ isTopMost }}
```

niik (dismissed):
I can't shake the feeling that the IsTopMostService feels out of place here but definitely not enough that I'm going to hold this PR up for it. I think it's the class structure that irks me the most and I think we could rewrite that into a function to make it feel more at home but no biggie.

app/src/lib/popup-manager.ts
```diff
@@ -98,11 +109,25 @@ export class PopupManager {
...
+    const indexLastError = this.popupStack.findIndex(
+      p => p.type === PopupType.Error
+    )
+
+    if (indexLastError === -1) {
+      this.popupStack.push(popup)
+      return
+    }
+
+    this.popupStack.splice(indexLastError, 0, popup)
```
niik: Now that we expose the internal `popupStack` directly via `allPopups` we need to ensure that it doesn't mutate. Consumers receiving a ReadonlyArray have a reasonable expectation that it won't change underneath them which can lead to subtle bugs in a snapshot-state model like ours. We should make the private `popupStack` be a `ReadonlyArray` as well to enforce this within this class.

```suggestion
    if (this.popupStack.at(-1)?.type !== PopupType.Error) {
      this.popupStack = this.popupStack.concat(popup)
      return
    }

    const errorPopups = this.getPopupsOfType(PopupType.Error)
    const nonErrorPopups = this.popupStack.filter(
      p => p.type !== PopupType.Error
    )
    this.popupStack = [...nonErrorPopups, popup, ...errorPopups]
```

app/src/ui/clone-repository/clone-repository.tsx
```diff
@@ -200,7 +216,11 @@ export class CloneRepository extends React.Component<
       this.updateUrl(initialURL)
     }
 
-    window.addEventListener('focus', this.onWindowFocus)
+    this.checkIsTopMostDialog(this.props.isTopMost)
+  }
+
+  public componentWillUnmount(): void {
+    window.removeEventListener('focus', this.onWindowFocus)
```
niik: Could/should we leverage the helper here as well instead of duplicating?

```suggestion
    this.checkIsTopMostDialog(false)
```

app/src/ui/add-repository/create-repository.tsx
```diff
@@ -158,7 +172,11 @@ export class CreateRepository extends React.Component<
     this.updateReadMeExists(path, this.state.name)
   }
 
-  public componentWillUnmount() {
+  public componentDidUpdate(): void {
+    this.checkIsTopMostDialog(this.props.isTopMost)
+  }
+
+  public componentWillUnmount(): void {
     window.removeEventListener('focus', this.onWindowFocus)
```
niik:
```suggestion
    this.checkIsTopMostDialog(false)
```

app/src/ui/dialog/is-top-most.tsx
```diff
@@ -0,0 +1,14 @@
+import memoizeOne from 'memoize-one'
+
+export function isTopMostDialog(
```
niik, line 6: This feels so much better, thanks for taking another swing at it ✨ Do you think you could add just a line or two of documentation for future us?

niik (changes requested):

niik (approved):
Thanks so much for fighting through this review @tidy-dev :sparkles:

MotoFitz (commented):
Hi, I'm looking for motofitz I'm Damon Fitzgerald I want to tell you I miss you and love you. Come home.

1096103013 (commented):
##`` / #14682
#15798

app/src/ui/about/about.tsx
```diff
@@ -67,18 +66,10 @@ interface IAboutState {
...
-export class About extends React.Component<IAboutProps, IAboutState> {
-  public static contextType = DialogStackContext
-  public declare context: React.ContextType<typeof DialogStackContext>
-
-  private checkWhetherDialogIsTopMost = memoizeOne((isTopMost: boolean) => {
-    if (isTopMost) {
-      this.onDialogIsTopMost()
-    } else {
-      this.onDialogIsNotTopMost()
-    }
```
Hamza99937: exist ..

Hamza99937 (commented):
اريد الوقت من التعليم

Hamza99937 (commented):
منبثقة

PR #15690, opened 2022-11-27
Open selected files

Shivareddy-Aluri:
Closes #[issue number]

## Description

tried to introduce feature which allows users to open multiple selected files in the default / selected editor.

## Release notes

Notes:

### DIFF
--- a/app/src/lib/editors/launch.ts
+++ b/app/src/lib/editors/launch.ts
@@ -28,19 +28,16 @@ export async function launchExternalEditor(
     // Desktop app is closed.
     detached: true,
   }
-
+  const pathsToOpen = fullPath.split(',')
   if (editor.usesShell) {
-    const pathsToOpen = fullPath.split(',')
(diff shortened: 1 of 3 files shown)


Shivareddy-Aluri:
Hi @sergiou87 i tried to introduce feature to open multiple selected files, which i felt would be very useful for users.
the code works for all the editor except intellij family editors. only in intellij idea family, if i try to open multiple files only few of them open.
i tried using  exec instead of spawn and looks like all the files will open in intellij family editors too. can you suggest anything here please?

sergiou87:
Sorry @Shivareddy-Aluri , it was a busy month and forgot to reply. I think you will need to find out how to make the JetBrains IDEs open multiple files at once (maybe it requires a very specific command line parameter) and then try to fit that _somehow_ in our editors code.

niik:
Hey @Shivareddy-Aluri, since we haven't heard from you in a while I'm going to close this to clean up our PR page a bit but if you get around to solving the issue you're having here feel free to open a new PR!

PR #15695, opened 2022-11-29
PR Preview: Create pull request button tweaks

tidy-dev:
## Description
While using this on beta, I noticed that the dialog doesn't close on submit (I probably didn't notice since it jumps out to browser on top of my Desktop.. but still 🤦 ). Also, if there is already a pull request open for the branch, the button should not saw "Create pull request" -> "View Pull Request". I added the external link octicon to the button in this case to make it a little more clear that the button will open it on dotcom.  Other thoughts are we could change the title to from " Open a Pull Request" to "[Title of Pull Request] #number"

## Release notes
Notes: [Improved] On Preview Pull Request dialog, submit button closes the dialog.
[Improved] On Preview Pull Request dialog, the dialog submit button opens existing pull request on GitHub when it exists..

### DIFF
--- a/app/src/ui/open-pull-request/open-pull-request-dialog.tsx
+++ b/app/src/ui/open-pull-request/open-pull-request-dialog.tsx
@@ -62,6 +62,9 @@ interface IOpenPullRequestDialogProps {
    * it's SHA  */
   readonly nonLocalCommitSHA: string | null
 
+  /** Whether the current branch already has a pull request*/
+  readonly currentBranchHasPullRequest: boolean
+
   /** Called to dismiss the dialog */
   readonly onDismissed: () => void
 }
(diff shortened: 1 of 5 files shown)


app/src/ui/open-pull-request/open-pull-request-dialog.tsx
```diff
@@ -179,25 +193,42 @@ export class OpenPullRequestDialog extends React.Component<IOpenPullRequestDialo
   }
 
   private renderFooter() {
-    const { mergeStatus, commitSHAs } = this.props.pullRequestState
-    const gitHubRepository = this.props.repository.gitHubRepository
+    const { currentBranchHasPullRequest, pullRequestState, repository } =
+      this.props
+    const { mergeStatus, commitSHAs } = pullRequestState
+    const isHostedOnGitHub = isRepositoryHostedOnGitHub(repository)
```
niik: Can we use the existing type guard `isRepositoryWithGitHubRepository` here instead?

niik (changes requested):

niik (approved):
:sparkles:

PR #15704, opened 2022-11-30
PR Preview: Handle no default branch.

tidy-dev:
## Description
By happen stance, I was doing some kicking of the tires and was just trying the PR previewer out on various repos of mine to see if I ran into anything funny. Found a repo where clicking Preview Pull Request just did nothing.... Found out it was because that repository did not have its default branch pulled locally, thus, the store said its default branch was null. I did not realize this as valid state at the time of development (thought my defaultBranch === null, return was just type checking). Turns out this is easily replicable by simply deleting the local default branch.

This PR adds a no default branch state to the pull request dialog so that the user picks the comparison branch when no default branch is found.

Other thoughts:
We could add a check for remote default branches to the [resolving of the default branch logic](https://github.com/desktop/desktop/blob/development/app/src/lib/stores/git-store.ts#L479) in the git store so that this scenario is not so easy to get into. I tested it out and tried a few things that I know rely on the default branch and it seems to work well (just shows origin/defaultBranchName) ... but it also felt like something that could impact lots of things so I didn't move forward with that solution.. and even still a user could technically get into a state where there is no default (remote doesn't resolve, and defaults to a global config or 'main' that just doesn't exist for the repo). 

## Screenshots
https://user-images.githubusercontent.com/75402236/204837911-1fe8eb50-794f-40e9-9fa5-9d53841df45a.mov

## Release notes
Notes: [Fixed] Preview Pull Request opens when there is not a local default branch

### DIFF
--- a/app/src/lib/app-state.ts
+++ b/app/src/lib/app-state.ts
@@ -967,7 +967,7 @@ export interface IPullRequestState {
    * The base branch of a a pull request - the branch the currently checked out
    * branch would merge into
    */
-  readonly baseBranch: Branch
+  readonly baseBranch: Branch | null
 
   /** The SHAs of commits of the pull request */
   readonly commitSHAs: ReadonlyArray<string> | null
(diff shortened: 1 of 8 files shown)


sergiou87 (approved):
Code looks good, tested it and works as expected 🎉 :shipit:

emt420769:
420

PR #15747, opened 2022-12-07
PR Preview: Add Preview Pull Request Metrics

tidy-dev:
## Description
This PR Adds the following metric tracking:
- `previewedPullRequestCount`: This is to see how many users are using the feature.
- `createPullRequestFromPreviewCount`: This is to see how many users are creating pull requests on dotcom from the feature. The existing metric of `createPullRequestCount` which will track all times this occurs.
   - Answer: Will users use this or just go back to always going straight to dotcom? 
  

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/lib/stats/stats-store.ts
+++ b/app/src/lib/stats/stats-store.ts
@@ -112,6 +112,7 @@ const DefaultDailyMeasures: IDailyMeasures = {
   guidedConflictedMergeCompletionCount: 0,
   unguidedConflictedMergeCompletionCount: 0,
   createPullRequestCount: 0,
+  createPullRequestFromPreviewCount: 0,
   rebaseConflictsDialogDismissalCount: 0,
   rebaseConflictsDialogReopenedCount: 0,
   rebaseAbortedAfterConflictsCount: 0,
@@ -219,6 +220,7 @@ const DefaultDailyMeasures: IDailyMeasures = {
   submoduleDiffViewedFromChangesListCount: 0,
   submoduleDiffViewedFromHistoryCount: 0,
   openSubmoduleFromDiffCount: 0,
+  previewedPullRequestCount: 0,
 }
 
 interface IOnboardingStats {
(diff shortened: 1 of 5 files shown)


IcebergVerse (commented):
add-preview-pull-request-metrics

app/src/lib/stats/stats-store.ts
```diff
@@ -1981,6 +1993,15 @@ export class StatsStore implements IStatsStore {
       log.error(`Error reporting opt ${direction}:`, e)
     }
   }
+
+  /**
+   * Increments the `createPullRequestFromPreviewCount` metric
```
sergiou87:
```suggestion
   * Increments the `previewedPullRequestCount` metric
```

app/src/ui/open-pull-request/open-pull-request-dialog.tsx
```diff
@@ -83,8 +83,8 @@ export class OpenPullRequestDialog extends React.Component<IOpenPullRequestDialo
       dispatcher.showPullRequest(repository)
     } else {
       dispatcher.createPullRequest(repository)
-      // TODO: create pr from dialog pr stat?
       dispatcher.recordCreatePullRequest()
+      dispatcher.recordCreatePullRequestFromPreview()
```
sergiou87, line 88: It'll be important to mention _somewhere_ that these two metrics will be tracked at the same time in this case.

tidy-dev: I added it to the `createPullRequest` metric documentation. To me if I when looking for "how many times do users create a pull request on dotcom", I would expect `createPullRequestCount` to answer that question. But, I can see that if I also so the metric below it, I could wonder if they were included in that count.. Thus, I think this is the most logical place to note it.

sergiou87 (dismissed):
Thanks for those changes! 🙌

sergiou87 (dismissed):

sergiou87 (approved):

PR #15754, opened 2022-12-08
Include remote branches in search for default branch

niik:
## Description

Spurred by @tidy-dev's question in https://github.com/desktop/desktop/pull/15704:

> We could add a check for remote default branches to the [resolving of the default branch logic](https://github.com/desktop/desktop/blob/development/app/src/lib/stores/git-store.ts#L479) in the git store so that this scenario is not so easy to get into

and apt observation of risk

> it also felt like something that could impact lots of things so I didn't move forward with that solution.. and even still a user could technically get into a state where there is no default (remote doesn't resolve, and defaults to a global config or 'main' that just doesn't exist for the repo).

I wanted to explore what it would look like to make a remote branch be able to be the default branch. I believe that was the original intention but it somehow never came to be.

This implementation is what I believe the default branch heuristics should have been all along. Next steps if we agree that this is a sane approach is to go through all places where we currently use `defaultBranch` and see if any of them erroneously make the assumption that it'll be local.

### Screenshots

image

## Release notes

Notes:


app/src/lib/find-default-branch.ts
```diff
@@ -0,0 +1,56 @@
...
+
+      if (remoteRef && branch.upstream === remoteRef) {
+        // Give preference to local branches that target the upstream
+        // default branch that also match the name. In other words, if there
+        // are two local branches which both track the origin default branch
+        // we'll prefer a branch which is also named the same as the default
+        // branch name.
+        if (!localTrackingHit || branch.name === defaultBranchName) {
+          localTrackingHit = branch
+        }
```
tidy-dev, line 55: Nit - not likely we have many branches. But, since we are giving preference to the `localTrackingHit`, seems we can short circuit the `for..loop` once it is found.

```suggestion
        if (!localTrackingHit || branch.name === defaultBranchName) {
          localTrackingHit = branch
          break
        }
```

app/src/lib/find-default-branch.ts
```diff
@@ -0,0 +1,56 @@
...
+ * contribution target remote HEAD we'll fall back to looking for the remote
+ * branch itself.
+ */
+export async function findDefaultBranch(
+  repository: Repository,
+  branches: ReadonlyArray<Branch>
+) {
+  const originHead = await getRemoteHEAD(repository, 'origin')
+  const defaultBranchName = originHead ?? (await getDefaultBranch())
+  const remoteRef = originHead ? `origin/${originHead}` : undefined
```
sergiou87: This should be `origin`/`upstream` depending on `getForkContributionTarget(repository)`, right?

sergiou87 (changes requested):
This looks good!! Just found it doesn't check the contribution target.

app/src/lib/find-default-branch.ts
flyi36: Originedeffault

app/src/lib/find-default-branch.ts
```diff
@@ -0,0 +1,56 @@
+import { Branch, BranchType } from '../models/branch'
+import { Repository } from '../models/repository'
+import { getRemoteHEAD } from './git'
+import { getDefaultBranch } from './helpers/default-branch'
```
Faisal1809, line 4: metadata

app/src/lib/find-default-branch.ts
niik: I don't think we can because `localTrackingHit` can be set more than once. It'll be set as soon as we find a branch that's tracking the remote HEAD but if there are more than one local branch tracking the remote HEAD and we see one where the local branch name that also matches the default branch name we'll re-assign it to that.

app/src/lib/find-default-branch.ts
niik: Gosh, this little innocent looking question had me on a wild goose chase for so long! I think you're right so I added this in https://github.com/desktop/desktop/pull/15754/commits/5389aa5ba865e056f2f26f6310c603a17a1d13ee.

The thing that tripped me up was the value of `upstreamDefaultBranch` post this change but I've come to the conclusion that it still has value since it only ever returns remote branches and `defaultBranch` can return both local and remote. I'm still not convinced that there isn't a path to get rid of `upstreamDefaultBranch` but that will have to wait for a different PR

app/src/lib/find-default-branch.ts
```diff
@@ -1,8 +1,23 @@
...
 
+async function getContributionTargetHead(
+  repository: Repository
+): Promise<[string, string | null]> {
+  if (getForkContributionTarget(repository) === ForkContributionTarget.Parent) {
+    const head = await getRemoteHEAD(repository, 'upstream')
+
+    if (head !== null) {
+      return ['upstream', head]
+    }
```
sergiou87: Shouldn't we always return here? The calling method will then set the `remoteRef` to `undefined` if there was no `head`. I _think_ it makes sense to stick to the `upstream` HEAD if the fork contribution target is the parent repo, even if for whatever reason we can't find its HEAD. Does that make sense?
```suggestion
    return ['upstream', head]
```

niik: Yep, it sure does. I took another stab this and at the same time addressed the (very rare) case when the default remote isn't named `origin`.

sergiou87 (approved):
✨ the refactor looks great and what I tested worked as expected! 🚀

niik:
Thanks for the thorough review here @sergiou87!

PR #15768, opened 2022-12-12
PR Preview: Only default remote branches as base branches

tidy-dev:
## Description
This PR does a few things:
1. Disables Previewing PRs on local repositories since you can't actually make a pull request. 
     - I had originally just disabled the create pr button in the dialog such that you could still use the tool.. but have changed my mind on that stance because it is not the intention of the tool to be merge comparison tool and now wouldn't work because we don't provide local branches as base branch options.
2. Filtered the base branch options down to branches that share the repositories default remote. The exception to this is if the repository is a forked repository contributing to the parent, then the base branches include origins of upstream.
3. In testing out forks, I realized that the create pull request button in the PR dialog did not target the right repository. This PR makes it so that if a fork targets its parent, it will open a pr from the fork repo to the parent repo. Otherwise, it will open the pr to the fork.
   
### Screenshots
https://user-images.githubusercontent.com/75402236/207122160-c797e6e1-b3dd-4fd3-9d58-fbd7b9d57eed.mov

## Release notes
Notes: [Improved] Pull request preview dialog only uses remote branches for base branch options.

### DIFF
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -47,7 +47,7 @@ import {
   isRepositoryWithGitHubRepository,
   RepositoryWithGitHubRepository,
   getNonForkGitHubRepository,
-  isRepositoryAForkContributingToParent,
+  isForkedRepositoryContributingToParent,
 } from '../../models/repository'
 import {
   CommittedFileChange,
@@ -6102,7 +6102,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const { parent, owner, name, htmlURL } = gitHubRepository
     const isForkContributingToParent =
-      isRepositoryAForkContributingToParent(repository)
+    isForkedRepositoryContributingToParent(repository)
 
     const baseForkPreface =
       isForkContributingToParent && parent !== null
@@ -7418,7 +7418,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
     /*  We only want branches that are also on dotcom such that, when we ask a
      *  user to create a pull request, the base branch also exists on dotcom.
      */
-    const remote = isRepositoryAForkContributingToParent(repository)
+    const remote = isForkedRepositoryContributingToParent(repository)
       ? UpstreamRemoteName
       : gitStore.defaultRemote?.name
     const prBaseBranches = allBranches.filter(
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -6102,7 +6102,7 @@ export class AppStore extends TypedBaseStore<IAppState> {
 
     const { parent, owner, name, htmlURL } = gitHubRepository
     const isForkContributingToParent =
-    isForkedRepositoryContributingToParent(repository)
+      isForkedRepositoryContributingToParent(repository)
 
     const baseForkPreface =
       isForkContributingToParent && parent !== null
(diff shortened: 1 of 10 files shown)


app/src/lib/stores/app-store.ts
```diff
@@ -6098,15 +6100,30 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    const compareForkPreface = isForkContributingToParent
+      ? `${owner.login}:${name}:`
+      : ''
+
+    const encodedCompareBranch =
+      compareForkPreface + encodeURIComponent(compareBranch.nameWithoutRemote)
+
     const compareString = `${encodedBaseBranch}${encodedCompareBranch}`
-    const baseURL = `${gitHubRepository.htmlURL}/pull/new/${compareString}`
+    const baseURL = `${htmlURL}/pull/new/${compareString}`
```
tidy-dev, line 6127: I found that the existing create pull request on dotcom path that does not pass in a base branch will always open to dotcom with a base branch of the parent repository. This surprised me kind of because in the scenario of not contributing to the parent this URL created here for my desktop fork will be `https://github.com/tidy-dev/desktop/pull/new/tidy-dev:desktop:Hello`., but, GitHub redirects to `https://github.com/desktop/desktop/compare/development...tidy-dev:desktop:Hello?expand=1`

However, if we were to pass a base branch in and apply the fork preface as this PR does - `https://github.com/tidy-dev/desktop/pull/new/tidy-dev:desktop:development...tidy-dev:desktop:Hello`; it will correctly go to `https://github.com/tidy-dev/desktop/compare/tidy-dev:desktop:development...tidy-dev:desktop:Hello?expand=1`

I may open a PR if this one goes through to do that because I think that is the expected behavior when not contributing to the parent.

app/src/models/repository.ts
```diff
@@ -213,3 +213,15 @@ export function getForkContributionTarget(
     ? repository.workflowPreferences.forkContributionTarget
     : ForkContributionTarget.Parent
 }
+
+/**
+ * Returns whether the fork is contributing to the parent
+ */
+export function isRepositoryAForkContributingToParent(
```
sergiou87: Just a suggestion (I try to avoid having `A` in identifiers 😳 )
```suggestion
export function isForkedRepositoryContributingToParent(
```

app/src/lib/stores/app-store.ts
sergiou87: 🤯 

> I may open a PR if this one goes through to do that because I think that is the expected behavior when not contributing to the parent.

Agree 👍 However, maybe this is a bug in dotcom and should be reported/fixed there?

sergiou87 (changes requested):
Looks good overall! 👏  I just had a suggestion for the name of a function.

Other than that, I wonder if the `Sorry, I can't find that branch` message in the branch list should make it clear to users that local branches have been filtered out. Otherwise, it might be confusing to look for a branch that you know exist, not finding it, and not knowing why exactly.

What do you think?

app/src/ui/open-pull-request/open-pull-request-header.tsx
```diff
@@ -104,6 +104,12 @@ export class OpenPullRequestDialogHeader extends React.Component<
             allBranches={prBaseBranches}
             recentBranches={prRecentBaseBranches}
             onChange={onBranchChange}
+            noBranchesMessage={
+              <>
+                Sorry, I can't find a that remote branch. <br />
```
sergiou87: Small typo:
```suggestion
                Sorry, I can't find that remote branch. <br />
```

sergiou87 (changes requested):
Nice, thank you! Just found a typo in the new message 😳

sergiou87 (approved):
Thanks for those changes!! The only thing that called my attention is the "not found" text is not centered now:
image

Happy to consider that nit fixed in a different PR, so I'll approve this one and let you decide what to do about it 😄

Issue #15772, opened 2022-12-13
Add Fleet as external editor option


valentinegb:
### The feature request

Currently, there is no external editor option for [Fleet](https://jetbrains.com/fleet), which is the editor I primarily use.

### Proposed solution

If Fleet was added as an option for the external editor setting, this would be a convenient way to quickly open projects in it.

niik:
Hey @valentinegb, thanks for reaching out. Given that we support a whole slew of JetBrains editors chances are adding support for Fleet wouldn't be all that tricky if you'd like to take a stab at it. There's [some documentation](https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md) on adding new editor which would be a good starting point.

valentinegb:
Alright, I'll take a look.

valentinegb:
It's not quite working... I got it to recognize I have Fleet installed and open Fleet, but it doesn't open the repository I'm on specifically, it just opens whatever my last project was.

Fleet does have a CLI that opens specific files and directories (`fleet [path]`). Is there a way to do that instead within GitHub Desktop?

tsvetilian-ty:
👋  @valentinegb JetBrains Fleet is only available through JetBrains Toolbox, and GitHub Desktop doesn't support it. I opened a PR with support for Toolbox and respectively Fleet.

valentinegb:
Pull request was merged, closing the issue

PR #15780, opened 2022-12-13
Support for JetBrains Toolbox and JetBrains Fleet editor

### LINKED ISSUE #12912: Support for latest Jetbrains IDE installed by Toolbox
### Describe the bug
Trying to choose one of Jetbrains IDE as default editor, but non of them is listed. Jetbrains IDE are PyCharm, WebStorm, Intellij IDEA (all of them installed by Jetbrains Toolbox)

### Version & OS

Version 2.9.3 (x64)
Windows 10 Pro, build 19043.1202

1. Go to "File" menu on top.
2. Click on "Options"
3. Go to "Integrations"
4. Open list of available external editors

### Expected behavior

At least, one of Jetbrains IDE should be listed

### Actual behavior

Only Notepad++ is listed`

### Screenshots

### Logs

tsvetilian-ty:
Closes #12912 #15772 

## Description
- Add support for [JetBrains Toolbox](https://www.jetbrains.com/toolbox-app/)
  - The implementation detects only the default configuration for the scripts
- Add [JetBrains Fleet](https://www.jetbrains.com/fleet/)

### Screenshots
fleet

### DIFF
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -144,6 +144,10 @@ const editors: IDarwinExternalEditor[] = [
     name: 'Lite XL',
     bundleIdentifiers: ['com.lite-xl'],
   },
+  {
+    name: 'JetBrains Fleet',
+    bundleIdentifiers: ['Fleet.app'],
+  },
 ]
 
 async function findApplication(
--- a/app/src/lib/editors/darwin.ts
+++ b/app/src/lib/editors/darwin.ts
@@ -145,7 +145,7 @@ const editors: IDarwinExternalEditor[] = [
     bundleIdentifiers: ['com.lite-xl'],
   },
   {
-    name: 'JetBrains Fleet',
+    name: 'Fleet',
     bundleIdentifiers: ['Fleet.app'],
   },
 ]
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -62,10 +62,12 @@ type WindowsExternalEditor = {
   readonly publisher: string
 
   /**
-   * Default shell script name for JetBrains Products
+   * Default shell script name for JetBrains Product
    * To get the script name go to:
    * JetBrains Toolbox > Editor settings > Shell script name
-   * Note: Don't use on JetBrains community and edu editions
+   *
+   * Go to `/docs/techical/editor-integration.md` for more information on
+   * how to use this field.
    */
   readonly jetBrainsToolboxScriptName?: string
 } & WindowsExternalEditorPathInfo
--- a/app/src/lib/editors/win32.ts
+++ b/app/src/lib/editors/win32.ts
@@ -18,16 +18,6 @@ interface IWindowsAppInformation {
 
 type RegistryKey = { key: HKEY; subKey: string }
 
-type WindowsExternalEditorJetbrainsToolbox = {
-  /**
-   * Default shell script name for JetBrains Products
-   * To get the script name go to:
-   * JetBrains Toolbox > Editor settings > Shell script name
-   * Note: Don't use on JetBrains community and edu editions
-   */
-  readonly toolboxShellScriptName?: string
-}
-
 type WindowsExternalEditorPathInfo =
   | {
       /**
@@ -70,8 +60,15 @@ type WindowsExternalEditor = {
 
   /** Value of the Publisher registry key that belongs to this editor. */
   readonly publisher: string
-} & WindowsExternalEditorPathInfo &
-  WindowsExternalEditorJetbrainsToolbox
+
+  /**
+   * Default shell script name for JetBrains Products
+   * To get the script name go to:
+   * JetBrains Toolbox > Editor settings > Shell script name
+   * Note: Don't use on JetBrains community and edu editions
+   */
+  readonly jetBrainsToolboxScriptName?: string
+} & WindowsExternalEditorPathInfo
 
 const registryKey = (key: HKEY, ...subKeys: string[]): RegistryKey => ({
   key,
@@ -323,7 +320,7 @@ const editors: WindowsExternalEditor[] = [
     name: 'JetBrains Webstorm',
     registryKeys: registryKeysForJetBrainsIDE('WebStorm'),
     executableShimPaths: executableShimPathsForJetBrainsIDE('webstorm'),
-    toolboxShellScriptName: 'webstorm',
+    jetBrainsToolboxScriptName: 'webstorm',
     displayNamePrefix: 'WebStorm',
     publisher: 'JetBrains s.r.o.',
   },
@@ -331,7 +328,7 @@ const editors: WindowsExternalEditor[] = [
     name: 'JetBrains Phpstorm',
     registryKeys: registryKeysForJetBrainsIDE('PhpStorm'),
     executableShimPaths: executableShimPathsForJetBrainsIDE('phpstorm'),
-    toolboxShellScriptName: 'phpstorm',
+    jetBrainsToolboxScriptName: 'phpstorm',
     displayNamePrefix: 'PhpStorm',
     publisher: 'JetBrains s.r.o.',
   },
@@ -339,7 +336,7 @@ const editors: WindowsExternalEditor[] = [
     name: 'Android Studio',
     registryKeys: [LocalMachineUninstallKey('Android Studio')],
     installLocationRegistryKey: 'UninstallString',
-    toolboxShellScriptName: 'studio',
+    jetBrainsToolboxScriptName: 'studio',
     executableShimPaths: [
       ['..', 'bin', `studio64.exe`],
       ['..', 'bin', `studio.exe`],
@@ -363,7 +360,7 @@ const editors: WindowsExternalEditor[] = [
     name: 'JetBrains Rider',
     registryKeys: registryKeysForJetBrainsIDE('JetBrains Rider'),
     executableShimPaths: executableShimPathsForJetBrainsIDE('rider'),
-    toolboxShellScriptName: 'rider',
+    jetBrainsToolboxScriptName: 'rider',
     displayNamePrefix: 'JetBrains Rider',
     publisher: 'JetBrains s.r.o.',
   },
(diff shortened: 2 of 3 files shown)


app/src/lib/editors/darwin.ts
```diff
@@ -144,6 +144,10 @@ const editors: IDarwinExternalEditor[] = [
     name: 'Lite XL',
     bundleIdentifiers: ['com.lite-xl'],
   },
+  {
+    name: 'JetBrains Fleet',
```
valentinegb: Other JetBrains IDEs on this list do not have the JetBrains prefix, neither should Fleet. This is only on Mac though, for some reason the JetBrains prefix has been normalized on Windows, apparently.

tsvetilian-ty: You're absolutely right 🙌

Kokomo1982:
> Closes #12912 #15772
> 
> ## Description
> * Add support for [JetBrains Toolbox](https://www.jetbrains.com/toolbox-app/)
>   
>   * The implementation detects only the default configuration for the scripts
> * Add [JetBrains Fleet](https://www.jetbrains.com/fleet/)
> 
> ### Screenshots
> fleet [ ](https://user-images.githubusercontent.com/9341546/207426925-7a428fc1-dfe2-4f4a-932b-51b2d7d45b31.gif) fleet [ ](https://user-images.githubusercontent.com/9341546/207426925-7a428fc1-dfe2-4f4a-932b-51b2d7d45b31.gif)

``

valentinegb (commented):
This currently only works on Windows.

tsvetilian-ty:
Does Fleet show in the dropdown?

valentinegb:
It does, but it doesn't use the CLI, it launches the app without opening the specified directory.

tsvetilian-ty:
Could you make a video and share it? Because I'm unable to reproduce the issue 🤔

valentinegb:
Sure, but I might not be able to get back onto my computer until about an hour-
Also, I'm not really familiar with this project or its stack but it looks to me like the code that enables using the generated Toolbox scripts is only in the `win32.ts` file, which, I would assume, is only used on Windows.

tsvetilian-ty:
No worries about the video, I just want to see how it behaves on your side. Here is a video of my VM
fleet-anim

Yes, `win32.ts` is Windows only, but that is where the issue with JetBrains Toolbox is. We use the registry to detect an editor on Windows, and IDEs installed through JetBrains Toolbox don't have registry entries, so they are invisible to GitHub  Desktop.

valentinegb:
Hmm, not quite what happens for me.

https://user-images.githubusercontent.com/35977727/207465886-771d64e2-d858-4316-8aa9-bcdfe22c5bb8.mov

*Edited so the video works better*

tsvetilian-ty:
It is slow because it's on VM, and it's a gif (15fps). 
Just a suggestion that might be why you have issues when you checkout the branch. It's a good idea to run `yarn build:dev` (or `npm run build:dev`) to ensure you have the latest state of the build cached.

valentinegb:
No I meant the video *I* sent- and I ran `yarn build:dev` before I started it, if that's what you mean.

tsvetilian-ty:
I couldn't see the video at all, so I had to open it on mobile 😆 Perhaps you could try to run GitHub Desktop from external terminal 🤔

valentinegb:
Oh haha- this is a new video, it should work better.

https://user-images.githubusercontent.com/35977727/207476235-685cdb96-31b2-43d0-b0a3-70336fd1d58b.mp4

So when I ran it in my terminal it did work. I still wonder why it didn't before... but I guess it probably doesn't really matter.

(Enjoy the mini-tour of my files lol)

app/src/lib/editors/win32.ts
```diff
@@ -60,7 +70,8 @@ type WindowsExternalEditor = {
 
   /** Value of the Publisher registry key that belongs to this editor. */
   readonly publisher: string
-} & WindowsExternalEditorPathInfo
+} & WindowsExternalEditorPathInfo &
+  WindowsExternalEditorJetbrainsToolbox
```
sergiou87: Since this is just a field you added, I don't think we need an additional type to `&` here. `WindowsExternalEditorPathInfo` was different because it has two different types inside.

However, I'm not sure about the new `toolboxShellScriptName` optional field because it's not tied/required for JetBrains IDEs, but I don't see an easy way of getting TypeScript to enforce that.

In any case, I think the field name should be more JetBrains specific, something like `jetBrainsToolboxScriptName`

app/src/lib/editors/win32.ts
```diff
@@ -18,6 +18,16 @@ interface IWindowsAppInformation {
 
 type RegistryKey = { key: HKEY; subKey: string }
 
+type WindowsExternalEditorJetbrainsToolbox = {
+  /**
+   * Default shell script name for JetBrains Products
+   * To get the script name go to:
+   * JetBrains Toolbox > Editor settings > Shell script name
+   * Note: Don't use on JetBrains community and edu editions
```
sergiou87: Why  not? 🤔  Probably worth mentioning that

app/src/lib/editors/win32.ts
```diff
@@ -465,6 +493,31 @@ async function findApplication(editor: WindowsExternalEditor) {
...
+        const exists = await pathExists(editorPathInToolbox)
+        if (exists) {
+          return editorPathInToolbox
+        }
+      }
+    }
+  }
+
   return null
 }
```
sergiou87: I think I'd extract this to a function:
```suggestion
  return findJetBrainsToolboxApplication(editor)
}

async function findJetBrainsToolboxApplication(editor: WindowsExternalEditor) {
  if (!editor.toolboxShellScriptName) {
    return null
  }
  
  const toolboxRegistryReference = [
    CurrentUserUninstallKey('toolbox'),
    Wow64LocalMachineUninstallKey('toolbox'),
  ]

  for (const { key, subKey } of toolboxRegistryReference) {
    const keys = enumerateValues(key, subKey)

    if (keys.length > 0) {
      const editorPathInToolbox = Path.join(
        getKeyOrEmpty(keys, 'UninstallString'),
        '..',
        '..',
        'scripts',
        `${editor.toolboxShellScriptName}.cmd`
      )
      const exists = await pathExists(editorPathInToolbox)
      if (exists) {
        return editorPathInToolbox
      }
    }
  }

  return null
}
```

docs/technical/editor-integration.md
```diff
@@ -203,6 +204,25 @@ location with an interface that doesn't change between updates.
 Desktop will confirm this file exists on disk before launching - if it's
 missing or lost it won't let you launch the external editor.
 
+### Support for JetBrains Toolbox editors
+
+Now GitHub Desktop support editors installed through JetBrains Toolbox. 
+The technique used to achieve that is using `toolboxShellScriptName` parameter
+to check if, in the default section for scripts in JetBrainsm Toolbox a script
```
sergiou87: Nit:
```suggestion
to check if, in the default section for scripts in JetBrainsm Toolbox, a script
```

sergiou87 (changes requested):
Thank you for this!! ❤️ 

To be honest, my intention was to thoroughly review #14843 but that's a bigger chunk of work. This PR addresses a long-standing issue with JetBrains Toolbox users in a more surgical way  🎉 

The main limitation I see is it only works with the default script names, but I'm not too worried since Desktop support for other external editors is quite rigid anyway.

And a nice to have would be making that new property required/available for JetBrains IDEs but not for the rest, but I'm not sure we can do that in TypeScript without over-engineering a type, so… 🤷‍♂️  Maybe @niik has some smart idea for that 😄

app/src/lib/editors/win32.ts
tsvetilian-ty: Initially, I thought it might require more fields, so I added the type. But in its final form, since it's a single and optional field, I added it to the `WindowsExternalEditor` type and renamed it as you suggested to `jetBrainsToolboxScriptName`

app/src/lib/editors/win32.ts
tsvetilian-ty: The main issue is that JetBrains Toolbox generated the same script name for products without considering their edition.
JetBrains PyCharm Example:
1. Install JetBrains PyCharm Community
2. At this point, JetBrains Toolbox will generate a shell script called `pycharm`
3. Install JetBrains PyCharm Professional
4. JetBrains Toolbox will generate a new script with the same name, `pycharm` and will override the script generated for the community version

I added this example to [editor-integration.md](https://github.com/desktop/desktop/pull/15780/files#diff-2a91c8695284b34dee3845d11a7595422c8dd1a773c7b43e31cdd8a89655104c) and more documentation on how the auto-generation process works

docs/technical/editor-integration.md
```diff
@@ -206,20 +206,32 @@ missing or lost it won't let you launch the external editor.
...
 {
   name: 'JetBrains PyCharm',
   ...
-  toolboxShellScriptName: 'pycharm',
+  jetBrainsToolboxScriptName: 'pycharm',
 },
 ```
 
+**Note:** Don't use `jetBrainsToolboxScriptName` field on JetBrains community and edu
+editions. When JetBrains Toolbox generates the scripts. It doesn't consider the
```
sergiou87: Small nit:
```suggestion
**Note:** Don't use `jetBrainsToolboxScriptName` field on JetBrains community and edu
editions. When JetBrains Toolbox generates the scripts, it doesn't consider the
```

Also maybe this should be reworded not to explicitly mention the community and edu editions, but to explain that, in general, Desktop can't handle multiple editions of the same JetBrains IDE installed simultaneously, since JetBrains Toolbox will use the same script name and override it for existing versions. Is that correct? Or is it really specific to community and edu editions?

tsvetilian-ty: You're right. Instead of using community and edu editions, I should use a more generic statement, so I changed it. There is nothing specific to the community and edu, and I just used them as examples of product editions. 

Yes, you're correct. GitHub Desktop can't detect multiple editions because a new installation overrides the existing script.

sergiou87 (approved):
Thank you for those changes!! I tested it and works as expected 👏

PR #15794, opened 2022-12-16
Validate Electron version before beta/prod builds

sergiou87:
## Description
After being afraid of accidentally releasing the wrong Electron version a few months ago, we decided to set up some mechanism to ensure we explicitly state which Electron version is allowed/expected to be released for the production and beta channels.

This PR adds a new script for just this that is ran in our CI workflow for release builds in the prod and beta channels. This will help us keeping one Electron version in beta builds for testing, and then a stable/tested one in prod builds.

## Release notes

Notes: no-notes

### DIFF
--- a/script/validate-electron-version.ts
+++ b/script/validate-electron-version.ts
@@ -1,0 +1,53 @@
+/* eslint-disable no-sync */
+/// <reference path="./globals.d.ts" />
+
+import * as distInfo from './dist-info'
+
+type ChannelToValidate = 'production' | 'beta'
+
+/**
+ * This object states the valid/expected Electron versions for each publishable
+ * channel of GitHub Desktop.
+ *
+ * The purpose of this is to ensure that we don't accidentally publish a
+ * production/beta/test build with the wrong version of Electron, which could
+ * cause really bad regressions to our users, and also the inability to go back
+ * to a previous version of GitHub Desktop without losing all settings.
+ */
+const ValidElectronVersions: Record<ChannelToValidate, string> = {
+  production: '19.0.0',
+  beta: '19.0.0',
+}
+
+const channel = getChannelToValidate()
+
+if (channel === null) {
+  console.log(
+    `No need to validate the Electron version of a ${distInfo.getChannel()} build.`
+  )
+  process.exit(0)
+}
+
+const expectedVersion = ValidElectronVersions[channel]
+const pkg: Package = require('../package.json')
+const actualVersion = pkg.devDependencies?.electron
+
+if (actualVersion !== expectedVersion) {
+  console.error(
+    `The Electron version for the ${channel} channel is incorrect. Expected ${expectedVersion} but found ${actualVersion}.`
+  )
+  process.exit(1)
+}
+
+console.log(
+  `The Electron version for the ${channel} channel is correct: ${actualVersion}.`
+)
+
+function getChannelToValidate(): ChannelToValidate | null {
+  const channel = distInfo.getChannel()
+  return isChannelToValidate(channel) ? channel : null
+}
+
+function isChannelToValidate(channel: string): channel is ChannelToValidate {
+  return Object.keys(ValidElectronVersions).includes(channel)
+}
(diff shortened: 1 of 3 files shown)


script/validate-electron-version.ts
```diff
@@ -0,0 +1,53 @@
...
+if (channel === null) {
+  console.log(
+    `No need to validate the Electron version of a ${distInfo.getChannel()} build.`
+  )
+  process.exit(0)
+}
+
+const expectedVersion = ValidElectronVersions[channel]
+const pkg: Package = require('../package.json')
+const actualVersion = pkg.devDependencies?.electron
```
tidy-dev, line 33: Minor concern - If someone uses the  `=19.0.0` in the package.json as opposed to the `19.0.0` will this fail?  We could just update one or the other if that were to occur, but, we could use the `semver` package to do the comparison like in the `endpoint-capabilities` file to not hit that bump/have to remember which one we used.

script/validate-electron-version.ts
```diff
@@ -0,0 +1,53 @@
...
+ * production/beta/test build with the wrong version of Electron, which could
+ * cause really bad regressions to our users, and also the inability to go back
+ * to a previous version of GitHub Desktop without losing all settings.
+ */
+const ValidElectronVersions: Record<ChannelToValidate, string> = {
+  production: '19.0.0',
+  beta: '19.0.0',
+}
+
+const channel = getChannelToValidate()
```
tidy-dev, line 22: Other thoughts - But just minor nit (feel free to ignore), we could probably reuse the `distInfo.isPublishable` by doing something like below and not have to make our own channel checking stuff in this file. Plus added benefit of clearing only checking for publishable builds. 

```
const expectedVersion = ValidElectronVersions[distInfo.channel];
const isExpectationDefinedForChannel = expectedVersion !== undefined
if(!distInfo.isPublishable() || !isExpectationDefinedForChannel) {
  // No need to validate console log
  process.exit(0)
}
```

script/validate-electron-version.ts
sergiou87: Using `semver` sounds like a nice idea! I kind of wanted to enforce the exact version to use, so didn't even consider other options.

script/validate-electron-version.ts
sergiou87: I rewrote this part several times and didn't quite like the combination of `isPushable` with additional checks 🤔 If, in the end, I needed to check the whether or not the channel is in `ValidElectronVersions`… 🤷

script/validate-electron-version.ts
sergiou87: After looking into this, I think it makes sense to be explicit about what we want. According to [the documentation](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#dependencies), we shouldn't use `=version` when we want to require a specific version, and I'm a bit concerned about letting `semver` doing something "smarter" by mistake (like considering the same major version but accepting patch revisions, for example).

So I think comparing these hardcoded strings should be more than fine. Do you agree? Or do you feel strongly that we should rely on `semver` for this?

tidy-dev: >So I think comparing these hardcoded strings should be more than fine. Do you agree?

Yes. They are more than fine. :) It is just a convenience factor.

Reason I brought it up is because I have seen in our history some devs have used `=` and some not. The documentation you referenced says to refer to [this semver documentation](https://github.com/npm/node-semver#ranges) for more on semver notation. This says using the `=` is valid and I think a sort of the "default" syntax as the doc says to assume it is present if not provided. The semver library has a `semver.eq()` which would make sure that for a version number it is its equivalent as opposed to `satifies()`. I believe `v19.0.0`, `=19.0.0`, and `19.0.0` are all equal in terms of versioning.  (Tho I haven't seen v19.0.0 used before)

For reference:
image

sergiou87: As discussed privately, `eq` doesn't seem to work well with `=...` versions :(

image

So we decided to use `===` which is completely clear and unambiguous, and with that avoid any possible confusion.

tidy-dev (approved):
✨  Thank you for hashing out my questions. :)

Issue #15774, opened 2022-12-13
protocol ' https' is not supported when spaces in URL


joerg1985:
### The problem

Adding a leading space to the git URL will cause an error `fatal: protocol ' https' is not supported` while fetching the repository. If the settings dialog is called again the space is not shown, but the ./git/config file will still contain the wrong URL.

### Release version

3.1.2 (x64)

### Operating system

Windows 10 x64

### Steps to reproduce the behavior

Repository > Repository settings... > Primary remote respository (origin)
add a space e.g. change "https://my.url" to " https://my.url"

steveward:
Closing as this was fixed in #15821.

zemgeez:
thanks for that

PR #15821, opened 2022-12-25
Trim white spaces before saving repository URL changes

Shivareddy-Aluri:
Closes #[issue number]
https://github.com/desktop/desktop/issues/15774

## Description

triming the trailing spaces around the URL in the repository path ,to avoid errors while fetching the repo info

## Release notes

Notes: [Fixed] Trim leading and trailing whitespaces in URLs of repository remotes

### DIFF
--- a/app/src/ui/repository-settings/repository-settings.tsx
+++ b/app/src/ui/repository-settings/repository-settings.tsx
@@ -289,20 +289,19 @@ export class RepositorySettings extends React.Component<
     const modifiedUrl = this.state.remote?.url.trim()
 
     if (this.state.remote && this.props.remote) {
-      if (modifiedUrl !== this.props.remote.url) {
-        try {
-          await this.props.dispatcher.setRemoteURL(
-            this.props.repository,
-            this.props.remote.name,
-            this.state.remote.url
-          )
-        } catch (e) {
-          log.error(
-            `RepositorySettings: unable to set remote URL at ${this.props.repository.path}`,
-            e
-          )
-          errors.push(`Failed setting the remote URL: ${e}`)
-        }
+      const trimmedUrl = this.state.remote.url.trim()
+      try {
+        await this.props.dispatcher.setRemoteURL(
+          this.props.repository,
+          this.props.remote.name,
+          trimmedUrl
+        )
+      } catch (e) {
+        log.error(
+          `RepositorySettings: unable to set remote URL at ${this.props.repository.path}`,
+          e
+        )
+        errors.push(`Failed setting the remote URL: ${e}`)
       }
     }
 
--- a/app/src/ui/repository-settings/repository-settings.tsx
+++ b/app/src/ui/repository-settings/repository-settings.tsx
@@ -286,7 +286,6 @@ export class RepositorySettings extends React.Component<
   private onSubmit = async () => {
     this.setState({ disabled: true, errors: undefined })
     const errors = new Array<JSX.Element | string>()
-    const modifiedUrl = this.state.remote?.url.trim()
 
     if (this.state.remote && this.props.remote) {
       const trimmedUrl = this.state.remote.url.trim()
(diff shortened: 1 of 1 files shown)


sergiou87:
Even if the change is this simple, would you mind filling out the template with the PR description, release notes and so on? 🙏

Shivareddy-Aluri:
> Even if the change is this simple, would you mind filling out the template with the PR description, release notes and so on? 🙏

@sergiou87 done

sergiou87 (changes requested):
After testing it, it feels a bit off because if you try to introduce a space _anywhere_ in an existing URL, it moves the caret to the end of the text.

I see two options here:
1. Trim the url at the time `this.props.dispatcher.setRemoteURL` is invoked (in the `onSubmit` method)
2. Use `error.push` as it's done in the `onCommitterNameChanged` method when the URL is not valid.

I think (1) should be enough for this: bothering the user just because they introduced extra whitespaces at the beginning or end of the URL sounds too much.

Happy to hear other thoughts, though!

Shivareddy-Aluri:
> After testing it, it feels a bit off because if you try to introduce a space _anywhere_ in an existing URL, it moves the caret to the end of the text.
> 
> I see two options here:
> 
> 1. Trim the url at the time `this.props.dispatcher.setRemoteURL` is invoked (in the `onSubmit` method)
> 2. Use `error.push` as it's done in the `onCommitterNameChanged` method when the URL is not valid.
> 
> I think (1) should be enough for this: bothering the user just because they introduced extra whitespaces at the beginning or end of the URL sounds too much.
> 
> Happy to hear other thoughts, though!

@sergiou87 yes make sense to go with 1 for this..
made the requested change.

app/src/ui/repository-settings/repository-settings.tsx
```diff
@@ -293,7 +293,7 @@ export class RepositorySettings extends React.Component<
           await this.props.dispatcher.setRemoteURL(
             this.props.repository,
             this.props.remote.name,
-            this.state.remote.url
+            this.state.remote.url.trim()
```
sergiou87: Works as expected, but could you extract this trimmed url to a constant, and then compare it with `this.props.remote.url` in the `if` above (line 291)?

sergiou87 (changes requested):

app/src/ui/repository-settings/repository-settings.tsx
Shivareddy-Aluri: done!

app/src/ui/repository-settings/repository-settings.tsx
```diff
@@ -286,9 +286,10 @@ export class RepositorySettings extends React.Component<
   private onSubmit = async () => {
     this.setState({ disabled: true, errors: undefined })
     const errors = new Array<JSX.Element | string>()
+    const modifiedUrl = this.state.remote?.url.trim()
```
sergiou87: This should be inside of the first `if`, so you don't need the `?`. Then, of course, you should use `modifiedUrl` (I'd name it `trimmedURL`) as parameter to `setRemoteURL` instead of `this.state.remote.url`.

sergiou87 (changes requested):

app/src/ui/repository-settings/repository-settings.tsx
Shivareddy-Aluri: done

app/src/ui/repository-settings/repository-settings.tsx
```diff
@@ -288,20 +288,19 @@ export class RepositorySettings extends React.Component<
...
-            this.state.remote.url
-          )
-        } catch (e) {
-          log.error(
-            `RepositorySettings: unable to set remote URL at ${this.props.repository.path}`,
-            e
-          )
-          errors.push(`Failed setting the remote URL: ${e}`)
-        }
+      const trimmedUrl = this.state.remote.url.trim()
```
sergiou87, line 291: Almost there! We still want the second `if`:
```suggestion
      const trimmedUrl = this.state.remote.url.trim()
      if (trimmedUrl !== this.props.remote.url) {
```

sergiou87 (changes requested):

sergiou87 (approved):
Code looks good and works as expected!! Thank you for your changes and patience 🙇‍♂️

PR #15828, opened 2022-12-27
Bump Electron to v21.3.3

tidy-dev:
## Description

This PR upgrades Electron to v21.3.3! 

A couple of pretty simple type changes were required:
1. `PathType` type used by `Electron.App.getPath(pathType)` no longer has a type of `'cache'` and has added a type of `'sessionData'`. This doesn't affect as we only use the PathTypes of `'document'` and `'home'`.
2. Electron repurposed the type of `Electron.Response` and replaced it with `Electron.CallbackResponse`. This does not have a programmatic impact on us as `Electron.CallbackResponse` has the same properties as `Electron.Response` used to have.

~~Node Notes: Node v16.14.0 is latest with arm64 builds and thus that is what I bumped it to; however, Electron 21.3.3 uses node v16.16.0 and so I have requested it in the [arm64 build thread](https://github.com/nodejs/build/issues/2540#issuecomment-1365940806). When we get that build, we will want to bump accordingly.~~ It was generated for us https://github.com/nodejs/build/issues/2540#issuecomment-1047935408 ❤️ 

## Release notes
Notes: no-notes


app/src/main-process/ordered-webrequest.ts
```diff
@@ -140,7 +140,7 @@ export class OrderedWebRequest {
     this.onBeforeRequest = new AsyncListenerSet(
       webRequest.onBeforeRequest.bind(webRequest),
       async (listeners, details) => {
-        let response: Response = {}
+        let response: CallbackResponse = {}
```
tidy-dev, line 143: Non-impacting change: 

Here is new interface:
image


If you look down at line 151, `cancel` and `redirectURL` are the expected properties.

image

app/src/ui/lib/app-proxy.ts
```diff
@@ -21,6 +20,7 @@ export type PathType =
   | 'recent'
   | 'logs'
   | 'crashDumps'
+  | 'sessionData'
```
tidy-dev, line 23: Non-impacting change: This `PathType ` is consumed in `ipc-shared.ts` by the `'get-path'` response channel definition. This is used in `main-process-proxy.ts` by `getPath` method. The `getPath` method is used in only used in `app-proxy.ts` and only uses the types of `'home'` and `'documents'`. 

image

sergiou87 (commented):
So good to see there are no major API changes that affect us 😌 

Thank you for your comments clarifying the changes! 💖

tidy-dev:
For testing - Release 3.1.3-test4 has the latest on it.

tidy-dev:
Planned to go ahead and move to next version since new Electron will be out soon.

PR #15831, opened 2022-12-27
Bump Electron to v22.0.3

tidy-dev:
## Description
We discussed moving to v21.3.3 to stay 1 version behind. I have that [PR open](https://github.com/desktop/desktop/pull/15828) and am testing it out (~~release 3.1.3-test1 is test build with it~~ release 3.1.3-test4 has newer node).

However... I think we will have some release churning with the start of the new year and since the Electron release cadence appears to be every 2 months.. A new version could be as soon as the end of January/first of February. Assuming we are going to keep Electron on beta for a few weeks. Moving to 22 for beta could likely land us 1 version behind by the time we release to prod.

Thus, I just spun up a this branch to see if moving to v22.0.0 brought any trouble and it appears pretty straight forward so far. 

This PR upgrades Electron to v22.0.3! 

A couple of pretty simple type changes were required:
1. `PathType` type used by `Electron.App.getPath(pathType)` no longer has a type of `'cache'` and has added a type of `'sessionData'`. This doesn't affect as we only use the PathTypes of `'document'` and `'home'`.
2. Electron repurposed the type of `Electron.Response` and replaced it with `Electron.CallbackResponse`. This does not have a programmatic impact on us as `Electron.CallbackResponse` has the same properties as `Electron.Response` used to have.
3. Using the `app. on('new-window'...` event has been [deprecated since 13 and removed in 22](https://github.com/electron/electron/pull/34526). Added the recommended fix. I am not 100% it works as expected as I couldn't replicate log from old version.. :/ 

~~Node Notes: Node v16.14.0 is latest with arm64 builds and thus that is what I bumped it to; however, Electron 22 uses node v16.17.1 and so I have requested it in the [arm64 build thread](https://github.com/nodejs/build/issues/2540#issuecomment-1365940806). When we get that build, we will want to bump accordingly.~~ It was generated for us https://github.com/nodejs/build/issues/2540#issuecomment-1047935408 ❤️


app/.npmrc
```diff
@@ -1,3 +1,3 @@
 runtime = electron
 disturl = https://electronjs.org/headers
-target = 19.0.0
+target = 21.3.3
```
sergiou87: I understand this is WIP, but just in case…
```suggestion
target = 22.0.0
```

package.json
```diff
@@ -155,7 +155,7 @@
     "@types/webpack-hot-middleware": "^2.25.6",
     "@types/webpack-merge": "^5.0.0",
     "@types/xml2js": "^0.4.0",
-    "electron": "19.0.0",
+    "electron": "20.0.0",
```
sergiou87: Oh and this should be 22.0.0, so maybe this is more WIP than I thought it was? 😆 
```suggestion
    "electron": "22.0.0",
```

tidy-dev: Woops! - yeah I kind of put this one together is a hurry at end of my day after doing 21.3.3. (But also didn't know if we wanted this yet - was going to check in with team next time we have everyone)

tidy-dev:
Since opening this, there was have been a few patches to 22. Thus, I have bumped to 22.0.3

app/src/main-process/ordered-webrequest.ts
```diff
@@ -140,7 +140,7 @@ export class OrderedWebRequest {
     this.onBeforeRequest = new AsyncListenerSet(
       webRequest.onBeforeRequest.bind(webRequest),
       async (listeners, details) => {
-        let response: Response = {}
+        let response: CallbackResponse = {}
```
tidy-dev, line 143: Same update to api as explained in https://github.com/desktop/desktop/pull/15828

tidy-dev:
Tested locally on my macOS - Test deployment pending and will test out on Windows.

sergiou87 (approved):
Code looks good and I tested it on macOS and Windows and worked as expected!

I tested not only the most basic use cases, but also those which involve doing some "system stuff" (opening the browser, handling our custom url scheme, notifications, moving repos to the recycle bin, "open file" dialog…)

I also opened the latest beta right after this test build and it kept my list of repositories and settings, so it seems to be backwards compatible in that regard 😌

Issue #5411, opened 2018-08-15
improve detection of programs that can open/edit files in Desktop


shiftkey:
Extracted from #4753

**Please describe the problem you think should be solved**
GitHub Desktop relies on the Electron APIs `shell.openInExternal` for opening a file in a chosen program. This covers a lot of cases, but as I outlined [here](https://github.com/desktop/desktop/issues/4753#issuecomment-392394769) some programs will use the `edit` verb when registering themselves. Electron is hard-coded to look for the `open` at the verb on Windows, and I'm not aware of whether there's a similar convention in macOS.

The other thing that's unclear is the context menu entry for this: `Open with Default Program`

<img width="392"  src="https://user-images.githubusercontent.com/359239/44158352-d7b9a900-a08a-11e8-844e-7651d7cc0f88.png">

Examples:

 - it's not clear which program will open this file extension if I click on it
 - if no program exists, should we disable the button?

**[Optional] Do you have any potential solutions in mind?**

Two ideas:

 - add APIs to Electron to support locating the program associated with a given extension (and optional verb depending on the OS) as well as passing a different verb to `openInExternal`
 - use a different method to find the program associated with a given extension

These require experience with how each OS that Desktops supports should work, and careful implementation to not introduce any security problems.

tom-n-pdx:
On both MacOS & Windows my editor is not deteted. And there is no method to provide a path to the editor I'd like to use.

On Windows I'm using NotePad++ installed with standard configuration - it's not detected.

Om MacOS I'm using Aquamacs installed with standard configuration - it's not detected.

Please improve your editor detection - or more reliable - just let provide a path to a editor.

shiftkey:
@tom-n-pdx this issue is related to opening specific files and the OS integration there, but it seems like you're talking about the **Repository** | **Open in [Editor]** feature? If that's the case, here's some extra context:

> On Windows I'm using NotePad++ installed with standard configuration - it's not detected.

We tried to add support for that in #4120 but we are unable to pass it a folder to open so that support is on hold (see https://github.com/desktop/desktop/pull/4120#issuecomment-369901927).

> Om MacOS I'm using Aquamacs installed with standard configuration - it's not detected.

I'm not familiar with this editor. Please check out the [macOS editor integration](https://github.com/desktop/desktop/blob/master/docs/technical/editor-integration.md#macos) to see what's involved with adding support.

> Please improve your editor detection - or more reliable - just let provide a path to a editor.

[...]

Issue #4785, opened 2018-05-24
Add support for launching Emacs.app as "Open in External Editor" entry


osculabo-carnifex:
**Is your feature request related to a problem? Please describe.**
Emacs.app is not picked up as an external editor

**Describe the solution you'd like**
Set arbitrary app as editor and pass it files as arguments

**Describe alternatives you've considered**
Painstakingly list every editor in existence or gently remind people to switch to Atom

**Teachability, Documentation, Adoption, Migration Strategy**
Use the current heuristic to list known editors, have a button to select the application/path to application outright below it.

MagicMarvMan:
Hello. I would like to take this issue. :+1: 
I will open a PR when I'm done.

@shiftkey 

Thank you,
Kind regards, 
MagicMarvMan

MagicMarvMan:
Hmm... I am trying to solve it, but it isn't detecting Emacs as an editor. An additionally the editor style looks really weird:
![Screenshot](https://user-images.githubusercontent.com/24806960/41856818-4e8fbafa-7896-11e8-89a1-baa75826e7e8.png)
Do you have an idea, @shiftkey ?
Thank you.

iAmWillShepherd:
@MagicMarvMan, thanks for picking this issue up. Would you mind @-mentioning me or `@desktop/eng` when you create your PR?

Rexogamer:
It’s been a year; I’ll take this up

billygriffin:
Since this has sat for so long without any additional demand or interest in working on it, I'm going to close this. We're definitely open to supporting additional external editors but for anything outside the largest ones, we rely on the community to build those integrations. If you're interested in doing this work and know how to implement it, please let us know and we can reopen. Thanks for the issue!

lcolen:
I'm setting up an environment for developing code on Linux (Ubuntu 20.04), am using emacs as my editor and currently using Eclipse as my IDE.  github desktop only gives me choices of Atom (a discontinued product), Visual Studio Code and Gnome text editor.  
Emacs has been one of the two "standard" *nix editors since not only before github, but probably since most of the people working on github desktop were born.  The basic emacs editing commands are standard in the shells that come with both Linux and Mac.  
It really needs to be added as a default editor.

fchenunsw:
If this happens, it'll be great!

PR #15857, opened 2023-01-03
Update linux.ts -- add editor Emacs

zipperer:
In `editors: ILinuxExternalEditor[]`, add entry for [Emacs](https://www.gnu.org/software/emacs/)

Relates to #4785 and #5411.

## Description

This commit follows the instructions at 
https://github.com/desktop/desktop/blob/development/docs/technical/editor-integration.md#linux
to add an editor for Linux. 

A forthcoming commit will add Emacs to the list of editors on that page. 

For editor Emacs I added these paths
 - /snap/bin/emacs (the path when a user installs GNU Emacs via snap on Ubuntu, i.e. when a user uses Ubuntu Software and installs  GNU Emacs. See also https://snapcraft.io/emacs)
 -  /usr/local/bin/emacs (the path when a user installs GNU Emacs according to [documentation for the project](https://www.gnu.org/software/emacs/manual/html_node/efaq/Installing-Emacs.html))
 - /usr/bin/emacs (a path I found when I searched the web for 'default emacs install location' -- see e.g. https://www.emacswiki.org/emacs/EmacsAsDaemon#h5o-2)

### DIFF
--- a/app/src/lib/editors/linux.ts
+++ b/app/src/lib/editors/linux.ts
@@ -70,6 +70,10 @@ const editors: ILinuxExternalEditor[] = [
     name: 'Jetbrains WebStorm',
     paths: ['/snap/bin/webstorm'],
   },
+  {
+    name: 'Emacs',
+    paths: ['/snap/bin/emacs', '/usr/local/bin/emacs', '/usr/bin/emacs'],
+  },
 ]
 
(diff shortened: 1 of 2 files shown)


sergiou87 (approved):
Thank you for your contribution!

PR #15864, opened 2023-01-04
Align no branches text in the center

tidy-dev:
## Description
Align no branches text to center.

Note this could impact main branches dropdown and anywhere else we use the branches list. But I checked and I think text center is what we want in all cases.

### Screenshots

image

## Release notes
Notes: no-notes


sergiou87 (approved):
Code looks good, tested it and works as expected! :shipit: ❗

Issue #15842, opened 2022-12-30
Clone Directory Messes Up When Changing URL


Look-Its-Sky:
### The problem

When adding 

### Release version

3.1.2

### Operating system

MacOS Ventura 13.1

### Steps to reproduce the behavior

1.) Clone from a URL
2.) Paste a URL and append two '/'s to the end of the URL
ex: 
Before - https://github.com/citra-emu/citra.git
After - https://github.com/citra-emu/citra.git//
3.) The clone directory should go back a folder each time a / is removed or added

### Screenshots

https://user-images.githubusercontent.com/67021158/210036453-a2881160-1e75-4c91-86ad-0aad6805de95.mp4

### Additional context

Also sorry if I did something wrong this is my first issue.

Shivareddy-Aluri:
@tidy-dev @sergiou87 what should be done here?
we are using `dirname` in app/src/ui/clone-repository/clone-repository.tsx -> updateUrl .. 
in Unix when we give `dirname /Users/xyz/Documents/GitHub/desktop//` it will give the output directory as `/Users/xyz/Documents/GitHub` .
this shouldn't happen and we have to avoid it ?

tidy-dev:
Without any investigation into current code, I would assume that we will want to sanitize the input from the repository URL before making modifications the the local path output. But, until this can be done, it is simple enough to tab down to the local path and modify it to whatever is your preferred target directory.

Shivareddy-Aluri:
so should i fix it? @tidy-dev

PR #15884, opened 2023-01-07
Avoid folder messup on url change while cloning repository via URL

Shivareddy-Aluri:
Closes #[issue number]
https://github.com/desktop/desktop/issues/15842

## Description

we are using dirname in app/src/ui/clone-repository/clone-repository.tsx -> updateUrl.
in Unix when we give dirname` /Users/xyz/Documents/GitHub/desktop//` it will give the output directory as `/Users/xyz/Documents/GitHub `. so avoiding with this code fix

-

## Release notes
no-notes

Notes:

### DIFF
--- a/app/src/ui/clone-repository/clone-repository.tsx
+++ b/app/src/ui/clone-repository/clone-repository.tsx
@@ -619,7 +619,11 @@ export class CloneRepository extends React.Component<
     const dirPath = tabState.path
     if (lastParsedIdentifier) {
       if (parsed) {
-        newPath = Path.join(Path.dirname(dirPath), parsed.name)
+        if (parsed?.name !== '/') {
+          newPath = Path.join(Path.dirname(dirPath), parsed.name)
+        } else {
+          newPath = dirPath
+        }
       } else {
         newPath = Path.dirname(dirPath)
       }


101123-007:
[EPContactsResults_1114011339312491.xlsx](https://github.com/desktop/desktop/files/10378769/EPContactsResults_1114011339312491.xlsx)

niik:
@Shivareddy-Aluri thanks for wanting to contribute, we really appreciate it! I apologize for the delay in reviewing your PR.

While this PR fixes the precise problem described in that issue where adding an extra trailing slash causes the suggested path to change unexpectedly you would run into the same problem again if you added another slash. Instead what we should be doing here is to change the parsing logic in `remote-parsing.ts` such that it stops capturing after encountering a valid name and owner pattern (neither repository names or user handles may contain slashes). That said the logic in there is heavily relied upon and hasn't been touched in a long time so I'm not convinced that fixing this issue is worth the risk.

I took a quick look at it before abandoning the work for the reasons outlined above but I think the below change could be a decent starting point for addressing this issue.

```diff
diff --git a/app/src/lib/remote-parsing.ts b/app/src/lib/remote-parsing.ts
index 737097d37..83707da8e 100644
--- a/app/src/lib/remote-parsing.ts
+++ b/app/src/lib/remote-parsing.ts
@@ -45,6 +45,21 @@ const remoteRegexes: ReadonlyArray<{ protocol: GitProtocol; regex: RegExp }> = [
 
 /** Parse the remote information from URL. */
 export function parseRemote(url: string): IGitRemoteURL | null {
+  try {
+    const { protocol, hostname, pathname } = new URL(url)
+    const nwo = /^\/([^\/]+)\/([^\/]+)/.exec(pathname)
+
+    const protoMap = new Map<string, GitProtocol>([
+      ['https:', 'https'],
+      ['ssh:', 'ssh'],
+    ])
+    const proto = protoMap.get(protocol)
+    if (nwo && proto) {
+      const name = nwo[2].replace(/\.git$/, '')
+      return { protocol: proto, hostname, owner: nwo[1], name }
+    }
+  } catch {}
+
   for (const { protocol, regex } of remoteRegexes) {
     const match = regex.exec(url)
     if (match !== null && match.length >= 4) {
```

In the meantime since this PR doesn't accomplish what we'd like for it to accomplish I'm going to close it out. If you feel like picking this up again feel free to open a new PR!

PR #15907, opened 2023-01-11
Pull-push-fetch button + dropdown

sergiou87:
## Description

This PR can be considered a follow up after our work in #15494 : instead of just having a menu from where users can fetch and force-push at any time those actions are available, we have considered making those actions more easily accessible via a new dropdown that can be expanded right below the push-pull-fetch button.

This new dropdown has been built by reusing the existing `Dropdown` component, which already implements three features we needed now:
- Floating element
- Dark translucent backdrop
- Focus trap

For now, this PR adds the following “dropdown actions”:

- For `Publish Branch`: Fetch
- For `Pull`: Fetch and Force Push (only if the current branch has diverged)
- For `Push`: Fetch
- For `Force push`: Fetch

Any other button state has no dropdown actions (and therefore no dropdown at all, they’re just buttons).

In the (near or not) future, we could consider other options like "Pull with rebase" or "Fast-forward" the current branch.

### Screenshots

https://user-images.githubusercontent.com/1083228/211846562-55644524-53d9-49cf-a6f2-c4e7bef81c58.mov

## Known issues

Fixed the focus issue by tweaking the style and making it behave as one would expect 🤞 

https://user-images.githubusercontent.com/1083228/211879621-a8c10d56-8119-4e88-a1fb-ad97d63b46fa.mov

~~Due to the focus trap, the app focuses the first item when the dropdown is shown, and it looks highlighted (and without the "focus ring"). I have to investigate further.~~

image

## Release notes

Notes: [Added] Add fetch and force-push actions in a dropdown as an alternative to the main Pull/Push/Publish action button

### DIFF
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -249,7 +249,8 @@ export class PushPullButton extends React.Component<IPushPullButtonProps> {
     const focusedIndex = Array.from(items).indexOf(focusedItem)
     const nextIndex =
       event.key === 'ArrowDown' ? focusedIndex + 1 : focusedIndex - 1
-    const nextItem = items[nextIndex % items.length]
+    // http://javascript.about.com/od/problemsolving/a/modulobug.htm
+    const nextItem = items[(nextIndex + items.length) % items.length]
     nextItem?.focus()
   }
 
--- a/app/src/ui/toolbar/push-pull-button.tsx
+++ b/app/src/ui/toolbar/push-pull-button.tsx
@@ -19,11 +19,11 @@ import {
   ToolbarDropdown,
   ToolbarDropdownStyle,
 } from './dropdown'
-import { Button } from '../lib/button'
 import { FoldoutType } from '../../lib/app-state'
 import { ForcePushBranchState } from '../../lib/rebase'
+import { PushPullButtonDropDown } from './push-pull-button-dropdown'
 
-const DropdownItemClassName = 'push-pull-dropdown-item'
+export const DropdownItemClassName = 'push-pull-dropdown-item'
 
 interface IPushPullButtonProps {
   /**
@@ -86,12 +86,12 @@ interface IPushPullButtonProps {
   readonly onDropdownStateChanged: (state: DropdownState) => void
 }
 
-enum DropdownItemType {
+export enum DropdownItemType {
   Fetch = 'fetch',
   ForcePush = 'force-push',
 }
 
-type DropdownItem = {
+export type DropdownItem = {
   readonly title: string
   readonly description: string | JSX.Element
   readonly action: () => void
@@ -142,7 +142,7 @@ function renderLastFetched(lastFetched: Date | null): JSX.Element | string {
  * This represents the "double arrow" icon used to show a force-push, and is a
  * less complicated icon than the generated Octicon from the `octicons` package.
  */
-const forcePushIcon: OcticonSymbol.OcticonSymbolType = {
+export const forcePushIcon: OcticonSymbol.OcticonSymbolType = {
   w: 10,
   h: 16,
   d:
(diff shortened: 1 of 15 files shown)


app/src/ui/toolbar/push-pull-button.tsx
```diff
@@ -273,51 +153,162 @@ const forcePushIcon: OcticonSymbol.OcticonSymbolType = {
...
+  ) {
+    return () => {
+      return (
+        <>
+          {/*
+          This <div> is not interactive, but its children are, so we need to
+          disable the jsx-a11y/no-static-element-interactions rule.
+          */}
+          {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
+          <div
```
tidy-dev: I had a similar problem with the items of the `DropdownSelectButton` needing to be up and down arrow navigable. Did a bit of looking around at other implementations on the web to make it accessible. Came to the conclusion that using Buttons instead of divs is more accessible. You get the up/down navigation for free and no disabling linters.  The negative about it is that these do not feel like buttons -> They feel like list items and will need styled to look like list items in this approach. Thus, it feels like it breaks the html convention of using the elements for what they present as. But, I think it is arguable that these are actually buttons because you click on them and expect an action just like a button. Thus we should be using an interact-able html element -> a button. 

However, I see you are doing some stuff with focusing and I am not sure this will solve all that, but I think that the onKeyDown will no longer complain to the linter for a button element if the focus logic is still needed.

tidy-dev: Hmm.. I think I may have misunderstood the code... cause the individual items are buttons.. disregard above. Still think we should try to find a way not to introduce more jsx-ally linters, but till I understand why it's needed I am not sure how to remove it. :D

sergiou87: Exactly, these divs are wrappers with a button inside of each of them. I can try to bubble those buttons up and see if it doesn't break the style 😳

tidy-dev: Alright.. so I did have this problem before.. :D I just was looking at the wrong part. Check out `avoiding-disabling-the-linter` to see how I went about it another way to not have to introduce a linter disabling.. (not sure if avoiding the linter is the best reason to be doing this.. but it also separates the concern out a little).

app/src/ui/toolbar/push-pull-button.tsx
```diff
@@ -273,51 +153,162 @@ const forcePushIcon: OcticonSymbol.OcticonSymbolType = {
...
+      case DropdownItemType.ForcePush:
+        return {
+          title: `Force push ${remoteName}`,
+          description: (
+            <>
+              Overwrite any changes on {remoteName} with your local changes
+              <br />
+              <br />
+              <div className="warning">
+                <span className="warning-title">Warning:</span> A force push
```
tidy-dev: Wondering if it makes sense to remove this when the user is going to get the confirmation popup anyway? I could see it feeling annoying to get the messaged twice. I could even see some users wanting to dismiss this warning too.. but I agree that it is a very dangerous operation and therefore warrants some indication in the UI that it should not be done lightly. 

Other thoughts are I think we are starting to have multiple different UI views for "Warning" messages. Should we should try to be consistent?

image

image

image

tidy-dev (changes requested):
💖 💖 💖  This is looking awesome! I can't wait for it to ship! 

I have commented on a couple of things. The linter comment, I am just keen on not introducing new disables if possible because we eventually want to remove them down the road. 

Other thoughts, should this be feature flagged?

app/src/ui/toolbar/push-pull-button.tsx
sergiou87: I love it, thank you!! ❤️ 

I don't know if the `window.addEventListener` is better or worse than `onKeyDown`, but definitely helps to avoid the lint issue 😄

app/src/ui/toolbar/push-pull-button.tsx
sergiou87: Given the user might have disabled the confirmation popup, and this warning doesn't add more steps, I don't feel strongly about removing the warning entirely. But I do agree we should look for consistency in those warnings 😅 I will use one of those styles (yellow octicon + smaller font with default font color) and see if it looks nice.

sergiou87: After a few attempts, I'm unable to find a combination for the warning hint that looks well in that button, other than what was mocked and implemented 😕 

So if you don't mind, I'm gonna leave it as it is for now, feature flag this, and we can discuss it at some other point (probably along with the rest of examples you presented in your comment 🙇‍♂️ )

sergiou87:
@tidy-dev thanks for your feedback! I included your changes to avoid the linter issue without disabling it, and also feature-flagged the dropdown. I couldn't think of a cleaner way of doing that, keeping two versions of `PushPullButton` or doing crazy things with types. As always, I'm open to suggestions!

app/src/ui/toolbar/push-pull-button.tsx
tidy-dev: > Given the user might have disabled the confirmation popup, and this warning doesn't add more steps, I don't feel strongly about removing the warning entirely

I don't feel super strongly on this. But to clarify cause I can see my wording wasn't great,  I did not mean to remove it entirely. Only remove it when the user is also going to get the confirmation popup. So that is not like, I read this warning, ok I am going to do it anyway, and then I get asked again. :P

tidy-dev (approved):
✨  Thank you for those changes. 

(As for my latest comment on the warning message, feel free to ignore or do in a different PR if you think it make sense)

app/src/ui/toolbar/push-pull-button.tsx
sergiou87: Ohhh ok, my bad I didn't get that 😳 That sounds very reasonable to me, I can tackle that in a different PR. Thanks!

Atkinson2646:
Patch

Atkinson2646 (commented):
Patch

Taingkeatek:
Help me earn mone

AbsolytnaiaLove (commented):
#15907

gazaliawol421:
Hi my name is gazali Awol

Natchaphonwongplianphan:
Please arrange to verify the right of ownership of the data I act on the ownership of the right I claim under international law You guys let me struggle myself why no one help me some loyalty check it  Want or pass for more than ten years that I tried to maintain the ecosystem that the world of the world together violates the rights of the license card to the way that I am a Thai life.  occupy without humans

PR #15939, opened 2023-01-15
Prefix local repositories sharing same name with path

### LINKED ISSUE #15937: Local Repositories with same name should show more of the path
### The feature request
Fork-based workflows can result in multiple local repositories sharing the same directory name, which can make them hard to differentiate. While it is possible see the full path of each repository by hovering them, a user may need to do this multiple times, which is an improvable user experience.

Desktop supports `Aliases` which can be used to assign a unique label to a repository in those cases, however this has to be configured per-repository.

### Proposed solution
A possible solution could be to show more of the path where the repository is located, up to the point where the paths are unique.
This approach would be similar to how many IDEs (e.g. JetBrains) handle multiple open tabs with the same file name. 

### Additional context
The following screenshots show the current behavior as well as the above proposed solution by example.

#### Current behavior
current behavior

#### Proposed behavior
proposed behavior

ghost:
Closes #15937

## Description

To implement a proof-of-concept for the above mentioned issue, I expanded the existing feature of prefixing repositories that need to be distinguished.

One should check if the implementation still works correctly with Enterprise repositories, for which the mechanism was initially intended for. However, I do not have access to any of those. (I believe so, but better be safe than sorry 😅)

### Screenshots

new

The screenshot shows an example of four repositories all sharing the same directory name `Core`. The repositories are located at:
* `/Users/[user]/Desktop/puppy/Core`
* `/Users/[user]/Desktop/penguin/Core`
* `/Users/[user]/Desktop/test/penguin/dir/Core/Core`
* `/Users/[user]/Desktop/test/puppy/dir/Core/Core`

## Release notes

Notes:

### DIFF
--- a/app/src/ui/repositories-list/group-repositories.ts
+++ b/app/src/ui/repositories-list/group-repositories.ts
@@ -8,6 +8,7 @@ import { getDotComAPIEndpoint } from '../../lib/api'
 import { caseInsensitiveCompare } from '../../lib/compare'
 import { IFilterListGroup, IFilterListItem } from '../lib/filter-list'
 import { IAheadBehind } from '../../models/branch'
+import path from 'path'
 
 /**
  * Special, reserved repository group names
@@ -30,6 +31,7 @@ export interface IRepositoryListItem extends IFilterListItem {
   readonly id: string
   readonly repository: Repositoryish
   readonly needsDisambiguation: boolean
+  readonly minimalUniquePrefix: string | null
   readonly aheadBehind: IAheadBehind | null
   readonly changedFilesCount: number
 }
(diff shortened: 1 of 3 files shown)


andrewhong04:
this is hype

sergiou87:
Thanks for your contribution @thsmdt ! 💖 

The problem with this approach is we already use it in the `Recents` section to differentiate between repos with the same name but different owner (user/org). If we used the same thing to differentiate paths, users wouldn't know when the added prefix is the owner or parts of the path. We're aware of the original problem you reported, and using the tooltips is the best idea we came up with in the existing context 😞 

In any case, we'd rather see UX discussions/proposals in the original issue before a PR is submitted, to make sure we don't waste contributors' time implementing something that hasn't been agreed with the team 😕 

Given this one in particular requires more discussion, I'm gonna close this PR and invite you to post some ideas in https://github.com/desktop/desktop/issues/15937 about how to tackle this problem. Sorry!!

Issue #15974, opened 2023-01-20
Unable to diff a file with specific diff function


esseivan:
### The problem

I am unable to diff a zip file (or any with diff=... in .gitattributes).
I set the command in the config (.git/config) file with 
```
[diff "zip"]
	textconv = unzip -v
```
Then in the .gitattributes located in the root of my repository I specified :
`*.zip      diff=zip`

When checking for a diff in Github Desktop, it doesn't load anything and prints an error in the developer tools window
Nothing interesting in the log file.

I tried with other stuff, as `diff=astextplain` but the same thing happens
Also to note that it works with the command line git commands

### Release version

Version 3.1.4 (x64)

### Operating system

Windows 11 x64

### Steps to reproduce the behavior

1. Clean repo with zip file
2. .gitattributes with :
`*.zip      diff=zip`
4. ./.git/config contains the diff :
```
[diff "zip"]
	textconv = unzip -v
```

5. Edit zip file
6. Check diff in Github Desktop for the zip file
7. No diff loaded, infinite loading. Developer tools shows the following error

### Screenshots

![image](https://user-images.githubusercontent.com/14168019/213652751-94715ca1-7012-442c-85ab-75ad22cd96b3.png)

esseivan:
After some debugging I found that the program exits at the line 135 of /app/src/lib/git/spawn.ts.
But I do not get any descriptive message in the console, however, the stderr contained the information I needed : 
`unzip -v: line 1: unzip: command not found\nfatal: unable to read files to diff\n`
So my issue just was that the command I set for the textconv was invalid (even though it works for the cmd-line git). To fix it, I added `C:\Program Files\Git\usr\bin` to the PATH variable.

It could maybe be possible to also print the stderr instead of just printing the default error message (spawn.ts:135) with something like `log.error(stderr.toString());`

tidy-dev:
Thanks for doing that debugging. I am going to close this out since it sounds like you found a resolution. :)

PR #16015, opened 2023-01-26
Add more information about the error in spawnAndComplete

esseivan:
When a diff fail because the command set in the .gitconfig file is invalid, it doesn't shows the stderr output. Just a standard output message with the exit code. This is only usefull for the devs but not the users.

According to what I discussed in #15974

## Description
When a diff fail because the command set in the .gitconfig (or config) file is invalid, it doesn't shows the "stderr" output from the process that ran the command. Just a standard output message with the exit code is shown in the console. This is only usefull for the devs but not the users. This is why I think adding this log might be useful.

### Screenshots

N/A

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/git/spawn.ts
+++ b/app/src/lib/git/spawn.ts
@@ -132,7 +132,7 @@ export async function spawnAndComplete(
         })
         return
       } else {
-        log.error(stderr.toString())
+        log.error(`Error in spawnAndComplete, stderr:\n${stderr}`)
         reject(
           new Error(
             `Git returned an unexpected exit code '${code}' which should be handled by the caller (${name}).'`
--- a/app/src/lib/git/spawn.ts
+++ b/app/src/lib/git/spawn.ts
@@ -132,6 +132,7 @@ export async function spawnAndComplete(
         })
         return
       } else {
+        log.error(stderr.toString());
         reject(
           new Error(
             `Git returned an unexpected exit code '${code}' which should be handled by the caller (${name}).'`
--- a/app/src/lib/git/spawn.ts
+++ b/app/src/lib/git/spawn.ts
@@ -132,7 +132,7 @@ export async function spawnAndComplete(
         })
         return
       } else {
-        log.error(stderr.toString());
+        log.error(stderr.toString())
         reject(
           new Error(
             `Git returned an unexpected exit code '${code}' which should be handled by the caller (${name}).'`


tidy-dev (dismissed):
✨ Thanks for that change. Doesn't hurt to add more context to the logs. :)

esseivan:
I fixed the formatting error (removed semicolon)

app/src/lib/git/spawn.ts
```diff
@@ -132,6 +132,7 @@ export async function spawnAndComplete(
         })
         return
       } else {
+        log.error(stderr.toString())
```
sergiou87: I'm a bit hesitant about this, mainly because there are no other logs in this file, and this log doesn't give any context about when or how was that `stderr` obtained. Maybe it's better to add it to the description of the `Error` in the `reject`?

sergiou87 (changes requested):

app/src/lib/git/spawn.ts
sergiou87: @niik you probably know better the whole chain of error handlers. What do you think about this ☝️ ?

Another idea is adding the `stderr` as the `cause` of the error:
`throw new Error('blah blah blah', { cause: stderr })`

tidy-dev: FWIW, I pulled this and tested (just by modifying the if to always fall to the reject, so not a real error scenario) to what it would look like for this to happen. I checked that we use `log.error` in other `lib/git` files.  We have a couple other places where we just `log.error(error)` in conjunction with throwing a more human readable error. My thought is that the human readable error would lead to checking the logs and you would now see the stderr output. 

That said, adding a preface to the error would make it even better and we do that way more than not in the code base. I think just something like `Error in spawnAndComplete, stderr: ${stdErr}` seems good here. 

I believe the code is working on the assumption that the any expected error would have been handled by the caller and no more context is needed (as the error given states :D ) and probably why we didn't think to log out the error before hand, but the error this user received here was due to a faulty environment  - not something we would usually expect to handle in a way other than to just log errors so that can be determined.
 
image

tidy-dev: TLDR: I think either log like pr suggests (adding prefacing as Sergio suggested) or go ahead and append to the dialog error since this scenario is very unexpected and you would want to go look for this extra context anyways.

esseivan: > That said, adding a preface to the error would make it even better and we do that way more than not in the code base. I think just something like `Error in spawnAndComplete, stderr: ${stdErr}` seems good here.

I added that in the preface :

image

niik: > @niik you probably know better the whole chain of error handlers. What do you think about this ☝️ ?

Honestly I think we should get this in line with the errors produced from the core.ts/git function so that it raises GitErrors with the rawMessage set to true for cases where we can't map the error to something we know could be going wrong (too large output for example). I know I've looked at this before but I can't remember why I didn't just do it then, either I got pulled away to something else or there was some piece of this that lived in dugite that was hard to extract. But we shouldn't have different error paths depending on whether Git is invoked through core.ts or spawn.ts.

esseivan: I don't really know how I would implement it, I'm not very aware how on the whole program works

tidy-dev:
@esseivan Thank you for wanting contribute and sorry it took so long to respond back to this. We discussed this in a team meeting today and would really prefer the approach @niik discussed in https://github.com/desktop/desktop/pull/16015#discussion_r1095944678. Thus, we closing this until we have time to circle back an implement it.

PR #16029, opened 2023-01-29
Add some IDEs from JetBrains Toolbox to Linux

adil192:
## Description

I've added the non-snap locations of PhpStorm and WebStorm, and I've also added IntelliJ IDEA, PyCharm and Android Studio.

I'm not sure about the snap locations but I've added them in for completeness. If I should remove them, let me know :)

## Release notes

Notes: Added editors installed with JetBrains Toolbox on Linux

### DIFF
--- a/app/src/lib/editors/linux.ts
+++ b/app/src/lib/editors/linux.ts
@@ -63,12 +63,24 @@ const editors: ILinuxExternalEditor[] = [
     paths: ['/usr/bin/lite-xl'],
   },
   {
-    name: 'Jetbrains PhpStorm',
-    paths: ['/snap/bin/phpstorm'],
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
Thanks for your contribution!

sergiou87 (changes requested):
There are linter issues :( Please fix them (`yarn lint:fix` should do it for you) and I'll trigger the CI jobs again

sergiou87 (approved):

jordan4ibanez:
I've read through this, and I still cannot get IntelliJ to show up. I've installed a desktop entry as both privileged and unprivileged and it still does not show up as an option. :(

PR #16045, opened 2023-02-01
Add axeDevTools extension

tidy-dev:
## Description
This makes it so when developing we can use the accessibility chrome web extension of [axe DevTools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) during local development.

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/main-process/main.ts
+++ b/app/src/main-process/main.ts
@@ -729,7 +729,13 @@ function createWindow() {
       electron: '>=1.2.1',
     }
 
-    const extensions = [REACT_DEVELOPER_TOOLS, ChromeLens]
+    const axeDevTools = {
+      id: 'lhdoppojpmngadmnindnejefpokejbdd',
+      electron: '>=1.2.1',
+      Permissions: ['tabs', 'debugger'],
+    }
+
+    const extensions = [REACT_DEVELOPER_TOOLS, ChromeLens, axeDevTools]
 
     for (const extension of extensions) {
       try {


app/src/main-process/main.ts
```diff
@@ -729,7 +729,13 @@ function createWindow() {
       electron: '>=1.2.1',
     }
 
-    const extensions = [REACT_DEVELOPER_TOOLS, ChromeLens]
+    const axeDevTools = {
+      id: 'lhdoppojpmngadmnindnejefpokejbdd',
+      electron: '>=1.2.1',
```
tidy-dev, line 734: Copied this electron line from the ChromeLens extension above. If I do not have it, the extension fails to load with an error saying the electron version doesn't make the required undefined range. Thus, that version number is arbitrary.

app/src/main-process/main.ts
```diff
@@ -729,7 +729,13 @@ function createWindow() {
...
+      Permissions: ['tabs', 'debugger'],
```
tidy-dev, line 735: Without this, we get Manifest 3 google chrome extension warnings and the scan feature in this tool fails to run.

tidy-dev: I believe this allows access to the "tab" that is open.

andrialexandrou:
Super!

sergiou87 (approved):
Nice addition!

PR #16095, opened 2023-02-09
Refactor Pull Request review dialog

sergiou87:
## Description

This work is part of supporting new notifications for PR comments in order to reuse the same dialog and look&feel we have for PR reviews.

Changes:
1. Extract all reusable code into a `PullRequestCommentLike` component
2. Remove the `number_of_comments` attribute (which wasn't used)
3. Use the right `key` for the dialog.

## Release notes

Notes: no-notes

PR #16096, opened 2023-02-09
Add new tool to easily test notifications from real data

sergiou87:
## Description

This PR is based on #16095 

It adds a new tool to test notifications with real data: it fetches all pull request reviews and comments and generates fake Alive events that trigger notifications just like if they came from real Alive events, making testing much easier.

It's not the prettiest thing in the world but it's not supposed to be either 😂  Just an internal tool to play with.

### Screenshots

https://user-images.githubusercontent.com/1083228/217815744-744fbb60-856b-479e-8b75-cccbbf9eed85.mov

## Release notes

Notes: no-notes

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -508,6 +508,15 @@ export interface IAPIPullRequestReview {
     | 'CHANGES_REQUESTED'
 }
 
+/** Represents both issue comments and PR review comments */
+export interface IAPIComment {
+  readonly id: number
+  readonly body: string
+  readonly html_url: string
+  readonly user: IAPIIdentity
+  readonly created_at: string
+}
+
 /** The metadata about a GitHub server. */
 export interface IServerMetadata {
   /**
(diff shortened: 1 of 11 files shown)


tidy-dev (approved):
✨  Awesome! Appears to do as intended. 

Some feedback, but since this is a dev tool, nothing to hold the PR back on:
- Took me a minute to find it cause the menu item is "Show notifications" but the dialog is entitled "Test notifications" (what I was looking for) :P (I also tested before reading the code)
- After clicking on a PR, it would be nice to see a heading or something telling you which PR comments you are looking at.
- Would be cool to also do pr check failed notifications through this (could wipe out my one since this is a lot more full fledged). 

Other notes: 
- (discussed privately) The pull request comments are not implemented yet (in another PR) so in the notification store so I only tested that so far as listing the comments out in the tool.

sergiou87:
Thank you for your feedback!! 💖 

I think you made great points there, I will tackle those changes in a different PR 🙇‍♂️ Check failed notifications are definitely in the roadmap, but required a bit of extra work, so I decided to "ship" this to unblock myself for PR comment notifications 😅

PR #16098, opened 2023-02-09
Make file status announced with voice over

tidy-dev:
## Description
This PR adds a `aria-label` to the parent div of the renderer of a changed file and adds an `aria-hidden = true` to the path label. The new aria-label is what will be read by the screen reader and includes the path label and status indicated by the icon.

The reason the `aria-hidden = true` is required is because this div is a child of a html element with a role of "option" (This div in the `ListRow` component). That will make it so that any descendant text is the label read. If we mark it as hidden, then, it will look for an `aria-label`. 

Note: This may not be a final solution, the goal will be to be able to provide a custom `aria` attributes to the div in the `ListRow` component. This would allow us to set any `aria` attributes that a given list could need.

### Screenshots
Tested on Windows with JAWS and NVDA.
Tested on MacOS voice over as shown below. 
https://user-images.githubusercontent.com/75402236/218134491-2eb94887-a2d9-4d61-a0e7-cb8229629f73.mov

## Release notes
Notes: [Improved] The file status of a changed file is announced by voice over software.

### DIFF
--- a/app/src/ui/changes/changed-file.tsx
+++ b/app/src/ui/changes/changed-file.tsx
@@ -58,18 +58,20 @@ export class ChangedFile extends React.Component<IChangedFileProps, {}> {
       filePadding -
       statusWidth
 
+    const ariaLabel = `${path} - File status: ${mapStatus(status)}`
+
     return (
-      <div className="file" onContextMenu={this.onContextMenu}>
+      <div
+        className="file"
+        onContextMenu={this.onContextMenu}
+        aria-label={ariaLabel}
+      >
         <TooltippedContent
           tooltip={checkboxTooltip}
           direction={TooltipDirection.EAST}
           tagName="div"
         >
           <Checkbox
-            // The checkbox doesn't need to be tab reachable since we emulate
-            // checkbox behavior on the list item itself, ie hitting space bar
-            // while focused on a row will toggle selection.
-            tabIndex={-1}
             value={this.checkboxValue}
             onChange={this.handleCheckboxChange}
             disabled={disableSelection}
--- a/app/src/ui/changes/changed-file.tsx
+++ b/app/src/ui/changes/changed-file.tsx
@@ -58,20 +58,18 @@ export class ChangedFile extends React.Component<IChangedFileProps, {}> {
       filePadding -
       statusWidth
 
-    const ariaLabel = `${path} - File status: ${mapStatus(status)}`
-
     return (
-      <div
-        className="file"
-        onContextMenu={this.onContextMenu}
-        aria-label={ariaLabel}
-      >
+      <div className="file" onContextMenu={this.onContextMenu}>
         <TooltippedContent
           tooltip={checkboxTooltip}
           direction={TooltipDirection.EAST}
           tagName="div"
         >
           <Checkbox
+            // The checkbox doesn't need to be tab reachable since we emulate
+            // checkbox behavior on the list item itself, ie hitting space bar
+            // while focused on a row will toggle selection.
+            tabIndex={-1}
             value={this.checkboxValue}
             onChange={this.handleCheckboxChange}
             disabled={disableSelection}
--- a/app/src/ui/changes/changed-file.tsx
+++ b/app/src/ui/changes/changed-file.tsx
@@ -58,14 +58,8 @@ export class ChangedFile extends React.Component<IChangedFileProps, {}> {
       filePadding -
       statusWidth
 
-    const screenReaderDescription = `${path}  ${mapStatus(status)}`
-
     return (
-      <div
-        className="file"
-        onContextMenu={this.onContextMenu}
-        aria-label={screenReaderDescription}
-      >
+      <div className="file" onContextMenu={this.onContextMenu}>
         <TooltippedContent
           tooltip={checkboxTooltip}
           direction={TooltipDirection.EAST}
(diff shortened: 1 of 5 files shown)


tidy-dev:
Received some feedback from Accessibility folks and what was I was saying about finding a way to keep just up/down arrow navigation with additionally announcement of checked/unchecked and file status is what the goal will be... just got to figure out how to do that and it may require a little more impactful change to the list component.

JoyceZhu:
Is there any reason in particular we definitely need a `label` element wrapping these menu items, which aren't inputs? I think part of why the screen reader isn't behaving as you might expect is because of this behavior causing the `label` contents to override anything else. (This is what the ESLint-disabled rule before the change in the diff is trying to prevent -- https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md). Based on [the specification for computing accessible names/descriptions](https://www.w3.org/TR/accname-1.2/#mapping_additional_nd_te), if you remove the wrapping `label`, I wonder if `aria-describedby`/`aria-label` will work better for you.

tidy-dev:
@JoyceZhu Thank you for the suggestion. It still appears to require the aria-hidden to function. But, it is true that this doesn't make sense as a label component at this time since it isn't not directly connected to the checkbox component. 

Reviewer notes:
- The path label component is used in the changes list, the history files list, and above the diff in the changes view to display the selected file. From visual inspection and looking at css, this change should not impact the usage.

app/src/ui/changes/changed-file.tsx
```diff
@@ -58,8 +58,14 @@ export class ChangedFile extends React.Component<IChangedFileProps, {}> {
...
       statusWidth
 
+    const screenReaderDescription = `${path}  ${mapStatus(status)}`
+
     return (
-      <div className="file" onContextMenu={this.onContextMenu}>
+      <div
+        className="file"
+        onContextMenu={this.onContextMenu}
+        aria-label={screenReaderDescription}
```
JoyceZhu: @tidy-dev looks like I missed that this is also trying to use `aria-label` on a generic HTML element (a `div` or a `span`) without any semantics attached to it (it's not marked up as `role="button"`, `role="link"`, etc.) That's likely also contributing to the mysterious behavior -- `aria-label` on a non-interactive element just doesn't consistently work. You can see [our guidance on a corresponding ERBLint rule](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-aria-label-misuse-counter.md) for some more info/suggested remediation (unfortunately I don't think the ESLint plugin we borrowed from elsewhere has this rule yet).

Based on this code, would you describe this element as a button (it seems there's a  click handler on it)? If so, I'd try adding `role="button"; `tabindex=0`; and a focus indicator. That'll help with keyboard accessibility and the explicit `role=button` would likely help with getting this label read out.

tidy-dev: Thank you for that link. Useful information.

Some context. The problems here originate from a html structure problem. We have a List component that is optimized for performance when rendering long list of data... and in doing so has sacrificed some things like normal HTML structuring for lists.

This list component places each row in a div that has the role of option. The div is somewhat agnostic from actual content which is render through a callback method. That is where this content is. 

Thus, I think the ideal solution is to be able to provide the aria-label (or aria-labelledby) on that containing div with the role of option. 

Since you have pointed out now that putting `aria-label` on a div is also against the standard practices and that the div does not have a truly have the role as you are suggesting (because it is a child of the one that does).. I think I will revisit my approach on this to see if I can solve the aria-label issue.

tidy-dev:
@niik  I know we discussed having an `aria-labelledby` and linking it to an id of a child below (doing the whole id generation/releasing stuff). But, I think that child below in this scenario would have to a hidden div. I think that is messier than using `aria-label`. (Not that there there couldn't be cases where it makes sense) Thus, for now, I am adding a callback method to be able to add the `aria-label` attribute to the proper parent div that has a `role` applied to it.  This solves the issue without [breaking standard practices](https://github.com/desktop/desktop/pull/16098#discussion_r1106356954) brought up by @JoyceZhu 

(Also, pinging you here as this issue may impact the list tickets you are working on).

app/src/ui/lib/path-label.tsx
```diff
@@ -41,19 +44,17 @@ export class PathLabel extends React.Component<IPathLabelProps, {}> {
         ? availableWidth / 2 - ResizeArrowPadding
         : undefined
       return (
-        // eslint-disable-next-line jsx-a11y/label-has-associated-control
-        <label {...props}>
+        <span {...props} aria-hidden={this.props.ariaHidden}>
```
tidy-dev, line 47: Talked with Markus on this, I think the historical context was that the goal was to tie this label to the checkbox in the changes list. But, now that is not applicable (plus it is used in other parts of the app that do not have check boxes like the history list). Therefore, this doesn't make sense as a span anymore.

tidy-dev: Note on aria-hidden, I found that even if I apply the `aria-label` attribute to the parent div that has the `role` of `option` as list items do. The default behavior is to use any child text as the accessibility label. Thus, this needs to be set to false in order for the `aria-label` to take precedence. 

Other things I tried was to apply an `aria-label` to the `svg` tag in the `octicon` component, but that was not picked up either. 

The only other thing that worked was to put a hidden div/span. That is still an option, but I think using the `aria-label` is programmatically more intentional/easier to understand what the purpose of at a glance.

niik (approved):

PR #16100, opened 2023-02-09
Commit Message/Misattributed commits Avatar Keyboard Accessible

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3269 (Sorry not external link)

## Description
The user avatar in the commit message area was not keyboard navigable to. Thus, when the misattributed warning was visible, a user could not access it with the keyboard. This updates the component to be a button component so that is is tab accessible.

Bonus: This gets rid of two accessibility linter disables!

### Screenshots

https://user-images.githubusercontent.com/75402236/217841381-361b8a29-f596-4121-8629-bd8d7eae5807.mov

## Release notes
Notes: [Improved] The misattributed warning popover is accessible through keyboard navigation.

### DIFF
--- a/app/src/ui/changes/commit-message-avatar.tsx
+++ b/app/src/ui/changes/commit-message-avatar.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable jsx-a11y/no-static-element-interactions */
-/* eslint-disable jsx-a11y/click-events-have-key-events */
 import React from 'react'
 import { Select } from '../lib/select'
 import { Button } from '../lib/button'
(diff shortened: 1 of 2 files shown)


sergiou87 (commented):
Looks good but I noticed something off… it's like there is a square (with rounded corners) around the avatar circle:

image

Left is prod, right is dev

If I use some tool to fill that "square" with red…
image

tidy-dev:
Thanks! I am staring really hard at those two first images and am not seeing the difference except maybe the right one is smaller.. but I can't tell if that is just the image. :D . I think you are saying that the outside button should have the same border radius as the inner avatar; I will see about doing that.

tidy-dev:
Here is button having the same border radius as the avatar image. This is also showing what it looks like when it is focused - which emphasizes that the button is not "square-ish", but is round to match the avatar.

image

image

sergiou87 (dismissed):
Works great now, thanks!! 🚀

sergiou87 (approved):
False alarm, it's all good! :shipit:

PR #15878, opened 2023-01-06
Use setup-node action to cache dependencies

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

Updated workflows to cache dependencies using [actions/setup-node](https://github.com/actions/setup-node#caching-global-packages-data). `setup-node@v3` or newer has caching **built-in**.

### AS-IS

```yaml
- name: Use Node.js ${{ matrix.node }}
  uses: actions/setup-node@v3
  with:
    node-version: ${{ matrix.node }}

- name: Get yarn cache directory path
  id: yarn-cache-dir-path
  run: echo "::set-output name=dir::$(yarn cache dir)"

- uses: actions/cache@v3
  id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
  with:
    path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
    key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
    restore-keys: |
      ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
      ${{ runner.os }}-yarn-
```

### TO-BE

```yml
- name: Use Node.js ${{ matrix.node }}
  uses: actions/setup-node@v3
  with:
    node-version: ${{ matrix.node }}
    cache: yarn
```

> It’s literally a one line change to pass the `cache: yarn` input parameter.

## References

- [https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

PR #16103, opened 2023-02-09
Use environment files instead of `set-output`

kieferro:
## Description

This PR replaces the use of `set-output`, which has been deprecated, in the Github Actions workflow. For more information regarding the deprecation, please see [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

## Release notes

Notes: no-notes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,8 @@ jobs:
 
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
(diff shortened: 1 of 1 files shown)


sergiou87:
Thank you for your contribution @kieferro ! But I think this won't be necessary once/if we accept #15878 which just removes the need for set-output entirely 😄

kieferro:
:+1: I had already suspected that this change might already exist somewhere, but I didn't stumble across this PR while searching.

PR #16108, opened 2023-02-10
Get rid of blank role attribute and rowgroup role on the innerGrid

niik:
This was incorrectly removed in https://github.com/desktop/desktop/commit/8562094757150be508ce07cf0847b746fcc07872. There's a difference between having a blank role attribute and not having the attribute at all (which is what we want here). Additionally we don't want the inner grid to have role `rowgroup` because we're using role listbox on the containing div.

### DIFF
--- a/app/src/ui/index.tsx
+++ b/app/src/ui/index.tsx
@@ -346,12 +346,21 @@ ipcRenderer.on('url-action', (_, action) =>
   dispatcher.dispatchURLAction(action)
 )
 
-// react-virtualized will use defaults if we pass undefined or omit
-// the aria-label, role, and containerRole props and we want to set
-// our own a11y props on the parent of the Grid component which is
-// why we're hacking our way around this here.
+// react-virtualized will use the literal string "grid" as the 'aria-label'
+// attribute unless we override it. This is a problem because aria-label should
+// not be set unless there's a compelling reason for it[1].
+//
+// Similarly the default props call for the 'aria-readonly' attribute to be set
+// to true which according to MDN doesn't fit our use case[2]:
+//
+// > This indicates to the user that an interactive element that would normally
+// > be focusable and copyable has been placed in a read-only (not disabled)
+// > state.
+//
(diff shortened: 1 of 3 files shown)


tidy-dev (dismissed):
✨  Thank you!

tidy-dev:
Note.. just noticed that this causes console errors.
image

niik:
> Note.. just noticed that this causes console errors.

Welp, this turned this small change into a bigger one. I ended up moving all the a11y props and the keyboard event handler down to the inner grid instead of the container div. I also added a hack to unset the default props and prop requirements of the Grid component because they don't make a lot of sense in our case.

app/src/ui/index.tsx
```diff
@@ -352,6 +353,26 @@ ipcRenderer.on('url-action', (_, action) =>
...
+// > state.
+//
+// 1. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label
+// 2. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-readonly
+;(function (defaults: Record<string, unknown>, types: Record<string, unknown>) {
+  ;['aria-label', 'aria-readonly'].forEach(k => {
+    delete defaults[k]
+    delete types[k]
+  })
+})(Grid.defaultProps, Grid.propTypes)
```
tidy-dev, line 374: Meh.. I was wondering if we were going to need to "manually" remove these props.

tidy-dev (approved):
✨ Works great!

This definitely resolves, [#3245](https://github.com/github/accessibility-audits/issues/3245) - announcing list item positioning.

And I believe it fixed [#3261](https://github.com/github/accessibility-audits/issues/3261) as well. When I tested that last one on windows, I got "list clickable" not "out of grouping list clickable" as described in the ticket. With this fix, I got "list". (On mac, it was "listbox", now it is "empty, listbox")

PR #16110, opened 2023-02-10
Make the close button a button in the dialog component; allow it to have focus

tidy-dev:
## Description
We have an accessibility issue in that our close button in our dialogs is not keyboard navigable to. This changes the anchor tag to a button tag so that it gets added to the normal tab order.  Additionally, I removed the css that prevented seeing an outline when it is focused as we do want users to know when it is focused.

As you can see, we [intentionally made this an anchor element in the past](https://github.com/desktop/desktop/pull/985). The stated reason is that it was grabbing focus from the element that we were trying to designated as focused. I have opened half a dozen modals and all them the focus immediately goes to the contents input. Thus, I am assuming that with all the chromium updates since that PR was entered in 2017, that this is not an issue anymore. 

Bonus.. we get rid of an ally linter disable

### Screenshots

https://user-images.githubusercontent.com/75402236/218148116-2bd0c261-8738-4e74-996c-7faf742199e0.mov

## Release notes
Notes: [Improved] The close button on dialogs is keyboard accessible.

### DIFF
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -291,6 +291,8 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
   }
 
   protected onDialogIsTopMost() {
+    console.log('onDialogIsTopMost')
+
     if (this.dialogElement == null) {
       return
     }
@@ -382,6 +384,8 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
    *
    *  4. Any remaining button
    *
+   *  5. The dialog close button
+   *
    */
   private focusFirstSuitableChild() {
     const dialog = this.dialogElement
@@ -411,6 +415,8 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
     // anchor tag masquerading as a button)
     let firstTabbable: HTMLElement | null = null
 
+    const closeButton = dialog.querySelector(':scope > header button.close')
+
     const excludedInputTypes = [
       ':not([type=button])',
       ':not([type=submit])',
@@ -423,6 +429,7 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
     const inputSelector = `input${excludedInputTypes.join('')}, textarea`
     const buttonSelector =
       'input[type=button], input[type=submit] input[type=reset], button'
+
     const submitSelector = 'input[type=submit], button[type=submit]'
 
     for (const candidate of dialog.querySelectorAll(selector)) {
@@ -444,16 +451,28 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
         candidate.matches(submitSelector)
       ) {
         firstSubmitButton = candidate
-      } else if (firstButton === null && candidate.matches(buttonSelector)) {
+      } else if (
+        firstButton === null &&
+        candidate.matches(buttonSelector) &&
+        candidate !== closeButton
+      ) {
         firstButton = candidate
       }
     }
 
-    const newActive =
-      firstExplicit[1] || firstTabbable || firstSubmitButton || firstButton
+    const focusCandidates = [
+      firstExplicit[1],
+      firstTabbable,
+      firstSubmitButton,
+      firstButton,
+      closeButton,
+    ]
 
-    if (newActive !== null) {
-      newActive.focus()
+    for (const focusCandidate of focusCandidates) {
+      if (focusCandidate instanceof HTMLElement) {
+        focusCandidate.focus()
+        break
+      }
     }
   }
 
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -291,8 +291,6 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
   }
 
   protected onDialogIsTopMost() {
-    console.log('onDialogIsTopMost')
-
     if (this.dialogElement == null) {
       return
     }
--- a/app/src/ui/dialog/header.tsx
+++ b/app/src/ui/dialog/header.tsx
@@ -47,7 +47,10 @@ interface IDialogHeaderProps {
  * might be necessary to use this component directly.
  */
 export class DialogHeader extends React.Component<IDialogHeaderProps, {}> {
-  private onCloseButtonClick = () => {
+  private onCloseButtonClick = (e: React.MouseEvent<HTMLButtonElement>) => {
+    /** This prevent default is a preventative measure since the dialog is akin
+     * to a big Form element. We wouldn't any surprise form handling. */
+    e.preventDefault()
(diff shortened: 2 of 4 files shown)


app/src/ui/dialog/header.tsx
```diff
@@ -58,21 +58,14 @@ export class DialogHeader extends React.Component<IDialogHeaderProps, {}> {
...
 
-    // We're intentionally using <a> here instead of <button> because
-    // we can't prevent chromium from giving it focus when the the dialog
-    // appears. Setting tabindex to -1 doesn't work. This might be a bug,
-    // I don't know and we may want to revisit it at some point but for
-    // now an anchor will have to do.
     return (
-      // eslint-disable-next-line jsx-a11y/anchor-is-valid, jsx-a11y/click-events-have-key-events, jsx-a11y/interactive-supports-focus
-      <a
+      <button
```
tidy-dev, line 65: I chose not to use our `<Button>` component here since this is clearly custom styled and I didn't want to introduce anything from our `<Button>` component.

sergiou87:
> I have opened half a dozen modals and all them the focus immediately goes to the contents input.

That is my experience when there are text boxes in the contents or "call-to-actions", but otherwise the close button gets focused. The only example of this that I've found is the app settings dialog: if I haven't signed out in one of my accounts, the focus goes to the `Sign in` button of such account. Otherwise, it focuses the close button.

https://user-images.githubusercontent.com/1083228/218426122-a754a97c-3fbc-4b7c-ad88-ebd7f1e0913c.mov

I wonder if we can make it programatically so that `Dialog` components focus its contents? 🤔

niik:
> The stated reason is that it was grabbing focus from the element that we were trying to designated as focused

I believe this got changed in https://github.com/desktop/desktop/pull/8599, we used to rely on browser heuristics only but since moved to an approach more suited to our dialog structure.

> I wonder if we can make it programatically so that `Dialog` components focus its contents? 🤔

As above, initial focus in dialogs is handled by `focusFirstSuitableChild` https://github.com/desktop/desktop/blob/b0c4b63ea6737270663cba67f38e4c0ef8b4a029/app/src/ui/dialog/dialog.tsx#L336-L386. That needs to be tweaked here to deprioritize the close button now that it's focusable. Should be a pretty straightforward change, I'll push it up here so that we can take a look at it.

For context we made the call to not have the button be focusable because it's not a common convention on either macOS or Windows to be able to reach dialog close buttons with keyboard focus. We instead opted to follow the platform conventions of being able to close a dialog with <kbd>Command+W</kbd> and <kbd>Escape</kbd>

sergiou87:
> For context we made the call to not have the button be focusable because it's not a common convention on either macOS or Windows to be able to reach dialog close buttons with keyboard focus. We instead opted to follow the platform conventions of being able to close a dialog with Command+W and Escape

That is actually a very good point 🤔

tidy-dev:
Asked in #accessibility about keyboard shortcuts/platform conventions:
> I can’t speak for the auditors but given that on the web, a dialog must close with the escape key but also have a focusable close button (X on our primer dialogs) I would recommend we follow that convention on the desktop app as well. In addition to sign-posting the close functionality for users, it will also give a common element to help screen reader users know when they’ve explored the entire dialog.

Seems to make sense to me. Thoughts @niik, @sergiou87?

(Also side question, I don't think there are any issues to address in this PR from your all's comments right? - I think @niik resolved @sergiou87 concern)

app/src/ui/dialog/dialog.tsx
```diff
@@ -291,6 +291,8 @@ export class Dialog extends React.Component<IDialogProps, IDialogState> {
   }
 
   protected onDialogIsTopMost() {
+    console.log('onDialogIsTopMost')
```
niik: Whoops! Let me take care of that 🤦

app/src/ui/dialog/header.tsx
```diff
@@ -47,7 +47,7 @@ interface IDialogHeaderProps {
  * might be necessary to use this component directly.
  */
 export class DialogHeader extends React.Component<IDialogHeaderProps, {}> {
-  private onCloseButtonClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
+  private onCloseButtonClick = () => {
```
niik: I _think_ we should prevent the default action of this event here now that we're turning into a button. It's _probably_ okay but I'm a bit wary given that the Dialog component is just a big `<Form />` with attached handlers for submission and resets. With the anchor tag we were safe but I'd be much more comfortable if we called preventDefault here.

niik (changes requested):

app/src/ui/dialog/header.tsx
tidy-dev: Added and tested. Makes sense.. I put a comment about it being as a preventative measure for this so that if we find it hurts something in the future, we know this was the reason.

niik (approved):

tidy-dev:
More Context from #accessibility. (Posting here for reference purposes).

Question: We use platform conventions of `Command+W` and `Escape` to allow a keyboard accessible way to close dialogs. Does Desktop need to also have the 'x' button keyboard focusable?

James Scholes
> Command+W (or Ctrl+W) isn't a convention to close dialogs on Windows, only on macOS.  On Windows it closes tabs in a tabbed UI, where such functionality is available to the user.  So we probably shouldn't be offering that keystroke on Windows.

> Secondly: it is true that Windows presents buttons to close and resize all windows, including dialogs, as well as to open the "System" menu.  These are not focusable with the keyboard, and are never directly used by keyboard users.
Instead, the following alternatives are available:
Alt+F4: close the top-level window, including dialogs.
Alt+Space: open the System menu, which has options like "Close", "Minimise", etc.
Note that these do not work as intended inside dialogs rendered via web content, regardless of whether we're talking about an Electron app or one in the browser.  That's because the functionality is provided by the window manager, and your custom dialogs inside a web page are not windows according to its logic.

> Escape also works to close native dialogs, and is easier to press than a keystroke involving a modifier (Alt+F4), and hence is commonly used.

> The key takeaway there is: the close button inside your dialogs must be keyboard-accessible, because you can't rely on the window manager to do your work for you.

> On macOS, the close button for regular windows is accessible with FKA and VoiceOver, and hence must be keyboard-and-screen-reader-accessible in your implementation too.

> If you want to avoid having to do any of this, find a way to render a proper Win32 and UIKit modal from Electron.  No idea if this is possible, but they managed native menu bar integration so...

> Also, one tiny point of Nuance: while the close button provided by the window manager in native Windows dialogs isn't keyboard-accessible, Windows dialogs almost always have a secondary means of dismissal.  E.g. a "No" or "Cancel" button, and these are always in the tab order.

> Also also: this only applies inside dialogs.  The actual window of your app benefits from everything I've outlined here already, and should have system-provided affordances that are accessible to close it, resize it, etc.

PR #16111, opened 2023-02-10
Avatar only a button when it needs to be

tidy-dev:
## Description
Follow up feedback on https://github.com/desktop/desktop/pull/16100 -> the avatar doesn't make sense as a button when it can't be clicked (when there isn't a misattributed commit warning) as the user has to tab there and is told it is a button, but hitting enter does nothing in a regular use case.

### Screenshots
https://user-images.githubusercontent.com/75402236/218152382-3cfab66b-f6e2-48e7-8405-4b01afc5d7d6.mov

## Release notes
(Use the note from last PR)
Notes: no-notes


sergiou87 (approved):
Changes look good and works as expected!

PR #16114, opened 2023-02-10
Invoking Context Menu from ListRow component

tidy-dev:
Based on https://github.com/desktop/desktop/pull/16108

(This diff looks funny because I have development merged into this branch and that pull request above does not.)

## Description
This PR refactor the callback of `onContextMenu` to be on the focus div element of each row of the changes list. That is the div in the `<ListRow>` component. The reason being is that the keyboard shortcuts (i.e. Shift + F10 on Windows and VO "VoiceOver" + Shift + M on MacOS) to invoke a context menu will only do so for the focused div which for a selectable list item is the one with the role of `option`

## Release notes
Notes: [Improved] The context menu for a file in changed files list can be invoked by keyboard shortcuts.

### DIFF
--- a/app/src/ui/lib/list/list-row.tsx
+++ b/app/src/ui/lib/list/list-row.tsx
@@ -53,6 +53,13 @@ interface IListRowProps {
     e: React.FocusEvent<HTMLDivElement>
   ) => void
 
+  /** Called back for when the context menu is invoked (user right clicks of
+   * uses keyboard shortcuts) */
+  readonly onContextMenu?: (
+    index: number,
+    e: React.MouseEvent<HTMLDivElement>
+  ) => void
+
   /**
    * Whether or not this list row is going to be selectable either through
    * keyboard navigation, pointer clicks, or both. This is used to determine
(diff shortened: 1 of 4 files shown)


tidy-dev:
After much trial by error.. I stumbled upon the fact that @niik changes in https://github.com/desktop/desktop/pull/16108 solved the problem of the incorrect row context menu being rendered on macOS when I went to start simplying the Grid rendering to see what thing could be impacting it and I applied some of @niik's changes and it just worked! I can't really explain why.. I am assuming something about roles where impacting this. :/

app/src/ui/lib/list/list-row.tsx
```diff
@@ -94,6 +101,12 @@ export class ListRow extends React.Component<IListRowProps, {}> {
     this.props.onRowBlur?.(this.props.rowIndex, e)
   }
 
+  private onContextMenu = (e: React.MouseEvent<HTMLDivElement>) => {
+    console.log(this.props.rowIndex)
+    console.log(e.currentTarget)
```
niik: Can we get rid of these please?
```suggestion
```

niik (changes requested):
One tiny request

app/src/ui/lib/list/list-row.tsx
tidy-dev: Woops.. Absolutely!

niik (approved):

PR #16120, opened 2023-02-12
Checkout a commit

### LINKED ISSUE #10068: Ability to checkout commit directly from History tab
Hello,

It would be nice to have the feature implemented so that when you right-click a commit in the History, it adds the options to checkout commit. Currently, there does not seem to have a simple way to implement a commit checkout, and having this feature would be nice to have and make it easier to go back to a previous commit. There is currently revert commit; however, that is not always what you want to do.

kitswas:
Closes #10068 

## Description

- All existing checkout infrastructure assumed that we will be checking out branches only.
  I modified it to allow both branches and commits.
  I created the helper functions for commit checkout following those for their branch counterparts.

### Screen recording

https://user-images.githubusercontent.com/90329875/218293541-35ce9186-3d7b-41e7-b445-fd4e1997f0b5.mp4

## Release notes

Notes: Added the much desired ability to checkout a single commit from the History tab. (#10068)


app/src/ui/history/commit-list.tsx
```diff
@@ -70,6 +70,12 @@ interface ICommitListProps {
    */
   readonly onCreateBranch?: (commit: CommitOneLine) => void
 
+  /**
+   * Callback to fire to create a branch from a given commit in the current
+   * repository
```
dpizzo-at-aops: I don't think this comment got updated?

kitswas: Thanks for pointing out. Fixed.

dpizzo-at-aops:
I'm literally just an innocent bystander who wants to see this feature as well and decided to browse the PR for fun... sorry if I got your hopes up about someone who actually could merge this reviewing it 😅

kitswas:
Thanks for the review.

> if I got your hopes up

You did but don't worry about it. Any bit of activity on this PR should help.

Gakk:
I am also just an innocent bystander who wants to see this feature implemented, and after watching your video I have one remark related to the user interface.

The feature is implemented so it is easy to checkout a commit — maybe too easy?

Most other tools shows a warning when user performs an action that will create a detached head. Maybe it will be easier to accept this pull request if a warning and choice is given to the user?

_Example from Atlassian Sourcetree:_

image

kitswas:
You have raised a good point. Let's wait and see what the maintainers have to say.

> Maybe it will be easier to accept

We won't know anything till the maintainers comment on what could be acceptable.

Andrej730:
Got some idea and just wanted to put it out there.

Now when you checkout commit you see "Current branch" field changing to "Detached HEAD on hash"

image

Maybe it's just me but noticing this makes me expect that I'd be able to also drag'n'drop commit to "Current branch" to check it out.

image

tidy-dev:
@kitswas Thank you for wanting to contribute and sorry for taking so long to respond. The team has just discussed this PR and we would like to see a warning like suggested in https://github.com/desktop/desktop/pull/16120#issuecomment-1508513704. If this is still something you are interested and able to work on, let us know.

kitswas:
~~<sub>I should be studying for my exams now. But, sometimes, programming wins over reason. I have no idea why I am doing this past midnight.</sub>~~

Here you go.

screenshot

app/src/lib/app-state.ts
```diff
@@ -204,6 +204,9 @@ export interface IAppState {
   /** Should the app prompt the user to confirm a discard stash */
   readonly askForConfirmationOnDiscardStash: boolean
 
+  /** Should the app prompt the user to confirm a discard stash */
+  readonly askForConfirmationOnCheckoutCommit: boolean
```
dpizzo-at-aops: Forgot to update the comment again?

kitswas:
@tidy-dev Ready for review.

app/src/lib/stores/app-store.ts
```diff
@@ -3839,18 +3849,70 @@ export class AppStore extends TypedBaseStore<IAppState> {
     this.hasUserViewedStash = false
   }
 
-  private async refreshAfterCheckout(repository: Repository, branch: Branch) {
+  private async refreshAfterCheckout(
```
tidy-dev, line 3868: We use the git concept of `commitish` elsewhere in the code where a string may be any manner of referencing a commit whether that be a branch name, commit sha, or other ways to denote a commit. 

Could we update this method to take a single parameter of `commitish: string` instead of two  possibly null parameters - update he use cases to send in the narrowed string ref?  Along that same lines, since checking out a branch and checking out a commit share the same progress interface. Could we also update the `ICheckoutProgress` interface in `progress.ts` to be `commitish` instead of target branch and update the doc to reflect that it could a branch name or a commit.

tidy-dev (changes requested):

tidy-dev:
Still looking through this PR a bit, some stuff overlaps heavily with checkout branch process and going to confer with team on where it makes sense to share logic.

tidy-dev (changes requested):
We usually pin new features behind a feature flag. See `feature-flag.ts` for examples. Any entry point to the new feature should be flagged.

kitswas:
> We usually pin new features behind a feature flag. See `feature-flag.ts` for examples. Any entry point to the new feature should be flagged.

With Development Features enabled / `GITHUB_DESKTOP_PREVIEW_FEATURES` environment variable set:

image

app/src/lib/stores/app-store.ts
kitswas: Addressed in 10f673f39260178ce94dcfccf2ee86bdb8d61233. Please check.

app/src/lib/git/checkout.ts
```diff
@@ -110,6 +117,72 @@ export async function checkoutBranch(
...
+          })
+        }
+      }
+    )
+
+    // Initial progress
+    progressCallback({ kind, title, value: 0, target: targetCommit })
+  }
+
+  const args = ['checkout', commit.sha]
```
tidy-dev: This needs to implement the first part of the `getCheckoutArgs` - the part about adding the `gitNetworkArguments()` and the `--progress` flag.

app/src/ui/history/commit-list-item.tsx
```diff
@@ -305,6 +309,17 @@ export class CommitListItem extends React.PureComponent<
       })
     }
 
+    if (enableCheckoutCommit()) {
+      items.push({
+        label: __DARWIN__ ? 'Checkout Commit' : 'Checkout commit',
+        action: () => {
+          if (this.props.onCheckoutCommit) {
+            this.props.onCheckoutCommit(this.props.commit)
+          }
```
tidy-dev:
```suggestion
          this.props.onCheckoutCommit?.(this.props.commit)
```

Just short hand

app/src/ui/history/compare.tsx
```diff
@@ -599,6 +601,21 @@ export class CompareSidebar extends React.Component<
     })
   }
 
+  private onCheckoutCommit = (commit: CommitOneLine) => {
+    const { repository, dispatcher, askForConfirmationOnCheckoutCommit } =
+      this.props
+    // console.log("Confirm checkout" + askForConfirmationOnCheckoutCommit)
```
tidy-dev:
```suggestion
```

app/src/lib/stores/app-store.ts
```diff
@@ -3847,18 +3857,72 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+        .catch(e =>
+          this.emitError(
+            // new CheckoutError(e, repository, branch)
+
+            new ErrorWithMetadata(e, {
+              gitContext: { kind: 'checkout', commitToCheckout: commit },
+              repository,
+            })
+          )
+        )
```
tidy-dev:
```suggestion
        .catch(e => this.emitError(new Error(e)))
```

In the `_checkoutBranch` method, the `CheckoutError` extending the `ErrorWithMetadata` is currently only needed to pass some metadata for specific error handling implemented for checking out a branch that is not handled in this PR (such as providing the ability to stash before checking out).

app/src/lib/git-error-context.ts
```diff
@@ -18,7 +19,16 @@ type CheckoutBranchErrorContext = {
...
+
+  /** The commit we were trying to checkout */
+  readonly commitToCheckout: CommitOneLine
+}
+
 /** A custom shape of data for actions to provide to help with error handling */
 export type GitErrorContext =
   | MergeOrPullConflictsErrorContext
   | CheckoutBranchErrorContext
+  | CheckoutCommitErrorContext
```
tidy-dev:
```suggestion
```

app/src/lib/git/checkout.ts
```diff
@@ -110,6 +117,72 @@ export async function checkoutBranch(
...
+        if (progress.kind === 'progress') {
+          const description = progress.details.text
+          const value = progress.percent
+
+          progressCallback({
+            kind,
+            title,
+            description,
+            value,
+            target: targetCommit,
```
tidy-dev: Could we use the `shortenSHA` helper method so that when it is briefly displayed in the toolbar it isn't the long sha which is hard to digest/doesn't fit well in the toolbar. Also would apply in the `app-store` when `refreshAfterCheckout` is called.

app/src/lib/git/checkout.ts
```diff
@@ -87,13 +88,19 @@ export async function checkoutBranch(
           const description = progress.details.text
           const value = progress.percent
 
-          progressCallback({ kind, title, description, value, targetBranch })
```
tidy-dev, line 90: Just a nit: If we changed the `const targetBranch = branch.name` above to `const target = branch.name`, then this could remain one easy to digest line.

app/src/lib/git/checkout.ts
```diff
@@ -110,6 +117,72 @@ export async function checkoutBranch(
...
+      account,
+      getFallbackUrlForProxyResolve(account, repository)
+    ),
+    expectedErrors: AuthenticationErrors,
+  }
+
+  if (progressCallback) {
+    const title = `Checking out commit ${commit.sha}`
+    const kind = 'checkout'
+    const targetCommit = commit.sha
```
tidy-dev:
```suggestion
    const target = commit.sha
```

Similar minor commit to above, if this is target, than below becomes `value, target})`

app/src/ui/toolbar/branch-dropdown.tsx
```diff
@@ -159,15 +159,17 @@ export class BranchDropdown extends React.Component<
     let progressValue: number | undefined = undefined
 
     if (checkoutProgress) {
-      title = checkoutProgress.targetBranch
-      description = __DARWIN__ ? 'Switching to Branch' : 'Switching to branch'
+      title = checkoutProgress.target
+      description = __DARWIN__
+        ? 'Checking out the Target'
```
tidy-dev: Prefer to keep this more descriptive. Since we are not using the `checkoutProgress.description`, in `checkout.ts` -> `checkoutBranch`, we can set the description of the `checkoutProgress` to 'Switching to Branch` and in `checkoutCommit` to `Checking out a Commit`. Then this bit of logic can use the `checkoutProgress.description` instead.

tidy-dev (changes requested):
@kitswas Thank you for all the changes thus far and sticking with us! I talked with the team and have some feedback.

1. In `checkout.ts`, the first part of  `getCheckoutArgs` needs applied in the `checkoutCommit` method. (network and progress flags)
2. In `checkout.ts` most of `checkoutCommit` and `checkoutBranch` share the same logic. We would like to have that reflected in the methods. Maybe introduce a `checkoutCommitish` that they both call. The `checkoutBranch` would need to have a way to make sure the submodule args are accounted for that `checkoutCommit` doesn't, but I believe that is your only difference aside for some progress wording in the progress title. That could be handled as a parameter.
3. The toolbar is the only place that consumes the `checkoutProgress` that is generated. At that point, we don't know if it is branch or commit so this PR is changing the description to "Checkout out the Target". Target is not typical git vernacular for describing a branch or commit. Thus, it would be nice to update the `checkoutProgress` at the `checkout.ts` level where branch vs commit is known and use that to be able to keep `Switching to a Branch` vs `Checkout out a Commit`. 
4. Remove all the formatting changes so reviewing this PR can be focused on code changes.

Note: These may also be mentioned in individual comments to point to where I am talking about.

tidy-dev (changes requested):
@kitswas Thank you for all the changes thus far and sticking with us! I talked with the team today and have some feedback.

1. In `checkout.ts`, the first part of  `getCheckoutArgs` needs applied in the `checkoutCommit` method. (network and progress flags)
2. In `checkout.ts` most of `checkoutCommit` and `checkoutBranch` share the same logic. We would like to have that reflected in the methods. Maybe introduce a `checkoutCommitish` that they both call. The `checkoutBranch` would need to have a way to make sure the submodule args are accounted for that `checkoutCommit` doesn't, but I believe that is your only difference aside for some progress wording in the progress title. That could be handled as a parameter.
3. The toolbar is the only place that consumes the `checkoutProgress` that is generated. At that point, we don't know if it is branch or commit so this PR is changing the description to "Checkout out the Target". Target is not typical git vernacular for describing a branch or commit. Thus, it would be nice to update the `checkoutProgress` at the `checkout.ts` level where branch vs commit is known and use that to be able to keep `Switching to a Branch` vs `Checkout out a Commit`. 
4. Remove all the formatting changes so reviewing this PR can be focused on code changes.

Note: These may also be mentioned in individual comments to point to where I am talking about.

tidy-dev (changes requested):
@kitswas Thank you for all the changes thus far and sticking with us! I talked with the team and have some feedback.

1. In `checkout.ts`, the first part of  `getCheckoutArgs` needs applied in the `checkoutCommit` method. (network and progress flags)
2. In `checkout.ts` most of `checkoutCommit` and `checkoutBranch` share the same logic. We would like to have that reflected in the methods. Maybe introduce a `checkoutCommitish` that they both call. The `checkoutBranch` would need to have a way to make sure the submodule args are accounted for that `checkoutCommit` doesn't, but I believe that is your only difference aside for some progress wording in the progress title. That could be handled as a parameter.
3. The toolbar is the only place that consumes the `checkoutProgress` that is generated. At that point, we don't know if it is branch or commit so this PR is changing the description to "Checkout out the Target". Target is not typical git vernacular for describing a branch or commit. Thus, it would be nice to update the `checkoutProgress` at the `checkout.ts` level where branch vs commit is known and use that to be able to keep `Switching to a Branch` vs `Checkout out a Commit`. 
4. Remove all the formatting changes so reviewing this PR can be focused on code changes.

Note: These may also be mentioned in individual comments to point to where I am talking about.

kitswas:
> Remove all the formatting changes so reviewing this PR can be focused on code changes.

1. How may I do that?
2. I execute `yarn lint:fix` before commiting to this project. Why does the linter add spaces randomly (for indentation) and then remove them when run again (without any code changes)?
3. GitHub Desktop has a handy 'Hide whitespace changes' option that might help the reviewers.

app/src/lib/git/checkout.ts
kitswas: I was wondering about it too. I went with 'minimal changes'.
Will change.
Atleast, it was not [this bad](https://thedailywtf.com/articles/changing-the-type-of-newbie).

app/src/lib/git-error-context.ts
kitswas: Removed `CheckoutCommitErrorContext`.

tidy-dev:
> How may I do that?

I use Visual Studio Code and have prettier extenstion installed, and that is what automatically applies the formatting for me. My assumption is you have a different formatter managing your code? For now focus on the code changes, we can figure out the formatting later.

app/src/ui/history/commit-list-item.tsx
kitswas: Didn't know this.

app/src/lib/git/checkout.ts
kitswas: > Could we use the shortenSHA helper method 

Done.
image

> Also would apply in the app-store when refreshAfterCheckout is called.

How should we figure out whether the `commitish` is a SHA or not?
We had changed to a commitish in 10f673f39260178ce94dcfccf2ee86bdb8d61233.

```ts
private async refreshAfterCheckout(
    repository: Repository,
    commitish: string
  ) {
    this.updateCheckoutProgress(repository, {
      kind: 'checkout',
      title: `Refreshing ${__DARWIN__ ? 'Repository' : 'repository'}`,
      description: 'Checking out',
      value: 1,
      target: commitish,
    })

    await this._refreshRepository(repository)
    return repository
  }
```

As a side-effect of the above, both commits and branches must now receive the same treatment.

And yes, the long SHA **is** _ugly_.

image

kitswas: > How should we figure out whether the commitish is a SHA or not?

We might treat any string of 40 digits as a SHA and call `shortenSHA()` on it.  
Has a chance of failure but a string 40 digits long will be truncated anyway.

Or we could leave it as it is and call it a feature. :P

app/src/lib/git/checkout.ts
kitswas: Do you mean `checkoutCommit` should have

```ts
  const baseArgs =
    progressCallback != null
      ? [...gitNetworkArguments(), 'checkout', '--progress']
      : [...gitNetworkArguments(), 'checkout']
```

with `commit.sha` appended in `args`?

tidy-dev: Yes. In other words the `baseArgs` in the `getCheckoutArgs` need to be applied to both.

app/src/lib/git/checkout.ts
tidy-dev: We could pass in progress target as parameter, if called from checking out branch, progressTarget = branch name and if called from checkout out a commit, progressTarget = shortenSha(commit.sha)

app/src/lib/git/checkout.ts
kitswas: Fixed.

app/src/lib/git/checkout.ts
kitswas: I handled it by shortening the SHA when it was passed as commitish.

kitswas: This was probably what you initially meant.

app/src/ui/checkout/confirm-checkout-commit.tsx
```diff
@@ -0,0 +1,104 @@
...
+        disabled={this.state.isCheckingOut}
+        onSubmit={this.onSubmit}
+        onDismissed={this.props.onDismissed}
+      >
+        <DialogContent>
+          <Row>
+            Checking out a commit will create a detached HEAD, and you will no
+            longer be on any branch.
+          </Row>
+          <Row>Are you sure you want to checkout this commit?</Row>
```
tidy-dev:
```suggestion
      <Dialog
        id="checkout-commit"
        type="warning"
        title={title}
        loading={this.state.isCheckingOut}
        disabled={this.state.isCheckingOut}
        onSubmit={this.onSubmit}
        onDismissed={this.props.onDismissed}
        ariaDescribedBy="checking-out-commit-confirmation"
        role="alertdialog"
      >
        <DialogContent>
          <Row id="checking-out-commit-confirmation">
            Checking out a commit will create a detached HEAD, and you will no
            longer be on any branch. Are you sure you want to checkout this commit?
          </Row>
```

Sorry. I missed this the first time around.  Confirmation dialogs require a role of `alertdialog` and an `ariaDescribedBy` property so that their messages are properly announced by screen readers. (This is a recent improvement and is not implemented across the board yet, but required for new features).

tidy-dev (changes requested):
@kitswas Making great progress! One more concern that is still present for me and then I will see if I can get another team member to take a look since this is pretty core functionality. 

There is quite a bit duplicated logic between `checkoutCommit` and `checkoutBranch` in the `checkout.ts`.  It would be awesome if we could make those functions more targeted and make a method or methods to house the shared logic.  [This commit is along the lines](https://github.com/desktop/desktop/commit/b3012fac79240e31b8f7e6d83ed4976be902458e) of what I am thinking. Tho, open to other solutions to share the "checkout" logic that is not unique to "commit" or "branch". 

Additionally, I added one more comment about the confirm dialog needing to have some additional markup to make sure our screenreader users are included. :)

Precha24 (commented):
Good, thanks

app/src/lib/git/checkout.ts
```diff
@@ -66,44 +115,60 @@ export async function checkoutBranch(
...
+  progressCallback?: ProgressCallback
+): Promise<true> {
+  const title = `Checking out ${__DARWIN__ ? 'Commit' : 'commit'}`
+  const opts = await getCheckoutOpts(
     repository,
-    branch,
     account,
+    title,
+    shortenSHA(commit.sha),
+    title,
```
sergiou87: Looking at the implementation, I guess this could be `undefined`? That's probably less confusing than seeing `title` twice (it definitely made me wonder why 😳 )
```suggestion
    undefined,
```

app/src/lib/stores/app-store.ts
```diff
@@ -3847,18 +3862,66 @@ export class AppStore extends TypedBaseStore<IAppState> {
...
+    commit: CommitOneLine
+  ): Promise<Repository> {
+    const repositoryState = this.repositoryStateCache.get(repository)
+    const { branchesState } = repositoryState
+    const { tip } = branchesState
+
+    // No point in checking out the currently checked out commit.
+    if (tip.kind === TipState.Valid && tip.branch.tip.sha === commit.sha) {
+      return repository
+    }
```
sergiou87: I guess we could also check for `tip.kind === TipState.Detached` and see if `tip.currentSha === commit.sha`, right? 🤔 

```suggestion
    if (
      (tip.kind === TipState.Valid && tip.branch.tip.sha === commit.sha) ||
      (tip.kind === TipState.Detached && tip.currentSha === commit.sha)
    ) {
      return repository
    }
```

sergiou87 (changes requested):
Code looks good and it works great! I left a couple of comments, but nothing major.

I did notice that when there are changes in the working directory, checking out a commit doesn't trigger the prompt suggesting to stash the changes and continue. However, I think that could be done in a different PR 😄 

Please, take a look at the suggested changes and let me know! 💖

app/src/lib/stores/app-store.ts
tidy-dev: @kitswas This lead me to check what happens when you try to checkout the already checked out commit. Definitely, we need to add this check. But, also, we need add this check in the history log, because right now you can invoke the context menu on the first commit and the check out commit is enabled. I click it and get the confirm, then nothing happens. It should be disabled for the first commit since you cannot check that one out/it is already checked out.

tidy-dev:
@sergiou87 In regards to,

> I did notice that when there are changes in the working directory, checking out a commit doesn't trigger the prompt suggesting to stash the changes and continue. However, I think that could be done in a different PR

We discussed this in a team sync a while ago and the reason we did not push for it in this PR is because if a user checks out a commit and then makes changes and attempts to check out a commit. A stash does not make sense because there is no branch to associate it with. Thus, implementing that needed more discussion around if we want to have a different behavior for when on branch versus on a detached head. Also as you say, if wanted, it can be done in a different PR.

app/src/lib/feature-flag.ts
```diff
@@ -73,6 +73,11 @@ export function enableResetToCommit(): boolean {
   return enableDevelopmentFeatures()
 }
 
+/** Should we allow checking out a single commit? */
+export function enableCheckoutCommit(): boolean {
+  return enableDevelopmentFeatures()
```
tidy-dev:
```suggestion
  return enableBetaFeatures()
```

Go ahead and set for next beta

tidy-dev:
@kitswas Do you feel this is ready for another review?

tidy-dev (approved):
✨  Thank you for all your work on this @kitswas and sorry for how long it has taken. The code looks great and works great! 

_(Going to let @sergiou87 have another looksee just to make sure I didn't miss something before merging)_

sergiou87:
@tidy-dev you're completely right, we discussed it and I was like " 💡 yeah! that makes total sense 🤯 " and then forgot about it 😂😂😂 I'm gonna give it a final review

sergiou87 (approved):
Beautiful work @kitswas !!! Thank you for your patience and your contribution 🙇‍♂️🙇‍♂️🙇‍♂️

Issue #16119, opened 2023-02-11
Arrow navigation doesn't work in emoji select window


m3r1v3:
### The problem

When navigating with the arrows (up/down) in the emoji window, the window closes instead of scrolling through the list. Previously, this problem was not present.

### Release version

3.1.6

### Operating system

Windows 10

### Steps to reproduce the behavior

1. Type `:` symbol in Summary/Description text field for opening emoji select window
2. Press Arrow Down key on keyboard
3. The emoji window is closed (in previous versions the list would scroll)

andrewhong04:
I've been looking into this and it seems like https://github.com/desktop/desktop/pull/15641 caused this bug. A pull request will be out shortly.

steveward:
Thanks for the report @merive, this is a duplicate of https://github.com/desktop/desktop/issues/16044 so I'm going to close this issue out. You can follow along there for updates.

PR #16121, opened 2023-02-12
Emoji arrow fix

### LINKED ISSUE #16044: Cannot click on the issue list scroll-bar
### The problem

This is probably easier to show a short video than it is to explain:

https://user-images.githubusercontent.com/164476/216118271-2b640bc7-e9d6-4fa6-b717-2d7282cb6793.mp4

Here you can see that I'm typing in a commit message. The message fixes an issue, so I type in `Fix #` and it helpfully brings up the pop-up list of issues. However:

1) Trying to drag the vertical scroll bar causes the list to vanish. You can see this in the video (the first thing I try)
2) Using the up/down arrow keys instead of the mouse causes this list to vanish (the second thing I try)

The only way I can scroll the list is to hover over it and use the mouse-wheel (the 3rd attempt in the video)

However, you can see it is responding to my mouse as I hover over the scrollbar. It just vanishes the moment you click before you get a chance to drag.

### Release version

3.1.5

### Operating system

mac os

1. Write a comment message including a #
2. Try to scroll the pop-up list using either the cursor keys or the vertical scroll bar

### Log files

_No response_

### Screenshots

_No response_

### Additional context

This may happen on Windows as well, but I haven't tested it yet.

I'm using an external keyboard and mouse on the Mac - they are not made by Apple, in case that makes any difference. I don't believe it should, but hey ... Apple.

andrewhong04:
Closes https://github.com/desktop/desktop/issues/16044

## Description
Not sure if this is a clean implementation, but yea.
 - Adds `ignoreOnRowRef` to `List` React component
 - Assigns `ignoreOnRowRef` to `true` for `autocomplete-text-input`

### Screenshots
amogus

## Release notes
Fix a bug where the emoji autocomplete would be unfocused when inputting arrow up and down keys.

Notes:

### DIFF
--- a/app/src/ui/autocompletion/autocompleting-text-input.tsx
+++ b/app/src/ui/autocompletion/autocompleting-text-input.tsx
@@ -200,6 +200,7 @@ export abstract class AutocompletingTextInput<
     return (
       <div className={className} style={{ top, left, height }}>
         <List
+          ignoreOnRowRef={true}
           rowCount={items.length}
           rowHeight={RowHeight}
           selectedRows={[selectedRow]}
--- a/app/src/ui/lib/list/list.tsx
+++ b/app/src/ui/lib/list/list.tsx
@@ -44,6 +44,7 @@ export interface IRowRendererParams {
 export type ClickSource = IMouseClickSource | IKeyboardSource
 
 interface IListProps {
+  readonly ignoreOnRowRef?: boolean
   /**
    * Mandatory callback for rendering the contents of a particular
    * row. The callback is not responsible for the outer wrapper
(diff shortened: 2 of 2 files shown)


steveward:
@novialriptide thanks for submitting a PR to fix this bug! ✨ I did a quick edit to the issue body to change the referenced issue from https://github.com/desktop/desktop/issues/16119 to https://github.com/desktop/desktop/issues/16044 where we are tracking this. Thanks again for your contribution, and someone should be around to review this soon!

niik:
Hey @novialriptide, thanks so much for opening this, we really appreciate it! I took a look and while this fixes the issue I don't feel comfortable just skipping that `onRowRef` method entirely. It has the potential of leading to more subtle bugs in the future as we add more functionality there. I opened #16219 to address this by not stealing focus when updating the `scrollToRow` prop which is only used by the `AutoCompletingTextInput` component. Please take a look and see what you think of that change. I'm going to close this PR in favor of #16219 so let's continue the discussion over there.

PR #16125, opened 2023-02-13
Focus on first suitable child in sign in flow

niik:
Spotted this while looking at https://github.com/desktop/desktop/pull/16110

When switching steps in the sign in flow we're essentially showing a new dialog (or at least it should behave as such) which means we'll need to reset the focus, if we don't then keyboard focus will just flow to the body whenever we change the dialog contents.

This is most noticeable in the GHES sign in flow where you first enter an endpoint, hit enter and then (before this change) you can't hit enter to open the browser.


tidy-dev (approved):
✨ Thanks for that fix!

PR #16151, opened 2023-02-17
Fix prod build crash due to propTypes not being defined

niik:
## Description

react-virtualized strips propTypes for prod builds so we're gonna have to be more careful here

Regression introduced in #16108

### Screenshots

image

## Release notes

Notes:


tidy-dev (approved):
✨ That fixed it! Thank you!

PR #16153, opened 2023-02-17
[Linux] update `electron-builder` to latest release and refresh configuration

shiftkey:
## Description

The `electron-builder` dependency is something that I have familiarity with, and while I've made much more changes in the fork I think bringing this up to date is a low-hanging change.

In terms of the config changes, I've removed the `snap` package because this requires a new [`multipass` package](https://multipass.run/) which also takes a long amount of time to create (compared to the others) and I've had some issues with supporting: https://github.com/shiftkey/desktop/issues/202

## Release notes

Notes: no-notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
     "@types/webpack-merge": "^5.0.0",
     "@types/xml2js": "^0.4.0",
     "electron": "22.0.3",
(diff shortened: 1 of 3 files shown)


niik (approved):
👍 Thanks!

PR #16018, opened 2023-01-27
Automatically fetch repositories cloned through the terminal. (Issue #14518)

Bavuett:
_**This PR is currenly a work in progress! This description will be changed as I work on the feature.**_

Closes #14518 

## Description

- On startup, check for a folder where the user stores his repositories and add them to GitHub Desktop.

## Release notes

Notes:

PR #16213, opened 2023-02-25
Add multiple repositories from parent folder.

### LINKED ISSUE #14518: Automatically fetch repositories cloned through the terminal.
### The feature request

I love GitHub Desktop, but sometimes I use the terminal to clone repositories because it's faster.

It would be great if GitHub Desktop would periodically check a known folder for newly cloned repositories and would automatically add them to the list, so people could manage them using this client without manually adding them everytime.

### Proposed solution

On startup, check for a folder where the user stores his repositories and add them to GitHub Desktop.

I have no experience with React so I can't work myself on the feature, but it would be awesome if someone worked on this.

### Additional context

I have no mockups, unfortunately.

Bavuett:
Closes #14518 (even though it's already closed, lol).

## Description

- Allow users to add to Desktop multiple repositories contained in a parent folder;
- Allow users to set a default parent directory in case they frequently add multiple repositories (for example, they clone repos through the terminal but still want to use Desktop from time to time).

This is just a draft of the feature. I would've liked to edit the classic "Add Existing Repository" feature because I think it would be better in terms of UX, but because of the discussions we had in the issues #14518 and #16018 (and since it would require me to edit a core functionality of the product) I thought it would be better to ask for your feedback first. 

As you surely have noticed, the scanning is not done on startup anymore: I took into account your considerations about performance. 

### Screenshots

https://user-images.githubusercontent.com/43916566/221365130-2929efa9-69b4-41f4-8fd7-54e4a127ae18.mp4


tidy-dev:
@bavuett Thank you for wanting to work on this and sorry it has taken a while to get back to you on it. We discussed this in a team meeting today and are looking at the issue two ways.

1. We have users who want the ability to sync to a folder. As mentioned before, this has performance caveat and would require the UI to handle the various errors, ability to choose, skip the repository found which comes with maintenance of that UI/UX. Since this is such a niche use case, we do not believe the benefits outweigh the cons. I think part of your solution is trying to keep that use case still in mind "Allow users to set a default parent directory in case they frequently add multiple repositories (for example, they clone repos through the terminal but still want to use Desktop from time to time)." Adding a setting for this is not something we would support (not to say we couldn't remember the last entry). 
2. We have users who have lots of repositories. They want a convenient way to add them all at once (our drag/drop feature allows for this but is not discoverable). However, we still feel this is a niche area and don't have a strong impetus to include it. 

If we were going to consider something in support use case #2, we think UI/UX needs a little more consideration.

Some concerns that arose from our discussion:
- "Add existing repositories from folder" may not differentiate it enough or explain the feature well enough. 
- The users who want this feature or more likely to be "power users" with multiple repositories and may have multiple hierarchical organization to their repositories. For example, Repositories > Work  and Repositories > Personal. The feature would need to be more an "Scan" feature that looks deeply for repositories, presents them to the user, allows the user to accept the repositories they want.

Lastly, if you would like to continue work on this, before opening anymore pull requests, please pitch your solution in the [associated issue](https://github.com/desktop/desktop/issues/14518). Words are great, but mocks up are even better. If the Desktop team agrees it is something we can support, then opening a PR is great. :)

PR #16219, opened 2023-02-27
Don't move focus when scrolling to row

### LINKED ISSUE #16044: Cannot click on the issue list scroll-bar

niik:
Closes #16044
Superseedes https://github.com/desktop/desktop/pull/16121

## Description

In https://github.com/desktop/desktop/pull/15641 we made the [default behavior of the scrollRowToVisible](https://github.com/desktop/desktop/pull/15641/files#diff-73d623888d5c34b2759c7e9d3fbb3c4f6820c1f6dd61b8496ec016117e41e8feR741) method be that it also moved focus. This works well when using keyboard navigation within a list but less so when dealing with a list controlled from a different component (as it the case with the `AutoCompletingTextInput`).

This changes the behavior when setting the `scrollToRow` prop (which is only used by the AutoCompletingTextInput component) to not steal focus.

## Release notes

Notes: [Fixed] Up/down arrow can be used to navigate autocomplete lists like emoji again

### DIFF
--- a/app/src/ui/lib/list/list.tsx
+++ b/app/src/ui/lib/list/list.tsx
@@ -813,7 +813,7 @@ export class List extends React.Component<IListProps, IListState> {
     if (scrollToRow !== undefined && prevProps.scrollToRow !== scrollToRow) {
       // Prefer scrollTop position over scrollToRow
       if (setScrollTop === undefined) {
-        this.scrollRowToVisible(scrollToRow)
+        this.scrollRowToVisible(scrollToRow, false)
       }
     }
 


andrewhong04:
Looks better than my bug fix.

sergiou87 (approved):
Code looks good and works as expected! 🙌

PR #16238, opened 2023-03-02
Fix Preview Pull Request Large Diff Image

tidy-dev:
## Description
This PR moves some duplicated CSS that is applied to the history diff and the changes diff.. that we also want in the pull request diff and likely will want it if we used the diff anywhere else in the future. Thus, instead of just duplicating again, I moved it into the _diff.scss

### Screenshots

Preview Pull Request
image

Changes
image

History
image

## Release notes
Notes: [Fixed] Large diff messaging styled consistently in pull request preview.

### DIFF
--- a/app/styles/ui/_commit-details.scss
+++ b/app/styles/ui/_commit-details.scss
@@ -11,25 +11,4 @@
     // look right alongside blank diff
     border-right: var(--base-border);
   }
-
-  .panel {
-    display: flex;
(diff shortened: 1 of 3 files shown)


sergiou87 (approved):
Oh! This is much better 😌 It's always nice to reuse code, and works as expected in changes, history and PR preview views!!

:shipit: ❗

PR #16240, opened 2023-03-02
Screen Reader Available Email label fields associated misattributed commits emails

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3280
## Description
This PR adds an `aria-label` to the other email input box in the `Git` preferences/options. This adds a "Email" label to the select input of the misattribute popover.

### Screenshots
macOS

https://user-images.githubusercontent.com/75402236/222504352-ea0de5ad-a948-4752-a1e4-d241b332e5cd.mp4

Windows:

https://user-images.githubusercontent.com/75402236/223109896-8daba977-aecb-4c2f-9594-cd436adff538.mp4

## Release notes
Notes: [Improved] Other email input in "Git" preferences/Options and misattributed popover email select have a screen readable label.

### DIFF
--- a/app/styles/ui/_commit-message-avatar.scss
+++ b/app/styles/ui/_commit-message-avatar.scss
@@ -38,6 +38,7 @@
       height: 10px;
       // With width=100%, the icon will be centered horizontally
       width: 100%;
+      vertical-align: unset;
     }
   }
 
--- a/app/src/ui/changes/commit-message-avatar.tsx
+++ b/app/src/ui/changes/commit-message-avatar.tsx
@@ -155,7 +155,7 @@ export class CommitMessageAvatar extends React.Component<
         </Row>
         <Row>
           <Select
-            label="Email"
+            label="Your Account Emails"
             value={this.state.accountEmail}
             onChange={this.onSelectedGitHubEmailChange}
           >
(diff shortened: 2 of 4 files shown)


app/styles/ui/_commit-message-avatar.scss
```diff
@@ -38,6 +38,7 @@
       height: 10px;
       // With width=100%, the icon will be centered horizontally
       width: 100%;
+      vertical-align: unset;
```
tidy-dev, line 41: Added this after taking windows screencast and seeing the warning icon out of place.

Before
CleanShot 2023-03-06 at 07 29 08

After: 
CleanShot 2023-03-06 at 07 29 30

sergiou87 (approved):
Code makes sense and works as expected! I'm surprised about the icon on Windows, I wonder if that broke on a recent (or not-so-recent) change 🤔

tidy-dev:
> I wonder if that broke on a recent (or not-so-recent) change 🤔

I am guessing it was on my recent PR that made the avatar a button when misattributed commit warning is present since it was `.button-component` css that was causing the problem.... but I thought I tested that on both windows and mac.. I must not have caught the regression then.

Kiaramary099 (commented):
#`

`

PR #16270, opened 2023-03-08
Announce popover header

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3271

## Description
This PR adds the role of dialog to and adds an optional `aria-labelledby` property to our popover component. Then utilizes the `aria-labelledby` for the misattributed commit warning popover.

Edit: Also adds the `aria-labelledby` to the other 4 popover use cases.

### Screenshots
MacOS Voiceover:
https://user-images.githubusercontent.com/75402236/223744979-51528218-3f5f-4949-b727-fc9b81367e55.mp4

Windows NVDA:
https://user-images.githubusercontent.com/75402236/223763836-077fbd5f-d057-4972-b025-9a2ada592171.mp4

## Release notes
Notes: [Improved] Popover titles are announced by screen readers.

### DIFF
--- a/app/src/ui/diff/diff-options.tsx
+++ b/app/src/ui/diff/diff-options.tsx
@@ -91,6 +91,7 @@ export class DiffOptions extends React.Component<
         caretPosition={PopoverCaretPosition.TopRight}
         onClickOutside={this.closePopover}
       >
+        <h3>Diff {__DARWIN__ ? 'Preferences' : 'Options'}</h3>
         {this.renderHideWhitespaceChanges()}
         {this.renderShowSideBySide()}
       </Popover>
@@ -107,7 +108,7 @@ export class DiffOptions extends React.Component<
   private renderShowSideBySide() {
     return (
       <section>
-        <h3>Diff display</h3>
+        <h4>Diff display</h4>
         <RadioButton
           value="Unified"
           checked={!this.props.showSideBySideDiff}
@@ -131,7 +132,7 @@ export class DiffOptions extends React.Component<
   private renderHideWhitespaceChanges() {
     return (
       <section>
-        <h3>Whitespace</h3>
+        <h4>Whitespace</h4>
         <Checkbox
           value={
             this.props.hideWhitespaceChanges
--- a/app/src/ui/diff/diff-options.tsx
+++ b/app/src/ui/diff/diff-options.tsx
@@ -88,10 +88,13 @@ export class DiffOptions extends React.Component<
   private renderPopover() {
     return (
       <Popover
+        ariaLabelledby="diff-options-popover-header"
         caretPosition={PopoverCaretPosition.TopRight}
         onClickOutside={this.closePopover}
       >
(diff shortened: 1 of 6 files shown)


app/src/ui/lib/popover.tsx
```diff
@@ -105,7 +106,13 @@ export class Popover extends React.Component<IPopoverProps> {
 
     return (
       <FocusTrap active={true} focusTrapOptions={this.focusTrapOptions}>
-        <div className={cn} ref={this.containerDivRef} style={this.props.style}>
+        <div
+          className={cn}
+          ref={this.containerDivRef}
+          style={this.props.style}
+          aria-labelledby={this.props.ariaLabelledby}
+          role="dialog"
```
tidy-dev, line 114: Impact notes for adding the `role="dialog"`:
I tested a few other existing use cases for the popover component including:
- popover dropdown component used in pull request preview
- diff settings
- check runs popover

I did not see any regression to using the rule and on the contrary it now announces the first heading on each of these. This works well for the check runs popover and popover dropdown component that have natural heading. On the diff settings, we probably need to either add a true heading "Diff Settings" or have an invisible one that the screen reader reads as right now it announces the first heading of "Whitespace".

Other thing to note is now the popovers are announced as  "dialog" when a popover is opened, but I believe that is correct behavior to inform a screenreader users of a "dialog like" thing on the screen.

app/src/ui/diff/diff-options.tsx
```diff
@@ -91,6 +91,7 @@ export class DiffOptions extends React.Component<
         caretPosition={PopoverCaretPosition.TopRight}
         onClickOutside={this.closePopover}
       >
+        <h3>Diff {__DARWIN__ ? 'Preferences' : 'Options'}</h3>
```
tidy-dev: CleanShot 2023-03-08 at 10 02 32

Added after making [this comment](https://github.com/desktop/desktop/pull/16270/files#r1129567830)

app/src/ui/lib/popover.tsx
tidy-dev: > on the diff settings, we probably need to either add a true heading "Diff Settings" or have an invisible one that the screen reader reads as right now it announces the first heading of "Whitespace".

Decided to [go ahead and add this;](https://github.com/desktop/desktop/pull/16270#discussion_r1129583497) totally open to discussion if any feels having a heading is overkill.

tidy-dev: Other notes -> Tho the popovers with `role` of `dialog` do read the first available header in macOS voice over. They do not on windows NVDA; thus, using the `aria-labelledby` will likely be a requirement for all our popovers.

tidy-dev: Since there were only 5 occurrences of popover, I went ahead and addressed it in this PR.
CleanShot 2023-03-08 at 11 17 56

niik (approved):
:sparkles:

PR #16272, opened 2023-03-08
Tooltip for Committing Message Avatar is a toggle tip

tidy-dev:
Now based on: https://github.com/desktop/desktop/pull/16449

xRef: https://github.com/github/accessibility-audits/issues/3012

## Description
This PR makes the avatar into a toggle tip. User can tab to it and toggle the button to see and screen reader announce the tooltip.

https://user-images.githubusercontent.com/75402236/230361020-a0fdd69a-9354-4358-a012-aa81b17e14e6.mp4

## Release notes
Notes: [Improved] The commit message avatar is now a toggle tip so that the commit author details are keyboard accessible.


tidy-dev:
Discussed in team sync, we decided to try turning the avatar into a button that opens the users preferences for changing their email. On focus, the tooltip will open and be announced.

tidy-dev:
> Discussed in team sync, we decided to try turning the avatar into a button that opens the users preferences for changing their email. On focus, the tooltip will open and be announced.

After doing the toggle tip work,  it seemed a simpler solution than making a button to open users preferences when we don't really have a reason to suspect users will want or expect a button there that does that.

app/src/ui/changes/commit-message-avatar.tsx
```diff
@@ -150,7 +176,7 @@ export class CommitMessageAvatar extends React.Component<
           <div>
             The email in your global Git config (
             <span className="git-email">{this.props.email}</span>) doesn't match
-            your GitHub{accountTypeSuffix} account.{' '}
+            your GitHub{accountTypeSuffix} account{userName}.{' '}
```
tidy-dev, line 179: Since the avatar no longer has a tooltip when there is a warning and therefore the user cannot see the GitHub user they are logged in as, I thought adding the username makes sense in this context.  (Tho open to opinions since technically users should only have one GitHub account.. but I think the reason this warning is here is because they are maybe using a personal and a work account :/) 

CleanShot 2023-04-06 at 07 09 55

niik (approved):

PR #16274, opened 2023-03-08
Commit attribution descriptive links

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3270, https://github.com/github/accessibility-audits/issues/3264

## Description
This updates all our "Learn more" links to be more descriptive for screen reader users.

### Screenshots
CleanShot 2023-03-08 at 14 21 24
CleanShot 2023-03-08 at 14 20 44
CleanShot 2023-03-08 at 14 28 20
CleanShot 2023-03-08 at 14 23 44
CleanShot 2023-03-08 at 14 22 26

## Release notes
Notes: [Improved] Learn more links are descriptive for screen readers.

### DIFF
--- a/app/src/ui/changes/commit-message-avatar.tsx
+++ b/app/src/ui/changes/commit-message-avatar.tsx
@@ -149,7 +149,7 @@ export class CommitMessageAvatar extends React.Component<
             <span className="git-email">{this.props.email}</span>) doesn't match
             your GitHub{accountTypeSuffix} account.{' '}
             <LinkButton uri="https://docs.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user">
-              Learn more.
+              Learn more about commit attribution.
             </LinkButton>
           </div>
         </Row>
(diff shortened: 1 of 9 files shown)


sergiou87 (commented):
Could this be the kind of thing where we leave the `Learn more` text for visual users and use the `aria-label` or  `aria-description` for non-visual users? It feels like they now look super verbose and redundant in the GUI.

tidy-dev:
The general guidance is to go for more descriptive first. Imho, that is generally just good practice for usability, but also accessibility. However, if size constraints are of concern then [using `aria-label` is an acceptable alternative](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8).

In these cases, I can see in the Misattributed warnings since we are trying to make them be less obtrusive to use them there. But in the repository settings, we already had the full sentence on one of them, all I did was extend the link tags. Thus, feel like keeping them where we have larger real estate makes sense.

Another alternative for the misattributed commits warning was to do like: 
> This email doesn't match your GitHub Account, so your commits will be `<link>`wrongly attributed`</link>.`

^ only concern there was that the "learn more" make it clear that you are likely going to some documentation.

Thoughts?

sergiou87 (approved):
I prefer using the `Learn more` verbiage, and the last changes you made look good to me!! :shipit:

PR #16286, opened 2023-03-09
Stream repositories available to the user from the API

### LINKED ISSUE #2770: cache user repositories locally
See https://github.com/desktop/desktop/pull/2436#issuecomment-326069611

We don't cache repositories retrieved from the API, so it's really slow on each usage (not just the first) when you have access to hundreds of repositories.

Might also need some design thought about managing the interactions (show cached repositories, while fetching new repositories to refresh, and then merging) so that the app refreshes without interrupting the user. 

cc @desktop/design for feels

### LINKED ISSUE #8955: Loading repositories is very slow

Opening a repository from GitHub is slow if the user has access to many repositories.

### Proposed solution

By caching the repositories locally, the user would quickly be able to open all but the most recently added repositories.

### Additional context

UI should indicate that it's presenting a cached view, allowing the user to wait for a refresh if the repository is not available in cache.

### LINKED ISSUE #16276: It is fetching all repositories in the organization and this is making it very slow
### The feature request

Hello, I recently tried using Github Desktop in my work but one of the most interesting features is being able to fetch my repositories from within it is impossible to use, I really didn't know whether to report it as a bug or a feature, so I leave it this question here too.

The problem is the following, it is trying to pull all the repositories of all the organizations that I am part of within Github, this is a problem because one of them has more than 140k of repositories, so the fetch takes about 1 hour to pull them all , which makes the application impossible to use in my case due to the delay in waiting for the fetch.

### Proposed solution

My suggestion: add a filter functionality, so that I can fetch only my personal repositories, since the filter that exists in the application requires that all repositories have been fetched first.

### Additional context

I participate in an organization with many repositories

niik:
Closes #16276
Closes #8955
Closes #2770

## Description

Improved the user experience in the clone repositories dialog for users with access to lots and lots of repositories.

Prior to this change we waited until we had loaded all repositories from the `/user/repos` endpoint (could be hundreds of pages long for some users) before we showed them in the UI. With this PR I'm intending to improve that experience for users with tons of repos while not making it any worse for the vast majority of users who have access to very few repositories.

The way I went about this was by showing the repositories in the UI _as each page from the API is coming in_ and by making multiple concurrent streaming requests in order to prioritize the users own repositories over repositories for organizations that they belong to (which is typically but not always the culprit).

Prior to this change it took me almost 5 minutes of waiting on the clone repository dialog before I could see any repositories and after this it's a matter of seconds until I at least see my own repositories. The overall load time is also improved _some_ by us loading owner, collaborator, and org affiliated repositories in three different concurrent streams.

Beyond those changes I noticed that we count any repository not owned by an organization as being "Your repository" in the filter list, that's been fixed to only include repository actually owned by the user. Additionally I've made it so that refreshing the list of repositories doesn't throw away all the repositories we've got access to.

#### Possible further improvements

I think an additional improvement that could be useful here is to add another fetch when refreshing repositories that only grab repositories created after the last fetch (i.e. `MAX(existing_repositories::created_at)`). That would let users who've just created a repository online be able to see it almost instantaneously after hitting refresh.

Showing a message when the user is searching but not finding what they're looking for would be nice (we're still loading repositories, your search results may be incomplete).

We could/should consider redesigning the clone screen to only show the users owned repositories by default. Ideally we'd let the user pick which org to load but the API support for that doesn't exists AFAICT.

For users with tons of repos there's a lot of overhead data from the API here that we don't care about. From a bandwidth conservation perspective GraphQL could help here.

## Release notes

Notes:

### DIFF
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -922,7 +922,10 @@ export class API {
         },
       })
     } catch (error) {
-      log.warn(`fetchRepositoriesByPage failed`, error)
+      log.warn(
+        `streamUserRepositories: failed with endpoint ${this.endpoint}`,
+        error
+      )
     }
   }
 
--- a/app/src/lib/api.ts
+++ b/app/src/lib/api.ts
@@ -26,6 +26,15 @@ if (envAdditionalCookies !== undefined) {
   document.cookie += '; ' + envAdditionalCookies
 }
 
+type AffiliationFilter =
+  | 'owner'
+  | 'collaborator'
+  | 'organization_member'
+  | 'owner,collabor'
+  | 'owner,organization_member'
+  | 'collaborator,organization_member'
+  | 'owner,collaborator,organization_member'
+
 /**
  * Optional set of configurable settings for the fetchAll method
  */
@@ -47,8 +56,9 @@ interface IFetchAllOptions<T> {
    * `(results) => results.length < 500`
    *
    * @param results  All results retrieved thus far
+   * @param page     The last fetched page of results
    */
-  continue?: (results: ReadonlyArray<T>) => boolean
+  continue?: (results: ReadonlyArray<T>, page: ReadonlyArray<T>) => boolean
 
   /**
    * Calculate the next page path given the response.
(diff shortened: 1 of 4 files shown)


app/src/lib/api.ts
```diff
@@ -47,8 +56,17 @@ interface IFetchAllOptions<T> {
    * `(results) => results.length < 500`
    *
    * @param results  All results retrieved thus far
+   * @param page     The last fetched page of results
    */
```
sergiou87: I think this line can be removed
```suggestion
   */
```

sergiou87 (dismissed):
Code makes sense and works **great**!!! ⚡ 

I only found one comment line that isn't needed anymore. I'll approve the PR anyway and leave it up to you whether you remove it here or not.

niik:
Gentle reminder here @sergiou87, I think this is all reviewed, I just applied that suggestion of yours to remove the comment

sergiou87 (approved):
Yup! Sorry, I forgot about that 😳

PR #16288, opened 2023-03-09
Add offset focus ring for buttons, vertical tabs etc

niik:
xref: https://github.com/github/accessibility-audits/issues/3288

## Description

This addresses an accessibility issue where the focus ring (or lack of really) for our primary buttons doesn't meet the required contrast. Prior to this PR we only tweaked the border color slightly when a button received keyboard focus. In order to address this I'm changing buttons as well as vertical tabs to use offset focus rings and I expect that we'll find more places to leverage this pattern going forward.

### Screenshots

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image | 

## Release notes

Notes:

### DIFF
--- a/app/styles/_variables.scss
+++ b/app/styles/_variables.scss
@@ -338,12 +338,6 @@ $overlay-background-color: rgba(0, 0, 0, 0.4);
   --focus-color: #{$blue};
 
   /**
-   * An alternative focus color to use when the background color of
-   * the focused element is the same as the regular focus color.
-   */
-  --alt-focus-color: #{$blue-800};
-
-  /**
    * Variables for form elements
    */
   --text-field-height: 25px;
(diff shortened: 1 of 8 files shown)


sergiou87 (commented):
Looks good overall! I only noticed something annoying in the commit button when its text is too long and is truncated:

Kapture 2023-03-10 at 11 21 39

As you can see the text moves slightly. It's a minor thing that didn't happen before this PR. I'm happy to approve it as it is, let me know what you want to do about it! 😄

niik:
> As you can see the text moves slightly. It's a minor thing that didn't happen before this PR. I'm happy to approve it as it is, let me know what you want to do about it! 😄

Great spot! 😍 I thought removing the border was safe but it wasn't and we can't remove it entirely because we want the primary and secondary button styles to have the same internal dimensions so that text line up when placed next to each other. I sorted that out and fixed the focus color on vertical tabs and added a little extra spacing to make things look better.

sergiou87 (approved):
Thanks for those changes! It works perfectly now 🙌

PR #16311, opened 2023-03-13
[wip] Trying to make windows app menu's announce

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3250

## Description
This is results of lots of poking and prodding in order to get a semi functional solution to getting the app menu items on windows to 1) announce the main menu label 2) announce the menu items.

Number 2 is really easy, the tab index simply needs to update to 0 when a menu item is selected.

Number 1 is supposed to be as easy. :D From what I can tell, it doesn't work because we immediately shift the focus to the menu instead of keeping the focus in the menu bar until a user hits a down arrow, space, or enter to open the menu. I made a work around where I just put an `aria-label` on the menu. This works tho not the generally recommended approach. Should we refactor to recommended approach or keep work around that seems to meet criteria?

Other notes.. I replace the use of the `ToolbarDropdown` and `ToolbarDropdownButton` with new components that strips away all the logic that is not applicable to the app windows (used for the repository and branch dropdowns) just so I could easily digest what was impactful to the use case I am analyzing. Wondering what others thoughts are on keeping is separate? They seem to me to be enough of distinct use cases with a lot of of "ifs" added to accomodate the multiple scenarios.. feels like it the current components could easily lead to fix one thing to break another.  The windows app menu only needs a much more trimmed down version (probably even more so than what I currently have). 

One more thing I haven't figured out, when you use the access keys to open a menu, it automatically focuses the first menu item but does not announce it. 

## Release notes
Notes: tbd

### DIFF
--- a/app/src/ui/app-menu/app-menu-bar-button.tsx
+++ b/app/src/ui/app-menu/app-menu-bar-button.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react'
 import { IMenu, ISubmenuItem } from '../../models/app-menu'
-import { MenuListItem } from './menu-list-item'
 import { AppMenu, CloseSource } from './app-menu'
-import { ToolbarDropdown } from '../toolbar'
 import { Dispatcher } from '../dispatcher'
+import { AppMenuDropdown } from './app-menu-dropdown'
+import { MenuListItem } from './menu-list-item'
 
 interface IAppMenuBarButtonProps {
   /**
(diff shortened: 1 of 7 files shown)


tidy-dev:
Talked with Accessibility folks and I think the goal of the actually existing behavior is acting correctly for Windows. I just misunderstood with looking at macOs behavior. Going to take another stab at this.

PR #16313, opened 2023-03-14
Preference dialog, now 99% less jarring

niik:
## Description

Our preference dialog has been far from ideal for a long time with the footer being visible on some tabs and not other and the dialog itself changing size when switching tabs. This wasn't always the case but it's become worse and worse as we've added things onto the dialog.

In this PR I'm making some changes to avoid the dialog changing size by

1. Creating a new section just for notifications. This might seem like overkill but it felt like the best option to pull out of advanced and I'm justifying giving it its own section so that users can easily find and turn it off. Plus it's not really advanced from a users perspective (tho that goes for more things in the advanced section)
2. Always showing the footer with the save and cancel buttons. In order to support this I've made it so that the theme selection is still live when switching in the preference tab but it reverts to the original if the user hits cancel. I haven't done the same for the accounts and I don't think that it's worth the effort nor do I think anyone will be surprised given that the sign in flow takes users out of the dialog which leaves only the logout flow.

## Release notes

Notes:


sergiou87:
I was thinking about this today, you read my mind! 🧠 👀 🤣 

I think Notifications could deserve their own section, and we could potentially allow users to disable specific ones (like PR reviews, CI failed or PR comments… or all of them like we do now). But for now I saw this one that might make more sense in the Prompts section:
2023-03-14 at 12 34

What do you think?

niik:
> What do you think?

I think that's an excellent idea!

image

sergiou87 (approved):
This is so much better!! :shipit:

PR #16315, opened 2023-03-14
Windows app menu screen reader fixes

tidy-dev:
Closes [github/accessibility-audits#3250](https://github.com/github/accessibility-audits/issues/3250)

## Description
1) This adds the `tabindex="-1"` to the dropdown menu items so they are announced screen readers.
2) This adds the `role="menuitem"` to the menu bar buttons so they are announced by screen readers when using `alt` and left and right arrows. 
3) This focuses the first menu item when "Enter" or "Space" is used to open the menu.
4) Removes focus of menu pane that steals focus from first focused menu item. 
-  This last bullet may need closer looking at, maybe an additional `if` condition that focuses only if there is not an menu item selected.. which I am not sure when is.. but if that case were to be, then this would ensure the user could navigate the menu items with arrow keys. But, before I investigate further, I want to verify with accessibility team that what this is doing now is expected behavior. 

### Screenshots

https://user-images.githubusercontent.com/75402236/225020390-dcd52f2c-4f57-4adf-bb57-87fae2b44d8d.mp4

## Release notes
Notes: [Fixed] On Windows, app level menu bar and menu items are announced by screen readers.

### DIFF
--- a/app/src/ui/app-menu/app-menu.tsx
+++ b/app/src/ui/app-menu/app-menu.tsx
@@ -333,7 +333,6 @@ export class AppMenu extends React.Component<IAppMenuProps, {}> {
     if (this.focusPane >= 0) {
       const pane = this.paneRefs[this.focusPane]
       if (pane) {
-        pane.focus()
         this.focusPane = -1
       }
     }
--- a/app/src/ui/app-menu/app-menu.tsx
+++ b/app/src/ui/app-menu/app-menu.tsx
@@ -89,51 +89,12 @@ function menuPaneClassNameFromId(id: string) {
 
 export class AppMenu extends React.Component<IAppMenuProps, {}> {
   /**
-   * The index of the menu pane that should receive focus after the
-   * next render. Default value is -1. This field is cleared after
-   * each successful focus operation.
-   */
-  private focusPane: number = -1
-
-  /**
-   * A mapping between pane index (depth) and actual MenuPane instances.
-   * This is used in order to (indirectly) call the focus method on the
-   * underlying List instances.
-   *
-   * See focusPane and ensurePaneFocus
-   */
-  private paneRefs: MenuPane[] = []
-
-  /**
    * A numeric reference to a setTimeout timer id which is used for
    * opening and closing submenus after a delay.
    *
    * See scheduleExpand and scheduleCollapse
    */
   private expandCollapseTimer: number | null = null
-
-  public constructor(props: IAppMenuProps) {
-    super(props)
-    this.focusPane = props.state.length - 1
-    this.receiveProps(null, props)
-  }
-
-  private receiveProps(
-    currentProps: IAppMenuProps | null,
-    nextProps: IAppMenuProps
-  ) {
-    if (nextProps.openedWithAccessKey) {
-      // We only want to react to the openedWithAccessKey prop once, either
-      // when it goes from false to true or when we receive it as our first
-      // prop. By doing it this way we save ourselves having to go through
-      // the dispatcher and updating the value once we've received it.
-      if (!currentProps || !currentProps.openedWithAccessKey) {
-        // Since we were opened with an access key we auto set focus to the
-        // last pane opened.
-        this.focusPane = nextProps.state.length - 1
-      }
-    }
-  }
 
   private onItemClicked = (
     depth: number,
@@ -157,9 +118,6 @@ export class AppMenu extends React.Component<IAppMenuProps, {}> {
       this.props.dispatcher.setAppMenuState(menu =>
         menu.withOpenedMenu(item, sourceIsAccessKey)
       )
-      if (source.kind === 'keyboard') {
-        this.focusPane = depth + 1
-      }
     } else if (item.type !== 'separator') {
       // Send the close event before actually executing the item so that
       // the menu can restore focus to the previously selected element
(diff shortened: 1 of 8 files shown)


app/src/ui/app-menu/app-menu.tsx
```diff
@@ -333,7 +333,6 @@ export class AppMenu extends React.Component<IAppMenuProps, {}> {
     if (this.focusPane >= 0) {
       const pane = this.paneRefs[this.focusPane]
       if (pane) {
-        pane.focus()
```
tidy-dev, line 336: @niik Do you remember context around this pane focusing? Currently, it is problematic and removing fixed my issue of stealing focus from the first selected menu item and therefore not announcing the first selected menu item.. but concerned it may break another scenario I am not testing. 

The problem I did encounter by removing this was that if a user hits `alt` and then opens a menu with `enter` , we did not select the first item and therefore we would need this to focus the pane so that up/down arrows would navigate it. I addressed this by doing what I think is the expectation (for windows) and focusing the first item on menu open. (Confirming with accessibility)

tidy-dev:
Feedback on this PR from James Scholes:
> Okay, I tried the test build.  The menu items are now wreading out as expected, which is great.  Some comments about potential improvements:
1. You had indicated that the individual menu containers were being given a name with aria-label, and I suggested aria-labelledby instead pointing to the ID of the corresponding menu bar item.  Neither seems to be in place in this test build.  While NVDA won't report the menu name when it opens, other screen readers (e.g. JAWS) will. And so all menus must have a name.
2. The menu bar items aren't indicated as having an associated pop-up menu, although I could swear they used to be in the beta build.  Either way, they need aria-haspopup="menu".
3. The entire menu bar can be accessed by a screen reader user at all times, which isn't standard behaviour in Windows apps.  It should be hidden for all users unless a user actually tries to reach it with mouse or keyboard.
4. Once focus is in the menu bar, Tab and Shift+Tab move through the menu bar items similar to how Left/Right Arrow do.  This is also non-standard on Windows; pressing Tab from a menu bar should hide it and place focus where it was before.
Hope these are helpful!


I have implemented 1 and 2 on this PR. I think 3 and 4 are fallouts of this not being a standard windows menu but a custom menu, and are not accessibility violations (if anything they just make the feature useable in more ways tho maybe unexpected -> if we get additional feedback that this should change we can followup.) Since it is reading with screen readers as expected, I think the focus logic is not needed now that on access key we select the first item so I have removed it. <- happy to change that if I am missing something.

app/src/ui/app-menu/app-menu.tsx
tidy-dev: @niik Continued this conversation in [added comment](https://github.com/desktop/desktop/pull/16315#issuecomment-1472318476).

app/src/ui/lib/aria-types.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export type AriaHasPopupType =
+  | boolean
+  | AriaHasPopup.Dialog
+  | AriaHasPopup.Menu
+  | AriaHasPopup.Listbox
+  | AriaHasPopup.Tree
+  | AriaHasPopup.Grid
+  | undefined
```
sergiou87: Another way to express this could be…
```suggestion
import { HTMLAttributes } from 'react'

export type AriaHasPopupType = HTMLAttributes<any>['aria-haspopup']
```

sergiou87 (commented):
Code makes sense to me, I just found something that can be simplified if you want. Not a fan of the `any` I had to add, but I don't know another way of doing that 😆 

I can test it next week!

app/src/ui/lib/aria-types.ts
```diff
@@ -0,0 +1,5 @@
+import { HTMLAttributes } from 'react'
+
+export type AriaHasPopupType = HTMLAttributes<
+  string | boolean | undefined
```
sergiou87: While it seems to be irrelevant, I don't think this is correct 🤔 AFAIK `HTMLAttributes` expects the type to be an Element (if you dive into `DomAttributes` you will see that).

Anyway, just a very minor thing.

tidy-dev: K. I went googling around for examples of using this. It seems like `HTMLAttributes<HTMLElement>` might make sense. What do you think?

sergiou87 (approved):
For some reason I have trouble making it work the first time I press <kbd>Alt</kbd> and trying to move around with my keyboard:

https://user-images.githubusercontent.com/1083228/226945568-a0c46cf7-3371-499c-a321-9e65980e1cde.mov

Once I opened one and press <kbd>Escape</kbd>, it works as expected. Not sure if you want to address this in this PR or prefer another one, so I'm gonna approve this in case you want to go ahead and fix that later.

Unrelated to this PR, but I noticed that extra padding menu panes have on the right side, also present in `development` 🤯

tidy-dev:
Yeah. I will look into this in a separate PR. ... and I know exactly why the extra padding is there.. easy fix for silly mistake 🤦 (overflow-y should be auto not scroll)

tidy-dev:
> For some reason I have trouble making it work the first time I press Alt and trying to move around with my keyboard:

@sergiou87 
I have not been able to reproduce this... Maybe there is some additional step.
I am:
1. Opening beta
2. Hit alt. I can:
    - left/right navigate to any menu bar item and they are announced
    - I can hit down arrow, space, or enter to open a menu and have first item focused
3. I can hit alt+access key and it opens the menu with first item focused.

Issue #10622, opened 2020-09-22
Keyboard Shortcuts for 


Lalo611:
### Describe the feature or problem you’d like to solve

By keayboard language limitations I can't press [Ctrl + ` ] to open the command prompt cause I need to press [Altgr] in order to the keyboard to read the key and that apparently breaks the shortcut

### Proposed solution

Let the users select in GitHub Desktop Settings [Shortcuts Conf 1], [Shortcuts Conf 2] and so on, so that we Latin Americans could use the shortcuts.

Or to fix the shortcut so that it doesn't break by pressing [Altgr].

### Additional context

![image](https://user-images.githubusercontent.com/69506035/93899130-64071300-fcca-11ea-99ef-6a233483421e.png)
*Note: I couldn't write the [Grave Accent] cause it would markdown all the text. I didn't find other way to write it down*

niik:
Hi @Lalo611, thanks for opening this!

It seems [french users have a similar issue](https://github.com/desktop/desktop/pull/2385#issuecomment-422323022). In hindsight it wasn't the best choice. It was a carry-over from the old GitHub Classic apps for Windows which used the shortcut as a sort of homage to old FPS games (https://github.com/desktop/desktop/issues/2138#issuecomment-312948687).

I'd be up for changing it but I'm vary of breaking existing users' flows. There was also a long debate about what would be a better alternative in the issue I linked to above but it never got resolved. I don't think full customization of shortcuts is going to be something we'll be able to prioritize in a while.

So if anyone has clever ideas I'm all ears. Could we detect these layouts somehow and dynamically change it? Or should we just bite the bullet and figure out a better shortcut for all? (I'm partial to the latter).

Lalo611:
Thank you for keeping the issue.
I'm not very familiarized coding, I mostly use GitHub for translating purposes, so it is not an annoying issue for me.
I'm aware that most users use that shortcut, that's why I proposed the other solution but I don't know how complex it could be.

Due to not being good at codes and all that stuff I don't know if the code has a feature to recognize the layout so that the shortcuts can be personalized in some way.

In my personal opinion changing the shortcut for all is not quite an option, there will always be dissatisfied people.

PR #2385, opened 2017-08-02
Adds short-cut to open terminal/console

iAmWillShepherd:
This PR fixes issue #2138.

Now you can open a terminal or console using ```Control/Ctrl + ` ```

PR #16332, opened 2023-03-16
Resize sidebar with keyboard shortcut

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3260

## Description
This adds a keyboard shortcut that allows users to increase or decrease the size of of any active resizable component.

## Release notes
Notes: [Improved] Users can resize the app sidebar via keyboard shortcuts.

### DIFF
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -12,6 +12,7 @@ import { AppMenu, MenuItem } from '../models/app-menu'
 import { hasConflictedFiles } from './status'
 import { findContributionTargetDefaultBranch } from './branch'
 import { enableStartingPullRequests } from './feature-flag'
+import { getLastResizableFocused } from '../ui/resizable'
 
 export interface IMenuItemState {
   readonly enabled?: boolean
@@ -380,7 +381,7 @@ function getMenuState(state: IAppState): Map<MenuIDs, IMenuItemState> {
 
   return getAllMenusEnabledBuilder()
     .merge(getRepositoryMenuBuilder(state))
-    .merge(getAppMenuBuilder())
+    .merge(getAppMenuBuilder(state))
     .merge(getInWelcomeFlowBuilder(state.showWelcomeFlow))
     .merge(getNoRepositoriesBuilder(state)).state
 }
@@ -431,18 +432,20 @@ function getNoRepositoriesBuilder(state: IAppState): MenuStateBuilder {
   return menuStateBuilder
 }
 
-function getAppMenuBuilder(): MenuStateBuilder {
+function getAppMenuBuilder(state: IAppState): MenuStateBuilder {
   const menuStateBuilder = new MenuStateBuilder()
 
-  const isResizableActive =
-    document.activeElement?.closest('.resizable-component') !== null
+  const lastResizableFocused = getLastResizableFocused(
+    state.documentFocusedElements
+  )
+
   menuStateBuilder.setEnabled(
     'increase-active-resizable-width',
-    isResizableActive
+    lastResizableFocused !== null
   )
   menuStateBuilder.setEnabled(
     'decrease-active-resizable-width',
-    isResizableActive
+    lastResizableFocused !== null
   )
 
   return menuStateBuilder
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -12,7 +12,6 @@ import { AppMenu, MenuItem } from '../models/app-menu'
 import { hasConflictedFiles } from './status'
 import { findContributionTargetDefaultBranch } from './branch'
 import { enableStartingPullRequests } from './feature-flag'
-import { getLastResizableFocused } from '../ui/resizable'
 
 export interface IMenuItemState {
   readonly enabled?: boolean
@@ -434,19 +433,10 @@ function getNoRepositoriesBuilder(state: IAppState): MenuStateBuilder {
 
 function getAppMenuBuilder(state: IAppState): MenuStateBuilder {
   const menuStateBuilder = new MenuStateBuilder()
+  const enabled = state.resizablePaneActive
 
-  const lastResizableFocused = getLastResizableFocused(
-    state.documentFocusedElements
-  )
-
-  menuStateBuilder.setEnabled(
-    'increase-active-resizable-width',
-    lastResizableFocused !== null
-  )
-  menuStateBuilder.setEnabled(
-    'decrease-active-resizable-width',
-    lastResizableFocused !== null
-  )
+  menuStateBuilder.setEnabled('increase-active-resizable-width', enabled)
+  menuStateBuilder.setEnabled('decrease-active-resizable-width', enabled)
 
   return menuStateBuilder
 }
--- a/app/src/lib/menu-update.ts
+++ b/app/src/lib/menu-update.ts
@@ -380,6 +380,7 @@ function getMenuState(state: IAppState): Map<MenuIDs, IMenuItemState> {
 
   return getAllMenusEnabledBuilder()
     .merge(getRepositoryMenuBuilder(state))
+    .merge(getAppMenuBuilder())
     .merge(getInWelcomeFlowBuilder(state.showWelcomeFlow))
     .merge(getNoRepositoriesBuilder(state)).state
 }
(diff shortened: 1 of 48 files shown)


app/src/main-process/menu/build-default-menu.ts
```diff
@@ -229,6 +229,16 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+-',
         click: zoom(ZoomDirection.Out),
       },
+      {
+        label: __DARWIN__ ? 'Decrease Sidebar Width' : 'Decrease sidebar Width',
+        accelerator: 'CmdOrCtrl+Shift+-',
+        click: emit('decrease-sidebar-width'),
+      },
+      {
+        label: __DARWIN__ ? 'Increase Sidebar Width' : 'Increase sidebar width',
```
tidy-dev: Open to suggestion on wording and accelerator. I based it on the zoom in/zoom out accelerator with the Shift.

Other options I thought of:
- "Increase/Decrease Sidebar"
- "Expand/Shrink Sidebar" (Width)

Invert and say: 
- "Increase View"
This is how visual studio code does it, but this doesn't make as much sense in desktop because the sidebar also represents the repository list width. 

Accelerator:
- ctrl+shift+i and ctrl+shift+d for increase/decrease.

app/src/lib/stores/app-store.ts
```diff
@@ -4898,6 +4903,25 @@ export class AppStore extends TypedBaseStore<IAppState> {
     return Promise.resolve()
   }
 
+  public _resizeSidebarWidth(resizeDirection: ResizeDirection): Promise<void> {
+    const { value, min, max } = this.sidebarWidth
+    const changedWidth =
+      resizeDirection === ResizeDirection.Decrease ? value - 1 : value + 1
```
tidy-dev: Could be more aggressive than 1? This hardly is noticeable if you use the menu item... but allows fine control if using keyboard shortcuts? Maybe menu item should have a higher factor?

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -229,6 +229,16 @@ export function buildDefaultMenu({
         accelerator: 'CmdOrCtrl+-',
         click: zoom(ZoomDirection.Out),
       },
+      {
+        label: __DARWIN__ ? 'Decrease Sidebar Width' : 'Decrease sidebar Width',
```
tidy-dev:
```suggestion
        label: __DARWIN__ ? 'Decrease Sidebar Width' : 'Decrease sidebar width',
```

niik:
While this solves the issue for the sidebar we do have other resizable panes that I imagine will encounter the same issue (the file list in the history view, the stash view, and soon the pull request dialog). We can't add keyboard shortcuts for all of these right? So I wonder if we should either just make the resizable component tab reachable and then handle keyboard events on that component or if we should make it programmatically focusable (tabindex=-1) and have the menu items move focus to the different resizables where the user then can use arrow keys to resize (and perhaps Shift+Arrow keys to increase by 10px increments).

The accessibility support for a resizable handle seems promising (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/separator_role#focusable_separator) at least?

tidy-dev:
> we do have other resizable panes

Yep. What you are saying sounds like a good idea to me.

Another idea I was thinking of is having a "resize settings" dialog where users could set all these different keyboard shortcuts. So in the view menu, you have a "Resize settings" link that opens to a new section in the preferences that allows setting of all these. But, this idea does open the can of worms of custom shortcuts.. so I like your better :D I will explore that.

app/src/main-process/menu/build-default-menu.ts
```diff
@@ -229,6 +229,20 @@ export function buildDefaultMenu({
...
+          : 'Contract active resizable',
+        accelerator: 'CmdOrCtrl+Shift+-',
+        click: emit('decrease-active-resizable-width'),
+      },
+      {
+        label: __DARWIN__
+          ? 'Expand Active Resizable'
+          : 'Expand active resizable',
+        accelerator: 'CmdOrCtrl+Shift+=',
+        click: emit('increase-active-resizable-width'),
```
tidy-dev, line 238: Again totally open to naming ideas.. I used to have "Increase active resizable width" and "Increase active resizable pane"... but those were both really long. 

Other thoughts. "Expand Active Pane" seems more relatable but maybe vague?

app/src/ui/resizable/resizable.tsx
```diff
@@ -77,6 +83,79 @@ export class Resizable extends React.Component<IResizableProps> {
...
+  }
+
+  /**
+   * Handler for when a user uses keyboard shortcuts to resize the size the
+   * active resizable
+   */
+  private handleMenuResizeEvent(resizeDirection: ResizeDirection) {
+    const { width } = this.props
+    const changedWidth =
+      resizeDirection === ResizeDirection.Decrease ? width - 5 : width + 5
```
tidy-dev, line 119: Decided on 5 because 1 felt way to slow.. but open to other suggestions

app/src/lib/menu-update.ts
```diff
@@ -430,6 +431,23 @@ function getNoRepositoriesBuilder(state: IAppState): MenuStateBuilder {
   return menuStateBuilder
 }
 
+function getAppMenuBuilder(): MenuStateBuilder {
+  const menuStateBuilder = new MenuStateBuilder()
+
+  const isResizableActive =
+    document.activeElement?.closest('.resizable-component') !== null
+  menuStateBuilder.setEnabled(
```
tidy-dev: Why does this crash the app? I must be completely overlooking something obvious...

tidy-dev: It says 'increase-active-resizable-width' is an unknown id?

tidy-dev: Yep it was something obvious. I had copied the zoom/zoom out  menu item entries for the menu directly above.. which we don't ever disable.. If you want to be able to "id"entify one of these you need to give it an "id". 🤦

niik:
I'm not sure I understand the purpose of keeping track of the last selected resizables here. The Windows menu should return focus to the element that had it prior to opening the menu before the menu event is dispatched so it should reach the correct resizable automatically. Are you finding that it's not? I did a quick test [here](https://github.com/desktop/desktop/compare/resize-pane) and it appeared to be working as intended.

tidy-dev:
> should reach the correct resizable automatically

I found that the functionally if the keyboard shortcuts worked perfectly for windows and mac. At that point, I had these menu items enabled at all times.

> purpose of keeping track of the last selected resizables here

At first, added event listeners to the resizable components focus and blur to disable/enable the expand/contract menu items when and resizable is in focus. I just used `document.activeElement.closest('.resizable-component')` to determine if a resizable was in focus for whether the item should be enabled. Thus, if you are in the diff, and open view the expand/contract menu items should be disabled since it is not a resizable. This worked for macOS because opening the system menu does not incur a change of focus. However, since on windows, the app menu is another rendered element, then it caused the `document.activeElement` to be the focused menu item and resulting in the contract/expand menu items to be already disabled when you opened the view menu. 


If you have a more deterministic method to achieve this, I am totally open to better suggestions. I considered just always having these enabled instead of going through this active element tracking hoops since the menu item is resize "active" pane.. but that does leave users with a button they can press that does nothing .. which I was like.. well could do some sort of notification that no active panes are in focus... but that felt ickier to me.. tho open to others opinions.

niik:
> This worked for macOS because opening the system menu does not incur a change of focus. However, since on windows, the app menu is another rendered element, then it caused the document.activeElement to be the focused menu item

That's what I was missing, thanks for adding that context! I would like to see if we could find a way to avoid storing DOM elements in the app state. While I don't see that happening here retaining DOM elements is a common way to "leak" memory that we don't have to hold on to. I took a stab at storing just the information we need (i.e whether a resizable component would likely react to a resize event being dispatched at a given time). Please take a look and see if that's an approach you'd be comfortable with here or if there's some other reason I'm missing that requires us to keep an array of elements in the app store: https://github.com/desktop/desktop/compare/resize-pane-with-keyboard-shorcut...avoid-retaining-dom-elements

Besides that I would love to try and find a different keyboard combination here than `CtrlOrCmd+[`. The `[` and `]` characters require the use of the AltGr key on many European keyboard layouts (see https://github.com/desktop/desktop/issues/10622, https://github.com/desktop/desktop/pull/2385#issuecomment-422323022) making it impossible to use this shortcut on those layouts.

tidy-dev:
> Please take a look and see if that's an approach you'd be comfortable with here or if there's some other reason I'm missing that requires us to keep an array of elements in the app store: https://github.com/desktop/desktop/compare/resize-pane-with-keyboard-shorcut...avoid-retaining-dom-elements

Much simpler! Thank you. 💖 Totally makes sense about the storing dom elements.. I wasn't a huge fan of that either (but my brain only went so far as to limit it to two :D) 

> The [ and ] characters require the use of the AltGr key on many European keyboard layouts

TIL of these keyboard layouts. I only switched from `CtrlorCmd`+`Shift`+`+/-` because when the shortcut was disabled I kept accidentally zooming in and out. I just made `CtrlorCmd`+`8\9` as I can't think of any letter keys that make sense and I think they should side by side.

niik (approved):
Love it, thanks for iterating on this!

PR #16335, opened 2023-03-17
Co-author revamped: less CodeMirror, more accessibility

sergiou87:
Closes https://github.com/github/accessibility-audits/issues/3287
Closes https://github.com/github/accessibility-audits/issues/3246
Closes https://github.com/github/accessibility-audits/issues/3244
Closes https://github.com/github/accessibility-audits/issues/3014

## Description

Up until now, the co-author input field was implemented using CodeMirror under the hood for _reasons_ (probably because of how nice it is to have a styled input text). However, due to CodeMirror, it's hard to make the component accessible.

This PR rewrites that component from the scratch, uses the existing `AutocompletingTextInput` component (that was made accessible in #16324), and adds other accessibility improvements needed:
- Screen readers now announce the autocomplete suggestion chosen by the user.
- Screen readers now announce the `Co-authors` label

It also changes a bit the UX:
- Now added co-authors are not part of the input text, but are siblings to it. Users can use arrow keys to navigate through them and delete them with <kbd>Backspace</kbd>/<kbd>Delete</kbd>.
- Now the input field always shows autocompletion suggestions when it's focused, since only known users _should_ be added.

One feature that has been lost (for now, at least) is undoing/redoing. Not sure how critical that is, but didn't want to make this PR bigger.

## Summary of Code Changes

- Replaced the `IAuthor` interface with the new `Author` type, that is the sum of `KnownAuthor` (for users we know exist in GH) and `UnknownAuthor` (for users that are either being searched for, or not found). This is what made this PR affect so many files, but it's not that bad!
- Added `CoAuthorAutocompletionProvider`. It reuses most of the code from our `UserAutocompletionProvider` (used for user autocompletion when you type `@` in the commit message textboxes) and only adds support for "unknown users" (AKA random strings that we check if match existing GH users). It also removes the requirement of `@` at the beginning, allowing us to always show autocomplete suggestions.
- Some new features to our `AutocompletingInput`:
   - Added support for `alwaysAutocomplete`: every time the input has focus, it will show the suggestions.
   - Added `autocompleteItemFilter` prop so we can filter out users that are already added
- `CommitMessage` now checks if there are unknown authors and shows a confirmation dialog warning about that (the `UnknownAuthors` dialog). Slightly hacky, but I made `CommitMessage` to pass `this.createCommit` as the `onCommitAnyway` callback of the said dialog.
- `AuthorInput` and `AuthorHandle` are all new, sorry.
- Added `AriaLiveContainer` as an easy & reusable way to force screen readers to re-read aria live sections. In this example: if you type something and get back exactly the same number of suggestions as before, this component will add/remove a `&nbsp;` to force the screen reader to read it.

### Screenshots

https://user-images.githubusercontent.com/1083228/227506420-be708c92-54a1-4085-99a2-ffe68e5597f0.mp4

## Release notes

Notes: [Improved] Make co-authors input text accessible


sergiou87:
👋 @github/accessibility-reviewers 

We have addressed this issue in https://github.com/desktop/desktop/pull/16335

Test build to try it out:
- macOS: https://centralreleases.blob.core.windows.net/github-desktop/releases/3.2.1-test12-edb30e85/GitHubDesktop-x64.zip
- Windows: https://centralreleases.blob.core.windows.net/github-desktop/releases/3.2.1-test12-edb30e85/GitHubDesktopSetup-x64.exe

Thanks in advance 🙏 ❤️

app/src/ui/accessibility/aria-live-container.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * This component encapsulates aria-live containers, which are used to
+ * communicate changes to screen readers. The container is hidden from
+ * view, but the screen reader will read the contents of the container
+ * when it changes.
+ *
+ * It also allows to make an invisible change in the content in order to force
+ * the screen reader to read the content again. This is useful when the content
+ * is the same but the screen reader should read it again.
+ */
+export class AriaLiveContainer extends Component<IAriaLiveContainerProps> {
```
sergiou87, line 21: Probably interesting to use in other places 😄

tidy-dev: Most definitely.. On the commit completion status I am tagging on the commit short sha for this purpose. I could do this instead, tho I guess the short sha does serve the purpose in clearly indicating to the user that is a different commit with the same summary (if they remember what the last short sha was :D)

tidy-dev (approved):
✨ Great work!

PR #16339, opened 2023-03-17
Add new RStudio bundle identifier

steveward:
We received a report that came in via the Docs feedback form that the RStudio integration was not working on macOS. I was able to reproduce the issue, and digging into it further found that the bundle identifier had changed.

## Description

- Adds a second bundle identifier for RStudio to `darwin.ts`

## Release notes

Notes:

Confirmed the change works in a test build:

Screenshot 2023-03-17 at 12 26 55 PM


ukn8:
هلو

sergiou87 (approved):
Thanks for this fix! ❤️

PR #16340, opened 2023-03-17
Announce Committing Status

tidy-dev:
Closes https://github.com/github/accessibility-audits/issues/3279

## Description
This adds an `aria-live` to announce the loading message and after 2 seconds of completion updates it to a Committed just now message. 

### Screenshots

macOS

Windows

## Release notes
Notes: [Improved] Committing status after submitting a commit is screen reader announced.

### DIFF
--- a/app/src/ui/changes/undo-commit.tsx
+++ b/app/src/ui/changes/undo-commit.tsx
@@ -27,6 +27,14 @@ export class UndoCommit extends React.Component<IUndoCommitProps, {}> {
   private button: HTMLButtonElement | null = null
 
   public componentDidMount(): void {
+    this.focusButtonIfJustCommitted()
+  }
+
+  public componentDidUpdate(): void {
+    this.focusButtonIfJustCommitted()
+  }
+
+  private focusButtonIfJustCommitted = () => {
     const diff = this.props.commit.author.date.getTime() - Date.now()
     const duration = Math.abs(diff)
 
--- a/app/src/ui/changes/undo-commit.tsx
+++ b/app/src/ui/changes/undo-commit.tsx
@@ -24,6 +24,21 @@ interface IUndoCommitProps {
 
 /** The Undo Commit component. */
 export class UndoCommit extends React.Component<IUndoCommitProps, {}> {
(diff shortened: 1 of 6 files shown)


app/src/ui/changes/undo-commit.tsx
```diff
@@ -24,6 +24,24 @@ interface IUndoCommitProps {
...
 /** The Undo Commit component. */
 export class UndoCommit extends React.Component<IUndoCommitProps, {}> {
+  private button: HTMLButtonElement | null = null
+
+  public componentDidMount(): void {
+    const diff = this.props.commit.author.date.getTime() - Date.now()
+    const duration = Math.abs(diff)
+    if (duration < 1000 && this.props.isCommitting === false) {
+      this.focusButton()
+    }
```
sergiou87: What's the reasoning behind this? Because I think this could not work if a "refreshing the repo" after a commit takes longer than 1s (which I think happens more often on Windows) 🤔

sergiou87:
I couldn't get the "move focus to Undo button" to work reliably. I'm also not sure about that behavior, since the `Undo` button is potentially destructive (specially if the user disabled the warning prompt). Looking at the accessibility issue, and unless I misread it, seems like moving the focus to the undo button is an option, but not mandatory, and just using the `aria-live` to inform about changes in the commit status is enough.

app/src/ui/changes/undo-commit.tsx
tidy-dev: Yeah. This part I wasn't too happy about, but was having trouble finding a better way.

tidy-dev:
> seems like moving the focus to the undo button is an option, but not mandatory, and just using the aria-live to inform about changes in the commit status is enough.

I was thinking the ticket indicated that we needed to announce when it finished too and a common way of doing that is to move focus.  However, it is a good point about Undo being somewhat destructive, I will take another look at just announcing the "Committed just now" part.

sergiou87 (approved):
As discussed privately, the screen reader sometimes interrupts the "committing" and "committed just now" messages, but this can work for now! :shipit:

PR #16345, opened 2023-03-20
Add option to create orphan branch like Batman

### LINKED ISSUE #7929: Create Empty/Orphan Branch
We can only create a branch based on current branches.
If we want to create an empty branch
We must use the command line.
Can desktop add this?
```
git checkout --orphan *EmptyBranchName*
```

andrewhong04:
Fixes https://github.com/desktop/desktop/issues/7929

### Description
Not entirely sure if this is the correct implementation as the only way to create an orphan branch is by doing `git checkout --orphan <branch_name>` while `createBranch()` uses the command `git branch ...`. In the meantime, please enjoy my Batman meme.

### Screenshots
imageimage

## Release notes

Notes: Add an option to create an orphan branch

### DIFF
--- a/app/src/lib/git/branch.ts
+++ b/app/src/lib/git/branch.ts
@@ -28,6 +28,15 @@ export async function createBranch(
   noTrack?: boolean,
   orphan?: boolean
 ): Promise<void> {
+  if (orphan) {
+    await git(
+      ['checkout', '--orphan', name],
+      repository.path,
+      'createOrphanBranch'
+    )
+    return
+  }
+
   const args =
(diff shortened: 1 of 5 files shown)


tidy-dev:
👋 @novialriptide Thanks for wanting to improve GitHub Desktop! This is not a commonly used feature that does not support the standard git workflow, and therefore, the GitHub Desktop team is not currently open to supporting it. 

Other feedback, if we were to support this type of feature, it would be in a lower profile way as to not encourage non-standard workflows that lead (possibly unknowing) users to be in unusual scenarios.

novialriptide:
Thanks for the feedback, should the issue relating to this pr also be closed?

PR #16409, opened 2023-03-28
Commit Detail Expander button linter cleanup

tidy-dev:
## Description
 An incidental positive is that now this is part of the tab order/keyboard accessible where it was not before.

Some easy cleaning up for SLA goal of removing these:
Bar graph of a11y linters in desktop/desktop repo

### Screenshots

CleanShot 2023-03-28 at 16 46 11

## Release notes
Notes: no-notes


sergiou87 (approved):
Looks and works as expected!

PR #16410, opened 2023-03-28
Stashed Changes Button: Remove aria-selected attribute and linter

tidy-dev:
## Description

This just removes the `aria-selected` attribute from the stashed changes button. I could not find a reason we had it. I tested visually and with voice over and did not observe a change from removing this... tried to use git history to find a reason.. but I got lost in the refactors. 

Some easy cleaning up for SLA goal of removing these:
Bar graph of a11y linters in desktop/desktop repo

## Release notes
Notes: no-notes

### DIFF
--- a/app/src/ui/changes/changes-list.tsx
+++ b/app/src/ui/changes/changes-list.tsx
@@ -884,12 +884,10 @@ export class ChangesList extends React.Component<
     )
 
     return (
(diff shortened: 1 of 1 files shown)


sergiou87 (approved):
With the little I've learned about a11y these days, this change makes sense to me. I can't tell why we would want that 🤔  Looks and works great!

PR #16425, opened 2023-03-31
Change list scrollable when zoomed

tidy-dev:
xref https://github.com/github/accessibility-audits/issues/3290
xref https://github.com/github/accessibility-audits/issues/3289

## Description
When the app is zoomed to 200%,
1. The undo button no longer overlaps the commit button.
2. The changes list now has a minimum height of 100px so a few list items are always visible.
3. The whole changes list panel starts scrolling when they all will no longer fit in the vertical height.
4. The description text area collapses to 20px height instead of 80px when zoomed in to reduce panel scrolling when zoomed.
5. Once changes list was visible, the file names were no longer visible due to some clipping logic based on an inaccurate available width.
     - That available width was the sidebar width was being clamped between a min and max. The max being window.innerWidth minus a fixed const of 460px that represents the toolbar buttons width. When the app is zoomed the window.innerWidth to be not much more than 460px. Thus, even tho the sidebar width value is never goes below 220 (the minimum), the max is around 50px and the clamped value was being used which would be 50px. 

### Screenshots

https://user-images.githubusercontent.com/75402236/229001686-5eb8b5b1-e006-4727-a82d-6d6b9c8d79d0.mp4

## Release notes

Notes: [Fixed] Clipping and overlapping of the changes list fixed at 200% zoom.


app/src/ui/repository.tsx
```diff
@@ -209,7 +208,7 @@ export class RepositoryView extends React.Component<
         : null) || null
 
     // -1 Because of right hand side border
-    const availableWidth = clamp(this.props.sidebarWidth) - 1
+    const availableWidth = this.props.sidebarWidth.value - 1
```
niik: This doesn't seem like the correct approach. The thought behind clamping here is that we attempt to preserve the user's preferred width as the window changes. So if they have set the sidebar width to 300px and have the horizontal real estate to accommodate that but then resize the app to the smallest width the sidebar will get compressed to its minimum width only for it to return to 300px again when the window is resized.

If we need to make the sidebar smaller when zoomed then that logic should get incorporated into the `updateResizableConstraints` method in AppStore.

Note that the same clamping operation happens for the repo list foldout as well.

Lastly, but only semi-related to this PR, we should probably change the constrain logic to match how CSS gives precedence to min-width.

```diff
diff --git a/app/src/lib/stores/app-store.ts b/app/src/lib/stores/app-store.ts
index a8702e62a..95eeec32c 100644
--- a/app/src/lib/stores/app-store.ts
+++ b/app/src/lib/stores/app-store.ts
@@ -7759,5 +7764,8 @@ function constrain(
   min = -Infinity,
   max = Infinity
 ): IConstrainedValue {
+  // Match CSS's behavior where min-width takes precedence over max-width
+  // See https://stackoverflow.com/a/16063871
+  max = max < min ? min : max
   return { value: typeof value === 'number' ? value : value.value, min, max }
 }
```

niik (changes requested):

app/src/ui/repository.tsx
tidy-dev: > This doesn't seem like the correct approach. The thought behind clamping here is that we attempt to preserve the user's preferred width as the window changes. So if they have set the sidebar width to 300px and have the horizontal real estate to accommodate that but then resize the app to the smallest width the sidebar will get compressed to its minimum width only for it to return to 300px again when the window is resized.

Thanks for that explanation.. I was trying to figure out why it was there since I didn't see a change in behavior.. but I didn't not thinking of that use case.

tidy-dev: I think your matching of css behavior will solve this actually because the constrain is applied in the `updateResizableConstraints `. Applying and testing now.

tidy-dev: Yep. Works great!

niik (approved):

Gotojail (commented):
Get your body, your fukn easte of space in my very living eccentric eccetra, to your PLACE WITH NONE FROM MY ECCSPECTATIONS OF LOFE AS I. ME. THE. AS I AM CALLED TO CAST OK NGS FROM THE PLACE ON HIGH AS I W24 AM TAXIZED CALLED EXEMPT. FROM  EXITS OF DAILY OPERATIONS FOR AFFAIRS IN BUSINESS BIZINESS BBIDNESSES WERE 24 ERE FORMULATED UN COMMISSIONED LARGE AREA WORKS... PAST. ALL IN THR FACTORS. INTEGRATIONS ARE MY PROPERTYS NOT ANY OTHER HUMAN IS.... TO ... GAIN.... BE FOREFRONT BE FOURTHERIGHT OR FOURTHRITE AS INFERIORITY OF MY ASSUMPTIONS AS YOU TAKE OVER INFOLTAMRATION IS UNDERMINING. MINING WAS OVERTURED. IM THE SAY SO. THIS IS MY AS I AM NOT A BLACK AND WHITE SKIN FPHARING BASTARD IN FLESH AS COSMIONAUT. YOU COZY? SUPREMIST? ARTAN BROTHER OF ALL HOODED? THAT WAS A WELDING SUIT FOR THE DEFLECTION OF BURNS .... TO.... SKIN FLESH.... UNDERWATER WELDING IS IN THE FUNNYJOKE JOSHING YA AS SEA PHARING ANNA  ZILCHNIGH ANN. A MAXFIELD  LINDSEY ANN, A MAXFIELD  IS OWNER AS FIRST, LINDSEY FOR HER EXHONERATANCES IN YOU GUESS NOT AS AN INVITED GURST, IT KEANS I NEVER ASKED YOU YOUR NAME IN MY SYSTEM ATANY REGISTRY YOU DID. IN MY NAKE LINDSEY UNDER THE TABLE ANNA, OVERMY HEAD WAS NOT TO BE OF ANY ISSUESI YOU MADE MY NAME AND STATE NUMBER THE ISSUER. OF MEDICARE MEDICAID MEDIGAP... INS OF MY ONS TAKEN IN OFF THE RECORD, CLAIMS.... BY FORM. IN MY LAST TITLES CATEGORY OR ALPHA DEXTUAL LETTER AS BET. IZOD  IZEVEN. AS IF I ASKED FOR MY THINGS AS THE ODD SIDE FROM EVEN. +- ARE MY JOUNT SIDES  I AM A BY I STAND SIT STATUESQUE OR ANIMATE. MOTION LESS OR SOGN MORE IN MOVEMENT AND MOVIES... ARE COMMISSIONED... AS A PAST THING IN DONE. I AM INFERIOR I AM T
YOUR SUPERIOR ANNA   SURPRISE ME WITH THAT BRING TO, ALL IS RE:LIVERABLE... Relience patterns  its the map to LINDSEY, WHEREVER I AM. ON ANY MAP AS I. NIT I MAP FROM NATIVES AS TEETH TAKES YOU TO PLYMOUTH TO GAIN MY MANDIBLES MANDIBULAR, MASTENOID AS ANNOYED YOU MAY ACT NOT AS ANN IN SECOND, OR IN HUMAN AT ALL FIR MY ANUAL... MANUAL... MANEUVERS... IN THIS THING CALLED LIFE  MY RELICS ARE ON THE SHELF AND NO FIGURINE HAS MY FACE.... FACADES CASCADES ANNA  IS NOT A GRADE FOR LYNNS SALOON. OR ANNAS SALON.ON TYSON STREET. VERSUS. VERSES. REVOCATION OF LIFE IS NON REVOKABLE MR OKLAHOMA. THE WOLF IS STANDING AS BEGORE ME AVTING AS A SHADE TREE AND YOU PAY HIM AS A FALSE IDOL... Arlington, youre a shame not me. Not i  you are the clever... So bring it. My doorsteps got a bell also. All so. Hearts are just a shape mine is live and has a police dive also. Its my drupals quadrupals and youre profane in it called the triconomissed beat. You get noundermined priveliges orprivelous ages... Of my childtime of little to paid mindes.... You take and take in w24 v2v4. Thats my accounts legumrs my legs to stand eith have feets. You show no known feits. My pay forfeits was made unremarkable as title transfers are and were firbidden. Ao you became bidders. Estimators. This is false requisites. No total balances were yo come to these persons of defeated in the past  my human i.p. used as married by proxy

PR #16449, opened 2023-04-03
Use toggle tip component for commit summary length hint

tidy-dev:
xref https://github.com/github/accessibility-audits/issues/3256

## Description
This PR introduces a toggle tooltip component to allow us to have tooltip that is activated "on click" of a button via keyboard and hover of a mouse. The toggle tip pattern is used for things like "info" icon tooltips. The toggle tip pattern allows something that may typically only be accessible by mouse hover to also be keyboard navigable. Just like tooltips it shouldn't be used for information crucial to completing the task at hand.

### Screenshots
macOS

https://user-images.githubusercontent.com/75402236/229596701-e48c1c3d-3585-4f6b-ab5c-62c2b3367403.mp4

Windows

https://user-images.githubusercontent.com/75402236/229597546-fff7624d-3687-4231-8616-70186b48b276.mp4

## Release notes
Notes: [Improved] The commit length hint is keyboard and therefore screen reader accessible.

### DIFF
--- a/app/src/ui/lib/toggletipped-content.tsx
+++ b/app/src/ui/lib/toggletipped-content.tsx
@@ -27,31 +27,43 @@ interface IToggledtippedContentProps
   readonly ariaLabel?: string
 }
 
+interface IToggledtippedContentState {
+  /** State of when the tooltip is visible */
+  readonly tooltipVisible: boolean
+}
+
 /**
  * A less keyboard toggleable version of the Tooltip content component for when
  * it's acceptable to add a wrapping element around the content. The content
  * will be wrapped in a button as it is toggleable. supports all the options
  * that the Tooltip component does without having to worry about refs.
  **/
-export class ToggledtippedContent extends React.Component<IToggledtippedContentProps> {
+export class ToggledtippedContent extends React.Component<
+  IToggledtippedContentProps,
+  IToggledtippedContentState
+> {
   private buttonRef = createObservableRef<HTMLButtonElement>()
   private shouldForceAriaLiveMessage = false
-  private tooltipVisible = false
+
+  public constructor(props: IToggledtippedContentProps) {
+    super(props)
+    this.state = {
+      tooltipVisible: false,
+    }
+
+    this.buttonRef.subscribe(this.onButtonRef)
+  }
 
   private onToggle = () => {
     this.shouldForceAriaLiveMessage = !this.shouldForceAriaLiveMessage
   }
 
   private onTooltipVisible = () => {
-    this.tooltipVisible = true
+    this.setState({ tooltipVisible: true })
   }
 
   private onTooltipHidden = () => {
-    this.tooltipVisible = false
-  }
-
-  public componentDidMount(): void {
-    this.buttonRef.subscribe(this.onButtonRef)
+    this.setState({ tooltipVisible: false })
   }
 
   private onButtonRef = (ref: HTMLButtonElement | null) => {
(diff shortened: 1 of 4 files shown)


app/src/ui/lib/toggletipped-content.tsx
```diff
@@ -0,0 +1,111 @@
...
+    this.tooltipVisible = false
+  }
+
+  public componentDidMount(): void {
+    this.buttonRef.subscribe(this.onButtonRef)
+  }
+
+  private onButtonRef = (ref: HTMLButtonElement | null) => {
+    if (ref === null) {
+      this.buttonRef.current?.removeEventListener(
```
niik: It doesn't look like we're ever subscribing to `tooltip-shown` so I don't understand why we're attempting to remove it here. Similarly it appears we never unsubscribe from the `tooltip-hidden` event?

app/src/ui/lib/toggletipped-content.tsx
```diff
@@ -0,0 +1,111 @@
...
+  private onTooltipVisible = () => {
+    this.tooltipVisible = true
+  }
+
+  private onTooltipHidden = () => {
+    this.tooltipVisible = false
+  }
+
+  public componentDidMount(): void {
+    this.buttonRef.subscribe(this.onButtonRef)
```
niik: We should subscribe in the constructor since the ref would already be assigned after the component has mounted.

app/src/ui/lib/toggletipped-content.tsx
```diff
@@ -0,0 +1,111 @@
...
+  private buttonRef = createObservableRef<HTMLButtonElement>()
+  private shouldForceAriaLiveMessage = false
+  private tooltipVisible = false
+
+  private onToggle = () => {
+    this.shouldForceAriaLiveMessage = !this.shouldForceAriaLiveMessage
+  }
+
+  private onTooltipVisible = () => {
+    this.tooltipVisible = true
```
niik: Given that we seem to be using this when rendering it should probably live in state so that we re-render the component when this changes right?

niik (changes requested):

app/src/ui/lib/toggletipped-content.tsx
tidy-dev: Hmm.. This looks messed up  from what I thought I landed on.... and am thinking.. how did this even work last time I tested it.  🤦 Sorry for definitely weird code.

tidy-dev: Here is what I had before and have updated too.. I am sorry for wasting your time.
Shows diff where remove/add listeners are applied to both visibility events

app/src/ui/lib/toggletipped-content.tsx
```diff
@@ -0,0 +1,123 @@
...
+    this.setState({ tooltipVisible: true })
+  }
+
+  private onTooltipHidden = () => {
+    this.setState({ tooltipVisible: false })
+  }
+
+  private onButtonRef = (ref: HTMLButtonElement | null) => {
+    if (ref === null) {
+      const currRef = this.buttonRef.current
```
niik: Unfortunately I don't think this will work or at least I would be very surprised if it did. The `current` property should be updated before the subscribers are called IIRC (or at least that would make the most sense).

So I think you're gonna have to keep a copy of the ref as an instance variable as well.

niik (changes requested):

app/src/ui/lib/toggletipped-content.tsx
Usrwrr: ۱

niik (dismissed):

niik (approved):

Issue #6836, opened 2019-02-13
unable to parse commit because commit message contains ASCII delimiter characters


ospaarmann:
## Description

When I open GitHub Desktop this error appears and keeps popping up: 

![screenshot 2019-02-13 at 10 19 36](https://user-images.githubusercontent.com/308086/52701008-79a64980-2f79-11e9-80f7-415ae65a8196.png)

The second part of the error message is a partial of a commit comment. The full commit looks like this:

```
commit <COMMIT_SHA>
Author: <Author Name> <HiddenName@users.noreply.github.com>
Date:   Fri Feb 8 14:50:24 2019 +0100

    Set queue_sid based on looked up queue (#546)

    Set queue_sid based on looked up queue
    Params sent by Twilio differ from usecase to usecase, so we need to make sure that a `^]queue_sid`^_ is always set, no matter how the call ended up in the queue.
```

This error also pops up whenever I click on the History tab or change the Mac OS Desktop to the one where GitHub Desktop is open.

## Version

* GitHub Desktop: 1.6.1

* Operating system: Mac OS 10.14.2

## Steps to Reproduce

I tried to reproduce it but this seems to only happen if another author commits with these characters in the commit message. Reproducing it locally wasn't possible for me. Since this is a private repo I cannot share it with you.

### Expected Behavior

No error message.

### Actual Behavior

Error message keeps popping up.

shiftkey:
@ospaarmann thanks for the report!

There's two places where we bubble up this error:

https://github.com/desktop/desktop/blob/697eb56865b084c915b1cb967bfeaa8a13d2e881/app/src/lib/git/log.ts#L131-L135

https://github.com/desktop/desktop/blob/f9b1afae4a2efd5cd48b59b6f56e4cbe7a1702d0/app/src/lib/git/for-each-ref.ts#L83-L87

In both of these cases we're asking Git to return each the values of each commit delimited, and it looks like a a character in your commit message is being parsed by us incorrectly.

Could you attach the full log files? The **Help** | **Show Logs in Finder** menu items should open the logs folder, and inside those are the log files generated by day. This will provide some more information to help troubleshoot which case is the problem one.

> Since this is a private repo I cannot share it with you.

Would you be able to extract the commit to a new repository that you could share? I'm only interested in the commit message and the specific characters inside and how they're tripping our parsing code.

The rough steps to do this:

 - `git checkout --orphan <branchname>` to create a new branch separate from the main history of your repository
[...]

PR #16451, opened 2023-04-04
Parse log output when retrieving stash entries using the appropriate parser

niik:
## Description

Noticed that getStashes parsed the output from `git log` using the old method of using delimiter characters. As of https://github.com/desktop/desktop/pull/11486 the preferred way is to use the log or for-each-ref parsers which will avoid problems such as https://github.com/desktop/desktop/issues/6836.

## Release notes

Notes: no-notes


sergiou87 (approved):
Nice cleanup! Looks and works great 🙌

PR #16472, opened 2023-04-06
Undo Confirmation Modal is an Alert Dialog

tidy-dev:
xref https://github.com/github/accessibility-audits/issues/3294

## Description
This PR implements the [Alert Dialog Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/) on our dialog components so we can use them for our alert (error) and confirmation modals. Then it uses it for the "Undo Commit" confirmation prompt.  

TLDR: Alert/Confirmation modals should have `role="alertdialog"` and have an `aria-describedby` property that holds the id of the element with the alert or confirmation message required by the user to proceed with whatever action was interrupted. 

If the typing stuff feels too messy to be worth it (or if the typing wizards here have a better approach to require the `ariaDescribedBy` prop when `alertdialog` is used), we can revert https://github.com/desktop/desktop/pull/16472/commits/2465f25d66328660b0c6544b855489ce1c9a79ba. 

### Screenshots
macOS

https://user-images.githubusercontent.com/75402236/230399602-70cab4f0-d800-44fc-9174-2bb4fcc10d20.mp4

windows:

https://user-images.githubusercontent.com/75402236/230401829-17a4a614-24ff-4ffd-ac1e-f49c5719b6ae.mp4

## Release notes
Notes: [Improved] The undo commit confirmation modal message is screen reader announced.

### DIFF
--- a/app/src/ui/dialog/dialog.tsx
+++ b/app/src/ui/dialog/dialog.tsx
@@ -130,12 +130,17 @@ interface IDialogProps {
    * of the loading operation.
    */
   readonly loading?: boolean
+}
 
+/**
+ * If role is alertdialog, ariaDescribedBy is required.
+ */
+interface IAlertDialogProps {
   /** This is used to point to an element containing content pertinent to the
    * users workflow. This should be provided for dialogs that are alerts or
    * confirmations so that that the information that is interrupting the user's
    * workflow is screen reader announced and acquire a response */
-  readonly ariaDescribedBy?: string
+  readonly ariaDescribedBy: string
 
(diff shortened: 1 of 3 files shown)


niik (changes requested):
> approach to require the ariaDescribedBy prop when alertdialog is used

I think we could make this just a tad nicer by having both `IAlertDialogProps` and `IDescribedByDialogProps` extend `IDialogProps` and then making `DialogProps` be just `IAlertDialogProps | IDescribedByDialogProps`. I think that makes it a little bit clearer.

niik (approved):

PR #16483, opened 2023-04-07
Remove windows controls from windows title accessibility flow

tidy-dev:
xref https://github.com/github/accessibility-audits/issues/3295

## Description
This PR adds `aria-hidden="true"` to the title bar window controls (minimize, maximize, close) for Windows OS so that they cannot be accessed screen reader technology such as NVDA browse mode. Note: We already have a `tabindex="-1"` that removes them from the tab order. 

## Release notes

Notes: [Improved] Windows title bar controls are not accessible by screen reader browse mode.

### DIFF
--- a/app/src/ui/window/window-controls.tsx
+++ b/app/src/ui/window/window-controls.tsx
@@ -102,6 +102,7 @@ export class WindowControls extends React.Component<{}, IWindowControlState> {
         tabIndex={-1}
         className={className}
         onClick={onClick}
+        aria-hidden="true"
       >
         <svg aria-hidden="true" version="1.1" width="10" height="10">
           <path d={path} />


app/styles/ui/window/_title-bar.scss
```diff
@@ -49,6 +49,7 @@
   // automatically even for borderless window so we only render
   // controls on Windows.
   .window-controls {
+    order: 3;
```
niik: This seems awfully fragile and easy for us to regress on. Could we at least add a comment about why this is important?

niik (changes requested):
Are regular window controls on native windows announced by NVDA et. al? Feels odd to me but 🤷 The acceptance criteria also seems to be to make sure that the controls doesn't disrupt when navigating through the app which I guess they'd still be doing with the PR right, only they'd come in a different order?

tidy-dev:
From issue comment:
> I see in the DOM that the windows-controls are placed after the app-menu-bar controls. It seems like we could switch the order in the DOM

This is why I went with this approach, but before merging this, I will hit up the accessibility team and see if removing it from the the order altogether is the correct approach.

From what I saw in the link provided about windows title bars (Not electron reference.. but Microsoft specs) the title bar would be expected to take "2 lines" worth of space with the windows control being above the menu items. Thus.. at least visually they are expected to be before the menu items. .. That said, I so far have been unsuccessful at navigating to windows controls in other windows apps. ( I have a hard time being sure I am in browse mode in my windows VM.. )

niik (approved):
✨

PR #16484, opened 2023-04-07
[Linux] Styling improvements to scrollbars and default layout

shiftkey:
## Description

This PR upstreams three fixes that I've been handling for a while on the Linux side:

 - toggling between light and dark themes does not update the application - 927bed218c0f7c73edfb3e10900e83ae05d33de7
 - restyle the scrollbar to look like it does on Windows - da438ba635f89d151a6f59caafe2a467edf53a28
 - remove the white space where the React menu is rendered on Linux - 611fa16ab0fe78b2d889b2b7ed9b6f8bf838b6b9

#### Current `development`

This is how the app looks on `development` when you switch the theme from dark to light:

And from light to dark:

Screenshot from 2023-04-07 12-48-28

**Note:** switching to/from high contrast theme seems unaffected. I haven't had a chance to dig into that, but I believe that's not GA'd just yet.

The `System` option is also not present - we enabled this back in [`2.9.9-linux2`](https://github.com/shiftkey/desktop/releases/tag/release-2.9.9-linux2). I've not had any reported problems reported recently, and I just tested this by switching between a light and dark OS theme on Ubuntu and saw the Desktop app react accordingly.

The default scrollbars on Linux are very chunky but otherwise suit the themes:

This white bar below the window menu is the HTML menu placeholder that we don't actually use on the Linux side:

Screenshot from 2023-04-07 12-50-58

Shout out to @jfgordon2 for the original version of the scrollbar fix https://github.com/shiftkey/desktop/pull/240

#### This branch

This is the light and dark mode with the new scrollbar designs (and the empty menu bar region removed)

As you hover over the scrollbar, it widens and becomes easier to click (like the Windows version):

## Release notes

Notes: Styling improvements on Linux to scrollbars and default layouts

### DIFF
--- a/app/styles/themes/_dark.scss
+++ b/app/styles/themes/_dark.scss
@@ -49,16 +49,16 @@ body.theme-dark {
    * Background color for custom scroll bars.
    * The color is applied to the thumb part of the scrollbar.
    *
-   * Note: Only applies to win32 platforms
+   * Note: Only applies to win32 and linux platforms
    */
   --scroll-bar-thumb-background-color: rgba(255, 255, 255, 0.2);
 
   /**
-   * Background color for custom scroll bars in their active state.
-   * The color is applied to the thumb part of the scrollbar.
-   *
-   * Note: Only applies to win32 platforms
-   */
+    * Background color for custom scroll bars in their active state.
+    * The color is applied to the thumb part of the scrollbar.
+    *
+    * Note: Only applies to win32 and linux platforms
+    */
   --scroll-bar-thumb-background-color-active: rgba(255, 255, 255, 0.5);
 
   // Box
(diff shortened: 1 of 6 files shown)


app/styles/themes/_dark.scss
```diff
@@ -49,16 +49,16 @@ body.theme-dark {
...
-   * Background color for custom scroll bars in their active state.
-   * The color is applied to the thumb part of the scrollbar.
-   *
-   * Note: Only applies to win32 platforms
-   */
+    * Background color for custom scroll bars in their active state.
+    * The color is applied to the thumb part of the scrollbar.
+    *
+    * Note: Only applies to win32 and linux platforms
+    */
```
niik:
```suggestion
   * Background color for custom scroll bars in their active state.
   * The color is applied to the thumb part of the scrollbar.
   *
   * Note: Only applies to win32 and linux platforms
   */
```

Seems there was some accidental indentation here

niik (changes requested):
Just a tiny request here @shiftkey and then I think we can take this

niik (approved):

Issue #13295, opened 2021-11-09
Codium not detected by github desktop


goyalyashpal:
copying 🐛 over from https://github.com/VSCodium/vscodium/issues/907

_update:_

* duplicate of: https://github.com/desktop/desktop/issues/10446
* fixed by: https://github.com/desktop/desktop/pull/16511

### Describe the bug
Github desktop (hereafter as ghd) is not recognising VSCodium

### Version & OS

- Ghd: 2.9.4 in "AppData/Local"
- OS: Windows 10 20H2
- Architecture: x64
- Codium: 1.61.2  
from [`VSCodium-x64-updates-disabled-1.61.2.msi`](https://github.com/VSCodium/vscodium/releases/download/1.61.2/VSCodium-x64-updates-disabled-1.61.2.msi) installed in "Program Files" (i.e. i guess for all users)
- NP++: 7.9.5 in "Program Files"
- Typora: 0.9.98 (beta) in "Program Files"

### Steps to reproduce the behavior

1. Open ghd
2. Go to "Files > Options" (<kbd>^</kbd><kbd>,</kbd>)
3. Then go to "Integrations > External Editor"
4. See that VSCodium is not there

### Expected behavior

VSCodium should be there

### Actual behavior

VSCodium is not there

### Screenshots

![image](https://user-images.githubusercontent.com/19423063/140774219-4e0f5d01-4a14-473f-9355-a020e406c6b5.png)

### Logs

Attach your logs by opening the `Help` menu and selecting `Show Logs...`, if applicable.

### Additional context

- I had ghd already installed since last month. I installed VSCodium only last week or so.
- I have restarted pc, and ghd ofc
- vscodium package used downloaded via https://github.com/VSCodium/vscodium/releases/tag/1.61.2
- similar issue: https://github.com/desktop/desktop/issues/7024

sergiou87:
Hi @yashpalgoyal1304 ! 👋 

Could you check in `regedit` if there is anything related to that VSCodium from these paths?

https://github.com/desktop/desktop/blob/da05a90a4b4310a3b98bf878794c0e9fa993f481/docs/technical/editor-integration.md?plain=1#L101-L113

That's where Desktop looks for the info about the integrated editors.

goyalyashpal:
Thanks @sergiou87 for getting back. 

Yes, VSCodium is there under the "64-bit Windows software":

[...]

PR #16511, opened 2023-04-11
fix VSCodium name

GitMensch:
as it never had a "Visual Studio" prefix, which is a registered name


niik (approved):
Thanks for the contribution @GitMensch!

@steveward Do you think you could take care of updating https://docs.github.com/en/desktop/installing-and-configuring-github-desktop/configuring-and-customizing-github-desktop/configuring-a-default-editor#supported-editors?

goyalyashpal:
i think this one fixes this one too: https://github.com/desktop/desktop/issues/13295

PR #16512, opened 2023-04-11
Recreate stash after renaming branch

### LINKED ISSUE #16442: Renaming Branch leaves stash behind in non-existing branch
### The problem

When having a stash (through the client) and renaming current branch, the stash stays under the old branch.
The old branch no longer exists and unless handled correctly, the stashed changes can be lost.

### Release version

3.2.0

### Operating system

Windows 10

Create a branch, do not publish it
Introduce changes, stash them
Rename branch.
You no longer have stashed changes

### Log files

_No response_

### Screenshots

_No response_

### Additional context

_No response_

### LINKED ISSUE #16848: Join
https://github.com/Jhoine/saeksunkhamsan-gmail.com/commit/ad6e20905bb7dcec6032df6e1707430bfa47745e

niik:
Closes #16442

## Description

I'm opening this a draft looking for thoughts and feedback. I wanted to see how complicated it would be to support "moving" (or re-associating) a Desktop stash entry when renaming a branch. This PR accomplishes this but I'm looking for flaws in my theory before I open this up for proper code review. Is this something we should support and is this an okay way to go about it?

#### A quick reminder of how stashes work

> A stash entry is represented as a commit whose tree records the state of the working directory, and its first parent is the commit at HEAD when the entry was created. The tree of the second parent records the state of the index when the entry is made, and it is made a child of the HEAD commit

From: https://git-scm.com/docs/git-stash#_discussion

Additionally a stash can have a third parent if created using the `--include-untracked` flag.

Stashes are then referenced by the special `refs/stash` reflog (and the most recent stash is referenced by the refs/stash ref as well). The reflog is what allows `git stash list` to show all created stashes. Desktop's custom stashes are built upon this system but use a specially formatted message to differentiate a Desktop stash from stashes created outside of Desktop as well as to associate them with a particular branch, ex: `!!GitHub_Desktop<main>`.

#### How this works

In order to "move a stash entry" we have to change the message (`!!GitHub_Desktop<main>`) so that it points to the new branch. We could get away with just changing the reflog but there are not reflog manipulation commands available so we have to drop the existing stash and create a new one. Applying the stash and the recreating it is not feasible since it's possible that the stash can't be applied cleanly. So instead we create a new, dangling, commit based on the same tree and parents but with a new message and store that in the reflog before we drop the old stash.

#### Leaky abstractions

- Recreating the stash entry will reorder the stash stack

#### Alternative approaches

- We could rewrite the stash reflog to keep the stash entry. Decided against this since there's no reflog manipulation plumbing commands available so we'd have to manually take care of it which is counter to our intention of only interfacing with the repo using Git commands. Even though the reflog format is straightforward there's more nuance than one would think around lock files, rewriting parent reflog entries etc etc.

## Release notes

Notes:


Alena32white:
`уволить`

sergiou87 (dismissed):
Other than changing the order of the stash list, I think this approach is quite clean, it doesn't feel that hacky as long as it's using documented APIs and in a quite straightforward way, I don't see anything weird that makes me feel uncomfortable, at all.

I like this implementation and love the feature, so for me it's a :shipit: !!!

tidy-dev (dismissed):
Makes sense to me. Thanks for all the explanation of how stashes work! 💖 

(Just to capture our sync convo somewhere, I think this PR is a great improvement on how we implemented stashes in Desktop as loosing a stash definitely feels like a bug/design flaw; and I think a more complete stash solution would have also solved this by allowing users to interact with all the stashes in their repo).

tidy-dev (approved):

PR #16525, opened 2023-04-12
Theme radio buttons refactor

tidy-dev:
xref: https://github.com/github/accessibility-audits/issues/3273

## Description
This PR updates the theme selector to look like radio buttons.

This PR: 
1. Introduces a `RadioGroup` component that we will likely need to wrap most of radio button usages with anyways because they should be in radio groups but are not currently.
2. Implements the `RadioGroup` component in the `Appearance` settings and adds custom styling to [mimic dotcom's theme selection.](https://github.com/settings/appearance).

### Screenshots
Preferences dialog opened to Appearance tab showing radio buttons with labels including an image swatch representative of dark, light, and system themes

## Release notes
Notes: [Improved] Make radio theme selection look like radio buttons.


app/src/ui/lib/radio-group.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * The currently selected item, denoted by its key.
+   */
+  readonly selectedKey: T
+
+  /**
+   * The keys of the radio buttons to display, in order of the radio buttons.
+   */
+  readonly radioButtonKeys: ReadonlyArray<T>
```
tidy-dev, line 16: This "in order of the radio buttons" doesn't feel happy to me since if you provide them in a different order than you provide the children to the `RadioGroup` then the radio buttons will behave weird about what is being checked. Granted one bout of testing should highlight that. The reason I did this was because I really like the straightforward readability of using the children in the `RadioGroup` as the labels for `RadioButtons` built in this component. If feels akin to if I were to do something like `<div role="radioGroup"><radio button><radio button></div>`; therefore easier to digest (especially for simple radio groups that just have text in the labels). 

An alternative pattern is to do like list row component and have a "renderLabel" prop that passes the key up. This would be more deterministic; and if anyone feels strongly that we should use that pattern I am open to changing it.

app/src/lib/feature-flag.ts
```diff
@@ -75,7 +75,7 @@ export function enableResetToCommit(): boolean {
 
 /** Should we allow high contrast theme option */
 export function enableHighContrastTheme(): boolean {
-  return enableBetaFeatures()
+  return false
```
sergiou87, line 78: Just curious: are we gonna remove this altogether or you're just hiding it until we have a proper "icon" for it?

app/src/ui/lib/radio-group.tsx
sergiou87: I'm a bit torn about this approach. I get what you mean about readability, but I also find it confusing because it doesn't reflect the actual hierarchy in the DOM after rendering. I do see why you did it, since it's not like a homogeneous list of items what we render, so using the `renderWhatever` approach is a bit more annoying, but I also think it's more React if that makes sense.

I think I prefer a `renderWhatever` approach, but I won't block the PR on that.

app/src/ui/preferences/appearance.tsx
```diff
@@ -112,15 +81,52 @@ export class Appearance extends React.Component<
...
+          </span>
+          <span>
+            <img src={darkThemeImage} alt="" />
+            <span className="theme-value-label">Dark</span>
+          </span>
+          {supportsSystemThemeChanges() && (
+            <span>
+              <span className="system-theme-swatch">
+                <img src={lightThemeImage} alt="" />
+                <img src={lightThemeImage} alt="" />
```
sergiou87: Do we need this image twice here? 🤔

app/styles/ui/_preferences.scss
```diff
@@ -79,6 +79,71 @@
...
+            position: absolute;
+            top: 0;
+
+            &:first-child {
+              position: inherit;
+            }
+
+            &:last-child {
+              clip-path: polygon(50% 0, 100% 0%, 100% 100%, 50% 100%);
+            }
```
sergiou87, line 141: Nice trick! 🤩

app/src/ui/preferences/appearance.tsx
tidy-dev: Yes. It is another trick, the first image is positioned relatively so that the label grows to the size of the image and the others are absolute so they can be clipped (and overlay the relatively positioned one).

app/src/lib/feature-flag.ts
tidy-dev: Meh.. I don't know. Right now it is because I don't have a proper icon, and I got the impression I was the only one onboard for this particular approach (particularly the customizable part)... and that maybe a better approach would be to just detect if the user has a high contrast system setting and apply accordingly. I think Markus dug into it a bit as part of the contrast accessibility issues work and on windows it was already surprisingly effective.  Thus, I figured to disable this until we have a consensus of an approach we would be willing to move forward into production with.

Other notes on this.. we currently have about 20 beta users that are making use of it which I find curious that that many would (of the small beta sample) unless they are just using it to have custom theme that is not necessarily high contrast.

app/src/ui/preferences/appearance.tsx
tidy-dev: Just added some documentation.

app/src/ui/lib/radio-group.tsx
tidy-dev: Done. Easy enough to change, and I was torn too so your vote for other approach was plenty enough to push me over. :D

app/src/lib/feature-flag.ts
niik: I put this on the agenda for our chat this afternoon so that we can discuss it but my take on this is we should remove it altogether until we have a decided upon way forward.

tidy-dev: Capturing team decision: We are going to pull out high contrast theme in favor of using system high contrast settings to customize desktop for high contrast users. Just changing flag in this PR, will open a follow up PR to fully remove.

sergiou87 (approved):
Code looks good and it works great! ✨ :shipit: ✨ 

The only thing that I would tweak is the size of the new "theme icons" to make them smaller, because right now this makes the `Appearance` section a bit taller than the rest, which feels like a step back from #16313 

But given we want to deploy a beta build with this, I think that change can be made in a different PR!

PR #16526, opened 2023-04-13
More autocompletion accessibility fixes

sergiou87:
Closes https://github.com/github/accessibility-audits/issues/3287
Closes https://github.com/github/accessibility-audits/issues/3246
Closes https://github.com/github/accessibility-audits/issues/3244

## Description
This work is a follow up of https://github.com/desktop/desktop/pull/16335 to address some accessibility issues that were reported after the PR was merged:
- The `Co-Authors` label right before the input should be a `label` element instead of a `div`.
- On Windows, NVDA wouldn't read the aria-live region with the number of suggestions the first time the autocomplete list is shown. I'd argue this is NVDA not behaving correctly, but the workaround was clean and easy enough: just move the aria-live region out of the auto-complete list and make it a sibling of the input, so it's always there.

## Release notes

Notes: [Fixed] NVDA reads number of suggestions when an autocompletion list shows up

### DIFF
--- a/app/src/ui/lib/author-input/author-input.tsx
+++ b/app/src/ui/lib/author-input/author-input.tsx
@@ -179,10 +179,14 @@ export class AuthorInput extends React.Component<
         <div className="sr-only" aria-live="polite" aria-atomic="true">
           {this.state.lastActionDescription}
         </div>
-        <div id="author-input-label" className="label">
-          Co-Authors&nbsp;
-        </div>
         <div className="shadow-input" ref={this.shadowInputRef} />
+        <label
+          id="author-input-label"
+          className="label"
+          htmlFor="added-authors"
+        >
+          Co-Authors&nbsp;
+        </label>
         {this.renderAuthors()}
         <AutocompletingInput<UserHit>
           placeholder="@username"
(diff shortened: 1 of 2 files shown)


app/src/ui/lib/author-input/author-input.tsx
```diff
@@ -207,6 +211,7 @@ export class AuthorInput extends React.Component<
   private renderAuthors() {
     return (
       <div
+        id="added-authors"
         className="added-author-container"
         ref={this.authorContainerRef}
         aria-labelledby="author-input-label"
```
niik, line 217: Wouldn't having a `label` with `for` mean that this is redundant?

niik (changes requested):
Just a small question

app/src/ui/lib/author-input/author-input.tsx
sergiou87: You would think so, but VoiceOver didn't read it, and I couldn't set a `label` without its `for` or otherwise we get a linter error 😞

niik: 🤦

niik (approved):

Sindhuja365 (commented):
Approve

PR #16527, opened 2023-04-13
Replace deprecated dependencies

niik:
## Description

I was trying to upgrade `xml2js` because we had a dependabot alert open for it but I couldn't because `azure-storage` required a specific (ancient) version of it. Investigating that I found that `azure-storage` itself was deprecated and so is `request` (which was a dependency of `azure-storage`).

So I ended up upgraded `azure-storage`, removing `request`, and upgrading `xml2js`.

I'm gonna do a test release from this branch to make sure the upload stuff works as intended I'd love a look from you @tidy-dev first since you have the most context around this blob storage stuff.

## Release notes

Notes:


tidy-dev (approved):
✨  Looks like it is accomplishing the same thing.

niik:
This seems to have worked just fine with my test deploy: https://github.com/desktop/desktop/actions/runs/4689112596/jobs/8310454418 :sparkles:

app/src/lib/file-system.ts
```diff
@@ -3,7 +3,6 @@ import * as Path from 'path'
 import { Disposable } from 'event-kit'
 import { Tailer } from './tailer'
 import byline from 'byline'
-import * as Crypto from 'crypto'
```
Alena32white, line 6: Write

